PR #2790, opened 2018-07-12
InputWithClearButton: Refactor clear button

agzuniverse:
Before:
search_before

After:
search_after

### DIFF
--- a/src/common/InputWithClearButton.js
+++ b/src/common/InputWithClearButton.js
@@ -6,18 +6,11 @@ import type { Context, LocalizableText } from '../types';
 import Input from './Input';
 import { BRAND_COLOR } from '../styles';
 import Icon from '../common/Icons';
-import Touchable from '../common/Touchable';
 
 const componentStyles = StyleSheet.create({
-  clearButtonContainer: {
-    alignItems: 'center',
-    justifyContent: 'center',
-    width: 22,
-    height: 22,
-    marginRight: 12,
-  },
   clearButtonIcon: {
     color: BRAND_COLOR,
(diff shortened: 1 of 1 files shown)


src/common/InputWithClearButton.js
```diff
@@ -14,10 +14,10 @@ const componentStyles = StyleSheet.create({
     justifyContent: 'center',
     width: 22,
     height: 22,
-    marginRight: 12,
   },
   clearButtonIcon: {
     color: BRAND_COLOR,
+    marginRight: 8,
```
borisyankov: When I look at this change, the first thing I am wondering is: Why is this value `8` if we had `12` before. Maybe there is a good reason but would be good if you mention it.

borisyankov: Even better though, since this is an element that is `touchable` it needs to be as big as possible. If you use `padding` instead of `margin` the touch area will increase, because the `padding` space is part of the component. A possible solution is to set the `height` and `width` to `22` of the `clearButtonInput` and don't use `padding`.

agzuniverse:
Thanks, redid the commit eliminating the `Touchable`. I had changed padding to `8` to match material design spec but on closer inspection of the docs ,`16` is the recommended value. Made that change as well.

gnprice:
Thanks @agzuniverse ! Merged -- this definitely makes the search bar look a lot nicer.

One important bit in the commit message: a "refactor" means it doesn't change how the app works on the outside. This isn't a refactor. :wink:

gnprice:
(also thanks @borisyankov for the original review!)

PR #2796, opened 2018-07-12
ModalSearchNavBar: Remove redundant code

agzuniverse:
This seems to be old logic left over from how search functioned.

### DIFF
--- a/src/nav/ModalSearchNavBar.js
+++ b/src/nav/ModalSearchNavBar.js
@@ -2,76 +2,33 @@
 import { connect } from 'react-redux';
 
 import React, { PureComponent } from 'react';
-import { View, StyleSheet } from 'react-native';
+import { View } from 'react-native';
 
 import type { Dispatch, Context, GlobalState } from '../types';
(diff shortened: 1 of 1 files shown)


gnprice:
Is this code really not doing anything now? Or if it is doing something, why is that redundant or unnecessary?

The history in the commit message is helpful, thanks -- and that might be the second question I'd ask, if it weren't there. But my first question is definitely how I can tell that this code really is redundant. :-)

(Or do you actually intend this to cause a change in the behavior of the app? In that case, I want to know more about that.)

gnprice:
Got it, thanks. Let's get that explanation into the commit. :-)

I think just putting that in the commit message would be OK. I'd explain it as: the props `searchBar` and `clearSearchInput` are never passed, so `showSearchInput` is always true, and a bunch of code disappears.

gnprice:
Thanks again @agzuniverse ! Sorry I didn't come back and merge this. If I had, it would have saved us some work over on #3163, where we eventually merged a very similar change as da6c43d4b .

PR #2798, opened 2018-07-12
timing: clarify identifier names.

roberthoenig:


roberthoenig:
Also added a commit for removing the misleading decimal places in the diagnostics timing screen.

src/diagnostics/TimeItem.js
```diff
@@ -25,10 +25,10 @@ export default class TimeItem extends PureComponent<TimingItemType> {
   props: TimingItemType;
 
   render() {
-    const { text, start, end } = this.props;
-    const startStr = format(start, 'HH:mm:ss.S'); // eslint-disable-line
-    const durationStr = numberWithSeparators((end - start).toFixed(2));
-    const timingStr = `Start: ${startStr}   Duration: ${durationStr} ms`;
+    const { text, startMs, endMs } = this.props;
+    const startStrMs = format(startMs, 'HH:mm:ss.S'); // eslint-disable-line
```
gnprice: What does it mean to call this `startStrMs` rather than `startStr`? It's no longer a number which needs out-of-band information about its units, but rather a string that identifies what's minutes, what's seconds, etc.

roberthoenig: My reasoning was that it is created by `format()`, which takes an ms argument. Now that I think about it, it doesn't make much sense, since that's already indicated by passing it the `startMs` parameter.

PR #2799, opened 2018-07-12
UnreadCards: Give visual feedback on pressing a topic

### LINKED ISSUE #2794: Add visual feedback on stream/topic button press.
Currently, when pressing on either a topic of a stream,  there is little to none visual feedback that this specific topic/stream was pressed:
ezgif com-optimize

Either there is only a quick green flash, or (when the narrow is fast enough) there is no feedback at all.

This makes it hard to be confident that the app recognized the press. It also makes it hard to be confident that the correct stream/topic has been pressed.

Two things to improve:
* The visual feedback should show as immediately as possible.
* The visual feedback should be visible for a couple hundred milliseconds, or until the narrow finishes (whatever is quicker).

agzuniverse:
Fixes #2794

### DIFF
--- a/src/unread/UnreadCards.js
+++ b/src/unread/UnreadCards.js
@@ -37,11 +37,11 @@ class UnreadCards extends PureComponent<Props> {
   };
 
   handleStreamPress = (stream: string) => {
-    this.props.dispatch(doNarrow(streamNarrow(stream)));
+    setTimeout(() => this.props.dispatch(doNarrow(streamNarrow(stream))));
   };
 
   handleTopicPress = (stream: string, topic: string) => {
-    this.props.dispatch(doNarrow(topicNarrow(stream, topic)));
+    setTimeout(() => this.props.dispatch(doNarrow(topicNarrow(stream, topic))));
   };
 
   render() {


src/unread/UnreadCards.js
```diff
@@ -42,7 +42,9 @@ class UnreadCards extends PureComponent<Props> {
   };
 
   handleTopicPress = (stream: string, topic: string) => {
-    this.props.dispatch(doNarrow(topicNarrow(stream, topic)));
+    setTimeout(() => {
```
jainkuniya: I thinks it will be better if we move this logic to `Touchable` component itself, as it will be then been applied to all the press throughout the app.

Thoughts? @borisyankov @gnprice @zulip/react-native-committers

borisyankov: I think we should even move it to `doNarrow` as grepping `dispatch(doNarrow` yields 24 results and many (all?) of them will benefit from giving a chance to the UI to respond.

I would extract everything from `dispatch(switchNarrow(narrow));` to the end of the file in a new function, maybe called something like `switchNarrowAndFetch()` and then call this function from `setTimeout`.

borisyankov: What I just suggested is quite enough, but as an additional reading, this is a great read (by the author of Redux):
https://stackoverflow.com/questions/35411423/how-to-dispatch-a-redux-action-with-a-timeout

gnprice:
Thanks @agzuniverse for this debugging! Glad such a simple change is enough to make it work :)

(Replying here to the discussion on the inline thread, because GitHub's UX is worse for inline threads than the main one and because this change is so small :tada: that the extra threading isn't needed.)

I agree with Boris that it'd be good to push this fix into the central `doNarrow`, so that we get the benefit of it in every place where we narrow. If you haven't already, I highly recommend reading the SO answer (really an essay / blog post / design doc, published in the form of an SO answer) that he linked -- a few weeks ago I found it to be a very helpful explanation of what's happening when our action creators (like `doNarrow`) return functions taking a Dispatch and GetState.

I'd go a step further: looking at the body of `doNarrow`, let's read through it and see which is the bit that is doing real work synchronously so that it doesn't return instantly. I think the answer is just the `dispatch(switchNarrow(narrow))` -- everything outside of the `dispatch` calls is just doing constant-time operations on data, and the `fetch*` actions already do their work asynchronously. So, let's make `switchNarrow` be asynchronous as well.

More generally, I think it'd be useful to have the following invariant be true:

> A `dispatch` call always returns quickly. If it causes something slow to happen, it does that asynchronously.

There may be a reason we can't apply that universally, but I think probably we can. And it'd help a lot in being able to reason about what will block the thread.

With that invariant in mind, one way of rephrasing your diagnosis of the cause of this bug is: `doNarrow` breaks the invariant. And then what I said above from reading the `doNarrow` implementation can be rephrased as: those `fetchMessages*` action creators are fine, but `switchNarrow` breaks the invariant, and that's the reason `doNarrow` breaks it.

I'm going to try to work up a fix at that level. I haven't previously made a change pushing through all the relevant layers here, so that will be useful for me to do before I try to help other people do so. :)

gnprice:
Hmm, OK, well: I made a series of refactors that cleaned up the code under `doNarrow`; see 513c0ca5b and its ancestors (e.g. `git log --stat -p 513c0ca5b`).

And then I tried this change, inside `doNarrow`:
```
-  dispatch(navigateToChat(narrow));
+  setTimeout(() => dispatch(navigateToChat(narrow)));
```
(That `navigateToChat` corresponds to `switchNarrow` in the discussion above, before my refactors.)

That helped -- but the touch feedback (in the Android emulator, on my powerful new desktop) was still kind of laggy. And the perf overlay ("Show Perf Monitor" in the RN in-app dev menu) showed us dropping a bunch of frames.

With this PR's fix, the feedback is still a bit laggy, and we still drop some frames, but it's better.

I also went and played around with other interactions in the app, and did those in the release build (current alpha, 15.0.92) that's on my Pixel 2 XL. We have a similar problem pretty badly on the "Create group" and "Search" buttons at the top of the PMs tab -- if you're lucky you might see a few belated frames of the feedback animation.

(The animation is kind of jerky in some other places too (including the "Switch" button on the settings tab, and a bit less so on the "Notifications" etc. items on the settings tab); but the new screen comes up very quickly, so it isn't the same kind of problem. Plus, the system settings app handles it about as well as our settings tab does, so that's a baseline.)

So I'm going to go ahead and merge this PR's fix, and I'll also go and make the same fix on those two buttons in PmConversationsCard.

We may want this in basically all our `onPress` handlers that do anything like navigate to a screen. If we can convince ourselves it's correct to put it *everywhere*, then we can put it in ZulipButton.js (like @jainkuniya suggested above) -- that'd help a lot to handle this cleanly.

PR #2802, opened 2018-07-13
docs/build-run: Troubleshooting note for Open JDK 8 not being default

punchagan:


gnprice:
Thanks @punchagan ! Merged.

I also did a bit more study to understand the context here; added a commit on top with some of that.

punchagan:
Thanks for the study and the updates! :+1:

Issue #2756, opened 2018-07-03
Add splash screen for before React has started up


prateek3255:
When the app loads it displays a white screen which seems a bit odd, so I guess you guys should add a splash screen. I would be happy do it for the android (since I don't have a mac) part using [this](https://medium.com/handlebar-labs/how-to-add-a-splash-screen-to-a-react-native-app-ios-and-android-30a3cec835ae) blog, because I have done it once before.

borisyankov:
That is true. A splash screen would look exactly as our current loading screen but without the spinning progress indicator. So a green screen with a centered Zulip logo.

prateek3255:
Ok, so I would get started with the android part.

prateek3255:
@zulipbot claim

zulipbot:
Hello @prateek3255, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

prateek3255:
@borisyankov I have already solved that issue #2705 and submitted a pull request #2743 for the same and I think that it's ready to be merged.

gnprice:
It'd definitely be good to add a splash screen, for the reasons said above.

Here's some notes on how to approach this, adapted from https://github.com/zulip/zulip-mobile/pull/2804#issuecomment-439520244 :

---

`@borisyankov` and I just took a quick look at [a third-party library for splash screens] -- and it's a pretty small amount of code. Here's the file that looks like the actual splash screen, on Android:
https://github.com/crazycodeboy/react-native-splash-screen/blob/master/android/src/main/java/org/devio/rn/splashscreen/SplashScreen.java

When the library is this small, our experience has been
* Like most things, it's probably not quite right.
* To fix the ways that it's not quite right, we're going to have to end up reading the respective Android and iOS documentation and figuring out how things are actually supposed to work.
* Once we do, it'll have been easier all around if we just write the code ourselves based on reading the docs :slightly_smiling_face: Because the library is so small, it's likely that the author themself didn't do much more than copy the relevant sample code from the platform docs + the RN docs, and put them together -- so we can do that instead.

[...]

PR #2804, opened 2018-07-13
added splash screen for android- #2756

prateek3255:
This PR is targeted to solve the issue #2756 

I updated a few files and added a new package `react-native-splash-screen` for adding splash screen for the android part( I couldn't do it for the ios part because I don't have a mac ) following [this](https://medium.com/handlebar-labs/how-to-add-a-splash-screen-to-a-react-native-app-ios-and-android-30a3cec835ae) blog.

Here is the gif on my device for the same -

zulip_splash


borisyankov:
I am wondering if we need to use `react-native-splash-screen`?
Android and iOS both have capabilities to add the splash screen.
I am always cautious when adding new dependencies because when they break or become incompatible with the latest React Native releases.

src/ZulipMobile.js
```diff
@@ -9,25 +9,33 @@ import CompatibilityChecker from './boot/CompatibilityChecker';
...
-            <StylesProvider>
-              <AppWithNavigation />
-            </StylesProvider>
-          </TranslationProvider>
-        </AppDataFetcher>
-      </AppEventHandlers>
-    </StoreProvider>
-  </CompatibilityChecker>
-);
+export default class ZulipMobile extends PureComponent<Props> {
```
jainkuniya: I guess changing this to 
```
export default class ZulipMobile extends PureComponent<void> 
```
will fix flow issue.

prateek3255:
@borisyankov this is probably the most easiest and best possible way to do it and this dependency is only required for the android part only because we need 2 splash screens there, 1 in native android code and the other one using the `react-native-splash-screen`, if we remove the other one there will be delay with a white screen between the android splash screen and the actual loaded app, you can find more details in [this](https://medium.com/handlebar-labs/how-to-add-a-splash-screen-to-a-react-native-app-ios-and-android-30a3cec835ae) blog. 

And `react-native-splash-screen` has around 2204 stars and the last commit was also only 3 days ago, so I don't think there will be any compatibility issue.

prateek3255:
Sorry for making a cluster of commits here, I am a newbie to open source and the CI tests, should I squash them into one?

gnprice:
Thanks @prateek3255 , and welcome! I haven't yet gotten a chance to review this code, but to your question about Git: our practice is to make each commit [clear and coherent](https://zulip.readthedocs.io/en/latest/contributing/version-control.html). So commits that just fix up test breakage in previous commits (like here, it looks like) should definitely be squashed in. On the other hand, when there are several logical changes that can be done separately, they're best as a series of separate commits.

Also, you can run the tests yourself before submitting -- no need to wait for Travis CI:
```
yarn test
```

prateek3255:
Can anyone please review my PR?

gnprice:
Thanks @prateek3255 for sending this!

@borisyankov and I just took a quick look at that library -- and it's a pretty small amount of code. Here's the file that looks like the actual splash screen, on Android:
https://github.com/crazycodeboy/react-native-splash-screen/blob/master/android/src/main/java/org/devio/rn/splashscreen/SplashScreen.java

When the library is this small, our experience has been
* Like most things, it's probably not quite right.
* To fix the ways that it's not quite right, we're going to have to end up reading the respective Android and iOS documentation and figuring out how things are actually supposed to work.
* Once we do, it'll have been easier all around if we just write the code ourselves based on reading the docs :slightly_smiling_face: Because the library is so small, it's likely that the author themself didn't do much more than copy the relevant sample code from the platform docs + the RN docs, and put them together -- so we can do that instead.

I'm going to close this PR, but we'd definitely like to have a splash screen for the reasons you say. If you'd like to take that on for Android -- reading Android and Android Studio docs, explaining what you find, and using Android Studio to make a splash screen -- I'd be glad to see that PR. :smile:

Issue #392, opened 2017-02-26
Use realm icons for quick mobile organization switcher


timabbott:
We just merged into master and deployed to chat.zulip.org https://github.com/zulip/zulip/pull/3740, which adds support for admin-uploaded realm icons.  Realm icons are the same size as avatars, 100x100.  

The URL is available in the /register request request, if we have `realm` in the `want` list, which I assume we already have since it contains useful settings data.

There are 2 fields that adds to the /register response:
* realm_icon_url (the URL for the realm's icon, all we should need).  This will always be present even if the user hasn't customized it (will fall back to gravatar), but only on the very latest server versions (effectively 1.6); we should put in some form of constant fallback I guess for older servers.
* realm_icon_source (needed only for the admin UI to change it, since this controls whether to show the "delete and go back to gravatar button")

borisyankov:
@timabbott can we get the `/favicon.png` type of alias?

timabbott:
yeah I can work on that next.

timabbott:
Actually, thinking about this a bit more, I'm not entirely sure that there's a purpose in creating that alias.  We need to cache several pieces of data about each realm:
* The server URL
* The realm icon URL
* The realm icon image itself (at least if we want to avoid needing Internet access to access the switcher)
* The email address and API key
* If we want to enhance the login screens the way we have on web, the realm's human-readable name (e.g. "The Zulip development community" and description.

Is there really benefit to having a convenience URL for getting the icon URL alone?  

It might instead make sense to add those extra fields to the payload returned by `/api/v1/get_auth_backends`, which is the thing that the app fetches from when a user enters a server URL.  What do you think of that approach?

timabbott:
Chatted with Boris briefly, and I think we're agreed to go with this approach.

I've implemented it as `GET /api/v1/server_settings`, soon to be merged and deployed to chat.zulip.org.  

An important detail is that there are 2 cases here:
* (1) The realm is known based on the URL, e.g. on a single-realm server (chat.zulip.org) or on a subdomains server (example.zulipchat.com)
* (2) The realm is not known based on the URL, e.g. a multi-realm server not using subdomains.  We will likely eventually deprecate this use case, since it's kinda a pain (or move it to require them using e.g. zulip.example.com/realm1, zulip.example.com/realm2 style login URLs).  But for now we need to support this.  In this case, the `realm_name` field will not be set.

Here's the (provisional) format for the structure, straight from our backend tests for the new endpoint:

```
[...]

PR #2805, opened 2018-07-13
WIP: account switcher: Add Realm Icons

punchagan:


borisyankov:
I think you might not be using `prettier` or you have configured it incorrectly.
The formatting shouldn't be jumping like that. Better figure it out early 👍

punchagan:
@borisyankov I was using `prettier` instead of `prettier-eslint` and that was causing some formatting to jump around, I fixed that now. Thanks! :smile:

borisyankov:
Nice. :)

One of the reasons I didn't add icons earlier is that there were several changes needed first, to best implement this. Some of the changes were made, some not.

I think the optimal approach is for you to continue working on this, while I do some of the remaining changes, then you finish up the work based on them.

Right now, our `Account` and `Auth` objects match 100% but that we can consider temporary. By adding icon information, we want to separate them. `Auth` should not have this property, while `Account` should. (I will do this)

Then we need to consider how and when we will update the icons if the organization administrator changes them.

I was also thinking about separately storing the `serverSettings` data for each account and getting the information via a selector.

chrisbobbe:
Hmm, this pull request looks a little old, but this would indeed be nice to add, as I don't currently see realm icons in master. 🙂

There are several conflicts, though, so we could either resolve them or throw out this PR and try again.

It appears that this (or a new PR) would fix https://github.com/zulip/zulip-mobile/issues/392 from early 2017.

chrisbobbe:
Closing as superseded by #4647.

PR #2807, opened 2018-07-14
Refactor baseSelector and bunch of navSelector.

jainkuniya:
* refactor: Rename `getChatScreenparams` to `getCurrentRouteNarrow`.  ee11b2f
This selector returns current route `params` if they are present, else
a object with `undefined` narrow.
Which is used at only one place to figure out to either play sound or
not on new message in `eventMiddleware`.
So `getCurrentRouteNarrow` looks better name for this instead of
`getChatScreenparams`.

* refactor: Move `getPrivateMessages` to `messageSelectors`.  8fabfa6
Before it was in `baseSelectors`, which is not a right place for it.
All selectors related to `message` are present in `messageSelectors`,
so it's better to move it.

* refactor: Move selectors from `baseSelectors` to `navSelectors`.  fd6d0b6
In continuation to previous commit, this selectors deals with
navigation state. So `navSelectors` looks better place for them.

* refactor: Rename `getCurrentRoute` to `getCurrentRouteName`.  570406e
As this selector, returns `routeName`, instead of route. So definately
`getCurrentRouteName` is more appropiate name for this.

* refactor: Use `createSelector` to create `getCurrentRouteName`. e8e6a24
Previously `routeName` was accessed directly, now `createSelector` is
used which will cache `routeName` whenever required. Thus, resulting
in better performance of the app and saving UI redundant repaints.


borisyankov:
Have a look at my comments of #8d647b2 
You don't need that commit.

Issue #2814, opened 2018-07-14
Show button in PasswordInput is aligned at bottom instead of right on Android API 28 only (checked on lower Android API & iOS, it works fine).


jainkuniya:
<img width="273" alt="screen shot 2018-07-14 at 9 03 47 pm" src="https://user-images.githubusercontent.com/18511177/42725929-d24ad006-87a9-11e8-9d64-7a4eab90d1fa.png">

@zulipbot claim

gnprice:
Thanks @jainkuniya for reporting this!

I believe it was fixed last month with Boris's 2fd7ba5ef.

PR #2815, opened 2018-07-14
layout: Align show button to right in PasswordInput on Android API 28.

jainkuniya:
Fixes: #2814.

Now

screen shot 2018-07-14 at 8 53 00 pm


borisyankov:
I am glad that this fixes the issue. 
Can you describe what change in API 28 made this break?

gnprice:
Closed because #2814 was fixed differently -- see there.

Issue #2677, opened 2018-06-13
Newly starred messages do not appear in the starred list.


jackrzhang:
To reproduce:

1. Star any message in any stream via long-press.
2. Navigate directly to the list of starred messages; the recently starred message does not appear.

This is likely connected to #2676.

jackrzhang:
@zulipbot claim

Issue #2676, opened 2018-06-13
A `starred` tag does not appear after starring a message in a stream.


jackrzhang:
To reproduce:

1. Star any message in any stream via long-press.
2. A tag containing `starred` text should appear below the message, but doesn't.

The same message does become marked within the web app, however. Also, if you long-press the same message again, the correct option 'Unstar message' is toggled.

This is on both Android and iOS.

jackrzhang:
@zulipbot claim

gnprice:
#2843 is a good fix which we should do (and it will put us ahead of the web app! see zulip/zulip#10111); but when the user did the starring/unstarring right here in this app in front of them, it will inevitably have too much latency for many (most?) users to feel responsive because it only gets to act when the server tells it about the event.

So we should also do a local echo of this state. The webapp gets away with doing this in a totally simple-minded way: it just updates its own state, then fires off a request to the server. See `toggle_starred` in `message_flags.js`, and its callers and callees.

That's not totally awesome behavior -- the request could fail, and then the user thought they'd safely stashed away that message in a way they'd remember to see it later, and we've let them down. And that case may come up a lot more often on mobile than on desktop. But doing better is certainly more complicated.

gnprice:
> That's not totally awesome behavior -- the request could fail, and then the user thought they'd safely stashed away that message in a way they'd remember to see it later

I just played a bit with reactions in FB Messenger, as one point of comparison to help gauge what our standards should be here. :wink: (Wait no, that's not on the FB "reaction" list. Maybe :open_mouth: ?) They did seem to be doing something, but it isn't great:
* I opened a thread, then went into airplane mode.
* Tried reacting to a message. The reaction appeared immediately. ... After I scrolled up a bit (scrolling that message offscreen), then scrolled back down, it was gone.
* Tried again. Curious how long it takes to disappear, so this time, stared at it for like 10 seconds... and it stayed. Scrolled up and down again, and again now it disappeared.
* Now I can't do any reactions at all; they just silently don't appear. Maybe it's decided it lacks a connection, and stopped doing local feedback.
* Minutes later (still in airplane mode), reactions work again. It really is about scrolling offscreen; I left one onscreen for a while and it stayed, then just dipped the message offscreen for the briefest time I could, and it was gone. And after that, it's back to not accepting reactions.
* OK, *now* when I open it again, there's a red "Airplane Mode On" banner. That wasn't there before -- I'd been surprised not to see it.
[...]

Issue #2627, opened 2018-06-04
message list: Show when a message is starred


gnprice:
In the message list, we should display an indication of whether a message has been starred. For consistency and ease of understanding, it should look a lot like what we do in the webapp, which is a "star" icon in the upper-right of the message.

This will fill in a significant UX gap in the interaction of starring a message, which is that there's basically no feedback to confirm to the user that it worked.

jackrzhang:
I'd be interested in trying to implement this.

@gnprice Can I claim this?

gnprice:
@jackrzhang Sure! You should be able to do it with zulipbot.

borisyankov:
We do support that actually. It is shown as a `starred` text at the bottom of the message.
Possibly an actual star will be an improvement.

borisyankov:
@jackrzhang The relevant file is:
https://github.com/zulip/zulip-mobile/blob/master/src/webview/html/messageTagsAsHtml.js

Another argument in favor of using an actual star - we are not currently translating this.

gnprice:
> We do support that actually. It is shown as a starred text at the bottom of the message.

I just tried this to make sure I wasn't imagining things, and I don't see it. So there's some other bug, then.

In any case, I think an actual star will be clearer. Good to know about that bit of code, though!

jackrzhang:
I see the `starred` tags, but they seem to be really unresponsive. When I star/unstar a message in the application, the menu that pops up shows the correct action - either 'Star message' or 'Unstar message' - but the message tag itself doesn't update.

Agreed, I think a visual star would be more intuitive - so for now, I'll just remove the `starred` tag.

@zulipbot claim
@zulipbot add "area: layout"

jackrzhang:
As discussed with @gnprice, I think it would be good to debug the behavior of the message tags first before moving forward with this. Will open a separate issue for that investigation.

jackrzhang:
Opened #2676 / #2677.

jackrzhang:
@gnprice I think this should have been closed with the resolution #2627 and #2677.

gnprice:
I've just filed #3593 for the remaining task of converting the design to an actual star. Closing this one -- thanks @jackrzhang for fixing these bugs!

PR #2818, opened 2018-07-18
Clean up type checking related to renderMessages.js.

jackrzhang:
My initial goal was to add a type definition to annotate the `renderedMessages` prop passed from `MessageList` into `MessageListWeb`, which caused a number of flow error messages leading to this series of changes.

The data flow around this area is buggy - see #2627, #2676, and #2677 - so hopefully this elucidates the current structure of the subsystem and provides insight into any possibly unintended behavior.

### DIFF
--- a/src/message/__tests__/renderMessages-test.js
+++ b/src/message/__tests__/renderMessages-test.js
@@ -2,13 +2,14 @@ import deepFreeze from 'deep-freeze';
 
 import { homeNarrow } from '../../utils/narrow';
 import renderMessages from '../renderMessages';
+import { NULL_MESSAGE } from '../../nullObjects';
 
 describe('renderMessages', () => {
   const narrow = deepFreeze([]);
 
   test('empty messages results in a single empty section', () => {
     const messageList = renderMessages([], homeNarrow);
-    expect(messageList).toEqual([{ key: 0, message: {}, data: [] }]);
+    expect(messageList).toEqual([{ key: 0, message: NULL_MESSAGE, data: [] }]);
   });
 
   test('renders time, header and message for a single input', () => {
--- a/src/types.js
+++ b/src/types.js
@@ -492,22 +492,25 @@ export type GetState = () => GlobalState;
 
 export type LocalizableText = any; // string | { text: string, values: Object };
 
-export type RenderedTimeDescriptor = {
+type RenderedTime = {
   type: 'time',
+  key: string,
   timestamp: number,
+  firstMessage: Message,
 };
 
-export type RenderedMessageDescriptor = {
+type RenderedMessage = {
   type: 'message',
-  message: Object,
+  key: number,
+  isBrief: boolean,
+  message: Message,
 };
 
-export type RenderedItemDescriptor = any; // MessageDescriptor | TimeDescriptor;
-
-export type RenderedSectionDescriptor = any; /* {
-  message: Object,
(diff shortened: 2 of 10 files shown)


src/types.js
```diff
@@ -489,18 +489,6 @@ export type GetState = () => GlobalState;
 
 export type LocalizableText = any; // string | { text: string, values: Object };
 
-export type RenderedTimeDescriptor = {
```
borisyankov: I am not sure you need to delete these items in a separate commit (even though they are indeed not used)
In the currently last commit, you are reintroducing them under slightly different names.

Just use the last commit to replace these and skip this commit altogether.

src/types.js
```diff
@@ -111,6 +111,7 @@ export type Message = {
   id: number,
   isOutbox: boolean,
   is_me_message: boolean,
+  last_edit_timestamp: number,
```
borisyankov: This prop is optional.

src/types.js
```diff
@@ -54,6 +54,11 @@ export type Account = Auth;
  *
  * See also EventReaction, which carries additional properties computed from
  * these.
+ *
+ * To compute emoji_code from emoji_name, use the mapping exported
+ * by `src/emoji/codePointMap`.
+ *
+ * TODO: Add note on how to determine `reaction_type`.
```
borisyankov: Maybe not leave the TODO in the commit, but do describe the `reaction_type` computation.

src/message/__tests__/renderMessages-test.js
```diff
@@ -2,13 +2,14 @@ import deepFreeze from 'deep-freeze';
 
 import { homeNarrow } from '../../utils/narrow';
 import renderMessages from '../renderMessages';
+import { NULL_MESSAGE } from '../../nullObjects';
```
borisyankov: Great! I like this commit a lot :)

src/types.js
borisyankov: If the `EditMessage` type does not make sense, just make the prop optional with `?`.

jackrzhang: Hm, okay. I'll look into the code paths for editing messages.

gnprice:
Thanks for these cleanups!

One thing I'd like us to do, while we're here studying these and figuring out what these types really end up being: I'd like to record a way that the next person to look at the code can look up for themselves what these are supposed to be, and also can go check if there've been changes on the server.

Ideally this would be API documentation for Zulip. But there isn't yet much in the way of that, and not for the stuff like `Message` where we really most need it. So the jsdoc on some of these types is going to be a little more complicated than that.

I think we still want it, for a lot of these types, to look like: this represents the data from such-and-such events the server sends. Then instead of docs, we'll just have to say where you look in the server code to find that.

(This thought comes especially to mind on the first commit in this series, with the `last_edit_timestamp` addition and the `timeEdited` change.)

gnprice:
BTW I spent a bit of time on this earlier today; and I've just pushed 563c9ae4e, which adds a big block comment on `Message` and some notes on specifics.

Still more to say on many more specifics! Like that `last_edit_timestamp`. Given the framework suggested in that block comment, on a commit like that I'd like to see a word about how you identified that this was the right type.

gnprice:
Thanks again @jackrzhang for this PR!

At this point it looks like the improvements this made have all gotten merged in one way or another, so closing.

PR #2819, opened 2018-07-18
WIP: Enhance API functions.

borisyankov:
This is work towards making our API calls better covered by types, more complete and better documented.

A secondary goal is to make them more self-contained and less coupled to the rest of the project.


src/api/apiTypes.js
```diff
@@ -1,11 +1,185 @@
...
   result: string,
   msg: string,
 };
 
+export type ApiResponseSuccess = {
+  result: 'success',
+  msg: '',
+};
+
+export type ApiErrorCode =
```
gnprice: This list is from the `ErrorCode` enum in the server's `zerver/lib/exceptions.py`, I think.

It'd be good to mention that in a short jsdoc comment.

src/api/users/getUsers.js
```diff
@@ -1,6 +1,14 @@
 /* @flow */
-import type { Auth, User } from '../../types';
+import type { Auth, ApiResponseSuccess, User } from '../apiTypes';
 import { apiGet } from '../apiFetch';
 
-export default (auth: Auth): Promise<User[]> =>
-  apiGet(auth, 'users', res => res.members, { client_gravatar: true });
+// Documentation at https://zulipchat.com/api/get-all-users
```
gnprice: This looks like prime material for jsdoc. How about making it a jsdoc comment on the exported function?

Doesn't need to be any longer than this, or
```
/** See https://zulipchat.com/api/get-all-users */
```

src/api/messages/getMessages.js
```diff
@@ -1,20 +1,29 @@
...
   narrow: Narrow,
   anchor: number,
   numBefore: number,
   numAfter: number,
   useFirstUnread: boolean = false,
-): Promise<Message[]> =>
-  apiGet(auth, 'messages', res => res.messages, {
+  applyMarkdown: boolean = true,
+): Promise<ApiResponseMessages> =>
+  apiGet(auth, 'messages', (res: ApiResponseMessages): ApiResponseMessages => res, {
```
gnprice: It'd be nice to get to leave out all the repetition in this `resFunc` argument.

What if you write just `res => res` -- or perhaps `undefined`, which gets you the default which is the same thing? We have a type signature on the outer function already, so its own type should come out just as clearly, right?

src/api/registerForEvents.js
```diff
@@ -1,13 +1,25 @@
...
-export default (auth: Auth) =>
+// Documentation at https://zulipchat.com/api/register-queue
+
+export default (
+  auth: Auth,
+  eventTypes: string[],
+  fetchEventTypes: string[],
+  applyMarkdown: boolean = true,
+  includeSubscribers: boolean = false,
+  clientGravatar: boolean = false,
```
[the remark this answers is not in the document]
borisyankov: This looks like a mistake. I will use `true` instead.

src/api/messages/getMessages.js
borisyankov: I was thinking something similar. I will remove each such function in one function per commit.

src/api/apiTypes.js
```diff
@@ -32,3 +206,8 @@ export type ApiServerSettings = {
   require_email_format_usernames: boolean,
   zulip_version: string,
 };
+
+export type ApiRegisterForEventsResponse = {
```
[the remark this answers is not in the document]
borisyankov: Indeed, part of another change. It will also be in the `register.js` file. Removed.

borisyankov:
As GitHub does not properly notify the concerned parties, I am clarifying that this PR has been revisited and the issues originally raised were resolved. Also many other, mostly straightforward commits were added.

gnprice:
@borisyankov Ah, thanks for commenting :) -- indeed, that's a classic GitHub gotcha. Reading now!

src/api/events/registerForEvents.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default (
+  auth: Auth,
+  eventTypes: string[],
+  fetchEventTypes: string[],
+  applyMarkdown: boolean = true,
+  includeSubscribers: boolean = false,
+  clientGravatar: boolean = true,
+  allPublicStreams: boolean = false,
+  narrow: Narrow = [],
```
gnprice: This doesn't actually match the server API, according to the doc:

> A JSON-encoded array of length 2 indicating the narrow for which you'd like to receive events for. For instance, to receive events for the stream Denmark, you would specify `narrow=['stream', 'Denmark']`. Another example is `narrow=['is', 'private']` for private messages. Defaults to `"narrow=[]"`.

whereas our `Narrow` is

```
export type NarrowElement = {
  operand: string,
  operator?: NarrowOperator, // TODO type: this shouldn't be absent.
};

export type Narrow = NarrowElement[];
```

Better to leave this out than to get it wrong, and I think it's complex enough not to be worth trying to throw in until there's a use case. Ideally include a comment mentioning that there is such a thing and we just haven't wrapped it, e.g.:
```
  allPublicStreams: boolean = false,
  // also `narrow`, not currently wrapped here because it's a bit complicated
  // and we don't currently have a use for it.
```

src/api/events/registerForEvents.js
```diff
@@ -0,0 +1,32 @@
...
+  last_event_id: number,
+  data: InitialData,
+};
+
+export default (
+  auth: Auth,
+  eventTypes: string[],
+  fetchEventTypes: string[],
+  applyMarkdown: boolean = true,
+  includeSubscribers: boolean = false,
```
gnprice: This and `allPublicStreams` don't match the server defaults (according to the doc.) That's fine for now, but let's clearly mark that in jsdoc -- otherwise it's the kind of thing that could give someone a nasty surprise (or frustrating debugging session) when they look at the API doc, then use this because they believe it's a thin wrapper around the server API, and it doesn't do as they expect.

src/api/events/registerForEvents.js
```diff
@@ -0,0 +1,32 @@
+/* @flow */
+import type { Auth, ApiResponseSuccess, Narrow } from '../apiTypes';
+import type { InitialData } from '../../types';
+import { apiPost } from '../apiFetch';
+
+/** See https://zulipchat.com/api/register-queue */
```
gnprice: This jsdoc winds up attached to ApiResponseRegister, rather than registerForEvents.

Let's keep jsdoc immediately above what it documents, without a blank line. I think the blank line helps make it easy for something else to get inserted in between, as happens with ApiResponseRegister over the course of this branch.

src/api/events/registerForEvents.js
gnprice: More explicitly:
* For the commit that adds these options, keeping the values we're currently passing (which are different from the default) is absolutely the right thing; it just needs to be called out in jsdoc.
* It'd be a good followup (in a separate commit) to make this function take the same defaults as the server API, and adjust the caller(s) to pass the desired non-default values. Doesn't have to be in the same branch, though.

src/api/apiTypes.js
```diff
@@ -1,11 +1,234 @@
...
+  | 'INVITATION_FAILED'
+  | 'INVALID_ZULIP_SERVER';
+
+export type ApiResponseError = {
+  code?: ApiErrorCode,
+  msg: string,
+  result: 'error',
+};
+
+export type ResponseExtractionFunc = (response: Object) => any;
```
gnprice: I appreciate the more specific type here (it was `any`), but given that the rest of this commit is a bulky mechanical change, I'd rather not have it in the same commit.

(This is small enough, and so long as the type-checker still passes the consequences are bounded enough, that it won't stop me from merging the commit as is, though.)

src/api/users/getUsers.js
```diff
@@ -1,6 +1,14 @@
...
-export default (auth: Auth): Promise<User[]> =>
-  apiGet(auth, 'users', res => res.members, { client_gravatar: true });
+/** See https://zulipchat.com/api/get-all-users */
+
+type ApiResponseUsers = ApiResponseSuccess & {
+  members: User[],
+};
+
+export default (auth: Auth, clientGravatar: boolean = true): Promise<ApiResponseUsers> =>
+  apiGet(auth, 'users', (res: ApiResponseUsers) => res, {
```
gnprice: This changes the API this function provides to callers, right -- won't that break something?

... Aha, it doesn't -- because this function has no callers.

Good then, but that's important information for the reader in the commit message. :-)

gnprice:
Oh also, in a commit message you wrote:

> The response often contains a program-readable `code` constant, but not always. In future, the back-end will likely have this set for all error responses.

I've deleted the second sentence, because it looks now like that won't be true. That was my intention when I added the `code` field over a year ago, but Steve started a conversation in chat recently (on #backend? some public stream) about it, and following that conversation I think we're all agreed this isn't especially likely.

(Arguably for a really top-quality API this is among the things we *should* do, but there's a lot of other things I'd want to spend effort on first -- and this is the kind of thing where mandating it without putting in the sustained effort to make it high-quality could be worse than not doing it at all.)

src/message/fetchActions.js
```diff
@@ -168,8 +168,8 @@ export const fetchEssentialInitialData = () => async (dispatch: Dispatch, getSta
   timing.end('Essential server data');
 
   dispatch(realmInit(initData));
-  if (narrow && messages) {
-    dispatch(messageFetchComplete(messages, narrow, 0, halfCount, halfCount));
+  if (narrow && messagesResponse) {
```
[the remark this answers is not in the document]
borisyankov: Good catch! I didn't consider that this is (potentially) changing the behavior.

But actually, the initial behavior is not really correct. It only checks for `messages` being `undefined` (it is true if `messages` is an empty array). But `tryUntilSuccessful` would not stop retrying until successful. If so it will always return `truthy` object.

The correct change is to remove the second check altogether.

src/api/messages/getMessages.js
borisyankov: #2933

src/api/users/getUsers.js
borisyankov: I did update the message.

gnprice:
(I see you've rebased, but it looks like you haven't yet addressed most of the feedback; as usual, ping with a comment when you have.)

gnprice:
Meanwhile: I just finished reading the rest of the branch, and merged 5 more commits, including the one to optionally skip the `Authorization` header. The followup to unify our code that does plain `fetch`es should be a nice improvement in itself -- good thought!

Most of the other commits in the later part of the branch look fine, except for a couple of the same comments as I made above on earlier commits.

One exception I notice: in the commit "api: Improve `deleteMessage` function", the fact that it changes the actual behavior of the app is very interesting! I'm happy the commit message mentions that. :smile:  I'd like the commit message to do two things further:
* The summary line should sound like an app-level behavior change; as is, it sounds like a bland refactor, just like "Improve `getMessages` API call" which just adds a parameter and a clear type. One version would be "api: Send a real DELETE request in `deleteMessage`."
* I'm curious to know more: how does the behavior change? Perhaps equivalently: how does the current behavior differ from the web app? Whatever the answer, the commit message should end up containing the answer.
   * One thing I happen to know from looking at this recently: there's two kinds of "delete". One is to just set the content to empty, which the server has a special case to translate to the 9-character string "(deleted)", and I believe this is what ordinary users "deleting" their messages in the web app end up doing. The other actually deletes the message so it no longer shows up at all, and typically only org admins have the power to do that one; it's intended for moderation use cases like dealing with spam, confidential info, offensive remarks, etc. Ultimately we should probably provide both for users that can do both.

borisyankov:
This turns into 'WIP' as I am extracting stuff out of it, and rebasing, but not necessarily making it ready-to-merge as-is.

PR #2820, opened 2018-07-19
tests: Make eslint error on warnings.

roberthoenig:
Do this to make the CI fail on improperly formatted code, straying
`console.log()`s, etc. that might not be detected by test:prettier.


gnprice:
Good thought, thanks @roberthoenig -- merged! We've definitely had things slip in occasionally because the warnings were easy to miss.

PR #2821, opened 2018-07-19
Replace `require()` with `import` where possible.

roberthoenig:
The `import` command from `ES6` gets transpiled by Babel to `require()`.
Babel is enabled for the src/ directory. Since we use `import` wherever
Babel is enabled, replace the few stray instances of `require()` in src/.


gnprice:
Thanks @roberthoenig for the cleanup and @borisyankov for the detailed (on the scale of the PR's size :) ) review -- merged!

PR #2822, opened 2018-07-19
redux-persist: Add compression.

roberthoenig:
Compress the state to mitigate storage limitations imposed by AsyncStorage
on Android.

In particular, AsyncStorage on Android uses either RocksDB or
SQLite. A limit of 6MB is set for both databases. If AsyncStorage uses
SQLite, the rows have a maximum length of ~2MB. redux-persist stores each
top-level reducer in a separate row. This means that big reducers such as
`users` can hit the storage limit in large orgs. When this happens, the
whole storage gets cleared.

On iOS, AsyncStorage uses a designated key-value storage with no size
limits.

Do not use the compression transformer redux-persist-transform-compress for
redux-persist, because it implements the compression algorithm in JS which
makes it slow and adds serious lag to the app.

Instead, implement a native module to do the compression in Java. Then,
wrap RN's AsyncStorage into CompressedAsyncStorage. CompressedAsyncStorage
overwrites AsyncStorage's getItem() and setItem() methods to compress and
decompress state that gets written to and read from the storage. Note that
we cannot do this with a transformer for redux-persist, because
transformers don't support async operations. Loading and compressing state
can take enough time to cause frame drops, so we need it to happen async.


src/boot/CompressedAsyncStorage.js
```diff
@@ -0,0 +1,46 @@
...
+import { AsyncStorage, NativeModules } from 'react-native';
+
+export default class CompressedAsyncStorage {
+  // HACK: This wrapper around AsyncStorage.getItem() is not perfectly compliant with the
+  // specification of AsyncStorage.getItem(): `callback` should be optional, and getItem()
+  // should return the result. Here, we declare `callback` as non-optional, because redux-persist
+  // always calls getItem() with a callback. redux-persist is the only library that uses
+  // CompressedAsyncStorage. Also, we don't return the result, because redux-persist never uses it
+  // (instead, it looks at the result passed to `callback`). This avoids decompressing the result
+  // twice.
```
gnprice: What if we say something like
```
const outerResult = await NativeModules.TextCompressionModule.decompress(result);
callback(error, outerResult);
return outerResult;
```
?

I think that'd let us return the result without extra work.

If we really can't, then that's fine, though.

src/boot/CompressedAsyncStorage.js
```diff
@@ -0,0 +1,46 @@
...
+  static async getItem(key: string, callback: (error: ?Error, result: ?string) => void) {
+    async function wrappedCallback(error, result) {
```
gnprice: I think this would actually be clearer by making this wrapper anonymous, like so:
```
await AsyncStorage.getItem(key, (error, result) => {
  // ... body of wrappedCallback ...
});
```

I was a bit confused at first reading this, until I saw below where this is passed to the underlying `getItem`. Swapping the order of presentation better matches the logical flow, I think.

(This is something that is awkward to do in Python, unfortunately, so we end up putting things in this inverted order a lot.)

src/boot/CompressedAsyncStorage.js
```diff
@@ -0,0 +1,46 @@
...
+      // n - null
+      if (!result.startsWith('e')) {
+        callback(error, result);
+        return;
+      }
+      callback(error, await NativeModules.TextCompressionModule.decompress(result));
+    }
+    await AsyncStorage.getItem(key, wrappedCallback);
+  }
+  static async setItem(key: string, value: string, callback: ?(error: ?Error) => void) {
```
gnprice: Style nit: put a blank line between each of these methods. That makes it easier to visually scan.

android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
```diff
@@ -0,0 +1,81 @@
...
+      while (!deflater.finished()) {
+        int byteCount = deflater.deflate(buffer);
+        outputStream.write(buffer, 0, byteCount);
+      }
+      deflater.end();
+      outputStream.close();
+      // The RN bridge currently doesn't support sending binary data, so send the compressed
+      // output as a string. Also, out of unknown reasons, using ISO-8859-1 to directly encode
+      // the output as a string doesn't work. Encoding it as a Base64 string does work, but it
+      // is unclear why. Curiously, Base64.encodeToString itself uses ISO-8859-1.
```
gnprice: Style nit: for good readability, prose should be wrapped at a narrower width than the maximum line length allowed for code.

About 70 columns is a good limit for the width of prose. Comment lines generally start a bit indented, so you can go past column 70 of the line; in VS Code, I set
```
    "editor.rulers": [76, 100],
```
to wrap text at 76 total columns. This comment starts 9 columns in, so going as deep as 79 would be reasonable.

Same principle applies to the block comments in the JS file.

android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
```diff
@@ -0,0 +1,81 @@
...
+    } catch (IOException e) {
+      promise.reject("IO_EXCEPTION", e);
+    }
+  }
+
+  @ReactMethod
+  public void decompress(String input, Promise promise) {
+    try {
+      Inflater inflater = new Inflater();
+      inflater.setInput(Base64.decode(input.getBytes("ISO-8859-1"), Base64.DEFAULT));
```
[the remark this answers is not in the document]
roberthoenig: Hrmm, okay. But does it make reasoning really easier for you? I find it super helpful to be reassured by this encoding that it'll just work with any bytes that get thrown at it, so I don't need to worry about it at all. For ASCII, I first need to know that the bytes generated by Base64 happen to make up a valid ASCII string.

src/boot/CompressedAsyncStorage.js
roberthoenig: Ah, right, that'd work!

src/boot/CompressedAsyncStorage.js
roberthoenig: So we could do that, but then we couldn't apply the solution of your comment above anymore:
```js
const outerResult = await NativeModules.TextCompressionModule.decompress(result);
callback(error, outerResult);
return outerResult;
```

roberthoenig: I think we should favor matching the real `getItem()` in behavior over the readability aspect.

roberthoenig: Oh wait, actually, the solution in your comment above makes creating a wrapper function unnecessary, so we won't even have the readability problem :)

src/boot/CompressedAsyncStorage.js
```diff
@@ -0,0 +1,46 @@
...
+  // twice.
+  static async getItem(key: string, callback: (error: ?Error, result: ?string) => void) {
+    async function wrappedCallback(error, result) {
+      // It's possible that getItem() is called on uncompressed state (e.g. when a user updates their
+      // app from a version without compression to a version with compression). Detect if the stored
+      // state is compressed, and return it unmodified if it isn't. Detect a compressed state by
+      // inspecting the first byte of the string. TextCompressionModule uses zlib compression and returns
+      // compressed strings that start with two bytes that specify the compression method. The
+      // bytes then get base64-encoded and turn into a string starting with 'e'. Luckily, uncompressed
+      // state must be valid JSON and hence cannot start with an e. The only possible JSON start characters
```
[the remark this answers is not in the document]
roberthoenig: Cool, makes sense.

One question: What is the "style of `mount`"? Are you referring to the unix `mount` command?

roberthoenig:
Updated. Here is the diff of changes since @gnprice commented:

```diff
diff --git a/android/app/src/main/java/com/zulipmobile/TextCompressionModule.java b/android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
index d1effabc..b2ef97fd 100644
--- a/android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
+++ b/android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
@@ -11,6 +11,8 @@ import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.zip.DataFormatException;
 import java.util.zip.Deflater;
 import java.util.zip.Inflater;
@@ -28,7 +30,16 @@ class TextCompressionModule extends ReactContextBaseJavaModule {
   }
 
   // TODO: Experiment what value gives the best performance.
-  private int bufferSize = 8192;
+  private final int bufferSize = 8192;
+
+  private final String header = "z|base64 zlib|";
+
+  @Override
+  public Map<String, Object> getConstants() {
+    final Map<String, Object> constants = new HashMap<>();
+    constants.put("header", header);
+    return constants;
+  }
 
   @ReactMethod
   public void compress(String input, Promise promise) {
@@ -44,11 +55,13 @@ class TextCompressionModule extends ReactContextBaseJavaModule {
       }
       deflater.end();
       outputStream.close();
-      // The RN bridge currently doesn't support sending binary data, so send the compressed
-      // output as a string. Also, out of unknown reasons, using ISO-8859-1 to directly encode
-      // the output as a string doesn't work. Encoding it as a Base64 string does work, but it
-      // is unclear why. Curiously, Base64.encodeToString itself uses ISO-8859-1.
-      promise.resolve(Base64.encodeToString(outputStream.toByteArray(), Base64.DEFAULT));
+      // The RN bridge currently doesn't support sending binary data, so
+      // send the compressed output as a string. Also, out of unknown
+      // reasons, using ISO-8859-1 to directly encode the output as a string
+      // doesn't work. Encoding it as a Base64 string does work, but it is
+      // unclear why. Curiously, Base64.encodeToString itself uses
+      // ISO-8859-1.
+      promise.resolve(header + Base64.encodeToString(outputStream.toByteArray(), Base64.DEFAULT));
     } catch (UnsupportedEncodingException e) {
       promise.reject("UNSUPPORTED_ENCODING_EXCEPTION", e);
     } catch (IOException e) {
@@ -60,7 +73,11 @@ class TextCompressionModule extends ReactContextBaseJavaModule {
   public void decompress(String input, Promise promise) {
     try {
       Inflater inflater = new Inflater();
-      inflater.setInput(Base64.decode(input.getBytes("ISO-8859-1"), Base64.DEFAULT));
+      byte[] inputBytes = input.getBytes("ISO-8859-1");
+      inflater.setInput(Base64.decode(inputBytes,
+                                      header.length(),
+                                      inputBytes.length - header.length(),
+                                      Base64.DEFAULT));
       ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
       byte[] buffer = new byte[bufferSize];
       while (inflater.getRemaining() != 0) {
diff --git a/src/boot/CompressedAsyncStorage.js b/src/boot/CompressedAsyncStorage.js
index eb6820d5..b7026830 100644
--- a/src/boot/CompressedAsyncStorage.js
+++ b/src/boot/CompressedAsyncStorage.js
@@ -2,38 +2,19 @@
 import { AsyncStorage, NativeModules } from 'react-native';
 
 export default class CompressedAsyncStorage {
-  // HACK: This wrapper around AsyncStorage.getItem() is not perfectly compliant with the
-  // specification of AsyncStorage.getItem(): `callback` should be optional, and getItem()
-  // should return the result. Here, we declare `callback` as non-optional, because redux-persist
-  // always calls getItem() with a callback. redux-persist is the only library that uses
-  // CompressedAsyncStorage. Also, we don't return the result, because redux-persist never uses it
-  // (instead, it looks at the result passed to `callback`). This avoids decompressing the result
-  // twice.
   static async getItem(key: string, callback: (error: ?Error, result: ?string) => void) {
-    async function wrappedCallback(error, result) {
-      // It's possible that getItem() is called on uncompressed state (e.g. when a user updates their
-      // app from a version without compression to a version with compression). Detect if the stored
-      // state is compressed, and return it unmodified if it isn't. Detect a compressed state by
-      // inspecting the first byte of the string. TextCompressionModule uses zlib compression and returns
-      // compressed strings that start with two bytes that specify the compression method. The
-      // bytes then get base64-encoded and turn into a string starting with 'e'. Luckily, uncompressed
-      // state must be valid JSON and hence cannot start with an e. The only possible JSON start characters
-      // are:
-      // { - Object
-      // [ - Array
-      // 0123456789 - number
-      // " - string
-      // f - false
-      // t - true
-      // n - null
-      if (!result.startsWith('e')) {
-        callback(error, result);
-        return;
-      }
-      callback(error, await NativeModules.TextCompressionModule.decompress(result));
+    let result = await AsyncStorage.getItem(key);
+    // It's possible that getItem() is called on uncompressed state, for
+    // example when a user updates their app from a version without
+    // compression to a version with compression. Detect if the stored state
+    // is compressed, and return it unmodified if it isn't.
+    if (result.startsWith(NativeModules.TextCompressionModule.header)) {
+      result = await NativeModules.TextCompressionModule.decompress(result);
     }
-    await AsyncStorage.getItem(key, wrappedCallback);
+    callback(undefined, result);
+    return result;
   }
+
   static async setItem(key: string, value: string, callback: ?(error: ?Error) => void) {
     await AsyncStorage.setItem(
       key,
@@ -41,6 +22,8 @@ export default class CompressedAsyncStorage {
       callback,
     );
   }
+
   static getAllKeys = AsyncStorage.getAllKeys;
+
   static clear = AsyncStorage.clear;
 }
```

android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
```diff
@@ -0,0 +1,98 @@
...
+    return "TextCompressionModule";
+  }
+
+  // TODO: Experiment what value gives the best performance.
+  private final int bufferSize = 8192;
+
+  private final String header = "z|base64 zlib|";
+
+  @Override
+  public Map<String, Object> getConstants() {
```
roberthoenig: The nice thing about getConstants is that React uses it to store the constants directy in JS and thus avoids making a roundtrip over the bridge when requesting the constants.

android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
```diff
@@ -0,0 +1,98 @@
...
+    return "TextCompressionModule";
+  }
+
+  // TODO: Experiment what value gives the best performance.
+  private final int bufferSize = 8192;
+
+  private final String header = "z|zlib base64|";
+
+  @Override
+  public Map<String, Object> getConstants() {
```
roberthoenig: The nice thing about getConstants is that React uses it to store the constants directy in JS and thus avoids making a roundtrip over the bridge when requesting the constants.

roberthoenig:
Updated:

* Added the `z|TRANSFORM|DATA` format.
* Updated the `getItem()` logic.
* Added newlines between function declarations and shortened my comments.

* Didn't change "ASCII" to "ISO-8859-1" -- see my comment.

android/app/src/main/java/com/zulipmobile/TextCompressionModule.java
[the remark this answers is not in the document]
gnprice: > returns a string that is constructed with the ISO-8859-1 charset

Hmm, what do you mean -- is there something in the chain there that describes it as ISO-8859-1? If so, I agree it could make more sense to match that description.

The actual string doesn't have any charset at all; it consists of a sequence of characters, i.e. Unicode codepoints, not of any particular encoding of those characters into bytes.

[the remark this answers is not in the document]
roberthoenig: > Hmm, what do you mean -- is there something in the chain there that describes it as ISO-8859-1?

Nothing in the chain, but rather Base64 itself: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Encoder.html#encodeToString-byte:A-

> This method first encodes all input bytes into a base64 encoded byte array and then constructs a new String by using the encoded byte array and the ISO-8859-1 charset.

That quote is also what I tried to express with "returns a string that is constructed with the ISO-8859-1 charset".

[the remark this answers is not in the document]
gnprice: > But does it make reasoning really easier for you? I find it super helpful to be reassured by this encoding that it'll just work with any bytes that get thrown at it, so I don't need to worry about it at all. For ASCII, I first need to know that the bytes generated by Base64 happen to make up a valid ASCII string.

Well, I think we found with a previous experiment you tried that encoding with iso-8859-1, in the context of the APIs handy here, didn't just work in that way. :)

More generally, ASCII is the set of characters where you can always (*) plug your ears and pretend there's just one true character encoding, Unicode was never needed, and the world is simple -- and things will work out just fine. So that's why when I know something is ASCII, it makes it easy to reason about what the code will do. Nothing else works quite so well for blissful ignorance, not even iso-8859-1.

(*) almost -- very occasionally you might run into EBCDIC.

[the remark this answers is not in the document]
roberthoenig: > Well, I think we found with a previous experiment you tried that encoding with iso-8859-1, in the context of the APIs handy here, didn't just work in that way. :)

Yeah, that makes sense as well. What's your final call?

gnprice:
Oh, I just thought of an important issue: what does this code do on iOS?

It's fine to not compress the state on iOS -- there isn't the same reason for urgency, and we can do it later if we like. But it is important that the app not fail because it can't find the code to do compression. :-)

I think the current version of this PR will try to find a `TextCompressionModule` regardless of platform, and so on iOS presumably blow up because that doesn't exist.

In order to make forward progress, I'm going to go ahead and merge this code but with a tweak to not actually invoke it. An important followup will be to go ahead and invoke it :grin:, conditionally.

gnprice:
OK, merged as 19dfdf2ef . Thanks for all the work on this!

Leaving this PR open just for that conditional-fallback logic; that should be the easy part :wink: . I think the logic I'd want is
* `boot/store.js` unconditionally refers to `CompressedAsyncStorage`. Or maybe we rename that, since it's more like "async storage supporting Zulip's custom encoding(s) where applicable"... `ZulipAsyncStorage` would be reasonable.
* That, in turn, looks for the compression module; when it finds it, it goes ahead like now, and when it doesn't, `setItem` just stores the raw data, and `getItem` gives an error if it finds it needs the module. (This should be impossible, unless.... someone downgrades on iOS to this version from some future version where we have compression on iOS. That's the same situation as if someone downgrades from this version to an older one, and it's fine to discard the state in that case.)

roberthoenig:
Cool, will update! One thought ahead: My view of the Base64 encoding is that its not so bad in terms of size but in terms of performance: Assuming a staggering 1 byte / cycle performance, for our 3 million byte state, that'd still mean 3 mio cycles, which can make up to a couple milliseconds. But that's not all. Looking at this line:
```js
promise.resolve(header + Base64.encodeToString(outputStream.toByteArray(), Base64.DEFAULT));
```
there is another problem: we concatenate the (now compressed) state string, ~x00,000 bytes, onto the header. This means allocating a new chunk of ~x00,000 bytes and writing the new string there. If we wouldn't apply the Base64 encoding, we could simply have the compression algorithm write its output to a preallocated byte array for both. On the other hand, these big byte arrays are probably cached, so copying them should be much, much faster.

roberthoenig:
Updated!

Hmmm, anyone looking at this PR will now only see the two new commits listed, but not the original 19dfdf2. Maybe it'd have been better after all to close the PR and open a new one?

src/boot/ZulipAsyncStorage.js
```diff
@@ -18,15 +21,32 @@ export default class CompressedAsyncStorage {
...
+      ) {
+        result = await NativeModules.TextCompressionModule.decompress(result);
+      } else {
+        // Panic! If we are confronted with an unknown format, there is
+        // nothing we can do to save the situation. Log an error, clear the
+        // storage and restart the app. This should not be possible unless a
+        // user downgrades their version of the app.
+        logErrorRemotely(new Error(`No decompression module found for format ${header}`), '');
+        AsyncStorage.clear();
+        RNRestart.Restart();
```
roberthoenig: This else clause turned out to be much harder to implement than expected. Unfortunately, redux-persist doesn't ship some reset-and-rehydrate option (which would've been perfect here). 

In light of this, I went on and tried to return NULL_OBJECT '{}' to reset the state manually this way. But NULL_OBJECT is still an object, and therefore no complete reset. As a result, some reducers that expected arrays failed when confronted with {}. The right thing to do that also happens on the first startup is to return `undefined`, which is a case that our reducers handle appropriately. But `undefined` doesn't exist in JSON, and whatever we return gets parsed by `JSON.parse()`... Same story with custom transformers. One solution to this would be to inspect the requested key and return either {} or [], whatever the null state for that reducer is. But that's hacky and brittle. And since we only expect this situation to occur extremely infrequently, we should just go ahead, reset the storage and do a clean restart. Even better, `RNRestart.Restart();` doesn't do a full restart, but only reloads the JS.

gnprice:
> If we wouldn't apply the Base64 encoding, we could simply have the compression algorithm write its output to a preallocated byte array for both. On the other hand, these big byte arrays are probably cached, so copying them should be much, much faster.

Yeah, I'd expect the compression or decompression to take more time here than the base64 encoding/decoding. I'm also not sure what other steps are involved under the hood of that Java <-> JS bridge -- I expect that also involves some copies already.

I think this is basically a textbook example of the kind of area where you don't want to try to optimize from a static look at the code, though -- instead measure, see what's slow, then optimize that. And before that: measure the numbers that matter for the app as a whole (like latency between returning to the event loop, or total CPU utilization), and see if this even is one of the significant contributors there.

gnprice:
(And if it is -- thanks to this work, we'll have a good basis on which to experiment with variations!)

gnprice:
> Hmmm, anyone looking at this PR will now only see the two new commits listed, but not the original 19dfdf2. Maybe it'd have been better after all to close the PR and open a new one?

That'd have been fine with me too. But I've basically given up on getting GitHub's UI to do a reasonable job of showing what happened -- really the only case where this works well is when a PR is sent, and then it's merged (with an actual Git merge, cluttering the history when that's done routinely for every small feature) without ever being revised. Once you make any revision, the old code comments' context gets thrown away, with the cheeky "We went looking everywhere, but couldn’t find those commits":
image

So the best I've found is to mention the commit ID at the tip of the version pushed, so that at least there's a cross-reference to that one.

src/boot/ZulipAsyncStorage.js
gnprice: Hmm, interesting!

What if we throw an error? (Given the look of this API, I think that'd mean first something like `callback(error, undefined)`, then `throw error`.) That seems like an appropriate response to this situation.

It looks like this API is intended to sometimes throw errors, so I'd hope the callers would handle that reasonably.

gnprice: A possible bit of data here: in my emulator, I had installed this version and then of course ended up "downgrading" (to current master d42189796 plus a short branch) to review a different PR.

I got this error:
image

That's `null is not an object (evaluating 'users.reduce')`, in this code:
```
export const getUsersById = createSelector(getUsers, (users = []) =>
  users.reduce((usersById, user) => {
    usersById[user.user_id] = user;
    return usersById;
  }, {}),
);
```

And `getUsers` is just `state => state.users`. So we're ending up with a state where `state.users` is `null`. Not sure what causes that, but it's clearly a pretty unreasonable result; better to ignore the persisted state entirely, as if it had disappeared.

roberthoenig: > What if we throw an error? (Given the look of this API, I think that'd mean first something like callback(error, undefined), then throw error.) That seems like an appropriate response to this situation.

Ah yeah, I looked in the handler code for this:

```js
  let restoredState = {}
  let completionCount = 0

  storage.getAllKeys((err, allKeys) => {
    if (err) {
      if (process.env.NODE_ENV !== 'production') console.warn('redux-persist/getStoredState: Error in storage.getAllKeys')
      complete(err)
    }

    let persistKeys = allKeys.filter((key) => key.indexOf(keyPrefix) === 0).map((key) => key.slice(keyPrefix.length))
    let keysToRestore = persistKeys.filter(passWhitelistBlacklist)

    let restoreCount = keysToRestore.length
    if (restoreCount === 0) complete(null, restoredState)
    keysToRestore.forEach((key) => {
      storage.getItem(createStorageKey(key), (err, serialized) => {
        if (err && process.env.NODE_ENV !== 'production') console.warn('redux-persist/getStoredState: Error restoring data for key:', key, err)
        else restoredState[key] = rehydrate(key, serialized)
        completionCount += 1
        if (completionCount === restoreCount) complete(null, restoredState)
      })
    })
  })
```

So in case of an error, redux-persist behaves in two different ways:

* if we are in development mode and receive an error, print a warning and restore the state `{}`.

* if we are in production mode, proceed as if the warning didn't exist.

roberthoenig: Both scenarios are undesirable for us: Just using `{}` doesn't work, (as explained in the comment above). And pretending that there is no error of course doesn't work either.

roberthoenig: > better to ignore the persisted state entirely, as if it had disappeared.

So that's what I meant with
>  One solution to this would be to inspect the requested key and return either {} or [], whatever the null state for that reducer is. But that's hacky and brittle.

This is brittle because we'd just ignore the persisted state for a single key. But say for some reason, we ignore it for `messages` but manage a rehydration for `caughtUp`, then we might end up thinking we have messages that we don't actually have. So it'd indeed be best to just ignore the **entire** state, but that's difficult with redux-persist's key-based interface.

roberthoenig: Thus, I was thinking that the best solution is to just `AsyncStorage.clear()` the whole thing and then restart the app. The error you're getting should not pop up in production, right? I was hoping that in production, the app would just silently restart.

roberthoenig: That said, I don't understand why the error pops up even though we already initiated the restart...

roberthoenig: Oh, wait, I think I now know the reason why you are getting this annoying error message: When you say you "downgraded" to master, it means that you actually are downgrading to a version that is completely unaware of compression, right? In fact, it is so unaware that it doesn't even know that the stored state could be compressed, and hence it doesn't know how to handle that situation. In such a situation, I think we can't do any better than this. We cannot change the code of already installed apps (except perhaps with "updates on the fly"? but that's a different story.)

The best we can do is to handle error from now on. You can test a  "downgrade", by commenting out 

```java
File: android/app/src/main/java/com/zulipmobile/ZulipNativePackage.java
29:         modules.add(new TextCompressionModule(reactContext));
```

roberthoenig:
Updated. Didn't really the error logic (see my comments in the now-collapsed discussion) but changed the Restart to an alert box that should be more informative and less error-prone.

src/boot/ZulipAsyncStorage.js
gnprice: Huh wait interesting:

> storage.getAllKeys

So, it looks like the `redux-persist` code you're looking at is operating on the *root* state -- not on a single property / subtree of that state. That is, it's operating on `accounts`, `drafts`, `messages`, `users`, etc., etc., all at once, not on any one of them.

Which means that `{}` is exactly the right answer if we have no data.

More generally, though: somehow this system does something reasonable if there really is no data to be had; so there had to be a way to get that result, whatever it is.

gnprice: Still more interesting, though:

> storage.getAllKeys

That is a method we're not overriding -- instead we're just carrying it through from `AsyncStorage`.

I took at look at the implementation of that method in `AsyncStorage`. It doesn't use the `getItem` method! Instead it goes straight to the Java backend:
```
  getAllKeys: function(callback?: ?(error: ?Error, keys: ?Array<string>) => void): Promise {
    return new Promise((resolve, reject) => {
      RCTAsyncStorage.getAllKeys(function(error, keys) {
        callback && callback(convertError(error), keys);
        if (error) {
          reject(convertError(error));
        } else {
          resolve(keys);
        }
      });
    });
  },
```

So, it seems like this ought to bypass the part where we so carefully decompress. Perhaps we should rewrite `getAllKeys`?

(This is also making me think maybe it would all be easier if we did the wrapping in Java anyway. `ZulipAsyncStorage` has very little logic of its own, so it's not like this would mean writing a ton of Java code.)

gnprice: OK, I think I found a solution. I am not impressed with redux-persist's treatment of errors here -- that thing where when it gets an error, it barrels on ahead with attempting to process whatever garbage it thinks it got (and only in production, too!) is really something -- but there's at least one important thing it gets right here: if the result for a given key is `null`, then the built-in rehydrate logic seems to handle that properly, so the state we wind up with under that key remains `{}` or `[]` or whatever it was that we'd initialized it to.

We just need to handle it properly in the few spots where we have custom rehydrate logic of our own.

One of those is `navReducers` -- that's the one where a failure means we get stuck on the loading screen forever. I tried fixing that, and now if we just raise an error in the normal way, everything degrades gracefully.

roberthoenig: Aah, great!

PR #2823, opened 2018-07-20
[WIP] android: Build RN from `zulip/react-native` from git.

jainkuniya:
Ready to initial review.


android/settings.gradle
```diff
@@ -23,3 +23,7 @@ project(':react-native-device-info').projectDir = new File(rootProject.projectDi
 include ':app'
 include ':react-native-vector-icons'
 project(':react-native-vector-icons').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-vector-icons/android')
+
+include ':ReactAndroid'
+project(':ReactAndroid').projectDir = new File(
```
jainkuniya: Should we add about this in our docs too? Thoughts?

borisyankov:
To be honest, I have no idea why each of the lines were changed. Would you give some details?

jainkuniya:
Ya will write comment on each line next.

On Jul 20, 2018 11:31 PM, "Boris Yankov" <notifications@github.com> wrote:

To be honest, I have no idea why each of the lines were changed. Would you
give some details?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/zulip/zulip-mobile/pull/2823#issuecomment-406680660>,
or mute the thread

android/build.gradle
```diff
@@ -3,9 +3,10 @@
 buildscript {
     repositories {
         jcenter()
+        google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.+'
```
borisyankov: That's pretty cool.
How confindent are we in making this change vs where the main RN project is?
Will this introduce any unwanted (or different) behavior?

android/app/build.gradle
```diff
@@ -92,7 +92,7 @@ def enableProguardInReleaseBuilds = false
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion '27.0.1'
```
gnprice: Plugin 2.3.0 needs build tools 25.0.0 or higher, so we'll still need to upgrade this one.

I checked the release notes for that:
https://developer.android.com/studio/releases/build-tools

and it looks quite compatible -- basically bug fixes and support for new language and Android features -- so that's fine.

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sun Mar 19 12:07:15 PDT 2017
+#Sat Jul 21 13:57:10 IST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions-snapshots/gradle-4.4-20171031235950+0000-all.zip
```
gnprice: Hmm. The build actually fails for me, with this error:
```
$ react-native run-android
Scanning folders for symlinks in /home/greg/z/mobile/node_modules (5ms)
JS server already running.
Building and installing the app on the device (cd android && ./gradlew installDebug)...
Downloading https://services.gradle.org/distributions-snapshots/gradle-4.4-20171031235950+0000-all.zip

Exception in thread "main" java.io.FileNotFoundException: https://downloads.gradle.org/distributions-snapshots/gradle-4.4-20171031235950+0000-all.zip
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1890)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
	at org.gradle.wrapper.Download.downloadInternal(Download.java:58)
	at org.gradle.wrapper.Download.download(Download.java:44)
	at org.gradle.wrapper.Install$1.call(Install.java:59)
	at org.gradle.wrapper.Install$1.call(Install.java:46)
	at org.gradle.wrapper.ExclusiveFileAccessManager.access(ExclusiveFileAccessManager.java:65)
	at org.gradle.wrapper.Install.createDist(Install.java:46)
	at org.gradle.wrapper.WrapperExecutor.execute(WrapperExecutor.java:126)
	at org.gradle.wrapper.GradleWrapperMain.main(GradleWrapperMain.java:61)
```

And in fact that URL gives me (a redirect to) a 404, either with `curl -I` or in a browser.

Can't be hard to fix, but this will be moot if we stick with Gradle 3.3.

android/build.gradle
```diff
@@ -3,10 +3,11 @@
...
-
+        classpath 'com.android.tools.build:gradle:3.1.0-alpha05'
```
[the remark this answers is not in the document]
jainkuniya: Okay!
Now using `2.3.+`

gnprice:
(I described this in PMs with @jainkuniya a couple of days ago, but recording here.)

This doesn't build for me. The reason is that our dependencies each have their own idea of SDK Build Tools version to use (i.e., `buildToolsVersion`), and many of them are 23.0.1 -- too old for version 2.3 of the Android Gradle Plugin, as mentioned above.

Here's what the error looks like when I try to build:
```
A problem occurred configuring project ':@remobile/react-native-toast'.
> The SDK Build Tools revision (23.0.1) is too low for project ':@remobile/react-native-toast'. Minimum required is 25.0.0
```

That just mentions the first dependency it happens to look at that has this issue, though; the same thing applies to lots of others:
```
$ find node_modules/ -name build.gradle | xargs grep -l buildToolsVersion.*23
node_modules/@remobile/react-native-toast/android/build.gradle
node_modules/react-native-photo-view/example/android/app/build.gradle
node_modules/react-native-sentry/android/build.gradle
node_modules/react-native-fetch-blob/android/build.gradle
node_modules/react-native-orientation/android/build.gradle
node_modules/react-native-orientation/demo/android/app/build.gradle
node_modules/react-native/ReactAndroid/build.gradle
node_modules/react-native/local-cli/templates/HelloWorld/android/app/build.gradle
node_modules/react-native-sound/android/build.gradle
```

We're not going to go and edit the files inside each of those dependencies to fix that. But Gradle configurations are just programs, and I think there should be some reasonable code we can add to our own `build.gradle` files that will fix this, by causing `buildToolsVersion` to effectively end up being whatever version we choose despite what those `build.gradle` files say.

As I said in my prototype commit (7ce4d0c2f): some debugging, documentation-reading, and experimenting will be required. 

Here's the Gradle user manual:
https://docs.gradle.org/current/userguide/userguide.html

Some headings I'd look under first would include "dependency management" and "multi-project builds".

android/build.gradle
```diff
@@ -22,3 +22,13 @@ allprojects {
         }
     }
 }
+
+subprojects {
+    afterEvaluate {project ->
+        if (project.hasProperty("android")) {
+            android {
+                buildToolsVersion '27.0.1'
```
gnprice: How did you choose the `subprojects` / `afterEvaluate` form for doing this?

From what I've read about Grade config I know there are several choices for how to make a patch like this, and they have subtly different meaning e.g. in what gets evaluated first, and what gets to override what. So I'd like to make sure we/I understand what choice we're making here.

jainkuniya: This version is adopted from https://docs.gradle.org/current/userguide/build_lifecycle.html#sec:project_evaluation

> subprojects {

Means only for subproject, we are adding this in the root project (which also contains our app as one of the project).

> afterEvaluate {project ->

This is a kinda callback (described as `closures` in the docs). Called as soon as project is evaluated, but before executing.

> if (project.hasProperty("android")) {

Making sure that we only override* variable value for android project/dependency, as `buildToolsVersion` is only required for android project.

* By doing this we are injecting/changing value after project is evaluating and waiting for executing (building project).


Another way is implementing interface listener (something like https://docs.gradle.org/current/javadoc/org/gradle/BuildListener.html).

Currently our gradle files follow `closures` approach only, so I thinks we should use `closures` only. 

like 
```
allprojects {
    repositories {
        mavenLocal()
        jcenter()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url "$rootDir/../node_modules/react-native/android"
        }
    }
}
```

I completely agrees with the all of this 😄 , and in case if something went wrong we can definitely improve in future 😅 . For now I thinks this is sufficient 😄 .

(Now updating commit msg and rearranging)

PR #2824, opened 2018-07-21
Translate strings used in non components (Lightbox and message action sheet)

kunall17:
Using the react-intl inject API a '_' called function (just like the web server) is being passed on to the context and used in various locations.

### DIFF
--- a/src/i18n/translations/messages_en.json
+++ b/src/i18n/translations/messages_en.json
@@ -141,6 +141,7 @@
   "Shared using Zulip!": "Shared using Zulip!",
   "Add reaction": "Add reaction",
   "Share message": "Share message",
+  "Error": "Error",
   "Failed to edit message": "Failed to edit message",
   "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
 }
--- a/src/i18n/translations/messages_en.json
+++ b/src/i18n/translations/messages_en.json
@@ -141,5 +141,6 @@
   "Shared using Zulip!": "Shared using Zulip!",
   "Add reaction": "Add reaction",
   "Share message": "Share message",
+  "Failed to edit message": "Failed to edit message",
   "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
 }
--- a/src/i18n/translations/messages_en.json
+++ b/src/i18n/translations/messages_en.json
@@ -143,5 +143,7 @@
   "Share message": "Share message",
   "Error": "Error",
   "Failed to edit message": "Failed to edit message",
+  "Storage permission needed": "Storage permission needed",
+  "To download images, allow Zulip to store files on your device.": "To download images, allow Zulip to store files on your device.",
   "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
 }
--- a/src/i18n/translations/messages_en.json
+++ b/src/i18n/translations/messages_en.json
@@ -137,5 +137,7 @@
   "Topics": "Topics",
   "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n": "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n",
   "Add subscribers": "Add subscribers",
+  "Download image": "Download image",
+  "Shared using Zulip!": "Shared using Zulip!",
   "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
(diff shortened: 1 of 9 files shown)


src/lightbox/LightboxActionSheet.js
```diff
@@ -1,5 +1,5 @@
 /* @flow */
-import type { Auth } from '../types';
+import type { Auth, TranslateStringType } from '../types';
```
borisyankov: We are not naming types `*Type`. Remove that part.
Also, try to think of a more descriptive name. Maybe `StringTranslator` or even `TextTranslator` (as 'text' is used in 'getText')

src/boot/TranslationAPIProvider.js
```diff
@@ -0,0 +1,28 @@
+/* @flow */
+import { PureComponent } from 'react';
+import { injectIntl, intlShape } from 'react-intl';
+
+import type { ChildrenArray } from '../types';
+
+type Props = {
+  children?: ChildrenArray<*>,
+  intl: intlShape.isRequired,
```
borisyankov: This does not look like a correct type.

src/boot/TranslationAPIProvider.js
```diff
@@ -0,0 +1,28 @@
...
+};
+
+class TranslationAPIProvider extends PureComponent<Props> {
+  static childContextTypes = {
+    _: string => {},
```
borisyankov: `string => {},` is not correct here - it defines a function with an argument **named** `string`. This should not reflect the function signature, it is a placeholder that we use in place of PropTypes. So just `() => {}` is enough.

Do not forget to update the `types.js` => `type Context` with the `_` value.

src/lightbox/LightboxActionSheet.js
kunall17: Done

src/boot/TranslationAPIProvider.js
kunall17: Yeah, that was a typo, and in this I'm returning a string so I changed to `() => '' `

kunall17:
Thanks for review @borisyankov, made the changes!

src/boot/TranslationAPIProvider.js
```diff
@@ -0,0 +1,28 @@
...
+import type { ChildrenArray } from '../types';
+
+type Props = {
+  children: ChildrenArray<*>,
+  intl: intlShape.isRequired,
+};
+
+class TranslationAPIProvider extends PureComponent<Props> {
+  static childContextTypes = {
+    _: () => '',
```
gnprice: I've never really quite understood what `contextTypes` is supposed to do. @borisyankov , can you explain why it's there and why the types all seem to be `() => {}`?

I did look at https://reactjs.org/docs/legacy-context.html which seems to be the relevant documentation, but the types there are all from `PropTypes`.

src/i18n/translations/messages_en.json
```diff
@@ -137,5 +137,7 @@
   "Topics": "Topics",
   "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n": "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n",
   "Add subscribers": "Add subscribers",
-  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
+  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream.",
```
gnprice: Small thing, but: let's always add to this list somewhere in the middle -- really anywhere other than the very last line. That avoids unnecessary churn in the diff due to the lack of a trailing comma.

src/boot/TranslationAPIProvider.js
gnprice: Let's keep this `() => {}` like everywhere else -- these lines are still puzzling to me :smile: , but there's no point in making it different in a way that we don't have a good story for why it'd help anything.

kunall17:
@gnprice Updated!

gnprice:
Hi @kunall17 -- thanks for the update!

I realized that the API I had in mind for handling translations would require some close reading of React docs and some experimentation... so I went and spent today doing that. I just pushed a series  c22dfee9b^..9eaa05c27 adding code to provide `_` and a new API for getting access to it, followed by 
a long second series c94586ca9^..d688069be making a number of refactors to the `messageActionSheet` code.

Would you try rebasing this branch on top of master, and using the new API for the lightbox code? I think that would be a good next step.

kunall17:
@gnprice Updated the branch :)

chrisbobbe:
This has gathered quite a few conflicts since 2018, so I'm closing it. Thanks for your efforts, @kunall17!

PR #2826, opened 2018-07-22
flow: Annotate MessageListEventLongPress.target.

jackrzhang:
This type corresponds to: https://github.com/zulip/zulip-mobile/blob/2024f14880ee60c0f06b7a564aeb3d1c5eb75dee/src/webview/js/js.js#L349.


gnprice:
Merged, thanks!

PR #2827, opened 2018-07-22
flow: Annotate ConstructActionButtonsType more precisely.

jackrzhang:
Use the `Message` and `FlagsState` aliases.


gnprice:
Merged, thanks!

PR #2828, opened 2018-07-22
Add a DrawerNavigator to house HomeTab buttons and settings

agzuniverse:
Having tabs at the bottom and buttons at the top for the `HomeTab` looked weird, and apps usually don't have settings as a major tab along with the other tabs.

This adds a `DrawerNavigator` that houses the buttons that was on top of `HomeTab` (except search, I left that in the app bar of the `HomeTab` for now.) and settings. Settings is removed from the bottom `TabNavigator`.

The drawer can be used for displaying more info if needed (Maybe the username and avatar of the current user, as in the twitter app?)

homewithdrawwer

draweropen


gnprice:
Thanks!

Quick design feedback, before I dig into the code:
* I like leaving search in the app bar.
* Let's put some info at the top of the drawer about the current user / account. In addition to Twitter, take a look at Gmail or Hangouts for examples of this.
  * That would also be a good place to put the "switch account" UI. That change can be a followup PR, though.
* What about the other two tabs? I think they should probably get an app bar with a navigation icon too.
  * For PMs, there's a search button now, so the app bar makes a nice place to move that. The "Create group" can turn into a FAB, the classic kind with just the + icon.
  * For streams... I guess the top tabs can go below the app bar. There's probably something better we can do, though.

borisyankov:
I think Twitter, Gmail, and Hangouts have drastically different UI than ours, thus this header thing makes way less sense in our context:

Hangouts and Gmail have a single view that lists the main items, Twitter is slightly more similar to Zulip in its layout, but auto-hides the header when scrolled down.

Putting a full search box there ala Facebook or Messenger looks more useful.

agzuniverse:
I've added user account information in the drawer and updated the screenshot.

I'll add proper app bars for the other two tabs soon.

agzuniverse:
The app bar moving along with the tab navigator is not commonly seen in apps. A single fixed app bar outside the tab navigator seems to be the norm.

So I created a wrapper around the tab navigator and moved the app bar into that.
I'm not entirely satisfied with how the results look, but here are the screenshots. Need some thoughts and ideas on these.

screenshot_1532706274
screenshot_1532706278

PR #2830, opened 2018-07-23
Simplify and clarify user selectors.

roberthoenig:
These commits can be merged one by one, each of them is a small improvement. However, they should be merged in order.

The first four commits are lightweight, the last two commits heavy.


src/account-info/AccountDetails.js
```diff
@@ -40,7 +40,7 @@ export default class AccountDetails extends PureComponent<Props, void> {
     return (
       <View>
         <Avatar
-          avatarUrl={getMediumAvatar(user.avatar_url)}
+          avatarUrl={user.avatar_url ? getMediumAvatar(user.avatar_url) : null}
```
borisyankov: Hmmm, that doesn't look like the correct behavior.
If `avatar_url` is `null` then we should return a differently sized gravatar.

We might want to handle that inside `getMediumAvatar`

src/common/Avatar.js
```diff
@@ -22,7 +22,7 @@ const componentStyles = StyleSheet.create({
 });
 
 type Props = {
-  avatarUrl: string,
+  avatarUrl: string | null,
```
borisyankov: Use a [maybe type](https://flow.org/en/docs/types/maybe/) here like `?string`.

roberthoenig: I considered that but a maybe type accepts the values `string`, `null` and `undefined`. In our codebase, only `string` and `null` seem to be used for avatarUrl, and I wanted to be as strict about that as possible.

src/account-info/AccountDetails.js
roberthoenig: So in `Avatar` we have this line:

```js
const fullAvatarUrl = avatarUrl ? getFullUrl(avatarUrl, realm) : getGravatarFromEmail(email);
```

So the avatar item explicitly accepts avatarUrls set to `null` and handles them by creating a gravatar. Isn't that the behavior that we want?

src/nav/navReducers.js
```diff
@@ -21,11 +21,22 @@ import {
...
-const rehydrate = (state: NavigationState, action: RehydrateAction): NavigationState =>
-  getInitialNavState(action.payload) || state;
+const rehydrate = (state: NavigationState, action: RehydrateAction): NavigationState => {
+  if (!getAuth(action.payload).apiKey) {
+    const accounts = getAccounts(action.payload);
+    // getStateForRoute can return null, but it is unclear under what
+    // conditions. Empirically, it doesn't return null on the initial start of
+    // the app, but this should be verified.
+    // $FlowFixMe: getStateForRoute may return null but it shouldn't.
+    return getStateForRoute(accounts && accounts.length > 1 ? 'account' : 'welcome');
```
roberthoenig: @gnprice I saw that in 9266bff170ee9cad0622c87885787b0717172024 you changed the behavior of getStateForRoute to return `null` when getActionForPathAndParams returns null. Do we know under what circumstances getActionForPathAndParams returns null?

src/common/Avatar.js
borisyankov: We definitely have the concept of `undefined` when talking about avatars. I am not so sure about `null` actually. 
Since `null` and `undefined` are often used interchangeably and officially there is no good separation in their meaning (though usually `undefined` means 'missing' or not yet defined, while `null` is a bit more explicit, but more rarely used) using the `?` as we do everywhere else would make sense.

roberthoenig: Okay, will use `?` then.

But are you sure avatars can actually be undefined?

Here is an example:

image

I wonder in what place of the code this could change to undefined.

borisyankov: I wonder the same. I am not sure if this used to not be included at all, but my preference would be to not include any key that has no data (even better, any key that has the default values)

roberthoenig:
Updated: Changed "string | null" to "?string" and fixed the merge conflict.

src/account-info/AccountDetails.js
gnprice: Seems reasonable to me.

As a separate refactor, it feels like the question of whether the avatar should be "medium" or what should be determined in the same place as the size of the gravatar -- which means that `getMediumAvatar`, which is for the case of a non-null `avatar_url`, and `getGravatarFromEmail` should be called from the same spot in the code. Probably inside `Avatar`, especially as we're already giving it a `size` prop.

That's orthogonal from this PR, though, I think. (Basically, the reason anything works now before this change is just that `getMediumAvatar`, given null input, happens to return the same thing for its output; this at least makes the behavior in the null case clearer.)

src/common/Avatar.js
gnprice: It's unfortunate that the one terse syntax Flow gives us here includes both null and undefined, when it's rare that we really want to allow both.

I'm happy to experiment with writing out things like `string | null` when in fact that's what we mean. I wouldn't want to go and do a massive sweep over the codebase for it right now, but we'll see how it looks after playing with it in a few places.

Because I don't see anywhere that we set this to null, I'm assuming that that's the way the server represents it for a user that hasn't set an avatar. That's pretty natural coming from Python -- the row in the database has `NULL` there, so it becomes `None` in Python, and JSON encoders translate that to `null`.

gnprice: Aha, in fact -- this is right there in [the API doc](https://zulipchat.com/api/get-all-users#response) that I linked to in the jsdoc for that type :grin:

> avatar_url: URL to the user's gravatar. `None` if the client_gravatar query paramater was set to  True.

That isn't quite right in a couple of ways -- it's `null`, not None, because this is JSON and not Python, and it's still non-null when the user has set a non-Gravatar avatar. But there it is -- it can be null. (And we do pass `client_gravatar: true`.)

src/realm/realmReducers.js
```diff
@@ -42,7 +42,11 @@ const initialState = {
 const realmInit = (state: RealmState, action: RealmInitAction): RealmState => ({
   ...state,
   canCreateStreams: action.data.can_create_streams,
-  crossRealmBots: action.data.cross_realm_bots,
+  crossRealmBots: action.data.cross_realm_bots.map(bot => ({
+    ...bot,
+    timezone: '',
+    avatar_url: null,
```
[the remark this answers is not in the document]
borisyankov: We should be handling correctly a 'missing value' case whether it is a missing key altogether or empty value. Then we will not need to add these at all. I think we are handling `avatar_url` correctly. We need to check for `timezone` too, but possibly we also handle it correctly.

roberthoenig:
I'm not so happy yet about the user state: Currently, we store it as an array. Ideally, it'd be a mapping from IDs to user objects. Not sure how high a priority this chane is, though.

PR #2702, opened 2018-06-20
Upgrade to React Navigation v2

borisyankov:
The PR migrates to React Navigation v2 without changing or introducing new behavior.

PR #2831, opened 2018-07-24
Use MaterialTopTabNavigator for main navigation

borisyankov:
This is based on the React Navigation 2 PR.

The last commit uses the new MaterialTopTabNavigator which is recommended to best follow Material design guidelines (it is also more configurable and nicer)


gnprice:
Thanks! As you see in my comment earlier today over here: https://github.com/zulip/zulip-mobile/pull/2702#discussion_r204930917 , I am interested in exactly this switch :)

For this PR, I'm looking at the tip commit since that's the one that's new relative to #2702 . That commit does a bunch of other things too, in addition to switching to MaterialBottomTabNavigator (btw the commit message and PR title seem to have the name flipped); those changes generally look like good improvements (like clarifying the naming around `StreamTabs`), but seem independent of the change of navigator component, and like they'd be clearer as separate commits.

When the other refactors and improvements are separated out, does the change that actually switches to MaterialBottomTabNavigator end up being very big? Perhaps it'd be simplest to skip moving MainTabs over to MaterialTopTabNavigator (one of the commits in #2702 ), and move it straight to MaterialBottomTabNavigator.

borisyankov:
The `Top` navigator switch is trivial complexity-wise so I think we can go ahead and merge them sooner and logically it makes sense as the purpose is to preserve the previous behavior.

The last commit that implements the `Bottom` one is marked as WIP yet, as there is a small behavioral change in the way tabs scroll - you can't scroll from one to the other as before. On the other hand, this new behavior is consistent with most apps work (YouTube, Twitter) so I am not so concerned.

Another thing I will need to figure-out - on transition to another tab, there is some color used to 'blend' to a new tab. When our background is white, it looks like a proper blending transition. When we are on the night theme, the same color results in a 'flashing' transition and not blending.

borisyankov:
Some of the changes in the last commit (and the purpose of this PR) are both an improvement in the structure of the app and also needed for this Bottom-Material component to work (it sets a background and doesn't work with our philosophy of 'transparent background simplifies stuff')

I'll split this into several commits too. From coding perspective, I think they will be simple but will be good that you test the behavior by running the branch to verify if it feels good in use.

gnprice:
The further work discussed above never happened; the PR still says the opposite of what it does, among everything else.

With the way we actually did the react-navigation v2 upgrade (in #3573), we never switched to the inappropriate `createMaterialTopTabNavigator` for our main, bottom, tabs in the first place as #2702 did. We're now using `createBottomTabNavigator` from `react-navigation-tabs`, and it seems fine. So I don't think there's any further useful work to do in this direction; closing.

PR #2832, opened 2018-07-24
Aesthetic improvement to welcome page

agzuniverse:
Side by side of before/after:
welcome_sidebyside

Thoughts?

### DIFF
--- a/src/start/WelcomeScreen.js
+++ b/src/start/WelcomeScreen.js
@@ -2,10 +2,23 @@
 import { connect } from 'react-redux';
 
 import React, { PureComponent } from 'react';
+import { StyleSheet, View } from 'react-native';
 
-import type { Dispatch } from '../types';
-import { Screen, ViewPlaceholder, ZulipButton } from '../common';
+import type { Context, Dispatch } from '../types';
+import { ViewPlaceholder, ZulipButton, Label, Logo } from '../common';
 import { navigateToAddNewAccount, navigateToWelcomeHelp } from '../actions';
+
+const componentStyles = StyleSheet.create({
+  title: {
+    fontSize: 30,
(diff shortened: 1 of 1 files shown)


borisyankov:
I have no preference for either of them. 
I think the start screen can be improved a lot but that would likely involve a graphics designer doing some cool graphics.

agzuniverse:
Seems like that is the better choice. I dropped the commit that added the `hideTitle` prop to screen and used a `View` with `styles.screen` instead.

agzuniverse:
@borisyankov further review?

src/start/WelcomeScreen.js
```diff
@@ -2,36 +2,63 @@
...
 
 import React, { PureComponent } from 'react';
+import { StyleSheet, View } from 'react-native';
 
-import type { Dispatch } from '../types';
-import { Screen, ViewPlaceholder, ZulipButton } from '../common';
+import type { Context, Dispatch } from '../types';
+import { ViewPlaceholder, ZulipButton, Label, Logo } from '../common';
 import { navigateToAddNewAccount, navigateToWelcomeHelp } from '../actions';
+import StylesProvider from '../boot/StylesProvider';
```
borisyankov: You don't need this import.

agzuniverse:
Updated, removing the unused import.

gnprice:
Thanks @agzuniverse for this PR!

@borisyankov and I just looked at it together, and as he said above we're not sure this is better than the existing screen. It would also definitely need some testing on small screens to make sure it still worked reasonably well.

We might at some point revisit this screen, but as Boris said above:

> that would likely involve a graphics designer doing some cool graphics.

agzuniverse:
Okay, that's cool!

PR #2833, opened 2018-07-24
jsdoc: Document user-related types.

roberthoenig:


src/types.js
```diff
@@ -458,6 +458,10 @@ export type UnreadState = {
   mentions: UnreadMentionsState,
 };
 
+/**
+ * Types the complete redux state. Each property corresponds to a top-level
```
borisyankov: Should we even mention that it is 'typing' something? 
Also it looks like we should be able to say a bit more here...

roberthoenig: Yeah, I'm not thrilled about the `Typing` preamble either. I'd be glad about better ways to start these docstrings.

roberthoenig: What more would you say here? I avoided adding @prop for all the props in GlobalState, because they should be documented where the type is defined.

gnprice:
Thanks @roberthoenig ! These are helpful.

I have some style changes I'd like to make (including removing those "Types ..." preambles :smile: ); I'll go ahead and merge and then make those on top.

gnprice:
See 1baf7aec1 .

PR #2835, opened 2018-07-25
session reducers: Simplify and explain initial fetch logic.

roberthoenig:


gnprice:
Thanks for these clarifications -- merged!

PR #2836, opened 2018-07-25
refactor: Simplify `needsInitialFetch` logic

borisyankov:
On `loginSuccess` we set `needsInitialFetch` to `true`.
This was introduced in:
https://github.com/zulip/zulip-mobile/commit/8edbf8562481d759a0a16daf0fac028ca54b6114

The code was `needsInitialFetch: !!action.apiKey` which might
suggest that more is going on, but not really. `apiKey` is always
a non empty value if login succeeds.

This PR replaces the code to `true` to make it simpler and more
clear, but no behavior is changed.

Also, adds a test for `sessionReducers\LOGIN_SUCCESS`.

### DIFF
--- a/src/session/__tests__/sessionReducers-test.js
+++ b/src/session/__tests__/sessionReducers-test.js
@@ -1,9 +1,14 @@
 import deepFreeze from 'deep-freeze';
 
-import { ACCOUNT_SWITCH, CANCEL_EDIT_MESSAGE, START_EDIT_MESSAGE } from '../../actionConstants';
+import {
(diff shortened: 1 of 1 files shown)


roberthoenig:
The first commit is a duplicate of the first commit in #2835 (probably opened a couple minutes earlier :wink: ) They have the same diff, so merging either should be fine.

gnprice:
As it happens, I just merged #2836 (and was reminded of this one through the link there) -- but the test looks useful too, thanks! Merged.

Issue #2837, opened 2018-07-26
Avoid blocking UI in onPress handlers


gnprice:
With #2799, we fixed a longstanding annoying issue in the app's UI: pressing a stream or topic in the unreads screen, the basic routine navigation move in the app, wouldn't give visual feedback on what you'd pressed. The cause was that in the onPress handler we were doing the navigation synchronously, so the animation didn't get to run until it was complete; the fix is to insert a `setTimeout` to let the UI respond before we do the navigation.

I just pushed similar fixes for a handful of other places in the app: see c003c77a9 and the four commits before it. (Note the `delay` helper I added. `setTimeout` works fine, except in a few places its return type doesn't match.)

There are still a bunch of places in the app where we have this issue -- if you go around into random parts of the UI deeper in the navigation, or just search for `onPress` and then find the corresponding widgets in the app and play with them, you find plenty. (In some of these places, if you're lucky you might see a few belated frames of the feedback animation; in others, I've never seen any.) We should fix them all.

We may want this in basically all our `onPress` handlers that do anything like navigate to a screen. If we can convince ourselves it's correct to put it *everywhere*, then we can put it in ZulipButton.js (like @jainkuniya suggested on #2799) -- that'd help a lot to handle this cleanly.

jainkuniya:
> We may want this in basically all our onPress handlers that do anything like navigate to a screen. If we can convince ourselves it's correct to put it everywhere, then we can put it in ZulipButton.js (like @jainkuniya suggested on #2799) -- that'd help a lot to handle this cleanly.

Let's try out this version 😅 

@zulipbot claim

sumukhah:
@zulipbot claim

sumukhah:
@zulipbot abandon

PR #2838, opened 2018-07-26
 touchable: Do not block UI while executing press.

jainkuniya:
Fix for #2837.

### DIFF
--- a/src/common/Touchable.js
+++ b/src/common/Touchable.js
@@ -39,6 +39,14 @@ export default class Touchable extends PureComponent<Props> {
     delay(onPress);
   };
 
+  handleLongPress = () => {
+    const { onLongPress } = this.props;
+    if (!onLongPress) {
+      return;
+    }
+    delay(onLongPress);
+  };
+
   render() {
     const { accessibilityLabel, style, children, onPress, onLongPress } = this.props;
 
@@ -61,7 +69,7 @@ export default class Touchable extends PureComponent<Props> {
           underlayColor={HIGHLIGHT_COLOR}
           style={style}
           onPress={this.handlePress}
-          onLongPress={onLongPress}
+          onLongPress={this.handleLongPress}
         >
           <View>{children}</View>
         </TouchableHighlight>
@@ -74,7 +82,7 @@ export default class Touchable extends PureComponent<Props> {
         style={style}
         background={androidBackground}
         onPress={this.handlePress}
-        onLongPress={onLongPress}
+        onLongPress={this.handleLongPress}
       >
         <View>{children}</View>
       </TouchableNativeFeedback>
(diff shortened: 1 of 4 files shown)


src/utils/async.js
```diff
@@ -2,7 +2,7 @@
 import { nullFunction } from '../nullObjects';
 
 /** Like setTimeout(..., 0), but returns a Promise of the result. */
-export function delay<T>(callback: () => T): Promise<T> {
+export function delay<T>(callback: () => void | Promise<any>): Promise<T> {
```
gnprice: I am not excited about adding an `any` here. In general, I want to avoid adding `any`.

What's the specific issue you ran into that made you want to make this type looser? Let's debug and find a better solution. I'm sure there are ways I'd be happy to adjust this type -- but using `any` is not going to be one of them. :stuck_out_tongue:

src/common/Touchable.js
```diff
@@ -30,8 +33,35 @@ type Props = {
...
+    handlePressWithDelay: true,
+  };
+
+  handlePress = () => {
+    const { onPress } = this.props;
+    if (!onPress) {
+      return;
+    }
+    delay(onPress);
+  };
```
gnprice: Hmm, this function is kind of wordy.

How about something like
```
onPress={onPress && () => delay(onPress)}
```
?

src/common/Touchable.js
```diff
@@ -22,6 +24,7 @@ type Props = {
  * Component to encapsulate our custom and platform-specific
  * settings applied to the built-in touchable components.
  *
+ * @prop [handlePressWithDelay] - Handle press with delay or not.
```
gnprice: This line of documentation really doesn't add anything to the name and type. :-) When that's what we have to say, it's better to just leave it off.

Also, this line should be at the end; the others, like `children` and `onPress`, are much more central to what the component does and should be shown to the reader first.

src/common/Touchable.js
```diff
@@ -30,8 +33,35 @@ type Props = {
 export default class Touchable extends PureComponent<Props> {
   props: Props;
 
+  defaultProps: {
+    handlePressWithDelay: true,
```
gnprice: What's the motivation for adding this option?

It looks like you don't actually have any callers turn it off.

gnprice:
In fact, I browsed around a bit, and here's a specific case where I think the experience will be made worse by delaying like this.

`Touchable` is used in `ComponentWithOverlay` -- itself a super generic utility component.

That in turn is used in `AvatarItem`.

And the `onPress` supplied by `AvatarItem` looks like this:
```
  handlePress = () => {
    const { email, onPress } = this.props;
    Animated.timing(this.animatedValue, {
      toValue: 0,
      duration: 300,
      useNativeDriver: true,
      easing: Easing.elastic(),
    }).start(() => onPress(email));
  };
```

So when we do call that, it should return very quickly; there's no cost to calling it immediately. And on the other hand, calling that is how we set up a UI animation that's going to run for 300ms, and which will be handled by the native animation driver, outside JS. So if we *delay* calling it... that animation is delayed in starting. Which is going to look janky, and which we didn't need to do. :-/

Issue #2840, opened 2018-07-26
Add subscribers button in stream screen is only visible to Admin.


jainkuniya:
Just checked on web app. Anyone can add anybody in the stream visible to him.

jainkuniya:
@zulipbot claim

PR #2841, opened 2018-07-26
streamScreen: Show 'Add subscribers' to all users.

jainkuniya:
Before it was only shown to admin. Checked on web app, anyone can add
anybody in the stream visible to him.

This makes behaviour consistent on web app and mobile app.

Fix: #2840


gnprice:
Looks good, merged -- thanks @jainkuniya for the fix!

PR #2842, opened 2018-07-26
Clean up message state and document message logic.

roberthoenig:
This is preparatory for making the message state relational. Mainly, these commits shall make reasoning about the message logic and state easier.


gnprice:
And just merged the two commits that were easiest to merge already; plus another commit on top expanding a bit on those helpful comments on `CaughtUp` and `CaughtUpState`. Thanks again!

roberthoenig:
> The rename should come with a migration, right? Is this branch intended to follow after the redux-persist version upgrade?

Ah, yup, that's correct.

roberthoenig:
Updated! Added explanatory comments for the anchor constants. Added a migration for `messages` -> `narrows`.

src/constants.js
```diff
@@ -7,3 +7,12 @@
...
 export const NEWEST_MESSAGE_ANCHOR = Number.MAX_SAFE_INTEGER;
+
+/**
+ * Pointer to the oldest message in a narrow.
+ *
+ * Usually, an anchor is a message ID. This constant can be used when we
+ * want to retrieve the oldest messages for a narrow, without actually
+ * knowing their IDs.
+ */
+export const OLDEST_UNREAD_ANCHOR = 0;
```
borisyankov: We do already use 'first unread' both in this project and in `zulip/zulip`.
I don't have any preference for this or your pick in a world where we haven't already picked one of the two. This is not that timeline :)
Let's just use `FIRST_UNREAD_ANCHOR` so we don't introduce inconsistent terminology.

src/constants.js
```diff
@@ -0,0 +1,3 @@
+/* @flow */
+export const NEWEST_MESSAGE_ANCHOR = Number.MAX_SAFE_INTEGER;
```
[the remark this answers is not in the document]
borisyankov: This constant is also worth naming consistently with the main project.
It seems we are using `last message` in `zulip/zulip`.

roberthoenig:
Updated with a better commit hygiene and renamed NEWEST_* OLDEST_* to FIRST_* LAST_*

roberthoenig:
We should include a copyright notice for redux-persist-migrate. I didn't do that because I didn't know how.

roberthoenig:
Updated: Added the big commit for normalizing the message state. Some caveats

* So far, I disregarded any state optimizations. Tests that tested for these optimizations (e.g. by using "toBe" instead of "toEqual") have been commented out. 

* messageReducers still needs tests.

Also, I'm experiencing some issues with redux-persist-migrate. Marking this PR as WIP, but reviews appreciated.

roberthoenig:
Updated: Fixed the migration system and added tests for the new messageReducers.js.

> So far, I disregarded any state optimizations. Tests that tested for these optimizations (e.g. by using "toBe" instead of "toEqual") have been commented out.

I removed the commented out tests. I think the PR is ready for review. @borisyankov can you test on an old device how the app performs without the message state optimizations?

gnprice:
Sorry this has taken a while! I just reviewed and merged the first few commits, adding the redux-persist-migrate code to our repo. I created a file for us to record third-party open-source licenses, following the pattern in zulip/zulip .

When I checked, the hash of the version of that file in your first commit and of the file you point to upstream didn't actually match -- turned out there was an extra blank line at the beginning. :slightly_smiling_face: So I touched that up.

The next commit ("Switch on ...") looks good too, except I don't like putting the version under `info`; that sounds like it could be anything, and in particular I think the natural assumption for someone seeing that would be that it's some kind of application-level information like just about everything else in the state. How about `migrations`?

gnprice:
From the next commit:

* Sounds like `version?: string` would be the right type in the previous commit.

* Can we have the migration actually preserve the data but move it? That seems like it'd often be desirable :wink: , so this is a good time to demonstrate as an example. Perhaps like this?

```
const migrations = {
  '0': state => {
    AsyncStorage.removeItem('reduxPersist:messages');
    const { messages, ...rest } = state;
    return { narrows: messages, ...rest };
  },
```

gnprice:
And merged the two easy commits after that. The big "normalize" commit deserves close attention, and what I should do right now is go sleep; I'll have to return to it later.

The other two commits should be quick to merge given the changes above, though.

Oh, one quick thing I see in that big last commit: would you split out the package.json/yarn.lock changes as a small preceding commit?

Thanks again for all your work on this! This is a complex change, and I think it's going to clean some things up for us significantly.

roberthoenig:
> Can we have the migration actually preserve the data but move it? That seems like it'd often be desirable wink , so this is a good time to demonstrate as an example. Perhaps like this?
```
const migrations = {
  '0': state => {
    AsyncStorage.removeItem('reduxPersist:messages');
    const { messages, ...rest } = state;
    return { narrows: messages, ...rest };
  },
```

I don't think we can, because the commit that introduces this migration also changes the 1-level reducers, so redux-persist won't actually know anymore that a `messages` state existed. The state it'll pass to the migration is `{narrows: {}, ...rest}` (without any `messages`!).

In this sense, the migration system is sort of incomplete, because the individual migrations don't include information about changes to the reducer tree, in particular the 1-level reducers. And the current model doesn't allow patching redux-persist-migrate with this functionality, because it is always called *after* redxu-persist retrieves the state from the storage. 

If we really wanted to move the old `messages` to `narrows`, we could probably do some hackery to retrieve it from the AsyncStorage and decompress it. But I don't think that's worth the effort for this migration.

roberthoenig:
@gnprice Updated with your suggestions.

gnprice:
> I don't think we can, because the commit that introduces this migration also changes the 1-level reducers, so redux-persist won't actually know anymore that a messages state existed. The state it'll pass to the migration is {narrows: {}, ...rest} (without any messages!).

Hmm, I see.

That seems like something we may want to fix! So I guess it's helpful that now we now about it. But definitely not critical for this particular migration.

I've just gone and pushed this migration commit and its parent which switches the system on.

gnprice:
Reading the main commit (at the end) now. Thanks for taking on this complex change!

* When we drop the existing `narrows` state, I'm concerned that we don't forget enough of the surrounding state, leaving the state inconsistent -- that we might get into a state where e.g. we believe our list of message IDs in that narrow (which is empty) is an up-to-date list, when in fact it isn't.
   * This isn't an issue if we do actually have the event queue expire, of course, for the reason you explain in the commit message -- but if someone is using Zulip, updates, and resumes using Zulip within 10 minutes, the app will keep using the existing queue.
   * Poking around, the `APP_REFRESH` action that causes so many reducers to reset their part of the state is only triggered on queue expiry; nothing else.
   * One particular state subtree I'm concerned about is `caughtUp` -- this has booleans for each narrow, for "we have the oldest message in this narrow" and "we have the newest message in this narrow (and have an open event queue where we should learn about new ones)".
   * For the reason you explain in the commit message, if we could force an `APP_REFRESH` on this migration, that'd be an acceptable result. I'm not sure there's an easy way to do that as a quick hack, though.
   * Alternatively, if we're going to discard the `narrows` state, the Right Thing to do would be for the migration to also appropriately reset all the related state: `caughtUp`, maybe `flags`, maybe others (would have to check the list). This is probably not hard to do, and easier than hacking in an `APP_REFRESH`, so I think we might as well do it this way.

(still reading, cont'd...)

gnprice:
* It looks like most of the changed lines here are in test code; I think there may be an opportunity there to break the change up to make it easier to read.
   * This change touches a lot of lines, which always makes things more challenging to read and understand -- for me reviewing now, for you in your own review/rereading as you developed it, and for future readers understanding what we did. Sometimes that's unavoidable; but this is complex enough that I'm looking for ways to reduce the reading-complexity of the change, to try to reach whatever level is unavoidable.
   * In this change, it looks like most of the lines are in test code, specifically two kinds of quite mechanical changes: a lot of example narrows/messages data changing format, and also a lot of tests moving from chatReducers to messageReducers.
   * Possibly also some non-mechanical changes in the tests? It's hard to tell because there are so many of those two mechanical kinds. :-)
   * We don't break things at any commit, so we can't just change the app code first, then update the tests -- they're changing because this change would break them. However, we can *disable* tests. This may be a good time for that.
   * Specifically, here's a possible sequence:
      * Make the changes in the app/non-test code, and just skip the affected tests. See [Jest docs](https://jestjs.io/docs/en/api#describeskipname-fn), and also see our `getFilteredEmojiList-test.js` for a small example. Here, I'd probably just use `describe.skip` at the top level, with a terse one-line comment. These skips are truly temporary, as they'll be lifted later in the same branch.
      * Update the data format in tests, and unskip as appropriate. This will be a highly mechanical commit.
      * Move tests from chatReducers to messagesReducers, and unskip as appropriate. Another highly mechanical commit.
      * If there are any remaining skipped tests (any tests that need some other change), fix them and unskip. (Or if some of these are interesting, possibly do them up front in the main commit. Easy to make that decision later by squashing.)
   * This way, we sacrifice a bit of readability in that the first commit isn't quite as self-contained as usual -- it requires some test changes that it no longer contains. The commit message needs to make clear that's what's happening, and make clear to look at the next few commits to see those changes. But given  the complexity of this change, I think the readability we gain from focusing the main commit on the essential bits is much greater.

(still reading)

src/chat/chatReducers.js
```diff
@@ -19,139 +16,50 @@ import {
...
-      msg[key] = [...state[key], action.message];
+      msg[key] = [...state[key], action.message.id];
     } else {
       msg[key] = state[key];
     }
-
     return msg;
   }, {});
 
-  return stateChange ? newState : state;
```
gnprice: This `stateChange` logic kind of looks like it's doing nothing, but it's actually providing an optimization of a kind that's very important in the React architecture: when the new state is substantively unchanged from the old, it causes it to actually be the same state object, equal under `===`.

That's important because it allows downstream users of this part of the state -- notably React in handling components that get it as a prop, and Reselect in handling `createSelector` selectors where it's an argument -- to see that it's unchanged and cut off any downstream processing right here.

The exact structure of this code feels a little bit clumsy to me, and I'd welcome something to tighten it. (Separately from the main commit here, though it'd be fine to be in the same branch if that's convenient.) But we should keep the optimization. :-)

gnprice: As a bit of commit-style feedback, this simplification would have been best as a separate, preparatory commit. That way (in a world where the simplification is correct) you'd get to take full advantage of it in simplifying the code to be edited in the main commit, and at the same time it has its own commit message and can be read and understood separately. And keeps the complexity of the main commit down by not adding to the changes that are part of that commit.

roberthoenig:
Re state invalidation:

Agreed, we need to check which subtrees that get persisted could et invalidated by removing the `narrows` state.

> One particular state subtree I'm concerned about is caughtUp -- this has booleans for each narrow, for "we have the oldest message in this narrow" and "we have the newest message in this narrow (and have an open event queue where we should learn about new ones)".

`caughtUp` is part of `discardKeys`, so that gets cleared anyway -- this is interesting, because it means that effectively, caching `narrows` doesn't achieve anything at all? Anyway, no need to clear subtrees that already get cleared.

Here are the subtrees that get persisted:

`'migrations', 'accounts', 'drafts', 'outbox', 'settings', 'mute', 'narrows', 'realm', 'streams', 'subscriptions', 'unread', 'userGroups', 'users'`

Out of those, I think we need to reset ... none actually? So the code that's currently in place should be sufficient?

roberthoenig:
Fwiw, to give this analysis a bit more backing: I've been running the upgrade process from "before normalized narrow" to "after normalized narrow" some times now, and nothing seemed to be fishy during my (rather brief) insepctions of the interface.

src/message/messageReducers.js
```diff
@@ -0,0 +1,148 @@
...
+                  prev_subject: state[action.message_id].subject,
+                  timestamp: action.edit_timestamp,
+                  user_id: action.user_id,
+                },
+            ...(state[action.message_id].edit_history || NULL_ARRAY),
+          ],
+          last_edit_timestamp: action.edit_timestamp,
+        },
+      }
+    : state;
```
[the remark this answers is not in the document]
roberthoenig: Hrmm okay. That'll require inverting `action.message_id in state` to `!(action.message_id in state)` though, and negation is harder to read as well.

src/chat/chatReducers.js
roberthoenig: So my original plan was to ignore all optimizations initially and wait for someone with an old phone to benchmark the app, and then decide on whether to put optimizations back in. But I'll take your word for this one being critical :)

src/message/messageReducers.js
```diff
@@ -0,0 +1,148 @@
...
+          ...state[action.message_id],
+          content: action.rendered_content || state[action.message_id].content,
+          subject: action.subject || state[action.message_id].subject,
+          subject_links: action.subject_links || state[action.message_id].subject_links,
+          edit_history: [
+            action.orig_rendered_content
+              ? action.orig_subject
+                ? {
+                    prev_rendered_content: action.orig_rendered_content,
+                    prev_subject: state[action.message_id].subject,
```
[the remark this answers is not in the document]
roberthoenig: Makes sense!

src/chat/chatReducers.js
roberthoenig: Also it just sorta felt nice to have the core-core event `eventNewMessage` be simply

```js
const eventNewMessage = (state: MessagesState, action: EventNewMessageAction): MessagesState => ({
  ...state,
  [action.message.id]: action.message,
});
```

:P

roberthoenig: Hmm, thinking about it a bit more, I'm not sure we should keep this optimization: For one, I don't know how to implement it: This commit splits chatReducers into chatReducers and messageReducers, and messageReducers only has access to the new `messages` state. It is therefore unaware of the `narrows` state that is needed for this optimization (`narrows` corresponds to `chatReducers`). 

For two, I'm not sure we wanna do the optimization here: If we do proper optimization by carefully detecting unchanged arrays of messages (as discussed here https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/message.20state/near/618189), this shouldn't be necessary. Also, we need to do the unchanged-arrays-of-messages-optimization in both cases. 

So I'll wait for your input on this.

roberthoenig:
@gnprice updated! I didn't address the migration or optimization logic (see my comments above), but I fixed the early return style nit, and I decoupled the test fixups from the non-test changes.

src/message/__tests__/messageReducers-test.js
```diff
@@ -0,0 +1,300 @@
...
+import {
+  MESSAGE_FETCH_COMPLETE,
+  EVENT_NEW_MESSAGE,
+  EVENT_MESSAGE_DELETE,
+  EVENT_UPDATE_MESSAGE,
+  EVENT_REACTION_ADD,
+  EVENT_REACTION_REMOVE,
+} from '../../actionConstants';
+
+describe('chatReducers', () => {
```
borisyankov: You probably want to rename this.

src/message/__tests__/messageReducers-test.js
```diff
@@ -0,0 +1,300 @@
+import deepFreeze from 'deep-freeze';
```
borisyankov: Does this include both the changes and the rename?
Doing it in two steps will make the diff readable.

src/message/__tests__/messageReducers-test.js
roberthoenig: Aah, right!

src/message/__tests__/messageReducers-test.js
roberthoenig: Hmm, yeah, I thought about doing that, but then didn't because there are already a lot of commits that really shouldn't be separate if it weren't for readability, and someone who want to see the diff can still easily generate it. But I guess I can just as well do it.

gnprice: > there are already a lot of commits that really shouldn't be separate if it weren't for readability

True. I'd suggest squashing together the commits that are all doing the same test fixes on different test files. The result is a bit big (8 files changed, 291 insertions(+), 205 deletions(-)), but it's highly uniform and boring -- and each file and indeed each test can be read and understood independently -- so that's OK.

As always, I don't really at all mind having commits split "too finely" -- squashing them back together is just about the easiest kind of change for me to make, and if I were otherwise about to merge this I'd simply do that. Far better to err on the side of splitting too finely than too coarsely. :slightly_smiling_face: 

(Maybe leave the `flagsReducer` one separate, to keep the handy extra info in its commit message right next to the change? But I think it'd also be fine to squash it and just include that paragraph in the squashed commit message.)

gnprice:
> caughtUp is part of discardKeys, so that gets cleared anyway

Ah, good point!

> Out of those, I think we need to reset ... none actually?

Yeah, after looking through those I agree. Thanks for this analysis!

---

I did find a different issue in the course of looking at that, though! `flags` is interesting, because it's another that stores info closely related to what messages we have... but `flags` is one we *discard*. So, I thought: how does that get re-populated when we start up and have some cached messages?

Turns out: we have some logic to do that, but it was broken by your rename `messages` -> `narrows`. The reason that slipped through is that there's a line turning off type-checking :cry: :
```
const rehydrate = (state: FlagsState, action: RehydrateAction): FlagsState => {
  // $FlowFixMe
  const arrayOfMessageArrays: Array<Message[]> = Object.values(
    action.payload && action.payload.messages ? action.payload.messages : {},
  );
  const flattenedMessages: Message[] = Array.prototype.concat(...arrayOfMessageArrays);
  return processFlagsForMessages(state, flattenedMessages);
};
```
I'll do the rename now since that's very simple. Then this code also needs an update for this normalization change.

src/chat/chatReducers.js
gnprice: But here on the `chat` aka `narrows` side, I think we still need essentially the same optimization we have in the old code, for the same reason: most importantly, there will be a lot of messages coming in where we don't have that narrow, or we have it somewhere in the past but aren't caught up. And I think it should be easy to keep with basically the same logic.

gnprice: ... Hmm, now I see what you mean by your remark about the split, though: there's a sort of emergent property the existing design has, which is that if a message comes in that doesn't fall into any narrow we're tracking, then we won't store the message itself. And that does become tricky to preserve.

Is that optimization essential? Let's see...
* On a busy day chat.zulip.org (an especially large and busy org) might see 1000 messages. Say 100/hr.
* At that rate, the messages we collect over even a few hours aren't a big deal. And nobody will use the app continuously, with no breaks longer than 10 minutes (the event queue timeout), for more than a few hours at a time.
* So we don't need it right now.
* In the future, we intend to use a longer-lived event queue for some events; this is mainly for non-messages, but could include some messages (PMs? pinned streams?)
* Even then, I think we'll only do that for messages that we want to keep in any case. And if not... we can cross that bridge when we come to it.

So for *that* optimization, I agree -- let's just go without it.

We should keep the ones that are easy to keep, though. :smile: Including the one on new messages in `chatReducers` that's implemented with this `stateChange` logic -- which is a bit clumsy but OK, and which we can basically keep unchanged.

src/message/messageReducers.js
```diff
@@ -0,0 +1,157 @@
...
+): MessagesState => ({
+  ...state,
+  ...groupItemsById(action.messages),
+});
+
+const eventReactionAdd = (state: MessagesState, action: EventReactionAddAction): MessagesState => {
+  const oldMessage = state[action.message_id];
+  if (!(action.message_id in state)) {
+    return state;
+  }
```
gnprice: Style nit: I'd define `oldMessage` after you've done the `action.message_id in state` check, rather than before.

Unless you use it to *do* that check. I think `oldMessage === undefined`, or even `!oldMessage`, would be fine for that.

src/message/messageReducers.js
```diff
@@ -0,0 +1,157 @@
...
+
+const eventNewMessage = (state: MessagesState, action: EventNewMessageAction): MessagesState => ({
+  ...state,
+  [action.message.id]: action.message,
+});
+
+const eventMessageDelete = (
+  state: MessagesState,
+  action: EventMessageDeleteAction,
+): MessagesState => omit(state, action.messageId);
```
gnprice: `lodash.omit` seems to make a new object even if the key to be omitted is already absent. So this should get the `if (!(action.messageId in state))` treatment too.

src/chat/chatReducers.js
roberthoenig: So after looking a bit more into this, I found out how it'd be possible to preseve the isInNarro optimization: We'd need to pass the `narrows` state as part of `EventNewMessageAction`. This is in fact already done for another subtree of the state, `caughtUp` (which enables the caughtUp optimization.

So this optimization is doable without too much work, but it should go into a new commit. I'll not implement that commit now, but it should be doable as a follow-up. Also, since the isCaughtUp optimization is tightly coupled to the isInNarrow optimization, I'll skip the isCaughtUp one for now as well (at least for messageReducers.js).

roberthoenig:
@gnprice updated.

* Split the creation of messageReducers-tests.js into two commits
* Added back some of the optimizations
* Fixed the flags state update.
* Addressed the nits.

src/chat/flagsReducers.js
```diff
@@ -91,11 +91,9 @@ const processFlagsForMessages = (state: FlagsState, messages: Message[]): FlagsS
 
 const rehydrate = (state: FlagsState, action: RehydrateAction): FlagsState => {
   const messagesState: { [narrow: string]: Message[] } =
-    action.payload && action.payload.accounts ? action.payload.narrows : {};
+    action.payload && action.payload.accounts ? action.payload.messages : {};
   // $FlowFixMe: Object.values
-  const arrayOfMessageArrays: Array<Message[]> = Object.values(messagesState);
-  const flattenedMessages: Message[] = Array.prototype.concat(...arrayOfMessageArrays);
-  return processFlagsForMessages(state, flattenedMessages);
+  return processFlagsForMessages(state, Object.values(messagesState));
```
gnprice: As the commit message on 9df067b23 which fixed the bug here says, let's make a practice of keeping `$FlowFixMe` as tightly focused as possible -- applying it to more than it has to is what caused this bug.

Here, that means applying to just the `Object.values` call. Could do that with a local variable, as in the old code; or I think it should also work to do something like
```
  return processFlagsForMessages(state,
    // $FlowFixMe: Object.values
    Object.values(messagesState));
```

src/boot/store.js
```diff
@@ -41,14 +41,18 @@ export const storeKeys = ['migrations', 'accounts', 'drafts', 'outbox', 'setting
...
-  'mute', 'narrows', 'realm', 'streams', 'subscriptions', 'unread', 'userGroups', 'users',
+  'messages', 'mute', 'narrows', 'realm', 'streams', 'subscriptions', 'unread', 'userGroups', 'users',
 ];
 
 const migrations = {
   '0': state => {
     AsyncStorage.removeItem('reduxPersist:messages');
     return state;
   },
+  '1': state => ({
```
gnprice: Hmm. What happens when a user upgrades across *both* of these migrations at once? Which of course will be the usual case for users that aren't us.

I think the answer is that they'll get a `messages` subtree which is the old `messages`. Which means we should drop the `messages` subtree in one of these migrations.

I think the logical place for that is migration 0, the one that removed `messages`. When we originally did that migration, we reasoned that because there was no `messages` top-level reducer anymore, redux-persist wouldn't even look at `messages` in the persisted store, so the state passed to the migrations would already lack `messages`; but that isn't true anymore.

(If that's right, then one lesson for future migration style is that we should have done that deletion in the original migration in the first place, even though it seemed redundant at the time -- that way in the event of re-using the key later, everything would work without depending on us noticing the re-use and going back and adding the deletion.)

So I guess something like
```
  const { messages, ...restState } = state;
  return restState;
```
is in order.

src/message/messagesActions.js
```diff
@@ -24,7 +25,10 @@ export const doNarrow = (narrow: Narrow, anchor: number = FIRST_UNREAD_ANCHOR) =
   dispatch(navigateToChat(narrow));
 
   const allNarrows = getAllNarrows(state);
-  const messagesForNarrow = allNarrows[JSON.stringify(narrow)] || NULL_ARRAY;
+  const messages = getMessages(state);
+  const messagesForNarrow = (allNarrows[JSON.stringify(narrow)] || NULL_ARRAY).map(
+    id => messages[id],
+  );
```
gnprice: Orthogonal to this change, but reading this code in the course of rereading this whole change I notice that these lines would be simpler as `getFetchedMessagesForNarrow(narrow)(state)`.

src/types.js
```diff
@@ -86,8 +86,19 @@ export type PmRecipientUser = {
...
+ * Indexer over all narrows for which we have locally stored messages.
+ * @prop narrow - Map from narrows to lists of message IDs.
+ */
 export type NarrowsState = {
-  [narrow: string]: Message[],
+  [narrow: string]: number[],
+};
+
+/**
+ * Indexer over all locally stored messages.
```
gnprice: Thanks for adding these jsdoc comments! We should do more of these. :smile: 

The term "indexer" doesn't mean much to me here; I'll make a pass post-merge at trying to come up with another way to describe these.

(The one sense that comes to mind is in the context of a search engine's backend, where the "indexer" is a throughput-optimized system that collates giant quantities of data to produce/maintain the "index", which in turn is the data structure that the latency-optimized "server" consults to answer a query. E.g. [livegrep](https://github.com/livegrep/livegrep/blob/master/src/indexer.cc) uses this term, as does anybody that's worked at Google. But that's clearly not the sense you mean.)

gnprice:
Oh, also CI is failing with a type error. :-) Should be easy to fix.

roberthoenig:
Ah, before I forget it: I didn't manually test this last round of changes -- I solely relied on the test  coverage (which is actually quite decent for the messages state). So that should definitely be done before merging any of this.

roberthoenig:
> Oh, also CI is failing with a type error. :-) Should be easy to fix.

Yeah, I think that's because of the hotfix you pushed for the flags state? It gave me a red line in vscode.

roberthoenig:
Nevermind the last comment, master  works for me.

src/types.js
gnprice: The Flow error that's in CI right now shows me where Flow uses this term in a different way! Perhaps this is what you had in mind.

```
 • Message [1] is incompatible with array type [2] in the indexer property.
 • number [3] is incompatible with string [4] in the indexer property's key.
...
 [4][2]  93│   const messagesState: { [narrow: string]: Message[] } =
```

Or at least the term "indexer property". By which it seems to mean this kind of pseudo-property/meta-property that describes the object functioning like a map.

As a term for the object itself (an object that has an "indexer property" and no others), I'd just call it a "map".

roberthoenig: `Indexer` is actually Flow terminology, it's explained here: https://flow.org/en/docs/types/objects/#toc-objects-as-maps

gnprice:
> the hotfix you pushed for the flags state

Yeah -- basically the error is that my fix made explicit what the type of `messagesState` should be (because the next thing we do is just hand it to `Object.values` which is practically `any`, so without that there's no real type-checking on the `messagesState` initializer -- which is where the bug had been), and that type changes with your normalization so needs to be updated.

src/types.js
roberthoenig: Ah, your comment was quicker :) Yeah, `map` sounds good as well, probably better in the sense that one doesn't need to know Flow for it.

src/boot/store.js
roberthoenig: Aah, yeah, good catch. 

>   const { messages, ...restState } = state;
  return restState;

I agree this is the right thing to do, in migration 0.

gnprice:
And merged! Thanks again @roberthoenig for all your work on this.

Quick notes on the tweaks I made before merging:
* Updated the old migration, as discussed above (this is a new commit at the beginning.)
* In the "Extract tests" commit:
   * Moved the copied tests to the top, matching their order in the old file -- this reduces the diff when comparing old and new files.
   * Matched the nesting of the old file with `describe('messageReducers'` and `describe('MESSAGE_FETCH_COMPLETE'` -- this reduces the diff in the next commit, especially by making indentation constant, and does the same for the diff when comparing old and new files.
 * In the "Fixup and enable tests for messageReducers" commit:
   * Added back the blank lines between different tests; this is helpful for visually scanning the file, and also reduces the diff.
   * Split the new test into a followup commit; this keeps the big diff strictly boring and mechanical, making both the big boring/mechanical change and the short interesting change easier to read.
* Fixed that type error mentioned above.
* Fixed a couple of lint errors -- one in master (oops!) and one from the branch.
* Fixed one last skipped test :tada: . (Yes, I put that emoji in the commit message, in UTF-8. [`unicode`](https://packages.debian.org/unicode) is my new favorite command-line tool. `unicode party` -> a nice display about U+1F389 PARTY POPPER.) (Jest makes it really surprisingly hard to locate skipped tests: https://github.com/facebook/jest/issues/4763 . I used `yarn jest --verbose 2>&1 | grep -iC2 skip` to gather clues.)
* Narrowed the `$FlowFixMe` mentioned above.

gnprice:
Also just made that pass over those two new jsdoc comments, revising and expanding.

... Also rereading this thread I see this comment, which I saw before but forgot:

> Ah, before I forget it: I didn't manually test this last round of changes -- I solely relied on the test coverage (which is actually quite decent for the messages state). So that should definitely be done before merging any of this.

I guess I'll go do some manual testing. :grin:

PR #2843, opened 2018-07-27
Update message state on EVENT_UPDATE_MESSAGE_FLAGS.

### LINKED ISSUE #2627: message list: Show when a message is starred

### LINKED ISSUE #2676: A `starred` tag does not appear after starring a message in a stream.

jackrzhang:
Prior to this commit, dispatching the EVENT_UPDATE_MESSAGE_FLAGS action only updated `state.flags` and did not correspondingly update any fields in `state.messages`.

---

Fix #2676 by linearly searching through all messages in MessagesState, and add or remove the specified flag.

Before:

starred-messages-bug-2676

After:

starred-messages-fixed-2676

---

For #2677:

Adding newly starred messages entails:
(1) Caching newly starred messages in the local scope of `eventUpdateMessageFlags` in an `id -> Message` map
(2) Appending newly starred messages to the list of messages for the 'is:starred' narrow
(3) Sorting the 'is:starred' narrow by timestamp to maintain chronological order

Removing starred messages entails filtering out the ids in `action.messages` from the 'is:starred' narrow.

Before:

starred-list-bug-2677

After:

starred-list-fixed-2677

---

Resolve #2627.

This logic should be updated with @roberthoenig's [refactor of the message state](https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/message.20state/near/617688).

@gnprice 
@borisyankov

### DIFF
--- a/src/chat/chatReducers.js
+++ b/src/chat/chatReducers.js
@@ -2,6 +2,7 @@
 import isEqual from 'lodash.isequal';
 
 import type {
+  Message,
   MessagesState,
   MessageAction,
   MessageFetchCompleteAction,
@@ -10,6 +11,7 @@ import type {
   EventNewMessageAction,
   EventMessageDeleteAction,
   EventUpdateMessageAction,
+  EventUpdateMessageFlagsAction,
 } from '../types';
 import {
   APP_REFRESH,
@@ -22,8 +24,9 @@ import {
   EVENT_REACTION_ADD,
   EVENT_REACTION_REMOVE,
   EVENT_UPDATE_MESSAGE,
+  EVENT_UPDATE_MESSAGE_FLAGS,
 } from '../actionConstants';
-import { isMessageInNarrow } from '../utils/narrow';
+import { isMessageInNarrow, STARRED_NARROW_STR } from '../utils/narrow';
 import { groupItemsById } from '../utils/misc';
 import chatUpdater from './chatUpdater';
 import { NULL_ARRAY, NULL_OBJECT } from '../nullObjects';
@@ -149,6 +152,65 @@ const eventUpdateMessage = (
     last_edit_timestamp: action.edit_timestamp,
   }));
 
+const eventUpdateMessageFlags = (
+  state: MessagesState,
+  action: EventUpdateMessageFlagsAction,
+): MessagesState => {
+  /*
+    In order to add new messages to the 'is:starred' narrow, we cache newly
+    starred messages.
+
+    This local variable will no longer be necessary when the messages state is
+    refactored into a map of message_id -> message, as narrows will then be
+    represented as lists of ids, and referencing action.messages will suffice.
+  */
+  const messageCache: { [id: number]: Message } = {};
+
+  // Update the flags fields
+  const result = Object.keys(state).reduce((updatedState, narrow) => {
+    updatedState[narrow] = state[narrow].map(message => {
+      if (!action.messages.includes(message.id)) {
+        return message;
+      }
+
+      const updatedMessage = Object.assign({}, message);
+      if (action.messages.includes(message.id)) {
+        if (action.operation === 'add') {
+          updatedMessage.flags = [...message.flags, action.flag];
+          if (action.flag === 'starred') {
+            // local side effect: cache newly starred messages
+            messageCache[message.id] = updatedMessage;
+          }
+        } else {
+          // action.operation === 'remove'
+          updatedMessage.flags = message.flags.filter(flag => flag !== action.flag);
+        }
+      }
+
+      return updatedMessage;
+    });
+
+    return updatedState;
+  }, {});
+
+  // Update the 'is:starred' narrow
+  if (action.flag === 'starred') {
+    if (action.operation === 'add') {
+      result[STARRED_NARROW_STR] = [
+        ...result[STARRED_NARROW_STR],
+        ...Object.values(messageCache),
+      ].sort((a, b) => a.timestamp - b.timestamp);
+    } else {
+      // action.operation === 'remove'
+      result[STARRED_NARROW_STR] = result[STARRED_NARROW_STR].filter(
+        message => !action.messages.includes(message.id),
+      );
+    }
+  }
+
+  return result;
+};
+
 export default (state: MessagesState = initialState, action: MessageAction): MessagesState => {
   switch (action.type) {
     case APP_REFRESH:
@@ -174,6 +236,9 @@ export default (state: MessagesState = initialState, action: MessageAction): Mes
 
     case EVENT_UPDATE_MESSAGE:
       return eventUpdateMessage(state, action);
+
+    case EVENT_UPDATE_MESSAGE_FLAGS:
+      return eventUpdateMessageFlags(state, action);
 
     default:
       return state;
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -157,3 +157,5 @@ export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
 export const parseNarrowString = (narrowStr: string): Narrow => JSON.parse(unescape(narrowStr));
 
 export const STARRED_NARROW = specialNarrow('starred');
+
+export const MENTIONED_NARROW = specialNarrow('mentioned');
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -155,3 +155,5 @@ export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
   Array.isArray(narrow1) && Array.isArray(narrow2) && isEqual(narrow1, narrow2);
 
 export const parseNarrowString = (narrowStr: string): Narrow => JSON.parse(unescape(narrowStr));
+
+export const STARRED_NARROW = specialNarrow('starred');
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -157,5 +157,6 @@ export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
 export const parseNarrowString = (narrowStr: string): Narrow => JSON.parse(unescape(narrowStr));
 
 export const STARRED_NARROW = specialNarrow('starred');
+export const STARRED_NARROW_STR = JSON.stringify(STARRED_NARROW);
 
 export const MENTIONED_NARROW = specialNarrow('mentioned');
--- a/src/chat/__tests__/chatReducers-test.js
+++ b/src/chat/__tests__/chatReducers-test.js
@@ -8,12 +8,14 @@ import {
   groupNarrow,
   streamNarrow,
   topicNarrow,
+  STARRED_NARROW_STR,
 } from '../../utils/narrow';
 import {
   MESSAGE_FETCH_COMPLETE,
   EVENT_NEW_MESSAGE,
   EVENT_MESSAGE_DELETE,
   EVENT_UPDATE_MESSAGE,
+  EVENT_UPDATE_MESSAGE_FLAGS,
   EVENT_REACTION_ADD,
   EVENT_REACTION_REMOVE,
 } from '../../actionConstants';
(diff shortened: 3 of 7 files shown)


jackrzhang:
@zulipbot add "area: redux"

src/utils/narrow.js
```diff
@@ -155,3 +155,12 @@ export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
   Array.isArray(narrow1) && Array.isArray(narrow2) && isEqual(narrow1, narrow2);
 
 export const parseNarrowString = (narrowStr: string): Narrow => JSON.parse(unescape(narrowStr));
+
+const starredNarrow: Narrow = [
```
borisyankov: You can export this too. We do export both the stringified and the array narrows.

src/utils/narrow.js
```diff
@@ -155,3 +155,12 @@ export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
...
+  {
+    operator: 'is',
+    operand: 'starred',
+  },
+];
+
+export const starredNarrowStr = JSON.stringify(starredNarrow);
```
borisyankov: This can get merged as a separate commit, if needed.
Not sure if we want to do that though, as are currently generating this narrow via `specialNarrow`.

borisyankov: If we want to keep it, maybe we would do something like `starredNarrow = specialNarrow('starred')` and then replace every use of `specialNarrow('starred')` with `starredNarrow`

jackrzhang: I'll update the commit to use `specialNarrow`.

jackrzhang: I can split the commit and update every use of `specialNarrow('starred')` to `starredNarrow`. For consistency, then, we should also have similar exports for every use of `specialNarrow`.

Should we use upper snake case to denote certain narrows that are constant? I.e. `HOME_NARROW`, `HOME_NARROW_STR`, `STARRED_NARROW`, etc.?

borisyankov: @jackrzhang yeah that UPPER_CASE naming is a good idea. Thanks for noticing that. 
I did go through the source code and created a PR for that #2848

gnprice: Yeah, this sounds like a good plan -- replace the uses of `specialNarrow('starred')` with a new `STARRED_NARROW`, as a separate preparatory commit.

I just merged #2848 , so CONSTANT_CASE will blend in nicely.

gnprice: And I think doing so for the other special narrows -- `mentioned` and `private` are the ones I see -- would be good too. In fact we already have `ALL_PRIVATE_NARROW` for one of those.

src/chat/chatReducers.js
```diff
@@ -149,6 +152,63 @@ const eventUpdateMessage = (
...
+        }
+      }
+
+      return updatedMessage;
+    });
+
+    return updatedState;
+  }, {});
+
+  // Update the 'is:starred' narrow
```
gnprice: The `is:mentioned` narrow needs the same logic, I think.

Hrm, but the webapp doesn't seem to do anything with flags other than `starred` and `read`. In fact the code is kind of beautifully simple, just a few lines:
https://github.com/zulip/zulip/blob/32438cde9/static/js/server_events_dispatch.js#L402

... wait. It doesn't actually work. Basically it has a type error. :-P And, empirically, if I'm looking at some messages in the webapp and I go and star or unstar some of them from mobile, or from another tab on web -- nothing happens. I guess I'll go file that bug in zulip/zulip.

There's another lesson here, though: the reason nobody noticed that was broken until now is that the webapp has *local echo* for starring a message.

We should do local echo too, independent of this fix; on a slow network, or if you're just on a different continent from your server, you don't want to wait 200ms to see the star appear.

src/chat/chatReducers.js
```diff
@@ -149,6 +152,63 @@ const eventUpdateMessage = (
...
+  if (action.flag === 'starred') {
+    if (action.operation === 'add') {
+      result[starredNarrowStr] = [
+        ...result[starredNarrowStr],
+        // $FlowFixMe
+        ...Object.keys(messageCache).map(id => messageCache[id]),
```
gnprice: How about `Object.values(messageCache)`?

src/chat/chatReducers.js
gnprice: Also, none of this kind of stuff is documented, but I think the server may not send this event (relying on `update_message` instead?) for a normal `mentioned` change. So let's not worry about that one for now.

borisyankov:
There is a big issue that needs discussing (or reworking) regarding this approach.
Unfortunately, this was not properly documented, nor did we discuss it in our several sessions with Greg or anyone.

Have a look at the `flagsReducers`. They were created while we used the native rendering, so there is a chance we can come up with another implementation that is now better (or not, not sure, just mentioning the point in time for context).

The undocumented, but a key aspect of `flags` is that we treat the `message.flags` data as read-only, load-on-start, then ignore. This is an important performance optimization. The important flag is the `read` flag. It changes **very often** for a message list; while scrolling past 50 messages, it might change once per message (we do mark several messages so possibly less), but none of the messages actually need re-rendering.

In an ideal world, we would have a data structure that contains message data that relates to how a message is rendered and anything that is not related is stored somewhere else, possibly in several normalized data stuctures etc. For simplicity and performance sake (we don't want to do too much data manipulation for too little benefit) we keep the messages as they come from the backend.

The `flags` store is organized in a way that in a component if you get a dependency on `flags.starred` you will need ro render only if `flags.starred` did change.

As I mentioned, this was conceived with native rendering when we had a single `Message` component per backend `message` object and the optimization worked a bit differently (every component was `connect()-ed`).

src/chat/chatReducers.js
jackrzhang: Makes sense, updated.

src/chat/chatReducers.js
```diff
@@ -149,6 +152,63 @@ const eventUpdateMessage = (
     last_edit_timestamp: action.edit_timestamp,
   }));
 
+const eventUpdateMessageFlags = (
+  state: MessagesState,
+  action: EventUpdateMessageFlagsAction,
+): MessagesState => {
+  /**
```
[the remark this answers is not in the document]
jackrzhang: Updated.

src/chat/chatReducers.js
```diff
@@ -149,6 +152,63 @@ const eventUpdateMessage = (
...
+    represented as lists of ids, and referencing action.messages will suffice.
+  */
+  const messageCache: { [id: number]: Message } = {};
+
+  // Update the flags fields
+  const result = Object.keys(state).reduce((updatedState, narrow) => {
+    updatedState[narrow] = state[narrow].map(message => {
+      const updatedMessage = Object.assign({}, message);
+
+      if (action.messages.includes(message.id)) {
```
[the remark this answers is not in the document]
jackrzhang: Sounds great! Updated.

src/chat/chatReducers.js
jackrzhang: I've opened #2852.

I've never heard of the term *local echo* before - a quick [wikipedia search](https://en.wikipedia.org/wiki/Echo_(computing)) provides context:
> ... **local echo**, where the sending device itself displays the sent data

If I'm understanding that correctly, I think what you're referring to here is identical in implementation to an '[optimistic update](https://stackoverflow.com/questions/33009657/what-is-optimistic-updates-in-front-end-development)'.

src/utils/narrow.js
jackrzhang: I've added two commits for `starred` and `mentioned`.

src/chat/chatReducers.js
gnprice: Ah, indeed!

"Optimistic update" isn't a term I know / that I see used in the communities I know. But yes, that's exactly what I mean.

As that Wikipedia article suggests, the term "local echo" comes from talking about interactions where the user is typing some text, and that text gets "echoed" locally without waiting for the server to echo it back. This turns out to make a big difference in the pleasantness of the interaction, if the server is more than a few ms away; it's one of the major features [mosh](https://mosh.org/) is proud of in contrast with `ssh`. :)

We use it a lot in the Zulip project in describing how when the user sends a message, we display the message immediately rather than wait for it to come back from the server. Search for "locally echo" here:
https://zulip.readthedocs.io/en/latest/subsystems/markdown.html

gnprice:
@borisyankov Thanks for that background!

There's an important point here which is maybe a bit implicit: currently, even though we're not updating the flags inside the message state, we are continuing to use it! Specifically, `renderMessagesAsHtml` has `flags: message.flags || []` in constructing a property that gets passed down through all the webview rendering path.

And that's basically what this bug is -- we're keeping around a stale version of the message's flags, and we're using that stale version for rendering the message.

It looks like that may be the one place we're using the `flags` off a message in the store; there's a place in `caughtUpReducers` that looks like it at first glance, but it's actually looking at the messages in a `MESSAGE_FETCH_COMPLETE` action.

So, as a fix for this bug, how about we fix the webview to use `flags` from the state instead of from the messages? I think that would also allow us to save all the code this PR was going to have to add to `chatReducers`. :-)

Then I'd want to follow that up immediately by deleting `flags` from the messages we put in the store. That doesn't seem like a lot of processing. And as this bug demonstrates, having an extra copy of the data that just gets stale is a source of confusion and bugs.

gnprice:
> So, as a fix for this bug, how about we fix the webview to use flags from the state instead of from the messages?

Oh, and then of course we'll need an appropriate optimization for the `read` flags, because those change so frequently.

I think this is basically a job for `MessageListWeb#shouldComponentUpdate`, or I guess for `getInputMessages` which it delegates this job to.

In fact, rereading `getInputMessages`, I'm reminded that it currently looks only at `renderedMessages`, which is a bit misnamed and is really sort of a rearranging of the messages and inserting of data for time and recipient headers.

So part of the fix, on this strategy, will also have to be to add to the condition in `getInputMessages` so that we update if the flags have changed, too.

And then this optimization will just be that we *don't* update on account of `flags.read` changing -- only the other flags.

jackrzhang:
> So, as a fix for this bug, how about we fix the webview to use flags from the state instead of from the messages? I think that would also allow us to save all the code this PR was going to have to add to chatReducers. :-)

Agreed 😄 

I distinctly recall wondering how or if the key-value pairs in `state.flags` were being used while I was making this PR. So it's great to have more context into that! Something I'm more curious about now is how `state.flags` is constructed when the app is initialized and how the data from the server API for that is shaped.

Based on my understanding of the design that's been proposed, here's a possible outline of the changes that should be made:
* webview: Refer to `state.flags` instead of `messages.flags`.
* (Rename `getInputMessages` to `processRenderedMessagesProp`(?))
* `getInputMessages`: Do not update on change in `state.flags`.
* Remove the `flags` property from messages in `state.messages`.

gnprice:
> Something I'm more curious about now is how `state.flags` is constructed when the app is initialized

So this part, you can read in `flagsReducers.js`; that's the code that initializes (to `initialState`) and maintains that data.

> and how the data from the server API for that is shaped.

You'll see there that it comes from several different places -- some with a list of messages (and this code just picks off the flags information), and some from an `update_message_flags` event.

> Based on my understanding of the design that's been proposed, here's a possible outline of the changes that should be made:

Good idea. Here's how I'd outline it:

* renderMessagesAsHtml: Refer to `state.flags` (via `getFlags`) instead of `message.flags`.
* `getInputMessages`: Update on changes to `flags` as well as to `renderedMessages` -- except don't update on changes to `flags.read`.
* Remove the `flags` property from messages in `state.messages`.

I'd also be glad to find a clearer name for `getInputMessages`. The thing is that "message" in that name doesn't mean what we mean by "message" in most of the app -- it's referring to a "message" in the [`window.postMessage` API](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) which we use to communicate with the JS inside the webview. The job of this function we call `getInputMessages` is to compute a list of "messages" in that sense for us to send to the webview.

gnprice:
Ongoing work in this direction is happening in #2922 ; closing this one. Thanks again!

PR #2847, opened 2018-07-28
notifications: Handle GCM to remove messages from the notification

kunall17:
(More information in the commit message)

Server PR https://github.com/zulip/zulip/pull/10094
Docs: https://github.com/zulip/zulip/blob/6c151c2459bc91a207eabdc50b2734df1e8e20f5/docs/production/mobile-push-notifications.md#removing-sent-notifications-from-mobile-as-messages-read-on-server


borisyankov:
I'll try to do more in-depth review later...
I think you should extract the notification-related code outside of `MainApplication.java` file and keep it lean (the new code and the code you did add previously)

borisyankov:
For clarity and to be consistent with our latest improved commit practices, it will be best if you extract the current code in a separate commit, and then add the new code in another commit.

kunall17:
Thanks for the quick review, i'll split out the commits @borisyankov

kunall17:
Cool! :) 👍

PR #2848, opened 2018-07-29
Use consistent constant upper case names

borisyankov:
There is a widely-used convention of naming constants in `UPPER_CASE`.

It is also officially recommended in MDN's guide on `const`:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

We mostly follow that convention (`actionConstants.js`, `nullObjects.js`, `styles/index.js` etc.) but not in all cases that we should.

I examined `ESLint` rules and plugins that might help us.
One plugin that was a good candidate is:
https://github.com/k03mad/eslint-plugin-const-case
Unfortunately, it didn't fit our needs, and additionally, it got confused by Flow types.

It is a bit tricky to determine what value should be UPPER_CASE automatically.
A lot of values are using the `const` keyword, but few of them need to be considered 'constants' in the sense we know. I have followed this:
 * they should be a value, not a function
 * they should not be a local variable in a function
 * they are almost always exported and reused elsewhere


gnprice:
Looks good, merged, thanks!

> It is a bit tricky to determine what value should be UPPER_CASE automatically.
> A lot of values are using the const keyword, but few of them need to be considered 'constants' in the sense we know.

Yeah. One JS style guide I often like is Google's -- it happens to have [a pretty nice little writeup](https://google.github.io/styleguide/jsguide.html#naming-constant-names) on this question:
> Every constant is a `@const` static property or a module-local `const` declaration, but not all `@const` static properties and module-local `const`s are constants. Before choosing constant case, consider whether the field really feels like a deeply immutable constant. For example, if any of that instance's observable state can change, it is almost certainly not a constant. Merely intending to never mutate the object is generally not enough.

(Then some examples.)

gnprice:
PS -
> It is also officially recommended in MDN's guide on `const`:

A thing that's not at all obvious looking at MDN but is sometimes important to know is that it's a lot like Wikipedia. Anyone can edit it, and many people do. Like certain subject areas on Wikipedia, it's generally accurate on concrete facts and generally the best available reference, partly because (I assume) there are some people who make it their thing to take good care of it; but there can be gaps in that, and it doesn't represent the official word from Mozilla or anything like that.

In this case, there are 76 usernames listed at the bottom of that page. Looks like this particular advice was mainly introduced in [this commit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const$compare?to=871843&from=845755) by one "agutoli", the [only change](https://developer.mozilla.org/en-US/dashboards/revisions?user=agutoli) that person has ever made on MDN. Which is fine, since the advice seems good from outside information; but a good illustration of the general point about MDN. :-)

Issue #2845, opened 2018-07-27
Search result highlights in topic headers render as literal HTML


gnprice:
In search results, we highlight the terms that matched the query. This works great in the message body; but when the match is in the topic, we show a literal `<span class="highlight">` and `</span>`, instead of highlighting the match as intended.

![image](https://user-images.githubusercontent.com/28173/43341373-23e7eaae-9194-11e8-91a0-1c7a558546f3.png)

This was probably broken by my 994145383 "webview: Escape *everything* that isn't intended as HTML.", which fixed a bug (and potential security vulnerability) where we'd interpret the actual text of the topic name as HTML. So the root cause of this one is that we're not distinguishing between
* a topic name `new job` with `new` highlighted (which should render as "**new** job"), and
* a topic name that really is `<span class="highlight">new</span> job` (which should render as "&lt;span class="highlight"&gt;new&lt;/span&gt; job").

PR #2849, opened 2018-07-30
webview: Fix search result in topic headers render as literal HTML.

jainkuniya:
Add `$!` before topic while generating header HTML, to include a literal.

Fix: #2845

Before
screen shot 2018-07-30 at 10 35 11 am

After
screen shot 2018-07-30 at 10 34 39 am

### DIFF
--- a/src/webview/html/messageHeaderAsHtml.js
+++ b/src/webview/html/messageHeaderAsHtml.js
@@ -59,7 +59,7 @@ export default ({
     # ${item.display_recipient}
   </div>
   <div class="header topic-text" data-narrow="${topicNarrowStr}">
-    ${item.match_subject || item.subject}
+    $!${item.match_subject || item.subject}
   </div>
 </div>
     `;


gnprice:
This is wrong -- it reintroduces the previous, much worse bug that was fixed by 994145383, the commit I linked to in #2845 .

As I wrote there:

> So the root cause of this one is that we're not distinguishing between
> * a topic name `new job` with `new` highlighted (which should render as "**new** job"), and
> * a topic name that really is `<span class="highlight">new</span> job` (which should render as "&lt;span class="highlight"&gt;new&lt;/span&gt; job").

A correct fix will need to treat those cases differently.

gnprice:
I've just pushed 46cacfce8 to fix this issue; take a look there to see how I avoided duplicating the template. Also note that there are two places we use `match_subject` -- it's important to cover them both so as to treat it consistently.

The bulk of the work I needed to do in order to make that fix was studying the server code (and checking my understanding with how the webapp handles it) to produce this information I put in the commit message:

```
Fortunately, the server correctly handles the interaction of `subject`
being text and `match_subject` needing to use HTML, in that before
marking up the topic, it encodes it (*) as HTML.  So, we can simply
take `match_subject` as the HTML to display.
```

In fact, in doing so I discovered a bug in the server's handling of this subject matter! zulip/zulip#10588 . Fortunately it only applies in a pretty obscure situation which we never even exercise (and in which the webapp ignores the affected data anyway), so we get to ignore the bug. But it illustrates how the answers to these questions can't just be assumed.

PR #2850, opened 2018-07-30
refactor: Convert ViewPlaceholder component to functional.

jainkuniya:
Functional component are stateless and in this component
there is no need of any state and lifecycle methods
so it's better to use functional component in such sisuation.

Also creation and updation of class based component is costly in terms
of time.

### DIFF
--- a/src/common/ViewPlaceholder.js
+++ b/src/common/ViewPlaceholder.js
@@ -1,5 +1,5 @@
 /* @flow */
-import React, { PureComponent } from 'react';
+import React from 'react';
 import { View } from 'react-native';
 
 type Props = {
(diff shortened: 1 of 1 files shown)


borisyankov:
Actually, 'changes requested` is not the correct status.
I would consider it more of 'should not be merged' a correct description.

jainkuniya:
Even I am hot sure about it.
But functional component simplifies a code.

I have just tried this few months back. But forgot to open PR.
If we all agrees with  `PureComponent ` we can close this PR :)

gnprice:
I'm pretty happy with our current style of always using a class. I think Boris's point about certain changes becoming a big diff because they have to change the form from function to class is a particularly strong reason.

Issue #2846, opened 2018-07-28
Realm emoji don't work in autocomplete or reactions


gnprice:
If you're composing a message and start typing an emoji's name, like `:smi`, we show a nice autocomplete popup with possibilities like `:smile:`.

If the emoji you have in mind is a realm emoji, though (or a Zulip extra emoji -- i.e. `:zulip:`, though in principle there could be more in the future), it won't appear. Also, if people have reacted to a message with a realm emoji or a/the Zulip extra emoji, those won't show up right (that's #2129, though I think we've since patched it up so at least we don't literally say "undefined".)

Happily both kinds work in the one most important place, which is rendering a message. But these other places are important too.

I think the basic root cause here is that our code for passing around the information about realm emoji is very confusing and it's hard to figure out what value is supposed to be what -- e.g. whether a prop called `realmEmoji` is supposed to be an object representing one emoji, or a map indexed by emoji code / id, or a map indexed by name. So the main step will be to fix that, which means refactoring and renaming a lot of pieces, adding type annotations where we don't already have them, and adding jsdoc in a few places for things to know that can't be expressed in the name and the types.

(I have some more specific thoughts on what a solution will look like, which I'll try to write up in a bit.)

Once that's done, I think the bugs as seen by the user will be straightforward to fix.

---
For cross-reference:

There's some work on this in the latter part of PR #2775, and in #2307 in commits a5a24a81e "realmEmoji: Fix realm emojis are not appearing in autocomplete" and c24432ed5 "reactions: Fix clicking on reactions doesn't work".

In addition to #2129, this may be the cause of #2806.

agzuniverse:
@zulipbot claim

jainkuniya:
@agzuniverse on which part are you planning to work 🤔. There is already PR for this #2307

agzuniverse:
Modifying the realm emoji selectors as I did in #2775 fixes the problem of realm emoji not showing up the app at all. This is the simplest fix to this problem as all the code in the app is already set up to work with a specific structure of the `relmEmoji` object, it's just that this is not the structure returned by the selector right now.

The commit with this modification was not merged in that PR, so I'll be doing that in a clearer way and opening a PR for that.

0rmi:
@agzuniverse @gnprice reading through these commits, do I assume correctly that the issue is fixed in agzunierse's fork and I could get working realm emotes already by building his code?

PR #2851, opened 2018-07-31
Add support for realm emoji

### LINKED ISSUE #2129: Realm emoji reactions showing as "undefined"
I'm running on v11.8.79 (android 7.1.1). This should have #1698 merged, right?.
My realm emoji's aren't showing in the app. What's going on?
screenshot

agzuniverse:
Created new selectors `getAllRealmEmojiByName` and `getActiveRealmEmojiByName` that returns the correct object structure expected by the rest of the application for the `realmEmoji` object.

Also added realm emoji support to  `EmojiPickerScreen`, `EmojiAutocomplete` and `EmojiRow` which was removed in 0c2d40f797c02a4cbd11515a3a99adc469fe2fc8

Fixes #2129 and partially #2846 (Zulip extra emoji is not yet supported)

The build is failing right now because the argument `emojiCode`  for `emojiReactionAdd()` and `emojiReactionRemove()` is expected to be a string, but in the case of realm emoji, the emoji ID is passed as `emojiCode` and this is a number.

### DIFF
--- a/src/api/emoji_reactions/emojiReactionAdd.js
+++ b/src/api/emoji_reactions/emojiReactionAdd.js
@@ -6,7 +6,7 @@ export default (
   auth: Auth,
   messageId: number,
   reactionType: string,
-  emojiCode: string,
+  emojiCode: string | number,
   emojiName: string,
 ): Promise<ApiResponse> =>
   apiPost(auth, `messages/${messageId}/reactions`, res => res, {
--- a/src/autocomplete/EmojiAutocomplete.js
+++ b/src/autocomplete/EmojiAutocomplete.js
@@ -8,11 +8,11 @@ import { Popup } from '../common';
 import EmojiRow from '../emoji/EmojiRow';
 import getFilteredEmojiList from '../emoji/getFilteredEmojiList';
 import type { GlobalState, RealmEmojiState } from '../types';
-import { getActiveRealmEmojiById } from '../selectors';
+import { getActiveRealmEmojiByName } from '../selectors';
 
 type Props = {
   filter: string,
-  realmEmoji: RealmEmojiState,
+  activeRealmEmojiByName: RealmEmojiState,
   onAutocomplete: (name: string) => void,
 };
 
@@ -20,8 +20,8 @@ class EmojiAutocomplete extends PureComponent<Props> {
   props: Props;
 
   render() {
-    const { filter, realmEmoji, onAutocomplete } = this.props;
-    const emojis = getFilteredEmojiList(filter, realmEmoji);
+    const { filter, activeRealmEmojiByName, onAutocomplete } = this.props;
+    const emojis = getFilteredEmojiList(filter, activeRealmEmojiByName);
 
     if (emojis.length === 0) {
       return null;
@@ -34,7 +34,13 @@ class EmojiAutocomplete extends PureComponent<Props> {
           initialNumToRender={12}
           data={emojis}
           keyExtractor={item => item}
-          renderItem={({ item }) => <EmojiRow name={item} onPress={() => onAutocomplete(item)} />}
+          renderItem={({ item }) => (
+            <EmojiRow
+              realmEmoji={activeRealmEmojiByName[item]}
+              name={item}
+              onPress={() => onAutocomplete(item)}
+            />
+          )}
         />
       </Popup>
     );
@@ -42,5 +48,5 @@ class EmojiAutocomplete extends PureComponent<Props> {
 }
 
 export default connect((state: GlobalState) => ({
-  realmEmoji: getActiveRealmEmojiById(state),
+  activeRealmEmojiByName: getActiveRealmEmojiByName(state),
 }))(EmojiAutocomplete);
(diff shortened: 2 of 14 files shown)


gnprice:
Thanks @agzuniverse ! This will be a really useful fix.

First bit of feedback: as always, it's really important that each commit make sense on its own, to make it easy for a reader to understand what's happening.

Two important cases of that in this branch:
* The first commit un-exports `getActiveRealmEmoji` and `getAllRealmEmoji` -- which means that everything that uses them will break. That can't be right.
* The first plus second commit together change everything that used those to use the `...ByName` selectors instead. But there were some parts of the app that worked correctly with realm emoji before -- so won't this break those? In particular, the message list.

Another thought on the code:
* I think one of the reasons this code was so broken for so long is that it was hard to tell what a value called `realmEmoji` was supposed to be. So let's fix that here. When it comes from the selector `getActiveRealmEmojiByName`, let's call that prop `activeRealmEmojiByName`; if there is a use for the old `getAllRealmEmoji`, let's call that `allRealmEmojiByCode`; etc. It's a little verbose... but verbosity is better than confusion. (Often, types save us from having to make that sort of thing explicit in the names. But our types don't help with distinctions like "all" vs. "active".)

After thinking through that, here's what I think would be a good sequence of commits for this PR:
1. First, add the new ByName selectors. Don't yet un-export the old ones.

2. Then, go through all the places where we use the old selectors. For each one,
   a. change the prop's name to what you think the contents should be -- like `activeRealmEmojiByName`
   b. change the selector to match
   c. change all the places that prop gets passed, down through other components and through helper functions, to also match.

   Where possible, make one commit for one call site of a selector, together with all the places underneath it. (The most important thing is that each commit makes sense on its own and doesn't break anything.)

3. Then if you do eliminate all uses of one or both of the old selectors, go ahead and un-export.

4. Then add the support for realm emoji and re-enable tests, etc., just like in your later commits in this branch.

agzuniverse:
PR has been updated as per the feedback you gave @gnprice

gnprice:
Thanks @agzuniverse ! This is definitely an improvement. I've just merged parts of this branch: 4 commits squashed from 6 of yours, making some preparatory changes and fixing one component; plus some other changes of my own to try to help make this area clearer.

Do take a look at 94234c324..7dc06ff87 to see those changes, and the adjustments I've made. In particular:
* Isolated fixups that are already needed and don't depend on any other changes can go at the beginning of the branch, making them easy to merge. Here, good examples include clarifying the existing tests and removing the `RealmEmoji = any`.
* I made a series of my own refactors, which I hope will make these selectors a bit less confusing to read. After your commit adding two new ones, I added a commit making the same changes there.

For the remaining changes, including the fixes to the actual behavior for users:
* As I said above, I really do want to make sure we use specific names like `activeRealmEmojiByName` rather than `realmEmoji`, in order to prevent future confusion like what led to this logic being broken in the first place.
   * In particular, I don't want to merge any attempted fixes that keep using the `realmEmoji` names; those just won't fix the problem in a durable way.
   * This includes the commits affecting `getFilteredEmojiList` and `EmojiRow`.
* Also the `getFilteredEmojiList` commit makes that function start using its `realmEmoji` argument, even though it's being passed garbage there. That's broken! You can actually demonstrate buggy behavior in the app that way (just start typing the numeric ID of some realm emoji.)
* In `EmojiRow`, the new `realmEmoji` prop doesn't make much sense. We seem to be treating it like just a boolean? But we pass a whole `RealmEmojiType` object.
   * Also, the `EmojiRow` commit adds that prop as required, but doesn't update its callers to pass it. That makes things broken at this commit.
   * Here's a design that I think would work: have `EmojiRow` get `activeRealmEmojiByName` with its own `connect` call.
   * Better yet: in `EmojiRow`'s `connect` call, say
     ```
     oneRealmEmoji: getActiveRealmEmojiByName(state)[props.name],
     ```
     so that it picks out just the one piece of that state it actually cares about. Note that will be `undefined` when the user chose a Unicode emoji.
* A good order to go about this in would be:
   * `EmojiRow` starts getting its own `activeRealmEmojiByName` (or `oneRealmEmoji`). Now it will work -- if only its callers can get the right list of emoji names.
   * `getFilteredEmojiList` starts calling its parameter `activeRealmEmojiByName`, and its two callers fix the selectors they use and the prop name to match.
      * It *doesn't* yet start behaving differently -- there's plenty of noise in this diff with all the renaming, so it's best if the interesting stuff is a separate commit.
   * The `addReaction` logic in `EmojiPickerScreen` gets updated to handle realm emoji.
   * Now `getFilteredEmojiList` starts including realm emoji -- now that everything that'll be affected is ready to handle it. (Or if I've missed some other spot, prepare that first too.)
* Similarly, the `SearchMessagesCard` commit breaks things. After all, why is that `realmEmoji` prop there in the first place? It's there to get passed to `MessageList`... which this doesn't yet update to expect the new name instead of the old.

This is a bug that's had a long life; I'll be glad to get it finally squashed, and I think this line of changes should do so pretty solidly when complete. :-)

agzuniverse:
Updated the commits as per you suggestions, then cherry picked the commits that re-enabled tests and updated the type for `emojiCode`. Squashed the commits for `SearchMessagesCard` and `MessageList`.

gnprice:
Thanks @agzuniverse for your work on this!

I looked again at this branch, and most of the remarks I made in my last comment still applied.

So I spent today carrying out my own plan described above :stuck_out_tongue: , and then continuing with some further changes: see #3170 for the whole thing. Closing this, as these issues are now resolved.

Issue #2751, opened 2018-07-02
Type error in webview's getMessageIdFromNode


borisyankov:
Webview throws an exception: `msgNode.getAttribute` is not a function.

This happens both on Android and iOS (the only other known issue is Android-specific)

chrisbobbe:
It looks like 65e3ea3bd0b5e451253a27e6696b0c56badc9527 types `node` as `?Node` instead of `Node`, but this wouldn't catch the problem Greg pointed out:

> So somehow we're getting a `msgNode` that is truthy, but doesn't have a `getAttribute` method.

Also, getMessageIdFromNode was deprecated in 7f968e5685372d38b6f370f89835f7a1c1fcf45c, but it still has six call sites.

gnprice:
Looking at Sentry events, we've had no cases of this in (at least) the past 90 days:
https://sentry.io/organizations/zulip/events/?project=191284&query=%22msgNode.getAttribute%22&statsPeriod=90d

To try to confirm that's not just a lack of data: the way these would get reported to Sentry is that the same `window.onerror` which puts up that red banner (or a less detailed one, in release builds) also posts an event through `sendMessage`, which gets processed here in `webViewEventHandlers.js`:
```js
    case 'error':
      logging.error(`WebView exception: ${JSON.stringify(event.details)}`);
      break;
```

And we do get other exceptions reported through that path:
https://sentry.io/organizations/zulip/events/?project=191284&query=%22WebView+exception%22+%22TypeError%22&statsPeriod=14d

So it seems this got fixed at some point. :man_shrugging: If we hear a fresh report, we'll reopen.

Issue #2558, opened 2018-05-23
SyntaxError on invalid token in webview


gnprice:
This looks a lot like #2505 , but the specific syntax error is different.

See https://sentry.io/zulip/zulip-mobile/issues/554709208/events/20750069553/ :
> "message":"Uncaught SyntaxError: Invalid or unexpected token"
> "line":1,"column":44,"error":{}

We also [got a report](https://chat.zulip.org/#narrow/stream/48-mobile/subject/illegal.20token/near/583456) from a user on 13.2.86, with a message that's worded differently but seems to mean the same thing, and the same line and column numbers:
![image](https://user-images.githubusercontent.com/28173/40453631-adf1d17e-5e9a-11e8-8ba9-98d79256acc7.png)

One key thing I don't know and am curious about is, what is this a syntax error *in*? What is being parsed, and what is trying to parse it?

gnprice:
Fixed in 787c4a53b (from #2854) a couple of weeks ago, along with #2505 and #2538.

Issue #2538, opened 2018-05-20
Fix rare exception in webview on Android


borisyankov:
Rendering some messages (very rarely, but consistently) throw an error on Android.

borisyankov:
It seems the error happens when we send some html messages to the phone (when transitioning from 'loading' to 'some messages loaded' state).
When the messages already exist, they are rendered properly.

gnprice:
@borisyankov says this issue may be related to #2505 and #2558.

gnprice:
Fixed in 787c4a53b (from #2854) a couple of weeks ago, along with #2505 and #2558.

Issue #2505, opened 2018-05-14
"SyntaxError: Unexpected identifier" in webview


borisyankov:
https://sentry.io/zulip/zulip-mobile/issues/554709208/

```
Error: {"message":"Uncaught SyntaxError: Unexpected identifier","source":"https://chat.zulip.org/","line":1,"column":212137,"error":{}}
    at src/webview/webViewEventHandlers.toUpperCase(webViewEventHandlers.js:172)
    logErrorRemotely(new Error(JSON.stringify(event.details)), 'WebView Exception');
    at src/webview/MessageListWeb.this(MessageListWeb.js:40)
    const handler = `handle${eventData.type.charAt(0).toUpperCase()}${eventData.type.slice(1)}`;
...
(13 additional frame(s) were not displayed)
```

gnprice:
More details in #2640 from @adambirds, who ran into an instance of this issue.

Felixeltoro:
VM1790:1 Uncaught SyntaxError: Unexpected identifier

gnprice:
> VM1790:1 Uncaught SyntaxError: Unexpected identifier

@Felixeltoro Is that a description of an error you're seeing? If so we'd be very interested to hear more details -- please file a new issue for it:
https://github.com/zulip/zulip-mobile/issues/new

PR #2854, opened 2018-08-01
Base64 encode message before sending to WebView

borisyankov:
This fixes the infamous red errors in the WebView with the very misleading error numbers.

Messages sent on Android are ` URL decoded` so a message containing `%22` (or maybe several other similar sequences) are interpreted and produce invalid data.

See the commits for more details.

Once I've figured out the issue I managed to search for and identify multiple bug reports to React Native that relate to this. (our JS code not being able to contain `//` comments is related)

There were multiple bug fix PR attempts, but none got merged.

Related reported issues:
https://github.com/facebook/react-native/issues/20069
https://github.com/facebook/react-native/issues/19611

Related bug-fix PRs:
https://github.com/facebook/react-native/pull/17394
https://github.com/facebook/react-native/pull/19655
https://github.com/facebook/react-native/pull/20366

### DIFF
--- a/src/utils/__tests__/encoding-test.js
+++ b/src/utils/__tests__/encoding-test.js
@@ -5,6 +5,7 @@ import {
   hexToAscii,
   asciiToHex,
   xorHexStrings,
+  strToBase64,
   extractApiKey,
 } from '../encoding';
 
@@ -61,11 +62,40 @@ describe('asciiToHex', () => {
   });
 });
 
+describe('strToBase64', () => {
+  test('can handle an empty string', () => {
+    const obj = '';
+    const expected = '';
+
+    const result = strToBase64(obj);
+
+    expect(result).toBe(expected);
+  });
+
+  test('can encode any string', () => {
+    const obj = {
+      key: 'ABCabc123',
+      empty: null,
+      array: [1, 2, 3],
+    };
+    const expected = 'W29iamVjdCBPYmplY3Rd';
+
+    const result = strToBase64(obj);
+
+    expect(result).toBe(expected);
+  });
+
+  test('supports unicode characters', () => {
+    const obj = { key: '😇😈' };
+    const expected = 'W29iamVjdCBPYmplY3Rd';
+    const result = strToBase64(obj);
+    expect(result).toBe(expected);
+  });
+});
 describe('extractApiKey', () => {
   test('correctly extracts an API key that has been XORed with a OTP', () => {
     const key = 'testing';
     const otp = 'A8348A93A83493';
-
     const encoded = xorHexStrings(asciiToHex(key), otp);
     expect(extractApiKey(encoded, otp)).toBe(key);
   });
--- a/src/utils/encoding.js
+++ b/src/utils/encoding.js
@@ -32,6 +32,11 @@ export const base64ToHex = (bytes: string) => asciiToHex(base64.decode(bytes));
 
 export const hexToBase64 = (hex: string) => base64.encode(hexToAscii(hex));
 
+// https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
+// `base64.encode` used instead of `btoa` to support Unicode input
+export const strToBase64 = (text: string): string =>
+  base64.encode(unescape(encodeURIComponent(text)));
+
 // Extract an API key encoded as a hex string XOR'ed with a one time pad (OTP)
 // (this is used during the OAuth flow)
 export const extractApiKey = (encoded: string, otp: string) =>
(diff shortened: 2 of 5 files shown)


src/utils/encoding.js
```diff
@@ -32,6 +32,9 @@ export const base64ToHex = (bytes: string) => asciiToHex(base64.decode(bytes));
 
 export const hexToBase64 = (hex: string) => base64.encode(hexToAscii(hex));
 
+export const objToBase64 = (obj: Object | Array<*>): string =>
+  base64.encode(unescape(encodeURIComponent(JSON.stringify(obj))));
```
gnprice: This is a pretty funny-looking sequence (encode, then *unescape*, then different encode?). It should have a comment pointing to an explanation. :-)

This one is pretty good:
https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding

Also, I'd rather separate this into two pieces:
* JSON.stringify -- the caller can do this, it knows its business
* this funny dance to turn what's already a perfectly good string, into something that will be inert under percent-decoding and basically any other reinterpretation.

I'd also want the name to be a bit more specific, because there's more going on to what this encoding means than "convert to base64". It's converting to base64 in a very particular way, which is that it's turning the string into UTF-8 (through this hack of `unescape(encodeURIComponent(...))`), and base64-encoding that bytestream. If the code interpreting this only knew "it's base64", that wouldn't be enough to interpret it correctly with confidence.

So I think the primitive I'd want here is something like this:

`export const base64Utf8Encode  = (text: string): string => base64.encode(unescape(encodeURIComponent(text)))`

gnprice:
Also, the commit message lists five of our issues:
> #2538 #2505 #2538 #2558 #2751

Those are actually four issues, as there's a duplicate :). And #2751 says it happens on iOS too -- so perhaps that one isn't fixed by this, and is something else?

gnprice:
Thanks!

> I did not call it base64Utf8Encode as the JS string format is not UTF-8 but one of the two UCS-2 or UTF-16 thus called it strToBase64; added the link to MDN and the spec for string encoding.

Right, so, this is why what it's doing is kind of odd and it's good to be explicit:

What `text => unescape(encodeURIComponent(text))` does is precisely to take the Unicode string that we had that is a nice normal JS string (with UTF-16 or UCS-2 internally, but that doesn't matter for this particular API)... and encode it as UTF-8. That is, anywhere that the string has a codepoint with a value greater than 255, it replaces that by a sequence of characters each no more than 255, and it uses UTF-8 to do it.

The reason that's what happens is that encodeURIComponent encodes things as UTF-8, and follows up by encoding those bytes with percent-encoding. The job of the `unescape` is to undo the percent-encoding; but it doesn't undo the UTF-8, it just leaves us with a sequence of characters that each represent one of those bytes of UTF-8.

Here's an example, which I did just now in the Chrome devtools console:
```
> '😈';
"😈"
> encodeURIComponent('😈');
"%F0%9F%98%88"
> unescape(encodeURIComponent('😈'));
"ð"
> unescape(encodeURIComponent('😈')) === '\xf0\x9f\x98\x88';
true
```
And in fact the four bytes `f0 9f 98 88` are the UTF-8 encoding of U+1f608, which is...
```
$ python3 -c 'print("\U0001f608")'
😈
```

gnprice:
And perhaps the other half of that explanation: the reason that that's what we *want* to happen, or anyway what we choose to have happen, is that unfortunate error "The string to be encoded contains characters outside of the Latin1 range." from either `btoa` or the equivalent from `base64.encode`.

In other words, what they really want to be encoding is bytes. They take their input in the form of a JS string, because that's the type that's available... but then they insist that all the characters in the input actually have values no more than 255, i.e. they can interpret their values as bytes. And one way of saying that is that they need to be "in the Latin1 range", because iso-8859-1 aka latin-1 is the character set that consists of the characters in the low 256 codepoints of Unicode (i.e. 0 to 255), each represented as the byte with the same value.

src/utils/__tests__/encoding-test.js
```diff
@@ -61,11 +62,40 @@ describe('asciiToHex', () => {
...
+    const expected = 'W29iamVjdCBPYmplY3Rd';
+
+    const result = strToBase64(obj);
+
+    expect(result).toBe(expected);
+  });
+
+  test('supports unicode characters', () => {
+    const obj = { key: '😇😈' };
+    const expected = 'W29iamVjdCBPYmplY3Rd';
```
gnprice: I think this does not test what you want it to test: :-p
```
$ echo W29iamVjdCBPYmplY3Rd | base64 -d ; echo '$'
[object Object]$
```

gnprice:
Merged as 787c4a53b -- thanks again!

PR #2855, opened 2018-08-01
Change the data structure for saving notifications data

kunall17:
Previously it was saved as

```
LinkedHashMap<String, Pair<String, Integer>> conversations;
```

Now it is saved as

```
LinkedHashMap<String, List<MessageInfo>> conversations;
```

MessageInfo contains the content and the message id.
This commit does no visual changes, only changes the underlaying data structure and functions related to it.

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/MainApplication.java
+++ b/android/app/src/main/java/com/zulipmobile/MainApplication.java
@@ -4,7 +4,6 @@ import android.app.Application;
 import android.app.NotificationManager;
 import android.content.Context;
 import android.os.Bundle;
-import android.util.Pair;
 
 import com.RNFetchBlob.RNFetchBlobPackage;
 import com.facebook.react.ReactApplication;
@@ -26,6 +25,7 @@ import com.wix.reactnativenotifications.core.notification.INotificationsApplicat
 import com.wix.reactnativenotifications.core.notification.IPushNotification;
 import com.zmxv.RNSound.RNSoundPackage;
 import com.zulipmobile.notifications.GCMPushNotifications;
+import com.zulipmobile.notifications.MessageInfo;
 import com.zulipmobile.notifications.PushNotificationsProp;
 import com.zulipmobile.RNSecureRandom;
 
@@ -40,7 +40,7 @@ import static com.zulipmobile.notifications.NotificationHelper.addConversationTo
 import static com.zulipmobile.notifications.NotificationHelper.clearConversations;
 
 public class MainApplication extends Application implements ReactApplication, INotificationsApplication {
-    private LinkedHashMap<String, Pair<String, Integer>> conversations;
+    private LinkedHashMap<String, List<MessageInfo>> conversations;
 
     private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
         @Override
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -39,9 +39,12 @@ public class GCMPushNotifications extends PushNotification {
     public static final String ACTION_NOTIFICATIONS_DISMISS = "ACTION_NOTIFICATIONS_DISMISS";
 
     /**
-     * {@link LinkedHashMap} is for holding the message notifications data and group them according to their narrow collection that contains.
-     * {@link String} for the narrow key of the message.
-     * {@link List<MessageInfo>} for the list of messages received in that narrow.
+     * The Zulip messages we're showing as a notification, grouped by conversation.
+     *
+     * Each key identifies a conversation; see @{link buildKeyString}.
+     *
+     * Each value is the messages in the conversation, in the order we
+     * received them.
      */
     private LinkedHashMap<String, List<MessageInfo>> conversations;
 
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -11,7 +11,6 @@ import android.os.Build;
 import android.os.Bundle;
 import android.text.TextUtils;
 import android.util.Log;
-import android.util.Pair;
 
 import com.wix.reactnativenotifications.core.AppLaunchHelper;
 import com.wix.reactnativenotifications.core.AppLifecycleFacade;
@@ -24,6 +23,7 @@ import com.zulipmobile.R;
 import java.io.IOException;
 import java.net.URL;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Locale;
 
 import me.leolin.shortcutbadger.ShortcutBadger;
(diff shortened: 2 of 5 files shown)


android/app/src/main/java/com/zulipmobile/notifications/NotificationHelper.java
```diff
@@ -65,22 +66,22 @@ public static String extractName(String key) {
...
             inboxStyle.addLine(sb);
         }
     }
 
-    public static int extractTotalMessagesCount(LinkedHashMap<String, Pair<String, Integer>> conversations) {
+    public static int extractTotalMessagesCount(LinkedHashMap<String, List<MessageInfo>> conversations) {
         int totalNumber = 0;
-        for (Map.Entry<String, Pair<String, Integer>> map : conversations.entrySet()) {
-            totalNumber += map.getValue().second;
+        for (Map.Entry<String, List<MessageInfo>> asd : conversations.entrySet()) {
```
gnprice: `asd`?

android/app/src/main/java/com/zulipmobile/notifications/MessageInfo.java
```diff
@@ -0,0 +1,27 @@
...
+    private String content;
+    private int messageId;
+
+    public MessageInfo(String content, int messageId) {
+        this.content = content;
+        this.messageId = messageId;
+    }
+
+    public String getContent() {
+        return content;
```
gnprice: Man, this makes me wish for [Kotlin](https://kotlinlang.org/docs/reference/data-classes.html) or something. :stuck_out_tongue:

```
data class MessageInfo(val content: String, val messageId: int)
```
96% fewer lines.

In fact, writing that out makes me think: do you want those setters? I think we actually don't -- I'd rather these things be immutable. That makes things easier to reason about.

(You don't seem to currently use the setters, happily.)

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -37,14 +37,14 @@
 
     public static final int NOTIFICATION_ID = 435;
     public static final String ACTION_NOTIFICATIONS_DISMISS = "ACTION_NOTIFICATIONS_DISMISS";
-    private LinkedHashMap<String, Pair<String, Integer>> conversations;
+    private LinkedHashMap<String, List<MessageInfo>> conversations;
```
gnprice: I'd really like a javadoc comment on this saying
* what these keys are
* that the lists have the latest message first.

That last bit, in particular, is a (useful!) bit of complexity that is added here, so this is a good time to start writing stuff down.

android/app/src/main/java/com/zulipmobile/notifications/NotificationHelper.java
```diff
@@ -65,22 +66,22 @@ public static String extractName(String key) {
         return key.split(":")[0];
     }
 
-    public static void buildNotificationContent(LinkedHashMap<String, Pair<String, Integer>> conversations, Notification.InboxStyle inboxStyle, Context mContext) {
-        for (Map.Entry<String, Pair<String, Integer>> map : conversations.entrySet()) {
+    public static void buildNotificationContent(LinkedHashMap<String, List<MessageInfo>> conversations, Notification.InboxStyle inboxStyle, Context mContext) {
+        for (Map.Entry<String, List<MessageInfo>> map : conversations.entrySet()) {
```
gnprice: (This is for a followup, because it's pre-existing)
This variable is not a map -- it's an entry in a map. `entry` would be a good name.

Similarly below.

android/app/src/main/java/com/zulipmobile/notifications/NotificationHelper.java
```diff
@@ -65,22 +66,22 @@ public static String extractName(String key) {
...
             String name = extractName(map.getKey());
 
             Spannable sb = new SpannableString(String.format(Locale.ENGLISH, "%s%s: %s", name,
-                    mContext.getResources().getQuantityString(R.plurals.messages, map.getValue().second, map.getValue().second),
-                    map.getValue().first));
+                    mContext.getResources().getQuantityString(R.plurals.messages, map.getValue().size(), map.getValue().size()),
```
gnprice: (This is for a followup, because it's pre-existing)
This repeated `map.getValue()` would be clearer... well, first by saying `entry`, but then by giving this expression a name:
```
List<MessageInfo> messages = entry.getValue();
```
I'd put that right at the top of the loop, just under the `...entry.getKey()` so the two of them are immediately under the `for` line.

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -37,14 +37,14 @@
 
     public static final int NOTIFICATION_ID = 435;
     public static final String ACTION_NOTIFICATIONS_DISMISS = "ACTION_NOTIFICATIONS_DISMISS";
-    private LinkedHashMap<String, Pair<String, Integer>> conversations;
```
gnprice: Separately: how about specifically saying something like `ArrayList` here?

android/app/src/main/java/com/zulipmobile/MainApplication.java
```diff
@@ -40,7 +41,7 @@
 import static com.zulipmobile.notifications.NotificationHelper.clearConversations;
 
 public class MainApplication extends Application implements ReactApplication, INotificationsApplication {
-    private LinkedHashMap<String, Pair<String, Integer>> conversations;
```
gnprice: Should clean up the now-unused `Pair` import.

android/app/src/main/java/com/zulipmobile/notifications/MessageInfo.java
kunall17: Yeah, I wonder if kotlin is possible with react native, never researched about this.

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
kunall17: Primarily I wrote List here because I think we might change it too LinkedList (details in the follow-up comment)

kunall17:
@gnprice How about we use LinkedList instead of ArrayList? 
It has to perform these main operations

- Traversing each message to find a certain message id (For deletion) 
- Deleting a MessageInfo
- Inserting message in the end

Deletion is faster in LinkedList but traversing is faster in ArrayList & other reason for LinkedList is after deletion the memory is freed up pretty easily and quickly which is not the case for ArrayList

android/app/src/main/java/com/zulipmobile/MainApplication.java
kunall17: Done

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
gnprice: Thanks for adding a comment! It didn't actually quite answer either of my two questions, though. :-)

I added a commit on top, cf1cd2a1a, to add those details; take a look to see what I had in mind.

gnprice:
Merged -- thanks!

> LinkedList instead of ArrayList

So actually, one thing that I noticed when working out what the answers were to my questions to put in that doc comment :grin: is: it looks like we're currently sticking each message on the end of the list; but what we should probably actually be doing is keeping them sorted by message ID. That way they're always in the order the sender sent them in. (I'm not sure if GCM promises things will be delivered in order; even if it does, between the Zulip server and the notification bouncer I don't think we have that guarantee.)

That operation feels a lot more natural on an ArrayList than a LinkedList.

Better yet, though: how about a TreeMap, indexed by message ID? That keeps things sorted, and also makes deletion very simple and efficient.


> other reason for LinkedList is after deletion the memory is freed up pretty easily and quickly which is not the case for ArrayList

This doesn't sound right to me -- I'd actually expect the opposite to be true, because a LinkedList will contain a separate piece of garbage for each entry in the list. But I don't think this is likely to be a big concern either way; this shouldn't be a lot of data.

kunall17:
Yeah, there is no sorting done according to the ID because currently the messages are being grouped into a separate list for different narrows, so in a particular narrow I don't think there can be a case where we receive a notification of a later message first than an older message, so If we're using a single list or a map (from message id to details) then it could have been a problem.

The main performance increase we can get if we group these into different narrows, a TreeMap would have been a good option, anyways mostly the number of messages would be really low so I don't think we should be worrying so much about the data structure.

gnprice:
> so in a particular narrow I don't think there can be a case where we receive a notification of a later message first than an older message

Why are you sure of that?

I don't see a reason to be sure messages will always come in order. If we don't have a good reason to be sure that assumption is true, we should write the code to behave correctly without making the assumption. Which means sorting the messages by ID (or inserting them into an ordered data structure), rather than just putting them on the end.

gnprice:
Just now I happened across this orange warning box [in the docs for FCM](https://firebase.google.com/docs/cloud-messaging/concept-options#collapsible_and_non-collapsible_messages):
> FCM does not guarantee the order of delivery.

That's the behavior I'd expect from a service like this, so it's nice to see it explicit in the docs. I'd be surprised if GCM makes a stronger guarantee here than FCM does.

In any case, I wouldn't want to make that guarantee on behalf of the Zulip app server and the Zulip push notification bouncer. We should assume that messages may come out of order.

PR #2864, opened 2018-08-04
TopicAutocomplete: Use simple string match for topic filtering

### LINKED ISSUE #2857: Crash in topic autocomplete due to parsing as regexp
To reproduce:
* Narrow to a stream.
* Open the compose box.
* Focus the topic input, and type a `(`.
* The app crashes.

Learned about this from a Sentry crash report ([link](https://sentry.io/zulip/zulip-mobile/issues/629372472/)). Here's the traceback:
```
SyntaxError: SyntaxError: Invalid regular expression: missing )

This error is located at:
    in t
    in Connect(t)
    in RCTView
[...]
  File "src/autocomplete/TopicAutocomplete.js", line 35
    const topicsToSuggest = topics.filter(x => x && x !== text && x.match(new RegExp(text, 'i')));
  File "src/autocomplete/TopicAutocomplete.js", line 32, in mode
    return null;
```

That was 15.0.92 on iOS. I reproduce on 16.1.94 on Android.

It looks like we're taking the text the user entered, and trying to parse it as a regular expression. That's doomed. If we wanted a regexp, we could take the text and escape it as a JS regexp; but also we're not particularly using regexp features here, so probably best is to just rewrite this without `RegExp`.

punchagan:
Using a RegExp on user input crashes the app on invalid inputs like entering a
single `(`. The search seems to have been using a RegExp search only for
enabling case-insensitive search. This has now been switched to a simple
sub-string test using lower-cased strings.

Fixes #2857

### DIFF
--- a/src/autocomplete/TopicAutocomplete.js
+++ b/src/autocomplete/TopicAutocomplete.js
@@ -32,7 +32,9 @@ class TopicAutocomplete extends PureComponent<Props> {
       return null;
     }
 
-    const topicsToSuggest = topics.filter(x => x && x !== text && x.match(new RegExp(text, 'i')));
+    const topicsToSuggest = topics.filter(
+      x => x && x !== text && x.toLowerCase().indexOf(text.toLowerCase()) > -1,
+    );
 
     return (
       <AnimatedScaleComponent visible={topicsToSuggest.length > 0}>


gnprice:
Thanks @punchagan for the swift fix! Merged.

Issue #2868, opened 2018-08-04
Emoji in message list don't render if names have changed. 


jainkuniya:
<img width="292" alt="screen shot 2018-08-05 at 12 41 50 am" src="https://user-images.githubusercontent.com/18511177/43679803-6bf01acc-9849-11e8-8720-1100e42e023b.png">

@zulipbot claim

gnprice:
This looks to me like a case of #2846. Does that fit what you're seeing?

jainkuniya:
#2846 if for realm emoji right?

But this issue is for unicode emoji. :)

armaanahluwalia:
Can confirm. Have a look at https://chat.zulip.org/#narrow/stream/179-Testing/topic/test.20topic on mobile 18.0.99

armaanahluwalia:
Additionally trying "Add a reaction" with ```umbrella_on_ground`` doesn't work at all.

gnprice:
Fixed in #2869 / c8f4d0030 .

PR #2869, opened 2018-08-04
emoji: Update codePointMap.

jainkuniya:
Generated map from the current zulip master
(1d2f8bed92e6f04f72dc7e472b815421436c73b2).

Run `tools/setup/emoji/build_emoji` to get map.
https://github.com/zulip/zulip/blob/master/docs/subsystems/emoji.md

Fix: #2868

Now
screen shot 2018-08-05 at 12 48 48 am


gnprice:
Thanks!
* Can we sort this list in the source code? That'd make it possible to glance over the diff and get a sense of what's changing for real.
* Specifically, let's have one commit that just sorts what we have, then another that does the update (still sorted).
* Not sure whether I'd rather have it sorted by name or by codepoint. (Thoughts?) Either would be way better than not sorted, though.
* Can you spell out exactly what steps you took to produce this file? That'd be extremely helpful information for the commit message, or maybe better yet a comment in this file.
* If you aren't already, I recommend playing with `jq` for those steps... in fact, `<static/generated/emoji/emoji_codes.js perl -lne 'print $1 if (/^exports.name_to_codepoint = (.*);$/)' | tr \' '"' | jq -S .` might be a good answer.

Also
* The one thing I'd like to be sure of before we merge this is that it won't cause problems for people using older versions of the server (... or at least I'd like to know just *what* problems it'll cause.) Can you explain what you think this will do in that case?

HarshitOnGitHub:
>If you aren't already, I recommend playing with jq for those steps... in fact, <static/generated/emoji/emoji_codes.js perl -lne 'print $1 if (/^exports.name_to_codepoint = (.*);$/)' | tr \' '"' | jq -S . might be a good answer.

We don't need to extract this mapping from `emoji_code.js` file. `static/generated/emoji/name_to_codepoint.json` file contains this mapping in JSON format.

>The one thing I'd like to be sure of before we merge this is that it won't cause problems for people using older versions of the server (... or at least I'd like to know just what problems it'll cause.) Can you explain what you think this will do in that case?

This should cause no problem to them except that they will have mismatch of names in mobile app and webapp just like users connecting to latest version of server.

gnprice:
> We don't need to extract this mapping from emoji_code.js file. static/generated/emoji/name_to_codepoint.json file contains this mapping in JSON format.

Ah indeed, thanks!

In that case, a better command is: `<static/generated/emoji/name_to_codepoint.json jq -S .`

> This should cause no problem to them except that they will have mismatch of names in mobile app and webapp just like users connecting to latest version of server.

Just to make this explicit so we can be sure I understand correctly: I think you're saying that with this change, a user using the updated app with an old server will see exactly the same kind of symptoms as users are seeing right now with an updated server. E.g., the autocomplete will offer `:peace:`, but then the server won't understand that (it thinks the name should be `:peace_symbol:`), so the message will literally say `:peace:` instead of :peace_symbol: . Is that right?

In that case I think that's perfectly acceptable to merge as a quick fix. Thanks for that clarification!

And also we should follow up by sorting out a better solution. :slightly_smiling_face: 

@jainkuniya Just one small point beyond what's above:
* CI is failing. It looks like a formatting issue, so should be easy to fix.

HarshitOnGitHub:
@gnprice 
>Just to make this explicit so we can be sure I understand correctly: I think you're saying that with this change, a user using the updated app with an old server will see exactly the same kind of symptoms as users are seeing right now with an updated server. E.g., the autocomplete will offer :peace:, but then the server won't understand that (it thinks the name should be :peace_symbol:), so the message will literally say :peace: instead of peace_symbol . Is that right?

Yep, exactly! Just to be more explicit they will also not be able to add new reactions with that emoji. By that I mean that if some has already reacted with that emoji then they will be able to click on the reaction and +1 to it but if they are the first person to do so they will get a emoji info mismatch exception. This has been explained more clearly here:
https://github.com/zulip/zulip/blob/412dbde2e0e6c56709689a311b8d3ea1077ec490/zerver/views/reactions.py#L52

@jainkuniya 
>Current zulip master is generating map with emoji code as key and name
as value. I have swaped key and value and sorted it via name to have
better git diff.

I think you are looking at wrong file. You should be looking at `name_to_codepoint.json` not `codepoint_to_name.json` the later maps emoji codes to their canonical names while the former maps supported names to corresponding emoji code. So if you will use that you will only get canonical names in mobile app not all the names we support.

jainkuniya:
Thanks @HarshitOnGitHub for the correction. Sorry, I missed it, my mistake 😢 .

Updated 👍

gnprice:
Thanks @jainkuniya and @HarshitOnGitHub ! Merged as c8f4d0030 .

Issue #2870, opened 2018-08-05
Error is not shown in alert when message edit is failed.


jainkuniya:
There is already outdated PR for this #2560.

@zulipbot claim

ray-kraesig:
Closing duplicate in favor of #2902.

PR #2871, opened 2018-08-05
message edit: Fix error msg is not shown when msg edit fails.

jainkuniya:
After a0cc3dc, it became quite easier to handle api fail. In this
commit, just get the correct object from error which contains error
message.

Fix: #2870

### DIFF
--- a/src/utils/info.js
+++ b/src/utils/info.js
@@ -2,14 +2,14 @@
 import { Alert } from 'react-native';
 import Toast from '@remobile/react-native-toast';
 
-export const showToast = (message: string) => {
+export const showToast = (message: string): void => {
   Toast.show(message, Toast.SHORT);
 };
 
-export const showErrorAlert = (message: string, title: string) =>
+export const showErrorAlert = (message: string, title: string): void =>
   Alert.alert(title, message, [{ text: 'OK', onPress: () => {} }], { cancelable: true });
 
-export const showErrorFromException = (error: Error, title: string) =>
+export const showErrorFromException = (error: Error, title: string): void =>
   // $FlowFixMe
   Alert.alert(title, error && error.data && error.data.msg, [{ text: 'OK', onPress: () => {} }], {
     cancelable: true,
--- a/src/utils/info.js
+++ b/src/utils/info.js
@@ -8,3 +8,9 @@ export const showToast = (message: string) => {
 
 export const showErrorAlert = (message: string, title: string) =>
   Alert.alert(title, message, [{ text: 'OK', onPress: () => {} }], { cancelable: true });
+
+export const showErrorFromException = (error: Error, title: string) =>
+  // $FlowFixMe
+  Alert.alert(title, error && error.data && error.data.msg, [{ text: 'OK', onPress: () => {} }], {
+    cancelable: true,
+  });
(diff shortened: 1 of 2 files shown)


src/compose/ComposeBox.android.js
```diff
@@ -235,7 +235,7 @@ class ComposeBox extends PureComponent<Props, State> {
     const subject = topic !== editMessage.topic ? topic : undefined;
     if (content || subject) {
       updateMessage(auth, { content, subject }, editMessage.id).catch(error => {
-        showErrorAlert(error.message, 'Failed to edit message');
+        showErrorAlert(error.data.msg, 'Failed to edit message');
```
borisyankov: You are making an assumption that the `error` object will always have a `data` property. If that assumption happens to be false, this will throw.

From the looks of the code, your assumption is currently true. Our API code should always throw through `makeApiError` and it seems that `removeEmptyValues` which is called from `updateMessage` is not going to throw...

Still, it might be a good idea to create a separate version of `showErrorAlert` that checks for `error.data`  before using the `msg` value. `showErrorFromException` or some similar name.

jainkuniya: @borisyankov Done 👍

src/utils/info.js
```diff
@@ -6,3 +6,10 @@ export const showToast = (message: string) => Toast.show(message, Toast.SHORT);
 
 export const showErrorAlert = (message: string, title: string) =>
   Alert.alert(title, message, [{ text: 'OK', onPress: () => {} }], { cancelable: true });
+
+export const showErrorFromException = (error: Error, title: string) =>
+  error
```
borisyankov: This code looks a bit too smart for its own good.
The meaning of it is to return the value but you don't want that. `Alert.alert` has a return type of `void`.
https://github.com/facebook/react-native/blob/master/Libraries/Alert/Alert.js#L47

Related to that `showErrorAlert` is also not that correct, just doesn't look as weird :)

src/utils/info.js
```diff
@@ -6,3 +6,10 @@ export const showToast = (message: string) => Toast.show(message, Toast.SHORT);
...
+  && error.data
+  && error.data.msg
+  // $FlowFixMe
+  && Alert.alert(title, error.data.msg, [{ text: 'OK', onPress: () => {} }], { cancelable: true });
```
borisyankov: You can do the `error && error.data && error.data.msg` as a second parameter.
Not only will this be more simple, will not make Flow complain (Flow is correct to complain!) but it will also show the error with the `title` text regardless if the error object is defined.

gnprice:
@jainkuniya I just rebased this onto master and tried it out. On trying to edit a message which is too old, I got a modal with an empty body:

image

Is that what you expect, or were you seeing a more informative error message? We probably want the latter :slightly_smiling_face: 

In any case, would you (a) rebase the branch, and (b) post a screenshot of an example of what the UI looks like after this change? Thanks!

jainkuniya:
> Is that what you expect, or were you seeing a more informative error message? We probably want the latter 🙂

Yup, we want more informative error message. Actually we are getting error message from server, we just need to use that.

agrawal-d:
Closed via https://github.com/zulip/zulip-mobile/pull/3797.

PR #2874, opened 2018-08-06
Only show the notification with an event message

kunall17:
Fixes a bug in which previously the notification was trying to show
messages with event remove which was showing up as `null:null` as it
didn't have any payload for the narrow and the content

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/MainApplication.java
+++ b/android/app/src/main/java/com/zulipmobile/MainApplication.java
@@ -85,11 +85,12 @@ public class MainApplication extends Application implements ReactApplication, IN
             clearConversations(conversations);
             NotificationManager nMgr = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
             nMgr.cancelAll();
-            return null;
         } else {
             PushNotificationsProp prop = new PushNotificationsProp(bundle);
-            addConversationToMap(prop, conversations);
-            return new GCMPushNotifications(context, bundle, defaultFacade, defaultAppLaunchHelper, new JsIOHelper(), conversations);
+            if (prop.getEvent().equals("message")) {
+                addConversationToMap(prop, conversations);
+            }
         }
+        return new GCMPushNotifications(context, bundle, defaultFacade, defaultAppLaunchHelper, new JsIOHelper(), conversations);
     }
 }
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -90,6 +90,10 @@ public class GCMPushNotifications extends PushNotification {
         // First, get a builder initialized with defaults from the core class.
         final Notification.Builder builder = super.getNotificationBuilder(intent);
 
+        if (!getProps().getEvent().equals("message")) {
+            return builder;
+        }
+
         String type = getProps().getRecipientType();
         String content = getProps().getContent();
         String senderFullName = getProps().getSenderFullName();
@@ -189,5 +193,13 @@ public class GCMPushNotifications extends PushNotification {
     @Override
     protected int createNotificationId(Notification notification) {
         return NOTIFICATION_ID;
+    }
+
+    @Override
+    public void onReceived() throws InvalidNotificationException {
+        if (!getProps().getEvent().equals("message")) {
+            throw new InvalidNotificationException("Cannot handle this notification.");
+        }
+        super.onReceived();
     }
 }
--- a/android/app/src/main/java/com/zulipmobile/notifications/PushNotificationsProp.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/PushNotificationsProp.java
@@ -71,4 +71,8 @@ public class PushNotificationsProp extends PushNotificationProps {
     public int getZulipMessageId() {
         return Integer.parseInt(mBundle.getString("zulip_message_id"));
     }
+
+    public Object getEvent() {
+        return mBundle.getString("event");
+    }
 }


android/app/src/main/java/com/zulipmobile/MainApplication.java
```diff
@@ -85,11 +85,13 @@ public IPushNotification getPushNotification(Context context, Bundle bundle, App
...
         } else {
             PushNotificationsProp prop = new PushNotificationsProp(bundle);
-            addConversationToMap(prop, conversations);
-            return new GCMPushNotifications(context, bundle, defaultFacade, defaultAppLaunchHelper, new JsIOHelper(), conversations);
+            if (prop.getEvent().equals("message")) {
+                addConversationToMap(prop, conversations);
+                return new GCMPushNotifications(context, bundle, defaultFacade, defaultAppLaunchHelper, new JsIOHelper(), conversations);
+            }
         }
+        return null;
```
gnprice: The main question I have on looking at this code is:

Hmm. Why is this returning null? What's the semantics of `getPushNotification` returning null?

I ended up with a lot to say and decided to make it visible for the whole team, so I went into detail [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/android.20notifications/near/626529). Please go ahead and read that discussion if you haven't already.

For the record here, the bottom line for this PR:
* Returning null where we currently do is OK.
* Returning null in the GCM case isn't right. But! We can instead throw an `IPushNotification.InvalidNotificationException`, and that will have the right effect. Let's do that for the immediate fix here.

kunall17: Throwing the exception is not straightforward here, as the `get` method in `PushNotification.java` is not handling the Exception, hence I'll see for some workarounds so we don't have to return null.

gnprice: I see, yeah.

Checked exceptions -- [not one of Java's better decisions](https://www.javaworld.com/article/3142626/core-java/are-checked-exceptions-good-or-bad.html).

I think the right way to do this in the API of this library is to have the `onReceive` on our `IPushNotification` implementation take its own look at things and throw that exception instead.

... Ha, wait, actually, instead of throwing the exception, that method can just... not do anything. This exception type is actually completely meaningless. The one method that has the option to throw it, can just return instead and get the same effect.

So yeah. I think we can just write
```
@Override
public void onReceived() {
  if (/* ... */) {
    return;
  }
  super.onReceived();
}
```
and that'll do the job.

kunall17: Another way can be, I'll implement my own GcmMessageHandlerService and handle the null returns?
I'll have to patch the AndroidManifest though.

kunall17: This AndroidManifest thing is not possible as it doesn't allow changing the attr `android:name` and that is the unique key as well.

kunall17:
@gnprice updated the PR, but haven't tested it. Have to rush to college.

android/app/src/main/java/com/zulipmobile/MainApplication.java
```diff
@@ -85,11 +85,12 @@ public IPushNotification getPushNotification(Context context, Bundle bundle, App
             clearConversations(conversations);
             NotificationManager nMgr = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
             nMgr.cancelAll();
-            return null;
```
gnprice: Hmm -- this adds a new wrinkle. Why change the behavior for the case of these dismiss intents?

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -90,6 +90,10 @@ public void onOpened() {
         // First, get a builder initialized with defaults from the core class.
         final Notification.Builder builder = super.getNotificationBuilder(intent);
 
+        if (!getProps().getEvent().equals("message")) {
+            return builder;
```
gnprice: I'm not sure what this is doing. Is this to handle something that starts happening in the dismiss-event case, because we no longer return `null` there?

kunall17: Basically to save some processing of the events as anyways the notification will be dumped afterward in the onReceived call.
So this avoids the modification of the builder, and gets dumped later on.

This is the stacktrace when a remove notification is recieved on the app now.
```
08-09 17:57:28.136 5819-6633/com.zulipmobile.debug V/ReactNativeNotifs: GCM message handling aborted
    com.wix.reactnativenotifications.core.notification.IPushNotification$InvalidNotificationException: Cannot handle this notification.
        at com.zulipmobile.notifications.GCMPushNotifications.onReceived(GCMPushNotifications.java:208)
        at com.wix.reactnativenotifications.gcm.GcmMessageHandlerService.onMessageReceived(GcmMessageHandlerService.java:20)
        at com.google.android.gms.gcm.GcmListenerService.handleIntent(Unknown Source:202)
        at com.google.firebase.iid.zzg.run(Unknown Source:26)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
```

android/app/src/main/java/com/zulipmobile/MainApplication.java
kunall17: To remove all the null return calls. After removing the return call we recieve a stacktrace.

```
08-09 18:03:08.885 3661-3661/? D/LogHelper: MXMNotificationListener - removed notification: com.zulipmobile.debug
```

I read about that stacktrace, and got to know its because the notification doesn't contain a data key which is required by each GCM message, whereas this is a fake one generated by the ProxyService, Very less functions are called for completing the function.
Although return null can also be used.

gnprice:
Closing this PR. Thanks for your work on it! I just pushed 8e9f1a6c5 and its ancestors to fix this issue instead.

Do take a look at those -- particularly that one and its parent, 96c61eb0a -- because they have some details with my analysis of how this wix/react-native-notifications library behaves.

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
gnprice: I've now studied the surrounding code quite a bit in the course of writing my own fix for this issue -- see linked commits in my main-thread comment.

Can you explain more about when you believe this `if` will be true and this bit of code will have an effect?

As far as I can see what's happening in the code, this `if` will never be true.
* The one place `getNotificationBuilder` is called is inside the base class's `buildNotification`; which is called only in its `postNotification`; which is called only in its `onPostRequest` (for local notifications, which we never do) and `onReceived`.
* With this commit's change to our own `onReceived`, the base class's `onReceived` is only called if this same condition is false.
* So this code is never reached if this condition is true (i.e. if `event` is not `message`.)

android/app/src/main/java/com/zulipmobile/MainApplication.java
gnprice: That line you quoted is not a stack trace. :-P A stack trace has a list with a function's caller, and its caller, and so on. Was there a stack trace after it? The actual stack trace would be good to paste if discussing it; I can't quite tell what you're saying.

The case where we enter this `if` body is with a "notification" that is a fake one generated not by the ProxyService, but by our own code. It's the bundle we add to `dismissIntent` in our `getNotificationBuilder`. ProxyService is just the place where we have chosen, in that same block of our own code, to point that intent at. It receives it, pulls out that bundle, and invokes this function.

kunall17: Haha yeah it's not a stack trace, I meant a debug log message.

kunall17:
Thanks for the fix @gnprice

PR #2875, opened 2018-08-06
fix: Improve focus ping

borisyankov:
Pass `new_user_input` as `true` instead of `false`.
I had incorrectly assumed before that this has something to do
with the user putting in new 'input' (typing)

The ditermination of wether it should be `true` or `false` is made
on the web if the user has actually interacted with the app.

On mobile we do not have a long running browser tab that might be
in the background, thus we should pass `true` always.

As discussed here:
https://chat.zulip.org/#narrow/stream/3-backend/topic/presence.20update.20fields

### DIFF
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -11,7 +11,7 @@ import { isPrivateOrGroupNarrow } from '../utils/narrow';
 let lastReportPresence = new Date();
 let lastTypingStart = new Date();
 
-export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean = false) => async (
+export const reportPresence = (hasFocus: boolean = true) => async (
   dispatch: Dispatch,
   getState: GetState,
 ) => {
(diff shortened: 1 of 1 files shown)


HarshitOnGitHub:
@borisyankov @gnprice Ping on this. I have been facing this annoying issue for quite sometime now and I am sure I am not alone.

gnprice:
Thanks @borisyankov !

The most important information I'd like to see here is: what is `newUserInput` (or `new_user_input`) supposed to mean? There's a line about what we'd previously incorrectly believed it was supposed to mean; that's helpful information, but even more important is what it *is* supposed to mean. That's what ought to determine what we choose to do here, after all. :slightly_smiling_face: 

Following the chat link, there's this from Tim:
> we should set it whenever we know the user has interacted with the app recently.

so that seems like an answer. That should go in the commit message.

Then, my next question: does this new version actually match that spec? We send these on an interval after `doInitialFetch`, and we never cancel the interval:
```
export const doInitialFetch = /* ... */
  setInterval(() => sendFocusPing(), 60 * 1000);
```

What happens when Zulip is on the screen and the user puts their phone down? It seems like we could potentially be sending these for a while.

gnprice:
@HarshitOnGitHub Can you elaborate on the issue you're seeing? (Perhaps file an issue for it?)

Based on the linked chat thread, the thing we're talking about is barely used -- it's not clear that it's used in the main product experience at all. That makes me think the issue that's annoying you is likely to be something else. I'd like to better understand the issue that's annoying you, so that we can make sure we fix it.

Issue #2876, opened 2018-08-06
Don't break `/thumbnail` images in message list


borisyankov:
Our app does not show images in WebView using the new thumbnailing coming in 1.9
We should fix that ASAP.

gnprice:
Ah, this was actually implemented by #2880 back in 2018 -- but it meant something rather narrower than the original title. Closing.

PR #2880, opened 2018-08-08
Add support for new thumbnail format

borisyankov:
Implements #2876

Our current code works fine with the images in messages used before.
It doesn't work with the new, upcoming thumbnails. The reason is that our `appendAuthToImages` function is too rigid and does not recognize the new `img` tags (which have some extra `data` attributes and the URL is a bit different)

Instead of trying to parse arbitrary HTML or trying to limit what HTML the server renders, we can do the smart thing. This PR moves the functionality inside the WebView. It has quite an unfair advantage - a full-blown browser that perfectly recognizes HTML.

### DIFF
--- a/src/utils/__tests__/url-test.js
+++ b/src/utils/__tests__/url-test.js
@@ -16,7 +16,6 @@ import {
   hasProtocol,
   fixRealmUrl,
   autocompleteUrl,
-  appendAuthToImages,
 } from '../url';
 
 import { streamNarrow, topicNarrow } from '../narrow';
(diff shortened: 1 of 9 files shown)


gnprice:
Thanks @borisyankov ! I definitely agree that this is a cleaner solution than the old regex-based solution. :-)

One comment on the behavior of the final code: it looks like this will try appending the API key to the source URL of any `img` element on the server's domain. I think that broadens it beyond uploads and thumbnails to some others where it doesn't apply: among them avatars and realm emoji, from a quick grep of our code so not including any that might appear in the server-rendered message content.

I guess those two examples are actually things we probably *should* be authenticating, so the right thing there is probably to go ahead and add the key (which current servers will ignore) and then make the server start requiring it in the future.

Still, "any `img` element" is potentially pretty broad, and I don't love the idea of sending random unexpected parameters to arbitrary routes on the server -- particularly when the parameter is sensitive data like this one, though certainly the fact that it's limited to the Zulip server's domain mitigates the risks a lot. So I'd like it to continue to be a whitelist of routes -- we can just add `/thumbnail` to that whitelist, and also `/user_avatars`. (Hilariously the latter appears to be also what we use for realm emoji.)

And one comment on the structure of the branch: as always :wink: , I want each commit to leave things in a non-broken state. Partway through this branch, it looks like we have a state where the `api_key` gets appended twice -- it's possible everything behaves fine from a user perspective, but that's definitely the kind of brokenness that makes the changes harder to understand because things don't seem to match up.

Given the hour, I'll see about adjusting both of those myself and then merging.

gnprice:
Merged, with those changes. Thanks again!

PR #2881, opened 2018-08-10
ios: Remove support for iOS 8, make 9 the minimum

borisyankov:
Updates the minimal iOS platform version supported up from
8.2 to 9.

This makes it official and the app will not be available for
download on iOS 8 devices. Our app currently crashes/exits on
startup without any reported error to Sentry or iTunes Connect.
This was verified both on a physical device and BrowserStack's
App Live.

We are not limiting our device support as all devices that run
iOS 8 can be upgraded to 9.

More information about devices and platform support:
http://iossupportmatrix.com/
https://everyi.com/by-capability/maximum-supported-ios-version-for-ipod-iphone-ipad.html


gnprice:
Looks good, merged -- thanks!

PR #2883, opened 2018-08-12
docs: Add instructions for easy unsigned android release builds to release guide.

Shayan-To:


gnprice:
Thanks @Shayan-To ! This will be quite useful.

I'd rather avoid instructions that involve editing code and editing it back -- that's kind of a fragile process and also adds hassle for the person using the instructions. So let's try to make this a version that's just another command you can run (or an option to a command would be cool too.)

Fortunately I think that should actually be pretty easy. Take a look at this page:
https://developer.android.com/studio/build/build-variants
and perhaps also this one it refers to for background: https://developer.android.com/studio/build/
In particular, we have build types `debug` and `release`. Let's add a third one, perhaps `releaseUnsigned`, that does this.

Shayan-To:
@borisyankov @gnprice I added a small property-switch to the gradle build script. Updated the instructions as well.

gnprice:
Huh interesting!

What's the motivation for using a project property instead of a build type?

I'm curious how the output filename is getting set to `app-release-unsigned.apk`. I like that the name is different -- but I don't see where the magic is.

Shayan-To:
@gnprice I imagine that the sentry thing is not something that is easy to control from within a build type. I took a look at the code that was being `apply`ed, looked like it is doing lots of strange things that I don't want to know how I can undo... So the only way I could imagine was to prevent it from being `apply`ed. And the only way I knew to do that was using a property.

About the different names, it is part of the build process. The build process first creates the unsigned file and names it `app-release-unsigned.apk`. Then if it is to sign it, it will create a signed version of that APK file and name it `app-release.apk`, and maybe removes the unsigned file. The file name difference comes from the fact that building and signing processes are completely separate. It is happening naturally, I didn't do anything about it.

gnprice:
Got it, thanks! That makes sense. Seems awkward that Sentry doesn't provide much of an off switch... but apparently it doesn't.

This sort of question ("why'd we do it that way instead of this other natural-seeming way?") is the kind of thing I like to make sure the commit message answers. Basically, because I was puzzled by it before asking the question and seeing the answer, I'd like to have the answer there up front so future readers don't have to be puzzled. :-)

That and a few other small things, and then I'll be glad to merge this!
* Specifically, a few words in the commit message like "We use a project property, rather than an alternate build type, because there doesn't seem to be a good way to turn off Sentry except at the root of the project config."
* Also a one-line comment inside that conditional, like "Disable Sentry because it requires an auth key.", since this part isn't really about signing anything so the connection requires a bit of explanation.
* Let's call the property "unsigned" -- that feels to me easier to parse than "noSigning".
* Does the `signingConfig null` do anything? Or could we equivalently just say `if (!...) { signingConfig ... }`?
* This is good information in the docs, but this is a confusing place for it -- the story around this is all about "how to make a release of Zulip Mobile to the iOS App Store, to the Google Play Store, and as an APK on the web", and this is something we *wouldn't* do for actually making a release.
   * How about in docs/howto/build-run.md , under "Android tips", perhaps a second command under `react-native run-android`?
   * Could also add a quick pointer at this spot in the release-guide doc in case someone finds themself there when they're looking for how to do this, but just a short sentence in parens like "(If you want to do a 'release build' but without the signing key, try `yarn build:android-unsigned`.)", with some of that as a link to the longer explanation.

gnprice:
Thanks again @Shayan-To for your work on this!

I found myself wanting this functionality recently, so because this PR has been dormant I went ahead and made the changes I described in my last comment above. Now merged.

I also added a few more changes on top -- I adjusted the logic a little further to make "unsigned" more like the default, and also switched the build in CI to work this way. See 6a36f8423 and ancestors.

PR #2885, opened 2018-08-13
Support new stream format for internal links

### LINKED ISSUE #2760: Internal links to streams and narrows don't work
Reported in chat [here](https://chat.zulip.org/#narrow/stream/48-mobile/subject/Hyperlinks.20within.20stream.20view.20don.E2.80.99t.20work/near/608487) and previously [here](https://chat.zulip.org/#narrow/stream/48-mobile/subject/Navigating.20Streams/near/559780).

We actually have logic that attempts to do this, but it's too narrow in the forms of URLs it handles, and in particular it doesn't handle the form the server/webapp migrated to some time ago where IDs are included.

#2312 is a previous version of this bug. #2313 was a previous attempted PR to address it. See comments on #2313 for some details on how to address the issue.

agzuniverse:
The first five commits reworks `isGroupLink`  and similar functions to avoid repetition of `isUrlInAppLink` and `getPathsFromUrl`.

The sixth commit introduces a function called `getStreamNameFromSlug` that removes the stream ID in the case of the new format and returns the stream name, or returns the stream name in the case of the old format.

This covers the case where, let's say a user has a stream named `test` with ID 7 and a stream named `7-test` with ID 8.
If the links are given in the new format, `7-test` will navigate to stream named `test` with ID 7 and `8-7-test` will navigate to stream `7-test` with ID 8, as expected.
If the link to stream `7-test` is given in the old format, it will navigate to stream `test` with ID 7 as the new format is prioritised.

A drawback I see with my own approach here is that the stream matching is still done by stream name and not stream ID. Matching by ID can be introduced by some changes in `isNarrowValid` or in a similar place.

TODO: Add tests for `getStreamNameFromSlug`.

Fixes #2760

### DIFF
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -54,15 +54,8 @@ export const isGroupLink = (paths: string[]): boolean =>
 export const isStreamLink = (paths: string[]): boolean =>
   paths.length === 2 && paths[0] === 'stream';
 
-export const isSpecialLink = (url: string, realm: string): boolean => {
-  const paths = getPathsFromUrl(url, realm);
-  return (
-    isUrlInAppLink(url, realm)
-    && paths.length === 2
-    && paths[0] === 'is'
-    && /^(private|starred|mentioned)/i.test(paths[1])
-  );
-};
+export const isSpecialLink = (paths: string[]): boolean =>
+  paths.length === 2 && paths[0] === 'is' && /^(private|starred|mentioned)/i.test(paths[1]);
 
 export const isEmojiUrl = (url: string, realm: string): boolean =>
   isUrlOnRealm(url, realm) && url.includes('/static/generated/emoji/images/emoji/unicode/');
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -52,14 +52,9 @@ export const isTopicLink = (url: string, realm: string): boolean => {
   );
 };
 
-export const isGroupLink = (url: string, realm: string): boolean => {
-  const paths = getPathsFromUrl(url, realm);
-  return (
-    isUrlInAppLink(url, realm)
-    && ((paths.length === 2 && paths[0] === 'pm-with')
-      || (paths.length === 4 && paths[0] === 'pm-with' && paths[2] === 'near'))
-  );
-};
+export const isGroupLink = (paths: string[]): boolean =>
+  (paths.length === 2 && paths[0] === 'pm-with')
+  || (paths.length === 4 && paths[0] === 'pm-with' && paths[2] === 'near');
 
 export const isStreamLink = (url: string, realm: string): boolean => {
   const paths = getPathsFromUrl(url, realm);
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -83,8 +83,10 @@ export const getEmojiUrl = (unicode: string): string =>
   `/static/generated/emoji/images/emoji/unicode/${unicode}.png`;
 
 export const getNarrowFromLink = (url: string, realm: string, users: User[]): Narrow => {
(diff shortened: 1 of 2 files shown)


src/utils/url.js
```diff
@@ -83,22 +83,23 @@ export const getEmojiUrl = (unicode: string): string =>
...
-  } else if (isTopicLink(url, realm)) {
-    return topicNarrow(
-      decodeURIComponent(transformToEncodedURI(paths[1])),
-      decodeURIComponent(transformToEncodedURI(paths[3])),
-    );
-  } else if (isStreamLink(url, realm)) {
-    return streamNarrow(decodeURIComponent(transformToEncodedURI(paths[1])));
-  } else if (isSpecialLink(url, realm)) {
-    return specialNarrow(paths[1]);
+  if (isUrlInAppLink(url, realm)) {
```
borisyankov: What happens if `isUrlInAppLink` returns `false`?
We will not return anything, and that looks like a bug.

You should prefer an 'early return`. 
So something like:
```js
if (!isUrlInAppLink(url, realm)) {
  return ...;
}
```

That makes the code simpler and the diff smaller.

src/utils/url.js
```diff
@@ -52,14 +52,9 @@ export const isTopicLink = (url: string, realm: string): boolean => {
...
 
-export const isGroupLink = (url: string, realm: string): boolean => {
-  const paths = getPathsFromUrl(url, realm);
-  return (
-    isUrlInAppLink(url, realm)
-    && ((paths.length === 2 && paths[0] === 'pm-with')
-      || (paths.length === 4 && paths[0] === 'pm-with' && paths[2] === 'near'))
-  );
-};
+export const isGroupLink = (paths: string[]): boolean =>
```
borisyankov: Hmm, that makes the function very confusing. 

Sure you've saved the `getPathsFromUrl` and `isUrlInAppLink` but the function makes no sense as I would not expect the question `Is this param a link to a group` be asked on a string array.

agzuniverse: I could inline the function, but perhaps just a name change, to `hasGroupLink` or something like that be better?

src/utils/url.js
agzuniverse: It returns `HOME_NARROW` if `isUrlInAppLink` is false.

Changed this to have a early return as you suggested.

gnprice:
I like the direction that this refactor is pointing in, with simplifications like this:
```
-export const isGroupLink = (url: string, realm: string): boolean => {
-  const paths = getPathsFromUrl(url, realm);
-  return (
-    isUrlInAppLink(url, realm)
-    && ((paths.length === 2 && paths[0] === 'pm-with')
-      || (paths.length === 4 && paths[0] === 'pm-with' && paths[2] === 'near'))
-  );
-};
+export const isGroupLink = (paths: string[]): boolean =>
+  (paths.length === 2 && paths[0] === 'pm-with')
+  || (paths.length === 4 && paths[0] === 'pm-with' && paths[2] === 'near');
```

However, this branch doesn't actually work! We're rearranging the interface of the helpers like `isGroupLink`, without changing their caller to match -- so it can't possibly work, and in fact both the tests and the type-checker show that it doesn't. I do put a high value on making the code easy to understand, but even more important than that is that it actually work :smile: . (And in this case, the same problems that make it not work also make it confusing anyway.)

This also doesn't address #2760; it adds a function `getStreamNameFromSlug`, but doesn't start using it anywhere.

The logic in that `getStreamNameFromSlug` is wrong; when there's a number, it just ignores it.

Let's close this PR, in favor of making a fresh go at #2760 .

Issue #2434, opened 2018-04-30
Chinese and Japanese input don't work on iOS (blocked on RN v0.57)


gnprice:
The normal input methods for users to write in Chinese or in Japanese don't work in our app, on iOS.

You don't need to know any Chinese or Japanese to work on this issue; I've described how to reproduce it below. Ask me here or in chat if something is confusing.

Steps to reproduce:
* In iOS settings, go to "General -> Keyboard -> Keyboards -> Add New Keyboard...". Select "Chinese (Simplified) -> Pinyin -- 10 Key".
* Go to Messages, or some other app. Hit the "globe" key on your keyboard, until you get the keyboard you just added -- it has keys in the middle for "ABC", "DEF", etc., and some keys labeled in Chinese. (You can always hit the "globe" key to get back to your normal keyboard.)
* Hit the keys for M G G A M, in that order. The input box should say "ni hao", which should be selected; and you should have suggestions including "你好".
  * This is "hello" in Chinese. The keyboard has guessed that out of all the letter sequences you could have meant with MNO GHI GHI ABC MNO, the most likely is "ni hao"; or in Chinese characters, "你好".
  * This is how this keyboard is supposed to work.
* Switch to the Zulip app. Narrow to a topic somewhere, and tap the "Reply" field to start composing a message.
* Hit the same five keys. The input box just says "mggam". For me, the suggestions read "is and of said has".
  * It's impossible to type Chinese this way.

The keyboard "Japanese -- Kana" is similarly broken, but the repro is harder to describe starting from no knowledge of Japanese. I'll consider this issue fixed when the Chinese keyboard is working :) -- that'll probably fix the Japanese keyboard too, and I'll file a separate issue if it doesn't.

borisyankov:
This is related:
https://github.com/facebook/react-native/issues/18403

hjiang:
I ran into this problem with the latest iOS app too. This is a video that shows the problem:
http://dropshare.47590191.mtmss.com/1Uh9o3wBA0.mov

All non-latin languages are probably broken.

hjiang:
Seems the upstream issue had been closed. Is there plan to making a new release soon? The current version is unusable for CJK users.

borisyankov:
The fix for the issue is in this commit:
https://github.com/facebook/react-native/commit/892212bad2daadd373f4be241e4cd9889b0a1005

It has not yet landed in a released version of React Native. 
[...]

PR #2789, opened 2018-07-12
Upgrade to React Native 0.56 & 0.57

borisyankov:
Upgrade dependencies, config and build files to RN 0.56

Previous upgrades were done manually by updating dependencies and
.flowconfig

This commit contains a full `diff` with previous versions.

What was done is:
 * run react-native-git-upgrade
 * resolve merge conflicts
 * update dependencies that were needed but missed by the script

Note: 
* In another commit I'll need to fix Flow issues before merging
* Jest test runner need fixing: https://github.com/facebook/react-native/issues/19859

PR #2886, opened 2018-08-13
Turn iOS compose box uncontrolled; dedupe with Android

borisyankov:
A fix for the iOS blocking issue is nowhere in sight, but a cool and simple workaround was discovered that allows us to implement the fix now, and not in an unspecified time in future.

Details are given in the commit messages.


gnprice:
Thanks! Glad to get these together again :)

A few questions:
* It's the first commit that fixes #2434, not the second, right?
* In the second commit, in the iOS case, when `text` is empty it looks like we end up calling `setTimeout(...setNativeProps(...))` with an empty `text` twice. Perhaps in `updateTextInput` it should just be like this?
```
if (text.length === 0) {
  // ...
} else {
  textInput.setNativeProps({ text });
}
```
* In these comments that mention different RN bugs, let's definitely have links -- either to the upstream bugs, or to our own issue that describes the RN bug when that's a better reference.
* Also, this isn't an "iOS bug" as far as I know -- it's an RN bug specific to iOS. Similarly "Android" doesn't have a bug here. You know that, but I think it's important to be clear about it in the code for other readers -- not only to keep an accurate mental tally of which platforms are responsible for bugs that affect us :stuck_out_tongue: , but also because that framing can make a big difference when following the link and reading more details to understand what the story is.
* How confident are you / how are you confident that the double-set trick consistently avoids the issue? The `setTimeout` in particular makes me wonder if there might be a race where sometimes the relevant bit of RN code runs in between the two sets, and other times it doesn't and the trick doesn't work -- like maybe that bit of RN code runs once per frame, and if we get an update just after that code ran for one frame, and the phone is nice and fast and not busy, the second set happens within the >=8ms before it runs for the next frame.
* On the other side of things, I wonder if a legitimate update could get squeezed in between the two sets, and get clobbered -- like if the user has just one letter there, and hits delete and then a letter in quick succession, and maybe if it's an old and overloaded phone so the JS engine takes a while to get around to the `setTimeout`'d second set.

borisyankov:
Indeed it is the first commit that solves #2434, I will reword that.

Hmm, while the code I did commit works correctly, I seem to have missed to change the `updateTextInput` function completely as intended. I totally wanted to do 

```js
  if (text.length === 0) {
    clearTextInput(textInput);
  } else {
    textInput.setNativeProps({ text });
}
```

borisyankov:
The `setTimeout` concern is pretty good to raise! I am researching to see how likely it is to cause this issue.

gnprice:
Quick update:
* @borisyankov [mentioned in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/RN.200.2E56.20.26.200.2E57/near/642976) that he's been able to reproduce cases where the `setTimeout` did in fact mess up the text input. So, chalk up another candidate workaround for upstream's bug facebook/react-native#18272 that doesn't quite make it. (There's a long line of them in that thread.)
* On the other hand, RN v0.57 has been released, and it has the fix for that bug! So once we upgrade to v0.57, we won't need such workarounds and can just straightforwardly switch to uncontrolled inputs like we have on Android.
* See #2789 for the current work on upgrading to RN v0.56; after that, we'll tackle v0.57.

gnprice:
I think this makes most sense to track as an issue at this point rather than a PR; when we come back to it I think we'll basically repeat the boring part of this PR and (happily!) skip the more interesting part. So I went and filed #3017 for that.

PR #2890, opened 2018-08-15
Do not include empty values in API calls

### LINKED ISSUE #2889: Subscribe / ubsubscribe self from streams does not work
The latest published app (and `master`) is not able to subscribe/unsubscribe the user from streams.

borisyankov:
Introduces a new function `encodeAsUriNoEmptyValues` and replaces the `encodeAsURI` used previously.

Now we do not pass empty values to the back-end API.

Also, fixes #2889

### DIFF
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -54,6 +54,8 @@ export const apiCall = async (
     if (response.ok && json !== undefined) {
       return resFunc(json);
     }
+    // eslint-disable-next-line no-console
+    console.error({ route, params, response });
     throw makeApiError(response.status, json);
   } finally {
     networkActivityStop(isSilent);
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -1,6 +1,6 @@
 /* @flow */
 import type { Auth, ResponseExtractionFunc } from '../types';
-import { getAuthHeader, encodeAsURI, isValidUrl } from '../utils/url';
+import { getAuthHeader, encodeAsUriNoEmptyValues, isValidUrl } from '../utils/url';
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks for this fix!

The exact behavior here is kind of confusing:
```
export const removeEmptyValues = (obj: Object): Object => {
  Object.keys(obj).forEach(key => obj[key] == null && delete obj[key]);
  return obj;
};
```

I think this removes values that are either `null` or `undefined`, but no others. Is that your analysis as well?

Neither of those is necessarily what I'd think of from seeing "empty", and on the other hand `{}` and `[]` and `''` are definitely "empty" in many contexts. So I'd want a comment explaining this function, if not a better name or other refactor.

Then where it's newly used in this branch, it seems like what you really want is to eliminate `undefined`, not `null`. We might actually pass a `null` through here sometimes, and if so then taking it out is probably a bug -- `null` is a perfectly good JSON-serializable value, which Python deserializes back as `None`. It definitely appears in the API at least in some return values, and there may be routes that give `None` a particular meaning for some parameter but will give an error if the parameter is simply not there.

I *think*, from a quick experiment with better Flow types, that we never pass `null` values here. (As far as Flow can tell, we never pass `undefined` either, but that's because `JSON.stringify` is annotated with a white lie that it returns `string`. In fact `JSON.stringify(undefined)` is `undefined`, which is how an `undefined` sneaks in here, but `JSON.stringify(null)` is `"null"`.) So I'll merge this; but it's pretty brittle, so I'll clean it up in some further commits.

gnprice:
Done -- see 94234c324 and ancestors for those followups. Thanks again!

Issue #1507, opened 2017-12-06
Warn when server isn't set up for push notifications


gnprice:
Many people installing a mobile app for a messaging service will assume that it'll get them push notifications -- they may see that as the main reason to install the mobile app.

Zulip does, of course, but only if the Zulip server they use has set up to [send push notifications through the bouncer](https://zulip.readthedocs.io/en/latest/production/mobile-push-notifications.html). And many servers out there have not. So if a user who hasn't read all about the push notification bouncer (who maybe isn't a server admin) tries installing the mobile app, they are likely to be disappointed.

We should detect whether the server supports push notifications. Not sure if there's an existing good way to detect this from the client side, or if we need to have the server export another bit of information. Then we should display a warning, at a minimum when the user first adds the server.

I think it would also be helpful to go further and have it in some ongoing fashion, if we can find the right place to put that in the UI.

---

Added later: When we introduce this sort of warning, we should also have it apply when the server nominally supports push notifications but is so old that it won't know how to send them in a format that we accept. #4967 will make it the case that this requires at least Zulip Server 1.9, and that minimum may increase in the future.

borisyankov:
I am also not aware if there is a way to detect notification support.
Getting the information via `server_settings` makes a lot of sense.

timabbott:
It would be easy to add the data to `server_settings`; just check whether `settings.PUSH_NOTIFICATION_BOUNCER_URL` is set.  Though the other approach to consider would be to have the response to the "register for push notifications" request sent to the local Zulip server return the state; I think that would be better, since it would correctly handle cases where the app was connected to a server and then push notification bouncing was configured in the server (I think in this case, the app currently requires you to logout and login again to get connected).

kunall17:
Let's do this in steps, 

1. Store the result of `users/me/apns_device_token|android_gcm_reg_id` (API for registering push from app) if successful let's show the result in the notification settings page that the device is registered for the notifications 

If not successful

1. Determine if server is configured for the push bouncer 
1. If the server is configured then show a retry button otherwise show server doesn't support notifications? 

@timabbott do we want to expose the push bouncer url and send the url of bouncer or just a boolean if its set in the API of `server_settings` for which the code changes should be here https://github.com/zulip/zulip/blob/73a5c392640c5c1f96f9c94f5f271d8e4d9251ad/zerver/context_processors.py#L99

kunall17:
[...]

PR #2893, opened 2018-08-16
Store and use server settings

borisyankov:
Crates a `serverSettingsReducer` and when we receive that data updates it.
Adds calls to update the `serverSettings`
Also adds selectors to get this data out of the state.

Used to display realm icons, notification status etc.


gnprice:
(#1507 is the issue for warning when the server isn't set up for notifications.)

borisyankov:
Rebased.

gnprice:
Thanks @borisyankov !

One thing I noticed just in doing a quick rebase of this:

> Extract a generic Notice component out of OffflineNotice
> refactor: Rework OfflineNotice to use Notice internally

This is a case where things would actually be clearer as a single commit. One commit copies a bunch of code, then the next one deletes (the bulk of) the original -- which means that what's really happening is the code is getting moved, and if they're squashed as one commit that becomes explicit.

In particular that means that when reading the green side of the diff, the code that's being copied is right there in the red side of the same commit, so the context is self-contained. It also means there's no worry when looking at the first commit that we're duplicating a bunch of code, because it's explicit that it's being moved.

gnprice:
> +export type ServerSettingsState = {

Let's be sure to put a jsdoc on this. Follow the style of the other `*State` types' jsdoc.

One important question to answer: what are the keys exactly? There are several different formats we could potentially use for a string to identify a realm.

> +const initialState: ServerSettingsState = NULL_OBJECT;

There's no optimization benefit to making this be the same `{}` object as other `{}` we have lying around -- this `const` initializer runs once -- so let's write this more simply and directly as `{}`.

```
... => ({
+  ...state,
+  [action.data.realm_uri]: {
+    ...state[action.data.realm_uri],
+    authentication_methods: action.data.realm_authentication_methods,
+    realm_description: action.data.realm_description,
+    realm_icon: action.data.realm_icon_url,
+    realm_name: action.data.realm_name,
+    realm_uri: action.data.realm_uri,
+  },
+});
+
```

I think this would benefit from destructuring and object-initializer shorthand. Like
```
const { realm_uri, authentication_methods, /* ... */ } = action.data;
return { ...state,
  [realm_uri]: {
    ...state[realm_uri],
    authentication_methods,
    /* ... */ }};
```

... Oh hey, in fact after typing that I notice that `authentication_methods` is an *exception* to the names matching up. That makes me further inclined to prefer the shorthand -- the exceptions will stand out because the boring stuff will be simpler. :-)

> export type GlobalState ...
> +  serverSettings: ServerSettingsState,

Hmm -- this should have come when adding this reducer to `reducers`, no?

```
+export const getCurrentServerSettings = createSelector(
+  getCurrentRealm,
+  getServerSettings,
+  (realm, serverSettings) => serverSettings[realm] || NULL_OBJECT,
+);
```

Hmm -- I think it isn't 100% obvious expected behavior that the caller of this selector might get an empty object.

And in fact as drafted, this API is quite dangerous -- the type inferred for the selector is that it returns an `ApiServerSettings`, which means that if you just take `serverSettings.push_notifications_enabled`, that is already (supposedly) a `boolean`. It'd be very reasonable for someone to use that directly and not compare against `false` -- and then the code will implicitly treat "don't know" as false.

So this needs to get a type that causes `serverSettings.push_notifications_enabled` to have a type like `?boolean` or `boolean | void` (or else to just be an error without checking `serverSettings` first.)

It'd also be good to mention in jsdoc that it returns `{}` when not found, if that is what it ends up doing.

gnprice:
This PR has been stale for a while. Given the last comments above, plus all the changes that have happened in master, I think the shortest route to fixing the issue will be to approach it fresh; so, closing this PR.

Issue #2863, opened 2018-08-04
Give the Android debug build a different icon for notifications, too


gnprice:
Following #2743 and #2745 , I am quite happy to have a debug build of the app installed on my phone alongside the release build. (It doesn't really work most of the time, as it needs the phone to be connected by e.g. USB to my computer where I'm running Metro with `react-native start`... but it's very handy for testing some things.)

One small annoyance is that when I get a notification, I get two copies -- one for each build. And I can't tell them apart in the UI, so I don't know which one will take me properly to the release build of the app, and which will take me to the debug build which typically doesn't work.

We solved this problem for the launcher / home screen by providing a modified icon in debug builds. Let's do something similar for the notification icon -- a similar "DEBUG" overlay would do nicely.

PR #2897, opened 2018-08-17
android notification: Use different icon for debug build.

jainkuniya:
Before both debug and release build were having same icon and both
build can coexist at the same time in the phone. So it was hard to
figure out from which build notification is generated.

For now reusing the laucher icon of debug build as notification icon
of debug build. As this icon is not as per notification icon pattern,
it looks quite odd. But it saves extra overhead of new icon. Only
contributors will be having debug build, so this can be compromissed.

Also next we should try not to include debug launcher icon in release
APK.

Fix: #2863


gnprice:
Thanks @jainkuniya ! Merged.

Copying here for the record some screenshots from a chat thread on the stream:

a notification, looking odd:
image

but in the status bar, looking fine:
image

PR #2899, opened 2018-08-17
fix: Do not open YouTube and Vimeo links in Lightbox

### LINKED ISSUE #2895: Embedded YouTube Videos Do Not Display When Clicked
Embedded YouTube videos do not appear as expected when users click on the video preview in the Desktop and Android/iOS apps (i.e., users just get a black screen).  This occurs in version 2.3.3 of the Desktop app and version 15.0.92 of the mobile app.  This issue does not occur when logged in via web browser.

borisyankov:
Fixes #2895

We do not support video in the Lighbox screen, nor do we intend to
(no benefit). Instead treat the links as normal ones if they are
to a video on YouTube or Vimeo. This matches the behavior of many
other mobile apps.

In future we might need to handle other types of links but analysis of
https://github.com/zulip/zulip/blob/master/static/js/lightbox.js
shows we recognize only YouTube and Vimeo yet.


borisyankov:
That is potentially a different feature that we might want to implement, but is not related to the issue :)

YouTube and Vimeo videos are just links to the page on YouTube/Vimeo, not a link to an actual file.

gnprice:
This seems pretty fragile. Can you elaborate on how you determined that these two patterns are the only exceptions that we currently need to handle?

Here's the diff:
```
-  if (e.target.matches('a[target="_blank"] > img')) {
+  if (
+    e.target.matches('a[target="_blank"] > img')
+    && !e.target.matches('.youtube-video > a[target="_blank"] > img')
+    && !e.target.matches('.vimeo-video > a[target="_blank"] > img')
+  ) {
```
so effectively we're saying that an image inside a blank-target link should pull up the lightbox... unless the link is in turn inside either a `.youtube-video` or a `.vimeo-video` element. Are there no other cases where we have a similar image-inside-link?

OK, and I grepped the server repo for `vimeo-video`, to see what generates this HTML. It's in `zerver/lib/bugdown/__init__.py`; it's unfortunately kind of messy (e.g. it handles Vimeo in a completely separate and non-parallel way from YouTube), but another thing that happens in that code is Twitter. So I tried mentioning a tweet, and sure enough -- if I click on the Twitter user's avatar inside the embedded tweet, we pull up a blank lightbox. So that's still a bug.

There's also something about Dropbox, and I'm not sure that's everything, either; would have to study more closely before saying so, because that bit of server code is such a mess.

I'll go ahead and merge this, because it does improve the situation; just with an adjustment of the commit message to the appropriate level of confidence. :-)

gnprice:
OK, pushed -- with also a commit on top to better explain what's going on, and to fix at least the Twitter bug, in order to serialize out what I learned from studying this code just now. Thanks!

PR #2904, opened 2018-08-19
webview: Make WebView friendly error red

borisyankov:
Currently the 'friendly error' we show in production can easily
be mistaken for a normal header because of its background color.

Make the background red instead.

| Before | After |
| ---| --- |
| simulator screen shot - iphone 8 plus - 2018-08-19 at 21 52 03 | simulator screen shot - iphone 8 plus - 2018-08-19 at 22 42 58 |

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -404,7 +404,7 @@ blockquote {
 #js-error-plain, #js-error-plain-dummy {
   z-index: 1000;
   width: 100%;
-  background: #555;
+  background: red;
   color: white;
   font-size: 15px;
   padding: 4px;


borisyankov:
> I think the CSS red is sort of tough on my eyes...
Wouldn't this be a feature in that case? :)

jackrzhang:
No, because I think the color is negatively affecting the readability of the message 😃

borisyankov:
BTW, we already use 'red' for other errors (this one should be super rare, ideally never seen).

Of course, we can change that color, the benefit of 'red' vs `rgb(255, 50, 50)` is that it is readable.

Alternatively, named colors here:
https://www.rapidtables.com/web/css/css-color.html

`crimson` or `firebrick` look like good alternatives, though less obvious what they are than `red`.

gnprice:
Thanks @borisyankov for sending this, and @jackrzhang for the review!

Just merged. I agree with both of you :slightly_smiling_face: -- this will be clearer than the gray by being obviously something bad and an error, and also the actual text in it feels harder to read to me than it was with the gray, or than it probably should.

@borisyankov do you think you might experiment with some other colors, perhaps less saturated reds, and see if you can balance those factors?

I think it's OK if the color doesn't have a nice name. Most of the greens in the webapp don't have a nice name -- I think that's what tends to happen when colors are carefully chosen, unfortunately.

PR #2906, opened 2018-08-20
webview: Do not use Array.from in WebView

### LINKED ISSUE #2905: Uncaught TypeeError: Array.from is not a function
`Uncaught TypeeError: Array.from is not a function` exception is sometimes thrown inside WebView.

borisyankov:
Fixes #2905

`Array.from` is not available on all platform's WebViews we support.
One way to do this is to include a polyfill like this one:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill

Instead we'll use a simlper and sufficient approach - using `Array`'s `slice`
method called on the imageTags array-like object.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -7,6 +7,10 @@ import type {
   MessageInputTyping,
 } from '../webViewHandleUpdates';
 
+// Converts an array-like object to an actual array
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects
+const arrayFrom = arrayLike => Array.prototype.slice.call(arrayLike);
+
 // We pull out document.body in one place, and check it's not null, in order
 // to provide that assertion to the type-checker.
 const documentBody = document.body;
@@ -241,7 +245,7 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
 
 const appendAuthToImages = auth => {
   const imageTags = document.getElementsByTagName('img');
-  Array.from(imageTags).forEach(img => {
+  arrayFrom(imageTags).forEach(img => {
     if (!img.src.startsWith(auth.realm)) {
       return;
     }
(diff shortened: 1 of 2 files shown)


src/webview/js/js.js
```diff
@@ -220,7 +220,8 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
 
 const appendAuthToImages = auth => {
   const imageTags = document.getElementsByTagName('img');
-  Array.from(imageTags).forEach(img => {
+  const imageTagsArray = Array.prototype.slice.call(imageTags);
```
gnprice: This is a neat trick!

I feel like it calls for some explanation, or at least some isolation as a thing with a name; it's a bit puzzling at first look. ("Why isn't this reduced to just `imageTags.slice()`?)

Perhaps a function named `arrayFrom`? Perhaps also with a link to this handy section on the MDN page, about exactly this trick:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects

borisyankov:
Just a ping that this PR was updated.

src/webview/js/js.js
```diff
@@ -7,6 +7,10 @@ import type {
   MessageInputTyping,
 } from '../webViewHandleUpdates';
 
+// Converts an array-like object to an actual array
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects
+const arrayFrom = arrayLike => Array.prototype.slice.call(arrayLike);
```
jackrzhang: The comment could mention that Array.from won't necessarily be supported in the WebView, which is why this function exists in the first place.

It's not too much extra effort to look at the commit body, but I think that's the other important detail here, and it would be more visible in the comment.

gnprice:
Thanks @borisyankov , and @jackrzhang for the comment -- merged!

I agree with Jack's remark, and added a commit on top to do that.

Issue #2365, opened 2018-04-20
[Android] Broken-looking context menus.


zingmars:
Context menus in some places seem to be very broken looking.

![screenshot-menu-1](https://user-images.githubusercontent.com/917126/39050687-ea4669ba-44ae-11e8-9003-ee612f5cbbdd.jpg)
![screenshot-menu-2](https://user-images.githubusercontent.com/917126/39050694-edf38976-44ae-11e8-8531-a34308d3148b.jpg)

This is on Android 8.0, OnePlus A3003. App version v12.3.83 from GitHub Releases.

borisyankov:
Thanks for reporting this. I thought we fixed that, I guess not.

pulkonet:
@borisyankov This would only be fixed via expo and couldn't be handled on our part. Expo's component do not support OnePlus Slate font very well and that's the only reason why this and other expo components used look broken on OnePlus and Oppo phones. 

A little suggestion @zingmars, if you wanna overcome this -  Change your font to Roboto (it is available in settings>font). That is the only fix until expo updates their components and provide support for OnePlus slate font.

zingmars:
Yup, the menus show up fine with Roboto. Is there an issue about this in expo's vcs that I can track?

pulkonet:
@zingmars [Here](https://github.com/expo/expo/issues/1186) it is. It is not for action sheet but similar as this issue exists in all of the expo components.

jainkuniya:
@zulipbot claim

gnprice:
The root issue here appears to be facebook/react-native#15114 .

Still not clear to me if this is the fault of RN, or of those fonts, or what.

gnprice:
Small update FFR, since we just ran into the same underlying upstream bug in #3533:

It looks like the key thing that triggers the bug is **setting `fontWeight`**. E.g., trying to make text bold.

In particular:
* Some recent debugging [on the upstream bug](https://github.com/facebook/react-native/issues/15114#issuecomment-495318368) finds that that's the key variable.
* We worked around this issue originally by updating `react-native-action-sheet` to v1.1.0... and one of the few things that changed in that version was that it stopped setting `fontWeight`: https://github.com/expo/react-native-action-sheet/pull/52/files So that fits.

ishammahajan:
Some speculation, this might be due to the text box not recalculating its size based on the changes `fontWeight`.

PR #2910, opened 2018-08-22
action-sheet: Update `react-native-action-sheet` to `^1.1.0`.

jainkuniya:
This fixes the long pending issue of text in options being cut on One
Plus and Oppo phones.

Fix: #2365

Before

After

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "update-snapshots": "yarn run test:unit -- -u"
   },
   "dependencies": {
-    "@expo/react-native-action-sheet": "1.0.2",
+    "@expo/react-native-action-sheet": "^1.1.0",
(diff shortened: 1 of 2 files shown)


borisyankov:
You also forgot to mention that the styling is now material design and is way better!

jainkuniya:
> You also forgot to mention that the styling is now material design and is way better! I am excited!

Ya! I missed it, but screenshots helps in this regard 😅 

Also thanks, updated commit message.

gnprice:
Oof, that spacing though -- there's way too much at the top and bottom. (In that screenshot, and also when I try it on Android 8.1 in the emulator.)

I'll merge because it's not greatly worse, and on your phone that bug it fixes is much worse than this; but I am not so excited about it.

PR #2911, opened 2018-08-23
Fix had a hard time exiting message editing. 

### LINKED ISSUE #2908: Had a hard time exiting message editing
I couldn't figure out how to exit message editing (I was in a PM). The up button (left arrow) took me to the previous screen. (I forget now what else I tried.)

Opening as per https://chat.zulip.org/#narrow/stream/48-mobile/topic/exiting.20message.20edit. Let me know if more info would be useful.

jainkuniya:
Fix #2908 .


armaanahluwalia:
I’m still seeing a dialog that says “failed to edit message”. Does not go away till you open another stream or reopen the app.
This is on 18.0.99

gnprice:
> I am still strongly in favor of having the Edit Message behavior be a separate navigation route. This will automatically implement many behaviors (Back Android button, 'back' in nav bar) and will separate it more logically in the code.

I agree with this.

Also I like the direction of this change! Keeping this information in the global Redux state has always been a pretty nasty hack, and this moves it to a prop passed down from the parent (which is much better React style) and removes that whole chunk of the state tree.

There's a lot of changes here, so it calls for a detailed code review.

src/chat/ChatScreen.js
```diff
@@ -19,27 +23,86 @@ type Props = {
...
+  state = {
+    editMessage: null,
+  };
 
   static contextTypes = {
     styles: () => null,
   };
 
+  componentDidMount() {
+    BackHandler.addEventListener('hardwareBackPress', this.handleBackPress);
```
borisyankov: Not a great fan of all this handling.
I suspect this will not work with the back gesture on iOS.
That is one of the reasons I wanted to base this on React Navigation's route.

src/chat/ChatScreen.js
```diff
@@ -19,27 +23,86 @@ type Props = {
...
     const { styles } = this.context;
     const { narrow } = this.props.navigation.state.params;
 
     return (
       <ActionSheetProvider>
         <View style={styles.screen}>
           <ZulipStatusBar narrow={narrow} />
-          <MainNavBar narrow={narrow} />
+          <MainNavBar
+            cancelEditMode={editMessage ? this.cancelEditMode : undefined}
```
borisyankov: I am concerned.
When looking at this change and `Chat`, it should be concerning you too.
We should try to localize a change and to not make all components aware of it.

src/compose/ComposeBox.android.js
```diff
@@ -15,13 +15,7 @@ import type {
...
   GlobalState,
 } from '../types';
-import {
-  addToOutbox,
-  cancelEditMessage,
-  draftUpdate,
-  fetchTopicsForActiveStream,
-  sendTypingEvent,
-} from '../actions';
+import { addToOutbox, draftUpdate, fetchTopicsForActiveStream, sendTypingEvent } from '../actions';
```
borisyankov: I guess you will want to rebase the changes to this file as iOS and Android got merged.

jainkuniya: Yup! rebased.

src/chat/ChatScreen.js
jainkuniya: Now, even if user uses iOS gesture then also he is out of `editMode` + navigateBack (totally what he expects) :)

PR #2915, opened 2018-08-24
Fold redux-persist into the codebase.

roberthoenig:
We often find us in a situation where the redux-persist API is not as
well-crafted as it could be, or where the redux-persist API is lacking
features, especially with regards to the internal storage
implementation. redux-persist v5 offers new features, but they change
the internal storage implementation in a way that would leave us off
worse. In particular, the built-in migration system in redux-persist
v5 does not seem future-proof. In general, the whole library seems to
lack rigor and thought-through-ness.

Therefore, fold redux-persist v4.10.2 into the codebase. (Prior to
this commit, we were using v4.10.0). This will allow us to quickly
iterate on enhancing the API, and to rework the internal storage
implementation.

This commit clones the https://github.com/rt2zz/redux-persist
repository at the v4.10.2 tag. All the relevant source code resides in
redux-persist/src, and all the relevant flow type definitions in
redux-persist/src/type-definitions/index.js.flow. All other files and
directories are discarded. The relevant dependencies of this pruned
redux-persist repo are copied to zulip-mobile's package.json. The
dependency on npm's `redux-persist` is removed, and all imports from
that package are rerouted. All source files in redux-persist are
unchanged, except the `import isPlainObject from
'lodash/isPlainObject'` -- it gets changed to `import isPlainObject
from 'lodash.isplainobject'` along with the respective dependency.

The redux-persist codebase does not adhere the eslint and prettier
style guides, so tests will fail. The following commit will fix this.


roberthoenig:
Updated to format redux-persist with prettier and eslint. Ready for review.

gnprice:
We've now finally merged in a version of this change! Namely e5409c57833c9239b98482f4a8ff4a56c730f85d / #4047. Thanks @roberthoenig for your work on investigating this and figuring out what we wanted to do.

Issue #2898, opened 2018-08-17
Do not include debug build launcher app icon in release APK/package.


jainkuniya:
Currently there is no logic which says, `ic_debug_launcher` is only used in debug build, so it is also present in release build. We can save few space (but very less). But for future this might be helpful, when we have more build based resources.

For reference:
https://stackoverflow.com/questions/33760518/android-gradle-have-both-debug-and-release-resources-in-different-flavors

jainkuniya:
@zulipbot label "good first issue"

prateek3255:
@zulipbot claim

prateek3255:
I want to work on this issue.

prateek3255:
@jainkuniya can you review my PR #2916 for this issue.

PR #2916, opened 2018-08-25
Fix issue 2898 - created seprarate resources for build and debug build 

prateek3255:
This PR fixes the issue #2898.

I refered these [docs](https://developer.android.com/studio/build/build-variants#sourcesets) and the comment no. 3 of [this](https://stackoverflow.com/a/33760583/8252081) StackOverflow post.

I found that all the common files such as manifest files, java files, common resources files and one of both the debug and release resources go into the main folder, and the different files for release and debug such as launcher icons in our case go into their respective debug and release source folders.

The resource files are overriden according to build type because the compiler expects them to be at these locations on the basis of build variant

```
debug
-----
Compile configuration: debugCompile
build.gradle name: android.sourceSets.debug
Java sources: [app\src\debug\java]
Manifest file: app\src\debug\AndroidManifest.xml
Android resources: [app\src\debug\res]
Assets: [app\src\debug\assets]
AIDL sources: [app\src\debug\aidl]
RenderScript sources: [app\src\debug\rs]
JNI sources: [app\src\debug\jni]
JNI libraries: [app\src\debug\jniLibs]
Java-style resources: [app\src\debug\resources]

main
----
Compile configuration: compile
build.gradle name: android.sourceSets.main
Java sources: [app\src\main\java]
Manifest file: app\src\main\AndroidManifest.xml
Android resources: [app\src\main\res]
Assets: [app\src\main\assets]
AIDL sources: [app\src\main\aidl]
RenderScript sources: [app\src\main\rs]
JNI sources: [app\src\main\jni]
JNI libraries: [app\src\main\jniLibs]
Java-style resources: [app\src\main\resources]

release
-------
Compile configuration: releaseCompile
build.gradle name: android.sourceSets.release
Java sources: [app\src\release\java]
Manifest file: app\src\release\AndroidManifest.xml
Android resources: [app\src\release\res]
Assets: [app\src\release\assets]
AIDL sources: [app\src\release\aidl]
RenderScript sources: [app\src\release\rs]
JNI sources: [app\src\release\jni]
JNI libraries: [app\src\release\jniLibs]
Java-style resources: [app\src\release\resources]
```


android/app/build.gradle
```diff
@@ -186,6 +186,8 @@ dependencies {
 
 }
 
+
```
jainkuniya: I guess this changes are done accidentally.

package-lock.json
```diff
@@ -0,0 +1,13320 @@
+{
```
jainkuniya: this too?

android/app/build.gradle
prateek3255: Yeah I guess this was made accidentally

package-lock.json
prateek3255: I didn't touch this file

prateek3255: how do I resolve this file?

android/app/build.gradle
jainkuniya: Please remove this change. :)

package-lock.json
jainkuniya: Do a `-i`  rebase and select `edit` for the commit in which this got commited, and do not add this file (add all other files) and then `git rebase --continue`.

Let me know on chat.zulip.org if you face any problem. :)

gnprice:
Thanks @prateek3255 for digging into this, and @jainkuniya for the previous review!

A couple of comments:
* This leaves us with two copies of each of these files: one in `.../src/main/`, and one in `.../src/debug/` or `.../src/release/`. We should have just one copy to avoid confusion.
* We can leave the "normal" version of each file under `src/main/`. That way it'll get used in the release build, and overriden in the debug build, for the same effect as in this version; but it feels a bit clearer to me, by explicitly marking which one we think of as "the" logo, the normal one to think of if you're not paying special attention to the release/debug distinction.

Those changes are super easy for me to make, so I'll just do them.

gnprice:
Oh also:
* Since we're using this nice clean source-sets mechanism, we don't need the `ic_debug_launcher` thing anymore :) I see @jainkuniya went and fixed that aspect up in the commit he included in #2897 , so I'll use that fix.

PR #2919, opened 2018-08-26
Fix and improve autocomplete popup size

### LINKED ISSUE #2917: Autocomplete popup behaves erraticly
A recent regression leads to visual issues in the autocomplete popup.
The actual issue is hard to describe but manifests in appearing/disappearing or varying heights of the item list.

borisyankov:
Fixes #2917 

Makes the sizing of the autocomplete popup more consistent and logical.

### DIFF
--- a/src/autocomplete/AutocompleteViewWrapper.js
+++ b/src/autocomplete/AutocompleteViewWrapper.js
@@ -1,6 +1,6 @@
 /* @flow */
 import React, { PureComponent } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { Dimensions, StyleSheet, View } from 'react-native';
 
 import type { InputSelectionType, Narrow } from '../types';
 import AutocompleteView from './AutocompleteView';
@@ -11,6 +11,7 @@ const styles = StyleSheet.create({
     position: 'absolute',
     bottom: 0,
     width: '100%',
+    justifyContent: 'flex-end',
   },
 });
 
@@ -39,8 +40,10 @@ export default class AutocompleteViewWrapper extends PureComponent<Props> {
       onMessageAutocomplete,
       onTopicAutocomplete,
     } = this.props;
+    const { height } = Dimensions.get('window');
+
     return (
-      <View style={[styles.wrapper, { marginBottom }]}>
+      <View style={[styles.wrapper, { height: height / 2 - marginBottom, marginBottom }]}>
         <TopicAutocomplete
           isFocused={isTopicFocused}
           narrow={narrow}
--- a/src/common/Popup.js
+++ b/src/common/Popup.js
@@ -7,12 +7,11 @@ import type { Context } from '../types';
 
 const styles = StyleSheet.create({
   popup: {
-    marginRight: 20,
-    marginLeft: 20,
-    marginBottom: 2,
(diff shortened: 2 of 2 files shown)


src/autocomplete/AutocompleteViewWrapper.js
```diff
@@ -39,8 +40,10 @@ export default class AutocompleteViewWrapper extends PureComponent<Props> {
       onMessageAutocomplete,
       onTopicAutocomplete,
     } = this.props;
+    const { height } = Dimensions.get('window');
+
     return (
-      <View style={[styles.wrapper, { marginBottom }]}>
+      <View style={[styles.wrapper, { height: height / 2 - marginBottom, marginBottom }]}>
```
jainkuniya: (Before we had it as 1/4 of height, to tackle landscape scenario too b4cd7c829a7d17888902f5468d4ea0a0f1084c42)

gnprice:
For me this results in some bad mis-layout as soon as I focus the compose box:
screenshot_1537823801
The app bar is gone; plus the input is partially obscured by the keyboard.

That's on an emulated Pixel 2 XL, on Android 8.

Perhaps the issue is that the `render` method with the `Dimensions.get` call isn't getting re-run when the keyboard pops up, and so we're still basing the size on the old dimensions?

More broadly, even if we find a way to do this that seems to work, I'd worry about edge cases where we don't get re-rendered -- even if we find a way that it's clearly documented should work, this feels like an area where RN is likely to have bugs in edge cases, because at least on Android it's a notoriously tricky area to get right. (What if the user enters split-screen mode? Or leaves it? Or moves their focus between split screens? Or if the keyboard itself aborts, or changes its height?)

So I'd be more confident if we can find a solution that doesn't involve our code having to find out the height.

gnprice:
Ah, playing with the `react-devtools` inspector a bit, I think I see the reason this is tricky: it's that we have `AutocompleteViewWrapper` as a child of `ComposeBox`, when the layout we're going for really calls for making it a sibling of `ComposeBox` and of `MessageList`.

Pulling it up to be a sibling will require some tricky rewiring of how the autocomplete communicates with the input fields, as it'll be one layer farther away. I think that's totally doable, but it'll be less work if we wait to do it until we've de-duplicated ComposeBox again -- which we had a previous effort at in #2886, but now is probably best done after upgrading to RN v0.57, which in turn will follow #2789 upgrading to v0.56.

Meanwhile, we should do something to make the autocomplete at least not look broken. How about setting its height to a fixed number of rows -- like 5? That should be an improvement over the status quo.

gnprice:
@borisyankov and I chatted a bit about this today.
* The second commit is a simple styling fixup that's independent of the major layout change. I'm going to go ahead and merge it.
* The other commit, we'll skip. As discussed above, we'll come back and fix this the right way after we lay some groundwork; in the meantime we'll do a quick fix like a fixed height.

gnprice:
As always for a UI change, some screenshots make a key piece of the PR. For this one, I took some just now.

These are for just the smaller (second) commit on its own.

Before:
image

After:
image

gnprice:
Merged that piece as 69ece87fa. Thanks @borisyankov !

PR #2920, opened 2018-08-27
Fix various typos.

roberthoenig:
These typos were found by an experimental upgrade to
eslint-plugin-spellcheck v0.0.9

### DIFF
--- a/src/streams/SubscriptionsCard.js
+++ b/src/streams/SubscriptionsCard.js
@@ -51,7 +51,7 @@ class SubscriptionsCard extends PureComponent<Props> {
 
 export default connect((state: GlobalState, props) => ({
   narrow: props.narrow || [],
-  // Main scrren long longer conatin drawer,
+  // Main screen no longer contains drawer,
   // so at any position we cannot show selected stream in the list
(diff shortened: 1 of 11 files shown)


src/streams/SubscriptionsCard.js
```diff
@@ -51,7 +51,7 @@ class SubscriptionsCard extends PureComponent<Props> {
 
 export default connect((state: GlobalState, props) => ({
   narrow: props.narrow || [],
-  // Main scrren long longer conatin drawer,
+  // Main screen no longer contains drawer,
```
borisyankov: I am not sure if that comment makes sense as part of the source code.
Maybe it did make sense as a commit message. I would prefer to completely remove it.

roberthoenig: Sure. Should be done in a follow-up commit?

borisyankov: Sure, a spellchecking vs removing a comment sounds like a different type of work.

gnprice:
Merged -- thanks @roberthoenig and reviewers!

Issue #2052, opened 2018-03-13
Cover our code with Flow types


gnprice:
Static types [can make a big difference](http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/) in making it easier to understand a codebase. That's valuable because it helps us write more reliable code with fewer bugs, and making the app more reliable with fewer bugs is our top priority right now. It's also valuable because it helps people learn the codebase more quickly, and we'll likely have several people learning the codebase over the next few months for GSoC.

We use [Flow](https://flow.org/en/docs/types/) for static types in the Zulip mobile app, but there's still a lot of code that doesn't have types. Let's add more. Here's one way to do it:
1. Look in `src/types.js` for a type definition that just says `any` and then has a real, more detailed type commented out.
2. Replace `any` with a real type. Try the one commented out next to it. This type is probably wrong but is a good guess to start from.
3. Run `yarn run test:flow` and see what errors you get. Study the errors and the surrounding code to figure out what the type should really be. Read [the Flow docs](https://flow.org/en/docs/types/) to help understand what the types mean.
4. Try fixing the types. This might mean editing the type definition you just added; or editing the code that uses it to refer to some other type instead; or editing other type definitions; or some other fix. Use what you've learned from studying the code!
5. Repeat steps 3 and 4 until you get no errors.
6. Commit your work as [clear, coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html). Use the commit messages to explain anything you learned that won't be obvious to someone reading the commits. Then send a PR :smile: 

Some other ways to find untyped code:
* You can run `yarn run test:flow-coverage` to get a nice list of files with lots of uncovered code, and click on the filenames to see specific lines. (Also helpful is `yarn run test:flow coverage --color` to see details on any one file.)
* You can run `yarn run test:flow-todo` to see a list of places we could have type annotations but don't, or have one like `any` that doesn't mean anything.

You can do either of those in place of step 1 above -- then steps 2 to 6 are similar.

Here's a few examples of what a good commit for this can look like:
- 835e117ab a simple example
- 212da36d7 where the right thing was to remove that type and fix the annotations
- c1c68cb75 a more complex example
- ac04b1099a another complex example (though [splitting into several commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) would have made it even better)

For tracking our progress: right now there are
* **44** uses of `any` or `Object` in `src/types.js`
[...]

PR #2928, opened 2018-08-29
Added type MapStateToProps in types.js

flowkraD:
With respect to issue #2052, types.js contained two type definition as any with a more detailed comment next to it. One of the types were replaced with their static types and were tested with Flow which indicated no errors. The code was formatted with `yarn prettier` after going through the CONTRIBUTING.md.

The Travis CI build failed for AnimatedValue type after 2 minutes. I apologise for overlooking it. The MapState ToProps passed and hence can be added without any problems. Again I would like to apologise for the inconveniences caused with the AnimatedValue type commit.

I hope the changes made lives up to the standards of the organization (^_^)

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -7,7 +7,7 @@ import type { AppStyles } from './styles/theme';
 
 export type { ChildrenArray } from 'react';
 
-export type AnimatedValue = any; // { AnimatedValue } from 'react-native';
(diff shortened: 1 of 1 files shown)


flowkraD:
Travis CI failed to identify MapStateToProps from `react-redux`.  Sorry for the inconveniences.

PR #2931, opened 2018-08-29
Improve draft saving reliability

borisyankov:
Fixes #2931 

Refactors draft saving code, creating a new action `updateDraft`.
This extracts the logic out of `ComposeBox` and makes it self-contained. Added tests for that too.

This also changes the behavior of the draft to save on each keystroke, thus ensuring no update is lost.


src/drafts/draftsActions.js
```diff
@@ -12,3 +12,14 @@ export const draftRemove = (narrow: Narrow): DraftRemoveAction => ({
   type: DRAFT_REMOVE,
   narrow,
 });
+
+export const updateDraft = (narrow: Narrow, message: string) => (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
+  if (message.trim().length === 0) {
```
jainkuniya: Just a thought: I guess we can also remove `draftRemove` and `draftAdd`, and completely rely on `draftUpdate`. 

And in `reducers`, if `message.trim().length === 0` then return state which does't have that `narrow` key in `drafts` else update drafts.

Basically moving this if condition to `reducers`.

Thoughts?

gnprice:
Thanks @borisyankov ! This will be a good fix, and I like how much simpler ComposeBox gets.

I think there's actually a significant bug this version may cause, though. There are a variety of places we call `handleMessageChange`, which this adds a `draftUpdate` dispatch to. One of those is `componentWillReceiveProps` (directly on iOS, indirectly on Android) -- and in particular, it looks like if we move from one narrow to another, this will cause us to overwrite *this* narrow's draft with whatever the draft might be for the *next* narrow (usually empty, I imagine.)

A version of the bug is actually already present in that we'll send one last typing event on the old narrow; the impact of that one is obviously a lot milder. :slightly_smiling_face: 

Looking at this, I think the root of the problem basically is that we're attempting to process `nextProps` while `this.props` is still the old props; to do that correctly, we'd need to carefully isolate what methods we call and how they all refer to `this.props`. How about moving this to a `componentDidUpdate` instead?

---

Smaller feedback:

Which issue did you mean to refer to? Not #2931, I think :wink: . Please mention it in the relevant commit message, in addition to the PR -- that helps cross-reference things when reading the Git history.

src/drafts/draftsActions.js
gnprice: I agree -- this logic feels like it'd be cleanest to push inside the reducer, and have just one action type like `DRAFT_UPDATE`. Not a blocker for this branch, though; it's already progress to move it out of ComposeBox.

Ideally the "skip if no change" would live in the reducer too -- after all, that's an important part of every reducer's job already. Does this logic really cut down that much on how many actions we end up dispatching? I don't see why we'd be getting often to `updateDraft` when the text hasn't changed.

jainkuniya:
> editMessage

(would like to mention that #M2911 also deals with `editMessage` :))

borisyankov:
Have you reproduced the 'wrong narrow' issue or are you basing this on code review alone?
I don't think this is a scenario that should happen.

We are getting the `narrow` value from a couple levels above from `ChatScreen` which itself gets it from `navigation.state.params`.

When the user navigates to a different narrow, React Navigation creates a different instance of `ChatScreen`.

gnprice:
> I don't think this is a scenario that should happen.
> [...]
> When the user navigates to a different narrow, React Navigation creates a different instance of `ChatScreen`.

In that case, does the code I'm referring to do anything at all? I'll quote it here in hopes of making the discussion clearer (this is the `ios` version):

```
  componentWillReceiveProps(nextProps: Props) {
    if (/* ... */) { /* ... */
    } else if (!isEqual(nextProps.narrow, this.props.narrow)) {
      if (!nextProps.draft) {
        this.clearMessageInput();
      }

      this.handleMessageChange(nextProps.draft);
    }
  }
```

If I understand your reply correctly, you're saying that this method will never be called with `narrow` changed, and therefore this whole block never runs and so this isn't a live bug.

If the only reason some code isn't a live bug is because it's never called at all, then the right thing is to delete it. If you believe that's the case here, then let's have the branch include a commit that deletes this block, and only after that make this change that would make this block a lot buggier.

Please also address the other, smaller thing I mentioned.

borisyankov:
That is a good point. I suspect this does not do anything. We used to get the `narrow` from state and there would have executed.

Also, my goal is definitely to make the `componentWillReceiveProps` function inside `ComposeBox completely.

If indeed, this code never runs, I will remove it in a separate commit.
I guess we want a definitive answer to this question before merging the PR.

gnprice:
> If indeed, this code never runs, I will remove it in a separate commit.
> I guess we want a definitive answer to this question before merging the PR.

Sounds good! I'd say that's one of the functions of making it a separate commit, in fact -- it neatly frames the question "does this in fact never run" disentangled from the other things we're doing in the branch, and gives that change its own commit message which is a place to put our answer to that question :smiley:

gnprice:
To make the current state of this PR explicit:
* #2996 is merged! :tada: Did that yesterday.
* These changes need rebasing to go on top of that.
* There's a smaller bit of feedback above that's still outstanding:
   > Which issue did you mean to refer to? Not #2931, I think :wink: . Please mention it in the relevant commit message, in addition to the PR -- that helps cross-reference things when reading the Git history.

Looking forward to getting this in!

(My suggested workflow was to put the changes that went into #2996 in *this* branch, at the beginning. One reason for that is that it lets you do that rebasing up front -- so you could add that change, rebase the rest of this on top of it, address the smaller bit of feedback too, and then I'd likely have been able to merge the whole thing yesterday.)

gnprice:
Thanks @borisyankov !

Reading the new version of this where the `narrow` case of `componentWillReceiveProps` is gone, I see there's still a `setMessageInputValue` call inside the remaining, `editMessage` case. I think that means that essentially the same bug I was concerned about above is still present in that case.

For that one, I think the code is not dead, so we don't get that extra-happy way of solving the issue. How about this solution I mentioned above, instead?

> I think the root of the problem basically is that we're attempting to process `nextProps` while `this.props` is still the old props; to do that correctly, we'd need to carefully isolate what methods we call and how they all refer to `this.props`. How about moving this to a `componentDidUpdate` instead?

As usual, that would go as a separate commit in this branch, at the beginning.

Smaller comment: can we skip adding the `if (drafts[JSON.stringify(narrow)] === message)` logic to `updateDraft`? I think we don't actually need that bit of complexity (small though it is).

borisyankov:
I did discard the 2nd commit completely as the 4th one was reworking it anyway. (the `if (drafts[JSON.stringify(narrow)] === message) logic to updateDraft`)

borisyankov:
While I want to completely revamp how we handle `editMessage`, I don't think there is a bug here.

```js
handleMessageChange = (message: string) => {
    this.setState({ message, isMenuExpanded: false });
    const { dispatch, narrow } = this.props;
    dispatch(sendTypingEvent(narrow));
  };
```

`draftUpdate` is called with `narrow` that will never change, and `message` that is gotten as function argument, gotten through `setMessageInputValue` again as argument.

PR #2933, opened 2018-09-01
Remove all 'response extraction' code

borisyankov:
Based on the API PR.

At this point, no function does use the response extraction concept and all use the `res => res` default null function.
    
Removing this is simple and risk-free.
    
    * remove any Flow types associated with this
    * remove the parameter from all core API functions
    * remove the parameter from all specific API functions


gnprice:
We'll sequence this after #2819 .

PR #2934, opened 2018-09-01
More robust caughtUp logic

### LINKED ISSUE #3338: 'All messages' view fails to show all messages
For `HOME_NARROW` if there are any unread messages the caught-up logic fails and considers incorrectly that `caughtUp.older` is `true` thus making the app not being able to load older messages.

borisyankov:
Fixes #3338

Previously we had to use custom caught-up logic to determine if we
have loaded the oldest or newest messages in a narrow.

Back-end now gives us that data while older versions do not.

Keep the old logic when the `found_newest` and `found_oldest`
values are not provided but use them directly if present.

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -222,8 +222,8 @@ type MessageFetchCompleteAction = {|
   anchor: number,
   numBefore: number,
   numAfter: number,
-  foundNewest: ?boolean,
-  foundOldest: ?boolean,
+  foundNewest: boolean | void,
+  foundOldest: boolean | void,
 |};
 
 type InitialFetchStartAction = {|
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -222,6 +222,8 @@ type MessageFetchCompleteAction = {|
   anchor: number,
   numBefore: number,
   numAfter: number,
+  foundNewest: ?boolean,
+  foundOldest: ?boolean,
 |};
 
 type InitialFetchStartAction = {|
--- a/src/api/messages/getMessages.js
+++ b/src/api/messages/getMessages.js
@@ -5,13 +5,20 @@ import { apiGet } from '../apiFetch';
 type ApiResponseMessages = {|
   ...ApiResponseSuccess,
   anchor: number,
-  found_anchor: boolean,
-  found_newest: boolean,
-  found_oldest: boolean,
+  found_anchor?: boolean,
+  found_newest?: boolean,
+  found_oldest?: boolean,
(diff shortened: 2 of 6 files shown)


borisyankov:
@gnprice This is the PR about the 'better caught up'. I guess I can unbundle it from the API PR.

borisyankov:
This is ready to go after #3246

gnprice:
Thanks @borisyankov !

The `messageFetchComplete` action creator feels like it's overdue now for switching from positional params to a params object. Want to try that change? (Could be a commit on top.)

---

```
+  if (action.foundNewest !== undefined && action.foundOldest !== undefined) {
```

The type written down for the action is supposed to mean this condition can never fail -- these are simply `boolean`. That's a sign that the types aren't right.

I think step 1 of the fix is that the types in the API should say like `foundNewest?: boolean`, with a jsdoc comment saying they're missing in Zulip server versions before vX.Y. (Are they indeed always present in newer servers' responses?) Then it's up to the application code using the API to decide how to handle the case where they're missing.

I haven't thought fully through the options for that layer, but certainly one answer is to propagate the `?:` typing to the action, and have exactly this version's code in the reducer.

Looks like we use `numBefore` and `numAfter` in one other place, in `fetchingReducers`. Can those be replaced by using these too?

If so, another answer might be to move this compatibility-fallback code into the action creator -- and leave `numBefore` and `numAfter` out of the action object.

If we go for that latter design, it'd be OK to move to it as followup commits after something a lot like this version. But the types being consistent, I'd like to see from the beginning.

borisyankov:
* rebased
* added proper type and docs for `found_newest` and `found_oldest` props
* now fixes an actual value (reproduced & verified this fixes it)

Making function params an object is implemented, but required a bit too many changes, so sitting in a branch on my machine in favor of merging this faster.

gnprice:
Thanks @borisyankov for the updates! Merged.

I added a few commits on top with some tweaks. Take a look in particular at the `?foo` one (46a818108), which is a pattern I recently became more conscious of.

> Making function params an object is implemented, but required a bit too many changes, so sitting in a branch on my machine in favor of merging this faster.

Sounds good -- will be glad to see it.

I'm also still curious about this question of mine above, which points to another possible followup:

> Looks like we use `numBefore` and `numAfter` in one other place, in `fetchingReducers`. Can those be replaced by using these too?
> 
> If so, another answer might be to move this compatibility-fallback code into the action creator -- and leave `numBefore` and `numAfter` out of the action object.

PR #2935, opened 2018-09-03
Rename 'peace_symbol' to 'peace'

agzuniverse:
peace_symbol was incorrect and did not produce the emoji when used in
a message.

Issue reported [here](https://chat.zulip.org/#narrow/stream/48-mobile/subject/Peace.20emoji/near/640962).

### DIFF
--- a/src/emoji/codePointMap.js
+++ b/src/emoji/codePointMap.js
@@ -45,7 +45,7 @@ export const codePointMap: { [name: string]: string } = {
   heart_with_arrow: '1f498',
   cooking: '1f373',
   new: '1f195',
-  peace_symbol: '262e',
+  peace: '262e',
   pancakes: '1f95e',
   herb: '1f33f',
   pouting_cat: '1f63e',


jainkuniya:
> We'll copy the current version of that data source. That's a quick fix I'll be happy with. :-)

Open PR fot this https://github.com/zulip/zulip-mobile/pull/2869 :)

HarshitOnGitHub:
@gnprice Sorry for a late reply! I was from laptop due to my mid semester examination. The file that you should be looking at is `static/generated/emoji/name_to_codepoint.json`. This file gets generated after provisioning. The idea of checking in this file seems wrong because if mobile app connects to some older server version(1.6 for example) this problem will still persist. Their webapp would still be serving the older names while mobile app will be having new ones.
`name_to_codepoint.json` is a json file so if it can be fetched from the server and persisted to disk when the user adds a new server it will solve the problem.

agzuniverse:
Thanks for the clarification @HarshitOnGitHub ! So this seems to be a slightly bigger issue than it initially seemed, because the mobile app currently uses a fixed set of emojis that comes with the app.

Leaving this PR open since it still applies as a quick fix for the original issue, but we need a better fix for a long term solution.

jainkuniya:
@agzuniverse You can close this PR. We will be going with #2869, which also fixes more issues like this :)

Issue #2791, opened 2018-07-12
Autocompletes pop up splashes for a moment just after typing prefix (@, #, :) if prefix is after some text in the input box.


jainkuniya:
**Steps to reproduce**
* Enter some text in the compose box.
* Enter prefix (Either `@`, `:` or `#`).
* Observe carefully that just after typing prefix, autocomplete pop is visible for very small time period.
It is only reproducible if prefix is after some text. 

[Autocomplete-splashes.mov](https://chat.zulip.org/user_uploads/2/37/TGA9eEmFLzO8gxDx0dMTa09-/Autocomplete-splashes.mov) 

@zulipbot claim

PR #2942, opened 2018-09-06
autocompleteView: Return null if prefix is neither :,@ nor #.

jainkuniya:
`getAutocompleteFilter` returns `lastWordPrefix` as either `:`, `@`,
`#` or ''. If it is a empty string that means user have entered normal
text, for which we are not interested and don't want to show
autocomplete popup. So return null from the component and render
nothing.

Fix: #2791

### DIFF
--- a/src/autocomplete/AutocompleteView.js
+++ b/src/autocomplete/AutocompleteView.js
@@ -34,10 +34,16 @@ export default class AutocompleteView extends PureComponent<Props> {
     const { text, selection } = this.props;
     const result = getAutocompleteFilter(text, selection);
 
-    const AutocompleteComponent = prefixToComponent[result.lastWordPrefix];
+    const { filter, lastWordPrefix } = result;
+
+    if (lastWordPrefix.length === 0) {
+      return null;
+    }
+
+    const AutocompleteComponent = prefixToComponent[lastWordPrefix];
 
     return (
-      <AnimatedScaleComponent visible={result.filter.length > 0}>
+      <AnimatedScaleComponent visible={filter.length > 0}>
         {AutocompleteComponent && (
           <AutocompleteComponent filter={result.filter} onAutocomplete={this.handleAutocomplete} />
         )}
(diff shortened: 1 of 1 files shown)


src/autocomplete/AutocompleteView.js
```diff
@@ -34,10 +34,16 @@ export default class AutocompleteView extends PureComponent<Props> {
     const { text, selection } = this.props;
     const result = getAutocompleteFilter(text, selection);
 
-    const AutocompleteComponent = prefixToComponent[result.lastWordPrefix];
+    const { filter, lastWordPrefix } = result;
```
borisyankov: I don't like the overuse of destructuring.
One reason we do it for `this.props` is that it saves us two levels of object access (`this` and `props`), and another is that it is simpler and consistent if we use it everywhere.

gnprice:
Thanks @jainkuniya !

This logic is pretty convoluted; it's hard to tell exactly what bug each change is intended to fix, what your diagnosis is of what's causing that bug, and what your reasoning is for why the change will fix it.

I just pushed a commit bda7b6627 that fixes #2987, which is the really awful symptom I've been seeing recently, and seeing other users complain about, that's in this area. That should put all the remaining issues into the category of things there's no pressure for us to fix immediately, so we have no reason not to take the time to understand what we're doing and why.

It looks like #2944 makes all the same substantive changes as this PR, among others. Let's close this one in order to focus discussion there.

src/autocomplete/AutocompleteView.js
gnprice: I think in this instance this is helpful; I just pushed 4fe1deae8 which does a version of this, with this explanation why:
```
    autocomplete: Tighten code slightly with destructuring-bind.
    
    Morally this helper `getAutocompleteFilter` is returning two separate
    values, named `lastWordPrefix` and `filter`; the object that collects
    them together doesn't have any real identity of its own, and only
    exists for the purpose of being something we can `return` because JS,
    like most languages, doesn't have a first-class concept of "return
    multiple values".
    
    So, treat it that way at the call site by immediately turning the
    returned object back into the two separate values it represents.
```

gnprice: > Also, when used for props and state only, it is easier to say 'oh, this variable without an object is either local or coming from props', while if using it for everything, that is not the case.

I don't really understand this objection. These are indeed local variables, coming from the return value of a function we called just like local variables often do... what's the notion of "local" that you have in mind that these don't count as?

Issue #2943, opened 2018-09-07
Animation on autocomplete popup are not visible.


jainkuniya:
We are using `AnimatedScaleComponent` to animate autocomplete popup, but there seems to be some issue in it. Animation is not visible.

@zulipbot claim

PR #2944, opened 2018-09-07
Fix #2791, #2943. Improve autocomplete pop animation and fix unwanted splash.

### LINKED ISSUE #2791: Autocompletes pop up splashes for a moment just after typing prefix (@, #, :) if prefix is after some text in the input box.

jainkuniya:
Ready to initial review.

Not completely sure, if we want last commit or not (b169953).

One thing which is remaining is: Show scale down animation when user selects item from the pop up. or shown when autocomplete hides because of such filter string is entered for which we don't have anything to show as suggestions.


gnprice:
(The commit message needs to answer those questions before it's merged; but feel free to start by trying to answer them here, or in chat.)

gnprice:
Thanks for the detailed explanation in that commit message! That was quite helpful in understanding what you'd learned about the issue. Based in part on that information (and after substantial discussion in chat), I just pushed 0baca8fce to fix #2791 ; see that commit message for my diagnosis, which isn't quite the same as your eventual conclusions.

gnprice:
I think some of the logic in your subsequent commits will want to change after that fix -- hopefully get simpler. Would you rebase, make those changes, and then edit the commit messages as needed to explain what they're doing in this new version?

jainkuniya:
(putting this on hold till #3027 get's merged/reviewed, because two commits of this are alternative of #3027)

PR #2945, opened 2018-09-07
Make 'Send' button easier to press

borisyankov:
Makes the 'Send' button easier to press by enlarging the area that is touchable.

First, it changes the `FloatingActionButton` so styling works better.
Then, adds 8 pixels padding around the button that are now 'touchable'.
(touchable are goes from 32x32 to 48x48)


borisyankov:
Duh. Did we figure out what were the issues with the latest Android? Maybe it is the same thing. I am seeing related layout issues on AccountPickerScreen. (no vertical space between the items)

borisyankov:
This should be fixed by #3013

gnprice:
Thanks @borisyankov for the fix and the detailed explanations in commit messages, and thanks @jainkuniya for the review! Just merged.

Issue #2940, opened 2018-09-05
Message position is sometimes incorrect


borisyankov:
It was reported that on our latest release we sometimes incorrectly calculate the message position.
One specific case is on apps startup the position would match message that was last read on the phone but not the actually latest read messages.

gnprice:
The specific issue described here, i.e.:

> One specific case is on apps startup the position would match message that was last read on the phone but not the actually latest read messages.

was fixed in f53f8311a, as part of #2843 . (Rather a while ago, soon after this issue was filed.)

The symptom "incorrect position" can happen with other causes, but we'll track those as separate issues. See #3457 and #3301 for two that are open now.

PR #2947, opened 2018-09-08
Improve message list positioning issues

borisyankov:
Improve cached data correctness by preserving the `flags` state instead of recalculating it (incompletely)

More details available in each commit message.

### DIFF
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -19,7 +19,7 @@ import createMigration from '../redux-persist-migrate/index';
  */
 // prettier-ignore
 export const discardKeys = [
-  'alertWords', 'caughtUp', 'fetching', 'flags', 'loading',
+  'alertWords', 'caughtUp', 'fetching', 'loading',
   'nav', 'presence', 'session', 'topics', 'typing',
 ];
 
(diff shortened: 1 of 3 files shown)


gnprice:
Sorry I missed this PR at the time!

We've since made this fix as f53f8311a -- in which the code changes appear in fact to be identical to squashing these two commits.

For handy reference in following up on #2940, here's the commit messages from this PR:
```
Do not discard but store `flags` data

Improves but does not completely resolve #2940

On app's start there is a potential (very likely) mismatch between
the stored messages and the flags associated with them (read, starred)

The first issue is that we do not update the messages themselves
but we do update the flags state. But we do not preserve the state!

The second issue is that the flags might have changed since the last
run, again very likely, especially for the `read` flag if the user
have used the web app.

This commit resolves the first issue. The second one requires more
work, mostly the code implementing short/long-lived queues.
```

```
state: Do not rehydrate 'flags' state

Previous commit makes sure we store the `flags` state.

This is better than trying to restore flags from `messages` as we
do not mutate this data. So any state that was changed during the
previous run of the app would not get restored properly.

The simpler solution of storing the `flags` is far superior to this
approach while not being a complete solution.
```

Issue #2866, opened 2018-08-04
Show a separator in the unread list, between pinned streams and other streams same as in subscribed stream tab.


jainkuniya:
@zulipbot claim

Issue #2865, opened 2018-08-04
Sometimes sorting of unread stream in unread screen is not correct.


jainkuniya:
How checkin is before than announce. 🤔 
<img width="323" alt="screen shot 2018-08-04 at 7 57 31 pm" src="https://user-images.githubusercontent.com/18511177/43677516-c84842b4-9820-11e8-8ba6-aee71df321f8.png">

@zulipbot claim

borisyankov:
Is #checkins pinned?

jainkuniya:
#checkins & #announce, both are pinned.

PR #2948, opened 2018-09-08
Improve unread screen. Fix: #2865 & #2866.

jainkuniya:
* unreadScreen: Sort pinned & non pinned streams internally. b5fed4a
Now both the sections (pinned and non-pinned) streams are sorted
internally in alphabetically order.
Fix: #2865

* unreadScreen: Show a swparator b/w pinned and non pinned streams. e916ffd
Fix: #2866.


src/unread/unreadSelectors.js
```diff
@@ -140,9 +140,15 @@ export const getUnreadStreamsAndTopics = createSelector(
       return totals;
     }, {});
 
-    const sortedStreams = Object.values(unreadMap)
-      .sort(caseInsensitiveCompareObjFunc('streamName'))
-      .sort((a: any, b: any): number => +b.isPinned - +a.isPinned);
+    const sortedStreams = Object.values(unreadMap).sort((a: any, b: any): number => {
+      // if both are pinned
```
borisyankov: Do you need three lines for this comment? :)

jainkuniya: I liked this break down, easy to read 😅 
Should I move to one line?

borisyankov: Yeah :)

gnprice:
Hi @jainkuniya !

This seems like a good UI change.

In the second commit, we start separating the array into two arrays for `isPinned` and `!isPinned` entries respectively.

Given that we're doing that, it seems like there's not much point in sorting the original array by `isPinned`. :slightly_smiling_face: Am I missing another place where it makes a difference?

If indeed it doesn't make a difference, let's just delete the sorting part.

src/unread/UnreadCards.js
```diff
@@ -52,7 +52,12 @@ class UnreadCards extends PureComponent<Props> {
         key: 'private',
         data: [{ conversations, ...restProps }],
       },
-      ...unreadStreamsAndTopics,
+      ...unreadStreamsAndTopics.filter(item => item.isPinned),
+      {
+        key: 'separator',
```
borisyankov: You don't need to add a dummy section. There are several ways to rework that.
One is to create two sections, one is 'Pinned' the other is 'Rest'.
Also, we should consider showing the header titles.

jainkuniya: Then, we will loss the floating heading of `StreamItem`, if we go with `pinned` and `rest` as sections. Currently each stream unread are treated as different section, and one section for all PM unread.

So if we go with
```
SectionSeparatorComponent={SectionSeparatorBetween}
```
then separator will be shown after each stream.

 Or are you suggesting to have nested `SectionList`?.

Issue #2949, opened 2018-09-08
Unread cards: Unread PM/GM is not shown if there are no unread stream messages.


jainkuniya:
<img width="339" alt="image" src="https://user-images.githubusercontent.com/18511177/45259263-030f8980-b3e6-11e8-9276-cde6c83047cd.png">

PR #2950, opened 2018-09-08
unreadScreen: Fix no PM/GM were visible if no stream unread msg.

jainkuniya:
Consider `conversations.length` too in determining if there are unread
messages or not.

Fix: #2949 

image


src/unread/UnreadCards.js
```diff
@@ -107,5 +107,5 @@ export default connect((state: GlobalState) => ({
   conversations: getUnreadConversations(state),
   presences: getPresence(state),
   usersByEmail: getAllUsersByEmail(state),
-  unreadStreamsAndTopics: getUnreadStreamsAndTopicsSansMuted(state),
+  unreadStreamsAndTopics: [], // getUnreadStreamsAndTopicsSansMuted(state),
```
borisyankov: WIP?

jainkuniya: Oh! ya. Thanks @borisyankov 
I used it for testing purpose 😅 
My mistake forgot to remove it before commiting. Will take care of it next time 👍

PR #2951, opened 2018-09-09
eslint: Update spellchecker and fix issues

borisyankov:
The performance issues with the spellcheck plugins were resolved
https://github.com/aotaduy/eslint-plugin-spellcheck/issues/44

This commit updates the dependency to the latest version and fixes
several spelling issues.


gnprice:
Thanks @borisyankov for this update (and for getting upstream to fix the performance issue!)

There's a bunch of new words in the exclude list. It looks like some of them are from URLs. The maintainer on that upstream thread mentioned
> I think you might need to add some extra skipWords and
> - 'https://[^s]*'
> to be able to skip https urls

If you try that, does it let us leave out a lot of those new terms?

borisyankov:
Just to confirm that I did fix the issues and did rerun locally and tests are passing (obviously CI agrees)

gnprice:
Great!

Now let's take advantage of that `https` pattern to drop all of the words from the list that were only needed for such URLs. :slightly_smiling_face:

borisyankov:
@gnprice I did remove the words that are caught by the `https` regex.

PR #2952, opened 2018-09-10
Fix #2808 Sometimes topic list is not rendered even after fetching it from server.

### LINKED ISSUE #2808: Topics are not rendered even after they are fetched & stored in state.
screen shot 2018-07-14 at 11 28 11 am

In the screenshot, `INIT_TOPICS` event got fired but still topics screen is showing loading screen.

@zulipbot claim

jainkuniya:


src/utils/redux.js
```diff
@@ -33,4 +33,4 @@ export const isStateGoingBack = (cur: GlobalState, prev: GlobalState): boolean =
   || isNavTransitionGoingOn(cur, prev)
   || isEqual(cur, prev);
 
-export const connectPreserveOnBackOption = { areStatesEqual: isStateGoingBack };
+export const connectPreserveOnBackOption = () => ({ areStatesEqual: isStateGoingBack });
```
borisyankov: This is a significant regression that might cause performance issues (may or may not, good to investigate).

The first version creates a constant and it is reused. The new version produces a new object on each call.

borisyankov:
I think the first changes we made that took `isTransitioning` into account were sensible.
We are now in a territory where we add and add logic and need it all over the place.

Maybe it is good to step back and examine what are the best practices that React Navigation suggests in these cases?

src/utils/redux.js
jainkuniya: > The first version creates a constant and it is reused. The new version produces a new object on each call.

To overcome this we can make another version of `connectPreserveOnBackOption` which only depends on `isStateGoingBack` and not on `isNavTransitionGoingOn`

jainkuniya:
>Maybe it is good to step back and examine what are the best practices that React Navigation suggests in these cases?

Now in the next major release, they are stopping redux integration. So this will be automatically taken care.

https://reactnavigation.org/docs/en/redux-integration.html

>Warning: in the next major version of React Navigation, to be released in Fall 2018, we will no longer provide any information about how to integrate with Redux and it may cease to work.

PR #2957, opened 2018-09-13
topicListScreen: Remove `connectPreserveOnBackOption` from connect params.

jainkuniya:
Why this is removed? Because it was acting over smartly. This was
added to save redundant re-renders which were caused by nav state
change on pushing/popping route. But what if topics got fetched while
navigation is in a state of `isTransitioning`. It will ignore that
event and will not get re-rendered.

Also `topicListScreen` is not depended on the nav state, so even if

 ```
cur.nav.routes.length < prev.nav.routes.length
```

is removed then also there is no problem. As the prop supllied to
`topicListScreen` will not change and `shouldComponentUpdate` (as
`PureComonent` is used) will return false and save re-render.

Instead of using `connectPreserveOnBackOption`, we should make our
selector not to listen to nav state change (#2810). Which is a better
alternative for `connectPreserveOnBackOption`.

Fix: #2808

Next will go through all other instances where
`connectPreserveOnBackOption` is used and check whether it is really
required or not.

### DIFF
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -7,7 +7,6 @@ import type { Auth, Dispatch, GlobalState, Orientation, User, PresenceState } fr
 import { getAuth, getSession, getAccountDetailsUserFromEmail, getPresence } from '../selectors';
 import { Screen } from '../common';
 import AccountDetails from './AccountDetails';
-import { connectPreserveOnBackOption } from '../utils/redux';
 
 type Props = {
   auth: Auth,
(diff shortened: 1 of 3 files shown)


gnprice:
(This shouldn't be merged until #2810 is in, right? Because until then it triggers a bug.)

gnprice:
Hmm, but #2808 is a nastier bug. So I guess it's best not to block on that.

borisyankov:
As discussed here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-navigation.20v2
we should completely remove the `connectPreserveOnBackOption` function.

From my testing, we should remove it in the nav bars too, at least, which leaves only `AccountDetailsScreen` (we use that in 6 places currently)

I will first do a PR removing the `getCanGoBack` usage and then none of the selectors would be `nav` state dependent.

gnprice:

It seems we all agree we should remove that value completely; great. :smile: I summarized our recent discussions and conclusions in a jsdoc on it, pushed as feff4bd17.

PR #2958, opened 2018-09-13
Upgrade ESLint libs and plugins and fix linting issues

borisyankov:
1. Upgrade ESLint libs and plugins.
2. Disable several rules that don't make sense.
3. Fix other issues that do.


gnprice:
> Based on the 'spell-check' PR.

(i.e., #2951 )

borisyankov:
Rebased. (also #2951 was merged)

gnprice:
I get this error:
```
/home/greg/z/mobile/src/types.js
  485:45  error  Prefer commas to semicolons in object and class types  flowtype/object-type-delimiter
  486:36  error  Prefer commas to semicolons in object and class types  flowtype/object-type-delimiter
```

which is on this code:
```
export interface Dispatch {
  <A: Action | NavigateAction>(action: A): A;
  <T>((Dispatch, GetState) => T): T;
}
```

If I make those into commas, then on save `prettier-eslint` turns them back into semicolons. Would you sort that out one way or another? I don't have a view on which answer is preferable.

---

Separately: it's awkward that the first commit here causes a huge number of lint errors; that can make it harder to diagnose other issues later, among other things. Can we put the rule changes first, and the upgrade at the end?

If it isn't easy to 100% do that, let's at least do as much of it as we easily can. Then we can take a look at what the hard part is.

borisyankov:
Closing as severely outdated.

PR #2959, opened 2018-09-15
Use `getCanGoBack` selector in `ModalSearchNavBar`

borisyankov:
We don't need to use the `getNav` selector in `ModalSearchNavBar`
as we use it only to check if we can go back which is the purpose
of `getCanGoBack` selector.

This is good for maintainability but also helps with a potential
future removal of the Redux-React Native integration. This is the
last component that directly relies on the whole `nav` state.

The other is `AppWithNavigation` but it serves only for Redux
integration.


gnprice:
Thanks, merged!

I made just small tweaks in the commit message; the biggest was that it's *nearly* the "last component" that did this, as the next paragraph points out there is one more. :wink:

PR #2963, opened 2018-09-18
docs/background: Recommend reading for Redux Persist.

jackrzhang:
Two resources are:
- [this Medium article](https://medium.com/async-la/redux-persist-your-state-7ad346c4dd07), and 
- the [documentation](https://github.com/rt2zz/redux-persist#redux-persist) from markdown files in the `rt2zz/redux-persist` repository.


chrisbobbe:
@gnprice This is one of the oldest PRs, but it looks pretty quick to review as it's just a small (and I think helpful) addition to the docs. 🙂 

A possible counterpoint (that may or may not be worth mentioning in the docs) is a concern raised here about v5 (you can tell I've mostly been looking through old PRs this afternoon!): https://github.com/zulip/zulip-mobile/pull/2915

chrisbobbe:
Closing, as we've grown ever more skeptical about the value of redux-persist, and we might switch to something else. 🙂

PR #2964, opened 2018-09-18
docs: Troubleshoot the error message for uninstalled dependencies.

jackrzhang:
Corresponding topic from chat [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Module.20.60lodash.2Eunion.60.20does.20not.20exist.20in.20the.20Haste.20module.20map).

### DIFF
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -89,6 +89,20 @@ next person with a setup like yours.
 
 ## Troubleshooting
 
+### Bundling Failure: Unable to resolve module...
+
+When trying to bundle the application via Metro after pulling new code, you may
+see an error like this:
+
+```
+error: bundling failed: Error: Unable to resolve module `lodash.union` from `.../zulip-mobile/src/chat/chatReducers.js`: Module `lodash.union` does not exist in the Haste module map
+```
+
+New dependencies may have been added to package.json. In the case for the
+message above, `lodash.union` was added. You can install any new packages by
+running `yarn`.
+
+
 ### Build failure: `aapt` / "error while loading shared libraries"
 
 When trying to build the Android app, you may see an error like this one
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -89,18 +89,24 @@ next person with a setup like yours.
 
 ## Troubleshooting
 
-### Bundling Failure: Unable to resolve module...
+### Bundling failure: Unable to resolve module ...
 
-When trying to bundle the application via Metro after pulling new code, you may
-see an error like this:
+When running the app, you might see in the output of the Metro bundler
+-- aka "the JS server", or `react-native start` -- an error like this
+(reformatted for readability):
 
 ```
-error: bundling failed: Error: Unable to resolve module `lodash.union` from `.../zulip-mobile/src/chat/chatReducers.js`: Module `lodash.union` does not exist in the Haste module map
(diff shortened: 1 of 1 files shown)


docs/howto/build-run.md
```diff
@@ -76,6 +76,20 @@ next person with a setup like yours.
...
+When trying to bundle the application via Metro after pulling new code, you may
+see an error like this:
+
+```
+error: bundling failed: Error: Unable to resolve module `lodash.union` from `.../zulip-mobile/src/chat/chatReducers.js`: Module `lodash.union` does not exist in the Haste module map
+```
+
+New dependencies may have been added to package.json. In the case for the
+message above, `lodash.union` was added. You can install any new packages by
+running `yarn run reinstall`.
```
borisyankov: `yarn reinstall` is enough.

docs/howto/build-run.md
```diff
@@ -76,6 +76,20 @@ next person with a setup like yours.
 
 ## Troubleshooting
 
+### Bundling Failure: Unable to resolve module...
+
+When trying to bundle the application via Metro after pulling new code, you may
```
borisyankov: It is good that you give a more concrete example and solution, but that removes the more general description that encompasses more use-cases, as this `reinstall` is not only useful for the 'new dependency added' case.

jackrzhang: I think 'new or updated dependencies' would be more encompassing and explicit; I've changed it to say that.

In what other cases other than new or updated dependencies might `reinstall` be used?

I agree that mentioning `reinstall` at the end as a possible general catch-all would be a helpful tip for troubleshooting; I could put that mention at the end back in, and reword the commit and this PR to be specific to the case of pulling new or updated dependencies.

gnprice:
Thanks @jackrzhang !

In the situation you describe, I generally run just `yarn` (a shortcut for `yarn install`). That *ought* to resolve it, and AFAICT it always does for me; its job is to install new and updated dependencies.

It's also a lot faster, so I'm reluctant to recommend blowing away the whole `node_modules` and also Yarn cache when not required. :smile: Here's `yarn reinstall` from our packages.json, just to make explicit in this thread what it does (it isn't from upstream):
```
    "reinstall": "rm -rf node_modules; yarn cache clean; yarn install",
```

Can you elaborate more on cases where just `yarn` (aka `yarn install`) hasn't worked for you, and `yarn reinstall` has been helpful?

gnprice:
(Also @borisyankov I am curious to hear about the earlier experiences that led you to add it and document it!)

borisyankov:
I also use `yarn` all the time and it seems I did not have to use `yarn reinstall` recently.
It definitely got use in the past. I can't remember the exact situations, I think it was mostly when we include new components that have native parts...

I have also seen this recommended in many places, so we are not being crazy having it as an option :) Just an option we don't need to use often.

jackrzhang:
@gnprice You're right that the command for the use case I've described here would be just `yarn`. 😄 Which means I haven't had any use for `yarn reinstall` thus far, then.

In case it might be helpful just to have the specific reference to Metro's error message, I've updated the commit accordingly.

gnprice:
Thanks again @jackrzhang ! Just merged.

The version in your branch deleted the mention of `yarn reinstall`; there are tweaks we might make to that text but I don't see the need to delete it, so I undid that part (`git checkout -p @^`). Also added a commit of my own on top with a bit of editing; take a look.

Issue #2967, opened 2018-09-18
layout: Selected style is not applied even if user is selected for new group/invite.


jainkuniya:
Current:
<img width="358" alt="screen shot 2018-09-18 at 11 56 53 pm" src="https://user-images.githubusercontent.com/18511177/45708384-eff77900-bb9e-11e8-8ecd-e804a43b87f7.png">

Expected:
<img width="358" alt="image" src="https://user-images.githubusercontent.com/18511177/45708435-14ebec00-bb9f-11e8-90a5-4e610f4c9529.png">

@zulipbot claim

PR #2968, opened 2018-09-18
userList: Fix reg in getting user id for calculating `isSelected`.

jainkuniya:
`user` has id with `user_id` instead of `id`.

Fix: #2967


gnprice:
Thanks -- merged!

PR #2970, opened 2018-09-18
refactor: Create a common method that should be called on stream, people, emoji onAutocomplete. Eliminate new function creation for every item of list.

jainkuniya:
StreamItem calls `onPress` with the stream name as parameter, so fit
that paramter in the common function which can be used for all stream
item autocomplete. Before, for every item a new function was created
at the time of rendering.

Create a function inside render should be avoided.
(https://codeburst.io/6-simple-ways-to-speed-up-your-react-native-app-d5b775ab3f16)

This is a expriemental change, if it goes goods then we should do this
at other similiar places too.

### DIFF
--- a/src/autocomplete/PeopleAutocomplete.js
+++ b/src/autocomplete/PeopleAutocomplete.js
@@ -25,8 +25,16 @@ type Props = {
 class PeopleAutocomplete extends PureComponent<Props> {
   props: Props;
 
+  handleUserGroupItemAutocomplete = (name: string): void => {
+    this.props.onAutocomplete(`*${name}*`);
+  };
+
+  handleUserItemAutocomplete = ({ fullName }): void => {
+    this.props.onAutocomplete(`**${fullName}**`);
+  };
+
   render() {
-    const { filter, ownEmail, users, userGroups, onAutocomplete } = this.props;
+    const { filter, ownEmail, users, userGroups } = this.props;
     const filteredUserGroups = getAutocompleteUserGroupSuggestions(userGroups, filter);
     const filteredUsers: User[] = getAutocompleteSuggestion(users, filter, ownEmail);
 
@@ -42,7 +50,7 @@ class PeopleAutocomplete extends PureComponent<Props> {
             key={item.name}
             name={item.name}
             description={item.description}
-            onPress={() => onAutocomplete(`*${item.name}*`)}
+            onPress={this.handleUserGroupItemAutocomplete}
           />
         ),
       },
@@ -55,7 +63,7 @@ class PeopleAutocomplete extends PureComponent<Props> {
             avatarUrl={item.avatar_url}
             email={item.email}
             showEmail
-            onPress={() => onAutocomplete(`**${item.full_name}**`)}
+            onPress={this.handleUserItemAutocomplete}
           />
         ),
       },
--- a/src/autocomplete/StreamAutocomplete.js
+++ b/src/autocomplete/StreamAutocomplete.js
@@ -18,8 +18,12 @@ type Props = {
 class StreamAutocomplete extends PureComponent<Props> {
   props: Props;
 
+  onAutocomplete = (name: string): void => {
+    this.props.onAutocomplete(`**${name}**`);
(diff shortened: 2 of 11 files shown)


borisyankov:
Good catch and good solution.
This is the exact reason I've been doing this so frequently.

> Create a function inside render should be avoided.

That is true but not at all costs. Depends on how often a given object is rendered (one new function created per render).

In that case, since it is a loop of potentially lots of items, this is a very good to have fixed!

src/autocomplete/StreamAutocomplete.js
```diff
@@ -18,8 +18,10 @@ type Props = {
 class StreamAutocomplete extends PureComponent<Props> {
   props: Props;
 
+  onAutocomplete = (name: string): void => this.props.onAutocomplete(`**${name}**`);
```
borisyankov: In a different PR, I decided we should be moving away from this in favor of more explicit but correct code like:

```js
onAutocomplete = (name: string): void => {
  const onAutocomplete = this.props;
  onAutocomplete(`**${name}**`);
}
```
The first is a new eslint rule that warns on destructurng, and the other issue, we should not return the result of `onAutcomplete` even though it is `void` as it might be confusing, why are we returning it? :)

jainkuniya: @borisyankov Done 👍

src/user-picker/UserPickerCard.js
```diff
@@ -37,6 +37,10 @@ type State = {
   selected: User[],
 };
 
+type HandleUserPressParams = {
+  email: string,
+};
```
gnprice: Let's inline types like this one. Like so:
```
handleUserPress = (params: { email: string }) => {
```
Otherwise they create clutter up here, and also make the actual function definition harder to read because the type that basically defines the signature is dozens of lines away up here.

src/user-picker/UserPickerCard.js
```diff
@@ -57,8 +61,9 @@ class UserPickerCard extends PureComponent<Props, State> {
     }
   };
 
-  handleUserPress = (email: string) => {
+  handleUserPress = (params: HandleUserPressParams) => {
```
gnprice: Like I said above, this `HandleUserPressParams` type is better off just written in line here.

But actually even better in this case: don't even write it at all! This is a good place for type inference. We can just write:
```
handleUserPress = ({ email }) => {
```
and because we have a nice type signature on `UserList`, Flow sees that the type of this method's parameter must be `{ email: string, fullName: string }`.

When doing that, it is essential that Flow really does succeed at the inference :smile: . Check that by hovering, or hitting C-k C-i to get the hover popup.

src/autocomplete/PeopleAutocomplete.js
```diff
@@ -22,11 +22,26 @@ type Props = {
...
 
+  handleUserGroupItemAutocomplete = (name: string): void => {
+    const { onAutocomplete } = this.props;
+    onAutocomplete(`*${name}*`);
+  };
+
+  handleUserItemAutocomplete = (params: HandleUserItemAutocompleteParams): void => {
+    const { onAutocomplete } = this.props;
+    const { fullName } = params;
+    onAutocomplete(`**${fullName}**`);
```
gnprice: This gets a bit verbose. I think this method body would be better as
```
    this.props.onAutocomplete(`**${params.fullName}**`);
```
Destructuring to give names to the various pieces is a helpful thing to do when the code is longer and when things get repeated; but here we use `this.props` just once and `params` just once, so the whole thing just gets longer by adding these intermediate names. And without them it all fits very plainly on one line, which is great.

(Better yet, combine with the remark I made on a different file, for
```
  handleUserItemAutocomplete({ fullName }) {
    this.props.onAutocomplete(`**${fullName}**`);
  }
```
.)

gnprice:
Thanks again @jainkuniya, and thanks for the revisions! Merged.

I made a few small adjustments to the branch, mainly to reorder the commits so that logically related changes were together; take a look.

PR #2974, opened 2018-09-19
peopleAutocomplete: Use `user_id` as key for each item in list.

jainkuniya:
Two user might have same name, but can't have same `user_id`. So use
id as key for each item of the list which is unique.


gnprice:
Merged, thanks @jainkuniya !

I have the same question as Boris -- and as I think he pretty strongly suspected, the answer is "yes". Take a look at `git grep key=`. The emails in particular can be improved by using an id. That'd be a nice followup from this!

PR #2975, opened 2018-09-19
android: Fix label of intent filter.

jainkuniya:
What is this label?
> the one set by the parent component, is used when the component is
> presented to the user as having the capability described by the
> filter.

(https://developer.android.com/guide/topics/manifest/intent-filter-element#label)

Currently this intent filter is only used in the oAuth functionality.
So after selecting account in the browser, user is redirected to the
app with this intent. If multiple app are capable of taking this
intent (like in our case when both release and debug builds are
installed in the phone), then a pop up is shown by the phone OS,
letting user to select with which app he wants to procced. And this
label is used in this pop up. `filter_react_native` is totally
irrelevant, so change it to something relevant, which can be used
to differentiate both the apps in the phone.

### DIFF
--- a/android/app/src/debug/res/values/strings.xml
+++ b/android/app/src/debug/res/values/strings.xml
@@ -1,0 +1,3 @@
+<resources>
+    <string name="app_name">Zulip (debug)</string>
+</resources>
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -42,7 +42,7 @@
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
         </intent-filter>
-        <intent-filter android:label="filter_react_native">
+        <intent-filter>
             <action android:name="android.intent.action.VIEW" />
             <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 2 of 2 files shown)


jainkuniya:
On current master, it shows like this.
screenshot_20180920-002614

gnprice:
Ah, thanks for tracking this down! This behavior had been bugging me. :-) And thanks for the link straight to the relevant docs.

Looking at that bit of docs, I think this isn't quite the right fix.
> The default value is the label set by the parent component.

And in fact, we already set a nice label on the parent component! Namely `@string/app_name`, which resolves to "Zulip".

So how about we just delete this `android:label` attribute entirely? I think that will cause this to say "Zulip", which is pretty spot on.

A bonus would be to make it "Zulip (debug)" in the debug version. That isn't strictly necessary, because we already have the nice distinct icon. But if we did do it, we should do it at the top level -- change `@string/app_name`.

I think that should actually be very straightforward, based on [this doc](https://developer.android.com/studio/build/build-variants#sourceset-build): just create a file `android/app/src/debug/res/values/strings.xml`, and set just the `app_name` value in it. That ought to override the value from our main `strings.xml`.

gnprice:
Thanks @jainkuniya for this PR and that update! Just rebased and merged. :smile:

PR #2977, opened 2018-09-19
 compose: Fix image fileName from uri for HEIC image files.

### LINKED ISSUE #2178: Convert HEIC/HEIF images on upload from iOS
(Summary by @gnprice:)

Images created by recent versions of iOS are often in the HEIF format (extension `.heic`), which is a patent-restricted Apple format. When those images are uploaded, they tend to be unreadable elsewhere.

When uploading an image on iOS that's in HEIF format, we should use platform facilities to get or create a JPEG version and upload that.

(Original report follows.)

---

Hi,

I setup zulip today on a Ubuntu 16.04.4 LTS following the installation guide, so far everything looks promising. I found one issue so far or it looks like one at least.

Some pictures on my Iphone are JPG everything works, but some pictures are in HEIC format and here the latest IOS client only show a link like IMG_7113.HEIC no image, if i click the link a web view opens up loading but shows only a white page.

I see the request in the nginx logfile with a 200 status code.

xxxxxxx - - [30/Mar/2018:22:12:11 +0000] "GET /user_uploads/2/92/wPKjGi9juRThwf-r-0bW_HsO/IMG_7113.HEIC HTTP/1.1" 200 65247 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 11_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.0 Mobile/15E148 Safari/604.1"

I thought maybe for the web view it would help to add the mime types but that did not change anything.

/etc/nginx/zulip-include/uploads.types
    image/heic                            heic;
    image/heif                            heif;

Thx

Markus

jackrzhang:
Copying the jsdoc comment I've written for `fixFileNameFromUri`:

```js
/**
 * In 2017, Apple adopted a new standard image format, HEIF, which stands
 * for High Efficiency File Format; Apple photos are named with the '.heic'
 * extension. When an image file with the extension '.heic' is uploaded to the
 * server,the image becomes unviewable within a message list. Interestingly,
 * the `file` command line utility reports that images saved under this format
 * are JPEG files, so there's likely some pecularity with how the server handles
 * file uploads based on the extension.
 *
 * If an image file name has the '.heic' extension, and the uri of the image on a
 * mobile device has an extension for the JPEG format, fixFileNameFromUri replaces
 * that extension with '.jpeg'.
 */
```

This resolves #2178.
@gnprice

### DIFF
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -20,22 +20,29 @@ type Props = {
 };
 
 /**
- * Adjust the fileName to reflect the real file format, according to uri.
+ * Adjust `fileName` to one with the right extension for the file format.
  *
- * Photos in an iPhone's camera roll (taken since iOS 11) are typically in
- * HEIF format and have filenames with the extension `.HEIC`.  When the user
- * selects one of these photos through the image picker, the file gets
- * automatically converted to JPEG format... but the `fileName` property in
- * the react-native-image-picker response still has the `.HEIC` extension.
+ * Sometimes we get an image whose filename reflects one format (what it's
+ * stored as in the camera roll), but the actual image has been converted
+ * already to another format for interoperability.
  *
  * The Zulip server will infer the file format from the filename's
- * extension, so we need to adjust the extension to match the format. The
- * clue we get in the image-picker response is the extension found in `uri`.
+ * extension, so in this case we need to adjust the extension to match the
+ * actual format.  The clue we get in the image-picker response is the extension
+ * found in `uri`.
  */
-export const fixFileNameFromUri = (uri: string, fileName: string): string => {
+export const chooseUploadImageFilename = (uri: string, fileName: string): string => {
+  /*
+  * Photos in an iPhone's camera roll (taken since iOS 11) are typically in
+  * HEIF format and have filenames with the extension `.HEIC`.  When the user
+  * selects one of these photos through the image picker, the file gets
+  * automatically converted to JPEG format... but the `fileName` property in
+  * the react-native-image-picker response still has the `.HEIC` extension.
+  */
   if (/\.jpe?g$/i.test(uri)) {
     return fileName.replace(/\.heic$/i, '.jpeg');
   }
+
   return fileName;
 };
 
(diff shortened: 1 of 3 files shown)


src/compose/ComposeMenu.js
```diff
@@ -19,6 +19,26 @@ type Props = {
...
+ * server,the image becomes unviewable within a message list. Interestingly,
+ * the `file` command line utility reports that images saved under this format
+ * are JPEG files, so there's likely some pecularity with how the server handles
+ * file uploads based on the extension.
+ *
+ * If an image file name has the '.heic' extension, and the uri of the image on a
+ * mobile device has an extension for the JPEG format, fixFileNameFromUri replaces
+ * that extension with '.jpeg'.
+ */
+export const fixFileNameFromUri = (uri: string, fileName: string): string => {
```
borisyankov: The name of the function can be made better. 'Fix' is a bit generic. 'FileName' might not be exact (we are talking about extension only). Would also adding 'image' somewhere in the name help? (I know I am a bit vague, but trying not to suggest you a concrete name :) )

I would also move this to some of the `/utils/*.js` files.

src/compose/ComposeMenu.js
```diff
@@ -19,6 +19,26 @@ type Props = {
...
+  if (/\.jpe?g$/i.test(uri)) {
```
borisyankov: Should we do an `if` for `Platform.OS === 'ios'`?
This might never be `true` on Android, but this `if` will make the behavior more exact to our intentions.

src/compose/ComposeMenu.js
borisyankov: Yeah, keeping it in the same directory but outside the `ComposeMenu` makes sense. We don't use that approach often right now, but it is perfectly valid and we might want to use it more frequently. 👍

src/compose/ComposeMenu.js
jackrzhang: The function doesn't depend on anything particular to iOS to run, so such a check would be unnecessary. In the case that a user on Android attempts to upload a HEIC file (though, I'm not sure if that's even possible), the logic will be used.

src/compose/ComposeMenu.js
jackrzhang: I could have been more specific; I'm ok with keeping the function in the `ComposeMenu` file.

src/compose/ComposeMenu.js
jackrzhang: According to Wikipedia, [HEIF is supported by Android Pie](https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format#Support).

gnprice:

I added a change to give the helper a perhaps more meaningful name, and revise the jsdoc accordingly.

PR #2978, opened 2018-09-19
createStreamScreen: Add `keyboardShouldPersistTaps="always"` to Screen.

jainkuniya:
Let's take an example, user just pushed create stream screen, now as
the details/fields are empty, button is disabled and on clicking
button keybaord is dismissed. Which is totally unexpected.

Basically as the button is disabled, so onPress is undefined and View
is rendered in the Touchable. So this touch gets propogated to
ScrollView, which dismisses the keyboard.

This improves the user exprience.

### DIFF
--- a/src/streams/CreateStreamScreen.js
+++ b/src/streams/CreateStreamScreen.js
@@ -25,7 +25,7 @@ class CreateStreamScreen extends PureComponent<Props> {
 
   render() {
     return (
-      <Screen title="Create new stream" padding>
+      <Screen title="Create new stream" padding keyboardShouldPersistTaps="always">
         <EditStreamCard
           isNewStream
           initialValues={{ name: '', description: '', invite_only: false }}


borisyankov:
Good analysis. Maybe we can do something to not propagate the touch? Are we doing something wrong in our component?

jainkuniya:
> Are we doing something wrong in our component?

By component do you mean Touchable?

Then if we remove this if statement, then it will not propagate to parent 
```
if (!onPress && !onLongPress) {
```

borisyankov:
@jainkuniya we were considering the behavior you are proposing with @gnprice and we are not sure that this is an improvement in all cases. 

You are correct that if you tap the disabled button the keyboard is dismissed but that seems OK (not good, not bad).

Remember that you are working on #3140 which will add more buttons to the screen and very likely they will not fit on the screen. Then we will want to dismiss the keyboard more easily. 😄 Even worse is that on iOS there is no button to hide the keyboard as is on Android.

Considering that I think we should not do that change.

Issue #2979, opened 2018-09-19
layout: weird green lines in the lower-left corner in the compose box (between menu expand button and input).


jainkuniya:
First reported here https://chat.zulip.org/#narrow/stream/48-mobile/subject/testflight/near/466844

Got fixed in https://github.com/zulip/zulip-mobile/pull/1810/files

But again, in some refactoring it got introduced.

@zulipbot claim

PR #2980, opened 2018-09-19
Fix: Compose menu icons are visible even in unexpanded state.

jainkuniya:
Fix: #2979

### DIFF
--- a/src/styles/composeBoxStyles.js
+++ b/src/styles/composeBoxStyles.js
@@ -67,6 +67,9 @@ export default ({ color, backgroundColor, borderColor }: Props) => ({
   composeMenu: {
     flexDirection: 'row',
     overflow: 'hidden',
+    // this backgroundColor style is a test and try hack to fix #2979,
+    // so it's hard to explain the behaviour
+    backgroundColor: 'transparent',
   },
   expandButton: {
     padding: 12,


src/styles/composeBoxStyles.js
```diff
@@ -67,6 +67,7 @@ export default ({ color, backgroundColor, borderColor }: Props) => ({
   composeMenu: {
     flexDirection: 'row',
     overflow: 'hidden',
+    backgroundColor: 'transparent',
```
borisyankov: The style is very weird (as I mentioned in the last PR review :) )
I'd add a comment here so we don't accidentally remove it again.

jainkuniya: @borisyankov Added comment :) 👍

src/styles/composeBoxStyles.js
```diff
@@ -67,6 +67,9 @@ export default ({ color, backgroundColor, borderColor }: Props) => ({
   composeMenu: {
     flexDirection: 'row',
     overflow: 'hidden',
+    // this backgroundColor style is a test and try hack to fix #2979,
```
borisyankov: Maybe it is better to know why this happens and if this hack works reliably? :)

gnprice:
:smile: Based on the comment in the new version it sounds like this is not something we should merge.

I definitely would be interested in a fix for #2979, though! Seems like some more debugging is required. :wink:

PR #2983, opened 2018-09-21
Fix emoji reaction data so removing works (fix #2784).

### LINKED ISSUE #2784: Unable to remove emoji reaction when just added
(edited by @gnprice after discussion below)

Tapping an emoji reaction you've made should remove it. But:

* It doesn't work if I just added the reaction here (the most common case in practice, I'm pretty sure).
* Also doesn't work if I just added it through the webapp.
* It *does* work if I previously added it, and only then went and loaded these messages.

---
(original report:)

Tapping on a reaction to a message by you does not remove it if it is the only reaction of that particular type on a message.

### LINKED ISSUE #2926: Emoji Reaction Picker: Frozen state / Can't delete reaction
This happened to me once so I'm assuming its possible it could happen to other users too and is indeed a bug that needs to be squashed. While scrolling through reactions the lag ( issue #2925 ) caused me to accidentally select an emoji. After this point I tried clicking the emoji in the message but it would not let me remove the emoji for some reason. I tried clicking the emoji icon several times but it wouldn't go away. 

On 16.2.96 for iOS.

jainkuniya:


src/actionTypes.js
```diff
@@ -347,8 +347,10 @@ export type EventUpdateMessageAction = ServerEvent & {
 };
 
 export type EventReactionCommon = ServerEvent & {
+  emoji_code: string,
   emoji_name: string,
   message_id: number,
+  reaction_type: 'unicode_emoji' | 'realm_emoji' | 'zulip_extra_emoji',
```
borisyankov:
```js
type ReactionType = 'unicode_emoji' | 'realm_emoji' | 'zulip_extra_emoji';
...
reaction_type: ReactionType,
```

jainkuniya: oh I misses, thanks @borisyankov  👍

gnprice:
So looking at the code in this change, my first reaction is: this is (if it's correct) clearly fixing a type error!
```
       reactions: oldMessage.reactions.concat({
         emoji_name: action.emoji_name,
         user: action.user,
+        reaction_type: action.reaction_type,
+        emoji_code: action.emoji_code,
       }),
```

Specifically, assuming this change is right, that means this bit of code was inserting completely the wrong type of object here.

So if this is fixing a type error: why wasn't our type-checker able to tell us that before? And then let's fix that problem too. :grin: That's the root cause of this bug, and there may be other bugs lurking that it also shows us how to fix.

gnprice:
> So if this is fixing a type error: why wasn't our type-checker able to tell us that before?

Hmm, well, this looks to be the core of the issue:
```
export type Message = { /* ... */
  reactions: SlimEventReaction[], /* ... */
};

export type SlimEventReaction = {
  emoji_name: string,
  user: any,
};
```

So our `Message` type only says the objects in `reactions` have an `emoji_name` and a `user`, and doesn't mention anything else.

Let's sort out what the server's API really is here, and nail that down clearly in the types so we can be sure we're fixing all the relevant code.
* What does `reactions` look like in the two different ways we get messages from the server? Quoting our `Message` jsdoc:
```
 *  * `message: Message` on a server event of type `message`, and our
 *    `EVENT_NEW_MESSAGE` Redux action for the event;
 *  * `messages: Message[]` in a `/messages` (our `getMessages`) response,
 *    and our resulting `MESSAGE_FETCH_COMPLETE` Redux action;
```
* Then at a minimum let's write that down in comments; if getting it into the types will require some work refactoring things, we can merge first and follow up with those changes after. But there's clearly something erroneous here and I'd like to find out what's really happening first.
* Related question: how exactly does this change in the reducer propagate into a change to what we pass to the server? There is probably some code along the way there that also should have caught this error.

gnprice:
I went and merged the first two commits, with the nice `ReactionType` type -- thanks again!

One change I made to both of them:
* The very helpful descriptions in the commit messages were information about the types, which make sense for anyone looking at the types in the future to want to see. They're relevant to the new version of the code on its own, and not specific to the comparison between the new version and the old version.
  * So I moved them out of the commit messages and into jsdoc on the types instead, so they'll be easier for more people to run across in the future. :smile:

jainkuniya:
> `EVENT_NEW_MESSAGE`

This will definitely will not contain any reaction, as when this event is dispatched from server, no reactions were present. (As this event is dispatched from server as soon as there is new message).
In case if someone reacts immediately after posting message then there will be BATCH EVENTS of `EVENT_NEW_MESSAGE ` and `EVENT_REACTION_ADD`.

> messages fetched via `/messages`.
It will contain reactions array with type `Reaction`.

> `EVENT_ADD_REACTION`
It contains `Reaction` & `ServerEvent` & `{message_id}` i.e `EventReactionCommon`.

gnprice:
> This will definitely will not contain any reaction, as when this event is dispatched from server, no reactions were present. (As this event is dispatched from server as soon as there is new message).

Hmm, it's true that there won't be any reactions in this case. But that doesn't mean there won't be a `reactions` property -- it could just always be an empty array. Have you checked empirically what happens? (I.e., looked in the debugger at an `EVENT_NEW_MESSAGE` action, or printed the event.)

> > `EVENT_ADD_REACTION`
> 
> It contains `Reaction` & `ServerEvent` & `{message_id}` i.e `EventReactionCommon`.

Just to avoid any ambiguity: you're saying that this is what you observed empirically, looking at an actual event (or the resulting `EVENT_ADD_REACTION` action) in the debugger?

jainkuniya:
> Have you checked empirically what happens?

Yup! checked in debugger, it contains empty array `reactions: []`

> EVENT_ADD_REACTION action in debugger

image


>Cool, so it looks like your main conclusion is:

Yup!

> What we've been doing with apiTypes is trying to keep there all the types that are really part of the server's API. When a type is part of that API and also is used elsewhere, we can import it elsewhere. So the Reaction type is a good example of that -- we'll put it in apiTypes.

Doing now :)

gnprice:
> > Have you checked empirically what happens?
> 
> Yup! checked in debugger, it contains empty array `reactions: []`

Great. :) And that really is cleaner for the API (make the messages as similar as possible between events and fetch), so I'm glad that's the answer, too.

jainkuniya:
@gnprice Done!

Moved `Reaction` type to `apiTypes`.

gnprice:
Awkwardly, Flow *still* doesn't see the type error here. :-/ This time, AFAICT we're doing everything right; we're running into this bug in Flow: facebook/flow#6871 .

Also illuminating is facebook/flow#7120 . And there are a lot of other issues with these "indexer properties" in Flow, i.e. the `{ [S]: T }` syntax -- e.g. [this search](https://github.com/facebook/flow/issues?utf8=%E2%9C%93&q=is%3Aissue+indexer+property+Object) finds a number of issues that have a similar flavor, where Flow with indexer properties is unsound, even when you're doing nothing at all exotic and just make a very ordinary programming error of the kind that a working type-checker should catch.

Worse, perhaps: neither of those two issues I linked has a response from anyone at Facebook or on the Flow team. :cricket: And the one that's our issue exactly was filed two months ago.

gnprice:
And merged -- thanks again @jainkuniya !

PR #2984, opened 2018-09-21
Update react-native-photo-view dependency to resolve #2654.

### LINKED ISSUE #2654: Lightbox on iOS fails to show uploaded images
This is similar to #2618 , but affects an additional part of the UI, and only on iOS.

To reproduce:
* On iOS, log into an org on zulipchat.com, or possibly any other server running a recent release and on Xenial. (Not chat.zulip.org, which is on Trusty.)
* Find a conversation with a user-uploaded image in it. (Or just upload one.)
* Tap the image, bringing up the lightbox.
* The image doesn't appear. We just have the lightbox's black background, and its top and bottom bars.

I just tested this with 13.4.88 on iOS 10. With 13.5.89 (which differs only in one unrelated bugfix) on Android, it works fine.

Given that this happens on a Xenial but not a Trusty server, it's almost certainly related to our feature to require authentication to see uploaded files (which doesn't work on Trusty and is disabled there.) We're either not passing the authentication info, or not following the redirect, or both.

jackrzhang:
(Reminder to run `yarn reinstall` after updating package.json)

@gnprice

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "react-native-image-picker": "^0.26.10",
     "react-native-notifications": "^1.2.0",
     "react-native-orientation": "^3.1.3",
-    "react-native-photo-view": "^1.5.2",
+    "react-native-photo-view": "alwx/react-native-photo-view#c58fd6b30",
     "react-native-safari-view": "2.0.0",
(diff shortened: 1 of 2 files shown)


gnprice:

PR #2989, opened 2018-09-22
webView: Update `generatedEs3.js` with current `js.js`.

jainkuniya:
Generate via running `yarn run generate-webview-js` in the terminal.


borisyankov:
Looks good. In our RN 0.56 PR when splitting, I forgot to include this, so we merged `js.js` version without the compiled part.

I am surprised that you need to include `instanceof` in the spellcheck though. Do you have to?

gnprice:
Thanks @jainkuniya ! I just pushed a couple of commits to make the same fix: 4e1d60f52, and then I noticed this PR, realized there was a lint issue :smile:, and pushed ee610bdba to fix that.

I fixed the lint issue by turning off the spellcheck rule entirely on that file; see the commit.

Issue #2988, opened 2018-09-22
webView: Messages are not mark as read if there are few messages in the list, i.e list is not scrollable.


jainkuniya:
Steps to reproduce:
* Narrow to such a topic/stream/pm, where messages are very less (2-3).
* Now as the list is not completely covering screen, it is not scrollable and there is no way to mark messages as read except manually using mark as read button.
* Ideally it should mark those messages as read.

I have worked on this issue and submitted #2579, will be updating that PR/submitting new PR next.

@zulipbot claim

gnprice:
If I understand your two PRs #2990 and #2991 correctly, your diagnosis is that the reason this is happening is that
* we are in fact calling `sendScrollMessage`, but
* `getStartAndEndNodes` is failing to find a message at the end, so we're sending `endMessageId: 0`.

Is that right?

Let's focus the PRs on that, then: make `getStartAndEndNodes` work reliably.

Your #2988 approaches one side of this, which is that this line looking for the last message might be probing empty space at the end of the list:
```
  const endNode = getMessageNode(document.elementFromPoint(200, window.innerHeight - 20));
```
I think the issue goes deeper than that, though. For example, as I look at this closely, I see that even when we don't get empty space off the end of the list, we'll get things wrong if we happen to hit a date separator, or a recipient bar. Here's a sample of the webview as seen via `chrome://inspect`:
![image](https://user-images.githubusercontent.com/28173/45985862-142edc00-c01d-11e8-8b8a-f31befccdf76.png)

That `timerow` div is 36px high. As a hack to avoid completely missing that there are messages there, we've given it a `data-msg-id` attribute that corresponds to the message after it... but that means that if the user scrolls down so that the `timerow` is just halfway visible, we'll mark *the following message* as read -- a message which the user has not seen a single pixel of. So that is a pretty unfortunate workaround.

Here's what I think would be a better approach:
* Find the first, and the last, *child of `documentBody`* that's visible. This is almost like what we're doing with `document.elementFromPoint` and `getMessageNode`, but if we don't find anything then use `firstChild` and `lastChild` to get the node. This solves the "empty at the end" problem.
* Instead of smearing out `data-msg-id` onto neighboring `timerow` and `header` divs, use `nextSibling` to turn "first visible child" into "first visible message", and similarly `previousSibling`/"last".
* That should do it.

Want to try that?

jainkuniya:
> Is that right?

Yup! 👍 

[...]

PR #2990, opened 2018-09-22
webView: Messages are not mark as read if there are few messages in the list, i.e list is not scrollable. Fix: #2988

jainkuniya:
Approach 1


gnprice:
Looking more closely at this, I have what I think will be a better plan to solve this problem reliably; see #2988 . Closing this PR in favor of discussion there / a new PR.

PR #2991, opened 2018-09-22
webView: Recursively find last message visible on screen. Fix: #2988

jainkuniya:
Before: Cordinate (200, window.innerHeight - 20) was used to find last
visible message on the screen (with current scroll position) and it's
id was passed to find messages which needs to be mark as read. But
what if list is small and there is no message node at
`window.innerHeight - 20` as y cordinate. Then `0` as message id was
passed, which is not making any sense (as now start msg id > end msg
id, see `msg.id >= fromId && msg.id <= toId` in
`filterUnreadMessagesInRange`).

After: Now recursively find last visible message node on screen with
current scroll position, i.e first find message node at (200,
window.innerHeight - 20), then (200, window.innerHeight - 40), then
(200, window.innerHeight - 60) and so on until y cordinate becomes
negative or message node is found.

Fix: #2988

Approach 2

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -122,17 +122,25 @@ window.addEventListener('resize', function (event) {
   height = documentBody.clientHeight;
 });
 
-var isMessageNode = function isMessageNode(node) {
-  return !!node && node.getAttribute && node.hasAttribute('data-msg-id');
+var getLastVisibleMsgIdOnScreen = function getLastVisibleMsgIdOnScreen() {
+  var x = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 200;
+  var y = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : window.innerHeight - 20;
+
+  if (x < 0 || y < 0) {
+    return 0;
+  }
+  var endNode = getMessageNode(document.elementFromPoint(x, y));
+  var msgId = getMessageIdFromNode(endNode, 0);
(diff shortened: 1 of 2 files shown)


gnprice:
I tried this on a narrow with one message in it, and got this red error banner:
image

> Uncaught TypeError: msgNode.getAttribute is not a function

This isn't a good place to use recursion; if we want to do this logic, it should be a loop. Recursion in this style depends critically on so-called "tail call optimization", which is a feature of a language implementation; and apparently the Chrome folks have [decided it's a bad idea](https://bugs.chromium.org/p/v8/issues/detail?id=4698), and their JS implementation isn't going to do it. Few [other implementations](https://kangax.github.io/compat-table/es6/#test-proper_tail_calls_(tail_call_optimisation)) have either.

I'm not really a fan of this "subtract 20 pixels" kind of logic; but apart from the comments above, this seems like an improvement on the existing code, so I'd be glad to take it once those are fixed.

gnprice:
Looking more closely at this, I have what I think will be a better plan to solve this problem reliably; see #2988 . Closing this PR in favor of discussion there / a new PR.

jainkuniya:
>This isn't a good place to use recursion; if we want to do this logic, it should be a loop. Recursion in this style depends critically on so-called "tail call optimization", which is a feature of a language implementation; and apparently the Chrome folks have decided it's a bad idea, and their JS implementation isn't going to do it. Few other implementations have either.

@gnprice Thanks for the info :)

PR #2995, opened 2018-09-24
CSS: Change style of disabled ZulipButton.

### LINKED ISSUE #2976: layout: Improve styles of ZulipButton.
Currently when Button is disabled, and on default theme, button text is not visible clearly. It's too light. Although, button is disabled, but I expected to know what is written on this component.

armaanahluwalia:
Tweaks the styling of the ZulipButton in its disabled state.
Can be tested on create a stream and when logging in.

Screenshot: 

Zulip Button Disabled Style

Fixes #2976

### DIFF
--- a/src/common/ZulipButton.js
+++ b/src/common/ZulipButton.js
@@ -58,7 +58,10 @@ const styles = StyleSheet.create({
     color: BRAND_COLOR,
   },
   disabled: {
-    opacity: 0.25,
+    backgroundColor: '#999',
+  },
+  disabledText: {
+    color: '#ddd',
   },
 });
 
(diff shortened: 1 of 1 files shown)


borisyankov:
Thanks for working on this. Also thanks for the video, it makes it way easier to review.

I would have also mentioned why we need this. Maybe you liked this better, maybe you think it does not look standard (true), maybe you think the text is not very readable (also true).

We try to follow Material Design when it is applicable. For example the guidelines for buttons are here:
https://material.io/develop/web/components/buttons/

You will find that they use:
```css
background-color: rgba(0, 0, 0, 0.12)
color: rgba(0, 0, 0, 0.37)
```

Which we can use verbatim. (this is light background with darker text, vs your PR which is the opposite)

gnprice:
Thanks @armaanahluwalia for sending this!

Following @borisyankov 's review above, I believe he's sent another approach to this in #3143. Let's follow up there.

PR #2998, opened 2018-09-25
Update is:starred narrow on EVENT_UPDATE_MESSAGE_FLAGS.

### LINKED ISSUE #2677: Newly starred messages do not appear in the starred list.

jackrzhang:
Fixes #2677.

In the screenshot below, #2676 is still unresolved.

<p align="center">
fix-2677-update-starred-list
</p>

I would label this with "bug" and "area: redux".

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -124,7 +124,8 @@ rules:
       const, wildcard, reactotron, camelcase, hydrator, perf, stacktrace, Dir, fs,
       avoider, octicons, centerer, ldap, gravatar, identicon, blueimp, filename, wildcards,
       jpeg, jpg, tif, mov, notif, diag, bmp, viewport, scalable, polyfill, rect, touchstart,
-      touchend, touchmove, remoteuser, sso, submessages, nbsp, args, px, na, heic]
+      touchend, touchmove, remoteuser, sso, submessages, nbsp, args, px, na, heic,
+      unstarred]
 overrides:
   - files:
       - generatedEs3.js
--- a/src/chat/chatReducers.js
+++ b/src/chat/chatReducers.js
@@ -7,6 +7,7 @@ import type {
   MessageFetchCompleteAction,
   EventNewMessageAction,
   EventMessageDeleteAction,
+  EventUpdateMessageFlagsAction,
 } from '../types';
 import {
   APP_REFRESH,
@@ -16,9 +17,10 @@ import {
   MESSAGE_FETCH_COMPLETE,
   EVENT_NEW_MESSAGE,
   EVENT_MESSAGE_DELETE,
+  EVENT_UPDATE_MESSAGE_FLAGS,
 } from '../actionConstants';
 import { LAST_MESSAGE_ANCHOR, FIRST_UNREAD_ANCHOR } from '../constants';
-import { isMessageInNarrow } from '../utils/narrow';
+import { isMessageInNarrow, STARRED_NARROW_STR } from '../utils/narrow';
 import { NULL_OBJECT } from '../nullObjects';
 
 const initialState: NarrowsState = NULL_OBJECT;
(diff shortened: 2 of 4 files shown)


.eslintrc.yaml
```diff
@@ -124,7 +124,8 @@ rules:
       const, wildcard, reactotron, camelcase, hydrator, perf, stacktrace, Dir, fs,
       avoider, octicons, centerer, ldap, gravatar, identicon, blueimp, filename, wildcards,
       jpeg, jpg, tif, mov, notif, diag, bmp, viewport, scalable, polyfill, rect, touchstart,
-      touchend, touchmove, remoteuser, sso, submessages, nbsp, args, px, na, heic]
+      touchend, touchmove, remoteuser, sso, submessages, nbsp, args, px, na, heic,
+      unstarred,]
```
gnprice: No trailing comma when `]` on same line.

src/chat/chatReducers.js
```diff
@@ -70,6 +72,29 @@ const eventMessageDelete = (
...
+      });
+    } else {
+      // operation === 'remove'
+      sources.push({
+        [STARRED_NARROW_STR]: state[STARRED_NARROW_STR].filter(id => !messages.includes(id)),
+      });
+    }
+  }
+
+  return Object.assign({}, ...sources);
```
gnprice: This will return a non-identical `state` even when no update happened -- e.g. when we just learned the user read some messages, which is very common.

I think this is another case where just systematically following the architecture's preferred pattern, avoiding updates unless there really is an update, doesn't appreciably complicate the code and saves complexity overall in understanding how things work.

Specifically, how about `const updates = [];`, and then `return Object.assign({}, state, ...updates);`... with an `if (!updates.length) { return state; }` just before it.

src/chat/chatReducers.js
```diff
@@ -70,6 +72,29 @@ const eventMessageDelete = (
...
 };
 
+const eventUpdateMessageFlags = (
+  state: NarrowsState,
+  action: EventUpdateMessageFlagsAction,
+): NarrowsState => {
+  const { messages, flag, operation } = action;
+  const sources = [state];
+
+  if (state[STARRED_NARROW_STR] && flag === 'starred') {
```
gnprice: Nit: I'd prefer `flag === 'starred'` first. That gives this the shape of an overall `switch/case` on `flag`, which is the appropriate shape for this to have in general.  The `state[STARRED_NARROW_STR]` is logic specific to the case when we're talking about the `starred` flag at all.

src/chat/chatReducers.js
```diff
@@ -70,6 +72,29 @@ const eventMessageDelete = (
...
+    if (operation === 'add') {
+      sources.push({
+        [STARRED_NARROW_STR]: [...state[STARRED_NARROW_STR], ...messages].sort(),
+      });
+    } else {
+      // operation === 'remove'
+      sources.push({
+        [STARRED_NARROW_STR]: state[STARRED_NARROW_STR].filter(id => !messages.includes(id)),
```
gnprice: Probably rarely a big deal in practice: but this is quadratic, in the number of already-known starred messages and the number we're learning were unstarred.

How about `const messagesSet = new Set(messages);` just above, and then use `messagesSet`?

.eslintrc.yaml
jackrzhang: Updated.

src/chat/chatReducers.js
jackrzhang: Updated.

src/chat/chatReducers.js
jackrzhang: Updated.

src/chat/chatReducers.js
jackrzhang: Updated.

gnprice:
Thanks again -- merged!

PR #3001, opened 2018-09-26
Update max iOS version to run on

borisyankov:
With the new iOS and device releases, update the 'max' version
of iPhone (newest, largest) to be 'iPhone XS Max'.


gnprice:
Merged, thanks!

I edited the commit message, because this subject line had scared me for a moment -- "max iOS version to run on" sounds like we didn't previously run on, say, iOS 12. Which is not what this turns out to be about. :slightly_smiling_face:

PR #3002, opened 2018-09-27
docs/push-notifications: Add missing step for setting bouncer url.

jackrzhang:
I split the changes into two commits for a cleaner diff.

### DIFF
--- a/docs/howto/push-notifications.md
+++ b/docs/howto/push-notifications.md
@@ -41,6 +41,11 @@ additional steps:
    python manage.py register_server --agree_to_terms_of_service
    ```
 
+   Registration should be one-time; that is, as long as you're using the
+   same `zproject/dev-secrets.conf` that was generated at provision time, you
+   shouldn't have to run the command again unless you build a new dev
+   environment from scratch.
+
    This is a variation of our [instructions for production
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks!

These steps are one-time setup, right? That is, once you do them once, you shouldn't have to do them again unless you build a new dev environment from scratch.

(Or more precisely, I think: as long as you're using the same `zproject/dev-secrets.conf` that was generated at provision time.)

These instructions sound a lot like this is something you have to do every time you want to run `tools/run-dev.py`, thanks to the "Before you run the server with `tools/run-dev.py`," bit. I think that impression is stronger after this change.

If these are one-time setup steps, want to try revising to make that clear? The "Now, follow ..." and thereafter are for *each* time, so some kind of separation is probably in order.

jackrzhang:
Will do!

jackrzhang:
Updated! @gnprice

jackrzhang:
Updated!

jackrzhang:
Hm, rereading this more, I think there are few more small tweaks that I want to make here. I'll post another comment when those are pushed :)

jackrzhang:
@gnprice This can be reviewed.

gnprice:
Thanks @jackrzhang ! Belatedly merged. :slightly_smiling_face:

Issue #2971, opened 2018-09-18
Cover selectors with types, using `Selector`


gnprice:
A large chunk of our remaining code that isn't covered with Flow types (#2052) is because most of our selectors don't have appropriate types.

Most of them, it seems like Flow ought to have the information to infer (see #2724 for details). That's likely a problem in Flow, but we can make it not be *our* problem; we're probably best off writing down explicit types on most of the selectors anyway, for maximum readability.

I worked out a type alias `Selector` recently that takes care of the complexity in the selector interfaces, and makes this very straightforward: `Selector<T>` means a selector that returns a `T`. Let's use that.

Here's an example:
```
export const getStreamFromId = (streamId: string): Selector<Stream> =>
  createSelector([getStreams], (streams, params) => {
```

gnprice:
I believe this is now complete! :tada: Thanks to all the contributors who helped out.

PR #3005, opened 2018-09-27
Altered types in NotificationCommon & EventReaction

flowkraD:
Issue in focus: #2052 
Type of content_truncated which was previously defined as `string` was changed to `boolean` as commented out next to it which proved to be iccurate.
Changed type of EventReaction.user from any to string.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -65,7 +65,7 @@ export type EventReaction = {
   emoji_code: string,
   emoji_name: string,
   reaction_type: string,
-  user: any,
+  user: string,
 };
 
 /** An aggregate of all the reactions with one emoji to one message. */
@@ -553,7 +553,7 @@ export type UnreadStream = {
 export type NotificationCommon = {
   alert: string,
   content: string,
-  content_truncated: string, // boolean
+  content_truncated: boolean,
   'google.message_id': string,
   'google.sent_time': number,
   'google.ttl': number,


src/types.js
```diff
@@ -553,7 +553,7 @@ export type UnreadStream = {
 export type NotificationCommon = {
   alert: string,
   content: string,
-  content_truncated: string, // boolean
+  content_truncated: boolean,
```
borisyankov: Are you sure this is correct?
I remember declaring it like that few months ago, because the incoming data was `string`, but we wanted it boolean.

flowkraD: I did a little digging before this change. I checked where `NotificationCommon` is being referred at and I found it's referenced in `NotificationPrivate` which is again referenced in `Notification`. Then I found that `Notification` is referred to in `config.js` to define the type of Config. Below (line 32), in config,js `undefined` was assigned to `notification`.
I headed on over to `notification.js` where the `notification` was assigned with a value `notificationData.getData()` and `notificationData` is defined as an `Object` whose contents were unknown to me. The TravisCI test showed no errors and i assumed that it was correct. I am not 100% sure that incoming data is actually `boolean`.

borisyankov: That would be a mistake on the Zulip's server payload, but I believe we didn't fix that. @jackrzhang might know something about that?

gnprice: The server code sure handles this field as a boolean: search for `content_truncated`, and it comes from `truncate_content` which returns a boolean.

There are several layers between that Python code in the Zulip server and this JS code, though. It wouldn't be a big surprise if one of them turns, say, the JSON value `true` into a string `'true'`.

Note that this is a field we never reference in our code: grep for `content_truncated`. So (a) running tests doesn't give much information about whether an annotation here is right, and (b) this isn't one of the higher-priority things for us to nail down the correct type for.

I've just gone and updated #2052 with current stats and a bit of guidance on finding the highest-value areas to add types now:
> #2971 is a great source of high-value places to add types. Running `yarn run test:flow-coverage` remains another great source.

Want to try those?

gnprice:
Thanks @flowkraD for looking at this!

I think the `EventReaction#user` annotation is not right; see e.g. `aggregateReactions.js`, where we pull `.user.email` off of one of these.

As I mentioned just now in an inline thread above, though, I think these two aren't among the higher-priority things for us to nail down the correct type for. So I'd actually prefer that instead of trying to sort out the right annotations for these, we set them aside for now (so I'll close this PR) and instead go after some of the higher-value annotations to add.

I've just gone and updated #2052 with current stats and a bit of guidance on finding the highest-value areas to add types now:
> #2971 is a great source of high-value places to add types. Running `yarn run test:flow-coverage` remains another great source.

Want to try those?

flowkraD:
I'll surely give it a try. I am a bit tight on time (still in freshman year of college) but I'll contribute whenever time permits. Thanks for bringing #2971 to my notice (^_^)

PR #3006, opened 2018-09-27
Add `IDEWorkspaceChecks.plist`

borisyankov:
Xcode 9.3 adds a new IDEWorkspaceChecks.plist file to a workspace’s
shared data, to store the state of necessary workspace checks.
Committing this file to source control will prevent unnecessary
rerunning of those checks for each user opening the workspace.

Quote from here:
https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-DontLinkElementID_7


gnprice:
Ah, thanks for tracking that explanation down. :slightly_smiling_face: Merged!

PR #3008, opened 2018-09-28
Refactor chatSelectors, chatReducers to narrowSelectors and narrowReducers respectively.

### LINKED ISSUE #2992: refactor code: Rename `chatReducer` to `narrowReducer`.
Currently we have `chatReducers` and  `messageReducers`. Firstly I got confused, why they are different, mostly both deals with message events. But then it became clear that: `messageReducer` is responsible for `message` state in global state and `chatReducer` is responsible for `narrow` in the global state. Thus then I was wondering we should rename `chatReducer` to something else (may be `narrowReducer`). As chat is very common name, chat itself can have many components, it doesn't convey any specific.

https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/refactoring.20code/near/645396

@zulipbot claim

jainkuniya:
Fix #2992


borisyankov:
I think we are using pluralization inconsistently `messageReducers` vs `flagsReducers`.

jainkuniya:
I guess singular will be better.

thoughts @gnprice ?

jackrzhang:
On an adjacent thought - I've opened #3016.

gnprice:
Thanks all -- merged!

Let's continue the plural/singular discussion on #3016 .

gnprice:
Hmm, I guess that's specifically about a different plural/singular question ("reducers" vs "reducer").

I agree with @jackrzhang that we should have the prefix in the module's name correspond to the name in the state. There are a few places we don't currently respect that:
```
accounts -> accountReducers
narrows -> narrowReducers
messages -> messageReducers
```
So we should just rename all those modules to match the state.

PR #3009, opened 2018-09-28
flow: Define Navigation state `params` properly

borisyankov:
`params` is an object and can contain anything. It is exacly as
Component's `props` but for navigation data.

It can contain:
 * narrow
 * email
 * streamId
 * serverSettings
 * realm
and more in future.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -293,9 +293,7 @@ export type NavigationState = {
     title: string,
     routeName: string,
     /** The fields in `params` vary by route; see `navActions.js`. */
-    params: {
-      narrow?: Narrow,
-    },
+    params: Object,
   }>,
 };
 


gnprice:
Here's the diff:
```
     /** The fields in `params` vary by route; see `navActions.js`. */
-    params: {
-      narrow?: Narrow,
-    },
+    params: Object,
```

So, what the current version says is:
* this is an object;
* there may be all kinds of properties on it (it's not a `{| ... |}` type, but plain `{ ... }`);
* but if there happens to be a property named `narrow`, it has type `Narrow`.

All of that is true at present; there's only ever a `narrow` when we went through `navigateToChat`, and that should indeed be of type `Narrow`.

For it to stay true depends on us not reusing the name `narrow` with a different type on some other route. I'm pretty OK with that assumption; even if we didn't have a type like this written down, I'd be pretty reluctant to do that because it seems quite confusing and easily avoidable. Unlike Components:
* there are only a few of these, and
* they're all defined right next to each other in the same file.

So this change seems like it's just giving up information that's useful; we lose coverage on that line in `getTopMostNarrow`.

borisyankov:
This change makes the type **correct**. It is an obviously too-simple to be useful, but changes an incorrect type to correct type, while in the future we might add more concrete type.

For example, `AccountDetailsScreen` gets an object for `params` like this:
```js
params: {
  email: string,
}
```

This is absolutely incompatible with:
```js
params: {
  narrow?: Narrow,
}
```

as this object does not contain an `email` property.

The correct type in that sense would be:
```js
params: { narrow: Narrow } | Object
```

gnprice:
Take a look at the discussion of "exact object types" in the Flow docs:
https://flow.org/en/docs/types/objects/#toc-exact-object-types

The type we have is correct. As I said above, it says
* this is an object;
* there may be all kinds of properties on it (it's not a `{| ... |}` type, but plain `{ ... }`);
* but if there happens to be a property named `narrow`, it has type `Narrow`.

All of which is true.

If we'd written `{| narrow?: Narrow |}` (an exact object type), then `{ email: string }` would not match. But we didn't, because that would be incorrect. :-)

Issue #2982, opened 2018-09-20
[Android] layout: Styles of account item is not consistent across Android & iOS.


jainkuniya:
on iOS
<img width="368" alt="image" src="https://user-images.githubusercontent.com/18511177/45840466-a04ab600-bd34-11e8-89e5-fbc0c880b70f.png">

on Android

![](https://chat.zulip.org/user_uploads/2/57/uyn-zqh6Nr531e3mW_uDSEZi/Screenshot_20180921-002434.jpg)

Notice the space between different account cards.

borisyankov:
Yeah, that is part of the annoying Android 8.x layout issues.

borisyankov:
Try my RN 0.57 PR and see if this issue is not magically fixed by two versions of React Native. 
Also would be a good signal if there are more issues to work on.

gnprice:
After #3013 , is this fixed now?

borisyankov:
Yes.

PR #3012, opened 2018-09-30
ui: Fix Account List spacing on Android

borisyankov:
This fixes the 'no spacing between account list items' issue. #2982

No visual changes to the iOS version, but Android now also matches it.

Also, more importantly, it makes the styling more clean and logical.

Now:
 * the AccountItem does not have a 'margin' by default, that didn't
make sense as it is not its job to provide that styling
 * the spacing is logically implemented via `ItemSeparatorComponent`

### DIFF
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -8,7 +8,6 @@ import { IconDone, IconTrash } from '../common/Icons';
 
 const styles = StyleSheet.create({
   wrapper: {
-    marginBottom: 8,
     justifyContent: 'space-between',
   },
   accountItem: {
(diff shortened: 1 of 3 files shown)


gnprice:
After #3013 , is this still needed or is it now redundant? Want to rebase it on that? It's possible the commit message is the only thing that changes.

(This issue isn't a priority compared to some others, I think.)

borisyankov:
It is not redundant, but also not critical. Rebased.

gnprice:
Hmm, I'm puzzled:
```
    ui: Fix Account List spacing on Android
    
    This fixes the 'no spacing between account list items' issue.
    No visual changes to the iOS version, but Android now also matches it.
```

Didn't #3013 make the spacing on Android match that on iOS?

That is what I meant when I suggested the commit message might change even if nothing else does. :slightly_smiling_face:

borisyankov:
I am picking this commit in the 'Deuglify UI' PR and closing this.

PR #3013, opened 2018-10-01
fix: Resolve layout issues with Touchable on Android

borisyankov:
`TouchableNativeFeedback` that we use on Android does not receive
any `style` props, that is quite inconsistent with `TouchableHighlight`
which we use on iOS.

Trying to set style props on a Touchable component does not have an
effect on Android, thus layout issues were introduced for Android-only in:
* AccountItem
* PasswordInput
* ZulipButton

### DIFF
--- a/src/common/Touchable.js
+++ b/src/common/Touchable.js
@@ -62,12 +62,11 @@ export default class Touchable extends PureComponent<Props> {
     return (
       <TouchableNativeFeedback
         accessibilityLabel={accessibilityLabel}
-        style={style}
         background={androidBackground}
         onPress={onPress}
         onLongPress={onLongPress}
       >
-        <View>{children}</View>
+        <View style={style}>{children}</View>
       </TouchableNativeFeedback>
     );
   }


gnprice:
What's an example of the layout issues this addresses?

It sounds like this is meant to change some behavior in the app. I'd like to have some idea of what the change is :) (and for anyone looking at the commit to be able to get some idea of what the change is.)

This may also be the kind of PR where before/after screenshots, for at least one of the widgets that changes, would be helpful.

borisyankov:
I did not provide a screenshot, mostly because it is a clear fix => several components look different than on iOS because the `style` we provide the component does not get applied at all :)

gnprice:
OK -- I'm still curious to hear what one of these changes actually is, though.

borisyankov:
The `show` button in the PasswordInput is on the next line on Android, while correctly on the same.

gnprice:
Got it. And on `AccountItem` we're missing the margin, which means there's no space between the options on the account switcher.

For `ZulipButton`, I don't see the difference. Are those styles doing something I'm missing?

We also pass a `style` prop to `Touchable` in one more place, namely `ImageAvatar`. I don't see an effect there either, so I have the same question as for `ZulipButton`.

borisyankov:
For `ZulipButton` it seems to be a bit unpredictable where we will encounter the issue, as we only apply this style to the Touchable:
```
  touchTarget: {
    flex: 1,
    alignSelf: 'stretch',
    justifyContent: 'center',
    alignItems: 'center',
  },
```

These may or may not have an effect, depending on the surrounding components and container.

gnprice:
Merged, thanks!

PR #3014, opened 2018-10-01
fix: WebView updated too frequently

borisyankov:
Revert a change made earlier in `webViewHandleUpdates`.
More details in #6bdb259

### DIFF
--- a/src/webview/__tests__/webViewHandleUpdates-test.js
+++ b/src/webview/__tests__/webViewHandleUpdates-test.js
@@ -82,6 +82,26 @@ describe('getInputMessages', () => {
     expect(messages[0].type).toEqual('content');
   });
 
+  test('when the rendered messages are equal deeply, do not return a new "content" message', () => {
+    const prevProps = {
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @borisyankov , especially for the detailed discussion in that commit message!

As I mentioned [in chat](https://chat.zulip.org/#narrow/stream/48-mobile/subject/Excessive.20Rendering.20Bug/near/649028) (after you were offline for the evening), this reverts more than the shallow/deep-equals change -- it also reverts the fix to #2676, re-opening that bug.

I've just pushed 9fc769802 to revert specifically the shallow/deep-equals change. See that commit message for some further analysis of what's going on here.

Based on what I learned that's described in that message (and at greater length in the same chat thread linked above), I think this also doesn't go far enough: the `fetching` and `typingUsers` props, which we switched to `===` in the same commit, suffer from the same issue and also need a revert. The impact of those naturally is smaller than of the content, but still good to avoid sending unnecessary `postMessage` messages.

PR #3019, opened 2018-10-03
Fix messages are not mark as read if list is short. Make `getStartAndEndNodes` more reliable. Fix #2988

### LINKED ISSUE #2988: webView: Messages are not mark as read if there are few messages in the list, i.e list is not scrollable.

jainkuniya:
Screencast
https://cl.ly/579af5465bb9
https://cl.ly/8b85cfee1fe7

Fix #2988

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -87,6 +87,9 @@ var getDocLevelNode = function getDocLevelNode(node) {
 var getMessageIdFromNode = function getMessageIdFromNode(node) {
   var defaultValue = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : -1;
 
+  if (!node) {
+    return defaultValue;
+  }
   var msgNode = getDocLevelNode(node);
   return msgNode && msgNode instanceof Element ? +msgNode.getAttribute('data-msg-id') : defaultValue;
 };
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @jainkuniya for taking a shot at this!

This logic is pretty convoluted, unfortunately. I think it shouldn't have to be appreciably more complicated than the existing logic it replaces.

I have a feeling that at this point the quickest way to communicate what I have in mind will be for me to just write it up myself, and you to read that; so I'm going to close this PR, and plan to do that soon.

A couple of specific things that jump out at me:
* I don't know what a "doc-level node" is. If anything it sounds like the whole document, whereas this function usually returns a child of the body.
* The recursion should be a loop; we discussed this on another PR recently.
* The use of `node[property]` to get either `node.previousSibling` or `node.nextSibling` is overly magical.

borisyankov:
I was going to comment on the original PR.
Since you are working on this code, you might want to do that change:
`getMessageNode` called directly is not needed in most (or all) cases now that we have `getMessageIdFromNode` (which of course uses that function)

So rework an callers of `getMessageNode` that then call `getMessageIdFromNode` to call the second function directly.

PR #3024, opened 2018-10-04
layout: Align option icon in Lightbox screen

### LINKED ISSUE #3003: layout: Improve style and position of option icon (3 dot icon) in Lightbox screen.
* Currently it is not aligned properly.

* Find better alternative for icon.

Related discussion https://chat.zulip.org/#narrow/stream/48-mobile/subject/UI/near/646695

AdityaHirapara:
Icon was not properly aligned due to overlay of unreadCount.

Fixes #3003.
screenshot_2018-10-04-17-40-08-244_com zulipmobile debug


borisyankov:
Cool, thanks for working on this. Your analysis of what is wrong is spot on.

I have found that using `absolute` positioning and magic numbers is rarely the right solution.

In this case, I don't think we need to use `NavButton` at all. We can directly use the `Icon` component. Would you try that?

AdityaHirapara:
@borisyankov, I 've made requested changes, Please review.

borisyankov:
Nice. Going in the right direction.
You didn't notice a particularly surprising thing about the `<Icon />` component... that it is 'touch-enabled' by default. So you don't need to add the Touchable at all. Just use the `onPress` of `Icon`.

AdityaHirapara:
I noticed that. Actually, I used Touchable here to display touch feedback. Should I remove that?

AdityaHirapara:
> Nice. Going in the right direction.
> You didn't notice a particularly surprising thing about the `<Icon />` component... that it is 'touch-enabled' by default. So you don't need to add the Touchable at all. Just use the `onPress` of `Icon`.

@borisyankov Done! Please review

PR #3027, opened 2018-10-08
Clean animatedScaleComponent and decide consistently whether to show unread notice or not.

jainkuniya:


src/chat/UnreadNotice.js
```diff
@@ -50,22 +50,24 @@ class UnreadNotice extends PureComponent<Props> {
 
   render() {
     const { limited, narrow, unreadCount } = this.props;
-
+    const shouldShow = unreadCount > 0;
```
borisyankov: I don't think this is adding anything to the code. If it was a complicated piece of code, sure.

src/chat/UnreadNotice.js
```diff
@@ -50,22 +50,24 @@ class UnreadNotice extends PureComponent<Props> {
...
-            <RawLabel
-              style={styles.unreadText}
-              text={limited ? unreadToLimitedCount(unreadCount) : unreadCount.toString()}
-            />
-            <Label
-              style={styles.unreadText}
-              text={unreadCount === 1 ? 'unread message' : 'unread messages'}
-            />
+      <AnimatedScaleComponent visible={shouldShow}>
+        {shouldShow && (
```
borisyankov: Why do we have this check inside every `AnimatedScaleComponent` children? We should move it to `AnimatedScaleComponent.js` instead.

gnprice:
Thanks @jainkuniya !

Have you tried out this branch with each of the UIs it affects?

I took a quick look:

* The FAB on the user picker (if you go to start a group PM conversation) now disappears abruptly, rather than shrink smoothly.
  * The smooth shrinking feels better to me. So perhaps this doesn't belong as a universal rule for this shared component.

* Looking at `UnreadNotice`: the effect is that when the messages are read, the notice now just disappears abruptly with no animation.
  * That doesn't feel like quite what we want. It feels pretty abrupt.
  * The status quo, with the white space and then a jump, feels more wrong. So maybe this strategy can be an improvement there. But, see above, it probably shouldn't be universal.

* Looking at the compose autocompleters, this makes it grow with an animation, but it disappears abruptly. Is that intentional? I'm not sure it's what we want.

(Curious also what thoughts @borisyankov has.)

jainkuniya:
Closing this, as this is now in stale state.

Issue #3022, opened 2018-10-03
Add event queue info and current polling status in `diagnostics` tab.


jainkuniya:
Currently if event queue breaks, or polling stops then there is no way to identity and confirm it except by human interpretation. Add some logging in the `diagnostics` tabs so that is becomes easier to confirm such cases.

Related chat: https://chat.zulip.org/#narrow/stream/243-mobile-team/subject/Event.20queue.20bug/near/649554

@zulipbot claim

borisyankov:
Expose the whole `session` state key.

PR #3028, opened 2018-10-09
 diagnostics: Add event queue details screen. Fixes: #3022

jainkuniya:
Fixes: #3022 

image


borisyankov:
We absolutely need to improve the 'debuggability' of the app... but I don't think doing it this way is effective or optimal.

There are two big issues for me in this approach:
 * a lot of noise added to the logs by adding `EVENT_NEW_EVENT_QUEUE_EVENT` on every server event!
 * a lot of custom code (and even strings to translate) but resulting in very little added information

Another issue - this being so custom, also makes it incompatible with the (soon coming, almost ready) two queue approach... I am also not convinced that the added data is helpful in debugging the exact issue we are having about a dead queue.

That is why I suggested this in the issue you had filed: 'export the whole session state'. That would be way shorter code-wise, would be compatible with any future changes, and would not add noise to our state management.

When debugging and gathering diagnostics data we need to be as broad as possible. We have no idea what is needed. Also, we need to keep this out of the state. 

Have a look at the `timing.js` diagnostics, to see how a much simpler approach is just as effective.

In order to debug the queue issues (and possibly most other issues), I suggest we implement the following two things:
 * a more general state display - so the session state but also any other key we want
 * an exception/error display - that is way more important when debugging the queue and possibly a lot of other issues
 * custom logging display - we can sprinkle some logs in places we want, maybe things like 'the queue was dead so I restarted', etc.

PR #3029, opened 2018-10-09
Fix: Incorrectly calculated unread mentions

borisyankov:
We are calculating the unread mentions to be used in the UI but
are currently not used (should do that soon)

Due to not exposing this we have not caught an issue - the logic to
determine if the current user was mentioned in an upcomming  message
was incorrect. In fact, there is no `is_mentioned` key on the `message`
object, but the data is passed via a `flags` array.

This logic is already used correctly in `eventMiddleware.js`.

This PR replaces the incorrect code and fixes the tests.

### DIFF
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -75,9 +75,14 @@ export type Message = {
   sender_domain: string,
 
   /**
-   * This is (always?) present in a `message` event; but we leave it out of
-   * the `messages` subtree of the Redux state, moving the information to
-   * the separate `flags` subtree.
+   * The `flags` story is a bit complicated:
+   *  * Absent in the `message` property of a `message` event... but instead
+   *    (confusingly) a `.flags` appears directly on the event.
+   *  * Present in the `EVENT_NEW_MESSAGE` Redux action for such an event.
+   *  * Present in the server's response to a `/messages` request (our
+   *    `getMessages`), and our `MESSAGE_FETCH_COMPLETE` action.
+   *  * Absent in the `messages` subtree of the Redux state; we move the
+   *    information to the separate `flags` subtree.
    */
   flags?: string[],
 
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -45,6 +45,13 @@ export type MessageEdit = {
  * complex; this is naturally a place where a large fraction of all the
  * features of Zulip have to appear.
  *
+ * Major appearances of this type include
+ *  * `message: Message` on a server event of type `message`, and our
+ *    `EVENT_NEW_MESSAGE` Redux action for the event;
+ *  * `messages: Message[]` in a `/messages` (our `getMessages`) response,
+ *    and our resulting `MESSAGE_FETCH_COMPLETE` Redux action;
+ *  * `messages: { [id]: Message }` in our global Redux state.
+ *
  * References include:
  *  * the two example events at https://zulipchat.com/api/get-events-from-queue
  *  * `process_message_event` in zerver/tornado/event_queue.py; the call
@@ -65,8 +72,8 @@ export type Message = {
   isOutbox: false,
 
   /**
-   * These don't appear in `message` events, but they appear in GET
-   * requests for messages in a narrow when a search is involved.
+   * These don't appear in `message` events, but they appear in a `/message`
+   * response when a search is involved.
    */
   match_content?: string,
   match_subject?: string,
@@ -76,13 +83,13 @@ export type Message = {
 
   /**
    * The `flags` story is a bit complicated:
-   *  * Absent in the `message` property of a `message` event... but instead
-   *    (confusingly) a `.flags` appears directly on the event.
-   *  * Present in the `EVENT_NEW_MESSAGE` Redux action for such an event.
-   *  * Present in the server's response to a `/messages` request (our
-   *    `getMessages`), and our `MESSAGE_FETCH_COMPLETE` action.
-   *  * Absent in the `messages` subtree of the Redux state; we move the
-   *    information to the separate `flags` subtree.
+   *  * Absent in `event.message` for a `message` event... but instead
+   *    (confusingly) there is an `event.flags`.
+   *  * Present under `EVENT_NEW_MESSAGE`.
+   *  * Present under `MESSAGE_FETCH_COMPLETE`, and in the server `/message`
+   *    response that action is based on.
+   *  * Absent in the Redux `state.messages`; we move the information to a
+   *    separate subtree `state.flags`.
    */
   flags?: string[],
 
--- a/src/events/__tests__/eventMiddleware-test.js
+++ b/src/events/__tests__/eventMiddleware-test.js
@@ -1,0 +1,28 @@
+/* @flow */
+import deepFreeze from 'deep-freeze';
+
+import eventMiddleware from '../eventMiddleware';
+
+describe('eventMiddleware', () => {
+  test('if `event.flags` key exist, move it to `event.message.flags`', () => {
+    const state = { session: {} };
+    const event = {
+      type: 'message',
+      flags: ['mentioned'],
+      message: {},
+    };
+    eventMiddleware(state, event);
+
+    expect(event.flags).not.toBeDefined();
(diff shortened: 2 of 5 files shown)


gnprice:
Hmm fascinating -- so one question this raises for me is: why is there also a `flags` property on the message itself?

After all, this diff itself says
```
 export type EventNewMessageAction = ServerEvent & {
   type: typeof EVENT_NEW_MESSAGE,
+  flags: string[],
   message: Message,
// [... others edited for brevity ...]
```
and `Message` has
```
  /**
   * This is (always?) present in a `message` event; but we leave it out of
   * the `messages` subtree of the Redux state, moving the information to
   * the separate `flags` subtree.
   */
  flags?: string[],
```
and in fact we have some code in `messageReducers.js` that dutifully scrubs that `flags` property of `action.message`, and on the other hand `flagsReducers.js` goes and consults `action.message.flags` to get its information.

The question behind that question is of course: is there *really* also a `flags` property on the message itself, or have we just been wrong about that the whole time?

gnprice:
After fruitlessly searching the code, then testing a couple of ways, I'm pretty sure the answer is: there is no `message.flags` on these events. Which means our `flags` data is just always blank for messages we learn about in a narrow we're caught up in.

borisyankov:
It is `flags` passed through the `event` itself.
So `event.flags` and not `event.message.flags`.

borisyankov:
That is likely to keep the `message` object the same for all users.

gnprice:
Not sure what you mean by "it".

Did you read the comment I quoted?
```
  /**
   * This is (always?) present in a `message` event; but we leave it out of
   * the `messages` subtree of the Redux state, moving the information to
   * the separate `flags` subtree.
   */
  flags?: string[],
```

The only reason that's there in the type is because it's supposedly in the `message` on the event.

gnprice:
(Note from an audio discussion with Boris):

On the other hand, our `MESSAGE_FETCH_COMPLETE` actions do have a `flags` property on each message. Plus, that's really the natural place for it in the first place. So I think the right direction to fix this is:
* The changes in this commit, plus:
* Leave `flags?` on `Message`, but fix the comment to explain the correct situation.
* Fix the code in `messageReducers`, `flagsReducers`, and possibly another place or two that is expecting to find this information in `action.message.flags`.

borisyankov:
Added a commit intercepting the event and 'fixing' it.

gnprice:
> `// fixfixfix`

This is not very helpful. :-) Is this a TODO? What should be "fixed" -- what would we rather be doing? Basically, when I look at this comment I don't know what I'm supposed to take from it / do about it. (That's actually true for me right now even though the context of this thread is fresh in my head -- it'll be more so for future readers, including future-me.)

This fix takes a different direction from what I described specifics for above. That's fine but it needs to take care of the appropriate details for this direction:
* The comment on `flags?` in `Message` is still wrong and needs to be fixed.
* Lower in this same `eventMiddleware`, we refer to `event.flags`; should be `event.message.flags`.
* Are there other places that look at `event.flags`?
* We should have the flags in just one place in the Redux action we're creating; if we're putting it on `event.message.flags`, then we should remove `event.flags`.

borisyankov:
So... the `// fixfixfix` line is weird because I don't remember writing it (though I guess I did) and serves no purpose. Removed it.

borisyankov:
Oh, I remembered what the `// fixfixfix` was supposed to be - I wanted to add a comment about the code added. Now did that.

borisyankov:
Also reworked to **move** the key and not just duplicate it.
Not sure if it is better than just duplicating, but after some thought, also not sure it is worse.
So just did it.

gnprice:
Thanks for the updates, and the point-by-point replies! This is really helpful.

> The comment on flags? in Message

So, here's how our Message type is documented:
```
/**
 * A Zulip message.
 *
 * This type is mainly intended to represent the data the server sends as
 * the `message` property of an event of type `message`.  Caveat lector: we
 * pass these around to a lot of places, and do a lot of further munging, so
 * this type may not quite represent that.  Any differences should
 * definitely be commented, and perhaps refactored.
...
 */
export type Message = { // ...
  /**
   * This is (always?) present in a `message` event; but we leave it out of
   * the `messages` subtree of the Redux state, moving the information to
   * the separate `flags` subtree.
   */
  flags?: string[],
```

So it's describing the data the server sends in that spot (the `message` property of an event of type `message`.) Where properties appear in this type without further discussion, we're saying they appear in the data the server sends.

And for `flags`, the comment specifically says we believe it appears there. We discovered in this thread that in fact it does not. That's why I say the comment is wrong.

---

Other than that, the first commit looks good! On the rest, just small comments:
* The second commit adds tests for exactly the functionality the first commit added; let's squash it into the first commit.
* In the third commit, the message seems unchanged. Given what we learned in this thread, I don't think it's accurate; it needs to be updated.

borisyankov:
I don't think we need to necessarily link any change to the type comment with this PR being merged.

The change in `eventMiddleware` makes the data more in line with what we expect, not less.

We (I?) knew it even before this PR that the server does not send a `flags` key even for a `/message` `GET` - `flags?` indicate that it is optional. The `flags` key was missing for an empty flags array, which may or may not be a good idea (having `flags: []` is definitely more orthodox)

borisyankov:
Updated:
* squished the first two commits, merging their commit message
* updated the second commit's message

gnprice:
> We (I?) knew it even before this PR that the server does not send a `flags` key even for a `/message` `GET` - `flags?` indicate that it is optional.

Well, that's not what the comment says -- it says it *is* there for events. Which it's not.

Also, what you "knew" is not true. :wink: The messages in a fetch result (which is from `/messages`) do have a `flags` key; that's in fact the only way we *ever* currently get flags. It's the reason one of those three messages in that screenshot does have the @-mention highlighting.

Which makes a very nice demonstration of why it's important that when we learn the answers to these questions, we write them down clearly.

(As does the bug this PR is fixing in the first place.)

> The `flags` key was missing for an empty flags array

Hmm, why do you say that?

I just went and empirically checked, by watching the Chrome devtools' network tab on the webapp. The server sends `flags: []` when there are no flags set, just like we'd both hope.

I'm interested in hearing why you say it was missing, because that suggests either (a) some exceptional case where it is indeed missing, which would be important to know, or (b) a source that's mistaken, which we should correct.

gnprice:
Your fix commit says:

> In a previous commit we 'normalize' the incoming data, moving it to the
> `message` object and now we can treat any message the same, no matter if
> it got to the app via a **GET** `/messages` API call or as a new message
> via the event system.

We don't actually handle `/messages` results here, though, right? Those generate `MESSAGE_FETCH_COMPLETE` actions, which I don't see a handler for in this reducer.

gnprice:
Merged -- thanks again @borisyankov ! I added a commit of my own to correct that jsdoc, as well as another to revise the `Message` jsdoc a bit more.

I had a couple of questions in my previous two comments -- those are independent of merging this PR, and I'm still interested in the answers.

borisyankov:
> > The flags key was missing for an empty flags array

> Hmm, why do you say that?

I did miss the fact that we are `omit()`-ting the `flags` key now.
Which means we should go ahead and omit even more keys that we don't use and split the keys into `ApiMessage` and `InernalMessage` (or just `Message`)
We are not using keys like: `content_type`, `client`, etc.

gnprice:
Ah, I see -- meaning you were looking at the message in the Redux state and saw that key wasn't there? That makes sense.

> Which means we should go ahead and omit even more keys that we don't use and split the keys into `ApiMessage` and `InernalMessage` (or just `Message`)
> We are not using keys like: `content_type`, `client`, etc.

Sure! I think separating the API's "message" type from our stored "message" type would be a good further improvement. For fields we don't use: I might be inclined to keep a few where it's at least clear what they mean and how/why we probably want to use them in the future, but there are a lot of them that I don't understand (and we don't use) or that I think I understand and they seem obsolete.

I do think keeping `flags` around was higher-risk, and removing it was higher-priority, than those, precisely because we *do* use the flags. Keeping them around meant keeping a stale copy of desirable information, which was a bug waiting to happen. But all these fields that we don't use are potential sources of confusion too.

Issue #2997, opened 2018-09-24
Autocomplete popup is huge and doesn't scroll


gnprice:
When I type into the compose box something like `:s`, or `@g`, there are a lot of possible completions. That's fine; we should display a reasonable scrolling view of them.

Instead, as of 17.1.98 (and some recent versions before it), the autocomplete popup seems to extend way beyond the top of the screen, and doesn't scroll. This looks broken, and makes it impossible to select any but the last few options.

gnprice:
#2919 is one PR aiming to address this. I think #2944 may also be intended to do something related. (@jainkuniya , is that right?)

timabbott:
Yeah, the mentions issue here is quite annoying and makes it impossible to mention some users.

gnprice:
@armaanahluwalia reports an additional, fascinating wrinkle:

> The autocomplete seems to stop at the bottom of the list... if i try to scroll up it doesnt seem to work unless i keep one finger on the screen and use another finger to scroll further up which reveals more emojis above which are inaccessible

I don't reproduce this on my device (17.1.98 on Android 9); IIRC Armaan uses iOS, so this may be OS-dependent. For me, it's just stuck at the bottom of the list and I'm unable to scroll at all, even by experimenting with multiple fingers.

This wrinkle isn't really an independent or additional problem relative to the issue described in the OP, but might be a helpful clue in resolving it.

armaanahluwalia:
Yes I’m on iOS 17.1.98. Let me know when fixed so I can retest.

gnprice:
Updating back here with discussion from #2919: fixing this properly will be tricky (details on that thread), and we'll come back and do that after we lay some groundwork, namely #3017 and the work that it in turn depends on.

In the meantime we'll do a quick fix, like a fixed max height. @jackrzhang is planning to do that.

gnprice:
This was fixed (back in October) by #3030. The height isn't always right, but it's never enormous, which is a big improvement.

PR #3030, opened 2018-10-09
autocomplete popup: Fix maximum height.

jackrzhang:
This is a quick fix to #2997. Set the height to 250 logical pixels, which is high enough to display about 5 list items.

Before:

autocomplete-popup-quick-fix-before

After:

autocomplete-popup-quick-fix-after


gnprice:

Issue #3032, opened 2018-10-09
travis: Build android build.


jainkuniya:
It's better to build Android build on Travis. This will helps us in tracking builds, and let us know when somethings fails (on some code change or dependency change).

I am able to build it. Will submit PR soon. https://travis-ci.org/jainkuniya/zulip-mobile/builds/439519182
@zulipbot claim

PR #3034, opened 2018-10-10
travis: Build Android build. Fixes: #3032

jainkuniya:
This will help in tracking down the Android build, and make native
Android development faster.

Now:

* Main language on travis is JAVA.

* node v8 is installed via nvm.

* all the Android components which are used by zulip-mobile and it's
dependencies are installed.

Fixes: #3032

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,44 @@
-language: node_js
+language: android
 
-node_js:
-  - 8
+jdk: oraclejdk8
+
+env:
+  global:
+    - ANDROID_HOME=${TRAVIS_BUILD_DIR}/android-sdk
+    - PATH=${ANDROID_HOME}/:${ANDROID_HOME}/tools/:${ANDROID_HOME}/platform-tools/:${PATH}
+
 
 cache:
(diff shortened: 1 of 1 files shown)


borisyankov:
I like that. I am also pleasantly surprised that it adds just a few minutes to the build time.
I am wondering can we do several builds per PR or something? Because Kunal's work on the 2e2 testing (that is totally on hold) depended on xcode-based and we might want to keep it working too, in the future (the .travis-e2e.yml)

gnprice:
Good idea, thanks!

I just rebased and pushed to the PR; I pushed a number of updates to our Android build config recently, so I want to see if Travis will be happy after these changes or if further updates are required.

gnprice:
Looks like some updates are needed. Want to try working out how to make those?

I'm also interested in figuring out how I would work out those updates. It isn't at all obvious to me what they should be... and I definitely wouldn't want to be iterating on a cycle where each cycle required going through Travis for an 8-minute build.

gnprice:
Thanks -- merged!

PR #3036, opened 2018-10-10
ios chinese input bug fix

### LINKED ISSUE #3033: I can not input chinese on ios
I can not input chinese on ios

x007007007:
#3033 bug fix

### DIFF
--- a/src/common/Input.js
+++ b/src/common/Input.js
@@ -1,6 +1,6 @@
 /* @flow */
 import React, { PureComponent } from 'react';
-import { TextInput } from 'react-native';
+import { Platform, TextInput } from 'react-native';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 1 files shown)


borisyankov:
Thanks for working on this issue.
We have another possible solution for this problem (making the component uncontrolled), but I am not 100% certain that it is solving the issue you are having.

To make the commit easier to review, would you do the following three changes:
 * run `yarn prettier` on this commit so the formatting is consistent with our established code style. This will make the diff smaller (few lines have unnecessary spaces) and easier to read
 * make sure the tests pass (currently Flow is complaining about missing props)
 * Add few lines to the commit message explaining what the issue you encounter is, and why your changes fix it.

x007007007:
#2434 will fix this bug

Issue #3035, opened 2018-10-10
travis: Check if generatedEs3.js is consistent with latest js.js or not.


jainkuniya:
`generatedEs3.js` is used by webView. We write js in the `js.js` and use babel to convert it to `generatedEs3.js`, so that it is compatible with mobile phones webView. But many times it is noticed that `js.js` is updated in the commit, but we forgot to update `generatedEs3.js`. Which will result in old js been executing, and we are thinking that new js is executing. To avoid such situations, there should be some check mechanism, which can warn such situations and maintain consistency between `generatedEs3.js` and `js.js`

@zulipbot claim

borisyankov:
A potentially better option is to generate the file itself automatically when the project starts (depends on whether this is now supported the Metro bundler).

jainkuniya:
Yup! in future it's better option.
But for now I guess this is sufficient and necessary test, at least for now. We can definitely improve in future :).

PR #3037, opened 2018-10-10
travis: Add verify generatedEs3 js script.

jainkuniya:
This script will help in maintaining consistency between js.js and
generatedEs3.js file.

Fixes: #3035

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ install:
 
 script:
   - npm run test:full
+  - tools/verify-webview-js
   - cd android && ./gradlew assembleDebug
 
 env:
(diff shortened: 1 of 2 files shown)


borisyankov:
If you use the Node script itself you can add something like `node verify-webview-js` and keep it cross-platform compatible.

gnprice:
Thanks!
* Miscellaneous developer-facing tools like this go in `tools/`.
* No need to `chmod +x` in setup; Git tracks that just fine, you just have to do it.
* No need for an `exit 0`.
* No need to say `bash foo` when `foo` is executable and has a shebang (`#!`) line; just `./foo`.
* Scripts that are to be executed as commands get names like commands; the language is an implementation detail that doesn't belong in the name. So `tools/verify-webview-js`.
* I'd be happy for this to be in JS (as Boris suggests), but I don't have a strong preference. This isn't something it's essential for most developers to run; really it only runs in CI, and when testing changes to this script itself.

gnprice:
Thanks @jainkuniya -- merged!

PR #3038, opened 2018-10-10
docs: Remove support for iOS 8.0.

jackrzhang:
XCode (I'm using 9.4.1 at the time of this commit) doesn't provide a simulator for iOS 8.0, and the vast majority of our iOS users stay updated to the latest OS, so the impact of this should be trivial.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -1,6 +1,6 @@
 # Developer guide
 
-We target operating systems >= Android 4.4 (API 19) and >= iOS 8.1.
+We target operating systems >= Android 4.4 (API 19) and >= iOS 9.0.
 
 ## Why React Native?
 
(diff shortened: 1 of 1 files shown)


borisyankov:
We should drop it to 9.x

https://everyi.com/by-capability/maximum-supported-ios-version-for-ipod-iphone-ipad.html

Even old devices support 9.x and we have 0 users on 8.x

gnprice:
Thanks @jackrzhang ! Merged, with an extra commit to bump further to 9.0.

In fact it turns out that we already did that in the actual metadata in Xcode, which controls what devices will see the app in the Play Store; so, good to update our docs so we don't have to rediscover the appropriate minimum version a third time :slightly_smiling_face:

PR #3039, opened 2018-10-10
flow: Remove handleImagePress from Lightbox props.

jackrzhang:
There is no `handleImagePress` prop that is passed to Lightbox. At the time of this commit, `handleImagePress` is a prototype method that updates local state.


gnprice:
Merged, thanks!

PR #3040, opened 2018-10-11
webview js: New algorithm for identifying messages in view

### LINKED ISSUE #2988: webView: Messages are not mark as read if there are few messages in the list, i.e list is not scrollable.

gnprice:
TODO items before merge:
* I think `Document.prototype.elementsFromPoint` (note plural) exists in an iOS 9 webview... but I'm not highly confident of that, so I'd like some manual testing that everything seems to work. @jackrzhang began work on this.
* There's a spot where I need an assert, and haven't sorted out how to express that; suggestions welcome. (@borisyankov ?)
* The last commit isn't really finished; I think it's not quite right in some cases.

The most important part of this branch is the first commit, which fixes #2988 ; and I'm pretty confident in it now apart from those first two points above. So I hope to get those resolved and get it merged soon.

---

The algorithm we have in `getStartAndEndNodes` for finding what
messages are visible on the screen (and therefore we should mark as
read) has some problems.  Among them:

 * Most egregiously, if there are no messages at the bottom of the
   viewport, we fail to find anything at all!  This happens in
   particular when you narrow to a conversation that only has a few
   messages and fits easily on the screen.  This is #2988.

 * When the bottom of the screen has a date separator, after which
   offscreen is a recipient bar, after which further offscreen is a
   message... we'll treat seeing that date separator as if it meant
   seeing the message.

 * When the *top* of the screen has a floating recipient bar, we'll
   treat that as seeing the message immediately after the recipient
   bar, as well as everything since then... which may extend many
   screenfuls into the past.

Fix all of those with an algorithm reworked more or less from scratch,
adapted from the algorithm used in the webapp.

 * Start from one visible message, and work up and down to the edges
   of the screen.  The webapp does this with the selected message (the
   one in the blue box); we need to work a bit harder, which is
   `someVisibleMessage` and its helper `midMessagePeer`.  Fixes #2988.

 * Because we're walking up and down as needed rather than counting on
   finding a message at a single fixed point, we can stop using the
   hack of finding `data-msg-id` on things that aren't messages, like
   date separators; we just pass over such things to focus on messages.

 * To get at an element laid out in sequence with messages, and be
   able to look past something drawn over them like a floating
   recipient bar, we use `elementsFromPoint` to get the whole sequence
   of elements behind the one returned by `elementFromPoint`.


jackrzhang:
I've tested this out on simulators running iOS 9.0 and iOS 9.3 - `Document.prototype.elementsFromPoint` doesn't exist in an iOS 9 webview:

<p align="center">
visible-messages-ios-9 0
<p/>

Here's the error message, which is a bit difficult to discern above:
image

Another interesting symptom is that specifically on the 'Stream' screen (the second icon from the left on the bottom navigation bar), I can't scroll past the top or bottom edge of the view, and there's no bouncing / snap-back animation like you can observe on the screen for private messages.

src/webview/js/js.js
```diff
@@ -624,7 +617,7 @@ const handleLongPress = (target: Element) => {
   sendMessage({
     type: 'longPress',
     target: target.matches('.header') ? 'header' : 'message',
-    messageId: getMessageIdFromNode(target),
+    messageId: ancestorMessageId(target),
```
borisyankov: I'd be curious to see if that will not throw the exception in `ancestorMessageId ` because of the `if (!messagePeerAncestor) {` clause.

gnprice: Thanks!

Moved this code over to #3097 .

gnprice:
OK -- and filed #3097 and #3098 for those followups.

PR #3041, opened 2018-10-11
Fix SpinningProgress issue with API 25+

borisyankov:
This PR contains two changes:
 * removes the 'thickness' from the public properties of `SpinningProgress`
 * replaces `SpinningProgress` ART-based implementation, with one based on static images

React Native's ART library is important and we probably will use
it more in the future. It is the only good option when we want to
achieve some visual elements / styles that we would use CSS or
SVG on the web instead.

Unfortunately, there is a serious bug on Android, that have existed
for a while. It becomes 'active' from API level 25 forward.

Visually, the issue manifests as black spinning square instead of
the usual progress spinner.

Bug reported to React Native here:
facebook/react-native#17565

A possible solution looks something like:
macdoum1/react-native@5878481

Possible alternatives evaluated:
 * Styles (ala CSS) - previous implementation used that, but was not
reliable between versions (due to RN issues)
 * use third-party component - a bit too complicated, and they suffer
from the same issues. For example:
bartgryszko/react-native-circular-progress#105
 * use static images - that is the chosen workaround

The exact look was recreated in Sketch and exported as PNG. We use
two versions - one with 'brand' color and one 'black'.

Once the bug in ART is fixed upstream, we will revert this change.


gnprice:

PR #3042, opened 2018-10-11
dependencies: Upgrade `react-native-fetch-blob`

borisyankov:
This fixes a blocker-issue with RN 0.57.2 that prevents the
project from building on Android.

The original project is no longer supported, but a very active
fork is being maintained and named as `rn-fetch-blob`.

This was manually tested and works fine. It contains multiple
fixes for issues that we do not have reports for, but feels nice
to auto-fix issues.


gnprice:
Hmm. I tried this just now, and I get the same result I described in this comment:
https://github.com/zulip/zulip-mobile/pull/2789#issuecomment-428418328

That's on Android 8.1 on an emulated Pixel 2 XL.

What do you think is causing that?

borisyankov:
Looking into that.

borisyankov:
Fixed!

gnprice:
Cool! WFM now, and that fix makes sense. Merged; thanks again!

PR #3045, opened 2018-10-12
Webview: Fix over strict type check on click handler.

armaanahluwalia:
This commit fixes a change introduced in d8352d where we're falsely
returning early if the element is not an instanceof HTMLElement.
Some elements like the down arrow are svg elements so we add
that condition for the return statement.

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -277,7 +277,8 @@ documentBody.addEventListener('click', function (e) {
 
   var target = e.target;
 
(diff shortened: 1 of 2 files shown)


gnprice:
Aha, thanks @armaanahluwalia for spotting and debugging that! Merged.

I added a commit on top to simplify the check again; just checking for the base class, `Element`, seems like it's what we really mean here.

armaanahluwalia:
Makes sense 🙂

PR #3047, opened 2018-10-12
Display !avatar images in message list

### LINKED ISSUE #3044: Make sure we display correctly !avatar images
One can include avatars via syntax:
```
!avatar(email@example.com)
```

Currently, rendering this does not work as expected.

borisyankov:
Fixes #3044

Avatars included in a message in the form of

```
!avatar(email@example.com)
```

are not being shown.

They are hosted at `/avatar/` and we do not provide auth for this
path. Fixing this is as simple as adding a check for these urls to
the `appendAuthToImages` function.


borisyankov:
Funnily and predictably different Babel versions produce different output, thus the failing tests (I did compile with version 7)

Issue #3048, opened 2018-10-13
layout: Limit the max height of compose box.


jainkuniya:
Currently compose box size can expand to whole chat screen. We should limit this height to something like every chat app does.

Both FB messenger and WhatsApp have limited this height to 6 lines of text (message).

on current master
<img width="485" alt="image" src="https://user-images.githubusercontent.com/18511177/46902580-58f3b780-cee5-11e8-9eb7-276de82e4c36.png">

<img width="369" alt="image" src="https://user-images.githubusercontent.com/18511177/46903594-e1c61f80-cef4-11e8-8926-2cdb43e63a53.png">

@zulipbot claim

gnprice:
Closing this, per discussion in the companion PR #3049.

PR #3049, opened 2018-10-13
composeBox: Limit maxHeight to 110.

jainkuniya:
Most chat apps limit the height of composeBox, and then text box
becomes scrollable on further entering text. FB messenger and WhatsApp
limit to 6 lines of text. Following the same pattern, limit the max
height of message input.

Fixes: #3048

screen shot 2018-10-13 at 1 04 07 pm

image

### DIFF
--- a/src/styles/composeBoxStyles.js
+++ b/src/styles/composeBoxStyles.js
@@ -52,6 +52,7 @@ export default ({ color, backgroundColor, borderColor }: Props) => ({
     backgroundColor,
     color,
     fontSize: 15,
+    maxHeight: 110,
     ...inputMarginPadding,
   },
   topicInput: {


borisyankov:
I am a bit suspicious that 110 equal 6 lines (would be true only if the line-height is `18.3333333` 😄 )

Regardless, we used to have this exact limit, but as I vaguely recall @timabbott wanted to be able to enter more text in the mobile app compared with other mobile apps (consistently with a tendency to write longer text on the web app).

I don't think the 'unlimited height' behavior is ideal, but neither is limiting it to 6-ish lines. Extra points for tracking the commit that did the change or the issue that it resolved to remind us of what was the reason to remove the limit,

armaanahluwalia:
I agree with @borisyankov. I don't think setting a max height really solves anything and takes away the ability to effectively use the ComposeBox as a space for drafting long messages. I've also made a small tweak to this code in #2887 

I just filed #3052 #3053 in case you're in the mood for text input stuff which to me are bigger problems while composing.

timabbott:
I've definitely composed messages longer than 6 lines of text in the current Zulip app (and I think copying Messenger and WhatsApp isn't right for our context, because Zulip, being a work tool, is generally going to have more long reason to send long messages explaining complex ideas).  I do think we need to have a reasonably scrolling situation for this, but I think limiting to 6 lines is counterproductive.

borisyankov:
Then we want to grow the compose box up to the height of the whole screen when the keyboard is popped up. (height = screen.height - keyboard.height).
If the user wants to see what was written, they can hide the keyboard. 
I'll do the change, as there are few more changes around this that I am working on.

PR #3050, opened 2018-10-13
refactor: Rename `getCurrentRoute` to `getCurrentRouteName`.

jainkuniya:
As this selector, returns `routeName`, instead of route (which also
contains other route details too like route params). So definately
`getCurrentRouteName` is more appropiate name for this, it's give more
clarity of what's it returns.

Pulling this commit out from #2807.

### DIFF
--- a/src/nav/__tests__/navSelectors-test.js
+++ b/src/nav/__tests__/navSelectors-test.js
@@ -1,14 +1,14 @@
 import deepFreeze from 'deep-freeze';
 
 import {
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks @jainkuniya ! Merged.

I agree with @borisyankov -- I'd love to just make this selector redundant and delete it. :smile: This seems like an unambiguous improvement from the status quo, though, so a good easy fix to make.

PR #3051, opened 2018-10-13
refactor: Move `getPrivateMessages` to `messageSelectors`.

jainkuniya:
Before it was in `baseSelectors`, which is not a right place for it.
All selectors related to `message` are present in `messageSelectors`,
so it's better to move it.

Pulling this commit out from #2807.


borisyankov:
Before considering merging this, I would say to wait for the 0.56 & 0.57 PR.
Then we can be a bit more confident we are not introducing any require cycles.
This was quite an issue with selectors. Until now we had to do it the check by trial and error.

This PR also helps reduce these a bit: #3070

gnprice:
@jainkuniya Now that the RN v0.57 upgrade is done (per @borisyankov 's comment above), want to try rebasing this and checking on require cycles? Would be glad to get this kind of improvement in. :slightly_smiling_face:

src/message/messageSelectors.js
```diff
@@ -2,10 +2,22 @@
...
+  getSubscriptions,
+} from '../directSelectors';
 import { getShownMessagesForNarrow } from '../chat/narrowsSelectors';
 import renderMessages from './renderMessages';
 import { findAnchor } from '../utils/message';
+import { ALL_PRIVATE_NARROW_STR } from '../utils/narrow';
+import { NULL_ARRAY } from '../nullObjects';
+
+export const getPrivateMessages = createSelector(getAllNarrows, getMessages, (narrows, messages) =>
+  (narrows[ALL_PRIVATE_NARROW_STR] || NULL_ARRAY).map(id => messages[id]),
```
jackrzhang: Are there any significant advantages to using `NULL_ARRAY` / `Object.freeze([])` here instead of just an array literal? It seems like it's an unnecessary layer here or I might be missing something.

borisyankov: Not sure how significant, but yeah. Here are few:
* Saves some memory (maybe too little to matter)
* Can not be mutated
* Most importantly `a === b` where both are `NULL_ARRAY` but `a !== b` when initialized with `[]`

gnprice:
Thanks @jainkuniya ! Merged.

Issue #2778, opened 2018-07-09
Improve re-login flow.


jainkuniya:
Do not navigate to realm screen if realm url is already present. (Basically when user taps on account item in account pick screen). As the server url is already validated. If url is found invalid pop back to realm screen.

@zulipbot claim

gnprice:
Closing this in favor of the newly-filed #3685. That one describes the underlying problem that the changes described here are intended to address.

PR #3055, opened 2018-10-15
Improve re-login flow. Fixes: #2778

jainkuniya:
https://cl.ly/587235c377db

### DIFF
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -1,10 +1,10 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { ActivityIndicator, StyleSheet, View } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 
+import type { ChildrenArray } from '../types';
 import { BRAND_COLOR } from '../styles';
 import { RawLabel, Touchable } from '../common';
-import { IconDone, IconTrash } from '../common/Icons';
 
 const styles = StyleSheet.create({
   wrapper: {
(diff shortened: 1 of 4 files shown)


src/account/AccountItem.js
```diff
@@ -1,6 +1,6 @@
 /* @flow */
```
borisyankov: > This resulted in too much automatically pushing and popping of screens, which doesn't look good.

Describe this in more detail. I think you might be referring to a behavior that is intended.

jainkuniya: Currently on clicking account item, we navigate to realm screen with auto fill realm url and tries to connect to it. So realm screen is pushed and within few seconds auth screen is pushed. 

On this PR, we are directly sending user to auth screen. Thus having less visual effects (sudden) which looks better. 

I guess we have discussed about this in India meet up :)

src/account/AccountPickScreen.js
```diff
@@ -56,7 +56,7 @@ class AccountPickScreen extends PureComponent<Props> {
             text="Add new account"
             style={styles.button}
             onPress={() => {
-              dispatch(navigateToRealmScreen(''));
+              dispatch(navigateToRealmScreen());
```
borisyankov: Good, we don't need that indeed, as we make sure to initialize the variable with '' if missing in `RealmScreen`.

src/account/AccountPickScreen.js
```diff
@@ -8,7 +8,7 @@ import type { Auth, Account, Dispatch, GlobalState } from '../types';
 import { getAuth, getAccounts } from '../selectors';
 import { Centerer, ZulipButton, Logo, Screen } from '../common';
 import AccountList from './AccountList';
-import { navigateToAddNewAccount, switchAccount, removeAccount } from '../actions';
+import { navigateToRealmScreen, switchAccount, removeAccount } from '../actions';
```
borisyankov: Good thinking. I like the rename.

borisyankov:
@gnprice I think the first two commits are a nice refactor and are independent of the 3d & 4th one and can be merged right now.
I am not completely convinced in the changes in the 3rd commit and will think about it again later.

src/account/AccountItem.js
```diff
@@ -63,7 +64,9 @@ export default class AccountItem extends PureComponent<Props> {
             <RawLabel style={styles.text} text={email} numberOfLines={1} />
             <RawLabel style={styles.text} text={realm} numberOfLines={1} />
           </View>
-          {!showDoneIcon ? (
+          {showProgress ? (
```
borisyankov: I am usually fine with a single ternary operator, but two of them usually make the logic too messy.

gnprice:
Merged those first two! ab9b52a2f.

Thanks @jainkuniya , and thanks @borisyankov for the review :smile:

src/account/AccountItem.js
jainkuniya: Moved this logic to `AccountList`

gnprice:
I've just filed #3685 for the underlying issue this is aimed at, which caused me to run across this PR again.

I don't think the code in this PR is the right direction -- it duplicates most of the logic that's in `tryRealm`, and that's a tricky stateful interaction that's well beyond the complexity of what I don't mind having duplicated.

Closing the PR. We'll use #3685 to continue to track the underlying issue.

gnprice:
Also, for cross-reference's sake: the original issue description this referred to was #2778 . (It's in a commit message, and the PR title... but neither of those cause GitHub to provide a link from the issue back to the PR.)

PR #3059, opened 2018-10-16
Simplify Screen component, use only ScrollView 

borisyankov:
When initially writing the Screen component, I was not aware of
the `scrollEnabled` property of `ScrollView`.

Replace the switching between a `ScrollView` and a `View` with
just setting the `scrollEnabled` to the correct value.


gnprice:
Neat. Thanks @borisyankov !

PR #3060, opened 2018-10-17
Reduce use of Redux with React Navigation

borisyankov:
Based on RN2 branch.

These are the first steps to remove React Navigation's dependency on Redux.
Some of them are a no-brainer. Some future ones would be a good idea too.
It is not yet clear if 100% Redux-less approach is ideal.


borisyankov:
Great. And also reducing the use of Redux + RNav will continue!

Issue #2861, opened 2018-08-03
Auto refresh in chat view loses the content of the message box


zingmars:
When an app has been closed (or switched away from) for a while with chat open and then re-opened the app will then refresh the whole chat view in order to get the latest messages. Sadly, if anything has been typed in the message box before the refresh (sometimes the refresh takes a few seconds to actually happen, in which case you already would've had enough time to write parts of your message) will disappear as the screen is refreshed. This can be quite frustrating as you now have to re-type the message you had already started to type.

gnprice:
I just hit this issue running 18.0.99 (a version I sent to our alpha channel yesterday), which has #2931. So apparently that didn't fix it as we'd hoped. :-/ Reopening this issue.

The specific sequence I happened to see today is a bit different from the OP:
* I'd been using the app, and started writing a reply message in a narrow.
* Then I got to my desk, and put down my phone for a while. (> 10 minutes.)
* I picked up my phone, and either it was already in Zulip or I went and switched to Zulip. It was still in that narrow, and showing my reply message.
* After a second or two, the refresh happened (the messages were replaced by the "loading" animation, then went back to showing the messages); and my draft reply was gone.

I.e., the draft message I lost was from before re-opening the app, not after. But I don't know of a reason the behavior would be different in those two cases.

borisyankov:
Thanks for the detailed explanation.
I did reproduce it in a controlled environment and working on it.

borisyankov:
Another update: The issue does not exist with a controlled input component.
But, I did find where and why this happens, so a fix is coming next.
In fact, it is kinda fascinating and unexpected why this happens (to be read about in an upcoming commit message)

borisyankov:
A fix is ready.

PR #3065, opened 2018-10-19
composebox: Do not unmount Input on hide

borisyankov:
When `canSend` property is `false` we hide the ComposeBox component.
Then after it appears again the contents of both topic and message
Inputs are missing.

This issue exists only with our uncontrolled Input.

The cause is quite curious. We correctly initialize the components
on `componentDidMount`. So the component is working correctly on
initial render. What is messing it up is that if `canSend` becomes
`false` we do render a `null` (so nothing). This happens when we
are refreshing the state, on `APP_REFRESH` called on broken queues,
but can happen in other cases too (any case that makes `canSend` be
`false`).

When we show the component again, since we handle the show/no show
logic inside the component, the lifecycle event `componentDidMount`
is not called again. Thus we do not initialize he values of the
Inputs that were unmounted.

There are several approaches to fixing this:
 * take the `canSend` logic ouside of the component. On each change
the whole component will get unmounted and the `componentDidMount`
will be called again - not a bad solution
 * handle the `false` to `true` transition of the value inside the
`componentWillReceiveProps` - would work, but we are trying to get
rid of this funcion, and also makes the logic too convoluted
 * do not unmount the Inputs when hiding - the chosen solution

When we need to hide the component, instead of not rendering it at
all, we just hide it - setting the `opacity` to 0 does that.
We also set `position` to `absolute` so it does not take up space
visually.

Tested both on iOS and Android and works well.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -277,10 +277,6 @@ class ComposeBox extends PureComponent<Props, State> {
       isSubscribed,
     } = this.props;
 
-    if (!canSend) {
-      return null;
-    }
-
     if (!isSubscribed) {
       return <NotSubscribed narrow={narrow} />;
     } else if (isAnnouncementOnly && !isAdmin) {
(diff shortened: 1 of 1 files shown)


gnprice:
Boris added in chat:

> After I completed he work, I realized it can be based on current `master` as it is specific to the Android version.

borisyankov:
Now that we have the RN upgrades, rebased on `master` anyway.
Ready to go in!

gnprice:
Thanks @borisyankov for this investigation and fix, and the very clear commit message! Merged.

gnprice:
Adding a cross-reference because this PR didn't have one: this fixed #2861.

Issue #3066, opened 2018-10-20
Safe areas are not handled correctly on iPhone X and Androids with notches


borisyankov:
We are currently handling, but not 100% correctly, the 'screen border offsets' known as 'safe areas' on iOS.

Few related bugs:
 * iPhone X with popped keyboard has some extra space below the compose box
 * the app on Android with notches may not take up the full height of the screen

gnprice:
@jackrzhang provided (on #3389) some handy screenshots that very clearly demonstrate one of the symptoms of this issue:

> On any iPhone X model, the upper edge of the screen covers part of the application UI in landscape orientation:
> 
> <img alt="" width="610" src="https://user-images.githubusercontent.com/12771126/54060091-76ddf380-41c9-11e9-9ec6-81b4326f60e7.png">
> 
> <img alt="" width="650" src="https://user-images.githubusercontent.com/12771126/54060080-6594e700-41c9-11e9-96e6-f46d48baa2bc.png">
> 
> [...]
> 
> I don't have a physical iPhone X device to use for testing, so it's also possible that this reproduces only on a simulator.

gnprice:
I think #3370 is probably also a symptom, indirectly:

> [From a video by @armaanahluwalia to demonstrate a different issue:]
> ![](https://user-images.githubusercontent.com/28173/53456805-9d8b7580-39e4-11e9-9f1f-373e9e1d8688.png)
> 
> A blank white band between the compose box and the keyboard.

Didn't repro for me on Android, and @jackrzhang reported it repro'd for him on a simulated iPhone XR but not on an iPhone 7. So that's consistent with it happening only on iOS devices that have a display cutout.

I suspect it's *caused by* the logic we have for attempting to avoid the cutout, i.e. by a bug in that logic.

gnprice:
Additional info from the dupe issue #3064 (filed just before this one), quoting @borisyankov:

> When iPhone X was released, new functionality to iOS was added to indicate `safe area` - to help with apps handling optional notches, rounded corners or the ...
> 
> A good article explaining all this here:
> https://medium.com/rosberryapps/ios-safe-area-ca10e919526f
> 
> We started using a component `react-native-safe-area` to handle this.
> 
> It isn't working perfectly and we need to fix it.
> Testing on iOS X or Simulator with it makes few layout issues obvious.
> 
> Better yet, use the now built-in component:
> https://facebook.github.io/react-native/docs/safeareaview

gnprice:
[...]

PR #3067, opened 2018-10-21
Use 'SafeAreaView' through the app

borisyankov:
Based on #2702 (as it changes internally how it uses `SafeAreaView`)


borisyankov:
Rebased and updated. 
Now going over all the changes another time and double-checking everything works fine and nothing is missing.

gnprice:
P1 because it fixes #3066 .

gnprice:
We're now on react-navigation v2 (see #3573), so this is unblocked!

gnprice:
Just rebased this branch. (Across over a year's changes!)

chrisbobbe:
Just rebased across another year's changes; I came across this while looking at https://github.com/zulip/zulip-mobile/issues/4267.

There are a few issues with this branch that will need to be resolved:

- Things break if, on a single screen, we use a `SafeAreaView` and a `ZulipStatusBar` while the `ZulipStatusBar` still sets its own height. But the commits that add `SafeAreaView` and the commit that removes the `ZulipStatusBar`'s height are separate. They're not that large, so they could probably be reasonably combined into one commit.
- The status bar's background color isn't always correct (see, in particular, the chat screen).
- Tapping the image in the lightbox is supposed to make the header and the footer slide away completely; now, they just retreat into the safe area (top and bottom) but are still quite visible.

chrisbobbe:
Also, it's worth noting that React Navigation has its own docs that go through how to handle safe areas—in a way that doesn't (at least doesn't directly) use the `SafeAreaView` component from React Native: https://reactnavigation.org/docs/4.x/handling-iphonex/.

chrisbobbe:
This has been superseded by our current plan for #3066. Currently, we have #4683 open for part of that.

Issue #2694, opened 2018-06-18
Fix misleading uses of "group"


gnprice:
As touched on in #2609, discussing where we used to say "Members" above a list of the other people in a group PM thread:

> I'm also not a fan of the term "members" here, as it sounds like a model where you could add and remove people to/from the group while the group keeps its identity -- like groups in FB Messenger and similar SMS-inspired systems.

We fixed that screen, but we have a few other places that sound like that model. One is "Create group", on the PMs screen of the main nav; for a new user, this is likely to look like what they want when they're actually looking for how to create a stream, because in those other systems the thing that's most like a Zulip stream is called a "group".

From a quick grep, the other strings that look suspicious are:

New group chat
Message group
No messages in this group

jackrzhang:
@gnprice 

Another two closely related items to adjust might be the `groupNarrow` and `isGroupNarrow` functions in `narrow.js`, which are identical to the `privateNarrow` and `isPrivateNarrow` functions, except that a `groupNarrow` refers to more than one email.

From the conversation in #2613, I think "recipients" is the preferred term here.

Here are possible rewordings I have in mind:
- "Create group" -> "Message new recipients"
- "New group chat" -> (also) "Message new recipients"
- "Message group" -> "Message recipients"
- "No messages in this group" -> "No messages for these recipients"

Thoughts?

borisyankov:
I never refer to people I want to talk to as 'recipients'. Does feel like a term we would use in code (we do) and not conversationally.

armaanahluwalia:
I agree recipients is ok in a specific situation but not as an overall label. Why not "New Group Chat"? or  "New Group Message"?

jackrzhang:
Hm - another possible criticism of "recipients" is that it's too broad here; in that every message sent has a recipient or group of recipients.

Maybe "group PM" instead? So rewordings might be:
- "Create group" -> "Create group PM"
- "New group chat" -> "New group PM"
- "Message group" -> "Message group PM"
- "No messages in this group" -> "No messages in this group PM", or "No messages in this thread", or I could see the term "recipients" being used here: "No messages for these recipients"

borisyankov:
PM means nothing to a normal person :)

borisyankov:
"New group chat" does not imply in any way that a "group" is created, the other names are misleading though.

jackrzhang:
The abbreviation of "private message" as PM, especially in the context of internet forums, is not new.

borisyankov:
[...]

PR #3071, opened 2018-10-23
Remove unused string "New group chat".

jackrzhang:
The string "New group chat" only appears in the translations files, which means that it isn't being displayed anywhere on the UI.

See https://github.com/zulip/zulip-mobile/issues/2694#issuecomment-432430078.

### DIFF
--- a/src/i18n/translations/messages_ar.json
+++ b/src/i18n/translations/messages_ar.json
@@ -78,7 +78,6 @@
   "Subscribed": "Subscribed",
   "No users found": "No users found",
   "No streams found": "No streams found",
-  "New group chat": "New group chat",
   "Diagnostics": "Diagnostics",
   "Toggle compose tools": "Toggle compose tools",
   "Terms of service": "Terms of service",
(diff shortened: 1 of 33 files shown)


borisyankov:
Did you have access to Transifex?
We generally do update the `/messages_en.json` then update the Transifex resource messages.
Then the rest translations are updated when we update translations from here.

That being said, I don't see any reason not to merge this as-is.

jackrzhang:
@borisyankov I can see the Zulip's dashboard at https://www.transifex.com/zulip/zulip/ (if that's what you're referring to)?

borisyankov:
Yeah, the resource you need for the mobile app is `mobile.json`
https://www.transifex.com/zulip/zulip/mobile/

borisyankov:
@jackrzhang I think I just did invite you (not sure, the UI of Transifex did not confirm, a bit buggy I think)

jackrzhang:
(Got the invitation - thanks!)

gnprice:
Thanks @jackrzhang and @borisyankov -- merged!

PR #3072, opened 2018-10-24
Authenticate image downloads by appending an 'api_key' parameter.

### LINKED ISSUE #2618: Download from lightbox fails to authenticate
Steps to reproduce:
* Log into an org on zulipchat.com, or any other server running a recent release and on Xenial. (Not chat.zulip.org, which is on Trusty.)
* Find a conversation with a user-uploaded image in it. (Or just upload one.)
* Tap the image, bringing up the lightbox.
* Tap the ellipsis in the lower right to bring up a menu; select "Download file".
* The download fails. It should succeed.

I tested this just now with 13.4.88 on Android. I'm pretty sure the cause is that we just haven't fixed the download to provide authentication to the server. For more details, see commit cb3198606, where we fixed the display of images within the message list webview.

jackrzhang:
Fixes #2618.

The Zulip server generally requires authentication for uploading
and downloading images. Since React Native's webview doesn't support
setting HTTP Authorization headers at the moment, we pass the API key
as a query parameter. See zulip/zulip@5ddf261.

There was previously an `appendAuthToImages` utility method that was
added in cb31986 and then removed in 9e28ef2. However,
`downloadFile.js` is used only for images, so opt to just append the
api_key via string concatenation, which is simple and easy.

### DIFF
--- a/src/api/downloadFile.js
+++ b/src/api/downloadFile.js
@@ -7,10 +7,12 @@ import { getAuthHeader, getFullUrl } from '../utils/url';
 import userAgent from '../utils/userAgent';
 
 export default (src: string, auth: Auth) => {
-  const fullUrlWithAPIKey = `${getFullUrl(src, auth.realm)}?api_key=${auth.apiKey}`;
+  const absoluteUrl = getFullUrl(src, auth.realm);
 
   if (Platform.OS === 'ios') {
-    return CameraRoll.saveToCameraRoll(fullUrlWithAPIKey);
+    const delimiter = absoluteUrl.includes('?') ? '&' : '?';
+    const urlWithApiKey = `${absoluteUrl}${delimiter}api_key=${auth.apiKey}`;
+    return CameraRoll.saveToCameraRoll(urlWithApiKey);
   }
   return RNFetchBlob.config({
     addAndroidDownloads: {
@@ -20,7 +22,7 @@ export default (src: string, auth: Auth) => {
       title: src.split('/').pop(),
       notification: true,
     },
-  }).fetch('GET', getFullUrl(src, auth.realm), {
+  }).fetch('GET', absoluteUrl, {
     'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8',
     'User-Agent': userAgent,
     Authorization: getAuthHeader(auth.email, auth.apiKey),
--- a/src/api/downloadFile.js
+++ b/src/api/downloadFile.js
@@ -7,8 +7,10 @@ import { getAuthHeader, getFullUrl } from '../utils/url';
 import userAgent from '../utils/userAgent';
 
 export default (src: string, auth: Auth) => {
+  const fullUrlWithAPIKey = `${getFullUrl(src, auth.realm)}?api_key=${auth.apiKey}`;
+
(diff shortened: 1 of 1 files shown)


src/api/downloadFile.js
```diff
@@ -7,8 +7,10 @@ import { getAuthHeader, getFullUrl } from '../utils/url';
 import userAgent from '../utils/userAgent';
 
 export default (src: string, auth: Auth) => {
+  const fullUrlWithAPIKey = `${getFullUrl(src, auth.realm)}?api_key=${auth.apiKey}`;
```
borisyankov: Is it guaranteed that the URL will not already have a `?` in it?

borisyankov: In `js.js` we do have this code:

```js
const delimiter = img.src.includes('?') ? '&' : '?';
img.src += `${delimiter}api_key=${auth.apiKey}`;
```

gnprice: Yeah, that logic seems more robust to me.

Too bad it isn't currently easy to share code between `js.js` and the rest of the app -- but fortunately this is only a line or so and unlikely to change, so it's not a big problem to duplicate it. :slightly_smiling_face:

gnprice:
Thanks @jackrzhang !

Style nit in the commit message: prefer writing 9 nibbles of Git commit IDs, rather than 7. It's not much longer, and is more future-proof. (You can set `git config --global core.abbrev 9` to make Git print out that many in places like `git log --oneline`.)

A question about the logic, as nicely explained in the commit message:

> Since React Native's webview doesn't support
> setting HTTP Authorization headers at the moment, we pass the API key
> as a query parameter.

But this download isn't happening in a webview. Can we pass an option somewhere to make these requests happen with the `Authorization` header?

ISTR we looked at this together on the iOS side, and the library we're using didn't have such an option (or was that just for the lightbox itself?). If that's right, then this solution is fine, just need to explain that reasoning. :wink: 

Hmm, but in fact on the Android side, it looks like we already do that!
```
     Authorization: getAuthHeader(auth.email, auth.apiKey),
```
Do you reproduce #2618 on Android, and does this change in fact fix it? If so that's interesting and puzzling -- what's going wrong with that header?

In any case, we shouldn't send *both* the key in the query string and also the header -- if for some reason we can't make the header work, we should cut it out. But first let's try to debug the header and get that working :smile: -- even if it's only on Android, that will reduce the number of cases where we do this kind of worrying thing where we put an API key right in the URL.

jackrzhang:
> But this download isn't happening in a webview. Can we pass an option somewhere to make these requests happen with the Authorization header?

Oops - that's correct! The specific interface that's being used to download an image on iOS isn't from a webview, it's React Native's `CameraRoll.savetoCameraRoll`, which is documented [here](https://facebook.github.io/react-native/docs/cameraroll#savetocameraroll). The `tag` parameter is a string; there's still no way to pass an Authorization header, so the current fix should be OK for iOS. I've updated the commit!

For progressing the Android implementation, see #3115.

gnprice:
Thanks @jackrzhang ! Merged.

On the inline-comments thread, perhaps my reply was confusing -- I meant to go ahead and duplicate that code that has more robust logic. I added a second commit on top to do that.

PR #3077, opened 2018-10-25
Added properties on gradle.properties for help in generating release apk

anirudhjain75:
The previous edition of the Gradle.properties didn't contain the properties required for generation of apk which sometimes may confuse the developer.

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -20,3 +20,8 @@
 # TODO remove this after RN 0.57 upgrade.
 #      See https://github.com/facebook/react-native/issues/16906 .
 android.enableAapt2=false
+
+MYAPP_RELEASE_STORE_FILE=keystore-file-name.keystore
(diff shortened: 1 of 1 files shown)


anirudhjain75:
The main reason behind adding these are so anyone can generate APK and sign it to use it on their phone. Instead of having to run node package manager and connecting the phone with a usb cable.

gnprice:
@anirudhjain75 Hmm, I don't think this makes sense as a solution, for a couple of reasons. Let's figure out the problem first, then we can work out a solution if needed.

If you're trying to run a release-style build of the app, take a look at our instructions on `yarn build:android-unsigned`: https://github.com/zulip/zulip-mobile/blob/master/docs/howto/build-run.md#android-tips and try those. Then install the APK with `adb install`. If something isn't working for you, ask in chat and let's debug.

Issue #3080, opened 2018-10-26
isReady may remain false in MessageListWeb


jainkuniya:
We are using `postMessage` to communicate between react app and message list (i.e WebView).
`postMessage` is acting as a bridge between whole app and webView.
Whenever there is a update in app e.g.: typing events or new message etc we send html content to WebView via this `postMessage` bridge. And similarly we get scroll, touch/click events from WebView (Message List) to react native app via the same bridge  `postMessage`.

https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/react.md#the-exception-messagelistweb

One thing we need to make sure is, bridge should be successfully established  before we start communication between two components. Else we will be sending messages from one side while the other side is not yet ready to receive/process messages. Thus this will cause in consistency and the updates will not be observed in the real time. 

On current master, we are sending `isReady` message whenever webView is ready to receive/send/process messages. But the the problem is, what if when webView send `isReady` message at that time RN part is not ready or this event is not received by RN? We are just ignoring this case currently.
Then according to this RN will always thinks that webView is not ready and will never send any new message to webView, and keep on waiting till it hears `isReady` from webView. 

On current master, one way hand shake is implemented (explained in above paragraph).

Which starts causing problems on Android 9. Issue like:
* #3078
* Typing indicator are never shown in the message list.
* More messages are never loaded on scrolling to top/bottom.
* White screen (or message loading) on narrowing to a narrow in which we don't have any message.
  Because on such narrow, we request new messages from server and send to webView via `postMessage`. So after fetching we want to send new html content to WebView, but RN thinks 
WebView is not ready, so it doesn't send. (https://chat.zulip.org/#narrow/stream/48-mobile/subject/messages.20don't.20load/near/655786)
> oh, it's a mix. Sometimes loading bars, sometimes not

(https://chat.zulip.org/#narrow/stream/48-mobile/subject/messages.20don't.20load/near/655849)
* etc etc (actually whole communication seems to be broken)

Solution: Improve/enhance this hand shaking mechanism. Implement a two way hand shake.
Keep sending a message from one side till a confirmation from other side is heard/received. 
I will prefer to keep this **one side** as RN one, because we have more control on RN one and it is easier to debug in RN.

(Sending PR in few minutes)

@zulipbot claim

borisyankov:
Great find, @jainkuniya Also, thanks for the quick reaction.
I have some comments regarding the code itself but will do it as a review in the PR.

[...]

PR #3081, opened 2018-10-26
webView: Improve hand shake mechanism with app.

jainkuniya:
Implement a two way hand shake. Send a ping from app to webView till a
ping is received from webView.

Refer #3080 for more details.

Fixes: #3080

### DIFF
--- a/src/webview/MessageListWeb.js
+++ b/src/webview/MessageListWeb.js
@@ -15,8 +15,8 @@ import { base64Utf8Encode } from '../utils/encoding';
 export default class MessageListWeb extends Component<Props> {
   context: Context;
   props: Props;
-  isWebViewReady: boolean;
   webview: ?Object;
+  sendMessagesIsReady: boolean;
   unsentMessages: WebviewInputMessage[] = [];
 
   static contextTypes = {
@@ -25,28 +25,24 @@ export default class MessageListWeb extends Component<Props> {
   };
 
   componentDidMount() {
-    /*
-     * Initiates a two-way handshake with the WebView. Ensures that both sides are
-     * ready to communicate before considering to be ready.
-     *
-     * This component communicates with webView with the help of `postMessage`.
-     * It needs to be ensured that this communication channel is established successfully
-     * on both side, else it will cause issue like #3080.
-     * Two way hand shake mechanism is used for this purpose.
-     *
-     * Send a hello event till a confirmation is reveived from webView.
-     */
+    this.setupSendMessages();
+  }
+
+  handleError = (event: Object) => {
+    console.error(event); // eslint-disable-line
+  };
+
+  /**
+   * Initiate round-trip handshakes with the WebView, until one succeeds.
+   */
+  setupSendMessages = (): void => {
     const intervalId = setInterval(() => {
-      if (!this.isWebViewReady) {
+      if (!this.sendMessagesIsReady) {
         this.sendMessages([{ type: 'ready' }]);
       } else {
         clearInterval(intervalId);
       }
     }, 30);
-  }
-
-  handleError = (event: Object) => {
-    console.error(event); // eslint-disable-line
   };
 
   sendMessages = (messages: WebviewInputMessage[]): void => {
@@ -58,7 +54,7 @@ export default class MessageListWeb extends Component<Props> {
   handleMessage = (event: { nativeEvent: { data: string } }) => {
     const eventData: MessageListEvent = JSON.parse(event.nativeEvent.data);
     if (eventData.type === 'ready') {
-      this.isWebViewReady = true;
+      this.sendMessagesIsReady = true;
       this.sendMessages(this.unsentMessages);
     } else {
       const handler = `handle${eventData.type.charAt(0).toUpperCase()}${eventData.type.slice(1)}`;
@@ -69,7 +65,7 @@ export default class MessageListWeb extends Component<Props> {
   shouldComponentUpdate = (nextProps: Props) => {
     const messages = getInputMessages(this.props, nextProps);
 
-    if (this.isWebViewReady) {
+    if (this.sendMessagesIsReady) {
       this.sendMessages(messages);
     } else {
       this.unsentMessages.push(...messages);
--- a/src/webview/MessageListWeb.js
+++ b/src/webview/MessageListWeb.js
@@ -15,14 +15,35 @@ import { base64Utf8Encode } from '../utils/encoding';
 export default class MessageListWeb extends Component<Props> {
   context: Context;
   props: Props;
+  isWebViewReady: boolean;
   webview: ?Object;
-  isReady: boolean;
   unsentMessages: WebviewInputMessage[] = [];
 
   static contextTypes = {
     styles: () => null,
     theme: () => null,
   };
+
+  componentDidMount() {
+    /*
+     * Initiates a two-way handshake with the WebView. Ensures that both sides are
+     * ready to communicate before considering to be ready.
+     *
+     * This component communicates with webView with the help of `postMessage`.
+     * It needs to be ensured that this communication channel is established successfully
+     * on both side, else it will cause issue like #3080.
+     * Two way hand shake mechanism is used for this purpose.
+     *
+     * Send a hello event till a confirmation is reveived from webView.
+     */
+    const intervalId = setInterval(() => {
+      if (!this.isWebViewReady) {
+        this.sendMessages([{ type: 'ready' }]);
+      } else {
+        clearInterval(intervalId);
+      }
+    }, 30);
+  }
 
   handleError = (event: Object) => {
     console.error(event); // eslint-disable-line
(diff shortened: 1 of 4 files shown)


src/webview/MessageListWeb.js
```diff
@@ -12,18 +12,42 @@ import { getInputMessages } from './webViewHandleUpdates';
...
+  *
+  * Synchronizing setup with webview
+  *
+  */
+  sendTestPing = (): void => {
+    if (this.state.isWebViewReady) {
+      return;
+    }
+    this.sendMessages([{ type: 'ping' }]);
+    setTimeout(this.sendTestPing, 10);
```
borisyankov: I am quite concerned about this value. Sending messages to the WebView is not slow, but it isn't just a normal a method call either. Everything goes through the bridge (somewhat slow) and then is executed as simple `MessageEvent` (maybe fast enough)

src/webview/MessageListWeb.js
```diff
@@ -12,18 +12,42 @@ import { getInputMessages } from './webViewHandleUpdates';
...
+  };
+
   static contextTypes = {
     styles: () => null,
     theme: () => null,
   };
 
+  /*
+  *
+  * Synchronizing setup with webview
```
borisyankov: That description is not clear.

src/webview/MessageListWeb.js
```diff
@@ -12,18 +12,42 @@ import { getInputMessages } from './webViewHandleUpdates';
...
+  *
+  */
+  sendTestPing = (): void => {
+    if (this.state.isWebViewReady) {
+      return;
+    }
+    this.sendMessages([{ type: 'ping' }]);
```
borisyankov: I would have called this event 'Hello'.
This is way more typical greeting when doing a handshake, IRL and digitally.
`ping` tends to mean 'are you still alive'.

borisyankov:
Latest findings point to this likely being related to Chrome 70 and not likely to be Android 9
If so, that is even more urgent issue as the Chrome browser will update for most users in 1-2 weeks (I suspect they are doing some staged rollouts thus not everyone is getting the update immediately)

src/webview/MessageListWeb.js
```diff
@@ -36,8 +60,10 @@ export default class MessageListWeb extends Component<Props> {
 
   handleMessage = (event: { nativeEvent: { data: string } }) => {
     const eventData: MessageListEvent = JSON.parse(event.nativeEvent.data);
-    if (eventData.type === 'ready') {
-      this.isReady = true;
+    if (eventData.type === 'ping') {
```
borisyankov: Looking at all this renaming, I am not sure my suggestion for 'hello' is optimal.
I think keeping the name 'ready' is superior both to 'hello' and 'ping' as it reveals the purpose for the event.

src/webview/js/js.js
```diff
@@ -491,18 +502,3 @@ documentBody.addEventListener('touchmove', (e: TouchEvent) => {
...
   lastTouchEventTimestamp = 0;
 });
-
-/*
- *
- * Synchronizing setup with outside
- *
- */
-
-const waitForBridge = () => {
```
borisyankov: Now, you are assuming the WebView will be immediately ready to send messages.
That might be true in Chrome 70 but not in previous (or even future) versions!

src/webview/MessageListWeb.js
jainkuniya: So should we increase or decrease this value?
```suggestion
    setTimeout(this.sendTestPing, 5);
```

or

```suggestion
    setTimeout(this.sendTestPing, 15);
```

src/webview/MessageListWeb.js
```diff
@@ -12,18 +12,42 @@ import { getInputMessages } from './webViewHandleUpdates';
 import * as webViewEventHandlers from './webViewEventHandlers';
 import { base64Utf8Encode } from '../utils/encoding';
 
-export default class MessageListWeb extends Component<Props> {
+type State = {
```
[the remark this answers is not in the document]
jainkuniya: > What is the reason to switch from a class instance property to state?

I thought it's better organisable if it is in state.
But as State is used for storing data on which we want render to be called again. But for this case we don't want render. So  class instance will be better, reverting back :)
👍

src/webview/js/js.js
jainkuniya: > you are assuming the WebView will be immediately ready to send messages.

kind of.

If it fails then React component will again send `hello` event and at that time it will respond.

Anyway we need to trust one side. So I thinks it better to have complexity on that side where we can debug easily, and make less debugging side with less logic.

src/webview/MessageListWeb.js
borisyankov: I was thinking somethinkg like 100 ;)

src/webview/MessageListWeb.js
```diff
@@ -15,15 +15,32 @@ import { base64Utf8Encode } from '../utils/encoding';
...
+  *
+  * Synchronizing setup with webview
+  *
+  */
+  sendTestPing = (): void => {
+    if (this.isWebViewReady) {
+      return;
+    }
+    this.sendMessages([{ type: 'hello' }]);
+    setTimeout(this.sendTestPing, 10);
```
borisyankov: Another point about `setTimeout`, we should use `setInterval` instead.
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval

borisyankov:
After testing the actual PR on an actual phone... :)
It seems to be working OK, but there is some weird delay when rendering the `ChatScreen` itself.

src/webview/MessageListWeb.js
borisyankov: To illustrate this, I did change our `waitForBridge` in #3088

src/webview/MessageListWeb.js
```diff
@@ -15,15 +15,32 @@ import { base64Utf8Encode } from '../utils/encoding';
...
     theme: () => null,
   };
 
+  /*
+  *
+  * Synchronizing setup with webview
+  *
+  */
+  sendTestPing = (): void => {
+    if (this.isWebViewReady) {
```
borisyankov: I don't think this will ever be `true` on the first run.
If we were to stick with `setTimeout` you could do something like
```js
if (!this.isWebViewReady) {
  setTimeout(this.sendTestPing, 10);
}
```

src/webview/webViewEventHandlers.js
```diff
@@ -15,8 +15,8 @@ import {
   messageLinkPress,
 } from '../actions';
 
-type MessageListEventReady = {
-  type: 'ready',
+type MessageListEventPing = {
```
borisyankov: After some more thinking, while 'helo' is better than 'ping', 'ready' is even better. Let's keep it 'ready'.

This means no change to this file, and few undos/changes to the other files

src/webview/MessageListWeb.js
```diff
@@ -15,15 +15,32 @@ import { base64Utf8Encode } from '../utils/encoding';
...
     styles: () => null,
     theme: () => null,
   };
 
+  /*
+  *
+  * Synchronizing setup with webview
+  *
+  */
+  sendTestPing = (): void => {
```
borisyankov: Rename to one of these:
 * sendHandshake
 * initiateHandshake
 * sendReadyHandshake
 * initiateReadyHandshake

The choice is yours 😄

src/webview/MessageListWeb.js
```diff
@@ -15,15 +15,35 @@ import { base64Utf8Encode } from '../utils/encoding';
...
-  isReady: boolean;
   unsentMessages: WebviewInputMessage[] = [];
 
   static contextTypes = {
     styles: () => null,
     theme: () => null,
   };
 
+  componentDidMount() {
+    /*
```
borisyankov:
```suggestion
/* Initiates a two-way handshake with the WebView. Ensures that both sides are
 * ready to communicate before considering to be ready. */
```

src/webview/js/js.js
```diff
@@ -327,7 +328,23 @@ const handleMessageTyping = (msg: MessageInputTyping) => {
...
+ *
+ * This webView communicates with react component (app) with the help of `postMessage`.
+ * It needs to be ensured that this communication channel is established successfully
+ * on both side, else it will cause issue like #3080.
+ * Two way hand shake mechanism is used for this purpose.
+ *
+ * Send a hello event as soon as hello event is reveived from React
+ * component to convey that webView is active.
+ *
+ */
```
borisyankov:
```suggestion
 /* In response to a `ready` message send back another `ready` message and thus
  * complete the handshake process. When it is received by the other side both sides are ready. */
```

gnprice:
Thanks @jainkuniya for the diagnosis and fix, and @borisyankov for the reviews! This looks generally great, apart from some small edits I'll make. (I'll include Boris's latest suggested changes.)

Just one thing in the behavior: 100ms feels to me like a long time to wait -- it's a very noticeable amount of delay for a UI. Let's go for 30ms. (Still kind of a while, but on trying 10ms and playing with it, we have plenty of other delay involved. :-/ )

gnprice:
(Confirmed the `length` values with Chrome 70 and Chrome 61, as I had the latter lying around in an Android 8.1 virtual device.)

gnprice:
And merged -- thanks again!

I'm going to go make a release from this, and push it to beta after a quick smoketest -- I'd like it to go to prod before the end of today. This is a pretty killer issue for anyone with the Chrome 70 update, which I think may be over half of users now :open_mouth: -- glad to have it fixed :tada:

PR #3084, opened 2018-10-29
home tab: Show unread mentions count on mentions icon.

jainkuniya:
Currently there is no way to get unread mentions count, So show it on
the icon itself as unread PM count is show over conversation icon.

image

### DIFF
--- a/src/nav/IconUnreadMentions.js
+++ b/src/nav/IconUnreadMentions.js
@@ -1,0 +1,34 @@
+/* @flow strict-local */
+
+import React, { PureComponent } from 'react';
+import { View } from 'react-native';
+
+import type { Dispatch } from '../types';
+import { connect } from '../react-redux';
+import { getUnreadMentionsTotal } from '../selectors';
+import { IconMention } from '../common/Icons';
+import { CountOverlay } from '../common';
+
+type Props = {|
+  dispatch: Dispatch,
+  unreadMentionsTotal: number,
+  color: string,
+|};
+
+class IconUnreadMentions extends PureComponent<Props> {
+  render() {
+    const { unreadMentionsTotal, color } = this.props;
+
+    return (
+      <View>
+        <CountOverlay unreadCount={unreadMentionsTotal}>
+          <IconMention size={24} color={color} />
+        </CountOverlay>
+      </View>
+    );
+  }
+}
+
+export default connect(state => ({
+  unreadMentionsTotal: getUnreadMentionsTotal(state),
+}))(IconUnreadMentions);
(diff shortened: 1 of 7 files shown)


borisyankov:
Cool. We need that!
Do we want to expand it for all other cases?

Have a look at `IconUnreadConversations.js`
We either want the same for 'mentions' or maybe we should generalize for other unreads?

jainkuniya:
> Have a look at `IconUnreadConversations.js`

Before I was thinking to do something similar as done in `Have a look at IconUnreadConversations.js`. Then I realised that `NavButton` can also have functionality to overlay component, so lets use it (minimal change 😅 )


> we should generalize for other unreads?

* Top tab bar
  * Will do for Home in another PR.
  * For star, search not required. (actually can't possible 😅 )
* Bottom tab bar
  * for home, not required. Will become too noisy, as count for this is same as of Home in top bar.
  * for streams, will no in another PR
  * for PM, already implemented
  * for setting, can't possible 😅 


image


cc @borisyankov @jackrzhang

borisyankov:
A related bug, now looking at the source code, the text of the `unread count` should be white, not black, separate PR though ;)

borisyankov:
@gnprice The code looks good and I have tested PR locally and it works as expected.

gnprice:
Thanks @jainkuniya !

Looking at this in the emulator, it feels odd how the count is detached from the icon -- way off to the upper right. (It's similar in the screenshot above).

I'd prefer to have it more similar to the count of unread PMs, where it overlaps part of the icon.

Looking closer... it looks like this code in `NavButton` has been unused! Before this branch, it was last used before 58b6bb768dcb50cd08196f6ffd2d3aa7fb3e4901 , over two years ago. So, it's unsurprising that it doesn't quite do what we want.

... I'll actually go cut it out, so it doesn't cause confusion. A good model for this PR would be the code in `IconUnreadConversations.js`, which implements the count on the PMs icon.

gnprice:
> ... I'll actually go cut it out, so it doesn't cause confusion.

Done as 180f36fed. See also commits up to 99ab6c1c8.

src/nav/IconUnreadMentions.js
```diff
@@ -0,0 +1,39 @@
...
+type Props = {|
+  dispatch: Dispatch,
+  unreadMentionsTotal: number,
+  color: string,
+|};
+
+class IconUnreadMentions extends PureComponent<Props> {
+  static defaultProps = {
+    color: BRAND_COLOR,
+  };
```
gnprice: Let's make this exactly parallel to `IconUnreadConversations`: the caller passes the color.

jainkuniya: Done 👍

gnprice:
Thanks @jainkuniya -- merged!

PR #3086, opened 2018-10-29
unread count overlay: Add scale animation to unread count.

jainkuniya:
It's looks much better with the animation, than just a re-render.

### DIFF
--- a/src/common/ComponentWithOverlay.js
+++ b/src/common/ComponentWithOverlay.js
@@ -6,6 +6,7 @@ import { StyleSheet, View } from 'react-native';
 import type { ChildrenArray, Style } from '../types';
 import { BRAND_COLOR } from '../styles';
 import Touchable from './Touchable';
+import AnimatedScaleComponent from '../animation/AnimatedScaleComponent';
 
 const styles = StyleSheet.create({
   wrapper: {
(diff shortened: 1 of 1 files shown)


borisyankov:
We use the `ComponentWithOverlay` component in 3 places currently. Two of them show an unread count, but the third is the `AvatarItem` which shows the `IconCancel`. I don't know if it looks good there. Also, we might want to keep this more generic, as it will likely be used with other types of overlays.

One option is to make the animating optional. Another is to creae an animated overlay... But if it looks good on `AvatarItem` I am fine to make all overlays animated. Could you test this?

gnprice:
@jainkuniya Would you try to make a quick screen-capture video of this animation? I think that would be the most helpful thing for reviewing it :slightly_smiling_face:

gnprice:
I'm not convinced we would want this UI change. Certainly it seems fine for us to continue to go without it.

If we did choose to make this change, I'd want to be sure to do it in a way that addressed the same concern Boris raised at the outset above:

> We use the `ComponentWithOverlay` component in 3 places currently. Two of them show an unread count, but the third is the `AvatarItem` which shows the `IconCancel`. I don't know if it looks good there. Also, we might want to keep this more generic, as it will likely be used with other types of overlays.

That is, either we would decide we really do want this on *all* cases of overlaying one thing on another (unlikely), and then make the change explicitly about that and not just unread counts; or we would rearrange how it works in the code so that we could choose specifically to do it for unread counts, and not for presence dots on avatars.

I don't understand this explanation of why this PR has no effect on the presence dots we show in AvatarItem:

> Because for Avatarltem, `AnimatedScaleComponent` never goes to `componentWillReceive` Props, props never changes. It only renders one time.

but even if it's true that it currently doesn't have such an effect, that's a fragile fact about the current (as of this PR) set of places we use `ComponentWithOverlay`. We might use it in different places in the future -- it has such a generic name, after all -- and so we should not have logic inside it that's secretly meant to be specific to one particular place we use it.

So if we do at some point choose to make this UI change, we'll want the implementation to be different from the approach in this PR, which I'm therefore closing.

PR #3090, opened 2018-10-30
Fix #3082. Highlight alert words in message content.

### LINKED ISSUE #3082: webView: Apply alert word css style for alert words in message list.
```
.alert-word {
  background-color: rgba(108,239,52,.5);
}
```

like web app does.

Current master
image

webapp
image

@zulipbot claim

jainkuniya:
Fix #3082. 

image

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -126,7 +126,7 @@ rules:
       avoider, octicons, centerer, ldap, gravatar, identicon, blueimp, filename, wildcards,
       jpeg, jpg, tif, mov, notif, diag, bmp, viewport, scalable, polyfill, rect, touchstart,
       touchend, touchmove, remoteuser, sso, submessages, nbsp, args, px, na, heic,
-      unstarred, gooooooooal, dogi]
+      unstarred, gooooooooal, dogi, ig]
 overrides:
   - files:
       - generatedEs3.js
--- a/src/message/MessageList.js
+++ b/src/message/MessageList.js
@@ -5,6 +5,7 @@ import React, { PureComponent } from 'react';
 import { connectActionSheet } from '@expo/react-native-action-sheet';
 
 import type {
+  AlertWordsState,
   Auth,
   Context,
   Debug,
@@ -52,6 +53,7 @@ export type OuterProps = {
 // TODO get a type for `connectActionSheet` so this gets fully type-checked.
 export type Props = {
   // From caller and/or `connect`.
+  alertWords: AlertWordsState,
   anchor: number,
   auth: Auth,
   debug: Debug,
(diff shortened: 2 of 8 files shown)


gnprice:
Thanks @jainkuniya !

Gosh, this logic looks really fragile. :worried: I see it's taken from the webapp. I'd like to avoid having to think hard about it and really be sure of its semantics, and instead just be able to as easily as possible look at it and say "this is exactly what the webapp does", so that there's definitely just one version of this very complicated logic to worry about.

So for example here's a command I'd like to be able to run, and have it assure me that the bulk of this code is totally identical:
```
git diff --no-index ../zulip/static/js/alert_words.js src/webview/html/processMessageContent.js
```
(Or
```
diff -u ../zulip/static/js/alert_words.js src/webview/html/processMessageContent.js
```
which is very similar.)

It's fine if the beginning and end of the file are different because of the different context -- but I'd like ideally every line after the `_.each(exports.words, function (word) {` and until the end of the function to be absolutely unchanged from the webapp version.

It looks like most of the differences are because of our style. This is a case where I think it makes sense to just turn off ESLint and Prettier. You can do that with comments like this:
https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments
https://prettier.io/docs/en/ignore.html#javascript

Let's also have a comment in the code saying this is taken from `static/js/alert_words.js` of the webapp.

Naming: this name is too general. The full name of this function in the webapp is `alert_words.process_message`, which is appropriate. Let's call this one... `processAlertWords`.

gnprice:
Thanks @jainkuniya !

This command I mentioned (but with the name updated) still shows a lot of differences:
```
$ git diff --no-index ../zulip/static/js/alert_words.js src/webview/html/processAlertWords.js
[... almost the whole file ...]
```

Would you try to make all of the code that has complicated logic in it match exactly? In particular, I'd ideally like every line after the `_.each(exports.words, function (word) {` and until the end of the function to be absolutely unchanged from the webapp version.

Also the `escape_user_regex` function, now that I look again.

gnprice:
Thanks again @jainkuniya !

This looks good. :smile: I'm merging, after having made a few changes. Some points of feedback:
* Tests generally go in the same commit as the thing they test. I've regrouped your changes as one commit adding `processAlertWords`, its tests, the CSS for the class it introduces, and the spellcheck change it requires; then a second commit adding the `processAlertWords` call site and the data flow down to it.
* How did you choose the color `rgba(108,239,52,.5)`? It looks like the webapp has `hsla(102, 85%, 57%, 0.5)`. Looks like yours is approximately the same in different units? Better to keep the same expression as in the webapp, I think.
* I made a few further refactors to make the copied code even closer to the webapp, and to simplify the code around it a bit; take a look.

gnprice:
BTW, seeing the number of layers you had to add a mention of `alertWords` to in order to pass that data down to the spot where you needed it here inspired me to go refactor some of that code... which led me step by step into a major refactoring of the message list code 🧹, which I hope will make things like this simpler in the future among other benefits. Take a look: #3156

PR #3093, opened 2018-10-31
Improve icon consistency

borisyankov:
* enables correct default icon on Android (non round) (release & debug)
* provides round icons (release & debug)
* makes the iOS app use the same icon


gnprice:
(#2993 is the previous PR on this subject)

gnprice:
Thanks again @borisyankov for your work on this!

The iOS commit looks great and I'd be glad to merge all those changes -- but I would like to record what that source image was. :smile: So that we don't re-create for our future selves / successors the same kind of messy situation we've found ourselves given by the past.

I think you mentioned a Sketch file that you made -- would you add that to the iOS commit, or as a separate commit?

On the Android side, I'm still not convinced by the switch from a round to a square icon. I would be glad to take a commit that fixed the debug icon, though; and I don't have strong preferences on exactly what that would look like.

Issue #2483, opened 2018-05-09
Visual jerkiness/rerendering of the message view after sending a message


timabbott:
More than half the time, if I send a message to a topic narrow I'm caught up on, the mobile app has the following sequence:
* I hit send, we hide compose, and the message appears (with a loading indicator in the timestamp slot)
* The message finishes sending, and the loading indicator disappears
* We get the message from the server, and it does a jerky, very visual rerender, that seems to bounce the screen up and down a bit.

borisyankov:
Yeah, I've noticed that for sure. It might be unavoidable in some cases (a difference between client and server rendering), but I think in the majority of cases it should be avoidable.

I think now that the loading and positioning issues are mostly resolved, this is important enough to tackle next?

timabbott:
Yeah, I think there's two things we can do with this:
* You can just compare the content between client and server and only do a rerender if they're different; that's 90% of what we do in the webapp
* We could get smarter about the way we execute the rerendering, to avoid having the scroll position bounce around.  The Zulip webapp certainly does this every time something is edited without a visual glitch, and since it's a webview in mobile see any reason we should have a problem doing that on mobile.

borisyankov:
@timabbott this is our 'outbox' feature where we render the message when you send it vs the message got from the server. I think the spinner might also be changing the widths. Might not be too complicated to fix.

timabbott:
Yeah, we do the exact same thing in the webapp (it's called "local echo" there; tons of code in `echo.js` for it).

PR #3094, opened 2018-10-31
Improve outbox handling

borisyankov:
Changes one of the intermediate steps when sending messages.
Fixes message list jumping and 'messages disappearing for a while'.

Our current outbox flow is:
1. Add outbox item and render it
2. Outbox item is sent and removed from outbox
3. New message given back by a server and we render

The problem is that at step 2 the message disappears.
On a fast network, this appears as a fast jump in the message list,
not ideal. On a slow network, this appears as a prolonged and buggy
looking 'message disappears after sending'.

Now we:
1. Add outbox item and render it
2. Send but do not remove from the list, instead mark as `isSent` to
prevent a duplicate if sending needs to be restarted
3. New message received, remove from outbox

### DIFF
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -54,7 +54,8 @@ export const trySendMessages = () => (dispatch: Dispatch, getState: GetState) =>
   }
   dispatch(toggleOutboxSending(true));
   const auth = getAuth(state);
-  state.outbox.forEach(async item => {
+  const outboxToSend = state.outbox.filter(outbox => !outbox.isSent);
+  outboxToSend.forEach(async item => {
     try {
       await sendMessage(
         auth,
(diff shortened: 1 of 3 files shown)


gnprice:
Awesome. Thanks @borisyankov for fixing this! Merged.

I added a reference to #2483, which is the messages-jerk-around issue that it fixes.

Looking at this code while reviewing, I discovered this logic in the delete case:
```
filterArray(state, item => item && item.timestamp !== +action.local_message_id)
```
and we looked together at it quickly and found the corresponding logic in `outboxActions` -- we generate both `id` and `timestamp` to equal the time (in whole seconds!), and we actually use `item.timestamp` as the `local_id` sent to the server.

The logic in your new code definitely seems like the right way for this to be:
```
item.id !== action.local_message_id
```
So because you tested it and it works, I went ahead and merged as is. A good followup will be to fix how we generate local IDs to match this new code, making them (a) conceptually separate from `timestamp`, and (b) reliably unique (e.g. a local counter -- see the webapp's `sent_messages.get_new_local_id`, which generates strings `loc-1`, `loc-2`, etc.)

PR #3095, opened 2018-11-01
Unify the two platform's ComposeBox code

### LINKED ISSUE #2434: Chinese and Japanese input don't work on iOS (blocked on RN v0.57)

### LINKED ISSUE #3017: Make ComposeBox uncontrolled on iOS too, and dedupe code
We duplicated this code as a temporary expedient in cb2589e2a (from #2738), because we needed to make the message and topic inputs uncontrolled on Android but leave them controlled on iOS, in order to simultaneously avoid two different React Native bugs that affected the two platforms respectively. See the above-mentioned commit and PR for details.

As we've known since then, we should de-duplicate them; we want uncontrolled inputs on both platforms. #2886 was one PR to do this, but its clever workaround for the RN bug that had blocked us the first time turned out to not always succeed.

Fortunately, RN 0.57 has a fix for that bug. So once that upgrade is done, this should be easy. See #2788 for the 0.56 upgrade, which looks like most of the work.

### LINKED ISSUE #3053: Autocorrect on iOS Compose Input
Autocorrect does not behave as expected on the iOS compose message input. Basically this is very easy to reproduce. See the video for a comparison of behavior on iMessage and Zulip.

Ideally Zulip would be able to autocorrect in the middle of a sentence like iMessage does.

https://streamable.com/8w70j
https://streamable.com/glcz7

If you notice at the end of the second video the cursor jumps to the end. That happened when I tried to type a character it wasn't me moving the cursor manually. It seems to lose its place ( this is not as easily repro-able as the autocorrect problem and may be related to #3052 )

Seen on on iOS 18.0.99

borisyankov:
Closes #3017. Fixes #2434. Fixes #3053.

Remove the iOS code which was our old 'controlled' version and
keep the new Android 'uncontrolled' version.

While iOS does not have the extreme performance issues the Android
app had because of bugs in React Native, it is still faster to not
have the component controlled - a controlled component would block
any time the UI thread blocks.

Another big benefit is removing the need to support two versions.


gnprice:

I believe #2434 was actually fixed by the RN v0.57 upgrade earlier today -- but the difference doesn't matter, of course. Glad they're all fixed. :slightly_smiling_face:

Issue #2956, opened 2018-09-12
Fetch emoji code/codepoints from server and store under realm and use it instead of using static codes.


jainkuniya:
Related discussion: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Emoji.20sync

@zulipbot claim

gnprice:
One example mismatch I noticed while fixing #3129 (described in a9f2666d6):

```
There's still a bug that remains, which is that we think `:0:`
etc. are names for the 10 single-digit emoji (in addition to the
spelled-out `￼:zero:`), but the server seems to disagree; as a result
autocomplete and adding a reaction don't work.
```

gnprice:
This is a followup to #2868.

Doing this is expected to resolve #3128.

gnprice:
Marking this as a priority. See https://github.com/zulip/zulip-mobile/issues/3128#issuecomment-802289882 , and the linked [chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/emoji.20list/near/1135704).

chrisbobbe:
We should do another ad-hoc fix for the `:slight_smile:` to `:smile` change, then remove the P1 label.

WesleyAC:
Removing P1, since I filed #4640. It's not merged yet, but I'd rather not forget to remove the label, so just doing it now.

gnprice:
> Removing P1, since I filed #4640. It's not merged yet, but I'd rather not forget to remove the label, so just doing it now.

It's all come out the same because #4640 is merged now, but for similar things in the future: if something's P1 to do, I'd like to make sure there's something open for it that has a P1 label until it's actually done. Forgetting to remove a P1 label is a much lower risk than forgetting to do the thing, because the P1 label itself keeps it visible as a thing we should take care of.

In this instance, one good option would have been to add a P1 label on #4640, and then remove it on this issue.

WesleyAC:
Bumping the priority on this again, since we expect to do a big emoji update on the server side soon (:crossed_fingers:), and it'll be very awkward to not have this then.

chrisbobbe:
Wordle is broken on mobile on CZO 😛; see https://chat.zulip.org/#narrow/stream/97-off-topic/topic/Wordle/near/1322606

chrisbobbe:
Just filed https://github.com/zulip/zulip/issues/21037; marking as blocked on that.

gnprice:
[...]

PR #3100, opened 2018-11-01
Issue #2956. Fetch codePointMap from server and use it

jainkuniya:
Issue #2956.


src/realm/realmReducers.js
```diff
@@ -26,13 +28,16 @@ import {
...
+import { codePointMap } from '../emoji/codePointMap';
+
 // Initial state
 const initialState = {
   canCreateStreams: true,
   crossRealmBots: [],
   twentyFourHourTime: false,
   pushToken: { token: '', result: '', msg: '' },
   emoji: {},
+  emojiNameToCodePoint: codePointMap,
```
jainkuniya: I am confused for this name:

* emojiNameToCodePoint
* emojiNameToCodePointMap
* codePointMap
* codePointEmojiMap
* any other

In this PR I have mostly used `codePointMap`

Thoughts?

@zulip/react-native-committers

jainkuniya: > What's the type of it intended to be again?

It will be an object same as the current one, hard coded in the app `codePointMap`
```
{
  '+1': '1f44d',
  '-1': '1f44e',
  '0': '0030-20e3',
  '1': '0031-20e3',
.....
};
```

borisyankov: We started giving these a bit more specific types.
This looks like a `{ [string]: string }`

src/api/emoji/getEmojiNameToCodePoint.js
```diff
@@ -0,0 +1,6 @@
+/* @flow strict-local */
+import type { Auth } from '../apiTypes';
+import { apiGet } from '../apiFetch';
+
+export default async (auth: Auth): Promise<string[]> =>
+  apiGet(auth, 'static/generated/emoji/name_to_codepoint.json', res => res, undefined, false, true);
```
andersk, line 6: Note that while this file still exists on some Zulip servers as a leftover from previous deployments, it’s no longer being updated as of zulip/zulip#13831. There’s now a static/generated/emoji/emoji_codes.json instead. (We don’t guarantee this as part of the public API though.)

PR #3105, opened 2018-11-02
Fix a performance issue with some lists

### LINKED ISSUE #3103: Ensure that we do size FlatLists and SectionLists correctly
Both `FlatList` and `SectionList` offer as the main performance benefit rendering a 'window' and not their whole contents. Sometimes in the past, by mistake, we have rendered one of these in full height but scrollable thanks to the `ScrollView` they are inside. We should review all such components and make sure that their `height` is constrained.

borisyankov:
Closes #3103

Both FlatList and SectionList offer as the main performance benefit
rendering a 'window' and not their whole contents. Sometimes in the
past, by mistake, we have rendered one of these in full height but
scrollable thanks to the ScrollView they are inside.

By analyzing all of our usage, I pinpointed the only major problem
we have, that resulted in a performance issues in the Emoji picker
and the `UsersScreen` (when one searches the 'all users' list),
which manifested as the list updating too infrequently during search.

Our default `Screen` behavior is to be 'scrollable' by default.
We should consider changing the default behavior. In the meantime
any screen that has a `FlatList` or `SectionList` directly inside
should switch to `contentScrollable={false}`.

### DIFF
--- a/src/chat/GroupDetailsScreen.js
+++ b/src/chat/GroupDetailsScreen.js
@@ -33,7 +33,7 @@ class GroupDetailsScreen extends PureComponent<Props> {
     const { recipients } = navigation.state.params;
 
     return (
-      <Screen title="Recipients">
+      <Screen title="Recipients" scrollEnabled={false}>
         <FlatList
           initialNumToRender={10}
           data={recipients}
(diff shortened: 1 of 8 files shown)


gnprice:
Some of these say `scrollableContent` -- should they all be `contentScrollable`?

borisyankov:
They should actually be `scrollEnabled` as the just merged PR renamed them.
I did the fix just now.

gnprice:
Great. :slightly_smiling_face: Merged -- thanks again!

PR #3109, opened 2018-11-04
Reactivate Reactotron

borisyankov:
Refresh our config and use of Reactotron in order to improve our
ability to debug the app.

This:
 * Updates dependencies to the latest Reactotron version
 * Simplifies the config, some of the `use` lines were not needed as
`useReactNative` already enabled them
 * enable again, but only for __DEV__


gnprice:
Thanks @borisyankov !

I've merged the first two commits here. For the third, as you mentioned just now, let's switch to a conditional `require` inside the `if (__DEV__)`.

Issue #2252, opened 2018-04-08
feature request: See who reacted with an emoji.


rishig:
Slack's implementation (long press the emoji) seems reasonable here?

jainkuniya:
I thinks we had a open issue for this one (no matter, can't find it right now)
FB messenger also handle this nicely.

@zulipbot claim

borisyankov:
I am adding it to 'Next' as it has already been requested several times, seems important enough.

pulkonet:
Requested as a part in #1555

jmandel:
just adding a perspective that having this functionality would be really useful! Our FHIR community group often uses emoji responses to coordinate things like... who is coming to dinner at a conference. Really helps to be able to check on that list when we're waiting for folks to arrive in the hotel lobby and trying to decide when we have quorum to walk to a restaurant. It has been surprisingly frustrating  not to be able to answer this question on mobile (native or web).

gnprice:
Thanks @jmandel for that extra detail! Definitely helps for making concrete why this will be a valuable feature.

gnprice:
With #3110, this is now implemented! :tada: It'll be in the next release.

(Well, I'm about to promote 26.16.139 from beta, where it hung out for a while due to the Thanksgiving holiday; but this feature will be in the next new release from master.)

jainkuniya:
Excited to see it in the release 😍

gnprice:
@jmandel @Vamshi99 and everyone: 26.17.140 is now out, with this feature!

It's available now on Android; on iOS it's out in beta, and is in Apple's review queue for the App Store.

jmandel:
Just tried it out on Android and it works beautifully :-) Thanks so much.

Vamshi99:
@gnprice Looks good. But some emojis don't show up correctly in the reacted people screen, they are just seen as a question mark for me.
https://imgur.com/a/vGWMtXM

jainkuniya:
On which device are you?

Vamshi99:
@jainkuniya Redmi Note 5 pro, Android 9

gnprice:
Thanks @Vamshi99 for that report! I've gone and filed #3716 to continue the discussion.

Issue #3092, opened 2018-10-31
message list: Show users associated with reactions.


jackrzhang:
This is a feature request! (for the longer term)

On the web application, users can see the users associated with a reaction via text that appears after hovering over the reaction.

On Slack's mobile application, a list of users associated with reactions to a message slides up from the bottom of the screen after a long press on one of the reactions to the message.

(@zulipbot label "area: message list")

jainkuniya:
(#2252)

timabbott:
Yeah, I'd really love to see this too; it's one of the few places where the mobile app just doesn't offer you information you might need to interpret message content.

jainkuniya:
(on it)

gnprice:
Closing as a duplicate of #2252 .

PR #3110, opened 2018-11-04
Issue #2252. feature request: See who reacted with an emoji.

jainkuniya:
Issue #2252.

Discussion: https://chat.zulip.org/#narrow/stream/48-mobile/topic/reaction.20recipients

image

image

image

image

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -7,6 +7,7 @@ import { isTopicMuted } from '../utils/message';
 import * as api from '../api';
 import { showToast } from '../utils/info';
 import { doNarrow, startEditMessage, deleteOutboxMessage, navigateToEmojiPicker } from '../actions';
+import { navigateToMessageReactionScreen } from '../nav/navActions';
 
 // TODO really this belongs in a libdef.
 export type ShowActionSheetWithOptions = (
@@ -110,6 +111,11 @@ const addReaction = ({ message, dispatch }) => {
 };
 addReaction.title = 'Add a reaction';
 
+const showReactions = ({ message, dispatch }) => {
+  dispatch(navigateToMessageReactionScreen(message.id));
+};
+showReactions.title = 'See who reacted';
+
 const cancel = params => {};
 cancel.title = 'Cancel';
 
@@ -123,6 +129,7 @@ const allButtonsRaw = {
   deleteMessage,
   starMessage,
   unstarMessage,
+  showReactions,
 
   // For headers
   unmuteTopic,
@@ -179,6 +186,9 @@ export const constructMessageActionButtons = ({
   narrow,
 }: ConstructSheetParams): ButtonCode[] => {
   const buttons = [];
+  if (message.reactions.length > 0) {
+    buttons.push('showReactions');
+  }
   if (!isAnOutboxMessage(message) && messageNotDeleted(message)) {
     buttons.push('addReaction');
   }
--- a/src/nav/navActions.js
+++ b/src/nav/navActions.js
@@ -96,8 +96,11 @@ export const navigateToStreamSubscribers = (streamId: number): NavigationAction 
 export const navigateToNotifications = (): NavigationAction =>
   StackActions.push({ routeName: 'notifications' });
 
-export const navigateToMessageReactionScreen = (messageId: number): NavigationAction =>
-  StackActions.push({ routeName: 'message-reactions', params: { messageId } });
+export const navigateToMessageReactionScreen = (
+  messageId: number,
+  reactionName?: string,
+): NavigationAction =>
+  StackActions.push({ routeName: 'message-reactions', params: { messageId, reactionName } });
 
 export const navigateToLegal = (): NavigationAction => StackActions.push({ routeName: 'legal' });
 
--- a/src/reactions/MessageReactionList.js
+++ b/src/reactions/MessageReactionList.js
@@ -20,6 +20,7 @@ import { objectFromEntries } from '../jsBackport';
 // reactions on the message.
 const getReactionsTabs = (
   aggregatedReactions: $ReadOnlyArray<AggregatedReaction>,
+  reactionName?: string,
   allUsersById: Map<number, UserOrBot>,
 ) => {
   // Each tab corresponds to an aggregated reaction, and has a user list.
@@ -42,7 +43,7 @@ const getReactionsTabs = (
     ]),
   );
 
-  return createMaterialTopTabNavigator(reactionsTabs, {
+  let navigatorConfig = {
     backBehavior: 'none',
     ...tabsOptions({
       showLabel: true,
@@ -51,7 +52,16 @@ const getReactionsTabs = (
         borderWidth: 0.15,
       },
     }),
-  });
+  };
+
+  // there can be cases where reaction have been removed by the time user see this screen
+  // or on re-render
+  // so confirm reaction `reactionName` still exists
+  if (reactionName !== undefined && reactionsTabs[reactionName]) {
+    navigatorConfig = { ...navigatorConfig, initialRouteName: reactionName };
+  }
+
+  return createMaterialTopTabNavigator(reactionsTabs, navigatorConfig);
 };
 
 type SelectorProps = $ReadOnly<{|
@@ -61,7 +71,7 @@ type SelectorProps = $ReadOnly<{|
 |}>;
 
 type Props = $ReadOnly<{|
-  navigation: NavigationScreenProp<{ params: {| messageId: number |} }>,
+  navigation: NavigationScreenProp<{ params: {| reactionName?: string, messageId: number |} }>,
 
   dispatch: Dispatch,
   ...SelectorProps,
@@ -69,7 +79,8 @@ type Props = $ReadOnly<{|
 
 class MessageReactionList extends PureComponent<Props> {
   render() {
-    const { message, ownUserId, allUsersById } = this.props;
+    const { message, navigation, ownUserId, allUsersById } = this.props;
+    const { reactionName } = navigation.state.params;
     const { reactions } = message;
 
     // useful when user is on this screen and reactions are revoked
@@ -85,7 +96,7 @@ class MessageReactionList extends PureComponent<Props> {
 
     const aggregatedReactions = aggregateReactions(reactions, ownUserId);
 
-    const TabView = getReactionsTabs(aggregatedReactions, allUsersById);
+    const TabView = getReactionsTabs(aggregatedReactions, reactionName, allUsersById);
 
     return (
       <Screen title="Reactions" scrollEnabled={false}>
--- a/src/reactions/MessageReactionList.js
+++ b/src/reactions/MessageReactionList.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
-import { TabNavigator, TabBarTop } from 'react-navigation';
+import { createMaterialTopTabNavigator } from 'react-navigation';
 import type { NavigationScreenProp } from 'react-navigation';
 
 import ReactionUserList from './ReactionUserList';
@@ -42,10 +42,8 @@ const getReactionsTabs = (
     ]),
   );
 
-  return TabNavigator(reactionsTabs, {
+  return createMaterialTopTabNavigator(reactionsTabs, {
     backBehavior: 'none',
-    tabBarComponent: TabBarTop,
-    tabBarPosition: 'top',
     ...tabsOptions({
       showLabel: true,
       showIcon: false,
--- a/src/reactions/MessageReactionList.js
+++ b/src/reactions/MessageReactionList.js
@@ -43,8 +43,17 @@ const getReactionsTabs = (
     ]),
   );
 
-  let navigatorConfig = {
+  // prettier-ignore
+  return createMaterialTopTabNavigator(reactionsTabs, {
     backBehavior: 'none',
+
+    // there can be cases where reaction have been removed by the time user see this screen
+    // or on re-render
+    // so confirm reaction `reactionName` still exists
+    ...(reactionName !== undefined && reactionsTabs[reactionName]
+      ? { initialRouteName: reactionName }
+      : {}),
+
     ...tabsOptions({
       showLabel: true,
       showIcon: false,
@@ -52,16 +61,7 @@ const getReactionsTabs = (
         borderWidth: 0.15,
       },
     }),
-  };
-
-  // there can be cases where reaction have been removed by the time user see this screen
-  // or on re-render
-  // so confirm reaction `reactionName` still exists
-  if (reactionName !== undefined && reactionsTabs[reactionName]) {
-    navigatorConfig = { ...navigatorConfig, initialRouteName: reactionName };
-  }
-
-  return createMaterialTopTabNavigator(reactionsTabs, navigatorConfig);
+  });
 };
 
 type SelectorProps = $ReadOnly<{|
--- a/src/reactions/MessageReactionList.js
+++ b/src/reactions/MessageReactionList.js
@@ -47,9 +47,8 @@ const getReactionsTabs = (
   return createMaterialTopTabNavigator(reactionsTabs, {
     backBehavior: 'none',
 
-    // there can be cases where reaction have been removed by the time user see this screen
-    // or on re-render
-    // so confirm reaction `reactionName` still exists
+    // The user may have originally navigated here to look at a reaction
+    // that's since been removed.  Ignore the nav hint in that case.
     ...(reactionName !== undefined && reactionsTabs[reactionName]
       ? { initialRouteName: reactionName }
       : {}),
@@ -77,6 +76,12 @@ type Props = $ReadOnly<{|
   ...SelectorProps,
 |}>;
 
+/**
+ * A screen showing who made what reaction on a given message.
+ *
+ * The `reactionName` nav-prop controls what reaction is focused when the
+ * screen first appears.
+ */
 class MessageReactionList extends PureComponent<Props> {
   render() {
     const { message, navigation, ownUserId, allUsersById } = this.props;
--- a/src/reactions/ReactionUserList.js
+++ b/src/reactions/ReactionUserList.js
@@ -13,6 +13,11 @@ type Props = $ReadOnly<{|
   allUsersById: Map<number, UserOrBot>,
 |}>;
 
+/**
+ * Component showing who made a given reaction on a given message.
+ *
+ * Used within `MessageReactionList`.
+ */
 class ReactionUserList extends PureComponent<Props> {
   handlePress = (email: string) => {
     const { dispatch } = this.props;
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -578,6 +578,15 @@ var compiledWebviewJs = (function (exports) {
 
   var handleLongPress = function handleLongPress(target) {
     hasLongPressed = true;
+
+    if (target.closest('.reaction')) {
+      sendMessage({
+        type: 'reactionDetails',
+        messageId: getMessageIdFromNode(target)
+      });
+      return;
+    }
+
     sendMessage({
       type: 'longPress',
       target: target.matches('.header') ? 'header' : target.matches('a') ? 'link' : 'message',
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -578,11 +578,13 @@ var compiledWebviewJs = (function (exports) {
 
   var handleLongPress = function handleLongPress(target) {
     hasLongPressed = true;
+    var reactionNode = target.closest('.reaction');
 
-    if (target.closest('.reaction')) {
+    if (reactionNode) {
       sendMessage({
         type: 'reactionDetails',
-        messageId: getMessageIdFromNode(target)
+        messageId: getMessageIdFromNode(target),
+        reactionName: requireAttribute(reactionNode, 'data-name')
       });
       return;
     }
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -701,10 +701,12 @@ const handleLongPress = (target: Element) => {
 
   hasLongPressed = true;
 
-  if (target.closest('.reaction')) {
+  const reactionNode = target.closest('.reaction');
+  if (reactionNode) {
     sendMessage({
       type: 'reactionDetails',
       messageId: getMessageIdFromNode(target),
+      reactionName: requireAttribute(reactionNode, 'data-name'),
     });
     return;
   }
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -701,6 +701,14 @@ const handleLongPress = (target: Element) => {
 
   hasLongPressed = true;
 
+  if (target.closest('.reaction')) {
+    sendMessage({
+      type: 'reactionDetails',
+      messageId: getMessageIdFromNode(target),
+    });
+    return;
+  }
+
   sendMessage({
     type: 'longPress',
     target: target.matches('.header') ? 'header' : target.matches('a') ? 'link' : 'message',
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -105,6 +105,7 @@ type MessageListEventError = {|
 type MessageListEventReactionDetails = {|
   type: 'reactionDetails',
   messageId: number,
+  reactionName: string,
 |};
 
 export type MessageListEvent =
@@ -236,9 +237,9 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
 
     case 'reactionDetails':
       {
-        const { messageId } = event;
+        const { messageId, reactionName } = event;
         const { dispatch } = props;
-        dispatch(navigateToMessageReactionScreen(messageId));
+        dispatch(navigateToMessageReactionScreen(messageId, reactionName));
       }
       break;
 
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -14,6 +14,7 @@ import {
   fetchOlder,
   fetchNewer,
   navigateToAccountDetails,
+  navigateToMessageReactionScreen,
   doNarrow,
   navigateToLightbox,
   messageLinkPress,
@@ -101,6 +102,11 @@ type MessageListEventError = {|
   },
 |};
 
+type MessageListEventReactionDetails = {|
+  type: 'reactionDetails',
+  messageId: number,
+|};
+
 export type MessageListEvent =
   | MessageListEventReady
   | MessageListEventScroll
@@ -110,6 +116,7 @@ export type MessageListEvent =
   | MessageListEventReaction
   | MessageListEventUrl
   | MessageListEventLongPress
+  | MessageListEventReactionDetails
   | MessageListEventDebug
   | MessageListEventError;
 
@@ -167,6 +174,7 @@ const handleLongPress = (
     showToast(_('Link copied to clipboard'));
     return;
   }
+
   const message = props.messages.find(x => x.id === messageId);
   if (!message) {
     return;
@@ -223,6 +231,14 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
         } else {
           api.emojiReactionAdd(auth, messageId, reactionType, code, name);
         }
+      }
+      break;
+
+    case 'reactionDetails':
+      {
+        const { messageId } = event;
+        const { dispatch } = props;
+        dispatch(navigateToMessageReactionScreen(messageId));
       }
       break;
 
--- a/static/translations/messages_en.json
+++ b/static/translations/messages_en.json
@@ -163,5 +163,6 @@
   "Clear": "Clear",
   "Reactions": "Reactions",
   "No reactions": "No reactions",
+  "See who reacted": "See who reacted",
   "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
 }
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -24,6 +24,7 @@ describe('constructActionButtons', () => {
   test('show star message option if message is not starred', () => {
     const message = deepFreeze({
       id: 3,
+      reactions: [],
     });
 
     const buttons = constructMessageActionButtons({
@@ -38,6 +39,7 @@ describe('constructActionButtons', () => {
   test('show unstar message option if message is starred', () => {
     const message = deepFreeze({
       id: 1,
+      reactions: [],
     });
 
     const buttons = constructMessageActionButtons({
@@ -47,6 +49,21 @@ describe('constructActionButtons', () => {
     });
 
     expect(buttons).toContain('unstarMessage');
+  });
+
+  test('show reactions option if message is has at least one reaction', () => {
+    const message = deepFreeze({
+      id: 1,
+      reactions: [{}],
+    });
+
+    const buttons = constructMessageActionButtons({
+      backgroundData: { auth, flags, mute, subscriptions },
+      message,
+      narrow,
+    });
+
+    expect(buttons).toContain('showReactions');
   });
 });
 


jackrzhang:
Referencing #3092.

borisyankov:
That's a lot of 'WIP's ;)
I'll give you a quick first review then 👍

src/webview/js/js.js
```diff
@@ -618,7 +618,18 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
   }
 });
 
-const getHtmlNodeType = (target: Element) => target.matches('.header') ? 'header' : 'message';
+const getHtmlNodeType = (target: Element) => {
+  if (target.matches('.header')) {
+    return 'header';
+  } else if (
+    target.matches('.reaction-list')
```
borisyankov: Examine how we use `element.closest('body > *');` for similar purposes and think if you can do the same here.

src/reactions/ReactionListItem.js
```diff
@@ -0,0 +1,42 @@
...
+
+  render() {
+    const { styles } = this.context;
+    const { name, users } = this.props.aggregatedReaction;
+    const usersNames = users.map((user: NamedUser) => user.full_name).join(', ');
+
+    return (
+      <View style={[styles.reactionListItemWrapper]}>
+        <View style={[styles.row]}>
+          <Emoji name={name} size={30} />
```
borisyankov: 30 => 32 size?

src/reactions/ReactionListItem.js
```diff
@@ -0,0 +1,42 @@
...
+          <View style={[styles.halfMarginLeft, styles.flexed]}>
```
borisyankov: Do you need two `Views` inside each other here?
Also, add more space around the icon, twice what it is (so not 'half' :) )

src/directSelectors.js
```diff
@@ -91,3 +92,6 @@ export const getRawRealmEmoji = (state: GlobalState): RealmEmojiState => state.r
 export const getNonActiveUsers = (state: GlobalState): User[] => state.realm.nonActiveUsers;
 
 export const getIsAdmin = (state: GlobalState): boolean => state.realm.isAdmin;
+
+export const getMessageFromId = (state: GlobalState, messageId: number): Message =>
```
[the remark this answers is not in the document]
gnprice: I agree with Boris on the name.

Let's also move this inside `messageSelectors`. It's the business of those selectors and the corresponding reducer how this data is laid out.

gnprice:
Thanks @jainkuniya for building this! And for the screenshots :smiley: 

Quick feedback on the UI in the screenshots: let's sort by number of people reacting, most people first.

Also I think the list of names as text gets pretty hard to scan at even 3-4 people -- especially with the paragraph-style wrapping. I'd rather have something more structured, and with the users' avatars.

Two ideas for how to do that:
* Facebook's UI for reactions on a post (FB proper, not Messenger, though maybe the latter is the same). I'm not immediately finding a handy screenshot of this on the web, but I'll take another couple of minutes shortly to look harder or make one. This UI works quite well in my experience.
* An alternative (but here I'm making up something I haven't seen a concrete implementation of) would be a vertical list like this, but each item is a horizontal list of just avatars (like Facebook does in lots of other contexts, I guess.) And when you tap an item to expand, it turns into a vertical list of avatar + name. This idea would need more iterating on, I think.

I haven't looked closely through the code at this stage, since you have a number of WIP commits and also an early review from Boris :slightly_smiling_face:

gnprice:
> Facebook's UI for reactions on a post

Ah here we are. This image looks to be desktop web, but the layout in the mobile app is very similar (to the window in the foreground):

Also it's from 2016, but looks familiar from when I've seen this UI recently.

Those are tabs at the top: "All", then each reaction in turn, sorted by number of people. I'd be happy even without the "All" tab. Each tab has a vertical list of people, with avatar + name.

gnprice:
Oh, and a question about the UI (that's hard to show in a screenshot): how does the user get to this screen?

I like the suggestion of long-pressing on a reaction.

src/reactions/aggregateReactions.js
```diff
@@ -23,7 +23,7 @@ export default (reactions: EventReaction[], ownEmail: string): AggregatedReactio
           }
         }
 
-        if (x.user && x.user.email === ownEmail) {
+        if (x.user && x.user.email === ownEmail && ownEmail !== undefined) {
```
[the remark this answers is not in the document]
jainkuniya: It can.

But in any case if `ownEmail` is undefined we should to consider it for `selfVote`. :)

src/webview/webViewEventHandlers.js
```diff
@@ -161,6 +162,11 @@ export const handleImage = (props: Props, event: MessageListEventImage) => {
 
 export const handleLongPress = (props: Props, event: MessageListEventLongPress) => {
   const { messageId, target } = event;
+  const { dispatch } = props;
+  if (target === 'reaction-list') {
```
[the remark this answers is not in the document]
jainkuniya: Done 👍

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,35 @@
...
+        <Text>{JSON.stringify(message)}</Text>
+      </Screen>
+    );
+  }
+}
+
+export default connect((state: GlobalState, props) => ({
+  message: getMessageFromId(
+    state,
+    (props.navigation
```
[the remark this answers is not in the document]
jainkuniya: Done 👍

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,35 @@
...
+  message: Message,
+};
+
+class MessageReactionList extends PureComponent<Props> {
+  props: Props;
+
+  render() {
+    const { message } = this.props;
+    return (
+      <Screen title="Reaction List" scrollableContent={false}>
```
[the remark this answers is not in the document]
jainkuniya: Done 👍

src/directSelectors.js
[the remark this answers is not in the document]
jainkuniya: Done 👍

src/webview/js/js.js
jainkuniya: Done, now I am using `.closest` 👍

src/reactions/MessageReactionList.js
```diff
@@ -29,14 +29,17 @@ class MessageReactionList extends PureComponent<Props> {
     const { styles } = this.context;
 
     const reactionsTabs = aggregatedReactions.reduce((tabs, aggregatedReaction) => {
+      const { width } = Dimensions.get('window');
```
borisyankov: Remember that this is not orientation-change proof.

src/message/messageSelectors.js
```diff
@@ -18,3 +18,6 @@ export const getAnchorForActiveNarrow = (narrow: Narrow) =>
     getMute,
     (messages, flags, subscriptions, mute) => findAnchor(messages, flags, subscriptions, mute),
   );
+
+export const getMessageById = (state: GlobalState, messageId: number): Message =>
```
borisyankov: Not sure if we can call this a selector.
Also `state.messages[messageId]` seems simple and straightforward.
Why did you decide to extract it to a function?

src/reactions/aggregateReactions.js
```diff
@@ -23,7 +23,7 @@ export default (reactions: EventReaction[], ownEmail: string): AggregatedReactio
           }
         }
 
-        if (x.user && x.user.email === ownEmail) {
+        if (ownEmail && x.user && x.user.email && x.user.email === ownEmail) {
```
borisyankov: Can `ownEmail` be undefined?

src/reactions/aggregateReactions.js
```diff
@@ -43,4 +43,7 @@ export default (
         return reactionMap;
       }, new Map())
       .values(),
+  ).sort(
+    (r1: AggregatedReaction, r2: AggregatedReaction) =>
+      r1.users.length < r2.users.length ? 1 : -1,
```
borisyankov: The comparer function does not need a `1` or `-1` but any positive or negative value will work.
Also, you need to pass `0` if equal.

The pattern to use is  to just return `r2.users.length - r1.users.length`

src/styles/tabs.js
```diff
@@ -4,9 +4,10 @@ import { BRAND_COLOR } from './';
 type Props = {
   showLabel: boolean,
   showIcon: boolean,
+  style?: Object,
 };
 
-export default ({ showLabel, showIcon }: Props) => ({
+export default ({ showLabel, showIcon, style = {} }: Props) => ({
```
borisyankov: `style` does not need to have a default value.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,90 @@
...
+  static contextTypes = {
+    styles: () => null,
+  };
+
+  getReactionsTabs = (aggregatedReactions: AggregatedReaction[]) => {
+    const { styles } = this.context;
+
+    const reactionsTabs = aggregatedReactions.reduce((tabs, aggregatedReaction) => {
+      tabs[aggregatedReaction.code] = {
+        // $FlowFixMe TODO
```
borisyankov: `TODO` is not a good description :)

src/reactions/__tests__/aggregateReactions-test.js
```diff
@@ -1,6 +1,7 @@
 import deepFreeze from 'deep-freeze';
 
 import aggregateReactions from '../aggregateReactions';
+import { NULL_USER } from '../../nullObjects';
```
borisyankov: Maybe just omit deactivated users?

src/styles/tabs.js
jainkuniya: Done 👍

src/reactions/aggregateReactions.js
jainkuniya: Done 👍

src/reactions/__tests__/aggregateReactions-test.js
jainkuniya: I guess, for v1 we can go with this. And afterward we can handle  `deactivated users` case. :)

borisyankov:
I am seeing some WIP commits. Let me know when you think the work is ready for a review and I'll do both manual testing and code review. ;)

borisyankov:
@jainkuniya ping me here and in chat when you think it is no longer WIP. 👍

src/reactions/MessageReactionList.js
jainkuniya: Removed 👍

src/reactions/aggregateReactions.js
jainkuniya: May be if we don't pass to the function.

Actually I was writing tests, and found this :)

src/message/messageSelectors.js
jainkuniya: Because it looks/feels good to have similar selectors at same place. :)

borisyankov:
Started testing this...
First issue: users' avatars are missing!

src/message/messageSelectors.js
borisyankov: Can confirm, not a selector :)

borisyankov:
Apparently this line:
```js
message: getMessageById(state, props.navigation.state.params.messageId),
```
might return `undefined` which throws an exception.

borisyankov:
I had the app crash once on clicking the second emoji reaction tab!
Not sure how that happened.

borisyankov:
> ExceptionsManager.js:84 Warning: Encountered two children with the same key, `-1`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version

This happens in `ReactionUserList` and might be related to the missing avatars.

borisyankov:
I think the action to show it "long-press on empty space" is very undiscoverable.
A better alternative to that would be adding a circle, at the end of the list of reactions with an `i` inside of them (the same styling as `#scroll-bottom` but smaller radius). On pressing of it, you show the details.

src/webview/js/js.js
```diff
@@ -644,7 +656,7 @@ const handleLongPress = (target: Element) => {
 
   sendMessage({
     type: 'longPress',
-    target: target.matches('.header') ? 'header' : 'message',
+    target: getHtmlNodeType(target),
```
borisyankov: `getHtmlNodeType` sounds a bit incorrec. Maybe `getTouchTargetType`?

src/reactions/aggregateReactions.js
borisyankov: I would prefer to figure out why `ownEmail` is undefined and fix this separately. 
It might be causing other issues too.
Then remove this commit.

borisyankov: Actually by 'I would prefer' I meant 'the correct way to approach this' :)

src/reactions/aggregateReactions.js
```diff
@@ -1,19 +1,24 @@
 /* @flow strict-local */
-import type { Reaction, AggregatedReaction } from '../types';
+import type { Reaction, AggregatedReaction, User } from '../types';
+import { NULL_USER } from '../nullObjects';
 
 export default (
   reactions: $ReadOnlyArray<Reaction>,
   ownEmail: string,
+  users?: User[],
```
borisyankov: If you really want to have the option of not passing users do it with a default value, like:
```js
users: User[] = []
```

Even better, instead of `getUsers` use the `getUsersById` selector and then  instead of `users.find` you can use `users[...]` to access the user.

src/reactions/aggregateReactions.js
```diff
@@ -1,19 +1,24 @@
...
 ): $ReadOnlyArray<AggregatedReaction> =>
   Array.from(
     reactions
       .reduce((reactionMap, x) => {
+        const user = users && users.find((u: User) => u.user_id === x.user.user_id);
+        const reactedUser = user || { ...NULL_USER, ...x.user };
```
borisyankov: When will we be unable to find a valid `user`?

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,89 @@
...
+  getReactionsTabs = (aggregatedReactions: $ReadOnlyArray<AggregatedReaction>) => {
+    const reactionsTabs = aggregatedReactions.reduce((tabs, aggregatedReaction) => {
+      const { width } = Dimensions.get('window');
+      const emojiSize: number = Math.min(width / aggregatedReactions.length * 0.32, 24);
+      const countSize: number = Math.min(width / aggregatedReactions.length * 0.4, 16);
+      tabs[aggregatedReaction.code] = {
+        screen: () => <ReactionUserList reactedUsers={aggregatedReaction.users} />,
+        navigationOptions: {
+          tabBarLabel: () => (
+            <View style={[styles.row]}>
```
borisyankov: That does to have to be an array.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+class MessageReactionList extends PureComponent<Props> {
+  props: Props;
+
+  static contextTypes = {
+    styles: () => null,
+  };
+
+  getReactionsTabs = (aggregatedReactions: $ReadOnlyArray<AggregatedReaction>) => {
```
borisyankov: You can extract that as a separate React component.

borisyankov:
While I did a lot of comments on this PR, I think we are getting close.
If you manage to address the issues before the meeting, ping me so I can have another look.

src/reactions/aggregateReactions.js
jainkuniya: Actually as soon as I added `user: { user_id: 2 },` in reaction object in `every duplicate reaction is aggregated to a single one with appropriate count` it started failing. Because now `x.user` is present but, `x.user.email` is null and also we aren't passing `ownEmail` in this test, it started failing.

src/reactions/aggregateReactions.js
jainkuniya: When user is deactivated, I guess :)

src/reactions/aggregateReactions.js
```diff
@@ -43,4 +43,4 @@ export default (
         return reactionMap;
       }, new Map())
       .values(),
-  );
+  ).sort((r1: AggregatedReaction, r2: AggregatedReaction) => r2.users.length - r1.users.length);
```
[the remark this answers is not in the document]
jainkuniya: Moved this commit to last :)

src/styles/tabs.js
```diff
@@ -4,9 +4,10 @@ import { BRAND_COLOR } from './constants';
 type Props = {|
   showLabel: boolean,
   showIcon: boolean,
+  style?: {}, // match this with upstream type
```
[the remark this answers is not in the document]
jainkuniya: Upstream (react-navigation) also as this type as `{}`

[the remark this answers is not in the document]
borisyankov: That makes sense then.

borisyankov:
This is still relevant:
>I think the action to show it "long-press on empty space" is very undiscoverable.
A better alternative to that would be adding a circle, at the end of the list of reactions with an 'i' inside of them (the same styling as #scroll-bottom but smaller radius). On pressing of it, you show the details.

src/reactions/ReactionListTabView.js
```diff
@@ -0,0 +1,45 @@
...
+export default (aggregatedReactions: $ReadOnlyArray<AggregatedReaction>) => {
+  const reactionsTabs = aggregatedReactions.reduce((tabs, aggregatedReaction) => {
+    const { width } = Dimensions.get('window');
+    const emojiSize: number = Math.min(width / aggregatedReactions.length * 0.32, 24);
+    const countSize: number = Math.min(width / aggregatedReactions.length * 0.4, 16);
+    tabs[aggregatedReaction.code] = {
+      screen: () => <ReactionUserList reactedUsers={aggregatedReaction.users} />,
+      navigationOptions: {
+        tabBarLabel: () => (
+          <View style={[styles.row]}>
```
borisyankov: That does not need to be an array.

gnprice:
Thanks @jainkuniya for building this, and @borisyankov for the reviews! Because Boris's latest review is short, I figure it's time for me to take a look too. :slightly_smiling_face: 

* I'm pretty happy with the UX, from playing with it a bit. @borisyankov , I'm curious what you mean when you say "long-press on empty space" -- but I see the interaction in this version as "long-press on the reactions", and that feels pretty natural and discoverable to me. (It also works to long-press on the space around a reaction, which is basically a way of making the touch target large.)

* One nit in the UX: when I long-press on, say, the 3rd out of 4 different reactions, I'd ideally like that to take me straight to the 3rd one, instead of to the 1st one. That doesn't have to come before merging this, though.

* Realm emoji don't seem to work -- they show up as "?". (So does the Zulip extra emoji, but that's another matter.)

* The tab labels feel a bit crowded -- like I want a little more space between the emoji and the count.

* The reactions can show up in a different order from how they do in the message list, which is confusing. The two UIs should be in the same order -- and preferably the same order as in the webapp. I don't have a strong view on what orderings are preferable beyond that, though.

I think that's what I have from trying it out; now looking through the code.

gnprice:
> flow: Fix type of reaction in Messsage.

Hmm interesting! And as Steve said in that chat thread, an unfortunate mismatch. Thanks for catching this.

```
+export type MessageReaction = $ReadOnly<{|
+  ...$Exact<Reaction>,
+  user: $ReadOnly<{|
+    email: string,
+    full_name: string,
+    user_id?: number,
+    id: number,
+  |}>,
+|}>;
```

Why is there a `user_id?:` there -- is there really sometimes a `user_id` property?

... Ah I see, is that related to the discussion in chat there that we might migrate to sending `user_id`? That'd be good, but it should be explicit, and there are some more pieces required for doing that; let's make it a separate commit, and focus this commit just on accurately modeling what the server currently sends.

Also `ReactionAdd` isn't the best name, because "add" contrasts with "remove", and add and remove work exactly the same here. Something like "ReactionForEvent" would be better...

... but better still, let's keep the name `Reaction` for the version with `user_id`, because that's the preferred version that we intend to migrate to. We can call the `id` version `LegacyReaction`. The jsdoc on `LegacyReaction` can link to that chat thread.


```
-        user: action.user,
+        user: {
+          ...action.user,
+          id: action.user.user_id,
+        },
```

Let's follow @showell 's suggestion in the chat thread, and prefer `user_id` -- so this conversion will go in the opposite direction.

(Which means this bit of code won't change -- instead, where we consume the `id` version let's convert that to `user_id`.)

Another thought I have looking at this commit is: surely either the version just before this change or just after it is a type error. If they both pass Flow, then our types somewhere aren't being effective, and we should figure out why. :slightly_smiling_face: If one of them doesn't, then either the previous commit or this one broke things.

```
    refactor, webView, js: Extract logic to `getTouchTargetType`.
...

+/**
+ * Return html node type in Zulip context like header or message node
+ */
+const getTouchTargetType = (target: Element) => target.matches('.header') ? 'header' : 'message';
```

Hmm interesting. But this isn't really about a "touch target" in general, is it? This is specific to a long-press.

Right above this code, in the `click` listener, we handle ordinary touches, and we have a bunch of different cases for that where it might be an avatar, a "header" (aka recipient bar), a link, etc. -- including a reaction.

In fact, thinking more about this code and comparing with that code (the `click` listener): I think I'd like to structure this code similarly to that code. Instead of sending always the same `type` value and having a `target` property to distinguish further, let's have a distinct `type` value -- say `reactionDetails`. This will have another type in the `MessageListEvent` union, say `MessageListEventReactionDetails`, and the `switch` in `handleMessageListEvent` can go straight to `navigateToMessageReactionScreen`.


```
+        const reactedUser: User = users[x.user.id.toString()] || {
+          ...NULL_USER,
+          user_id: x.user.id,
+          full_name: x.user.full_name,
+          email: x.user.email,
+        };
```

Yikes -- what's all this doing?

The baseline here is to simply use `x.user`. Anything more complex than that, there needs to be a clear reason for.

There's the `id` / `user_id` issue -- that conversion has to happen somewhere, and here could be an OK place for it (though probably it should happen sooner than here, so the thing we get here already has `user_id`.)

The other thing this code seems to be doing is to pull in a bunch of garbage data, i.e. NULL_USER. Pretty much by definition, if we ever actually touch any of that data, we have a bug -- so if there's some other code that complains or fails when we leave it out, we should fix that code.

```
+  style?: {}, // match this with upstream type
```

I don't quite understand the comment -- is it saying this *does* match the upstream type, or that as a TODO item it'd be nice to come back and *change* it to match it with the upstream type?

Also it's not at all clear from looking at this file what "upstream" might be. From where it's used, the answer is "`TabNavigator` from `react-navigation`". That'd be good context to put into jsdoc just above the `export default` line -- that'd be a helpful improvement to the existing code here.


```
+  props: Props;
```

We don't need to write this anymore :smiley: -- see `git log --stat -S 'props: Props;'`.


```
+  static contextTypes = {
+    styles: () => null,
+  };
```

We don't need this in most places either :tada: (and in fact it won't work where it's not needed, including for `halfMarginTop`.) See `git log --stat -p --grep styles`, and in particular 
9565accff and 2127349a0 .

---

OK, I haven't read through the commits after that, but this should be enough for this round of review. Thanks again!

gnprice:
Thanks @jainkuniya !

I'm going to focus in this review on just the first few commits, so we can get those merged and then focus on the rest. Specifically these four:
```
d9d86cf3e flow: Fix type of reaction in Messsage.
235ffaebd flow, groupItemsById: Make function param type more precise.
026e0ad9c messagesReducer: Make sure reaction in msg always have `user.user_id`.
# f3e1d6cd3 webView: Use `.closest` instead of `.matches` in `getTouchTargetType`.
4c3453179 aggregatedReactions: Include reacted users list.
```
(Ignoring the webview `.closest` one for now, just because it isn't related to the other four.)

`d9d86cf3e flow: Fix type of reaction in Messsage.`

* Thanks for the revisions on this!

* One thing I mentioned last time still applies:

  > Another thought I have looking at this commit is: surely either the version just before this change or just after it is a type error. If they both pass Flow, then our types somewhere aren't being effective, and we should figure out why. :slightly_smiling_face: If one of them doesn't, then either the previous commit or this one broke things.

  Can you identify where the type error is? That'd help us make sure we take care of whatever the problem was that caused our types not to be effective in catching the issue.


`235ffaebd flow, groupItemsById: Make function param type more precise.`

* Huh interesting. It kind of confuses me that this is necessary (Flow is already inferring what type `T` is -- why doesn't it infer this detail for itself?), but so be it.


```
commit 026e0ad9c

    messagesReducer: Make sure reaction in msg always have `user.user_id`.

    User object in reaction of message has `id` key in it, but reaction
    event has `user_id` in it. So copy `user_id` to `id`, and then we can
    always use `id`.
    
    In future we will be migrating to `user_id`, so on receiving messages
    from server migrate to new type that is `user_id`.
    
    Also added new migration (`5`). So that old stored messages also get
    migrated.
```

* Huh, I'm a bit confused -- we're going to copy to `id`, but also migrate to the type that has `user_id`?

  Those sound like going in two opposite directions. The latter is the one I'm hoping to see, though.

* Reading through the code: this has the effect of making a lot of our application code need to worry about legacy vs non-legacy messages, and call `migrateMessages` in various spots.

  The way I'd like to arrange this would be: *none* of the code outside of `api/`, and I guess a store migration, knows that "legacy" messages ever existed. In other words: everything on the inside, everything that isn't directly interacting with the messy outside world of version skew, deals with data in nice consistent types.

  It's OK that the commit that introduces the legacy types causes the word "legacy" to appear in some of the app code (outside of `api/`), because that's just putting a more accurate name on what was already the truth. But what I'd like to avoid, even in the middle of the branch, is having actual conversion logic show up in the app code.

* Specifically, this line worries me:
```
       this.setState({
-        messages,
+        messages: migrateMessages(messages),
```

  Can you arrange things so that `migrateMessages` only appears in `src/api/`, and the store migration?

* Also: looking at `migrateMessages`, it looks like it won't work right when the server *does* start sending the preferred format. Let's fix that.

* Also, `migrateMessages` is the kind of function that would benefit greatly from having its own unit tests. :wink: 


```
4c3453179 aggregatedReactions: Include reacted users list.
```

* This previous comment applies to this code:

```
+        const reactedUser: User = (users && users.get(x.user.user_id)) || {
+          ...NULL_USER,
+          ...x.user,
+        };
```

  > Yikes -- what's all this doing?
  >
  > The baseline here is to simply use `x.user`. Anything more complex than that, there needs to be a clear reason for.
  > [...]
  > [The work] this code seems to be doing is to pull in a bunch of garbage data, i.e. NULL_USER. Pretty much by definition, if we ever actually touch any of that data, we have a bug -- so if there's some other code that complains or fails when we leave it out, we should fix that code.

jainkuniya:
> Can you identify where the type error is? That'd help us make sure we take care of whatever the problem was that caused our types not to be effective in catching the issue.

We have pre assumed that it will always be `user_id`, in both the cases. And we are using same [Reaction](https://github.com/zulip/zulip-mobile/blob/master/src/api/modelTypes.js#L235) in both the places i.e in [Message](https://github.com/zulip/zulip-mobile/blob/master/src/api/modelTypes.js#L372) and in [reaction action](https://github.com/zulip/zulip-mobile/blob/master/src/actionTypes.js#L339). So there is no flow error.

jainkuniya:
> Huh interesting. It kind of confuses me that this is necessary (Flow is already inferring what type T is -- why doesn't it infer this detail for itself?), but so be it.

Without this commit, flow gives error:
```
Running flow...
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/boot/store.js:99:32

Cannot call groupItemsById with migrateMessages(...) bound to items because property id is read-only in
object type [1] but writable in object type [2] in array element.

     src/boot/store.js
      96│     const { messages } = state; // migrate message from LegacyMessage to Message (LegacyReaction -> Reaction)
      97│     return {
      98│       ...state,
      99│       messages: groupItemsById(migrateMessages(Object.keys(messages).map(id => messages[id]))),
     100│     };
     101│   },
     102│ };

     src/message/messagesReducers.js
 [1]  20│ export const migrateMessages = (messages: LegacyMessage[]): Message[] =>

     src/utils/misc.js
 [2]  24│ export function groupItemsById<T: { id: number }>(items: T[]): { [id: number]: T } {


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/message/messagesReducers.js:136:27

Cannot call groupItemsById with migrateMessages(...) bound to items because property id is read-only in
object type [1] but writable in object type [2] in array element.

     src/message/messagesReducers.js
 [1]  20│ export const migrateMessages = (messages: LegacyMessage[]): Message[] =>
        :
     133│     case MESSAGE_FETCH_COMPLETE:
     134│       return {
     135│         ...state,
     136│         ...groupItemsById(migrateMessages(action.messages)),
     137│       };
     138│
     139│     case EVENT_REACTION_ADD:

     src/utils/misc.js
 [2]  24│ export function groupItemsById<T: { id: number }>(items: T[]): { [id: number]: T } {
```

jainkuniya:
> Huh, I'm a bit confused -- we're going to copy to id, but also migrate to the type that has user_id?

Sorry my mistake. It was supposed to be:
```
So copy `id` to `User_id`, and then we can
    always use `user_id`.
```

* Now `migrateMessages` is only limited to `src/api`.

> Also: looking at `migrateMessages`, it looks like it won't work right when the server does start sending the preferred format. Let's fix that.

* Now I have added a condition to handle it :)

* Added tests for `migrateMessages`

jainkuniya:
Removed the usage of `NULL_USER`

gnprice:
Thanks @jainkuniya !

Again focusing on those 4 early commits.

#### f49bf1fee flow: Fix type of reaction in Messsage.

* You mention a migration, and link to [a chat thread](https://chat.zulip.org/#narrow/stream/3-backend/topic/key.20mismatch.20in.20msg.20reaction). But that conversation doesn't seem to agree with the new `Reaction` type in this diff. Here's what Steve said there:

  > In this specific case I would just want to send user_id and no email or full_name, ideally.

  Did any version of this migration end up happening on the server?

* I asked:

  > Another thought I have looking at this commit is: surely either the version just before this change or just after it is a type error. If they both pass Flow, then our types somewhere aren't being effective, and we should figure out why. :slightly_smiling_face: If one of them doesn't, then either the previous commit or this one broke things.
  >
  > Can you identify where the type error is?

  You offered an answer (thanks!):

  > We have pre assumed that it will always be `user_id`, in both the cases. And we are using same [Reaction](https://github.com/zulip/zulip-mobile/blob/master/src/api/modelTypes.js#L235) in both the places i.e in [Message](https://github.com/zulip/zulip-mobile/blob/master/src/api/modelTypes.js#L372) and in [reaction action](https://github.com/zulip/zulip-mobile/blob/master/src/actionTypes.js#L339). So there is no flow error.

  But it didn't really address what I was wondering about, so I guess the question wasn't clear.

  So, I checked for myself. The answer seems to be:
  * First, Flow does indeed pass both just before this commit (good, because that's master!), and just after.
  * OK, so apparently our types before this commit are just wrong; and Flow says our code agrees with the types. So, why hasn't that caused problems already? E.g., crashing because we look for a property that isn't there.
  * It turns out we *weren't using* this property at all, anywhere. So we never bumped into the fact that we'd written down the wrong (in some contexts) name for it.
  * Two ways to see that efficiently, using Flow:
    * Comment out the property, so the type says it isn't there at all. Run Flow -- no error. If we'd had any code looking for this `user_id`, it would have shown up here as an error.
    * Make the entire `user` property have the type `mixed`, which means "could be anything, can't make any assumptions".
      * Run Flow, and there are 3 errors. These are all the places we try to look inside the `user` property on a reaction.
      * Read the code shown for each error. In each case, the only part we look at is the `user.email`.

#### 759d8f102 flow, groupItemsById: Make function param type more precise.

* Ah, I see now. I think the point is that this *isn't* really more precise -- it's actually more general. In particular, we're allowing it to have a property `+id: number`, not only a property `id: number`.

  * First off: this notation `$ReadOnly<{ id: number }>` is [exactly equivalent to](https://flow.org/en/docs/types/utilities/#toc-readonly) `{ +id: number }`. Let's prefer the latter here -- I think it deflates some of the mystery of what `$ReadOnly` means.

  * Then: I say `{ id: number }` **is a subtype** of `{ +id: number }`. Why? If you have a `{ id: number }`, you can pass it to something that takes a `{ +id: number }` -- after all, the thing is just giving you an extra promise that it won't write to the object's `id`. But if you have a `{ +id: number }`, then you *made* a promise not to write to the object's `id` -- so you can't pass it to something that takes `{ id: number }`, because that other code might write to it and break the promise you made.

  * So, we're changing the bound on this parameter from `{ id: number }` to a supertype, `{ +id: number }`.

  * That makes it no longer mysterious why we need the change -- it's because we're going to use the function in a new place that needs the more general interface.


#### b234cfacd messagesReducer: Make sure reaction in msg always have `user.user_id`.

* Thanks for the updates!

* Can we avoid splitting up the definition of `Message`? (Splitting it makes it more complicated to read.) And instead write

```
export type LegacyMessage = $ReadOnly<{|
  ...$Exact<Message>,
  reactions: $ReadOnlyArray<LegacyReaction>,
|}>;
```

just overriding the property from plain `Message`.



#### 192463070 aggregatedReactions: Include reacted users list.

* Much improved, thanks! :smile: 

* This code can be simplified:
```
+        const reactedUser: User | ReactionUser = (users && users.get(x.user.user_id)) || {
+          ...x.user,
+        };
```

to just:
```
        const reactedUser: ReactionUser = x.user;
```

(And then, just inline that where `reactedUser` is used.)

That'd let some types be simplified downstream, too.

jainkuniya:
> Did any version of this migration end up happening on the server?

This is how we get in message fetch call:
```

reactions: Array(33)
0:
  emoji_code: "1f476"
  emoji_name: "baby"
  reaction_type: "unicode_emoji"
  user:
    email: "***@gmail.com"
    full_name: "****"
    id: 1057
    __proto__: Object
__proto__: Object
```

And this is how we get in reaction add event:
```
user:
  email: "****@gmail.com"
  full_name: "****"
  user_id: 161
  __proto__: Object
```

So, seems none of the migration happened.

> But that conversation doesn't seem to agree with the new Reaction type in this diff.

In the conversation, focus was on discussing between `user_id` and `id`. So I migrated considering that only. BTW if we just have `user_id`, then we should have complete list of users all the time. I mean its size is too large in large realms and and we cannot show reactionList screen until we have full list.

So should we migrate to only `user_id`? I don't think we should, your thoughts?

> it's because we're going to use the function in a new place that needs the more general interface.

Yup, making it generic so that it can be used in other places a well.

> just overriding the property from plain Message.

Done 👍 

> (And then, just inline that where reactedUser is used.)

Done 👍

gnprice:
#### b0f1f69c1 flow: Fix type of reaction in Messsage.

* Hmm. Your comments are predicting that a certain migration will happen. But nobody in that thread suggested that we might do that particular migration :wink:, and there's no sign of it happening. So I don't think that prediction is a good one -- certainly not as it stands.

  Steve's first suggestion there was that we just handle it as it currently is:

  > Unfortunately, I think it's a bit difficult to fix this in a backward compatible way, so I recommend for the proximal issue that you just deal with it (and perhaps try to encapsulate it with a helper function of some sort, or just immediately mutate it at the "edge").

  Tim then suggested that we (1) choose something to migrate to, then (2) do that:

  > decide which we prefer, and then start sending both; this would set us up to migrate mobile + deprecate sending both in several months.

  To which Steve replied that his preferred answer to what to migrate to was plain `user_id`:

  > In this specific case I would just want to send user_id and no email or full_name, ideally.

  So if I'd predict any migration based on this thread, it would be to that.

* I think it's best if we try to get this PR moving without worrying about a migration -- I don't think a migration of this is a priority for either us or the backend folks. So let's take Steve's original suggestion: let's just convert the `.user.id` form to the `.user.user_id` form at the very edge. Which means as we get each message from the server.


#### 163963049 flow, groupItemsById: Make function param type more generic.

* Cool, I'll go and merge this. I'll want to add a bit to the commit message, because it's pretty mysterious as it stands.


#### cb8763aa9 messagesReducer: Make sure reaction in msg always have `user.user_id`.

* This appeared in your newest revision:

```
+++ src/reduxTypes.js
+export type LegacyMessagesState = {|
+  [id: number]: $Exact<LegacyMessage>,
+|};
```

That would mean "legacy" types spreading within our internal application code. We don't want to do that.

Fortunately it doesn't look like you use this, so can just take it back out.


#### e08a867a0 aggregatedReactions: Include reacted users list.

* Hmm, I think you misunderstood me :wink:

  The point was that we don't need all that multi-conditional goo with the `users &&`! We can just write `users: [...prevReaction.users, x.user]`.

gnprice:
(And merged that one commit, as 0b8920aea .)

jainkuniya:
>  just convert the .user.id form to the .user.user_id form at the very edge. 

That is what we doing right now, right? I have written migration, so that older message will also have `user_id` instead of `id`. I guess I am not getting what you mean to say here. :)


> Fortunately it doesn't look like you use this, so can just take it back out.

Done 👍 

> Hmm, I think you misunderstood me 😉

Yup! :)

Do you mean to always use `x.user`, means always take user from reaction object? 
```
export type AggregatedReaction = {|
  code: string,
  count: number,
  name: string,
  selfReacted: boolean,
  users: $ReadOnlyArray<User | ReactionUser>,
  type: string,
|};
```
I have made it `User | ReactionUser`. So if user is deactivated, we can get his name from reaction object user and show it. By looking up first in `users` will make sure that if someone has changed his name, we show his updated name.

gnprice:
> So if user is deactivated, we can get his name from reaction object user and show it. By looking up first in `users` will make sure that if someone has changed his name, we show his updated name.

Pretty sure the info in the reaction object reflects the server just looking up that user's details at the time it's giving us the data -- so if they've changed their name etc., it's already reflected there.

Or do you mean if they've changed it since we got the reaction from the server? That's a good point... and really it highlights the problem with keeping all this data denormalized, scattered in copies across lots of miscellaneous objects. :stuck_out_tongue: We shouldn't keep around data that we're not going to keep up to date.

(The principle is very much like with the bug where we used to keep a `flags` property on the messages under `state.messages` but let it get stale didn't update that, and so `flags.starred` would be stale -- the cause of #2676 and #2677, and explained in https://github.com/zulip/zulip-mobile/pull/2843#issuecomment-409509922 .)

Hmm. I think in fact what that really means is that (a) we should always use the user ID here, and look it up in the `users` data -- never get the other details from the reaction object; (b) as a further cleanup, when we get a reaction object from the server we should completely throw away the `user` object inside it, and just keep a `user_id` property!

(Earlier you alluded to a possible future where we don't keep a comprehensive user list all the time. In that future, we might add some logic like: (b') when we get a reaction from the server containing a `user` object, we go and store that in our user list... as part of maintaining an invariant that we have a comprehensive list of users *that we have other data referring to*, like they sent a message. But we don't need to worry about anything like that now, because at present we do just keep a comprehensive list of users in the realm.)

jainkuniya:
(continuing after the call here, conclusion of the call was to find what webapp does for deactivated users and do same in the mobile app )
Just tested on webapp, others users can see deactivated users messages, reactions as it  is without any difference. 
Deactivated status is visible in user profile and at the time of sending PM.
image
image

gnprice:
Thanks! That's a helpful finding on deactivated users.

I just filed #3551 with the things you found we don't do that we probably should. Not a priority, though, and we have plenty of other things to focus on -- so let's focus for now on the things we're already doing.

I also just added some jsdoc, in eeb0c3859 (and its parent). I hope that helps clarify when to include deactivated users and when not to -- please take a look and ask any follow-up questions.

Coming back to emoji reactions: I think the answer this gives us is that we simply don't have to worry or care about whether a user is deactivated or not. Just be sure to use `getAllUsers` and friends, not `getUsers*` or `getActiveUsers*`. 

And then that gets us back to this:

> I think in fact what that really means is that (a) we should always use the user ID here, and look it up in the `users` data -- never get the other details from the reaction object; (b) as a further cleanup, when we get a reaction object from the server we should completely throw away the `user` object inside it, and just keep a `user_id` property!

So I think that's the next step for this PR.

src/reactions/ReactionListTabView.js
```diff
@@ -0,0 +1,49 @@
...
+import styles from '../styles';
+
+export default (
+  aggregatedReactions: $ReadOnlyArray<AggregatedReaction>,
+  initialRouteName: string,
+) => {
+  const reactionsTabs = aggregatedReactions.reduce((tabs, aggregatedReaction) => {
+    const { width } = Dimensions.get('window');
+    const emojiSize: number = Math.min((width / aggregatedReactions.length) * 0.32, 24);
+    const countSize: number = Math.min((width / aggregatedReactions.length) * 0.4, 16);
```
gnprice: This looks like it will make things giant when there's just one or two reactions. Will it, or is there something I'm missing?

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,58 @@
...
+type Props = {
+  message: Message,
+  navigation: NavigationScreenProp<*> & {
+    state: {
+      params: {
+        src: string,
+        message: Message,
+      },
+    },
+  },
```
gnprice: See a36814e80 for a better way to write this type.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,58 @@
...
+  ownUserId: number,
+  users: Map<number, User>,
+};
+
+class MessageReactionList extends PureComponent<Props> {
+  render() {
+    const { message, navigation, ownUserId, users } = this.props;
+    const { initialRouteName } = navigation.state.params;
```
gnprice: Wait, is it `src` or `initialRouteName`?

src/webview/js/generatedEs3.js
```diff
@@ -551,9 +551,19 @@ documentBody.addEventListener('click', function (e) {
 
 var handleLongPress = function handleLongPress(target) {
   hasLongPressed = true;
+
+  if (target.closest('.reaction-list')) {
+    sendMessage({
+      type: 'reactionDetails',
+      messageId: getMessageIdFromNode(target),
+      targetReactionName: target.getAttribute('data-name') || ''
```
gnprice: Hmm -- what if `target` isn't the element with `data-name` but rather some descendant? For example, the `img` for an image emoji.

I think you want to get this from `target.closest('.reaction')`.

static/translations/messages_en.json
```diff
@@ -161,5 +161,6 @@
   "Add subscribers": "Add subscribers",
   "Set yourself to away": "Set yourself to away",
   "Clear": "Clear",
+  "Reactions": "Reactions",
```
gnprice: This should happen in a commit that actually uses this string.

src/reactions/ReactionUserList.js
```diff
@@ -0,0 +1,49 @@
...
+import { FlatList } from 'react-native';
+import { connect } from '../react-redux';
+
+import type { Dispatch, UserOrBot } from '../types';
+import UserItem from '../users/UserItem';
+import { navigateToAccountDetails } from '../actions';
+
+type Props = {
+  dispatch: Dispatch,
+  reactedUsers?: $ReadOnlyArray<?UserOrBot>,
```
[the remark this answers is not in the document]
jainkuniya: I guess decision to previous one, will take care of this as well.

src/reactions/aggregateReactions.js
```diff
@@ -1,36 +1,52 @@
...
-      const prevReaction = reactionMap.get(x.emoji_name);
-      if (prevReaction) {
-        reactionMap.set(x.emoji_name, {
-          ...prevReaction,
-          count: prevReaction.count + 1,
-        });
-      }
-    }
+  ownUserId: number,
+  users?: Map<number, User>,
```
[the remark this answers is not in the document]
gnprice: Hmm, let's see.

It looks like what this `users` parameter is being used for is just for `users.get(x.user.user_id)`, and as a sort of flag by testing `users` for truthiness.

How about simply skipping that and letting the `AggregatedReaction` have an array of user IDs rather than user objects? Then the parameter can go away entirely, and the consumer of the data can look up for itself the user IDs as needed.

src/reactions/MessageReactionList.js
jainkuniya: Type was wrong, it is `initialRouteName` 😞

src/styles/tabs.js
```diff
@@ -1,12 +1,17 @@
 /* @flow strict-local */
 import { BRAND_COLOR } from './constants';
 
+/**
+ * react-navigation tabNavigation also treat styles as {}
+ * so match the same type here too
+ */
```
gnprice: Nit: Let's reserve the `/**` jsdoc marker for comments that are at the level of clarity and information that makes them a proper doc comment. This comment isn't and that's 100% fine -- this type doesn't need a doc comment -- but we'll just write it with plain `/*` or `//` to keep that clear.

See for comparison [this style-guide item](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use--for-public-quality-private-documentation) from Flutter. The context is a little different, but this point applies just as well for us.

gnprice: Separately: let's put this comment immediately above the `style` property itself.

src/emoji/EmojiRow.js
```diff
@@ -29,25 +29,39 @@ type Props = {|
...
-          {imageEmoji ? <ImageEmoji emoji={imageEmoji} /> : <Emoji name={name} size={20} />}
-          <RawLabel style={styles.text} text={name} />
+          {imageEmoji ? (
+            <ImageEmoji emoji={imageEmoji} size={size} />
+          ) : (
+            <Emoji name={name} size={size} />
+          )}
+          {showEmojiName && <RawLabel style={styles.text} text={name} />}
         </View>
       </Touchable>
```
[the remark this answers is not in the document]
gnprice: bump

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,105 @@
...
+      showLabel: true,
+      showIcon: false,
+      style: {
+        borderWidth: 0.15,
+        borderColor: 'gray',
+      },
+    }),
+  });
+};
+type NavigationsProps = {|
```
gnprice: nit: blank line between top-level declarations

(ditto below)

src/nav/navActions.js
```diff
@@ -97,6 +97,15 @@ export const navigateToEditStream = (streamId: number): NavigateAction =>
 export const navigateToStreamSubscribers = (streamId: number): NavigateAction =>
   NavigationActions.navigate({ routeName: 'invite-users', params: { streamId } });
 
+export const navigateToMessageReactionScreen = (
```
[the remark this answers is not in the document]
gnprice: bump

src/reactions/ReactionListTabView.js
gnprice: There was indeed something I was missing -- I was reading `Math.min` the wrong way around. :slightly_smiling_face: That's taken care of, then.

On the other hand we should also have the thing that I *thought* these were doing -- we should have minimum values so that when there are a *lot* of reactions, they don't become too tiny to read. Better to show an incomplete list that can be read, than a complete list where nothing can be read.

Stepping back a level (and as we discussed by voice today): instead of doing this computation based on `Dimensions.get('window')`, it'd be much cleaner and more robust to do it with flexbox style props. I know CSS flexbox has ample features to do what we want here, and hopefully RN's flexbox does too. As is, this version makes an assumption that the component has the whole width of the window, and that makes it more fragile than it would be with flexbox layout.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,105 @@
...
+  navigation: NavigationScreenProp<{ params: {| initialRouteName: string, messageId: number |} }>,
```
gnprice: How about calling this `targetReactionName` throughout (like you do in the webview-to-RN event), instead of `initialRouteName`?

The latter is kind of confusing from the perspective of outside this component -- it only makes sense as sort of an implementation detail, of how the component has chosen to name the "routes" in the TabNavigator it creates within itself. But `targetReactionName` describes the meaning pretty well.

src/webview/js/js.js
```diff
@@ -740,9 +740,22 @@ const handleLongPress = (target: Element) => {
...
 
+  if (target.closest('.reaction-list')) {
+    const reactionNode = target.closest('.reaction');
+    if (!reactionNode) {
+      return;
+    }
+    sendMessage({
+      type: 'reactionDetails',
+      messageId: getMessageIdFromNode(target),
+      targetReactionName: reactionNode.getAttribute('data-name') || '',
```
gnprice: As we discussed by voice today: let's use `requireAttribute`, like the other similar code around this does.

src/webview/js/js.js
```diff
@@ -740,9 +740,22 @@ const handleLongPress = (target: Element) => {
 
   hasLongPressed = true;
 
+  if (target.closest('.reaction-list')) {
+    const reactionNode = target.closest('.reaction');
+    if (!reactionNode) {
+      return;
+    }
```
gnprice: Also from our voice chat today: why have this `return` case at all? How about just starting with `target.closest('.reaction')` and not looking for the `.reaction-list`? We don't seem to actually use the latter once we find it.

src/webview/js/js.js
```diff
@@ -740,9 +740,22 @@ const handleLongPress = (target: Element) => {
...
+    });
+    return;
+  }
+
   sendMessage({
     type: 'longPress',
-    target: target.matches('.header') ? 'header' : target.matches('a') ? 'link' : 'message',
+    target: target.closest('.header') ? 'header' : target.closest('a') ? 'link' : 'message',
     messageId: getMessageIdFromNode(target),
     href: target.matches('a') ? requireAttribute(target, 'href') : null,
```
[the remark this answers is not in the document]
jainkuniya: Okay, let's take it separately.

src/emoji/EmojiRow.js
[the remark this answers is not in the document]
jainkuniya: Done, created new component `Emoji`

src/reactions/ReactionListTabView.js
jainkuniya: For now, let's drop this commit. 

Actually we don't even want `flex` here. React navigation has its own measurement. It make sure that tab bar occupy whole space available to it. And tabs content is in centre. 

Add minWidth, maxWidth doesn't work as expected. 

Anyway it is rare case when messages have so many reactions. Let's work on this on another PR.

src/reactions/aggregateReactions.js
```diff
@@ -15,14 +15,18 @@ export default (
         code: x.emoji_code,
         count: 0,
         selfReacted: false,
+        users: [],
       };
       reactionMap.set(x.emoji_name, item);
     }
 
     item.count += 1;
+    item.users.push(x.user_id);
```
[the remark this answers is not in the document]
jainkuniya: Cool, me too 😍

src/emoji/EmojiRow.js
```diff
@@ -39,21 +37,17 @@ class EmojiRow extends PureComponent<Props> {
...
           <RawLabel style={styles.text} text={name} />
         </View>
       </Touchable>
     );
   }
 }
 
-export default connect((state, props): SelectorProps => ({
-  imageEmoji: getActiveImageEmojiByName(state)[props.name],
-}))(EmojiRow);
```
gnprice: Please look again at this and clean up everything that's no longer being used here.

src/emoji/Emoji.js
```diff
@@ -1,9 +1,37 @@
...
+type Props = {|
+  name: string,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|};
+
+type ComponentProps = {|
+  name: string,
+|};
```
gnprice: Let's avoid this redundancy -- there's no need for it. Search for `SelectorProps` to see examples.

src/reactions/ReactionUserList.js
```diff
@@ -0,0 +1,50 @@
...
+
+  render() {
+    const { reactedUserIds, users } = this.props;
+
+    return (
+      <FlatList
+        data={reactedUserIds}
+        keyExtractor={userId => `${userId}`}
+        renderItem={({ item }) => {
+          const user: ?UserOrBot = users.get(item);
```
gnprice: No need to annotate locals when type inference can handle them.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,101 @@
...
+      <Screen title="Reactions" scrollEnabled={false}>
+        <View style={styles.flexed}>
+          <TabView />
+        </View>
+      </Screen>
+    );
+  }
+}
+
+export default connect((state: GlobalState, props: NavigationsProps): SelectorProps => ({
```
gnprice: Here also, see examples by searching for `SelectorProps`.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,101 @@
...
+  ...SelectorProps,
+|};
+
+class MessageReactionList extends PureComponent<Props> {
+  render() {
+    const { message, navigation, ownUserId, usersById } = this.props;
+    const { targetReactionName } = navigation.state.params;
+    const { reactions } = message;
+
+    const aggregatedReactions: $ReadOnlyArray<AggregatedReaction> = aggregateReactions(
```
gnprice: Ditto re: annotating locals.

src/emoji/Emoji.js
```diff
@@ -1,9 +1,37 @@
...
+    const { name, imageEmoji } = this.props;
+
+    // TODO: this only handles Unicode emoji (shipped with the app)
+    // and realm emoji, but not Zulip extra emoji.  See our issue #2846.
+    return imageEmoji ? <ImageEmoji emoji={imageEmoji} /> : <UnicodeEmoji name={name} size={20} />;
+  }
+}
+
+export default connect((state, props: ComponentProps): SelectorProps => ({
+  imageEmoji: getActiveImageEmojiByName(state)[props.name],
```
gnprice: Is this still the right set of emoji? For an old reaction, there might be an inactive emoji, I think.

(Any change should be a separate commit from the refactor that moves this code, though.)

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,101 @@
...
+      aggregatedReaction.name,
+      {
+        screen: () => (
+          <ReactionUserList reactedUserIds={aggregatedReaction.users} users={usersById} />
+        ),
+        navigationOptions: {
+          tabBarLabel: () => (
+            <View style={styles.row}>
+              <Emoji name={aggregatedReaction.name} />
+              <RawLabel text={`  ${aggregatedReaction.count}`} />
```
gnprice: These space characters look awfully suspicious as a way of trying to control this layout. :wink: Can we use, say, a style with padding instead?

src/emoji/Emoji.js
jainkuniya: yup, let's take it in another commit.

Let's do it in follow up PR.
Because `<Emoji` is used in `<EmojiRow`. And `<EmojiRow` is used in `EmojiPIckerScreen` and `EmojiAutocomplete`, where we only wants active emojis.

jainkuniya: Added c1058092

src/emoji/Emoji.js
```diff
@@ -1,9 +1,33 @@
...
+|};
+
+type Props = {|
+  name: string,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|};
+
+class Emoji extends PureComponent<Props, void> {
```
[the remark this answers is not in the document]
gnprice: ... Right, that's the `State` type-parameter.

There's no reason to specify a state type explicitly here. Taking that out before merging.

src/reactions/ReactionUserList.js
```diff
@@ -0,0 +1,50 @@
...
+  dispatch: Dispatch,
+  reactedUserIds: $ReadOnlyArray<number>,
+  users: Map<number, UserOrBot>,
+};
+
+class ReactionUserList extends PureComponent<Props> {
+  handlePress = (email: string) => {
+    const { dispatch } = this.props;
+    dispatch(navigateToAccountDetails(email));
+  };
```
gnprice: I think this is best squashed into the same commit that made this component.

src/reactions/aggregateReactions.js
```diff
@@ -15,14 +15,18 @@ export default (
...
     if (x.user_id === ownUserId) {
       item.selfReacted = true;
     }
   });
-  return Array.from(reactionMap.values());
+  return Array.from(reactionMap.values()).sort(
+    (r1: AggregatedReaction, r2: AggregatedReaction) => r2.count - r1.count,
+  );
```
gnprice: This affects the reactions shown in the message list, too, right?

Does this match how the webapp orders these?

Whatever change we make here, probably best put at the start of the branch (in particular before creating the new screen) -- that way the screen looks the way we want it to from the beginning.

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,98 @@
...
+      showIcon: false,
+      style: {
+        borderWidth: 0.15,
+      },
+    }),
+  });
+};
+
+type NavigationsProps = {|
+  navigation: NavigationScreenProp<{ params: {| targetReactionName: string, messageId: number |} }>,
```
gnprice: How about just `reactionName`? I don't think the word "target" adds much information here, or even makes a lot of sense outside the few lines in js.js that fire off the event to the outside.

src/nav/AppNavigator.js
```diff
@@ -45,6 +46,7 @@ export default createStackNavigator(
     'emoji-picker': { screen: EmojiPickerScreen },
     loading: { screen: LoadingScreen },
     main: { screen: MainScreenWithTabs },
+    'message-reaction': { screen: MessageReactionList },
```
gnprice: How about `message-reactions`? This is after all for *all* the reactions on a message, which is very often several of them.

(OTOH `MessageReactionList` is just fine -- it is indeed a "reaction list", i.e. a list of reactions.)

src/message/messageActionSheet.js
```diff
@@ -110,6 +111,11 @@ const addReaction = ({ message, dispatch }) => {
 };
 addReaction.title = 'Add a reaction';
 
+const showReactions = ({ message, dispatch }) => {
+  dispatch(navigateToMessageReactionScreen(message.id));
+};
+showReactions.title = 'Reactions';
```
gnprice: Let's make this "See who reacted".

gnprice:
Got an error playing with this in dev:
image

That was while looking at the reaction screen, and adding and removing reactions from another user (from the webapp)

gnprice:
> That was while looking at the reaction screen, and adding and removing reactions from another user (from the webapp)

Mmm, I think specifically I'd removed the reaction that was the one I'd navigated to. I.e. the other user had reacted with `slight_smile`; I'd long-pressed that reaction to get to this screen; then the other user removed that reaction.

gnprice:
Hmm also this is puzzling:
image

They aren't sorted there by count. That's while they are indeed ordered that way in the message list:
image

jainkuniya:
> Mmm, I think specifically I'd removed the reaction that was the one I'd navigated to. I.e. the other user had reacted with slight_smile; I'd long-pressed that reaction to get to this screen; then the other user removed that reaction.

Interesting situation 😅

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    const aggregatedReactions = aggregateReactions(reactions, ownUserId);
+
+    // there can be cases where reaction have been removed by the time user see this screen
+    // or on re-render
+    // so confirm reaction `reactionName` still exists
+    const TabView = getReactionsTabs(
+      aggregatedReactions,
+      aggregatedReactions.find(reaction => reaction.name === reactionName) ? reactionName : '',
```
gnprice: This would be a bit cleaner after `reactionsTabs` has been constructed, inside `getReactionsTabs`.

jainkuniya: Updated 👍

src/reactions/MessageReactionList.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+    ]),
+  );
+
+  // there can be cases where reaction have been removed by the time user see this screen
+  // or on re-render
+  // so confirm reaction `reactionName` still exists
+  const initialRouteName = aggregatedReactions.find(reaction => reaction.name === reactionName)
+    ? reactionName
+    : '';
```
gnprice: So, the reason I said this would be cleaner after `reactionsTabs` had been constructed is because the code can actually be different and simpler. :slightly_smiling_face: It can use `reactionsTabs`. That's also more directly connected to the point of why this logic is here.

jainkuniya:
I tried with multiple users, reacting with reactions even when user is on reactions list. It is getting re rerendered correctly.

src/reactions/MessageReactionList.js
jainkuniya: oh ya!

Updated 👍

gnprice:
> I tried with multiple users, reacting with reactions even when user is on reactions list. It is getting rerendered correctly.

OK, thanks. It seems like whatever I saw is uncommon, and it isn't critical anyway; so I'll merge without worrying about it. If we see it later, we can handle it as a bug.

I'm about to merge most of this branch. We're not yet *quite* done, because as I prepared to merge I noticed a new issue. I looked into the relevant `react-navigation` API, and found:
* The use of `TabNavigator` is deprecated. See [the 2.x release notes](https://reactnavigation.org/blog/2018/05/07/react-navigation-2.0.html), and more generally the discussion in our commits 98836b84d, 845834c1d, 0d2066f41.
* Using `TabNavigator` gives different behavior on iOS vs. Android! I think both you and I have been mostly looking at it on Android. The iOS behavior seems like not what we want.
  * So, let's use `createMaterialTopTabNavigator`.
* In the API doc: https://reactnavigation.org/docs/en/2.x/material-top-tab-navigator.html#initialroutename for the `initialRouteName` option, there's no mention of `''` as a special value. Rather, it's supposed to be the name of some route. (This is the question that originally caused me to look the API up.)
  * When we don't have a useful `reactionName`, let's just not pass this option.

gnprice:
I've just pushed :tada: the following commits:
e438f82b2 reactions: Sort aggregated reactions by most numerous.
943c31ba1 styles/tabs [nfc]: Add catch-all option `style`.
595f1dcb3 reactions: Add new component ReactionUserList.
a49b1f1f8 reactions: Add new screen `MessageReactionList`.
0feabcc03 reactions [nfc]: Assimilate style of props type definitions.
327ace399 reactions [nfc]: Rename some props to explicit `allUsersById`.

The first four are from this branch; the last two are commits I added on top.

I've also pushed back to this branch some revisions, so it's now the following commits:
596965dd3 webview: On long-press on reactions, navigate to reactions list.
bb8c05c52 MessageReactionList: Focus tab for reaction that user long-pressed on.
c7ea86008 message action sheet: Add option to show reaction list.
9f265ecfd reactions [nfc]: Add some jsdoc and revise a comment.

That includes one commit of mine at the end, some adjustments to commit messages, and one substantive change described in a commit message:
```
commit bb8c05c52e292839dbf8b3d734ee3c40fdcbf7a5
Author: Vishwesh Jainkuniya <gitvishwesh@gmail.com>
Date:   Thu Jan 10 16:13:31 2019 +0530

    MessageReactionList: Focus tab for reaction that user long-pressed on.
    
    [greg: fixed `reactionName` type in nav prop and inward; type error
     was hidden by lack of type-checking on `StackActions.push`]
```

The net effect is:
* The new screen is merged!
* But it isn't actually reachable in the UI; the commits that do that are in the remaining branch.
* That's because of the issues I mentioned in the previous comment in the use of the `react-navigation` API.

Please fix that, and then I think we'll be able to merge the whole thing and close this out for real :smile:

gnprice:
OK -- this is now merged! :checkered_flag: :tada:

I made a couple more edits to the branch before merge:
* Moved the `createMaterialTopTabNavigator` commit to the start. That fixes an issue I wanted to fix before adding UI to get to this screen, so it goes before the commits that add that UI.
* Added a new commit refactoring the conditional buildup of the tabs options object -- demonstrates a bit of a trick for keeping those reasonably clean.

Glad to have this finally complete. :sweat_smile:

PR #3111, opened 2018-11-05
Always enable notifications + Fix them

borisyankov:
Previously we did disable notifications on emulators. That is not
only not needed but also counter-productive as we want to be able
to test them on different device configurations.
    
Removes the config flag `enableNotifications` completely and the
check that uses this flag to enable/not enable the notifications.

Also, fixes notifications not working in Android emulator!


gnprice:
Merged. Thanks @borisyankov for debugging this!

Issue #3107, opened 2018-11-03
yarn install error:detox: Command failed. Exit code: 65,and yarn ios error


xiaomingplus:
Hello,the master branch,run ```yarn install`` failed.

node:v8.12.0
yarn :1.10.1

```bash
 yarn install
yarn install v1.10.1
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning " > react-intl@2.4.0" has unmet peer dependency "prop-types@^15.5.4".
[4/4] 📃  Building fresh packages...
[-/3] ⠈ waiting...
[2/3] ⠈ @sentry/cli
error /Users/mintvivid/xiaomingplus/inbox/zulip-mobile/node_modules/detox: Command failed.
Exit code: 65
Command: scripts/postinstall.sh
Arguments:
Directory: /Users/mintvivid/xiaomingplus/inbox/zulip-mobile/node_modules/detox
Output:
/Users/mintvivid/Library/Detox/ios/f09d03bc547fdab9fe6559dff6ddb8d755bfe68d was found, but could not find Detox.framework inside it. This means that the Detox framework build process was interrupted.
         deleting /Users/mintvivid/Library/Detox/ios/f09d03bc547fdab9fe6559dff6ddb8d755bfe68d and trying to rebuild.
Extracting Detox sources...
Building Detox.framework from ./ios_src...
info Visit https://yarnpkg.com/en/docs/cli
```

then ,run ```yarn ios```,still failed:

```bash
** BUILD FAILED **


The following build commands failed:

	CompileC /Users/mintvivid/xiaomingplus/inbox/zulip-mobile/ios/build/Build/Intermediates.noindex/React.build/Debug-iphonesimulator/double-conversion.build/Objects-normal/x86_64/strtod.o /Users/mintvivid/xiaomingplus/inbox/zulip-mobile/node_modules/react-native/third-party/double-conversion-1.1.6/src/strtod.cc normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler
[...]

PR #3121, opened 2018-11-08
dependecies: Fix 'yarn' dependency installation

borisyankov:
Fixes: #3107

Currently, running `yarn` to install/update depencencies is failing
on iOS with XCode 10.
The cause is `Detox`, one of the dependencies fail to build.
Upgrading to the latest version fixes that problem.

Note on Detox:
We are not actively using it but we should, and that is why I did
update the dependency instead of removing it.

They are also finishing up their Android support:
https://github.com/wix/Detox/blob/master/docs/More.AndroidSupportStatus.md


borisyankov:
@gnprice this is short-n-simple + urgent ⏩

Issue #3075, opened 2018-10-25
Target Android SDK version 26


gnprice:
Beginning on 2018-11-01 -- that's next week -- any new release we upload to the Play Store must have `targetSdkVersion` set to 26 or higher.

This is a major change. It causes all systems running newer versions of Android than our previous target (22) to treat our app with a number of behavior changes introduced by the versions since then, up to the new target -- which means 6.0, 7.0, 7.1, and 8.0, aka Marshmallow, Nougat, and Oreo.  That's years of changes, with known and, surely, unknown changes that affect us. Here's a concise list of just the most important changes:
https://developer.android.com/distribute/best-practices/develop/target-sdk

I committed this change to master in 53970d46d, and released 19.0.100 to the alpha channel (on Android) with that change. That will help us find issues empirically. Now we need to find and fix them. :stuck_out_tongue: 

**UPDATE**: I've released 19.1.101 to Android alpha. It fixes the critical known issue that notifications completely didn't work (#3076). More testing needed for all the "known unknown" potential issues below, as well as to spot any others!

Here's a list, which I'll keep updated, of the things we need to do:
- [x] (@gnprice) Read through the [list of high-risk changes](https://developer.android.com/distribute/best-practices/develop/target-sdk) to find likely issues.
- [x] (@gnprice) Read through the more detailed "behavior changes" docs ([this one](https://developer.android.com/about/versions/marshmallow/android-6.0-changes) and later) for each release to find likely issues.
- [x] (everyone!) Play around with the ~19.0.100~ 19.1.101 alpha and look for issues.
- [x] (@gnprice, #3076) [Set channels on our notifications](https://developer.android.com/guide/topics/ui/notifiers/notifications#ManageChannels). In 19.0.100, notifications ~don't~ didn't work at all.
- [x] (filed as #3576) Test permissions behavior.
  - [x] (@gnprice) Read through the permissions in our output manifest, and compare to docs to find "dangerous" permissions and other potential pitfalls.
  - [x] (-> #3576) Test our use of the camera; test scenarios where the user first installs the app, we request permission, where the user then revokes permission, etc. See discussion of [dangerous permissions](https://developer.android.com/distribute/best-practices/develop/target-sdk#premarshmallow); and more detail [here](https://developer.android.com/guide/topics/permissions/overview#dangerous-permission-prompt) and usage notes [here](https://developer.android.com/training/permissions/usage-notes).
[...]

Issue #2812, opened 2018-07-14
Simplify how we translate strings outside components


gnprice:
For user-facing text whose source code is in the code of a component, `Label` and friends give us a nice convenient way to pass the text through translation.

When the code is in a function elsewhere, we have this way of doing it:
```
    const getString = value => this.context.intl.formatMessage({ id: value });
```
and then pass `getString` down the stack. We only do this in `MessageList` and its helpers, though -- everywhere else, we just don't translate. #1538 and #1539 cover two such areas.

This pattern is the right basic architecture, but it's wordy. The state we ought to get to is that basically every string in the source code that looks like it's meant for the user is clearly marked right there for translation. To do that, we'll be writing this a lot; we need to make it concise and low-noise.

This only needs two small changes:
* Change the name to `_`. This is a traditional name that's been widely used for this since at least the '90s, and the backend uses it.
* Put the handy wrapper (`getString` above, now to be called `_`) directly on `context`.

A good followup will be the equivalent of #1538 and #1539 with this style, and then to sweep to find untranslated strings that remain.

gnprice:
See also https://github.com/zulip/zulip-mobile/issues/1481#issuecomment-348678750 and subsequent comments for some previous discussion.

borisyankov:
I agree with the use of `context` for this vs what was used in the PRs. That is the reason I didn't merge earlier.

We need to figure out a more systemic approach to get everything translated. Would @kunall17 want to work on all the other related translation tasks?

gnprice:
(See #2824 for a PR to do this.)

PR #3123, opened 2018-11-09
android: Request WRITE_EXTERNAL_STORAGE permission.

### LINKED ISSUE #3115: android: Fix use of `WRITE_EXTERNAL_STORAGE` permission.
This is part of https://github.com/zulip/zulip-mobile/issues/3075.

This error from React-Native comes up on the debug build when I try to download an image on an emulator running API 28:

 

On an installation from Play Store, the app crashes.

See [this document](https://developer.android.com/training/permissions/requesting).

(@zulipbot add "area: android" "crash")
(@zulipbot claim)

jackrzhang:
The first time a user attempts to download an image, permission to write to external storage (the SD card) should be requested. If permission is granted by the user, the image should proceed to download.

If permission is denied, the image will not be fetched, and the corresponding text 'Storage permission denied' is shown. Any subsequent time that a user attempts to download an image, a 'rationale' modal will appear with explaining why the app needs permissions, and the permission will once again be requested.

<p align="center">
android-request-permissions-write-external-storage-1
</p>

If a user instructs that the permission not be asked for again, then subsequent attempts to download an image will only result in showing the text 'Storage permission denied', though the permission can always be changed via Android Settings.

<p align="center">
android-request-permissions-write-external-storage-2
</p>

For reference, see [React Native 0.57's PermissionsAndroid API](https://facebook.github.io/react-native/docs/permissionsandroid) and [Android's developer documentation on requesting permissions](https://developer.android.com/training/permissions/requesting).

This fixes #3115 and progresses #3075.
@gnprice

### DIFF
--- a/src/lightbox/LightboxActionSheet.js
+++ b/src/lightbox/LightboxActionSheet.js
@@ -1,6 +1,6 @@
 /* @flow */
 import type { Auth } from '../types';
-import downloadFile from '../api/downloadFile';
+import downloadImage from './downloadImage';
 import share from './share';
 import shareImage from './shareImage';
 import { showToast } from '../utils/info';
@@ -32,9 +32,9 @@ type ButtonType = {
   onPress: (props: ButtonProps) => void | boolean | Promise<any>,
 };
 
-const downloadImage = async ({ src, auth }: DownloadImageType) => {
+const tryToDownloadImage = async ({ src, auth }: DownloadImageType) => {
   try {
-    await downloadFile(src, auth);
+    await downloadImage(src, auth);
     showToast('Download complete');
   } catch (error) {
     showToast('Can not download');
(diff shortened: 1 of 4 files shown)


src/lightbox/LightboxActionSheet.js
```diff
@@ -50,7 +50,7 @@ const shareImageDirectly = ({ src, auth }: DownloadImageType) => {
 };
 
 const actionSheetButtons: ButtonType[] = [
-  { title: 'Download file', onPress: downloadImage },
+  { title: 'Download image', onPress: tryToDownloadImage },
```
borisyankov: Make sure to update the `messages_en.json` file.

jackrzhang: Will update!

I wonder if there's any way that this caught via the test suite.

jackrzhang:
(Also, this moves forward #2618.)

gnprice:
Thanks @jackrzhang ! Merged.

I made a few edits on top that I think make the code easier to understand -- please take a look.

I missed @borisyankov 's comment above (sorry!) until I'd already merged:
> Make sure to update the messages_en.json file.

 But actually I think the situation is that these messages -- all the labels in the lightbox action sheet -- are already untranslated even when our volunteer translators have kindly provided translations. :disappointed: Note the lack of any code to act on these strings to find their translations. I also quickly verified just now with German.

So that would be good to fix. Also while we're at it make the new messages in this new flow get translated too. :-)

gnprice:
(#2812 is some background for that translation issue.)

PR #3125, opened 2018-11-09
Display indicator for unread messages to the user

borisyankov:
Adds a visual indicator of which messages are read, which are unread, and animates that visual when a message becomes read.

### DIFF
--- a/src/session/sessionReducers.js
+++ b/src/session/sessionReducers.js
@@ -77,7 +77,6 @@ const initialState: SessionState = {
     top: 0,
   },
   debug: {
-    highlightUnreadMessages: false,
     doNotMarkMessagesAsRead: false,
   },
 };
--- a/src/settings/DebugScreen.js
+++ b/src/settings/DebugScreen.js
@@ -25,11 +25,6 @@ class DebugScreen extends PureComponent<Props> {
     return (
       <Screen title="Debug">
         <OptionRow
-          label="Distinguish unread messages"
-          defaultValue={debug.highlightUnreadMessages}
-          onValueChange={() => this.handleSettingToggle('highlightUnreadMessages')}
-        />
-        <OptionRow
           label="Do not mark messages read on scroll"
           defaultValue={debug.doNotMarkMessagesAsRead}
           onValueChange={() => this.handleSettingToggle('doNotMarkMessagesAsRead')}
--- a/src/types.js
+++ b/src/types.js
@@ -86,7 +86,6 @@ export type EditMessage = {|
 |};
 
 export type Debug = {|
-  highlightUnreadMessages: boolean,
   doNotMarkMessagesAsRead: boolean,
 |};
 
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -171,11 +171,10 @@ class MessageList extends Component<Props> {
       showMessagePlaceholders,
     } = this.props;
     const messagesHtml = renderMessagesAsHtml(backgroundData, narrow, renderedMessages);
-    const { auth, debug } = backgroundData;
+    const { auth } = backgroundData;
     const html = getHtml(messagesHtml, theme, {
       anchor,
       auth,
-      highlightUnreadMessages: debug.highlightUnreadMessages,
       showMessagePlaceholders,
     });
 
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -44,11 +44,11 @@ import {
 import { withGetText } from '../boot/TranslationProvider';
 
 import type { ShowActionSheetWithOptions } from '../message/messageActionSheet';
-import type { WebviewInputMessage } from './webViewHandleUpdates';
+import type { WebViewUpdateEvent } from './webViewHandleUpdates';
 import type { MessageListEvent } from './webViewEventHandlers';
 import getHtml from './html/html';
 import renderMessagesAsHtml from './html/renderMessagesAsHtml';
-import { getInputMessages } from './webViewHandleUpdates';
+import { getUpdateEvents } from './webViewHandleUpdates';
 import { handleMessageListEvent } from './webViewEventHandlers';
 import { base64Utf8Encode } from '../utils/encoding';
 
@@ -102,8 +102,8 @@ export type Props = {|
 class MessageList extends Component<Props> {
   context: Context;
   webview: ?{ postMessage: (string, string) => void };
-  sendMessagesIsReady: boolean;
-  unsentMessages: WebviewInputMessage[] = [];
+  sendUpdateEventsIsReady: boolean;
+  unsentUpdateEvents: WebViewUpdateEvent[] = [];
 
   static contextTypes = {
     styles: () => null,
@@ -111,7 +111,7 @@ class MessageList extends Component<Props> {
   };
 
   componentDidMount() {
-    this.setupSendMessages();
+    this.setupSendUpdateEvents();
   }
 
   handleError = (event: mixed) => {
@@ -121,27 +121,27 @@ class MessageList extends Component<Props> {
   /**
    * Initiate round-trip handshakes with the WebView, until one succeeds.
    */
-  setupSendMessages = (): void => {
+  setupSendUpdateEvents = (): void => {
     const intervalId = setInterval(() => {
-      if (!this.sendMessagesIsReady) {
-        this.sendMessages([{ type: 'ready' }]);
+      if (!this.sendUpdateEventsIsReady) {
+        this.sendUpdateEvents([{ type: 'ready' }]);
       } else {
         clearInterval(intervalId);
       }
     }, 30);
   };
 
-  sendMessages = (messages: WebviewInputMessage[]): void => {
-    if (this.webview && messages.length > 0) {
-      this.webview.postMessage(base64Utf8Encode(JSON.stringify(messages)), '*');
+  sendUpdateEvents = (uevents: WebViewUpdateEvent[]): void => {
+    if (this.webview && uevents.length > 0) {
+      this.webview.postMessage(base64Utf8Encode(JSON.stringify(uevents)), '*');
     }
   };
 
   handleMessage = (event: { nativeEvent: { data: string } }) => {
     const eventData: MessageListEvent = JSON.parse(event.nativeEvent.data);
     if (eventData.type === 'ready') {
-      this.sendMessagesIsReady = true;
-      this.sendMessages(this.unsentMessages);
+      this.sendUpdateEventsIsReady = true;
+      this.sendUpdateEvents(this.unsentUpdateEvents);
     } else {
       const { _ } = this.props;
       handleMessageListEvent(this.props, _, eventData);
@@ -149,12 +149,12 @@ class MessageList extends Component<Props> {
   };
 
   shouldComponentUpdate = (nextProps: Props) => {
-    const messages = getInputMessages(this.props, nextProps);
+    const uevents = getUpdateEvents(this.props, nextProps);
 
-    if (this.sendMessagesIsReady) {
-      this.sendMessages(messages);
+    if (this.sendUpdateEventsIsReady) {
+      this.sendUpdateEvents(uevents);
     } else {
-      this.unsentMessages.push(...messages);
+      this.unsentUpdateEvents.push(...uevents);
     }
 
     return false;
--- a/src/webview/__tests__/webViewHandleUpdates-test.js
+++ b/src/webview/__tests__/webViewHandleUpdates-test.js
@@ -1,8 +1,8 @@
-import { getInputMessages } from '../webViewHandleUpdates';
+import { getUpdateEvents } from '../webViewHandleUpdates';
 import { flagsStateToStringList } from '../html/messageAsHtml';
 import { HOME_NARROW } from '../../utils/narrow';
 
-describe('getInputMessages', () => {
+describe('getUpdateEvents', () => {
   test('missing prev and next props returns no messages', () => {
     const prevProps = {
       backgroundData: {},
@@ -11,7 +11,7 @@ describe('getInputMessages', () => {
       backgroundData: {},
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toEqual([]);
   });
@@ -26,7 +26,7 @@ describe('getInputMessages', () => {
       fetching: { older: false, newer: true },
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toEqual([
       {
@@ -47,7 +47,7 @@ describe('getInputMessages', () => {
       fetching: { older: false, newer: false },
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toEqual([]);
   });
@@ -62,7 +62,7 @@ describe('getInputMessages', () => {
       typingUsers: [{ id: 10 }],
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toHaveLength(1);
     expect(messages[0].type).toEqual('typing');
@@ -78,7 +78,7 @@ describe('getInputMessages', () => {
       renderedMessages: [],
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toEqual([]);
   });
@@ -111,7 +111,7 @@ describe('getInputMessages', () => {
       ],
     };
 
-    const messages = getInputMessages(prevProps, nextProps);
+    const messages = getUpdateEvents(prevProps, nextProps);
 
     expect(messages).toHaveLength(1);
     expect(messages[0].type).toEqual('content');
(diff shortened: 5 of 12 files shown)


borisyankov:
I did update the PR with the visual improvements we did when reviewing with @rishig and @gnprice 

Now ready for code review.

borisyankov:
Rebased.

gnprice:
Reading this code. Generally looks good! I will want to play with it, and poke at the code a bit harder.

> `const selector = msg.messageIds.map(id => `[data-msg-id="${id}"]`).join();`

Huh interesting! Just concatenating them makes an OR?

I wonder also if there's a more efficient way to do this:
```
const updateMessagesReadAttributes = rangeHull => {
  const messageElements = arrayFrom(document.querySelectorAll('[data-msg-id]'));
  messageElements.forEach(element => {
    const msgId = +element.getAttribute('data-msg-id');
    if (msgId >= rangeHull.first && msgId <= rangeHull.last) {
```

so we don't have to walk every message in the DOM in JS. One key fact which may help: the `first` and `last` really are the IDs of actual messages, or else `first > last` so you know the range is empty:
```
/**
 * Returns the IDs of the first and last visible messages, if any.
 *
 * If no messages are visible, the return value has first > last.
 */
```
So perhaps find `first`, then walk until reach `last`?

> webview: Fix generated JS code if it contains a `\`

I don't actually recall what specifically in the source code triggered this situation. Is this commit placed earlier in the branch than whatever caused the problem? If not, it should be :slightly_smiling_face: 

Also, if the problem is caused by something other than a literal `\` in the source code (i.e. something else causes Babel to emit code with a `\` in it), it'd be good to give a hint in the commit message about what that is/was.

> MessageInputReadReceipt,

I was concerned about this name, and then I saw in a subsequent commit you renamed it. :-) Let's give it the desired name (except the MessageInput -> WebViewUpdateEvent part) from the beginning -- that will (a) avoid an alarming earlier commit, and (b) make the big rename more regular and boring, which is good.

Also perhaps ...MessagesRead, plural? Since it carries a list of them.

gnprice:
@borisyankov What's the status of this PR -- is it ready for the next review, or is there more work you'd like to do on it first?

I just bumped the priority on it, because this is UI feedback that I think will help a lot to make people confident they understand what the app is doing. It will also probably surface some bugs that we get reports of but don't yet understand.

borisyankov:
> `const selector = msg.messageIds.map(id => [data-msg-id="${id}"]).join();`
>
> Huh interesting! Just concatenating them makes an OR?

Almost. Concatenating them with a comma does make it an 'and' as in `.class1,.class2 { }`
The default delimiter for `join` with no parameter is a comma.
I am changing the code to explicitly include the comma.

borisyankov:
> So perhaps find first, then walk until reach last?

Good idea for an optimization. I did the change, it looks like this:

```js
const updateMessagesReadAttributes = rangeHull => {
  let element = document.querySelector(`[data-msg-id='${rangeHull.first}']`);
  while (element && +element.getAttribute('data-msg-id') <= rangeHull.last) {
    element.setAttribute('data-read', 'true');
    element = element.nextElementSibling;
  }
};
```

borisyankov:
> I don't actually recall what specifically in the source code triggered this situation. Is this commit placed earlier in the branch than whatever caused the problem? If not, it should be 🙂

I see the commit message is descriptive now, I might have updated it after you've commented:
> In `generate-webview-js` Babel does its job perfectly but then we
put it inside a template literal. This changes the `\` symbol's
meaning to try and escape the next symbol instead of being itself.
>
> Before including the generated code, replace any single `\` with a
double `\\` to make sure the escaping results in a single `\`.

borisyankov:
> Let's give it the desired name (except the MessageInput -> WebViewUpdateEvent part) from the beginning -- that will (a) avoid an alarming earlier commit, and (b) make the big rename more regular and boring, which is good.

I did the change. It is a bit of a foreshadowing of the upcoming commit, as it is the first thing to be called a 'WebViewUpdateEvent'

> Also perhaps ...MessagesRead, plural? Since it carries a list of them.

Done. The final name is `WebViewUpdateEventMessagesRead`. A bit long, but I find this to disambiguate well the concepts of messaging between the WebView and the Zulip messages.

borisyankov:
@gnprice I think I addressed all of your comments. Time for another look?

gnprice:
Thanks @borisyankov for these revisions and replies!

#### 807d79d3d diagnostics: Remove 'Distinguish unread messages' option

* Looks good. I'll wait to merge until I'm ready to merge the new functionality.


#### 391a4cfcb css: Use 'loading' class in `MessageLoadingAsHtml`

* Cool, looks good. I may merge this now.


#### 463bee85c webview: Update message node's `data-read` when scrolling

```
+  setTimeout(() => {
+    updateMessagesReadAttributes(rangeHull);
+  }, 1000);
```

* What's the motivation for the 1-second delay?
  * If it's about the animation, can we control that with CSS instead?


```
+/**
+ * Update the 'data-read' attribute to a given range of message elements.
+ * This is styled with css to indicate visually what messages are being read.
+ */
+const updateMessagesReadAttributes = rangeHull => {
+  let element = document.querySelector(`[data-msg-id='${rangeHull.first}']`);
+  while (element && +element.getAttribute('data-msg-id') <= rangeHull.last) {
+    element.setAttribute('data-read', 'true');
+    element = element.nextElementSibling;
+  }
+};
```

* Thanks for the jsdoc!

* The name "update" is ambiguous about what it does; "set" would be a bit clearer that it goes one way.

* Some sibling elements of messages aren't themselves messages. (In the visible-messages algorithms above, I call these "message-peers".) They may not have `data-msg-id` attributes, which makes this logic tricky to reason about. And we shouldn't be setting `data-read` on them; might not be a live bug, but it's messy and is a latent bug.

* Thinking about this and rereading the code around it: I think probably the best way to implement this may actually be to stick this `setAttribute` line in the middle of the loop at the core of `visibleMessageIds`.
  * Just need to add a parameter to `visibleMessageIds` to control that behavior -- we don't want it at the other call site, initializing `prevMessageRange`.
  * The name and jsdoc should also update to match, but I can take care of that.

* Alternatively, we'd borrow a bit more of the logic from that loop, notably the `element.classList.contains`. Also, we can terminate the loop with an *equality* test for `last` -- that way we don't ever have to walk past potentially the very last message into the random other elements at the end.


#### 144901871 webiew: Add CSS for read/unread messages

```
 .message:not([data-read="true"]) {
+  box-shadow: inset 0.25em 0 ${BRAND_COLOR};
+}
+.message[data-read="true"] {
+  box-shadow: inset 0.25em 0 transparent;
+  transition: box-shadow 0.3s ease-out;
 }
```

* OK, this is definitely one of the places where the right units are `px`, not `em`. :stuck_out_tongue_closed_eyes: Looks like the webapp has 3px; perhaps do the same? Or 4px, or another value you think better.

gnprice:
> > I don't actually recall what specifically in the source code triggered this situation. Is this commit placed earlier in the branch than whatever caused the problem? If not, it should be slightly_smiling_face
> 
> I see the commit message is descriptive now, I might have updated it after you've commented:
> 
> > In `generate-webview-js` Babel does its job perfectly but then we
> > put it inside a template literal. This changes the `\` symbol's
> > meaning to try and escape the next symbol instead of being itself.
> > Before including the generated code, replace any single `\` with a
> > double `\\` to make sure the escaping results in a single `\`.

Sure, that's informative, but it doesn't at all answer the question you quoted. :wink: 

The question is easy to answer by taking this change out and running `tools/generate-webview-js`. The answer is, it's this line in the next commit:
```
+  const selector = msg.messageIds.map(id => `[data-msg-id="${id}"]`).join(',');
```
So, cool -- this commit is in the right spot in the branch.

Then that also helps answer this other question I asked:

> Also, if the problem is caused by something other than a literal `\` in the source code (i.e. something else causes Babel to emit code with a `\` in it), it'd be good to give a hint in the commit message about what that is/was.

There is indeed no `\` in the source code. What there is is a `"` inside a template literal, and for that Babel emits code with a `\` in it -- code like
```
`see: " a quote`
```
gets turned into `"see: \" a quote"`.

gnprice:
> > Let's give it the desired name (except the MessageInput -> WebViewUpdateEvent part) from the beginning -- that will (a) avoid an alarming earlier commit, and (b) make the big rename more regular and boring, which is good.
> 
> I did the change. It is a bit of a foreshadowing of the upcoming commit, as it is the first thing to be called a 'WebViewUpdateEvent'

OK. My suggestion (as quoted here) was to avoid the foreshadowing :wink: , but this is probably fine.


> > Also perhaps ...MessagesRead, plural? Since it carries a list of them.
> 
> Done. The final name is `WebViewUpdateEventMessagesRead`. A bit long, but I find this to disambiguate well the concepts of messaging between the WebView and the Zulip messages.

I agree -- sounds good!

gnprice:
#### 38a17f8e3 Send read updates from outside to the WebView

* Thanks for the updates! See also reply in previous comment.

* Nit: these names should match (note "message" singular):
`+const handleUpdateEventMessageRead = (msg: WebViewUpdateEventMessagesRead) =>`


#### 87123b502 Rename WebView 'messages' entity to 'UpdateEvent'

* Looks good!

* A few missed spots to rename -- places where we still call these "messages":

```
+  const messages: WebViewUpdateEvent[] = JSON.parse(decodedData);
+  messages.forEach((msg: WebViewUpdateEvent) => {
```
("messages", "msg")

```
+const handleUpdateEventContent = (msg: WebViewUpdateEventContent) => {
```
(and similar handlers' "msg")

```
webViewHandleUpdates.js
104:  const messages = [];

MessageList.js
105:  sendMessagesIsReady: boolean;
106:  unsentMessages: WebViewUpdateEvent[] = [];
124:  setupSendMessages = (): void => {
134:  sendMessages = (messages: WebViewUpdateEvent[]): void => {
152:    const messages = getUpdateEvents(this.props, nextProps);

js.js
564:  const messages: WebViewUpdateEvent[] = JSON.parse(decodedData);
565:  messages.forEach((msg: WebViewUpdateEvent) => {
```

borisyankov:
> What's the motivation for the 1-second delay?
> If it's about the animation, can we control that with CSS instead?

Yes, indeed. The delay gives the user a chance to notice the initial unread state before the indicator starts fading out. (I have added this explanation to the commit message)

I think both in-code and in-css are fine as solutions. I did make the change though, the JS no longer has the setTimeout, and the CSS now includes the `transition-delay` property. I have also unfolded the shorthand `transition` into 4 lines for clarity.

borisyankov:
> OK, this is definitely one of the places where the right units are px, not em. 😝 Looks like the webapp has 3px; perhaps do the same? Or 4px, or another value you think better.

That is actually one of the places where it makes sense. It was supposed to be 1/4th of the 'padding' which is 1em.

This PR #3188 makes the padding 16px so that would make the width '4px'. I did that change.

borisyankov:
> The name "update" is ambiguous about what it does; "set" would be a bit clearer that it goes one way.

Done.

borisyankov:
> A few missed spots to rename -- places where we still call these "messages":

Done. In addition, I did several more searches and found a couple of more missed variables to rename.

borisyankov:
> we'd borrow a bit more of the logic from that loop, notably the element.classList.contains. 

I did prefer that solution and did implement a check for 'message' class.


> Also, we can terminate the loop with an equality test for last -- that way we don't ever have to walk past potentially the very last message into the random other elements at the end.

We are already doing that or I did not understand you correctly?

borisyankov:
I think I have addressed the comments, so ready for another review.

gnprice:
Thanks @borisyankov for the revisions and replies!


#### 3077eb009 webview: Set message node's `data-read` when scrolling

> > What's the motivation for the 1-second delay?
> > If it's about the animation, can we control that with CSS instead?
> 
> Yes, indeed. The delay gives the user a chance to notice the initial unread state before the indicator starts fading out. (I have added this explanation to the commit message)
> 
> I think both in-code and in-css are fine as solutions. I did make the change though, the JS no longer has the setTimeout, and the CSS now includes the `transition-delay` property.

* Sounds good, thanks!

  One reason I like the CSS solution better is that it inherently makes it clear the purpose of the delay is just to adjust the animation -- because it no longer affects anything else. Info in comments and commit messages can be helpful, but it's always better to make things clear in the code itself when possible.

  The other main reason is that it lets the `data-read` attribute function purely as data -- it represents just the fact that the message has been read. With the `setTimeout`, there was also this intermediate state where we've identified it as read (and even sent a message/event to the outside, and posted the information to the server), but still have `data-read` as absent/false. That makes it more complicated to reason about the behavior, and seems like a recipe for confusion and future bugs.

  (In principle another reason would be that using CSS saves the JS engine from having to wake up later to do this, for a possible perf improvement. If that factor had gone the other way, though, I'd gladly trade it for the things above, in the absence of a known perf issue and measurements identifying this as a cause.)


> I have also unfolded the shorthand `transition` into 4 lines for clarity.

* Cool -- yeah, with two different time values I think that explicitness is helpful.

* Small nit: I think these are a bit easier to read when presented in a logical order, rather than alphabetical -- in particular I think `transition-property` is helpful to start with (just like in the shorthand) as it sets what the other properties are about.



> > we'd borrow a bit more of the logic from that loop, notably the element.classList.contains.
> 
> I did prefer that solution and did implement a check for 'message' class.

* Cool. I think a bit more is still needed here because of this point:

  > They may not have `data-msg-id` attributes, which makes this logic tricky to reason about.

  One approach would be this suggestion I made:

  > > Also, we can terminate the loop with an equality test for last -- that way we don't ever have to walk past potentially the very last message into the random other elements at the end.
  > 
  > We are already doing that or I did not understand you correctly?

  To answer this question: say `last` happens to be the last message ID of the list. Then we'll get to that message, and then go to its `nextElementSibling`... which right now is the `message-loading` element, and in general might be some random widget we've put at the end of the list.

  It has no `data-msg-id` attribute. Empirically, `getAttribute` then returns `null`... so unary `+` returns 0... and `0 <= rangeHull.last`, so the `while` condition is satisfied and the loop keeps going. We'll keep walking through the random end-of-list widgets, until the last one gives a `nextElementSibling` of `null`.

  I think this doesn't currently cause any live bugs... but it sure is a latent one, as it doesn't fit with how the loop looks like it's supposed to work.

  In short: let's write this so that we never try to interpret `null` as a number. (I'd also like to avoid assuming that the timerows and recipient headers, or other non-`message` elements, have `data-msg-id` attributes; they currently do but that's been its own source of bugs and I want to remove that attribute there.)


#### 4af9e7be7 webiew: Add CSS for read/unread messages

> > OK, this is definitely one of the places where the right units are px, not em. :stuck_out_tongue_closed_eyes: [..]
> 
> That is actually one of the places where it makes sense. It was supposed to be 1/4th of the 'padding' which is 1em.

* Ah, fair enough. What I was thinking is that this value inherently should be in px because it has nothing to do with text size... but because we currently use em for a closely-related value (which also inherently should be in px), there's a good reason locally to prefer em too.

  The real fix is, of course, that that other value (and others that aren't about text size) should be in px too. :wink: So:

> This PR #3188 makes the padding 16px so that would make the width '4px'. I did that change.

* Thanks for the revision and especially for that reference!

  That PR is mostly marked WIP so I won't try to review/merge most of it, but I may look in a bit to see if I can cherry-pick the part that makes that matching change.


#### b5f4fbe06 webview: Fix generated JS code if it contains a `\`

* (Reminder to myself to add to this commit message the answers to my questions, discussed above.)


#### a002b7448 Send read updates from outside to the WebView

> * Nit: these names should match (note "message" singular):

* This should get fixed within this commit, rather than in the unrelated next commit.


#### d06d325c8 Rename WebView 'messages' entity to 'UpdateEvent'

> > A few missed spots to rename -- places where we still call these "messages":
> 
> Done. In addition, I did several more searches and found a couple of more missed variables to rename.

* Thanks!

* I think there's one spot where you renamed something that actually shouldn't get this name:

```
-const sendMessage = (msg: MessageListEvent) => {
-  window.postMessage(JSON.stringify(msg), '*');
+const sendMessage = (uevent: MessageListEvent) => {
+  window.postMessage(JSON.stringify(uevent), '*');
```

That's in `js.js` and is for sending data from the WebView out to the enclosing MessageList, rather than vice versa. As the type annotation says, the object isn't an "UpdateEvent" but rather a `MessageListEvent`.

We can likely improve on the name `msg` here too, but it shouldn't be `uevent`.

borisyankov:
> I think these are a bit easier to read when presented in a logical order...

Done. Now ordered by the shorthand order.

borisyankov:
>  let's write this so that we never try to interpret null as a number. 

Done. Now we check with `hasAttribute` before getting the value and converting it to a number.

borisyankov:
> We can likely improve on the name msg here too, but it shouldn't be uevent.

I did undo this change completely. The name should be something like `msgData` or `eventData` which is the name used in `MessageList.js` (mostly because it comes from event.nativeEvent.data)

borisyankov:
> > Nit: these names should match (note "message" singular):
> This should get fixed within this commit, rather than in the unrelated next commit.

Fixed.

borisyankov:
Addressed the latest comments.
Wanted to address '(Reminder to myself to add to this commit message the answers to my questions, discussed above.)' but I guess you best know what you wanted to add? (as the commit message seemed sufficient to me)

gnprice:
Thanks for these revisions! Merged. :tada: 

> (I'd also like to avoid assuming that the timerows and recipient headers, or other non-`message` elements, have `data-msg-id` attributes; they currently do but that's been its own source of bugs and I want to remove that attribute there.)

I inserted a new commit 17dab9f28 to take care of this concern.


> Wanted to address '(Reminder to myself to add to this commit message the answers to my questions, discussed above.)' but I guess you best know what you wanted to add? (as the commit message seemed sufficient to me)

Yeah. It seemed like I hadn't successfully communicated what I was looking for here, and I decided the most effective way to communicate that would be to write it myself. :-) Take a look in ec8ad64bf.

PR #3127, opened 2018-11-09
Use a trailing slash when doing a SSO login

### LINKED ISSUE #3126: Fix "Login with SSO" in the mobile app.
From https://github.com/zulip/zulip/pull/10788#issuecomment-437301602: @gnprice @borisyankov it appears that the mobile implementation of SSO login accidentally has the trailing slash stripped for `/accounts/login/sso`; my guess is the right fix here is to just change that end (since the general convention for the Zulip webapp is trailing slashes in URLS; it'll redirect to add the slash if omitted, but probably that redirect isn't working with the mobile HTTP library).  

I suspect @codebit-de would be up for testing.

(This is for https://zulip.readthedocs.io/en/latest/production/authentication-methods.html#apache-based-sso-with-remote-user, which I thought we'd tested manually when we added the code, but I can easily see a typo having gotten in there between testing and merge and it never worked after merge, since that backend isn't widely used, and a lot of sites using it have Zulip running behind a corporate firewall that they don't let phones on).

borisyankov:
Fixes #3126

Adds a trailing slash to the `accounts/login/sso` path when doing
OAuth. The convention for the Zulip webapp is to include trailing
slashes in URLS.

The `accounts/login/social/github` just next to it is left as-is
because it actually does not work with a trailing slash currently.

### DIFF
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -98,7 +98,7 @@ class AuthScreen extends PureComponent<Props> {
   };
 
   handleSso = () => {
-    this.beginOAuth('accounts/login/sso');
+    this.beginOAuth('accounts/login/sso/');
   };
 
   render() {


timabbott:
Re; GitHub: yeah, the python-social-auth URLs seem to violate our convention for auth pages (which actually has a bunch of violations within that context; and our API routes of course use a no-trailing-slash convention).  May be worth a bit of a discussion on what URL convention we want and migrating towards that...

gnprice:
Merged -- thanks @borisyankov for the swift fix!

PR #3134, opened 2018-11-12
flow, userItem: Make `avatarUrl` optional.

jainkuniya:
If user have not set custom (image) avatar then, we are required to
generate gavatar url (hashing email addredd) on the client side. So in
such cases we don't have avatar url. UserItem takes avatarUrl and
passes it to Avatar, in which it can be null/undefined. So make it
optional in UserItem too.

### DIFF
--- a/src/users/UserItem.js
+++ b/src/users/UserItem.js
@@ -28,7 +28,7 @@ const componentStyles = StyleSheet.create({
 type Props = {
   email: string,
   fullName: string,
-  avatarUrl: string,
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @jainkuniya ! Merged.

One small but meaningful detail I fixed in the commit message: this prop is not "optional", meaning that the caller is always expected to pass a value for it.

But it is a "maybe" type (or "nullable" as some languages call it), in that the value might be `null`, or `undefined`.

PR #3137, opened 2018-11-13
Fix message loading indicators not showing

borisyankov:
When we navigate to a new narrow if there are three cases regarding
the existance of messages:
 * there are no messages
 * there are messages
 * there are currently no messages but fetching

In case of 3) we are showing a 'message loading indicator'.

Currently though, we are showing it quite rarely. This is caused
by us first navigating to the narrow and then staring to load new
messages (the logic is 'show if no messages but fetching').

We are smart enough to update that state even if the order is
navigation then loading, but it takes some time. Our new handshake
algorithm slows things down a bit too. Thus it seems we no longer
show the loading indicator in many cases.

The solution is simple and effective. First do the fetching, then
do the actual navigation to the narrow.

### DIFF
--- a/src/message/__tests__/messageActions-test.js
+++ b/src/message/__tests__/messageActions-test.js
@@ -33,9 +33,9 @@ describe('messageActions', () => {
 
       expect(actions).toHaveLength(3);
       expect(actions[0].type).toBe('FETCH_STATE_RESET');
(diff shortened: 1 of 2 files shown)


gnprice:
Merged -- thanks @borisyankov for the fix and the detailed explanation!

Issue #3131, opened 2018-11-10
Stream creation is missing some recent features


gnprice:
@rishig mentioned that some recently-added features that show up when creating a stream haven't been implemented for mobile.
* One is "private with shared history": #2683 
* I don't know what the others are.

@rishig , would you comment with the others you know about?

gnprice:
Recent (unreleased, post-2.1) server versions no longer have `announcement_only` and instead have a `stream_post_policy` enum. Discussion here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Add.20functionality.20to.20create.20announcement.20streams/near/855688
and docs here:
https://zulipchat.com/api/add-subscriptions

gnprice:
Bumping to note that with #4079 / #4049 we now have the server's feature level in Redux. That was something we expected we'll want as part of doing this correctly, so this is now more ripe to implement.

PR #3140, opened 2018-11-14
 createStream: Add functionality to create announcement streams. Part of #3131

jainkuniya:
Part of #3131

image

### DIFF
--- a/src/api/streams/createStream.js
+++ b/src/api/streams/createStream.js
@@ -9,11 +9,13 @@ export default (
   description?: string = '',
   principals?: string[] = [],
   inviteOnly?: boolean = false,
+  streamPostPolicy?: number = 1,
   announce?: boolean = false,
 ): Promise<ApiResponse> =>
   apiPost(auth, 'users/me/subscriptions', {
     subscriptions: JSON.stringify([{ name, description }]),
     principals: JSON.stringify(principals),
     invite_only: inviteOnly,
+    stream_post_policy: streamPostPolicy,
     announce,
   });
--- a/src/streams/CreateStreamScreen.js
+++ b/src/streams/CreateStreamScreen.js
@@ -14,10 +14,15 @@ type Props = $ReadOnly<{|
 |}>;
 
 class CreateStreamScreen extends PureComponent<Props> {
-  handleComplete = (name: string, description: string, isPrivate: boolean) => {
+  handleComplete = (
+    name: string,
+    description: string,
+    isPrivate: boolean,
+    streamPostPolicy: number,
+  ) => {
     const { dispatch, ownEmail } = this.props;
 
-    dispatch(createNewStream(name, description, [ownEmail], isPrivate));
+    dispatch(createNewStream(name, description, [ownEmail], isPrivate, streamPostPolicy));
     dispatch(navigateBack());
   };
 
@@ -26,7 +31,12 @@ class CreateStreamScreen extends PureComponent<Props> {
       <Screen title="Create new stream" padding>
         <EditStreamCard
           isNewStream
-          initialValues={{ name: '', description: '', invite_only: false }}
+          initialValues={{
+            name: '',
+            description: '',
+            invite_only: false,
+            stream_post_policy: 1,
+          }}
           onComplete={this.handleComplete}
         />
       </Screen>
--- a/src/streams/EditStreamCard.js
+++ b/src/streams/EditStreamCard.js
@@ -4,10 +4,18 @@ import { StyleSheet, View } from 'react-native';
 
 import type { Dispatch } from '../types';
 import { connect } from '../react-redux';
-import { Input, Label, OptionRow, ZulipButton } from '../common';
+import { Input, Label, OptionRow, ZulipButton, SelectionList } from '../common';
 import styles from '../styles';
 import { IconPrivate } from '../common/Icons';
 import { getIsAdmin } from '../selectors';
+
+type StreamPostPolicy = $ReadOnly<{| key: number, value: string |}>;
+
+const STREAM_POST_POLICIES: StreamPostPolicy[] = [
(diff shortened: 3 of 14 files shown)


src/api/streams/createStream.js
```diff
@@ -8,11 +8,13 @@ export default (
   description?: string = '',
```
borisyankov: The last line of the commit message might be a bit premature :) (and can be omitted!)

src/streams/streamsActions.js
```diff
@@ -43,6 +43,14 @@ export const updateExistingStream = (
   if (initialValues.invite_only !== newValues.isPrivate) {
     await updateStream(getAuth(getState()), id, 'is_private', newValues.isPrivate);
   }
+  if (initialValues.is_announcement_only !== newValues.isAnnouncementOnly) {
```
borisyankov: This line of code looks OK but I am starting to be concerned we are calling `updateStream` too many times. Is there a 'batch update' version of this API call?

src/streams/CreateStreamScreen.js
```diff
@@ -16,10 +16,15 @@ type Props = {
 class CreateStreamScreen extends PureComponent<Props> {
   props: Props;
 
-  handleComplete = (name: string, description: string, isPrivate: boolean) => {
+  handleComplete = (
+    name: string,
+    description: string,
+    isPrivate: boolean,
+    isAnnounceOnly: boolean,
```
borisyankov: Should this be `isAnnouncementOnly`? (announce => announcement)

src/streams/EditStreamCard.js
```diff
@@ -79,6 +91,11 @@ export default class EditStreamCard extends PureComponent<Props, State> {
           defaultValue={initialValues.invite_only}
           onValueChange={this.handleIsPrivateChange}
         />
+        <OptionRow
+          label="Restrict posting to organization administrators"
```
borisyankov: We will need this in `messages_en.json` too.

borisyankov:
Do we also need to expose the option of `history_public_to_subscribers`?

Note about English :) ... `it's === it is` and you mean `its` a few times in the commit messages.

gnprice:
Thanks @jainkuniya ! I agree with all @borisyankov 's comments.

Another thought:
* This list of fields `name`, `description`, `inviteOnly`, and now `isAnnouncementOnly` appears in a lot of places, as your commit "createStream: Add functionality to ..." (btw "to to" -> "to") demonstrates. Also a lot of those places are positional arguments, and increasingly long lists of positional arguments make me nervous.
* How about we move these four into an object type of their own, like `StreamOptions`? Then all the places that pass around those four fields together can instead pass around one `streamOptions: StreamOptions`.
* That can come after this change. But it will be a good thing to do before we add a fifth option `historyPublicToSubscribers` :slightly_smiling_face: .

borisyankov:
PS. The 'Deuglify' PR #3143 was inspired by these changes here and that the screen is getting busy.

@jainkuniya you should keep an eye on this PR :)

src/streams/CreateStreamScreen.js
```diff
@@ -14,10 +14,17 @@ type Props = {|
 |};
 
 class CreateStreamScreen extends PureComponent<Props> {
-  handleComplete = (name: string, description: string, isPrivate: boolean) => {
+  props: Props;
```
borisyankov: We no longer include `props: Props;`

src/streams/streamsActions.js
jainkuniya: submitted #3224 :)

src/streams/CreateStreamScreen.js
jainkuniya: Done.

I missed it while rebasing.

gnprice:
Hi @jainkuniya ! I was just looking back through our P1 issues, and I see the one this PR is for. I'd love to pick it back up and get it merged. :slightly_smiling_face:

Would you rebase, and also take a look through and make any updates that seem like a good idea from looking at it again a while later? Then I'll give it a fresh review.

gnprice:
There's some recent chat discussion in this thread:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Add.20functionality.20to.20create.20announcement.20streams/near/847205

gnprice:
Thanks @jainkuniya ! Some comments from our discussion today:
* Let's call it something other than "picker", because it turns out that means something different from this in both the Apple and Google guidelines.

  There isn't a great standard name for this. But we could call it something like "SelectionList".

* Let's put them both in one file. They're so closely related -- each one only makes sense together with the other.

* Instead of Zulip green, let's make the text gray -- same as you see here: https://material.io/components/lists and in the Android system settings app.

* For the text of the options, use the same text that's in the webapp

* The spacing looks off. Like there's a bunch of space above "Private", and much less space between that and who-can-post, or between the latter and the "Create" button. Let's fix the spacing -- add more vertical margin or padding.

  Take a look at https://material.io/components/lists#specs and specifically "Two line", and check we're conforming to that.

* The alignment feels a bit off to me too, like the horizontal padding and margin should be adjusted as well. So it'd be good to check on that too.

---

A couple of further thoughts, on workflow:
* Let's make a fresh PR and close this one, because so much has changed from the version that was reviewed in this thread.
* In the next revision, please post screenshots.

jainkuniya:
Github is not allowing me to create new PR without closing this, so closing this. And for reference will link other PR number in another comment

jainkuniya:
This is moved to https://github.com/zulip/zulip-mobile/pull/4066

gnprice:
> Github is not allowing me to create new PR without closing this

(I think specifically this is because you were using the same branch name.)

Closing this thread seems like the right thing, so :+1: -- discussion continues on #4066.

PR #3141, opened 2018-11-14
WIP: Exception diagnostics

borisyankov:
This clearly needs more work.
Pushed more for a discussion, that we missed doing.


borisyankov:
Closing, as too early in the process and unlikely to be implemented.

PR #3143, opened 2018-11-15
Deuglify UI

borisyankov:
Multiple minor tweaks to the UI.
No UI element is changed or new concept introduced.
Just much-needed polish.

### DIFF
--- a/src/chat/UnreadNotice.js
+++ b/src/chat/UnreadNotice.js
@@ -12,7 +12,7 @@ import AnimatedScaleComponent from '../animation/AnimatedScaleComponent';
 
 const styles = StyleSheet.create({
   unreadContainer: {
-    padding: 4,
+    padding: 8,
     backgroundColor: '#96A3F9',
     flexDirection: 'row',
     justifyContent: 'space-between',
@@ -22,10 +22,14 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     alignItems: 'center',
   },
+  unreadNumber: {
+    fontSize: 14,
+    color: 'white',
+    paddingRight: 4,
+  },
   unreadText: {
     fontSize: 14,
     color: 'white',
-    padding: 2,
   },
 });
 
@@ -48,7 +52,7 @@ class UnreadNotice extends PureComponent<Props> {
       <AnimatedScaleComponent visible={unreadCount > 0}>
         <View style={styles.unreadContainer}>
           <View style={styles.unreadTextWrapper}>
-            <RawLabel style={styles.unreadText} text={unreadCount.toString()} />
+            <RawLabel style={styles.unreadNumber} text={unreadCount.toString()} />
             <Label
               style={styles.unreadText}
               text={unreadCount === 1 ? 'unread message' : 'unread messages'}
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -5,6 +5,7 @@ import type { Text } from 'react-native';
 import Feather from 'react-native-vector-icons/Feather';
 import IoniconsIcon from 'react-native-vector-icons/Ionicons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
+import SimpleLineIcons from 'react-native-vector-icons/SimpleLineIcons';
 
 /*
  * This corresponds to the documented interface:
@@ -31,6 +32,7 @@ export const IconWarning: IconType = props => <Feather name="alert-triangle" {..
 export const IconSend: IconType = props => <MaterialIcon name="send" {...props} />;
 export const IconMute: IconType = props => <MaterialIcon name="volume-off" {...props} />;
 export const IconStream: IconType = props => <Feather name="hash" {...props} />;
+export const IconPin: IconType = props => <SimpleLineIcons name="pin" {...props} />;
 export const IconPrivate: IconType = props => <Feather name="lock" {...props} />;
 export const IconPrivateChat: IconType = props => <Feather name="mail" {...props} />;
 export const IconDownArrow: IconType = props => <Feather name="chevron-down" {...props} />;
--- a/src/common/ZulipButton.js
+++ b/src/common/ZulipButton.js
@@ -47,8 +47,15 @@ const styles = StyleSheet.create({
   secondaryIcon: {
     color: BRAND_COLOR,
   },
-  disabled: {
-    opacity: 0.25,
+  disabledPrimaryFrame: {
+    backgroundColor: 'rgba(127, 127, 127, 0.4)',
+  },
+  disabledSecondaryFrame: {
+    borderWidth: 1.5,
+    borderColor: 'rgba(127, 127, 127, 0.4)',
+  },
+  disabledText: {
+    color: 'rgba(127, 127, 127, 0.8)',
   },
 });
 
@@ -88,11 +95,19 @@ export default class ZulipButton extends PureComponent<Props> {
     const { style, text, disabled, secondary, progress, onPress, Icon } = this.props;
     const frameStyle = [
       styles.frame,
-      secondary ? styles.secondaryFrame : styles.primaryFrame,
-      disabled && styles.disabled,
+      disabled
+        ? secondary
+          ? styles.disabledSecondaryFrame
+          : styles.disabledPrimaryFrame
+        : secondary
+          ? styles.secondaryFrame
+          : styles.primaryFrame,
       style,
     ];
-    const textStyle = [styles.text, secondary ? styles.secondaryText : styles.primaryText];
+    const textStyle = [
+      styles.text,
+      disabled ? styles.disabledText : secondary ? styles.secondaryText : styles.primaryText,
+    ];
     const iconStyle = [styles.icon, secondary ? styles.secondaryIcon : styles.primaryIcon];
 
     if (progress) {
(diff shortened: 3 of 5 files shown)


borisyankov:
Ready for review... I'd say manual test, as this is hard to code-review, and its changes are unlikely to break anything. (and very likely to deuglify many things)!

gnprice:
I just rebased this branch, resolving various small conflicts.

NB there were some conflict markers already in the existing version of the branch; I didn't go in and attempt to sort those out.

gnprice:
Looking through these. Thanks @borisyankov for these changes!

#### daa7334f4 ui: Refactor Account List spacing on Android

* Looks good, thanks!


#### ee789dbf4 ui: Align correctly `OptionRow` in `EditStreamCard`

* Hmm, the alignment of this at the tip of the branch looks actually pretty similar to what it is on master. So, not sure I understand.

* But this is a nicely circumscribed change, and the screen looks fine in the end, so shrug, happy to merge as part of the branch.


#### a9f0f7218 ui: Style disabled buttons according to Material Design guidelines

* Hmm, here's what this looks like:

  image

  That's pretty low contrast -- I think it'd be hard for many people to read. Is this really what Material recommends?

  Ah, in fact here's what I see [in the Material guidelines](https://material.io/design/components/buttons.html#toggle-button), and indeed it looks pretty different:

  image

  So I'd be happy to change our disabled-button style to match that.


#### 7403e42f2 ui: Add icons to important OptionRow components

* I like this idea in general. And definitely happy with the "private stream", "muted" and "notifications" icons.

* The icon here for "pinned stream" is one I strongly associate with "location on a map" -- e.g. on a calendar event it identifies the field for where in the world / the city the event will happen.

  image

  In the webapp, we use an icon that doesn't have that association and feels compatible for me with "pin to top / keep as important":

  image


#### 72eb2f749 stream screen: Remove "Topics" button

* Sounds good!


#### d0d9a53f7 StreamScreen: Add icons to buttons and style differently

* Looks good, thanks!


#### d3246a0ff ui: Add space between stream icon and name in nav bar

* Oh wow, I hadn't noticed that :-/ -- yep, definitely good to fix.


#### 3006968ed ui: Space the `OptionRow` icon better

* This one has merge-conflict markers in it, so it breaks the app.

* I'm kind of confused reading it, too -- why the name `optionRowIcon` when you're adding a use of it in `OptionButton`? And it's also used in `OptionRow`, but that site is left still saying `settingsIcon` which doesn't exist anymore.

  I suspect this is all due to a mismerge; if you'd like to make the change afresh on the current code, I think that'd help a lot.


#### c0c038160 ui: Add needed spacing in EditStreamCard

* Hmm -- the resulting space actually looks pretty excessive, at least on Android. And in fact on master it feels like we have plenty of spacing here already, again on Android.

  Perhaps the change should be platform-dependent? That's a bit messy, though.

* Also, there are various other places we have text fields with labels; notably the auth screens. Are those good already, or do they also need this change? Can we do it in a shared way?


#### 99410e78a ui: Add better spacing in StreamCard

* LGTM, thanks!


#### 196ab9b92 ui: Less space dividers

* Sure. This is rather narrower than the commit message says, though -- just one divider, on one screen, is affected :wink: 


#### 9b854c460 ui: Better spacing in `UnreadNotice`

* Hmm, the existing left padding aligns with the topics in topic headers. Not sure this feels more aligned to me on that side, rather than less.

* Not sure I like the extra vertical space taken up, either.

  ... In fact, trying it with the keyboard open, in a stream narrow where I have some unreads, we're painfully short of vertical space:

  image

  That's an emulated Nexus 5, at default display settings (vs. the make-things-smaller settings I keep on my own device.) It's clearly going to be pretty bad even without this change, but every little counts.

* The handling of "N unread messages" is already bad -- it's fundamentally impossible to translate properly. ... Oh, rereading, I guess this doesn't change the behavior of the padding between the two pieces (which substitutes for a proper text space). So, sure, a refactor there is fine. Though really we should replace it with proper i18n anyway.

gnprice:
I think the 7 commits in this branch that LGTM all work fine without the other commits that aren't ready -- so, just merged those!

9ff14aa2b ui: Refactor Account List spacing on Android
08864673d ui: Align correctly `OptionRow` in `EditStreamCard`
fe7ec662e stream screen: Remove "Topics" button
265100e3a stream screen: Add icons to buttons and style differently
a7cc98ad8 ui: Add space between stream icon and name in nav bar
ce3411a25 ui: Add better spacing in StreamCard
435f403eb ui: Remove the divider in stream screen.

gnprice:
And updated the PR to rebase the remaining 5 commits:

f8efc46c8 ui: Style disabled buttons according to Material Design guidelines
d282be5aa ui: Add icons to important OptionRow components
127cd8b4f ui: Space the `OptionRow` icon better
745a5e334 ui: Add needed spacing in EditStreamCard
faca55096 ui: Better spacing in `UnreadNotice`

I'm pretty sure most of these are changes we want, modulo just the specific comments above -- so, would be quite glad to see a further revision of them in order to merge some more!

borisyankov:
> Ah, in fact here's what I see in the Material guidelines, and indeed it looks pretty different:

That is a toggle button. Maybe you did link the incorrect image because there is the correct button on the same page next to it.

This is what a button + disabled (2) looks like:



I will check if the values I did use are the same, if not will change to the exact ones.

borisyankov:
> I'm kind of confused reading it, too -- why the name optionRowIcon when you're adding a use of it in OptionButton? 

I think we did merge another part of this few months ago and now it is a less useful commit, I did drop it.

borisyankov:
The 'ui: Style disabled buttons' commit is entirely redone. Now looks like the guidelines, looks good on night theme and includes styles for primary/secondary.

borisyankov:
> The icon here for "pinned stream" is one I strongly associate with "location on a map" -- e.g. on a calendar event it identifies the field for where in the world / the city the event will happen.

Fair enough. Changed to this one:

image

borisyankov:
> Hmm, the existing left padding aligns with the topics in topic headers. Not sure this feels more aligned to me on that side, rather than less.

I think the topics need realignment, but I got your point about the label not having much space to work with. Still, the previous padding of 4 was too low. I did reduce it from 16 to 8 as 'better than 4, way less than 16' value.

borisyankov:
> Hmm -- the resulting space actually looks pretty excessive, at least on Android. And in fact on master it feels like we have plenty of spacing here already, again on Android.

I think you are just used to looking at the uncomfortably stuffed UI of the web app, and places of the mobile app.

I did compare screenshots of this commit vs Slack. Amazingly they almost align perfectly:

image

So I think this is good on Android. Without that, on iOS it looks ridiculous.

gnprice:
#### cb31d8204 ui: Style disabled buttons according to Material Design guidelines

> That is a toggle button. Maybe you did link the incorrect image because there is the correct button on the same page next to it.
> 
> This is what a button + disabled (2) looks like:

* Ah got it :slightly_smiling_face: -- and thanks for the image, and the specific links in the commit message. I've found the Material docs unfortunately kind of hard to find things in :-/ (... and in fact sometimes I end up referring to [the pre-2018 "archive" version](https://material.io/archive/guidelines/) because it's better organized.) So those links are extra helpful for partially working around that.

* And thanks for the update -- LGTM!


#### f28300511 ui: Add icons to important OptionRow components

* New icon LGTM; thanks!


#### 652330d93 ui: Better spacing in `UnreadNotice`

(out of order, discussing the simpler ones first)

* > I think the topics need realignment, but I got your point about the label not having much space to work with. Still, the previous padding of 4 was too low. I did reduce it from 16 to 8 as 'better than 4, way less than 16' value.

  Cool, the alignment looks good now -- thanks!

* I'll take out the change to vertical padding, and merge.


#### 65de43644 ui: Add needed spacing in EditStreamCard

> > Hmm -- the resulting space actually looks pretty excessive, at least on Android. And in fact on master it feels like we have plenty of spacing here already, again on Android.
> 
> I think you are just used to looking at the uncomfortably stuffed UI of the web app, and places of the mobile app.

Haha, possible. :smile: 


> I did compare screenshots of this commit vs Slack. Amazingly they almost align perfectly:
> 
> image

Hmm. Is this with your changes? Here's what I see:

image

That's at this commit, in an emulated Nexus 5, with default display settings.

Your screenshot above has quite a bit less spacing between the label and the field than this screenshot does. I think yours is consistent with what I see on master. That makes it look like evidence that we should leave this as is. :wink: 

With the change, one concern I have is that the field ends up as close to the label on the *next* field as to its own -- it ends up looking ambiguous. And in fact if starting from the label, I think in the screenshot above, "Description" actually looks *closer* to the field above it than to the one below it.


> Without that, on iOS it looks ridiculous.

* On iOS, the input field is styled quite differently, and I agree the label is shoved up too close to the input field and looks cramped.

  So I think the right fix here will be platform-dependent... or possibly there's an existing platform dependence that we'll want to adjust.

  That definitely feels like this component isn't the right home for it, though. Which leads to:

* > Also, there are various other places we have text fields with labels; notably the auth screens. Are those good already, or do they also need this change? Can we do it in a shared way?

gnprice:
And merged 3 of those 4 remaining commits! Thanks again.

3d2c83f81 ui: Style disabled buttons according to Material Design guidelines
f68735596 ui: Add icons to important OptionRow components
3225d1915 ui: Better spacing in `UnreadNotice`

The only one still outstanding is now

c1893cfa8 ui: Add needed spacing in EditStreamCard

borisyankov:
Apparently, I have compared our current app with Slack!
So, indeed our previous spacing between label and input is better!

The only improvement in future would be to make the label smaller and dimmer (75% of font size and maybe opacity: 0.75)

PR #3144, opened 2018-11-15
Upgrade to React Native 0.57.7

borisyankov:
Fixes the iOS build issues.

Changelog here:
https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md

All dependencies that come with a fresh RN install are updated.


gnprice:
Thanks @borisyankov ! Merged, after making my own pass through reading the changelog.

See also the commit I added on top (pushed here.)

PR #3148, opened 2018-11-16
apiFetch: Fix regression in using `console.warn`.

jainkuniya:
`console.warn` takes string as parameter, not a object. Before we were
passing object, which again raised an exception and it was direclty
going to catch, instead of creating and throwing an error from
`makeApiError.`

### DIFF
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -62,7 +62,7 @@ export const apiCall = async (
       return resFunc(json);
     }
     // eslint-disable-next-line no-console
-    console.warn({ route, params, httpStatus: response.status, json });
+    console.log({ route, params, httpStatus: response.status, json });
     throw makeApiError(response.status, json);
   } finally {
     networkActivityStop(isSilent);


borisyankov:
Are you sure about that?
I did expect `log`, `warn` and `error` to accept the same parameters.
I did test a `console.warn({ a: 1, b: 2 });` in the console just now and it works fine.

jainkuniya:
> Are you sure about that?

I also expected the same, but on mobile it is not working.

it is throwing this error
image

borisyankov:
Wow. That is unpleasant. Good catch.
I wonder if it is Sentry that is overwriting that function. If not, it is React Native.

It is great if you can figure out what part of the stack is causing this issue.

gnprice:
Thanks @jainkuniya for catching this!

There's an unfortunate downside to this version of the fix -- calling `console.log` and friends with an object, when you're using something like the Chrome Developer Tools to debug, produces a nice pretty-printed and interactively expandable/collapsible display. It'd be a shame to lose that in the case where it's available.

I agree with Boris :grin: :

> It is great if you can figure out what part of the stack is causing this issue.

Would you dig into this some more? And then see how we might be able to fall back to a string where we have to, while sticking with an object where we can.

borisyankov:
I think `console.warn` really expects a string.
I am dubious that calling `indexOf` on an array, while not throwing an error, will preserve the intended by RN behavior 😄 

`JSON.stringify` is certainly a reasonable option, but makes reading longer objects less pleasant.
I think going to a basic `console.log` is the best option here.

borisyankov:
Also, bonus points for examining all the other `console.warn` and `console.error` (likely suffers from the same issue) calls for usage of an object and fixing this too.

jainkuniya:
@borisyankov 
We are using `console.error` at only one place:
In `MessageList` when it is called, red screen pops in the mobile. So for the release build, I thinks app will crash, and gets recored on the Sentry.  If we change it to `console.log` then we need to keep track on chrome debugger for such errors.
So I thinks we can keep it as it is.

And for `console.warn`:
it is also used at only one place, `apiFetch`.

>I think going to a basic console.log is the best option here.

As with the use of `array` we can get advantage of warn + better reading formats, we can go with this :)

gnprice:
I don't quite understand the title line:
> apiFetch: Fix reg in `console.warn`.

What do you mean by "reg"?

> Generally`console.warn` should take string/object as parameter, but with string it is throwing error

Do you mean "with an object"? We aren't passing it a string.

> > YellowBox can only display strings, not objects.

Well, if this were true this fix wouldn't be right :slightly_smiling_face: since it's passing an Array, which is still not a string.

What's the behavior you observe with this solution?

---

In the other answer you got on that SO thread:

> w3schools.com cannot be used as reliable source of information.

Oh man, so meta :stuck_out_tongue: -- StackOverflow cannot be used as a reliable source of information either. (Though often a very useful one for getting a possible answer, or helping find the information in a reliable source!)

But MDN is indeed a much better source than "w3schools". In general I recommend doing a Google query like `mdn console.warn` to quickly navigate to the good stuff.

borisyankov:
I think the optimal and simplest solution is to use `.log` everywhere :)

jainkuniya:
> What do you mean by "reg"?

By "reg", I mean "regression".

By taking help from `Array` in `console.warn` (previous version of this PR) will still be a hack and that format will also be not readable in YellowBox (which is the main purpose of `console,warn`)

Now, I have replaced `console.warn` with `console.log`.

gnprice:
> By "reg", I mean "regression".

Got it, thanks. I had not ever heard that usage before. :slightly_smiling_face: 

Merging -- thanks for the fix!

PR #3149, opened 2018-11-16
messageEditDialog: Fix error message is not shown.

jainkuniya:
We get error message as a result from api call. In apiFetch if any
request fails, we are throwing and error where `error.data` contains
response data.

So get error msg from this data object and show it to the user.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -233,7 +233,7 @@ class ComposeBox extends PureComponent<Props, State> {
     const subject = topic !== editMessage.topic ? topic : undefined;
     if (content || subject) {
       updateMessage(auth, { content, subject }, editMessage.id).catch(error => {
-        showErrorAlert(error.message, 'Failed to edit message');
+        showErrorAlert(error.data && error.data.msg, 'Failed to edit message');
       });
     }
     dispatch(cancelEditMessage());


src/compose/ComposeBox.js
```diff
@@ -233,7 +233,7 @@ class ComposeBox extends PureComponent<Props, State> {
     const subject = topic !== editMessage.topic ? topic : undefined;
     if (content || subject) {
       updateMessage(auth, { content, subject }, editMessage.id).catch(error => {
-        showErrorAlert(error.message, 'Failed to edit message');
```
borisyankov: I wonder if we should be setting `error.message` correctly instread. Also, is `error` ever not defined?

jainkuniya: > error ever not defined?

Just wanted to be extra safe 😅

jainkuniya: > I wonder if we should be setting error.message correctly instread.

@gnprice Thoughts?

borisyankov: > > error ever not defined?
> 
> Just wanted to be extra safe 😅

Don't be :)
In fact, any check, in this case, might be 'too safe' as the only exception we are throwing is an API exception.

jainkuniya: > Don't be :)

Okay, cool! updated 👍

gnprice:
Hmm interesting - this sounds kind of like your earlier PR #2871 , which I just looked at with @borisyankov . :slightly_smiling_face: 

What's the relationship between these two PRs? Or between the issues they're addressing?

Also, as with the other one -- would you post a screenshot of an example of what the UI looks like after this change?

jainkuniya:
@gnprice 
I have rebased and updated #2871.
This PR is just same as that of.

jainkuniya:
Closing now, as it is duplicate of #2871

PR #3151, opened 2018-11-16
Invalidate old messages less abruptly (was: Fixes 'No messages' on queue restart)

borisyankov:
Not yet ready, but I consider it a P0 bug fix and I will finish up first thing next week.

### DIFF
--- a/src/message/messagesActions.js
+++ b/src/message/messagesActions.js
@@ -19,7 +19,7 @@ export const doNarrow = (narrow: Narrow, anchor: number = FIRST_UNREAD_ANCHOR) =
   }
 
   dispatch({ type: DO_NARROW, narrow });
-  dispatch(fetchMessagesInNarrow(narrow, anchor));
+  dispatch(fetchMessagesInNarrow(narrow, anchor, false));
   dispatch(navigateToChat(narrow));
 };
 
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -212,6 +212,7 @@ export type MessageFetchCompleteAction = {
   anchor: number,
   numBefore: number,
   numAfter: number,
+  replaceExisting: boolean,
 };
 
 export type MarkMessagesReadAction = {
--- a/src/chat/__tests__/flagsReducers-test.js
+++ b/src/chat/__tests__/flagsReducers-test.js
@@ -50,6 +50,32 @@ describe('flagsReducers', () => {
 
       expect(actualState).toBe(initialState);
     });
+
+    test('if `replaceExisting` is `true` replace old state completely', () => {
+      const initialState = deepFreeze({
+        read: {
+          1: true,
+          2: true,
+          3: true,
+        },
+      });
+
+      const expectedState = {
+        read: {
+          3: true,
+        },
+      };
+
+      const action = deepFreeze({
+        type: MESSAGE_FETCH_COMPLETE,
+        messages: [{ id: 3, flags: ['read'] }],
+        replaceExisting: true,
+      });
+
+      const actualState = flagsReducers(initialState, action);
+
+      expect(actualState.read).toEqual(expectedState.read);
+    });
   });
 
   test('flags are added or replace existing flags', () => {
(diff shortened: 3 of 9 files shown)


gnprice:
(Chat discussion here: https://chat.zulip.org/#narrow/stream/48-mobile/subject/no.20messages/near/661974 That's the diagnosis; reports are earlier in thread.)

src/message/messagesActions.js
```diff
@@ -21,24 +18,8 @@ export const doNarrow = (narrow: Narrow, anchor: number = FIRST_UNREAD_ANCHOR) =
...
-  const isCaughtUp = caughtUp.newer && caughtUp.older;
-
-  if (anchor === FIRST_UNREAD_ANCHOR && tooFewMessages && !isCaughtUp) {
-    dispatch(fetchMessagesAtFirstUnread(narrow));
-  } else if (anchor !== FIRST_UNREAD_ANCHOR) {
-    dispatch(fetchMessagesAroundAnchor(narrow, anchor));
-  }
-
+  dispatch({ type: DO_NARROW, narrow });
+  dispatch(fetchMessagesInNarrow(narrow));
```
gnprice: This needs to pass `anchor`; otherwise the `anchor` passed to `doNarrow` gets dropped, and this fetch always uses the default.

gnprice:
Hmm, and I guess I'll leave the second-to-last commit for now too, since it's infrastructure for the last one and it's not totally obvious to me that we'll want it in its current form.

gnprice:
OK, those 4/6 commits pushed, as d39148fd9 .

borisyankov:
The first three commits are ready to go in. 
I am reconsidering something on the last 5.

gnprice:
Thanks @borisyankov ! Just merged the first two commits, as 69f05fe91 and 611aabd24. Glad to have this bug fixed :smile:

gnprice:
I edited the title and priority to reflect that the part that fixed #3162 is now merged.

borisyankov:
I am still seeing #3162

gnprice:
Yeah, I've been increasingly seeing symptoms similar to #3162 recently. Not sure the behavior is quite the same as that one, though, and the cause is likely at least a bit different -- probably good to make a fresh issue for it.

Thanks for the ping on this PR. This thread is a bit confusing now, though -- it looks like back in November I merged 4 out of 6 commits, and then in early December I merged revised versions of the remaining two. So all the previous discussion here is about changes that are actually now merged.

(Or if it's not, it's confusing to sort that out. :stuck_out_tongue: )

Would you make a fresh PR with the new/remaining changes? I think that will make it easier to discuss them. (In retrospect, that would have been a good idea back in early December when I merged the remaining part of the original PR.)

chrisbobbe:
> This thread is a bit confusing now, though -- it looks like back in November I merged 4 out of 6 commits, and then in early December I merged revised versions of the remaining two. So all the previous discussion here is about changes that are actually now merged.

It may be time to close this, as I think was the intention back in March.

chrisbobbe:
Closing as stale; please let me know if this was done in error.

gnprice:
Yeah, I think closing this PR is the right thing.

The state of it was really what ought to be two PRs:
* "PR 3151(a)" is what was in the original version of the PR. That turned into 6 commits which I merged, as described above. So PR 3151(a) was merged.
* "PR 3151(b)" is some additional commits (also 6 of them) that got added to the PR later, but before the first 6 were merged. Those should have been pulled out into a new PR, but never were, and they never got properly discussed and merged. So PR 3151(b) is now quite stale and it's time to close it.

Looking briefly now at those commits, it looks like they were intended to address #3802 and perhaps related issues. I think the main interesting bit to take away from them for our current work on those issues is this point made in the first commit message:

    Do not reset `narrows`, `messages`, and `flags` on a dead queue
    
    [...]    
    This change brings the behavior consistent with what happens if
    the app was killed and restarted. It is important that we treat
    both cases the same way, as the often substitute each other -
    the app may be resumed with a dead queue and may try reconnecting,
    or it might have been killed producing a full restart.

PR #3154, opened 2018-11-17
Added flow type in actionTypes.js

VisibleMarkov:
Added flow type to `FetchMessageAction` and `EventAlertWordsAction`.

For `FetchMessageAction`:
`FetchMessageAction` in a function that is used to invoke Redux dispatch. This function is defined in both synchronous and asynchronous sense in `fetchActions.js`: [async](https://github.com/zulip/zulip-mobile/blob/08f1bfdd0b4dcb2aaf734ef590e6a78ca386618a/src/message/fetchActions.js#L75) and [sync](https://github.com/zulip/zulip-mobile/blob/08f1bfdd0b4dcb2aaf734ef590e6a78ca386618a/src/message/fetchActions.js#L88)

Tests Ran:
* [x] npm run test:flow
* [x] npm run test:lint
* [x] npm run test:full

Issue in focus: #2052

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -86,6 +86,7 @@ import type {
   PresenceState,
   Presence,
   RealmEmojiState,
+  SettingsState,
   CaughtUpState,
   MuteState,
   AlertWordsState,
@@ -515,8 +516,7 @@ export type FlagsAction =
 
 export type SettingsChangeAction = {
   type: typeof SETTINGS_CHANGE,
-  key: string,
-  value: boolean | string,
+  update: $Shape<SettingsState>,
 };
 
 export type DraftUpdateAction = {
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -516,7 +516,7 @@ export type FlagsAction =
 export type SettingsChangeAction = {
   type: typeof SETTINGS_CHANGE,
   key: string,
-  value: any,
+  value: boolean | string,
 };
 
 export type DraftUpdateAction = {
--- a/src/settings/LanguageScreen.js
+++ b/src/settings/LanguageScreen.js
@@ -19,7 +19,7 @@ class LanguageScreen extends PureComponent<Props> {
 
   handleLocaleChange = (value: string) => {
     const { dispatch } = this.props;
-    dispatch(settingsChange('locale', value));
+    dispatch(settingsChange({ locale: value }));
   };
 
   render() {
--- a/src/settings/NotificationsScreen.js
+++ b/src/settings/NotificationsScreen.js
@@ -27,7 +27,7 @@ class NotificationsScreen extends PureComponent<Props> {
       opp: 'offline_notification_change',
       value: !offlineNotification,
     });
-    dispatch(settingsChange('offlineNotification', !offlineNotification));
+    dispatch(settingsChange({ offlineNotification: !offlineNotification }));
   };
 
   handleOnlineNotificationChange = () => {
@@ -37,7 +37,7 @@ class NotificationsScreen extends PureComponent<Props> {
       opp: 'online_notification_change',
       value: !onlineNotification,
     });
-    dispatch(settingsChange('onlineNotification', !onlineNotification));
+    dispatch(settingsChange({ onlineNotification: !onlineNotification }));
   };
 
   handleStreamNotificationChange = () => {
@@ -47,7 +47,7 @@ class NotificationsScreen extends PureComponent<Props> {
       opp: 'stream_notification_change',
       value: !streamNotification,
     });
-    dispatch(settingsChange('streamNotification', !streamNotification));
+    dispatch(settingsChange({ streamNotification: !streamNotification }));
   };
 
   render() {
--- a/src/settings/SettingsCard.js
+++ b/src/settings/SettingsCard.js
@@ -51,7 +51,7 @@ class SettingsCard extends PureComponent<Props> {
 
   handleThemeChange = () => {
     const { dispatch, theme } = this.props;
-    dispatch(settingsChange('theme', theme === 'default' ? 'night' : 'default'));
+    dispatch(settingsChange({ theme: theme === 'default' ? 'night' : 'default' }));
   };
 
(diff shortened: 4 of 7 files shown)


VisibleMarkov:
Also added flow type to `EventAlertWordsAction`.
It is an object to the best of my knowledge and [this](https://github.com/zulip/zulip-mobile/blob/61d3027123dc5adf2a1a9e7957a0bfb60be6942e/src/events/eventToAction.js#L99) piece of code helped me get type `EventAlertWordsAction`.
I wanted to clarify one small thing: The object field `alertWords` is basically an array of `string` but [here](https://github.com/zulip/zulip-mobile/blob/0d41e29fd036e7f6541f4c30eb4b69c0673a940e/src/alertWords/alertWordsReducer.js#L16), in `alertWordsReducer.js`, it is being returned as `AlertWordsState` so I have kept the convention while defining the type and used `AlertWordsState` instead of `string[]`. I hope this was right. I'll change it back to `string[]` if that wasn't the right thing to do.
Would like your opinion on this @gnprice

VisibleMarkov:
Added flow type to `SettingChangeAction.value`.
`SettingChangeAction` is the return type of function `settingChange` in `settingsActions.js` and it's second arameter is always assigned to `SettingChangeAction.value`. This can be seen in it's definition [here](https://github.com/zulip/zulip-mobile/blob/2c1caa2a2aa5e574fca3acfb4d4fda0ca2adde5c/src/settings/settingsActions.js#L5).
Following this we get that `SettingChangeAction.value` get's assigned a `boolean` value [here](https://github.com/zulip/zulip-mobile/blob/0d2b8850339297f2cce0d4550b852b275d81ee16/src/settings/NotificationsScreen.js#L30) as well as a `string` [here](https://github.com/zulip/zulip-mobile/blob/b0730a0636862f09d195c33be09d47776e449080/src/settings/LanguageScreen.js#L22) and [here](https://github.com/zulip/zulip-mobile/blob/abe64f10b34dc0f9e76e7a1eb75917605e35b1d5/src/settings/SettingsCard.js#L54)

src/actionTypes.js
```diff
@@ -211,7 +215,7 @@ export type MessageFetchCompleteAction = {
   numAfter: number,
 };
 
-export type FetchMessagesAction = any;
+export type FetchMessagesAction = (dispatch: Dispatch, getState: GetState) => Promise<void> | void;
```
borisyankov: The definition you are providing is actually an action creator.
Have a look at the distinction between actions and action creators (the second creates the action 😄 )
https://redux.js.org/basics/actions

borisyankov: What seems to be confusing in this case is that the name is actually incorrect, and likely I did introduce that error :)
In `fetchActions` the return type of `fetchMessages` is not `FetchMessagesAction`
Compare that to `messageFetchComplete` where the returned type is indeed an action.

Maybe the right choice, for now, is to remove the `FetchMessagesAction` type.

VisibleMarkov: Shall I add this type as `ActionCreator` and replace the instances of `FetchMessagesAction` with `ActionCreator`?
Or the preferred way would be to remove the type entirely and define it inline?

borisyankov: Have a look at the `redux_v4.x.x.js` file which contains the definitions for Redux (coming from the FlowTyped project).

It already contains an `ActionCreator` type which you will want to reuse.

We haven't used it anywhere so you can't look in our source code, but Googling for examples will probably work fine.

VisibleMarkov: On it. Thanks a lot. Will implement as soon as i can.

VisibleMarkov: @borisyankov  Sorry for getting back so late but I needed some clarification here:
In `redux_v4.x.x.js`, `ActionCreator` is defined as follows: `ActionCreator<A, B> = (...args: Array<B>) => A;`. I had a hard time interpreting `...args: Array<B>` because we aren't passing any array here. 
I tried to just fit my above definition of ActionCreator into this and this is what I thought worked: `FetchMessagesAction = ActionCreator<void | Promise<void>, GetState | Dispatch>;` and flow gave no errors but then i went ahead and made a small change `FetchMessagesAction = ActionCreator<void | Promise<void>, GetState>;` removing `Dispatch` and flow still didn't give me an error which led me to conclude that this implementation of mine might be wrong. Seeking some guidance on this.

gnprice:
Thanks @VisibleMarkov ! The `EventAlertWordsAction` commit looks great; merged, as 7a80a9cb9.

For FetchMessagesAction, I agree with @borisyankov 's suggestion above that it's probably best just to delete it. I've pushed a commit to do that: de70ffd45 .

For SettingsChangeAction, one of the couple of places we use that type is here:
```
export const settingsChange = (key: string, value: any): SettingsChangeAction => ({
  type: SETTINGS_CHANGE,
  key,
  value,
});
```
which has an `any` in it. Would you fix that too, and anywhere else that type is involved? That'd make it easier to confirm at a glance that everything that uses that type handles it consistently with the new, clearer definition.

src/actionTypes.js
VisibleMarkov: Fixed in de70ffd thanks to @gnprice

VisibleMarkov:
Reverted fdb946b, the addition of flow type to `FetchMessageAction` through c35945a - removed merge conflict
Made the necessary changes in 0846a17 for `SettingChangeAction.value`
I think this commit isn't responsible for failing the CI test. If a squash is required,please let me know.

VisibleMarkov:
Cleaned up the commits. I have still kept the `EventAlertWordAction` and the `SettingChangeAction` separate though. 
Is this fine?
P.S. Thanks for the link to the docs, helped a ton!

VisibleMarkov:
Thanks @gnprice, Being new to git, I'm yet to discover the wonderful thing it holds. Apologies for the inconvenience. I'll make sure the PRs in the future will be better.
Thank you for your guidance.

gnprice:
Not a problem; thanks, and you're welcome!

Issue #3073, opened 2018-10-25
Improve CompatibilityScreen text.


jackrzhang:
Two things that @gnprice and I discussed in person:
- Refer to 'Play Store' on Android and 'App Store' on iOS.
- Include links to the latest versions.

(@zulipbot claim)

borisyankov:
I'd say an important thing to test first is if this component even behaves OK still :)
I don't think we've tested it recently (or ever since its creation)

borisyankov:
After that, we definitely want to update the URLs 👍

gnprice:
Bumped the priority because completing #2634 will require us to activate this mechanism for old Android clients -- and before we do that we want to complete this change, and get a release with it out to the Play Store, so that when someday we activate this mechanism for a second time the experience is nicer. :)

gnprice:
Now complete, with #3158 . Thanks @jackrzhang !

PR #3158, opened 2018-11-20
[WIP] Implement and smoothen UX around compatibility checks.

jackrzhang:
This PR corresponds to #3073.

This is a work in progress; I'm opening this now for reference - no need for review yet!


gnprice:
Thanks @jackrzhang ! I've just merged the first of these commits, because that one is ready -- see 6f77a4a64 and its parent. (I made a tweak on top.)

jackrzhang:
@gnprice With the last 4 commits, I've pushed updates to the `CompatibilityScreen` UI, which renders if a global check to https://zulipchat.com/compatibility fails.


<p align="center">
android-compatibility-screen-global-check
<p/>

borisyankov:
That link does not look clickable.
We need the 'App Store' and 'Play Store' buttons:




These are just examples, not sure where is the the official source of these resources.

jackrzhang:
The color and other visual properties of the linked text can be changed. I don't think the buttons are *needed*, though they're recognizable.

@borisyankov Where do you get those examples from?

jackrzhang:
I've found the guidelines for the badges:
- https://developer.apple.com/app-store/marketing/guidelines/#section-products
- https://play.google.com/intl/en_us/badges/

(I'll go ahead and add them?)

borisyankov:
I was a bit concerned about file sizes but the Google Play Store badge is just 13kb so I think that is reasonable.

jackrzhang:
The (current) first 7 commits of this branch update the compatibility screen UI by adding touchable badges, as suggested by @borisyankov above, and phrasing the text a bit more concisely.

I ran `yarn jest` locally at  549d573; the tests passed.

I tested the screen on iOS 12.1 by running the build on my iPhone 7.

Here's a screen recording from an AVD with Google Play:

<p align="center">
android-compatibility-screen-global-check-with-badge
<p/>

@gnprice

gnprice:
Thanks @jackrzhang ! This looks good.

Would you include also the original SVG download of the App Store badge? Good for reproducibility, as that is more like the real "source" version of the image.

gnprice:
Merged, with a few edits. Thanks @jackrzhang !

PR #3159, opened 2018-11-21
SwitchAccountButton: Remove unused prop `accounts`.

jackrzhang:
This is a silent type error that seems to have been introduced via https://github.com/zulip/zulip-mobile/commit/b37352ade06779ad77f334c091300c2370cd9ce2#diff-73a6ef41677e3bf406fce93366e99735.

### DIFF
--- a/src/account-info/SwitchAccountButton.js
+++ b/src/account-info/SwitchAccountButton.js
@@ -6,7 +6,7 @@ import { StyleSheet } from 'react-native';
 
 import type { Auth, Dispatch, GlobalState } from '../types';
 import { ZulipButton } from '../common';
-import { getAuth, getAccounts, getPushToken } from '../selectors';
+import { getAuth, getPushToken } from '../selectors';
 import { unregisterPush } from '../api';
 import { logErrorRemotely } from '../utils/logging';
 import { deleteTokenPush, navigateToAccountPicker } from '../actions';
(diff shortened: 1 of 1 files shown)


borisyankov:
Nice. Is there a way to make these errors non-silent?

jackrzhang:
Unfortunately not, I think :/ 

There is a `react/no-unused-prop-types` eslint rule, but that seems to be specific to the syntax `React.propTypes` and not flow.

jackrzhang:
According to https://github.com/yannickcr/eslint-plugin-react/pull/1412, this should be supported. Maybe there's something that's not being done with the eslint configuration?

gnprice:
Thanks @jackrzhang ! Merged.

I share @borisyankov 's reaction :slightly_smiling_face: -- it would be great to have this automatically checked. I played with it a bit just now, but didn't find a way; in particular, making that `Props` an exact type was no help. I suspect unfortunately a Flow-based approach won't be helpful until `connect` has more of a real-type signature:
```
  declare export function connect<
    Com: ComponentType<*>,
    S: Object,
    SP: Object,
    RSP: Object,
    CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>,
    ST: {[_: $Keys<Com>]: any}
    >(
    mapStateToProps: MapStateToProps<S, SP, RSP>,
    mapDispatchToProps?: null
  ): (component: Com) => ComponentType<CP & SP> & $Shape<ST>;
```
Look at all those `Object`s and `any`s. :cry: 

Alternatively an ESLint-based fix would be great, if someone can find a rule that works for this.

PR #3160, opened 2018-11-21
Rename `auth` to `account`.

jackrzhang:
With the exception of code corresponding to the `AuthenticationMethods` type and configuration files, all references to `auth` or the `Auth` type are also of the `Account` type:

https://github.com/zulip/zulip-mobile/blob/8cb2dc8e363e1c84b6fa7e5811345722f451f05f/src/api/apiTypes.js#L4-L8

https://github.com/zulip/zulip-mobile/blob/8cb2dc8e363e1c84b6fa7e5811345722f451f05f/src/types.js#L64

Refer to `account`, or `Account`, which I think is more accurate than `auth` or `Auth`.

(This PR touches a lot of files, so there's a high likelihood of merge conflicts.)


borisyankov:
I strongly disagree with this analysis and PR.

This line is important (if not documented):
https://github.com/zulip/zulip-mobile/blob/master/src/types.js#L64
... and hints at the purpose of the two types.

`Auth` is different than `Account` but is incidentally and temporarily the same.

There are plans to enhance `Account` that just weren't implemented (yet?), but you can imagine stuff like `lastLogin`, `realmIcon` or maybe the fact that `apiKey` is optional (a maybe type).

`Auth` on the other hand will not implement any of these new properties and maybe should keep `apiKey` mandatory.

gnprice:
I agree directionally with @borisyankov 's reply, and this in particular:

> Auth is different than Account but is incidentally and temporarily the same.

The way I'd primarily frame the difference: `Auth` contains the user's API key, which means the power to go and do things as them on the server. In the existing codebase, we often pass an `Auth` around to parts of the code that have no business doing anything on the server; what they do legitimately need is the user's identity (by email or ID), or the base URL of the realm. Passing around an API key all the time like that makes me nervous, because even if the probability of a bug where it gets someplace it really shouldn't is fairly low (and AFAIK we have never had a bug like that), the impact would be high.

So the distinction between an Auth (which is the power to go and act as them -- a [capability](https://en.wikipedia.org/wiki/Capability-based_security), in the jargon of a pretty handy security concept) and an Account (which just identifies them) is an important one which we ought to make.

Until we do, as long as the values contain an API key, I'm happy with them being called "auth" even where what the code really needed was merely an "account". That at least serves as a reminder that this thing we're passing around is a thing with power.

gnprice:
> So the distinction between an Auth (which is the power to go and act as them -- a capability, in the jargon of a pretty handy security concept) and an Account (which just identifies them) is an important one which we ought to make.

OK, I went and made a separate type for this! And a corresponding selector. See 28b1177d3.

I demoed using it in one place, but there are many others where it would be good to use.

Also added a comment :smile: as @jackrzhang requested above, in the parent commit f5a2603a4.

PR #3163, opened 2018-11-21
Remove usage of connectPreserveOnBackOption.

jainkuniya:
Removed from `MainNavBar`, `ModalSearchNavBar`, where it will always be true.
Currently WIP, need to handle cases where it will be false in `ModalNavBar`. (Also a rebase)

Edit: Completed.

### DIFF
--- a/src/nav/ModalNavBar.js
+++ b/src/nav/ModalNavBar.js
@@ -4,13 +4,11 @@ import { connect } from 'react-redux';
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
 
-import type { Dispatch, Context, GlobalState, LocalizableText } from '../types';
+import type { Dispatch, Context, LocalizableText } from '../types';
 import styles, { NAVBAR_SIZE } from '../styles';
 import Label from '../common/Label';
-import { getCanGoBack } from '../selectors';
 import NavButton from './NavButton';
 import { navigateBack } from '../actions';
-import { connectPreserveOnBackOption } from '../utils/redux';
 
 type Props = {|
   dispatch: Dispatch,
@@ -51,11 +49,4 @@ class ModalNavBar extends PureComponent<Props> {
   }
 }
 
-export default connect(
-  (state: GlobalState) => ({
-    canGoBack: getCanGoBack(state),
-  }),
-  null,
-  null,
-  connectPreserveOnBackOption,
-)(ModalNavBar);
+export default connect()(ModalNavBar);
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -43,7 +43,12 @@ class AccountPickScreen extends PureComponent<Props> {
     const { accounts, dispatch, auth } = this.props;
 
     return (
-      <Screen title="Pick account" centerContent padding>
+      <Screen
+        title="Pick account"
+        centerContent
+        padding
+        canGoBack={!!(auth.apiKey && auth.apiKey.length > 0)}
+      >
         <Centerer>
           {accounts.length === 0 && <Logo />}
           <AccountList
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -39,16 +39,23 @@ class AccountPickScreen extends PureComponent<Props> {
     this.props.dispatch(removeAccount(index));
   };
 
+  // We can get here three ways:
+  //  * the "switch accounts" button
+  //  * the "log out" button
+  //  * as the initial screen, if we have a known account but no API key.
+  //
+  // The "log out" button is a bit exceptional because it's the user
+  // taking a navigational action... but the screen they just left
+  // required the login they've just discarded, so they can't go back.
+  //
+  // So, show a "navigate back" UI in the first case, but not the other two.
+  canGoBack = !!this.props.auth.apiKey;
+
   render() {
     const { accounts, dispatch, auth } = this.props;
 
     return (
-      <Screen
-        title="Pick account"
-        centerContent
-        padding
-        canGoBack={!!(auth.apiKey && auth.apiKey.length > 0)}
-      >
+      <Screen title="Pick account" centerContent padding canGoBack={this.canGoBack}>
         <Centerer>
           {accounts.length === 0 && <Logo />}
           <AccountList
--- a/src/common/Screen.js
+++ b/src/common/Screen.js
@@ -15,10 +15,11 @@ import type {
 import KeyboardAvoider from './KeyboardAvoider';
 import OfflineNotice from './OfflineNotice';
 import ZulipStatusBar from './ZulipStatusBar';
-import { getSession } from '../selectors';
+import { getCanGoBack, getSession } from '../selectors';
 import ModalNavBar from '../nav/ModalNavBar';
 import ModalSearchNavBar from '../nav/ModalSearchNavBar';
 import styles from '../styles';
+import { connectPreserveOnBackOption } from '../utils/redux';
 
 const componentStyles = StyleSheet.create({
   wrapper: {
@@ -37,6 +38,7 @@ const componentStyles = StyleSheet.create({
 
 type Props = {|
   autoFocus: boolean,
+  canGoBack: boolean,
   centerContent: boolean,
   children: ChildrenArray<*>,
   safeAreaInsets: Dimensions,
@@ -89,6 +91,7 @@ class Screen extends PureComponent<Props> {
   render() {
     const {
       autoFocus,
+      canGoBack,
       centerContent,
       children,
       keyboardShouldPersistTaps,
@@ -109,7 +112,7 @@ class Screen extends PureComponent<Props> {
         {search ? (
           <ModalSearchNavBar autoFocus={autoFocus} searchBarOnChange={searchBarOnChange} />
         ) : (
-          <ModalNavBar title={title} />
+          <ModalNavBar canGoBack={canGoBack} title={title} />
         )}
         <OfflineNotice />
         <KeyboardAvoider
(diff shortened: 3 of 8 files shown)


src/nav/ModalNavBar.js
```diff
@@ -76,11 +78,4 @@ class ModalNavBar extends PureComponent<Props> {
...
 
-export default connect(
-  (state: GlobalState) => ({
-    canGoBack: getCanGoBack(state),
-  }),
-  null,
-  null,
-  connectPreserveOnBackOption,
-)(ModalNavBar);
+export default connect((state: GlobalState) => ({}), null, null)(ModalNavBar);
```
borisyankov: `(state: GlobalState) => ({}), null, null)` is that part even needed?
If there is no `dispatch` even the `connect` is not needed.

gnprice:
Thanks @jainkuniya !

> `<ModalNavBar dispatch={dispatch} title={title} />`

I'm not a fan of passing `dispatch` as an explicit prop like this. In general I'd rather either:
* get `dispatch` from a `connect` call -- even if that's the only thing `connect` does; or
* pass specific callbacks to do the specific things the child component needs to do.

In this case, I think a callback prop like `onBack` would be a good choice. Then the parent can have something like `onBack = () => { this.props.dispatch(navigateBack()); }`.

---

Zooming out: how are we going to have this code decide whether the back/Up button should be shown?

Maybe the first question: are there any screens at all where we use `Screen` and *don't* want to show it, i.e. where `canGoBack` should be false?

I am not sure there are. I think the only screen where that should be false may be `MainScreenWithTabs`... which doesn't use `Screen`, so this code doesn't get involved.

If that's right, we can effectively just hard-code `canGoBack = true` throughout these components. (Which means we remove `canGoBack` completely, replacing it with `true` and then simplifying the code accordingly.)

gnprice:
Later, if we need to revive some logic resembling `getCanGoBack`: starting with react-navigation v2 there is a direct way to do that [from the component's props](https://reactnavigation.org/docs/en/2.x/navigation-prop.html#dangerouslygetparent-get-parent-navigator), specifically `props.navigation`. Which would be fine; the important thing is it needs to come from the props.

gnprice:
Thanks @jainkuniya !

I agree about those two screens.

I like the first commit; I plan to merge it in a minute.

Starting at the second commit:
* "It is hard to image a case [where it's false]" is not very reassuring. :smile: Can you give an explanation why it definitely will always be true?
* Why keep the `connectPreserveOnBackOption`, and why have a `canGoBack` defaultProp?

Third commit:
* It's not totally clear to me we weren't using this; certainly we seem to look at `state` in `render`, and use it to decide what to show. Also you're refactoring the API to change the type of a prop. Maybe this would be easier to explain as a couple of separate commits?

Fourth commit (ModalNavBar):
* What @borisyankov said above. :slightly_smiling_face: 

Fifth commit ("flow, modalNavBar"):
* Same; why the `connect` at all?
* Also, it sounds like Flow was broken at the previous commit? Remember, every commit should leave things as non-broken.

Last few commits:
* It looks like these are adding back functionality that was broken in previous commits. Every commit should leave things as non-broken. Maybe add this optional prop earlier on.

gnprice:
And merged that first commit, as c31a1c50a.

jainkuniya:
> Can you give an explanation why it definitely will always be true?

It is hard to imagine a case when `canGoBack` will be false for this
component, so the selector is almost redundant for this. This modal is
used in search screens, where we need to show search input in the
navBar. As per our current app flow, there is always a possibility of
going back from a search screen. So we always need to show back icon
in this modal.

>It's not totally clear to me we weren't using this;

Before we were using different type of navBar, there was a search icon
in the right of navBar and on clicking this icon we were showing
search input in the navBar else screen name was shown. But now we are
not using this style. Now in all screens, by default searchInput is
shown. So remove this code of toggling search state.

>it looks like these are adding back functionality...

Now on every commit is is working, i.e not broken :)

src/nav/ModalNavBar.js
borisyankov: What I meant was that you don't need to pass the three empty parameters.
```js
export default connect((state: GlobalState) => ({}), null, null)(ModalNavBar);
```
can become
```js
export default connect()(ModalNavBar);
```

jainkuniya: Cool! got it 👍

src/nav/ModalNavBar.js
```diff
@@ -4,17 +4,15 @@ import { connect } from 'react-redux';
...
 import Label from '../common/Label';
-import { getCanGoBack } from '../selectors';
 import NavButton from './NavButton';
 import { navigateBack } from '../actions';
-import { connectPreserveOnBackOption } from '../utils/redux';
 
 type Props = {|
   dispatch: Dispatch,
-  canGoBack: boolean,
+  canGoBack?: boolean,
```
borisyankov: This should not be optional because you always pass a value thanks to `defaultProps`

borisyankov:
@gnprice I did a review. Besides the single comment I made, I think the PR is good to merge.

src/nav/ModalNavBar.js
jainkuniya: Done 👍

gnprice:
Thanks @jainkuniya !

> refactor: Remove unused code from `ModalSearchNavbar`.

The new explanation is helpful, thanks.

This commit is still doing several different things, and some of them don't match the commit message. In particular, it adds an extra condition `&& searchBarOnChange` in `Screen.js`. If that prop is missing, do you really want to make a `ModalNavBar` instead? Why?

The explanation also doesn't quite say why this code really is dead. "By default" is not very reassuring. :-) I think the real point is that at the only call site for this component, we don't pass these three props -- so in fact they are always `undefined`, and that leads to all these simplifications.

> modalSeachNavbar: Remove dependency of `getCanGoBack`.

Again "hard to imagine" and "almost redundant" are not very reassuring. :wink: The thing I really want to know is that the value is *always* true, because we only ever use this on screens where it's always true.

> modalNavBar: Remove the dependency of `getCanGoBack`.

Again, why does `ModalNavBar` have a default value for `canGoBack`? Or for `title`?

This commit is also doing two very different things -- these would be better as separate commits:
* It moves the determination of `canGoBack` from `ModalNavBar` up to its caller, `Screen`. This is a straightforward plumbing change -- it's a refactor that doesn't require any careful thinking about the app's UI to understand and verify.
* It moves the determination of `canGoBack` from `Screen` up to its callers. This requires thinking globally about the entire app's interface, and trying to be sure not to miss some other screen that could possibly need to not have a back icon.

The first of these two could just move the `getCanGoBack` call to `Screen`. Then the second one would remove that call, in favor of having the default and passing the exceptional values down from those two screens.

--

Looking back at these, I think two of my comments can be addressed together:
* For both kinds of `*NavBar` component, make `canGoBack` a prop passed by the caller; make `Screen` use `getCanGoBack`.
* Then, replace that last `getCanGoBack` with passing `canGoBack` from the callers of `Screen` (with a default). I think it'll happen to be the case that the only places we pass `false` there will be places we don't pass `search` as true... but that's kind of by accident.

I also don't feel 100% confident yet that these two are the only two screens that should be passing `canGoBack={false}`. Can you explain why it's really only those two screens?

borisyankov:
> I also don't feel 100% confident yet that these two are the only two screens that should be passing canGoBack={false}. Can you explain why it's really only those two screens?

There are two things here.

First, optimally we don't need to reason about what Screen can go back, as these components are in the root of the Stack Navigator and will have a `navigation` passed to them. We can use similar logic to what we have in Redux, or something ready-made and will look something like `this.props.navigation.canGoBack()` (in this case, `canGoBack` is pseudo code, needs some digging in the documentation).

But second, it isn't too hard to determine what screens can be navigated back from.
We have two main states:
* logged in to a realm - the main screen is `MainScreenWithTabs` and it does not have a back button
* login flow - this can reliably be analyzed as it includes only a few screens. I think @jainkuniya analysis here is conclusive and only these screens really can not have a 'back' button

gnprice:
> First, optimally we don't need to reason about what Screen can go back, as these components are in the root of the Stack Navigator and will have a `navigation` passed to them. We can use similar logic to what we have in Redux, or something ready-made and will look something like `this.props.navigation.canGoBack()`

Hmm, I like this idea! Very much agree that it'd be preferable not to have to reason about which screens can go back.


> * logged in to a realm - the main screen is `MainScreenWithTabs` and it does not have a back button
> * login flow - this can reliably be analyzed as it includes only a few screens. I think @jainkuniya analysis here is conclusive and only these screens really can not have a 'back' button

Well, I guess I don't yet see an analysis of the login flow, only a conclusion :slightly_smiling_face: . What are all the different screens in the login flow, and why is that a complete list?

Or, what are all the different ways we can land on a screen without navigating in a way we should be able to go back from, and why is that a complete list?

jainkuniya:
> What are all the different screens in the login flow, and why is that a complete list?

Welcome Screen, RealmScreen, AuthScreen, PasswordAuthAcreen, DevAuthScreen.

There are 3 possibility:
* User freshly opens app, after install. Then welcome screen will be at the root of the navigation till user is logged in. 
* After logging/user opens app (already login case): `MainScreenWithTabs` will be at root (Main). 
* User logs out/opens app which logged out from all out: Then `AccountPickScreen` will be at the root of the navigation.

So we only need to handle 1 & 3 case, where user can not go back and don't want to show back icon in nav bar.

cc @gnprice

gnprice:
And merged! :confetti_ball: 

I made some edits to the branch -- primarily
* splitting up some commits
* adding more reasoning on a couple of commit messages
* adding a couple of extra commits to clarify some aspects.

Do take a look at the merged version; it's 6c7fffc76..cd7b25757.

PR #3164, opened 2018-11-24
Stricter Flow

borisyankov:
Fixes issues that were found by experimentally enabling `@flow strict` on all files.
Unfortunately, we can not do that as any imported file needs to also be strictly typed.

While the non-strict Flow does not recognize these as issues it is good to do these fixes.

### DIFF
--- a/src/account/AccountList.js
+++ b/src/account/AccountList.js
@@ -9,7 +9,7 @@ type Props = {
   auth: Auth,
   accounts: Account[],
   onAccountSelect: number => void,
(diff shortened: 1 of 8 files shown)


gnprice:

gnprice:
This also inspired me to make some further investigations of making Flow stricter, which I just pushed; see bcca910a7..6efa7980c . In particular, there seems to be a new feature `strict-local`, which looks very handy for making progress on strictness!

Issue #2722, opened 2018-06-25
Group PM shows up on PMs screen with blank name, just commas


gnprice:
On a user's screen when they were showing me something else in the app, I saw the PMs screen in the main nav. Out of several conversations, one of them was labeled just ", ," -- comma, space, comma -- where the names of the other people should go. It looked like what you might get with a group PM thread with 3 other people, if their names were all the empty string.

I've seen this symptom myself in the past a couple of times. It'd be good to nail down the cause and fix it.

This was on an iPhone in 13.4.88, the latest. The org was zephyr.zulipchat.com, which is somewhat special on the server side -- but I don't see an obvious way that would cause this issue (and when I've seen this myself, it wasn't on zephyr), and in any event this is a buggy thing to show.

gnprice:
Here's the plan I'd suggest for how to do this:

First, some cleanups to help make this part of the code easier to understand and work with. This will make the later steps easier, and will also help us out every time we come back to this code in the future.
* Put an appropriate type annotation on the `conversations` prop on the various components here -- `Object[]` isn't very helpful.
* ConversationGroup isn't really a group, and it's not about just any conversation. Let's give it a better name, and at least a one-line JSDoc comment saying what it really is. A good one would be `GroupPmConversationItem` and a comment saying
```
/** A list item describing one group PM conversation. */
```
* Similarly ConversationList (`PmConversationList`) and ConversationsCard (`PmConversationsCard`).

Then, to actually fix the issue:
* I like having the users' names get looked up from the `users` state at render time -- that keeps the `users` state as the single source of truth about who a user is.
* The issue, then, is just that that state is missing some users that we have messages with.
* I don't have a full understanding of the flows that cause the `users` state to update, and why we don't have this state or what logic we might have that should cause us to have it. Clearly we don't consistently have it now. If we did fix things so that we did, there would probably still sometimes be bugs, causing gaps.
* On the other hand, as #1879 exploited, we have information about these users right there in `display_recipient` on the message. We can use that to fill these gaps.
* The key thing I'd like to do differently from the approach in #1879 is: let's fill the gaps *in the `users` state itself*, our single source of truth.
* I take inspiration for this from the webapp, which follows this approach. See `extract_people_from_message` in `people.js`.
* Like the webapp (see the `report_late_add` call), we can also use this in the future to log when this happens, so it helps us resolve the causes of these gaps too.

borisyankov:
[...]

PR #3165, opened 2018-11-26
Fix, add a test, improve `getRecipientsInGroupNarrow`

borisyankov:
While adding Flow types, I found a bug in `getRecipientsInGroupNarrow`.
If a user is not found by the `allUsers.find` call, it is replaced with
an empty array `[]`. This was likely intended to be an empty object.

Funnily, doing a `user.email` returns an `undefined` regardless if
`user` is an object or an array. Neither will satisfy a strictly
typing object as a `User` type. This is a good argument in favor of
the `NULL_USER` constant. For that reason, I do not fully agree with
deprecating it.

Instead of adding a dummy/null object we can filter the user out.
So do that, and add a test to confirm this works.

### DIFF
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -4,6 +4,7 @@ import {
   getFirstMessageId,
   getLastMessageId,
   getLastTopicForNarrow,
+  getRecipientsInGroupNarrow,
   getMessagesForNarrow,
   getStreamInNarrow,
   isNarrowValid,
@@ -260,6 +261,27 @@ describe('getLastTopicForNarrow', () => {
     const actualLastTopic = getLastTopicForNarrow(narrow)(state);
 
     expect(actualLastTopic).toEqual('Some subject');
+  });
+});
+
+describe('getRecipientsInGroupNarrow', () => {
+  test('returns recipients in group narrow skipping non-existing users', () => {
+    const state = deepFreeze({
+      narrows: {},
+      users: [
+        {
+          email: 'john@example.com',
+        },
+      ],
+      realm: {
+        nonActiveUsers: [],
+      },
+    });
+    const narrow = groupNarrow(['john@example.com', 'nonexisting@example.com']);
+
+    const actualRecipients = getRecipientsInGroupNarrow(narrow)(state);
+
+    expect(actualRecipients).toEqual([{ email: 'john@example.com' }]);
   });
 });
 
--- a/src/chat/narrowsSelectors.js
+++ b/src/chat/narrowsSelectors.js
@@ -100,7 +100,9 @@ export const getLastTopicForNarrow = (narrow: Narrow): Selector<string> =>
 
 export const getRecipientsInGroupNarrow = (narrow: Narrow) =>
   createSelector(getAllUsers, allUsers =>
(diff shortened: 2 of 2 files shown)


gnprice:
> Funnily, doing a `user.email` returns an `undefined` regardless if
> `user` is an object or an array. Neither will satisfy a strictly
> typing object as a `User` type. This is a good argument in favor of
> the `NULL_USER` constant. For that reason, I do not fully agree with
> deprecating it.

Hmm. I definitely agree that this code is buggy!

```
    emailsOfGroupNarrow(narrow).map(r => allUsers.find(x => x.email === r) || [])
```

I think you're saying you would have liked it to be
```
    emailsOfGroupNarrow(narrow)
      .map(r => allUsers.find(x => x.email === r) || NULL_USER)
```
Is that right?

And instead, because I've asked to deprecate `NULL_USER`, the fix you've written is
```
    emailsOfGroupNarrow(narrow)
      .map(r => allUsers.find(x => x.email === r))
      .filter(user => user)
```

It comes down to how we want the code to behave in this error case. I don't really love either of these behaviors, actually; both of them seem likely to lead to buggy behavior. Fundamentally they both return to the caller some data that's partly corrupted.
* In the `NULL_USER` version, there's a bunch of invalid values for `user_id`, `email`, `avatar_url`. As a result:
  * When the caller says `key={user.email}` in some list (as both existing callers do: `TitleGroup` and, after passing from `InfoNavButtonGroup` through navigation, `GroupDetailsScreen`), they're liable to get duplicate keys if multiple users are missing. Similarly if they used `user.user_id`.
  * When `user_id` or `email` is used to index into some other data structure (e.g. `presence` in `TitleGroup`), the result will be `undefined` or some other corrupt data.
  * Whatever the caller goes on to render in the UI based on the corrupt data is likely to look highly buggy -- like #2722 .
  * In general: if the caller tries to use the data, it's likely not to get a result that's actually an appropriate way to handle missing user data.
* In the `filter` version, the returned recipient list is shorter than the real one. As a result:
  * It might be empty, which is likely to produce buggy results.
  * It might be the same as the recipient list of a different conversation which appears on the same screen.
  * If the list of recipients is used to identify the conversation, it will refer to a different conversation.

I think there are basically two ways to make a helper like this handle this situation that make it possible to use it non-buggily:
* Throw an error. This is appropriate when we're sure it would represent a bug, and not a data/network issue. I think that isn't quite true here, but it might be once we make the navigation on launching from a notification block on learning about the conversation.
* Return a sentinel value, and require the caller to explicitly handle it. For this `undefined` would work, though perhaps `null` is a bit cleaner. So this code would be
  ```
    emailsOfGroupNarrow(narrow)
      .map(r => allUsers.find(x => x.email === r) || null)
  ```
  and the caller would get an `Array<User | null>` instead of `Array<User>`, and would have to handle that appropriately -- perhaps skipping nulls, perhaps rendering placeholders in the UI, perhaps raising an error, but in any event making the choice explicitly.

gnprice:
> I think there are basically two ways to make a helper like this handle this situation that make it possible to use it non-buggily:
> 
> * Throw an error.

I ended up taking this path, in ca0df5122c98af5e96b56d64a6627318c884d5bc a little after this PR.

Also fixed the linear search with de62b0f96c665dd78de3dac19a2a35bae0309294.

PR #3166, opened 2018-11-26
Add Flow for all selectors

borisyankov:
Selectors are one of the most important parts of the app that are still not typed. So do that for:
* accountsSelectors
* caughtUpSelectors
* fetchingSelectors
* narrowsSelectors
* messageSelectors
* pmConversationsSelectors
* subscriptionSelectors
* topicSelectors
* typingSelectors
* unreadSelectors
* userSelectors

`navSelectors` are intentionally nov covered as we are trying to get rid of them.

### DIFF
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -4,6 +4,7 @@ import {
   getFirstMessageId,
   getLastMessageId,
   getLastTopicForNarrow,
+  getRecipientsInGroupNarrow,
   getMessagesForNarrow,
   getStreamInNarrow,
   isNarrowValid,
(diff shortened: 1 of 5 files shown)


borisyankov:
Updated the commit message and added a commit for `navSelectors`

gnprice:
Thanks @borisyankov ! Merging the commits here that are easy to merge.

Exceptions:

* I left out the `unreadSelectors` commit because it gave a warning.
* I dropped the `getAllUsersByEmail` hunk because it didn't look right -- why are the values `User` and not `User | RealmBot`? I think you need to type the initial value for `reduce`, to avoid `reduce` laundering a type error.
* I left out the `navSelectors` commit because it added a vague type -- saying `getCurrentRouteParams: Selector<Object>` isn't really better than not saying a type -- and when I tried dropping that hunk, I got a type error on a different selector there.
* I left out the `narrowsSelectors` commits, because they seem to depend on / include #3165; let's follow up there.

gnprice:
#2971 is now complete, and I think everything in this branch has been done.

PR #3167, opened 2018-11-26
avatar: Fix rounded corners not used for group chat icon

borisyankov:
The `TextAvatar` component is used for the avatar of group chats.
It did not show the same rounded corners as the single PMs.

This is supported and the issue was that we are not passing a `shape`
prop thus accidentally defaulting the value to `square`.

Set the default value for `shape` to `rounded` for both avatar
components to fix the issue and to better reflect our intent.


gnprice:

PR #3174, opened 2018-11-28
Reuse the destination narrow logic when uploading images

### LINKED ISSUE #3130: Image upload in stream narrow goes to wrong topic
Steps:
* Narrow to a stream, and open the compose area.
* Edit the topic.
* Upload an image -- I just now tested with an existing image, but I believe using the camera has the same result.
* The message doesn't go to the topic you asked for! Instead it goes to... the topic of the latest message visible, maybe.

This is a common experience when intending to start a new conversation.

borisyankov:
Fixes #3130

Extract the code in `ComposeBox => handleSend()` that determines
the destination narrow to `getDestinationNarrow()`. Use this function
to pass the narrow to `ComposeMenu` which then uses that value when
uploading an image.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -212,15 +212,17 @@ class ComposeBox extends PureComponent<Props, State> {
     this.setState({ isMenuExpanded: false });
   };
 
+  getDestinationNarrow = (): Narrow => {
+    const { narrow } = this.props;
+    const { topic } = this.state;
+    return isStreamNarrow(narrow) ? topicNarrow(narrow[0].operand, topic || '(no topic)') : narrow;
+  };
+
   handleSend = () => {
-    const { dispatch, narrow } = this.props;
-    const { topic, message } = this.state;
+    const { dispatch } = this.props;
+    const { message } = this.state;
 
-    const destinationNarrow = isStreamNarrow(narrow)
-      ? topicNarrow(narrow[0].operand, topic || '(no topic)')
-      : narrow;
-
-    dispatch(addToOutbox(destinationNarrow, message));
+    dispatch(addToOutbox(this.getDestinationNarrow(), message));
 
     this.setMessageInputValue('');
   };
@@ -302,7 +304,7 @@ class ComposeBox extends PureComponent<Props, State> {
         <View style={styles.composeBox} onLayout={this.handleLayoutChange}>
           <View style={styles.alignBottom}>
             <ComposeMenu
-              narrow={narrow}
+              destinationNarrow={this.getDestinationNarrow()}
               expanded={isMenuExpanded}
               onExpandContract={this.handleComposeMenuToggle}
             />
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -15,7 +15,7 @@ import { getNarrowToSendTo } from '../selectors';
 type Props = {
   dispatch: Dispatch,
   expanded: boolean,
-  narrow: Narrow,
+  destinationNarrow: Narrow,
   onExpandContract: () => void,
 };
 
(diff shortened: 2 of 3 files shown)


src/compose/ComposeBox.js
```diff
@@ -302,7 +304,7 @@ class ComposeBox extends PureComponent<Props, State> {
         <View style={styles.composeBox} onLayout={this.handleLayoutChange}>
           <View style={styles.alignBottom}>
             <ComposeMenu
-              narrow={narrow}
+              narrow={this.getDestinationNarrow()}
```
gnprice: Hmm, this feels kind of surprising. Usually we use a `narrow` prop to mean "the narrow the user is looking at". What does `ComposeMenu` do with this prop -- will it do something unexpected because it assumes this is the narrow in view?

Looking at that component, the answer seems to be:
* Indeed it does assume that
* but what it does with that is to go apply some other logic to try to pick a topic to send to!
* so what this change ends up doing is bypassing that logic, because it looks for `isStreamNarrow` and this will no longer ever pass a stream narrow.

Let's avoid confusion here. :-)
* Cut the now-dead `getNarrowToSendTo` over in that code.
* Rename this prop to, say, `narrowToSendTo` (maybe `sendNarrow` if you want to be terse), so it's clear it doesn't have the same meaning as we conventionally give to a `narrow` prop.

borisyankov:
Added a commit to remove `getNarrowToSendTo` and renamed the `narrow` prop in `ComposeMenu` to `destinationNarrow`.

gnprice:
Hmm, this isn't right, in that first commit:
```
compose menu: Cleanup unused `getNarrowToSendTo` selector

We are not currently using it, and the logic was always a bit flaky.
```
We definitely are -- in fact the selector's result is *replacing* the `narrow` prop that the parent passes in. And when the narrow that's passed in is a stream narrow, that's important.

borisyankov:
You are right, I swapped the order of the two commits a bit too hastily.
Now reverted back.
After the `Reuse the destination narrow logic when uploading images` commit, the `getNarrowToSendTo` selector is now not needed.

gnprice:
Sounds good! I think you didn't push the new version, though :wink:

borisyankov:
Ha, indeed. Pushed.

gnprice:
Thanks!

Hmm, the first commit does two or three things now, and one is it "renames" the `narrow` prop of `ComposeMenu`... but there's still a `connect` call passing `narrow`. That feels a bit off.

And in fact Flow gives an error at that commit:
```
Cannot create ComposeMenu element because ST [1] is not a React component.

     src/compose/ComposeBox.js
     303│         />
     304│         <View style={styles.composeBox} onLayout={this.handleLayoutChange}>
     305│           <View style={styles.alignBottom}>
     306│             <ComposeMenu
     307│               destinationNarrow={this.getDestinationNarrow()}
     308│               expanded={isMenuExpanded}
     309│               onExpandContract={this.handleComposeMenuToggle}
     310│             />
     311│           </View>
     312│           <View style={styles.composeText}>
     313│             {this.getCanSelectTopic() && (

     flow-typed/npm/react-redux_v5.x.x.js
 [1] 119│     ST: {[_: $Keys<Com>]: any}
```

Which sounds familiar -- we have a couple of fixmes for those, which we added at the last RN version upgrade because they were baffling to figure out. So now that's a clue about what was causing those errors!

```
$ rg 'is not a React component'
src/main/MainScreenWithTabs.js
21:        {/* $FlowFixMe-56 Cannot create ZulipStatusBar element because ST is not a React component. */}

src/common/Screen.js
104:        {/* $FlowFixMe-56 Cannot create ZulipStatusBar element because ST is not a React component. */}
```

This is an example of why I always want fixmes to have an explanatory comment. :-) I'm going to go see about fixing those two, now that we have this clue to what the error means.

---

Back to this branch: as always I want every commit to leave things non-broken. I think this is also an example where the change is harder to understand for reasons that overlap with why the intermediate code is broken, and making things non-broken will make it easier to understand.

Maybe first rename; then factor out and use that instead of `connect`; then delete the selector definition itself?

borisyankov:
The rename is a separate commit now.
Makes more sense. 👍

gnprice:
Looks great -- thanks for the revisions! Merged.

PR #3177, opened 2018-11-29
Fix prettier error and run prettier locally

borisyankov:
* Run `yarn prettier` to fix formatting
* Include `prettier` test in `yarn test`

Include the `yarn run test:prettier` test when running tests locally
with the default `yarn test`. This will help reduce the commits with
incorrect formatting (for these who run the command)

The order of the command is changed to be consistent with the
`test:full` command that is run on the CI server with the only
the difference is we are not running Coveralls locally.


gnprice:
Thanks @borisyankov ! After merging the first commit, I followed up with some other changes -- see 977596d9e, and also other commits up to 0833e1377.

Issue #2559, opened 2018-05-24
iOS: message list doesn't respect system text size


gnprice:
On my iPhone 6 running iOS 10, I had the Zulip app open to the message list screen, then went over to system settings and changed the text size. The text in the app bar responded appropriately, but the message list didn't change.

This would make most of the app's functionality totally unusable for someone with a visual impairment who needs a large text size.

Compare #444, where a version of the app a year ago didn't respond to text size changes at all except on restart.

I haven't investigated exactly when this happens. If it is (as I assume) mainly about the message list, we're probably dropping the information about the text size when interacting with the webview.

gnprice:
​Just to be clear: this is definitely not the appropriate behavior for our
app. The user doesn't know or care that we've used a webview to implement
the message list -- they just want the text size to be the way they've
configured it.

If you'd like to work up a fix (by making the text in the webview scale in
tandem with the rest of the app), that'd be great. I agree, this is an area
where changes should be tested on a variety of OS versions.

I just tested on Android (8.1 O, with 13.4.88). The text in the message
list scales with my system font size just like the rest of the text in the
app. (When I change the setting and come back to the app, it jumps back to
the unreads screen in the nav; that's slightly annoying but should rarely
matter for users other than ourselves.)

gnprice:
Bumped the priority because we just heard (on support@) from a user who's running into this issue -- making messages unreadable.

PR #3188, opened 2018-12-03
WIP: Webview scaling

borisyankov:


borisyankov:
Ready for review. Commits are independent of each other.

gnprice:
Thanks @borisyankov !

> -  font-size: 14px;
> +  font-size: 0.9em;

Why not 0.83333em?

```
 body {
+  font: ${font};
   font-family: sans-serif;
   line-height: 1.4;
   margin: 0;
   width: 100%;
   max-width: 100%;
-  font-size: 15px;
 }
```

`font` is a shorthand for several different properties, right?
https://developer.mozilla.org/en-US/docs/Web/CSS/font
Yeah, looks like it.

I'm not 100% sure how this interacts with the `font-family` line just after it -- I'd guess that the latter overrides this? But I'd want to know.

What happens if you just delete the `font-size` line?

The last commit also does a number of other things, which should be their own commits so they can be explained in commit messages.

> Make `border-radius` values consistent and scalable

Oh, backing up: do we want to be making all our `px` measurements in terms of `em` instead? My sense (but I'm not certain of this) is that that is not the standard or recommended practice -- that that's what we should do for everything that's directly related to the size of text, but other design elements should be in `px`.

Note that `px` already mean logical pixels, not real physical pixels -- at least I sure hope they mean the same `px` we have on the RN side -- and that there are separate system settings where the user can adjust the size of a logical pixel.

borisyankov:
> Why not 0.83333em?
Do you mean 0.93333em = 14 / 15 ?
Because 14 was an arbitrary number 'a bit less than 15'. 0.9 is the same, while 0.93333 suggests too much specificity (that isn't there)

borisyankov:
> Oh, backing up: do we want to be making all our px measurements in terms of em instead? My sense (but I'm not certain of this) is that that is not the standard or recommended practice -- that that's what we should do for everything that's directly related to the size of text, but other design elements should be in px.

That is actually quite a recommended practice on the web, if you want to scale things more easily. More often than not, you do want to scale everything proportionally, whether because of a setting or maybe because of responsible design. Often `rem` is even better, as nested `em`s can mess up things, and the case we want to scale twice is way rarer than scaling once from the root (`rem` is newer than `em` but still quite old and supported on all browsers we care about)

borisyankov:
> Note that px already mean logical pixels, not real physical pixels -- at least I sure hope they mean the same px we have on the RN side -- and that there are separate system settings where the user can adjust the size of a logical pixel.

Pixels in WebView work exactly as they work on the web. They are totally physical pixels, though sometimes it is a scale of 2x, 3x or some fraction, depending on the zoom.

The power of using `em`/`rem` we can say `body: 32px` and scale everything twice proportionally.

borisyankov:
`font-family: sans-serif;` will indeed mess up with `font` and I had it removed previously. Doing this now.

borisyankov:
Rebased and removed the `font-family` line.

gnprice:
> That is actually quite a recommended practice on the web, if you want to scale things more easily.

Hmm. I guess here's an example on Android: go into system settings -> Accessibility -> Font size, and move it to "Largest".

Elements like the app bar, its icons, and the switch controls in settings don't change size at all, while the text labels do.

That fits with the upstream advice for Android:
https://developer.android.com/training/multiscreen/screendensities
> When defining text sizes, however, you should instead use scalable pixels (sp) as your units (but never use sp for layout sizes). ... For example, when you specify spacing between two views, use dp ... When specifying text size, always use sp[.]

Same results when I try a similar experiment on iOS.

So I think we should probably match the platform's behavior.

gnprice:
> They are totally physical pixels, though sometimes it is a scale of 2x, 3x or some fraction, depending on the zoom.

That sounds like exactly what I mean by "logical pixels, not real physical pixels". :stuck_out_tongue: I'm not actually sure what you mean when you say "totally physical pixels" but they can have a scale factor.

There's some background here that I wrote a few months ago:
https://github.com/zulip/zulip-mobile/blob/master/docs/background/layout.md

> Dimensions in React Native are written without explicit units, and are implicitly in units of **logical pixels**, which we abbreviate `px`. (Android calls this a `dp` or "dip"; iOS calls it a "point".)  One logical pixel corresponds to some number of real physical pixels on the display. [...]

For the normal Android and iOS SDKs, and for React Native which follows them, that scale factor is rarely 1 -- that would be a very low-density display today.

> The ratio of physical pixels to a logical pixel may be an integer like 1, 2,
> or 3 (the default settings on many devices), or another value like 2.61 or
> 3.5 (the defaults on the iPhone 6+/7+/8+ and the Pixel 2 XL respectively),
> or even 4.5 (the largest setting on a Pixel 2 XL).

[This handy guide](https://www.paintcodeapp.com/news/ultimate-guide-to-iphone-resolutions) refreshes me on the parameters of that -- the iPhone 3GS was the last iPhone where the scale factor was 1.

The thing I don't know is whether the `px` inside a WebView matches the `px` in React Native. (Which I do know matches the `dp` or "point" in the underlying platform.)

gnprice:
I'm still a bit confused by the last commit. It seems like it's doing two pretty different things -- we may well want both of them but it seems like the reasons would be pretty separate:
* Stop setting `font-size`; take whatever the browser's default is.
* Switch from `font-family: sans-serif` to either `font-family: -apple-system-body` or `font-family: Roboto, Droid Sans`, depending on platform.

In particular both of these values look like they're for `font-family`, so the switch from saying `font:` to `font-family:` looks like it shouldn't mean anything.

Again:

> What happens if you just delete the `font-size` line?

gnprice:
> We actually are. The only UI element that should not scale is the #scroll-bottom button, everything else is 'text'.

Well, I asked this question in the first place on the "border-radius" commit. That includes changes to selectors like `.avatar img` and `.loading-avatar`. Those don't sound like text to me. Can you explain more what you mean?

gnprice:
> For example, on a desktop with lower rez display they are very likely exactly the physical pixel size, and they originated as such. Only after a lot of years, they started having a multiplier added to them.

Hmm OK but we were talking about mobile. You did say "the web" but it did not occur to me that by that you might mean desktop. The web exists just as much on mobile as it does on desktop -- I think a lot of websites see more traffic from mobile than from desktop these days.

> In comparison, the size on mobile almost immediately started as a 'not an actual pixel'.

This isn't true, though. As I said above, on the iPhone up through iPhone 3GS (so until iPhone 4, announced in 2010), a logical pixel is (normally) the same as a physical pixel. I believe the same was typically true for Android devices in the same period.

There is a feature to change the scaling -- "Zoom" I think Apple calls it -- but I don't think it's commonly used. And it doesn't represent anything that desktop web doesn't have -- your browser has zoom with Ctrl-+ and Ctrl--.

Anyway. This is all rather afield from the pertinent question.

> It don't either, but at least comparing a screenshot of the a 32px in RN 'Send' button and the 32px in WebView Scroll to bottom button, they are the same pixel size in iOS.

That's helpful, thanks.

Here's confirmation in Android docs:
https://developer.android.com/guide/webapps/targeting
> The WebView class and most web browsers on Android convert your CSS pixel values to density-independent pixel values

For iOS, the relevant bit appears to be the `magnification` property here:
https://developer.apple.com/documentation/webkit/wkwebview
which defaults to 1.0, and which isn't mentioned in the RN source. Presumably that means a scale factor relative to the device's normal notion of a pixel, in which case that means it behaves as expected on that side too.

borisyankov:
> That includes changes to selectors like .avatar img and .loading-avatar. Those don't sound like text to me. 

To me, they feel not really like 'text' but they are part of the 'document'. They are not UI elements.
On the other hand some apps I tested do not scale their avatars - Slack, Disqord. The new FB Messenger actually does not scale anything!

borisyankov:
I am more willing to agree avatars should not scale inside the WebView, but I don't like that because we will break the consistency between iOS and Android. Android scales everything without even this PR, iOS does not scale anything. In this context it might make sense to make the scroll to bottom button `em` sized so it is the same for iOS.

borisyankov:
After some more testing on iOS/Android with different sizes... Android also behaves similarly with regards to `px` so it makes sense to turn the avatars to pixels. Will do that.

gnprice:
Merged the first three commits -- thanks! I believe the next few are WIP.

gnprice:
Renamed this PR to reflect its broadened scope. :slightly_smiling_face: 

There are a bunch of changes here that look like they're doing things I'd love to merge. Most of them are marked WIP, so it'd be very helpful if you'd go through them and make whatever further revisions you think they need, and add explanations of what they're doing and why.

Of the two non-WIP commits:

* avatars: This code change looks good!

  The commit message doesn't match, though -- it says 35px seems good, but the code says 40px. Would you update the commit message so it reflects your thinking?

  (Also, the summary line just says "not scale with font" but the actual change makes the avatars a good bit larger at baseline, which is a more conspicuous change. "webview: Make avatars bigger, and not scale with font" would be good.)

* In the "use system fonts" commit, my [comment above](https://github.com/zulip/zulip-mobile/pull/3188#issuecomment-448176793) about the two different things it's doing is the main question I have.

borisyankov:
This is ready for review. 
I have left a few commits that are 'WIP' and are not essential to implementing the issue, but are good-to-have.

I have split the commit that was doing two things, the 'font-size' and system size change. Only the font-size change turned out to be essential, the other one I am experimenting with and might leave out.

The commits are mostly independent.

gnprice:
Thanks @borisyankov ! This looks very helpful.

#### f7b1a0cf3 webview: Use 'rem' instead of 'em' for styling sizes

* Makes sense, thanks! And thanks for the helpful discussion and the links.

* The commit message says this changes all our `em` units to `rem`, but that doesn't seem to be true:
```
$ rg '[0-9]em' src/webview/css/css.js 
110:  padding: 0 1em 1em 4rem;
205:  border-top: 1em solid transparent;
206:  border-bottom: 1em solid transparent;
207:  border-left: 1em solid green;
263:  margin: 1em auto;
293:  margin: 0.2em 0 0 0;
318:  margin: 0.4em 0.4em 0.2em 0;
329:  margin: 0.25em 0;
333:  padding: 0.25em 0.5rem;
350:  margin: .25em 0.5rem .25em 0;
```

(That's from right after this commit. There are also some matches at the tip of the branch, too.)

Should those also be changed the same way?

* The commit message asserts that "in our use" `rem` and `em` are equivalent. This is a key fact to understanding the reasoning of this commit, so thanks for mentioning it. I think one more step of the reasoning would be quite helpful too, which is: why is it true that in our use `rem` and `em` are equivalent? Or: how did you determine that that's true?

  I *think* the answer is that we never have two `font-size` properties on elements nested inside each other, except `body`. But it would be good to confirm that that's your thinking, or even more so to hear what your thinking is if that isn't it. :slightly_smiling_face: 

#### 4d60e29f4 webview: Make avatars not scale with font

* Cool, thanks for the edit.

* This bit still applies:

  > the summary line just says "not scale with font" but the actual change makes the avatars a good bit larger at baseline, which is a more conspicuous change. "webview: Make avatars bigger, and not scale with font" would be good.

#### a7ba6f57c webview: Make some UI elements have size relative to font size

* Cool, this makes sense, except that when the explanation refers to `border-radius`, it's not really about `border-radius` vs. other properties, is it? If I'm understanding the rationale correctly, the key distinction is between elements that are basically text and that aren't.

* Just scanned through this to see that the affected selectors all feel like "text"... and I see one I don't recognize: `.typing-list`.

  Grepping, it looks like the name `typing-list` isn't mentioned anywhere else, in either our repo or the server.

  Do you recognize this? If it isn't something that currently exists, we should just delete the block.


#### 086104633 webview: Stop setting font-size in the 'html' element

* Cool, is this the change that fixes #2559?

  * If so, a line `Fixes: #2559` will be quite helpful for multiple reasons.

* Looking at the code, I thought I understood what it would do based on the previous discussion in this PR... but then looking at the commit message I'm not sure I do after all:

     The default value of the 'font-size' property is 100%. That makes
      it scale on iOS with system settings. (it already scales this way
      even with the previously set 15px).

  The "it already scales this way ..." sounds like it's saying that already, with the existing code before this change, the font size scales according to system settings. But that's exactly the behavior we didn't have and this is intended to get us, right?

  Hmm, perhaps it's missing some words like "On Android,"?

* The commit message says the `html` element, but the code seems to be about the `body`.

* It seems like this will have two effects:
  * Our font size starts scaling with the system font size, fixing #2559.
  * In the default case (for the majority of users who haven't made changes to their system font-size setting), our font size will increase from 15px to 16px.

  I don't have a strong view on whether our default font size should be 15px or 16px, but I definitely don't want to make that change by accident -- it should be a choice we make.

  What are the reasons as you see them for preferring one size over the other? What were your original reasons for choosing 15px? (That was a change you made in 7a6ec3f13 / #1346, when the WebView message list was in alpha.)

* In particular, if I'm correctly understanding your explanations of how this works (in this commit message and earlier in this thread), it should be easy to keep the old default size while getting the scaling: write `font-size: 93.75%`, where I got 93.75% from computing 15/16.

  So in fact I'd like to make these changes as two separate changes (assuming you think we should do both) -- that way it's explicit that we're changing the size by choice and not by accident, and also the separation makes it much easier to clearly explain the reasons for each change.

#### f7b412184 webview: Make padding and margin not scalable with font size

* (This comes after a WIP change to use system fonts -- is it also intended as WIP?)

* Thanks for the helpful explanation of the pros-and-cons reasoning here!

* Thanks also for mentioning that some of the values are being slightly adjusted!

* There are a few spots where there are behavior changes (or at least what *look* in the code like behavior changes) that seem more substantial than rounding like 0.2rem == 3.2px -> 4px:

```
+  padding: 8px;
...
 .topic-header,
 .private-header {
+  padding: 0 8px;
...
-  border-radius: 10px;
+  border-radius: 0.25rem;
...
-  right: 5px;
-  bottom: 15px;
+  right: 8px;
+  bottom: 8px;
```
(possibly a few more.)

Those may all be good changes... but it's a lot harder to understand and think about them when they're mixed in with a larger (in lines of code), more sweeping and boring change that is basically a refactor. So it'd be better to have them separate. And unlike the roundings, for a lot of these there's no obvious reason they're related to the rem->px refactor at all.

Plus, these more substantial changes aren't explained at all in the commit message, or even mentioned. And it'd be hard to keep an explanation coherent if mixed in with the pure refactor + small roundings that are explained there now.

So, let's do those separately, with their own explanation.


#### 62a09b2b5 docs: Add webview.md

* Cool, this looks helpful! Want to move it forward in the branch to before the WIP changes?

borisyankov:
> I think the answer is that we never have two font-size properties on elements nested inside each other, except body. But it would be good to confirm that that's your thinking, or even more so to hear what your thinking is if that isn't it. 

That is exactly the reason. Historically, only 'em' existed. It was used predominantly for 'root font-size scaling' with the added unwanted benefit that it wasn't relative to the root if anyone had set the font size of another container. 'rem' is a newer unit that solves the problem (but not that new to be a concern for us)

https://caniuse.com/#feat=rem

borisyankov:
> The commit message says this changes all our em units to rem, but that doesn't seem to be true:

My search/replace procedure had failed me for the last time! I did search the commit again, also the next commits. I pushed an updated version.

borisyankov:
At the next commit, I did add the suggested change 'webview: Make avatars **bigger, and** not scale with font" would be good.`

borisyankov:
> Grepping, it looks like the name typing-list isn't mentioned anywhere else, in either our repo or the server.
>
> Do you recognize this? If it isn't something that currently exists, we should just delete the block.

Indeed, not used. Added a commit that deletes this and another style that I also found out is not used.

gnprice:
Thanks @borisyankov for the replies and updates!

For b737eb5cd "webview: Make some UI elements have size relative to font size" and later, many of my comments above are still open -- so I think those are still for you to take the next steps on.

For the 4 commits before that:

#### 7b95518ad docs: Add webview.md

* Looks good, thanks! About to merge, except:

* I'm adding a quick remark in the commit message that this is talking about our near-future practices, *not* actually our current practices as of this commit. Otherwise it could be pretty confusing for someone trying to understand it and comparing to the code.


#### eaedf788d webview: Cleanup unused styles

* Looks good, thanks! About to merge.


#### c2b5cfe84 webview: Use 'rem' instead of 'em' for styling sizes

* Thanks for the reply about why this is equivalent for us!

* As always, answers to "why" questions in review belong in the commit message, so they benefit future readers too. I was preparing to merge this and edited the commit to add that explanation, but then...

* Rereading the explanation in context, I noticed a mismatch:

  > * I _think_ the answer is that we never have two `font-size` properties on elements nested inside each other, except `body`.

  > 'rem's are based on 'root em' - the font size of `<html>`

  And indeed, [MDN confirms](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size#Values) `rem` is based on the font size at the `<html>` element... but we've been changing the `font-size` at **`<body>`**, which is a child of `<html>`. So as I understand what the docs are saying, it should actually be the case that `rem` is *not* equivalent to `em` for us -- one is based on the root font-size, and the other is based on the slightly smaller font-size we've had set on the `<body>`.

  Can you help untangle what the story is here? Do you actually find empirically that this change has no effect? If so, can you explain the apparent discrepancy with the docs?

  Alternatively, this change really should have no effect if it comes after the "webview: Stop setting font-size in the 'body' element". So if it does have an effect where it currently is, let's finish that other change and stick it after that one.


#### c103a5cba webview: Make avatars bigger, and not scale with font

* Thanks for the edit! About to merge.

gnprice:
#### [c2b5cfe](https://github.com/zulip/zulip-mobile/commit/c2b5cfe841e2db6bab5713f2dd43dda695faf73c) webview: Use 'rem' instead of 'em' for styling sizes

For reference, my edit to the commit message:
```
    -    This commit changes all our 'em' units to 'rem'. While they are
    -    not equivallent in all cases, in our use they are. This change
    -    is strictly to make the styling more robust and does not change
    -    any elements' size.
    +    This commit changes all our 'em' units to 'rem'.  While they are
    +    not equivalent in all cases, in our use they are, because we never
    +    have a `font-size` property on two elements (other than `body`)
    +    nested inside each other.  This change is strictly to make the
    +    styling more robust and does not change any elements' size.
     
         'em's are relative to current element's font-size while 'rem's
    -    are based on 'root em' - the font size of `<html>` which is
    +    are based on 'root em' - the font size of `<html>` - which is
         exactly what we want.
```

gnprice:
And just merged those three ready commits! Thanks again.

8f5e8fb62 webview: Clean up unused styles.
6bed7dea4 docs: Add webview.md
3dd5fb4de webview: Make avatars bigger, and not scale with font

borisyankov:
> Can you help untangle what the story is here? Do you actually find empirically that this change has no effect? If so, can you explain the apparent discrepancy with the docs?

There indeed is an effect, and the docs are correct. It is just hard to notice the difference between 15px and 16px. (I did confirm it with 150px font-size)

I did move the `font-size` style from `body` to `html` in the same commit to make the unit change really have no effect. Also updated the commit message.

borisyankov:
Dropped the 'WIP: webview: Use system fonts on both iOS and Android' commit, not sure it adds anything.

gnprice:
#### 62c99ffb8 webview: Use 'rem' instead of 'em' for styling sizes

> > Can you help untangle what the story is here? Do you actually find empirically that this change has no effect? If so, can you explain the apparent discrepancy with the docs?
> 
> There indeed is an effect, and the docs are correct. It is just hard to notice the difference between 15px and 16px. (I did confirm it with 150px font-size)
> 
> I did move the `font-size` style from `body` to `html` in the same commit to make the unit change really have no effect. Also updated the commit message.

* Great -- thanks for checking in on that, and that sounds like a good solution!

* One further adjustment: I'm going to split the move of `font-size` from `body` to `html` out as its own separate commit to come first. I think that makes the explanation a fair bit easier to follow.

* Then about to merge this commit.

---

In the second and later commits, I think many of my comments above are still open. We probably won't get through all of them in our remaining time, so I'll probably end up finishing them up myself later, which is fine.

There are a few points where I'd be especially glad to hear your thinking (vs. others where I *think* I understand already and it's just a matter of editing):

#### fa6ffa992 webview: Stop setting font-size in the 'html' element

* > What are the reasons as you see them for preferring one [font] size over the other? What were your original reasons for choosing 15px? (That was a change you made in [7a6ec3f](https://github.com/zulip/zulip-mobile/commit/7a6ec3f133dc4bd31dd06f913c205cd245464fb2) / #1346, when the WebView message list was in alpha.)

* In this bit of the commit message:

      The default value of the 'font-size' property is 100%. That makes
      it scale on iOS with system settings. (it already scales this way
      even with the previously set 15px).

  if I insert "On Android" so it ends with "(On Android, it already scales this way even with the previously set 15px)", does that reflect what you intended to say there? Or did you mean something else?

  If you did mean something other than that, then I don't understand, and please explain. :-)

  > The "it already scales this way ..." sounds like it's saying that already, with the existing code before this change, the font size scales according to system settings. But that's exactly the behavior we didn't have and this is intended to get us, right?

gnprice:
Oh, and another point where I'd be especially glad to hear your thinking:

#### 68828f424 webview: WIP: collapse margins FTW

* I have a vague idea what this is doing, but if you could explain your understanding of what it's doing and why that would be quite helpful. :slightly_smiling_face: 

  In particular: I think at least some of this is intended to cancel out with no net effect on the visible layout. Is that true of all of it, or are there some changes this is intended to make to how things show up for the user?

(The rest of the WIP commits, I think I understand at a mechanical level, and their visual effects will speak for themselves.)

gnprice:
#### 62c99ffb8 webview: Use 'rem' instead of 'em' for styling sizes

* Hmm, in fact as I edit this commit message (with the benefit of having separated out the body->html change first for clarity), I see another discrepancy in the logic of why this has no effect!

  The reasoning I'd offered above was

  > * I _think_ the answer is that we never have two `font-size` properties on elements nested inside each other, except `body`.

  and we then corrected `body` to `html` there.

  But in fact that's still not right -- what we really need in order for `rem` to mean the same as `em` is that we never *use `em` or `rem`* on *any* property on an element that's nested inside any non-root element (i.e. any element other than `html`) that has its own `font-size`. That's rather broader.

* So, I went back through the places we set `font-size`, to check if that criterion is actually met. There are 7 places:
```
html {
code {
pre {
.timestamp {
.loading-spinner, .outbox-spinner {
.message-tags {
#js-error-detailed {
```

Other than `html`, I believe most of these we don't stick any elements inside, only text: `code`, `pre`, `.timestamp`, `.loading-spinner`, `.outbox-spinner`.

Then `#js-error-detailed` we stick text and `<br>` inside, and we have no styling on `<br>`.

Then there's `.message-tags`. We do actually put elements inside those! Namely `span.message-tag`... and in turn we have
```
.message-tag {
  padding: 0.25em 0.5em;
```

So I believe this padding is actually affected by the change: as `em` it's relative to the `font-size: 0.7em` on `.message-tags`, but as `rem` it bypasses that and is relative to the root. So this change makes that padding 1/0.7 = 43% wider.

gnprice:
And merged the change to `rem`, as cacd51e0e and 6ef229e42 ! Thanks for all your work on this.

As I mentioned in the last few comments, there are a couple of questions where I'd be especially glad to hear your thinking -- other than that, I think this is in good shape for me to finish up later, and other PRs will be the priority where I'd most like to have your help in the time remaining.

borisyankov:
> I have a vague idea what this is doing, but if you could explain your understanding of what it's doing and why that would be quite helpful. 🙂

The name of the commit is coming from it using the 'collapsable margin' property of block level emenets in css (so not true for flexbox, css grid, or inline). 

Two blocks next to each other vertically with margins will not have the distance between them be the margin1 + margin2 but the max(margin1, margin2) - so they collapse onto each other.

This leads to simplification and improvement to the layout, in our case and in general. This is true only for the 'margin' property and not 'padding'.

An example of something desirable: a list of a dozen of these `.message-row { margin: 16px; }` will result a list with margin 16px on each side and 16px between each row (instead of 16 + 16 = 32px if it was padding, or any other layout that would not result in collapsing)

> In particular: I think at least some of this is intended to cancel out with no net effect on the visible layout. Is that true of all of it, or are there some changes this is intended to make to how things show up for the user?

The intention is not to preserve the layout visually 100% but to improve it slightly.

borisyankov:
I agree that not all commits are as important, definitely the ones with 'WIP', the reason I didn't polish them or elaborated on them yet (can always do in future).

gnprice:
Very helpful explanation -- thanks!

gnprice:
I've just pushed several changes based on the next few commits in this PR:

642b2d43b msglist: Let font size on iOS scale with system settings.
cbf2e6ce9 msglist: Use default 16px font-size, rather than 15px.
3a9d2aa23 msglist: Make emoji height/width scale with font size.
4955ea5da msglist: Adjust various lengths to multiples of 1/16 rem.
ebaffb456 msglist: Make padding and margin independent of font size.

Thanks again @borisyankov for your work on this!

I'll push here a rough rebased version of the rest.

Issue #3186, opened 2018-12-03
Cannot cancel unsent messages


zingmars:
Might have gone overboard with the blurring, but here's a screenshot -
![image](https://user-images.githubusercontent.com/917126/49382321-d1753880-f71e-11e8-8b61-6c9a6375742a.png)

Basically I have two messages that failed to send for some reason and I cannot cancel them (the cancel button in the menu that comes up does nothing). I've tried force closing and clearing the cache (haven't tried reinstalling the app yet) but the messages always pop back up after I open the chat. Basically the end result is that I have two messages that are always stuck at the bottom that won't send and can't be cleared.

borisyankov:
I think the 'Cancel' option is a bit misleading here. It is duplicating a functionality that is standard in IOS and means 'Cancel this menu'.
You should see a 'Delete' option and that should work for you.

Restarting the app should trigger a proper resend, though it is a bit weird that you have such old messages that are still not sent.

zingmars:
There is no "Delete" option though. All I'm offered are the "Copy to Clipboard", "Share" and "Cancel" options.

borisyankov:
Do you mean that they recently got sent? Or at the time you sent them?

So the bad news is we have few bugs around the unsent messages retrying, duplicating, and now I guess not cleaning up. The good news is that they happen rarely so we didn't got many reports about them.

zingmars:
I actually have no idea when exactly they got sent. In the other person's log the messages appear to be in the right order though so either they got sent when they should have been or during the initial retry. I'm actually not quite sure why they didn't send in the first place (maybe I had a bad signal?).

Anyway, other than clearing all app data and relogging, how would clear those messages out?

borisyankov:
I just did fix another bug, the fix to which I hope will be available soon.
We actually do support the deletion of unsent messages but we incorrectly hide the option!

zingmars:
Alright, I'll wait for that then.

gnprice:
It sounds like the bug here was that the "Delete message" option wasn't showing; that was fixed in #3189.

PR #3189, opened 2018-12-03
outbox: Fix not being able to delete unsent messages

borisyankov:
As pointed in #3186 sometimes our unsent messages might get stuck
and not get sent. Also they can not be deleted.

The cause for the messages getting stuck is yet unknown but is
important to figure out).

Not being able to delete the messages has a simple fix though.
In fact, we already support that but incorrectly determine the
visibility of that opiton.

Remove the `isSentMessage` as a requirement for the 'Delete' option
and allow deletion of the outbox messages.


gnprice:
Thanks @borisyankov ! Merged.

PR #3190, opened 2018-12-03
Rework and fix `progressiveTimeout`

borisyankov:
After a closer look at `progressiveTimeout` with Greg, we found a
piece of logic is wrong, the timeout would not reset if it has
been more than 5 minutes since last timeout instead of the other
way around. Also, the logic was more convoluted than it could be.

This commit does:
 * fix the timeout reset bug
 * simplify the logic
 * tweaks the timing values


gnprice:
Thanks! Merged.

PR #3192, opened 2018-12-04
draftsSelectors: Flow typed selector

VisibleMarkov:
Issue in focus: #2971 

I'm aware of the PR #3166, but `draftsSelectors` wasn't a part of original PR message. I went through the discussion and didn't find any mention of removing this entirely so I went ahead and typed the selector present in `draftsSelectors.js`.

Tests ran:
* [x] npm run test:flow
* [x] npm run test:lint

### DIFF
--- a/src/drafts/draftsSelectors.js
+++ b/src/drafts/draftsSelectors.js
@@ -2,7 +2,7 @@
 import { createSelector } from 'reselect';
 
 import { getDrafts } from '../directSelectors';
-import type { Narrow } from '../types';
+import type { Narrow, Selector } from '../types';
 
-export const getDraftForActiveNarrow = (narrow: Narrow) =>
+export const getDraftForActiveNarrow = (narrow: Narrow): Selector<string> =>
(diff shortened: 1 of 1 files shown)


gnprice:
Looks good, merged -- thanks!

> `npm run test:flow`
> `npm run test:lint`

I'm curious, how did you come across those commands to use? Generally we've recommended `yarn test:flow` and `yarn test:lint`, both for consistency with use of `yarn` elsewhere and because it's one fewer word to type.

Though since yesterday, those are just aliases for a still shorter command: `tools/test` will run all of them, and only on files you've changed so it's very fast. :smile: Or `tools/test flow` etc. for running just one kind of check.

gnprice:
> the way auto-complete works on my terminal, to get yarn, i have to go all the way till the 3rd character but `npm run` is the first thing that pops up when I type in `n` so I just went with that

Maybe this means you want something like `alias y=yarn` in your `.bashrc` :wink: , so you can type just `y flow`.

VisibleMarkov:
That's really smart! I wonder why I din't notice it before :sweat_smile:
Thanks a lot @gnprice

Issue #3178, opened 2018-11-30
npm run test not working


mbuotidem:
I just cloned the latest version of zulip-mobile to work on it and after setting everything up (I.e getting it to run on my android) I decided to run 'npm run test' before making any changes. However, the command does not run and states "Windows Subsystem for Linux has no installed distributions. 

I found this weird because I've successfully run the test before so I restored the old version of zulip-mobile from my Recycle bin, moved it to my Desktop and tried to run the test. It ran without issues. Not sure what I am missing here. 

Please see attached screenshot with both cmd windows running side by side. 
<img width="1347" alt="zulip npm run test" src="https://user-images.githubusercontent.com/4692052/49299891-6e697480-f486-11e8-93f0-fa9b1f782040.PNG">

mbuotidem:
@gnprice Happy to help. Before trying these commands I decided to install WSL and see what difference that would make, if any. Here is the result of that effort. 

1. From Windows cmd

- The `npm run test` command proceeded to setup a 'flow' server but then failed. Screenshot below : 

- Individual runs of `yarn run test:lint`, `yarn run jest`,  and `yarn run test:prettier` all failed. 

![npm run test after wsl install](https://user-images.githubusercontent.com/4692052/49333306-1a43ba80-f582-11e8-9a63-8c0beed8e811.PNG)

2. From Git Bash on Windows:

- The `npm run test` command did not get as far as setting up a 'flow' server. Screenshot below : 

- Individual runs of `yarn run test:lint`, `yarn run jest`,  and `yarn run test:prettier` all failed. 

![git bash wsl on](https://user-images.githubusercontent.com/4692052/49333310-20399b80-f582-11e8-9f0c-39389b6f3043.PNG)

3. From WSL on a WSL specific zulip-mobile folder

- The `npm run test` command proceeded to setup a 'flow' server but then failed. Screenshot below : 

- `yarn run test:lint` also failed **WHILE** `yarn run jest` and `yarn run test:prettier` succeeded.


![wsl](https://user-images.githubusercontent.com/4692052/49333313-2596e600-f582-11e8-8802-bf89979941db.PNG)

I am now going to remove WSL and try the instructions in the link you posted.

gnprice:
Thanks @mbuotidem !  I will be curious to hear back on how the instructions I posted go for you. :-)

---

Separately, I am interested in getting a good set of instructions for setting up the dev environment on WSL. We actually have a rough version already:
[...]

PR #3193, opened 2018-12-04
Update WSL Setup documentation

mbuotidem:
Issue in focus: #3178. 

Updated the WSL draft guide with step by step instructions and re-organized the document flow for better readability.  Also removed some suggestions that could lead users down some frustrating dead ends.

### DIFF
--- a/docs/howto/windows.md
+++ b/docs/howto/windows.md
@@ -1,18 +1,17 @@
 # Developing on Windows using WSL
 
 This is a guide on doing Zulip Mobile development on a Windows machine
-using WSL. Please note that you should be running Windows 10 Fall
-Creators Update or newer.
+using WSL.
 
 Using WSL is not a requirement for working on Zulip Mobile. If you
 follow our step-by-step [main install instructions](build-run.md) on
 Windows, they should work just fine.  You'll use the basic Windows
 "Command Prompt", aka `cmd.exe`, for your command line.
 
-However, If you'll be spending a lot of time developing Zulip Mobile, or if you're
-already very familiar with the command line on Linux or macOS,
-it's highly recommended that you install [WSL][wsl-home] / "Ubuntu for
-Windows", and use `bash` instead of `cmd.exe` for your main shell prompt.
+However, if you'll be spending a lot of time developing Zulip Mobile, or if
+you're already very familiar with the command line on Linux or macOS, it's
+highly recommended that you install [WSL][wsl-home] / "Ubuntu for Windows",
+and use `bash` instead of `cmd.exe` for your main shell prompt.
 
 WSL is a [pretty magical][wsl-announce-ubuntu] piece of engineering, and it
 gives you access to the same ecosystem of command-line tools that people on
@@ -40,7 +39,6 @@ approach to get some of the same benefits is [Cygwin](http://www.cygwin.com/).
 If you try it out with Zulip Mobile, a PR to add a section to this guide
 would be welcome!)
 
-
 [wsl-faq]: https://docs.microsoft.com/en-us/windows/wsl/faq
 
 
@@ -50,11 +48,11 @@ wide a range of contributors as most of our developer documentation.
 If you're trying them, we would love to hear from you [in #mobile on
 chat.zulip.org][czo]!
 
-Again, if you aren't already comfortable with the Linux (or macOS) command
-line, or just want to get things working quickly and
+Again, if you aren't already comfortable with the Linux (or macOS)
+command line, or just want to get things working quickly and
 straightforwardly, you may prefer to skip these beta instructions and
-instead follow our step-by-step [main install
-instructions](build-run.md) natively on Windows.
+instead follow our step-by-step [main install instructions](build-run.md)
+natively on Windows.
 
 [czo]: ../../README.md#discussion
 
@@ -71,7 +69,6 @@ You'll want to use the Windows side of things for:
 * Android Studio
 * the Android emulator
 * the location of your development tree in the filesystem
-
 
 What are the reasons to put a given piece on the Linux side or the
 Windows side?  (See the [WSL FAQ][wsl-faq] for more information on
@@ -91,13 +88,22 @@ many of these.)
   but it may be a bit simpler to use the Linux version.
 * One significant limitation of WSL (as of 2018) is that doing a large
   amount of filesystem I/O tends to be slow.  This is a reason to do
-  operations like a Gradle build (Gradle is the Java program that does the Android build), perhaps `git` commands, and perhaps
-  the occasional `yarn install` using the Windows versions.
+  operations like builds of the app (i.e. running Gradle), perhaps
+  `git` commands, and perhaps the occasional `yarn install` using the
+  Windows versions.
 
 [so-guide]: https://stackoverflow.com/questions/42614347/running-react-native-in-wsl-with-the-emulator-running-directly-in-windows
 
 
 ### Step-by-step instructions for setting up WSL
+
+These are primarily one Zulip contributor's notes from setting up
+their dev environment on WSL.  Again, if you're trying them, please
+come and say hello [in #mobile on chat.zulip.org][czo]!
+
+0. Update Windows to the latest version.  To use WSL you'll need at least
+   the Windows 10 Fall Creators Update, released in late 2017.  (For
+   options on Windows 8 and older, see above.)
 
 1. Open Powershell as an Administrator and run the command below to enable WSL.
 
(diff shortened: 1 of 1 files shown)


docs/howto/windows.md
```diff
@@ -79,14 +68,6 @@ You'll want to use the Windows side of things for:
...
 * the location of your development tree in the filesystem
 
-We're not sure yet which side you'll want to use for the following;
-try them on both the Linux side and the Windows side, and let us know
-which you prefer:
-* the `git` command line (from your `bash` shell either way, but you
-  might have it run the Windows version)
-* Gradle, which is the Java program that does the Android build
-  (again, invoking it from your `bash` shell either way)
-* the JDK, which is used for running Gradle
```
gnprice: Do you have a conclusion on which side you prefer for each of these to run on? If so, what did you learn that led to that conclusion?

More broadly, what's your thinking behind removing this paragraph?

docs/howto/windows.md
```diff
@@ -106,28 +87,132 @@ many of these.)
...
 
-### Accessing the filesystem
-You can access your Windows local drives from `bash` and other Linux
-programs by navigating to `/mnt/<drive-letter>` .
 
-### Running `yarn` commands
-First, install Node.js and Yarn in your Linux environment.  You'll follow
-the same steps as in [our main setup guide](build-run.md) -- just use the
-Linux versions. :-)
+### Step-by-step instructions for setting up WSL
```
gnprice: Awesome, thanks for these detailed step-by-step instructions!

Can you say more about how you worked out these steps? Were there existing instructions online that you were able to use for parts of them? Is it sort of a cleaned-up transcript of the steps you ended up taking to set things up?

The answer to how you worked them out doesn't necessarily go in the doc, but I definitely want to know :) and it probably goes in the commit message too. One reason is that inevitably there will be things that don't work for some people, and then it'll be super helpful to know why that step was there as we're figuring out how to fix it for them. Also, the right instructions will change over time, so knowing why the different pieces are there will again be super helpful for keeping it up to date with changes in WSL and the various tools we install.

docs/howto/windows.md
```diff
@@ -106,28 +87,132 @@ many of these.)
...
+
+1. Open Powershell as an Administrator and run the command below to enable WSL. 
+
+    `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux`
+
+2. Install [Chocolatey][chocolatey]. After installing Chocolatey, use it to install the JAVA SE Development Kit with
```
gnprice: This is one I'm particularly curious about, because I have actually never heard of Chocolatey!

From the web, I see it's a popular package manager for Windows. Did you mention Chocolatey here because it's just what you use routinely to install software? Or is there something more specific about why Chocolatey in particular?

docs/howto/windows.md
```diff
@@ -106,28 +87,132 @@ many of these.)
...
-Linux versions. :-)
+### Step-by-step instructions for setting up WSL
+
+1. Open Powershell as an Administrator and run the command below to enable WSL. 
+
+    `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux`
+
+2. Install [Chocolatey][chocolatey]. After installing Chocolatey, use it to install the JAVA SE Development Kit with  
+
+    `choco install -y jdk8`
```
gnprice: Huh, also I am curious why you needed this step at all. I'd expect that the Android Studio installer would install a suitable JDK if you don't already have one.

docs/howto/windows.md
```diff
@@ -106,28 +87,132 @@ many of these.)
...
+
+3. Follow the [instructions][android-studio] here to install Android Studio. 
+
+4. In Step 3 above, you saw how to add the `ANDROID_HOME` environment variable. Return to the Environment Variables screen. Scroll down in the User variables section where you added `ANDROID_HOME` until you find a variable called `Path`.
```
gnprice: Dang, this step is a bit messy.

That is a reason I'm glad to have detailed step-by-step instructions like these!

It also means this is a step where I'm curious to better understand what it's doing (what goes wrong if you leave out one of the substeps here?), and also how you determined this was the set of actions to take to handle that.

docs/howto/windows.md
```diff
@@ -106,28 +87,132 @@ many of these.)
...
+
+5. Go to the Windows Store and type the name of the linux distro you would like to use with WSL. These instructions have been tested to work on Ubuntu and DebianGNU/Linux. When the page for your selected distro opens. Click on 'Get' to install it. 
+
+6. You will receive a notification when the installation has completed. Click on 'Launch' to open it. When it opens, you will see the message "Installing, this may take a few minutes...". Soon after that, you will be prompted to enter the username and password you wish to use. If you haven't used a UNIX based system you will notice that nothing shows up while you are typing your password. This is normal and expected behavior.  
+
+7. After setting up your username and password, run 
+
+    `sudo apt update && sudo apt -y upgrade`
+
+    This will initialize your WSL installation and update your distro. This could take a while so grab a cup of coffee. (You can paste this and the rest of the commands in this guide into the WSL window using your right-click button. Ctrl + V will not work. )
```
gnprice: Huh interesting. Important tip!

Does Shift + Insert work? (That's an alternate key combo for "paste" that I remember using a lot on Windows. Also Ctrl + Insert for "copy" and Shift + Delete for "cut", if I recall right.) If so, that might be worth mentioning too.

docs/howto/windows.md
```diff
@@ -136,12 +221,24 @@ export PATH="$HOME/bin:$HOME/.local/bin:/usr/bin:$PATH"
...
 ```
 
-### Run emulator
-Run your app in the emulator using `react-native run-android`.  Now you can
-perform any terminal commands inside the Linux environment, including
-`react-native start`, and the emulator will load the bundle correctly.
+### Possible Issues
+
+#### JAVA_HOME is set to an invalid directory
+If you run into an error that says that 'JAVA_HOME is set to an invalid directory' consider using WSLENV to share the `JAVA_HOME` environment variable with your WSL installation. This only works if you have  Windows 10 Build 17064 or newer.
```
gnprice: Huh, I can't find much about Build 17064 on the web.

Do you perhaps mean 17063? Here's an announcement of WSLENV:
https://blogs.msdn.microsoft.com/commandline/2017/12/22/share-environment-vars-between-wsl-and-windows/

gnprice:
Merged -- thanks again @mbuotidem !

As I said above, I am still quite interested in the above questions, and further improvements to these instructions; just don't want those to get in the way of forward progress like this change :smile:

docs/howto/windows.md
mbuotidem: I left the paragraph because I simply was unable to get Gradle and the JDK working on the WSL side. I encountered several errors that I was unable to figure out despite some extensive googling. I am still trying to figure it out since I'm unhappy with the number of steps required to get setup when you have to install the JDK and Android Studio on the Windows side and I know if everything was done from WSL I could probably half the steps. But until we have a clear path for a new user to follow, I hesitate to have them spend as many hours as I did and give up halfway.

docs/howto/windows.md
mbuotidem: Not a problem. I figured the steps through a combination of the initial draft that was already here, and then searching for answers whenever I encountered an error on the way. A lot of the steps in the documentation are therefore just hedges. Step 4 is a case in point. One one computer I worked on while testing these instructions, some of the steps in 4 were not needed. However,  a different Windows machine required every single step before it was happy to proceed.

docs/howto/windows.md
mbuotidem: I had never heard of Chocolatey before this project either. It was actually recommended by the [React Native](https://facebook.github.io/react-native/docs/getting-started.html#node-python2-jdk) Official Docs.

docs/howto/windows.md
mbuotidem: This is one part of the documentation I really don't like and plan eliminate. I will be taking out each of the substeps and trying to see if it still works and then I can be certain of what's necessary and what's superfluous. Also, I just discovered that Path can be set via the Powershell so I'd probably change that section to include the necessary Powershell commands as these steps are indeed messy.

docs/howto/windows.md
mbuotidem: Good catch! I meant Build 17063.

docs/howto/windows.md
mbuotidem: I tried those out but they don't work. I discovered the right click trick by accident so I never went on to search for a solution for that. Just did and found that there's an [option](https://blogs.msdn.microsoft.com/commandline/2018/04/13/copy-and-paste-arrives-for-linuxwsl-consoles/) to enable "Ctrl+Shift+C/V" for copy paste so I'll include a link to that in the next version of the doc.

PR #3194, opened 2018-12-05
Initial load rework

borisyankov:
Now standalone PR. Separated the `LOADING_*` actions in another PR.


gnprice:
Merged one of these as 80348112f -- thanks!

borisyankov:
This is now ready for a review (and hopefully a merge).
Most commits are independent.

gnprice:
```
    Fetch users outside 'register' to quickly re-establish queue
    
    Stop fetching the users list on the initial `register` call.
    That allows us to re-establish the event queue faster, especially
    with large amount of users, and avoid missing events.
---

     'realm_emoji',
     'realm_filters',
-    'realm_user',
     'realm_user_groups',
     'subscription',
```

This breaks the basic invariant of the event system: that we get a snapshot of the state we're interested in, which is atomic with the start of our event queue, so that by applying future events as they come in we maintain a consistent, and up-to-date, view of the data.

Losing that means we introduce race conditions -- when something changes around the time we start the event queue, we might effectively see the change twice or zero times, instead of exactly once. Whatever data we get wrong may stay permanently wrong until the next time we completely reload that data from scratch.

It's possible that in certain situations, after carefully reasoning out what the possible consequences are we can decide they're (a) OK for reasons specific to that data and how we handle it, or (b) an acceptable sacrifice for performance. But that careful reasoning is essential. (And, like all important reasoning about our code: if it isn't explicit in the code or a commit message, it didn't happen.) Just breaking the invariants casually like this, without even mentioning the fact, is not acceptable.

Would you like to try again, this time with a clear explanation of that essential reasoning?

I think some of the other changes are mergeable; more comments and/or merging coming separately.

gnprice:
I believe with #3262 this has been superseded. Thanks again!

PR #3197, opened 2018-12-07
Fix password input

borisyankov:
This fixes an annoying issue that makes the size of the input
about half what it should be (on inspection 16px vs 33px for the
default one).

Tested on iOS and Android.


gnprice:
Thanks @borisyankov for the swift fix and the clear explanations! Merged.

I like the ViewPlaceholder idea -- I don't think I'd spotted that pattern before.

PR #3198, opened 2018-12-08
Fix PasswordInput not passing onChangeText prop to Input

agzuniverse:
In 4dd4752d8 which made the props passed down to Input from `PasswordInput` explicit, `onChangeText` was not passed down.

This commit fixes this by passing down `onChangeText` from `PasswordInput` to `Input`.

Reported in chat [here](https://chat.zulip.org/#narrow/stream/48-mobile/subject/new.20login.20screen/near/669460).

### DIFF
--- a/src/common/PasswordInput.js
+++ b/src/common/PasswordInput.js
@@ -50,12 +50,13 @@ export default class PasswordInput extends PureComponent<Props, State> {
   };
 
   render() {
-    const { placeholder } = this.props;
+    const { placeholder, onChangeText } = this.props;
     const { isHidden } = this.state;
 
     return (
       <View>
         <Input
+          onChangeText={onChangeText}
           placeholder={placeholder}
           secureTextEntry={isHidden}
           autoCorrect={false}


gnprice:
Thanks @agzuniverse !

This fix isn't quite right, though -- and `yarn flow` (or `tools/test`) tells you why :smile: :
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/common/PasswordInput.js:53:26

Property onChangeText is missing in Props [1].

 [1] 40│   props: Props;
       :
     50│   };
     51│
     52│   render() {
     53│     const { placeholder, onChangeText } = this.props;
     54│     const { isHidden } = this.state;
     55│
     56│     return (
```

Basically the root cause of this bug in the first place is that the type we'd written for `Props` didn't include all the props that this component actually required -- so when I wrote out the props explicitly in 4dd4752d8, and relied on the type, I left one out. In fixing it, we should avoid propagating that problem :wink: 

... Oh gosh, and in fact when I make that type exact, Flow informs me that there are a bunch more props that were also not mentioned! I'm going to go fix these -- and leave behind a type that says what's really happening so there's no such confusion in the future.

gnprice:
Fixed in 960d9ae67; see also the other commits in 3631e2548^..a15c446d2.  Thanks again!

PR #3199, opened 2018-12-09
webView, css, dark mode: Add BG color for message containing mention.

jainkuniya:
Background color is same as web app is using i.e `hsla(8, 78%, 43%, 0.15)`.

https://github.com/zulip/zulip/blob/master/static/styles/night_mode.scss#L348

Before
image

After
image

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -455,6 +455,9 @@ body {
 .highlight {
   background-color: hsla(51, 100%, 64%, 0.42);
 }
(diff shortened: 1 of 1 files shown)


src/webview/css/css.js
```diff
@@ -455,6 +455,9 @@ body {
 .highlight {
   background-color: hsla(51, 100%, 64%, 0.42);
 }
+[data-mentioned="true"], [data-wildcard_mentioned="true"] {
```
borisyankov: We do already have styles for this.
Did you mean to edit these instead of overwriting?

PR #3208, opened 2018-12-12
Remove sent messages from outbox on initial fetch

### LINKED ISSUE #3203: Unsent message stays stuck with spinner even after sent
@armaanahluwalia observed the following on 20.0.104 (as summarized by @gnprice after some debugging):
"""
* A message didn't send initially,
* but then it did send almost an hour later,
* and we got new messages in that thread (maybe via the event queue, maybe via a fetch when you went back to that narrow, not sure yet) -- including that message itself,
* but we didn't match up that message with the message we'd tried to send, so we've kept showing the trying-to-send message.
"""

The "unsent" message is always shown at the bottom, so it makes it look like we're also not getting new messages -- they just show up above the old, "unsent" message.

---
Original description:

An issue about messages failing to be sent has been reported several times in the chat:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Unread.20message.20bug
https://chat.zulip.org/#narrow/stream/48-mobile/subject/messages.20not.20loading

Starting investigating the issue and marking this as P0 as it looks frequent enough thus severe.
If it is a regression it a release blocker.
If not a regression but an existing behavior, still a high-priority issue to work on.

borisyankov:
Fixes #3203

We do not delete sent messsages from outbox immediately on a
successful sending, but keep them in the outbox until we receive
the message from the server via our event queue.

There are two cases where this is a problem:
 * the queue dies / gets stuck and we do not reveive that message
 * the user quits the app before the response

The second option should be rare but totally possible. The first
option should be rare but currenlty is more frequent than optimal.

Removing the sent messages on initial fetch is a very safe and
conservative behavior. We might decide to do this more aggressively
in future but that might also lead to behavior confusing to the
user like "Where did my message disappear?"

### DIFF
--- a/src/outbox/__tests__/outboxReducer-test.js
+++ b/src/outbox/__tests__/outboxReducer-test.js
@@ -1,10 +1,34 @@
 import deepFreeze from 'deep-freeze';
 
 import outboxReducers from '../outboxReducers';
-import { MESSAGE_SEND_START, EVENT_NEW_MESSAGE } from '../../actionConstants';
+import {
+  INITIAL_FETCH_COMPLETE,
+  MESSAGE_SEND_START,
+  EVENT_NEW_MESSAGE,
+} from '../../actionConstants';
 import { streamNarrow } from '../../utils/narrow';
 
 describe('outboxReducers', () => {
+  describe(INITIAL_FETCH_COMPLETE, () => {
+    test('filters out isSent', () => {
+      const initialState = deepFreeze([
+        { content: 'New one' },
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @borisyankov !

What's the reasoning in the choice of INITIAL_FETCH_START as the trigger for this? I think my initial thought would be to do it at INITIAL_FETCH_COMPLETE.

gnprice:
Thanks again @borisyankov ! Merged.

PR #3210, opened 2018-12-13
Fix Autocomplete sometimes not responding to touch

### LINKED ISSUE #3209: Autocomplete sometimes doesn't respond to touch
When an autocomplete popup appears (for any of @-mentions, stream mentions, or emoji) and I touch one of the results to choose it, sometimes it doesn't work. Instead, the touch seems to go through to whatever's behind the popup -- so the effect might be to hide the keyboard, or take me to the lightbox, or to the info screen for a user, etc.

This doesn't always happen (sometimes it works fine) and I'm not sure of the exact pattern, but here's a recipe which seems to reproduce it 100% of the time, on my Android device and in the Android emulator:
* Switch accounts.
* Navigate to a stream narrow. (Probably any conversation will do? Haven't tried.)
* Type `@g`, or some other prefix, to cause an autocomplete popup to appear.
* Try tapping on one of the autocomplete suggestions.

This is a regression in 21.0.104 (and 21.1.105). Using `git bisect` to do a binary search, I found that it was introduced in 824c908cc aka #3065. It doesn't seem obvious how that caused this issue, though, even when I reread it now knowing that it somehow did.

Marking as priority because this is pretty annoying when you want to @-mention someone -- effectively it will mostly mean the user can't. I would revert that change until we figure it out, if it weren't itself a fix for a priority issue.

borisyankov:
Fixes #3209

Back in #3065 we did introduce a new behavior for `ComposeBox` to
not unmount when one can't write to a narrow. I would not call it
a 'hack' as this approach should have worked fine (and would on web)
Seemingly it is throwing React Native off and showing the Autocomplete
visually above the WebView but below in the stack of 'touchable' targets.

Luckily, a better, less hacky approach is possible. Just use a familiar
style from the web: `display: 'none'`

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -284,7 +284,7 @@ class ComposeBox extends PureComponent<Props, State> {
     const placeholder = getComposeInputPlaceholder(narrow, auth.email, users);
     const style = {
       marginBottom: safeAreaInsets.bottom,
-      ...(canSend ? {} : { opacity: 0, position: 'absolute' }),
+      ...(canSend ? {} : { display: 'none' }),
     };
 
     return (


gnprice:
Merged -- thanks @borisyankov for the swift fix!

PR #3211, opened 2018-12-13
webView: Fix `generatedEs3.js` is not updated with `js.js`.

jainkuniya:
`466329641c026807b0f0f497ed7a02830e0c1fa3` updated the `js.js` but
forgot to generate webview js file. Thanks to travis for pointing out
that `generatedEs3.js` is not updated with `js.js`.


gnprice:
Ah indeed, thanks @jainkuniya -- merged!

Issue #3025, opened 2018-10-04
Loading indicator on unreads screen


gnprice:
After the app launches, it takes us too long to get to showing up-to-date information on the unreads screen (the home screen of the nav). After the spinner completes and the unreads screen shows up, it at first shows stale data from when we last ran, while asking the server for newer data; the update to newer data takes ~2s for me on a good connection, often 3-5s, and for some users longer.

The fact that it takes that long is a problem in itself, and we should do better. That's a matter for another issue; it won't be simple to fix.

A thing that exacerbates the effect of this slowness is that for those seconds while we're showing stale data, we don't show any indication of this at all -- the screen looks exactly like it would have when that information was fresh, hours or days ago. That undermines users' trust in what they see in the app.

So, in this situation we should show an indication that we're still connecting. One reasonable form for this would be a banner a lot like our `OfflineNotice`, but instead of "No Internet connection", saying something like "Connecting to server...".

gnprice:
The title just says "unreads screen" for brevity, but I'm pretty sure we want it on at least the PMs and streams screens too for similar reasons.

armaanahluwalia:
@gnprice I’d like to work on this. @zulipbot claim.

sstangl:
Instinctively, if I know that the screen is out of date, I usually attempt swiping up, to scroll down "past the end" of the screen, which in many native applications induces refresh behavior.

The app does not support scrolling past the end -- normally, apps will let you do that, and then give a little flick up back to the locked scroll position. Secondly, that action doesn't do anything.

Even if the server took a while to load messages, that would be more OK if I felt that I could instruct the app to try to fetch more messages by swiping down / scrolling up. At least then I would know that the messages are coming. Currently I don't know what the internal app state is at all.

gnprice:
@sstangl Yeah, I agree. I forgot to cross-reference when I filed this issue: #2712 and #2713 call for pull-to-refresh behavior like you describe.

I think even if we already had pull-to-refresh, this loading indicator would still be important. In particular for the path from "load the app for the first time in a while" -> "see up-to-date conversations, and be confident they're up to date", which is a common interaction that it's important to make quick and easy wherever conditions make it possible. Once we have a loading indicator and a user learns to expect it, they'll get that confidence in a few hundred ms when it doesn't appear.

gnprice:
[...]

PR #3212, opened 2018-12-14
Fix #2712, #3025.  WIP: Handle unreadCards onRefresh.

### LINKED ISSUE #2712: Pull-to-refresh on unreads screen and streams/PMs screens
We often have bugs with the symptom that the app doesn't update its view of what messages the user has gotten. This can be frustrating, but as a user what really makes it frustrating rather than mildly annoying is that there's nothing you can do about it, short of force-closing the app and reopening. This has been a recurring complaint from our most active users for basically as long as the app has existed.

In many apps like Gmail, the presented information (e.g., the emails in your inbox) may update on its own but also the user can force a refresh right now, with a pull-to-refresh gesture -- when scrolled to one end of the list, pull as if to scroll further. The app refreshes the info, and a spinner appears while it does so.

We should borrow the same feature.

Ideally the experience we would like to present would make this unnecessary -- the user would see the app automatically keep itself up to date on all the information it's presenting, and would never need this. In the past we've rejected requests for this feature (#1473; #1611 if I'm correctly interpreting that issue; many discussions in chat), often on these grounds. Empirically we have not lived up to that promise, and we should give the user a reasonable tool to handle that.

This issue is for providing this feature on the screens of the main nav: the unreads screen, and the streams and PMs screens. A second issue will track providing this feature on the message list, which is likely to work differently.

jainkuniya:
https://cl.ly/1fb5f6cf11b8

@timabbott Is there any endpoint to get only `unread` data from server, which can be called on refreshing unread screen. Currently we are getting unread data through `/register` endpoint.


src/unread/UnreadCards.js
```diff
@@ -28,7 +28,24 @@ type Props = {|
   unreadStreamsAndTopics: any /* UnreadStream[] */,
 |};
 
-class UnreadCards extends PureComponent<Props> {
+type State = {|
+  refreshing: boolean,
```
borisyankov: That should be coming from the `loadingReducers` state, likely.

borisyankov:
I still do not agree with the generic 'refresh stuff' approach.
We currently do not have a known bug that will desync the unread count.
Providing a 'refresh' functionality will only be confusing to the users, as no effect will take place in the absence of a discrepancy between server and client.

jainkuniya:
This will also show a nice indicator that app is updating/fetching unread data on startup. #3025

jainkuniya:
This loading indicator is only shown one time (just after login). For further app launch, this condition is always false.
```
if (unreadStreamsAndTopics.length === 0 && conversations.length === 0) {
      return isLoading ? (
```
Because we are storing `unreadStreamsAndTopics` in redux store.

jainkuniya:
And currently there is no indicator that, app is fetching/updating unread data. #3025

jainkuniya:
> Is there any endpoint to get only unread data from server, which can be called on refreshing unread screen. Currently we are getting unread data through /register endpoint.

Currently there is no such endpoint. It is easy to make such API, but using such api can lead to race conditions. 

> Steve Howell: @jainkuniya No, we don't have a direct endpoint for aggregating unread counts, such as you found in via register in the unread_data field.
The current paradigm is that you get aggregated counts up front, then maintain your data as you see messages coming in from fetches (they have an unread flag) as well as message updates.
Steve Howell: It wouldn't take much effort to create an endpoint for this, but we'd want to use it very judiciously, as it's expensive, and it's a bit prone to race conditions if you use it in the middle of a session.

https://chat.zulip.org/#narrow/stream/3-backend/topic/api.20to.20get.20only.20unread.20data

So, should we re-register on such refresh?

PR #3213, opened 2018-12-14
Clean state on `LOGOUT`.

jainkuniya:
Do we also need test for this? :)

### DIFF
--- a/src/alertWords/alertWordsReducer.js
+++ b/src/alertWords/alertWordsReducer.js
@@ -5,7 +5,7 @@ import type {
   RealmInitAction,
   EventAlertWordsAction,
 } from '../types';
-import { REALM_INIT, INIT_ALERT_WORDS } from '../actionConstants';
+import { REALM_INIT, INIT_ALERT_WORDS, ACCOUNT_SWITCH, LOGOUT } from '../actionConstants';
 import { NULL_ARRAY } from '../nullObjects';
 
 const initialState = NULL_ARRAY;
@@ -21,6 +21,10 @@ export default (
   action: AlertWordsAction,
 ): AlertWordsState => {
   switch (action.type) {
+    case ACCOUNT_SWITCH:
+    case LOGOUT:
+      return initialState;
+
     case REALM_INIT:
       return realmInit(state, action);
 
--- a/src/chat/flagsReducers.js
+++ b/src/chat/flagsReducers.js
@@ -15,6 +15,7 @@ import {
   EVENT_UPDATE_MESSAGE_FLAGS,
   MARK_MESSAGES_READ,
   ACCOUNT_SWITCH,
+  LOGOUT,
 } from '../actionConstants';
 import { deeperMerge } from '../utils/misc';
 
@@ -104,6 +105,7 @@ const markMessagesRead = (state: FlagsState, action: MarkMessagesReadAction): Fl
 
 export default (state: FlagsState = initialState, action: FlagsAction): FlagsState => {
   switch (action.type) {
+    case LOGOUT:
     case DEAD_QUEUE:
     case ACCOUNT_SWITCH:
       return initialState;
--- a/src/mute/muteReducers.js
+++ b/src/mute/muteReducers.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import type { MuteState, MuteAction, RealmInitAction, EventMutedTopicsAction } from '../types';
-import { REALM_INIT, ACCOUNT_SWITCH, EVENT_MUTED_TOPICS } from '../actionConstants';
+import { REALM_INIT, ACCOUNT_SWITCH, EVENT_MUTED_TOPICS, LOGOUT } from '../actionConstants';
 import { NULL_ARRAY } from '../nullObjects';
 
 const initialState: MuteState = NULL_ARRAY;
@@ -13,6 +13,7 @@ const eventMutedTopics = (state: MuteState, action: EventMutedTopicsAction): Mut
 
 export default (state: MuteState = initialState, action: MuteAction): MuteState => {
   switch (action.type) {
+    case LOGOUT:
     case ACCOUNT_SWITCH:
       return initialState;
 
--- a/src/realm/realmReducers.js
+++ b/src/realm/realmReducers.js
@@ -97,6 +97,7 @@ const eventRealmEmojiUpdate = (
 
 export default (state: RealmState = initialState, action: RealmAction): RealmState => {
   switch (action.type) {
+    case LOGOUT:
     case ACCOUNT_SWITCH:
       return initialState;
 
@@ -109,7 +110,6 @@ export default (state: RealmState = initialState, action: RealmAction): RealmSta
     case DELETE_TOKEN_PUSH:
       return deleteTokenPush(state, action);
 
-    case LOGOUT:
     case LOGIN_SUCCESS:
       return loginChange(state, action);
 
--- a/src/streams/streamsReducers.js
+++ b/src/streams/streamsReducers.js
@@ -13,6 +13,7 @@ import {
   EVENT_STREAM_REMOVE,
   EVENT_STREAM_UPDATE,
   ACCOUNT_SWITCH,
+  LOGOUT,
 } from '../actionConstants';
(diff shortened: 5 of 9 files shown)


gnprice:
Hmm, interesting. It does feel a bit haphazard what state we reset on LOGOUT, what on ACCOUNT_SWITCH, and what on LOGIN_SUCCESS. Even in the case of LOGOUT after this branch, there are some exceptions to the pattern (at least `topics`); see transcript below.

@borisyankov do you have thoughts on what state changes we should be making at each event? I think it'd be good to (a) write down something consistent (perhaps in jsdoc on the action type?), and then (b) sweep through making them all match that.

```
$ rg --sort-files -C1 -e LOGOUT: -e LOGIN_SUCCESS: -e ACCOUNT_SWITCH: src/**/*Reducers.js
src/account/accountsReducers.js
87-
88:    case ACCOUNT_SWITCH:
89-      return accountSwitch(state, action);
90-
91:    case LOGIN_SUCCESS:
92-      return loginSuccess(state, action);
93-
94:    case LOGOUT:
95-      return logout(state, action);

src/caughtup/caughtUpReducers.js
68-    case DEAD_QUEUE:
69:    case LOGOUT:
70:    case LOGIN_SUCCESS:
71:    case ACCOUNT_SWITCH:
72-      return initialState;

src/chat/fetchingReducers.js
55-    case DEAD_QUEUE:
56:    case LOGOUT:
57:    case LOGIN_SUCCESS:
58-    case DO_NARROW:
59:    case ACCOUNT_SWITCH:
60-      return initialState;

src/chat/flagsReducers.js
107-  switch (action.type) {
108:    case LOGOUT:
109-    case DEAD_QUEUE:
110:    case ACCOUNT_SWITCH:
111-      return initialState;

src/chat/narrowsReducers.js
105-    case DEAD_QUEUE:
106:    case LOGOUT:
107:    case LOGIN_SUCCESS:
108:    case ACCOUNT_SWITCH:
109-      return initialState;

src/drafts/draftsReducers.js
25-  switch (action.type) {
26:    case LOGOUT:
27-      return initialState;

src/loading/loadingReducers.js
58-    case DEAD_QUEUE:
59:    case ACCOUNT_SWITCH:
60-      return initialState;

src/message/messagesReducers.js
144-    case DEAD_QUEUE:
145:    case LOGOUT:
146:    case LOGIN_SUCCESS:
147:    case ACCOUNT_SWITCH:
148-      return initialState;

src/mute/muteReducers.js
15-  switch (action.type) {
16:    case LOGOUT:
17:    case ACCOUNT_SWITCH:
18-      return initialState;

src/nav/navReducers.js
60-
61:    case ACCOUNT_SWITCH:
62-      return accountSwitch(state, action);
63-
64:    case LOGIN_SUCCESS:
65-      return loginSuccess(state, action);
--
69-
70:    case LOGOUT:
71-      return logout(state, action);

src/outbox/outboxReducers.js
48-
49:    case ACCOUNT_SWITCH:
50:    case LOGOUT:
51-      return initialState;

src/presence/presenceReducers.js
40-  switch (action.type) {
41:    case LOGOUT:
42:    case LOGIN_SUCCESS:
43:    case ACCOUNT_SWITCH:
44-      return initialState;

src/realm/realmReducers.js
99-  switch (action.type) {
100:    case LOGOUT:
101:    case ACCOUNT_SWITCH:
102-      return initialState;
--
112-
113:    case LOGIN_SUCCESS:
114-      return loginChange(state, action);

src/session/sessionReducers.js
153-    case DEAD_QUEUE:
154:    case ACCOUNT_SWITCH:
155:    case LOGIN_SUCCESS:
156-      return loginSuccess(state, action);

src/streams/streamsReducers.js
56-
57:    case LOGOUT:
58:    case ACCOUNT_SWITCH:
59-      return initialState;

src/subscriptions/subscriptionsReducers.js
69-  switch (action.type) {
70:    case LOGOUT:
71:    case LOGIN_SUCCESS:
72:    case ACCOUNT_SWITCH:
73-      return initialState;

src/topics/topicsReducers.js
45-  switch (action.type) {
46:    case ACCOUNT_SWITCH:
47-      return initialState;

src/typing/typingReducers.js
97-    case DEAD_QUEUE:
98:    case LOGOUT:
99:    case LOGIN_SUCCESS:
100:    case ACCOUNT_SWITCH:
101-      return initialState;

src/unread/unreadHuddlesReducers.js
87-  switch (action.type) {
88:    case LOGOUT:
89:    case ACCOUNT_SWITCH:
90-      return initialState;

src/unread/unreadMentionsReducers.js
73-  switch (action.type) {
74:    case LOGOUT:
75:    case ACCOUNT_SWITCH:
76-      return initialState;

src/unread/unreadPmsReducers.js
74-  switch (action.type) {
75:    case LOGOUT:
76:    case ACCOUNT_SWITCH:
77-      return initialState;

src/unread/unreadStreamsReducers.js
83-  switch (action.type) {
84:    case LOGOUT:
85:    case ACCOUNT_SWITCH:
86-      return initialState;

src/user-groups/userGroupsReducers.js
86-  switch (action.type) {
87:    case LOGOUT:
88:    case LOGIN_SUCCESS:
89:    case ACCOUNT_SWITCH:
90-      return initialState;

src/users/usersReducers.js
25-  switch (action.type) {
26:    case LOGOUT:
27:    case LOGIN_SUCCESS:
28:    case ACCOUNT_SWITCH:
29-      return initialState;
```

gnprice:
Hmm, that output is much clearer with colors than it is here. I recommend running that command yourself. :stuck_out_tongue:

borisyankov:
I think we can go in the other direction actually.
For caching purposes, we should clean up on `LOGIN_SUCCESS` only if the account is different to the one we have cached.
We do a pretty thorough reset on successful initial fetch anyway.

borisyankov:
In any case, we should be very consistent between the reducers, almost all should behave the same.

gnprice:
> For caching purposes, we should clean up on LOGIN_SUCCESS only if the account is different to the one we have cached.

Hmm, sure -- if that's a case that happens regularly in practice. If it's rare or shouldn't happen, then the optimization won't be worth the complexity.

Can it actually happen that we end up dispatching a `LOGIN_SUCCESS` event for the same account we already have data cached for?

---

What about for the other events?
* `LOGOUT` I think you're saying we should not reset state on. We currently do that in a lot of cases -- so that'd mean we should remove those. (With perhaps a few exceptions where there's a clear reason.)
* What about `ACCOUNT_SWITCH`?
* `DEAD_QUEUE` is a little more subtle, and we're working on that one separately.

borisyankov:
> Can it actually happen that we end up dispatching a LOGIN_SUCCESS event for the same account we already have data cached for?

I think the only case is when the user has logged out (state changes to `apiKey: ''`), then logs in to the same account.

borisyankov:
> `LOGOUT` I think you're saying we should not reset state on. 

Yeah, possibly.
The only reason we are doing it right now is kinda for 'security'.

gnprice:
OK, just spent a few minutes looking at this with Boris.
* The three action types `LOGOUT`, `ACCOUNT_SWITCH`, `LOGIN_SUCCESS` don't appear in any sequence -- they correspond to completely independent actions in the UI, namely the "Logout" button; choosing an already-authenticated account in the account picker screen; and authenticating to an account we didn't have a key for.

* The action creators for those three action types are plain-old-object creators; they don't have any logic of their own.

* I think it'd be a useful refactor to consolidate one action type like `CLEAR_ACCOUNT_DATA`.
  * The action creators for these three (in `accountActions.js`) can dispatch that first, then also a `LOGOUT` etc.
  * Almost all the reducers would respond only to the `CLEAR_ACCOUNT_DATA`.
  * There are a couple of exceptions like `accountActions` and `navActions`. Those would be the only ones to continue responding to `LOGOUT` etc.; and they'd stand out better as a result.
  * Then there are further changes we might make to those, but that'll be easier to see after that.

I think that refactor would accomplish the purpose of this PR very well. @jainkuniya , want to try doing that?

jainkuniya:
Cool! Thanks, will do it.

gnprice:
> I think that refactor would accomplish the purpose of this PR very well.

I've filed this task as #4446, and that subsumes this old PR.

PR #3214, opened 2018-12-14
i18n: Add missing strings in `messages_en`.

jainkuniya:
So with the the help of `Transifex`, corresponding tranlated strings
can also be added to the app.


gnprice:
Ah, thanks for catching these!

One small style nit in editing this file: let's add new entries anywhere *but* the very end. JSON doesn't allow trailing commas, so adding a new last element means editing the line before it, making the diff more complex than it needs to be.

gnprice:
And merged -- thanks again!

PR #3215, opened 2018-12-14
Fix issue introduced in b82ef4d.

jainkuniya:
Will update this PR in few hours :)

### DIFF
--- a/src/directSelectors.js
+++ b/src/directSelectors.js
@@ -1,4 +1,5 @@
 /* @flow strict-local */
+import { NULL_ARRAY } from './nullObjects';
 import type {
   GlobalState,
   SessionState,
@@ -61,7 +62,7 @@ export const getFetching = (state: GlobalState): FetchingState => state.fetching
 export const getFlags = (state: GlobalState): FlagsState => state.flags;
 
 export const getReadFlags = (state: GlobalState): { [messageId: number]: boolean } =>
-  state.flags.read;
+  state.flags.read || NULL_ARRAY;
 
 export const getAllNarrows = (state: GlobalState): NarrowsState => state.narrows;
 
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -194,7 +194,7 @@ export const constructMessageActionButtons = ({
     buttons.push('deleteMessage');
   }
   if (!isAnOutboxMessage(message)) {
-    if (message.id in flags.starred) {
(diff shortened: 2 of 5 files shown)


jainkuniya:
Updated!

Should we also include tests?

src/directSelectors.js
```diff
@@ -61,7 +61,7 @@ export const getFetching = (state: GlobalState): FetchingState => state.fetching
 export const getFlags = (state: GlobalState): FlagsState => state.flags;
 
 export const getReadFlags = (state: GlobalState): { [messageId: number]: boolean } =>
-  state.flags.read;
+  state.flags.read || [];
```
borisyankov: || NULL_ARRAY

jainkuniya: Done 👍

gnprice:
Thanks @jainkuniya !

The real fix is actually just that that commit was wrong -- I had just misread things when I wrote it. I just pushed b1c9859af as a fix; see below.

```
commit b1c9859afeea73cc96705eef9ed6fe5774eeb9c1
Date:   Mon Dec 17 01:45:27 2018 -0800

    flags: Revert `initialState` change; make type more specific.
    
    In commit b82ef4d0d "flags: Fix bogus `initialState`" I completely
    misread the FlagsState type; as the bit of test data there shows, I
    thought it was saying it was a map from message IDs to maps from flag
    names to booleans.  In fact it's the transpose of that, and the
    `initialState` value we had was perfectly reasonable.
    
    Revert that change; and make FlagsState explicit about what flags are
    expected, to match.  This gives us type-checking for e.g.
    `flags.starrred`, though any place we do `flags[foo]` sadly just turns
    straight into `any` with current Flow.
    
    Thanks to @jainkuniya for pointing out that commit was broken.
---
 src/chat/__tests__/flagsReducers-test.js | 28 ++++++++++++++++++++++++++--
 src/chat/flagsReducers.js                | 16 +++++++++++++++-
 src/types.js                             | 21 ++++++++++++++++-----
 3 files changed, 57 insertions(+), 8 deletions(-)
```

borisyankov:
These changes pointed me in a direction we can do some improvements... if @jainkuniya is interested in working on them.

We should more selectively provide the flags data, as in `getReadFlags` we should do that for `starred`, `alert_words` and anything else. Then we will prevent renders from generic `flags` changes that are not affecting the actual data we depend upon.

jainkuniya:
yup! @borisyankov got it. Will work on it 👍

PR #3217, opened 2018-12-15
apiTypes: Extract items of `Stream` type to `StreamOptions`.

jainkuniya:
As this list of streamOptions (name, description, inviteOnly,
isAnnouncementOnly, history_public_to_subscribers) will keep on
increase as we make streams more customizable. So it is good to have a
separate type for stream options.

### DIFF
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -182,13 +182,17 @@ export type RealmEmojiState = $ReadOnly<{
 
 export type RealmFilter = [string, string, number];
 
-export type Stream = {|
-  stream_id: number,
+export type StreamOptions = {|
   description: string,
   name: string,
   invite_only: boolean,
   is_announcement_only: boolean,
   history_public_to_subscribers: boolean,
+|};
+
+export type Stream = {|
+  ...$Exact<StreamOptions>,
+  stream_id: number,
 |};
 
 export type Subscription = {|


jainkuniya:
(part of #3140)

borisyankov:
Hmmm, I don't think I clearly understand your reasoning. 
And if I am understanding it, not sure I agree with it :) 
Clarify & elaborate.

jainkuniya:
@borisyankov 
I guess better reasoning can be found [here](https://github.com/zulip/zulip-mobile/pull/3140#issuecomment-438875187) (by @gnprice) :).

gnprice:
Thanks @jainkuniya ! The quoted suggestion was:

> * This list of fields `name`, `description`, `inviteOnly`, and now `isAnnouncementOnly` appears in a lot of places, as your commit "createStream: Add functionality to ..." (btw "to to" -> "to") demonstrates. Also a lot of those places are positional arguments, and increasingly long lists of positional arguments make me nervous.

So the spots I was concerned about were like this one in that linked PR:
```
    dispatch(createNewStream(name, description, [ownEmail], isPrivate, isAnnouncementOnly));
```

and I'd like to group a lot of those positional arguments into an object, so they can have names.

In this spot, on the other hand, they're already part of an object and so have names attached already.

PR #3219, opened 2018-12-18
Make VSCode recommend extensions we recommend

borisyankov:
We do recommend seveal VS Code extensions in our documentation:
https://github.com/zulip/zulip-mobile/blob/32795a8ba71e290efabc1ef36a27e654d3daf3a1/docs/howto/editor.md

It turns out VS Code supports an automatic recommendation of these
extensions if we specify them in `.vscode/extensions.json` file.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+	"recommendations": [
+		"vsmobile.vscode-react-native",
+		"flowtype.flow-for-vscode",
+		"dbaeumer.vscode-eslint",
+		"esbenp.prettier-vscode"	
+	]
+}


ishammahajan:
Why was this closed? Seems like something good to implement (and easy enough), and will help new contributors 🙂.

gnprice:
> Why was this closed?

I think mainly because it'd been idle for a number of months.

I agree, this would be helpful to have! Would you like to pick it up?

Rereading, I think the main things I'd want before merging it are
* Link to those docs somewhere (this part is easy)
* Update the discussion in docs/howto/editor.md in light of the new logic. This hopefully means making it simpler :slightly_smiling_face: . This is a writing task.

Could start from this version (you can use `../zulip/tools/reset-to-pull-request 3219` to fetch it), and add a commit on top for the docs change.

ishammahajan:
> Would you like to pick it up?

I would! Thanks, pushed as #3510 😄

Issue #3205, opened 2018-12-10
Add ultra basic support for the SubMessage-based polls feature 


timabbott:
Zulip has an experimental "polls" feature, which uses the special SubMessage table in the database for its storage (https://zulip.readthedocs.io/en/latest/subsystems/widgets.html#poll-todo-lists-and-games).  You can create one on web by sending a message with `/poll` as the content.  The UI is currently pretty janky; we have a separate issue on web (https://github.com/zulip/zulip/issues/11010) for improving that.

The polls feature has been getting a decent amount of interest from folks who want it in Zulip Cloud, so we need to take steps to make them display in a not-totally broken fashion on mobile.

Structurally, the `widgets`  / `submessage` feature has some important properties for mobile: The API only sends and receives JSON, via `submessage` events (or the `.submessage` field on a message object fetched from the server).  The client is completely responsible (and empowered) to render the data as it pleases, as well as enabling UI. 

In terms of staging, I think there's a few levels of mobile-side support for this widget (or any other widget for that matter, though the others are more prototypes and don't have a lot of demand so we shouldn't worry about them).  

* First, mobile can literally look for messages with a `poll` submessage/widget entry as the most immediate hack to detect and say "This message is a poll, which is not supported on mobile yet.". 
* From there, it can process the data from those events properly, and do a basically textual rendering of the poll results, so you can at least understand the message.
* And then from there we could work on rendering the results and turning on UI to actually participate in the poll from your device.

borisyankov:
@timabbott To clarify some aspects that seem to be true:
 * we can differentiate between normal messages and widgets by `submessages.length > 0` ? 
 * `msg_type` of a submessage can be only `widget` currently?
 * submessages can only be added to, no update/delete?
 * `content` is always intended to be JSON string?

timabbott:
> we can differentiate between normal messages and widgets by submessages.length > 0 ?

yes.

> `msg_type` of a submessage can be only widget currently?

yes; the client can ignore any other submessage until if/when we change that.

> submessages can only be added to, no update/delete?

currently, yes, though that could change in the future.

> `content` is always intended to be JSON string?

yes.

timabbott:
@borisyankov @gnprice how close are we to merging this?  We've been getting complaints from users about the frankly poor situation that polls are invisible in the mobile apps.

borisyankov:
@timabbott we have the #3220 ready to merge.
It does support the processing, understanding and storage of submessages.

[...]

PR #3220, opened 2018-12-18
Add basic support for submessages/widgets

borisyankov:
These steps add tracking of the sub-message data correctly and showing a placeholder that the widgets are not yet supported.

Implements the basics of #3205


gnprice:
Thanks @borisyankov ! This looks good. Several small points:
* Let's put the jsdoc directly above what it documents, with no blank line. Helps prevent getting separated later.
* The detailed jsdoc on the SubMessage type is super helpful -- thanks! :smile: 
* In this in a commit message:
   > Currently the only opperation is addition, we do not update or delete submessages.
  I'm not sure if the "we" is just the mobile app, so far, or is Zulip as a whole. Good to be explicit.
* This test should be redundant according to the type:
  `message.submessages && message.submessages.length > 0`
  Oh hmm but I guess `message` could be an `Outbox` and not `Message`. ... Maybe add `submessages: Array<empty>` to `Outbox`? That's the type of arrays whose elements belong to the type with no values... in other words, the type consisting only of the empty array, because there are no possible values to be its elements. And add a `submessages: []` to where we make outbox objects.
* I'd squash the widget CSS commit into the widget rendering, since the idea is that the plain text doesn't look right.

borisyankov:
> This test should be redundant according to the type:
message.submessages && message.submessages.length > 0

This is not only for `Outbox` but also to support older versions of the back-end which do not have the `submessages` property.

gnprice:
Thanks! Changes all look good.

> > This test should be redundant according to the type:
> > `message.submessages && message.submessages.length > 0`
> 
> This is not only for `Outbox` but also to support older versions of the back-end which do not have the `submessages` property.

Hmm, I see.

This is similar to the situation I just made a comment on in #2934 , then. I'd like this sort of variability to be explicit in the types; basically if it's real enough that we need code to handle it, then it's real enough to put in the types. The key factor in how much value the types give us is how much we can believe them, so wherever we see a reason not to believe them we should fix that.

An industrial-strength version of this principle would mean we should be systematically validating the API responses against the types we expect. That'd be pretty useful too, but probably also a lot of work to get going; for now I'm happy with assuming the server is some reasonably upstream-like version of Zulip.

Here, a `?:` type in the API, with a line of jsdoc saying what Zulip server version introduced it, should be good enough.

borisyankov:
I did add `submessages?: $ReadOnlyArray<SubMessage>` and `stream_id?: number,` but then that requires weird `if`s to check for null values in cases where we are absolutely certain there will not be.

For example:
`/messagesReducers.js`
`     110│       submessages: [...message.submessages, submessage],`

This one now needs a check, though this code will never execute on a server that does not have support for submessages.

Two places that we need to add weird code for `stream_id` too, that really don't need it.

I suspect we have many places in our code that can become a maybe type if we are to reflect the truth in the differences between server versions objects. That seems to be out of scope and capability of a type system, at least if we don't version the types themselves.

gnprice:
Writing down some quick notes here from an audio conversation with @borisyankov today. Part of this is also copied from a [chat conversation](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Queue.20.26.20race.20conditions/near/680549) a few days ago.

> I did add `submessages?: $ReadOnlyArray<SubMessage>` and `stream_id?: number,` but then that requires weird `if`s to check for null values in cases where we are absolutely certain there will not be.
> 
> For example:
> `/messagesReducers.js`
> ` 110│ submessages: [...message.submessages, submessage],`
> 
> This one now needs a check, though this code will never execute on a server that does not have support for submessages.

The general pattern I'd like to use for this is:

* **Do the conversions at the edge, and have a consistent set of types on the inside.**

In this instance, what that translates to looks like:
* Let's separate the `Message` types into two types, quite similar to each other but with an important difference:
  * `ServerMessage` models the full complexity (the lack of valid assumptions) of what the server can send -- `submessages` is optional, etc.
  * `Message` is the type of a nice cleaned-up object where the variation between server versions, between different routes that aren't quite consistent, etc., has been canonicalized away.
  * There's a bit of code, in one central place, that converts the chaotic `ServerMessage` form to the canonicalized `Message` form.
* I think both types belong inside the API bindings `src/api/`; basically they are part of its interface on two different sides, server-facing and client-application-code-facing, respectively. So the normalizing/canonicalizing code does too.
* When that code (`canonicalizeMessage`?) sees there's no `submessages` property:
  * It adds `submessages: []`.
  * This isn't just a placeholder or "null" value, representing "we don't know if there are submessages". If it were, that'd be a problem.
  * Rather, it really is saying "there are no submessages" -- because that is the *truth*. The server evidently doesn't do submessages, because it returned a message with no `submessages` property.
  * And now the whole rest of the code doesn't have to deal with that variation. There's always a `submessages` on the internal `Message` type; when it's `[]`, that always means "there are no submessages"; and nothing needs to remember or care that the way we learned that is that the server didn't mention `submessages` at all.

---

Separate from this PR's work but as nice followup opportunities:

There are a number of other cleanups we can make on the app-facing `Message` type once we separate it from `ServerMessage`; `flags`, `sender_domain`, `isOutbox`. They haven't been important enough on their own to do the work of that separation, but once that's done they'll be easy and good wins.

A less simple cleanup but that will be quite helpful is to all the properties that differ between stream messages and PMs; basically we get to make up how to handle that on the internal `Message`, whatever we think best, and then have `canonicalizeMessage` convert to that and all the app code get to enjoy it. For example, we might make a single `recipient` property which is a union of two object types, and shove all the stream-vs-PM-dependent properties inside that: like `{| type: 'stream', stream_id: number, /* ... */ |} | {| type: 'private', users: PmRecipientUser[] |}`. That'd let us untangle all the biggest messes in the `Message` type, like `display_recipient`.

---

As those optional followups illustrate, there's also a short-term-easier solution which is perfectly reasonable as a way of getting this feature shipped:
* Keep a single `Message` type for now, which basically looks like `ServerMessage` would.
* I.e. it has `submessages?: $ReadOnlyArray<SubMessage>` -- and jsdoc explaining what's known about when that's present and absent.
* Now the application code that cares about submessages is exposed to the complexity of the fact that the property might or might not be present -- and it has to decide how to handle its absence, i.e. do the job that `canonicalizeMessage` would do. In the case of `submessages`, it should treat that exactly as if it were present and empty.
* If there are only one or two such places, this isn't a high cost. That's what we've done so far with `stream_id` and others.

This simpler solution is actually what I had in mind in my previous comment here. I hope this description of the more fully-worked-out set of types we can move to in the future is helpful even if we do take the simpler solution for now, though.

timabbott:
Just a quick note: We just changed the poll widget API internals a bit in zulip/zulip#11375 as part of making it more reasonable before we turn it on outside chat.zulip.org (which is happening essentially now, though we'll only advertise it to the handful of orgs that were really asking for it).  Looking at this PR, it looks like you weren't interacting with the element that changed yet (`new_comment` becoming `new_option`) anyway, so this is mostly just a note that you should look at master when doing further work on this.

borisyankov:
I did revisit the PR. Each commit had suffered a bit from significant merge conflicts from the several refactors (types, reducers, moving of content from a file to file).

Also, to explicitly acknowledge that I agree with the conclusions of the discussion we had, and I also agree that these bigger changes can wait.

I have implemented the changes in the 'easier solution':
* documented the `submessage` value
* made it optional and made the code deal with non-existent value for it

I consider it ready for merging.

borisyankov:
Ping. This is ready for review & merge.

gnprice:
Thanks @borisyankov !

Made some changes, and now merging.

* Although the server database model is `SubMessage`, everything else in the server and docs, and everything in the API, seem to have it as one word, "submessage". This branch mostly went for two words "sub message", as in `...SubMessage` and `...SUB_MESSAGE`; I've changed it to "submessage" to match the API.

* Like in #3263 and in particular 2c206a919, I'd like us to document server event types explicitly, in the new `eventTypes.js`; and then use those as near-verbatim as possible in the Redux actions we make from them. I wrote this as a new commit.

* In the comment on `Message.submessages`, added the crucial point about what it means when it's absent: it's equivalent to being empty.

* Some small adjustments to the reducer code, explained in new commits making them.

* The CSS had duplicate `display: flex;` lines; deleted one.

* The CSS was added at the very end of our CSS, which isn't a logical place for it; other CSS code there is for the `#scroll-bottom` button, the error banners, and other things that live outside the messages themselves. This is CSS for a particular type of message, so I moved it to be next to the other CSS for messages.

* I made a pass over the "sorry, unsupported" message, making it hopefully a bit more helpful.

PR #3223, opened 2018-12-18
MainScreenWithTabs: Render CompatibilityScreen based on server_settings.

jackrzhang:
This is a continuation of #3158.

When a user switches accounts or adds an account for the first time, a request to `api/v1/server_settings` is made, specifically for the response's `is_incompatible` boolean. If the app is incompatible with the specific server, then `CompatibilityScreen` is rendered, along with `SwitchAccountButton`.

Here's a screenshot from the iOS simulator:

<p align="center">

<p/>

Here's a screen recording from an AVD with Google Play:

<p align="center">
android-compatibility-screen-global-check-with-badge
<p/>


src/account-info/SwitchAccountButton.js
```diff
@@ -44,8 +49,20 @@ class SwitchAccountButton extends PureComponent<Props> {
   };
 
   render() {
+    const { isOnCompatibilityScreen } = this.props;
+
+    const buttonStyle = [styles.button];
+    if (isOnCompatibilityScreen) {
+      buttonStyle.push(styles.whiteBorder);
```
borisyankov: You can combine like this:
`[styles.button, isOnCompatibilityScreen && styles.whiteBorder]`

RN ignores `null`s and `false`s.

Issue #3222, opened 2018-12-18
refactor updateExistingStream: Reduce api calls on changing stream details.


jainkuniya:
Currently if user changes stream name and description, then we are making two API calls. On the other hand `streams` API endpoint is generic and can handle multiple properties at the same time. So redactor `updateExistingStream` method, so that in such cases only one request is send to server.

https://chat.zulip.org/#narrow/stream/3-backend/topic/api.20.2Fstreams.2F.24.7Bid.7D

@zulipbot claim

gnprice:
This was fixed by be12766bc2451afda2ff99b26b219b617c9055be / #5341 .

PR #3224, opened 2018-12-18
updateExistingStream: Reduce api calls on changing stream details.

jainkuniya:
Before inidividual api calls were made for each stream property which
is changed. But on server, this api is capable of handeling multiple
propery at a time. So just call api once with all the properties which
are changed.

Fixes: #3222

### DIFF
--- a/src/api/streams/updateStream.js
+++ b/src/api/streams/updateStream.js
@@ -2,7 +2,8 @@
 import type { ApiResponse, Auth } from '../apiTypes';
 import { apiPatch } from '../apiFetch';
 
-export default (auth: Auth, id: number, property: string, value: string): Promise<ApiResponse> =>
-  apiPatch(auth, `streams/${id}`, {
-    [property]: value,
-  });
+export default (
+  auth: Auth,
+  id: number,
+  newValues: { [string]: string | boolean | number },
+): Promise<ApiResponse> => apiPatch(auth, `streams/${id}`, { ...newValues });
--- a/src/streams/streamsActions.js
+++ b/src/streams/streamsActions.js
@@ -3,6 +3,7 @@ import type { GetState, Dispatch, Stream, InitStreamsAction } from '../types';
 import { createStream, updateStream, toggleMuteStream, togglePinStream } from '../api';
 import { INIT_STREAMS } from '../actionConstants';
 import { getAuth } from '../selectors';
+import { getChangedValues } from '../utils/misc';
 
 export const initStreams = (streams: Stream[]): InitStreamsAction => ({
   type: INIT_STREAMS,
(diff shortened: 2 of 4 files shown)


src/api/streams/updateStream.js
```diff
@@ -2,7 +2,11 @@
...
-    [property]: value,
-  });
+export type ParamsType = {
+  new_name?: string,
+  description?: string,
+  is_private?: boolean,
+};
+
+export default (auth: Auth, id: number, newValues: ParamsType): Promise<ApiResponse> =>
+  apiPatch(auth, `streams/${id}`, res => res, { ...newValues });
```
borisyankov: Do you need to spread the values?
```{ ...newValues }`` vs just `newValues`

File: src/api/streams/updateStream.js [hunk 2]
```diff
@@ -2,7 +2,11 @@
...
 import { apiPatch } from '../apiFetch';
 
-export default (auth: Auth, id: number, property: string, value: string): Promise<ApiResponse> =>
-  apiPatch(auth, `streams/${id}`, res => res, {
-    [property]: value,
-  });
+export type ParamsType = {
+  new_name?: string,
+  description?: string,
+  is_private?: boolean,
```
borisyankov: Is this an exhaustive list?

jainkuniya: yup! stream have limited properties.

src/api/streams/updateStream.js
jainkuniya: I tried with `newValues `, then flow is giving error.
````
Cannot call apiPatch with newValues bound to params because:
 • string [1] is incompatible with boolean [2] in property description.
 • string [1] is incompatible with number [3] in property description.
 • boolean [4] is incompatible with string [5] in property is_private.
 • boolean [4] is incompatible with number [3] in property is_private.
 • string [6] is incompatible with boolean [2] in property new_name.
 • string [6] is incompatible with number [3] in property new_name.

           src/api/streams/updateStream.js
       [6]  6│   new_name?: string,
       [1]  7│   description?: string,
       [4]  8│   is_private?: boolean,
            9│ };
           10│
           11│ export default (auth: Auth, id: number, newValues: ParamsType): Promise<ApiResponse> =>
           12│   apiPatch(auth, `streams/${id}`, res => res, newValues);
           13│

           src/utils/url.js
 [5][2][3] 16│ export type UrlParams = { [string]: string | boolean | number };



Found 6 errors
```


In `toggleMobilePushSettings` also we are creating new object.
```
export default async ({
  auth,
  opp,
  value,
}: {
  auth: Auth,
  opp: string,
  value: boolean,
}): Promise<ApiResponse> =>
  apiPatch(auth, 'settings/notifications', res => res, {
    ...getRequestBody(opp, value),
  });
```
So I thought of doing same 😅

src/streams/streamsActions.js
```diff
@@ -26,21 +27,20 @@ export const updateExistingStream = (
   initialValues: Object,
   newValues: Object,
 ) => async (dispatch: Dispatch, getState: GetState) => {
+  const apiParams: UpdateStreamApiParamsType = {};
```
borisyankov: This looks like a candidate for a utility function.
The chances are that we will add more fields and we will need this logic in other places.
What if you do create a `currentValues` object with the values needed. 
Then we can diff it against the current version (via a utility function)

src/api/streams/updateStream.js
borisyankov: `ParamsType` does not sound like concrete or a good name.
First, we don't name types `*Type` :)
Also, `Param`? It is a parameter/argument to a function but you can do better.

src/streams/streamsActions.js
jainkuniya: Then we need to do some refactors, like
* currently we have `name` key (for stream name) in `initialValues` and `newValues`
But at the time of sending parameters to API, we want key `new_name`, so again a if/else will be required.
Suggestions?

src/api/streams/updateStream.js
jainkuniya: any name you want to suggest? :)
`ApiBody`, `ApiParameters`??

src/streams/streamsActions.js
borisyankov: I would suggest you do an object diffing function and then pass the resulting object (if it is not empty!) to this new function:
https://github.com/zulip/zulip-mobile/blob/master/src/api/apiFetch.js#L14

gnprice:
We ended up making this change in be12766bc / #5341, with a slightly more explicit implementation.

Issue #3225, opened 2018-12-18
Add mobile support for AzureAD authentication integration


timabbott:
I think it should be able to use the exact same flow we use for GitHub authentication, since on the backend it's just 2 lines of code inheriting from the same subclass.

This was added in the backend via https://github.com/zulip/zulip/pull/10618.  Eventually, we'll want to rethink the interface through which we communicate about which social auth backends are supported (since they'll all work the same), but it's probably best to not block on that until adding new Social auth backends is a common occurence.

@sschritt and @Weihnachtshase should be able to help with testing, since they use this auth backend.

PR #3227, opened 2018-12-20
Add Azure AD authentication

sschritt:
Authentication with Microsoft AzureAD SSO (see zulip/zulip#10618)

Fixes: #3225

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -36,6 +36,7 @@ export const IconPrivateChat: IconType = props => <Feather name="mail" {...props
 export const IconDownArrow: IconType = props => <Feather name="chevron-down" {...props} />;
 export const IconGoogle: IconType = props => <IoniconsIcon name="logo-google" {...props} />;
 export const IconGitHub: IconType = props => <Feather name="github" {...props} />;
+export const IconWindows: IconType = props => <IoniconsIcon name="logo-windows" {...props} />;
 export const IconCross: IconType = props => <Feather name="x" {...props} />;
 export const IconDiagnostics: IconType = props => <Feather name="activity" {...props} />;
 export const IconNotifications: IconType = props => <Feather name="bell" {...props} />;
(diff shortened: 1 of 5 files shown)


src/common/Icons.js
```diff
@@ -20,6 +20,7 @@ export const IconPrivateChat = (props: Object) => <Feather name="mail" {...props
 export const IconDownArrow = (props: Object) => <Feather name="chevron-down" {...props} />;
 export const IconGoogle = (props: Object) => <IoniconsIcon name="logo-google" {...props} />;
 export const IconGitHub = (props: Object) => <Feather name="github" {...props} />;
+export const IconAzureAD = (props: Object) => <IoniconsIcon name="logo-windows" {...props} />;
```
borisyankov: Maybe call it `IconWindows` even though we use it for Azure.
The logo for Azure is different, but I think Windows' logo is a reasonable replacement.

borisyankov:
@sschritt the PR looks good. Would you resolve the merge conflicts?
I think we want to merge it ASAP as discussed here https://chat.zulip.org/#narrow/stream/48-mobile/topic/Azure.20SSO

We don't currently have a server with that authentication set up to test on but I have high confidence the code is correct and should work.
We can merge it as-is.

borisyankov:
@gnprice I did add a P1 to make the PR more visible.
It is solving a problem a customer is encountering.

gnprice:
@borisyankov I just went and did the rebase -- the conflicts were straightforward to resolve. Also squashed the name change into the main commit.

gnprice:
Two things that need to be fixed:
* A test is broken.
* The `AuthenticationMethods` type should be updated to include the new case we're acting on.

Those should be easy, so I'll just fix them and merge.

gnprice:
Merged! @sschritt , please try it out and confirm everything works for you.

Vilinthril:
Just to clarify, this has not been promulgated to the Google Play Store yet, right?

Issue #3228, opened 2018-12-20
account screen layout issue: Component under `ComponentList` are not rendered properly.


jainkuniya:
![image](https://user-images.githubusercontent.com/18511177/50283369-b7a45700-047b-11e9-83b2-9c2d3e07a6de.png)

gnprice:
Huh fascinating -- thanks for catching and diagnosing this!

I'm curious what broke it. Do you know anything about what caused this to break? You use the term "regression" in your PR, which sounds like saying that you do.

gnprice:
Also I get different behavior, even on the parent commit of your PR:
![image](https://user-images.githubusercontent.com/28173/50469516-9d79e580-0961-11e9-9097-71c7d0f52ccb.png)

Still broken layout, but the information is all there; just not spaced appropriately, and the "Send private message" button is missing. That's on both Android 8 and 9, on an emulated Pixel 2 and a real Pixel 2 XL respectively. So whatever's causing this is at least a little subtle.

PR #3229, opened 2018-12-20
accountDetails: Fix layout regression.

jainkuniya:
We are using `ImageBackground`, which uses styles such as `position:
absolute`. Because it is capable of displaying text (/componenets)
over it. So styles such as `alignItems: center, flex: 1` were
resulting items to be present at vertically center of screen instead
of space below image.

By removing this styles and only keeping `justifyContent: center`,
works as expected.

Fix: #3228

image


gnprice:
Thanks @jainkuniya ! Merged.

Issue #2611, opened 2018-06-02
settings: Build a better UI for choosing language


gnprice:
This will be a bit complex to build, and I think right now we have a number of equally complex areas of UI work to do that will be higher-value. (Most users will only ever touch this screen either zero or one times.) But I'm getting down some thoughts now, since I was looking into this thanks to #2610.

Our screen for choosing the UI language looks like this:
![image](https://user-images.githubusercontent.com/28173/40869056-72e1a48e-65c9-11e8-92f3-d4beccbbc32b.png)
Apart from some small layout issues (those are #2610) which shouldn't be hard to fix, there are a few problems with this:
* It uses national (more or less) flags to represent languages. This is confused about the world (many languages are spoken widely in several countries, and many countries have several primary languages), and isn't super helpful anyway.
* It names all the languages in English only.
* It's a long list, in an unavoidably arbitrary order (it can't be in the order of the user's own language, because the point of the screen is we don't know that yet), with no search.

Good language selectors I see in mobile apps:
* don't use flags
* name each language in the current UI language, and in itself
* provide a search box -- which searches through both those names for each language

The Wikipedia app (the official one from Wikimedia) does an excellent job of this. It's actually significantly different in details between their Android and iOS apps, but both are good models.

anh2111:
@zulipbot claim

Hi, I am new to Zulip repo and looking for an issue to get started. This one looks doable. I'll follow Wikipedia's language screen on Android and make a WIP PR soon.

anh2111:
The PR is ready for review.

gnprice:
Now done, with #3231. Thanks again @anh2111 !

PR #3231, opened 2018-12-21
[WIP] Issue #2611. settings: Build a better UI for choosing language

anh2111:
Issue #2261. WIP: removed flags, named each language in English and in itself, added a search box. I'm learning the translation part to show language names in the current UI language.

Native language names are taken from https://github.com/nohamelin/simple-locale-switcher/wiki/Language-Names-for-Locale-Codes

Doing my first PR to the repo, feedback is welcome.


anh2111:
Sure :)

image

src/settings/LanguagePicker.js
```diff
@@ -15,23 +15,53 @@ const styles = StyleSheet.create({
...
 |};
 
-export default class LanguagePicker extends PureComponent<Props> {
+type State = {|
+  data: Array<*>,
+|};
+
+export default class LanguagePicker extends PureComponent<Props, State> {
+  state = {
+    data: languages,
```
borisyankov: You don't need that state.

src/settings/LanguagePicker.js
```diff
@@ -15,23 +15,53 @@ const styles = StyleSheet.create({
...
+  };
+
+  handleFilterChange = (text: string) => {
+    const newData = languages.filter(item => {
```
borisyankov: Run this code in `render()` and you can remove `componentDidUpdate()`

src/settings/LanguagePicker.js
```diff
@@ -15,23 +15,53 @@ const styles = StyleSheet.create({
...
+      const itemData = `${item.name.toUpperCase()} ${item.nativeName.toUpperCase()}`;
+      const textData = text.toUpperCase();
+      return itemData.indexOf(textData) > -1;
```
borisyankov: Use `string.prototype.includes` to make this a little bit clear.

src/settings/LanguagePicker.js
anh2111: Updated, thanks for pointing it out!

src/settings/LanguagePicker.js
anh2111: Updated!

src/settings/LanguagePicker.js
anh2111: Updated!

src/settings/LanguagePicker.js
```diff
@@ -15,26 +19,57 @@ const styles = StyleSheet.create({
...
+            name={item.name}
+            nativeName={item.nativeName}
           />
         )}
       />
     );
   }
 }
+
+export default connect()(withGetText(LanguagePicker));
```
borisyankov: You don't need `connect` to use `withGetText`.
We use it to extract data out of the Redux store and/or to get the `dispatch` function.

gnprice: Hmm interesting. I wrote `withGetText`, recently; I agree with this:

> Logically, `withGetText` will inject prop `_` to `LanguagePicker`, so there is no need to pass `_` here.

and I think I wrote its Flow type wrong. :grin: 

I'll look at that and try to fix it, but in the meantime there's an alternate way to use our translation API that might be a little nicer here anyway. We don't have any examples of this yet because the API is quite new, but here's what it looks like at the tip of this PR branch:
```
@@ -4,7 +4,7 @@ import { FlatList, StyleSheet, View } from 'react-native';
 import { connect } from 'react-redux';
 
 import type { GetText } from '../types';
-import { withGetText } from '../boot/TranslationProvider';
+import { withGetText, TranslationContext } from '../boot/TranslationProvider';
 
 import languages from './languages';
 import LanguagePickerItem from './LanguagePickerItem';
@@ -20,13 +20,16 @@ type Props = {|
   value: string,
   onValueChange: (locale: string) => void,
   filter: string,
-  _: GetText,
 |};
 
-class LanguagePicker extends PureComponent<Props> {
+export default class LanguagePicker extends PureComponent<Props> {
+  context: GetText;
+
   getTranslatedLanguages = () =>
     languages.map(language => {
-      const translatedName = this.props._(language.name);
+      const _ = this.context;
+      const translatedName = _(language.name);
       return {
         ...language,
         name: translatedName,
@@ -71,5 +74,3 @@ class LanguagePicker extends PureComponent<Props> {
     );
   }
 }
-
-export default connect()(withGetText(LanguagePicker));
```

Try that way?

gnprice: One quick important workflow point, on this topic: when you fix something, use `git rebase -i` to squash it into the commit that it's fixing. That's part of how we make [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html).

gnprice:
Thanks @anh2111, and welcome!

This generally looks good -- looking forward to merging it! A few comments (in addition to the thread I replied to above):
* Let's also remove the `flag` lines from the data file, now that we're no longer using. This can be a separate commit.
* That link for your source for the native names of languages is very helpful. Let's put that in the commit message for that commit.
* I'm a little dubious of a couple of items there, like lower-case "magyar". Let's use Wikipedia (the UI, not the articles) as the gold standard for the appropriate way to identify a language -- that's a project that for every language has speakers who care about making sure it's right, and are empowered to make it right.
  * For Hungarian, the Wikipedia website and mobile app both use "Magyar" in a language-selection UI like this, so let's follow them. The [Wikipedia article](https://en.wikipedia.org/wiki/Hungarian_language) suggests it's normally written in lowercase, so I think that reflects the difference between prose and title text.
  * For Hindi, I think the text you have says "Hindi (India)", as in the locale `hi_IN`. We just want "Hindi"; again consult Wikipedia.
  * Similar story for Portuguese.
* Just to highlight here in the main thread something I mentioned above: please use `git rebase -i` to squash fixes into the commits they fix. See Zulip's [Git style guide](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) for more on how we make our commits individually clear and coherent.

src/settings/LanguagePicker.js
anh2111: Thanks for the suggestion, that works perfectly :) 

As a side note, I also tried following:
```
/**
 * Usually called `_`, and invoked like `_('Message')` -> `'Nachricht'`.
 *
 * Use `context: TranslationContext` in a React component; then in methods,
 * say `const _ = this.context`.
 *
 * Alternatively, for when `context` is already in use, use `withGetText`
 * and then say `const { _ } = this.props`.
 *
 * @prop intl - The full react-intl API, for more complex situations.
 */
export type GetText = {
  (string): string,
  intl: IntlShape,
};
```
And it doesn't work. I think the comment may be misleading here.

src/i18n/translations/messages_en.json
```diff
@@ -137,5 +137,26 @@
   "Topics": "Topics",
   "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n": "Welcome to Zulip!\n\nYou'll need to first create an account from your computer. If you're not sure where to start, go to zulipchat.com from your web browser.\n\nHope to see you back here soon!\n",
   "Add subscribers": "Add subscribers",
-  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
+  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream.",
+  "English": "English",
```
borisyankov: I am not sure we need to add the language names to be translated into all languages.
Do you think we should remove these?

src/settings/LanguagePicker.js
anh2111: Also, I've used `rebase` to squash commits. Thanks for suggesting!

src/i18n/translations/messages_en.json
borisyankov: Sure. But then, maybe we don't need to have every language translated into every language? 
Even right now, this is 20 ^ 2 = 400 total translations. 
The idea of this work is that the user would recognize the language name in their own language.

anh2111: I see. If so, we may need to edit the requirements described in issue #2611. 

@gnprice your thoughts?

src/settings/LanguagePicker.js
gnprice: Yeah, indeed that comment is mistaken; I'll fix it to match the instructions I gave above (and that you've just validated actually work :slightly_smiling_face: ). Thanks!

gnprice:
Thanks @anh2111 -- this is getting close!

* This is a bit confusing; I think the style is for the not-native name?
```
+            <RawLabel text={nativeName} />
+            <RawLabel text={name} style={componentStyles.nativeName} />
```

* Small tweak: the Wikipedia UI here makes the current-UI-language name for each language a bit smaller as well as lighter-weight than the self-name, and that seems to work well. Perhaps try that?

* I still see a commit that adds state, and then a later commit that takes it out again. Happy to help with how to use `git rebase -i` to squash changes together so as to clean this up; perhaps in chat over on chat.zulip.org?

* Similarly, the fixes to names should be squashed into the commit that adds them in the first place (so there's no commit adding the bogus versions). And the explanation of the data source should be in the commit message of the commit that adds the data.

gnprice:
BTW I just pushed f932a1da5 and 266640e28 to fix that jsdoc and the `withGetText` type signature, respectively. Thanks again for pointing those out!

Issue #3063, opened 2018-10-19
design: The same 'home' icon is used for the 'All messages' screen and the 'Unread' screen.


jackrzhang:
I think higher-level changes to how navigation currently works would be beneficial; at the moment though, it's pretty confusing just to see the same icon twice on the 'Unread' screen.

The web app uses a 'home' icon for 'All messages', so I think a quick improvement could be to use a different icon for the 'Unread' screen on the bottom navigation bar, even if it doesn't represent the concept of a sidebar or 'Unread' screen super accurately.

<p align="center">'Home' screen:<p/>

<p align="center">
<img src="https://user-images.githubusercontent.com/12771126/47196422-43332780-d315-11e8-83cc-32404caaf25d.png" width="350" />
<p/>

<p align="center">'All messages' screen:<p/>

<p align="center">
<img src="https://user-images.githubusercontent.com/12771126/47196547-f439c200-d315-11e8-9b8d-350425a968b8.png" width="350" />
<p/>

<p align="center">Web app:<p/>

<p align="center">
<img src="https://user-images.githubusercontent.com/12771126/47196513-bc327f00-d315-11e8-9e7e-0b8c7d707328.png" width="550" />
<p/>

borisyankov:
I agree that changing the 'Unread' icon is better. Any ideas what icon would be appropriate?

I decided to use 'Home' for that tab initially as it was supposed to have other key info in the screen, though it still doesn't ;)

jackrzhang:
Maybe the [comments icon](https://fontawesome.com/icons/comments?style=regular) could be appropriate here?

![image](https://user-images.githubusercontent.com/12771126/48460801-84d4b800-e785-11e8-96aa-bf155b3a4143.png)

@gnprice Any thoughts?

rishig:
I like the idea of the unreads screen being the "Home" screen, and de-emphasizing "All messages". 
And we can replace the icon on the webapp as well; the reason it's a Home icon is for historical reasons. 

Maybe some options:

The top half of
![image](https://user-images.githubusercontent.com/890911/48462215-2f4fd980-e78c-11e8-8da0-f48202c79c7e.png)

![image](https://user-images.githubusercontent.com/890911/48462311-805fcd80-e78c-11e8-905d-fef7fe423969.png)

![image](https://user-images.githubusercontent.com/890911/48462328-95d4f780-e78c-11e8-8807-ce0ce01ab819.png)

[...]

PR #3232, opened 2018-12-21
Stop using the same icon for two different buttons on the main screen.

jackrzhang:
This is a proposal for a quick resolution to #3063, discussed with and suggested by @gnprice.

Before:

After:

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -20,7 +20,7 @@ export const Icon: ComponentType<{| ...IconProps, name: string |}> = Feather;
 
 export type IconType = ComponentType<IconProps>;
 
-export const IconHome: IconType = props => <Feather name="home" {...props} />;
+export const IconInbox: IconType = props => <Feather name="inbox" {...props} />;
 export const IconStar: IconType = props => <Feather name="star" {...props} />;
 export const IconMention: IconType = props => <Feather name="at-sign" {...props} />;
 export const IconSearch: IconType = props => <Feather name="search" {...props} />;
--- a/src/main/HomeTab.js
+++ b/src/main/HomeTab.js
@@ -33,7 +33,7 @@ class HomeTab extends PureComponent<Props> {
       <View style={styles.wrapper}>
         <View style={styles.iconList}>
           <NavButton
-            name="home"
+            name="globe"
             onPress={() => {
               dispatch(doNarrow(HOME_NARROW));
             }}
--- a/src/main/MainTabs.js
+++ b/src/main/MainTabs.js
@@ -9,7 +9,7 @@ import HomeTab from './HomeTab';
 import StreamTabs from './StreamTabs';
 import PmConversationsCard from '../pm-conversations/PmConversationsCard';
 import SettingsCard from '../settings/SettingsCard';
-import { IconHome, IconSettings, IconStream } from '../common/Icons';
+import { IconInbox, IconSettings, IconStream } from '../common/Icons';
 import { OwnAvatar } from '../common';
 import IconUnreadConversations from '../nav/IconUnreadConversations';
 import ProfileCard from '../account-info/ProfileCard';
@@ -21,7 +21,7 @@ export default TabNavigator(
       navigationOptions: {
         tabBarLabel: 'Home',
         tabBarIcon: (props: TabNavigationOptionsPropsType) => (
-          <IconHome size={24} color={props.tintColor} />
+          <IconInbox size={24} color={props.tintColor} />
         ),
       },
     },
--- a/src/title/TitleSpecial.js
+++ b/src/title/TitleSpecial.js
@@ -8,7 +8,7 @@ import { Icon } from '../common/Icons';
 import styles from '../styles';
 
 const specials = {
-  home: { name: 'All messages', icon: 'home' },
+  home: { name: 'All messages', icon: 'globe' },
   private: { name: 'Private messages', icon: 'mail' },
   starred: { name: 'Starred', icon: 'star' },
   mentioned: { name: 'Mentions', icon: 'at-sign' },


gnprice:
Ah, I was thinking actually of "globe" for the upper left -- the one that means "everything" or "all messages". I think the "house" icon is pretty OK for the "home" view aka unreads.

gnprice:
> Inbox

That'd be this:

image

I'd be OK with using that too for "All messages". I remember discussing that now that you mention it, but when @jackrzhang and I looked back through #3063 we didn't see it.

Thinking about "globe" vs. "inbox" for the icon for "All messages", I think my preference is "globe". In particular both "All messages" and the home screen i.e. "Unreads" have sort of the nature of an inbox, and "Unreads" feels like it's more strongly an inbox than "All messages" is. (The "All messages" view does start at your first unread, but it will also include a lot of stuff you've already seen.) But it's not a strong preference. @borisyankov what are your thoughts between those two?

Whatever change we make, we'll also see what feedback we get after sending out the beta; it's easy to change again at that point if we're persuaded of a different choice.

jackrzhang:
Ah, gotcha @gnprice. (updating)

borisyankov:
The argument against `globe` is that users need to associate it in the other direction that we think of it, not 'what icon should we set to that concept' but 'what does this icon mean' which almost never works for metaphors and not direct depictions of things.

I think that globe means 'world', means 'localization', in fact, we already use it `Settings -> Language`.

Replacing the bottom icon with 'Inbox' achieves is enough to achieve the goal of having different icons for top and bottom.

gnprice:
> I think that globe means 'world', means 'localization'

Hmm interesting.

I don't think this is likely to be a common association -- I don't think I often see it in applications, and also "localization" is not a concept that people not involved in writing software are likely to think about. :-)

It is a pretty reasonable answer for the inverse problem that you mention ("what icon can we use for this 'what language' setting?"), where it supplements a text label.

In any case "house" (-> "home") is completely perverse for the "all messages" screen -- a screen that we don't expect most users to often, or ever, use. So even if we replaced it with an icon that was utterly mystifying for most users, that would be an improvement because at least it wouldn't actively mislead them. But I think "globe" -> "the whole world" -> "everything" is likely to do better than that.

borisyankov:
`localization` => I meant `change language`.
This is the exact icon used on iOS's Keyboard to switch languages.

jackrzhang:
Based on the conversation above, I've updated this PR, using the globe icon for "All messages" and the inbox icon for "Unread". Here's how it looks:



Thoughts?

armaanahluwalia:
I think the nav layout needs a rework since the presence of nav links both up top and below with seemingly no thought as to which go where is confusing at best. The fact that swiping right from the  “home” screen removes the top menu altogether suggests it is a sub menu to the home section. If that is our thinking then here’s the changes I would like to see:

1) Add a red circle on top of the home icon ( bottom left ) denoting the number of unread messages. This would have two positive effects; 
    * make the eye travel there first
    * signal it’s the focus of the application
    * provide instant information on unreads

2) Change the home icon on the top left to a “list” icon. I think this more accurately denotes the chronological and flat nature of all messages.

Here are some longer term changes I’d like to see to nav ( a little off topic from this discussion but worth mentioning anyway ):

1) Make the styling of the top nav in the home section more clearly a secondary menu.

2) Make the secondary menu up top in the home section a tabbed view instead launching new sections altogether. This disrupts the mental model of what’s going on because they feel like modals rather than content sections and we lose the navibility from them.

3) Add a default icon to the tabbed secondary menu up top for unread messages which should be the view you initially land on.

jackrzhang:
@armaanahluwalia I agree that higher-level changes to how navigation works would be beneficial.

At the moment, it's pretty confusing just to see the same icon for two different buttons on the same screen, and I think the commits in this PR are specifically a simple and incremental improvement for that specific aspect of the layout.

armaanahluwalia:
I agree that changes to make them not the same icon are probably good. What did you think about the rest of the stuff I said?

Also: I think instead of inbox icon + globe icon we should use home envelope icon + list icon.

Here’s my defence of my icon choices:

The envelope is the same metaphor for messages only it’s more universal and obvious that it’s messages.

The list more accurately depicts all messages because it’s obvious it’s a collection of things ( more obviously a list of messages since it’s supposed to be a secondary menu within messages )

gnprice:
Thanks again @jackrzhang ! Merged, with a couple of tweaks -- take a look.

borisyankov:
??????????????

borisyankov:
How did we agree to replace the 'All messages' icon with the 'Change language' keyboard icon from iOS?

borisyankov:
The purpose of the PR was as the name suggests "Stop using the same icon for two different buttons on the main screen. "

That was achieved with the first two commits.

The second change was utterly ridiculous for two reasons:
 * it changes the icon from what is in the web app
 * it is a very well known system icon on iOS

gnprice:
> How did we agree to replace the 'All messages' icon with the 'Change language' keyboard icon from iOS?

I think that won't be a salient association for most users.

We'll see what feedback we get! Certainly easy to change to another icon later. But this is a big improvement over the highly misleading "home" icon.

borisyankov:
> But this is a big improvement over the highly misleading "home" icon.

You mean the icon we use on the web?

> I think that won't be a salient association for most users.

I can't imagine how that would be. It is on **every** iOS keyboard:

https://photos5.appleinsider.com/gallery/22924-28346-leftiekeys-l.jpg

armaanahluwalia:
I thought my pitch for a notification count and icon choices was supported by pretty good reasoning. 🤷‍♂️

gnprice:
I don't love the icon on the web either. But on the web, the icon is accompanied by a text label explaining what it actually means:
image

That makes it a lot lower-stakes if the icon is misleading. We don't have that luxury on mobile.


> It is on **every** iOS keyboard

Huh, I'd have expected it would only be present when you actually have multiple languages / input methods configured. Which I don't think most users do.

Is it really there out of the box, with just one keyboard configured? That would surprise me.

timabbott:
> > But this is a big improvement over the highly misleading "home" icon.

> You mean the icon we use on the web?

FWIW, we're probably going to make that icon go with "All unreads" on the web within a month or so.

gnprice:
@armaanahluwalia I think broader changes to the nav are certainly in order!

But they will take some work. That means they should not block fixing this longstanding complaint which takes just a few lines to fix. :smile:

borisyankov:
@timabbott I am talking about the 'Change language' icon that is now used for 'All messages'.

Anyone who knows another language would use a secondary keyboard.

armaanahluwalia:
Globe appears on all iPhones I believe out of the box.

gnprice:
Here's a screenshot of an iOS keyboard without the globe:
image

In fact it comes in a demonstration of typing multiple languages at once! Since iOS 10, even as a multilingual user it's been unnecessary to configure additional keyboards unless the different languages you want to type use different scripts.
https://www.idownloadblog.com/2017/01/16/how-to-multilingual-typing-iphone/

armaanahluwalia:
Can we create a ticket for rethinking the nav a little bit? I agree this is a positive change but there's a lot to do to make the whole thing less confusing. I made quite a few suggestions I think are valuable and would appreciate any sort of feedback on those :)

borisyankov:
Yeah, I agree not everyone needs a second keyboard.
360 million people speak English as their first language. That is 4.78% of all people on Earth.

gnprice:
> 360 million people speak English as their first language

Yes, and the screenshot above is of someone typing both Spanish and English, with a single keyboard.

I've made a decision based on the information we have so far, and I think this discussion is no longer adding new information, so let's wrap it here. We'll have more information after users see 23.2.111 with this change and give us feedback (or don't); we can pick the discussion up again when we know more.

gnprice:
> Can we create a ticket for rethinking the nav a little bit? I agree this is a positive change but there's a lot to do to make the whole thing less confusing. I made quite a few suggestions I think are valuable and would appreciate any sort of feedback on those :)

Sure, more discussion sounds good. How about starting a chat thread?

I think that'll be a somewhat better medium for that discussion, as there are several different ideas and also there's discussion to be had before a direction is fully specced out.

Issue #3150, opened 2018-11-16
Use a distinct Zulip notification sound, on Android


kalzekdor:
I've recently started using the Zulip client for Android. One thing that seems missing is the option to use a custom notification sound. Any notifications use the default notification sound. I heavily prefer having different notification sounds for different things, so I can tell without looking at my phone what's asking for my attention. The ability to customize the notification sound is a major QoL feature for me, even better would be the ability to set a per-stream notification sound.

Are there any plans along these lines?

borisyankov:
Thank you for raising this issue. I can see how that is important.
We have thought about that but I don't think we considered this a priority.
The more we hear about a certain feature requested, the more we bump it up.
Do you expect the default sound to be different/better or customizable and be able to be set by you? (I think the first option, being simpler has better chance of being implemented sooner)

kalzekdor:
I'm not sure what you mean by a different default sound. It's using the default notification sound as dictated by Android. I can change this in the Android settings, so it's not a problem with the particular sound. I would definitely like to be able to set a customized sound particular to Zulip.

Having this on the roadmap is good enough for me. I don't have any experience with Android development, though I've been meaning to pick it up. Might see if I can take a crack at it if I can find the time.

borisyankov:
I think I did explain it incorrectly. I meant different default sound for the Zulip app (not the system).

 But if you are interested in trying to do it yourself, would be even cooler. 👍

kalzekdor:
No, I was referring to the sound played by the push notifications.

gnprice:
Thanks @kalzekdor for writing in!

Zulip actually has a custom notification sound specific to Zulip, which we've used in the webapp for a long time and which we recently fixed our notifications on iOS to use. There's code in the app that's intended to do this for Android too -- so the fact that it isn't is a bug.

It sounds like if we fixed that bug it would largely cover what you really want here, so I've taken the liberty of retitling this issue to reflect that bug. Conveniently we *also* just added support in the webapp for customizing the notification sounds, which in the future the mobile app will also support; see #3204 for that.

Personally I normally have notification sounds disabled system-wide on my phone ("ring volume" turned down to zero), so I've never noticed. I tested just now (on Android 9), and I reproduce -- I get the system default sound just as you do. So we just need to debug and fix.

gnprice:
PS: here's that code in the app:
```
android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java

[...]

PR #3233, opened 2018-12-22
android, notif: Fix custom sound is not played. Fixes: #3150

jainkuniya:
Fixes: #3150

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -50,6 +50,9 @@ public class GCMPushNotifications {
             CharSequence name = context.getString(R.string.notification_channel_name);
             int importance = NotificationManager.IMPORTANCE_HIGH;
             NotificationChannel channel = new NotificationChannel(CHANNEL_ID, name, importance);
+            AudioAttributes audioAttr = new AudioAttributes.Builder()
+                    .setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
+            channel.setSound(getNotificationSoundUri(context), audioAttr);
             getNotificationManager(context).createNotificationChannel(channel);
         }
     }
@@ -185,7 +188,9 @@ public class GCMPushNotifications {
         }
 
         final Uri soundUri = getNotificationSoundUri(context);
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+        if (android.os.Build.VERSION.SDK_INT >= 26) {
+            // Managed by the channel.
+        } else if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
             AudioAttributes audioAttr = new AudioAttributes.Builder()
                     .setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
             builder.setSound(soundUri, audioAttr);
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -22,6 +22,7 @@ import com.zulipmobile.R;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.List;
 import java.util.Locale;
 
 import me.leolin.shortcutbadger.ShortcutBadger;
@@ -36,7 +37,8 @@ import static com.zulipmobile.notifications.NotificationHelper.TAG;
 
 public class GCMPushNotifications {
 
-    private static final String CHANNEL_ID = "default";
+    private static final String CHANNEL_ID = "default_v2";
+
     private static final int NOTIFICATION_ID = 435;
     static final String ACTION_CLEAR = "ACTION_CLEAR";
     static final String EXTRA_NOTIFICATION_DATA = "data";
@@ -47,6 +49,17 @@ public class GCMPushNotifications {
 
     public static void createNotificationChannel(Context context) {
         if (Build.VERSION.SDK_INT >= 26) {
+            // for now we only support one default channel
+            // it is not possible to change channel properties after creation
+            // so new channel is created in such cases
+            // which can result in many old unused channel, so delete them
+            List<NotificationChannel> channels = getNotificationManager(context).getNotificationChannels();
+            for (NotificationChannel channel : channels) {
+                if (!channel.getId().equals(CHANNEL_ID)) {
+                    getNotificationManager(context).deleteNotificationChannel(channel.getId());
+                }
+            }
+
             CharSequence name = context.getString(R.string.notification_channel_name);
             int importance = NotificationManager.IMPORTANCE_HIGH;
             NotificationChannel channel = new NotificationChannel(CHANNEL_ID, name, importance);


android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -172,12 +180,13 @@ private static void updateNotification(
...
 
+        // Note: This code only set custom sound for API < 26 (from higher API sound is managed by channels)
         if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
             AudioAttributes audioAttr = new AudioAttributes.Builder()
                     .setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
-            builder.setSound(Uri.parse("android.resource://" + context.getPackageName() + "/" + R.raw.zulip), audioAttr);
+            builder.setSound(Uri.parse(ContentResolver.SCHEME_ANDROID_RESOURCE +"://" + context.getPackageName() + NOTIFICATION_SOUND_PATH), audioAttr);
         } else {
-            builder.setSound(Uri.parse("android.resource://" + context.getPackageName() + "/" + R.raw.zulip));
+            builder.setSound(Uri.parse(ContentResolver.SCHEME_ANDROID_RESOURCE +"://" + context.getPackageName() + NOTIFICATION_SOUND_PATH));
```
gnprice: I like the idea of deduplicating the `"/raw/zulip"` string. Let's take that farther -- let's move the whole `Uri.parse(...)` call into a little static method that returns a `Uri`.

That will also let us bring the `"/raw/zulip"` string constant back into that code -- the code is easier to read if it doesn't refer to a static constant defined 100 lines up in the file.

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -36,6 +37,7 @@
 
     private static final String CHANNEL_ID = "default";
     private static final int NOTIFICATION_ID = 435;
+    private static final String NOTIFICATION_SOUND_PATH = "/raw/zulip";
```
gnprice: Hmm. I'd prefer the `R.raw.zulip` form, because it's more transparent to tools, like Android Studio.

That form [is documented](https://developer.android.com/reference/android/content/ContentResolver#openAssetFileDescriptor(android.net.Uri,%20java.lang.String,%20android.os.CancellationSignal)), with no sign of deprecation. Seems like it should work.

You mention in the commit message an error about `android.resource://com.zulipmobile.debug2131165184`. That is missing a slash. Perhaps that's an unrelated issue?

android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
```diff
@@ -172,12 +180,13 @@ private static void updateNotification(
             builder.addAction(action);
         }
 
+        // Note: This code only set custom sound for API < 26 (from higher API sound is managed by channels)
         if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
             AudioAttributes audioAttr = new AudioAttributes.Builder()
                     .setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
-            builder.setSound(Uri.parse("android.resource://" + context.getPackageName() + "/" + R.raw.zulip), audioAttr);
+            builder.setSound(Uri.parse(ContentResolver.SCHEME_ANDROID_RESOURCE +"://" + context.getPackageName() + NOTIFICATION_SOUND_PATH), audioAttr);
```
gnprice: I agree with using `SCHEME_ANDROID_RESOURCE`, but I don't think the reasons in your commit message are realistic. :-)

The upstream docs [say this constant](https://developer.android.com/reference/android/content/ContentResolver#SCHEME_ANDROID_RESOURCE) was introduced in API 1, they specify its value (as "android.resource"), and they don't mention it ever having changed. So this is specified to be a fixed constant.

If a future API version (or a rogue device vendor) wants to do something differently here, it's likely to involve different semantics that require code changes to adapt to anyway; in the unlikely event that they change the existing constant's value, it's not clear that we'd be closer to working by following the constant vs. following the value.

What using the constant *does* do for us is explicitly draw the connection with what bit of the system API we're relying on. Helps catch typos, and better yet makes it easier to look up relevant docs. This is a good example: I actually found it hard to google for information about these resource URIs (to answer my questions about `R.raw.zulip` vs. `"raw/zulip"`), but then I looked up [`ContentResolver.SCHEME_ANDROID_RESOURCE`](https://developer.android.com/reference/android/content/ContentResolver#SCHEME_ANDROID_RESOURCE), and the helpful docs I linked to above which answered my questions are there on the same page.

borisyankov:
Before we decide on a better default sound, I would consider this to be a regression (even if code-wise it is perfect).

gnprice:
> Before we decide on a better default sound, I would consider this to be a regression (even if code-wise it is perfect).

Oh interesting -- what don't you like about this sound? Is there a discussion somewhere about a better one?

This is the same sound we use for notifications on the desktop. That is a different environment, so it might make sense to differ; but it'd also be valuable to be recognizably "the same" sound, so ideally it'd be just a variation of the same thing with some appropriate adjustment.

gnprice:
Ah, there's discussion [in chat](https://chat.zulip.org/#narrow/stream/101-design/subject/new.20sounds/near/672052) (see linked message and also a few messages earlier, two polls) of new sounds to offer in the webapp. So whichever of those we end up liking and adding, those will be a good candidate pool.

Other notes from discussing with @borisyankov just now:
* The current sound file is pretty quiet. That part is straightforward to fix, but...
* The low pitch, and the lack of a distinct tone, make the sound pretty subtle even when the amplitude is cranked up to just before the point of clipping.
* I think that was intentional in the sound's original design -- to make it less disruptively interrupting. That makes sense when on headphones at a desk in an office; it doesn't work so well when the point is to be heard from inside one's pocket, potentially in a noisy environment.
* Many of the new sound candidates in that thread are more like classic mobile notification sounds, with higher pitches, tonal quality, and a bit longer duration with gradual fall-off -- in short, more bell-like. So there should be several we end up with that would be suitable here.

For next steps for this PR: @jainkuniya , how about you go ahead and make the code revisions discussed above. These are good cleanups, and I'd be glad to merge them -- with perhaps a commit on top to disable the sound for now. Then in a week or two when we have a sound we like, we'll be all set to drop it in and re-enable.

jainkuniya:
(Need to do testing)

gnprice:
Thanks again @jainkuniya !

I've just pushed three of these commits, as 6becc6e91..356a88dce .

I've left out:
* The actual removing of `setDefaults`. This affects not only sound, but also vibration. We have a separate bug in our manifest (really, we're affected by a bug in the wix notification library's manifest) which means we don't request the `VIBRATE` permission -- so our `setVibrate` call is a permissions error, which the `setDefaults` has been covering up for us.

  On my device the `setVibrate` doesn't do anything anyway because I'm on Android 9; but when I simulate it by using `Vibrator#vibrate` to try to run the same vibration, the app crashes. Based on my reading of the docs, I expect that on API < 26 the app would crash in the same way (or at best, the vibration simply wouldn't work) with the version in this PR.

  I have a fix for this bug on a branch of mine, so I'll see about merging that fix.

* The channels part. This design has the consequence that each previous channel ID we've used accumulates in the code as something we need to track forever. That's avoided by the approach I suggested above:

  > [...] change CHANNEL_ID, and then in this code, look for any existing channels with other IDs:
  > https://developer.android.com/reference/android/app/NotificationManager.html#getNotificationChannels()
  > and delete those.

  Would you try reworking those commits to use that design?

gnprice:
OK, merged my fix for the VIBRATE permission, and also took care of `setDefaults`. See e2decb69e^..458ef8832.

Note two changes I made there:
* I adjusted how we identify the system default notification sound. See commit message.
* I kept DEFAULT_VIBRATE and DEFAULT_LIGHTS, and added a separate TODO item to remove DEFAULT_VIBRATE.

I'd be glad to merge a change to handle channels the way I described in the previous comment. Then the vibration settings would be a good followup after that.

gnprice:
Sorry this PR seems to have fallen through the cracks. Just bumped the priority (following the issue, #3150), to help myself spot it to get back to it.

gnprice:
This was subsumed by #5221 and the issue was fixed, earlier this year.

PR #3234, opened 2018-12-24
Register for all events

borisyankov:
Rework `registerForEvents` call completely.
Start registering for all server events.

### DIFF
--- a/src/api/__tests__/apiFetch-test.js
+++ b/src/api/__tests__/apiFetch-test.js
@@ -1,6 +1,48 @@
-import { getFetchParams } from '../apiFetch';
+import { objectToParams, getFetchParams } from '../apiFetch';
 
 global.FormData = jest.fn();
+
+describe('objectToParams', () => {
+  test('object consisting of numbers, strings or booleans is not changed', () => {
+    const obj = {
+      numKey: 123,
+      strKey: 'str',
+      boolKey: true,
+    };
+    const result = objectToParams(obj);
+    expect(result).toEqual(obj);
+  });
+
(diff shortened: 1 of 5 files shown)


gnprice:
Thanks @borisyankov !

I'm concerned about the duplication of `registerForEvents` plus `registerForAllEvents` -- particularly of the flags like
```
    apply_markdown: true,
    include_subscribers: false,
    client_gravatar: true,
```
They're identical for now, but this feels like the sort of setup that can very easily lead to one diverging from the other, and then to extreme confusion when the observed behavior seems to differ from the flags (because the flags we're actually using are in the other copy of this code.) This is exacerbated because the two copies are in completely separate files, making it very easy while looking at one to never realize the other exists.

I don't have a strong preference on whether we have one or two different functions exposed to the application code, but one way or another I definitely want those flag values to be written down in one shared place. Could have both import them from some "constants" or "defaults" module, or could give one of these modules a non-default export that does most of the work and then both modules' default exports just wrap that, or could put both of these in one module, etc.

borisyankov:
Good points.

I did try to merge some work from the `api` PR, then iterated several times, then reworked this completely and this is the result.

gnprice:
Thanks! I like this direction, moving the specific options we use into the application code.

A couple of comments on specifics:
* Let's leave `null` values in; those translate fine into JSON, and on the Python side into `None`.
* In the commit that moves the options to the application code, another option `all_public_streams: false` is added. That should be a separate commit, if we want to do it.

gnprice:
Thanks again @borisyankov ! Merged.

PR #3235, opened 2018-12-24
android notif: Create Notification channels for group, private, stream

kunall17:
This would allow users to use the android features for different
reciepeint types (stream, private, group)

Android offers them to change importance, sound vibration
notification dot and other customization they can do
for different type of messages

Categorize the notification props according to the message type
and assign the specific channel ID string to the notification.

These three channels are being created.
device-2018-12-24-105857

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -20,6 +20,8 @@ import com.zulipmobile.R;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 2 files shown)


jainkuniya:
Or another way can be realm based channels, like Gmail app.

kunall17:
For multiple realms we don't support notifications yet anyways.
And grouping can be done for multiple realms, like different channel categories can be there for that.

gnprice:
We have issue #3401 for tracking this feature, which we definitely want in some form. We're not going to use this implementation, though, so closing the PR.

(One thing we'll want to do differently from this version: the selection of channels should correspond to categories we choose from a product perspective, rather than from an implementation perspective where 1:1 PMs are a distinct category from group PMs. They should also be separate per Zulip account; we long ago fixed notifications so that they work across multiple accounts in multiple realms.)

Issue #3236, opened 2018-12-24
screen, layout: There is some white BG in screen even in dark theme. 


jainkuniya:
Current master:
![image](https://user-images.githubusercontent.com/18511177/50405195-08e37c00-07d7-11e9-9a86-eebb98d9fcb3.png)

Ideally in dark theme, whole screen should have dark theme BG color.

@zulipbot claim

PR #3237, opened 2018-12-24
screen, layout: Fix white BG/panel at the bottom of screen.

jainkuniya:
Apply padding instead of margin in account for safeAreaInsets, so that
theme BG color is painted on whole screen.

Fixes: #3236

image


borisyankov:
While we are fixing this more thoroughly in #3067 this is straightforward enough to merge independently.

PR #3240, opened 2018-12-26
Fix app links

### LINKED ISSUE #2760: Internal links to streams and narrows don't work

borisyankov:
Fixes #2760 .

Ready for review and merge separately.


gnprice:
Thanks @borisyankov ! On a quick read (as I have to run off shortly), this looks generally good.

Small things:
* The commit that deletes `is*Link` is best squashed into the one that inlines them into their call site; that makes it extra clear where all that code is coming from.
* The new `getLinkType` is still a bit messy. I'd be glad to see a follow-up to clean it up some -- e.g. `if (paths.length == 2) { /* ... group, stream, special ... */`. Or perhaps `switch (paths[0])` (after checking length is nonzero) would be the best outermost test.
* I like the change to return `null` when not parseable! :smile: And for the caller to then handle that as appropriate for its context. I guess `getLinkType` is another place you'll want to do that (you mention already that there are more cases you have in mind.)

borisyankov:
Merged the two suggested commits.
Polished significantly the last two commits.
Ready for review.

src/utils/internalLinks.js
```diff
@@ -0,0 +1,117 @@
...
+
+export const isInternalLink = (url: string, realm: string): boolean =>
+  isUrlOnRealm(url, realm) ? /^(\/#narrow|#narrow)/i.test(url.split(realm).pop()) : false;
+
+export const isMessageLink = (url: string, realm: string): boolean =>
+  isInternalLink(url, realm) && url.includes('near');
+
+type LinkType = 'external' | 'home' | 'group' | 'topic' | 'stream' | 'special';
+
+export const getLinkType = (url: string, realm: string): LinkType => {
```
gnprice: Some other cases I'm curious about:
* With `/near/ID` on a stream
* A PM with one person, not a group
* A search
* A sent-by query
* A search or sent-by query, combined with a stream and/or topic, or with a group PM thread, etc.

More generally, the URLs we support in the webapp are a lot richer than this code is built to accept. A couple of the cases above should probably be fixed even with the current structure of this code, but I think this is not the right architecture and in the end we should do something that better matches how the URLs actually work.

(None of that is an obstacle to merging this PR, though! I see that it's refactoring the logic that exists, and that's a valuable thing to do. Improvements to the behavior may or may not land in this PR, but they should definitely be separate commits.)

src/utils/internalLinks.js
```diff
@@ -0,0 +1,117 @@
...
+  return 'home';
+};
+
+export const getNarrowFromLink = (
+  url: string,
+  realm: string,
+  users: User[],
+  streams: Stream[],
+): ?Narrow => {
+  const type = getLinkType(url, realm);
```
gnprice: The commit that switches to using this is best squashed with its parent which creates this helper. That makes it a lot clearer why the details of the helper are as they are -- they're just a factoring-out of existing logic in this function.

If that helper `getLinkType` were being introduced fresh, I would have a lot of questions to address before wanting to merge it. ;-) See comment above for some of them. But when it's just moving logic we already have and making it better, it gets a pass on a lot of that.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -0,0 +1,270 @@
...
+    ).toBe('group');
+    expect(getLinkType('https://example.com/#narrow/stream/jest', 'https://example.com')).toBe(
+      'stream',
+    );
+    expect(getLinkType('https://example.com/#narrow/stream/stream/', 'https://example.com')).toBe(
+      'stream',
+    );
+  });
+
+  test('when a url is not a topic narrow return false', () => {
```
gnprice: This and some of the other test names should be updated in the commit that changes a lot of the test cases, "Start testing `getLinkType`" -- they no longer make sense as they are.

src/utils/internalLinks.js
```diff
@@ -0,0 +1,117 @@
...
+
+  return 'home';
+};
+
+export const getNarrowFromLink = (
+  url: string,
+  realm: string,
+  users: User[],
+  streams: Stream[],
+): ?Narrow => {
```
gnprice: This is never `undefined`, right? Only `null`, or a `Narrow`.

If that's right, let's write `Narrow | null` to describe the situation precisely. Flow's shorthand `?Narrow` unfortunately means `Narrow | null | void`, which is virtually never what we want -- the reality is almost always something more specific, like `Narrow | null` or `Narrow | void`.

Where this becomes really important is when the main type is `string` or `number` or `boolean`, something that can itself be falsy -- because then `if (foo)` is bad, and if we write `foo: ?Foo` we have to write redundant checks like `if (foo !== null && foo !== undefined)`, when in reality only one of `null` and `undefined` can ever happen. But I think it's simpler to have a blanket policy of writing the specific `Foo | null` or `Foo | void` and never `?Foo`, even when we could get away with `if (foo)`.

gnprice:
Bump. From my comments above, this looks pretty close! Would be great to see it merged.

gnprice:
Also I just edited the description to link to the issue -- because I ran across the issue earlier today, and just spent a couple of minutes searching to find this issue because there wasn't a link.

That is why I asked for a link in one of my very first comments:

> We have an issue for this, right? Would you link to the issue from the PR thread?

I'm disappointed that you didn't. This really makes a difference in smoothing the workflow -- I would really appreciate if you would acquire a habit of linking PRs to the issues they're about.

borisyankov:
> The commit that switches to using this is best squashed with its parent which creates this helper.
Done. Squished into one.

borisyankov:
> This is never undefined, right? Only null, or a Narrow.

Changed to `Narrow | null`

borisyankov:
I've added new commit 'Rename isGroupLink to isPmLink'.

borisyankov:
Also, the link type that was previously 'group' is now 'pm' to properly reflect the actual value.

src/utils/internalLinks.js
borisyankov: Squashed.

src/utils/internalLinks.js
borisyankov: A PM with one person, not a group

'group' now renamed to 'pm' to reflect the correct usage.

borisyankov:
> * With /near/ID on a stream
> * A search
> * A sent-by query
> * A search or sent-by query, combined with a stream and/or topic, or with a group PM thread, etc.

These remain unsupported. Should we document this somehow?

gnprice:
Thanks again @borisyankov !

These updates look good; I'll expect to merge (possibly with small tweaks) soon.



> > * With /near/ID on a stream
> > * A search
> > * A sent-by query
> > * A search or sent-by query, combined with a stream and/or topic, or with a group PM thread, etc.
> 
> These remain unsupported. Should we document this somehow?

Ah, good thought! Yeah -- jsdoc on `getNarrowFromLink` would be a reasonable place. Can always move from there to somewhere else if we think of someplace better later.

gnprice:
I've now come back to this.

The first 10 commits, performing various refactors, look good now and I'm about to merge.

With the various revisions having made these changes much more readable, I've now read closely the last two commits:

#### c4d8697f6 Refactor and cleanup `getNarrowFromLink` tests

This commit makes each test case much more verbose.

Probably not coincidentally, it also deletes the majority of the test cases. Some of them arguably were redundant, but many were testing important edge cases.

There are things I'd change about the existing tests here, but this change is mainly in the wrong direction. So I'm not going to merge it.

One thing this change does that I like is using proper helpers like `groupNarrow` and `starredNarrow` to construct narrow objects in some places where the existing code uses narrow literals. This (a) makes the tests more concise and clearer, and (b) better respects the `Narrow` type as an abstraction. I'd have merged this as a commit of its own, and I may make a similar change separately.


#### e74390cfb getNarrowFromLink: Parse stream IDs in links

I don't love the internal structure of the new code here.

But a more serious problem is that this commit adds no tests at all for the rather subtle new case it's handling.

And more serious still is that a minute's effort writing such a test, while looking closely at the code, finds a bug in the new logic. This test fails:
```
  test('old-style stream link but with hyphen', () => {
    const url = 'https://example.com/#narrow/stream/with-hyphen';
    const expected = streamNarrow('with-hyphen');
    const result = getNarrowFromLink(url, 'https://example.com', usersById, streams);
    expect(result).toEqual(expected);
  });
```

---

The last commit is the one that actually aims to fix #2760, so that will remain open after I merge those first 10 commits. Happily the refactors in those commits will be helpful, though, as I go on to edit this code to fix that bug.

gnprice:
Merged those first 10 commits as 92455a9c1..2aa6c2264, with some small adjustments.

Thanks again @borisyankov for all the work you did on this!

Issue #2691, opened 2018-06-18
notifs: Use Android's fancy "notification group" feature


gnprice:
Android has a handy feature we should use to make all our notifications a "group", with one for each conversation.

Upstream [tutorial docs here](https://developer.android.com/training/notify-user/group). An illustration from that doc, to help make clear what feature I'm referring to:
![image](https://user-images.githubusercontent.com/28173/41548988-4e5286b6-72d9-11e8-996a-851f02894ff5.png)

One key benefit of this is that each conversation's notification can take the user to the respective conversation.

This works great for Gmail (as in the illustration) and FB Messenger, and we should do the same thing.

(This Android feature was introduced in Android 7.0 Nougat, and we support some lower versions, so we'll need an appropriate conditional and fallback. But [as of 2020-04](https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/863756), >95% of our users are on Android 7 or higher and so will benefit from this feature.)

jainkuniya:
I am working on this and submitted WIP PR #3407

gnprice:
Android 11 was just released today, and it introduces some further fancy features in this direction: "conversations" and "bubbles".
https://developer.android.com/guide/topics/ui/conversations
We should target those too. I haven't yet read about these new API features in detail; it may be that it'll make most sense to target them at the same time as this older "notification group" feature. (Which after all is aimed at a highly overlapping use case.) If not, we'll make a new issue for these new Android 11 features.

In general our userbase is [fairly quick to upgrade](https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/1012192) to new OS releases; if the Android 11 rollout goes like Android 10 did before it, then a sizeable chunk of our Android users will get it over the next few months, and a majority will within a year.

gnprice:
This will probably require some improvements to the data structure we keep as `ConversationMap` in the notifications code. There's some comments pointing in that direction by the definition of that type. I'll file a specific issue for that in a bit.

PR #3241, opened 2018-12-27
[WIP] Show grouped notifications using messaging style 

kunall17:
This changes the way notifications are displayed, previously only the last message content was being displayed in a summarised format.
This changes to a grouped notification and now each notification is being separately notified in a messaging style format. This adds the possibility of narrowing to that particular notification, action events (replies or cancellation) of each narrow. 
Now it shows all the messages of a narrow saved in our data structure.

**Screenshots:** 
Expanded
expanded
In collapsed way
device-2018-12-27-113238

The titles can be discussed and changed as per discussion.

The code is mostly taken from the android docs
For grouping the notifications: https://developer.android.com/training/notify-user/group
For Messaging style notifications: https://developer.android.com/training/notify-user/expanded#message-style

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/GCMPushNotifications.java
@@ -20,6 +20,8 @@ import com.zulipmobile.R;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


gnprice:
This seems to have been a mistyped "Fixes:" reference. :-/

gnprice:
This was subsumed by #4842, which completed #2691.

PR #3243, opened 2018-12-27
Add new API functions

borisyankov:
These were extracted from #2819 for easier merging.

### DIFF
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -47,6 +47,17 @@ export type Reaction = $ReadOnly<{|
   emoji_code: string,
 |}>;
 
+export type MessageHistory = $ReadOnly<{|
+  content: string,
+  rendered_content: string,
+  timestamp: number,
+  topic: string,
+  user_id: number,
+  prev_content?: string,
+  prev_rendered_content?: string,
+  content_html_diff?: string,
+|}>;
+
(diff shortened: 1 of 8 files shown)


gnprice:
Thanks @borisyankov ! These generally look good. A few small points:
* jsdoc should go on the functions (not the types), so hover works; and with no blank line in between, so it doesn't end up getting separated later.
* Let's use exact object types; and spread rather than intersection-type syntax, because the latter seems to have puzzling semantics in Flow (even in the non-exact case, where as far as I can see they ought to be equivalent.)
* Let's put some quick jsdoc on the `UserProfile` type; even just a link to the `get-profile` API doc would be helpful.
* The doc for `create-user` says the user requesting it has to be an organization administrator -- it's not for anonymously creating an account from scratch. So (a) we should pass a real `auth` after all, and (b) as that's (empirically :slightly_smiling_face: ) a surprising requirement, let's mention it in the jsdoc for the function.

borisyankov:
Feedback implemented. Ready for a final review.

gnprice:

I made just two changes:
* The commit message on `createUser` hadn't been updated to reflect our learning that it's for org admins' use :stuck_out_tongue: ; I updated it.
* I added a commit on top adjusting getMessageHistory's type, and adding a bit of jsdoc, based on my reading of its API doc page.

PR #3246, opened 2018-12-28
Use full API response

borisyankov:
These commits were extracted from #2819 for easier merging.

### DIFF
--- a/src/alertWords/alertWordsActions.js
+++ b/src/alertWords/alertWordsActions.js
@@ -11,5 +11,6 @@ export const initAlertWords = (alertWords: string[]) => ({
 
 export const fetchAlertWords = () => async (dispatch: Dispatch, getState: GetState) => {
   const auth = getAuth(getState());
-  dispatch(initAlertWords(await getAlertWords(auth)));
+  const { alert_words } = await getAlertWords(auth);
+  dispatch(initAlertWords(alert_words));
 };
--- a/src/api/alert_words/getAlertWords.js
+++ b/src/api/alert_words/getAlertWords.js
@@ -1,6 +1,11 @@
 /* @flow strict-local */
-import type { Auth } from '../apiTypes';
+import type { Auth, ApiResponseSuccess } from '../apiTypes';
 import { apiGet } from '../apiFetch';
 
-export default async (auth: Auth): Promise<string[]> =>
-  apiGet(auth, 'users/me/alert_words', res => res.alert_words);
+type ApiResponseAlertWords = {|
+  ...ApiResponseSuccess,
+  alert_words: string[],
+|};
+
+export default async (auth: Auth): Promise<ApiResponseAlertWords> =>
+  apiGet(auth, 'users/me/alert_words');
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -7,10 +7,6 @@ import { networkActivityStart, networkActivityStop } from '../utils/networkActiv
 
 const apiVersion = 'api/v1';
 
-type ResponseExtractionFunc = (response: Object) => any;
-
-const defaultResFunc: ResponseExtractionFunc = res => res;
-
 export const objectToParams = (obj: Object) => {
   const newObj = {};
   Object.keys(obj).forEach(key => {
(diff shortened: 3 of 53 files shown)


gnprice:
Thanks @borisyankov !

Flow finds errors in a couple of these.

Some changes made in #3243 also apply to each of these: jsdoc immediately on function, and spread rather than intersection types.

```
-  dispatch(initAlertWords(await getAlertWords(auth)));
+  const alertWordsResponse = await getAlertWords(auth);
+  dispatch(initAlertWords(alertWordsResponse.alert_words));
```

Here's what I wrote about this [back in September](https://github.com/zulip/zulip-mobile/pull/2819#pullrequestreview-153541600):

"""
So this one and a bunch of others pose a bit of an interesting API design problem: the response contains just one meaningful field (one field other than `{'result': 'success', 'msg': ''}`), and that field's name just repeats part of the name of the API call. It feels like the calling code gets a little worse when it has to extract that out -- it's just a bit of extra ceremony and repetition that doesn't add any clarity.

For these cases, I wonder if it'd be better to keep the extraction, but just do it in more idiomatic style?  E.g.:
```
export default async (auth: Auth): Promise<string[]> =>
  apiGet(auth, 'users/me/alert_words').then((r: ApiResponseAlertWords) => r.alert_words);
```

The price is a bit of non-uniformity between different API functions -- some do an extraction, some don't. I think the non-uniformity would make the extraction not worth it, if we could only apply it in a few cases. But it looks like in fact the majority of API calls follow this pattern (which makes sense -- lots of questions naturally have an answer that's just one thing, or one uniform list of things), so I think keeping the extraction for those probably is worth it.

What do you think?
"""

borisyankov:
I think the `apiCall().then(result => result.field)` is a better approach than an extractor function.

But the reason to remove that function was exactly for consistency of the return value and also to make sure adding a field to a previously single-value response not be a breaking change.

I expect this to happen more in the future, but previously `getMessages` was returning only `messages` and now it returns several other fields `found_newest`, `found_oldest` etc.

What makes it a bit prettier is the use of destructuring like:
```js
const { messages } = getMessages(...);
```

borisyankov:
I went and changed the code from the type 
```js
const messagesResult = getMessages(...);
```

borisyankov:
Also made all other minor changes. Should be ready.

gnprice:
Hmm, the destructuring does mitigate it. Still feels kind of awkward to me, but with this I think it'll be OK and I don't mind trying it out.

Smaller comments:

* The `messagesResponse` appears in an early commit, then gets fixed in a later one:
```
-  const [messagesResponse, streams] = await Promise.all([
+  const [{ messages }, { streams }] = await Promise.all([
```

* The `getUsers` commit has [this old comment](https://github.com/zulip/zulip-mobile/pull/2819#discussion_r206848322) apply. Also the commit message is a bit confusing by referring to "all params" when there's only one; made me wonder if there were more, and go check.

* The `getMessages` call is now one that would especially benefit from a link to the API doc, with all those new return properties.

* As usual, let's have the commit summaries start like `api: ` rather than `refactor:`.

borisyankov:
About the API docs links, have a look at this PR consisting of a single commit:
#3250

gnprice:
> About the API docs links, have a look at this PR consisting of a single commit:
> #3250

Yep, but that doesn't include `getMessages` which is the one I mentioned :slightly_smiling_face:

borisyankov:
I did:
 * rebase
 * made the change to `getMessages` earlier
 * added a link in `getMessages` to https://zulipchat.com/api/get-messages
 * `getUsers` have a better commit message

gnprice:
Thanks!

This comment still applies:

> * The `getUsers` commit has [this old comment](https://github.com/zulip/zulip-mobile/pull/2819#discussion_r206848322) apply.

I think I can merge up to that point, though.

gnprice:
(And thanks for the list of changes you made!)

borisyankov:
I did undo the change of adding `client_gravatar` as an argument to `getUsers`.

gnprice:
Merged -- thanks again @borisyankov !

PR #3248, opened 2019-01-01
AccountDetailsScreen: Adding flow type and updating to strict-local

VisibleMarkov:
Removed weakly typed reference and added `ConnnectorProps` type to `AccountDetailsScreen.js`. The file was updated with the tag `strict-local`.

Clarifications Required:
* Given `Props` was already being used as a type in the same document, I had to name the new type as `ConnectorProps`. If this can be improved, please let me know and I'll take care of it in a jiffy (^_^)

### DIFF
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -1,8 +1,9 @@
-/* @flow */
+/* @flow strict-local */
 import { connect } from 'react-redux';
 
 import React, { PureComponent } from 'react';
 
+import type { NavigationScreenProp } from 'react-navigation';
 import type { Dispatch, GlobalState, User, PresenceState } from '../types';
 import { getAccountDetailsUserFromEmail, getPresence } from '../selectors';
 import { Screen } from '../common';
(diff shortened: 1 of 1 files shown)


src/account-info/AccountDetailsScreen.js
```diff
@@ -33,7 +34,17 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connect((state: GlobalState, props: Object) => ({
+type ConnectorProps = {|
+  navigation: NavigationScreenProp<*> & {
+    state: {
+      params: {
+        email: string,
```
jainkuniya: Hey @VisibleMarkov Thanks!
This is not working as intended. Just replace it with `emails: string,` then run `npm run test:flow`, it should give error but it is not giving. In short, this change is not adding value, it is just fooling us that **code is under flow coverage, but isn't**.

It will be great if you cloud find some better way, so that we can get error if something goes wrong 👍

VisibleMarkov: Sorry for the delay at getting back.
I did reference other part of the code-base before  using `NavigationScreenProp<*>` and thought that was the way to go, however I didn't realize changing `email` to `emails` will not trigger an error.
I tried removing `NavigationScreenProp<*>` altogether and that didn't trigger any flow warning but in VSCode using the flow extension, it shows the part of the code to be redundant by graying it out so I thought that approach also fools flow in some way.
Trying to find a better way currently.

gnprice:
Thanks @VisibleMarkov ! And thanks @jainkuniya for the review. :slightly_smiling_face: 

I'll be interested to see what solution you find -- there are a number of spots in our code that use these `react-navigation`-related types, and I think a lot of them can be improved once we figure out how to write the types correctly. The kind of check that @jainkuniya described will be an important one in making sure any solution is a good one.

PR #3249, opened 2019-01-01
migrations: Fix regression introduced in `2` migration.

jainkuniya:
In `2` migration (c2ed78e7e6c9b46ed4c1cf0d0b66ae87b2b0ce40), we
dropped `result` & `msg` from the `state.relam.pushToken`. But now if
someone freshly installs the app, then there is not any initilization
of `realm` state, so it will throw undefined (as at the time of
applying `2` migration there is only `messages` & `narrows` in the
`state`, which are initialized in `1` migration). So make sure realm
state is present before accessing `state.realm.pushToken.token`.

### DIFF
--- a/src/redux-persist-migrate/index.js
+++ b/src/redux-persist-migrate/index.js
@@ -49,11 +49,10 @@ export default function createMigration(manifest, versionSelector, versionSetter
   const migrationDispatch = next => action => {
     if (action.type === REHYDRATE) {
       const incomingState = action.payload;
-      let incomingVersion = parseInt(versionSelector(incomingState), 10);
+      const incomingVersion = parseInt(versionSelector(incomingState), 10);
       if (Number.isNaN(incomingVersion)) {
-        incomingVersion = null;
-        // first launch after install, so initial state is empty object
-        // migration not required, just update verion
+        // first launch after install, so incoming state is empty object
+        // migration not required, just update version
         action.payload = versionSetter(incomingState, currentVersion);
         return next(action);
       }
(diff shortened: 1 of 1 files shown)


gnprice:
Hmm interesting -- thanks for catching this!

It doesn't feel to me like the right API for us to be running all these migrations at all in the case of a fresh install -- I'd rather fresh installs be governed just by the `initialState` in the various reducers, and the migrations only be relevant for updating old persisted state.

Fortunately, a while ago we took control of `redux-persist-migrate` by "vendoring" it into our own repo! So this is easy for us to fix there.

What if we do that, with something like this (which I haven't tested)?
```
diff --git src/redux-persist-migrate/index.js src/redux-persist-migrate/index.js
index caf3c467e..25732a4e3 100644
--- src/redux-persist-migrate/index.js
+++ src/redux-persist-migrate/index.js
@@ -49,12 +49,8 @@ export default function createMigration(manifest, versionSelector, versionSetter
   const migrationDispatch = next => action => {
     if (action.type === REHYDRATE) {
       const incomingState = action.payload;
-      let incomingVersion = parseInt(versionSelector(incomingState), 10);
-      if (Number.isNaN(incomingVersion)) {
-        incomingVersion = null;
-      }
-
-      if (incomingVersion !== currentVersion) {
+      const incomingVersion = parseInt(versionSelector(incomingState), 10);
+      if (!Number.isNaN(incomingVersion) && incomingVersion !== currentVersion) {
         const migratedState = migrate(incomingState, incomingVersion);
         action.payload = migratedState;
       }
```
Would you try that patch instead, and see if the issue still reproduces? Or fix it up if I've made a mistake you see how to fix.

jainkuniya:
Will try it out

gnprice:
Thanks @jainkuniya !

Hmm, looking harder at the code, I think this isn't 100% right either: it looks like it will mean new installs never get the version set in them, so they won't get future migrations either. So that needs adjustment.

gnprice:
Looks great, merged -- thanks @jainkuniya !

I did go and add more discussion to the commit message, as you suggested. :slightly_smiling_face: But it's telling basically the same story your version in 2e17012a5 was already telling.

PR #3250, opened 2019-01-02
api: Add JSDoc links to API calls

borisyankov:
Adds links to all API calls that have a corresponding online documentation page.


gnprice:
Thanks @borisyankov!

This isn't actually all of them, right? At a quick look at the left-sidebar nav on https://zulipchat.com/api/ and spot-checking the ones that look likely: there are several in messages/ and several in subscriptions/ that this doesn't cover. Possibly others; I haven't checked comprehensively.

I'll go merge this but with the commit message edited to not say it's all of them.

PR #3254, opened 2019-01-05
Logout if API key is invalid

### LINKED ISSUE #3253: Handle invalid API key by logging the user out
On a successful login, we receive an API key from the server to be used for API calls.
There are cases where this API key might become invalid.

Currently, one way to do that is to request a new API key from `Settings => Your account => Api key => Show/change your API key`.
There might be other events that might require the API key expiration (security reasons etc).

In any case, when the API key is no longer valid the app is not handling this at all.
There is no way for the app to recover from that state, it can't make API calls nor request a valid API key.

The correct behavior is to log out the user and let them login again.

borisyankov:
If a 4xx error is encountered when registering for an event, log out.

Fixes #3253

Details in each commit.

### DIFF
--- a/src/events/eventActions.js
+++ b/src/events/eventActions.js
@@ -3,6 +3,7 @@ import { batchActions } from 'redux-batched-actions';
 
 import type { Action, Dispatch, GeneralEvent, GetState, GlobalState } from '../types';
 import { pollForEvents } from '../api';
+import { logout } from '../account/accountActions';
 import { deadQueue } from '../session/sessionActions';
 import eventToAction from './eventToAction';
 import eventMiddleware from './eventMiddleware';
@@ -70,6 +71,12 @@ export const startEventPolling = (queueId: number, eventId: number) => async (
 
       lastEventId = Math.max.apply(null, [lastEventId, ...events.map(x => x.id)]);
     } catch (e) {
+      if (e.httpStatus === 401) {
+        // 401 Unauthorized -> our `auth` is invalid.  No use retrying.
+        dispatch(logout());
+        break;
+      }
+
       // protection from inadvertent DDOS
       await progressiveTimeout();
 
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -25,6 +25,7 @@ import { addToOutbox, sendOutbox } from '../outbox/outboxActions';
 import { realmInit } from '../realm/realmActions';
 import { reportPresence } from '../users/usersActions';
 import { startEventPolling } from '../events/eventActions';
+import { logout } from '../account/accountActions';
 
 const messageFetchStart = (narrow: Narrow, numBefore: number, numAfter: number): Action => ({
   type: MESSAGE_FETCH_START,
@@ -162,14 +163,22 @@ const fetchInitialData = () => async (dispatch: Dispatch, getState: GetState) =>
   dispatch(initialFetchStart());
   const auth = getAuth(getState());
 
-  const initData = await tryUntilSuccessful(() =>
-    registerForEvents(auth, {
-      fetch_event_types: config.serverDataOnStartup,
-      apply_markdown: true,
-      include_subscribers: false,
-      client_gravatar: true,
-    }),
-  );
+  let initData;
+  try {
+    initData = await tryUntilSuccessful(() =>
+      registerForEvents(auth, {
+        fetch_event_types: config.serverDataOnStartup,
+        apply_markdown: true,
+        include_subscribers: false,
+        client_gravatar: true,
+      }),
+    );
+  } catch (e) {
+    // This should only happen on a 4xx HTTP status, which should only
+    // happen when `auth` is no longer valid.  No use retrying; just log out.
+    dispatch(logout());
+    return;
+  }
 
   dispatch(realmInit(initData));
   dispatch(fetchTopMostNarrow());
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -159,7 +159,7 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState) 
   }
 };
 
-const fetchInitialData = () => async (dispatch: Dispatch, getState: GetState) => {
+export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
   dispatch(initialFetchStart());
   const auth = getAuth(getState());
 
@@ -193,11 +193,6 @@ const fetchInitialData = () => async (dispatch: Dispatch, getState: GetState) =>
   }
 
   dispatch(sendOutbox());
-};
-
-export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
-  dispatch(fetchInitialData());
-
   dispatch(initNotifications());
   dispatch(reportPresence());
   setInterval(() => dispatch(reportPresence()), 60 * 1000);
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -41,22 +41,4 @@ describe('tryUntilSuccessful', () => {
 
     expect(result).toEqual('hello');
   });
-
-  test('does not retry a call if maxRetries = 0', async () => {
-    let callCount = 0;
-    const thrower = () => {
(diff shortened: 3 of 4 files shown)


borisyankov:
No longer WIP. Ready for review + merge.

gnprice:
Thanks @borisyankov !

```
+  } catch (e) {
+    dispatch(logout());
+    return;
+  }
```

This will skip the rest of `fetchInitialData`, but there's still some things its caller `doInitialFetch` will do:
```
export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
  dispatch(fetchInitialData());

  dispatch(initNotifications());
  dispatch(reportPresence());
  setInterval(() => dispatch(reportPresence()), 60 * 1000);
};
```

The right solution might be to just fold those two functions together, like we recently did to form `fetchInitialData`. There is potentially a meaningful distinction to be made here ("fetch" vs. other stuff), but it's not clear it's a super helpful one -- and we're not totally observing it anyway, with e.g. `trySendMessages` in the "fetch" part.

Alternatively it could return true/false, or something else.

gnprice:
Also, to write down a comment I made in our call today: I think the meaning of `tryUntilSuccessful` gets a little complicated with this change, and also no longer matches the name.

One thing I think would be quite helpful is to try writing jsdoc for `tryUntilSuccessful` -- just a sentence or two saying what it means. That's helpful
(a) for the usual reasons -- to help people using it, or reading the code that uses it;
(b) because I think it will actually be challenging to write an accurate explanation in a reasonable amount of text, and that is a useful sign that the interface is getting complicated :stuck_out_tongue_winking_eye: .

borisyankov:
Rebased & ready.

gnprice:
Thanks @borisyankov !

####  "Use `progressiveTimeout` in `tryUntilSuccessful`"

```
-/* @flow strict */
+/* @flow */
```

We've made a lot of progress recently in eliminating `any` and `Object` and making our code `strict-local`. I want to make that a ratchet -- make a practice of never regressing on that.

In this case, I think the old version already solved the relevant challenges in finding the right generic types here -- it looks like that just got discarded in your rebase. So please clean that up.

####  "tryUntilSuccessful: Do not retry if error is 4xx (Client Error)"

* Thanks for adding jsdoc for this! The jsdoc should get squashed into the same commit that makes it necessary :wink: 

* Now that the description exists:

```
+/**
+ * Calls an async function and if unsuccessful retries the call.
+ *
+ * If the function is an API call and the response has HTTP status code 4xx
+ * the error is considered unrecoverable and the exception is rethrown, to be
+ * handled further up in the call stack.
+ */
 export const tryUntilSuccessful = async (func: any) => {
```

I think there's some evident tension in the semantics: the name, and the summary line, suggest something very general, but then the paragraph explains there's some much more specific logic that knows details about our API layer and the exceptions in its interface.

One next question is: currently this
* fails immediately on 4xx API errors;
* retries on other API errors;
* retries on any other errors unrelated to the API.

Is that quite the behavior we want? It seems like if there's an exception unrelated to an API call (e.g., the expression in an argument to an API call throws an exception), that's probably also non-temporary and shouldn't be retried. I think what we really want here is something like
* retry on network failures from `fetch`, and on 5xx errors from our API layer
* fail immediately on any other exception

I'll be OK merging this without that further change, but in that case I'll be very interested in a followup.

#### "Logout if registering for events throws 4xx"

* This logic looks good, thanks!

* I think this code is harder to understand than it has to be -- it's basically not possible to understand without going and reading either the history, or the `tryUntilSuccessful` jsdoc to learn the things that are in tension with the name. (Though the fact that information is in the jsdoc makes a big improvement over reading the implementation and puzzling over why it doesn't seem to match the name.)

  A local fix would be to take some of the commit message's info and move it into the `catch` block: say that this should only happen on 4xx, which should only happen when our auth is no longer good.

  Potentially with a different name for `tryUntilSuccessful` this would be clear with no comment, which would be even better :slightly_smiling_face: 

* There's one other important spot to have logic like this: around our call to `pollForEvents`, in `startEventPolling`. That way we'll learn about an invalid API key before too long even if the app keeps running for a while.

#### "Merge `fetchInitialData` and `doInitialFetch`"

Looks good, thanks!

borisyankov:
Updated the PR. Here are the changes:

>  ```
> -/* @flow strict */
> +/* @flow */
> ```

This inadvertent regression in type strictness.

> The jsdoc should get squashed into the same commit that makes it necessary 😉

Done.

borisyankov:
I did experiment with several approaches suggested in the review.

Following are comments on other work that is in-progress or needs further discussion

borisyankov:
> Is that quite the behavior we want? It seems like if there's an exception unrelated to an API call (e.g., the expression in an argument to an API call throws an exception), that's probably also non-temporary and shouldn't be retried. I think what we really want here is something like
> 
> * retry on network failures from fetch, and on 5xx errors from our API layer
> * fail immediately on any other exception
>
> I'll be OK merging this without that further change, but in that case I'll be very interested in a followup.

I agree that we should differentiate between them.
It is a bit tricky to differentiate currently, I did explore several ways to do that.
Opening a new issue to track all thesе aspects you commented on that are not a blocker to merge this.
#3482

borisyankov:
> not possible to understand without going and reading either the history, or the tryUntilSuccessful jsdoc to learn the things that are in tension with the name. 

I think some refactoring of the 'tryUntilSuccessful` function is a good idea. Not a blocker for merging this though. I have opened a new issue #3483 to track further improvements to this function.

borisyankov:
> Is that quite the behavior we want? It seems like if there's an exception unrelated to an API call (e.g., the expression in an argument to an API call throws an exception), that's probably also non-temporary and shouldn't be retried.

I'd say long-term we want to improve that with #3482 or #3483 

Short-term - yes, we absolutely do want that.

We currently use that function in only two cases:
 * the initial `register` call
 * the initial `fetchPrivateMessages`

Both should not fail. If they do, we are in an unrecoverable situation (especially the first call).

A temporary failing back-end might respond with html content or something weird. We don't want to mistake that for something unrecoverable. I want the default behavior in 'uncharted territories' to be retry and to err on the side of retrying, which is accompanied by a timeout now.

gnprice:
Thanks @borisyankov for these small fixes!

Would you please act on the two other comments quoted below?

> * I think this code is harder to understand than it has to be -- it's basically not possible to understand without going and reading either the history, or the `tryUntilSuccessful` jsdoc to learn the things that are in tension with the name.

You quote this one and then refer to not blocking on a refactoring of `tryUntilSuccessful`. But I specifically offered two possible solutions, neither of which is that:

>   A local fix would be to take some of the commit message's info and move it into the `catch` block: say that this should only happen on 4xx, which should only happen when our auth is no longer good.
>
>   Potentially with a different name for `tryUntilSuccessful` this would be clear with no comment, which would be even better slightly_smiling_face

Even if you can't think of a more accurate name for `tryUntilSuccessful`, you can definitely move the info into a comment.


> * There's one other important spot to have logic like this: around our call to `pollForEvents`, in `startEventPolling`. That way we'll learn about an invalid API key before too long even if the app keeps running for a while.

This seems easy to do, and just as relevant as when I originally made this comment.

borisyankov:
It was hard to express why I considered this non-blocking, it was because of the follow-up commits would have moved the target for the comments elsewhere.

I thought the work will be done later but found out it clarifies the code lot and improves it enough that I did it next.

Referring to this PR #3484 

This is the commit I had in mind when thinking the documentation can wait:
https://github.com/zulip/zulip-mobile/pull/3484/commits/dda3e41e42a03fe49a75a44b3c7e7aefbb6f463b

Now the logic is in a separate function and the function itself is documented.

gnprice:
Cool, thanks for the extra context.

I took a quick look at #3484, and it doesn't appear to address either of my two comments above:
* It doesn't touch that confusing code added by the "Logout if registering for events throws 4xx"; in fact it doesn't touch that file, `fetchActions.js`, at all. 
* It doesn't add anything around `pollForEvents` to detect an auth failure there.

I went ahead and made those fixes to this branch, in a minimal way, including a new commit for the `pollForEvents` / `startEventPolling` change. Merging with those adjustments. Thanks again for these changes!

PR #3255, opened 2019-01-05
WIP: More precise (and more correct!) loading state

borisyankov:


chrisbobbe:
Hmm, looks like this PR has become stale and it may be time to close it; what are your thoughts on this, @borisyankov, @gnprice?

Specifically, it looks like `state.loading` was folded into `state.session.loading` in https://github.com/zulip/zulip-mobile/commit/e9ffd5e910e266b5ebbd69ec94bb7a92df2d350e#diff-2ec6b30ae213c723ddfa3f4d08b88e36.

`state.session.loading` is true during the initial fetch (INITIAL_FETCH_START to INITIAL_FETCH_COMPLETE), with DEAD_QUEUE, LOGOUT, and ACCOUNT_SWITCH all setting it to false.

loadingReducers.js was renamed to loadingReducer in https://github.com/zulip/zulip-mobile/commit/e137e29b5d80ea3c534192615d5b2b9dd5f38630#diff-28bdd53c7f1b8958a719c62f6bff5c4b and removed in the commit linked above.

On the other hand, PR would still have added a convenient, central way to set loading state for any string we want (action.entity), with the most obvious example being a narrow, I think, e.g.,

```
{
    [NARROW_STR_1]: false,
    [NARROW_STR_2]: true,
    // (etc.)
}
```

But `state.fetching` is already a good source of truth for all the loading state specific to a narrow, I believe, including the required .older and .newer parts. This couldn't coexist with a structure like the one proposed without a good chance for bugs to be introduced (keeping them both in sync, etc.), and what we have now seems to work well.

Are there other loading states that would benefit from something like what this PR provides?

chrisbobbe:
Oops, I see I've misunderstood this PR a bit -- editing now.

chrisbobbe:
Closing as stale; please let me know if I’ve done this in error.

gnprice:
Thanks @chrisbobbe! I agree with your reasoning.

PR #3257, opened 2019-01-06
PR #3231 follow-up. settings: Build a better UI for choosing language.

anh2111htd:
- Let's sort the languages, by locale, rather than put English first and the rest in English alpha order.
- When I search, then try to make a choice, it seems to take two taps -- one to dismiss the keyboard, the next to actually choose it. Let's make it take just one tap.

### DIFF
--- a/src/settings/LanguagePicker.js
+++ b/src/settings/LanguagePicker.js
@@ -52,7 +52,7 @@ export default class LanguagePicker extends PureComponent<Props> {
     return (
       <FlatList
         ItemSeparatorComponent={OptionDivider}
-        initialNumToRender={languages.length}
+        initialNumToRender={15}
         data={data}
         keyboardShouldPersistTaps="always"
         keyExtractor={item => item.tag}


gnprice:
Thanks @anh2111htd for this (long-ago) contribution! Sorry it got overlooked.

(For cross-reference: this was a follow-up to #3231. GitHub linkifies PR/issue references in comments, but not in titles.)



> * When I search, then try to make a choice, it seems to take two taps -- one to dismiss the keyboard, the next to actually choose it. Let's make it take just one tap.

This was a good change. It looks like we ended up making it in #3260, around the same time as this PR.


> * Let's sort the languages, by locale, rather than put English first and the rest in English alpha order.

We fixed this much later, in 1a03dc7477359d51a89df0c0989d65d68a49de33 / #4285. We did it a bit differently: it's always sorted by each language's own name for itself, which seems to be standard in these UIs.


> Prop initialNumToRender was added to fix an issue described in #1350. Since the issue no longer appears when taking initialNumToRender back out, we remove this prop to get the normal benefit of FlatList.

Huh yeah, wacky -- good find.

The [docs say](https://reactnative.dev/docs/flatlist#initialnumtorender):
> This should be enough to fill the screen but not much more.

On a moderately tall test device, I get 11 items on the screen, and the default for this prop is 10. So we do actually want somewhat more than the default. I'll make it 15, and merge.

PR #3258, opened 2019-01-06
AccountDetailsScreen: Aded and tested flow typing and updated to strict-local

Kriyszig:
Previous PR: #3248 
The previous PR was flawed as it fooled flow into believing the code was under flow coverage when it wasn't.
I followed this [comment](https://github.com/react-navigation/react-navigation/issues/912#issuecomment-327032278) in the `react-navigation` repository where the state is passed as the parameter rather than using intersection types in flow. 
This passed flow test and even gave an error when I changes `email` to `emails`, thus proving to an extent the code is covered under flow.
I would love your reviews on this change - I presume it was the using of `&` operator that led to the weird behavior.
I also did some further digging and found out that this problem persists even in `/src/lightbox/LightboxScreen.js` where if we change `params` to `param`, flow will not throw any errors.

If this implementation is fine, I would be more than happy to change the usage of `NavigationScreenProp` throughout the code-base.

### DIFF
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -1,8 +1,9 @@
-/* @flow */
+/* @flow strict-local */
 import { connect } from 'react-redux';
 
 import React, { PureComponent } from 'react';
 
+import type { NavigationScreenProp } from 'react-navigation';
 import type { Dispatch, GlobalState, User, PresenceState } from '../types';
 import { getAccountDetailsUserFromEmail, getPresence } from '../selectors';
 import { Screen } from '../common';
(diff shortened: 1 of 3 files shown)


VisibleMarkov:
@gnprice, Would love your review on this PR (^_^)

VisibleMarkov:
Thank you for the Green flag @jainkuniya, Sorry about the commit massage, will amend it ASAP. Thank you (^_^)

VisibleMarkov:
The LightBox and ChatScreen implementations are hard to convert this was as the previous flow type definition of the value `key` comes in direct conflict with one of our tests. So far, these two work as expected. I have been searching for ways to override previously defined types in flow but I haven't found any luck at that so far.

VisibleMarkov:
P.S. This PR can be merged now because I haven't found a way around the previously defined types in flow-typed. If there are any changes required please let me know @gnprice

PR #3260, opened 2019-01-08
languagePicker, search: Fix language is not selected on single tap.

jainkuniya:
List was capturing first tap to hide keyboard. So set
`keyboardShouldPersistTaps` as `always` in `FlatList`, to avoid
capturing of tap event by list.

References:
https://facebook.github.io/react-native/docs/scrollview#keyboardshouldpersisttaps

### DIFF
--- a/src/settings/LanguagePicker.js
+++ b/src/settings/LanguagePicker.js
@@ -60,6 +60,7 @@ export default class LanguagePicker extends PureComponent<Props> {
         ItemSeparatorComponent={() => <View style={styles.separator} />}
         initialNumToRender={languages.length}
         data={data}
+        keyboardShouldPersistTaps="always"
         keyExtractor={item => item.locale}
         renderItem={({ item }) => (
           <LanguagePickerItem


gnprice:
Thanks @jainkuniya ! This change looks good -- about to merge.

This explanation confused me for a bit, as it seemed to be saying something that's backward from what I now think you actually meant:

> List was capturing first tap to hide keyboard. So set
> `keyboardShouldPersistTaps` as `always` in `FlatList`, to avoid
> capturing of tap event by list.

The symptom was that the list *wasn't* responding to the tap, to do its thing of selecting a new item in the list. So I read "capturing of tap event by list" to refer to the thing that *wasn't* happening, and that we wanted to happen -- which made it confusing why you were talking about avoiding it. :wink: 

Reading how this logic is implemented (using a command `rg -C20 keyboardShouldPersist ../react-native/` to look in my local clone of the RN tree), I see that the default behavior involves a `ScrollView` component, in code from a mixin called `ScrollResponder`, capturing the first tap.

So I think you were saying something like "the list was capturing the tap event, preventing the list item from getting it". The distinction here between the list and its items is a pretty subtle one, so one way to make that clearer would be to make the contrast explicit: "The list was capturing the first tap to hide the keyboard, preventing the list item from getting the tap to make a selection." (but see next point)

Also, a more precise reference helps: it's the `ScrollView`, more specifically than "the list" in general.

Here's the version I ended up with:

```
language picker: Let first tap make a selection and dismiss keyboard.

When the search box was focused and the keyboard visible, tapping a
language would just unfocus the input and hide the keyboard, without
actually selecting the new language; you'd have to tap a second time
to make a selection.

This is the default behavior of `ScrollView`, found inside the
`FlatList` we use to show the languages.

Instead, set `keyboardShouldPersistTaps` to `always`.  This disables
`ScrollView`'s handling of the tap, letting our list item receive it.
  https://facebook.github.io/react-native/docs/scrollview#keyboardshouldpersisttaps
```

Plus a note on a curious additional fact:

    (The search box still gets unfocused and the keyboard hidden, even
    though `ScrollView` is no longer doing anything to cause that.  This
    seems to be triggered by the settings change... ah, and it's caused
    by the hack in `TranslationProvider` described in a comment there,
    where we change `key` when the locale changes.)

In other words, the behavior after this change doesn't actually match what the documentation for `keyboardShouldPersistTaps="always"` says should happen! Did you notice that? :wink:

I noticed the discrepancy when trying to understand what the docs were saying, and how that related to what your commit message said and what the code was intended to do. After I noticed the discrepancy, I wanted to be sure to understand it because it could easily have meant we didn't yet understand what `keyboardShouldPersistTaps` actually did.

PR #3262, opened 2019-01-08
Load 'streams' data via a 'register' call

borisyankov:
Avoids a race condition when doing the `register` call.


borisyankov:
Code-wise this PR is done and ready for review/discussion.

borisyankov:
Updated and rebased. Ready for a review.

borisyankov:
Rebased & ready.

gnprice:
Thanks @borisyankov ! I like this change.

Would you add a commit to remove the INIT_STREAMS action type?

I think it's important to keep the codebase clear of things that no longer play a role in how the app works -- we've had in the past a habit of leaving those around, and I've found them to be a recurring source of confusion. So I'd like to make a habit of taking care of that in the same PR where we make the code dead in the first place, because that's the easiest time to do it and by far the easiest time to remember to do it.

(The API binding `getStreams` is different, though -- our API layer is explicitly intended to describe the server API, not only to serve our app's current needs, so it's fine to leave things there we don't currently use.)

borisyankov:
Done. Added another commit with cleanup.

gnprice:
Thanks @borisyankov for the revision! Looks great.

Merged, with a couple of small additional commits: your cleanup of INIT_STREAMS drew attention to some parts of the `loading` state that are also unused, so I followed that trail and then took those out too.

PR #3263, opened 2019-01-10
Away status

borisyankov:
Implements #3266


borisyankov:
While the PR itself is WIP (waiting for clarification and discussion on some back-end functionality) all three current commits can be merged already.

gnprice:
Thanks @borisyankov !

There are some rebase conflicts with the refactors to action and event types that I made last week. I think all of the actual conflicts will be easy to resolve.

One thing that would be great is in the commit that does this:
```
+export type EventUserStatusUpdateAction = {|
+  ...ServerEvent,
+  type: typeof EVENT_USER_STATUS_UPDATE,
+  away: boolean,
+  user_id: number,
+|};
+
```
let's use the new `src/api/eventTypes.js` to write down exactly what the server actually sends in these events. Then the action type can refer to that -- much like `EventPresenceAction` does now.

gnprice:
(In general, whenever we're adding a new event type -- or for another reason doing the work of nailing down exactly what the server sends in a given type of event -- I'd like to use the new `eventTypes.js` to write that information down while it's fresh.)

gnprice:
(It looks like my comments above about `eventTypes.js` still apply.)

src/api/setAwayStatus.js
```diff
@@ -0,0 +1,11 @@
...
+import type { ApiResponseSuccess, Auth } from './apiTypes';
+import { apiPost, objectToParams } from './apiFetch';
+
+type SetAwayStatusParams = {
+  away?: boolean,
+  status_text?: string,
+};
+
+export default (auth: Auth, params: SetAwayStatusParams): Promise<ApiResponseSuccess> =>
+  apiPost(auth, 'users/me/status', objectToParams(params));
```
gnprice: What's the reasoning for using `objectToParams` here?

I believe it should have no effect at all -- neither of these properties are permitted to be an array, or `undefined`. So it seems simpler to just say `params`.

borisyankov:
This is ready for review.

src/api/setAwayStatus.js
borisyankov: The function is used for its ability to remove keys with `undefined` values.

gnprice: This isn't much of a reply. I wrote above:

> I believe it should have no effect at all -- neither of these properties are permitted to be an array, or `undefined`.

If that's mistaken, would you explain why? If not, why do you take steps to remove keys with `undefined` values?

src/events/eventToAction.js
```diff
@@ -145,6 +146,12 @@ export default (state: GlobalState, event: Object): EventAction => {
         type: opToActionUserGroup[event.op],
       };
 
+    case 'user_status':
+      return {
+        ...event,
+        type: EVENT_USER_STATUS_UPDATE,
+      };
```
gnprice: Take a look at the block of 8 event types up above, with `actionTypeOfEventType`. This should go there.

src/api/eventTypes.js
```diff
@@ -110,6 +110,13 @@ export type PresenceEvent = {|
   presence: UserPresence,
 |};
 
+export type UserStatusEvent = {|
+  ...EventCommon,
+  user_id: number,
+  away?: boolean,
+  status_text?: string,
```
gnprice: Huh interesting! Are both of these really optional?

Also: Is your description here based on empirical observation, or reading the server code, or discussions with Steve, or something else?

src/api/setAwayStatus.js
gnprice: Aha, I see in a later commit that you do pass it params objects that have `undefined` values. I was confused to see that Flow accepted that; it turns out that the documented semantics of `?:` properties is [like so](https://flow.org/en/docs/types/objects/#toc-optional-object-type-properties):
> In addition to their set value type, these optional properties can either be `void` or omitted altogether. However, they cannot be `null`.

So indeed this is doing something useful, then.

(This reply remains not much of a reply.)

src/directSelectors.js
```diff
@@ -54,6 +55,8 @@ export const getTopics = (state: GlobalState): TopicsState => state.topics;
 
 export const getUserGroups = (state: GlobalState): UserGroup[] => state.userGroups;
 
+export const getUserStatus = (state: GlobalState): UserStatusState => state.userStatus;
```
gnprice: This should go in the same commit that adds this branch to the state tree.

src/user-status/userStatusReducers.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default (state: UserStatusState = initialState, action: Action): UserStatusState => {
+  switch (action.type) {
+    case LOGOUT:
+    case LOGIN_SUCCESS:
+    case ACCOUNT_SWITCH:
+      return initialState;
+
+    case REALM_INIT:
+      return action.data.user_status || initialState;
```
gnprice: The commit message points out:

    Note: Older back-ends do not have a `user_status` key and we handle
    this correctly by leaving the state in a valid, empty state.

Quite so! This is an important bit of logic to have.

As we've discussed recently, if it's important enough to have code for, it's important enough to have in the types. The type should look like `user_status?: UserStatusMap`. Ideally with a line of jsdoc saying what this commit message says: that older servers (older than what version?) don't send that property.

src/user-status/userStatusReducers.js
```diff
@@ -0,0 +1,41 @@
...
+
+    case EVENT_USER_STATUS_UPDATE: {
+      const newUserStatus = { ...state[action.user_id] } || {};
```
gnprice: Can this `|| {}` ever do anything?

Note that an empty object is already truthy:
```
> ({} || false)
{}
```

src/user-status/userStatusReducers.js
```diff
@@ -0,0 +1,41 @@
...
+      if (action.away !== undefined) {
+        newUserStatus.away = action.away;
+      }
+      if (action.status_text !== undefined) {
+        newUserStatus.status_text = action.status_text;
+      }
+      return {
+        ...state,
+        [action.user_id]: newUserStatus,
+      };
```
gnprice: Hmm, this is all kind of awkwardly verbose.

How about we let the values in our state be `undefined` rather than absent? Then we can write something like
```
const status = state[action.user_id] || {};
return {
  ...state,
  [action.user_id]: {
    away: action.away !== undefined ? action.away : status.away,
    status_text: action.status_text !== undefined ? action.status_text : status.status_text,
  },
};
```

src/types.js
```diff
@@ -129,6 +129,17 @@ export type UserIdMap = {
   [userId: string]: User,
 };
 
+export type UserStatus = {|
+  away?: boolean,
+  status_text?: string,
```
gnprice: What's the semantics of this?

In particular, how does `away: undefined`, or a missing `away`, relate to `away: false` or `away: true`?

And is there a distinction in the meanings of `away: undefined` vs. a missing `away`? (And same question for `user_status`.)

src/user-status/userStatusReducers.js
gnprice: (I don't love that version either, but it feels a little more compact and direct.)

src/user-status/userStatusReducers.js
```diff
@@ -0,0 +1,41 @@
...
+import {
+  LOGOUT,
+  LOGIN_SUCCESS,
+  ACCOUNT_SWITCH,
+  REALM_INIT,
+  EVENT_USER_STATUS_UPDATE,
+} from '../actionConstants';
+import { NULL_OBJECT } from '../nullObjects';
+
+const initialState: UserStatusState = NULL_OBJECT;
```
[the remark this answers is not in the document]
borisyankov: Completely disagree because:
 * we are currently using this pattern everywhere
 * there are comparisons happening, any selector that uses that or any component that uses that as props does a comparison against the old value
 * as a bonus this object is frozen and will throw if we mutate it, revealing a potential bug

src/user-status/userStatusReducers.js
borisyankov: This code remained from a version that was like `state[action.user_id] || {}`. Fixed.

src/events/eventToAction.js
borisyankov: Oh, didn't notice these `case` refactorings.
I am not sure this is an improvement, removing simple duplication by increasing the complexity of the code...
Anyway, did the change.

src/api/eventTypes.js
```diff
@@ -110,6 +110,13 @@ export type PresenceEvent = {|
   presence: UserPresence,
 |};
 
+export type UserStatusEvent = {|
```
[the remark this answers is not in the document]
gnprice: Cool, I see this is done in an updated version of the branch. Small thing there -- the pattern I've set in this `EventTypes` enum is for the names to exactly match the values:
```
  static update_message_flags: 'update_message_flags' = 'update_message_flags';
```
I'd like to continue that pattern here, so `EventTypes.user_status`.

src/user-status/userStatusReducers.js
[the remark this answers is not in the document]
gnprice: > there are comparisons happening, any selector that uses that or any component that uses that as props does a comparison against the old value

Right, but that's perfectly well served by `const initialState: UserStatusState = {};`. 

(That's why I wrote "with some *other* unrelated object that might happen to be empty".)

If there's ever a place where we compare one of these with a value that comes from some *other* mention of `NULL_OBJECT` -- like some other reducer for a different part of the state tree -- that would indicate a very worrying bug.


> we are currently using this pattern everywhere

Hmm, true:
```
$ perl -lne 'print $1 if (/^const initialState.* = (.*)/)' src/*/*Reducer*.js | sort | uniq -c
      5 {
      1 getStateForRoute('loading') || NULL_NAV_STATE;
     12 NULL_ARRAY;
      9 NULL_OBJECT;
```

So *shrug* I guess it's fine for this PR to go along with the pattern, then. I might go through at some point and see about changing them all together, though.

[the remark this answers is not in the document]
gnprice: > as a bonus this object is frozen and will throw if we mutate it, revealing a potential bug

This is certainly something that could be useful in general!

I am not sure there is much value, though, in only having it on the empty values. I'd rate it a lot more likely to catch bugs if we were to do it systematically, on all the objects and arrays in our Redux state trees. The Reselect cached selector values would also be a good target.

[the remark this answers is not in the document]
borisyankov: I suspect the operation on deep object is slow as `Object.freeze()` does not work deeply and the `deepFreeze` we use in the unit-tests is iterating children calling `Object.freeze()` on them if array or object.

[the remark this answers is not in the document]
borisyankov: It seems the source code for that is shorter than an explanation:
https://github.com/substack/deep-freeze/blob/master/index.js

src/types.js
borisyankov: As other API calls both for input/output we can specify one or more of these at once. Currently, there are two but maybe more of these would be added in future. So any combination of the keys existing is possible except for no keys specified.

src/user-status/userStatusReducers.js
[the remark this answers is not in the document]
gnprice: > It seems the source code for that is shorter than an explanation

Like so much of the JS library ecosystem. :stuck_out_tongue_closed_eyes: Thanks for the link!

Yeah, that implementation may not be a suitable one for normal use. I think a good version of this would probably have to look something like: freeze each object/array as we create it, and count on its property values having been respectively frozen when they were created, without trying to iterate through them to double-check that.

Then the one potentially tricky bit is finding a way to spell that that doesn't make all our reducers (and cached selectors) annoyingly verbose.

src/types.js
gnprice: OK, but my question was about the semantics -- what does it *mean* for `away` to be `true`, vs. `false`, vs. `undefined`, vs. not present? 

In particular, is there a difference in meaning between `undefined` and not present, for either property? And for `away`, does `undefined` mean the same thing as `false`, or what is the contrast in their meanings?

gnprice: Also, note this is a type definition in `src/types.js`, which we use for our own internal data structures. What the same-named parameters mean in the API call to the server to *set* this user's own status is a different question. (One I'd also be glad to have a clear answer to! But that one I feel like I have a more confident guess at than this one -- I'm guessing/hoping that when a parameter is left out there, it simply doesn't change the value the server has for that field.)

src/types.js
```diff
@@ -706,6 +718,10 @@ export type InitialDataUpdateMessageFlags = {|
   },
 |};
 
+export type InitialDataUserStatus = {|
+  user_status: UserStatusMap,
```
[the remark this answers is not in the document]
borisyankov: I did update the commit message and made the `user_status` value optional to reflect that it might not exist.

src/api/eventTypes.js
borisyankov: Based on all three. One or two of them exist. Never none (as expected).

src/types.js
borisyankov: This is actually API-related, as the `initialData` structures are dependent on the back-end and in another PR these types are moved to `apiTypes` (maybe closed)

borisyankov: I did change the `away` type to `away?: true,` which is the precise type. The value never exists as `false`.

src/user-status/userStatusReducers.js
borisyankov: This code has some drawbacks.
If previously we hadn't set `away` or `status_text` it will now be set to `undefined`.

I am going in the other direction. The value `away` will never be `false` but will miss completely, `status_text` will never be `null`, `undefined` or an empty string with length=0 but will also be deleted.

gnprice: > If previously we hadn't set `away` or `status_text` it will now be set to `undefined`.

Well yes, that is what I said :slightly_smiling_face: :

> How about we let the values in our state be `undefined` rather than absent?

More precisely, that would mean that each of those properties would always be set, whether to `undefined` or to something else.


> I am going in the other direction. The value `away` will never be `false` but will miss completely, `status_text` will never be `null`, `undefined` or an empty string with length=0 but will also be deleted.

OK, cool -- I'll look at the resulting code shortly.

src/events/eventToAction.js
gnprice: Cool.

> I am not sure this is an improvement, removing simple duplication by increasing the complexity of the code...

I think the main thing I liked about this particular small refactoring was that we had all these different cases which were doing essentially the same thing... but that wasn't obvious because they were scattered around this `switch` statement and interleaved with cases that worked differently. And so each of them added complexity as you had to look at each case individually to see what it did. Now it's very clear that these all have the exact same boring structure, and the exceptions which do something more interesting stand out better.

There are other ways that could be done. Simply reordering cases to bring these together would have accomplished that and certainly not added any complexity, and would have been fine though remained verbose. When I went to make it less verbose, I chose this particular style largely because... this is what the existing code in the file already does for a bunch of other cases! :slightly_smiling_face:

src/api/eventTypes.js
gnprice: Ah cool, I see that the `UserStatus` types in the app's own data structures now has some quite helpful jsdoc, and of course that overlaps with this information.

No need to duplicate info. Any given piece of information can be written down in one of the places where it's important, and then write "See also `UserStatus` in src/types.js" or whatever in each of the other places.

I think the key bit of info that that doesn't cover and is important for this event type is: what does it mean when one of these is missing? Also `away: false`, which can't occur in the other place.

(Ideally I think the information on that `UserStatus` type would be in `api/` somewhere -- particularly as we seem to use it in `InitialData`, which is really mainly a matter of describing the API. But it's fine to defer doing that rearrangement.)

src/types.js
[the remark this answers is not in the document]
gnprice: Thanks! I appreciate the jsdoc on the `UserStatus` type, too.

FWIW my inclination would be to leave the description of `user_status` out of the commit message, as it's covered quite nicely in the jsdoc.

src/types.js
gnprice: Thanks for the jsdoc on this type, which answers most of my questions!

> This is actually API-related, as the `initialData` structures are dependent on the back-end

Hmm, it's true that we're using the same type to describe that structure. 

I think that doesn't really change most of what I discussed in my comments above, though. It's fine and good to reuse the same type in our Redux state's type as in describing parts of the server API, where that makes sense... but it doesn't relieve us of our responsibility for defining what the objects in our Redux state tree are supposed to mean. Which the jsdoc you've added largely does. :smile: 

(I have a specific follow-up question or two, which I'll ask as a separate comment-thread.)

src/api/eventTypes.js
```diff
@@ -17,7 +17,7 @@ import type { Message, Stream } from './apiTypes';
 //
 
 /** See ClientPresence, and the doc linked there. */
-export type UserStatus = 'active' | 'idle' | 'offline';
+export type PresenceStatus = 'active' | 'idle' | 'offline';
```
gnprice: This is a good rename! I think in fact it belongs at the very beginning of the branch -- then that properly clears things up for starting to use the term "user status" for this new feature.

Ditto the other related rename.

src/user-status/userStatusReducers.js
gnprice: Thanks for adjusting the type!

I'd still like to see this part too:
> Ideally with a line of jsdoc saying what this commit message says: that older servers (older than what version?) don't send that property.

src/types.js
```diff
@@ -126,9 +126,29 @@ export type MessagesState = {|
...
 };
 
+/**
+ * Specifies user status related properties
+ * @prop away - present if we are to override user's presence status
+ *       * when `true` the user is always `offline`
+ *       * when missing the presence is not changed
+ *       * can not be `false`
+ * @prop status_text - a string representing information the user decided to
+ *   manually set as his 'current status'
```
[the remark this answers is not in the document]
gnprice: Aha, I see the commit message for the later commit "Handle 'EVENT_USER_STATUS_UPDATE' action" has this information!

    Currently the web app does update the `away` value separately from
    the `status_text` string and that is the immediate plan for mobile
    too. It is possible though that both values are changed at once.
    
     * if an event contains `away: false` this has the meaning of 'clear
       away overriding'
     * if an event contains `status_text: ''` this means 'clear status text'

    There is value in keeping our internal state simpler and consistent
    with what the back-end gives us as values. To achieve this we make
    sure if we get `away: false` and `status_text: ''` we remove the
    corresponding keys.

So it sounds like the answer is:
* can't be `undefined`, nor `''`
* instead, will just be absent.

This is essential info for understanding the code around this type. Not only the reducer which maintains it, but also code which consumes it. And -- key test for what belongs in the code + comments, and not only commit message -- it's not at all specific to the *change* between this commit and its parent, and instead is 100% relevant to understanding the new code from scratch as it is, and for any future version of the codebase where these bits of code looks more or less like as they do here.

So let's stick that in the jsdoc too.

(There's also some information in there answering my questions about the semantics of the events. The same thing applies.)

src/directSelectors.js
gnprice: bump

src/user-status/userStatusReducers.js
gnprice: OK, now that I see the intended semantics for these (`away` has just two states, and `status_text` as empty-string is synonymous with missing), I think probably the cleanest thing would actually be to have a type like this in our Redux state tree: `{| away: boolean, status_text: string |}`. Then this code would look like
```
/** The status a user implicitly has when not present in the state. */
export const DEFAULT_STATUS = { away: false, status_text: '' };
// ...

const status = state[action.user_id] || DEFAULT_STATUS;
return {
  ...state,
  [action.user_id]: {
    away: action.away !== undefined ? action.away : status.away,
    status_text: action.status_text !== undefined ? action.status_text : status.status_text,
  },
};
```

One advantage of this would be that it makes explicit what it means when we don't have this kind of data on a given user -- unlike for many other kinds of data where it would mean "we don't know", here it means "they have these default values". The `export` is there for a selector to use to encapsulate that fact.

I'm OK merging a version with the existing design, though.

gnprice:
OK, and just merged 5 of these commits, as ac71a733d..f4283ee16!

I made just one edit to their code, along with small edits to the commit messages -- fixed one straggler mention of "user status" to mean a user's presence status:
```
diff --git src/account-info/AccountDetails.js src/account-info/AccountDetails.js
index f06dcdfb1..66833bb24 100644
--- src/account-info/AccountDetails.js
+++ src/account-info/AccountDetails.js
@@ -17,7 +17,7 @@ const componentStyles = StyleSheet.create({
   componentListItem: {
     alignItems: 'center',
   },
-  userStatusWrapper: {
+  statusWrapper: {
     justifyContent: 'center',
     flexDirection: 'row',
   },
@@ -49,7 +49,7 @@ export default class AccountDetails extends PureComponent<Props, void> {
           shape="square"
         />
         <ComponentList outerSpacing itemStyle={componentStyles.componentListItem}>
-          <View style={componentStyles.userStatusWrapper}>
+          <View style={componentStyles.statusWrapper}>
             <PresenceStatusIndicator presence={presence} hideIfOffline={false} />
             <RawLabel style={[styles.largerText, styles.halfMarginLeft]} text={user.email} />
           </View>
```

Found that with `git grep -iE 'user.?status'`.

gnprice:
(And for convenience, I just rebased the remaining 6 commits on top of those, and pushed that to the PR branch.)

src/api/eventTypes.js
borisyankov: Indeed. I had a commit somewhere that I think I dropped because it was causing too many merge conflicts that moved InitialData and all related types to `apiTypes`.

src/user-status/userStatusReducers.js
borisyankov: I don't even know what the next version will be called and the current official version does not support it. ¯\(ツ)/¯

src/api/eventTypes.js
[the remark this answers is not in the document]
borisyankov: Updated, haven't noticed that.
In lights of this, I think this is just a way more verbose version of  the type:
`update_message_flags` | `user_status`
which will also not allow us to specify an invalid value and hopefully will provide auto-suggest as well.

borisyankov:
Updated.

src/api/eventTypes.js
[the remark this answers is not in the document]
gnprice: Hmm, I don't think I'm quite picturing what alternative code you have in mind. Want to send a small PR to demonstrate? Happy to experiment with other ways of writing this.

gnprice:
Merged -- thanks again @borisyankov !

I added under `InitialDataUserStatus` the comment discussed in a thread above.

I also squashed together the commit that asks the server to send that initial data, and the one that processes it in the reducer. Those feel to me like they each make more sense to understand in combination with the other, and the "request initial data" one especially is simple enough that the squashed one isn't so long as to get in the way of reading.

PR #3267, opened 2019-01-11
Profile tab and UI for setting away status

borisyankov:
Based on #3263

Includes a new Profile card, and UI for setting 'away' and 'status text'.


gnprice:
And #3263 is merged!

borisyankov:
The first 5 commits are ready for review (one is the same as the other PR)
The rest could be reviewed but are still being changed a bit.

src/common/OwnAvatar.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Renders an image of the current user's avatar
+ *
+ * @prop [size] - Sets width and height in pixels.
+ * @prop [user] - Current presence for this user used to determine status.
+ * @prop [realm] - Current realm url, used if avatarUrl is relative.
+ * @prop [shape] - One of 'square', 'rounded', 'circle'.
+ */
+class OwnAvatar extends PureComponent<Props> {
```
gnprice: Huh. What's the motivation for adding this component? I.e., why not just say `<Avatar user={ownUser} ...`?

I'm not a fan of this duplication -- there are a bunch of details like "`shape` defaults to `rounded`", "`avatarUrl` from `avatar_url` or from gravatar like so", and others, that feel like they should only live in one place.

I appreciate that some of the wrinkles of `Avatar` are left out -- but that seems like it only lessens the duplication, and doesn't really provide a reason to do it.

Also I think this bit in the commit message is mistaken:
> does not render TextAvatar as it is only for group conversations

In fact, `GroupPmConversationItem` doesn't use this path -- it refers to `TextAvatar` directly.

Perhaps that `TextAvatar` path in `Avatar` never gets used at all? It might be something that applies only if we pass bogus data like an empty `email` etc., which we might no longer do and should probably stop doing if there's anywhere that we do. This is a bit of a tangled area of our code, so it'd be great to simplify it down.

src/users/userSelectors.js
```diff
@@ -54,6 +61,9 @@ export const getUsersSansMe: Selector<User[]> = createSelector(
   (users, ownEmail) => users.filter(user => user.email !== ownEmail),
 );
 
+export const getOwnUser: Selector<User> = (state: GlobalState) =>
```
gnprice: Perhaps `getSelfUser`? You'd raised yesterday the idea of making the naming of own/self/me/etc. consistent, and I think "self" seems like the best single term for that.

src/account-info/ProfileCard.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  handleUpdateAwayStatus = () => {};
+
+  render() {
+    const { ownUser, presence } = this.props;
+
+    return (
+      <View>
+        <AccountDetails user={ownUser} presence={presence[ownUser.email]} />
```
gnprice: You'll want to simplify this when rebasing atop your more recent refactor that nicely did that for existing components.

borisyankov:
Closing as the original review was done on 4 commits that are very outdated compared to what is currently in. Opening #3344

Issue #3264, opened 2019-01-10
Scroll position got disturbed when keyboard show/hides


jainkuniya:
Ideally: When compose box is focussed, keyboard pops and message list should scroll so that content shift upwards.
And then when keyboard hides, again message list should scroll to make sure that content shift downwards. 
This behaviour is normal and almost all chat apps follow this.

When this feature was implemented, this was working as expected. But recently, it seems to be broken.
Before it was working as expected on iOS. And on Android there was a small issue when list scroll is near to bottom. But now it is completely broken.

I have been working on it, and I guess I have figured out the problem. Will submit PR soon :)

@zulipbot claim

borisyankov:
Another related issue is that the position is affected with `UnreadNotice` shown/hidden.

PR #3268, opened 2019-01-11
 webView,css:Fix scroll got disturbed when viewport height changes.

jainkuniya:
Before `img` height was in `vh`. So when viewport height changes,
`img` height also gets changed. Thus scroll position gets affected.

Now provide `img` height in `px`, so that it doesn't change on
viewport height change.

Note: on unread notice/keyboard show/hide, webView viewport height
changes.

References: https://developer.mozilla.org/en-US/docs/Web/CSS/length

Fix: #3264

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -294,7 +294,7 @@ hr {
 .message_inline_ref img,
 .twitter-image img {
   width: 100%;
-  height: 30vh;
+  height: 160px;
   object-fit: contain;
 }
 blockquote {


src/webview/css/css.js
```diff
@@ -257,8 +257,8 @@ hr {
 .message_inline_ref img,
 .twitter-image img {
   width: 100%;
-  height: 30vh;
-  object-fit: contain;
+  height: 200px;
+  object-fit: cover;
```
borisyankov: I don't think we can just switch to `cover` here. In many cases (wide landscape images) it cuts out the side parts of the image, which in some cases might be OK, but in others result in a weird zoom. Maybe we need more logic around this. If it would cut more than 20-50% of the image do not do this...

borisyankov:
I think this is addressing only a small part of the issue and in a way that is not clearly an improvement - it fixes the jump but regresses the look a bit.

Also, this does not address the `UnreadNotice` jump.

If you add a listener inside the WebView for `resize` events and make sure the bottom position remains the same through the event, that is probably the solution we are looking for.

gnprice:
```
    Before `img` height was in `vh`. So when viewport height changes,
    `img` height also gets changed. Thus scroll position gets affected.
```

* Good catch, thanks! This seems like something we definitely need to fix, regardless of what other fixes are necessary.

* How did you choose the number 200px?

* Also, I'm tempted to say the height should still depend on the size of the screen -- we just need to turn that into a `px` value from outside the WebView. (Using `Dimensions`, I think, like we do in `Lightbox.js`.)

---

```
    webView, img, css: Crop image and fit in the component.
    
    Almost all chat apps like twitter, whatsapp renders the cropped image
    in the message list. It looks good, having all images same height and
    width.
```

* As Boris says, this is trickier. And it's unrelated to the scrolling issue, right?

  I do think we can likely improve on our current choice, but more design work will be needed.

  Also, I feel like we discussed this somewhere recently (possibly following this PR), with some details that took effort to produce. Would you please track down that conversation and link to it from here? That way we can avoid repeating ourselves.

---

```
    webview, js: Scroll `smoothly` on window resize.
    
    Message list scrolls to adjust the position when keyboard pops in/out.
```

* Hmm interesting. I'm not sure I like this effect -- it feels pretty heavy. It also doesn't look like other messaging apps do this, on a quick and partial survey.

---

```
    android,webView: Fix scroll position on resize if at near bottom.
```

* Hmm, this looks rather subtle. More effort will be needed to explain it, I think. :slightly_smiling_face: 

* It looks like `nearBottom` can more simply be written as `belowContentHeight < 300`. Is that right, or what am I missing?

  On the other hand, in `handleScrollEvent` one of the two is set in a `setTimeout` while the other is set immediately. What's the motivation behind that? It seems pretty complicated to reason about -- I challenge you to write a clear explanation of what `nearBottom` and `belowContentHeight` *mean*, that's actually accurate for this code. :wink: 

```
+  if (
+    difference > 0
...
+  } else if (belowContentHeight > 0) {
+    const realdiff = -1 * Math.min(-1 * difference, belowContentHeight);
```

* There seem to be a lot of possible combinations here of different conditions for what's greater than what. (Having recently filed my US tax return, it reminds me unhappily of forms like [this one](https://www.irs.gov/pub/irs-prior/i1040gi--2018.pdf#page=40).) I think some of them may be impossible, which simplifies things -- but I don't want to have to do the careful math to find out. How about unpacking this into a more explicit set of conditions?

* Also please no commented-out code. That's what version control is for. :stuck_out_tongue:

borisyankov:
> Also, I'm tempted to say the height should still depend on the size of the screen -- we just need to turn that into a px value from outside the WebView. (Using Dimensions, I think, like we do in Lightbox.js.)

Hmm, is `Dimensions` for Screen size or viewport size. We definitely want the first. If it is the latter we would be replicating the `vh` behavior.

gnprice:
> > Also, I'm tempted to say the height should still depend on the size of the screen -- we just need to turn that into a px value from outside the WebView. (Using Dimensions, I think, like we do in Lightbox.js.)
> 
> Hmm, is `Dimensions` for Screen size or viewport size. We definitely want the first. If it is the latter we would be replicating the `vh` behavior.

Take a look at the Lightbox example I mentioned, or at the docs:
https://facebook.github.io/react-native/docs/dimensions.html

It doesn't even know anything about a WebView -- so it won't care whether e.g. `UnreadNotice` is crowding the message list.

Now, there is an important question of whether it's affected by the keyboard. We'd certainly want to pin that down before using it.

But an absolute size in `px` would also be perfectly reasonable. If we can't easily get a screen-size measurement that doesn't depend on the keyboard, then that's probably what we'll want instead.

gnprice:
To shorten my review, after rereading:

* First commit just needs explanation -- why 200px?
  * Having thought about this again (and after the discussion with @borisyankov just above): Let's go ahead with a fixed number of px, either 200px or some other value. That's simpler than trying to adapt to screen size, and I'm not sure that the latter would be better even if we found a way to do it reliably.
* Let's drop the second and third commits.
* Last commit needs to be clearer in the actual code... and one thing I think might help you do that is to attempt to write a good explanation of what those two new globals mean. That'll be hard. Then make changes so that they mean something that's easier to clearly explain.

gnprice:
Hmm -- in my emulated Nexus 5, this actually makes images quite a bit *taller*:

image

Compare from master (well, the base of this branch):

image

That's even though the commit message explains it was supposed to make them *smaller*, by a ratio of 2/3.

    Make it as 200px, because usually whole screen is around ~1000px and
    200px is 20%.

I think you have overestimated how big a phone screen is. :wink: For one thing, remember these are "logical" pixels.

Please take a look at this background doc:
https://github.com/zulip/zulip-mobile/blob/master/docs/background/layout.md

gnprice:
Also remember the old `30vh` is 30% of the *viewport*'s height -- that's the original point of this branch, after all! And the viewport is shorter by at a minimum the status bar, the app bar, the compose box, and (on Android) the system buttons.

gnprice:
Merged -- thanks again @jainkuniya !

A couple of things that I had to adjust in the commit message first:
* The message said 180px, though the code said 160px.
* The message said

  > Tested on Android Pixel 3 emulator, One plus six physical device,
  > and iOS X simulator, 180px is resulting in same size of image as old 30vh.

  but surely it wasn't actually the same size, right? For one thing it's unlikely it was quite the same size across those three. (Also I kept reading "iOS X simulator" as "iOS simulator", which is meaningless here as different models will surely have different results. Now I realize you possibly meant to write "iPhone X simulator"... but surely that would be much taller, no?)

  So instead I wrote what the actual size came to in the emulator I was using, and said how that roughly generalized.

PR #3275, opened 2019-01-14
Fix Action Sheet popping under Lightbox Footer on Android

### LINKED ISSUE #3274: [Android] Action sheet in Lightbox pops under the footer
As reported here:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.20click.20image.20more.2C.22cancel.22.20is.20hidden

borisyankov:
Fixes #3274

We are setting `zIndex` on the styles applied to header & footer
were making them go over the Action Sheet.

This property will not be needed if we move `PhotoView` before
these two components. This does not change the positioning as they
are overlaying it, with position `absolute`.

### DIFF
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -27,7 +27,6 @@ const styles = StyleSheet.create({
     backgroundColor: 'black',
     opacity: 0.8,
     position: 'absolute',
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @borisyankov ! Merged.

PR #3280, opened 2019-01-15
Fix broken layout in ChatScreen on iOS

### LINKED ISSUE #3273: Keyboard overlaps the compose box on iOS
This is a release-blocker issue on `master` and `107`

Discussed here
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Keyboard.20placement.20on.20iOS

borisyankov:
Fixes #3273

A previous commit fixing a weird issue on Android inadvertantly
broke the iOS layout:
https://github.com/jainkuniya/zulip-mobile/commit/f142452b4fa66ea00fc49048c4b31b89fb4dbe6c

The ZulipStatusBar compoennt  is a difference between Android and iOS
as it has actual height on iOS but not on Android, we can never paint
in the status bar area of Android, we can change the color only.

This commit keeps the fix from the previous commit but simplifies it:
* removes the extra View, no need for the extra wrapper
* simplified the style, no need for the extra `flex`
* moved `ZulipStatusBar` to bottom because of `flexDirection: 'column-reverse'

### DIFF
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -22,7 +22,6 @@ type Props = {|
 const componentStyles = StyleSheet.create({
   /** A workaround for #3089, by letting us put Chat (with MessageList) first. */
   reverse: {
-    flex: 1,
     flexDirection: 'column-reverse',
   },
 });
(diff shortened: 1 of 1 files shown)


gnprice:
Merged. Thanks @borisyankov !

PR #3282, opened 2019-01-16
Fix narrow from config

### LINKED ISSUE #3284: "No messages" on following a notification
This appears to happen 100% of the time on 22.0.107 if:
* the app isn't already running, and
* it's a conversation we haven't previously seen.

This is the version that's currently the beta on iOS, and went to prod yesterday on Android.

(I think if it's a conversation we *have* seen, we'll show a stale list of messages, i.e. omitting the new one that actually caused the notification. Also bad! Not sure of that, though.)

Report here, and extensive debugging discussion below:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications/near/682058

borisyankov:
Reverts the reversal of two previous commits.

Fixes #3284 .

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -12,6 +12,7 @@ type Config = {|
   startup: {
     narrow: ?Narrow,
     anchor: number,
+    shouldNarrowToNotification: boolean,
     notification: ?Notification,
   },
   messagesPerRequest: number,
@@ -31,6 +32,7 @@ const config: Config = {
   startup: {
     narrow: undefined,
     anchor: FIRST_UNREAD_ANCHOR,
+    shouldNarrowToNotification: false,
     notification: undefined,
   },
   messagesPerRequest: 50,
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -19,7 +19,6 @@ import {
   getLastMessageId,
   getCaughtUpForActiveNarrow,
   getFetchingForActiveNarrow,
-  getTopMostNarrow,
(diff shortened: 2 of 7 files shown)


gnprice:
Just to briefly summarize from the chat thread (linked from #3284):
* I've just merged #3283 as an immediate fix for the critical issue here.
* But that adds back some code that, though it works, we're very much agreed is not the right design. So I am still quite interested in a better fix, more along these lines.

borisyankov:
Not only it is not the right design, but it is also not the right fix.
The notification opening still show a "No messages" text but loads the messages after that which is still an improvement.

borisyankov:
I have added a commit that prevents an accidental repeat opening of the notification and this PR is ready for a review.

borisyankov:
Added reversal commits of the earlier reverts.

gnprice:
Thanks @borisyankov for this PR and the updates!

In our audio discussion today, I realized that `config.startup.notification` can be replaced completely by `doNarrow`, now that thunk actions are buffered (since 42adfb348).

I suggested then that you pick up making that change, as a revision of this PR. But then later today as I resumed doing some of the notifications work I've been doing, I realized that those changes will overlap with some changes I'd been planning yesterday to make in my next steps. So I propose a small change of plan: I'll go and make that change myself. (In fact I have a draft from just now, in my `notif` branch.)

Closing this PR in favor of that plan. Thanks again for helping explore this! Please do mention if there's something else you notice as I'm making these changes.

PR #3288, opened 2019-01-17
Retry on failed fetch messages

### LINKED ISSUE #3281: Message does not load if Internet is back.
If you open a message during weak internet connection, grey bars are displayed instead of the real meassage.
This state lasts forever even if Internet connection is back or change to a WiFi.

It also happens to me that the message was removed from the new message list after turning back. I wanted to go back and forth to trigger reload. I cannot confirm this now though. It is Zulip 21.2.106 on Android.

borisyankov:
Fixes #3281


gnprice:
Thanks @borisyankov ! Merged.

gnprice:
FTR, I just reverted this because it broke our tests. Details in #3281 .

Issue #3289, opened 2019-01-17
[Autocomplete] People autocomplete doesn't pops if `@` is typed in new line (after enter).


jainkuniya:
![image](https://user-images.githubusercontent.com/18511177/51335188-6e115100-1aa7-11e9-954d-ccb17e2d8ccd.png)

https://chat.zulip.org/#narrow/stream/48-mobile/topic/Problems.20mention.20pop.20up.20on.20iOS
@zulipbot claim

PR #3290, opened 2019-01-17
people autocomplete: Fix it doesn't pop if `@` entered in new line.

jainkuniya:
We only consider `@` when the preceding char is either ` `, `\n`, `:`
or `#`. Because every email as `@` in it and we don't want to trigger
suggestions in such cases.

Fixes: #3289

image

### DIFF
--- a/src/autocomplete/__tests__/getAutocompleteFilter-test.js
+++ b/src/autocomplete/__tests__/getAutocompleteFilter-test.js
@@ -27,6 +27,18 @@ describe('getAutocompleteFilter', () => {
     });
   });
 
+  test('get respective lastWordPrefix even when keyword is entered in new line', () => {
+    selection = { start: 1, end: 1 };
+    expect(getAutocompleteFilter('\n@', selection)).toEqual({ filter: '', lastWordPrefix: '' });
+    expect(getAutocompleteFilter('\n#', selection)).toEqual({ filter: '', lastWordPrefix: '' });
+    expect(getAutocompleteFilter('\n:', selection)).toEqual({ filter: '', lastWordPrefix: '' });
+
+    selection = { start: 2, end: 2 };
+    expect(getAutocompleteFilter('\n@', selection)).toEqual({ filter: '', lastWordPrefix: '@' });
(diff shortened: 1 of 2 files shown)


src/autocomplete/__tests__/getAutocompleteFilter-test.js
```diff
@@ -11,6 +11,12 @@ describe('getAutocompleteFilter', () => {
     selection = { start: 1, end: 1 };
     expect(getAutocompleteFilter('@', selection)).toEqual({ filter: '', lastWordPrefix: '@' });
 
+    selection = { start: 1, end: 1 };
```
borisyankov: These would be better in a new test case.

armaanahluwalia:
Try entering more than 5 lines of text and then trying an @ mention. The popup is cutting off content at the top. You may want to look at that since you're fixing this anyway.

jainkuniya:
image

@armaanahluwalia On Android or iOS? I can't reproduce. Can you please share the screenshot?

borisyankov:
@armaanahluwalia The current implementation definitely has issues with height which we can improve but is surprisingly tricky so we delayed it for later.

armaanahluwalia:
Its a pretty bad bug for a pretty common use case aka typing longer messages on zulip with an @ mention.

See [Bug Video 2](https://streamable.com/vo6ym)

borisyankov:
Thanks for the video @armaanahluwalia
I agree it is worse than I thought and worth fixing.

src/autocomplete/__tests__/getAutocompleteFilter-test.js
jainkuniya: Done 👍

gnprice:
Thanks @jainkuniya for this fix! Merged.

This function in general is pretty convoluted and hard to follow, and this change unfortunately makes it a little bit more so. I think there are more issues lurking that are hard to see because of the complexity of it. So I'd love to see a followup to make it simpler and clearer. Filed #3345 with more details on that.

gnprice:
@armaanahluwalia Thanks for that helpful video!

Please go ahead and file that as an issue. Here on a loosely related PR thread, it's liable to get forgotten. :wink:

Issue #3295, opened 2019-01-18
Placeholder Text padding shows irregularity


Jigar3:
![whatsapp image 2019-01-18 at 10 51 59 am](https://user-images.githubusercontent.com/32068121/51367146-3f2ec580-1b0f-11e9-8aa1-a8a78bb7d0b9.jpeg)
The placeholder text padding is not perfect when I opened it for the first time.

![whatsapp image 2019-01-18 at 10 52 00 am](https://user-images.githubusercontent.com/32068121/51367143-3d650200-1b0f-11e9-8a18-ed197937ef8b.jpeg)
After clicking on the inputbox, the padding becomes correct.

I want to take up this issue. Can anyone explain what might cause this behaviour to happen?
@zulipbot claim

borisyankov:
@timabbott has indicated that this is an important issue to fix quickly.

I suggest we give @Jigar3 a chance to find the issue if he has time in the next few days, if not @jainkuniya you should find and fix the issue 👍 

@Jigar3 indeed `git bisect` is your best bet. Also, a great opportunity to learn how it works (I am not assuming you already do, as most people don't).

You can see all the tags with `git tag`. This was likely introduced in one of the most recent releases. Check 106, or 105 and once you've found a commit that this issue does not exist use bisect to (surprisingly) quickly find what commit did introduce this bug.

Jigar3:
@borisyankov I will definitely try to resolve this issue thanks for the chance

gnprice:
Thanks @Jigar3 for reporting this issue! We'd been noticing it for a little while, but hadn't gotten around to taking care of it. And thanks @jainkuniya for some good systematic debugging and for the fix :smile: .

FTR I'll quote here @jainkuniya 's observations from debugging, as recorded in commit 6351c8e07 which fixed it:

```
* Only reproducible when compose box is shown after message loading
  (GIF like loader) after narrowing on Android.  Can't reproduce on
  narrows whose messages are cached and immediately visible after
  narrowing.

* If on the topic input (a) the `this.getCanSelectTopic()` condition
  is removed and (b) the `multiline` prop is set, it starts showing
  the issue, just like the message input.

So from this it can be inferred that:

* It is a `multiline` input issue, which occurs when its parent first
  got rendered with `display: none`.  Normally it is not reproducible
  on topic input because at the time of first render (when style is
  `display: none`), topic input is not rendered (because
  `this.getCanSelectTopic()` is false).

[...]

PR #3300, opened 2019-01-21
android, composeBox: Fix placeholder text padding irregularity.

jainkuniya:
Few observations:

* Only reproducible when compose box is shown after message loading
(GIF like loader) after narrowing on Android. Can't reproduce on
narrows whose messages are cached and immediately visible after
narrowing.

* If `this.getCanSelectTopic()` condition is removed from topic input
and `multiline` is added to `Input`, it start behaving like `Message`
input.

So from this it can be inferred that:

* It is a `multiline` input issue, which occurs when its parent first
got rendered with `display: none`. Normally it is not reproducible on
topic input because at the time of first render (when style is
`display: none`), topic input is not rendered (because
`this.getCanSelectTopic()` is false).

So to fix this issue, this commit replace `display: none` with `height:0,
width:0`, without breaking any functionality.

Fixes: #3295

I tried to use `git bisect`, but on some commits I wasn't able to build app. Was taking lots of time to check on multiple steps. So I thought instead of finding what caused it let's go for finding a solution to fix it. 

This is mostly a hack and I don't have complete reasoning. This is just based on observation and facts.

@Jigar3 in case you find better solution, we would love to take it.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -293,7 +293,7 @@ class ComposeBox extends PureComponent<Props, State> {
     const placeholder = getComposeInputPlaceholder(narrow, auth.email, users);
     const style = {
       marginBottom: safeAreaInsets.bottom,
-      ...(canSend ? {} : { display: 'none' }),
+      ...(canSend ? {} : { height: 0 }),
     };
 
     return (
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -351,7 +351,7 @@ class ComposeBox extends PureComponent<Props, State> {
             <FloatingActionButton
               style={styles.composeSendButton}
               Icon={editMessage === null ? IconSend : IconDone}
-              size={32}
+              size={canSend ? 32 : 0}
               disabled={message.trim().length === 0}
               onPress={editMessage === null ? this.handleSend : this.handleEdit}
             />


borisyankov:
Do you need to set both `height` and `width` to `0`? Isn't `height` enough?

borisyankov:
Not sure if that works 100% on Ansdroid but on iOS it fails to hide the `SendButton` component.

jainkuniya:
> Isn't `height` enough?

Yup! so should we keep only `height`?

> Not sure if that works 100% on Ansdroid but on iOS it fails to hide the `SendButton` component.

Got it. For me it is working 100% on Android.

jainkuniya:
Actually it is also not able to hide `SendButton` on Android too.

gnprice:
Fascinating -- thanks for this debugging!

I don't feel great about this solution, because it makes the logic more complicated -- in particular how it opens up the need to go in and patch up things like the `size` of the send button. There's already kind of a lot of interaction between different pieces that makes it hard to reason about the behavior of this whole component, and this change makes that a bit more so.

How about if instead of `display: 'none'` we simply don't render all these sub-components at all? I.e. like
```
if (!canSend) {
  return null;
}
```
That's the standard React solution for this kind of situation, where we conditionally want to just not show something. 

In general the one reason it can make sense to instead do something like `display: 'none'` (or this `height: 0` hack) is in order to keep the child components' state if they get hidden and re-shown. But I think that doesn't apply here:
* The inputs are uncontrolled, and we already track their values in the state of `ComposeBox`.
* For the topic input, we already do exactly the standard thing! If you blur the message and topic inputs, we stop rendering the topic input at all. And indeed this works fine when I try it: edit the topic, tap the message list so the topic input gets dropped, then tap the message input so it gets shown again, and the edited value is right there.

borisyankov:
The implementation is as is because of:
https://github.com/zulip/zulip-mobile/commit/824c908ccdb1cfcc6de4f9711dbb63eab39adeec#diff-e83898e461a984b9837e188409ba3707

jainkuniya:
So let's go with this:

* Only render `ComposeBox` component if `canSend`.
* Do not use `return null`, because `componentDidMount` will not be called again and text will get lost on toggling `canSend` (824c908#diff-e83898e461a984b9837e188409ba3707)
* Do something like `{canSend && <ComposeBox narrow={narrow} />}` in `Chat.js`

jainkuniya:
(submitted #3310, with this^ approach)

gnprice:
> The implementation is as is because of:
> [824c908#diff-e83898e461a984b9837e188409ba3707](https://github.com/zulip/zulip-mobile/commit/824c908ccdb1cfcc6de4f9711dbb63eab39adeec#diff-e83898e461a984b9837e188409ba3707)

Ah, thanks for pulling that up! That's helpful context.

I spent some time later yesterday looking at the ComposeBox logic, and drafting up some refactors. One of the things I noticed was that `componentDidMount`; I think what it's doing can more straightforwardly be handled by passing a `defaultValue` prop to Input/TextInput. With this additional context, then, a bonus of doing that would be that we can go back to the standard "just don't render" solution for the `!canSend` condition.

gnprice:
Oh, though #3310 also makes a lot of sense :slightly_smiling_face: . That bonus will become moot, then.

gnprice:
Closing in favor of #3310. Thanks again @jainkuniya for the debugging, and for both PRs!

PR #3302, opened 2019-01-21
webview: Fix small scroll issues with message list

### LINKED ISSUE #3301: Message list scroll position is often slightly messed up
When showing/hiding `UnreadNotice` the message list scroll position is moved by the height of that panel.

As a result, when you narrow to a conversation with one or a few unread messages and at least a screenful of total messages (which is a very common scenario when staying caught up in a busy org), what you typically see is the screen jumps a couple of times, and then settles scrolled slightly too far up, with the last message only partly visible. This is annoying and looks pretty broken.

[summary expanded by @gnprice]

borisyankov:
Fixes #3301

When showing/hiding UnreadNotice the message list scroll position
is moved by the height of that panel.

This (and other similar problems) can be fixed by a sophisticated
algorithm that 'anchors' the scroll position of the webview to the
bottom instead of the default to top. That turned out to be a much
more tricky solution, but a much simpler one that is a subset of
it would do the trick for now.

We add an event handler on the webview's 'resize' event. This will
be called in these situations:
 1. the keyboard is shown/hidden
 2. the phone orientation is changed
 3. the compose box shows/hides the topic input
 4. the UnreadNotice is shown/hidden

This code is an effective fix for (4)
I was a bit concerned that (3) might be negatively affected by
the fix but after testing it few times I think the behavior is an
improvement.

Not only that but we should look for even more cases where we can
scroll to the very bottom of the list (and maybe expand the 100px
distance that triggers it)

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -522,4 +522,7 @@ documentBody.addEventListener('touchmove', function (e) {
 documentBody.addEventListener('drag', function (e) {
   lastTouchEventTimestamp = 0;
 });
+window.addEventListener('resize', function (event) {
+  scrollToBottomIfNearEnd();
+});
 `;
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -684,3 +684,7 @@ documentBody.addEventListener('touchmove', (e: TouchEvent) => {
 documentBody.addEventListener('drag', (e: DragEvent) => {
   lastTouchEventTimestamp = 0;
 });
+
+window.addEventListener('resize', event => {
+  scrollToBottomIfNearEnd();
+});


gnprice:
Thanks @borisyankov !

Playing around a bit before and after this change:
* It definitely improves things in the (quite common) situation where the user is caught up, and scrolled to the bottom, when a new message arrives
* It doesn't affect the behavior when the user has scrolled up, and then a new message arrives (causing UnreadNotice to appear) -- the scroll position jumps, and stays jumped
* I am not convinced it fixes #3301, because the latter seems like part of #3301 
* The behavior in the caught-up case is still jumpy and deserves improvement

So I think this is a useful hack for now but "fixes #3301" doesn't seem accurate. And I would love to see some more thorough fixes in this area :slightly_smiling_face:

gnprice:
Oh also, though:
* Putting this code here at the end of the file is messy -- the surrounding code is completely unrelated. It should be next to the other related code.
* As I go looking for that related code... here it is!
```
/* Cached to avoid re-looking it up all the time. */
let viewportHeight = documentBody.clientHeight;

window.addEventListener('resize', event => {
  const difference = viewportHeight - documentBody.clientHeight;
  if (documentBody.scrollHeight !== documentBody.scrollTop + documentBody.clientHeight) {
    window.scrollBy({ left: 0, top: difference });
  }
  viewportHeight = documentBody.clientHeight;
});
```

Note the `scrollBy`.

This is obviously closely related logic. Can you explain your understanding of how these two interact?

It's OK if on investigation we're puzzled and we put in an interim fix that empirically seems to be an improvement -- but I do want us to have actually looked at the question.

This also underlines why it is really quite an important thing to put the new bit of code next to (or in some other way cross-reference it with) the existing related code :wink:

gnprice:
This thread seems to have gone stale. The issue is still one I'd love to see a fix for; closing to focus discussion back on the issue thread, #3301, and to open things up for a future PR.

Issue #117, opened 2016-11-24
Handle sharing from other apps (Android)


borisyankov:
Register app so other apps can send it content (text, urls, photos).

vishwesh3:
@zulipbot claim

zulipbot:
Hello @vishwesh3, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for a week. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zulipbot:
Hello @vishwesh3, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for a week. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zulipbot:
Hello @vishwesh3, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

vishwesh3:
@zulipbot claim 
will resume work on https://github.com/zulip/zulip-mobile/pull/674

gnprice:
Some later discussion in #1102 and #1351.

ftribrunn:
Is anyone working on this?

jainkuniya:
Will submit a PR soon for this.

gnprice:
Just edited to make this issue specific to Android, and filed a separate issue #3277 for the iOS side. I expect a lot of the work will be separate between the two platforms; and we hear this request a lot more often for Android than iOS, I think related to the platform conventions making this a more common thing on Android.

Here's upstream docs on the platform feature:
https://developer.android.com/training/sharing/receive

andreasoc:
[...]

PR #3304, opened 2019-01-22
Handle sharing from other apps (Android)

jainkuniya:
Issue #117

I guess for v1 only `text` will work. This is ready for initial review.


gnprice:
Thanks @jainkuniya !

I'm going to close this because it isn't quite what we want, and I think it will be more work to turn it into a version we want than to build it fresh.

Some quick high-level observations, for concreteness:
* Before implementing something here, we should discuss a bit what we want the UX to be like -- either on the issue thread or in chat.  Then we should also discuss what that will look like in Android platform terms and in the design of the code.
* In particular, IIUC this will open the app, taking you to the home screen as usual, and then you navigate somewhere and the text shows up as effectively a draft? That sounds confusing, and I think we can do better.
* As always, each commit should leave things in a state where everything works correctly. There are a couple of spots here that leave things in a broken in-between state; e.g. after the first commit, we advertise that we can handle sharing (of `text/plain`) but if the user tries to use that, we just drop it on the floor.

Issue #3176, opened 2018-11-29
home screen: Jitter on Scroll.


jainkuniya:
@armaanahluwalia reported:
> I'm seeing some very distracting and weird behaviour is observed while scrolling the unreads screen on 20.0.103. The title bars position seems to be moving around.  
[Link to video](https://streamable.com/4zfjg)

This is a upstream issue.

Read more at https://chat.zulip.org/#narrow/stream/48-mobile/topic/Jitter.20on.20Scroll

borisyankov:
The upstream issue is here:
https://github.com/facebook/react-native/pull/22025

This affects our unread screen `UnreadCards`, all users screen and user picker screen (that use the `UserList` component)

gnprice:
Fixed upstream in facebook/react-native@e4fd9babe, which landed in v0.59. So this should be resolved when we complete #3399 .

gnprice:
And in fact with #3399 complete, this is now fixed! I just went and scrolled a ways through the list of users on czo (from "people" icon -> "Search" button), and the heading stayed firmly in place.

PR #3305, opened 2019-01-22
No sticky headers in Unread tab

### LINKED ISSUE #3286: Scrolling through topics on iOS is not working
Trying to scroll, topics jump around and won't stop scrolling. https://photos.app.goo.gl/hK1bj9RmvLTBF8i2A demonstrates the behavior on @vietdnguyen's phone (and @rhausam is experiencing the same).

borisyankov:
Fixes: #3286 and #3176

Related fix upstream but not ready to merge:
https://github.com/facebook/react-native/pull/22025

A planed minor redesign (discussed with Rishi) to make the Unread
tabs less flashy and 'circus-colors-like' requires us to remove
the stickiness from the headers.

Even though though the design is still WIP, it makes sense to use
this change as a fix to an issue that was already reported multiple
times.

This makes the UnreadCards SectionList have normal, non-sticky
headers. We have to set `stickySectionHeadersEnabled` explicitly
to `false` as the defaults differ for iOS (true) and Android (false)

### DIFF
--- a/src/unread/UnreadCards.js
+++ b/src/unread/UnreadCards.js
@@ -57,7 +57,7 @@ class UnreadCards extends PureComponent<Props> {
 
     return (
       <SectionList
-        stickySectionHeadersEnabled
+        stickySectionHeadersEnabled={false}
         initialNumToRender={20}
         sections={unreadCards}
         keyExtractor={item => item.key}


borisyankov:
The upstream fix is merged into React Native, so this PR is no longer needed.

PR #3306, opened 2019-01-22
Render `/me ...` messages

### LINKED ISSUE #1773: /me command doesn't work in mobile app
**App Version:**
Android 6.1.48

borisyankov:
Fixes #1773

* Add `processMeMessage` function
* Run `processMeMessage` on each message's content before rendering.
* Add simple 'italics' style to '/me ...' messages.

The styling is the exact same one Slack uses (just italics) and
saves on having to support another completely different message
layout on all the various device sizes.

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -123,6 +123,9 @@ hr {
   width: 100%;
   border-radius: 3px;
 }
+.message-me {
+  font-style: italic;
+}
 .content {
   width: 100%;
   max-width: 100%;
--- a/src/webview/html/__tests__/processMeMessage-test.js
+++ b/src/webview/html/__tests__/processMeMessage-test.js
@@ -1,0 +1,33 @@
+import processMeMessage from '../processMeMessage';
+
+describe('processMeMessage', () => {
+  test('empty string is not changed', () => {
+    const message = '';
+    const result = processMeMessage(message);
+    expect(result).toEqual(message);
+  });
+
+  test('a simple string is returned as is', () => {
+    const message = 'Some message';
+    const result = processMeMessage(message);
+    expect(result).toEqual(message);
+  });
+
+  test('if the string contains a "/me" in <p> tag replace it', () => {
+    const message = '<p>/me is happy</p>';
+    const result = processMeMessage(message);
+    expect(result).toEqual('<span class="message-me">is happy</span>');
+  });
+
+  test('no "<p>" tag, no change', () => {
+    const message = '/me is happy';
+    const result = processMeMessage(message);
+    expect(result).toEqual(message);
+  });
+
+  test('the "/me" has to come in the first paragraph or else no change', () => {
+    const message = '<p>Hey its</p><p>/me Mario</p>';
+    const result = processMeMessage(message);
+    expect(result).toEqual(message);
+  });
+});
--- a/src/webview/html/messageAsHtml.js
+++ b/src/webview/html/messageAsHtml.js
@@ -15,6 +15,7 @@ import { shortTime } from '../../utils/date';
 import aggregateReactions from '../../reactions/aggregateReactions';
 import { codeToEmojiMap } from '../../emoji/data';
 import processAlertWords from './processAlertWords';
+import processMeMessage from './processMeMessage';
 
 const messageTagsAsHtml = (isStarred: boolean, timeEdited: number | void): string => {
   const pieces = [];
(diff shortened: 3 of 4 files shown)


borisyankov:
@gnprice ping for a review.

borisyankov:
> I think the word processing for /me would be better in processAlertWords itself 

That wouldn't be very true to the function name though, wouldn't it? This has nothing to do with alert words :)

borisyankov:
> The style of display would be better off exactly the same as the one displayed in the webapp IMO -- maybe some trickery with display: inline or :after, and CSS variables would achieve that?

No need for tricky CSS, it can be achieved by relatively straight-forward CSS... and then we need to support another way of rendering messages that is barely used and barely needed (the feature request has been open for years, the PR for a half, and no one is really asking about this, this is a 'barely worth it PR', IMHO).

Also, to reiterate, Slack do render it exactly like this PR. I would argue the web app needs to revert to this rendering too (for simplicity and maintainability).

ishammahajan:
> Slack do render it exactly like this PR

Oh I wasn't aware of that, I misread the commit. Sorry about that.

> This has nothing to do with alert words

`/me` isn't an alert word? I was under the impression that it was, my mistake. Even then though, processing another potential word would be cumbersome. Perhaps adding both of the processing to one function (e.g. `processMessageContent(content)`) in `messageAsHtml` would be better for the sake of modularity.

borisyankov:
> /me isn't an alert word? I was under the impression that it was, my mistake.

Honest error. I barely know what they are :)

https://zulipchat.com/help/add-an-alert-word

gnprice:
Thanks @borisyankov for this PR and @ishammahajan for the review!

On the logic: in the webapp code (thanks for [the link](https://github.com/zulip/zulip/blob/8aee6a1dd/static/js/message_list_view.js)!), it has this:
```
    _maybe_format_me_message: function (message_container) {
        if (message_container.msg.is_me_message) {
            // Slice the '<p>/me ' off the front, and '</p>' off the first line
            // 'p' tag is sliced off to get sender in the same line as the
            // first line of the message
            var msg_content = message_container.msg.content;
            var p_index = msg_content.indexOf('</p>');
            message_container.status_message = msg_content.slice('<p>/me '.length, p_index) +
                                                msg_content.slice(p_index + '</p>'.length);
```

Note in particular the `is_me_message` check.

So, I think we should consult that; it looks like this version doesn't. I think that will even mitigate to some extent the brittleness that you point out in the error message :slightly_smiling_face: 

Looks like we have the plumbing to sling it around; it's a flag.

ishammahajan:
> zulipchat.com/help/add-an-alert-word

Helpful link! :)
Looks like I'll be getting a lot more notifications now 😛.

borisyankov:
> So, I think we should consult that; it looks like this version doesn't. I think that will even mitigate to some extent the brittleness that you point out in the error message 🙂
>
> Looks like we have the plumbing to sling it around; it's a flag.

I am not sure what you mean.

gnprice:
> > So, I think we should consult that; it looks like this version doesn't. I think that will even mitigate to some extent the brittleness that you point out in the error message slightly_smiling_face
> > Looks like we have the plumbing to sling it around; it's a flag.
> 
> I am not sure what you mean.

In the webapp, the corresponding logic only fires if the `is_me_message` flag is set on the message.

Digging a bit more to see what that works out to... it's set from the function `Message.is_status_message` in the server... and in particular that rejects any message with a newline in the raw Bugdown content. So e.g. a message
```
/me writes a second
line
```
will *not* be treated as a me-message.

(There actually seems to be a bug where the webapp's local echo disagrees with the server -- the former does treat it as a me-message.)

I don't love the implementation of this feature, and don't have a strong view on the edge cases of what *should* count as a me-message; but of course whatever we do, the behavior should agree across platforms.

So to do that here, we just need to consult the `is_me_message` flag on the message. Which I believe we already have in our `flags` data structure, alongside the other flags.

gnprice:
Quick update -- I reported this:

> (There actually seems to be a bug where the webapp's local echo disagrees with the server -- the former does treat it as a me-message.)

as zulip/zulip#12450, and it was fixed in zulip/zulip#12619 and in particular zulip/zulip@c03efd297. The backend logic was updated to match the webapp frontend.

Looking briefly back at the code in this PR, it's based on the webapp frontend, not the backend, so hopefully it's already aligned with the behavior that the backend and webapp frontend now share.

gnprice:
Closing because this implementation direction doesn't work. The issue remains open, and can be fixed with the approach described at https://github.com/zulip/zulip-mobile/pull/3306#issuecomment-497481741 above.

PR #3309, opened 2019-01-24
clean up, chat: Remove unused `scrollOffset` variable.

jainkuniya:
It got introduced in 9dc95ccba4cfb257a44e5e062ea7542035e34f9e, when we
were using native message list. Now we are no longer using this
variable, so remove it.


gnprice:
Thanks @jainkuniya -- merged!

PR #3310, opened 2019-01-24
Fixes #3295  chat: Render `ComposeBox` only if `canSend`.

### LINKED ISSUE #3295: Placeholder Text padding shows irregularity

jainkuniya:
Fixes #3295


borisyankov:
I fully endorse this approach!
No 'hack' involved. Just make sure to test this well, but from the looks of the code it should work fine!

gnprice:
Merged. Thanks again @jainkuniya for the debugging, the detailed notes, and the fix!

I made only small edits to the commit message; take a look.

I also squashed the two commits into one: together they have the effect of moving some logic from `ComposeBox` to `Chat`, and whenever code is being moved it's usually helpful to make the move in one commit rather than separately add and delete, so that it's self-contained to see that the old and new code match up and to understand where the new code comes from (in particular, that it wasn't newly made up.) Sometimes in complex cases it's a better tradeoff to do them separately for other reasons, but in this case a single commit works great.

Issue #3147, opened 2018-11-16
Inconsistency between activity text and activity indicator in title.


jainkuniya:
<img width="352" alt="image" src="https://user-images.githubusercontent.com/18511177/48637878-f9615e00-e9f4-11e8-8d5b-82d936047299.png">

In the above image, activity text is "Active Now", but indicator color is not green. It should be green.

@zulipbot claim

borisyankov:
I suspect the color indicator is the incorrect one 😄

borisyankov:
We might want to get rid of the intermediate yellow state altogether and be either no indicator or a green dot.

gnprice:
Thanks @jainkuniya for flagging this!

In the webapp, if I
* go through the people in the right sidebar with orange "idle" half-dots,
* click on one of the chevrons to get a user's profile,
* and hover on the orange half-dot within the profile,
* the hover info consistently says "Just now".

I suspect the behavior reported here is what we're doing in the exact same situation. In the presence system's terms:
https://zulip.readthedocs.io/en/latest/subsystems/presence.html
I think what's happening is that the user has a client which is online now... but they haven't actually interacted with the client recently.

I'd say the combination of an orange "idle/inactive" indicator (AFAICT we don't have any words within the app to explain what it means) with the text "Just now" is pretty confusing in itself, and probably isn't a fully-considered product choice.

So, we should figure out how we want this system to behave as a product matter, and describe that clearly. Then the mobile app and webapp can both implement the behavior we decide we want.

@rishig , do you have a sense of how you'd like this area to work? /cc also @timabbott who may have more to add on how the underlying system is working. It looks like zulip/zulip#11607 (an ongoing open issue) is related, too.

rishig:
yeah, I think https://github.com/zulip/zulip/issues/11607 has my updated thoughts. Agreed that "last seen" should have something like "N minutes ago" if they are not green; not sure if the current webapp behavior is a bug or a previous product decision.

timabbott:
@rishig can you collaborate with @vinitS101 on getting the design for https://github.com/zulip/zulip/pull/11882 (the PR for https://github.com/zulip/zulip/issues/11607) to something you're happy with?  I think that's probably what's blocked there.  

[...]

PR #3311, opened 2019-01-25
activityText: Fix inconsistency with user status indicator.

jainkuniya:
If `status` in `aggregated` is `idle`, then show `Idle
[presenceToHumanTime]` and likewise for `active` & `offline`.

Fixes: #3147

image

### DIFF
--- a/src/title/ActivityText.js
+++ b/src/title/ActivityText.js
@@ -24,9 +24,19 @@ class ActivityText extends PureComponent<Props> {
 
     const activity = presenceToHumanTime(presence);
 
-    return (
-      <RawLabel style={[style, color !== undefined && { color }]} text={`Active ${activity}`} />
-    );
+    const activityText = {
+      active: `Active ${activity}`,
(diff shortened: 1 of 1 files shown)


borisyankov:
That is not consistent with the goal of the activity information nor with how the feature works in FB Messenger which this was based on.

The goal is to say how long ago a user was active, not what their current activity status is.

borisyankov:
In that sense, there is no inconsistency. One can be both 'Idle' and 'Last active 5 minutes ago'.
The only problem is if they are 'Idle' and 'Active now'.

jainkuniya:
Then we need to again search for `active` in whole `presence` object and find difference.

gnprice:
Thanks @jainkuniya !

I think this isn't going to be quite the logic we want either; and "Idle now" is pretty paradoxical-sounding in itself. We'll figure out more on #3147 how we want this area to work, and then we can make the code changes to implement that.

PR #3313, opened 2019-01-25
vscode, extension: Autogenerate `generatedEs3.js` on saving `js.js`.

jainkuniya:
This is done using [Save and Run](https://marketplace.visualstudio.com/items?itemName=wk-j.save-and-run).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,4 +13,11 @@
     "[javascript]": {
         "editor.formatOnSave": true,
     },
+    "saveAndRun": {
+        "commands": [{
+            "match": "/js.js",
+            "isAsync": true,
+			"cmd": "tools/generate-webview-js"
+        }]
+    },
 }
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks! It'd be good to automate this.

I am not a fan of this implementation, though.
* It goes and pops up the "Terminal" side area, even though there's no useful information there. That's leaking the implementation into the UI.
* I took a quick look through the code of this extension, and it is fragile and buggy -- what it actually does is to paste the text of the command into the terminal window.
  * This is basically equivalent to blindly typing the command into your nearest Terminal.app window at random times! If you're lucky, or perhaps if you never use the terminal for anything else, this will work... but if there's anything else going on, it could have absolutely arbitrary effects.

VS Code has a built-in feature intended for this use case, called ["Tasks"](https://code.visualstudio.com/docs/editor/tasks). Want to try using that instead?
* A good step 1 would be to just make a "task" that runs `tools/generate-webview-js`. It won't yet run automatically on save, but you'll have the keyboard shortcut Ctrl+Shift+B (or Mac equivalent; "Run Build Task") for it.
* To make it run automatically on save, the recommended thing is to have the task itself watch for changes. We can handle this after step 1 is done.

PR #3314, opened 2019-01-25
android: Trust user-added CAs for all secure connections

### LINKED ISSUE #3312: Android app can no longer connect to secure server trusted by user-added CA
We have a Zulip server set up with a secure SSL certificate, but the certificate is trusted only by an internal Certificate Authority of ours so far, not yet a public CA.

In previous versions of the Android app, I was able to connect to the server over from the Android app as long as my phone was connected to our VPN and I had [added our internal CA](https://support.google.com/nexus/answer/2844832?hl=en) to my Android device. 

This no longer works in the latest versions of the app. Zulip now says "Cannot connect to server".

I'm fairly confident the fix is fairly simple: add a [network security configuration file](https://developer.android.com/training/articles/security-config) to the app, and [set it up to trust user-added CAs for secure connections](https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html), but I'm not sure if there's any philosophical objections to re-enabling that. Hopefully not! It has been a handy workflow while we're internally testing Zulip before we are able to expose our server publicly.

Thanks!

borisyankov:
Fixes #3312

Apps that target API Level 24 and above no longer trust user or
admin-added CAs for secure connections, by default.

This config allows the app to trust user-added certificate
authorities for all secure connections

More details here:
 * [Changes to Trusted Certificate Authorities in Android Nougat](https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html)
 * [Network security configuration](https://developer.android.com/training/articles/security-config)

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -32,7 +32,9 @@
       android:allowBackup="true"
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
-      android:theme="@style/AppTheme">
+      android:theme="@style/AppTheme"
+      android:networkSecurityConfig="@xml/network_security_config"
+    >
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
--- a/android/app/src/main/res/xml/network_security_config.xml
+++ b/android/app/src/main/res/xml/network_security_config.xml
@@ -1,0 +1,10 @@
+<network-security-config>
+  <base-config>
+    <trust-anchors>
+      <!-- Trust preinstalled CAs -->
+      <certificates src="system" />
+      <!-- Trust user added CAs -->
+      <certificates src="user" />
+    </trust-anchors>
+  </base-config>
+</network-security-config>


gnprice:
Hmm. I don't love this solution of reverting to the old defaults. The problem of securing TLS and the CA system is complex and has a number of smart people working on it, and in general I'd much rather take the solutions shipped by major browsers and other expert security teams like those at Google, rather than second-guess them.

For the request in #3312, I think my ideal solution would be to make it a user opt-in: a switch in settings somewhere. Unfortunately there doesn't seem to be a good way to control this from code -- if you need to control it from code, it looks like you're stuck with the notorious [`TrustManager` API](https://developer.android.com/training/articles/security-ssl), and I don't even see how to express in that API the equivalent of this configuration. (Surely it's there somehow? But it's going to be complicated to pin down.)

A solution that I think I'd be OK with would be to set this configuration *except* for domains that we know should be identified by certs from global trusted CAs. In particular: zulipchat.com and its subdomains, and chat.zulip.org. That should mitigate the risk substantially. In the blog post, under "Trusting user-added CAs for all domains except some" there's an example of how to do this.

What do you think?

gnprice:
(Also: thanks for sending this!)

gnprice:
Closing because stale. The underlying issue remains tracked as #3312.

PR #3316, opened 2019-01-27
safariView: Replace `\` with `/`.

jainkuniya:
While opening links consisting of `\`, `SafariView` was throwing
error. So replace it with `/`.

Fixes: #3315

### DIFF
--- a/src/utils/openLink.js
+++ b/src/utils/openLink.js
@@ -4,6 +4,7 @@ import SafariView from 'react-native-safari-view';
 
 export default (url: string): void => {
   if (Platform.OS === 'ios') {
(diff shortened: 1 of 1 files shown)


borisyankov:
I think you can simply close that.
While there are few cases where blindly replacing `\` with `/` might work, that is way too often not the case.

gnprice:
Thanks @jainkuniya !

Largely echoing @borisyankov 's reply: this change doesn't work as a solution. The new URL this makes is basically a random one that doesn't have much to do with the original -- it almost certainly won't actually work for getting the intended resource.

PR #3317, opened 2019-01-28
android, unicode urls: Fix correct url not passed to custom tabs.

jainkuniya:
Before `getFullUrl` was adding `realm` even if unicode url string
starts with `http`. This issue is completely wired. It can not be
reproducible when remote debugger is on, on Android, but can
otherwise.

Replacing `startsWith` with `indexOf` doesn't make any logically
change, but have internal implementation difference. Thus this got
fixed.

### DIFF
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -45,7 +45,7 @@ export const encodeParamsForUrl = (params: UrlParams): string =>
     .join('&');
 
 export const getFullUrl = (url: string = '', realm: string): string =>
-  !url.startsWith('http') ? `${realm}${url.startsWith('/') ? '' : '/'}${url}` : url;
+  url.indexOf('http') !== 0 ? `${realm}${url.startsWith('/') ? '' : '/'}${url}` : url;
 
 export const isUrlOnRealm = (url: string = '', realm: string): boolean =>
   url.startsWith('/') || url.startsWith(realm) || !/^(http|www.)/i.test(url);


borisyankov:
Sounds logical that this could be a fix.
Do you happen to have found this issue described or discussed somewhere online?
Would be a good addition to the commit message and would also make it more clear that this is the right fix.

gnprice:
Hmm. Can you say more about exactly what you observed?

The theory in this commit appears to be that `String.prototype.startsWith`, in the JS environment that RN pulls in, has a bug that can cause it to give a completely wrong answer.

There's a programming cliché that "it's never a compiler bug":
https://blog.plover.com/prog/compiler-error.html
https://blog.codinghorror.com/the-first-rule-of-programming-its-always-your-fault/
The cliché applies equally well to the OS, to the Android platform and iOS SDK... and to the JS standard library. :slightly_smiling_face:

Compilers do have bugs! And so do JS standard-library implementations, I'm sure. But these are systems a lot of people have banged on and a lot of people have put good work into making solid -- and so when you're tempted to say it's a compiler bug, or a JS stdlib bug, it's generally much more likely to be *your* bug, or something you don't understand yet.

And on the rare occasion you do find a JS stdlib bug... you'll want to have it well documented! So the path to resolving the situation well is basically the same either way.

borisyankov:
The most likely place this bug can be is in the polyfill:
https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/polyfills/String.prototype.es6.js

borisyankov:
... or more likely a JSC issue!

Very interesting thread here:
https://github.com/facebook/react-native/issues/11370

Looks like JSC will be upgraded in RN 0.59 
https://github.com/facebook/react-native/issues/19737

And that would be a very likely fix for our issue.

jainkuniya:
Let's close this. As I am not able to reproduce the issue. And there is no report for any such or similar issue from users. I am actively testing this and re-open it, if this is still a issue.

Actually the issue was only on Android, `startsWith ` was not working as intendant.

PR #3320, opened 2019-01-29
Fix dev auth login error display

borisyankov:
Since we introduced `makeApiError` the error displayed to the user
is saying simply 'API'. This change extracts the error message the
server returns, which is in `err.data.message`.


borisyankov:
@gnprice ping for a review for this small PR.

gnprice:

Issue #3321, opened 2019-01-30
Emoji being shown twice when enclosed within a bold or italic markup


SivagiriVisakan:
When a message is sent with an emoji wrapped inside a bold or italic markup , for example as `** :tada: some text**`, then the emoji is shown twice.(see screenshot attached)

![image](https://user-images.githubusercontent.com/27341353/51995219-4e7f1d00-24d8-11e9-8474-b2a2ac2c3759.png)

See discussion: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Emoji.20issue.20when.20enclosed.20within.20bold.20or.20italics.20markup

jainkuniya:
This is a server side issue #11621. So will be fixed there, no need to do any changes in the `zulip-mobile` codebase.

gnprice:
Fixed on the server, as zulip/zulip#11621 . Thanks @SivagiriVisakan for the report!

Issue #2693, opened 2018-06-18
Emoji cut off at bottom


gnprice:
A user reported that it "seems iOS app emoji are being cut off at the bottom", providing the following screenshot:

![](https://chat.zulip.org/user_uploads/2/e/jxuYnpeMdQaBZDGbPndSYV49/IMG_0054.PNG)

Looks like this is a message containing only an emoji, as the last message shown. Not sure exactly what range of conditions lead to the bug, but it shouldn't be hard to reproduce and play with.

jackrzhang:
Hm, the bug is not present in the generated HTML/CSS from `MessageListWeb`, which makes me think this has to be specific to iOS.

armaanahluwalia:
Sorry for not being more descriptive. This was a private message inbox with one other user ( hence the blacking out by me ). The emoji was the last message received from the other user.

gnprice:
We learned more about this in #2706 ; see comment thread there, and [this chat discussion](https://chat.zulip.org/#narrow/stream/2-general/topic/Unicode.20Emoji.20Accessibility).

PR #2706, opened 2018-06-20
webview: Layout fixes.

jackrzhang:
See #2696, #2693.

PR #3325, opened 2019-02-04
webView, emoji, css: Fix Emoji rendering issue (twice, cut down).

jainkuniya:
Use `overflow: hidden` in the emoji `span` tag (webapp also uses it)
and increase `width` & `height` to `20`.

Fixes: #2693, #3321.

Before:
On iPhone 5s
image

On iPhone X
image

This PR:
On iPhone 5s
image

iPhone X
image

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -343,8 +343,8 @@ blockquote {
 }
 .emoji {
   display: inline-block;
-  height: 18px;
-  width: 18px;
+  height: 20px;
+  width: 20px;
   white-space: nowrap;
   color: transparent;
   vertical-align: text-top;


borisyankov:
Why do we need `overflow: hidden;`?

jainkuniya:
Because `span` which contains `emoji` class also has text in the `span` tag. 

```
<span aria-label="smiley" class="emoji emoji-1f603" role="img" title="smiley">:smiley:</span>
```

So to avoid **:smiley:** get's rendered, we want `overflow: hidden;`

web app handles this like
```
{
    display: inline-block;
    background-image: url('sheet-google-64.png');
    -webkit-background-size: 5200%;
    -moz-background-size: 5200%;
    background-size: 5200%;
    background-repeat: no-repeat;

    /* Hide the text. */
    text-indent: 100%;
    white-space: nowrap;
    overflow: hidden;
}
```

borisyankov:
But we are not rendering the `:smiley:` now either!?!

jainkuniya:
Because of the current width and height in the css CSS, it is not being
rendered.

borisyankov:
Does it render in 20px x 20px ?

gnprice:
From the snippet of the webapp's CSS quoted above:
```
    /* Hide the text. */
    text-indent: 100%;
    white-space: nowrap;
    overflow: hidden;
```
Isn't `text-indent: 100%` the key bit there? What does `overflow: hidden` do for us without that?

In our CSS, we have `color: transparent`. I think that's the bit that is hiding the text for us.

gnprice:
Apart from the `overflow: hidden` bit:

* Why 20px? Do you know a reason to believe that that ought to be enough? Or did you just try it on a couple of devices and find that 20px was enough, while 18px wasn't?

  If the latter, the change is still an improvement and good to merge. But it's important to know the difference. :-) Also, it'd be good to record exactly why we chose this value / believe it's good: "Tested on an iPhone 5s and an iPhone X, and this size is enough."

gnprice:
Oh huh, and I just went and read #3321. Your screenshots don't illustrate that part. :slightly_smiling_face: 

Do you believe there's an interaction of some kind between #3321 and #2693? At first glance, to me they look completely unrelated. They should be separate commits, unless there's some reason they can't be. That way, each commit can make clear what problem *that* change is intended to solve, and why.

I'm guessing now that you added the `overflow: hidden` line because you found that it resolves #3321? I'd like to understand that better before merging a fix. Why does #3321 appear when bold or italics is involved, but not otherwise? What exactly is the effect of this `overflow: hidden`, and how is it connected to the issue?

gnprice:
(Also: thanks @jainkuniya for these fixes and the screenshots!)

gnprice:
Oh, and see also this previous attempt at fixing #2693:
https://github.com/zulip/zulip-mobile/pull/2706/commits/6a362aa72b07bcb7214da2a9e7c3a81cda903f48
and the discussion on that PR thread.

borisyankov:
@gnprice @jainkuniya I just spent an hour trying to figure out the issues and trying to improve our styling. Here are my findings:

We should not increase randomly the `width` and `height` as it does mess up the line height (for example `25px` is terrible etc) But `20px` is OK as 1.4 line-height * 15px = 21 < 20px.

So, the commit: [webview, css: Fix emoji cut off at bottom.](/zulip/zulip-mobile/pull/3325/commits/595b41b7dab80cd1cfc0d286d5a8d688ebbea4d5)
is good and we should merge it.

borisyankov:
That being said, on Safari, the emojis are not being cut off without it. I guess it is needed on Chrome.

borisyankov:
Now, the other commit: [webview, css: Fix emoji in bold gets rendered twice.](/zulip/zulip-mobile/pull/3325/commits/8930bd6c2dc7223c8cd8496ac848398088a07f74)

Adding `overflow` actually cuts off the emoji if it does not fit in the height-width size.

Addressing the commit message:
> Use `overflow: hidden` in the emoji `span` tag (webapp also uses it).

This is really not a good reason to bring up. The web app does this completely differently. They have an emoji sprite. We do use the system font. Also the contents are in an `.emoji:before` (see `cssEmojis.js`)

borisyankov:
```html
<span aria-label="tada" class="emoji emoji-1f389" role="img" title="tada">
    <span aria-label="tada" class="emoji emoji-1f389" role="img" title="tada">
        🎉
    </span>
</span>
```

That looks like a back-end issue. I can not imagine this is a correct markup that we should be expected to handle.

gnprice:
* Thanks @jainkuniya for the split and the additional information in commit messages! That's very helpful.

* I think you meant to mention #2693 in the first commit. :wink: 

* Small thing in that commit message:

  > Tested on few iPhone 5, x and Android 9

  From earlier in this discussion I know you tested it on an iPhone X, so I think that's what you mean by the "x". But if I didn't already know that, I wouldn't have understood. Clearer to just write explicitly, even though it looks a little redundant: "iPhone 5, iPhone X, and on Android 9".

* But earlier you said "iPhone 5S". That's not the same thing as an iPhone 5, and it's not really even a kind of iPhone 5; it just has a similar name. Do you mean iPhone 5S?

* (Separately, the version of iOS is probably more relevant than the model of the device.)

> @gnprice @jainkuniya I just spent an hour trying to figure out the issues and trying to improve our styling. Here are my findings:
> 
> We should not increase randomly the `width` and `height` as it does mess up the line height (for example `25px` is terrible etc) But `20px` is OK as 1.4 line-height * 15px = 21 < 20px.
> 
> So, the commit: [webview, css: Fix emoji cut off at bottom.](/zulip/zulip-mobile/pull/3325/commits/595b41b7dab80cd1cfc0d286d5a8d688ebbea4d5)
> is good and we should merge it.
> [...]
> That being said, on Safari, the emojis are not being cut off without it. I guess it is needed on Chrome.

* Thanks @borisyankov for that investigation -- this is very helpful information!

  I'd add only: that means we should merge that commit, *after* adding to it this additional information. :smile:

  Also that's very interesting that it differs between Chrome and Safari. I wonder what's the root cause of the difference there.

gnprice:
> > `text-indent: 100%;` in `.emoji` class
> 
> Not needed. Because it is indenting image too.
> 
> With ` text-indent: 100%;` [... image ...]
> 
> without ` text-indent: 100%;` [... image ...]

* Right, sorry, to be clear: I don't mean that we should add `text-indent` to our CSS. We have a different scheme than the webapp does for showing these emoji.

  I mean that I don't see why we should add the `overflow: hidden`. Because, again, we have a different scheme than the webapp does for showing these emoji. :slightly_smiling_face: 

  And specifically, I was analyzing how the webapp's CSS works, and I was saying that the key bit of it (or anyway *a* key bit of it) that causes the text to get hidden is the `text-indent: 100%`. Without that, there's no overflow, so there's nothing for `overflow: hidden` to act on in the first place.

gnprice:
* As for the #3321 case, where it's enclosed in bold -- I have the same comment as @borisyankov . (Including that HTML in the commit message was very helpful for explaining your diagnosis -- thanks!) That is clearly a bug in the server, that just happens to not have any effect in the webapp.

  Would you go file a bug in zulip/zulip explaining that given a certain input, it produces that HTML you quoted with the duplicate `emoji` spans? And mentioning zulip/zulip-mobile#3321 so things get properly cross-referenced.

  Then it can get fixed on the server side.

  Thanks for tracking that down!

jainkuniya:
Filed https://github.com/zulip/zulip/issues/11621 for backend issue.

gnprice:
And merged! After adding @borisyankov 's observations above to the commit message. Thanks again to you both :smile: 

(And thanks @jainkuniya for filing that server issue!)

Issue #3324, opened 2019-02-04
Self-PM appears in other PM conversations


jainkuniya:
Steps to reproduce.
* Open any PM narrow other than self, so that `newer` `caughtup` for that narrow become true.
* Now open self narrow and send new message.
* Now again  narrow back to same narrow which you picked in step 1.

**Observation**
The message which we posted in step two (to self) is also visible in other narrow (which we picked in step 1,3)

Restart the app, all works fine.

Issue is why message got added to list of other narrow?

Because there is an issue in our `matchRecipients` of `isMessageInNarrow` in `narrow.js`.

As for all PM narrow, `normalizedRecipients === ownEmail` is true for self message.
Because in this condition we are not taking care of `narrow`, it just compare recipient in message and `ownEmail`.

@zulipbot claim

Will send a PR soon for this.

borisyankov:
Is that a bug though? This is how that works in the web app.

jainkuniya:
@borisyankov By "all private narrow" , I doesn't mean "All private message".
I mean all other private narrows.

(editing title to ` all other private narrows` )

gnprice:
Thanks @jainkuniya !

Following up on my comments on #3328, here's what I think would make a good fix for this bug:
* Add some unit tests that clearly demonstrate the bug. They should fail if run on master.
* I think the root cause of the bug is that `matchRecipients` is too complicated. So let's fix that.
* I think one of the reasons it became too complicated is that it's handling both 1:1 PMs and group PMs. Let's write each case separately.
* The group case can be a lot simpler than that `matchRecipients` function. Try writing it very simply.
* The 1:1 case can also be simpler. Try writing it out more explicitly. I think a conditional like `email === ownEmail ? ...` may help.

gnprice:
FTR, although #3328 refers to a "regression in `matchRecipients`", this bug is much older, and older than that function. It dates back to 2017-01 with commit d8faae9d7; it's present at that commit, and absent at its parent.

Before that point, the code had the opposite bug on self-1:1 messages: they didn't appear even in their own conversation narrow. That commit aimed to fix that, and introduced the puzzling `normalizedRecipients === ownEmail` condition mentioned above, and with it introduced this bug.

PR #3328, opened 2019-02-05
isMessageInNarrow: Fix regression in `matchRecipients`.

jainkuniya:
Remove `normalizedRecipients === ownEmail` condition, as it is not
taking `narrow` in consideration. For self message, `display_recipient`
will only consist of `ownEmail`. Thus for calling `isMessageInNarrow`
with self message and any private narrow was returning `true` (because
of `normalizedRecipients === ownEmail`). So remove this condition.

Also improve `normalizedNarrow`, so that it consist of only unique
values. Now this change will take care of logic which was intended to be
taken by `normalizedRecipients === ownEmail` (which is getting removed,
by this commit).

Fixes: #3324

### DIFF
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -1,6 +1,7 @@
 /* @flow strict-local */
 import isEqual from 'lodash.isequal';
 import unescape from 'lodash.unescape';
+import uniqby from 'lodash.uniqby';
 
 import type { Narrow, Message } from '../types';
 import { normalizeRecipients } from './recipient';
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @jainkuniya ! This bug will be good to fix.

I don't love this solution. The existing code is too complicated and lacks conceptual clarity, which is the root cause of the issue -- and the same thing is still true after this change.

One way to see that: what does this `matchRecipients` function mean? How would you describe it in a jsdoc comment? I think it will be challenging to write a description that is completely accurate and precise and yet is reasonably short, because the function's meaning is somewhat confused. (Which, again, was already true before this change. :slightly_smiling_face: )

Relatedly: a bugfix like this definitely calls for unit tests. :stuck_out_tongue: There should be a test added that demonstrates the bug, and would fail on the old code.

gnprice:
(Let's continue on the issue thread.)

PR #3329, opened 2019-02-05
composeBox, UI: Paint the safe area with BG color of composeBox.

jainkuniya:
Master:
image
image

This PR:
image
image


gnprice:
Thanks @jainkuniya for this fix! Looks great.

Merged, with just one adjustment to the commit message: the "safe area" is the area on the inside of the screen, i.e. the opposite of the part this change is about. :wink: It's called the "safe area" because it's the area that's nice and safe to draw whatever we want in because no notches, etc., are invading it.

See e.g. in Apple's docs:
https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/
image

PR #3330, opened 2019-02-05
webview, openLink: Get the AWS res file path and open it in browser.

jainkuniya:
Auth needs to be passed to access user_upload, or we can also access it
via appending `api_key` to the url. So append the `api_key` to the url,
and then get the redirected url (which is of AWS). And finally open it
in browser, so that user don't have to login to realm in browser.

So the external browser only sees the "expires in 1 minute" S3 URL for
accessing the file. And the `api_key` is also not exposed in the url via
accessing the file in browser.

https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/view.20uploaded.20files

### DIFF
--- a/src/message/messagesActions.js
+++ b/src/message/messagesActions.js
@@ -23,7 +23,10 @@ export const doNarrow = (narrow: Narrow, anchor: number = FIRST_UNREAD_ANCHOR) =
   dispatch(navigateToChat(narrow));
 };
 
-export const messageLinkPress = (href: string) => (dispatch: Dispatch, getState: GetState) => {
+export const messageLinkPress = (href: string) => async (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
   const state = getState();
   const auth = getAuth(state);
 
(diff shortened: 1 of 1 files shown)


src/message/messagesActions.js
```diff
@@ -34,6 +34,24 @@ export const messageLinkPress = (href: string) => (dispatch: Dispatch, getState:
...
+      const srcPath = url.substring(auth.realm.length);
+      if (
+        srcPath.startsWith('/user_uploads/')
+        || srcPath.startsWith('/thumbnail?')
+        || srcPath.startsWith('/avatar/')
+      ) {
+        const delimiter = url.includes('?') ? '&' : '?';
+        url += `${delimiter}api_key=${auth.apiKey}`;
+
+        fetch(url).then(response => {
```
borisyankov: async + await > then

src/message/messagesActions.js
```diff
@@ -34,6 +34,24 @@ export const messageLinkPress = (href: string) => (dispatch: Dispatch, getState:
...
+          openLink(response.url);
```
borisyankov: This is likely not enough. I don't thick all realms' resources are behind a redirect to URL AWS.

src/message/messagesActions.js
jainkuniya: Done 👍

chrisbobbe:
To clear a path forward on this important issue, I'm closing both this PR and #3495, with a solution outlined at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/view.20uploaded.20files/near/819777 that will involve some backend work.

There, Greg lists these three requirements that the new solution fulfills, and offers an implementation plan:

```quote
The goal basically is to get
* (a) a plain old URL that can be passed to a browser (not a URL plus some cookies or other headers that need to be set)
* (b) that gets the user the file immediately, without requiring a fresh login
* (c) but doesn't include a secret with much wider powers (i.e. their API key), which would risk a nasty surprise if they innocently show someone else the file by sending them the link.
```

Issue #3303, opened 2019-01-22
Viewing uploaded files fails, or requires logging in again


zingmars:
Right now if I share a file (i.e. a document) and click on the link that's sent in the Android app it will open a webview asking me to log into the Zulip instance. This is pretty annoying to be honest as the expected behaviour here would be to offer me to download the file instead.

drumilpatel2000:
@zulipbot  claim

borisyankov:
I consider the original issue to be a **bug report** not a **feature request**

For documents (for example PDFs) that are not hosted on a zulip server, we do open the document in a Chrome Custom Tab / Safari View. This screen immediately displays the document. In addition, a menu would allow downloading the document.

This matches the behavior of several chat apps that I tested, including Slack (also several non-chat apps!)

So, we need to fix the auth issue, not add a direct download functionality, which already exists in the screen that shows the document.

gnprice:
Thanks @borisyankov for the investigation! I agree with your conclusions.

Folding #3487 into this as a dupe -- you've determined it's the same underlying issue, with a difference in symptoms by platform:
> * on Android - after opening fails we go back to the message list
> * on iOS - we get a login screen

gnprice:
> you've determined it's the same underlying issue, with a difference in symptoms by platform:
> 
> > * on Android - after opening fails we go back to the message list
> > * on iOS - we get a login screen

Hmm actually this doesn't quite fit the data -- the original report here on this thread was on Android, and described a login screen.

I still think these are the same issue so I'll dedupe them; we'll just keep in mind that we have a somewhat confusing set of reports.

gnprice:
Current state:

* We believe all the reported symptoms are ultimately caused by trying to open an auth-requiring URL and not providing auth; similar to the cause of #3124 .
  * Background: normal auth in the mobile app is by including an `Authorization` header, i.e. by HTTP basic auth. We do this for all API requests made directly by our own code, but it's trickier when other libraries are making HTTP requests. Some we successfully teach to send that header; for others, we resort to sticking the user's API key in the query-part of the URL.

* Reported symptoms include:
  * on Android, a login screen (in the OP on this thread);
  * on Android, a "Cannot display PDF" toast (on #3487 -- my repro, and a similar report from a user)
  * on Android, it all working! (on #3487, report from a user; details unknown)
  * using chat.zulip.org (from Android), it all works fine (my repro, on #3487)
  * on iOS, a login screen (Boris comment on #3487)

[...]

PR #3331, opened 2019-02-07
 msg list: Download file on clicking file url.

jainkuniya:


gnprice:
Thanks @jainkuniya ! Quick comments (copying what I told you today in PMs and audio):
* This needs more explanation. What was the old behavior; what have you learned about it; what changes in the behavior with your code changes; and why are those changes the right ones?
* I think #M3487 is related. Discussing that relationship would be especially helpful. (Perhaps it's even exactly the issue this fixes?)
* Also please fix the build :slightly_smiling_face:

borisyankov:
This is implementing #3303

src/utils/url.js
```diff
@@ -172,3 +172,9 @@ export const autocompleteUrl = (value: string = '', protocol: string, append: st
     : '';
 
 export const isValidUrl = (url: string): boolean => urlRegex({ exact: true }).test(url);
+
+export const isUrlFilePath = (url: string) =>
```
borisyankov: This is certainly not robust enough.
For example, `mysite.com/welcome.php` does match it but should not be downloaded.

Usually, the distinction between a downloadable file and viewable file is made based on the mime type, and often the server can 'enforce' (or maybe more correctly 'suggest') the downloadability of a file.

borisyankov:
It is not clear to me if the behavior, even implemented perfectly, is desirable. This does not match the behavior of serval chat apps I did test, including Slack

borisyankov:
I consider this a better (though different) solution to the same problem #3495

src/utils/url.js
jainkuniya: Yup! I agree it's not robust.

But to get ` mime type` (from header) we need to do a fetch (send a request), and let's say if file is 1mb, then first 1mb (and time) is gone for checking ` mime type` and if it is downloadable file then download it (again 1mb + time).

gnprice:
Cool, thanks @borisyankov for the review and the debugging on #3303 , and thanks @jainkuniya for the updates.

I agree with Boris that this isn't the right direction for a fix; we'll continue on the issue #3303 and on #3495 .

PR #3332, opened 2019-02-08
WIP:  Rename 'getAllUsersByEmail' to 'getUsersAndBotsByEmail' and type it

borisyankov:
I wonder what would be the most practical way to handle the:
`Cannot get user.avatar_url because property avatar_url is missing in RealmBot [1].`

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -2,7 +2,7 @@
 import React, { PureComponent } from 'react';
 import { View, Dimensions, StyleSheet } from 'react-native';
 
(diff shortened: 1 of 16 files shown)


gnprice:
Thanks @borisyankov !

Did you look at the history of this file? I mentioned in https://github.com/zulip/zulip-mobile/pull/3326#discussion_r254864735 that there is some relevant background there:
`git log --stat -p src/users/userSelectors.js`

We can certainly change things, but the commit that makes the change has to point to the previous discussion and explain why we're doing something different.

Issue #3334, opened 2019-02-10
lightBox: Sudden splash of the header and footer in & out on pushing route.


jainkuniya:
Whenever user routes to `Lightbox` screen, just after component renders there is a sudden splash in the footer and header. Header and footer comes into the screen and goes out 2 times before settling down.

@zulipbot claim I have diagnosed this issue and will submit PR soon.

EDIT
To confirm:
* Put a `console.log` in the `render` method. This `log` can be observed multiple times just after pushing route and on every event we get from server.

PR #3335, opened 2019-02-11
Fixes: #3334 Fix splash of header & footer in light-box screen pushing route and getting new event from server.

jainkuniya:
Fixes: #3334

### DIFF
--- a/src/account/accountsSelectors.js
+++ b/src/account/accountsSelectors.js
@@ -100,13 +100,12 @@ export const getPartialAuth: Selector<Auth> = createSelector(getActiveAccount, a
  *  * `getPartialAuth` for use in authentication flows, where there is an
  *    active account but it may not be logged in.
  */
-export const tryGetAuth = (state: GlobalState): Auth | void => {
-  const account = tryGetActiveAccount(state);
+export const tryGetAuth: Selector<Auth | void> = createSelector(tryGetActiveAccount, account => {
   if (!account || account.apiKey === '') {
     return undefined;
   }
   return authOfAccount(account);
-};
+});
 
 /**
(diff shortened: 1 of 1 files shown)


gnprice:
On the same principles I described in #3337: caching/memoizing `tryGetAuth` makes sense because it calls `authOfAccount`, making a new object.

Caching `getAuth`, on the other hand, doesn't -- it just returns the object it gets from `tryGetAuth`.

Similarly `tryGetActiveAccount`. Can you explain why you believe that one fixes #3334?

gnprice:
More specifically: I believe that the only one of these three that should have any real effect is the one on `tryGetAuth`.

Would you check to see if just that commit alone fixes #3334?

gnprice:
Thanks @jainkuniya for that additional testing! Sorry I seem to have missed the update, and thanks for the ping.

Two things which are important but easy to fix, and so I'll do them and then merge:
* Our selectors should have types. :smile: This one actually did in the existing code, and with your change it loses the type annotation:
```
-export const tryGetAuth = (state: GlobalState): Auth | void => {
-  const account = tryGetActiveAccount(state);
+export const tryGetAuth = createSelector(tryGetActiveAccount, account => {
```

* The commit message no longer makes a lot of sense, now that the previous commit in the branch has been dropped:
```
    accountsSelectors: Use `reselect` in `tryGetAuth` selector.
    
    In continuation to previous commit, this too helps in saving lots to
    redundant re-renders.
```
  In particular the reference to #3334 was very helpful in understanding why this change is made; so now that the previous commit isn't there, this one should mention it directly.

gnprice:
And merged -- thanks again!

PR #3336, opened 2019-02-11
accountPickerScreen: Fix re-render on every event of long poll.

jainkuniya:
Use `reselect` in `getAccountStatuses`, so that it memorize accounts and
return same object for same value and saves re-render.


gnprice:

PR #3337, opened 2019-02-11
 accountsSelectors: Use `reselect` in the remaining selector (left out from #3335, #3336)

jainkuniya:


gnprice:
Thanks @jainkuniya !

Reading through these, most of them are examples where there isn't any benefit to caching. E.g. here:
```
-export const hasAuth = (state: GlobalState): boolean => !!tryGetAuth(state);
+export const hasAuth: Selector<boolean> = createSelector(tryGetAuth, auth => !!auth);
```
If `tryGetAuth` returns an object, we're going to return `true`; and nothing `createSelector` can do will be cheaper than just going ahead and computing `!!auth`.

I think in fact all of these are examples where the selector body is very cheap to compute. There is a reason that some of them are useful, though -- when the selector is constructing a new object, then caching can let us return the identical object and save a re-render. I'll go ahead and merge the commits where that applies.

gnprice:
And merged! Took 2 of the commits and dropped 3; take a look.

Issue #3339, opened 2019-02-12
Avoid linear scans through all users, streams, etc.


gnprice:
There are a number of places where we make a linear search through an array of potentially thousands of items, when we really just want a specific item with a known ID or name.

This is a pattern that can be devastating for performance -- in particular, as soon as it happens in a loop, we have quadratic behavior. This is a common class of performance bug found in lots of software:
https://accidentallyquadratic.tumblr.com/post/113840433022/why-accidentally-quadratic
and we should avoid it.

Here's a rough catalog of where we do this:
```
$ git grep -ohP '\w+\.find\(.*?\)' src/ | sort | grep --color '^\w*'
actionSheetButtons.find(x => x.title === title)
allStreams.find(stream => stream.stream_id === subscription.stream_id)
allUsers.find(u => u.email === narrow[0].operand)
allUsers.find(x => x.email === email)
allUsers.find(x => x.email === r)
handlers.find(x => x.isFunc(narrow)
messages.find(msg => !isMessageRead(msg, flags, subscriptions, mute)
messages.find(x => x.id === messageId)
messages.find(x => x.id === messageId)
messages.find(x => x.isFunc(narrow)
mute.find(x => x[0] === stream.name && x[1] === name)
selected.find(user => user.user_id === item.user_id)
selected.find(x => x.email === email)
state.find(item => item.timestamp === action.outbox.timestamp)
state.find(y => x.stream_id === y.stream_id)
state.find(y => x.stream_id === y.stream_id)
streams.find(s => s.name === narrow[0].operand)
streams.find(s => s.name === narrow[0].operand)
streams.find(s => s.name === narrow[0].operand)
streams.find(s => s.name === narrow[0].operand)
streams.find(s => s.name === narrow[0].operand)
streams.find(stream_ => narrow[0].operand === stream_.name)
streams.find(sub => narrow[0].operand === sub.name)
streams.find(x => x.name === narrow[0].operand)
streams.find(x => x.name === narrow[0].operand)
streams.find(x => x.name === narrow[0].operand)
streams.find(x => x.name === narrow[0].operand)
streams.find(x => x.name === narrow[0].operand)
streams.find(x => x.stream_id === streamId)
streams.find(y => x && x.stream_id === y.stream_id)
subscriptions.find(sub => narrow[0].operand === sub.name)
subscriptions.find(sub => streamName === sub.name)
subscriptions.find(x => x.name === display_recipient)
[...]

Issue #3015, opened 2018-10-01
Fix caching in "for narrow" selectors


gnprice:
When we want to pass an argument to a selector, like `props.narrow` to `getRenderedMessages`, we currently do so in a way that totally defeats Reselect's caching. We should fix this.

See [discussion in chat](https://chat.zulip.org/#narrow/stream/48-mobile/subject/Excessive.20Rendering.20Bug/near/649036). In particular, on what to do instead:

> The solution recommended in the Reselect docs is to push the inspection of `props` inside the selector:
https://github.com/reduxjs/reselect#accessing-react-props-in-selectors
>
> I don't love that solution -- it feels awfully implicit to me. [...] I'd much rather stick to passing specifically the narrow down to the selectors, like we currently do.
>
> I guess there's nothing in the Reselect code (AFAICT) that depends on that "props" argument *actually* being the `props` of any React component. So we could just say `getRenderedMessages(state, props.narrow)`, and have all the `...ForNarrow` selectors take a second argument of type `Narrow`, and that may just all work great.
>
> (We'd be treating that `Narrow` as what the Reselect docs expect to be "props".)

So instead of defining selectors so that their call sites look like `getRenderedMessages(props.narrow)(state)`, we should define them so they're called like `getRenderedMessages(state, props.narrow)`.

I don't think this is currently a top priority, though.

gnprice:
An update on how to approach this, something I learned from studying `reselect` and `react-redux` after reviewing #2450: in many of these selectors, we really wouldn't get a meaningful benefit from `createSelector` even if we used it correctly! For example:
```
export const getFetchingForActiveNarrow = (narrow: Narrow) =>
  createSelector(getFetching, fetching => fetching[JSON.stringify(narrow)] || NULL_FETCHING);
```
This should just be
```
export const getFetchingForActiveNarrow = (state: GlobalState, narrow: Narrow) =>
  getFetching(state)[JSON.stringify(narrow)] || NULL_FETCHING;
```
(Also it should be renamed by s/Active//.) Even if we used `createSelector` correctly, what the resulting selector would do when called is basically
* (We're here because either `state` or the component's `props` has changed.)
* Check if either `state` or `narrow` has changed.
* If so, recompute `fetching`; else we're done.
* Check if either `fetching` or `narrow` has changed.
* If so, recompute the final result.

Whereas the simpler thing, when called, does
* (We're here because either `state` or the component's `props` has changed.)
* Compute `fetching`.
[...]

PR #3341, opened 2019-02-12
part of #3339. Migrate from get*User to get*UserById / get*UserByEmail for efficient search. Also fix caching in typingSelector. part of #3015

jainkuniya:


gnprice:
Thanks @jainkuniya !  Comments below.

> getActiveUsersByEmail

* This can use `getActiveUsers` for the first step, rather than repeating what that one does.
* I'm not a fan of this `reduce` pattern; I think it's generally less clear than alternatives. Here, let's use the pattern demonstrated in `getUsersById` and `getUsersByEmail` (which you will note are a lot shorter :smile: )


> getUsersSansMeByEmail

Hmm. This is very similar to `getUsersByEmail`. Rather than have a separate version of this data structure, I think it will make more sense for calling code to just use `getUsersByEmail`, and just not look up the "self"/"me" user.

Yeah, for example, UserPickerCard can just use `getUsersByEmail` here. It's fine that the map has some entries in it which it won't look at.

> getCurrentTypingUsers

```
-export const getCurrentTypingUsers = (narrow: Narrow): Selector<User[]> =>
-  createSelector(getTyping, getUsers, getOwnEmail, (typing, users, ownEmail) => {
+export const getCurrentTypingUsers: Selector<$ReadOnlyArray<User>, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  getTyping,
+  getUsers,
+  getOwnEmail,
+  (narrow, typing, users, ownEmail): User[] => {
```

Cool, so, there's one subtlety here which you'll see if you look back at the commits I made recently and linked in #3015:
* As is, this will cause Reselect to invoke e.g. `getTyping` like `getTyping(state, narrow)`.
* That happens to be harmless, as `getTyping` will just ignore the second argument...
* ... but it's kind of worrying. What if we accidentally use a selector here that does look at the second argument, and it has an effect we weren't thinking of here?
* So I've been writing `state => getTyping(state)`, to avoid that.


> typingSelector: Use `getUsersById` for fast user search.

```
+    // $FlowFixMe Cannot return currentTyping.userIds.map(...) because inexact undefined is incompatible with exact User in array element.
+    return currentTyping.userIds.map(userId => usersById.get(userId));
```

Hmm, this looks like a real type error! Let's not paper over it with a `$FlowFixMe`. :wink: 

What is the error telling you? What possible situation is it describing?

> getNarrowFromLink: Use `usersById` for efficient user search.

-      recipients.map((recipient: string) => getUserById(users, parseInt(recipient, 10)).email),
+      recipients.map((recipient: string) => (usersById.get(parseInt(recipient, 10)) || {}).email),

What's the effect of the `|| {}`? Is that the effect you want, and why?

For this commit, one reasonable approach would be to keep the exact same behavior in this case as the existing behavior, in `getUserById`. That could be done by using `NULL_USER` instead of `{}`.

gnprice:
Thanks @jainkuniya for these revisions!

> userSelectors: Refactor `getActiveUsers` to return map by email.

Ah, interesting that we don't seem to have any usages of `getActiveUsers` that don't actually just want a map!

* I like keeping the naming convention consistent and boring, though. So I'd rather call this selector `getActiveUsersByEmail`, even if there doesn't happen to be a `getActiveUsers` exported.

* Followup point: why do we use "active users" here instead of "all users"? What is the situation where the difference between the two has an effect?
  * ... I think the answer is that in that situation we have a bug, and the fix to the bug is to switch to `getAllUsersByEmail`. If you would test that out, file an issue, and send a PR to fix it, that would be great.

```
user-picker-screen: Avoid linear scans through all users.
typingSelectors: Fix caching in "for narrow" selectors.
typingSelector: Use `getUsersById` for fast user search.
userHelpers, clean: Remove `userById` method.
```

These look good, thanks!

* A further followup: as I suggested above, these commits propagate `NULL_USER` to a couple of sites where we had been using `getUserById`. What is the effect of `NULL_USER` in each of those places? Let's write the behavior explicitly... and then decide if that's really the fallback behavior we want, and write what we want instead.
  * In `getNarrowFromLink`, I'm pretty sure the effect of the fallback is in fact completely broken. Hmm, that one is fixed in #3240, which seems to be stalled.
  * In `getCurrentTypingUsers`, the effect is also pretty broken. I think the fix is to just drop any users we can't find.

gnprice:
And merged -- thanks again! I added one commit to address my comment on naming above.

See the followup notes in my previous comment -- would be glad to see those.

Issue #2787, opened 2018-07-11
Typing certain emoji crashes the app


borisyankov:
This happens if using `Gboard - the Google Keyboard`.
This is reported as being the fix:
https://github.com/facebook/react-native/commit/41975f75d96ef4b606b4618461bf24d5db063b77 

The commit is included in React Native 0.56

gnprice:
Sweet -- glad this is already fixed upstream!

Looks like https://github.com/facebook/react-native/issues/18405#issuecomment-373624413 is the RN issue report. (Followed a link from the commit message on the commit mentioned above; and the repro steps look like a perfect match.)

Though... hmm. That issue is still open.

... OK, but the linked commit (which came from https://github.com/facebook/react-native/pull/18443) is supposed to be a workaround that applies if you don't pass an `onKeyPress` prop to your `TextInput`. Which we currently don't, so I guess when we upgrade to RN 0.56 that will fix the issue for us.

gnprice:
BTW that upstream issue thread helps clarify what the conditions are that trigger the issue: it looks like it's not specific to Gboard, but it happens on Android when you use any keyboard to type... actually any character outside [the Unicode BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane), i.e. a character that doesn't fit into one Java `char`. In practice that mostly means an emoji that isn't one of the oldest emoji.

gnprice:
Unassigning because I believe there isn't any ongoing work to do on this issue; it should just get fixed when we upgrade RN. (Though part of doing that upgrade is we should confirm that.) Thanks again @borisyankov for writing up the issue and for the research!

I just filed #2788 to track the RN upgrade. If you want to take that, please do :)

gnprice:
Also: thanks @adnrs96 for originally reporting this issue!

gnprice:
... Hmm, on the one hand I should have reread the thread more closely:

> BTW that upstream issue thread helps clarify what the conditions are that trigger the issue: it looks like it's not specific to Gboard, but it happens on Android when you use any keyboard to type... actually any character outside [the Unicode BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane), i.e. a character that doesn't fit into one Java `char`. In practice that mostly means an emoji that isn't one of the oldest emoji.

[...]

PR #3344, opened 2019-02-14
UI set away status

borisyankov:
Adds a Profile tab and allows setting user's away status and status text.

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { View, Dimensions, StyleSheet } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { connect } from 'react-redux';
 
 import type { GlobalState, User } from '../types';
@@ -22,6 +22,8 @@ const componentStyles = StyleSheet.create({
   },
 });
 
+const AVATAR_SIZE = 200;
+
 type Props = {|
   realm: string,
   user: User,
@@ -30,16 +32,17 @@ type Props = {|
 class AccountDetails extends PureComponent<Props, void> {
   render() {
     const { realm, user } = this.props;
-    const screenWidth = Dimensions.get('window').width;
 
     return (
       <View>
-        <UserAvatar
-          avatarUrl={getAvatarFromUser(user, realm, screenWidth)}
-          size={screenWidth}
-          shape="square"
-        />
         <ComponentList outerSpacing itemStyle={componentStyles.componentListItem}>
+          <View>
+            <UserAvatar
+              avatarUrl={getAvatarFromUser(user, realm, AVATAR_SIZE)}
+              size={AVATAR_SIZE}
+              shape="circle"
+            />
+          </View>
           <View style={componentStyles.statusWrapper}>
             <PresenceStatusIndicator email={user.email} hideIfOffline={false} />
             <RawLabel style={[styles.largerText, styles.halfMarginLeft]} text={user.email} />
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -1,12 +1,14 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 import { View, Dimensions, StyleSheet } from 'react-native';
+import { connect } from 'react-redux';
 
-import type { User } from '../types';
+import type { GlobalState, User } from '../types';
 import { UserAvatarWithPresence, ComponentList, RawLabel } from '../common';
+import { getCurrentRealm } from '../selectors';
 import PresenceStatusIndicator from '../common/PresenceStatusIndicator';
 import ActivityText from '../title/ActivityText';
-import { getMediumAvatar } from '../utils/avatar';
+import { getAvatarFromUser } from '../utils/avatar';
 import { nowInTimeZone } from '../utils/date';
 import styles from '../styles';
 
@@ -21,18 +23,19 @@ const componentStyles = StyleSheet.create({
 });
 
 type Props = {|
+  realm: string,
   user: User,
 |};
 
-export default class AccountDetails extends PureComponent<Props, void> {
+class AccountDetails extends PureComponent<Props, void> {
   render() {
-    const { user } = this.props;
+    const { realm, user } = this.props;
     const screenWidth = Dimensions.get('window').width;
 
     return (
       <View>
         <UserAvatarWithPresence
-          avatarUrl={typeof user.avatar_url === 'string' ? getMediumAvatar(user.avatar_url) : null}
+          avatarUrl={getAvatarFromUser(user, realm, screenWidth)}
           email={user.email}
           size={screenWidth}
           shape="square"
@@ -58,3 +61,7 @@ export default class AccountDetails extends PureComponent<Props, void> {
     );
   }
 }
+
+export default connect((state: GlobalState) => ({
+  realm: getCurrentRealm(state),
+}))(AccountDetails);
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 
 import type { GlobalState, User } from '../types';
 import { UserAvatar, ComponentList, RawLabel } from '../common';
-import { getCurrentRealm } from '../selectors';
+import { getCurrentRealm, getUserStatusTextForUser } from '../selectors';
 import PresenceStatusIndicator from '../common/PresenceStatusIndicator';
 import ActivityText from '../title/ActivityText';
 import { getAvatarFromUser } from '../utils/avatar';
@@ -20,6 +20,9 @@ const componentStyles = StyleSheet.create({
     justifyContent: 'center',
     flexDirection: 'row',
   },
+  statusText: {
+    textAlign: 'center',
+  },
 });
 
 const AVATAR_SIZE = 200;
@@ -27,11 +30,12 @@ const AVATAR_SIZE = 200;
 type Props = {|
   realm: string,
   user: User,
+  userStatusText: string | void,
 |};
 
 class AccountDetails extends PureComponent<Props, void> {
   render() {
-    const { realm, user } = this.props;
+    const { realm, user, userStatusText } = this.props;
 
     return (
       <View>
@@ -47,6 +51,12 @@ class AccountDetails extends PureComponent<Props, void> {
             <PresenceStatusIndicator email={user.email} hideIfOffline={false} />
             <RawLabel style={[styles.largerText, styles.halfMarginLeft]} text={user.email} />
           </View>
+          {userStatusText !== undefined && (
+            <RawLabel
+              style={[styles.largerText, componentStyles.statusText]}
+              text={userStatusText}
+            />
+          )}
           <View>
             <ActivityText style={styles.largerText} email={user.email} />
           </View>
@@ -64,6 +74,7 @@ class AccountDetails extends PureComponent<Props, void> {
   }
 }
 
-export default connect((state: GlobalState) => ({
+export default connect((state: GlobalState, props) => ({
   realm: getCurrentRealm(state),
+  userStatusText: getUserStatusTextForUser(state, props.user.user_id),
 }))(AccountDetails);
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -4,7 +4,7 @@ import { View, Dimensions, StyleSheet } from 'react-native';
 import { connect } from 'react-redux';
 
 import type { GlobalState, User } from '../types';
-import { UserAvatarWithPresence, ComponentList, RawLabel } from '../common';
+import { UserAvatar, ComponentList, RawLabel } from '../common';
 import { getCurrentRealm } from '../selectors';
 import PresenceStatusIndicator from '../common/PresenceStatusIndicator';
 import ActivityText from '../title/ActivityText';
@@ -34,9 +34,8 @@ class AccountDetails extends PureComponent<Props, void> {
 
     return (
       <View>
-        <UserAvatarWithPresence
+        <UserAvatar
           avatarUrl={getAvatarFromUser(user, realm, screenWidth)}
-          email={user.email}
           size={screenWidth}
           shape="square"
         />
--- a/src/account-info/ProfileCard.js
+++ b/src/account-info/ProfileCard.js
@@ -3,8 +3,10 @@ import React, { PureComponent } from 'react';
 import { ScrollView, StyleSheet, View } from 'react-native';
 import { connect } from 'react-redux';
 
-import type { GlobalState, User } from '../types';
+import type { Dispatch, GlobalState, User } from '../types';
 import { getSelfUserDetail } from '../selectors';
+import { ZulipButton } from '../common';
+import { navigateToUserStatus } from '../actions';
 import AccountDetails from './AccountDetails';
 import AwayStatusSwitch from './AwayStatusSwitch';
 import SwitchAccountButton from './SwitchAccountButton';
@@ -16,9 +18,13 @@ const componentStyles = StyleSheet.create({
     flexDirection: 'row',
     marginHorizontal: 8,
   },
+  setStatusButton: {
+    marginHorizontal: 16,
+  },
 });
 
 type Props = {|
+  dispatch: Dispatch,
   selfUserDetail: User,
 |};
 
@@ -29,6 +35,11 @@ type Props = {|
  * The user can still open `AccountDetails` on themselves via the (i) icon in a chat screen.
  */
 class ProfileCard extends PureComponent<Props> {
+  handleSetUserStatus = () => {
+    const { dispatch } = this.props;
+    dispatch(navigateToUserStatus());
+  };
+
   render() {
     const { selfUserDetail } = this.props;
 
@@ -36,6 +47,12 @@ class ProfileCard extends PureComponent<Props> {
       <ScrollView>
         <AccountDetails user={selfUserDetail} />
         <AwayStatusSwitch />
+        <ZulipButton
+          style={componentStyles.setStatusButton}
+          secondary
+          text="Set a status"
+          onPress={this.handleSetUserStatus}
+        />
         <View style={componentStyles.accountButtons}>
           <SwitchAccountButton />
           <LogoutButton />
--- a/src/account-info/ProfileCard.js
+++ b/src/account-info/ProfileCard.js
@@ -1,0 +1,50 @@
+/* @flow strict-local */
+import React, { PureComponent } from 'react';
+import { ScrollView, StyleSheet, View } from 'react-native';
+import { connect } from 'react-redux';
+
+import type { GlobalState, User } from '../types';
+import { getSelfUserDetail } from '../selectors';
+import AccountDetails from './AccountDetails';
+import AwayStatusSwitch from './AwayStatusSwitch';
+import SwitchAccountButton from './SwitchAccountButton';
+import LogoutButton from './LogoutButton';
+
+const componentStyles = StyleSheet.create({
+  accountButtons: {
+    alignItems: 'flex-end',
+    flexDirection: 'row',
+    marginHorizontal: 8,
+  },
+});
+
+type Props = {|
+  selfUserDetail: User,
+|};
+
+/**
+ * This is similar to `AccountDetails` but used to show the current users account.
+ * It does not have a 'Send private message` but has `Switch` and `Logout` buttons.
+ *
+ * The user can still open `AccountDetails` on themselves via the (i) icon in a chat screen.
+ */
+class ProfileCard extends PureComponent<Props> {
+  render() {
+    const { selfUserDetail } = this.props;
+
+    return (
+      <ScrollView>
+        <AccountDetails user={selfUserDetail} />
+        <AwayStatusSwitch />
+        <View style={componentStyles.accountButtons}>
+          <SwitchAccountButton />
+          <LogoutButton />
+        </View>
+      </ScrollView>
+    );
+  }
+}
+
+export default connect((state: GlobalState) => ({
+  selfUserDetail: getSelfUserDetail(state),
+}))(ProfileCard);
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -148,7 +148,7 @@ export type Message = $ReadOnly<{
   flags?: $ReadOnlyArray<string>,
 
   /** The rest are believed to really appear in `message` events. */
-  avatar_url: ?string,
+  avatar_url: string | null,
   client: string,
   content: string,
   content_type: 'text/html' | 'text/markdown',
@@ -247,7 +247,7 @@ export type Topic = {|
  *   https://zulipchat.com/api/get-all-users#response
  */
 export type User = {|
-  avatar_url: ?string,
+  avatar_url: string | null,
   bot_type?: number,
   bot_owner?: string,
   email: string,
--- a/src/common/ComponentList.js
+++ b/src/common/ComponentList.js
@@ -32,18 +32,13 @@ export default class ComponentList extends PureComponent<Props> {
     const { children, itemStyle, spacing, outerSpacing, style } = this.props;
     const outerStyle = outerSpacing ? { margin: spacing } : {};
     const marginStyle = { marginBottom: spacing };
-
-    return (
-      <View style={[style, outerStyle]}>
-        {React.Children.map(
-          children,
-          (child, i) =>
-            child
-            && React.cloneElement(child, {
-              style: [child.props.style, i + 1 < children.length ? marginStyle : {}, itemStyle],
-            }),
-        )}
-      </View>
+    const childrenToRender = React.Children.toArray(children).filter(child => child);
+    const childrenWithStyles = childrenToRender.map((child, i) =>
+      React.cloneElement(child, {
+        style: [child.props.style, i + 1 < childrenToRender.length ? marginStyle : {}, itemStyle],
+      }),
     );
+
+    return <View style={[style, outerStyle]}>{childrenWithStyles}</View>;
   }
 }
--- a/src/common/GroupAvatar.js
+++ b/src/common/GroupAvatar.js
@@ -28,16 +28,17 @@ type Props = {
 };
 
 /**
- * Renders a text avatar based on user initials and a color
- * calculated as a hash from their name.
+ * Renders a text avatar based on a single or multiple user's
+ * initials and a color calculated as a hash from their name.
+ * We are currently using it only for group chats.
  *
- * @prop name - User's full name to extract initials from.
+ * @prop name - User's names to extract initials from.
  * @prop size - Sets width and height in pixels.
  * @prop shape - One of 'square', 'rounded', 'circle'.
  * @prop children - If provided, will render inside the component body.
  * @prop onPress - Event fired on pressing the component.
  */
-export default class TextAvatar extends PureComponent<Props> {
+export default class GroupAvatar extends PureComponent<Props> {
   static defaultProps = {
     shape: 'rounded',
   };
--- a/src/common/GroupAvatar.js
+++ b/src/common/GroupAvatar.js
@@ -4,10 +4,8 @@ import { Text, StyleSheet, View } from 'react-native';
 
 import type { ChildrenArray } from '../types';
 import Touchable from './Touchable';
-import { colorHashFromName } from '../utils/color';
-
-export const initialsFromName = (name: string) =>
-  (name.match(/\S+\s*/g) || []).map(x => x[0].toUpperCase()).join('');
+import { colorHashFromString } from '../utils/color';
+import { initialsFromString } from '../utils/misc';
 
 const styles = StyleSheet.create({
   frame: {
@@ -20,7 +18,7 @@ const styles = StyleSheet.create({
 });
 
 type Props = {
-  name: string,
+  names: string,
   size: number,
   shape: string,
   children?: ChildrenArray<*>,
@@ -32,7 +30,7 @@ type Props = {
  * initials and a color calculated as a hash from their name.
  * We are currently using it only for group chats.
  *
- * @prop name - User's names to extract initials from.
+ * @prop names - The name of one or more users, used to extract their initials.
  * @prop size - Sets width and height in pixels.
  * @prop shape - One of 'square', 'rounded', 'circle'.
  * @prop children - If provided, will render inside the component body.
@@ -44,14 +42,14 @@ export default class GroupAvatar extends PureComponent<Props> {
   };
 
   render() {
-    const { children, name, size, shape, onPress } = this.props;
+    const { children, names, size, shape, onPress } = this.props;
 
     const frameSize = {
       height: size,
       width: size,
       borderRadius:
         shape === 'rounded' ? size / 8 : shape === 'circle' ? size / 2 : shape === 'square' ? 0 : 0,
-      backgroundColor: colorHashFromName(name),
+      backgroundColor: colorHashFromString(names),
     };
     const textSize = {
       fontSize: size / 3,
@@ -60,7 +58,7 @@ export default class GroupAvatar extends PureComponent<Props> {
     return (
       <Touchable onPress={onPress}>
         <View style={[styles.frame, frameSize]}>
-          <Text style={[styles.text, textSize]}>{initialsFromName(name)}</Text>
+          <Text style={[styles.text, textSize]}>{initialsFromString(names)}</Text>
           {children}
         </View>
       </Touchable>
--- a/src/common/OwnAvatar.js
+++ b/src/common/OwnAvatar.js
@@ -5,8 +5,7 @@ import { connect } from 'react-redux';
 import type { GlobalState, User } from '../types';
 import { getCurrentRealm, getSelfUserDetail } from '../selectors';
 import UserAvatar from './UserAvatar';
-import { getFullUrl } from '../utils/url';
-import { getGravatarFromEmail } from '../utils/avatar';
+import { getAvatarFromUser } from '../utils/avatar';
 
 type Props = {|
   user: User,
@@ -22,11 +21,7 @@ type Props = {|
 class OwnAvatar extends PureComponent<Props> {
   render() {
     const { user, size, realm } = this.props;
-    const fullAvatarUrl =
-      typeof user.avatar_url === 'string'
-        ? getFullUrl(user.avatar_url, realm)
-        : getGravatarFromEmail(user.email);
-
+    const fullAvatarUrl = getAvatarFromUser(user, realm);
     return <UserAvatar avatarUrl={fullAvatarUrl} size={size} shape="circle" />;
   }
 }
--- a/src/common/UserAvatarWithPresence.js
+++ b/src/common/UserAvatarWithPresence.js
@@ -20,7 +20,7 @@ const componentStyles = StyleSheet.create({
 });
 
 type Props = {|
-  avatarUrl: ?string,
+  avatarUrl: string | null,
   email: string,
   size: number,
   realm: string,
--- a/src/common/UserAvatarWithPresence.js
+++ b/src/common/UserAvatarWithPresence.js
@@ -7,8 +7,7 @@ import { StyleSheet } from 'react-native';
 import type { GlobalState } from '../types';
 import { getCurrentRealm } from '../selectors';
 import UserAvatar from './UserAvatar';
-import { getFullUrl } from '../utils/url';
-import { getGravatarFromEmail } from '../utils/avatar';
+import { getAvatarUrl } from '../utils/avatar';
 import PresenceStatusIndicator from './PresenceStatusIndicator';
 
 const componentStyles = StyleSheet.create({
@@ -49,8 +48,7 @@ class UserAvatarWithPresence extends PureComponent<Props> {
 
   render() {
     const { avatarUrl, email, size, onPress, realm, shape } = this.props;
-    const fullAvatarUrl =
-      typeof avatarUrl === 'string' ? getFullUrl(avatarUrl, realm) : getGravatarFromEmail(email);
+    const fullAvatarUrl = getAvatarUrl(avatarUrl, email, realm);
 
     return (
       <UserAvatar avatarUrl={fullAvatarUrl} size={size} onPress={onPress} shape={shape}>
--- a/src/common/index.js
+++ b/src/common/index.js
@@ -32,7 +32,7 @@ export { default as SectionSeparatorBetween } from './SectionSeparatorBetween';
 export { default as SlideAnimationView } from './SlideAnimationView';
 export { default as SmartUrlInput } from './SmartUrlInput';
 export { default as SpinningProgress } from './SpinningProgress';
-export { default as TextAvatar } from './TextAvatar';
+export { default as GroupAvatar } from './GroupAvatar';
 export { default as Touchable } from './Touchable';
 export { default as UnreadCount } from './UnreadCount';
 export { default as PresenceStatusIndicator } from './PresenceStatusIndicator';
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -113,7 +113,7 @@ class Lightbox extends PureComponent<Props, State> {
           <LightboxHeader
             onPressBack={this.handlePressBack}
             timestamp={message.timestamp}
-            avatarUrl={getAvatarFromMessage(message)}
+            avatarUrl={getAvatarFromMessage(message, auth.realm)}
             senderName={message.sender_full_name}
           />
         </SlideAnimationView>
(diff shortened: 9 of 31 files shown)


borisyankov:
Ready.

gnprice:
Thanks @borisyankov !

> Add `OwnAvatar` component

This comment still applies: https://github.com/zulip/zulip-mobile/pull/3267#discussion_r255269225

> Add `AwayStatusSwitch` component

```
+type Props = {|
+  ...PropsFromState,
+  dispatch: Dispatch,
+|};
```

* On another PR we discussed this name `PropsFromState`, which should be something else.

* I'm also curious what the intended use and meaning of the type is exactly. The `dispatch` prop is also provided by `connect`... so if this gets called something like `PropsFromConnect`, then it should probably go there. Also if the idea is that the props mentioned explicitly in `Props` will correspond to props the caller can, or should, pass in, then `dispatch` doesn't belong in that list.

* Separately: this in the commit message is good information that I think would be best captured in jsdoc on the component:
```
     * retrieves the current user's `user status` data and presents it
     * allows by switching it to control the `away` status
```

> Add `ProfileCard` component

Similarly, this would be good to capture as jsdoc on the component:
```
    This is very similar to `AccountDetails` but used to show the
    current users account.
```

> Rework `AccountDetails` to be more generic
> 
> Move "Send private message" out of `AccountDetails`.
> This allows us to reuse `AccountDetails` in `ProfileCard` but not
> show the "Send private message" button that is not appropriate there.

* Hmm interesting. Shouldn't this go before we create `ProfileCard`, then? Or at least before we wire it up in the navigation.

* Also this looks in the code like it will change the layout -- the button was previously inside a `ComponentList`. Haven't yet compared visually.


> AccountDetails: show user status text if one is set

```
+    // backgtoundColor: 'red',
```

* You'll want to take this out, presumably.

  (Also it's misspelled.)

```
+    const statusText = userStatus[user.user_id] && userStatus[user.user_id].status_text;
[...]
+          {statusText !== undefined && (
+            <RawLabel style={[styles.largerText, componentStyles.statusText]} text={statusText} />
+          )}
```

* This logic feels to me like it's calling for a (very simple) selector. Best if the component doesn't have to deal with the nuances of what it means for the whole object to be missing from the map, vs. the object to be present but have `undefined` for `status_text`.

  E.g. `getUserStatusText: (state: GlobalState, userId: number) => string | void`, making use of the kind of signature discussed in #3015.

> Add `UserStatusScreen`

```
    Values are imported from a separate file as Prettier does not
    'like' unicode and messes up the formatting.
```

Huh indeed!

Less euphemistically: Prettier has a bug in handling Unicode. :stuck_out_tongue:

I suspect that it's specifically about non-BMP code points, which includes most emoji -- those are the code points that in JavaScript (and Java) if you're not careful you end up treating as if they're two code points. Like the RN bug that underlay #2787.

Searching their tracker, this looks a lot like prettier/prettier#770. ... But that was supposedly fixed in Prettier 1.3.0, and we're on 1.12.1. Hm.

Anyway:

* At a minimum, there should be a comment inside that tiny file saying why it's there -- it's a peculiar way for the code to be, and so it should provide an answer to the reader's natural question.

* Better, let's be good citizens and do something about the issue :wink: . Looks like there are newer releases of Prettier -- see if the issue goes away on upgrade? If not, file a bug with Prettier upstream.

```
    Possible presets are given for a one-click entry.
```

Good idea!

```
+  getInitialStatusText = () => {
+    const { selfUserDetail, userStatus } = this.props;
+    const selfUserStatus = userStatus[selfUserDetail.user_id];
+    return (selfUserStatus && selfUserStatus.status_text) || '';
+  };
```

Another use case for that selector :slightly_smiling_face: 

```
+  handleStatusTextUpdate = () => {
+    const { dispatch } = this.props;
+    const { statusText } = this.state;
+    dispatch(updateUserStatusText(statusText));
+    dispatch(navigateBack());
+  };
+
+  handleStatusTextClear = () => {
+    const { dispatch } = this.props;
+    dispatch(updateUserStatusText(''));
+    dispatch(navigateBack());
+  };
```

Hmm -- perhaps the "clear" case should change the text to `''`, and then call the "update" case?

```
+  handleStatusTextChangeText = (statusText: string) => {
+    this.setState({
+      statusText,
+    });
+  };
```

This is a bit confusing -- it's basically named after one call site, the `onChangeText` of the input, but then we use it in another place too where `onChangeText` is definitely *not* what just happened.

Perhaps it'd be clearer to name it after what it does? E.g. `setStatusTextState`. In particular, if it weren't for the unfortunate issue of identity across re-renders, I think we wouldn't have a name for this at all and would just write e.g. `statusText => this.setState({ statusText })`.

... Hmm, in fact it looks like you are already giving up identity across re-render!
```
            onPress={() => {
              this.handleStatusTextChangeText(statusSuggestion);
            }}
```
If you can fix that reasonably, that'd be good.

If you can't... might as well at least make it clearer to read, by writing the `setState` directly.

```
+export default [
+  '📅 In a meeting',
+  '🚌 Commuting',
+  '🤒 Out sick',
+  '🌴 Vacationing',
+  '🏠 Working remotely',
+];
```

* These should probably get translated. See LanguagePicker (since #3231) for an example of how to do that cleanly.

  Hmm, more precisely: I think in your current version, `OptionButton` will try to translate them for the *labels*, but the actual value *entered* will be untranslated. (And no translation will happen anyway because they're not in the messages list.) Anyway, we should translate them.

* Separately: some of these feel like they ought to go with an "away" state. But this UI is no more awkward than the webapp seems to be for this, so *shrug*.

> Add `Set a status` button to `ProfileCard`

Backticks mean code :slightly_smiling_face: . You want double-quotes, I think.

Also there's an unclosed backtick in the commit-message body. Small but it sticks out at me.


> Add `getAvatarFromUser` and `getAvatarFromMessage`

Good idea!

* What about in `Avatar`? Very similar logic there too.

```
-            avatarUrl={message.avatar_url || getGravatarFromEmail(message.sender_
email)}
+            avatarUrl={getAvatarFromMessage(message)}
```

* This code differs slightly from old to new: in the old version, an `avatar_url` of `''` would fail over to `getGravatarFromEmail`. Was that case impossible? Or, is the new behavior the right thing in that case anyway? Deserves a few words in the commit message.

* Also, that would have been impossible in the first place if that file had `@flow strict-local`. A good followup would be to go and make it that way now. Looks like there's just one obstacle to that, and it looks like an easy one.

borisyankov:
> * On another PR we discussed this name `PropsFromState`, which should be something else.

That was written at the time when I wrote the other commit. We have agreed on `PropsFromConnect` but also it does not matter for this commit so I reverted it to the usual `Props`.

borisyankov:
> Hmm -- perhaps the "clear" case should change the text to `''`, and then call the "update" case?

We close the screen after the button press, so that step is not needed.

borisyankov:
About this: https://github.com/zulip/zulip-mobile/pull/3267#discussion_r255269225

I did simplify the `OwnAvatar` even further, removing the `shape` as prop.
After refactoring the avatar/gravatar logic in a later commit it is almost as simple as just using `Avatar` (maybe 1 line of code difference). We need the component to get the data out of the user anyway.

Also, currently we don't have a way to not show the `PresenceStatusIndicator` in `Avatar` but we don't want it in `OwnAvatar`. 

These are the reasons that justify this component's existence for me.


As for whether `Avatar` is needed, I am not sure, maybe we can refactor it later indeed.

> Perhaps that `TextAvatar` path in `Avatar` never gets used at all?

It is currently used in `TitleGroup`. It makes sense to make the usage in that file and in `GroupPmConversationItem` the same. There is some (minor) value in abstracting away how an avatar is rendered by just using `Avatar` in both places.

borisyankov:
As for whether `Avatar` is needed, I am not sure, maybe we can refactor it later indeed.

> Perhaps that `TextAvatar` path in `Avatar` never gets used at all?

It is currently used in `TitleGroup`. It makes sense to make the usage in that file and in `GroupPmConversationItem` the same. There is some (minor) value in abstracting away how an avatar is rendered by just using `Avatar` in both places.

borisyankov:
> Searching their tracker, this looks a lot like [prettier/prettier#770](https://github.com/prettier/prettier/issues/770). ... But that was supposedly fixed in Prettier 1.3.0, and we're on 1.12.1. Hm.

I did update to the latest one `1.16.4` and tried it, the same result.

borisyankov:
> * Hmm interesting. Shouldn't this go before we create `ProfileCard`, then? Or at least before we wire it up in the navigation.

I did move it before the creation of `ProfileCard` and moved the changes to this file to its creation.

> * Also this looks in the code like it will change the layout -- the button was previously inside a `ComponentList`. Haven't yet compared visually.

It doesn't change visually.

borisyankov:
I did create several user status selectors (with docs) and did rework the other commits to use them.

borisyankov:
After several smaller changes according to comments and few visual improvements, this should be ready now.

gnprice:
> About this: [#3267 (comment)](https://github.com/zulip/zulip-mobile/pull/3267#discussion_r255269225)
> 
> I did simplify the `OwnAvatar` even further, removing the `shape` as prop.
> After refactoring the avatar/gravatar logic in a later commit it is almost as simple as just using `Avatar` (maybe 1 line of code difference). We need the component to get the data out of the user anyway.
> 
> Also, currently we don't have a way to not show the `PresenceStatusIndicator` in `Avatar` but we don't want it in `OwnAvatar`.
> 
> These are the reasons that justify this component's existence for me.

Cool, thanks for this reply!

One of my concerns in the linked comment was that part of the commit message is mistaken, namely this bit:

> does not render TextAvatar as it is only for group conversations

That part is still there. Regardless of what we do with the code, we need to make the commit message accurate.


> After refactoring the avatar/gravatar logic in a later commit it is almost as simple as just using `Avatar` (maybe 1 line of code difference).

This sounds to me like further reason why we *shouldn't* have this duplication. The "1 line of code difference" describes the extra cost of using the common `Avatar` vs. using `OwnAvatar`. But the cost of *having* `OwnAvatar` is much greater than that: for a start, this file is 34 lines long.

But:

> We need the component to get the data out of the user anyway.

Hmm, given the context in which we invoke this component (from `MainTabs`, rather than from an ordinary component), I agree that it'd be inconvenient to pass this data in from the caller. Shrug, then.

---

That still leaves

* the commit message needs to be accurate
* the jsdoc mentions three optional parameters -- only one of which is a parameter (the others come from `connect`), and none of which are optional.

The first of those two issues, I mentioned in my original review, which I linked to again last time. It's frustrating to have to keep repeating it.

gnprice:
In general:

Thanks @borisyankov for these revisions! I appreciate the factoring out of those selectors (complete with jsdoc! :heart: :writing_hand:), and the moving some details into jsdoc and reordering of some commits.

The first 7 of these commits (up through f5dd46993 "Remove 'Switch` and `Logout` buttons from `Settings`") all look good, with the exception of my comments above about "Add `OwnAvatar` component".

In the remaining 4 commits:
```
0d602b340 AccountDetails: show user status text if one is set
9bbc2ad65 Add `UserStatusScreen`
2a800e6dc Add 'Set a status' button to `ProfileCard`
a44282c7d Add `getAvatarFromUser` and `getAvatarFromMessage`
```
I'm a bit frustrated because each one of them has things I asked for in my previous review [above](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-464245998) that you've neither changed nor replied about. Would you please look back at those?

gnprice:
> > Hmm -- perhaps the "clear" case should change the text to `''`, and then call the "update" case?
> 
> We close the screen after the button press, so that step is not needed.

We do, but the field is still visible for a couple hundred ms during the animation.

I just played with this -- enter some text, then hit "Clear". It is disconcerting as the text stays there unchanged, looking like my request to clear it has been ignored. I then tried making the change I suggested above, and the immediate feedback of seeing the text get cleared is a definite improvement.

It also reduces duplication in the code.

gnprice:
I just tried playing with the UI from this branch.

The new ProfileCard and UserStatusScreen have a problem with vertical overflow -- they can be taller than the screen, and then they don't seem to scroll. Here's `ProfileCard` after I've set a status:
<div align="center"></div>
That's on an emulated Pixel 2.

... And actually now I can't even get to `UserStatusScreen`, because the button has been scrolled completely off the bottom. (Before setting a status, the button is visible but partly cut off.) But it was bad too -- the buttons get covered up by the keyboard, which makes the UI extremely confusing unless you already know (or guess) that they're hiding there and try closing the keyboard.

Should also check what it's like for `AccountDetailsScreen` when the user has set an away status.

gnprice:
> > Perhaps that `TextAvatar` path in `Avatar` never gets used at all?
> 
> It is currently used in `TitleGroup`.

Is it? How's that?

I just looked again at the relevant logic. I'm now pretty confident the truth is that the `TextAvatar` codepath is unreachable.

Here's that logic:
```
    const fullAvatarUrl =
      typeof avatarUrl === 'string' ? getFullUrl(avatarUrl, realm) : getGravatarFromEmail(email);
    const AvatarComponent = fullAvatarUrl ? ImageAvatar : TextAvatar;
```

So the question is if `fullAvatarUrl` can ever be falsy.

Here's `getGravatarFromEmail`:
```
export const getGravatarFromEmail = (email: string = '', size: number = 80): string =>
  `https://secure.gravatar.com/avatar/${md5(email.toLowerCase())}?d=identicon&s=${size}`;
```

Nope, never falsy.

Here's `getFullurl`:
```
export const getFullUrl = (url: string = '', realm: string): string =>
  !url.startsWith('http') ? `${realm}${url.startsWith('/') ? '' : '/'}${url}` : url;
```

Rather a mess in a number of ways, but setting that aside -- this will never be falsy either!
* If `url` is a nonempty string, neither branch will be falsy.
* If `url` is `''`, we get the first branch... which will be a nonempty string. (Even if `realm` is something bogus, we'll get `/`.)
* Even if a `null` sneaks in here, that will just blow up at `startsWith`. And `undefined` would do the same except it gets replaced with `''` before that.

---

Perhaps I'm missing something here! If so I will be glad for you to show it to me.

Even then, though, I'm disappointed that you didn't make the effort to set it out clearly in the previous comment. It's rather a lot more work in general to demonstrate that a given code path *isn't* ever taken than it is to demonstrate that it *is* -- the former has to cover all possibilities, while the latter need only cover one -- but here I am doing that work on the harder side, and you haven't done the courtesy of sharing your thinking on the easier side.

borisyankov:
> Perhaps I'm missing something here! If so I will be glad for you to show it to me.

I was referring to `Avatar` being used there... assuming that we were using the `TextAvatar` for a group icon in the title bar. After your comments and reading through code/history I realized I was misremembering that.

Apparently, originally the intention was not to decide between User / Group for `TextAvatar` but user vs user without an avatar. Currently, indeed, we do not have this case. Not sure if that ever was true, but if so, was a very long time ago.

I have started refactoring `Avatar`, removing the `AvatarComponent`, fixing some of the types, for example https://github.com/zulip/zulip-mobile/commit/ef209dbb8c9399a9250ea51f496faefa9905aed6#diff-253b256c35ad836307b741a080fd2b1a fixed several issues but made the `=== 'string'` check incorrect in `Avatar`.

I am also making it take a `user` prop instead of the other params.

borisyankov:
Another refactor I think makes sense after that is to rename:
`Avatar` => `AvatarWithPresence`
'ImageAvatar` => `UserAvatar`
`TextAvatar` => `GroupAvatar`

gnprice:
> Apparently, originally the intention was not to decide between User / Group for `TextAvatar` but user vs user without an avatar. Currently, indeed, we do not have this case. Not sure if that ever was true, but if so, was a very long time ago.

Yeah, that story makes sense.


> I have started refactoring `Avatar`, removing the `AvatarComponent`, fixing some of the types [...] I am also making it take a `user` prop instead of the other params.
> 
> Another refactor I think makes sense after that is to rename:
> 
> * `Avatar` => `AvatarWithPresence`
> * `ImageAvatar` => `UserAvatar`
> * `TextAvatar` => `GroupAvatar`

These all sound good!


> I think we should merge this independently and should not delay this PR.

Sure. The things currently standing between this PR and a merge are
* the items after the line in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465407576), on `OwnAvatar`
* the still-outstanding items from the previous review, referred to in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465412909)
* the small UI issue in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465414341) (could merge with this, but I'd definitely want it addressed in followup)
* the major UI issues in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465416092)

borisyankov:
A summary of what id the 'diff' with the latest reviewed version:

> * the items after the line in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465407576), on `OwnAvatar`

Major rework of how the `getGravatar*` functions work. 
Fixes inconsistencies, potential bugs and size issues.

> * the still-outstanding items from the previous review, referred to in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465412909)

For `OwnAvatar` i did fix the commit message, the jsdoc was fixed too

> * the small UI issue in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465414341) (could merge with this, but I'd definitely want it addressed in followup)

I did fix it so the 'Clear' action does show in the UI. Could have added even a bit of delay, since I had to use `setTimeout` anyway, but not sure it was a good idea, so I didn't.

> * the major UI issues in [this comment](https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-465416092)

Major rework of the component. Now everything fits in small screens too.

borisyankov:
I am not sure if I am missing anything as the comments started hiding and links stopped working...

gnprice:
Thanks @borisyankov for these revisions!

* Re comments and links to them: as we discussed in audio chat today, what I'd like to be standard procedure -- and this is quite important to me -- is that when resubmitting a branch for review, you go back through **every review comment, item by item**, and make sure you've addressed each one. (Meaning: either done as suggested, or mentioned that you aren't doing it and why.)

  When you do that, it won't matter if some links to comments don't work -- because you're going through every comment anyway.

  This might require explicitly expanding some comments in the UI, because GitHub is being GitHub. A bit annoying of a UI, but it works fine.

---

* I'm going to merge these commits in a bit:
```
bccde1d37 userStatusSelectors: Create three user-status selectors
88d7ff6f2 Add `OwnAvatar` component
304fab86b Add `AwayStatusSwitch` component
d79bf1bbd Rework `AccountDetails` to be more generic
e2cc51ebf Avatar: Do not use `TextAvatar` and simplify
8d4941e65 Rename `ImageAvatar` to `UserAvatar`
4cc9bed38 Rename `Avatar` to `UserAvatarWithPresence`
```

* I quite like the direction of the changes to the avatar code! Specific comments below on the commits I'm not yet merging, but in general I think this makes very good progress to making that code more understandable and robust.

* I'm still not fully satisfied with the various avatar components, but this is now good enough that I'm willing to merge `OwnAvatar`. I know you have some further refactors coming on that, which I'm looking forward to; and there are some specific things I'm interested in, which I can describe later.

* Small bit of commit-message style, in "e2cc51ebf Avatar: Do not use `TextAvatar` and simplify":

  For mentioning a commit, instead of a long link like `https://github.com/zulip/zulip-mobile/commit/ef209dbb8c9399a9250ea51f496faefa9905aed6`, let's write just the commit ID (the "sha1"). And just a short form -- 7 hex digits is Git's default, and I use 9, which will last even for a giant repo.

  This makes it much more readable in prose. It also makes it much more convenient for use at the command line, and transparent for other tools like `gitk` -- every tool gets to linkify it internally in its own UI, rather than making you go to a browser and wait for GitHub to load.

---

* "b567f28be Add `getAvatarFromUser` and `getAvatarFromMessage`"

  This is one with some unaddressed earlier comments.

* "4c6b2601e Rename `TextAvatar` to `GroupAvatar`"

  This looks good! Except that a couple of these lines should raise mental question marks :stuck_out_tongue: -- most of all this one:

```
+ * @prop name - User's names to extract initials from.
```

  What is this `name` prop anyway? It's a name, but also several names?

  Whatever the real answer is, the text and likely the name of the prop should be adjusted to all agree on it.

* "82dcdc3cd Add `getAvatarFromParams` function"

  * Huh, "from params" is kind of a funny thing to say in a function's name. :slightly_smiling_face: What is that intended to mean in this case? Naturally it is using the parameters it's passed... what is this name intended to say that isn't contained in a name like `getAvatar`?

  Perhaps you were aiming to distinguish it from a selector? I think the "FromFoo" doesn't totally accomplish that anyway; but `computeAvatar` would be one way to do it. Also "...AvatarUrl" would be more accurate.

  * `avatarUrl: ?string` -- this type is too vague, permitting `null` when that's not what we mean. Let's say `string | void` instead.

  * This combines a nontrivial refactor with a change in behavior for `getAvatarFromMessage`. Both changes are good, but combining them in one commit makes them both confusing for the reader to unpack -- better to split them into two commits.

* "2dd1179e5 Make all avatar-genrating functions take a 'size' parameter"

  This makes a lot of sense, thanks! As a small bit of style: I'm not a fan of duplicating the choice of default value (80) to four different places. Most of these don't actually care at all what the default is, only that it's optional -- how about just writing `size?: number` there? Then when it's absent, the single default at the very inside will take effect.

* "947637ccb Use `getMediumAvatar` if larger avatar needed"

  Makes sense. Let's move that explanation (especially about the two sizes `default` and `medium` and where the number 100 comes from) into a comment inside the function -- it's something that's important for understanding the new code as it is, not only the diff between old and new.

  Also in general: no need to duplicate information between a commit message and a comment. When you move information like that into a comment, go ahead and take out the details in the commit message that just repeat it.

* "ee4f669bb Redesign `AccountDetails` to fit screens better"

  Cool, I like this UI change!

  The commit is a bit confusing because it silently makes a mostly-unrelated (though also good) change: it cuts where we'd been showing the presence indicator overlaid on the avatar.

  That is certainly a good change! It's basically one of the payoffs of the previous refactors to the avatar components. Best to make it a separate commit, so its commit message can properly celebrate. :smiley:

  (If we were determined to keep it one commit, it would in any case absolutely need to be mentioned in the commit message. But by far the clearest way to do that is to make it its own commit.)

* The next three commits look good (as they did before); leaving unmerged because they depend on others above.

* The next three each have unaddressed comments from before:
```
86944d053 AccountDetails: show user status text if one is set
bc55720e9 Add `UserStatusScreen`
dc90c8ff1 Add 'Set a status' button to `ProfileCard`
```

  So please read through my comments above and take care of those.

* More on "bc55720e9 Add `UserStatusScreen`":

  Thanks for the change to the "clear" case here!

  I'm curious about the `setTimeout`. Do you believe it should be needed in principle; or did you find that in a certain environment it was empirically needed? When I tried this change the other day, it WFM without that; but I'm not sure whether it should be needed in principle.

gnprice:
Pushed those 7 commits as 70d406709..f37856207. Thanks again!

gnprice:
(Quick thing I neglected to mention on Friday as I was rushing out the door:

> in general: no need to duplicate information between a commit message and a comment. When you move information like that into a comment, go ahead and take out the details in the commit message that just repeat it.

I went ahead and did this on a few of the commits I merged, before merging them. You had kindly put information there into comments in response to a previous round of review (thanks!), and I hadn't yet mentioned this finer point. As you revise the remaining commits, please make this change for them too.)

borisyankov:
I did push an update. Changes from the last version:

Added a commit 'ComponentList: Fix a layout issue adding extra space at bottom'

It fixes a bug that adds an extra space at the bottom of the list of components if the last component is `null`. While that seems it would occur rarely, if ever, this happens in `AccountDetails` if the `timezone` prop is not set (we render a `null`).

> - Small bit of commit-message style, in "e2cc51ebf Avatar: Do not use `TextAvatar` and simplify":
>   For mentioning a commit, instead of a long link like `https://github.com/zulip/zulip-mobile/commit/ef209dbb8c9399a9250ea51f496faefa9905aed6`, let's write just the commit ID (the "sha1"). And just a short form -- 7 hex digits is Git's default, and I use 9, which will last even for a giant repo.

Makes sense. I will use 7 digits from now on.

> - "b567f28be Add `getAvatarFromUser` and `getAvatarFromMessage`"
>   This is one with some unaddressed earlier comments.

I am not sure what this is referring to. If it is the comments about the `Avatar` component that is now called `UserAvatarWithPresence` - the comments are referring to an old implementation. Currently we reuse the same code used everywhere else via `getAvatarFromParams`.

> - "4c6b2601e Rename `TextAvatar` to `GroupAvatar`"
>   This looks good! Except that a couple of these lines should raise mental question marks 😛 -- most of all this one:
>
> ```
> + * @prop name - User's names to extract initials from.
> ```
>
> What is this `name` prop anyway? It's a name, but also several names?
>
> Whatever the real answer is, the text and likely the name of the prop should be adjusted to all agree on it.

I did several changes that resulted in a new commit 'GroupAvatar: rename prop `name`, clarify usage'

> - "82dcdc3cd Add `getAvatarFromParams` function"
>
> - Huh, "from params" is kind of a funny thing to say in a function's name. 🙂 What is that intended to mean in this case? Naturally it is using the parameters it's passed... what is this name intended to say that isn't contained in a name like `getAvatar`?
>
> Perhaps you were aiming to distinguish it from a selector? I think the "FromFoo" doesn't totally accomplish that anyway; but `computeAvatar` would be one way to do it. Also "...AvatarUrl" would be more accurate.

The weird naming was selected to correspond to the `getAvatarFromUser` and `getAvatarFromMessage` functions.
If it wasn't that clear it is the similar (the same signature) the drawback of that name is not justified.
Now I renamed it to `getAvatarUrl`.

> - `avatarUrl: ?string` -- this type is too vague, permitting `null` when that's not what we mean. Let's say `string | void` instead.

I left this as-is as there are 7-8 other places that refer to that same type.
I added another commit called 'flow: Specify `avatar_url` more precisely' that addresses this.

> - This combines a nontrivial refactor with a change in behavior for `getAvatarFromMessage`. Both changes are good, but combining them in one commit makes them both confusing for the reader to unpack -- better to split them into two commits.

I did extract the change that made `getAvatarFromMessage` support a `realm` parameter into a separete commit:
'getAvatarFromMessage: Take into account current realm'
The refactor itself (and only it) remains in the previous commit.

> - "2dd1179e5 Make all avatar-genrating functions take a 'size' parameter"
>   This makes a lot of sense, thanks! As a small bit of style: I'm not a fan of duplicating the choice of default value (80) to four different places. Most of these don't actually care at all what the default is, only that it's optional -- how about just writing `size?: number` there? Then when it's absent, the single default at the very inside will take effect.

Good find. I did remove the defaults. Added details in the commit message.
Also added a change to `getGravatarFromEmail` in another commit right after it:
'getGravatarFromEmail: Remove default value for `size`'

> - "947637ccb Use `getMediumAvatar` if larger avatar needed"
>   Makes sense. Let's move that explanation (especially about the two sizes `default` and `medium` and where the number 100 comes from) into a comment inside the function -- it's something that's important for understanding the new code as it is, not only the diff between old and new.
>   Also in general: no need to duplicate information between a commit message and a comment. When you move information like that into a comment, go ahead and take out the details in the commit message that just repeat it.

I did add a JSDoc with that information to `getMediumAvatar` and removed the duplications from the commit message.

> - "ee4f669bb Redesign `AccountDetails` to fit screens better"
>   Cool, I like this UI change!
>   The commit is a bit confusing because it silently makes a mostly-unrelated (though also good) change: it cuts where we'd been showing the presence indicator overlaid on the avatar.
>   That is certainly a good change! It's basically one of the payoffs of the previous refactors to the avatar components. Best to make it a separate commit, so its commit message can properly celebrate. 😃
>   (If we were determined to keep it one commit, it would in any case absolutely need to be mentioned in the commit message. But by far the clearest way to do that is to make it its own commit.)

I did split this into two commits.
The new one is called:
'AccountDetails: Use `getAvatarFromUser` and `realm` within'

> - This logic feels to me like it's calling for a (very simple) selector. Best if the component doesn't have to deal with the nuances of what it means for the whole object to be missing from the map, vs. the object to be present but have `undefined` for `status_text`.

I did rework the commit and extracted the logic as a selector.

> - At a minimum, there should be a comment inside that tiny file saying why it's there -- it's a peculiar way for the code to be, and so it should provide an answer to the reader's natural question.

I did add comment in `userStatusTextSuggestions.js`

> Backticks mean code 🙂 . You want double-quotes, I think.
> Also there's an unclosed backtick in the commit-message body. Small but it sticks out at me.

Fixed.

> - More on "bc55720e9 Add `UserStatusScreen`":
>   Thanks for the change to the "clear" case here!
>   I'm curious about the `setTimeout`. Do you believe it should be needed in principle; or did you find that in a certain environment it was empirically needed? When I tried this change the other day, it WFM without that; but I'm not sure whether it should be needed in principle.)

I tested this and apparently I was wrong we needed it for the state to be reflected in the UI.
What is funny, we still needed it to have the state be mutated before the `updateStatusText` call.
The state mutations are batched so that could have been expected.
I wasn't sure `setTimeout` guarantees us the state would be mutated, though I think it does.
I also experimented with `setState` and a second parameter being a call to `updateStatusText` - that is now guaranteed to be after the mutation.

Instead of keeping this, I reworked the code a bit to be more straight-forward and not require that deep knowledge of how state mutations work.

gnprice:
Thanks @borisyankov for all these revisions! I especially appreciate that you went through previous comments and addressed most of them.

This is quite close -- if it were earlier in the day I'd be going and merging the bulk of it.

* One key thing that's still open is the UserStatusScreen portion of this UI issue:

  > The new ProfileCard and UserStatusScreen have a problem with vertical overflow -- they can be taller than the screen, and then they don't seem to scroll. [... in UserStatusScreen] the buttons get covered up by the keyboard, which makes the UI extremely confusing unless you already know (or guess) that they're hiding there and try closing the keyboard.

---

Replying to one question of yours above:

> I am not sure what this is referring to. If it is the comments about the `Avatar` component that is now called `UserAvatarWithPresence` [...]

* I was thinking of this comment, which is about code in `Lightbox` (quoted above the comment):

  > This code differs slightly from old to new: in the old version, an `avatar_url` of `''` would fail over to `getGravatarFromEmail`. Was that case impossible? Or, is the new behavior the right thing in that case anyway? Deserves a few words in the commit message.

* Also this comment which suggests a followup:

  > Also, that would have been impossible in the first place if that file had `@flow strict-local`. A good followup would be to go and make it that way now. Looks like there's just one obstacle to that, and it looks like an easy one.

---

Other items that potentially we can leave for followups, but would be good to take care of:

* In the pre-supplied values for `UserStatusScreen`:

  > These should probably get translated. See LanguagePicker (since #3231) for an example of how to do that cleanly.
  > Hmm, more precisely: I think in your current version, `OptionButton` will try to translate them for the _labels_, but the actual value _entered_ will be untranslated. (And no translation will happen anyway because they're not in the messages list.) Anyway, we should translate them.

* In "Redesign `AccountDetails` to fit screens better": it'd be good to write "200" just once, as it's an arbitrary size that it's important to keep matching between the two places we mention it. Just a local variable like `const avatarSize = 200` would do nicely, I think.

* About the Prettier issue in `UserStatusScreen`:

  > Better, let's be good citizens and do something about the issue :wink: . Looks like there are newer releases of Prettier -- see if the issue goes away on upgrade? If not, file a bug with Prettier upstream.

gnprice:
It's the end of my day now, but when I get to my desk tomorrow I plan to go about merging this.

If by that point you've resolved the "covered up by the keyboard" issue, then I expect to merge the whole thing.

If you can also take care of the other items in my previous comment, that would be great. If not, that will just mean a few more followup items. Thanks again!

borisyankov:
One key thing that's still open is the UserStatusScreen portion of this UI issue:

> The new ProfileCard and UserStatusScreen have a problem with vertical overflow -- they can be taller than the screen, and then they don't seem to scroll. [... in UserStatusScreen] the buttons get covered up by the keyboard, which makes the UI extremely confusing unless you already know (or guess) that they're hiding there and try closing the keyboard.

I did rework the `UserStatusScreen` - now it uses a `FlatList` that only takes as much of the screen as there is space. Also, the buttons are now always on screen. That seems to work quite well on smaller screens.

For `ProfileCard` I replaced the `View` with a `ScrollView` and that is enough to fix the now rare case that we need to scroll this screen (after making sure in previous changes to not add extra space)

> I was thinking of this comment, which is about code in Lightbox (quoted above the comment):
>
> This code differs slightly from old to new: in the old version, an avatar_url of '' would fail over to getGravatarFromEmail. Was that case impossible? Or, is the new behavior the right thing in that case anyway? Deserves a few words in the commit message.

Supposedly impossible. We either have an actual avatar url or `null`.

> Also, that would have been impossible in the first place if that file had @flow strict-local. A good followup would be to go and make it that way now. Looks like there's just one obstacle to that, and it looks like an easy one.

I think we have `@flow strict-local` now or we are talking about different files?

> Other items that potentially we can leave for followups, but would be good to take care of:

> In the pre-supplied values for UserStatusScreen:

> These should probably get translated. See LanguagePicker (since #3231) for an example of how to do that cleanly.
> Hmm, more precisely: I think in your current version, OptionButton will try to translate them for the labels, but the actual value entered will be untranslated. (And no translation will happen anyway because they're not in the messages list.) Anyway, we should translate them.

We should ship as-is but will do this in subsequent commit. 

> In "Redesign AccountDetails to fit screens better": it'd be good to write "200" just once, as it's an arbitrary size that it's important to keep matching between the two places we mention it. Just a local variable like const avatarSize = 200 would do nicely, I think.

Done.

> About the Prettier issue in UserStatusScreen:
>
> Better, let's be good citizens and do something about the issue 😉 . Looks like there are newer releases of Prettier -- see if the issue goes away on upgrade? If not, file a bug with Prettier upstream.

Tested with the new version and it wasn't fixed.
But I did try many variations in order to reproduce a minimal verrsion for a bug report.
A breakthrough! It seems the default eslint/prettier settings do not cause the issue.
I'll continue testing to see what is the exact thing that causes it.

gnprice:
Thanks again @borisyankov for these revisions!

On UserStatusScreen, I'm still seeing the buttons obscured by the keyboard, and no scroll. This is 
with commit b32103e34 (the one you pushed a few minutes ago), with a `FlatList` there. Screenshot:
image

gnprice:
Huh, the plot thickens! It's not 100% of the time -- here's another screenshot, where the buttons are nicely visible:
image
And the suggestions list scrolls.

That's what I saw immediately after switching apps back to Zulip, from the system settings where I was browsing around for some details.

I'll poke a bit more. Perhaps you also have ideas?

gnprice:
Hmm, now it *mostly* doesn't reproduce, in either of two emulators. I'm not sure what the pattern is.

Some observations and tentative patterns:
* I observed it several times after reloading in the dev popup menu, and after re-running `react-native run-android` to reinstall a current build of the APK. So that rules out some ways it could have been just stale code from a different version of the branch.
  * But, at some point the Metro bundler (`react-native start`) crashed and I restarted it... and I think I haven't seen it since then. So possibly the bundler was serving stale code? ... alternately with serving good code? Grr, if so.
* I think I only ever saw it when navigating to `UserStatusScreen` was the very first time the keyboard showed up since starting the app. That could be a clue.
* The keyboard often behaved funny, as it sometimes does in debug builds in the emulator.
  * E.g., I'd hit `r r` to reload... and that would bring up the keyboard, and it'd stay up when the app loaded and showed the main nav screen, even though the keyboard makes no sense there.
  * So it's possible whatever bug this was only applies in the emulator and/or in a debug build.

gnprice:
Cool, I'm going to go ahead and merge -- :crossed_fingers: that that puzzling symptom won't recur in the wild. If it does, we'll debug it then.

---

Thanks again for the revisions!

I think you have several followup items, some of them quick -- and you mentioned some in your last comment.

To the question in your last comment:

> I think we have `@flow strict-local` now or we are talking about different files?

I meant `Lightbox.js`. Which doesn't yet, so this is one open followup item.

borisyankov:
> I meant `Lightbox.js`. Which doesn't yet, so this is one open followup item.

Oh yeah, sure, will do that next.

Working on a few follow-up commits, but seeing the agreed changes have not made it to the web yet, we would be fine for that release without them (different icon for 'unavailable' etc.)

gnprice:
And merged! Thanks again for all your work on this.

I went and split up a couple of the commits that needed splitting. In one case, one of the needed commits was basically one of the commits in @jainkuniya 's #3362... so I cherry-picked that commit here :stuck_out_tongue_closed_eyes: .

There are various follow-up items mentioned in the thread above, and there are some further cleanups to the avatar logic that I have in some terse notes locally but I don't think I've described. I'll try to collect all those followups in a comment after this one, so they don't fall through the cracks. (I know you're already working on some of them.)

And then I'll make a new release with this change!

gnprice:
Follow-up items, with details in various comments above (Ctrl+F is effective after you tell GitHub to expand the middle of the thread; that widget in turn can be found by Ctrl+F for "hidden item"):

* Lightbox strict-local
* Translations for pre-supplied values for UserStatusScreen
* File a bug with Prettier upstream for that Unicode (perhaps specifically non-BMP) issue

---
Other follow-up items to clean up the `*Avatar*` components:
* A number of props are redundantly described across several components out of `UserAvatar`, `UserAvatarWithPresence`, and `OwnAvatar`; in types, jsdoc, and defaults. I think this is the root cause of a number of specifics mentioned below.
  * Fortunately, this is a solvable problem! `OwnAvatar` and `UserAvatarWithPresence` can defer to `UserAvatar` for these details, getting things DRY.
  * For examples, see the various `*Input*` components that wrap `Input`. In particular `InputWithClearButton`, `PasswordInput`, and commits a15c446d2 and 31523e10b on the late `MultilineInput`.
* Absolute vs. potentially-relative URLs:
  * `UserAvatar` takes absolute only, while `UserAvatarWithPresence` takes potentially-relative. They should be made consistent, one way or the other -- pushing the call to `getAvatarUrl` either inward to `UserAvatar` or outward to callers.
  * Inward makes `OwnAvatar` still simpler, and it better aligns `UserAvatar`'s interface with its name. So I think that's probably preferable.
  * Also the jsdoc on `UserAvatar` falsely says it takes potentially-relative.
* Defaults for `shape`:
  * Redundant defaults to `rounded` at several layers -- should be at just one, similar to `size` in this thread.
  * Divergent defaults, to `rounded` and `circle`. There's no reason showing the self-user's own avatar should inherently be associated with a different shape from showing another user's -- so the choice of `circle` is better put at the caller.
    * (With the `*Input*` technique mentioned above for typing a wrapper component, it becomes basically zero-boilerplate for `OwnAvatar` to offer the `shape` prop, which helps here.)
* Type for `shape`: it's written as `string` on the inner component, but really has a more specific type.
* Defaults of bogus empty-string values, in `UserAvatarWithPresence` for `avatarUrl`, `email`, and `realm`.


---

Other follow-up items:
* The logic in `getMediumAvatar` is fairly messy. For one thing -- why is there a `//g` option on that regexp? That seems like if it had any effect it would be a bug.
* `realm_message_retention_days: ?number` in `InitialDataRealm` can't be quite right; see below.
* I plan to make a pass over the various jsdoc involved in this branch for some small style adjustments.
* I'll do a `tx` push and pull; that will conveniently sync the new strings into all the languages' message catalogs, as well as present them to our kind translators.

---

Other comments:
* In one of today's revisions I noticed the commit that narrowed the `avatar_url` types changed from saying `string | void` to saying `string | null` -- because it turns out the server sends `null`. That is good information that I'm glad we learned! That's the kind of detail that I hope to learn and record when making the effort to write down good specific types. :slightly_smiling_face: 
  * Hmm, in fact that reminds me of a general point: there is no `undefined` in JSON. There's `null`, and there's a property just not existing. So any time we have a `?foo` or a `foo | void` in our types for data we get from the server, that's almost certainly mistaken.
  * At a quick grep, I see just one remaining example of that: `realm_message_retention_days: ?number` in `InitialDataRealm`. So that'd be good to clear up too.

* Thanks for this description:

  >  What is funny, we still needed it to have the state be mutated before the `updateStatusText` call.
  > The state mutations are batched so that could have been expected.
  > I wasn't sure `setTimeout` guarantees us the state would be mutated, though I think it does.
  > I also experimented with `setState` and a second parameter being a call to `updateStatusText` - that is now guaranteed to be after the mutation.
  > Instead of keeping this, I reworked the code a bit to be more straight-forward and not require that deep knowledge of how state mutations work.

  This all makes sense! And I like the resulting code.

* Thanks for this change in `AccountDetails`! I appreciated the additional cleanup:

  > I did split this into two commits.
  > The new one is called:
  > 'AccountDetails: Use `getAvatarFromUser` and `realm` within'

  If you look back I think you'll see that that the split I was asking for was something different and independent:

  > The commit is a bit confusing because it silently makes a mostly-unrelated (though also good) change: it cuts where we'd been showing the presence indicator overlaid on the avatar.

  So I took care of that split before I merged today. That was the commit I took from #3362 . :wink:

gnprice:
Oh, one other small cleanup that this is likely a good time for, which I just happened to see in the codebase:
```
$ rg -iA1 fixme.*realmbot
src/title/TitlePrivate.js
21:    // $FlowFixMe: sort out RealmBot
22-    const avatarUrl: string | null = user.avatar_url;

src/pm-conversations/PmConversationList.js
53:            // $FlowFixMe: sort out RealmBot
54-            const avatarUrl: string | null = user.avatar_url;
```

In reality I think cross-realm bots have avatars? But we're always falling back to gravatars for them. Seems the data is getting misplaced somewhere.

PR #3350, opened 2019-02-16
Issue #3015 Fix caching in "for narrow/streamId" selectors.

jainkuniya:
part of #3015
Converted all to new style.
```
➜  zulip-mobile git:(issue-3015) perl -0ne '
          print "$ARGV: $1\n" while (/^export const (\w+) = [^\n]*=>\s*createSelector/msg)
        ' src/**/*.js
```

### DIFF
--- a/src/chat/narrowsSelectors.js
+++ b/src/chat/narrowsSelectors.js
@@ -103,48 +103,48 @@ export const getRecipientsInGroupNarrow: Selector<UserOrBot[], Narrow> = createS
     }),
 );
 
+// Prettier mishandles this Flow syntax.
+// prettier-ignore
 // TODO: clean up what this returns.
-export const getStreamInNarrow = (
-  narrow: Narrow,
-): Selector<Subscription | {| ...Stream, in_home_view: boolean |}> =>
-  createSelector(
-    getSubscriptions,
-    getStreams,
-    (subscriptions, streams) => {
-      if (!isStreamOrTopicNarrow(narrow)) {
-        return NULL_SUBSCRIPTION;
-      }
-
-      const subscription = subscriptions.find(x => x.name === narrow[0].operand);
-      if (subscription) {
-        return subscription;
-      }
-
-      const stream = streams.find(x => x.name === narrow[0].operand);
-      if (stream) {
-        return {
-          ...stream,
-          in_home_view: true,
-        };
-      }
-
+export const getStreamInNarrow: Selector<Subscription | {| ...Stream, in_home_view: boolean |}, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  state => getSubscriptions(state),
+  state => getStreams(state),
+  (narrow, subscriptions, streams) => {
+    if (!isStreamOrTopicNarrow(narrow)) {
       return NULL_SUBSCRIPTION;
-    },
-  );
+    }
 
-export const isNarrowValid = (narrow: Narrow): Selector<boolean> =>
-  createSelector(
-    getStreams,
-    getAllUsersByEmail,
-    (streams, allUsersByEmail) => {
-      if (isStreamOrTopicNarrow(narrow)) {
-        return streams.find(s => s.name === narrow[0].operand) !== undefined;
-      }
+    const subscription = subscriptions.find(x => x.name === narrow[0].operand);
+    if (subscription) {
+      return subscription;
+    }
 
-      if (isPrivateNarrow(narrow)) {
-        return allUsersByEmail.get(narrow[0].operand) !== undefined;
-      }
+    const stream = streams.find(x => x.name === narrow[0].operand);
+    if (stream) {
+      return {
+        ...stream,
+        in_home_view: true,
+      };
+    }
 
-      return true;
-    },
-  );
+    return NULL_SUBSCRIPTION;
+  },
+);
+
+export const isNarrowValid: Selector<boolean, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  state => getStreams(state),
+  state => getAllUsersByEmail(state),
+  (narrow, streams, allUsersByEmail) => {
+    if (isStreamOrTopicNarrow(narrow)) {
+      return streams.find(s => s.name === narrow[0].operand) !== undefined;
+    }
+
+    if (isPrivateNarrow(narrow)) {
+      return allUsersByEmail.get(narrow[0].operand) !== undefined;
+    }
+
+    return true;
+  },
+);
--- a/src/topics/topicSelectors.js
+++ b/src/topics/topicSelectors.js
@@ -4,6 +4,7 @@ import { createSelector } from 'reselect';
 import type {
   MuteState,
   Narrow,
+  GlobalState,
   Selector,
   Stream,
   StreamsState,
@@ -66,8 +67,7 @@ export const getTopicsForStream: Selector<?(TopicExtended[]), number> = createSe
   },
 );
 
-export const getLastMessageTopic = (narrow: Narrow): Selector<string> =>
-  createSelector(
-    state => getShownMessagesForNarrow(state, narrow),
-    messages => (messages.length === 0 ? '' : messages[messages.length - 1].subject),
-  );
+export const getLastMessageTopic = (state: GlobalState, narrow: Narrow): string => {
+  const messages = getShownMessagesForNarrow(state, narrow);
+  return messages.length === 0 ? '' : messages[messages.length - 1].subject;
+};
--- a/src/topics/topicSelectors.js
+++ b/src/topics/topicSelectors.js
@@ -19,23 +19,23 @@ import { getStreamsById } from '../subscriptions/subscriptionSelectors';
 import { NULL_ARRAY } from '../nullObjects';
 import { isStreamNarrow } from '../utils/narrow';
 
-export const getTopicsForNarrow = (narrow: Narrow): Selector<string[]> =>
-  createSelector(
-    getTopics,
-    getStreams,
-    (topics: TopicsState, streams: StreamsState) => {
-      if (!isStreamNarrow(narrow)) {
-        return NULL_ARRAY;
-      }
-      const stream = streams.find(x => x.name === narrow[0].operand);
+export const getTopicsForNarrow: Selector<string[], Narrow> = createSelector(
+  (state, narrow) => narrow,
+  state => getTopics(state),
+  state => getStreams(state),
+  (narrow: Narrow, topics: TopicsState, streams: StreamsState) => {
+    if (!isStreamNarrow(narrow)) {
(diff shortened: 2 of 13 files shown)


gnprice:
Thanks @jainkuniya !

```
48f4589f3 draftSelector: Fix caching in "for narrow" selectors.
```

* This is one where `createSelector` isn't helping us; let's just cut it out instead.

  (Having a selector is still helpful, because it encapsulates the logic "no draft means empty".)

```
6b535b7a7 unreadSelector: Fix caching in "for narrow" selectors.
536ca8d24 userSelectors: Fix caching in "for email" selectors.
```

* Looks good, thanks!

```
04f3dbb77 messageSelectors: Fix caching in "for narrow" selectors.
```

* This bit:
```
+  (state, narrow) => getShownMessagesForNarrow(state, narrow),
```
can more conventionally be written as just `getShownMessagesForNarrow`.

Ditto in the next selector.


* You don't appear to use this first sub-selector at all:
```
+export const getAnchorForNarrow: Selector<number, Narrow> = createSelector(
+  (state, narrow) => narrow,
```
So leave it out. :slightly_smiling_face: 

Ditto at least once in a later commit.


* Huh -- was there an error you got without this bit? What was it?
```
-) => {
+): number => {
```
If you just happened to notice that it could use a type annotation, that's fine; but do mention in the commit message. But if there was an error, that may be a sign this is papering over an issue.


```
78bf09a44 topicsSelectors: Fix caching in "for narrow" selectors.
```

Couple of the same points as above.


```
d748c3c8b topicsSelectors: Fix caching in "streamId" selectors.
```

* This one can be further improved like so:
```
+export const getTopicsForStream: Selector<?(TopicExtended[]), number> = createSelector(
+  (state, streamId) => streamId,
+  state => getTopics(state),
[...]
+  state => getStreamsById(state),
```
becomes
```
export const getTopicsForStream: Selector<?(TopicExtended[]), number> = createSelector(
  (state, streamId) => getTopics(state)[streamId],
  (state, streamId) => getStreamsById(state)[streamId],
```

That's probably best left as a separate commit, though -- just happened to notice it while I read.

```
1390c4152 [narrows, fetching]Selector: Fix caching in "for narrow" selectors.
```

* Couple of the same points as above.

* Also, this might be simpler to read as two separate commits.

```
bb983d5d0 subscriptionSelectors: Fix caching in "for narrow" selectors.
dcea418f1 subscriptionSelectors: Fix caching in "for streamId" selectors.
```

* Looks good, thanks!

* The "FromId" selectors are also calling out to be addressed as cases of #3339. :smile: Also `getTitleBackgroundColor`.

```
c6057bd50 titleSelectors: Fix caching in "for narrow" selectors.
```

* One annoying wrinkle in Flow:
```
-export const getTitleBackgroundColor = (narrow?: Narrow): Selector<string> =>
+export const getTitleBackgroundColor: Selector<string, ?Narrow> = 
```

Writing `?: Narrow` and `?Narrow` are unfortunately not equivalent!  The latter is less specific -- it includes `null`.

We don't intend to pass `null` here, so let's keep that information by writing a more specific type: `Narrow | void`.

gnprice:
And just merged 5 of these commits, as 830160202..414e48cc6 ! Thanks again.

Two of them have opportunities for further improvements as followups. Because they might be otherwise easy to miss, I'll highlight them here -- search above for:

> This one can be further improved like so

> The "FromId" selectors are also calling out to be addressed as cases of #3339. ￼

jainkuniya:
> That's probably best left as a separate commit, though -- just happened to notice it while I read.

Let's do it in a separate PR :)

jainkuniya:
>  Also, this might be simpler to read as two separate commits.


Both selectors are too interconnected, and it was easy to fix flow errors for single commit. So I made single commit 😉

gnprice:
> Two of them have opportunities for further improvements as followups.

I was looking back at this PR just now and noticed this line. To avoid these falling through the cracks, I went and did them; pushed as 9319af040..bea1ee587.

gnprice:
Thanks again @jainkuniya for your work on this! I'd like to complete the issue this addresses, because there are other changes we want to make to many of these selectors.

So, would you like to try rebasing this series onto current master? Please also take a look back through my comments above; some of them still apply to the current version of the PR.

I'll also go ahead and address the issue on individual selectors when they come up. But it'd be good to get ahead of that and fix it for all of them.

jainkuniya:
sure will rebase it 👍 
lots of conflicts 😅

src/topics/topicSelectors.js
```diff
@@ -66,8 +66,8 @@ export const getTopicsForStream: Selector<?(TopicExtended[]), number> = createSe
...
-export const getLastMessageTopic = (narrow: Narrow): Selector<string> =>
-  createSelector(
-    state => getShownMessagesForNarrow(state, narrow),
-    messages => (messages.length === 0 ? '' : messages[messages.length - 1].subject),
-  );
+export const getLastMessageTopic: Selector<string, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  (state, narrow) => getShownMessagesForNarrow(state, narrow),
+  (narrow, messages) => (messages.length === 0 ? '' : messages[messages.length - 1].subject),
+);
```
gnprice: This is an example where the caching isn't doing anything useful, even if we fix it so the caching happens. That's because the body isn't doing anything expensive: it's doing O(1) work. So this can turn into a plain function with no `createSelector`.

src/topics/topicSelectors.js
```diff
@@ -66,8 +66,8 @@ export const getTopicsForStream: Selector<?(TopicExtended[]), number> = createSe
...
 
-export const getLastMessageTopic = (narrow: Narrow): Selector<string> =>
-  createSelector(
-    state => getShownMessagesForNarrow(state, narrow),
-    messages => (messages.length === 0 ? '' : messages[messages.length - 1].subject),
-  );
+export const getLastMessageTopic: Selector<string, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  (state, narrow) => getShownMessagesForNarrow(state, narrow),
+  (narrow, messages) => (messages.length === 0 ? '' : messages[messages.length - 1].subject),
```
gnprice: Moot because of the other comment, but note that `narrow` isn't being used here. So that and the `(state, narrow) => narrow` can be dropped.

src/chat/fetchingSelectors.js
```diff
@@ -1,9 +1,13 @@
...
 /** The value implicitly represented by a missing entry in FetchingState. */
 export const DEFAULT_FETCHING = { older: false, newer: false };
 
-export const getFetchingForNarrow = (state: GlobalState, narrow: Narrow): Fetching =>
-  getFetching(state)[JSON.stringify(narrow)] || DEFAULT_FETCHING;
+export const getFetchingForNarrow: Selector<Fetching, Narrow> = createSelector(
+  (state, narrow) => narrow,
+  (state, narrow) => getFetching(state),
+  (narrow, fetching) => fetching[JSON.stringify(narrow)] || DEFAULT_FETCHING,
+);
```
gnprice: Similar to `getLastMessageTopic`: this is doing O(1) work, nothing expensive, so better to drop `createSelector` entirely and just make it a function.

src/chat/narrowsSelectors.js
```diff
@@ -104,47 +104,46 @@ export const getRecipientsInGroupNarrow: Selector<UserOrBot[], Narrow> = createS
...
-  narrow: Narrow,
-): Selector<Subscription | {| ...Stream, in_home_view: boolean |}> =>
-  createSelector(
-    getSubscriptions,
-    getStreams,
-    (subscriptions, streams) => {
-      if (!isStreamOrTopicNarrow(narrow)) {
-        return NULL_SUBSCRIPTION;
-      }
+// prettier-ignore
```
gnprice: What's the reason for the prettier-ignore?

These are definitely sometimes the right thing -- just curious what the reason is in this instance.

src/title/titleSelectors.js
```diff
@@ -14,14 +14,14 @@ export const DEFAULT_TITLE_BACKGROUND_COLOR = 'transparent';
...
+  (state, narrow) => narrow,
+  state => getSubscriptions(state),
+  (narrow, subscriptions) => {
+    if (!narrow || !isStreamOrTopicNarrow(narrow)) {
+      return DEFAULT_TITLE_BACKGROUND_COLOR;
+    }
+    const streamName = narrow[0].operand;
+    return (subscriptions.find(sub => streamName === sub.name) || NULL_SUBSCRIPTION).color;
+  },
+);
```
gnprice: See the first couple of commits in #3914 :smile: 

Actually, in particular https://github.com/gnprice/zulip-mobile/commit/954e6b72b8e430dfa880b755d9c42b60f7b52140 is a good example to look at for the comments I made above about a couple of selectors where the caching isn't useful.

gnprice:
And I just did a grep at the tip of this branch. It looks like there's just one case of #3015 left! Namely:
```
src/drafts/draftsSelectors.js
7-export const getDraftForNarrow = (narrow: Narrow): Selector<string> =>
8:  createSelector(
```

(I started with `rg createSelector`. Almost all the results were clearly good, as far as #3015 is concerned. The exceptions were a few where the line was just `  createSelector(`, so I got more context: `rg -C8 createSelector`, or in fact `-C1` would have been enough. Two of the three were also good, and this was the other.)

src/chat/narrowsSelectors.js
jainkuniya: as below line is too long, prettier splits it into two line like:
```
export const getStreamInNarrow: Selector<Subscription | {| ...Stream, in_home_view: boolean |},
  Narrow, > = createSelector(
  (state, narrow) => narrow,
```

And then lint complains:
```
Running lint

  107:41  error  There must be no space at end of "Selector" generic type annotation  flowtype/generic-spacing
```

gnprice: Got it. The same issue as in a couple of other spots in this file, then:
```js
// Prettier mishandles this Flow syntax.
// prettier-ignore
export const getShownMessagesForNarrow: Selector<$ReadOnlyArray<Message | Outbox>, Narrow> =
```

I'll add the same comment here.

gnprice:
Merged! Thanks again @jainkuniya .

I made some small adjustments before merge, mostly to the summary lines -- take a look at the resulting commits:

923f31862 message [nfc]: Fix caching in getRenderedMessages.
f0f2b43ee topics [nfc]: Drop unhelpful caching on getLastMessageTopic.
cdd39ae6f topics [nfc]: Fix caching on getTopicsForNarrow.
0b0e7a702 narrows [nfc]: Fix caching on getStreamInNarrow.
993112c60 drafts [nfc]: Drop unhelpful caching on getDraftForNarrow.

Issue #3353, opened 2019-02-19
Travis CI build failure due to lint


AB261:
The command "tools/test --full --coverage" is exiting with return value 1 and is preventing PR from getting accepted. The problem lies in lint 
![lint_problems_zulip_mobile](https://user-images.githubusercontent.com/17261046/52987970-73a1e400-3423-11e9-9ccc-096af9876ff6.png)

AB261:
@zulipbot claim

gnprice:
Fixed in #3533. Thanks @AB261 for the report!

PR #3356, opened 2019-02-19
Fix Icon type issues

borisyankov:


gnprice:
You scared me for a moment by saying this was about "type issues"! :wink: I went and quick reran `yarn flow` to check that I hadn't somehow overlooked an issue with that.

These are not type errors. They're lint errors, complaining about a naming convention. Should certainly be fixed because they cause annoyance -- but a very different thing.

gnprice:
The wording of this lint message is puzzling. Did you look at the documentation for it?

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md

The point is that in the files where it's complaining, we're saying something like
```
import Icon from '../common/Icons';
```
Once we start having an export named `Icon` from that module, the lint rule fires because that looks like we meant to say `import { Icon } from '../common/Icons';`.

I think `IconType` is an appropriate name, so I'll take this change. (It's parallel to `ComponentType`; a value of this type is not a specific instance of a React component, but the component type in general.)

But it isn't addressing the real point the lint rule is making, which is that our names are inconsistent here about the plural. We're importing the default export, but giving it a name which is *almost* but not quite the same as the name of the module -- and that's confusing. So I'll fix that, too.

gnprice:
Merged, with an additional commit to fix the underlying naming issue. Thanks @borisyankov !

gnprice:
(And oops, I got the issue number wrong in my edited commit message. Went and closed the issue manually -- which was #3353.)

Issue #3361, opened 2019-02-22
accountDetails: Presence status is being shown on avatar.


jainkuniya:
In account details screen, we already show presence status below avatar. So there is no need of showing it on avatar image too.
![image](https://user-images.githubusercontent.com/18511177/53250905-6154ce00-36e1-11e9-92a4-290b9ca3e7b6.png)

@zulipbot claim

gnprice:
Thanks @jainkuniya for reporting this!

@borisyankov , I think one of the changes you had queued up in a PR was going to fix this. Please mention this issue in that commit. :slightly_smiling_face:

PR #3362, opened 2019-02-22
  accountDetails: Do not show presenceStatus indicator in avatar.

jainkuniya:
Issue #3361

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 import { View, Dimensions, StyleSheet } from 'react-native';
 
 import type { User } from '../types';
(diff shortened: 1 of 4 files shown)


borisyankov:
This will be resolved with #3344

You better wait with your work on avatar-related stuff, as the changes in that PR definitely will conflict.

gnprice:
Thanks @jainkuniya !

This is an area where @borisyankov happens to have some open refactors from just the past few days, which include taking care of this. I believe those will be in soon, so I'll close this. Sorry for the race condition.

gnprice:
BTW, just now I ended up including the tip commit of these two commits in the merged version of #3344 (as d1a39768b), because that turned out to be exactly the change needed there :smile: . Thanks again!

PR #3363, opened 2019-02-23
 presenceStatusIndicator: Subscribe presence to more deeper in the state.

jainkuniya:
Presence state changes in every few minutes. So to make sure that this
components only gets re-rendered when presence of relevant person
changes, subscribe it deeper in the presence state.


gnprice:
Thanks @jainkuniya !

* This is an example where `createSelector` doesn't do us any good:
  * The selector isn't doing any work.
  * It isn't making a new object.

  So let's leave that bit out.

* Let's name this `...ForEmail`. That helps us keep a consistent pattern:
  * `getFooByBar` gets a map of *all* foo, indexed by bar.
  * `getFooForBar` takes bar as a second parameter, and gets just the one foo that corresponds to that bar.

  We only recently started having selectors of that second type; see `narrowsSelectors` for existing examples.

* Hmm, but actually: given that `createSelector` isn't helpful, we can just write something like this:
```
  presence: getPresence(state)[props.email],
```

  That's what we have in `ActivityText` now, for example.

* Want to do the same thing for `usersByEmail` while you're here?

PR #3364, opened 2019-02-24
Cleaned up flow type for selectors in narrowSelector

AB261:
Issues in focus: #2971, #3166
Removed redundant messages word in the logical operation performed under getIfNoMessages and added flow type for it

Added flow type for getStream in narrow. This one was different because the selector returned either Subscription if the Stream was subscribed to before, or Stream + in_home_view=true if it was not subscribed to before. A type could have been declared for the latter case but it seemed unnecessary so instead, I used spread syntax to define the return type. 

Tests ran:
- [x] tools/test

### DIFF
--- a/src/chat/narrowsSelectors.js
+++ b/src/chat/narrowsSelectors.js
@@ -2,7 +2,17 @@
 import isEqual from 'lodash.isequal';
 import { createSelector } from 'reselect';
 
-import type { GlobalState, Message, Narrow, Outbox, RealmBot, Selector, User } from '../types';
+import type {
+  GlobalState,
+  Message,
+  Narrow,
+  Outbox,
+  RealmBot,
+  Selector,
+  User,
+  Stream,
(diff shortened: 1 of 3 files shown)


gnprice:
Looks good -- thanks @AB261 !

Two small things:
* Because each of these is a bit more than something totally boring and mechanical, and each one makes sense on its own, it's clearest to split them as two separate changes.
* The type `Stream` is already an exact object type:
```
export type Stream = {|
  stream_id: number,
[...]
|};
```
  so the `$Exact` is redundant and can be left out. (We probably have some of those elsewhere in the codebase, left over from before we went and made almost all our object types exact -- probably you saw one of those. They can be taken out too.)

I'll go ahead and merge, with those tweaks.

gnprice:
Merged -- thanks again!

I also added a commit on top, 75b7381cc. Take a look at that and the two commits preceding it.

PR #3366, opened 2019-02-24
userGroupItem, icon: Set `color` according to the theme.

jainkuniya:
So that it can match color with the color of the text in both the
themes.

This PR:
image
image

on master:
image
image


gnprice:
Thanks @jainkuniya, looks great -- merged!

The main thing I tweaked in the commit message is to link back to this PR, for the nice screenshots. :slightly_smiling_face:

PR #3368, opened 2019-02-25
drafts: Cleanup state when switching accounts

borisyankov:
We previously did not reset the state on switching accounts but
only on logout.

This fix has the added benefit that a user can cleanup the drafts
by switching accounts (but this is also a aside effect that we will
not persist the drafts if the user switches accounts and gets back).

### DIFF
--- a/src/drafts/draftsReducers.js
+++ b/src/drafts/draftsReducers.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import type { DraftState, Action } from '../types';
-import { DRAFT_UPDATE, LOGOUT } from '../actionConstants';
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @borisyankov !

* My first thought on this change (and the issue it fixes) is: this looks like it was really easy to get wrong. How can we make it less error-prone?

* Concrete immediate versions of that are:
  * Are there more actions that should be in this list?
  * Are there more reducers that should get a similar change?

* Those immediate versions (which are pretty straightforward to go answer), I'd really like to see in the commit message. Otherwise, in general there's a high probability that there are more things we should similarly fix. And once we're looking, it's so easy to fix them. :slight_smile:

borisyankov:
Closing as not likely to get worked on further, and could be addressed as bigger rethinking/reworking.

Issue #3307, opened 2019-01-23
Enhancing architecture docs


AB261:
Many of the folders in the /src directory are not mentioned in the architecture.md file such as lightbox. It would help beginners understand the structure more easily if this was done.

AB261:
@zulipbot claim

chrisbobbe:
Thanks @AB261 for reporting this, and I'm sorry it took so long to respond! As you said at https://github.com/zulip/zulip-mobile/pull/3371#issuecomment-576723345, you're too busy to keep working on this issue right now, which is very understandable because you went so long without receiving feedback.

So, I've closed that PR, and this issue is now open to someone who would like to get more familiar with the /src directory and help others to understand it too. 🙂 The doc in question is at https://github.com/zulip/zulip-mobile/blob/master/docs/architecture.md.

A good solution to this will be to write a brief but accurate description on each of the files in the /src directory, which has grown much larger since that part of the doc was last updated. _This will require digging into the directory and reading the code itself._ Also, please note that some of those directories (like redux-persist-migrate) are local copies of NPM packages that we've cloned to adapt to our own needs, so that should be clear in the descriptions, with links to the packages themselves, and the reason for the clone.

If something isn't clear after doing some investigation, please post about it here or on chat.zulip.org. And if something in that doc outside the /src section could be improved, feel free to do that in a separate commit.

chrisbobbe:
I see that this hasn’t attracted much activity, which is fine — @gnprice mentioned today that the maintenance burden of documenting everything in the /src directory is quite large, considering each file should have its own code comments anyway, and I agree. So, I’m closing this. 🙂

PR #3371, opened 2019-02-27
Update architecture.md

AB261:
Issue #3307 
I think having an index of sorts of what is inside the source directory would help tremendously for beginners to feel more comfortable with the codebase. Although some folders are self-explanatory, new people might not have the same coding background or might not understand as easily, therefore, I am going through all the folders and trying to get an idea of the function they perform( it will help me also :smile: ). This is the first time through and so I wanted to make a pr to ask if any changes are required in the way I have written the description or if something needs to be added/deleted/updated.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -45,9 +45,20 @@ checker (unused)
 The source directory is broken up into subdirectories corresponding to
 components of the app:
 * `account` - login, logout, and user account
+* `account-info` - account details,status switiching, logoutbutton
 * `api` - clients for the Zulip server API
+* `animation` - rotating and scaling animation components
+* `autocomplete` - autocomplete items for people,stream and topics
+* `boot` - sets up store, translationprovider, aysncstorage,logging , reducers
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Hi @AB261, thanks for this, and I agree, it would be great to update this page in the docs, if you're still around! If not, maybe someone else might like to take on this project? 🙂 

Since it took so long for you to get a response (sorry!) there have been several updates to that directory that it would be good to include as well. Many of the labels proposed in this PR could be improved for accuracy and polish, but this is a good start to adding documentation that's long-overdue. 🙂

Some of those directories (like redux-persist-migrate) are local copies of NPM packages that we've cloned to adapt to our own needs, so that should be clear in the descriptions, with links to the packages themselves.

AB261:
Hey! I am a bit busy right now so I won't be able to take it up. If someone else is willing then they can go ahead and take it up.

chrisbobbe:
OK! I'll close this PR and bump the issue, then. Thanks for reporting this!

PR #3377, opened 2019-03-01
userStatus, input: Set max length to 60.

jainkuniya:
Server only supports status message with max length 60, and otherwise
throws error on updating it. So set max length in the input itself so
that user can enter status of only valid length.


gnprice:
Thanks @jainkuniya -- merged!

Issue #3378, opened 2019-03-02
[Android] composeBox: Nothing happens after selecting item in autocomplete pop up.


jainkuniya:
This is Android specific. On iOS it works absolutely fine.

With the awesome `git` tool, `git bisect` I was able to find out which commit introduced it. And it was my commit only. I have never imagined that just `backgroundColor` can affect in such a way.

Commit which introduced this is: 041ce55e

I guess issue is in RN, when there is `absolute` component inside a `view` which has `backgroundColor` property on Android, then it doesn't receive any touchable event. And thus in our case nothing happens on touching items in autocomplete list. Also not able to scroll the autocomplete list.

Reported: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Problem.20with.20mention.20popup

@zulipbot claim

gnprice:
Thanks @jainkuniya for tracking this down!

And glad `git bisect` worked well for you. It is indeed a powerful tool -- especially helpful for diagnosing an issue like this one that is triggered by something totally surprising that seems like it shouldn't have any connection. :smile: 

One further observation: it's not quite that *nothing* happens when you tap an item in the popup. Instead, the touch event goes right "through" the popup, and hits the message list. For example, if you tap a spot in the popup that's hiding a sender avatar, or a topic recipient bar, you'll navigate to the account details or the topic narrow respectively.

We had a previous bug with exactly the same symptoms, of taps on the autocomplete popup going through to the message list. Let's see if I can find it...

gnprice:
The previous issue was #3209 . That one happened only some of the time.

That one was also Android-only, and like this one it was pretty mysterious:

> This is a regression in 21.0.104 (and 21.1.105). Using `git bisect` to do a binary search, I found that it was introduced in 824c908cc aka #3065. It doesn't seem obvious how that caused this issue, though, even when I reread it now knowing that it somehow did.

That one was fixed by #3210 aka 3910f2d69 . Hmm, and that fix also involved avoiding `position: 'absolute'`:
```
-      ...(canSend ? {} : { opacity: 0, position: 'absolute' }),
+      ...(canSend ? {} : { display: 'none' }),
```

gnprice:
Copying back here a comment you made on your PR #3379 to fix this:

> I tried to reproduce it on new RN project, but can't. May be I am not able to get exact scenario where it will be reproducible or may be there is not RN issue, just we are doing something wrong.
> 
> But on the current master if we just remove
> 
> ```js
[...]

PR #3379, opened 2019-03-02
[Android] autocomplete: Fix text doesn't autocompletes on selecting item

jainkuniya:
This got introduced in 41ce55, where background color is set to the
parent of `absolute` autocomplete view.

This seems to be a RN issue that, `absolute` positioned components
doesn't receive touchable events if inside component which has
background color set. (will try to verify it on new RN project)

So set background color to compose input view, sibling of autocomplete
view instead of parent of autocomplete view.

Fixes: #3378

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -332,13 +332,8 @@ class ComposeBox extends PureComponent<Props, State> {
     };
 
     return (
-      <View style={style}>
-        <View
-          style={[
-            this.styles.autocompleteWrapper,
-            { marginBottom: safeAreaInsets.bottom + height },
-          ]}
-        >
+      <View>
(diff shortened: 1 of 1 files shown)


jainkuniya:
I tried to reproduce it on new RN project, but can't. May be I am not able to get exact scenario where it will be reproducible or may be there is not RN issue, just we are doing something wrong.

But on the current master if we just remove 
```javascript
backgroundColor: 'rgba(127, 127, 127, 0.1)',
```
from `style` then it is not reproducible.

gnprice:
Thanks @jainkuniya ! Merged as 3d6968523. (Neglected to update this PR on push, so it'll show as closed rather than merged, sorry.)

I think it'd be quite useful to spend some more time pursuing the underlying RN issue, because this looks to be the second time we've run into it. Filed #3383 for that.

PR #3380, opened 2019-03-02
ios, touchable: Set style to view wrapper instead of Touchable.

jainkuniya:
On master:
image

This PR:
image

### DIFF
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -2,7 +2,7 @@
 import { connect } from 'react-redux';
 
 import React, { PureComponent } from 'react';
-import { View } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 
 import type { Dispatch, GlobalState, Narrow } from '../types';
 import styles, { BRAND_COLOR } from '../styles';
@@ -12,6 +12,15 @@ import TitleNavButtons from '../title-buttons/TitleNavButtons';
 import { DEFAULT_TITLE_BACKGROUND_COLOR, getTitleBackgroundColor } from '../title/titleSelectors';
 import { foregroundColorFromBackground } from '../utils/color';
 import { navigateBack } from '../actions';
+
+const componentStyles = StyleSheet.create({
+  navContent: {
+    flex: 1,
+    justifyContent: 'space-between',
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+});
 
 type Props = {|
   dispatch: Dispatch,
@@ -36,8 +45,10 @@ class ChatNavBar extends PureComponent<Props> {
             dispatch(navigateBack());
           }}
         />
-        <Title color={color} narrow={narrow} />
-        <TitleNavButtons color={color} narrow={narrow} />
+        <View style={componentStyles.navContent}>
+          <Title color={color} narrow={narrow} />
+          <TitleNavButtons color={color} narrow={narrow} />
+        </View>
       </View>
     );
   }
--- a/src/title/TitlePrivate.js
+++ b/src/title/TitlePrivate.js
@@ -28,14 +28,16 @@ class TitlePrivate extends PureComponent<Props> {
     // $FlowFixMe: sort out CrossRealmBot
     const avatarUrl: string | null = user.avatar_url;
     return (
-      <Touchable onPress={this.handlePress} style={styles.navWrapper}>
-        <UserAvatarWithPresence size={32} email={user.email} avatarUrl={avatarUrl} />
-        <ViewPlaceholder width={8} />
-        <View>
-          <Text style={[styles.navTitle, { color }]} numberOfLines={1} ellipsizeMode="tail">
-            {user.full_name}
-          </Text>
-          <ActivityText style={styles.navSubtitle} color={color} email={user.email} />
+      <Touchable onPress={this.handlePress}>
+        <View style={styles.row}>
+          <UserAvatarWithPresence size={32} email={user.email} avatarUrl={avatarUrl} />
+          <ViewPlaceholder width={8} />
+          <View>
+            <Text style={[styles.navTitle, { color }]} numberOfLines={1} ellipsizeMode="tail">
+              {user.full_name}
+            </Text>
+            <ActivityText style={styles.navSubtitle} color={color} email={user.email} />
+          </View>
         </View>
       </Touchable>
     );


jainkuniya:
(This got introduced in 8f99a62220a16182a89896548cddb93356e138e7)

borisyankov:
I would be extremely careful doing a change like this to `Touchable` which is used in 24 places in the app. It is quite possible that styling the internal `View` does not result always to the same style.

src/common/Touchable.js
```diff
@@ -53,7 +54,7 @@ export default class Touchable extends PureComponent<Props> {
           onPress={onPress}
           onLongPress={onLongPress}
         >
-          <View>{children}</View>
+          <View style={wrapperStyle}>{children}</View>
```
jainkuniya: `TouchableHighlight` is different from `TouchableNativeFeedback`. 
`TouchableHighlight` don't have `flex:1` by default on iOS (or may on Android too). But `TouchableNativeFeedback` have, at least on Android. So if we directly apply `style` to `view` then, then it will break in components like `Buttons` as then touchable will only be region on text not full button. 

And if we always provide `flex: 1` to `TouchableHighlight`, then it will break where there are 2-3 touchable in single view, as then all touchable will take equal space. 

So we can't say that we want `flex:1` always on `TouchableHighlight`, but sometimes we want. Thus this need to be configurable.

And styles like `flexDirection` needs to be provided to `view`, because `TouchableHighlight` has only one child so   `flexDirection` has no effect. 

So it's better to have `style` & `wrapperStyle`.

src/nav/ChatNavBar.js
```diff
@@ -36,8 +36,10 @@ class ChatNavBar extends PureComponent<Props> {
             dispatch(navigateBack());
           }}
         />
-        <Title color={color} narrow={narrow} />
-        <TitleNavButtons color={color} narrow={narrow} />
+        <View style={styles.navContent}>
```
jainkuniya: This `view` wrapper makes easy to understand the components structure/arrangements.

gnprice:
Thanks @jainkuniya for spotting and fixing this!

* That screenshot does look pretty bad. I'm not seeing that symptom, though. Do you have any ideas on how to reproduce it, or what controls whether it happens?

  * Oh, I see your second commit mentions iOS. Does this happen all the time on iOS?

#### navBar: Create a wrapper around nav content.

* Let's keep the style here in the same file as the `render` that uses it -- that makes things much easier to follow, and avoids adding to global interfaces.

  (We used to have tons of styles in these global places, and I've moved most of them to the individual files where they're used, for this reason.)

* I'm not clear on the purpose of this commit. The commit message is very helpful in identifying the key line that it's intended to add: `justifyContent: 'space-between'`. But I still don't see the step before that -- what problem is being solved by adding that?

#### ios, touchable: Set style to view wrapper instead of Touchable.

* Again it's not clear to me what problem this is solving. Perhaps it empirically resolves the issue? (But can we do that with just one of these two changes?) Even if it does, though, I'd really like to have some theory of why that is, or why this way is more correct than the other way. Otherwise, we're just as likely to run into a similar problem by accident in the future too. And this change adds complexity, so there's a real cost to it.

* In the Android case, we apply the given style to the inner View instead of the outer `TouchableNativeFeedback`. This change basically makes us do the same thing in the iOS case... but only sort of optionally, and only opting in in this one caller. Maybe we should do the same thing in the iOS case all the time?

jainkuniya:
> Does this happen all the time on iOS?

Yup, this is iOS specific.

> But I still don't see the step before that -- what problem is being solved by adding that?

This doesn't solve any problem directly. But this makes design layout more clear.
That at upper layer you have you have `<View style={[styles.navBar` which is just divided into two more components. First one is nav button and other one is rest. So at this point we only need to take care of button and content style. And inside content we can do whatever it will not affect button style.

Currently we have `flex: 1,` in `Title` style via `navWrapper`, which is helping us to achieve what we want, i.e `TitleNavButtons` at right. So f we remove `flex:1`, then it will mess up this title nav button. That means style of title is effecting position of other components, which should not be the case.

So I added this wrapper.

> Maybe we should do the same thing in the iOS case all the time?

No we don't want to this in all cases. We only want this to be done in cases where we want to apply styles like `fledDirection` etc. 

https://github.com/zulip/zulip-mobile/pull/3380#discussion_r261855356

gnprice:
Hmm. Here's what I get on Android after this change:

image

I.e., the exact same symptom as this fixes on iOS. :grin: 

I think the reason is that with your change, we pass `styles.navWrapper` only as `wrapperStyle`, not as `style` -- which means on Android it's ignored.

It's still not clear to me what's the intended meaning of `wrapperStyle`, and how it differs from `style`. I think that unclarity is likely at the root cause of this issue.

Here's one question: why do we need to pass `Touchable` a style to put on an inner `View` at all? Instead of
```
<Touchable style={style} ...>
  ...
</Touchable>
```
why don't we just say
```
<Touchable ...>
  <View style={style}>
    ...
  </View>
</Touchable>
```
at the caller? (And then perhaps drop the extra `View` in `Touchable`.)

That seems more explicit -- the answer to how that `style` is being used and what it means is right there in the caller's code.

In this case, that would mean in `TitlePrivate`, we'd say
```
      <Touchable onPress={this.handlePress}>
        <View style={styles.navWrapper}>
          <UserAvatarWithPresence size={32} email={user.email} avatarUrl={avatarUrl} />
          <ViewPlaceholder width={8} />
          ...
        </View>
      </Touchable>
```

Is there something that goes wrong if we take that approach? If so, what? That would be informative, I think.

jainkuniya:
> (And then perhaps drop the extra View in Touchable.)

Will be very risky. As always `Touchable` expects only one child.
One thing we can do it, add a extra view wrapper when you want styles for `view`. Same as you suggested above.

gnprice:
> As always `Touchable` expects only one child.

(and in your new commit message:)

> `Touchable` can only contain one view inside it

Hmm interesting -- why do you say that?

The prop is called `children`, and the jsdoc says "Components to turn into 'touchable' ones.". So if it really can only be one component, we should fix the jsdoc and find a way to make that clear.

But I don't see why there should be a problem with having several children. In fact that would be pretty puzzling -- because even if there's some quirk with `TouchableHighlight` or `TouchableNativeFeedback`, in both cases the implementation of `Touchable` will give either of those just one child anyway, by wrapping its own `children` in a wrapper `View`.

gnprice:
I've just pushed b78b49d62 to address this, based on the discussion here. Please take a look and confirm that that resolves the issue for you.

Thanks again for spotting this issue and this debugging!

gnprice:
OK, and I spent a good chunk more time today digging into `Touchable` and the underlying upstream components to understand what was actually going on here. The main result is ec32af1a4 , along with some smaller cleanup commits elsewhere. Please take a look!

gnprice:
One outcome is that this prophecy has come true :smile: :

> `Touchable` can only contain one view inside it

That was already true of both the underlying upstream components `TouchableHighlight` and `TouchableNativeFeedback`; but our `Touchable` had had an attempted feature of inserting its own wrapper View to satisfy that condition, allowing multiple children.

Trouble is, that feature made things very confusing on iOS where we used `TouchableHighlight`, because the wrapper View caused the `style` properties which you would use to lay out those multiple children -- like `flexDirection`, or `justifyContent: 'space-between'` -- to not work. As a result, the multiple-children feature effectively wasn't usable.

So I cut it out, and now `Touchable` really does require exactly one child; and it gives you a nice error if you get that wrong. :slightly_smiling_face:

PR #3386, opened 2019-03-06
 Reduce waiting for messages to load by fetching more messages earlier

borisyankov:
The original values were changed several times and the currently
used ones are chosen conservatively to reduce issues we had in
the past.

We no longer have the positioning issues we used to have so it
makes sense to revisit these and change for better user experience.

These changes are a result of extensive testing and analysis.


gnprice:
Thanks @borisyankov ! Merged.

I will be very eager to see the documentation promised in the commit message :smile: :

    These changes are a result of extensive testing and analysis.
    The reasoning for the exact values and the platform-dependent
    behavior will be added as documentation later.

Issue #3345, opened 2019-02-14
Clarify and improve autocomplete-parsing logic


gnprice:
The function `getAutocompleteFilter` controls when we show an autocomplete popup in composing message content. It's currently pretty convoluted and hard to follow, as highlighted in #3290. I see that as the root cause of issue #3289, and I think there are more issues lurking in that logic that are hard to see because of the complexity of the code. So I'd love to see a followup to make it simpler and clearer.

Concretely: There are a lot of conditionals in it that are various special cases. Also it's not at all clear what the various pieces are supposed to mean. (`lastWordPrefix` doesn't really have anything to do with a word, let alone a last word. And `lastIndex` isn't particularly the last index of anything.)

To fix it, I think the main steps are:
* Write jsdoc saying what the function is supposed to mean. What do the return values mean? In particular, what return values signify that there's nothing to autocomplete?
  * Part of this answer will be complicated to explain. That points at the next step, improving the interface so it's less complicated. In the actual branch, the jsdoc can come after the better interface.
* Improve the return value interface. Replace `lastWordPrefix` with a meaningful name, e.g. `code` or `sigil`. In the case where there's nothing to autocomplete, return `null` rather than a mishmash of partial values.
* Replace the convoluted internal logic with a regular expression. See MDN: [Regular Expressions guide](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions), and [RegExp reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp). In particular you'll want [capturing groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references), and the `m` [flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Parameters).
* Think about the edge cases in the logic, and adjust them.
  * What is the condition `['\n', ' ', '#', ':'].includes(text[text.lastIndexOf('@') - 1])` doing? Are there similar cases that should also be included? How can we tell when we're done adding cases there? What does the webapp do?
  * What is the condition `!['\n', ' '].includes(text[lastIndex + 1])` doing? Again, are there similar cases that should also be included? How can we tell when we're done adding cases? And what does the webapp do?

AB261:
Is someone working on this? If not can I work on this?

AB261:
@zulipbot claim

PR #3390, opened 2019-03-09
Fix: Clarify and improve autocomplete-parsing logic

AB261:
Issue: #3345
I created a RegExpression that should make the logic less convoluted. There are a lot of edge cases and I tried to make sure all are covered. In the issue, it was mentioned to use m flag and capturing groups but I found using g flag with m flag easier because reg expression returns the first match and we want the latest match which requires the g flag. I have explained my logic through comments especially matching the @ part which is a bit difficult. I added a test case for corner case wherein @ was preceded by words such as asva@gmail.com . Finally, I will be sending more pr in regards to making the variable names easier to understand. I think point 2 in the things to do for the issue and making the jsdoc.  
Tests ran:
- [x] tools/test


AB261:
Hey **@jainkuniya** ! Thanks for the swift review and I looked into how slack does it and these are my views.
Screenshot_20190309-174619

Screenshot_20190309-173245

Screenshot_20190309-175319
So slack doesn't allow for autocomplete of emoji or stream autocomplete without a space before it.
If we allow for autocomplete without a space before then it could cause problems like where the ** leads to one word being bold and other not (pic 3). I think it is better if we go the way slack is doing it and keep a space before the autocomplete text. I also think it would look nicer to read as well because the emojis would be mushed together.
Let me know what the best course is.

gnprice:
Thanks @AB261 for sending this, and @jainkuniya for the review!

* First: in a change like this that reworks some tricky existing code and also changes its behavior, it's important to make small, separate, individually clear and coherent commits. See our doc on this:

  So your first commit should be at least two separate commits -- one for the refactoring, one for the behavior change. Most likely it should be several separate commits, for different aspects of the refactoring (and different behavior changes, if there's more than one.)

* I agree with @jainkuniya that part of this is pretty complicated.
  * I like the structure with a series of `if / else if` conditions! That is nice and straightforward, and I think an improvement on the existing structure.
  * That regexp is *way* too dense, though.
  * One factor is actually a fact I missed when recommending a regexp in #3345: JS apparently lacks the `x` flag for regexps (as in Perl `/x` or Python `re.X`), which means no comments or whitespace. Sad but true. That puts a tight limit on how much logic can go into a JS regexp before it's unreadably complex.
  * But, I think we don't actually need this complex a regexp here. In particular:
    * Reread the reference docs on what `[]` means. For example, you don't generally need it when there's just one character.
    * Is `[\w\s|@]` really the logic we want there? Maybe change it to be the same as the others. (If that changes the behavior, then have a separate commit first to make the behavior change.) Then you can factor those out, and make the whole thing a lot simpler.
  * Your browser dev console is a great place to experiment with regexps.
    image

Would you try revising with those points in mind?

AB261:
Thanks for the review @gnprice! I will consider your recommendations and make the required changes.

AB261:
I made the regex and the overall logic a lot clearer now. All the filtering is now based on the text itself and I am only using regex match to match only the sigils. I also removed all the unnecessary brackets in the regex making it much much smaller.

gnprice:
Thanks @AB261 for the revision! This regex is definitely a lot simpler :smiley: :
```
/:|[^\w]@|#/gm
```

A few bits of quick high-level feedback:
* The branch has a lot of commits that fix up earlier commits in the branch. You should squash those in -- the goal is clear and coherent commits that describe your eventual understanding of the changes, not the sequence of drafts that led you there. See our guide on commit style: https://zulip.readthedocs.io/en/latest/contributing/version-control.html
  * Note the renaming still belongs as a separate commit. There may be several separate commits in here -- see that guide for discussion.
* The big comment above the regex seems to describe the previous version.
* The `lastIndexOf` stuff should be possible to make much simpler by using the data on the match object that the regexp match returns -- that's what it's for. :slightly_smiling_face:

AB261:
Thanks for the review @gnprice! I am not sure what you mean by making lastIndexOf simpler. The match does not return an index ( as I am using the global flag) so I cannot start building a substring from that point on. I could add the original [\w|\s] to all the RegExp parts ( after the: #, and @) making it capture all the content after the sigil, but I believe the priority is to keep the RegExp simple. Two options can be taken either I can do as I said in the above method making the RegExp hard to grasp, use lastIndexOf based on the sigil found in the expression. Which is preferred? I updated the comments.

PR #3391, opened 2019-03-09
chat: Add types to navigation props in GroupDetailsScreen.js

gayashanbc:
### Description
Add types to navigation props in GroupDetailsScreen.js
### Issues in focus
#2052


gayashanbc:
@gnprice

Issue #3374, opened 2019-02-28
Add support for composing "silent mentions"


timabbott:
We recently added a new markdown syntax for "silent mentions", spelled `@_**Tim Abbott**`.  We should extend the mobile app typeahead to complete that if the user starts their mention typeahead with `@_`.  

(The HTML generated is very similar to the HTML for normal mentions, displayed as a bubble around "Tim Abbott" (no `@`) with markup matching a normal mention + an extra `silent` CSS class that we can use for future styling changes if desired, so I expect no changes are required to support displaying these properly)

ungps:
Hello.

I would like to work on this.
@zulipbot claim

chrisbobbe:
It looks like we should

- add a line like `'@_': PeopleAutocomplete,` to `prefixToComponent` in src/autocomplete/AutocompleteView.js
- study src/autocomplete/getAutocompleteFilter.js and work out a sensible way for it to give '@_' (which has the particular trait of not being a single character) as `lastWordPrefix`.

(Or look at ways to clean up the code more dramatically; but I think I'd try the above approach first.)

PR #3393, opened 2019-03-11
autocomplete: Add support for composing "silent mentions".

ungps:
Add markdown syntax for "silent mentions", which happens when
a user starts their mention with "@_". Add empty "silent" class
for future styling changes.

There is also a PR with some changes on zulip-markdown-parser repo. https://github.com/zulip/zulip-markdown-parser/pull/1

Issue: https://github.com/zulip/zulip-mobile/issues/3374

### DIFF
--- a/src/autocomplete/__tests__/getAutocompleteFilter-test.js
+++ b/src/autocomplete/__tests__/getAutocompleteFilter-test.js
@@ -11,6 +11,18 @@ describe('getAutocompleteFilter', () => {
     selection = { start: 1, end: 1 };
     expect(getAutocompleteFilter('@', selection)).toEqual({ filter: '', lastWordPrefix: '@' });
 
+    selection = { start: 2, end: 2 };
+    expect(getAutocompleteFilter('@_', selection)).toEqual({ filter: '', lastWordPrefix: '@' });
+
+    selection = { start: 4, end: 4 };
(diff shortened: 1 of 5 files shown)


gnprice:
Thanks @ungps !

```

   const filter: string =
     text.length > lastIndex + 1 && !['\n', ' '].includes(text[lastIndex + 1])
-      ? text.substring(lastIndex + 1, text.length)
+      ? text.substring(
+          lastIndex + 1 + (text[lastIndex] === '@' && text[lastIndex + 1] === '_'),
+          text.length,
+        )
       : '';
```

Oof, this is pretty complicated. :weary: 

This code is already too complicated -- see #3345 for discussion. It's already hard to understand, and I think that's hiding some bugs already. It gets significantly more so with this change.

I'm going to declare a moratorium on adding features to this code without making it simpler to compensate. @AB261 is working on that right now in #3390. If you can make small local changes to this spot to keep the complexity down, we can move forward on this independent of #3390; otherwise, let's wait for that work to play out.

Separately:

```
+.silent {
+}
```

This has no effect and isn't necessary; so just leave it out.

gnprice:
Closing, for the reasons above. @ungps , thanks again, and please feel free to send a new PR, now or later, as described here:

> If you can make small local changes to this spot to keep the complexity down, we can move forward on this independent of #3390; otherwise, let's wait for that work to play out.

PR #3396, opened 2019-03-12
Use the built-in types from 'rn-fetch-blob'

borisyankov:
* Use `objectToParams` in `downloadImage` to ensure we are not passing `undefined` to `Authorization`.
* Do a minor upgrade of `rn-fetch-blob` and remove the auto-generated type file `rn-fetch-blob_vx.x.x.js`.
    
`rn-fetch-blob` includes Flow types now, so use them instead.


gnprice:

PR #3405, opened 2019-03-15
usability: Add the possibility to copy only a link from a message.

### LINKED ISSUE #3385: Copying Message to Clipboard

ungps:
Fixes #3385 3385. Long pressing on a link displays in the menu the option
to copy only the link ("Copy link to clipboard") instead of copying
the whole message ("Copy message to clipboard").


gnprice:
Thanks @ungps for working on this!

I'm not a fan of this UX, which feels pretty subtle and easy to miss. I'd rather have long-press on a link just unconditionally copy the link. That's the approach in #3404 , so I'm closing this PR in favor of continuing discussion there and on the issue (#3385).

I see there looks to have been some confusion on the issue thread #3385 about who was working on it; sorry the communication on that didn't work better. FFR, writing `@zulipbot claim` is a good way to make unambiguous that you plan to work on an issue.

PR #3407, opened 2019-03-16
Improve Android notifications.

jainkuniya:
This PR will take following points in account:
* Beautify Android notification #2691
image
* Convert Java code to Kotlin as per requirement.


android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
```diff
@@ -187,19 +187,38 @@ internal data class MessageFcmMessage(
...
+
+                    // `realm_id` was added in the same commit as `server`.
+                    realmId = data.require("realm_id").parseInt("realm_id"),
+
+                    // `realm_uri` was added in server version 1.9.0
+                    // (released 2018-11-06; commit 5f8d193bb).
+                    realmUri = data["realm_uri"]?.parseUrl("realm_uri"),
+
+                    // `user` was already present in server version 1.6.0 .
+                    email = "" //TODO
```
jainkuniya: Remove FCM message does't have this data (`user` i.e`email`) so how we are going to handle this.

With the new data structure of the conversation map where key is the combination of `Identity` and `zulipMessageId`. Now how we are going to compare the key (especially, `Identity`) when we receive remove FCM message event, so that we can remove it from conversation map.

cc @gnprice

gnprice:
Thanks @jainkuniya !

Quick comments on the Kotlin conversion:

* All these static methods on `NotificationHelper` would be perfect to move to package level -- i.e. kill `class NotificationHelper` and its companion object, and just say `fun addConversationToMap` etc. at the top level of the file.

  See docs: https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#package-level-functions

  To make things extra smooth, add `@file:JvmName("NotificationHelper")` at the top of the file -- then `NotificationHelper.addConversationToMap` etc. should all work just like before.

* It looks like the code in `NotificationHelper` got reordered? Please don't do that -- or rather, do it in a separate commit, either before or after. Keeping the code in the same order makes a big difference for comparing before and after, to see what changed.

* Some good opportunities for followup improvements are in the handling of `null`. This doc is worth reading early: https://kotlinlang.org/docs/reference/null-safety.html (Also applies to a couple of spots in the new code later in the PR.)

* Another cleanup opportunity: `extractName`.  See discussion of a similar line in 9fd7be59f, cleaned up in cbd6f8cde .

gnprice:
And early feedback on the new data structure (high-level because I see it's WIP):

* It's important that this is a map with a "conversation" as the key! That is the most important job of this data structure. :slightly_smiling_face: So let's not lose that.
  * The `String` key we have, using `buildKeyString`, is fine. That function can probably get a clearer name, though.
  * Better yet would be a nice algebraic data type -- i.e. something defined using the same pattern as our `Recipient` type. This should be a separate commit from other changes, though, like adding `Identity`.
* Also the `MutableList` is basically fine. A [`TreeMap`](https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html) would probably be an improvement -- though again, this should be a separate commit from other changes.
* So in short, something like `Map<Identity, Map<String, MutableList<MessageFcmMessage>>>` would be good.
  * It'd be good to make the change from `MessageInfo` to `MessageFcmMessage`, and the change adding `Identity`, as separate commits.
* Better yet might be `Map<Identity, Map<ConversationId, TreeMap<MessageFcmMessage>>>`, where `ConversationId` is an algebraic data type like I mentioned above. But that change isn't a priority.

Also:
* No need to start labelling things "legacy" while you're only changing the data structure -- we're going to use the same data structure regardless of what UI the platform version will give us the support to show. (So e.g. `buildKeyString` isn't "legacy"-specific.)
  * When you do start adding the new UI, "legacy" is a good name for the old one. Let's actually put the legacy-UI-specific code in its own separate file -- including especially `getNotificationBuilder`. Then that file can have "legacy" in its name.
  * (Also before making those changes, we'll want to convert FCMPushNotifications.java to Kotlin too.)

android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
gnprice: Yeah, this is something we should fix on the server. :-/ But of course we still have to handle the case where it's missing.

To do that, we can just iterate through all the keys in the map. (The top-level keys should be just `Identity` -- see more detailed comment I just made in this PR's main thread.) There won't be a ton of them. When the `RemoveFcmMessage` lacks an `email`, we'll just have to apply it to *every* identity that has the right server and realm. For almost all normal users, there will be just one of those anyway.

gnprice: Also, maybe you were planning to change this anyway (I know this is a WIP commit), but to be explicit:
* Let's never use an empty string (or zero, or empty list/array, etc.) as a symbol for "don't know". This is a notorious source of bugs. Instead, use an explicit "no data" marker -- in Kotlin or Java, that's `null`.
* This means you'll have to change the types! The type of `Identity#email` will go from `String` to `String?`. The fact you have to change this type is almost the same as the whole point :slightly_smiling_face: -- it means everywhere we use this field, we'll have to confront the possibility that it's `null`. Which is a good thing, because that is a real possibility!

gnprice:
Separately, a quick workflow tip in case it's helpful: `tools/test` these days will build the Android code and also run the Android unit tests. When making changes to this code, I definitely recommend regularly running the tests either that way, or in Android Studio (e.g. with Ctrl+Shift+F10).

(Maybe you already are; I know this is WIP, so it's fine that there are build failures. But I figured the specific ways to do it conveniently might not be obvious.)

gnprice:
Also, a small but important GitHub workflow point: please make sure in a PR to link to any relevant issues. I know we have one for the fancy notification UI. :slightly_smiling_face: Several, even, for different aspects of it.

This is important not only for finding the issue (and any previous discussion there) from the PR, but also for finding the PR from the issue.

android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
jainkuniya: cool! Submitted https://github.com/zulip/zulip/issues/11961 to track this on server. :)

jainkuniya:
> Also, a small but important GitHub workflow point:

Oh sorry I missed, now done.

>  Ctrl+Shift+F10

Doesn't seems to work for me on Mac, even tried with Cmd 🤔

gnprice:
> > Also, a small but important GitHub workflow point:
> Oh sorry I missed, now done.

thanks!


> I use `CMD + Shift + T` to find test of any function, then `Ctrl + Shift + R` to run them.

Cool, that sounds effective too!


> > Ctrl+Shift+F10
> 
> Which keymap? I use Mac OS X 10.5+ copy
> Doesn't seems to work for me on Mac, even tried with Cmd 

Looks like the name of this action is "Run Context Configuration". If you're looking at a test file and you right-click some code, you'll see it in the context menu.


> Anyway I am already using `tool/test` :)

Great :)

gnprice:
```
-        Log.v(TAG, "getPushNotification: " + data.toString(), new Throwable());
+        Log.v(NotificationHelper.getTAG(), "getPushNotification: " + data.toString(), new Throwable());
```

Quick tip on this small annoyance: the thing that solves this is `@JvmField`, which is nicely explained here:
https://developer.android.com/kotlin/interop#companion_constants

That whole page is probably worth quickly skimming (at least the heading and first paragraph or so of each section), so that you have an idea of what kinds of problems it offers solutions to. Then when you see one of those problems you can refer back to this page to read about the solution.

(Also, just yesterday I posted [a doc](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/android.md) with some advice on Kotlin and links to resources, and added a few more links today. That page is one of those links. :slightly_smiling_face: )

gnprice:
```
    android, notif: Override `equals` func of `Identitiy`.
    
    Currenlty for all events, server doesn't send `email` so it's better to
    avoid `Identity` comparison based on `id`.
```

Let's not do this -- this will be very confusing, and also it will make it hard to use `Identity` properly in the case where we do have the user's email.

Instead, in the exceptional case where we don't, we can just explicitly iterate through the keys and match only on the other fields.

gnprice:
(I'll hold off on doing a proper review until you say the word -- those are just things I saw in a quick skim. Thanks again!)

gnprice:
Thanks @jainkuniya for all your work on this!

In the first commit, converting a file to Kotlin, there's one thing I mentioned above which is really important, because it makes the change much harder to read and review:

> It looks like the code in NotificationHelper got reordered? Please don't do that -- or rather, do it in a separate commit, either before or after. Keeping the code in the same order makes a big difference for comparing before and after, to see what changed.

It looks like it's still reordered in this version, unfortunately. Would you please undo that? As I said in the initial review, this really makes a big difference.

(I'll go look through the subsequent commits too. I wanted to make this its own comment to help highlight it.)

gnprice:
On the remaining commits after that first one:

```
af1902f26 android, notifHelper, refactor: Replace multiple if with when.
b1f9070af android, notifHelper, refactor: Lift out try statement.
bb63088ac android, notif: Make conversation DS value type to `MessageFcmMessage`.
f0976b953 android, notif: Make `RemoveFcmMessage` public.
21fe91366 android, notif: Change signature of the `removeMessagesFromMap`.
```

* These refactors generally look good, thanks!

* I especially like the very smooth, one step at a time, change to `ConversationMap` in the third of these commits.

* Nit: it'd be a bit cleaner to put the commit that deletes `MessageInfo` right after the commit that removes the last places we use it.

```
-    try {
-        return URL(url, "?$query")
+    return try {
+        URL(url, "?$query")
```

* Nit: it's the `return` that is getting lifted out -- it's moving to the outer, or higher-level, part of the syntax tree. The `try` is getting pushed inward and deeper. So the commit message could say something like "Push `try` inside `return`", or "Lift `return` out from `try`".

```
d7cf318eb android, notif: Change signature of `buildNotificationContent`.
```

* Hmm, I don't think I quite understand the intention of this one. In particular:

```
-fun buildNotificationContent(conversations: ConversationMap, inboxStyle: Notification.InboxStyle, mContext: Context) {
+fun buildNotificationContent(conversations: ConversationMap, inboxStyle: Notification.InboxStyle, mContext: Context, identity: Identity?) {
+    if (identity == null) {
+        return
+    }
```

Why is this a change we want to make? It looks like the effect is just to ignore notifications that come from a server that doesn't specify the identity. I don't think that's what we want; and in any case it definitely requires more explanation in the commit message.


```
3f6ad17b3 android, notif: Change signature of `extractNames`.
5d040da93 android, notif: Change signature of `extractTotalMessagesCount`.
```

Same remark as the previous commit.


```
75fd7d395 android, notif: Change conversationMap key to `Identity`.
```

Ah, OK, I think I see what wrong turn was taken in the preceding few commits.

The new interfaces here feel a lot more complicated than they have to be for the job they're doing, and (as I mentioned above) the special cases they have for `identity == null` appear buggy. But I think I can suggest how to make these cleaner.

* First: several of these methods don't really want to take a `ConversationMap` and an `Identity?` -- they want to take the result of `getNarrowMapForIdentity(conversations, identity)`. After all, that's basically the first thing they do with those parameters.

  And that makes their interfaces (a) simpler, and (b) closer to the status quo, for a simpler change.

  So, let's give a name to that type. It's basically just the previous `ConversationMap`. Perhaps something like

```
class ByConversation : LinkedHashMap<String, MutableList<MessageFcmMessage>>()
class AllConversations : LinkedHashMap<Identity, ByConversation>()
```

I don't love those names, but they'll do; and I think that setup of two named types for the different layers is what we want.

* For a first commit in this direction, you might introduce the two types but have them still basically mean the same thing:

```
class ByConversation : LinkedHashMap<String, MutableList<MessageFcmMessage>>()
class AllConversations : ByConversation()
```

Then change each mention of `ConversationMap` to one or the other, to separate out which places should be which. The methods `addConversationToMap` and `removeMessagesFromMap` want `AllConversations`; the methods `buildNotificationContent`, `extractNames`, `extractTotalMessageCount` want `ByConversation`.

In fact, the big method `getNotificationBuilder` which calls several of those really just wants a `ByConversation`.

* Then change `AllConversations` to be a map with key `Identity`. Most of the code that changes in the current version of this PR shouldn't have to change. Some important aspects to think about and test:
  * What are the possible scenarios for `identity` being null, and for its fields being null? See comments in `extractIdentity`.
  * What is the user experience when actually getting notifications on two different identities? (E.g., from chat.zulip.org and from some test org on zulipchat.com.) In particular:
    * We want to show two independent notifications.
    * If you hit "clear" on one, or if you go read all the messages on one, it should only clear that one -- it shouldn't cause collateral damage on the other, unrelated notification.
    * In general, it should feel as cleanly separated as if running two separate copies of the Zulip app -- it shouldn't feel like one is interfering with or contaminating the other.
    * This probably requires a bit more work in `FCMPushNotifications.kt`, using the `NotificationManager` API. The calls to `cancelAll` will change, and the hard-coded `NOTIFICATION_ID` will have to become something that depends on `identity`.

* The parts mentioned just above involving `NotificationManager` -- those can probably be a separate commit, before actually changing `AllConversations`. That'll help keep different kinds of changes separated.


```
b07b60dfd android, notif, clean up: Delete MessageInfo class.
```

Cool! As mentioned above, it'd be a bit cleaner to move this to just after we stop using `MessageInfo`.

```
7a55dc593 android, notif, refactor: Convert FCMPushNotifications to kotlin.
```

* Cool! And I quite appreciate that this code is mostly kept in the same order :slightly_smiling_face: , at least for now. (If you later want to reorder it in some separate commits, that's fine.)

* I'd love to merge this commit already. Can you try rebasing it forward, to come before the changes to the `ConversationMap` data structure?
  * If it only involved a few manual cleanups after the auto-conversion, the easy way to "rebase" it might be to just re-do it.

* Some opportunities for good further cleanups, taking advantage of Kotlin (but definitely keep these in commits separate from this one):
  * Use more string interpolation, for things like `"msgid:$messageId"`, and also where `String.format` is used.
  * Inline a bunch of locals like `recipient`, now that their definitions are super concise.
  * The null-handling in `fetchAvatar` is confusing. Probably its caller should do the checking for null. Try to use a Kotlin idiom like `?.` to make that nice and concise -- see `FcmMessage.kt` for examples.

gnprice:
Thanks @jainkuniya for these revisions!

```
1e5ea1ae6 refactor, android: Convert `NotificationHelper` to Kotlin.
```

* This is a lot closer! Thanks for matching the order of the existing code :)

* I think the one remaining thing I'd like to see in this first commit is: in the commit message, mention what changes you made manually.

  That way when I'm reading it and looking at a given bit of the diff, I know whether it's something the auto-converter did or something you as a human chose to do. Which can go a long way in helping understand what the idea is.

  No need for a lot of justification - just a list of brief mentions is good, saying just enough so the reader can tell when they're looking at the bit of code it's about.

  Because I couldn't always tell what was a human choice vs. the way the auto-converter works, I actually went and ran the auto-converter myself so I could compare :wink: . There were a couple of subtle ones!

* Also I think there's at least one unintentional change: the commit drops the javadoc on ConversationMap.

* If you can add that list, then I think this change will be ready to merge!

```
34646c9b3 android, notifHelper, refactor: Replace multiple if with when.
cf2d06edb android, notifHelper, refactor: Lift `return` out from `try`.
```

* These look good, thanks! Will be ready to merge as soon as the preceding commit is.

```
2afed4250 android, FCMPushNotif, refactor: Convert FCMPushNotifications to kotlin.
```

* Looks good! This should be ready to merge as soon as the previous conversion is.

* I also appreciate that you fixed the one spot where the previous version of this commit had reordered the code :heart: 

* The commit message says this:

  > Simply use Android Studio plugin to convert and then small clean up so that functions/variable can be directly used in other classes.

  A good improvement would be to say what changes you had to make in that cleanup. I notice
  * `@file:JvmName`, and removing the outer `object FCMPushNotifications { ... }`
  * `@JvmField`, twice
  * ... I think those are the only ones I notice. It would be good to confirm that! Or to point out any others.

  Because those are all relatively boring, and because there are no changes at all to other files, it's not as essential to definitely list them as for the previous conversion. But it'd be nice.

```
6037cd3af android, FCMPushNotif, refactor: Use string interpolation.
fa590f3dd android, FCMPushNotif, refactor: Inline bunch of variable.
```

* These look good, thanks! Should be ready to merge as soon as the preceding commits are (so really just that one first conversion.)

* I think there are still a number of additional opportunities for each of these kinds of improvements. E.g. search for ` + `, and you'll find at least a couple more string concatenations that would be better as template interpolation. (Note that things like `toString` happen automatically in a template interpolation.)

```
103e5a046 android, FCMPushNotif, refactor: Use let for inline null check.

-        val avatar = fetchAvatar(sizedURL(context,
-            fcmMessage.sender.avatarURL, 64f))
-        if (avatar != null) {
-            builder.setLargeIcon(avatar)
-        }
+        val avatar = fetchAvatar(sizedURL(context, fcmMessage.sender.avatarURL, 64f))
+        avatar?.let { builder.setLargeIcon(avatar) }
```

* Cool, so -- `let` is actually even better than this :slightly_smiling_face: . The lambda expression (the `{ ... }`) gets the receiver of `let` as an argument -- so when you call `(/* some expression */).let { ... }`, you can refer to that whole receiver expression's value, in the body of the lambda, by saying `it`. See
  * docs on `it`: https://kotlinlang.org/docs/reference/lambdas.html#it-implicit-name-of-a-single-parameter
  * an article on Kotlin idioms I found just now via Google: https://phauer.com/2017/idiomatic-kotlin-best-practices/#consider-let
  * docs on `let` -- note the type of the argument, as `(T) -> R` rather than just `() -> R`, so it's a callback that takes an argument; and note where else `T` appears in that type: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/let.html

  So in this case, this becomes `avatar?.let { builder.setLargeIcon(it) }`.

gnprice:
Thanks for the update!

Just pushed the first 3 commits, converting `NotificationHelper`. I made some edits to the commit messages, particularly on the first commit (as 7d96c7f7d) -- take a look.

Now looking at the next few.

gnprice:
Comments on later commits:

```
-    val uri = Uri.fromParts("zulip", "msgid:" + Integer.toString(messageId), "")
+    val uri = Uri.fromParts("zulip", "msgid:${Integer.toString(messageId)}", "")
```

The `Integer.toString` here is redundant, which makes it confusing. (I mentioned above that it's redundant.)

In general, `toString` methods already get called on whatever you put into a Kotlin string template.

```
+        val avatar = fetchAvatar(sizedURL(context, fcmMessage.sender.avatarURL, 64f))
+        avatar?.let { builder.setLargeIcon(it) }
```

Cool, this is helpful.

There are a few more things that can be done here to handle the `null` case more cleanly. I've gone and added new commits to the branch to do those -- please take a look.

---

And I've merged those next few commits! These correspond to all the "focus" commits you identified above. The results are da08ccba3..909ad1629 .

gnprice:
Quick comments on the remaining commits (which I know are a draft):
* Careful with that search-and-replace :wink:  -- you have "allConversations" in some user-facing text.
* Also, you're changing a lot of things to "allConversations" that don't actually want "all" conversations, which means you just have to change them again later. And at the end of the branch a lot of things are still "allConversations" that in the end should be just `conversations: ByConversation`.

jainkuniya:
> The Integer.toString here is redundant, which makes it confusing. (I mentioned above that it's redundant.)

Oh, I missed it :(

gnprice:
Thanks @jainkuniya !

These changes all look good -- about to merge. A few comments and things I adjusted:
* The three similar one-line changes from `ConversationMap` to `ByConversationMap` are easiest to understand together, so I squashed them.

* The last two commits, replacing the only use of `MessageInfo` and then deleting it, are also easiest to understand together, so I squashed them. In particular, the commit that replaced the use of it had this in the commit message:

  > Where `MessageInfo` only consists of message id and content.

  but once the same commit also deletes `MessageInfo`, that information is self-evident because it's right there in the diff.

  (In both of these cases, it helps that the commits are really small. If they were more complex, there'd be more value in breaking them up.)

* In general, basically any data type should have a doc comment saying what it means. (For example, see the classes in `FcmMessage.kt`.) We have one of those for the old `ConversationMap` (new `ByConversationMap`); we should keep it accurate any time we make changes, and add a comment for any new data type we add.

  So, the commit renaming `ConversationMap` to `ByConversationMap` and adding a new `ConversationMap` should also add a comment for the new type, and should update the old one as needed.

  In your commit message, there is some prime material for these doc comments!

  > Idea of `ByConversations` is to map messages to their respective narrow
    of that particular realm (identity).
    Now in the next step `Conversations` will start mapping from
    this `ByConversations` to `Identity`.

  I went and added appropriate comments as an additional commit. I think some of your upcoming changes will merit updates to those comments, so please take a stab at them when you do.

Issue #3384, opened 2019-03-05
Correctly display user's status when 'unavailable'


borisyankov:
We support the user setting themselves to unavailable, but we don't yet correctly display when a user is unavailable.

This means two things:
* The icon should be the one we've chosen and now have on the web (rather than a gray disk.)
* The user shouldn't be described as "Active now", or "Active 15 minutes ago" etc.

More discussion here:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/user.20status/

[edited by @gnprice ; corresponding original text below]

---

 * introduce the new 'Unavailable' status and do not treat it as 'Offline'
 * Render that new state exactly the same as on the web (a bit more tricky than the purely css solution)
 * Change the text for 'last active' if 'unavailable'

[See also #3406 for another followup previously included here.]

gnprice:
Thanks @borisyankov for filing this!

I'm going to narrow this issue down to the most critical/urgent points, the better to track getting those taken care of; I'll file a separate issue for the rest.

gnprice:
This was implemented by #3384.

PR #3417, opened 2019-03-19
Correctly display user's status when 'unavailable'

borisyankov:
Implements #3384

* introduce the new 'Unavailable' status and do not treat it the same as 'Offline'
* render that new state exactly the same as on the web (a bit more tricky than the purely css solution)
* change the text for 'last active' if 'unavailable'

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -57,7 +57,7 @@ class AccountDetails extends PureComponent<Props, void> {
             />
           )}
           <View>
-            <ActivityText style={styles.largerText} email={user.email} />
+            <ActivityText style={styles.largerText} user={user} />
           </View>
           {user.timezone ? (
             <View>
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -84,9 +84,8 @@ export type UserGroup = {|
 /**
  * Specifies user status related properties
  * @prop away - present if we are to override user's presence status
- *       * when `true` the user is always `offline`
- *       * when missing the presence is not changed
- *       * can not be `false`
+ *   * This is the "user status" / "unavailable" feature added in early 2019.
+ *     (At time of writing, there are no docs to link to.)
  * @prop status_text - a string representing information the user decided to
  *   manually set as his 'current status'
  */
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -87,7 +87,7 @@ export type UserGroup = {|
  *   * This is the "user status" / "unavailable" feature added in early 2019.
  *     (At time of writing, there are no docs to link to.)
  * @prop status_text - a string representing information the user decided to
- *   manually set as his 'current status'
+ *   manually set as their 'current status'
  */
 export type UserStatus = {|
   away?: true,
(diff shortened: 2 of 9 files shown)


gnprice:
Thanks @borisyankov !

#### 60af8d0dc Add a discrete 'unavailable' presence status

* We have a bit of text you wrote that says what `userStatus.away` means, right? That text was pretty specific about the exact meaning, so it won't be true anymore with this change. That should get fixed.

  Looks like that's on `UserStatus`, in `modelTypes`.

  I'd like to say that comment should link to some existing docs for the details, like the `ClientPresence` jsdoc links to the presence subsystem doc. ... Unfortunately AFAICT we have zero documentation of this system either in developer docs, user docs, or API docs -- e.g. https://zulipchat.com/help/check-whether-someone-is-here-or-away (and the corresponding file in master) still describes the old system without this "unavailable" concept.

```
+): PresenceStatus | 'unavailable' =>
```

* This line feels like a bit of a code smell to me. OTOH we have the type-checker to propagate this type through, so it's fairly harmless. I think if we started repeating this union a few times, I'd start definitely wanting to give it a name.

#### dabe47202 PresenceStatusIndicator: Rework to match web visuals

* I like the use of tiny within-file helper components! I think that works really well here. (Perhaps predictable that I would like it :slightly_smiling_face: -- you probably read cb418f134 last week.)

* Nit: I'd rather name them to uniformly match the `status` values they correspond to -- so s/Online/Active/ to match `active`.

* Nit: It's slightly confusing that the definitions and the `switch` statement come in different orders.


#### 0845e4136 ActivityText: Remove 'color' property

Good cleanup, thanks!


#### e7e24a237 ActivityText: Take a 'user' prop instead of 'email'

Cool, good plan!


#### 4b71c6db0 presenceToHumanTime: Consider user status if given

* You mentioned in chat that this "today" logic doesn't quite make sense -- it'll describe something 20 hours ago as "today", even if it's definitely not "today" but "yesterday" in normal terms.

  Is this just matching the webapp? If so, definitely worth mentioning, to make clear why we're choosing this somewhat odd behavior -- probably a line or two of comment on the `return 'today';` would be the right place.


#### 93e327503 ActivityText: Pass 'userStatus' to 'presenceToHumanTime'

* Hmm, it looks like this fixes a type error introduced by the last commit. We shouldn't break things at any step -- it makes the changes harder to follow, as well as to investigate with things like `git bisect`.

  It looks like you've written the implementation of this function to make `status` optional. So one perfectly adequate fix would be to make that explicit in the type.

---

Playing with the UI, a couple of small things I notice -- not blockers but would be good followups:

* The status dot isn't aligned well on the full-screen profile:
  image

* The "unavailable" icon lacks contrast on top of an avatar:
  image

  * Possible fix: give it a partly-opaque background? Or even fully opaque.

  * Something similar would probably benefit the other icons too -- if someone's avatar happens to have green or similar bright colors in the lower right, even the "active" color might not stand out. Perhaps a contrasting (white?) border would be good.

gnprice:
Merged! Because it's past the end of your day and we want to move quickly on this, I just went and made the above fixes myself.

The followup items in the UI are still open, though. Would you take those on next?

Also though I added a note about the misleading "today" logic, I didn't actually change it.

Thanks again!

gnprice:
(I happened to notice those two followups in the review comment when I found this old PR while populating the "a-avatar" label I just created. Doesn't look like we ever either handled them or filed them, so I've gone and filed them now as #4306 and #4307 .)

PR #3418, opened 2019-03-19
messageHeaderAsHtml: Modify topic header to include date.

### LINKED ISSUE #1336: It's really hard to figure out the date of the messages you're looking at
When you're looking at a set of messages, it can be really hard to figure out what date they were sent at.  While I think this problem is much more annoying because of our scroll position issues which can often lead to looking at messages from a few days ago seemingly randomly, I think it's still relevant regardless.

I wonder if we should in the mobile app, start displaying timestamps in the sender rows that are not for today in the format of e.g. "Jun 5 15:50".  It feels like we have the space, and I don't see a better way to surface this information.  Since most of the time folks will be reading messages from today, there shouldn't be much added clutter from doing this.

(Though I suppose the problem here is actually that stick headers aren't working, since those are where we had the date before?  So fixing that could be a better tactical solution)

ishammahajan:
Many times when people are scrolling through messages they need to see 
what date it is. This modification is made for this use-case. Users can 
simply see the header to determine the date a particular message was sent.

This is implemented using some basic `css` and `js`. When writing the css 
selector for the header of the topic, there was an issue arising where the 
sticky header (because a small part of it gets cut out) was showing 
misalignment. This is the reason why `margin-top: 0.2em` was added to 
`.topic-header` selector.

Tests have been conducted, and behaviour is checked for both Android and 
iOS, light and dark mode.

Fixes #1336

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -76,6 +76,10 @@ hr {
   justify-content: space-between;
   margin-bottom: 0.25em;
 }
+.topic-header {
+  margin-top: 0.2em;
+  margin-bottom: 0.00em;
+}
 .timerow {
   text-align: center;
   color: #999;
@@ -99,6 +103,11 @@ hr {
   color: #999;
   font-size: 0.9em;
   white-space: nowrap;
+}
+.topic-timestamp {
+  font-size: 0.9em;
+  white-space: nowrap;
+  margin-right: 0.5em;
 }
 .message {
   display: flex;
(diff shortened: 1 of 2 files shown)


gnprice:
Hi @ishammahajan -- thanks for this contribution!

* First quick bit of feedback: for a change like this one that's aimed at the UI, please include screenshots in the PR. They can help a lot in making it quicker for people to review.

* For this design overall, I'm not yet convinced. In particular, these won't show up at all when narrowed to a single topic, or to a PM (1:1 or group) conversation. So it doesn't really fix #1336 -- we need something that applies to those too.

* Some specific bits that would be important to fix if we did go with a version of this:
  * Should say "Today" for today; in general, should have the same logic as the date dividers.
  * It isn't aligned with the topic name.
  * Touching a topic header (to narrow to the topic) no longer works.

gnprice:
Closing this PR in favor of continuing the overall design discussion in #1336 . Thanks again!

Issue #3372, opened 2019-02-27
Allow updating own profile 


borisyankov:
Implement the functionality that exists on the web at:
https://chat.zulip.org/#settings/your-account

More important parts:
* Email, Full name, Password
* Upload new avatar

Secondary:
* Deactivate account
* Custom profile fields

anh2111htd:
@zulipbot claim

AdityaHirapara:
I would like to work on this issue

AdityaHirapara:
I am thinking of implementing it on profile tab like this:
![Screenshot_2019-03-18-19-20-50-35](https://user-images.githubusercontent.com/27485533/54538534-6f350080-49ba-11e9-9241-154c8ea40480.png)
Then pressing button will navigate to new screen like user status.

@borisyankov any suggestions?

borisyankov:
@AdityaHirapara we have some problems with vertical space here - some phone screens might not be high enough to fit this. I have an idea for a different style of 'Edit profile' button, but it is perfectly fine to start working on this with this implementation and we can fix that before merging it.

timabbott:
We probably don't need to show local time for your own profile, so there may be some space optimization options available too.

AdityaHirapara:
> We probably don't need to show local time for your own profile, so there may be some space optimization options available too.

Yes, That can be done.

borisyankov:
We probably need this information somewhere though.
If there is a timezone set it is not necessarily what the actual time of the device is.

Also, I haven't had set this for myself so my concern for the space was even without that line.

anh2111htd:
Hi @AdityaHirapara, I've already been working on this issue :) I claimed 2 weeks ago but didn't update frequently as it took a while to learn the codebase and debug.

For now, user can edit their own full name, I'm working on allowing them to edit avatar and password. @borisyankov can you help me do a quick review for #3421 as it may relate to your PR #3403?

AdityaHirapara:
Oh okay, Let me know if I can help you with anything :)

timabbott:
We should be sure to disable the widget when changing a user's avatar is disabled (see zulip/zulip#12251) because the organization is syncing them from LDAP.

borisyankov:
@timabbott should we introduce a `capabilities` or `permissions` object with flags on `register`?
The purpose will be to move the biz logic of determining what the client can do to the back-end.

So stuff like 'if ldap => canEditProfile: false`, `if admin => canDeleteStream: false` etc.

[...]

PR #3421, opened 2019-03-20
[WIP] Issue #3372. settings: Allow updating own profile.

anh2111htd:
Screenshot for the account screen:

image

Screenshot for the edit screen, currently users can only edit full name, I'm adding the option to edit avatar and password.

image

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -51,6 +51,9 @@ class AccountDetails extends PureComponent<Props, void> {
             <PresenceStatusIndicator email={user.email} hideIfOffline={false} />
             <RawLabel style={[styles.largerText, styles.halfMarginLeft]} text={user.email} />
           </View>
+          <View style={componentStyles.statusWrapper}>
+            <RawLabel style={[styles.largerText, styles.halfMarginLeft]} text={user.full_name} />
+          </View>
           {userStatusText !== undefined && (
             <RawLabel
               style={[styles.largerText, componentStyles.statusText]}
--- a/src/account-info/ProfileCard.js
+++ b/src/account-info/ProfileCard.js
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 import type { Dispatch, GlobalState, User } from '../types';
 import { getSelfUserDetail } from '../selectors';
 import { ZulipButton } from '../common';
-import { navigateToUserStatus } from '../actions';
+import { navigateToUserStatus, navigateToEditProfile } from '../actions';
 import AccountDetails from './AccountDetails';
 import AwayStatusSwitch from './AwayStatusSwitch';
 import SwitchAccountButton from './SwitchAccountButton';
@@ -20,6 +20,10 @@ const componentStyles = StyleSheet.create({
   },
   setStatusButton: {
     marginHorizontal: 16,
+  },
+  editProfileButton: {
+    marginHorizontal: 16,
+    marginTop: 8,
   },
 });
 
@@ -40,6 +44,11 @@ class ProfileCard extends PureComponent<Props> {
     dispatch(navigateToUserStatus());
   };
 
+  handleEditProfile = () => {
+    const { dispatch } = this.props;
+    dispatch(navigateToEditProfile());
+  };
+
   render() {
     const { selfUserDetail } = this.props;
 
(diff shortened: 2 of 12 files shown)


gnprice:
Thanks @anh2111htd !

#### 128461369 settings: Display full name in Account Details.

* Ah, thanks for adding the user's name! That probably should have been there a long time ago. :slightly_smiling_face: 

* This screen is getting a bit busy with so many pieces of information as rows of text -- email, name, status text, activity, local time. Also the email is oddly over-emphasized.

  I think I'll go ahead and merge this first small commit (adding the name), then tweak to improve that situation a bit. There's probably more we can do there, though.

#### 81f89591f settings: Add edit profile screen to edit full name.

* As @borisyankov mentioned on #3372 , the new button exacerbates the vertical-space shortage too. Boris, I'd be interested to hear suggestions for how we could arrange that better.

```
 type EventUserUpdateAction = {|
   type: typeof EVENT_USER_UPDATE,
-  // In reality there's more -- but this will prevent accidentally using
-  // the type before going and adding those other properties here properly.
+  email: string,
+  newAccountDetails: {| full_name: string |},
 |};
```

* Cool, this is good to add. A few important general points:
  * Is this in fact the complete contents of these events? (I kind of doubt it, with just `full_name`.) Whenever we're not telling the full story, that needs to be explicit.
  * Changes like this that handle additional data from the server should be in separate commits from UI changes. They're pretty independent, they have different things to think about in writing and reviewing them, and so they should be separated so they can be read separately and can each have their own commit messages.
  * And more specifically -- as you mentioned on #3372, this will be covered by #3403 . :slightly_smiling_face:  Fine to include those changes in order to make progress on this issue, but it should be explicit -- e.g. mention that other PR in this one.

```
+class EditProfileScreen extends PureComponent<Props> {
```

* So I know this is following a pattern we have elsewhere in the code, but just since last week we've started to shift away from it -- see cb418f134, then 6bc623c3d, 7f7620811, and more recently Boris's d5328f5d0.

  Basically it's this:

  > In a lot of places we have some tiny helper components that exist entirely to serve a specific parent component.  Most of those would be a lot easier to read, understand, and maintain if the helpers were in the same file where they're used -- less boilerplate code; less need to cross-reference back and forth between files; and most of all, the helper component typically implicitly assumes things about the context it's used in, and that context would be explicit and ready at hand.

  except that in this example, it's the parent that is the tiny helper, rather than the child.

  So let's combine these into one file. `EditProfileScreen.js` would be a fine name -- the only thing exported will be what your `EditProfileScreen.js` exports now.

  See `ProfileCard.js` for an example of some of the style for this, like how to write the props type.

```
+  updateDetailsChanged = () => {
+    const { selfUserDetail } = this.props;
+    const { fullName } = this.state;
+    const hasDetailsChanged = fullName !== selfUserDetail.full_name;
+
+    this.setState({ hasDetailsChanged });
+  };
```

* Hmm, I don't love handling this with an additional state property -- it makes the control flow more complex, and raises the possibility of an inconsistent state where `fullName` has in fact changed but `hasDetailsChanged` is still false. In fact: it *guarantees* that we will pass through that inconsistent state! So at best it's going to be complicated to reason through why this works correctly, if it does.

  * Instead, how about just making a function with a name like `hasChangedData`, which looks at the props and state and returns a boolean? Then `render` will call that function.

---

(There are some more bits I haven't reviewed, which I think will be covered by #3403 instead.)

gnprice:
```
 type EventUserUpdateAction = {|
   type: typeof EVENT_USER_UPDATE,
-  // In reality there's more -- but this will prevent accidentally using
-  // the type before going and adding those other properties here properly.
+  email: string,
+  newAccountDetails: {| full_name: string |},
 |};
```

Oh wow, no, now I see what this is doing -- you're actually writing a new action, internal to the app, which you just invented.

This is not at all the right place for that, or the right name. Take a look at `eventToAction.js` (which you edited in this commit) to see what all the other `EVENT_*` action types are for. They represent events sent by the server. (See [server-side docs](https://zulip.readthedocs.io/en/latest/subsystems/events-system.html), or [a client-oriented doc](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/realtime.md).)

The bit where you deleted some lines from `eventToAction.js` would be a good moment to ask yourself the question: what is this codepath for? Is there a reason it makes sense to replace it with this new thing?

anh2111htd:
Thanks @gnprice for reviewing. As you suggested, there are parts covered by PR #3403, so I will put this on hold and continue once #3403 has been merged.

Issue #3387, opened 2019-03-07
Loading indicator on message list when data is stale


gnprice:
Similar to #3025, but for the message list. This is an especially important UI surface for this situation because users often start the app by opening a notification, which navigates straight to the message list past the unreads screen.

One simple, good design for an indicator would be a banner a lot like our `OfflineNotice`, but instead of "No Internet connection", saying something like "Connecting to server...". That's a common design among other messaging apps.

thedeveloperr:
New contributor to zulip mobile. Have only worked on zulip backend. Is this a first good issue for new contributor?

anh2111htd:
@gnprice I notice this animation when loading messages:

![image](https://user-images.githubusercontent.com/19589970/55006391-9f168080-5018-11e9-8b15-d478f9d82faa.png)

It indicates the list is loading to me. In `MessageList`, by default,  `showMessagePlaceholders` is true, and this animation will show whenever `isFetching && noMessages`. For this issue, should I use the same check used for this animation and add a banner to make the indication clearer?

anh2111htd:
@gnprice Fixed this in #3430, please review.

chrisbobbe:
Hmm, once https://github.com/zulip/zulip-mobile/issues/3802 is fixed (which will probably be soon), will this still be a shortcoming in handling the loading state, or will it maybe be time to close this rather old issue and its PR?

ray-kraesig:
I don't believe a fix for #3802 would help with this; the user would still be looking at old state for some time after clicking the notification.

chrisbobbe:
> the user would still be looking at old state for some time after clicking the notification.

Mmm, right! The proposed fix for #3802 eliminates the "No messages" message during the period when we are re-registering, by showing the stale data, but it still doesn't indicate that the messages are stale. I like the idea of a banner for this, as mentioned here, since it's still helpful to see even stale messages when possible.

gnprice:
> If I've understood the scope of this issue, this particular loading period (during, and only during, the /register request) is global and should be shown no matter what screen you're looking at (similar to the "No Internet connection" banner mentioned above). Actually, if this is correct, I believe #3025 may be identical to this issue.

Hmm, yeah, this may be the best way to do it. Certainly there's a long tail of screens that show data from the server, and that should have a loading indicator when our data from the server is stale; and it'd be best not to try to handle them one by one.

There are a few screens that shouldn't have this banner, because they aren't about any one server.
[...]

PR #3430, opened 2019-03-28
Loading indicator on message list when data is stale

anh2111htd:
Issue #3387. The check if data is stale is `const hasCaughtUp = caughtUp.newer;` where `caughtUp` is  `getCaughtUpForNarrow(state, props.narrow)`. Based on the use case describe in #3387 , I understand we want to indicate to the users if they have had all new message (when they open from a notification for example), so I check for `caughtUp.newer` here. Let me know if this check is not sufficient or incorrect.

Screenshot:
image

### DIFF
--- a/src/chat/Chat.js
+++ b/src/chat/Chat.js
@@ -9,6 +9,7 @@ import MessageList from '../webview/MessageList';
 import NoMessages from '../message/NoMessages';
 import ComposeBox from '../compose/ComposeBox';
 import UnreadNotice from './UnreadNotice';
+import LoadingNotice from './LoadingNotice';
 import styles from '../styles';
 import { canSendToNarrow } from '../utils/narrow';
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Hi @anh2111htd, thanks for your work on this, and I'm sorry we didn't get around to reviewing this — not your fault! 🙂

I'm closing this, as https://github.com/zulip/zulip-mobile/pull/3901 also shows the notice on the unreads screen (far left on the tabs at the bottom; an "inbox" icon), which fixes https://github.com/zulip/zulip-mobile/issues/3025, and several other screens that use data from the event system. Also, the loading state we're concerned with here is stored in `session.loading`, not `state.caughtUp`; you can read the JSDoc about `state.caughtUp` in reduxTypes.js. 🙂

Issue #3406, opened 2019-03-16
Translate status text suggestions


gnprice:
This is one of the followups to #3344. For details on this and other followups, see this comment:
https://github.com/zulip/zulip-mobile/pull/3344#issuecomment-468465263

anh2111htd:
@gnprice Fixed in PR #3432, please review.

PR #3432, opened 2019-03-29
Translate status text suggestions 

anh2111htd:
Issue #3406.

### DIFF
--- a/src/user-status/UserStatusPicker.js
+++ b/src/user-status/UserStatusPicker.js
@@ -1,0 +1,43 @@
+/* @flow strict-local */
+import React, { PureComponent } from 'react';
+import { FlatList } from 'react-native';
+
+import type { GetText } from '../types';
+import { TranslationContext } from '../boot/TranslationProvider';
+import { OptionButton } from '../common';
+import statusSuggestions from './userStatusTextSuggestions';
+
+type Props = {|
+  onPress: (status: string) => void,
+|};
+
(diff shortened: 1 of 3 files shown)


src/user-status/UserStatusPicker.js
```diff
@@ -0,0 +1,43 @@
...
+
+type Props = {|
+  onPress: (status: string) => void,
+|};
+
+export default class UserStatusPicker extends PureComponent<Props> {
+  static contextType = TranslationContext;
+  context: GetText;
+
+  getTranslatedSuggestions = () =>
```
jainkuniya: Option button already uses `Label` so there won't be any problem. https://github.com/zulip/zulip-mobile/blob/master/src/common/OptionButton.js

It will get translated automatically, just u need to put them in message_en.json

anh2111htd: @jainkuniya I tried it, and while the option text shown on the screen is translated, the value of this option in `UserStatusScreen` is still in English, so when we send an update request to Zulip, the status is still in English.

chrisbobbe:
Thanks for the PR, @anh2111htd! It looks like the status text suggestions were already added to messages_en.json, in 640d4414282606c489573a389e1578730f87a3bf. I also see some changes to the UI components in that area, but we've gone in a different direction with that (see #4683).

For those reasons, I'm closing this. Thanks again for your work, though!

PR #3434, opened 2019-03-29
webview js: Improve `longPress` detection logic.

ishammahajan:
The existing `longPress` logic is very convoluted. It makes the user
study the code for a long amount of time to follow the state machine
and figure out exactly how the logic works. There are also a few
inconsistencies with respect to how the code handles events, in
particular `lastTouchEventTimestamp = Date.now()` in the `touchend`
listener, and a few more inconsistencies.

This commit aims to fix these inconsistencies by removing the
timestamp factor from `lastTouchEventTimestamp` altogether. It (the
replaced name being `touchActive`) acts as a flag for if the user is
touching the screen at present, which is an accurate indicator for
if the user wants to `longPress` (when the handler for longPress
triggers he would still be touching the screen if he wants to
longPress.)

Tested on Android and iOS emulators and a physical Android device.

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -230,13 +230,13 @@ var sendScrollMessageIfListShort = function sendScrollMessageIfListShort() {
 };
 
 var scrollEventsDisabled = true;
-var lastTouchEventTimestamp = 0;
 var hasLongPressed = false;
+var longPressTimeout;
 var lastTouchPositionX = -1;
 var lastTouchPositionY = -1;
 
 var handleScrollEvent = function handleScrollEvent() {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 
   if (scrollEventsDisabled) {
     return;
@@ -421,7 +421,7 @@ var requireAttribute = function requireAttribute(e, name) {
 
 documentBody.addEventListener('click', function (e) {
   e.preventDefault();
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 
   if (hasLongPressed) {
     hasLongPressed = false;
@@ -497,11 +497,6 @@ documentBody.addEventListener('click', function (e) {
 });
 
 var handleLongPress = function handleLongPress(target) {
-  if (!lastTouchEventTimestamp || Date.now() - lastTouchEventTimestamp < 500) {
-    return;
-  }
-
-  lastTouchEventTimestamp = 0;
   hasLongPressed = true;
   sendMessage({
     type: 'longPress',
@@ -520,9 +515,9 @@ documentBody.addEventListener('touchstart', function (e) {
 
   lastTouchPositionX = e.changedTouches[0].pageX;
   lastTouchPositionY = e.changedTouches[0].pageY;
-  lastTouchEventTimestamp = Date.now();
   hasLongPressed = false;
-  setTimeout(function () {
+  clearTimeout(longPressTimeout);
+  longPressTimeout = setTimeout(function () {
     return handleLongPress(target);
   }, 500);
 });
@@ -537,13 +532,13 @@ var isNearPositions = function isNearPositions() {
 
 documentBody.addEventListener('touchend', function (e) {
   if (isNearPositions(lastTouchPositionX, lastTouchPositionY, e.changedTouches[0].pageX, e.changedTouches[0].pageY)) {
-    lastTouchEventTimestamp = Date.now();
+    clearTimeout(longPressTimeout);
   }
 });
 documentBody.addEventListener('touchmove', function (e) {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 });
 documentBody.addEventListener('drag', function (e) {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 });
 `;
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -535,6 +535,9 @@ documentBody.addEventListener('touchend', function (e) {
     clearTimeout(longPressTimeout);
   }
 });
+documentBody.addEventListener('touchcancel', function (e) {
+  clearTimeout(longPressTimeout);
+});
 documentBody.addEventListener('touchmove', function (e) {
   clearTimeout(longPressTimeout);
 });
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -350,13 +350,13 @@ const sendScrollMessageIfListShort = () => {
  */
 let scrollEventsDisabled = true;
 
-let lastTouchEventTimestamp = 0;
 let hasLongPressed = false;
+let longPressTimeout;
 let lastTouchPositionX = -1;
 let lastTouchPositionY = -1;
 
 const handleScrollEvent = () => {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
   if (scrollEventsDisabled) {
     return;
   }
@@ -557,7 +557,7 @@ const requireAttribute = (e: Element, name: string): string => {
 
 documentBody.addEventListener('click', (e: MouseEvent) => {
   e.preventDefault();
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 
   /* Without a flag `hasLongPressed`, both the short press and the long
    * press actions get triggered. See PR #3404 for more context. */
@@ -641,23 +641,11 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
 });
 
 const handleLongPress = (target: Element) => {
-  // The logic that defines a "long press" in terms of raw touch events
-  // is pretty subtle.  The `lastTouchEventTimestamp` and surrounding logic
-  // are an attempt to define a long press.
-  //
-  // TODO: The logic around this "timestamp" is a bit obscure; it's
-  // sometimes a real timestamp, other times 0 as sort of a boolean flag.
-  // It would be good to clean it up to be clearer.
-  //
-  // At the same time, the logic is believed not to cover all the cases it
-  // should; for example, multi-touch events.  Better would be to either find
-  // a library we can use which strives to handle all that complexity, or
+  // The logic is believed not to cover all the cases it should; for
+  // example, multi-touch events. Better would be to either find a
+  // library we can use which strives to handle all that complexity, or
   // get long-press events from the platform.
-  if (!lastTouchEventTimestamp || Date.now() - lastTouchEventTimestamp < 500) {
-    return;
-  }
 
-  lastTouchEventTimestamp = 0;
   hasLongPressed = true;
 
   sendMessage({
@@ -676,9 +664,9 @@ documentBody.addEventListener('touchstart', (e: TouchEvent) => {
 
   lastTouchPositionX = e.changedTouches[0].pageX;
   lastTouchPositionY = e.changedTouches[0].pageY;
-  lastTouchEventTimestamp = Date.now();
   hasLongPressed = false;
-  setTimeout(() => handleLongPress(target), 500);
+  clearTimeout(longPressTimeout);
+  longPressTimeout = setTimeout(() => handleLongPress(target), 500);
 });
 
 const isNearPositions = (x1: number = 0, y1: number = 0, x2: number = 0, y2: number = 0): boolean =>
@@ -693,14 +681,14 @@ documentBody.addEventListener('touchend', (e: TouchEvent) => {
       e.changedTouches[0].pageY,
     )
   ) {
-    lastTouchEventTimestamp = Date.now();
+    clearTimeout(longPressTimeout);
   }
 });
 
 documentBody.addEventListener('touchmove', (e: TouchEvent) => {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 });
 
 documentBody.addEventListener('drag', (e: DragEvent) => {
-  lastTouchEventTimestamp = 0;
+  clearTimeout(longPressTimeout);
 });
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -685,6 +685,10 @@ documentBody.addEventListener('touchend', (e: TouchEvent) => {
   }
 });
 
+documentBody.addEventListener('touchcancel', (e: TouchEvent) => {
+  clearTimeout(longPressTimeout);
+});
+
 documentBody.addEventListener('touchmove', (e: TouchEvent) => {
   clearTimeout(longPressTimeout);
 });


src/webview/js/generatedEs3.js
```diff
@@ -537,13 +537,13 @@ var isNearPositions = function isNearPositions() {
 
 documentBody.addEventListener('touchend', function (e) {
   if (isNearPositions(lastTouchPositionX, lastTouchPositionY, e.changedTouches[0].pageX, e.changedTouches[0].pageY)) {
-    lastTouchEventTimestamp = Date.now();
+    touchActive = false;
```
ishammahajan: The `Date.now()` here wasn't required, as the `!lastTouchEventTimestamp` would capture the cases it covered anyways if it was set to `0` instead.

src/webview/js/generatedEs3.js
```diff
@@ -520,7 +520,7 @@ documentBody.addEventListener('touchstart', function (e) {
 
   lastTouchPositionX = e.changedTouches[0].pageX;
   lastTouchPositionY = e.changedTouches[0].pageY;
-  lastTouchEventTimestamp = Date.now();
+  touchActive = true;
```
ishammahajan: Like below, setting `lastTouchEventTimestamp` to `1` here would have the same effect as before since the handler only gets triggered after 500ms in any case.

ishammahajan:
@gnprice ready for review! 🙂

gnprice:
Thanks @ishammahajan !

So, the thing that's missing in this commit is to make it easy for the reader to see why the new logic has the same behavior as the old. Actually, it isn't totally clear even that it's *meant* to have the same behavior -- I know that's the intent from our discussion on the #mobile stream, but just looking at the commit it looks like it's probably meant to change the behavior somehow.

And then the reader sees that one `lastTouchEventTimestamp = Date.now()` turns into `touchActive = false` while the other turns into `touchActive = true`, and it sure looks like the behavior is changing -- and the reader may spend quite a while puzzling over exactly what changed, and why we believed that the change was a good one, before they realize that despite appearances it didn't change after all. So what I'd like to do is avoid setting puzzles like that for the next person studying this. :wink: 

Some of the needed explanation is in the comments you made above on the PR -- so that information just needs to go into a commit message.

I think if you try adding that explanation in the commit message, you'll also find that another step that makes it much clearer is to break the change into two commits:
* One just changes one `Date.now()` to 0 -- it's a one-line change, with a commit message explaining why it doesn't change the behavior.
  * Possibly also cuts out the `Date.now() - lastTouchEventTimestamp < 500` condition, since that's closely related.
* Then another commit replaces `lastTouchEventTimestamp` with the boolean `touchActive`.

One hint of this is that your explanations above already discuss a hypothetical intermediate state. :slightly_smiling_face: So the reader is basically expected to have that intermediate state in mind in order to understand the explanation of why the change makes sense... so it's clearer to just actually write the intermediate state down explicitly.

ishammahajan:
@gnprice I agree with your assessment! The intermediate state should certainly be explained so that it becomes clear to the reader.

I have changed the commits as per request, and added in the contents of the comments into the commits :). I initially thought the comments were part of the commits and were considered appropriate places to explain what exactly the changes were supposed to mean, but it is correct that a casual reader of the commits must also be able to understand the commits from the content of the commits itself rather than having to go to check on the changes explicitly and perhaps finding comments explaining the change there.

Ready for review!

ishammahajan:
So I was using the app on this build, and found out that this logic doesn't actually work as intended. Since the timeout actually runs all the time without being cancelled and the flag inside `handleLongPress` actually does the handling of control, if we repeatedly click on one message on this commit, the js registers the clicks as one long click if the finger happened to be pressed at the 500th millisecond of the first click, since the handling of timeout is not able to recognise this.

ishammahajan:
@gnprice The commit details what has been done to fix the bug.

I was confused as to what the order of the commits should be, or if I should discard the commit prior to this altogether. Came to the conclusion that all three commits are required to clear the confusion about changing this piece of logic.

I also wasn't sure about changing the comments inside `handleLongPress`. It would be great if you could have a look at them too.

This has been tested for both platforms and is ready for review :)

gnprice:
Thanks @ishammahajan ! And glad you were testing it. :-)

I'm actually not 100% sure I understand your new findings -- I think you're saying that the previous version of this PR (which is now the two commits fd19c90f9 "webview js: Refactor logic to make it more readable." and 3b2d2f45d "webview js: Improve `longPress` detection logic.") turns out to introduce a bug?

If so, then the general principle that applies is that we try to never knowingly merge a commit that leaves things in a broken state. This turns out to make a big difference when trying to understand the commit history -- every commit you look at, you can be confident the author and reviewer believed that it didn't break anything, and that helps a lot as you're mentally putting the pieces together and making sense of what they were trying to do.

Or is the bug one that already existed, and the previous version just didn't fix it? In that case, it's fine to fix bugs one at a time. But sometimes it's simpler to go straight to the final version, and when that is simpler it's better.

ishammahajan:
@gnprice I was a little confused with respect to the commit messages, but the explanation makes a lot of sense. The reason I initially did not remove the previous commit was because I thought the commits were more coherent that way. However they do introduce a bug, which is why they're not viable.

Rebased this onto master (and tested for both platforms, works as expected)! :)

gnprice:
Thanks, that's helpful!

I like the idea of replacing all the `lastTouchEventTimestamp` logic with remembering a timeout object and calling `clearTimeout` on it. That feels like it more directly uses the underlying system in the intended way, and should be easier to reason about. Definitely interested in getting that improvement merged.

I still don't feel I quite understand what this change is doing and why. Partly that's because there seem to be two different things it's doing:
* Instead of having this timestamp-cum-flag `lastTouchEventTimestamp`, checking it at the top of `handleLongPress`, and clearing it when we want to suppress a long-press... we suppress a pending long-press by just clearing the timeout that would have called `handleLongPress`.
  * This feels like basically a refactor -- it doesn't necessarily involve changing the outward behavior at all, just changing the internal logic to make it easier to think about.
  * The commit message is purely about this side, about making the code easier to think about.

* But then also, this change deletes some of the places where we used to clear `lastTouchEventTimestamp`, or set it to `Date.now()`.
  * Namely `handleScrollEvent`, the `'click'` handler, and the `touchstart` handler just before we set a new timeout.
  * This sure looks like it probably changes the behavior.
  * Do you believe it actually doesn't change the behavior? If so, that's quite interesting -- and should be explained in the commit message.
  * Or, do you believe it *does* change the behavior, and the new behavior is better? Again that's interesting, and should definitely be explained. :-) In particular, I'd like to understand exactly how the behavior is different; and why the old behavior is wrong and/or the new behavior is an improvement.

Does that match your understanding of what this change is doing? (If not, again that's interesting and I'd like to hear more!)

One thing that I think might help a lot in explaining this is to separate it into two commits -- one that is a pure refactor, just changes the structure without changing the behavior, and one that changes the behavior without changing the underlying structure. (This is a very general and very often helpful move to make.)

Specifically here, that might mean:
* One commit replaces `lastTouchEventTimestamp` with `longPressTimeout` and `clearTimeout`. But it *doesn't* change the list of places we reset things -- every `lastTouchEventTimestamp = ...` becomes a `clearTimeout`.
  * In this commit message, you mention that it is a pure refactor with no change in the visible behavior, and explain why you believe that's true.
  * As we've both said, the existing logic is pretty subtle. So this explanation might mostly be about explaining the subtleties of the existing logic, to see how the new logic matches it.
* Another commit deletes some of the places we reset things.
  * This might come either before or after the refactor commit.
  * In this commit message, you explain how the visible behavior changes and why that's good. Basically the old behavior (is there a bug you can describe?), the new behavior, and why the new behavior is better.
  * Or maybe this still doesn't change the visible behavior? In that case, explain why.

Is that separation possible? I actually don't yet understand this change well enough to know whether it's possible. :wink: So in part, even just for you to answer saying that it is or isn't possible (and if not possible, then why) will be helpful explanation.

If the separation is possible, I think it will almost certainly help for explaining the changes.

If it *isn't* possible to switch to `clearTimeout` without also changing the behavior, then you'll just have to explain both changes at once.

OK, those are specific suggestions -- the end goal is really just that I need to understand the change; and then the commits should contain whatever explanation I ended up needing, so the next reader can understand them too. So take what I've written as a source of questions to answer and some possible tips for how to answer them, but do whatever's best for taking the understanding in your head and communicating it into mine :slightly_smiling_face: , and then into other readers'.

ishammahajan:
Thanks for the detailed explanation and the suggestions!

I think the reason as to why I didn't explain this adequately is because with the amount of time I've worked with this piece of code, I thought it would be fairly obvious to everyone who read the diffs (which I should not have assumed.)

The changes I made when deleting the line in `handleScrollEvent` and related is because it was redundant pieces of code in the current logic as well (or so is what I have deduced) because the event listeners we use (`touchstart`, `touchend` (which covers the behaviour covered within the error catching in `click` event listener), and `touchmove` (which covers behavioural errors which might occur during scrolling)) already cover the cases which have been deleted in this commit.

The deletion doesn't change the behaviour because the lines didn't have any impact on the behaviour earlier. 
In the case that it used to, it most probably doesn't have any impact now, since the timeout gets cancelled on `touchmove` and `touchend`, which is why the flag / status catching logic we had in `handleLongPress` (for which the `lastTouchEventTimestamp` was set and reset) is also no longer required. Removal of the status catching logic negates the requirement of the deleted lines too.

With regards to the commit separation, it is definitely required 🙂 . The first commit to change the behaviour, and the second to remove the apparent redundant code. The redundancy of the code should be discussed before taking any actions though, I think.

gnprice:
> with the amount of time I've worked with this piece of code

Yep -- the goal now is to convert the understanding you've built in your head over that time into something that's in the repo (the code, and the history) and can be loaded into other people's heads :slightly_smiling_face: 


> The changes I made when deleting the line in `handleScrollEvent` and related is because it was redundant pieces of code in the current logic as well (or so is what I have deduced) because the event listeners we use (`touchstart`, `touchend` (which covers the behaviour covered within the error catching in `click` event listener), and `touchmove` (which covers behavioural errors which might occur during scrolling)) already cover the cases which have been deleted in this commit.
> 
> The deletion doesn't change the behaviour because the lines didn't have any impact on the behaviour earlier (but separating them is certainly possible.)
> In the case that it used to, it most probably doesn't have any impact now, since the timeout gets cancelled on `touchmove` and `touchend`,

Cool, interesting.

On the first paragraph here, I would ask if you're sure that's the case (and then why) -- but from the next couple of sentences, it looks like the answer to that is no, you're not sure :wink: 

Have you read through the documentation on how `click` interacts with `touchend` and friends? (There are some links in #3439, which I filed after studying this area in order to review the first version of this PR.) It gets a bit subtle. It's not obvious to me that these are redundant -- they might be, but getting to that conclusion would require some more careful reasoning about the possible sequences of events here.


> With regards to the commit separation, it is definitely required  . The first commit to change the behaviour, and the second to remove the apparent redundant code. The redundancy of the code should be discussed before taking any actions though, I think.

Well, I think being uncertain about some changes is actually an excellent reason to go ahead and separate them out from changes that you're more sure of! Just be sure to put the changes you're most confident of first, and the more uncertain ones later in the branch. That way, I can go ahead and merge the change we're sure of -- I'd love to merge the conversion to `clearTimeout`, but I can't while the commit it's part of makes changes that we're not yet confident about.

Here, it sounds like if you added back some `clearTimeout` calls in places where this version deletes `lastTouchEventTimestamp = 0`, you could make a commit that you're very confident doesn't change any of the visible behavior -- there's nothing a user could try that would tell the difference between before and after the commit. Is that right? If so, that would be a great first commit for the branch.

Then perhaps a later commit goes and deletes those calls, because you believe they're probably redundant. That will require some more careful explanation of why they're redundant.

But also, getting to that may require doing the work to understand this touch-event system as a whole, which is what #3439 is about. Instead of cutting those two lines, it might be better to go and do the broader changes discussed in #3439 . (Again, that'd be after making a mergeable version of the conversion to `clearTimeout`! I'd quite like to merge that.)

ishammahajan:
> the goal now is to convert the understanding you've built in your head over that time into something that's in the repo

My apologies 🙂

In this commit, my belief was that there was in fact one change which I can confidently say would be redundant with `clearTimeout`, the reason for which I have explained in the commit. 

I went through #3439, and filing that was an excellent choice! It details some very important changes which should take place, in order to improve code quality and understandability, and I think solving that issue would probably surface one or two bugs (which would be fixed within fixing it) making the app robust! I think the redundancy of the code should be handled in the PR which fixes that issue. I will check out the interaction of `touchevent` with the normal mouse events and default click events. I have some thoughts on the ideas suggested in that issue, which I will leave over there after researching a bit more.

I've taken the liberty of fixing a small part of that issue in a subsequent commit here. Hope you don't mind 😄.

gnprice:
Thanks for the revision, and the added bugfix! :slightly_smiling_face: . And nothing to apologize for.

I think in the revised version there's still an error of logic, in this hunk of the diff:
```
@@ -676,9 +664,8 @@ documentBody.addEventListener('touchstart', (e: TouchEvent) => {
 
   lastTouchPositionX = e.changedTouches[0].pageX;
   lastTouchPositionY = e.changedTouches[0].pageY;
-  lastTouchEventTimestamp = Date.now();
   hasLongPressed = false;
-  setTimeout(() => handleLongPress(target), 500);
+  longPressTimeout = setTimeout(() => handleLongPress(target), 500);
 });
```

We're setting a new timeout with a `handleLongPress` callback, but now we aren't canceling the old one. Moreover, because we're overwriting `longPressTimeout` with the new timeout, we're putting it beyond the reach of all our `clearTimeout` calls elsewhere -- so if we get here with a live timeout in `longPressTimeout`, we turn it into an unstoppable juggernaut that will soon trigger a long-press no matter what else happens.

In particular this means that if you touch with two fingers in sequence, then promptly lift them -- like quickly typing `we` on a QWERTY keyboard -- you'll see a long-press happen a half-second after the first touch, even though both fingers have been off the screen by then for several hundred ms.

Fortunately, I think I now understand your thinking well enough (in particular, the crucial point that this commit is *intended* to make no change at all to the external behavior) that I can just fix it, and merge: before clobbering `longPressTimeout` with a new timeout, we need to call `clearTimeout` on it.

This matches the effect that the `lastTouchEventTimestamp = Date.now()` line in this hunk had in the old/existing code -- it would cause any *older* callbacks to have no effect, while not affecting the one we created a couple of lines later.

(Going AFK just now, but I expect to make that fix and merge when back at my desk.)

gnprice:
And merged! Thanks again @ishammahajan for all your work on this.

> I went through #3439, and filing that was an excellent choice! It details some very important changes which should take place, in order to improve code quality and understandability, and I think solving that issue would probably surface one or two bugs (which would be fixed within fixing it) making the app robust! I think the redundancy of the code should be handled in the PR which fixes that issue. I will check out the interaction of `touchevent` with the normal mouse events and default click events. I have some thoughts on the ideas suggested in that issue, which I will leave over there after researching a bit more.

And this sounds like a good plan to me. :smiley: 

In addition to the logic fix I mentioned above, I made a few adjustments to the commit messages -- do take a look at them:
 * in the first commit, cut out the part that explains why you believed that one `clearTimeout` wasn't required -- because it turned out it was :wink: 
 * in the second commit, added a link to the specific page of docs you were referring to
 * in the second commit, I think "error-catching" isn't quite the right way to think about this event -- things like a modal alert or going out of focus are normal things to happen -- so I revised that framing a bit

ishammahajan:
The revisions were helpful in clarifying my understanding of commit messages a bit more, thanks!

I think this point needs to be drilled in - commits should be independently intelligible. I blatantly avoid this in the first commit here. Should have separated the refactor and the subsequent change in two different commits so you could have just ignored one of the commits if you wanted to.

Another softer but equally as important point, a commit message should be of three parts - one describing why the change was required (so readers could get a sense of the problem being solved), one describing how the problem is being solved in this particular place, and an optional third where concerns are expressed. Will try to stick to this format in later significant commits.

Thanks again! 🙂

gnprice:
> I think this point needs to be drilled in - commits should be independently intelligible.

Yep, exactly!

This is a really important point in Zulip's approach to commits, code review, etc. (and common in a lot of open-source projects with high code-quality standards, including the Linux kernel, and in industry environments when they have high code-quality standards.) I often summarize it as "clear and coherent commits", and we try to describe it in our [Git style guide](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) -- but for a lot of people it's a new practice, and quite natural not to be used to it at first.


> I blatantly avoid this in the first commit here. Should have separated the refactor and the subsequent change in two different commits so you could have just ignored one of the commits if you wanted to.

Yeah. And in fact it's more than that: even in a case where I do happily merge both changes, the separation helps a lot in understanding each change. Which is the main work that a reviewer needs to do before merging. :slightly_smiling_face: 

> Another softer but equally as important point, a commit message should be of three parts - one describing why the change was required (so readers could get a sense of the problem being solved), one describing how the problem is being solved in this particular place, and an optional third where concerns are expressed. Will try to stick to this format in later significant commits.

Yep, that's a good way of putting it!

There's another description, with some additional details, in the [latter half](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages) of the same style guide I linked to above.

(Hmm, one small thing I notice there that I don't quite agree with: 76 characters/columns is a bit too wide for prose. What I ask for is about 68 columns, up to 70 columns max.)

PR #3435, opened 2019-03-29
webview header: Modify topic header to include date.

### LINKED ISSUE #1336: It's really hard to figure out the date of the messages you're looking at

ishammahajan:
Many times when people are scrolling through messages they need to
see what date it is. This modification is made for this use-case.
Users can simply see the header to determine the date a particular
message was sent.

Header is converted into a flexbox and js for timestamp is added.
The date is human readable, and header is truncated if the topic
name is very large.

This would, in it's current iteration, only introduce the header
in the streams narrow. This would bring problems where users would
not be able to get the date for a certain message in 1:1 or in
the topic narrow. The complete PR, however, aims to fix this by
introducing headers to these narrows as well, the benefits of 
which are detailed in https://github.com/zulip/zulip-mobile/issues/1336#issuecomment-475026033.

Tasklist :-
- [x] Add date in the header for streams.
- [x] Add header for 1:1 / group messages.
- [x] Confirm functionality for 1:1 / group messages.
- [x] Add header for topic messages.
- [x] Fix bug appearing for messages from same person but different date.

([this](https://imgur.com/T2WfbIx) is the current state of this PR)

Fixes #1336

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -152,11 +152,6 @@ hr {
 .header-wrapper {
   cursor: pointer;
 }
-.stream-header {
-  padding: 0;
-  display: flex;
-  flex-direction: row;
-}
 .stream-text,
 .topic-header,
 .private-header {
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -163,9 +163,6 @@ hr {
   line-height: 2;
   white-space: nowrap;
 }
-.private-header {
-  padding: 0 0.5em;
-}
 .topic-header {
   background: #ccc;
   min-width: 30%;
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -170,14 +170,14 @@ hr {
 .stream-text {
   padding: 0 0.5em;
 }
-.topic-text {
+.header-text {
   flex: 1;
   padding: 0 0.5em;
   overflow: hidden;
   text-overflow: ellipsis;
   pointer-events: none;
 }
-.topic-date {
+.header-date {
   opacity: 0.5;
   padding: 0 0.5em;
   pointer-events: none;
--- a/src/webview/html/messageHeaderAsHtml.js
+++ b/src/webview/html/messageHeaderAsHtml.js
@@ -26,7 +26,7 @@ export default (
   type HeaderStyle = 'none' | 'topic+date' | 'date' | 'full';
   const headerStyle: HeaderStyle = caseNarrow(narrow, {
     stream: () => 'topic+date',
-    topic: () => 'none',
+    topic: () => 'topic+date',
 
     pm: () => 'date',
     groupPm: () => 'full',
--- a/src/webview/html/messageHeaderAsHtml.js
+++ b/src/webview/html/messageHeaderAsHtml.js
@@ -72,8 +72,8 @@ export default (
 
     return template`
 <div class="header-wrapper header topic-header"
-    data-msg-id="${item.id}"
-    data-narrow="${topicNarrowStr}">
+    data-narrow="${topicNarrowStr}"
+    data-msg-id="${item.id}">
   <div class="header stream-text"
        style="color: ${textColor};
               background: ${backgroundColor}"
--- a/src/webview/html/messageHeaderAsHtml.js
+++ b/src/webview/html/messageHeaderAsHtml.js
@@ -71,7 +71,7 @@ export default (
     const topicHtml = renderSubject(item);
 
     return template`
-<div class="header-wrapper header stream-header topic-header"
+<div class="header-wrapper header topic-header"
     data-msg-id="${item.id}"
     data-narrow="${topicNarrowStr}">
   <div class="header stream-text"
--- a/src/webview/html/messageHeaderAsHtml.js
+++ b/src/webview/html/messageHeaderAsHtml.js
@@ -23,13 +23,13 @@ export default (
   narrow: Narrow,
   item: Message | Outbox | {||},
 ) => {
-  type HeaderStyle = 'none' | 'topic+date' | 'full';
+  type HeaderStyle = 'none' | 'topic+date' | 'date' | 'full';
   const headerStyle: HeaderStyle = caseNarrow(narrow, {
     stream: () => 'topic+date',
     topic: () => 'none',
 
-    pm: () => 'none',
-    groupPm: () => 'none',
+    pm: () => 'date',
+    groupPm: () => 'full',
 
     home: () => 'full',
     starred: () => 'full',
@@ -52,8 +52,8 @@ export default (
   data-narrow="${topicNarrowStr}"
   data-msg-id="${item.id}"
 >
-  <div class="topic-text">$!${topicHtml}</div>
-  <div class="topic-date">${humanDate(new Date(item.timestamp * 1000))}</div>
+  <div class="header-text">$!${topicHtml}</div>
+  <div class="header-date">${humanDate(new Date(item.timestamp * 1000))}</div>
 </div>
     `;
   }
@@ -80,13 +80,13 @@ export default (
        data-narrow="${streamNarrowStr}">
     # ${item.display_recipient}
   </div>
-  <div class="topic-text">$!${topicHtml}</div>
-  <div class="topic-date">${humanDate(new Date(item.timestamp * 1000))}</div>
+  <div class="header-text">$!${topicHtml}</div>
+  <div class="header-date">${humanDate(new Date(item.timestamp * 1000))}</div>
 </div>
     `;
   }
 
-  if (item.type === 'private' && headerStyle === 'full') {
+  if (item.type === 'private') {
     const recipients =
       item.display_recipient.length === 1 && item.display_recipient[0].email === ownEmail
         ? item.display_recipient
@@ -98,16 +98,30 @@ export default (
         : groupNarrow(recipients.map(r => r.email));
     const privateNarrowStr = JSON.stringify(narrowObj);
 
-    return template`
+    if (headerStyle === 'date') {
+      return template`
 <div class="header-wrapper private-header header"
      data-narrow="${privateNarrowStr}"
      data-msg-id="${item.id}">
-  ${recipients
+  <empty></empty>
+  <div class="header-date">${humanDate(new Date(item.timestamp * 1000))}</div>
+</div>
+      `;
+    }
+
+    if (headerStyle === 'full') {
+      return template`
+<div class="header-wrapper private-header header"
+     data-narrow="${privateNarrowStr}"
+     data-msg-id="${item.id}">
+  <div class="header-text">${recipients
     .map(r => r.full_name)
     .sort()
-    .join(', ')}
+    .join(', ')}</div>
+  <div class="header-date">${humanDate(new Date(item.timestamp * 1000))}</div>
 </div>
-    `;
+      `;
+    }
   }
 
   return '';
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -207,6 +207,14 @@ function walkToMessage(
   return element;
 }
 
+function walkToHeaderAbove(start: ?Element): ?Element {
+  let element: ?Element = start;
+  while (element && !element.classList.contains('header')) {
+    element = element.previousElementSibling;
+  }
+  return element;
+}
+
 function firstMessage(): ?Element {
   return walkToMessage(documentBody.firstElementChild, 'nextElementSibling');
 }
@@ -284,6 +292,27 @@ function visibleMessageIds(): { first: number, last: number } {
   return { first, last };
 }
 
+function getFirstVisibleMessage(): Element {
+  const top = 0;
+  const bottom = viewportHeight;
+  let message = document.createElement('null');
+
+  function walkElements(start: ?Element) {
+    let element = start;
+    while (element && isVisible(element, top, bottom)) {
+      if (element.classList.contains('message')) {
+        message = element;
+      }
+      element = element.previousElementSibling;
+    }
+  }
+
+  const start = someVisibleMessage(top, bottom);
+  walkElements(start);
+
+  return message;
+}
+
 /** DEPRECATED */
 const getMessageNode = (node: ?Node): ?Node => {
   let curNode = node;
@@ -333,6 +362,18 @@ const sendScrollMessage = () => {
   prevMessageRange = messageRange;
 };
 
+const handleDateInHeader = () => {
+  const firstVisibleMessage = getFirstVisibleMessage();
+
+  if (firstVisibleMessage) {
+    const replaceableDate = firstVisibleMessage.getAttribute('msg-human-date');
+    const closestHeader = walkToHeaderAbove(firstVisibleMessage);
+    if (replaceableDate && closestHeader) {
+      closestHeader.getElementsByClassName('header-date').item(0).innerHTML = replaceableDate;
+    }
+  }
+};
+
 // If the message list is too short to scroll, fake a scroll event
 // in order to cause the messages to be marked as read.
 const sendScrollMessageIfListShort = () => {
@@ -362,6 +403,7 @@ const handleScrollEvent = () => {
   }
 
   sendScrollMessage();
+  handleDateInHeader();
 
   const nearEnd = documentBody.offsetHeight - window.scrollY - window.innerHeight > 100;
   showHideElement('scroll-bottom', nearEnd);
--- a/src/message/renderMessages.js
+++ b/src/message/renderMessages.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import type { Message, Narrow, Outbox, RenderedSectionDescriptor } from '../types';
-import { isTopicNarrow, isPrivateOrGroupNarrow } from '../utils/narrow';
+import { isTopicNarrow } from '../utils/narrow';
 import { isSameRecipient } from '../utils/recipient';
 import { isSameDay } from '../utils/date';
 
@@ -9,7 +9,7 @@ export default (
   narrow: Narrow,
 ): RenderedSectionDescriptor[] => {
   let prevItem;
-  const showHeader = !isPrivateOrGroupNarrow(narrow) && !isTopicNarrow(narrow);
+  const showHeader = !isTopicNarrow(narrow);
 
   return messages.reduce(
     (sections, item) => {
--- a/src/message/renderMessages.js
+++ b/src/message/renderMessages.js
@@ -1,6 +1,5 @@
 /* @flow strict-local */
 import type { Message, Narrow, Outbox, RenderedSectionDescriptor } from '../types';
-import { isTopicNarrow } from '../utils/narrow';
 import { isSameRecipient } from '../utils/recipient';
 import { isSameDay } from '../utils/date';
 
@@ -9,7 +8,6 @@ export default (
   narrow: Narrow,
 ): RenderedSectionDescriptor[] => {
   let prevItem;
-  const showHeader = !isTopicNarrow(narrow);
 
   return messages.reduce(
     (sections, item) => {
@@ -25,7 +23,7 @@ export default (
         });
       }
       const diffRecipient = !isSameRecipient(prevItem, item);
-      if (showHeader && diffRecipient) {
+      if (diffRecipient) {
         sections.push({
           key: `header${item.id}`,
           message: item,
--- a/src/title/TitleStream.js
+++ b/src/title/TitleStream.js
@@ -6,7 +6,6 @@ import { StyleSheet, Text, View } from 'react-native';
 import type { Narrow, Stream, Subscription, Dispatch } from '../types';
 import { connect } from '../react-redux';
 import StreamIcon from '../streams/StreamIcon';
-import { isTopicNarrow } from '../utils/narrow';
 import { getStreamInNarrow } from '../selectors';
 import styles from '../styles';
 
@@ -37,7 +36,7 @@ class TitleStream extends PureComponent<Props> {
   });
 
   render() {
-    const { narrow, stream, color } = this.props;
+    const { stream, color } = this.props;
 
     return (
       <View style={this.styles.outer}>
@@ -52,11 +51,6 @@ class TitleStream extends PureComponent<Props> {
             {stream.name}
           </Text>
         </View>
-        {isTopicNarrow(narrow) && (
-          <Text style={[styles.navSubtitle, { color }]} numberOfLines={1} ellipsizeMode="tail">
-            {narrow[1].operand}
-          </Text>
-        )}
       </View>
     );
   }
--- a/src/webview/html/messageAsHtml.js
+++ b/src/webview/html/messageAsHtml.js
@@ -11,7 +11,7 @@ import type {
 } from '../../types';
 import type { BackgroundData } from '../MessageList';
 import { getAvatarFromMessage } from '../../utils/avatar';
-import { shortTime } from '../../utils/date';
+import { shortTime, humanDate } from '../../utils/date';
 import aggregateReactions from '../../reactions/aggregateReactions';
 import { codeToEmojiMap } from '../../emoji/data';
 import processAlertWords from './processAlertWords';
@@ -80,6 +80,7 @@ export default (backgroundData: BackgroundData, message: Message | Outbox, isBri
      class="message ${isBrief ? 'message-brief' : 'message-full'}"
      id="msg-${id}"
      data-msg-id="${id}"
+     msg-human-date="${humanDate(new Date(message.timestamp * 1000))}"
      $!${flagStrings.map(flag => template`data-${flag}="true" `).join('')}
     >`;
 
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -125,6 +125,16 @@ function walkToMessage(start, step) {
   return element;
 }
 
+function walkToHeaderAbove(start) {
+  var element = start;
+
+  while (element && !element.classList.contains('header')) {
+    element = element.previousElementSibling;
+  }
+
+  return element;
+}
+
 function firstMessage() {
   return walkToMessage(documentBody.firstElementChild, 'nextElementSibling');
 }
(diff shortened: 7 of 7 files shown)


src/webview/js/js.js
```diff
@@ -662,7 +670,13 @@ const handleLongPress = (target: Element) => {
...
   sendMessage({
     type: 'longPress',
-    target: target.matches('.header') ? 'header' : target.matches('a') ? 'link' : 'message',
+    target:
+      target.matches('.header')
+      || (target.parentNode instanceof Element && target.parentNode.matches('.header'))
+        ? 'header'
+        : target.matches('a')
+          ? 'link'
+          : 'message',
```
gnprice: Hmm, this is getting messy.
* If we do end up with logic looking like this, it should definitely get pulled out into a local variable -- it's too complex to belong inline in a function call like this one.
* How about using `closest` instead? See [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest), and example usage for `.message_inline_image a` and friends above.
  * This could also simplify the other part of the diff, above.

src/webview/css/css.js
```diff
@@ -170,6 +170,8 @@ hr {
   padding-left: 0.5em;
   background: #ccc;
   min-width: 30%;
+  display: flex;
```
gnprice: nit: trailing space on this line

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -39,7 +40,8 @@ export default (
   data-narrow="${topicNarrowStr}"
   data-msg-id="${item.id}"
 >
-  $!${topicHtml}
+  <div>$!${topicHtml}</div>
+  <div>$!${humanDate(new Date(item.timestamp * 1000))}</div>
```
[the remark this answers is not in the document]
ishammahajan: I see. Thanks, fixed!

src/webview/js/js.js
ishammahajan: I agree, was going to ask about that. Fixed! 😄.

src/webview/css/css.js
ishammahajan: Fixed 🙂

ishammahajan:
@gnprice Thanks for the review, have addressed the concerns on my machine! Sorry about the screenshot, I wasn't ready for a review right now 😅, regardless I'll add screenshots/suitable recordings to this shortly. Been busy lately, didn't have enough time to dedicate to the project.

Another thing, I had originally marked this as a WIP because the CSS and some JS might change because of the complexity of implementing this in topics/1:1 pms, which is why I was holding back on asking for a review.

gnprice:
Got it, no problem :slightly_smiling_face:  When you'd like another review, just leave a comment here!

ishammahajan:
I think this is ready for review! I have detailed what has been done to solve each checkbox in the relevant commit, which hopefully is understandable 🙂 .

Recordings are as follows :-
special-date-in-header
topic-date-in-header
pmgroup-date-in-header

src/webview/js/js.js
```diff
@@ -593,6 +593,14 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.parentNode instanceof Element && target.parentNode.matches('.header')) {
```
borisyankov: Is this change required only because you've added a `div` inside the '.header'?

Try using CSS: `pointer-events: none;` for that `div` and it will stop handling click/touch events.

src/webview/css/css.js
```diff
@@ -157,20 +159,22 @@ hr {
...
 }
-.topic-text {
+.topic-content {
   flex: 1;
   padding-left: 0.5em;
   background: #ccc;
   min-width: 30%;
 }
+.topic-text {
+  overflow: hidden;
```
borisyankov: You definitely want some margin on the left. 
Check out how longer topics almost blend with the date div
Try `0.5em` or `1em`

src/webview/css/css.js
```diff
@@ -157,20 +159,22 @@ hr {
...
+  text-overflow: ellipsis;
+}
```
borisyankov: Adding some transparency to the text (about 0.5) will make it look better, as this is secondary information to the topic name.

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -29,22 +31,23 @@ export default (
...
 >
-  $!${topicHtml}
+  <div class="topic-text">$!${topicHtml}</div>
+  <div class="human-date">$!${humanDate(new Date(item.timestamp * 1000))}</div>
 </div>
     `;
   }
 
-  if (item.type === 'stream') {
+  if (item.type === 'stream' && (isSpecialNarrow(narrow) || isHomeNarrow(narrow))) {
```
gnprice: Is this in fact redundant? (And in the `private` case below, is the new condition in fact equivalent to the old?) How do you know -- or, how would you find out for sure?

Also, there's already kind of a long list of possible types of narrows. What if a new one gets added -- how can you write this so that it will *stay* correct then?

I think what you probably want here is `!isStreamOrTopicNarrow`. That's what this block is really about -- that's the description that will stay true regardless of what other miscellaneous types of narrows exist.

Another solution which would make sure you've comprehensively covered the cases, and make sure it stays correct if/when more cases are added, would be to use `caseNarrow`. (See the example uses of it in `utils/narrow.js`. See also `git log --stat -p fa6134aa6^..e9fe1e801` .) But I think in this case the simpler `isStreamOrTopicNarrow` will do just fine.

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -29,22 +31,23 @@ export default (
...
 
-  if (isStreamNarrow(narrow)) {
+  if (isStreamNarrow(narrow) || isTopicNarrow(narrow)) {
     const topicNarrowStr = JSON.stringify(topicNarrow(item.display_recipient, item.subject));
     const topicHtml = renderSubject(item);
 
     return template`
 <div
-  class="header-wrapper header topic-text"
+  class="header-wrapper header topic-content"
```
gnprice: This name is kind of confusing -- we generally use "content" for the content of the message itself.

Perhaps `topic-header`?

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -57,45 +60,49 @@ export default (
...
+<div class="header-wrapper header stream-header" data-msg-id="${item.id}">
   <div class="header stream-text"
        style="color: ${textColor};
               background: ${backgroundColor}"
        data-narrow="${streamNarrowStr}">
     # ${item.display_recipient}
   </div>
-  <div class="header topic-text" data-narrow="${topicNarrowStr}">
-    $!${topicHtml}
+  <div class="header-wrapper header topic-content" data-narrow="${topicNarrowStr}">
```
gnprice: Why the nested `header-wrapper`, and the nested `header`? That seems likely to cause confusion and/or a bug, if it doesn't cause a live bug already.

gnprice: Also `topic-content` is *definitely* going to be confusing if what it really means is "topic plus date". The date isn't really anything to do with the topic, so `topic-content` doesn't give a hint of that.

If you want a common class for "topic + date" in this and the other case, you could if nothing else call it "topic-and-date". But before that, I would ask why they're grouped together in the first place. :slightly_smiling_face: Why not add the "date" div as a sibling of both the `stream-text` and the `topic-text`?

src/webview/html/messageHeaderAsHtml.js
gnprice: Oh, or here's an alternate way of organizing this logic: put a `caseNarrow` at the *outside*, before looking at `item.type`. Then you get to be super explicit about what the behavior is for each possible type of narrow.

Because much of the logic will be shared across several types of narrow, I'd probably structure that something like this:
```
  const headerStyle = caseNarrow(narrow, {
    stream: () => 'topic+date',
    pm: () => 'date',
    search: () => 'full',
    // ... etc.    
  });

  // Now we're completely done with `narrow` -- the information we need from it is all in `headerStyle`.

  if (item.type === 'stream' && headerStyle === 'topic+date') {
    // ...
  } else // ...
```

The exact set of possible values for `headerStyle` will be up to what you're doing, and might change in later commits in the branch. This will be a way to separate that decision-making in its own bit of code where it's easy to see what's happening.

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -57,45 +60,49 @@ export default (
...
-    && !isGroupNarrow(narrow)
-    && !isTopicNarrow(narrow)
-  ) {
-    const recipients =
-      item.display_recipient.length === 1 && item.display_recipient[0].email === ownEmail
-        ? item.display_recipient
-        : item.display_recipient.filter(r => r.email !== ownEmail);
+  if (item.type === 'private') {
+    let recipientsStr;
+    let privateNarrowStr;
```
gnprice: I don't like these mutable variables -- they're a smell, a sign that something's off in the logic.

One observation: you're including the div for "recipients" even when there's nothing there. That seems likely to cause trouble -- if nothing else, confusion.

Probably better to just have a separate case for the new situation where we're showing this header but don't want to show the recipients.

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -57,45 +60,49 @@ export default (
...
-<div class="header-wrapper private-header header"
+<div class="header-wrapper header private-header"
      data-narrow="${privateNarrowStr}"
      data-msg-id="${item.id}">
-  ${recipients
-    .map(r => r.full_name)
-    .sort()
-    .join(', ')}
+     <div class="topic-text">$!${recipientsStr}</div>
+     <div class="human-date">$!${humanDate(new Date(item.timestamp * 1000))}</div>
```
gnprice: Dangerous -- this is really important!

Same thing I mentioned in a comment in the last round:
https://github.com/zulip/zulip-mobile/pull/3435#discussion_r271046563

Essentially this is a type error -- you are taking content that is plain text, and using it in a place that expects HTML. The only correct way to do that is to *encode* the text as HTML.

Otherwise when some user changes their name in Zulip to something like, say, `<script>alert("lol pwned");</script>`, and then sends you a PM -- something very much the wrong thing happens. :stuck_out_tongue_winking_eye: 

The term for this is "HTML injection". Here's one reasonable piece of background on the issue:
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md
In that page's terms, "Rule # 1" is the applicable rule here.

In our `template` function which we're using here, the normal way to encode text as HTML is to use a plain `${...}`. The `!` in `$!$` is a bright red warning sign! Please heed the warning sign.

gnprice:
Also, thanks for those recordings! Very helpful :slightly_smiling_face: 

In the 1:1 PM case, the bar looks oddly blank with just the date at the end -- it looks like there's some kind of bug. (Reminds me of how some paper exams, forms, etc. will say "This Page Intentionally Left Blank" rather than actually leave a page blank, to avoid people worrying that there was some error in printing.) Not immediately sure what'd be the best way to address that, though.

src/webview/css/css.js
ishammahajan: > some margin on the left.

I agree, it's definitely blending in a lot. Also I assume you mean margin on the right of the topic-text?

src/webview/css/css.js
ishammahajan: Agreed, done!

src/webview/html/messageHeaderAsHtml.js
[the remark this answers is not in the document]
ishammahajan: So I see that I hadn't addressed this in the latest commit (for reasons I don't know, the diff got erased from WD.)

Anyways it's fixed now.

src/webview/html/messageHeaderAsHtml.js
ishammahajan: My original intention of grouping them as separate div's was to make it clear that there will be different classes to represent both the text and the date, and I'll admit my inexperience with CSS; I don't know how to not bring two divs and still maintain the flexbox with `justify-content: space-between;` The date manages to overflow no matter what changes I make to the CSS to prevent it from doing so. Making two divs solves this problem, so I went with it.
Screenshot 2019-04-28 at 11 54 13 AM

Suggestions about changes to the CSS are welcome! 😄

ishammahajan: The `header-wrapper` and `header` nesting was definitely redundant (although no observable bugs could be found by me) and has been taken care of, thanks! Also changed the name for `topic-content` to `topic-header` in order to correctly represent what type of narrow the header belongs to (like `stream-header` and `private-header` represent their narrows.)

src/webview/js/js.js
ishammahajan: Done! Thanks for the advice! 🙂

src/webview/html/messageHeaderAsHtml.js
ishammahajan: I agree with this. Introduced another case instead of checking with variables (now checking the narrow with `isPrivateNarrow() && !isSpecialNarrow()` to check for if the screen the user is on is a private conversation.)

src/webview/html/messageHeaderAsHtml.js
ishammahajan: Thanks for the detailed analysis!

I agree with all of the points you made. I do think the second solution is what we would want to implement. Unfortunately I did see the review on this commit after I saw the others, so I'll take some time to implement this (and fix conflicts) and push the final version.

src/webview/html/messageHeaderAsHtml.js
ishammahajan: Thanks for the explanation and the link (helped a lot more than understanding HTML injection too :) )!

I was actually aware of HTML injection being a real thing, and I had fixed this issue in WD before (as explained in the comment above) must have forgotten to commit it for some reason.

(That link is very resourceful!)

ishammahajan:
@borisyankov @gnprice thanks for the reviews! I have addressed all the points covered and made the appropriate changes.

The failed check is because of the changes introduced in the first commit with using `caseNarrow`. Unfortunately I do not have enough time today to look into fixing it, but I think it should be done by the day after tomorrow.

I will be posting updated screenshots after the design is finalised and implemented, so that work done in capturing and posting these is not repeated. 

The design finalisation I am talking about is the problem concerning how dates should be shown in the PM narrow (as said in the last comment). I've given it some thought and my opinion is to keep the entirety of the header (the background) and the header text is aligned to the middle (optionally giving it a defined width with rounded corners, although I don't think this would go well with the rest of the app). The reasoning behind this is that we cannot remove the name from the appbar because that would involve removing the `last active` indicator, which is not a desired trait, and we cannot stick the date to the right due to the reasons explained in the comment above.

ishammahajan:
I'm not too sure about the design though, if there are better ideas I would like to hear them.

borisyankov:
I was definitely not a fan of adding the headers in the narrows that previously didn't have one.
Especially disliked the one in topic narrows, that moves the topic name there too.

The first issue I have is that they are just very confusing, you expect this to be a separator that just got stuck at top temporarily but in these narrows it is a permanent thing that is also empty (besides the date on the right).

A smaller issue, these take vertical space that previously was free for display of messages.

Another thing that we have to improve:
 * now we have duplication of 'time rows' and these dates
 * we are still not showing the actual time of the message (the sticky headers are not a great place for this, just noting it as an outstanding issue)

I don't see a reason not to merge the changes for the narrows already with sticky headers, as these look great.

We can iterate on the other narrows more though. Maybe a sticky date that only takes the right side of the screen (basically a header that is something like `{ width: 100, right: 0 }` in css.

gnprice:

* The first commit, making that `headerStyle` refactor, looks great. I'm about to merge it, with two adjustments:
  * The two test failures it caused were because a set of tests are passing half-formed, ill-typed data, causing this function to get `undefined` for `narrow`. (Many of our tests don't pass the type-checker, and so they don't have the `/* @flow */` marking that causes the type-checker to report errors in them.) I fixed up the tests to pass a valid `Narrow` value.
  * I added a small second commit on top that organizes the different narrow types to I think make the patterns a bit easier to see, and also adds a type annotation to help Flow more precisely understand the code.

* On the design:

  > I will be posting updated screenshots after the design is finalised and implemented, so that work done in capturing and posting these is not repeated.

  The approach I think works best is: any time you make visible changes to the design which you'd like to discuss, make new screenshots and post them. They really help a lot in getting everyone (even just you and one reviewer) on the same page about what designs we're talking about.

  To keep the amount of work down, it'd be perfectly fine to post just one or two still screenshots if that illustrates what's changed. The original set of three anigif captures is great for seeing the design in full context, but the reader can combine those with a still to get a picture of the new design.

* Haven't yet read in detail the second and later commits -- I'll do that next.

gnprice:
> * The first commit, making that `headerStyle` refactor, looks great. I'm about to merge it, with two adjustments:

And merged! This is a6b4b5ede and b1b873f2c .

gnprice:
On the next commit, modifying the header only for stream messages:

* Careful with changing where we do and don't HTML-encode things:

```
-  $!${topicHtml}
+  <div class="topic-text">${topicHtml}</div>
```

It's true that `$!$`, which uses data without HTML-encoding it, is dangerous because skipping HTML encoding means HTML injection... but if the data is already HTML, then HTML-encoding it a second time is *also* a bug, though a less severe one.

With this change, if the topic is actually "2 + 2 < 5", it gets doubly encoded to become "2 + 2 &amp;lt; 5", which means it displays as "2 + 2 &lt; 5".

I just wrote and pushed to master a quick test that demonstrates this -- see #3467. When you rebase on the new master, you'll find that this change breaks that test.

Same thing in the other case here, even though I didn't add a test for that today.

* This is still a bit confusing to me:

```
-<div class="header-wrapper stream-header" data-msg-id="${item.id}">
+<div class="header-wrapper header stream-header topic-header"
+    data-msg-id="${item.id}"
+    data-narrow="${topicNarrowStr}">
   <div class="header stream-text"
        style="color: ${textColor};
               background: ${backgroundColor}"
        data-narrow="${streamNarrowStr}">
     # ${item.display_recipient}
   </div>
-  <div class="header topic-text" data-narrow="${topicNarrowStr}">
-    $!${topicHtml}
-  </div>
+  <div class="topic-text">${topicHtml}</div>
+  <div class="topic-date">${humanDate(new Date(item.timestamp * 1000))}</div>
 </div>
```

One thing: what is `topic-header` intended to mean?

Also the nesting of `header` is confusing. I think the effect will be that if you tap on the stream name, it'll still narrow to the stream, but if you tap anywhere else in the header (the topic, the blank space, the date, or possibly some stray padding somewhere even next to the stream name) it'll narrow to the topic.

Is it intentional that tapping on the date will narrow to the topic? If so, that should definitely be explicit in the commit message. There may also be a cleaner way to do that that doesn't involve this nesting (and the worry that maybe some edge around the stream name will also narrow to the topic.)

* In the CSS, there are two `.topic-header` blocks, and they contradict each other on `background`.

ishammahajan:
@borisyankov thanks for the insight on design!

> not a fan of adding the headers in the narrows that previously didn't have one

I have continued the discussion in [zulip](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/header.20design/near/737817) about this, let's discuss the points further over there. The gist of my opinion about this is that, while it does sacrifice a tiny bit of the vertical space, it brings uniformity to the design, since now headers appear everywhere and actually have a function which makes them non-static. It also gives the added benefit of being in sync with the webapp's design.

The possible designs are the two following and everything in between.
Screenshot 2019-05-02 at 11 58 09 PM
header date mockup

> you expect this to be a separator that just got stuck at top temporarily but in these narrows it is a permanent thing that is also empty

I do agree with this, we don't yet have a solution for showing dates in private headers, and while I have posted some mockups of a possible design in chat, I do think there may be better solutions out there.

> * now we have duplication of 'time rows' and these dates

I do think both are necessary, since if there is a case where the timerow is in the middle of the screen, the date on the header will not reflect that of the timerow which is on the screen, it will rather reflect the date of the timerow above it.

It would also be applicable in another case where the topic of the conversation does not change as the clock strikes midnight. In this case since the topic doesn't change, another header is not created, but a timerow is still required because we do still need to display the date.
Screenshot 2019-05-02 at 11 19 13 PM
The webapp handles this similarly as well.

> * we are still not showing the actual time of the message (the sticky headers are not a great place for this, just noting it as an outstanding issue)

I'll be working on that next, as soon as design on this is finalised and this PR is merged :)

ishammahajan:
@gnprice thanks for pushing the fix for the first commit!

> With this change, if the topic is actually "2 + 2 < 5", it gets doubly encoded to become "2 + 2 &lt; 5", which means it displays as "2 + 2 < 5".

It gets doubly encoded to become "2 + 2 &amp;lt; 5", right? Single encoding would bring it to "2 + 2 &lt; 5".

> I think the effect will be that if you tap on the stream name ...

That's exactly right! I wanted to make the structure such that the area a user taps to get to the stream (the stream name `# stream-name`) would get superimposed on the header for the topic itself. This way the headers are more modular and if there was another change required to header (say) it would be easy to think of/make. If the headers are not nested, this modularity is non-existent. The `topic-header` is a 'type' of header, just like `stream-header` and `private-header`.

> * In the CSS, there are two `.topic-header` blocks, and they contradict each other on `background`

They do contradict each other on colours, but that's because one of the colours is for dark theme and one is for light (I haven't changed anything with respect to the colours. As the commit says, I have simply renamed what was `topic-text` to `topic-header`, topic-text also had apparently contradictory blocks.)

--
I have rebased onto master and made changes accordingly. Tested again, seems to work well. Screen recordings are almost exactly the same as the previous with changes to transparency of the date and a padding introduced between topic-text and the date so as to not make the header congested.
Screenshot 2019-05-03 at 12 04 54 AM
Screenshot 2019-05-03 at 12 07 53 AM

gnprice:
> @gnprice thanks for pushing the fix for the first commit!
> 
> > With this change, if the topic is actually "2 + 2 < 5", it gets doubly encoded to become "2 + 2 < 5", which means it displays as "2 + 2 < 5".
> 
> It gets doubly encoded to become "2 + 2 &lt; 5", right? Single encoding would bring it to "2 + 2 < 5".

Oof, this is confusing because GitHub decodes these entities in comments... so in the rendering of my comment things come out differently than how I'd typed them. I didn't notice that until just now -- let me try again, using the "preview" feature to make sure it comes out right:

With this change, if the topic is actually "2 + 2 < 5" (with a less-than), it gets doubly encoded to become `2 + 2 &amp;lt; 5` (the less-than replaced with a sequence of 8 characters including two semicolons), which means it displays as "2 + 2 &amp;lt; 5" (the less-than replaced with 4 characters including one semicolon.)

(On top of everything else, in fact GitHub apparently has an HTML-encoding bug of its own :rofl: :cry:  where the "Quote reply" feature copies the displayed text version, without encoding things back as HTML entities... so they get decoded a second time, and your quote from my comment comes out differently from my actual comment, and the same with my quote from your comment.)

Does that make better sense now? Thanks for pointing out the confusion :slightly_smiling_face:

gnprice:
> > * In the CSS, there are two `.topic-header` blocks, and they contradict each other on `background`
> 
> They do contradict each other on colours, but that's because one of the colours is for dark theme and one is for light (I haven't changed anything with respect to the colours. As the commit says, I have simply renamed what was `topic-text` to `topic-header`, topic-text also had apparently contradictory blocks.)

Ah I see, makes sense, thanks. I saw the two in the diff (with just the usual 3 lines of context), and mistakenly assumed they were both in the one giant block.


> > I think the effect will be that if you tap on the stream name ...
> 
> That's exactly right! I wanted to make the structure such that the area a user taps to get to the stream (the stream name `# stream-name`) would get superimposed on the header for the topic itself. This way the headers are more modular and if there was another change required to header (say) it would be easy to think of/make. If the headers are not nested, this modularity is non-existent. The `topic-header` is a 'type' of header, just like `stream-header` and `private-header`.

Hmm, I see! And thanks for the added explanation in the commit message; that's quite helpful.

I like the idea of this being a modular piece that's dropped in. A couple of things that make it not quite there even with this change (and these would be a good target for followup commits):
* The `stream-header` class. I'm not certain this is actually doing anything for us at all -- the `display: flex` is definitely redundant, and the two other attributes may be too.
* The order of the `data-narrow` and `data-msg-id` attributes, and the whitespace before the closing `>`. These have no semantic effect at all... but they do get in the way of seeing the parallelism between these cases, as a human reader :wink:

gnprice:
I just pushed the next commit, adding dates in the existing recipient headers for stream messages. Thanks again @ishammahajan !

I also added a small commit on top that fixes the bug introduced. In general we avoid having any given commit break things, even if a later commit in the PR fixes it -- and this is an illustration of one reason why, as we're not ready to merge that next commit. :slightly_smiling_face:

Another reason we avoid having any given commit break things is that it's generally a lot easier to read and understand the commit when you can assume the author and reviewer(s) believed it all made sense together and didn't break anything. (There can always still be bugs they missed, of course, but at least you have a much better shot at following what those humans were thinking.) So the clear identification of the bug in the commit message mitigates that, and was crucial in it being OK to merge the commit with the known bug. But I still didn't want to merge it without also promptly fixing it. :wink:

gnprice:
(For design discussion for the next steps, I'll move to [that chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/header.20design/near/737817) you created.)

Also, FTR: the commits I just merged were ada59a7d8 and 123fd7a10.

gnprice:
Oh, one next step that I think we can do before settling further design questions: PM headers, in the cases where they already appear (namely "special narrows" like all-messages, starred, and search.) In those narrows, it looks a bit incongruous to have dates on headers for stream messages but not PMs.

This is basically one half of the changes that appear in your next commit:
```
msglist header: Modify pm/group headers to show date.
```
but that commit also starts showing those headers in PM *narrows*, which I'd prefer to hold off on for a bit more design discussion. Would you split that change out as a separate first commit?

I think that split makes things a bit clearer even if we were about to merge them both, though it wouldn't be absolutely essential in that case. (One thing I would insist on even then, though: the commit message should definitely be clear that it's making both those changes. In the current one it's easy to miss that and read it as only being about adding dates to existing headers.)

ishammahajan:
> Does that make better sense now? Thanks for pointing out the confusion

It does 😂 ! My reply to your comment was based on the me trying to understand the test anyway.

> * The order of the `data-narrow` and `data-msg-id` attributes

I'll fix that promptly, thanks!

> * The `stream-header` class. I'm not certain this is actually doing anything for us at all -- the `display: flex` is definitely redundant, and the two other attributes may be too.

Just tested it, it appears as if the class is completely redundant, since removing it from `messageHeaderAsHtml.js` doesn't change anything.

> Would you split that change out as a separate first commit?

I'll do that, unfortunately not today though, since there's not enough time.

ishammahajan:
Closing due to #3496.

PR #3437, opened 2019-04-01
Improve Flow typing of 'navigation' props, rework how params are extracted

borisyankov:
Prerequisite for #3422

More details in each commit.

* connect: Use `navigation.getParam()` instead of `navigation.state.params`
*  Replace `navigation.state.params` with `getParam`
* flow: Use better types for `navigation` properties

### DIFF
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -53,5 +53,5 @@ class AccountDetailsScreen extends PureComponent<Props> {
 }
 
 export default connect((state: GlobalState, props) => ({
-  user: getAccountDetailsUserFromEmail(state, props.navigation.state.params.email),
+  user: getAccountDetailsUserFromEmail(state, props.navigation.getParam('email')),
 }))(AccountDetailsScreen);
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 import { View, StyleSheet } from 'react-native';
-import type { NavigationScreenProp } from 'react-navigation';
+import type { NavigationScreenProp, NavigationStateRoute } from 'react-navigation';
 import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 
 import type { Context, Narrow } from '../types';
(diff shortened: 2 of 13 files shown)


borisyankov:
@gnprice ready for review/discussion/merge.

gnprice:
Thanks @borisyankov ! These navigation types have certainly bugged me for a while.

I don't love the `getParam` style, because it seems like it'll be much harder to get proper types on than the `params` style. The latter style, it clearly *should* be possible in principle to describe the types of in Flow, even if we currently don't and perhaps the libdefs we have in `flow-typed/npm/` currently don't make it possible. Specifically, `params` is some particular object type, and in fact it corresponds (when everything's correct!) to the type of the actual object we pass in the corresponding action creator in `navActions.js`.

But the fact that `navigation.getParam('messageId')` in `EmojiPickerScreen` returns `number` while `navigation.getParam('message')` in `LightboxScreen` returns `Message`... expressing that fact, in the type system in a checkable way, just seems impossible.

Given that we don't currently have working type-checking on these, I could see making the change, knowing that we'd end up changing back if we later see an opportunity to get working type-checking. But I'd want to better understand what we're gaining from the change.

Re `RealmScreen.js`:

     * we don't actually need to verify if `navigation` is defined,
       it always is, when the component is used as a navigation screen
     * we don't need to check if `navigation.state.param` is defined,
       as `getParam` will do that for us

What we do in most components is just directly get properties off of `props.navigation.state.params`. If we've somehow reached the component without appropriate navigation, that can blow up... but the component is already written to assume that we get there in a certain way, and if that assumption fails then it's best to fail early anyway rather than get into the interior of the component's code and fail in some unpredictable way midway through. So we can just simplify to
```
  initialRealm: props.navigation.state.params.realm || '',
```
independent of this change.

gnprice:
BTW as it happens I did some work on these types last week, as part of my cleaning up of most of the remaining `Object` types. See https://github.com/zulip/zulip-mobile/compare/master...gnprice:nav-params-types for the unpolished rough branch -- before ending the day I cleaned up the other changes in the branch and merged them (as dd03939cb^..d298669ef), but didn't get to those.

If it'll help with #3422, I can go back in the next day or two and finish and merge those changes. What was the issue you were seeing?

gnprice:
As promised above, a couple of weeks ago I went and made a bunch of cleanups and fixes to the navigation types; see 05bfc6c4e^..729386058.

I think that resolves what this PR was aiming for, and without sacrificing any type-checking (as `getParam` does); so closing. If there's an aspect of this PR you think remains useful, please go ahead and open a new one atop master. Thanks again!

PR #3442, opened 2019-04-03
Add Flow types required for RN 0.59 migration

borisyankov:
This PR adds Flow types that if missing, result in errors in the React Native 0.59 upgrade PR (not because of RN but because of Flow 0.92)

Can be merged separately. Prerequisite for #3422

### DIFF
--- a/src/autocomplete/PeopleAutocomplete.js
+++ b/src/autocomplete/PeopleAutocomplete.js
@@ -28,8 +28,12 @@ class PeopleAutocomplete extends PureComponent<Props> {
     this.props.onAutocomplete(`*${name}*`);
   };
 
-  handleUserItemAutocomplete = ({ fullName }): void => {
-    this.props.onAutocomplete(`**${fullName}**`);
+  handleUserItemAutocomplete = (email: string): void => {
+    const { users, onAutocomplete } = this.props;
+    const user = users.find(x => x.email === email);
+    if (user) {
+      onAutocomplete(`**${user.full_name}**`);
+    }
   };
 
   render() {
--- a/src/boot/AppEventHandlers.js
+++ b/src/boot/AppEventHandlers.js
@@ -6,7 +6,7 @@ import SafeArea from 'react-native-safe-area';
 import Orientation from 'react-native-orientation';
 
 import type { Node as React$Node } from 'react';
-import type { Dispatch, Orientation as OrientationT } from '../types';
+import type { Dispatch, Orientation as OrientationT, ConnectionType } from '../types';
 import { connect } from '../react-redux';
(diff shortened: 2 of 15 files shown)


borisyankov:
@gnprice this should be ready to merge independently of RN 0.59

gnprice:
Thanks @borisyankov ! These look like good improvements in general, in addition to being required by the newer Flow :smile: 

Small comments:
* Some lint errors; should fix.
* Several of these are writing down a type that comes from RN upstream. I think it's important for the reader to see what bit of upstream API they're intended to correspond to and what source you were consulting to write them -- both to understand the code, and also to make it easy to reproduce your work to see if e.g. the upstream API has changed a year from now, and apply updates.
  * Each change already provides this information quite nicely in the commit messages -- thanks! Particularly with the links to upstream docs. This is a kind of information that I think actually belongs in the tree itself, in jsdoc -- it's important information even for understanding the new code as it is, not only understanding how it compares to the old code. Would you move it in there?
  * The `ConnectionType` is one that does this well already. (Except I guess a smaller nit: since the link is there in the comment, I'd leave it out of the commit message.)

---

Not a blocker to merge, but a followup-thought that this brought to mind: where we're writing types that really belong to RN, or to another third-party library, the ideal place for them would actually be in a libdef, rather than inside our app code. We generally haven't done that because we get most of our libdefs from the `flow-typed` command/repo, and we haven't developed a workflow for making changes to that, so there's a lot of friction. I think it may not actually be hard to start doing it, though; I might sit down and try it sometime.

gnprice:
Thanks again @borisyankov !

You mentioned this PR in an audio conversation the other day, so I came back to look at it. It looks like you've resolved the lint errors; but my other comment is still outstanding:

> Several of these are writing down a type that comes from RN upstream. I think it's important for the reader to see what bit of upstream API they're intended to correspond to and what source you were consulting to write them [...]

Would you please address that? Looking forward to getting to merge this :slightly_smiling_face:

gnprice:
Also, I'll just add a link here to make this PR discoverable directly from the issue: this is part of #3399 .

borisyankov:
This is ready for review and merge, separately from the RN 0.59 PR.

The changes are mostly minor commit message improvements but also this commit is new:
`UserItem and related components: onPress now takes just 'email'`

gnprice:
Thanks for these revisions.  The new commit looks great.

In the places I requested changes, I'm disappointed because it looks as if you didn't really read what I said or make much effort to respond.  This matters to me because it means I feel like I wasn't heard.  I'll follow up with you separately.

In this instance, the requested changes are pretty straightforward to make; so I'll go ahead and merge, perhaps with some adjustments.

gnprice:
And merged! Thanks again.

Issue #3433, opened 2019-03-29
Make and post updated screenshots (2019)


gnprice:
The existing screenshots we use on the Zulip website and in the app stores are a couple of years old (some of them older). They look different from the current app, and are also visibly dated because they show contemporary system UI widgets and device bezels. It'd be great to replace them with fresh shots.

Subtasks:
- [ ] One Android shot for https://zulipchat.com/apps/android
- [ ] Several Android shots for https://play.google.com/store/apps/details?id=com.zulipmobile , showing different features (you can use the existing screenshots as a guide)
- [ ] One iOS shot for https://zulipchat.com/apps/ios
- [ ] Several iOS shots for https://itunes.apple.com/us/app/zulip/id1203036395 , showing different features (you can use the existing screenshots as a guide)

Points to have in mind:
* Make sure to check with the authors of any messages included in the screenshots and get their OKs. Even though the messages are in principle publicly visible if they're on a public stream on a server like chat.zulip.org, there's a real difference in visibility when posted in these screenshots, so people may feel differently.
  * To make that easier, you might look for a shot where all of the people in the frame are longtime and/or active contributors. They're likely to be easy to reach, and are more likely to feel good about appearing in one of these screenshots.
* On Android you can use ["System UI demo mode"](https://developer.android.com/studio/debug/dev-options) to get a generic boring status bar for a slightly cleaner screenshot.
  * For iOS, it looks like there isn't a system feature for this. There are a couple of suggestions in [this post](https://congsun.wordpress.com/2015/07/20/take-perfect-screenshots-before-submit-your-app-to-the-app-store/)... but if trying to do that takes more than like an hour, just skip it and let the status bar be in its natural state.

alexandraciobica:
Hello! I'll do this :)

schezfaz:
hey @alexandraciobica ! If you aren't working on it, let me know. I'd like to work on this as well!

alexandraciobica:
Hey! I am doing it. I had to wait for the people in the screenshots to answer if they are ok with it. I already took all the screenshots.

alexandraciobica:
I made the PR for /apps and I also made screenshots for Android, iPhone, iPad Pro and iPad3. 

I used the system UI demo mode for Android and the quicktime hack for iphone and iPad Pro. For the iPad3 it didn't work but I tried to make the top bar as clean as possible. I also asked all the people in the screenshots, only Neil Pilgrim didn't answer. 

[...]

Issue #3322, opened 2019-01-31
Add screenshots/GIFs to README.md


armaanahluwalia:
I think it would be helpful for new users to have screenshots/GIFs on the main README file to demonstrate how the app looks and behaves. This gives new/potential users a quick glance at what they would get if they went through the signup/download/install process.

devbihari:
let me handle this one. :fire:

PR #3443, opened 2019-04-04
readme: Add screenshots to readme.

alexandraciobica:
Fixes: 
- https://github.com/zulip/zulip-mobile/issues/3322
- which is also the last point from this one https://github.com/zulip/zulip-mobile/issues/3433

Added screenshots of the ios application to the README.md. I chose to add only these ones because the Android ones look the same, except from the toggles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@
 Zulip Mobile is the official mobile Zulip client, supporting both iOS
 and Android.
 
+![ios-screenshots](https://user-images.githubusercontent.com/18381652/55577814-b4746480-5714-11e9-8d92-5631922d22e3.png)
+
+![night-mode](https://user-images.githubusercontent.com/18381652/55577813-b4746480-5714-11e9-85a8-a1124d295231.png)
+
 ## Get the app
 
 Release versions of the app are available here:


gnprice:
Thanks @alexandraciobica ! These look great.

For the README, a few comments:
* Let's keep it to one row, of two or three screenshots.
* I think one of the unreads screen, and one of the message list, will be good. We can skip the "All messages" message list (that's a less commonly used feature) and the settings screen.
  * One of those can be in light mode and one in dark mode.
* I think I'd prefer to put the image(s) themselves in the repo, so it's self-contained.
* Can you experiment with different compression settings and try to make the images smaller? One of them is currently 1.9 MB; our whole tree is 6.9 MB. This is mostly solid colors and some text (no big detailed photos), so I'd hope it can compress a lot better without losing much image quality. If we can't manage that, though, maybe it'll be best to leave the images outside the repo after all.

alexandraciobica:
I managed to reduce the file sized to 96kb and 77kb respectively. And I think the quality is good enough too. I added them below so you can see the quality.

Where should I add them in the repo? Directly in the root, where also the readme is?
@gnprice 
1_small
6small

chrisbobbe:
Thanks, @alexandraciobica! I'm closing this because the app looks different now; for example, we have a different primary color (#4544).

Issue #3436, opened 2019-04-01
Provide "reply with mention" option


gnprice:
I.e., reply to message with an @-mention of its author -- what the `@` shortcut does in the webapp. This is handy in user-support kinds of use cases, particularly in an org like chat.zulip.org where there are tons of users, many with similar names, and so finding the author in autocomplete isn't always a great option. @timabbott has requested this a few times... as someone who frequently makes support-type replies on chat.zulip.org :wink: 

A simple version of this would be as yet another option in the long-press menu. Maybe only on messages you didn't send -- that will help slightly because the menu can be longer on your own messages (with delete and edit).

Spitballing, a fancier version would be to turn "Reply" into "Reply...", and have it pull up a *second* menu with options like "Reply", "Reply @${sender.full_name}", "Reply quoting message" (see #1975). But that might be best done separately after merging a first version of this.

ungps:
Hello,

I was working on adding an option to reply & quote, but I got stuck with 3 homework assignments at university and I couldn't focus anymore. Anyway, I will submit a PR for this. I would like to work on this.

Currently, selecting "Reply" opens up the thread. I guess the behaviour of "Reply with mention" should be almost the same (open up a narrow and autocomplete the textfield with @sender.full_name).

@zulipbot claim

sumukhah:
@gnprice, @ungps 
can I work on this?

chrisbobbe:
@gnprice [said](https://github.com/zulip/zulip-mobile/issues/3436#issue-427855536):
> Spitballing, a fancier version would be to turn "Reply" into "Reply...", and have it pull up a second menu with options like "Reply", "Reply @${sender.full_name}", "Reply quoting message" (see #1975). But that might be best done separately after merging a first version of this.

Makes sense to start simpler, and I don't think the simpler version will lengthen the path to the fancier version.

AlaaElden98:
Any update on this ?

chrisbobbe:
> Any update on this ?

You're welcome to take this issue, if you'd like. There's been some work toward this already, in #4190, rebase it, and grab what's useful from there. 🙂

gnprice:
> You're welcome to take this issue, if you'd like. There's been some work toward this already, in #4190, rebase it, and grab what's useful from there.

See also this thread with more on how to do that appropriately:

https://chat.zulip.org/#narrow/stream/48-mobile/topic/taking.20up.20abandoned.20issues/near/1156928

[...]

PR #3444, opened 2019-04-05
message list: Provide "reply with mention" option.

ungps:
Add the option to "reply with mention" in the long press menu.
The option is available only on other people's messages.

#3436


src/message/messageActionSheet.js
```diff
@@ -39,6 +45,13 @@ const reply = ({ message, dispatch, auth }) => {
 };
 reply.title = 'Reply';
 
+const replyWithMention = ({ message, dispatch, auth }) => {
+  const narrow = getNarrowFromMessage(message, auth.email);
+  dispatch(doNarrow(narrow, message.id));
+  dispatch(draftUpdate(narrow, `@**${message.sender_full_name}** `));
+};
+replyWithMention.title = 'Reply @%user%';
```
ishammahajan: How does `Reply to @user` sound here?

src/message/messageActionSheet.js
```diff
@@ -230,9 +247,10 @@ export const showActionSheet = (
       ...params,
     });
   };
+  const decorateTitle = title => title.replace('%user%', params.message.sender_full_name);
```
ishammahajan: This would be handled better where the title is assigned, for the sake of clarity.

src/message/messageActionSheet.js
```diff
@@ -230,9 +247,10 @@ export const showActionSheet = (
...
   showActionSheetWithOptions(
     {
-      options: optionCodes.map(code => _(allButtons[code].title)),
+      options: optionCodes.map(code => decorateTitle(_(allButtons[code].title))),
```
ishammahajan: Same as above, to use the function for every title where it is only applicable to one seems unwise, don't you think?

src/message/messageActionSheet.js
borisyankov: Sounds way better than without the 'to' :)
But I don't think we need the '@' in either case.

ungps: Indeed, it sound better. Thanks!

src/message/messageActionSheet.js
ungps: I agree with you here. The most elegant solution here is to use LocalizableString. The problem is that this must be a string and `_` does not replace variables either.

src/message/messageActionSheet.js
ungps: I agree with you here. The most elegant solution here is to use LocalizableString. The problem is that this must be a string and `_` does not replace variables either.

src/message/messageActionSheet.js
```diff
@@ -39,6 +45,13 @@ const reply = ({ message, dispatch, auth }) => {
 };
 reply.title = 'Reply';
 
+const replyWithMention = ({ message, dispatch, auth }) => {
+  const narrow = getNarrowFromMessage(message, auth.email);
+  dispatch(doNarrow(narrow, message.id));
+  dispatch(draftUpdate(narrow, `@**${message.sender_full_name}** `));
```
gnprice: Hmm, this looks fragile if it works at all -- we narrow, and only *then* we set the draft? Consider how and when the draft gets used.

Also, how does this behave in the case where we're already in the relevant narrow?

PR #3445, opened 2019-04-07
Fix a message loading bug and improve behavior

### LINKED ISSUE #3441: Messages sent while closed are missing and stay missing
Very nicely detailed report in chat here:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/android.20app.20show.20new.20messages/near/727083
(This post is a bit terse as I'm running out the door -- but there are plenty of good details at the link.)

That user reports it regularly happens to them. On v24.0.113 on Android, and previously on v23.3.112. Using their own server, v2.0.0, though I'd guess that isn't a factor.

Marking as a priority because this is a pretty bad class of bug, and it seems possible this user is one of many who regularly see it. If we determine it's really widespread, it'll belong at P0.

Cf previous issue #3284, and other issues and PRs with label `area: realtime`.

borisyankov:
The behavior of 'needFetchAtFirstUnread' is quite wrong.
First, we do the minimal work to fix #3441 in the first commit.
The rest of the work implements correctly the originally intended behavior.

### DIFF
--- a/src/message/__tests__/fetchActions-test.js
+++ b/src/message/__tests__/fetchActions-test.js
@@ -1,6 +1,6 @@
 import mockStore from 'redux-mock-store'; // eslint-disable-line
 
-import { fetchMessages, fetchMessagesAtFirstUnread, fetchOlder, fetchNewer } from '../fetchActions';
+import { fetchMessages, fetchOlder, fetchNewer } from '../fetchActions';
 import { streamNarrow, HOME_NARROW, HOME_NARROW_STR } from '../../utils/narrow';
 import { navStateWithNarrow } from '../../utils/testHelpers';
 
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @borisyankov ! This indeed looks like an effective fix.

The PR summary says later commits implement the intended "do we need to fetch" behavior. I see some welcome simplifications which might make it easier to do that, but I don't see the actual new logic in the current branch. Are there more changes you intended to push?

The last commit moves the function to a selectors file. I'm not sure that's actually helpful; perhaps it'll be clearer one way or the other in the presence of the change that adds back more logic. It's true that the function is a pure function, one of whose arguments is the state... but semantically it feels to me like part of the decision-making of the "fetch messages" logic here in `messages/fetchActions`.

I.e., the choices we've made that are embedded in it are choices about when and what we fetch (and the other code with those choices is in this file), and not about what data structures we maintain (which is the set of choices that are embedded in selectors and reducers).

gnprice:
Merged, except that last commit (moving to a selectors module), and with a quick commit on top adding a comment. Thanks again!

If you already have in your head (or on your disk) a clear idea of what the more sophisticated logic ought to be, I'd be glad to see that in a further PR.

PR #3448, opened 2019-04-08
stream list: Prevent creating streams with the same name

### LINKED ISSUE #3447: Trying to create stream with a name in-use does not display warning
If someone tries to create a stream with a name that is already in-use, the user is not warned that the stream already exists and will be redirected back to stream list.

@zulipbot  claim

ungps:
Display toast message when user tries to create a stream with the
same as an already existing one.

Fix #3447.


ungps:
I am not sure whether this is the best approach. Instead of showing a toast message, we could display an error above the "Create" button. This is the approach followed in the realm screen when we are trying to connect with an invalid organisation URL (like "a.zulipchat.com").

ungps:
I force pushed some changes because:

* I found it hard to spot the toast message when the keyboard is up. I think a red error message is more suitable in this case.

* The warning should also appear when we try to edit a stream, not only when we create a stream.

gnprice:
Thanks @ungps for sending this!

A few comments:
* I agree that a red error message sounds better than a toast.
* For a PR like this where UI is an important part, please post screenshots in the PR thread; they help a lot for discussion/review.
* I think adding a separate property to the state isn't the preferred React-idiomatic way to do this.
  * One UI consequence is that this version gives an error only when you go hit the submit button -- potentially after typing a bunch of description. Better would be when you blur the name field, or maybe even while you're typing (so in `handleNameChange`).
  * This post looks to have a pretty good description of what I think is a more preferred way to handle this: https://medium.freecodecamp.org/how-to-use-reacts-controlled-inputs-for-instant-form-field-validation-b1c7b033527e
* Does this behave right when editing an existing stream? I think it may not. Be sure to test and think about that case too (which also uses this code, as its name suggests.)

ungps:
Thank you @gnprice for review. I pushed a new version. 

 * In both create / edit stream page, the error message is not displayed and the button is disabled when the "Name" field is empty.
 * In edit stream page, the error message is not displayed and the button is disabled when name field wasn't modified.  If description, access (private / public) or the case of the stream name was modified, the button is enabled (the error message still hidden)

Previews:

https://drive.google.com/file/d/13gJJwZvAV-5FLo6PETmw5vruVa-KDzXZ/view?usp=sharing

https://drive.google.com/file/d/16ayxOeDiLQ7Y1H2IUMhUYRmDJvNyuddU/view?usp=sharing

src/streams/EditStreamCard.js
```diff
@@ -47,8 +53,12 @@ export default class EditStreamCard extends PureComponent<Props, State> {
   };
 
   render() {
-    const { initialValues, isNewStream } = this.props;
-    const { name } = this.state;
+    const { initialValues, isNewStream, streams } = this.props;
+    const { name, description, isPrivate } = this.state;
+
+    const isInvalidStreamName =
+      caseInsensitiveCompareFunc(name, initialValues.name) !== 0
```
borisyankov: Nice that you are reusing a function but this seems a bit convolutrd.
hat would read better as:
```js
name.toLowerCase() === initialValues.name.toLowerCase()
```

That being said, I am not sure this part is even needed.
On a new stream creation, you are comparing with `''`, and on stream edit, you are comparing with stream's previous name which should be valid.

src/streams/EditStreamCard.js
```diff
@@ -47,8 +53,12 @@ export default class EditStreamCard extends PureComponent<Props, State> {
   };
 
   render() {
-    const { initialValues, isNewStream } = this.props;
-    const { name } = this.state;
+    const { initialValues, isNewStream, streams } = this.props;
+    const { name, description, isPrivate } = this.state;
+
+    const isInvalidStreamName =
```
borisyankov: Often I prefer the positive variant `isValid` as I find it reads better and we don't have to do the double negative to achieve valid `isValid` vs `!isInvalid`.

src/streams/EditStreamCard.js
```diff
@@ -75,10 +86,18 @@ export default class EditStreamCard extends PureComponent<Props, State> {
         <ZulipButton
           style={styles.marginTop}
           text={isNewStream ? 'Create' : 'Update'}
-          disabled={name.length === 0}
+          disabled={
+            name.length === 0
+            || isMatch(initialValues, { name, description, invite_only: isPrivate })
```
borisyankov: This looks like a lot of logic put in a property value assignment. Would be better as a flag near the `isInvalidStreamName` with a good name like `canUpdateStream` or any other appropriate and helpful name.

borisyankov: `isMatch` looks like overcomplicating code that can be implemented with few `===`s.
While 'lodash' is popular and we do not shy away from using it, I definitely had to go an read the documentation to be able to understand if the code is correct.

This code looks simpler and more exact as:
````js
initialValues.name === name 
&& initialValues. description === description  
&& initialValues. invite_only === isPrivate
```

src/streams/EditStreamCard.js
ungps: Thank you @borisyankov for review.

Indeed, this condition is a little bit counterintuitive, but without it, we would not allow user to change stream name case (ie from "New_Stream" to "NEW_STREAM"). Moreover, the error message would be displayed even if the user did not change stream name field.

src/streams/EditStreamCard.js
ungps: I changed it from negative to positive logic.

src/streams/EditStreamCard.js
ungps: Makes sense. I pushed a new version. Thank you!

src/streams/EditStreamCard.js
```diff
@@ -75,10 +90,14 @@ export default class EditStreamCard extends PureComponent<Props, State> {
         <ZulipButton
           style={styles.marginTop}
           text={isNewStream ? 'Create' : 'Update'}
-          disabled={name.length === 0}
+          disabled={canUpdateStream}
```
borisyankov: You meant to do a negative here, right?
`disabled={!canUpdateStream}`

And probably the logic would need changing then?

ungps: Right. Sorry. I force pushed a new version.

gnprice:
Thanks again @ungps !

> In edit stream page, the error message is not displayed and the button is disabled when name field wasn't modified. If description, access (private / public) or the case of the stream name was modified, the button is enabled (the error message still hidden)

* This sounds like a basically independent change, right? Let's make this a separate commit, either before or after.

  Then that change can also be explained in its own commit message. Everything we change in a commit should be covered by something described in the commit message -- you could satisfy that by adding a mention of this change to the one commit message, but it's easier to understand if it's separated out.

* Nit: Let's use a name like `canSubmit` rather than `canUpdateStream`. The latter sounds like it might be referring to the contrast between "Update" and "Create", which after all appears in the user text on the very same button.

* Thanks for the videos -- but would you also post an actual screenshot, right here in the thread (so the image shows up right here on the GitHub PR page)? That makes it a lot easier to see and discuss. (In fact, one of the two videos doesn't even load for me.)

  In particular I'd like to see a screenshot of the error message -- that's a new bit of design introduced by this PR, so we'll want it to be part of the discussion.

ungps:
Thank you too, @gnprice ! I pushed new changes to

* split the commit into two commits
* Changed `canUpdateStream` to `canSubmit` because is more suitable.
* I am pretty crowded with homework assignments at university and I couldn't keep up with #3450; but I hope I fixed merge issues correctly.    

Below are a couple of screenshots (with caption):

Creating new stream - name field not filled (no error message, button disabled)


Creating new stream - stream name already in-use (display error message, disable button)


Editing stream - no field changed (no error message, button disabled)


Editing stream - name changed in upper case (no error message, button enabled)


Editing stream - stream name ("announce") already in use (error displayed, button disabled)
In both night mode / day mode:

gnprice:
Thanks @ungps for those screenshots!

One thing they show is that this layout is confusing:
image

The error is much too far from the field it's about; it's actually closer to the "Description" field below, so it looks like it's about that.

For some pictures of good design for this situation, search for "error" in the Material guidelines:
https://material.io/design/components/text-fields.html#anatomy
Our text fields aren't Material Design in general (although they'd probably be better if they were); but that may help illustrate something to move toward.

ungps:
Thanks again, @gnprice!

Indeed, the message was way too far from the field. I hope this is better now. I also changed the colour of underline in case of error.

Screenshot_20190427-181058
Screenshot_20190427-181119
Screenshot_20190427-181331
Screenshot_20190427-181630

PR #3450, opened 2019-04-11
Fix type of `connect` to fully work, and to work with newer Flow.

gnprice:
This is part of #3399. See also discussion in #3422 and #3428 .

The new RN version (v0.59) comes with a new Flow version (v0.92). That Flow version, in turn, comes with [more consistent enforcement](https://medium.com/flow-type/asking-for-required-annotations-64d4f9c1edf8) of an annotation requirement that (a) on the one hand means fewer bogus "Santa Claus" types ("everything the code wants to be true, is true!", aka `empty`) but (b) means a lot of boilerplate when using `connect`, if relying on the libdef for it in `flow-typed` upstream.

Moreover, the new upstream libdef for `connect` doesn't understand `defaultProps`, causing a lot of spurious type errors.

So with this branch, we bring in the new libdef... but then add our own wrapper around it, with an utterly trivial implementation -- if you erase all the types, it looks like this:
```
export function connect(mapStateToProps) {
  return connectInner(mapStateToProps);
}
```
but with a bunch of complexity in the types to describe what `connect` really does, plus about 500 words of comments.

Also marks with a fixme the remaining places where we have type errors hidden by the ineffective old type on `connect`. I fixed a bunch of such errors last week and the week before, before I realized how much work was needed on the type of `connect` itself; for the rest, this just marks them so we can move on and fix them separately.

---

Posting this as a PR because (a) I'd like to shift to more of a practice of posting my own changes for discussion before merge, and (b) this change in particular deserves some discussion, because I want everyone to be aware of it so there are no surprises when Flow complains about a type error on `connect` in the future.

### DIFF
--- a/flow-typed/npm/react-redux_v5.x.x.js
+++ b/flow-typed/npm/react-redux_v5.x.x.js
@@ -1,40 +1,27 @@
-// flow-typed signature: b872fdc4da6b0c1021c55df6fab87e73
-// flow-typed version: f8afc4cfdd/react-redux_v5.x.x/flow_>=v0.68.0 <=v0.84.x
+// flow-typed signature: f06f00c3ad0cfedb90c0c6de04b219f3
+// flow-typed version: 3a6d556e4b/react-redux_v5.x.x/flow_>=v0.89.x
 
-declare module "react-redux" {
-  import type { ComponentType, ElementConfig } from 'react';
+/**
+The order of type arguments for connect() is as follows:
 
-  // These types are copied directly from the redux libdef. Importing them in
-  // this libdef causes a loss in type coverage.
-  declare type DispatchAPI<A> = (action: A) => A;
-  declare type Dispatch<A: { type: $Subtype<string> }> = DispatchAPI<A>;
-  declare type Reducer<S, A> = (state: S | void, action: A) => S;
-  declare type Store<S, A, D = Dispatch<A>> = {
-    dispatch: D;
-    getState(): S;
-    subscribe(listener: () => void): () => void;
-    replaceReducer(nextReducer: Reducer<S, A>): void
-  };
+connect<Props, OwnProps, StateProps, DispatchProps, State, Dispatch>(…)
 
-  declare export class Provider<S, A, D> extends React$Component<{
-    store: Store<S, A, D>,
-    children?: any
-  }> {}
+In Flow v0.89 only the first two are mandatory to specify. Other 4 can be repaced with the new awesome type placeholder:
 
-  declare export function createProvider(
-    storeKey?: string,
-    subKey?: string
-  ): Provider<*, *, *>;
+connect<Props, OwnProps, _, _, _, _>(…)
 
-  /*
+But beware, in case of weird type errors somewhere in random places
+just type everything and get to a green field and only then try to
+remove the definitions you see bogus.
 
+Decrypting the abbreviations:
+  WC = Component being wrapped
   S = State
-  A = Action
+  D = Dispatch
   OP = OwnProps
   SP = StateProps
   DP = DispatchProps
   MP = Merge props
-  MDP = Map dispatch to props object
   RSP = Returned state props
   RDP = Returned dispatch props
   RMP = Returned merge props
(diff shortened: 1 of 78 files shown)


gnprice:
Also filed #3451 for the followup task of fixing the newly-exposed errors.

gnprice:
Ah, I forgot there is one type of boilerplate that's still required for the new Flow version. It's rather milder than some of the alternatives, though. Specifically:
* When `mapStateToProps` actually takes a `props` parameter,
* then its *return* type needs to be annotated.

For example:
```
type SelectorProps = {|
  backgroundColor: string,
|};

type Props = {|
  narrow: Narrow,
  dispatch: Dispatch,
  ...SelectorProps,
|};

class ChatNavBar extends PureComponent<Props> {
/* ... */
}

export default connect((state, props): SelectorProps => ({
  backgroundColor: getTitleBackgroundColor(props.narrow)(state),
}))(ChatNavBar);
```

Even this annotation seems like it shouldn't fundamentally be necessary; but at least for now, and with Flow v0.92, I didn't succeed in finding a way to get Flow to infer this.

gnprice:
Thanks @borisyankov for the brief discussion in audio chat the other day. Merged, after adjusting the main commit's message for the fact I mentioned above, that we do need a few annotations even with this version.

gnprice:
BTW, just to help me close a tab with confidence -- [here's](https://flow.org/try/#0C4TwDgpgBAglC8UAkAlCBDAJgeQHYBsQAeAbwB8oAPAfgC4pcBXAWwCMIAnAGihHqbaceALzpQAzsA4BLXAHMoZAL4A+ANwAoDaEhQAIuITI90gGamiMHuSr1JM+YtWaA9C6gBjAPa5JUTOL0BkYkvPQAjFBKmtrg0CiGiKgQkpbWFJR2UrIKyuoabp4+fhyBUAkhYVCR0Vo60HrMRkgm5mlQNplQzNKUEJhO+d6+wP7MQU2IoXzVUTH1UACS4gDKjJAc9UQrPAAq9CsqRiuaC8trG-UATNt7B0eIu6dxS6uMrFs7ULsPrxec1yIux4h2eumWAFEAI6MdD4W7fe5Gc7rAFxIEg-KxXQADWRb0u6IE7G4UAgzDAoCGxVGXTxiAAzPMXgBNfFrD7o5JYPCEUgZMTEoRVHp9AZ5HgwfKYCAefDoDjQABuCqqLLUUEKwAAFtJDOIANboYApKCsEAaYZ+EAzNmIEAawroXADCBKiC4KA6vUFdxW0Y220hGYM6Ka9ysRijb2GDzoRjiU3OskcDheUkgdAW7HQABa7NRm3RVmQaB5BGIJC6Qo4qnyhX9UGE9HzUyqNRihUjo3EWEItBzUAA6tIdeEEQAFX4nS00qjN4ej7Xjpj4fA8SEwuFEew5Hg1lS-Vf4TsRqO8LMAQkHI51GO+4XoPyMu3Crj9c4A7vRb9qiMeN3EaFYXhACyQpKkj0YNdmV0AAhOQvC8AYpgoGYaxEMRd0cZQwWgODWCwEI0PoagMKbLDshw2oFjg5h02gVCqgZTCsgcXIaJeOivD8JiZnJSkQFYiQqI408oAAChIGZwiUegUU5SAiG5HAK35KAADotJLLSNIIoiJVgFQAEpNCkmS5L+d4thU3lKwoXSdK0hCkPFVRJRM2DoAhfBgE-PUIBuL5n0QBSbLLVS+RsXSvl03ZBkxMzpIiSyfL8gKbhLFzkM831JOS6pUt8-zE0ynh9MwXLBzSkqIAZBEQqswklK+Wy1OirTYq0+K8nycyUvoGqAvqrLEJy0y8v6wrBuK4b2gqqqpIXABydBlsslozAsGx0JYElhOwjiuA0ZBsFYAArABZdAwHU-iIJAJweAk4yECOUV+kPCbBzg1oLAAVR4PRfk2tohsTerAf0FQeCQM6rpuog9Ge174He3pPvrFwlvoVb1voX6tvmwjKomnGoDxyzCbaUbXNy8nKYJv75voxV6ZIFa1qp5msvoyQqsHCdIlCgk0SUpAfPJD1gF2YseGWkBlphk7kEl5hpdlpSKvlxXD3fIoRigMBHygIWjDfLRBauZp4eusBpCIbWVagIgAGkeAANRUCT3agD3UaOML0Q92G1Y1uWoFdw9nagKAsYNvwwCuegJ2ttsZJEXG1rmOoXlThQkltm6HeyzBjtjt3Pe933-beqzFIgIgQ9V-ApdwGWI6jlQY9j6lDaTuQU6uAuOnbTOKez2pB0hVv1fb2qOU+PYHhVuGLrth2dhjyu-erz2A-rrZm4l2fw6U4FI+j3u8NNhl8TD+eAsX9FtaM-XGzABkU7vqYLJvicGQjxnm3dKiZn5azGmXN+s5+6AO-iPAqsl35zk-nIVg8Dgz8GgvgKeMo5QKmVKqT+mB5JAVPo-MB1kX6QI8vrCSxCNL3UEhNWOrC2HsMKPkdh3DWE9x4bHIAA) a Flow playground link with some of my experiments that led to the type-wrapper in f8c981028. Most of the ideas wound up in f8c981028, and with much more explanation than I wrote down in the experiments; but there are a few ideas there that didn't end up being necessary.

gnprice:
> Even this annotation seems like it shouldn't fundamentally be necessary; but at least for now, and with Flow v0.92, I didn't succeed in finding a way to get Flow to infer this.

FTR based on some quick experiments just now, I can't actually reproduce why this appeared to be necessary!

There is #3768, which is a Flow unsoundness bug (that we just recently spotted) affecting these places in our code; but adding `SelectorProps` doesn't seem to do any good with that. And when I apply the silly `const y = () => <X />;` workaround described on that issue, everything seems to get inferred perfectly fine with no `SelectorProps` annotation.

PR #3453, opened 2019-04-12
More type fixes for RN v0.59 / Flow v0.92

gnprice:
This is another piece of #3399; sort of a continuation of #3450.

Basically I took a branch based on #3422, plus my changes in #3450, and looked at the type errors that remained. The errors that jumped out at me as either (a) potentially involving something subtle in the type system, or (b) being about code which has some subtlety and I remember writing, I investigated and fixed.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -23,7 +23,7 @@ export type Dimensions = {|
   top: number,
 |};
 
-export type InputSelectionType = {|
+export type InputSelection = {|
   +start: number,
   +end: number,
 |};
--- a/src/types.js
+++ b/src/types.js
@@ -24,8 +24,8 @@ export type Dimensions = {|
 |};
 
 export type InputSelectionType = {|
-  start: number,
-  end: number,
+  +start: number,
+  +end: number,
 |};
 
 /**
--- a/src/autocomplete/AutocompleteView.js
+++ b/src/autocomplete/AutocompleteView.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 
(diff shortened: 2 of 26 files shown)


src/types.js
```diff
@@ -24,8 +24,8 @@ export type Dimensions = {|
 |};
 
 export type InputSelectionType = {|
-  start: number,
-  end: number,
+  +start: number,
```
borisyankov: Duh, this is the most subtle thing a diff can show `-  start` vs `+  +start`.

gnprice: I'm not sure if that's a request to do something different, or just a lament about Flow's syntax. :slightly_smiling_face: 

I do think this is a good reason to keep this change separate from large other changes, or other changes to this chunk of code! As is, though, as the reader you know *something* changed in this block because there's a diff shown here... so it should get your attention, and then hopefully it isn't too hard to spot what.

One other tool that helps: `diff-highlight`. See my .gitconfig for how to enable it:
https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3
The tool itself is distributed with Git; the Debian package puts it in `/usr/share/doc/git/contrib/diff-highlight`, and in general `locate diff-highlight` should find it on your system. In cases like this, it causes the one small piece of each line that actually changed to get highlighted in `git diff`, `git log -p`, etc.

image

The next commit is actually a yet stronger example of where `diff-highlight` is super helpful -- every one of the diff hunks has the word `Type` in red, which helps me see at a glance that the exact same boring thing is happening in each one of them.

image

gnprice:
Merged. Thanks @borisyankov for taking a look!

Issue #3454, opened 2019-04-14
Diagnostics: Allow setting a custom nickname that we send to Sentry


borisyankov:
We often have problems identifying what bugs in Sentry match which user when reported by our customers.

The issue is privacy, we can't just log user emails or anything identifying.
On the other hand, phone models are not unique enough to help us with that.

This feature will live in the Diagnostics screen. It will allow us to set an optional custom username/nickname that will be send along with any Sentry error report.

PR #3459, opened 2019-04-16
Diagnostics nickname

borisyankov:
Implements #3454


borisyankov:
Ready for review.

gnprice:
Thanks @borisyankov ! This seems like a potentially useful feature.

* The main thing that makes me uneasy about this implementation is the going and sticking a hook in the central rehydrate flow, in `boot/StoreProvider`:
```
 export default class StoreHydrator extends PureComponent<Props> {
   componentDidMount() {
     timing.start('Store hydration');
-    restore(() => {
+    restore((err?: Error, state?: GlobalState) => {
       timing.end('Store hydration');
+      if (state) {
+        sentrySetUsername(state.settings.diagnosticNickname);
+      }
     });
   }
```
This puts into that very central and global bit of code some logic that encodes intimate details of this setting and how we have it interact with Sentry. Those details should be confined in one area close to each other, in `settings/settings*.js`.

A secondary, related, point is that it's awkward to have to try to cover all the places where we set out to modify this field and remember to put `sentrySetUsername` on each of them -- it'd be better to have that automatically get called when it changes.

One approach that I think would address both of these cleanly is to use [`store.subscribe`](https://redux.js.org/api/store). If nothing else, we could even have in `boot/store.js` a line like
```
store.subscribe(sentryListener);
```
where `sentryListener` contains those details, and is imported from a definition that's next to the other places they appear (in `settings/settings*.js`.)

* Small things in the new component:
```
+export default connect((state: GlobalState, props) => ({
```
(a) We no longer need to repeat `GlobalState`! Our `connect` type-wrapper encodes that.
(b) We do on the other hand need to write an explicit `SelectorProps`. Or rather, with the coming upgrade we'll need to, so we might as well keep up on new components.

* I'd squash the three Redux commits together, and the two new-UI commits together. I think in each case the changes are all closely enough connected that they're easier to understand in combination with each other.

PR #3461, opened 2019-04-23
WIP: Improve 'Scroll to bottom' button

borisyankov:


gnprice:
Thanks @borisyankov ! The use of `pointer-events` is a nice cleanup.

* There's a couple of different ways this code refers to the `a` element in the scroll-to-bottom widget -- in the CSS like so:
```
.scroll-bottom-link * {
  pointer-events: none;
}
#scroll-bottom a {
```
and in the JS with
```
+  if (target.matches('.scroll-bottom-link')) {
```
It takes a bit of close reading of the HTML to work out that those are just different ways of talking about the same element. Can we make them the same?

Specifically, how about dropping the `scroll-bottom-link` class entirely, and writing that CSS block like `#scroll-bottom a * { ... }` and the target test like `if (target.matches('#scroll-bottom a'))`?

(Or another style, but if so then applying that consistently to the other selectors too.)

* Ha, I see, in fact in the next commit you move toward more classes and less IDs. :slightly_smiling_face: OK, but again let's make it consistent among closely related elements: so `#scroll-bottom svg` shouldn't coexist with `.scroll-bottom-circle`. (It should at least be `.scroll-bottom-circle svg` so the relationship is evident -- plus they should be adjacent rather than separated by `.scroll-bottom-text`.)

* Separately, on that next commit: why is the padding in em? What's the resulting size of the touch target -- does it in fact reach the desired 44px? It's pretty hard to tell.

* Let's push it even a bit bigger, to 48px: https://material.io/design/usability/accessibility.html

borisyankov:
> It takes a bit of close reading of the HTML to work out that those are just different ways of talking about the same element. 

This PR actually removes the `#scroll-bottom a {` completely, so there aren't multiple ways to refer to that element.

> Specifically, how about dropping the scroll-bottom-link class entirely, and writing that CSS block like #scroll-bottom a * { ... } and the target test like if (target.matches('#scroll-bottom a'))?

That would be against modern best practices where the majority (all?) of selectors should just be against classes. For example, BEM (which we don't follow, but might, though our webview CSS is tiny):

http://getbem.com/naming/

> I see, in fact in the next commit you move toward more classes and less IDs.

As a note, for anyone that might be concerned about performance - selecting by class vs selecting by id is similar in performance. That might be a bit counter-intuitive for people thinking strictly about 'id's as unique identifiers that are accessed in an id-to-element map. Once you take into account that there could be multiple elements with the same ID (totally not a good practice, but also totally possible) then the implementation internally to the browser is likely similar.

> let's make it consistent among closely related elements: so #scroll-bottom svg shouldn't coexist with .scroll-bottom-circle.

Yeah, I think a separate commit that will improve upon that would make sense.

> Separately, on that next commit: why is the padding in em? What's the resulting size of the touch target -- does it in fact reach the desired 44px? It's pretty hard to tell.

Everything is padded in 'ems' in the `webview`. Not sure if we should change that, but it if we do, we would do it in a separate commit.

> Let's push it even a bit bigger, to 48px

Mission almost accomplished. We are now at 47px (0.5em * 2 + 32px)

gnprice:
Thanks for the replies!

I see you haven't pushed a new version of the branch, and from your reply I'm not sure if you plan to. To reiterate an essential point of our code standards, and of the purpose of code review: We want people to read and understand the commits we merge. When there's something in the commits I don't understand and I ask a question like "why", or "what exactly does this do", the answers ultimately need to be made clear in the actual commits (either by changes to the code, or comments, or info in commit messages.) So several points in your answers identify changes to make in the branch.

See also this comment, which also points toward making the code easier to understand:
> plus they should be adjacent rather than separated by `.scroll-bottom-text`.

gnprice:
> Everything is padded in 'ems' in the `webview`. Not sure if we should change that, but it if we do, we would do it in a separate commit.

Hmm -- not everything, but indeed most things are. (This is easy to check -- `git grep padding src/webview/`.)

*shrug* The essential thing for me is that it's clear what's going on -- both (a) it's clear in this commit that it actually meets the desired target size, and (b) it's clear in the code, for the next person making changes here, what constraint the value was intended to satisfy, so that it isn't completely a magic number that can't be touched for fear of unknown consequences. If you want to put it in em, then one way to satisfy both (a) and (b) would be a comment like `// want >=48px touch target; 0.6em = 9px, so 32px + 2 * 9px = 50px` .

But using em feels pretty indirect and brittle when we know our actual intention is expressed in px. Even without converting any of the existing uses elsewhere of em to use px, it seems like it'd be an improvement to make at least this one component be less brittle by just saying `8px`.

borisyankov:
The current implementation is absolutely the correct way to express the button size and margins. There is nothing wrong or brittle in mixing em/px/rem as long as it is intentional. Here it is.

* Pixel size for the button means it does not grow/shrink with the system's font size
* Padding in 'em's means that this does adapt with font size change

Expressing the padding with the same units and values as the other items mean that the button will not be misaligned.

This PR will change the padding to pixels #3188 and I will make a point of changing this padding to pixels too.

borisyankov:
I have changed the 'em' to 'px'. These commits will make sense right after #3188 is merged.

gnprice:
> I have changed the 'em' to 'px'. These commits will make sense right after #3188 is merged.

Sounds great! Looking now at #3188.

PR #3462, opened 2019-04-26
Cleanup SubscriptionCard

borisyankov:
Removes several lines of code that have had no effect for a very
long time (since we had a drawer in the app)

The purpose of the code being removed was to show as 'selected'
a subscription, if it was the currently active narrow (this
component was rendered in the drawer)
  
This component now only serves as a Tab in StreamTabs. It never
receives a value for the 'narrow' property, thus it remains the
default `[]`. This in turn always sets `selected` to `false`.

None of the removed code had any effect except cluttering the code.

### DIFF
--- a/src/streams/SubscriptionsCard.js
+++ b/src/streams/SubscriptionsCard.js
@@ -3,10 +3,10 @@
 import React, { PureComponent } from 'react';
 import { View, StyleSheet } from 'react-native';
 
-import type { Dispatch, Narrow, Subscription, GlobalState } from '../types';
+import type { Dispatch, Subscription, GlobalState } from '../types';
 import { connectFlowFixMe } from '../react-redux';
 import StreamList from './StreamList';
-import { isStreamNarrow, streamNarrow } from '../utils/narrow';
+import { streamNarrow } from '../utils/narrow';
 import { getUnreadByStream } from '../selectors';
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @borisyankov !

* Why does this leave the `narrow` property in `Props`? That's part of the dead code, right?

* I think you mean `StreamTabs`, not `MainTabs`.

gnprice:
Oh one more:
* The natural next question this raises is: can that `selected` prop be removed entirely?

borisyankov:
> * Why does this leave the narrow property in Props? That's part of the dead code, right?
> * I think you mean StreamTabs, not MainTabs.

Correct. Thanks for catching these issues. Fixed and pushed an update.

gnprice:
Thanks for those updates! Those two items look good.

Please take a look at the other comment I made above.

borisyankov:
> The natural next question this raises is: can that selected prop be removed entirely?

We can always do that but I prefer not to do it right now.

We have a few future tasks that would require it (admittedly not immediate ones).
For example, a view that shows which streams a user is subscribed to, or maybe if we move the stream list in a place that can be seen when still narrowed, we would also need the 'selected' functionality.

gnprice:
Thanks for the reply! I'll go ahead and merge this because it's clearly a step forward; then we can continue the discussion about that followup step.

PR #3463, opened 2019-04-27
WIP: Sparse array types

borisyankov:
PR intended for discussion

### DIFF
--- a/src/pm-conversations/pmConversationsSelectors.js
+++ b/src/pm-conversations/pmConversationsSelectors.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import { createSelector } from 'reselect';
 
-import type { Message, PmConversationData, Selector } from '../types';
+import type { Message, PmConversationData, Selector, SparseArray } from '../types';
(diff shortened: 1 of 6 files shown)


gnprice:
Hmm interesting. Can you say a few words about the motivation for this? My first reaction is that "array" seems like a confusing way to have to think about these data structures -- among other things there's no particular sequence to them.

borisyankov:
The initial motivation was fixing the discrepancy between few places that declared `{ [number]: ... }` but initialized it as `[]`...

Our usage was curiously independent of wether the value was object or an array.
After some time (yesterday at our call already) I decided things should go the other way around - the initialization with `[]` should be fixed to `{}`.

After some research, I do not think there is a good reason to use sparse arrays over objects. Likely the implementation internally is similar. This was an interesting read that I found regarding this topic:
https://news.qooxdoo.org/javascript-array-performance-oddities-characteristics-d8139757b238

gnprice:
Sounds good. Thanks for the update, and the link!

PR #3465, opened 2019-04-29
Add 'InjectedDispatch' type

borisyankov:


gnprice:
We discussed this in our video conversation yesterday; just to reflect that in quick notes here:

* There's a bit of cost to this -- for anyone looking at something like `...InjectedDispatch`, that's an indirection they have to go follow to find out what it means, where `dispatch: Dispatch` is more transparent. So I'll want a clear reason for the change.

* You mentioned it's part of refactoring all `Props` types to a combination of props from the Redux state, from the caller, etc. In that context if `dispatch: Dispatch` becomes the last thing that isn't a spread, I could see this change making sense. I think I wouldn't want to do it without that refactor.

* Jumping up to a meta level: looking at this PR without the extra context in our conversation, I wouldn't have had a clear idea what feedback you're looking for. (Right now the PR description is empty and the two commit messages are just a few words -- nothing wrong with that for a WIP, just means there isn't much I can say.)

  I think the best feedback I'd be able to give would be something like "Can you please say more about what this is aiming to do and why?"

borisyankov:
I am taking the feedback into account. This will be part of #3471
Closing this stand-alone PR.

PR #3468, opened 2019-05-01
tests: Add example-data helpers for Redux state, more users, and cross-realm bots

gnprice:
This builds on #3467 / 415f0546b, with the same purpose of helping us to write
unit tests that
 * pass genuine well-typed data,
 * and don't have a lot of repetitive boilerplate for the boring bits.

We have a lot of tests for selectors, and they generally need a value
of type `GlobalState` to pass to the selector.  This lets us start
doing that.

Also convert `userSelectors-test.js` entirely to the new form, for demonstration -- note the full type-checking and the highly negative diffstat.

My approach was to look for test files we've repeatedly made changes to in the last 6 months:
```
$ git log --after=2018-12-01 --format= --numstat -- src/**/__tests__/ | sort -k 3
```
and I found `userSelectors-test.js` was one of the leaders. So then I went about seeing what it would take to convert it to well-typed data, and doing that.

### DIFF
--- a/src/__tests__/exampleData.js
+++ b/src/__tests__/exampleData.js
@@ -2,7 +2,14 @@
 import deepFreeze from 'deep-freeze';
 import { createStore } from 'redux';
 
-import type { CrossRealmBot, Message, PmRecipientUser, Stream, User } from '../api/modelTypes';
+import type {
+  CrossRealmBot,
+  Message,
+  PmRecipientUser,
+  Stream,
+  Subscription,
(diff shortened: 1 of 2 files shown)


gnprice:
I merged the commits described in this PR description a while ago, as the series leading up to 307ff91e0a895a600358bfd775be950b940a6a97 .

Looks like I'd forgotten then that I'd made this PR attached to the branch. Closing it now.

PR #3470, opened 2019-05-02
Flow simplify convo list

borisyankov:
When adding the `InjectedDispatch` type in #3465 I found that we are not injecting it in `PmConversationList` but passing it from the parent.

This is a stand-alone PR though. Details and motivation in the respective commits:
* UnreadCards: Simplify implementation and types
* PmConversationList: Get 'dispatch' injected by 'connect'


gnprice:
Thanks @borisyankov !

These changes look generally good. Some comments/questions:
* In the first commit, we're actually changing the set of properties passed, right? Specifically, no longer passing `dispatch`. That isn't at all clear in the commit message -- it sounds like a message for a pure refactor. That should be mentioned, so you can explain why it's OK.
* In fact, is it OK? It looks like it should break the `PmConversationList` call. (Note there's a `$FlowFixMe` which knocks out all type-checking on the whole `SectionList` call, so Flow won't tell us about that.)
* In the second commit, the message says this is the only instance we still pass `dispatch` manually. How did you determine that? In a search, there look to be a handful more:
```
$ rg dispatch=
src/account-info/ProfileCard.js
86:          <SetStatusButton dispatch={this.props.dispatch} />
89:          <SwitchAccountButton dispatch={this.props.dispatch} />
90:          <LogoutButton dispatch={this.props.dispatch} />
```

borisyankov:
> In the second commit, the message says this is the only instance we still pass dispatch manually. How did you determine that? 

Hmm, by searching for `dispatch={dispatch}` which obviously missed that.

Looking at these components that you've folded in one file a while back, I think another commit here that would completely fold them into a single component makes sense. What do you think?

gnprice:
> Looking at these components that you've folded in one file a while back, I think another commit here that would completely fold them into a single component makes sense. What do you think?

Plausible!

I'm not sure it would simplify much, though. The `onPress` definitions wouldn't get any shorter, I think, and that's most of the complexity in these definitions.

And on the other side of the ledger, the `onPress` definitions would each get separated farther from the UI elements that use them, and also we'd have to make up names for them and the reader would have to cross-reference by the names to see what's going on. Not a huge cost, but may be more than the small benefit.

In a different direction, I'm curious what these can become using [hooks](https://reactjs.org/docs/hooks-intro.html) -- it seems like they might be candidates to become a bit terser and simpler that way. We'll get to use those once the RN v0.59 upgrade is done. I think taking advantage of that here would likely mean keeping them separate.

Issue #3451, opened 2019-04-11
Fix type errors hidden by `connect`


gnprice:
Historically the `connect` of `react-redux` has had a `flow-typed` libdef that largely defeats type-checking on it -- so e.g. a component can have a prop with a type that doesn't actually match the selector that supplies it, or a `mapStateToProps` callback can assume that its `props` argument has some property that is never actually specified.

With #3450 we get a version of `connect` that makes type-checking work. It turns out there are a number of errors out there that this was hiding! In addition to the many such errors that I fixed a week or two ago (accounting for most of the commits in cae19530f^..f89f317ff), there are 13 more files with these errors, now marked with `commitFlowFixMe`.

A workflow to fix them:
1. Change all uses of `connectFlowFixMe` to plain `connect`: `perl -i -0pe 's/connectFlowFixMe/connect/g' src/*/*.js`
2. Run `yarn flow` to see errors.
3. Pick a file with errors; fix the errors.
4. Commit just that file.
5. Repeat from step 2.

Then send a PR with any one or more commits made this way.

PR #3471, opened 2019-05-02
WIP: Flow coverage of 'connect' calls

borisyankov:
This is very WIP and is being opened only so we can refer to it via other PRs.

### DIFF
--- a/docs/architecture/react.md
+++ b/docs/architecture/react.md
@@ -78,7 +78,7 @@ are created with Redux's `connect`, and the code of ours that's involved is
 typically highly structured and pretty short, as in this example:
 
 ```
-export default connect((state: GlobalState) => ({
+export default connect((state) => ({
   unreadHuddlesTotal: getUnreadHuddlesTotal(state),
   unreadPmsTotal: getUnreadPmsTotal(state),
 }))(IconUnreadConversations);
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -2,7 +2,7 @@
 import React, { PureComponent } from 'react';
 import { View, StyleSheet } from 'react-native';
 
-import type { User, Dispatch } from '../types';
+import type { User, InjectedDispatch } from '../types';
 import { connect } from '../react-redux';
 import { UserAvatar, ComponentList, RawLabel } from '../common';
 import { getCurrentRealm, getUserStatusTextForUser } from '../selectors';
@@ -35,7 +35,7 @@ type SelectorProps = {|
 type Props = {|
   user: User,
 
-  dispatch: Dispatch,
+  ...InjectedDispatch,
   ...SelectorProps,
 |};
 
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -27,15 +27,18 @@ const componentStyles = StyleSheet.create({
 
 const AVATAR_SIZE = 200;
 
+type OwnProps = {|
+  user: User,
+|};
+
 type SelectorProps = {|
   realm: string,
   userStatusText: string | void,
 |};
 
 type Props = {|
-  user: User,
-
   ...InjectedDispatch,
+  ...OwnProps,
   ...SelectorProps,
 |};
 
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -82,7 +82,7 @@ class AccountDetails extends PureComponent<Props, void> {
   }
 }
 
-export default connect((state, props): SelectorProps => ({
+export default connect((state, props: OwnProps): SelectorProps => ({
   realm: getCurrentRealm(state),
   userStatusText: getUserStatusTextForUser(state, props.user.user_id),
 }))(AccountDetails);
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -18,10 +18,18 @@ const styles = StyleSheet.create({
   },
 });
 
-type Props = {|
-  user: UserOrBot,
-  ...InjectedDispatch,
+type OwnProps = {|
   navigation: NavigationScreenProp<{ params: {| email: string |} }>,
+|};
+
+type SelectorProps = {|
+  user: UserOrBot,
+|};
+
+type Props = {|
+  ...InjectedDispatch,
+  ...OwnProps,
+  ...SelectorProps,
 |};
 
 class AccountDetailsScreen extends PureComponent<Props> {
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -62,6 +62,6 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state, props) => ({
+export default connectFlowFixMe((state, props: OwnProps): SelectorProps => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
 }))(AccountDetailsScreen);
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -4,7 +4,7 @@ import { StyleSheet } from 'react-native';
 import type { NavigationScreenProp } from 'react-navigation';
 
 import type { InjectedDispatch, UserOrBot } from '../types';
-import { connectFlowFixMe } from '../react-redux';
+import { connect } from '../react-redux';
 import { getAccountDetailsUserForEmail } from '../selectors';
 import { Screen, ZulipButton } from '../common';
 import { IconPrivateChat } from '../common/Icons';
@@ -62,6 +62,6 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state, props: OwnProps): SelectorProps => ({
+export default connect((state, props: OwnProps): SelectorProps => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
 }))(AccountDetailsScreen);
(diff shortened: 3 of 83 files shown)


gnprice:
Thanks @borisyankov ! On just a quick skim:

> Split 'Props' into 'OwnProps' and 'StateProps'

As I said on a previous version of this change, I don't like the name `StateProps` because it sounds like a reference to the React component's own state and `State` type. When I added a few of these in #3450 / 165e848ea and thereafter, I called them `SelectorProps`, which I think points better at what they actually do.

I'm also not a fan of the name `OwnProps` -- it just raises the question, "*whose* own props?" If anything, what I'd expect "own props" to mean in one of these files is the props belonging to the component the code is mainly about... i.e. an emphatic way of referring to `Props` in *contrast* to some other set of props, like the props passed to the `connect`-made wrapper. One name I think can work for this concept is `OuterProps`.

Separately: this commit re-introduces `GlobalState` boilerplate in some places.

gnprice:
Also on that same commit: there are a ton of changes that are near-mechanical, and then a few that are more complex -- like this:
```
-// TODO get a type for `connectActionSheet` so this gets fully type-checked.
```
(is that TODO really eliminated?) and this:
```
-            hidden={hidden && Platform.OS !== 'android'}
+            hidden={!!hidden && Platform.OS !== 'android'}
```
As always in that situation, it'd be a big help to pull out the interesting changes as one or more relatively small commits. Otherwise it's quite hard to effectively read or review.

borisyankov:
`OwnProps` and `StateProps` are the names and types used by Redux / React-Redux.
These names are reflected in examples, types, and documentation.

gnprice:
Yeah, the names make some sense inside the context of react-redux. In that context, things are talking about the wrapper component that `connect` will make, and about Redux which has a state.

But this isn't inside react-redux -- in fact it's up at the opposite end of the file from where we invoke `connect`. These types and their names are read as part of the definition of the component. And these "own props" are the opposite of the component's own props, and the state/selector props are completely unrelated to anything in the component's state.

borisyankov:
> Also on that same commit: there are a ton of changes that are near-mechanical, and then a few that are more complex

That is part of the 'WIP'-iness. Shouldn't have done these in the same commit at all. The commit will remain purely the 'split' types changes and this is moving to another commit.

borisyankov:
>  I called them SelectorProps, which I think points better at what they actually do.

I don't think `SelectorProps` is better, as these are 'props coming from the global state, extracted by selectors, injected via `connect`'. I would have called them `GlobalStateProps`. 

Currently renamed to `SelectorProps` though, if you insist. What would be better than `OwnProps`?

borisyankov:
What is your opinion on this:
```js
 user: getAllUsersByEmail(state).get(props.email),
```

When typed correctly, `user` can be `undefined`. `|| NULL_USER` is the simplest solution. The alternative is not that clean.

borisyankov:
Not only did I not split the `hidden={!!hidden && Platform.OS !== 'android'}` line, it is now unneeded, so I did undo the change.

gnprice:
> Currently renamed to `SelectorProps` though, if you insist.

Thanks!


> I would have called them `GlobalStateProps`.

I think this would be an OK name. I have another thought to add about `SelectorProps`, though:


> > I called them SelectorProps, which I think points better at what they actually do.
> 
> I don't think `SelectorProps` is better, as these are 'props coming from the global state, extracted by selectors, injected via `connect`'.

One fact that perhaps might help: the `mapStateToProps` argument we pass to `connect` is itself a selector. (After all, what's a selector? It's just some pure function of the Redux state, and optionally of some second parameter. Called "selector" because a typical one just selects some particular piece of the sprawling state tree. All of which is exactly what `mapStateToProps` is about.)

I think I thought of `SelectorProps` as a possible name before I noticed that fact, but once I did notice it it helped me become quite comfortable with the name.

What do you think in light of that?

gnprice:
> What would be better than `OwnProps`?

One name I think is better (mentioned above) is `OuterProps`.

Not an awesome name, but I think it at least resolves the ambiguity that I think is fatal for `OwnProps` in this context: there are two components, one outer and one inner, and it makes clear these are the props of the *outer* one.

gnprice:
> What is your opinion on this:
> 
> ```js
>  user: getAllUsersByEmail(state).get(props.email),
> ```
> 
> When typed correctly, `user` can be `undefined`. `|| NULL_USER` is the simplest solution. The alternative is not that clean.

Ah, thanks for asking!

The question I always ask in a situation like this is: is this case actually possible when this code works as intended? Or is it impossible, because of an invariant that the type system just doesn't know about?

If it really is possible, then the code should handle it in a way that gives the user a reasonable result. I think `NULL_USER` rarely does that -- if the `NULL_USER` case actually happens, it's likely to produce a blank and buggy-looking widget on the screen. Instead, either `undefined` or `null` can be a good way for selector code like this to signal that case, and then some logic in a place like a `render` method can make an appropriate decision.

If it's impossible -- or, if there just isn't a reasonable thing the app can do in the context other than give up -- then the right thing to do is to throw an exception. Take a look [in `accountsSelectors.js`](https://github.com/zulip/zulip-mobile/blob/aada8313c/src/account/accountsSelectors.js#L118), at `getAuth` and `tryGetAuth`:
```
/**
 * The auth object for the active, logged-in account, or undefined if none.
 *
 * The "active" account is the one currently foregrounded in the UI, if any.
 * The active account is "logged-in" if we have an API key for it.
 *
 * This is for use in early startup, onboarding, account-switch,
 * authentication flows, or other times where there may be no active account
 * or it may not be logged in.
 *
 * See:
 *  * `getAuth` for use in the bulk of the app, operating on a logged-in
 *    active account.
 *  * `getPartialAuth` for use in authentication flows, where there is an
 *    active account but it may not be logged in.
 */
export const tryGetAuth: Selector<Auth | void> = /* ... */ ;

/**
 * The auth object for the active, logged-in account; throws if none.
 *
 * For use in all the normal-use screens and codepaths of the app, which
 * assume there is an active, logged-in account.
 *
 * See:
 *  * `tryGetAuth` for the meaning of "active, logged-in".
 *  * `tryGetAuth` again, for use where there might not be such an account.
 */
export const getAuth = (state: GlobalState): Auth => {
  const auth = tryGetAuth(state);
  if (auth === undefined) {
    throw new Error('Active account not logged in');
  }
  return auth;
};
```

A closer example, though it seems to be lacking jsdoc (oops! I should probably fix that), is `getStreamFromId` -- takes a `GlobalState` and a `number` stream ID, returns the stream, throws if none.

gnprice:
At a quick skim, the 5 small commits at the front of the branch all look good (thanks!). Would you like me to look closer and merge?

On the next one, `flow: Remove 'GlobalState' type from 'connect' calls` -- definitely in favor of this direction. But most of these are call sites of `connectFlowFixMe`, which has `$FlowFixMe` aka `any` as its parameter type, so the annotation is actually helpful there. So what I'd prefer to do for all of those is to remove the `GlobalState` annotation at the same time as fixing them each to proper `connect`.

Oh, I guess speaking of which: those first few commits fix type errors. Are those type errors that are suppressed by `connectFlowFixMe`, and can those call sites be converted to proper `connect`? If so, let's do that at the same time as fixing each error -- that helps add extra context to why the error was possible, and also puts a ratchet on the fix.

borisyankov:
Yeah, indeed the first 5 commits are independent and can be merged.

The issues were uncovered by the [flow: Fully type 'connect' calls](https://github.com/zulip/zulip-mobile/pull/3471/commits/dc40c0fc2fe9983a5aee513e4d3dc055aeee8926) commit

It seems that this had a great overlap with the issues `connectFlowFixMe` was ignoring. I did temporarily replace them all with a `connect` and there was just one new error.

I am fixing it now and will add to this PR, together with a commit removing the `connectFlowFixMe`.

gnprice:
Cool. To make the status of things explicit: for those first few commits, this comment I made above applies:

> Oh, I guess speaking of which: those first few commits fix type errors. Are those type errors that are suppressed by `connectFlowFixMe`, and can those call sites be converted to proper `connect`? If so, let's do that at the same time as fixing each error -- that helps add extra context to why the error was possible, and also puts a ratchet on the fix.

Generally I write reviews in the order I see things, and usually don't go back and polish the review as a unified whole. This helps me get them out efficiently, and I hope at least some of the time it's helpful for you and others to see how I approach things. It does mean that if you don't read to the end there's a possibility of confusion.

gnprice:
Adding cross-reference links, because seeing what we've said before is useful:
* #3451 is the issue this addresses
* #3428 is a previous PR that this is a followup to.

chrisbobbe:
This PR may be more exhaustive than one like https://github.com/zulip/zulip-mobile/pull/3789, but it’s probably OK if work on https://github.com/zulip/zulip-mobile/issues/3451 doesn’t all get done at once.

This PR, aside from having conflicts, proposes some patterns that may hinder readability, so it may be time to close it. Specifically, I agree with Greg’s positions against using OwnProps and InjectedDispatch, e.g., at https://github.com/zulip/zulip-mobile/pull/3471#issuecomment-490640211 and 
https://github.com/zulip/zulip-mobile/pull/3465#issuecomment-488070565.

gnprice:
Yep, I agree -- let's close this PR. Those first 5 commits (discussed in the last few comments above) would have been helpful if fixed up to be self-contained by removing the fixmes, but #3789 now covers most of the same ground and provides much more clarity about what it's doing and why.

It looks like in fact the only use of `connectFlowFixMe` (i.e. the only case of #3451) which #3789 doesn't yet cover and this does is the one in `ZulipStatusBar`. I think the first commit in this PR is the one that covers that. But in addition to the comment I made at the time that it should remove the fixme, rereading it now, I don't love the API choice it makes for `ZulipStatusBar` (with "no background" meaning "black background"); so in that case too we'll do better just making a fresh fix.

Thanks!

chrisbobbe:
> the one in ZulipStatusBar

I believe Ray has a fix for this one in progress, with a plan to remove the connectFlowFixMe definition altogether.

PR #3475, opened 2019-05-05
UnreadNotice: Remove disused 'limited' property

borisyankov:
Standalone, but also a prerequisite for #3471 because it removes the need for a Flow fix.


gnprice:
Thanks @borisyankov ! Merged.

PR #3476, opened 2019-05-07
Improve android notifications.

jainkuniya:
Beautify Android notification #2691 (in continuation with #3407)

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
@@ -174,11 +174,11 @@ private fun getNotificationBuilder(
     return builder
 }
 
-internal fun onOpened(application: ReactApplication, conversations: ConversationMap, data: Bundle) {
+fun onOpened(application: ReactApplication, conversations: ConversationMap, data: Bundle, identity: Identity) {
     logNotificationData(data)
     NotifyReact.notifyReact(application, data)
     getNotificationManager(application as Context).cancelAll()
-    clearConversations(conversations)
+    clearConversations(conversations, identity)
     try {
         ShortcutBadger.removeCount(application as Context)
     } catch (e: Exception) {
@@ -187,7 +187,7 @@ internal fun onOpened(application: ReactApplication, conversations: Conversation
 
 }
 
-internal fun onClear(context: Context, conversations: ConversationMap) {
-    clearConversations(conversations)
+fun onClear(context: Context, conversations: ConversationMap, identity: Identity) {
+    clearConversations(conversations, identity)
     getNotificationManager(context).cancelAll()
 }
--- a/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
@@ -22,7 +22,6 @@ import com.zulipmobile.BuildConfig
 import com.zulipmobile.R
 
 private val CHANNEL_ID = "default"
-private val NOTIFICATION_ID = 435
 
 @JvmField
 val ACTION_CLEAR = "ACTION_CLEAR"
@@ -68,6 +67,10 @@ internal fun onReceived(context: Context, conversations: ConversationMap, mapDat
         updateNotification(context, conversations, fcmMessage)
     } else if (fcmMessage is RemoveFcmMessage) {
         removeMessagesFromMap(conversations, fcmMessage)
+        val byConversationMap = conversations[fcmMessage.identity]
+        if (byConversationMap == null || byConversationMap.isEmpty()) {
+            getNotificationManager(context).cancel(getNotificationTag(fcmMessage.identity), getNotificationId(fcmMessage.identity))
+        }
         if (conversations.isEmpty()) {
             getNotificationManager(context).cancelAll()
         }
@@ -76,13 +79,20 @@ internal fun onReceived(context: Context, conversations: ConversationMap, mapDat
 
 private fun updateNotification(
     context: Context, conversations: ConversationMap, fcmMessage: MessageFcmMessage) {
-    if (conversations.isEmpty()) {
+    val byConversationMap = conversations[fcmMessage.identity]
+    if (byConversationMap == null || byConversationMap.isEmpty()) {
+        getNotificationManager(context).cancel(getNotificationTag(fcmMessage.identity), getNotificationId(fcmMessage.identity))
+        return
+    } else if (conversations.isEmpty()) {
         getNotificationManager(context).cancelAll()
         return
     }
-    val byConversationMap = conversations[fcmMessage.identity] ?: return
     val notification = getNotificationBuilder(context, byConversationMap, fcmMessage).build()
-    getNotificationManager(context).notify(NOTIFICATION_ID, notification)
+    getNotificationManager(context).notify(
+        getNotificationTag(fcmMessage.identity),
+        getNotificationId(fcmMessage.identity),
+        notification
+    )
 }
 
 private fun getNotificationSoundUri(context: Context): Uri {
@@ -177,7 +187,7 @@ private fun getNotificationBuilder(
 fun onOpened(application: ReactApplication, conversations: ConversationMap, data: Bundle, identity: Identity) {
     logNotificationData(data)
     NotifyReact.notifyReact(application, data)
-    getNotificationManager(application as Context).cancelAll()
+    getNotificationManager(application as Context).cancel(getNotificationTag(identity), getNotificationId(identity))
     clearConversations(conversations, identity)
     try {
         ShortcutBadger.removeCount(application as Context)
@@ -189,5 +199,5 @@ fun onOpened(application: ReactApplication, conversations: ConversationMap, data
 
 fun onClear(context: Context, conversations: ConversationMap, identity: Identity) {
     clearConversations(conversations, identity)
-    getNotificationManager(context).cancelAll()
+    getNotificationManager(context).cancel(getNotificationTag(identity), getNotificationId(identity))
 }
--- a/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FCMPushNotifications.kt
@@ -80,7 +80,8 @@ private fun updateNotification(
         getNotificationManager(context).cancelAll()
         return
     }
-    val notification = getNotificationBuilder(context, conversations, fcmMessage).build()
+    val byConversationMap = conversations[fcmMessage.identity] ?: return
+    val notification = getNotificationBuilder(context, byConversationMap, fcmMessage).build()
     getNotificationManager(context).notify(NOTIFICATION_ID, notification)
 }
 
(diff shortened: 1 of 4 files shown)


gnprice:
Thanks @jainkuniya !

```
e81de2e9c android, kt, doc: Fix linking of element/class in comments
```

Ah, good catch, thanks :smile: 


```
b1d50c109 android, notif: Make `removeMessagesFromMap` to accept `removeFcmMessage`.
```

Cool, looks like a good refactor.


```
8a728d3ee android, notif: Create fun to get notificationId.

+fun getNotificationId(identity: Identity?): Int {
+    return if (identity == null)  NOTIFICATION_ID else Integer.parseInt("${identity.realmId}${identity.userId}")
+}
```

A couple of thoughts:
* That scheme for generating IDs looks messy. It turns the two input numbers into strings (in base 10); then concatenates the strings together; then interprets that back as a number.
  * In general, turning things into strings and back is a bad sign -- it usually leads to bugs.
  * Here, very concretely: perhaps one identity has realmId 12 and userId 3, and another has realmId 1 and userId 23. They'd collide, at 123.
* Also, it's important to distinguish between realmId 1 on one server, and realmId 1 on another.
  * So we can either use `serverHost` plus `realmId`, or use `realmUri`. (Plus `userId`, either way.)
  * I don't see an important reason to prefer one or the other. Minor reasons include (a) that `realmUri` is a bit newer (so using the others works with server 1.8.0, and it requires 1.9.0); (b) that `realmUri` is just one field instead of two.
  * Oh but (a) doesn't really apply anyway, because we don't have `userId` before 2.0.0.
  * (Do you know where to find this version information? Important info to be able to find ........... I'm consulting the comments in `FcmMessage.kt`. This is why I made sure to take careful notes there. :smile: )
* Hmm, one thing I see in [the API docs](https://developer.android.com/reference/android/app/NotificationManager): instead of just an int ID, we can pass a string "tag" plus an int ID, and the pair (tag, id) is the unique identifier. (The tag defaults to `null`.)
  * Let's do that, with `realmUri` as the tag.
* Make sure to think through what happens if various fields are `null`. Again, consult the comments in `FcmMessage.kt` to understand what combinations of `null` fields are possible.

```
b77728bb0 android, notif: Create fun `dataForClearAction`.
```

I'll be better able to understand this in context, so I won't try to review in detail until you have the code that uses it. Small things in the meantime:
* No blank line between doc comment and the thing it documents.
* This doesn't refer to its receiver at all, does it? In that case it doesn't belong as a method here.
  * Feel free to make it just a standalone, top-level function if you don't see a more helpful alternative.

gnprice:
And merged those first two commits! They're 5366d8b2b and parent.

jainkuniya:
> In general, turning things into strings and back is a bad sign -- it usually leads to bugs.

Yup, got it.

 - Another way can be: use of `hashCode` fun
```
    override fun hashCode(): Int {
        var result = serverHost.hashCode()
        result = 31 * result + realmId
        result = 31 * result + (realmUri?.hashCode() ?: 0)
        result = 31 * result + (userId ?: 0)
        return result
    }
``` 
It will uniquely identify `Identity` plus we don't have to much think about `null`, because  for older server version it might fallback to same `id` (hashcode), but will definitely start working fine for latest server. Not much code, and no custom handling.

- Yup, tag (`realmUri`) and id (`userId`) also sounds good. 
  - When both are null (<1.9.0) use `NOTIFICATION_ID`. On notification click, fallback to currently selected account, no account switch (current behaviour). A single notification for whole Zulip app.
 - When `realmUri` is present, but `userId` not (1.9.0<server <2.0.0). Separate notifications based only on `realmUri` . 
 - When both are present all good 😍 Separate notifications based on each account and switch to it on clicking them.


If we go with later one, then we will end up in null checks and all. But with first one they will get handled automatically.  


> (Do you know where to find this version information? Important info to be able to find ........... 

We have server info details in RN part, but not on native side.
BTW there should be some way to access RN state from native, it will help a lot for feature like:
- reply from notification (`auth` is required to send messages from native)
- show streams/pm directly in `share with`, like Whatsapp, messenger
- widgets etc

jainkuniya:
```
b77728bb0 android, notif: Create fun `dataForClearAction`.
```
This data  will be used in `NotificationIntentService`. To identify which notification is actually clicked. So that `onClear` and `onOpened` can be called for that `byConversation` (not on whole conversations).

Also I have some doubts: why are we using service (`NotificationIntentService`). Services are basically which are long lived and runs in background. Receiver (`BroadcastReceiver`) are something which are there to receive particular events, and this is our use case.

https://developer.android.com/guide/components/services
https://developer.android.com/guide/components/broadcasts

gnprice:
> > (Do you know where to find this version information? Important info to be able to find ...........
> 
> We have server info details in RN part, but not on native side.

Hmm, I guess perhaps that was ambiguous.

I *don't* mean having the code look at what version the server is. In fact, that is almost always a bad idea and we shouldn't do it. (Instead, we should look at the actual relevant behavior and just handle the thing that actually happened -- like a given field is present or missing.)

What I meant was, I'd said some things like

> `realmUri` is a bit newer (so using the others works with server 1.8.0, and it requires 1.9.0);

which refers to a set of facts about what behavior appears in which server versions. Those facts are useful in making decisions about what server behaviors to take care to handle and what behaviors to not worry so much about, even though the code will just be looking at the actual behavior and not trying to interpret a server version number.

And so I was asking if you knew how to look up those facts, as a human. Is that question clearer? :slightly_smiling_face:

gnprice:
> * Yup, tag (`realmUri`) and id (`userId`) also sounds good.
>   
>   * When both are null (<1.9.0) use `NOTIFICATION_ID`. On notification click, fallback to currently selected account, no account switch (current behaviour). A single notification for whole Zulip app.
> * When `realmUri` is present, but `userId` not (1.9.0<server <2.0.0). Separate notifications based only on `realmUri` .
> * When both are present all good heart_eyes Separate notifications based on each account and switch to it on clicking them.
> 
> If we go with later one, then we will end up in null checks and all. But with first one they will get handled automatically.

Let's use tag `realmUri` and id `userId`.

I think with the right use of `?:` and perhaps `?.`, the handling of nulls won't involve any more complexity than it does in the `hashCode` implementation you showed above.

gnprice:
> ```
> b77728bb0 android, notif: Create fun `dataForClearAction`.
> ```
> 
> This data will be used in `NotificationIntentService`. To identify which notification is actually clicked. So that `onClear` and `onOpened` can be called for that `byConversation` (not on whole conversations).

Sure -- again, I'll be able to review it in detail once I see the code that'll be using it.


> Also I have some doubts: why are we using service (`NotificationIntentService`). Services are basically which are long lived and runs in background. Receiver (`BroadcastReceiver`) are something which are there to receive particular events, and this is our use case.
> 
> https://developer.android.com/guide/components/services
> https://developer.android.com/guide/components/broadcasts

Hmm, good question! I agree that from their descriptions, a receiver sure sounds like a better fit than a service.

Another place I'd look is the guide documentation for doing the specific thing we're doing, which is responding to an action on a notification:
https://developer.android.com/training/notify-user/build-notification#Actions
That suggests a `BroadcastReceiver`, though it's only an example and doesn't say this is the *only* reasonable thing to do.

One observation that I think complicates the picture: for receiving FCM messages (the things that come from the server and cause us to show a notification in the first place), we have a service `FcmListenerService`, because that's what the FCM API prescribes. Also for learning our FCM instance ID -- which is a very infrequent event, probably once ever for most installs of the app -- we have a service `InstanceIDListenerService`, again because that's what the FCM API prescribes. Both of those also sound more like use cases for Receivers than Services, just like `NotificationIntentService` does... but the people who designed the FCM API probably understand the Android system better than you or I do, and they chose to make them Services.

In any case, I think a service is fine -- maybe I'd pick a receiver if rewriting it, but AFAICT the service isn't causing problems.

gnprice:
Thanks @jainkuniya for the revisions!

Only managed a few minutes with this PR today, but quick comments:

## android, notif: Change structure of the ConversationMap.

* This generally makes sense!

* One thing to adjust:
```
 private fun getNotificationBuilder(
     context: Context, conversations: ConversationMap, fcmMessage: MessageFcmMessage): Notif
ication.Builder {
+    var byConversationMap = conversations[fcmMessage.identity]
```

Better to have the caller do this, and pass only `byConversationMap`. One reason is that it avoids accidentally using the full `conversations` in the rest of the function...

```
         inboxStyle.setSummaryText("${conversations.size} conversations")
```
... like here, for example :wink: 

```
+    if (byConversationMap == null) {
+        byConversationMap = ByConversationMap()
+    }
```

* Also, is this `null` case really possible? If it did happen, would this really be doing the right thing? I think the answers are "no" and "no". :slightly_smiling_face: (Consider what the resulting notification would actually look like.)

  We should never have a case that just generates nonsense. If it can happen, do something reasonable. If it can't and there isn't a reasonable thing to do, raise an exception.

## android, notif: Create fun `dataForClearAction`.

```
     }
+    /**
+     * All the data to dismiss notification for this particular identity
+     */
+
+    fun dataForClearAction(): Bundle {
```

* Take a look at how this appears visually. Which function is it attached to? Then follow the style of the similar comments around it.

## android, notif: Set uri as well for action clear intent.

    Next we need to send bundle data (Identity) with this intent so that
    only messages for this notification are removed from conversation  map.
    Note: without setting uri, sending of bundle was not working.

* Hmm interesting -- can you say more about what problem you were seeing?

```
+        val uri = Uri.fromParts("zulip", "clear:${fcmMessage.zulipMessageId}", "")
```

* Compare this URI with the other, existing, one, where the name comes from `"msgid:${fcmMessage.zulipMessageId}"`. What does `msgid` mean here? What makes it more or less appropriate in one of these contexts?


## android, notif: Create seperate notifications for each Identity.

```
     } else if (fcmMessage is RemoveFcmMessage) {
         removeMessagesFromMap(conversations, fcmMessage)
+        val byConversationMap = conversations[fcmMessage.identity]
+        if (byConversationMap == null || byConversationMap.isEmpty()) {
+            getNotificationManager(context).cancel(getNotificationTag(fcmMessage.identity), getNotificationId(fcmMessage.identity))
+        }
         if (conversations.isEmpty()) {
             getNotificationManager(context).cancelAll()
         }
```

* This feels kind of verbose and redundant. Redundancy that doesn't have a clear purpose tends to harbor bugs.

  Do we need the `cancelAll`? Why?

  Is the `byConversationMap.isEmpty()` case actually possible? If so, can we make it impossible?

jainkuniya:
```
android, notif: Change structure of the ConversationMap.
```
> Also, is this null case really possible? 

very unlikely to happens. But IDE (Kotlin) forces to handle it, so let's just return when `byConversationMap` is null.

gnprice:
> > Also, is this null case really possible?
> 
> very unlikely to happens. But IDE (Kotlin) forces to handle it, so let's just return when `byConversationMap` is null.

Cool, thanks for that revision.

When you next have a chance to work on this, would you address my other comments?

I think the last one (about "android, notif: Create seperate notifications for each Identity.", with several questions about it) in particular gets at some important conceptual aspects in making this logic all fit together and work reliably.

gnprice:
Thanks again @jainkuniya for all your work on this!

As we discussed today, let's set aside this line of work; we have lots of other things to pursue that are a better fit for you and that I'm eager to have you working on instead :slightly_smiling_face:

PR #3479, opened 2019-05-08
subscriptionsReducers: Delete subscriptions when a stream is deleted

### LINKED ISSUE #3474: Subscription list does not update when a stream is deleted.
Hello,

Last week when I was working on #3448, I found that subscription list does not get updated when a stream is deleted (although the stream list is updated).

Steps to reproduce:
1. Open the mobile app and the web app.
2. Delete a stream (which you are subscribed to) from the web app. 
3. The deleted stream should disappear from subscription list, but it does not.

I tried to trace the root cause of this problem and it seems that `getSubscriptions` in `getSubscribedStreams` in  `src/subscriptions/subscriptionSelectors.js` returns the same list before & after deleting a stream. This could be easily fixed like I did in this [commit](https://github.com/ungps/zulip-mobile/commit/e13cba1644e6db9a4c78eb99eb49169fc770871f), but I think this shouldn't be fixed here. It seems it is more of a server problem and should probably be posted on https://github.com/zulip/zulip/issues, but I want a confirmation before posting there. 

P.S: The subscription list is updated properly .when a user unsubscribes.

borisyankov:
Fixes #3474

We are not recieving a subscription 'unsubscribe' event from the
server when a stream we are subscribed to is deleted and thus we
keep subscribed to non-exiting stream until the state is refreshed.

The ideal fix for this issue should be implemented on the back-end.

This code is compatible with eventual server-side fix that would
issue an unsubscription on stream deletion.

We require this code regardless of such a fix, as we would need
to support old and new back-end versions.

### DIFF
--- a/src/subscriptions/__tests__/subscriptionsReducers-test.js
+++ b/src/subscriptions/__tests__/subscriptionsReducers-test.js
@@ -1,6 +1,7 @@
 import deepFreeze from 'deep-freeze';
 
-import { REALM_INIT, EVENT_SUBSCRIPTION, ACCOUNT_SWITCH } from '../../actionConstants';
+import { EventTypes } from '../../api/eventTypes';
+import { REALM_INIT, EVENT_SUBSCRIPTION, ACCOUNT_SWITCH, EVENT } from '../../actionConstants';
 import subscriptionsReducers from '../subscriptionsReducers';
 
 describe('subscriptionsReducers', () => {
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @borisyankov ! Two comments:

* Please look again at the switch/case syntax in the diff:
```
             case 'create':
             case 'delete':
+              return filterArray(
+                state,
+                // $FlowFixMe Flow is unable to infer the correct version of StreamEvent
+                x => !action.event.streams.find(y => x && y && x.stream_id === y.stream_id),
+              );
             case 'occupy':
             case 'vacate':
               return state;
```

I don't think what this does is what you wanted. :wink: 

* Why give up so quickly on types? Take a look at what the neighboring code does:
```
    case EVENT: {
      const { event } = action;
      switch (event.type) {
        case EventTypes.stream:
          switch (event.op) {
            case 'update':
              return updateSubscription(state, event);

            case 'create':
            case 'delete':
```

And in fact if we do the same and say just `event`, then Flow successfully relies on the `case` results just like usual.

borisyankov:
> Why give up so quickly on types? Take a look at what the neighboring code does:

Thanks for pointing this out. Never give up on types :)
I did the necessary change and now the FlowFixMe is gone.

borisyankov:
> Please look again at the switch/case syntax in the diff:

Were you referring to the same Flow type issue, or was it something else, that I am not getting?

gnprice:
Thanks for the update!

> > Please look again at the switch/case syntax in the diff:
> 
> Were you referring to the same Flow type issue, or was it something else, that I am not getting?

This code runs in the `create` case as well as `delete`. The commit message and the tests are all about `delete`.

(Also, there should be a blank line before the next `case`, to help the reader see the structure of which cases are grouped together and run the same code.)

borisyankov:
Hmm, I did the change right after your comment but forgot to bing that it is done.

borisyankov:
The needed change is actually this new commit: d622abbf3c4

gnprice:
Cool, thanks.

In the new commit:
```
    Also we can get rid of the
    eslint ignore rule.


             default:
-              (event: empty); // eslint-disable-line no-unused-expressions
               return state;
```

So, there's a reason this is here. (After all, it'd be very easy to remove it -- just delete that line!)

It's there because it recruits the type-checker to verify that our `case` statements here are exhaustive. In particular, this means that if in the future we add another possible value for `op`, the type-checker will identify for us every `switch` statement like this one that needs to be updated.

Try it out -- on master, delete the `occupy` and `vacate` cases. You'll immediately see a Flow error on that `event: empty`.  Then try that after this change, and no such luck.

If you do a quick search in the history like `git log --stat -p -S no-unused-expr`, you'll see this explained:
```
commit 93c7a4fc92323d374509bfeee669ad409b238aa7
Date:   Tue Mar 19 17:35:15 2019 -0700

    PresenceStatusIndicator: Ensure the `switch` is exhaustive.
    
    This incantation will cause a type error on any future change that
    would make this `switch` statement non-exhaustive (or more precisely,
    would make it reach this dummy default case.)
...
 
       default:
+        (status: empty); // eslint-disable-line no-unused-expressions
         return null;
```

but it is pretty obscure when looking at the code in the tree. I'll see about doing something to make it more explicit there.

gnprice:
Merged -- thanks again!

Edits I made:
* Dropped the refactor discussed in the previous comment.
* Rewrote the commit message to reflect my reply on #3474 .
* Added a small commit on top with style tweaks.

gnprice:
> but it is pretty obscure when looking at the code in the tree. I'll see about doing something to make it more explicit there.

And done, as 6dcbdf951 . Also added a few more uses of that trick, now that it's a lot simpler to use and to read: 438408a37 .

Issue #3482, opened 2019-05-13
Improve API error handling


borisyankov:
Implement some of the suggestions in the review of #3254
Some ideas listed here are enablers of these suggestions, others are just related and good to do.

* Add custom exception on API error (change `makeApiError') We are not doing this because React Native's build tools and/or Flow were having issues with custom error object. We should check if this is still true and if so whether there are workarounds. An alternative solution, that I do not prefer, would be is to check the form of the error object (is there an 'httpStatus' property).

* Type the responses and error object better and more strictly. The response from the server is a bit too varied (in some error cases it is even an html response and not json) and we have not documented or typed that extensively enough.

chrisbobbe:
I believe this was resolved with https://github.com/zulip/zulip-mobile/pull/3639, which was marked as superseding https://github.com/zulip/zulip-mobile/pull/3484, but please let me know if I'm closing this in error.

PR #3484, opened 2019-05-13
Make API errors better

borisyankov:
These are changes that were intended as part of #3254 but while necessary I decided it would be a feature creep to bundle these with the original PR.

These are the first steps to implementing #3482

### DIFF
--- a/src/api/__tests__/apErrors-test.js
+++ b/src/api/__tests__/apErrors-test.js
@@ -1,0 +1,21 @@
+import { isErrorBadEventQueueId, EventQueueError } from '../apiErrors';
+
+describe('isErrorBadEventQueueId', () => {
+  test('not any Error object is a "bad event queue id" error', () => {
+    const error = new Error();
+    const result = isErrorBadEventQueueId(error);
+    expect(result).toBe(false);
+  });
+
+  test('custom error of type "EventQueueError" and "BAD_EVENT_QUEUE_ID" returns true', () => {
+    const data = {
+      code: 'BAD_EVENT_QUEUE_ID',
+      msg: 'Bad event queue id: 123',
+      queue_id: '123',
+      result: 'error',
+    };
+    const error = new EventQueueError(400, data);
+    const result = isErrorBadEventQueueId(error);
+    expect(result).toBe(true);
+  });
+});
--- a/src/api/__tests__/apErrors-test.js
+++ b/src/api/__tests__/apErrors-test.js
@@ -1,4 +1,4 @@
-import { isErrorBadEventQueueId, EventQueueError } from '../apiErrors';
+import { isErrorBadEventQueueId, isClientError, ApiError, EventQueueError } from '../apiErrors';
 
 describe('isErrorBadEventQueueId', () => {
   test('not any Error object is a "bad event queue id" error', () => {
@@ -16,6 +16,14 @@ describe('isErrorBadEventQueueId', () => {
     };
     const error = new EventQueueError(400, data);
     const result = isErrorBadEventQueueId(error);
+    expect(result).toBe(true);
+  });
+});
+
+describe('isClientError', () => {
+  test('an API error with error code between 400 and 499 is a "client error"', () => {
+    const error = new ApiError(404, {});
+    const result = isClientError(error);
     expect(result).toBe(true);
   });
 });
--- a/src/api/apiErrors.js
+++ b/src/api/apiErrors.js
@@ -33,3 +33,17 @@ export class EventQueueError extends BaseApiError<ApiEventQueueResponseError> {}
  */
 export const isErrorBadEventQueueId = (e: Error | ApiError | EventQueueError): boolean =>
   (e instanceof ApiError || e instanceof EventQueueError) && e.data.code === 'BAD_EVENT_QUEUE_ID';
+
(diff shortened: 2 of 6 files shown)


borisyankov:
This is ready for review.

gnprice:
Thanks @borisyankov ! Comments below.

```
-  code?: ApiErrorCode,
+  code: ApiErrorCode,
```

* Huh interesting. What's the reason for this change?

  A `git log --stat -p -S ApiErrorCode` finds that when you originally added this type in 060a7762e, we wrote

      The response often contains a program-readable `code` constant,
      but not always.

  Is there a reason to no longer believe that's true?

```
+/**
+ * Response from the event queue in case of an error.
+ *
+ * @prop queue_id - String uniquely identifying the queue. Available in addition 
to
+ *   the usual error message information.
+ */
+export type ApiEventQueueResponseError = {|
+  ...ApiResponseError,
+  queue_id: string,
+|};
```

* Cool, good to make distinct. This should specify the appropriate `code` -- it's a lot like events, or for that matter Redux actions, in that one property's value identifies what other properties and types to expect.

```
    api: Add custom errors classes 'ApiError' and 'EventQueueError'
    
    We are currently hackily injecting some API data to a standard
    Error object. This works fine but creates issues with Flow, as
    it rightly complains of misuse.
    
    The right approach is a custom class, descendent of the Error
    object and enhanced with the needed properties. This used to
    be tricky because of the Babel / React Native build pipeline
    but improvements in the past year have made this possible.
```

* Ah nice! Glad to see this cleaned up. :-)

* What were the build pipeline changes that made this possible? Or: what would go wrong before?

* The jsdoc on `BaseApiError` appears to be incomplete. :wink: 

* Hmm, I feel like `EventQueueError` should be a subtype of `ApiError`.

  I guess an ingredient in that is that `ApiEventQueueResponseError` should be a subtype of `ApiResponseError`.

  And in turn what that's telling me is that `ApiResponseError` shouldn't be an exact object type. Which makes sense -- the reality is that API error responses may have additional fields. It's only when you know a specific error code (like `code: 'BAD_EVENT_QUEUE_ID'`) that you can know what other fields will be present (in that example, `queue_id: string` and no others) and have an exact type.

```
    Replace 'makeApiError' with a custom error class 'ApiError'
```

* Cool. I think this is best squashed into the parent commit, which defines `ApiError` -- it actually makes the parent easier to understand, by showing that choices like having those two fields with their particular names have just been moved from this other, previous code.

```
+export const isErrorBadEventQueueId = (e: Error | ApiError | EventQueueError): boolean =>
+  (e instanceof ApiError || e instanceof EventQueueError) && e.data.code === 'BAD_EVENT_QUEUE_ID';
```

* Yeah, it feels odd having these `ApiError | EventQueueError` etc. I think this is in part a symptom of having the one of them not be a subtype of the other, when it naturally ought to be.

* Also it's confusing that... if I'm reading this right, we never actually make an `EventQueueError`? Those only exist in tests. Instead, we make `ApiError` that just happens to have `.data.code === 'BAD_EVENT_QUEUE_ID'`.

  Ideally when we see one of those, we'd make it a type that matches, and tells Flow that there'll be a `queue_id: number`. But we don't actually use `queue_id`, so it'd be fine not to have this mechanism...

  but we definitely shouldn't have this extra type, and then not use it when we're handling the thing it's obviously intended for. That's super confusing. If we're not going to wrangle Flow so as to actually use it, we should delete it and just have `ApiError`.

borisyankov:
> ```
> -  code?: ApiErrorCode,
> +  code: ApiErrorCode,
> ```
>
> ```
> The response often contains a program-readable `code` constant,
> but not always.
> ```
>
> Is there a reason to no longer believe that's true?

Yeah, and that reason was I thought I was wrong originally, but actually not.
I did confirm that is still true, changed the commit to not do the change and added a line to clarify that indeed not all error responses include the 'code' value.

borisyankov:
> What were the build pipeline changes that made this possible? Or: what would go wrong before?

From what I remember Babel 6 did not allow extending built-in types. (and maybe there was some React Native-related issue, maybe not)

A plugin like this could have been used though:
https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend

But with the current dependencies, everything just works as expected.

borisyankov:
> I guess an ingredient in that is that ApiEventQueueResponseError should be a subtype of ApiResponseError.

It should but I have problems expressing the only difference between them is the data thrown, as the implementation is the same. `EventQueueError extends ApiError<ApiEventQueueResponseError>` didn't work, as ApiError is not generic. 

I expected something like this to be the solution:
```js
export class ApiError<T> extends BaseApiError<T | ApiResponseError> {}
```
But that is not valid Flow. Would be great if you know of a solution to this.

borisyankov:
> It's only when you know a specific error code (like code: 'BAD_EVENT_QUEUE_ID') that you can know what other fields will be present (in that example, queue_id: string and no others) and have an exact type.

That is not entirely correct. We can predict the response will be of that type if it is thrown by the `pollForEvents` API call that is targeting the `/events` endpoint. But that is tricky to express, and will definitely require some refactoring that might make the code a bit too messy for little benefit.

ray-kraesig:
Superseded by #3639.

Issue #3375, opened 2019-02-28
Show diverse post times for multiple posts from the same user


athena-metis:
In zulip web interface the post times for consecutive posts from the same user are individually time stamped however in zulip-mobile only the time of the first post is listed even if the posts are many hours apart. This behaviour is the same in both portrait and landscape layouts.

**Expected behaviour:** Show time for either:-
- each post (may clutter the interface for faster moving conversation)
- show the time for each post where the previous post was X time after the previous post.

Example in the web version
![web-version](https://user-images.githubusercontent.com/32957443/53596107-edaa3980-3b96-11e9-93c9-93deacb4e861.png)

Example in zulip-mobile (v22.1.108 on android, this is also unchanged in v23.1.110)
![mobile-version](https://user-images.githubusercontent.com/32957443/53596635-0ff08700-3b98-11e9-9bad-7d23104cc206.png)

athena-metis:
Just to confirm this behaviour is also the same in v23.1.110

borisyankov:
That is intentional. There is no space to loose on a mobile phone's screen.
We might want to add some way to display this data, but it will definitely not be the same way as web.

athena-metis:
I appreciate that space is limited and you wouldn't want to replicate the web approach, but not showing the time for more than the first post when they are hours apart is also not communicating valuable information.

armaanahluwalia:
Agree with this feedback. Could probably just add a thin divider with a timestamp centered? Since vertical space isn't as much of an issue.

timabbott:
I tend to agree with the reporter that we should include the timestamps.  It's often essential information for understanding what someone is communicating to see that they followed up 3 times an hour apart vs. 2 minutes apart.  And as is shown in that screenshot, our integrations are generally designed to not repeat the timestamp of an event because that information is expected to be displayed by the Zulip app itsefl.

gnprice:
Thanks @athena-metis for writing this up! I agree -- some way of showing this information is a recurring thing I notice that I wish I had, and we should add it.

FWIW WhatsApp shows the time of every message. They stick it in the lower-right corner of the message's box; they manage to do it so that it takes no horizontal space, except on the last line of the message (which after all doesn't typically reach the very end of the line anyway.)

Some other apps (Android Messages; FB Messenger) take an approach where if you touch a message, a small area expands to show its timestamp.

I tend to prefer WhatsApp's approach. It may be a bit tricky to implement, but if we can I think it'd be pretty great.

[...]

PR #3488, opened 2019-05-19
brief-messages: Add timestamp.

ishammahajan:
There have been multiple complaints about differing timestamps not
being available for messages sent at differing times by the same
user in succession. This commit aims to solve that problem by adding
a timestamp which comes right after a brief message (on the same
line, so as to conserve space). This timestamp uses the
`float:right` property, and clears itself from timestamps appearing
in further messages.

One thing of concern here is that the `top-margin` of the `::after` pseudo-element, based on calculation should be 0.2em, because the `font-size` is set to 0.8em. Inspite of this, 0.3em top-margin appears to be slightly better in terms of keeping the text aligned to the bottom of the line and not the top.

0.2

0.3

0.4em seems to make it fit perfectly!

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -95,7 +95,7 @@ hr {
 .timerow-right {
   background: -webkit-linear-gradient(left, #999 0%, transparent 90%);
 }
-.timestamp {
+.timestamp, .message-brief > .content > :last-child::after {
   color: #999;
   font-size: 0.9rem;
   white-space: nowrap;
@@ -108,6 +108,14 @@ hr {
 }
 .message-brief {
   padding: 0 1rem 1rem 4rem;
+}
+.message-brief > .content > :first-child {
+  clear: right;
+}
+.message-brief > .content > :last-child::after {
+  content: var(--time);
+  float: right;
+  margin: 0.1rem 0 0 2rem;
 }
 .message p + p {
   margin-top: 1rem;
--- a/src/webview/html/messageAsHtml.js
+++ b/src/webview/html/messageAsHtml.js
@@ -75,19 +75,22 @@ $!${message.content}
  ><p>Interactive message</p
  ><p>To use, open on web or desktop</p
 ></div>
+<div class="dummy-element-to-align-timestamp"></div>
 `;
 
 export const flagsStateToStringList = (flags: FlagsState, id: number): string[] =>
   Object.keys(flags).filter(key => flags[key][id]);
 
 export default (backgroundData: BackgroundData, message: Message | Outbox, isBrief: boolean) => {
-  const { id } = message;
+  const { id, timestamp } = message;
   const flagStrings = flagsStateToStringList(backgroundData.flags, id);
+  const messageTime = shortTime(new Date(timestamp * 1000), backgroundData.twentyFourHourTime);
   const divOpenHtml = template`
     <div
      class="message ${isBrief ? 'message-brief' : 'message-full'}"
      id="msg-${id}"
      data-msg-id="${id}"
+     $!${isBrief ? template`style="--time:'${messageTime}'"` : ''}
      $!${flagStrings.map(flag => template`data-${flag}="true" `).join('')}
     >`;
 
@@ -106,7 +109,7 @@ $!${divOpenHtml}
 `;
   }
 
-  const { sender_full_name, sender_email, timestamp } = message;
+  const { sender_full_name, sender_email } = message;
   const avatarUrl = getAvatarFromMessage(message, backgroundData.auth.realm);
   const subheaderHtml = template`
 <div class="subheader">
@@ -114,7 +117,7 @@ $!${divOpenHtml}
     ${sender_full_name}
   </div>
   <div class="timestamp">
-    ${shortTime(new Date(timestamp * 1000), backgroundData.twentyFourHourTime)}
+    ${messageTime}
   </div>
 </div>
 `;


ishammahajan:
A sample -- keep in mind this screenshot is with 0.4em top-margin.

src/webview/css/css.js
```diff
@@ -108,6 +108,16 @@ hr {
 .message-brief {
   padding: 0 1em 1em 4em;
 }
+.message-brief > .content > :first-child {
+  clear: right;
+}
+.message-brief > .content > :last-child::after {
```
borisyankov: Are you sure you need the `:after' pseudo-element? Can you achieve this in a more conventional way?
You should definitely use the already existing `.timestamp' styling, just adding the differentce in styles for `.message-brief `. That will ensure the styles change together.

src/webview/css/css.js
```diff
@@ -108,6 +108,16 @@ hr {
...
+  content: var(--time);
+  float: right;
+  color: #999;
+  font-size: 0.8em;
```
borisyankov: Not a good idea visually to have the sizes of the timestamps being different. Either use 0.9 as is the timestamp in 'message-full' or change both to 0.9

src/webview/css/css.js
```diff
@@ -108,6 +108,16 @@ hr {
 .message-brief {
   padding: 0 1em 1em 4em;
 }
+.message-brief > .content > :first-child {
+  clear: right;
+}
+.message-brief > .content > :last-child::after {
+  content: var(--time);
```
borisyankov: Have you confirmed that CSS variables are supported on all platforms that we support?

src/webview/html/messageAsHtml.js
```diff
@@ -80,6 +80,14 @@ export default (backgroundData: BackgroundData, message: Message | Outbox, isBri
      class="message ${isBrief ? 'message-brief' : 'message-full'}"
      id="msg-${id}"
      data-msg-id="${id}"
+     $!${
```
borisyankov: Extract this and the similar code in `message-full` to a common function.

ishammahajan: Sure! Thanks.

src/webview/css/css.js
ishammahajan: I had not, thanks for reminding! I have since confirmed that it works from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).

src/webview/css/css.js
ishammahajan: That makes sense, thanks, I have changed it!

src/webview/css/css.js
ishammahajan: Implementing this conventionally would be more complicated and would require an `::after` pseudo element in any case in order to give a sort of `left-margin` to the message because of the reasons mentioned in #3375 
>One part of how to pull that off is that it's important the actual message text not overlap the timestamp, or even crowd it too closely

>will ensure the styles change together.

Makes sense, thanks!

ishammahajan:
This push (since font-size has been increased to 0.9em on @borisyankov 's suggestion ) changes the scale of the margin-top issue mentioned in the first message on this PR by 1/2, but that problem still exists.

gnprice:
Thanks @ishammahajan !

Quick thought on this bit:

> One thing of concern here is that the `top-margin` of the `::after` pseudo-element, based on calculation should be 0.2em, because the `font-size` is set to 0.8em. Inspite of this, 0.3em top-margin appears to be slightly better in terms of keeping the text aligned to the bottom of the line and not the top.

This sounds like the sort of thing that gets confusing as a result of nested use of `em`. I suspect your 0.2em or 0.3em may be relative to the *new* size that's already multiplied by 0.8em -- in which case, (1.0 - 0.8) / 0.8 = 0.25em may be the value that gets it exactly right.

But also this is maybe a sign that it'd be good to switch to `rem` :slightly_smiling_face: .

gnprice:
> But also this is maybe a sign that it'd be good to switch to `rem`  .

(Which is happening in #3188 -- so my suggestion is to not worry about that wrinkle for now, until that PR is merged and you rebase on top of it.)

ishammahajan:
Thanks @gnprice for the heads up about the incoming merge! I have rebased on top of master and this is ready for a thorough review if this is the design we're going with.

I discovered a bug during refactoring this to `rem`. When used with a widget, because of their dummy nature, the timestamp wasn't aligned properly.
Screenshot 2019-06-01 at 1 24 09 PM

This has been fixed by adding a dummy element after the widget html, which I'm assuming is okay since it is presumably only a temporary element and going to be replaced soon with code which displays them properly.
Screenshot 2019-06-01 at 1 16 32 PM

gnprice:
Thanks @ishammahajan !

Just spent some time playing with this, reading through conversations. Also read the code.

* I find I quite appreciate the indication of where one message ended and the next began! :smile: IOW, it's successfully meeting this desire I mentioned on #3375:

  > in mobile Zulip there is (typically) no visual indication at all of the distinction between two consecutive messages from the same user. That's sometimes confusing too. Most mobile messaging apps have some kind of bubble around each message; and in the Zulip webapp, the timestamp serves this function.

* It's a little awkward that for sender-showing messages (`message-full`), the timestamp is at the top, but for sender-implicit messages (`message-brief`), it's at the bottom. Potentially confusing, as the pattern isn't consistent.

  This also means that between the first and second message for a given sender + recipient + date, that boundary doesn't gain any visual indication.

  Perhaps just put the timestamp at the bottom for all messages?

* Huh, this CSS syntax is new to me!

```
+.message-brief > .content > :last-child::after {
+  content: var(--time);
```

(The `var(...)`.) Seems powerful and pretty neat.

I wonder, though, if it'd be simpler to just write this element directly into the HTML.

* > This has been fixed by adding a dummy element after the widget html,

  Hmm. I don't mind adding a wrinkle like this to the widget's HTML... but I would want to understand why it was needed. In particular, it isn't obvious to me what about the widget's HTML (which looks pretty ordinary/boring to me) causes the issue in your screenshot -- which means I wonder if an ordinary message's body could also trigger the issue.

  Do you understand more of what's happening there? [Ah, I see -- more below.]

* The interaction with edited/starred tags doesn't look right:

  image

* Also odd with a reaction:

  image

* And when the message itself ends with a bullet point, it's OK but it loses the squeeze-onto-last-line magic:

  image

* Meanwhile, I think I understand now what's going on with the widgets! And it does not totally reassure me that there won't also be trouble with some messages' content. :wink: (E.g. an embedded tweet? We have a few other fancy cases too, which cause fancy HTML to be emitted by the server as the message content.)

  It's that the `.content > :last-child::after` sticks the timestamp as a pseudo-element *inside* the last child of the `.content`. Which works great so long as that last child is an ordinary `p` element.

  Hmm. It would certainly be nice to pull this off without having to muck with the HTML of the actual message body, as sent by the server in `message.content`. This CSS does that, except in the cases it doesn't work.

  One step would be to stick a `div` immediately around the actual message body, that doesn't include the tags, reactions, etc. That would fix the brokenness involving those. That leaves the case where the actual body doesn't end with a normal `p`.

  Can you perhaps write the CSS so that (a) if the last child really is a `p`, then the timestamp goes at `::after` of that `p`; (b) otherwise, it goes at `::after` of the whole body? I think that would cleanly address any un-planned-for case like an embedded tweet. If we treat the widget placeholder text the same as a message body, then it'll cleanly take care of that too.

  That still doesn't get us the magic for a `ul`, but that's OK.

* nit: In the commit message, let's be sure to mention the issue it fixes (#3375) -- that's helpful for finding the issue from the commit and vice versa.

gnprice:
> `var(...)`

Hmm, I went and checked the compatibility / newness of this feature:
https://developer.mozilla.org/en-US/docs/Web/CSS/--*

It's in Safari on iOS since an unspecified version... and in Chrome for Android since v49.

Referring back to js.js to remind myself what versions we support:
```
 * * On Android, core functionality needs to work on Chrome 44 (conveniently
 *   available for testing in a stock Android 6.0 Marshmallow image.)
 *   Graceful degradation is acceptable below Chrome 58 (ditto 8.0 Oreo.)
 *   When aware of issues down to Chrome 30 (shipped with 4.4 KitKat), we
 *   fix them if trivial, or else record in comments here.
```

I'd say it's fine for these timestamps not to show up on those older versions. It'll be important to test on Chrome 44 to make sure that the degradation is graceful, though.

Similarly important to test on iOS 9. (If you pin down whether `var()` itself is supported there, you can also save the next person from the same testing by editing that MDN page.)

gnprice:
Closing, as I've just merged your #3491 with the alternate design. Thanks again for all your work on this!

ishammahajan:
Thanks for the extensive reviews @gnprice! 😄

gnprice:
Reopening, as I've just sent this out to our beta testers for feedback, as version 25.5.119. (On Android, anyway; working on it for iOS.) See [discussion in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps/near/753634) about #3491 as well as this PR.

gnprice:
Some design feedback on this particular PR, and other discussion:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps.3A.20119/near/753791

(We'll try to list the main points back here on this thread, but see there for details.)

jainkuniya:
Here is a issue in the, which needs to be addressed before it goes to the master.
**How to reproduce:**
Turn of internet and send a message, message timestamp will rotate with loader in right bottom corner. Refer below message
image
https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps.3A.20119/near/753816

gnprice:
Thanks @jainkuniya for the screenshot of that fascinating symptom!

I have a pretty strong guess about what's causing it -- I was puzzled the other day when you first reported that symptom (in chat), but I think I have it now. It's this bit:

> * [...] It's that the `.content > :last-child::after` sticks the timestamp as a pseudo-element _inside_ the last child of the `.content`. Which works great so long as that last child is an ordinary `p` element.

I bet what's happening is that the `.content > :last-child` winds up being the spinner, specifically an element that's actually rotating its contents, and so the timestamp winds up getting put inside that.

If that's right, then I think it'll get fixed by the same thing that fixes the interaction with "starred"/"edited" tags, etc.:

> * [...] One step would be to stick a `div` immediately around the actual message body, that doesn't include the tags, reactions, etc. That would fix the brokenness involving those.

ishammahajan:
Is this design still on the roadmap?; I remember it being removed from the beta for some reason (don't exactly remember why though, might have been due to the above bug).

chrisbobbe:
> Is this design still on the roadmap?; I remember it being removed from the beta for some reason (don't exactly remember why though, might have been due to the above bug).

Hmm, I'm not 100% sure what happened! I'll try to reconstruct some history here. 🙂 

On June 7th, when your [commit](https://github.com/zulip/zulip-mobile/commit/8a1099fcca1e682aa7aabf1f6656d5cf3d901ad0) in https://github.com/zulip/zulip-mobile/pull/3491 was merged, it was marked with "Fixes #3375." As a result, the issue #3375 was automatically closed. Since the current PR is meant to fix the same issue, that made its status and priority unclear, since it no longer had an open issue in the issue list. The same day, Greg resolved this by closing the current PR, because it looked like https://github.com/zulip/zulip-mobile/pull/3491 was meant to replace it.

Then, with the current PR still not corresponding to an open issue, the PR was re-opened to await the results of some beta testing. Quite a lot of discussion followed, but I believe (and I may be wrong) that it was generally resolved that this PR was no longer necessary. I may be wrong about this. That discussion happened here: https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps/near/756150

What do you think, @ishammahajan ? @gnprice ?

chrisbobbe:
Until this PR is completed or closed, though, I have to say it blocks https://github.com/zulip/zulip-mobile/issues/3593 because that's a UI issue in the same area (we'd like a tappable star icon on each message).

ishammahajan:
@chrisbobbe I think considering the blocked work, this PR might be a good design choice. I have been very busy with academics & project work since the summer ended last year, which was why I wasn't able to complete my pending PRs. I can try completing this PR over the weekend if it's required though.

What do you think?

chrisbobbe:
> I have been very busy with academics & project work since the summer ended last year

Totally understandable! 🙂 

> I think considering the blocked work, this PR might be a good design choice.

Ah, I should have been more clear — completion of this pull request is not strictly required for https://github.com/zulip/zulip-mobile/issues/3593 to continue. I just don't want to start changing this same area of the UI when it's not clear whether or not this PR is intended to be completed.

I think what would be most helpful now — and it's fine if you don't have the time! — is, rather than spending your time on code changes, if you could briefly describe what this PR does differently to how the UI works in that area currently, and how it's better. It looks like your own completed work on https://github.com/zulip/zulip-mobile/pull/3491 was meant to replace this PR, so it's not 100% clear to me that this PR is still necessary. Some discussion on that ended with the beta 25.6.120 release, as Greg noted in https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps/near/756150, which I took as a hint that this was resolved, but I may be wrong.

ishammahajan:
> Ah, I should have been more clear — completion of this pull request is not strictly required for #3593 to continue.

Ohh. In that case I don't think this was intended to be completed in any case. I was under the impression that as a result of certain internal discussions this PR became blocking.

This can be closed then, I think. :)

chrisbobbe:
> This can be closed then, I think. :)

Ok, done!

Again, thanks for all the work you did on this PR and that other one. 🙂

PR #3493, opened 2019-05-22
webview: Do not throw on <img /> with no src attribute

borisyankov:
It does not seem we have `img` tags with no `src` attributes,
these would be quite useless. Yet, such tag is completely valid
html and could exist.

Accessing a missing attribute on an html element almost always
returns `null` (except older browsers where the more useful empty
string is returned) so calling `img.src.startsWith` will throw.

This change adds a check to prevent an exception in such cases.


chrisbobbe:
I'm closing this PR; as the description says, this isn't something we've actually seen, but a hypothetical.

If we introduce a mistake in how we render an image, we should fix that mistake.

If the server makes a mistake in how it tells us to render an image, that mistake should be fixed. Also, in principle, if we found we really wanted to validate something like this, the best place to do so would probably be at the edge, [crunchy-shell](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md) style.

PR #3495, opened 2019-05-22
webview: Add auth to local documents

### LINKED ISSUE #3303: Viewing uploaded files fails, or requires logging in again

### LINKED ISSUE #3487: Uploaded PDF can't be viewed
A user reports:

> I did send out several of PDFs to private messages of my members.
> Everybody that uses the App could not open the PDFs. The message was "Cannot open PDF".
> The members with web application had no problem opening the PDFs

The recipients they've heard from are all on Android; we don't know if the issue is also present on iOS.

Apparently it did work for another recipient, though, also on Android!

> now I got feedback from somebody with Android and the App and it worked! How about that?

So that could be a very interesting clue.

I briefly experimented trying to reproduce it: I uploaded a PDF from the webapp, sent the message, saw it on the mobile app, and tried following the link. For me with v24.0.113 on Android 9 aka P, the issue
* **does** reproduce on an org on zulipchat.com;
* **does not** reproduce on chat.zulip.org -- things work fine there.

When it reproduces, the behavior is:
* The PDF viewer app appears on screen, and shows its "working" animation
* Shortly thereafter, the PDF viewer goes away and the Zulip message list returns to the foreground
* A "toast" (a temporary bubble near the bottom of the screen) appears, with the message "Cannot display PDF ( sagas.pdf cannot be opened)" (where `sagas.pdf` is the filename of the PDF; extra space _sic_)

The difference between servers is presumably because chat.zulip.org hosts uploaded files locally, while zulipchat.com hosts them on S3. In particular on zulipchat.com (and with the S3 backend generally), when the client requests an uploaded file, the server returns a redirect to a URL in S3; we may not be handling the redirect properly.

If so, this may be another symptom of the same underlying issue as the cause of #3124 .

borisyankov:
Fixes #3487
Fixes #3303 (but does not implement the exact suggestion)

Just as we do with images, check if a link is local and is poining
to `/user_uploads/`. If so, adds the `api_key` value to the url
to enable opening it in Chrome Custom Tab / Safari View.

The file can then be downloaded via the oveflow menu (three-dots)
on Android. On iOS, similar actions can be taken via the 'Share'
button available at the bottom of the screen.

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -346,6 +346,24 @@ var appendAuthToImages = function appendAuthToImages(auth) {
   });
 };
 
+var appendAuthToFileLinks = function appendAuthToFileLinks(auth) {
+  var aTags = document.getElementsByTagName('a');
+  arrayFrom(aTags).forEach(function (a) {
+    if (!(a.href && a.href.startsWith(auth.realm))) {
+      return;
+    }
+
+    var hrefPath = a.href.substring(auth.realm.length);
+
+    if (!hrefPath.startsWith('/user_uploads/')) {
+      return;
+    }
+
+    var delimiter = a.href.includes('?') ? '&' : '?';
+    a.href += delimiter + "api_key=" + auth.apiKey;
+  });
+};
+
 var handleUpdateEventContent = function handleUpdateEventContent(uevent) {
   var target;
 
@@ -368,6 +386,7 @@ var handleUpdateEventContent = function handleUpdateEventContent(uevent) {
 
   documentBody.innerHTML = uevent.content;
   appendAuthToImages(uevent.auth);
+  appendAuthToFileLinks(uevent.auth);
 
   if (target.type === 'bottom') {
     scrollToBottom();
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -475,6 +475,23 @@ const appendAuthToImages = auth => {
   });
 };
 
+const appendAuthToFileLinks = auth => {
+  const aTags = document.getElementsByTagName('a');
+  arrayFrom(aTags).forEach(a => {
+    if (!(a.href && a.href.startsWith(auth.realm))) {
+      return;
+    }
+
+    const hrefPath = a.href.substring(auth.realm.length);
+    if (!hrefPath.startsWith('/user_uploads/')) {
+      return;
+    }
+
+    const delimiter = a.href.includes('?') ? '&' : '?';
+    a.href += `${delimiter}api_key=${auth.apiKey}`;
+  });
+};
+
 const handleUpdateEventContent = (uevent: WebViewUpdateEventContent) => {
   let target: ScrollTarget;
   if (uevent.updateStrategy === 'replace') {
@@ -494,6 +511,7 @@ const handleUpdateEventContent = (uevent: WebViewUpdateEventContent) => {
   documentBody.innerHTML = uevent.content;
 
   appendAuthToImages(uevent.auth);
+  appendAuthToFileLinks(uevent.auth);
 
   if (target.type === 'bottom') {
     scrollToBottom();


ishammahajan:
I am certainly new to this, but isn't the `api_key` supposed to be secret? https://en.m.wikipedia.org/wiki/Application_programming_interface_key
> The API key often acts as both a unique identifier and a secret token for authentication, and will generally have a set of access rights on the API associated with it

Any user can access it if given through URL.

borisyankov:
We should definitely strip that on 'Copy URL to clipboard'.

borisyankov:
BTW we already use it for images. The ideal case would be to have the back-end support a scheme that relies on timestamp key that are valid for say 10mins and we salt with the api_key.

borisyankov:
Another security concern is not an issue: if the user copies the URL of the viewed document it will not contain the api_key as it will be redirected to an exipiring s3 URL.

gnprice:
Thanks @borisyankov !

I agree with your diagnosis of the underlying issue, and that the right solution is to fix auth for the existing "open this link" path.

Like @ishammahajan , it does make me a bit nervous to copy the API key into more places, like these links -- if we do this, I'd worry about scenarios where that auth-carrying URL ends up propagating to more places, and the user innocently posts it somewhere for others to see. The fact that these URLs contain a key that gives access to all the user's data, not just the one file, could be a nasty surprise.

> We should definitely strip that part on 'Copy URL to clipboard'.

This would help and could potentially resolve that worry...

... but can we instead do this at another layer, more closely targeted to the desired effect? When the user follows one of these links, the webview just posts an event to the outside, a `url` event... which in `webViewEventHandlers` results in a `messageLinkPress` call... which goes and invokes either `SafariView` or `NativeModules.CustomTabsAndroid.openURL`.

And just before that last step we already call `getFullUrl` to get an appropriate kind of URL -- it's messily implemented but basically tries to turn relative URL references into real absolute URLs.

So, a small improvement over this PR's implementation would be to append the API key in `messageLinkPress`, just before calling `SafariView` or `openURL`.

Better: can we teach either `SafariView` or `openURL` to provide appropriate auth? That's a question that comes down to their respective APIs.

gnprice:
> Another security concern is not an issue: if the user copies the URL of the viewed document it will not contain the api_key as it will be redirected to an exipiring s3 URL.

This applies only with the S3 backend, unfortunately.

With the local-files backend for uploaded files -- which is the default -- I believe there's no redirect, so the URL we tell the browser to open will be the one that stays as the file is viewed.

(Even on the S3 backend, there's still a window where the redirect hasn't happened yet. If the user is on a slow connection (or their Zulip server is having trouble!) and it hangs there for a while, it's easy to imagine them copying/sharing the URL somewhere else while it's still the one with the API key in it.)

gnprice:
Just to highlight the open question I think is most interesting for the next steps of investigation:

> Better: can we teach either `SafariView` or `openURL` to provide appropriate auth? That's a question that comes down to their respective APIs.

The answer will then tell us what sort of workaround we'll end up doing, or settling for.

borisyankov:
> Better: can we teach either SafariView or openURL to provide appropriate auth? That's a question that comes down to their respective APIs.

We can, but likely need back-end support.

I did examine both options:
* passing auth via custom auth headers
* setting a cookie ala the web app

I did read some documentation about both:
* Android Custom Tab https://developer.android.com/reference/android/support/customtabs/CustomTabsIntent.html 
* Safari View Controller https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller

The situation with iOS is bad - we can't manually set a cookie, nor it looks like we can specify custom headers. Cookies are shared with main Safari browsers in the past, not anymore since iOS 11 (so a user logged in their browser could have had this working previously)

Android - maybe no way to set cookies manually, cookies shared with browser though. Also, there is a way to set custom http headers but they are dependent on new Android version. (using Browser.EXTRA_HEADERS)


A universal solution that would depend on browser support is to pass the API somehow and expect the browser as a response to set auth cookies, then proceed to the url.

gnprice:
Thanks @borisyankov for that investigation!

Android first:

> Android Custom Tab https://developer.android.com/reference/android/support/customtabs/CustomTabsIntent.html

Ah, I see now that `CustomTabsAndroid` is a small module of our own. The logic is more convoluted than it needs to be, but the heart of it seems to be `Intent(Intent.ACTION_VIEW, Uri.parse(url))`; docs [here](https://developer.android.com/guide/components/intents-common) on what that does.

> Android - maybe no way to set cookies manually, cookies shared with browser though. Also, there is a way to set custom http headers but they are dependent on new Android version. (using Browser.EXTRA_HEADERS)

Can you say more about this? What docs are you relying on?

The bit of docs I find is this: https://developer.android.com/reference/android/provider/Browser#EXTRA_HEADERS

That says "Added in API level 8", i.e. in ancient times -- so that's actually quite encouraging on the Android side.

borisyankov:
> That says "Added in API level 8", i.e. in ancient times -- so that's actually quite encouraging on the Android side.

That is the part that is confusing. It seems the 'API level 8' is for the `android.provider.Browser`

When looking I found this answer:
https://stackoverflow.com/questions/35419041/how-do-you-add-custom-http-headers-to-chrome-custom-tabs

Maybe the answer is referring to using `Bundle` to pass the extra headers it to the `CustomTabsIntent` as a relatively new feature.

gnprice:
Maybe? Still confused what people are referring to when they mention a new feature.

`CustomTabsIntent` is a pretty thin abstraction, it turns out on reading the code -- it really just boils down to
```
val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
context.startActivity(intent)
```
with optionally some extras set on the `Intent`. Also, `CustomTabsIntent` is in the support library -- which we build into the app -- so its version is entirely up to us and independent of the platform version.

OTOH, this bug is not encouraging:
https://bugs.chromium.org/p/chromium/issues/detail?id=772966

gnprice:
(Here's what looks to be the code that looks at that `EXTRA_HEADERS` extra and interprets it, FWIW:
https://cs.chromium.org/chromium/src/chrome/android/java/src/org/chromium/chrome/browser/IntentHandler.java?l=788&rcl=41eb854d9b0186421b65a347645b80d5ec25d92f
)

gnprice:
Re iOS:

> The situation with iOS is bad - we can't manually set a cookie, nor it looks like we can specify custom headers.

Yeah. :-/


> Cookies are shared with main Safari browsers in the past, not anymore since iOS 11 (so a user logged in their browser could have had this working previously)

Interesting -- can you share what reference you saw that told you about this?

---

Having read through some of the SafariView documentation and reflected on it, I think the preferred solution from the platform's perspective is probably that when we want this kind of control, we should be using a `WKWebView`, not a `SafariView`. The `SafariView` is meant to look like a normal browser -- so it is actually a bit odd, and potentially confusing, if it behaves behind the scenes in subtly different ways.

On a `WKWebView`, we certainly can control the headers; the [`load(_:)` method](https://developer.apple.com/documentation/webkit/wkwebview/1414954-load), which is the main method for "navigate to some URL", takes a `URLRequest`, which [has a dict of HTTP headers](https://developer.apple.com/documentation/foundation/urlrequest). Though it's inconvenient that RN doesn't expose that for us, or at least it didn't when we were dealing with auth in our main webview last year.

I think a similar solution is likely the ideal one on Android, for similar reasons: when viewing an upload or similar file on the Zulip server, do it in a webview.

In both cases that's likely a bit of work to implement, though.

gnprice:
> > Cookies are shared with main Safari browsers in the past, not anymore since iOS 11 (so a user logged in their browser could have had this working previously)
> 
> Interesting -- can you share what reference you saw that told you about this?

Ah, just read more closely [the doc you linked](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller) and I see it now:

> In iOS 9 and 10, it shares cookies and other website data with Safari. The user's activity and interaction with SFSafariViewController are not visible to your app, which cannot access AutoFill data, browsing history, or website data. You do not need to secure data between your app and Safari. If you would like to share data between your app and Safari in iOS 11 and later, so it is easier for a user to log in only one time, use [SFAuthenticationSession](https://developer.apple.com/documentation/safariservices/sfauthenticationsession) instead.

And the point of that last-mentioned feature seems to be to use credentials the user has in the browser to grant their privileges to the app, not vice versa.

Later, it is pretty explicit in recommending the approach I described in the previous comment:

> If your app lets users view websites from anywhere on the Internet, use the `SFSafariViewController` class. If your app customizes, interacts with, or controls the display of web content, use the [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview) class.

borisyankov:
The problem with reimplementing everything in a WebView, apart from being a non trivial task (Android's CustomTab gives the Send file. Open with, Download, Print, functionality for free) is that it is not clear what would be the behavior of opening a file inside a WebView. Will it be opened in the appropriate app or will it behave like something that is contained inside a component. This can be verified experimentally though.

gnprice:
> The problem with reimplementing everything in a WebView, apart from being a non trivial task (Android's CustomTab gives the Send file. Open with, Download, Print, functionality for free)

Yeah. A lazy version of this would be to just have either an "Open in browser..." (or "Open in Chrome...") action, or a "Share..." action; then the user can get to a full-featured browser to do anything else. I find myself doing this regularly as an Android user.

Those actions might not work in the redirect case... but I'm not confident the various actions will work in the redirect case in the CustomTab, either. Another experimental question, I guess.


> is that it is not clear what would be the behavior of opening a file inside a WebView. Will it be opened in the appropriate app or will it behave like something that is contained inside a component. This can be verified experimentally though.

Yeah, definitely agree.

chrisbobbe:
To clear a path forward on this important issue, I'm closing both this PR and #3330, with a solution outlined at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/view.20uploaded.20files/near/819777 that will involve some backend work.

There, Greg lists these three requirements that the new solution fulfills, and offers an implementation plan:

> The goal basically is to get
> * (a) a plain old URL that can be passed to a browser (not a URL plus some cookies or other headers that need to be set)
> * (b) that gets the user the file immediately, without requiring a fresh login
> * (c) but doesn't include a secret with much wider powers (i.e. their API key), which would risk a nasty surprise if they innocently show someone else the file by sending them the link.

PR #3502, opened 2019-05-28
WIP: A very WIP upgrade to React Navigation v3

borisyankov:
Based on and a follow up to #2702


gnprice:
Ah, good thought, thanks!

I think the thing to prioritize here will be braindumping what you've learned about this upgrade. Doing that in code is great where it's easy; in general,
* What are the main things changed in react-navigation v3 that will be relevant to us, or potential pitfalls?
* What questions are still open / what things are you still uncertain about as you're investigating?

That can be in commit messages, or in this PR thread.

borisyankov:
The commit `Add and configure 'react-native-gesture-handler'` is mergeable independently.

borisyankov:
> What are the main things changed in react-navigation v3 that will be relevant to us, or potential pitfalls?

I think the upgrade is straightforward, they lib just tends to churn a bit on the API and some navigation objects have had the format of the config objects changed a bit. Nothing major.

> What questions are still open / what things are you still uncertain about as you're investigating?

The 'very WIP' part comes from the fact that I worked on this PR a long time ago, but now that I have reviewed the code, it needs a bit of cleanup and would likely be a fast and painless merge.

gnprice:
> The commit `Add and configure 'react-native-gesture-handler'` is mergeable independently.

Good to know, thanks!

Is it mergeable independently of the v2 upgrade #2702, or does it depend on that? (It comes after it in this branch.)

And thanks for the other notes too; that's helpful.

borisyankov:
Yeah, it does not depend on anything.
It setups this library:
https://github.com/kmagiera/react-native-gesture-handler

Which is a dependency of RN3. It can be used separately from it (but unlikely to)

chrisbobbe:
Hi, first comment on Zulip code here! :)

One thing I ran into a few months ago with React Navigation 3 (I'm 95% sure it was introduced in version 2 actually), was that if you try to `.navigate(routeName)` or `.push(routeName)`, and `routeName` is not "registered somewhere in the app's router" (https://reactnavigation.org/docs/en/2.x/navigation-prop.html#navigate-link-to-other-screens), then it will **fail with no error or warning**. I don't think the upgrades to 2 or 3 touched any calls to `.push` or `.navigate`, so those commits themselves should be fine, but it's something to know for the future if/when new routes are added (or there are new ways of naming routes; I saw a suggestion to use JSON.stringify: https://github.com/zulip/zulip-mobile/pull/2702#issuecomment-497484938).

It seems a PR was made to React Navigation to add a helpful message or error, but closed with very little ceremony: https://github.com/react-navigation/react-navigation/pull/2140.

Maybe someone has run into this (or you've seen that they've fixed the issue in the last few months)?

gnprice:
Thanks @chrisbobbe !

Looks like the underlying issue for that is https://github.com/react-navigation/react-navigation/issues/1053 , which in turn they copied to https://react-navigation.canny.io/feature-requests/p/throw-warning-or-error-when-navigate-to-an-invalid-route and closed. That is an annoying pitfall, and it sure seems like they should be able to mitigate it with a warning.

For our project, we largely deal with this issue by centralizing the `.push` calls in one file that provides tiny wrappers for them: https://github.com/zulip/zulip-mobile/blob/ca20e816a/src/nav/navActions.js Then most everything else that needs to refer to a route does so by calling one of those wrappers. So as long as we get the spellings right in that file, we get the benefits of type-checking when we use them. (Same story for the parameter types, too.)

chrisbobbe:
Closing as superseded by https://github.com/zulip/zulip-mobile/pull/4249, which has already been merged.

PR #3504, opened 2019-05-29
Remove 'Detox' and related files

borisyankov:
Removes the Detox end-to-end testing library from the project.
https://github.com/wix/Detox

This was introduced in #825

This is a heavy depenency and sometimes the 'yarn' command fails on
installing it (noncritically, but still).

Since we never made any use of it in the last 2 years (still a good
idea, but still, not used) a good idea is to remove it.


gnprice:

PR #3511, opened 2019-06-05
message list: Fix difference in spacing.

### LINKED ISSUE #3508: Differing left spacing for `message-full` and `message-brief` in WebView.
After a recent merge (for pull request #3188), the spacing on the messages on the left hand side got a little wonky. This might be a result of switching sizing to px instead of em, or due to changes in margin/padding.

Screenshot 2019-06-01 at 12 04 04 PM

Notice that a full message has different spacing than a brief message.

@zulipbot claim.

ishammahajan:
The apparent spacing of the left margin is differing between a
`message-full` and a `message-brief` at the time of this
commit. This was due to merging a previous commit which incorrectly
replaced `2rem` to `40px` instead of the `32px` correct value. This
small commit fixes that.

Fixes #3508.

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -114,9 +114,9 @@ hr {
 }
 .avatar,
 .loading-avatar {
-  min-width: 40px;
-  width: 40px;
-  height: 40px;
+  min-width: 32px;
+  width: 32px;
+  height: 32px;
   margin-right: 1rem;
 }
 .avatar img {


gnprice:
Thanks @ishammahajan ! This change looks good (at least for now -- more below), and I'm about to merge it. The debugging and the mention of the lengths 2rem, 40px, and 32px is very helpful.

* The first question I had on reading this commit is: what was that previous commit exactly? We'll definitely want to mention it in the commit message -- that makes very helpful context for someone reading.

So I used the command:
```
git log --stat -p -S 40px src/webview/css/
```
to see commits that added or removed the string `40px` in that directory.

(Actually, I typed `git lsp -S40px src/webview/css/` -- I [have a Git alias](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3) `lsp` for `log --stat -p` which I use constantly, and the space after `-S` is optional.)

The answer is, commit 3dd5fb4de:
```
commit 3dd5fb4de6cd72898178f8238bff81a670186c46
Author: Boris Yankov <borisyankov@gmail.com>

    webview: Make avatars bigger, and not scale with font
    
    After comparing the size of the WebView avatar to the web app and
    few competitors, I found out we are showing them rather small.
    
    Web app has them 35px x 35px, but several other mobile chat apps
    have them at around 40px and that is what we chose.
    
    Also, reuse the same css style between .avatar and .loading-avatar .
```

* Hmm! That commit message is actually quite explicit about the intention -- it intentionally made the avatars bigger, as well as making them independent of font size.

  (This commit was part of #3188, and one revision I asked for [in review](https://github.com/zulip/zulip-mobile/pull/3188#issuecomment-494140195) was that explicitness. Better still would have been to split it into two commits, to completely avoid confusion between the two things it was doing; I often ask for a revision like that, but in this case I didn't. Since the dual-purpose commit did end up confusing at least one person -- namely you -- I guess in retrospect we should indeed have split it up!)

  Because making them bigger caused this regression where things weren't aligned, I'll go ahead and merge this change to revert that part. The net effect is that the avatars don't scale with font size (one of the two things that previous commit did), but they're still a bit small.

* I think Boris's research and reasoning for making them a bit bigger makes sense, so I'd still like to make that change too. I'll follow up about that.

ishammahajan:
> (Actually, I typed `git lsp -S40px src/webview/css/` -- I [have a Git alias](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3) `lsp` for `log --stat -p` which I use constantly, and the space after `-S` is optional.)

So I have copied all of your aliases 😆, hope you don't mind!

> ```
> git log --stat -p -S 40px src/webview/css/
> ```

Thanks for the tip!

> * so I'd still like to make that change too. I'll follow up about that.

I personally don't agree with this, have left comments on the relevant issue.

gnprice:
> So I have copied all of your aliases :laughing:, hope you don't mind!

That's what the gist is for! :slightly_smiling_face: 

> > ```
> > git log --stat -p -S 40px src/webview/css/
> > ```
> 
> Thanks for the tip!

Oh that reminds me -- we have a short doc I wrote with this and a few other [key tips for reading Git history](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md), including the "secret" to using `git log -p` effectively. I predict it will have a few other bits you find useful. :wink: 


> > * so I'd still like to make that change too. I'll follow up about that.
> 
> I personally don't agree with this, have left comments on the relevant issue.

Cool, thanks -- we'll discuss there.

PR #3512, opened 2019-06-05
Mentions: spacing breaks when it appears at the beginning of line.

ishammahajan:
Mentions while writing this PR get cut off if they appear in the
beginning of a message. This commit adds a very slight margin to
mentions to prevent this from happening.

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -140,6 +140,7 @@ hr {
   background-color: rgba(0, 0, 0, 0.1);
   border-radius: 3px;
   padding: 0 2px;
+  margin: 0 1px;
   box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.2);
 }
 .header-wrapper {


ishammahajan:
Before: 


After:

ishammahajan:
@gnprice here is the second fix as discussed in [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/message.20spacing/near/748978).

gnprice:
Thanks @ishammahajan !

* One thing I'm curious about in an issue like this is: why was it getting cut off in the first place? For the most part, the box for the message contents is supposed to hold whatever's there.

  Playing with it, I think a thing that's happening specifically is that the `box-shadow` is getting cut off. We have this:

      box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.2);

  If in Chrome DevTools (which are pretty neat -- do [take a look](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#webview-chrome) at our notes on using those, if you haven't already) I put my cursor on that 1px and hit the "down" key to change it to 0px... there's no longer anything cut off. If instead I hit "up" a few times to make it big, the vertical cutoff line is right at the left edge of the mention pill. If I change the alpha of `0.2` to 1, so the shadow is a sharp black, this all becomes clearer, too.

* Another question I ask is: how did you choose the value 0.1rem?

  I think the box-shadow observation provides a good way to answer that: make it 1px instead :slightly_smiling_face: , explaining that it's to make sure there's room for the shadow. Also the shadow goes in all directions, so probably cleanest to make the margin do so too, with just `margin: 1px`... though vertical margin doesn't seem to actually do anything, I'm guessing because the element has inline layout.

ishammahajan:
> * If in Chrome DevTools

I do use those, @jainkuniya advised me to use them. I didn't know about changing values in there and seeing the changes reflected in real time though, thanks for that, it's going to make my life much easier 😄.

> * how did you choose the value 0.1rem

I didn't put much thought into it, to be honest. I just took the unit of padding and inserted the smallest accepted value of that unit into margin. 1px does sound better though, changed!

gnprice:
> > * If in Chrome DevTools
> 
> I do use those, @jainkuniya advised me to use them. I didn't know about changing values in there and seeing the changes reflected in real time though, thanks for that, it's going to make my life much easier .

Great :smiley: A related tip: if you hover on one of those values while it's focused, you get a reminder about how holding Ctrl, Shift, or Alt modify the effect! They make it +/- 100, 10, or 0.1 respectively instead of 1. (I'd been loosely aware there were modifiers like those, but had never quite remembered which did what until I saw that reminder this week.)


> > * how did you choose the value 0.1rem
> 
> I didn't put much thought into it, to be honest. I just took the unit of padding and inserted the smallest accepted value of that unit into margin. 1px does sound better though, changed!

:+1:

gnprice:
And merged! Thanks again.

Made one substantive change in the commit message: I added the answers to my questions above. :slightly_smiling_face: 

    msglist: Fix @-mention pills getting cut off.
    
    The pill around an @-mention gets slightly cut off at the left if it
    appears at the beginning of a message.  Fix that by adding a very
    slight margin... equal to the size of the box-shadow, because that
    turns out to be the part that's getting cut off.

Issue #3516, opened 2019-06-06
Convert all colors in our CSS to HSL


gnprice:
(Expanded from https://github.com/zulip/zulip-mobile/pull/3491#issuecomment-498802902 .)

The webapp makes a practice of using HSL instead of RGB (there are a few exceptions, but I think those are where people just didn't know and it wasn't caught in review.)

There are good reasons for this, because it makes it easier to understand what the color is, and also to make changes to it. See [here](https://stackoverflow.com/questions/26059228/css-hsl-or-rgba-colors) and [here](https://css-tricks.com/hsl-hsla-is-great-for-programmatic-color-control/) for some good explanations.

We aren't currently consistent about this ourselves -- in the CSS for the message list WebView, it's about half and half, ignoring colors that are just flat gray. Let's convert to all HSL.

Here's a search that I think finds all the colors expressed as RGB:
```
$ git grep -Phoi -e 'rgb.*?\)' -e '#[0-9a-f]+' src/webview/css/ | perl -pe 's/, ?/, /g' | sort
```
Or to get them in context / with line numbers:
```
$ rg -i -e 'rgb.*?\)' -e '#[0-9a-f]+' src/webview/css/
```

A way to go about this:
* First, convert all the zero-saturation colors, where R == G == B. These will be pretty mechanical, I think, so one or a handful of commits can sweep through them. As the first command (the one with `grep -o` and `sort`) shows, these are a large fraction of the total.
* Then, take one color at a time (or a few closely-related colors, like `#008000` and `#00a000`), and convert it everywhere. Find an RGB->HSL converter on the web to compute the new values. Be sure to identify how you did the conversion (e.g. link to the converter), in the PR and at least the first commit message.

gnprice:
A good follow-up would be to do the same for colors in our RN code. The syntax is basically the same:
  https://facebook.github.io/react-native/docs/colors

Let's do that as a separate series of commits, though -- the diffs will look a little different, so I think it'll be easier to read that way.

ishammahajan:
#3519 fixes this issue (commenting for notification).

PR #3519, opened 2019-06-07
RGB > HSL

ishammahajan:
These three commits change the color schema as described in #3516.

The second commit rounds off the colors in `css.js` and there is no equivalent commit for rounding off values used in the rest of the app. My reasoning for such is that in CSS.js it's quite helpful for developers to see the changes made in the color instantly in the file itself (using the extension 'Color Highlight'), mostly because the 'hot reload' feature in react-native doesn't apply to WebViews. Hot reloading however works well in react-native environments and hence doesn't need to be rounded off in order to see the color changes properly and fast. In any case, the second commit is optional, and may be dropped depending on whether the maintainers are okay with the very slight color changes caused due to round off (±0.5 in every case).

### DIFF
--- a/docs/howto/editor.md
+++ b/docs/howto/editor.md
@@ -29,19 +29,8 @@ codebase. Each extension page has install instructions at the top.
 * [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
 * [prettier-vscode](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)
 
-For the extensions to work properly, make sure that VS Code has recognized
-[our settings file](https://github.com/zulip/zulip-mobile/tree/master/.vscode/settings.json)
-in the zulip-mobile repository. Usually, this should happen automatically
-when you open the `zulip-mobile` folder in VS Code. You can verify that the
-settings were recognized by opening your Workspace Settings with
-`Ctrl+Shift+P` -> `Open Workspace Settings`. If the editor on the right-hand
-side starts with the comment
-```js
-// Workspace Settings for the zulip-mobile repo.
-```
-you're all set - VS Code has recognized your workspace settings. If not,
-make sure that you opened VS Code as a folder. Do this by pressing
-`Ctrl+K Ctrl+O` and navigating to your local `zulip-mobile` clone.
+If something doesn't seem right, see the items under "Troubleshooting" below.
+
 
 ### Getting familiar with VS Code
 
@@ -64,7 +53,27 @@ Take a look through some of VS Code's docs.  In particular:
   enable this feature, and configure it to use our standard formatting
   rules.
 
+
 ### Troubleshooting
+
+#### Confirm our workspace settings took effect
+
+For the extensions to work properly, make sure that VS Code has recognized
+[our settings file](https://github.com/zulip/zulip-mobile/tree/master/.vscode/settings.json)
+in the zulip-mobile repository. Usually, this should happen automatically
+when you open the `zulip-mobile` folder in VS Code.
+
+You can verify that the settings were recognized by opening your
+Workspace Settings with `Ctrl+Shift+P` -> `Open Workspace
+Settings`. If the editor on the right-hand side starts with the
+comment
+```js
+// Workspace Settings for the zulip-mobile repo.
+```
+you're all set - VS Code has recognized your workspace settings. If not,
+make sure that you opened VS Code as a folder. Do this by pressing
+`Ctrl+K Ctrl+O` and navigating to your local `zulip-mobile` clone.
+
 
 #### Saving causes bad reformatting of code
 
--- a/docs/howto/editor.md
+++ b/docs/howto/editor.md
@@ -31,6 +31,11 @@ codebase. Each extension page has install instructions at the top.
 
 If something doesn't seem right, see the items under "Troubleshooting" below.
 
+Perhaps try the following extensions too, which can be handy:
+
+* [Color Picker](https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color)
+* [Color Highlight](https://marketplace.visualstudio.com/items?itemName=naumovs.color-highlight)
+
 
 ### Getting familiar with VS Code
 
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -13,7 +13,7 @@ const styles = StyleSheet.create({
   accountItem: {
     flexDirection: 'row',
     alignItems: 'center',
-    backgroundColor: 'hsla(177.1, 69.7%, 46.7%, 0.1)',
+    backgroundColor: 'hsla(177, 70%, 47%, 0.1)',
     borderRadius: 4,
     height: 72,
   },
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -13,7 +13,7 @@ const styles = StyleSheet.create({
   accountItem: {
     flexDirection: 'row',
     alignItems: 'center',
-    backgroundColor: 'rgba(36, 202, 194, 0.1)',
+    backgroundColor: 'hsla(177.1, 69.7%, 46.7%, 0.1)',
     borderRadius: 4,
     height: 72,
   },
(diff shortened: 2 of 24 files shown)


agzuniverse:
It seems like the whenever `rgb` form of representing color is present it used the full `rgba` form and sets alpha to 1, when it could have been omitted, and I actually like that approach. Maybe similarly the full `hsla` form should be used everywhere for consistency? What do you think @jainkuniya ?

gnprice:
Thanks @ishammahajan !

* On this question (@agzuniverse):

> It seems like the whenever `rgb` form of representing color is present it used the full `rgba` form and sets alpha to 1, when it could have been omitted, and I actually like that approach. Maybe similarly the full `hsla` form should be used everywhere for consistency?

I looked around for CSS style guides on the web, and didn't find any that had a recommendation on whether to say `hsl` vs. `hsla`. (Or `rgb` vs. `rgba`.) Then I looked at the webapp:
```
$ grep -PRho -i -e 'hsl.*?\)' ../zulip/static/styles/ | sort
```
With a handful of exceptions, the style there is quite consistent that when possible, it's `hsl` with three arguments, and when that's not possible (when alpha is different from 1), it's `hsla` with four. So let's follow the same style. (That's also what the current version of this PR does.)


#### b0311919f css: Change `rgb(a)` and `#` colors to `hsl(a)`.

    [...] this commit changes all colors present in
    the CSS file to match the HSL standard -- by using the extension
    available in the market called 'Color Picker' which also allows for
    conversion of color (type 'Convert Color' in the command palette).

Thanks! Is this referring to a VS Code extension? If so, is it this one?
https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color
If so, let's call it "the VS Code extension `anseki.vscode-color`" -- that will be a good unambiguous way to identify it.

Hmm I see -- and that does look like a super handy tool for this purpose! :-)
https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color#convert-color
Good find.


#### c2f757f3b css: Round off HSL internal values.

* Looks good. This commit strategy is a good one, too -- do the mechanical change that a tool can automatically do for you, commit that verbatim, then make cleanups in a separate commit.
  * (We generally avoid merging commits that we know break something, so when the tool's output is actually broken then that's an exception: we fix it in the same commit.)

* I'm curious about this extension!

      [...] in order
      to make them more readable and functional for the vscode extension
      'Color Highlight'.

  Is it this one? https://marketplace.visualstudio.com/items?itemName=naumovs.color-highlight


#### 41f44c733 Change color scheme from RGB or HEX to HSL.

* Cool. Was this also done with the "Color Picker" extension?

* Want to round these off as well? (Again as a separate commit.)

ishammahajan:
Thanks for the review @gnprice!

> So let's follow the same style.

Sure.

> is it this one?
> [marketplace.visualstudio.com/items?itemName=anseki.vscode-color](https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color)

Yes, that is the one. Will mention it in the commits.

> * Is it this one? [marketplace.visualstudio.com/items?itemName=naumovs.color-highlight](https://marketplace.visualstudio.com/items?itemName=naumovs.color-highlight)

Correct. I was concerned whether the changes in values will be accepted by you though, since they might change the look -- just for the sake of developer productivity. Should I go ahead and try to make the fix upstream to accept decimal values?



> * Cool. Was this also done with the "Color Picker" extension?

Yep!

> * Want to round these off as well? (Again as a separate commit.)

As mentioned in OP, 

> The second commit rounds off the colors in `css.js` and there is no equivalent commit for rounding off values used in the rest of the app. My reasoning for such is that in CSS.js it's quite helpful for developers to see the changes made in the color instantly in the file itself (using the extension 'Color Highlight'), mostly because the 'hot reload' feature in react-native doesn't apply to WebViews. Hot reloading however works well in react-native environments and hence doesn't need to be rounded off in order to see the color changes properly and fast. In any case, the second commit is optional, and may be dropped depending on whether the maintainers are okay with the very slight color changes caused due to round off (±0.5 in every case).

gnprice:
> I was concerned whether the changes in values will be accepted by you though, since they might change the look -- just for the sake of developer productivity.

Ah, I see. Two answers, I guess:
* In general, in the Zulip project we take developer productivity as a high priority! It's critical to how we're able to make an open-source project like this work, and more broadly one thing I take from my experience in the industry (and Tim has a similar view on this) is that taking developer productivity seriously can make a huge difference to the quality of the software written.

* In this case, I think I'd prefer the rounded versions even if the tools worked just as well with non-integers.
  * The tenths place doesn't reflect any intention that anyone had when writing this code -- they were working in whole integers, just in a different coordinate system.
  * Most conspicuously, nobody decided "let's make this a 49.8% gray" -- what they surely had in mind was "let's make this a 1/2 gray", and `rgb(127, 127, 127)` was just the convenient way to close-enough express that in the syntax they were using. So rounding 49.8% to 50% actually *restores* the original intention of the code.
  * If you grep for colors in the webapp:
    ```
    $ perl -nle 'print $1 for (/hsla?\((.*?)\)/g)' ../zulip/static/styles/*.scss | sort -n | uniq -c

    # or more concisely, just hues:
    $ perl -nle 'print $1 for (/hsla?\((.*?),/g)' ../zulip/static/styles/*.scss | sort -n | uniq -c
    ```

    you see that they're all in integers. In particular, you'll see that 170 is a widespread hue -- and 169.8 is not! So again nobody was thinking "hue 169.8, saturation 47.9%, lightness 54.1%" when they wrote `rgba(82, 194, 175, 0.25)` -- they were surely thinking of the color that the webapp specifies as `hsl(170, 48%, 54%)`.

You've probably observed that it's common for me to notice that some commit in a PR will cause a change in behavior, and to ask for revisions related to that. What I'm really going for there is that there shouldn't be an *accidental*, or unexplained, or otherwise hard-to-understand change.

So that's why I'm pleased with the strategy in this PR of first a commit that may make slightly ugly code but is 100% mechanical and surely causes no perceptible behavior change -- and then a separate second commit that makes other adjustments and might slightly change the behavior. Isolating the behavior changes in a separate commit from the behavior-preserving pure refactor (i.e. the change of units, with that extra digit of precision to keep the values practically equal) is very often a good strategy for making both steps easier to understand, and I think this is an example of that.


> > * Want to round these off as well? (Again as a separate commit.)
> 
> As mentioned in OP,
> 
> > The second commit rounds off the colors in `css.js` and there is no equivalent commit for rounding off values used in the rest of the app. My reasoning for such is that in CSS.js it's quite helpful for developers to see the changes made in the color instantly in the file itself (using the extension 'Color Highlight'), mostly because the 'hot reload' feature in react-native doesn't apply to WebViews. Hot reloading however works well in react-native environments and hence doesn't need to be rounded off in order to see the color changes properly and fast. In any case, the second commit is optional, and may be dropped depending on whether the maintainers are okay with the very slight color changes caused due to round off (±0.5 in every case).

Ah sorry -- I read the commits (including the commit messages) but apparently didn't notice this text in the PR description. Thanks for pointing back at it. :slightly_smiling_face: 

I think my explanation above applies to this, then. As an especially salient example:
```
-export const BRAND_COLOR = 'rgba(82, 194, 175, 1)';
+export const BRAND_COLOR = 'hsl(169.8, 47.9%, 54.1%)';
```
One of the most common colors in the webapp styles is `hsl(170, 48%, 54%)` -- this slight variation of it doesn't appear at all. :wink: So that's the color that was actually intended here in the first place.

ishammahajan:
Thanks for the explanation, @gnprice !
> * in the Zulip project we take developer productivity as a high priority!

That makes sense.

> * If you grep for colors in the webapp:
>   ```
>   $ perl -nle 'print $1 for (/hsla?\((.*?)\)/g)' ../zulip/static/styles/*.scss | sort -n | uniq -c
>   
>   # or more concisely, just hues:
>   $ perl -nle 'print $1 for (/hsla?\((.*?),/g)' ../zulip/static/styles/*.scss | sort -n | uniq -c
>   ```

>   you see that they're all in integers.

That also makes a lot of sense, should have realised this beforehand.

I'll round the colors and push an update ASAP, then, thanks!

ishammahajan:
> `$ grep -PRho -i -e 'hsl.*?\)' ../zulip/static/styles/ | sort`

Just as a note, thanks for these tips about command line instructions you keep giving. They help a lot! 😄 
__
Rebased onto master and ready for a merge!

gnprice:
And merged. Thanks again @ishammahajan !


> Just as a note, thanks for these tips about command line instructions you keep giving. They help a lot! :smile: 

Glad to hear it :smile: Sure thing.

PR #3527, opened 2019-06-18
time pills: Make slide-in time pills quieter.

ishammahajan:
Shadows on time slide in timepills have been modified to be centric
instead of leaning toward the left. The changes in the shadow could
be found in the material.io design guidelines/source code, with 2dp
elevation
https://github.com/material-components/material-components-web/blob/master/packages/mdc-elevation/_mixins.scss.

The background matches the default message-list background, but has
a little more added contrast in order to differenciate the top of
the pill with the message list.

@gnprice here's the time pills edit asked for in [chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/message.20timestamps/near/754237)! (screenshots present there)

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -105,6 +105,11 @@ hr {
 .message-brief {
   padding: 0 16px 16px 64px;
 }
+.static-timestamp {
+  color: #999;
+  font-size: 0.9rem;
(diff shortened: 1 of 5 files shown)


gnprice:
Merged -- thanks @ishammahajan !

Issue #3526, opened 2019-06-18
master build error


dkvbabu:
HI,

I am facing below problem while running react-native run-android

> Task :app:processDebugResources FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:processDebugResources'.
> Android resource linking failed
  /home/ubunto/.gradle/caches/transforms-1/files-1.1/drawee-1.10.0.aar/bfe2e3ba3748905ed6faebb6039cbe7e/res/values/values.xml:3:5-58:857: AAPT: error: resource android:attr/fontVariationSettings not found.
      
  /home/ubunto/.gradle/caches/transforms-1/files-1.1/drawee-1.10.0.aar/bfe2e3ba3748905ed6faebb6039cbe7e/res/values/values.xml:3:5-58:857: AAPT: error: resource android:attr/ttcIndex not found.
      
  error: failed linking references.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 33s

Please look into this and reply.

Regards
dk venkatesh babu

wsupajirakit:
I found the same problem.

gnprice:
Hello! Thanks @dkvbabu for filing this.

I see the same thing on my own machine, and we're seeing it also in CI. The exact same commit that's in master now built just fine when I originally pushed it:
https://travis-ci.org/zulip/zulip-mobile/builds/546002328
so this seems to be a change somewhere in our dependencies. It's unfortunate that we seem to be exposed to a change uploaded by dependency's maintainer causing our own build to behave differently and fail.

I'm looking into fixing the immediate issue now.

gnprice:
Update: this isn't totally trivial to fix, awkwardly.

The required step to fix it seems to be to bump `compileSdkVersion` to 28:
https://github.com/flutter/flutter/issues/27226#issuecomment-458650933
Apparently the cause is that some dependency (`drawee`, I guess) began to use, at compile time, a feature introduced in API 28.

And indeed, that resolves this error.

---

But then there's another error (and using `cd android && ./gradlew :assembleDebug --continue` to have the build plow past the first error, one can confirm that this error is already present on master, so it's not caused by the `compileSdkVersion` bump):
```
* What went wrong:
Execution failed for task ':app:mergeExtDexDebug'.
[...]

PR #3529, opened 2019-06-18
Avatar sizes increase to 48px globally.

ishammahajan:
Changes the avatar sizes found everywhere in the application to 48px. Since this is the size used by most other comparable messaging applications, 48px seems like a good option to go for (other reasons explained in the commit/s).

In the first commit `.message-brief` is set to `padding: 0 16px 16px 73px`. While this 73 might seem like an arbitrary number, it's confusing to me and I don't know why this is the case, but that is the only one which makes the messages perfectly aligned.

### DIFF
--- a/src/pm-conversations/GroupPmConversationItem.js
+++ b/src/pm-conversations/GroupPmConversationItem.js
@@ -9,7 +9,7 @@ import styles from '../styles';
 const componentStyles = StyleSheet.create({
   text: {
     flex: 1,
-    marginLeft: 8,
+    marginLeft: 16,
     marginRight: 8,
   },
 });
@@ -46,7 +46,7 @@ export default class GroupPmConversationItem extends PureComponent<Props> {
     return (
       <Touchable onPress={this.handlePress}>
         <View style={styles.listItem}>
-          <GroupAvatar size={32} names={allNames} />
+          <GroupAvatar size={48} names={allNames} />
           <RawLabel
             style={componentStyles.text}
             numberOfLines={2}
--- a/src/users/UserItem.js
+++ b/src/users/UserItem.js
@@ -10,7 +10,7 @@ const componentStyles = StyleSheet.create({
     backgroundColor: BRAND_COLOR,
   },
   text: {
-    marginLeft: 8,
+    marginLeft: 16,
   },
   selectedText: {
     color: 'white',
@@ -54,7 +54,7 @@ export default class UserItem extends PureComponent<Props> {
       <Touchable onPress={this.handlePress}>
         <View style={[styles.listItem, isSelected && componentStyles.selectedRow]}>
           <UserAvatarWithPresence
-            size={32}
+            size={48}
             avatarUrl={avatarUrl}
             email={email}
             onPress={this.handlePress}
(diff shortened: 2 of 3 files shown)


ishammahajan:
(The build failing doesn't have anything to do with the changes made in this code, it seems like. My other PRs also have the exact same error happening.)

Screenshots:

Screenshot_20190618-232202
Screenshot_20190618-232126
Screenshot_20190618-232159
Screenshot_20190618-232038

gnprice:
> (The build failing doesn't have anything to do with the changes made in this code, it seems like. My other PRs also have the exact same error happening.)

(Yup -- see also #3526 .)

gnprice:
> In the first commit `.message-brief` is set to `padding: 0 16px 16px 73px`. While this 73 might seem like an arbitrary number, it's confusing to me and I don't know why this is the case, but that is the only one which makes the messages perfectly aligned.

Huh, puzzling. That shouldn't be.

And in fact, when I run this, it looks slightly misaligned to me. Here's a screenshot, super zoomed in (a feature of the Android emulator -- Ctrl+Z, or the "zoom" icon in the emulator toolbar):
image

The PM conversations screen looks great, though!

gnprice:
Oh wow, here's a clue about the alignment issue, though -- in an emulated Nexus 5, the avatar winds up being some funny number of pixels wide! Here's the `.avatar` element:
image

Not 48px, but rather 38.828px.

(The actual avatar, the `img`, then gets scaled to that width and the corresponding height.)

I'm not sure what's controlling that... but that will be something for you to debug! :wink: 

One related odd thing I notice: we set `min-width`. If that's doing anything at all, either before or after this change, then it's probably at least a clue as to what's going on here.

gnprice:
In the second commit:
```
@@ -42,7 +42,7 @@ class UserAvatarWithPresence extends PureComponent<Props> {
   static defaultProps = {
     avatarUrl: '',
     email: '',
-    size: 32,
+    size: 48,
```

Why change the default?

Are there other call sites this affects? If so, we'll want to look and make sure the effect is something we want.

ishammahajan:
> One related odd thing I notice: we set `min-width`. If that's doing anything at all, either before or after this change, then it's probably at least a clue as to what's going on here.

You're right! I don't know why, but setting `min-width: 48` sets the width properly. I suspect it has something to do with the width of the phone (and hence the viewport), because on my device the width of the avatar before I changed `min-width` was ~ 40.821px. Something to do with [used value](https://developer.mozilla.org/en-US/docs/Web/CSS/used_value) in css, I think.

Checking [width](https://developer.mozilla.org/en-US/docs/Web/CSS/width) on MDN -
> The min-width and max-width properties override width.

Which doesn't appear to be true, because if you remove the `width` property, the avatars become huge -- I believe the uploaded picture's width, and consequently the height (which is set to 48px!).  In any case, setting width to a value should set the width of the element in question as that value, without consideration! It's rather strange; I'm not able to figure out how the CSS works here.

> Why change the default?
> 
> Are there other call sites this affects?

Shouldn't have changed the default. It does affect multiple sites, but at all call sites, for whatever reason, we override the default (to the same value as the default 😆). I've removed the change in the default size.

gnprice:
Thanks @ishammahajan for the PR and these updates! Looks good -- just merged.

PR #3530, opened 2019-06-18
codehilite: Update values for syntax highlighting from webapp.

### LINKED ISSUE #3524: JSON quote rendering is not colorblind friendly
I can't read the red text in
Screenshot_20190617-081433

ishammahajan:
This commit essentially takes the colors used in the webapp's SCSS
file for syntax highlighting (`pygments.scss`), and copies that into
the application. This makes the colors between the webapp and the
mobile application exactly the same and hence provides better parity
of experience!

Fixes #3524.

Relevant File from the webapp: https://github.com/zulip/zulip/blob/master/static/styles/pygments.scss
This has also been linked through a comment in the code.

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -1,6 +1,7 @@
 /* @flow strict-local */
 import type { ThemeName } from '../../types';
 import { BRAND_COLOR } from '../../styles';
+import cssPygments from './cssPygments';
 import cssEmojis from './cssEmojis';
 import cssNight from './cssNight';
 
@@ -317,17 +318,6 @@ blockquote {
 .message ul + p {
   margin-top: 16px;
 }
-.codehilite .gi { color: hsl(120, 100%, 31%); }
-.codehilite .gd { color: hsl(0, 100%, 31%); }
-.codehilite .k { color: hsl(120, 100%, 25%); font-weight: bold; }
-.codehilite .kd { color: hsl(120, 100%, 25%); font-weight: bold; }
-.codehilite .nf { color: hsl(240, 100%, 50%); }
-.codehilite .s2 { color: hsl(0, 70%, 43%); }
-.codehilite .cp { color: hsl(39, 100%, 37%); }
-.codehilite .kt { color: hsl(338, 100%, 35%); }
-.codehilite .nc { color: hsl(240, 100%, 50%); font-weight: bold; }
-.codehilite .nb { color: hsl(120, 100%, 25%); }
-.codehilite .s1 { color: hsl(0, 70%, 43%); }
 .twitter-tweet {
   border: 2px solid hsla(203, 89%, 53%, 0.5);
   background: hsla(203, 89%, 53%, 0.1);
@@ -498,6 +488,7 @@ export default (theme: ThemeName) => `
 <style>
 ${cssBase}
 ${theme === 'night' ? cssNight : ''}
+${cssPygments(theme === 'night')}
 ${cssEmojis}
 </style>
 <style id="style-hide-js-error-plain">
--- a/src/webview/css/cssPygments.js
+++ b/src/webview/css/cssPygments.js
@@ -18,9 +18,6 @@ export default (isNightMode: boolean): string => `
 .codehilite .hll {
     background-color: hsl(60, 100%, 90%);
 }
-.codehilite {
-    background-color: hsl(0, 0%, 98%);
-}
 .codehilite .c {
     color: hsl(180, 33%, 37%);
     font-style: italic;
@@ -232,11 +229,7 @@ ${
     .codehilite code,
     .codehilite pre {
         color: hsl(212, 100%, 82%);
-        background-color: hsl(212, 28%, 18%);
-    }
-
-    .codehilite {
-        background-color: hsl(212, 25%, 17%);
+        background-color: hsl(212, 25%, 15%);
     }
 
     .codehilite .hll {
--- a/src/webview/css/cssPygments.js
+++ b/src/webview/css/cssPygments.js
@@ -1,0 +1,414 @@
+/* @flow strict */
+
+/**
+ * This entire code has been copied near verbatim from the WebApp's `pygments.scss` file.
+ * https://github.com/zulip/zulip/blob/master/static/styles/pygments.scss.
+ * The main change is in the logic of the `isNightMode` flag. The reason why we have to do
+ * that is because the WebApp uses SCSS, which can detect this natively, while we use CSS.
+ *
+ * Use this command to get the diff between the WebApp code vs this file for debugging:
+ * `git diff --no-index ../zulip/static/styles/pygments.scss src/webview/css/cssPygments.js`
+ */
+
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks @ishammahajan !

I like the idea of splitting this out as its own source file -- that helps a lot in comparing it against the webapp version. (Plus it helps in keeping things organized, just because it's so long.)

One more thing that helps when copying something near verbatim like this: try to make it so that the diff when comparing them is as small as possible. For instance:
```
$ git diff --no-index ../zulip/static/styles/pygments.scss src/webview/css/cssCodeColors.js
```

Ideally that would be just a few lines at the top, with the handy comment; and a line or two in the middle where the logic happens, and at the end.

Can you try to tweak this so that that's the case? Sometimes that means turning off linters, because the webapp has a different set of lint configurations; see our `processAlertWords.js` for an example where we did that.

Similarly: how about giving this a name like `cssPygments.js`? That just makes it a little easier to see (before even reading the file itself) that it might correspond to the `pygments.scss` in the webapp.

ishammahajan:
> One more thing that helps when copying something near verbatim like this: try to make it so that the diff when comparing them is as small as possible. For instance:
> 
> ```
> $ git diff --no-index ../zulip/static/styles/pygments.scss src/webview/css/cssCodeColors.js
> ```

Ah, thanks for the handy git command! Helped me while refactoring the file :)
Also I agree! I didn't think of this before, but even if the linter gave errors (it didn't) we should disable it in order to properly compare it with the webapp version at a later point of time. 

> Can you try to tweak this so that that's the case?

Sure, its done! I noticed there was a diff I had added in order to fix a bug that was introduced by pure copy paste (which I'm assuming the webapp takes care of in another manner) which I hadn't noted down in the commit! So that was taken care of as well.

> Similarly: how about giving this a name like `cssPygments.js`?

Hmm, I see how giving this name could be appealing. We don't use [Pygments](pygments.org) though, so I don't know how I feel about it. I agree with the general sentiment though, so the file name has been changed.

gnprice:
Excellent, thanks!

    There was an issue while copy-pasting the code from the webapp,
    where the background-color on a piece of code written in Night mode
    was white, which is why there is a little diff [...]

Hmm, interesting. One thing I notice in that diff:

```
     .codehilite code,
     .codehilite pre {
         color: hsl(212, 100%, 82%);
         background-color: hsl(212, 25%, 15%);
     }
 
+    .codehilite {
+        background-color: hsl(212, 25%, 17%);
+    }
```

is that there's a very similar `background-color` just above it! And set for selectors `.codehilite code, .codehilite pre`, which sound like they're meant to be quite broad.

How did you pick 17% vs the 15% there?

Also, I guess: is that `.codehilite code` etc. rule really applying in the webapp? If so, why is it not applying for us?

(If it's *not*, then that's quite interesting and should be fixed, possibly by just deleting it in the webapp code if the effect that's currently happening is what we decide we want.)

gnprice:
* Small style comment: Let's provide that `git diff` command line right there in the comment at the top of the file. That way it's maximally easy for someone to reproduce it in the future and update as needed.

* > We don't use Pygments though

  Hmm, do we not? There's certainly a bunch of mentions of it in the server code. Remember that the HTML we use for the message contents (and that this CSS is styling) comes from the server... so whatever *is* involved in looking at some code in a message and assigning the CSS classes to highlight it for syntax, it's something that runs on the server.

ishammahajan:
> Also, I guess: is that `.codehilite code` etc. rule really applying in the webapp? If so, why is it not applying for us?

So, the problem is this.
Screenshot 2019-07-01 at 2 13 24 AM

The `codehilite` div is the outermost div (just under `.content`, inside which `pre` and `code` exists. The `pre` and `code` render properly, but because they are slightly rounded on the edges, this happens.
Screenshot 2019-07-01 at 2 11 54 AM

Which is why we add the `.codehilite { ... }` in order to fill that whiteness. My reasoning before was different, but in the recent push I have made a change to `background-color: hsl(212, 28%, 18%);` because it makes sense to put the background-color as the same the rest of the app contains -- I am assuming this background-color change is done by another manner in the webapp (since they use SCSS, which I am not familiar with, I can't directly tell). (previously the reasoning was that if you see the light mode colors in the same file `cssPygments`, the difference between `.codehilite { ... }` and `.codehilite pre { ... }` is of 2% if copied directly from the webapp, but I now realize that was stupid.)
Screenshot 2019-07-01 at 2 14 03 AM

ishammahajan:
> * Small style comment: Let's provide that `git diff` command line right there in the comment at the top of the file. That way it's maximally easy for someone to reproduce it in the future and update as needed.

Sure, that makes sense. Thanks!

> * , do we not? There's certainly a bunch of mentions of it in the server code.

Ah, we do, and that's from the server. Turns out I didn't understand what `pygments` was properly. Thanks for the explanation!

ishammahajan:
Just bumping this as well.

gnprice:
Thanks for the reminder :wink: , as well as the updates!

* Nit: this should say `cssPygments.js`:
```
 * `git diff --no-index ../zulip/static/styles/pygments.scss src/webview/css/cssCodeColors.js`
```

So here's the substantive diff in the current version:
```
// if isNightMode:
     .codehilite code,
     .codehilite pre {
         color: hsl(212, 100%, 82%);
-        background-color: hsl(212, 25%, 15%);
+        background-color: hsl(212, 28%, 18%);
+    }
+
+    .codehilite {
+        background-color: hsl(212, 25%, 17%);
     }
```

* I think it looks better the way the webapp has it -- the background is darker than the general background -- so I'd like to find a way to follow that.

> The `codehilite` div is the outermost div (just under `.content`, inside which `pre` and `code` exists. The `pre` and `code` render properly, but because they are slightly rounded on the edges, this happens. [screenshot]
> 
> Which is why we add the `.codehilite { ... }` in order to fill that whiteness. [...] I am assuming this background-color change is done by another manner in the webapp (since they use SCSS, which I am not familiar with, I can't directly tell).

Hmm, let's investigate.

Inspecting one of these in the webapp in the Chrome devtools, here's what I see:
image

So in particular
```
.codehilite {
￼    display: block !important;
￼    border: none !important;
￼    background: none !important;
}
```
and that's coming from `rendered_markdown.scss` line 411. (There's an ellipsis in the screenshot, but on mouseover it shows the full name.)

Indeed, here's the relevant snippet from `rendered_markdown.scss`:
```
.codehilite {
    display: block !important;
    border: none !important;
    background: none !important;
}
```

The `!important` feels like a bit of a crude hack to override `pygments.scss` -- I suspect whoever wrote that block didn't 100% understand how the styles worked either. :stuck_out_tongue: . But the simple version is we can just leave out `background-color` on a plain `.codehilite`, in both themes.

Then we should be able to leave the colors just as they are in the webapp.

(A nice bonus followup would be to make that `.codehilite { ... }` change in the webapp, too.)

gnprice:
OK, made those changes and merged! Thanks again.

The trivial comment fix (and another like it) I made by editing the commit; then the substantive color fix I made as a new commit on top.

Issue #2447, opened 2018-05-03
Bundle a font in the app and use it consistently


gnprice:
There are a lot of spots in the app where our layout is affected by nuances of the font used.
* An extreme example is #2365 -- a component we get from one of our dependencies comes out completely broken with the default font on certain devices.
* A more subtle example was uncovered in chat at [#mobile > input height discrepancy](https://chat.zulip.org/#narrow/stream/48-mobile/subject/input.20height.20discrepancy/near/572286). Fonts differ in how much vertical padding they build in -- for the sake of ascenders and descenders in Latin letters, or more so for some other scripts like देवनागरी that use more vertical space -- and that can cause the same text widget, with all the same properties right down to the font size, to lay out with a different height when given a different font. That in turn makes it quite difficult to design a layout that consistently looks right.

In the web app, we solve this by specifying a specific font, and providing the font in case the browser doesn't have it. We should do the same thing in the mobile app.

In the web app, the font we use is Source Sans Pro. We could use that, or another free font -- see the [Google font directory](https://fonts.google.com/?category=Sans+Serif) for many choices, but two particularly likely candidates would be [Noto](https://www.google.com/get/noto/) and Roboto. 

A quick feasibility check: the full download for Noto Sans is 16MB, but that's easily cut down to 2MB by dropping exotic variants like "SemiCondensedLightItalic", or 1MB with modern compression (`xz`). I think even my 1MB list has more variants than we probably need.

chrisbobbe:
https://github.com/zulip/zulip-mobile/issues/3885 is targeted at older Android devices to get all emoji glyphs to display. It outlines a way to fetch a font dynamically on Android, instead of bundling it with the app.

gnprice:
Ah, thanks for drawing that connection! I'd forgotten we had this old issue filed.

One difference between the situation for emoji and for text is this:

> A quick feasibility check: the full download for Noto Sans is 16MB, but that's easily cut down to 2MB by dropping exotic variants like "SemiCondensedLightItalic", or 1MB with modern compression (`xz`). I think even my 1MB list has more variants than we probably need.

Whereas the Noto emoji are a single font file `NotoColorEmoji.ttf`, and it's 9.4 MB (in the latest/largest version I have lying around, from the Android SDK v29 aka for Android 10) and compresses down only to 8.7 MB.

[...]

PR #3533, opened 2019-06-20
settings card: Add header - Label and separator.

ishammahajan:
The settings card by itself looks a little empty. @timabbot
suggested that this was due to a label being absent from the
settings screen.
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/mascot/near/756544

Also added a separator to make things cleaner. Used a `Label`
component for including translations.

### DIFF
--- a/src/settings/SettingsCard.js
+++ b/src/settings/SettingsCard.js
@@ -14,6 +14,7 @@ import {
   IconLanguage,
   IconMoreHorizontal,
 } from '../common/Icons';
+import ModalNavBar from '../nav/ModalNavBar';
 import {
   settingsChange,
   navigateToNotifications,
@@ -44,6 +45,7 @@ class SettingsCard extends PureComponent<Props> {
 
     return (
       <ScrollView style={styles.optionWrapper}>
+        <ModalNavBar canGoBack={false} title="Settings" />
         <OptionRow
           Icon={IconNight}
           label="Night mode"


gnprice:
Thanks!

Let's make this an app bar with the same design as the app bar elsewhere -- see in particular the other settings screens (e.g. "Notifications") for examples.

ishammahajan:
Fair warning: The second commit (Make titles bold) isn't quite ready for a merge. This is because on certain devices from brands like Oneplus/Oppo which use custom fonts cut off the text at certain places. I'm guessing #2447 is a blocker for this for the moment -- in the case that we do want the titles to be bold.

The first and the third commit is ready though, in case we need to fix this ASAP.

Screenshot_20190620-230342

ishammahajan:
Updated: There were some issues happening where on iOS an `88px` gap would appear on top of the settings card. This was because instead of applying the safe area insets as `44px` (normal), because in the previous version I simply used the `Screen` component to `SettingsCard` which nested two `Screen` components -- hence adding two `ZulipStatusBar`s resulting in two safe area insets padding.

ishammahajan:
To resolve this I removed the use of `Screen` and directly added the `ModalNavBar` which is the component used in `Screen` to display the nav bar.

gnprice:
> This was because instead of applying the safe area insets as `44px` (normal), because in the previous version I simply used the `Screen` component to `SettingsCard` which nested two `Screen` components -- hence adding two `ZulipStatusBar`s resulting in two safe area insets padding.

Hmm interesting. Can we change the existing invocation of `Screen` on that screen to take care of this, then? Currently we only invoke `ModalNavBar` from just one place, namely `Screen`, so it's probably cleanest to maintain that.

(I haven't gone and looked closely at the code in this PR yet.)

gnprice:
Also, one potentially useful bit of context: we used to always keep files/modules split up real small so there was just one React component defined per file.

Pretty recently we reversed that, and have started putting several closely related components in the same file where that makes the code easier to read and understand. But most code you see in the codebase now still reflects the old way.

I think `SettingsCard` + `SettingsScreen` are likely an excellent example of two components that would be easier to read and understand if combined in the same file. :slightly_smiling_face: Feel free to go ahead and combine them, as a commit either early or late in the branch, if that seems to help clarify things.

gnprice:
> Fair warning: The second commit (Make titles bold) isn't quite ready for a merge. This is because on certain devices from brands like Oneplus/Oppo which use custom fonts cut off the text at certain places. I'm guessing #2447 is a blocker for this for the moment -- in the case that we do want the titles to be bold.

(BTW see the comment I [just added](https://github.com/zulip/zulip-mobile/issues/2365#issuecomment-504578340) on #2447 -- we hadn't known this when we first hit that bug, but it looks like it is in fact specifically related to making things bold. So, yeah.)

ishammahajan:
> Hmm interesting. Can we change the existing invocation of `Screen` on that screen to take care of this, then? Currently we only invoke `ModalNavBar` from just one place, namely `Screen`, so it's probably cleanest to maintain that.

The `Screen` widget would require another parameter if we were to implement it that way (perhaps `isCard`?), which is why I was reluctant to change anything. I think in hindsight I agree though, adding that parameter would be better than changing a whole bunch of structure and the cleanness achieved by using `ModalNavBar` at one place.

> I think `SettingsCard` + `SettingsScreen` are likely an excellent example of two components that would be easier to read and understand if combined in the same file. 🙂 Feel free to go ahead and combine them, as a commit either early or late in the branch, if that seems to help clarify things.

Actually, I don't think `SettingsScreen` is actually used anywhere!
```shell
zulip-mobile isham$ rg -C4 SettingsScreen
src/nav/AppNavigator.js
12-import UsersScreen from '../users/UsersScreen';
13-import SubscriptionsScreen from '../subscriptions/SubscriptionsScreen';
14-import ChatScreen from '../chat/ChatScreen';
15-import LoadingScreen from '../start/LoadingScreen';
16:import SettingsScreen from '../settings/SettingsScreen';
17-import LanguageScreen from '../settings/LanguageScreen';
18-import PasswordAuthScreen from '../start/PasswordAuthScreen';
19-import DebugScreen from '../settings/DebugScreen';
20-import DiagnosticsScreen from '../diagnostics/DiagnosticsScreen';
--
51-    realm: { screen: RealmScreen },
52-    search: { screen: SearchMessagesScreen },
53-    subscriptions: { screen: SubscriptionsScreen },
54-    users: { screen: UsersScreen },
55:    settings: { screen: SettingsScreen },
56-    language: { screen: LanguageScreen },
57-    lightbox: { screen: LightboxScreen },
58-    group: { screen: CreateGroupScreen },
59-    'invite-users': { screen: InviteUsersScreen },
zulip-mobile isham$
```
Confirmed this by deleting that file and checking behaviour. Doesn't seem to change!

gnprice:
> Actually, I don't think `SettingsScreen` is actually used anywhere!

Well, your search results there show this line in `AppNavigator.js`:
```
55:    settings: { screen: SettingsScreen },
```
So that's some kind of reference to it! It's not totally obvious what it means until you dig more... but that just means that, until you work out what this is doing, it could be doing anything at all.

In context, this turns out to mean that navigating to the `settings` route opens up `SettingsScreen`. Which, it turns out, is what we do on touching the "settings" icon in the bottom nav bar in the main nav of the app. In other words: that's the component that defines our settings screen. :wink:

gnprice:
One tool that can be quite handy for seeing the tree of React components we're using: the [React Developer Tools](https://facebook.github.io/react-native/docs/debugging.html#react-developer-tools).

If you open that up when on the settings screen, I expect it will show you that there's a `SettingsScreen`, and a `SettingsCard` inside it.

I think it'll also show you that with the current version of this PR, there are now two `Screen` components, one nested inside another. That will be a clue as to why we were double-counting the safe area. :wink: The one in `SettingsScreen` is the one I had in mind above as the existing invocation of `Screen`.

gnprice:
> In context, this turns out to mean that navigating to the `settings` route opens up `SettingsScreen`. Which, it turns out, is what we do on touching the "settings" icon in the bottom nav bar in the main nav of the app. In other words: that's the component that defines our settings screen.

Oh huh! I seem to be mistaken -- and it looks like this is still more confusing than that.

The `settings` route would be reached by the `navigateToSettings` action... but we never actually invoke that action:
```
$ rg -A1 navigateToSettings
src/nav/navActions.js
32:export const navigateToSettings = (): NavigateAction =>
33-  NavigationActions.navigate({ routeName: 'settings' });
```

And unless this makes a super confusing exception to our normal convention, we never navigate to a route except by invoking one of those action creators in `navActions.js`. (The route names are simple words that are a pain to usefully grep for.)

So maybe this really isn't used after all? If that's true, we should cut it out.

gnprice:
Anyway, backing up from the SettingsScreen issue:
* I'd like to reserve the use of `Screen` for just the one outermost component that draws stuff visibly on the screen -- let's not introduce it inside a more-nested component like `SettingsCard`.
  * Especially let's not have one `Screen` nested inside another. If you look with the React Developer Tools, I think you'll find there's a `Screen` somewhere above `SettingsCard` in the tree.
  * The `isCard` flag feels like it makes the meaning of `Screen` pretty complicated; and I think trying to stick it in a more-nested place is the root cause of that complication.

#### 93d2cfe15 screen: Change `canGoBack` to `isRootNav`.

* Hmm interesting. Can you say more about why this feels like a better name for this value?

  I don't totally love the existing name, but it's not clear to me this new one is less confusing.

* In `AccountPickScreen` it looks like it didn't get flipped.

* Also in the return value of `handleBackButtonPress` on `BackNavigationHandler`.


#### e42cd9224 navStyles title: Make titles bold.

* This is the one that triggers the Oneplus/Oppo bug, right?


#### 6fc21d0a1 nav bar: Increase left margin for text if `isRootNav={true}`.

* Hmm. Honestly this line is already pretty puzzling:
```
      isRootNav ? { marginLeft: 16 } : { marginRight: NAVBAR_SIZE },
```
Can you work out why we apparently want a `marginRight` in one case, but not the other? The difference between the two cases seems like it should only be about the *left* side, where there might or might not be a back button.

Perhaps we should add a right margin in both cases? Also is `NAVBAR_SIZE` an appropriate width for it? That's pretty wide.

I guess this change is independent of that, though.

* > The desired behavior (which this commit accomplishes) is having the label stick further to the right, starting from the vertical line drawn from the middle of the icons under it.

  I went and briefly played with the system settings app in Android 9, as a model for this kind of UI.

  The titles there seem to be left-aligned with the labels on the settings, so to the right of the icons. That looks pretty good. I think a line through the middle of the icons is less helpful for alignment.

ishammahajan:
Thanks for the review @gnprice !

> If you open that up when on the settings screen, I expect it will show you that there's a `SettingsScreen`, and a `SettingsCard` inside it.

Hmm, in `MainTabs.js`, the component we use to show the main UI, has some code like this.
```js
export default TabNavigator(
  {
    home: {
      screen: HomeTab,
      navigationOptions: {
        tabBarLabel: 'Home',
        tabBarIcon: (props: TabNavigationOptionsPropsType) => (
          <IconInbox size={24} color={props.tintColor} />
        ),
      },
    },
	[...],
    settings: {
      screen: SettingsCard,
      navigationOptions: {
        tabBarLabel: 'Settings',
        tabBarIcon: (props: TabNavigationOptionsPropsType) => (
          <IconSettings size={24} color={props.tintColor} />
        ),
      },
    },
	[...]
  },
  {
    backBehavior: 'none',
    tabBarComponent: TabBarBottom,
    tabBarPosition: 'bottom',
    ...tabsOptions({
      showLabel: !!Platform.isPad,
      showIcon: true,
    }),
  },
);
```
So, I think in this case we call `SettingsCard` directly. The two `Screens` come from one in `MainScreenWithTabs.js` which renders like this.
```js
render() {
    const { styles: contextStyles } = this.context;

    return (
      <View style={[styles.flexed, contextStyles.backgroundColor]}>
        <ZulipStatusBar /> <--- This has a screen under it
        <OfflineNotice />
        <MainTabs />
      </View>
    );
  }
```
This is why I introduced the `isCard` property to `Screen`.

> * The `isCard` flag feels like it makes the meaning of `Screen` pretty complicated; and I think trying to stick it in a more-nested place is the root cause of that complication.

Do you think we should instead introduce a `ModalNavBar` directly? Although as you said before (and I agree),

> Currently we only invoke `ModalNavBar` from just one place, namely `Screen`, so it's probably cleanest to maintain that.

So in this case I don't know what the best thing to do might be.

> * but it's not clear to me this new one is less confusing

Well, I didn't make it so that this was less confusing per se, just to make it more applicable. For example, all the usages of `canGoBack` (adding of the back nav button) are actually pointing towards the fact that it is a root navigation. It isn't something related to this PR, and in retrospect I don't even think it's a good name change, since it's current reflection does only reflect only one aspect of `Screen`. I think I'll drop this commit.

> * This is the one that triggers the Oneplus/Oppo bug, right?

Yep, this is the one. I haven't found a solution yet (except including a font bundled in the app itself #2447, which I think is a very good thing to do because of other reasons as well -- primarily because every messaging app now-a-days has a unique font, which adds a certain unique flavor to it, although what font to choose might be a huge debate in itself).

> Can you work out why we apparently want a `marginRight` in one case, but not the other?

I'm not exactly sure, but I believe it is because of the hamburger menu on the right for the list of topics in a stream (`ExtraButton` in `ChatNavBar.js`)
Screenshot 2019-07-02 at 5 18 23 PM
This becomes apparent when you look at a stream with a really long stream name.
Screenshot 2019-07-02 at 5 20 34 PM


> * The titles there seem to be left-aligned with the labels on the settings, so to the right of the icons. That looks pretty good. I think a line through the middle of the icons is less helpful for alignment.

True, what I picked was a little out of the random. I'll fix it, thanks for looking into it! :)

ishammahajan:
Hmm. So if I align the title to the left, the screen for picking accounts and the welcome screen get disturbed.
Screenshot 2019-07-02 at 6 16 18 PM
Made use of the `isCard` here as well to fix it (the previous usage was also due to this, since if we set the things to dynamically change based on `canGoBack`, it affects both this one and the welcome screen, because both of them don't need to have a `goBack` button and yet are different in the way they handle the status bars (because one is a card and another is a screen)).

I don't think we can do without `isCard` because it is an important distinction between aforementioned screens and this Card usage.

ishammahajan:
Added more details in the separated commit 🙂 .

gnprice:
A few weeks after the last comments above, I was studying our navigation and I confirmed that `SettingsScreen` isn't used anywhere, and deleted it: 2f0ac45d4 .

So that removes one source of confusion here.

gnprice:
> The two `Screens` come from one in `MainScreenWithTabs.js` which renders like this.
> 
> ```js
> render() {
>     const { styles: contextStyles } = this.context;
> 
>     return (
>       <View style={[styles.flexed, contextStyles.backgroundColor]}>
>         <ZulipStatusBar /> <--- This has a screen under it
> ```

Hmm, does it, though? I don't see one:

```
        <View style={style}>
          <StatusBar
            animated
            showHideTransition="slide"
            hidden={hidden && Platform.OS !== 'android'}
            backgroundColor={Color(statusBarColor).darken(0.1)}
            barStyle={getStatusBarStyle(statusBarColor)}
          />
        </View>
```

Here `StatusBar` is an RN upstream component.

It does try to muck with a safe-area inset, though, so it may explain double-counting there:
```
    const style = { height: hidden ? 0 : safeAreaInsets.top, backgroundColor };
```

But it looks like there may indeed be no existing `Screen` involved on the settings screen! Hmm.

gnprice:
#### 68e42174f screen: Add `isCard` property.

* One essential thing that this commit should do is to actually implement the new prop! I.e. this bit that's currently in the next commit:
```
-        <ZulipStatusBar />
+        {!isCard && <ZulipStatusBar />}
```

Without that, this commit just adds some false advertising :slightly_smiling_face: . It also makes it impossible to understand on its own, because as is it doesn't actually work -- the prop doesn't do what it says it does.

gnprice:
#### 1e81a5746 settings card: Add header - Label and separator.

OK, after looking closer at `Screen` vs. `MainScreenWithTabs` (as described above), I'm convinced that using `Screen` is definitely the wrong approach here. Instead we should invoke `ModalNavBar` directly.

Consider all the stuff that `Screen` does:
```
      <View style={[contextStyles.screen, { paddingBottom: safeAreaInsets.bottom }]}>
        {!isCard && <ZulipStatusBar />}
        {search ? (
          <ModalSearchNavBar autoFocus={autoFocus} searchBarOnChange={searchBarOnChange} />
        ) : (
          <ModalNavBar canGoBack={canGoBack} isCard={isCard} title={title} />
        )}
        <OfflineNotice />
        <KeyboardAvoider
          behavior="padding"
          style={[componentStyles.wrapper, padding && styles.padding]}
        >
          <ScrollView
            contentContainerStyle={[styles.flexed, centerContent && componentStyles.content, style]}
            style={componentStyles.childrenWrapper}
            keyboardShouldPersistTaps={keyboardShouldPersistTaps}
            scrollEnabled={scrollEnabled}
          >
            {children}
          </ScrollView>
        </KeyboardAvoider>
      </View>
```
(as seen at the tip of your branch; very similar in master.)

That's a lot of things! Do we want all of those things? You've added a workaround (the `isCard`) to eliminate one of them, but there are a couple more I'm pretty sure we don't want:
* the `paddingBottom`
* the `OfflineNotice`

and then there's all the `KeyboardAvoider` and `ScrollView` stuff that we don't know any reason to think we want here.

The original motivation was to add an app bar, which we call `ModalNavBar`; so let's just add `ModalNavBar`.

I think that simplifies these changes a fair bit, too.

ishammahajan:
> * One essential thing that this commit should do is to actually implement the new prop!

Ah, thanks for noticing that! Fixed.

> I think that simplifies these changes a fair bit, too.

I agree. The offline notice bit is why I kept using the `Screen`, but since you specified we don't actually need it, removing `Screen`.

ishammahajan:
I have also dropped the left-margin on the ModelNavBar because I feel like it doesn't suit there + there's no unnecessary flag added just for that.

> * The titles there seem to be left-aligned with the labels on the settings, so to the right of the icons.

Maybe it is because of Android 10, or because of the OEM of my device, but if I check now the settings label on the `Settings` app on my device appear to not be aligned (left padding 16px) if they don't have any Back Arrow icon for navigating to the previous screen. That looks good too, IMO.

gnprice:
Thanks @ishammahajan !  I like how very much simpler this gets, too :smile:  Looking now.

gnprice:
> Maybe it is because of Android 10, or because of the OEM of my device, but if I check now the settings label on the `Settings` app on my device appear to not be aligned (left padding 16px) if they don't have any Back Arrow icon for navigating to the previous screen.

Here's the system settings app on stock Android 10, in an emulator:

Screenshot_1568664279

That's a screen with a Back arrow. But I can't actually find a screen there that has an app bar with a title and doesn't have a Back arrow. The main screen of the settings app looks like this:

Screenshot_1568664596

gnprice:
And merged -- thanks again @ishammahajan !

Glad this turned out so very simple in the end. :slightly_smiling_face: 

One bit of commit-message style (or two, I guess):
* Tim's username is @timabbott, with a double "t" at the end
* But actually on GitHub @-mentioning someone in a commit message is kind of a misfeature, unfortunately. GitHub turns that into a notification for them every time a version of the commit is rebased and pushed somewhere, which is usually totally irrelevant for them -- e.g. someone took a long stretch of history and rebased it and pushed that to their own clone. That person is doing something perfectly reasonable because it's off in their own sandbox, and the author (you) had a perfectly reasonable intention to mention the person when making a change in the project, but GitHub turns the combination of those into spam.

  For that reason I usually avoid GitHub @-mentions in commit messages. Good alternatives include:
  * just their name -- can be unambiguous with further context (like the chat link in this case), or referring to a maintainer
  * for complete unambiguity, the Linux-kernel / Git-upstream convention -- see the `Reported-By:` in [this random recent kernel commit](https://github.com/torvalds/linux/commit/72dbcf72156641fde4d8ea401e977341bfd35a05) or similar in [lots of others](https://github.com/torvalds/linux/commits/master), or see 338036e0c where I recently wrote `Suggested-by: Anders Kaseorg <anders@zulipchat.com>`.
  * to actually send a notification, @-mention the person in the PR thread. This doesn't get duplicated over and over like commits can, and so this problem doesn't arise.

gnprice:
> The offline notice bit is why I kept using the `Screen`, but since you specified we don't actually need it, removing `Screen`.

Just to clarify this bit: the reason we don't want it from `Screen` here is that we already have one!

Try it out -- go to the settings screen and enter airplane mode, and you should get the red "No Internet connection" banner at the top.

That's supplied by `MainScreenWithTabs`, as quoted in your helpful analysis [up here](https://github.com/zulip/zulip-mobile/pull/3533#issuecomment-507658426).

So I believe invoking `Screen` in `SettingsCard` would produce a second one.

gnprice:
> * But actually on GitHub @-mentioning someone in a commit message is kind of a misfeature, unfortunately.

(I just went and turned this into a bit of docs: https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#mentioning-people as 9164c2b60 .)

Issue #3106, opened 2018-11-02
After creating a new stream a user seems 'unsubscribed'


borisyankov:
Right after a user creates a new stream, he is subscribed to it. 
But they appear unsubscribed in the UI.
Restarting the app does fix this incorrect state.

PR #3534, opened 2019-06-20
zulipSwitch: Derive `valueControlled` from props change.

jainkuniya:
Before we were only deriving `valueControlled` (which is value of the
switch) on component mount (at the time of initializing state).

But if defaultValue props gets changed, then it's value will not get
reflected. Thus when subscription add event arrives, it's value is
propagated till switch, but we aren't taking defaultValue expect at the
time of mounting.

If we want to go with the convention that defaultValue can only be used
at the first time, then we need to use value prop. Which will make
things complicated, as when switch wants to update it will receive both
defaultValue and value from parent, and it will add some logic to handle
such situations.

Actually it is not default value, rather it is original value which parent
caller wants. Basically original value before user (may) make changes.
So `originalValue` better fits here.

So use `getDerivedStateFromProps` which will derive new value of
`valueControlled` and returns an object to update the state.

> getDerivedStateFromProps is invoked right before calling the render
method, both on the initial mount and on subsequent updates.

Reference: https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

Fixes: #3106

### DIFF
--- a/src/account-info/AwayStatusSwitch.js
+++ b/src/account-info/AwayStatusSwitch.js
@@ -29,7 +29,7 @@ class AwayStatusSwitch extends PureComponent<Props> {
     return (
       <OptionRow
         label="Set yourself to away"
-        defaultValue={awayStatus}
+        value={awayStatus}
         onValueChange={this.handleUpdateAwayStatus}
       />
     );
--- a/src/common/OptionRow.js
+++ b/src/common/OptionRow.js
@@ -36,7 +36,7 @@ export default class OptionRow extends PureComponent<Props> {
         {!!Icon && <Icon size={18} style={this.styles.icon} />}
         <Label text={label} style={styles.flexed} />
         <View style={styles.rightItem}>
-          <ZulipSwitch defaultValue={value} onValueChange={onValueChange} />
+          <ZulipSwitch value={value} onValueChange={onValueChange} />
         </View>
       </View>
     );
--- a/src/common/OptionRow.js
+++ b/src/common/OptionRow.js
@@ -12,7 +12,7 @@ import styles, { ThemeContext } from '../styles';
 type Props = {|
   Icon?: IconType,
   label: string,
-  defaultValue: boolean,
+  value: boolean,
   style?: ViewStyleProp,
   onValueChange: (newValue: boolean) => void,
 |};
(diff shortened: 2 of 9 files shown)


gnprice:
Thanks @jainkuniya for the investigation and that detailed write-up! I agree, this code looks quite buggy and looks like it would cause #3106 exactly.

Looking at the linked doc, it's pretty firm in saying that `getDerivedStateFromProps` is a sign we probably want to be doing something else. I think the something-else will make the code simpler, too, so let's try to get to that.

As I mentioned in our call today, I think the point in that doc that applies to this code is:
> If you want to “reset” some state when a prop changes, consider either making a component fully controlled or fully uncontrolled with a key instead. [and note the original has links to more docs]

And looking at the code, the logic looks more complicated than what it's actually doing. A first observation: in
```
    const switchValue = typeof valueControlled !== 'undefined' ? valueControlled : value;
```
I'm pretty sure the condition is always true. So for a first commit, try eliminating that.

Then: I don't see why this component should be getting itself involved in the question of controlled vs. uncontrolled. The jsdoc says it just "provides consistent styling of the built-in Switch component", and that seems like the useful part of what it does. So, let's try having it pass all props straight through to `Switch`, except for `trackColor` which is the thing it actually wants to change, and let the caller figure out `value` and `onValueChange` on its own.

For examples of passing all props through to an upstream component, see `RawLabel` (the smallest and cleanest), and perhaps also `Label` and `Input`. Note in particular the types and the jsdoc.

jainkuniya:
Updated! 👍 

Now as switch is not persisting value in its state: One difference that can be observed is, if internet if off and user toggles switch then it will change state for sometime and will come back to original state as `value` is not changed. Which I guess is expected as Internet is not available.

gnprice:
Thanks @jainkuniya ! I just pushed two of these three commits, which were very nice clean focused refactors. I then wrote two new commits of a similar kind, based on observations about the code that became clearer after those two commits of yours. See 604c34d85^..d941dec9a.

I think those changes may also help in simplifying the remaining change in this branch.

As we discussed in our call today, please also clarify what the change means for OptionRow. I think the best way to do that is probably to have an earlier commit that changes OptionRow from having a `defaultValue` prop to having `value`.

After today's simplifications (both yours and mine), I also think I have a clearer understanding of what we're really doing here: the point is that *`ZulipSwitch` is always a controlled component*.

(After that's all cleared up, I think the thing I said above about passing all miscellaneous props straight through to the underlying component is probably also right. But that will be much simpler to do after the changes in this PR are complete.)

gnprice:

Issue #3133, opened 2018-11-10
PMs screen doesn't show many conversations


gnprice:
The PMs screen shows the PM conversations we've been in recently, but "recently" may be as few as the last 100 PM messages. For a user that heavily uses PMs, this can leave out conversations even from a day or two ago.

One user reported an example as a typical experience: they're looking for a specific conversation, and the last message in it was 3pm yesterday. It's 3:30pm at the time they look, just over 24hr later. It doesn't show up on the PMs screen, though 6 other conversations do.

When there's plenty of blank space on the screen, this looks wrong and flaky/broken. It also makes it more annoying to get to the desired conversation.

In particular, for a group PM conversation, it can look like there's no way to do that at all. (The existing answer is the "Create group" button, which is highly misleading; see #2694.)

timabbott:
With in part this issue in mind, the Zulip backend as of Zulip 1.9 added a nice database index for PMs that a given user has received; so we can very quickly do queries on that data.  

The really dumb version that could be done with the API today would be to do a `GET /messages` `is:private` narrow query, but that would be a lot of data over the network.  Probably what we should actually do is write a function similar to the "more topics" function that takes advantage of this index to get you e.g. the (PM recipient User ID list, last_message_id) tuples covering e.g. the last 1K PMs you've had, which probably is a <10ms query on the database end, and either include the data in `/register` or a separate endpoint (depending what's better for mobile's architecture); I expect it'd be <1hr work for me to add backend support for this if we had confirmation of what format the mobile app wanted.  And the app should be able to maintain that data set after `/register` as new messages come in, since it's basically just "look up the user_id_list, and bump the last_message_id".

(Until the database index work in 1.9, this would have been prohibitively expensive, since we'd need to scan all messages to find enough PMs, which for low-PM users in an organization with a lot of streams traffic could have easily involved the database needing to look at 50K+ messages of history and thus adding 100sms of latency).  

I don't think this data set would need unreads data, since the client should have that and be able to splice it in directly and it's probably cleaner to have one source of truth there, but it wouldn't be hard to include.

borisyankov:
Yeah, indeed the mobile app does not need the unread data. One of the bug that @rishig encountered was that he knew there were unread PMs (correctly) but couldn't see them because of the 100 messages limit.

It would be great to return the first message's raw content in the response for each PM entry so we can include a 'conversation preview' in the list like most other apps do.

timabbott:
[...]

PR #3535, opened 2019-06-21
Add `recent_private_conversations` key to `/register`.

ishammahajan:
WIP PR. Just for review/general help.

### DIFF
--- a/src/pm-conversations/recentPmConversationsReducer.js
+++ b/src/pm-conversations/recentPmConversationsReducer.js
@@ -1,9 +1,32 @@
 /* @flow strict-local */
+import isEqual from 'lodash.isequal';
+
 import type { Action, RecentPrivateConversationsState } from '../types';
-import { REALM_INIT } from '../actionConstants';
+import { REALM_INIT, EVENT_NEW_MESSAGE } from '../actionConstants';
 import { NULL_ARRAY } from '../nullObjects';
 
 const initialState: RecentPrivateConversationsState = NULL_ARRAY;
+
+const newMessage = (state, action) => {
+  if (action.message.type !== 'private') {
+    return state;
+  }
+
+  const userIds = action.message.display_recipient.map(recipient => recipient.id);
+  const index = state.findIndex(item => isEqual(item.user_ids, userIds));
+  const oldMaxMsgId = index < 0 ? null : state[index].max_message_id;
+  const item = {
+    user_ids: userIds,
+    max_message_id: Math.max(action.message.id, oldMaxMsgId ?? -Infinity),
+  };
+
+  const unsorted =
+    index < 0
+      ? [item, ...state] /* force linebreak */
+      : [item, ...state.slice(0, index), ...state.slice(index + 1)];
+
+  return unsorted.sort((a, b) => -(a.max_message_id - b.max_message_id));
+};
 
 export default (
   state: RecentPrivateConversationsState = initialState,
@@ -12,6 +35,8 @@ export default (
   switch (action.type) {
     case REALM_INIT:
       return action.data.recent_private_conversations || initialState;
+    case EVENT_NEW_MESSAGE:
+      return newMessage(state, action);
     default:
       return state;
   }
--- a/src/pm-conversations/recentPmConversationsReducer.js
+++ b/src/pm-conversations/recentPmConversationsReducer.js
@@ -1,0 +1,18 @@
+/* @flow strict-local */
+import type { Action, RecentPrivateConversationsState } from '../types';
+import { REALM_INIT } from '../actionConstants';
+import { NULL_ARRAY } from '../nullObjects';
+
+const initialState: RecentPrivateConversationsState = NULL_ARRAY;
+
+export default (
+  state: RecentPrivateConversationsState = initialState,
+  action: Action,
+): RecentPrivateConversationsState => {
+  switch (action.type) {
+    case REALM_INIT:
+      return action.data.recent_private_conversations || initialState;
+    default:
+      return state;
+  }
+};
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -337,6 +337,7 @@ export const action = deepFreeze({
       realm_users: [],
       user_id: 4,
       realm_user_groups: [],
+      recent_private_conversations: [],
       streams: [],
       never_subscribed: [],
       subscriptions: [],
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -147,7 +147,7 @@ export const stream: Stream = makeStream({
   description: 'An example stream.',
 });
 
-const displayRecipientFromUser = (user: User): PmRecipientUser => {
+export const displayRecipientFromUser = (user: User): PmRecipientUser => {
   const { email, full_name, user_id: id } = user;
   return deepFreeze({
     email,
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -4,6 +4,7 @@ import type {
   CrossRealmBot,
   RealmEmojiById,
   RealmFilter,
+  RecentPrivateConversation,
   Stream,
   Subscription,
   User,
@@ -105,6 +106,10 @@ export type InitialDataRealmUser = {|
 
 export type InitialDataRealmUserGroups = {|
   realm_user_groups: UserGroup[],
+|};
+
+export type InitialDataRecentPmConversations = {|
+  recent_private_conversations: RecentPrivateConversation[],
 |};
 
 type NeverSubscribedStream = {|
@@ -273,6 +278,7 @@ export type InitialData = {|
   ...InitialDataRealmFilters,
   ...InitialDataRealmUser,
   ...InitialDataRealmUserGroups,
+  ...InitialDataRecentPmConversations,
   ...InitialDataStream,
   ...InitialDataSubscription,
   ...InitialDataUpdateDisplaySettings,
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -536,3 +536,26 @@ export type Message = $ReadOnly<{
   subject: string,
   subject_links: $ReadOnlyArray<string>,
 }>;
+
+//
+//
+//
+// ===================================================================
+// Summaries of messages and conversations.
+//
+//
+
+/**
+ * Describes a recent PM conversation.
+ *
+ * See https://github.com/zulip/zulip/commit/4c3c669b4#diff-2c2aa234cb4a0120fa5f2eeaf8a94fa2R283
+ * for the structure of this type and the meaning of its properties.
+ *
+ * `user_ids` does not contain the `user_id` of the current user. Consequently,
+ * a user's conversation with themselves will return [], which is unlike the
+ * behaviour found in other parts of codebase.
+ */
+export type RecentPrivateConversation = {|
+  max_message_id: number,
+  user_ids: number[],
+|};
--- a/src/boot/reducers.js
+++ b/src/boot/reducers.js
@@ -21,6 +21,7 @@ import nav from '../nav/navReducer';
 import outbox from '../outbox/outboxReducer';
 import presence from '../presence/presenceReducer';
 import realm from '../realm/realmReducer';
+import recentPrivateConversations from '../pm-conversations/recentPmConversationsReducer';
 import session from '../session/sessionReducer';
 import settings from '../settings/settingsReducer';
 import streams from '../streams/streamsReducer';
@@ -50,6 +51,7 @@ const reducers = {
   outbox,
   presence,
   realm,
+  recentPrivateConversations,
   session,
   settings,
   streams,
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -44,9 +44,9 @@ export const storeKeys: Array<$Keys<GlobalState>> = [
  * don't have to re-download it.
  */
 // prettier-ignore
-export const cacheKeys: Array<$Keys<GlobalState>> = [
-  'flags', 'messages', 'mute', 'narrows', 'realm', 'streams',
-  'subscriptions', 'unread', 'userGroups', 'users',
+export const cacheKeys = [
+  'flags', 'messages', 'mute', 'narrows', 'realm', 'recentPrivateConversations',
+  'streams', 'subscriptions', 'unread', 'userGroups', 'users',
 ];
 
 /**
--- a/src/config.js
+++ b/src/config.js
@@ -32,6 +32,7 @@ const config: Config = {
     'realm_filters',
     'realm_user',
     'realm_user_groups',
+    'recent_private_conversations',
     'stream',
     'subscription',
     'update_display_settings',
--- a/src/directSelectors.js
+++ b/src/directSelectors.js
@@ -23,6 +23,7 @@ import type {
   Subscription,
   Stream,
   Outbox,
+  RecentPrivateConversation,
   User,
   UserGroup,
   UserStatusState,
@@ -82,6 +83,9 @@ export const getSubscriptions = (state: GlobalState): Subscription[] => state.su
  * or `getStreamsById` instead.
  */
 export const getStreams = (state: GlobalState): Stream[] => state.streams;
+
+export const getRecentPrivateConversations = (state: GlobalState): RecentPrivateConversation[] =>
+  state.recentPrivateConversations;
 
 export const getPresence = (state: GlobalState): PresenceState => state.presence;
 
--- a/src/pm-conversations/__tests__/pmConversationsSelectors-test.js
+++ b/src/pm-conversations/__tests__/pmConversationsSelectors-test.js
@@ -1,16 +1,13 @@
 import deepFreeze from 'deep-freeze';
 
 import { getRecentConversations } from '../pmConversationsSelectors';
-import { ALL_PRIVATE_NARROW_STR } from '../../utils/narrow';
 
 describe('getRecentConversations', () => {
-  test('when no messages, return no conversations', () => {
+  test('when no recent conversations, return no conversations', () => {
     const state = deepFreeze({
       realm: { email: 'me@example.com' },
+      recentPrivateConversations: [],
       users: [{ user_id: 0, email: 'me@example.com' }],
-      narrows: {
-        [ALL_PRIVATE_NARROW_STR]: [],
-      },
       unread: {
         pms: [],
         huddles: [],
@@ -25,54 +22,17 @@ describe('getRecentConversations', () => {
   test('returns unique list of recipients, includes conversations with self', () => {
     const state = deepFreeze({
       realm: { email: 'me@example.com' },
+      recentPrivateConversations: [
+        { max_message_id: 3, user_ids: [1] },
+        { max_message_id: 2, user_ids: [2] },
+        { max_message_id: 0, user_ids: [1, 2] },
+        { max_message_id: 4, user_ids: [] },
+      ],
       users: [
         { user_id: 0, email: 'me@example.com' },
         { user_id: 1, email: 'john@example.com' },
         { user_id: 2, email: 'mark@example.com' },
       ],
-      narrows: {
-        [ALL_PRIVATE_NARROW_STR]: [0, 1, 2, 3, 4],
-      },
-      messages: {
-        1: {
-          id: 1,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-          ],
-        },
-        2: {
-          id: 2,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 2, email: 'mark@example.com' },
-          ],
-        },
-        3: {
-          id: 3,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-          ],
-        },
-        4: {
-          id: 4,
-          type: 'private',
-          display_recipient: [{ id: 0, email: 'me@example.com' }],
-        },
-        0: {
-          id: 0,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-            { id: 2, email: 'mark@example.com' },
-          ],
-        },
-      },
       unread: {
         pms: [
           {
@@ -132,62 +92,17 @@ describe('getRecentConversations', () => {
   test('returns recipients sorted by last activity', () => {
     const state = deepFreeze({
       realm: { email: 'me@example.com' },
+      recentPrivateConversations: [
+        { max_message_id: 4, user_ids: [1] },
+        { max_message_id: 3, user_ids: [2] },
+        { max_message_id: 5, user_ids: [1, 2] },
+        { max_message_id: 6, user_ids: [] },
+      ],
       users: [
         { user_id: 0, email: 'me@example.com' },
         { user_id: 1, email: 'john@example.com' },
         { user_id: 2, email: 'mark@example.com' },
       ],
-      narrows: {
-        [ALL_PRIVATE_NARROW_STR]: [1, 2, 3, 4, 5, 6],
-      },
-      messages: {
-        2: {
-          id: 2,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-          ],
-        },
-        1: {
-          id: 1,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 2, email: 'mark@example.com' },
-          ],
-        },
-        4: {
-          id: 4,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-          ],
-        },
-        3: {
-          id: 3,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 2, email: 'mark@example.com' },
-          ],
-        },
-        5: {
-          id: 5,
-          type: 'private',
-          display_recipient: [
-            { id: 0, email: 'me@example.com' },
-            { id: 1, email: 'john@example.com' },
-            { id: 2, email: 'mark@example.com' },
-          ],
-        },
-        6: {
-          id: 6,
-          type: 'private',
-          display_recipient: [{ id: 0, email: 'me@example.com' }],
-        },
-      },
       unread: {
         pms: [
           {
--- a/src/pm-conversations/__tests__/pmConversationsSelectors-test.js
+++ b/src/pm-conversations/__tests__/pmConversationsSelectors-test.js
@@ -1,161 +1,155 @@
-import deepFreeze from 'deep-freeze';
+/* @flow strict-local */
 
 import { getRecentConversations } from '../pmConversationsSelectors';
+import * as eg from '../../__tests__/lib/exampleData';
 
 describe('getRecentConversations', () => {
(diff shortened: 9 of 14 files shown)


gnprice:
Thanks @ishammahajan ! This generally looks like the right structure.

A few comments:
* Let's first do a version that just solves #3133 , without yet showing a snippet of the latest message.
  * Once that's done: to add snippets, I think we can get the server to start sending those as part of the `recent_private_conversations` data. That'll save us from going and individually fetching a bunch of messages, which I'm guessing is what the new `getMessage` and the hypothetical `fetchPrivateMessagesv2` is intended for.
* On the `RecentPrivateConversations` type, let's have a bit of jsdoc about what it means:
  * Link to the comment added in zulip/zulip@4c3c669b4 that describes the format.
  * Have a summary line (like the other examples there) that briefly describes it in a sentence.
  * Mention on `user_ids` the detail that it's of *other* users -- that's something that isn't totally consistent across the system.
* I think the next pieces are to add logic to the new reducer to keep it up to date as `EVENT_NEW_MESSAGE` and other relevant events come in (see the events handled in the messages and narrows reducers)... and then to use this new data to fix #3133!

gnprice:
(Also note e137e29b5 -- I think you had a version based on before that and rebased past it, and you'll want to adjust the rebase a bit.)

ishammahajan:
Thanks for the early review!

> * `getMessage` and the hypothetical `fetchPrivateMessagesv2` is intended for.

The `getMessage` was my attempt at fetching only one message from the `/messages` endpoint, not get the raw message, but I can see where the confusion arose from. Likewise, the `fetchPrivateMessagesv2` was just added to test functionality alongside the existing behaviour to confirm correct behaviour.

> * let's have a bit of jsdoc about what it means

I'll add it, thanks!

> * I think the next pieces are to add logic to the new reducer to keep it up to date as `EVENT_NEW_MESSAGE` and other relevant events come in

Hmm. That sounds about right, but my consideration for this was that we don't really have any message data stored in the `messages` which we can refer to though the `max_message_id` while displaying data using a selector -- I think my next move would be to fetch all these messages `max_message_id` refers to, using a loop. What do you think?

gnprice:
> > * `getMessage` and the hypothetical `fetchPrivateMessagesv2` is intended for.
> 
> The `getMessage` was my attempt at fetching only one message from the `/messages` endpoint, not get the raw message, but I can see where the confusion arose from. Likewise, the `fetchPrivateMessagesv2` was just added to test functionality alongside the existing behaviour to confirm correct behaviour.

Ah OK, if it's just for your testing in development then never mind.

What I mean is just that if you were thinking these would be part of the PR to merge, I think we can actually do without them. (We don't need that data for solving #3133 , and for showing snippets we'll get it another way that's more efficient.)


> Hmm. That sounds about right, but my consideration for this was that we don't really have any message data stored in the `messages` which we can refer to though the `max_message_id` while displaying data using a selector -- I think my next move would be to fetch all these messages `max_message_id` refers to, using a loop. What do you think?

Right, so: for #3133 we don't actually need any message data! So it's fine that the message IDs in `recent_private_conversations` may refer to messages we don't have the message data for. The IDs are needed just so we can sort the conversations by recency, and then the only other data we need on a given conversation is the user IDs. (And the user data, but that we maintain the complete set of.)

ishammahajan:
> (Also note [e137e29](https://github.com/zulip/zulip-mobile/commit/e137e29b5d80ea3c534192615d5b2b9dd5f38630) -- I think you had a version based on before that and rebased past it, and you'll want to adjust the rebase a bit.)

Ah, I thought there was something wrong about that! I was confused when the linter gave errors for the exact same lines. Turns out there was an 's' missing for each. 😄

gnprice:
> * I think the next pieces are to add logic to the new reducer to keep it up to date as `EVENT_NEW_MESSAGE` and other relevant events come in (see the events handled in the messages and narrows reducers)... and then to use this new data to fix #3133!

(I used terms a bit loosely here, so to clarify just in case because you haven't already spent a lot of time with this part of the system: `EVENT_NEW_MESSAGE` is a `type` value we use on certain Redux actions we create within the app. So anything that involves `EVENT_NEW_MESSAGE` is about a Redux action, not directly about an event. But, like all our action types with names `EVENT_*`, the actions of this type are created one-to-one to correspond to events we get from the Zulip server. In this case, consulting `eventToAction.js`, you'll see that an `EVENT_NEW_MESSAGE` action corresponds to a `message` event. Because one of these actions corresponds one-to-one to an event, we'll sometimes conflate the two casually, like I did above.)

ishammahajan:
> Right, so: for #3133 we don't actually need any message data! So it's fine that the message IDs in `recent_private_conversations` may refer to messages we don't have the message data for. The IDs are needed just so we can sort the conversations by recency, and then the only other data we need on a given conversation is the user IDs. (And the user data, but that we maintain the complete set of.)

From that issue:
>The PMs screen shows the PM conversations we've been in recently, but "recently" may be as few as the last 100 PM messages. For a user that heavily uses PMs, this can leave out conversations even from a day or two ago.

It makes sense now! Would you confirm whether the code I have written in `messageSelectors.js` is correct? I'll get to adding the extra logic.

gnprice:
> I'll get to adding the extra logic.

Sounds good!


> Would you confirm whether the code I have written in `messageSelectors.js` is correct? 

Well, so the first question might be: what is that function `getPrivateMessages` supposed to mean?

We don't seem to have any jsdoc on it, so this is a bit implicit. But you can search for where it's used:
```
$ rg -C4 getPrivateMessages
...
10-export const getRecentConversations: Selector<PmConversationData[]> = createSelector(
11-  getOwnEmail,
12:  getPrivateMessages,
...
159-export const getUnreadPrivateMessagesCount: Selector<number> = createSelector(
160:  getPrivateMessages,
161-  getReadFlags,
162-  (privateMessages, readFlags) => countUnread(privateMessages.map(msg => msg.id), readFlags),
```

It's used in two places. The first of those is probably used only for the PM conversations screen (I haven't looked to confirm that), and I'm guessing is the path you meant to affect.

The second one, though, seems to have a different job: it's counting how many of those known PMs are unread.

So for that job, it really wants a full list of private *messages*, or their IDs anyway -- not just the latest one message per conversation.

Which does pretty well fit the name `getPrivateMessages`.

So let's leave `getPrivateMessages` as it is, doing basically what the name says it does. I think you'll want to make some changes to `getRecentConversations`, though, to use the new data.

gnprice:
One other thing I notice, looking a bit more at the existing code:

We do use in a few places an unread count for each conversation. We're computing that from looking back through the actual messages we have.

For now, for fixing #3133 , I think we can leave that aspect of the logic as it is: the unread count for a conversation will just count unreads in the messages we happen to have (often, just the latest 100 PMs overall).

As a further improvement, we might see about getting unread counts added to the `recent_private_conversations` data sent by the server. Then we can use those instead. But we're already today showing unread counts based only on those most recent messages; so it's OK to leave that as is while we start at least showing the older conversations at all, fixing #3133.

gnprice:
Oh btw, as you're writing that reducer logic -- here's a remark I made over on #3133 that may be helpful:

> > After fetching, clients will need to correctly maintain these data structures as new messages are sent or deleted (Though for deleted, probably you can just not worry about it too much? Unclear).
> 
> Hmm -- yeah, in fact we can't in general maintain them exactly on deletion. If the message that was latest is deleted, we don't know what the next-latest would be.
> 
> I think it's OK for us to just say that the semantics of our data structure are slightly relaxed: `max_message_id` is actually an upper bound on the message IDs, one that's "close enough" that it's suitable for use in sorting the conversations by recency.

Basically there's a case where you may get stuck trying to see how to write the reducer logic... and I think the right answer there is to [punt](https://english.stackexchange.com/a/388166) on it and just let the data be slightly wrong in a certain way.

gnprice:
> EDIT: Now that I think of it, if the user has had conversations in the WebApp but not on mobile, and they want to have a look at the conversation from the last 101th message, they wouldn't be able to. We should add the relevant code to fetch the messages as well, I think (though perhaps in a later commit).

Right, so if the user goes and actually navigates to the conversation, we'll need to go fetch the messages for it.

This is something we already do, though, and the existing mechanism should handle it just fine. If for example you go to the stream list, you can navigate to a stream where potentially the last message was a very long time ago, so we don't have any of those messages at all -- and we go and fetch them. Even in the PM conversations screen today, we might have only a handful of messages from any given conversation we show, so if the user navigates to it we promptly go fetch more.

See `fetchMessagesInNarrow` (and its freshly-added jsdoc) for one key layer of where that happens. See also its callers -- the most important is `doNarrow`. We invoke `doNarrow` from all the different places where the user chooses to navigate to some narrow; and it's the only place where we call `navigateToChat`, which does the actual navigation.

gnprice:
I just went and wrote some more jsdoc too, expanding on the subject of that last comment. Take a look at 089adebe0 and 2f4482572.

ishammahajan:
> Which does pretty well fit the name `getPrivateMessages`.
> 
> So let's leave `getPrivateMessages` as it is, doing basically what the name says it does. I think you'll want to make some changes to `getRecentConversations`, though, to use the new data.

That makes sense, I think. Thanks!

> As a further improvement, we might see about getting unread counts added to the `recent_private_conversations` data sent by the server. Then we can use those instead.

Hmm. That's a good idea. Do we get something similar to show the unread counts in the streams?

> Basically there's a case where you may get stuck trying to see how to write the reducer logic... and I think the right answer there is to [punt](https://english.stackexchange.com/a/388166) on it and just let the data be slightly wrong in a certain way.

Yeah, I did 'punt' on it since it's difficult to get the last message of a conversation (specially if the messages of that conversation have not been retrieved. 

Actually, the act of a user deleting a message should make it a more recently active conversation than others with no activity (similar for `EVENT_ADD_REACTION` and others). So I wonder if we should artificially increase `max_message_id` to reflect the recency, since that number has only one use -- to sort the conversations by recency.

> Right, so if the user goes and actually navigates to the conversation, we'll need to go fetch the messages for it.

Understood, thanks! 🙂 

> I just went and wrote some more jsdoc too, expanding on the subject of that last comment. Take a look at [089adeb](https://github.com/zulip/zulip-mobile/commit/089adebe063a793d07c778501212f6769ead012e) and [2f44825](https://github.com/zulip/zulip-mobile/commit/2f4482572a2f0150d61dc608df0622223cddb6cb).

Thanks for that, helped a lot!

timabbott:
> Hmm. That's a good idea. Do we get something similar to show the unread counts in the streams?

So, we actually already have the `unread_msgs` data structure that has data on unread messages everywhere; so if we want to display unread messages data in this UI widget, we can just splice that together client-side (I don't think it makes sense to have the server send two copies of the unread PMs data).

ishammahajan:
> So, we actually already have the `unread_msgs` data structure that has data on unread messages everywhere;

> > the unread count for a conversation will just count unreads in the messages we happen to have (often, just the latest 100 PMs overall).

Hmm. Now that I check up on it, we do already use this in the app. Not sure what @gnprice was referring to there.

gnprice:
> Hmm. Now that I check up on it, we do already use this in the app. Not sure what @gnprice was referring to there.

I was looking, I think, at the `getUnreadPrivateMessagesCount` function which I quoted from above. As I described when quoting from it, it does in fact go and count through the messages we have.

But: searching in turn for that function, it looks like we don't actually ever invoke it! And we haven't since bd88566b6, nearly two years ago.

I'll go cut it out so it doesn't cause any further confusion.

gnprice:
Thanks @ishammahajan for this update!

Reading the code:

* In this new data structure in the state, let's have an invariant that the list of conversations is sorted (by reverse `max_message_id`, i.e. the order we want to use them in.) The initial value comes to us that way, and that simplifies using it.

  So let's document that in jsdoc on the `RecentPrivateConversationsState` type, and then adjust the reducer logic to maintain that.

* In a couple of places there are conditions like `if (action.ownEmail &&` and `.filter(conversation => conversation !== undefined)` that look defensive against data that should be disallowed by the types. Are those situations actually possible? If so, we should think through what they actually mean and how we'd want to handle them. If not, let's leave out the extra code.

* Does this bit actually work? I think the `===` will never be true:
```
  const conversationIndex = state.findIndex(
    conversation => conversation.user_ids === newConversation.user_ids,
  );
```

* Speaking of "does this actually work": please write unit tests. :grin: The reducer is an especially good target for unit tests: a well-specified, self-contained interface, and the logic inside can be tricky.

  Take a look at `src/users/__tests__/userSelectors-test.js` and `src/webview/html/__tests__/messageHeaderAsHtml-test.js` for examples of unit tests in a recommended style. A lot of our existing unit tests are in a style that is both more work to write, and gives lower-quality results.

  In particular, let's apply `@flow strict-local` to new tests, just like we do for all our non-test code. The `exampleData` library used in those two examples will help you do that.

* Speaking of existing unit tests in a style that is a lot of work and also not as helpful as it should be: sorry the `pmConversationsSelectors` tests were such a mess to update. Looking at the diff there, though, I notice two changes that look like real changes in the interface and potentially concerning.

  First: a bunch of `timestamp: 0`, which looks like dummy garbage data. If we don't need those, can we just remove them entirely? How about even making that a preparatory commit, before the main change?

  If there's a reason that wouldn't work, then that would suggest something more complicated is going on that'd be good to explain more.

* Second: the change in the convention of whether the current user's ID is in the list. The commit message says we "don't need" that to be there... but why is that, and in particular why would whether we need it change because we changed the source of the data?

  Either convention is fundamentally OK -- but which convention is used is part of the interface between different parts of the system. If this function switches from one convention to the other in what it returns, that's a change to its interface, and callers may break because they were assuming the previous convention.

  One fine solution would be to adjust this change so it doesn't change the results. Another would be to have a separate commit that changes from one convention to the other, adjusting callers as needed... and again adjust this change so it doesn't change the results. (Either before or after the commit that does change the results.)

* I think this sequence of `filter`, `map`, `map` would be a lot clearer as a single loop:
```
+    const recipients = recentPrivateConversations
+      .filter(conversation => conversation !== undefined)
+      .map(
+        conversation =>
+          conversation.user_ids.length !== 0
+            ? conversation
...
```

* Is this `filter` really producing appropriate behavior, in the cases where it does anything? I'm pretty sure that if it does anything the result will always be quite buggy:
```
+          conversation.user_ids
+            .map(id => ({ email: emails[id] }))
+            .filter(recipient => recipient.email !== undefined),
```
What would be appropriate behavior in that case?

(I think correctly implementing appropriate behavior in that case is one of the things that becomes easier if the enclosing filter, map, map sequence is turned into simply a loop -- in addition to just understanding the code as it is.)

* What's this bit of string-processing mean?
```
+        || unreadHuddles[currentUser.user_id.toString().concat(',', recipient.ids)],
```

In particular, does that put the user IDs in the right order? What order does `unreadHuddles` have them in? This is another thing where many conventions are perfectly reasonable, but "which one?" is part of the interface, and different pieces of code need to agree on it.

If logic like this is indeed needed, this line is too crowded a place for it -- this logic is too gnarly, and too interface-sensitive, to happen anonymously inside a property lookup like this. We'd want to make it its own small function with a name, and a line or two of jsdoc saying what it means.

* Commit style: let's put the `EVENT_NEW_MESSAGE` handler before the change to the selector. Before that handler, the data in the new state subtree is correct only just after the `/register` call, and starts becoming wrong as new messages come in; so it wouldn't be good to switch to using it.

ishammahajan:

> * In this new data structure in the state, let's have an invariant that the list of conversations is sorted (by reverse `max_message_id`, i.e. the order we want to use them in.) The initial value comes to us that way, and that simplifies using it.

That makes total sense, I don't know why I didn't think of it that way. Changed the commit to reflect that. :)

> * In a couple of places there are conditions like `if (action.ownEmail &&` and `.filter(conversation => conversation !== undefined)` that look defensive against data that should be disallowed by the types. Are those situations actually possible?

I believe the first (`action.ownEmail`) was a goof by me, and the second was a piece which I did not remove after iterating through the reducer a bit. Sorry about that.

(next responses will be a bit out of order because of the nature of the comments.)

> * First: a bunch of `timestamp: 0`, which looks like dummy garbage data.

I had added that as a placeholder for the data in `PmConversationData` since its type consisted of the timestamp as well. The reason was that I did not want to disturb the functioning of the other callers of the functions which used the same type -- which in hindsight was also foolish since setting timestamp to 0 makes timestamp attribute as useless as just making it optional in the type.

Anyway, now that I went and checked the usage of timestamp, it isn't actually used anywhere -- `PmConversationData`s usage lies in `UnreadCard` and `PmConversationList`, both of which don't use the attribute `timestamp` anywhere, as far as I can understand from the code. So perhaps it will be best to remove them from the type?

...and yes! Removing the attribute from the type only results in a `flow` error in `pmConversationsSelectors` where I was using it to set `timestamp: 0`. After removing that, `flow` passed successfully.

ishammahajan:
> * the change in the convention of whether the current user's ID is in the list. The commit message says we "don't need" that to be there... but why is that, and in particular why would whether we need it change because we changed the source of the data?

I tried to keep the tests faithful to how we would get data from the server. The server would send it with no user ID of the current user. The selector though, converts an empty field into one that contains the current user's ID, while leaving the fields with multiple user ids untouched -- this was the pattern in the previous reducer as well (`normalizeRecipientsSansMe` demonstrated) to prevent the usage of the current user's name in the displayed list.

> * Is this `filter` really producing appropriate behaviour, in the cases where it does anything? I'm pretty sure that if it does anything the result will always be quite buggy:
> 
> ```
> +          conversation.user_ids
> +            .map(id => ({ email: emails[id] }))
> +            .filter(recipient => recipient.email !== undefined),
> ```

The reason I put that is because for some reason, the server was giving me a conversation with a user with user id `3` (I don't remember the exact number correctly), but such a user does not exist in the users tree of our redux state. I initially thought that was `Welcome Bot` but that isn't the case since my conversation with it does get displayed even after adding this line. Perhaps a deleted user?

... this is so strange! Now that I try out the code without the line, for getting what the error exactly was, since I'm very positive that such an error existed, the error disappeared! I think this should be investigated more, but I don't know how I would do such a thing.

ishammahajan:
(not ready for review yet)

ishammahajan:
> * Speaking of "does this actually work": please write unit tests. 😁 The reducer is an especially good target for unit tests: a well-specified, self-contained interface, and the logic inside can be tricky.

Done! Also made the `pmConversationsSelectors.js` conform to those rules.

ishammahajan:
Alright, the rest of it should be fixed too.

The problem I'm having at the moment though, is that the `PmConversationsCard` doesn't update the entire state, and only the extra entry which comes in when you try to update it real time. The exact problem, I think, is that the state gets updated (tested by reloading the app -- the list is OK), but the list doesn't get updated properly from the state for some reason. I cannot figure out why.

Here's what I mean -- https://imgur.com/tm1ffFU

A recording which I had recorded prior to the above had a bug which related to this.

> > * the change in the convention of whether the current user's ID is in the list. The commit message says we "don't need" that to be there... but why is that, and in particular why would whether we need it change because we changed the source of the data?
> 
> I tried to keep the tests faithful to how we would get data from the server.

So it turns out I didn't really understand what you mean in the above paragraph -- you're talking about the selector where my reply was about the reducer. I have fixed the selectors (and tests) to match the convention of `unreadHuddles`.

The problem with `PmConversationsCard` not updating the order of the list properly does still exists though.

ishammahajan:
Oh, just an update: I might know the source of this issue now; attempting to resolve it.

gnprice:
#### 74f820468 Add `recent_private_conversations` key in `register` endpoint.

* Nit / GitHub style tip: in the commit message, instead of
```
    [#11944 on zulip/zulip](https://github.com/zulip/zulip/issues/11944)
```
you can write `zulip/zulip#11944`. Fits better into prose, and linkifies just as well on GitHub: zulip/zulip#11944 .

Also `zulip/zulip@4c3c669b4`, for linking to a commit: zulip/zulip@4c3c669b4 .

* Thanks for the detailed and clear jsdoc on the type! I especially appreciate the highlighting of the potentially-surprising bit of the semantics.

* Nit there: s/parameters/properties/. Parameters are what a function consumes; properties are (the JS term for) what an object has.

  (The term "property" isn't standard across languages: Python says "attribute", C and C++ say "member" or "data member", Java says "field". But "parameter" is quite standard.)

* Another nit: s/does not consist of/does not contain/. "Consist of" would mean it contained that and *only* that. It's true that the list doesn't consist of that one user ID, but that isn't very informative. :wink: 

* One more nit, at a different layer: look around at the context of the code around where you've put this definition. It's right between `ClientPresence` and `UserPresence` -- those are closely related to each other (one just consists of a bunch of the other), so it's helpful to keep them right next to each other and not split them.

  This type doesn't have much to do with those or the others nearby. There's also an attempt at overall thematic organization in that file, with header comments like this one:
```
//
//
//
// ===================================================================
// Users and data about them.
//
//
```

Of the existing headings, I think the most appropriate is "Messages and things attached to them", at the end (indeed the very end, the end of that section after `Message`.)

But actually even more appropriate would be a new section at the end, with a heading like "Summaries of messages and conversations." Let's do that. Just copy-paste the header format, and replace the heading text.

* On `RecentPrivateConversationsState`, let's have jsdoc mention the invariant that it's sorted by `max_message_id` descending. (Which I see you now have logic for in the next commit, thanks!)

* Oh also, that's a better place for the "last 1000" stuff that's in the jsdoc on `RecentPrivateConversation`. The latter type is really just about one conversation.

#### ae3a34d0d recent pms reducer: Add `EVENT_NEW_MESSAGE`.

```
+  if (action.ownEmail === action.message.sender_email) {
+    return state;
+  }
```

* Hmm, do we want this? I think not -- the typical behavior, which seems helpful, is that if I send a message in an old conversation that bumps it to the top of my list.

```
+  const conversationIndex = state.findIndex(conversation =>
+    isEqual(conversation.user_ids, newConversation.user_ids),
+  );
+
+  if (conversationIndex !== -1) {
+    return [
+      newConversation,
```

* So this will maintain the desired invariant... under the assumption that the new message always has a greater message ID than any of the existing conversations.

  I think that assumption *might* be true, but I'm not sure of it. Even if it should be true in principle, it's easy to imagine a bug causing it not to be.

  So I think I'd feel a bit more comfortable if we sorted this list explicitly before returning it. At worst that shouldn't be expensive because this list won't be very long. And if the sort implementation is optimized in some common ways (as e.g. [Python's is](https://en.wikipedia.org/wiki/Timsort)), then when working on an already-sorted or already-almost-sorted list, it'll only take O(n) time -- which we already incurred with the `findIndex`.

  (There are fancy ways to set up data structures for this, avoiding the `findIndex` as well as the sort, so that it's log(N) time at worst and I think even O(1) in the (common) case where the entry being updated is one of the top O(1) of them already. But I don't think any of that is needed here.)

* Oh, similarly, on avoiding that assumption: instead of just overwriting the `max_message_id` with `action.message.id`, it should be the max of the old value and the new ID.

```
+    user_ids: action.message.display_recipient.map(recipient => recipient.id),
```

* Hmm, is this right? In particular, doesn't `display_recipient` contain the self user's ID?


#### 7199132be Remove `timestamp` attribute from `PmConversationData`.

Looks good, thanks! Will merge.


#### 4043b10b6 pms selectors: Change selectors for `getRecentConversations`.

```
+export const getAllUserEmails: Selector<$ReadOnly<{ [user_id: number]: string }>>
 = createSelector(
+  getAllUsers,
+  (users = []) =>
+    users.reduce((array, user) => {
+      array[user.user_id] = user.email;
+      return array;
+    }, {}),
+);
```

* So this matches a style we used to have lots of in the codebase, but I finally went and got rid of a couple of weeks ago after it'd been bugging me for a long time. :slightly_smiling_face: For a direct translation, look at `objectFromEntries` and the places we use it. (See `git log --stat -p --grep reduce e768ebb5f` for the history of removing the old pattern, which I call "false `reduce`".)

  Better yet, let's use a `Map` rather than an object-as-map. At current master, `src/users/userSelectors.js` has several good examples.

* Also the name should end in `ById`.

(OK, stopping here though I haven't yet read the rest of this commit. I think the above should give you a good queue of revisions to make.)

gnprice:
> #### [7199132](https://github.com/zulip/zulip-mobile/commit/7199132bef5ac54f887e2911641c0021799b688f) Remove `timestamp` attribute from `PmConversationData`.
> Looks good, thanks! Will merge.

Done: 53b56f336

gnprice:
(I've just rebased this branch onto master, and also fixed a couple of tests at commit 3/4.)

gnprice:
OK, and pushed a version with a number of revisions:
* Addressed all but one of my comments above.
* Made a few small other changes in the first commit, mainly keeping things alphabetized where we have them that way.
* Made some bigger changes in the commit "pms selectors: Change selectors for getRecentConversations.", mostly as added commits (wouldn't want to merge it that way, but may keep things clearer for now.)

The remaining major thing I think it's essential to do before merging this is:
I'm not certain this correctly handles the various forms in which we identify a PM conversation. (Include self, or not; user IDs, emails; arrays, comma-joined strings.) This is an unfortunately unnecessarily-complex area of the Zulip API.

In particular there's my one remaining comment from above:

> ```
> +    user_ids: action.message.display_recipient.map(recipient => recipient.id),
> ```
> 
> 
> 
> * Hmm, is this right? In particular, doesn't `display_recipient` contain the self user's ID?

The newly rewritten `getRecentConversations` selector has some logic to deal with this patchwork too, and I'm not yet confident that that's all correct.

gnprice:
Just pushed 35be10b90, which confirms that `display_recipient` always contains the self user's ID.

Also confirmed that, as jsdoc in this branch mentions, the user-ID lists in `recent_private_conversations` never contain the self user's ID. (And that they're always sorted:
```python
    # Sort to prevent test flakes and client bugs.
    for rec in recipient_map.values():
        rec['user_ids'].sort()
```
though that isn't a documented guarantee.)

Then the third bit of API surface area involved here is the unreads data structure, which `getRecentConversations` queries in order to add the information of how many in each conversation are unread. Still to do is to pin that down, as part of #4035.

ray-kraesig:
Picking this up, as #3133 is on my plate now:

> Also confirmed that, as jsdoc in this branch mentions, the user-ID lists in `recent_private_conversations` [are] always sorted:
> 
> ```python
>     # Sort to prevent test flakes and client bugs.
>     for rec in recipient_map.values():
>         rec['user_ids'].sort()
> ```

Alas, this is not *quite* so. The above code was only added in 2.2-dev-53-g405a529340 (zulip/zulip@405a529340), which was backported to the 2.1 branch as 2.1.1-50-gd452ad31e0 (zulip/zulip@d452ad31e0).

Zulip server versions 2.1.0 and 2.1.1 will therefore not sort user IDs.

src/pm-conversations/recentPmConversationsReducer.js
```diff
@@ -1,17 +1,42 @@
...
+  if (action.message.type !== 'private') {
+    return state;
+  }
+
+  const userIds = action.message.display_recipient.map(recipient => recipient.id);
+  const index = state.findIndex(item => isEqual(item.user_ids, userIds));
+  const oldMaxMsgId = index < 0 ? null : state[index].max_message_id;
+  const item = {
+    user_ids: userIds,
+    max_message_id: Math.max(action.message.id, oldMaxMsgId ?? -Infinity),
```
chrisbobbe, line 20: Hmm, have we thought through storing `-Infinity` in the Redux state; in particular, what happens when it's saved and then loaded from storage? It looks like it doesn't round-trip through JSON:

```
JSON.parse(JSON.stringify({ a: -Infinity }))
// {a: null}
```

chrisbobbe (commented):
I've just made a rebased version of this PR, but I haven't pushed it to this PR branch in case it might accidentally remove something important to the direction of the PR. You can see it at my `recentprivateconvo-rebased` branch.

chrisbobbe:
> I've just made a rebased version of this PR, but I haven't pushed it to this PR branch in case it might accidentally remove something important to the direction of the PR. You can see it at my `recentprivateconvo-rebased` branch.

Ah, and I've just come across #4104, which I'd forgotten about—it's a more recent PR that may indeed replace this one, so it may not have been super helpful for me to rebase this one (which I did in my `recentprivateconvo-rebased` branch, as mentioned). Ah, well.

gnprice:
I've just sent #4392 as a new approach to this issue.

PR #3536, opened 2019-06-22
dependency: Upgrade react-native-webview to ^6.0.0.

### LINKED ISSUE #3518: Message content shows through in compose area
Reported by @zbenjamin, who's been seeing this on 25.0.114 and for a while before that (presumably 24.0.113, which was the latest beta for some time.)

That's on an iPhone 7. I was also able to reproduce it, on an iPhone 7 in 25.2.116:

It doesn't always happen. Zev suggested

> I think it happens when you open the compose box while messages are still loading

I found that hard to pull off at first, and as I played around with narrowing to various streams and opening the compose box I didn't see it for a while. I first reproed it after following a link in a message, navigating back from the resulting browser view, and quickly opening the compose box again.

Then once it had started happening, it seemed to persist indefinitely (at least in a given narrow): even though it was a conversation where we had all the messages already, if I opened the compose box, I got this symptom every time.

Another time I saw it: I'd followed a link in a message; navigated back from the resulting browser view; the message list was showing the loading animation; I opened the compose box; and the symptom appeared.

jainkuniya:
There has been many improvement since inital release 2.0.0.
This also fixes #3518.

I used git bisect and found that issue #3518 got introduced in
32b6bc490b5fab5065f3a0e0bb457ff1a6fbb109 where we shifted from RN core
webview to react-native-webview first release.

From then there has been many fixes, to take them in let's
use latest release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "react-native-sound": "^0.10.9",
     "react-native-text-input-reset": "^1.0.2",
     "react-native-vector-icons": "^4.6.0",
-    "react-native-webview": "2.0.0",
+    "react-native-webview": "^6.0.0",
     "react-navigation": "^1.5.12",
     "react-navigation-redux-helpers": "^1.1.1",
     "react-redux": "^5.0.7",
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -44,7 +44,7 @@ var escapeHtml = function escapeHtml(text) {
 };
 
 var sendMessage = function sendMessage(msg) {
-  window.postMessage(JSON.stringify(msg), '*');
+  window.ReactNativeWebView.postMessage(JSON.stringify(msg), '*');
 };
 
 window.onerror = function (message, source, line, column, error) {
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -92,7 +92,7 @@ const escapeHtml = (text: string): string => {
 };
 
 const sendMessage = (msg: MessageListEvent) => {
-  window.postMessage(JSON.stringify(msg), '*');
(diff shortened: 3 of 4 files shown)


gnprice:
Hmm fascinating!
* Have you confirmed that this fixes #3518?
* The release notes say some changes to the Android build are required. Have you tried this out on Android?

jainkuniya:
Update on this:
Tested 6.x.x on Android, it is a breaking change. Because of `androidx`. I am more inclined to first fix #3518 and then move to 6.x.x
Because 5.x.x is also a breaking change, they have re-written communication between RN and webview code https://github.com/react-native-community/react-native-webview/releases/tag/v5.0.0
Thus it becomes hard to debug if something goes wrong by taking multiple breaking changes in one shot.

So let's  first test and verify 5.x.x before taking another big/breaking change.

But the problem is I tested 5.9.1 on Android, it works fine. Just we need to take care of 
```
 window.ReactNativeWebView.postMessage(data);
```
But on iOS 
```
document.addEventListener('message'
```
doesn't work. It need `window.addEventListener('message'`, which doesn't work on Android. :(
https://github.com/react-native-community/react-native-webview/issues/363#issuecomment-467730963

BTW #3518 is even fixed in 5.9.1

So just wanted to know your thoughts, should we go with 6.x.x in one shot or should I find a version <= 5.9.1 where all seems to work.

gnprice:
Cool -- thanks for the investigation and that writeup!

AndroidX is supported (indeed required) in RN v0.60, which just came out last week. So let's upgrade to 6.x after we upgrade to RN v0.60 -- which I've just filed as #3548 .

I think it would be productive to do the `react-native-webview` upgrade in smaller steps. Can you find a 4.x version where the upgrade works smoothly?

... In fact, let's move this to an issue. That will be a better way to track this, because we're not really talking about the particular code in this PR.

Issue #3517, opened 2019-06-06
Keypad emoji fail to render (sometimes)


gnprice:
We fixed #3129 a while ago, and the keypad emoji work great in a simulated Nexus 5 running Android 9 and a debug build from a recent master:

![image](https://user-images.githubusercontent.com/28173/59069488-1b74dd80-886c-11e9-8197-886edab1c19b.png)

But on my physical Pixel 2 XL, running also Android 9, with the recent release build v25.2.116, I see the symptoms of #3129 almost exactly: on the same messages from the screenshot above,
* the :one: and :hash: emoji in the messages are invisible (presumably the plain digits in white)
* the reaction with one :one: emoji renders as "1? 1", except I've written "?" here to stand for with an X-ed out rectangle (presumably a replacement glyph).

Perhaps a fonts issue? Note that as our unit tests show:
```
describe('nameToEmojiMap', () => {
  const check = (name, string1, string2) => {
    expect(string1).toEqual(string2);
    expect(nameToEmojiMap[name]).toEqual(string1);
  }; // ...

  test('works for some multi-codepoint emoji', () => {
    check('0',        '0⃣', '0\u{20e3}');
    check('asterisk', '*⃣', '*\u{20e3}');
    check('hash',     '#⃣', '#\u{20e3}');
  });
});
```
a keypad emoji like :hash: expands to the plain `#` character followed by U+20e3... and the latter is
```
$ unicode u+20e3
U+20E3 COMBINING ENCLOSING KEYCAP
```

so basically the font used on my phone might lack that combining character.

gnprice:
Also works great on an iPhone 7 with iOS 12.3.1, running v25.4.118.

gnprice:
Quick additional bit of data: the webapp in Chrome works great on the same messages and same device where the mobile app has the above symptom. That's my physical Pixel 2 XL running Android 9, with v25.6.120, the latest beta.

gnprice:
@jainkuniya reported some additional information on #3537, a PR:
* if I understand correctly, he saw the issue also on a OnePlus 6, on Android 9.
* the problem is fixed by using a codepoint sequence like `0030-fe0f-20e3` rather than `0030-20e3`.

[...]

PR #3537, opened 2019-06-23
webview, android: Fix Keypad emoji fail to render.

jainkuniya:
There are 11 emojis (*, #, 0-9) which contain COMBINING ENCLOSING KEYCAP
(U+20E3) and codePoint of such emoji contains '-' in it. We parse
codePoint by splitting '-', then getting string from it and finally
joining all strings. But in this processing we were missing out '-', which was
causing invalid char render on phones like Pixel XL, One Plus Six
Android 9. Now joining with corresponding unicode value of '-' i.e '\uFE0F',
renders whole emoji correctly.

https://www.fileformat.info/info/unicode/char/20e3/index.htm

Fixes: #3517

image

### DIFF
--- a/src/emoji/data.js
+++ b/src/emoji/data.js
@@ -10,7 +10,7 @@ const parseUnicodeEmojiCode = (code: string): string /* force line */ =>
   code
     .split('-')
     .map(hex => String.fromCodePoint(parseInt(hex, 16)))
-    .join('');
+    .join('\uFE0F');
 
 export const nameToEmojiMap = unicodeEmojiNames.reduce((obj, name) => {
   obj[name] = parseUnicodeEmojiCode(unicodeCodeByName[name]);


gnprice:
Thanks @jainkuniya !

Can you explain how you determined that `'\uFE0F'` should go here? The explanation that it corresponds to '-' doesn't mean much to me.

gnprice:
Here's a relevant reference:
http://www.unicode.org/reports/tr51/#def_emoji_keycap_sequence

I think the upshot is that instead of
```
'0': '0030-20e3',
```
etc., we should have
```
'0': '0030-fe0f-20e3',
```
etc.

gnprice:
Thanks again for this investigation!

The best place to track what we've learned about the issue is the issue thread, so I've copied the key information back there (#3517).

The change in this PR isn't the right direction for the code, so I'm closing the PR; let's have further discussion in the issue thread (and the chat thread linked from there.)

Issue #3531, opened 2019-06-19
Another colorblind accessibility issue


jmandel:
Similar to #3524 I can't read the highlighted text in:
![Screenshot_20190618-225014](https://user-images.githubusercontent.com/313089/59735557-ab773780-921b-11e9-9fb7-562c8226f661.png)

Incidentally, I'm not sure why this particular word is highlighted or what the highlighting indicates.

ishammahajan:
@jmandel thanks for reporting this! Is it perhaps that you're using the 'Alert Words' feature? For reference for the other devs, here's what it looks like in the WebApp.

<img width="381" alt="Screenshot 2019-06-19 at 1 26 50 PM" src="https://user-images.githubusercontent.com/33200024/59747165-f46bd200-9295-11e9-8893-08a74b06d9f2.png">

As it goes, I think the highlight on the mobile app is very harsh, and should be toned down.

jmandel:
Yes, good call. It's an alert word - and indeed, I couldn't read it :)

gnprice:
Thanks @jmandel for the report!

@ishammahajan , if you'd like to claim this and take it the rest of the way, I think it's another one you're well equipped to take care of :slightly_smiling_face: 

The highlighting in the webapp (in Isham's screenshot above, in night mode) doesn't look great to me either. @jmandel in the same message you have above, can you read the word well in the webapp? If so then we should certainly copy the webapp here as an immediate accessibility fix, but either way it's possible it'd be good to then adjust the design on both platforms, at least in night mode.

gnprice:
I just edited the title to reflect what we figured out above about what this feature is :slightly_smiling_face: 

@jmandel , if you'd be up for checking how this works for you in day/light mode, that'd be helpful too.

jmandel:
Thanks! In light mode, it looks fine for me.

jmandel:
Also, to be complete: In dark mode the mobile app and the web app are both unreadable for me.

gnprice:
> Also, to be complete: In dark mode the mobile app and the web app are both unreadable for me.

Ah, OK! I'll go turn this into an issue on zulip/zulip, then, for a bit of a wider audience -- and whatever we end up with, we can implement on both web and mobile.

ishammahajan:
> @ishammahajan , if you'd like to claim this and take it the rest of the way, I think it's another one you're well equipped to take care of 🙂

Sure!

[...]

PR #3538, opened 2019-06-24
css: Make alert-words more readable.

ishammahajan:
The previous iteration of alert words was not ready accessable for
people with color blindness. As @jmandel reports, he wasn't able to
see the white text on the bright green background while in dark
mode.

This commit is a temporary fix for the issue, by reducing the alpha
of the background color in order to make the white text on it more
readable (in dark mode).

References #3531.

(does not fix the issue, just provides a temporary patch for it).


gnprice:
Thanks @ishammahajan ! Just merged :slightly_smiling_face:

PR #3543, opened 2019-07-01
Improve the Realtime docs very slightly.

ishammahajan:
Details are mentioned in the commit messages :).

(this doc is written excellently otherwise, just these few points were a little hard to understand; corrected it so that things are easier to understand for those who read this doc after me)

### DIFF
--- a/docs/architecture/realtime.md
+++ b/docs/architecture/realtime.md
@@ -112,7 +112,8 @@ things:
 * A snapshot (the "initial state") of the relevant application-level
   state.
 * An event finger -- the ID of a newly-created event queue, and a last
-  event ID.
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @ishammahajan for these changes! Merged, with an additional commit on top that hopefully clarifies a bit further.

The one bit of added information in the new commit: the 10-minute timeout is the default, but can be configured. From `rg EVENT_QUEUE_TIMEOUT` in zulip/zulip (and then `rg lifespan_secs` which that leads to), I believe it's controlled by a `queue_lifespan_secs` option to `/register`... though we don't seem to [have documented](https://zulipchat.com/api/register-queue) that one.

Issue #3544, opened 2019-07-07
webview: Getting lots of exception because of selector being empty  


jainkuniya:
![image](https://user-images.githubusercontent.com/18511177/60766673-b7437480-a0ca-11e9-9707-a07c96fa8769.png)

![image](https://user-images.githubusercontent.com/18511177/60766799-8c5a2000-a0cc-11e9-8281-3057554bf8df.png)

I can reproduce it while narrowing to PM from notification. This may also be responsible for #3117. 

In `js.js` we have 
```
 const selector = uevent.messageIds.map(id => `[data-msg-id="${id}"]`).join(',');
  const messageElements = arrayFromNodes(document.querySelectorAll(selector));
```
if `messageIds` is empty then `selector` will be "" which is throwing error. 

**Solution:** Check in webView for `messageIds` is not empty.

But what is point of sending empty array to webView, better fix it on RN side.

Currently we are simply using `===` to check equality (in webViewHandleUpdates.js)
```javascript
  if (
    prevProps.backgroundData.flags
    && prevProps.backgroundData.flags.read !== nextProps.backgroundData.flags.read
  ) {
    uevents.push( updateRead(prevProps, nextProps));
  }
```
Which can not be used in js, as they are objects.
![image](https://user-images.githubusercontent.com/18511177/60766938-0048f800-a0ce-11e9-9484-899b7fbdc123.png)

**Solution:** Use `lodash.isequal`
```javascript
if (
    prevProps.backgroundData.flags
    && !isEqual(prevProps.backgroundData.flags.read, nextProps.backgroundData.flags.read)
  )
```

gnprice:
Thanks for the detailed report! This will be good to fix.

A clarification: the long messy red banner at the top in your screenshot is what we show in a debug build. In a release build (so for normal users), it just says "Oops! Something went wrong. Try again?"

Can you say more about this next bit?

> This may also be responsible for #3117.

I don't see why it would. Was your idea that the exception might stop us processing the update to add the new message? That would just as likely apply to new messages from elsewhere as newly-sent messages... and in the case of a newly-sent message, it would cause the outbox version (with the spinner) to stay visible, never cause the message to disappear... and in any case, `getUpdateEvents` puts the content update first in the list. So by the time we process the read-status update and hit this exception, the content update has already happened.

gnprice:
BTW this issue is already in the tracker, as #3539. I'm going to close this as a duplicate, and point to it from there.

Issue #3539, opened 2019-06-24
Heavy message traffic leads to "Oh no! Something went wrong. Try Again?"


swallitsch:
When a user is looking at a stream/topic and a lot of messages are being posted to that stream or topic, the red "Oh no! Something went wrong! Try again?" banner will briefly appear, despite all the messages updated and no connection issues occurring.

This seems related to the "New Messages" banner, as users don't seem to be able to self activate the error- it only occurs when someone else is sending messages.

Tested by typing gibberish and hitting return a bunch.

![image](https://user-images.githubusercontent.com/32879218/60040958-01f6d280-966f-11e9-8ebc-b16bf6760175.png)

Observed on:
* Google Pixel 3 XL - Android 9 - Build number: PQ3A.190605.003
* iPhone Xs, OS version 12.3.1

Latest Zulip app as of this message, Zulip server is 2.0.3

This issue was only seen on mobile- desktop users do not see the banner or have any connection issues.

gnprice:
Thanks @swallitsch for the report!

The next step for someone to debug this is to reproduce it, and look at the app's error log, e.g. with `adb logcat`. That will have more detail on the error.

If reproducing in a debug build of the app, the banner will also be longer and have more detail on the error.

We also log these errors to Sentry. Looking, [here's one](https://sentry.io/organizations/zulip/issues/812537889/?project=191284) that looks like a match -- the relevant bit of JSON blob is
```
{"message":"Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty.",
"line":436,"column":49,
"error":{
"stack":"Error: Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty.\n    at Object.handleUpdateEventMessagesRead [as read] ...",
```

That should be a solid lead for debugging.

gnprice:
Some detailed debugging by @jainkuniya in the duplicate issue #3544 .

PR #3545, opened 2019-07-07
Fix: #3544. Fix empty selector is being provided to querySelectorAll.

jainkuniya:
* messageList, getUpdateEvents: Use `lodash.isequal` to compare object.
* web, msg read event: Make sure querySelectorAll never get empty selector

For more debug information, please refer #3544

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -415,6 +415,10 @@ var handleUpdateEventReady = function handleUpdateEventReady(uevent) {
 };
 
 var handleUpdateEventMessagesRead = function handleUpdateEventMessagesRead(uevent) {
+  if (uevent.messageIds.length === 0) {
+    return;
+  }
+
   var selector = uevent.messageIds.map(function (id) {
     return "[data-msg-id=\\"" + id + "\\"]";
   }).join(',');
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks @jainkuniya !

Quick remark, having not yet looked at the code: please update the issue references to point at the original issue #3539, rather than #3544 .

gnprice:
We chatted about this briefly in our video call today. Other than the nit above about issue references, just one small comment:
* The condition `!uevent.messageIds` is always false, so let's take that out. The type is `Array<number>`, and an `Array` is always truthy.
  * In general, for everything internal to the app, I want to rely on the types we've expressed in Flow, and not insert side conditions at runtime. If this value were `null` or `undefined` for some reason, that'd be a bug inside the app -- and a hole somewhere in our type-checking -- and we should just fix that bug.
  * (In some cases it could potentially be appropriate to have a runtime check of a condition that would represent a bug in the app. But if we did, it should be an error or at least get logged to Sentry. And in this case I don't see a particular reason to worry that there'd be a type error here that this would catch.)

After that, looks ready to merge!

jainkuniya:
Updated 👍

gnprice:
Thanks @jainkuniya ! Merged.

On reading again, I decided the optimization that switched to `!isEqual` from `!==` wasn't actually necessarily a good optimization. So I added a new commit on top to change it: ce4ed6ab2 .

PR #3559, opened 2019-07-17
[WIP]fix type error hidden by connect

keshav234156:
issue #3451
Had done changes in 2 file from connectFlowfixme to connect and fixed the error given by yarn flow

### DIFF
--- a/src/directSelectors.js
+++ b/src/directSelectors.js
@@ -57,6 +57,11 @@ export const getUserGroups = (state: GlobalState): UserGroup[] => state.userGrou
 
 export const getUserStatus = (state: GlobalState): UserStatusState => state.userStatus;
 
+/**
+ * WARNING: despite the name, only (a) `is_active` users (b) excluding cross-realm bots.
+ *
+ * See `getAllUsers`.
+ */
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks @keshav234156, and welcome !

The change to `ActivityText` looks good, modulo some style things you can fix:

* There are lint errors. See [our testing doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/testing.md#linting) for more.

  I recommend [setting up VS Code](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/editor.md) so that the code you edit gets formatted automatically whenever you save.

* See our guide on [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html), especially the section on commit messages.

  This change would also be helpful to split up into two commits -- one for ActivityText and one for TopicListScreen, since those are basically independent to explain and understand.

Then in the other changes:

* In `TopicListScreen` you change the actual runtime behavior of the code. Is that change correct? Why? Does the code work correctly after the change, and did it work correctly already?

  (I think the answer is that this change breaks this code.)

Issue #3509, opened 2019-06-04
"Unread messages" banner stays when scrolled to end


gnprice:
Pretty often, when I'm catching up on a conversation and the "N unread messages" banner is showing, I'll scroll through it to the end... and then the banner is still there. Usually with just a handful of messages unread.

Then if I scroll slightly back up from the end, the banner goes away.

Now that #3125 is merged, this should get substantially easier to investigate.

GonzRon:
I think I will file a ticket for this if it's not already reported:
I am using the android mobile client, friend is using IOS, we are both experiencing the same behavior.
This message ("unread messages") appears for a topic / stream even when it is you that has entered the message, the mark all messages as read button that appears seems to do nothing.

gnprice:
Hi @GonzRon ! Please do file issues for those -- those reports will be helpful. Let's keep this issue thread focused on the specific issue described in the description at the top.

gnprice:
I just reproduced this on v25.6.120, the latest beta:
* Went to `#test here` on czo, a stream I subscribe to, have muted, and always have tons of unreads in.
* Went to the list of topics.
* Picked a topic that had a larger number of unreads -- 15 unreads -- and navigated there.
* That message list opened. The banner read "6 unread messages". I tried moving around a bit, and I was at the end -- it wasn't possible to scroll down.

So, probably the unreads were older than the messages on the screen? That seems like a clue.

jainkuniya:
There are two possibilities for this issue:
* Unread count in the banner comes from the unread state. The only way it is updated is via server events. So if user scrolls, we send messageFlags to server and server again sends `messageFlags` event and we update it in the store. So if there is internet connectivity issue (slow) then there is high possibility that there will be no effect on unread count in banner even after scrolling multiple times.
I have reproduced it when my mobile data is slow. 
**Solution:**  Something in the direction of #1422. So on message scroll, update unread count state locally as well (and send messageFlag update to server). So that we don't wait until messages are really marked as read on server. And our api calls will update messageFlags in background (current behaviour).

* Another issue is in our `queueMarkAsRead.js`
```javascript
if (Date.now() - lastSentTime > 2000) {
    console.log('Date.now() - lastSentTime > 2000', unsentMessageIds);
    messagesFlags(auth, unsentMessageIds, 'add', 'read');
    unsentMessageIds = [];
    lastSentTime = Date.now();
  }
``` 
[...]

PR #3560, opened 2019-07-18
queueMarkAsRead: Set timeout if there is mark request in less than 2s.

jainkuniya:
If there are multiple requests to mark messages as read with interval
less than 2s, and if there is no further request then later were
stuck in the queue. Becuase there was only one caller to
`messagesFlags`, which was only called by user scroll event.
So even after reaching at the end of message list, unread banner is
visible with some unread count.

So now set timeout to send read message flag to server.

Fixes: #3509

### DIFF
--- a/src/__tests__/exampleData.js
+++ b/src/__tests__/exampleData.js
@@ -4,9 +4,10 @@ import { createStore } from 'redux';
 
 import type { CrossRealmBot, Message, PmRecipientUser, Stream, User } from '../api/modelTypes';
 import type { GlobalState, RealmState } from '../reduxTypes';
-import type { Account } from '../types';
+import type { Auth, Account } from '../types';
 import { ACCOUNT_SWITCH, LOGIN_SUCCESS } from '../actionConstants';
 import rootReducer from '../boot/reducers';
+import { authOfAccount } from '../account/accountMisc';
 
 // TODO either fix Jest test-discovery patterns, or rename this file,
 // so this dummy test isn't required.
@@ -66,6 +67,7 @@ const makeAccount = (user: User): Account => ({
 
 export const selfUser: User = makeUser({ name: 'self' });
 export const selfAccount: Account = makeAccount(selfUser);
+export const selfAuth: Auth = authOfAccount(selfAccount);
 
 export const otherUser: User = makeUser({ name: 'other' });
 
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks @jainkuniya !

Looking carefully at this logic for a few minutes, I went back and forth about whether I thought it worked correctly. I now *think* it does...

but that's a good reminder to write a test for it! Take a look at this Jest documentation for some very handy features to help write clean tests for `setTimeout`-using code:
https://jestjs.io/docs/en/timer-mocks

gnprice:
Oh, in fact also:

In that webapp `send_read` function I mentioned in the issue, the line that handles this is:
```
    start = _.throttle(server_request, 1000);
```
I suggested writing timing logic... but I have a better idea now. :wink: Somehow I didn't make the connection that we have just as good a `throttle` function handy too -- we'd call it `lodash.throttle`. So let's use that.

I think a couple of unit tests would still be valuable, partly to help make sure we have the behavior clear and partly as a warmup to use the same time-mocking feature in some other tests we should have.

gnprice:
Thanks @jainkuniya !

I'd like all new files to have `@flow strict-local`, including tests. Let's do that here.

Adding it as a quick experiment, there are two things to do:

* The auth objects need to be objects of the right type.

  It's a small amount of work to make an appropriate object... but fortunately it's work we only really need to do once for the whole codebase. See the files listed by `git grep -l exampleData` (and then see `exampleData.js`) for where I've done this for a variety of objects.

  Here, let's add to `exampleData` an object `selfAuth`, right after `selfAccount`, that is just `authOfAccount(selfAccount)`. Then you can write `eg.selfAuth` instead of `{}` in each call.

* The assignment to `messagesFlags.default` gives an error because it's not writable.

  It looks like this is meant to basically mock out the `messagesFlags` module. Here's what the Jest docs suggest for doing that job: https://jestjs.io/docs/en/mock-functions#mocking-modules It looks cleaner, too.

  Want to try that? Or if you have tried it and ran into an obstacle: what did you find?

---
```
+    await sleep(3000);
```

Ouch :stopwatch: :face_with_head_bandage: . This makes this test take 3 seconds longer to run than it should. That's a long time for a test!

Again, take a look at Jest's solution for this problem:
https://jestjs.io/docs/en/timer-mocks

For where we use `Date.now`, you may also need the mocking-modules technique linked above, to mock out `Date`.

PR #3564, opened 2019-07-23
dependency: Upgrade react-native-webview to `^5.0.0`

### LINKED ISSUE #3550: Upgrade react-native-webview to 3.x, 4.x, 5.x
We're using `react-native-webview` version 2.0.0. In the months since that release, they've made several more with major changes.

In particular @jainkuniya found with `git bisect` that #3518 was introduced by switching from the RN-included WebView to `react-native-webview`. And he found that upgrading all the way to 5.9.1 fixes #3518. (Described on #3536). AFAIK he didn't test versions in between, so it's possible a much earlier version already fixes that issue. Anyway, we should get to the latest.

The latest version of `react-native-webview`, from last week, comes as a pair ([see NPM](https://www.npmjs.com/package/react-native-webview)): 6.3.1 with AndroidX, and 5.12.1 without. Until #3548 , we'll want the non-AndroidX version, so that means the latest 5.x version.

@jainkuniya also described on #3536 finding that upgrading to 5.9.1 wasn't smooth. So let's do this in steps:
* First, to the latest 3.x.
* ~~Then the latest 4.x.~~ -> looking at the version history, 5.0.1 came less than 3 days after 4.0.0, so perhaps skip this.
* Then the latest 5.x.

For each version:
* Read the release notes. Figure out what breaking changes the maintainers say they made.
* Try the upgrade.
* Fix any issues.
* In the commit message, include:
  * a link to the release notes
  * a mention of any breaking changes that we think might matter
  * a description of what you did about them.

A couple of things we know about already:
* @jainkuniya mentioned on #3536 something changed in 5.x with `postMessage`, and linked to https://github.com/react-native-community/react-native-webview/issues/363#issuecomment-467730963 .
* That thread links in turn to another thread: https://github.com/react-native-community/react-native-webview/issues/66#issuecomment-459804483
* In short it looks like `postMessage` has been dropped in favor of an API that's said to be better. We should switch to that new API.

jainkuniya:
There is a complete rewrite of communication between webview and rn
component starting from 5.0.0

So use following as stated in the release notes
```
window.ReactNativeWebView.postMessage(data)
```
https://github.com/react-native-community/react-native-webview/releases/tag/v5.0.0

Also due to following change in https://github.com/react-native-community/react-native-webview/commit/f3bdab5
```
-￼    stringWithFormat:@"document.dispatchEvent(new MessageEvent('message', %@));",
+    stringWithFormat:@"window.dispatchEvent(new MessageEvent('message', %@));",
```

on iOS `document.addEventListener` stopped working. So change it to `window.addEventListener`.

This also fixes #3550, #3518.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "react-native-sound": "^0.10.9",
     "react-native-text-input-reset": "^1.0.2",
     "react-native-vector-icons": "^4.6.0",
-    "react-native-webview": "2.0.0",
+    "react-native-webview": "^5.0.0",
     "react-navigation": "^1.5.12",
     "react-navigation-redux-helpers": "^1.1.1",
     "react-redux": "^5.0.7",
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -44,7 +44,7 @@ var escapeHtml = function escapeHtml(text) {
 };
 
 var sendMessage = function sendMessage(msg) {
-  window.postMessage(JSON.stringify(msg), '*');
+  window.ReactNativeWebView.postMessage(JSON.stringify(msg));
 };
 
 window.onerror = function (message, source, line, column, error) {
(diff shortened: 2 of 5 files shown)


gnprice:
Thanks @jainkuniya ! Merged.

A couple of comments:
* There was a lint error. I added an `eslint-ignore-line` comment; and then a `prettier-ignore` to prevent that comment from getting moved around to the wrong place.
* I found a cleaner way to distinguish iOS from Android here :slightly_smiling_face:, and added a new commit on top that does that.

PR #3580, opened 2019-08-01
 accountSelector: Export user_id from realm.

jainkuniya:
* realm state: Persist user_id int REALM_INIT.
* accountSelector: Export user_id from realm.

### DIFF
--- a/src/realm/__tests__/realmReducer-test.js
+++ b/src/realm/__tests__/realmReducer-test.js
@@ -22,6 +22,7 @@ describe('realmReducer', () => {
         canCreateStreams: true,
         crossRealmBots: [],
         email: '',
+        user_id: 0,
         isAdmin: false,
         twentyFourHourTime: false,
         emoji: {},
--- a/src/realm/realmReducer.js
+++ b/src/realm/realmReducer.js
@@ -16,6 +16,7 @@ const initialState = {
   canCreateStreams: true,
   crossRealmBots: [],
   email: undefined,
+  user_id: undefined,
   twentyFourHourTime: false,
   emoji: {},
   filters: [],
@@ -38,6 +39,7 @@ const initialState = {
 const fakeBlankState = {
   ...initialState,
   email: '',
+  user_id: 0,
 };
 
 const convertRealmEmoji = (data): RealmEmojiById =>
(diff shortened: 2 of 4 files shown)


gnprice:
Thanks @jainkuniya !

#### afd4fe65f realm state: Persist user_id int REALM_INIT.

This looks good, with one tweak: best to mention `user_id` directly after `email` (in each of the places it appears), because they're so closely related.


#### 1bdd3d8dd accountSelector: Export user_id from realm.

* So the commit message describes this as adding a selector `getOwnUserId`; and then removing another selector -- which happens to have the same name -- on the grounds that it does the same thing.

  Given that the name and also the behavior are the same, I think it's more natural to think of this change as modifying the existing selector `getOwnUserId`. You're changing its implementation, and then also moving it from one file to another.

* Then that makes it natural to ask:
  * Why the edit to the jsdoc to no longer mention `getOwnEmail`?
  * There was a TODO comment. The first part of it is done now (great!). But why delete the second part?

* Also I think `userSelectors` is actually the better home for this. I should probably move `getOwnEmail`, too; it's only in `accountsSelectors` because that's where it originally was, and when that stopped making as much sense in c5b93700e0a5e8e016ae45a78993785782d450d8 I didn't go and move it.

#### ...

I'll be happy to merge this with the fixes above. But the real payoff is what comes next, in that TODO comment that this does the first part of:
```
// TODO get this directly from initial data, like we do `getOwnEmail`;
//   then we can switch `getOwnUser` and other uses of `getOwnEmail` to use
//   user IDs instead!
```

So that would be great to do next. :slightly_smiling_face:

gnprice:
> * Also I think `userSelectors` is actually the better home for this. I should probably move `getOwnEmail`, too; it's only in `accountsSelectors` because that's where it originally was, and when that stopped making as much sense in [c5b9370](https://github.com/zulip/zulip-mobile/commit/c5b93700e0a5e8e016ae45a78993785782d450d8) I didn't go and move it.

And done; see 2f8c7b82a..f528a7465 .

gnprice:
Thanks @jainkuniya ! Merged.

I dropped the last commit, the one deleting `getOwnUser`. It's true it has no callers... but note that just below it we have `getSelfUserDetail`, which does the same job and does have several callers. And I have a "DEPRECATED" note on that one saying we should use `getOwnUser` instead. :slightly_smiling_face: So the fix to `getOwnUser` not having callers is to convert the `getSelfUserDetail` callers to use it.

PR #3586, opened 2019-08-10
Convert all object types to exact

keshav234156:
fixes#3452
this commit convert object type into exact.Have only made change to those that didn't show any error in error with yarn flow.After this only src/message/messageActionSheet.js is left for the changes which shows couple of errors.


ray-kraesig:
Several of these object types are deliberately inexact, sometimes even with adjacent explanatory comments saying so. These should not be blindly made exact, but instead should be marked with Flow's new explicit inexact object syntax.

See the linked issue (and its links) for more details.

Issue #3548, opened 2019-07-09
Upgrade to RN v0.60


gnprice:
Released last week:
https://facebook.github.io/react-native/blog/2019/07/03/version-60

~First we'll want to complete the upgrade to v0.59 `<_<` `>_>`, aka #3399.~ (done!)

From the release notes, notable changes include:
* Migrated to AndroidX -- this will take some work on upgrading.
* [edit:] CocoaPods. See #3983.
* There's a new recommended upgrade process.
* (I should read the rest of the release notes more completely -- getting this down quickly for now.)

Also:
* Flow v0.98.

---

The bulk of this work is now done. Details in https://github.com/zulip/zulip-mobile/issues/3548#issuecomment-638372833 and in #4133 .

Some known changes that we already take advantage of in #4133:
* We can start using "autolinking": #4026
* We upgrade react-native-webview, fixing in particular #4093.

Other known changes that this will open up the option to do -- in particular I went through this thread and followed all the backlinks to related issues:
* We can start trying Hermes: #4131
* We can probably upgrade react-native-unimodules: #4091
  * That may also let us drop a bit of Gradle code: https://github.com/zulip/zulip-mobile/issues/3548#issuecomment-607558197
* We'll gain the feature we need for solving #2772.
* The react-navigation v3 upgrade #3649 becomes a bit simpler.
* We'll gain support for Android's "Downloadable Fonts" feature, which will give us what we need to fix #3885.
* We'll gain a better way to help the user to the permissions menu on iOS: #3814

Known cleanups we should probably follow this with:
* #4118 react-native-orientation

gnprice:
#3399 is now done! We're on RN v0.59, and so this upgrade to v0.60 is no longer blocked.

ray-kraesig:
A list of known dependencies whose update is blocked by the transition to RN v0.60:
* react-native-image-picker
* *(add more as identified)*

gnprice:
(Belatedly updating the assignment metadata.)

gnprice:
Good news on the AndroidX upgrade! I sat down yesterday afternoon to try it, and it looks like it will be (a) a small change that (b) does not depend on the RN upgrade, and that in fact we could have done at any time in the last year or so.

I have a draft commit to do this; I think it's most likely complete as is, but there are a few more things I want to test.

Detailed discussion in this (long!) commit message: 5cec78fe4

chrisbobbe:
[...]

PR #3587, opened 2019-08-11
 netinfo: Migrate to community NetInfo v3.2.1.

### LINKED ISSUE #3568: Migrate to community NetInfo
This is a followup to the RN v0.59 upgrade #3399, and a blocker for the RN v0.60 upgrade #3548.

When launching the app with the debugger, the following message now goes to the console:

> Warning: NetInfo has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-community/netinfo' instead of 'react-native'. See https://github.com/react-native-community/react-native-netinfo

In v0.60, the core module appears to indeed be gone.

The README says the new module's API is compatible, so this may be easy.

jainkuniya:
Fixes #3568.


android/app/src/main/java/com/zulipmobile/MainApplication.java
```diff
@@ -38,6 +39,7 @@ public boolean getUseDeveloperSupport() {
         protected List<ReactPackage> getPackages() {
             return Arrays.asList(
                     new MainReactPackage(),
+            new NetInfoPackage(),
```
gnprice: nit: match indentation

package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@expo/react-native-action-sheet": "^1.1.0",
+    "@react-native-community/netinfo": "3.2.1",
```
gnprice: Please use a leading `^` -- that way a plain `yarn upgrade` in the future will get the latest 3.x.

(Notice that almost all our dependencies use that style.)

flow-typed/npm/react-native-document-picker_vx.x.x.js
```diff
@@ -0,0 +1,33 @@
+// flow-typed signature: e70237f3f0f0f7216ac66fa336d6aecf
+// flow-typed version: <<STUB>>/react-native-document-picker_vgit://github.com/Elyx0/react-native-document-picker.git#v3/flow_v0.92.0
```
gnprice: flow types: Add type for react-native-document-picker.
    
    This dependency was added in d9bbb5e5f047a6de17eef1870fbad17c8debf5f9
    but flow type were not added.
    
    So now just run `flow-typed update` and it is generated automatically.

Heh, we actually do have types for this! Look in the parent directory, and then at the history :slightly_smiling_face: 

Or most practically: go to `ComposeMenu.js` where we use this library, in the `handleFilePicker` function, and try editing `response.name` or something to misspell the property's name. You'll get a Flow error.

I did leave things a little messy in that `yarn update-libdefs` creates this file -- really we want to leave things in a state where it makes no changes unless there are updates to be had. So thanks for tidying that up!

I'll take a few minutes to see if I can teach `yarn update-libdefs` to ignore this library, and if not then I'll merge this commit.

gnprice: Merged, as 6219e5a23 . Thanks again!

gnprice:
(BTW I just edited the PR description to point at the issue. The reference in the commit message is important, but a reference in the PR description is helpful to for cross-linking them here on the GitHub web UI.)

gnprice:
Oh, one more remark on the commit message:

    Not upgrading to latest 4.x.x because it comes with AndroidX, which will
    not work for us yet.

Dang, and that is good to know!

When there's a "yet" like this, I always want to answer the question: "until when?" "what's it blocked on?" In this case, I think the answer is the RN v0.60 upgrade: #3548 . Does that agree with your understanding? If so, a link to that issue would be a good way to answer the question, like: "which will not work for us until we upgrade to RN v0.60 (#3548)".

package.json
jainkuniya: Yup!

But I thought, there will not be any new 3.x.x release, as they already moved to 4.x.x

Anyway I agree it to have `^`, doing it 👍

gnprice:
Looks good, merged -- thanks @jainkuniya !

PR #3606, opened 2019-09-05
lint: Add `flowtype/no-dupe-keys` to eslint

ray-kraesig:
Activate a linter in `eslint-plugin-flowtype` to warn about duplicate
keys in type declarations. Without this lint, flow does not warn when
a later-declared key in the same type shadows an earlier one, even
when the two are declared with incompatible types.

This presently only catches a duplicate parameter in modelTypes.js
(also fixed in this commit).


gnprice:
Thanks @ray-kraesig ! Merged.

PR #3609, opened 2019-09-11
Typify url-tests.js

ray-kraesig:
This was slightly more involved than it probably should have been due to a tacit use of `undefined`. (See description of 88dd8d910.)


gnprice:
Thanks @ray-kraesig ! Merged.

The Travis failure looks like it's caused by my recent 338036e0c . WFM locally with `tools/test jest --full`, but something's different in the Travis environment. :-/ I'll investigate.

Issue #3567, opened 2019-07-24
Trailing slash on realm URL can happen, and breaks `/events`


gnprice:
[Original title: "Message list won't update, on iOS on physical device". Better characterized below after some debugging.]

AFAIK this doesn't appear in any actual release we've made -- but it happens on 25.7.121 and 25.8.122 (aka current master), in their respective alpha release builds, on my iPhone 7 running iOS 12.3. So we can't make a release until it's fixed.

It also happens in a debug build on that same device.

It *doesn't* happen in a debug build on an iPhone XR simulator running iOS 12.4... nor on an iPhone 7 simulator running iOS 12.4.

The symptoms are:
* When looking at a message list, it (usually!) fails to update when something changes -- a message arrives, a reaction, etc.
* When you send a message, it goes through fine... but the "outbox" object, which shows up as a message with a spinner on it, sticks around. It's still there even if you leave the message list and come back.

gnprice:
Ohhh -- it's that we're just not getting events.

First clue to this was on looking in the debug console, seeing this happening over and over:
```
{route: "events?queue_id=1563767601%3A50342&last_event_id=-1", params: {…}, httpStatus: 404, json: undefined}
```

But wait, why was I seeing updates if I navigated out of the message list and then back to it? ... It's because we were doing a new fetch. I'm not entirely sure why that is -- don't we avoid those if we're all caught up? the log shows we had `caughtUp` of `{older: false, newer: true}` for the narrow in question, hmm -- but the log shows we were, and that certainly explains getting the updates then.

I thought I'd spotted an update another time, too, while still looking at the message list. Maybe I'd somehow provoked another fetch? Or maybe I'd just confused that with seeing an update after navigating away and back.

Anyway, that certainly explains all the symptoms. Next question: what's in turn causing it.

gnprice:
Puzzling. Odd facts include:
* It works fine for (my usual account on) chat.zulip.org. Where it doesn't work is a test org on kandra-test.zulipchat.com.
  * Possibly another symptom of #3196, somehow? That test org still has that setting enabled.
* I'm not seeing anything corresponding to these requests when I look in the zulipchat.com server logs. There are no 404s at all in `/var/log/zulip/tornado.log` (which has numerous 200 lines for successful requests of the same type); presumably that means these requests aren't making it to Tornado, but I don't see any sign of them in `/var/log/nginx/{access,error}.log` either.

gnprice:
[...]

PR #3611, opened 2019-09-12
Allow user to specify "localhost:9991" as a realm

ray-kraesig:
... and in the process, unify the URL parsing logic between `SmartUrlInput` and `url.js`.

This permits simpler use of the iOS simulator, which has also been documented.

### DIFF
--- a/src/common/SmartUrlInput.js
+++ b/src/common/SmartUrlInput.js
@@ -6,6 +6,7 @@ import type { NavigationEventSubscription, NavigationScreenProp } from 'react-na
 
 import type { Context } from '../types';
 import { autocompleteRealmPieces, autocompleteRealm, fixRealmUrl } from '../utils/url';
+import type { Protocol } from '../utils/url';
 import RawLabel from './RawLabel';
 
 const styles = StyleSheet.create({
@@ -31,7 +32,7 @@ type Props = {|
    * The protocol which will be used if the user doesn't specify one.
    * Should almost certainly be "https://".
    */
-  defaultProtocol: string,
+  defaultProtocol: Protocol,
   /**
    * The example organization name that will be displayed while the
    * entry field is empty. Appears, briefly, as the initial (lowest-
--- a/src/common/SmartUrlInput.js
+++ b/src/common/SmartUrlInput.js
@@ -5,7 +5,7 @@ import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet
 import type { NavigationEventSubscription, NavigationScreenProp } from 'react-navigation';
 
 import type { Context } from '../types';
-import { autocompleteUrl, fixRealmUrl, hasProtocol } from '../utils/url';
+import { autocompleteRealmPieces, autocompleteRealm, fixRealmUrl } from '../utils/url';
 import RawLabel from './RawLabel';
 
 const styles = StyleSheet.create({
@@ -27,9 +27,22 @@ const styles = StyleSheet.create({
 
 type Props = {|
   defaultValue: string,
+  /**
+   * The protocol which will be used if the user doesn't specify one.
+   * Should almost certainly be "https://".
+   */
+  defaultProtocol: string,
+  /**
+   * The example organization name that will be displayed while the
+   * entry field is empty. Appears, briefly, as the initial (lowest-
+   * level) component of the realm's domain.
+   */
   defaultOrganization: string,
-  protocol: string,
-  append: string,
+  /**
+   * The default domain to which the user's input will be appended, if
+   * it appears not to contain an explicit domain.
+   */
+  defaultDomain: string,
   navigation: NavigationScreenProp<mixed>,
   style?: ViewStyleProp,
   onChangeText: (value: string) => void,
@@ -38,6 +51,10 @@ type Props = {|
 |};
 
 type State = {|
+  /**
+   * The actual input string, exactly as entered by the user,
+   * without modifications by autocomplete.
+   */
   value: string,
 |};
 
@@ -70,8 +87,8 @@ export default class SmartUrlInput extends PureComponent<Props, State> {
   handleChange = (value: string) => {
     this.setState({ value });
 
-    const { append, protocol, onChangeText } = this.props;
-    onChangeText(fixRealmUrl(autocompleteUrl(value, protocol, append)));
+    const { onChangeText, defaultProtocol: protocol, defaultDomain: domain } = this.props;
+    onChangeText(fixRealmUrl(autocompleteRealm(value, { protocol, domain })));
   };
 
   urlPress = () => {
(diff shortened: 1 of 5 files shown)


src/utils/url.js
```diff
@@ -67,11 +90,40 @@ const mimes = {
...
+
+/**
+ * Given user input purporting to be a Zulip server's host name, provide
+ * a prefix, derived value, and suffix which may suffice to turn it into
+ * a real URL.
+ *
+ * Presently, the derived value will always be equal to the input value;
+ * this should not be relied on, as it may change in future.
+ */
+export const autocompleteHostnamePieces = (
```
gnprice: I don't think "hostname" is the right term to use here. The input might
* be an IP address, with no hostname present
* contain a port
* contain a scheme (and `//`)

But what these (and the hostname case) all have in common is that they're fragments of a URL.

So I think "autocomplete URL" is the right central words to describe what this is doing -- it's taking a fragment of a URL (that contains the most salient bits), and filling it out into a real, complete URL.

Of course it's not just any URL, and the UI context that makes this particular autocompletion helpful wouldn't make sense for lots of kinds of URLs. It's really about specifically autocompleting a Zulip organization URL, or "realm" URL as it's often called in the API and code.

So a good more specific name might be `autocompleteRealmUrl` (and `...Pieces`).

src/utils/url.js
```diff
@@ -44,10 +44,33 @@ export const getResource = (
...
+};
+
+/** DEPRECATED */
+export const fixRealmUrl = (url: string = '') => {
+  if (url === '') {
+    return '';
+  }
+  const trimmedUrl = url
+    .replace(/\s/g, '') // strip any spaces, internal or otherwise
+    .replace(/\/+$/g, ''); // eliminate trailing slashes
```
gnprice: This looks like a behavior change! Should be its own (tiny) commit separate from the refactors.

Certainly it's a helpful change -- see #3567 and in particular https://github.com/zulip/zulip-mobile/issues/3567#issuecomment-514844325 .

src/utils/url.js
```diff
@@ -67,11 +90,40 @@ const mimes = {
...
+export const autocompleteHostnamePieces = (
+  value: string,
+  data: AutocompletionData,
+): AutocompletionPieces => {
+  const [protocol, nonProtocolValue] = parseProtocol(value);
+
+  const prefix: string | null = protocol === null ? data.protocol : null;
+
+  const suffix: string | null =
+    nonProtocolValue.includes('.') || nonProtocolValue.includes(':') ? null : `.${data.hostname}`;
```
gnprice: The behavior change on `:` here would be great to pull out as a (very small) commit of its own, coming after the refactor.

If you do, and also the small other behavior change I mentioned above, does the main commit become a completely pure refactor? If so, that's extra helpful.

src/utils/url.js
```diff
@@ -67,11 +90,40 @@ const mimes = {
...
+ */
+export const autocompleteHostnamePieces = (
+  value: string,
+  data: AutocompletionData,
+): AutocompletionPieces => {
+  const [protocol, nonProtocolValue] = parseProtocol(value);
+
+  const prefix: string | null = protocol === null ? data.protocol : null;
+
+  const suffix: string | null =
```
gnprice: I think this and the annotation on `prefix` can be left quite well for type inference.

(Perhaps in an earlier version there was an error related to the inhomogeneous list they get put into? Seems fine now -- perhaps your explicit annotation on the return type took care of that.)

src/common/SmartUrlInput.js
```diff
@@ -96,22 +98,22 @@ export default class SmartUrlInput extends PureComponent<Props, State> {
...
     } = this.props;
-    let { value } = this.state;
-    let showAppend = value.indexOf('.') === -1;
-    if (defaultValue && value.length === 0) {
-      showAppend = false;
-      value = defaultValue;
-    }
+    const { value } = this.state;
+
+    const [prefix, , suffix] = autocompleteHostnamePieces(value ?? defaultValue, {
```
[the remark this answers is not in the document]
ray-kraesig: It is, though! I'm reluctant to condemn TC-39-blessed syntax just because it looks unusual at first. At [the top of that ESlint page](https://eslint.org/docs/rules/no-unused-vars), there's even an example using it.

Aesthetically, I also prefer `_`-as-ignored-value (as also seen in at least Rust and Haskell), but you'd still get complaints if you're ignoring more than one value. (Unless you name them... but then you have to name them, and then someone *will* eventually come along and blithely use those names.)

Also, while it's not in use in zulip-mobile, some people do bind `_` to lodash.

The argument for `const { prefix, suffix } = ...` is a lot stronger, of course (as anyone who's ever used [the perl `stat` function](https://perldoc.perl.org/functions/stat.html) can attest). I think returning a tuple is okay for this particular case, since it's chopping a string (which is inherently sequential) into a very specific and very small number of pieces. But it's marginal, at best.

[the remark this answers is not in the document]
ray-kraesig: Remnant of a previous version. Thanks for the catch.

src/utils/url.js
ray-kraesig: The autocompletion mechanism is theoretically somewhat more general than just Zulip realms, but it *is* very strongly realm-oriented. Arguably, I should have hardcoded `zulipchat.com` at this level – I left it general because it started that way, but I don't think this particular bit of configurability will ever buy anyone anything.

At any rate, I'll use that name (or `autocompleteRealm`, possibly).

src/utils/url.js
ray-kraesig: It is, but after reconsideration, I'm not sure it's the right thing to do, at least on non-Zulip domains: I'm pretty sure it's possible to configure a(n admittedly bizarre) proxy server such that the extra slashes are needed.

I've left it out for now; it's an easy change.

ray-kraesig:
Note that despite GitHub's ordering of these commits, the docs update is indeed a child commit of the "do not append domains" commit.

I've left off converting to `const {prefix, suffix} = ...`, on the theory that I've done enough violence to recorded history as it is; that can easily be added in a separate commit, if you still think it's worth doing.

gnprice:
> Note that despite GitHub's ordering of these commits, the docs update is indeed a child commit of the "do not append domains" commit.

Yeah, GitHub unfortunately has a bug that they've ignored reports of forever, where they show commits inappropriately ordered by author date.

(My guess has always been that someone decided topo-sorting was too hard and/or complicated, and didn't understand Git committer dates -- sorting by committer date would be implemented almost exactly the same, and fix the problem in almost all cases. But who knows.)

Workaround is to ignore the lies on the GitHub UI :disappointed: and take one more reason to look at PRs with `git` itself.

src/utils/url.js
```diff
@@ -67,11 +101,47 @@ const mimes = {
...
+  protocol: Protocol,
+  domain: string,
+|};
+
+export type AutocompletionPieces = [Protocol | null, string, string | null];
+
+/**
+ * A short list of some characters not permitted in subdomain name elements.
+ */
+const disallowedCharacters: Array<string> = [...'.:/'];
```
gnprice: Ah, this is a neat trick. :-)

PR #3612, opened 2019-09-14
android, notif: Narrow to conversation on click even if app is in bg.

### LINKED ISSUE #3582: notification: On clicking doesn't narrow down to conversation if app was in background.
I was using Zulip app few minutes ago, now I switched to another app. But still Zulip app is running in background. And as I gets a notification of PM message, I clicked on it. It doesn't took me to the PM conversation, rather ended up on the screen where I was when I switched to another app.

I reproduced it on the Android 9 One Plus Six Android Phone.

jainkuniya:
Notify react about the notification when app is:
- in forground and notification comes (LifecycleState = RESUMEND)
when user is using app and notification comes
- in background and notification comes (LifecycleState = BEFORE_RESUME)
when user is not using app. But app is present in the recent apps.

and finally launch (bring app to foreground)
MainActivity after emiting event.

Fixes #3582

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/NotifyReact.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/NotifyReact.java
@@ -28,29 +28,37 @@ class NotifyReact {
     static void notifyReact(ReactApplication application, final Bundle data) {
         NotificationsModule.initialNotification = data;
 
-        if (!emitIfResumed(application, "notificationOpened", Arguments.fromBundle(data))) {
-            // The app will check initialNotification on launch.
-            Log.d(TAG, "notifyReact: launching main activity");
-            launchMainActivity((Context) application);
-        }
+        emitIfResumed(application, "notificationOpened", Arguments.fromBundle(data));
+        Log.d(TAG, "notifyReact: launching main activity");
+        bringAppToForegroundIfNot(application);
+
     }
 
-    private static boolean emitIfResumed(ReactApplication application, final String eventName, final @Nullable Object data) {
+    private static void bringAppToForegroundIfNot(ReactApplication application) {
+        final ReactNativeHost host = application.getReactNativeHost();
+        final ReactContext reactContext = host.getReactInstanceManager().getCurrentReactContext();
+        if (reactContext != null && reactContext.getLifecycleState() == LifecycleState.RESUMED) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @jainkuniya !

Can you say a bit more about what you learned from debugging?

Also, this change appears to do two things:
* When the lifecycleState is `BEFORE_RESUME`, we'd previously not `emit`, and would `launchMainActivity`. Now we'll `emit`, and then `launchOrResumeMainActivity`.
* When the lifecycleState is `RESUMED`, we'd previously `emit`, and would not `launchMainActivity`. Now we'll `emit` and then also `launchOrResumeMainActivity`.

Is the change in the `RESUMED` case intentional?

gnprice:
Can you say a bit more about what you learned from debugging? This doesn't appear to answer that question to any greater degree than the previous version did :slightly_smiling_face: 

There are a number of comments I'd make on the specifics of this implementation. But it's premature to get into the details of a solution before being clear about what we think the problem is.

jainkuniya:
Problem is:
On master, we are only emitting event to react if `LifecycleState == RESUMED`, which only works in one scenario, i.e user is using App and notification comes (in this case LifecycleState is `RESUMED`).

But what if `LifecycleState` is `BEFORE_RESUME`, i.e user was using Zulip app few minutes back and currently on other app (Zulip app is present in the recent apps). In this case we weren't notifying react. We were just launching activity, which is more sort of bringing activity to foreground (as we have set activity to have single instance).

So now I am notifying react every time except when app is in the state of `BEFORE_CREATE`, because in this case activity will be created and react will go in the flow of `initialNotification`.

gnprice:
Cool, thanks!

From that diagnosis, it seems like the case where we want to change the behavior is where the lifecycleState is `BEFORE_RESUME`. The existing behavior is correct in the other cases, right?

This branch (and the single main commit at the start of it) changes the logic rather more than that, and makes it more complex. Among other things, it duplicates the application -> host -> reactContext logic... and leaves out the `if (!host.hasInstance())` condition, which according to the comment there means it's likely to crash sometimes with an assertion failure.

I think from here it may be easiest for me to go ahead and see about implementing a fix myself -- it may be simpler to demonstrate a way of structuring this code than to explain. Thanks for the work investigating this issue! It's been very helpful.

PR #3613, opened 2019-09-14
types: Describe current state of reaction object in redux.

jainkuniya:
When messages are fetched from server via GET `/messages`, then it
contains `id` as key in `user` of the reaction object. While reaction
add event has `user_id` in it.

As we weren't using this attributes anywhere, so there was no problem
and fortunately app never crashed.


gnprice:
Thanks @jainkuniya ! Merged.

PR #3615, opened 2019-09-16
dependencies: use ordinary npm version of document-picker

ray-kraesig:
The existing unusual github reference for `react-native-document-picker`
was causing `flow-typed` to generate a stub file it thereafter couldn't
read.

Replace this (which seems to have effectively been 3.2.0) with an
ordinary ref to ^3.2.4.


gnprice:
Thanks @ray-kraesig ! Merged.

PR #3616, opened 2019-09-16
docs: add troubleshooting section to setup

ray-kraesig:
Add a paragraph concerning a misleading error message observed during
initial setup.

### DIFF
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -199,10 +199,10 @@ When trying to build the Android app, you may see this error:
 Task 'installDebug' not found in project ':app'.
 ```
 
-This (somewhat misleading) error message has been seen when the Android SDK is
-not fully configured. Creating a new React Native project (per [the upstream
-docs](https://facebook.github.io/react-native/docs/getting-started.html)) and
-running `react-native android` may give clearer error messages.
+This (somewhat misleading) error message can occur when the Android
+SDK is not fully configured.  Creating a new React Native project (per
+[the upstream docs][rn-installation]), and running `react-native
+android` there, may give clearer error messages for debugging.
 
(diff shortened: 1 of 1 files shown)


docs/howto/build-run.md
```diff
@@ -190,6 +190,24 @@ version number starts with "1.8", and with JDK 11, it starts with
...
+
+When trying to build the Android app, you may see this error:
+
+```
+* What went wrong:
+Task 'installDebug' not found in project ':app'.
+```
+
+This has been known to occur when React Native has not yet been set up to use
+the Android Development Kit. Check the [React Native installation
```
gnprice: This line doesn't mean much to me: "React Native has not yet been set up to use the Android Development Kit." Partly because AFAICT "the Android Development Kit" isn't the name of anything; I think you probably mean the Android SDK?

Then for "setting up React Native to use" that... I suspect there's a more concrete description we can make. Is this referring to the `PATH` and `ANDROID_HOME` bit in the upstream docs?

ray-kraesig: a) Yes, lost a word. Tentatively changed to "has not yet been configured to use the Android SDK", although "the Android SDK has not yet been configured for React Native's use" is arguably equally true.

b) That was part of it, and may even have been the final part, but I'm pretty sure it wasn't the only part. I'm afraid I only made a note of the fact that starting an empty project gave me clear and precise error messages, not what those messages were.

With that in mind, how about...

> This (somewhat misleading) error message has been seen when the Android SDK is not fully configured. Creating a new React Native project (per [the upstream docs](https://facebook.github.io/react-native/docs/getting-started.html), at **Creating a new application**) and running `react-native android` within it may give clearer error messages.

gnprice: Sure, that sounds good.

gnprice:
Thanks! Merged.

PR #3619, opened 2019-09-18
Make new mentioned/starred messages appear in mentions/starred narrows

### LINKED ISSUE #3592: New messages don't appear in mentions narrow
Forking from [here](https://github.com/zulip/zulip-mobile/pull/3542#issuecomment-523670583), where I noticed this in reading the code.

The `narrowsReducer` relies on `isMessageInNarrow`, which has this logic:

```
  return caseNarrow(narrow, {
    home: () => true,
    stream: name => name === message.display_recipient,
    // ...
    starred: () => message.type === 'starred',
    mentioned: () => message.type === 'mentioned',
    allPrivate: () => message.type === 'private',
```

Those `starred` and `mentioned` conditions can't ever be true.

And indeed, if you open the @-mentions narrow and then (from some other user on another device) send a message with an @-mention that should appear there, it doesn't.

This "mostly" works, because most of the time the user will enter that narrow cold and so we'll load the messages from the server. But we should fix it.

I think it's trickier to reproduce buggy behavior here for the `starred` case, but we should fix that too.

Note that the correct condition for `mentioned` will include both the `mentioned` flag and the `wildcard_mentioned` flag; see [this other comment](https://github.com/zulip/zulip-mobile/pull/3542#issuecomment-523673571) .

In addition to the new-message case, we should fix the `EVENT_UPDATE_MESSAGE_FLAGS` case too. That one currently handles the "starred" narrow but not the "mentioned" narrow.

chrisbobbe:
Fixes #3592.

I ran into some Babel issues with Jest and haven't been able to run the tests yet; could you please take a quick look at the logs (attached) to see if this is a known issue? Since I'm assuming everything works for people with development environments that have been set up for a while, and the error was thrown from code I didn't touch, there may have been a breaking change in a minor version of Babel or Jest that was triggered by the carat in package.json allowing silent minor version upgrades (e.g., "jest": "^24.8.0"), but that's just a guess until I can dig deeper or learn that it's a known issue.

[Issue 3592 Test Output (babel issue?).txt](https://github.com/zulip/zulip-mobile/files/3624392/Issue.3592.Test.Output.babel.issue.txt)

I'm proposing a change to the eventMiddleware logic (first commit) that would ensure that `flags` is always set on `action.message.flags`, at least as an empty array, if `event.flags` does not exist, to help with the "crunchy shell" pattern: https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md.

### DIFF
--- a/src/chat/__tests__/narrowsReducer-test.js
+++ b/src/chat/__tests__/narrowsReducer-test.js
@@ -83,7 +83,13 @@ describe('narrowsReducer', () => {
 
       const action = deepFreeze({
         type: EVENT_NEW_MESSAGE,
-        message: { id: 3, type: 'stream', display_recipient: 'stream', subject: 'topic', flags: [] },
+        message: {
+          id: 3,
+          type: 'stream',
+          display_recipient: 'stream',
+          subject: 'topic',
+          flags: [],
+        },
         caughtUp: {},
       });
 
--- a/src/chat/__tests__/narrowsReducer-test.js
+++ b/src/chat/__tests__/narrowsReducer-test.js
@@ -37,7 +37,7 @@ describe('narrowsReducer', () => {
 
       const action = deepFreeze({
         type: EVENT_NEW_MESSAGE,
-        message: { id: 3, display_recipient: 'some stream', subject: 'some topic' },
+        message: { id: 3, display_recipient: 'some stream', subject: 'some topic', flags: [] },
         caughtUp: {},
       });
 
@@ -57,7 +57,7 @@ describe('narrowsReducer', () => {
 
       const action = deepFreeze({
         type: EVENT_NEW_MESSAGE,
-        message: { id: 3 },
+        message: { id: 3, flags: [] },
         caughtUp: {
           [HOME_NARROW_STR]: {
             older: false,
@@ -83,7 +83,7 @@ describe('narrowsReducer', () => {
 
       const action = deepFreeze({
         type: EVENT_NEW_MESSAGE,
-        message: { id: 3, type: 'stream', display_recipient: 'stream', subject: 'topic' },
+        message: { id: 3, type: 'stream', display_recipient: 'stream', subject: 'topic', flags: [] },
         caughtUp: {},
       });
 
@@ -104,6 +104,7 @@ describe('narrowsReducer', () => {
       id: 1,
       type: 'private',
       display_recipient: [{ email: 'me@example.com' }, { email: 'a@a.com' }, { email: 'b@b.com' }],
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
@@ -132,6 +133,7 @@ describe('narrowsReducer', () => {
       type: 'stream',
       display_recipient: 'some stream',
       subject: 'some topic',
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
@@ -167,6 +169,7 @@ describe('narrowsReducer', () => {
     const message = {
       id: 1,
       display_recipient: [{ email: 'me@example.com' }],
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
@@ -204,6 +207,7 @@ describe('narrowsReducer', () => {
       type: 'stream',
       display_recipient: 'some stream',
       subject: 'some topic',
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
@@ -240,6 +244,7 @@ describe('narrowsReducer', () => {
       type: 'stream',
       display_recipient: 'stream name',
       subject: 'some topic',
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
@@ -274,6 +279,7 @@ describe('narrowsReducer', () => {
       type: 'private',
       sender_email: 'someone@example.com',
       display_recipient: [{ email: 'me@example.com' }, { email: 'mark@example.com' }],
+      flags: [],
     };
     const action = deepFreeze({
       type: EVENT_NEW_MESSAGE,
--- a/src/chat/narrowsReducer.js
+++ b/src/chat/narrowsReducer.js
@@ -2,6 +2,7 @@
 import union from 'lodash.union';
 
 import type { NarrowsState, Action } from '../types';
+import { ensureUnreachable } from '../types';
 import {
   DEAD_QUEUE,
   LOGOUT,
@@ -75,6 +76,7 @@ const eventUpdateMessageFlags = (state, action) => {
           [narrowStr]: state[narrowStr].filter(id => !messagesSet.has(id)),
         });
       } else {
+        ensureUnreachable(operation);
         throw new Error(
           `Unexpected operation ${operation} in an EVENT_UPDATE_MESSAGE_FLAGS action`,
         );
--- a/src/chat/narrowsReducer.js
+++ b/src/chat/narrowsReducer.js
@@ -74,6 +74,10 @@ const eventUpdateMessageFlags = (state, action) => {
         updates.push({
           [narrowStr]: state[narrowStr].filter(id => !messagesSet.has(id)),
         });
+      } else {
+        throw new Error(
+          `Unexpected operation ${operation} in an EVENT_UPDATE_MESSAGE_FLAGS action`,
+        );
       }
     }
 
--- a/src/chat/narrowsReducer.js
+++ b/src/chat/narrowsReducer.js
@@ -13,7 +13,7 @@ import {
   EVENT_UPDATE_MESSAGE_FLAGS,
 } from '../actionConstants';
 import { LAST_MESSAGE_ANCHOR, FIRST_UNREAD_ANCHOR } from '../constants';
-import { isMessageInNarrow, STARRED_NARROW_STR } from '../utils/narrow';
+import { isMessageInNarrow, MENTIONED_NARROW_STR, STARRED_NARROW_STR } from '../utils/narrow';
 import { NULL_OBJECT } from '../nullObjects';
 
 const initialState: NarrowsState = NULL_OBJECT;
@@ -64,17 +64,26 @@ const eventUpdateMessageFlags = (state, action) => {
   const messagesSet = new Set(messages);
   const updates = [];
 
-  if (flag === 'starred' && state[STARRED_NARROW_STR]) {
-    if (operation === 'add') {
-      updates.push({
-        [STARRED_NARROW_STR]: [...state[STARRED_NARROW_STR], ...messages].sort(),
-      });
-    } else {
-      // operation === 'remove'
-      updates.push({
-        [STARRED_NARROW_STR]: state[STARRED_NARROW_STR].filter(id => !messagesSet.has(id)),
-      });
+  const updateFlagNarrow = narrowStr => {
+    if (state[narrowStr]) {
+      if (operation === 'add') {
+        updates.push({
+          [narrowStr]: [...state[narrowStr], ...messages].sort(),
+        });
+      } else if (operation === 'remove') {
+        updates.push({
+          [narrowStr]: state[narrowStr].filter(id => !messagesSet.has(id)),
+        });
+      }
     }
+
+    return updates;
+  };
+
+  if (flag === 'starred') {
+    updateFlagNarrow(STARRED_NARROW_STR);
+  } else if (['mentioned', 'wildcard_mentioned'].includes(flag)) {
+    updateFlagNarrow(MENTIONED_NARROW_STR);
   }
 
   if (!updates.length) {
--- a/src/events/__tests__/eventMiddleware-test.js
+++ b/src/events/__tests__/eventMiddleware-test.js
@@ -1,9 +1,8 @@
-import deepFreeze from 'deep-freeze';
-
 import eventMiddleware from '../eventMiddleware';
+import { NULL_ARRAY } from '../../nullObjects';
 
 describe('eventMiddleware', () => {
-  test('if `event.flags` key exist, move it to `event.message.flags`', () => {
+  test('if `event.flags` key exists, move it to `event.message.flags`', () => {
     const state = { session: {} };
     const event = {
       type: 'message',
@@ -16,12 +15,13 @@ describe('eventMiddleware', () => {
     expect(event.message.flags).toEqual(['mentioned']);
   });
 
-  test('if `event.flags` do not exist, do not mutate event', () => {
+  test('if `event.flags` does not exist, set message.flags to an empty array', () => {
     const state = { session: {} };
-    const event = deepFreeze({
+    const event = {
       type: 'message',
       message: {},
-    });
-    expect(() => eventMiddleware(state, event)).not.toThrow();
+    };
+    eventMiddleware(state, event);
+    expect(event.message.flags).toEqual(NULL_ARRAY);
   });
 });
--- a/src/events/eventMiddleware.js
+++ b/src/events/eventMiddleware.js
@@ -5,6 +5,7 @@ import type { GeneralEvent, GlobalState, MessageEvent } from '../types';
 import { isHomeNarrow, isMessageInNarrow } from '../utils/narrow';
 import { getActiveAccount, getChatScreenParams, getOwnEmail, getIsActive } from '../selectors';
 import { playMessageSound } from '../utils/sound';
+import { NULL_ARRAY } from '../nullObjects';
 
 export default (state: GlobalState, event_: GeneralEvent) => {
   switch (event_.type) {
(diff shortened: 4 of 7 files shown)


chrisbobbe:
Also, I used a rebase to reorder the commits so that the eventMiddleware commit (ensuring the existence of message.flags; ca5f351) was done first, since technically it is required to avoid an error that would have been introduced in the 2fa1684 commit, since that commit has a call to .includes on message.flags without first checking that message.flags exists. You can see this reorder in my fork: https://github.com/chrisbobbe/zulip-mobile/commits/issue-3592 but I'm puzzled why my original order (the incorrect order) is showing up here.

chrisbobbe:
Oh interesting, I can see the test results from Travis; it says, "TypeError: Cannot add property flags, object is not extensible".

If `event.message` not extensible now, I wonder why it wasn't treated as extensible before my commits; it was still assigning .flags to `event.message` where it was not defined before, right? I'm assuming this test passed before my commits; once I get Jest working locally, I can dig deeper into this. At a glance right now, I've traced `event` back to the JSON from the fetch() call I and haven't found a call to Object.freeze() or Object.preventExtensions().

Summary of all failing tests
 FAIL  src/events/__tests__/eventMiddleware-test.js
  ● eventMiddleware › if `event.flags` does not exist, set message.flags to an empty array
    TypeError: Cannot add property flags, object is not extensible
      18 |       if (!event.message.flags) {
      19 |         // $FlowFixMe Message is readonly to serve our use of it in Redux.
    > 20 |         event.message.flags = event.flags || NULL_ARRAY;
         |                               ^
      21 |         delete event.flags;
      22 |       }
      23 | 
      at _default (src/events/eventMiddleware.js:20:31)
      at Object.<anonymous> (src/events/__tests__/eventMiddleware-test.js:26:34)

chrisbobbe:
Oops -- I didn't see your comment from yesterday on the issue:

> Over on the linked thread we're still sorting out exactly how to handle wildcard_mentioned. So for this issue, let's not worry about it.

until now, having already implemented an opinion about 'wildcard_mentioned' (i.e., that it's a separate flag that needs to be checked everywhere we check 'mentioned'). Sorry about that; I'll make changes of course, as necessary, just let me know.

gnprice:
Thanks @chrisbobbe !

> Oops -- I didn't see your comment from yesterday [about `wildcard_mentioned`]

No problem. This looks like a good simple-to-implement way of handling it; if we end up combining the flags earlier in the pipeline, we can adjust this code then.

---

> Oh interesting, I can see the test results from Travis; it says, "TypeError: Cannot add property flags, object is not extensible".

Ah, so: look closely at the top and bottom of the failure text you quoted. This is happening when running `eventMiddleware-test.js`... and specifically at this line:
```
      at Object.<anonymous> (src/events/__tests__/eventMiddleware-test.js:26:34)
```
That's the `eventMiddleware` call in the "if `event.flags` does not exist" test case.

So I think it's caused by the `deepFreeze` in that test case. The bulk of our tests use `deepFreeze` on their fixture data, in order to catch unintended mutations. In this case the mutation is intended, so the simple fix is to remove the `deepFreeze`.

(Whether using mutation here is a *good idea* is another question. Most likely it isn't and we'd be better off keeping things unmutated and making a new object here, just as we do in the bulk of the app. But that's probably a larger refactor.)

---

The bigger thing the CI run finds is actually a series of Flow errors. These are all of the form:

> Cannot call `message.flags.includes` because property `includes` is missing in undefined [1].

That's because the `Message` type still allows `flags` to be absent, which is because it's:
```
   *  * Absent in the Redux `state.messages`; we move the information to a
   *    separate subtree `state.flags`.
```

You can run our test suite with `tools/test`, or just Flow with `tools/test flow` (or `yarn flow`). More info [here](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/testing.md). That will be helpful for iterating to get to a tests-clean state.

Specific comments coming next.

gnprice:
> Also, I used a rebase to reorder the commits so that the eventMiddleware commit (ensuring the existence of message.flags; [ca5f351](https://github.com/zulip/zulip-mobile/commit/ca5f351b050f74497816095a694e68d3764bd474)) was done first, since technically it is required to avoid an error that would have been introduced in the [2fa1684](https://github.com/zulip/zulip-mobile/commit/2fa16849ba559bf72871a737e3a711550366258e) commit, since that commit has a call to .includes on message.flags without first checking that message.flags exists.

Perfect, thanks!

> You can see this reorder in my fork: https://github.com/chrisbobbe/zulip-mobile/commits/issue-3592 but I'm puzzled why my original order (the incorrect order) is showing up here.

Yeah, this is unfortunately a longtime bug in GitHub: they'll show commits inappropriately ordered by author date, making them out of order compared to the actual Git history.

(My guess has always been that someone decided topo-sorting was too hard and/or complicated, and wasn't fully aware of Git committer dates -- sorting by committer date would be implemented almost exactly the same as by author date, and fix the problem in almost all cases. But who knows.)

When reviewing a PR, I fetch it locally and mainly read it with `git log --stat -p` and `git log --oneline` (or actually [my aliases](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3) `git usp` and `git uv`), rather than the GitHub web UI. So I usually don't actually even notice the order GitHub displays the commits in. This issue is one of the reasons for that approach.

gnprice:
#### ca5f351b0 eventMiddleware: Set event.message.flags to [] even if absent on event.

I like this change! It makes the behavior simpler to describe, and also brings it in line with what that comment on `Message.flags` said it already was.

Test failure needs to be fixed; discussed above.

This step isn't enough to make it possible for the rest of the app to become a "soft center" with respect to this property's presence, because in the Redux state we leave it off and maintain the flags data separately. But it's still a move in the right direction.


#### 2fa16849b isMessageInNarrow: Fix check for starred or mentioned messages.

This looks good, except the Flow error. I believe that at each call site it's actually the case that this `Message` has a `flags` property... but because we use one `Message` type for a few different purposes, Flow doesn't see that.

I think the best immediate fix is to just include the presence checks that the type says are necessary.

Hmm, or for a more principled alternative: assert the property is there (i.e. throw an error if not.) That way if we do introduce a bug where the property is missing here because the flags are located elsewhere, we'll almost certainly notice in dev, rather than silently behave as if all flags are false. (But if that gets cumbersome, no need; as the next commit shows, the status quo is already that we just do the presence checks and silently assume that if there were flags set they'd be present here.)


#### 3e53bf7de Fix 2 checks of 'mentioned' flag  without checking 'wildcard_mentioned'

Thanks for this fix! I agree this is the right behavior in both of these spots.

This version causes Flow errors much like the parent, for the same reason.


#### 4638c8c25 narrowsReducer/eventUpdateMessageFlags: Handle mentioned/wildcard_mentioned

This looks right!

I'd like to avoid duplicating the code in this block, though. How about factoring it out into a function so that we can say something like
```
  if (flag === 'starred') {
    return updateFlagNarrow(state, STARRED_NARROW_STR, operation, messages);
  }
  return state;
```

Then to handle mentions is a short `else if (['mentioned', 'wildcard_mentioned'].includes(flag))` block.

chrisbobbe:
Oh OK, yes, I can see now that the test results showing the Flow issues and the "object is not extensible" message do indeed show up when testing in my local development environment. I saw the 20 occurrences of "SyntaxError: Unexpected token import" (shown in the .txt doc in my first post here) and assumed that there was some Babel mis-configuration interfering with all tests, without looking closer to see that some tests did run properly.

It looks like this error is thrown in all 20 cases by this line in src/compose/ComposeMenu.js, a file not touched by my commits:

`const DocumentPicker = (await import('react-native-document-picker')).default;`

and it appears that that import has caused issues before; it's preceded by this comment:

```
    // Defer import to here, to avoid an obnoxious import-time warning
    // from this library when in the test environment.
```

Deferring the import doesn't seem to have helped much, or else it fixed the original warning and caused this new error; could it be that there was an additional workaround outside source control, like (somehow?) some more Babel configuration, that most developers have, but I don't, since I just recently cloned the project?

chrisbobbe:
Oh OK, yes, I can see now that the test results showing the Flow issues and the "object is not extensible" message do indeed show up when testing in my local development environment. I saw the 20 occurrences of "SyntaxError: Unexpected token import" (shown in the .txt doc in my first post here) and assumed that there was some Babel mis-configuration interfering with all tests, without looking closer to see that some tests did run properly.

It looks like this error is thrown in all 20 cases by this line in src/compose/ComposeMenu.js, a file not touched by my commits:

`const DocumentPicker = (await import('react-native-document-picker')).default;`

and it appears that that import has caused issues before; it's preceded by this comment:

```
    // Defer import to here, to avoid an obnoxious import-time warning
    // from this library when in the test environment.
```

Deferring the import doesn't seem to have helped much, or else it fixed the original warning and caused this new error; could it be that there was an additional workaround outside source control, like (somehow?) some more Babel configuration, that most developers have, but I don't, since I just recently cloned the project?

gnprice:
> I saw the 20 occurrences of "SyntaxError: Unexpected token import" (shown in the .txt doc in my first post here) and assumed that there was some Babel mis-configuration interfering with all tests, without looking closer to see that some tests did run properly.

Ah, apologies, I looked right past that yesterday.

I believe this means you're using Node 8, and the fix is to [use Node 10](https://github.com/zulip/zulip-mobile/blob/5c0b3a09a/docs/howto/build-run.md#main-steps) (the current LTS). The incompatibility with Node 8 was introduced last week in 338036e0c; it actually broke things in CI until 091bdc3ee a few commits later.

The error is pretty unhelpful. Because the incompatibility is only a week old, I think you may be the first person to have gotten bitten by it. We should make that easier to debug; I might try putting a Node version check right in `tools/test`, or failing that we'll add a troubleshooting entry in build-run.md .

I'd appreciate seeing what `node --version` says for you, to confirm that that's the issue.

gnprice:
> Deferring the import doesn't seem to have helped much, or else it fixed the original warning and caused this new error; could it be that there was an additional workaround outside source control, like (somehow?) some more Babel configuration, that most developers have, but I don't, since I just recently cloned the project?

And to answer these questions directly (though you might be able to deduce some of these answers from the diagnosis above):
* Yep, deferring the import fixed the original warning. That happened originally in b20dbcf02 .
* Then in 338036e0c we adopted fancier syntax for deferring the import... and that caused this new error.
* The known way to avoid the error is to be using Node 10 rather than Node 8. This is in our setup instructions but easy to miss, especially if you already have Node lying around.
* I'm actually puzzled why the Node version matters here, because I'd have thought it's Babel's job to compile the fancy syntax out. As that commit message on 338036e0c says, our Babel config appears to have the plugin enabled that would compile this out. I haven't dug into the discrepancy; when CI failed my first question was what was different between CI and my machine, and when I saw we were using an old Node version in CI and that upgrading that fixed the issue, I left it at that. There probably is something wrong in our Babel config (perhaps only for Jest?), but I expect that part doesn't differ between your machine and others'.

gnprice:
> there may have been a breaking change in a minor version of Babel or Jest that was triggered by the carat in package.json allowing silent minor version upgrades (e.g., "jest": "^24.8.0"), but that's just a guess until I can dig deeper or learn that it's a known issue.

Ah, and *this* at least shouldn't be a problem -- avoiding that kind of discrepancy is Yarn's job, with `yarn.lock`. :slightly_smiling_face: In particular these lines:
```
jest@^24.8.0:
  version "24.8.0"
  resolved "https://registry.yarnpkg.com/jest/-/jest-24.8.0.tgz#d5dff1984d0d1002196e9b7f12f75af1b2809081"
  integrity sha512-o0HM90RKFRNWmAWvlyV8i5jGZ97pFwkeVoGvPW1EtLTgJc2+jcuqcbbqcSZLE/3f2S5pt0y2ZBETuhpWNl1Reg==
```
should mean that `yarn install` (or plain `yarn`) gets exactly version 24.8.0 (and with exactly that SHA-512 hash) for that `"jest": "^24.8.0"`.

Upgrades are done explicitly with e.g. `yarn upgrade`, and are reflected in `yarn.lock` so that everyone else gets the same new versions once they rerun `yarn`.

gnprice:
> All three Flow errors remained. Despite the assertion, apparently Flow can't be reasonably sure that `message.flags` will still be defined in the three sites where `.includes()` is called -- my current level of understanding is that these sites are just "too far away" from the assertion, which I guess is defensible on Flow's part, if a little annoying.

Yeah. I think specifically the issue is that `message` is an object, it's mutable, and there's no way for Flow to know that it doesn't get mutated sometime in `caseNarrow` or before one of those callbacks gets invoked. (Maybe it could with a sufficiently smart whole-program analysis, but for scalability it doesn't do that; all inference is local, and the only things it knows from the rest of the program are what's in the types.)

The most common fix for this kind of thing is to say something like
```
const { flags } = message;
```
and then have both the check and the use refer to `flags`. Then it's clear they both see the same `flags`, even if `message` is mutated in between.

In our codebase this is mostly a (small) annoyance, because for most objects the truth is they never get mutated. In codebases where mutation is routine, and especially those with a lot of concurrency, this pattern is actually a very important one for avoiding some frustrating bugs.

gnprice:
> That did fix it for the first two sites of `.includes()` -- I guess it was satisfied that `throw new Error('...')` isn't returned, as in the previous example, it's just executed -- but apparently the third site, `message.flags.includes('wildcard_mentioned')`, is too far away from the assertion for Flow to be informed by it, and it alone caused the error (below). Is there a guiding principle that explains why it would work for the first two sites of `.includes()` but not this one, other than my vague "too far away" explanation?

Ah, yeah. It's the same principle at work as in your first version:

> I think specifically the issue is that `message` is an object, it's mutable, and there's no way for Flow to know that it doesn't get mutated sometime in `caseNarrow` or before one of those callbacks gets invoked.

Here, it's that once the first `includes` method is called, that's some arbitrary code running -- and Flow can't tell whether that code might go and mutate other things in the program, including our `message` object.

So the general fix to make a guard condition stay good is to make a local `const` binding for whatever it's inspecting. Otherwise, it'll last just until the first function call.

chrisbobbe:
Yep, that was it, thank you! `node --version` gives v8.12.0, and switching to v10.10.0 fixes the issue completely.

> The most common fix for this kind of thing is to say something like
> 
> const { flags } = message;
> and then have both the check and the use refer to flags. Then it's clear they both see the same flags, even if message is mutated in between.

Brilliant, this worked! Yes, I was referring to the possibility of mutation with "which I guess is defensible on Flow's part, if a little annoying"; I just took a step back when I saw that an assertion _is_ respected for code that _immediately_ follows it, so I wasn't sure how far away was "too far away," and how that immediacy is determined.

> Here, it's that once the first includes method is called, that's some arbitrary code running -- and Flow can't tell whether that code might go and mutate other things in the program, including our message object.

This sounds correct for what's going on here; some supporting examples to follow.

> Otherwise, it'll last just until the first function call.

And this sounds reasonable as the policy to determine where "too far away" is!

Some tweaking reveals that you _can_ introduce arbitrary code between the check and the call to `.includes`, without triggering the error, as long as it doesn't include a function call:

```
      // no Flow error
      if (message.flags) {
        const asdf = 'asdf';
        const o = { a: 123, b: 456 };
        delete o.a;
        for (let i = 0; i < 10; i++) {
          let jkl = 'jkl';
        }
        return false || message.flags.includes('wildcard_mentioned')
      } else {
```

Slightly upsettingly, it still doesn't trigger the error in the following case, presumably because there are still no function calls:

```
      // still no Flow error (oops!)
      if (message.flags) {
        delete message.flags; // <---- message.flags is gone, but still no error!
        return false || message.flags.includes('wildcard_mentioned')
      } else {
```

gnprice:
> Slightly upsettingly, it still doesn't trigger the error in the following case, presumably because there are still no function calls:
> 
> ```
>       // still no Flow error (oops!)
>       if (message.flags) {
>         delete message.flags; // <---- message.flags is gone, but still no error!
>         return false || message.flags.includes('wildcard_mentioned')
>       } else {
> ```

Hmm, indeed!

It looks like the Flow bug here is centered on `delete`. If we mutate the property to some other value rather than delete it, Flow gives an error appropriately:
```
const m = ({x: 1}: {| x: number | null |});
if (m.x !== null) {
  m.x = null;
  m.x.toFixed(); // error: Cannot call `m.x.toFixed` because property `toFixed` is missing in null
}
```

Looks like facebook/flow#6567 probably covers this, and facebook/flow#7089 seems closely related.

chrisbobbe:
> Upgrades are done explicitly with e.g. yarn upgrade, and are reflected in yarn.lock so that everyone else gets the same new versions once they rerun yarn.

Nice; I had a vague feeling the lock file could also be updated with a `yarn install` (or perhaps a `yarn install` following the initial install), but I suppose not, which is nice.

I've made some changes to address your feedback; the plus sign (+) tip at https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#pushing-commits-after-tidying-them really helped me out when I thought I was stuck, with no way to avoid introducing a merge commit! :)

gnprice:
And merged. Thanks again @chrisbobbe for the thorough fix and for the revisions!

Seeing the `'add' | 'remove'` assert in your last commit, I added a small commit on top that puts a static type-level assert there as well (which isn't a substitute here for the runtime one); take a look.

gnprice:
> We should make that easier to debug; I might try putting a Node version check right in `tools/test`, or failing that we'll add a troubleshooting entry in build-run.md .

Did both of these! Just pushed commits 9f82dfb46 and a5a064f7e.

PR #3621, opened 2019-09-19
search: prevent race conditions between subsequent queries

### LINKED ISSUE #3058: Search can race, showing results for a previous query
image

As user entered `h` and then `i`. 
After entering `h` app releases a fetch request with search param as `h`. Before it's response is received another request is fired with search param as `hi`. Unfortunately this second request response arrived first, messages in the state are updated and shown to the user. Now result of first request arrives, which as no messages. Thus again state is updated with `[]` as new messages and **no result** text is shown. Which also resulted in sudden splash of message list.

(Debugging logs are shown in above image)

@zulipbot claim

ray-kraesig:
Enforce an ordering on resolution of asynchronous search requests by discarding any delayed, out-of-order responses.

Fixes #3058.

### DIFF
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -29,7 +29,11 @@ export default class SearchMessagesCard extends PureComponent<Props> {
     const { isFetching, messages } = this.props;
 
     if (isFetching) {
-      return <LoadingIndicator size={40} />;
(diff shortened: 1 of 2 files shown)


ray-kraesig:
(Temporarily withdrawn for further testing.)

PR #3623, opened 2019-09-19
search: prevent race conditions between subsequent queries

### LINKED ISSUE #3058: Search can race, showing results for a previous query

ray-kraesig:
Enforce an ordering on resolution of asynchronous search requests by discarding any delayed, out-of-order responses.

Fixes #3058.

### DIFF
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -2,19 +2,14 @@
 
 import React, { PureComponent } from 'react';
 import { StyleSheet, View } from 'react-native';
-import throttle from 'lodash.throttle';
 import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 
-import type { Auth, Message, Dispatch } from '../types';
-import { connect } from '../react-redux';
+import type { Message } from '../types';
 import { LoadingIndicator, SearchEmptyState } from '../common';
-import { HOME_NARROW, SEARCH_NARROW } from '../utils/narrow';
+import { HOME_NARROW } from '../utils/narrow';
 import MessageList from '../webview/MessageList';
-import * as api from '../api';
 import renderMessages from '../message/renderMessages';
 import { NULL_ARRAY } from '../nullObjects';
-import { getAuth } from '../selectors';
-import { LAST_MESSAGE_ANCHOR } from '../constants';
 
 const styles = StyleSheet.create({
   results: {
@@ -23,60 +18,23 @@ const styles = StyleSheet.create({
 });
 
 type Props = {|
-  dispatch: Dispatch,
-  auth: Auth,
-  query: string,
-|};
-
-type State = {|
+  queryIsEmpty: boolean,
   messages: Message[],
   isFetching: boolean,
 |};
 
-class SearchMessagesCard extends PureComponent<Props, State> {
-  state = {
-    messages: [],
-    isFetching: false,
-  };
-
-  handleQueryChange = async (query: string) => {
-    const { auth } = this.props;
-
-    throttle(async () => {
-      this.setState({ isFetching: true });
-      const { messages } = await api.getMessages(
-        auth,
-        SEARCH_NARROW(query),
-        LAST_MESSAGE_ANCHOR,
-        20,
-        0,
-        false,
-      );
-      this.setState({
-        messages,
-        isFetching: false,
-      });
-    }, 500).call(this);
-  };
-
-  componentDidUpdate(prevProps: Props) {
-    if (prevProps.query !== this.props.query) {
-      this.handleQueryChange(this.props.query);
-    }
-  }
-
+export default class SearchMessagesCard extends PureComponent<Props> {
   static NOT_FETCHING = { older: false, newer: false };
 
   render() {
-    const { isFetching, messages } = this.state;
-    const { query } = this.props;
+    const { queryIsEmpty, isFetching, messages } = this.props;
 
     if (isFetching) {
       return <LoadingIndicator size={40} />;
     }
 
     if (messages.length === 0) {
-      return query.length > 0 ? <SearchEmptyState text="No results" /> : null;
+      return !queryIsEmpty ? <SearchEmptyState text="No results" /> : null;
     }
 
     const renderedMessages = renderMessages(messages, []);
@@ -98,7 +56,3 @@ class SearchMessagesCard extends PureComponent<Props, State> {
     );
   }
 }
-
-export default connect(state => ({
-  auth: getAuth(state),
-}))(SearchMessagesCard);
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -18,8 +18,7 @@ const styles = StyleSheet.create({
 });
 
 type Props = {|
-  queryIsEmpty: boolean,
-  messages: Message[],
+  messages: Message[] | null,
   isFetching: boolean,
 |};
 
@@ -27,14 +26,18 @@ export default class SearchMessagesCard extends PureComponent<Props> {
   static NOT_FETCHING = { older: false, newer: false };
 
   render() {
-    const { queryIsEmpty, isFetching, messages } = this.props;
+    const { isFetching, messages } = this.props;
 
     if (isFetching) {
       return <LoadingIndicator size={40} />;
     }
 
+    if (!messages) {
+      return null;
+    }
+
     if (messages.length === 0) {
-      return !queryIsEmpty ? <SearchEmptyState text="No results" /> : null;
+      return <SearchEmptyState text="No results" />;
     }
 
     const renderedMessages = renderMessages(messages, []);
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -29,7 +29,11 @@ export default class SearchMessagesCard extends PureComponent<Props> {
     const { isFetching, messages } = this.props;
 
     if (isFetching) {
-      return <LoadingIndicator size={40} />;
+      // Display loading indicator only if there are no messages to
+      // display from a previous search.
+      if (!messages || messages.length === 0) {
+        return <LoadingIndicator size={40} />;
+      }
     }
 
     if (!messages) {
--- a/src/search/SearchMessagesScreen.js
+++ b/src/search/SearchMessagesScreen.js
@@ -17,15 +17,13 @@ type Props = {|
 |};
 
 type State = {|
-  query: string,
-  messages: Message[],
+  messages: Message[] | null,
   isFetching: boolean,
 |};
 
 class SearchMessagesScreen extends PureComponent<Props, State> {
   state = {
-    query: '',
-    messages: [],
+    messages: null,
     isFetching: false,
   };
 
@@ -47,20 +45,19 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
   };
 
   handleQueryChange = (query: string) => {
-    this.setState({ query });
-    this.performQuery(query);
+    if (query !== '') {
+      this.performQuery(query);
+    } else {
+      this.setState({ messages: null, isFetching: false });
+    }
   };
 
   render() {
-    const { query, messages, isFetching } = this.state;
+    const { messages, isFetching } = this.state;
 
     return (
       <Screen search autoFocus searchBarOnChange={this.handleQueryChange} style={styles.flexed}>
-        <SearchMessagesCard
-          queryIsEmpty={query === ''}
-          messages={messages}
-          isFetching={isFetching}
-        />
+        <SearchMessagesCard messages={messages} isFetching={isFetching} />
       </Screen>
     );
   }
--- a/src/search/SearchMessagesScreen.js
+++ b/src/search/SearchMessagesScreen.js
@@ -1,30 +1,71 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 
+import * as api from '../api';
+import type { Auth, Dispatch, Message } from '../types';
 import { Screen } from '../common';
 import SearchMessagesCard from './SearchMessagesCard';
 import styles from '../styles';
+import { SEARCH_NARROW } from '../utils/narrow';
+import { LAST_MESSAGE_ANCHOR } from '../constants';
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks @ray-kraesig !

Quick preliminary thoughts:

#### 268848ddc search: refactor search screen to be more principled

I like this strategy. Have not yet read this commit in detail.


#### 22e7c3d0a SearchMessagesCard: remove `queryIsEmpty` prop

Commit message says:

> Clear `messages` appropriately when needed.

but I think it currently doesn't? I imagine that'd be a `setState` in `handleQueryChange`.

... Alternatively, maybe it vacuously does because we never want to clear `messages`? There's probably a simpler way to say that in the commit message, though. :stuck_out_tongue: 

... though even in that interpretation, we should initialize it to `null` now, right?

Terminology nit: I'm accustomed to reserving "prop" for the React concept by that name, and saying "property" for the general JS concept. So there's no `query` "prop" -- that confused me for a bit when reading the code after the commit message. Could call it a "state property".


#### 75b126015 search: enforce ordering on network requests

    This commit introduces non-React/Redux state in unmanaged fields of
    the search card.  This should probably be folded into the Redux store
    at some point, but should be fine for now.

I think they won't belong in Redux. The Redux store is basically a way of organizing state that belongs to a near-root component in the tree. This state belongs to the search screen; e.g. we'd want it to all disappear if you navigated back to the main screen, and be created from scratch if you returned.

Probably should be folded into React state, though.

```
+    } else {
+      // The empty query can be resolved without a network call.
+      messages = null;
+    }
```

How about pushing this inside `performQueryRaw`, as an early return there?

I think that would simplify reasoning about the different control paths here. It took me a few minutes to convince myself that everything behaves correctly on this path in the current version, and I feel like that would be a lot simpler if this conditional were contained inside the function call. It's also where this fact about the empty query sort of logically belongs.

The cost would be a trip up to the event loop and back. But that's quite cheap compared to an actual network request.


#### 836762b69 search: change isFetching semantics and behavior

```
     this.lastIdReceived = id;
 
     // A query is concluded. Report the message-list.
-    this.setState({ messages, isFetching: false });
+    this.setState({ messages, isFetching: this.lastIdReceived !== id });
```

This will always be false, won't it? Seeing as we just updated `this.lastIdReceived`.



#### 77fa78a9c search: throttle search requests

```
 
+import throttle from 'lodash.throttle';
```

Nit: I've been keeping imports from external libraries first, then imports from our own code as a separate stanza.

Hmm, doesn't this need an interval length? If this works as is, I guess there's some default... but a default in such a generic library doesn't make a lot of sense, so I'd rather have an explicit value.

ray-kraesig:
> #### [268848d](https://github.com/zulip/zulip-mobile/commit/268848ddc0c70bd36b8117b870bd07a9421823a7) search: refactor search screen to be more principled
> I like this strategy. Have not yet read this commit in detail.

Considering that every other commit in this series has had at least one notable issue, you probably should. :disappointed:

----

> #### [22e7c3d](https://github.com/zulip/zulip-mobile/commit/22e7c3d0aa10ac6f41c6b2cbad08243d80fdb00e) SearchMessagesCard: remove `queryIsEmpty` prop
> Commit message says:
>
> > Clear `messages` appropriately when needed.
> 
> but I think it currently doesn't? I imagine that'd be a `setState` in `handleQueryChange`.
>  [...]
> ... though even in that interpretation, we should initialize it to `null` now, right?

I appear to have lost some of my changes in the commit-factorization process; both of those things should indeed be present. Fixed.

> Terminology nit: I'm accustomed to reserving "prop" for the React concept by that name, and saying "property" for the general JS concept. [...] Could call it a "state property".

It just used to be a prop in the previous version, and I failed to invalidate that bit of cache in my head. I'd have called it a "field" if I'd been thinking about it. Fixed (though open to other names).

----

> #### [75b1260](https://github.com/zulip/zulip-mobile/commit/75b12601590d178d5581df660a019a16b157e338) search: enforce ordering on network requests
> ```
> This commit introduces non-React/Redux state in unmanaged fields of
> the search card.  This should probably be folded into the Redux store
> at some point, but should be fine for now.
> ```
> 
> I think they won't belong in Redux. [...] 

Removed. Thought I'd _already_ removed, in fact – the comments in the file itself are certainly from an old version.

> ```
> +    } else {
> +      // The empty query can be resolved without a network call.
> +      messages = null;
> +    }
> ```
> 
> How about pushing this inside `performQueryRaw`, as an early return there?
> 
> I think that would simplify reasoning about the different control paths here. It took me a few minutes to convince myself that everything behaves correctly on this path in the current version, and I feel like that would be a lot simpler if this conditional were contained inside the function call. It's also where this fact about the empty query sort of logically belongs.

That makes `performQueryRaw` itself somewhat harder to reason about – it's intended to be just the network request. Should `network` or `api` should be in the name?

There's actually a bug here, though it's not in evidence in this single commit: empty-string "queries" shouldn't be throttled, as `performQuery` itself will be. Given that, I've moved the logic to `handleQueryChange`, which may also cut out an additional event-loop cycle (depending on exactly how the entry to an `async` function is implemented).

Alternatively, it could be left in place (would comments help?) and only `performQueryRaw` could be throttled; this would have been my first preference, as it keeps all `lastId*` and `isFetching` manipulation localized to a single function.

(My _zeroth_ preference was to make that section of code a `do`-expression ([TC39](https://github.com/tc39/proposal-do-expressions), [Babel](https://babeljs.io/docs/en/babel-plugin-syntax-do-expressions/)) — `const messages = do { ... }` — but on review, I doubt that syntax is ever getting out of phase 1.)

----

> #### [836762b](https://github.com/zulip/zulip-mobile/commit/836762b69d30b5c4b7972820bc3818748035ada5) search: change isFetching semantics and behavior
>
> This will always be false, won't it? Seeing as we just updated `this.lastIdReceived`.

Should be `this.lastIdSent !== id`, of course – or formally equivalently but semantically more clearly, `this.lastIdSent === this.lastIdReceived`. Fixed.

----

> #### [77fa78a](https://github.com/zulip/zulip-mobile/commit/77fa78a9c362df53c1cd716482cdd0249ecc3583) search: throttle search requests
>
> Hmm, doesn't this need an interval length?

Yes; that should read `}, 500);`. Fixed (again).

src/search/SearchMessagesCard.js
```diff
@@ -18,23 +18,26 @@ const styles = StyleSheet.create({
...
 
     if (isFetching) {
       return <LoadingIndicator size={40} />;
     }
 
+    if (messages === null) {
+      return null;
+    }
+
     if (messages.length === 0) {
```
gnprice: Small style bit: perhaps make these three conditionals an "if / else if / else if" chain? Saves a few lines of vertical space, and semantically they're quite closely related so I think it's helpful in that respect too.

In the later commit that adds a condition inside `isFetching`, that'll mean the extra condition needs to be an `&&` conjunct instead of a nested `if`; I think that's a (nit-scale) improvement too.

src/search/SearchMessagesScreen.js
```diff
@@ -13,10 +13,16 @@ import { getAuth } from '../account/accountsSelectors';
...
+  // Warning: do not add new props without considering their effect on the
+  // behavior of this component's non-React internal state. See comment below.
 |};
 
 type State = {|
+  /** The list of messages returned for the latest query, or `null` if there is
+   *  effectively no "latest query" to have results from.
+   */
   messages: Message[] | null,
+  /** Whether there is currently an active valid network request. */
```
gnprice: This isn't true as of this commit, right? This is the meaning that a later commit gives it.

If you have this commit document it as what it means at this point (or actually even just not document it yet at all -- that'd be fine), and then change the doc comment to this in the same commit that changes the meaning, I think that will help clarify both this and the later commit.

src/search/SearchMessagesScreen.js
```diff
@@ -83,7 +84,7 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
       messages,
       isFetching: this.lastIdSent !== this.lastIdReceived,
     });
-  };
+  }, 500);
```
gnprice: Perhaps shorter, like 200ms or 250ms? I imagine 500ms may feel pretty sluggish.

Easy to tweak later, though.

src/search/SearchMessagesCard.js
```diff
@@ -18,23 +18,26 @@ const styles = StyleSheet.create({
...
 
   render() {
-    const { queryIsEmpty, isFetching, messages } = this.props;
+    const { isFetching, messages } = this.props;
 
     if (isFetching) {
       return <LoadingIndicator size={40} />;
     }
 
+    if (messages === null) {
```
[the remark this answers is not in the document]
ray-kraesig: So it does! And so I shall. (I've been bitten by the Python semantics before.)

src/search/SearchMessagesScreen.js
ray-kraesig: > This isn't true as of this commit, right? This is the meaning that a later commit gives it.

That's correct — this was an error in the commit refactoring.

src/search/SearchMessagesScreen.js
ray-kraesig: Possibly. Average smartphone typing speed [seems to be ≤30 wpm](https://www.quora.com/What-is-the-average-smartphone-users-typing-speed), or [approximately](https://www.yorku.ca/mack/RN-TextEntrySpeed.html) 250ms/character, though; if that's accurate, a 250ms throttle sounds like it'd be functionally useless.

I *have* observed sluggishness, but it seems to have been due to `TextInput` itself delaying the messages. (Seen in simulation/emulation only; I haven't timed it on real hardware yet.)

src/search/SearchMessagesScreen.js
```diff
@@ -1,30 +1,70 @@
...
 import styles from '../styles';
+import { SEARCH_NARROW } from '../utils/narrow';
+import { LAST_MESSAGE_ANCHOR } from '../constants';
+import { connect } from '../react-redux';
+import { getAuth } from '../account/accountsSelectors';
 
-type Props = {||};
+type Props = {|
+  auth: Auth,
+|};
```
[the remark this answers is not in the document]
ray-kraesig: That... doesn't seem to be the problem that the `$FlowFixMe` is (currently?) suppressing? Flow produces identical error output with or without `dispatch: Dispatch,` after that `$FlowFixMe` is removed:

```
 • in property search.screen:
    • Either property navigationOptions is missing in React.ComponentType [1] but exists in
      withOptionalNavigationOptions [2].
    • Or React.ComponentType [1] is incompatible with React.StatelessFunctionalComponent [3].
```

I've added it anyway, though; `connect` [is documented as supplying it](https://react-redux.js.org/using-react-redux/connect-mapdispatch), and probably _should_ be complaining about it already at its call site in this file.

src/search/SearchMessagesScreen.js
```diff
@@ -79,7 +79,10 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
     this.lastIdReceived = id;
 
     // A query is concluded. Report the message-list.
-    this.setState({ messages, isFetching: false });
+    this.setState({
+      messages,
+      isFetching: this.lastIdSent !== this.lastIdReceived,
```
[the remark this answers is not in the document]
ray-kraesig: I don't agree at all! I find that much more confusing – it sends me into a mental trainwreck of "okay, wait, is that _actually_ the right way around?" and "so is `lastIdSent - lastIdReceived < 0` actually a supported case?"

The English text you've provided doesn't seem to match up particularly well with either operator; it certainly wasn't the way I was thinking about it when I wrote it. (Which, if I had to put it into English, would be something more like "this device is at equilibrium exactly when...".)

ray-kraesig:
An additional commit to add some missing error-handling is forthcoming.

src/search/SearchMessagesScreen.js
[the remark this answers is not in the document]
gnprice: Yeah, those nav-related errors are what the fixme is aimed at suppressing. I think the failure to give an error here isn't so much because of the fixme as because the `react-navigation` types are no good (i.e. for broadly the same reason the fixme is there.)

In particular I think if the `connect` result were actually used to instantiate a component, then Flow would notice the missing `dispatch`. Ditto if it were wired through some well-typed higher-order components and ultimately used to instantiate a component.

... Then again if I try it on a random other component (`UnreadCards`, the first I saw in a quick grep), deleting either `dispatch` or another `connect`-supplied property from the `Props` type gives an error right at the `connect` call site. ... Oho, but if I delete the one call site of that component -- even though that's off in a different file -- the error disappears. So yeah, that seems to be what's happening.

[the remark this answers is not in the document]
gnprice: The commit 776df74ff "search: enforce ordering on network requests" now deletes this line; I'm guessing that's a rebase error.

src/search/SearchMessagesScreen.js
gnprice: > Average smartphone typing speed [seems to be ≤30 wpm](https://www.quora.com/What-is-the-average-smartphone-users-typing-speed), or [approximately](https://www.yorku.ca/mack/RN-TextEntrySpeed.html) 250ms/character, though; if that's accurate, a 250ms throttle sounds like it'd be functionally useless.

Hmm. That speed does sound pretty plausible.

Still that doesn't mean it's the wrong value. There's always the possibility that whoever originally decided to stick a `throttle` call in here wasn't quite thinking it through. :wink:

gnprice:
Merged! :checkered_flag: 

As we discussed in person just now, I dropped the commit that adds throttling, as the UX of it seems materially slower at least some of the time with that. And as you observed, the putative throttling in the status quo wasn't actually doing anything :upside_down_face: so that's not going to be any regression in the bursts of search requests we might send to servers.

PR #3625, opened 2019-09-21
docs/dev server: Add explicit vagrant restart command

ray-kraesig:
A naïve `vagrant halt` + `vagrant up [--provision]` doesn't suffice.


gnprice:
Thanks! Merged.

PR #3626, opened 2019-09-24
Small search-screen followups 

gnprice:
This makes a few small changes following up on #3623 . The main one is:

    search: Move request-ID allocation out of inner helper.
    
    This is a pure refactor.  But if handleQueryChangeInner were wrapped
    in a throttle or debounce in the most obvious way, there'd be a bug
    which this fixes.  The possibility of doing that is the main reason
    this function is separate, so seems best to clean that up while it's
    fresh in our minds.


ray-kraesig:
Those last two commits look like they belong somewhere else, but we do need them. LGTM; committed.

gnprice:
Thanks!

> Those last two commits look like they belong somewhere else

This is related to our preference for reading (and writing) history as commits rather than as PRs. :-)

These could have been three independent PRs. But there's a fair amount of overhead in making, and merging, lots of little PRs, compared to the work in Git of making the commits. So when a given session of work looking at a given piece of code results in some loosely-related small changes in addition to the main ones, it's typical to include them in the same PR.

Then as long as they're in nicely separate commits, they remain nicely separated in the history as seen from the future.

In this case, the commit cutting lodash.throttle came naturally as I was looking at this code. Then when I ran `yarn update-libdefs` to sync the libdefs, I saw it removed the react-native-vector-icons stub too, so that produced the other commit.

PR #3627, opened 2019-09-25
android: remove manually-added font assets

ray-kraesig:
Remove from the Android source directory some font assets that, as of
cd899bbe0, should be added to the APK automatically by a Gradle
script.


gnprice:
Thanks! Merged.

I expanded the commit message somewhat, most of all to explicitly make the connection with the particular library these all come from.

PR #3628, opened 2019-09-26
Add Android NDK installation instructions

ray-kraesig:
Its absence is precisely diagnosed by relevant error messages and not difficult to fix, but we should still note it somewhere.


gnprice:
Looks good, merged -- thanks!

Issue #3590, opened 2019-08-15
Investigate whether mobile app doesn't send an immediate presence update on open


timabbott:
Based on the report in
https://chat.zulip.org/#narrow/stream/9-issues/topic/messages.20before.20status.20change/near/779759:

> I've vaguely noticed this issue several times now, and it just happened again. Sometimes I'll receive a message from someone (either in a stream or as a PM) or see that they're typing a PM while they still appear as offline; then, several seconds later (mostly ~5-20 seconds if I'm recalling correctly), they'll change to online. Everyone but me pretty much uses the mobile app exclusively, whereas I mostly use the desktop app, which is where I notice this. Does anyone know what might be going on?

, we should investigate whether the mobile app needs to be doing something new to send an immediate presence update to the server when a user opens the app and starts using it.  It's possible something else is going on (or this is an artifact of networking perf), but it seems worth investigating because we do have special code for preventing this failure mode in the webapp.

cyphase:
I was the reporter, if any questions come up. I'll try to pay a bit closer attention to when this happens in case that helps.

cyphase:
This just happened again (I'm not going to report every occurrence, but figured I'd provide at least one specific instance). My eyes caught a flicker as a reaction was added to a message, and within about 2-3 seconds I started counting, for about 5 seconds, then the user's status changed to online.

I have a simple daemon running (connected as me via API) to relay select real-time events to a notification service. It echos all events, so I can see that there as well, the `reaction` event came before the `presence` event; in between was an `update_message_flags` event that occurred when I focused the desktop client and a message was marked as read. Unfortunately the `reaction` event doesn't include a timestamp, and my code isn't printing it's own, but I don't think it would provide much additional information in this case.

gnprice:
Vishwesh has a diagnosis of the issue, in a commit message on #3629 : what's happening is that
* we only send presence updates once a minute,
* so we have a variable that tracks the last time we did so, and skip sending it if it's been less than a minute,
* ... but at app start, we initialize that value to the current time.
* And that initialization is at import time, before we run any of the code that would actually send a presence update.
* So we'll never send a presence update in the first 60 seconds after the app starts.

The fix, then, is that that initial value should never have been the current time; it should be a "zero" value in the distant past, say the Unix epoch. (If indeed that logic should be there at all; it's not clear it should.)

The bug appears to have been there since presence updates were first implemented for the app, in 
[...]

PR #3629, opened 2019-09-28
Fix immediate presence was not updated on app open 

jainkuniya:
Fixes: #3590

### DIFF
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -219,6 +219,9 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState) 
  * we want to do when starting up, or regaining a network connection.
  */
 export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
+  dispatch(reportPresence());
+  setInterval(() => dispatch(reportPresence()), 60 * 1000);
+
   dispatch(initialFetchStart());
   const auth = getAuth(getState());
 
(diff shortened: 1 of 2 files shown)


src/message/fetchActions.js
```diff
@@ -222,6 +222,9 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
   dispatch(initialFetchStart());
   const auth = getAuth(getState());
 
+  dispatch(reportPresence());
+  setInterval(() => dispatch(reportPresence()), 60 * 1000);
```
gnprice: These should go above the `initialFetchStart` and `const auth` lines. Those are closely tied to the `tryUntilSuccessful ... api.registerForEvents` call.

src/users/usersActions.js
```diff
@@ -7,8 +7,8 @@ import { PRESENCE_RESPONSE } from '../actionConstants';
 import { getAuth, tryGetAuth } from '../selectors';
 import { isPrivateOrGroupNarrow } from '../utils/narrow';
 
-let lastReportPresence = new Date();
-let lastTypingStart = new Date();
+let lastReportPresence = 0;
+let lastTypingStart = 0;
```
gnprice: Hmm. This makes these variables inconsistently-typed: they're a number at first, and later a Date.

Let's make them consistently one or the other. That way I (and the next reader) don't have to wonder what semantics we're giving to a date subtraction between a `Date` and a number. :-)

Based on MDN's doc:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Syntax
I think if you say `new Date(0)`, that will give a `Date` object referring to the Unix epoch aka the first moment of 1970 in UTC. That's the thing that I *hope* a plain zero (as in this version) ends up working out to as well.

gnprice:
Thanks @jainkuniya ! Merged.

I added a couple of commits on top making further improvements to this code; take a look.

PR #3630, opened 2019-09-28
typingEvent: Report stop event to server on sending message.

jainkuniya:
So that other user doesn't think that user is typing another message.

Set `lastTypingStart` as 0, when typing operation is stop, so that
immediate start typing event can be send to the server, if required.

### DIFF
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -34,7 +34,7 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
   });
 };
 
-export const sendTypingEvent = (narrow: Narrow) => async (
+export const sendTypingStart = (narrow: Narrow) => async (
   dispatch: Dispatch,
   getState: GetState,
 ) => {
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -53,3 +53,16 @@ export const sendTypingStart = (narrow: Narrow) => async (
   const auth = getAuth(getState());
   api.typing(auth, narrow[0].operand, 'start');
 };
+
+export const sendTypingStop = (narrow: Narrow) => async (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
+  if (!isPrivateOrGroupNarrow(narrow)) {
+    return;
+  }
+
+  const auth = getAuth(getState());
+  api.typing(auth, narrow[0].operand, 'stop');
+  lastTypingStart = new Date(0);
+};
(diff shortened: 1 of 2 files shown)


src/api/modelTypes.js
```diff
@@ -456,3 +456,5 @@ export type Message = $ReadOnly<{
   timestamp: number,
   type: 'stream' | 'private',
 }>;
+
+export type TypingOperation = 'start' | 'stop';
```
gnprice: This doesn't really belong at this spot.

Let's put it in `eventTypes.js`. Give it the name `TypingEventOp`, and put it after `StreamEvent` -- when we later add `TypingEvent` to that file, we'll put it at that spot alphabetically.

src/users/usersActions.js
```diff
@@ -33,17 +33,19 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
   });
 };
 
-export const sendTypingEvent = (narrow: Narrow) => async (
+export const sendTypingEvent = (narrow: Narrow, opertation: TypingOperation) => async (
```
gnprice: spelling: "operation"

src/users/usersActions.js
```diff
@@ -33,17 +33,19 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
...
 
-  if (differenceInSeconds(new Date(), lastTypingStart) > 15) {
+  const isStopEvent: boolean = opertation === 'stop';
+
+  if (differenceInSeconds(new Date(), lastTypingStart) > 15 || isStopEvent) {
     const auth = getAuth(getState());
-    api.typing(auth, narrow[0].operand, 'start');
-    lastTypingStart = new Date();
+    api.typing(auth, narrow[0].operand, opertation);
+    lastTypingStart = isStopEvent ? 0 : new Date();
```
gnprice: Hmm. Seeing the ways this control flow comes out differently for 'stop' vs. 'start', I think it'd make most sense to make this two different functions rather than one function with an argument.

Say, `sendTypingStart` and `sendTypingStop`.

src/api/modelTypes.js
jainkuniya: I have updated this PR, created new method `sendTypingStop`. I don't think in this new version we want to move out type from api, let it be present where it is actually used (used only once). No unnecessary re-directions.

gnprice:
Thanks for the updates! Looks good -- merged.

PR #3633, opened 2019-09-30
Migrate logging to new functions `logging.error` and `logging.warn`.

gnprice:
See commit messages for details.

I was moved to go and do this by some other work where I found I wanted to log warnings to Sentry, and was dissatisfied with the existing API.


ray-kraesig:
As can perhaps be deduced from the above: looks good!

PR #3617, opened 2019-09-17
icons: update react-native-vector-icons

ray-kraesig:
Update react-native-vector-icons to v6.6.0.

This updates Feather, in particular, to include the `terminal` icon which is used on the "Log in with dev account" option.

Fixes #2209. (See also PR #2892. This fixes the issue seen there by adding a call to a gradle file provided by `react-native-vector-icons`, as suggested in [their README](https://github.com/oblador/react-native-vector-icons/blob/07b156bb513dda45ca7468a92b4ffc04b072a588/README.md).)

PR #3635, opened 2019-10-03
Icons: Extend type-checking to the icon names, and small other cleanups.

gnprice:
Before this change, if we write a `name` prop that mistypes an icon's
name, or refers to an icon that for any reason doesn't exist, Flow
won't give an error.  Now it will.

Was looking at this a couple of weeks ago in reviewing #3617 and experimenting with the code. Most of these changes are a set of followups I'd drafted then, and just cleaned up now.

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -41,12 +41,12 @@ export type IconNames = FeatherGlyphs;
 export const Icon = fixIconType<IconNames>(Feather);
 
 /** A (type for a) component-type like `Icon` but with `name` already specified. */
-export type IconType = ComponentType<$Diff<IconProps<empty>, {| name: mixed |}>>;
+export type SpecificIconType = ComponentType<$Diff<IconProps<empty>, {| name: mixed |}>>;
 
 const makeIcon = <Glyphs: string>(
   iconSet: ComponentType<IconPropsBusted<Glyphs>>,
   name: Glyphs,
-): IconType => props => {
+): SpecificIconType => props => {
   const FixedIcon = fixIconType<Glyphs>(iconSet);
   return <FixedIcon name={name} {...props} />;
 };
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -2,7 +2,7 @@
 import React from 'react';
 import type { ComponentType } from 'react';
 import type { Text } from 'react-native';
-import type { Color } from 'react-native-vector-icons';
+import type { Color, IconProps as IconPropsBusted } from 'react-native-vector-icons';
 import Feather from 'react-native-vector-icons/Feather';
 import type { FeatherGlyphs } from 'react-native-vector-icons/Feather';
 import IoniconsIcon from 'react-native-vector-icons/Ionicons';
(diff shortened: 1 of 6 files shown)


src/common/Icons.js
```diff
@@ -28,46 +30,62 @@ type IconProps<Glyphs: string> = {|
   color?: Color,
 |};
 
+/** A variation of `IconProps` closer to what upstream's .js.flow says. */
+type IconPropsBusted<Glyphs: string> = {| ...IconProps<Glyphs>, allowFontScaling: boolean |};
```
ray-kraesig: Why not just `import { IconProps as IconPropsBusted } from 'react-native-vector-icons'`?

Or alternatively, instead `import { Icon as IconBusted } from 'react-native-vector-icons'`, and use `$Class<IconBusted<Glyphs>>` (I think) in place of `ComponentType<IconPropsBusted<Glyphs>>` below.

src/common/Icons.js
```diff
@@ -44,12 +44,12 @@ export type IconNames = FeatherGlyphs;
 export const Icon = fixIconType<IconNames>(Feather);
 
 /** A (type for a) component-type like `Icon` but with `name` already specified. */
-export type IconType = ComponentType<$Diff<IconProps<empty>, {| name: mixed |}>>;
+export type SpecificIconType = ComponentType<$Diff<IconProps<empty>, {| name: mixed |}>>;
```
ray-kraesig: I read this as a specific (icon-type), rather than a (specific icon)'s type.

I don't think it improves legibility at any use site.

src/common/Icons.js
```diff
@@ -9,15 +9,17 @@ import IoniconsIcon from 'react-native-vector-icons/Ionicons';
...
- * https://github.com/oblador/react-native-vector-icons/blob/v6.6.0/README.md#properties
+ * This corresponds to the documented interface:
+ *   https://github.com/oblador/react-native-vector-icons/blob/v6.6.0/README.md#properties
  *
- * Unfortunately the upstream implementation of this type explicitly includes
- * `allowFontScaling?: boolean` -- a misrespecification of the React Native
- * property (which RN itself specifies to be `allowFontScaling?: ?boolean`).
+ * Upstream has a .js.flow file which is meant to describe this.  But the
+ * type written there is erroneously much narrower:
+ *  * it's missing most of the properties of `Text`;
```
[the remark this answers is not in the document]
gnprice: Oof, indeed -- I'd noticed that earlier, but then forgotten it.

That is of course a problem of its own :slightly_smiling_face: Will fix the comment to reflect that.

src/common/Icons.js
gnprice: > Why not just `import { IconProps as IconPropsBusted } from 'react-native-vector-icons'`?

Hmm, indeed that seems to work! I'll do that.

I thought I'd tried something like that and it didn't, but something must have been different.


> Or alternatively, instead `import { Icon as IconBusted } from 'react-native-vector-icons'`, and use `$Class<IconBusted<Glyphs>>` (I think) in place of `ComponentType<IconPropsBusted<Glyphs>>` below.

It feels like something on these lines ought to work too, but if I try it (with [`Class`](https://flow.org/en/docs/types/utilities/#toc-class)) I get a bunch of Flow errors like this:
```
Cannot call makeIcon with Feather bound to iconSet because Icon [1] is
incompatible with Icon [2].

     src/common/Icons.js
 [2] 54│   iconSet: Class<IconBusted<Glyphs>>,
       :
     90│ export const IconPlusSquare = makeIcon(Feather, 'plus-square');

     node_modules/react-native-vector-icons/Feather.js.flow
 [1]  9│ declare export default Class<Icon<FeatherGlyphs>>;
```

Nominal typing? Something else? Dunno.

Possibly this is the thing I tried that I remembered as also being `IconProps as IconPropsBusted`.

src/common/Icons.js
gnprice: > I read this as a specific (icon-type), rather than a (specific icon)'s type.

You've read it correctly!

This name can probably be improved. The thing that's really wrong with the existing names is the relationship between `Icon` and this `IconType`: with those two names next to each other in the same small piece of API, it sure looks like `IconType` should be the type of `Icon`, but it's not... and in fact the difference between them runs in a completely different direction from that.

One question here is: why is "Type" in the name at all? Yes this is the name of a type... but if we had a convention that types' names said "Type", we'd have a lot more names like that. I'm taking "Type" here as parallel to the relationship in React (as described with Flow) between `Component` and `ComponentType`. One of those is a class -- so already a type -- and the other is the type *of* a thing like `Component`.

So `IconType` would be a natural name for the type of `Icon` and things like it... and this isn't that, but instead is a certain variation on that. "Specific" is one attempt to gesture at that variation.

gnprice:
Thanks for the review -- updated! I believe I've taken care of the first two comments.

gnprice:
(ping)

ray-kraesig:
Sorry! This one fell through the cracks. Looking now.

chrisbobbe:
Small bump on this one, as I'm getting fresh errors from Flow 0.105.0, which will be part of the RN v0.61 upgrade. In particular, the inconsistency between `react-native-vector-icons`, with the `allowFontScaling` prop. Lots of these:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/common/Icons.js:61:46

Cannot create Feather element because null or undefined [1] is incompatible with boolean [2] in property
allowFontScaling.

     src/common/Icons.js
     58│ export const IconLanguage: IconType = props => <Feather name="globe" {...props} />;
     59│ export const IconNight: IconType = props => <Feather name="moon" {...props} />;
     60│ export const IconSettings: IconType = props => <Feather name="settings" {...props} />;
     61│ export const IconRight: IconType = props => <Feather name="chevron-right" {...props} />;
     62│ export const IconPlusCircle: IconType = props => <Feather name="plus-circle" {...props} />;
     63│ export const IconLeft: IconType = props => <Feather name="chevron-left" {...props} />;
     64│ export const IconPeople: IconType = props => <Feather name="users" {...props} />;

     node_modules/react-native-vector-icons/index.js.flow
 [2] 57│   allowFontScaling?: boolean,

     node_modules/react-native/Libraries/Text/TextProps.js
 [1] 59│   allowFontScaling?: ?boolean,
```

And it seems like this PR is intended to address that, and some further inconsistencies.

src/common/Icons.js
```diff
@@ -30,46 +30,59 @@ type IconProps<Glyphs: string> = {|
...
-export const IconFile: IconType = props => <Feather name="file" {...props} />;
-export const IconTerminal: IconType = props => <Feather name="terminal" {...props} />;
-export const IconMoreHorizontal: IconType = props => <Feather name="more-horizontal" {...props} />;
-export const IconEdit: IconType = props => <Feather name="edit" {...props} />;
-export const IconPlusSquare: IconType = props => <Feather name="plus-square" {...props} />;
+export const IconInbox = makeIcon(Feather, 'inbox');
+export const IconStar = makeIcon(Feather, 'star');
+export const IconMention = makeIcon(Feather, 'at-sign');
+export const IconSearch = makeIcon(Feather, 'search');
+export const IconDone = makeIcon(Feather, 'check');
```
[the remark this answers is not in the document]
gnprice: That'd feel pretty cluttered to me -- adding `: SpecificIconType` to each of these lines would feel like repetitive boilerplate.

src/common/Icons.js
gnprice: Yeah, renaming to `FeatherIcon` might be a good further cleanup. That would also help with the awkward mixture we have of places where `Icon` means
* sometimes a value like `IconDone` or `IconTerminal` or `IconPrivateChat`, of type `SpecificIconType`;
* other times the `Icon` export of this module, which has a different type -- it requires a `name` prop.

Precisely because it's an awkward mixture it's a little annoying to go in and fix, though. So I'm going to go ahead and merge this without blocking on that additional fixup.

gnprice:
Thanks for the ping! Merged; just had to resolve some conflicts on rebase.

Hopefully that takes care of the issue you're seeing with the new Flow, too; if not we can debug in chat.

chrisbobbe:
> Hopefully that takes care of the issue you're seeing with the new Flow, too; if not we can debug in chat.

Yep, it absolutely did! Thank you!

PR #3636, opened 2019-10-04
Enable use of static locally-bundled files from WebView

ray-kraesig:
Set up a build subsystem allowing static assets to be pushed to the bundled app for use in the WebView, and enable their use from the WebView itself.

Although only static assets are currently supported, there is a place marked out in the build scripts for future build-time generation of these assets.

Note that the last two commits in this PR are not really part of the above; they're instead an example and a test, transforming some of the existing dynamically-injected-but-otherwise-static CSS into a static file.


ray-kraesig:
Temporarily withdrawn while I track down an unrewritten resource. (When a message is starred or unstarred, the resulting reload tries to load a remote resource locally.)

ray-kraesig:
... and upon investigation, that seems to be just the new effect of an existing attempt to load untransformed URLs, rather than a completely new bug. Reactivating!

android/app/webviewAssets.gradle
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+gradle.projectsEvaluated {
+    // The root of our git repository.
+    def repoDir = project.file('../..').absolutePath
+
+    android.applicationVariants.all { variant ->
+        // Accurate as of Android Gradle Plugin 3.4.0. Known to move between
+        // releases; but alas, no stable API is known to the Internet at large.
+        def assetsDir = "${buildDir}/intermediates/merged_assets/${variant.name}/out"
```
gnprice: Alas indeed.

I think a helpful breadcrumb to add here, too, would be something like `See commits 8f8a266af and 833d083a9 for previous changes.`

android/app/webviewAssets.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+        def variantTask = tasks.create(
+            name: "buildStaticWebviewAssetsFor${variant.name.capitalize()}",
```
gnprice: I think the naming style used by upstream would be like `"build${variant.name.capitalize()}StaticWebviewAssets"`. Like `bundleDebugJsAndAssets`, or `compileReleaseSources`, to take a couple of examples from `./gradlew tasks`.

src/webview/static/README.md
```diff
@@ -0,0 +1,15 @@
+# Webview static files
+
+This directory holds static files to be packaged for use by the embedded
+WebView. Files herein (other than those named `README.md`) will be copied out to
+the packaged bundle at build time, into a subdirectory named `webview`.
+
+_N.B._: When editing and testing, changes made to this directory will **not**
+be uploaded to an emulator after a mere reload! You must perform a full rebuild
+(`react-native run-${TARGET}`) to repackage any new or altered assets.
```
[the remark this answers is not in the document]
ray-kraesig: As I remember it, not terribly long — on the order of 15s, as opposed to 5s. There's no actual recompilation of other components being done, just a very quick scan of the timestamps of their penultimate dependencies (which will normally all be up-to-date). Memory probably is not the best thing to go on for that, though; I'll double-check.

It may be a moot point, though: even for dynamic elements, I find that I've been doing most of my development of this sort by fiddling around in the WebView inspector, and only "freezing" it into code or static assets when I'm nearly done. ¯\_(ツ)_/¯

src/webview/static/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Although these files are static at runtime, they may still be generated
+dynamically at _build_ time. See [the build-webview script][build-webview] for
+an appropriate hook.
```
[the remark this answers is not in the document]
ray-kraesig: No, it actually is these files (meaning "files in webview/static"). I'm expecting the sources for any generated files that will be built into this directory will live elsewhere (_e.g._, js.js stays where it is.)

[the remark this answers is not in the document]
gnprice: Ah, got it. So IIUC the plan, then:
* this directory has some of the sources for what goes in the built webview-static directory (namely, those sources that are just copied verbatim)
* other sources may be elsewhere, like src/webview/js/
* this directory will also be the place to stage the built artifacts from those various other sources -- or rather, all the build products we can build in a platform-independent way

I think the main reason I found this paragraph confusing is that I read it as talking about what's in the tree in the Git repository -- it actually didn't occur to me that "these files" was meant to include files that might be in this directory in the worktree, but aren't in Git. In general I guess I think of the tree that's in Git as the primary thing we act on and talk about. :-)

That confusion can be cleared up with just some more words to make that explicit, maybe like "Although the files in this directory are static at runtime, some of them may be generated dynamically at _build_ time (and not tracked in Git). See ..."

I also don't totally love that layout where this directory does double duty: I think I'd find things a bit cleaner if build products weren't intermingled in the same directory with source files, and instead were tucked away in some dedicated build directory (like the Android build does with `android/build/` and `android/app/build/`.) But the point where that would really matter is a ways off.

src/webview/css/css.js
```diff
@@ -5,488 +5,16 @@ import cssPygments from './cssPygments';
...
-#scroll-bottom svg {
-  width: 32px;
-  height: 32px;
-  fill: hsla(0, 0%, 100%, 0.75);
-}
-`;
+/** Common CSS properties. */
+const cssVars = `
+:root {
+  --brand-color: ${BRAND_COLOR}
```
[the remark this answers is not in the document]
ray-kraesig: Per MDN: Chrome 49, Safari 9.1. That puts us in the safe zone for Safari, but the danger zone for Android. I've confirmed that it does _not_ work on the stock KitKat image. It won't work on stock Lollipop, either (Chrome 37), but it will work on a vaguely-updated Lollipop.

... also, it turns out that the code to absolutize relative URLs in a previous commit needs a couple of KitKat polyfills. :sweat:

It might be better to go ahead and do this the "hard" way; it shouldn't actually be that hard. (And either way, despite the branch name, this commit and the next needn't block the rest of the PR.)

tools/build-webview
```diff
@@ -0,0 +1,137 @@
...
+    esac
+done
+
+if  [ -z "${TARGET-}" ]; then
+    usage; exit 2
+elif [ -z "${DEST-}" ]; then
+    usage; exit 2
+fi
+
+# XXX case-ify
```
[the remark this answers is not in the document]
ray-kraesig: Nope! (Removed.)

tools/build-webview
```diff
@@ -0,0 +1,137 @@
...
+# Sanity-check DEST and TARGET.
+case "$TARGET" in
+    ios)
+        if [ "$(uname)" != "Darwin" ]; then
+            err "iOS builds only supported on macOS";
+            exit 1;
```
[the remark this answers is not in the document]
ray-kraesig: Yup! (Removed.)

src/webview/css/css.js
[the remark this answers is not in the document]
gnprice: Rereading myself what we say there, I think it's acceptable as long as this works in Chrome 58, which comes on a stock Oreo image. (The WebView's browser is updated as an APK, and empirically is usually recent even when the OS is old.) And even core functionality can break on older than Chrome 44, found in stock Marshmallow.

I'll start a chat discussion about our Chrome version support, because it's complicated and I don't think we've really thought about it in a year.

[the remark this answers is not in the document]
gnprice: That discussion here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Chrome.20versions/near/790982

src/webview/static/README.md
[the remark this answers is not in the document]
gnprice: Looks good!

> As a note for that distant future:

FWIW my experience is that finding a remark I remember making in a GitHub comment can be frustratingly laborious. So I recommend putting any note you want to find in the distant future into either a comment, a commit message, or a chat thread :wink:.

(One reason for the stream `#mobile-team`'s existence is to be a suitable place for such notes, no matter how technically in-the-weeds. This particular note, I'd probably most recommend sticking either in a comment by the `rsync` command in the script, or in this README.)

android/app/webviewAssets.gradle
gnprice: Good thought -- thanks!

src/webview/static/README.md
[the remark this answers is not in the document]
gnprice: > even for dynamic elements, I find that I've been doing most of my development of this sort by fiddling around in the WebView inspector, and only "freezing" it into code or static assets when I'm nearly done. ¯\_(ツ)_/¯

True enough! I generally do the same thing. That's a much faster cycle still, which is great.

I think in my experience when I'm making webview changes there's typically at least a couple of cycles of actually putting the changes into the tree and wanting to refresh from that. It could be just one if I consistently did that translation correctly (or zero if it were so consistent I didn't feel a need to check), but it's fiddly enough that I think most often I don't. So if doing that took like a minute, or several minutes, that'd be bad.

15s seems OK, though. Thanks for checking :slightly_smiling_face: 

Maybe s/full //? To me "full rebuild" suggests specifically a build that's *not* using any cached intermediate build products.

src/webview/static/README.md
```diff
@@ -0,0 +1,15 @@
+# Webview static files
+
+This directory holds static files to be packaged for use by the embedded
+WebView. Files herein (other than those named `README.md`) will be copied out to
+the packaged bundle at build time, into a subdirectory named `webview`.
```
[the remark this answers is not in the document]
gnprice: New version looks good, thanks!

The other information in your reply is interesting and sounds useful, and I was going to suggest putting it in comments in appropriate places. Then I saw that the info about the iOS root seems to already be in a comment on `assetsPath` in one of your later commits, which looks like a perfect spot.

For the point about keeping things consistent: I think that means consistency between `assetsPath`; the paths passed at the call sites of `tools/build-webview`; and the validation in the latter script itself. Is that right?

Given your point about Xcode :stuck_out_tongue: , perhaps that comes down to the script's own validation, and `assetsPath` and/or other app code that relies on the paths. I think a pointer at least from the script to `assetsPath` would be helpful -- I know when I was looking at the mention of `file:///android_asset/webview` yesterday I was curious where that path came from, or what it has to be consistent with.

src/webview/js/js.js
```diff
@@ -81,6 +83,31 @@ if (!Element.prototype.closest) {
   };
 }
 
+/* Polyfill String#startsWith. Native in Mobile Safari 9, Chrome 49. */
+if (!String.prototype.startsWith) {
+  // $FlowFixMe (polyfill)
+  Object.defineProperty(String.prototype, 'startsWith', {
```
gnprice: Huh interesting. What's the reasoning for choosing between this and `String.prototype.startsWith = ...`, like you do with `includes` below?

android/app/build.gradle
```diff
@@ -76,6 +76,7 @@ import com.android.build.OutputFile
 
 project.ext.react = [
     entryFile: "index.js",
+    // quasi-unstable: see commits 8f8a266af and 833d083a9
```
gnprice: Good thought, thanks :)

src/webview/js/js.js
```diff
@@ -81,6 +83,31 @@ if (!Element.prototype.closest) {
...
+
+/* Polyfill String#includes. Native in Mobile Safari 9, Chrome 41. */
+if (!String.prototype.includes) {
+  // $FlowFixMe (polyfill)
+  String.prototype.includes = function includes(search, start = 0) {
+    if (start + search.length > this.length) {
+      return false;
+    } else {
+      return this.indexOf(search, start) !== -1;
+    }
```
gnprice: Hmm, why not just `indexOf` and test the return value?

At a quick test, I think this length-check case is inconsistent with the `includes` semantics:
```
$ node
> "abc".includes("", 100)
true
> "abc".indexOf("", 100)
3
```

src/webview/js/js.js
```diff
@@ -466,31 +493,48 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
...
+    const imageTags = parent.getElementsByTagName('img');
+    arrayFrom(imageTags).forEach(img => {
+      // Get the raw `src` value from the DOM. (We can't easily use `img.src`,
+      // since it's absolutized by the browser.)
+      //
+      /* $FlowFixMe: upstream `getNamedItem` return type should be `Attr | null` */
+      const actualSrc: string | null = img.attributes.getNamedItem('src')?.value;
+
+      // Skip elements with non-relative sources.
+      if (actualSrc && !actualSrc.startsWith('/')) {
```
gnprice: This is a bit more specific than "relative", right? Like if the value were `../foo.png` or `foo.png` or `?x=y` or `#sec-1`, or in the other direction `//chat.example.com/user_uploads/foo.png`, those would be relative and not matched here.

Basically this assumes that any relative URL reference we'll see as the `src` of an `img` is one replacing the whole path, and no more or less. That assumption sounds quite plausible, and I think it's fine to make if we're explicit about it and we confirm the server code meets it. Mostly I just want to be sure to be explicit about it. :slightly_smiling_face:

src/webview/js/js.js
```diff
@@ -466,31 +493,48 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
...
-      return;
-    }
+/**
+ * Rewrite the source URLs of all relative-URL <img> tags beneath the specified
+ * parent element to have an absolute path referencing the Zulip realm.
+ *
+ * DEPRECATED: If no parent element is specified, we transform every <img> in
+ * the entire document.
+ */
+const appendAuthToImages: (auth: Auth, parent?: Element | Document) => void = (() => {
```
gnprice: This name becomes out of date, I think. Perhaps `rewriteImageUrls`?

src/webview/MessageList.js
```diff
@@ -99,6 +102,21 @@ export type Props = {|
   _: GetText,
 |};
 
+/**
+ * Path to the webview assets.
```
gnprice: This isn't really a "path" so much as a URL, right?

I think a few more words saying it's specifically a URL (/ path / etc.) *as seen from inside the webview* would be helpful -- otherwise it sounds like perhaps a filesystem path that the RN code might use.

src/webview/MessageList.js
```diff
@@ -180,13 +198,78 @@ class MessageList extends Component<Props> {
...
+      };
+    })();
+
+    // Note: neither `originWhitelist` nor `onShouldStartLoadWithRequest` are
+    // significant security barriers; they're checked only against the URL of
+    // the document itself. They cannot be used to validate the URL of any
+    // resource the WebView loads.
+    //
+    // Worse, as of at least v7.4.1, `originWhitelist` is completely broken: see
+    // https://github.com/react-native-community/react-native-webview/pull/697.
```
gnprice: "v7.4.1 of react-native-webview" would be helpful for interpreting the version number; even though it is possible to guess (as I'm guessing) based on that PR link :slightly_smiling_face:

src/webview/MessageList.js
```diff
@@ -180,13 +198,78 @@ class MessageList extends Component<Props> {
...
+    expected base URL = ${baseUrl}`);
+        }
+        return ok;
+      };
+    })();
+
+    // Note: neither `originWhitelist` nor `onShouldStartLoadWithRequest` are
+    // significant security barriers; they're checked only against the URL of
+    // the document itself. They cannot be used to validate the URL of any
+    // resource the WebView loads.
```
gnprice: Ah, it took me a bit to realize that there's still at least one plausible attack scenario this isn't excluding: if some malicious code tries to load a *new* document with e.g. `window.location = ...`, then the callback will fire.

Initially I thought this was saying these checks only applied to the *initial* load -- which made it hard to see how it could be worth the complexity of that `onShouldStartLoadWithRequest` callback (which I haven't read yet; that's next.)

As is: as you say, it's not a hard barrier, but it very plausibly might interrupt some attempted exploit.

src/webview/MessageList.js
```diff
@@ -180,13 +198,78 @@ class MessageList extends Component<Props> {
...
+     * here is not loaded, and doesn't even need to exist. It serves only as a
+     * placeholder, so that relative URLs and cross-domain security restrictions
+     * have somewhere to believe that this document originates from.
+     */
+    const baseUrl = `${assetsPath}/index.html`;
+
+    /**
+     * Paranoia^WSecurity: only load `baseUrl`, and only load it once. Any other
+     * requests should be handed off to the OS, not loaded inside the WebView.
+     */
```
gnprice: Nit: I like to reserve the `/**` jsdoc marker for comments that live up to being documentation. Here's a nice description of the principle, in just a slightly different context:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use--for-public-quality-private-documentation
(That style guide in general is pretty great, and a lot of especially the documentation section generalizes quite well.)

This comment is a fine one here -- this isn't an interface that needs documentation -- but should just be marked with plain `/*` or `//`.

src/webview/static/README.md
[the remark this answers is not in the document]
ray-kraesig: Added a few pointers in the relevant locations. (I've somewhat arbitrarily decided that `tools/build-webview` gets to be the "owner" of this decision: it seems a bit too fiddly to go in the current README, which is vaguely targeted at people producing assets rather than people maintaining the build system. That could be changed, though.)

src/webview/MessageList.js
ray-kraesig: Removed the version number, instead. (Given the PR link, it's probably of very little utility.)

src/webview/js/js.js
gnprice: Yeah -- MDN is a public wiki, and it's a very good one but can still suffer from the weaknesses of the genre.

Basically I think of it a lot like the stronger subject areas of Wikipedia.

Also, this means you can Be Bold and fix it there for the next person :slightly_smiling_face:

gnprice: Also also, I find it helpful to have a comment like "Adapted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes".

src/webview/js/js.js
```diff
@@ -466,31 +487,57 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
...
-    }
+      // Compute the absolute URL as though this page were the basis. (This
+      // should be equivalent to `img.src`, and may even be identical.)
+      const deducedSrc: string = new URL(actualSrc).href;
+
+      // Compute the absolute URL as though `auth.realm` were the basis.
+      let fixedSrc: string = new URL(actualSrc, realm).href;
+
+      // If the corrected URL is on this realm...
+      if (!fixedSrc.startsWith(auth.realm)) {
```
gnprice: Wait, is this the right way around?

gnprice: Also I'm not sure offhand whether `auth.realm` ends in a slash. If not, then this string-prefix check can go wrong.

In the next line, we'll check if `actualSrc` matches any of several patterns, each of which starts with `/`. So we could actually just drop this outer check.

src/webview/js/js.js
```diff
@@ -466,31 +487,57 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
...
+      }
 
-    const delimiter = img.src.includes('?') ? '&' : '?';
-    img.src += `${delimiter}api_key=${auth.apiKey}`;
-  });
-};
+      // Apply changes, if any.
+      if (fixedSrc !== deducedSrc) {
+        img.src = fixedSrc;
+      }
```
gnprice: Why not `if (fixedSrc !== img.src)`?

src/webview/js/js.js
ray-kraesig: There's so little of it left, I'm not sure that's true anymore! But it certainly holds for `startsWith`, so I'll add a note.

src/webview/static/README.md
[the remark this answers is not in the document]
gnprice: Looks good!

> I've somewhat arbitrarily decided that `tools/build-webview` gets to be the "owner" of this decision: it seems a bit too fiddly to go in the current README

Sounds good to me.

src/webview/MessageList.js
gnprice: Hmm -- in the Android case it begins with `file:///`, right? Which isn't a path; it's a URL.

I think I don't fully understand how this is ultimately getting interpreted in the iOS case. It's used to construct a URL -- it seems like it's effectively a URL reference that's relative to whatever the base URL would be by default? Would be glad to understand better. What sources have you found helpful in understanding what `baseUrl` on the iOS webview means?

src/webview/js/js.js
gnprice: > On the other hand, this is inside the WebView, and URL objects have been around for a very long time. Rewritten to explicitly let the browser parse URLs rather than doing it by hand,

SGTM!

src/webview/js/js.js
ray-kraesig: It's not, no. I need to write actual tests for this function.

`auth.realm` does not end in a slash unless something has already gone wrong. See `fixRealmUrl` in `url.js`, and its tests in `url-test.js`. (Also, this is exactly what the old code does.)

Oof. We can't drop this outer check, as the server may have sent us an image URL which is already explicitly absolutized to `auth.realm`, but does not have the API key attached. Going on to compare the raw `src` to strings which begin with `'/'` is a bug.

src/webview/js/js.js
ray-kraesig: See comment near `deducedSrc`. Equivalence is not equality, and I'm a lot less certain that `img.src` and `fixedSrc` are normalized identically than that `deducedSrc` and `fixedSrc` are.

(They might be! It's entirely possible that all relevant URLs have unique normalized forms. I just didn't want the correctness of this code to depend on that fact.)

It's probably worth reordering or recommenting the code to make that clear here, though.

src/webview/js/js.js
gnprice: > Oof. We can't drop this outer check, as the server may have sent us an image URL which is already explicitly absolutized to `auth.realm`, but does not have the API key attached.

Hmm, yeah, that probably can happen.

That isn't helped by this outer check, though, right? Ah, I guess you mean something like: we need to first check the scheme+authority against `realm` (i.e. a version of this outer check), and then check the *path* -- pulled out from the fixed URL -- against the patterns.

src/webview/js/js.js
gnprice: > Equivalence is not equality

Can you expand on what you mean by this? These are strings, so `===` is a by-value comparison (unless `img.src` is more complicated?)... but I'm not sure I'm understanding what you mean by "equivalence" here (and/or perhaps "equality").

gnprice:
I've just merged the first 5 of the 11 commits in this branch, as:
```
bfb15d794 Xcode project: update with new version
e6b193607 webview build: add new script `tools/build-webview`
690731545 webview build: call new script from Gradle
63fe17925 gradle: Add comment-pointers to commits fixing related issues
5ca2fbd2b webview build: call new script from Xcode
```

Thanks! You've also sent a version of the 6th as #3644, which I expect I'll merge shortly.

src/webview/js/js.js
ray-kraesig: Sorry: I mean to say that I'm aware that URL equivalence is not _quite_ string equality (trailing slashes after bare hostnames, percent-escaping, explicit specification of port 80/443 for http/https respectively).

It's _very likely indeed_ that any browser will compute `img.src` from the raw `src` with the same normalization as is used by `new URL(..., document.location)`; but many more horrible things have been done in the name of backwards compatibility than choosing a different normalization form in different contexts.

(Of course, overwriting `img.src` with a formally-distinct-but-equivalent value wouldn't exactly be particularly harmful, in this context...)

gnprice: That's helpful, thanks.

I guess the place I start from is: the obvious way to write this is `img.src = fixedSrc;`, with no conditional. That would guarantee the `src` gets set to the URL we've computed for what it should be.

Is there a case where it'd be semantically wrong to do that? If there is, I don't yet see it; and I don't yet see why this condition would correspond to avoiding such a case. Conversely, if we can say "the `src` will always be the URL that's computed by this logic, here in this function", then that'd be a nice and relatively-simple situation to reason about.

What I've assumed this condition is doing is an optimization -- DOM operations can be slow, so avoid setting the value on the DOM if it'd be a no-op. (I'm not 100% sure this even matters; you might hope that the DOM implementation is smart enough that it can do the same obvious optimization, and avoid doing any of the expensive operations if it can see it's a no-op.)

From that perspective of "the semantic thing we want is `img.src` becomes `fixedSrc`, and we're just optimizing that", this seems like a tricky way to write the optimization. Maybe it's broader, and maybe the broader situations it applies to are also correct for the optimization... but it seems like there's more to understand before being confident of that than if it were just `if (fixedSrc != img.src)`.

IOW: I understand you're saying the two conditions could have different results for complicated reasons. AFAICT the correctness we're talking about is that of an optimization, so a version that just optimizes in fewer cases can only be more correct, not less so. If that's right, then out of two different ways to write it where the difference is complicated to understand, I'd rather have the version where it's easier to tell that every case where it applies the optimization is a correct one. Even if that gives up the optimization in some other cases where, upon closely studying the semantics, it'd turn out that the optimization is correct.

src/webview/js/js.js
ray-kraesig: (And the MDN update has been submitted.)

src/webview/MessageList.js
ray-kraesig: Ah, I see the confusion. Well, part of it, at least.
* URLs and filesystem paths are not formally disjoint. A string starting with `./` may potentially be both. (This is the case for the iOS value of `baseUrl`.)
* URLs and filesystem paths are not _colloquially_ disjoint. A string starting with `file://` may potentially be both. (This is the case for the Android value of `baseUrl`.)

... admittedly, the second of these two may not actually be true in the world outside my head. >_>

----

Formally, `baseUrl` is a potentially-relative URL with the `file:` scheme. As the `file:` scheme does not permit relative paths (normatively: [RFC 8089](https://tools.ietf.org/html/rfc8089#section-2); informatively: [Stack Overflow](https://stackoverflow.com/a/53082960) (first comment, not the answer)), the scheme is simply not given if the URL is relative. Either way, `baseUrl` names the filesystem directory containing the webview's static assets.

Relative URLs, being indistinguishable from relative paths, are interpreted in exactly the same way: that is, relative to the process's current directory (for iOS, see [currentDirectoryPath](https://developer.apple.com/documentation/foundation/nsfilemanager/1410766-currentdirectorypath?language=objc)). On iOS this is initially – and for us, always – the [root directory of the bundle](https://developer.apple.com/documentation/appkit/nsrunningapplication/1535500-bundleurl), which is mounted on some subdirectory of the primary volume in a mostly-unsurprising fashion.

----

I'm not sure what subset of the above, or what rewording, would be appropriate to include.

(I also haven't attempted to change the process's `$CWD` in order to verify that WebView does what I think it does rather than always using `bundleURL` or something.)

src/webview/js/js.js
ray-kraesig: Either adopted or made irrelevant in the rewrite; I forget which.

ray-kraesig:
Ready for (re-)review.

(If you've looked at the version from three days ago, the only difference between that and this is a typo fix.)

src/webview/MessageList.js
gnprice: > * URLs and filesystem paths are not formally disjoint. A string starting with `./` may potentially be both.

To this bit I was about to reply, based on my recollection of the RFCs, that a string beginning with `./` would be only a relative URL reference and not an actual URL.

... But this is another bit of terminology where the new, modern-web-style, standard has shifted things!
https://url.spec.whatwg.org/#valid-url-string
(Probably for the best, and certainly in the direction of common usage.) So yep, such a thing would be a "valid URL string" (specifically a "relative-URL-with-fragment string" that begins with a "path-relative-scheme-less URL string"); and that term in turn is [recognized](https://url.spec.whatwg.org/#concept-url) as the thing the single word "URL" means if it doesn't mean a `URL` object.


> * URLs and filesystem paths are not _colloquially_ disjoint. A string starting with `file://` may potentially be both. (This is the case for the Android value of `baseUrl`.)
>
> ... admittedly, the second of these two may not actually be true in the world outside my head. >_>

But yeah, this one isn't true in my head :smile: and I think it isn't an interpretation most people would apply to this description either.

The way I'd describe a string like `file:///foo/bar` is:
* it's a URL;
* specifically it's a URL that *encodes* the filesystem path `/foo/bar`.

It's certainly closely related to that filesystem path! But it's an encoding of it that is different from the canonical "filesystem path" serialization as the eight bytes `/foo/bar`.

One reason I wouldn't say it *is* a filesystem path is: if you pass the 15-byte string `file:///foo/bar` to any system call that expects a path, or to a command like `cat` that expects a path (to pass on to some syscall), it won't work:
```
$ cat /etc/hostname
dark-matter

$ cat file:///etc/hostname
cat: 'file:///etc/hostname': No such file or directory
```

It's only if you pass it to something that knows it's expecting a URL -- like a browser -- that the program will decode the string as a URL, extract the filesystem path that's inside, and use that as a path.

gnprice: > As the `file:` scheme does not permit relative paths (normatively: [RFC 8089](https://tools.ietf.org/html/rfc8089#section-2); informatively: [Stack Overflow](https://stackoverflow.com/a/53082960) (first comment, not the answer))

Small comment on this: now that we've discovered the modern URL specification (:100:), we can link here: https://url.spec.whatwg.org/#url-writing

I think that also may be clear enough to be able to skip the other link, too.

gnprice: OK, reading your description paragraphs above, and now rereading the code, I think I understand the story.

Would it be accurate to say that this is always a URL for the assets? In the one case it's an absolute URL, with the `file` scheme; in the other it's a relative URL (and will be interpreted relative to some `file`-scheme URL.)

---

It seems kind of mysterious what the relative URL actually gets resolved relative to. It cannot itself serve as [the document's actual base URL](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url) -- that, naturally enough, has to be an absolute URL. So it's actually kind of puzzling that a relative URL works here at all -- something must be resolving it to an absolute one.

The [docs](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#source) aren't much help... then grepping the implementation, it goes straight to [the `baseURL` parameter](https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring) sent to the WKWebView, and that doc doesn't help either.

A couple of clues from searching the web:
* A normal value in iOS-land to use there [might](https://www.hackingwithswift.com/articles/112/the-ultimate-guide-to-wkwebview) be [`Bundle.main.resourceURL`](https://developer.apple.com/documentation/foundation/bundle/1414821-resourceurl).
* Another normal thing to do in iOS-land [might be to](https://stackoverflow.com/questions/39336235/wkwebview-does-load-resources-from-local-document-folder) construct it like `URL(fileUrlWithPath: Bundle.main.bundlePath, isDirectory: true)`.
  * Seems like `bundleURL` would be simpler... but that is a hint that the URL could just end in `/` without something like an `index.html`.

How did you discover that a relative URL here was a possibility?

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+  // Parse URL into origin and path-plus-search.
+  const splitUrl = (url: URL): { head: string, rest: string } => {
+    const head = origin(url);
+    const rest = url.href.slice(head.length);
+    return { head, rest };
+  };
+
+  // List of routes which accept the API key appended as a GET parameter.
+  const inlineApiRoutes: string[] = ['/user_uploads/', '/thumbnail?', '/avatar/'];
+
```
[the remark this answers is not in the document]
ray-kraesig: Should be doable. The current layout is a relict of the original code's inline position.

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+ *   1. Make relative URLs absolute, with a path based on the Zulip realm rather
+ *      than the document location.
+ *   2. If the source URL names an endpoint known to require authentication,
+ *      inject an API key into its query parameters.
+ *
+ * DEPRECATED: If no parent element is specified, transform every <img> in the
+ * entire document.
+ */
+const rewriteImageUrls: (auth: Auth, parent?: Element | Document) => void = (() => {
+  // Ponyfill for URL.origin (Chrome 52, Safari 10).
```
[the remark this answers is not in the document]
ray-kraesig: It's like a monkeypatch, only completely different. Also I deny all responsibility in the matter!

* https://ponyfoo.com/articles/polyfills-or-ponyfills
* https://github.com/sindresorhus/ponyfill

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+      // Break it into `origin` and `path` components.
+      const fixedSrc: URL = new URL(actualSrc, realm);
+      const { head: fixedOrigin, rest: fixedPath } = splitUrl(fixedSrc);
+
+      // If the corrected URL is on this realm...
+      if (fixedOrigin === origin(realm)) {
+        // ... check to see if it's a route that needs the API key...
+        if (inlineApiRoutes.some(route => fixedPath.startsWith(route))) {
+          // ... and append it, if so.
+          const delimiter = actualSrc.includes('?') ? '&' : '?';
```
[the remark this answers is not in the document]
ray-kraesig: It doesn't show up well here, thanks to the fact that I did the rename and rewrite in one commit (which was a bad idea) – but, modulo some renaming, this logic remained essentially unchanged through the rewrite.

This is worth a follow-up commit to clean up, though.

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+
+      // Compute the absolute URL as though `auth.realm` were the basis.
+      // Break it into `origin` and `path` components.
+      const fixedSrc: URL = new URL(actualSrc, realm);
+      const { head: fixedOrigin, rest: fixedPath } = splitUrl(fixedSrc);
+
+      // If the corrected URL is on this realm...
+      if (fixedOrigin === origin(realm)) {
+        // ... check to see if it's a route that needs the API key...
+        if (inlineApiRoutes.some(route => fixedPath.startsWith(route))) {
```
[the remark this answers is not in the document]
ray-kraesig: As the preceding, on all counts. (This will also eliminate the need for `splitURL`.)

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+ * entire document.
+ */
+const rewriteImageUrls: (auth: Auth, parent?: Element | Document) => void = (() => {
+  // Ponyfill for URL.origin (Chrome 52, Safari 10).
+  const origin = (url: URL) => {
+    if (url.origin) {
+      return url.origin;
+    }
+    const { href, pathname, search } = url;
+    return href.slice(0, href.length - pathname.length - search.length);
```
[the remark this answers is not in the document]
ray-kraesig: It does not! I've added `hash.length` for now, but a rewrite along the lines you suggested is a good idea for followup work.

src/webview/js/rewriteImageUrls.js
```diff
@@ -0,0 +1,102 @@
...
+      const fixedSrc: URL = new URL(actualSrc, realm);
+      const { head: fixedOrigin, rest: fixedPath } = splitUrl(fixedSrc);
+
+      // If the corrected URL is on this realm...
+      if (fixedOrigin === origin(realm)) {
+        // ... check to see if it's a route that needs the API key...
+        if (inlineApiRoutes.some(route => fixedPath.startsWith(route))) {
+          // ... and append it, if so.
+          const delimiter = actualSrc.includes('?') ? '&' : '?';
+          fixedSrc.search += `${delimiter}api_key=${auth.apiKey}`;
```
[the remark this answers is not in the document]
ray-kraesig: (See following comment.)

ray-kraesig:
This push does not (yet?) include the refactoring commits mentioned above.

Also, GitHub won't let me respond to the URL/path comment chain... which is probably for the best, honestly. The important thing is that I've adjusted the comments in that area, and hopefully in a constructive manner!

src/webview/js/rewriteImageUrls.js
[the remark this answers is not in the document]
gnprice: I see. It's like the old `arrayFrom`, then! Seems like a fine concept to have a name for.

The latter link (aka ponyfill.com ) seems to explain the coinage, too.

gnprice:
> Also, GitHub won't let me respond to the URL/path comment chain

FFR on some other issue where you'd like to do so (no need to do so here, it's a long comment-chain and the branch is the place where edits ultimately matter): I think sometimes when a reply-comment is part of a review, it lacks a reply box where it appears as part of the review. Like here:

image

but at the same time it *also* appears upthread underneath the OP comment it's a reply to, and it has a reply box there:

image

GitHub. *shrug*

So that's what I'm seeing for that comment. Does that match what you see?

gnprice:
Thanks for the updates! This looks good. I think what I'd like to do next with this is:
* I'll go ahead and merge it. (After a filter-branch with tools/generate-webview-js -- that's the bit that's failing in CI.)
* Then I'll take a stab at the URL-related refactorings we've discussed above.

That'll let us swap places on this code, since it's been for a few weeks in a mode of you writing and me reviewing. I'll get to try to apply and serialize the things I learned yesterday from studying the spec and docs, which is probably most efficient by me writing and you reviewing; and it'll help you be done and move on to the next phase.

---

The comment on `assetsPath` makes much more sense to me now, thanks. :slightly_smiling_face: I think a more apropos name to go with that description would be like `assetsPathUrl` -- something that identifies that it's in the form of a URL, not directly as a path. But the comment helps a lot, and it's not like this is a name in a broad API that has a bunch of callsites scattered all over, which would make the name much higher-stakes. So I'm happy merging, and possibly tweaking in followup.

It's also still quite mysterious what this is interpreted relative *to* when it's a relative URL, on iOS. But as I found in https://github.com/zulip/zulip-mobile/pull/3636#discussion_r340343686 , that mystery seems to go all the way down through the WebKit API into the Apple secrecy-bubble. I may try turning that PR comment into one in the tree and/or a commit message.

gnprice:
> * (After a filter-branch with tools/generate-webview-js -- that's the bit that's failing in CI.)

FTR the command was:
`git filter-branch --tree-filter tools/generate-webview-js @{u}..master`

That's something that should be more automated, has never been quite annoying enough to automate further... but one of the side benefits of having `tools/build-webview` is that it should be pretty easy to automate that away soon!

gnprice:
> It's also still quite mysterious what this is interpreted relative _to_ when it's a relative URL, on iOS. But as I found in [#3636 (comment)](https://github.com/zulip/zulip-mobile/pull/3636#discussion_r340343686) , that mystery seems to go all the way down through the WebKit API into the Apple secrecy-bubble. I may try turning that PR comment into one in the tree and/or a commit message.

Done, as 067a5e356. Please take a look!

(Because this is entirely in a comment, as I said recently elsewhere about docs I think it's best to handle it with primarily post-commit review.)

src/webview/MessageList.js
gnprice: Also expanded this comment to resolve this confusion, in e5bce1d2f .

src/webview/MessageList.js
ray-kraesig: > How did you discover that a relative URL here was a possibility?

By the simple expedient of failing to consider that it might _not_ be permitted. :upside_down_face: The actual resolution, as it happens, is performed by `+ (NSURL *)NSURL:(id)json` in [`React/Base/RCTConvert.m`](https://github.com/facebook/react-native/blob/4b487ba50025becb6a83c805b99d45651db6b8c1/React/Base/RCTConvert.m#L113); it assumes `resourceURL` as the base.

As you note, this isn't documented. Explicit use of `resourceURL` (or `bundleURL`, which [seems to be identical on iOS](https://old.reddit.com/r/swift/comments/4p33iq/resourcepath_vs_bundleurl_vs_bundlepath/)?) would be preferable.

(I'm still reluctant to bring in `react-native-fs` just for that, but I'm much less reluctant than I was when I thought it used `file://${CWD}` as the base.)

ray-kraesig:
> The comment on `assetsPath` makes much more sense to me now, thanks.  I think a more apropos name to go with that description would be like `assetsPathUrl` [...]

Works for me.

> > It's also still quite mysterious what this is interpreted relative _to_ when it's a relative URL, on iOS. But as I found in [#3636 (comment)](https://github.com/zulip/zulip-mobile/pull/3636#discussion_r340343686) , that mystery seems to go all the way down through the WebKit API into the Apple secrecy-bubble. I may try turning that PR comment into one in the tree and/or a commit message.
> 
> Done, as [067a5e3](https://github.com/zulip/zulip-mobile/commit/067a5e35602fc32a7e712cf7271eefc60d653db0). Please take a look!
> 
> (Because this is entirely in a comment, as I said recently elsewhere about docs I think it's best to handle it with primarily post-commit review.)

This appears to be evidence against that proposition. :upside_down_face: I'll have a revision along shortly for review.

PR #3638, opened 2019-10-07
Try sharing some code from the webapp

gnprice:
This is the mobile-side companion to zulip/zulip#13253 .

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -29,6 +29,12 @@
 
 
 [include]
+; Effectively this line warns Flow that this path might be a symlink
+; (because the developer might be using `yarn link` to have this point into
+; their zulip.git worktree), and it should watch/inotify at the target.
+node_modules/@zulip/shared/
+; Add more as needed!  See also metro.config.js and docs/howto/yarn-link.md .
+
 
 [libs]
 node_modules/react-native/Libraries/react-native/react-native-interface.js
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -6,10 +6,10 @@ import type { Dispatch, GetState, Narrow } from '../types';
 import * as api from '../api';
 import { PRESENCE_RESPONSE } from '../actionConstants';
 import { getAuth, tryGetAuth } from '../selectors';
-import { isPrivateOrGroupNarrow } from '../utils/narrow';
+import { isPrivateOrGroupNarrow, caseNarrowPartial } from '../utils/narrow';
+import { getAllUsersByEmail } from './userSelectors';
 
 let lastReportPresence = new Date(0);
-let lastTypingStart = new Date(0);
 
 export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean = false) => async (
   dispatch: Dispatch,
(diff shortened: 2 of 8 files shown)


gnprice:
This is now all ready except it's waiting for `@zulip/shared` to get uploaded after zulip/zulip#13253 is merged. (Until then, it relies on `../zulip` from the project root being a worktree for the zulip/zulip repo, with the #13253 branch checked out.)

But I think that makes it now ready for review! @ray-kraesig , PTAL.

.flowconfig
```diff
@@ -29,6 +29,12 @@
 
 
 [include]
+; Effectively this line warns Flow that this path might be a symlink
+; (because the developer might be using `yarn link` to have this point into
+; their zulip.git worktree), and it should watch/inotify at the target.
+node_modules/@zulip/shared/
+; Add more as needed!  See also metro.config.js .
```
ray-kraesig: This (and the comment in `metro.config.js`, as well) should probably point "up" to the `yarn-link` document, rather than "sideways" to a sibling.

(Also, the space before the period shouldn't be necessary.)

ray-kraesig:
Looks good, up to the NOMERGE. 

(I'd appreciate it if you could push in the Jest config relocation early; I'm looking to upgrade us to 24.9.0 and/or add `jest-extended`.)

gnprice:
Thanks for the review!

> (I'd appreciate it if you could push in the Jest config relocation early; I'm looking to upgrade us to 24.9.0 and/or add `jest-extended`.)

Sure -- pushed, as ba57c6b23^..0f829fe78.

gnprice:
> Looks good, up to the NOMERGE.

Oh also, to avoid ambiguity: the branch looks like this:
```
jest [nfc]: Move config out of package.json.
jest [nfc]: Organize config a bit.
metro [nfc]: Drop comment that made better sense in the template.
wip deps: Add @zulip/shared, for code shared from the webapp.
yarn: Make it possible to use `yarn link` to do work on dependencies.
NOMERGE: set resetCache to simplify debugging Metro config
NOMERGE reduce log noise for debugging typing notifs
[wip] typing status: Use @zulip/shared in place of our reimplementation.
```

The last commit comes after the two NOMERGE commits, but is just as intended for merge (once `@zulip/shared` is uploaded) as the earlier commits. Have you reviewed that one too, or were you saying you hadn't?

.flowconfig
gnprice: > This (and the comment in `metro.config.js`, as well) should probably point "up" to the `yarn-link` document, rather than "sideways" to a sibling.

Sure, good idea -- that also gives an answer to the question of where to link to that doc from. (As it stands, there are no links to it other than browsing the tree, and in our heads.)

I think the "sibling" link is actually still helpful, though: if you have the general idea and just need a reminder of the specifics, that is a lot fewer words (about 400 fewer) than the doc.

> (Also, the space before the period shouldn't be necessary.)

I generally prefer to minimize ambiguity about where the name ends and the surrounding sentence's punctuation begins.

In this example, if in my terminal I have this sentence without that space, then double-clicking on the filename selects (and puts in the clipboard for middle-click paste) the whole "word" including the period -- IOW, the ambiguity is enough to fool the terminal's heuristic for conveniently selecting and copying in units of words.

ray-kraesig:
> > Looks good, up to the NOMERGE.
> 
> The last commit comes after the two NOMERGE commits, but is just as intended for merge (once `@zulip/shared` is uploaded) as the earlier commits. Have you reviewed that one too, or were you saying you hadn't?

Sorry — I'd missed your reference to the Zulip branch with the proposed contents of `@zulip/shared`, and so was saying that I hadn't. I'll take a(nother) look.

.flowconfig
ray-kraesig: > double-clicking on the filename

I suspected; and I _thought_ I'd tested exactly that and had it select the filename without the period – but now I can't reproduce it. `>_<` Let me retract that, and instead suggest ``` `metro.config.js` ``` (or some other quotative mark), just so it doesn't look like a typo.

(But it's not important.)

ray-kraesig:
... and having looked, I'm going to say that, in its current state, that last commit is too `[wip]` and `TODO`-y. The API that's being shared feels a bit awkward in its own right; its use here, even more so.

Concretely, I would suggest first:
  * refactoring the `typing_status` module into a class (whether an ES6-style `class` or otherwise) with approximately the module's current interface, instantiated when it's handed a `worker`; or
  * effectively doing the above locally by wrapping the imported shared module, with future work to include something like "pushing (a version of) the wrapper into `@zulip/shared` as the new interface".

(Given the TODO-notes, I infer you were already planning something similar to the first option?)

gnprice:
Thanks for the further review!

I've pushed a revised version that wraps `typing_status` in an interface that better reflects how I'm logically thinking of it, with jsdoc. Please take a look.

gnprice:
And we now have `@zulip/shared` up on NPM! https://www.npmjs.com/package/@zulip/shared

Pushed an update to switch to that from `file:`.

src/users/usersActions.js
```diff
@@ -35,6 +35,76 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
   });
 };
 
+const typingWorkerBase = auth => ({
```
ray-kraesig: It seems a bit odd to call this a `Base` when it's not being used for inheritance.

(It _is_ being used for Self-style inheritance, though; I suppose that's close enough. :upside_down_face:)

src/users/usersActions.js
```diff
@@ -35,6 +35,76 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
...
+  get_current_time: () => new Date().getTime(),
+
+  notify_server_start: user_ids_array => {
```
ray-kraesig: These two functions look like they want explicit types on their arguments.

You may also want to have them return `Promise<void>` rather than `void` – not in this commit, of course, but possibly in the later refactoring step you mention. (I'm not sure whose responsibility error-reporting should be, but it seems like it should happen somewhere around here.)

src/users/usersActions.js
```diff
@@ -43,27 +113,23 @@ export const sendTypingStart = (narrow: Narrow) => async (
...
+    }
+    return user.user_id;
+  });
 
   const auth = getAuth(getState());
-  api.typing(auth, narrow[0].operand, 'start');
+  maybeNotifyTyping(auth, recipientIds);
 };
 
+// TODO call this on more than send: blur, navigate away, etc.
```
ray-kraesig: In "etc." perhaps include "user deletes all unsent text".

src/users/usersActions.js
gnprice: > (It _is_ being used for Self-style inheritance, though; I suppose that's close enough. )

That's more or less what I was thinking :slightly_smiling_face: , though I didn't have Self's name specifically in mind.

I was thinking more like "base layer" or "base model" than object-oriented inheritance... but it's not really a coincidence that the same word appears there too.

src/users/usersActions.js
gnprice: > These two functions look like they want explicit types on their arguments.

Sure.


> You may also want to have them return `Promise<void>` rather than `void` – not in this commit, of course, but possibly in the later refactoring step you mention. (I'm not sure whose responsibility error-reporting should be, but it seems like it should happen somewhere around here.)

Hmm, good thought. I'll have that in mind as I refactor on the webapp side.

src/users/usersActions.js
gnprice: Sure.

gnprice:
Thanks for the re-review!

I've made the requested changes, and merged.

Issue #2781, opened 2018-07-09
[iOS] Navigation to Lightbox hangs in the between.


jainkuniya:
On iOS when image in the message list is clicked, navigations hangs in the between (half screen is pushed & has screen is occupied by chat screen).

Most prob this is due to `PhotoView` component we are using from `react-native-photo-view`.

borisyankov:
This looks like the issue:
https://github.com/alwx/react-native-photo-view/issues/141

jackrzhang:
I notice that before the screen hangs, the Lightbox background is black, and afterwards, the photo is shown in the Lightbox.

![lightbox-halfway-delay](https://user-images.githubusercontent.com/12771126/45852465-e3853480-bcf4-11e8-9cdf-2287b14c2e40.gif)


The delay seems to occur while, in some sense, the image data is being loaded.

tmjordan:
i have the same problem

PR #3639, opened 2019-10-08
Make API errors better (v2)

### LINKED ISSUE #2780: User list lags in create group screen.
If user list is long (like as of chat.zulip.org) on clicking user list there is a lag in the UI feedback. Now filter the user list, and the lag is also reduced proportionally.

@zulipbot claim

### LINKED ISSUE #2902: Editing a message outside of the valid edit window should show a helpful message.
Our organization is on zulipchat.com with a 10-minute edit window.

In the app, if a user attempts to edit a message that is outside of this window, a modal dialog pops up with just the text `API`. The user didn't know that they couldn't edit the message due to policy reasons -- they then thought that there was an error in the app or with the server.

Two options:
1. The "Edit" button should be grayed-out if the edit cannot be performed due to policy reasons, possibly with an explanation like "Edit (10 minutes only)", or something like that.
2. If the user manages to make an Edit request anyway, the returned message should explain something about the organization Edit policy.

ray-kraesig:
Update of PR #3484. Addresses most (hopefully all) review feedback therein.

Fixes #2902. Supersedes PRs #2781 and #3484.

### DIFF
--- a/src/api/apiErrors.js
+++ b/src/api/apiErrors.js
@@ -43,3 +43,15 @@ export const makeErrorFromApi = (httpStatus: number, data: mixed): Error => {
   logging.warn(`Bad response from server: ${JSON.stringify(data)}`);
   return new Error('Server responded with invalid message');
 };
+
+/**
+ * Is exception caused by a Client Error (4xx)?
+ *
+ * Client errors are often caused by incorrect parameters given to the backend
+ * by the client application.
+ *
+ * A notable difference between a Server (5xx) and Client (4xx) errors is that
+ * a client error will not be resolved by waiting and retrying the same request.
+ */
+export const isClientError = (e: Error): boolean =>
+  e instanceof ApiError && e.httpStatus >= 400 && e.httpStatus <= 499;
--- a/src/api/apiErrors.js
+++ b/src/api/apiErrors.js
@@ -1,0 +1,45 @@
+/* @flow strict-local */
+import type { ApiErrorCode, ApiResponseErrorData } from './transportTypes';
+import * as logging from '../utils/logging';
+
+/** Runtime class of custom API error types. */
+export class ApiError extends Error {
+  code: ApiErrorCode;
+  data: $ReadOnly<{ ... }>;
+  httpStatus: number;
+
+  constructor(httpStatus: number, data: $ReadOnly<ApiResponseErrorData>) {
+    // eslint-disable-next-line no-unused-vars
+    const { result, code, msg, ...rest } = data;
+    super(msg);
+    this.data = rest;
+    this.code = code;
+    this.httpStatus = httpStatus;
+  }
+}
+
+/**
+ * Given a server response (allegedly) denoting an error, produce an Error to be
+ * thrown.
+ *
+ * If the `data` argument is actually of the form expected from the server, the
+ * returned error will be an {@link ApiError}; otherwise it will be a generic
+ * Error.
+ */
+export const makeErrorFromApi = (httpStatus: number, data: mixed): Error => {
+  // Validate `data`, and construct the resultant error object.
+  if (typeof data === 'object' && data !== null) {
+    if (data.result === 'error' && typeof data.msg === 'string') {
+      // If `code` is present, it must be a string.
+      if (!('code' in data) || typeof data.code === 'string') {
+        // Default to 'BAD_REQUEST' if `code` is not present.
+        return new ApiError(httpStatus, { code: 'BAD_REQUEST', ...data });
+      }
+    }
+  }
+
+  // Server has responded, but the response is not a valid error-object.
+  // (This should never happen, even on old versions of the Zulip server.)
+  logging.warn(`Bad response from server: ${JSON.stringify(data)}`);
+  return new Error('Server responded with invalid message');
+};
--- a/src/api/transportTypes.js
+++ b/src/api/transportTypes.js
@@ -31,31 +31,42 @@ export type ApiResponseSuccess = {|
   msg: '',
 |};
 
-/** List of error codes at https://github.com/zulip/zulip/blob/master/zerver/lib/exceptions.py */
-export type ApiErrorCode =
-  | 'BAD_REQUEST'
-  | 'REQUEST_VARIABLE_MISSING'
-  | 'REQUEST_VARIABLE_INVALID'
-  | 'BAD_IMAGE'
-  | 'REALM_UPLOAD_QUOTA'
-  | 'BAD_NARROW'
-  | 'MISSING_HTTP_EVENT_HEADER'
-  | 'STREAM_DOES_NOT_EXIST'
-  | 'UNAUTHORIZED_PRINCIPAL'
-  | 'UNEXPECTED_WEBHOOK_EVENT_TYPE'
-  | 'BAD_EVENT_QUEUE_ID'
-  | 'CSRF_FAILED'
-  | 'INVITATION_FAILED'
-  | 'INVALID_ZULIP_SERVER';
+/**
+ * A list of current error codes can be found at:
+ *   https://github.com/zulip/zulip/blob/master/zerver/lib/exceptions.py
+ *
+ * Unfortunately, the `code` entry is a relatively late addition to the Zulip
+ * API: prior to (about) version 1.7.0, it hasn't always been present in
+ * responses. [1]  The modern default, when no other code has been defined, is
+ * 'BAD_REQUEST'; we therefore synthesize that value when connecting to old
+ * servers that don't provide an error code.
+ *
+ * It's tempting to make ApiErrorCode an enumerated type, save for the dual
+ * problem: when connecting to _newer_ Zulip servers, we may see values of
+ * `code` not known to this version of the client! In particular, new error
+ * codes are occasionally assigned to existing classes of error which previously
+ * returned 'BAD_REQUEST'.
+ *
+ * (Note that "newer" here doesn't necessarily mean "newer than this client",
+ * but "newer than the last time someone updated the error code list from the
+ * server". We have no mechanism in place to share the set of error codes --
+ * and, given the facts above, it's not clear that the existence of such a
+ * mechanism would be helpful anyway.)
+ *
+ * Though unstable in the general case, `code` is still useful. It _is_
+ * guaranteed to be returned, and stable, for certain known classes of errors
+ * arising from certain known endpoints. Nonetheless: let its user beware.
+ *
+ * [1] Reference: https://github.com/zulip/zulip/commit/709c3b50fc
+ */
+export type ApiErrorCode = string;
 
 /**
  * Response from the API in case of an error.
  *
- * @prop code - A string error code, one of several predefined values. See
- *   ApiErrorCode. (Some older API endpoints, or older Zulip servers, may not
- *   include this value; in this case it will be the modern default code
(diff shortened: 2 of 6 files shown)


ray-kraesig:
(Withdrawn pending CI fixes.)

src/api/transportTypes.js
```diff
@@ -31,25 +31,47 @@ export type ApiResponseSuccess = {|
...
-  | 'CSRF_FAILED'
-  | 'INVITATION_FAILED'
-  | 'INVALID_ZULIP_SERVER';
+/**
+ * A list of current error codes can be found at:
+ *   https://github.com/zulip/zulip/blob/master/zerver/lib/exceptions.py
+ *
+ * Unfortunately, the `code` entry is a relatively late addition to the Zulip
+ * API: it hasn't always been present in responses, and even on current versions
+ * of the server, may still not be present in some. Since we also can't reliably
```
gnprice: In fact the situation is just like this except a bit more so: most errors don't have a `code`, and it's not clear that most ever will. From a grep in the server repo, first for things with codes and then without:
```
$ rg 'class.*\(.*JsonableError' | wc -l
20

$ rg 'JsonableError\(' | wc -l
212
```

I think that's probably even the right strategy: if we did try to make a policy of "all errors have codes", realistically it would lead to a lot of error codes with ill-thought-out semantics. Instead we add codes lazily, whenever there's a known use case, and so we do OK on actually thinking through the API choices of the code's name, the boundaries of what it covers, etc.

src/api/transportTypes.js
```diff
@@ -31,25 +31,47 @@ export type ApiResponseSuccess = {|
...
+ *
+ * (Note, in particular, that "newer" doesn't necessarily mean "newer than this
+ * client", but "newer than the last time someone updated the error code list
+ * from the server". We have no mechanism in place to share the set of error
+ * codes... and, given the preceding facts, it is not clear that the existence
+ * of such a mechanism would be helpful.)
+ *
+ * For all that, `code` is not completely useless. It _is_ guaranteed to be
+ * returned, and stable, for certain known classes of errors arising from
+ * certain known endpoints, and all new errors should have a `code` value.
```
gnprice: (same discussion as above)

Oh just for precision in talking about it, though: I think in fact all errors from non-ancient servers (since 1.7.0 -- specifically 709c3b50f and near ancestors) do have *a* `code`; it's just that for most of them, it's a generic  'BAD_REQUEST'.

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
+/* @flow strict-local */
+import type { ApiResponseErrorData } from './transportTypes';
+import * as logging from '../utils/logging';
+
+/**
+ * Runtime class of custom API error types.
+ */
+export class ApiError extends Error {
+  data: $ReadOnly<ApiResponseErrorData>;
+  httpStatus: number;
```
gnprice: One thing I told Boris over on #3484 I think applies: the commit that deletes `makeApiError` would be best squashed into the one that adds this, because it actually makes both commits easier to understand. In particular when reading this definition, it makes clear that choices like having those two fields with their particular names have just been moved from that other, previous code.

If you want to keep the `isEventQueueIdError` change in a separate commit (which sounds like a good idea to me), this first commit could just change `e.message === 'API'` to `e instanceof ApiError`, I think.

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
...
+ * returned error will be an {@link ApiError}; otherwise it will be a generic
+ * Error.
+ */
+export const makeErrorFromApi = (httpStatus: number, data: ?mixed): Error => {
+  // Validate `data`, and construct the resultant error object.
+  if (data !== null && typeof data === 'object') {
+    if (typeof data.msg === 'string' && data.result === 'error') {
+      if ('code' in data && typeof data.code === 'string') {
+        /* $FlowFixMe, in the unlikely event we can ever narrow types with if
+        statements on properties. */
```
gnprice: nit: indent so it doesn't look like the start of a non-comment line

Perhaps just use `//` style comments? Actually, though, I'm (pleasantly) surprised that `$FlowFixMe` works when there's another comment line in between like this; not sure if that might vary with the comment style.

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Given a server response (allegedly) denoting an error, produce an Error to be
+ * thrown.
+ *
+ * If the `data` argument is actually of the form expected from the server, the
+ * returned error will be an {@link ApiError}; otherwise it will be a generic
+ * Error.
+ */
+export const makeErrorFromApi = (httpStatus: number, data: ?mixed): Error => {
```
gnprice: This `?mixed` is the same as `mixed`, right?

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
...
+      if ('code' in data && typeof data.code === 'string') {
+        /* $FlowFixMe, in the unlikely event we can ever narrow types with if
+        statements on properties. */
+        return new ApiError(httpStatus, data);
+      }
+    }
+  }
+
+  // Server has responded, but the response is not a valid error-object.
+  // (This should never happen, even on old versions of the Zulip server.)
```
gnprice: I think not having `code` is actually normal pre-1.7.0.

Logic here seems fine, just adjust comment and perhaps the warning message. The logged warning will in turn help us gauge if we can forget pre-1.7.0 servers yet!

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
...
+  logging.warn(`Bad response from server: ${JSON.stringify(data)}`);
+  return new Error('Server responded with invalid message');
+};
+
+/**
+ * The event queue is too old or has been garbage collected.
+ * https://zulipchat.com/api/get-events-from-queue#bad_event_queue_id-errors
+ */
+export const isEventQueueIdError = (e: Error): boolean =>
+  e instanceof ApiError && e.data.code === 'BAD_EVENT_QUEUE_ID';
```
gnprice: Like with `ApiError` itself, the commit adding this would be clarified a bit by squashing in the few-line change that switches to it. Much of the content of this is taken from the code deleted there, so that'd make the comparison a bit easier to look at.

src/api/apiErrors.js
```diff
@@ -0,0 +1,62 @@
...
+ * Error.
+ */
+export const makeErrorFromApi = (httpStatus: number, data: ?mixed): Error => {
+  // Validate `data`, and construct the resultant error object.
+  if (data !== null && typeof data === 'object') {
+    if (typeof data.msg === 'string' && data.result === 'error') {
+      if ('code' in data && typeof data.code === 'string') {
+        /* $FlowFixMe, in the unlikely event we can ever narrow types with if
+        statements on properties. */
+        return new ApiError(httpStatus, data);
```
gnprice: (Take the below as thoughts on a possible followup -- what's here is already an improvement, and will be a fine clean step in the history.)

Oh alternatively: one thing you could do here is say
```
        return new ApiError(httpStatus, { result: 'error', code: data.code, msg: data.msg });
```

That appears to pass Flow just fine.

I guess literally *that*, verbatim, isn't great because there are potentially those other properties on `data` that might be interesting. But it points in the direction of I think a pretty decent alternative: that version already cries out for ditching `result` as redundant, and pulling `code` and `msg` out onto the `ApiError` itself. Then the `data` property of `ApiError` could be just of type `{}`, any object at all; it could get either the original `data`, or perhaps better the result of a `const { result, code, msg, ...restData } = data`.

src/api/transportTypes.js
ray-kraesig: That _completely_ changes my preferred approach here. If the server's filling in `BAD_REQUEST` for "code not otherwise specified", then we might as well do the same on our end for pre-1.7.0 servers, and just document that it's an explicit catch-all.

src/api/apiErrors.js
ray-kraesig: If `ApiError` is going to be exposed to non-testing code anyway, I'm not sure a one-use-site trivial predicate like `isEventQueueIdError` helps at all; I might as well strip it entirely.

Which is convenient, because it's the only reason I separated out the commits in this particular way.

gnprice: Sure; sgtm!

src/api/transportTypes.js
gnprice: Huh -- yeah, I definitely agree! Had not put the pieces together to think of that. :slightly_smiling_face: That will simplify some things.

src/api/apiErrors.js
gnprice: > I'm afraid it does vary with the comment style. It seems to look for the next statement after the comment – including both compound statements and other comments.

Welp. Good to know that's the rule, though.

src/api/apiErrors.js
gnprice: > The existing logic _is_ incorrect, since it'll sometimes throw away a `msg` provided by a pre-1.7.0 server!

Hmm, fair enough!

src/api/apiErrors.js
ray-kraesig: Unfortunately, I'd better do this now.

There's a bug in my inlined implementation of `isEventQueueEtc.` that this will fix and Flow won't catch, because [caught exceptions are implicitly `any`, and _we can't turn that off_](https://github.com/facebook/flow/issues/2470).

gnprice: > [caught exceptions are implicitly `any`, and _we can't turn that off_](https://github.com/facebook/flow/issues/2470).

Womp womp.

Want to make a quick remark of that on `#mobile-team`? That will make it more likely we find this discussion if we see something that reminds us of it later... and also it's just an interesting fact for others to see.

src/api/apiErrors.js
```diff
@@ -0,0 +1,57 @@
...
+  data: $ReadOnly<{ ... }>;
+  httpStatus: number;
+
+  constructor(httpStatus: number, data: $ReadOnly<ApiResponseErrorData>) {
+    // eslint-disable-next-line no-unused-vars
+    const { result, code, msg, ...rest } = data;
+    super(msg);
+    this.data = rest;
+    this.code = code;
+    this.httpStatus = httpStatus;
```
gnprice: Picky nit: I find it's slightly easier to read when a series of properties (etc.) comes in the same order in different places -- makes it a bit less work to look back and forth between them and compare.

Here, that'd mean initializing these three in the same order they're declared above.

src/api/apiErrors.js
```diff
@@ -0,0 +1,57 @@
...
+ * Error.
+ */
+export const makeErrorFromApi = (httpStatus: number, data: mixed): Error => {
+  // Validate `data`, and construct the resultant error object.
+  if (typeof data === 'object' && data !== null) {
+    if (data.result === 'error' && typeof data.msg === 'string') {
+      // If `code` is present, it must be a string.
+      if (!('code' in data) || typeof data.code === 'string') {
+        // Default to 'BAD_REQUEST' if `code` is not present.
+        return new ApiError(httpStatus, { code: 'BAD_REQUEST', ...data });
```
gnprice: Huh neat, no fixme!

Seems like it's the spread that does it.

... Oh but it's that Flow is just unsound here :-/ -- it passes just as well if this line goes before any of those refinements. Ah well, so be it.

PR #3640, opened 2019-10-09
flow: downgrade all `@flow strict` to `@flow strict-local`

ray-kraesig:
Perhaps someday more of our dependencies will be `@flow strict`. At
present, though, even (our version of) React Native isn't -- and
trying to keep islands of `strict` afloat in a sea of `strict-local`
is probably counterproductive.

Performed automatically with:
```
rg 'flow strict[^-]' -l -g'*.js' |
    xargs sed -i 's/@flow strict/@flow strict-local/'
```


gnprice:
Looks like there's a few more:
```
$ git grep '@flow strict$' -- src/**/*.js
src/api/eventTypes.js: * @flow strict
src/api/modelTypes.js: * @flow strict
src/jsBackport.js: * @flow strict
Binary file src/webview/js/generatedEs3.js matches
```

Want to add that case to the pattern?

gnprice:
Also `sed -i` is fast enough you can skip the grep and xargs :-)

(So is `perl -i`.)

ray-kraesig:
Done! The `xargs` isn't for speed, though; I just find left-to-right dataflow more natural. (:

gnprice:
Thanks! Looks good -- merged.

PR #3642, opened 2019-10-09
gradle: run tasks only in `:app` namespace, where applicable

ray-kraesig:
`gradlew testDebugUnitTest` runs all known instances of tasks named
`testDebugUnitTest`. `gradlew :app:testDebugUnitTest` just runs ours.

(Also apply similar changes to other implicit uses of the `:app`
namespace.)

Fixes the entirely silent requirement we'd had on Java 8 to run our
tests: `react-native-image-picker` relies on `robolectric:3.3.2`,
which in turn relies on a Java 8 implementation detail. [1]

[1] https://community.oracle.com/thread/4011800


gnprice:
Thanks for tracking this down, and sorry you had to! Merging.

This means I also get to go drop that `JAVA_HOME` setting, to a JDK 8 install, that I've had on my desktop and had forgotten about until today. <_< >_>

Searching history for references to `JAVA_HOME`, I find the JDK 8 requirement *used* to be documented... until 0e45fda87, a few hours after you started :slightly_smiling_face: (and almost certainly prompted by your own setting up of a dev environment). In which I took it out because I thought it was already untrue, because I thought I'd been using JDK 11 for most of a year. In reality I was according to `java --version`... but it's apparently a convention in the Java world to have wrapper scripts find the JRE (the `java` executable) from `JAVA_HOME` in preference to `PATH`, and `gradlew` implements that convention, and so I was getting JDK 8 in everything under Gradle.

gnprice:
Merged, with a note appended to the commit message mentioning this bit:

> Searching history for references to `JAVA_HOME`, I find the JDK 8 requirement _used_ to be documented... until 0e45fda87 [...] [i]n which I took it out because I thought it was already untrue

in hopes of making the history a bit less confusing in light of that past mistake.

PR #3644, opened 2019-10-11
webview js: add polyfills for Android KitKat

ray-kraesig:
Neither `String.startsWith` nor `String.includes` existed in the stock Android KitKat browser.

(Extracted from PR #3636.)

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -33,6 +33,20 @@ if (!Element.prototype.closest) {
   };
 }
 
+if (!String.prototype.startsWith) {
+  String.prototype.startsWith = function startsWith(search, rawPos) {
+    var pos = rawPos > 0 ? rawPos | 0 : 0;
+    return this.substring(pos, pos + search.length) === search;
+  };
+}
+
+if (!String.prototype.includes) {
+  String.prototype.includes = function includes(search) {
+    var start = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 0;
+    return this.indexOf(search, start) !== -1;
+  };
+}
+
 var documentBody = document.body;
 
 if (!documentBody) {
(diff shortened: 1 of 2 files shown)


gnprice:
Travis failure was for lack of this diff :slightly_smiling_face: :
```
diff --git src/webview/js/generatedEs3.js src/webview/js/generatedEs3.js
index 361325a03..c717cf029 100644
--- src/webview/js/generatedEs3.js
+++ src/webview/js/generatedEs3.js
@@ -34,7 +34,7 @@ if (!Element.prototype.closest) {
 }
 
 if (!String.prototype.startsWith) {
-  String.prototype.startsWith = function value(search, rawPos) {
+  String.prototype.startsWith = function startsWith(search, rawPos) {
     var pos = rawPos > 0 ? rawPos | 0 : 0;
     return this.substring(pos, pos + search.length) === search;
   };
```

Made that change, and merging!

gnprice:
Also I'd encourage you to edit https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Polyfill with your improved version -- that'd be a nice small public service. (I would probably not touch the type-check stuff, unless I could do it accurately with no further research -- just fix the bogus `if (start + search.length > this.length)`.)

timabbott:
FWIW, on the webapp frontend we're using the `core-js` module for polyfills, with webpack's auto-detection of which features to include based on a `.browserslistrc` file.  @andersk did the integration so may have more advice on this, but that seems like definitely a better way to manage polyfills than doing them manually for things we notice we're using like this.

andersk:
> with webpack's auto-detection

More specifically, Babel’s, with the [`useBuiltIns: usage`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) option of `@babel/preset-env`.

```
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "corejs": 3,
        "useBuiltIns": "usage"
      }
    ]
  ]
}
```

I agree it’s likely to be a better solution for zulip-mobile too.

ray-kraesig:
Fortuitously enough, I've been working on integrating [rollup](https://rollupjs.org/) into our webview JS build today for other reasons... and, probably not coincidentally, have set up almost exactly that.

PR #3645, opened 2019-10-11
tools/build-webview: Let DEST be any path, not only absolute.

gnprice:
This brings this script's interface in line with normal CLI
conventions.  It doesn't matter when called from the build process
in Gradle or Xcode -- those can just pass an absolute path -- but
is helpful when running it by hand.

### DIFF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -35,10 +35,6 @@ err() {
 # Parameters and environment
 ################################################################################
 
-# chdir to the current git repo's root
-ROOT="$(git rev-parse --show-toplevel)"
-cd "$ROOT"
-
 # Parse arguments. Sloppy, but sufficient for now.
 unset TARGET
 unset DEST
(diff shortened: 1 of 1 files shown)


tools/build-webview
```diff
@@ -62,6 +58,14 @@ elif [ -z "${DEST-}" ]; then
     usage; exit 2
 fi
 
+# If DEST is relative (or contains symlinks), make it absolute for the
+# sake of the glob checks below.
+DEST=$(readlink -m "$DEST")
```
ray-kraesig: `readlink` doesn't appear to take a `-m` option on OS X.

gnprice: Oof, thanks. :-/ I'll go looking for a workaround, though not today.

gnprice:
This was taken care of by #3663.

(FTR, the way we handled the `readlink -m` issue there is that we added a call to `ensure-coreutils.sh`.)

PR #3646, opened 2019-10-11
xcode: Add a "run release mode" scheme.

gnprice:
Also a couple of small cleanups.

Details, and background (most of which I just learned), in the first commit's long commit message.

These are byproducts of attempting today to make it possible to test push notifications in a build made directly from Xcode, without going through TestFlight. Haven't gotten that part working yet, but these seem independently helpful.


ray-kraesig:
Looks good! Amended only to convert `http` to `https` in one of the commit messages.

PR #3653, opened 2019-10-17
Use WebP images on Android

ray-kraesig:
... which are slightly smaller, and have been fully supported since late Jelly Bean.


gnprice:
Sounds good!

Would you do two things in the commit message?
* Mention why we want to do this. :slightly_smiling_face: (It's in the PR description -- the files are slightly smaller -- but that's a draft and the repo is our primary record.)
* Saying "supported on Android 15+" is rather disorienting, as this year's new version is Android 10. :stuck_out_tongue: "API 15" works; or "Android API level 15" would be a formal, explicit way to put it.
* Actually here's a third thing: presumably you learned these facts about what's supported where from some reference. Would you link to that reference?

ray-kraesig:
Done!

gnprice:
Thanks -- merged!

Issue #3563, opened 2019-07-23
Update Android targetSdkVersion to 28, by 2019-11 deadline


gnprice:
We should update our `targetSdkVersion` to 28, aka Android 9.0 Pie. It's currently 26, since #3075.

Starting 2019-11-01, Google Play will require this for any new updates to our app.

Upstream doc here:
https://developer.android.com/about/versions/pie/android-9.0-migration
and some useful background here:
https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd

The important steps are
1. Read about the potentially-breaking changes, and identify those that might affect us.
2. Make a WIP change to `targetSdkVersion`.
3. Test the WIP change thoroughly, especially in the areas highlighted in step 1. Fix things as needed, and repeat.

Fortunately this is a less overwhelming set of changes than we had to deal with in #3075 -- because we aren't catching up on 4 years of platform changes at once, as we were then.

More details to fill in later. Filing now in order to have an issue to link to.

gnprice:
This can also be seen as a followup task from the RN v0.59 upgrade #3399 -- the upstream RN template app made this change in that interval.

gnprice:
I was just going through related docs to close out #3075 (and disaggregate the followup tasks that remain from it), so here are some more links.

A short outline of important changes in 9.0 / 28:
https://developer.android.com/distribute/best-practices/develop/target-sdk#prepie
That includes changes that already apply to us on 9.0, in addition to changes that are opt-in with `targetSdkVersion`.

A more detailed list, and focused exclusively on changes relevant to this issue (i.e., changes that will come into effect when we update `targetSdkVersion`):
https://developer.android.com/about/versions/pie/android-9.0-changes-28

A quick skim of that list finds:
* Likely requires action: `isCleartextTrafficPermitted` -- see also 1f55c035a
* Might require action: WebView changes re MIME types and re document scrolling
* Several others I'd have to investigate more to be confident they don't affect us.

Of the changes that already apply to us on Android 9 devices:
* [App Standby Buckets](https://developer.android.com/about/versions/pie/power#buckets) can affect ability to receive FCM messages (and hence show notifications.)
  * Confirms the importance of not overusing `priority: high` in FCM! Fortunately we've always gotten that right; see #3185.

gnprice:
Rereading [that list](https://developer.android.com/about/versions/pie/android-9.0-changes-28):

[...]

PR #3658, opened 2019-10-21
android: Update target SDK version

ray-kraesig:
Per [1] and [2], update our target SDK version before the November
deadline.

Fortunately, there appear to be no changes listed that affect us. [3]

[1] https://developer.android.com/distribute/best-practices/develop/target-sdk
[2] https://android-developers.googleblog.com/2019/02/expanding-target-api-level-requirements.html
[3] https://developer.android.com/about/versions/pie/android-9.0-changes-28

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,7 +13,7 @@ buildscript {
         // but some testing is required and code changes are often required.
         // See upstream for background and more discussion:
         //   https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd
-        targetSdkVersion = 26
+        targetSdkVersion = 28
 
         // Should be the latest SDK version available.  See upstream recommendation:
         //   https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd


ray-kraesig:
Note that I'm not as sure about this commit as I was ten minutes ago: on the Google Play Console there's a list of currently-greylisted APIs we're using. None of them seem to be on the (very short) [Pie blacklist](https://android.googlesource.com/platform/frameworks/base/+/pie-release/config/hiddenapi-force-blacklist.txt), but they should probably still tracked down and verified harmless before a full release.

gnprice:
Thanks! The other cross-reference here is #3563 .

As we discussed in person, I think the next thing we want to do to validate this, following the investigations you've done of each of the areas where things have changed (and in parallel with the last steps like those greylisted APIs), is to get it out to our beta users. That will help surface any unknown unknowns.

Let's send out a beta tomorrow, I think; would do it now except it's getting late in the day (and I'd like to tie up a few unrelated loose ends before the day's over.)

In the commit message, it'd be helpful to mention #3563. Perhaps leave it open until we've completed the validation we want to do, though.

gnprice:
Going to merge and send this to beta! :crossed_fingers: 

Let's use #3563 for continuing the discussion of those graylisted APIs and anything else that comes up.

PR #3663, opened 2019-10-23
Fix invocation of bash script from Gradle on Windows

ray-kraesig:
Sadly, this was more complicated than it looked, due to WSL/MSYS disagreements. (See comments.)

As a side effect, for the sake of script-maintainers' sanity, macOS users are now required to install `coreutils`.

### DIFF
--- a/android/app/webviewAssets.gradle
+++ b/android/app/webviewAssets.gradle
@@ -1,3 +1,6 @@
+import org.apache.tools.ant.taskdefs.condition.Os
+import java.nio.file.Paths
+
 /**
  * Build and merge static webview assets.
  *
@@ -6,6 +9,49 @@
  * processes. See comments therein (or $ROOT/src/webview/static/README.md) for
  * more information.
  */
+
+/*
+ * Windows support functions.
+ *
+ * On Windows, we have three different possible setups:
+ *   1. Windows Gradle + Git Bash.
+ *   2. Windows Gradle + WSL's bash.
+ *   3. WSL Gradle + WSL's bash.
+ *
+ * Case 3, fortunately, should behave exactly like Linux.
+ *
+ * In cases 1 and 2, however, all paths Gradle provides are backslash-separated.
+ * This is problematic on its own, as the script we're invoking doesn't grok
+ * backslashes, so we normalize those on Windows. That's enough for case 1.
+ *
+ * Case 2 is worse, though. The `bash` of Git Bash groks both "C:/" and "C:\";
+ * more generally, it operates on the Windows filesystem with Windows path
+ * names. The WSL `bash` _doesn't_ -- it's a standard Linux `bash` (or close
+ * enough), and it only sees the Windows disks as ordinary mounts under the lone
+ * root path '/'. Paths beginning with either "C:/" or "C:\" will confuse it
+ * terribly.
+ *
+ * And, of course, Gradle has no good way of distinguishing between cases 1 and
+ * 2. As far as it's concerned, `bash.exe` is `bash.exe`.
+ *
+ * Since we can't predict the environment of our callee, we can't use absolute
+ * paths. Fortunately, with only a little extra work within the script itself,
+ * we _can_ use relative paths; the script must derelativize them appropriately.
+ */
+
+/** Normalize the path-separators of `path` to be forward slashes. */
+def normalizePath(String path) {
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        return path.replace('\\', '/')
+    } else {
+        return path
+    }
+}
+
+/** Return `path`, made relative to `basePath`. */
+def relativizePath(String path, String basePath) {
+    return Paths.get(basePath).relativize(Paths.get(path)).toString();
+}
 
 gradle.projectsEvaluated {
     // The root of our git repository.
@@ -24,8 +70,13 @@ gradle.projectsEvaluated {
             name: "build${variant.name.capitalize()}StaticWebviewAssets",
             type: Exec
         ) {
-            executable "${repoDir}/tools/build-webview"
-            args "android", "--destination", "${assetsDir}/webview"
+            // All arguments to our script must be relative to `workingDir`.
+            workingDir repoDir
+            executable "bash"
+            args "./tools/build-webview",
+                "android", "--destination",
+                // See above note on Windows compatibility.
+                normalizePath(relativizePath("${assetsDir}/webview", repoDir))
         }
 
         // Run this task as part of assets merging for this variant.
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -4,9 +4,10 @@ This guide describes how to build and run the app so you can develop it.
 
 ## Main steps
 
-(First, if using **macOS**: Upgrade to the latest version of the OS and then
-to the latest Xcode.  In particular, Xcode versions before 9.0 are known to
-definitely not work.)
+(First, if using **macOS**: Upgrade to the latest version of the OS and then to
+the latest Xcode.  In particular, Xcode versions before 9.0 are known to
+definitely not work. You'll also need the GNU coreutils package from Homebrew,
+installable via `brew install coreutils`.)
 
 (If using **Windows**: The step-by-step instructions below should work
 fine on Windows.  Alternatively if you'd like a richer command-line
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -6,8 +6,10 @@ This guide describes how to build and run the app so you can develop it.
 
 (First, if using **macOS**: Upgrade to the latest version of the OS and then to
 the latest Xcode.  In particular, Xcode versions before 9.0 are known to
-definitely not work. You'll also need the GNU coreutils package from Homebrew,
-installable via `brew install coreutils`.)
+definitely not work.  You'll also need [GNU coreutils][] installed, e.g.
+with `brew install coreutils`.)
+
+[GNU coreutils]: https://www.gnu.org/software/coreutils/
 
 (If using **Windows**: The step-by-step instructions below should work
 fine on Windows.  Alternatively if you'd like a richer command-line
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -25,7 +25,7 @@ EOF
 
 # print error and die
 err() {
-    echo "$0:" $'\e[31merror:\e[0m' "$1" >&2
+    echo -e "$0:" $'\e[31merror:\e[0m' "$@" >&2
     exit 1;
 }
 
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -29,8 +29,6 @@ err() {
     exit 1;
 }
 
-
-
 ################################################################################
 # Parameters and environment
 ################################################################################
@@ -38,6 +36,8 @@ err() {
 # chdir to the current git repo's root
 ROOT="$(git rev-parse --show-toplevel)"
 cd "$ROOT"
+
+. "$ROOT"/tools/lib/ensure-coreutils.sh
 
 # Parse arguments. Sloppy, but sufficient for now.
 unset TARGET
@@ -61,6 +61,9 @@ if  [ -z "${TARGET-}" ]; then
 elif [ -z "${DEST-}" ]; then
     usage; exit 2
 fi
+
+# Make $DEST absolute, if it isn't already.
+DEST=$(readlink -m "$DEST")
 
 # Sanity-check DEST and TARGET.
 case "$TARGET" in
--- a/tools/lib/ensure-coreutils.sh
+++ b/tools/lib/ensure-coreutils.sh
@@ -4,17 +4,34 @@
 # Ensures GNU coreutils are available in PATH.
 # May add to PATH, or exit with a message to stderr.
 #
-# We don't yet have good docs for handling failure here if on
-# Windows or macOS, so until then this isn't suitable for scripts
-# used in normal development -- only for maintainer-facing scripts.
+# The GNU coreutils are a basic piece of infrastructure for having
+# a reasonable 21st-century shell scripting environment, so we
+# freely invoke this in any of our scripts that need `readlink -f`
+# or other features not always found without them.
+#
+# On any GNU/Linux system this is of course a non-issue.  Likewise
+# on Windows: we inevitably require Git, and Git for Windows comes
+# with a GNU environment called "Git BASH", based on MSYS2.
+#
+# So this is really all about macOS.  Fortunately it's easy to get
+# coreutils installed there too... plus, many people already have
+# it installed but just not in their PATH.  We write our scripts
+# for a GNU environment, so we bring it into the PATH.
 
 check_coreutils() {
+    # Check a couple of commands for GNU-style --help and --version,
+    # which macOS's default BSD-based implementations don't understand.
     fmt --help >/dev/null 2>&1
+    readlink --version 2>&1 | grep -q GNU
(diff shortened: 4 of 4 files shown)


tools/build-webview
```diff
@@ -25,10 +25,31 @@ EOF
...
+            # The BSD-based readlink on OSX doesn't have an equivalent to the
+            # GNU `-f` option. Check for `greadlink` instead, and die if not
+            # found.
+            if type greadlink &> /dev/null; then
+                greadlink -m $1
+            else
+                err "\`greadlink\` not found; please install GNU coreutils: " \
+                    "\n   brew install coreutils"
+            fi
+        ;;
```
gnprice: Take a look at tools/lib/ensure-coreutils.sh :smile: 

Given the insight that this is not a problem in the first place on Windows, I'm happy to see coreutils become a requirement for normal development (cf the comment in that file). Probably most people who clone this repo on macOS have it installed somewhere anyway, via homebrew or otherwise.

gnprice:
(Looks like this will also subsume #3645.)

android/app/webviewAssets.gradle
```diff
@@ -7,6 +10,45 @@
  * more information.
  */
 
+/**
+ * Windows support functions.
+ *
+ * On Windows, we have three different possible setups:
+ *   1. Windows Gradle + MSys Git's bash.
```
gnprice: Naming nit: "MSys Git" doesn't appear to be quite the name of anything.

There was an "msysGit": https://github.com/msysgit/msysgit whose name often got regularized to "msysgit". Looks like that's a few years obsolete, though. The new stack includes things named "Git for Windows"; "Git BASH"; "the Git SDK"; and ["MSYS2" or "msys2"](http://www.msys2.org/).

Does the "msys"/"MSYS" name ever show up in actually interacting with this stack (as far as you've seen)? Or are you just remembering, like me, the days when "msysgit" was the way people got Git on Windows?

I've mainly been referring to this GNU-like environment as "Git Bash", just regularizing the capitalization of "Git BASH" which seems to be what it calls itself for users.

I've seen enough references over the years to ~KSplice~ (the actual name was and is Ksplice) that in any case I'd like to not add internal capitals to other people's names. :wink:

android/app/webviewAssets.gradle
```diff
@@ -7,6 +10,45 @@
...
+ * Case 3, fortunately, should behave exactly like Linux.
+ *
+ * In cases 1 and 2, however, all paths Gradle provides are backslash-separated.
+ * This is problematic on its own, as the bash script we're invoking doesn't
+ * grok backslashes, so we normalize those on Windows. That's enough for case 1.
+ *
+ * Case 2 is worse, though. MSys's `bash` groks both "C:/" and "C:\"; more
+ * generally, it operates on the Windows filesystem with Windows path names. The
+ * WSL `bash` doesn't -- it's a standard Linux `bash` (or close enough), and it
+ * only sees the Windows disks as ordinary mounts under the lone volume '/'.
```
gnprice: I had to read this line a couple of times to understand what's meant by "the lone volume '/'". I think in the world where '/' is a thing, it doesn't make a lot of sense to call it a "volume" -- if anything that'd refer to one of the several things that might get mounted as filesystems at various paths in the overall path hierarchy that begins at '/'.

Would "ordinary mounts under the Unix-style single root path '/'" express what you mean here?

Or "only sees the Windows disks as ordinary mounts at paths like '/mnt/c/'" might add some helpful further concreteness.

android/app/webviewAssets.gradle
```diff
@@ -7,6 +10,45 @@
...
+def normalizePath(String path) {
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        return path.replace('\\', '/')
+    } else {
+        return path
+    }
+}
+
+def relativizePath(String path, String src) {
+    return Paths.get(src).relativize(Paths.get(path)).toString();
```
gnprice: It's not obvious to me from the names which of these parameters is expected to be which.

Perhaps something like "path" and "base"? Or a line of javadoc could do it, too.

android/app/webviewAssets.gradle
```diff
@@ -7,6 +10,45 @@
  * more information.
  */
 
+/**
```
gnprice: Nit: This is formatted to make it be javadoc for `normalizePath`, but it's really not documentation for either of these helpers (it doesn't describe their interface).

Writing a plain `/*` would be a sufficient fix.

android/app/webviewAssets.gradle
```diff
@@ -7,6 +10,45 @@
...
+ * And, of course, Gradle has no good way of distinguishing between cases 1 and
+ * 2. As far as it's concerned, `bash.exe` is `bash.exe`.
+ *
+ * Since we can't predict the environment of our callee, we can't use absolute
+ * paths. Fortunately, with only a little extra work in the bash script, we
+ * _can_ use relative paths, letting the script derelativize them appropriately.
+ */
+def normalizePath(String path) {
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        return path.replace('\\', '/')
```
gnprice: Hmm... is there a possible case where there actually is already a `/` in one of the path components? :grin: If so, that will surely cause confusion down the line.

Not sure there'd be anything sensible to do with that other than throw an error.

docs/howto/build-run.md
```diff
@@ -4,9 +4,10 @@ This guide describes how to build and run the app so you can develop it.
 
 ## Main steps
 
-(First, if using **macOS**: Upgrade to the latest version of the OS and then
-to the latest Xcode.  In particular, Xcode versions before 9.0 are known to
-definitely not work.)
+(First, if using **macOS**: Upgrade to the latest version of the OS and then to
+the latest Xcode.  In particular, Xcode versions before 9.0 are known to
+definitely not work. You'll also need the GNU coreutils package from Homebrew,
+installable via `brew install coreutils`.)
```
gnprice: I hesitate to say "you must use Homebrew", given how bad Homebrew is.

The error message in `tools/lib/ensure-coreutils.sh` says
```
This script requires GNU coreutils.

Install from your favorite package manager, or from:
  https://www.gnu.org/software/coreutils/
```

Perhaps "You'll also need GNU coreutils installed, either from your favorite package manager or [upstream](https://www.gnu.org/software/coreutils/)."?

tools/build-webview
ray-kraesig: I did see that! Those are maintainer-only scripts, as you note – most contributors won't be mucking about in there to be confused by it – so I'm content to leave cleaning it out for another day and another PR.

(Also I've only checked `readlink`; if you need any other utilities from `coreutils`, they'll have to be verified separately.)

android/app/webviewAssets.gradle
ray-kraesig: MSYS is thus named on [its official page](https://www.mingw.org/wiki/MSYS), as is [msys2](https://www.msys2.org/). "MSys" is all my fault.

>  Or are you just remembering, like me, the days when "msysgit" was the way people got Git on Windows?

Mostly this.

android/app/webviewAssets.gradle
ray-kraesig: Nope! NTFS and FAT32 both disallow `0x2F` as a character in file and directory names.

docs/howto/build-run.md
ray-kraesig: We do already have a reference to `brew install coreutils` in debugging.md, and the [React Native instructions we reference](https://facebook.github.io/react-native/docs/getting-started.html) also give direct `brew` commands.

I'm not sure that's a battle worth fighting. :confused:

ray-kraesig:
:ping_pong:

ray-kraesig:
(No changes in the commit itself in that last push, just a mention of the issue it fixes.)

gnprice:
(I've pushed a version with the changes we just discussed in person. Still to read the other bits.)

docs/howto/build-run.md
gnprice: Lol, I'd forgotten these lines (which I wrote):

    (On macOS, your `date` command may not have this feature, because
    it's a version whose UI hasn't changed since the '80s.  You want
    GNU `date`.  `brew install coreutils` will install it, with the name
    `gdate`.)

That still feels to me quite different from "you need the thing from Homebrew". It says "you need this thing; one way to get it is this Homebrew command".

I'm not particularly interested in taking on the battle of fighting Homebrew-normativity in other projects. :stuck_out_tongue: But I'd like to avoid contributing to it ourselves, and I think we can pretty easily accomplish that.

(OTOH I might at some point make the effort to switch to something else on our own office Mac, at least. Might leave Homebrew around just to refer to how it works when helping a contributor debug their setup... but could eliminate actually relying on it.)

android/app/webviewAssets.gradle
gnprice: Great! Good to know, thanks.

A one-line comment to that effect might be nice in case someone has the same question (though I'm sure many people already know it well.)

tools/lib/ensure-coreutils.sh
```diff
@@ -4,17 +4,34 @@
...
-# Windows or macOS, so until then this isn't suitable for scripts
-# used in normal development -- only for maintainer-facing scripts.
+# The GNU coreutils are a basic piece of infrastructure for having
+# a reasonable 21st-century shell scripting environment, so we
+# freely invoke this in any of our scripts that need `readlink -f`
+# or other features not always found without them.
+#
+# On any GNU/Linux system this is of course a non-issue, and nor
+# is it on Windows: we inevitably require Git, and Git for Windows
+# comes with a GNU environment called "Git BASH", based on MSYS2.
```
ray-kraesig: `s/and nor/nor/;`

tools/build-webview
```diff
@@ -60,6 +37,8 @@ absolutify() {
 ROOT="$(git rev-parse --show-toplevel)"
 cd "$ROOT"
 
+. "$ROOT"/tools/lib/ensure-coreutils.sh
```
ray-kraesig: This could use a comment, I think.

docs/howto/build-run.md
gnprice: > Perhaps "You'll also need GNU coreutils installed, either from your favorite package manager or [upstream](https://www.gnu.org/software/coreutils/)."?

Or a version that mentions `brew install` without saying Homebrew is mandatory could be

"You'll also need [GNU coreutils](https://www.gnu.org/software/coreutils/) installed, e.g. with `brew install coreutils`."

gnprice:
Thanks for the updates!

All looks good modulo the changes I added to the branch (one independent commit, and one meant to be squashed in) and the one subthread above about how we say to get coreutils.

tools/lib/ensure-coreutils.sh
```diff
@@ -4,17 +4,34 @@
...
 ensure_coreutils() {
+    # If we already have it, then great.
     check_coreutils && return
 
+    # Homebrew provides names like `greadlink` on the PATH,
+    # but also provides the standard names in this directory.
+    # Try that.
     homebrew_gnubin=/usr/local/opt/coreutils/libexec/gnubin
     if [ -d "$homebrew_gnubin" ]; then
         export PATH="$homebrew_gnubin":"$PATH"
```
ray-kraesig: This is technically incorrect: `$PATH` may be empty. :upside_down_face:

tools/lib/ensure-coreutils.sh
gnprice: Hmm, somehow that version feels ungrammatical to me! Yet it's true "and nor" is not common.

I think it's related to the fact the first clause isn't really in the negative. If this version bugs you, I can reword.

tools/build-webview
gnprice: Could do, but: what would you like a comment to say? The obvious thing that comes to my mind is what the name already aims to say :smile: : "ensure we have coreutils".

For details in turn of what *that* means, I think it's going to be hard to compete in this spot with the comment at the top of that file itself.

tools/lib/ensure-coreutils.sh
gnprice: Hmmm -- I guess that can be possible :stuck_out_tongue: but it seems like a fairly pathological situation to be in.

Do you have an idiom for smoothly handling that situation? I am not sure I've ever seen a script that took account of that possibility. Could handle it if it doesn't add much complication... but I'm inclined to have a low threshold on that level of complexity. We also assume things like: you don't have a `readlink` that is actually some version from the '80s and lacks a feature like `-m`.

tools/lib/ensure-coreutils.sh
gnprice: (done)

gnprice:
> , so long as the build process has been tested on a Mac with `check_coreutils` forced to fail.

Glad this caused me to double-check!

I'd previously tried it on the office Mac, where `check_coreutils` fails the first time and succeeds after the PATH fix...

... but apparently only with the change to `build-webview` to *use* `ensure-coreutils.sh`, not with the changes subsequently made to the latter. Which were small, so small... but changing that `fmt` to `command fmt` apparently alters the `set -e` behavior, or something. Oh Bash.

I'll debug that and make a fix.

gnprice:
> changing that `fmt` to `command fmt` apparently alters the `set -e` behavior, or something. Oh Bash.

... It appears to be a bug in the ancient `bash` that Apple ships (because they hate freedom):
```
$ bash -ec 'bash --version | head -2; if command false; then echo t; else echo f; fi'
GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
f

$ ssh zulip-mac.local -- "bash -ec 'bash --version | head -2; if command false; then echo t; else echo f; fi'"
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

Best workaround looks to be to just drop the `command` bit. :-/ At least that should only matter in a fairly pathological setup anyway.

gnprice:
Based on your OK above, I've gone ahead and made the remaining changes and then merged!

Thanks for the PR and then the review :slightly_smiling_face:

tools/lib/ensure-coreutils.sh
ray-kraesig: I believe replacing the hard colon with `${PATH:+:}` would do it.

(Not worth the trouble here, mind, but it might be useful in other contexts.)

tools/build-webview
ray-kraesig: I'd been thinking something along the lines of "Ensure we have `readlink`" (rather than the moderately nebulous "coreutils").

tools/lib/ensure-coreutils.sh
gnprice: > (Not worth the trouble here, mind, but it might be useful in other contexts.)

:+1:

tools/build-webview
gnprice: > I'd been thinking something along the lines of "Ensure we have `readlink`" (rather than the moderately nebulous "coreutils").

Sure, that'd be reasonable.

OTOH now that we have a version of this we believe we *can* just count on for all dev scripts, I think it could be helpful to go ahead and add this to the top of basically all of them (that are in shell). Effectively that makes it part of our shell-scripting environment: we just write all our shell scripts in "Bash, with GNU coreutils". Which is a helpful simplification for thinking about editing any of these scripts -- if you need `readlink -f` or whatever, no need to consider whether this is one of the scripts that has coreutils or doesn't, just do it because they all have it.

PR #3671, opened 2019-10-29
notifications: normalize case of realm_uri and account.realm

rk-for-zulip:
`account.realm` is an only-lightly-normalized form of whatever the
user typed in; what little normalization is done does not include
case-folding.

On the other hand, based on errors in Sentry, the notification's
provided URL *is* case-normalized.

N.B.: case-normalization may be destructive to non-ASCII URLs or in
Turkish locales.

* Non-ASCII URL comparison is probably already broken, because we
  don't do punycode encoding on our end.

* Turkish locale users with ASCII realm URLs should be mostly
  unaffected, unless they have typed a capital (dotless) I in their
  realm URL. In this case they will need to replace it with a
  lowercase (dotted) i.

### DIFF
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -59,9 +59,18 @@ export const getAccountFromNotificationData = (
     return null;
   }
 
+  // The `realm_uri` we get from the server should be case-normalized already;
+  // but better safe than sorry.
+  const normalized_realm_uri = realm_uri.toLowerCase();
   const urlMatches = [];
   identities.forEach((account, i) => {
-    if (account.realm === realm_uri) {
+    // `account.realm` is _not_ case-normalized; it's straight from user input,
+    // which (judging from error reports) autocorrect seems to be forcibly
+    // autocapitalizing on some devices.
+    //
+    // (TODO: instead of performing case-mangling, just compare against the
+    // value of `realm_uri` we get from `/server_settings` at login time.)
+    if (account.realm.toLowerCase() === normalized_realm_uri) {
       urlMatches.push(i);
     }
   });


ray-kraesig:
... and I've forgotten to mark this as a draft PR – it's untested! :fearful: (Please avoid actually merging for now.)

gnprice:
Thanks for tracking this down!

The realm URL that appears in the notification data is the one the server has for itself (for that realm).

Thinking about... you know what, I'm going to move this [to chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/realm.20URL/near/795201). :slightly_smiling_face:

ray-kraesig:
It's worth noting that the right fix here is probably not to do any sort of URL mangling: it's to ask the server _what it calls itself_ at login (or thereabouts), then use that as a comparand. We've been getting a number of Sentry errors indicating that people are connecting to renamed servers.

gnprice:
Sorry, it seems I dropped this thread. Added more just now [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/realm.20URL/near/800766) after you pinged me in person; thanks!

Sounds like we both agree on the right direction for a complete fix: we'd drop any attempt at canonicalizing the URL ourselves, and just quote the server's idea of it from `server_settings`.

That shouldn't be super complicated, but this commit is very simple, so it's a fine quick mitigation. Code LGTM; my only comments are
* adjust the TODO comment to reflect the new plan
* adjust the commit message ditto

ray-kraesig:
:ping_pong:

gnprice:
We're now accomplishing this by parsing both URLs as `URL` objects, with ebbab7e99 and previously #4235 .

In the chat thread linked above, we'd discussed that as the ideal way to solve this problem. At the time we didn't have a useful `URL` implementation handy, because RN provides only a skeleton of one. Since #4081, we have a suitable `URL` implementation as a polyfill, and this is one of many handy applications for it.

Issue #3670, opened 2019-10-28
Support data-driven auth methods, e.g. SAML


gnprice:
The server is gaining a new API feature to let `/server_settings` describe auth methods with a bit more detail, making it possible to use auth methods that aren't in a hardcoded list in the client code.

The immediate use case for this is SAML authentication, where there's a single protocol that may have any number of providers with different names, so that it makes less sense than it already does for the client to have to hardcode the possible names in advance.

The server-side implementation just landed as zulip/zulip#13307. Docs are said to be coming soon; previous discussion [in chat here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SAML.20authentication/near/790808).

It would be good for us to implement our end soon, in order to better end-to-end validate the API choices before they're made permanent by a server release.

We'll need to keep the existing code for backwards-compatibility with older Zulip servers, but should plan to ignore the old data if the new key is present in the `/server_settings` data.

/cc @timabbott @mateuszmandera

timabbott:
The docs updates were merged in https://github.com/zulip/zulip/pull/13339.

gnprice:
I'm going through this now to implement the new part of the `server_settings` API -- just for the existing methods like Google and GitHub auth -- since that seems to be up now on chat.zulip.org.

@timabbott what's a way we can test actual SAML auth? I think you said something about a test machine on a DigitalOcean droplet.

Some thoughts on the API as I have my hands in it:
* The name `social_backends` for this property is pretty odd:
  * These aren't particularly "social". Google and GitHub auth are marginally so; auth with AzureAD, or `remoteuser` SSO, or a SAML provider like Okta, is pretty much exclusively about authenticating as a corporate employee, or another institutional affiliate.
    * I know there's a library we're using with a name like "django-social-auth", but that inaptly-named dependency within the server seems like very much an implementation detail from any client's perspective.
  * These aren't "backends" in any obviously meaningful sense. They're methods for authenticating, or providers of authentication or identity... but the very next thing we're going to do with any of these is go make an HTTP request from a browser. Which will typically redirect to some other site, for the user to go enter their credentials into a web page. None of that seems particularly more backend-y than basically anything else a user might do on the Internet.
[...]

PR #3674, opened 2019-11-01
Clean up auth code in preparation for adding SAML

gnprice:
These are all pure refactors, or tests.

They came as I was refactoring this code to make it suitable to incorporate the new data-driven list of auth methods (#3670) which we'll use for SAML... and then as I studied it harder and unraveled what some parts of it were really doing, and sought to make the names and organization better match that.


src/start/AuthScreen.js
```diff
@@ -19,52 +19,52 @@ import { generateOtp, openBrowser, closeBrowser } from './oauth';
 import { loginSuccess, navigateToDev, navigateToPassword } from '../actions';
 
 type AuthenticationMethodDetails = {|
-  method: string,
   name: string,
+  displayName: string,
   Icon: IconType,
   handler: string,
 |};
```
ray-kraesig: Given the commit message, it sounds like `'id'` or `'methodName'` might be a better key than `'name'`.

(Either way, these two items could use docstrings, or at least docstringishly descriptive comments.)

src/utils/__tests__/encoding-test.js
```diff
@@ -78,11 +77,3 @@ describe('base64Utf8Encode', () => {
...
   });
 });
-describe('extractApiKey', () => {
-  test('correctly extracts an API key that has been XORed with a OTP', () => {
-    const key = 'testing';
-    const otp = 'A8348A93A83493';
-    const encoded = xorHexStrings(asciiToHex(key), otp);
-    expect(extractApiKey(encoded, otp)).toBe(key);
-  });
-});
```
ray-kraesig: This test was removed, but doesn't seem to have been replaced in a new location.

src/start/AuthScreen.js
gnprice: Yeah, I agree `name` is ambiguous on identifier-like name vs. display name. I think it helps that the other name in the pair is unambiguous, though.

To me `id` would suggest a numeric ID -- like our user IDs, realm IDs, stream IDs, etc. Realms and streams have identifier-like names, even, but then also "IDs" which are numbers. I think `methodName` doesn't add any information beyond `name`; to the extent an auth method's "name" could mean its display name, "method name" could just as well. So I don't think either of those would help.

Some jsdoc is a good idea. I'll add some.

src/utils/__tests__/encoding-test.js
gnprice: Thanks. That was intentional, but I should mention it in the commit message. (The test just recites the implementation, and I didn't see value in keeping it. Once `authFromCallbackUrl` gets some tests a couple of commits later, that effectively includes a new test of this code that actually specifies the input.)

src/start/AuthScreen.js
ray-kraesig: > To me `id` would suggest a numeric ID -- like our user IDs, realm IDs, stream IDs, etc.

Ah, that's understandable. And noted. (The meaning I was thinking of was from the DOM:  `document.getElementById('blorple')`, etc.)

gnprice:
Thanks @ray-kraesig for the review! Merged, with revisions from your comments.

src/start/AuthScreen.js
gnprice: > (The meaning I was thinking of was from the DOM: `document.getElementById('blorple')`, etc.)

Ah, makes sense! Yeah, within Zulip I think we pretty consistently say "ID" only for a numeric ID.

The one exception I can think of is within the server codebase -- we've managed to keep it out of the API -- where `Realm.string_id` is an identifier-style string, better known as the "subdomain".

PR #3677, opened 2019-11-05
webview [nfc]: demystify the relative base URL

ray-kraesig:
Make it clear who's doing the resolution, and with respect to what.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -105,35 +105,36 @@ export type Props = {|
 |};
 
 /**
- * The path to the webview assets, represented as a `file:`-scheme URL.
+ * A URL-like magic string denoting the webview-assets folder.
  *
- * This may be a relative URL, relative to whatever base URL a `WebView`
- * component will use to interpret its own `source.baseUrl` prop.  (If
- * relative, the scheme naturally is left implicit.)
- *
- * See the `tools/build-webview` script for more details.
- *
- * Because this URL may be relative, it shouldn't be used anywhere *other*
- * than the `source.baseUrl` of a `WebView` component.  In particular, it
- * shouldn't be used for URLs in the browser environment within the
- * `WebView`; to reach the webview assets there, just use `./`.
+ * Not suitable for reuse or export.
  */
-// On iOS, a typical value to use here might be expressed like (in Swift)
-// `Bundle.main.resourceURL`.  But there's no convenient way to get that
-// from inside RN (and it doesn't seem worth linking something like
-// react-native-fs for); and unlike on Android, it's not fixed enough to be
-// able to just hardcode.
-//
-// Fortunately, on iOS the `WebView` will accept a relative URL for this
-// value!  This behavior seems a bit mysterious -- it doesn't appear to be
-// documented -- but the mysterious base URL for the base URL takes a
-// convenient enough value we can just use that.  (The `WebView`
-// implementation on iOS passes `source.baseUrl` straight through to
-// `baseURL` here:
-//   https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring
-// but there's no indication that that can be relative, or if so what it is
-// treated as relative to.)
 const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/webview';
+// What the value above probably _should_ be, semantically, is an absolute
+// `file:`-scheme URL naming the path to the webview-assets folder. [1]
+//
+// * On Android, that's exactly what it is. Different apps' WebViews see
+//   different (virtual) root directories as `file:///`, and at runtime, the
+//   APK's `assets/` directory is mounted at `file:///android_asset/`. We can
+//   easily hardcode that, so we do.
+//
+// * On iOS, it's not: it's a relative path. (Or relative URL, if you prefer.)
+//   We can't make it absolute here, because neither React Native itself nor any
+//   of our current dependencies directly expose the Foundation API that would
+//   tell us the absolute path that our bundle is located at [2].
+//
+//   Instead, for now, we exploit the fact that (the iOS version of) React
+//   Native will resolve it with respect to the bundle's absolute path. [3] This
+//   fact is not known to be documented, however, and should not be taken for
+//   granted indefinitely.
+//
+// [1] See `tools/build-webview` for more information on what folder that is.
+//
+// [2] Specifically, `Bundle.main.bundleURL` (aka `[[NSBundle mainbundle]
+//     bundleURL]`). Alternatively, the value of `resourceURL` is (believed to
+//     be) equivalent in this context.
+//
+// [3] https://github.com/facebook/react-native/blob/0.59-stable/React/Base/RCTConvert.m#L85
 
 class MessageList extends Component<Props> {
   context: Context;


src/webview/MessageList.js
```diff
@@ -105,34 +105,32 @@ export type Props = {|
 |};
 
 /**
- * The path to the webview assets, represented as a `file:`-scheme URL.
+ * The path to the webview assets, represented as a `file:`-schema URL. This may
+ * be a relative URL; if so, it is relative to the bundle or package root, and
+ * the scheme is necessarily omitted as implicit. [1]
```
gnprice: A bit of comment-writing style: I find it quite helpful to try to have a one-line summary at the beginning, as its own paragraph. Here's a style guide that has a nice writeup of this principle:
https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

Here, I think the first sentence will work fine for that, so just adding a blank line suffices.

src/webview/MessageList.js
```diff
@@ -105,34 +105,32 @@ export type Props = {|
 |};
 
 /**
- * The path to the webview assets, represented as a `file:`-scheme URL.
+ * The path to the webview assets, represented as a `file:`-schema URL. This may
```
gnprice: Jargon nit: the term [is "scheme"](https://url.spec.whatwg.org/#concept-url-scheme), with an "e" at the end.

src/webview/MessageList.js
```diff
@@ -105,34 +105,32 @@ export type Props = {|
...
- * This may be a relative URL, relative to whatever base URL a `WebView`
- * component will use to interpret its own `source.baseUrl` prop.  (If
- * relative, the scheme naturally is left implicit.)
+ * Because this URL may be relative, it shouldn't be used for URLs in the
+ * browser environment _within_ the `WebView`; to reach the webview assets
+ * there, just use `./`.
  *
- * See the `tools/build-webview` script for more details.
+ * [1] Unlike some other URL schemas (notably `https:`), `file:` URLs don't
+ *     support use of an explicit schema with a relative path.
```
gnprice: Huh, does `https:` really support this? I'm not finding it [in the spec](https://url.spec.whatwg.org/#url-writing), or in some brief experimentation.

src/webview/MessageList.js
```diff
@@ -105,34 +105,32 @@ export type Props = {|
 |};
 
 /**
- * The path to the webview assets, represented as a `file:`-scheme URL.
+ * The path to the webview assets, represented as a `file:`-schema URL. This may
+ * be a relative URL; if so, it is relative to the bundle or package root, and
```
gnprice: > the bundle or package root

I just spent a few minutes studying to see if I can unpack what this means. I think what's meant here is
* on iOS, the root directory of the app's bundle
  * `foo.bundleURL` is ["[t]he full URL of the receiver’s bundle directory"](https://developer.apple.com/documentation/foundation/bundle/1415654-bundleurl)
  * Looking in particular into what "bundle" means -- it's apparently [a pretty generic Apple concept](https://developer.apple.com/documentation/foundation/bundle). There might be several bundles floating around; `Bundle.main` gets specifically the one that contains our app.
* on Android... not sure. What's "the package root" mean concretely? And if we did use a relative URL here, would it get interpreted relative to that?
  * If it isn't trivial to definitively answer this, probably cleanest to just not try to generalize the description in that direction.

src/webview/MessageList.js
```diff
@@ -105,34 +105,32 @@ export type Props = {|
...
-// documented -- but the mysterious base URL for the base URL takes a
-// convenient enough value we can just use that.  (The `WebView`
-// implementation on iOS passes `source.baseUrl` straight through to
-// `baseURL` here:
-//   https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring
-// but there's no indication that that can be relative, or if so what it is
-// treated as relative to.)
+// [α] Specifically, `Bundle.main.bundleURL` (aka `[[NSBundle mainbundle]
+//     bundleURL]`). Alternatively, the value of `resourceURL` is equivalent in
+//     this context.
```
gnprice: Huh interesting -- are you saying that `resourceURL` gets the same value as `bundleURL`?

src/webview/MessageList.js
```diff
@@ -105,35 +105,33 @@ export type Props = {|
...
+//
+// * On iOS, it's not: it's a relative path. We can't make it absolute here,
+//   because neither React Native itself nor any of our current dependencies
+//   directly expose the Foundation API that would tell us the absolute path
+//   that our bundle is located at [2].
+//
+//   Instead, for now, we exploit the fact that (the iOS version of)
+//   react-native-webview will resolve a relative path with respect to the
+//   bundle's absolute path. This fact is not known to be documented, however,
+//   and should not be taken for granted indefinitely.
```
gnprice: A small remaining quibble: I think the two mentions of a "relative path" here would be more accurate to say "relative URL". E.g. [here's](https://github.com/react-native-community/react-native-webview/blob/v5.12.1/ios/RNCWKWebView.m#L371) where that conversion happens:
```
- (void)visitSource
{
    // Check for a static html source first
    NSString *html = [RCTConvert NSString:_source[@"html"]];
    if (html) {
        NSURL *baseURL = [RCTConvert NSURL:_source[@"baseUrl"]];
        if (!baseURL) {
            baseURL = [NSURL URLWithString:@"about:blank"];
        }
        [_webView loadHTMLString:html baseURL:baseURL];
        return;
    }
```
Specifically the `RCTConvert NSURL:` call. So that code certainly seems to want to think of it as a URL.

src/webview/MessageList.js
```diff
@@ -105,35 +105,33 @@ export type Props = {|
 |};
 
 /**
- * The path to the webview assets, represented as a `file:`-scheme URL.
+ * A pathlike magic string denoting the webview-assets folder.
```
gnprice: I appreciate how this very effectively makes clear that it's not making any particularly general promises. :smile: 

A small remaining quibble: rather than "pathlike", I think it might fit better to call it "path- or URL-like", or even just "URL-like".
* The way it's used is basically to hand it to something that expects a URL -- i.e. as a property called `baseUrl` -- so given that that makes sense, it seems like it must be accurate to think of this string as URL-like, whatever else it might be like.
* As the comment below says, the thing this would ideally be is a URL. A `file`-scheme URL (and with no host, which apparently `file`-scheme URLs can have) -- so certainly a URL that is pretty pathlike.
* In the case where it's not overtly a URL, it's a relative path... which also qualifies as a path-relative URL, so describing that as a URL is consistent too.

src/webview/MessageList.js
gnprice: > I think that's at least slightly too short: "relative" is potentially surprising key information if you're thinking of it as a URL.

That's fair.

Just as an exercise in docs-writing style, then, because this is moot given the other changes: a version that would be almost as short as I'd like, and would take care of that, might be

> The path to the webview assets, represented as a possibly-relative `file:`-scheme URL.

I.e. just add "possibly-relative" to the existing version. (As well as any desired information at greater length in the prose that follows.) I'd be pretty OK with a summary line like that.

If writing that, I'd probably go on to tighten it a bit: s/represented // would get most of the way, and a final version might be:

> Path to the webview assets, as a possibly-relative `file:` URL.

src/webview/MessageList.js
gnprice: > ... but that seems to be because validation errors don't actually terminate URL parsing. (See also line 5 in the table at [the top of that section](https://url.spec.whatwg.org/#urls).)

Hmm! Good to know.

src/webview/MessageList.js
gnprice: > but the documentation [here](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW17) at least strongly implies that this is true for all iOS apps.

Ah indeed, thanks!

I think in fact combining what that says:
> In an iOS application, nonlocalized resources are located at the top-level of the bundle directory, along with the application’s executable file and the Info.plist file.

with the doc'd meaning [of `resourcePath`](https://developer.apple.com/documentation/foundation/nsbundle/1417723-resourcepath):
> The full pathname of the bundle’s subdirectory containing resources.

pretty clearly says that for an iOS application, the `resourcePath` will be the same as the `bundlePath`.

---

> the location of the "unbundled" bundle

Ah but I think this isn't the right way to describe that directory -- rather, the directory *is* the bundle. A bundle is a particular kind of directory. It's "bundled" just by virtue of having that ".app" at the end of its name.

How a bundle gets serialized up into a single file is another matter.

I started writing some more details and then decided it'd be good to put them somewhere more reusable, in chat:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/apple.20app.20bundles/near/796432

gnprice: > (The `assets` directory which is also present in that folder is a React Native thing... and also an Android thing, because it remounts that directory as `file:///android_asset/`. I don't know why.)

Yeah, I think:
* the `assets` directory itself is an Android thing ([docs](https://developer.android.com/guide/topics/resources/providing-resources)), and RN is just going along with that
* the `file:///android_asset/` thing is probably a WebView special? Not sure. Discussion [here in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/android.20filesystem/near/796440).

src/webview/MessageList.js
gnprice: > An iOS app bundle and an Android APK are, on some level, the same thing – they're a directory structure with a bunch of files: some executable, some resources, some metadata. An APK is a zip file of that structure, while... well, I assume there's some similar packaging mechanism for iOS, but I have no idea how it's customarily serialized.
> 
> What I mean here is the root of that directory structure, _as it is mounted at runtime_.

Yeah, so in the Android case I think there's a difference between
* the APK, which is a zip file used as a distribution format for the app
* the various directories belonging to the app once installed

I've been fuzzy on the details... so I went and investigated, details [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/android.20filesystem/near/796411).

One thing that seems to come out of that is that `/android_asset/` doesn't seem to be any kind of directory in the real filesystem at all! My best guess is that URLs under `file:///android_asset/` are specially interpreted by the WebView implementation (meaning the underlying Android WebView.)

And in comparing the iOS and Android cases, it looks like:
* an iOS bundle is an actual directory
* but on Android when an app is installed, the APK mostly doesn't get unpacked, nor mounted as a filesystem -- so with a few exceptions, there's no actual directory in the filesystem where the files inside it live. Instead, you get at the things inside it through a variety of APIs like [`Resources`](https://developer.android.com/reference/android/content/res/Resources.html) and [`AssetManager`](https://developer.android.com/reference/android/content/res/AssetManager).

gnprice:
(Also made some longer comments on delving into the semantics of paths and filesystems and so on on iOS and Android; visible upthread in replies to the relevant previous comments. But those are now all moot for this particular change, I think.)

src/webview/MessageList.js
```diff
@@ -105,35 +105,33 @@ export type Props = {|
...
-//   https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring
-// but there's no indication that that can be relative, or if so what it is
-// treated as relative to.)
 const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/webview';
+// What the value above probably _should_ be, semantically, is an absolute
+// `file:`-scheme URL naming the path to the webview-assets folder. [1]
+//
+// * On Android, that's exactly what it is. Different apps see different root
+//   directories for `file://`, and at runtime, the APK's `assets/` directory is
+//   mounted at `file://android_asset/`. We can easily hardcode that; so we do.
```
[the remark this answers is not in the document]
ray-kraesig: Done. (Also limited claims about `android_asset` to being WebView-specific.)

src/webview/MessageList.js
ray-kraesig: I would say it instead occurs [within that RCTConvert NSURL](https://github.com/facebook/react-native/blob/0.59-stable/React/Base/RCTConvert.m#L85), which uses the label "path" for its input. Admittedly, that's React Native rather than `react-native-webview` specifically.

Amended, with a link to the conversion code included.

> I think the two mentions of a "relative path" here would be more accurate to say "relative URL".

Amended to be agnostic on the matter.

src/webview/MessageList.js
ray-kraesig: Done, as "URL-like".

src/webview/MessageList.js
ray-kraesig: I'm somewhat surprised that APKs are never mounted into the underlying filesystem proper... but ultimately I see that as an irrelevant implementation detail. The filesystem seen by the WebView (or WebView-interpreted code) under `file://` may be mostly or entirely virtual, but it's still a (read-only) filesystem.

> My best guess is that URLs under `file:///android_asset/` are specially interpreted by the WebView implementation (meaning the underlying Android WebView.)

I can't find real documentation for that, but there are a couple of Chromium bugs where developers have given offhanded confirmation of this:
* https://bugs.chromium.org/p/chromium/issues/detail?id=346718
* https://bugs.chromium.org/p/chromium/issues/detail?id=387898

src/webview/MessageList.js
gnprice: > there's no possibility that it could include search-parameters or a fragment; and this fact is, technically, exploited.

If "pathlike magic string" is interpreted broadly enough to not preclude a leading `file://` before the actual path part, I don't think it precludes some trailing `?foo#bar` either :wink: So I don't think "pathlike" really helps with that.

If it were important to be explicit about this aspect, I think I'd probably add a line in the prose of the jsdoc (after the summary line) saying something like "The URL has null query and fragment."

But as you say, the literal value is right there, so I think it's fine as is.

src/webview/MessageList.js
```diff
@@ -105,35 +105,36 @@ export type Props = {|
...
-//   https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring
-// but there's no indication that that can be relative, or if so what it is
-// treated as relative to.)
 const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/webview';
+// What the value above probably _should_ be, semantically, is an absolute
+// `file:`-scheme URL naming the path to the webview-assets folder. [1]
+//
+// * On Android, that's exactly what it is. Different apps' WebViews see
+//   different (virtual) root directories as `file:///`, and at runtime, the
+//   APK's `assets/` directory is mounted at `file:///android_asset/`. We can
```
gnprice: The remaining semantic quibble I have is: to me this still says this `assets/` directory is in an actual filesystem, one that could be accessed with normal filesystem APIs.

I agree (following on [this](https://github.com/zulip/zulip-mobile/pull/3677#discussion_r344965636) comment thread elsewhere in this PR) that the assets are presented here in an interface that's abstractly *like* a filesystem in some ways. But this reads to me as saying something more concrete than that.

I think in particular
* "mount" has to me a pretty specific jargon meaning when talking about paths and directories: it means putting the thing in the filesystem as the kernel sees it, i.e. as seen through syscalls like `open`.
* Saying "that's exactly what it is" seems like a clear statement that this description isn't a metaphorical one.

src/webview/MessageList.js
gnprice: > I can't find real documentation for that, but there are a couple of Chromium bugs where developers have given offhanded confirmation of this:

Cool -- I copied that finding over to the linked chat thread :)

gnprice:
> I might then send a further followup change.

Done, as #3686.

Issue #3242, opened 2018-12-27
Can't use the app offline (when there's no internet connection)


zingmars:
When trying to access private messages or streams when my phone either has a bad (or no) internet connection internet connection I can't see any messages that might be there (nothing is cached). It just gives me the loading screen spinner. 
I.e. - 
If I had the private message view open before I disconnected, I can still view the messages and switch between streams and private messages if I want, but if I close the app while outside the private message view (meaning it will work if I close the app while viewing my PMs) and re-open it again I will no longer be able to view them and get the spinning circle again.

![image](https://user-images.githubusercontent.com/917126/50488041-4a5b6d80-0a09-11e9-9476-b0ecea365787.png)

borisyankov:
Thanks for reporting this.
It is expected and somewhat intended behavior but I think we can improve this.

zingmars:
Can't say I'm a fan of this design. What if I want to view a conversation I've had while in a situation where internet isn't available?

zingmars:
As it is right now, it's a really horrible experience when roaming and in places that have a spotty network connection. Every other alternative seems to work just fine in these conditions. Implementing some sort of caching would be really useful even if it's read-only.

gnprice:
Thanks @zingmars for reporting this!

We've just fixed the PM-conversation-list issue in dfca21f81, thanks to @jainkuniya . As described in that commit message, there was a time when the loading-spinner was necessary because we weren't actually caching some of the important information we needed... but that time was over in mid-2018, so by the time of your original report we could have fixed it by just removing the spinner. Oops; but now we have.

You also mention trouble with streams, but @jainkuniya wasn't able to reproduce that part (and we don't have the same kind of loading-spinner logic there.) Perhaps that was fixed by some other change since your original report? Please try it out, and if you find there's still trouble outside the PM-conversation-list screen, we'd be glad to see a followup issue with details on that.

PR #3678, opened 2019-11-05
pmConversationCard: Do not show loader while initial fetch is going on.

jainkuniya:
Because users are already cached, so that conversation list can be shown
by getting users from it.

This loader was added in 06c4de8e110fa5939c83b949333fae37e82fcfb3, at
that time user list was not cached in the redux.

Fixes: #3241


jainkuniya:
Next I guess we can remove `users` from loading state as well, no body is using it.

gnprice:
Merged -- thanks!

> Next I guess we can remove `users` from loading state as well, no body is using it.

Sure. I'd in fact go further:
(a) reduce the whole `loading` state-subtree to just a single boolean;
(b) move that boolean inside the `session` tree.

gnprice:
Hmm but I think you meant #3242 for the issue this fixes. :wink:

PR #3680, opened 2019-11-06
Clean loading state and move it to session.

jainkuniya:
Follow up of dfca21f816ee5492105c6e4794a73d19402e3cc6.


src/session/sessionReducer.js
```diff
@@ -145,9 +148,16 @@ export default (state: SessionState = initialState, action: Action): SessionStat
...
+    case INITIAL_FETCH_START:
+      return {
+        ...state,
+        loading: true,
+      };
+
     case INITIAL_FETCH_COMPLETE:
       return {
         ...state,
+        loading: false,
```
jainkuniya: This commit is WIP, because not sure why we don't have session reducer test for this case. Should I add it?

src/loading/loadingReducer.js
```diff
@@ -1,32 +0,0 @@
...
-  unread: false,
-  users: false,
-};
-
-export default (state: LoadingState = initialState, action: Action): LoadingState => {
-  switch (action.type) {
-    case DEAD_QUEUE:
-    case LOGOUT:
-    case ACCOUNT_SWITCH:
-      return initialState;
```
gnprice: Note these cases -- be sure to preserve the behavior in these cases too.

src/session/sessionReducer.js
gnprice: Sure, why not.

src/loading/loadingReducer.js
jainkuniya: Yups, updated 👍

gnprice:
I just read the updated version at 98b91a9 -- the logic looks good!

I see you've still marked it as WIP. I'm guessing that's because you plan to add tests.

src/session/sessionReducer.js
jainkuniya: Hey, I have added few tests. Can you please review and let me know, whether I am going in the right direction.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -38,4 +38,39 @@ describe('sessionReducer', () => {
...
+
+  test('APP_ONLINE', () => {
+    const newState = sessionReducer(baseState, eg.action.app_online);
+    expect(newState).toEqual({ ...baseState, isOnline: true });
+  });
+
+  test('APP_STATE', () => {
+    const newState = sessionReducer(baseState, eg.action.app_state);
+    expect(newState).toEqual({ ...baseState, isActive: true });
+  });
```
gnprice: In each of these, notice that the test can't be understood by reading just the test -- it's dependent on the specific details of `eg.action.do_narrow` and so on.

E.g., why should we have `isOnline: true`? That's not a general fact about any `APP_ONLINE` action -- it's only when the action carries a `true` value.

By contrast in the existing `LOGIN_SUCCESS` test, notice that it completely doesn't matter for this test (or the code it's testing) what the specific data in the action is -- only that it's some `LOGIN_SUCCESS` action. That's the case where `exampleData` is super useful: providing some valid value of a given type, for tests where the details of the value don't matter.

For these tests where the specific data does matter, see the `START_EDIT_MESSAGE` test for a good pattern to follow.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -38,4 +38,39 @@ describe('sessionReducer', () => {
...
+
+  test('DEAD_QUEUE', () => {
+    const newState = sessionReducer(baseState, eg.action.dead_queue);
+    expect(newState).toEqual({ ...baseState, needsInitialFetch: true });
+  });
+
+  test('LOGOUT', () => {
+    const newState = sessionReducer(baseState, eg.action.logout);
+    expect(newState).toEqual(baseState);
+  });
```
gnprice: For these two and also `INITIAL_FETCH_COMPLETE` below, the action contains no data at all. I think it'd be cleaner to make that transparent -- see the `CANCEL_EDIT_MESSAGE` test for a model.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -1,9 +1,9 @@
-/* @flow strict-local */
```
gnprice: No, leave this here -- see our [test style guide](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/testing.md#test-style-guide).

I don't totally follow the commit message:

    So that there is no need to dump unwanted data in tests, and thus make
    tests clener.

but I think this might be describing what `exampleData` is the solution for.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -38,4 +38,39 @@ describe('sessionReducer', () => {
...
   });
+
+  test('DEAD_QUEUE', () => {
+    const newState = sessionReducer(baseState, eg.action.dead_queue);
+    expect(newState).toEqual({ ...baseState, needsInitialFetch: true });
+  });
+
+  test('LOGOUT', () => {
+    const newState = sessionReducer(baseState, eg.action.logout);
+    expect(newState).toEqual(baseState);
```
gnprice: It's not very informative to say that this has no effect when starting from this state :wink: . Better to start from a state where it does do something. See the `CANCEL_EDIT_MESSAGE` test for an example.

src/session/__tests__/sessionReducer-test.js
jainkuniya: Thanks for the link.

I was on completely on opposite page 😅

src/session/__tests__/sessionReducer-test.js
jainkuniya: Updated 👍

src/session/__tests__/sessionReducer-test.js
```diff
@@ -38,4 +60,149 @@ describe('sessionReducer', () => {
...
+  test('LOGOUT', () => {
+    const state = deepFreeze({
+      ...baseState,
+      lastNarrow: [],
+      needsInitialFetch: true,
+      loading: true,
+    });
+    const newState = sessionReducer(state, deepFreeze({ type: LOGOUT }));
+    expect(newState).toEqual(baseState);
+  });
```
gnprice: This looks reasonable; but when this test is first added, it tests not doing anything :slightly_smiling_face:

src/session/__tests__/sessionReducer-test.js
```diff
@@ -38,4 +60,149 @@ describe('sessionReducer', () => {
...
+
+  test('DO_NARROW', () => {
+    const state = deepFreeze({
+      ...baseState,
+      lastNarrow: [],
+    });
+    const action = deepFreeze({
+      type: DO_NARROW,
+      narrow: privateNarrow('a@a.com'),
+    });
```
gnprice: formatting nit: Here and on a bunch of others where the state and/or action are quite simple expressions, it'd be helpful to squash them on one line. Just makes things closer together, and easier to scan through.

Quick keyboard tip for making that change in VS Code: I go to the end of the first line (after the `{`), and hit delete, end, delete, end, delete, then Ctrl+S to save-and-format.

src/__tests__/exampleData.js
```diff
@@ -219,4 +219,107 @@ export const action = deepFreeze({
     email: selfAccount.email,
     apiKey: selfAccount.apiKey,
   },
+  realm_init: {
+    type: REALM_INIT,
+    data: {
+      last_event_id: 34,
+      msg: 'msg',
+      queue_id: 1,
```
gnprice: Yikes this is long :astonished: But that's how the reality is, so that's cool.

How did you get this particular set of values -- was it the result of some real request? ... Ah, no, that can't be in the case of `realm_uri` and some others.

One thing that doesn't matter in this test but I think would be helpful: use `selfAccount.realm` and similar values, to make this fit together consistently with the other example data.

src/session/__tests__/sessionReducer-test.js
gnprice: Aha, and then when the change is made in the app code, the diff here is bigger. So that would be good to separate.

Looks like ACCOUNT_SWITCH is similar.

gnprice:
Thanks for the revisions!

I've gone and merged most of this, as the following commits:
70d6b4d4a loading reducer [nfc]: Remove unused `users` property.
24e175831 session reducer: Improve test for ACCOUNT_SWITCH.
ae81101ca session reducer: Add test for REALM_INIT.
baadd8209 session reducer: Add tests for most actions.
318ca92d8 exampleData [nfc]: Add a check that `eg.action` has the right type.
6c5f752e0 session reducer [nfc]: More-compact formatting for tests.

This includes a couple of new commits of mine at the end, plus I squashed together all the similar tiny test-case commits.

I've pushed back to the branch the remaining two commits, including the main refactor commit:
34874f392 reducers [nfc]: Fold `loading` into `session`, as just a boolean.
f2c1a8cd2 session reducer: Add test for INITIAL_FETCH_START.

In the first of these, take a look at the test cases. They should be explicit about the part they're actually checking; several of them leave that implicit in the use of `baseState`.

Also this commit introduces a case for `INITIAL_FETCH_START`, so it should add the test case for it too. (I.e. the two commits should be squashed together.)

gnprice:
Looks good -- merged! Thanks again.

PR #3681, opened 2019-11-07
docs: add a link to the server docs

ray-kraesig:
As pointed out yesterday.


gnprice:
Thanks -- merged!

(This kind of pure-docs change is also one where I'd typically push directly, more wiki-style, and would encourage you to do the same. I'll see it in any case, so there'll be post-commit review.)

PR #3682, opened 2019-11-07
Apply new lint: `flowtype/require-readonly-react-props`

ray-kraesig:
Upgrade `eslint-plugin-flowtype`, and enable the `flowtype/require-readonly-react-props` lint.

This requires (mostly automated) changes to most of our Props types. It's also not perfect enforcement of immutability, but it may be better than nothing.

### DIFF
--- a/src/boot/TranslationProvider.js
+++ b/src/boot/TranslationProvider.js
@@ -53,9 +53,9 @@ const makeGetText = (intl: IntlShape): GetText => {
  * See https://reactjs.org/docs/context.html
  * vs. https://reactjs.org/docs/legacy-context.html .
  */
-class TranslationContextTranslator extends PureComponent<{
-  children: React$Node,
-}> {
+class TranslationContextTranslator extends PureComponent<{|
+  +children: React$Node,
+|}> {
   context: { intl: IntlShape };
 
   static contextTypes = {
--- a/src/boot/TranslationProvider.js
+++ b/src/boot/TranslationProvider.js
@@ -73,11 +73,11 @@ class TranslationContextTranslator extends PureComponent<{
   }
 }
 
-type Props = {|
+type Props = $ReadOnly<{|
   dispatch: Dispatch,
   locale: string,
   children: React$Node,
-|};
+|}>;
 
 class TranslationProvider extends PureComponent<Props> {
   render() {
--- a/src/common/ZulipStatusBar.js
+++ b/src/common/ZulipStatusBar.js
@@ -24,14 +24,14 @@ export const getStatusBarStyle = (statusBarColor: string): BarStyle =>
     ? 'light-content'
     : 'dark-content';
 
-type Props = {
+type Props = $ReadOnly<{
   dispatch: Dispatch,
   hidden: boolean,
   theme: ThemeName,
   backgroundColor: string,
   safeAreaInsets: Dimensions,
   orientation: Orientation,
-};
+}>;
 
 /**
  * Controls the status bar settings depending on platform
--- a/src/react-redux.js
+++ b/src/react-redux.js
@@ -71,7 +71,7 @@ export function connect<SP, P, C: ComponentType<P>>(
   mapStateToProps?: (GlobalState, OwnProps<C,
     // Error "property `foo` is missing"?  Add to inner component's props.
     SP>) => SP,
-): C => ComponentType<OwnProps<C, SP>> {
+): C => ComponentType<$ReadOnly<OwnProps<C, SP>>> {
   return connectInner(mapStateToProps);
 }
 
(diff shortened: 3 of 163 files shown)


src/common/ZulipStatusBar.js
```diff
@@ -24,14 +24,14 @@ export const getStatusBarStyle = (statusBarColor: string): BarStyle =>
     ? 'light-content'
     : 'dark-content';
 
-type Props = {
+type Props = $ReadOnly<{
```
gnprice: nit: In the commit message:

    This is an automated transformation, performed by the following:
    
      # type Props = {| ... |};
      find src -name '*.js' | \
        xargs perl -i -0pe 's/(type Props \= )(\{\|.*?\|\})(;)/$1\$ReadOnly<$2>$3/
sg'
    
      # type Props = {| ... |};
      find src -name '*.js' | \
        xargs perl -i -0pe 's/(type Props \= )(\{.*?\})(;)/$1\$ReadOnly<$2>$3/sg'

I think the latter comment was meant to not have the `|` pair.

src/boot/TranslationProvider.js
```diff
@@ -54,7 +54,7 @@ const makeGetText = (intl: IntlShape): GetText => {
  * vs. https://reactjs.org/docs/legacy-context.html .
  */
 class TranslationContextTranslator extends PureComponent<{
-  children: React$Node,
+  +children: React$Node,
```
gnprice: nit: the commit message mentions "nonempty exact types defined inline" but this one is inexact. Perhaps s/exact/object/ in that line?

... Also I don't see a reason why this *shouldn't* be exact. Probably good to make it so, either before or after this commit. Most likely it just got missed when I swept through making props types exact, because it was defined inline.

Oh, I guess the message says "fall largely", so it's not untrue as is. Still, easy to make it complete so might as well. And following the links, I believe all the same facts are just as true of inexact object types. Ah in fact here's the docs for `$ReadOnly` itself:
So effectively it seems to be *defined*, at an interface/docs level, as just applying `+` to each key.

src/react-redux.js
```diff
@@ -71,7 +71,7 @@ export function connect<SP, P, C: ComponentType<P>>(
   mapStateToProps?: (GlobalState, OwnProps<C,
     // Error "property `foo` is missing"?  Add to inner component's props.
     SP>) => SP,
-): C => ComponentType<OwnProps<C, SP>> {
+): C => ComponentType<$ReadOnly<OwnProps<C, SP>>> {
```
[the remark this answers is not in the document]
ray-kraesig: > [...] which we know because you've prepared those follow-on commits and they pass tests.

Well, that and the fact that React requires props to be immutable anyway. If this commit resulted in any declared types being tightened beyond their existing use, that would reveal an actual bug in the code.

(Clarified accordingly, though.)

src/boot/TranslationProvider.js
ray-kraesig: Made the relevant change. Added that link to footnote 2.

ray-kraesig:
(Also rebased to resolve conflicts.)

src/react-redux.js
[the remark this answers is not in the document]
gnprice: Mmm, true enough!

src/boot/TranslationProvider.js
gnprice: Looks good! Just one quibble remaining, in the commit message:

    The former have been manually wrapped with `$ReadOnly`. The latter
    have had their entries prefixed with `+`, which for exact types has
    essentially the same effect.

I think this could more simply say "[...] prefixed with `+`, which means the same thing." That is: I believe it's exactly the same effect, by definition, and independent of whether the object type is exact. (That's based on that linked bit of docs.)

Does that sound right to you?

ray-kraesig: It sounds plausible, but not definite: I read your phrasing "means the same thing" as "is identical", and the docs merely state that the two are "equivalent". Equivalent modulo what?

(I'll punt on that if they will. Re-revised.)

gnprice: New text LGTM!

    The former have been manually wrapped with `$ReadOnly`. The latter
    have had their entries prefixed with `+`, which is functionally
    equivalent. (See, e.g., the Flow documentation [2].)

---

FWIW, I think by "equivalent" they mean that they're completely interchangeable for the question of whether a program is well-typed or not. IOW from a type-theory perspective they are different ways of spelling the exact same type; the type-checker might remember which way you spelled it, but that's only a concession to the needs of good error messages for humans.

That interpretation comes partly from scanning for other uses of "equivalent" on that page, and partly from knowing that the Flow people are pretty type-theory-oriented and applying my mental model of how people talk and write in academic PL/types communities.

PR #3683, opened 2019-11-11
Upgrade sentry (redux^Wpart 2)

ray-kraesig:
Rewrite all (untouched) Sentry code. There is no significatnt continuity between the old and new SDK code; neither, then, is any called for in our invocations thereof.

(Draft PR, as this is largely untested.)


src/sentry.js
```diff
@@ -1,15 +1,75 @@
...
+//
+//   A better method may be to call `init()` again, even though the "Static API"
+//   documentation [5] says this is undefined. A still-better method might be to
+//   disregard the "Static API" and use `getCurrentHub()` everywhere.)
+
+// [1] https://github.com/getsentry/sentry-javascript/blob/7a792f4e/packages/types/src/options.ts#L25
+// [2] https://github.com/getsentry/sentry-javascript/blob/7a792f4e/packages/types/src/options.ts#L19
+// [3] https://github.com/getsentry/sentry-javascript/issues/2039#issuecomment-486674574
+// [4] https://github.com/getsentry/sentry-javascript/blob/7a792f4e/packages/core/src/baseclient.ts#L76
+// [5] https://docs.sentry.io/development/sdk-dev/unified-api/#static-api
```
gnprice: Can we just never call Sentry when we have it disabled? It seems like it wouldn't be too hard to have our own boolean defined in this file to track that; and then in our handful of direct call sites to Sentry, we could just check that boolean first. Those are already concentrated into a handful of wrappers, for good reasons independent of this aspect.

If that'd let us not have to worry about any of the complexity described in this comment, that seems like it'd be an excellent choice.

flow-typed/npm/@sentry/react-native_v1.x.x.js
```diff
@@ -14,7 +14,70 @@
...
+  // * https://github.com/getsentry/sentry-react-native/blob/master/src/js/backend.ts
+  // * https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/backend.ts
+  // * https://github.com/getsentry/sentry-javascript/blob/f4e59bcdd/packages/types/src/options.ts
+  //
+  // Please add them to this declaration as (and if) they're needed.
+  declare export type Options = {|
+    enabled?: boolean,
+    dsn?: string,
+    ignoreErrors?: Array<string | RegExp>,
+  |};
```
gnprice: This `flow-typed/npm/` directory is automatically maintained, with `yarn update-libdefs` (which in turn is a thin wrapper around `flow-typed update`.)

For hand-written libdefs, to keep them clear of getting clobbered the next time we run that, we put them in the parent `flow-typed/` directory. For examples that override the third-party libdefs, see `flow-typed/react-intl.js` and `git log --stat -p flow-typed/**/*document-pick*` (or replace `**/` with `{,npm/}` if you don't have `shopt -s globstar` in your bashrc.)

src/utils/logging.js
```diff
@@ -22,12 +24,53 @@ import config from '../config';
...
+ *
+ * The error will be logged to Sentry, including a stack trace. The stack trace
+ * is taken from `err` if an `Error` object, and otherwise synthesized from the
+ * call site.
+ *
+ * To match the function name, the default severity is 'log'.
+ *
+ * Returns a Sentry event_id, although this is not expected to be useful.
+ */
+export const logToSentry = (event: string | Error, level: SeverityType = Severity.Log): string => {
```
gnprice: I'd like to keep this unexported, for basically the reason described in fd8f5cdae:

    logging: Migrate all `logErrorRemotely` calls to new `logging.*` API.
    
    Two reasons to get "remotely" out of the function's name:
    
     * It was already inaccurate -- this function was just as much for
       logging locally as remotely.
    
     * If accurate, it wouldn't belong in the API.  The choice that
       makes sense to express at individual error sites is between
       message severity levels (error, warning, etc.), while specific
       reporting routes (local logs, Sentry, etc.) are a policy choice
       to make centrally per severity level.

Naturally we can always revise that if we find we really want to log specifically to Sentry somewhere. In that case I'd also want to probe a bit into what that use case is really trying to do, and see if we can come up with an appropriate bit of API to describe that.

gnprice: (Very helpful to have this as a separate helper function, though! I'd waffled on whether to do that even when writing the current version, where the corresponding code is pretty simple.)

src/sentry.js
```diff
@@ -1,15 +1,75 @@
...
+
+Sentry.init({
+  enabled: config.sentryKey === null,
+  dsn: config.sentryKey ?? 'none',
+  ignoreErrors: ['Network request failed'],
+});
+
+// This is normal behavior when running locally; only builds shipped to the
+// storefronts will have a Sentry key.
```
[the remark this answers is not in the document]
ray-kraesig: Fixed!

src/sentry.js
ray-kraesig: We could certainly check isEnabled at call sites, but we'll still have to worry about this eventually: at some point we'll need to be able to completely enable or disable Sentry at runtime, not just startup time, and then all this becomes relevant again.

(In particular, merely toggling that hypothetical boolean wouldn't suffice, as it wouldn't disable the unhandled-exception handlers.)

flow-typed/npm/@sentry/react-native_v1.x.x.js
ray-kraesig: Fi—

—well, this turned out to be more interesting than I thought. See new commits.

src/utils/logging.js
ray-kraesig: That's fair. Done!

ray-kraesig:
> I think you mentioned that it wasn't logging things right, but I don't recall the particulars. Does it log the exceptions at all? What's missing or wrong in them?

The code in `logging.js` was making calls using the wrong (i.e., old) function signatures. I don't recall whether it actually threw a new exception in mid-reporting attempt when it tried to access nonexistent fields, or whether it just sort of staggered through and managed to make a half-valid report anyway.

Since the broken version never made it out into an actual release (meaning that we won't have to wade through half-valid events in Sentry) I'm not too concerned about what the exact symptoms were.

gnprice:
> The code in `logging.js` was making calls using the wrong (i.e., old) function signatures. I don't recall whether it actually threw a new exception in mid-reporting attempt when it tried to access nonexistent fields, or whether it just sort of staggered through and managed to make a half-valid report anyway.
> 
> Since the broken version never made it out into an actual release (meaning that we won't have to wade through half-valid events in Sentry) I'm not too concerned about what the exact symptoms were.

Cool, makes sense!

This would be good to add to the commit message, then. :slightly_smiling_face:

src/sentry.js
gnprice: > at some point we'll need to be able to completely enable or disable Sentry at runtime, not just startup time

Will we? I'm curious about the use cases you have in mind; and will be interested in thinking about what solutions we can come up with for them.

Possibly a chat thread would be a good place for that discussion.

.flowconfig
```diff
@@ -40,6 +40,10 @@ node_modules/@zulip/shared/
 [libs]
 node_modules/react-native/Libraries/react-native/react-native-interface.js
 node_modules/react-native/flow/
+; If the following line is not present, Flow ignores this file in favor of
+; flow-typed/npm/@sentry/react-native_vx.x.x.js. (Contrast commit 6219e5a23,
+; where the opposite behavior is exhibited!)
+flow-typed/@sentry/react-native_v1.x.x.js
```
gnprice: Huh! Peculiar. I have no good hypotheses to explain this after a couple of minutes' looking; I might poke at it a bit more later.

flow-typed/@sentry/react-native_v1.x.x.js
```diff
@@ -0,0 +1,187 @@
+// flow-typed signature: a24669b78520403133fc3a2153278527
+// flow-typed version: <<STUB>>/@sentry/react-native_v1.0.9/flow_v0.92.0
+
+/**
+ * This is an autogenerated libdef stub for:
```
gnprice: We can delete all the autogenerated goo below, and this autogeneration header with it. The goo is only useful for making things omni-permissive, and with your hand-written types that's no longer needed :slightly_smiling_face: or desired.

.flowconfig
ray-kraesig: It is concerning. In particular I worry that it's only pseudodeterministic – based on the ordering of keys in a hashmap, or what-have-you. 

(I spent some time hunting through the source myself before writing this hack, but no luck. :sweat:)

ray-kraesig:
> This would be good to add to the commit message, then.

... yes? It's (supposed to be) attached to the "`sentry: complete rewrite...`" commit. What's missing from it?

gnprice:
> > This would be good to add to the commit message, then.
> 
> ... yes? It's (supposed to be) attached to the "`sentry: complete rewrite...`" commit. What's missing from it?

The commit message I see for that commit is:

    sentry: complete rewrite of all Sentry-invoking code (+ Flow types)
    
    The Sentry API has undergone a complete overhaul. The old `raven-js`
    internals have gone away, replaced by a unified JavaScript SDK that
    dispatches to native backends where possible.
    
    Additionally Sentry has gone to some lengths to unify their API across
    platforms: see https://docs.sentry.io/development/sdk-dev/ for
    details. Unfortunately, the new API is in many ways only _subtly_
    incompatible with the old -- and although TypeScript type definitions
    are provided, Flow type definitions are not. Any old code in place
    looked like it still worked, even though it should have generated
    errors.
    
    This effectively calls for a complete reanalysis and rewrite of every
    part of our code that directly invokes Sentry. One should perhaps
    think of this transition not as an upgrade so much as a movement to a
    completely different SDK.

Which is the same version that was there when I asked the question in the first place. So from that you can at least infer that I didn't find the answer obvious from this description. :wink: 

I think the key bits in your reply that I found helpful were:
* The version before this change (but post-upgrade) probably doesn't really work, or at best half-works.
  * (I think maybe "it should have generated errors" was meant to convey something like that? But that "should have" sounds to me like a counterfactual -- like it's suggesting that it didn't generate errors. And I don't see anything else indicating what actually did go wrong.)
* That version hasn't been in a release. (Which is why we don't mind being vague about the exact symptoms.)

Oh, relatedly: it would be good to mention commit `ec101ddcb` and the fact that we recently upgraded from an 0.x version to a 1.x version. That's information I did have, but is also important context for understanding this commit.

src/sentry.js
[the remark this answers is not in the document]
gnprice: (Perhaps you didn't commit something? As of 7f62ed555 it seems to be unchanged.)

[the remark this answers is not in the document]
ray-kraesig: ... apparently not. Fixed – for real this time, I hope. 🤞

ray-kraesig:
>   * (I think maybe "it should have generated errors" was meant to convey something like that? But that "should have" sounds to me like a counterfactual -- like it's suggesting that it didn't generate errors. And I don't see anything else indicating what actually did go wrong.)

Less counterfactual than ambiguous: it's entirely possible that the failure was silent everywhere. I've attempted to clarify that.

> * That version hasn't been in a release. (Which is why we don't mind being vague about the exact symptoms.)
> 
> Oh, relatedly: it would be good to mention commit `ec101ddcb` and the fact that we recently upgraded from an 0.x version to a 1.x version. That's information I did have, but is also important context for understanding this commit.

Added (in both cases).

ray-kraesig:
Withdrawn while I sort out (additional!) divergences in Sentry behavior from expectations.

Fortunately, it looks like the revision will have tests attached.

PR #3684, opened 2019-11-12
webview js: fix prelude JS function call syntax

ray-kraesig:
Extraneous trailing commas in function calls only became legal in ECMAScript 2017.


gnprice:
Thanks! Merged.

PR #3686, opened 2019-11-12
webview [nfc]: Further write up what we've learned about asset paths.

gnprice:
This follows up on b2b274c4b with some further things we learned in
the discussion on its PR #3677, and in a pair of related chat threads:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/near/796432

Also add a couple more reference links, and adjust wording around
`file:///android_asset/` to avoid seeming to assert that it
corresponds to a directory really mounted somewhere in the filesystem.


gnprice:
For the reasons mentioned on #3687 , I'm going to go ahead and merge this now. @ray-kraesig if you find there's something in the resulting version that you don't think is right (and it isn't something with an obvious fix, like a typo), then I think we agree we should handle that without doing additional investigation -- I think we should be able to mark the uncertainty in the text, without deleting the information.

PR #3687, opened 2019-11-13
webview [nfc]: Remove contentious comment; isolate its subject

ray-kraesig:
Having this comment block in place has already cost us more time and
effort than it could ever reasonably save. It is a _fractal can of
worms._ Remove all attempts at explanation from it.

Since it's been firmly established to be useless for all other
purposes, move the value to its use-site, instead of placing it near
the top of the file where it's more likely to distract people.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -104,38 +104,6 @@ export type Props = {|
   _: GetText,
 |};
 
-/**
- * A URL-like magic string denoting the webview-assets folder.
- *
- * Not suitable for reuse or export.
- */
-const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/webview';
-// What the value above probably _should_ be, semantically, is an absolute
-// `file:`-scheme URL naming the path to the webview-assets folder. [1]
-//
-// * On Android, that's exactly what it is. Different apps' WebViews see
-//   different (virtual) root directories as `file:///`, and at runtime, the
-//   APK's `assets/` directory is mounted at `file:///android_asset/`. We can
-//   easily hardcode that, so we do.
-//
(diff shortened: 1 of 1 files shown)


gnprice:
Hi @ray-kraesig! Coming back to this PR now after a while, because Chris mentioned the contemporaneous #3686 over at https://github.com/zulip/zulip-mobile/pull/3866#issuecomment-582316854 .

I agree that we spent a lot of time investigating the subject matter this comment is about, more time than the value the information has for us. But: at this point we *have* spent that time, and I think the information does have real value. So I'd like to keep our results from that investigation around for handy reference, even though ex ante the best use of time would have been to do less.

(Specifically, I think the right strategy would have been to take only the information we'd already learned in order to write this code, and serialize that into a less-detailed comment that stopped there and just clearly marked the boundary of what we knew.)

For that reason, I'm going to close this PR and merge #3686, which writes up what we learned in the last phase of our research into this area. That'll be a return here to our normal practice for docs and comments of editing them more wiki-style than we edit code, pushing changes before review. If you find there's something in the resulting version that you don't think is right (and it isn't something with an obvious fix, like a typo), then I agree I want to handle that without doing additional investigation -- but we can mark the uncertainty rather than deleting the information.


> move the value to its use-site, instead of placing it near
> the top of the file where it's more likely to distract people.

Yeah, I agree this belongs next to its use-site. But I think actually that use-site itself is somewhat distracting in its current location -- in the last week or two when reading PRs in this file `MessageList.js`, I realized that the whole chunk of logic with `baseUrl` and `onShouldStartLoadWithRequest` (which contains a lot of detail) is pretty much independent of the other work that `MessageList` is doing, and had the thought that it'd be helpful if those were pulled out to their own spot outside that React component's definition. Maybe I'll try making that rearrangement.

PR #3688, opened 2019-11-13
Deduplicate indirect deps with yarn-deduplicate.

gnprice:
Also add a linter (in the form of another tools/test "suite")
to keep it that way.
    
Inspired by zulip/zulip@e57ede942 (zulip/zulip#13118). (thanks @andersk !)

### DIFF
--- a/tools/test
+++ b/tools/test
@@ -64,6 +64,10 @@ cd "$rootdir"
 
 PATH=node_modules/.bin:"$PATH"
 
+# Intersect $files with the set of our JS files in src/.
+#
+# Prints a list of newline-terminated paths; either files, or
+# directories meaning their whole subtrees.
 files_js() {
     case "$files" in
         full)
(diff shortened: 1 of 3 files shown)


gnprice:
This script in general is not the awesomest code; I think it's quietly outgrown the scale at which its current style (e.g. a bunch of names that are short and kind of cryptic) is good enough. This branch adds a doc-comment on a function that needed it, but doesn't try to comprehensively address that right now.

Just after writing this, I went and drafted up a branch to start using shellcheck -- inspired by this as well as other recent shell-script-touching PRs. That won't directly help with things like too-short names, but it'll be another step toward better code here.

tools/test
```diff
@@ -162,6 +171,22 @@ run_prettier() {
...
 }
 
+run_deps() {
+    files_check package.json yarn.lock \
+        || return 0
+
+    if ! yarn-deduplicate --fail --list; then
+        cat >&2 <<EOF
+
+Duplicates could be shared in yarn.lock.
```
ray-kraesig: This error message is more than a little cryptic; I'd suggest something more explicit, along the lines of "`Mergeable duplicate dependencies found in yarn.lock`".

gnprice: Good idea. I went for even more explicit than that:
> Found duplicate dependencies in yarn.lock which could be dedup'd.

gnprice:
Thanks @ray-kraesig for the review! Merged, with the edit mentioned above.

Issue #3614, opened 2019-09-16
Keep compose box scrolled to bottom, not top, as you type a long message


gnprice:
Since we fixed #3369 with #3595 , the compose box is now scrollable so you can see what you're typing even in the later part of a long message.

But it doesn't automatically scroll itself that way; by default it sticks at the top, and you have to scroll it yourself. See report in chat, [with a video](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Issues.20with.20long.20messages/near/786652).

I'm not sure what the right logic might be for a complicated case where the user is moving to and fro in the message and editing different parts of it. Let's declare the problem of getting that exactly right to be out of scope for this issue.

The most common case is that the user mostly types additional text at the end, and perhaps occasionally scrolls up to edit an earlier part of the message. Let's make that case work such that as long as you're at the end, there's no manual scrolling needed.

jainkuniya:
This issue is easily reproducible on iOS. On Android it gets automatically scrolled up on typing.

If we revert https://github.com/zulip/zulip-mobile/pull/3595, then it will get fixed. But then #3369 will occur on Android. 

Next will try to find out a simple common solution.

gnprice:
Thanks @jainkuniya ! Just to make things explicit:

> This issue is easily reproducible on iOS. On Android it gets automatically scrolled up on typing.

I believe you're saying this issue doesn't reproduce on Android.

The original user report of the issue was on iOS, so that's consistent too.

jainkuniya:
This compose box issue is only issue on screen when there are multiple input on the screen, like stream narrow (not topic or PM narrow).

Root cause is: Topic input always try to remain in the window/screen, even when message input expands. So only way left for message input to expand is to expand at bottom, instead of top.

gnprice:
> Root cause is: Topic input always try to remain in the window/screen, even when message input expands. So only way left for message input to expand is to expand at bottom, instead of top.

Ah interesting. Are there particular experiments you've done, or things you've observed, that lead to this diagnosis?

gnprice:
There were a couple of different attempted fixes in the PR #3690, but they didn't 100% work, and it seems like we don't yet understand the cause of this issue. Let's focus first on investigating the issue, so we can understand what's going on; then once we understand it, we can fix it.

Copying here a version of some of my comments there, as well as more summary of our conversation just now:

[...]

Issue #3369, opened 2019-02-26
Compose box obscured by keyboard on long message, in stream view


armaanahluwalia:
When replying to a message in the stream view, if the message you’re typing is longer than a certain line length, the message gets cut off.

This makes drafting long messages extremely difficult. See [this video](https://streamable.com/m9kho) for a reproduction.

gnprice:
Thanks @armaanahluwalia for this report! Very interesting.

To highlight one detail: this specifically happens when narrowed to a whole stream -- it doesn't happen when narrowed to a topic.

To describe here in words the exact symptom: when you type a long (draft) message, the compose box grows to make room... and in the video, it eventually grows to try to use some space that's hidden by the keyboard.

Some further observations in the video:
* The send button moves along with the bottom of the compose box... and right to the end of the video, you can still see a bit of the top of it. So the box is growing just a constant amount too far down.
* Even before you type anything, there's a different symptom present! ![image](https://user-images.githubusercontent.com/28173/53456805-9d8b7580-39e4-11e9-9f1f-373e9e1d8688.png)

  That blank white band between the compose box and keyboard is doing nobody any good.
* You appear to be on iOS. Helpful to mention for issues involving layout and/or the keyboard :slightly_smiling_face: , as they often vary by platform.

Further notes:
* With 23.1.110 (the current beta) on Android 9, I reproduce this issue -- but I don't reproduce the blank white band. So that's probably an unrelated issue.
* Gosh, though -- the height of that blank white band is just about equal to the amount we're going too low. So I still wonder if they're related somehow, even though one can happen without the other.

gnprice:
Oh also -- this symptom is reminiscent of #3273 "Keyboard overlaps the compose box on iOS". That one was fixed in 23.0.109, but the history may be a clue for what to look into.

gnprice:
(I went and filed #3370 for the blank white band.)

gnprice:
Some discussion on a duplicate issue, #3414.

One direction to fix this would be to limit the compose box's height. But this shouldn't be a small fixed limit like 6 lines; see discussion on #3049. Instead it should fill the space available. See @borisyankov 's comment on #3414 about how that would ideally work.

Also @jainkuniya points out that the fact this happens only on a stream narrow means it only happens when the topic input is present. That's probably a clue. The compose box's layout already works just fine when there's no topic input; if we can make it work equally well when there is, that solves the problem.

gnprice:
[...]

PR #3690, opened 2019-11-16

jainkuniya:
Now this is how it looks:

Fixes: #3614
This reverts #3595 and include fix for #3369.

image
image

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { Platform, View, TextInput, findNodeHandle, ScrollView } from 'react-native';
+import { Platform, View, TextInput, findNodeHandle } from 'react-native';
 import type { LayoutEvent } from 'react-native/Libraries/Types/CoreEventTypes';
 import TextInputReset from 'react-native-text-input-reset';
 
@@ -292,6 +292,7 @@ class ComposeBox extends PureComponent<Props, State> {
       alignItems: 'flex-end',
     },
     composeText: {
+      flex: 1,
       paddingVertical: 8,
     },
     composeSendButton: {
@@ -360,7 +361,7 @@ class ComposeBox extends PureComponent<Props, State> {
             expanded={isMenuExpanded}
             onExpandContract={this.handleComposeMenuToggle}
           />
-          <ScrollView contentContainerStyle={this.styles.composeText}>
+          <View style={this.styles.composeText}>
             {this.getCanSelectTopic() && (
               <Input
                 style={this.styles.topicInput}
@@ -392,7 +393,7 @@ class ComposeBox extends PureComponent<Props, State> {
               onSelectionChange={this.handleMessageSelectionChange}
               onTouchStart={this.handleInputTouchStart}
             />
-          </ScrollView>
+          </View>
           <FloatingActionButton
             style={this.styles.composeSendButton}
             Icon={editMessage === null ? IconSend : IconDone}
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @jainkuniya !

From the commit messages, I think this is a WIP branch. (There's a "TODO add more details here" in one of them :wink:) It's best to make that explicit in the PR thread too.

There's an issue this fixes, right? Please mention it in the PR description, and in the commit message.

---

It looks like what this does in the code is:
* add a wrapper `View` around each of the topic input and message-content input
* add `justifyContent: 'space-around'` to the style for the common wrapper `View` we have for those two inputs together
* replace `marginBottom` on the topic input with `marginTop` on the new wrapper `View` of the message-content input

It's not obvious to me why any of those is related to the issue. Do you have a theory for why this changes the behavior?

Also, are all three (or more) of the changes actually relevant? Try removing bits of it one at a time:
* no new wrapper `View` for the topic input
* keep the `marginBottom` rather than `marginTop`
* no new wrapper `View` for the content input -- can put the style on the input itself instead of wrapper
* etc.

If a simpler change works just as well, then it may help for understanding what actually drives it -- which will make it easier to avoid introducing a similar issue in the future. Also makes a smaller, easier-to-understand diff :smile:

jainkuniya:
Ya, will add more details next.

Actually this current state is just based on try and run 😅. Not sure about the exact reasons, will try to find them out.

Just curious, have you tried this branch?

gnprice:
I haven't yet tried the branch; it seems like this is still in the investigation phase, and I'm relying on you to do the investigation :slightly_smiling_face: 

I'll definitely be interested in hearing what results you get when you try removing bits of this change one at a time (including the bits I suggested above). Among other reasons, it's likely that will help us understand what's going on.

gnprice:
From our chat discussion, I now understand your screenshots better: those are both from a stream narrow. In the second one, the message content input has grown so tall it's covering over the topic input.

Trying the branch out, I got a slightly different result!
image
That's on a simulated iPhone 8 running iOS 13.2.2. The content input has grown tall... but it's partly *under* the topic input, instead of over it. (It seems to be over some of the padding of the topic input, but under the topic input itself.)

So that's interesting that it behaves differently. Can you describe what conditions you're using that get the result you showed above?

Both of these behaviors seem like they're improvements on the status quo, so I'd be OK merging a version that still has this behavior.

I would still like you to do the experiment of removing bits of this change one at a time. That will be informative, and may make the change simpler.

src/compose/ComposeBox.js
```diff
@@ -292,7 +292,9 @@ class ComposeBox extends PureComponent<Props, State> {
       alignItems: 'flex-end',
     },
     composeText: {
+      flex: 1,
       paddingVertical: 8,
+      justifyContent: 'flex-end',
```
jainkuniya: Have never imagined fix will be this short 🤦‍♂ 

Now I guess this solution is ideal as we wanted 😍

gnprice:
Summarizing our conversation just now:

It looks like the core issue here is that the content input is too tall. In the extreme case the input is too tall for the whole input to be shown at all; in other cases it's itself shown in full, but it crowds out the topic input, or the unreads banner.

The iOS screenshots above are on an iPhone X-family model, with a notch, which pulls in an unrelated bug that causes overlapping symptoms. Best to focus on a model without a notch. I use a simulated iPhone 8. On that, this branch still shows symptoms caused by the input being too tall.

It seems like we still don't yet really understand what's causing the issue, or why these attempted fixes have the effects they do. I'd like you to focus at this stage on investigating the issue so we can understand what's going on, and not on the code of an attempted fix. That's best tracked over on the issue thread #3614; I'll close this PR.

PR #3691, opened 2019-11-16
webview: work around WebKit bug involving CSS transitions

### LINKED ISSUE #3689: Timestamp pills start out visible on stream entry
Upon entering a stream, timestamp pills for visible messages start out extended, and immediately slide inwards.

`git bisect` suggests this behavior was introduced relatively recently, at commit 8fdb93bd4dac796834b639846867c41296653557, titled
`webview css: move base CSS into actual CSS file` – which should not have been a functional change.

This has been confirmed to occur (on `master`) on iOS 11.4, iOS 13.2, and Android Pie and Q. It does _not_ appear to occur on Android Lollipop or Marshmallow.

ray-kraesig:
Fixes #3689.

### DIFF
--- a/src/webview/html/html.js
+++ b/src/webview/html/html.js
@@ -11,6 +11,34 @@ type InitOptionsType = {|
   showMessagePlaceholders: boolean,
 |};
 
+/**
+ * Workaround for a WebKit CSS transitions bug affecting both iOS and Android.
+ *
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks!

(I accidentally put my review on the issue thread #3689, oops.)

PR #3693, opened 2019-11-19
prettier: Remove old VSCode setting

ray-kraesig:
New versions of the Prettier VSCode plugin (in particular, 3.8) have
started to complain about certain out-of-date settings in VSCode
config files.  Remove the relevant (or rather, now-irrelevant) config
option from our workspace.

Users may need to manually update their Prettier VSCode extension to
continue to format code with Prettier on save.

See, e.g:
- https://github.com/prettier/prettier-vscode/blob/62004a02b4/README.md#user-content-error-messages

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,8 +8,7 @@
     // https://github.com/Microsoft/vscode-react-native/blob/master/doc/intellisense.md .
     "flow.useNPMPackagedFlow": true,
 
-    // Format on save, with prettier-eslint.
-    "prettier.eslintIntegration": true,
+    // Format on save (with prettier-eslint).
     "[javascript]": {
         "editor.formatOnSave": true,
     },


gnprice:
> Users may need to manually update their Prettier VSCode extension to continue to format code with Prettier on save.

Hmm. What's the concrete action item here?

FWIW I seem to have version 3.8.0 of that extension. I haven't seen a warning about this; maybe it's just not surfacing for some reason.

gnprice:
> FWIW I seem to have version 3.8.0 of that extension. I haven't seen a warning about this; maybe it's just not surfacing for some reason.

Ah, OK, then later today I saw it. :-) Must have just taken the update.

I think the key desideratum here is to make sure that our config still works properly -- that if you open an arbitrary file in the codebase, in particular one where we've used `prettier-eslint` to give a more reasonable result than Prettier wants to do on its own, hitting save won't mangle the formatting.

The linked doc sort of gestures at saying it'll work, but it kind of hems and haws and I'm not 100% sure it means to say it will. In any case it'd be good to check empirically for ourselves. :wink:

ray-kraesig:
> Hmm. What's the concrete action item here?

Ensure that version 3.8 is installed, and possibly restart VSCode if it says "restart required".

>  In any case it'd be good to check empirically for ourselves.

Confirmed that:
&nbsp; &nbsp; a) neutering `tools/formatting.eslintrc.yaml` and running `yarn prettier --all` made dozens of formatting changes across the codebase;
&nbsp; &nbsp; b) VSCode accepted the new normal when viewing one of those changes (i.e., add-whitespace-and-save was a visual no-op); and
&nbsp; &nbsp; c) resetting `tools/formatting.eslintrc.yaml` and saving a changed file in VSCode caused that file to revert to the pre-reformat state.

ray-kraesig:
(No content changes in the latest version; I've just removed a reference to another GitHub project from the commit message, where it would result in reference spam.)

gnprice:
> Confirmed that:

Thanks for checking that! I am now reassured. :slightly_smiling_face: 

Merged, with some edits to the commit message. Mainly I wanted to make it clear there this fact you checked -- that the setup still works fine, it's not that the plugin dropped this option because it dropped the functionality, rather it's that the functionality is now there automatically.

Issue #3570, opened 2019-07-25
Migrate to community AsyncStorage


gnprice:
This is a followup to the RN v0.59 upgrade #3399. It isn't required before upgrading to v0.60, and might not be for v0.61.

When running unit tests, e.g. with `tools/test` , or when launching the app with the debugger, the following now goes to the console:
> Warning: Async Storage has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-community/async-storage' instead of 'react-native'. See https://github.com/react-native-community/react-native-async-storage

The core module is still there in RN master when I look now.

Marking as a priority, though, because we shouldn't get used to having a big red warning in our tests.

The README doesn't say anything about upgrading from the core module. Hopefully it's easy.

PR #3694, opened 2019-11-19
Switch to community async-storage module

ray-kraesig:
... and eliminate the angry red message from our build logs.


ray-kraesig:
The test failure here appears to be due to the Jest harness reading the unprocessed Flow code in the referenced file as plain JavaScript – which is more than a little surprising, as the build process itself has no problem with it.

ray-kraesig:
This looks like a simple Jest configuration issue, but resolving that yields further problems which are less trivial to satisfy.

Withdrawing until we have some solution for, or at least better characterization of, the issue of `flow-typed` ignoring type definition files outside of `flow-typed/npm` in favor of those within.

ray-kraesig:
> Withdrawing until we have some solution for, or at least better characterization of, the issue of `flow-typed` ignoring type definition files outside of `flow-typed/npm` in favor of those within.

Well... we don't really have that. But do we have a list started in `.flowconfig`, and that'll probably do for now.

gnprice:
SGTM; merged!

gnprice:
(For the sake of GitHub cross-referencing: this fixed #3570.)

PR #3696, opened 2019-11-20
Revert "api: augment console.log with proper Sentry breadcrumb"

ray-kraesig:
This reverts commit a60a97a1ab5440e1b8221998a4c84623ded035eb.

Despite the commit order, this was originally written and tested
against the 0.x interface. In 1.x, these identifiers are no longer
available with this import method.

A future commit is planned which will incorporate this functionality
correctly, alongside Flow type declarations to ensure this does not
recur elsewhere.


gnprice:
Merged, thanks!

Issue #3685, opened 2019-11-12
On logout and re-login, the org-URL screen flashes by


gnprice:
If you log out of an account, and then from the "switch accounts" screen choose the same account again, there's a quick flash of something before we take you to the "choose how to log in" screen.

That's `RealmScreen` -- what's happening is we take you there, and then it promptly redirects on to the next screen. It fetches from `server_settings` first, which is the main reason we don't just go past it already.

This is a low-priority issue, because this flow is pretty uncommon -- I expect very few users ever go through it, other than people working on Zulip itself.

A previous related issue was #2778. A previous PR to do this was #3055 .

sumukhah:
@gnprice 
can i take this?

sumukhah:
@gnprice 
![ezgif-2-fc1567eb8ec2](https://user-images.githubusercontent.com/23723464/69441504-de949880-0d70-11ea-93ea-a4b0be0c1fb3.gif)
seems working,
can i make a pull request?

gnprice:
Some thoughts on how to approach this, copied from my and @ray-kraesig 's comments on #3712:

## UI

I think a good UI here would be:
* When the user hits the button, we immediately navigate to the auth screen, and make the request.
* While the request is pending, the auth screen shows a "loading" animation. A simple version of this would be that it's mostly blank, with a spinner.
  * In a fancier version, the "still loading" version of the auth screen would have a layout made to look like the real thing, with placeholder gray rectangles. This requires more design work to get right; strategy-wise, probably best to leave as a possible followup.

Then if the user wants to cancel, they just navigate back. And if the request fails, the loading animation gets replaced with an error message, and a button to retry; the user can either hit the button, or navigate back.

The translucent overlay with spinner would also work. I think that was a very common UI style a few years ago.

Aside from just feeling more contemporary, I think the main thing the "transition to mockup loading version" approach accomplishes, relative to the translucent overlay with spinner, is that the intermediate state is visually similar to the final state -- so it doesn't much add to the amount of visual busy-ness.  In particular if the request finishes quickly (which is a common case), this makes the visual effect very close to the ideal of just transitioning straight to the fully-loaded state.

---

(@ray-kraesig:)

* You can't just hide the single button the user clicked; you have to hide – or disable – _all_ the buttons, for the same reason. (Hence the suggestion of a translucent overlay.)
* The user must be able to cancel out of the action with the standard back button/gesture (on Android) and/or a "Cancel" button (on iOS).
[...]

PR #3698, opened 2019-11-22
Re-Login: Fix Navigation to LoginScreen ratherthan RealmScreen on Acc…

sumukhah:
on Selecting an existing AcountScreen would take me to RealmScreen then it would navigate to Login by dispatching navigateToAuth action,
I have fixed this bug by replacing the dispatching function in AcountPickScreen.

I have made a manual testing it worked fine.
I have used an Alert.alert which will be visible if any error raises on selecting an account
Fixes: #3685

### DIFF
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -1,5 +1,4 @@
 /* @flow strict-local */
-import Sentry from '@sentry/react-native';
 import type { UrlParams } from '../utils/url';
 import type { Auth } from './transportTypes';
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -0,0 +1,16089 @@
+{
```
jainkuniya: Just curious how this file got added?
Are you using `npm`?

sumukhah: yes i am using npm

sumukhah: should i delete it?

Issue #3585, opened 2019-08-09
Sentry breaks build on Android, unless disabled


gnprice:
Attempting a normal release build fails with this error:
```
$ time yarn build:android --stacktrace
yarn run v1.17.3
$ cd android && ./gradlew assembleRelease -Psigned -Psentry --stacktrace

// ... familiar warning noise ...

> Task :app:bundleReleaseJsAndAssets FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:bundleReleaseJsAndAssets'.
> A problem occurred starting process 'command '/home/greg/z/mobile/android/app/build/generated/assets/react/release/index.android.bundle.map''

// ... giant stack-trace ...

Caused by: java.io.IOException: Cannot run program "/home/greg/z/mobile/android/app/build/generated/assets/react/release/index.android.bundle.map" (in directory "/home/greg/z/mobile"): error=2, No such file or directory
        at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:25)
        ... 7 more
Caused by: java.io.IOException: error=2, No such file or directory
        ... 8 more
```

It's not obvious this is caused by Sentry, but the workaround turns out to be to skip it, by running just `cd android && ./gradlew assembleRelease -Psigned`.

First saw the issue 2019-07-19 when building v25.7.121; it was fine when building v25.6.120 on 2019-06-18.

As a result I've been leaving Sentry out of all Android releases since then. It'd be good to debug the issue and sort it out so we can start using Sentry again.

gnprice:
This is presumably back, now that we've reverted the Sentry upgrade.

gnprice:
Fixed by #3702.

PR #3702, opened 2019-11-25
Rewrite Sentry use (v3)

ray-kraesig:
This patchset upgrades the Sentry SDK, rewrites our use of it, and adds supporting tests and Flow types.

Replacement for reverted PR #3676 and failed PR #3683.

Fixes issue #3585.


ray-kraesig:
Updated to fix CI by running `yarn-deduplicate`. No changes outside `yarn.lock`.

src/sentry.js
```diff
@@ -1,11 +1,36 @@
...
+    // eslint-disable-next-line no-console
+    console.log(`Sentry key ${displayKey} provided; initializing`);
+
+    Sentry.init({
+      dsn: key,
+      ignoreErrors: ['Network request failed'],
+    });
+  } else {
+    // This is normal behavior when running locally; only builds shipped to the
+    // storefronts will have a Sentry key.
```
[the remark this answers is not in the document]
ray-kraesig: Well, for once I know exactly how this happened. This actually got fixed for the old PR... but the fix didn't get carried over to the new one, because I didn't pull everything over from the Mac – just a few cherry-picked commits. :man_facepalming:

src/__tests__/sentry-test.js
```diff
@@ -0,0 +1,30 @@
...
+    };
+
+    test('breadcrumbs', () => {
+      expectSentryUninitialized();
+
+      Sentry.addBreadcrumb({
+        message: 'test message; should not result in a real breadcrumb',
+        level: Sentry.Severity.Debug,
+      });
+    });
```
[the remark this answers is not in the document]
ray-kraesig: That's fair.

The second one should probably stay essentially as is, though, to prevent some hypothetical future reader of Sentry event logs dismissing `Jest test error` as harmless expected noise. I've tagged it with a comment.

src/sentry.js
```diff
@@ -1,11 +1,36 @@
...
-if (config.enableSentry) {
-  Sentry.config(config.sentryKey, {
-    deactivateStacktraceMerging: true,
-    ignoreErrors: ['Network request failed'],
-  }).install();
-}
+/** Initialize Sentry into its default configuration. */
+const initSentry = () => {
+  // Under no circumstances should Jest ever execute this function.
+  if (process.env.NODE_ENV === 'test') {
```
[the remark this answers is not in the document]
ray-kraesig: (Mostly discussed offline; but for the record...)

I'm reluctant to do this – there are very, very few reasons that one should ever call `isTest`. I'd actually be more comfortable dropping this commit entirely and running the risk of sending Jest events to Sentry than implying that it's vaguely acceptable to `import { isTest } from '../config'` from arbitrary files.

As far as it being _the_ right way, I'm not sure there is one; but this one is, just barely, documented: https://jestjs.io/docs/en/24.0/getting-started.html#using-babel

Following offline discussion I've instead added a somewhat awkward, but non-`export`, `isSentryUnsafe` function to `sentry.js`, and tied that in. This comes with bonus commentary on what happens if you run Sentry in debug mode! (Spoiler: not much.)

ray-kraesig:
> Thanks @ray-kraesig ! Will be great to have this upgraded.
> 
> > [9d8f329](https://github.com/zulip/zulip-mobile/commit/9d8f329bcb71da7b906de217157132a2d54092fe) sentry: disable Sentry exactly when DSN is not specified
> 
> It feels a little odd doing this when Sentry isn't quite wired up at all -- feels a bit unmoored as I try to reason about it.
> 
> This would make sense to do before the upgrade, right? That seems clearer.

It's fallen where it has mostly because this is the point at which there's a minimal amount of Sentry code that needs editing in order to make the change. This is indeed a pretty awkward location, and I don't have any objection to pushing it back one.

(Also, Sentry _is_ wired up at this point, just incompletely: as the previous commit message notes, the native SDK and uncaught exception handler are fully functional.)

> I think some of these could likely also go before the upgrade:
> [be54c32](https://github.com/zulip/zulip-mobile/commit/be54c32cd8dacf4b99d0d812b091471c53af9c9f) sentry: check for, and guard against, Jest

* This one _could_, certainly, but I don't see that it belongs any more naturally before the upgrade than after.

> [033f60d](https://github.com/zulip/zulip-mobile/commit/033f60d376a15dec669d682d47a28e47518ed2cb) sentry: refactor initialization sequence

* The functional components of this one could, but the comment is written against Sentry 1.x. (And, again, this one doesn't seem to be more natural before than after.)

> [087498c](https://github.com/zulip/zulip-mobile/commit/087498c5a06b294ba15eea7a11607f9d1c518327) sentry: add tests to verify assumed behavior

* This one doesn't commute. These tests are written for Sentry 1.x; I have not tested it, nor anything analogous, under Sentry 0.x. Indeed, given the explicit tests for `config.enableSentry` around the old error-reporting code, I would be completely unsurprised if it failed.

  Thankfully, we don't need to care whether such tests would pass or not. Sentry 0.x's behavior in this regard is no longer relevant.


All that said: I suspect that what might make more sense to you – and _would_ trivially commute – would be moving `logging: recreate Sentry logging in utils/logging.js` backward to before the three commits you've mentioned. This would mean we'd have logging code that doesn't check whether Sentry's enabled in place before the tests that mean it's sane to write the logging code without checking whether Sentry's enabled, but a that's relatively minor inversion.

(Not done in this version.)

> "Strip stack frames" is a bit of a confusing description: much of the point is to _give_ us stack frames, right?

Well, not to give us the particular stack frames that `trimHeadFrames` stripped. Or trimmed. (I suspect if I had merely said "trim" instead of "strip" here, there would have been much less confusion.)

Reworded based on your later suggestion, though.

> Speaking of the old code, another thought: it would be helpful to separate out the "remove nonessential Sentry calls" (i.e. stop actually using Sentry) change as its own commit, coming before "upgrade Sentry". One reason is that it'd be much smaller, making it much easier to refer back to that when looking at this "recreate Sentry logging" commit which replaces that code.

:+1: (Well, :+1: except for your parenthetical comment there.)

In retrospect, it would make even more sense to have three commits: one of which removes all the Sentry 0.x code, one of which does the dependency replacement, and the last of which adds all the Sentry 1.x code. ... actually I'll give that a whirl; it should be feasible.

ray-kraesig:
> In retrospect, it would make even more sense to have three commits: one of which removes all the Sentry 0.x code, one of which does the dependency replacement, and the last of which adds all the Sentry 1.x code. ... actually I'll give that a whirl; it should be feasible.

Done. This involved pulling the logging code backward in the commit history (as mentioned earlier), for simplicity's sake.

No changes to the final codebase.

src/sentry.js
```diff
@@ -3,7 +3,7 @@ import { Sentry } from 'react-native-sentry';
 
 import config from './config';
 
-if (config.enableSentry) {
+if (config.sentryKey != null) {
```
gnprice: This one is `!= null` and the others are `!== null`. Is there a reason to make them different?

Given that the type excludes `undefined` I'm inclined to make them all `!==`.

src/__tests__/sentry-test.js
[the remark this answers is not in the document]
gnprice: Ah, I see!
```
+      // The text here is intended to prevent some hypothetical future reader of
+      // Sentry event logs dismissing the error as harmless expected noise, in
+      // case Sentry is somehow actually initialized at this point.
+      const err = new Error('Jest test error; should not result in a Sentry event
');
```

Makes sense; and thanks for the comment explaining it.

src/sentry.js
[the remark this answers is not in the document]
gnprice: This solution works for me!


> than implying that it's vaguely acceptable to `import { isTest } from '../config'` from arbitrary files.

As I said in person but FTR: the way I'd typically address this is by making the name a lot less innocuous than `isTest`. E.g. `unsafeIsInsideTest`, using something like "unsafe" that makes clear it's not OK to use without some really good reasoning for why.

It's always possible for someone to ignore that and blithely use it anyway... but someone who'd cheerfully do that can also refer directly to `process.env.NODE_ENV`, and that doesn't even have such a verbal guardrail.

The new version encapsulates this knowledge at a different boundary, and that accomplishes what I'm looking for here too.

src/sentry.js
```diff
@@ -1,11 +1,55 @@
...
+  //   [1] https://jestjs.io/docs/en/24.0/getting-started.html#using-babel
+  //   [2] https://stackoverflow.com/a/52231746
+  if (process.env.NODE_ENV === 'test' || process.env.JEST_WORKER_ID !== undefined) {
+    return 'Sentry should not be initialized while testing!';
+  }
+
+  // Running Sentry in debug mode is technically *safe*, although it's probably
+  // not *useful*. (The build process doesn't ordinarily create bundles or .map
+  // files in debug mode, so you'll probably get nonsensical stack traces.)
+  /* if (process.env.NODE_ENV === 'development') {
```
gnprice: I'm a bit confused by this pair of comments: why is it unsafe in tests, while merely not useful in debug mode?

What happens if it runs in tests? I'm imagining that it'd put some noise data in the Sentry console. But that's the same thing that would happen in a dev build, right?

src/sentry.js
ray-kraesig: No, this is just an alienism that happens not to have been caught by the linter. Fixed.

src/sentry.js
ray-kraesig: That's one scenario. Another, arguably slightly better scenario is that the calls raise (mildly cryptic) exceptions because Sentry isn't properly mocked.

Jest has no good reason to even *try* to initialize Sentry, no matter what the value of `sentryKey` is. (And because `initSentry` is invoked as a side effect of importing this file, it's possible that we might unknowingly do so when testing something else.) There is some support from Sentry for Jest-mocking... but until and unless we actually have a specific reason to jump through those hoops, I'd rather just yell.

That said, the other side was still wrong. As the comment below notes, it's acceptable and even normal to call `initSentry` while in debug mode; it should always have a null `sentryKey` in this case (unless, perhaps, *&lt;more complicated condition&gt;*), but the attempt will be made, just as is done in release.

And of course, "safe" was never the right word here. I've renamed it to "preventNoise", and simplified the call site.

ray-kraesig: Addendum: if there are still issues with this commit, I'd rather just drop it than hold up the remainder of this commit. Additional checks are generally nice, but we've done fine without these so far.

ray-kraesig:
> I think by "gross" you mean "large, obvious", but it ambiguously could mean "disgusting" -- that was actually my reading at first glance. Like "the differences are subtle, and then disgusting once you learn about them", rather than "some subtle differences, and some big obvious ones."

I've spent long enough mucking around in the SDKs' _[sic]_ guts that I'm okay with that.

> If you mean "large, obvious", how about using a different word? Many potential readers might not even know that sense of "gross" -- I think it's pretty uncommon outside some set phrases in the jargon of law and commerce, like "gross margin" and "gross negligence" -- and will only have the "disgusting" reading.

"both subtle and gross" is a sufficiently-well-established set phrase itself, I think. <sup>[[1]](https://books.google.com/ngrams/graph?content=both+subtle+and+gross&year_start=1800&year_end=2008&corpus=15&smoothing=3) [[2]](https://www.google.com/search?q=%22both+subtle+and+gross%22+-%22satiate+desires%22)</sup>

> Heh, I see there's a 1.1.0 now. Does the difference look nontrivial? If so, pretty reasonable to defer that upgrade to after finishing this big one; OTOH if it's small, it could be good to get the bugfixes. Could even include a fix that simplifies the big upgrade!

There doesn't seem to be anything in [the changelog](https://github.com/getsentry/sentry-react-native/blob/master/CHANGELOG.md) worth delaying this for. (Although we should probably look into the user-feedback API mentioned there; I vaguely recall at least one complaint on one of the storefronts about there not being a good user-feedback channel.)

> Took me a moment to realize "previously" meant "in the parent commit". It'd be good to say that; otherwise, especially for a future reader, this could mean "in some chat thread or GH issue or something".

:+1:

gnprice:
Thanks for the revisions! LGTM, and merged.

src/sentry.js
gnprice: Very helpful, thanks!

Issue #3699, opened 2019-11-22
Presence-reporting logic is confused, in overlapping ways


gnprice:
**EDIT**: Title reflects further discussion in thread below. Previous title:

### ~Failing to report presence inactive when put in background~

and original report from before that:

---

###  ~Can report presence active even after in background~

This is potentially one of the causes of #3659: which would mean it's causing people not to get notifications when they should.

It hasn't been confirmed empirically -- that'll be an important step in addressing the issue.

Found from studying the source. As I wrote there:

> > [@ray-kraesig] Or at least that the mobile app should be better about declaring its lack of presence. (We could possibly put something appropriate in `handleAppStateChange`... but there's a `reportPresence` call in there that looks like it's supposed to handle exactly this. )
> 
> Hmm. There is indeed that:
> 
> ```js
>   handleAppStateChange = (state: 'active' | 'background' | 'inactive') => {
>     const { dispatch, unreadCount } = this.props;
>     dispatch(reportPresence(state === 'active'));
> ```
> 
> But then there's also this:
> 
> ```js
> export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
>   dispatch(reportPresence());
>   setInterval(() => dispatch(reportPresence()), 60 * 1000);
> ```
> 
> That doesn't know anything about being in the background or not. And the `hasFocus` argument to `reportPresence` defaults to true. So one thing that may be happening is
> 
> * app is put in background
> * `handleAppStateChange` gets called; calls `reportPresence(false)`
> * sometime in the next 60 seconds, the timer set up by `doInitialFetch` fires
>   
>   * ... and the OS hasn't yet totally put us on ice. So it runs, and does effectively `reportPresence(true)`, which tells the server we're active after all.
> 
> So that, at least, seems like straightforwardly a bug.

gnprice:
(If empirically it turns out this *doesn't* seem to happen, then that'll be interesting and it'd be good to understand why. Also that fact may be a fragile one, dependent on facts about the platforms that aren't evident in our code, and might not be stable over time -- in which case it'd be good to adjust the code so that it definitely behaves right even if those external facts change.)

ray-kraesig:
At least on my Android test setup, it's both worse and better than that: the `setInterval()`-triggered `reportPresence` doesn't seem to be firing while the app is inactive... but `handleAppStateChange` isn't kicking off a presence report _at all_.

[...]

Issue #3659, opened 2019-10-21
Notifications not received on Android with app in background


ray-kraesig:
Repro:

1\. Log in to the Android app as User A.
2\. Navigate to the "PMs with User B" screen.
3\. Press the Home button to go directly to the Home screen. *Do not force-quit the app;* at this stage it should still be running in the background.
4\. As User B, quickly send a PM to User A.

No notification will appear on the Android device.

5\. Bring up the app list. Slide Zulip off into oblivion.
6\. As User B, send a PM to User A.

Exactly one notification will (eventually) appear on the Android device, for only the PM sent in step 6.

ray-kraesig:
_N.B._: This may be related to [a note on the Android Pie changelog](https://developer.android.com/about/versions/pie/android-9.0-changes-28#suspended-apps) describing how notifications from "suspended" apps have hitherto been canceled.

If so, I'd expect that on Pie devices the user will receive one notification containing both the step 4 and step 6 PMs. (In that case we'll probably still want some sort of workaround for pre-Pie systems, though.)

gnprice:
Thanks for this report!

I didn't repro this on my phone (Pixel 2 XL, running Android 10 and the current Zulip beta) when we tried earlier today.

I'm not sure if we satisfied this bit, though:
> 4. As User B, quickly send a PM to User A.

How quickly does it seem to need to be to trigger the issue?

ray-kraesig:
> Thanks for this report!
> 
> I didn't repro this on my phone (Pixel 2 XL, running Android 10 and the current Zulip beta) when we tried earlier today.

I don't recall if you had the PMing user's message-list open, though – I think you might have been at the PMs-list page.

(For reference, I tested on a Samsung with Marshmallow. The emulator had similar issues, but I've found in-emulator notifications to be of questionable reliability anyway.)

> I'm not sure if we satisfied this bit, though:
> 
> > 4\. As User B, quickly send a PM to User A.
> 
> How quickly does it seem to need to be to trigger the issue?

I haven't tested this extensively; I've just tried to keep it down to 5s or so.

gnprice:
> I don't recall if you had the PMing user's message-list open, though – I think you might have been at the PMs-list page.

Hmm, yeah.

I just tried again making sure of that step too, though -- so it was steps 1-4 as written above exactly -- and it still didn't reproduce; the notification arrived promptly as usual.

(Same Android 10 device as mentioned above; the app version is 26.11.134.)

[...]

PR #3704, opened 2019-11-26
Presence: Heartbeat

ray-kraesig:
~~Draft PR to centralize presence-reporting code. Not yet suitable for merge due to, at a minimum, the reasons noted in the final commit.~~

This may not be the heartbeat pattern we actually want, but if not, it should be easy to replace.

[edit by @gnprice: Fixes (or moots) #3699 ; part of addressing #3659 .]


ray-kraesig:
Commenting to make it unambiguous that this is no longer a draft PR.

(The exact heartbeat pattern desired is [still under discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/presence.20and.20notifications), but the structure introduced here isolates that, and should be separately reviewable.)

src/presence/heartbeat.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Heartbeat: Perform callbacks at regular intervals.
+ *
+ * While active, calls `callback` with `true` every `milliseconds` milliseconds.
+ * While inactive, does nothing.
+ *
+ * When transitioning between active and inactive states, performs one
+ * additional edge-triggered callback with `true` (if starting) or `false` (if
+ * stopping).
```
gnprice: Per the discussion linked and quoted above, I think we want one of these and not the other.

That probably makes most sense as a separate later commit anyway, though -- trying to make the switch to this implementation as close to NFC as it reasonably can be.

src/presence/__tests__/heartbeat-test.js
```diff
@@ -0,0 +1,162 @@
...
+  const HEARTBEAT_TIME = 60 * 60 * 1000;
+
+  // before running tests: set up fake timer API
+  beforeAll(() => {
+    jest.useFakeTimers();
+  });
+
+  // before each test: reset fake-timers state
+  beforeEach(() => {
+    jest.useFakeTimers();
```
gnprice: Do we want `useFakeTimers` in both of these, or just one?

src/boot/AppEventHandlers.js
```diff
@@ -131,7 +125,12 @@ class AppEventHandlers extends PureComponent<Props> {
   }
 
   render() {
-    return <View style={styles.wrapper}>{this.props.children}</View>;
+    return (
+      <>
```
gnprice: Huh interesting! What's this syntax do? Looks potentially handy elsewhere too.

(I searched for `<>` in the two most obvious pages of JSX docs -- https://reactjs.org/docs/introducing-jsx.html and https://reactjs.org/docs/jsx-in-depth.html -- with no luck.)

src/presence/HeartbeatComponent.js
```diff
@@ -0,0 +1,63 @@
...
+
+  heartbeat: Heartbeat = new Heartbeat(this.onHeartbeat, 1000 * 60);
+
+  componentDidMount() {
+    this.onStateChange();
+    AppState.addEventListener('change', this.onStateChange);
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener('change', this.onStateChange);
```
gnprice: What's the thinking behind using this vs. the `isActive` we maintain in Redux? (With the listener in `AppEventHandlers`.)

src/presence/HeartbeatComponent.js
```diff
@@ -0,0 +1,63 @@
...
+    this.heartbeat.stop();
```
gnprice: This looks asymmetrical. Should we be starting it in `componentDidMount`?

... Ah I see, we do that conditionally via `onStateChange`.

How about naming the latter something like `updateHeartbeatState`? I.e. named for what it does rather than when it's called. There are a couple of call sites that aren't the "state change" callback, and this name is a little puzzling at both of them.

src/presence/HeartbeatComponent.js
```diff
@@ -0,0 +1,63 @@
...
+|}>;
+
+/**
+ * Component providing a recurrent `presence` signal.
+ */
+// Note that "PureComponent" is of questionable veracity: this component's
+// entire purpose is to emit network calls for their observable side effects.
+// However, it is at least true that there is never a need to call `render()` if
+// the props haven't changed.
+class PresenceHeartbeat extends PureComponent<Props> {
```
gnprice: I think that's actually 100% consistent with what `PureComponent` is intended to mean:
https://reactjs.org/docs/react-api.html#reactpurecomponent

After all, almost all components are meant to cause something to appear in the UI; and any component with a widget like a button (that does something) is capable of causing quite general side effects, often a network request. All of that is independent of being a `PureComponent`. So the sense of "pure" in `PureComponent` has to be a lot narrower than "this is side-effect-free code".

I think in fact it really runs only in the dual direction: it says it's free of *suffering* certain kinds of side effects, but nothing about what side effects it may *cause*.

It could perhaps have a more lucid name than it does. :slightly_smiling_face:

src/boot/AppEventHandlers.js
ray-kraesig: It's just a grouping.
* https://reactjs.org/docs/fragments.html
* https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

src/presence/heartbeat.js
ray-kraesig: Ironically, the current observed (buggy) behavior is closer to the quote above than this implementation of Heartbeat is. It would be more `[nfc]`, from a certain perspective, to go straight there.

(I'd still rather fix things _before_ changing them, though.)

src/presence/__tests__/heartbeat-test.js
ray-kraesig: Just the first one. Thanks for the catch.

src/presence/HeartbeatComponent.js
ray-kraesig: Mostly the fact that `isActive` isn't really application-_controlled_ state. There are a lot of things that may happen in between that handler and a call to `render()` here.

(For example: if we get rehydrated with the wrong `isActive` state, what happens? If that's not possible, why not?)

src/presence/HeartbeatComponent.js
ray-kraesig: Works for me. (The asymmetry is worth a couple of comments, too.)

src/presence/HeartbeatComponent.js
ray-kraesig: > After all, almost all components are meant to cause something to appear in the UI; and any component with a widget like a button (that does something) is capable of causing quite general side effects, often a network request.

This is true, as far as it goes – but `PureComponent.render()` in particular is expected to do precisely none of those things, and instead only to return a specification of how and when one might later go about doing them.

That expectation is flagrantly violated by this component, whose `.render()` can, and often will, fire off a network request before returning. It's less pure than many `Component.render()` functions. The kindest thing one can say about it, I think, is that it is idempotent even when considered as a function from `World` to `World`.

(Well, modulo modifications of `this` and its prototype chain. But if you're breaking the equivalence of `this.foo.bind(this)` and `() => this.foo()`, then nothing holds anyway.)

ray-kraesig:
> > So I think this means that on going into the background, we should
> > 
> > * not send any update about that; but
> > * stop sending `active` updates, for as long as we're continuously in the background.
> > 
> > And when we come back to the foreground, if it's been over a minute, we should immediately send an `active` update.

I'm not entirely convinced that this is the right approach: it seems like the server could synthesize that from the more precise and accurate information stream (i.e., one with "idle" events) with very little overhead.

An implementing commit ~~is nonetheless attached~~ will follow nonetheless, once I have a sane test approach implemented. (And possibly once discussion settles down over on the Zulip channel, since I'm afraid I kicked it back up again.)

>  Not sure it simplifies things to pre-emptively take [the throttling] out of this layer as this separate commit, though.

It's one less thing in the `HeartbeatComponent` commit, at least. (I suppose it could be squashed into that one, but I'm not sure _that_ would simplify things, either. ¯\\\_(ツ)_/¯)

src/presence/HeartbeatComponent.js
gnprice: > (For example: if we get rehydrated with the wrong `isActive` state, what happens? If that's not possible, why not?)

That's not possible because `isActive` is more fully `state.session.isActive`, and the `session` subtree is one we don't persist. See `src/boot/store.js`:
```
/**
 * Properties on the global store which we explicitly choose not to persist.
 *
 * All properties on the global store should appear either here or in the
 * lists of properties we do persist, below.
 */
// prettier-ignore
export const discardKeys: Array<$Keys<GlobalState>> = [
  'alertWords', 'caughtUp', 'fetching', 'loading',
  'nav', 'presence', 'session', 'topics', 'typing', 'userStatus',
];
```


> Mostly the fact that `isActive` isn't really application-_controlled_ state.

Hmm, can you unpack what you mean by "controlled"? I might not understand what you mean. Or does the bit above about persistence affect this?

This state is completely determined by our code, specifically `sessionReducer.js`.

src/boot/AppEventHandlers.js
gnprice: TIL, thanks!

... I'll move to [a chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react.20fragments/near/800635).

src/presence/heartbeat.js
gnprice: > Ironically, the current observed (buggy) behavior is closer to the quote above than this implementation of Heartbeat is. It would be more `[nfc]`, from a certain perspective, to go straight there.

Hmm, I guess that's right! That might be a clean way to do it too, then.


> (I'd still rather fix things *before* changing them, though.)

Not sure I understand; isn't a fix a change?

src/presence/HeartbeatComponent.js
gnprice: > That expectation is flagrantly violated by this component, whose `.render()` can, and often will, fire off a network request before returning.

Hmm, I see.

I think that's a sign that `render` isn't the right place to put that call -- quite independently of `PureComponent` vs. its base class `Component`.

Consulting the API docs:
https://reactjs.org/docs/react-component.html
I think `componentDidUpdate` is the intended place to put this kind of call. That's based on its own description, and also on "use this instead" pointers to it in the descriptions of `getDerivedStateFromProps` and the deprecated `UNSAFE_componentWillReceiveProps`.

gnprice:
> it seems like the server could synthesize that from the more precise and accurate information stream (i.e., one with "idle" events) with very little overhead.

The thing is that "idle" doesn't mean "the user went away just now". Rather it means "hi, still here, but btw the user hasn't interacted with me (this client) recently."

... But really the better place for that design discussion is the chat thread; here we're in danger of overflowing what works well on a GitHub thread, as well as just getting mixed with the details of code. Want to follow up there?

src/presence/heartbeat.js
ray-kraesig: > Not sure I understand; isn't a fix a change?

A change to the code, but not necessarily to its specification. Perhaps read as: I'd rather fix the existing code (leaving the intended specification unchanged) before separately changing that specification (and the code along with it).

(This is decidedly situational. If, for example, the second step would have involved eliminating all the code in the first step, I would probably have a very different opinion.)

src/presence/HeartbeatComponent.js
ray-kraesig: :+1:

(The docs aren't explicit as to whether `componentDidUpdate` will fire if the `post-render()` reconciliation phase says there are no changes, but [it seems that it will](https://github.com/reactjs/reactjs.org/pull/1230).)

gnprice: Oof, that's a regrettable form for documentation to be in. :-/ Glad you tracked it down -- please mention it somewhere so that we, at least, can hang onto that information.

src/presence/heartbeat.js
gnprice: Got it. If I now understand you right, I think that means that

> (I'd still rather fix things _before_ changing them, though.)

was agreeing with

> That probably makes most sense as a separate later commit anyway, though

Is that right? I didn't read it that way the first time. :slightly_smiling_face:

src/presence/HeartbeatComponent.js
ray-kraesig: Sorry – the `.isActive` field itself is certainly owned and controlled by our app, but its value is a representation and a cache of underlying state that very much is not.

Notably, if the app is *not* active, it affects the cached value's staleness period in ways that are nontrivial to reason about – here, based on the decisions of two different schedulers, one of which is OS-dependent. To me that's an indication that this data probably shouldn't be in the store at all.

In this particular use case... honestly, it's probably not that important if we don't get the notice promptly. If there's a reason it *should* go through Redux, I've got a short patch I can squash into the relevant commit.

(On the other hand, I'd argue that the lone existing use of `.isActive`, over in `src/events/eventMiddleware.js`, is a bug! Or at least is closely involved in one. But that's not directly relevant here; I'll elaborate somewhere else.)

ray-kraesig:
Updated with new presence protocol, new tests, and a (hopefully very temporary) new dev-dependency.

package.json
```diff
@@ -108,6 +108,7 @@
     "jest-environment-jsdom": "^24.9.0",
     "jest-environment-jsdom-global": "^1.2.0",
     "jest-extended": "^0.11.2",
+    "lolex": "^5.1.1",
```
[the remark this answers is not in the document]
gnprice: Ahh, I see, the next commit has the answer.

Simplest thing may be to just squash the two, as the deps commit is very small.

Also unfortunately my reading of [that Jest PR](https://github.com/facebook/jest/pull/7776) would not be as optimistic as this. AFAICT there's been no reply from anyone at FB in the 10 months that PR has been open. Landing this PR, and doing so for Jest v25, is the author's plan but it's not clear it's anyone else's.

src/presence/__tests__/heartbeat-test.js
```diff
@@ -55,7 +55,7 @@ class Lolex {
 let lolex: Lolex;
 
 // type alias for Jest callback functions of type (boolean) => void
-type CallbackType = JestMockFn<$ReadOnlyArray<boolean>, void>;
+type CallbackType = JestMockFn<$ReadOnlyArray<void>, void>;
```
gnprice: This comment is out of date :slightly_smiling_face: 

Simplest fix is to update it in tandem.

But I think this also makes an excellent example of why it's good to avoid repeating details of the code in the comments. Here, the information this is really trying to add is something like "those type arguments are the types of the mock function's parameters and return." One way to say that would be
```
// Type parameters are: JestMockFn<TArguments: $ReadOnlyArray<*>, TReturn>
```

src/presence/__tests__/heartbeat-test.js
```diff
@@ -0,0 +1,243 @@
...
+    heartbeat: Heartbeat,
+  |} => {
+    const callback: CallbackType = jest.fn();
+    const heartbeat = new Heartbeat(callback, HEARTBEAT_TIME);
+    return { callback, heartbeat };
+  };
+
+  // Heartbeat erases its callback type (and it should probably be private
+  // anyway), so the `expectRunning` and `expectNotRunning` helpers take the
+  // callback mock as a separate well-typed argument.
```
gnprice: Huh, I'm puzzled -- what do you mean by saying it erases its callback type? Here's the definition in the early part of the branch:
```
class Heartbeat {
  intervalId: IntervalID | null = null;

  callback: (state: boolean) => void;
  milliseconds: number;
```

and a later commit does `s/state: boolean//`.

It seems like the type is right there. What am I missing?

src/presence/heartbeat.js
```diff
@@ -0,0 +1,68 @@
...
+  /** Start the heartbeat. Idempotent. */
+  start() {
+    if (this.isActive()) {
+      return;
+    }
+    if (this.previousTime + this.milliseconds <= Date.now()) {
+      this.doCallback();
+    }
+    this.intervalId = setInterval(this.doCallback, this.milliseconds);
+  }
```
gnprice: I think this can have a gap of up to twice nominal, if the state flaps shortly before the regular timer would fire.

gnprice:
BTW I also just added a link in the PR description to #3699, so that that thread gets an autolink back here. I was making a comment there, wanted to refer to this PR, and found I didn't have a reference to it :slightly_smiling_face:

src/presence/heartbeat.js
ray-kraesig: 2 - ε, yes, [as mentioned in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/presence.20and.20notifications/near/800273)... whereupon we talked a great deal about the possibility of too many presence signals, but much less about the possibility of not enough. :/

After a bit of poking and refactoring... hilariously, it looks like the simplest way to minimize that gap *also* removes any reason to call `Date.now`, and therefore makes Lolex useless here. Yay? ¯\\\_(ツ)_/¯

ray-kraesig: Addendum: the above is not true, and Lolex remains useful.

src/presence/__tests__/heartbeat-test.js
ray-kraesig: That's not the full type, though; it's just the maximal supertype. The "underlying type" of the `callback` object here is `CallbackType`.

*Not* erasing the type would involve giving `Heartbeat` a generic type parameter `T: (boolean) => void` (later `() => void`). This isn't difficult, of course; but it's also pointless unless `callback` is part of `Heartbeat`'s public interface. (Which it really shouldn't be!)

src/presence/__tests__/heartbeat-test.js
ray-kraesig: Unfortunately `*` is [deprecated](https://flow.org/en/docs/linting/rule-reference/#toc-deprecated-type). In particular I'm not sure that wouldn't have the effect of `ReadOnlyArray<any>`, which – while accurate for Jest – is probably not what I want here.

src/presence/HeartbeatComponent.js
ray-kraesig: It is now the case that this component performs side effects as a side effect of `render()`, rather than in `render()` directly.

Modulo that change, I think this warning remains appropriate.

src/boot/AppEventHandlers.js
```diff
@@ -131,7 +125,12 @@ class AppEventHandlers extends PureComponent<Props> {
...
+        <HeartbeatComponent />
+        <View style={styles.wrapper}>{this.props.children}</View>
```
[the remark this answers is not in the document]
ray-kraesig: (*from offline*) More like four or five different sibling components, I think. Agreed otherwise.

package.json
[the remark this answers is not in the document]
ray-kraesig: > Simplest thing may be to just squash the two, as the deps commit is very small.

Done!

> Also unfortunately my reading of that Jest PR would not be as optimistic as this.

The author in question is a collaborator for Jest (i.e., has commit privileges on the repo), so I have somewhat greater hopes of seeing this in for v25. 

(The new message for the squashed commit is somewhat more measured, nonetheless.)

ray-kraesig:
The new version is rather larger, as it contains more tests than were originally thought viable... and this GitHub PR is already approaching the limit of usefulness. Closing, to be reopened with a fresh slate and a pointer.

PR #3705, opened 2019-11-30
build-webview: script cleanup

ray-kraesig:
Some cleanup done to the `tools/build-webview` script, partly in preparation for more interesting functional changes.

This includes a new option (for testing) and a minor bugfix.

### DIFF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -45,6 +45,7 @@ cd "$ROOT"
 # Parse arguments. Sloppy, but sufficient for now.
 unset TARGET
 unset DEST
+SANITY_CHECKS=1
 while (( $# )); do
     case "$1" in
         --help|-h|help)
@@ -55,6 +56,9 @@ while (( $# )); do
         --destination)
             # note: this doesn't permit an equals sign after `--destination`
             shift; DEST="$1"; shift;;
+        --no-sanity-checks)
+            # disables target-directory sanity checks, for testing's sake
+            shift; SANITY_CHECKS=0;;
         *) usage; exit 2;;
     esac
 done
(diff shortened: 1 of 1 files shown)


ray-kraesig:
(Amended final commit to remove comment alterations that had crept in from another version. No functional changes.)

tools/build-webview
```diff
@@ -30,90 +32,108 @@ err() {
...
+# paths.
+root="$(git rev-parse --show-toplevel)"
+readonly root
+
+# This is our local staging directory; once it's assembled, it'll be synced over
+# to $dest.
+readonly staging="$root"/src/webview/static
+if [ ! -d "$staging" ]; then
+    err "cannot find asset staging directory '$staging'";
+fi
```
gnprice: I think this bit belongs more naturally around where it was, shortly before where it's used. One reason is that this isn't the right spot for this error check; if the command is `tools/build-webview --help`, or ill-formed, the right behavior is to exit with a syntax message, which means doing CLI parsing first.

Another is just that this feels closely tied to the details of how the build and sync work, so a natural thing to want to consult when reading that. That isn't a huge deal for this variable, because it's pretty simple; in the other PR, I felt it more strongly on the `z_rsync` function definition, which was also up here separated from the code that uses it.

ray-kraesig: Will do. (In the very near future, I'd like to pull the staging directory out of the middle of the source tree, and let `src/webview/static` be just another source directory for it. Once that's done, it would make some sense to be able to specify an arbitrary staging directory anyway – at least for testing – so there's another reason to keep it in that general area!)

gnprice: SGTM!

ray-kraesig:
I am quite certain that that button said "Rebase", _not_ "Squash", when I clicked it. :sweat:

gnprice:
(FTR in case it's helpful for anyone finding this via commit 37bd3e31a: Ray accidentally hit GitHub's "squash and merge" button, intending to rebase and merge instead. The individual steps of the branch can be seen in this PR thread, just above here.

We also then went and found the repo settings for those buttons, and turned off "squash and merge" entirely.)

ray-kraesig:
(Addendum: the first button I hit was indeed clearly labeled "Rebase and merge". The _second_ button I hit was merely labeled "Try Again".)

gnprice:
> (Addendum: the first button I hit was indeed clearly labeled "Rebase and merge". The _second_ button I hit was merely labeled "Try Again".)

Weird! Well, hopefully disabling "Squash and merge" will prevent that from behaving the same way in the future... but also this is maybe another in the line of reasons to treat the GitHub UI with suspicion. :-/

PR #3711, opened 2019-12-06
notifications: fix type of registration failure event

ray-kraesig:
The registration failure event object is not a string; treating it as
such yielded only `[Object object]` in the console. Use the actual
type.

This type has been confirmed accurate for rn-notifications@1.5.0 (our
current version) and rn-notifications@2.1.6 (the latest available at
time of writing), as well as by inspection in the console after this
change.


gnprice:
LGTM, thanks -- merged!

PR #3713, opened 2019-12-07

jainkuniya:


src/chat/Chat.js
```diff
@@ -31,6 +31,9 @@ const componentStyles = StyleSheet.create({
     flex: 1,
     flexDirection: 'column-reverse',
   },
+  composeBox: {
+    maxHeight: '60%',
```
jainkuniya: relative max height, so we don't have to worry about split screen.

gnprice:
Closing this. As described on #3614, the next step on this issue needs to be that we do some work to understand what's actually going wrong here.

If something were really seriously wrong from a user perspective and we had a hack that made it a lot better, then we'd certainly be open to merging the hack while we investigated the underlying cause. But that's what we've already done on this issue, with #3595 fixing #3369. This new hack is rather messier than the one we have, and it introduces new user-facing problems. It's time to focus on fixing the actual issue.

PR #3715, opened 2019-12-09
actions: fix typos in type of `notification_name`

ray-kraesig:
... and manually ensure that its declaration and use-site match.

(As seen Friday evening.)


gnprice:
Merged, thanks!

PR #3725, opened 2019-12-12
Linter tweaks

ray-kraesig:
Allow leading single underscores in object properties, as a stand-in for real private fields.

Also allow `for..of` loops and labeled statements. (I run into this about once a month, and end up going down a rabbit hole wondering why.)

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -65,6 +65,11 @@ rules:
   no-plusplus: off
   no-nested-ternary: off
 
+  # We also partly repeal AirBnB's forbidden-syntax subrules. (This must be done
+  # in the `overrides:` section.)
+  #
+  # no-restricted-syntax: ...
+
   # Likely-wrong code
   no-unused-vars: [warn, {vars: local, args: none}]
   # no-param-reassign seems well-intentioned... but fires on common uses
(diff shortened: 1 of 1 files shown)


ray-kraesig:
The CI failure occurs due to a parse error in `.eslintrc.yaml`. This parse error only occurs if `--full` is passed to the linter test – without it, the linter is not even run.

(Arguably, this is a bug in `tools/test` – a full lint should probably always be run if `.eslintrc.yaml`, or one of a handful of other files, has been touched.)

gnprice:
Wow, that anti-loop (!!?) stance in the thread airbnb/javascript#1103 is truly baffling to read. A fascinating find!

Particularly with the reasoning that combines it with being against GOTO. One has to wonder if the maintainers have ever actually read Dijkstra's [letter](http://www.u.arizona.edu/~rubinson/copyright_violations/Go_To_Considered_Harmful.html), or know what it was about.

gnprice:
> (Arguably, this is a bug in `tools/test` – a full lint should probably always be run if `.eslintrc.yaml`, or one of a handful of other files, has been touched.)

Yeah, I agree. It'd be pretty reasonable to add logic for that. I haven't just because it's fairly infrequent to be editing the lint config (and basically only maintainers do so), and so just remembering to do so by hand hasn't bothered me enough.

PR #3727, opened 2019-12-17
Presence heartbeat (v2)

### LINKED ISSUE #3699: Presence-reporting logic is confused, in overlapping ways

ray-kraesig:
Rewrite, and possibly unnecessarily thoroughly test, the presence-heartbeat code.

Resumption of PR #3704.

Fixes #3699.

### DIFF
--- a/src/presence/PresenceHeartbeat.js
+++ b/src/presence/PresenceHeartbeat.js
@@ -25,9 +25,9 @@ type Props = $ReadOnly<{|
 // changed -- which is exactly what we want.)
 class PresenceHeartbeat extends PureComponent<Props> {
   /** Callback for Heartbeat object. */
-  onHeartbeat = (state: boolean) => {
+  onHeartbeat = () => {
     if (this.props.auth) {
-      this.props.dispatch(reportPresence(state));
+      this.props.dispatch(reportPresence(true));
     }
   };
 
--- a/src/presence/PresenceHeartbeat.js
+++ b/src/presence/PresenceHeartbeat.js
@@ -1,0 +1,72 @@
+// @flow strict-local
+import { PureComponent } from 'react';
+import { AppState } from 'react-native';
+import type { Auth, Dispatch } from '../types';
+
+import { connect } from '../react-redux';
+import { tryGetAuth } from '../account/accountsSelectors';
+import { reportPresence } from '../actions';
+import Heartbeat from './heartbeat';
+
+type Props = $ReadOnly<{|
+  dispatch: Dispatch,
+  auth: Auth | void,
+|}>;
+
+/**
+ * Component providing a recurrent `presence` signal.
+ */
+// The name "PureComponent" is potentially misleading here, as this component is
+// in no reasonable sense "pure" -- its entire purpose is to emit network calls
+// for their observable side effects as a side effect of being rendered.
+//
+// (This is merely a misnomer on React's part, rather than a functional error. A
+// `PureComponent` is simply one which never updates except when its props have
+// changed -- which is exactly what we want.)
+class PresenceHeartbeat extends PureComponent<Props> {
+  /** Callback for Heartbeat object. */
+  onHeartbeat = (state: boolean) => {
+    // N.B.: If `auth` changes, we do not send out a final `false` presence
+    // status for the previous `auth`. It's the responsibility of our logout
+    // handler to determine whether that's necessary.
+    //
+    // (TODO: ensure that our logout handlers actually do that.)
+    if (this.props.auth) {
+      this.props.dispatch(reportPresence(state));
+    }
+  };
+
+  heartbeat: Heartbeat = new Heartbeat(this.onHeartbeat, 1000 * 60);
+
+  componentDidMount() {
+    this.updateHeartbeatState(); // conditional start
+    AppState.addEventListener('change', this.updateHeartbeatState);
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener('change', this.updateHeartbeatState);
+    this.heartbeat.stop(); // unconditional stop
+  }
+
+  // React to any state change.
+  updateHeartbeatState = () => {
+    // heartbeat.toState is idempotent
+    this.heartbeat.toState(AppState.currentState === 'active' && !!this.props.auth);
+  };
+
+  // React to props changes.
+  //
+  // Not dependent on `render()`'s return value, although the docs may not yet
+  // be clear on that. See: https://github.com/reactjs/reactjs.org/pull/1230.
+  componentDidUpdate() {
+    this.updateHeartbeatState();
+  }
+
+  render() {
+    return null;
+  }
+}
+
+export default connect(state => ({
+  auth: tryGetAuth(state),
+}))(PresenceHeartbeat);
--- a/src/presence/PresenceHeartbeat.js
+++ b/src/presence/PresenceHeartbeat.js
@@ -26,11 +26,6 @@ type Props = $ReadOnly<{|
 class PresenceHeartbeat extends PureComponent<Props> {
   /** Callback for Heartbeat object. */
   onHeartbeat = (state: boolean) => {
-    // N.B.: If `auth` changes, we do not send out a final `false` presence
-    // status for the previous `auth`. It's the responsibility of our logout
-    // handler to determine whether that's necessary.
-    //
-    // (TODO: ensure that our logout handlers actually do that.)
     if (this.props.auth) {
       this.props.dispatch(reportPresence(state));
     }
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -1,5 +1,4 @@
 /* @flow strict-local */
-import differenceInSeconds from 'date-fns/difference_in_seconds';
 import * as typing_status from '@zulip/shared/js/typing_status';
 
 import type { Dispatch, GetState, Narrow } from '../types';
@@ -9,8 +8,6 @@ import { getAuth, tryGetAuth } from '../selectors';
 import { isPrivateOrGroupNarrow, caseNarrowPartial } from '../utils/narrow';
 import { getAllUsersByEmail } from './userSelectors';
 
-let lastReportPresence = new Date(0);
-
 export const reportPresence = (isActive: boolean = true, newUserInput: boolean = false) => async (
   dispatch: Dispatch,
   getState: GetState,
@@ -19,13 +16,6 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
   if (!auth) {
     return; // not logged in
   }
-
-  const now = new Date();
-  if (differenceInSeconds(now, lastReportPresence) < 60) {
-    // TODO throttle properly; probably fold setInterval logic in here
-    return;
-  }
-  lastReportPresence = now;
 
   const response = await api.reportPresence(auth, isActive, newUserInput);
   dispatch({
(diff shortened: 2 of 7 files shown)


ray-kraesig:
The CI failure is due to a lint complaint involving `no-loop-func`. Which is fascinating, as I could have sworn I'd used that construct before... 

... ah, no, [it was inside a `.forEach`](https://github.com/zulip/zulip-mobile/blob/148767ab1/src/webview/js/__tests__/appendAuthToImages.js) then.

Disabled via linter pragmata, at least for now.

ray-kraesig:
The linter failure in the previous CI failure has been addressed by linter pragmata for now. (I thought I'd used that construct previously; but no, [that was inside a `.forEach`](https://github.com/zulip/zulip-mobile/blob/148767ab1/src/webview/js/__tests__/appendAuthToImages.js).)

The current CI failure is due to a tests-only auxiliary file being (not unreasonably) mistaken for an actual test file by Jest, because it's in a subdirectory of the `__tests__` folder that it's being used from. It clearly needs to be moved, but I see nowhere appropriate for it at present, and have no strong opinion about where it *should* go. A `dev-src/` directory at the root, perhaps?

ray-kraesig:
(Addendum: I haven't simply done that because, on reflection, the additional tests they're in support of may be more code than is worth having around. I'd prefer a second opinion on that.)

chrisbobbe:
I'm not sure I can speak to the necessity of those tests, but maybe for the Jest failure -- it's a bit of a patch, but perhaps the code in src/presence/__tests__/aux/interval-set.js doesn't need to be in a separate file; it's not very long, it's only imported by one file, heartbeat-test.js, and that file includes similar helper code (the Lolex class) that hasn't been deemed large enough to separate.

ray-kraesig:
> [...] and that file includes similar helper code (the Lolex class) that hasn't been deemed large enough to separate.

It probably should be, though: there are other timer-dependent tests (see: `rg 'Date\.now' -g'**/__tests__/**'`) which could, and probably should, make use of Lolex.

They may also want to make use of interval trees – in which case I'll have to properly separate out... well, a lot of things – but even then the interval-tree library wrapper should probably be a sub-library, just for the sake of separation of concerns.

(If I *just* wanted to silence the Jest failure, I could put a dummy test in that file.)

ray-kraesig:
Pinging @gnprice.

gnprice:
Marking priority, following the issue.

gnprice:
Thanks again @ray-kraesig for these changes! Now that the key parts of #3886 are in, want to rebase this to take advantage of Lolex being in master? (And the remaining commits of #3886 should be landing shortly, too, but I think those won't interact with this branch.)

I think the now-merged commits from #3886 also resolve the Jest file-layout questions discussed above. But please mention if I've missed something and that's not the case.

I'm also about to go read through this code -- I'll just skip over for now the bits that look like they were covered by #3886.

src/users/usersActions.js
```diff
@@ -20,14 +17,7 @@ export const reportPresence = (hasFocus: boolean = true, newUserInput: boolean =
     return; // not logged in
   }
 
-  const now = new Date();
-  if (differenceInSeconds(now, lastReportPresence) < 60) {
-    // TODO throttle properly; probably fold setInterval logic in here
```
gnprice: Commit-level comments on this commit:
e12c452df presence: strip throttling from `reportPresence` wrapper

Nit: "would fix #3699 as literally reported" is no longer quite right, as we've edited #3699 so it describes the problem more broadly.

Related nit: "would fix #..." is a dangerous phrase to put in a commit message, because GitHub sees the "fix" and doesn't understand the "would". :-) Should basically always get reworded to hide from GitHub's auto-close rules, with e.g. a circumlocution like "would be a fix for #...".

Stepping back, though, I think my comment on this commit from the previous PR still applies:
https://github.com/zulip/zulip-mobile/pull/3704#pullrequestreview-325863916
> This looks like a good implementation of what it says on the tin. I think it may not be part of what we want following today's discussion, though.

src/presence/PresenceHeartbeat.js
```diff
@@ -0,0 +1,66 @@
...
+
+  heartbeat: Heartbeat = new Heartbeat(this.onHeartbeat, 1000 * 60);
+
+  componentDidMount() {
+    this.updateHeartbeatState(); // conditional start
+    AppState.addEventListener('change', this.updateHeartbeatState);
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener('change', this.updateHeartbeatState);
```
gnprice: We had a comment thread on this bit over on the previous PR: https://github.com/zulip/zulip-mobile/pull/3704#discussion_r352959826

Your explanation there makes sense to me! I'm convinced.

I also wanted to mention that thread here because you'd said:
> (On the other hand, I'd argue that the lone existing use of `.isActive`, over in `src/events/eventMiddleware.js`, is a bug! Or at least is closely involved in one. But that's not directly relevant here; I'll elaborate somewhere else.)

and I don't want that thought to get lost. :slightly_smiling_face: Want to elaborate on that -- perhaps in a new issue, or a chat thread?

src/presence/PresenceHeartbeat.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Component providing a recurrent `presence` signal.
+ */
+// Note that "PureComponent" is of questionable veracity: this component's
+// entire purpose is to emit network calls for their observable side effects
+// as a side effect of being rendered.
+//
+// However, it is at least true that there is never a need to call `render()`
+// (and therefore `componentDidUpdate()`) if the props haven't changed.
+class PresenceHeartbeat extends PureComponent<Props> {
```
gnprice: On the previous PR we discussed a previous version of this comment: https://github.com/zulip/zulip-mobile/pull/3704#discussion_r352962392 Thanks for the updates!

I think it's pretty reasonable to say that the term "pure" here is confusing and of questionable veracity. I think what my thought here boils down to now is that that complaint is really all about the naming of `PureComponent`. What we're doing is 100% consistent with the expectations of a React `PureComponent`, which are [documented here](https://reactjs.org/docs/react-api.html#reactpurecomponent) -- it's just that those expectations don't include all the things that you might (reasonably!) expect the word "pure" to mean, or to be true of "a pure component".

A comment about that mismatch of `PureComponent` vs. the meaning of "pure" in general would be fine to have here. (Certainly it wouldn't be the first comment in our codebase that complains about a design choice that happened somewhere else :wink: -- I've written plenty of those.) What I don't want is a warning that we're hacking around or breaking some API expectation that potentially could expose us to problems when the React implementation changes, in a spot where in fact we aren't.

src/presence/PresenceHeartbeat.js
ray-kraesig: Eh, I might as well express my current thoughts on that in the form of commit messages.

Opened PR #3936.

src/presence/PresenceHeartbeat.js
ray-kraesig: > What I don't want is a warning that we're hacking around or breaking some API expectation that potentially could expose us to problems when the React implementation changes, in a spot where in fact we aren't.

Ah, I see the ambiguity now. Amended significantly.

src/users/usersActions.js
ray-kraesig: > Stepping back, though, I think my comment on this commit from the previous PR still applies:
> [#3704 (review)](https://github.com/zulip/zulip-mobile/pull/3704#pullrequestreview-325863916)
> 
> > This looks like a good implementation of what it says on the tin. I think it may not be part of what we want following today's discussion, though.

The fix for this is basically to squash the second commit into the fourth commit. As there's not much substance to it, done. (Although this does make the commit message uncomfortably crowded. :confused:)

ray-kraesig:
> I think it might be simplest to reserve those for a subsequent PR.

Fair enough! Snipped for now.

src/presence/PresenceHeartbeat.js
gnprice: New version looks great!

src/users/usersActions.js
gnprice: I like the result, thanks!

The commit message does get a bit longer. I think it might actually on net be simpler than before, though. One way it's certainly simpler is that the "If the user logs in and out" bullet previously had to tell a multi-step story ("This was previously concealed by ... which was itself removed in a previous patch") and that's now a simple before/after.

gnprice:
Thanks @ray-kraesig for the revisions!

One question and a nit, both in the commit messages.

> Fixes (the original version of) #3699, and is work towards #3659.

Seems to me like it actually fixes #3699. Is there a particular aspect of that that's still open that you have in mind?

Maybe this is a sign that my revised title was too vague :slightly_smiling_face: : "Presence-reporting logic is confused, in overlapping ways". I think the bar for closing the issue isn't that we become sure there are no bugs or confusions in the new version -- just that it's no longer confused in the ways we knew about. And of course a related good thing to achieve (as this PR does) is that it's written in a way that we can hope is less prone to confusion. :wink: 

    heartbeat: change presence-reporting protocol

    Following discussion on chat.zulip.org's #mobile-team channel, adjust

s/channel/stream/ :smile: 

Oh, and I guess a second nit: s/heartbeat/presence/ in that summary line would feel a bit more on-point to me. I think it's because this change is mainly aimed at the server-facing reporting of presence, rather than the internal implementation choices (even though it involves making changes to the latter to accomplish its changes to the former.)

ray-kraesig:
> Seems to me like it actually fixes #3699. Is there a particular aspect of that that's still open that you have in mind?

As of that commit, yes: `idle` messages are still sent, under the (existing) confusion that they should be. The following commit should have the relevant fix-note. (Done.)

gnprice:
> As of that commit, yes: `idle` messages are still sent, under the (existing) confusion that they should be. The following commit should have the relevant fix-note. (Done.)

Ah, got it. Thanks!

I'll merge shortly. I think I'll make one small further edit to the commit messages: on this "Fixes (the original version of) #3699" commit, I'll add a mention that a fix to the remainder of that issue is coming up next. Reduces the suspense a bit for the reader. :wink:

PR #3733, opened 2019-12-20
Add side-channel for Sentry logging

ray-kraesig:
... and utilize it to (hopefully) collapse many of our existing errors into single issues.

### DIFF
--- a/src/message/messageSelectors.js
+++ b/src/message/messageSelectors.js
@@ -9,16 +9,56 @@ import {
   getMute,
   getSubscriptions,
 } from '../directSelectors';
+import * as logging from '../utils/logging';
 import { getShownMessagesForNarrow } from '../chat/narrowsSelectors';
 import renderMessages from './renderMessages';
 import { findAnchor } from '../utils/message';
 import { ALL_PRIVATE_NARROW_STR } from '../utils/narrow';
 import { NULL_ARRAY } from '../nullObjects';
 
+/**
+ * Truncate a potentially-very-long array for logging and/or reporting purposes.
+ * Returns something which may or may not be an array, but is at least JSONable
+ * and human-readable.
+ */
+function truncateForLogging<T>(arr: Array<T>, len = 10) {
+  if (arr.length <= 2 * len) {
+    return arr;
+  }
+  return {
+    length: arr.length,
+    start: arr.slice(0, len),
+    end: arr.slice(-len),
+  };
+}
+
 export const getPrivateMessages: Selector<Message[]> = createSelector(
   getAllNarrows,
   getMessages,
-  (narrows, messages) => (narrows[ALL_PRIVATE_NARROW_STR] || NULL_ARRAY).map(id => messages[id]),
+  (narrows, messages) => {
+    const privateMessages: Message[] = [];
+    const unknownIds: number[] = [];
+
+    const pmIds = narrows[ALL_PRIVATE_NARROW_STR] || NULL_ARRAY;
+    pmIds.forEach(id => {
+      const msg = messages[id];
+      if (msg !== undefined) {
+        privateMessages.push(msg);
+      } else {
+        unknownIds.push(id);
+      }
+    });
+
+    // BUG (#3749): all messages in `narrows` _should_ also be in `messages`.
+    // Error reports indicate that, somehow, this isn't always so.
+    if (unknownIds.length > 0) {
+      logging.error('narrow IDs not found in state.messages', {
+        all_ids: truncateForLogging(pmIds),
+        unknown_ids: unknownIds,
+      });
+    }
+    return privateMessages;
+  },
 );
 
 export const getRenderedMessages = (narrow: Narrow): Selector<RenderedSectionDescriptor[]> =>
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -65,6 +65,7 @@ export const getAccountFromNotificationData = (
       urlMatches.push(i);
     }
   });
+
   if (urlMatches.length === 0) {
     // No match.  Either we logged out of this account and didn't
     // successfully tell the server to stop sending notifications (possibly
@@ -72,9 +73,10 @@ export const getAccountFromNotificationData = (
     // there's some confusion where the realm_uri we have is different from
     // the one the server sends in notifications.
     const knownUrls = identities.map(({ realm }) => realm);
-    logging.warn(
-      `notif realm_uri not found: ${realm_uri}, known values are ${JSON.stringify(knownUrls)}`,
-    );
+    logging.warn('notification realm_uri not found in accounts', {
+      realm_uri,
+      known_urls: knownUrls,
+    });
     return null;
   }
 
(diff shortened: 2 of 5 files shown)


src/utils/logging.js
```diff
@@ -52,61 +45,118 @@ const logToSentry = (event: string | Error, level: SeverityType): string => {
...
+  const toConsole = consoleMethod.bind(console);
+
+  return (event: string | Error, extras: Extras = {}) => {
+    logToSentry(event, Severity.Error, extras);
+
+    if (config.enableErrorConsoleLogging) {
+      toConsole(event);
+
+      // avoid sending auxiliary data to Sentry twice
+      if (!isSentryActive()) {
```
gnprice: Hmm! What's the story here?

It looks like this means that when Sentry is active, this information will *only* be available in Sentry -- not in `adb logcat` etc. If this information is useful when debugging at the console, then that could be quite annoying when e.g. trying to debug a published release version on someone's device.

src/notification/index.js
```diff
@@ -65,25 +65,30 @@ export const getAccountFromNotificationData = (
...
-    // except we don't store user IDs in `accounts`, only emails.  Until we
-    // fix that, just ignore the information.
-    logging.warn(`notif realm_uri ambiguous: ${urlMatches.length} matches for ${realm_uri}`);
+    // The user has several accounts in the notification's realm.  We should be
+    // able to tell the right one using the notification's `user_id`... except
+    // we don't store user IDs in `accounts`, only emails.  Until we fix that,
+    // just ignore the information.
+    logging.warn('notification realm_uri ambiguous; multiple matches found', {
+      match_count: urlMatches.length,
+      realm_uri,
```
gnprice: nit: Let's use the same name for the "`realm_uri` from notification" value in both the "<1 matches" and ">1 matches" cases.

src/message/messageSelectors.js
```diff
@@ -18,7 +19,30 @@ import { NULL_ARRAY } from '../nullObjects';
...
+      } else {
+        unknownIds.push(id);
+      }
+    });
+
+    // BUG: all messages in `narrows` _should_ also be in `messages`.
+    // Error reports indicate that, somehow, this isn't always so.
+    if (unknownIds.length > 0) {
+      logging.error('narrow IDs not found in state.messages', {
+        all_ids: pmIds,
```
gnprice: This seems like it could be a lot of data, and it's not obvious it'd be essential in studying the issue. Perhaps just the length?

src/message/messageSelectors.js
```diff
@@ -18,7 +19,30 @@ import { NULL_ARRAY } from '../nullObjects';
...
+        excess_ids: unknownIds,
```
gnprice: nit: Really more like "missing" than "excess", no?

src/message/messageSelectors.js
```diff
@@ -18,7 +19,30 @@ import { NULL_ARRAY } from '../nullObjects';
...
+      const msg = messages[id];
+      if (msg !== undefined) {
+        privateMessages.push(msg);
+      } else {
+        unknownIds.push(id);
+      }
+    });
+
+    // BUG: all messages in `narrows` _should_ also be in `messages`.
+    // Error reports indicate that, somehow, this isn't always so.
```
gnprice: Interesting! And puzzling.

Is there an issue for this? This would be good to file an issue for.

gnprice:
And I've pushed the first four commits from this branch, as:
841efbbf7 logging [nfc]: remove stray references in comments
281bf5f7d sentry [nfc]: explicitly initialize Sentry in index.js
52b6dffcc sentry [nfc]: move is-active test into the main code base
9f2e1be90 sentry [nfc]: add Scope and related functions to Sentry defs

src/utils/logging.js
```diff
@@ -52,61 +45,118 @@ const logToSentry = (event: string | Error, level: SeverityType): string => {
...
+  consoleMethod: mixed => void,
+  severity: SeverityType,
+|};
+type LogFunction = (event: string | Error, extras?: Extras) => void;
+
+const makeLogFunction = ({ consoleMethod, severity }: LogParams): LogFunction => {
+  const toConsole = consoleMethod.bind(console);
+
+  return (event: string | Error, extras: Extras = {}) => {
+    logToSentry(event, Severity.Error, extras);
```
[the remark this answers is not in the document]
ray-kraesig: Done! (Not sure why I didn't think to do that originally, really.)

src/utils/logging.js
```diff
@@ -52,61 +45,118 @@ const logToSentry = (event: string | Error, level: SeverityType): string => {
...
  * The error will be logged to Sentry and/or the console as appropriate.
  *
- * In a debug build, this pops up the RN error red-screen.  This is
- * appropriate when the condition should never happen and definitely
- * represents a bug.  For conditions that can happen without a bug (e.g. a
- * failure to reach the server), consider `logging.warn`.
+ * In a debug build, this pops up the RN error red-screen.  This is appropriate
+ * when the condition should never happen and definitely represents a bug.  For
+ * conditions that can happen without a bug (e.g. a failure to reach the
+ * server), consider `logging.warn`.
```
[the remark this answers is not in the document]
ray-kraesig: It doesn't; the changes here and below are only due to the fact that I hit <kbd>Shift</kbd>-<kbd>Meta</kbd>-<kbd>W</kbd> to clean up the parameter descriptions, and the main body came along for the ride.

As for the others – I only found two – I'm neither sure how they got rewrapped, nor how I missed noticing them at commit time. (I do habitually use `git add -p`, so they should have shown up.) Reverted.

(Personally, I find 70 characters to be on the uncomfortably-short side; even the ~90-character lines in this message box feel a bit cramped. Ah well.)

src/utils/logging.js
ray-kraesig: Having done some digging, I think this deserves some additional testing and discussion which is mostly outside the scope of this PR.

Removed for other, more disappointing reasons.

src/message/messageSelectors.js
ray-kraesig: I suspect we may want a little more information than that. Trimmed to also provide the first and last 10 entries.

src/message/messageSelectors.js
ray-kraesig: My line of thinking was that `pmIds` was the minuend and `messages` the subtrahend. But sure, that works too.

(Or actually, `unknown_ids`, to match the variable name.)

src/message/messageSelectors.js
ray-kraesig: It would, and I should have filed one. Done, as #3749.

src/utils/logging.js
```diff
@@ -1,28 +1,20 @@
...
+
+// Wrapper for `Sentry.withScope`, allowing callbacks to return values.
+function withScope<R>(callback: Scope => R): R {
+  let ret: R;
+  withScopeImpl(scope => {
+    ret = callback(scope);
+  });
+  // Flow can't know that `ret` has actually been initialized
+  return ((ret: $FlowFixMe): R);
+}
```
gnprice: nit: this (and the imports) should be in the next commit

src/utils/logging.js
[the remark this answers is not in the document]
gnprice: :+1: Thanks!

src/message/messageSelectors.js
gnprice: Thanks!

Let's put that issue number in the comment, too.

gnprice:
And merged -- thanks again!

PR #3734, opened 2019-12-20
Permit use of commas in stream names

### LINKED ISSUE #3729: Can't send messages to streams with commas in the name
If the user attempts to send a message to a stream which has at least one comma in the name, that message will never be sent.

(This is arguably a server-side bug, arising from the Zulip server parsing the `to` field as a comma-separated list even in `stream`-targeted messages. We'll still need a workaround, though.)

ray-kraesig:
Should fix #3729.

This is just a draft PR because it doesn't take into account the problem of what to do about existing messages in the outbox. (We *probably* don't want to flush a potential backlog of hundreds of messages all at once. Perhaps, during the following migration, we should drop messages stalled due to this bug that are more than a day old?)

### DIFF
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -52,15 +52,14 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
   const outboxToSend = state.outbox.filter(outbox => !outbox.isSent);
   try {
     outboxToSend.forEach(async item => {
-      await api.sendMessage(
-        auth,
-        item.type,
-        isPrivateOrGroupNarrow(item.narrow) ? item.narrow[0].operand : item.display_recipient,
-        item.subject,
-        item.markdownContent,
-        item.timestamp,
-        state.session.eventQueueId,
-      );
+      await api.sendMessage(auth, {
+        type: item.type,
+        to: isPrivateOrGroupNarrow(item.narrow) ? item.narrow[0].operand : item.display_recipient,
+        subject: item.subject,
+        content: item.markdownContent,
+        localId: item.timestamp,
+        eventQueueId: state.session.eventQueueId,
+      });
       dispatch(messageSendComplete(item.timestamp));
     });
     return true;
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -64,9 +64,21 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
         return; // i.e., continue
       }
 
+      const to = ((): string => {
+        const { narrow } = item;
+        // TODO: can this test be `if (item.type === private)`?
+        if (isPrivateOrGroupNarrow(narrow)) {
+          return narrow[0].operand;
+        } else {
+          // HACK: the server attempts to interpret this argument as JSON, then
+          // CSV, then a literal. To avoid misparsing, always use JSON.
+          return JSON.stringify([item.display_recipient]);
+        }
+      })();
+
       await api.sendMessage(auth, {
         type: item.type,
-        to: isPrivateOrGroupNarrow(item.narrow) ? item.narrow[0].operand : item.display_recipient,
+        to,
         subject: item.subject,
         content: item.markdownContent,
         localId: item.timestamp,
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -50,8 +50,20 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
   const state = getState();
   const auth = getAuth(state);
   const outboxToSend = state.outbox.filter(outbox => !outbox.isSent);
+  const oneWeekAgoTimestamp = Date.now() / 1000 - 60 * 60 * 24 * 7;
   try {
     outboxToSend.forEach(async item => {
+      // If a message has spent over a week in the outbox, it's probably too
+      // stale to try sending it.
+      //
+      // TODO: instead of just throwing these away, create an "unsendable" state
+      // (including a reason for unsendability), and transition old messages to
+      // that instead.
+      if (item.timestamp < oneWeekAgoTimestamp) {
+        dispatch(deleteOutboxMessage(item.id));
+        return; // i.e., continue
+      }
+
       await api.sendMessage(auth, {
         type: item.type,
         to: isPrivateOrGroupNarrow(item.narrow) ? item.narrow[0].operand : item.display_recipient,
--- a/src/api/messages/sendMessage.js
+++ b/src/api/messages/sendMessage.js
@@ -6,18 +6,20 @@ import { apiPost } from '../apiFetch';
 /** See https://zulipchat.com/api/send-message */
 export default async (
   auth: Auth,
-  type: 'private' | 'stream',
-  to: string,
-  subject: string,
-  content: string,
-  localId: number,
-  eventQueueId: number,
+  params: {|
+    type: 'private' | 'stream',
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks! Ouch, yeah.

I think a good solution to that last question is: in a commit before the one that fixes this bug, we add some logic for `trySendMessages` to just drop outbox messages that are ancient, like >1wk old. (And log something to Sentry when they do.)

ray-kraesig:
On reflection, the last two commits ("send messages sequentially" and "continue sending messages on client errors") are arguably a little out of scope.

If there are issues with them, but not with the first three commits, it's probably better to close this PR with just those three merged – the other two can wait for their own issue+PR.

src/outbox/outboxActions.js
```diff
@@ -46,23 +47,57 @@ export const messageSendComplete = (localMessageId: number): Action => ({
...
+
+      const to = ((): string => {
+        const { narrow } = item;
+        // TODO: can this test be `if (item.type === private)`?
+        if (isPrivateOrGroupNarrow(narrow)) {
+          return narrow[0].operand;
+        } else {
+          // HACK: the server attempts to interpret this argument as JSON, then
+          // CSV, then a literal. To avoid misparsing, always use JSON.
+          return JSON.stringify([item.display_recipient]);
```
gnprice: Huh, interesting that this is a list -- so I think it'll be like `["general"]`. Is that really the right form?

The API docs, fwiw, say (at https://zulipchat.com/api/send-message):
> The destination stream, or a CSV/JSON-encoded list containing the usernames (emails) of the recipients.

which suggests a list is only applicable for PMs.

When changing the format of what we send in a request, it'd also be good to confirm that older server versions already accept the new format; then if they don't, we'll figure out how to handle that.

src/outbox/outboxActions.js
```diff
@@ -46,23 +47,57 @@ export const messageSendComplete = (localMessageId: number): Action => ({
...
+      // that instead.
+      if (item.timestamp < oneWeekAgo) {
+        dispatch(deleteOutboxMessage(item.id));
+        continue;
+      }
+
+      const to = ((): string => {
+        const { narrow } = item;
+        // TODO: can this test be `if (item.type === private)`?
+        if (isPrivateOrGroupNarrow(narrow)) {
```
gnprice: Yeah, it's kind of odd that this code (already in the status quo) uses a mix of `narrow` and `display_recipient`.

I feel like `narrow` shouldn't be necessary -- that information should be covered by `display_recipient`. Besides which, `narrow` is kind of a confused thing to use here, because a narrow doesn't necessarily correspond to any one conversation.

src/outbox/outboxActions.js
```diff
@@ -46,23 +47,57 @@ export const messageSendComplete = (localMessageId: number): Action => ({
   local_message_id: localMessageId,
 });
 
-export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean => {
+export const trySendMessages = async (dispatch: Dispatch, getState: GetState): Promise<boolean> => {
   const state = getState();
   const auth = getAuth(state);
   const outboxToSend = state.outbox.filter(outbox => !outbox.isSent);
+  const oneWeekAgo = Date.now() / 1000 - 60 * 60 * 24 * 7;
```
gnprice: It always makes me a little nervous to see unit-conversion factors like this `/ 1000` appear silently -- it's too easy to compare one thing that's secretly in seconds to another that's secretly in milliseconds, and wind up like that [muffed Mars meteorology mission](https://www.wired.com/2010/11/1110mars-climate-observer-report/).

If we were writing in Haskell, I'd encode this in the type system, with a `newtype` that was an instance of the numeric typeclasses. In Rust or C++ I'd try to do something equivalent.

In the type system we have, though, I don't know of a good way to do that. So what we can't say with types, we say with names: I've often suffixed time variables with `Ms` to indicate they're in milliseconds.

For seconds, we could say `oneWeekAgoSec`. Or `oneWeekAgoTimestamp`, and declare that "timestamp" for us means "Unix seconds-since-epoch" -- since that's what it means in the `timestamp` property on `Message` which we get straight from the Zulip API.

src/outbox/outboxActions.js
gnprice: (This doesn't call for any code change in this PR; just replying to the thought in the TODO comment.)

ray-kraesig: I think the *right* thing to do here is to split private messages from stream messages in the type system, and provide a canonical discriminator method.

I tried to do that at one point, only to discover that it's going to require a later version of Flow than we have. (v0.100, I think? One of the versions with object-spread improvements, anyway.)

src/outbox/outboxActions.js
ray-kraesig: JSON has been accepted since at least zulip/zulip@97d7d31b684aae36d1406b9157fc230d6d5f8e93 – sometime in the Humbug 0.1.4 period, in mid-2013 – and given the presence of `json_to_list` in the "before" side of that diff, it's probably been accepted for longer than that.

(I'll make a note in the commit message.)

> [...] which suggests a list is only applicable for PMs.

Arguably it's a server bug that it tries to interpret `to` as a list even for stream messages, but that is indeed what it does. (Reported as zulip/zulip#13836.)

ray-kraesig:
> Let's move the last two commits to a separate PR, as you suggested -- they have some complexity, so let's narrow the scope for this bugfix.

After poking at the relevant code a bit, I'm going to have to abandon those commits entirely. The bug they're trying to fix presently masks the potentially severe effects of other bugs. :disappointed: I don't think there's a simpler way to fix the control flow here than a total rewrite.

I'll file a new issue with some notes. In the meantime, :ping_pong:.

src/outbox/outboxActions.js
gnprice: > I think the _right_ thing to do here is to split private messages from stream messages in the type system, and provide a canonical discriminator method.

That would certainly also be good!

src/outbox/outboxActions.js
gnprice: Cool. That sure is reassuringly old. :-)

I think then my remaining question is: the commit message says "Always use JSON" for this field; but it looks like the code actually uses JSON just for the stream case, and continues to use a simple CSV for the PMs case.

I think I'd be OK with that difference between the two cases; but I'd want to have the commit message, and the comments, be in agreement with what's happening.

Alternatively, I think it would work well to do as the commit message says: always use JSON. More specifically, always use a JSON-encoded list. In the PMs case, I think this would be just something like
```
- return narrow[0].operand;
+ return JSON.stringify(narrow[0].operand.split(','));
```
(And then perhaps celebrate the shared `JSON.stringify` part by pulling it out to deduplicate.)

gnprice:
Thanks for the revisions! Replied above.

ray-kraesig:
> I think then my remaining question is: the commit message says "Always use JSON" for this field [...]

Clarified. (The other option would have required testing.)

ray-kraesig:
Withdrawn pending [research and discussion.](https://chat.zulip.org/#narrow/stream/3-backend/topic/messages.20api.20%28commas%29/near/815377). The short version is that we may want to use a stream ID here rather than a stream name.

ray-kraesig:
The ability to use stream IDs was only added in 2.0 (zulip/zulip@b8221555d2225594c6fb8b9bbcbb63ce3cedbc87). While it might be a good idea once we can do version detection, we'll have to make do with JSONified names for now.

gnprice:
Looks good, thanks -- merged!

PR #3735, opened 2019-12-20
webview css: use `-webkit-animation` properties for compatibility

### LINKED ISSUE #3730: Unsendable messages have frozen spinners
Unsendable messages appear in the message list with a frozen spinner, as shown below.

Screenshot_20191217_163124

(This is a reprise, but apparently not a duplicate, of #1546.)

ray-kraesig:
The `animation` property doesn't appear to have been available without this prefix until Chrome 43, and we presently try to support as far back as Chrome 37.

Fixes #3730.

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -250,11 +250,16 @@ hr {
     hsla(0, 0%, 50%, 0.5) 100%
   );
   background-size: 200% 200%;
+  -webkit-animation: gradient-scroll 1s linear infinite;
   animation: gradient-scroll 1s linear infinite;
 
   border-radius: 10px;
   height: 8px;
   margin-bottom: 10px;
+}
+@-webkit-keyframes gradient-scroll {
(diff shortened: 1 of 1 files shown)


gnprice:
Hmm, this is quite a fair bit of duplication: 23 lines across 8 places in the file.

Here's what we say (in `js.js`) about old Chrome versions:
```
 *   Graceful degradation is acceptable below Chrome 58. [...]
 *
```

I think the complexity of this fix falls well above the "trivial" threshold, in terms of its maintenance cost for ongoing changes to our CSS styles.

Happily the issue is also fairly minor: it doesn't affect the actual functionality, and most users will rarely to never see it at all. So I think it's the kind of "graceful degradation" I'd be OK with at any Chrome version below 58, when the fix isn't trivial.

Thanks for looking into this! I think based on the information collected, the best path is to close the PR and issue (as WONTFIX), and turn attention to other bugs and features.

PR #3736, opened 2019-12-20
WebView: Implement passive event listeners for scroll in webviews

### LINKED ISSUE #3440: Make our event listeners passive, for smoother scrolling
Our message list generally scrolls quite smoothly for me, on my Pixel 2 XL. But we sometimes hear reports that it's not.

And, it turns out, there is a simple thing we can do that should improve scrolling! Most of our event handlers should be "passive". See [explainer doc](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md) on the feature for description and rationale. For specifics, see MDN API doc on [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener).

NB this feature apparently gained support in Safari only with iOS 10 (per the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Browser_compatibility) of that last doc.) That is over 99% of our iOS users; and virtually all our Android users are covered, because Chrome auto-updates. So it's OK that the benefit of this will only be for users on those newer browsers -- but we do support iOS 9, so we should make sure we don't just break, by doing feature detection as described in those docs.

agrawal-d:
This PR improves scroll performance while scrolling through message list WebViews, by replacing event listeners with passive event listeners.

Closes #3440

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -719,18 +719,54 @@ const handleLongPress = (target: Element) => {
   });
 };
 
-documentBody.addEventListener('touchstart', (e: TouchEvent) => {
-  const { target } = e;
-  if (e.changedTouches[0].pageX < 20 || !(target instanceof Element)) {
-    return;
+/**
+ * Feature-detection for passive event listeners, to improve scrolling
+ * performance.
+ *
+ * The method addEventListener() can accept an optional boolean flag, passive,
+ * which if true, indicates that the function specified by listener will never
+ * call preventDefault(), which can improve scroll performance in some cases.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener.
+ *
+ * Not all events gain performance benifits from this. Lighthouse, the Chrome
+ * performance audit tool, has identified the following events that can benefit
+ * from this flag : wheel, mousewheel, touchstart, and touchmove. See
+ * https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners#recommendations.
+ */
+const passiveIfSupported: EventListenerOptionsOrUseCapture = (() => {
+  let checkPassive = false;
+  try {
+    // $FlowFixMe (https://github.com/facebook/flow/issues/285)
+    const opts = Object.defineProperty({}, 'passive', {
+      get() {
+        checkPassive = true;
+        return;
+      },
+    });
+    window.addEventListener('testPassive', null, opts);
+    window.removeEventListener('testPassive', null, opts);
+  } catch (e) {
+    // no manipulation required for the caught exception.
   }
+  return checkPassive ? { passive: true } : false;
+})();
 
-  lastTouchPositionX = e.changedTouches[0].pageX;
-  lastTouchPositionY = e.changedTouches[0].pageY;
-  hasLongPressed = false;
-  clearTimeout(longPressTimeout);
-  longPressTimeout = setTimeout(() => handleLongPress(target), 500);
-});
+documentBody.addEventListener(
+  'touchstart',
+  (e: TouchEvent) => {
+    const { target } = e;
+    if (e.changedTouches[0].pageX < 20 || !(target instanceof Element)) {
+      return;
+    }
+
+    lastTouchPositionX = e.changedTouches[0].pageX;
+    lastTouchPositionY = e.changedTouches[0].pageY;
+    hasLongPressed = false;
+    clearTimeout(longPressTimeout);
+    longPressTimeout = setTimeout(() => handleLongPress(target), 500);
+  },
+  passiveIfSupported,
+);
 
 const isNearPositions = (x1: number = 0, y1: number = 0, x2: number = 0, y2: number = 0): boolean =>
   Math.abs(x1 - x2) < 10 && Math.abs(y1 - y2) < 10;
@@ -752,9 +788,13 @@ documentBody.addEventListener('touchcancel', (e: TouchEvent) => {
   clearTimeout(longPressTimeout);
 });
 
-documentBody.addEventListener('touchmove', (e: TouchEvent) => {
-  clearTimeout(longPressTimeout);
-});
+documentBody.addEventListener(
+  'touchmove',
+  (e: TouchEvent) => {
+    clearTimeout(longPressTimeout);
+  },
+  passiveIfSupported,
+);
 
 documentBody.addEventListener('drag', (e: DragEvent) => {
   clearTimeout(longPressTimeout);
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -597,6 +597,25 @@ var compiledWebviewJs = (function (exports) {
     });
   };
 
+  var passiveIfSupported = function () {
+    var checkPassive = false;
+
+    try {
+      var opts = Object.defineProperty({}, 'passive', {
+        get: function get() {
+          checkPassive = true;
+          return;
+        }
+      });
+      window.addEventListener('testPassive', null, opts);
+      window.removeEventListener('testPassive', null, opts);
+    } catch (e) {}
+
+    return checkPassive ? {
+      passive: true
+    } : false;
+  }();
+
   documentBody.addEventListener('touchstart', function (e) {
     var target = e.target;
 
@@ -611,7 +630,7 @@ var compiledWebviewJs = (function (exports) {
     longPressTimeout = setTimeout(function () {
       return handleLongPress(target);
     }, 500);
-  });
+  }, passiveIfSupported);
 
   var isNearPositions = function isNearPositions() {
     var x1 = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 0;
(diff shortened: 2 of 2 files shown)


agrawal-d:
@ray-kraesig I have tried to implement the changes you requested. Please review them. The feature detection method mentioned here https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection causes an error to be shown by Flow, so I have added the comment `$FlowFixMe` to supress it.

> In turn, this suggests that we need some confirmation that the suggested change actually helps with our slow scrolling issues, rather than taking it for granted that it will because we think it should.

There is a way to inspect the FPS of the WebView using Chrome Dev Tools Remote Debugging.  I wlil try to see if there are visible performance improvements by using passive event listeners. (https://developers.google.com/web/tools/chrome-devtools/evaluate-performance#analyze_frames_per_second).

Although the Google Developers article here (https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners) mentions that passive event listeners *do* improve performance.

ray-kraesig:
> @ray-kraesig I have tried to implement the changes you requested. Please review them.

Looks OK, except as noted inline and below.

> Although the Google Developers article here (https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners) mentions that passive event listeners _do_ improve performance.

The event listener that citation uses as an example is `touchstart`, not `scroll`. I don't think we have any reason to believe that using a passive event listener on `scroll` helps at all – it's still my understanding that, since scroll events are [already noncancelable](https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event), all `scroll` event listeners are effectively "passive" anyway.

agrawal-d:
You are right, it appears that adding the passive flag to event listeners for scroll event offers no performance gains. It is explicitly mentioned here : (https://developers.google.com/web/updates/2016/06/passive-event-listeners).

So, should I continue to work on this ( make the changes you requested) or should the related issue be closed?

src/webview/js/js.js
```diff
@@ -413,7 +413,29 @@ const handleScrollEvent = () => {
   showHideElement('scroll-bottom', nearEnd);
 };
 
-window.addEventListener('scroll', handleScrollEvent);
+// Feature-detection for passive event listeners, to improve scrolling performance
+(() => {
+  let supportsPassive = false;
```
agrawal-d: IIFE was added to prevent leaking of `supportsPassive` into the global scope.

ray-kraesig:
> You are right, it appears that adding the passive flag to event listeners for scroll event offers no performance gains.
> 
> So, should I continue to work on this ( make the changes you requested) or should the related issue be closed?

Making the *`scroll`* event handler passive may not be helpful – but making the various cancelable `touch*` (_etc._?) event handlers passive, as the documentation suggests, is another matter. I don't think the related issue should be closed unless that's also determined not to be useful.

> IIFE was added to prevent leaking of `supportsPassive` into the global scope.

Yes, but a bare `{}`-block would work just as well for that, in this context – unlike `var`, `let` and `const` [are block-scoped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Description).

(Of course, depending on whether and how you reorganize the code to apply to multiple event handlers, the IIFE may yet be useful.)

agrawal-d:
Oops! I didn't see that there were other event listeners in this file!

So, I am thinking of doing this: I will allow the ```supportsPassive``` variable to remain in the global scope, and then add the flag to all the event listeners in the file that would benefit from it ( since the listeners are initiated at different line numbers in the file).

Does this sound right? If so, I will proceed with it. If not,what do you suggest?

ray-kraesig:
Offhand, it seems like it would be cleaner to put the result of testing the flag in the global scope (e.g., `return supportsPassive ? {...} : false` from the IIFE), rather than testing it for every event handler. Other than that, sounds good.

A potentially useful fact (as the resultant variable will probably need a type): although the obvious type of the value `false` is `boolean`, Flow also has a [literal type](https://flow.org/en/docs/types/literals/) `false` – in effect, the type `boolean` is equivalent to the type `false | true`.

agrawal-d:
I have made the required changes.

Passive flag has only been added to those events, in which a visible performance benefit has been observed in the Chrome Lighthouse performance tool as mentioned [here](https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners).

agrawal-d:
Will make some additional changes.

agrawal-d:
I have made some changes.

You mentioned that flow supports literal types. When I tried adding one to the `passiveIfSupported` variable, it seems flow got confused, giving the following error wherever it was used with an event listener.

Variable declaration [Line 417] :
 `const passiveIfSupported: boolean | {| passive: boolean |} = (() => {...`

Flow error : http://p.ip.fi/a1pt

But I think that not adding the type here can be acceptable, because flow is able to infer the type automatically ( as can be seen by hovering over the variable name in VS Code ). 

Let me know what you think, and if any changes are needed.

agrawal-d:
I _think_ I understand the cause of the error now.

Flow requires  type 
```
type EventListenerOptionsOrUseCapture =
  | boolean
  | { capture?: boolean, once?: boolean, passive?: boolean } 
``` 
to be used with `addEventListener`

So, if any of the properties ( in this case, _capture_ and _once_) is explicitly missing, it will not be able to match the types.

If adding a type to the variable `passiveIfSupported` is a *must*, then using ```EventListenerOptionsOrUseCapture```  as a type works.

Please suggest the best recourse.

ray-kraesig:
Well, that's an interesting and disappointing bug. 

Its rejection of `false | {| passive: boolean |}` as incompatible is correct, at least going by the type signatures it has available: as far as the function signature says, `addEventListener` could mutate its final argument by giving it a `capture` value. `false | { passive: boolean }` is also formally incompatible, as `addEventListener` could remove the `passive` property.

However, it also rejects something as broad as `boolean | { passive?: boolean, ... }` which doesn't seem right at all.

(Minimal repro, for future reference:

https://flow.org/try/#0C4TwDgpgBA8mwEsD2A7AzlAvFA3gKCihAjQH4AuKAIySQBsIBDFAGgKhSQutoebYC+Abjx4AZgFcUAY0SoojWQAok8ZOkpw56AJS5ho6ajTAoq4Bmw4iJSsABOE6AaPpT5tACZK14mUo09EysUAB04QJYuDZodo7OIniKwCrwaDoiyakWnhl4QA)

ray-kraesig:
> But I think that not adding the type here can be acceptable, because flow is able to infer the type automatically ( as can be seen by hovering over the variable name in VS Code ).

Flow will infer the type you actually wrote, whether or not it's the type you intended. This probably isn't a problem in this particular case and at this particular time; but there are enough untyped, too-broadly-typed, and correctly-typed-but-poorly-designed APIs in the world that it's probably best to use explicit types for any interesting intermediate values.

> If adding a type to the variable `passiveIfSupported` is a _must_, then using `EventListenerOptionsOrUseCapture` as a type works.

This is actually a much better solution than just giving it an inline anonymous type, as it also gives the reader a reference to help explain why the type is what it is. (Not a great reference, admittedly, but better than none.)

> Let me know what you think, and if any changes are needed.

In addition to the above, this should probably be moved down to where the touch-event handlers are registered.

(The name `passiveIfSupported` is also a bit opaque, but acceptable if it's close to all of its use-sites. The EventListener API being what it is, I'm not sure there's a good name for that value at all. :disappointed:)

agrawal-d:
@ray-kraesig I have made the changes.

gnprice:
Thanks @agrawal-d , and thanks @ray-kraesig for the reviews!

One key piece I'd like to see added to this commit before it's merged: we should include in it the explanation of how we determined that those two event types (`touchstart` and `touchmove`) are the two that we wanted to apply this option to, and that we don't want to apply it to any others. That's for a few, partly overlapping reasons:
* It helps anyone reading the commit -- like ourselves in the future, when making other changes here -- see what our thinking was.
* Writing down how we made a choice helps make sure we thought it through in the first place. :wink: 
* Whenever we add other event listeners in the future, writing this down helps us figure out whether they should use this option.

Probably the best immediate place for that explanation is in a comment on `passiveIfSupported`. I'm imagining it might be a paragraph or two, with one or a few links to relevant sources we relied on.

@agrawal-d if you'd like to make a go at adding that explanation, that'd be great. If you don't get to it, it's also a normal part of our process that the reviewer (@ray-kraesig) might add this kind of explanation themselves, in the process of merging.

agrawal-d:
@ray-kraesig Added an explanatory comment block.

ray-kraesig:
I was just about ready to merge... and then I found this:

https://developers.google.com/web/updates/2017/01/scrolling-intervention
https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html

This indicates that, at the time #3440 was filed (2019-04-01), the vast majority of our users should already have been on a default-passive version of Chrome (56.0.2924, released 2017-02-01) or Safari (11.1, released with iOS 11.3, 2018-03-29).

@gnprice: you mentioned in #3440 that users had been complaining about scrolling being slow. Was that contemporary to the bug being filed, or were those complaints a collection of older issues?

gnprice:
> I was just about ready to merge... and then I found this:

Gosh -- good catch! Replied on the issue, #3440; it looks like we don't need to / shouldn't make this change after all.

Thanks for the investigation, and the reviews before that -- and thanks again @agrawal-d for writing up this PR. Sorry the issue pointed you on a path that turned out not to be a change that made sense to make. Hopefully we'll see you take on another one next. :smile:

PR #3738, opened 2019-12-24
DevAuthScreen: Use FlatList to list administrators login options.

22sujata8d:
Before:
Screenshot 2019-12-24 at 3 58 55 PM

After:
Screenshot 2019-12-24 at 4 03 08 PM

### DIFF
--- a/src/start/DevAuthScreen.js
+++ b/src/start/DevAuthScreen.js
@@ -92,13 +92,14 @@ class DevAuthScreen extends PureComponent<Props, State> {
             style={[styles.field, componentStyles.heading2, componentStyles.heading]}
             text="Administrators"
           />
-          {directAdmins.map(admin => (
-            <ZulipButton
-              key={admin.email}
-              text={admin.email}
-              onPress={() => this.tryDevLogin(admin.email)}
-            />
-          ))}
+          <FlatList
+            data={directAdmins.map(admin => admin.email)}
+            keyExtractor={(item, index) => item}
+            ItemSeparatorComponent={() => <View style={componentStyles.accountItem} />}
+            renderItem={({ item }) => (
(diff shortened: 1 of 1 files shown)


ray-kraesig:
Codewise, this looks good. Unfortunately, on a smaller device – in particular, one of my usual testing setups, an emulated Nexus 4 (4.7" 768x1280) running Android 5.0 (API 21) – it renders like this:

Screenshot_1577215523-half

The two `FlatList`s end up being separately scrollable, which is less than ideal. It would probably be better for only the bottom list to scroll, or for the entire card to scroll as a single unit.

22sujata8d:
@ray-kraesig thank you for reviewing. For the upper list can I do something like the following:
```
directAdmins.map(admin => (
    <View>
        <ZulipButton ...>
        <View style={componentStyles.accountItem} />
    </View>
)
```
I will try this approach and see the result.

22sujata8d:
I tried the following code:

```
          {directAdmins.map(admin => (
            <View>
              <ZulipButton
                key={admin.email}
                text={admin.email}
                onPress={() => this.tryDevLogin(admin.email)}
              />
              <View style={componentStyles.accountItem} />
            </View>
          ))}
```

Following is the result for Nexus 4 (4.7" 768x1280)

image

ray-kraesig:
That's visually reasonable, although I'm leery of the unbalanced trailing `<View style={componentStyles.accountItem} />`. Does that serve a purpose?

Codewise, it's also usually better to use `<>...</>` instead of `<View>...</View>` as a dummy element.

(Also, please make proposed code changes like this available as a Git commit – whether as part of the PR or just to a different branch on your own repository – so that they can be easily pulled and tested.)

PR #3739, opened 2019-12-26
Fix decoding of internal narrow-links

ray-kraesig:
Links to narrows containing non-ASCII characters were not being fully decoded.

This should fix an error reported in Sentry: https://sentry.io/organizations/zulip/issues/1200774371/

### DIFF
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -61,9 +61,11 @@ export const getLinkType = (url: string, realm: string): LinkType => {
   return 'home';
 };
 
-/** PRIVATE -- exported only for tests. */
-export const transformToEncodedURI = (string: string): string =>
-  string.replace(/\.\d/g, (match, p1, p2, offset, str) => `%${match[1]}`);
+/** Decode a dot-encoded string. */
+// The Zulip webapp uses this encoding in narrow-links:
+// https://github.com/zulip/zulip/blob/1577662a6/static/js/hash_util.js#L18-L25
+export const decodeHashComponent = (string: string): string =>
+  decodeURIComponent(string.replace(/\./g, '%'));
 
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
@@ -78,11 +80,11 @@ const parseStreamOperand = (operand, streamsById): string => {
 
   // Old format: just stream name.  This case is relevant indefinitely,
   // so that links in old conversations continue to work.
-  return decodeURIComponent(transformToEncodedURI(operand));
+  return decodeHashComponent(operand);
 };
 
 /** Parse the operand of a `topic` or `subject` operator. */
-const parseTopicOperand = operand => decodeURIComponent(transformToEncodedURI(operand));
+const parseTopicOperand = operand => decodeHashComponent(operand);
 
 /** Parse the operand of a `pm-with` operator. */
 const parsePmOperand = (operand, usersById) => {
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -1,4 +1,5 @@
 /* @flow strict-local */
+import { addBreadcrumb } from '@sentry/react-native';
 import type { Narrow, Stream, User } from '../types';
 import { topicNarrow, streamNarrow, groupNarrow, specialNarrow } from './narrow';
 import { isUrlOnRealm } from './url';
@@ -64,8 +65,20 @@ export const getLinkType = (url: string, realm: string): LinkType => {
 /** Decode a dot-encoded string. */
 // The Zulip webapp uses this encoding in narrow-links:
 // https://github.com/zulip/zulip/blob/1577662a6/static/js/hash_util.js#L18-L25
-export const decodeHashComponent = (string: string): string =>
-  decodeURIComponent(string.replace(/\./g, '%'));
+export const decodeHashComponent = (string: string): string => {
+  try {
+    return decodeURIComponent(string.replace(/\./g, '%'));
+  } catch (err) {
+    // `decodeURIComponent` throws strikingly uninformative errors
+    addBreadcrumb({
+      level: 'info',
+      type: 'decoding',
+      message: 'decodeHashComponent error',
+      data: { input: string },
+    });
+    throw err;
+  }
+};
 
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -8,7 +8,7 @@ import {
   getLinkType,
   getNarrowFromLink,
   getMessageIdFromLink,
-  transformToEncodedURI,
+  decodeHashComponent,
 } from '../internalLinks';
 import * as eg from '../../__tests__/exampleData';
 
@@ -128,7 +128,7 @@ describe('getLinkType', () => {
     ).toBe('pm');
     expect(
       getLinkType(
-        'https://example.com/#narrow/pm-with/a.40b.2Ecom.c.d.2Ecom/near/3',
+        'https://example.com/#narrow/pm-with/a.40b.2Ecom.2Ec.2Ed.2Ecom/near/3',
         'https://example.com',
       ),
     ).toBe('pm');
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks!

FTR the linked Sentry error has a stack trace that begins like so:
```
URIError: URI error
  at decodeURIComponent(app:///[native code]:0:0)
  at getNarrowFromLink(src/utils/internalLinks.js:85:38)
  at next(src/message/messagesActions.js:44:18)
```

(Sentry items disappear after a pretty short window, I think because of the free plan we're on.)

gnprice:
Haven't yet read this in detail, but one thought I have is: this looks like (a) a pretty fiddly piece of logic, which is (b) self-contained -- naturally a pure, platform-independent function, modulo the `addBreadcrumb` bit -- and (c) a problem the web frontend also needs to solve. So it seems like a strong candidate for sharing with the webapp.

Want to take a look at the corresponding web-frontend code, and see what it would look like to make that code shared? Who knows, it might even have a bug related to this one :stuck_out_tongue: and then we'd get to fix it in both places at once.

Separately, but at a similar level of detail: I'm not sure what a "dot component" is. Want to add a bit of jsdoc saying what `decodeDotComponent` is meant to do? Possibly another name would convey more of the meaning, too. (This may be moot, depending on how the sharing prospects look.)

timabbott:
`static/js/hash_util.js` is the relevant webapp code, which seems like a module that'd make sense to share.

ray-kraesig:
*timabbot wrote:*
> `static/js/hash_util.js` is the relevant webapp code, which seems like a module that'd make sense to share.

Alas, no. The majority of that file is concerned with _constructing_ links to narrows, which isn't a thing the mobile app presently ever does. Even if it did, it's also the case that the majority of that file is tightly coupled to the types and structures of the webapp (`stream_data.id_to_slug(...)`), and factoring that out would probably not be worthwhile.

----

*gnprice wrote:*
> this looks like (a) a pretty fiddly piece of logic,

Yes, but the fiddly parts are all uselessly paranoid error-checking; in the cold light of a new day, it seems clear that it all needs to be set on fire.

> Separately, but at a similar level of detail: I'm not sure what a "dot component" is.

I was kind of hoping you'd have a better idea for a name there. :sweat: Renamed to `decodeHashComponent`, which – while arguably no more informative – is at least [the name used by the webapp](https://github.com/zulip/zulip/blob/61180020c1a32b46868b8d0c5a7686eb43eef6c6/static/js/hash_util.js#L58-L60).

> Want to add a bit of jsdoc saying what `decodeDotComponent` is meant to do?

:+1:

src/utils/internalLinks.js
```diff
@@ -61,9 +62,23 @@ export const getLinkType = (url: string, realm: string): LinkType => {
   return 'home';
 };
 
-/** PRIVATE -- exported only for tests. */
-export const transformToEncodedURI = (string: string): string =>
-  string.replace(/\.\d/g, (match, p1, p2, offset, str) => `%${match[1]}`);
+/** Decode MediaWiki-style dot-encoding. */
```
gnprice: Hmm interesting -- I don't recognize the MediaWiki connection. Is there a useful link?

Some googling finds references to a few kinds of URL encoding in MediaWiki, but they seem to all use percents:
https://en.wikipedia.org/wiki/Help:URL#Fixing_links_with_unsupported_characters
https://www.mediawiki.org/wiki/Manual:PAGENAMEE_encoding

src/utils/internalLinks.js
```diff
@@ -61,9 +62,23 @@ export const getLinkType = (url: string, realm: string): LinkType => {
...
+// The Zulip server and webapp use this encoding for names with special
+// characters in narrow-links.
+export const decodeHashComponent = (string: string): string => {
+  try {
+    return decodeURIComponent(string.replace(/\./g, '%'));
```
gnprice: A pointer to `decodeHashComponent` in `static/js/hash_util.js` would be a helpful one-line comment.

(The same probably goes for some of the existing functions below, oops. But also the value is greater here, because the implementation shares so much verbatim.)

src/utils/internalLinks.js
```diff
@@ -61,9 +62,23 @@ export const getLinkType = (url: string, realm: string): LinkType => {
   return 'home';
 };
 
-/** PRIVATE -- exported only for tests. */
-export const transformToEncodedURI = (string: string): string =>
-  string.replace(/\.\d/g, (match, p1, p2, offset, str) => `%${match[1]}`);
+/** Decode MediaWiki-style dot-encoding. */
+// The Zulip server and webapp use this encoding for names with special
+// characters in narrow-links.
```
gnprice: Let's not say "special character" to describe the overwhelming majority of characters :wink: including ordinary letters used to write most people's languages.

I think simplest is to just say "use this encoding in narrow-links". That's also I think the most accurate description of the situation; it's just that for some names, the encoding is a nop and so it's not obvious it was there.

timabbott:
> Even if it did, it's also the case that the majority of that file is tightly coupled to the types and structures of the webapp (stream_data.id_to_slug(...)), and factoring that out would probably not be worthwhile.

I specifically was referring to `exports.parse_narrow` from `has_util.js`, which is a pure function.

src/utils/internalLinks.js
ray-kraesig: _b)_ [It's gone now.](https://www.mediawiki.org/wiki/Manual:%24wgExperimentalHtmlIds)

_a)_ It wasn't used for all URL encoding – just for encoding the names of [section headers](https://web.archive.org/web/20071203202308/http://en.wikipedia.org/wiki/Joseph-Louis_Lagrange#M.C3.A9canique_analytique) and the like. HTML4 (!) required that all `name` attributes consist of [a restricted set of characters](https://www.w3.org/TR/html4/types.html#type-name) which excluded `%` but included `.`.

> Is there a useful link?

The best I can find is https://stackoverflow.com/questions/36880245. I think omitting the word "MediaWiki" and just linking to `hash_util.js` (as suggested below) would be more useful.

src/utils/internalLinks.js
ray-kraesig: Done! (In my (weak) defense, at the time I was thinking less about देवनागरी or кирилиця, and more about `/[./#%:"']/` and the like.)

ray-kraesig:
> > Even if it did, it's also the case that the majority of that file is tightly coupled to the types and structures of the webapp (stream_data.id_to_slug(...)), and factoring that out would probably not be worthwhile.
> 
> I specifically was referring to `exports.parse_narrow` from `has_util.js`, which is a pure function.

It is pure, but it's still coupled to the webapp's types and capabilities. In particular, it returns values with a `negated` field, which the mobile app knows nothing about.

(It would make quite a bit of sense, though, to core out `getNarrowFromLink` and replace its interior  with a thin transformation layer wrapped around `parse_narrow`.)

src/utils/internalLinks.js
gnprice: Ah, that looks familiar now!

Chasing links from there, this might be the best description:
https://phabricator.wikimedia.org/T152540
(That was the issue for removing it.)

That solution SGTM.

gnprice:
Thanks again! Merged.

> I specifically was referring to `exports.parse_narrow` from `has_util.js`, which is a pure function.

If you follow the call graph in the source, you'll find that a lot of that function's work gets done by `decode_operand`, and that relies on `people.slug_to_emails` and `stream_data.slug_to_name`, which in turn consult the app's data structures. So that will require some refactoring to share: it'll need to take those data structures (perhaps better, getters for them) explicitly.

That's an example of what I had in mind with:

> I think the refactoring needed to share more of the `hash_util` module will be worthwhile to do eventually, but it's tangential to the issue at hand so we might as well pursue other priorities first.

---

Just so it's not lost in the GitHub thread noise, I'll repeat the followup suggestion I made above, though:

> I think actually the main opportunity now is that these newly-written tests are much more thorough and informative than [those in the frontend](https://github.com/zulip/zulip/blob/8f281c4fc9680f8000a7dafeaebe3d3b33982474/frontend_tests/node_tests/hash_util.js#L28-L35). 
> 
> For that reason, and to get further experience sharing code, I think it'd be a good followup to... move `decodeHashComponent` and its inverse `encodeHashComponent`, say, into a shared module.

gnprice:
> Thanks again! Merged.
> 
> > I specifically was referring to `exports.parse_narrow` from `has_util.js`, which is a pure function.
> 
> If you follow the call graph in the source, you'll find that a lot of that function's work gets done by `decode_operand`, and that relies on `people.slug_to_emails` and `stream_data.slug_to_name`, which in turn consult the app's data structures. So that will require some refactoring to share: it'll need to take those data structures (perhaps better, getters for them) explicitly.
> 
> That's an example of what I had in mind with:
> 
> > I think the refactoring needed to share more of the `hash_util` module will be worthwhile to do eventually, but it's tangential to the issue at hand so we might as well pursue other priorities first.
> 
> Just so it's not lost in the GitHub thread noise, I'll repeat the followup suggestion I made above, though:
> 
> > I think actually the main opportunity now [to share code in this area] is that these newly-written tests are much more thorough and informative than [those in the frontend](https://github.com/zulip/zulip/blob/8f281c4fc9680f8000a7dafeaebe3d3b33982474/frontend_tests/node_tests/hash_util.js#L28-L35).
> > For that reason, and to get further experience sharing code, I think it'd be a good followup to... move `decodeHashComponent` and its inverse `encodeHashComponent`, say, into a shared module.

timabbott:
Ahh, makes sense.  Such a refactoring sounds good.  We could name that module `narrow_encoding.js` or something.

gnprice:
Filed #3757 for that pair of sharing opportunities.

PR #3740, opened 2019-12-27
Test data: Eliminate possibility of collisions in randomly-generated user IDs

ray-kraesig:
(Yes, this actually happened.)

Screenshot_20191226_141439

### DIFF
--- a/src/__tests__/exampleData.js
+++ b/src/__tests__/exampleData.js
@@ -18,9 +18,39 @@ describe('nothing', () => {
 /** Return an integer 0 <= N < end, roughly uniformly at random. */
 const randInt = (end: number) => Math.floor(Math.random() * end);
 
+/**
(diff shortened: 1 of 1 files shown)


gnprice:
Cool, thanks! Will merge.

I guess, in retrospect, thinking about these probabilities quantitatively and the number of times we run tests makes clear that there would eventually be a collision here. Looks like the probability of this test failing this way is ~0.03% (any of three user IDs colliding from a universe of size 10k); there's probably several other tests that can independently fail this way with similar probabilities; so if you run the whole suite dozens of times a day, you'll see a failure like this every few weeks to months. In practice we (at least I) usually run `tools/test` and let Jest selectively choose only relevant tests, which makes the failure rate somewhat less, but it's still non-negligible.

I like the nice efficient algorithm, too! And the discussion of pros and cons of different Latin terms to use <3 .

Issue #1234, opened 2017-09-29
Some text & icons are not visible in night theme.


vishwesh3:
Text in DevAuthScreen and icons in SettingOptions are not visible is night theme.

<img width="377" alt="screen shot 2017-09-29 at 5 14 20 pm" src="https://user-images.githubusercontent.com/18511177/31017140-2d25a804-a545-11e7-99c3-1a5308198970.png">
<img width="303" alt="screen shot 2017-09-29 at 5 16 55 pm" src="https://user-images.githubusercontent.com/18511177/31017147-334cf048-a545-11e7-8715-1efd23c9298d.png">

vishwesh3:
@zulipbot label "bite size" "newcomers" "Hacktoberfest"

zulipbot:
**ERROR:** Labels "newcomers", "Hacktoberfest" do not exist and were thus not added to this issue.

borisyankov:
Using our custom `Label` component instead of the generic `Text` will solve this issue.

cdikibo:
I want to take this issue.

kunall17:
@cdikibo Cool, go ahead and comment `@zulip bot claim` without the space b/w zulip & bot to claim and assign it to you!

cdikibo:
@zulipbot claim

zulipbot:
Welcome to Zulip, @cdikibo! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-mobile/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](http://zulip.readthedocs.io/en/latest/index.html#code-contribution-guide), as well as the rest of the developer documentation on that site.


See you on the other side (that is, the pull request side)!

PR #3742, opened 2019-12-30
tools: add `pull-fetch-request`

ray-kraesig:
Add a script to fetch a given PR (e.g. #1234) into a branch of its own (e.g. pr/1234).

Inspired by the Zulip server's script of the same name.

### DIFF
--- a/tools/fetch-pull-request
+++ b/tools/fetch-pull-request
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -eu
+
+usage() {
+cat <<EOF -
+Usage:
+    $(basename "$0") <PR number> [<remote name>]
+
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks!

I'd like to keep the scripts of this flavor consolidated in one place, namely tools/ over in zulip/zulip. Would you send a version of this PR there? I'll be happy to review it.

Also please take a look at the `reset-to-pull-request` script that's in that directory. That's the script I myself use constantly.

ray-kraesig:
[The existing documentation](https://zulip.readthedocs.io/en/latest/git/zulip-tools.html#fetch-a-pull-request-and-rebase) (which I had not been not aware of) suggests that such a PR would need to be significantly more complicated than this.

Closing without action for now, just to clean up.

PR #3748, opened 2020-01-01
ComposeBox: Prevent message with whitespace-only topic from sending.

### LINKED ISSUE #3743: Messages with whitespace-only topics should not be sent
It should not be possible to attempt to send a message with a topic consisting only of whitespace.

----

Note: this bug interacts with #3731. If an attempt is made to post such a message, then – while such messages remain in the outbox – any future messages with *empty* topic will be unified with that topic, instead of being given the topic `(no topic)` as they ordinarily would.

hashirsarwar:
Fixes #3743.
This trims the topic string whenever the message text input is focused. Apart from preventing message with whitespace-only topic from sending, it will also remove whitespace from the start and end of the topic string like Zulip web version.

**GIF:**
ezgif com-optimize

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -397,7 +397,7 @@ class ComposeBox extends PureComponent<Props, State> {
             style={this.styles.composeSendButton}
             Icon={editMessage === null ? IconSend : IconDone}
             size={32}
-            disabled={message.trim().length === 0}
+            disabled={message.trim().length === 0 || topic.trim().length === 0}
             onPress={editMessage === null ? this.handleSend : this.handleEdit}
           />
         </View>


hashirsarwar:
@gnprice kindly review this PR.

ray-kraesig:
If you type a message first, enter several spaces into the topic, and immediately press Send, the message will still be sent with an all-whitespace topic.

`handleMessageFocus` is definitely not the right place to do this. `handleTopicBlur` would be better, but is probably not the most user-friendly option on its own – and, depending on details of the event dispatch system, may not execute in time. I'd suggest ensuring that the Send button is simply not active until a valid topic is present, as it already is when there is no valid message.

hashirsarwar:
@ray-kraesig kindly review the changes once again.

ray-kraesig:
Looks reasonable – but please rebase onto the latest `master`, for ease of testing.

hashirsarwar:
@ray-kraesig done!

ray-kraesig:
Merged, with minor edits to the commit message. Thanks!

PR #3750, opened 2020-01-02

jainkuniya:
Fixes: #3614

Now it looks like this:
- When unread notice is present, then it shrinks to a line. 
I am not sure, in ideal situation will it be visible fully or should we make it scroll up (hide behind navbar)
image
image

- When unread notice is not present
image

Behaviour is same on both Android and iOS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,10 @@ env:
     - COVERALLS_REPO_TOKEN=4eYQDtWoBJlDz2QkxoQ2UcnmJFcOB7zkv
   jobs:
     - TEST_SUITE=''
-    - TEST_SUITE=pirlo
+
+    # Pirlo suite disabled until we adjust the test script to match
+    # changes in order to accomodate #3750.
+    # - TEST_SUITE=pirlo
 
 android:
   components:
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -282,6 +282,9 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   styles = {
+    wrapper: {
+      flexShrink: 1,
+    },
     autocompleteWrapper: {
       position: 'absolute',
       bottom: 0,
@@ -290,6 +293,7 @@ class ComposeBox extends PureComponent<Props, State> {
     composeBox: {
       flexDirection: 'row',
       alignItems: 'flex-end',
+      flexShrink: 1,
     },
     composeText: {
       flex: 1,
@@ -309,6 +313,7 @@ class ComposeBox extends PureComponent<Props, State> {
       borderWidth: 0,
       borderRadius: 5,
       fontSize: 15,
+      flexShrink: 1,
       ...this.inputMarginPadding,
       ...this.context.styles.backgroundColor,
     },
@@ -340,7 +345,7 @@ class ComposeBox extends PureComponent<Props, State> {
     };
 
     return (
-      <View>
+      <View style={this.styles.wrapper}>
         <View style={[this.styles.autocompleteWrapper, { marginBottom: height }]}>
           <TopicAutocomplete
             isFocused={isTopicFocused}
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { Platform, View, TextInput, findNodeHandle, ScrollView } from 'react-native';
+import { Platform, View, TextInput, findNodeHandle } from 'react-native';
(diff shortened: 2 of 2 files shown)


jainkuniya:
> When unread notice is present, then it shrinks to a line.

This issue is due to `AnimatedScaleComponent`, if we remove it from `UnreadNotice`, then unread notice completely scrolls up to make space for the compose box.

gnprice:
Thanks @jainkuniya ! Glad we seem to have found a good path. (FTR [here's our chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/compose.20box.20issue.20.23M3614/near/805218) that led us to this.)

That last screenshot looks great. I think dealing properly with `UnreadNotice` can be a followup task.

Another followup task: when the input gets tall, we probably want to keep the message list (the webview) at some min height, so that it's still possible to refer to existing messages while you're writing the new one. I think the one case where we might not want that is when the screen is so crowded (because of a tiny screen?) that that can't fit with even the *minimum* height of the input; in that case, the input is more important than the message list.

It looks like the build failed inside the new end-to-end test suite. Would you try to debug that? Please discuss in chat as you do; this is new to all of us and there'll surely be things to learn. If we can't figure it out, then the PR will need to disable it for now (in .travis.yml) before merge, but I'd definitely like to spend some time trying to debug first -- this is a good opportunity for us to start understanding what it'll take to keep this test suite maintained.

Just one comment in the code: do we need all three of these `flexShrink` properties? I suspect that at most one or two of them are doing anything at all. Try removing them one at a time to see which are actually needed; that will help make the code clearer about what's really going on, which will be helpful for future maintenance.

jainkuniya:
>  but I'd definitely like to spend some time trying to debug first

Sure, will debug this.

> try removing them one at a time to see which are actually needed;

Tried removing one by one. It either don't fix issue or introduce some other layout issue.

> I think dealing properly with UnreadNotice can be a followup task.

Sure, I have done some debugging on our animation components perviously. I don't think the current implementation is good, mainly how we maintain the visibility.

> we probably want to keep the message list (the webview) at some min height

What should be the min/max height of webview/compose box? We can provide it on `%` basis (based on parent). Like this https://github.com/zulip/zulip-mobile/pull/3713/commits/27f07bc8cf9e089822f46fea1a99454000e6fe15

gnprice:
> > try removing them one at a time to see which are actually needed;
> 
> Tried removing one by one. It either don't fix issue or introduce some other layout issue.

Hmm interesting!

I went and played with it a bit myself, to help understand better. I think the point is that these three are the successive layers of wrapper components, one inside another, from the multiline `Input` outward to the component that's a sibling of the (parent of the) `MessageList` component. And I guess each one of them needs to have the appropriate flex properties in order to propagate the constraints between that `MessageList` on one end of a chain of layout constraints, and the multiline `Input` on the other.

There's actually one other wrapper component in between there, which is the `View` that has the `this.styles.composeText` style... but the reason that's not involved is that the flex it's participating in is in the row direction (thanks to `flexDirection: row` at its parent), which I guess means the constraints in the column direction just get propagated through directly.

So that's all good, then. Thanks for investigating!

That leaves the one open question on this (other than followup tasks) as the Pirlo test failure. Have you been able to reproduce it and investigate at all? (It won't reproduce properly on master right now, because it's broken at an earlier step by #3762. But it'd still be good to avoid introducing an additional failure with this change.)

jainkuniya:
>  And I guess each one of them needs to have the appropriate flex properties in order to propagate the constraints between that MessageList on one end of a chain of layout constraints, and the multiline Input on the other.

Yes!

>  as the Pirlo test failure. Have you been able to reproduce it and investigate at all?

One thing is, I am not able to re-run it on CI. I have to push code in order to re-trigger it. Which increase iteration circle. Thus not sure it is just flaky or something wrong.

I tried few times, but can't find much. 

Also just saw that we have disabled it on master.

jainkuniya:
> https://github.com/zulip/zulip-mobile/pull/3774

Oh, actually Pirlo tests are re-enabled.

gnprice:
Merged. Thanks again @jainkuniya !

We looked at the Pirlo failure today while on a video call. What we learned was that it failed at a step with this description:
> Executing Step 11 of type CLICK
>
> Finding Locator XPATH = //android.widget.FrameLayout[@resource-id="android:id/content"]/android.widget.FrameLayout/android.view.ViewGroup/android.view.ViewGroup/android.view.ViewGroup/android.view.ViewGroup/android.view.ViewGroup[3]/android.widget.ScrollView/android.view.ViewGroup/android.widget.EditText

which looks like it's intended to tap the message input box. Note the `android.widget.ScrollView` in there. Presumably that corresponds to the `ScrollView` component which this branch replaced with a plain `View` -- and so this path stopped describing anything in the layout.

PR #3751, opened 2020-01-02
types [nfc]: remove optionality from Narrow::operator

ray-kraesig:
Narrow's `operator` field was originally optional to encompass a hack in the `TitleSpecial` and (former) `TitleHome` components. As that's long gone (see commit 806a131), we can tighten the type a bit.


gnprice:
LGTM, merged -- thanks!

Issue #3752, opened 2020-01-02
Starred and mentioned narrows can get out of order


gnprice:
This isn't common, and on a busy long-lived server it's generally fairly difficult to trigger. I haven't observed it directly, but it seems clear in the code.

---

@showell pointed out that `Array#sort` when not passed an explicit comparison function [has surprising behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort):
> The default sort order is ascending, built upon converting the elements into strings, then comparing their sequences of UTF-16 code units values.

In particular:
```
> [11,2,22,1].sort()
[ 1, 11, 2, 22 ]
```

In general, this means `sort` with no argument is acceptable only when the array elements have type `string` (or a subtype), and moreover only when the strings are being sorted only for uniqueness and not presentation. (For user-facing strings, we'd want [String#localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) or similar.)

The proper invocation for numbers is instead:
```
> [11,2,22,1].sort((a, b) => a - b)
[ 1, 2, 11, 22 ]
```

---

From a quick search of our code:
```console
$ rg -F 'sort()'
```

we have a handful of no-argument `sort()` calls. Some are emails (ok), some are emoji names (dubious, but sorting entirely by name isn't awesome in the first place), one is locale identifiers (ok)... and one is message IDs, for the `narrows` state subtree:
```
const updateFlagNarrow = (state, narrowStr, operation, messageIds): NarrowsState => {
  if (!state[narrowStr]) {
    return state;
  }
  switch (operation) {
    case 'add':
      return {
        ...state,
        [narrowStr]: [...state[narrowStr], ...messageIds].sort(),
```

This code is invoked for the "starred" and "mentioned" narrows, when a relevant flag is added to a message after we already have state for the narrow. When it is, it'll put the messages for the whole narrow in the "1, 11, 2" order rather than "1, 2, 11".

The reason this is often hard to trigger even if you try is that if all the message IDs involved have the same number of digits -- which is true for long periods once there are lots of total messages -- both orders are equivalent.

Easy to fix, fortunately.

PR #3753, opened 2020-01-02
narrows reducer: Fix bad `sort` in an edge case.

gnprice:
It turns out that `sort` without an explicit comparison function
has surprising behavior -- it converts everything to strings, and
compares the UTF-16 sequences.  So e.g.

  > [11, 2, 22, 1].sort()
  [ 1, 11, 2, 22 ]

This makes it acceptable really only for sorting strings, and even
then only for identifier-like strings.

For numbers, the right pattern is:

  > [11, 2, 22, 1].sort((a, b) => a - b)
  [ 1, 2, 11, 22 ]

A quick grep for nullary `sort` finds we've mostly handled this
correctly; but a quick grep finds one exception, where we're using
nullary `sort` on numbers.  It's the message IDs in a narrow, so
in fact we really are relying on keeping them in numeric order.
Fix the sort.

Happily this isn't a very common codepath, so it isn't likely many
users have hit this bug.

Fixes: #3752

### DIFF
--- a/src/chat/narrowsReducer.js
+++ b/src/chat/narrowsReducer.js
@@ -68,7 +68,7 @@ const updateFlagNarrow = (state, narrowStr, operation, messageIds): NarrowsState
     case 'add':
       return {
         ...state,
-        [narrowStr]: [...state[narrowStr], ...messageIds].sort(),
+        [narrowStr]: [...state[narrowStr], ...messageIds].sort((a, b) => a - b),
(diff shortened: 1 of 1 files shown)


ray-kraesig:
Tested and confirmed against a set of starred messages on CZO. :+1:

gnprice:
Thanks for the review and test!

PR #3755, opened 2020-01-03
Add Types (#2052): for url-parse and in js.js

awasthishubh:
I spent a lot of time at understanding the code and types but I was not able to discover any major new types as I feel most of them are already covered and many are just uncovered due to use of objects like window, document which does not have any typing.

This is my first PR at Zulip. I am very new to this codebase as well as TypeFlow due to which I have a few doubts which may be silly but I want someone to help me understand.

* In `src/common/icon.js`, we haven't mentioned types of `props`. Can anyone help me in extracting the possible properties of this object `props`? I traced its references and found some properties like `size`, `colour`. 
I also found `style` property in `props` which makes it really hard or even impossible to determine the type. Please correct me if I'm wrong.
Also, I found an `IconProps` type defined in `icon.js`. It looks the same but it's not been used so I assume this is not the right type of props.

* In `src/api/modelTypes.js` `$FlowFixMe` is used for `display_recipient` property of Message. This left around 50 uncovered types in src/webview/html/messageHeaderAsHtml.js and many other places (like narrow.js)
Comment at `modelTypes.js` says its type is either 'string' or PmRecipientUser[] so why can't we use type as `string|PmRecipientUser[]`?

### DIFF
--- a/src/start/webAuth.js
+++ b/src/start/webAuth.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import { NativeModules, Platform } from 'react-native';
 import SafariView from 'react-native-safari-view';
-import parseURL from 'url-parse';
+import parseURL, { type Url } from 'url-parse';
 
 import type { Auth } from '../types';
 import openLink from '../utils/openLink';
@@ -69,7 +69,7 @@ export const authFromCallbackUrl = (
   otp: string,
   realm: string,
 ): Auth | null => {
-  const url = parseURL(callbackUrl, true);
+  const url: Url = parseURL(callbackUrl, true);
 
   // callback format expected: zulip://login?realm={}&email={}&otp_encrypted_api_key={}
   if (
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -71,7 +71,7 @@ if (!Element.prototype.closest) {
    Taken (with minor edits) from the relevant MDN page. */
 if (!String.prototype.startsWith) {
   // $FlowFixMe (polyfill)
-  String.prototype.startsWith = function startsWith(search, rawPos) {
+  String.prototype.startsWith = function startsWith(search: string, rawPos: number) {
     const pos = rawPos > 0 ? rawPos | 0 : 0;
     return this.substring(pos, pos + search.length) === search;
   };
@@ -82,7 +82,7 @@ if (!String.prototype.startsWith) {
      https://tc39.es/ecma262/#sec-string.prototype.includes */
 if (!String.prototype.includes) {
   // $FlowFixMe (polyfill)
-  String.prototype.includes = function includes(search, start = 0) {
+  String.prototype.includes = function includes(search: string, start: number = 0) {
     /* required by the spec, but not worth the trouble */
     // if (search instanceof RegExp) { throw new TypeError('...'); }
     return this.indexOf(search, start) !== -1;
@@ -108,7 +108,7 @@ const sendMessage = (msg: MessageListEvent) => {
   window.ReactNativeWebView.postMessage(JSON.stringify(msg));
 };
 
-window.onerror = (message, source, line, column, error) => {
+window.onerror = (message: string, source: string, line: number, column: number, error: Error) => {
   if (window.enableWebViewErrorDisplay) {
     const elementJsError = document.getElementById('js-error-detailed');
     if (elementJsError) {


src/types.js
```diff
@@ -263,3 +263,27 @@ export type PmConversationData = {|
   recipients: string,
   unread: number,
 |};
+
+
+/**
+ * URL return type of parse-url
+ */
+export type UrlType={
```
ray-kraesig: The right place for this is not in `src/` but in `flow-types/npm/url-parse_v1.3.x.js` – where it can already be found.

ray-kraesig: Most of this can already be found in `flow-types/npm/url-parse_v1.3.x.js`. If that definition file is notably inaccurate, the a new one should be created in `flow-types/`, not in `src`.

src/types.js
```diff
@@ -263,3 +263,27 @@ export type PmConversationData = {|
...
+  slashes: boolean,
+  protocol: string,
+  hash: string,
+  query: {
+    realm: string,
+    email: string,
+    otp_encrypted_api_key: string
+  },
```
ray-kraesig: This should probably be `query: { [key: string]: string }`.

(The existing definition file has `query: Object`, which isn't quite right.)

src/types.js
awasthishubh: I checked that file but it uses the object as a type for query properties which will still leave types uncovered.
Is flow-types/npm/url-parse_v1.3.x.js auto-generated? Can I modify type in this file or should I create a new one?
From 4401df005c744251972f062dc1631eee02b492c7 I guess It's not auto-generated.

src/types.js
awasthishubh: I created the type for our particular use case. I will modify it to query: { [key: string]: string }`

awasthishubh:
> The bits in `js.js` look good. I'll have to dig a bit to answer your question about `icon.js`, but I do know about `.display_recipient` offhand:
> 
> > * In `src/api/modelTypes.js` `$FlowFixMe` is used for `display_recipient` property of Message. [...]
> >   Comment at `modelTypes.js` says its type is either 'string' or PmRecipientUser[] so why can't we use type as `string|PmRecipientUser[]`?
> 
> The right fix here appears to be factoring `.display_recipient` out into a fragment of `Message`/`Outbox`, and conditioning its type on the value of `.type` (as we do for action types and so forth): `'stream'` implies `display_recipient: string`, `'private'` implies `display_recipient: PmRecipientUser[]`. However, there are bugs in our version of Flow that prevent us from factoring it out that way.
> 
> We could alternatively use the explicit union type you suggest – but that would involve writing a bunch of code to test its type at every use site just to satisfy this version of Flow, and then taking it all back out once we _can_ do that refactor.
> 
> Hopefully that's not worth the trouble, especially since a Flow upgrade should be feasible soon.

Got it. Thanks :)

ray-kraesig:
Hmm? There's no new code for me to review.

... although to follow up about `src/common/icon.js` – that code is broken partly due to the `allowFontScaling` issue discussed in the comment inline. More recent versions of Flow than the one we're currently using complain about it.

awasthishubh:
@ray-kraesig Sorry to ping you like this. I guess you missed my reply.
Regarding the type defined at flow-types/npm/url-parse_v1.3.x.js for the object returned by url-parser:

> I checked that file but it uses the object as a type for query properties which will still leave types uncovered.
> Is flow-types/npm/url-parse_v1.3.x.js auto-generated? Can I modify type in this file or should I create a new file at flow-typed?
> From [4401df0](https://github.com/zulip/zulip-mobile/commit/4401df005c744251972f062dc1631eee02b492c7) I guess It's not auto-generated.

ray-kraesig:
> @ray-kraesig Sorry to ping you like this. I guess you missed my reply.

No apologies necessary: something seems to have eaten *my* reply from earlier! A reconstruction follows.

`flow-types/npm/url-parse_v1.3.x.js` is not autogenerated, but it is third-party; it belongs to [the flow-typed repository](https://github.com/flow-typed/flow-typed). Their README indicates that they accept improvement PRs, so one route forward (probably the best route forward) would be to fix this on their end, and then update us to the latest version.

Alternatively, if the file `flow-types/url-parse_v1.3.x.js` is created (note the lack of an `npm/`!), Flow should use it in preference to the one in `flow-types/npm/`. This would be somewhat more awkward to maintain, however.

awasthishubh:
Instead of removing I imported url type from flow-typed. I will also send a PR to flow-type to update their definition.

Issue #3760, opened 2020-01-07
Button UI in CreateStream is not working properly 


sumukhah:
Users get confused whether the button is active or not,
So fixing this issue will solve the problem.
_the actual value is getting updated, but only the UI is not supporting_
![ButtonFixZulip](https://user-images.githubusercontent.com/23723464/71878269-3afd1d00-3151-11ea-8b95-89ee88d2d915.gif)

PR #3761, opened 2020-01-07
Buttons on Create Stream: Fix Button UI not turns On clicking over,

sumukhah:
button was referencing initialValue, when user clicks on that button it was not updating it's value
I have fixed it by changing current value of the button
Fixes:#3760

### DIFF
--- a/src/streams/EditStreamCard.js
+++ b/src/streams/EditStreamCard.js
@@ -79,7 +79,7 @@ export default class EditStreamCard extends PureComponent<Props, State> {
           style={componentStyles.optionRow}
           Icon={IconPrivate}
           label="Private"
-          value={initialValues.invite_only}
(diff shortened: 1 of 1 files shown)


ray-kraesig:
The code fix looks good, but there are several issues with the commit message. See https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages for more details.

In addition, although not explicitly called out in that document:
* The second line of the commit message should always be blank, separating the one-line summary from the message proper.
* While GitHub will recognize [many things](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) as issue-closing remarks, it does seem to require a space between the keyword and the issue number.

ray-kraesig:
(Adjusted the commit message formatting before merging.) Thanks!

PR #3762, opened 2020-01-07
Cut "welcome screen"; make enter-URL screen more helpful

gnprice:
This cuts out the rather bare WelcomeHelpScreen, as well as WelcomeScreen which existed as a way to get to it; then makes some changes to RealmScreen to hopefully make it more directly helpful instead.


gnprice:
Travis failed on the Pirlo suite: the Pirlo test script goes through the onboarding flow (an important part of its job!); this PR changes the onboarding flow; and so the script doesn't find what it expects and isn't able to proceed. Here's the detailed result (on one of the two devices):
https://app.pirlo.io/result/8a8e0576-1be7-40ea-adb3-adb524ea9bda/details?authToken=9mf3-dtyE1lMeDFFoRjm391oY-0%3D

This will be an important sort of thing to handle to make the Pirlo test suite useful over time, so I'll be interested in figuring it out in general. It seems like what it's really telling us is that the test script ought to be versioned together with the code -- so the commit that changes the flow also changes the script -- but I think that's not really how Pirlo is designed to work, so more discussion is needed. /cc @kgritesh @vinayjain404

For now, because we're in a hurry for this particular change, I'll add to this PR a commit that just temporarily disables the Pirlo suite in CI.

src/start/RealmScreen.js
```diff
@@ -76,7 +77,13 @@ class RealmScreen extends PureComponent<Props, State> {
     const { progress, error, realm } = this.state;
 
     return (
-      <Screen title="Welcome" padding centerContent keyboardShouldPersistTaps="always">
+      <Screen
+        title="Welcome"
+        canGoBack={!this.props.navigation.state.params?.initial}
```
ray-kraesig: This is what `dangerouslyGetParent` is documented to be for.

It's still a hack, mind (as if the name didn't give it away). The "right" way is probably to use a react-navigation `AppContainer`, rather than trying to generalize this somehow.

static/translations/messages_en.json
```diff
@@ -17,6 +17,7 @@
   "Pick account": "Pick account",
   "Welcome": "Welcome",
   "Enter your Zulip server URL:": "Enter your Zulip server URL:",
+  "e.g. zulip.example.com": "e.g. zulip.example.com",
```
ray-kraesig: This helper text isn't actually helpful – it's redundant with the existing placeholder-slash-autocomplete text.

ray-kraesig:
Our CI test failed on the previous run – not Pirlo, but our Jest tests; it seems `src/utils/__tests__/async-test.js`'s test for `sleep` is flaky. Created PR #3763 to provide more information for next time (though it's not hard to guess what probably happened).

static/translations/messages_en.json
gnprice: It's a different example, though; my thought is to make this bit of UI less zulipchat.com-centric.

src/start/RealmScreen.js
gnprice: Hmm fascinating! I'll take a look.

gnprice: Hmm, so here's the docs:
https://reactnavigation.org/docs/en/2.x/navigation-prop.html#dangerouslygetparent-get-parent-navigator

It looks like this will get us the `StackNavigator` object we call `AppNavigator`. And from there, we'd take its `routes` property or something like that?

I think that's the same purpose we currently serve with the `nav` reducer and selectors -- note `getNavigationRoutes` in particular.

One reason we moved away from that is that it didn't have the right behavior on transitions. You'd be on one screen, and you'd hit a button that navigated to another... and for a few hundred ms, the old screen is still visible (at least in part), but the *overall* nav state has already changed, so if the screen uses the overall nav state to decide how to render then we get glitches. The `props.navigation.state.params` value, on the other hand, remains at the value that belonged to *this* screen.

vinayjain404:
@gnprice Currently our platform does not have branching or versioning built in it. Our clients update the test case independently after changes to the UI and ignore the test failure.

I am happy to update the test case once the new flow is updated and we can think more about designing versioning into the system.

src/start/RealmScreen.js
gnprice: On `AppContainer`: reading up on that, it looks like it's part of what we already have (it's implicit in `createStackNavigator` up through 2.x, and that's what we use in `AppNavigator.js`); but the part of it that handles showing the app bar and back-navigation button, we turn off by passing `headerMode: 'none'`.

Seems quite likely that an eventual good answer here would look like providing our own implementation of the app bar, integrated with the navigation framework, rather than disabling it. Docs for that appear to be
https://reactnavigation.org/docs/en/2.x/navigation-views.html
https://reactnavigation.org/docs/en/navigation-views.html
and in particular the latter links to
https://github.com/react-navigation/stack/blob/1.0/src/views/Header/Header.tsx

I'll augment the comment to mention that.

static/translations/messages_en.json
gnprice: From a bit of in-person discussion: AIUI you think this example will actually be confusing, but I'm not persuaded.

Given that we want to get this set of changes out quickly, I'm going to go ahead and merge, and then see about a release. This is certainly a screen where it makes sense to iterate further on the UI, though; we can always change it again next.

PR #3763, opened 2020-01-07
tests: push some complex expectations into Jest helpers

ray-kraesig:
Use jest-community helpers to compute certain properties, to provide
more precise information in case of failure.

(Inspired by a test flake for `sleep` in `async-tests`.)

### DIFF
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -7,7 +7,8 @@ describe('sleep', () => {
     const start = Date.now();
     await sleep(expectedMs);
     const durationMs = Date.now() - start;
-    expect(expectedMs <= durationMs && durationMs < 10 * expectedMs).toBeTruthy();
+    expect(expectedMs).toBeLessThanOrEqual(durationMs);
+    expect(durationMs).toBeLessThan(10 * expectedMs);
   });
 });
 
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/async-test.js
```diff
@@ -7,7 +7,8 @@ describe('sleep', () => {
     const start = Date.now();
     await sleep(expectedMs);
     const durationMs = Date.now() - start;
-    expect(expectedMs <= durationMs && durationMs < 10 * expectedMs).toBeTruthy();
+    expect(expectedMs).toBeLessThanOrEqual(durationMs);
+    expect(durationMs).toBeLessThanOrEqual(10 * expectedMs);
```
gnprice: There's a small semantic change here: `<` becomes "less than or equal".

ray-kraesig: ... I mean, I don't think being a millisecond more permissive is really meaningful here; but okay?

gnprice:
Merged -- thanks!

src/utils/__tests__/async-test.js
gnprice: Thanks for the revision!

This is one of those cases where I don't have much of a preference between the possibilities if writing from scratch... but in a refactoring commit that's about making one kind of change, I definitely prefer that it makes only that kind of change, and doesn't have any other semantic effects, even very small ones. It's a form of hygiene that I find very helpful for reading and understanding the diffs.

In this instance, with the revision I believe it's the case that this only affects what details are printed, and has no effect on whether the checks pass or fail.

If you find the double-closed interval nicer here, I'd be perfectly fine with a commit switching to that. Just as a separate commit from any wider, orthogonal refactoring. :slightly_smiling_face:

PR #3769, opened 2020-01-08
Minor tweaks

ray-kraesig:
Fix up a couple of minor inefficiencies discovered while poking at an unrelated bug.


gnprice:
Merged -- thanks!

PR #3773, opened 2020-01-09
[nfc] Revert indentation changes

ray-kraesig:
Revert some (apparently) accidental indentation changes to support files, to somewhat simplify the hopefully-upcoming upgrade to RN 0.60.

(This PR contains only whitespace changes.)


gnprice:
The invariant I want to maintain for formatting like indentation is that we can always have the IDE auto-format the code (with one keystroke), and it doesn't make any changes. That minimizes formatting churn, and I also often find it convenient while writing code.

For our main codebase, we check this invariant in CI. For the Android codebase we don't, and so it's incompletely adhered to, but it's generally close. For this file, it looks like if I do that at the tip of this branch, it
* reverts this branch's changes;
* expands `getConversations` out from its single line.

That story on Android isn't as well documented or fully automated as in the main codebase, mainly because over the past year or so the Android work has mostly been just me. So it's probably getting to time to fix that. :slightly_smiling_face: There's some discussion in the commit messages of
83134808803920467488fa333b70bdd970d87bab
8c86b820b564f2a0d3d5556fc6b0ec64959ac66c
436800ecdaeab9000005892292995ac17d6849a7

So I think for the Android file here, we should just leave it as is.

---

For the `Info.plist` file:

    Inadvertently changed in c8818f0a33224f6226667cff4d73e3a5f3832875.
    (Understandably so -- this file, uniquely, uses tabs as indentation.)

It looks like that commit actually made the *same* indentation change you're making.

Then 456352b82 changed it back.

Before that, e5cbe3cc5 also changed it.

I don't really care which it is, but I would like it to stop changing. :-) What happens if you go into Xcode and edit one of the values in this file?

I don't believe the name "React Native standard"; I think what you're describing is just "the sample app in the `react-native init` template". RN doesn't care what the indentation is.

ray-kraesig:
(Stale, and Chris has the RN 0.60 upgrade anyhow.)

Issue #3776, opened 2020-01-09
Build doesn't work from the Windows Command Prompt


gnprice:
Running a command like `react-native run-android` fails with an error like
```
> Task :app:buildDebugStaticWebviewAssets FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:buildDebugStaticWebviewAssets'.
> A problem occurred starting process 'command 'bash''
```

The problem seems to be that our bit of Gradle code that starts the Bash script `tools/build-webview` can't find the `bash` executable to run.

For now, the workaround is to run from the Git Bash prompt rather than the Command Prompt.

But we should also keep the build working smoothly from the Command Prompt. It should be possible to adjust that Gradle code so that it can find the Git Bash version of `bash` even when run from the Command Prompt.

agrawal-d:
Running the tests from `./tools/test` in a non-bash command-line ( like cmd ) also fails on Windows.

gnprice:
> Running the tests from `./tools/test` in a non-bash command-line ( like cmd ) also fails on Windows.

The tests will include the build, so it's expected that if the build fails, the tests will fail.

(If you haven't touched anything in `android/`, then by default a plain `tools/test` skips the Gradle build, so it might pass. But `tools/test --full` will run the build and will fail.)

What does the failure look like when you run `./tools/test`? If there's something else failing in addition to the Gradle build, let's make an issue for that too.

agrawal-d:
I do not use Windows; a new developer had asked me for help with running the tests, and I suggested them to use `git bash` instead of `cmd`, and the tests worked.

The error message they got was:
`Program 'test' failed to run:, No application is associated with the specified file for the operation`

gnprice:
Hmm. OK, I think the best way to proceed on that would be for the person who's seeing the issue to say hello in `#mobile` on chat.zulip.org, and we can debug with them there.

That'll help us effectively figure out specifically what's happening. Then we can open new issues here as needed for anything else that turns out to need fixing.

gnprice:
Thinking about this again, I decided the practical solution is for us to just desupport running commands from the Windows Command Prompt -- we can recommend people use Git Bash instead, which our setup instructions already have people install. Pushed 3aadff09f making that change in our docs. Closing this bug as WONTFIX.

PR #3779, opened 2020-01-09
Windows build issues

ray-kraesig:
Address three mostly unrelated Windows issues:

* Fix up the webview build script.
* Document a new explicit requirement to install `rsync`.
* Warn users to prefer Git Bash, rather than the Command Prompt, for now (#3776).


gnprice:
Thanks @ray-kraesig ! Merged.

Both the macOS-specific and Windows instructions feel like they've overgrown the format a bit at this point; I'll make a pass now at reorganizing them some.

gnprice:
> Both the macOS-specific and Windows instructions feel like they've overgrown the format a bit at this point; I'll make a pass now at reorganizing them some.

Pushed as b46c429bb.

PR #3783, opened 2020-01-10
webview build: unbreak recent macOS breakage

ray-kraesig:
The recent commit ca71acfbfc9843bce156bb161bd30f7ef076242a broke
webview builds on macOS by assuming that `readlink -m` was already
available.

Unfortunately, there's a Catch-22 here: we need $root to ensure we
have access to `readlink -m` in the first place! On the other hand, we
don't need a normalized form of $root, just something that will let us
run a script.

Delay normalization of $root until after `ensure-coreutils.sh` is run.

### DIFF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -35,19 +35,20 @@ err() {
 # Environment
 ################################################################################
 
-# The project root, for absolutizing paths. (Assumed to be one directory up from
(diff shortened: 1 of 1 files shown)


gnprice:
Merged -- thanks!

Issue #3700, opened 2019-11-23
Group chat screen is broken when there are many members in the chat.


jainkuniya:
![Screenshot_20191124-032042](https://user-images.githubusercontent.com/18511177/69485988-9ce91800-0e6c-11ea-87ce-6a7992f67b7d.jpg)

awasthishubh:
@zulipbot claim

Maskedman99:
![Screenshot-20200417025912-339x222](https://user-images.githubusercontent.com/31368194/79508582-70ba7780-8057-11ea-87f3-c7c03ec5ceb1.png)

It's also broken when the recipient's name is large.

PR #3786, opened 2020-01-10
Group Nav: limit avatars & hide overflow

awasthishubh:
Fixes issue: #3700

**Before**
020ef3fa-5a60-47ed-a558-91507f142451
---
**After**
7e721bb2-0bb6-4b62-ab45-f32916c5e20d

`navWrapper` style is being used multiple places but I guess it's safe to set `overflow: hidden` as overflow anyways goes beyond the screen in other cases.

This fixes the issue but I also limited the number of avatars to be rendered as in a large group without this limit, 100s of avatar images will stay in memory without being displayed.
* I used screen height as well so as to have enough avatars for landscape mode. 
* I didn't consider navigation and info icon's width so as to have 1 or 2 extra avatars just to be on a safer side.

### DIFF
--- a/src/title/TitleGroup.js
+++ b/src/title/TitleGroup.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 
 import React, { PureComponent } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { StyleSheet, View, Dimensions } from 'react-native';
 
 import type { Dispatch, UserOrBot } from '../types';
 import { connectFlowFixMe } from '../react-redux';
@@ -29,10 +29,12 @@ class TitleGroup extends PureComponent<Props> {
 
   render() {
     const { recipients } = this.props;
+    const { width, height } = Dimensions.get('window');
+    const numOfAvatar = Math.ceil(Math.max(width, height) / (32 + 16)); // ImgSize: 32, margin: 16
 
     return (
       <View style={styles.navWrapper}>
-        {recipients.map((user, index) => (
+        {recipients.slice(0, numOfAvatar).map((user, index) => (
           <View key={user.email} style={this.styles.titleAvatar}>
             <UserAvatarWithPresence
               onPress={() => this.handlePress(user)}
--- a/src/styles/navStyles.js
+++ b/src/styles/navStyles.js
@@ -8,6 +8,7 @@ export const statics = {
     flexDirection: 'row',
     alignItems: 'center',
     justifyContent: 'flex-start',
+    overflow: 'hidden',
   },
   navSubtitle: {
     fontSize: 13,


src/title/TitleGroup.js
```diff
@@ -29,10 +29,12 @@ class TitleGroup extends PureComponent<Props> {
 
   render() {
     const { recipients } = this.props;
+    const { width, height } = Dimensions.get('window');
+    const numOfAvatar = Math.ceil(Math.max(width, height) / (32 + 16)); // ImgSize: 32, margin: 16
```
jainkuniya: I would suggest you to not not sure `Dimensions.get('window');`, as it will give Dimensions of window, but if app is running in split screen, it might break.

Also this solution doesn't seems to be future proof. As in future if you increase margin or size, you have to reconfigure this numbers.

awasthishubh:
@jainkuniya Thanks for the review :)

>Set justifyContent to space-between in navBar (basically create wrappers around title and buttons and logically relate them with space-between or space-around) and set flex:1 to the avatar list.

I tried. It modified the positioning of Title and buttons but this issue is being caused due to overflow in Title which still persisted.
Just to be sure, is this what you asked?

WhatsApp Image 2020-01-13 at 6 10 54 PM

awasthishubh:
> navWrapper is also used in TitleSpecial, so try creating new style.

as I have mentioned
> navWrapper style is being used multiple places but I guess it's safe to set overflow: hidden as overflow anyways goes beyond the screen in other cases.

Do you think setting `overflow: hidden` can break `TitleSpecial` in some case? 
If so can I set the style as 
`<View style={...styles.navWrapper, overflow: 'hidden'}>`
instead of creating a whole new style?

src/title/TitleGroup.js
awasthishubh: I can assure you that this will not break on split-screen. :)
`Dimensions.get('window')` will either be equal or larger than the actual width we want. In the latter case, this will render more avatars which is fine.

> Also this solution doesn't seems to be future proof. As in future if you increase margin or size, you have to reconfigure this numbers.

If I put a comment at those values, will that be fine? I can't think of any other workaround.

chrisbobbe:
Greg mentions at https://github.com/zulip/zulip-mobile/issues/3940#issuecomment-593583112 that it would be nice to show an ellipsis — it doesn't have to be pressable — just to indicate that there are more avatars, and the list hasn't ended.

src/title/TitleGroup.js
gnprice: > will either be equal or larger than the actual width we want. In the latter case, this will render more avatars which is fine.

If it doesn't actually affect what gets displayed, then the simpler thing is just to not do the slice at all (i.e. leave this effectively at infinity.) That will also avoid the risk that it could break due to changes in the assumptions (like the size and margin of these images) that this calculation is relying on.

awasthishubh:
Sorry, I'm currently not available to make changes. I would like someone else to take this issue.

Issue #3768, opened 2020-01-08
Flow loses track of prop types when we say `export default connect(...`


gnprice:
This appears to be a bug in Flow itself. I can't find a report of it in the Flow tracker -- for example, this search comes up empty:
https://github.com/facebook/flow/issues?utf8=%E2%9C%93&q=is%3Aissue+export+empty
so probably I should turn it into a report there. (Though we shouldn't get our hopes up for that leading to a fix in bounded time.)

The symptom is that Flow winds up identifying many of our React components as having type `ComponentType<empty>`, which you can see by hovering over them in the editor wherever they're imported and used. The Flow type called `empty` is the bottom of the type poset, the subtype of all types; in our experience, it's often produced by bugs in Flow, and has the effect of papering over them and hiding related type errors.

Here, if you go and edit a call to such a component to pass some completely bogus props:
```
-        <AccountDetails user={user} />
+        <AccountDetails user={32} foo="bar" />
```
you find that Flow gives no error. So that's bad; there may be some real type errors this is hiding.

In at least a few spot-checked cases, the common pattern in the affected components is that we use `react-redux` on them, saying something like
```
export default connect((state, props): SelectorProps => ({
  realm: getCurrentRealm(state),
  userStatusText: getUserStatusTextForUser(state, props.user.user_id),
}))(AccountDetails);
```

It turns out that to fix the issue, it's enough to edit this like so:
```
const X = connect(/* ... */); // the exact same expression as before

// $FlowFixMe
const y = () => <X />;

export default X;
```
Now when the component is imported in some other module, it has the type it should have, like `ComponentType<{| +user: User |}>`; and a call with bogus props gets the error messages it should get.

It's not enough to separate the `const X = ...` and `export default`; if you comment out the `const y` line, it's back to the status quo with `ComponentType<empty>`.

The mention of `X` in `const y = ...` *should* have no effect on the type of `X`. After all, we don't even mention any props and the call doesn't type-check (hence the `$FlowFixMe`), so this doesn't give Flow any information it didn't already have. That's why I'm confident this is accurately described as a bug in Flow, and not in e.g. the type of `connect`. (A priori, it sure could have been a bug in the type of `connect`; that type is pretty complex.)

We should probably deploy some workaround across all our components that use the `export default connect(...` pattern; either the one above, or some variation we find that's equally effective.

ray-kraesig:
[...]

Issue #2585, opened 2018-05-29
TypeError: TypeError: undefined is not an object (evaluating 'e.name.toLowerCase')


borisyankov:
https://sentry.io/zulip/zulip-mobile/issues/562779649/

```
TypeError: TypeError: undefined is not an object (evaluating 'e.name.toLowerCase')

This error is located at:
    in t
    in Connect(t)
    in t
    in RCTView
    in n
    in t
    in n
    in RCTView
    in RCTView
    in t
    in RCTView
    in e
    in r
    in Unknown
    in n
    in n
    in t
    in Connect(t)
    in Unknown
    in t
    in Connect(t)
    in IntlProvider
    in t
    in Connect(t)
    in t
    in Connect(t)
    in RCTView
    in n
    in Connect(n)
    in n
    in t
    in t
    in Unknown
    in RCTView
    in RCTView
    in t
    at src/topics/TopicListScreen.(TopicListScreen.js:49)
    <TopicList topics={filteredTopics} onPress={this.handlePress} />
    at src/topics/TopicListScreen.ReactNativeRenderer$3(TopicListScreen.js:49)
    <TopicList topics={filteredTopics} onPress={this.handlePress} />
...
(17 additional frame(s) were not displayed)
```

gnprice:
We marked this as fixed by 8811fa813, but I don't see how that could have fixed the issue -- it changed purely the type annotations, not the runtime behavior. So, if this has been fixed, it was fixed at another time.

Checking in Sentry, it looks like this bug is still live. [Here's](https://sentry.io/organizations/zulip/discover/zulip-mobile:9c01aef1eafc4d948824db5796184ddb) a recent event from it, on 26.22.145 which is the current version. The message is:
```
TypeError: TypeError: undefined is not an object (evaluating 't.name.toLowerCase')
```
and the stack trace puts it here in `TopicListScreen.js`:
```
      topics && topics.filter(topic => topic.name.toLowerCase().includes(filter.toLowerCase()));
```

Pretty rare, though.

chrisbobbe:
Oh no! My bad. I was aware that I only changed the type annotations, but I diligently went through to make sure the runtime checks were already done, and said so in the commit message. But I didn't indicate which commit those checks were done in, and, to cap it all, apparently, I was wrong. :(

PR #3789, opened 2020-01-10
Connect type fix

chrisbobbe:
These are various type fixes mostly addressing the use of connectFlowFixMe, so that connect is used instead. They were discovered while working on https://github.com/zulip/zulip-mobile/pull/3765 but were unrelated enough to motivate a separate PR.

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -43,16 +43,15 @@ class AccountDetails extends PureComponent<Props> {
   render() {
     const { realm, user, userStatusText } = this.props;
 
-    // The set of timezone names in the tz database is subject to change over
-    // time. Handle unrecognized timezones by quietly discarding them.
-    let localTime: string | null;
-    try {
-      localTime =
-        user.timezone !== undefined && user.timezone !== ''
-          ? `${nowInTimeZone(user.timezone)} Local time`
-          : null;
-    } catch (err) {
-      localTime = null;
+    let localTime: string | null = null;
+    // See comments at CrossRealmBot and User at src/api/modelTypes.js.
+    if (user.timezone !== '' && user.timezone !== undefined) {
+      try {
+        localTime = `${nowInTimeZone(user.timezone)} Local time`;
+      } catch (err) {
+        // The set of timezone names in the tz database is subject to change over
+        // time. Handle unrecognized timezones by quietly discarding them.
+      }
     }
 
     return (
@@ -70,11 +69,11 @@ class AccountDetails extends PureComponent<Props> {
         <View>
           <ActivityText style={styles.largerText} user={user} />
         </View>
-        {user.timezone !== '' && user.timezone !== undefined ? (
+        {localTime !== null && (
           <View>
-            {localTime !== null && <RawLabel style={styles.largerText} text={localTime} />}
+            <RawLabel style={styles.largerText} text={localTime} />
           </View>
-        ) : null}
+        )}
       </ComponentList>
     );
   }
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -2,7 +2,7 @@
 import React, { PureComponent } from 'react';
 import { View, StyleSheet } from 'react-native';
 
-import type { User, Dispatch } from '../types';
+import type { UserOrBot, Dispatch } from '../types';
 import { connect } from '../react-redux';
 import { UserAvatar, ComponentList, RawLabel } from '../common';
 import { getCurrentRealm, getUserStatusTextForUser } from '../selectors';
@@ -33,7 +33,7 @@ type SelectorProps = {|
 |};
 
 type Props = $ReadOnly<{|
-  user: User,
+  user: UserOrBot,
 
   dispatch: Dispatch,
   ...SelectorProps,
@@ -47,7 +47,10 @@ class AccountDetails extends PureComponent<Props> {
     // time. Handle unrecognized timezones by quietly discarding them.
     let localTime: string | null;
     try {
-      localTime = `${nowInTimeZone(user.timezone)} Local time`;
+      localTime =
+        user.timezone !== undefined && user.timezone !== ''
+          ? `${nowInTimeZone(user.timezone)} Local time`
+          : null;
     } catch (err) {
       localTime = null;
     }
@@ -67,7 +70,7 @@ class AccountDetails extends PureComponent<Props> {
         <View>
           <ActivityText style={styles.largerText} user={user} />
         </View>
-        {user.timezone ? (
+        {user.timezone !== '' && user.timezone !== undefined ? (
           <View>
             {localTime !== null && <RawLabel style={styles.largerText} text={localTime} />}
           </View>
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -2,8 +2,9 @@
 import React, { PureComponent } from 'react';
 import { StyleSheet } from 'react-native';
 
-import type { Dispatch, User } from '../types';
-import { connectFlowFixMe } from '../react-redux';
+import type { NavigationScreenProp } from 'react-navigation';
+import type { Dispatch, UserOrBot } from '../types';
+import { connect } from '../react-redux';
 import { getAccountDetailsUserForEmail } from '../selectors';
 import { Screen, ZulipButton, Label } from '../common';
 import { IconPrivateChat } from '../common/Icons';
@@ -25,11 +26,13 @@ const styles = StyleSheet.create({
 });
 
 type SelectorProps = $ReadOnly<{|
-  user: User,
   isActive: boolean,
+  user: UserOrBot,
 |}>;
 
 type Props = $ReadOnly<{|
+  navigation: NavigationScreenProp<{ params: {| email: string |} }>,
+
   dispatch: Dispatch,
   ...SelectorProps,
 |}>;
@@ -67,7 +70,7 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state, props) => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
   isActive: getUserIsActive(state, props.navigation.state.params.email),
 }))(AccountDetailsScreen);
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -2,7 +2,7 @@
 import React, { PureComponent } from 'react';
 import { StyleSheet } from 'react-native';
 
-import type { Dispatch, GlobalState, User } from '../types';
+import type { Dispatch, User } from '../types';
 import { connectFlowFixMe } from '../react-redux';
 import { getAccountDetailsUserForEmail } from '../selectors';
 import { Screen, ZulipButton, Label } from '../common';
@@ -24,10 +24,14 @@ const styles = StyleSheet.create({
   },
 });
 
-type Props = $ReadOnly<{|
+type SelectorProps = $ReadOnly<{|
   user: User,
   isActive: boolean,
+|}>;
+
+type Props = $ReadOnly<{|
   dispatch: Dispatch,
+  ...SelectorProps,
 |}>;
 
 class AccountDetailsScreen extends PureComponent<Props> {
@@ -63,7 +67,7 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state: GlobalState, props) => ({
+export default connectFlowFixMe((state, props) => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
   isActive: getUserIsActive(state, props.navigation.state.params.email),
 }))(AccountDetailsScreen);
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -100,7 +100,11 @@ export type User = {|
   bot_type?: number,
   bot_owner?: string,
 
-  timezone: string,
+  // The ? is for future-proofing. Greg explains in 2020-02, at
+  // https://github.com/zulip/zulip-mobile/pull/3789#discussion_r378554698, that
+  // both human and bot Users will likely end up having a missing timezone
+  // instead of an empty string.
+  timezone?: string,
 
   // avatar_url is synthesized on the server by `get_avatar_field`.
   avatar_url: string | null,
@@ -137,7 +141,12 @@ export type CrossRealmBot = {|
   is_bot: true,
   user_id: number,
 
-  // timezone included since commit 58ee3fa8c (in 1.9.0)
+  // The timezone field has been included since commit 58ee3fa8c (in 1.9.0). Tim
+  // mentions in 2020-02, at
+  // https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-581218576,
+  // that, as of the time of writing, it is always '' for bots, but it may be
+  // omitted later to reduce payload sizes. So, we're future-proofing this field
+  // by making it optional. See comment on the same field in User.
   timezone?: string,
 |};
 
--- a/src/title/ActivityText.js
+++ b/src/title/ActivityText.js
@@ -9,12 +9,17 @@ import { getPresence, getUserStatus } from '../selectors';
 import { presenceToHumanTime } from '../utils/presence';
 import { RawLabel } from '../common';
 
-type Props = $ReadOnly<{
-  dispatch: Dispatch,
+type SelectorProps = $ReadOnly<{|
   presence: UserPresence,
-  style: TextStyleProp,
   userStatus: UserStatus,
-}>;
+|}>;
+
+type Props = $ReadOnly<{|
+  style: TextStyleProp,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
 
 class ActivityText extends PureComponent<Props> {
   render() {
--- a/src/title/ActivityText.js
+++ b/src/title/ActivityText.js
@@ -3,8 +3,8 @@
 import React, { PureComponent } from 'react';
 import type { TextStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 
-import type { UserPresence, UserStatus, Dispatch } from '../types';
-import { connectFlowFixMe } from '../react-redux';
+import type { UserPresence, UserStatus, Dispatch, UserOrBot } from '../types';
+import { connect } from '../react-redux';
 import { getPresence, getUserStatus } from '../selectors';
 import { presenceToHumanTime } from '../utils/presence';
 import { RawLabel } from '../common';
@@ -16,6 +16,7 @@ type SelectorProps = $ReadOnly<{|
 
 type Props = $ReadOnly<{|
   style: TextStyleProp,
+  user: UserOrBot,
 
   dispatch: Dispatch,
   ...SelectorProps,
@@ -35,7 +36,7 @@ class ActivityText extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state, props) => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   presence: getPresence(state)[props.user.email],
   userStatus: getUserStatus(state)[props.user.user_id],
 }))(ActivityText);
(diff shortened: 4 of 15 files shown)


chrisbobbe:
> The changes themselves look alright, but their ordering and partition make it more difficult to tell than it really needs to be.

Ah, I probably should have labeled this WIP, as I'm still learning how to best organize my Zulip commits. For context (and I should have included this in the opening comment), this was already an attempt to address a similar problem that Greg discovered in https://github.com/zulip/zulip-mobile/pull/3765#discussion_r365032960 and which we discussed in person.

Factoring out and aggregating the "pure refactor to extract SelectorProps" changes seems like a fine direction to go; @gnprice, what do you think?

gnprice:
Reading this in detail now, but a quick reply to the thread:
> (I'd probably say "typing" or "types" for these, despite the ambiguity of the former.)

Yep, for that reason I think we've pretty consistently said `types:` when it's about types, and reserved `typing:` for the typing-status feature :slightly_smiling_face: . More recently I've also avoided the ambiguity from the other direction, by saying `typing status:` for the latter.

In this instance, `types:` is pretty good but I might try to make it a bit more specific. When I added this `connect` type-wrapper and `connectFlowFixMe`, and migrated call sites to them, it looks like I said `connect:`:
165e848ea connect: Demonstrate SelectorProps pattern.
81932ca0e connect: Add an error-help comment.
f8c981028 connect: Update libdef, and add our own nontrivial wrapper.
7208ed03b connect: Convert remaining error sites to use "fixme" version.
fb3f71b4e connect: Add a "fixme" version, `connectFlowFixMe`.
cf7ada9f3 connect: Mass-migrate to our type-fixing wrapper.
b58df6071 connect: Make a trivial wrapper that just narrows down the type.

Could also combine those, like `connect types:`.

src/account-info/AccountDetailsScreen.js
```diff
@@ -52,6 +59,6 @@ class AccountDetailsScreen extends PureComponent<Props> {
   }
 }
 
-export default connectFlowFixMe((state: GlobalState, props) => ({
+export default connect((state, props): SelectorProps => ({
```
gnprice: This line will change slightly after #3842, the fix to #3768. Similarly elsewhere.

As is, I think this `SelectorProps` annotation may not actually have any effect! I'm no longer sure what problem I believed it worked around, or why, when I added that pattern last year in #3450. But the new form:
```
export default connect<SelectorProps, _, _>((state, props) => ({
```
successfully fixes #3768 (works around the underlying Flow issue), so that'll be good.

gnprice:
> In particular, the various `This also includes a pure refactor to extract SelectorProps` changes seem a good target for themselves being factored out and aggregated together. (Pure refactors are always easier to verify, and removing them from functionally-active changes makes the latter simpler as well.)

Yeah, I like this idea. In particular I think after this change, what's left would mostly be just a couple of isolated lines at a time: fix a type, maybe add an import for it, and s/connectFlowFixMe/connect/. Which would be great for reading those diffs and immediately seeing exactly what had changed.

src/account-info/AccountDetails.js
```diff
@@ -58,7 +58,7 @@ class AccountDetails extends PureComponent<Props> {
         <View>
           <ActivityText style={styles.largerText} user={user} />
         </View>
-        {user.timezone ? (
+        {user.timezone !== '' && user.timezone !== undefined ? (
```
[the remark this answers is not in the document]
chrisbobbe: Hmm. I don't remember what the semantics was, but I think Greg requested this change, if I'm remembering correctly. @gnprice?

src/title/TitlePrivate.js
```diff
@@ -4,22 +4,31 @@ import React, { PureComponent } from 'react';
...
+  dispatch: Dispatch,
+  ...SelectorProps,
 }>;
 
 class TitlePrivate extends PureComponent<Props> {
   handlePress = () => {
     const { dispatch, user } = this.props;
+    if (!user) {
+      return;
+    }
```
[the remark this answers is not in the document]
chrisbobbe: Done :) (not pushed yet)

src/account-info/AccountDetails.js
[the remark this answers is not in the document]
chrisbobbe: Hmm. I wonder if it's actually the undefined and null cases that are puzzling. user.timestamp's type is string, and some example user objects [here](https://zulipchat.com/api/get-all-users) show that the empty string is a real case.

chrisbobbe:
OK, thanks, both of you!

I've put all the `This also includes a pure refactor to extract SelectorProps` changes together in a single commit. It's long (11 files), but, as you say, it's easier to review NFC commits if they're isolated from functional changes.

I do see some repetitions in some of the other commits; in particular, e49cd8f..9fc4f5e all include "Correctly type the navigation prop using NavigationScreenProp." In two of these two commits, that's the only change, but in the other two, other changes are necessary to remove the connectFlowFixMe, and I think those can't be split out from the NavigationScreenProp change to allow putting all the NavigationScreenProp changes in the same commit.

ray-kraesig:
#3842 is in, which will probably require some merge-conflict resolutions.

chrisbobbe:
> I think I may need some help from one of you to do that investigation, as the API doc just says "The time zone of the user," and I'm not sure if it's meant to say "The time zone of the user _or bot_."

@tabbott do all bots have time zones?

timabbott:
@chrisbobbe I believe bot users will have a `timezone` field of `''`, based on the data model.  We may in the future want to just not send the field for bot users (since it's immutably that value) as part of optimizing payload sizes, so I'd recommend writing the mobile app in a way that treats the value being missing the same as `''`.

chrisbobbe:
OK, just resolved those merge conflicts, and added a commit in the spirit of https://github.com/zulip/zulip-mobile/pull/3842/commits/48d944d2435c19ec064a4bcfa6290cba753e6fcd.

> Can the user.timezone === '' case ever actually happen? (If it does, does it have semantics?)

@ray-kraesig, Tim's comment clears up the confusion about that timezone check, so the check should be correct as-is, meaning our CrossRealmBot type was correct to have timezone be an optional property. I've added a commit with a comment explaining that difference between CrossRealmBot and User (timezone is required in User). 🙂 I also noticed some error handling logic to handle unrecognized timezones, so I tried to consolidate that logic with the "user might not even have a timezone" logic, with comments, so things are hopefully clear.

This should be ready for another review.

chrisbobbe:
(sorry for the long list of force-pushes; essentially, I marked the TopicListScreen commit as fixing https://github.com/zulip/zulip-mobile/issues/2585, but I accidentally said 3585 instead of 2585, then fixed it.)

chrisbobbe:
Oh, it appears that this PR would make progress on https://github.com/zulip/zulip-mobile/issues/3451! It’s not comprehensive, though; there will still be more of these to do.

src/api/modelTypes.js
```diff
@@ -117,6 +117,10 @@ export type CrossRealmBot = {|
   user_id: number,
 
   // timezone included since commit 58ee3fa8c (in 1.9.0)
+  // Currently, bots have a `timezone` field of ''. In future, as Tim mentions
+  // at https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-581218576,
+  // that field may be omitted to reduce payload sizes. So timezone is correctly
+  // an optional property in CrossRealmBot, but correctly required in User.
   timezone?: string,
```
gnprice: "Currently" is perilous in comments. :stuck_out_tongue: The future may eventually come along, and when it does the comment will tend to still be there and so will be wrong.

A simple fix is to say "As of 2020-02".

It'd also be good to integrate what this comment is saying and what the existing comment says, as they're part of the same story (the behavior of this field through the server's history.) I think it amounts to: the `timezone` field is included since that commit, but at least as of 2020-02, it's always `''` for bots. And so as Tim mentions at that link, in the future it may be omitted etc.

gnprice: Oh, one thing that's a bit subtle (as our terminology is confusing) that I see reading @timabbott's comment again: he's talking about *bots*, and that's a wider category than *cross-realm bots*. The latter are a fixed handful like Welcome Bot. The former can be created by users in a given realm... and they show up as values of type `User`. So I think in fact when it comes to this possible future, there isn't a difference between `CrossRealmBot` and `User`, because the latter includes bots too.

Which I guess means it'd be good to make it optional in `User`! That needn't block these other changes, though -- that's just some future-proofing based on Tim's comment in this thread. Should in any case make the comment reflect that, though.

src/account-info/AccountDetails.js
```diff
@@ -43,13 +43,16 @@ class AccountDetails extends PureComponent<Props> {
...
-    // The set of timezone names in the tz database is subject to change over
-    // time. Handle unrecognized timezones by quietly discarding them.
-    let localTime: string | null;
-    try {
-      localTime = `${nowInTimeZone(user.timezone)} Local time`;
-    } catch (err) {
-      localTime = null;
+    let localTime: string | null = null;
+    // `user` will be '' or undefined if it is a bot (see comment at src/api/modelTypes);
+    // `user` will be '' if a human user with no timezone set.
```
gnprice: This sort of knowledge is great to write down at the type definition in `modelTypes.js` -- that way it's there for when we refer to this property from some other code, unrelated to this file.

(Also, per my previous comment interpreting what Tim said, I think the story is a bit more homogeneous than this: it may be `''` or missing/undefined if there's no timezone set, which in particular is true of all bots. The "missing" case is the past for cross-realm bots, and a possible future for... I think probably for all types of users.)

src/topics/TopicListScreen.js
```diff
@@ -2,19 +2,26 @@
...
 import { Screen } from '../common';
 import { topicNarrow } from '../utils/narrow';
 import { getTopicsForStream } from '../selectors';
 import { getStreamForId } from '../subscriptions/subscriptionSelectors';
 import TopicList from './TopicList';
 import { fetchTopics, doNarrow } from '../actions';
 
+type SelectorProps = $ReadOnly<{|
+  stream: Stream,
+  topics: ?(TopicExtended[]),
```
gnprice: nit: commit message says `TopicExtended` where it means `TopicExtended[]`. Also best to parenthesize `?(TopicExtended[])`, same as in the code. (I'm not sure if `?TopicExtended[]` means the same thing, or means `(?TopicExtended)[]`.)

src/title/ActivityText.js
```diff
@@ -3,18 +3,24 @@
...
-  dispatch: Dispatch,
+type SelectorProps = $ReadOnly<{|
   presence: UserPresence,
-  style: TextStyleProp,
   userStatus: UserStatus,
-}>;
+|}>;
+
+type Props = $ReadOnly<{|
+  style: TextStyleProp,
```
gnprice: The first commit puts this in `SelectorProps`; should stay here instead.

src/title/TitlePrivate.js
```diff
@@ -28,8 +38,17 @@ class TitlePrivate extends PureComponent<Props> {
...
+    if (prevProps.user && !this.props.user) {
+      logging.warn('user prop disappeared in TitlePrivate. Not a crash, but unexpected behavior.');
+    }
+  }
+
   render() {
     const { user, color } = this.props;
+    if (!user) {
+      return null;
+    }
```
gnprice: Can shorten the commit's summary line: s/if getAllUsersByEmail returns no user/if no user found/.

src/topics/TopicListScreen.js
```diff
@@ -52,7 +59,7 @@ class TopicListScreen extends PureComponent<Props, State> {
   }
 }
 
-export default connectFlowFixMe((state: GlobalState, props) => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
```
gnprice: It'd be a little cleaner if we apply the `<SelectorProps, _, _>` pattern to each of these as we convert from `connectFlowFixMe`. That way, we're maintaining an invariant that that's the new pattern and that's how we always write `connect` calls that have a `props`-taking callback.

Because you've prepared the content of this change already and it's sitting at this later point in your branch, it's possible to quite efficiently move the pieces of it forward to the earlier points where they'd ideally go: in a `git rebase -i`, edit those earlier commits, and do a command like `git checkout master src/topics/TopicListScreen.js` (or replace "master" with whatever branch you're using) to pull forward just one file's changes at a time.

src/api/modelTypes.js
gnprice: Also just added some jsdoc which hopefully clarifies this some: 60807bde2

src/title/TitlePrivate.js
```diff
@@ -10,11 +10,17 @@ import ActivityText from './ActivityText';
 import { getAllUsersByEmail } from '../users/userSelectors';
 import styles from '../styles';
 import { navigateToAccountDetails } from '../nav/navActions';
+import * as logging from '../utils/logging';
```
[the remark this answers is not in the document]
chrisbobbe: Indeed, thanks for teaching me that!!

src/account-info/AccountDetails.js
chrisbobbe: > it may be `''` or missing/undefined if there's no timezone set, which in particular is true of all bots.

Hmm, are you sure about the first part? I don't think it follows from Tim's [comment](https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-581218576) that human users will ever have it missing/undefined. But I do see that bots may have it missing/undefined, and that all `CrossRealmBot`s, and some, `User`s are bots (the single-realm bots); thanks for explaining the subtlety there and for https://github.com/zulip/zulip-mobile/commit/60807bde20d7e6a43db17d30b75c14f56c540676. 🙂

If I'm understanding correctly, then, these comments by chance ended up being conceptually correct, but I bet it would be more practical to refer to our type aliases, by name, right, so something like this:

```
// `user.timezone` will be '' or undefined for CrossRealmBots (see CrossRealmBot comment at src/api/modelTypes);
// `user.timezone` will be '' for Users, but in future may be '' or undefined (see User comment at modelTypes).
```

src/topics/TopicListScreen.js
chrisbobbe: Wow, this probably saved me >=20m of work. 🙂

chrisbobbe: Oh, how should I explain the change to `<SelectorProps, _, _>` in all those commit messages?

chrisbobbe:
OK, I've made those changes, hopefully correctly! 🙂 

My two questions, I think, are https://github.com/zulip/zulip-mobile/pull/3789#discussion_r378008966 and https://github.com/zulip/zulip-mobile/pull/3789#discussion_r378003799.

src/topics/TopicListScreen.js
gnprice: Great :slightly_smiling_face: 

I think it doesn't actually need explanation in the commit messages, because it's the pattern we now universally use for this kind of call.

src/api/modelTypes.js
```diff
@@ -100,6 +100,12 @@ export type User = {|
   bot_type?: number,
   bot_owner?: string,
 
+  // Tim mentions in 2020-02, at
+  // https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-581218576,
+  // that, as of the time of writing, it is always '' for bots, but it may be
+  // omitted later to reduce payload sizes. TODO: So, we should probably
+  // future-proof this field by making it optional, since realm-specific bots
+  // are assigned the User type, with is_bot true.
```
gnprice: nit: TODO should always be at the start of its line of text. (I.e. after the `//`, or the line's leading `*` in the middle of a `/*`-style comment, but before any text.) That helps it stand out.

I think basically I think of it as introducing a new paragraph.

src/account-info/AccountDetails.js
gnprice: > I don't think it follows from Tim's [comment](https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-581218576) that human users will ever have it missing/undefined.

Hmm, yes, on rereading the comment I agree with that reading.

Still, I expect it's possible for human users to have no meaningful timezone set -- just from an onboarding UX perspective. And spot-checking the right sidebar on czo, I found a user with no "local time" line in their profile popup, which seems likely to mean that.

Then searching through the server's `zerver/models.py` for the timezone field on users, here it is with a nice comment:
```
    # A timezone name from the `tzdata` database, as found in pytz.all_timezones.
    #
    # The longest existing name is 32 characters long, so max_length=40 seems
    # like a safe choice.
    #
    # In Django, the convention is to use an empty string instead of NULL/None
    # for text-based fields. For more information, see
    # https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.Field.null.
    timezone = models.CharField(max_length=40, default=u'')  # type: str
```
So "no known timezone" is represented as `''` inside the server. Which is probably why it's the same way, for now, in the API.

And although Tim's comment about the future only mentioned bots, if we were to make that change I don't see why it'd be conditioned on being a bot or not -- suppressing `timezone: '',` would have the same size benefits for human users as for bots. So as a predictive matter, I think it's likely we'd want to do it for human users the same as for bots. In any case when we follow Tim's actionable advice on the matter:

> I'd recommend writing the mobile app in a way that treats the value being missing the same as `''`.

the easiest way to do that is to apply it the same way for all users, and happily that's also the most broadly future-proof way.

src/api/modelTypes.js
```diff
@@ -137,7 +143,12 @@ export type CrossRealmBot = {|
   is_bot: true,
   user_id: number,
 
-  // timezone included since commit 58ee3fa8c (in 1.9.0)
+  // The timezone field has been included at least since commit 58ee3fa8c (in
+  // 1.9.0). Tim mentions in 2020-02, at
```
gnprice: Not "at least" -- that's the actual commit that introduced it :slightly_smiling_face:

src/api/modelTypes.js
```diff
@@ -100,6 +100,12 @@ export type User = {|
   bot_type?: number,
   bot_owner?: string,
 
+  // Tim mentions in 2020-02, at
```
gnprice: Small thing in this commit message: I've liked using the prefix "api types" for changes here. The "api" name is helpful for saying it's inside `src/api/`, the bindings to the Zulip server API. So in this case `api types [nfc]:`.

(At one level, all types changes are NFC because they don't affect what happens at runtime. But this is more so because it's only in the comments: it doesn't affect what future changes would type-check. So when it already says "types" I generally reserve "[nfc]" for changes, like this one, that make no functional changes even to the types.)

src/account-info/AccountDetails.js
gnprice: > but I bet it would be more practical to refer to our type aliases, by name, right

Yep -- I was just about to add a small comment about this, and then saw you'd already mentioned it here :wink:

chrisbobbe:
OK, I've pushed the latest revision. In particular, as mentioned in person, I was able to go ahead and make timezone optional in User, with no follow-on changes necessary. 🙂

gnprice:
Thanks @chrisbobbe for all your work on this! Merged :slightly_smiling_face: 

I made two small changes just to the summary lines before merge:
* Several summary lines looked like "TitlePrivate types: ...", but made changes to the runtime behavior and not only to the type annotations. I'd like to reserve the "types" prefix for saying the commit touches *only* the types. (This wasn't explicit in my earlier comment https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-578350535 .)
* Several others looked like "ActivityText types: ...", and it's true that they only touched the types, so it's perfectly accurate. But I decided what I'd like to have there even more than "types", to be most helpful when scanning back through the log, is "[nfc]".

  I see my previous comment https://github.com/zulip/zulip-mobile/pull/3789#issuecomment-578350535 about prefixes like "connect types:"; I guess one thing that's different here is that when it isn't a series all with the same prefix like "connect types:", but a bunch of different names of different lengths, the "types" doesn't stand out visibly against all the words but the "[nfc]" does.

So here's the form this series ended up taking:

`$ git log --oneline --reverse @~13..`
628a683a4 connect [nfc]: Extract SelectorProps from Props in 10 files, for clarity.
8cb6aa8a6 EditStreamScreen [nfc]: Fix connectFlowFixMe.
03a025eb4 InviteUsersScreen [nfc]: Fix connectFlowFixMe.
1b090bc06 api types: Make timezone optional on User, for future-proofing.
085771cd5 AccountDetailsScreen: Fix connectFlowFixMe.
5b67d49be AccountDetails: Centralize logic to show/hide localTime.
8811fa813 TopicListScreen [nfc]: Fix connectFlowFixMe.
053e9ebe1 TitleGroup: Fix connectFlowFixMe.
5f3159245 TopicAutocomplete [nfc]: Fix connectFlowFixMe.
660887be0 NotSubscribed [nfc]: Fix connectFlowFixMe.
addc8f69d ActivityText [nfc]: Fix connectFlowFixMe.
3dbb314ac SubscriptionsCard: Fix connectFlowFixMe.
84a60063e TitlePrivate: Fix connectFlowFixMe, render null if no user found.

PR #3795, opened 2020-01-13
Show alert dialog when ActionSheet action fails

### LINKED ISSUE #3788: No error message is shown when deletion fails
If a user tries to delete a message which is too old to be deleted, no error message is shown when the attempt fails.

(See also #2902.)

agrawal-d:
Fixes #3788.

An alert dialog is now shown, instead of silently ignoring the error.

Attached below is a screenshot showing this change:
image

The dialog is similar to what is shown when editing a message fails:
image


agrawal-d:
Why is the Travis CI build failing? I haven't changed anything that should caused build failures.

```
FAILED: android
The command "TERM=dumb tools/test --full ${TEST_SUITE}" exited with 1.
```

src/message/messageActionSheet.js
```diff
@@ -58,7 +58,9 @@ const deleteMessage = async ({ auth, message, dispatch }) => {
   if (isAnOutboxMessage(message)) {
     dispatch(deleteOutboxMessage(message.timestamp));
   } else {
-    api.deleteMessage(auth, message.id);
+    api.deleteMessage(auth, message.id).catch((err: Error) => {
```
jainkuniya: A good follow up of this will be to make sure this message is translated based on user selected language (also for fail to edit message).

agrawal-d: Can you please tell me how I can do that? Th [/howto/translations.md](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/translations.md) does not explain much.

agrawal-d:
Thanks for the review @ray-kraesig . Have made some changes.

Now works for all possible actions, and all errors are handled in a single place.

 A more generic message is shown as the alert title, while the alert body message itself is extracted from the error thrown.
image

@ray-kraesig please review.

src/message/messageActionSheet.js
```diff
@@ -12,7 +12,7 @@ import { navigateToMessageReactionScreen } from '../nav/navActions';
 // TODO really this belongs in a libdef.
 export type ShowActionSheetWithOptions = (
   { options: string[], cancelButtonIndex: number },
-  (number) => void,
+  (number) => Promise<void>,
```
ray-kraesig: This isn't our type to modify, unfortunately – as the comment notes, it shouldn't really be here at all. (Also, this modification results in a Flow error.)

On the other hand, `ButtonDescription` (seen below) *is* ours.

src/message/messageActionSheet.js
```diff
@@ -232,15 +232,19 @@ export const showActionSheet = (
...
+      await allButtons[optionCodes[buttonIndex]]({
+        dispatch,
+        subscriptions: params.backgroundData.subscriptions,
+        auth: params.backgroundData.auth,
+        ownEmail: params.backgroundData.ownEmail,
+        _,
+        ...params,
+      });
+    } catch (err) {
+      Alert.alert('Couldnt perform the requested action', err.message);
```
ray-kraesig: Typo: `Couldn't` (although preferentially `Could not`).

That said, it's probably worth the trouble to do better than a generic `Could not perform the requested action` here. We can associate a failure message with each action (`Could not delete message`, etc.) the same way we're associating a title.

agrawal-d: I have replaced the message with a simpe 'Error'. This will not be confusing because the alert message contains context-specific detailed message.

src/message/messageActionSheet.js
agrawal-d: Fixed

agrawal-d:
Thanks for reviewing, @ray-kraesig. Have made some changes.

agrawal-d:
image

src/message/messageActionSheet.js
ray-kraesig: Counterpoint:

Screenshot_20200116_130132

agrawal-d:
@ray-kraesig , so how do you propose we do it? Hardcode the alert titles for each action?

agrawal-d:
This looks better:
image

src/message/messageActionSheet.js
```diff
@@ -233,14 +233,21 @@ export const showActionSheet = (
...
+        await allButtons[optionCodes[buttonIndex]]({
+          dispatch,
+          subscriptions: params.backgroundData.subscriptions,
+          auth: params.backgroundData.auth,
+          ownEmail: params.backgroundData.ownEmail,
+          _,
+          ...params,
+        });
+      } catch (err) {
+        const titleString = allButtons[optionCodes[buttonIndex]].title;
```
ray-kraesig: It's harmless right now, but it's probably better to cache `allButtons[optionCodes[buttonIndex]]` before the `await` is performed.

(`allButtons` *shouldn't* ever be modified – but if it is, we might end up accessing different values of `allButtons` before and after the `await`, if two actions are performed in quick succession before the first one fails.)

agrawal-d: Doesn't the callback have a closure over `allButtons[optionCodes[buttonIndex]]`, so that even if another action is performed successively, the previous callback isn't affected?

agrawal-d:
@ray-kraesig have pushed some changes. Please review.

src/message/messageActionSheet.js
agrawal-d: Not being used in the latest change at all, so haven't cached it.

src/message/messageActionSheet.js
```diff
@@ -141,6 +141,22 @@ const allButtonsRaw = {
...
+  'Failed to edit message',
+  'Failed to delete message',
+  'Failed to star message',
+  'Failed to unstar message',
+  'Failed to show reactions',
+  'Failed to unmute topic',
+  'Failed to mute topic',
+  'Failed to mute stream',
+  'Failed to unmute stream',
+];
```
ray-kraesig: There's nothing keeping this lined up with the actions themselves. If your earlier `delete topic` PR were applied, for example, this would be suddenly and quietly broken.

It would be safer to associate actions with failure messages in the same way they're associated with titles, as I suggested earlier.

agrawal-d: Can you please give me an example of how you expect the alert titles to be associated, because I am unable to understand what we want here. Thanks. 

I Hardcoded the error titles because you mentioned that this is expected in a comment. But I guess you meant something else.

static/translations/messages_en.json
```diff
@@ -119,6 +119,19 @@
...
+  "Failed to share message":"Failed to share message",
+  "Failed to edit message":"Failed to edit message",
+  "Failed to delete message":"Failed to delete message",
+  "Failed to star message":"Failed to star message",
+  "Failed to unstar message":"Failed to unstar message",
+  "Failed to show reactions":"Failed to show reactions",
+  "Failed to unmute topic":"Failed to unmute topic",
+  "Failed to mute topic":"Failed to mute topic",
+  "Failed to mute stream":"Failed to mute stream",
+  "Failed to unmute stream":"Failed to unmute stream",
```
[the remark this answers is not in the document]
agrawal-d: Im sorry, I am used to the prettier taking care of formatting, but I guess it doesn't add space after colons. Will keep this in mind in the future.

agrawal-d:
Thank you for the review, @ray-kraesig. I have (hopefully) made the changes your requested, and added a commit message.

I have not added any documentary comments in the code, because I did not feel that this change needs to be documented, but if required, I will add some.

Please review these changes, thanks.

static/translations/messages_en.json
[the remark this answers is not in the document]
ray-kraesig: `prettier` is intended for code formatting, and not necessarily data formatting. We don't let it touch `.json` files: those are typically both generated and consumed by tooling.

agrawal-d:
Thanks for the review, @ray-kraesig. Have made some changes.

The `errorMessage` property in type `ButtonDescription` is optional because the callback associated with the button `Cancel` ( and possibly other buttons we may add in the future ) can not throw an error.

Please review these changes, thanks!

agrawal-d:
@ray-kraesig I went ahead with the straightforward solution. Please review, thanks!

> The commit summary is also a few too many characters long for GitHub to be happy with it. (Our current documented limit is 76, but perhaps we should reduce this slightly.)

Probably because of the 'Verified' commit badge, as I have added a GPG key to my GitHub account.

src/message/messageActionSheet.js
```diff
@@ -54,77 +56,95 @@ const copyToClipboard = async ({ _, auth, message }) => {
...
+    await api.toggleMuteStream(auth, sub.stream_id, true);
   }
 };
 muteStream.title = 'Mute stream';
+muteStream.errorMessage = 'Failed to mute stream';
 
-const starMessage = ({ auth, message }) => {
-  api.toggleMessageStarred(auth, [message.id], true);
+const starMessage = async ({ auth, message }) => {
+  await api.toggleMessageStarred(auth, [message.id], true);
```
ray-kraesig: It's not obvious, but this is `await`ing on void.

(That's really a bug in `toggleMessageStarred`, though.)

ray-kraesig:
– and merged. Thanks!

PR #3797, opened 2020-01-13
MessageActionSheet: Remove reply option from ActionSheet when in a topic narrow

### LINKED ISSUE #3528: Don't provide 'Reply' option in message action sheet when narrowed to the `topic` narrow.
This is unexpected behaviour and bad design to have a feature in the app which essentially does nothing -- in particular situations. 

The 'Reply' option in `messageActionSheet` should not be displayed when in the topic narrow since it can potentially confuse users.

See [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Reply.20not.20working/near/754610) for reference.

agrawal-d:
Fixes #3528 
report-reply

Earlier, a `reply` option was also shown in the messageActionSheet, which essentially performed no action, since the user is already in a topic narrow.

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -2,7 +2,13 @@
 import { Clipboard, Share } from 'react-native';
 import type { Auth, Dispatch, GetText, Message, Narrow, Outbox, Subscription } from '../types';
 import type { BackgroundData } from '../webview/MessageList';
-import { getNarrowFromMessage, isHomeNarrow, isSpecialNarrow } from '../utils/narrow';
+import {
+  getNarrowFromMessage,
+  isHomeNarrow,
+  isPrivateNarrow,
+  isSpecialNarrow,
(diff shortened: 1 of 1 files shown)


agrawal-d:
@gnprice @ray-kraesig can you please review this pull request?

ray-kraesig:
Looks mostly good. However:

* The commit message needs the text `Fixes #3528.`
* The commit summary is too long; it should be [76 characters max](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages).

(You may also want to address my recent comment on the linked issue, but it's not strictly necessary.)

agrawal-d:
Thanks for the review. I have made the changes.
>(You may also want to address my recent comment on the linked issue, but it's not strictly necessary.)
This too has been implemented now.

ray-kraesig:
Merged, with minor formatting tweaks. Thanks!

PR #3801, opened 2020-01-14
docs: add react-devtools information

agrawal-d:
Adds information about `react-devtools`.

### DIFF
--- a/docs/howto/debugging.md
+++ b/docs/howto/debugging.md
@@ -11,6 +11,7 @@ A variety of tools are available to help us do that.
   especially React and Redux
   * ... [with the Chrome Developer Tools](#chrome-devtools) /
     "Debug JS remotely"
+  * ... [with React DevTools](#react-devtools)
   * ... [with Reactotron](#reactotron)
   * ... [with redux-logger](#redux-logger)
 * [Debugging the message list](#webview) in its WebView
@@ -79,6 +80,23 @@ See discussion of `redux-logger` below.
 See also [in the "Troubleshooting" section below](#debug-remotely-never-loads).
 
 [chrome-devtools-device]: https://facebook.github.io/react-native/docs/debugging.html#debugging-on-a-device-with-chrome-developer-tools
+
+
+<div id="react-devtools">
+
+## React Devtools
+
+The standalone version of React DevTools, which is an electron app, can be used
+to debug  the react-native component hierarchy. Refer to
+[the docs][react-devtools-docs] for instructions on usage with React Native.
+
+As of 2020-01, the latest version of react-devtools (v4) does not support any
+released version of React Native, and v3 needs to be installed. This can be
+done using the command `yarn global add react-devtools@3`. See
+[this blog post][reactjs-blog-post] for complete compatibility information.
+
+[react-devtools-docs]: https://facebook.github.io/react-native/docs/debugging.html#react-developer-tools
+[reactjs-blog-post]: https://reactjs.org/blog/2019/08/15/new-react-devtools.html#which-versions-of-react-are-supported
 
 
 <div id="reactotron" />
--- a/docs/howto/debugging.md
+++ b/docs/howto/debugging.md
@@ -84,19 +84,19 @@ See also [in the "Troubleshooting" section below](#debug-remotely-never-loads).
 
 <div id="react-devtools">
 
-## React Devtools
+## React DevTools
 
-The standalone version of React DevTools, which is an electron app, can be used
-to debug  the react-native component hierarchy. Refer to
-[the docs][react-devtools-docs] for instructions on usage with React Native.
+The standalone version of the React Developer Tools, an Electron app,
+can be used to debug the React component hierarchy in our app.  See
+[RN's docs][rn-react-devtools] on how to use it with RN.
 
-As of 2020-01, the latest version of react-devtools (v4) does not support any
-released version of React Native, and v3 needs to be installed. This can be
-done using the command `yarn global add react-devtools@3`. See
-[this blog post][reactjs-blog-post] for complete compatibility information.
+As of 2020-01, the latest version (v4) of React DevTools does not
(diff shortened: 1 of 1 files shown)


agrawal-d:
@gnprice please review.

docs/howto/debugging.md
```diff
@@ -81,6 +82,13 @@ See also [in the "Troubleshooting" section below](#debug-remotely-never-loads).
 [chrome-devtools-device]: https://facebook.github.io/react-native/docs/debugging.html#debugging-on-a-device-with-chrome-developer-tools
 
 
+<div id="react-devtools">
+
+## React Devtools
+
+The standalone version of React DevTools, which is an electron app, can be used to inspect the react-native components of the app. Since the version of react-native we use is `0.59`, the latest version (v4) of react-devtools is not supported, and v3 needs to be installed.  This can be done using the command `npm i -g react-devtools@^3` See [this blog post](https://reactjs.org/blog/2019/08/15/new-react-devtools.html#which-versions-of-react-are-supported) for complete compability information. Refer to [it's npm page](https://www.npmjs.com/package/react-devtools) for instructions on how to use with React Native.
```
agrawal-d: `react-native` 0.61.5  is the current version, and so 0.62 should be released soon. The wording of this doc makes it future-proof.

gnprice: As discussed in chat, this isn't really about the version of RN we use. Rather, it's that v4 *does not support any released version of React Native*, as of 2020-01.

This also touches on a useful tactic for writing docs: where possible, avoid writing specifics which will change over time, like "we use RN v0.59". When we upgrade, that either becomes one more thing we need to update, or it becomes wrong and potentially misleading. (And we'll have two upgrades *before* the v0.62 upgrade that lets us delete this bit of docs entirely.) Instead, say things that stay true. When that's tricky, you can always add "as of *today's month*" so readers in the future know *when* the statement was true.

gnprice: Thanks for adding this section! We should have a section on this tool -- I was a bit surprised to see that we don't have one already. :slightly_smiling_face: 

As a matter of information architecture: the focus of this section should be documenting the fact that there *is* this tool, and what it's for.

That part doesn't need to be very long -- the first and last sentences you have here are basically enough. But two things to make it clearer:
* Should add a link to the section of the RN debugging doc: https://facebook.github.io/react-native/docs/debugging.html#react-developer-tools that discusses this tool.
* Those sentences that describe the tool as a whole should be their own paragraph, right at the start of this new section.

  Then the rest of what's here, about compatibility with different versions, should be a separate paragraph after that one. That's a more specific detail, so it should be separated to reflect that.

gnprice: Rather than the `npm` CLI, we use `yarn`. For consistency, in this doc let's recommend `yarn global add` rather than `npm i -g`.

agrawal-d:
Thanks for the detailed review, @gnprice! It would have taken you much less time to make the changes yourself, so I really appreciate it - the reviews from the Zulip developer community are very educational! :D

docs/howto/debugging.md
```diff
@@ -81,6 +82,23 @@ See also [in the "Troubleshooting" section below](#debug-remotely-never-loads).
...
+for instructions on how to use with React Native.
+
+As of today's month, we do not support the latest version (v4) of
+react-devtools, and v3 needs to be installed.  This can be done using the
+command `yarn global add react-devtools@3`. See
+[this blog post](https://reactjs.org/blog/2019/08/15/new-react-devtools.html#which-versions-of-react-are-supported)
+for complete compability information.
+
+
+
```
agrawal-d: All lines are less than 80 columns now, so some lines may look tiny because the words following the newline are long.

agrawal-d:
@gnprice can you please have a look at this? Thanks!

docs/howto/debugging.md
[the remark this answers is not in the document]
gnprice: Bump -- this is still needed.

docs/howto/debugging.md
gnprice: Doesn't look tiny to me!

In any case, yep, it's normal that some lines will be short when paragraphs are properly wrapped :slightly_smiling_face:

docs/howto/debugging.md
gnprice: > As of today's month, we do not support the latest version (v4) of
> react-devtools, and v3 needs to be installed.

Hmm, I guess this point wasn't clear. This sentence is still something that will change over time, and stop being true. What I meant was to write down the actual month that you're writing the text in, like I did in the description above: "as of 2020-01".

But again, this really isn't about what *we* do or don't support. Rather it's that react-devtools v4 *does not support any released version of React Native*, quite beyond our control. So that's the thing we should say.

docs/howto/debugging.md
```diff
@@ -81,6 +82,23 @@ See also [in the "Troubleshooting" section below](#debug-remotely-never-loads).
 [chrome-devtools-device]: https://facebook.github.io/react-native/docs/debugging.html#debugging-on-a-device-with-chrome-developer-tools
 
 
+<div id="react-devtools">
+
+## React Devtools
+
+The standalone version of React DevTools, which is an electron app, can be used
+to debug  the react-native component hierarchy. Refer to
+[the docs](https://facebook.github.io/react-native/docs/debugging.html#react-sdeveloper-tools)
```
gnprice: An important step when writing documentation with links in it: test out the links to make sure you don't have a typo in them :wink:

gnprice:
Thanks for the revisions!

See the comment just above, and also replies further up in the thread -- GitHub can make those easy to miss.

agrawal-d:
Thanks for the review, @gnprice. Have pushed some changes.

gnprice:
Merged -- thanks @agrawal-d ! I added a commit on top with some small edits; take a look.

Issue #3784, opened 2020-01-10
Properly handle alien 4xx responses for `/register` calls


ray-kraesig:
When calling `/register`, it's entirely possible that the server specified as a realm _exists_, but is not a valid Zulip realm... and that it will respond to `/register` with no JSON or with invalid JSON.

Presently, this is not propagated as an `ApiError(404, ...), but as a generic `Error`... which the `tryUntilSuccessful` invoked by `doInitialFetch` fails to understand.

(Possibly related: #3483.)

gnprice:
Hmm true!

I think this is a pretty unlikely scenario to get into -- because we only attempt `/register` after the user has successfully logged in, and that requires the server to have played along with `/server_settings` and then with one of the auth flows the client understands. So that means it was... either a Zulip server, or basically some kind of imitation of one, at least at the time we logged in.

Probably the most likely way this could happen is if it *was* once a Zulip server, and then got repurposed, or turned into a domain parking page or something.

The relationship between `tryUntilSuccessful` and specific classes of errors is kind of a mess, and a gotcha when looking at its call site. If we make any changes to this logic, we should probably start by finding a way to make that interface more explicit about what it really means. (But unless there's a reason this situation is less unlikely than it looks, I think it's not a priority compared to other issues.)

ray-kraesig:
I should clarify that this is not hypothetical: it's based on Sentry logs. We were fortunate enough to have a small number of such servers respond with parseable JSON.

(I can't easily determine how many of the generic "Bad response from server: undefined" reports are giving us parking pages, but a quick sampling suggests it's probably not many of them.)

gnprice:
Ah, that's good to know!

Want to paste a traceback and/or a Sentry link? (Even though the linked-to page won't be publicly readable, it's helpful at least for others of us with access.)

ray-kraesig:
https://sentry.io/organizations/zulip/issues/1422486962/
https://sentry.io/organizations/zulip/issues/1414991211/
https://sentry.io/organizations/zulip/issues/1412778322/

And perhaps more generally, [the search query](https://sentry.io/organizations/zulip/issues/?project=191284&query=is%3Aunresolved+%22Bad+response+from+server%22) that presently finds these.

gnprice:
Thanks! Looking at those three:
* Two are actually at `/server_settings`, not `/register`.
[...]

PR #3807, opened 2020-01-15
account details screen: Display deactivation information for user

### LINKED ISSUE #3551: Say a user is deactivated when showing their profile or composing PM
@jainkuniya looked at how the webapp handles deactivated users, and found two helpful things we should do:

> Deactivated status is visible in user profile and at the time of sending PM.
> image
> image

(Moved this from https://github.com/zulip/zulip-mobile/pull/3110#issuecomment-505607319 .)

Let's do those two things.
* Our profile screen is a lot like the user popover in the webapp, so we can use basically the same design.
* For the compose case: let's show a message above the compose box, if you focus the compose box while narrowed to a PM thread that includes a deactivated user.

This isn't a priority, so let's leave it alone for now. Filing it just so it's easy to pick up in the future.

agrawal-d:
Fixes #3551.

Two changes:
1. Deactivation status is shown.
1. Button text is "View private messages" instead of "Send private message"

**Before**: (Assume Cyphase is deactivated ) 
before

**After**:
after

I had to fiddle around a  lot to understand to get the relevant data, here is the reasoning why it is done in this manner:
 - The get-all-users *is not* used by the mobile app to fetch user lists ( this is the list that contains is_deactivated data ). Instead, /register event queue is used.
- So, there is no direct way to get if a user is deactivated. However, `getActiveUsersByEmail()` returns only activated users, comparing it with current user can give us `is_deactivated`.
- A lot of informative and detailed comments are present in `/src/api/modelTypes.js:47` and  `/src/users/userSelectors.js`, the reviewer may go through them.

I *really* enjoyed doing this :D

### DIFF
--- a/src/account-info/AccountDetailsScreen.js
+++ b/src/account-info/AccountDetailsScreen.js
@@ -5,20 +5,28 @@ import { StyleSheet } from 'react-native';
 import type { Dispatch, GlobalState, User } from '../types';
 import { connectFlowFixMe } from '../react-redux';
 import { getAccountDetailsUserForEmail } from '../selectors';
-import { Screen, ZulipButton } from '../common';
+import { Screen, ZulipButton, Label } from '../common';
 import { IconPrivateChat } from '../common/Icons';
 import { privateNarrow } from '../utils/narrow';
 import AccountDetails from './AccountDetails';
 import { doNarrow } from '../actions';
+import { getUserIsActive } from '../users/userSelectors';
 
 const styles = StyleSheet.create({
   pmButton: {
     marginHorizontal: 16,
   },
+  deactivatedText: {
+    textAlign: 'center',
+    paddingBottom: 20,
+    fontStyle: 'italic',
+    fontSize: 18,
+  },
 });
 
 type Props = $ReadOnly<{|
   user: User,
+  isActive: boolean,
   dispatch: Dispatch,
 |}>;
 
@@ -29,7 +37,7 @@ class AccountDetailsScreen extends PureComponent<Props> {
   };
 
   render() {
-    const { user } = this.props;
+    const { user, isActive } = this.props;
     const title = {
       text: '{_}',
       values: {
@@ -41,9 +49,12 @@ class AccountDetailsScreen extends PureComponent<Props> {
     return (
       <Screen title={title}>
         <AccountDetails user={user} />
+        {!isActive && (
+          <Label style={styles.deactivatedText} text="(This user has been deactivated)" />
+        )}
         <ZulipButton
           style={styles.pmButton}
-          text="Send private message"
+          text={isActive ? 'Send private message' : 'View private messages'}
           onPress={this.handleChatPress}
           Icon={IconPrivateChat}
         />
@@ -54,4 +65,5 @@ class AccountDetailsScreen extends PureComponent<Props> {
 
 export default connectFlowFixMe((state: GlobalState, props) => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
+  isActive: getUserIsActive(state, props.navigation.state.params.email),
 }))(AccountDetailsScreen);
--- a/src/users/__tests__/userSelectors-test.js
+++ b/src/users/__tests__/userSelectors-test.js
@@ -6,6 +6,7 @@ import {
   getAllUsersById,
   getUsersById,
   getUsersSansMe,
+  getUserIsActive,
 } from '../userSelectors';
 import * as eg from '../../__tests__/exampleData';
 
@@ -122,5 +123,27 @@ describe('getUsersSansMe', () => {
       realm: eg.realmState({ email: eg.selfUser.email }),
     });
     expect(getUsersSansMe(state)).toEqual([eg.otherUser]);
+  });
+});
+
+describe('getUserIsActive', () => {
+  const state = eg.reduxState({
+    users: [eg.selfUser],
+    realm: {
+      ...eg.baseReduxState.realm,
+      crossRealmBots: [eg.crossRealmBot],
+      nonActiveUsers: [eg.makeUser(), eg.makeUser()],
+    },
+  });
+  test('returns false for a user that has been deactivated', () => {
+    expect(getUserIsActive(state, state.realm.nonActiveUsers[0].email)).toBeFalse();
+  });
+
+  test('returns true for a user that has not been deactivated', () => {
+    expect(getUserIsActive(state, state.users[0].email)).toBeTrue();
+  });
+
+  test('returns true for a cross realm bot', () => {
+    expect(getUserIsActive(state, state.realm.crossRealmBots[0].email)).toBeTrue();
   });
 });
--- a/src/users/userSelectors.js
+++ b/src/users/userSelectors.js
@@ -173,3 +173,19 @@ export const getAccountDetailsUserForEmail: Selector<UserOrBot, string> = create
     );
   },
 );
+
+/**
+ * The activation status of a user - whether or not a user/bot has
+ * been deactivated (`is_active` in the Zulip API).
+ */
+export const getUserIsActive: Selector<boolean, string> = createSelector(
+  (state, email) => email,
+  state => getActiveUsersByEmail(state),
+  (email, usersByEmail) => {
+    if (!email) {
+      return false;
+    } else {
+      return usersByEmail.has(email);
+    }
+  },
+);
--- a/static/translations/messages_en.json
+++ b/static/translations/messages_en.json
@@ -86,6 +86,8 @@
   "Jot down something": "Jot down something",
   "Message {recipient}": "Message {recipient}",
   "Send private message": "Send private message",
+  "View private messages": "View private messages",
+  "(This user has been deactivated)": "(This user has been deactivated)",
   "Forgot password?": "Forgot password?",
   "Members": "Members",
   "Recipients": "Recipients",


agrawal-d:
On second thoughts, removing the PM button is a wrong idea, because one might want to view past messages. I will revert the PM button, and instead, create a separate PR for disabling the send button on deactivated profile PM, and also attempt to fix the @undefined bug.

agrawal-d:
In a separate Pull Request, I will fix the _PM deactivated person screen_, as there are three bugs when PMing a deactivated profile
image

   1. The user shows up as @undefined in the compose box.
   2. Actually sending a message results in an indefinite spinner being shown.
   3. While the send button is disabled, it is still possible to send attachments/images ( which get infinite spinners, this is tracked by a different issue (https://github.com/zulip/zulip-mobile/issues/3784)

All three will be fixed by simply removing the compose box, and showing a view similar to the #announce one:
image

I have created an issue for this : https://github.com/zulip/zulip-mobile/issues/3812

agrawal-d:
@gnprice @agrawal-d  please review :)

src/account-info/AccountDetailsScreen.js
```diff
@@ -29,29 +36,51 @@ class AccountDetailsScreen extends PureComponent<Props> {
...
     };
-
-    return (
-      <Screen title={title}>
-        <AccountDetails user={user} />
+    let deactivatedText = null;
+    let privateMessageButton = null;
+    if (isDeactivated) {
+      deactivatedText = (
+        <RawLabel style={styles.deactivatedText} text="(This user has been deactivated)" />
```
jainkuniya: Can we use `<Label />` instead so that this text gets translated in user selected language.

src/account-info/AccountDetailsScreen.js
```diff
@@ -29,29 +36,51 @@ class AccountDetailsScreen extends PureComponent<Props> {
...
+      );
+      privateMessageButton = (
```
jainkuniya: Here both the components are same, just the difference is in `text`. This will look much more cleaner this this
```
text={isDeactivated ? "View private messages" : "Send private message"}
```

src/account-info/AccountDetailsScreen.js
```diff
@@ -29,29 +36,51 @@ class AccountDetailsScreen extends PureComponent<Props> {
...
+          text="View private messages"
+          onPress={this.handleChatPress}
+          Icon={IconPrivateChat}
+        />
+      );
+    } else {
+      privateMessageButton = (
         <ZulipButton
           style={styles.pmButton}
           text="Send private message"
```
jainkuniya: I will prefer `Open chat`. This will work in both the cases.

src/account-info/AccountDetailsScreen.js
```diff
@@ -29,29 +36,51 @@ class AccountDetailsScreen extends PureComponent<Props> {
...
+        {deactivatedText}
+        {privateMessageButton}
       </Screen>
     );
   }
 }
 
 export default connectFlowFixMe((state: GlobalState, props) => ({
   user: getAccountDetailsUserForEmail(state, props.navigation.state.params.email),
+  isDeactivated:
```
jainkuniya: Can you extract this to a selector? or better move this logic to `getAccountDetailsUserForEmail `, which by name says account details for email (and deactivated information belongs to this).

src/account-info/AccountDetailsScreen.js
agrawal-d: I had used label first. It does not adapt to theme changes ( dark mode ).

jainkuniya: Can you please provide more details on this? maybe screenshot.

Then if there are some layout issues, then it will be better to first enhance `Label` to make it look better in dark theme.

agrawal-d: Is there a way that text can be translated, before it is added to the RawLabel?

agrawal-d:
@jainkuniya this is how test/label looks in dark mode, which is why I used RawLabel:
image

agrawal-d:
@jainkuniya  Have pushed some changes, please review. I have separated the deactivation text into a variable, so that it can be translated, and RawLabel can be used.

src/account-info/AccountDetailsScreen.js
agrawal-d: Done

src/account-info/AccountDetailsScreen.js
agrawal-d: Changed to `Open chat`

src/account-info/AccountDetailsScreen.js
agrawal-d: **Fixed**

src/account-info/AccountDetailsScreen.js
```diff
@@ -37,13 +43,20 @@ class AccountDetailsScreen extends PureComponent<Props> {
...
+    }
 
     return (
       <Screen title={title}>
         <AccountDetails user={user} />
+        {deactivatedText}
         <ZulipButton
           style={styles.pmButton}
-          text="Send private message"
+          text="Open chat"
```
ray-kraesig: The webapp prefers "private messages", eschewing "chat" in this context (and almost all others); we should follow suit. Your original suggestion is probably a better option.

(Also, a new string will need to be added to `messages_en.json` regardless.)

src/users/userSelectors.js
```diff
@@ -159,13 +159,14 @@ export const getActiveUsersByEmail: Selector<Map<string, UserOrBot>> = createSel
 export const getAccountDetailsUserForEmail: Selector<UserOrBot, string> = createSelector(
   (state, email) => email,
   state => getAllUsersByEmail(state),
-  (email, allUsersByEmail) => {
+  (state, email) => getActiveUsersByEmail(state).has(email),
+  (email, allUsersByEmail, is_active) => {
```
ray-kraesig: Style nit: prefer `camelCase` to `snake_case` for anything that isn't being serialized to the network.

src/account-info/AccountDetailsScreen.js
```diff
@@ -37,13 +43,20 @@ class AccountDetailsScreen extends PureComponent<Props> {
         _: user.full_name || ' ',
       },
     };
+    let deactivatedText = null;
+    if (user.is_active === false) {
+      deactivatedText = (
+        <Label style={styles.deactivatedText} text="(This user/bot has been deactivated)" />
```
ray-kraesig: This string needs to go in `messages_en.json`.

(Also, let's stick with `(This user has been deactivated)`, to mirror the webapp.)

agrawal-d:

src/users/userSelectors.js
```diff
@@ -156,6 +156,23 @@ export const getActiveUsersByEmail: Selector<Map<string, UserOrBot>> = createSel
...
+    const deactivatedUsersByEmail = [];
+    allUsers.forEach(user => {
+      if (!activeUsersAndBots.includes(user)) {
+        deactivatedUsersByEmail.push([user.email, user]);
+      }
+    });
+    return new Map(deactivatedUsersByEmail);
+  },
+);
+
```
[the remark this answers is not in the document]
agrawal-d: Would a selector that only returns a `boolean` be acceptable? Even I thought that n^2 is too much, but I thought that a selector must return a subset of the list of users.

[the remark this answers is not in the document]
agrawal-d: I think It should be acceptable, because there exists `getOwnEmail`.

agrawal-d:
Thanks for reviewing @ray-kraesig. Have pushed some changes. Please have a look.

agrawal-d:
@ray-kraesig thanks for the review. I have made the changes that you requested:

- wrote a test for the new selector and separated that into a new commit
- fixed the flow type for the selector
- changed the docstring comment for the selector to be more clear as you suggested
- changed commit message to show change is in harmony with the Zulip webapp.

Please review. Thanks!

src/users/__tests__/userSelectors-test.js
```diff
@@ -124,3 +125,12 @@ describe('getUsersSansMe', () => {
...
   });
 });
+
+describe('getUserIsActive', () => {
+  test('returns false for a user that has been deactivated', () => {
+    const users = [eg.makeUser(), eg.makeUser(), eg.makeUser()];
+    const deactivatedUser = eg.makeUser();
+    const state = eg.reduxState({ users });
+    expect(getUserIsActive(state, deactivatedUser.email)).toBeFalse();
+  });
```
ray-kraesig: You also need to check that it returns `true` for a user who has *not* been deactivated! And, less obviously, for a bot (`realm.crossRealmBots`).

Although the current implementation doesn't rely on it, you should probably include `deactivatedUser` in the redux-state's `realm.nonActiveUsers`. Alternatively, you can just reuse the state setup from several of the tests above.

(Optional useful trick: you can set up a shared `state` in the `describe` block and use it in its sibling `test` blocks.)

agrawal-d:
Yikes :scream: thanks to your suggestion to add a test for `cross realm bots`, I quickly realized that these was a small bug in the selector I created - I used `getUsersByEmail` instead of `getActiveUsersByEmail`. `getUsersByEmail` does not contain cross reamp bots!

ray-kraesig:
Merged. (I've added `Fixes #3551.` to the latter commit, for ease of reference later, but the commits are otherwise unchanged.)

Thanks!

PR #3811, opened 2020-01-16
Modify/don't provide 'Reply' option for messages that can't be replied to 

### LINKED ISSUE #3810: Don't provide 'Reply' option for messages that can't be replied to
... *e.g.* (and possibly *i.e.*), messages in admin-only streams when the active user is not an admin.

(See also #3528.)

agrawal-d:
Fixes #3810
- Instead of removing the reply option in an announcement stream, I have modified it to show 'Navigate to topic', since one might still want to go to the topic narrow, where the reply option is already disabled.
- Option remains 'Reply' is the app user is an admin.
- Also removes reply option in group PMs.

### Screenshots
image
image


agrawal-d:
@ray-kraesig please review.

src/webview/MessageList.js
```diff
@@ -75,6 +75,7 @@ export type BackgroundData = $ReadOnly<{
   mute: MuteState,
   ownEmail: string,
   ownUserId: number,
+  ownIsAdmin: boolean,
```
ray-kraesig: `isAdmin` will do. (The `own` affix here is of questionable utility anyway, IMO, even where it's grammatical.)

Also, please make sure the ordering of properties is consistent here and below.

src/message/messageActionSheet.js
```diff
@@ -199,6 +201,9 @@ export const constructMessageActionButtons = ({
     buttons.push('addReaction');
   }
   if (!isAnOutboxMessage(message) && !isTopicNarrow(narrow) && !isPrivateNarrow(narrow)) {
+    if (isAnnouncementOnly) {
+      reply.title = 'Navigate to topic';
+    }
```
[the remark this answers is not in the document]
agrawal-d: I feel that `Navigate to topic` is much better than `Reply` for this action anyway, because it **does not** open up a reply dialog as expected, and actually performs the action of navigating to topic.

This has often been confusing to me, and to other users as well, I think.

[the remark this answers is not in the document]
agrawal-d: Will it be alright if I completely replace the `Reply` button with `Navigate to topic` button?

agrawal-d:

agrawal-d:
Blocked until https://github.com/zulip/zulip-mobile/pull/3795 is merged.

agrawal-d:
@ray-kraesig please review these changes, thank you!

ray-kraesig:
Code is good. Some commit-message issues:

* Both commit summaries are `section: uncapitalized sentence with a period.`, which is neither the preferred Zulip official standard (capitalization required) nor the less-preferred git.git style (no period).
* The second commit doesn't actually have anything to do with #3810 and shouldn't say it fixes it.
* The second commit's message says it eliminates "Reply" in topic narrows... but that's already true in `master`.

The first and second are uninteresting (if it were just those two I would do the copyediting myself before merging), but the third is odd enough that I have to ask what the intent was there.

agrawal-d:
For your second point - I thought the second commit's message made sense because it _is_ removing the Reply option in a place where it shouldn't be possible to reply - group private messages. I have removed it now.

For your third point - I made a mistake 😓 - Instead of documenting the change I made, I documented the entire feature.

I have reworded both the commits.

ray-kraesig:
> For your second point - I thought the second commit's message made sense because it _is_ removing the Reply option in a place where it shouldn't be possible to reply - group private messages. I have removed it now.

It's entirely possible to reply to a group PM, though – you can still type into the compose box and send a message and have it be seen; it's just that the current "Reply" option doesn't have an effect there.

(This contrasts with announcement-only streams, where even if you send a message to the server it won't be accepted.)

> For your third point - I made a mistake  - Instead of documenting the change I made, I documented the entire feature.

I figured that was all it was, but I had to ask.

> I have reworded both the commit messages.

Missed the commit summary on the second one, though. :slightly_smiling_face: Amended, and merging. Thanks!

src/webview/MessageList.js
gnprice: > (The `own` affix here is of questionable utility anyway, IMO, even where it's grammatical.)

I think it's important for "email" and "user id" and "user". There are lots and lots of places we have a variable (or parameter or property) named `email` or `userId` or `user`, and they normally mean "(the email or ID of) the user this widget, or this scrap of data, is about". When it's instead "the user that is looking at the app", that has a very different feel and it's good to make the contrast explicit.

I agree it's not helpful for "is admin", and probably most other possible names.

src/message/messageActionSheet.js
```diff
@@ -217,8 +226,12 @@ export const constructMessageActionButtons = ({
...
   }
-  if (!isAnOutboxMessage(message) && !isTopicNarrow(narrow) && !isPrivateNarrow(narrow)) {
-    buttons.push('reply');
+  if (!isAnOutboxMessage(message) && !isTopicNarrow(narrow) && !isPrivateOrGroupNarrow(narrow)) {
+    if (isAnnouncementOnly && !isAdmin) {
+      buttons.push('narrowToTopic');
+    } else {
+      buttons.push('reply');
+    }
   }
```
gnprice: This is too many conditionals in one place :slightly_smiling_face: -- it's more than the reader, in particular the author, can reasonably keep track of in their head.

One symptom of that is that this causes "Narrow to topic" to appear only for non-outbox messages. That doesn't make sense.

This also leaves "Reply" in place when the message is in an announcement-only stream, any time you're in a narrow that isn't specific to the stream. You could be in the all-messages narrow; or starred messages, @-mentions, or a search.

src/webview/webViewEventHandlers.js
```diff
@@ -180,11 +181,18 @@ const handleLongPress = (
...
+    backgroundData,
+    narrow,
+    isAnnouncementOnly,
+  } = props;
   showActionSheet(target === 'header', dispatch, showActionSheetWithOptions, _, {
     backgroundData,
     message,
     narrow,
+    isAnnouncementOnly,
   });
```
gnprice: One bonus of using `backgroundData` for the plumbing: it saves us from making a longer and longer list here of things we just pass through without change. (That was part of the origin of the `backgroundData` object in this message-list code.)

gnprice:
Reverted that commit, as 8f58f462d.

PR #3817, opened 2020-01-16
Fixed Gallery/Camera permissions

### LINKED ISSUE #3787: First attempt to open camers/gallery fails
On the first run, when we open camera/photos to send an image:

**Expected Behaviour:**
* Ask for permission
* Open Camera or Gallery

**Actual Behaviour:**
* Asks for permission
* Nothing happens after that. Need to tap camera again.

rheo-chiti:
This fixes #3787

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -17,6 +17,7 @@
         tools:node="remove"/>
     <uses-permission android:name="android.permission.CAMERA" android:required="false" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
     <uses-feature android:name="android.hardware.touchscreen" android:required="false" />
 
     <application
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { Platform, StyleSheet, View } from 'react-native';
+import { Platform, StyleSheet, View, PermissionsAndroid } from 'react-native';
 import type { DocumentPickerResponse } from 'react-native-document-picker';
 // $FlowFixMe
 import ImagePicker from 'react-native-image-picker';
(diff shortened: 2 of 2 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -17,6 +17,7 @@
         tools:node="remove"/>
     <uses-permission android:name="android.permission.CAMERA" android:required="false" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
```
ray-kraesig: Why is this needed now (when apparently it wasn't previously)?

src/compose/ComposeMenu.js
```diff
@@ -51,6 +51,13 @@ export const chooseUploadImageFilename = (uri: string, fileName: string | void):
   return name;
 };
 
+async function requestCameraPermission() {
+  await PermissionsAndroid.requestMultiple([
+    PermissionsAndroid.PERMISSIONS.CAMERA,
+    PermissionsAndroid.PERMISSIONS.READ_EXTERNAL_STORAGE,
+  ]);
```
ray-kraesig: What happens when this is called on iOS?

src/compose/ComposeMenu.js
```diff
@@ -100,7 +108,7 @@ class ComposeMenu extends PureComponent<Props> {
         waitUntilSaved: true,
       },
     };
-
+    requestCameraPermission();
```
ray-kraesig: Is `READ_EXTERNAL_STORAGE` really necessary for just the camera? We shouldn't ask for it in this flow if it's not.

(Same with `CAMERA` in the `handleImagePicker` flow.)

src/compose/ComposeMenu.js
```diff
@@ -80,6 +87,7 @@ class ComposeMenu extends PureComponent<Props> {
   };
 
   handleImagePicker = () => {
+    requestCameraPermission();
```
ray-kraesig: This looks like it will launch the permission request dialog asynchronously, and then immediately start the image picker. What happens if the image picker launch completes (presumably with a failure) before the user does anything with the permission request dialog?

android/app/src/main/AndroidManifest.xml
rheo-chiti: Earlier it wasn't needed because React Native automatically called it for the first time,but once declined it has to be called manually.

src/compose/ComposeMenu.js
rheo-chiti: Even if we don't call ```READ_EXTERNAL_STORAGE```  app automatically asks for it on pressing the second time,it's unusual.

src/compose/ComposeMenu.js
rheo-chiti: I haven't tested that yet, should I put that in try/catch block?

ray-kraesig:
You'll need to make and push changes before I can do another review.

(In particular, I'm not sure what you propose to do with a try/catch block here.)

PR #3818, opened 2020-01-16
ComposeMenu.js: Handle null for fileName in response from ImagePicker.

### LINKED ISSUE #3813: [iOS]: Crash on adding photo from camera if “Access to photos” permission not granted
I’ll fill this in with more detail tomorrow, but on iOS it seems that we have a crash when trying to add photos from the **camera** (not the media library!) when read access to photos (i.e., the media library) is not granted. Seen on iPhone 6s with iOS 13.3 on v26.20.143.

chrisbobbe:
Fixes #3813.

### DIFF
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -2,8 +2,7 @@
 import React, { PureComponent } from 'react';
 import { Platform, StyleSheet, View } from 'react-native';
 import type { DocumentPickerResponse } from 'react-native-document-picker';
-// $FlowFixMe
-import ImagePicker from 'react-native-image-picker';
+import * as ImagePicker from 'react-native-image-picker';
 
 import type { Dispatch, Narrow } from '../types';
 import { connect } from '../react-redux';
@@ -57,14 +56,16 @@ class ComposeMenu extends PureComponent<Props> {
     dispatch(uploadFile(destinationNarrow, uri, chooseUploadImageFilename(uri, fileName)));
   };
 
-  handleImagePickerResponse = (response: {
-    didCancel: boolean,
-    // Upstream docs are vague:
-    // https://github.com/react-native-community/react-native-image-picker/blob/master/docs/Reference.md
-    error?: string | void | null | false,
-    uri: string,
-    fileName: string,
-  }) => {
+  handleImagePickerResponse = (
+    response: $ReadOnly<{
+      didCancel: boolean,
+      // Upstream docs are vague:
+      // https://github.com/react-native-community/react-native-image-picker/blob/master/docs/Reference.md
+      error?: string | void | null | false,
+      uri: string,
+      fileName: string | void,
+    }>,
+  ) => {
     if (response.didCancel) {
       return;
     }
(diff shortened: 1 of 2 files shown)


src/compose/ComposeMenu.js
```diff
@@ -63,7 +63,7 @@ class ComposeMenu extends PureComponent<Props> {
     // https://github.com/react-native-community/react-native-image-picker/blob/master/docs/Reference.md
     error?: string | void | null | false,
     uri: string,
-    fileName: string,
+    fileName: ?string,
```
ray-kraesig: This shouldn't have typechecked anyway – this isn't a supertype of `Response`!

Flow isn't catching this because of the `$FlowFixMe` up top: `ImagePicker` is defined as `any`. Changing the import to `import * as ImagePicker` (and removing the suppression comment) lets Flow see the definitions.

It seems that either import method works, at least on Android. Which is surprising, to say the least – but perhaps there's some React linkage magic going on. Does it work both ways on iOS?

chrisbobbe: Hmm, I think I see what you mean -- using `import * as ImagePicker` reveals that the callback we're passing, `this.handleImagePickerReponse`, isn't accepted by Flow because of the types specified by react-native-image-picker (specifically, .launchCamera expects to pass a Response to that callback, which doesn't include fileName being null, and ours does). I remember being vaguely surprised that no Flow errors were introduced when I made this change, and that I didn't have to add any suppressions. I should probably have at least made a note of my surprise there. Turns out it's because of this very general suppression, as you point out.

I wonder if we can avoid doing such a general suppression. If you get a moment, would you like to go over this in person? I've poked around a bit here and I'm not sure how best to use a modified Response type and a modified type for ImagePicker.launchCamera without making a local copy of react-native-image-picker, which would complicate staying up to date with newer versions. (Maybe we'd just need to make that local copy anyway.)

chrisbobbe:
OK, I've updated the pull request to reflect our chat earlier.

ray-kraesig:
Haven't tested yet, but looks good codewise.

* The change itself needs an inline comment, lest someone come along one day and unknowingly undo all your hard work by tightening up the the callback's type to match the `flow-typed` declarations.
* When describing a change in the commit message body, it should also use the *to*-less infinitive, to match the summary.
* The line lengths in your commit summaries are a bit long – one of them is 81 characters. [Our standard](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#message-body) is 70.

chrisbobbe:
> The line lengths in your commit summaries are a bit long – one of them is 81 characters. Our standard is 70.

Indeed, and I [should know](https://github.com/zulip/zulip/commit/04e6c3bd18e17fbfb1f59e6aedc060fa7f82e957#diff-9a4e02aa1eebc2316c73dc4724082323) 🤣. I have been meaning to configure Nano to do this for me automatically, but I wasn't sure what any available shortcuts might be to ease enforcing different limits on the summary and the body; I'll ping you or Greg tomorrow on this, and address your other comments.

Thank you!

chrisbobbe:
OK, this should be ready for another review, when convenient. 🙂

ray-kraesig:
:heavy_check_mark:

chrisbobbe:

PR #3821, opened 2020-01-17
compose box: enable Send button in PMs

### LINKED ISSUE #3820: Send button in private messages is disabled
It looks like https://github.com/zulip/zulip-mobile/commit/a2d6793aeac19259c205a175048f4017bf89ddcb was made to fix https://github.com/zulip/zulip-mobile/pull/3748, but it introduced the side effect of preventing people from sending private messages! Private messages don't have topics, so the added conditional disables the send button.

ray-kraesig:
Inadvertently broken in a2d6793aeac19259c205a175048f4017bf89ddcb.

Fixes #3820.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -337,6 +337,10 @@ class ComposeBox extends PureComponent<Props, State> {
       paddingBottom: safeAreaInsets.bottom,
       backgroundColor: 'hsla(0, 0%, 50%, 0.1)',
     };
+    const hasTopic: boolean = this.getCanSelectTopic();
+    // Empty messages are never sendable. Messages with empty topics are
+    // sendable only if there shouldn't be a topic anyway.
+    const isSendable: boolean = !!(message.trim() && (topic.trim() || !hasTopic));
 
     return (
       <View>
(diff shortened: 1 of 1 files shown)


ray-kraesig:
Pinging @gnprice.

gnprice:
Thanks!

Is this quite the right behavior? `getCanSelectTopic` is specifically meant to answer the question "should we show the topic input box", and it has some wrinkles that don't seem like they belong in the answer to "should the 'send' button be enabled". In particular `state.isFocused`, and I'm not sure about the `editMessage` wrinkle either.

Let's revert a2d6793ae / #3748 to take the urgency out (as you mentioned a bit ago in person), and then we can follow up by sorting out the exact logic we should have here.

ray-kraesig:
> [...] and I'm not sure about the `editMessage` wrinkle either.

On review, I'm not sure the `editMessage` logic is even correct for the existing use case! It prevents you from editing a message's topic if you're viewing it in a topic narrow – which I think *is* something you might reasonably want to do, to spin a tangent off from an existing discussion.

That aside, as I mentioned offline, I'm now leaning toward the belief that "the correct topic for a PM is `undefined`, rather than `""`", which should make #3820-like bugs significantly less likely.

gnprice:
> On review, I'm not sure the `editMessage` logic is even correct for the existing use case!

In general the `editMessage` logic around the compose box is all pretty messy :-/, so that wouldn't be surprising.


> It prevents you from editing a message's topic if you're viewing it in a topic narrow – which I think _is_ something you might reasonably want to do, to spin a tangent off from an existing discussion.

Yeah, I agree this is a pretty reasonable thing to want to do. The one thing I'd want to check if removing that condition is that the app behaves in a reasonable way when you do it -- e.g. it might be pretty confusing as the message disappears. Though that's I guess what the webapp does in this case.


> "the correct topic for a PM is `undefined`, rather than `""`", which should make #3820-like bugs significantly less likely.

I agree.

PR #3822, opened 2020-01-17
redux: Discard messages/narrows on REALM_INIT, not DEAD_QUEUE.

### LINKED ISSUE #3802: "No messages" when in fact we're loading messages
This is perhaps the most conspicuous bug I regularly see at present when using the app. I see it on v26.20.143, and I think I've been seeing it on previous versions for weeks at least, perhaps longer.

The typical sequence is:
* You had the app open, and had been looking at some narrow.
* You go do something else, and come back to the app after a while -- like over 10 min. (The time for an event queue to expire.)
* You re-open the app; it was still in the background, so still has its navigation state and shows you that narrow.
* But for a few seconds, the message list in the narrow says "No messages! Why not start the conversation?", before switching to the "loading" animation.

I think basically what's going on is:
* We try to poll the event queue, and learn it's expired.
* OK, so we go and set up a new event queue, which will come with a new snapshot of all the data.
* Along with that, we have to discard all our existing message data as stale -- any of those messages could have been edited, deleted, had reactions added or removed, etc., since we last polled the queue, and we wouldn't know.
* But we do that in a way that leaves us acting as if we knew there were *no* messages in those narrows, etc. This might be as simple as an ordering issue; e.g. perhaps we shouldn't discard the message data at `DEAD_QUEUE` (when we learn the event queue has expired), but at `REALM_INIT` (kind of misnamed; dispatched when the new `/register` completes.)

Some history (which I haven't fully reread and absorbed) for cross-reference:
* #3162 was a previous bug with some similar symptoms -- much worse, though, as it doesn't just fix itself after a few seconds
* #3151 was meant to address that; the PR is open, but the #3162 part of it was merged
* #3284 was similar to the current bug, though a different repro sequence (involving a notification)
* #3319 described the symptoms in detail (though again the repro involves a notification)

See also [chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/MessageReactionList.20absent.20message.20prop/near/809132). There's an issue affecting MessageReactionList, where `state.messages` has no message with the expected message ID, which potentially is how the same underlying issue shows up if in step 1 you were looking at a reaction list rather than a message list.

chrisbobbe:
Fixes: #3802.

When returning to the app after the event queue has expired, there's
a moment after stale data is purged when the app incorrectly
believes that an empty message list is the up-to-date state, and "No
messages! Why not start the conversation?" displays until fresh data
arrives.

Currently, state.messages and state.narrows are emptied on the
DEAD_QUEUE action, which is dispatched before the /register request.
Instead, preserve the stale data until /register completes, clearing
it in response to REALM_INIT.

In doInitialFetch, fetchTopMostNarrow, fetchPrivateMessages, and
(sometimes) fetchMessagesInNarrow are also dispatched after
/register to ensure the stale data is refreshed. All of these
already handle their loading states correctly, on a per-narrow basis
in state.fetching.

Clearing the messages in MESSAGE_FETCH_START instead of REALM_INIT
was considered but rejected because MESSAGE_FETCH_START is
dispatched before the /register request.

### DIFF
--- a/src/caughtup/caughtUpReducer.js
+++ b/src/caughtup/caughtUpReducer.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import type { CaughtUp, CaughtUpState, Action } from '../types';
 import {
-  DEAD_QUEUE,
+  REALM_INIT,
   LOGOUT,
   LOGIN_SUCCESS,
   ACCOUNT_SWITCH,
@@ -56,7 +56,7 @@ const legacyInferCaughtUp = (prevCaughtUp: CaughtUp | void, action) => {
 
 export default (state: CaughtUpState = initialState, action: Action): CaughtUpState => {
   switch (action.type) {
-    case DEAD_QUEUE:
+    case REALM_INIT:
     case LOGOUT:
     case LOGIN_SUCCESS:
     case ACCOUNT_SWITCH:
--- a/src/chat/Chat.js
+++ b/src/chat/Chat.js
@@ -11,10 +11,12 @@ import ComposeBox from '../compose/ComposeBox';
 import UnreadNotice from './UnreadNotice';
 import styles from '../styles';
 import { canSendToNarrow } from '../utils/narrow';
+import { getLoading } from '../directSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow } from './narrowsSelectors';
 
 type SelectorProps = {|
+  loading: boolean,
   fetching: Fetching,
   haveNoMessages: boolean,
 |};
@@ -37,9 +39,9 @@ const componentStyles = StyleSheet.create({
 
 class Chat extends PureComponent<Props> {
   render() {
-    const { fetching, haveNoMessages, narrow } = this.props;
+    const { fetching, loading, haveNoMessages, narrow } = this.props;
 
-    const isFetching = fetching.older || fetching.newer;
+    const isFetching = fetching.older || fetching.newer || loading;
     const showMessagePlaceholders = haveNoMessages && isFetching;
     const sayNoMessages = haveNoMessages && !isFetching;
     const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
@@ -59,6 +61,7 @@ class Chat extends PureComponent<Props> {
 }
 
 export default connect<SelectorProps, _, _>((state, props) => ({
+  loading: getLoading(state),
   fetching: getFetchingForNarrow(state, props.narrow),
   haveNoMessages: getShownMessagesForNarrow(state, props.narrow).length === 0,
 }))(Chat);
--- a/src/chat/fetchingReducer.js
+++ b/src/chat/fetchingReducer.js
@@ -1,7 +1,6 @@
 /* @flow strict-local */
 import type { FetchingState, Action } from '../types';
 import {
-  DEAD_QUEUE,
   LOGOUT,
   LOGIN_SUCCESS,
   ACCOUNT_SWITCH,
@@ -42,7 +41,6 @@ const messageFetchComplete = (state, action) => {
 
 export default (state: FetchingState = initialState, action: Action): FetchingState => {
   switch (action.type) {
-    case DEAD_QUEUE:
     case LOGOUT:
     case LOGIN_SUCCESS:
     case DO_NARROW:
--- a/src/chat/flagsReducer.js
+++ b/src/chat/flagsReducer.js
@@ -6,6 +6,7 @@ import {
   EVENT_NEW_MESSAGE,
   EVENT_UPDATE_MESSAGE_FLAGS,
   LOGOUT,
+  LOGIN_SUCCESS,
   ACCOUNT_SWITCH,
 } from '../actionConstants';
 import { deeperMerge } from '../utils/misc';
@@ -100,6 +101,7 @@ export default (state: FlagsState = initialState, action: Action): FlagsState =>
   switch (action.type) {
     case REALM_INIT:
     case LOGOUT:
+    case LOGIN_SUCCESS:
     case ACCOUNT_SWITCH:
       return initialState;
 
--- a/src/chat/flagsReducer.js
+++ b/src/chat/flagsReducer.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import type { Action, FlagsState, Message } from '../types';
 import {
-  DEAD_QUEUE,
+  REALM_INIT,
   MESSAGE_FETCH_COMPLETE,
   EVENT_NEW_MESSAGE,
   EVENT_UPDATE_MESSAGE_FLAGS,
@@ -98,7 +98,7 @@ const eventUpdateMessageFlags = (state, action) => {
 
 export default (state: FlagsState = initialState, action: Action): FlagsState => {
   switch (action.type) {
-    case DEAD_QUEUE:
+    case REALM_INIT:
     case LOGOUT:
     case ACCOUNT_SWITCH:
       return initialState;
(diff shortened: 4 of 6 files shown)


chrisbobbe:
This is WIP because, while my manual testing hasn't revealed any unintended side effects, this touches some central logic, and I have a few lingering questions about redux-persist (to be posted shortly).

chrisbobbe:
Specifically, from reading the code, it seems like redux-persist is used to rehydrate the Redux store with data from AsyncStorage, right? Are `state.messages` and/or `state.narrows` persisted in this way?

I'm not sure what the order is of the REHYDRATE action and our other actions, including the ones in doInitialFetch. 42adfb34810468f1a011086c9480093aa768a013 describes createActionBuffer as a way to postpone other actions so REHYDRATE always happens first, but I'm not sure that's what it does. From the [doc](https://github.com/rt2zz/redux-action-buffer):

> A middleware for redux that buffers all actions into a queue until a breaker condition is met, at which point the queue is released (i.e. actions are triggered).

Our breaker condition is the REHYDRATE action, but from that description, it doesn't seem like there's any reordering of the queue (e.g., for our case, to put REHYDRATE at the front), only that the queue doesn't start emptying until the breaker condition is met.

Does this seem right, and if so, might it matter (either in general or for this PR)?

chrisbobbe:
(Just made some edits to the commit message.)

gnprice:
> Are `state.messages` and/or `state.narrows` persisted in this way, and if so, does that matter?
> 
> I'm not sure what the order is of the REHYDRATE action and our other actions, [...]

We discussed these questions yesterday in person, and I think I answered them then. I've also just sent #3907 which adds a bunch of jsdoc and other comments to `store.js`, which hopefully will help make these (and many other questions) easy for the next person to find the answers to in the future.

chrisbobbe:
Ah, we were writing comments at the same time! 🙂 I'll post mine anyway, just to make sure I've understood correctly.

> Does this seem right, and if so, might it matter (either in general or for this PR)?

This makes more sense, after discussing in person. The REHYDRATE action marks the completion of retrieving the state from ZulipAsyncStorage, for the whitelisted fields specified in storeKeys and cacheKeys. It's clear from reading the source code that all of the actions that would have preceded REHYDRATE are entered into a queue, which is released _after_ REHYDRATE (not before).

In fact, I may have been a little too harsh on the docs yesterday — it turns out that the "breaker condition" can be anything, not just a Redux action. The fact that it happens to be an action for us made me think, ok, here's a library intended to manage a homogeneous queue of Redux actions, but it just likes to break the definition of a queue, to suit some purpose, by reordering a particular action to the front.

chrisbobbe:
OK, I've read and merged PR https://github.com/zulip/zulip-mobile/pull/3900. I also read PR https://github.com/zulip/zulip-mobile/pull/3907.

gnprice:
Thanks @chrisbobbe!

I just reread this PR, in light of #3900. (Which, for self-containedness of this thread, was this:
> I sat down this afternoon to look at #3822 and #3802, and found I wasn't 100% sure under what conditions we show that "No Messages" message, and that it wasn't immediate to find out.
>
> I dug into the code to answer that question, and this series of changes is the result. :slightly_smiling_face:

) The main question in my mind was: after this PR, when you open the app with an expired event queue, what's the sequence of changes that would affect whether we show "No messages"? And should we expect that there will still be a stage at which we do show that?

I think the answer is: this change narrows that window by a lot, but doesn't quite close it; and there's an additional tweak we can do that will then close it.

Specifically, this is what now decides when that's shown (in `Chat.js`):
```js
    const isFetching = fetching.older || fetching.newer;
    const showMessagePlaceholders = haveNoMessages && isFetching;
    const sayNoMessages = haveNoMessages && !isFetching;
```
where `fetching` is the per-narrow value from `state.fetching`, and `haveNoMessages` is `msgs.length === 0` where `msgs` is the per-narrow array of messages we would show in the message list for this narrow.

The issue this PR is aimed at is basically that we're making `sayNoMessages` true when we shouldn't.

The PR works by delaying the point when `haveNoMessages` becomes true. In the sequence in `doInitialFetch`:
```js
export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetState) => {
  // ...

  dispatch(initialFetchStart());

  // ...
    initData = await tryUntilSuccessful(() =>
      api.registerForEvents(auth, {
        fetch_event_types: config.serverDataOnStartup,
        apply_markdown: true,
        include_subscribers: false,
        client_gravatar: true,
        client_capabilities: {
          notification_settings_null: true,
        },
      }),
    );

  dispatch(realmInit(initData));
  dispatch(fetchTopMostNarrow());
  dispatch(initialFetchComplete());
```

it moves to the `realmInit` step, whereas in master it's happening just before `doInitialFetch` is even called.

Just after the `realmInit` step, I'd expect this would still leave us saying "No messages" -- at least if `Chat#render` happens to get called at that moment. We're clearing out the `narrows` and `messages` state trees, so `haveNoMessages` will be true; but `isFetching` will still be false. It will be made true by `fetchTopMostNarrow`, which will start a fetch for the narrow currently shown on screen.

Hmm, though perhaps it's not actually possible for `render` to get called in between those! I think `realmInit` is all synchronous, and so is the parts of `fetchTopMostNarrow` up through dispatching the `messageFetchStart` event that causes `isFetching` to become true. In principle, all those `dispatch` calls give Redux an opportunity to run any code it likes, including a `react-redux` listener that in turn causes React to call `render`... but it's possible that Redux or `react-redux` makes a guarantee that won't happen. In which case this does already completely fix the issue.

Still, even if that turns out to be the case, the whole chain of reasoning feels a bit fragile -- dependent on fine details of the sequence of calls in `doInitialFetch` and of the exact places `await` appears in the definitions of its various helpers.

I think a good fix would be to augment the definition of `isFetching` in that `Chat.js` logic quoted above, so that it also considers `state.session.loading`, using `getLoading`. That's the flag which is set by `initialFetchStart` and cleared by `initialFetchComplete`; it basically covers "we're doing a fresh /register call, loading a fresh copy of the overall app data". I think it makes semantic sense that if we're in the middle of that and we don't have any messages, we should emphasize "we're loading messages" rather than "we have no messages" -- which is exactly the effect of `showMessagePlaceholders` vs. `sayNoMessages` in that `Chat#render` code.

gnprice:
A separate (simpler :slightly_smiling_face: ) question I looked at this to answer: is this acting together on all the data that's closely related and we want to act together on?

So I went and looked at `reduxTypes.js`, which has some useful cross-references.
```js
 * These `Message` objects lack the `flags` property; we store that
 * information separately, as `FlagsState`.
 *
 * See also `NarrowsState`, which is an index on this data that identifies
 * messages belonging to a given narrow.
 */
export type MessagesState = {|
// ...

 * See also:
 *  * `MessagesState`, which stores the message data indexed by ID;
 *  * `CaughtUpState` for information about where this data *is*
 *    complete for a given narrow;
 *  * `FetchingState` for information about which narrows we're actively
 *    fetching more messages from.
 */
export type NarrowsState = {
```

So I think at least `FlagsState` i.e. `state.flags` should be tracked the same way as `MessagesState` i.e. `state.messages`. ... Looks like there's already an unrelated discrepancy there, in the handling of `LOGIN_SUCCESS`. But in any case `flagsReducer.js` should get the same update in this diff as `messagesReducer.js` does.

And then also `caughtUp` should get cleared when `narrows` is.

As for `fetching`... I'm not convinced it should properly be cleared on either `DEAD_QUEUE` or `REALM_INIT`! Neither one actually causes us to stop an ongoing fetch. So probably delete that `case` line from `fetchingReducer`.

chrisbobbe:
All good points, thanks for explaining all of that!! I've made changes to address these things.

gnprice:
And merged -- thanks!

Issue #3806, opened 2020-01-15
Feature request: delete topics


andreasoc:
Some users in my Company use only the APP and the possibility to delete or edit the topics is necessary.
thanks

ray-kraesig:
Editing topics can be done already: long-press on any message. (The interface could definitely be improved, though.)

Deleting topics... isn't something we support at all, as far as I'm aware. Do you just mean "moving all the messages in one topic into another existing topic"?

timabbott:
We do support that in the webapp these days:

![image](https://user-images.githubusercontent.com/2746074/72487751-db8fb480-37c3-11ea-8695-fe57825831df.png)

The API endpoint is `streams/(?P<stream_id>\d+)/delete_topic`; not part of the documentation yet.

andreasoc:
> Editing topics can be done already: long-press on any message. (The interface could definitely be improved, though.)
> 
> Deleting topics... isn't something we support even on the web interface, as far as I'm aware. Do you mean "moving all the messages in one topic into another existing topic"?

I confirm that the topic deletion is present in the desktop app, but not in the mobile app

agrawal-d:
I have made a PR that implements this feature : https://github.com/zulip/zulip-mobile/pull/3824

ray-kraesig:
> I have made a PR that implements this feature : #3824

For future reference: there's no need to post this sort of thing in the GitHub issue, so long as you've used the magic words in the PR itself (which you have). GitHub's automatic backlink suffices.

PR #3824, opened 2020-01-18
action sheet: add and implement delete topic option

agrawal-d:
Closes https://github.com/zulip/zulip-mobile/issues/3806

delete-topic

image


agrawal-d:
@gnprice I need to be an administrator to test this. Since you are an administrator, can you please test this PR and let me know if it works? Thanks!

agrawal-d:
Also, there is a small issue: In line `messageActionSheet.js:88`, flow assumes `message` is of type `Outbox` even though the definition is `message: Message | Outbox`. This causes a flow error.

Is this fixable, or is thus a flow bug?

agrawal-d:
Only admins can do this. Works on my test realm `https://divyanshutest.zulipchat.com/`

agrawal-d:
@ray-kraesig  @gnprice please review.

static/translations/messages_en.json
```diff
@@ -50,6 +50,7 @@
   "Copy to clipboard": "Copy to clipboard",
   "Link copied to clipboard": "Link copied to clipboard",
   "Mute topic": "Mute topic",
+  "Delete topic":"Delete topic",
```
ray-kraesig: Space after the colon, for consistency.

src/message/messageActionSheet.js
```diff
@@ -79,6 +79,26 @@ const muteTopic = ({ auth, message }) => {
 };
 muteTopic.title = 'Mute topic';
 
+const deleteTopic = ({ auth, message }) => {
+  Alert.alert(
+    `Are you sure you want to delete the topic '${message.subject}' ?`,
```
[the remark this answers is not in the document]
agrawal-d: I'll wait for @gnprice 's response, before pushing any changes.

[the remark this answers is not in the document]
ray-kraesig: Erratum: this JS construct can't be localized, but the i18n library we have can help.

```js
  const alertTitle = _.intl.formatMessage(
    { id: "Are you sure you want to delete the topic '{topic}'?" },
    { topic: message.subject },
  );
```

agrawal-d:
@ray-kraesig I'm not sure how to proceed with this, with respect to the flow issue and handling the corner cases you mentioned https://github.com/zulip/zulip-mobile/pull/3824#pullrequestreview-345587020.
Can you please tell me in what I should do next to resolve that? Thanks!

ray-kraesig:
It's not really a Flow issue; it's a functional issue that Flow is recognizing. `message` may actually be an `Outbox`; that case will need to be tested for and handled separately, much like the delete-message callback does. 

However, the delete-message action isn't a great template here, as it has the bug I mentioned: if a user long-presses a message before it's sent, then the delete-message callback will still attempt to delete it from the local Outbox (from which it's already been removed), rather than from the server, and no error will be given.

Resolving that (for either delete-message or delete-topic) will require some thought and some testing.

src/topics/topicActions.js
```diff
@@ -34,3 +35,19 @@ export const fetchTopicsForActiveStream = (narrow: Narrow) => async (
   }
   dispatch(fetchTopics(stream.stream_id));
 };
+
+export const deleteMessagesForTopic = (narrow: Narrow) => async (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
+  const messages: $ReadOnlyArray<Message | Outbox> = getMessagesForNarrow(getState(), narrow);
```
agrawal-d: @ray-kraesig can you please help me out here? `getMessagesForNarrow()` is returning an empty array here, even though the narrow does contain several messages.
Output of the console statement: 
image

Am I using the selector incorrectly?

agrawal-d: Posted a discussion on CZO here: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3824/near/815558

agrawal-d:

src/message/messageActionSheet.js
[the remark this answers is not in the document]
agrawal-d: The value of `id` is being taken literally, without substitution for topic. Using `defaultMessage` appears to work.

I referred to https://github.com/formatjs/react-intl/blob/master/docs/API.md#formatmessage

agrawal-d:
Tests are failing, I will update them if the current change is acceptable.

andreasoc:
sigh

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,41 @@ const muteTopic = async ({ auth, message }) => {
...
+  const alertTitle = _.intl.formatMessage(
+    {
+      id: 'delete_topic_warning',
+      defaultMessage: "Are you sure you want to delete the topic '{topic}' ?",
+    },
+    { topic: message.subject },
+  );
+  Alert.alert(
+    alertTitle,
+    'This will also delete all messages in the topic.',
```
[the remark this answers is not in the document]
agrawal-d: Done

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,41 @@ const muteTopic = async ({ auth, message }) => {
 muteTopic.title = 'Mute topic';
 muteTopic.errorMessage = 'Failed to mute topic';
 
+const deleteTopic = async ({ auth, message, dispatch, ownEmail, _ }) => {
+  const alertTitle = _.intl.formatMessage(
+    {
+      id: 'delete_topic_warning',
+      defaultMessage: "Are you sure you want to delete the topic '{topic}' ?",
+    },
```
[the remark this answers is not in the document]
agrawal-d: Changed.

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,41 @@ const muteTopic = async ({ auth, message }) => {
...
+    { topic: message.subject },
+  );
+  Alert.alert(
+    alertTitle,
+    'This will also delete all messages in the topic.',
+    [
+      {
+        text: 'Delete topic',
```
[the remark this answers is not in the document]
agrawal-d: Done

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,41 @@ const muteTopic = async ({ auth, message }) => {
...
+        onPress: async () => {
+          dispatch(deleteMessagesForTopic(getNarrowFromMessage(message, ownEmail), message)).catch(
+            err => {
+              Alert.alert(`${_(deleteTopic.errorMessage)} '${message.subject}'`, err.message);
```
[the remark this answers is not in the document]
agrawal-d: Fixed.

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,41 @@ const muteTopic = async ({ auth, message }) => {
...
+            },
+          );
+        },
+        style: 'destructive',
+      },
+      {
+        text: 'Cancel',
```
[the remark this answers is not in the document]
agrawal-d: Done.

agrawal-d:
@ray-kraesig I've pushed some changes. I've also created a new test for the 'Delete toipc' button.

agrawal-d:
>Please ensure tests pass before resubmitting for review.

Should tests always pass before requesting a review?
Is it ever okay if tests are not passing when requesting an initial review? I'm asking because the initial review may result in significant changes requiring the tests to be changed again.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -147,11 +162,43 @@ describe('constructHeaderActionButtons', () => {
...
     const buttons = constructHeaderActionButtons({
-      backgroundData: { mute, subscriptions },
+      backgroundData: { mute, subscriptions, ownUser },
       message,
     });
 
     expect(buttons).toContain('muteStream');
   });
+
+  test('show delete topic option if current user is an admin', () => {
```
ray-kraesig: A test for the case where the user is not an admin is also needed.

src/message/messageActionSheet.js
```diff
@@ -90,6 +91,46 @@ const muteTopic = async ({ auth, message }) => {
...
+            text: _('Cancel'),
+            onPress: () => {
+              resolve('ok');
+            },
+            style: 'cancel',
+          },
+        ],
+        { cancelable: true },
+      );
+    });
```
ray-kraesig: If instead you do nothing but `resolve(true)` and `resolve(false)` in the two `onPress` handlers, you can move all the application logic out of the interior by testing the return value of `await AsyncAlert`. This should result in nearly-symmetric branches in the `Alert.alert` call, more linear control flow for the whole function, and exception forwarding essentially for free.

(Also, `AsyncAlert` could use a type annotation.)

ray-kraesig:
> Should tests always pass before requesting a review?

Almost always, yes. (Although see below.)

> Is it ever okay if tests are not passing when requesting an initial review? I'm asking because the initial review may result in significant changes requiring the tests to be changed again.

That is sometimes the case! However, when it is, the review request should always be accompanied with an explanation of why the tests don't pass, and what questions you need to have answered in order to fix them. Otherwise, I'm probably going to infer that you're still working on the PR, and that it's not actually ready for review.

Putting off fixing the tests until you get a first review may be appropriate when those fixes would be a significant and potentially wasteful undertaking in themselves. (Although one would definitely still need a comment stating and explaining this.) However, in this case:
* Fixing the tests is straightforward – not quite as straightforward as it should be (cleanup is badly needed!), but still, not terribly intrusive.
* Essentially any implementation of a `Delete topic` button-insertion test is going to need to know whether the user is an admin. Even a total rewrite would be very unlikely to change that.

src/message/__tests__/messageActionSheet-test.js
agrawal-d: Done

src/message/messageActionSheet.js
agrawal-d: Done!

agrawal-d:
@ray-kraesig this is ready for a re-review.

gnprice:
Hi @agrawal-d !

In `src/message/__tests__/messageActionSheet-test.js`, I notice it's one of our test files that isn't yet type-checked. Since we're making significant changes to it in this PR, let's have an earlier part of the PR get it passing the type-checker -- and I think some of the same changes that help with that will also help make the new tests nice and shorter and simpler to read.

There's some discussion of this in our docs on testing:
https://github.com/zulip/zulip-mobile/blob/master/docs/howto/testing.md#unit-tests-js
We also happened to have a [chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Writing.20Tests/near/820706) about it just today, after you asked about test-writing strategy in the context of a different PR :slightly_smiling_face:

agrawal-d:
@gnprice I'm having some trouble converting `messageActionSheet-test.js` to a type checked one.

1. I added `// @flow strict-local` to the top.
2. This causes the following error in several places:

```
For:
const store = mockStore({ ...

Cannot call `mockStore` with object literal bound to `middlewares` because  object literal [1] is incompatible with  array type [2].Flow(InferError)
messageActions-test.js(14, 31): [1] object literal
redux-mock-store_v1.2.x.js(26, 42): [2] array type
```

I think flow is not not using our mockStore defined in `redux-mock-store.js` and instead usinng `redux-mock-store_v1.2.x.js` to infer the type - moving the contents of `redux-mock-store.js` into `messageActions-test.js` somehow fixes this error.

3. Assuimg I fixed the error in step 2, I get several more errors of these two types:
```
For :
store.dispatch(doNarrow(streamNarrowObj));

Cannot call `store.dispatch` with `doNarrow(...)` bound to `action` because property `type` is missing in  function [1] but exists in  object type [2].Flow(InferError)
messagesActions.js(24, 83): [1] function
redux-mock-store_v1.2.x.js(14, 28): [2] object type
```

```
For 
expect(actions[0].type).toBe('DO_NARROW');

Cannot get `actions[0].type` because property `type` is missing in  function [1].Flow(InferError)
```

I'm can't think of how resolve these :'(

agrawal-d:
@ray-kraesig I've removed the `async` keyword in the latest force-push.

ray-kraesig:
> @gnprice I'm having some trouble converting `messageActionSheet-test.js` to a type checked one.

The lines of code quoted here appear to be from `messageActions-test.js`, not `messageActionSheet-test.js`. That file's a little out of scope for this PR, I think. :slightly_smiling_face:

agrawal-d:
> The lines of code quoted here appear to be from `messageActions-test.js`, not `messageActionSheet-test.js`. That file's a little out of scope for this PR, I think. slightly_smiling_face

Oh, this is hilarious! Thanks!

agrawal-d:
@gnprice @ray-kraesig I've made all the changes - the tests are type-checked now. Can you please review this? Thanks!

agrawal-d:
Not sure why the ordering of commits is different in GitHub UI, here's the output of
```git log --pretty=oneline --abbrev-commit```

```
0bf5b440 (HEAD -> delete-topic, agrawal-d/delete-topic) action sheet: Add delete topic feature.
4227e799 topic actions: Create action 'deleteMessagesForTopic'.
fc435354 api: Create API function for 'delete topic'.
acf3fb75 messageActionSheet [nfc]: Refactor tests to make them type-checked.
51be05fe exampleData: Add sample data of types 'BackgroundData' and 'Narrow'.
```

**Comment by chrisbobbe** [comment]
> Not sure why the ordering of commits is different in GitHub UI

Well spotted! 🙂 Greg explained this to me [here](https://github.com/zulip/zulip-mobile/pull/3619#issuecomment-532884971); it's a bug in GitHub where they're sorting by author date instead of committer date.

File: src/message/__tests__/messageActionSheet-test.js [hunk 11]
```diff
@@ -1,30 +1,31 @@
...
 describe('constructActionButtons', () => {
   const narrow = deepFreeze([]);
 
   test('show star message option if message is not starred', () => {
     const message = deepFreeze({
+      ...eg.streamMessage(),
       id: 3,
       reactions: [],
+      avatar_url: '',
+      client: '',
```
ray-kraesig: Since you're using `eg.streamMessage()` now, you can just do `const message = eg.streamMessage({ id: 3 })` here (and similarly elsewhere) – there's no need to explicitly specify any properties which aren't relevant to the test.

src/__tests__/exampleData.js
```diff
@@ -360,3 +368,44 @@ export const action = deepFreeze({
...
+    summarize_in_stream: {},
+    force_expand: {},
+    force_collapse: {},
+    has_alert_word: {},
+    historical: {},
+    is_me_message: {},
+  },
+  mute: [['announce', 'stream events']],
+  ownUser: selfUser,
+  subscriptions: [
```
ray-kraesig: *Idem*.

src/__tests__/exampleData.js
```diff
@@ -137,6 +138,13 @@ const displayRecipientFromUser = (user: User): PmRecipientUser => {
   });
 };
 
+export const narrow: Narrow = [
```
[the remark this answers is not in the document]
agrawal-d: You're right - thanks for the explanation!

agrawal-d:
@ray-kraesig thank you for the review. I have made the changes you requested. Please review it again.

agrawal-d:
>Greg explained this to me here; it's a bug in GitHub where they're sorting by author date instead of committer date.

Oh - that's interesting. Thank's @chrisbobbe !

src/topics/topicActions.js
```diff
@@ -34,3 +36,25 @@ export const fetchTopicsForActiveStream = (narrow: Narrow) => async (
   }
   dispatch(fetchTopics(stream.stream_id));
 };
+
+export const deleteMessagesForTopic = (narrow: Narrow, message: Message | Outbox) => async (
```
ray-kraesig: I should have noticed this during one of the preceding passes, but – why does this function take a `Message | Outbox`? All the information it needs is in the `Narrow`.

(For that matter, it would be fairly simple just to pass the stream name and topic as parameters – there's no need to wrap them up in a `Narrow` object which you then have to check, or assume, to be of the right shape.

`Narrow` *should* be a type we'd want here. Unfortunately it's really a server API type that we've pressed into service for insufficiently-closely-related purposes in the mobile app. :disappointed: We've been discussing its replacement for some time.)

ray-kraesig:
And merged. Thanks!

PR #3825, opened 2020-01-20
Lightbox: fix support for GIFs

### LINKED ISSUE #3497: Animated GIFs are static in lightbox
@ishammahajan reported that animated GIFs don't work in the lightbox, posting this video:
https://chat.zulip.org/user_uploads/2/b/ZwHl_biOscMtw27V8GeNyJAu/Screen-Recording-2019-03-21-at-2.11.20-AM.mov

My description of that video:
* The animated image (presumably a gif?) works fine in the message list.
* But on touching it to go to the lightbox, there's just black -- the image doesn't show at all.

On further investigation, Isham added:

> It does come up, the initial image, after about 15-20 seconds

(Which I think means that after that loading time, a still frame showed up, but it still wasn't animated.)

@borisyankov [posted a test anigif](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Gifs.20don't.20get.20displayed.20in.20mobile.20apps/near/722158), and
* Boris then reported that it worked for him, on both iOS and Android
* But when I tried just now, on Android in v24.0.113, it didn't work:
  * The animation works just fine in the message list -- same as in Isham's video.
  * But when I go to the lightbox, I get just a still frame.

So, more debugging to be done! I think the first question to answer is: why did it work for Boris and not for me? Or put another way: in what conditions does it work and not work?

agrawal-d:
Closes #3497.
GIFs weren't working on Android. This patch fixes that.

**Reason for changes**:
The earlier dependencies in build.gradle worked only for RN Version>0.60.
This configuration works for RN Version 0.59.

This is documented in the official docs for React Native 0.59 (https://facebook.github.io/react-native/docs/0.59/image#gif-and-webp-support-on-android)

gif

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -210,7 +210,7 @@ dependencies {
     implementation "com.google.firebase:firebase-messaging:17.3.4"
     implementation "me.leolin:ShortcutBadger:1.1.16@aar"
     implementation "com.facebook.react:react-native:+"  // From node_modules
-    implementation 'com.facebook.fresco:animated-gif:1.3.0' // For animated GIF support
+    implementation 'com.facebook.fresco:animated-gif:1.10.0' // For animated GIF support
     implementation "com.android.support:customtabs:${rootProject.ext.supportLibVersion}"
 
     testImplementation 'junit:junit:4.12'


agrawal-d:
@gnprice please review :)

agrawal-d:
Have made some changes, documenting the motivation behind the patch, and added commit summary as well.

agrawal-d:
Sorry, I shouldn't have worded it like that. I'm not sure that it is incompatible, rather, I'm sure that using it definitely causes GIFs to *not* work, and that using the version mentioned in the docs fixes that.

ray-kraesig:
> Sorry, I shouldn't have worded it like that. I'm not sure that it is incompatible, rather, I'm sure that using it definitely causes GIFs to **not** work, and that using the version mentioned in the docs fixes that.

Best not to say that in the commit message, then. :slightly_smiling_face: Amended and merged. Thanks!

PR #3826, opened 2020-01-20
action sheet: add border radius

agrawal-d:
Add border radius to ActionSheet on Android. Makes it look better ( IMO ).
action_sheet


gnprice:
Hi @agrawal-d!

I seem to recall there being a chat conversation about this idea not long ago. Would you link to that here? That'd help us avoid repeating discussions.

My immediate reaction is that this doesn't look like a style that's familiar from other apps. So I'd want some more reasoning before adding it.

agrawal-d:
Here's the link to the conversation: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Action.20Sheet/near/809519

PR #3828, opened 2020-01-21
account details: Add error handling for unknown time zones

### LINKED ISSUE #3816: App crashes on viewing user profile when timezone is unknown
**Steps to reproduce**:
1. Navigate here in the mobile app https://chat.zulip.org/#narrow/stream/48-mobile/topic/App.20crashing/near/809664
1. Tap on Steve Howell's profile picture.

**What happens**
The app crashes. Doesn't crash on viewing any other user's profile

**Error log** :http://p.ip.fi/ywPP

agrawal-d:
Fixes #3816

**Before**: App crashed when viewing profile of a user whose timezone was invalid ( i.e Deprecated ).

**Now**: App does not display the user's local time, and does not crash.

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -43,6 +43,15 @@ class AccountDetails extends PureComponent<Props> {
   render() {
     const { realm, user, userStatusText } = this.props;
 
+    // The set of timezone names in the tz database is subject to change over
+    // time. Handle unrecognized timezones by quietly discarding them.
+    let localTime: string | null;
+    try {
+      localTime = `${nowInTimeZone(user.timezone)} Local time`;
+    } catch (err) {
+      localTime = null;
+    }
+
     return (
       <ComponentList outerSpacing itemStyle={componentStyles.componentListItem}>
         <View>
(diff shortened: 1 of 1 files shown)


agrawal-d:
@ray-kraesig please review.

agrawal-d:
@ray-kraesig have made some changes based on your comments on this PR, and some other PRs. Please review. Thanks!

agrawal-d:
@ray-kraesig thanks for the review. Have pushed some changes.

ray-kraesig:
Merged (with slight emendation to include more detail about why). Thanks!

gnprice:
Thanks @agrawal-d and @ray-kraesig for this fix!

A small comment on [this bit](https://github.com/zulip/zulip-mobile/pull/3828#discussion_r371999897) for future reference:

> The fact that the app _does_ crash probably doesn't deserve mention here, since it's dependent on nonlocal information

I agree the crash shouldn't be mentioned in the comment, for the reason Ray described.

OTOH the fact that this was causing a crash is a perfect tidbit to mention in the *commit message*.

* The comments in the code are for understanding the code as it is. The fact that unrecognized timezones can exist is already a good enough reason that we should handle them correctly, regardless of what the exact consequences of getting them wrong would be.

* The commit message describes the *change* between the old and new code. When thinking about the change, the fact that it fixes a crash is quite an interesting one! It's part of the "why" of the commit -- it's always good to know why the change matters.

  We do often fix bugs even when we discovered them purely from (re)reading the code and it's probable that nobody has ever actually hit the bug, and even when something at another layer means they have no user-visible effect at all. But when reading the history to understand what changed, it's good to know which commits are like that and which are fixing a live bug, and in particular a crash.

Issue #2937, opened 2018-09-04
Act on "long timeout" warning, re-enable "yellow box" to show warnings


gnprice:
We have the following line in `ZulipMobile.js`:
```
console.disableYellowBox = true; // eslint-disable-line
```

This silences a useful mechanism RN offers for showing warnings when in debug mode, i.e. in development. We should turn this mechanism back on.

There are two warnings we currently get regularly that would make this noisy. Here they are as shown in `adb logcat *:S ReactNativeJS:V` (on Android; for iOS, use Xcode to see logs). One is about isMounted:
```
Warning: isMounted(...) is deprecated in plain JavaScript React classes. Instead, make sure to clean up subscriptions and pending requests in componentWillUnmount to prevent memory leaks.
```
It looks like this one may be an upstream issue, which they're working on: https://github.com/facebook/react-native/issues/18868 So let's just accept this one for now.

The other is about `setTimeout` with long durations:
```
Setting a timer for a long period of time, i.e. multiple minutes, is a performance and correctness issue on Android as it keeps the timer module awake, and timers can only be called when the app is in the foreground. See https://github.com/facebook/react-native/issues/12981 for more info.
(Saw setTimeout with duration 86497.55859375ms)
```
This seems like probably our code, and definitely an informative warning: whatever we're doing with such a long timer, we have a bug if we're assuming that timer will fire in any reasonable time. So we should track down what code of ours is doing that, and fix it.

gnprice:
Not needed for fixing this issue, but notes which may be handy for subsequent steps if we try to further improve our warnings in development:

[Example code](https://github.com/facebook/react-native/issues/4045#issuecomment-386408388) for customizing the handling of unhandled promise rejections.

[Background](https://github.com/then/promise/pull/116/files#diff-1e290ac8433d555bce009b162cb869d0) on how the concept of an "unhandled promise rejection" is tracked in the first place.

chrisbobbe:
I've just tried removing `console.disableYellowBox = true; // eslint-disable-line` to see these errors, and from some brief interaction with the apps on the iOS and Android simulators, these particular errors don't seem to be happening anymore! 🙂 

The `setTimeout` with long durations issue may have been in our code, but I'm familiar with this popping up from a [dependency](https://github.com/firebase/firebase-js-sdk/issues/97).

[...]

PR #3832, opened 2020-01-22
ComposeBox.js: On componentWillUnmount, clear 200ms timeout on input …

chrisbobbe:
…blurs.

Discovered in https://github.com/zulip/zulip-mobile/issues/2937#issuecomment-577368784.

To avoid calling setState on an unmounted component (yellow box warning
observed at https://github.com/zulip/zulip-mobile/issues/2937#issuecomment-577368784),
this clears the timeouts on componentWillUnmount that get set on handleMessageBlur
and handleTopicBlur to delay the call to this.updateIsFocused (which calls
this.setState) by 200ms.

From a quick search through the code, I don't see any other obvious cases where a
setTimeout delays a call to setState, but there may be other asynchronous tasks
that would trigger this warning.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -107,6 +107,8 @@ class ComposeBox extends PureComponent<Props, State> {
   messageInput: ?TextInput = null;
   topicInput: ?TextInput = null;
 
+  inputBlurTimeoutId: ?TimeoutID = null;
+
   static contextTypes = {
     styles: () => null,
   };
@@ -121,6 +123,11 @@ class ComposeBox extends PureComponent<Props, State> {
     message: this.props.draft,
     selection: { start: 0, end: 0 },
   };
+
+  componentWillUnmount() {
+    clearTimeout(this.inputBlurTimeoutId);
+    this.inputBlurTimeoutId = null;
+  }
 
   updateIsFocused = () => {
     this.setState(state => ({
@@ -201,7 +208,9 @@ class ComposeBox extends PureComponent<Props, State> {
       isMessageFocused: false,
       isMenuExpanded: false,
     });
-    setTimeout(this.updateIsFocused, 200); // give a chance to the topic input to get the focus
+    // give a chance to the topic input to get the focus
+    clearTimeout(this.inputBlurTimeoutId);
+    this.inputBlurTimeoutId = setTimeout(this.updateIsFocused, 200);
   };
 
   handleTopicFocus = () => {
@@ -219,7 +228,9 @@ class ComposeBox extends PureComponent<Props, State> {
       isTopicFocused: false,
       isMenuExpanded: false,
     });
-    setTimeout(this.updateIsFocused, 200); // give a chance to the message input to get the focus
+    // give a chance to the message input to get the focus
+    clearTimeout(this.inputBlurTimeoutId);
+    this.inputBlurTimeoutId = setTimeout(this.updateIsFocused, 200);
   };
 
   handleInputTouchStart = () => {


src/compose/ComposeBox.js
```diff
@@ -122,6 +124,11 @@ class ComposeBox extends PureComponent<Props, State> {
     selection: { start: 0, end: 0 },
   };
 
+  componentWillUnmount() {
+    // So this.setState is not called when the component is unmounted.
+    clearTimeout(this.updateIsFocusedTimeout);
```
ray-kraesig: This needs no justification. :slightly_smiling_face:

src/compose/ComposeBox.js
```diff
@@ -201,7 +208,8 @@ class ComposeBox extends PureComponent<Props, State> {
       isMessageFocused: false,
       isMenuExpanded: false,
     });
-    setTimeout(this.updateIsFocused, 200); // give a chance to the topic input to get the focus
+    // give a chance to the topic input to get the focus
+    this.updateIsFocusedTimeout = setTimeout(this.updateIsFocused, 200);
```
ray-kraesig: This will clobber an existing value of `this.updateIsFocusedTimeout`; if two blur events are handled in quick succession before the unmount, the first one's invocation will still be called.

chrisbobbe:
> I'm not really happy with the debounce hack at all.

It's a good point, but I didn't introduce the timeouts myself; this is just a surface-level attempt to clean up after them. 🙂 Before further review is needed, I'll look into using Animation to avoid having to do this.

src/compose/ComposeBox.js
chrisbobbe: Thanks for catching this!

ray-kraesig:
> It's a good point, but I didn't introduce the timeouts myself; this is just a surface-level attempt to clean up after them.  Before further review is needed, I'll look into using Animation, or the timestamp strategy you described, to avoid having to use the setTimeout hack.

No worries! As I said, this is (modulo my other concern) an acceptable fix. The `Animation` approach in particular is likely to have additional subtleties not worth addressing yet. (I've filed a new issue for it.)

chrisbobbe:
OK, sounds good! This should be ready for another review, then, when convenient. It cleans up the commit message and uses two separate instance properties, which I believe I've better named, to address the clobbering issue.

src/compose/ComposeBox.js
```diff
@@ -122,6 +125,11 @@ class ComposeBox extends PureComponent<Props, State> {
     selection: { start: 0, end: 0 },
   };
 
+  componentWillUnmount() {
+    clearTimeout(this.messageBlurTimeoutId);
+    clearTimeout(this.topicBlurTimeoutId);
```
ray-kraesig: I'm surprised this works! Apparently `clearTimeout({anything: "goes"})` is accepted without comment. (In the browser, at least.)

src/compose/ComposeBox.js
```diff
@@ -201,7 +209,8 @@ class ComposeBox extends PureComponent<Props, State> {
       isMessageFocused: false,
       isMenuExpanded: false,
     });
-    setTimeout(this.updateIsFocused, 200); // give a chance to the topic input to get the focus
+    // give a chance to the topic input to get the focus
+    this.messageBlurTimeoutId = setTimeout(this.updateIsFocused, 200);
```
ray-kraesig: This still won't work: if two blur events are triggered in quick succession, this will get double-updated.

(Which isn't likely under ordinary circumstances – but if the phone's responding sluggishly for one reason or another, the user's inputs could be buffered – topic, topic, topic, try to select something near the start of the message, tap the topic again just in case, give up and start mashing the back button.)

Just adding a `clearTimeout` before the `setTimeout` should handle all cases, though. (And then there's no need for two separate timeout variables.)

chrisbobbe: Ohhh yep, good catch, thanks a lot.

chrisbobbe: OK, I've made that fix.

ray-kraesig:
Merged. Thanks!

(It would be a nice bit of followup work to reduce the `?Type` fields here down to the more-specific `Type | void` (or `Type | null`), but that's more than a bit out of scope for this PR.)

PR #3834, opened 2020-01-23
compose box: enable Send button in PMs 

Maskedman99:
Fixes Issue #3820 Send button in private messages is disabled.
When the state 'topic' is initialized, it's value will be set to this.props.lastMessageTopic if this value is not empty, else the state's value will be set to the string 'topic'. This will avoid the topic in personal messages becoming empty, and hence won't disable the Send button.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -117,7 +117,7 @@ class ComposeBox extends PureComponent<Props, State> {
     isFocused: false,
     isMenuExpanded: false,
     height: 20,
-    topic: this.props.lastMessageTopic,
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks for your effort on this! Unfortunately, I’m closing this, as https://github.com/zulip/zulip-mobile/pull/3821 is already open and provides a better solution, as a) it uses existing logic to check if there is a topic, and b) it doesn’t rely on a potentially confusing [“magic string”](https://en.m.wikipedia.org/wiki/Magic_string) that might be a way for bugs to slip in. 🙂

Issue #3833, opened 2020-01-23
Emoji reaction search only checks beginning of name


chrisbobbe:
Along with https://github.com/zulip/zulip-mobile/issues/2927, this would really improve the ease of reacting with an emoji.

I just noticed that if I enter the term “smile” in the search box for choosing a message reaction emoji, the results are “smile,” “smile_cat,” “smiley,” “smiley_cat,” and “smiley_hacker.” But I’m actually looking for “slight_smile.” To get that to show up in the results, I have to start typing “slight...” — looks like it’s only searching the available emoji by the beginning of the name.

chrisbobbe:
I expect this will be a small, easy change to the predicate passed in the call to `names.filter` in the exported getFilteredEmojiNames function in src/emoji/data.js. Marking as a good first issue. 🙂 

Hmm! I just noticed that GitHub has the same limitation. Is there a reason we would want to exclude results where the search term matches something other than the beginning of the name?

sumukhah:
Working on the issue

chrisbobbe:
Thanks @sumukhah!

chrisbobbe:
Hmm, as @gnprice points out, there's a nice feature in the web app where if you start typing an emoji's name in the compose box (not the reaction search box!), it brings up a typeahead to search the emoji:

![image](https://user-images.githubusercontent.com/22248748/73016289-6c711c00-3dd2-11ea-87f6-f18535d31120.png)

The logic for that search is more sophisticated than I originally had in mind for this issue, but I think it would be good to implement here. In particular, see `#2`:

1. It searches the entire name, not just the beginning (as the current issue proposes), and
2. It also _reorders_ the search results, so that if they _happen_ to start with the search term, they are listed first.

`#2` can be seen in the screenshot, above: If I've started my search term with "light," I may want "slight smile" to appear, but I'm probably more interested in results that actually do start with "light": "light bulb," "lightning," etc., so they appear first.

To revisit my original example, you'll notice that if you try ":smile" in the web app, `#2` will push "slight smile" so far down the list (because "smile," "smiley," etc., appear first) that it actually doesn't appear. This is only because the typeahead is limited to 5 items, which isn't currently the case on mobile.

chrisbobbe:
[...]

PR #3836, opened 2020-01-23
Emoji reaction search: Fix emoji query search

sumukhah:
Emoji reaction search was querying emoji whose name begins with search term.
But if a search term is valid but not present at the beginning of that emoji
name, then searching was inefficient to query the required emoji.
This issue has resolved by querrying of emoji whose name contains the search
-Term, no matter where the search term is present.
manual testing has done.
Fixes:#3833

### DIFF
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -36,7 +36,7 @@ describe('getFilteredEmojiNames', () => {
   });
 
   test('returns a sorted list of emojis starting with query', () => {
-    const list = getFilteredEmojiNames('go', []);
+    const list = getFilteredEmojiNames('go', {});
     expect(list).toEqual([
       'go',
       'goal',
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -36,7 +36,7 @@ describe('getFilteredEmojiNames', () => {
   });
 
   test('returns a sorted list of emojis starting with query', () => {
-    const list = getFilteredEmojiNames('go', {});
+    const list = getFilteredEmojiNames('go', []);
     expect(list).toEqual([
       'go',
       'goal',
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Hi @sumukhah, thanks for this PR! This correctly fixes the issue I mentioned in my first comment, but it looks like some tests are failing. Please update the tests, in the same commit, to match the new behavior. That way, future commits won't accidentally interfere with the improvement you've made here. 🙂 

An extra challenge has become part of this issue, as described here: https://github.com/zulip/zulip-mobile/issues/3833#issuecomment-577837272. This will most likely mean passing a compareFunction argument to `.sort()` in the same area of the code you were looking at. (Docs can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)). The tests should also reflect the new sorting, and they should all pass. Also, please update the commit message to describe the sorting.

Tim and Greg have pointed out that the eventual goal is to share code between the web app and the mobile app, but getting this PR merged with the sorting functionality described in that comment would be a good short-term improvement.

sumukhah:
@chrisbobbe 
I'm facing some errors with testing,
`.sort((a, b) => a.indexOf(query) - b.indexOf(query))` is not working in test, but it gives result in devise.
Screenshot from 2020-01-25 01-18-32
Screenshot from 2020-01-25 00-58-00

sumukhah:
my last commit (tests) ran successfully in my system,
But got  errors in Travis-CI
can someone help me finding the error?
Thank you

gnprice:
Hi @sumukhah !

For help with the tests, see our docs on testing:
https://github.com/zulip/zulip-mobile/blob/master/docs/howto/testing.md

You don't mention what command you ran or what the output was, but I think if you run the tests yourself with the recommended command `tools/test`, you'll get the same error as in CI. That is:
```
Running prettier...
src/emoji/data.js
success formatting 1 file with prettier-eslint
478 files were unchanged
[...]
FAILED: prettier
```

For how to fix it, look for "prettier" on that docs page.

---

Meanwhile, it looks like we're moving faster than I'd expected on sharing this code from the webapp! So I'd recommend you put work on this PR on hold for now, as we might get that done soon.

chrisbobbe:
> Meanwhile, it looks like we're moving faster than I'd expected on sharing this code from the webapp! So I'd recommend you put work on this PR on hold for now, as we might get that done soon.

Thanks for your hard work on this, @sumukhah! Following Greg's comment, and as I explain at https://github.com/zulip/zulip-mobile/issues/3833#issuecomment-580053555, and I'm afraid I'll be closing this, with my apologies, at least for now. We're working on making this logic shared between the web app and the mobile app, and a lot of effort has gone into this already on the web side. 🙂

PR #3837, opened 2020-01-23
Revert "ComposeBox: Don't send message with whitespace-only topic."

### LINKED ISSUE #3820: Send button in private messages is disabled

ray-kraesig:
This reverts commit a2d6793aeac19259c205a175048f4017bf89ddcb.

Fixes #3820 (Send button never enabled in PMs), which was a regression
induced by the reverted commit.

Reopens #3743. Obviates and closes #3821.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -403,7 +403,7 @@ class ComposeBox extends PureComponent<Props, State> {
             style={this.styles.composeSendButton}
             Icon={editMessage === null ? IconSend : IconDone}
             size={32}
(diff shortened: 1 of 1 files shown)


gnprice:
Merged, thanks!

PR #3842, opened 2020-01-25
connect [nfc]: Work around Flow bug by making types more explicit.

### LINKED ISSUE #3768: Flow loses track of prop types when we say `export default connect(...`

gnprice:
This fixes #3768, using the pretty clean workaround we eventually got to.

### DIFF
--- a/src/account-info/AccountDetails.js
+++ b/src/account-info/AccountDetails.js
@@ -71,7 +71,7 @@ class AccountDetails extends PureComponent<Props> {
   }
 }
 
-export default connect((state, props): SelectorProps => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   realm: getCurrentRealm(state),
   userStatusText: getUserStatusTextForUser(state, props.user.user_id),
 }))(AccountDetails);
--- a/src/chat/Chat.js
+++ b/src/chat/Chat.js
@@ -52,6 +52,6 @@ class Chat extends PureComponent<Props> {
   }
 }
 
-export default connect((state, props): SelectorProps => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   canSend: canSendToNarrow(props.narrow) && !getShowMessagePlaceholders(props.narrow)(state),
 }))(Chat);
--- a/src/chat/UnreadNotice.js
+++ b/src/chat/UnreadNotice.js
@@ -66,6 +66,6 @@ class UnreadNotice extends PureComponent<Props> {
   }
 }
 
-export default connect((state, props): SelectorProps => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   unreadCount: getUnreadCountForNarrow(state, props.narrow),
 }))(UnreadNotice);
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -412,7 +412,7 @@ class ComposeBox extends PureComponent<Props, State> {
   }
 }
 
-export default connect((state, props): SelectorProps => ({
+export default connect<SelectorProps, _, _>((state, props) => ({
   auth: getAuth(state),
   ownEmail: getOwnEmail(state),
   usersByEmail: getActiveUsersByEmail(state),
--- a/src/emoji/Emoji.js
+++ b/src/emoji/Emoji.js
@@ -39,6 +39,6 @@ class Emoji extends PureComponent<Props> {
   }
 }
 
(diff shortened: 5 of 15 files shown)


ray-kraesig:
Sadly it doesn't seem to be easy to make `flow dump-types` produce `empty` (or `ComponentType<empty>`, etc.) for the types this fixes. (It would be nice to have a semi-automated before/after check.)

Still, looks good. Thanks!

gnprice:
> Sadly it doesn't seem to be easy to make `flow dump-types` produce `empty` (or `ComponentType<empty>`, etc.) for the types this fixes. (It would be nice to have a semi-automated before/after check.)

Huh, TIL about `flow dump-types`!

It looks like the thing that's really missing here is for `flow dump-types` to include values that appear as the ~"element type" in JSX, like the `AccountDetails` in `<AccountDetails user={user} />`.

I played around a bit with this. One workaround that works is to add a line like this, in say `AccountDetailsScreen.js` (which already imports and uses `AccountDetails`):
```
const c = AccountDetails;
```
Then in the "before", this produces
```
/home/greg/z/mobile/src/account-info/AccountDetailsScreen.js:66:7: ComponentType<empty>
/home/greg/z/mobile/src/account-info/AccountDetailsScreen.js:66:11-24: ComponentType<empty>
```
(that's one for `c`, and one for `AccountDetails`), and in the "after" it produces:
```
/home/greg/z/mobile/src/account-info/AccountDetailsScreen.js:66:7: ComponentType<{|+user: User|}>
/home/greg/z/mobile/src/account-info/AccountDetailsScreen.js:66:11-24: ComponentType<{|+user: User|}>
```
So, success!

It'd be a kind of awkward degree of manual effort to go and add those in near a call site of each of the affected components respectively. And I wouldn't want to do it inside the modules where they're defined, because the original bug seems to involve exports and module boundaries.

I had the thought that it could be reasonable to make a single file that just imports and mentions each of these components, in a boring mechanical way. Unfortunately the most boring version doesn't work -- if I say something like:
```
import AccountDetails from '../AccountDetails';
AccountDetails;
```
then the type comes out as `ComponentType<$ReadOnly<OwnProps<class AccountDetails, SelectorProps>>>`. Ditto with `const c = AccountDetails;` (plus, then the `c` comes out as `any (implicit)`.)

Oh well. Fun tool for another time.

PR #3845, opened 2020-01-28
Windows WSL doc tweaks

ray-kraesig:
Some minor tweaks to, and updating of, our experimental documentation for WSL.


gnprice:
Thanks! Merging.

    The existing name `windows.md` implies, despite the content, that it
    applies to all Windows users. (At present WSL is still marked
    "experimental".)

I think when we first added these instructions, my hope was that it could soon become our recommended normal way of doing it. That didn't work out. And considering it now, I'm not sure I see a strong reason to want to use it over Git Bash, anyway.

PR #3846, opened 2020-01-28
logging: truncate list of unknown message IDs

ray-kraesig:
Followup to #3733. (This list isn't necessarily any shorter than the
list of all private messages.)


gnprice:
Merged, thanks!

Issue #2914, opened 2018-08-24
[iOS] white screen visible when switching views in dark mode


lukasIO:
When I have dark mode activated and click on either of the home, starred or mentioned buttons,I can briefly see a white flash from what seems to be the opened view having a white background md then switching colors to the drk mode colors

chrisbobbe:
![Jan-28-2020 10-41-27](https://user-images.githubusercontent.com/22248748/73294614-0b26bf80-41bb-11ea-8cac-049b7fbc4139.gif)

Hmm! It's this, right? I've started to notice this disruption; in a dark room, I like to use night mode so I don't hurt my eyes, and this flicker is quite bright.

I'll open a PR for this soon; it seems to be that the WebView for the message list just shows white by default before it's finished initializing. It looks like one working solution is to use the startInLoadingState and the renderLoading props of WebView, to make the background dark in night mode during that time.

Issue #3026, opened 2018-10-08
Images not cached in WebView


armaanahluwalia:
I did a bit of digging and it turns out we are re-rendering the WebView on any change to messages. This causes a re-request for all images within a thread anytime the state of messages is changed. This is expensive both in terms, of performance as well as bandwidth usage.  A thread with more than a few images will cause a lot of wasted data.

2 possible solutions I can think of are -
1) Do not re-render WebView on all changes to messages and instead update, append or remove messages as needed.
2) Enable proper cacheing in WebViews. As per my quick research this is not as easy as it sounds. Probably different for Android and iOS.

Tested on both iOS and android. New requests are made to the server on any change to messages.

Relevant discussion in this thread https://chat.zulip.org/#narrow/stream/48-mobile/topic/Excessive.20Rendering.20Bug

gnprice:
See also #3790. The issues are independent, but each one basically multiplies the impact of the other.

gnprice:
The original description is a bit loosely worded, which I found confusing when re-reading this today. Here's some clarification, based on rereading the (quite detailed) linked chat thread, and also the code.

We're not in fact having a new React `render()` call that affects the WebView, or re-creating the React `WebView` component, or anything like that. At least, AFAICT there's no sign of that in this report.

What does happen is that, inside the `WebView` that provides the message list, we replace the entire content of the `body` with an updated version. See `handleUpdateEventContent` in `js.js`.

This happens whenever a new message comes in that should appear in the current message list, or another change happens that affects one of the messages in the current message list. See `MessageList#shouldComponentUpdate`, and the `getUpdateEvents` function (in `webViewHandleUpdates.js`) which it invokes.

So far, everything I've described above is by design. Then the observation that makes this issue is that apparently the browser handles images very inefficiently in this case: it fetches all the same images again, rather than get them from cache.

Separately, it sure feels inefficient to be replacing the entire body when only one message has changed (which is the usual case). Maybe we should change that. That's solution 1 in the description above, because it'd also resolve this issue.

Separately separately, there could be bugs where we replace the `body` much more often, in cases where in fact nothing in it has changed at all. The linked chat thread started with a report of a bug of that form. We promptly fixed that, in commit 9fc769802. If there are cases now where that happens, we should file new issues for those and fix them too.

chrisbobbe:
[...]

Issue #1946, opened 2018-02-19
Refactor our context providers to the new context API


borisyankov:
React 16.3 will introduce a new official API for working with context.
When React Native migrates to this React version (likely in the next version 0.54) we can rework our contexts to use this new API.

divyabaid16:
I would like to work on this issue. Can you plz guide me ?

borisyankov:
I am not yet familiar enough with the new context API :)
It is not urgent enough, so I am focusing my efforts on other areas of the app.

This issue would be interesting and appropriate for a student that wants to learn this on his own.

robgev:
@zulipbot claim

zulipbot:
Welcome to Zulip, @robgev! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-mobile/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

parshva-b:
Hey, I am Parshva Barbhaya, new to opensource and this community and want to participate in GSoC under this organization. I would like to work on this issue, but do need some guidance.

gnprice:
Some quick notes on what this means, partly to refresh myself:
* The only thing we use the old context API for is for a `styles` object. You can find uses with `git grep contextTypes`.
  * There's one exception: `TranslationProvider` uses the old API. A block comment there explains why that one is OK.
  * There are 14 of these left.
* See commit 51dd1b3b2, specifically its changes to `LineSeparator.js` and `miscStyles.js`, for an example of converting from the old to the new API. (Ignore the other files that commit touches; those were one-time changes to make such conversions possible. In retrospect I'd have done better to split it into two commits.)

PR #3848, opened 2020-01-28
MessageList: Fix white flicker in dark mode on iOS while WebView is i…

chrisbobbe:
Fixes: #2914.

Use startInLoadingState and renderLoading props to present a blank,
transparent View, during the first load.

react-native-webview does not specify exactly what work is being
done during this "first load," but it's fairly clear
(https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#startinloadingstate)
that it only happens (and our blank View will only show) on the
first load of the WebView, and not during API calls, etc.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { Component } from 'react';
-import { Platform } from 'react-native';
+import { Platform, View } from 'react-native';
 import { WebView } from 'react-native-webview';
 import type { WebViewNavigation } from 'react-native-webview';
 
@@ -199,6 +199,15 @@ class MessageList extends Component<Props> {
     return false;
   };
 
+  renderLoading = () => {
+    const style = {
+      backgroundColor: 'transparent',
+      width: '100%',
+      height: '100%',
+    };
+    return <View style={style} />;
+  };
+
   render() {
     const { styles: contextStyles } = this.context;
     const {
@@ -290,6 +299,8 @@ class MessageList extends Component<Props> {
     return (
       <WebView
         useWebKit
+        startInLoadingState
+        renderLoading={this.renderLoading}
         source={{ baseUrl, html }}
         originWhitelist={['file://']}
         onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -3,7 +3,6 @@ import React, { Component } from 'react';
 import { Platform, View } from 'react-native';
 import { WebView } from 'react-native-webview';
 import type { WebViewNavigation } from 'react-native-webview';
-
 import { connectActionSheet } from '@expo/react-native-action-sheet';
 
 import type {
@@ -46,7 +45,6 @@ import {
   getRealm,
 } from '../selectors';
 import { withGetText } from '../boot/TranslationProvider';
-
 import type { ShowActionSheetWithOptions } from '../message/messageActionSheet';
 import type { WebViewUpdateEvent } from './webViewHandleUpdates';
 import type { MessageListEvent } from './webViewEventHandlers';
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -9,7 +9,6 @@ import { connectActionSheet } from '@expo/react-native-action-sheet';
 import type {
   AlertWordsState,
   Auth,
-  Context,
   Debug,
   Dispatch,
   Fetching,
@@ -25,6 +24,8 @@ import type {
   ThemeName,
   User,
 } from '../types';
+import type { ThemeColors } from '../styles';
+import { ThemeContext } from '../styles';
 import { connect } from '../react-redux';
 import {
   getAuth,
@@ -137,15 +138,12 @@ const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/
 // [3] https://github.com/facebook/react-native/blob/0.59-stable/React/Base/RCTConvert.m#L85
 
 class MessageList extends Component<Props> {
-  context: Context;
+  static contextType = ThemeContext;
+  context: ThemeColors;
+
   webview: ?WebView;
   sendUpdateEventsIsReady: boolean;
   unsentUpdateEvents: WebViewUpdateEvent[] = [];
-
-  static contextTypes = {
-    styles: () => null,
-    theme: () => null,
-  };
 
   componentDidMount() {
     this.setupSendUpdateEvents();
@@ -209,7 +207,6 @@ class MessageList extends Component<Props> {
   };
 
   render() {
-    const { styles: contextStyles } = this.context;
     const {
       backgroundData,
       renderedMessages,
@@ -304,7 +301,8 @@ class MessageList extends Component<Props> {
         source={{ baseUrl, html }}
         originWhitelist={['file://']}
         onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
-        style={contextStyles.webview}
+        /* eslint-disable react-native/no-inline-styles */
+        style={{ backgroundColor: this.context.backgroundColor, borderWidth: 0 }}
         ref={webview => {
           this.webview = webview;
         }}
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -300,7 +300,7 @@ class MessageList extends Component<Props> {
         originWhitelist={['file://']}
         onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
         /* eslint-disable react-native/no-inline-styles */
-        style={{ backgroundColor: this.context.backgroundColor, borderWidth: 0 }}
+        style={{ backgroundColor: this.context.backgroundColor }}
         ref={webview => {
           this.webview = webview;
         }}
--- a/src/styles/miscStyles.js
+++ b/src/styles/miscStyles.js
@@ -88,8 +88,4 @@ export default ({ color, backgroundColor }: ThemeColors) => ({
     // alignItems: 'stretch',
     backgroundColor,
   },
-  webview: {
-    borderWidth: 0,
-    backgroundColor,
-  },
 });


chrisbobbe:
Fix on iOS, night mode:

fixed iOS, night mode

chrisbobbe:
iOS light mode (prior behavior preserved):

fixed iOS, light mode

src/webview/MessageList.js
```diff
@@ -199,6 +199,8 @@ class MessageList extends Component<Props> {
     return false;
   };
 
+  getBlankWebView = () => <View style={this.context.styles.webview} />;
```
gnprice: This name doesn't feel quite right -- after all, what this provides isn't a webview.

I think having the name start with "render" would also be helpful, as it's basically another render method: a method whose job is to prepare and return a bunch of components to get put into the tree.

In fact `renderLoading` (same name as the prop it's passed to) would be pretty good.

src/webview/MessageList.js
```diff
@@ -290,6 +292,12 @@ class MessageList extends Component<Props> {
     return (
       <WebView
         useWebKit
+        {...Platform.select({
+          ios: {
```
gnprice: Ah, this is a neat way of expressing this conditional!

chrisbobbe: I like it, if you do!

Maybe worth noting: It took some effort of reading the code for me to confirm that this way of selecting iOS doesn't contribute to unwanted rerenders of the WebView (which is not a PureComponent), which could be caused by new objects created and passed as props when MessageList's `render` is called. A new object _is_ created, but then that object is spread, to give the `startInLoadingState` and `this.getBlankWebView` props that don't change or get new references on MessageList's `render`.

The existing `source`, `originWhitelist`, and `ref` (there's a [new ref API](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) that doesn't tempt the creation of a new callback on each render) would seem to suffer from this problem. In practice, MessageList's shouldComponentUpdate returning false probably prevents most or all of the re-renders, but it's [noted](https://reactjs.org/docs/react-component.html#shouldcomponentupdate) that this isn't a 100% effective way to prevent re-renders, and its behavior may change:

> In the future React may treat shouldComponentUpdate() as a hint rather than a strict directive, and returning false may still result in a re-rendering of the component.

It looks like there is an existing issue relating to re-renders of the WebView, but I'm not fully sure what's causing those re-renders specifically: https://github.com/zulip/zulip-mobile/issues/3026.

chrisbobbe:
> And I think you said you were going to look into whether we can make the loading view transparent instead, which would be even better.

Yes, I've done some digging there. It'll be possible as long as we're not using react-native-webview on v5.6.0 or v5.6.1. The very latest version is v8.0.5 and we're on v5.0.0 right now. That issue was https://github.com/react-native-community/react-native-webview/issues/475. A `backgroundColor: 'white'` was introduced in the .container style and the default loading and error views in https://github.com/react-native-community/react-native-webview/pull/472/commits/f6c821e9ea1f9e46956ba016143631dc2618f9e0, and it was removed from the .container style in https://github.com/react-native-community/react-native-webview/commit/5ec74ea9bb93f04a90b0e8845d756496fceb6e9e#diff-a32cd1cea4d19bbfff56369224e62299. The .container style, it turned out, could not be overwritten, which is why there's no workaround for v5.6.0 and v5.6.1. The `backgroundColor: 'white'` still remains on the default loading and error views, but it still defaulted to white before that was introduced, and the only way of customizing them is to completely replace them; so,

tl;dr:

the complete workaround for <v5.6.0 and >=v5.6.2 is the same:

- include a `backgroundColor: 'transparent'` in the `styles` prop, and
- supply our own function as `renderLoading` that returns a View with `backgroundColor: 'transparent'`.

chrisbobbe:
Potentially worth taking a look at, but I don't have a competent estimate of the actual performance cost (of showing transparent things on iOS): https://github.com/facebook/react-native/commit/e4c53c28aea7e067e48f5c8c0100c7cafc031b06

src/webview/MessageList.js
chrisbobbe: Sure, `renderLoading` sounds good to me.

I'm not sure I'm entirely on board with react-native-webview's naming choices; as far as I can tell, `renderLoading` is only called to render the _initial_ load, not any subsequent ones, and is entirely irrelevant to something like `onShouldStartLoadWithRequest`, which we use. But at least we can be consistent with the names they've chosen (as, indeed, we've done with `onShouldStartLoadWithRequest).

chrisbobbe:
What do you think of the choice to use 100% width and 100% height on the View returned by `renderLoading`? I know we've used `flex: 1` in several places, but in this case, that puts our loading view only in the bottom half of the area covered by the WebView, because it's one of two siblings:

(output of the render method of WebView in react-native-webview/lib/WebView.ios.js)
```
      return (
        <View style={styles.container}>
          {webView}
          {otherView}
        </View>
      );
```

(otherView is our loading view; it could also be an error view.)

chrisbobbe:
OK, I believe I've addressed all the comments on here and in person, so this is ready for another review when it's convenient. :)

src/webview/MessageList.js
```diff
@@ -290,10 +298,13 @@ class MessageList extends Component<Props> {
     return (
       <WebView
         useWebKit
+        startInLoadingState
+        renderLoading={this.renderLoading}
```
gnprice: For the commit message, one perhaps-helpful phrase in the linked docs is "show the loading view on the first load." I think that helps clarify what it's meant to do.

In particular if in

    Use startInLoadingState and renderLoading props to present a blank,
    transparent View, during the initial "loading state."

we say

    [...] during "the first load."

I think that's already somewhat more reassuring. I agree it's still not the world's clearest documentation.

Tiny nit, but I also feel like I want a paragraph break between that sentence and what comes after. The first is saying what we do, and the rest is offering a caveat and expanding on that with some details.

src/webview/MessageList.js
```diff
@@ -137,15 +140,12 @@ const assetsPath = Platform.OS === 'ios' ? './webview' : 'file:///android_asset/
...
+  context: ThemeColors;
   webview: ?WebView;
   sendUpdateEventsIsReady: boolean;
   unsentUpdateEvents: WebViewUpdateEvent[] = [];
 
-  static contextTypes = {
-    styles: () => null,
-    theme: () => null,
-  };
+  static contextType = ThemeContext;
```
gnprice: Ah, well spotted finding #1946! I'd forgotten we had an issue on file for that migration. I just added a small comment there with things I learned from a quick look at the history (just now, and also earlier today when we were looking at this.)

Nit here: let's put `static contextType = ...` right above `context: ...`, and then a blank line setting the pair of those off from the other declarations; compare 51dd1b3b2. These two are closely linked. (In fact even with the old API, the old pair of declarations should have been kept together too.)

I want to link to this commit from that issue as an example, so polishing the style matters more than usual :)

src/webview/MessageList.js
```diff
@@ -290,10 +298,13 @@ class MessageList extends Component<Props> {
...
       <WebView
         useWebKit
+        startInLoadingState
+        renderLoading={this.renderLoading}
         source={{ baseUrl, html }}
         originWhitelist={['file://']}
         onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
-        style={contextStyles.webview}
+        /* eslint-disable react-native/no-inline-styles */
+        style={{ backgroundColor: this.context.backgroundColor }}
```
gnprice: Looks good!

In the commit message, a bit about how I look at this change:

    borderWidth defaults to 0. No visible changes were seen on iOS or
    Android after removing it. 32ce2086c102a4071b689574ed3a4e3577a48a20,
    which introduced this, doesn't have detail about the need for it.

A fact about that previous commit that's a key part of my reasoning that this is a good change is that that previous commit was a huge one doing lots of things (in fact introducing the entire WebView-based message list.) So for a small, odd detail like this with no explanation, there's little reason to think it was carefully considered in the first place when that change was made.

(If it were a focused change that did nothing but this, we could still end up concluding it didn't seem necessary... but I'd want to spend a *bit* more effort trying to reconstruct what the author then was thinking, and what they might have known that we now don't.)

So adding just a few words like "was a large change that introduced the whole WebView-based message list, and [doesn't have detail ...]" would help make clear we've covered that avenue.

Then a small style bit: 40 hex digits is a lot, and often doesn't flow well in prose (though you've managed it unusually well here!) I generally truncate commit IDs to 9 digits, which is still long enough to make ambiguity extremely unlikely.

src/webview/MessageList.js
```diff
@@ -25,6 +24,10 @@ import type {
   ThemeName,
   User,
 } from '../types';
+
+import type { ThemeColors } from '../styles';
+import { ThemeContext } from '../styles';
+
 import { connect } from '../react-redux';
```
gnprice: No blank lines between imports, except between the block of those from external libraries, at top, and from our own code, after that.

(I see this file already has a stray blank line, oops. But I think we're generally pretty consistent on this.)

chrisbobbe:
Thanks for the detailed review! I've pushed the latest iteration. N.B., a too-hasty use of tab-completion meant that I did a `git rebase --committer-date-is-author-date` instead of a `git rebase --continue`, and I couldn't easily find out how to revert that... So, just in case something looks odd with the committer or author dates. But the mistake didn't appear to be fatal.

chrisbobbe:
This is totally unrelated, but it addresses the same file you're reviewing, so it might (?) be helpful to post here.

It looks like it's possible for the interval that's set to retry the initial handshake with the WebView (by sending WebViewUpdateEventReady events) on componentDidMount to never be cleared, even when MessageList unmounts. I'm not sure how likely it is for the WebView to get stuck and never complete the handshake. But we are at least depending on MessageList not to be unmounted (e.g., by the user navigating away) within the first 30ms of mounting, and the fragility would increase if that 30ms interval is reconfigured to be greater for some reason:

```
  componentDidMount() {
    this.setupSendUpdateEvents();
  }

  /**
   * Initiate round-trip handshakes with the WebView, until one succeeds.
   */
  setupSendUpdateEvents = (): void => {
    const intervalId = setInterval(() => {
      if (!this.sendUpdateEventsIsReady) {
        this.sendUpdateEvents([{ type: 'ready' }]);
      } else {
        clearInterval(intervalId);
      }
    }, 30);
  };
```

I could open a PR to define intervalId as an instance property on MessageList instead of within the setupSendUpdateEvents closure, and call clearInterval on componentWillUnmount, if you like.

I can't be sure what the current effect of this is in the wild. But I did confirm in a small reproduction (increase the interval to 2000ms and navigate away during that time) that no error is thrown, and even though MessageList is unmounted, it's not garbage-collected:

```
  _isMounted = false;

  componentDidMount() {
    this.setupSendUpdateEvents();
    this._isMounted = true;
  }
  componentWillUnmount() {
    this._isMounted = false;
  }

  /**
   * Initiate round-trip handshakes with the WebView, until one succeeds.
   */
  setupSendUpdateEvents = (): void => {
    const intervalId = setInterval(() => {
      if (!this.sendUpdateEventsIsReady) {
        console.log('not ready; this:', this); // every 2 seconds, logs the MessageList instance with _isMounted: false
        this.sendUpdateEvents([{ type: 'ready' }]);
      } else {
        console.log('ready, clearing');
        clearInterval(intervalId);
      }
    }, 2000);
  };
```

gnprice:
Merged. Thanks!

gnprice:
> It looks like it's possible for the interval that's set to retry the initial handshake with the WebView (by sending WebViewUpdateEventReady events) on componentDidMount to never be cleared, even when MessageList unmounts.

Hmm, indeed! Well spotted. I agree it'd be good to be less fragile in that respect, even though with the exact current code (with the 30ms interval) it's unlikely to happen this way.


> I could open a PR to define intervalId as an instance property on MessageList instead of within the setupSendUpdateEvents body, and call clearInterval on componentWillUnmount, if you like.

SGTM, thanks!

PR #3849, opened 2020-01-28
about: Request navless versions of /terms and /privacy.

gnprice:
This option causes the server (since zulip/zulip#13758) to render
these pages' content alone, without the navigation header or footer
linking to other pages.

Older servers without this feature simply ignore the extra query
parameter, so this is a no-op there.


gnprice:
(No use merging this until zulip/zulip#13758 has been merged. And if the interface changes in discussion there, I'll want to update this one.)

gnprice:
The server PR is merged, and deployed on chat.zulip.org. I've confirmed empirically that with this PR branch, these two screens show no site navigation:

image

chrisbobbe:
Merged as 283d78587.

PR #3850, opened 2020-01-30
docs/howto/debugging: Move "you can skip this step" to top of the step.

chrisbobbe:
So people don't do the step unnecessarily.


gnprice:
Merged, thanks!

Issue #3857, opened 2020-02-01
Unread notice doesn't shrink completely when composeBox expands full screen.


jainkuniya:
On iOS:
![image](https://user-images.githubusercontent.com/18511177/73594459-a5c52c80-4534-11ea-83ee-345c06ee7461.png)

This is also reproducible on Android.

*Steps to reproduce*:
1. Open any chat which have unread messages. 
2. Press enter many times in compose box message input, so that compose box expands.
3. Unread notice doesn't goes off completely.

jainkuniya:
This is not completely fixed.

Even after 4ef9a13, in topic narrow it is reproducible 
![image](https://user-images.githubusercontent.com/18511177/75607274-61b94d80-5b1b-11ea-94a6-7b237a545305.png)

PR #3858, opened 2020-02-01
unreadNotice: Shrink height to 0 when compose box expands.

jainkuniya:
The height of UnreadNotice which was stucked on the top was
equal to the `paddingVertical` applied to it. So when it comes to shrink
RN was persisting padding.

Actually, this `View` wrapper is just a add on, not required. Styles can
be directly applied to `Animated.View`. So remove whis extra wrapper
from UnreadNotice, which fixes the issue.

Fixes: #3857

On this PR
image


gnprice:
Merged -- thanks!

PR #3859, opened 2020-02-01
compose box: Trim leading and trailing whitespace in topic.

### LINKED ISSUE #3743: Messages with whitespace-only topics should not be sent

agrawal-d:
Fixes #3743.

Trim the leading and trailing whitespace from topic when it is used by `getDestinationNarrow()` in `ComposeBox`. A consequence of this is that a whitespace-only topic gets converted to `(no topic)`.

Tested that:
- It works as expected.
- This mirrors the webapp behaviour, so it wont break things.
- Send button does not break in PM or group PM screens.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -228,7 +228,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
   getDestinationNarrow = (): Narrow => {
     const { narrow } = this.props;
-    const { topic } = this.state;
+    const topic = this.state.topic.trim();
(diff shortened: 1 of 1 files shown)


agrawal-d:
@ray-kraesig Can you please review this PR?

gnprice:
Bumping priority to match the issue.

ray-kraesig:
Looks good; merging. Thanks!

Issue #3860, opened 2020-02-01
Colour of Status bar in splash screen doesn't match with the brandcolor


Maskedman99:
![Screenshot_20200202-004353_Zulip](https://user-images.githubusercontent.com/31368194/73597787-f5682000-4555-11ea-836f-cc11a893f7b8.jpg)

Change the color of status bar to match with that of the splash screen color.

Maskedman99:
Can I work on this?

ray-kraesig:
For reference, this is only true on Android. The iOS status bar is effectively transparent:

![Screen Shot 2020-02-03 at 14 29 11](https://user-images.githubusercontent.com/54861490/73696888-1c2b7100-4692-11ea-8e03-2378d00c344b.png)

PR #3861, opened 2020-02-01
ui: Fix color of statusbar in loading screen.

Maskedman99:
fix the status bar color in loading screen issue in https://github.com/zulip/zulip-mobile/issues/3860
Screenshot_20200202-014022_Zulip (debug)

### DIFF
--- a/src/start/LoadingScreen.js
+++ b/src/start/LoadingScreen.js
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 import { StyleSheet, View } from 'react-native';
 
 import { BRAND_COLOR } from '../styles';
-import { LoadingIndicator } from '../common';
+import { LoadingIndicator, ZulipStatusBar } from '../common';
 
 const styles = StyleSheet.create({
   center: {
@@ -18,6 +18,7 @@ export default class LoadingScreen extends PureComponent<{||}> {
   render() {
     return (
       <View style={styles.center}>
+        <ZulipStatusBar backgroundColor={BRAND_COLOR} />
         <LoadingIndicator color="0, 0, 0" size={80} showLogo />
       </View>
     );


src/start/LoadingScreen.js
```diff
@@ -18,6 +18,7 @@ export default class LoadingScreen extends PureComponent<{||}> {
   render() {
     return (
       <View style={styles.center}>
+        <StatusBar backgroundColor={BRAND_COLOR} barStyle="light-content" />
```
agrawal-d: `backgroundColor` is an android-only prop. Have you tested how `barStyle="light-content"` looks in an IOS device, since `backgroundColor` is no-op for IOS?

agrawal-d:
@Maskedman99 thanks for the PR! I have left a single comment.

Your commit message needs to be changed. Please have a look at [our commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages).

Also, since I am a new contributor, a more experienced contributor will be able to review this change better. You might want to request a review from @ray-kraesig or @jainkuniya.

src/start/LoadingScreen.js
Maskedman99: I don't have access to iOS devices, but a quick check in snacks.expo.io showed the icons in status bar aren't visible on white background in iOS, changing ``barStyle="light-content"`` to ``barStyle="dark-content"`` the icons remain black and visible, will fix this now.

ray-kraesig:
* On Android, unlike on iOS, we customarily set the status bar off from the app background. It would probably be better to use our `ZulipStatusBar` component to do that (which also gets you the correct value of `barStyle` automatically).

* The commit message's body is 87 characters wide, which (per [the commit guidelines linked earlier](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages)) is too much.

* The commit message should be clear that nothing is done for iOS because nothing needs to be. (Its current wording gives the impression that a different fix is needed on iOS.)

ray-kraesig:
> lighten(0.1) was added because the ZulipStatusBar automatically
darkens any given background-color by 0.1, therefore passing a
color after lightening it by 0.1 balances this effect and gives the
desired output.

* Obtaining that darkening effect was exactly why I suggested using `ZulipStatusBar`: as I noted, we consistently have that darkening effect everywhere else on the Android app.

* In general, performing the inverse of part of an aggregate operation before performing the aggregate operation itself is a good way to make code too intertwined and tangled to easily understand or modify.

* `color.lighten(0.1).darken(0.1)` is not actually `color`, but `color.darken(0.01)`.

* Please don't close and reopen PRs while you're working on them. Just push if and when you have something new.

Maskedman99:
@ray-kraesig I've done the suggested modifications.

ray-kraesig:
The final code looks right, but you've added a new commit rather than correcting the old commit. You'll need to [squash](https://www.google.com/search?hl=en&q=git%20squash) the two commits into one.

Maskedman99:
@ray-kraesig I tried to squash and was able to merge the previous 2 commits to one but for some reason the previous previous commit is now being shown in the PR, making it show again 2 commits.

ray-kraesig:
> @ray-kraesig I tried to squash and was able to merge the previous 2 commits to one but for some reason the previous previous commit is now being shown in the PR, making it show again 2 commits.

This appears to be because you've also squashed two preceding unrelated commits together. You'll need to discard those and rebase your commit onto (preferably) the latest `master`.

Maskedman99:
Fixed it, I had to delete my previous commit, make a new one and push it. Deleting the previous one closed the PR automatically like before and I had to re-open it.

ray-kraesig:
Merged. Thanks!

(I've taken the liberty of tweaking the commit message slightly.)

> Fixed it, I had to delete my previous commit, make a new one and push it. Deleting the previous one closed the PR automatically like before and I had to re-open it.

For future reference, this shouldn't be necessary. The usual procedure is to fix the branch on your local machine, then force-push the branch to GitHub; this will leave the PR open and updated.

chrisbobbe:
https://zulip.readthedocs.io/en/latest/git/fixing-commits.html is a handy doc for fixing commits on a local branch :)

PR #3862, opened 2020-02-02
Lightbox: Show spinner while image is loading.

### LINKED ISSUE #2896: Show a spinner in lightbox when slow to load
As @sstangl suggested in #2888, when the image in the lightbox takes a while to load,

> It would be useful to:
> * Display something more meaningful than a black screen -- possible a spinner, that a background download is in progress?

This'll be somewhat less common in the medium term as we roll out server-side thumbnailing, but still a fairly common case when people are on slow network connections.

This is a polish feature that would be nice to add.

agrawal-d:
Show a spinner while image is loading in Lightbox.

lightbox

**Why did I not use the built-in prop `loadingIndicatorSource`?**
The said prop **does not work** on Android.  It might work for IOS, but I have not tested it.
I replaced `source` with a sample image, which worked, but the same image does not work for `loadingIndicatorSource`, which means that it's not an issue with opening the spinner resource file. A [developer here confrims](https://github.com/alwx/react-native-photo-view/issues/173) that it only works in IOS, as do [other issues in that repository](https://github.com/alwx/react-native-photo-view/issues?utf8=%E2%9C%93&q=is%3Aissue+loadingindicatorsource).

**Another note:**
The library we use, `react-native-photo-view` appears to be unmaintained (
Last  commit on 6 Apr 2018 ) and we should consider replacing it.

Fixes #2896.

### DIFF
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -10,7 +10,7 @@ import { connect } from '../react-redux';
 import type { ShowActionSheetWithOptions } from '../message/messageActionSheet';
 import { getAuth } from '../selectors';
 import { getResource } from '../utils/url';
-import { SlideAnimationView } from '../common';
+import { SlideAnimationView, LoadingIndicator } from '../common';
 import LightboxHeader from './LightboxHeader';
 import LightboxFooter from './LightboxFooter';
 import { constructActionSheetButtons, executeActionSheetAction } from './LightboxActionSheet';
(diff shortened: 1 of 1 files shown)


agrawal-d:
@ray-kraesig please review.

ray-kraesig:
Alas, this breaks the lightbox on iOS – the spinner appears, but when it goes away, the photo doesn't appear in its place.

agrawal-d:
> Alas, this breaks the lightbox on iOS – the spinner appears, but when it goes away, the photo doesn't appear in its place.

Strange! If the spinner does go away after some time, it means the callback for `onLoadEnd` has fired. The photo should appear then. What do you think might be the problem?

agrawal-d:
I'll close this as I don't have an IOS device to reliably test this.

Issue #3796, opened 2020-01-13
Image share from MessageList does not work as expected


agrawal-d:
**Steps to reproduce**
1. Long press on an image in message list
2. Tap Share in the ActionSheet.

**What is expected**
The image would be shared.

**What actually happens**
The name of the image file is shared ( the markdown title, if image is posted using the website ). 

![report](https://user-images.githubusercontent.com/7714968/72272756-2b00a400-364f-11ea-8476-1be002412bea.gif)

This works fine if I tap the image, open it in Lightbox and then share it.

rheo-chiti:
Similar to issue #3793 ,when the app crashes 
Sharing image button is showing unexpected behavior

agrawal-d:
@rheo-chiti I think these two are unrelated issues. The problem here is that it's not a bug, but a UX issue, most likely.

Issue #3793, opened 2020-01-12
App crashes on pressing share image button for the second time


rheo-chiti:
When share image button is pressed for the first time ,the image is downloaded and then we are shown the options where to share but on  pressing it the second time,the app crashes

![20200112_230133 (1)](https://user-images.githubusercontent.com/43493203/72222902-9daa4a80-358f-11ea-926b-5a08be20ab92.gif)

chrisbobbe:
Hmm, I wasn't able to reproduce this on a quick test on the Android emulator. What device/version are you seeing this on?

PR #3865, opened 2020-02-03
[WIP] messageActionSheet: Add option to copy link of message to clipboard

chrisbobbe:
Continuation of https://github.com/zulip/zulip-mobile/pull/3412.

In its current state, this PR isn't ready to review. It's the same as the previous one, except up-to-date with the current master, with some merge conflict resolutions. Now I'll work on Greg's latest comments in https://github.com/zulip/zulip-mobile/pull/3412#issuecomment-524109454.


gnprice:
Adding a cross-reference: I think this is meant to fix #2623.

gnprice:
> These commits do introduce user-facing strings; should I get set up with Transifex or is that something you do, @gnprice?

The essential thing in the PR branch is just that it adds to `messages_en.json`; that's the data we then feed to Transifex. I can sync to Transifex and back, but also we should get you set up with it :-)

src/message/messageActionSheet.js
```diff
@@ -120,14 +126,28 @@ const unstarMessage = async ({ auth, message }) => {
...
 shareMessage.errorMessage = 'Failed to share message';
 
+const copyMessageLink = ({ _, auth, message }) => {
+  if (message.isOutbox === true) {
+    throw new Error('Message has not been sent.');
+  }
+  Clipboard.setString(getLinkToMessage(auth.realm, message));
+  showToast(_('Copied link to message'));
+};
+copyMessageLink.title = 'Copy message link';
```
gnprice: I looked at this and thought "gosh, that sounds ambiguous -- what about when the message *contains* a link? won't it sound like this means to copy that link?"

Then I found there'd been discussion of exactly that on the previous PR: https://github.com/zulip/zulip-mobile/pull/3412#issuecomment-502324323 :slightly_smiling_face: and I guess my conclusion was:
> I think it's still a little ambiguous what "message link" means [...] Is it "Copy a link to the message", or "Copy the link in the message"? But it's good enough I'll be happy to merge it, and we can wordsmith further later.

So that seems reasonable -- I'll stand by that. If you have some ideas for that further wordsmithing, though, I'm interested!

static/translations/messages_en.json
```diff
@@ -47,7 +47,10 @@
   "Narrow to conversation": "Narrow to conversation",
   "Reply": "Reply",
   "Add a reaction": "Add a reaction",
-  "Copy to clipboard": "Copy to clipboard",
+  "Copy message content": "Copy message content",
+  "Copied message content": "Copied message content",
+  "Copy message link": "Copy message link",
+  "Copied link to message": "Copied link to message",
```
gnprice: I think there's a few more new/edited strings, like "Failed to copy message content to clipboard".

src/message/messageActionSheet.js
```diff
@@ -52,15 +58,15 @@ const reply = ({ message, dispatch, ownEmail }) => {
 reply.title = 'Reply';
 reply.errorMessage = 'Failed to reply';
 
-const copyToClipboard = async ({ _, auth, message }) => {
+const copyMessageContent = async ({ _, auth, message }) => {
   const rawMessage = isAnOutboxMessage(message) /* $FlowFixMe: then really type Outbox */
     ? message.markdownContent
     : (await api.getRawMessageContent(auth, message.id)).raw_content;
```
gnprice: In the last commit when `getRawMessageContent` gets introduced, this spot should switch to it. Basically that new function is the result of factoring out this statement.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -363,3 +364,78 @@ describe('getMessageIdFromLink', () => {
...
+
+  test('properly encodes topics with tricky characters', () => {
+    const { realm } = eg.selfAccount;
+    const message = eg.streamMessage({
+      stream_id: 123,
+      display_recipient: 'general',
+      subject: '.%%2E/#\ud83d\udc4b',
+    });
+    expect(getLinkToMessage(realm, message)).toEqual(
+      `${realm}/#narrow/stream/123-general/topic/.2E.25.252E.2F.23.F0.9F.91.8B/near/${message.id}`,
```
ray-kraesig: You may also want to add a non-ASCII Unicode character in the BMP (i.e., one not represented by a pair of surrogates in UTF-16).

src/message/messageActionSheet.js
ray-kraesig: Considering that the success string just above is "Copied link to message", I'd say present-tensing that as "Copy link to message" is probably for the best.

src/message/messageActionSheet.js
```diff
@@ -43,6 +43,11 @@ type ButtonDescription = {
 
 const isAnOutboxMessage = (message: Message | Outbox): boolean => message.isOutbox;
 
+const getRawMessageContent = async (auth: Auth, message: Message | Outbox) =>
+  isAnOutboxMessage(message) /* $FlowFixMe: then really type Outbox */
```
ray-kraesig: This should be fixable simply by testing `message.isOutbox` inline, rather than calling `isAnOutboxMessage`.

chrisbobbe:
> One UI comment: I think we can remove the "Share" option. It seems like a lot to have three of these:

I tend to agree, but just for completeness, here's a consolidated list of issues related to the "Share" option (and thus that would be affected by its removal), and @jainkuniya's comment in support of keeping it:

#3796
#3793
#3577
#2122
https://github.com/zulip/zulip-mobile/issues/2623#issuecomment-395811892

static/translations/messages_en.json
chrisbobbe: Good catch, thanks

chrisbobbe:
I've made the other changes; just my question in https://github.com/zulip/zulip-mobile/pull/3865#issuecomment-590566082 remains, I think.

ray-kraesig:
Probably what we need is some sort of hierarchical menu:
```
┎━━━━━━━━━━━━━━┓      ┌──────────────┐
┃ Copy...    ▶ ┃      │ Copy Link    │
┡━━━━━━━━━━━━━━┩      ├──────────────┤
│ Edit Message │      │ Copy Message │
├──────────────┤   ⇒  └──────────────┘
│ Share        │
├──────────────┤
│     ....     │
└──────────────┘
```
... and while I'm sure there's an easy-ish way to do that in React Native, I'm not familiar with it.

PR #3866, opened 2020-02-04
emoji: Cut misleading ignored reference to an emoji font.

gnprice:
Per the docs of the react-native-vector-icons package, this argument
is supposed to be the name of a font.  But there's no sign of a font
by this name actually appearing in the package, so it's hard to see
what passing it could actually mean -- instead, it just misleads the
reader by suggesting that this font is involved somehow.
    
And, empirically, the behavior is the same when this argument is
removed.  Tested in the reaction-list screen, in an emulated Pixel 3
running Android 10 and in an emulated iPhone 8 running iOS 13.

### DIFF
--- a/src/emoji/Emoji.js
+++ b/src/emoji/Emoji.js
@@ -10,7 +10,7 @@ import { getAllImageEmojiByName } from './emojiSelectors';
 
 /* $FlowFixMe: `nameToEmojiMap` is mistyped upstream; elements of
   `glyphMap` may be either `number` or `string`. */
-const UnicodeEmoji = createIconSet(nameToEmojiMap, 'AppleColorEmoji');
+const UnicodeEmoji = createIconSet(nameToEmojiMap);
 
 type SelectorProps = {|
   imageEmoji: ImageEmojiType | void,


chrisbobbe:
Mmm, thanks, you must have seen my comment at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/emoji.3A.20AppleColorEmoji.3F/near/814546 🙂

ray-kraesig:
For reference, for the curious: `AppleColorEmoji` is the name of a font available [here](https://poomsmart.github.io/repo/emoji10.html); but we've never shipped it, and I'm not sure if Apple has ever made those emoji glyphs available under a license that would permit us to.

Merging. Thanks!

chrisbobbe:
Great! Sadly this was not the happy path for me to get that 🧱 to show up on an Android device with out-of-date system fonts. Do you think it would unduly increase the maintenance burden if we included an up-to-date font file? Currently, users who see an empty square instead of 🧱 won't think of updating their system fonts; they'll see it as a bug in Zulip.

> Mmm, thanks, you must have seen my comment at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/emoji.3A.20AppleColorEmoji.3F/near/814546 🙂

^ Also I just realized that we were investigating the same thing separately, sorry about that; I didn't see this PR.

chrisbobbe:
If the info is still correct (I haven’t looked closely) this seems like a useful PR to merge, before work starts on shipping a font file with the app, as we discussed in person: https://github.com/zulip/zulip-mobile/pull/3686

PR #3867, opened 2020-02-04
message list [nfc]: Small cleanups around BackgroundData to prepare for adding more.

gnprice:
Inspired by looking back at this code while studying #3811.

@agrawal-d, after this PR you should be able to get the `is_admin` bit by just picking out `ownUser` and saying `ownUser.is_admin`, without any additional plumbing at the layers above the spot where you actually need it.


src/webview/MessageList.js
```diff
@@ -334,15 +334,15 @@ export default connect<SelectorProps, _, _>((state, props: OuterProps) => {
...
     auth: getAuth(state),
     debug: getDebug(state),
     flags: getFlags(state),
     mute: getMute(state),
     ownEmail: getOwnEmail(state),
     ownUserId: getOwnUser(state).user_id,
-    allImageEmojiById: getAllImageEmojiById(state),
-    subscriptions: getSubscriptions(state),
     twentyFourHourTime: getRealm(state).twentyFourHourTime,
+    subscriptions: getSubscriptions(state),
```
ray-kraesig: `"t" < "s"`?

src/webview/MessageList.js
```diff
@@ -72,6 +72,7 @@ export type BackgroundData = $ReadOnly<{
   flags: FlagsState,
   mute: MuteState,
   ownUser: User,
+  theme: ThemeName,
```
ray-kraesig: I'm not sure what jsdoc you're referring to when you say, in the commit message,

> as explained in the jsdoc on the type.

The jsdoc on `BackgroundData` in particular states: `Data about the user, the realm, and all known messages.`, which doesn't seem consonant with including `theme`.

src/webview/MessageList.js
```diff
@@ -78,7 +82,12 @@ export type BackgroundData = $ReadOnly<{
 }>;
 
 type SelectorProps = {|
+  // Data about the user, the realm, and all known messages.
```
ray-kraesig: This seems likely to diverge confusingly from its source.

src/webview/MessageList.js
gnprice: Ha, thanks for the catch. The two lists disagreed, but then I settled it the wrong way :)

src/webview/MessageList.js
gnprice: The theme is a global choice made by the user, so I think it fits broadly under "about the user".

The main reason I think it belongs here is the other side of the distinction, in the next paragraph: it's independent of the specific narrow and specific messages we're displaying.

Which also goes along with being suitable to include in the bundle that's wired through to a diverse set of helpers below: many of those helpers deal with a particular message, and when they also get data about the current narrow, that creates a risk of confusion (just like in #3811) about which information applies.

src/webview/MessageList.js
gnprice: Edited to be in terms of the other side of the distinction. (Though in practice I think this description is likely to be pretty stable too.)

gnprice:
Thanks for the review! Pushed a revised version.

PR #3868, opened 2020-02-05
SearchMessagesScreen: Store fetched messages in Redux store.

### LINKED ISSUE #3856: Opening reactions list from search history results in app crash
Steps to reproduce:
1. Search for "happy new year" on CZO using the app.
1. Long press on any reaction.

What happens:
App crashes immediately. Doesn't happen in case of all messages, only for certain messages.

What is expected:
I should be able to see the reactions screen.

Error log:
```js
TypeError: TypeError: Cannot read property 'reactions' of undefined
 
This error is located at:
    in MessageReactionList (created by Connect(MessageReactionList))
    in Connect(MessageReactionList) (at SceneView.js:9)
    in SceneView (at StackViewLayout.js:483)
    in RCTView (at View.js:45)
[...]
```
Complete log: http://p.ip.fi/iAcw

My device is running v26.20.143 (Android Oreo).

agrawal-d:
Fixes #3856.
The app no longer crashes on viewing the reactions from search results now.

### DIFF
--- a/src/search/SearchMessagesScreen.js
+++ b/src/search/SearchMessagesScreen.js
@@ -10,6 +10,7 @@ import { SEARCH_NARROW } from '../utils/narrow';
 import { LAST_MESSAGE_ANCHOR } from '../anchor';
 import { connect } from '../react-redux';
 import { getAuth } from '../account/accountsSelectors';
+import { messageFetchComplete } from '../message/fetchActions';
 
 type Props = $ReadOnly<{|
   auth: Auth,
@@ -36,18 +37,29 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
   /**
    * PRIVATE.  Send search query to server, fetching message results.
    *
-   * Does not update any state, or read any of the component's data except
-   * `props.auth`.
+   * Stores the fetched messages in the Redux store via a
+   * `messageFetchComplete` dispatch. Does not read any of the component's
+   * data except `props.auth` and `props.dispatch`.
    */
   fetchSearchMessages = async (query: string): Promise<Message[]> => {
-    const { auth } = this.props;
-    const { messages } = await api.getMessages(auth, {
+    const { auth, dispatch } = this.props;
+    const fetchArgs = {
       narrow: SEARCH_NARROW(query),
       anchor: LAST_MESSAGE_ANCHOR,
       numBefore: 20,
       numAfter: 0,
-      useFirstUnread: false,
-    });
+    };
+    const { messages, found_newest, found_oldest } = await api.getMessages(auth, fetchArgs);
+
+    dispatch(
+      messageFetchComplete({
+        ...fetchArgs,
+        messages,
+        foundNewest: found_newest,
+        foundOldest: found_oldest,
+      }),
+    );
+
     return messages;
   };
 


agrawal-d:
@ray-kraesig please review, thanks!

agrawal-d:
@ray-kraesig please review.

src/search/SearchMessagesScreen.js
```diff
@@ -37,17 +38,17 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
    *  Performs a network request associated with a query. Does not
    *  update or access internal state (except `auth`).
    */
-  performQueryRaw = async (query: string): Promise<Message[]> => {
+  performQueryRaw = async (query: string) => {
```
agrawal-d: The type 'ServerApiResponseMessages' is not exported and can not be used outside of `api/getMessages.js` so I have removed the flow type here.

ray-kraesig:
I'm going to turn this one over to @gnprice and/or @chrisbobbe for review, since they've been looking at the Redux message store recently.

src/search/SearchMessagesScreen.js
```diff
@@ -98,13 +100,25 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
...
     // A query is concluded. Report the message-list.
     this.setState({
-      messages,
+      messages: networkResponse.messages,
       isFetching: this.lastIdSent !== this.lastIdReceived,
     });
+
+    // store the fetched messages in the Redux store.
+    dispatch(
+      messageFetchComplete(
```
chrisbobbe: One thing we've noticed about messageFetchComplete is that it takes a lot of arguments, so it can be tricky to remember what each one is for — it's not obvious from looking at this call site what, for example, the 20 passed as the fourth argument means.

So, it would be helpful, in a separate commit before this one, to change messageFetchComplete so that it has just one parameter, which is an object, where the properties are the names of the current parameters. The main advantage of doing it this way is that when you pass that argument, you're forced to label its properties at or near the call site, and people reading the code there can more easily understand what's happening. 🙂 

Concretely, messageFetchComplete should look something like this:

```
export const messageFetchComplete = (args: {
  messages: Message[],
  narrow: Narrow,
  anchor: number,
  numBefore: number,
  numAfter: number,
  foundNewest?: boolean,
  foundOldest?: boolean,
}): Action => {
  const { messages, narrow, anchor, numBefore, numAfter, foundNewest, foundOldest } = args;
  return {
    type: MESSAGE_FETCH_COMPLETE,
    messages,
    narrow,
    anchor,
    numBefore,
    numAfter,
    foundNewest,
    foundOldest,
  };
};
```

You'll also want to update all call sites of the messageFetchComplete action creator accordingly, in the same commit. (Just the action creator, not the handlers in fetchingReducer.js and narrowsReducer.js with the same name.)

src/search/SearchMessagesScreen.js
```diff
@@ -37,17 +38,17 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
...
     const { auth } = this.props;
-    const { messages } = await api.getMessages(
+    const response = await api.getMessages(
       auth,
       SEARCH_NARROW(query),
       LAST_MESSAGE_ANCHOR,
       20,
       0,
       false,
     );
```
chrisbobbe: Following from my comment at messageFetchComplete, please also do a similar refactor for `api.getMessages` (the default export of src/api/messages/getMessages.js), in its own commit. 🙂 In that case, though, `auth` can be left as the first param because it's kind of an ambient piece of data that's not specific to any API request, and the rest can be brought together into a single object parameter.

src/search/SearchMessagesScreen.js
chrisbobbe: Then, after those two refactors I mentioned (and before your original commit), you can share code for passing arguments to api.getMessages and messageFetchComplete, since it takes exactly the same four pieces of data, with the same names! This can go near the top of handleQueryChangeInner. You can call it, for example, fetchArgs:

```
const fetchArgs = {
  narrow: SEARCH_NARROW(query),
  anchor: LAST_MESSAGE_ANCHOR,
  numBefore: 20,
  numAfter: 0,
}
```

To make that sharing easy, let's remove the `performQueryRaw` method. It's not very large, and it only has one call site, so it's OK to run inline. Replace its call site with `const networkPromise = api.getMessages(auth, fetchArgs);`, using the fetchArgs you've just defined.

Then, in your original commit, when you call messageFetchComplete, you can spread fetchArgs.

```
dispatch(
  messageFetchComplete({
    ...args,
    messages: networkResponse.messages,
    foundNewest: networkResponse.found_newest,
    foundOldest: networkResponse.found_oldest,
  }),
);
```

I think it's much easier to read than before. It's easy to see that the same four pieces of data are used in api.getMessages and messageFetchComplete, without having to read each one of them and compare them. 🙂

agrawal-d:
@chrisbobbe thanks for the review and explanations. I have pushed some changes.

agrawal-d:
@chrisbobbe ping

src/api/messages/getMessages.js
```diff
@@ -74,19 +74,22 @@ const migrateResponse = response => {
...
 ): Promise<ApiResponseMessages> => {
+  const { narrow, anchor, numBefore, numAfter, useFirstUnread } = args;
   const response: ServerApiResponseMessages = await apiGet(auth, 'messages', {
     narrow: JSON.stringify(narrow),
     anchor,
     num_before: numBefore,
     num_after: numAfter,
     apply_markdown: true,
-    use_first_unread_anchor: useFirstUnread,
+    use_first_unread_anchor: useFirstUnread || false,
```
chrisbobbe: You can actually still set a default value for useFirstUnread, in the destructuring assignment, like so, so you don't have to modify code wherever it's consumed. In this case, that's just one place, but this would be useful if a lot of things used it. 🙂 

```
const { narrow, anchor, numBefore, numAfter, useFirstUnread = false } = args;
```

Here's a doc describing this syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Setting_a_function_parameters_default_value

chrisbobbe:
Thanks for the ping, and for those improvements! It looks mostly good to me; I've added one small comment. We haven't forgotten about this, but Greg is working on https://github.com/zulip/zulip-mobile/pull/3911 and will want to coordinate that with this.

src/search/SearchMessagesScreen.js
```diff
@@ -78,33 +62,46 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
...
+    let apiResponse;
+    const fetchArgs = {
+      narrow: SEARCH_NARROW(query),
+      anchor: LAST_MESSAGE_ANCHOR,
+      numBefore: 20,
+      numAfter: 0,
+    };
+
+    // if the promise's construction fails, we let the exception
+    // propagate immediately
```
gnprice: This comment is explaining something that's no longer true after this change. It's about how the `performQueryRaw` call is made outside the `try`, and only the `await` is inside.

I think in fact it's fine, and clearer, to put them together as you do in this commit :-) See https://github.com/gnprice/zulip-mobile/commit/db8fbd16e2c0c96b3a128b861e61369ffe9f60fa (part of #3911) which makes the same change. The comment should be deleted to match, though.

src/search/SearchMessagesScreen.js
```diff
@@ -78,33 +62,46 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
...
+
+    // store the fetched messages in the Redux store.
+    dispatch(
+      messageFetchComplete({
+        ...fetchArgs,
+        messages: apiResponse.messages,
+        foundNewest: apiResponse.found_newest,
+        foundOldest: apiResponse.found_oldest,
+      }),
+    );
```
gnprice: * I think this should go before the `setState`, not after.

  I *think* the difference doesn't currently matter... but the `setState` may trigger a `render` call, and so in principle if that means a `render` call (and a series of further `render` calls on descendant components) before we have the new data stowed away in our Redux state then something could get confused by not finding data about a message.

* This will cause a `MESSAGE_FETCH_COMPLETE` action to be dispatched for this search narrow. Do all our reducers handle that appropriately?

  Tip: searching for `case MESSAGE_FETCH_COMPLETE` is great for finding all the places we act on a given action type in our reducers.

  I think in the `narrows` reducer we probably want to skip search narrows. We're not going to keep them up to date there (as new messages come in as events), and we'd accumulate a bunch of old searches.

  The `caughtUp` and `fetching` reducers should do the same as `narrows`.

  The `messages` reducer is the reason we're doing this PR, and `flags` goes along with it. I think those don't need any adjustment.

src/api/messages/getMessages.js
agrawal-d: Done! Thanks for the doc link!

src/message/fetchActions.js
```diff
@@ -62,16 +65,23 @@ export const fetchMessages = (
...
   dispatch(
-    messageFetchComplete(messages, narrow, anchor, numBefore, numAfter, found_newest, found_oldest),
+    messageFetchComplete({
+      messages,
+      narrow,
+      anchor,
+      numBefore,
+      numAfter,
+      found_newest,
+      found_oldest,
```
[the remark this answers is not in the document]
agrawal-d: Fixed!

src/api/messages/getMessages.js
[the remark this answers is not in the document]
agrawal-d: Will keep this in mind!

src/search/SearchMessagesScreen.js
agrawal-d: Deleted the comment.

src/search/SearchMessagesScreen.js
agrawal-d: Changed the reducers.

agrawal-d:
Thanks for the review, @chrisbobbe @gnprice! I have pushed some changes, and they are ready for a re-review.

agrawal-d:
I wasn't sure about the commit message for the changes to the reducers - so I've only added a commit summary to each commit:
`xyz reducer: Discard fetched messages from search narrow.`

agrawal-d:
Updated the new tests - they are properly typed now. @chrisbobbe  @gnprice please review, thanks!

src/message/fetchActions.js
```diff
@@ -62,16 +65,23 @@ export const fetchMessages = (
 ) => async (dispatch: Dispatch, getState: GetState) => {
   const useFirstUnread = anchor === FIRST_UNREAD_ANCHOR;
   dispatch(messageFetchStart(narrow, numBefore, numAfter));
-  const { messages, found_newest, found_oldest } = await api.getMessages(
-    getAuth(getState()),
+  const { messages, found_newest, found_oldest } = await api.getMessages(getAuth(getState()), {
```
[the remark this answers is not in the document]
chrisbobbe: There's also a call to api.getMessages in `src/search/SearchMessagesScreen` that should be updated in the same commit. It does get fixed and moved in the following commit ("SearchMessagesScreen: Store fetched messages in Redux store."), but the fix and the move are separate changes.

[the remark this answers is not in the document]
gnprice: I like having the changes to the two different functions (`messageFetchComplete` and `api.getMessages`) happen in separate commits. Definitely want the call sites of each function to get updated at the same time as the function, though. :-) Then the net is that the changes to one function plus its call sites are in one commit, and to the other in another commit, which makes it easy as the reader to see how they correspond to each other.

I recommend `git checkout -p`, along with `git rebase -i`, for rearranging which changes happen in which commit. It should be possible to do it without really having to edit the code further in your editor -- just `git rebase -i`, mark "edit" for the two commits in question, and use a combination of commands like `git checkout -p @^` and `git checkout -p search` (or whatever name you use for the branch) to remove or add specific changes in each commit.

src/chat/__tests__/fetchingReducer-test.js
```diff
@@ -97,4 +98,19 @@ describe('fetchingReducer', () => {
       expect(newState).toEqual(expectedState);
     });
   });
+
+  test('if fetched messages are from a search narrow, discard them', () => {
```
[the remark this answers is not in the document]
gnprice: Yeah. I think "ignore" in place of "discard" would work pretty well.

src/caughtup/__tests__/caughtUpReducer-test.js
```diff
@@ -65,6 +66,24 @@ describe('caughtUpReducer', () => {
...
+
+      const action = deepFreeze({
+        ...eg.action.message_fetch_complete,
+        narrow: [{ operator: 'search', operand: 'some query' }],
+      });
+
+      const newState = caughtUpReducer(initialState, action);
+
+      expect(newState).toEqual(initialState);
+    });
```
gnprice: Reading a test, a key question to ask is: does this really test what it's meant to test? This is especially important for tests, like this one, that test that nothing happens -- it'd be easy to accidentally write it so that some other aspect of the test case prevents the thing from happening, and so the test would pass even if the logic it's meant to test was completely missing.

It's hard to tell from reading just this test whether that's the case, because it's not obvious whether the example value for this action type would be one that makes `caughtUp` become true or not.

In general, one good way to make that clear is: make the test case's input data look exactly like another test case that tests something *does* happen, except for just one difference, namely whatever it is that you want to test prevents the thing from happening.

Here, that'd mean taking one of the test cases with "true" in the expected new state, copying it, and just changing the narrow to a search narrow.

src/chat/__tests__/fetchingReducer-test.js
gnprice: This should be nested in the `describe('MESSAGE_FETCH_COMPLETE'` to make clearer the context of what it's testing.

src/chat/fetchingReducer.js
```diff
@@ -51,7 +52,7 @@ export default (state: FetchingState = initialState, action: Action): FetchingSt
       return messageFetchStart(state, action);
 
     case MESSAGE_FETCH_COMPLETE:
-      return messageFetchComplete(state, action);
+      return isSearchNarrow(action.narrow) ? state : messageFetchComplete(state, action);
```
gnprice: These should be parallel to each other.

chrisbobbe:
One thing @gnprice and I noticed this morning, when looking at https://github.com/zulip/zulip-mobile/issues/3956, is that the messages we'll now be storing in Redux will be rendered with the search term highlighted, no matter which narrow they appear in! So if you do a search for "hello," and then you finish your search and go on to browse as normal, you'll be surprised to see that some messages appear with "hello" highlighted. This is something we want to avoid. 😛

Greg and I agreed that it makes sense to just omit the 'match_content' and 'match_subject' fields in the messagesReducer. (Take a look at the comment above these fields in the Message type at src/api/modelTypes.) Something like this should work:

```diff
    case MESSAGE_FETCH_COMPLETE:
      return {
        ...state,
-       ...groupItemsById(action.messages.map(message => omit(message, 'flags'))),
+       ...groupItemsById(action.messages.map(message => omit(message, ['flags', 'match_content', 'match_subject']))),
      };
```

src/chat/fetchingReducer.js
agrawal-d: I'm not sure what you mean by 'parallel' here.
Do you mean to bring `? state : mes ...` to it's own line? If so, prettier is automatically bringing them back. 
If not, can you please explain?

src/chat/__tests__/fetchingReducer-test.js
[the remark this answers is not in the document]
agrawal-d: Fixed!

src/chat/fetchingReducer.js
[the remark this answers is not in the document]
agrawal-d: Done!

agrawal-d:
@chrisbobbe @gnprice thanks for the review! I have pushed some changes.
Please review them.

src/caughtup/caughtUpReducer.js
```diff
@@ -63,18 +64,22 @@ export default (state: CaughtUpState = initialState, action: Action): CaughtUpSt
...
 
     case MESSAGE_FETCH_COMPLETE: {
-      const key = JSON.stringify(action.narrow);
-      let caughtUp;
-      if (action.foundNewest !== undefined && action.foundOldest !== undefined) {
-        /* This should always be the case for Zulip Server v1.8 or newer. */
-        caughtUp = { older: action.foundOldest, newer: action.foundNewest };
+      if (isSearchNarrow(action.narrow)) {
+        return state;
       } else {
```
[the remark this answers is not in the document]
agrawal-d: Sure!

src/message/fetchActions.js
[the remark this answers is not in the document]
agrawal-d: Done!

src/caughtup/__tests__/caughtUpReducer-test.js
agrawal-d: > In general, one good way to make that clear is: make the test case's input data look exactly like another test case that tests something _does_ happen, except for just one difference, namely whatever it is that you want to test prevents the thing from happening.

Thanks -this is helpful!

agrawal-d: I've changed the tests for `caughtUpReducer`, would you like me to also change the tests for other reducers I've modified? That would also make the commit `example data: Add sample action 'message_fetch_complete'` un-necessary.

chrisbobbe: @gnprice [said](https://github.com/zulip/zulip-mobile/pull/3868#discussion_r387931258):
> Here, that'd mean taking one of the test cases with "true" in the expected new state, copying it, and just changing the narrow to a search narrow.

In the status quo, this does sort of undo the progress toward getting more tests using exampleData. I wonder about, in a separate commit, getting all the existing tests in the file to use the action from `exampleData`, and then applying the change I've just quoted.

chrisbobbe: FYI, there will be merge conflicts with anything that uses `exampleData` because of https://github.com/zulip/zulip-mobile/commit/50a07b8b9c9105f18958568a49b709e1cdce3328.

chrisbobbe:
I'm sorry for the delay in responding; I'm not sure how this fell through the cracks!

I like the change you made to address https://github.com/zulip/zulip-mobile/pull/3868#issuecomment-595978666, and, at a first glance, it made sense to see it in its own commit.

Looking more closely, I've changed my mind; I think it belongs squashed into the main commit, "SearchMessagesScreen: Store fetched messages in Redux store.", since that's the commit where the unwanted behavior is introduced. The commit message can then be updated to explain the change.

This way, the unwanted behavior never appears in any commits, so it would still be theoretically possible to cut a release from any commit in the repository, without fear that things are broken. 🙂Also, while it's considerate to the reviewer to separate changes into different commits where possible, Greg pointed out (I'm trying to find a link...) that it can actually be *more* readable if these kinds of changes (ones that are needed to avoid breakage) are in the same commit as the main change. That way, it's clearer to the reviewer whether or not something is missing, without even having to check other commits.

Does this sound right, @gnprice?

src/caughtup/__tests__/caughtUpReducer-test.js
gnprice: > In the status quo, this does sort of undo the progress toward getting more tests using exampleData. I wonder about, in a separate commit, getting all the existing tests in the file to use the action from `exampleData`, and then applying the change I've just quoted. @gnprice, thoughts?

Yeah, I agree -- that combination would be ideal.

Doesn't need to mean getting the whole test files to be type-checked -- just using the new `eg.action.message_fetch_complete` in the places where it fits.

src/chat/fetchingReducer.js
gnprice: Sorry, I should have been clearer and used more words. It took me a minute myself just now to figure out what I meant. :wink: 

By "these" I meant the two cases for `MESSAGE_FETCH_START` and `MESSAGE_FETCH_COMPLETE`. They're currently quite parallel -- meaning they behave very similarly, with differences that naturally correspond to the difference between "start" and "complete". So when making changes the changes should continue to treat them in parallel.

src/chat/fetchingReducer.js
```diff
@@ -51,7 +52,8 @@ export default (state: FetchingState = initialState, action: Action): FetchingSt
       return messageFetchStart(state, action);
 
     case MESSAGE_FETCH_COMPLETE:
-      return messageFetchComplete(state, action);
+      // We don't want to accumulate old searches that we'll never need again.
+      return isSearchNarrow(action.narrow) ? state : messageFetchComplete(state, action);
```
gnprice, line 56: nit: Let's add the new logic around `isSearchNarrow` (and the comment that goes with it) inside the `messageFetchComplete` handler function, instead of here.

Similarly for `messageFetchStart` (when adding to that per my other comment about keeping them "parallel"), and in `narrowsReducer`.

In general, the way we structure all these reducers is:
* If the code for a given case is simple enough that it feels like it fits inline in the `switch` in the reducer function, we put it there.
* Otherwise we have one of these handler functions named after the action type, and that case in the handler just calls the handler which does all the work.

That's helpful because either way it keeps all the logic in one place -- if you see a function with a name like `messageFetchComplete` in a reducer file, you know it completely describes what we do in that reducer for a `MESSAGE_FETCH_COMPLETE` action.

gnprice:
> Looking more closely, I've changed my mind; I think it belongs squashed into the main commit, "SearchMessagesScreen: Store fetched messages in Redux store.", since that's the commit where the unwanted behavior is introduced. The commit message can then be updated to explain the change.

Sure, I agree this would be helpful.

The existing revision of the branch (to f81c7ba92) is already OK on this aspect, because this part:

> This way, the unwanted behavior never appears in any commits, so it would still be theoretically possible to cut a release from any commit in the repository, without fear that things are broken.

is already true -- the fix to omit `match_content` and `match_subject` happens before the main commit, so we're already omitting them before we start dispatching MESSAGE_FETCH_COMPLETE actions that would contain them.

But this point:

> Also, while it's considerate to the reviewer to separate changes into different commits where possible, Greg pointed out (I'm trying to find a link...) that it can actually be _more_ readable if these kinds of changes (ones that are needed to avoid breakage) are in the same commit as the main change. That way, it's clearer to the reviewer whether or not something is missing, without even having to check other commits.

applies, in that when looking at the commit that omits those fields, that code isn't yet ever actually processing messages for search narrows, so it's potentially confusing because the change doesn't yet have an effect.

Another solution, given that in the current revision of the branch we already fix the problem before it starts appearing, would be to just edit the commit message on the fix so that it mentions that this isn't happening yet but will start happening in an upcoming commit.

src/caughtup/__tests__/caughtUpReducer-test.js
agrawal-d: Done.

src/chat/fetchingReducer.js
agrawal-d: Thanks for the explanation.

agrawal-d: Done.

src/chat/fetchingReducer.js
agrawal-d: Done.

agrawal-d:
@chrisbobbe, @gnprice Thanks for reviewing! I've pushed some changes, and it's ready for a re-review.

src/caughtup/__tests__/caughtUpReducer-test.js
```diff
@@ -1,6 +1,7 @@
 /* @flow strict-local */
 import deepFreeze from 'deep-freeze';
 
+import * as eg from '../../__tests__/lib/exampleData';
 import caughtUpReducer from '../caughtUpReducer';
 import { MESSAGE_FETCH_START, MESSAGE_FETCH_COMPLETE } from '../../actionConstants';
```
chrisbobbe: Looks like the MESSAGE_FETCH_COMPLETE import isn't used anymore.

EDIT: Ah, looks like it was removed in the next commit. Let's remove it in the same commit that removes all uses, so `tools/test lint` passes at each commit. I caught this with `git rebase -i --exec 'tools/test --diff HEAD^`, which checks every commit on the branch after `upstream/master` (if the branch is set to track `upstream/master`). GitHub only runs CI tests for the single commit at the tip of the branch, I guess to save resources.

src/chat/fetchingReducer.js
```diff
@@ -51,6 +56,7 @@ export default (state: FetchingState = initialState, action: Action): FetchingSt
       return messageFetchStart(state, action);
 
     case MESSAGE_FETCH_COMPLETE:
+      // We don't want to accumulate old searches that we'll never need again.
```
chrisbobbe, line 59: Cool, looks like the `isSearchNarrow` logic was moved so it happens inside `messageFetchComplete`, following https://github.com/zulip/zulip-mobile/pull/3868#discussion_r411815301. It seems to me that this comment should only appear once (not twice), inside `messageFetchComplete`, right above the code it explains the reason for.

agrawal-d: Oops, I copied this line above, but forgot to delete it here.

src/caughtup/__tests__/caughtUpReducer-test.js
agrawal-d: Thanks, I'll keep this in mind. You had mentioned this before, but I forgot to use it.

chrisbobbe:
Nice, thanks for those fixes! Looks good to me. Pinging @gnprice with a note to see https://github.com/zulip/zulip-mobile/pull/3868#pullrequestreview-406858094.

gnprice:
Thanks @agrawal-d and @chrisbobbe ! All the preparatory commits look great.

For the final, main commit, I need to stare at it a bit harder (and perhaps #3911 to help refresh my memory) to remind myself how this all works.

Ah, I think one thing that's going on in that commit is that it's doing some refactoring to the control flow of this code -- like moving the equivalent of `this.performQueryRaw(query)` to inside the `try` block. And then when I reviewed an earlier version of this PR and started looking closely to convince myself that those made sense, I ended up producing a series of small isolated refactors to take this code in a similar direction, and #3911 was the result of that work. So probably what I want to do is
* maybe first go ahead and merge the first N-1 commits of this PR
* then merge the changes in #3911 to do those refactors
* then rebase this PR's main change on top of that -- which should produce a more isolated change that doesn't also contain those refactors.

gnprice:
OK! I've merged #3911, and also merged the first 8 of this PR's 9 commits:
6a41b4bfe example data: Add sample action 'message_fetch_complete'.
febcf2c94 caughtup reducer tests [nfc]: Use exampleData where possible.
5e3d59fb8 caughtup reducer: Don't store fetched messages from search narrow.
9c7705ebe fetching reducer: Don't store fetched messages from search narrow.
84efaafa9 narrows reducer: Don't store fetched messages from search narrow.
a7e527a79 messages reducer: Omit fields 'match_content' and 'match_subject'.
2d8bfc35f fetchActions [nfc]: Restructure messageFetchComplete parameters.
354cbeb49 getMessages [nfc]: Restructure getMessages arguments.

For the remaining commit in this branch, I think the code is now ready for a version of that change which doesn't engage in any broader refactoring. Specifically I think you can do the change entirely within the helper method `fetchSearchMessages`:
```js
  fetchSearchMessages = async (query: string): Promise<Message[]> => {
    const { auth } = this.props;
    const { messages } = await api.getMessages(auth, {
      narrow: SEARCH_NARROW(query),
      anchor: LAST_MESSAGE_ANCHOR,
      numBefore: 20,
      numAfter: 0,
      useFirstUnread: false,
    });
    return messages;
  };
```

It can dispatch the `messageFetchComplete` action, and then go ahead and return just `messages`, just as it currently does. This also keeps it very similar to the thunk actions in `fetchActions` that also fetch some messages and then dispatch a `messageFetchComplete` action.

@agrawal-d would you make that adjustment to the remaining commit?

agrawal-d:
>@agrawal-d would you make that adjustment to the remaining commit?

Done!

src/search/SearchMessagesScreen.js
```diff
@@ -40,14 +41,24 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
...
   fetchSearchMessages = async (query: string): Promise<Message[]> => {
-    const { auth } = this.props;
-    const { messages } = await api.getMessages(auth, {
+    const { auth, dispatch } = this.props;
+    const fetchArgs = {
       narrow: SEARCH_NARROW(query),
       anchor: LAST_MESSAGE_ANCHOR,
       numBefore: 20,
       numAfter: 0,
-      useFirstUnread: false,
```
agrawal-d, line 49: ` useFirstUnread` defaults to `false`.

gnprice:
Merged! Thanks again.

PR #3870, opened 2020-02-05
Cocoapods preparation

ray-kraesig:
Clean up a little, to make the life of anyone testing experimental branches a little easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,9 @@ npm-debug.log
 # Test files
 /pirlo-cli
 
+# CocoaPods (not currently used)
+/ios/Pods
+
 # Transifex, for a maintainer syncing translations up and down
 /.transifexrc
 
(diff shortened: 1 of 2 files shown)


gnprice:
Looks good, thanks! Two small nits:

* The tool's name is "CocoaPods", capitalized the same way as "GitHub" (and unlike "Dropbox" :wink: .) This applies to the commit messages, and the gitignore comment.

* For the first commit's summary line, I like the prefix `ios build:`. Generally the Android build and the iOS build (and the small amount of other build stuff) are very widely separated parts of the codebase from each other and have quite different flavors, so the distinction between them is a high-order bit that I like having in the prefix when scanning a commit log.

ray-kraesig:
Done!

gnprice:
Thanks!

Merged, with two small tweaks:
* fixed the name's capitalization in the gitignore comment
* s/under git control/under source control/ in the gitignore commit message, because the former doesn't really parse for me.

gnprice:
Also fixed this PR's title -- just to minimize the number of places that might contribute to putting the wrong capitalization in people's heads :slightly_smiling_face:

PR #3872, opened 2020-02-06
MessageActionSheet: Fix 'Reply' option to work correctly in different scenarios.

agrawal-d:
Fixes https://github.com/zulip/zulip-mobile/issues/3810.

* Show 'Reply' option for Outbox messages.
* Modify 'Reply' option for announcement-only narrow messages based on whether the user is an admin.

Tested in various narrows, including announcement-only narrows, PM narrows, search narrow, starred-messages narrow.


agrawal-d:
The commit summary is very long, but I couldn't think of any valid summary shorter than this.

ray-kraesig:
There are a number of test failures in CI. (Please run `tools/test jest` locally to observe and resolve these.)

However, I would advise against the general approach of this PR regardless. `constructMessageActionButtons` is already much too complicated (one of the most complicated in our codebase, in fact), and needs to be broken up and/or simplified before it's made more so.

gnprice:
Thanks @agrawal-d !

Here's some suggestions for the question you asked here:

> The commit summary is very long, but I couldn't think of any valid summary shorter than this.

Perhaps they'll be helpful for future commits.

---

The current summary line is

> MessageActionSheet: Fix 'Reply' option to work correctly in different scenarios.

* The meaning of "work correctly" is pretty much covered by "Fix". So you could delete that:

> MessageActionSheet: Fix 'Reply' option for different scenarios.

* That's pretty short! 63 columns. You could also cut "option" -- putting 'Reply' in quotes already makes it pretty clear in context that it's an option in the UI:

> MessageActionSheet: Fix 'Reply' for different scenarios.

* At this point, there's plenty of room in the summary line. :slightly_smiling_face: So you could expand "different scenarios" to give a bit more detail:

> MessageActionSheet: Fix 'Reply' for outbox messages and announcement-only streams.

* OK, that's a bit long again -- 82 columns. How about cut "messages":

> MessageActionSheet: Fix 'Reply' for outbox and for announcement-only streams.

* and I've also used the shorter "message actions" as a prefix for this subsystem:

> message actions: Fix 'Reply' for outbox and for announcement-only streams.

* That's back down to 74 columns, which is within the upper end of OK for a summary line.

---

In a different direction:

* When you find yourself writing something like "different scenarios" in the summary line, that's a sign that the change is probably several logically independent changes and should be several separate commits. See our guide on [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html).

  (And then most of the advice above would still apply to each of the resulting commits.)

hanhnguyen121088:
I cannot find reply fuction on zulip mobile. Could you plz help me on this issue

Issue #3874, opened 2020-02-06
MessageActionSheet title breaks for PMs in multi-stream narrows.


agrawal-d:
![Screenshot from 2020-02-06 20-55-20](https://user-images.githubusercontent.com/7714968/73955516-7b56e300-4929-11ea-8016-32417979ff01.png)

agrawal-d:
I'm working on this.

PR #3875, opened 2020-02-06
messageActionSheet: Fix action sheet title for private messages.

agrawal-d:
Fixes https://github.com/zulip/zulip-mobile/issues/3874

Before: 
image

Now:
Screenshot from 2020-02-06 21-45-10

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -1,6 +1,15 @@
 /* @flow strict-local */
 import { Clipboard, Share, Alert } from 'react-native';
-import type { Auth, Dispatch, GetText, Message, Narrow, Outbox, Subscription } from '../types';
+import type {
+  Auth,
+  Dispatch,
+  GetText,
+  Message,
+  Narrow,
+  Outbox,
+  Subscription,
+  User,
+} from '../types';
 import type { BackgroundData } from '../webview/MessageList';
 import {
   getNarrowFromMessage,
@@ -14,6 +23,7 @@ import * as api from '../api';
 import { showToast } from '../utils/info';
 import { doNarrow, startEditMessage, deleteOutboxMessage, navigateToEmojiPicker } from '../actions';
 import { navigateToMessageReactionScreen } from '../nav/navActions';
+import filteredRecipientsForPM from '../pm-conversations/filteredRecipientsForPM';
 
 // TODO really this belongs in a libdef.
 export type ShowActionSheetWithOptions = (
@@ -246,6 +256,19 @@ export const constructMessageActionButtons = ({
   return buttons;
 };
 
+/** Returns the title for the action sheet. */
+const getActionSheetTitle = (message: Message | Outbox, ownUser: User): string => {
+  if (message.type === 'private') {
+    const recipients = filteredRecipientsForPM(message, ownUser);
+    return recipients
+      .map(r => r.full_name)
+      .sort()
+      .join(', ');
+  } else {
+    return `#${message.display_recipient} > ${message.subject}`;
+  }
+};
+
 /** Invoke the given callback to show an appropriate action sheet. */
 export const showActionSheet = (
   isHeader: boolean,
@@ -278,7 +301,7 @@ export const showActionSheet = (
     {
       ...(isHeader
         ? {
-            title: `#${params.message.display_recipient} > ${params.message.subject}`,
+            title: getActionSheetTitle(params.message, params.backgroundData.ownUser),
           }
         : {}),
       options: optionCodes.map(code => _(allButtons[code].title)),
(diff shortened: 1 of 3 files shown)


agrawal-d:
Travis test is failing because of some dependency download issue.
> Could not download junit.jar (junit:junit:4.12): Skipped due to earlier error

I think this is temporary.

agrawal-d:
@ray-kraesig can you please review this, thanks!

src/message/messageActionSheet.js
```diff
@@ -274,11 +283,28 @@ export const showActionSheet = (
...
+    params.message.display_recipient.forEach(
+      (recepient: PmRecipientUser, idx: number, recepients: PmRecipientUser[]) => {
+        title += `${recepient.full_name}`;
+        if (idx < recepients.length - 2) {
+          title += ', ';
+        } else if (idx === recepients.length - 2) {
+          title += ' & ';
+        }
+      },
+    );
```
ray-kraesig: There are several issues here.

* We typically exclude the current user from the names of a group PM.
* `` `${someStringValue}` `` is redundant; it's equivalent to a bare `someStringValue`.
* The variable name `recepient` is misspelled (correctly: `recipient`).
* This is entirely localization-incompatible. In Japanese, for example, this would need to be something like `John Doeとのプライベートメッセー` and/or `John Doe、Welcome Bot、donut bot間のプライベートメッセー`: not only is different punctuation used, but the fundamental word order is different. String concatenation is, in general, unacceptable for localized text.

  At present, we actually give up on localization for this and use bare `names.join(', ')`, see e.g. [here](https://github.com/zulip/zulip-mobile/blob/28bd508eaedbbed35ca29486f5d363b222ff4cc3/src/pm-conversations/GroupPmConversationItem.js#L44). This is... not great,  but it mostly sidesteps the issue. Perhaps in the future we'll move to `Intl.ListFormat`.

  Also, Zulip prefers use of the Oxford comma (at least in the en-US locale).

src/message/messageActionSheet.js
```diff
@@ -274,11 +283,28 @@ export const showActionSheet = (
...
+  } else {
+    title = `#${params.message.display_recipient} > ${params.message.subject}`;
+  }
   showActionSheetWithOptions(
     {
       ...(isHeader
         ? {
-            title: `#${params.message.display_recipient} > ${params.message.subject}`,
+            title,
```
ray-kraesig: The new code always computes the title, even when unnecessary, and leaves a mutable variable floating around. Extracting title computation to a new function (instead of directly into this one) would avoid both of these.

ray-kraesig:
> Travis test is failing because of some dependency download issue.
> 
> > Could not download junit.jar (junit:junit:4.12): Skipped due to earlier error
> 
> I think this is temporary.

Confirmed. (I believe you should be able to restart a suspected-spuriously-failed CI run by clicking "Restart build" in the upper-right of the CI page. That may require permissions, though.)

agrawal-d:
Made some revisions. Current logic is largely borrowed from `src/webview/html/messageHeaderAsHtml.js`

src/message/messageActionSheet.js
```diff
@@ -246,6 +255,23 @@ export const constructMessageActionButtons = ({
...
+  // This means the message is a private message.
+  if (Array.isArray(message.display_recipient)) {
+    const recipients =
+      message.display_recipient.length === 1 && message.display_recipient[0].email === ownUser.email
+        ? message.display_recipient
+        : message.display_recipient.filter(r => r.email !== ownUser.email);
+    return recipients
+      .map(r => r.full_name)
+      .sort()
+      .join(', ');
```
ray-kraesig: This logic should probably be unified with that in `src/webview/html/messageHeaderAsHtml.js`, rather than merely copied from it.

I'm not sure there's a good place for that to go at the moment, though; in the absence of somewhere better, I'd suggest a new file in `src/pm-conversations`. :confused:

src/message/messageActionSheet.js
```diff
@@ -246,6 +255,23 @@ export const constructMessageActionButtons = ({
   return buttons;
 };
 
+/** Returns the title for the action sheet */
```
ray-kraesig: (Missing period.)

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -87,11 +88,7 @@ export default (
...
 
   if (item.type === 'private' && headerStyle === 'full') {
-    const recipients =
-      item.display_recipient.length === 1 && item.display_recipient[0].email === ownUser.email
-        ? item.display_recipient
-        : item.display_recipient.filter(r => r.email !== ownUser.email);
-
+    const recipients = filteredRecipientsForPM(item, ownUser);
     const narrowObj =
       recipients.length === 1
```
agrawal-d: `recipients` is being used here too, before map, sort aand join. If I include map, sort, and join in the common function I created, I won't be able to reuse it here. Do you still want me to include map, sort, and join in the new function I created?

ray-kraesig: Sorry, I'd missed that on first readthrough – and unfortunately, that value seems to flow into a morass of insufficiently-undocumented assumptions.

Ah well. Perhaps in some future version.

ray-kraesig:
Merged. Thanks!

PR #3876, opened 2020-02-06
Fix Button Position: Options position in Message Actions

sumukhah:
"See who reacted" option was listed first in message action options. then "Add reaction" was listed. This wasnt convinient for users. so interchanged their postions for the reason (i.e 'Add reaction'-'See who reacted').

Fixes: zulip#3871

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -210,11 +210,11 @@ export const constructMessageActionButtons = ({
   narrow,
 }: ConstructSheetParams): ButtonCode[] => {
   const buttons = [];
-  if (message.reactions.length > 0) {
-    buttons.push('showReactions');
-  }
   if (!isAnOutboxMessage(message) && messageNotDeleted(message)) {
     buttons.push('addReaction');
+  }
+  if (message.reactions.length > 0) {
+    buttons.push('showReactions');
   }
   if (!isAnOutboxMessage(message) && !isTopicNarrow(narrow) && !isPrivateOrGroupNarrow(narrow)) {
     buttons.push('reply');


chrisbobbe:
Thanks, @sumukhah!

I confirmed with the user that this is an Android-only issue; in fact, on iOS, the "Add Reaction" button is already in the same place every time, whether the message has reactions or not. So, we'll only want to do this re-ordering on Android. Here's one way this could be done; just be sure to include a code comment explaining the difference in behavior between iOS and Android. As @ray-kraesig points out, this file will need some careful rewriting, but it'll be good to get this fix out before we eventually do that.

```
  buttons.push(
    ...(() => {
      const reactionButtons = [
        message.reactions.length > 0 ? 'showReactions' : undefined,
        !isAnOutboxMessage(message) && messageNotDeleted(message) ? 'addReaction' : undefined,
      ];
      return Platform.OS === 'android'
        // On Android, keep "Add Reaction" consistently in one place (at the top) whether or not a message has reactions
        ? reactionButtons.reverse().filter(b => !!b) 
        // iOS already puts the "Add Reaction" in the same place, whether or not there are reactions
        : reactionButtons.filter(b => !!b);
    })(),
  );
```

I think the commit message could be made even more specific. You could put "Android message action sheet" before the colon, since it specifies what part of the code the change affects, and on what platform. And then after the colon, something that expresses the purpose of the change, like "Keep 'Add Reaction' button in the same place every time." The commit message can describe the change in more detail; the previous button positioning was inconvenient for a specific reason: very frequent users develop a habit of hitting the same spot on their screen every time, and it's annoying when this causes them to press the wrong button. 🙂 Also, we'd like to keep line lengths to a maximum of 70 characters. You can read more about commit message guidelines at https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages.

Another tip is that you can "rewrite history" on your fork, before a PR is merged into master, so that you don't have to make whole new branch and pull request when you want to change your commits on a PR. You can read about that [here](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#reordering-commits). 🙂

sumukhah:
Thank you so much @chrisbobbe

chrisbobbe:
Thanks, @sumukhah! I polished the commit message a bit and merged.

Issue #3654, opened 2019-10-18
PMs from new sources not immediately displayed in PMs list


ray-kraesig:
When a user receives a PM from a new source (_i.e._, one that had not previously sent them a PM), the center icon will correctly display a notification for a new message... but when tapped, that conversation-source does not appear in the PMs list; it remains unrefreshed.

(Force-stopping and reentering the app will also force a refresh.)

gnprice:
Yikes, that's a pretty bad symptom!

Would definitely be good to fix. The bug will very probably be in one of the "reducer" files maintaining the data structures that PM conversation list relies on.

chrisbobbe:
Aha! The 'narrows' state is not being updated to reflect the new messages.

Here's why: When a PM arrives, `action.caughtUp[ALL_PRIVATE_NARROW_STR].newer` is used by the narrowsReducer in the eventNewMessage function that handles the EVENT_NEW_MESSAGE action, to decide whether to update the state. `action.caughtUp[ALL_PRIVATE_NARROW_STR].newer` is just taken from the previous state, seen in the 'message' case in eventToAction.js. If the user has seen the most recent message, i.e., if `action.caughtUp[ALL_PRIVATE_NARROW_STR].newer` is true for that narrow, the state updates. If not, then the user is scrolling through old messages, probably from a link to some ancient history, and it would be confusing to append new messages directly after those messages, so the state does not update:

```
    const isInNarrow = isMessageInNarrow(action.message, JSON.parse(key), action.ownEmail);
    const isCaughtUp = action.caughtUp[key] && action.caughtUp[key].newer;
    const messageDoesNotExist = state[key].find(id => action.message.id === id) === undefined;

    if (isInNarrow && isCaughtUp && messageDoesNotExist) {
      stateChange = true;
      newState[key] = [...state[key], action.message.id];
    } else {
      newState[key] = state[key];
    }
```

But `action.caughtUp[ALL_PRIVATE_NARROW_STR].newer` is always false, despite the fact that (I believe) it should actually be always true. There's an initial fetch specifically of private messages that should be getting all of the most recent ones, and should be changing that .newer to true. (/messages is called from fetchPrivateMessages which is dispatched during doInitialFetch in src/message/fetchActions.js.) But the response to that /messages call has false for found_newest.

Further investigation will almost certainly involve https://github.com/zulip/zulip-mobile/issues/3133 (currently open); I just wanted to jot this finding down before I forget.

[...]

PR #3877, opened 2020-02-06
Set LAST_MESSAGE_ANCHOR to 10000000000000000 (sixteen zeroes).

chrisbobbe:
Fixes: #3654.

The server correctly interprets this special number as indicating
that we want the latest messages (0 is used to request the oldest
messages). Later, we can conditionalize on the server version to
support the updated API merged in
https://github.com/zulip/zulip/pull/13747, that use the strings
'oldest' and 'newest' instead of 0 and 10000000000000000.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -16,4 +16,6 @@ export const FIRST_UNREAD_ANCHOR = 0;
  * want to retrieve the newest messages for a narrow, without actually
  * knowing their IDs.
  */
+// This special value is understood by the server, corresponding to
+// LARGER_THAN_MAX_MESSAGE_ID there.  See #3654.
 export const LAST_MESSAGE_ANCHOR = 10000000000000000; // sixteen zeroes
--- a/src/constants.js
+++ b/src/constants.js
@@ -16,4 +16,4 @@ export const FIRST_UNREAD_ANCHOR = 0;
  * want to retrieve the newest messages for a narrow, without actually
  * knowing their IDs.
  */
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @chrisbobbe !

One point in the commit message: in the API, 0 doesn't mean the oldest unread messages, but simply the oldest. The legacy API for "oldest unread" is a separate boolean flag `use_first_unread_anchor: true`, which causes `anchor` to be ignored.

... Looking closer through our code for this, now that I have those facts about the API in my head from reading zulip/zulip#13747 recently, I see that *within* our code we do it the way the commit message describes, though. (And I guess we never make a request for simply the oldest messages.) It's confusing that we do something that's so almost, but not quite, like the server's API. I may see if there's a quick refactor I can do to clean that up.

Anyway, I'll make that adjustment in the commit message and then merge.

gnprice:
And pushed 7a3a84547 to try to clean up that confusion a bit:

```
commit 7a3a845477127fdb7a10b2dd682650c06f93ebaf
Date:   Thu Feb 6 20:45:28 2020 -0800

    anchor [nfc]: Correct the jsdoc on FIRST_UNREAD_ANCHOR.
    
    This jsdoc didn't match the name at all!  It turns out the name was
    right... and the jsdoc described what, confusingly, the very same
    value means in the actual Zulip API.
```

I also explored a refactor to actually stop using 0 in this confusing way. I have a draft that more or less does it, but more work is required before that's mergeable.

PR #3878, opened 2020-02-06
Add feature to view profile on tapping @mention for a user

### LINKED ISSUE #3879: MessageList : Tapping on @mention for a user should open the profile.
I expect that tapping on an @mention in the message list should take me to that user's profile page. However, it does nothing. 

In every other messaging app/ collaboration app that I use, this is the behaviour.

agrawal-d:
Closes #3879.

This is a common pattern in messaging apps, and a feature I felt was missing.

ezgif-3-258db77d92e6

I also created a selector  - `getUserForId` and wrote tests for it.

Tested that it works in varoius narrows, such as stream narrows, topic narrows, PM narrows, starred messages narow.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -604,6 +604,20 @@ const requireAttribute = (e: Element, name: string): string => {
   return value;
 };
 
+/**
+ * Returns the integer parsed value of a DOM element attribute.
+ *
+ * Throws if parsing fails.
+ */
+const requireNumericAttribute = (e: Element, name: string): number => {
+  const value = requireAttribute(e, name);
+  const parsedValue = parseInt(value, 10);
+  if (Number.isNaN(parsedValue)) {
+    throw new Error(`Could not parse attribute ${name} value '${value}' as integer`);
+  }
+  return parsedValue;
+};
+
 documentBody.addEventListener('click', (e: MouseEvent) => {
   e.preventDefault();
   clearTimeout(longPressTimeout);
@@ -629,7 +643,7 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
   if (target.matches('.avatar-img')) {
     sendMessage({
       type: 'avatar',
-      fromUserId: requireAttribute(target, 'data-sender-id'),
+      fromUserId: requireNumericAttribute(target, 'data-sender-id'),
     });
     return;
   }
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -656,6 +656,14 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.matches('.user-mention')) {
+    sendMessage({
+      type: 'mention',
+      userId: requireNumericAttribute(target, 'data-user-id'),
+    });
+    return;
+  }
+
   /* Should we pull up the lightbox?  For comparison, see the web app's
    * static/js/lightbox.js , starting at the `#main_div` click handler. */
   const inlineImageLink = target.closest('.message_inline_image a');
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -108,6 +108,11 @@ type MessageListEventReactionDetails = {|
   reactionName: string,
 |};
 
+type MessageListEventMention = {|
+  type: 'mention',
+  userId: number,
+|};
+
 export type MessageListEvent =
   | MessageListEventReady
   | MessageListEventScroll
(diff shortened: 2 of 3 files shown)


src/nav/navActions.js
```diff
@@ -103,3 +103,14 @@ export const navigateToLegal = (): NavigationAction => StackActions.push({ route
...
+
+export const navigateToUserScreenForUserId = (userId: number) => (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
+  const state = getState();
+  const user = getUserForId(state, userId);
+  if (user && user.email) {
+    dispatch(navigateToAccountDetails(user.email));
+  }
```
ray-kraesig: If this fails, it shouldn't do so silently; there should be some sort of notification (`Alert.alert`, perhaps?) and error handling (at least `logging.warn`).

Note that a failure here does not necessarily indicate a bug! I believe it may be possible, on some servers, to craft an @&zwj;-mention containing an invalid or unused user-id. (Or at least have historically been possible, in which case there may still be messages floating around in the history with invalid user-ids.)

src/nav/navActions.js
```diff
@@ -103,3 +103,14 @@ export const navigateToLegal = (): NavigationAction => StackActions.push({ route
 
 export const navigateToUserStatus = (): NavigationAction =>
   StackActions.push({ routeName: 'user-status' });
+
+export const navigateToUserScreenForUserId = (userId: number) => (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
+  const state = getState();
+  const user = getUserForId(state, userId);
```
ray-kraesig: As @gnprice noted [on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Selector/near/815187), this would probably be better handled via the existing `getAllUsersById` selector.

agrawal-d:
Here's how the Alert dialog looks for an invalid `user_id`
image

agrawal-d:

src/webview/js/js.js
```diff
@@ -639,6 +639,16 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.matches('.user-mention')) {
+    sendMessage({
+      type: 'mention',
+      userId: parseInt(requireAttribute(target, 'data-user-id'), 10),
+      // `innerText` is always present for @mention target. $FlowFixMe.
+      userName: target.innerText[0] === '@' ? target.innerText.substring(1) : target.innerText,
```
agrawal-d: Only used in the alert dialog, in case the user is not found,

src/webview/webViewEventHandlers.js
```diff
@@ -251,6 +259,18 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
       logging.error(`WebView exception: ${JSON.stringify(event.details)}`);
       break;
 
+    case 'mention': {
```
ray-kraesig: This should go with the rest of the normal event types, and above the special events like `'debug'` and `'error'`.

ray-kraesig:
Incidentally, I'd suggest holding off on this PR for now. There are plans to use IDs, rather than emails, almost everywhere as user identifiers; after #3789 is merged, it should be easy to convert `navigateToAccountDetails` in particular, which should save a fair amount of trouble.

agrawal-d:
@ray-kraesig https://github.com/zulip/zulip-mobile/pull/3789 has been merged. What changes should I make now?

ray-kraesig:
Something similar to that. Refactoring `navigateToAccountDetails` to use user IDs should probably go hand-in-hand with refactoring `getUserIsActive` and `AccountDetailsScreen` to do the same.

That's enough to be worth submitting on its own as a separate PR, though. (We don't have an issue for it specifically, but it's part of #&zwj;3764.)

agrawal-d:
https://github.com/zulip/zulip-mobile/pull/3968 is now merged :tada:! I've made some changes to this PR to use its results. Additionaly, the first commit addresses https://github.com/zulip/zulip-mobile/pull/3968#discussion_r398221311.

@ray-kraesig @gnprice please review.

ray-kraesig:
Merged – and thanks! (I've taken the liberty of amending the first commit to strike `[nfc]' from the summary and apply the mentioned grammar nit.)

agrawal-d:
Thanks for merging!
>and apply the mentioned grammar nit

Which nit are you referring to here?

ray-kraesig:
... what on *earth*, GitHub. :disappointed: *Hopefully,* it's been posted now.

(My other comment in the review was about `[nfc]` not being applicable, since the error paths were very different. It's gone, apparently, but that's no great loss.)

PR #3880, opened 2020-02-06
message actions [nfc]: Clarify which narrows offer "edit message".

gnprice:
This conditional does not mean the same thing as the old version!
The old version produced true for search narrows.

That would have been a bug, but... when we show a search narrow,
we actually pass `narrow={HOME_NARROW}` to `MessageList` (in
`SearchMessagesCard.js`), and so the `narrow` value that makes it
here is the home narrow.  So the bug is only latent, and this
commit makes no functional change.  Still good to fix latent bugs,
and to be explicit about the narrows where we *are* prepared to
offer the edit-message UI rather than try to list where we aren't.

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -4,9 +4,8 @@ import type { Auth, Dispatch, GetText, Message, Narrow, Outbox, Subscription } f
 import type { BackgroundData } from '../webview/MessageList';
 import {
   getNarrowFromMessage,
-  isHomeNarrow,
   isPrivateOrGroupNarrow,
-  isSpecialNarrow,
+  isStreamOrTopicNarrow,
   isTopicNarrow,
 } from '../utils/narrow';
 import { isTopicMuted } from '../utils/message';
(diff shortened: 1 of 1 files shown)


ray-kraesig:
Is there a particular reason that we disallow editing messages in the other narrows? `search` and `mentioned` are reasonable, but it would seem to make some sense in the `home` and `starred` narrows.

LGTM, though; merging. Thanks!

gnprice:
> Is there a particular reason that we disallow editing messages in the other narrows? `search` and `mentioned` are reasonable, but it would seem to make some sense in the `home` and `starred` narrows.

Yeah, I agree there are use cases for that -- particularly "home" / "all messages".

But the edit-message UI as currently implemented won't work there. In particular it depends on `ComposeBox`, which we don't show there.

timabbott:
Seeing this thread reminds me that `message_edit.get_editability`, `message_edit.get_deletability`, etc. are probably functions we should move to shared code.

gnprice:
> Seeing this thread reminds me that `message_edit.get_editability`, `message_edit.get_deletability`, etc. are probably functions we should move to shared code.

Hmm, I agree! Filed #3898 and a companion zulip/zulip issue.

PR #3882, opened 2020-02-07
types [nfc]: Add convenience type MessageLike.

ray-kraesig:
Replace the hack in `Outbox` (which typed it as having certain fields
which it never possessed in practice) with a slightly more
sophisticated hack in the form of a compatible auxiliary type.

This closes a minor soundness hole (values of type `Outbox` could
incorrectly have been assigned the value `undefined` in those fields).
Perhaps more importantly, it also makes `Outbox` a `JSONable` type.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -182,14 +182,43 @@ export type Outbox = {|
   subject: string,
   timestamp: number,
   type: 'stream' | 'private',
-
-  // These fields are always absent here, but can appear in `Message`.  We
-  // mention them here only to reassure Flow that they'll never be something
-  // more interesting than `undefined`, in order to type-check accesses on
-  // values of type `Message | Outbox`.
-  last_edit_timestamp?: void,
-  match_content?: void,
 |};
+
+/**
+ * MessageLike: Imprecise alternative to `Message | Outbox`.
+ *
+ * Flow reasonably dispermits certain classes of access on union types. In
+ * particular,
+ * ```
+ * const { sender_id } = (message: Message | Outbox);  // error
+ * ```
+ * is not allowed. However, as long as you're prepared to handle values of
+ * `undefined`, it's both JavaScript-legal to do so and occasionally convenient.
+ *
+ * We therefore construct an intermediate type which Flow recognizes as a
+ * subtype of `Message | Outbox`, but which Flow will permit us to directly (and
+ * soundly) destructure certain `Message`-only fields from:
+ * ```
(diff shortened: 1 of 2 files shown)


gnprice:
Interesting, thanks!

Making `Outbox` be `JSONable` sounds good. Do you understand why in the status quo it isn't? It seems like `{| foo?: void |}` contains just the same values as `{| |} | {| foo: void |}`, a union of two JSONable types, so it ought to be JSONable itself.

I'm not sure I follow the bit about a "minor soundness hole". Do you mean the fact that if there were an `Outbox` value floating around with one of those properties present, then that'd make the bit in the comment saying "These fields are always absent here" untrue? I don't see how that can lead to a value at runtime that doesn't fit in its static type, though, which is what unsoundness means.

ray-kraesig:
>  Do you understand why in the status quo it isn't?

Becuase `{| foo: void |}` is *not* `JSONable`: `JSON.stringify({foo: undefined}) === "{}"`.

As noted in its docstring, `JSONable` doesn't just mean "is accepted by `JSON.stringify`", but the rather stronger "is safely round-trippable through `JSON.stringify` / `JSON.parse`". (This was a deliberate design decision on my part; it could be changed, but I think it's the semantics we actually want.)

> I'm not sure I follow the bit about a "minor soundness hole".

PEBKAC. It's not a soundness hole at all, just a (trivial) incorrect typing. (It's possible to do `outboxMessage.last_edit_timestamp = undefined`, leaving you with an object that *does* have that property, as could be witnessed by `Object.keys` _et al._) Will fix.

gnprice:
> Because `{| foo: void |}` is _not_ `JSONable`:

Ah, indeed, thanks!

I was looking at the actual definition of `JSONable` when I wrote that. But I was focusing on the object part, thinking that you were pointing at the optional-ness of these properties, and evidently I skimmed too fast over the list of primitive types above that. (Which has `null | string | number | boolean`, but not `void`.)

ray-kraesig:
:ping_pong:. (The most recent push only corrects a similar lexical failure in a comment.)

Elisabe74:
Todos ustedes son una porquería.

Elisabe74:
Empezando por la hija de puta que utiliza mi cuenta

Elisabe74:
Todos ustedes son una porquería.

gnprice:
Merged -- thanks!

PR #3886, opened 2020-02-10
Add Lolex for tests involving time

ray-kraesig:
Separated out from #3727, with some improvements.

### DIFF
--- a/src/__tests__/aux/lolex.js
+++ b/src/__tests__/aux/lolex.js
@@ -47,6 +47,31 @@ export class Lolex {
 
   constructor() {
     this._clock = LolexModule.install();
+
+    // Wrap Lolex's setTimeout and setInterval with additional checks.
+    //
+    // In Node.js, where Jest runs, a timeout or interval of greater than
+    // INT32_MAX is reduced to exactly 1. (In browsers, and probably in React
+    // Native, it is instead truncated to an int32 value.)
+    //
+    // This can lead to subtle bugs. Rather than risk this, cause any test which
+    // would hit this check to fail immediately.
+    const maxTimer = 2 ** 31 - 1;
+    const wrapTimerFunction = fn => {
+      const wrapped = (_cb, delay: number, ...rest) => {
+        // do not refactor this conditional without considering NaN
+        if (!(delay <= maxTimer)) {
+          throw new Error(`timer too large (${delay}; max ${maxTimer})`);
+        }
+        // eslint-disable-next-line prefer-rest-params
+        return fn.apply(this, [_cb, delay, ...rest]);
+      };
+      return wrapped;
+    };
+
+    const clock = this._clock;
+    clock.setTimeout = wrapTimerFunction(clock.setTimeout);
+    clock.setInterval = wrapTimerFunction(clock.setInterval);
   }
 
   clearAllTimers(): void {
--- a/src/__tests__/metatests.js
+++ b/src/__tests__/metatests.js
@@ -75,7 +75,7 @@ describe('jsdom-global', () => {
   });
 });
 
-describe('lolex', () => {
+describe('Lolex', () => {
   const lolex: Lolex = new Lolex();
 
   afterAll(() => {
--- a/src/__tests__/metatests.js
+++ b/src/__tests__/metatests.js
@@ -1,6 +1,7 @@
 /** @jest-environment jest-environment-jsdom-global */
 // @flow strict-local
 import { Lolex } from './aux/lolex';
+import objectEntries from '../utils/objectEntries';
 
 /**
  * This file should not test any part of the application. It exists to test that
@@ -174,5 +175,31 @@ describe('Lolex', () => {
     expect(lolex.getTimerCount()).toBe(1);
     // ... and it has been fired appropriately many times.
     expect(count).toBeCloseTo((end - start) / interval, 1);
+  });
+
+  describe('timers fail on overflow', () => {
+    /** Helper function: return delayed do-nothing invocation of the supplied function. */
+    const invoke = (
+      fn: typeof setTimeout | typeof setInterval,
+      time: number,
+    ): (() => void) => () => {
+      fn(() => {}, time);
+    };
+
+    for (const [name, fn] of objectEntries({ setTimeout, setInterval })) {
+      test(name, () => {
+        expect(invoke(fn, 0)).not.toThrow();
+        expect(invoke(fn, 1)).not.toThrow();
+        expect(invoke(fn, 123456789)).not.toThrow();
+        expect(invoke(fn, 2 ** 31 - 1)).not.toThrow();
+
+        expect(invoke(fn, 2 ** 31)).toThrow();
+        expect(invoke(fn, 10e10)).toThrow();
+        expect(invoke(fn, Infinity)).toThrow();
+        expect(invoke(fn, NaN)).toThrow();
+
+        lolex.clearAllTimers();
+      });
+    }
   });
 });
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -2,6 +2,14 @@
 import { sleep, tryUntilSuccessful } from '../async';
 import { Lolex } from '../../__tests__/aux/lolex';
 
+const sleepMeasure = async (expectedMs: number) => {
+  const start = Date.now();
+  await sleep(expectedMs);
+  const actualMs = Date.now() - start;
+
+  return actualMs;
+};
+
 describe('sleep (ideal)', () => {
   const lolex: Lolex = new Lolex();
 
@@ -13,14 +21,6 @@ describe('sleep (ideal)', () => {
     // clear any unset timers
     lolex.clearAllTimers();
   });
-
-  const sleepMeasure = async (expectedMs: number) => {
-    const start = Date.now();
-    await sleep(expectedMs);
-    const actualMs = Date.now() - start;
-
-    return actualMs;
-  };
 
   test('waits for exactly the right number of milliseconds', async () => {
     const expectedMs = 1000;
@@ -40,10 +40,7 @@ describe('sleep (ideal)', () => {
 describe('sleep (real)', () => {
   test('waits for approximately the right number of milliseconds', async () => {
     const expectedMs = 1000;
-
-    const start = Date.now();
-    await sleep(expectedMs);
-    const actualMs = Date.now() - start;
+    const actualMs = await sleepMeasure(expectedMs);
 
     // `sleep` should sleep for the specified number of milliseconds (and not,
     // for example, that many seconds or microseconds).
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -4,11 +4,23 @@ import { sleep, tryUntilSuccessful } from '../async';
 describe('sleep', () => {
   test('waits for a given time in milliseconds', async () => {
     const expectedMs = 1000;
+
     const start = Date.now();
     await sleep(expectedMs);
-    const durationMs = Date.now() - start;
-    expect(expectedMs).toBeLessThanOrEqual(durationMs);
-    expect(durationMs).toBeLessThan(10 * expectedMs);
+    const actualMs = Date.now() - start;
+
+    // `sleep` should sleep for the specified number of milliseconds (and not,
+    // for example, that many seconds or microseconds).
+    expect(expectedMs).toBeLessThanOrEqual(actualMs);
+    expect(actualMs).toBeLessThan(10 * expectedMs); // [α]
+
+    // [α] In theory, we can't be sure of this test; the time between the
+    // specified timeout expiring and a timeslice actually becoming available
+    // may be arbitrarily long.
+    //
+    // In practice, 10x really is enough of a padding factor that it's never
+    // been an issue, even with real-world timers on shared CI hardware with
+    // unpredictable loads.
   });
 });
 
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -1,8 +1,44 @@
 /* @flow strict-local */
 import { sleep, tryUntilSuccessful } from '../async';
+import { Lolex } from '../../__tests__/aux/lolex';
 
-describe('sleep', () => {
-  test('waits for a given time in milliseconds', async () => {
+describe('sleep (ideal)', () => {
+  const lolex: Lolex = new Lolex();
+
+  afterAll(() => {
+    lolex.dispose();
+  });
+
+  afterEach(() => {
+    // clear any unset timers
+    lolex.clearAllTimers();
+  });
+
+  const sleepMeasure = async (expectedMs: number) => {
+    const start = Date.now();
+    await sleep(expectedMs);
+    const actualMs = Date.now() - start;
+
+    return actualMs;
+  };
+
+  test('waits for exactly the right number of milliseconds', async () => {
+    const expectedMs = 1000;
+    const sleepPromise: Promise<number> = sleepMeasure(expectedMs);
+
+    lolex.runOnlyPendingTimers();
+
+    // If `sleepPromise` hasn't resolved already, it never will; this await will
+    // hang. In this case, Jest will eventually time out and report failure.
+    // https://jestjs.io/docs/en/jest-object.html#jestsettimeouttimeout
+    const actualMs = await sleepPromise;
+
+    expect(actualMs).toEqual(expectedMs);
+  });
+});
+
+describe('sleep (real)', () => {
+  test('waits for approximately the right number of milliseconds', async () => {
     const expectedMs = 1000;
 
     const start = Date.now();


ray-kraesig:
Previous test breakage (only *mostly* due to late refactor) is now fixed.

src/__tests__/metatests.js
```diff
@@ -73,3 +74,109 @@ describe('jsdom-global', () => {
...
+      flag = true;
+    }, 3e9);
+    setTimeout(() => {
+      flag = true;
+    }, 6e9);
+
+    lolex.advanceTimersByTime(5.5e9);
+
+    expect(flag).toBe(true);
+  });
```
gnprice: Perhaps make the second timeout set the flag to `false`? That'd make the test slightly more informative; otherwise, it seems like it'd be more informative if the second timeout were deleted.

src/__tests__/metatests.js
```diff
@@ -73,3 +74,109 @@ describe('jsdom-global', () => {
...
+    const start = Date.now();
+
+    // set up eternal interval
+    let count = 0;
+    const interval = 10;
+    setInterval(() => ++count, interval);
+
+    const maxTime = 1000;
+
+    /* set up several timeouts, of length at most `maxTime` */ {
```
gnprice: nit: please put this open-brace on its own line, not after a comment :-)

Or just delete the braces -- I don't think they really add to the clarity here.

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,47 @@
...
+    const expectation: Promise<boolean> = expectationPromise();
+
+    // `expectation` should have set up its timer before returning control.
+    expect(lolex.getTimerCount()).toBe(1);
+
+    // (This would cause the commented test in `expectationPromise` to fail.)
+    lolex.advanceTimersByTime(expectedMs * 100);
+
+    // If this stalls, Jest will report failure.
+    expect(await expectation).toBeTrue();
```
gnprice: Hmm interesting! What does this mean -- how does Jest decide that it's stalled, vs. that it just hasn't completed yet (but later will)?

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,47 @@
...
+      const start = Date.now();
+      await sleep(expectedMs);
+      const durationMs = Date.now() - start;
+
+      expect(expectedMs).toBeLessThanOrEqual(durationMs);
+      // cannot be guaranteed: arbitrary delays are possible
+      // expect(durationMs).toBeLessThan(10 * expectedMs);
+
+      return true;
+    };
```
gnprice: I'd really love to avoid nesting stuff inside this promise, to the extent we can -- it makes the control flow a lot more complicated to reason about. Or if it really does need to be inside here to get us things we want, I'd like to understand clearly why that's the case, which I don't yet.

For a first step: what happens if we pull the `expect` call outside the promise, to be done after `await`ing it? And have this function return `durationMs`. Does that change the test's behavior, and if so how?

(Maybe this interacts with the answer to my question about "if this stalls".)

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,47 @@
...
-    expect(durationMs).toBeLessThan(10 * expectedMs);
+
+    const expectationPromise = async () => {
+      const start = Date.now();
+      await sleep(expectedMs);
+      const durationMs = Date.now() - start;
+
+      expect(expectedMs).toBeLessThanOrEqual(durationMs);
+      // cannot be guaranteed: arbitrary delays are possible
+      // expect(durationMs).toBeLessThan(10 * expectedMs);
```
gnprice: This seems like a regrettable loss in test coverage: it means that if our `sleep` acquired a bug where, say, we were confused about units and always slept for as many seconds as we'd intended milliseconds, it'd still pass the test. Is there a reason that switching to Lolex means we need to sacrifice our test coverage of that failure mode?

It's true that a `sleep` in real life can't guarantee a timely wakeup. That's why the test only asks that the interval be within 10x of the target, which I think has been enough to make flakes nonexistent even on the very messy real systems that Travis CI runs our tests in. When we move to a spiffy mock setup where we have total control of the (virtual) environment, what I'd expect to happen is that the behavior only becomes more predictable, not less -- so that we could tighten this to like `1.01 * expectedMs`, and it would pass every time. That would be a very appealing kind of thing to be able to do when trying to test tricky timer-using code.

src/__tests__/metatests.js
```diff
@@ -73,3 +74,109 @@ describe('jsdom-global', () => {
     }
   });
 });
+
+describe('lolex', () => {
```
gnprice: These tests are great!

I'd like to have them added in the same commit as the implementation -- that way the implementation comes with tests from the beginning, which I think is generally desirable when there isn't a reason that would make things too complex. In particular, as you note on the tests, they make a nice set of examples for using the new API.

.eslintrc.yaml
```diff
@@ -169,3 +173,7 @@ overrides:
         # [2]: https://github.com/tc39/proposals/commit/cb9c6e50
 
     no-with: error
+# Test file overrides
+- files: ['**/__tests__/**']
+  rules:
+    no-restricted-imports: off
```
gnprice: nit: missing final newline

src/__tests__/metatests.js
ray-kraesig: > That'd make the test slightly more informative

This is incorrect: it makes the test *much* more informative.

It reveals a bug, you see. :disappointed:

.eslintrc.yaml
```diff
@@ -80,7 +80,11 @@ rules:
   import/prefer-default-export: off
   import/no-extraneous-dependencies:
   - error
-  - devDependencies: ['**/__tests__/*.js', tools/**]
+  - devDependencies: ['**/__tests__/**/*.js', tools/**]
```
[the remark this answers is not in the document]
ray-kraesig: Indeed not. Broken out.

.eslintrc.yaml
```diff
@@ -169,3 +173,7 @@ overrides:
         # [2]: https://github.com/tc39/proposals/commit/cb9c6e50
 
     no-with: error
+# Test file overrides
+- files: ['**/__tests__/**']
```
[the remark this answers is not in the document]
ray-kraesig: Done. (This seemed to call for a new blank line above `# Global overrides`, for symmetry.)

src/__tests__/metatests.js
ray-kraesig: **Addendum:** The bug involves Lolex quietly converting timeout values greater than 2^31 - 1 into very short timeouts, to match Node's behavior. (Allegedly the cap here is part of the [HTML spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers), but I can't find language implying it. Modern Chrome and Firefox use the 32 least significant bits rather than rolling to 0.)

src/__tests__/metatests.js
ray-kraesig: As intended! This style of block-comment-plus-brace-pair is useful with code-folding.

That said, this block has been totally rewritten since it got that wrapper (it used to be a manually-specified list, rather than a for-loop), so folding it away is no longer helpful. Braces removed.

src/utils/__tests__/async-test.js
ray-kraesig: There's a configurable timeout, defaulting to 5000ms. [[1]](https://jestjs.io/docs/en/configuration.html#testtimeout-number) [[2]](https://stackoverflow.com/q/49603939)

src/utils/__tests__/async-test.js
ray-kraesig: No, that works just fine. Shuffled appropriately.

src/utils/__tests__/async-test.js
ray-kraesig: That would be reasonable... so long as about half of this comment is in the code near the 10x test. Especially the bit where a 10x test is specifically intended to guard against a unit-scaling bug, rather than just being an unusually-large fudge factor.

(Done.)

ray-kraesig:
The new version has an additional final commit (`lolex: Wrap setTimeout and setInterval to error on overflow.`) which is, in retrospect, probably a little too intrusive. If there are issues with it, I'd rather drop it than let it block the rest of this; we can relegate improvements to followup work if desired.

src/__tests__/metatests.js
gnprice: Well then! Glad that was informative. :slightly_smiling_face:

src/__tests__/metatests.js
gnprice: > useful with code-folding.

Hmm interesting!

I do like having the code be clear when read as-is, though. :wink: And a lot of what makes that work is expectations, and when I see a line-long comment, having an important bit of live code structure slip in at the end of the line is highly unexpected.

Anyway, happy it's moot in this case. If another case comes up in the future, we can discuss more in depth.

src/utils/__tests__/async-test.js
gnprice: Well, I think the origin of it really is as a large fudge factor. :slightly_smiling_face: It had been 10%, and in 935c1f52f9795c2f516a9a1e8531ff0a50bf8143 we loosened it to 10x because that had flaked.

The possible unit-scaling bug is just a nice example of how some code like this plausibly could actually have a bug that this test with a large fudge factor would catch.

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,53 @@
...
+    const actualMs = await sleepPromise;
+
+    // `sleep` should sleep for the specified number of milliseconds (and not,
+    // for example, that many seconds or microseconds).
+    expect(expectedMs).toBeLessThanOrEqual(actualMs);
+    expect(actualMs).toBeLessThan(10 * expectedMs); // [α]
+
+    // [α] In theory, we can't be sure of this test; the time between the
+    // specified timeout expiring and a timeslice actually becoming available
+    // may be arbitrarily long. (To simulate this with Lolex, one can set
```
gnprice: Is this true with Lolex? It seems like one of the great advantages of systematically mocking out the clock as well as timers, provided the mock implementation delivers this, is that we could expect the timers to behave ideally.

Which effectively means that for timer-intensive code like this, we'd have written much more of a true *unit test* than was ever possible with real clocks.

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,53 @@
...
+
+      return actualMs;
+    };
+    const sleepPromise: Promise<number> = sleepAsync();
+
+    // `sleepPromise` should have set up its timer before returning control.
+    expect(lolex.getTimerCount()).toBe(1);
+
+    const FUDGE_FACTOR = 1.01; // see discussion of [α], below
+    lolex.advanceTimersByTime(expectedMs * FUDGE_FACTOR);
```
gnprice: The FUDGE_FACTOR here seems like it makes this test loose in a way that it isn't now. I still don't see why any tradeoff like that should be a consequence of adopting a nice spiffy mock setup, where we get to control the clock's behavior in a way we didn't have before.

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,53 @@
...
+      const start = Date.now();
+      await sleep(expectedMs);
+      const actualMs = Date.now() - start;
+
+      return actualMs;
+    };
+    const sleepPromise: Promise<number> = sleepAsync();
+
+    // `sleepPromise` should have set up its timer before returning control.
+    expect(lolex.getTimerCount()).toBe(1);
```
gnprice: Is there something this tests about the function's behavior beyond what's tested below? This feels to me like it's only poking into the specifics of the function's implementation. (What if there were two timers involved? Not super likely for this particular function, given the simplicity of its spec... but I don't see how it'd be the caller's business if there were.)

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,53 @@
...
-    expect(durationMs).toBeLessThan(10 * expectedMs);
+
+    const sleepAsync = async () => {
+      const start = Date.now();
+      await sleep(expectedMs);
+      const actualMs = Date.now() - start;
+
+      return actualMs;
+    };
+    const sleepPromise: Promise<number> = sleepAsync();
```
gnprice: I like this simplification, thanks!

Can we make this another step more straightforward from here? I think it'd be helpful to basically inline this single-use function, like so:
```js
  const start = Date.now();
  const sleepPromise: Promise<void> = sleep(expectedMs);

  // ...

  await sleepPromise;
  const actualMs = Date.now() - start;
```

src/__tests__/metatests.js
```diff
@@ -73,3 +74,115 @@ describe('jsdom-global', () => {
...
+
+  test('timers fail on overflow', () => {
+    expect(() => {
+      setTimeout(() => {}, 10e10);
+    }).toThrow();
+
+    expect(() => {
+      setInterval(() => {}, 10e10);
+    }).toThrow();
+  });
```
gnprice: Why 10e10 in these tests and not something close to the required threshold, like 3e9 or 2**31?

src/__tests__/aux/lolex.js
```diff
@@ -0,0 +1,110 @@
...
+  _clock;
+
+  constructor() {
+    this._clock = LolexModule.install();
+
+    // Wrap Lolex's setTimeout and setInterval with additional checks.
+    //
+    // In Node.js, a timeout or interval of greater than INT32_MAX is reduced to
+    // exactly 1. (In browsers, and probably in React, it is instead truncated
+    // to an int32 value.)
```
gnprice: "in React" is a confusing expression here -- React also runs in browsers :-) and I think also in Node. I think you probably mean "React Native"?

src/utils/__tests__/async-test.js
gnprice: Cool, thanks.

Just "stalls" doesn't communicate that to me. How about "If this never completes, Jest will time out and report failure."?

gnprice:
Most of these commits LGTM, including the main one, so I've gone and merged them! Thanks again.

dd53b0d19 eslint: Extend test-file import permissions to subdirectories.
2faad06a8 eslint: Disallow imports from __tests__ directories.
6d1d1df89 jest: Add Lolex for proper time mocking.
6dea1330d queueMarkAsRead: Adapt to use Lolex in tests.

That leaves two commits in progress: the `sleep` test, and the newest commit on top.

src/utils/__tests__/async-test.js
ray-kraesig: > Is this true with Lolex?

It's not – hence the trailing parenthetical. That could stand to be made more clear, though.

> It seems like one of the great advantages of systematically mocking out the clock as well as timers, provided the mock implementation delivers this, is that we could expect the timers to behave ideally.

Alas, "ideally" contrasts with "in reality", making this simultaneously one of the great *disadvantages* of mocking out the clock. We don't have any simple way to ensure that our code behaves properly under real-world conditions like load-induced greater-than-epsilon timeslice delays.

> Which effectively means that for timer-intensive code like this, we'd have written much more of a true *unit test* than was ever possible with real clocks.

That's true, but it's not an unalloyed good. :disappointed: An integration test may well be more informative than a unit test.

src/utils/__tests__/async-test.js
ray-kraesig: Quite the opposite: it's intended to replicate the looseness that *is* present now. (Clarifying the preceding would also clarify this, I expect.)

I suppose we could have a separate test that's "sharp", so to speak. I'm not sure it would be terribly informative, though.

src/utils/__tests__/async-test.js
ray-kraesig: This was intended not as a test of `sleep`'s behavior, but as a metatest of `sleepPromise`'s behavior (resulting from an actual failure in its original implementation).

It's probably not necessary.

src/utils/__tests__/async-test.js
ray-kraesig: That's very much *less* straightforward, not more – it mixes the two control paths, and in subtly fragile ways.

In particular, `sleepPromise` will hang unless you advance the timer before `await`ing on it, and doing so would make the computation of `const actualMs` [0] meaningless.

src/__tests__/metatests.js
ray-kraesig: I like 10e10 more than 3e9. That's about it, really.

It's probably worth adding tests that are just barely on either side of the limit, as well as versions with `NaN` and `Infinity`.

src/__tests__/aux/lolex.js
ray-kraesig: ... yes. :man_facepalming:

ray-kraesig:
The previous muddled Lolex-based `sleep` test has been replaced with a sharp Lolex-based `sleep` test. This only tests `sleep`'s behavior under an ideal clock; rather than trying to write a Lolex-based test for less-ideal clocks, the original code from `master` has been left in place, continuing to use the real clock for this purpose.

src/utils/__tests__/async-test.js
ray-kraesig: (Expanded somewhat more than even that, including a documentation link.)

src/utils/__tests__/async-test.js
```diff
@@ -1,14 +1,59 @@
...
+  });
+
+  afterEach(() => {
+    // clear any unset timers
+    lolex.clearAllTimers();
+  });
+
+  test('waits for exactly the right number of milliseconds', async () => {
+    const expectedMs = 1000;
+    const sleepPromise: Promise<number> = sleepMeasure(expectedMs);
```
gnprice: I think this line appears in a commit where `sleepMeasure` hasn't yet :slightly_smiling_face: 

Maybe put the "factor out sleepMeasure" commit earlier, with just one callsite, and then add this new callsite right after that? Or other sequences would be good too.

src/__tests__/aux/lolex.js
```diff
@@ -47,6 +47,31 @@ export class Lolex {
...
+    // INT32_MAX is reduced to exactly 1. (In browsers, and probably in React
+    // Native, it is instead truncated to an int32 value.)
+    //
+    // This can lead to subtle bugs. Rather than risk this, cause any test which
+    // would hit this check to fail immediately.
+    const maxTimer = 2 ** 31 - 1;
+    const wrapTimerFunction = fn => {
+      const wrapped = (_cb, delay: number, ...rest) => {
+        // do not refactor this conditional without considering NaN
+        if (!(delay <= maxTimer)) {
```
gnprice: Fun :stuck_out_tongue:

ray-kraesig: Ain't that the truth. &#x1f522;&#x20e0;

src/utils/__tests__/async-test.js
ray-kraesig: Restored(?) to what I thought it was originally: `sleepMeasure` created in this commit, within the unit-test's `describe` block.

gnprice:
And merged. Thanks again!

Issue #3672, opened 2019-10-29
iOS app can send malformed notif tokens to server


gnprice:
This is a bug that appears to have been present for most of this year. There are no known user-visible symptoms; we spotted it in server records on zulipchat.com.

The symptoms seen on the server are:
* The `zerver_pushdevicetoken` table may have a number of records with `kind = 1` -- meaning APNs -- and `token = '[object Object]'`. Normally `token` should be a 64-digit hex string.
* The value '[object Object]' is of course bogus as an APNs device token. The server regularly learns this when talking to APNs, and deletes these records in the same way it would a device token that had expired. So there's only ever a few of these records, all recent.
* The deletion produces log messages like the following -- either to `/var/log/zulip/events_missedmessage_mobile_notifications.log` for a Zulip app server talking directly to APNs, or to `/var/log/zulip/server.log` for the notification bouncer service:
```
2019-10-29 00:28:46.536 INFO [zerver.lib.push_notifications] APNs: Removing invalid/expired token [object Object] (BadDeviceToken)
```

OTOH we continue to get new perfectly-valid tokens too, and notifications seem to work for people in practice, including for new users.

Still, we should debug this -- not least because whatever bug is causing this, there may be some users that it's causing notifications not to work for.

gnprice:
Some thoughts on how to investigate:

* In the app, we basically pass these tokens straight through to the server from whatever the `react-native-notifications` library hands to the callback we register as a `remoteNotificationsRegistered` listener.
  * (This is a rather muddled third-party library that I got rid of on Android around the beginning of this year, but that we still use on iOS.)
  * See the `NotificationListener` class in `src/notification/index.js`, particularly `handleDeviceToken` and its references and callees.

* So it seems like that object is sometimes the expected string, and sometimes an object value. One line of debugging would ask: when is it the latter -- and what's in the object when it is?

* That means poking around in client-side code for notifications. We have documentation [here](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/push-notifications.md#testing-client-side-changes-on-ios) on how to do that.
  * On Android this is very straightforward, but unfortunately on iOS Apple makes it complicated.

* This might be a good time to pursue the "dev certificate" approach sketched out in that doc.
  * This issue is an especially good fit for that approach, because the easiest and likely first form of that will be with a dev server... and this issue doesn't really benefit from using a prod server (no UI involved where realistic data helps), and on the other hand probably does benefit from using a dev server.

PR #3888, opened 2020-02-10
APNs: log invalid device tokens, rather than sending them

ray-kraesig:
APNs device tokens, as received through out iOS notifications library,
should be a hex string. Unfortunately, they're sometimes being sent to
the server as "[Object object]".

This is most likely due to the argument of `handleDeviceToken`
actually being an object, despite its typing as `string`, and being
improperly stringified by `encodeParamsForUrl`. However, it may
alternatively be the case that someone is mis-stringifying the object
before we get ahold of it, so log that case too.

This is work toward the resolution of #3672.

### DIFF
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -4,6 +4,7 @@ import NotificationsIOS from 'react-native-notifications';
 
 import type { Auth, Dispatch, Identity, Narrow, User } from '../types';
 import { topicNarrow, privateNarrow, groupNarrow } from '../utils/narrow';
+import type { JSONable } from '../utils/jsonable';
 import * as api from '../api';
 import * as logging from '../utils/logging';
 import {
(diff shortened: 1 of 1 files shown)


ray-kraesig:
:warning: This PR is untested, due to the difficulty (noted elsewhere) of testing iOS notifications code.

src/notification/index.js
```diff
@@ -209,9 +209,15 @@ export class NotificationListener {
...
-    await this.dispatch(sendAllPushToken());
+  handleDeviceToken = async (deviceToken: string | { ... }) => {
+    // A device token should normally be a string of hex digits. Sometimes,
+    // however, we appear to receive objects here. Log this. (See issue #3672.)
+    if (typeof deviceToken === 'string' && deviceToken !== '[Object object]') {
+      this.dispatch(gotPushToken(deviceToken));
+      await this.dispatch(sendAllPushToken());
+    } else {
+      logging.error('Received non-string device token', { deviceToken });
+    }
```
gnprice: Let's write this error path in the early-return pattern:
```js
  if (isBad(thing)) {
    logging.error('Thing is bad', { thing });
    return;
  }

  processNormally(thing);
```

This keeps the code that handles each error next to the code that tests for it. A related effect is that it helps the reader to study the behavior in a given error case, or the behavior in the "normal"/happy case, or to understand the full behavior by divide-and-conquer.

Doesn't hugely matter for a function this short with just one error case. But it's good to have the pattern consistent for if the function grows, or if another contributor copies the pattern in another context. And even at this scale I think it already helps.

ray-kraesig: (I actually did write it that way in the first draft; I switched to the above because the negated logic seemed more straightforward.)

ray-kraesig:
Done. Also, more honesty about being dishonest about types.

gnprice:
Thanks -- merged!

src/notification/index.js
gnprice: Cool. Yeah, I think the main thing that can make this nested style appealing is that it lets you write down the things that are true in the happy path, rather than the things that aren't.

But then I think the effect on the code's overall structure generally ends up outweighing that, even at this scale.

Thanks for the revision!

Issue #3764, opened 2020-01-07
Convert from emails to user IDs to identify users


gnprice:
This is a meta-issue which covers a large number of independent tasks. Filing it in order to have an identifier (ha) we can use to refer to the overall effort.

In general, Zulip originally used a user's email address as the primary way of identifying the user, throughout the API and the client data structures. This isn't good -- among other reasons, because people change their email addresses. So, years ago, we shifted toward using numeric user IDs. This was a slow migration, but the server and the webapp are now pretty far along; just about the whole API can now operate in terms of user IDs.

In the mobile app, though, we still pass around emails to identify users in lots of places. This often complicates the code, as different layers look up the email for a user ID and vice versa in order to communicate with other layers; it also leads to bugs when a user changes their email address. So we should fix it.

In general, this means:
* Almost everywhere that we pass an email in an API request, it should instead be a user ID.
  * If the API route doesn't accept user IDs, there should be an issue for that on zulip/zulip and a comment linking to that next to our code that passes an email.
* Almost everywhere that a React component takes a prop like `email`, it should instead be a `userId`.
  * Possibly everywhere. Anyplace where we have a good reason to take an `email` prop -- other than the boring reason that we haven't gotten around to converting it yet -- there should be a comment explaining why.
* Almost everywhere that we have an object or `Map` data structure where the keys are emails, they should instead be user IDs.
  * The main exception is the data structures, like the one returned by `getAllUsersByEmail`, that we use only within code we haven't converted yet. Once other code is all converted, those should disappear too, or be used only in rare exceptional spots.

Tasks for specific instances of this migration include #3501 and #3525.

agrawal-d:
Hi!
I'm going to work on migrating `navigateToAccountDetails` and `getUserIsActive` to user IDs from emails, to make pull #3878 mergeable, as per https://github.com/zulip/zulip-mobile/pull/3878#issuecomment-589911019.

gnprice:
As an update to see (most of) what remains for this task, I tried commenting out the `email` property on `User`:
```diff
 export type User = $ReadOnly<{|
   user_id: UserId,
-  email: string,
+  // email: string,
```
and seeing where Flow reported we're still using it.

- In several places we still use emails because old server versions didn't accept user IDs:
  - get-messages
  - sending PMs
  - subscribing other users to streams
  - sending typing status

  I think those are all old enough now that we can drop that logic.

[...]

PR #3889, opened 2020-02-10
narrow [nfc]: Cut `Narrow` from the `Outbox` type.

gnprice:
Our `Narrow` type is kind of a mess: it's defined as just an array of operator/operand string pairs, but in practice most of our code assumes it has a lot more structure than that. There's also a lot of scattered places where we parse that structure, and so express a variety of those assumptions. Separately, it's a major source of #3764 in that it identifies users by email, not ID, for PM threads (as well as streams by name, not ID); and the many scattered places we parse its implicit structure are each obstacles to fixing that.

I spent some time last week sketching out what it would look like to migrate `Narrow` to something that would use IDs rather than emails or stream names, and also be more explicitly structured so that e.g. after conditioning on being a PM conversation we'd have something like `userIds: number[]` rather than having to parse a string. I found it actually seems surprisingly doable, and I have a rough draft in a branch.

One piece is that there are places where we just shouldn't be using a "narrow" value at all, because we're not actually talking about a feed of messages or a possible message list -- we're talking about one message and who it goes to. The cleanest way to migrate those is to just get them off `Narrow` entirely. One of those places is a property on `Outbox`; this branch fixes that.

### DIFF
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -163,7 +163,6 @@ export const addToOutbox = (narrow: Narrow, content: string) => async (
   const localTime = Math.round(new Date().getTime() / 1000);
   dispatch(
     messageSendStart({
-      narrow,
       isSent: false,
       ...extractTypeToAndSubjectFromNarrow(narrow, getUsersByEmail(state), userDetail),
       markdownContent: content,
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -64,17 +64,13 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
         return; // i.e., continue
       }
 
-      const to = ((): string => {
-        const { narrow } = item;
-        // TODO: can this test be `if (item.type === private)`?
-        if (isPrivateOrGroupNarrow(narrow)) {
-          return narrow[0].operand;
-        } else {
-          // HACK: the server attempts to interpret this argument as JSON, then
-          // CSV, then a literal. To avoid misparsing, always use JSON.
-          return JSON.stringify([item.display_recipient]);
-        }
-      })();
+      // prettier-ignore
+      const to =
+        item.type === 'private'
+          ? item.narrow[0].operand
+            // HACK: the server attempts to interpret this argument as JSON, then
+            // CSV, then a literal. To avoid misparsing, always use JSON.
+          : JSON.stringify([item.display_recipient]);
 
       await api.sendMessage(auth, {
         type: item.type,
--- a/src/outbox/outboxActions.js
+++ b/src/outbox/outboxActions.js
@@ -67,7 +67,10 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
       // prettier-ignore
       const to =
         item.type === 'private'
-          ? item.narrow[0].operand
+            // This will include the self user, possibly twice.  That's
+            // fine; on send, the server (since at least 2013) drops dupes
+            // and normalizes whether to include the sender.
+          ? item.display_recipient.map(r => r.email).join(',')
             // HACK: the server attempts to interpret this argument as JSON, then
             // CSV, then a literal. To avoid misparsing, always use JSON.
           : JSON.stringify([item.display_recipient]);
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -248,19 +248,24 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
     search: () => false,
   });
 
-/** True just if `haystack` contains all possible messages in `needle`. */
-export const narrowContains = (haystack: Narrow, needle: Narrow): boolean => {
-  if (isHomeNarrow(haystack)) {
-    return true;
-  }
-  if (isAllPrivateNarrow(haystack) && isPrivateOrGroupNarrow(needle)) {
-    return true;
-  }
-  if (isStreamNarrow(haystack) && needle[0].operand === haystack[0].operand) {
-    return true;
-  }
-  return JSON.stringify(needle) === JSON.stringify(haystack);
-};
+export const narrowContainsOutbox = (haystack: Narrow, needle: Outbox): boolean =>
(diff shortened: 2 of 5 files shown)


src/outbox/outboxActions.js
```diff
@@ -67,7 +67,10 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
       // prettier-ignore
       const to =
         item.type === 'private'
-          ? item.narrow[0].operand
+            // This will include the self user, possibly twice.  That's
+            // fine; on send, the server (since at least 2013) drops dupes
+            // and normalizes whether to include the sender.
+          ? item.display_recipient.map(r => r.email).join(',')
```
ray-kraesig: I wouldn't want to just `.slice` off an element – but it does feel like the self-user should be `.filter`ed out here, rather than relying on the permissiveness of the server implementation.

Also, if these are emails rather than IDs, we should probably use JSON rather than CSV. (If I'd realized that's what they were, I would have done the additional testing on #3734. :disappointed:) I'm not about to ask you to do that for this PR, but tagging it with an inline `TODO:` would be appreciated.

src/utils/narrow.js
```diff
@@ -248,19 +248,24 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
-  return JSON.stringify(needle) === JSON.stringify(haystack);
-};
+export const narrowContainsOutbox = (haystack: Narrow, needle: Outbox): boolean =>
+  caseNarrowPartial(haystack, {
+    stream: name => needle.type === 'stream' && needle.display_recipient === name,
+    topic: (streamName, topic) =>
+      needle.type === 'stream'
+      && needle.display_recipient === streamName
+      && needle.subject === topic,
+    pm: emails => emails === needle.display_recipient.map(r => r.email).join(','),
```
ray-kraesig: This line (`pm: emails => ...`) doesn't look like it will work, for multiple reasons:
* It'll fail if `narrow[0].operand` and `display_recipient` are not in the same order. (In practice, we seem to frequently rely on serializations being consistent. :disappointed:)
* If one of your comments in a later commit is accurate – and, more questionably, if I'm understanding the code correctly –  the left-hand side will lack the self-user, while the right-hand side will have it.
* Both of the above bug(?)s are concealed by the fact that there's no `groupPm` case in this `caseNarrowPartial`. (I'd suggest using `caseNarrow` here instead.)
* It's a `TypeError` when `needle.type === 'stream'`, as `"stream name".map is not a function`.

gnprice: Eek, good catch re: missing the `groupPm` case!

I suspect what happened there is that this change (and the others in this branch) was originally taken from a longer `Narrow`-refactoring series where it came after a refactor to unify the `pm` and `groupPm` cases. I have a couple of longer `Narrow`-refactoring series as drafts, and I'm pretty sure I made that unification somewhere in there. Definitely `caseNarrow`, the total version, is the right one to use here for just that sort of reason; I have no recollection of what I might have been thinking when I wrote `caseNarrowPartial` instead.


> * It'll fail if `narrow[0].operand` and `display_recipient` are not in the same order. (In practice, we seem to frequently rely on serializations being consistent. disappointed)

Yeah, and I believe that's already the case in the existing version of this code, as a result of just string-comparing the results of `JSON.stringify`. Given that (which I'll re-check empirically, because although I think I did so originally, it's been some months since then and I'm not certain I did), I'm OK with preserving that behavior in the course of this basically orthogonal refactor. Probably deserves a comment about it, though.


> * If … the left-hand side will lack the self-user, while the right-hand side will have it.
> * It's a `TypeError` when `needle.type === 'stream'`, as `"stream name".map is not a function`.

Hmm. Will check on these. And probably write some unit tests for this function, as I seem to have demonstrated the need for them >_>

gnprice: > `JSON.stringify` … which I'll re-check empirically

Ah, no, here it is more directly:
* In the old code, we're comparing the `JSON.stringify` of each narrow, which means the sequences of operators and operands must be equal. In particular, `narrow[0].operand` must be equal.
* The `display_recipient` on an outbox message is constructed from the same `Narrow` value (passed to `addToOutbox`) that in the old code was included as a property. A quick look at the private helpers of `addToOutbox` shows that it's computed like so:
```js
  narrow[0].operand
    .split(',')
    .map(item => {
      const user = usersByEmail.get(item) || NULL_USER;
      return { email: item, id: user.user_id, full_name: user.full_name };
    })
    .concat({ email: selfDetail.email, id: selfDetail.user_id, full_name: selfDetail.full_name });
```
So yep, we've been counting on the order of these.

I think that same look at the code also confirms that the display_recipient will have the self user added at the end, like you said, while the narrow's list of emails won't.

chrisbobbe:
@gnprice, would you like me to look closely at this one?

gnprice:
Not yet, thanks -- still needs some further revisions. I'll be sure to post a comment on this thread when it's ready for another look.

gnprice:
I've pushed the first commit of this branch, which is a pure jsdoc change, as 31ecb8d5b.

(Still working on the more substantial changes.)

gnprice:
Closing this in favor of #4299 . (And previously #4294, which also grew out of this.) That does all the same things more solidly, with a number of other changes.

src/utils/narrow.js
gnprice: In #4299 I ended up dropping this implementation entirely... because there's already an `isMessageInNarrow`! Which I noticed partly as a consequence of https://github.com/zulip/zulip-mobile/pull/3889#discussion_r377400072 above. That function needed some significant work to become well tested, and then that testing turned up some bugs which needed fixing; those changes were #4294, and #4299 returns to the outbox code to switch to using that.

src/outbox/outboxActions.js
gnprice: In #4299 I ended up taking care of that duplicate self-user by fixing how we construct `display_recipient` in the first place, elsewhere in this file. Having it wrong there turns out to cause its own buggy behavior, before any request even reaches the server.

PR #3892, opened 2020-02-11
webview: Fix message list is scrolled when unread notice hides.

jainkuniya:
On webview resize listner, do not scroll if difference is less than 100.
This listner is only meant to sroll list when keyboard pops in or out,
and for such event difference will be > 100.

### DIFF
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -139,7 +139,7 @@ var compiledWebviewJs = (function (exports) {
   window.addEventListener('resize', function (event) {
     var difference = viewportHeight - documentBody.clientHeight;
 
-    if (documentBody.scrollHeight !== documentBody.scrollTop + documentBody.clientHeight) {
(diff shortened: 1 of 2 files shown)


gnprice:
Hmm, interesting.

Can you say more about the problem this is intended to solve? Maybe in a chat thread for discussion, or else an issue. Please be sure to link to related issues, as well, so the background is readily available for everyone.

The existing behavior definitely isn't great, as it feels like the screen jumps around... but with this change it still feels to me like it jumps around. That doesn't make me excited to add a hack like this.

And I don't feel like I understand yet what's actually happening -- for example, what's the exact sequence of `resize` events we're getting and `window.scrollBy` calls we're making. It should be totally doable to find that out. Then we can discuss what the sequence should instead be, and then what the logic in the code should be that would get that sequence.

In short, this problem is a good one to try to solve, but it's still at the stage where what's needed is debugging, not code. It's not ripe for a PR until we've done that debugging (or perhaps some design discussion).

PR #3893, opened 2020-02-11
compose box: Show message when PM recipient(s) have been deactivated.

### LINKED ISSUE #3812: Dont show compose box in deactivated user PM screen
It is not possible to send PMs to deactivated users, and the compose box breaks in the PM screen for such users.

Instead, show a message in a from similar to :
image

agrawal-d:
Fixes #3812.

For PMs
image

For Group PMs
image


ray-kraesig:
I'm going to pass this one off to @gnprice, as he's been looking at the compose-box more recently.

src/compose/ComposeBox.js
```diff
@@ -423,6 +448,22 @@ class ComposeBox extends PureComponent<Props, State> {
...
+  }
+  const recipientEmail = narrow[0].operand;
+  return !getUserIsActive(state, recipientEmail);
+};
+
+const getGroupPmrecipientIsDeactivated = (state: GlobalState, narrow: Narrow): boolean => {
+  if (!isGroupNarrow(narrow)) {
+    return false;
+  }
+  const recipientEmails: string[] = narrow[0].operand.split(',');
```
agrawal-d: I considered using the selector `getRecipientsInGroupNarrow` here, but it would create un-necessary User objects when we only require emails.

chrisbobbe: In `getRecipientsInGroupNarrow`, it looks like various containers for the `User` objects (arrays, and a `Map`) do get re-created (though not the `User` objects themselves, I think). I guess the performance cost is linear with the number of (total?) users, and we do avoid that concern by inspecting the operand here. Some caching is done in `getRecipientsInGroupNarrow`; that may help performance if we were to use that.

One thing about doing `narrow[0].operand.split(',');` here is that it's dealing directly with the raw data, so this code would need to be revisited if the shape of that data changed. Not the end of the world, but I like to reach first for reusable functions that do that work, if possible. I wonder if `emailsOfGroupNarrow` might be appropriate, if `getRecipientsInGroupNarrow` has too high a performance cost? @gnprice, maybe you have some thoughts? 🙂

src/compose/ComposeBox.js
```diff
@@ -357,6 +376,12 @@ class ComposeBox extends PureComponent<Props, State> {
 
     return (
       <View style={this.styles.wrapper}>
+        {showDeactivatedGroupPMWarning && (
+          <Label
+            text="One or more recipients may have been deactivated"
```
chrisbobbe, line 381:
```diff
- "One or more recipients may have been deactivated"
+ "One or more recipients have been deactivated"
```
I think? It's not uncertain: we know that at least one has been deactivated.

src/compose/ComposeBox.js
```diff
@@ -328,6 +338,11 @@ class ComposeBox extends PureComponent<Props, State> {
       ...this.inputMarginPadding,
       ...this.context.styles.backgroundColor,
     },
+    deactivatedGroupPMWarning: {
+      padding: 10,
+      backgroundColor: 'hsl(45, 100%, 51%)',
```
chrisbobbe, line 343: Could you say more about how you arrived at these numbers? I wonder if there's a constant we can use for the `backgroundColor`, and if it might be slightly more readable to say `'black'` instead of `'hsl(0, 0%, 0%)'`.

src/compose/ComposeBox.js
```diff
@@ -423,6 +448,22 @@ class ComposeBox extends PureComponent<Props, State> {
...
 
+const getPmrecipientIsDeactivated = (state: GlobalState, narrow: Narrow): boolean => {
+  if (!isPrivateNarrow(narrow)) {
+    return false;
+  }
+  const recipientEmail = narrow[0].operand;
+  return !getUserIsActive(state, recipientEmail);
+};
+
+const getGroupPmrecipientIsDeactivated = (state: GlobalState, narrow: Narrow): boolean => {
```
chrisbobbe, line 459: Nit: I'd capitalize `Recipient` in `getPmRecipientIsDeactivated` and `getGroupPmRecipientIsDeactivated`

src/compose/ComposeBox.js
agrawal-d: @chrisbobbe this is moot now, since `getUserIsActive` now accepts user IDs only.

src/compose/ComposeBox.js
```diff
@@ -434,4 +475,6 @@ export default connect<SelectorProps, _, _>((state, props) => ({
   editMessage: getSession(state).editMessage,
   draft: getDraftForNarrow(state, props.narrow),
   lastMessageTopic: getLastMessageTopic(state, props.narrow),
+  showDeactivatedPMWarning: getPmrecipientIsDeactivated(state, props.narrow),
+  showDeactivatedGroupPMWarning: getGroupPmrecipientIsDeactivated(state, props.narrow),
```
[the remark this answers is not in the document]
agrawal-d: Yeah, this sounds better.

src/compose/ComposeBox.js
agrawal-d: Fixed.

src/compose/ComposeBox.js
agrawal-d: IIRC it was google material orange colour. But i've removed it now, since I realized that its not possible to send a message in a group PM where even one member has been deactivated.

agrawal-d:
Thanks for the review @chrisbobbe. I've made the changes you suggested.
One thing I realized now is that you cant send a message to a group PM where even a single recipient has been deactivated. So, I've changed the PR so that instead of simply showing a warning, the compose box is entirely removed.

src/compose/ComposeBox.js
agrawal-d: Yep, fixed.

chrisbobbe:
I'm curious about the taxi from the second commit! 😛 https://github.com/zulip/zulip-mobile/blob/411a2e09183d85cd58eb80b066390d4dee398ae8/static/img/taxi.png

src/utils/narrow.js
```diff
@@ -182,6 +182,12 @@ export const isGroupNarrow = (narrow?: Narrow): boolean =>
 export const emailsOfGroupNarrow = (narrow: Narrow): string[] =>
   caseNarrowPartial(narrow, { groupPm: emails => emails });
 
+/**
+ * The recipients' email if a Private narrow; else error.
```
chrisbobbe, line 186: Nit: "recipient's" (apostrophe before the "s"); if I understand correctly, this is one person's email.

src/message/GroupPmRecipientDeactivated.js
```diff
@@ -0,0 +1,21 @@
...
+import { Label } from '../common';
+import styles from '../styles';
+
+class GroupPmRecipientDeactivated extends PureComponent<{||}> {
+  render() {
+    return (
+      <View style={styles.disabledComposeBox}>
+        <Label
+          style={styles.disabledComposeText}
+          text="You can't post a message because one or more recipient accounts have been deactivated"
```
chrisbobbe, line 14: nit: "recipients' accounts", I think.

chrisbobbe:
This looks good! I've made a few small comments above. 🚕 

One other thing I noticed this time is that the `getPmRecipientIsDeactivated` and `getGroupPmRecipientIsDeactivated` selectors are defined in the same file they're used, instead of living in one of our many files dedicated to selectors.

I think this is probably okay; they're quite specialized, and I can't immediately think of which selectors file they would go in. We can always move them to one of those files if we end up needing them in multiple places. @gnprice, thoughts?

agrawal-d:
> I'm curious about the taxi from the second commit!  https://github.com/zulip/zulip-mobile/blob/411a2e09183d85cd58eb80b066390d4dee398ae8/static/img/taxi.png

:smile: :taxi: Oops! I reall I drew this using `mtPaint` some time ago - how it ended up in the `static/img` folder **and** got commited, I have no idea! :rofl: :fearful: 

image

(btw, `mtPaint` is a really fun, free software)

agrawal-d:
Thanks for the review @chrisbobbe! I've made the changes.

Issue #3895, opened 2020-02-12
Usable and used languages should be (better) synchronized


ray-kraesig:
At present, we have mostly-complete translations for Finnish, Korean, and a few other languages that aren't present in the language selection screen. We also have some currently-unusable translations being loaded in `messages.js`, along with their locales in `locale.js`.

The sets of usable, selectable, and imported languages should be synchronized; and, more generally, it should be easy to *keep* them synchronized in the future.

gnprice:
Indeed!

#3896 is merged; and see the WIP PR #3897, which will do a long-overdue sync. Even after that, it will be useful to have more automation so we more easily do better at keeping these synced in the future. (It won't be *all* automation, though; there are a few human choices to make, mainly in the best names to use for some languages.)

See also the bit of docs added in 6775b4f44b35e70d03a93307ccd63af69b8c1d86.

gnprice:
#3897 has been merged, and #3959 which took another step to get these better in sync.

#3915 is an open PR to automate keeping them in sync for the future.

PR #3896, opened 2020-02-12
i18n [nfc]: add all known languages to the translation database

ray-kraesig:
There are a number of currently-unselectable languages in our
translations directory. We don't want to hook them all up
automatically, as several of them are incomplete; but we should have
the language names themselves available for translation in advance of
doing so.

This is work towards #3895.

The new list of languages was generated by the following script:

```bash
  curl https://pkgstore.datahub.io/core/language-codes/language-codes_json/data/97607046542b532c395cf83df5185246/language-codes_json.json > langcodes.json;
  (
    CODES=$(
      ls static/translations/messages_* |
      # strip sublocales
      perl -ne '/messages_(..)/ or die; print "$1\n"'
    );
    for code in $CODES; do
      jq "(.[] | select(.alpha2==\"$code\")).English | sub(\"[;,].*\"; \"\")" langcodes.json;
    done |
      # reinject sublocales (a)
      sed 's/Chinese/Chinese (Simplified)/' |
      sed 's/English/English (U.S.)/' ;

    # reinject sublocales (b)
    echo '"English (U.K.)"'
    echo '"Chinese (Traditional)"'

    # in use until en-GB is hooked up; reserved thereafter
    echo '"English"'
  ) | sort | uniq | perl -pe 'chomp; $_="  $_: $_,\n"'
```


gnprice:
Thanks @ray-kraesig ! Merged.

PR #3897, opened 2020-02-12
Update list of offered languages

gnprice:
This is part of #3895; this PR will update the current list to what it should be.

Doesn't yet work, I believe only because `locale.js` and `messages.js` aren't updated. Because @ray-kraesig is preparing a script to update those, I figure I'll hold off for a bit for that to be ready and save some effort. :)

See also the series of recent commits ending at 6775b4f44b35e70d03a93307ccd63af69b8c1d86, which documents this list and where the nontrivial parts of its data come from.

### DIFF
--- a/docs/howto/translations.md
+++ b/docs/howto/translations.md
@@ -1,5 +1,16 @@
 # Managing translations
 
+A person using the Zulip app can choose from a large number of
+languages for the app to present its UI in.
+
+Within the running app, we use a library `react-intl` to get the
+appropriate translation for a given string ("message") used in the UI.
+
+To manage the set of UI messages and translations for them, and
+provide a nice workflow for people to contribute translations, we use
+(along with the rest of the Zulip project) a service called Transifex.
+
+
 ## Maintainers: syncing to/from Transifex
 
 ### Setup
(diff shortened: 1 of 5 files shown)


gnprice:
In particular this makes it possible to select Finnish, Galician, Korean, Lithuanian, Persian, Romanian, Swedish, and Ukrainian as the UI languages, which it wasn't before. Some of those have near-100% translations, with everything translated but a small number of recently-added (or -edited) strings.

gnprice:
OK -- this now works! Thanks @ray-kraesig for providing some of the updates (as described in the relevant commit message), based on the script you've recently drafted for automating this :slightly_smiling_face: 

In addition to the two files mentioned above, Ray's branch also reminded me of the third nearly-mechanical file that needed to be updated, namely `flow-typed/translations.js`.

This branch now also adds docs on all the follow-up steps involved, as well as some background and context on how our translations work.

gnprice:
One task this leaves open for a followup is adding `zh-Hant` aka Chinese (Traditional), and renaming Chinese (Simplified) from its current internal name of `zh` to the more specific `zh-Hans`. Chinese (Traditional) is at 14% translated, well above the 5% threshold where we offer a translation to the user, so there may be users who would notice and appreciate that change (in addition to its value for making these names a bit less confused within our code.)

The reason I don't include that in this branch is, as mentioned in a commit message: it'll need a migration for existing users who have `zh` in their settings, so it's not as completely trivial as the rest of these changes.

chrisbobbe:
@gnprice I know it's a top priority to get some support for these languages out soon, so don't let this block progress — but you may already know that @ray-kraesig's https://github.com/zulip/zulip-mobile/pull/3915, designed to subsume this one, includes adding zh-Hant and that migration you mentioned, and it looks good to me. One possible option is to merge and release everything on Ray's branch except the automation script, if you'd prefer to have more stress-testing done on that.

gnprice:
Merged, after chatting with @ray-kraesig just now.

He's also going to extract from #3915 a commit with the `zh-Hans` migration so we can go ahead and add `zh-Hant`.

PR #3900, opened 2020-02-13
no-messages [nfc]: Clean up "fetching" selectors and "No Messages" UI logic.

gnprice:
I sat down this afternoon to look at #3822 and #3802, and found I wasn't 100% sure under what conditions we show that "No Messages" message, and that it wasn't immediate to find out.

I dug into the code to answer that question, and this series of changes is the result. :slightly_smiling_face: They're all NFC refactors, and each step is simple; at the end, I think it's appreciably easier to reason about what's going on in this part of our UI.

### DIFF
--- a/src/chat/Chat.js
+++ b/src/chat/Chat.js
@@ -16,7 +16,7 @@ import { getShownMessagesForNarrow } from './narrowsSelectors';
 
 type SelectorProps = {|
   fetching: Fetching,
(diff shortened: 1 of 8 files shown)


chrisbobbe:
Brilliant! It's so nice to be able to clear the path for others, and ourselves, by clarifying things, like this. Merged.

Issue #3899, opened 2020-02-12
Debug AnimatedComponent getting stuck with rapid changes in state


chrisbobbe:
Good animations can go a long way to smooth out rough edges, and currently we're being blocked from using more of them.

876f22d1aec9580ac053849b2c0ffaf1891b99fb reverted the use of AnimatedComponent for the OfflineNotice because of this issue:

> When disconnecting from a VPN on Android 9 and later, the connection state goes from 'vpn' to 'none' to 'wifi' very quickly. This can result in the online banner getting stuck in mid-animation, until some other event (e.g. the user scrolling) causes a redraw.

I haven't had time to reproduce it, yet, but we'll need a consistent reproduction before we can continue work on this. I'd also like to use AnimatedComponent for a loading banner in https://github.com/zulip/zulip-mobile/issues/3387.

ray-kraesig:
> > When disconnecting from a VPN on Android 9 and later, the connection state goes from 'vpn' to 'none' to 'wifi' very quickly. This can result in the online banner getting stuck in mid-animation, until some other event (e.g. the user scrolling) causes a redraw.
> 
> I haven't had time to reproduce it, yet, but we'll need a consistent reproduction before we can continue using AnimatedComponent. I'd especially like to use it for a loading banner in #3387.

For the record, I originally tested by building a copy of TunProxy from a fork (raise-isayan/TunProxy, I think?), installing that onto a physical Pie device, and running a trivial HTTP proxy to forward onward to my local HTTP server.

Note that a physical Pie device was required! This bug did not occur in the emulator, possibly because the connection state transition sequence was different.

chrisbobbe:
It's been [proposed](https://github.com/zulip/zulip-mobile/issues/2795) to use ["Fluid Transitions for React Navigation"](https://github.com/fram-x/FluidTransitions). That project doesn't look to be actively maintained, and I'd prefer to stick with AnimatedComponent.

One thing its README.md does point out, [here](https://github.com/fram-x/FluidTransitions), is that Animated can't use the native driver for width and height values. That's covered in the Animated doc [here](https://reactnative.dev/docs/animations#caveats). It's not clear to me from that paragraph in Fluid Transitions whether Fluid Transitions uses the native driver and thus solves the problem any better than Animated.

Does the fact that Fluid Transitions says "The library is JavaScript only - no linking required" mean that it's impossible for it to use the native driver? Or could it access the native driver through an existing link set up by React Native?

[...]

Issue #2725, opened 2018-06-25
Conversation list should use cache if no connection


borisyankov:
If there is data for the conversation list, but trying to load users a loading indicator is shown.
Instead, it should show the previously loaded data.

chrisbobbe:
https://github.com/zulip/zulip-mobile/issues/3802 is part of this, but it doesn't include showing a loading indicator; PR https://github.com/zulip/zulip-mobile/pull/3901 is in progress for that.

gnprice:
I think in this issue the loading indicator is (was) the status quo, and it's asking to show the cached data.

We now do show the cached data, so this was done some time ago. And that's a problem :wink:, namely #3387, because we don't give any indication that we're loading new data and what's visible is stale.

chrisbobbe:
> We now do show the cached data, so this was done some time ago.

12 days ago, I think, in https://github.com/zulip/zulip-mobile/pull/3822.

gnprice:
Hmm, yeah, thanks for the correction! I guess I was misled by too-hastily reading #3387, which is a year old and asks for the loading indicator (but doesn't actually say whether the status quo was that we showed cached data, or prematurely-cleared data, or what -- just, we didn't indicate we were loading fresh data.)

PR #3901, opened 2020-02-13
Loading banner during /register

chrisbobbe:
This would pair really well with https://github.com/zulip/zulip-mobile/pull/3822, if they could both go out in the same release, but we'll see how it goes. 🙂

As mentioned in person, it would be awesome to see the banner animate its entrance/exit, but I don't think it's realistic that https://github.com/zulip/zulip-mobile/issues/3899 will be resolved before this goes out, which is fine. 🙂 

Marking as P1 because the issue is marked P1.

Fixes: #3025 
Fixes: #3387


chrisbobbe:
Just rebased onto master and fixed merge conflicts.

src/common/LoadingBanner.js
```diff
@@ -0,0 +1,92 @@
...
+class LoadingBanner extends PureComponent<Props> {
+  static contextType = ThemeContext;
+  context: ThemeColors;
+
+  render() {
+    const {
+      loading,
+      spinnerColor = 'default',
+      textColor = this.context.color,
+      backgroundColor,
```
gnprice: I like this way of expressing the default for `textColor`, which I think you've basically introduced to our codebase. Can we do the same thing for `backgroundColor`?

src/common/LoadingBanner.js
```diff
@@ -0,0 +1,92 @@
...
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
+
+/**
+ * Displays a notice that the app is connecting to the server.
+ * Rendered when session.loading is true (as of 2020/02, only
+ * during the /register request).
+ *
```
gnprice: I appreciate the date marker for the uncertainty of how this may have changed in the future. :-)

The strategy I generally prefer, though, is to go perhaps a bit further:
* Say less here: perhaps like "Display a notice that the app is connecting to the server, when appropriate."
  * The idea is to only say what's in the interface of this code -- what it promises, for other code to rely on -- and not the specifics of the implementation choices it makes in order to live up to that interface.
* Then for the implementation choices, make them as clear as possible *inside* the code.
  * If the reader wants to know the details of how this code tries to live up to the "when appropriate" promise, they'll want to look at the code in any case, and the code should make their questions as easy to answer as it can.
  * Sometimes this involves comments.

src/common/Screen.js
```diff
@@ -116,6 +120,7 @@ class Screen extends PureComponent<Props> {
           <ModalNavBar canGoBack={canGoBack} title={title} />
         )}
         <OfflineNotice />
+        {shouldShowLoadingBanner === true && <LoadingBanner />}
```
gnprice: nit: simplify to drop the `=== true` -- it's already boolean :slightly_smiling_face:

src/common/Screen.js
```diff
@@ -85,6 +87,7 @@ class Screen extends PureComponent<Props> {
     search: false,
     autoFocus: false,
     searchBarOnChange: (text: string) => {},
+    shouldShowLoadingBanner: true,
```
gnprice: Let's squash this commit with the one after it, which passes false for this in several places. That way we don't have an intermediate state where the banner appears in some places where it doesn't make sense.

Both changes are small, so that doesn't make it a lot to read. If this were a more complicated transition, there are other ways it could be split up -- e.g., add the option default-off, then add explicit opt-outs one by one or in groups of related callsites, then flip the default. But happily none of that is needed here.

src/nav/ChatNavBar.js
```diff
@@ -31,19 +32,46 @@ class ChatNavBar extends PureComponent<Props> {
...
+          /* eslint-disable react-native/no-inline-styles */
+          style={[
+            {
+              flexDirection: 'row',
+              height: NAVBAR_SIZE,
+              alignItems: 'center',
+            },
+            { backgroundColor },
+          ]}
+        >
```
gnprice: How about making this split in its own little NFC commit, and then adding the `LoadingBanner` after that? I think that'd make it easier to read, separately, the part that's just pulling these two pieces apart from each other and the part that's inserting the `LoadingBanner` (and the `spinnerColor` logic for it.)

chrisbobbe:
(Note to self: Mark a commit as fixing https://github.com/zulip/zulip-mobile/issues/2725.)

src/common/LoadingBanner.js
```diff
@@ -0,0 +1,92 @@
...
+      justifyContent: 'center',
+      alignItems: 'center',
+      backgroundColor: backgroundColor ?? this.context.backgroundColor,
+    };
+    return (
+      <View key={key} style={style}>
+        <View
+          /* eslint-disable react-native/no-inline-styles */
+          style={{ flexDirection: 'row', alignItems: 'center' }}
+        >
```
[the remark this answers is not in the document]
chrisbobbe: Ah, removing this inner `View` has no effect, so I'll remove it.

src/common/LoadingBanner.js
```diff
@@ -0,0 +1,92 @@
...
+          <View>
+            <LoadingIndicator size={14} color={spinnerColor} />
+          </View>
```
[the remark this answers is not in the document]
chrisbobbe: Ah, so it turns out that `LoadingIndicator`'s `View` has `flex: 1`. The wrapper `View` here prevents the LoadingIndicator's box from stretching out to occupy as much of the banner as it can, with "Connecting..." fitting on the right, like so:

image

src/nav/ModalNavBar.js
```diff
@@ -17,22 +17,26 @@ type Props = $ReadOnly<{|
...
     const { dispatch, canGoBack, title } = this.props;
     const textStyle = [
       styles.navTitle,
       canGoBack ? { marginRight: NAVBAR_SIZE } : { marginLeft: 16 },
     ];
 
     return (
-      <View style={[contextStyles.navBar]}>
+      <View
+        /* eslint-disable react-native/no-inline-styles */
```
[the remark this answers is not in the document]
chrisbobbe: Makes sense. I've added some more justification in the commit message that I hope makes sense.

src/common/Screen.js
chrisbobbe: Hmm, but it's an optional property of Props (`shouldShowLoadingBanner?: boolean`), so it could be `undefined`.

Flow says this:

```
Sketchy null check on boolean [1] which is potentially false. Perhaps you meant to check for null or undefined [1]?
(sketchy-null-bool)
```

chrisbobbe:
https://github.com/zulip/zulip-mobile/pull/3901#discussion_r384229204 and https://github.com/zulip/zulip-mobile/pull/3901#discussion_r384140387 came up while making these changes. 🙂

.eslintrc.yaml
```diff
@@ -106,7 +106,7 @@ rules:
 
   # React Native.  This plugin isn't included in the airbnb config, and
   # nothing is enabled by default, so we enable its rules.
-  react-native/no-inline-styles: error
+  react-native/no-inline-styles: off # Noisy and a bad guide to rerendering optimizations.
```
chrisbobbe: e2dc155335100efca9c8256c07e0f03f3520c1b3

I realized this morning that this may be more detail than is currently, and may be distracting. The `style` prop is only one prop, and there are plenty of other props that can be guilty of causing unnecessary rerenders, and this rule has never caught those.

One argument for keeping all this detail (to the extent that it's correct and readable) is that we may use other eslint rules, like https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md, that handle those other props, narrowing things down to a focus on the style prop.

gnprice:
> (Note to self: Mark a commit as fixing #2725.)

As I commented just now on #2725 , that issue was actually asking for what we do today; we had a loading indicator and not the cached data, and it asked to show the latter. So I've closed it as already-done.

src/nav/ModalNavBar.js
```diff
@@ -32,7 +32,18 @@ class ModalNavBar extends PureComponent<Props> {
...
-      <View style={[contextStyles.navBar]}>
+      <View
+        style={[
+          {
+            borderColor: 'hsla(0, 0%, 50%, 0.25)',
+            flexDirection: 'row',
+            height: NAVBAR_SIZE,
+            alignItems: 'center',
+            borderBottomWidth: 1,
+            backgroundColor: contextStyles.navBar.backgroundColor,
```
gnprice: This looks a bit funny because we're inlining `contextStyles.navBar`... but keep using one piece of `contextStyles.navBar` :-)

There's another style `contextStyles.backgroundColor`, which consists exactly of the same `backgroundColor` property found in `contextStyles.navBar`. So this and the other similar site could use that instead.

Then that also makes it possible to actually remove the `navBar` definition in this same commit. That makes the change a bit simpler to understand, because (a) the existing definition and the inlined versions are both present in the diff (in red `-` and green `+` respectively), and (b) it becomes clear this is inlining all the uses.

Still good to have a second, followup commit (just like this version of the branch does) to switch from `contextStyles.backgroundColor` to `this.context.backgroundColor` and the new context API.

src/session/sessionReducer.js
```diff
@@ -37,6 +37,11 @@ export type SessionState = {|
   isActive: boolean,
   isHydrated: boolean,
   lastNarrow: ?Narrow,
+
+  /**
+   * Whether the /register request is in progress; i.e., on startup, or on
+   * re-init following a dead event queue after 10 minutes of inactivity.
+   */
   loading: boolean,
   needsInitialFetch: boolean,
```
gnprice: nit: add blank line after `loading`

Otherwise it looks as if the comment might apply to all the next few properties.

src/session/sessionReducer.js
```diff
@@ -37,6 +37,11 @@ export type SessionState = {|
   isActive: boolean,
   isHydrated: boolean,
   lastNarrow: ?Narrow,
+
+  /**
+   * Whether the /register request is in progress; i.e., on startup, or on
+   * re-init following a dead event queue after 10 minutes of inactivity.
+   */
   loading: boolean,
```
gnprice: A good example for putting into the "summary line, then prose" style we discussed yesterday in the context of `ZulipVersion`. :slightly_smiling_face:

gnprice: branch-level nit: put this commit before the one that adds `LoadingBanner`, because this helps explain what the latter is doing

src/common/index.js
```diff
@@ -15,6 +15,7 @@ export { default as KeyboardAvoider } from './KeyboardAvoider';
 export { default as Label } from './Label';
 export { default as LineSeparator } from './LineSeparator';
 export { default as LoadingIndicator } from './LoadingIndicator';
+export { default as LoadingBanner } from './LoadingBanner';
```
gnprice: nit: should go in the same commit that adds the module.

src/common/Screen.js
gnprice: Hmm, that's odd -- that warning should be specific to a type like `boolean | void`, not `boolean`...

Ah, I see. It's because the type is written as optional in `Props`. Instead, when a value is supplied in `defaultProps`, that prop should be non-optional in `Props`.

IOW -- well, the context is that there's really two versions of what the type of "the props of this component" are:
* the version that applies to callers;
* the version that applies to the implementation, after `defaultProps` has had its say.

And then another way to say the above rule is that `Props` (or more exactly, the type argument to `Component` or `PureComponent`, which we conventionally call `Props`) describes the latter, inward-facing version of "the props of this component". Flow then automatically looks at `defaultProps` and uses it to figure out the former, outward-facing version. Here's the relevant bit of the Flow docs:
https://flow.org/en/docs/react/components/#toc-using-default-props

Then Flow should see that this `shouldShowLoadingBanner` local variable is indeed already boolean :slightly_smiling_face:

src/common/LoadingBanner.js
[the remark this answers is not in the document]
gnprice: Got it, that makes sense. Thanks!


> Perhaps `LoadingIndicator`'s outermost `View` should not have `flex: 1`, and that way, we could remove this puzzling `View`. But we'd need to review all uses of `LoadingIndicator` to make sure it doesn't break things elsewhere.

Yeah, agreed on both counts.

This smells to me like an indication that it doesn't make sense for `LoadingIndicator` to have `flex: 1` on its outermost component. That's a property that's really about how the *parent* gets laid out, and how the given component participates in that -- so when a given component has that at its top level, it becomes effectively part of its interface, and I don't see a good reason why that'd naturally be part of the interface of `LoadingIndicator`.

.eslintrc.yaml
```diff
@@ -106,7 +106,7 @@ rules:
...
   # react-native/no-unused-styles: error  # This is buggy on the `this.styles` pattern.
   # react-native/no-color-literals: error  # TODO eliminate these and enable.
```
gnprice: nit: the rule is already off by default, so can just delete the line; or make it like the two below it, with `error` but commented out.

gnprice: The next thing I notice here is that this means we're no longer using this ESLint plugin at all. :-) So should probably drop it entirely.

Looking back through the [list of all its rules](https://github.com/Intellicode/eslint-plugin-react-native/tree/master/docs/rules)... I think the only two we actually want are the two commented out below this one. Maybe we'd want to keep these commented-out config lines, as a reminder for a possible future where we try to bring one or both of those rules into use. But in the meantime, it'll be good to make `node_modules` a bit smaller, and `tools/test lint` perhaps a bit faster.

.eslintrc.yaml
gnprice: I think a good medium for the material that makes up most of this commit message would be a chat thread, on perhaps `#mobile-team`. That's a good place for putting a lot of detail, and for people to reply and discuss; and Zulip's search makes it pretty accessible later, much more accessible than e.g. a thread on GitHub.

Then for the commit message, I think a summary at the scale of a couple of paragraphs will be good. I think the key points are basically just:
* Some styles are most readable when inlined.
* You might think this is a performance issue, because we're making new style objects on each render. But the reality is we have tons of code that does that already, because the lint rule doesn't catch even simple variations. Moreover a lot of them are for fundamental reasons, like that the style depends on props, state, and context -- so there's not a clear path to turning this rule into something useful.

About the details: does it really matter whether the child is a `PureComponent`? I thought the `PureComponent` optimization was only a shallow check of `props` itself, so that it checks identity on each individual prop value. But perhaps that chat thread will be the best place to discuss :slightly_smiling_face:

src/common/Screen.js
gnprice: Yup! Quite handy.

I've also just pushed a version of this comment as f3694dce2.

chrisbobbe: Great, thanks!

src/nav/ModalNavBar.js
chrisbobbe: > There's another style `contextStyles.backgroundColor`, which consists exactly of the same `backgroundColor` property found in `contextStyles.navBar`. So this and the other similar site could use that instead.

Mm, I missed this in your previous review; I'll use this. It seems like it's actually spelled `contextStyles.backgroundColor.backgroundColor`, which is weird but doable until the next commit lands. 🙂

src/common/LoadingBanner.js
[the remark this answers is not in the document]
chrisbobbe: Makes sense; as mentioned in person, this will be an important fix, but outside this PR's scope.

src/common/LoadingBanner.js
chrisbobbe: > Then for the implementation choices, make them as clear as possible inside the code.

I think we landed on LoadingState being the appropriate place for this comment; I'll add one there.

src/unread/UnreadCards.js
```diff
@@ -49,11 +47,7 @@ class UnreadCards extends PureComponent<Props> {
     ];
 
     if (unreadStreamsAndTopics.length === 0 && conversations.length === 0) {
-      return isLoading ? (
-        <LoadingIndicator size={40} />
-      ) : (
-        <SearchEmptyState text="No unread messages" />
-      );
```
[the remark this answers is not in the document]
chrisbobbe: Yes and yes; in fact, #3025 is also fixed in that same commit so I've adjusted that too.

chrisbobbe:
> One other comment, applying to several commit messages: take a look through them for some long summary lines to shorten (and move details into the prose body as needed.)

Done.

chrisbobbe:
(I just force-pushed, but I'm not quite done with these changes; I'll comment when I am.)

chrisbobbe:
OK, this is ready for another review.

gnprice:
Thanks @chrisbobbe for the revisions! Merged, with a few small commit-message edits we just discussed.

I've also dropped the new commit that removes the eslint react-native plugin, because I realized that it doesn't yet remove it from package.json and yarn.lock :slightly_smiling_face: -- only from the eslintrc. That can be a good quick followup.

PR #3906, opened 2020-02-13
logging: Disregard null device token.

ray-kraesig:
This is normal on (at least) the Android emulator when running without
Google Play services. (Presumably this would also be true on physical
devices without notification services.)

### DIFF
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -210,11 +210,20 @@ export class NotificationListener {
     this.dispatch(narrowToNotification(data));
   };
 
-  /** Private. */
+  /**
+   * Private.
+   *
+   * @param deviceToken This should be a `?string`, but there's no typechecking
+   *   at the registration site to allow us to ensure it. As we've been burned
+   *   by unexpected types here before, we do the validation explicitly.
+   */
   handleDeviceToken = async (deviceToken: mixed) => {
-    // A device token should normally be a string of hex digits. Sometimes,
-    // however, we appear to receive objects here. Log this. (See issue #3672.)
-    if (typeof deviceToken !== 'string' || deviceToken === '[Object object]') {
+    // Null device tokens are known to occur (at least) on Android emulators
+    // without Google Play services, and have been reported in other scenarios.
+    // See https://stackoverflow.com/q/37517860 for relevant discussion.
+    //
+    // String device tokens are the normal and expected case.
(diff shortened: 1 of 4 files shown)


gnprice:
Thanks @ray-kraesig!

The commit message says "logging", but it looks like this actually bails out from the non-logging logic after that, too. Seems like that might be fine, given the token is null -- but I'd want to look through to be sure of that. Would you confirm that's the right thing, and update the commit message? (And/or the code, if a variation turns out to be better.)

ray-kraesig:
It bailed out mostly because the logging logic that it would currently have hit would also bail out. The only effective change there would have been to not log if the device token is null.

However, that _is_ a change from the pre-#3888 code's handling of `null`, and the ultimate behavior is subtly different... although not really in any end-user-visible way. I've restored it anyway.

(I'm weakly of the opinion that the right thing to do on `null` is to add a mostly-passive "notification" in Settings, warning the user that push-notifications couldn't be set up for some reason – but that implies a whole heap of questionable UI fiddling that I probably wouldn't believe to be worth the trouble yet even if I were *strongly* of that opinion.)

src/notification/index.js
```diff
@@ -211,16 +211,41 @@ export class NotificationListener {
...
+      }
+
+      // If we reach this point, `deviceToken` must be a string.
+      return (deviceToken: string);
+    };
+
+    const token: ?string = validateToken();
+    if (token === undefined) {
       return;
     }
```
gnprice: I'm afraid I find this way of writing the logic really twisty. :-/ This code is significant work for me to follow and be confident I understand; the logic it's ultimately implementing isn't complicated, so it doesn't seem like that should have to be the case.

One thing I think is contributing to that effect is the use of a closure, where this helper function takes `deviceToken` implicitly from an enclosing scope rather than explicitly as a parameter. Another is the sort of alternation in the control flow: first there's a path that accepts, then one that rejects, then one that accepts... and that combines with the accept/reject distinction being expressed in a non-obvious and implicit way.

How about we try some edits to this code while both looking at the same screen together sometime in the next day or two? That might be an effective way to discuss.

ray-kraesig:
Updated based on offline discussion. Notable functional changes include loss of the `[Object object]` check, which removes any use for additional circumlocution.

ray-kraesig:
This PR – or something close enough to it – was merged as part of #3910. Closing.

PR #3907, opened 2020-02-13
store [nfc]: Add a bunch of comments in store.js; refactor slightly.

gnprice:
This documents a bunch of facts about how we set up our Redux store, and the various third-party add-ons we have surrounding it, which weren't previously really written down.

It also refactors this code slightly for better organization and easier reading.

(The refactors are why I'm sending this as a PR rather than just pushing it to master, as I typically do for pure comments and docs changes.)


gnprice:
(Oops, I see I put this PR branch into the main repo rather than my fork. Well, the mess will be temporary; I normally delete my PR branches after merge in any case.)

ray-kraesig:
The documentation all looks good. I don't think moving the middleware list – and the mess of imports needed to create it – into `store.js` is an improvement, though.

Speaking of a mess of imports, `store.js` as it exists is a good candidate for the following transformation:
```flow
import { a, b } from 'file'
import type { T } from 'file'
```
                        ⇓
```flow
import { a, b, type T } from 'file'
```

src/boot/store.js
```diff
@@ -118,24 +133,104 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
...
+ * The config options to pass to redux-persist.
+ *
+ * See upstream docs:
+ *   https://github.com/rt2zz/redux-persist/tree/v4.10.2
+ *
+ * (Why v4?  We studied v5 and determined it doesn't make sense for us to
+ * use because its design choices on migrations don't support some important
+ * kinds of migrations.  See commit 081a09806, PR #2762, and this discussion:
+ *   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux-persist.20v5/near/624465
+ * all from 2018-08.)
```
chrisbobbe, line 220: Also e5409c578.

chrisbobbe: Ah, looks like https://github.com/zulip/zulip-mobile/commit/e5409c57833c9239b98482f4a8ff4a56c730f85d has all three of these references, so I bet we can just refer to https://github.com/zulip/zulip-mobile/commit/e5409c57833c9239b98482f4a8ff4a56c730f85d by itself.

chrisbobbe:
I just fixed the rebase conflicts but couldn't push to this branch (maybe you didn't check the "allow edits" box?). Would you mind checking out my own branch `gregs-pr-store-comments`, where my revision is, to check that I resolved the conflicts appropriately and make any other changes/recommendations as you find them?

PR #3908, opened 2020-02-14
Scroller in Account pick screen

sumukhah:
after_1_commit
(Hide scroller is implemented)
lastFix


ray-kraesig:
First: it's a little difficult to review this PR because there are about 10 commits in it. These need to be squashed into 2. (I almost said 1; but see below.) You may want to take a look at our [git documentation](https://zulip.readthedocs.io/en/latest/git/). (In particular, **@&zwj;gnprice** notes that [having a graphical client set up](https://zulip.readthedocs.io/en/latest/git/setup.html#get-a-graphical-client) can often prevent this sort of issue, by making the branch structure visible.)

Second, this PR has an additional change – making the entire Account Pick screen scrollable – that's not clearly indicated in the PR text. It's a good change, though, and would be accepted if it were separated out rather than being buried among several similar-looking commits.

Third, the new property on `Screen` is named `verticalScrollIndicatorEnabled`, but the underlying `ScrollView` prop is named `showsVerticalScrollIndicator`, but there's no indication given of why this rename is necessary or helpful. It's probably best to keep it aligned with the underlying prop names, otherwise.

sumukhah:
Thank you @ray-kraesig ,
I am a little unclear about 
> Second, this PR has an additional change – making the entire Account Pick screen scrollable – that's not clearly indicated in the PR text. It's a good change, though, and would be accepted if it were separated out rather than being buried among several similar-looking commits.

So if i need to do any change please tell me..

src/common/Screen.js
```diff
@@ -122,7 +122,7 @@ class Screen extends PureComponent<Props> {
           contentContainerStyle={[padding && styles.padding]}
         >
           <ScrollView
-            contentContainerStyle={[styles.flexed, centerContent && componentStyles.content, style]}
+            contentContainerStyle={[centerContent && componentStyles.content, style]}
```
ray-kraesig: Now that I can see this more clearly, I realize that it may not be a good idea. This would change the style for *all* `Screen`-using components, not just `AccountPickScreen`, and that may have unwanted effects on other screens – especially since this style was itself introduced as a bugfix in commit 9fa95a9b790ddbac5193f3cce42d9bdd5ef48c3e.

That bugfix seems like something of a hack itself, so removing it might actually be the right thing to do – but that would have to be explained and/or verified (preferably both).

sumukhah: Thank you @ray-kraesig ,
>We shouldn't hide the scrollbar, here or on other scrollable pages, as it removes important functionality:

That is a great point, i have created an alternative solution for that,
added some margin for "Add new account" button, so that it escapes from scroll bar overlay,

> This would change the style for all Screen-using components, not just AccountPickScreen, and that may have unwanted effects on other screens – especially since this style was itself introduced as a bugfix in commit 9fa95a9.

>That bugfix seems like something of a hack itself, so removing it might actually be the right thing to do – but that would have to be explained and/or verified (preferably both).

I have not modified Screen this time, as it may give side effects,
If you think  that code needs to be replaced then I will add a separate PR
ezgif-2-105859cc225a

src/account/AccountPickScreen.js
```diff
@@ -10,6 +11,12 @@ import { Centerer, ZulipButton, Logo, Screen, ViewPlaceholder } from '../common'
 import AccountList from './AccountList';
 import { navigateToRealmScreen, switchAccount, removeAccount } from '../actions';
 
+const styles = StyleSheet.create({
+  accountList: {
+    flex: 0,
+  },
+});
```
ray-kraesig: In general, performing the inverse of part of an aggregate operation before performing the aggregate operation itself is a good way to make code too intertwined and tangled to easily understand or modify. This is a last-resort kind of hack, and would need to be very well documented as such. (Also, I believe the correct value to prevent `flex` from being set  here would actually be the string `"unset"`. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/unset).)

The ideal fix here is probably to do what I hinted at last time, and remove `flexed` from `Screen` – except with documentation, analysis, and some sort of fix so that [this bug](https://github.com/zulip/zulip-mobile/commit/026814edd7f4a3a19b9486bd3c01b20f91c52cd0#r31336633) doesn't recur.

Failing that, though, I'm pretty sure it should still be possible to adjust the layout of `AccountPickScreen` and/or `AccountList` so that the button isn't pushed off the screen, even without CSS hackery.

src/account/AccountPickScreen.js
```diff
@@ -15,6 +15,9 @@ const styles = StyleSheet.create({
   accountList: {
     flex: 0,
   },
+  addAccountButton: {
+    margin: 5,
+  },
```
ray-kraesig: This would work, but it needs an explanatory comment.

sumukhah: Thank you @ray-kraesig ,
 from https://github.com/zulip/zulip-mobile/commit/b0730a0636862f09d195c33be09d47776e449080, 

>and some sort of fix so that this bug doesn't recur.

I have  tried implementing "nestedScrollEnabled", but it has not worked properly.( https://reactnative.dev/docs/scrollview )
 in this commit,in Screen component, children which has flatList/SectionList is wrapped in View,
if not they are wrapped in ScrollView.

sumukhah: Hello @ray-kraesig, can you review this commit?
if it is ok I will Squash them into 2 commits as you have suggested

PR #3911, opened 2020-02-18
search: Fix some error handling, and simplify a bit.

gnprice:
I was looking back at this code last week while studying #3868, which will make some other changes to it.  The logic was a bit complex to understand, but I didn't want to disturb the error-handling it's doing.  Then as I looked closer, though, I found that the error-handling behavior wasn't quite right!  Hence this series, which fixes that and also simplifies the code somewhat along the way.

This is ready for normal review, but I'd like to hold off actually merging in order to minimize conflicts for #3868.  Following review, I'll look at that one and merge only commits I think will make that PR easier on net to complete.  I can rebase the rest after that's landed.  (There's also one non-mergeable commit at the tip of this branch, just adding a small comment referring to how to rebase onto #3868.)

### DIFF
--- a/src/search/SearchMessagesScreen.js
+++ b/src/search/SearchMessagesScreen.js
@@ -53,8 +53,10 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
   };
 
   // Non-React state. See comment following.
-  lastIdSent: number = 1000;
+  // Invariant: lastIdSuccess <= lastIdReceived <= lastIdSent.
+  lastIdSuccess: number = 1000;
   lastIdReceived: number = 1000;
+  lastIdSent: number = 1000;
 
   // This component is less pure than it should be. The correct behavior here is
   // probably that, when props change, all outstanding asynchronous requests
@@ -71,6 +73,7 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
     if (query === '') {
       // The empty query can be resolved without a network call.
       this.lastIdReceived = id;
+      this.lastIdSuccess = id;
       this.setState({ messages: null, isFetching: false });
       return;
     }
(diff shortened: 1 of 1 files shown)


src/search/SearchMessagesScreen.js
```diff
@@ -49,6 +49,9 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
       0,
       false,
     );
+
+    // TODO: on rebasing atop #3868, fetchMessagesComplete etc. will go here
```
chrisbobbe: Hmm, I don't see anything called fetchMessagesComplete in the current PR or in #3868; where does this come from?

gnprice:
> One thing that may be relevant is that, apparently, we can't guarantee that these two updates will be batched. Some logging in a componentDidUpdate in this commit and the parent shows two separate updates, with the switch in ordering you've described. It's true that we expect all setState calls that are run in a React event handler to result in a single, batched update [...]

Hmm, interesting! Thanks for looking into that.


> I did track the call stack all the way to an onChangeText passed to a TextInput, in src/common/Input.js, so at least our updates happen as a result of code run in an event handler. But I think our local async/await logic [...]

Yeah. I think the consequence of this is that we're "in a React event handler" at the start of `handleQueryChange`, but only up to the point where we `await`.

It sounds like this means we're actually getting a `render` call with the intermediate state where one of these `setState` calls has happened and the other hasn't yet.

I think this means the change isn't actually NFC! ... But, the new behavior is better than the old behavior, where it differs. So the change is good but I should revise the commit message :-)

Specifically, the old behavior was that we could have intermediate state saying:
* We're done fetching. Here's the messages: (the old value for `messages`)

Which is untrue and misleading. In the new behavior, the possible intermediate state is instead:
* We're still fetching. But here's the messages we have: (the new value for `messages`)

That's much better; at worst, it's not saying things as confidently as it could.

And in the particular UI choices we have in `SearchMessagesCard`, these states would actually be communicated differently if it happens that this is the first query to get a result (or the first one after an empty query), so that the old `messages` is null. In that case the old behavior would be
* ("loading" animation) -> blank -> results

whereas the new behavior is
* ("loading" animation) -> results

and so in fact indistinguishable from the ideal, batched behavior.

src/search/SearchMessagesScreen.js
gnprice: Hmm, not sure. I may have just meant `messageFetchComplete`.

If not that, it would have been some kind of refactored version of that that I may have had in mind after our conversation about #3868. Fortunately this is a WIP note to myself (in this tip commit of the branch); I wouldn't want to merge something that elliptical :-)

chrisbobbe:
Great, I'm glad to be helpful! 🙂 And I'm glad the new behavior is better than the old, and indistinguishable from the ideal, batched behavior; I didn't follow through with that investigation.

src/search/SearchMessagesScreen.js
chrisbobbe: Haha, makes sense.

gnprice: Edited to `messageFetchComplete`. If I did have some refactor in mind, the "etc." can encompass that. :wink:

gnprice:
And updated!

chrisbobbe:
Cool! I like the new commit message. So, this PR looks good to me, but it looks like this is pending a review from @ray-kraesig as well, and we'll want to coordinate with #3868 as you've mentioned.

src/search/SearchMessagesScreen.js
```diff
@@ -109,6 +105,13 @@ class SearchMessagesScreen extends PureComponent<Props, State> {
     });
   };
 
+  // The real work to be done on a query is async.  This wrapper exists
+  // just to fire off `handleQueryChangeInner` without waiting for it.
+  // TODO do we even need this wrapper?
```
ray-kraesig: As written, I believe, the answer is "no". It may be that React Native will quietly *wait* on `Promise`s returned by event handlers (if only to funnel rejections through its usual error-reporting paths), but it certainly isn't going to *block* on them.

gnprice:
Merged, along with most of #3868 -- see https://github.com/zulip/zulip-mobile/pull/3868#issuecomment-625606453 .

PR #3912, opened 2020-02-18
Remove `connectFlowFixMe`

### LINKED ISSUE #3451: Fix type errors hidden by `connect`

ray-kraesig:
Remove the last use of `connectFlowFixMe` from the codebase, followed immediately by `connectFlowFixMe` itself.

Fixes #3451.

(This PR also includes, as mostly-unrelated cleanup in a final [nfc] commit, a handful of `a ? a : b` ⇒ `a ?? b` conversions. This can be omitted if desired.)

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -370,9 +370,7 @@ export default connect<SelectorProps, _, _>((state, props: OuterProps) => {
   return {
     backgroundData,
     initialScrollMessageId:
-      props.initialScrollMessageId !== undefined
-        ? props.initialScrollMessageId
-        : getFirstUnreadIdInNarrow(state, props.narrow),
+      props.initialScrollMessageId ?? getFirstUnreadIdInNarrow(state, props.narrow),
     fetching: props.fetching || getFetchingForNarrow(state, props.narrow),
     messages: props.messages || getShownMessagesForNarrow(state, props.narrow),
     renderedMessages: props.renderedMessages || getRenderedMessages(props.narrow)(state),
--- a/src/__tests__/exampleData.js
+++ b/src/__tests__/exampleData.js
@@ -52,7 +52,7 @@ const randString = () => randInt(2 ** 54).toString(36);
 
 const randUserId: () => number = makeUniqueRandInt('user IDs', 10000);
 const userOrBotProperties = ({ name: _name }) => {
-  const name = _name !== undefined ? _name : randString();
+  const name = _name ?? randString();
   const capsName = name.substring(0, 1).toUpperCase() + name.substring(1);
   return deepFreeze({
     avatar_url: `https://zulip.example.org/yo/avatar-${name}.png`,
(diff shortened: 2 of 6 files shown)


gnprice:
Thanks @ray-kraesig !

I'd prefer to avoid making the code around `ZulipStatusBar` more complex, if we can. It'd be good to use the same pattern as everywhere else, and to not have to worry (now, and through future changes) about whether the upstream libdef is working correctly for us or has started silently accepting some type errors.

I think in fact we can handle this remaining case by making the code simpler, instead. `getTitleBackgroundColor` is kind of a mess in a couple of ways:
* it's attempting to provide a cache, but the cache doesn't actually work -- it's a case of #3015
* also there isn't actually any expensive work to cache there -- well, there's a case of #3339, but that's easy to eliminate -- so there's no need to involve `createSelector` at all
* it's expressing a UI choice, not really anything about the data, so it most appropriately belongs in UI code and not in the data-fetching code that feeds it

Combining those points, the logic in the body of `getTitleBackgroundColor` (i.e. the last argument passed to `createSelector`) can be pulled out as a function that the `render` method of `ZulipStatusBar` calls directly. (Ditto at the other call site of `getTitleBackgroundColor`.) Then the problem is dissolved.

gnprice:
(Then as an unrelated followup to those simplifications of `getTitleBackgroundColor`, I think its two call sites don't really want to be invoking a shared helper to make this decision at all! They're already separately and differently overriding basically the one thing it contributes beyond "find the stream, read its `color` property"; namely, what value to default to when no stream. That then leaves the code making the UI choices entirely within the UI code, where it naturally belongs.)

src/webview/MessageList.js
```diff
@@ -370,9 +370,7 @@ export default connect<SelectorProps, _, _>((state, props: OuterProps) => {
   return {
     backgroundData,
     initialScrollMessageId:
-      props.initialScrollMessageId !== undefined
-        ? props.initialScrollMessageId
-        : getFirstUnreadIdInNarrow(state, props.narrow),
+      props.initialScrollMessageId ?? getFirstUnreadIdInNarrow(state, props.narrow),
```
gnprice: I went through checking the types at each of these spots to confirm the `??` form was equivalent... and there's this one case where it isn't! The type of this property on `props` is `initialScrollMessageId?: number | null`. So if `null` is passed, this form behaves differently.

The others look good, though, and I'd be glad to have this change for the rest of them.

gnprice:
OK, and just sent #3914, along the lines I sketched above!

That branch leaves open some of the cleanups mentioned above. I have some drafts of doing more of that; but #3914 is enough of it to resolve the type issue.

ray-kraesig:
That's reason enough to close this PR entirely. (The cleanups are out of scope, and can get a mini-PR of their own.)

gnprice:
Sounds good. Thanks!

Issue #3851, opened 2020-01-31
MessageList: Avoid entirely replacing html body if possible.


chrisbobbe:
Forking from https://github.com/zulip/zulip-mobile/issues/3026, which is otherwise resolved.

Currently, the entire html body in the WebView is replaced on any change to the message list (a new message is received, a new emoji reaction is made, a message is edited, etc.). While the largest possible symptom of this, the fresh fetching of images every time this happens, has been substantially remedied, by WebView's [added ability to cache](https://github.com/react-native-community/react-native-webview/releases/tag/v4.0.0), it would certainly be nice not to have to do this.

Incremental changes are possible; as @gnprice points out, we don't want to have to maintain exhaustive diff-checking logic. A good approach is probably to allow the entire replacement of a single message on any change to it. So if it's added, edited, reacted to, starred, etc., treat all these the same way -- just replace that message entirely.

sevmonster:
This functionality makes the mobile app really clunky to use. Any action results in a half second long delay as the DOM is re-rendered in its entirety, at least on my Pixel 6 Pro—especially since it keeps so many posts loaded on the page. After enough messages are sent the delay gets larger and larger as the posts stack up, and I assume this also takes up more memory. Images reloading constantly is also very painful, moreso when on a metered or slow connection.

I now know thankfully that the images reloading can be resolved by adjusting my proxy config to add caching headers, but it's still annoying that such a thing could happen in the first place.

PR #3097, opened 2018-11-01
[WIP] Followups on visible-messages algorithm (#3040)

gnprice:
This contains the two still-WIP followup cleanup commits I mentioned on #3040 .

PR #3913, opened 2020-02-19
WIP for #3851

chrisbobbe:
As mentioned today, I've done some refactoring to clarify and simplify, and added the diffing algorithm. I'd like to iron out wrinkles here before getting too far into the code that will respond to the content diff events, which will also necessitate a rewrite of some important scrolling logic.


chrisbobbe:
https://github.com/zulip/zulip-mobile/pull/3097 may be relevant for the DOM manipulation work, if only to be aware of potential merge conflicts. That PR makes some changes to how messageIds are associated with elements, which should be mostly orthogonal because our plan for the rest of #3851 involves working only with indexes of the children of a new <div /> we'll introduce.

chrisbobbe:
I've paused before rebasing and fixing conflicts because of https://github.com/zulip/zulip-mobile/commit/6d1d1df89c0a09f4ce947e06cd3fb3fe0f28c19d#r37482253.

chrisbobbe:
Today I took another stab at the DOM manipulation part, since I figured the choice to use the three inbound events introduced in this PR (insert, delete, replace) will probably be about the same after this PR is reviewed, even if their names and some details change.

One thing I ran into, though, is that we need to handle getting the initial set of messages onto the screen; our diffing is based on this.props and nextProps in shouldComponentUpdate, and both of those will always be defined. We could have a fourth event dedicated to rendering these multiple messages at the same time, or we could do some plumbing, being really careful of race conditions, to use our existing diff algorithm to produce an edit sequence from an empty array to the initial messages (i.e., a sequence of inserts). Either way, this seems like something we should decide before I keep working on the DOM manipulation.

chrisbobbe:
Following from our conversation yesterday:

`getFirstUnreadIdInNarrow` is making a check to the mute state of a message (specifically, a call to `shouldBeMuted`) that I believe is unnecessary. If the following analysis is correct, we know scrolling will fail on entering a stream narrow for the first time, where the stream is muted.

`getFirstUnreadIdInNarrow` already only considers the "shown" messages in a narrow, i.e., those returned by `getShownMessagesForNarrow`. `getShownMessagesForNarrow`'s filtering is entirely determined by a call to `shouldBeMuted`, passing the narrow in question. It's a little hard to read, but from what I can tell, the logic in that call is correct. (Even if it's not correct, it's equivalent to the filtering used to populate the MessageList; `getShownMessagesForNarrow` is also used for that.)

But then, `getFirstUnreadIdInNarrow` takes those correctly filtered results, and, through the single-call-site functions `findFirstUnread` and `isMessageRead`, makes an *additional* call to `shouldBeMuted` — with the difference that HOME_NARROW is always passed for `narrow`, instead of the narrow in question. It's like this call is faking that we're in the HOME_NARROW, and it seems unwise here.

When the narrow in question is a muted stream, that `shouldBeMuted` call returns true for all messages; this is consistent with the desire not to see messages from a muted stream when you're *actually* in the HOME_NARROW. That makes `isMessageRead` return true for all messages, `findFirstUnread` returns undefined, `getFirstUnreadIdInNarrow` returns null, so nextProps.initialScrollMessageId is null, and a `scroll-to-anchor` event is sent to the WebView with null for `scrollMessageId`.

chrisbobbe:
This still isn't quite ready for review (but of course all comments are welcome) and may not be until we've finished working on Sign in with Apple. But there were some merge conflicts, so I fixed those and pushed.

ray-kraesig:
Are there any fragments of this (that you think would be improvements even without the rest of the patchset) that could be reviewed and cherry-picked separately?

PR #3914, opened 2020-02-19
Fix last `connectFlowFixMe`; and delete it!

### LINKED ISSUE #3451: Fix type errors hidden by `connect`

gnprice:
Fixes #3451, as the final cleanup after most call sites were resolved in #3789.

The commit message on the most interesting step is:

---

status bar [nfc]: Fix connectFlowFixMe.

The thing that had made it gnarly to use our well-typed `connect` here was that we both took `backgroundColor` in as a prop of the outer component and then returned a `backgroundColor` from the `mapStateToProps` callback, to override the former as a prop on the inner component.

The same fact is also a source of complexity and surprise when trying to read the code as a human, because we're giving the same name to two slightly different things right next to each other.  Doubly so because those two things are a prop of a component passed to `connect` and the same-named prop of the resulting component -- and it's routine across the rest of our codebase that when a prop exists on both, it's passed straight through verbatim.

So, disentangle that by introducing a distinct name.  And then lift the fixme!

This whole "title background color" logic is still rather tanglier than I'd like: in how `narrow` and a forced `backgroundColor` are implicitly mutually exclusive as props to `ZulipStatusBar`; in the use of DEFAULT_TITLE_BACKGROUND_COLOR aka `'transparent'` both as a value with its own substantive meaning and as a sentinel meaning "no stream here"; in how this component and `ChatNavBar` share this bit of UI logic, but under the covers via a putative selector.  It'd be good to disentangle those.  But now at least we have types here.

### DIFF
--- a/src/subscriptions/subscriptionSelectors.js
+++ b/src/subscriptions/subscriptionSelectors.js
@@ -27,6 +27,11 @@ export const getSubscriptionsById: Selector<Map<number, Subscription>> = createS
     new Map(subscriptions.map(subscription => [subscription.stream_id, subscription])),
 );
 
+export const getSubscriptionsByName: Selector<Map<string, Subscription>> = createSelector(
+  getSubscriptions,
+  subscriptions => new Map(subscriptions.map(subscription => [subscription.name, subscription])),
+);
+
 export const getIsActiveStreamSubscribed: Selector<boolean, Narrow> = createSelector(
   (state, narrow) => narrow,
   state => getSubscriptions(state),
--- a/src/common/ZulipStatusBar.js
+++ b/src/common/ZulipStatusBar.js
@@ -24,13 +24,20 @@ export const getStatusBarStyle = (statusBarColor: string): BarStyle =>
     ? 'light-content'
     : 'dark-content';
 
-type Props = $ReadOnly<{
-  dispatch: Dispatch,
-  hidden: boolean,
+type SelectorProps = $ReadOnly<{|
+  safeAreaInsets: Dimensions,
   theme: ThemeName,
   backgroundColor: string,
-  safeAreaInsets: Dimensions,
   orientation: Orientation,
+|}>;
+
+type Props = $ReadOnly<{
+  backgroundColor?: string,
+  narrow?: Narrow,
+  hidden: boolean,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
 }>;
 
 /**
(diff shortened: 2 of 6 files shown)


src/subscriptions/subscriptionSelectors.js
```diff
@@ -27,6 +27,11 @@ export const getSubscriptionsById: Selector<Map<number, Subscription>> = createS
     new Map(subscriptions.map(subscription => [subscription.stream_id, subscription])),
 );
 
+export const getSubscriptionsByName: Selector<Map<string, Subscription>> = createSelector(
```
chrisbobbe: Is `name` (the stream name) a unique identifier for `Subscription`s? In subscriptionsReducer, it looks like `stream_id` is considered unique:

```
    case EVENT_SUBSCRIPTION:
      switch (action.op) {
        case 'add':
          return state.concat(
            action.subscriptions.filter(x => !state.find(y => x.stream_id === y.stream_id)),
          );
```

Perhaps they're both unique? The existence of the [`get_stream_id` endpoint](https://zulipchat.com/api/get-stream-id) suggests so. Maybe it would be useful to have code comments on certain API types about any unique identifiers they have.

chrisbobbe:
This looks mostly good to me; I just posted a comment with a question, above.

This is a partial rewrite of https://github.com/zulip/zulip-mobile/pull/3912, right, but there are still some nice fixes that I think Ray has in mind on that one.

src/subscriptions/subscriptionSelectors.js
gnprice: Yeah, they are. It'd be better to use the numeric ID, but that's a long-open ongoing migration. I just filed #3918 to describe it.

The one real disadvantage of stream names vs. IDs -- other than the disadvantages of having *two* different ways, which complicates the code -- is that names can change.

The reason this codepath is doing its lookup by name is ultimately that it has a `Narrow`, and our `Narrow` type uses emails for users (cf #3764) and names for streams. We should fix that, but it'll be rather a project in itself.

gnprice:
Thanks @chrisbobbe for the review! Merged.


> This is a partial rewrite of #3912, right, but there are still some nice fixes that I think Ray has in mind on that one.

Yeah, there's a commit which sweeps through to use `??` where applicable, which would be good to have. After your comment above, Ray closed #3912 and suggested he might continue that as its own small PR, which makes sense. (There's a tweak or two to be made from the version in #3912.)

PR #3915, opened 2020-02-19
Automate translation-database integration

ray-kraesig:
Add a script to simplify pulling translation data.

Closely related to #3897, which it may subsume and/or should be integrated into.


tools/generate-txn-import.js
```diff
@@ -0,0 +1,218 @@
...
+  /** The human-readable name of this script. */
+  const thisScript = path.relative(root, __filename).replace(/^\.\//, '');
+
+  // ===================================================================
+  // Data acquisition
+  // ===================================================================
+
+  /** All the locales for which we have translation databases. */
+  const rawLocales /* : string[] */ = await (async () => {
+    // $FlowFixMe: Flow doesn't know about promises.opendir
```
chrisbobbe: > // $FlowFixMe: Flow doesn't know about promises.opendir

Neither, apparently, does [Node 10](https://nodejs.org/docs/v10.0.0/api/fs.html#fs_fspromises_opendir_path_options), which I think is the latest version we explicitly require for anything (see 9f82dfb46bd66b30a97307783ba293f2deab1879).

So it might be nice to have a friendly error message if the version is less than [v12.12.0](https://nodejs.org/docs/v12.12.0/api/fs.html#fs_fspromises_opendir_path_options), or at least a comment saying that at least that version is required. I get this, currently:

```
$ node tools/generate-txn-import.js 
(node:83083) UnhandledPromiseRejectionWarning: TypeError: fs.promises.opendir is not a function
```

src/settings/languages.js
```diff
@@ -40,22 +40,30 @@ const languages: $ReadOnlyArray<Language> = [
...
+  { locale: 'fi', name: 'Finnish', nativeName: 'Suomi' },
   { locale: 'fr', name: 'French', nativeName: 'Français' },
   { locale: 'de', name: 'German', nativeName: 'Deutsch' },
+  { locale: 'gl', name: 'Galician', nativeName: 'Galego' },
   { locale: 'hi', name: 'Hindi', nativeName: 'हिन्दी' },
   { locale: 'hu', name: 'Hungarian', nativeName: 'Magyar' },
   { locale: 'id-ID', name: 'Indonesian', nativeName: 'Bahasa Indonesia' },
   { locale: 'it', name: 'Italian', nativeName: 'Italiano' },
   { locale: 'ja', name: 'Japanese', nativeName: '日本語' },
+  { locale: 'ko', name: 'Korean', nativeName: '한국어' },
```
chrisbobbe: https://github.com/zulip/zulip-mobile/pull/3897/commits/e5ac53299bda6eaa7e3bc94815e405794b36f8ff also adds Lithuanian and no others. (Your choice to leave out Lithuanian may have been intentional, as a consequence of using a particular threshold that Greg didn't use.)

ray-kraesig: Exactly that. Serbian and Swedish already fall below that arbitrary threshold, though, so it makes sense to lower it. (Done.)

tools/generate-txn-import.js
ray-kraesig: Fixed – but I still get a warning:
```
(node:8216) ExperimentalWarning: The fs.promises API is experimental
```
... but I'm also running a somewhat older version of Node 10. If this is an issue for other people I can swap in the relevant `sync` calls instead.

chrisbobbe: > I'm also running a somewhat older version of Node 10.

I'm puzzled that it would work on any version of Node 10; the docs at v12.12.0 say it was introduced in v12.12.0. 🤷‍♂

ray-kraesig:
(The following commit message is no longer in the set of PR commits (as the commit it was attached to was subsumed into #3897). It's got some scripts I used, though, so I'm stashing it here for now.)

````
languages: Add more languages to the language-selection screen.

Manually add all languages with more than (somewhat arbitrarily) 10
translated entries.

This was aided significantly by the following script, which prints
(among other things) the number of non-default entries in each
Transifex messages-database:

```bash
(for file in $(ls static/translations/messages_*); do
    echo $file: $(git diff --no-index --shortstat -- static/translations/messages_en.json $file);
done) | sort -k 5rn
```

... and also by the following script, which takes the language data
found at https://stackoverflow.com/a/20623472 (as "./langdata.json")
and prints all the potentially-relevant entries.

```bash
(
    CODES=$(
    ls static/translations/messages_* |
    perl -ne '/messages_(..)/ or die; print "$1\n"' |
    sort | uniq | xargs echo
    );
    node -e "
    const codes = ('${CODES[@]}').split(' ');
    first = s => s.split(/[;,]|( \()/)[0];
    data = require('./langdata.json');
    out = codes.map(code => {
        const entry = data.filter(s => s.code == code)[0];
        return Object.fromEntries( Object.entries(entry)
        .map(([k,v]) => [k==='code'?'locale':k,first(v)])
        );
    });
    out.forEach(s => s.nativeName = s.nativeName.replace(/./, c=>c.toUpperCase()));
    console.log(out);
    "
)
```

Note that the language data therein is not quite what we want (for
example, "Russian" is given as "Русский язык" rather than "Русский",
and there is no distinction between traditional and simplified
Chinese), but all added entries were verified against Wikipedia and/or
other common language-selection lists on the wider Internet.
````

ray-kraesig:
(This is still sort of "in progress" – although not actively, of late – but that's better represented by marking it as a draft.)

ray-kraesig:
So here's a relatively milquetoast version of the translation-automation script. It does little more than generate the existing files that look like they already should be autogenerated.

Because it doesn't have access to the data in `languages.js`, it can't do additional checks for whether we really need or want the additional imports. As such, the script also adds a few new useless imports from `react-intl` and from unused-but-extant `messages_*.json` files. (We already have a fair number of these, so this isn't a *new* problem.)

I have some additional (unpolished, but functional) work that more or less pulls the data currently in `languages.js` out into a YAML file, which is then read by the new script to generate exactly the imports and settings-data that we actually want. I was originally planning to submit that as a successor to this PR, but it's starting to look like it'll be cleaner to squash them together and present it as an alternate PR instead.

PR #3917, opened 2020-02-19
api types [nfc]: Remove sole reference to deprecated 'pointer' in use…

chrisbobbe:
…rs/me response.

See
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/removing.20the.20pointer/near/818607,
and the corresponding issue
https://github.com/zulip/zulip/issues/8994.

The pointer will no longer be a user-facing concept. Luckily, we
weren't using it, except to describe this type.

### DIFF
--- a/src/api/users/getUserProfile.js
+++ b/src/api/users/getUserProfile.js
@@ -10,9 +10,9 @@ type ApiResponseUserProfile = {|
   is_admin: boolean,
   is_bot: boolean,
   max_message_id: number,
-  pointer: number,
   short_name: string,
   user_id: number,
+  // pointer: number, /* deprecated 2020-02; see zulip/zulip#8994 */
 |};
 
 /** See https://zulipchat.com/api/get-profile */


ray-kraesig:
The archaeologists of the future may appreciate a comment left in its place noting that `pointer` has been deprecated (removed?) as of such-and-such date or server version. (It's not terribly important for this field, but it's a nice habit to be in.)

chrisbobbe:
> The archaeologists of the future may appreciate a comment left in its place noting that `pointer` has been deprecated (removed?) as of such-and-such date or server version. (It's not terribly important for this field, but it's a nice habit to be in.)

Makes sense. We don't have a server version yet (the change is planned, but hasn't been made), so I put 2020-02, in Greg's preferred date format.

ray-kraesig:
This might be a bit much. :sweat_smile: I was only thinking something like...

```diff
   max_message_id: number,
-  pointer: number,
+  // pointer: number,  /* deprecated 2020-02; see zulip/zulip#8994 */
   short_name: string,
```

If there's a need (or want) for a more comprehensive explanation, it would probably be better to move `pointer` to the end of the type, as a rarely-referenced component.

chrisbobbe:
Ah.  😅Sounds good, here you go.

ray-kraesig:
Merged. Thanks!

Issue #3919, opened 2020-02-20
Deleted Messages can be edited in message lists


sumukhah:
Normally we don't want to change contents in deleted messages,
but currently we can edit deleted messages.
![before](https://user-images.githubusercontent.com/23723464/74962604-41600380-5436-11ea-8283-8869db35ac40.jpg)

sumukhah:
Working on the issue

agrawal-d:
It's possible to edit "deleted messages" in the webapp, though.

sumukhah:
> It's possible to edit "deleted messages" in the web app, though.
in the web app, deleting a message actually removes that message from List.
But in mobile-app deleting a message changes it's body to "(deleted)"..
if we deleted a message in mobile-app , that message still appears as "(deleted)" in web-app, web app won't consider it as a deleted message.

gnprice:
As @ray-kraesig wrote over on the PR #3920:
> Per [discussion on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Actions.20for.20deleted.20messages/near/819409), this is currently our intended behavior (also seen on the webapp).
> 
> Discussion about whether those deletion semantics are really desirable is welcome (preferably, but not necessarily, also on CZO) – but such a discussion should probably precede further revisions of this PR. Closing, at least for now.

Closing the issue for the same reason.

As I said in the chat thread, the model Zulip has for "deleted" messages certainly can be confusing, and there may be a better one we could be doing!

But we want the mobile app to follow the webapp here -- this is an aspect of the product, not of the specific UI each user sees for interacting with it, and so we want the mobile app and webapp to behave the same way. So the place to pursue changing this behavior would be on either the zulip/zulip tracker, or a chat stream like `#general` or `#feedback` that isn't specific to the mobile app.

PR #3920, opened 2020-02-20
hide "Edit Message" option for deleted messages

sumukhah:
editing deleted message makes a deleted action reversible at any point.
This is not as expected.

Fix: #3919 
after


ray-kraesig:
Per [discussion on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Actions.20for.20deleted.20messages/near/819409), this is currently our intended behavior (also seen on the webapp).

Discussion about whether those deletion semantics are really desirable is welcome (preferably, but not necessarily, also on CZO) – but such a discussion should probably precede further revisions of this PR. Closing, at least for now.

PR #3922, opened 2020-02-21
notifications: Correctly stringify pmUsers to fix navigation to group…

chrisbobbe:
… PMs.

Navigation to a group PM on pressing a notification was broken
because pmUsers was incorrectly stringified in
GroupPm.getPmUsersString. Fix, and add a test for it.

### DIFF
--- a/android/app/src/test/java/com/zulipmobile/notifications/FcmMessageTest.kt
+++ b/android/app/src/test/java/com/zulipmobile/notifications/FcmMessageTest.kt
@@ -52,6 +52,19 @@ class FcmMessageTest : FcmMessageTestBase() {
     }
 }
 
+class RecipientTest : FcmMessageTestBase() {
+    @Test
+    fun `GroupPm#getPmUsersString gives the correct string`() {
+        expect.that(Recipient.GroupPm(pmUsers = setOf(123, 234, 345)).getPmUsersString())
+            .isEqualTo("123, 234, 345")
+    }
+    @Test
+    fun `GroupPm#getPmUsersString correctly reorders user ids`() {
+        expect.that(Recipient.GroupPm(pmUsers = setOf(234, 123, 23, 345)).getPmUsersString())
+            .isEqualTo("23, 123, 234, 345")
+    }
+}
+
 class MessageFcmMessageTest : FcmMessageTestBase() {
     object Example {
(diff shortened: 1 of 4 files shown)


android/app/src/test/java/com/zulipmobile/notifications/FcmMessageTest.kt
```diff
@@ -52,6 +52,19 @@ class FcmMessageTest : FcmMessageTestBase() {
     }
 }
 
+class RecipientTest : FcmMessageTestBase() {
+    @Test
+    fun `GroupPm#getPmUsersString gives the correct string`() {
+        expect.that(Recipient.GroupPm(pmUsers = setOf(123, 234, 345)).getPmUsersString())
+            .isEqualTo("123, 234, 345")
```
gnprice: Let's eliminate the spaces too.

If this already is enough that the JS code understands it, that's good :) -- but the format used elsewhere is purely comma-separated, so may as well make it consistent.

android/app/src/test/java/com/zulipmobile/notifications/FcmMessageTest.kt
```diff
@@ -52,6 +52,19 @@ class FcmMessageTest : FcmMessageTestBase() {
...
+    }
+    @Test
+    fun `GroupPm#getPmUsersString correctly reorders user ids`() {
+        expect.that(Recipient.GroupPm(pmUsers = setOf(234, 123, 345)).getPmUsersString())
+            .isEqualTo("123, 234, 345")
+    }
```
gnprice: This test can be made a bit stronger by testing that e.g. 123 sorts after 23 -- i.e. the sorting is of the numbers, and not a lexicographic sort of their ASCII representations.

chrisbobbe:
OK, I've pushed these changes. :)

chrisbobbe:
I'm resolving merge conflicts and clarifying a commit message for this now. ("Incorrectly stringified" is vague; I'm tracking down what exactly the output was, which I should have done in the first place.)

gnprice:
Merged! Thanks again. Just made small tweaks to the commit messages.

Evidently I dropped this PR back in March, oops. >_> Looks like it probably got lost in the chaos of all suddenly switching to work from home, early in the pandemic. I ran across the bug today, it looked familiar, and I found this PR again.

PR #3923, opened 2020-02-21
msglist [nfc]: Warn about passing non-corresponding messages and rend…

chrisbobbe:
…erMessages.

I mentioned maybe last week that I'd seen a case where out-of-view messages were being marked as read. Looking through the code, though, I actually didn't see how this could be happening; the warning this commit adds is already being heeded, as far as I can tell. I'll do more testing soon, but I at least wanted to add this warning in the code.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -339,6 +339,13 @@ type OuterProps = {|
 
   /* Remaining props are derived from `narrow` by default. */
 
+  // `messages` and `renderedMessages` SHOULD correspond to the same set of
+  // messages, but `messages` MUST be a subset of `renderedMessages`, or
+  // scrolling will wrongly mark out-of-view messages as read! `startMessageId`
+  // and `endMessageId` in MessageListEventScroll are derived from
+  // `renderedMessages`, but are used as bounds on `messages`; see markRead in
+  // webViewEventHandlers.js. For the same reason, both arrays MUST be ordered
+  // by messageId.
   messages?: Message[],
   renderedMessages?: RenderedSectionDescriptor[],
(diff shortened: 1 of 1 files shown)


chrisbobbe:
(It may be the case that we want code changes to enforce this, or make it something we never have to think about.)

Also, I've assigned lots of recent PRs to Greg, so I'm assigning this one to Ray.

ray-kraesig:
In this case, wouldn't it be simpler to remove `renderedMessages` as a prop entirely, and compute it as `renderMessages(props.messages ?? getShownMessagesForNarrow(props.narrow)(state), props.narrow)` inside `MessageList`'s `mapStateToProps`?

The only caller that supplies `renderedMessages`, AFAICT, is `SearchMessagesCard`, and that's basically what it's doing – and I'm not convinced the caching done by our `getRenderedMessages` selector in the current case is effective.

Some not-necessarily-strict subset of this comment would still be worth making, but it'd be in `SelectorProps` rather than `OuterProps`, and/or possibly in `renderMessages`. (Which, itself, probably belongs in `src/webview/`.)

chrisbobbe:
Yes, that looks like a clear way forward to accomplish "make illegal states unrepresentable" (citation...somewhere?).

https://github.com/zulip/zulip-mobile/pull/3913 touches this code (and in fact renames renderedMessages, renderMessages, getRenderedMessages, and, I think, renderMessagesAsHtml), so perhaps I'll wait for that to land. That's rather substantial, though, and it's only half of the work needed for https://github.com/zulip/zulip-mobile/issues/3851.

chrisbobbe:
I expect this to be made irrelevant by #4465; closing.

Issue #2138, opened 2018-03-26
Per-realm notification settings (formerly "Fine grained notification settings")


mkaito:
At the highest level, per-ream notification settings seem like they would be pretty useful.

In a more specific use-case, per-stream or even per-topic settings would be nice. Specifically, we're looking to use Zulip as part of our alerting system for the devops team, and being able to "mute everything except this stream/topic" would help.

anirudhjain75:
Yes it will be a good improvement, @borisyankov Should I start working on this ?

anirudhjain75:
Though I don't think per topic notification would be good, but per stream would be nice

pulkonet:
This feature can be given to a user as an option to choose from. This would be pretty sweet if implemented correctly. I would like to work on it, but I don't think this is anywhere on the priority list.

chrisbobbe:
I don't have an update on per-realm notification settings, but there's been some work on stream-level notification settings.

When https://github.com/zulip/zulip-mobile/issues/3449 was closed with the merge of https://github.com/zulip/zulip-mobile/pull/3765 (commit https://github.com/zulip/zulip-mobile/commit/3b9246735d0f68b3bb14426fc4c73637ade5ed63), now, stream notifications without an explicit setting will default to the user's general push notification setting. So you could "mute everything except this stream" by setting your user-level setting to off, and explicitly marking one stream as on. Hope this helps. 🙂

chrisbobbe:
Ah, actually, looking at the code and with some manual testing, it appears that "Notifications when offline" and "Notifications when online" are being correctly set per-realm, but "Stream notifications" are not. That'll be a quick PR that, I think, should close this issue. @gnprice, thoughts?

```diff
export default (state: SettingsState = initialState, action: Action): SettingsState => {
  switch (action.type) {
    case REALM_INIT:
      return {
        ...state,
        offlineNotification: action.data.enable_offline_push_notifications,
        onlineNotification: action.data.enable_online_push_notifications,
+        streamNotification: action.data.enable_stream_push_notifications,
      };
```

chrisbobbe:
Opened as PR https://github.com/zulip/zulip-mobile/pull/3926.

[...]

PR #3926, opened 2020-02-21
Update stream notification setting when switching accounts

chrisbobbe:
Discovered in https://github.com/zulip/zulip-mobile/issues/2138#issuecomment-589806328.


ray-kraesig:
Well, this seems like a clear bug and a straightforward fix. And it even comes with more typechecking in tests! (Those last few tests look like they could reasonably be deduplicated with a for-loop, but that can be done in a follow-up if desired.)

Updated to reference the new location of `exampleData`, and merged. :tada: Thanks!

chrisbobbe:
Yay, thanks!

PR #3928, opened 2020-02-24
js [nfc]: Simplify expressions using `??`.

ray-kraesig:
Originally part of #3912, in which this was performed manually. Candidate expressions were identified with a multiline search for `/==\s*undefined\s*\?/` and manually checked. (`null` was also checked for, but no simplifiable expressions using it were found.)

This revised commit was generated in part automatically, with a
temporary installation of jscodeshift. The resultant script:

 * correctly excluded one item (`initialScrollMessageId`) noted to be incorrect in #3912;
 * arguably correctly excluded another item (`ackedPushToken`) *not* noted to be incorrect;
 * converted one item (`match_content`) in a technically-semantically-correct fashion that caused Flow to complain and required manual adjustment to fix;
 * had to repeatedly reinvoke Flow via the shell to get type information; and
 * was ~140 lines of ugly code (more, if properly cleaned up) that doesn't belong anywhere in our repo.

### DIFF
--- a/src/__tests__/exampleData.js
+++ b/src/__tests__/exampleData.js
@@ -52,7 +52,7 @@ const randString = () => randInt(2 ** 54).toString(36);
 
 const randUserId: () => number = makeUniqueRandInt('user IDs', 10000);
 const userOrBotProperties = ({ name: _name }) => {
-  const name = _name !== undefined ? _name : randString();
+  const name = _name ?? randString();
   const capsName = name.substring(0, 1).toUpperCase() + name.substring(1);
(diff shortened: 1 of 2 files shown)


ray-kraesig:
... I cannot locally reproduce this test failure.

gnprice:
LGTM, thanks!

I'm curious to hear more of the story of the jscodeshift experiment. Perhaps in a stream chat thread?

gnprice:
Merged -- with a tweak to the summary line after I realized that it was ambiguous :-)

PR #3929, opened 2020-02-24
docs/build-run: Add more troubleshooting cases.

chrisbobbe:


docs/howto/build-run.md
```diff
@@ -268,6 +280,16 @@ If you've run `react-native link`, you can discard the edits it made
 (along with any other edits you've made) by running `git reset --hard`.
 
 
+### Build failure: missing `/node_modules/react-native/third-party/double-conversion-1.1.6/src/fast-dtoa.cc`
+
+When trying to build the iOS app, if you get an error like
+
+```
+error: Build input file cannot be found: '/Users/chrisbobbe/dev/zulip-mobile/node_modules/react-native/third-party/double-conversion-1.1.6/src/bignum.cc' (in target 'double-conversion' from project 'React')
```
chrisbobbe: Unsure about having my name in here (I don't really mind); removing it might lose clarity.

ray-kraesig: Apple occasionally uses `jappleseed` for this purpose. ([Example.](https://developer.apple.com/documentation/foundation/filemanager/1407693-url#discussion))

chrisbobbe: Done 👍

docs/howto/build-run.md
```diff
@@ -268,6 +280,23 @@ If you've run `react-native link`, you can discard the edits it made
...
+```
+
+then try restarting Xcode.
+
+
+### App gets stuck on "Downloading JavaScript bundle 100%" with no loading spinner
+
+Try using the "shake" gesture and turning off remote debugging. (You can turn it
+on later.)
+
```
gnprice: Huh interesting. I think I see this regularly, and just reload -- i.e., use the "shake" gesture and choose reload there, or just hit `r` `r` in the Android emulator. (Or IIRC Cmd+R in the iOS simulator.)

One side effect of turning remote debugging off, or on, is that it'll cause a reload. So I kind of suspect that that's what's making this tip work. Is that consistent with what you've seen?

chrisbobbe: No, reloading wasn't working for me. I came across this solution from Stack Overflow; see https://stackoverflow.com/a/54318120/6792075.

gnprice:
> (There are also some style adjustments I want to make, but I'll just do those as a commit or two on top.)

Done:
75dba08a0 docs/build-run: Add further context on old Detox errors.
501ab5828 docs/build-run: Small edits to new troubleshooting items.
2fed12775 docs/build-run: Add further context on Xcode 11 / old-RN issue.

PR #3931, opened 2020-02-25
recipients [nfc]: Small refactors to new filteredRecipientsForPM helper.

gnprice:
This is a quick followup to #3874. It turns out this new helper function has some close relatives already existing, collected in one module, so we move it next to them. Then make a couple of small tweaks within the function.

There's plenty that could be improved about this whole family of very similar functions, and their names and documented interfaces. Tomorrow I may make a go at trying to find a minimal PR in that direction that doesn't block on doing any big broader refactors. In any case, I think this branch is helpful as is.

/cc @agrawal-d


gnprice:
Also, just to say this explicitly: thanks @ray-kraesig and @agrawal-d for pulling this function out in the first place! That was a necessary step in anything we might do to try to clean up this cluster of logic.

gnprice:
Thanks @ray-kraesig for the review!

Yeah, `display_recipient` is definitely a mess. Documenting would be a very good start; normalizing (at the app's outward boundary, i.e. at a [crunchy shell](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md)) would be great too. I suspect normalizing will happen in stages (rather than in one go), and will probably involve some iteration on finding the shape we want to normalize it into.

PR #3932, opened 2020-02-26
recipient [nfc]: Split up pm{Ui,Key}RecipientsFromMessage, and start on "unreads" keys.

gnprice:
This is a more substantial follow-up to #3931.

There's still plenty that's confusing about these interfaces, but I think this branch improves the situation a good bit.

I have some draft other changes, including to give `getRecipientsIds` a better name (turns out what it's really about is specifically for the "unread messages" data structure) and some jsdoc; to start cleaning up `normalizeRecipients` and `normalizeRecipientsSansMe`; and even to write down exactly what the story is for what appears in `display_recipient`! Those still require some further polish on a few points before they're ready to merge, though. And this branch is already more than I'd quite planned to do on this today. ;-)

### DIFF
--- a/src/unread/unreadHelpers.js
+++ b/src/unread/unreadHelpers.js
@@ -7,19 +7,12 @@ type SomeUnreadItem = { unread_message_ids: number[] };
 export function removeItemsDeeply<T: SomeUnreadItem>(objArray: T[], messageIds: number[]): T[] {
   let changed = false;
   const objWithAddedUnreadIds = objArray.map(obj => {
-    const newIds = removeItemsFromArray(obj.unread_message_ids, messageIds);
-    if (newIds.length === obj.unread_message_ids) {
-      return obj;
-    }
-
     const filteredIds = removeItemsFromArray(obj.unread_message_ids, messageIds);
-
     if (filteredIds.length === obj.unread_message_ids.length) {
       return obj;
     }
 
     changed = true;
-
     return {
       ...obj,
       unread_message_ids: filteredIds,
--- a/src/unread/unreadHelpers.js
+++ b/src/unread/unreadHelpers.js
@@ -33,7 +33,6 @@ function addItemsDeeply<T: SomeUnreadItem>(input: T[], itemsToAdd: number[], ind
   const item = input[index];
 
   const unreadMessageIds = addItemsToArray(item.unread_message_ids, itemsToAdd);
-
   if (item.unread_message_ids === unreadMessageIds) {
     return input;
   }
@@ -42,7 +41,7 @@ function addItemsDeeply<T: SomeUnreadItem>(input: T[], itemsToAdd: number[], ind
     ...input.slice(0, index),
     {
       ...item,
-      unread_message_ids: addItemsToArray(item.unread_message_ids, itemsToAdd),
+      unread_message_ids: unreadMessageIds,
     },
     ...input.slice(index + 1),
   ];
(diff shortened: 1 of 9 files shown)


src/utils/recipient.js
```diff
@@ -85,12 +85,17 @@ export const getRecipientsIds = (message: Message, ownEmail?: string): string =>
...
-    ? recipients.filter(r => r.email !== ownEmail)[0].id.toString()
-    : recipients
-        .map(s => s.id)
-        .sort((a, b) => a - b)
-        .join(',');
+  if (recipients.length === 2) {
+    if (ownEmail === undefined) {
+      throw new Error('getRecipientsIds: got 1:1 PM, but ownEmail omitted');
+    }
+    return recipients.filter(r => r.email !== ownEmail)[0].id.toString();
```
ray-kraesig: I mean, it's certainly better than the status quo in which there's no check at all, but... :nauseated_face:

If you can't reasonably split this into two separate functions, can you at least give it an explanatory (and, presumably, abjuratory) docstring? Ideally with a `TODO:` with a suggested path forward. (...and possibly rename it not to have a plural as a [noun adjunct](https://en.wikipedia.org/wiki/Noun_adjunct).)

Unrelatedly, the summary on this commit is a little too long. (GitHub wraps them after 72 characters.)

src/unread/unreadHelpers.js
```diff
@@ -7,19 +7,12 @@ type SomeUnreadItem = { unread_message_ids: number[] };
 export function removeItemsDeeply<T: SomeUnreadItem>(objArray: T[], messageIds: number[]): T[] {
   let changed = false;
   const objWithAddedUnreadIds = objArray.map(obj => {
-    const newIds = removeItemsFromArray(obj.unread_message_ids, messageIds);
-    if (newIds.length === obj.unread_message_ids) {
```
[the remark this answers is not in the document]
gnprice: :-/ Yeah.

One thing is that this isn't *unsound* -- it's just always false. So it should be basically a lint-style warning.

And then one tricky thing, I guess, is that it's often intentional to have a check that the types say should always be false -- because there's some un-type-checked code elsewhere in the system and you're not sure if it behaves properly. Ideally you'd express that uncertainty in the types of that code's interfaces, so that you'd necessarily check at the boundary and then the type-checker would guarantee you'd adequately checked, but that's often not how it works out. So I think I'd really want a way to silence that lint warning explicitly on a given check, with a hammer smaller than `$FlowFixMe` that won't also suppress unrelated actual type errors if they arise.

Still, none of that seems particularly insurmountable, and I agree it'd be a good check to have.

src/utils/recipient.js
gnprice: Yeah, there's still further improvements this function needs. As I mentioned above:

> I have some draft other changes, including to give `getRecipientsIds` a better name (turns out what it's really about is specifically for the "unread messages" data structure) and some jsdoc

---

> Unrelatedly, the summary on this commit is a little too long. (GitHub wraps them after 72 characters.)

What we say [in our docs](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages)￼, following some discussion in the last few months, is that the summary:
> is kept short (max 76 characters, ideally less), while concisely explaining what the commit does

This is 75ch, so I guess it's just under the wire.

gnprice:
Thanks @ray-kraesig for the review!

Based on your comments, I've gone ahead and merged this branch. There's still further changes I want to make here; I'll see about sending a followup PR next.

PR #3936, opened 2020-02-29
Remove `isActive`

ray-kraesig:
Remove `isActive` from our Redux state.

~~(Blocked on #3727.)~~

### DIFF
--- a/src/actionConstants.js
+++ b/src/actionConstants.js
@@ -3,7 +3,6 @@ export const REHYDRATE: 'persist/REHYDRATE' = 'persist/REHYDRATE';
 
 export const APP_ONLINE: 'APP_ONLINE' = 'APP_ONLINE';
 export const APP_ORIENTATION: 'APP_ORIENTATION' = 'APP_ORIENTATION';
-export const APP_STATE: 'APP_STATE' = 'APP_STATE';
 export const DEAD_QUEUE: 'DEAD_QUEUE' = 'DEAD_QUEUE';
 
 export const REALM_ADD: 'REALM_ADD' = 'REALM_ADD';
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -2,7 +2,6 @@
 import {
   REHYDRATE,
   APP_ONLINE,
-  APP_STATE,
   DEAD_QUEUE,
   INIT_SAFE_AREA_INSETS,
   APP_ORIENTATION,
@@ -114,11 +113,6 @@ type RehydrateAction = {|
 type AppOnlineAction = {|
   type: typeof APP_ONLINE,
   isOnline: boolean,
-|};
-
-type AppStateAction = {|
-  type: typeof APP_STATE,
-  isActive: boolean,
 |};
 
 type DeadQueueAction = {|
@@ -596,7 +590,6 @@ type RealmAction = RealmInitAction | UnackPushTokenAction | AckPushTokenAction;
 
 type SessionAction =
   | RehydrateAction
-  | AppStateAction
   | AppOnlineAction
   | InitSafeAreaInsetsAction
   | AppOrientationAction
--- a/src/boot/AppEventHandlers.js
+++ b/src/boot/AppEventHandlers.js
@@ -15,7 +15,7 @@ import {
   NotificationListener,
   notificationOnAppActive,
 } from '../notification';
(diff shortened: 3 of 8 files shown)


gnprice:
Thanks! I like all these changes. I particularly like the tactic of cleaning up some of that tangled logic in eventMiddleware by a series of small `[nfc]` commits up front.

One nit in the last commit message:

    However, that underlying state affects the cached value's staleness
    period in ways that are nontrivial to reason about: its updates are
    delayed based on the decisions of two different schedulers, one of
    which is OS-dependent.

Would you name what two schedulers you have in mind? I think one is the OS deciding whether to run our process; the other is... the JS implementation's event loop? I haven't 100% thought through what does happen, because the fact that it's complicated is enough to confirm your point :slightly_smiling_face: 

What's the sense in which this is blocked on #3727? It looks to me like it's good as it stands (modulo that one nit.) We got to this from discussing whether to use `isActive` in #3727 (or its predecessor), but the conclusion was just to not add a new use of it; there wasn't an existing use that that PR would take out.

ray-kraesig:
> Thanks! I like all these changes. I particularly like the tactic of cleaning up some of that tangled logic in eventMiddleware by a series of small `[nfc]` commits up front.

Fortunately the Alexandrian strategy wasn't necessary here.

(I'm actually not terribly happy with it as is: presently it relies on the fact that checking to play a sound is the last thing it does, by using `return` to skip that rather than something more tightly scoped. But I suspect you would have liked my solutions to that rather less than the status quo. :upside_down_face:)

> Would you name what two schedulers you have in mind? I think one is the OS deciding whether to run our process; the other is... the JS implementation's event loop?

Well, its evaluator-dispatch mechanism, anyway – which may or may not be unified with the thing I'd usually call the "event loop". Left vague mostly because *a)* "evaluator-dispatch mechanism" is a poor and unwieldy description (the established term of art, if one exists, escapes me), *b)* I don't want to describe the particulars of an irrelevant implementation, and *c)* as you go on to note, it doesn't really matter.

Amended, however, to more correctly read "*at least* one of which is OS-dependent": for all I know, RN's version of JavaScriptCore uses different strategies on iOS and Android, and who's to say we won't end up moving to Hermes on one OS before the other?

(It's possible that someone somewhere makes guarantees about callback ordering that could, in theory, allow us to analyze this as being more deterministic, but I wouldn't consider such guarantees themselves to be reliable even if they're in a language spec – it's exactly the sort of thing that I can see a particular implementation explicitly deciding to diverge from the spec from for performance reasons, given that 99.9% of programmers don't (and 99.999% don't need to) write code that depends on things like that.)

> What's the sense in which this is blocked on #3727? [...] [T]here wasn't an existing use [of `isActive`] that that PR would take out.

... somehow I'd gotten it into my head that there *was* such a use. I don't know why; obviously the tests would have failed if there actually were one.

ray-kraesig:
Rebased, and reamended to avoid raising the question.

gnprice:
Hmm, I think you may have not pushed your amended version? It doesn't look like the message has changed -- commit 72cfff8a9 still says:

    However, that underlying state affects the cached value's staleness
    period in ways that are nontrivial to reason about: its updates are
    delayed based on the decisions of two different schedulers, one of
    which is OS-dependent.

Just to record here what we discussed in person earlier (reconstructed from what I recall):
* It's 100% fine for the commit message to be uncertain about things you're uncertain about, and that don't seem worth the effort required to become certain about.
* What I want to avoid here is the sense that there's something you do know, that's part of your reasoning and the reader has to know too in order to follow your reasoning, but that the reader is left to guess.
  * Or something you believe, even with moderate or weak confidence, but that's part of your reasoning; the reader shouldn't have to guess at that either. Uncertainty is fine, just make it explicit.
* At a tactical level for how to express this thinking without sounding more certain than you want to: consider expanding a noun conglomerate like "evaluator-dispatch mechanism" into a description, like "how the JS implementation decides which tasks to execute when." Noun conglomerates tend to sound jargony and technically specific.

ray-kraesig:
> Hmm, I think you may have not pushed your amended version?

Well, this is puzzling. I *did* push the amended version – as b40ff77f7, still visible above. I'm not sure how the following version ended up being the pre-amendment one – perhaps due to out-of-date refs on my home machine?

Re-rebased and pushed.

gnprice:
Huh, indeed.

Looks good, thanks -- merged!

PR #3937, opened 2020-02-29
docs/howto/push-notifications: Add sandbox APNs workaround.

chrisbobbe:
It will still be really valuable to explore the solutions proposed
near the end of this file.

### DIFF
--- a/docs/howto/push-notifications.md
+++ b/docs/howto/push-notifications.md
@@ -170,6 +170,85 @@ idiosyncratic capitalization shown.
 
 ### Current workaround
 
+This workaround means using a development server. If that won't work
+for you, you'll have to try the second workaround, below.
+
+You can tell your development server to talk to Apple's APNs "sandbox"
+server, instead of its server meant for production, but you'll need a
+certificate signed by Apple authorizing you to do so. Some background
+on that is
+[here](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns/#2947606).
+
+1. First, generate a Certificate Signing Request (CSR). Apple
+   Developer's instructions are
+   [here](https://help.apple.com/developer-account/#/devbfa00fef7).
+   You can use something like "John Appleseed APNs Sandbox" for the
+   "Common Name."
+
+2. Greg is authorized in Apple Developer to upload the CSR and obtain
+   the actual certificate, so you should send it to him and ask him to
+   do that. He'll follow something like [these
+   instructions](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/establishing_a_certificate-based_connection_to_apns)
+   to upload your CSR and obtain an APNs SSL (Sandbox) certificate
+   (not Sandbox & Production). Greg will send you the certificate.
+   Open the file, so Keychain Access will import it into your
+   keychain.
+
+   (Note: When you created the CSR, Keychain Access also created a
+   private key. The CSR itself is not the private key, so it doesn't
+   require extra care to send over the Internet. As of 2020-03-02, the
+   APNs certificate doc incorrectly conflates the CSR and the private
+   key. The CSR file contains only the public key and a signed hash of
+   the request itself; this can be confirmed by extracting its
+   contents with OpenSSL:
+
+   `openssl asn1parse -i -in CertificateSigningRequest.certSigningRequest`
+
+   `openssl req -in CertificateSigningRequest.certSigningRequest -text -noout`
+
+   .)
+
+3. Keychain Access will recognize that the certificate corresponds to
+   your private key. Select them both and choose to export them to a
+   single .p12 (PKCS #12 format) file.
+
+4. Move the .p12 file into your zulip.git clone, in /zproject. Then
+   ssh into the Vagrant container (`vagrant ssh`) and convert the .p12
+   file into a .pem file, with
+
+   `openssl pkcs12 -in Certificates.p12 -out apns-sandbox.pem -nodes`
+
+   (replacing Certificates.p12 with whatever your filename is). If
+   openssl isn't installed, run `sudo apt install openssl`. Then
+   delete the .p12 file; you won't need it anymore.
+
+5. The .pem file contains your private key, so be sure you don't push
+   it to GitHub! One way this could be done is with the .gitignore
+   file, but .gitignore is itself version-controlled, and other people
+   probably don't have a .pem file with the same name. The
+   .git/info/exclude file lets you tell Git what to ignore, but it
+   isn't in version control, so other people won't be confused by its
+   contents. Add a line with "zproject/apns-sandbox.pem" (or whatever
+   the path is for you). Confirm that the .pem file isn't being
+   tracked by Git; it should not show up at all when you run `git
+   status`.
+
+6. Add a line with `APNS_CERT_FILE = "zproject/apns-sandbox.pem"` to
+   zproject/dev_settings.py. This lets Python use the certificate to
+   communicate with the APNs sandbox server.
+
+Now, restart the server, and you should be receiving notifications on
+your iOS development build! Be sure mobile notification settings are
+on, using the web app's settings interface. (You're all set to do this
+if your branch is up-to-date with a version of `master` from 2020-03
+or later; if not, you'll need the changes in [this
+commit](https://github.com/zulip/zulip/commit/23ba2b63c5c10f43b02a2bb2c470cc6ff597d839)).
+If it's not working, please say so in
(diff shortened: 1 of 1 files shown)


docs/howto/push-notifications.md
```diff
@@ -170,6 +170,64 @@ idiosyncratic capitalization shown.
...
+   something like "Johnny Appleseed APNs Sandbox" for the "Common Name."
+
+2. Greg is authorized in Apple Developer to upload the CSR and obtain the actual
+   certificate, so you should send it to him and ask him to do that. He'll
+   follow something like [these
+   instructions](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/establishing_a_certificate-based_connection_to_apns)
+   to upload your CSR and obtain an APNs SSL (Sandbox) certificate (not Sandbox
+   & Production). (Note: When you created the CSR, Keychain Access also created
+   a private key. The CSR itself is not the private key, so it doesn't require
+   extra care to send over the Internet. The APNs certificate doc **wrongly**
```
chrisbobbe: The offending doc is at https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/establishing_a_certificate-based_connection_to_apns; Greg pointed out to me that this doc is, indeed, in error — not us.

Still, if there's a good way to do it, I'd like to succinctly allay a contributor's fear that _we're_ the ones who are wrong here. Ideas?

chrisbobbe:
Note that https://github.com/zulip/zulip/commit/23ba2b63c5c10f43b02a2bb2c470cc6ff597d839 on the server is required; should I mention this in the doc?

docs/howto/push-notifications.md
ray-kraesig: Less bold text, for one. Perhaps something like the following?

> Note that, as of 2020-03-02, the linked Apple documentation incorrectly conflates the CSR and the private key. The CSR contains only *<the public key / a hash of the private key>*, which can be confirmed by *<whatever you and/or Greg did to confirm this>*.

docs/howto/push-notifications.md
```diff
@@ -170,6 +170,64 @@ idiosyncratic capitalization shown.
...
+
+We can tell our development server to talk to Apple's APNs "sandbox" server,
+instead of its server meant for production, but you'll need a certificate signed
+by Apple authorizing you to do so. Some background on that is
+[here](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns/#2947606).
+
+1. First, generate a Certificate Signing Request (CSR). Apple Developer's
+   instructions are
+   [here](https://help.apple.com/developer-account/#/devbfa00fef7). You can use
+   something like "Johnny Appleseed APNs Sandbox" for the "Common Name."
```
ray-kraesig: Customarily "John Appleseed" in this context. (Comparison: [John](https://www.google.com/search?hl=en&q=%22john+appleseed%22+site%3Adeveloper.apple.com) vs. [Johnny](https://www.google.com/search?hl=en&q=%22johnny+appleseed%22+site%3Adeveloper.apple.com).)

docs/howto/push-notifications.md
```diff
@@ -170,6 +170,64 @@ idiosyncratic capitalization shown.
 
 ### Current workaround
 
+This workaround means using a development server. If that won't work for you,
+you'll have to try the second workaround, below.
+
+We can tell our development server to talk to Apple's APNs "sandbox" server,
```
ray-kraesig: > *You can tell *your development server [...]

perhaps? (I'm not sure how it's relevant that we can tell CZO to do this.)

chrisbobbe: Haha indeed; I've gotten so used to using "we," in commit messages, and this looks like spillover from that habit.

docs/howto/push-notifications.md
chrisbobbe: > <whatever you and/or Greg did to confirm this>

It was from Greg's knowledge of how CSR's work in general.

I believe it was this: 1) everyone who sets up their website to use SSL issues a CSR to a Certificate Authority over the Internet; 2) assume the CSR contains the private key; 3) either people's private keys are intercepted and stolen all the time, or no one issues CSR's because that's a possibility, and 4) contradiction, we're not seeing step 3.

@gnprice But this is a bit verbose; I think even a link to a contradictory doc from Apple would be fine.

ray-kraesig: There's [a relevant bit on Wikipedia](https://en.wikipedia.org/w/index.php?title=Certificate_signing_request&oldid=940627604#Structure) – although that's not very meaningful; generally if Wikipedia and Apple documentation conflict I would believe the latter.

In particular, it's not completely beyond belief that Apple might have a custom CSR format and process. (It would, arguably, be very Apple of them to do so.) If this involves a more secure method of submitting keys to Apple than plaintext, that would (initially appear to) avoid 3.

Fortunately, this is not only not so, but also easily mostly verifiably not so. Amending the previous suggestion:

> The CSR file contains the only the public key and a signed hash of the request itself; this can be confirmed by extracting its contents with OpenSSL:
>
> `openssl asn1parse -i -in CertificateSigningRequest.certSigningRequest`
> `openssl req -in CertificateSigningRequest.certSigningRequest -text -noout`

(I suppose someone might believe that the signature actually encodes the private key somehow; but demonstrating otherwise is well beyond the scope of this parenthetical, and also not too difficult for the interested paranoid to confirm without further explanation than this.)

The above lines may also serve as a reference for us: we can use these commands to peek at the inside of any received CSRs, to confirm that they are what they should be and nothing more.

chrisbobbe: Nice, thanks, I'll use this!

I was hoping we could delight readers' ears with the phrase "contradictory doc", but this is a better solution. 😝

chrisbobbe:
OK, thanks for the review, and those `openssl` commands, in particular.

docs/howto/push-notifications.md
```diff
@@ -170,6 +170,74 @@ idiosyncratic capitalization shown.
...
+   zproject/dev_settings.py. This lets Python use the certificate to communicate
+   with the APNs sandbox server.
+
+Now, restart the server, and you should be receiving notifications on your iOS
+development build! Be sure mobile notification settings are on, using the web
+app's settings interface (this will be easiest if your branch contains
+[zulip/zulip@23ba2b6](https://github.com/zulip/zulip/commit/23ba2b63c5c10f43b02a2bb2c470cc6ff597d839)).
+If it's still not working, please contact someone on the mobile team before
+trying the next workaround, as that one will mean a worse development experience
+for you.
```
gnprice: This warning isn't necessary, because the next one is quite impossible to do without our involvement :slightly_smiling_face: -- it involves uploading to TestFlight.

Possibly it'd be good to add a mention of that fact at the top of that next section.

I think the key thing to say here is: if it's not working, please say so so we can debug.

gnprice: Also, re "contact someone on the team": what we generally want to steer people toward is asking in a stream (like `#mobile`).

docs/howto/push-notifications.md
```diff
@@ -170,6 +170,74 @@ idiosyncratic capitalization shown.
...
+   by Git; it should not show up at all when you run `git status`.
+
+6. Add a line with `APNS_CERT_FILE = "zproject/apns-sandbox.pem"` to
+   zproject/dev_settings.py. This lets Python use the certificate to communicate
+   with the APNs sandbox server.
+
+Now, restart the server, and you should be receiving notifications on your iOS
+development build! Be sure mobile notification settings are on, using the web
+app's settings interface (this will be easiest if your branch contains
+[zulip/zulip@23ba2b6](https://github.com/zulip/zulip/commit/23ba2b63c5c10f43b02a2bb2c470cc6ff597d839)).
```
gnprice: Maybe helpful to say "merged 2020-02"? Or otherwise to communicate: if your branch is atop a version of `master` from 2020-03 or later, then you're set as far as this part is concerned and there's no need to go checking for it.

docs/howto/push-notifications.md
gnprice: > In particular, it's not completely beyond belief that Apple might have a custom CSR format and process. (It would, arguably, be very Apple of them to do so.)

It could be very Apple of them to have something custom -- but it'd be very un-Apple of them to have a system that involved them touching everyone's secret keys, when it's cryptographically possible to avoid that.

Helping the reader inspect things for themself, with these `openssl` commands, is a good solution.

chrisbobbe:
Thanks for the review, both of you! I've made the requested changes.

gnprice:

Mentioning here just because GitHub code-comment threads can be easy to lose things in: there's a followup item discussed at https://github.com/zulip/zulip-mobile/pull/3937#discussion_r388610471 .

Issue #3941, opened 2020-03-02
navigating to screens repeatedly between 2 screens, makes navigating back hard.  


sumukhah:
If there are 2 interlinked screens(say A, B).
if user navigate  to "mainScreen => A => B => A => B ..."
makes hard to retrace into mainScreen(it has to go back in B => A =>B => A => mainScreen)
![beforeIssue](https://user-images.githubusercontent.com/23723464/75688598-c3b5b680-5cc5-11ea-89b5-b33bba4b7340.gif)

sumukhah:
Working on the issue

jgabomination:
Hey @sumukhah is the issue fixed?

sumukhah:
@jgabomination 
There's been some discussion of navigation patterns like this one, around https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux.20.2F.20navigation/near/811675.

PR #3943, opened 2020-03-02
Fix back navigation 

sumukhah:
resolved

Fixes: #3941

### DIFF
--- a/src/nav/navSelectors.js
+++ b/src/nav/navSelectors.js
@@ -45,13 +45,13 @@ export const getCanGoBack = (state: GlobalState) => state.nav.index > 0;
 export const getSameRoutesCount: Selector<number> = createSelector(
   getNav,
   nav => {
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Hi @sumukhah, thanks for the new issue and PR! There's been some discussion of navigation patterns like this one, around https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux.20.2F.20navigation/near/811675.

Greg's comment at that link, in particular, addresses a case where I think this solution might add behavior that may be seen as buggy in some cases. In particular, if it's _not_ a repeated `a => b` pattern, e.g., `mainScreen => a => b => c => b`, and I click "back," I would expect to land at `c`. Instead, I would land at `a` and I might have no idea why. (Perhaps I'd forgotten I'd been at `b` before `c`.)

PR #3946, opened 2020-03-03
Extend exampleData to further clean up messageActionSheet tests

gnprice:
Thanks @agrawal-d for upgrading the message action sheet tests in #3824 / 4670210f8 to be well-typed! And thanks @ray-kraesig for reviewing and merging that PR.

After reading that commit, I saw some ways that that test file could be further improved to reduce boilerplate and also to make the test data more typical and more fully coherent (e.g., stream name and ID properties agreeing with each other.) Some of those involve extending `exampleData` with new features. This PR is the result.

(@agrawal-d , I can't seem to add you as a requested reviewer -- perhaps you're not in the GitHub org? But please do take a look. :slightly_smiling_face: )

### DIFF
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -8,27 +8,15 @@ const baseBackgroundData = deepFreeze({
   alertWords: [],
   allImageEmojiById: eg.action.realm_init.data.realm_emoji,
   auth: eg.selfAuth,
-  debug: {
-    doNotMarkMessagesAsRead: false,
-  },
+  debug: eg.baseReduxState.session.debug,
   flags: {
-    read: {},
+    ...eg.baseReduxState.flags,
     starred: {
       // Key has to be number according to the type declaration. $FlowFixMe.
       1: true,
       // Key has to be number according to the type declaration. $FlowFixMe.
       2: true,
     },
-    collapsed: {},
-    mentioned: {},
-    wildcard_mentioned: {},
-    summarize_in_home: {},
-    summarize_in_stream: {},
-    force_expand: {},
-    force_collapse: {},
-    has_alert_word: {},
-    historical: {},
-    is_me_message: {},
   },
   mute: [],
   ownUser: eg.selfUser,
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -48,14 +48,7 @@ describe('constructActionButtons', () => {
 
   test('show reactions option if message is has at least one reaction', () => {
     const message = eg.streamMessage({
-      reactions: [
-        {
-          user_id: 12345,
-          emoji_name: 'haha',
-          reaction_type: 'unicode_emoji',
-          emoji_code: '',
-        },
-      ],
+      reactions: [eg.unicodeEmojiReaction],
     });
 
     const buttons = constructMessageActionButtons({
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -20,20 +20,7 @@ const baseBackgroundData = deepFreeze({
   },
   mute: [],
   ownUser: eg.selfUser,
-  subscriptions: [
-    {
-      ...eg.makeStream(),
-      color: '#ffffff',
-      in_home_view: false,
-      pin_to_top: false,
-      audible_notifications: false,
-      desktop_notifications: false,
-      email_address: 'stream@email.com',
-      is_old_stream: false,
-      push_notifications: true,
-      stream_weekly_traffic: 2,
-    },
-  ],
+  subscriptions: [],
   theme: 'default',
   twentyFourHourTime: false,
 });
@@ -130,7 +117,7 @@ describe('constructHeaderActionButtons', () => {
 
     const subscriptions = deepFreeze([
       {
-        ...baseBackgroundData.subscriptions[0],
+        ...eg.subscription,
         name: 'electron issues',
         in_home_view: false,
       },
@@ -157,7 +144,7 @@ describe('constructHeaderActionButtons', () => {
 
     const subscriptions = deepFreeze([
       {
-        ...baseBackgroundData.subscriptions[0],
+        ...eg.subscription,
         name: 'electron issues',
         in_home_view: true,
       },
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -30,7 +30,7 @@ const baseBackgroundData = deepFreeze({
     historical: {},
     is_me_message: {},
   },
-  mute: [['announce', 'stream events']],
+  mute: [],
   ownUser: eg.selfUser,
   subscriptions: [
     {
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -98,17 +98,7 @@ describe('constructHeaderActionButtons', () => {
   });
 
   test('show Unmute stream option if stream is not in home view', () => {
-    const message = eg.streamMessage({
-      display_recipient: 'electron issues',
-    });
-
-    const subscriptions = deepFreeze([
-      {
-        ...eg.subscription,
-        name: 'electron issues',
-        in_home_view: false,
-      },
-    ]);
+    const subscriptions = [{ ...eg.subscription, in_home_view: false }];
 
     const backgroundData = deepFreeze({
       ...baseBackgroundData,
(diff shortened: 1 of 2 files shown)


agrawal-d:
I have a few examinations this week, I'll take a look once they are over.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -1,5 +1,6 @@
 // @flow strict-local
 import deepFreeze from 'deep-freeze';
+
```
agrawal-d: Oh! I didn't know about this!

I have one question - In several files, there is a newline after `// @flow strict-local`, in some files there isn't. I couldn't decipher any pattern. What is the correct practice?

src/__tests__/exampleData.js
```diff
@@ -125,6 +132,24 @@ export const stream: Stream = makeStream({
...
+export const makeSubscription = (args: { stream?: Stream } = {}): Subscription => {
+  const { stream: streamInner = stream } = args;
+  return deepFreeze({
+    ...streamInner,
+    color: '#123456',
+    in_home_view: true,
+    pin_to_top: false,
+    audible_notifications: false,
+    desktop_notifications: false,
+    email_address: '??? make this value representative before using in a test :)',
```
agrawal-d: I had a look at `zulip/zulip`'s `zerver/lib/email_mirror_helpers.py:32`, here's how the stream email is generated:

```python
def encode_email_address_helper(name: str, email_token: str, show_sender: bool=False) -> str:
    # Some deployments may not use the email gateway
    if settings.EMAIL_GATEWAY_PATTERN == '':
        return ''

    # Given the fact that we have almost no restrictions on stream names and
    # that what characters are allowed in e-mail addresses is complicated and
    # dependent on context in the address, we opt for a simple scheme:
    # 1. Replace all substrings of non-alphanumeric characters with a single hyphen.
    # 2. Use Django's slugify to convert the resulting name to ascii.
    # 3. If the resulting name is shorter than the name we got in step 1,
    # it means some letters can't be reasonably turned to ascii and have to be dropped,
    # which would mangle the name, so we just skip the name part of the address.
    name = re.sub(r"\W+", '-', name)
    slug_name = slugify(name)
    encoded_name = slug_name if len(slug_name) == len(name) else ''

    # If encoded_name ends up empty, we just skip this part of the address:
    if encoded_name:
        encoded_token = "%s.%s" % (encoded_name, email_token)
    else:
        encoded_token = email_token

    if show_sender:
        encoded_token += ".show-sender"

    return settings.EMAIL_GATEWAY_PATTERN % (encoded_token,)
```

I'm not at all familiar with Python, but it looks like we use a combination of the stream name in ASCII and some server configs to generate the stream name. Perhaps, this might help us in using a representative value here.

src/__tests__/exampleData.js
```diff
@@ -125,6 +132,24 @@ export const stream: Stream = makeStream({
...
+    is_old_stream: true,
```
agrawal-d: I couldn't find the meaning of `is_old_stream` in the codebase - can you please tell me? Also, it doesn't appear to be used anywhere at all ( I am probably wrong but a search in VS Code for `is_old_stream` gives only declarations, no uses. )

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -42,9 +36,11 @@ describe('constructActionButtons', () => {
 
   test('show unstar message option if message is starred', () => {
     const message = eg.streamMessage({ id: 1 });
+    // eslint-disable-next-line no-useless-computed-key
+    const flags = { ...baseBackgroundData.flags, starred: { [1]: true, [2]: true } };
```
agrawal-d: This is a nice trick! :-)

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -117,25 +107,15 @@ describe('constructHeaderActionButtons', () => {
...
-    });
-
-    const subscriptions = deepFreeze([
-      {
-        ...eg.subscription,
-        name: 'electron issues',
-        in_home_view: true,
-      },
-    ]);
+    const subscriptions = [{ ...eg.subscription, in_home_view: true }];
```
agrawal-d: Did you add `in_home_view: true` for readability, since it is already true in `makeSubscription()`?

src/message/__tests__/messageActionSheet-test.js
gnprice: > In several files, there is a newline after `// @flow strict-local`, in some files there isn't. I couldn't decipher any pattern. What is the correct practice?

Reasonable question!

I think we don't really have a set style. :-) I can't think of a particular preference I really have, either.

I did just now go look for examples of how I've done it myself, when adding a new file or adding type-checking to a file. I used the command `git lsp --author greg -S '@flow strict-local'` (i.e. `git log --stat -p --author greg -S '@flow strict-local'` but I [have an alias](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3) for `log --stat -p`), and used ["the secret to `git log -p`"](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md#2-the-secret-to-using-git-log--p) to conveniently browse the results.

The result is... I do both. :smile:

src/__tests__/exampleData.js
gnprice: Yeah -- the easiest way to get a representative value is probably to look at some empirical data, e.g. in the console when debugging the app.

But also it's not a priority to fill in this particular property's value, because we don't have any tests that care about it. I suspect it's something we actually completely don't care about in the app :wink:, and so ideally would just drop and not even store in Redux.

src/__tests__/exampleData.js
gnprice: From a scan of the server code, it seems to mean "this stream is old enough that the `stream_weekly_traffic` number is meaningful", and concretely "this stream is at least 7 days old".

It's not a great name :wink: 

Won't matter to us unless/until we start doing something with `stream_weekly_traffic` too.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -157,7 +144,7 @@ describe('constructHeaderActionButtons', () => {
 
     const subscriptions = deepFreeze([
       {
-        ...baseBackgroundData.subscriptions[0],
+        ...eg.subscription,
```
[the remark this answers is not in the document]
gnprice: Yeah -- specifically, I left it explicit here because it's part of what this test is about.

If it were omitted, then looking at this test you wouldn't be able to tell whether the test actually tests the case it's meant to test unless you knew that `eg.subscription` definitely has `in_home_view: true`. That is the logical default, and so you could perhaps figure that it ought to be the default... but even to know it's the logical default, you have to know, and think about, some context about what it *means*.

Whereas when it's explicitly here, then you can look at this test code very locally, and think through it at just a very straightforward step-by-step level, and see that it does exactly what it says it does.

src/message/__tests__/messageActionSheet-test.js
gnprice: (same answer)

src/api/modelTypes.js
```diff
@@ -440,7 +440,7 @@ export type Submessage = $ReadOnly<{|
  * See also `MessagesState` for discussion of how we fetch and store message
  * data.
  */
-export type Message = $ReadOnly<{
+export type Message = $ReadOnly<{|
```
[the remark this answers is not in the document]
chrisbobbe: > Really, *none* of the server-sent data types should be exact – future revisions of the server will almost certainly add new fields to some of them, and we shouldn't make promises that we know won't be kept. :confused: 

When those future revisions arrive, though, we'll try to update the types accordingly? Having exact types with temporary out-of-date states feels better, to me, than making them all inexact. We have server and client capabilities to handle migrations gracefully.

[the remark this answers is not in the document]
chrisbobbe: > ```diff
> +  // TODO: destructure and validate message data received from the Zulip server,
> +  // transforming it into a normalized, exact `Message` type strictly for app-
> +  // internal use. (See docs/architecture/crunchy-shell.md.)
> ```

This sounds right as an eventual goal, though.

chrisbobbe:
@gnprice I'm taking a look at this now, and will likely merge soon unless you know of obstacles and haven't written them down yet. 🙂

chrisbobbe:
Merged, thanks, @gnprice! I made two small tweaks:

I linked to facebook/flow#380 as Ray suggested at https://github.com/zulip/zulip-mobile/pull/3946#discussion_r390631209.

I got a Flow warning in

7d9e7f38c exampleData: In eg.streamMessage, take an optional stream.

that probably wasn't showing up when this PR was first composed; we've upgraded Flow since then. I addressed its complaint.

gnprice:
Thanks @chrisbobbe !

That Flow warning is pretty odd. I spent a few minutes investigating and wound up with a diagnosis and a cleaner workaround -- so I pushed that just now as a small tweak, d141be763.

Issue #3647, opened 2019-10-12
ios: Opening a notification doesn't navigate if app running in background


gnprice:
This is a lot like #3582... except on iOS. Observed just now when testing other notification changes, on my iPhone 7 running iOS 12.4.1. First observed in an alpha-only release made for testing a branch, but then reproduced on 26.8.131 (installed from TestFlight). Haven't yet tried older versions, but it may well have been present for a long time.

We have very little code of our own involved in this path in the iOS case. That code attempts to listen for an event emitted by library code when the user opens a notification, but we might have something wrong there; or there might be an issue in the library code. Debugging required.

Here's the current behavior I observe when the user gets a notification-causing message:
* If app is in foreground: No notification at all. (That seems like an issue of its own.)
* If app is in background: Notification appears. Opening it brings the app to foreground, but no navigation happens -- we stay on whatever screen we were on when the user switched away. That's this issue.
* If app isn't running at all: Notification appears. Opening it launches the app, and causes navigation to the appropriate screen, as desired.

chrisbobbe:
Linking conversation at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20push.20notifications/near/825050, which see.

A workaround has been found; `NotificationsIOS.consumeBackgroundQueue()` must be called after the app is put in the foreground. This could be implemented with an interval of a few seconds, or, much better, with a listener that tells us that the app has just been put in the foreground. Either way, this is a really superficial fix, and more permanent fixes should be considered first. That API has long been deprecated in the react-native-notifications library, and we may even be switching to a different library.

PR #3947, opened 2020-03-03
ios notifs: Consume "background queue" so notifs navigate from backgr…

chrisbobbe:
…ound.

This is a short-term fix that relies on a deprecated API. We should
upgrade to a newer version of react-native-notifications or use
@react-native-community/push-notification-ios instead.

For now-outdated docs on this, see
https://github.com/wix/react-native-notifications/commit/fe30ae5e03990eb46dc7c04a9ebe86ae0338c7d6#diff-d455ef90f34ba49d565c96e9f2cc8b98R53-R58.
The API was removed in release 2.0.6-shapshot-8 with commit
https://github.com/wix/react-native-notifications/commit/800669120bb7d5f5220ad821a323c3ab7f94342f,
presumably in response to
https://github.com/wix/react-native-notifications/issues/339.

Process an opened notification from the "background queue" so our
handleNotificationOpen will run, and the app will navigate, for a
notification opened while the app was in the background.

Fixes: #3647

### DIFF
--- a/src/boot/AppEventHandlers.js
+++ b/src/boot/AppEventHandlers.js
@@ -10,7 +10,11 @@ import type { Node as React$Node } from 'react';
 import type { Dispatch, Orientation as OrientationT } from '../types';
 import { connect } from '../react-redux';
 import { getUnreadByHuddlesMentionsAndPMs } from '../selectors';
-import { handleInitialNotification, NotificationListener } from '../notification';
+import {
+  handleInitialNotification,
+  NotificationListener,
+  notificationOnAppActive,
+} from '../notification';
 import { appOnline, appOrientation, appState, initSafeAreaInsets } from '../actions';
 import PresenceHeartbeat from '../presence/PresenceHeartbeat';
 
@@ -88,6 +92,9 @@ class AppEventHandlers extends PureComponent<Props> {
     dispatch(appState(state === 'active'));
     if (state === 'background' && Platform.OS === 'android') {
       NativeModules.BadgeCountUpdaterModule.setBadgeCount(unreadCount);
+    }
+    if (state === 'active') {
+      notificationOnAppActive();
     }
   };
 
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -132,6 +132,29 @@ export const handleInitialNotification = async (dispatch: Dispatch) => {
   dispatch(narrowToNotification(data));
 };
 
+export const notificationOnAppActive = () => {
+  if (Platform.OS === 'ios') {
+    try {
+      // Allow 'notificationOpened' events to be emitted when pressing
+      // a notification when the app is in the background.
+      //
+      // TODO: This API is deprecated in react-native-notifications release
+      // 2.0.6-snapshot.8; see #3647.
+      //
+      // We don't know the behavior if this is called before
+      // NotificationsIOS.requestPermissions(), so, catch any errors
+      // silently. Ray's investigation shows that it *shouldn't*
+      // throw, but may (https://github.com/zulip/zulip-mobile/pull/3947#discussion_r389192513).
+      NotificationsIOS.consumeBackgroundQueue();
+    } catch (e) {
+      logging.warn(e, {
+        message:
+          'Call to NotificationsIOS.consumeBackgroundQueue failed; pressed notification failed to navigate',
+      });
+    }
+  }
+};
+
 /**
  * From rn-notifications@1.5.0's RNNotifications.m.
  */


chrisbobbe:
If, soon, we can use a later version of react-native-notifications, or switch to @react-native-community/push-notification-ios, then this PR should be closed in favor of that solution.

src/boot/AppEventHandlers.js
```diff
@@ -95,6 +96,17 @@ class AppEventHandlers extends PureComponent<Props> {
     if (state === 'background' && Platform.OS === 'android') {
       NativeModules.BadgeCountUpdaterModule.setBadgeCount(unreadCount);
     }
+    if (state === 'active' && Platform.OS === 'ios') {
+      // TODO: This API is deprecated in react-native-notifications release
+      // 2.0.6-snapshot.8; see #3647.
+      try {
```
chrisbobbe: See motivation for try/catch at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20push.20notifications/near/825135.

ray-kraesig: I dug into this a bit – but, unfortunately, not enough to have a solid answer to that question.

I have learned that it probably *shouldn't* throw. The [definition of `consumeBackgroundQueue`](https://github.com/wix/react-native-notifications/blob/v1.4.0/RNNotifications/RNNotifications.m#L549) uses [`RCT_EXPORT_METHOD`](https://github.com/facebook/react-native/blob/v0.59.10/React/Base/RCTBridgeModule.h#L134-L175), which creates [an asynchronous method](https://github.com/facebook/react-native/blob/v0.59.10/React/Base/RCTBridgeModule.h#L208-L210) which [doesn't return any sort of value](https://github.com/facebook/react-native/blob/v0.59.10/React/Base/RCTModuleMethod.mm#L557-L562). (Even via callbacks, since it has no parameters.)

However, the invocation of a native method is eventually done [here](https://github.com/facebook/react-native/blob/v0.59.10/React/CxxModule/RCTNativeModule.mm#L91-L120), and the comment [here](https://github.com/facebook/react-native/blob/v0.59.10/React/CxxModule/RCTNativeModule.mm#L108) implies that some `NSException`s are propagated back to JS. I haven't found where that would happen.

src/boot/AppEventHandlers.js
```diff
@@ -95,6 +96,17 @@ class AppEventHandlers extends PureComponent<Props> {
...
+        // We don't know the behavior if this is called before
+        // NotificationsIOS.requestPermissions(), so, catch any errors silently.
+        NotificationsIOS.consumeBackgroundQueue();
+      } catch (e) {
+        // Do nothing
+      }
```
ray-kraesig: At the very least we need to feed these errors to Sentry. In particular, if we upgrade `react-native-notifications` to a later version without fixing this call, it *will* start throwing, and we may never notice! (Flow has no functional typing information for this module, so won't complain.)

ray-kraesig:
> If, soon, we can use a later version of react-native-notifications, or switch to @react-native-community/push-notification-ios, then this PR should be closed in favor of that solution.

Not soon enough, I'm sorry to say. :frowning_face: Fortunately, this fix is pretty small and isolated, and shouldn't cause us any problems when upgrading or switching.

src/boot/AppEventHandlers.js
```diff
@@ -89,6 +91,22 @@ class AppEventHandlers extends PureComponent<Props> {
...
+        // silently. Ray's investigation shows that it *shouldn't*
+        // throw, but may (https://github.com/zulip/zulip-mobile/pull/3947#discussion_r389192513).
+        NotificationsIOS.consumeBackgroundQueue();
+      } catch (e) {
+        logging.warn(e, {
+          message:
+            'Call to NotificationsIOS.consumeBackgroundQueue failed; pressed notification failed to navigate',
+        });
+      }
+    }
```
gnprice: Let's move all this into its own function, in our notifications code in `src/notifications/index.js` -- I think that's a better home for this complexity to live, next to the other code that has to deal with the intricacies of these libraries. Perhaps right after `handleInitialNotification`, which has a similar job.

Probably the `Platform.OS === 'ios'` condition belongs inside that function, too. Then this can just look like
```js
  if (state === 'active') {
    notificationOnAppActive();
  }
```
or some other name that indicates this is the function for doing notification-related things that we do when the app moves into the foreground / into state `active`. (And the details of what those things are is the job of the function's implementation, as part of the notification code.)

chrisbobbe: OK, sounds good!

chrisbobbe:
Thanks for the review! 🙂

> Have you tested this, and it seems to fix the issue?

Yes, I did. I'm setting up the dev server again right now, to test that these latest changes I've made haven't introduced bugs.

chrisbobbe:
OK, this should be ready to go. I changed the comments slightly; I put them inside the `try` block and added a bit about why the code is needed in the first place.

gnprice:

PR #3954, opened 2020-03-05
WIP: Something to play with on iOS, for navigation.

chrisbobbe:
EDIT: Huh, not sure why I named the branch "notification" ideas instead of "navigation" ones. 🤷 

@gnprice @ray-kraesig

Following discussions [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux.20.2F.20navigation) and [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation), here's something concrete to play with, made especially for iOS.

I think it would have been a good idea to settle some of those discussions before I moved the conversation to hypothetical "a -> b -> a -> b" sequences, etc., so hopefully this WIP PR can help toward that. 🙂 I don't think the concerns of a large stack building up are more urgent right now than fixing the unexpected "back" and "up" behaviors, which I still find really confusing and unhelpful after a lot of time with the UI.

I think we've [found](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/809169) that Apple doesn't have a strong, consistent rule that would require the "up" button/action, so I've left it out here, hopefully without too much controversy. The concern I raised in my [long story](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812671) about the #consciousness stream is resolved by putting the topic list button in the topic nav, not just the stream nav.

I think some disagreement remains on the question of the "up" button/action on Android, which I'll quote here.

Greg [said](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812610):
> > > * to pop the history stack: Left-arrow nav-bar button.
> > 
> > This is, IIRC, the opposite of the recommended behavior on Android; this button should take you "up", rather than chronologically "back".
>
> > The "back button in a browser" functionality is explicitly [the purpose of the system Back button/gesture.](https://stuff.mit.edu/afs/sipb/project/android/docs/design/patterns/navigation.html)
>
> So, it's absolutely right that that's the doctrine in the Android docs.
>
> I've grown skeptical that that's actually a great UI choice, though. I've also found it hard to locate real examples of apps seriously following that design.
>
> So I think it may be the right choice for us to leave that left-arrow in the nav bar as a "back" navigation. (Like the status quo, and like in the dictionary I described above.)

(See also Greg's example with a diagram.)

Ray [said](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812651):
> >Hmm, and with which behavior? Do you mean having it go "back" on iOS, but "up" on Android?>
> Exactly so.
> 
> > I think an "up" icon wouldn't greatly help. Nobody who hasn't been an Android developer knows about the contrasting terms "back" and "up" -- at least I'd never heard of them, or put together the concepts quite that way, in my years as an Android user before starting to develop for it.
> 
> Yes, that's *why* the different icon. :slight_smile: At least on the topic screen, being located next to the stream-name-over-topic-name construct should be a strong hint to its behavior. I'm less sure about the stream screen, though.

Before resuming this directly, I think it may help to return to my analysis of that Japanese dictionary app, and, from it, some guidelines I've offered to help reduce confusion for Zulip users. Greg, it looks like you [agree](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812606) with them. Ray, you [responded](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812594) to an implementation idea that followed them, but I'm curious to know what you think about the guidelines themselves. They seem broad enough to be a good place to explore any disagreements before moving forward, perhaps in the ["navigation" topic](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/826490) on CZO.

I [said](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/812590):
> A) It lets you navigate by the static structure of the pages, which I think you've called "the DAG of the overall navigation."
> B) It lets you navigate back through the user's history as it's created by some (maybe all) A-type navigation actions.
> C) A- and B-type navigation actions have distinct interfaces, so you're never unsure where a button or swipe will take you.

It seems to me that the Android [diagrams](https://stuff.mit.edu/afs/sipb/project/android/docs/design/patterns/navigation.html) we've discussed observe these.

I may have misunderstood, but some of the recent [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux.20.2F.20navigation/near/824899) on how many (2, N, 2N, etc.) routes would accumulate in the stack may be influenced by differing assumptions, with one being that a single back or up button could fulfill both A and B, which violates C. If I'm jumping to conclusions here, please let me know, but *after* you've responded to the above. 😉

### DIFF
--- a/src/nav/navActions.js
+++ b/src/nav/navActions.js
@@ -10,10 +10,9 @@ import type {
   UserOrBot,
   ApiResponseServerSettings,
 } from '../types';
-import { getSameRoutesCount } from '../selectors';
 
 export const navigateBack = () => (dispatch: Dispatch, getState: GetState): NavigationAction =>
-  dispatch(StackActions.pop({ n: getSameRoutesCount(getState()) }));
+  dispatch(StackActions.pop({}));
 
 // Other stack routes reached through `navReducer`:
 //    StackActions.push({ routeName: 'loading' });
@@ -24,6 +23,8 @@ export const navigateBack = () => (dispatch: Dispatch, getState: GetState): Navi
 /** Only call this via `doNarrow`.  See there for details. */
 export const navigateToChat = (narrow: Narrow): NavigationAction =>
   StackActions.push({ routeName: 'chat', params: { narrow } });
+
+export const navigatePopToTop = (): NavigationAction => StackActions.popToTop({});
 
 export const navigateToUsersScreen = (): NavigationAction =>
   StackActions.push({ routeName: 'users' });
--- a/src/title-buttons/ExtraNavButtonStream.js
+++ b/src/title-buttons/ExtraNavButtonStream.js
@@ -6,7 +6,7 @@ import type { Dispatch, Narrow, Stream } from '../types';
 import { connect } from '../react-redux';
 import { getStreams } from '../selectors';
 import NavButton from '../nav/NavButton';
-import { navigateToTopicList } from '../actions';
+import { navigatePopToTop } from '../nav/navActions';
 
 type Props = $ReadOnly<{|
   dispatch: Dispatch,
@@ -17,17 +17,13 @@ type Props = $ReadOnly<{|
 
 class ExtraNavButtonStream extends PureComponent<Props> {
   handlePress = () => {
-    const { dispatch, narrow, streams } = this.props;
-    const stream = streams.find(x => x.name === narrow[0].operand);
-    if (stream) {
-      dispatch(navigateToTopicList(stream.stream_id));
-    }
+    this.props.dispatch(navigatePopToTop());
   };
 
   render() {
     const { color } = this.props;
 
-    return <NavButton name="list" color={color} onPress={this.handlePress} />;
+    return <NavButton name="home" color={color} onPress={this.handlePress} />;
   }
 }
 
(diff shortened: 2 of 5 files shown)


ray-kraesig:
Well, I mostly agree; and in particular, I agree that this is an ideal that should be striven for. However, C) is really two propositions...

> C1) A- and B-type navigation actions have distinct interfaces[...].
> C2) [...] you're never unsure where a button or swipe will take you.

... and C1 is not *quite* sufficient to guarantee C2, because:

> B) It lets you navigate back through the user's history as it's created by ***some (maybe all)*** A-type navigation actions.

... the complexity of the rule determining which A-type actions are not invertible by a B-action is pretty much the degree to which proposition C2 fails to hold for B-actions.

Unfortunately, there are some pretty compelling technical and usability reasons not to let a stack of A-type actions build up indefinitely; and preventing that requires either 1) a structurally-guaranteed maximum navigation depth or 2) some non-invertable A-type actions. (Most apps are happy with option 1, but I don't think anyone thinks it's an appropriate choice for Zulip.)

chrisbobbe:
Thanks, Ray! I've replied [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation/near/833626).

chrisbobbe:
Closing; let's revisit these issues with a clean slate and fresh eyes.

Issue #3540, opened 2019-06-28
Remove invalid code for nonexistent realm/user_groups endpoint


timabbott:
While updating https://github.com/zulip/zulip/issues/10044, which originally had a bunch of endpoints sourced from the mobile app codebase, I determined that two of these endpoints don't exist and never have:

```
tabbott@coset:~/zulip-mobile$ git grep 'realm/user_groups/'
src/api/user_groups/getUserGroupById.js:export default (auth: Auth, id: number): Promise<mixed> => apiGet(auth, `realm/user_groups/${id}`);
tabbott@coset:~/zulip-mobile$ git grep 'users/me/user_groups'
src/api/user_groups/getUserGroups.js:export default (auth: Auth): Promise<mixed> => apiGet(auth, 'users/me/user_groups');
```

I'll also add the following two endpoints probably have no use in the mobile app, since it should be getting all the data for them from `register` anyway.
```
tabbott@coset:~/zulip-mobile$ git grep 'realm/emoji'
src/api/realm/getRealmEmojis.js:export default async (auth: Auth): Promise<ApiResponseRealmEmojis> => apiGet(auth, 'realm/emoji');
tabbott@coset:~/zulip-mobile$ git grep 'realm/filters'
src/api/realm/createRealmFilter.js:  apiPost(auth, 'realm/filters', { pattern, url_format_string: urlFormatString });
src/api/realm/getRealmFilters.js:  apiGet(auth, 'realm/filters');
```

Arpitpundir:
can i work on this one?

timabbott:
Just a quick bump on this (mostly because I'm closing the zulip/zulip issue that links to it).

gnprice:
Weird. Looks like these were added in #1431; no explanation given.

That was around the time that some user-group-related endpoints were added on the server, in zulip/zulip@b1603d289 and preceding commits. The actual URLs were different; but perhaps at some point a draft version of those changes had these URLs? (Which doesn't make a good reason -- the change should have waited for the server-side change to land, and should have documented what source it was relying on in any case -- but possibly explains how it happened.)

PR #3955, opened 2020-03-05
Remove unused code for various endpoints (two of which don't exist).

chrisbobbe:
Fixes: #3540 

I found a few more endpoints that @timabbott mentioned in https://github.com/zulip/zulip/issues/10044 as candidates for cleaning up, not mentioned in #3540. Also, while all of these endpoints are currently unused in the mobile app, it's likely that that will change for some of them, and I've marked them in the commit messages.

Two endpoints (`GET users/me/user_groups` and `GET realm/user_groups/${id}`) just don't exist and never have.


agrawal-d:
> Two endpoints (`GET users/me/user_groups` and `GET realm/user_groups/${id}`) just don't exist and never have.

I wonder how those got added in the first place 😂.

gnprice:
Thanks @chrisbobbe!

All of these removals LGTM; feel free to merge after editing the "don't merge" from the commit messages.

Specific thoughts:

    "I believe the app doesn't support editing these organization
    settings. Further, it should be just getting these data from
    /register with the right event_types value, so it's probably a
    mobile bug if we're accessing these:"
    
    This endpoint is included, but I don't think it's for fetching data,
    so I wanted to confirm that we should remove this code.

It's not for fetching data, but it is for editing some organization settings, and we don't have any UI for editing org settings. Eventually perhaps we'll want to build one, or at least some fragments of one, but in general I think it's a low priority. (This applies to several commits.)

Deleting a stream is a more likely nearer-term thing to add, and it's also a natural pair with `createStream` which we do use. So maybe that would be the one of these to keep around. It's also cheap to add later when we are going to use it, though, so perhaps best to stick to routes we're using or have concrete plans to use.

    toggleMuteStream exists, and is already being used. Favor that one,
    since other code that we do use follows that pattern; see
    togglePinStream, toggleStreamNotifications, etc.
    
    muteTopic and unmuteTopic is an exception; perhaps we would like to
    refactor that into a toggleMuteTopic.

I agree on both points. One thing to add: "toggle" is a bad name for what all those functions do. Typically in software use it means specifically to set a boolean flag to the opposite of whatever value it currently has; here, we've been using it inappropriately to just mean setting the flag to a value.

One good name might be `setTopicMute`. For `toggleMuteStream` and friends, a second independent improvement would be to take the opportunity to have fewer idiosyncratic names of our own, and use more names verbatim from the API -- so, together, perhaps a `setSubscriptionProperty` that takes a `property: 'is_muted' | 'pin_to_top' | ...` and a `value: boolean`.

chrisbobbe:
> I agree on both points. One thing to add: "toggle" is a bad name for what all those functions do. Typically in software use it means specifically to set a boolean flag to the opposite of whatever value it currently has; here, we've been using it inappropriately to just mean setting the flag to a value.

Yep, I agree. It seems odd that we've used "toggle" in any of these, given that the server isn't handling any of them as a toggle. On looking a bit closer than I did before, I think one good theory for the origin of these names is that the names were influenced by the UI; the functions named with 'toggle' are called when a RN toggle switch is pressed; `muteTopic` and `unmuteTopic` aren't named with 'toggle', and it turns out that they aren't associated with toggle switches. 

> One good name might be `setTopicMute`.

Makes sense.

> For `toggleMuteStream` and friends, a second independent improvement would be to take the opportunity to have fewer idiosyncratic names of our own, and use more names verbatim from the API -- so, together, perhaps a `setSubscriptionProperty` that takes a `property: 'is_muted' | 'pin_to_top' | ...` and a `value: boolean`.

This sounds good too.

chrisbobbe:
OK, I've pushed two commits that get us those `setTopicMute` and `setSubscriptionProperty` bindings, so I'll remove the WIP marker. :)

chrisbobbe:
I've just rebased this past a171a051f, which did a lot of the work toward fixing #3540. I also made more commits out of the last few, so it should be more readable now. 🙂

src/api/subscriptions/setSubscriptionProperty.js
```diff
@@ -2,11 +2,18 @@
...
+  value: boolean,
+): Promise<ApiResponse> =>
   apiPost(auth, 'users/me/subscriptions/properties', {
     subscription_data: JSON.stringify([
       {
-        property: 'in_home_view',
+        property,
         stream_id: streamId,
         value,
       },
```
gnprice: I do still like this consolidation! Even though it needs a bit of a wrinkle to maintain that "mute" vs. `in_home_view` facade mentioned in my previous comment.

I think a quick conditional within this function can take care of that reasonably cleanly, though. So this will take `is_muted` | `pin_to_top` | `push_notifications` as an argument, but if it's given `is_muted` it'll convert to `in_home_view` for what to actually pass to the server.

src/api/subscriptions/setSubscriptionProperty.js
```diff
@@ -2,11 +2,18 @@
 import type { ApiResponse, Auth } from '../transportTypes';
 import { apiPost } from '../apiFetch';
 
-export default async (auth: Auth, streamId: number, value: boolean): Promise<ApiResponse> =>
+export type SubscriptionProperty = 'in_home_view' | 'pin_to_top' | 'push_notifications';
+
+export default async (
```
gnprice, line 7: Let's give this a quick bit of jsdoc that points to the API documentation:
https://chat.zulip.org/api/update-subscription-settings

That has quite a high ratio of helpfulness to effort, so it's a good minimal form of documentation for these API bindings. (And is really all they need when they don't have significant wrinkles of their own to document -- no sense duplicating what's in the server API docs.)

gnprice (commented):
Thanks for the revision! Comments below.

src/message/messageActionSheet.js
```diff
@@ -103,15 +103,13 @@ deleteMessage.title = 'Delete message';
 deleteMessage.errorMessage = 'Failed to delete message';
 
 const unmuteTopic = async ({ auth, message }) => {
-  invariant(message.type === 'stream', 'unmuteTopic: got PM');
```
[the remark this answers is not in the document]
chrisbobbe: Ooh, thanks for the catch—I think this was a rebase conflict-resolution error.

chrisbobbe:

chrisbobbe:
New revision pushed! 🙂

WesleyAC (approved):
Looks great, thanks for doing this!

chrisbobbe:
Thanks for the reviews! Merged.

Issue #3939, opened 2020-02-29
Profile icon breaks for group PMs with large no. of recipients


agrawal-d:
PMs list screen:

![image](https://user-images.githubusercontent.com/7714968/75606143-dab2a800-5b0f-11ea-9bee-cb897908fcf3.png)

agrawal-d:
@gnprice said
>One step toward addressing these: I'm curious how group PMs like these wind up looking in the webapp.

The web app doesn't display icons at all, so it's not an issue there:
Screenshot of list of private messages -
![image](https://user-images.githubusercontent.com/7714968/75606205-780ddc00-5b10-11ea-8b62-571264910693.png)

gnprice:
Thanks @agrawal-d for spotting and reporting this!

I've never been much of a fan of this initials-based pseudo-icon in any case. I'd be happy to see it completely reimagined. But in any case we should fix this bug. :-)

Maybe:
* Use just one initial, not two: use the very first letter of the name (so from the person's first name, in the case of Western name conventions)
* If there are up to 4, write the initials.
* If there are more than that, then write only 3, followed by a plus sign.

Beyond about 4 I think the lists of initials get pretty unreadable anyway. :wink:

PR #3958, opened 2020-03-07
group-pm: Fix profile icon breaking for large no.of recepients

Maskedman99:
When there are large no.of recipients or people with large
no.of initials in their names, the profile Icon is overflowed
with letters. This PR fixes this by taking only the first Initial
from each name and if no.of recipients is > 4 adding a '+'
charector to the end.

Fixes: #3939

### DIFF
--- a/src/common/GroupAvatar.js
+++ b/src/common/GroupAvatar.js
@@ -5,7 +5,6 @@ import { Text, StyleSheet, View } from 'react-native';
 import type { Node as React$Node } from 'react';
 import Touchable from './Touchable';
 import { colorHashFromString, foregroundColorFromBackground } from '../utils/color';
-import { initialsFromString } from '../utils/misc';
 
 const styles = StyleSheet.create({
   frame: {
@@ -18,12 +17,23 @@ const styles = StyleSheet.create({
 });
 
 type Props = $ReadOnly<{|
-  names: string,
+  names: string[],
   size: number,
   shape: 'rounded' | 'square',
   children?: React$Node,
   onPress?: () => void,
 |}>;
+
+/** Private; exported only for tests. */
+export const initialsForGroupIcon = (names: string[]): string => {
+  const initials = names.map(item =>
+    item === '' ? '' : String.fromCodePoint(item.codePointAt(0)).toUpperCase(),
+  );
+  if (initials.length > 4) {
+    initials[3] = '…';
+  }
+  return initials.slice(0, 4).join('');
+};
 
 /**
  * Renders a text avatar based on a single or multiple user's
@@ -48,7 +58,7 @@ export default class GroupAvatar extends PureComponent<Props> {
       height: size,
       width: size,
       borderRadius: shape === 'rounded' ? size / 8 : 0,
-      backgroundColor: colorHashFromString(names),
+      backgroundColor: colorHashFromString(names.join(', ')),
     };
     const textSize = {
       fontSize: size / 3,
@@ -58,7 +68,7 @@ export default class GroupAvatar extends PureComponent<Props> {
     return (
       <Touchable onPress={onPress}>
         <View style={[styles.frame, frameSize]}>
-          <Text style={[styles.text, textSize]}>{initialsFromString(names)}</Text>
+          <Text style={[styles.text, textSize]}>{initialsForGroupIcon(names)}</Text>
           {children}
         </View>
       </Touchable>
--- a/src/utils/misc.js
+++ b/src/utils/misc.js
@@ -24,9 +24,6 @@ export function deeperMerge<K, V>(obj1: { [K]: V }, obj2: { [K]: V }): { [K]: V 
   );
 }
 
-export const initialsFromString = (name: string): string =>
-  (name.match(/\S+\s*/g) || []).map(x => x[0].toUpperCase()).join('');
-
 export function groupItemsById<T: { +id: number }>(items: T[]): { [id: number]: T } {
   return objectFromEntries(items.map(item => [item.id, item]));
 }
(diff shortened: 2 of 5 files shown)


Maskedman99:
Here's how the solution looks after implementing.
Screenshot_20200307-131302_Zulip (debug)

src/utils/misc.js
```diff
@@ -32,3 +32,18 @@ export function groupItemsById<T: { +id: number }>(items: T[]): { [id: number]:
 }
 
 export const isValidEmailFormat = (email: string = ''): boolean => /\S+@\S+\.\S+/.test(email);
+
+export const initialsForGroupIcon = (name: string): string => {
```
ray-kraesig: Despite `initialsFromString` using it, `name` isn't a good description of what this function takes as an argument.

src/utils/misc.js
```diff
@@ -32,3 +32,18 @@ export function groupItemsById<T: { +id: number }>(items: T[]): { [id: number]:
...
+  let x = '';
+  for (let i = 0; i < Name.length; i++) {
+    if (i < 4) {
+      x += Name[i][0].toUpperCase();
+    } else {
+      x = x.slice(0, -1);
+      x += '+';
+      break;
+    }
+  }
```
ray-kraesig: This can be simplified significantly by testing `Name.length` once, and using `.map()` rather than an explicit loop.

Extracting a single character from a string in JavaScript with `string[n]` is almost invariably always a bug; this should use `.codePointAt(0)` instead. (Its absence is a bug in `initialsFromString`!) If a user has a name beginning with a character not in the Basic Multilingual Plane (e.g., an emoji), `string[0]` will break up a surrogate pair, which will cause many problems.

src/common/GroupAvatar.js
```diff
@@ -57,7 +57,7 @@ export default class GroupAvatar extends PureComponent<Props> {
     return (
       <Touchable onPress={onPress}>
         <View style={[styles.frame, frameSize]}>
-          <Text style={[styles.text, textSize]}>{initialsFromString(names)}</Text>
+          <Text style={[styles.text, textSize]}>{initialsForGroupIcon(names)}</Text>
```
ray-kraesig: Since this was the only occurrence of `initialsFromString`, this commit should probably also remove it.

src/utils/misc.js
ray-kraesig: **Erratum:** `.codePointAt(0)` returns the numeric value of the first code point. There are several options to extract the first character; perhaps the most straightforward is to check for a high surrogate, and if one is found, to take the following low surrogate as well.

Maskedman99:
@ray-kraesig thanks for reviewing, I've done the suggested modifications.

Maskedman99:
@ray-kraesig Thanks, I've modified the code, now GroupPmConversationItem.js will send an array of names to GroupAvatar.js this will solve the problem around users having ", " in their names as their is no longer splitting of strings.

Maskedman99:
@ray-kraesig I resolved the conflicts in the PR, but Travis CI is showing a test failed inside ```src/utils/__tests__/async-test.js``` which wasn't happening before. 
Also surprisingly, the build had passed for https://github.com/zulip/zulip-mobile/commit/b6c64ee90b8c7e9a59185dcfc40c465f9cfbd6b0 (https://travis-ci.org/github/zulip/zulip-mobile/builds/667574878?utm_medium=notification&utm_source=github_status) which differs with the present commit in that I removed an unused empty line in it.

ray-kraesig:
> @ray-kraesig I resolved the conflicts in the PR, but Travis CI is showing a test failed inside `src/utils/__tests__/async-test.js` which wasn't happening before.

That's quite strange, and almost certainly not related to anything you've done. The tests succeeded on a rerun; I've filed a separate issue.

src/common/GroupAvatar.js
```diff
@@ -36,6 +35,17 @@ type Props = $ReadOnly<{|
  * @prop children - If provided, will render inside the component body.
  * @prop onPress - Event fired on pressing the component.
  */
+
+export const initialsForGroupIcon = (listOfNames: Array<string>): string => {
+  const x = listOfNames.slice(0, 4);
+  const initials = x.map(item => item.codePointAt(0));
+  if (listOfNames.length > 4) {
+    initials.pop();
+    initials.push(43); // Decimal CodePoint of '+' is 43
```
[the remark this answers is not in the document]
gnprice: "+" was my suggestion here: https://github.com/zulip/zulip-mobile/issues/3939#issuecomment-593581780

I think I was influenced by citation practices in academic computer science :grin: -- with commonly-used LaTeX styles, if you cite a paper with one author it'll look like "[Kra20]" with the first few letters of their name; if two to four (maybe three?) authors, it'll use their initials; if more, it'll use the first three authors' initials and a "+".

I think if you want to read the list of initials as like a name, then a "+" works better in a name than a "…" does.

But maybe that isn't the best way to read them. I don't have a strong preference.

src/common/GroupAvatar.js
```diff
@@ -18,13 +17,24 @@ const styles = StyleSheet.create({
...
   shape: 'rounded' | 'square',
   children?: React$Node,
   onPress?: () => void,
 |}>;
 
+export const initialsForGroupIcon = (names: string[]): string => {
+  const initialsInCodePoint = names.map(item => item.codePointAt(0));
+  if (initialsInCodePoint.length > 4) {
+    initialsInCodePoint[3] = 8230; // Decimal CodePoint of '…' is 8230
+  }
```
ray-kraesig, line 31: This is still, as I noted earlier, "trying to work with data that's partly in one world and partly in another."

I'm not seeing a good reason not to put the `String.fromCodePoint(...)` call in the first `.map`.

ray-kraesig:
... and I see you've expanded the `no.` in the commit at some point. Excellent. Merging!

Note that there are still issues here (`Z̈ed` will be shown as a `Z`, rather than `Z̈`; a prefixed `🏳️‍🌈` will be truncated to `🏳`), but that's a problem we have everywhere in our codebase that involves initials. :slightly_frowning_face:

PR #3959, opened 2020-03-09
Support `zh-Hant`

ray-kraesig:
Migrate `zh` to `zh-Hans`, and add support for `zh-Hant`.

Extracted from existing PR #3915.


gnprice:
Thanks @ray-kraesig !

These changes look great. The migration calls for manual testing that it works; you confirmed just now that you've done that in the context of #3915, and that sounds good to me. Merging!

PR #3960, opened 2020-03-09
eslint: Stop using react-native plugin.

chrisbobbe:
In abc043253, we disabled the only rule that remained enabled from
this plugin. So, remove the plugin. Leave the two existing
commented-out lines to enable other rules, which describe how we'd
like to potentially use them in the future. See:
  https://github.com/zulip/zulip-mobile/pull/3901#discussion_r388026654

----------------

From https://github.com/zulip/zulip-mobile/pull/3901#issuecomment-596799693, where Greg said:

> I've also dropped the new commit that removes the eslint react-native plugin, because I realized that it doesn't yet remove it from package.json and yarn.lock 🙂 -- only from the eslintrc. That can be a good quick followup.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -6,7 +6,7 @@ extends:
   - ./tools/formatting.eslintrc.yaml
 
 plugins:
- - react-native
+# - react-native
  - jest
  - flowtype
 
--- a/package.json
+++ b/package.json
@@ -97,7 +97,6 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-react": "^7.14.2",
-    "eslint-plugin-react-native": "^3.7.0",
     "eslint-plugin-spellcheck": "0.0.14",
     "flow-bin": "^0.92.0",
     "flow-coverage-report": "^0.6.0",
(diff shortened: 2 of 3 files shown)


ray-kraesig:
If we're going to leave the two unused rules as comments, it probably makes sense to leave the `react-native` plugin inclusion in place as a comment as well.

That said, since we're looking at it – are there any rules under [the list of supported rules](https://github.com/Intellicode/eslint-plugin-react-native#list-of-supported-rules) that we *would* want?

chrisbobbe:
Greg identified those two from the list, as the only ones we might want. I'll quote here, since GitHub's UI made the two links in my opening comment appear identical (and therefore easy to skip over the first one); they're links to two different comments on the same PR:

Greg [says](https://github.com/zulip/zulip-mobile/pull/3901#discussion_r388026654):
> The next thing I notice here is that this means we're no longer using this ESLint plugin at all. :-) So should probably drop it entirely.
> 
> Looking back through the [list of all its rules](https://github.com/Intellicode/eslint-plugin-react-native/tree/master/docs/rules)... I think the only two we actually want are the two commented out below this one. Maybe we'd want to keep these commented-out config lines, as a reminder for a possible future where we try to bring one or both of those rules into use. But in the meantime, it'll be good to make `node_modules` a bit smaller, and `tools/test lint` perhaps a bit faster.

Leaving the `react-native` plugin inclusion in place, but commented out, sounds good to me.

chrisbobbe:
OK, I've re-added the react-native line and commented it out.

gnprice:
Thanks both!

I've just pushed 88cafe4df, too, to add a bit to the comments here.

PR #3963, opened 2020-03-10
emoji search: Search for query-string anywhere in the emoji name.

### LINKED ISSUE #3948: emoji autocomplete only searches at start of each emoji name
on web:
image

on mobile:
image

Issue: 
We filter emoji using `x.indexOf(query) === 0`, which only checks if emoji name starts with `query`. We should use `x. includes(query)`, so that all emojis which contains `query` are included. 

Moved here from:
https://github.com/zulip/zulip-mobile/pull/3847#discussion_r371967822

ray-kraesig:
Previously, emoji queries required an exact initial match. Instead,
match against the query-string's occurrence anywhere in the name.

This imperfectly mimicks the webapp's behavior, in that prefix-matches
are considered better than normal matches. The webapp leaves some
sort-ordering unspecified, however, and also has a set of "popular
emoji" which are always considered a good match if they're considered
a match at all. Properly mimicking the webapp is left as #3833.

Test data has been taken from the original PR #3836 by @sumukhah,
whose choice of sort order is also maintained to simplify tests.

Fixes #3948.

Co-Authored-By: sumukhah <sumukha214@gmail.com>

### DIFF
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -51,16 +51,16 @@ describe('getFilteredEmojiNames', () => {
       'gorilla',
       'got_it',
       'agony',
+      'all_good',
       'dango',
       'dragon',
       'dragon_face',
-      'virgo',
-      'all_good',
+      'easy_come_easy_go',
+      'heart_of_gold',
+      'merry_go_round',
       'octagonal_sign',
       'synagogue',
-      'merry_go_round',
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks @ray-kraesig ! Merged, with one more commit on top to give it the desired sort order -- it turned out to be just about as simple to do as the other one. :slightly_smiling_face:

Issue #2834, opened 2018-07-25
Fix bogus year-1970 timestamp in how we aggregate user presence


gnprice:
See `getAggregatedPresence`:
https://github.com/zulip/zulip-mobile/blob/1baf7aec1/src/utils/presence.js#L16

As the comment there says, this should have the same behavior that the webapp implements in `status_from_timestamp`.

Comparing those two functions, there are two differences in behavior where we should adjust to match:
* When a user doesn't have any presence info more recent than OFFLINE_THRESHOLD_SECS, i.e. 140 seconds, we return a timestamp of 0. The webapp instead records the time of the latest presence we have -- making it possible to distinguish between someone being away for 3 minutes, 3 hours, or 3 weeks.
* We include a `client` property in the returned object, corresponding to whichever client we ended up taking the information from. This can be pretty arbitrary if a user has several clients online, so it doesn't mean much; probably anything that used that property would be making a mistake. We should follow the webapp in not returning that.

On the other hand, we can continue to skip computing the `mobile` property; as our jsdoc on `ClientPresence` says, the feature that was meant for was never completed as planned and is now legacy.

The timestamp fix may confuse some other code of ours that relies on the current behavior; I haven't traced through everything. Part of resolving this issue will be to check for that, and if so to (a) when first making this fix, adjust that code in the same commit so its behavior doesn't change; then (b) as a followup, make everything take full advantage of the new more-complete information.

gnprice:
A few weeks ago, @rishig reported (out of band):

> ​​I feel like active/inactive state on mobile is sometimes not correct (e.g. currently it thinks I’m offline and have been for 49 years, and also got the report from [a user])

And just now I was looking for some other issue, and happened across this issue's title! And this detail above, which I spotted from reading the code, looks awfully suspicious:

> When a user doesn't have any presence info more recent than OFFLINE_THRESHOLD_SECS, i.e. 140 seconds, we return a timestamp of 0.

A quick `git diff 1baf7ae src/utils/presence.js` shows that that logic hasn't changed since the version I referred to above, so we're still doing that. And that timestamp of 0 sounds awfully likely to cause the symptom Rishi reported.

Bumping this to a priority issue, since we now know of user-visible symptoms that this is probably causing.

PR #3966, opened 2020-03-10
Fix bogus year-1970 timestamp in how we aggregate user presence

chrisbobbe:
presence: Fix timestamp being Unix zero (50 years ago).

This is a user-visible symptom, but it's kind of a rare case that we
wouldn't expect to see normally. Sometimes, a user is displayed to
be "last active" 50 years ago, which is bogus. It turns out that it
only happens when a presence event is sent from the server (meaning
it should be as up-to-date as possible; we wouldn't expect it to be
more than a few seconds off), but it's more than 140 seconds old.

Presence events coming from the server don't have the `aggregated`
field, like presence objects do in the /register response and
presence report responses. So we have our own function,
getAggregatedPresence, to calculate it for us. This is analogous to
the status_from_timestamp function in the web app.

As suggested in the parent, a bug seems to have been introduced into
this function by a hasty use of `.reduce`, when a `.forEach` would
work perfectly well.

So, to fix:

1) Use .forEach instead of reduce (that's the parent).

2) Assume that there is at least one ClientPresence in `presence`,
and that its timestamp is sane. This is handled by the server; it
seems reasonable that no event would have been emitted if there
weren't sane data to transmit.

3) Make the returned timestamp be whatever the latest timestamp is,
out of all the ClientPresence objects.

4) Maintain `status` as the greatest status reported within
OFFLINE_THRESHOLD_SECS (with `active` > `idle` > `offline`), or
`offline` if none was within the threshold, as explained in the
comment and implemented in the web app.

Note that `#4` surfaced a subtlety in the intent stated in the
comment, and the behavior of the web app, which prompted changing a
test case. The `timestamp` and the `status` fields of the returned
object won't necessarily be sourced from the same ClientPresence
object from the input. Take an example where one object has status
`idle`, and another has status `active`, and both timestamps are
within OFFLINE_THRESHOLD_SECS, but the `idle` one's timestamp is
later than the `active` one's. In that case, the aggregated `status`
would be `active`, but the timestamp would be the timestamp from the
`idle` object. This may or may not be exactly what we want; the
information conveyed is inaccurate.

Also, note that further investigation or broader refactors were
avoided because of https://github.com/zulip/zulip/issues/13734,
"Improve presence system's scalability for large realms", which
signals substantial future changes to the Presence API. See also the
discussion at
https://chat.zulip.org/#narrow/stream/3-backend/topic/presence.20API/near/811593.

Fixes: #2834

### DIFF
--- a/src/utils/presence.js
+++ b/src/utils/presence.js
@@ -54,12 +54,14 @@ export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
   let client = '';
   let timestamp = 0;
 
+  const dateNow = Date.now();
+
   for (const [device, devicePresence] of objectEntries(presence)) {
     if (device === 'aggregated') {
       continue;
     }
 
-    const age = Date.now() / 1000 - devicePresence.timestamp;
+    const age = dateNow / 1000 - devicePresence.timestamp;
     if (timestamp < devicePresence.timestamp) {
       timestamp = devicePresence.timestamp;
     }
--- a/src/utils/presence.js
+++ b/src/utils/presence.js
@@ -28,7 +28,7 @@ const OFFLINE_THRESHOLD_SECS = 140;
  *   https://zulip.readthedocs.io/en/latest/subsystems/presence.html
  *
  * This logic should match `status_from_timestamp` in the web app's
- * `static/js/presence.js`.
+ * `static/js/presence.js` at 1ae07b93d^.
  */
 export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
   /* Out of the ClientPresence objects found in `presence`, we consider only
--- a/src/utils/presence.js
+++ b/src/utils/presence.js
@@ -5,6 +5,7 @@ import distanceInWordsToNow from 'date-fns/distance_in_words_to_now';
 
 import type { ClientPresence, UserPresence, PresenceStatus, UserStatus } from '../types';
 import { ensureUnreachable } from '../types';
+import objectEntries from './objectEntries';
 
 /** The relation `>=`, where `active` > `idle` > `offline`. */
 const presenceStatusGeq = (a: PresenceStatus, b: PresenceStatus): boolean => {
@@ -30,7 +31,7 @@ const OFFLINE_THRESHOLD_SECS = 140;
  * This logic should match `status_from_timestamp` in the web app's
  * `static/js/presence.js` at 1ae07b93d^.
  */
-export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
+export const getAggregatedPresence = (presence: UserPresence): ClientPresence => {
   /* Out of the ClientPresence objects found in `presence`, we consider only
    * those with a timestamp newer than OFFLINE_THRESHOLD_SECS; then of
    * those, return the one that has the greatest PresenceStatus, where
@@ -39,20 +40,28 @@ export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
    * If there are several ClientPresence objects with the greatest
    * PresenceStatus, an arbitrary one is chosen.
    */
-  Object.keys(presence)
-    .filter((client: string) => client !== 'aggregated')
-    .reduce(
-      (aggregated, client: string) => {
-        const { status, timestamp } = presence[client];
-        if (Date.now() / 1000 - timestamp < OFFLINE_THRESHOLD_SECS) {
-          if (presenceStatusGeq(status, aggregated.status)) {
-            return { client, status, timestamp };
-          }
-        }
-        return aggregated;
-      },
-      { client: '', status: 'offline', timestamp: 0 },
-    );
+
+  let status = 'offline';
+  let client = '';
+  let timestamp = 0;
+
+  for (const [device, devicePresence] of objectEntries(presence)) {
+    if (device === 'aggregated') {
+      continue;
+    }
+
+    const age = Date.now() / 1000 - devicePresence.timestamp;
+    if (age < OFFLINE_THRESHOLD_SECS) {
+      if (presenceStatusGeq(devicePresence.status, status)) {
+        client = device;
+        status = devicePresence.status;
+        timestamp = devicePresence.timestamp;
+      }
+    }
+  }
+
+  return { client, status, timestamp };
+};
 
 export const presenceToHumanTime = (presence: UserPresence, status?: UserStatus): string => {
   if (!presence || !presence.aggregated) {
--- a/src/utils/presence.js
+++ b/src/utils/presence.js
@@ -32,13 +32,22 @@ const OFFLINE_THRESHOLD_SECS = 140;
  * `static/js/presence.js` at 1ae07b93d^.
  */
 export const getAggregatedPresence = (presence: UserPresence): ClientPresence => {
-  /* Out of the ClientPresence objects found in `presence`, we consider only
-   * those with a timestamp newer than OFFLINE_THRESHOLD_SECS; then of
-   * those, return the one that has the greatest PresenceStatus, where
-   * `active` > `idle` > `offline`.
+  /* Gives an artificial ClientPresence object where
+   * - `timestamp` is the latest `timestamp` of all ClientPresence
+   *   objects in the input, and
+   * - `status` is the greatest `status` of all *recent* ClientPresence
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Marking P1 because the issue is marked P1.

gnprice:
Thanks! I like this code, assuming our new understanding is right.

I have a couple of questions:

    2) Assume that there is at least one ClientPresence in `presence`,
    and that its timestamp is sane. This is handled by the server; it
    seems reasonable that no event would have been emitted if there
    weren't sane data to transmit.

This sure makes sense. Can you confirm this is in fact the server's behavior? (I.e. there's always at least one item in the `presence` event.)

Alternatively or additionally, we could make that part of the interface/contract of `getAggregatedPresence`, and its caller (the reducer) could just ignore an event with no data, treating it as a no-op.

    4) Maintain `status` as the greatest status reported within
    OFFLINE_THRESHOLD_SECS (with `active` > `idle` > `offline`), or
    `offline` if none was within the threshold, as explained in the
    comment and implemented in the web app.
    
    Note that `#4` surfaced a subtlety in the intent stated in the
    comment, and the behavior of the web app, [...]

Is the new behavior the same as the webapp's? That'd be a good fact to know.

gnprice:
Smaller things:

    presence [nfc]: Use .forEach instead of .reduce in getAggregatedPresence.

This could say just "instead of a .reduce"; I think the "presence" prefix already carries the useful information for the summary that the name does.


    Preparation for the child commit

There's an asymmetry between a commit's message referring to a parent and a child: the parent is an immutable fact about the commit, whereas it can always gain more children. There's a new child any time you make a PR/topic branch and happen to start from that commit; there might even be a new child added to the permanent shared history, if we make a stable-release branch and happen to start it from that commit.

So for that reason I prefer saying "the next commit", "an upcoming commit", and so on, even though I often say "the parent commit" as well as things like "the preceding commit". Saying "next commit" or "upcoming commit" is still a bit indeterminate, but the terms match the indeterminacy, whereas "child" is jargon.


    Preparation for the child commit, where we squash a bug where a
    timestamp can end up up being Unix time zero, which is 50 years ago.

I think the key point as of this commit isn't the impact of the bug, but the fact that it was caused by this loop being too terse / trying to be too clever. Or put another way, that we fix it by decoupling some of this logic.

The impact can be left as suspense for the reader to resolve when they look at that next commit :wink:

src/utils/presence.js
```diff
@@ -39,20 +39,30 @@ export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
...
-      (aggregated, client: string) => {
-        const { status, timestamp } = presence[client];
-        if (Date.now() / 1000 - timestamp < OFFLINE_THRESHOLD_SECS) {
-          if (presenceStatusGeq(status, aggregated.status)) {
-            return { client, status, timestamp };
-          }
+    .filter((c: string) => c !== 'aggregated')
+    .forEach(c => {
+      const devicePresence = presence[c];
+      const age = Date.now() / 1000 - devicePresence.timestamp;
```
gnprice: One followup thought: `Date.now()` should be called once at the top of the loop, like `const now = Date.now()`. That way the whole function is working from the same value of "now", which helps keep it sane to reason about.

src/utils/presence.js
```diff
@@ -39,20 +39,30 @@ export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
...
-    .reduce(
-      (aggregated, client: string) => {
-        const { status, timestamp } = presence[client];
-        if (Date.now() / 1000 - timestamp < OFFLINE_THRESHOLD_SECS) {
-          if (presenceStatusGeq(status, aggregated.status)) {
-            return { client, status, timestamp };
-          }
+    .filter((c: string) => c !== 'aggregated')
+    .forEach(c => {
+      const devicePresence = presence[c];
```
gnprice: Another followup thought ... OK, actually, three closely related thoughts, which I started writing as separate comments but are better as one. This loop can be further improved by:
* moving the condition in `filter` into an ordinary if-statement inside the loop;
* using `objectEntries` (i.e. `Object.entries` but with a useful type signature) to get `c` and `devicePresence` simultaneously;
* using a real language-level loop, a `for .. of` loop, instead of `.forEach`. See 245ea9c4c and the comment that added to our ESLint config.

src/utils/presence.js
```diff
@@ -39,20 +39,30 @@ export const getAggregatedPresence = (presence: UserPresence): ClientPresence =>
...
+      const age = Date.now() / 1000 - devicePresence.timestamp;
+      if (timestamp < devicePresence.timestamp) {
+        /* eslint-disable prefer-destructuring */
+        timestamp = devicePresence.timestamp;
```
gnprice: That is kind of wacky that the lint rule wants to change this!

It looks like that's optional:
https://eslint.org/docs/rules/prefer-destructuring
and can be disabled with something like
```yaml
    prefer-destructuring:
    - error
    - AssignmentExpression: {array: false, object: false}
```
So maybe let's do that.

src/utils/presence.js
chrisbobbe: > using objectEntries (i.e. Object.entries but with a useful type signature) to get c and devicePresence simultaneously;

Great that we've taken the care to make a useful type signature, I'll use that! Right; the web app uses a `for .. of` loop with Object.entries, if I recall, and I really wanted to mirror that, but Flow didn't like it.

chrisbobbe:
OK, I think I've resolved the smaller things you've pointed out.

For https://github.com/zulip/zulip-mobile/pull/3966#issuecomment-598511784, there's an additional wrinkle that I've noted in the commit messages... Apparently, when I first authored this, I wasn't looking at the most up-to-date master, and the particular function this is based on has been removed, and its logic handled elsewhere. The web app has inevitably opted in to "slim presence" before we have. This PR should still accurately reflect what was going on before that, but I think using "slim presence" is out of scope of this fix.

> > `[...] it seems reasonable that no event would have been emitted if there weren't sane data to transmit.`

> Can you confirm this is in fact the server's behavior?

The way to establish this most strongly is to read the server code at every commit, and while I remember doing a little of that, and concluding that this is somewhat likely to be true, I think it's much easier to do what you suggested, along the lines of the "crunchy shell":

> Alternatively or additionally, we could make that part of the interface/contract of `getAggregatedPresence`, and its caller (the reducer) could just ignore an event with no data, treating it as a no-op.

chrisbobbe:
> ```
> 4) Maintain `status` as the greatest status reported within
> OFFLINE_THRESHOLD_SECS (with `active` > `idle` > `offline`), or
> `offline` if none was within the threshold, as explained in the
> comment and implemented in the web app.
> 
> Note that `#4` surfaced a subtlety in the intent stated in the
> comment, and the behavior of the web app, [...]
> ```
> 
> Is the new behavior the same as the webapp's? That'd be a good fact to know.

Yes, I noticed an oddness in the webapp's behavior and adapted our code to mimic it. I've edited the commit message and the comment to make this (hopefully) clearer.

src/utils/presence.js
```diff
@@ -28,31 +29,52 @@ const OFFLINE_THRESHOLD_SECS = 140;
...
-        return aggregated;
-      },
-      { client: '', status: 'offline', timestamp: 0 },
-    );
+
+  let status = 'offline';
+  let client = '';
+  let timestamp = 0;
+
+  const dateNow = Date.now();
```
gnprice, line 57: small thing in commit message:

    This is a fresh optimization that wasn't done before this series of
    commits, so it gets its own commit.

This isn't really an optimization -- it's about correctness and consistent behavior. In this function in particular, it avoids the possibility of treating one presence report as too old and another as not too old, when the latter is actually older than the former. That's... harmless in this case, I think, but it's best not to have to worry about that class of things in general.

src/utils/presence.js
```diff
@@ -28,31 +29,52 @@ const OFFLINE_THRESHOLD_SECS = 140;
...
+
+  for (const [device, devicePresence] of objectEntries(presence)) {
```
gnprice, line 59: This is great but the commit message should be updated to match :wink:

chrisbobbe:
Thanks for the review! Here are my latest changes.

gnprice:
And merged! Thanks again.

src/utils/presence.js
chrisbobbe: Absolutely!

Issue #3457, opened 2019-04-15
Narrow sometimes lands far in the past


gnprice:
A user reports:

>  if Zulip Chat is open and a new message is incoming, that it not direct scroll to the message if you tab it in the unread message page, the message is not there. If i close, open Zulip, it works.
Similar is if you open any topic it won't scroll to bottom to newest message, you have do it manually.

When asked

> Is the message there if you scroll down to it, or is the message not present even when you scroll?

they add:

> Message is present, but need to scroll down multiple times until bottom to reach the new message. On each scroll step i see the arrow down icon to jump to the bottom. If press that icon it just jumps som already loaded messages down and so on.... after repeat scroll down multiple times, bottom is reached.

From further discussion I think the scenario is that there's just one (or maybe a handful) of unread messages, in a conversation with many older messages; and the narrow lands far in the past, many messages before the unreads.

They're using v23.3.112 on Android 9.

The symptom is reminiscent of some ancient bugs: #1545, #2146 (minus the mark-as-read part), #2699. The latest of those was fixed last June, though, and I don't think we've seen such symptoms since then, so the cause is likely unrelated to those.

gnprice:
A user just reported a pattern that might be a description of this issue:

> Occasionally conversations "autoscroll" to the top (or at least quite far back) in a topic, requiring manually scrolling back down

I'm not sure whether they're referring to what they first see when narrowing to the conversation, which is what would make it the same as this issue. It kind of sounds like they mean they're already looking at the messages when they get scrolled way up, which would be a different set of symptoms.

gnprice:
User feedback today (on a related [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/start.20position.20when.20reading.20topics/near/746531)):
> BTW, I still get this almost every time I use the app and I think it contributes to a perception that the Zulip app is broken and janky

I've also found in my own use that this is one of the top annoyances I've noticed in recent weeks.

In my experience it's often just 1-2 screenfuls back.

gnprice:
I just tried a few times to reproduce this, narrowing to various streams and PM conversations on chat.zulip.org. I wasn't able to do so.

So one very useful form of progress on this would be to experiment more, and find a way to reproduce the issue at will.

gnprice:
Some user reports from that same [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/start.20position.20when.20reading.20topics/near/746531):

[...]

PR #3967, opened 2020-03-12
Fix instance of #3457, "Narrow sometimes lands far in the past", for muted streams.

chrisbobbe:
Please see the motivation for this at https://github.com/zulip/zulip-mobile/pull/3913#issuecomment-597910545, and in the commit message.

The second and third commits are marked [don't merge]. In removing these single-use helpers, I'm not sure whether the advantage of immediately simplifying the code (no more flipping back and forth between files, mainly) outweighs the concerns of a) losses in test coverage and b) future reusability, especially since a) the core logic of each is simple, and b) we haven't opted to reuse them since they were introduced over two years ago.

### DIFF
--- a/src/message/messageSelectors.js
+++ b/src/message/messageSelectors.js
@@ -12,7 +12,6 @@ import {
 import * as logging from '../utils/logging';
 import { getShownMessagesForNarrow } from '../chat/narrowsSelectors';
 import renderMessages from './renderMessages';
-import { findFirstUnread } from '../utils/message';
 import type { JSONable } from '../utils/jsonable';
 import { ALL_PRIVATE_NARROW_STR } from '../utils/narrow';
 import { NULL_ARRAY } from '../nullObjects';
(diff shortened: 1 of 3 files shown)


gnprice:
This looks great, thanks!

I don't much mind losing those tests. This logic is, as you say, quite simple once you get down to it (and especially after the bug is removed :stuck_out_tongue: ); and the tests weren't doing us much good.

I also don't mind at all any loss of reusability here. If we want to say `!flags.read[msg.id]` elsewhere, I think we can just say that. If we want to say `messages.find(msg => !flags.read[msg.id])` elsewhere in the future... that's maybe enough to be worth having in a function. But we can always make a function again later.

It'd be nice to have some tests for `getFirstUnreadIdInNarrow`. Though actually, the more interesting thing would be tests for `getShownMessagesForNarrow`, which we don't seem to have. I don't think that blocks merging this.

chrisbobbe:
OK, I've removed the "[don't merge]" on those commits! 🙂

ray-kraesig:
Merged, with one typo-fix in the first commit's message ("seem" → "seems").

PR #3969, opened 2020-03-12
prettier: use Babylon (babel-flow) as the parser

ray-kraesig:
The `babel-flow` and `flow` parsers produce subtly different ASTs, and Prettier's support for the latter is relatively poor.

In particular, when using the `flow` parser:
  * ~~prettier/prettier#&zwj;5288: the private field sigil `#` is stripped.~~ *(Resolved.)*
  * prettier/prettier#2597: type-declaration comments are promoted to inline type declarations.

This commit will enable some simple uses of private fields, but not all. See our commit 86a42582aace1b13f1e250b0a34835866b794e4a and its associated PR (#3725) for more detail.

The more important fix (and the one that can't be obtained by updating Prettier, at least not yet) is the second. This will allow us to rely on Flow's typechecking in scripts which will be executed directly by Node without a preliminary Babel run -- e.g., scripts residing in `tools/`, or plugins for support tools such as Jest or ESLint.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,6 @@
 {
   "printWidth": 100,
-  "parser": "flow",
+  "parser": "babel-flow",
   "singleQuote": true,
   "trailingComma": "all",
 }


ray-kraesig:
(This PR was motivated by the fact that I've needed to make this change for two separate scripts currently in work – the translation automation script, and a better-for-our-purposes import-rules ESLint plugin.)

gnprice:
Thanks @ray-kraesig ! Merged. Glad to enable that tools use-case.

I tweaked the commit message slightly so "prettier issue NNNN" becomes "prettier/prettier#NNNN", just to save the reader having to look up the exact location of Prettier's issue tracker.

ray-kraesig:
> Thanks @ray-kraesig ! Merged. Glad to enable that tools use-case.
> 
> I tweaked the commit message slightly so "prettier issue NNNN" becomes "prettier/prettier#NNNN", just to save the reader having to look up the exact location of Prettier's issue tracker.

That was thoughtful and intentional on my part; I'd wanted to avoid spamming the Prettier bug-tracker with additional backtrack-notices. :disappointed:

gnprice:
> I'd wanted to avoid spamming the Prettier bug-tracker with additional backtrack-notices.

I see. Looking at those two particular Prettier issue threads now, they look fine to me. For the one that's still open, I think the trackback link is actually helpful -- it provides some signal "hey, people are running into this bug".

Certainly there can be cases where it's annoying, though. When you have the goal of defeating that GitHub feature, I'd still appreciate saving the reader from having to look up the exact location of some other project's tracker. One way to do so would be to write something like `https:// github.com/prettier/prettier/issues/12345`, with a mangled but readable link. Another would be like "prettier/prettier issue 12345".

Issue #3881, opened 2020-02-07
Sending outbox messages is fraught with issues


ray-kraesig:
There is a small constellation of intertwined problems with message-sending.

* All outbox messages are sent immediately, with no throttling.
* Outbox messages may be sent in any order, regardless of their input order.
* Server-rejected messages are not marked as such and will be repeatedly re-sent, wasting bandwidth and battery.
* Outbox messages may be _successfully_ sent multiple times.
* If the app is interrupted and killed while trying to send messages, it may become *permanently* stuck, and will never send messages again unless it is uninstalled or its data is wiped.
* (_... etc.?_)

gnprice:
> * If the app is interrupted and killed while trying to send messages, it may become permanently stuck, and will never send messages again unless it is uninstalled or its data is wiped.

Huh fascinating! How does this one happen?


> * Outbox messages may be _successfully_ sent multiple times.

Definitely possible in principle: I think the scenario would be that a send request succeeds on the server, but the response doesn't make it to us; and the new-message event (which we make into `EVENT_NEW_MESSAGE`) doesn't reach us either before we retry.

Is that the kind of scenario you have in mind, or is there another path?


All of these would be good to address. I imagine a single solution may address several of them at once.

gnprice:
See also #3829, #3731, #3584, #2374. And #3247 is an issue with a similar nature on a different codepath: double-sends of posting an emoji reaction to a message.

gnprice:
OK, and in particular:

> > * Outbox messages may be _successfully_ sent multiple times.
> 
> Definitely possible in principle: I think the scenario would be that a send request succeeds on the server, but the response doesn't make it to us; and the new-message event (which we make into `EVENT_NEW_MESSAGE`) doesn't reach us either before we retry.

This one is #2374. There's some nice discussion on that thread, too.


> * Server-rejected messages are not marked as such and will be repeatedly re-sent, wasting bandwidth and battery.

I think this is the same issue as #3731, describing another aspect of the symptoms. (Which would be a nice point to add to that thread.)

ray-kraesig:
> > * If the app is interrupted and killed while trying to send messages, it may become permanently stuck, and will never send messages again unless it is uninstalled or its data is wiped.
> 
> Huh fascinating! How does this one happen?

`outboxSending` is stored in Redux. If the app is dehydrated with that set to `true`, then killed, there is no mechanism that will clear it.

[...]

PR #3970, opened 2020-03-12
Block imports of files via a whitelist

rk-for-zulip:
Add and use a custom ESLint plugin to prevent certain files from being imported unless they're on a whitelist.

This also sets up a minimal framework for other custom ESLint rules.

### DIFF
--- a/.eslintplugin/zulip-disallow-imports.js
+++ b/.eslintplugin/zulip-disallow-imports.js
@@ -1,0 +1,379 @@
+/* @flow strict-local */
+
+// This is an almost complete rewrite of ESLint's `no-restricted-imports` rule.
+//
+// At present, this is not a full replacement for all uses of that rule:
+// `no-restricted-imports` allows blacklisting patterns of identifiers as well.
+// This has not been reimplemented, but the facilities used by the original to
+// do so have been left in place, in expectation of future requirements.
+//
+// The original is available at:
+//
+//   github.com/eslint/eslint/blob/0243549db4/lib/rules/no-restricted-imports.js
+
+'use strict';
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+// $FlowFixMe: `ignore` has no flow-typed typedef file
+const ignore = require('ignore');
+
+// from node.js; Flow provides these types in core
+const path = require('path');
+
+/**
+ * Trivial wrapper around `ignore` library. Partly this is for type-checking,
+ * but mostly it's because "ignores" is a terrible way to spell "matches".
+ */
+class Matcher {
+  /*:: _ignore; */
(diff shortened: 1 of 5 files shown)


ray-kraesig:
So this is me coming back from the zoo with an armful of yak hair. 

Ultimately, this is in support of #3881. From the top (bottom?):

* #3881 involves modifying `outboxReducer`. New tests are needed.
* `outboxReducer-tests` isn't properly typed yet. In fact there are bugs in the test code because of that.
* Writing well-typed testing code for a reducer is an absolutely *miserable* experience if you don't have access to the types of the actions it's reducing. (Particularly server-originating actions.)
* Just adding `export` to the action types would result in those types leaking into places where `Action` should be opaque. (See, e.g., commit 6b703b6dd28f2c09f0d65db3facf86f7f9166e3c.)
* Moving them to a new conditionally-import-restricted file is possible, but makes the `overrides` property in `.eslintrc` much more complicated to write and test.
  * There are two available ESLint rules that look like they *should* be able to do this, but neither one actually can. (`import/no-restricted-paths` is particularly confusingly documented.)

And at that point I'd spent enough time digging around in the guts of ESLint that I figured it would be easy enough to adapt one of those rules into a custom rule, and, well, here we are.

This isn't even the end of the subtask chain – as that first commit indicates, that's actually #3969 (which is also needed for, or at least badly wanted by, the translation automation tool).

ray-kraesig:
Updates (if you've looked at this before):
* Moved the free-floating `arrayOfStrings` inherited from `no-restricted-imports` into the JSON schema as a definition.
* Removed `importerBlacklist`, since synthesizing it from `importerWhitelist` is relatively easy.

.eslintplugin/zulip-disallow-imports.js
```diff
@@ -0,0 +1,379 @@
...
+/**
+ * Basic set of error messages.
+ *
+ * These error messages also have automatically-generated versions with a
+ * trailing `{{customMessage}}`; see `messages`, below.
+ */
+const baseMessages = {
+  fileImport: "'{{importee}}' may not be imported.",
+  fileWhitelist: "'{{importer}}' may not import '{{importee}}' (not whitelisted).",
+  fileBlacklist: "'{{importer}}' may not import '{{importee}}' (blacklisted).",
```
ray-kraesig: (Note to self: remove this line.)

gnprice:
This would be a useful lint feature, but not important enough to match the complexity cost of the implementation, when that's amortized only over our small codebase.

I think it would make a lot of sense for this feature to exist in ESLint upstream, or in an ESLint plugin advertised to the broader JS community. Also for any one codebase with maybe a few hundred people working in it, I think it'd be worthwhile for someone working on code quality there to build something like this.

PR #3971, opened 2020-03-13
eslint [nfc]: Give comments more room to breathe; regularize indentation.

gnprice:
Things I noticed in the course of making the edits in 88cafe4df, that can perhaps make future edits to this config a little easier.

This is `eslint [nfc]` because it makes no functional changes even *to* the ESLint config; it just touches comments and whitespace.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -112,8 +112,6 @@ rules:
 
   camelcase: off
 
-  no-await-in-loop: off # Promise.all is not always desirable
-
   # Permit dangling underscores in class property names, to denote "private"
   # fields. (These should be replaced with true private fields per the TC39
   # class fields proposal [1], once that's available to us.)
@@ -139,6 +137,7 @@ rules:
   # Repeal some rather absurd rules that make some code impossible to write in
   # the most reasonable way.
   no-else-return: off
+  no-await-in-loop: off # Promise.all is not always desirable
   prefer-destructuring:
   - error
   - AssignmentExpression: {array: false, object: false}
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -28,10 +28,12 @@ settings:
     node:
       extensions: [.js, .ios.js, .android.js, .json]
 
+
 rules:
 
   #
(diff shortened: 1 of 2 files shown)


gnprice:
@ray-kraesig , one other thought that came to mind looking at this: what's the reason why the overrides on `files: ["*.js"]` are there rather than in the main `rules` block? A comment says they need to be, but I don't recall if you mentioned at the time what went wrong if they're normal rules, and the commit 245ea9c4cdf8fb9459ba64dff3e71dab89b36c75 and PR #3725 don't say.

.eslintrc.yaml
```diff
@@ -76,47 +91,74 @@ rules:
...
   - patterns: ['**/__tests__/**']
+
   import/no-cycle: off  # This would be nice to fix; but isn't easy.
 
-  # Jest
+
+  #
+  # Jest; plugin `jest`.
+  #
+
```
ray-kraesig: This doesn't really look like it's supposed to be a sort of section header.

I'd suggest something like

```
# ===================================================================
# Jest

# This rule could be useful if...
```

or, with somewhat more brevity,

```
####
# Jest

# This rule could be useful if...
```

or even the Markdown-style

```
# ## Jest
#
# This rule could be useful if...
```

... although personally I'd include the `#` in all cases, not just the final one.

.eslintrc.yaml
```diff
@@ -76,47 +91,74 @@ rules:
...
-  # React Native.
+
+  #
+  # React Native; plugin `react-native`.
   #
+
   # This plugin isn't included in the airbnb config, and it doesn't
   # itself enable any rules by default, so this config is it.
-  #
+
```
ray-kraesig: ?

gnprice:
I've just pushed a refreshed version of this: rebased across the changes since it was originally written, addressing Ray's comment above at https://github.com/zulip/zulip-mobile/pull/3971#discussion_r393253375 , and adding a few new commits.

chrisbobbe:
Looks great, thanks! Merged.

Issue #3961, opened 2020-03-10
In "add reaction" search, accept actual emoji as well as names


gnprice:
A user [reports](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Reactions/near/829192):
> When you click "add reaction" you get the emoji drop down list and the search field.
>
> On the search field, if you add the smile emoji from the Android keyboard (not the "smile" word) you can't add it as a reaction. The only available action is to search, but it can't find the emoji on the drop down list.

This is a perfectly reasonable thing to want to do -- system emoji keyboards can be pretty great, and many people are accustomed to them and they have the user's frequently-used emoji. So we should make it work.

To do this, I think what we need to do is basically to search through the actual Unicode values of Unicode emoji, as well as the names of the emoji.

7lym:
Hello, this would be a useful alternative to revamping the emoji picker menu (by using groups for example, just like system emoji keyboards do). 
Is this feature still part of the release roadmap? Thank you.

PR #3972, opened 2020-03-13
emoji search: Show result for literal emoji search.

agrawal-d:
Works for both single-codepoint (:+1:) and multi-codepoint emoji (:hash:).

image

Fixes https://github.com/zulip/zulip-mobile/issues/3961

### DIFF
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -57,6 +57,20 @@ describe('getFilteredEmojiNames', () => {
     expect(getFilteredEmojiNames('qwerty', { qwerty: {} })).toEqual(['qwerty']);
   });
 
+  test('returns the emoji name for a literal single-codepoint emoji search', () => {
+    const list = getFilteredEmojiNames('🤔', {});
+    expect(list).toEqual([
+      'thinking',
+    ]);
+  });
+
+  test('returns the emoji name for a literal multi-codepoint emoji search', () => {
+    const list = getFilteredEmojiNames('#️⃣', {});
+    expect(list).toEqual([
+      'hash',
+    ]);
+  });
+
   test('remove duplicates', () => {
     expect(getFilteredEmojiNames('dog', {})).toEqual(['dog', 'dogi']);
     expect(getFilteredEmojiNames('dog', { dog: {} })).toEqual(['dog', 'dogi']);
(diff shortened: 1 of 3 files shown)


src/emoji/data.js
```diff
@@ -29,6 +29,32 @@ export const getFilteredEmojiNames = (
   query: string,
   activeImageEmojiByName: $ReadOnly<{ [string]: ImageEmojiType }>,
 ): string[] => {
+  // Check if the query itself is an emoji, and is known to Zulip. If
+  // so, return the emoji name in an array. Emoji with single-codepoint and
+  // multi-codepoint have length 2 and 3 respectively.
```
chrisbobbe, line 34: "Emoji with single-codepoint and multi-codepoint have length 2 and 3 respectively."

Hmm, that's surprising to me! I wonder why? Maybe add a link to some documentation in this code comment?

src/emoji/data.js
```diff
@@ -29,6 +29,32 @@ export const getFilteredEmojiNames = (
...
+    );
+    if (emojiName !== undefined) {
+      return [emojiName];
+    }
+  } else if (query.length === 3) {
+    const emojiName = unicodeEmojiNames.find(
+      key =>
+        unicodeCodeByName[key]
+        === `${query
+          .codePointAt(0)
```
chrisbobbe, line 47: This indentation looks confusing to me; maybe Prettier and ESLint are making it look like this. One thing that might help is to give it some breathing room by using the curly-brace and `return` statement syntax with the lambda, e.g.,

```javascript
key => {
  // ...
  return /* ... */;
}
```

Then, in there, you could have separate variables corresponding to the first and second code points, just so it's a bit clearer what's going on.

src/emoji/__tests__/data-test.js
```diff
@@ -57,6 +57,20 @@ describe('getFilteredEmojiNames', () => {
...
   });
 
+  test('returns the emoji name for a literal single-codepoint emoji search', () => {
+    const list = getFilteredEmojiNames('🤔', {});
+    expect(list).toEqual([
+      'thinking',
+    ]);
+  });
+
+  test('returns the emoji name for a literal multi-codepoint emoji search', () => {
```
chrisbobbe, line 67: Since we have checks on the length of `query`, if it's 2 or 3, I think it would be smart to test some ordinary, non-emoji queries with those lengths, e.g., 'sad', just to make sure their behavior is the same.

src/emoji/data.js
agrawal-d: Even I am wrong. Emoji length can vary a lot depending on modifiers :sweat: 
https://hsivonen.fi/string-length/

agrawal-d:
Thanks, @chrisbobbe, for the review. 
Overall, this looks like a very hacky code to me, now that I've read a bit more about emoji. I'll close this PR for now.

PR #3973, opened 2020-03-14
presence: Fix some unlikely cases; add another test; draw state machine.

gnprice:
I reread this code closely today when thinking about [this bug report](https://chat.zulip.org/#narrow/stream/9-issues/topic/messages.20before.20status.20change/near/831348).

I didn't find anything that seemed like it could have caused it... a good outcome, because it turned out that came from the previous release, when this code looked quite different. :slightly_smiling_face: I did think for a while that I'd found something that possibly could have, because the way we avoid a certain problem is kind of subtle. Add a comment explaining the subtlety (complete with a state-machine diagram!), and a test for that case.

Also make a couple of changes that are mostly NFC in practice, but would improve the behavior in edge cases (or a hypothetical implementation change in RN), and that make the code a bit simpler to reason about.

### DIFF
--- a/src/presence/PresenceHeartbeat.js
+++ b/src/presence/PresenceHeartbeat.js
@@ -34,8 +34,8 @@ class PresenceHeartbeat extends PureComponent<Props> {
   heartbeat: Heartbeat = new Heartbeat(this.onHeartbeat, 1000 * 60);
 
   componentDidMount() {
-    this.updateHeartbeatState(); // conditional start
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks @ray-kraesig for the audio review yesterday! I've just pushed a revision to adjust the state diagram so the arrows pointing into "on" are less crowded, as you requested. LMK if there are any other revisions you'd like to see.

ray-kraesig:
Thanks; looks good! Merged.

PR #3974, opened 2020-03-14
group-pm: Improve the text color on group icons

Maskedman99:
Determine the color of text on group avatar by using foregroundColorFromBackground() this will help users to read the text properly when the background color is light.
Screenshot_20200314-130555_Zulip (debug)

Also replace the for loop in colorHashFromString() with a single line equation that will reduce the complexity from O(n) to O(1).
The for loop before resulted in a recursive equation: 
```hash(N) = hash(N-1) * 31 + name.charCodeAt(1)       , hash(0) = 0```
This actually is the sum of the first N terms in a Geometric Progression, for which a mathematical equation is already present.
The for loop ran till ```names.length```, whose value depends on the no.of members in a group. For a certain large value and above, JS wouldn't be able to calculate the value of ```hash``` and returns ```infinity```,  this happens in the equation too, hence the value ```names.length % 100``` was used instead of ```names.length```.

### DIFF
--- a/src/common/GroupAvatar.js
+++ b/src/common/GroupAvatar.js
@@ -4,7 +4,7 @@ import { Text, StyleSheet, View } from 'react-native';
 
 import type { Node as React$Node } from 'react';
 import Touchable from './Touchable';
-import { colorHashFromString } from '../utils/color';
+import { colorHashFromString, foregroundColorFromBackground } from '../utils/color';
 import { initialsFromString } from '../utils/misc';
 
 const styles = StyleSheet.create({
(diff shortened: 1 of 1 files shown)


ray-kraesig:
The use of `foregroundColorFromBackground` is a good idea, and will be accepted if resubmitted on its own.

However, `colorHashFromString` is *completely* broken, in multiple ways; optimizing a poorly-chosen algorithm only moves it farther away from something useful. Please omit this part. (Thanks for calling our attention to it, though.)

Maskedman99:
@ray-kraesig I've done the suggested modifications.

ray-kraesig:
Merged (after some minor proofreading in the commit message). Thanks!

PR #3978, opened 2020-03-17
[WIP?] Add cyclomatic complexity lint (+ various refactors)

ray-kraesig:
Refactor a few large functions, and add a lint to prevent future functions from getting too unwieldy. (In particular, this refactors the already-problematic `constructMessageActionButtons` to split the Outbox and Message cases, fixing a bug.)

The cyclomatic complexity lint has a few arguable false positives and a few items that would just be nontrivial to do. These have just been marked with exhortative `TODO` markers for now.

This PR should not be merged as-is; some decisions need to be made first.

 - [ ] Should `fromAPNsImpl` actually be broken up?
   * The right way to remove it from lint-consideration is probably to make a generalized typechecked validator (via [ajv](https://github.com/epoberezkin/ajv) or the like). That's going to involve some code generation somewhere, though, if we want to keep it Flow-checked.
   * Breaking it up doesn't actually make it clearer or more testable; it might be better just to tag it with a marker.
-  [ ] What exactly should the limit be?
   * ESLint defaults to 20. That's high enough that nothing we have hits it (at least after the refactors), but I'm of the opinion that this is far too high.
   * 15 (my preference) is doable, but requires either tagging a few functions with `// eslint-disable`s or doing some less trivial refactoring.
   * [The original study suggested 10.](https://en.wikipedia.org/wiki/Cyclomatic_complexity#Limiting_complexity_during_development) This is technically reachable, but would either take a lot of questionably-productive effort or would require throwing around sufficiently many `// eslint-disable`s that, I think, we run the risk of becoming cavalier with them.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -65,7 +65,7 @@ rules:
   no-plusplus: off
   no-nested-ternary: off
 
-  complexity: ["error", {max: 20}]
+  complexity: ["error", {max: 15}]
 
   # We also partly repeal AirBnB's forbidden-syntax subrules. (This must be done
   # in the `overrides:` section.)
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -65,6 +65,8 @@ rules:
   no-plusplus: off
   no-nested-ternary: off
 
+  complexity: ["error", {max: 20}]
+
   # We also partly repeal AirBnB's forbidden-syntax subrules. (This must be done
   # in the `overrides:` section.)
   #
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -33,31 +33,31 @@ import {
 } from '../actionConstants';
 import { getOwnEmail } from '../users/userSelectors';
 
-const opToActionUser = {
+const opToActionUser = Object.freeze({
   add: EVENT_USER_ADD,
   remove: EVENT_USER_REMOVE,
   update: EVENT_USER_UPDATE,
-};
+});
 
-const opToActionUserGroup = {
+const opToActionUserGroup = Object.freeze({
   add: EVENT_USER_GROUP_ADD,
   remove: EVENT_USER_GROUP_REMOVE,
   update: EVENT_USER_GROUP_UPDATE,
   add_members: EVENT_USER_GROUP_ADD_MEMBERS,
   remove_members: EVENT_USER_GROUP_REMOVE_MEMBERS,
-};
+});
 
-const opToActionReaction = {
+const opToActionReaction = Object.freeze({
   add: EVENT_REACTION_ADD,
   remove: EVENT_REACTION_REMOVE,
-};
+});
 
-const opToActionTyping = {
+const opToActionTyping = Object.freeze({
   start: EVENT_TYPING_START,
   stop: EVENT_TYPING_STOP,
-};
+});
 
-const actionTypeOfEventType = {
+const actionTypeOfEventType = Object.freeze({
   update_message: EVENT_UPDATE_MESSAGE,
   subscription: EVENT_SUBSCRIPTION,
   presence: EVENT_PRESENCE,
@@ -68,7 +68,7 @@ const actionTypeOfEventType = {
   update_global_notifications: EVENT_UPDATE_GLOBAL_NOTIFICATIONS_SETTINGS,
   update_display_settings: EVENT_UPDATE_DISPLAY_SETTINGS,
   user_status: EVENT_USER_STATUS_UPDATE,
-};
+});
 
 // This FlowFixMe is because this function encodes a large number of
 // assumptions about the events the server sends, and doesn't check them.
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -73,6 +73,14 @@ const actionTypeOfEventType = Object.freeze({
 // This FlowFixMe is because this function encodes a large number of
 // assumptions about the events the server sends, and doesn't check them.
 export default (state: GlobalState, event: $FlowFixMe): EventAction => {
+  // simple case: conversion by lookup-table
+  const actionType: $Values<typeof actionTypeOfEventType> | void =
+    actionTypeOfEventType[event.type];
+  if (actionType !== undefined) {
+    return { ...event, type: actionType };
+  }
+
+  // complicated case: additional event translation needed
   switch (event.type) {
     case 'alert_words':
       return {
@@ -98,21 +106,6 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction => {
       return {
         type: EVENT,
         event,
-      };
-
-    case 'update_message':
-    case 'subscription':
-    case 'presence':
-    case 'muted_topics':
-    case 'realm_emoji':
-    case 'realm_filters':
-    case 'submessage':
-    case 'update_global_notifications':
-    case 'update_display_settings':
-    case 'user_status':
-      return {
-        ...event,
-        type: actionTypeOfEventType[event.type],
       };
 
     case 'realm_user':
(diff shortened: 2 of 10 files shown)


ray-kraesig:
(No functional changes since last push; just some commit message edits.)

Note that several of the refactors included here are probably worth including in their own right, regardless of whether or when the cyclomatic complexity lint is adopted. The patchset has been arranged so that they can be individually cherry-picked.

gnprice:
Hmm. I'll say I'm dubious of this kind of metric in general. And the changes here that feel most like they're driven by the metric and the lint don't seem to me like improvements.

Patch by patch:

07f54909d action sheet [nfc]: Refactor constructMessageActionButtons (1/2).
5c26b071e action sheet: Refactor constructMessageActionButtons (2/2).
51765d2f8 action sheet [nfc]: Remove `isAnOutboxMessage`.

I like the cleanup in this subseries!

I'd want to confirm that those are the right set of actions to offer on an outbox message. But knowing the type explicitly definitely seems helpful, and I like the technique of parameterizing `ConstructSheetParams`.

2064228d2 subscriptions [nfc]: Refactor reducer (1/2).
c10e6e92b subscriptions [nfc]: Refactor reducer (2/2).
8078a6529 notifications [nfc]: Break up message-validator function.

These move around code but they don't particularly disentangle anything (unlike the `constructMessageActionButtons` case), and they don't feel to me like they really help things. If anything they add some indirections to follow; and in the case of the last one, 19 additional lines (in net) of boilerplate. I'd be OK making some refactors like this if it helped us apply a lint that really helped a lot elsewhere, but I'm not seeing that in these results.

907a88dd1 events [nfc]: Reduce cyclomatic complexity of eventToAction.

This one makes the code shorter but denser and less homogeneous -- I think on net it becomes less clear. Before this change, to see what we do with a given event type you can just scan down the cases and find its name.


9b1282ebf event actions [nfc]: Freeze lookup objects.

    This allows Flow to recognize these as $ReadOnly, with highly-specific
    key- and value-types.

This sounds handy!

I'm not actually reproducing the effect, though -- I tried removing the `Object.freeze` on `actionTypeOfEventType`, and `yarn flow` passes and also I see the same very specific type when I hover on `const actionType` as I did with the `Object.freeze` present.

gnprice:
Staring at that `constructMessageActionButtons` function some more, I think a major reason it felt so messy (and still feels fairly messy even after the improvement made by this branch) is that it has a long series of conditions each of which (a) is complicated, often two or three different conditions joined by boolean operators, and (b) isn't particularly parallel to the others. That all makes it pretty complex to reason about.

By contrast in our reducers and in `eventToAction` we tend to have `switch` statements with sometimes lots of cases. The conditions are then quite simple and quite parallel to each other -- they're just switching on the value of one tag property, or in some cases a couple of tag properties with nesting semantics (i.e. given a certain value for the top-level tag, you switch on a certain set of values for the next-level tag.) They're also manifestly mutually exclusive, which simplifies things a lot -- only one is relevant at a time.

Maybe there could be a metric getting at that characteristic, and a lint checking for that. I'd still be doubtful of good results in practice until I really saw it, though.

gnprice:
> I like the cleanup in this subseries!
> 
> I'd want to confirm that those are the right set of actions to offer on an outbox message.

OK, looking closer:
* yep, "delete" sure makes sense
* I don't see a reason why the offering of "copy" and "share" should be different for outbox messages vs. real full-blown messages. I don't totally love that set of options as it is (and we have some open issues about changing it), but unless there's a clear reason for variation I think it'll be cleanest to leave those parallel between the two cases -- so, include `shareMessage` in the outbox case.


Oh, and then in the next commit message, though not particularly central:

    (We could alternatively annotate it with `%checks`. In this case,
    however, that would merely upgrade it from "worse than useless" to
    "useless".)
    
From my reading of https://github.com/facebook/flow/issues/34#issuecomment-459595992 and the thread around it -- as so often happens, the best Flow docs are the issue threads :confused: -- I think `%checks` wouldn't actually even work here. Certainly agree inlining this is better.

ray-kraesig:
> [...] But knowing the type explicitly definitely seems helpful, and I like the technique of parameterizing `ConstructSheetParams`.

Amusingly enough, I'm not a huge fan. Parameterizing `ConstructSheetParams` was a mechanical change; I think the resultant code would probably be slightly cleaner if `message` were extracted from it instead.

(But it would also be more involved to execute and more troublesome to verify the correctness of such a change; and the expected improvement is tiny. Maybe if someone's really bored someday.)

I'll amend it with your suggestion about "Share" and submit it as a separate PR.

----

> ... in the case of [8078a6529721a8716e8c45a3fa4c0ef47e2fb335], 19 additional lines (in net) of boilerplate.

As I noted in the commit message, I wasn't happy with that one either.

(Also as I noted there, I think if we had a good JSON schema validator, plus an up-to-date JSON-schema-to-Flow-type-definition translator – or vice versa, I'm not picky – it would cover more than 90% of our input validation needs. 

I spent some time looking into JavaScript JSON validation libraries we could use, but didn't find any I liked enough to recommend. Perhaps after RN 0.60 I'll take some time to look into native-code JSON-schema validators.)

----

> [907a88d](https://github.com/zulip/zulip-mobile/commit/907a88dd135a1a9b7af0d7600e72daaabf242276) events [nfc]: Reduce cyclomatic complexity of eventToAction.
> 
> This one makes the code shorter but denser and less homogeneous -- I think on net it becomes less clear. Before this change, to see what we do with a given event type you can just scan down the cases and find its name.

Can't really argue that.

If I were going to redo this (in the expected continued absence of full validation of events), it would probably start like this...

```typescript
type ZerverEventType = ...;

type ActionCreationMethod =
  // A method to construct the action, in full.
  | ((event: $FlowFixMe, state: GlobalState) => EventAction)
  // A string value `EVENT_TYPE_NAME` is shorthand for the common case
  //   `event => { ...event, type: EVENT_TYPE_NAME }`.
  | string;

const dispatchTable: { +[ZerverEventType]: ActionCreationMethod } = {
  alert_words: event => ({
    type: EVENT_ALERT_WORDS,
    alertWords: event.alertWords,
  }),

  presence: EVENT_PRESENCE,

  ...
};
```

which would actually remove a layer of indirection. But there's not much impetus to do so now.

----

> I'm not actually reproducing the effect, though -- I tried removing the `Object.freeze` on `actionTypeOfEventType`, and `yarn flow` passes and also I see the same very specific type when I hover on `const actionType` as I did with the `Object.freeze` present.

Same here, now. :confused:  I suspect that this was to block one of those cases where Flow cheerfully decides to add a mistake to its inference set, and then produce an error message three functions later and twenty-nine-and-a-half times the size it needs to be.

If I manage to reproduce it, I'll resubmit with more details.

----

> From my reading of [facebook/flow#34 (comment)](https://github.com/facebook/flow/issues/34#issuecomment-459595992) and the thread around it -- as so often happens, the best Flow docs are the issue threads  -- I think `%checks` wouldn't actually even work here.

That's not a hypothetical; it *does* work. Try it on `master` – it'll start complaining about an unneeded suppression comment.

gnprice:
> > I think `%checks` wouldn't actually even work here.
> 
> That's not a hypothetical; it _does_ work. Try it on `master` – it'll start complaining about an unneeded suppression comment.

Good to know!

I also hadn't been aware Flow had that feature before seeing this PR. Worth being on the lookout for places it might be useful for us.



> If I were going to redo this (in the expected continued absence of full validation of events), it would probably start like this...

Plausible!

See also `src/api/eventTypes.js`, and the history in `git log --stat -p --full-diff src/api/eventTypes.js` -- that's a sketch I made early last year of another way to make these more structured.

(But probably the place for any detailed followup is in chat.)


> I'll amend it with your suggestion about "Share" and submit it as a separate PR.

Thanks! #4065, I see.

Issue #3979, opened 2020-03-18
Support UI "alerting a word in a message" feature


sumukhah:
The current app highlights alerted messages/words. But a user cannot add/remove them. The web app has this feature

![issue](https://user-images.githubusercontent.com/23723464/76960102-d0324400-6940-11ea-9704-4af5ff569682.jpg)

sumukhah:
Working on the issue

PR #986, opened 2017-08-05
Created Alert words screen in Setting details screen.

jainkuniya:
screen shot 2017-08-05 at 12 02 44 pm
screen shot 2017-08-05 at 12 03 06 pm
screen shot 2017-08-05 at 12 03 23 pm

PR #3981, opened 2020-03-18
alert word: Add UI to support add, delete alert word

sumukhah:
web app supports adding and deleting a alert words. In mobile app
alert words are only highlighted with a background color. User
cannot add or remove an alert word.

so Create an option for this purpose in setting screen. Support
adding and deleting an alert word.

Fixes: #3979


src/settings/AlertWordScreen.js
```diff
@@ -0,0 +1,99 @@
...
+    const { auth, alertWords } = this.props;
+    const { text } = this.state;
+    try {
+      addAlertWord(auth, [...alertWords, text]);
+    } catch (error) {
+      /* eslint-disable no-console */
+      console.warn('unable to connect server');
+    } finally {
+      this.setState({ text: '' });
+    }
```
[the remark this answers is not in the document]
sumukhah: some thing like this??
ezgif-6-cb375c33311f

sumukhah:
@ray-kraesig I have changed some code in the commit. Please review

src/api/alert_words/removeAlertWord.js
```diff
@@ -0,0 +1,9 @@
+/* @flow strict-local */
+
+import type { ApiResponse, Auth } from '../transportTypes';
+import { apiDelete } from '../apiFetch';
+
+export default (auth: Auth, alertWord: Array<string>): Promise<ApiResponse> =>
```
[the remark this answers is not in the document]
sumukhah: Yes, I thought `{add, remove}Alertword.js` is capable of adding/remove one-word on a single call. The overall result from these functions is the addition/deletion of a single word, whereas `getAlertWords` receive all the words. I thought this would explain very well even though it takes multiple words as an argument during addition/deletion.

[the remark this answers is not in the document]
sumukhah: >The actual filenames should also be plural here ({add,remove}AlertWords.js), since the functions they export accept multiple items.

Yes, `{add, remove}Alertword.js` is capable of adding/remove one-word on a single call. The overall result of these functions is the addition/deletion of a single word. I thought this would explain very well even though it takes multiple words as an argument during addition/deletion.

jainkuniya:
Just to mention, there was similar work done long time back #986. Just in case, it might help in any way.

src/settings/AlertWord.js
```diff
@@ -0,0 +1,48 @@
...
+
+const styles = StyleSheet.create({
+  containerStyle: {
+    backgroundColor: 'white',
+    margin: 5,
+    padding: 5,
+    shadowColor: 'black',
+    shadowOpacity: 0.8,
+    elevation: 6,
+    shadowRadius: 15,
```
ray-kraesig, line 15: I'd suggest formatting this style as is done in [this StackOverflow answer](https://stackoverflow.com/a/51302156), to make it clear to the reader what the purpose of these styles are (and that, *e.g.*, changing the `shadowColor` isn't going to have any effect on Android).

... well, assuming this is the UI style we decide to go with. I think it's largely OK as is (possibly modulo some spacing tweaks), but I'm going to call @gnprice in for that.

src/settings/AlertWordScreen.js
```diff
@@ -0,0 +1,108 @@
...
+    } catch (error) {
+      logging.error(error);
+      showToast(error.message);
+    }
+    this.setState({ progress: false });
+  };
+
+  handleRemoveAlertWord = async (word: string) => {
+    const { auth, alertWords } = this.props;
+    const filteredWords = alertWords.filter(alertWord => alertWord === word);
```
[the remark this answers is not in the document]
sumukhah: > What's the purpose of this filter step?
> Why does this call include ...alertWords?

Yes, they do not require those steps in fact. That was my mistake :(

sumukhah:
Fixed sathe issues pls review@ray-kraesig

src/api/alert_words/addAlertWords.js
```diff
@@ -0,0 +1,7 @@
+/* @flow strict-local */
+import type { ApiResponse, Auth } from '../transportTypes';
+import type { AlertWordsState } from '../../types';
+import { apiPost } from '../apiFetch';
+
+export default (auth: Auth, alertWords: AlertWordsState): Promise<ApiResponse> =>
```
ray-kraesig, line 6: `AlertWordsState` describes our Redux state, which is an internal implementation detail of the app. If we change how we're storing that state in Redux, this API call should *not* change. (And conversely, if the API were ever to change (though it shouldn't), we also wouldn't necessarily want to change our internal state to be the same type.)

This really should be exactly `string[]`.

src/settings/AlertWordScreen.js
```diff
@@ -0,0 +1,123 @@
...
+    const { alertWords, auth } = this.props;
+
+    if (alertWords.includes(text.trim())) {
+      this.setState({ errorMessage: 'This word is already present' });
+    } else {
+      this.setState({ progress: true });
+      try {
+        await addAlertWords(auth, [text]);
+        this.setState({ text: '' });
+        showToast('Created a new alert word');
```
ray-kraesig, line 67: This toast only serves to give the user [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue). The alert word appears in the list; there shouldn't be a need for any further notifications.

src/settings/AlertWordScreen.js
```diff
@@ -0,0 +1,123 @@
...
+      text,
+    });
+  };
+
+  handleSubmit = async () => {
+    const { text, errorMessage } = this.state;
+    const { alertWords, auth } = this.props;
+
+    if (alertWords.includes(text.trim())) {
+      this.setState({ errorMessage: 'This word is already present' });
```
ray-kraesig, line 61: Consider an early return here to reduce nesting depth.

Also, it's confusing to report some errors with persistent error message text, and others with a toast.

src/settings/AlertWordScreen.js
```diff
@@ -0,0 +1,123 @@
...
+  state = {
+    text: '',
+    progress: false,
+    errorMessage: '',
+  };
+
+  handleTextChange = (text: string) => {
+    this.setState({
+      text,
+    });
```
ray-kraesig, line 53: Editing the text should probably clear the persistent error-message, if that's kept.

src/api/alert_words/getAlertWords.js
```diff
@@ -1,10 +1,11 @@
 /* @flow strict-local */
 import type { Auth, ApiResponseSuccess } from '../transportTypes';
 import { apiGet } from '../apiFetch';
+import type { AlertWordsState } from '../../types';
 
 type ApiResponseAlertWords = {|
   ...ApiResponseSuccess,
-  alert_words: string[],
+  alert_words: AlertWordsState,
```
ray-kraesig, line 8: As with the previous occurrence of this: `AlertWordsState` is internal; this type is whatever it's specified to be in the Zulip API documentation. In this case that's exactly `string[]`.

PR #3984, opened 2020-03-19
Added email address to each stream settings view

### LINKED ISSUE #3887: Show a stream's email address
The mobile app should show a stream's email address, used for sending an email to a stream. Currently, a mobile app user who wants to use the incoming email gateway has to log on to the webapp to retrieve the address, or ask someone to share the address.

An obvious place to include this information is in the stream settings view; possibly it could be in a collapsed section or an "advanced" view with other things that have not been added yet.

drigio:
Fixes #3887 
- Stream Settings now shows email address of the stream.
- Clicking on the email address, copies it to clipboard.

I've used `ToastAndroid` for showing that email address is copied. I would love to know an alternative choice we can use to show that status on *iOS* and *Android* as welll without using the `ToastAndroid` module.
Please let me know of any other changes that are needed.


src/streams/StreamScreen.js
```diff
@@ -1,12 +1,12 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
-import { View } from 'react-native';
+import { View, TouchableNativeFeedback, Platform, Clipboard, ToastAndroid } from 'react-native';
```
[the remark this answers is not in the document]
chrisbobbe: I like React Native's naming of `ToastAndroid` better, since it's very clear that it only works on Android, and I see that you're looking for something that works on both platforms. You'll find it in src/utils/info.js, called `showToast`.

Take a look at some other places where `showToast` is used; you'll find that some (it should be all!) pass `_('some string literal')` instead of `'some string literal'`. The `_` ensures that the string displays in the user's chosen language. You'll need access to `_`, and I think the right way to get that is to put this at the top of the `StreamScreen` class:

```
  static contextType = TranslationContext;
  context: GetText;
```

This makes StreamScreen a consumer of the `TranslationContext`, which was created with `React.createContext` in the [new React Context API](https://reactjs.org/docs/context.html#legacy-api). It might be helpful to follow the example of the `LanguagePicker` component (src/settings/LanguagePicker.js), which does the same thing. The strategy is also described in the JSDoc on the `GetText` type in src/types.js.

Then, in `handleCopy`, declare `const _ = this.context;`. The naming of `_` is to make it as unobtrusive as possible, while still ensuring that it shows in its translated form.

Then, you should add an entry into static/translations/messages_en.json. That way, the English version will be available, and when we sync to Transifex, it will be available for translators to translate into other languages (there's a [doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/translations.md) about that, if you're interested). But you don't have to worry about that; the essential thing for this PR branch is that an entry gets added into messages_en.json.

Also, I'd like to be consistent with other user-facing strings, with capitalization and punctuation; what do you think of "Copied email address" (and maybe @gnprice, do you have thoughts on this)?

src/streams/StreamScreen.js
```diff
@@ -104,6 +109,23 @@ class StreamScreen extends PureComponent<Props> {
...
+          <View style={styles.padding}>
+            <Label
+              style={[styles.marginTop, { fontSize: 16, }]}
+              text="Email Address"
+            />
+          </View>
+          <TouchableNativeFeedback
+            onPress={() => this.handleCopy(subscription.email_address)}
+            background={Platform.OS === 'android' ? TouchableNativeFeedback.SelectableBackground() : ''}
+          >
```
chrisbobbe: Taking a quick look at how the web app handles this, it looks like it's also done in the stream settings view, so I think it makes sense to put it here. One thing I noticed is that the web app has a question mark icon with a helpful link to a page ([this one](https://chat.zulip.org/help/message-a-stream-by-email)) that describes what the email address is for:

image

In the mobile app, currently, I don't know of a similar pressable question mark icon with a link to the realm's Help page, but it would definitely be useful here, and potentially elsewhere; @gnprice, what do you think? That Help page is realm-specific, so it would need to be connected to the `accounts` state in Redux to get the current realm and point to the specific help page for that realm. Something reusable would be better than a one-off solution, I think.

src/streams/StreamScreen.js
```diff
@@ -63,10 +63,10 @@ class StreamScreen extends PureComponent<Props> {
     dispatch(toggleStreamNotification(stream.stream_id, !currentValue));
```
chrisbobbe: Please squash this commit with its parent, to maintain the invariant that all commits pass all tests.

I was really excited when I first learned about the tricks for tidying commits on a PR, from [this doc](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html). That doc is really the key piece for enabling the standard of [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html); without it, I'd be stuck, feeling like I just had one shot at making good commits. It's tricky at first, but with practice, it becomes second-nature. 🙂

src/streams/StreamScreen.js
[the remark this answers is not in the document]
drigio: Thanks for pointing me out to the right places! Will definitely take care of the `TouchableNativeFeedback` and the `AndroidToast`.  Thanks on the great explanation of the translation workflow, (was wondering why `_` was being used like that in codebase while quickly skimming through it.)
About the "Copied email address" - I agree, being consistent with other strings is better and maybe let's hear about it from @gnprice as well?

src/streams/StreamScreen.js
drigio: I think having a reusable component will be good! I've seen it at many places like [this](https://chat.zulip.org/help/manage-inactive-streams) and [this](https://chat.zulip.org/help/create-your-organization-profile) as well (all those are redirects to the respective help page). We need to create another issue maybe for that component then?

About the help page being realm specific - isn't that same for all the realms?

src/streams/StreamScreen.js
drigio: Sure will do it. Haha squashing commits definitely helps tidy the huge commit hell xD

drigio:
@chrisbobbe I have made the necessary changes. Please have a look at it. Also, do you need me to squash the last commit as well?

static/translations/messages_en.json
```diff
@@ -199,5 +199,6 @@
   "Reactions": "Reactions",
   "No reactions": "No reactions",
   "See who reacted": "See who reacted",
-  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream."
+  "Only organization admins are allowed to post to this stream.": "Only organization admins are allowed to post to this stream.",
+  "Copied email address": "Copied email address"
```
agrawal-d: A better string might be "Stream email copied to clipboard" or simply "Copied to clipboard".

src/streams/StreamScreen.js
```diff
@@ -104,6 +115,17 @@ class StreamScreen extends PureComponent<Props> {
             secondary
             onPress={() => delay(this.handleEditSubscribers)}
           />
+          <View style={styles.padding}>
+            <Label style={[styles.marginTop, { fontSize: 16 }]} text="Email Address" />
+          </View>
+          <Touchable onPress={() => this.handleCopy(subscription.email_address)}>
+            <Label
+              style={[
+                { borderColor: BRAND_COLOR, borderWidth: 1.5, borderRadius: 20, padding: 16 },
```
agrawal-d: The padding and the border-radius seem too much to me. 5px for borderRadius and 8px for the padding can be better.

agrawal-d: Also, the styles should be applied to `Touchable`, not to `Label`, otherwise, the ripple feedback goes beyond the round borders. I checked, it does get fixed if applied to `Touchable`.

drigio: @agrawal-d I applied the styles to `Touchable` instead of `Label`, however as you can see in the screenshot below the ripple still goes beyond the round borders. Did you check on Android or iOS device?


Also the borderRadius and padding that you suggested seem a bit off from other designs of zulip maybe? What do you say?

static/translations/messages_en.json
drigio: > Unfortunately, this change introduces a bug that causes the app to crash when you view the stream info for an unsubscribed stream. I think this is because stream email is unavailable for such streams, but you should investigate that further.

Yes the bug was indeed caused because of retrieving email address of an unsubscribed streams. Now I've put a check on it and it renders only if the stream is subscribed.

Now there are two options here 
1. We can show a message to user maybe in red, stating that he needs to subscribe to the stream to see the email address (like the image below) OR 
2. We won't render the email address view if stream is not subscribed.

ray-kraesig: Definitely not red. That usually signals an actual error; it's not appropriate for something that's part of normal operation, and that many users won't need or want to change. Semantically, this seems to call for some flavor of grey.

Option 2 seems like it'll confuse users who *are* looking for the email address. (Or even users that aren't, and who are just exploring the stream-settings cards.)

ray-kraesig:
I'm wondering if perhaps this shouldn't be pulled out into its own card, accessible here via an `OptionButton`. The stream's email address seems like something few mobile users will ever want to reference; it probably doesn't need more than a single option's worth of space on the Settings card, and certainly not an eye-catching mostly-unique widget.

Moving it to its own card would permit an unambiguous "Copy to Clipboard" action button, instead of relying on an implicit touch-to-copy. We could also have a one-line help description and an outbound "Help" link.

static/translations/messages_en.json
drigio: @ray-kraesig Sure will work on making it grey.

drigio:
@ray-kraesig If I am not wrong what you are suggesting is add a `OptionButton` here and route it to its own `StreamEmailAddressScreen` which has its own copy `ZulipButton` and a `Label` describing it. What shall we be doing for the "Help" link? 
> Something reusable would be better than a one-off solution, I think.

Create a separate component for it as @chrisbobbe suggested?

ray-kraesig:
> @ray-kraesig If I am not wrong what you are suggesting is add a `OptionButton` here and route it to its own `StreamEmailAddressScreen` which has its own copy `ZulipButton` and a `Label` describing it. What shall we be doing for the "Help" link?

Probably just another `ZulipButton`, labeled "More info..." or the like.

A reusable custom widget sounds like it might be a good idea for other areas of the app, where we can't reasonably fit intrusively-sized help widgets everywhere; if we already had one, I'd say to use it here, too, for consistency's sake. If it's on a mostly-empty card, though, we don't need to worry too much about how much space a Help link takes up.

Issue #3964, opened 2020-03-10
Support "Sign In with Apple" auth [iOS]


gnprice:
This is for the client side, on iOS, of zulip/zulip#14168.

In addition to the links there, here are Apple docs specific to iOS:
* https://developer.apple.com/documentation/authenticationservices
* sample code: https://developer.apple.com/documentation/authenticationservices/implementing_user_authentication_with_sign_in_with_apple

Here's an implementation for RN which we may be able to use:
* https://github.com/expo/expo/tree/master/packages/expo-apple-authentication

Note that although it's for React Native, it only works on iOS. It's a thin wrapper around the system-provided components.

We'll also want to support this on Android, because people may sign up using it on iOS and then also want to use Zulip on an Android device. That may have a separate implementation, so I'll file a separate issue for it.

mateuszmandera:
I think this one should be possible to work on without having anything done on the server yet? The first part of the native flow should be happening only between the app and Apple server - at the end of that the app should have a JWT token. All that's left at that point is to send the token with `mobile_flow_otp` to a server endpoint (probably `/complete/apple-id`) and the server will give the standard `zulip://` redirect response - testing this part will require an implementation on the server.

gnprice:
Hmm -- yeah, after looking into this side of things some more, I agree. There's enough complexity involved before the point where we talk to the server that we can productively start working through that even while the server implementation is still being prepared.

OTOH, there is a different prerequisite this is blocked on! That handy existing implementation for RN requires something called [react-native-unimodules](https://github.com/unimodules/react-native-unimodules), which in turn requires us to set up CocoaPods for our build. That's something we've been meaning to do, but will take some work to accomplish. I've just filed #3983 to give it an issue of its own; now that it's blocking this work, we should get on that soon.

So, thanks @mateuszmandera for poking us on this!

chrisbobbe:
@gnprice, when you get a chance, there are some necessary Apple Developer Console actions that you're authorized to take. From Expo's AppleAuthentication doc ([link to "Configuration" section](https://docs.expo.io/versions/latest/sdk/apple-authentication/#configuration)):

> 2. Log into the Apple Developer Console, go to "Certificates, Identifiers, & Profiles" and then "Identifiers".
> [...]
> 4. In the list of identifiers, click on the one corresponding to your primary app. Enable the "Sign In with Apple" capability, click "Edit", and choose the "Enable as a primary App ID" option. Save the new configuration.
> [...]
[...]

PR #881, opened 2017-07-16
Initial Implementation of Sentry.

vishwesh3:
Removed fabric from Android.

Removed fabtic from ios

PR #3987, opened 2020-03-20
Use CocoaPods

### LINKED ISSUE #3983: Set up CocoaPods
CocoaPods is a tool for managing iOS libraries. It's been since forever a thing that many people use in their React Native apps; for example, it's in the standard instructions for integrating RN into an existing app:
https://reactnative.dev/docs/integration-with-existing-apps#3-install-cocoapods
It wasn't in the template sample app, though (the one `react-native init` uses), and so naturally lots of other people didn't use it. We briefly had our own app set up to use it, but then took that out, all back in 2017; see history in 861744be8 (#3870).

In RN v0.60, some work landed that made it work more smoothly with RN than it did before, and that was then followed up by making the template sample app (used by `react-native init`) include it in the build. In RN v0.61, they went further so that IIUC (#3781) it becomes required to use CocoaPods to manage the build for any RN app. So for a while we've planned to switch back to it as one of the things we'll do in the course of #3548, upgrading to RN v0.60.

Now there is a fresh reason to want to do this, with its own deadline! Specifically, for #3964 Apple auth, the handy implementation someone's already made is [a package in Expo](https://github.com/expo/expo/tree/master/packages/expo-apple-authentication) and its install instructions require something called [react-native-unimodules](https://github.com/unimodules/react-native-unimodules) which in turn requires CocoaPods. So, past time we split out an issue specifically for this.

Here's two different sources of instructions on how to go about doing this:
* RN upstream's docs for [integrating into an existing app](https://reactnative.dev/docs/0.59/integration-with-existing-apps), particularly the section ["Configuring CocoaPods dependencies"](https://reactnative.dev/docs/0.59/integration-with-existing-apps#configuring-cocoapods-dependencies).
  * Note that there are tabs near the top of that page for choosing from "iOS (Objective-C)", "iOS (Swift)", and Android. The two iOS tabs have different sample Podfiles there, even though this has nothing to do with which language you write your code in. So that provides two different examples, using the Podspec features in somewhat different ways, which is helpful information for building an understanding of what's going on.
* The install instructions for that [react-native-unimodules](https://github.com/unimodules/react-native-unimodules) library have some pointers of their own for setting up CocoaPods. Naturally this has some overlap with the RN upstream instructions.
  * Note in particular the spot where these instructions have one version for RN <= v0.59 and another for RN >= v0.60. We'll want the former now, and later as part of #3548 we'll presumably move to the latter.

See also some previous discussion [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/cocoapods/near/807626).

chrisbobbe:
EDIT: Adding unimodules was split out into #4016.

Fixes #3983. Also adds react-native-unimodules, to prepare for https://github.com/zulip/zulip-mobile/issues/3964.

Tested on the simulator and device, and all features seem to be working. Possibly a little slower at runtime than before, at least initially, maybe not significantly.

Note in the second commit that we're stuck on a version of react-native-unimodules (0.6.0, no caret) that's not the latest, because at some point between 0.6.0 and 0.7.0-rc.4 (reportedly), support for RN v0.59 was dropped. 0.7.0-rc.4 is reportedly the earliest version that supports AndroidX, so it may be that our AndroidX work (PR #3852) will have to be done simultaneously with the RN v0.60 upgrade (issue #3548). Apparently, backwards compatibility for those on older RN versions is not a priority; see https://github.com/unimodules/react-native-unimodules/issues/100#issuecomment-566436685.


gnprice:
Great! Glad to see this so quickly. :smile: 

I'll have to go connect the Mac I have here (taken home from the office when we all started working from home) to a monitor, so I can play with this... and hmm, first clear some desk space for said monitor. :slightly_smiling_face: 

> 0.7.0-rc.4 is reportedly the earliest version that supports AndroidX

Hrm. As I wrote over on #3852, when an app updates to use AndroidX, it can do so even while it's still using third-party libraries that are still built to use the Android Support Library; so the migration path is, apps go first; then libraries update at their leisure.

There are a lot of people on the Internet confused about that, though -- I didn't really understand it until I did the research that produced #3852. It's possible there's something nasty react-native-unimodules is doing that actually interferes with that nice smooth migration path engineered by the AndroidX authors. But I suspect that instead the library's maintainers are just among the people confused, and that in fact it'll work just fine. I guess we can find out!

chrisbobbe:
> It's possible there's something nasty react-native-unimodules is doing that actually interferes with that nice smooth migration path engineered by the AndroidX authors.
> [...]
> But I suspect that instead the library's maintainers are just among the people confused, and that in fact it'll work just fine.

Mm! Good point. The deprioritization of backwards compatibility and semantic versioning made me leery to dig deeper into the code. It seemed possible that the best record for the current state of the package was in its maintainers' heads, so I tended to trust them when they said certain versions wouldn't work nicely with AndroidX. But you're right, and now I remember the results of your research into AndroidX: it should Just Work (unless, like you said, react-native-unimodules has done something unexpected).

gnprice:
> I'll need to update the docs in that first commit, also, because `pod install` is a required command now. Just tested on a freshly cloned project; the build fails without running `pod install` (of course) but running `pod install` (after running `yarn`; order matters, since the Podfile uses paths in node_modules) is sufficient to get it up and running.

Can we put that in `postinstall` in our `package.json`? See https://github.com/zulip/zulip-mobile/pull/3852/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 for an example.

If there's a part that's hard to automate, then let's document; but when we can automate, automating is better :wink: 

Hmm, I guess we probably need to add "install CocoaPods" to the dev setup instructions, too.

And then the `postinstall` should only run `pod install` if present / if you're on macOS. Probably that's too much logic to stuff into a package.json line, so we should just say `"postinstall": "tools/postinstall"` there and have a script at `tools/postinstall` that handles it. In that case, perhaps best to first merge a version where `pod install` is just in the docs, and follow up to automate it.

ios/Podfile
```diff
@@ -0,0 +1,37 @@
...
+    'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
+    'RCTText',
+    'RCTNetwork',
+    'RCTWebSocket', # Needed for debugging
+    'RCTAnimation', # Needed for FlatList and animations running on native UI thread
+    'RCTImage', # Included in react-native-unimodules' example Podfile
+    # Add any other subspecs you want to use in your project
+  ]
+  # Explicitly include Yoga if you are using RN >= 0.42.0
+  pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'
```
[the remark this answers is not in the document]
chrisbobbe: OK. It's very difficult to find the officially recommended example Podfile for RN v0.59 (see https://github.com/facebook/react-native-website/issues/1603#issuecomment-601491956), so including all of its details (with those outdated comments) was a goal, in case we encountered any issues with these dependencies in the future.

But I do link to that officially recommended example Podfile for RN v0.59, in a code comment at the top, which I guess is enough. And it's true that we're >= each of those versions. I would just hope that if, e.g., 'CxxBridge' starts causing an issue, we don't find some SO answer that says we can just remove it, and do that.

ios/Podfile
```diff
@@ -0,0 +1,37 @@
...
+
+  # Third party deps podspec link
+  pod 'DoubleConversion', :podspec => '../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec'
+  pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
+  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
+
+  use_unimodules!
```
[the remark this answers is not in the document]
chrisbobbe: Mm, I thought I caught that, but obviously not.

Originally, I did everything at once — the CocoaPods work and adding unimodules — and made a single commit. Then, when I thought it would be good to split that into two commits, my strategy was to create a commit after the original commit, deleting the code that added unimodules. And then to reverse the commits.

But to accomplish that reversal, I knew that a simple interactive rebase with the two `pick` commands reversed wouldn't work. That's because the "pick" command applies a computed patch, it doesn't just effect the selected commit's snapshot. Usually applying a patch is what we want, but in this case, I had snapshot A and snapshot B, and I simply wanted to reorder them so B came before A.

I ended up doing an interactive rebase and breaking before the first commit, then doing `git checkout <secondCommit> -- .`, committing that, then `git checkout <firstCommit> -- .`, committing that, then `git rebase --edit-todo`, I dropped the two original commits, and finished the rebase. I'm not sure if this was exactly where this mistake was introduced, but it sure seemed like a lot of maneuvering just to reverse the two commits. Do you know of a better way?

ios/Podfile
```diff
@@ -0,0 +1,37 @@
...
+  pod 'React', :path => '../node_modules/react-native', :subspecs => [
+    'Core',
+    'CxxBridge', # Include this for RN >= 0.47
+    'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
+    'RCTText',
+    'RCTNetwork',
+    'RCTWebSocket', # Needed for debugging
+    'RCTAnimation', # Needed for FlatList and animations running on native UI thread
+    'RCTImage', # Included in react-native-unimodules' example Podfile
+    # Add any other subspecs you want to use in your project
```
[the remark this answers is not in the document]
chrisbobbe: > That is to name here all the subspecs we need
> [...]
> It looks like you can find all the available subspecs by looking in `node_modules/react-native/React.podspec`.

In the Podfile, should I aim to reference all of the subspecs that are listed in React.podspec? I took the example Podfile from the RN v0.59 doc thinking that was the minimum set of subspecs necessary, and that anything on top of that might be pointless or introduce bugs. I can see the argument for being more complete than we may need to be, but I wonder why the example Podfile doesn't include all the subspecs in React.podspec.

Do you think "all the subspecs in React.podspec" includes just the outer level of nesting, here, or might we also need to get the "subspecs of the subspecs" (if that makes sense):

`$ grep '\.subspec \"[a-zA-Z]*\" do' node_modules/react-native/React.podspec`
```
  s.subspec "Core" do |ss|
  s.subspec "CxxBridge" do |ss|
  s.subspec "DevSupport" do |ss|
  s.subspec "RCTFabric" do |ss|
  s.subspec "tvOS" do |ss|
  s.subspec "jsinspector" do |ss|
  s.subspec "jsiexecutor" do |ss|
  s.subspec "jsi" do |ss|
  s.subspec "PrivateDatabase" do |ss|
  s.subspec "cxxreact" do |ss|
  s.subspec "fabric" do |ss|
    ss.subspec "activityindicator" do |sss|
    ss.subspec "attributedstring" do |sss|
    ss.subspec "core" do |sss|
    ss.subspec "debug" do |sss|
    ss.subspec "graphics" do |sss|
    ss.subspec "scrollview" do |sss|
    ss.subspec "text" do |sss|
    ss.subspec "textlayoutmanager" do |sss|
    ss.subspec "uimanager" do |sss|
    ss.subspec "view" do |sss|
  s.subspec "RCTFabricSample" do |ss|
  s.subspec "ART" do |ss|
  s.subspec "RCTActionSheet" do |ss|
  s.subspec "RCTAnimation" do |ss|
  s.subspec "RCTBlob" do |ss|
  s.subspec "RCTCameraRoll" do |ss|
  s.subspec "RCTGeolocation" do |ss|
  s.subspec "RCTImage" do |ss|
  s.subspec "RCTNetwork" do |ss|
  s.subspec "RCTPushNotification" do |ss|
  s.subspec "RCTSettings" do |ss|
  s.subspec "RCTText" do |ss|
  s.subspec "RCTVibration" do |ss|
  s.subspec "RCTWebSocket" do |ss|
  s.subspec "fishhook" do |ss|
  s.subspec "RCTLinkingIOS" do |ss|
  s.subspec "RCTTest" do |ss|
```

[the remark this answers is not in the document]
ray-kraesig: > Hmm, so, I think there's a next step to be done before we're really fully _using_ CocoaPods. That is to name here all the subspecs we need... and remove all the corresponding pieces from our own Xcode project file.

That's certainly something we're going to have to do at some point, but it may be more complexity than is needed for #3964. Does `react-native-unimodules` need a *full* CocoaPods conversion, or will a bare-bones Podfile without the React Native core components suffice? (The latter would probably still be a significant step towards #3983.)

ios/Podfile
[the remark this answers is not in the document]
gnprice: Yeah. There's that link; and also when asking that kind of "why did we put this here?" question, that's always a good prompt to go looking in the Git history, so that'll give us the full context of why we added these lines.

Also ideally when we've fully digested this migration, each subspec we mention here will be here because we really do need it, and have a clear reason why. :slightly_smiling_face: For a chunk of them the reason might be like "these are core pieces of RN used for everything", but which ancient version they were introduced in isn't really part of that information. In our current state of knowledge, basically that link is the explanation.

ios/Podfile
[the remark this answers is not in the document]
gnprice: > Usually applying a patch is what we want, but in this case, I had snapshot A and snapshot B, and I simply wanted to reorder them so B came before A.

One way I might do this is:
```
git reset --soft @^^
# Now your HEAD includes neither commit, but your worktree and index are still at B.

git commit -c @{1}
# Now you have one commit, with tree/snapshot B.

git checkout @{2}^ .
# Now the worktree and index are at A.

git commit -c @{2}^
# Now you have both commits.
```

The `-c` are optional, but convenient for preserving the author dates and commit messages (but still give you an editor to edit the commit messages.)

The @^^ and @{2}^ and so on could be replaced with explicit commit IDs copied from `git log`, but make handy ways to refer to the commits without needing to copy-paste.

When the commits in question are in the middle of a longer series, I might use `git rebase -i` with just a `b` / `break` command inserted after the two commits, and then do the above steps there.

ios/Podfile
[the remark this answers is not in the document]
gnprice: > Does `react-native-unimodules` need a _full_ CocoaPods conversion, or will a bare-bones Podfile without the React Native core components suffice? (The latter would probably still be a significant step towards #3983.)

Yeah, if the fuller conversion turns out to be more complex, I'd be happy merging whatever smaller version suffices to unblock #3964.

Though again, I'll be more confident that we've successfully fully wired up CocoaPods when there are pieces of the build that we rely on and don't have some other way they get included :wink: Without that, it seems like we might attempt #3964 and then discover there's something else we need to do in order for a library we add there to actually get built into the app.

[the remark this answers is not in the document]
gnprice: > In the Podfile, should I aim to reference all of the subspecs that are listed in React.podspec? I took the example Podfile from the RN v0.59 doc thinking that was the minimum set of subspecs necessary, and that anything on top of that might be pointless or introduce bugs. I can see the argument for being more complete than we may need to be, but I wonder why the example Podfile doesn't include all the subspecs in React.podspec

What I think I'd ideally like is to just specify the exact same set as we currently do in the Xcode project file, and at the same time stop specifying those there. I.e. just convert over *how* we include those pieces, separate from any changes to which ones we include.

It's also pretty likely that there are some things we include in the build that we don't need to, largely because the Xcode config has felt so opaque and it's felt so finicky, and liable to cause problems that are annoying to debug, to modify it. So after they're all in this Podfile instead, we might go through and identify some we don't need and remove those.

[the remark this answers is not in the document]
chrisbobbe: > and remove all the corresponding pieces from our own Xcode project file

This sounds easier to solve than determining the correct contents of the Podfile. So far, I haven't found a clear way forward for how to cleanly remove things from the project.pbxproj file. I think this area of the Xcode UI will be helpful:

image

Note that `libPods-ZulipMobile.a` is in that list, and a lot of other, mostly RN-related stuff is also in the list. It seems plausible to want to shrink that list down to just the Pods item, but I'd need a clean mapping from a given `.a` item in that list, to a corresponding added entry in the Podfile.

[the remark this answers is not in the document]
chrisbobbe: Mm, our posts crossed! 🙂Still, to accomplish 

> specify the exact same set as we currently do in the Xcode project file

it sounds like the next step is for me to find a mapping from those `.a` files in that screenshot (assuming that screenshot controls the Xcode project file in the ways we want) to pods/subspecs that can be added to the Podfile.

chrisbobbe:
Yay! Thanks to Greg's help, I'm well on my way to getting this working. I'm investigating two other issues now (a postinstall script to avoid double-counting some font files for RN Vector icons, motivated by https://github.com/oblador/react-native-vector-icons/issues/1074#issuecomment-596761869, and I've seen an instance where Metro doesn't start automatically, but there are clues for a workaround with an extra Build Phase in https://github.com/facebook/react-native/commit/cd8064bc5), but one seems ripe for others to work on independently, so I thought I'd jot it down before lunch here.

One of our dependencies, `@remobile/react-native-toast`, lacks a `.podspec` file. It's very easy to add one (this [article](https://engineering.brigad.co/demystifying-react-native-modules-linking-ae6c017a6b4a#4b1c) agrees, and gives a call to action for PRs to do that), but it hasn't been done yet. In fact, there's a three-year-old PR to do that: https://github.com/remobile/react-native-toast/pull/20, but it hasn't been merged. It appears that there haven't been any updates to that repo at all, for three years. So it may be time to vendor it (make our own local fork with the desired changes), or to find another library that gives us toasts on iOS, that's well-maintained and either has a `.podspec` file or would readily accept a PR that adds one. (Toasts on Android are handily provided by [ToastAndroid from React Native](https://reactnative.dev/docs/0.59/toastandroid).)

Note that the `.podspec` file changes with an upgrade to RN v0.60.0, and it's easy for library maintainers to miss this, or introduce breaking changes by mistake; I'll dump a candidate commit message here with links to see an example:

```
    deps: Lock rn-fetch-blob@0.10.15; breaking change at 0.10.16.
    
    See the warning and apology for this in their README at
    https://github.com/joltup/rn-fetch-blob#version-compatibility-warning.
    
    Specifically, at
    https://github.com/joltup/rn-fetch-blob/commit/0f6c3e3cc (released
    in rn-fetch-blob@0.10.16), the Podspec was updated to match RN's new
    pattern (added in
    https://github.com/facebook/react-native/commit/2321b3fd7, released
    in react-native@0.60.0) of having separate podspecs for each Xcode
    project, instead of using subspecs for them.
    
    We can allow upgrading to 0.10.16 when we switch to
    react-native@0.60.0 (that's open as #3548).
```

chrisbobbe:
Adding "blocked on other work" for #4008.

gnprice:
> to avoid double-counting some font files for RN Vector icons, motivated by [oblador/react-native-vector-icons#1074 (comment)](https://github.com/oblador/react-native-vector-icons/issues/1074#issuecomment-596761869),

Looking at the comment above that one:
> it seems like the crux of the problem is that the font files are being referred to by both the `react-native.config.js` and the `podspec`.
> 
> the reference to the font files as assets in the `react-native.config.js` will cause the `react-native link` to load the fonts as a resource for the _App_ project.

points to a cleaner solution, I think: it's saying that because the font files appear in `react-native.config.js`, they've wound up in our Xcode project config for our app. And indeed they seem to be there:
```
$ git grep vector-icons ios/
ios/ZulipMobile.xcodeproj/project.pbxproj:              2E41C86FA25744F998C2BB02 /* Zocial.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = Zocial.ttf; path = "../node_modules/react-native-vector-icons/Fonts/Zocial.ttf"; sourceTree = "<group>"; };
ios/ZulipMobile.xcodeproj/project.pbxproj:              51E1EDA028654E17A35C5BCA /* MaterialIcons.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = MaterialIcons.ttf; path = "../node_modules/react-native-vector-icons/Fonts/MaterialIcons.ttf"; sourceTree = "<group>"; };
[...]
```

So if we delete those references and let the podspec take care of it, that should resolve the conflict.

gnprice:
And it perhaps says something about the sprawling opaque complexity of doing anything with Xcode build config :stuck_out_tongue_closed_eyes: that people in that thread are finding it easier to run a sed script on a dependency's podspec file than to delete the items from their own app's Xcode config.

chrisbobbe:
> because the font files appear in `react-native.config.js`, they've wound up in our Xcode project config for our app

But `node_modules/react-native-vector-icons/react-native.config.js ` is in node_modules, so a `postinstall` script is the most reliable way we have of making sure it doesn't contain those references, right? I'm not fully sure how, but it sounds like we're both seeing that their presence there is why they wind up in our Xcode config. So just removing them from the Xcode config won't be enough.

chrisbobbe:
Separately, one person believes that using CocoaPods with React Native allows us to use the "Parallelize Build" option in Xcode. I haven't thought through this enough to say whether it's likely to be good long-term, but I did verify that the build doesn't fail by doing so, and I don't notice any runtime errors either. However, one trial without parallelization took 26s, and one trial with parallelization took 24s, so it doesn't seem to shave off that much time anyway. But here's the article that gave me the idea, anyway: https://engineering.brigad.co/demystifying-react-native-modules-linking-ae6c017a6b4a#4f80

gnprice:
> making sure it doesn't contain those references, and that they never get re-introduced, right? I'm not fully sure how, but it sounds like we're both seeing that their presence there is why they wind up in our Xcode config.

I think `react-native link` is the thing that looks at `react-native.config.js`, and acts on it by putting something into our Xcode config.

In any case, the Xcode config is in version control -- so once the references are out, we'll notice if they come back in, and can easily take them out again :slightly_smiling_face: (and debug why they were re-introduced)

chrisbobbe:
> I think `react-native link` is the thing that looks at `react-native.config.js`, and acts on it by putting something into our Xcode config.

That makes sense. `react-native link` can be called with no arguments, just like `yarn install`, so it's tempting to approach it similarly to `yarn install`, as in "Well, I've tried a bunch of specific things, why don't I run this all-purpose, idempotent command that's supposed to fix all kinds of issues with my dependencies?" Basically, we can't recommend that approach. But I don't think we've ever explicitly recommended it, so there's not much lost there. And the question will be moot once we move from [automatic linking](https://reactnative.dev/docs/0.59/linking-libraries-ios#automatic-linking), with `react-native link`, to [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md), without it.

> In any case, the Xcode config is in version control -- so once the references are out, we'll notice if they come back in, and can easily take them out again 🙂 (and debug why they were re-introduced)

Yep, that's true. And modifying things in node_modules in-place adds its own headaches that we've had to think about, even [very recently](https://github.com/zulip/zulip-mobile/pull/4008#issuecomment-605310916). So it'll be nice to avoid doing that; I'm on board with not making a `postinstall` script. 🙂

gnprice:
Re parallelizing the build:
* Sure, we probably should!
* I'm not sure this is related to switching to CocoaPods, at least not directly. From what I can tell from the web -- and the closest thing I can find to actual documentation on this is [a WWDC talk](https://developer.apple.com/videos/play/wwdc2018/408/?time=115) :disappointed: -- the unit of parallelization that this is about is the "target". In particular that means that... in that post's example, the box has no effect at all, because there's just one target shown there.
  * Within a target, there are generally tons of different subtasks, compiling individual files. I can't immediately tell whether Xcode already parallelizes those, or does after this is enabled. I suspect that it already did -- that it builds a whole target, with individual files in parallel where applicable, then the next target, etc. -- and that it continues doing so with this box checked. But that's just a guess.
  * It may be that most of our parallelization opportunities are within a large target, rather than between many small ones. In that case this setting may not make a big difference.
* One thing that will strongly affect how much this helps is how many cores your CPU has. On the office Mac (which I have with me for this WFH period), there's just one physical core, two virtual "hyperthreaded" cores. YMMV.

chrisbobbe:
Makes sense. I've separated that into its own commit, of course; I'm so glad to have something that doesn't have to be included in the giant commit I'm putting together.

Just to jot down a note that I found some Expo packages that will be available to us with unimodules, which we can use instead of two of our third-party dependencies if we so choose. Expo packages will be much better maintained, especially compared with any non-Expo, non-RN-Community counterparts that are designed to do the same thing.

chrisbobbe:
OK, I've pushed my work for today -- this includes the branch from https://github.com/zulip/zulip-mobile/pull/4008, but that's only because it's necessary to get working code; that PR should be reviewed and merged before this one.

chrisbobbe:
Current status on this is that we have a way forward, I just have a few things to do following tons of discussion and help from Greg near [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Android.20build.3A.20unimodules/near/845042) before this can be reviewed again.

chrisbobbe:
OK, following Greg's comment [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Android.20build.3A.20unimodules/near/845050), I think the simplest way for me to test that `unimodules` is working on this PR is to replace our use of `react-native-device-info` with [`expo-application`](https://github.com/expo/expo/tree/master/packages/expo-application) and [`expo-device`](https://github.com/expo/expo/tree/master/packages/expo-device).

chrisbobbe:
Looks like Expo isn't keen on Flow: https://forums.expo.io/t/flow-support-for-expo/1824, which is a shame because, otherwise, I expect Expo packages to be very well maintained. I'm looking at flow-typed now to see what there is.

docs/howto/build-run.md
```diff
@@ -222,6 +222,73 @@ version number starts with "1.8", and with JDK 11, it starts with
 [jdk-lts-roadmap]: https://adoptopenjdk.net/support.html#roadmap
 
 
+### Build failure: java.lang.UnsupportedClassVersionError, "Unsupported class file major version 57"
+
+When trying to build the Android app, you may see this error:
+
+```
+* Where:
+Build file '/Users/jappleseed/dev/zulip-mobile/node_modules/<package path>/android/build.gradle'
```
chrisbobbe, line 231: I'll revise this soon, as I expect your comment at https://github.com/zulip/zulip-mobile/pull/4008#discussion_r401288426 applies here too.

chrisbobbe:
OK, and I think I'm ready for another review! I decided to just test by adding `expo-application`, instead of both `expo-application` and `expo-device`, so I haven't removed `react-native-device-info`. This way it's easier to back out of, if necessary.

chrisbobbe:
> In that case, perhaps best to first merge a version where `pod install` is just in the docs, and follow up to automate it.

Ah, oops! I still need to add this to the docs.

ios/Podfile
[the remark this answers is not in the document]
chrisbobbe: Thanks, I've bookmarked this!

docs/howto/build-run.md
```diff
@@ -222,6 +226,73 @@ version number starts with "1.8", and with JDK 11, it starts with
...
+   - If you used the installation included in Android Studio, take
+     that path minus the "/jre/bin/java" at the end, e.g.,
+     "/Applications/Android Studio.app/Contents/jre/jdk/Contents/Home".
+   - If you installed the Debian/Ubuntu package, then
+     `ls -l /usr/lib/jvm/` should list `java-8-openjdk-amd64` or
+     something similar to that.
+3. Using the path from step 2, run `JAVA_HOME=[that path]`. Then, in
+   the same terminal window, try building again. If it works, add
+   `export JAVA_HOME=[that path]` to your `~/.bashrc` or equivalent,
+   so you don't have to repeat it in the future.
```
gnprice, line 290: So the way this syntax works is a bit counterintuitive:

(a) `$ FOO=bar` just sets `FOO` to `bar` as a variable in the shell, so `"$FOO"` gets turned to `bar`; it doesn't affect the environment, for commands run from the shell.

(b) `$ FOO=bar some-command` runs the command with `FOO` set to `bar` in its environment. It doesn't affect the shell variable `FOO`, or the environment for future commands.

(c) `$ export FOO=bar` sets the shell variable `FOO` just like `FOO=bar`, but also "exports" it, meaning that the same key-value pair `FOO=bar` will be added to the environment for future commands executed.

I think I gave you a command like (b) to use in the course of debugging, something like `JAVA_HOME=... android/gradlew ...`. And (c) is what you'd want to do once debugging/investigation is complete. The form (a) is only really useful in shell programming :-)

[Docs](https://www.gnu.org/software/bash/manual/bash.html#Environment); quick demo:
```
# (a), within shell
$ ( FOO=bar; echo $FOO; )
bar
# (a), within command
$ ( FOO=bar; bash -c 'echo $FOO' )

# (b)
$ ( FOO=bar bash -c 'echo $FOO' )
bar
# (c)
$ ( export FOO=bar; bash -c 'echo $FOO' )
bar
```

Oh, one wrinkle, I guess: if the shell already has `FOO` marked as "exported", then changes to the shell variable's value (with form (a)) will apply to the environment for future commands executed. In particular if `FOO` was in the shell's own environment when it was itself started, then that will have caused it to mark it as exported, so form (a) works:

```shell
# (a), within a shell that got a different value in the environment
$ ( export FOO=bar; bash -c 'FOO=zzz; echo $FOO' )
zzz
```

I think I wouldn't want to rely on that, though -- when the value is intended to appear in the environment, better to be explicit with either (c) or (b).

docs/howto/build-run.md
```diff
@@ -74,6 +74,10 @@ Native upstream on
 You want the tab "React Native CLI Quickstart"; the "Expo CLI Quickstart" does
 not apply.
 
+If you're starting with iOS development, be sure to install CocoaPods,
+as the guide instructs. After you've done that, run
+`cd ios && pod install && cd ..`.
```
gnprice, line 79: People working with iOS will also need to run this after each time they run `yarn install` (or plain `yarn`). There's no great place to document that -- really because we should automate it :slightly_smiling_face: as a near-term followup task -- but probably adding it to this paragraph is the best place.

I ran `rg yarn docs/` to scan for other relevant spots. Probably `docs/howto/ios-tips.md` is a good spot to repeat this point.

And then the release instructions is the other spot I noticed... but actually, better than that is `tools/ios`. That already has a line that runs `yarn`; right after that is a good spot to add:
```bash
    ( cd ios && pod install )
```
(The parens mean these commands are run in a ["subshell"](https://www.gnu.org/software/bash/manual/bash.html#Command-Grouping), a new shell that's a child process of the one running the rest of the script. That keeps the `cd` from mutating the main shell, which is slightly cleaner.)

Or oh hey, actually, better:
```bash
    pod install --project-directory=ios
```
Found that from `pod install --help`. The version with two `cd` commands is likely still the right one for these instructions, for their short lifetime before we automate it away, because I think it successfully optimizes for being easy to casually understand.

chrisbobbe:
> Thanks @chrisbobbe for all this work!
> 
> I think I'd like to split this into two PRs:
> 
> * CocoaPods, and the simple or docs-only followups to that;
> * unimodules (as a new PR).
> 
> That's because I think the CocoaPods part is close to merging, whereas for unimodules I feel like there's a fair amount going on there where I just need to catch up on some of the reading you've done of its docs and code.

Done! Just split that off as #4016.

package.json
```diff
@@ -74,7 +76,7 @@
     "redux-persist": "^4.10.2",
     "redux-thunk": "^2.1.0",
     "reselect": "^3.0.1",
-    "rn-fetch-blob": "^0.10.15",
```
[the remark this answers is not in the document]
chrisbobbe: Sure, sounds good. I've been keeping track of some other things like this, too, so I'll post them there as well, and we can change them if necessary depending on what the final version of this work looks like.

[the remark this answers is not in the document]
chrisbobbe: OK, that draft is at https://github.com/zulip/zulip-mobile/issues/3548#issuecomment-607558197.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -1749,6 +377,94 @@
...
+			inputPaths = (
+			);
+			name = "Start Metro";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
```
gnprice, line 396: So, I think the reason this is newly necessary is that `React.xcodeproj` has its own "Start Packager" build phase, within the "React" target.

I took a look at that one for comparison. It has some helpful features this version lacks, which would be nice to incorporate. (Notably, it avoids bothering to launch the packager for a release build.) ... ha, apparently I was looking at a v0.60 version of React.xcodeproj, oops :slightly_smiling_face: Anyway, that is a good feature.

This is an unfriendly spot to have to edit it, though. It'd be much cleaner as a separate script file.

Anyway, doing anything about that can be a followup. (And then a good PR upstream, I think!)

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -1749,6 +377,94 @@
...
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/FontAwesome5_Solid.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Fontisto.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Foundation.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Ionicons.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/MaterialCommunityIcons.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/MaterialIcons.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Octicons.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/SimpleLineIcons.ttf",
+				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Zocial.ttf",
+			);
```
gnprice, line 421: Hmm -- do you understand why these are appearing in our own Xcode project config? Seems like this should be taken care of by CocoaPods somehow -- like this task should get pushed into the generated Pods project.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -1749,6 +377,94 @@
...
+			name = "[CP] Check Pods Manifest.lock";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+				"$(DERIVED_FILE_DIR)/Pods-ZulipMobileTests-checkManifestLockResult.txt",
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
```
gnprice, line 465: Ooh, clever -- glad CocoaPods has this mechanism.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -1809,6 +525,7 @@
 /* Begin XCBuildConfiguration section */
 		00E356F61AD99517003FC87E /* Debug */ = {
 			isa = XCBuildConfiguration;
+			baseConfigurationReference = 047663CF32A867699B13342E /* Pods-ZulipMobileTests.debug.xcconfig */;
 			buildSettings = {
```
gnprice, line 529: There are some mentions of our newly-podified dependencies in several HEADER_SEARCH_PATHS lists below. Those can probably get deleted too.

(Can't highlight the actual lines, because GitHub.)

Found with `rg react-native ios/ZulipMobile.xcodeproj/`.

docs/howto/build-run.md
chrisbobbe: Oh, OK, thanks! This helps my understanding a lot.

https://github.com/zulip/zulip-mobile/commit/0e45fda87

> I think I gave you a command like (b) to use in the course of debugging, something like `JAVA_HOME=... android/gradlew ...`.

That's right. So, I'll change the instructions so they say to do that.

> And (c) is what you'd want to do once debugging/investigation is complete.

What do you think about the additional recommendation to put it in `~/.bashrc`? This idea (and some other details, like JDK 8 installation instructions) was from https://github.com/zulip/zulip-mobile/commit/0e45fda87.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -923,110 +176,22 @@
...
-		CFCFE50D1F0011FA00C295CF /* Products */ = {
-			isa = PBXGroup;
-			children = (
-				CFCFE52F1F0011FA00C295CF /* libRNFetchBlob.a */,
-			);
-			name = Products;
+			path = Pods;
 			sourceTree = "<group>";
 		};
 		F9E946BE712842C5A02C7F51 /* Resources */ = {
```
gnprice, line 197: This list retains references to the font files from react-native-vector-icons.

I think this particular list doesn't *do* anything -- I think it just controls the "Resources" expandable folder-like thing in the project view in the left sidebar. Not really sure, though.

Looks like these references can be deleted from that same left-sidebar UI, too.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -592,84 +86,14 @@
 			buildActionMask = 2147483647;
 			files = (
 				ED297163215061F000B7C4FE /* JavaScriptCore.framework in Frameworks */,
```
gnprice, line 88: Kind of puzzled why this line remains here. But, it sure is there in the template project! So be it, then.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -592,84 +86,14 @@
...
-				832341BD1AAA6AB300B99B32 /* libRCTText.a in Frameworks */,
-				00C302EA1ABCBA2D00DB3ED1 /* libRCTVibration.a in Frameworks */,
-				139FDEF61B0652A700C62182 /* libRCTWebSocket.a in Frameworks */,
-				05C34EA2C79448BB9767887A /* libRNVectorIcons.a in Frameworks */,
-				D6D40984ED07473499B3F0C9 /* libRNDeviceInfo.a in Frameworks */,
-				78FA95D8058C4372AB199525 /* libRNSound.a in Frameworks */,
-				A83E9861943143E6A10A689E /* libSafariViewManager.a in Frameworks */,
-				ADE3B79538D54689B14FEB10 /* libRNFetchBlob.a in Frameworks */,
-				2318BF63A79049CA9F97E2E0 /* libRNSentry.a in Frameworks */,
 				840D44FCCBB14F97B77D9443 /* libz.tbd in Frameworks */,
```
gnprice, line 89: Also puzzled about this one.

This one isn't in the upstream template -- but it also looks like it never was. It entered our project file in a18c95b59 (#881).

Well, I guess that makes any cleanup of it orthogonal to this PR, then.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -1809,6 +525,7 @@
 /* Begin XCBuildConfiguration section */
 		00E356F61AD99517003FC87E /* Debug */ = {
 			isa = XCBuildConfiguration;
+			baseConfigurationReference = 047663CF32A867699B13342E /* Pods-ZulipMobileTests.debug.xcconfig */;
```
gnprice, line 528: What led to this line (and the similar ones below) being here?

It looks like a useful mechanism. I'm curious about it partly because I can't find anything in the Xcode UI that corresponds to it. There's another reference to the same file in the "Pods" group, and that appears in the left sidebar... but when I go to the ZulipMobile target, then to the Build Settings tab which seems to be the thing this affects, I can't find anywhere that says this file is consulted. Though I *do* see there the setting values that reflect this file being taken into account.

Was this in some install instructions? Or added by something automatic? Or something else? :slightly_smiling_face:

docs/howto/build-run.md
chrisbobbe: Mmm — it also sounds right, even if we do automate it, to simulate not having successfully run `pod install`, and add a troubleshooting entry for that.

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: Noted, good idea!

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: I've responded at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/cocoapods/near/846340.

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: Hmm, I've just found out that some of these can't be deleted without there being build failures; I'm investigating now.

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: > Well, I guess that makes any cleanup of it orthogonal to this PR, then.

OK, noted.

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: I'm about 80% sure I saw someone commenting that it was necessary to preserve this even after CocoaPods; the other 20% is that I guess I might have been thinking of something like [this](https://stackoverflow.com/a/54844185/6792075), which was just someone pointing out its addition in v0.57 or v0.58 (we're already on v0.59).

ios/ZulipMobile.xcodeproj/project.pbxproj
chrisbobbe: Answered this at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/cocoapods/near/846635.

docs/howto/build-run.md
```diff
@@ -222,6 +222,73 @@ version number starts with "1.8", and with JDK 11, it starts with
 [jdk-lts-roadmap]: https://adoptopenjdk.net/support.html#roadmap
 
 
+### Build failure: java.lang.UnsupportedClassVersionError, "Unsupported class file major version 57"
```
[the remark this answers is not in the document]
chrisbobbe: Ah, OK — I had been avoiding first person singular at all costs, frequently favoring the academic "we", but I agree that first person singular is helpful for something anecdotal like this.

docs/howto/build-run.md
chrisbobbe: Mm, to save you the trouble of looking, I didn't get around to doing this today. But I expect there will be other changes I'll need to make from your review, so I can lump this one in with those. 🙂

tools/ios
```diff
@@ -41,6 +41,7 @@ archive_path="${rootdir}"/ios/build/ZulipMobile.xcarchive
 
 do_build() {
     yarn
+    pod install --project-directory=ios
```
gnprice, line 44: Hmm, I suggested this above, but: I'd forgotten that in this script, the working directory is `ios/`! See the `cd` command above.

Probably this is a sign that I should revise this script to use the repo root as its working directory, same as our other scripts. :slightly_smiling_face: For the moment, I'll just fix this up while merging.

gnprice:
Followup items collected from above (and please comment if you see one I've missed):
* small, good to do soon: https://github.com/zulip/zulip-mobile/pull/3987#discussion_r402009262:
  > it also sounds right, even if we do automate it, to simulate not having successfully run `pod install`, and add a troubleshooting entry for that; I can do that tomorrow.
* bigger, for another time: https://github.com/zulip/zulip-mobile/pull/3987#discussion_r401997969
  (move "Start Metro" script into a script file, make it not run on release build)

And then of course #4016 is the major next step :slightly_smiling_face: , taking us toward #3964.

chrisbobbe:
> Followup items collected from above (and please comment if you see one I've missed):

- The actual automation of `pod install` with a `postinstall` script (probably small)

- Small (hopefully?), not introduced in this PR: potential cleanup of "libz.tbd in Frameworks": https://github.com/zulip/zulip-mobile/pull/3987#discussion_r402002988
 
 > This one isn't in the upstream template -- but it also looks like it never was. It entered our project file in a18c95b59 (#881).
 > 
 > Well, I guess that makes any cleanup of it orthogonal to this PR, then.

tools/ios
chrisbobbe: Ah, right, makes sense!

chrisbobbe:
I responded about `libz` at https://github.com/zulip/zulip-mobile/issues/4026#issuecomment-610696643.

PR #3990, opened 2020-03-20
Move exampleData.js into `auxiliary`

ray-kraesig:
... thereby allowing us to get rid of its dummy test.


chrisbobbe:
I'm glad we found a good place to put this! It wasn't always known (https://github.com/zulip/zulip-mobile/pull/3727#issuecomment-566719420) that we could have Jest ignore certain files so they aren't required to have dummy tests.

I bet there are other places we can take advantage of this, but I don't think they should block merging this — in fact, due to the (fortunately) somewhat increasing activity in `exampleData`, it would probably be good to get this merged as quickly as possible, or this PR will gather some rather annoying merge conflicts.

Looks good to me.

chrisbobbe:
Maybe, including something like

> `__tests__/auxiliary` is skipped by Jest (see `jest.config.js`)

also in the first commit might be useful, since this I think the first commit is a little more visible to people wondering why the move was done.

gnprice:
Thanks @ray-kraesig for this cleanup!

I like the idea of putting this file along with `lolex.js` in a subdirectory. Looking at this diff is kind of making me regret the name `auxiliary`, though -- it's a bit of a mouthful to type, and I don't want these test helpers to feel like a pain to go reach for.

How about `lib`? That's short and feels quite conventional -- and in Zulip in particular it's already what we have for `tools/lib/` and the server/webapp repo's `tools/lib/` and `scripts/lib/`.

I'll push a revision of this branch with a quick rename added in front. Please take a look.

ray-kraesig:
> I bet there are other places we can take advantage of this, but I don't think they should block merging this — in fact, due to the (fortunately) somewhat increasing activity in `exampleData`, it would probably be good to get this merged as quickly as possible, or this PR will gather some rather annoying merge conflicts.

Indeed. I've got at least one other PR that's been waiting on this one, specifically to avoid those merge conflicts. :disappointed:

Amended my commit messages per [your suggestion, above](https://github.com/zulip/zulip-mobile/pull/3990#issuecomment-608737583), and merged (including Greg's rename).

Issue #3985, opened 2020-03-20
`colorHashFromString` is almost nonfunctional


ray-kraesig:
`colorHashFromString`:
  * ignores all characters in the string except the second;
  * returns `#NaN` on strings of length zero or one;
  * returns `#NaN` on very long strings;
  * cannot return any color with no red component; and
  * cannot return `#ffffff`.

Maskedman99:
@zulipbot claim

PR #3991, opened 2020-03-21
group-pm: fix colorHashFromString

Maskedman99:
colorHashFromString() now considers all charecters in the string it can also now return colors with red component in it. Also fixed the cases where the function returned #NaN for very long or empty strings.

Fixes: #3985

### DIFF
--- a/src/utils/__tests__/color-test.js
+++ b/src/utils/__tests__/color-test.js
@@ -17,13 +17,21 @@ describe('foregroundColorFromBackground', () => {
 
 describe('colorHashFromString', () => {
   test('returns a 6 digit hex number to use as a color ', () => {
-    const hash = colorHashFromString('John Doe');
-    expect(hash).toHaveLength(7);
+    expect(colorHashFromString('')).toHaveLength(7);
+    expect(colorHashFromString('😃')).toHaveLength(7);
+    expect(colorHashFromString('John Doe')).toHaveLength(7);
+    expect(colorHashFromString('abcdefghijklmnopqrstuvwxyz'.repeat(50))).toHaveLength(7);
   });
 
   test('produces the same output for the same input', () => {
     const hash1 = colorHashFromString('John Doe');
     const hash2 = colorHashFromString('John Doe');
     expect(hash1).toEqual(hash2);
+  });
+
+  test('produces different output for similar inputs', () => {
+    const hash1 = colorHashFromString('John Doe, Juan Pérez');
+    const hash2 = colorHashFromString('John Doe, Jean Dupont');
+    expect(hash1).not.toEqual(hash2);
   });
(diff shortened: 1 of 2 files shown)


Maskedman99:
@ray-kraesig since this function is only used in GroupAvatar.js should I move it into that.

Maskedman99:
@jainkuniya the difference isn't exactly visible in the UI

### **Before:**
Screenshot_20200322-094923_Zulip (debug)

### **After:**
Screenshot_20200322-094812_Zulip (debug)

Maskedman99:
since this function is only used in GroupAvatar.js should I move it into that?
Also, 
> desirable properties are tested – e.g., that similar strings have different hashes,or that short and long strings generate valid values.

you meant the opposite right similar strings have similar hashes

ray-kraesig:
> since this function is only used in GroupAvatar.js should I move it into that?

While that's the only place it's used now, it's conceptually generic and could reasonably be used elsewhere if there were any call for it. Unless that's going to change, it might as well stay where it is.

(It probably wouldn't be worth the churn to move *out*, if it were already there. But it's not.)

> Also,
> 
> > desirable properties are tested – e.g., that similar strings have different hashes,or that short and long strings generate valid values.
> 
> you meant the opposite right similar strings have similar hashes

Decidedly not! That's often [exactly the property you want a hash function *not* to have](https://en.wikipedia.org/wiki/Avalanche_effect); it makes finding collisions easier.

This application doesn't exactly call for a cryptography-grade hash, but it's still desirable that (e.g.) `"John Doe, Juan Pérez"` and `"John Doe, Jean Dupont"` have different colors, even though under the current function they wouldn't. Obviously the [pigeonhole principle](https://en.wikipedia.org/wiki/Pigeonhole_principle) guarantees that this won't be true in the general case, but matches should generally be uncommon.

Maskedman99:
@ray-kraesig thanks for clarifying, It seems I interpreted 'similar strings' as ``"John Doe"`` and 
``"John  Doe"``. I've added the suggested modifications.

chrisbobbe:
Bumping in priority because it's apparent that the bug described in the issue is causing a P0 crash, https://github.com/zulip/zulip-mobile/issues/4038

ray-kraesig:
> Bumping in priority because it's apparent that the bug described in the issue is causing a P0 crash, #4038. I'll try for a simple fix that can ignore the #NaN output, unless @ray-kraesig if you happen to know if this is very close to mergeable (or you can merge with a few tweaks).

As of the last pass, I *think* all issues have been resolved. I'll run a couple of additional tests, but it should be mergeable.

chrisbobbe:
Thanks @Maskedman99 and @ray-kraesig!

PR #3993, opened 2020-03-21
chatScreen [nfc]: Cut down Chat.js, move to ChatScreen.js

jainkuniya:
fixes: #3857

when unread notice is present:
image
image

when unread notice not present:
image
image

Changes:
- Compose box will take up max height as possible.
- Never `UnreadNotice` will be visible in half truncated state.


gnprice:
Quick notes from our video call just now:
* This says `[nfc]` but also that it fixes a bug. Those don't agree with each other. :wink: 
* There are several nontrivial changes in the code, beyond pushing Chat up to ChatScreen:
  * `OfflineNotice` outside `KeyboardAvoider`, becomes inside
  * compose box was outside the reversed container, now inside
  * ...
* Recommended strategy:
  * Make a commit that folds Chat into ChatScreen, but that really is NFC -- take care to not change anything about which components are inside which.
  * Then experiment with these other changes, to see what fixes the bug.

jainkuniya:
@gnprice updated.

Now the first two commits are `nfc` and third one is the fix, and fourth one is improvement.

gnprice:

I left out the last commit, and let's make that a followup. It'd be great to do that cleanup; I just want us to have tested on iOS too (where `KeyboardAvoider` is active).

Issue #3977, opened 2020-03-17
composeBox: Limit maxHeight.


jainkuniya:
Compose box should have some sort of maxHeight. So that user can see some part of messages and type/reply to it. Currently compose box expands to whole screen, completely hiding message list. 

Chat thread: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/limiting.20max.20height.20of.20the.20compose.20box

PR #4001, opened 2020-03-26
composeBox: Limit maxHeight to '50%' of parent.

jainkuniya:
- Provide max height in percentage so that it works in split screen mode,
and even when keyboards appears.

Fixes: #3977


gnprice:
Merged -- thanks!

I tweaked it to 60% first, following our conversation. As you said, we can iterate based on user feedback from here.

PR #4005, opened 2020-03-27
Add LaTeX math support (v3)

### LINKED ISSUE #2660: LaTeX Support in mobile
The Zulip app currently does not render LaTeX formatting correctly.

A simple example from a user in my realm:
* Correctly rendering visual of post in desktop app:
image

* Incorrect render on Android:
image

* source code:
````latex
\text{DRYLAB WEEK OF 6/10/18} 
```
```latex
General
```
- These updates will either come in the form of Hardware/Software/Simulations or Chromastat/Simulations or Chromastat/Game/Simulations

```latex
Chromastat
```
We will be meeting on Tuesday to talk about expectations, deadlines, and get started with work. We have gone in depth with how the device is built and programmed so far. 

```latex
Simulations
````

### PR DESCRIPTION
**Pull request by gnprice**
This is a successor to #3744 and #3679, and fixes #2660.  It sets us up so that the message-list webview now uses KaTeX to show math formulas, just like the webapp does.

The commits that actually set up KaTeX, and then add some fixes to how it displays in certain cases, are @ray-kraesig's and taken from those previous PRs.  The major difference from those PRs is that it doesn't include the complex build-script infrastructure that those PRs added; instead it simplifies our existing `tools/build-webview` script somewhat and then extends it more modestly in order to accomplish what we need for KaTeX.  The approach is based on my [`rsync-refactor`](https://github.com/zulip/zulip-mobile/compare/master...gnprice:rsync-refactor) branch (to commit gnprice/zulip-mobile@6d98d01fb), from the #3679 thread.

Smaller differences include:
 * Updated the KaTeX version.
 * Applied this comment: https://github.com/zulip/zulip-mobile/pull/3679#discussion_r344940954
 * Various changes to commit messages and comments.

I've tested the branch on Android and iOS, and it works in my testing.

I'd appreciate review from @ray-kraesig (and others) on this PR.  Ray and I have discussed exhaustively the strategy question of whether to develop now an infrastructure for complex build scripts, and while he disagrees, I've made the decision that we won't.  But there are plenty of more-specific choices here that I'd be glad to see comments on.

There's one commit toward the end that I still have questions about (left over from #3679), which I'll comment on below.  Up to that point, and including the main commit that fixes #2660, I consider this branch ready to merge.

### DIFF
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -1,8 +1,26 @@
 /* @flow strict-local */
+import { Platform } from 'react-native';
 import type { ThemeName } from '../../types';
 import cssPygments from './cssPygments';
 import cssEmojis from './cssEmojis';
 import cssNight from './cssNight';
+
+/**
+ * Fix KaTeX frac-line elements disappearing.
+ *
+ * This is a hack, but it's probably better than not having fraction lines on
+ * low-resolution phones. It's only known to be useful under Chrome and Android,
+ * so we only include it there.
+ *
+ * See, among others:
+ *   https://github.com/KaTeX/KaTeX/issues/824
+ *   https://github.com/KaTeX/KaTeX/issues/916
+ *   https://github.com/KaTeX/KaTeX/pull/1249
+ *   https://github.com/KaTeX/KaTeX/issues/1775
+ */
+const katexFraclineHackStyle = `<style id="katex-frac-line-hack">
+.katex .mfrac .frac-line { border-bottom-width: 1px !important; }
+</style>`;
 
 export default (theme: ThemeName) => `
 <link rel='stylesheet' type='text/css' href='./base.css'>
@@ -17,5 +35,6 @@ ${cssEmojis}
   display: none;
 }
 </style>
+${Platform.OS === 'android' ? katexFraclineHackStyle : '<!-- Safari -->'}
 <style id="generated-styles"></style>
 `;
--- a/src/webview/css/css.js
+++ b/src/webview/css/css.js
@@ -6,6 +6,7 @@ import cssNight from './cssNight';
 
 export default (theme: ThemeName) => `
 <link rel='stylesheet' type='text/css' href='./base.css'>
+<link rel='stylesheet' type='text/css' href='./katex/katex.min.css'>
 <style>
 ${theme === 'night' ? cssNight : ''}
 ${cssPygments(theme === 'night')}
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -158,8 +158,7 @@ sync() {
 
 # Sync the directory structure, preserving metadata.
 sync "${root}/src/webview/static" "${dest}" <<EOF
-# Ignore files that won't be needed at runtime.
-- README.md
-
-+ *
++ /index.html
++ /base.css
+- *
 EOF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -138,11 +138,28 @@ fi  # if $check_path_name
 # (Build and) sync
 ################################################################################
 
-# Sync the directory structure, preserving metadata. (We ignore any files named
-# 'README.md'; these should be strictly informative.)
+# Sync files from src to dest, reading rsync filter rules from stdin.
 #
-# We use `rsync` here, as it will skip unchanged files.
-mkdir -p "$dest"
-rsync -a --no-D --delete --delete-excluded \
-      --exclude=README.md \
-      "$root/src/webview/static/." "$dest"
+# Both src and dest should be directories.
+#
+# Files in dest not mentioned will be deleted by default.  A rule
+# like `-,r /foo` will override this to leave `foo` in place.
+#
+# For documentation on rsync filter rules, see `man rsync` or:
+#   https://dyn.manpages.debian.org/testing/rsync/rsync.1.en.html#FILTER_RULES
+sync() {
+    # "_dest" because bash's `readonly` means can't shadow with `local`
+    local src="$1" _dest="$2"
+    mkdir -p "${_dest}"
+    rsync -aR --no-D --delete --delete-excluded \
+          "${src}/./" "${_dest}/." \
+          --filter='. /dev/stdin'
+}
+
+# Sync the directory structure, preserving metadata.
+sync "${root}/src/webview/static" "${dest}" <<EOF
+# Ignore files that won't be needed at runtime.
+- README.md
+
++ *
+EOF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -57,7 +57,7 @@ readonly root
 # Parse arguments. Sloppy, but sufficient for now.
 unset target
 unset dest
-sanity_checks=1
+check_path_name=0
 while (( $# )); do
     case "$1" in
         --help|-h|help)
@@ -68,9 +68,10 @@ while (( $# )); do
         --destination)
             # note: this doesn't permit an equals sign after `--destination`
             shift; dest="$1"; shift;;
-        --no-sanity-checks)
-            # disables target-directory sanity checks, for testing's sake
-            shift; sanity_checks=0;;
+        --check-path-name)
+            # enable persnickety checks of our caller's choice of paths,
+            # on the assumption it's the actual app build script
+            shift; check_path_name=1;;
         *) usage; exit 2;;
     esac
 done
@@ -86,9 +87,9 @@ fi
 dest=$(readlink -m "$dest")
 
 # Argument parsing has concluded; argument variables are no longer mutable.
-readonly target dest sanity_checks
+readonly target dest check_path_name
 
-if (( sanity_checks )); then
+if (( check_path_name )); then
     case "$target" in
         ios)
             if [ "$(uname)" != "Darwin" ]; then
@@ -127,7 +128,7 @@ if (( sanity_checks )); then
     if [[ "$dest" != *"/webview" ]]; then
         err "unexpected destination directory '$dest' (expected basename \"webview\")"
     fi
-fi  # if $sanity_checks
+fi  # if $check_path_name
 
 # Set $staging. This is our local staging directory; once it's compiled, it will
 # be synced over to $dest.
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -15,13 +15,18 @@ unset CDPATH
 # print usage message; do not die
 usage() {
     cat >&2 <<EOF
-usage: $0 [ios | android] --destination DEST
+usage: $0 [ios | android] --destination DEST [--check-path-name]
 
-Builds a platform-local intermediate directory containing files used by the
-message-list WebView. The target platform (either \`ios\` or \`android\`) must
-be specified, along with the expected destination directory.
+Build our webview assets, into directory DEST.
 
-(This script is usually run automatically by the build system.)
+This script is mainly for internal use by our build scripts.  They
+go on to use the contents of DEST as the webview-assets folder,
+which our message-list WebView relies on for its static assets.
+See src/webview/MessageList.js for details.
+
+With --check-path-name, make some defensive checks that DEST has
+the form expected from the actual build scripts.
+
 EOF
 }
 
@@ -69,8 +74,6 @@ while (( $# )); do
             # note: this doesn't permit an equals sign after `--destination`
             shift; dest="$1"; shift;;
         --check-path-name)
-            # enable persnickety checks of our caller's choice of paths,
-            # on the assumption it's the actual app build script
             shift; check_path_name=1;;
         *) usage; exit 2;;
     esac
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -133,6 +133,11 @@ if (( check_path_name )); then
     fi
 fi  # if $check_path_name
 
+
+################################################################################
+# (Build and) sync
+################################################################################
+
 # Set $staging. This is our local staging directory; once it's compiled, it will
 # be synced over to $dest.
 readonly staging="$root/src/webview/static"
@@ -141,35 +146,9 @@ if [ ! -d "$staging" ]; then
 fi
 
 
-################################################################################
-# Pre-sync: additional build steps
-################################################################################
-
-# Currently none, but this is where they'll go.
-#
-# TODO: Use something like `make -f src/webview/static.Makefile`, to ensure that
-#   complicated build steps are not run unconditionally.
-#
-# TODO: split assets into build-time-generated and build-time-static?
-#   - rsync can sync from multiple directories, but has no collision-detection
-#   - might be too many layers of abstraction
-
-
-################################################################################
-# Sync
-################################################################################
-
 # Sync the directory structure, preserving metadata. (We ignore any files named
 # 'README.md'; these should be strictly informative.)
 #
 # We use `rsync` here, as it will skip unchanged files.
 mkdir -p "$dest"
 rsync -a --no-D --delete --delete-excluded --exclude=README.md "$staging/." "$dest"
-
-
-
-################################################################################
-# Post-sync: additional build steps
-################################################################################
-
-# None yet.
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -138,17 +138,11 @@ fi  # if $check_path_name
 # (Build and) sync
 ################################################################################
 
-# Set $staging. This is our local staging directory; once it's compiled, it will
-# be synced over to $dest.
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/webview/css/css.js [hunk 1]
```diff
@@ -1,11 +1,76 @@
...
+/*
+  By default, KaTeX renders (non-inline) math into a div of fixed precomputed
+  width. This will be cut off by the edge of the screen when the formula is too
+  long -- and on a mobile device, that's very nearly always.
+
+  The naïve solution of simply giving some part of the KaTeX fragment itself an
+  `overflow-x: auto` style breaks terribly:
+    * Margin collapsing no longer works, causing rendering artifacts. (This is
+      particularly visible on integral signs, which are truncated into near-
+      illegibility.)
```
gnprice: (Continuing from https://github.com/zulip/zulip-mobile/pull/3679#discussion_r344937078 .)

I appreciate the extra few words here, "which are truncated into near-illegibility".

I still don't understand what is meant by "margin collapsing no longer works", though, and I don't understand what goes wrong if you do this "naive" solution.

One reason I appreciate the extra words is in fact that they provide something concrete, so that I can tell if I'm seeing the same thing you're describing! I did the following experiment:
* Checked out this branch just before this commit -- so there's KaTeX, but not this scrolling fixup.
* Opened the app in the Android emulator, and went to a message list with a display-math integral in it.
* The integral looked fine.
* Opened the Chrome devtools on the webview, and edited the stylesheet so that the `.katex-display` element would have `overflow-x: auto`.
* Result: the integral still looked fine and normal. I didn't see anything truncated.
* Another display formula, this one horizontally long so it overflowed the screen, also looked fine, and scrolling it worked as expected.

As a variation, I made one of the integral endpoints unreasonably tall, in case that was the situation you had particularly in mind. Still showed up fine:
Screenshot from 2020-03-26 20-01-42
The ℵ (aleph) is potentially hard to read because it's small, but that seems expected given where it is in the formula; and in particular nothing appears to be truncated.

ray-kraesig: > I still don't understand what is meant by "margin collapsing no longer works", though

Describing this as being caused by margin collapsing [[1]](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) no longer working [[2]](https://stackoverflow.com/questions/19718634/how-to-disable-margin-collapsing/19719427#19719427), is, I *vaguely* recall, based on a comment in a KaTeX GitHub issue. (Which, naturally, I can no longer find. :disappointed:)

> * Checked out this branch just before this commit -- so there's KaTeX, but not this scrolling fixup.
> * Opened the app in the Android emulator, and went to a message list with a display-math integral in it.
> * The integral looked fine.
> * Opened the Chrome devtools on the webview, and edited the stylesheet so that the `.katex-display` element would have `overflow-x: auto`.

I followed this procedure, and got this:
Screenshot_1585442774_scaled
Which is about what I remember seeing. (This would probably not be a bad image to provide a link to (a thousand words, etc.), if GitHub is a reliable-enough imagehost.)

You're probably not seeing it because you're on the latest WebView, where I couldn't reproduce it in this way. (I customarily test on API 21, for essentially this reason.) I think it also breaks on Safari, or at least that it did three months ago; but I'm not sure, and can't presently test it.

gnprice: Ah! Very interesting, thanks -- both the screenshot (a thousand words, etc.), and the repro detail about what platform you tried it on.

I'll go try this on some older versions. I think it's also possible to describe the situation in the comment, and then link to this screenshot as a helpful bonus -- something like: "Some formulas have rendering errors on older browsers (before Chrome NN and Safari MM); in particular, `\int` signs get clipped halfway through."

Another key aspect of this behavior is that it's apparently a browser bug. From the earlier descriptions, I couldn't tell if the issue (as you understood it) was that doing this triggered browser bugs, or that it just interacted badly, through the legitimately-specced arcanities of CSS and of margin-collapsing in particular, with what KaTeX was doing with its CSS.

When the behavior is faithful to the spec, I tend to feel we should understand what's going on better before doing a hack to work around it, because it's likely there's a cleaner way which the spec authors (or people giving them feedback) had in mind. But when it's a browser bug -- and when some broken-seeming behavior is fixed in newer browser versions, that's prima facie evidence it's a browser bug -- then a hack is much more likely to be the best solution available.

gnprice: > API 21

Looking this up, I think you're referring to Android 5.0 Lollipop.

Here's what we have in the "supported platforms" comment in js.js:
```
/*
 * Supported platforms:
 *
 * * We support iOS 10.  So this code needs to work on Mobile Safari 10.
 *   Graceful degradation is acceptable below iOS 12 / Mobile Safari 12.
 *
 *   Graceful degradation is acceptable below Chrome 58.
 *
 *   * These versions are found in stock images for Android 6 Marshmallow
 *     and Android 8 Oreo, respectively, for convenient testing.
 *
 *   * (Note that Android's Chrome auto-updates independently of the OS, and
 *     the large majority of Android users have a fully-updated Chrome --
 *     more recent than the Safari on most iOS devices.)
 *
 *
 * * See docs/architecture/platform-versions.md for data and discussion
 *   about our version-support strategy.
 */
```

So the Chrome 37 that shipped with that release (and that appears in stock emulator images for it) is in the status of "we sometimes fix issues affecting those versions, only when trivial".

I think it'd be quite reasonable to test routinely on Chrome 44, which you can get in an Android 6.0 Marshmallow (aka API 23) stock emulator image. If you test on Chrome 37 and find something broken, it'd be wise to test on Chrome 44 before spending more than a few minutes seeking to fix it.

This particular misrendering, I wouldn't describe as "core functionality" failing to work, either -- it looks ugly, clearly something's broken, but you can still pretty much read everything. So by our policy, it's OK to have (and doesn't justify a complex hack to fix) if it appears only below Chrome 58 (found on an Android 8.0 Oreo stock image) and iOS 12.

ray-kraesig: > Another key aspect of this behavior is that it's apparently a browser bug. From the earlier descriptions, I couldn't tell if the issue (as you understood it) was that doing this triggered browser bugs, or that it just interacted badly, through the legitimately-specced arcanities of CSS and of margin-collapsing in particular, with what KaTeX was doing with its CSS.

Let me clarify then: at the time, I understood it to be the *latter*, and I am not certain even now that it's not. (The fact that Chrome's behavior has changed is certainly evidence in favor, but – CSS being, as you note, CSS – not proof.)

Even if it's not, though – as long as Safari v${latest - 1} (which, as I mentioned, I can't test right now) and stock Android 8.0 (which I can and will) are happy with `overflow: auto`, I expect I will be, too.

My initial tests have [something else going wrong](https://user-images.githubusercontent.com/54861490/77966390-c8a36100-7297-11ea-82cf-3b9be62c0212.png), though, so there may be a bit of a delay there. (This is on Android 10, on a pristine commit 0b3d3b90b362572f3406f8052cce48067c4f96ba, which I'm pretty sure worked on Saturday. :disappointed:)

> So the Chrome 37 that shipped with that release [...] is in the status of "we sometimes fix issues affecting those versions, only when trivial".

Hmm. I'd thought our support for Lollipop was more firm than that.

That said, I typically *do* test interesting bugs against one or more of a small collection of Android images; I just didn't have them to hand, for obvious reasons. If this breaks on Chrome 37 but not Chrome 44, it'll be the first thing I can recall which does.

ray-kraesig: ... and, indeed, it *does* break on stock Marshmallow (API 23, Chrome 44). It doesn't break on Oreo (API 27, Chrome 61).

I haven't tested any intermediate images, because it's moot: even on Oreo, the second issue in that comment – the fact that the equations also become vertically scrollable – is still a problem. (I haven't tested an Android 10 image specifically, but it happens on desktop Chrome 80.)

gnprice: Cool. Certainly worth studying more, then. I didn't get to testing this issue today as much as I'd hoped, but I did just now try this:

> even on Oreo, the second issue in that comment – the fact that the equations also become vertically scrollable – is still a problem.

and I don't reproduce. Same procedure as above, this time with a `\sum... \int... \frac...` that looks like one of your examples in the screenshot above, and (on an Android 8 Oreo emulator image) it looks normal and scrolls horizontally but not vertically, as desired.

So, definitely interested in pinning that down.

... Hm! But `navigator.userAgent` in the dev console there tells me it's Chrome 72. Must have auto-updated :-/ -- normally a good thing, but not for this kind of testing. I'll have to revert that, or if nothing else then make a fresh image.

---

Any comments on the branch before this point? I think even without horizontal scrolling of display math, this is a major improvement over the status quo, both for display math and (what I think is more common) inline math. So I would be very pleased to merge those n-2 commits soon, and then have a followup task to get horizontal scrolling for wide display formulas.

ray-kraesig: > and I don't reproduce. Same procedure as above, this time with a `\sum... \int... \frac...` that looks like one of your examples in the screenshot above, and (on an Android 8 Oreo emulator image) it looks normal and scrolls horizontally but not vertically, as desired.

The exact text was:
````
```math
\lim_{x \to \infty} \int_{0}^{1} \frac{1 + 2x + 3x^2 + 4x^3 + 5x^4 + 6x^5 \cdots}{e^x} \; dx
```
````

Here's a shot of it with the equation scrolled upwards slightly. Note the truncated exponents.
Screenshot_20200330_180355

(The problem is not so much that the exponents are truncated; it's that the equations are all wobbly to the touch. It's not a good look.)

> Any comments on the branch before this point? 

A few things. I had to cancel my review to respond here, but I have them saved; I'll finish them up and post them shortly.

> I think even without horizontal scrolling of display math, this is a major improvement over the status quo, both for display math and (what I think is more common) inline math. So I would be very pleased to merge those n-2 commits soon, and then have a followup task to get horizontal scrolling for wide display formulas.

Wait, what? Why *not* the last two? The existing horizontal-scrolling code is functional everywhere without issues, as far as I can tell. If investigation shows that something simpler is feasible, we can swap it out later without affecting users.

Besides which, I don't think leaving them out leaves the app in a good state. It's possible for an embedded equation to be cut off at a point where it's not obvious that it's not all there.

File: tools/build-webview [hunk 2]
```diff
@@ -68,9 +68,10 @@ while (( $# )); do
         --destination)
             # note: this doesn't permit an equals sign after `--destination`
             shift; dest="$1"; shift;;
-        --no-sanity-checks)
-            # disables target-directory sanity checks, for testing's sake
-            shift; sanity_checks=0;;
+        --script)
```
ray-kraesig: It's a good idea at its core; but `--script` is not a meaningful name, even with the documentation added in the following commit. 

(Why not just `--sanity-checks`?)

tools/build-webview
```diff
@@ -138,17 +138,11 @@ fi  # if $sanity_checks
...
-fi
-
-
 # Sync the directory structure, preserving metadata. (We ignore any files named
 # 'README.md'; these should be strictly informative.)
 #
 # We use `rsync` here, as it will skip unchanged files.
 mkdir -p "$dest"
-rsync -a --no-D --delete --delete-excluded --exclude=README.md "$staging/." "$dest"
+rsync -a --no-D --delete --delete-excluded \
```
ray-kraesig: The staging directory was never intended as a speed optimization; the first two paragraphs in this commit are attacking a strawman.

(It was primarily intended for transparency and verifiability, as part of a check against stale files being left in the build directories. The details are probably no longer relevant enough to discuss here.)

src/webview/css/css.js
```diff
@@ -6,6 +6,7 @@ import cssNight from './cssNight';
 
 export default (theme: ThemeName) => `
 <link rel='stylesheet' type='text/css' href='./base.css'>
+<link rel='stylesheet' type='text/css' href='./katex/katex.min.css'>
```
ray-kraesig: This is the only line of any part of this commit — summary, message, or diff — that I can reasonably be said to have written.

Please adjust its `Author:` field accordingly.

src/webview/css/css.js
gnprice: Great -- thanks for the repro, and for the comments on the rest!

> Besides which, I don't think leaving them out leaves the app in a good state. It's possible for an embedded equation to be cut off at a point where it's not obvious that it's not all there.

The webapp has the same behavior. Here's [two messages](https://chat.zulip.org/#narrow/stream/7-test-here/topic/latex/near/844938) with the same content `1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 + 11 + 12 + 13 + 14 + 15 + 16 + \cdots`, except one is in an inline math formula and the other in display math:

image

We've heard very clearly from users that the current state in the mobile app is that LaTeX doesn't work, that they're very happy they can use LaTeX in the webapp, and that they want it in the mobile app too. It'll be good to fix this edge case, but I am not going to block on it for shipping the feature.

gnprice: (Meant to reply directly to this part too, and forgot.)

> Why *not* the last two? The existing horizontal-scrolling code is functional everywhere without issues, as far as I can tell.

Complexity. The app has had a number of complex hacks in it that were made without clearly understanding, or simply without clearly explaining, what problem they were solving and how and why. And it's suffered from that.

There's a place for complex hacks, and there may well be a place for this one. But we don't have a clear enough understanding at this point to be ready to make it.

src/webview/css/css.js
gnprice: OK -- changed.

Here's why I left your `Author:` line in place, though: this commit incorporates your work determining
* we need the CSS but not the JS, despite the install instructions just saying to use both
* we need the WOFF2 and not the other font formats, and the browser support details that make that the case
* the choice of where and how to lay out the KaTeX files

So my inclination if marking myself as author would be to have a `Co-authored-by:` line for you. I've left that out in this revision, because that feels most consistent with the spirit of your request here, and I want to respect your preferences on whether to claim this authorship. But I'd be glad to add such a line if you're OK with that.

tools/build-webview
gnprice: Yeah, I don't love the name. I wanted something non-specific about *what* the checks are, basically a name that summarizes the description in the new comment / in the subsequent usage message, but "script" isn't very clear in doing that.

I didn't go for `--sanity-checks` because I don't like the implication there's something wild or wrong about a different destination path. In particular when you ask some code to do something in conditions that it has "sanity checks" that would reject if you enabled them, that feels like the code is free to do wacky things that don't correspond to what it would do in proper conditions. Here, that would undermine the usefulness of being able to pass a convenient nonce directory and inspect the output there.

How about `--check-path-name`? Revised to that. (I like "name" a bit better than just `--check-path` or `--check-dest`, because either of the latter sound like they might check things like what's *inside* the dest directory.)

tools/build-webview
gnprice: Interesting -- that reason is news to me :slightly_smiling_face: I guess the sharing was the best explanation I could reverse-engineer for it -- as the message said, there are circumstances not *too* different from these where something like it would be helpful.

Revised to:

    This "staging" step was intended as a way to help the developer
    hand-inspect the output for stale files.  Better to prevent them
    automatically, by continuing to use tools like `rsync --delete`.
    Meanwhile, the extra indirection complicates the code and
    complicates reasoning about it.  So, simplify it away.

gnprice:
Pushed just now the revisions mentioned above, and without those last two commits.

I'll go file an issue about the fact that overfull hboxes^W^W wide display-math formulas can leave part of the formula invisible. It's a zulip/zulip issue as it stands, and once this is merged it'll be a mobile issue too, which we might or might not solve in the same way.

src/webview/css/css.js
gnprice: Filed zulip/zulip#14422 for the issue in the existing webapp.

gnprice:
OK -- this is now merged!

I've tested it back to iOS 10 and Android 6 Marshmallow (with Chrome 44), and it works beautifully, with that known issue that's the equivalent of zulip/zulip#14422. I'll go file a bug for that one.

Thanks @ray-kraesig for the review, and thanks again for your work on this earlier.

src/webview/css/css.js
ray-kraesig: For the record, very belatedly: I do indeed feel that co-authorship would not have been appropriate here.

What I would probably write in a similar case would be something along the lines of "&zwj;(Based in part on/Supersedes/Inspired by) earlier work by...", with no pseudoheader. (As a concrete example, see the commit message of 988736b9862542aa9f9e6ca8e3ebe07bb52fc95b.)

PR #4012, opened 2020-03-31
UI improvement proposal: minimalize message interface

tinnieteal:
1. left align message and avatar pic for messages from myself
2. hidden redundant information in chat interface, i.e. user name
@gnprice 
First PL from gsoc 2020 candidate Tina Chen, not so sure about my UI design and coding practice. looking forward to your comments!


chrisbobbe:
Hi, @tinnieteal, welcome! 🙂 

Since these are changes to the visual design, could you please post some screenshots here? That'll be helpful for an initial review.

ray-kraesig:
* The two things this commit is alleged to do are very different; at the very least they should be separated out into different Git commits. They should probably also be considered as separate UI design changes (perhaps with one contingent on the other).
* There are several [CI failures](https://travis-ci.org/github/zulip/zulip-mobile/builds/669316060) which will need to be addressed. These look like linting and typechecking errors; the former can be resolved by running `prettier`, but the latter will take some work.
  * There are also a number of nonfunctional spacing changes to the template strings. Are you using a different formatter than we are, by any chance?
* As Chris notes, screenshots are a must for UI changes. (Preferably before/after comparisons.)

tinnieteal:
Hello @chrisbobbe and @ray-kraesig

Thanks for your comments and suggestions!
1. I have attached the UI change screenshot before and after 
2. Since the two UI changes are interconnected, I wonder if you would still suggest me to separate the commit for each changes. I think the two changes have to make sense when sticking together. 
3. I have fixed the linter test, but this is still test failing
- flow test: that Outboxed does not property `sender_id`, and it seems that I cannot simply fix it by adding this field to flow type definition, since there are other part of the code depending on this original definition. I have not look into that part of code but I wonder if you know a simple fix to this. 
- mock test: I wonder what  practice I should follow to fix mock test. Should I change each of these test definition to reflect the UI change?

Regards,
Tina

ray-kraesig:
> 1. I have attached the UI change screenshot before and after

:+1:

Looking at the changes makes me suspect that this isn't really a change we want, though. It'd be a significant departure from the webapp, and I think also from other team-collaboration software (Slack, Mattermost, etc.), on both mobile and desktop.

It might be best to post these screenshots on [chat.zulip.org's #mobile stream](https://chat.zulip.org/#narrow/stream/48-mobile) for slightly more lively discussion.

> 2. Since the two UI changes are interconnected, I wonder if you would still suggest me to separate the commit for each changes.

Almost certainly!

* In the first commit, you can reverse the presentation of self-sent messages.
* In the second commit, you can remove usernames in one-on-one PMs.

These two things probably wouldn't go in separate *pull requests*, but you can include a series of related Git commits in a single PR.

If there's still some reason that this can't easily be done... well, I for one would find that very surprising, and worthy of some discussion and explanation here.

> * flow test: that Outboxed does not property `sender_id`, and it seems that I cannot simply fix it by adding this field to flow type definition, since there are other part of the code depending on this original definition. I have not look into that part of code but I wonder if you know a simple fix to this.

That's Flow pointing out (what I assume is) a real bug: `Outbox` objects really don't have `sender_id`s, so that comparison will always fail. The consequences of this are visible at the bottom of your last screenshot, where the unsent message has your avatar on the left.

You'll want to test `message.isOutbox`, which Flow understands as a discriminant between the two types, and (presumably) ensure that `isMyself` is true for it. (Flow should even understand something along the lines of `const isMyself = message.isOutbox || message.sender_id === ...`)

> * mock test: I wonder what  practice I should follow to fix mock test. Should I change each of these test definition to reflect the UI change?

Exactly so: if the expected behavior changes, the tests need to as well.

Issue #4013, opened 2020-03-31
Translation errors appears on some screen 


sumukhah:
Some words are not added for translation resulting in error.

![1](https://user-images.githubusercontent.com/23723464/78057249-0a213280-73a4-11ea-9feb-27992553172c.jpg)

sumukhah:
Working on the issue

ray-kraesig:
Closed by #4013... for now, at least. There are doubtless a few untranslated strings still floating around, but this does clear up a nice chunk of them.

PR #4015, opened 2020-04-01
Translation: Fix translation error on some screens

sumukhah:
Some strings are used in the app but not added to the messages_en.json.
Due to this application fails to translate them and throws console error.
Add those strings to messages_en.json. Sometimes in the future these
strings will be pushed to Transifex which generates translated results
for all other languages.

strings containing emojis are added as it is to avoid complication in the
codebase.

Fixes: #4013


chrisbobbe:
Thanks, @sumukhah, I saw your screenshot of the error on the issue page, and this sure looks like something that should be fixed! 🙂 

Could you explain how you determined which strings were missing? That can help us be more confident that we have exactly the strings we need (no more, no less).

sumukhah:
pick any language (I chose Hindi).
navigate to different screens, console error will occur and displays the exact word which has not been added for translation.
Add that word to **messages_*.json**(in mycase messages_hin.json)
for checking further.

ray-kraesig:
> Could you please explain how you chose which strings to add? That can help us be more confident that we have exactly the strings we need (no more, no less).

Alas, we are probably never going to be confident that there aren't more strings that will need translating.

> Add that word to **messages_*.json**(in my case messages_hin.json)
> to check further.

Unfortunately this introduces (or, arguably, surfaces) a bug. If (*e.g.*) `"🌴 Vacationing": "🌴 छुट्टी पर"` is added to `messages_hi.json`, then "🌴 छुट्टी पर" will correctly appear as the label for an option on the UserStatusScreen – but tapping it will cause "🌴 Vacationing" to be filled in as the status, not "🌴 छुट्टी पर"!

sumukhah:
Thank you @ray-kraesig, 
I have committed changes that deal with the issue mentioned above.

src/user-status/UserStatusScreen.js
```diff
@@ -109,4 +113,4 @@ class UserStatusScreen extends PureComponent<Props, State> {
 
 export default connect(state => ({
   userStatusText: getSelfUserStatusText(state),
-}))(UserStatusScreen);
+}))(injectIntl(UserStatusScreen));
```
ray-kraesig, line 116: We have a less intrusive convention for this. See the documentation for `type GetText` in `src/types.js`.

static/translations/messages_en.json
```diff
@@ -199,5 +199,17 @@
...
+  "User status": "User status",
+  "Muted": "Muted",
+  "Log in with GitLab":"Log in with GitLab",
+  "No topics found":"No topics found",
+  "What's your status?":"What's your status?",
+  "📅 In a meeting": "📅 In a meeting",
+  "🚌 Commuting": "🚌 Commuting",
+  "🤒 Out sick": "🤒 Out sick",
+  "🌴 Vacationing":"🌴 Vacationing",
+  "🏠 Working remotely":"🏠 Working remotely"
```
ray-kraesig, line 214: These should follow the same formatting as the other entries: a single space after the colon.

Maskedman99:
Looks like fixing this should also fix #3406

ray-kraesig:
I've made some edits to the commit messages, and (in a separate commit) regrouped some text strings, more for aesthetic reasons than functional ones. :upside_down_face: Merging. Thanks!

> Looks like fixing this should also fix #3406

Good catch! Added during the commit-message editing pass.

sumukhah:
Thank you @ray-kraesig

PR #4016, opened 2020-04-02
Add `react-native-unimodules` so we can use Expo packages

chrisbobbe:
Split off from #3987, following Greg's [comment](https://github.com/zulip/zulip-mobile/pull/3987#pullrequestreview-385942754):

> I think I'd like to split this into two PRs:
> * CocoaPods, and the simple or docs-only followups to that;
> * unimodules (as a new PR).
> 
> That's because I think the CocoaPods part is close to merging, whereas for unimodules I feel like there's a fair amount going on there where I just need to catch up on some of the reading you've done of its docs and code.


gnprice:
P1 because this is on our path to #3964.

ios/Podfile.lock
```diff
@@ -1,6 +1,18 @@
 PODS:
   - boost-for-react-native (1.63.0)
   - DoubleConversion (1.1.6)
+  - EXApplication (2.1.0):
+    - UMCore
```
gnprice, line 5: IIUC how this works, I think this part should happen only in the second commit.

(In this instance, the main value of getting this right is probably that it helps this commit serve as a clean demonstration of how unimodules work! But also I think it avoids having the first commit be broken. So it's also an example of how "every commit keeps things working" helps with "each commit is understandable in itself".)

src/utils/userAgent.js
```diff
@@ -1,9 +1,10 @@
 /* @flow strict-local */
 import { NativeModules } from 'react-native';
 import DeviceInfo from 'react-native-device-info';
+import { nativeApplicationVersion } from 'expo-application';
 
-const { getVersion, getSystemName, getSystemVersion } = DeviceInfo;
+const { getSystemName, getSystemVersion } = DeviceInfo;
 
 export default !NativeModules.RNDeviceInfo
```
gnprice, line 8: Huh, this conditional is odd. Dates back to 9d9720c74ccf0aebb35a51e018ced676dbc7970a.

ios/Podfile.lock
chrisbobbe: Hmm, it looks like you've selected lines 4 and 5:

```diff
+   - EXApplication (2.1.0):
+     - UMCore
```

and these lines were added only in the second commit, if I'm reading my graphical Git client and the GitHub UI correctly. 🙂 

Or are you referring to all of these additions to the Podfile.lock, e.g., EXAppLoaderProvider, EXConstants, EXFileSystem, EXPermissions, and a bunch of things prefixed with UM? If so:

It looks like the `use_unimodules` script called from the Podfile (it's at node_modules/react-native-unimodules/cocoapods.rb) goes through node_modules and, for any module that has a `unimodule.json` file in it, includes it with `pod "#{pod_name}", path: podspec_directory` (the same way other pods are included in the Podfile), which explains how these lines are getting added in the Podfile.lock. So, after running Yarn at this commit, to make sure node_modules is in the correct state, we'll want to be sure we know the reason why any module with a `unimodule.json` is in node_modules.

We can take exactly the same input, or nearly — just those `unimodule.json` files, without looking elsewhere in a given module — and ask Yarn why they're there:

`for f in node_modules/*/unimodule.json; do yarn why $(cat $f | jq -r '.name'); done` (I'd appreciate feedback on this little script, if you have any 🙂)

and all of them tell me the same story, namely:

```
info Reasons this module exists
   - "react-native-unimodules" depends on it
```

And 'react-native-unimodules' is, correctly, the only new dependency we're adding to the `package.json` in this commit.

chrisbobbe:
> Looks like it's basically doing the same thing that's more recently arrived for the broader RN ecosystem with "autolinking". Does that sound right from your reading?

Yep! The main work is being done by a Ruby script that calls `pod "#{pod_name}", path: podspec_directory` for "unimodule" modules (i.e., those that have a `unimodule.json` in them) in node_modules. The `pod ...` syntax is the same as we've been using to manually add pods to the Podfile, so it saves that bit of manual work.

ios/ZulipMobile/AppDelegate.m
```diff
@@ -39,6 +44,14 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   return YES;
 }
 
+- (NSArray<id<RCTBridgeModule>> *)extraModulesForBridge:(RCTBridge *)bridge
+{
+  NSArray<id<RCTBridgeModule>> *extraModules = [_moduleRegistryAdapter extraModulesForBridge:bridge];
```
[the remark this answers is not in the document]
chrisbobbe: > Do you?

No! 😄

Yep, that reasoning looks good to me; I just read the relevant parts.

chrisbobbe:

OK, I've just made the requested changes, and left a query about clean commits at https://github.com/zulip/zulip-mobile/pull/4016#discussion_r404473651. 🙂

ios/Podfile.lock
gnprice: > and these lines were added only in the second commit, if I'm reading my graphical Git client and the GitHub UI correctly. :slightly_smiling_face: 

Hmm! Indeed they were. Sorry about that -- I misread the branch. Those lines (and the other `EXApplication`-related lines in this file) were the ones I had in mind.

I think I was actually fooled by the `Podfile.lock -diff` attribute setting which I asked for on the previous PR <_< >_>... or perhaps by the combination of that and the fact that I have `yarn.lock diff` in my `.git/info/exclude`, causing `yarn.lock` and `Podfile.lock` to behave differently in the diffs I see.

So in the second commit (which is now the third commit -- the one adding `expo-application`), I saw it appearing in `yarn.lock`, meaning it wasn't in `node_modules/` before that point... but didn't see any diff in the `Podfile.lock` -- there would have been a block like this:
```diff
diff --git ios/Podfile.lock ios/Podfile.lock
index 24f59e6bd..3c581ef7b 100644
Binary files ios/Podfile.lock and ios/Podfile.lock differ
```
but that's easy to skim past -- and inferred it was already added in the first commit.

I think the immediate lesson is I'll be sure to keep those two files either both at our default of `-diff`, or both overriden to `diff`, in my `.git/info/exclude` so I don't see a spurious mismatch between them.

src/utils/userAgent.js
gnprice: Thanks for the diagnosis and cleanup!

ios/Podfile.lock
chrisbobbe: No problem! 😄

ios/ZulipMobile/AppDelegate.m
```diff
@@ -17,11 +17,16 @@
...
+#import <UMReactNativeAdapter/UMNativeModulesProxy.h>
+#import <UMReactNativeAdapter/UMModuleRegistryAdapter.h>
 
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+  self.moduleRegistryAdapter = [[UMModuleRegistryAdapter alloc] initWithModuleRegistryProvider:[[UMModuleRegistryProvider alloc] init]];
+
   RCTSetLogThreshold(RCTLogLevelError);
```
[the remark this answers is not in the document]
gnprice: (This fix happened in the expo-application commit; moved to this first commit, with `git rebase -i` / `git co -p master $file`.)

gnprice:
Thanks for the revisions! Sorry for the confusion about that Podfile.lock change.

I've made a small tweak mentioned above, and about to merge... oh, and I guess I should actually build and run from the branch first. :stuck_out_tongue_closed_eyes: Doing that now, then plan to merge assuming all's as expected.

ios/ZulipMobile/AppDelegate.m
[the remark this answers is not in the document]
chrisbobbe: > (This fix happened in the expo-application commit; moved to this first commit, with `git rebase -i` / `git co -p master $file`.)

Oh goodness, you're right. Thanks for catching and fixing that!

gnprice:
Oh, actually, there's this warning:
```
$ yarn
yarn install v1.21.1
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "linux" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > expo-application@2.1.0" has unmet peer dependency "@unimodules/core@~1.0.0".
[4/4] Building fresh packages...
Done in 3.44s.
```

The unmet peer dependency. Would you find a combination of package versions that avoids that? It's probably fine, this is a simple package, but that general class of issue can cause bugs -- and I like to keep routine flows like `yarn install` free of warnings.

chrisbobbe:
> The unmet peer dependency. Would you find a combination of package versions that avoids that?

Sure! I'll move this to [CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/unimodules.20unmet.20peer.20dependency/near/849254) as I expect this to be slightly complicated, given that we're stuck on react-native-unimodules@0.6.0 (no caret)

gnprice:
And merged. Thanks!

Issue #4019, opened 2020-04-04
Autocomplete popup is no longer appearing


Maskedman99:
i.e I am not able to see this popup anymore
![image](https://user-images.githubusercontent.com/31368194/78453092-2cb99100-76ad-11ea-8d25-a3a5d90b6cec.png)

This is happening in the new beta version as well as in the master branch.

Maskedman99:
O.K I see where it went, It's hiding in the navbar
![Screenshot_20200404-195902_Zulip (debug)](https://user-images.githubusercontent.com/31368194/78453435-5a073e80-76af-11ea-8b53-b736bdb1f02b.jpg)

Maskedman99:
Apparently in ChatScreen.js message list is being rendered after compose box, so whenever the autocomplete is rendered the message list is being rendered above it hiding the autocomplete.

Reversing the changes in the commit https://github.com/zulip/zulip-mobile/commit/955534d75b7140688325c004011dc62b26cb87e8 I am able to bring the auto-complete back, This commit was made to address issue https://github.com/zulip/zulip-mobile/issues/3857 which I think is no longer possible to reproduce because of this commit https://github.com/zulip/zulip-mobile/commit/54cf185eee3fb37a63eacd13f966446e04b7a568

jainkuniya:
Thanks @Maskedman99 for the repot.

Can you please once verify this bug on 5328d3e1, we are planning to red ride of `reverse` style.

Maskedman99:
Actually, I am using ```reverse``` style in the PR, if we're getting rid of ```reverse``` , I will make the necessary changes in the PR and squash it

PR #4020, opened 2020-04-04
composeBox: fix auto-complete not showing

Maskedman99:
In ChatScreen.js, message list is being rendered after compose box auto-complete is implemented inside composeBox, so whenever the autocomplete is rendered the message list is being rendered above it thus hiding the autocomplete.

This is fixed in this PR by rendering the MessageList before compose box.

Screenshot-20200404233021-945x654
----------------Before------------------------------------------------------------------------------After----------------


gnprice:
Fixes #4019. Merging -- thanks @Maskedman99 for the fast bug report and clean fix!

PR #4021, opened 2020-04-06
chatNavBar: fix gap between navBar and topmost header

Maskedman99:
There is a gap of 1px between navigation Bar and the top header in chat screen. This is caused by applying  bottomBorder on navBar. This PR removes this.

Screenshot-20200407015649-881x318
---------------------Before-----------------------------------------------------------------------After----------------------------

### DIFF
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -2,6 +2,7 @@
 
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
+import Color from 'color';
 
 import type { Dispatch, Narrow } from '../types';
 import { LoadingBanner } from '../common';
@@ -40,7 +41,10 @@ class ChatNavBar extends PureComponent<Props> {
     return (
       <View
         style={{
-          borderColor: 'hsla(0, 0%, 50%, 0.25)',
+          borderColor:
+            backgroundColor === 'transparent'
+              ? 'hsla(0, 0%, 50%, 0.25)'
+              : Color(backgroundColor).darken(0.1),
           borderBottomWidth: 1,
         }}
       >


ray-kraesig:
I'm less sure about this. I *think* the purpose of this border is to provide a visible separation between the nav-bar and the content background when the former is a color that blends in too well with the latter. For example, pure white, in the default theme:

verona-side-by-side

As @gnprice noted in the screenshot, though, in other contexts it looks more like a gap than a border. Could that be fixed with just a color adjustment?

Maskedman99:
@ray-kraesig I've made changes, the borderColor now matches with the StatusBar Color, this should avoid the blending in.

gnprice:
Thanks for this debugging! I would not have guessed that was the cause.

This doesn't seem to have the right effect on PM threads -- the border disappears:
Screenshot_1586897329

I think the reason is that `backgroundColor` there is `'transparent'`, and so a darkened version is still transparent.

Maskedman99:
@gnprice thanks for pointing it out, I've made changes

Screenshot-20200415125331-940x218

ray-kraesig:
Looks good; merging. Thanks!

I've made some edits to the commit summary and message – solely for formatting purposes, as all the necessary information was present. You may want to review our [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages)... and, on the other hand, *we* may want to consider explicitly noting that there should be two newlines after a paragraph.

gnprice:
> our [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages)... and, on the other hand, _we_ may want to consider explicitly noting that there should be two newlines after a paragraph.

We do say on that page:
> The body is written in prose, with full paragraphs; each paragraph should be separated from the next by a single blank line.

I'm sure it's possible to make the guidelines clearer, though.

ray-kraesig:
> We do say on that page:
> 
> > The body is written in prose, with full paragraphs; each paragraph should be separated from the next by a single blank line.

zulip/zulip@8b72de1d4dce791fd6a655a48fd65ed5d8f2e1bd

gnprice:
Haha, got it. Thanks. :wink: And glad to have the cross-reference here now!

PR #4022, opened 2020-04-07
typing: Serialize recipients based on server version.

### LINKED ISSUE #3732: Typing notifications are mistyped, resulting in rejection (for servers < 2.0)
Per [the docs for the typing-notifications API](https://zulipchat.com/api/typing), the Zulip server expects `to` to be a list of email addresses, not a list of user IDs. Unfortunately, we're sending the latter.

The current version of the Zulip server appears to accept numeric IDs (@andersk, can you confirm?) but errors in Sentry indicate that responses from servers along the lines of `Bad value for 'to': [17,42,105]` do occur in the wild.

agrawal-d:
Zulip server versions >= 2.0.0-rc1 require user ID arrays in the
'to' argument for the typing endpoint, whereas we currently send
email arrays.

Check the Zulip version and use the appropriate format of the
argument, i.e., for versions >= 2.0.0-rc1 or when the Zulip server
version is undetermined, use user ID arrays, otherwise, use email
arrays.

Fixes #3732.

### DIFF
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -34,12 +34,7 @@ const typingWorker = (state: GlobalState) => {
   // (zulip/zulip@2f634f8c0). For versions before this, email arrays
   // are used. If current server version is undetermined, user ID
   // arrays are optimistically used.
-  let useEmailArrays: boolean;
-  if (serverVersion === undefined) {
-    useEmailArrays = false;
-  } else {
-    useEmailArrays = !serverVersion.isAtLeast(new ZulipVersion('2.0.0-rc1'));
-  }
+  const useEmailArrays = !!serverVersion && !serverVersion.isAtLeast(new ZulipVersion('2.0.0-rc1'));
 
   const getRecipients = user_ids_array => {
     if (useEmailArrays) {
--- a/src/users/usersActions.js
+++ b/src/users/usersActions.js
@@ -1,12 +1,13 @@
 /* @flow strict-local */
 import * as typing_status from '@zulip/shared/js/typing_status';
 
-import type { Dispatch, GetState, Narrow } from '../types';
+import type { Auth, Dispatch, GetState, GlobalState, Narrow } from '../types';
 import * as api from '../api';
 import { PRESENCE_RESPONSE } from '../actionConstants';
-import { getAuth, tryGetAuth } from '../selectors';
+import { getAuth, tryGetAuth, getServerVersion } from '../selectors';
 import { isPrivateOrGroupNarrow, caseNarrowPartial } from '../utils/narrow';
-import { getAllUsersByEmail } from './userSelectors';
+import { getAllUsersByEmail, getUserForId } from './userSelectors';
+import { ZulipVersion } from '../utils/zulipVersion';
 
 export const reportPresence = (isActive: boolean = true, newUserInput: boolean = false) => async (
   dispatch: Dispatch,
@@ -25,17 +26,40 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
   });
 };
 
-const typingWorker = auth => ({
-  get_current_time: () => new Date().getTime(),
+const typingWorker = (state: GlobalState) => {
+  const auth: Auth = getAuth(state);
+  const serverVersion: ZulipVersion | void = getServerVersion(state);
 
-  notify_server_start: (user_ids_array: number[]) => {
-    api.typing(auth, JSON.stringify(user_ids_array), 'start');
-  },
+  // User ID arrays are only supported in server versions >= 2.0.0-rc1
+  // (zulip/zulip@2f634f8c0). For versions before this, email arrays
+  // are used. If current server version is undetermined, user ID
+  // arrays are optimistically used.
+  let useEmailArrays: boolean;
+  if (serverVersion === undefined) {
+    useEmailArrays = false;
+  } else {
+    useEmailArrays = !serverVersion.isAtLeast(new ZulipVersion('2.0.0-rc1'));
+  }
 
-  notify_server_stop: (user_ids_array: number[]) => {
-    api.typing(auth, JSON.stringify(user_ids_array), 'stop');
-  },
-});
+  const getRecipients = user_ids_array => {
+    if (useEmailArrays) {
+      return JSON.stringify(user_ids_array.map(userId => getUserForId(state, userId).email));
+    }
+    return JSON.stringify(user_ids_array);
+  };
+
+  return {
+    get_current_time: () => new Date().getTime(),
+
+    notify_server_start: (user_ids_array: number[]) => {
+      api.typing(auth, getRecipients(user_ids_array), 'start');
+    },
+
+    notify_server_stop: (user_ids_array: number[]) => {
+      api.typing(auth, getRecipients(user_ids_array), 'stop');
+    },
+  };
+};
 
 export const sendTypingStart = (narrow: Narrow) => async (
   dispatch: Dispatch,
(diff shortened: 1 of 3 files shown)


agrawal-d:
@gnprice @ray-kraesig please review.

src/users/usersActions.js
```diff
@@ -25,17 +26,41 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
...
+const typingWorker = (state: GlobalState) => {
+  const auth: Auth = getAuth(state);
 
-  notify_server_start: (user_ids_array: number[]) => {
-    api.typing(auth, JSON.stringify(user_ids_array), 'start');
-  },
+  const getRecipients = user_ids_array => {
+    // User ID arrays are only supported in server versions >= 2.0.0-rc1, for
+    // versions before this, email arrays are used. If current server version is
+    // undetermined, user ID arrays are optimistically used.
```
ray-kraesig, line 35: This is... probably the right decision. The server started providing its version in `1.5.0-547-gd9b10727fa`, in early 2017; it's more likely that we failed to parse a new, exotic version-string format than that someone's actually trying to connect to a Zulip server that doesn't provide one.

It may be worth making a note of that reasoning somewhere, though. (Or of yours, if it was different.)

src/users/usersActions.js
```diff
@@ -25,17 +26,41 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
...
+    let useEmailArrays: boolean;
+    const activeAccount: Account | void = tryGetActiveAccount(state);
 
-  notify_server_stop: (user_ids_array: number[]) => {
-    api.typing(auth, JSON.stringify(user_ids_array), 'stop');
-  },
-});
+    if (!activeAccount || activeAccount.zulipVersion === undefined) {
```
ray-kraesig, line 39: `activeAccount?.zulipVersion === undefined`, I think.

src/users/usersActions.js
```diff
@@ -25,17 +26,41 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
...
+      useEmailArrays = false;
+    } else {
+      useEmailArrays = !new ZulipVersion(activeAccount.zulipVersion).isAtLeast(
+        new ZulipVersion('2.0.0-rc1'),
```
ray-kraesig, line 43: This could use a comment referencing the actual server commit in which this capability was added.

src/users/usersActions.js
```diff
@@ -25,17 +26,41 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
...
+      );
+    }
+    if (useEmailArrays) {
```
ray-kraesig, line 46: `useEmailArrays` doesn't need to be recomputed every time either of `notify_server_{start,stop}` is called; it should be constant for a given `typingWorker`, just as `auth` is.

(Admittedly, under our current implementation, that won't save much effort.)

src/users/usersActions.js
agrawal-d: If I use `?.`, flow complains below because `activeAccount` is `Account | void`, i.e I can't use `activeAccount.zulipVersion` then.

src/users/usersActions.js
agrawal-d: That's 2f634f8c06eaece269b9a2692c8ec8edc5af6ac5

agrawal-d:
I tested this in CZO, and it worked -the app sends user ID arrays, and the typing notifications work.
I tried to test this on a local server before commit 2f634f8, but the server wasn't starting up due to some configuration errors. I locally changed the server version to `1.0` in the app, and we do send email arrays then, so the logic is working.

src/users/usersActions.js
ray-kraesig: Well, that's an interesting bug! Not much to be done about it for now.

For future reference, it appears to specifically be `github.com/facebook/flow/issues/6613`, which was fixed in v0.114.0. (At time of writing, the latest version of React Native is v0.62.2, using Flow v0.113.0.)

src/users/usersActions.js
ray-kraesig: (For later readers' reference, this is `1.9.0-1173-g2f634f8c06`, or equivalently, `2.0.0-rc1~476`.)

src/users/usersActions.js
```diff
@@ -25,17 +26,42 @@ export const reportPresence = (isActive: boolean = true, newUserInput: boolean =
...
+const typingWorker = (state: GlobalState) => {
+  const auth: Auth = getAuth(state);
 
-  notify_server_start: (user_ids_array: number[]) => {
-    api.typing(auth, JSON.stringify(user_ids_array), 'start');
-  },
+  // User ID arrays are only supported in server versions >= 2.0.0-rc1, for
+  // versions before this, email arrays are used. If current server version is
+  // undetermined, user ID arrays are optimistically used.
+  // Support for user ID arrays was added in zulip/zulip commit 2f634f8.
```
gnprice, line 35: nit: use 9 hex digits for a commit ID (`git config --global core.abbrev 9` helps make this easy.)

For discussion see https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#mentioning-commits , which I just wrote :stuck_out_tongue: to explain why I've done this, and recommended it, for a long time.

gnprice, line 35: nit: mention the commit ID right next to the version number -- they're two ways of describing the same fact

ray-kraesig:
The only thing I'll add here is –

... *ha!* I literally got the notification for [this commit](https://github.com/zulip/zulip-mobile/commit/83ffe04bcddcd8b3cd8122eca89eb05f57f617e1) as I was typing this sentence. As Greg notes, at least 9 hex digits, both in the commit message and in source comments. (In the commit message, for cross-repository commit references, you can use something along the lines of `zulip/zulip@2f634f8c06`; GitHub renders this as zulip/zulip@2f634f8c06.)

I don't actually recommend the `1.5.0-547-gd9b10727fa` style of reference for most purposes. If you're going to promote that comment to the commit message, I'd suggest something more like `... providing its version as early as zulip/zulip@d9b10727fa (released in 1.6.0, in 2017-06). It's more likely...`

Other than that, LGTM.

----

> I tested this in CZO, and it worked -the app sends user ID arrays, and the typing notifications work.
> I tried to test this on a local server before commit 2f634f8, but the server wasn't starting up due to some configuration errors, (I have limited internet connectivity right now, and cant destroy and re-provision.) I locally changed the server version to `1.0` in the app, and we do send email arrays then, so the logic is working.

I recently attempted to do the same and failed. I'm *not* having internet connectivity issues, as far as I can tell; I fear that whatever cloud-side resources 1.9.x depended on may have vanished. I'll bring it up on CZO.

I don't think this needs to block on that, though. Given that sending email arrays successfully reports typing presence in 2.2-dev, I feel safe in assuming that it'll work on previous versions.

agrawal-d:
@gnprice thanks for the review. I've pushed some changes. Please take a look.

agrawal-d:
You may notice that GitHub UI still truncates the commit hash to 7 digits in the commit message, despite me having used the recommended 9 digits.

ray-kraesig:
> You may notice that the GitHub UI still truncates the commit hash to 7 digits in the commit message, despite me having used the recommended 9 digits.

Unfortunately, yes. The underlying commit message still has the correct number of digits in it, though, as can be seen by hovering on the link to your commit above. (I believe GitHub's linkifier also *uses* all provided digits, even if it doesn't present them.)

My concerns are addressed, and it looks like Greg's have been, too – I'll give him some time to take a second look, but I'm looking forward to having this fix in. Thanks!

gnprice:
Merged -- thanks again both!

I added a couple of commits on top: one a code-style fixup, and one in part a jsdoc-style fixup and in part just adding some information to the jsdoc that we hadn't quite written down yet. I also slightly adjusted the commit messages' summary lines.

chrisbobbe:
> adding some information to the jsdoc that we hadn't quite written down yet

Thanks, @gnprice!

PR #4029, opened 2020-04-08
webview build: Build to an intermediate staging directory for Android.

### LINKED ISSUE #4018: webview assets are missing on first build, on Android
Quick repro:
```shell-session
$ rm -rf android/app/build
$ android/gradlew -p android -q :app:assembleDebug \
    && unzip -l android/app/build/outputs/apk/debug/app-debug.apk \
        | grep css$
```

This should print:
```
     8963  1980-00-00 00:00   assets/webview/base.css
    22113  1980-00-00 00:00   assets/webview/katex/katex.min.css
```

Instead it prints nothing. (Other than a line of Gradle spam from one of our dependencies.)

If you run the build a second time (without deleting the build directory again first), it works fine.

The effect in the app is quite conspicuous. If you instead do
```shell-session
$ rm -rf android/app/build
$ react-native run-android
```

and then go to a message list, it looks very broken.

Same effect if you do a release build with a fresh build directory. This was the cause of the original 26.24.147 build [being broken](https://github.com/zulip/zulip-mobile/releases/tag/26.24.147).

The issue reproduces back to 26.13.136, the first version that had the `base.css` mechanism:
```shell-session
$ git checkout 26.13.136
$ yarn \
  && rm -rf android/app/build \
  && android/gradlew -p android -q :app:assembleDebug \
  && unzip -l android/app/build/outputs/apk/debug/app-debug.apk \
     | grep css$
```

ray-kraesig:
... instead of stepping on the AGP's toes and hoping its dance card is
empty tonight.

The assets-merge task provided by the Android Gradle Plugin builds to
an undocumented output directory. Unfortunately, we'd been writing to
that same destination directory; this is not safe if done before the
assets-merge task runs.

Up until now, we hadn't noticed because the assets-merge had normally
been a no-op for us -- we have no asset directories declared in our
Android Studio project, so nothing for the merge task to merge. On the
first run, though, it's _not_ a no-op: it'll wipe the output directory
as it sets up the other directories in ${buildDir}/intermediates/,
even though it does nothing thereto thereafter.

We'd originally been writing directly to that directory in mimicry of
react-native-vector-icons, which does the same; but r-n-v-i does some
additional violence to the task-dependency graph in order to run
*after* the assets-merge task, rather than before it.

We could follow its lead there, but it's much simpler for us just to
work with the AGP's assets-merge system rather than work around it. We
instead build to a temporary asset-directory dynamically added in the
Gradle project.

Fixes #4018.


ray-kraesig:
Tagging as high-priority, to match the relevant issue.

A bit of hackery used to help diagnose this issue may be of interest, and is available [as a gist](https://gist.github.com/ray-kraesig/54420b865fa852b306e4b844cf05ce10).

ray-kraesig:
(The newly-pushed commit has some minor editing done to the commit message and the PR's OP. There are no changes to the diff compared to the original version.)

gnprice:
Great -- thanks for this debugging!

I'm curious to learn more about some of these pieces; I'll make a chat thread.

ray-kraesig:
Updated, incorporating most of the suggestions from CZO.

gnprice:
Thanks for the revisions! And I tested this version too, and it WFM -- after `rm -rf android/app/build/` and a fresh build, I get:
```
$ unzip -l android/app/build/outputs/apk/release/app-release.apk | grep css$
     8963  1980-00-00 00:00   assets/webview/base.css
    22113  1980-00-00 00:00   assets/webview/katex/katex.min.css
```

Merging.

PR #4031, opened 2020-04-09
Add proper typing to Outbox reducer tests

ray-kraesig:
Rewrite `outboxReducer-test.js` tests to be well-typed.

This incidentally fixes a few tests which weren't actually testing what they said they were testing. (Fortunately, fixing the tests didn't require changes to the code being tested.) Otherwise, this patchset should involve no interesting changes to the tests.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -5,7 +5,12 @@ import { createStore } from 'redux';
 import type { CrossRealmBot, Message, PmRecipientUser, Stream, User } from '../../api/modelTypes';
 import type { Action, GlobalState, RealmState } from '../../reduxTypes';
 import type { Auth, Account, Outbox } from '../../types';
-import { ACCOUNT_SWITCH, LOGIN_SUCCESS, REALM_INIT } from '../../actionConstants';
+import {
+  ACCOUNT_SWITCH,
+  LOGIN_SUCCESS,
+  REALM_INIT,
+  EVENT_NEW_MESSAGE,
+} from '../../actionConstants';
 import rootReducer from '../../boot/reducers';
 import { authOfAccount } from '../../account/accountMisc';
 
@@ -463,3 +468,17 @@ export const action = deepFreeze({
 // which one has which specific type.)
 /* eslint-disable-next-line no-unused-expressions */
 (action: { [string]: Action });
+
+/* ========================================================================
+ * Action fragments
+ *
+ * Partial actions, for those action types whose interior will almost always
+ * need to be filled in with more data.
+ */
+
+export const eventNewMessageActionBase /* \: $Diff<EventNewMessageAction, {| message: Message |}> */ = {
+  type: EVENT_NEW_MESSAGE,
+  id: 1001,
+  caughtUp: {},
+  ownEmail: selfAccount.email,
+};
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -9,6 +9,12 @@ import { ACCOUNT_SWITCH, LOGIN_SUCCESS, REALM_INIT } from '../../actionConstants
 import rootReducer from '../../boot/reducers';
 import { authOfAccount } from '../../account/accountMisc';
 
+/* ========================================================================
+ * Utilities
+ *
+ * Generators for primitive data types.
+ */
+
 /**
  * Generate a new (nonzero) timestamp, suitable for many uninteresting purposes.
  *
@@ -54,6 +60,10 @@ const makeUniqueRandInt = (itemsType: string, end: number): (() => number) => {
 
 /** Return a string that's almost surely different every time. */
 export const randString = () => randInt(2 ** 54).toString(36);
+
+/* ========================================================================
+ * Users and bots
+ */
 
 const randUserId: () => number = makeUniqueRandInt('user IDs', 10000);
 const userOrBotProperties = ({ name: _name }) => {
@@ -139,6 +149,10 @@ export const otherUser: User = makeUser({ name: 'other' });
 
 export const crossRealmBot: CrossRealmBot = makeCrossRealmBot({ name: 'bot' });
 
+/* ========================================================================
+ * Streams
+ */
+
 const randStreamId: () => number = makeUniqueRandInt('stream IDs', 1000);
 export const makeStream = (args: { name?: string, description?: string } = {}): Stream => {
   const name = args.name ?? randString();
@@ -157,6 +171,10 @@ export const stream: Stream = makeStream({
   name: 'a stream',
   description: 'An example stream.',
 });
+
+/* ========================================================================
+ * Messages
+ */
 
 const displayRecipientFromUser = (user: User): PmRecipientUser => {
   const { email, full_name, user_id: id } = user;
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Re: `test data [nfc]: Rename \`action\` to \`actions\`` — just a heads up that https://github.com/zulip/zulip-mobile/commit/781af12098fa5acf8093f7c9dbe2b07959ff9817 just landed, and it uses `eg.action` instead of `eg.actions`.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -61,7 +62,7 @@ describe('sessionReducer', () => {
   });
 
   test('LOGIN_SUCCESS', () => {
-    const newState = sessionReducer(baseState, eg.action.login_success);
+    const newState = sessionReducer(baseState, actions.login_success);
```
gnprice, line 65: The object has many actions, but each one is a single example action. The name is chosen so that an example action of type `LOGIN_SUCCESS` is `eg.action.login_success`.

The pattern `eg.action.foo` is what appears all over the tests, so that's what the name is optimized for; we don't really ever use it as a collection.

src/outbox/__tests__/outboxReducer-test.js
```diff
@@ -83,59 +66,48 @@ describe('outboxReducer', () => {
...
-      ]);
+    test('remove message if local_message_id matches', () => {
+      const message1 = makeOutboxMessage({ timestamp: 546 });
+      const message2 = makeOutboxMessage({ timestamp: 150238512430 });
+      const message3 = makeOutboxMessage({ timestamp: 150238594540 });
+      const initialState = deepFreeze([message1, message2, message3]);
 
       const action = deepFreeze({
-        type: EVENT_NEW_MESSAGE,
+        ...eventNewMessageActionBase,
```
gnprice, line 89: similarly, keep the names `eg.foo`

src/__tests__/lib/exampleData.js
```diff
@@ -3,12 +3,28 @@ import deepFreeze from 'deep-freeze';
...
+ * Generators for primitive data types.
+ */
+
+/**
+ * Generate a new (nonzero) timestamp, suitable for many uninteresting purposes.
+ *
+ * Counts up approximately forever in increments of 1000.
+ */
+const makeTime: () => number = (() => {
+  const startTime = 10 ** 13; // 2286 Nov 20
```
gnprice, line 23: ISO dates, please :wink:

src/__tests__/lib/exampleData.js
```diff
@@ -244,6 +272,56 @@ export const streamMessage = (extra?: $Rest<Message, {}>): Message => {
...
+  // In real data, the .id and .timestamp of an Outbox object are identical.
+  // Complain if a test tries to violate that constraint.
+  const { id, timestamp } = data;
+  if (id !== undefined && timestamp !== undefined && id !== timestamp) {
+    throw new Error(
+      `outbox messages should have identical .id ${id} and .timestamp (${timestamp})`,
+    );
+  }
+
+  const outputTimestamp = id ?? timestamp ?? makeTime() / 1000;
```
gnprice, line 312: Given this, there's no use case for the test to choose one or the other to set. Keep the contract and the implementation simpler by just disallowing `id`.

src/__tests__/lib/exampleData.js
```diff
@@ -244,6 +272,56 @@ export const streamMessage = (extra?: $Rest<Message, {}>): Message => {
...
+ *
+ * (Only stream messages for now. Feel free to add PMs, if you need them.)
+ */
+
+/** An outbox message with no interesting data. */
+const outboxMessageBase: $Diff<Outbox, {| id: mixed, timestamp: mixed |}> = deepFreeze({
+  isOutbox: true,
+  isSent: false,
+
+  avatar_url: 'https://example.com/avatar.png',
```
gnprice, line 286: should match the other sender properties

src/__tests__/lib/exampleData.js
```diff
@@ -244,6 +272,56 @@ export const streamMessage = (extra?: $Rest<Message, {}>): Message => {
...
+  content: '<p>Test.</p>',
+  display_recipient: 'test',
+  // id: ...,
+  markdownContent: 'Test.',
+  narrow: [{ operator: 'stream', operand: 'test' }],
+  reactions: [],
+
```
gnprice, line 293: does this blank line signify something? I'm missing it. (The other one does, and is helpful.)

src/session/__tests__/sessionReducer-test.js
```diff
@@ -30,7 +31,7 @@ describe('sessionReducer', () => {
       needsInitialFetch: false,
       loading: true,
     });
-    const newState = sessionReducer(state, eg.action.account_switch);
+    const newState = sessionReducer(state, actions.account_switch);
```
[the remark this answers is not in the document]
gnprice: > > we can just re-export it to maintain the "`eg.foo` is an example foo" convention.
> 
> ... this won't quite work. `exampleActions` imports and depends on `exampleData`, and importing the former from the latter would induce a circular dependency.

Mmm, I see. I guess that means when we eventually split this file up, we'll need to break it into a few more pieces. Potentially as simple as: one file for everything but actions, one for the actions, and then a tiny `index.js` to re-export from them both.

> Leaving `exampleData` as monolithic for now.

Sounds good!

src/__tests__/lib/exampleData.js
gnprice: Looks good but I think the implementation comment
```
  // In real data, the .id and .timestamp of an Outbox object are identical.
  // Complain if a test tries to violate that constraint.
```
no longer applies and can be deleted too.

src/outbox/__tests__/outboxReducer-test.js
gnprice: ditto on `makeOutboxMessage` :slightly_smiling_face:

gnprice:
Thanks for the revisions! Looks good except a couple of small comments, in threads above.

One thought on the "action fragments" direction, though I'm fine merging that part as it is in this branch:
```js
/* ========================================================================
 * Action fragments
 *
 * Partial actions, for those action types whose interior will almost always
 * need to be filled in with more data.
 */

export const eventNewMessageActionBase /* \: $Diff<EventNewMessageAction, {| message: Message |}> */ = {
  type: EVENT_NEW_MESSAGE,
  id: 1001,
  caughtUp: {},
  ownEmail: selfAccount.email,
};
```

One way these could be structured, which might help for more convenient typing as we add more of them, is as factory functions rather than skeleton objects. For example the signature could look a lot like `makeStream`:
```js
export const makeStream = (args: { name?: string, description?: string } = {}): Stream => {
```

Not all the properties need to be offered in the parameters -- those two are there because there are computed defaults for them, and in particular the one for `description` depends on the `name`. If a test wants to customize some other property, it does it with a spread like this:
```js
        const collider = { ...eg.makeStream({ name: 'collider' }), stream_id: 311 };
```

One nice thing about that signature for the "action fragments" section is that then we could use a variant of the same trick we use for type-checking `eg.action`:
```js
// Ensure every `eg.action.foo` is some well-typed action.  (We don't simply
// annotate `action` itself, because we want to keep the information of
// which one has which specific type.)
/* eslint-disable-next-line no-unused-expressions */
(action: { [string]: Action });
```
because all such factory functions will have types that are subtypes of, not quite `Action` itself, but `(...empty) => Action`.

ray-kraesig:
> Thanks for the revisions! Looks good except a couple of small comments, in threads above.

Done!

----

> One way these could be structured, which might help for more convenient typing as we add more of them, is as factory functions rather than skeleton objects. [...]

This is fine, as far as it goes; but...

> One nice thing about that signature for the "action fragments" section is that then we could use a variant of the same trick we use for type-checking `eg.action`:

... that's not a "nice thing", in my book. What you're calling a trick, I would call a *hack* – a workaround for the fact that we don't have access to the actual `FooAction` types here. I would very much prefer not to extend it.

gnprice:
Thanks -- merged!


> What you're calling a trick, I would call a _hack_ – a workaround for the fact that we don't have access to the actual `FooAction` types here. I would very much prefer not to extend it.

OK -- we can cross that bridge when we come to it.

PR #4034, opened 2020-04-10
[REVIEWABLE] Sign in with Apple on iOS

chrisbobbe:
This is a rough, incomplete draft of this feature (focused on iOS) intended to check my understanding of what needs to be done, broadly. Even if I've got some things wrong, I think I've reached a point where it would be helpful to test with a server-side implementation, using a key that @gnprice has.

I have several questions, some larger, some smaller. How about I write a separate message for each of them, here, and if it's big enough for a lot of discussion, we can go to CZO?

Fixes: #3964


chrisbobbe:
After we [decided](https://chat.zulip.org/#narrow/stream/107-kandra/topic/Apple.20auth/near/850089) to use a custom button everywhere, and never to use the system-provided button ([Apple's HIG document on these](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/)), I haven't yet done the work to comply with the official rules for custom buttons. Part of me wishes this were acceptable, since the code makes it so easy for it to be uniform with the other social auth buttons, which (IMHO) looks better than having it be different, anyway. For the real version, it looks like the biggest differences we'll have to make are:
  - to use the SVG Apple provides, instead of the "apple-logo" SVG icon from [`Ionicons`](https://ionicons.com/).
  - to match the height of the logo file to the height of the button
  - to not use our Zulip green color on the logo, or any custom color

It looks so nice with the others, now, though!
Screen Shot 2020-04-10 at 1 46 55 PM

chrisbobbe:
You can pass a "state" to the authorization request to Apple, which is "Data that’s returned to you unmodified in the corresponding credential after a successful authentication." ([Apple doc](https://developer.apple.com/documentation/authenticationservices/asauthorizationopenidrequest/3153063-state?language=objc); `state` is a property on an [ASAuthorizationOpenIDRequest](https://developer.apple.com/documentation/authenticationservices/asauthorizationopenidrequest?language=objc), which [ASAuthorizationAppleIDRequest
](https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidrequest?language=objc) extends, and [here](https://github.com/expo/expo/blob/master/packages/expo-apple-authentication/ios/EXAppleAuthentication/EXAppleAuthenticationRequest.m#L31)'s the line where Expo uses an ASAuthorizationAppleIDRequest, also, [Expo doc](https://docs.expo.io/versions/latest/sdk/apple-authentication/#appleauthenticationsigninasyncoptions)). It's optional. Expo says it's useful to help prevent replay attacks. Would it be appropriate to use an OTP from our existing OTP generator, perhaps? 🤷‍♂️

chrisbobbe:
There's also a `nonce` property we can apparently use, which Expo exposes, and I'm having trouble distinguishing it from what `state` does. In the first of [two](https://docs.expo.io/versions/latest/sdk/apple-authentication/#arguments) [attempts](https://docs.expo.io/versions/latest/sdk/apple-authentication/#appleauthenticationappleauthenticationsigninoptions), Expo is, surprisingly, more helpful than the [Apple doc](https://developer.apple.com/documentation/authenticationservices/asauthorizationopenidrequest/3362520-nonce?language=objc) about this property; Expo links to the [OpenID Connect specification](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowSteps). There may be a standard, correct answer for what to do with this, but that spec says it's optional, so I'm appealing to others' judgment on this.

chrisbobbe:
On purpose, in this PR, I haven't made any code changes to the web flow logic, with a glass-half-full mindset (and some thinking through) that it will Just Work, since that's what the existing web auth flow is designed for. The web flow will always be used on Android, and always on iOS versions before 13. For iOS >= 13, it will be used for non-Kandra realms.

chrisbobbe:
As noted in a code comment, I think we'll need to carefully handle errors from the `fetch` of `${this.props.realm}/complete/apple/?mobile_flow_otp=${nativeAppleOTP}`. I have it in my notes from a conversation with @gnprice that, when using the native flow, it would be nice to open Zulip's "create account" flow in the browser if it turns out that the user hasn't created a Zulip account yet. The mobile app doesn't have to be in charge of this for any of the existing auth flows (I think maybe the purely web-auth ones handle it this way, though?).

chrisbobbe:
We can use `fetch`, right, or does [this limitation](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/view.20uploaded.20files/near/849454) apply?

chrisbobbe:
I'm looking at [this doc](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms) for this question. For the body of the POST request to `/complete/apple/` (which we make after successful Apple auth), I'm trying to exactly replicate the shape of the request that Apple will send there, in the web flow. In the web flow, we get to tell Apple what we'd like their request to look like, with a `response_mode` flag, which can be `query`, `fragment`, or `form_post`. I'm having trouble finding where we do this in https://github.com/zulip/zulip/pull/14235, but I expect it will be `form_post`, because the docs say, "If you requested any scopes, the value must be form_post", and I think we'll probably at least want the email scope, maybe full name too. Then,

> For the `form_post` value, an HTTP POST request containing the results of the authorization is sent to the redirectURI. The HTTP body contains the result parameters with application/x-www-form-urlencoded content type.

Thus, the body I've chosen to use. Do I seem to be on the right track here?

chrisbobbe:
I haven't thought much about the "Prevent Duplicate Accounts" strategy outlined at [this Apple doc](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple). It doesn't seem to block a basic implementation, but it might be nice to investigate.

chrisbobbe:
Apple has [documentation](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/verifying_a_user) (at "Manage the User Session") on, I think, a way to kick someone out of a Zulip account that they logged into with Apple, upon discovering that their account is no longer authenticated at the device level. It doesn't sound like we do something similar with Google on Android, but some users might feel more comfortable if this were done?

chrisbobbe:
Just rebased and fixed some merge conflicts.

chdinesh1089:
> I'm having trouble finding where we do this in zulip/zulip#14235, but I expect it will be form_post, because the docs say, "If you requested any scopes, the value must be form_post", and I think we'll probably at least want the email scope, maybe full name too.

You're right about it. That is taken care by Python-social-auth here https://github.com/python-social-auth/social-core/blob/master/social_core/backends/apple.py#L55
We send the scope as `email name`. We only get name the first time user makes a request, so Only on the initial try to register we get name but we get email every time the user tries to logs in.

> I haven't thought much about the "Prevent Duplicate Accounts" strategy outlined at this Apple doc. Omitting that strategy doesn't seem to block a basic implementation,

I think a user can only make two accounts, one with original email through normal email registration and another with apple auth with a private email. And we get the same private email everytime as apple returns user data based on initial request as said in [Retrieve the User’s Information from Apple ID Servers](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple)
> The server returns the credentials and user information based on the initial request.

Looking at "Prevent Duplicate Accounts" in that doc, If my understanding is right, we send a request to the apple keychain(an apple user uses this to save their login credentials of a website/app in this afaik) credential sharing which helps us detect existing account only if it's credentials with url are saved in keychain. But I'm not really sure about that. I didn't really understand how to implement [`ASAuthorizationPasswordProvider`](https://developer.apple.com/documentation/authenticationservices/asauthorizationpasswordprovider) from their docs.

chrisbobbe:
Thanks, @chdinesh1089! It's helpful for me see that code in python-social-auth. 🙂

I'd better go help make dinner, but I just pushed some changes to add the custom button. Initially, I had thought it would be best to use the custom button for everything, since we had to make one anyway. But actually, I think we can be more targeted than that. In particular, @gnprice, what do you think about

- on Android, using the existing ZulipButton that's already consistent with the other social auth buttons, since Apple's Human Interface Guidelines (which guide what the button should look like) don't apply on Android
- on iOS, using the system-provided button when possible (i.e., iOS >=13), and the custom button otherwise? The custom button is kind of annoying because (at least currently) I don't know how to make the logo, which is a PNG, work well with any of RN's `Touchable` components that have feedback, since they all use opacity, and none seem to apply the opacity properly to the PNG

chrisbobbe:
Here's the official button:

image

And the custom button:

image

I guess one thing that's nice about the custom button is that we're allowed to increase the size of the Apple logo to match the other logos/icons ([doc](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/)):

> Also, the left-aligned logos are available in small, medium, and large sizes, so you can match logo sizes in all the sign-up buttons you display.

src/start/AuthScreen.js
```diff
@@ -220,12 +231,47 @@ class AuthScreen extends PureComponent<Props> {
...
+      `${this.props.realm}/complete/apple/?mobile_flow_otp=${nativeAppleOTP}`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+        body: encodeParamsForUrl(credential),
+      },
+    );
+
```
gnprice, line 255: I don't think this will work as is. I'm not sure we want to be using the `/complete/apple/` route at all.

Here's what Apple will cause to be sent to that route (under "Handle the Response"):
https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms

Here's the data we have at this point, in the `AppleAuthenticationCredential` (from Expo, and corresponding closely to [this](https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidcredential) from Apple):
https://docs.expo.io/versions/latest/sdk/apple-authentication/#appleauthenticationappleauthenticationcredential

Hmm, looking closer, I guess they actually align quite well in what information is present. They just aren't quite the same shape -- so that line `body: encodeParamsForUrl(credential)` will need to be expanded to do a bit more work :wink: 

Then, one thing that simplifies things (and may also call for adding a separate route) is: we don't have a need for the redirect to `zulip://`, or the OTP mechanism which exists to encrypt the API key through that channel. That redirect is needed for getting from a browser back to the app's code. If we're making our own `fetch` here, we can just get the data from that directly.

Alternatively, for the new-account case maybe we do want to open a browser. If we do open a browser, then we need the `zulip://` redirect and the `mobile_flow_otp` mechanism. For simplicity we might, at least at first, just always open a browser, regardless of whether the user might already have an account. In that case we'll want to call something like `webAuth.openBrowser` instead of `fetch`.

chrisbobbe: > For simplicity we might, at least at first, just always open a browser, regardless of whether the user might already have an account. In that case we'll want to call something like `webAuth.openBrowser` instead of `fetch`.

Mmm, this does make sense as a place to start; I was [a bit worried](https://github.com/zulip/zulip-mobile/pull/4034#issuecomment-612268425) about how to handle basically any case other than the completely successful login path, with the browserless `fetch` way. All those contingencies (no Zulip account, etc.) could be handled in the browser. We would indeed need the `zulip://` redirect, but I'm not sure I've understood whether you've decided to add the separate route and use it here.

Using the browser for for everyone is not ideal for the long-term, though, right; users with an existing account would want to log in seamlessly, without the browser. I'd basically want to open an issue for this immediately upon merging this simpler approach; does that sound right? (And I hope opening the browser for people with existing accounts won't be enough to trigger an App Store review rejection.) One way this more complicated, but smoother, process could look is

1. We use `fetch` on either a) a route that never gives the `zulip://` redirect, since, as you've said, we don't need it (in this case, adding a route is necessary), or b) a route that we can pass a flag to, saying we don't want the redirect (here, it would be possible, but maybe awkward, to use an altered form of the existing route).
2. Then, the result would either contain an API key, and we'd be ready to go, or (e.g., if there's no Zulip account) a signal that we need to try again, this time with `webAuth.openBrowser`.

gnprice: > Using the browser for everyone is not ideal for the long-term, though, right; users with an existing account would want to log in seamlessly, without the browser. I'd basically want to open an issue for this immediately upon merging this simpler approach; does that sound right?

Yeah, that sounds right.


> (And I hope opening the browser for people with existing accounts won't be enough to trigger an App Store review rejection.)

Yeah, me too. Good thing we're not right up against the deadline.


> One way this more complicated, but smoother, process could look is [...]

This sounds like a good plan. The `mobile_flow_otp` option already serves as a flag that controls whether it does the `zulip://` redirect, vs. redirecting to the webapp (in addition to providing a one-time pad to use in the `zulip://` redirect.) So I think it'd probably make sense to go for (b) and just add another flag with a meaning like "just return the API key in a JSON response". But @mateuszmandera may have more to say about what'd be a convenient arrangement on the server side.

mateuszmandera: Hmm, so the plan would be to do the browser flow only for now?

For the native approach - I think `/complete/apple/` is still a good endpoint to use for this as on the server side we can override the class method handling the endpoint to be able to deal with the "native flow" request as well. The upside of that being that then the rest of the pipeline for processing "social authentication" can run as normal (avoiding code duplication). And then we'd have two cases for the response you'll get from the server:
1. Complete success - you get the `zulip://` redirect (yes, it's kind of redundant to do the OTP thing, but on the other hand - code reuse is nice)
2. Some kind of failure - you get a https redirect. If you get this, you could open the redirect in the browser and that will handle the error. "Bad" errors will get the same handling as in browser flow (show an error message if we're currently doing that in the browser flow). And the "user doesn't have an account" will take the user through the registration flow in the browser and upon success redirect back to the app with the newly created account through the `zulip://` redirect.

How does the above sound? @gnprice  @chrisbobbe

gnprice: > Hmm, so the plan would be to do the browser flow only for now?

Yes but there's a couple of things that can mean, so to be explicit:
* For users on Android and on iOS <=12, we'll do the whole auth flow in a browser, same way we do Google and GitHub and SAML and the rest.
* For users on iOS >=13 where we can, we'll
  * use the system-provided flow to get a credential from the user -- so the user gets some system UI that says like "do you want to sign into this app with your identity, what name do you want, do you want to use your real email address or an opaque alias", and when they say yes we get one of these [`AppleAuthenticationCredential`](https://docs.expo.io/versions/latest/sdk/apple-authentication/#appleauthenticationappleauthenticationcredential) objects;
  * then we'll turn that into a URL like `https://example.zulipchat.com/complete/apple?id_token=…&mobile_flow_otp=…&…` and open a browser with that URL, and pick up the browser flow from there. In particular if the user already has an account and the server accepts the login, it'll redirect the browser to a `zulip://…` URL with the encrypted API key.


> Some kind of failure - you get a https redirect. If you get this, you could open the redirect in the browser and that will handle the error. "Bad" errors will get the same handling as in browser flow (show an error message if we're currently doing that in the browser flow).

Hmm. A couple of questions/thoughts I'd have are:
* Will the URL in the redirect be enough to just pass to a fresh browser and it'll pick up the flow smoothly from there? No cookies or other state needed?
* That seems like it wouldn't cleanly support the app trying to understand anything about the failure and give a structured error of its own -- it'd just always have to go to a browser. That might be OK, but it puts a ceiling on how polished the experience can be with this approach.

In any case I think we can discuss more and experiment when we get there. I would really like to get to a point where we have one flow that works solidly; among other things I think that will help us a lot in concretely understanding how all the steps of this are supposed to work. So I will be very happy for us to just get the flow with `/complete/apple/` in a browser working.

src/start/AuthScreen.js
```diff
@@ -220,12 +231,47 @@ class AuthScreen extends PureComponent<Props> {
...
   };
 
-  handleAuth = (method: AuthenticationMethodDetails) => {
+  handleNativeAppleAuth = async () => {
+    const { dispatch, realm } = this.props;
+    const nativeAppleOTP = await webAuth.generateOtp();
+    const credential = await AppleAuthentication.signInAsync({
+      state: nativeAppleOTP,
+    });
+    if (credential.state !== nativeAppleOTP) {
```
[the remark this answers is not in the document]
chrisbobbe: Do you mean rename `webAuth.generateOtp`? It's currently got a useful, OTP-specific comment on it:

```
// Generate a one time pad (OTP) which the server XORs the API key with
// in its response to protect against credentials intercept
```

I wonder about extracting the body into a separate function, maybe `generateRandomToken`:

```
  if (Platform.OS === 'android') {
    return new Promise((resolve, reject) => {
      NativeModules.RNSecureRandom.randomBase64(32, (err, result) => {
        if (err) {
          reject(err);
        }
        resolve(base64ToHex(result));
      });
    });
  } else {
    const rand = await NativeModules.UtilManager.randomBase64(32);
    return base64ToHex(rand);
  }
```

chrisbobbe:
OK, just pushed my latest revision, @gnprice. Still a draft, and Flow has some issues with

9812a3c24 [draft] Use "Sign in with Apple".

But just in case you wanted to check in on the code.

chrisbobbe:
Fixed the Flow issues (and filed an Expo issue that should clarify some things for us)!

chrisbobbe:
I just pushed with some rebase conflicts resolved, and removed the edits of lots of translation files, as that'll be handled in Transifex as long as `messages_en.json` is correct.

gnprice:
Tested this just now with Mateusz's test server, as discussed [in chat here](https://chat.zulip.org/#narrow/stream/3-backend/topic/apple.20auth/near/916048). And with one fix, it works!

```diff
diff --git src/start/AuthScreen.js src/start/AuthScreen.js
index a246baed9..af8714691 100644
--- src/start/AuthScreen.js
+++ src/start/AuthScreen.js
@@ -245,6 +245,7 @@ class AuthScreen extends PureComponent<Props> {
     const state = await webAuth.generateRandomToken();
     const credential: AppleAuthenticationCredential = await AppleAuthentication.signInAsync({
       state,
+      requestedScopes: [AppleAuthentication.AppleAuthenticationScope.FULL_NAME, AppleAuthentication.AppleAuthenticationScope.EMAIL],
     });

**Comment by gnprice** [comment]
Specifically with that change it worked on a simulated iPhone 11 running iOS 13.1.

It also worked, even without that change, on iOS 12 -- on a simulated iPhone 7 running iOS 12.2. As expected, on that OS version the "Sign in with Apple" button went straight to a browser view; I went through the auth flow on appleid.apple.com, got redirected back to the Zulip server, and then the app was successfully logged in.

**Comment by chrisbobbe** [comment]
Probably worth a try on Android, as well; this PR is also meant to fix #3965.

File: ios/ZulipMobile/ZulipMobile.entitlements [hunk 3]
```diff
@@ -4,5 +4,9 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>com.apple.developer.applesignin</key>
+	<array>
+		<string>Default</string>
+	</array>
```
gnprice, line 10: What's this bit mean? (Did it come from some instructions somewhere?)

src/start/AuthScreen.js
```diff
@@ -251,16 +293,24 @@ class AuthScreen extends PureComponent<Props> {
...
+                key={auth.name}
+                style={styles.halfMarginTop}
+                onPress={() => this.handleAuth(auth)}
+              />
+            ) : (
+              <ZulipButton
+                key={auth.name}
+                style={styles.halfMarginTop}
+                secondary
+                text={`Sign in with ${auth.displayName}`}
```
gnprice, line 308: Small issue independent of this PR, but I notice it as I look at this code and particularly the set of translated strings at the same time:

This can't work when `auth.displayName` isn't one of the values we know about in advance, so in particular with an open-ended SSO like SAML. We should instead make this a fixed string with a placeholder, and pass the value of `auth.displayName` in through `values`.

src/start/AuthScreen.js
```diff
@@ -1,27 +1,40 @@
...
 import RealmInfo from './RealmInfo';
-import { getFullUrl } from '../utils/url';
+import { getFullUrl, encodeParamsForUrl } from '../utils/url';
 import * as webAuth from './webAuth';
 import { loginSuccess, navigateToDev, navigateToPassword } from '../actions';
+import IosCompliantAppleAuthButton from './IosCompliantAppleAuthButton';
+import openLink from '../utils/openLink';
 
+/* eslint-disable-next-line no-unused-vars */
+const KANDRA_APPLE_KID = 'asdfjkl;'; // TODO, of course (likely won't live here)
```
gnprice, line 37: Discussion [in chat](https://chat.zulip.org/#narrow/stream/3-backend/topic/apple.20auth/near/916095) about what we want this check to be.

For where the relevant constant (or constants) live, whatever they end up being: if we end up finding that we can't straightforwardly ask RN to tell us them based on the app's own metadata, then `src/config.js` is the place we've collected values that feel like this one.

src/api/settings/getServerSettings.js
```diff
@@ -14,14 +14,28 @@ export type AuthenticationMethods = {
...
 |};
 
+export type AppleExternalAuthenticationMethod = {|
+  name: 'apple',
+  apple_kid: string,
+  ...BaseExternalAuthenticationMethod,
+|};
+
+export type OtherExternalAuthenticationMethod = {|
+  name: string, // I'd like this to be "all strings except 'apple'"
```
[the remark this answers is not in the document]
gnprice: I think this is a reasonable wish and there unfortunately isn't a way to fulfill it in Flow's type system. (But I don't blame Flow, because I think it'd be hard to support this without making the type system a lot more complicated.) Here's a previous occasion where I've wanted the same thing:
```
commit f3c6ff71768774834cd3584b832ebb35419bd31c
Date:   Thu Apr 4 15:21:50 2019 -0700

    persist: Write down types on our vendored redux-persist-migrate.
    
    This version involves a couple of compromises:
    
    * The types aren't generic, as the library itself actually is --
      rather they're specialized to our particular `Action` and
      `GlobalState` types.  (And, to play nice with our `Action` type,
      we get the `REHYDRATE` constant from our own `actionConstants`.)
    
      On my first attempt at this (back on 2019-01-08), I tried to type
      this properly generically, in a way that would make sense for the
      upstream library.  I didn't find a way to make it work -- in
      particular, to give an appropriate type for Action.  The type
      should be "either RehydrateAction, or else an object where `type`
      is *not* `REHYDRATE`"... and I think there there may not actually
      be a way in Flow's type system to write that generically,
      i.e. short of listing out every possible value for the `type`
      property.  Using our concrete `Action` type is basically a way of
      doing that.
```

In that instance, we could solve it by actually listing out every possible value for `type`, because ultimately it was all internal to the app. In this instance, that solution isn't available.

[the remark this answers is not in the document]
chrisbobbe: Cool, thanks! I remember coming across that commit message recently and making the connection.

One nice thing about our current strategy of enumerating the Kandra-hosted domains in `src/config.js` is that we don't need to make a separate `AppleExternalAuthenticationMethod` type that has stuff like `apple_kid` or `team_id` or `app_id`. So `ExternalAuthenticationMethod` can just stay how it is.

So this problem just vanishes. 🎉

src/config.js
```diff
@@ -39,6 +40,7 @@ const config: Config = {
     'update_message_flags',
     'user_status',
   ],
+  kandraHostedDomains: ['zulip.com', 'zulipchat.com', 'chat.zulip.org'],
```
gnprice, line 43: Thinking about the name for this variable: I'd like it to be more generically about the relationship of this (build of the) app to some domains, rather than Kandra by name. In principle if someone else makes and distributes their own build with their own App ID, they'd want to edit this to refer to domains they control, even though these would still be the ones hosted by Kandra.

(Even if nobody ends up ever doing that, I think it keeps the logic around this variable clearer to think about if it's named so it works in that general case.)

Perhaps `appAssociatedDomains`? `appOwnDomains`?

src/start/AuthScreen.js
```diff
@@ -227,12 +240,66 @@ class AuthScreen extends PureComponent<Props> {
...
+        // which is basically the URL the user entered on the first
+        // screen, is Kandra-hosted.
+        host = new WhatwgURL(this.props.realm).host;
+      } catch (e) {
+        // `this.props.realm` invalid. TODO: Check this much sooner.
+      }
+      if (config.kandraHostedDomains.includes(host)) {
+        shouldUseNativeAppleFlow = true;
+      }
+    }
```
gnprice, line 295: This is enough code it should probably get moved to its own little helper method. Then call like `} else if (this.shouldUseNativeAppleFlow(method)) {`.

Oh, or even
```
  } else if (method.name === 'apple'
    && this.canUseNativeAppleFlow()) {
```

That's nice because it separates the part that's about this auth method from the more-complicated rest.

src/start/AuthScreen.js
```diff
@@ -227,12 +240,66 @@ class AuthScreen extends PureComponent<Props> {
...
+      let host: string | void;
+      try {
+        // Check that the realm we're actually sending requests to,
+        // which is basically the URL the user entered on the first
+        // screen, is Kandra-hosted.
+        host = new WhatwgURL(this.props.realm).host;
+      } catch (e) {
+        // `this.props.realm` invalid. TODO: Check this much sooner.
+      }
+      if (config.kandraHostedDomains.includes(host)) {
```
gnprice, line 292: This will need to be a bit more complex -- we want to accept any foo.zulip.com or foo.zulipchat.com.

docs/howto/ios-tips.md
```diff
@@ -84,3 +84,87 @@ It seems like there's some caching strategy to avoid fetching
...
+  These will be associated with the Kandra team, so
+  [please](https://chat.zulip.org/#narrow/stream/3-backend/topic/apple.20auth/near/915391)
+  let him know when you're finished with the credentials so he can
+  revoke them. Please don't abuse them with deliberate spam, as
+  that goes on our reputation.
+- Use the domain `zulipdev.com` where Apple asks for a domain;
+  [`localhost` won't
+  work](https://chat.zulip.org/#narrow/stream/3-backend/topic/Apple.20Auth/near/831533).
+  - Add a line like `127.0.0.1 zulipdev.com` to your `/etc/hosts` as
+    described [here](https://zulip.readthedocs.io/en/latest/subsystems/realms.html?highlight=zulipdev.com#working-with-subdomains-in-development-environment). Notice
```
gnprice, line 108: This usually shouldn't be necessary -- on the public Internet, zulipdev.com resolves to 127.0.0.1. It looks like those instructions are for the case where you have something unusual in your network configuration that interferes with that. (It says something about a proxy server... but even that alone shouldn't make this necessary, nor be fixed by this alone.)

docs/howto/ios-tips.md
```diff
@@ -84,3 +84,87 @@ It seems like there's some caching strategy to avoid fetching
...
+    there's already a line mapping 127.0.0.1 to `localhost`; this is what
+    makes the iOS simulator work with `http://localhost:9991`, as
+    described [here](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/dev-server.md#ios-simulator-macos-only), without
+    having to configure `EXTERNAL_HOST`. Likewise, you won't have to
+    set `EXTERNAL_HOST` to get it to work on the simulator.
+  - A physical device won't be aware of the `zulipdev.com` domain;
```
gnprice, line 114: Because the domain resolves on the public Internet to the right address, the physical device will be aware of the domain; the trouble is that it'll mean the wrong thing. 127.0.0.1 is a [loopback](https://en.wikipedia.org/wiki/Localhost) address, pointing at the current machine you're on -- so when you're on a physical device, that's the device itself. (Arguably when you're in the iOS simulator, it ought to mean the simulated device, but empirically it doesn't.)

chrisbobbe:
OK, just pushed my latest changes, including the new way of establishing that a server is Kandra-hosted, as discussed around [here](https://chat.zulip.org/#narrow/stream/3-backend/topic/apple.20auth/near/918790).

jest/jestSetup.js
```diff
@@ -36,3 +36,43 @@ jest.mock('react-native-device-info', () => ({
   getSystemName: jest.fn().mockReturnValue('ios'),
   getSystemVersion: jest.fn().mockReturnValue('13.3.1'),
 }));
+
+jest.mock('expo-apple-authentication', () => ({
+  AppleAuthenticationButton: jest.fn(),
```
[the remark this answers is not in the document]
chrisbobbe: Huh, apparently, this time, everything works fine without adding anything to `transformModulesWhitelist` or mocking it. I wonder what's different.

src/start/AuthScreen.js
chrisbobbe: Ah, right—I should address this in my next revision, or as a separate PR (it didn't make it into the current revision).

ios/ZulipMobile/ZulipMobile.entitlements
chrisbobbe: Yeah, it's step one of https://github.com/expo/expo/tree/1c5eb9818/packages/expo-apple-authentication#setup-ios-project, adding the "Sign in with Apple" entitlement.

src/config.js
chrisbobbe: Sure, makes a lot of sense!

docs/howto/ios-tips.md
chrisbobbe: > on the public Internet, zulipdev.com resolves to 127.0.0.1

Interesting! I did not know this at all...nor did I suspect that it might be a thing you could even do? 🤔

chrisbobbe:
@gnprice, thanks for the review! I've just pushed my latest changes, for another one.

jest/jestSetup.js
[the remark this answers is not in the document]
chrisbobbe: Hmm, and it looks like the errors are back. So, I re-added the mock and the entry in `transformModulesWhitelist` (additionally adding `@unimodules/` there as well, in its own commit).

chrisbobbe:
Sorry, just pushed again—was getting those errors in Jest again; see https://github.com/zulip/zulip-mobile/pull/4034#discussion_r447928183.

ios/ZulipMobile/ZulipMobile.entitlements
gnprice: Great, and I see you've added this info to the relevant commit message.

gnprice:
And merged -- thanks! Added a few small changes.

chrisbobbe:
Great, thanks, and thanks for those changes! I just took a look and they definitely make sense.

PR #4043, opened 2020-04-15
Fix scroll-on-resize to avoid jumping from bottom

gnprice:
When the message list's viewport grows, we generally want to scroll up to keep its bottom edge aligned within the content, but we want to skip that when we're at the very bottom. We have logic to do all that, but it's pretty cryptic; this series starts by unpacking it.

If we're at the bottom and scroll up anyway, the effect would be exactly the symptoms in #3301.

Then it turns out that when we're at the bottom, the `scrollTop` value may not actually quite line up with where you'd think it should be. Empirically:
 * On iOS (in Safari), it tends to be 1px beyond the bottom.
 * On Android (in Chrome), it's not an integer, and can be a fraction of a px either beyond or short of the bottom.

So, allow for that when checking to see if we're at the bottom. We'd been testing for exact numerical equality.

Thanks to @jainkuniya for suggesting a related direction in #4003.

Fixes: #3301

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -167,9 +167,9 @@ const showHideElement = (elementId: string, show: boolean) => {
 let viewportHeight = documentBody.clientHeight;
 
 window.addEventListener('resize', event => {
-  const difference = viewportHeight - documentBody.clientHeight;
+  const heightChange = documentBody.clientHeight - viewportHeight;
   if (documentBody.scrollHeight !== documentBody.scrollTop + documentBody.clientHeight) {
-    window.scrollBy({ left: 0, top: difference });
+    window.scrollBy({ left: 0, top: -heightChange });
   }
   viewportHeight = documentBody.clientHeight;
 });
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -184,8 +184,9 @@ window.addEventListener('resize', event => {
   //
   // If that happened, we'll be at the very bottom now.
   const maxScrollTop = documentBody.scrollHeight - documentBody.clientHeight;
-  if (documentBody.scrollTop === maxScrollTop) {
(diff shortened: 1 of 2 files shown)


jainkuniya:
> On iOS (in Safari), it tends to be 1px beyond the bottom.
On Android (in Chrome), it's not an integer, and can be a fraction of a px either beyond or short of the bottom.

This is interesting.

Tried this version, on IOS it looks great. But on Android there is some jerkiness, when unread notice hides, although no scroll effect (I guess it was in #4003 as well). Next we should eliminate that too.

gnprice:
> But on Android there is some jerkiness, when unread notice hides, although no scroll effect

Yep -- specifically what I see is that when the unread notice disappears, the content jerks up and back down, and lands right back in the right spot.

jainkuniya:
What I thinks going on is (on Android):
as unread notice disappears, it releases space. RN moves messageList component up, and then it realises that now it has more room lets expand MessageList. So it looks like content is slight up for some moment and again back to bottom.

I tried moving Unread Notice to bottom, i.e between MessageList and ComposeBox. Now when unread notice disappears, messageList just moves/shift down.

src/webview/js/js.js
```diff
@@ -184,8 +184,8 @@ window.addEventListener('resize', event => {
   //
   // If that happened, we'll be at the very bottom now.
   const maxScrollTop = documentBody.scrollHeight - documentBody.clientHeight;
-  if (documentBody.scrollTop === maxScrollTop) {
-    // We're at the very bottom of the content.  Don't scroll.
+  if (documentBody.scrollTop >= maxScrollTop - 1) {
+    // We're at (or beyond) the very bottom of the content.  Don't scroll.
```
ray-kraesig, line 188: The "-1" here really needs to have an inline explanation.

PR #4048, opened 2020-04-17
Icon: Fix IconPlus pointing to "plus-circle".

sumukhah:
'IconPlus' is reffering to "plus-circle" icon. But a plane plus
icon is available. So rename current 'IconPlus' to 'IconPlusCircle',
Add plus icon and name it "IconPlus".


ray-kraesig:
Thanks! Merging.

PR #4054, opened 2020-04-18
[WIP] Consume user IDs rather than emails from typing-status events.

### LINKED ISSUE #3930: Consume user IDs rather than emails from typing-status events
This is an instance of #3764. In the `typing` events from the server, which we turn into `EVENT_TYPING_START` and `EVENT_TYPING_STOP` actions, we should switch from looking at the `email` properties (on both `action.recipients` elements and `action.sender`) to looking at the `user_id` properties. Those have been present since ancient server versions; see https://github.com/zulip/zulip/pull/13998#issuecomment-590611227 .

In turn on the server we'd like to stop sending emails here, in order to simplify the API and the code. See zulip/zulip#13998. That's blocked on making this client-side change, and then on an appropriate interval to allow users to update the app.

In order to make this change, I think we'll want to do something like:
* Change `TypingState` (the type of `state.typing`) to use user IDs rather than emails in the keys.
  * This requires no data migration, because we don't persist this part of the state tree; see `src/boot/store.js`.
* Change consumers of `state.typing` to deal appropriately with user IDs rather than emails. This includes:
  * `getCurrentTypingUsers` -- this takes a `Narrow`, which has emails, so it'll need to use `getAllUsersByEmail` to convert to IDs.
  * `clearTypingNotification` -- I think this one may actually not need to change. (Though it looks pretty dubious architecturally; it may be good to cut it out entirely in any case.)

agrawal-d:
I also manually tested that typing notifications are working, and the changes are reflected in the Redux state. Relevant tests are also made well-typed.
Closes #3930.
________

Currently, tests are failing due to (I believe) a Flow bug:

The type TypingState is defined as 
```js
export type TypingState = {|
  [normalizedRecipients: string]: {
    time: number,
    userIds: number[],
  },
|};
```

So, `initalState` should be a valid instance of TypingState:
```js
      const initialState = deepFreeze({
        '1': { time: 123456789, userIds: [1] },
      });

      const newState = typingReducer(initialState, action);
```
But Flow complains:
```
Cannot call `typingReducer` with `initialState` bound to `state` because property `1` is missing in  `TypingState` [1] but exists in  object literal [2].Flow(InferError)
```
That's because Flow is able to infer the key, i.e. '1'. But if I change it to something that Flow can not infer, such as:

```js
      const normalizedRecipients = "1".toString();
      const initialState = deepFreeze({
        [normalizedRecipients]: { time: 123456789, userIds: [1] },
      });

      const newState = typingReducer(initialState, action);
```

Flow no longer complains. So I think this is a bug in Flow. So, should I add `FlowFixMe` in each case of the bug? ( in the file `typingReducer-test.js` )

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -354,7 +354,7 @@ type EventPresenceAction = {|
 
 type EventTypingCommon = {|
   ...ServerEvent,
-  ownEmail: string,
+  ownUserId: number,
   recipients: Array<{
     user_id: number,
     email: string,
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -31,7 +31,7 @@ import {
   EVENT_SUBSCRIPTION,
   EVENT,
 } from '../actionConstants';
-import { getOwnEmail } from '../users/userSelectors';
+import { getOwnEmail, getOwnUserId } from '../users/userSelectors';
 
 const opToActionUser = {
   add: EVENT_USER_ADD,
@@ -151,7 +151,7 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction => {
     case 'typing':
       return {
         ...event,
-        ownEmail: getOwnEmail(state),
+        ownUserId: getOwnUserId(state),
         type: opToActionTyping[event.op],
         time: new Date().getTime(),
       };
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -270,12 +270,12 @@ export type TopicsState = {|
   [number]: Topic[],
 |};
 
-export type TypingState = {|
+export type TypingState = {
   [normalizedRecipients: string]: {
     time: number,
     userIds: number[],
   },
-|};
+};
 
 // These four are fragments of UnreadState; see below.
 export type UnreadStreamsState = StreamUnreadItem[];
--- a/src/typing/__tests__/typingReducer-test.js
+++ b/src/typing/__tests__/typingReducer-test.js
@@ -1,3 +1,5 @@
+/* @flow strict-local */
+
 import deepFreeze from 'deep-freeze';
 
 import { EVENT_TYPING_START, EVENT_TYPING_STOP } from '../../actionConstants';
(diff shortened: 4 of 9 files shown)


agrawal-d:
@gnprice @ray-kraesig please review.

agrawal-d:
Bumping priority to match the issue.

ray-kraesig:
This appears to be the specific Flow bug discussed in `src/utils/jsonable.js`. The usual workaround is to make the relevant type (`TypingState`) inexact.

src/actionCreator.js
```diff
@@ -1,17 +0,0 @@
...
-import { EVENT_TYPING_START } from './actionConstants';
-
-export default (dispatch: Dispatch, actions: $ReadOnlyArray<Action>, state: GlobalState) => {
-  actions.forEach(action => {
-    switch (action.type) {
-      case EVENT_TYPING_START:
-        if (Object.keys(state.typing).length === 0) {
-          dispatch(clearTypingNotification());
-        }
-        break;
```
gnprice, line 13: So, it might be perfectly right to delete this, but it'll take a bit more reasoning than that :slightly_smiling_face: 

The commit message says:

    typing: Delete file 'actionCreator.js'.
    
    This file does not appear to do anything useful. So, delete it
    and remove references to it from all the places where it's called.

It sure looks like it's doing *something*, though! It's causing us to invoke `clearTypingNotification` in certain circumstances.

Is that not actually happening? Or, is it not useful to be doing that? In either case: why? :wink:

agrawal-d:
@ray-kraesig @gnprice, thanks for the review!
I've made some changes. Please review them.

src/actionCreator.js
gnprice: Cool, sounds good. Thanks for drawing that connection!

gnprice:
One nit in the commit messages:

    typing: Create helpers to compute normalized recipients as user IDs.

Just "typing" is ambiguous -- it sounds like it's about static types :slightly_smiling_face:. I've generally written "typing status" for this.

I'll make that tweak and then merge. Thanks again!

Issue #4056, opened 2020-04-20
Sharing messages that contain '>', '<', and '&' results in '&lt;', '&gt;' and  '&amp;'


Maskedman99:
![Screenshot-20200420192019-774x94](https://user-images.githubusercontent.com/31368194/79759378-3b09dd00-833c-11ea-9343-796743a949f0.png)

Maskedman99:
Further ares where this edge-case occurs

![Screenshot-20200420193648-714x64](https://user-images.githubusercontent.com/31368194/79761259-d13f0280-833e-11ea-8976-83fc9ee2add1.png)
This is the same case in webapp,

gnprice:
Hmm, yeah, that'd be nice to fix.

The reason is that the text we share is meant to be a "plain text" version of the message's content. For your example message, the `content` property (which is HTML) will be `"<p>vue &lt; react-native &gt; flutter &quot; ionic</p>"`. And here's the "share" implementation:
```js
const shareMessage = ({ message }) => {
  Share.share({
    message: message.content.replace(/<(?:.|\n)*?>/gm, ''),
  });
};
shareMessage.title = 'Share';
shareMessage.errorMessage = 'Failed to share message';
```

So it just takes the HTML and strips all tags.

We can probably improve that. The first step would be to extract that expression `message.content.replace(/<(?:.|\n)*?>/gm, '')` as a function and then add some tests. Then we can add things like translating HTML entities (like `&lt;`) -- along with careful tests.

ray-kraesig:
> We can probably improve that. The first step would be to extract that expression `message.content.replace(/<(?:.|\n)*?>/gm, '')` as a function and then add some tests. Then we can add things like translating HTML entities (like `&lt;`) -- along with careful tests.

Are you sure? I think just using the Markdown source (`api.getRawMessageContent`) might be a better option, even if you're not sharing to a Markdown-capable target. Simplicity-of-implementation aside, most people understand \*emphasis\* markers.

gnprice:
> I think just using the Markdown source (`api.getRawMessageContent`) might be a better option, even if you're not sharing to a Markdown-capable target.

I don't have a real strong view on this product/UX choice. I think people's preferences may differ depending on what else is in the message beyond `*`, though. For a detailed discussion the right place would probably be #mobile... but also I'd rather not spend a lot of energy discussing this product choice in the abstract -- we have plenty of issues to work on which we know a lot of people who use the app care about.

[...]

PR #4060, opened 2020-04-21
message-share: Add tests for the messages being shared

Maskedman99:
This PR moves the codes to remove html tags from the message that is to be shared, into a function ```messageFilter()```, and adds tests for it.

See Issue #4056


src/message/messageActionSheet.js
```diff
@@ -169,9 +169,12 @@ const unstarMessage = async ({ auth, message }) => {
 unstarMessage.title = 'Unstar message';
 unstarMessage.errorMessage = 'Failed to unstar message';
 
+export const messageFilter = (messageContent: string): string =>
```
ray-kraesig, line 173: This name is much too generic for this context, and it doesn't actually describe this function's purpose, anyway. (This is a "filter" for the *content* of messages, rather than for the messages themselves; and in the following commit it becomes more than a filter.)

src/message/messageActionSheet.js
```diff
@@ -169,8 +169,13 @@ const unstarMessage = async ({ auth, message }) => {
...
 unstarMessage.errorMessage = 'Failed to unstar message';
 
+/** Private; exported only for tests. */
 export const messageFilter = (messageContent: string): string =>
-  messageContent.replace(/<(?:.|\n)*?>/gm, '');
+  messageContent
+    .replace(/<(?:.|\n)*?>/gm, '')
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>');
```
ray-kraesig, line 178: I suspect (but have not confirmed) that there will also need to be clauses for `&quot;` and `&#39;` here.

Also, this deserves comments.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -230,4 +230,10 @@ describe('messageFilter', () => {
     expect(messageFilter('<p>😄John <a></br>Doe</p>')).toEqual('😄John Doe');
     expect(messageFilter('1<div clas="a">2<p>3<a href="a" /></br>4</p>5')).toEqual('12345');
   });
+
+  test("'&amp;','&lt;' and '&gt' should be replaced with '&', '<' and '>'", () => {
+    expect(messageFilter('&lt;&gt;&amp;')).toEqual('<>&');
+    expect(messageFilter('&amp;&amp;amp;')).toEqual('&&amp;');
+    expect(messageFilter('<p>&lt;p&gt;</p>&amp;<a href="p" />')).toEqual('<p>&');
```
ray-kraesig, line 237: Not that it matters much here, but self-closing non-void-element tags aren't legal HTML5.

PR #4061, opened 2020-04-21
async: Fudge test value to defeat flake in `sleep` test.

### LINKED ISSUE #4010: Test of `sleep` has flaked in CI
In a recent PR with no changes to either `async.js` or `async-test.js`, one of the tests for "sleep" flaked, mysteriously sleeping for only 999 milliseconds rather than 1000.

```
 FAIL  src/utils/__tests__/async-test.js
  ● sleep (real) › waits for approximately the right number of milliseconds
    expect(received).toBeLessThanOrEqual(expected)
    Expected: <= 999
    Received:    1000
      45 |     // `sleep` should sleep for the specified number of milliseconds (and not,
      46 |     // for example, that many seconds or microseconds).
    > 47 |     expect(expectedMs).toBeLessThanOrEqual(actualMs);
         |                                ^
      48 |     expect(actualMs).toBeLessThan(10 * expectedMs); // [α]
      49 | 
      50 |     // [α] In theory, we can't be sure of this test; the time between the
      at _callee2$ (src/utils/__tests__/async-test.js:47:32)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at invoke (node_modules/regenerator-runtime/runtime.js:152:20)
      at node_modules/regenerator-runtime/runtime.js:162:13
      at tryCallOne (node_modules/promise/lib/core.js:37:12)
      at node_modules/promise/lib/core.js:123:15
      at flush (node_modules/asap/raw.js:50:29)
```

Screenshot_20200328_153252

ray-kraesig:
Sometimes, `setTimeout(1000)` will fire after only 999 milliseconds.
This appears to be due to an upstream bug in Node.js itself [1].

Closes #4010.

[1] https://github.com/nodejs/node/issues/26578

### DIFF
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -44,7 +44,7 @@ describe('sleep (real)', () => {
 
     // `sleep` should sleep for the specified number of milliseconds (and not,
     // for example, that many seconds or microseconds).
-    expect(expectedMs - 1).toBeLessThanOrEqual(actualMs); // [α]
+    expect(actualMs).toBeGreaterThanOrEqual(expectedMs - 1); // [α]
     expect(actualMs).toBeLessThan(10 * expectedMs); // [β]
 
     // [α] The fudge term of -1 in this test is a workaround for a Node.js
--- a/src/utils/__tests__/async-test.js
+++ b/src/utils/__tests__/async-test.js
@@ -44,16 +44,20 @@ describe('sleep (real)', () => {
 
     // `sleep` should sleep for the specified number of milliseconds (and not,
     // for example, that many seconds or microseconds).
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks for tracking that down!

Reassuring, in a way, that it's a specific bug in Node with an understood cause that puts an upper bound on its effect. Less reassuring that the maintainers seem pretty blasé about whether to do anything about it. :-/

Merged, with an additional commit to fix Jest's failure messages when the test fails.

Issue #3767, opened 2020-01-08
Separate StreamList component into two components


chrisbobbe:
It seems that the StreamList component is reused for two purposes that may be different enough to motivate separating it into two different components. It's used to list streams in both the "Subscribed" tab and the "All streams" tab, on the "streams" screen (accessible from the "#" icon at the bottom). Here are some reasons motivating a separation:

* The component uses a React Native SectionList, but only the version on the "Subscribed" tab actually separates the streams into sections: sections for "pinned" and "unpinned" streams. In "All streams," there is just one list of all the streams. This is rather obscurely accomplished by lumping all streams into the "unpinned" section, taking advantage of the fact that section titles aren't displayed.
* The two instances are currently differentiated by certain props being passed, such as "showDescriptions" and "showSwitch." This means there is slightly laborious amount of conditionalizing within this component.
* An unreadByStream prop, which supplies unread counts for the "Subscribed" instance, has an empty defaultProp to avoid crashing on the "All streams" tab. If StreamList were separated, then a reader wouldn't even have to think about unread count logic for "All streams."

Other miscellaneous issues include:

* The 'selected' prop isn't used in either place. Is this a broken feature that should be working, or the remnant of an incomplete removal of a feature?

rheo-chiti:
@chrisbobbe 
Can I work on this issue?

chrisbobbe:
Sure, thank you! Feel free to open a topic in the mobile stream on chat.zulip.org :)

rheo-chiti:
@chrisbobbe  should I make two components ```StreamListCardList``` and ```SubscriptionsCardList``` and then pass the necessary props from ```SubscriptionsCard``` to ```SubscriptionsCardList``` and from ```StreamListCard``` to ```StreamListCardList``` and delete ```StreamList```.

gnprice:
This came up again recently: see [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Stream.2FSubscription.20split/near/1209112), and discussion at https://github.com/zulip/zulip-mobile/pull/4718#discussion_r627779508 .

PR #4062, opened 2020-04-24
UnreadCards: Avoid collision between strings from users and a magic string

chrisbobbe:
We'd like to clarify and untangle most of the UI code in
UnreadCards, StreamItem, and elsewhere (part of that is #3767), but
the most pressing task is to minimally fix an issue where a stream
named "private" causes bugs for other users who are subscribed to it
and are viewing UnreadCards on mobile.

So, fix it at the source, where the `key` in the data is first
assigned, by adding a prefix. Fortunately, consumers of this data
don't ever expect that the `key` will be equal to the stream's name.

The magic string should never have been introduced in the first
place, and it should be removed as part of those larger refactors.
But the temptation to use it would probably not have existed, if
more careful thought had gone into the UI semantics here.

In particular, private messages and stream/topic messages are not
organized the same way, and they have different semantics. It looks
like the magic string was added to deal with those differences in a
world where both kinds of messages must coexist in a single
SectionList. The cleaner solution would have been to use two
separate lists; for example, a FlatList for PMs, and a SectionList
for the stream/topic messages. PMs just don't fit in a heading–item
structure, like streams and topics do, much less one with the same
semantics.

### DIFF
--- a/src/unread/__tests__/unreadSelectors-test.js
+++ b/src/unread/__tests__/unreadSelectors-test.js
@@ -363,7 +363,7 @@ describe('getUnreadStreamsAndTopics', () => {
           },
         ],
         isMuted: true,
-        key: 'stream 0',
+        key: 'stream:stream 0',
         streamName: 'stream 0',
         unread: 5,
       },
@@ -379,7 +379,7 @@ describe('getUnreadStreamsAndTopics', () => {
           },
         ],
         isMuted: true,
-        key: 'stream 2',
+        key: 'stream:stream 2',
         streamName: 'stream 2',
         unread: 2,
       },
@@ -431,7 +431,7 @@ describe('getUnreadStreamsAndTopics', () => {
         ],
         isMuted: false,
         isPrivate: undefined,
-        key: 'stream 0',
+        key: 'stream:stream 0',
         streamName: 'stream 0',
         unread: 2,
       },
@@ -447,7 +447,7 @@ describe('getUnreadStreamsAndTopics', () => {
           },
         ],
         isMuted: false,
-        key: 'stream 2',
+        key: 'stream:stream 2',
         streamName: 'stream 2',
         unread: 2,
       },
@@ -480,7 +480,7 @@ describe('getUnreadStreamsAndTopics', () => {
 
     expect(unreadCount).toEqual([
       {
-        key: 'stream 0',
+        key: 'stream:stream 0',
         streamName: 'stream 0',
         color: 'red',
         unread: 5,
@@ -497,7 +497,7 @@ describe('getUnreadStreamsAndTopics', () => {
         ],
       },
       {
-        key: 'stream 2',
+        key: 'stream:stream 2',
         streamName: 'stream 2',
         color: 'blue',
         unread: 2,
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks for spotting this and fixing it!

I'll go and merge, with just some edits to the commit message: the use of a string here is fine, because that's what React expects for a `key`, and the use of `'private'` as one of the keys is fine. The issue was that a piece of more-or-less arbitrary string data (the stream name) was used without properly encoding it to remain unique against other keys (namely the constant `'private'`).

The technique of prefixing some data with a fixed identifier is a common one for letting arbitrary data  coexist in a namespace alongside other keys or data. Here's some examples from our use of memcached in the Zulip server:
```python
def bot_profile_cache_key(email: str) -> str:
    return "bot_profile:%s" % (make_safe_digest(email.strip()),)

def user_profile_by_id_cache_key(user_profile_id: int) -> str:
    return "user_profile_by_id:%s" % (user_profile_id,)

def user_profile_by_api_key_cache_key(api_key: str) -> str:
    return "user_profile_by_api_key:%s" % (api_key,)
```

chrisbobbe:
@gnprice [said](https://github.com/zulip/zulip-mobile/pull/4062#issuecomment-618742037):
> the use of a string here is fine

Oh, OK, I guess that makes sense. It seemed to match the profile of a "magic string"; Wikipedia [says](https://en.wikipedia.org/wiki/Magic_string), " a magic string is an input that a programmer believes will never come externally and which activates otherwise hidden functionality".

Perhaps this broad definition of a magic string, as a *symptom* (from whatever cause), would still match what we were seeing? I guess I saw the string literal hard-coded there, assumed it was a code smell that had to be gotten rid of, eventually, and named the occurrence (not the symptom) as a magic string.

I do understand the rest —

> because that's what React expects for a `key`, and the use of `'private'` as one of the keys is fine. The issue was that a piece of more-or-less arbitrary string data (the stream name) was used without properly encoding it to remain unique against other keys (namely the constant `'private'`).

— I just want to check my understanding on what a magic string actually is.

gnprice:
Yeah, I think "magic string" is a reasonable diagnosis of what the author(s) of that bit of code might have been thinking when they wrote it that way.

... Or I dunno, if they'd asked themselves the question of whether that "input [would] []ever come externally", would they really have answered "no"? That seems unlikely -- I'd expect that line of thought to lead to something like `@private` or `!!_PRIVATE` (even if those aren't ultimately good answers), rather than just `private`. Seems more like they probably just didn't contemplate the fact that this string `private` was going into the same place as some unadorned inputs go -- they treated it as an identifier, and came up with a perfectly reasonable identifier.

Maybe one way of framing it: a magic string is where you put an identifier in a place that belongs to data. When you put data in a place that belongs to identifiers, or some other particular structure... that's something else, and maybe the best general name for it is "injection". Here quite a mild one, but an HTML injection or SQL injection or XSS are basically the same kind of thing in different contexts. When you have both identifiers and data going in the same place, which kind of problem it is comes down to which one is supposed to be there, and which one is out of place.

chrisbobbe:
Mmm, OK! Thanks for that explanation.

PR #4063, opened 2020-04-24
Automate translation-database integration (alternate)

### LINKED ISSUE #3895: Usable and used languages should be (better) synchronized

ray-kraesig:
Add a script to simplify pulling translation data. 

Alternative of #3915.

Fixes #3895.


ray-kraesig:
Updated and rebased, in light of the recent Transifex sync. No changes to non-generated code were needed.

PR #4064, opened 2020-04-24
converting tests to type check

sumukhah:
Some helper functions are used in order to improve code readability, reduce lines of code. some helper functions like(`mapToMakeUser`, `extractNameAndEmail`, e.t.c). 
But normally with the test files, this won't be implemented. So if any changes regarding helper functions are needed I would like to refactor.


src/users/__tests__/userHelpers-test.js
```diff
@@ -180,24 +217,78 @@ describe('sortUserList', () => {
...
+    Object.keys(presences).forEach(user => {
+      const userStatus = {};
+      const email: string = user;
+      const userPresence = presences[email];
+      Object.keys(userPresence).forEach(key => {
+        const presenceKey: string = key;
+        userStatus[presenceKey] = userPresence[presenceKey];
+      });
+      modifiedPresences[email] = userStatus;
+    });
```
sumukhah, line 288: These steps are required for converting to exact Objects with specific indexers.

sumukhah, line 288: These steps are required for converting to exact Objects with specific indexers.

sumukhah:
Please review @ray-kraesig, @gnprice, @chrisbobbe, @jainkuniya

src/users/__tests__/userHelpers-test.js
```diff
@@ -271,30 +368,75 @@ describe('groupUsersByStatus', () => {
...
```
ray-kraesig, line 432: If this were needed, it would definitely need to be encapsulated in its own function, rather than used inline twice.

src/users/__tests__/userHelpers-test.js
```diff
@@ -271,30 +368,75 @@ describe('groupUsersByStatus', () => {
...
+
+    const groupedUsers = groupUsersByStatus(mapToMakeUser(users), modifiedPresences);
+    const extractedUsers = {};
+    Object.keys(groupedUsers).forEach(state => {
+      extractedUsers[state] = extractNameAndEmail(groupedUsers[state]);
+    });
```
ray-kraesig, line 438: This, too, is unneeded, due to the semantics of `toMatchObject`.

src/users/__tests__/userHelpers-test.js
```diff
@@ -180,24 +217,78 @@ describe('sortUserList', () => {
...
+
     const presences = {
-      'mark@example.com': { aggregated: { status: 'offline' } },
+      'mark@example.com': {
+        aggregated: { status: 'offline', client: 'ZulipMobile', timestamp: Date.now() },
+        ZulipMobile: {
+          status: 'offline',
+          client: 'ZulipMobile',
+          timestamp: Date.now(),
+          pushable: false,
```
ray-kraesig, line 228: `pushable` can't cleanly be supplied, and probably isn't needed anyway. See the comments at its declaration.

src/__tests__/lib/exampleData.js
```diff
@@ -45,7 +45,7 @@ const makeUniqueRandInt = (itemsType: string, end: number): (() => number) => {
 export const randString = () => randInt(2 ** 54).toString(36);
 
 const randUserId: () => number = makeUniqueRandInt('user IDs', 10000);
-const userOrBotProperties = ({ name: _name }) => {
+const userOrBotProperties = ({ name: _name, email: _email }) => {
```
ray-kraesig, line 55: No need for the initial `_` on `_email`. (We only have one on `name` because of the lint error that would otherwise result.)

src/__tests__/lib/exampleData.js
```diff
@@ -142,6 +149,24 @@ export const makeStream = (args: { name?: string, description?: string } = {}):
...
 
+const randGroupId: () => number = makeUniqueRandInt('group IDs', 1000);
+export const makeGroup = (args: { name?: string, description?: string } = {}): UserGroup => {
+  const name = args.name ?? randString();
+  const description = args.description ?? `This is a group for discussing ${randString()}`;
+  const members = [];
+
+  for (let i = 0; i < randInt(10) + 1; i++) {
+    members.push(randInt(1000));
+  }
```
ray-kraesig, line 160: As written, this means the same user could be present in a group twice. I suspect that might violate some expectations.

Additionally, `i < randInt(10) + 1` is being executed on every pass through the loop; this means the output of `randInt` will be different every time, and the average number of loops will be much less than 5.

sumukhah:
committed some changes, can you review this? @rk-for-zulip

PR #4065, opened 2020-04-24
Refactor constructMessageActionButtons

ray-kraesig:
Break `constructMessageActionButtons` up into two functions, depending on whether the message being processed is an Outbox message or not. Type this properly.

Extracted from PR #3978.


gnprice:
Thanks -- merged!

PR #4066, opened 2020-04-25
streams: Add functionality to create and edit stream with stream post policy.

jainkuniya:
Continuing from [pervious PR](https://github.com/zulip/zulip-mobile/pull/3140#issuecomment-617976221)

- now it uses the same option text as web, just trimmed the  "can post" part to save some space. 

image

image

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -169,6 +169,7 @@ export const makeStream = (args: { name?: string, description?: string } = {}): 
     invite_only: false,
     is_announcement_only: false,
     history_public_to_subscribers: true,
+    stream_post_policy: 1,
   });
 };
 
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -242,6 +242,7 @@ export type Stream = {|
   invite_only: boolean,
   is_announcement_only: boolean,
   history_public_to_subscribers: boolean,
+  stream_post_policy: number,
 |};
 
 export type Subscription = {|
--- a/src/api/streams/createStream.js
+++ b/src/api/streams/createStream.js
@@ -9,11 +9,13 @@ export default (
   description?: string = '',
   principals?: string[] = [],
   inviteOnly?: boolean = false,
+  streamPostPolicy?: number = 1,
   announce?: boolean = false,
 ): Promise<ApiResponse> =>
   apiPost(auth, 'users/me/subscriptions', {
     subscriptions: JSON.stringify([{ name, description }]),
     principals: JSON.stringify(principals),
     invite_only: inviteOnly,
+    stream_post_policy: streamPostPolicy,
     announce,
   });
--- a/src/common/SelectionList.js
+++ b/src/common/SelectionList.js
@@ -1,0 +1,155 @@
+/* @flow strict-local */
+import React, { PureComponent } from 'react';
+import { FlatList, StyleSheet, View } from 'react-native';
+import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
+import type { NavigationScreenProp } from 'react-navigation';
+
+import type { Dispatch } from '../types';
+import type { ThemeColors } from '../styles';
+import { connect } from '../react-redux';
+import Label from './Label';
+import { IconDone, IconRight } from './Icons';
+import Touchable from './Touchable';
+import { BRAND_COLOR, ThemeContext } from '../styles';
+import { navigateToZulipPickerOptionsScreen, navigateBack } from '../nav/navActions';
+import OptionDivider from './OptionDivider';
+import Screen from './Screen';
+
+const componentStyles = StyleSheet.create({
+  selectorWrapper: {
+    marginTop: 4,
+  },
+  selectorRow: {
+    flex: 1,
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+  },
+  selectorValue: {
+    color: 'gray',
+    marginTop: 4,
+  },
+  selectorIcon: {
+    alignSelf: 'center',
+  },
+  selectorLabel: {
+    fontWeight: '400',
+  },
+  optionItem: {
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    paddingTop: 12,
+    paddingBottom: 12,
+    paddingLeft: 16,
+    paddingRight: 16,
+  },
+  optionSelectedValue: { color: BRAND_COLOR },
+});
+
+type OptionsScreenProps = $ReadOnly<{|
+  dispatch: Dispatch,
+  navigation: NavigationScreenProp<{
+    params: {|
+      label: string,
+      options: $ReadOnlyArray<{| key: number, value: string |}>,
+      selectedIndex: number,
+      onOptionSelect: (value: number) => void,
+    |},
+  }>,
+|}>;
+
+/**
+ * A selector option list component.
+ *
+ * @prop label - label for selectionl list.
+ * @prop options - all available options to select from.
+ * @prop selectedIndex - index of currently selected option.
+ * @prop onOptionSelect - Event called when any option is selected.
+ */
+class OptionsScreen extends PureComponent<OptionsScreenProps> {
+  onOptionSelect = (index: number) => {
+    const { dispatch, navigation } = this.props;
+    const { onOptionSelect } = navigation.state.params;
+    onOptionSelect(index);
+    dispatch(navigateBack());
+  };
+
+  render() {
+    const { label, options, selectedIndex } = this.props.navigation.state.params;
+
+    return (
+      <Screen title={label}>
+        <FlatList
+          ItemSeparatorComponent={OptionDivider}
+          data={options}
+          keyExtractor={item => `${item.key}`}
+          renderItem={({ item, index }) => (
+            <Touchable onPress={() => this.onOptionSelect(index)}>
+              <View style={componentStyles.optionItem}>
+                <Label
+                  style={selectedIndex === index && componentStyles.optionSelectedValue}
+                  text={item.value}
+                />
+                {selectedIndex === index && <IconDone size={16} color={BRAND_COLOR} />}
+              </View>
+            </Touchable>
+          )}
+        />
+      </Screen>
+    );
+  }
+}
+
+type SelectorProps = $ReadOnly<{|
+  dispatch: Dispatch,
+  label: string,
+  options: $ReadOnlyArray<{| key: number, value: string |}>,
+  selectedIndex: number,
+  style?: ViewStyleProp,
+  onOptionSelect: (value: number) => void,
+|}>;
+
+/**
+ * An selector component, which shows the the label and selected value
+ *
+ * @prop [style] - Style object applied to the outermost component.
+ * @prop label - label of the picker.
+ * @prop options - all available options to select from.
+ * @prop selectedIndex - index of currently selected option.
+ * @prop onOptionSelect - Event called when any option is selected.
+ */
+class Selector extends PureComponent<SelectorProps> {
+  static contextType = ThemeContext;
+  context: ThemeColors;
+
+  styles = { icon: { color: this.context.color } };
+
+  navigateToZulipPickerOptions = () => {
+    const { dispatch, options, label, selectedIndex, onOptionSelect } = this.props;
+    dispatch(navigateToZulipPickerOptionsScreen(label, options, selectedIndex, onOptionSelect));
+  };
+
+  render() {
+    const { label, options, selectedIndex, style } = this.props;
+
+    return (
+      <Touchable
+        style={[style, componentStyles.selectorWrapper]}
+        onPress={this.navigateToZulipPickerOptions}
+      >
+        <View style={componentStyles.selectorRow}>
+          <View>
+            <Label style={componentStyles.selectorLabel} text={label} />
+            <Label style={componentStyles.selectorValue} text={options[selectedIndex].value} />
+          </View>
+          <IconRight size={24} style={[this.styles.icon, componentStyles.selectorIcon]} />
+        </View>
+      </Touchable>
+    );
+  }
+}
+
+export default {
+  Selector: connect()(Selector),
+  OptionsScreen: connect()(OptionsScreen),
+};
--- a/src/common/index.js
+++ b/src/common/index.js
@@ -42,4 +42,5 @@ export { default as ViewPlaceholder } from './ViewPlaceholder';
 export { default as WebLink } from './WebLink';
 export { default as ZulipButton } from './ZulipButton';
 export { default as ZulipStatusBar } from './ZulipStatusBar';
+export { default as SelectionList } from './SelectionList';
 export { default as ZulipSwitch } from './ZulipSwitch';
--- a/src/streams/EditStreamCard.js
+++ b/src/streams/EditStreamCard.js
@@ -4,10 +4,18 @@ import { StyleSheet, View } from 'react-native';
 
 import type { Dispatch } from '../types';
 import { connect } from '../react-redux';
-import { Input, Label, OptionRow, ZulipButton } from '../common';
+import { Input, Label, OptionRow, ZulipButton, SelectionList } from '../common';
 import styles from '../styles';
 import { IconPrivate } from '../common/Icons';
 import { getIsAdmin } from '../selectors';
+
+type StreamPostPolicy = $ReadOnly<{| key: number, value: string |}>;
+
+const STREAM_POST_POLICIES: StreamPostPolicy[] = [
+  { key: 1, value: 'All stream members' },
+  { key: 2, value: 'Only organization administrators' },
+  { key: 3, value: 'Only organization full members' },
+];
 
 const componentStyles = StyleSheet.create({
   optionRow: {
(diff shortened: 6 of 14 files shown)


src/streams/EditStreamCard.js
```diff
@@ -2,9 +2,20 @@
...
 import { IconPrivate } from '../common/Icons';
+import { getIsAdmin } from '../selectors';
+
+type StreamPostPolicy = $ReadOnly<{| key: number, value: string |}>;
+
+const STREAM_POST_POLICIES: StreamPostPolicy[] = [
+  { key: 1, value: 'All stream members' },
+  { key: 2, value: 'Only organization administrators' },
+  { key: 3, value: 'Only organization full members' },
+];
```
gnprice, line 18: The keys here are part of the server API, right? I.e. this won't work right if we used some other numbers.

It's important to draw that connection explicitly. The most natural way to do that is:
* Put the definitions of these constants -- giving names for 1, 2, and 3 -- in the types in `src/api/`.
* Then from there link to the relevant bit of API docs, which is here: https://zulipchat.com/api/add-subscriptions

Concretely, try this in `modelTypes.js`:
```js
// See stream_post_policy at https://zulipchat.com/api/add-subscriptions
export class StreamPostPolicies {
  static everyone: 1 = 1;
  static admins: 2 = 2;
  static restrict_new_members: 3 = 3;
}

export type StreamPostPolicy =
  | typeof StreamPostPolicies.everyone
  | typeof StreamPostPolicies.admins
  | typeof StreamPostPolicies.restrict_new_members;

export type Stream = {|
  // ...
  stream_post_policy: StreamPostPolicy,
```

That's modeled on what we have at `EventTypes` in `eventTypes.js`. There may be other ways to do it but that's one reasonably-clean way.

Then anyplace you would say something like `1`, say instead `StreamPostPolicies.everyone`.

src/streams/EditStreamCard.js
```diff
@@ -13,33 +24,55 @@ const componentStyles = StyleSheet.create({
   },
 });
 
+type SelectorProps = {|
+  isAdmin: boolean,
+|};
+
 type Props = $ReadOnly<{|
+  dispatch: Dispatch,
```
gnprice, line 32: follow style for where to place this -- see other examples

src/streams/EditStreamCard.js
```diff
@@ -13,33 +24,55 @@ const componentStyles = StyleSheet.create({
...
+    super(props);
+
+    this.state = {
+      name: this.props.initialValues.name,
+      description: this.props.initialValues.description,
+      isPrivate: this.props.initialValues.invite_only,
+      streamPostPolicyIndex:
+        STREAM_POST_POLICIES.findIndex(
+          policy => policy.key === this.props.initialValues.stream_post_policy,
+        ) || 0,
```
gnprice, line 68: This alternation between "index" and "key" seems pretty fragile -- a sharp edge in the API of `SelectionList.Selector`.

Can you revise it so that instead of `selectedIndex` it takes `selectedKey`? And then everything can work with keys and forget about indexes, except the `renderItem` callback in `OptionsScreen`.

src/api/modelTypes.js
```diff
@@ -242,6 +242,7 @@ export type Stream = {|
   invite_only: boolean,
   is_announcement_only: boolean,
   history_public_to_subscribers: boolean,
+  stream_post_policy: number,
```
gnprice, line 245: This should be optional! It sure isn't always there.

Also should have a comment saying when it was introduced.

src/api/modelTypes.js
```diff
@@ -242,6 +242,7 @@ export type Stream = {|
   invite_only: boolean,
   is_announcement_only: boolean,
```
gnprice, line 243: This should become optional now, with a comment saying when it was removed.

src/api/modelTypes.js
gnprice: Also should be next to `is_announcement_only`, as they're closely related.

src/common/SelectionList.js
```diff
@@ -0,0 +1,155 @@
...
+ * @prop label - label of the picker.
+ * @prop options - all available options to select from.
+ * @prop selectedIndex - index of currently selected option.
+ * @prop onOptionSelect - Event called when any option is selected.
+ */
+class Selector extends PureComponent<SelectorProps> {
+  static contextType = ThemeContext;
+  context: ThemeColors;
+
+  styles = { icon: { color: this.context.color } };
```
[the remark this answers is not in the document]
jainkuniya: It is only one style, I would say let's inline it.

src/common/SelectionList.js
```diff
@@ -0,0 +1,155 @@
...
+  selectorValue: {
+    color: 'gray',
+    marginTop: 4,
+  },
+  selectorIcon: {
+    alignSelf: 'center',
+  },
+  selectorLabel: {
+    fontWeight: '400',
+  },
```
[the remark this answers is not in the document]
jainkuniya: sure

src/streams/EditStreamCard.js
jainkuniya: Initially I was passing `selectedKey`, actually it was making `Selector` complicated. As now it has to lookup in whole list to find label for the selected item. 
So I thought going with `selectedIndex`.

src/api/modelTypes.js
jainkuniya: > Also should be next to is_announcement_only, as they're closely related.

I prefer alphabetically order, because when list is long it is easy to guess and locate desired item. You don't have to go through whole list.

src/api/modelTypes.js
jainkuniya: I checked on `2.2-dev-2006-g0cf4c89713`, we are still getting `is_announcement_only `. 
My main concern is, should we make it optional as part of this change because it will affect `getIsActiveStreamAnnouncementOnly` which is used in `ComposeBox.`

src/streams/EditStreamCard.js
gnprice: I'm not sure what you mean by "label" -- that isn't a term used in this code, except for something quite different.

There shouldn't be a need to go look through the whole list to find something. If you're referring to `onOptionSelect` -- it can just pass the key to the callback.

src/api/modelTypes.js
gnprice: I see. OK, if the server is still sending it then it wasn't removed and we don't need to mark it optional. Someday (zulip/zulip#14188) there'll be API docs for this.

gnprice:
@jainkuniya You've replied to some comments mentioning changes, but haven't pushed a new revision -- not sure if you meant to but here's a reminder in case you did.

src/streams/EditStreamCard.js
jainkuniya: By `label` I mean text which is show to user for each option.

So, I am referring to the currently selected option text shown in the selector.

Let's define a data structure like this:
```js
STREAM_POST_POLICIES = {
    StreamPostPolicies.everyone: {label: "All stream members"},
    .........
}
```

and pass this to `SelectionList`. And it can take `selectedKey`.

Not in `Selector` it will be easy to get selected option text (label), 
```
options[selectedKey].label
```

now but in `OptionsScreen ` we have to do `Objects.keys` to convert it into array

jainkuniya:
Ya, I haven't pushed the code yet. I commented about the changes, just to mark them as I have accommodated them.

(Still in between of the rebase)

src/common/SelectionList.js
```diff
@@ -0,0 +1,155 @@
+/* @flow strict-local */
```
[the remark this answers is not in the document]
jainkuniya: Sure, updated 👍

src/streams/EditStreamCard.js
jainkuniya: Created `streamPostPolicy.js`

src/streams/EditStreamCard.js
jainkuniya: Changed the data structure of the selection list options, now it is
```
export type OptionsType = $ReadOnly<{| [key: string]: { label: string } |}>;
```

and `selectedKey: string,` is passed everywhere.

src/common/SelectionList.js
```diff
@@ -0,0 +1,176 @@
...
+        </View>
+      </Touchable>
+    );
+  }
+}
+
+const SelectionScreen: {
+  Selector: ComponentType<SelectorOuterProps>,
+  OptionsScreen: ComponentType<OptionsScreenOuterProps>,
+} = {
```
gnprice, line 171: Hmm, when you have to write a cast like this, that's a sign something's gone wrong with our type-checking :slightly_smiling_face: 

It looks like no-argument `connect` falls into an `empty`-type trap. The result is that you could write something quite untrue for the props types in this cast, and Flow would accept it.

I just experimented a bit, and found a workaround that avoids having to repeat ourselves, and gets Flow successfully type-checking the props we pass in when we use the component. Write the `connect` call like so:
```
connect<{||}, _, _>()
```

So e.g.
```js
export default {
  Selector: connect<{||}, _, _>()(Selector),
  OptionsScreen: connect<{||}, _, _>()(OptionsScreen),
};
```

I'll go fix that in the existing places we have no-argument `connect` in the tree, too.

src/streams/EditStreamCard.js
```diff
@@ -2,9 +2,19 @@
...
 import styles from '../styles';
 import { IconPrivate } from '../common/Icons';
+import { getIsAdmin } from '../selectors';
+import { StreamPostPolicies } from '../api/streams/streamPostPolicy';
+
+const STREAM_POST_POLICIES = {
+  [StreamPostPolicies.everyone]: { label: 'All stream members' },
+  [StreamPostPolicies.admins]: { label: 'Only organization administrators' },
+  [StreamPostPolicies.restrict_new_members]: { label: 'Only organization full members' },
+};
```
gnprice, line 17: Let's try using a real JS `Map` here, rather than an object-as-map.

One reason is that that's just cleaner in general about our intentions. Related to that is that it allows the keys to be any values we like -- in particular, they can be the actual `StreamPostPolicies.everyone` etc. -- rather than having to convert them to strings.

And then a bonus is that I suspect it will let us get rid of the need to ever convert a string *back* to a `StreamPostPolicy` value.

src/api/streams/streamPostPolicy.js
```diff
@@ -0,0 +1,22 @@
...
+      case this.everyone:
+        return this.everyone;
+      case this.admins:
+        return this.admins;
+      case this.restrict_new_members:
+        return this.restrict_new_members;
+      default:
+        return this.everyone;
+    }
+  };
```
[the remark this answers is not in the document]
gnprice: OK, and now that I've read the whole thing -- I think this function can actually be dispensed with entirely! See my comment below about using a real `Map` vs. an object-as-Map. That will let us pass actual `StreamPostPolicy` values down through the layers here, instead of strings. And that should be able to go all the way to this line:
```
            <Touchable onPress={() => this.onOptionSelect(item.key)}>
```
so that `item.key` at this point is already of the right type and there's no need for the `onOptionSelect` callback to turn it back from a string.

We'll still need to stringify the keys to use as actual React keys, but that can be done just at this one line:
```
          keyExtractor={item => `${item.key}`}
```
and not interfere with the types anywhere else.

src/common/SelectionList.js
```diff
@@ -0,0 +1,176 @@
...
+import { connect } from '../react-redux';
+import Label from './Label';
+import { IconDone, IconRight } from './Icons';
+import Touchable from './Touchable';
+import { BRAND_COLOR, ThemeContext } from '../styles';
+import { navigateToSelectionListOptionsScreen, navigateBack } from '../nav/navActions';
+import OptionDivider from './OptionDivider';
+import Screen from './Screen';
+
+export type OptionsType = $ReadOnly<{| [key: string]: { label: string } |}>;
```
gnprice, line 19: Let's make this like
```js
export type SelectionOptions<K> = Map<K, {| +label: string |}>;
```

That is:
* use a real `Map`, as discussed in previous comment
* use an exact object type
* name doesn't end in "Type" just because it's a type -- same reason we don't say `class OptionsScreenClass`, etc. Rather a name ending in "Type" means a type *of things which are themselves types*.
* "Options" is quite generic; add the word "Selection" from the name of this module.

src/common/SelectionList.js
gnprice: > I'll go fix that in the existing places we have no-argument `connect` in the tree, too.

Done: 1e0006e5b

src/api/streams/streamPostPolicy.js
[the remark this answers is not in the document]
jainkuniya:
```
keyExtractor={item => `${item.key}`}
```

^ flow won't allow this, throwing following error

```
Cannot coerce `item.key` to string because  `K` [1] should not be coerced.
```

But Flow is happy with `String(item.key)`, which is using `String` constructor
image

jainkuniya:
Pushed an updated version.

There is one flow error, which I still need to figure out
```
Cannot call this.handleOptionSelect with item[0] bound to selectedKey because K [1] is incompatible with K [2].

     src/common/SelectionList.js
 [2]  64│   handleOptionSelect = (selectedKey: K) => {
        :
      78│           data={Array.from(options)}
      79│           keyExtractor={item => String(item[0]).toString()}
      80│           renderItem={({ item, index }) => (
      81│             <Touchable onPress={() => this.handleOptionSelect(item[0])}>
      82│               <View style={this.styles.item}>
      83│                 <Label
      84│                   style={selectedKey === item[0] && this.styles.selectedValue}

     /private/tmp/flow/flowlib_203af7c0/core.js
 [1] 544│     @@iterator(): Iterator<[K, V]>;


```

From last version, changed one behaviour: now on selecting an option, we don't navigate back automatically. I checked in the iOS (from where we got inspiration of this component), they too don't navigate back
image

gnprice:
> There is one flow error, which I still need to figure out

I poked around a bit, and the code seems fine here. Looks like there's a bug in Flow that could be causing this:
https://github.com/facebook/flow/issues/8242

So just add a fixme, with a comment next to it pointing to that Flow issue.

Also there's an encouraging response on the bug! From 2019-12:
> This looks like a bug with generics. We are planning a big overhaul of how generic typing works in the first half of 2020, so that should address this.

... Oh! Actually, better: use the workaround described in the last comment:
```
  handleOptionSelect(selectedKey: K) {
```

That seems to work.

jainkuniya:
Thanks @gnprice for pointing out the workaround.

src/api/modelTypes.js
```diff
@@ -244,12 +244,34 @@ export type UserPresence = {|
...
+  /**
+   * Defines who can post to the stream.
+   * is_announcement_only is also one of the policy.
+   * But going forward there can be multiple post polices, see StreamPostPolicies.
+   * So it is suggested to use stream_post_policy instead of is_announcement_only,
+   * which might get removed in future
+   *
+   * This is available on Zulip server running >= 2.2
+   */
```
gnprice, line 273: (I'll adjust the jsdoc style some before I merge.)

src/api/streams/createStream.js
```diff
@@ -9,11 +11,13 @@ export default (
...
   principals?: string[] = [],
   inviteOnly?: boolean = false,
+  streamPostPolicy?: StreamPostPolicy = StreamPostPolicies.everyone,
   announce?: boolean = false,
 ): Promise<ApiResponse> =>
   apiPost(auth, 'users/me/subscriptions', {
     subscriptions: JSON.stringify([{ name, description }]),
     principals: JSON.stringify(principals),
     invite_only: inviteOnly,
+    stream_post_policy: streamPostPolicy,
```
gnprice, line 21: This shouldn't be labeled NFC -- we're changing what we send to the server.

How does this behave on older servers that don't expect this parameter? I'd expect an error.

src/__tests__/lib/exampleData.js
```diff
@@ -169,6 +169,7 @@ export const makeStream = (args: { name?: string, description?: string } = {}):
     description,
     invite_only: false,
     is_announcement_only: false,
+    stream_post_policy: 1,
```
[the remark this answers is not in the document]
gnprice: > everywhere else that provides a value like 1 for these

There are still other places like this. I believe in your ability to find them. :slightly_smiling_face: I don't want to have to make a GitHub comment on each one, which is why I wrote this.

src/common/SelectionList.js
```diff
@@ -0,0 +1,176 @@
...
+          renderItem={({ item, index }) => (
+            <Touchable onPress={() => this.handleOptionSelect(item[0])}>
+              <View style={this.styles.item}>
+                <Label
+                  style={selectedKey === item[0] && this.styles.selectedValue}
+                  text={item[1].label}
+                />
+                {selectedKey === item[0] && <IconDone size={16} color={BRAND_COLOR} />}
+              </View>
+            </Touchable>
```
gnprice, line 94: The `item[0]` and `item[1]` business here is kind of messy and opaque. (Though I sure am glad that Flow understands it and successfully provides type-checking!)

Instead, we can replace the `Array.from(options)` with something like
```js
Array.from(options, ([key, option]) => ({ ...option, key }))
```
so that the items have properties `key` and `label`.

src/common/SelectionList.js
```diff
@@ -0,0 +1,176 @@
...
+ * @prop selectedKey - key of currently selected option.
+ * @prop onOptionSelect - callback when any option is selected.
+ */
+export class OptionsScreen<K> extends PureComponent<OptionsScreenProps<K>, OptionsScreenState<K>> {
+  constructor(props: OptionsScreenProps<K>) {
+    super(props);
+    this.state = {
+      selectedKey: props.navigation.state.params.selectedKey,
+    };
+  }
```
gnprice, line 49: We normally write this like
```js
    state = {
￼      selectedKey: this.props.navigation.state.params.selectedKey,
￼    };
```

src/common/index.js
```diff
@@ -42,4 +42,5 @@ export { default as ViewPlaceholder } from './ViewPlaceholder';
 export { default as WebLink } from './WebLink';
 export { default as ZulipButton } from './ZulipButton';
 export { default as ZulipStatusBar } from './ZulipStatusBar';
+export { default as SelectionList } from './SelectionList';
 export { default as ZulipSwitch } from './ZulipSwitch';
```
gnprice, line 46: Notice these are alphabetized :slightly_smiling_face: -- please keep them that way

static/translations/messages_en.json
```diff
@@ -211,5 +211,9 @@
   "🚌 Commuting": "🚌 Commuting",
   "🤒 Out sick": "🤒 Out sick",
   "🌴 Vacationing": "🌴 Vacationing",
-  "🏠 Working remotely": "🏠 Working remotely"
+  "🏠 Working remotely": "🏠 Working remotely",
```
gnprice, line 214: one more nit: please add somewhere other than the end of this list. That way there isn't the extra removed/added line in the diff from the added comma.

(The root problem here is with JSON's syntax, but in this context it's easy to work around.)

Issue #4067, opened 2020-04-25
Narrow contains email 'welcome-bot@zulip.com' that does not map to any user.


Maskedman99:
![Screenshot_20200425-153722_Zulip (debug)](https://user-images.githubusercontent.com/31368194/80277799-2d26d400-870f-11ea-94d6-6686893bc121.jpg)

Issue happens while trying to chat with Welcome Bot. Happening in the master branch.

@zulipbot claim

Maskedman99:
Looks like https://github.com/zulip/zulip-mobile/commit/531670135f5f2b7ad411873a27679d62c0d059ac is what introduced the bug, ```getUsersByEmail``` was used instead ```getAllUsersByEmail```

ray-kraesig:
Confirmed. Raising priority.

PR #4068, opened 2020-04-25
typingSelectors: Fix selecting welcome bot in pm crashes the app

Maskedman99:
Issue was caused beacuse ```getUsersByEmail``` couldn't determine the user-id of Welcome Bot. This is fixed in the PR by making use of ```getAllUsersByEmail``` instead.

Fixes: #4067.

### DIFF
--- a/src/typing/__tests__/typingSelectors-test.js
+++ b/src/typing/__tests__/typingSelectors-test.js
@@ -90,4 +90,30 @@ describe('getCurrentTypingUsers', () => {
 
     expect(typingUsers).toEqual([expectedUser]);
   });
+
+  test('when in a private narrow with a deactivated user, return valid data', () => {
+    const deactivatedUser = eg.makeUser();
+    const state = eg.reduxState({
+      typing: {},
+      realm: eg.realmState({ nonActiveUsers: [deactivatedUser] }),
+    });
+
+    const getTypingUsers = () => getCurrentTypingUsers(state, privateNarrow(deactivatedUser.email));
+
+    expect(getTypingUsers).not.toThrow();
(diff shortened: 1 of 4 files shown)


ray-kraesig:
For reference, this was introduced in #4054.

LGTM at first glance, but pinging @gnprice and @agrawal-d for comments.

agrawal-d:
Thanks for fixing this, @Maskedman99!
You can also modify the tests by adding some `CrossRealmBots` to `state.realm.crossRealmBots` in `src/typing/__tests__/typingSelectors-test.js`, by using `eg.makeCrossRealmBot`.
Looks good otherwise.

ray-kraesig:
As this is an active crash bug that hasn't yet made it out to a release, I'll go ahead and add tests myself.

gnprice:
Thanks all! Merged -- glad to have this caught and fixed.

I'd describe the underlying cause of this as the fact that we were using the `getUsersBy*` selectors here, which are in fact limited to just some users, instead of `getAllUsersBy*` which cover all users. Then the change in #4054 increased the impact of the bug by a lot, because it caused this code to start using the user data structures earlier and for all the users in the thread, instead of just the ones that are typing at the moment.

So I've added a commit on top which fixes that underlying bug too. This is a good example of how a bug which has a small direct impact because it's very hard to trigger in itself can be a "latent bug" which starts having a much bigger impact later on after an otherwise-innocent change to the code.

And then the *root* cause is more like: the `getUsersBy*` selectors sure sound like they're the thing to use, as they have the obvious, unmarked names. I'll file an issue for fixing that generally.

PR #4069, opened 2020-04-25
group-pm: Fix breaking of navbar when there is a large no.of recipients

Maskedman99:
With large number of recipients in group-pm the navbar overflows with avatars. This is fixed by displaying the avatars in a ```Flatlist``` which hides the overflow effectively by letting users scroll through the avatars, while also providing the memory enhancements that are needed with the large
data.

Fixes: #3700.

### DIFF
--- a/src/title/TitleGroup.js
+++ b/src/title/TitleGroup.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 
 import React, { PureComponent } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { StyleSheet, View, FlatList } from 'react-native';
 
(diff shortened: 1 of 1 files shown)


Maskedman99:
---------------------------------------------------------Before--------------------------------------------------------------
Screenshot-20200425212122-365x111
-------------------------------------------------------------After-------------------------------------------------------------
InShot-20200425-211640155

ray-kraesig:
Functional, and would be an improvement as is – but when the list is scrolled all the way to the end, there's some unbalanced trailing space. Can that be easily fixed?

Maskedman99:
@ray-kraesig I've fixed the trailing space

Maskedman99:
@ray-kraesig review

Issue #4042, opened 2020-04-15
iOS: message list doesn't respect system text size


gnprice:
This is just like #2559, but I saw it just now in a recent version from master. This was on a simulated iPhone 8 running iOS 13.3.

> I had the Zulip app open to the message list screen, then went over to system settings and changed the text size. The text in the app bar responded appropriately, but the message list didn't change.
>
> This would make most of the app's functionality totally unusable for someone with a visual impairment who needs a large text size.

Not sure what changed between 642b2d43b, where I previously fixed this bug, and now. The only change since then that looks at all related is the very next commit, cbf2e6ce9, which deleted the `font-size:` line entirely.

Or maybe there's some other variable involved? I.e. I'm just testing it differently than I did when I wrote 642b2d43b, and the old way would still work fine now while the new way would already have reproed back then.

gnprice:
> The only change since then that looks at all related is the very next commit, [cbf2e6c](https://github.com/zulip/zulip-mobile/commit/cbf2e6ce94ca4d9f3744d59030ff1671dd690cac), which deleted the `font-size:` line entirely.

Just to be sure, I tested just now the hypothesis that this somehow caused the bug: in Safari's Web Inspector, I added a `font-size: 93.75%` attribute back to the `html` CSS block, just like the one that commit removed. The bug still reproed -- the font just got slightly smaller, as you'd expect.

gnprice:
The bug is already present in 26.21.144, which is the oldest release that Apple still keeps around in TestFlight even for me as the app's maintainer to install for testing. That's on my iPhone 7 running iOS 12.4.1; I've also confirmed the bug is present on that version in the latest release.

The release that originally had (what I sure thought at the time was) the fix was 25.3.117.

It would be great for someone to take this on, and debug and fix it. The next steps I'd pursue are:
* Try to find an older version of the app where this works (where the bug wasn't there.) In particular, confirm it worked in 25.3.117, or at commit 642b2d43b.
  * If you run into trouble running older versions of the app in development, see our [troubleshooting steps](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/build-run.md). If you run into anything that isn't there, please describe it in chat -- debugging any issues of that kind will be valuable in itself.
* If it doesn't work at the commit that was supposed to fix it, that'll be interesting and we'll have to debug further from there.
[...]

PR #4071, opened 2020-04-26
webview, css: Apply font: -apple-system-body; to body.

jainkuniya:
Fixes: #4042

This is based on https://www.interactiveaccessibility.com/blog/text-resizing-web-pages-ios-using-dynamic-type#.XqVIvpMzZ25

- Change font size in iOS setting to test this PR, Setting -> General -> Accessibility -> Larger text
- After changing text size, reload the webview. Changes are only effective when webview is re-rendered.
- It only works for text whose html element doesn't override `font-size` (like `.code`) and rely on parent (`body`) css styles.

Screenshots (iPhone X), when `font: -apple-system-body;` is applied to body:
- Max font size (Larger accessibility toggle is off)
image

- Min font size (Larger accessibility toggle is off)
image

Also works fine when Larger accessibility toggle is on
image

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -5,6 +5,7 @@ html {
   user-select: none; /* Standard syntax */
   -khtml-user-select: none;
   -webkit-touch-callout: none;
+  font: -apple-system-body;
 }
 body {
   font-family: sans-serif;


gnprice:
Very interesting!

Would you please post screenshots of what this looks like?

At the code level: this attribute should appear next to the existing `font-family` attribute.

gnprice:
> * It only works for text whose html element doesn't override `font-size` (like `.code`) and rely on parent (`body`) css styles.

Hmm. Here's what we have for `code`:
```css
code {
  font-size: .857rem;
  white-space: pre-wrap;
  padding: 0 4px;
}
```

That uses `rem` units, which is what our style guide [calls for](https://github.com/zulip/zulip-mobile/blob/master/docs/background/webview.md) for font sizes (and `px` for the sizes of everything that isn't text), precisely because it should do the right thing here.

... Oh, I see. It's because the PR puts the size on `body`, which isn't the root element. Try putting it on `html` instead.

Here's a nice short tutorial on the subject:
https://dev.to/colingourlay/how-to-support-apple-s-dynamic-text-in-your-web-content-with-css-40c0
I am still (a) puzzled why this sure seemed to work before, and (b) for that matter, puzzled that according to these sources, Apple seems to have arranged things so that these accessibility settings don't work on web pages unless the developer uses this arcane undocumented syntax to opt in. But, glad you've found it!

jainkuniya:
(Updated with the screenshots)

jainkuniya:
After applying it to html tag, yes code block is also adopting size from system
image
image

jainkuniya:
https://dev.to/colingourlay/how-to-support-apple-s-dynamic-text-in-your-web-content-with-css-40c0
Anyway we are overwriting `font-family: sans-serif;`, I did get how `@supports` can help us

gnprice:
Great!

Would you test how this looks on Android? And check that the system font settings do work there (I found they did in the past, but it's a good time to check.)

If this works fine on Android -- meaning it has no effect there -- then great. If it interferes with anything, let's use that `@supports` approach to limit the effect to iOS.

jainkuniya:
So I tested this change on Android (Android 10 running on One plus 6), here are the results:
Summary: It works on Android as well. So we don't require `@supports`

On default font size:


On larger font size:




On small font size:

gnprice:
Thanks!

Did you see my code comment above?

> At the code level: this attribute should appear next to the existing `font-family` attribute.

Also the question I most want answered about the Android behavior is whether this change has any effect. Would you compare how the app looks on Android before and after this change?

jainkuniya:
> At the code level: this attribute should appear next to the existing font-family attribute.

This comment was made when we were applying `font: -apple-system-body;` to `body`, right? 🤔 

In current version it is in `html`
```
html {
  -webkit-user-select: none; /* Safari 3.1+ */
  -moz-user-select: none; /* Firefox 2+ */
  -ms-user-select: none; /* IE 10+ */
  user-select: none; /* Standard syntax */
  -khtml-user-select: none;
  -webkit-touch-callout: none;
  font: -apple-system-body;
}
```

jainkuniya:
This PR vs on master with same font size

gnprice:
> This comment was made when we were applying `font: -apple-system-body;` to `body`, right?

Hmm, yes -- and I see, there's no `font-family` in the block it's now in, so this doesn't make sense anymore. Perhaps I was looking at a previous version of the code.

gnprice:
And merged. Thanks!

Issue #2859, opened 2018-08-03
given action "REALM_INIT", reducer "userGroups" returns undefined


gnprice:
Sentry [shows this error](https://sentry.io/zulip/zulip-mobile/issues/626034816/) hitting a bunch of users, ever since 15.0.92 came out:

> Given action "REALM_INIT", reducer "userGroups" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.

(That is an admirably clear error message. Thanks, Redux!)

Here's the relevant code:
```
const realmInit = (state: UserGroupsState, action: RealmInitAction): UserGroupsState =>
  action.data.realm_user_groups;
```

Seems like that field isn't included in the response we got from the server.

I suspect these are older servers that just don't have this feature. Checked one at random, and it's on 1.7.1:
```
$ curl -s https://[...]/api/v1/server_settings | jq .zulip_version
"1.7.1"
```

We should be compatible, of course, so I guess we need to have this code handle it properly.

borisyankov:
Started working on this. It seems we are not consistent in how we handle this. I'll rework all reducers to do the same reasonable thing in this case.

PR #4073, opened 2020-04-27
api types: `realm_user_groups` is absent prior to v1.8.0.

chrisbobbe:
This bite-size PR is forked from https://github.com/zulip/zulip-mobile/pull/4046#discussion_r414889862.

---------------

api types: `realm_user_groups` is absent prior to v1.8.0.

This was already handled in runtime code, as of d64b7c84b, on
PR #2878, for issue #2859.

On the server, `git log --stat -p --reverse -S realm_user_groups`
reveals that 4c6a376fd, which is v1.8.0-rc1~2711, is the first
occurrence of 'realm_user_groups'.

Also comment on UserGroup type to highlight difference from group
PMs feature.

### DIFF
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -104,7 +104,10 @@ export type InitialDataRealmUser = {|
 |};
 
 export type InitialDataRealmUserGroups = {|
-  realm_user_groups: UserGroup[],
+  /**
+   * Absent in servers prior to v1.8.0-rc1~2711 (or thereabouts).
+   */
+  realm_user_groups?: UserGroup[],
 |};
 
 type NeverSubscribedStream = {|
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -155,6 +155,14 @@ export type CrossRealmBot = {|
  */
 export type UserOrBot = User | CrossRealmBot;
 
+/**
+ * For mentioning multiple users at once.
+ *
+ * When you mention a user group, everyone in the group is notified as
+ * though they were individually mentioned.
+ *
+ * This feature is not related to group PMs.
+ */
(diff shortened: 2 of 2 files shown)


gnprice:
Merged -- thanks!

gnprice:
And yeah, that naming is (perhaps unavoidably) confusing. :-/ Good to clarify.

One other thought: often a helpful way to clarify this kind of thing is a link to the user-facing docs for the feature. Here that's this page:
https://zulipchat.com/help/user-groups

... In fact if I'd thought of that before merging, I would have added it to the new bit of jsdoc. I guess that means I'll go add it now.

chrisbobbe:
> One other thought: often a helpful way to clarify this kind of thing is a link to the user-facing docs for the feature.

Hmm, OK! I did think of doing that — helpfully, right at the top, the user-facing doc confirms my claim in the jsdoc. I didn't add the link, since I think we'd like to encourage/expect that contributors will go to the docs, of their own accord, when they feel they don't know enough about a feature. But I suppose it's nice to pave the way for them to get to the doc, by including a link, and I'm sure I've been guilty of not looking at a doc because I felt the way to it wasn't well-paved. 🙂

gnprice:
> I think we'd like to encourage/expect that contributors will go to the docs, of their own accord, when they feel they don't know enough about a feature.
> 
> But I suppose it's nice to pave the way for them to get to the doc, by including a link

Yeah -- and the other thing this does is, if you're thinking of going to the user docs, it unambiguously identifies exactly what feature we're talking about. This is especially helpful when the terms we use in the code and API aren't the same as in the UI and user-facing docs, which sometimes happens. Here those are pretty well aligned, but a reader might reasonably wonder about that and so it's nice to pin it down explicitly.

PR #4076, opened 2020-04-28
tests: Remove Pirlo.

ray-kraesig:
The Pirlo API is not currently available, and looks to remain that way for some time.


.travis.yml
```diff
@@ -10,10 +10,6 @@ env:
   jobs:
     - TEST_SUITE=''
```
gnprice, line 11: We can also eliminate this bit (and the unquoted `${TEST_SUITE}` below) -- it was introduced in 788758ba8 to support the Pirlo suite.

ray-kraesig: Is that also true for the `${TEST_SUITE} added in 788758ba8db6bf6c5fab98b78c729c71d829b39d?

PR #4078, opened 2020-04-29
Stop using fake blank data at login; and restore #4040

### LINKED ISSUE #4077: App uses fake blank/empty data on login, until data loads from server
I'm about to fix this; filing as an issue to have a nice identifier to refer to it by, and to explain the whole story all in one place.

In general the app's main UI relies on actually having data from the server, because after all without that there's not much useful it could do.

In most situations, we meet that expectation by using the latest data we have from the server, if we have any, and showing the green loading screen (while fetching data in the background) if we don't. In particular, on switching between accounts we show the loading screen.

But currently on login we behave differently: we go straight to the main UI. This means we try to render all the UI of the main tabs -- and potentially more if the user navigates around before the loading completes -- without actually having data for them.

---

The most visible effect of this is that the first thing a user sees on login is a broken-looking screen where the lower-right corner where a tab icon should be is missing/blank. (It's meant to be the user's avatar, and we don't know their avatar URL yet.) We also wrongly say there are no unreads, and no streams if you get that far, etc.

The only reason we get even that far is a sprinkling of hacks that supply fake data when we don't have any. The most immediate is in the realm reducer, providing a user email of `''`. Then `UserAvatar`, which is responsible for that lower-right tab icon, and much of the other avatar-related code it relies on, turns that into an avatar URL of `''` and, if I'm reading right, ultimately equal to the realm URL, and tries to use that as the URL of an image. Barring an elaborate practical joke by the Zulip server admin, the realm URL does not return a valid image and so the `ImageBackground` component remains blank.

As a result, the cascading effect of this is that when we make any changes to the code that cause the main-screen UI to call on data that we *don't* supply fake versions of, logging into the app simply crashes. In particular 79c867166 (#4040) triggered this by switching a selector from using the self-user's email (which we fake) to the corresponding `User` object (which we don't.)

We shouldn't have to tiptoe around what data we can use -- the main UI of the app is all about interacting with a particular Zulip server logged in as a particular account, and its code should be able to rely on the server data for that account being present. Once this is fixed (and modulo any unknown other bugs), there'll be no need for such tiptoeing.

gnprice:
This fixes #4077, which was the root cause of the crash bug brought out by #4040; then restores the changes made originally in #4040.

See background in #4077 and details in commit messages.

### DIFF
--- a/src/realm/realmReducer.js
+++ b/src/realm/realmReducer.js
@@ -35,7 +35,6 @@ export default (state: RealmState = initialState, action: Action): RealmState =>
 
     case REALM_INIT: {
       return {
-        ...state,
         canCreateStreams: action.data.can_create_streams,
         crossRealmBots: action.data.cross_realm_bots,
         email: action.data.email,
--- a/src/realm/realmReducer.js
+++ b/src/realm/realmReducer.js
@@ -31,10 +31,6 @@ const initialState = {
  * `navReducer`) even though we're still loading the actual data from the
  * server.  So we need some fake data that the UI code will swallow.
  * TODO: Probably stop doing that.
- *
- * Also: On `ACCOUNT_SWITCH`, during the transition animation, some old
- * components can still be mounted from the UI for the previous account that
- * make no sense without server data.  Probably ditto `LOGOUT`.
  */
 const fakeBlankState = {
   ...initialState,
--- a/src/realm/realmReducer.js
+++ b/src/realm/realmReducer.js
@@ -12,15 +12,17 @@ import {
 import { objectFromEntries } from '../jsBackport';
 
 const initialState = {
-  canCreateStreams: true,
   crossRealmBots: [],
+
+  nonActiveUsers: [],
+  filters: [],
+  emoji: {},
+
   email: undefined,
   user_id: undefined,
   twentyFourHourTime: false,
-  emoji: {},
-  filters: [],
+  canCreateStreams: true,
   isAdmin: false,
-  nonActiveUsers: [],
 };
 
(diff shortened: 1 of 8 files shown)


src/realm/realmReducer.js
```diff
@@ -35,7 +35,6 @@ export default (state: RealmState = initialState, action: Action): RealmState =>
 
     case REALM_INIT: {
       return {
-        ...state,
```
chrisbobbe, line 57: Since this involves spreading a value and type-checking in a reducer, I was vaguely reminded of [this Flow bug](https://github.com/zulip/zulip-mobile/pull/4046#discussion_r412562505).

I verified empirically that it isn't in play after this commit. (Making a property `undefined` trips Flow.) The next commit helps verify that everything's working as expected by reading the code. So in this case it wasn't necessary to go re-digest what exactly that Flow bug is.

chrisbobbe: On re-reading the Flow issue Ray linked to, it sounds like maybe the bug was happening *before* this commit? Anyway, the important thing is that it's not happening after it. 😆

chrisbobbe: This PR looks good to me, but Ray may have more to say.

ray-kraesig:
> This PR looks good to me, but Ray may have more to say.

Nope! :slightly_smiling_face:  Merging.

chrisbobbe:
@gnprice [said](https://github.com/zulip/zulip-mobile/issues/4077#issue-608723332):
> The most visible effect of this is that the first thing a user sees on login is a broken-looking screen where the lower-right corner where a tab icon should be is missing/blank.

Quite glad to be rid of this! 🙂

PR #4079, opened 2020-04-30
Store server feature level in Redux

### LINKED ISSUE #4049: Store server feature level in Redux
(This isn't ripe to do any work on until the server side has been implemented.)

Per zulip/zulip#14618, we plan to add a "feature level" number to the server's API, which appears alongside `zulip_version` but can be used at a finer grain.

Once it's there, we'll want to observe it on `/server_settings` and `/register`, and maintain it in Redux. This will work just like #3839.

Then we'll want to use it instead of the version number anywhere we can -- basically, for any new server feature that's after 2.1.0. In particular:
* `stream_post_policy`, for #3140 
* as an optimization on how we use zulip/zulip#14500 (discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/view.20uploaded.20files/near/855422))
* new Markdown features as discussed [here](https://chat.zulip.org/#narrow/stream/6-frontend/topic/common.20frontend.20markdown/near/816929)

Priority because of #3140.

hashirsarwar:
Following [#14658](https://github.com/zulip/zulip/pull/14658), we now obtain `feature level` from Zulip server and store it in Redux. This is designed to provide a simple way for mobile apps to decide whether the server supports a given feature or API change.

Closes #4049.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -127,6 +127,8 @@ export const realm = 'https://zulip.example.org';
 
 export const zulipVersion = new ZulipVersion('2.1.0-234-g7c3acf4');
 
+export const zulipFeatureLevel = 1;
+
 export const makeAccount = (
   args: {
     user?: User,
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -473,6 +473,7 @@ export const action = deepFreeze({
       realm_uri: selfAccount.realm,
       realm_video_chat_provider: 1,
       realm_waiting_period_threshold: 3,
+      zulip_feature_level: 1,
       realm_emoji: {},
       realm_filters: [],
       avatar_source: 'G',
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -135,6 +135,7 @@ export const makeAccount = (
     email?: string,
     realm?: string,
     apiKey?: string,
+    zulipFeatureLevel?: number | null,
     zulipVersion?: ZulipVersion | null,
     ackedPushToken?: string | null,
   } = {},
@@ -144,6 +145,7 @@ export const makeAccount = (
     email = user.email,
     realm: realmInner = realm,
     apiKey = randString() + randString(),
+    zulipFeatureLevel: zulipFeatureLevelInner = zulipFeatureLevel,
     zulipVersion: zulipVersionInner = zulipVersion,
     ackedPushToken = null,
   } = args;
@@ -151,6 +153,7 @@ export const makeAccount = (
     realm: realmInner,
     email,
     apiKey,
+    zulipFeatureLevel: zulipFeatureLevelInner,
     zulipVersion: zulipVersionInner,
     ackedPushToken,
   });
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -228,6 +228,7 @@ describe('accountsReducer', () => {
         email: 'newaccount@example.com',
         realm: 'https://new.realm.org',
         zulipVersion: null,
+        zulipFeatureLevel: null,
       });
 
       const action = deepFreeze({
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -26,33 +26,27 @@ describe('accountsReducer', () => {
     });
 
     test('if no account with this realm exists, prepend new one, with empty email/apiKey', () => {
-      const newAccount = eg.makeAccount({
-        realm: 'https://new.realm.org',
-        email: '',
-        apiKey: '',
-      });
+      const newRealm = 'https://new.realm.org';
 
       const action = deepFreeze({
         ...baseAction,
-        realm: newAccount.realm,
+        realm: newRealm,
       });
 
-      expect(accountsReducer(prevState, action)).toEqual([newAccount, account1, account2]);
+      expect(accountsReducer(prevState, action)).toEqual([
+        eg.makeAccount({ realm: newRealm, email: '', apiKey: '' }),
+        account1,
+        account2,
+      ]);
     });
 
     test('if account with this realm exists, move to front of list', () => {
-      const newAccount = eg.makeAccount({
-        realm: account2.realm,
-        email: account2.email,
-        apiKey: '',
-      });
-
       const action = deepFreeze({
         ...baseAction,
-        realm: newAccount.realm,
+        realm: account2.realm,
       });
 
-      expect(accountsReducer(prevState, action)).toEqual([newAccount, account1]);
+      expect(accountsReducer(prevState, action)).toEqual([account2, account1]);
     });
 
     describe('if an account with this realm exists', () => {
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -162,6 +162,28 @@ describe('accountsReducer', () => {
       expect(newState).toEqual(expectedState);
       expect(newState).not.toBe(prevState);
     });
+
+    test('records zulipFeatureLevel on active account', () => {
+      const prevState = deepFreeze([account1, account2, account3]);
+      const newZulipFeatureLevel = 6;
+      const action = deepFreeze({
+        ...eg.action.realm_init,
+        data: {
+          ...eg.action.realm_init.data,
+          zulip_feature_level: newZulipFeatureLevel,
+        },
+      });
+
+      const expectedState = [
+        { ...account1, zulipFeatureLevel: newZulipFeatureLevel },
+        account2,
+        account3,
+      ];
+
+      const newState = accountsReducer(prevState, action);
+
+      expect(newState).toEqual(expectedState);
+    });
   });
 
   describe('ACCOUNT_SWITCH', () => {
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -107,38 +107,26 @@ describe('accountsReducer', () => {
     const account3 = eg.makeAccount();
 
     test('records zulipVersion on active account', () => {
-      const prevState = deepFreeze([account1, account2, account3]);
       const newZulipVersion = new ZulipVersion('2.0.0');
-      const action = deepFreeze({ ...eg.action.realm_init, zulipVersion: newZulipVersion });
-
-      const expectedState = [{ ...account1, zulipVersion: newZulipVersion }, account2, account3];
-
-      const newState = accountsReducer(prevState, action);
-
-      expect(newState).toEqual(expectedState);
-      expect(newState).not.toBe(prevState);
+      expect(
+        accountsReducer(
+          deepFreeze([account1, account2, account3]),
+          deepFreeze({ ...eg.action.realm_init, zulipVersion: newZulipVersion }),
+        ),
+      ).toEqual([{ ...account1, zulipVersion: newZulipVersion }, account2, account3]);
     });
 
     test('records zulipFeatureLevel on active account', () => {
-      const prevState = deepFreeze([account1, account2, account3]);
       const newZulipFeatureLevel = 6;
-      const action = deepFreeze({
-        ...eg.action.realm_init,
-        data: {
-          ...eg.action.realm_init.data,
-          zulip_feature_level: newZulipFeatureLevel,
-        },
-      });
-
-      const expectedState = [
-        { ...account1, zulipFeatureLevel: newZulipFeatureLevel },
-        account2,
-        account3,
-      ];
-
-      const newState = accountsReducer(prevState, action);
-
-      expect(newState).toEqual(expectedState);
+      expect(
+        accountsReducer(
+          deepFreeze([account1, account2, account3]),
+          deepFreeze({
+            ...eg.action.realm_init,
+            data: { ...eg.action.realm_init.data, zulip_feature_level: newZulipFeatureLevel },
+          }),
+        ),
+      ).toEqual([{ ...account1, zulipFeatureLevel: newZulipFeatureLevel }, account2, account3]);
     });
   });
 
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -16,16 +16,15 @@ import * as eg from '../../__tests__/lib/exampleData';
 
 describe('accountsReducer', () => {
   describe('REALM_ADD', () => {
-    const accountWithZulipVersion = eg.makeAccount({
+    const account1 = eg.makeAccount({
       apiKey: '',
       realm: 'https://realm.one.org',
     });
-    const accountWithoutZulipVersion = eg.makeAccount({
+    const account2 = eg.makeAccount({
       apiKey: '',
-      zulipVersion: null,
       realm: 'https://realm.two.org',
     });
-    const prevState = deepFreeze([accountWithZulipVersion, accountWithoutZulipVersion]);
+    const prevState = deepFreeze([account1, account2]);
 
     test('if no account with this realm exists, prepend new one, with empty email/apiKey', () => {
       const newAccount = eg.makeAccount({
@@ -40,18 +39,33 @@ describe('accountsReducer', () => {
         zulipVersion: eg.zulipVersion,
       });
 
-      const expectedState = [newAccount, accountWithZulipVersion, accountWithoutZulipVersion];
+      const expectedState = [newAccount, account1, account2];
 
-      const newState = accountsReducer(prevState, action);
-
-      expect(newState).toEqual(expectedState);
-      expect(newState).not.toBe(prevState);
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
     });
 
     test('if account with this realm exists, move to front of list', () => {
       const newAccount = eg.makeAccount({
-        email: accountWithoutZulipVersion.email,
-        realm: accountWithoutZulipVersion.realm,
+        email: account2.email,
+        realm: account2.realm,
+        apiKey: '',
+      });
+
+      const action = deepFreeze({
+        type: REALM_ADD,
+        realm: newAccount.realm,
+        zulipVersion: eg.zulipVersion,
+      });
+
+      const expectedState = [newAccount, account1];
+
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
+    });
+
+    test('if account with this realm exists and zulipVersion is non-null, move to front of list', () => {
+      const newAccount = eg.makeAccount({
+        email: account2.email,
+        realm: account2.realm,
         apiKey: '',
         zulipVersion: eg.zulipVersion,
       });
@@ -62,12 +76,28 @@ describe('accountsReducer', () => {
         zulipVersion: eg.zulipVersion,
       });
 
-      const expectedState = [newAccount, accountWithZulipVersion];
+      const expectedState = [newAccount, account1];
 
-      const newState = accountsReducer(prevState, action);
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
+    });
 
-      expect(newState).toEqual(expectedState);
-      expect(newState).not.toBe(prevState);
+    test('if account exists and zulipVersion is null, move to front of list and include zulipVersion', () => {
+      const newAccount = eg.makeAccount({
+        email: account2.email,
+        realm: account2.realm,
+        apiKey: '',
+        zulipVersion: null,
+      });
+
+      const action = deepFreeze({
+        type: REALM_ADD,
+        realm: newAccount.realm,
+        zulipVersion: eg.zulipVersion,
+      });
+
+      const expectedState = [account2, account1];
+
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
     });
   });
 
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -36,6 +36,7 @@ describe('accountsReducer', () => {
       const action = deepFreeze({
         type: REALM_ADD,
         realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
         zulipVersion: eg.zulipVersion,
       });
 
@@ -54,6 +55,7 @@ describe('accountsReducer', () => {
       const action = deepFreeze({
         type: REALM_ADD,
         realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
         zulipVersion: eg.zulipVersion,
       });
 
@@ -73,6 +75,7 @@ describe('accountsReducer', () => {
       const action = deepFreeze({
         type: REALM_ADD,
         realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
         zulipVersion: eg.zulipVersion,
       });
 
@@ -92,10 +95,51 @@ describe('accountsReducer', () => {
       const action = deepFreeze({
         type: REALM_ADD,
         realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
         zulipVersion: eg.zulipVersion,
       });
 
       const expectedState = [account2, account1];
+
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
+    });
+
+    test('if no account with this realm exists and account has zulipFeatureLevel, prepend new one, with empty email/apiKey', () => {
+      const newAccount = eg.makeAccount({
+        email: '',
+        apiKey: '',
+        realm: 'https://new.realm.org',
+        zulipFeatureLevel: eg.zulipFeatureLevel,
+      });
+
+      const action = deepFreeze({
+        type: REALM_ADD,
+        realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
+        zulipVersion: eg.zulipVersion,
+      });
+
+      const expectedState = [newAccount, account1, account2];
+
+      expect(accountsReducer(prevState, action)).toEqual(expectedState);
+    });
+
+    test('if account with this realm exists and account has zulipFeatureLevel, move to front of list', () => {
+      const newAccount = eg.makeAccount({
+        email: account2.email,
+        realm: account2.realm,
+        apiKey: '',
+        zulipFeatureLevel: eg.zulipFeatureLevel,
+      });
+
+      const action = deepFreeze({
+        type: REALM_ADD,
+        realm: newAccount.realm,
+        zulipFeatureLevel: eg.zulipFeatureLevel,
+        zulipVersion: eg.zulipVersion,
+      });
+
+      const expectedState = [newAccount, account1];
 
       expect(accountsReducer(prevState, action)).toEqual(expectedState);
     });
--- a/src/account/__tests__/accountsReducer-test.js
+++ b/src/account/__tests__/accountsReducer-test.js
@@ -16,52 +16,43 @@ import * as eg from '../../__tests__/lib/exampleData';
 
 describe('accountsReducer', () => {
   describe('REALM_ADD', () => {
-    const account1 = eg.makeAccount({
-      apiKey: '',
-      realm: 'https://realm.one.org',
-    });
-    const account2 = eg.makeAccount({
-      apiKey: '',
-      realm: 'https://realm.two.org',
-    });
+    const account1 = eg.makeAccount({ realm: 'https://realm.one.org', apiKey: '' });
+    const account2 = eg.makeAccount({ realm: 'https://realm.two.org', apiKey: '' });
     const prevState = deepFreeze([account1, account2]);
+    const baseAction = deepFreeze({
+      type: REALM_ADD,
+      zulipFeatureLevel: eg.zulipFeatureLevel,
+      zulipVersion: eg.zulipVersion,
+    });
 
     test('if no account with this realm exists, prepend new one, with empty email/apiKey', () => {
       const newAccount = eg.makeAccount({
+        realm: 'https://new.realm.org',
         email: '',
         apiKey: '',
-        realm: 'https://new.realm.org',
       });
 
       const action = deepFreeze({
-        type: REALM_ADD,
+        ...baseAction,
         realm: newAccount.realm,
-        zulipFeatureLevel: eg.zulipFeatureLevel,
-        zulipVersion: eg.zulipVersion,
       });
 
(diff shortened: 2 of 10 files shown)


hashirsarwar:
@gnprice @chrisbobbe this is ready for a review. :)

chrisbobbe:
Thanks, @hashirsarwar!

Just a note to @gnprice that `zulipFeatureLevel: number | null`, on `Account`, instead of `zulipFeatureLevel?: number`, was on my advice, based on https://github.com/zulip/zulip-mobile/pull/4047/commits/e026c6a46d1e9e9846febbf68419544962bebb9e (in PR #4047, currently a draft) — so if this is wrong, please let me know too. 🙂

src/__tests__/lib/exampleData.js
```diff
@@ -114,6 +114,8 @@ export const realm = 'https://zulip.example.org';
 
 export const zulipVersion = '2.1.0-234-g7c3acf4';
 
+export const zulipFeatureLevel = 100;
```
chrisbobbe, line 117: Nit in commit message for

e3ca22d27 RealmScreen: Include feature level from /server_settings in REALM_ADD.

"accessed"

src/__tests__/lib/exampleData.js
```diff
@@ -461,6 +461,7 @@ export const action = deepFreeze({
       },
       user_status: {},
     },
+    zulipFeatureLevel,
```
chrisbobbe, line 467: Nit in commit message for 

16b81678a realmActions: Include feature level in REALM_INIT.

"already"

src/__tests__/lib/exampleData.js
```diff
@@ -122,6 +122,7 @@ export const makeAccount = (
     email?: string,
     realm?: string,
     apiKey?: string,
+    zulipFeatureLevel?: number | null,
```
chrisbobbe, line 125: Nits in commit message for 

f514d6a7a Account type: Add a property featureLevel.

- Let's be consistent with the name we use as an identifier in the code: `zulipFeatureLevel`. Setting it off with \`backticks\` makes it quite clear (in addition to the camelCase capitalization) that we're referring to an identifier. There may be other places where this comment applies.
- Let's specify that we're talking about `loginSuccess` and `realmAdd` in the `accountsReducer`, not the functions with the same names in `accountActions`. It could be inferred from context and by looking at the diff, but it's nice to help the reader avoid ambiguity.

hashirsarwar:
@chrisbobbe thanks for the review.

As far as commit messages are concerned, I think we use imperative mood in summaries. In case of message bodies, there is no as such restriction. See the commit message example [here](https://github.com/zulip/zulip/commit/cd5b38f5d8bdcc1771ad794f37262a61843c56c0) provided in the documentation.

However, I'll make the requested changes soon. :slightly_smiling_face:

chrisbobbe:
> As far as commit messages are concerned, I think we use imperative mood in summaries. In case of message bodies, there is no as such restriction. See the commit message example [here](https://github.com/zulip/zulip/commit/cd5b38f5d8bdcc1771ad794f37262a61843c56c0) provided in the documentation.

Hmm, interesting. I would have thought the distinction is made on a different axis — not whether the words are in the summary or in the body, but rather, whether they describe some work that gets done in the commit, as opposed to background information, motivations for the work, plans for the future, maybe even close or distant effects of the work, etc.

But the [doc](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages) does indeed mention "imperative" several times in the section about the summary, and zero times in the section about the body.

@gnprice, @ray-kraesig, thoughts?

hashirsarwar:
@chrisbobbe here are some other examples of recent commit messages from the web repository.

Screenshot from 2020-05-01 00-29-13

Still, I would like to know what others think about it. :slightly_smiling_face: 
Thanks once again!

timabbott:
That's correct, imperative mood is only expected in the summary line.  The paragraph body should just be a plain English paragraph explaining relevant background for the change.

chrisbobbe:
Hmm, the commits in [that screenshot](https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622063597) also seem to follow both patterns. 🙂

@timabbott [said](https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622066301):
> That's correct, imperative mood is only expected in the summary line. The paragraph body is a more normal paragraph explaining relevant background/motivation/etc. for the change.

Hmm, OK. @timabbott, it sounds like there's an additional constraint that the entire work of the change (excluding background, motivation, etc.) be condensed into the summary line. With the summary line's character limit, might it sometimes be impossible to preserve the required amount of detail in that class of explanation, even with minimal, coherent commits? When follow-on changes on a main change are needed in the same commit, they often deserve their own explanation, right, which I'm not sure would count as background info?

Anyway, @hashirsarwar, for the commit I mentioned [here](https://github.com/zulip/zulip-mobile/pull/4079#pullrequestreview-403723583):

f514d6a7a Account type: Add a property featureLevel.

That bit I highlighted from the body, "This adds a new property featureLevel in the type Account.", seems redundant; you've got it (in the imperative, good 🙂) in the summary line.

src/types.js
```diff
@@ -94,6 +94,20 @@ export type Account = {|
    */
   zulipVersion?: string,
 
+  /**
+   * An integer indicatating what features are available on the server.
```
chrisbobbe, line 98: indicating

gnprice:
> The paragraph body is a more normal paragraph explaining relevant background/motivation/etc. for the change.

I think if you look again you'll agree that this is a stronger statement than you really intend. Among your own recent commit messages, it's not uncommon to have some "how" or high-level "what" in the commit message body, in addition to "why" and background:

    While working on this, I fixed various issues where feature levels
    could be mentioned or endpoints didn't properly document changes.

    [...]  We fix this inconsistency by making
    the database model case-insensitive.

    Fix this by ensuring all the files we generate are in this special
    subdirectory.

    Also improve the headings for it.

and I think that's for good reasons -- sometimes there's just more to say than makes sense to put in the summary line.


> imperative mood is only expected in the summary line

It looks like the practice on this is mixed. At a quick survey of recent zulip.git history:
* As quoted above, your messages sometimes put the "what" in the same syntactic form we use in summary lines, and sometimes in the plain indicative like we do for background in the bodies.
* Mine generally have the "what" match the summary line.
* Anders tends to write shorter messages; when they do include some "what", it matches the summary line.
* Steve includes some "what" in the body more frequently, and consistently has it in the indicative, like the background in the body.

In the mobile repo, we've more or less consistently used the same form for describing "what" in the body (if present there) as we do in the summary line.

I don't feel an urgency to make the style uniform one way or another in zulip.git, but unsurprisingly I do like better the style I use. :wink: One reason is that a sentence in the indicative can easily be ambiguous about whether it's describing something already true before this commit (as background) or something newly made true by this commit.

In any case we've been pretty consistent in zulip-mobile, and so I'd like to generally maintain that consistency there.

chrisbobbe:
@hashirsarwar, I like these changes!

@gnprice and/or @ray-kraesig will likely have more to add. @gnprice, there's a query [here](https://github.com/zulip/zulip-mobile/issues/4049#issuecomment-622082621), which perhaps I should have posted on the PR thread, and in any case will quote here:

> @gnprice [said](https://github.com/zulip/zulip-mobile/issues/4049#issue-602308429):
> > Then we'll want to use it instead of the version number anywhere we can -- basically, for any new server feature that's after 2.1.0.
> 
> @gnprice, do we know the particular minimum server version that will have this flag? The above implies that such a minimum server version exists, but it's not 2.1.0. Maybe 2.1.1?
> 
> Just so we can put the right comment on `zulip_feature_level` in `ApiResponseServerSettings`.

src/api/settings/getServerSettings.js
```diff
@@ -34,6 +34,7 @@ export type ApiResponseServerSettings = {|
   realm_name: string,
   realm_uri: string,
   require_email_format_usernames: boolean,
+  zulip_feature_level: number,
```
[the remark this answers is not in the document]
gnprice: I think in this case if the server's response has no `zulip_feature_level`, I'd like to translate that to a value of 0 in the `Account` object. The reason is that if the server doesn't send a `zulip_feature_level`, that means it predates the introduction of that property -- which can be treated implicitly as one very broad feature level of 0, preceding the first explicit feature level which is 1.

We'll still need the type to be `number | null`, I think, to handle accounts that the user hasn't looked at since upgrading to an app version that stores feature levels.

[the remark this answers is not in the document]
gnprice: See docs at https://zulipchat.com/api/server-settings :
> New in Zulip 2.2. We recommend using an implied value of 0 for Zulip servers that do not send this field.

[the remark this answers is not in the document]
chrisbobbe: Hmm, that does make sense! So, I think that would mean the `REALM_INIT` and `REALM_ADD` actions have `zulipFeatureLevel: number`, while the `Account` still has `number | null`, right?

gnprice:
> do we know the particular minimum server version that will have this flag?

Ah, my wording here was confusing:

> basically, for any new server feature that's after 2.1.0.

In general new features don't appear in a new Zulip patch release, i.e. a version a.b.(c+1) vs. a.b.c. Rather they only appear in a new major release, which is a.(b+1).0 or (a+1).0.0. So a new server feature that's after 2.1.0 means one whose first release is 2.2.0. (Or 3.0.0 if we were to go straight there, but we won't.)

---

For this purpose, though:

> Just so we can put the right comment on `zulip_feature_level` in `ApiResponseServerSettings`.

I think the answer is:
* In the past (and in the future, for server features introduced in the past), we'd identify a zulip.git commit and the release it appeared in.
* Thanks to this very feature, and to improvements in the Zulip API documentation, I think we can shift to something that's both easier and usually more helpful!

  Specifically, this can say something like "Added in server version 2.2, feature level 1."

  And there's already a link in that file to https://zulipchat.com/api/server-settings , which documents that fact. (I guess in this instance it's not 100% clear about the "feature level 1" detail -- but in general the API docs will now say things like "New in Zulip 2.2 (feature level 2)" or whatever.)

src/message/fetchActions.js
```diff
@@ -249,7 +249,13 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
     return;
   }
 
-  dispatch(realmInit(initData, serverSettings.zulip_version));
+  dispatch(
+    realmInit(
+      initData,
+      serverSettings.zulip_feature_level === undefined ? null : serverSettings.zulip_feature_level,
```
gnprice, line 255: Instead of getting this from `serverSettings`, we can actually get the `zulip_feature_level` from `initData`. (Which also means this line and `realmInit` don't need to change, and the reducer can handle it.)

That's preferable because for recent servers we can also get `zulip_version` from there. We make this `getServerSettings` call here because older servers send `zulip_version` only there; but `zulip_feature_level` appears on the same set of servers in one place as the other place.

This isn't well documented, sorry -- not your fault that you didn't spot that. The initial data from `/register` is one of the couple of main areas we haven't yet covered in our API documentation.

src/api/settings/getServerSettings.js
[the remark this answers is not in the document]
gnprice: Yep, that sounds right!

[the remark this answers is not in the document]
gnprice: Err, actually `REALM_ADD` will, as it's based on `server_settings` -- but `REALM_INIT` only needs the initial data from `/register` (its `realm` property), per my other comment just now.

src/message/fetchActions.js
chrisbobbe: > Instead of getting this from `serverSettings`, we can actually get the `zulip_feature_level` from `initData`. (Which also means this line and `realmInit` don't need to change, and the reducer can handle it.)

Mm, nice! Perhaps the type `InitialDataRealm`, or even `InitialDataBase`, might be a place to add `zulip_feature_level?: number`, to make `initData` well-typed?

src/__tests__/lib/exampleData.js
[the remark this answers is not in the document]
hashirsarwar: @ray-kraesig thanks for the feedback. I've updated the PR.

src/api/initialDataTypes.js
```diff
@@ -78,6 +78,13 @@ export type InitialDataRealm = {|
   realm_uri: string,
   realm_video_chat_provider: string,
   realm_waiting_period_threshold: number,
+  /**
```
gnprice, line 81: nit: blank line to separate comment from item above (and ditto in `getServerSettings.js`)

src/api/settings/getServerSettings.js
```diff
@@ -35,6 +35,10 @@ export type ApiResponseServerSettings = {|
   realm_uri: string,
   require_email_format_usernames: boolean,
   zulip_version: string,
+  // zulip_feature_level added for server v2.2, feature level 1
+  // See https://github.com/zulip/zulip/commit/e3b90a5ec8 for
+  // the server-side implementation.
```
gnprice, line 40: Instead of the implementation, we can link to the docs:
  https://zulipchat.com/api/server-settings

Implementation works as a substitute when there are no docs, but docs are better when available.

src/api/initialDataTypes.js
```diff
@@ -78,6 +78,13 @@ export type InitialDataRealm = {|
...
+   * Added in server version 2.2, feature level 1. Older servers
+   * don't send this.
```
gnprice, line 83: nit: "Older servers don't send this" is redundant -- "Added in ..." covers that information

src/api/initialDataTypes.js
```diff
@@ -78,6 +78,13 @@ export type InitialDataRealm = {|
...
+   * See https://github.com/zulip/zulip/commit/e3b90a5ec8 for the
+   * server-side implementation.
```
gnprice, line 85: This one isn't strictly documented (the initial data is pretty undocumented in general); but this field is exactly like the corresponding one in `/server_settings`. So I think the cleanest thing is to say that -- like
> Same meaning as in the `server_settings` response:
>   https://zulipchat.com/api/server-settings

src/api/settings/getServerSettings.js
```diff
@@ -35,6 +35,10 @@ export type ApiResponseServerSettings = {|
...
+  zulip_feature_level?: number,
```
gnprice, line 41: This change and the one in `initialDataTypes` are best squashed -- they're two manifestations of the same thing.

A good prefix for the commit-message summary would be `api types:`.

src/start/RealmScreen.js
```diff
@@ -55,7 +55,15 @@ class RealmScreen extends PureComponent<Props, State> {
 
     try {
       const serverSettings: ApiResponseServerSettings = await api.getServerSettings(realm);
-      dispatch(realmAdd(realm, new ZulipVersion(serverSettings.zulip_version)));
+      dispatch(
+        realmAdd(
+          realm,
+          serverSettings.zulip_feature_level === undefined
+            ? null
```
gnprice, line 62: See discussion above: https://github.com/zulip/zulip-mobile/pull/4079#discussion_r418365612

This can be 0 instead of null, and then the action type can have just `zulipFeatureLevel: number`.

src/account/accountsReducer.js
```diff
@@ -32,6 +33,7 @@ const realmAdd = (state, action) => {
       apiKey: '',
       email: '',
       ackedPushToken: null,
+      zulipFeatureLevel: action.zulipFeatureLevel,
```
gnprice, line 36: An intermediate commit makes this always `null`, and a fixup commit at the end fills in the proper logic. Instead it can get the proper logic to begin with -- just squash the fixup commit into the other one.

src/types.js
```diff
@@ -95,6 +95,20 @@ export type Account = {|
...
+   * An integer indicating what features are available on the server.
+   *
+   * The feature level increases monotonically; a value of N means the
+   * server supports all API features introduced before feature level N.
+   * This is designed to provide a simple way for mobile apps to decide
+   * whether the server supports a given feature or API change.
+   *
+   * Like zulipVersion, we learn the feature level from /server_settings
+   * at the start of the login process, and again from /register when
+   * setting up an event queue.
```
gnprice, line 108: This is what we should do, but we don't in this version :wink: 

Look at where `zulipVersion` gets set in `accountsReducer.js`. One of those is missing the corresponding logic for `zulipFeatureLevel`.

hashirsarwar:
@gnprice thanks for the feedback. Can you review this once again?

src/api/initialDataTypes.js
gnprice: bump -- same thing in `getServerSettings.js`

src/types.js
gnprice: This is still missing. Without it, this PR really doesn't work properly at all.

Search for `zulipVersion` in `accountsReducer.js`. Notice the mention in the `realmInit` function. That's how we implement this part:
> and again from /register when setting up an event queue.

We need to do the same for `zulipFeatureLevel`.

gnprice:
Thanks for the revisions! Comments as replies in threads above.

chrisbobbe:
Hi @hashirsarwar! It looks like there's a `prettier` failure being caught in CI.

src/types.js
```diff
@@ -95,6 +95,20 @@ export type Account = {|
...
+   */
+  zulipFeatureLevel: number | null,
```
chrisbobbe, line 110: **edit**: Hmm, I'm actually not sure about what I say below the line (below); @gnprice said, at https://github.com/zulip/zulip-mobile/pull/4079#discussion_r418363632, that the `null` does serve a purpose:

> We'll still need the type to be `number | null`, I think, to handle accounts that the user hasn't looked at since upgrading to an app version that stores feature levels.

I think maybe it's clearer if we do forget about `null`, and just make a migration that sets it to `0` if it's not there? Maybe not, or maybe I'm missing something. 🙂 What do you think, Greg?

-----

This should also be just `number`, to add to the changes you've made in response to https://github.com/zulip/zulip-mobile/pull/4079#discussion_r421867948.

Correspondingly, the `zulipFeatureLevel` param in `makeAccount` in exampleData.js should be `zulipFeatureLevel?: number`. The `?` is so that we can call `makeAccount` without deciding what `zulipFeatureLevel` we want; the implementation can pick one.

src/account/accountsReducer.js
```diff
@@ -41,6 +43,7 @@ const realmAdd = (state, action) => {
 const realmInit = (state, action) => [
   {
     ...state[0],
+    zulipFeatureLevel: state[0].zulipFeatureLevel,
```
chrisbobbe, line 46: Instead of getting the `zulipFeatureLevel` from whatever the state was before this action, we should grab it from the new data that the action includes. In particular, I believe, `action.data.zulip_feature_level`.

Since `action.data` comes from the server data without really being transformed, `zulip_feature_level` might be undefined because the server doesn't provide it. So, we'll want to do, e.g., `action.data.zulip_feature_level ?? 0`.

src/types.js
chrisbobbe: **edit**: See above.

A few places still have `zulipFeatureLevel: null` being set explicitly (you can search for these); they should be replaced with `zulipFeatureLevel: 0`.

gnprice: > I think maybe it's clearer if we do forget about `null`, and just make a migration that sets it to `0` if it's not there?

The thing about that is that it might not be true -- the server might actually be on a later version, and we just haven't talked to it since upgrading the app to a version that stores the feature level.

In particular, anyone using the app now with chat.zulip.org or zulip.com has accounts where we've recorded that the version is 3.0-rc1-something. There is no server (unless a willfully confusing one) that has that version number and a feature level of 0.

chrisbobbe: Ah, right—thanks for reminding me about that possibility. So my comments here should be disregarded, except:

I think maybe we'll need a migration that sets it to `null` (otherwise it'll be missing altogether)? Something like:

```javascript
  // Accounts.zulipFeatureLevel is `number | null`
   '14': state => ({
    ...state,
    accounts: state.accounts.map(a => ({
      ...a,
      zulipFeatureLevel: a.zulipFeatureLevel !== undefined ? a.zulipFeatureLevel : null,
    })),
  }),
```

hashirsarwar:
Hi @chrisbobbe, I have updated the PR. Kindly have a look.

chrisbobbe:
Thanks, @hashirsarwar!

@gnprice, the migration I recommended at https://github.com/zulip/zulip-mobile/pull/4079#discussion_r452366955 is correct, I think, right?

Another thing that just occurred to me is that it would be good to have some tests in `accountsReducer-test.js` that specifically make sure the `zulipFeatureLevel` gets stored when we want it to be.

One bit of trouble is that the current tests for `zulipVersion` aren't a very clear model for what to do about `zulipFeatureLevel`; sorry about that:

- I've even got a PR to remove one: https://github.com/zulip/zulip-mobile/pull/4198.
- For `'REALM_INIT' -> 'records zulipVersion on active account'`, I could imagine a similar one like `'REALM_INIT' -> 'records zulipFeatureLevel on active account'`, just that the shape is different (the `zulipFeatureLevel` is in `action.data`, not `action`).
- For `REALM_ADD`... it looks like the existing tests aren't really clear and coherent; sorry. I think I was trying to reduce repetition. There are two relevant axes, it seems: (A) whether "an account with this realm exists", and (B) whether the affected account (i.e., the new or existing account) has a `zulipVersion` or not. So, properly, there should be 2 * 2 = 4 tests, I suppose, but I didn't quite get there. Throw in (C) whether the affected account has `zulipFeatureLevel` or not, and I guess it's 2 * 2 * 2 = 8 tests. That seems...excessive. But I'm wary of compromising, as I seem to have done with the existing tests, and left them in this confusing state. @gnprice, do you see a nice balance somewhere?

src/types.js
gnprice: Yeah, that looks right. We do need a migration, or else the `zulipFeatureLevel` property would be simply missing (contrary to what the type says.)

The code can be tightened a bit by using `??`:
```
  zulipFeatureLevel: a.zulipFeatureLevel ?? null
```

... Actually, better: it's impossible for `zulipFeatureLevel` to be present at this point. This migration runs only when we're processing data that was stored by an older version of the app that didn't know anything about `zulipFeatureLevel` here. So we can just say:
```
  zulipFeatureLevel: null
```
and that's clearer about about what's happening because it doesn't try to handle a situation that should be impossible.

(Looking back at previous migrations, I see we did the redundant-conditional thing with the last two, oops.)

src/boot/store.js
```diff
@@ -163,6 +163,15 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
         typeof a.zulipVersion === 'string' ? new ZulipVersion(a.zulipVersion) : a.zulipVersion,
     })),
   }),
+
+  // Accounts.zulipFeatureLevel is `number | null`
+  '14': state => ({
+    ...state,
+    accounts: state.accounts.map(a => ({
+      ...a,
+      zulipFeatureLevel: a.zulipFeatureLevel !== undefined ? a.zulipFeatureLevel : null,
```
gnprice, line 172: As mentioned at https://github.com/zulip/zulip-mobile/pull/4079#discussion_r459144052 , this can just be:
```suggestion
      zulipFeatureLevel: null,
```

src/boot/store.js
```diff
@@ -163,6 +163,15 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
         typeof a.zulipVersion === 'string' ? new ZulipVersion(a.zulipVersion) : a.zulipVersion,
     })),
   }),
+
+  // Accounts.zulipFeatureLevel is `number | null`
```
gnprice, line 167: One thing that would help in these comments is that it's just as helpful (or almost as helpful) to know the *old* status of the bit of data in question as the new status.

So e.g. for this one: "Added Accounts.zulipFeatureLevel as `number | null`"

gnprice:
> * For `REALM_ADD`... it looks like the existing tests aren't really clear and coherent; sorry. I think I was trying to reduce repetition. There are two relevant axes, it seems: (A) whether "an account with this realm exists", and (B) whether the affected account (i.e., the new or existing account) has a `zulipVersion` or not. So, properly, there should be 2 * 2 = 4 tests, I suppose, but I didn't quite get there. Throw in (C) whether the affected account has `zulipFeatureLevel` or not, and I guess it's 2 * 2 * 2 = 8 tests. That seems...excessive. But I'm wary of compromising, as I seem to have done with the existing tests, and left them in this confusing state. @gnprice, do you see a nice balance somewhere?

One thing that tends to help is to make each test case very short and focused. (Where possible -- and I think it should be possible here.) When each one is just a few lines, tests one thing, and it's clear what that one thing is, that helps make it comfortable to have a number of different test cases and really think systematically through having them cover all the interesting different situations.

One step that helps in that is to cut out unnecessary lines (and names, and steps) in each test case. The `.not.toBe(prevState)` is redundant -- prevState is frozen, so it can't have been mutated, so if the new state *were* prevState then it couldn't have been equal to expectedState, which we checked on the previous line. Then there's no need to give a name to `newState`, which saves a couple of lines, or to `expectedState`, which saves another line or two.

Another step that helps is to have separate, small, test cases for separate concerns. The existing test cases have descriptions that make them about how the accounts array gets reordered, and how that interacts with the kind of subtle way that we fill in the email and API key. Then they also at the same time do some testing of what happens to the `zulipVersion`. It's a bit hard to tell what they're really testing there -- effectively it's a lot like they're not really consciously testing it, and just have `zulipVersion` properties present to keep the tests and the type-checker passing. (Consider what would happen if someone made a change that broke the `zulipVersion` aspect of these tests -- it's hard to tell what they're testing in that respect, so it's quite likely the person would just treat it as one of the situations where a test is too sensitive to irrelevant details, and update the test to match the new behavior.) Instead, there could be:
* a couple of test cases that test what the existing two test descriptions say, and without doing anything in particular about `zulipVersion` -- these can be shorter by just having an account or two on different realms and not mentioning `zulipVersion`
* then a couple more test cases that are specifically about `zulipVersion` -- which can be shorter by just having one account
* then a couple more test cases that are specifically about `zulipFeatureLevel`, the same way.

chrisbobbe:
@gnprice [said](https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-662756156):
> Instead, there could be: [...]

Ah OK, makes sense. I guess I was too strongly thinking that the one true, pedantically correct thing to do, for full coverage, would be these 8 tests:

- account exists in state; account is without `zulipVersion` and without `zulipFeatureLevel`
- account exists in state; account is with `zulipVersion` and without `zulipFeatureLevel`
- account exists in state; account is without `zulipVersion` and with `zulipFeatureLevel`
- account exists in state; account is with `zulipVersion` and with `zulipFeatureLevel`
- account is not in state; account is without `zulipVersion` and without `zulipFeatureLevel`
- account is not in state; account is with `zulipVersion` and without `zulipFeatureLevel`
- account is not in state; account is without `zulipVersion` and with `zulipFeatureLevel`
- account is not in state; account is with `zulipVersion` and with `zulipFeatureLevel`

and that anything less is unprincipled. But seeing them all listed out like that (and imagining the list grow exponentially with each new field we might ever add to Account!), I think, shows that it distracts from what we really want to test.

Your bullet points there make sense. To be more specific about the second and third, though, might the full set of tests look like this:

- The two we already have:
  - account exists in state
  - account is not in state

- a couple about `zulipVersion`:
  - account exists in state; account has `zulipVersion`
  - account is not in state; account has `zulipVersion`

- a couple about `zulipFeatureLevel`:
  - account exists in state; account has `zulipFeatureLevel`
  - account is not in state; account has `zulipFeatureLevel`

That's...six test cases, instead of eight. Nice that we're not growing exponentially with each new field, though. Do you think the repetition of "account exists / is not in state" is unnecessary? If so, do you have something else in mind by "a couple more," which I take as meaning there should be multiple for `zulipVersion` and multiple for `zulipFeatureLevel` (instead of just one each)?

gnprice:
> That's...six test cases, instead of eight. Nice that we're not growing exponentially with each new field, though.

Yeah. The other aspect is that each one is likely to be smaller than the existing cases, because it sets up a less complicated situation. Ideally each one should be just a very few lines.


> * a couple about `zulipVersion`:
>   * account exists in state; account has `zulipVersion`
>   * account is not in state; account has `zulipVersion`

I was thinking of
* account exists in state, and has non-null `zulipVersion`
* account exists in state, and has null `zulipVersion`

The case where the account doesn't exist in the state can I think be covered by the "account is not in state" test case in the first pair, because that can just do an equality test on the whole resulting state.

chrisbobbe:
Hi, @hashirsarwar! Do you have a moment to come back to this?

I think the remaining things are a few tweaks to the migration, as Greg mentions at https://github.com/zulip/zulip-mobile/pull/4079#pullrequestreview-453757154, then Greg has described a way forward for the tests.

src/account/__tests__/accountsReducer-test.js
```diff
@@ -16,16 +16,15 @@ import * as eg from '../../__tests__/lib/exampleData';
 
 describe('accountsReducer', () => {
   describe('REALM_ADD', () => {
-    const accountWithZulipVersion = eg.makeAccount({
+    const account1 = eg.makeAccount({
```
hashirsarwar, line 19: I think changing `accountWithZulipVersion/accountWithoutZulipVersion` to a more general name `account1/account2` is better in terms of readability and coherence. Please tell me if you don't like this idea. :slightly_smiling_face:

hashirsarwar:
@chrisbobbe please have a look.

src/account/__tests__/accountsReducer-test.js
chrisbobbe: I like it!

chrisbobbe:
Thanks for the revisions! Answering here instead of in PMs, so others can see 🙂.

Yes, it would be good if the "make existing tests more coherent" part were in a different commit than the commit where the tests are added. Might be helpful to put the clarifying commit before the commit that adds tests. If I'm cleaning things up, I often do so before adding new things, if it makes sense.

The commit where the tests are added should be the same commit that makes them necessary. That looks like it's

c1a861748 accountsReducer: Store server feature level in Redux.

Also, in your PM, you linked to one of my comments, https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-662772504, but Greg made some adjustments in https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-662826178, which means the tests should look slightly different from what I said. 🙂

chrisbobbe:
Thanks, @hashirsarwar! I've made a few tweaks as commits on top of this branch, and I pinged @gnprice on CZO to double-check them.

gnprice:
Thanks @chrisbobbe ! These look good. Merged, with a few additional commits on top; take a look.

chrisbobbe:
> Thanks @chrisbobbe ! These look good. Merged, with a few additional commits on top; take a look.

Thanks, and thanks for those additional commits, I read them all! 🙂

Issue #4081, opened 2020-05-01
Start using proper URL objects instead of string concatenation


chrisbobbe:
~~EDIT: Blocked on #3781, the RN 0.61 upgrade, unfortunately; see my comments here, up until https://github.com/zulip/zulip-mobile/issues/4081#issuecomment-622606771, below.~~ Looks like this is unblocked; see https://github.com/zulip/zulip-mobile/issues/4081#issuecomment-623716266 for a way to avoid depending on a file newly introduced in RN v0.61.

-----

Arising especially from https://github.com/zulip/zulip-mobile/pull/3819#issuecomment-576487586, https://github.com/zulip/zulip-mobile/pull/3819#issuecomment-578293246, and https://github.com/zulip/zulip-mobile/issues/3567#issuecomment-622575914, which all deal with something going wrong with constructing URLs, here's an issue for using something more reliable and consistent than string concatenation and various regexes.

Issues like https://github.com/zulip/zulip-mobile/issues/4058 may benefit from this too?

#4082 is a workaround that has our first TODO pointing to the eventual resolution of this issue.

It looks like React Native has a polyfill for the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) Web API, but it's been cut down to reduce JS bundle size. https://github.com/charpeni/react-native-url-polyfill#why-do-we-need-this is a more complete polyfill (it says it would add 73.41 KB to the bundle), and it lists some bugs that people have run into with React Native core's polyfill.

chrisbobbe:
Ha — in fact, a React Native maintainer closed an issue, at https://github.com/facebook/react-native/issues/16434#issuecomment-580094423, citing `react-native-url-polyfill` as (apparently) the official workaround.

chrisbobbe:
I think https://github.com/facebook/react-native/issues/25717, and probably others, listed there, is enough to say that we shouldn't use React Native's polyfill.

Consider the bug in https://github.com/zulip/zulip-mobile/issues/3567#issuecomment-622574435:

```
webAuth.openBrowser(`${this.props.realm}/${url}`, otp);
```

incorrectly doubles a slash if `url` (which is a `.login_url` on an item in `.external_authentication_methods` from `/server_settings`) starts with a slash:

```
https://chat.zulip.org//accounts/login/social/google
```

All the ones I've seen do, but we shouldn't expect that they always will. If it's missing, then https://github.com/facebook/react-native/issues/25717 means we'll get

```
[...]

PR #4082, opened 2020-05-02
Deduplicate slashes in social auth login endpoints

chrisbobbe:
This fixes an issue noted in https://github.com/zulip/zulip-mobile/issues/3567#issuecomment-622574435 where we're duplicating slashes when we call social auth endpoints.

A better fix depends on https://github.com/zulip/zulip-mobile/issues/4081. The commit messages reflect that issue being blocked by the RN v0.61 upgrade, but Ray points out that we may be able to do it without the available JS library, and instead make a native plugin, in which case the issue is not blocked; discussion is ongoing.

### DIFF
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -204,7 +204,10 @@ class AuthScreen extends PureComponent<Props> {
    */
   beginWebAuth = async (url: string) => {
     otp = await webAuth.generateOtp();
-    webAuth.openBrowser(`${this.props.realm}/${url}`, otp);
+    // TODO: Following #4081, use, e.g.,
+    //   `new URL(url, this.props.realm)`
+    const absoluteURL = `${this.props.realm}/${url.replace(/^\/+/, '')}`;
+    webAuth.openBrowser(absoluteURL, otp);
   };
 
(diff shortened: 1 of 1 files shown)


ray-kraesig:
The PR text and comment say RN v0.61, but the commit message says RN v0.62.

Other than that, LGTM.

chrisbobbe:
@ray-kraesig [said](https://github.com/zulip/zulip-mobile/pull/4082#issuecomment-623650592):
> The PR text and comment say RN v0.61, but the commit message says RN v0.62.
> 
> Other than that, LGTM.

Good catch. 😆

I've removed both mentions, following our new knowledge that it isn't actually blocked at all.

ray-kraesig:
I'm a bit behind on PRs, I'm afraid; this one's fallen through. ... much like #3819 did, which seems to have fixed the same bug. I'll be happy to merge the JSDoc, of course; do we still want the first commit?

chrisbobbe:
> which seems to have fixed the same bug

They're similar but different bugs; #3819 was for trailing slashes on the realm URL; this is for leading slashes in the login URL. #4081 will much reduce our woes in both places (and many more), but for now, I think the fix in this PR would be good to have.

Issue #3864, opened 2020-02-03
Distinguish API errors in Sentry


gnprice:
In our Sentry logs, the [most frequent "issue"](https://sentry.io/organizations/zulip/issues/1275118205) by far is:
> **Error** ApiError(src/api/apiErrors)

Flipping through the events that make it up, they turn out to be a wide variety of different errors:
* Bad value for 'to': [30] (on POST `/typing`; [event](https://sentry.io/organizations/zulip/issues/1275118205/events/359eee6c9b30470d9ac2b7ad61bbcf7a))
* Topic can't be empty (on POST `/messages`, with e.g. `subject=%20` -- so #3743)
* Zulip server auth failure: f2f0[... a UUID] is not registered (on POST `/users/me/android_gcm_reg_id`; [event](https://sentry.io/organizations/zulip/issues/1275118205/events/4d4c58efe832446b999e7112012cdecb))
* Cannot send to multiple streams (on POST `/messages`; [event](https://sentry.io/organizations/zulip/issues/1275118205/events/e0913567d2174379ab07423b742cfdb3))
* ... that covers the last few dozen of these events, but there are probably other errors that are just a bit less frequent.

Confusingly, the main dashboard, which shows one big row per "issue", includes the error message from the one most recent event, making it look like that one error is generating that many events, when in fact it's all these errors combined.

I'm not sure exactly how Sentry is deciding to aggregate these. In addition to the "**Error** ApiError(src/api/apiErrors)" part which they all have in common, they also all have the same stack trace, so that may be a factor; it's
```
Error: Cannot send to multiple streams
  at construct(app:///[native code]:0:0)
  at _construct(node_modules/@babel/runtime/helpers/construct.js:30:21)
  at Wrapper(node_modules/@babel/runtime/helpers/wrapNativeSuper.js:26:14)
  at ApiError(src/api/apiErrors.js:14:5)
  at makeErrorFromApi(src/api/apiErrors.js:36:20)
  at call(src/api/apiFetch.js:76:11)
  at tryCatch(node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:45:40)
[... then a bunch more framework frames, which don't vary either]
```
where only the message in the first line varies.

We have a lot of useful metadata about these errors, so we should be able to do pretty well at disambiguating. A few items that probably always mean a different error when they differ are:
* the route, i.e. the path portion of the URL (like `/api/v1/typing`) or better yet the part after `/api/v1/`
[...]

PR #4086, opened 2020-05-03
Tag API errors to distinguish them in Sentry

ray-kraesig:
Mark Sentry events for `ApiError`s with additional information. See issue #3864 for more details on the symptoms this alleviates.

For evidence that this actually *does* disaggregate messages with identical call stacks but distinct error messages, compare [this issue](https://sentry.io/organizations/zulip/issues/1643886301) and [this issue](https://sentry.io/organizations/zulip/issues/1643887841).


gnprice:
Thanks!

In those examples, it looks like the actual Sentry page doesn't show some of the important information about the error that it currently does show -- like the URL path, and the API error code.

I think that may be an artifact of how you made those test events, but I'd like to see what information we'll have on the errors in real life. Can you try a demo that's representative on that axis? It's probably enough to just make with `apiFetch` some request that resembles a good one but has some bogus arguments.

ray-kraesig:
Eh? Those examples seem to be at parity with the status quo, in that respect. As far as I can see (referencing [this recent real event](https://sentry.io/organizations/zulip/issues/1275118205/events/3f3f36e51df74a1cab96fca3d98f20e5/)), the data in question isn't part of the error event itself; it's part of the attached context, as custom `api` breadcrumbs.

In both of the examples, the error code is "200 OK", because I hijacked *successful* API calls to generate the bogus error events. The relevant API calls are visible [in the breadcrumb list](https://sentry.io/organizations/zulip/issues/1643887841/?project=191284#breadcrumbs) – though admittedly it's easy to miss there, thanks to all the spurious `console` breadcrumbs from the Redux debug output. ([We don't normally generate breadcrumbs for 200s](https://github.com/zulip/zulip-mobile/blob/bc4283a76433b362b69244da5df96f0291716c1a/src/api/apiFetch.js#L66), but it was done here as part of inducing the error reports.)

There is still the problem that we can't necessarily tell which `api` breadcrumb goes with which error: compare [this event](https://sentry.io/organizations/zulip/issues/1643886301/events/8929b52784de40d4ab3ef97c0e9bf525/?project=191284) and [this event](https://sentry.io/organizations/zulip/issues/1643886301/events/78a37db9b73a4ec8882b48a817fd9d09/?project=191284). This is a preexisting issue, though. (We probably want to use the existing [`ExtraErrorData` integration](https://docs.sentry.io/platforms/javascript/#extraerrordata) to deal with this, but I think our `flow-typed` libdefs will need to be cleaned up first.)

> It's probably enough to just make with `apiFetch` some request that resembles a good one but has some bogus arguments.

(Ironically, if I did this, the error information *wouldn't* be present – the breadcrumbs [are generated by `apiCall`](https://github.com/zulip/zulip-mobile/blob/bc4283a76433b362b69244da5df96f0291716c1a/src/api/apiFetch.js#L71-L74), one level above `apiFetch`.)

gnprice:
> all the spurious `console` breadcrumbs from the Redux debug output.

Yes -- those make those examples look quite different from a normal API error event, and much noisier. You hadn't said anything about what modifications you'd made to produce them, and I didn't want to guess. (The stack trace is also in an unusual spot.)

With that explanation, I'm happy to merge this.

> There is still the problem that we can't necessarily tell which `api` breadcrumb goes with which error: compare [this event](https://sentry.io/organizations/zulip/issues/1643886301/events/8929b52784de40d4ab3ef97c0e9bf525/?project=191284) and [this event](https://sentry.io/organizations/zulip/issues/1643886301/events/78a37db9b73a4ec8882b48a817fd9d09/?project=191284). This is a preexisting issue, though.

I don't think ambiguity has been a problem here in practice, but yeah, certainly it'd be nice to have the data attached to the event proper in a bit more structured of a way.


>  (We probably want to use the existing [`ExtraErrorData` integration](https://docs.sentry.io/platforms/javascript/#extraerrordata) to deal with this,

Docs say:

> This integration extracts all non-native attributes from the Error object and attaches them to the event as the `extra` data.

IIUC that's equivalent to adding a line like this to the `beforeSend` added in this branch:
```js
  event.extras = { httpStatus: exception.httpStatus, code: exception.code, data: exception.data };
```

(Plus doing something similar for other `Error` objects with extra properties on them... but we don't make any of those, except `ApnsMsgValidationError` which has an `extras`.)

That's helpful but I think doesn't cover the information we most want: the route and params. And I'm not sure we want the `data`.

Can we get the same effect by adding an `extras` property on the `ApiError` object? And then we can choose what goes there specifically with this in mind.

gnprice:
Hmm, no, actually, this isn't ready as is.

Here's the data this adds to the fingerprint:
```
+    event.fingerprint.push(exception.message, exception.code, exception.httpStatus.toString());
```

That has one item in common with the items I suggested in #3864, which are the route, the HTTP method, and the `code`. In particular it doesn't have the route, which is the most interesting of them. The main thing it has instead is the `message`. 

So I went back through some of the recent events on [the Sentry issue](https://sentry.io/organizations/zulip/issues/1275118205) linked from #3864, to see how this logic would do with them.

The result: it would split them much too finely. The Zulip server's messages don't try to come from a finite set of constants -- they don't try to, the messages aren't the machine-readable thing, that's what `code` and any extra properties are for. In particular:
* "Bad value for 'to': [42]" ([event](https://sentry.io/organizations/zulip/issues/1275118205/events/e08cf4beac5e4048a04b885005a1fedb/?project=191284))
* "Zulip server auth failure: (... a UUID ...) is not registered" ([event](https://sentry.io/organizations/zulip/issues/1275118205/events/72b3583e3a65468ead94677c0c301e22/?project=191284))

So adding the message to the fingerprint would make a significant regression in the usability of our Sentry dashboard.

Adding the code and HTTP status would be better than nothing -- I'd be happy to merge a revision that did that -- but not enough to resolve #3864.

How about you revise this to just add the code and HTTP status, so we can get that in, and then follow up with a PR to finish #3864 by providing the route? And the HTTP method, while we're there -- the Zulip API has several places where basically different API calls are at the same URL with different HTTP methods.

ray-kraesig:
> > This integration extracts all non-native attributes from the Error object and attaches them to the event as the `extra` data.
>
> IIUC that's equivalent to adding a line like this to the `beforeSend` added in this branch:
> 
> ```js
>   event.extras = { httpStatus: exception.httpStatus, code: exception.code, data: exception.data };
> ```

Something like that – although the [actual implementation](https://github.com/getsentry/sentry-javascript/blob/f4e59bcdd08f62983e2390e03c8ae69e392d9765/packages/integrations/src/extraerrordata.ts#L42) is a little more elaborate, and won't (*e.g.*) clobber any existing `extras` on `event`.

> That's helpful but I think doesn't cover the information we most want: the route and params. And I'm not sure we want the `data`.

Adding the route and params is trivial... but I take your point about `data`; a separate property is probably a better option.

> So adding the message to the fingerprint would make a significant regression in the usability of our Sentry dashboard.

So I was about to say it wouldn't be any worse than it was before the aggregation mechanism changed – `Bad value for 'to'` spam was a problem even then – but this turns out not to be true: under the new Discover, I haven't been able to find a way to do negative queries!

Anyway, even if it were true, dropping the message from the fingerprint would probably still be an improvement. I'll push a fix shortly.

ray-kraesig:
> How about you revise this to just add the code and HTTP status, so we can get that in, and then follow up with a PR to finish #3864 by providing the route?

If you take too long to merge this as-is, you may find this work attached to this PR. :wink: (Testing it now.)

ray-kraesig:
Nope! The simple version will still fracture events into too many issues – 'get' methods already have the params encoded in the route at that level.

(... Also, Sentry decided to filter out [all my data as PII](https://sentry.io/organizations/zulip/issues/1647378657/events/ea192817a1c342ec83bd17d7c6537bdb/) for a couple of hours. I'm still not sure what was up with that.)

gnprice:
> Nope! The simple version will still fracture events into too many issues – 'get' methods already have the params encoded in the route at that level.

Mmm, true!

One quick solution that would take us quite a way is to just take the first path component after `/api/v1/`.

There's more we could do, but it might need more extensive refactoring. For any more-complex design, let's discuss in a stream.

ray-kraesig:
Closing as superseded by #4096.

PR #4090, opened 2020-05-06
Remove `objectToParams`

ray-kraesig:
This function did both less and more than it really needed to.

Pulled out of plumbing work for #4086.


gnprice:
Thanks! Merged.

PR #4094, opened 2020-05-07
people auto complete: Send user ID for users having same full name.

agrawal-d:
Previously, our @-mention autocomplete had a bug, wherein it
failed to correctly mention the correct user if more than one user
had the same full name.

Now, if more than one users have the same full name, we now send
the user ID as well to ensure that the correct user is
@-mentioned.

### DIFF
--- a/src/autocomplete/PeopleAutocomplete.js
+++ b/src/autocomplete/PeopleAutocomplete.js
@@ -32,6 +32,14 @@ class PeopleAutocomplete extends PureComponent<Props> {
     const { users, onAutocomplete } = this.props;
     const user = users.find(x => x.email === email);
     if (user) {
+      // If another user with the same full name is found, we send the
+      // user ID as well, to ensure the mentioned user is uniquely identified.
+      if (users.find(x => x.full_name === user.full_name && x.user_id !== user.user_id)) {
+        // See the `get_mention_syntax` function in
+        // `static/js/people.js` in the webapp.
+        onAutocomplete(`**${user.full_name}|${user.user_id}**`);
+        return;
+      }
       onAutocomplete(`**${user.full_name}**`);
     }
   };


agrawal-d:
@chrisbobbe, @gnprice, @ray-kraesig this PR is ready for a review.

src/autocomplete/PeopleAutocomplete.js
```diff
@@ -32,6 +32,12 @@ class PeopleAutocomplete extends PureComponent<Props> {
     const { users, onAutocomplete } = this.props;
     const user = users.find(x => x.email === email);
     if (user) {
+      // If another user with the same full name is found, we send the
+      // user ID as well, to ensure the mentioned user is uniquely identified.
+      if (users.find(x => x.full_name === user.full_name && x.user_id !== user.user_id)) {
```
chrisbobbe, line 37: I expect this will look different when https://github.com/zulip/zulip-mobile/issues/3339 is resolved, but it looks good to me as a fix before then.

agrawal-d:
@chrisbobbe I've pushed some changes.

This version uses binary search and is faster than the linear one.
On CZO, the linear scan takes ~18ms or more.
This version takes  1ms or lesser (Date.now() has a resolution of 1ms)

chrisbobbe:
Hmm, interesting! I actually didn't have binary search in mind when I commented at https://github.com/zulip/zulip-mobile/pull/4094#discussion_r421843264.

Instead, I was thinking that we wouldn't be working with arrays here anymore, and instead something like an `Immutable.Map` (now much more possible with https://github.com/zulip/zulip-mobile/pull/4047#issuecomment-624881408 merged).

At #3339, @gnprice [said](https://github.com/zulip/zulip-mobile/issues/3339#issue-409048098):
> Instead, we should make these lookups through efficient, indexed data structures. A JS object works great. A `Map` works equally great, and can be a bit cleaner in the code.

However! The natural thing to do for #3339, to get `users` using an "efficient, indexed data structure" is, I think to do an [`Immutable.Map`](https://immutable-js.github.io/immutable-js/docs/#/Map) where keys are user IDs — not full names, which is what we care about here. So unless we have a separate index for full names, we're kind of worse off than we are with an array; we'd be awkwardly (and still linearly) traversing all entries in the user ID-keyed `Map`.

One solution, as you've shown here, is to use binary search on an array. (One step that I think is missing currently is to ensure `users` is sorted by `full_name` before passing it to `binarySearch`.) Better than linear, true, but a lookup in an `Immutable.Map`, as I understand it, is near-constant.

@gnprice, as someone who used to do research on algorithms, will know what to do here. But the one thing that hasn't come up in my discussions with him is the need to maintain indexes on things besides a canonical, unique ID.

agrawal-d:
>Instead, I was thinking that we wouldn't be working with arrays here anymore, and instead something like an Immutable.Map (now much more possible with #4047 (comment) merged).

I did consider maps. But JavaScript does not support MultiMaps, so if we attempt to hash with `full_name` as keys, the last value will be overwritten. So we can't use a `Map` in this use case.

If we hash using `user_id` - well, in this case, we will have to iterate over the map, which is again O(n).

src/autocomplete/PeopleAutocomplete.js
```diff
@@ -32,6 +33,17 @@ class PeopleAutocomplete extends PureComponent<Props> {
     const { users, onAutocomplete } = this.props;
     const user = users.find(x => x.email === email);
     if (user) {
+      const idx = binarySearch(users, elem => {
+        if (elem.full_name === user.full_name && elem.user_id === user.user_id) {
+          return 1;
+        }
+        return elem.full_name.toLowerCase().localeCompare(user.full_name.toLowerCase());
+      });
```
ray-kraesig, line 41: This won't work: the given comparator doesn't describe a consistent ordering.

Consider `user = { full_name: "John Doe", user_id: 4, ... }`, and a `users` array matching the following:
```javascript
users =~ [
  { full_name: "A", user_id: 0 }, // -1
  { full_name: "B", user_id: 1 }, // -1
  { full_name: "C", user_id: 2 }, // -1
  { full_name: "D", user_id: 3 }, // -1
  { full_name: "John Doe", user_id: 4 }, // 1
  { full_name: "John Doe", user_id: 5 }, // 0
  { full_name: "X", user_id: 6 }, // 1
  { full_name: "Y", user_id: 7 }, // 1
  { full_name: "Z", user_id: 8 }, // 1
];
```
(Trailing comments show the expected return value of the comparator for each element.)

`binary_search` will check `users[4]` first, which is the user in question; but the comparator will return `1`, so it will limit further searches to the range `[0, 3]`, and therefore never find `{ full_name: "John Doe", user_id: 5 }`.

src/utils/__tests__/binarySearch-test.js
```diff
@@ -0,0 +1,59 @@
...
+describe('binarySearch', () => {
+  const toFind = 100;
+  const comparator = (element: number) => {
+    if (element < toFind) {
+      return -1;
+    } else if (element > toFind) {
+      return 1;
+    }
+    return 0;
+  };
```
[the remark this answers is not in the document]
agrawal-d: Changed.

src/autocomplete/PeopleAutocomplete.js
agrawal-d: Fixed now.

agrawal-d:
@rk-for-zulip thanks for the review. I've made some changes to fix the issues you pointed out. Can you take a look again?

gnprice:
Seeing this add an implementaion of binary search, my first reaction is that binary search is notoriously tricky to implement completely correctly. (Perhaps I'll go make a thread in #learning about that story -- it might be fun to read about.)

So I'm inclined to look for other solutions for that reason.

From an algorithmic perspective, one thing I notice here is that this is one of several places that we do a linear pass through this same array. If you look at `getAutocompleteSuggestion`, which is called in the `render` method of this component:
```js
  const allAutocompleteOptions = getUsersAndWildcards(users);
  const startWith = filterUserStartWith(allAutocompleteOptions, filter, ownEmail);
  const initials = filterUserByInitials(allAutocompleteOptions, filter, ownEmail);
  const contains = filterUserThatContains(allAutocompleteOptions, filter, ownEmail);
  const matchesEmail = filterUserMatchesEmail(users, filter, ownEmail);
```
Each of those five lines represents a linear pass through all this data. The first one, in fact, actually allocates a copy:
```js
export const getUsersAndWildcards = (users: User[]) => [
  { ...NULL_USER, full_name: 'all', email: '(Notify everyone)' },
  { ...NULL_USER, full_name: 'everyone', email: '(Notify everyone)' },
  ...users,
];
```
which tends to be especially expensive.

So all of those together are going to do much more work than we'd do by a linear search here. In algorithmic terms, that means the most we could gain by doing something fancier is a constant factor -- O(n) + O(log n) is still O(n), just like O(n) + O(n) is.

And actually the situation is sharper than that -- we're doing all that work every time the user types another letter, to search again with the new query. This code, on the other hand, only runs once at the end when the user chooses an item from the list.

So, let's just not do anything fancy here for now. :slightly_smiling_face:

I do find, empirically on czo, that the people autocomplete can be pretty slow to respond with results. That would be good to improve, and it might involve using some more-clever algorithms and/or data structures in `getAutocompleteSuggestion` and its helpers. Happy to talk more about that on #mobile-team, if you're interested.

agrawal-d:
@gnprice Thanks for the review, I've reverted it back to the simple version. Can you take a look again? 

CC @chrisbobbe

chrisbobbe:
Great, thanks! This looks good to me, though it just occurred to me that it would be good to refer to an authoritative source on this syntax, in the code comment you've added. I believe the `get_mention_syntax` function in `static/js/people.js` is what we want here.

agrawal-d:
@chrisbobbe I've updated the commit message.

chrisbobbe:
Merged, thanks, @agrawal-d! I also made a small tweak: adding the reference to the web app in a code comment, in addition to the commit message.

Issue #3782, opened 2020-01-09
Upgrade to RN v0.62


gnprice:
Released 2020-03-26. Release notes:
https://reactnative.dev/blog/2020/03/26/version-0.62
and shorter version: https://github.com/facebook/react-native/releases/tag/v0.62.0

(~~This is for a version that isn't released yet :slightly_smiling_face:, so there's no work to do on it right now. Filing to have an identifier to refer to this upgrade by.~~)

~~We also won't start real work on this upgrade until after #3781, the upgrade to RN v0.61, is complete.~~ Which it is!

Changes include:
* Flow v0.113. In particular this [means](https://github.com/facebook/flow/blob/master/Changelog.md#01130):
  * (Flow v0.106) Changes to object spreads
  * (Flow v0.111) More changes to object spreads; see comment below
  * (Flow v0.112) Support for method calls in optional chains, like `foo?.bar()`

gnprice:
One thing v0.62 is projected to get us is the latest version of the React Developer Tools. See #3801, our docs [here](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#react-devtools), and that version's [release notes](https://reactjs.org/blog/2019/08/15/new-react-devtools.html).

ray-kraesig:
RN 0.62 is expected to come with a version of Flow which will, at long last, make exact object types with indexer properties usable. (See notes in [the current jsonable.js](https://github.com/zulip/zulip-mobile/blob/27271d94626fa936a82aa20d5da60ca3b1202d99/src/utils/jsonable.js) for more information.)

gnprice:
Upstream's [list of breaking changes](https://reactnative.dev/blog/2020/03/26/version-0.62#breaking-changes) is short, and from a quick grep we don't appear to be affected by any of them.

Advertised improvements include:
* [Support for Flipper](https://reactnative.dev/blog/2020/03/26/version-0.62#flipper-by-default), a developer tool. This is the change upstream seems most excited to talk about. May be handy -- certainly we should play with it once we're on this version!

* A small API for getting the user's system-wide [preference for a light or dark color scheme](https://reactnative.dev/blog/2020/03/26/version-0.62#new-dark-mode-features). I'd been thinking we should do this, particularly as it's now an OS feature (both Android and iOS) to have the theme change automatically on a daily cycle. I guess since this upgrade is coming and will make it easy, it makes sense to wait for that.

[...]

Issue #3452, opened 2019-04-11
Convert all object types to exact, or explicitly-inexact


gnprice:
~~(This becomes possible only with Flow v0.84, which we'll get as part of #3399; so it's blocked on that. And completely finishing it is blocked on later work in upstream Flow.)~~

Generally almost all our object types should be [exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types). See discussion in 61d2e3426 and 703739338... and see the many examples in `git log --grep exact` of type errors that had been concealed by inexact object types and were discovered by making the object types exact.

The Flow team agrees, and they have been making changes to make it easier to work that way across a codebase:
https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf

We've already gone and made the bulk of our object types exact; see 61d2e3426, 703739338, and 8370002d4 . With Flow v0.84, it becomes possible to explicitly mark as inexact the few types we actually want to describe that way. This task is to
* do that;
* for the remaining types that should be exact but aren't, make them exact (and fix whatever type errors stopped us from doing that before);
* make sure we really have gotten every case of the previous two steps (with that flowlint promised in "Step 1" of the roadmap post linked above -- which doesn't seem to exist yet).

gnprice:
#3399 is now done.

gnprice:
Flow upstream recently posted an update on how to do this upgrade:
https://medium.com/flow-type/how-to-upgrade-to-exact-by-default-object-type-syntax-7aa44b4d08ab

Looking through the steps there:

> 1. **Upgrading dev tools.** Ensure all your tools support the new syntax.

We already have a few explicit inexact object types and haven't seen them break anything, so this is mostly set.

Checking the four package versions mentioned in the post, we're good on three of them, but we should

- [ ] upgrade babel-eslint to at least v11-beta.

> 2. **Codemodding.** Upgrade your codebase to use the new syntax.

The flowlint rule `implicit-inexact-object`, which would find places we have the old syntax, doesn't appear until Flow v0.101. We're on v0.92. So this will be easier to complete, and make it stay completed, once we're up to that Flow version.

Looks like RN v0.60 brings us Flow v0.98, and RN v0.61 brings Flow v0.105. So that Flow upgrade, in turn, will come with the RN v0.61 upgrade #3781.

OTOH that doesn't mean we need to wait to upgrade our code! There's a tool called `flow-upgrade` linked from that post, which will make the transformation automatically across our code. Might try that.

[...]

PR #4101, opened 2020-05-12
Warn on @-mentioning someone who won't see it because not subscribed

### LINKED ISSUE #3373: Warn on @-mentioning someone who won't see it because not subscribed
Referenced users do not receive notifications from unsubscribed streams. See github issue. Unfortunately there is no warning at mobile-client :-( This is a ugly pitfall if you think you have notice someone, but the person didn't get any information.
See https://chat.zulip.org/#narrow/stream/48-mobile/topic/Referenced.20users.20on.20unsubscribed.20streams/near/709487

agrawal-d:
Fixes #3373.

**Summary of Changes**
* We use the endpoint `/users/{user_id}/subscriptions/{stream_id}` to get the subscription status.
* On non-PM screens, we show a warning when someone unsubscribed is mentioned.
* This warning shows up with a nice animation.
* Pressing the subscribe button subscribes the user and dismisses the warning.
* Tapping on the warning dismisses it.
* Multiple warnings can be shown at a time.

Screenshot_20200512-133929

Web app equivalent, for comparison:
image

### DIFF
--- a/src/api/subscriptions/getSubscriptionToStream.js
+++ b/src/api/subscriptions/getSubscriptionToStream.js
@@ -1,0 +1,21 @@
+/* @flow strict-local */
+import type { Auth, ApiResponseSuccess } from '../transportTypes';
+import { apiGet } from '../apiFetch';
+
+type ApiResponseSubscriptionStatus = {|
+  ...ApiResponseSuccess,
+  is_subscribed: boolean,
+|};
+
+/**
+ * Get whether a user is subscribed to a particular stream.
+ *
+ * See https://zulip.com/api/get-subscription-status for
+ * documentation of this endpoint.
+ */
+export default (
+  auth: Auth,
+  userId: number,
+  streamId: number,
+): Promise<ApiResponseSubscriptionStatus> =>
+  apiGet(auth, `users/${userId}/subscriptions/${streamId}`);
--- a/src/autocomplete/AutocompleteView.js
+++ b/src/autocomplete/AutocompleteView.js
@@ -19,14 +19,24 @@ type Props = $ReadOnly<{|
   isFocused: boolean,
   text: string,
   selection: InputSelection,
-  onAutocomplete: (input: string) => void,
+
+  /**
+   * The callback that is called when the user taps on any of the suggested items.
+   *
+   * @param input The text entered by the user, modified to include the autocompletion.
+   * @param completion The suggestion selected by the user. Includes markdown formatting,
+      but not the prefix. Eg. **FullName**, **StreamName**.
+   * @param lastWordPrefix The type of the autocompletion - valid values are keys of 'prefixToComponent'.
+   */
+  onAutocomplete: (input: string, completion: string, lastWordPrefix: string) => void,
 |}>;
 
 export default class AutocompleteView extends PureComponent<Props> {
   handleAutocomplete = (autocomplete: string) => {
     const { text, onAutocomplete, selection } = this.props;
+    const { lastWordPrefix } = getAutocompleteFilter(text, selection);
     const newText = getAutocompletedText(text, autocomplete, selection);
-    onAutocomplete(newText);
+    onAutocomplete(newText, autocomplete, lastWordPrefix);
   };
 
   render() {
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -196,8 +196,12 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   // See JSDoc on 'onAutocomplete' in 'AutocompleteView.js'.
-  handleMessageAutocomplete = (message: string, completion: string, lastWordPrefix: string) => {
-    this.setMessageInputValue(message);
+  handleMessageAutocomplete = (
+    completedText: string,
+    completion: string,
+    lastWordPrefix: string,
+  ) => {
+    this.setMessageInputValue(completedText);
 
     if (lastWordPrefix === '@') {
       if (this.mentionWarnings.current) {
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -16,6 +16,8 @@ import type {
   Dimensions,
   CaughtUp,
   GetText,
+  Subscription,
+  Stream,
 } from '../types';
 import { connect } from '../react-redux';
 import { withGetText } from '../boot/TranslationProvider';
@@ -35,6 +37,7 @@ import ComposeMenu from './ComposeMenu';
 import getComposeInputPlaceholder from './getComposeInputPlaceholder';
 import NotSubscribed from '../message/NotSubscribed';
 import AnnouncementOnly from '../message/AnnouncementOnly';
+import MentionWarnings from './MentionWarnings';
 
 import {
   getAuth,
@@ -43,6 +46,7 @@ import {
   getLastMessageTopic,
   getActiveUsersByEmail,
   getCaughtUpForNarrow,
+  getStreamInNarrow,
 } from '../selectors';
 import {
   getIsActiveStreamSubscribed,
@@ -64,6 +68,7 @@ type SelectorProps = {|
   draft: string,
   lastMessageTopic: string,
   caughtUp: CaughtUp,
+  stream: Subscription | {| ...Stream, in_home_view: boolean |},
 |};
 
 type Props = $ReadOnly<{|
@@ -116,7 +121,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
   messageInput: ?TextInput = null;
   topicInput: ?TextInput = null;
-
+  mentionWarnings: React$ElementRef<MentionWarnings> = React.createRef();
   inputBlurTimeoutId: ?TimeoutID = null;
 
   state = {
@@ -190,8 +195,15 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(draftUpdate(narrow, message));
   };
 
-  handleMessageAutocomplete = (message: string) => {
+  // See JSDoc on 'onAutocomplete' in 'AutocompleteView.js'.
+  handleMessageAutocomplete = (message: string, completion: string, lastWordPrefix: string) => {
     this.setMessageInputValue(message);
+
+    if (lastWordPrefix === '@') {
+      if (this.mentionWarnings.current) {
+        this.mentionWarnings.current.getWrappedInstance().handleMentionSubscribedCheck(completion);
+      }
+    }
   };
 
   handleMessageSelectionChange = (event: { +nativeEvent: { +selection: InputSelection } }) => {
@@ -261,6 +273,11 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(addToOutbox(this.getDestinationNarrow(), message));
 
     this.setMessageInputValue('');
+
+    if (this.mentionWarnings.current) {
+      this.mentionWarnings.current.getWrappedInstance().clearMentionWarnings();
+    }
+
     dispatch(sendTypingStop(narrow));
   };
 
@@ -354,6 +371,7 @@ class ComposeBox extends PureComponent<Props, State> {
       isAdmin,
       isAnnouncementOnly,
       isSubscribed,
+      stream,
     } = this.props;
 
     if (!isSubscribed) {
@@ -370,6 +388,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
     return (
       <View style={this.styles.wrapper}>
+        <MentionWarnings narrow={narrow} stream={stream} ref={this.mentionWarnings} />
         <View style={[this.styles.autocompleteWrapper, { marginBottom: height }]}>
           <TopicAutocomplete
             isFocused={isTopicFocused}
@@ -450,4 +469,5 @@ export default connect<SelectorProps, _, _>((state, props) => ({
   draft: getDraftForNarrow(state, props.narrow),
   lastMessageTopic: getLastMessageTopic(state, props.narrow),
   caughtUp: getCaughtUpForNarrow(state, props.narrow),
+  stream: getStreamInNarrow(state, props.narrow),
 }))(withGetText(ComposeBox));
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -1,0 +1,186 @@
+/* @flow strict-local */
+
+import React, { PureComponent } from 'react';
+import { connect } from 'react-redux';
+
+import type { Auth, Stream, Dispatch, Narrow, UserOrBot, Subscription, GetText } from '../types';
+import { TranslationContext } from '../boot/TranslationProvider';
+import { getActiveUsersById, getAuth } from '../selectors';
+import { isPrivateNarrow } from '../utils/narrow';
+import * as api from '../api';
+import { showToast } from '../utils/info';
+
+import AnimatedScaleComponent from '../animation/AnimatedScaleComponent';
+import MentionedUserNotSubscribed from '../message/MentionedUserNotSubscribed';
+
+type State = {|
+  unsubscribedMentions: Array<number>,
+|};
+
+type SelectorProps = {|
+  auth: Auth,
+  usersById: Map<number, UserOrBot>,
+|};
+
+type Props = $ReadOnly<{|
+  narrow: Narrow,
+  stream: Subscription | {| ...Stream, in_home_view: boolean |},
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
+
+class MentionWarnings extends PureComponent<Props, State> {
+  static contextType = TranslationContext;
+  context: GetText;
+
+  state = {
+    unsubscribedMentions: [],
+  };
+
+  /**
+   * Tries to parse a user object from an @-mention.
+   *
+   * @param completion The autocomplete option chosend by the user.
+      See JSDoc for AutoCompleteView for details.
+   */
+  getUserFromMention = (completion: string): UserOrBot | void => {
+    const { usersById } = this.props;
+
+    const unformattedMessage = completion.split('**')[1];
+    const [userFullName, userIdRaw] = unformattedMessage.split('|');
+
+    // We skip user groups, for which autocompletes are of the form
+    // `*<user_group_name>*`, and therefore, message.split('**')[1]
+    // is undefined.
+    if (unformattedMessage === undefined) {
+      return undefined;
+    }
+
+    if (userIdRaw !== undefined) {
+      const userId = Number.parseInt(userIdRaw, 10);
+      return usersById.get(userId);
+    }
+
+    for (const user of usersById.values()) {
+      if (user.full_name === userFullName) {
+        return user;
+      }
+    }
+
+    return undefined;
+  };
+
+  showSubscriptionStatusLoadError = (mentionedUser: UserOrBot) => {
+    const _ = this.context;
+
+    const alertTitle = _.intl.formatMessage(
+      {
+        id: "Couldn't load information about {fullName}",
+        defaultMessage: "Couldn't load information about {fullName}",
+      },
+      { fullName: mentionedUser.full_name },
+    );
+    showToast(alertTitle);
+  };
+
+  /**
+   * Check whether the message text entered by the user contains
+   * an @-mention to a user unsubscribed to the current stream, and if
+   * so, shows a warning.
+   *
+   * This function is expected to be called by `ComposeBox` using a ref
+   * to this component.
+   *
+   * @param completion The autocomplete option chosend by the user.
+      See JSDoc for AutoCompleteView for details.
+   */
+  handleMentionSubscribedCheck = async (completion: string) => {
+    const { narrow, auth, stream } = this.props;
+    const { unsubscribedMentions } = this.state;
+
+    if (isPrivateNarrow(narrow)) {
+      return;
+    }
+    const mentionedUser = this.getUserFromMention(completion);
+    if (mentionedUser === undefined || unsubscribedMentions.includes(mentionedUser.user_id)) {
+      return;
+    }
+
+    let isSubscribed: boolean;
+    try {
+      isSubscribed = (await api.getSubscriptionToStream(
+        auth,
+        mentionedUser.user_id,
+        stream.stream_id,
+      )).is_subscribed;
+    } catch (err) {
+      this.showSubscriptionStatusLoadError(mentionedUser);
+      return;
+    }
+
+    if (!isSubscribed) {
+      this.setState(prevState => ({
+        unsubscribedMentions: [...prevState.unsubscribedMentions, mentionedUser.user_id],
+      }));
+    }
+  };
+
+  handleMentionWarningDismiss = (user: UserOrBot) => {
+    this.setState(prevState => ({
+      unsubscribedMentions: prevState.unsubscribedMentions.filter(
+        (x: number) => x !== user.user_id,
+      ),
+    }));
+  };
+
+  clearMentionWarnings = () => {
+    this.setState({
+      unsubscribedMentions: [],
+    });
+  };
+
+  render() {
+    const { unsubscribedMentions } = this.state;
+    const { stream, narrow, usersById } = this.props;
+
+    if (isPrivateNarrow(narrow)) {
+      return null;
+    }
+
+    const mentionWarnings = [];
+    for (const userId of unsubscribedMentions) {
+      const user = usersById.get(userId);
+
+      if (user === undefined) {
+        continue;
+      }
+
+      mentionWarnings.push(
+        <MentionedUserNotSubscribed
+          stream={stream}
+          user={user}
+          onDismiss={this.handleMentionWarningDismiss}
+          key={user.user_id}
+        />,
+      );
+    }
+
+    return (
+      <AnimatedScaleComponent visible={mentionWarnings.length !== 0}>
+        {mentionWarnings}
+      </AnimatedScaleComponent>
+    );
+  }
+}
+
+// $FlowFixMe. TODO: Use a type checked connect call.
+export default connect(
+  state => ({
+    auth: getAuth(state),
+    usersById: getActiveUsersById(state),
+  }),
+  null,
+  null,
+  { withRef: true },
+)(MentionWarnings);
--- a/src/message/MentionedUserNotSubscribed.js
+++ b/src/message/MentionedUserNotSubscribed.js
@@ -35,11 +35,14 @@ const styles = StyleSheet.create({
   },
   text: {
     flex: 1,
-    color: 'white',
+    color: 'black',
   },
   button: {
     backgroundColor: 'orange',
     padding: 6,
+  },
+  buttonText: {
+    color: 'black',
   },
 });
 
@@ -68,7 +71,12 @@ class MentionedUserNotSubscribed extends PureComponent<Props> {
             }}
             style={styles.text}
           />
-          <ZulipButton style={styles.button} text="Subscribe" onPress={this.subscribeToStream} />
+          <ZulipButton
+            style={styles.button}
+            textStyle={styles.buttonText}
+            text="Subscribe"
+            onPress={this.subscribeToStream}
+          />
         </TouchableOpacity>
       </View>
     );
--- a/src/message/MentionedUserNotSubscribed.js
+++ b/src/message/MentionedUserNotSubscribed.js
@@ -23,7 +23,7 @@ type Props = $ReadOnly<{|
 |}>;
 
 const styles = StyleSheet.create({
-  mentionedUserNotSubscribed: {
+  outer: {
     flexDirection: 'row',
     alignItems: 'center',
     justifyContent: 'space-around',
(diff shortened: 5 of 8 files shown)


agrawal-d:
@chrisbobbe, @gnprice, @ray-kraesig this PR is ready for a review.

agrawal-d:
Blocked until https://github.com/zulip/zulip/pull/14966 is resolved.

agrawal-d:
zulip/zulip@14966 is now fixed, and I can continue work on this.

agrawal-d:
WIP, as I have to make some improvements, but can still be reviewed.

src/styles/composeBoxStyles.js
```diff
@@ -16,4 +16,22 @@ export default {
...
+    borderTopColor: 'rgb(214, 161, 54)',
+  },
+  mentionedUserNotSubscribedText: {
+    flex: 1,
+    color: 'white',
+  },
+  mentionedUserNotSubscribedButton: {
+    backgroundColor: 'orange',
+    padding: 6,
+  },
```
gnprice, line 36: These should go in the same file as they're used -- much easier to read and understand that way.

src/message/MentionedUserNotSubscribed.js
```diff
@@ -0,0 +1,71 @@
...
+  auth: Auth,
+  stream: { ...Stream },
+  user: UserOrBot,
+|};
+
+type Props = $ReadOnly<{|
+  narrow: Narrow,
+  userId: number,
+  onDismiss: (user: UserOrBot) => void,
+  dispatch: Dispatch,
```
gnprice, line 24: nit: blank line to separate the props the caller cares about from the ones supplied by wrappers

src/message/MentionedUserNotSubscribed.js
```diff
@@ -0,0 +1,71 @@
...
+import { withGetText } from '../boot/TranslationProvider';
+import { connect } from '../react-redux';
+import * as api from '../api';
+import { ZulipButton, RawLabel } from '../common';
+import { getAuth, getStreamInNarrow, getUserForId } from '../selectors';
+import styles from '../styles';
+
+type SelectorProps = {|
+  auth: Auth,
+  stream: { ...Stream },
```
gnprice, line 16: Prefer exact object types wherever possible; see discussion at #3452.

src/message/MentionedUserNotSubscribed.js
```diff
@@ -0,0 +1,71 @@
...
+        id: '{username} will not be notified unless you subscribe them to this stream.',
+        defaultMessage: '{username} will not be notified unless you subscribe them to this stream.',
+      },
+      { username: user.full_name },
+    );
+
+    return (
+      <View>
+        <TouchableOpacity onPress={this.handleDismiss} style={styles.mentionedUserNotSubscribed}>
+          <RawLabel style={styles.mentionedUserNotSubscribedText} text={alertTitle} />
```
gnprice, line 55: Hmm, I see. Let's use `Label` instead -- that's how we normally handle translation, so that will be more explicit about how we're translating this string.

Note that its argument is a `LocalizableText` -- you can pass an object there like `{ text: '{username} will not …', values: { username: user.full_name } }`.

(That probably is not an awesome way to write that interface, and we should make it take a `values` prop instead. But that's an independent improvement -- the API should work just fine already.)

I think that will also be a bit simpler in this code than calling `_.intl.formatMessage` directly is.

src/autocomplete/AutocompleteView.js
```diff
@@ -20,12 +20,15 @@ type Props = $ReadOnly<{|
   text: string,
   selection: InputSelection,
   onAutocomplete: (input: string) => void,
+  processAutoComplete: (completion: string, completionType: string) => void,
```
gnprice, line 23: Having two different callbacks here for the same event doesn't feel like the right interface. Better to adjust the one callback's interface to do what we need it to.

src/compose/ComposeBox.js
```diff
@@ -184,6 +195,64 @@ class ComposeBox extends PureComponent<Props, State> {
...
+    this.setState(prevState => ({
+      unsubscribedMentions: prevState.unsubscribedMentions.filter(
+        (x: number) => x !== user.user_id,
+      ),
+    }));
+  };
+
+  clearMentionWarnings = () => {
+    this.setState({ unsubscribedMentions: [] });
+  };
```
gnprice, line 248: The `ComposeBox` component already handles a ton of different things, and that's already made it a lot more challenging to understand than it ought to be. So I'd like to avoid adding more, and instead put new functionality in appropriate new components and  have `ComposeBox` just do the minimum to wire them in.

Here, I think a good way to do that would be to have a component that
* has zero or more individual `MentionedUserNotSubscribed` components as children
* has the `unsubscribedMentions` state
* has the implementations of these three methods

Then from `ComposeBox` we just need to be able to invoke `handleMentionSubscribedCheck` and `clearMentionWarnings`. I think the React way to do this is to have `ComposeBox` hold a "ref":
https://reactjs.org/docs/refs-and-the-dom.html
using `createRef`, that points to that child component.

(Alternatively the state could live up at `ComposeBox`, and it could just pass the child some callbacks to get at it. But that would expose somewhat more of the details to `ComposeBox`, and because of the situation where `ComposeBox` has too many details of too many things already, I'd rather avoid that.)

src/api/subscriptions/getSubscriptionToStream.js
```diff
@@ -0,0 +1,15 @@
...
+type ApiResponseSubscriptionStatus = {|
+  ...ApiResponseSuccess,
+  is_subscribed: boolean,
+|};
+
+export default (
+  auth: Auth,
+  userId: number,
+  streamId: number,
+): Promise<ApiResponseSubscriptionStatus> =>
```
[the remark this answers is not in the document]
agrawal-d: Added a JSDoc, and also a link to the relevant commit.

src/message/MentionedUserNotSubscribed.js
agrawal-d: Changed to `Subscription | {| ...Stream, in_home_view: boolean |}`, directly copied from the selector.

src/message/MentionedUserNotSubscribed.js
agrawal-d: Fixed.

src/styles/composeBoxStyles.js
agrawal-d: Moved to the component

src/message/MentionedUserNotSubscribed.js
agrawal-d: Changed to `Label`.

src/autocomplete/AutocompleteView.js
agrawal-d: Modified the current callback

src/compose/ComposeBox.js
agrawal-d: Done. The new component is `MentionWarnings`.

agrawal-d:
@gnprice thanks for the review. I've pushed some changes.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,131 @@
+/* @flow strict-local */
+
+import React, { PureComponent } from 'react';
+// import { connect } from '../react-redux';
```
agrawal-d, line 4: See TODO comment above the `connect` call.

agrawal-d:
>One other thing: Do you think a cancel button would fit in there somewhere, like the "x" button in the web app? It would just call clearMentionWarnings, I think. People might get frustrated that the warnings are hiding the message list from view, with no way to dismiss them except to send a message or subscribe the user.

It is possible to dismiss the warning by tapping on it.

>Even better would be to automatically dismiss a warning if the user removes the mention of that user from the input text, if that's easy.

I agree, but that would be difficult to implement, and even the webapp does not do this.

chrisbobbe:
> It is possible to dismiss the warning by tapping on it. An 'x' button would be even better, but there is no space :'(

Ah OK, got it. Yep, I just tested this on my physical iOS device and it works, though it takes several seconds. I expect this is because the app is slower in development than production, even on a physical device. But I looked at the code and see how this is happening, so, sounds good.

src/api/subscriptions/getSubscriptionToStream.js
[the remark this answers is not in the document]
agrawal-d: **Update**: The endpoint is now documented at https://zulipchat.com/api/get-subscription-status

src/compose/ComposeBox.js
```diff
@@ -106,6 +111,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
   messageInput: ?TextInput = null;
   topicInput: ?TextInput = null;
+  mentionWarnings: ?MentionWarnings = null;
```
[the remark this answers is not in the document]
agrawal-d: Thanks for the diff. I did try this earlier during debugging, but it did not make it to the final version. 
Used it now.

src/compose/ComposeBox.js
```diff
@@ -184,8 +190,14 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(draftUpdate(narrow, message));
   };
 
-  handleMessageAutocomplete = (message: string) => {
+  handleMessageAutocomplete = (message: string, completion: string, completionType: string) => {
```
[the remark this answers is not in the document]
agrawal-d: >If that's the case, let's see if we can say that explicitly with the types, so instead of string, it would be ':' | '#' | '@'. A JSDoc would certainly help with this as well, as I've mentioned in another comment. slightly_smiling_face

I had tried `':' | '#' | '@'` earlier as the type, but it causes the Flow error, because the value is derived from some plumbing in `getAutocompleteFilter`, and Flow is not intelligent enough to figure out that only these three are possible values, instead of a complete string.

agrawal-d:
Thanks for the review, @chrisbobbe! I've pushed some changes

src/compose/ComposeBox.js
```diff
@@ -250,6 +263,11 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(addToOutbox(this.getDestinationNarrow(), message));
 
     this.setMessageInputValue('');
+
+    if (this.mentionWarnings) {
+      this.mentionWarnings.getWrappedInstance().clearMentionWarnings();
```
gnprice, line 272: Hmm -- at the other spot this is `this.mentionWarnings.current.getWrappedInstance()`.

Is there a reason for the difference? I'm guessing it's unintentional.

Do they both work?

The examples in the docs at https://reactjs.org/docs/refs-and-the-dom.html seem to use `current`. Probably we should too, then, even if the other way also seems to work.

src/compose/ComposeBox.js
```diff
@@ -184,8 +190,15 @@ class ComposeBox extends PureComponent<Props, State> {
...
   };
 
-  handleMessageAutocomplete = (message: string) => {
+  // See JSDoc on 'onAutocomplete' in 'AutocompleteView.js'.
+  handleMessageAutocomplete = (message: string, completion: string, lastWordPrefix: string) => {
     this.setMessageInputValue(message);
+
+    if (lastWordPrefix === '@') {
+      if (this.mentionWarnings.current) {
+        this.mentionWarnings.current.getWrappedInstance().handleMentionSubscribedCheck(completion);
```
gnprice, line 203: This pattern can be shortened with `?.`:
`this.mentionWarnings.current?.getWrappedInstance()?.handleMentionSubscribedCheck(completion)`.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+  }
+}
+
+// TODO: Use a type checked connect call.  $FlowFixMe.
+export default connect(
+  (state, props) => ({
+    auth: getAuth(state),
+    usersById: getActiveUsersById(state),
+  }),
```
gnprice, line 136: Better to leave out the `props` parameter here when it's not going to be used. React Redux inspects this callback to see if it takes a second parameter, and if it doesn't then it applies an optimization where it knows it doesn't need to re-evaluate anything if the props change (and instead only if the state changes.)

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+    // is undefined.
+    if (unformattedMessage === undefined) {
+      return;
+    }
+    const [userFullName, userId] = unformattedMessage.split('|');
+    const unsubscribedMentions = this.state.unsubscribedMentions.slice();
+    let mentionedUser: UserOrBot;
+
+    // eslint-disable-next-line no-unused-vars
+    for (const [id, user] of usersById) {
```
gnprice, line 65: Can more simply say `for (const user of usersById.values())`.

See MDN:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+      if (userId !== undefined) {
+        if (user.user_id === userId) {
+          mentionedUser = user;
+          break;
+        }
```
gnprice, line 70: Also all this could be a *lot* simpler: `usersById.get(userId)` :smile: That's the great thing about a map data structure!

We'll still need the loop in the case where we're looking for a full name.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @param message The message text entered by the user in the compose box.
+   */
+  handleMentionSubscribedCheck = async (message: string) => {
+    const { usersById, narrow, auth, stream } = this.props;
+
+    if (isPrivateNarrow(narrow)) {
+      return;
+    }
+    const unformattedMessage = message.split('**')[1];
```
gnprice, line 52: This logic looks quite suspicious to me.

It seems like we're trying to parse the message's markup to look for an @-mention. That's a tricky thing to do accurately.

If we were to write logic to do that, it should be in a function of its own with a nice clean interface, and with thorough unit tests. For example: what about a message that has some `**bold**` text and then has an @-mention? Or, what about a message that has several @-mentions in it?

Looking at the place this is called, though, it looks like the argument it's passed may not actually be the message text? That's contrary to this function's jsdoc if so. But `handleMessageAutocomplete` passes it the `completion` it received, which in turn is described at the `onAutocomplete` prop's jsdoc as:
```
   * @param completion The suggestion selected by the user.
```
Which, I'm still not 100% sure what that means -- does it look like `@**Divyanshu Agrawal**` or `Divyanshu Agrawal` or something else? (Based on the code here in this function, I'm guessing the former.) Either way, that makes this problem a lot more tractable than it would be if it were really an entire message.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+      if (userId !== undefined) {
+        if (user.user_id === userId) {
+          mentionedUser = user;
+          break;
+        }
+      } else if (user.full_name === userFullName) {
+        mentionedUser = user;
+        break;
+      }
+    }
```
gnprice, line 75: One thing that would I think help significantly with reading this code and understanding what's going on is to take all the logic that parses the text of an @-mention and pull it out as its own function.

The function signature could be: `(mentionText: string, usersById: Map<number, UserOrBot>) => UserOrBot | void`. Where `mentionText` would be the argument passed to this function, which is called `message` but as I discovered above seems like it's (fortunately) not really a whole message's text at all.

That way it doesn't get entangled with this component's state, or with API calls or anything else.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+    if (!mentionedUser || unsubscribedMentions.includes(mentionedUser)) {
+      return;
```
gnprice, line 77: This has a bug because it's going to compare the user objects for identity. If `unsubscribedMentions` contains an object describing user 1234, and `mentionedUser` is user 1234, but they're different objects, then it will not see the match. One way this could happen is if the user object gets updated in Redux as a result of that user changing something (like their avatar URL) after they've already been added to `unsubscribedMentions` -- then `unsubscribedMentions` will still have the old object.

Instead, the comparison should just look at user IDs.

Probably the simplest solution is that `unsubscribedMentions` should just *be* an array of user IDs. When we need full user objects, we can get them by looking them up in `usersById` (with `usersById.get`, as mentioned in another comment.)

gnprice:
> > It is possible to dismiss the warning by tapping on it. An 'x' button would be even better, but there is no space :'(
> 
> Ah OK, got it. Yep, I just tested this on my physical iOS device and it works, though it takes several seconds. I expect this is because the app is slower in development than production, even on a physical device.

Hmm, that sounds real slow!

In general I'm not sure if the app is slower in a debug build than a release build; it certainly could be in principle, but I'm not sure I notice it. But it is not normally *that* slow; if something is taking several seconds in a debug build, it's going to be painfully slow in a release build too and we should fix that. @chrisbobbe Is it still this slow for you in the current revision?

agrawal-d:
(@chrisbobbe for me, that action is instantaneous)

src/message/MentionedUserNotSubscribed.js
```diff
@@ -0,0 +1,86 @@
...
+    flex: 1,
+    color: 'white',
+  },
+  mentionedUserNotSubscribedButton: {
+    backgroundColor: 'orange',
+    padding: 6,
+  },
+});
+
+class MentionedUserNotSubscribed extends Component<Props> {
```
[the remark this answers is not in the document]
agrawal-d: No - I just read about `PureComponent` in performance.md; will use it now.

src/compose/MentionWarnings.js
agrawal-d: Oh yeah - the jsdoc is incorrect. Here, i'm only removing the `**` from the @-mention, not the complete message.

agrawal-d: I've added relevant JSDocs to make it more clear what the `completion` looks like. ( It's of the form `**Divyanshu Agrawal**` ).

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    if (
+      !(await api.getSubscriptionToStream(auth, mentionedUser.user_id, stream.stream_id))
```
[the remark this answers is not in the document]
agrawal-d: I am inclined towards silently ignoring the error in this case, because the error will only confuse the user - did the @-mention get sent, or did it not?

src/compose/ComposeBox.js
agrawal-d: I'm getting this flow error:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:196:7

Flow does not yet support method or property calls in optional chains.

     193│     this.setMessageInputValue(message);
     194│
     195│     if (lastWordPrefix === '@') {
     196│       this.mentionWarnings.current?.getWrappedInstance()?.handleMentionSubscribedCheck(completion);
     197│     }
     198│   };
     199│


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:196:7

Flow does not yet support method or property calls in optional chains.

     193│     this.setMessageInputValue(message);
     194│
     195│     if (lastWordPrefix === '@') {
     196│       this.mentionWarnings.current?.getWrappedInstance()?.handleMentionSubscribedCheck(completion);
     197│     }
     198│   };
     199│



Found 2 errors
```

agrawal-d:
Thanks for the review @chrisbobbe, @gnprice! I've pushed some changes.

src/compose/ComposeBox.js
```diff
@@ -261,6 +277,11 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(addToOutbox(this.getDestinationNarrow(), message));
 
     this.setMessageInputValue('');
+
+    if (this.mentionWarnings) {
+      this.mentionWarnings.current.getWrappedInstance().clearMentionWarnings();
```
gnprice, line 282: This condition still says just `this.mentionWarnings` while the other says `this.mentionWarnings.current`, and after the method call inside this one was edited (following https://github.com/zulip/zulip-mobile/pull/4101#discussion_r463396159 ) to use `.current`.

Again they should probably all match -- they should if there isn't a reason for them to vary.

But also, if the code works in this version, then that means the condition isn't actually needed at all, because the condition in this version isn't doing anything. Perhaps just take these conditions out? They don't appear in the examples in the docs:
https://reactjs.org/docs/refs-and-the-dom.html

src/compose/ComposeBox.js
gnprice: Dang, so be it then.

It looks like this is https://github.com/facebook/flow/issues/4303 ... which, good news, is closed as fixed! In v0.112.0, looks like. Which means we'll get that when we upgrade to RN v0.62: #3782.

src/compose/MentionWarnings.js
gnprice: > It's of the form `**Divyanshu Agrawal**`

Neither of the things I guessed! Glad the answer is written down :wink:

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,162 @@
...
+    // is undefined.
+    if (unformattedMessage === undefined) {
+      return undefined;
+    }
+
+    if (userIdRaw !== undefined) {
+      const userId = Number.parseInt(userIdRaw, 10);
+      const user = usersById.get(userId);
+      if (user !== undefined) {
+        return user;
```
gnprice, line 59: Let's leave out the `if (user !== undefined)` and just unconditionally return `user` at this point.

Consider what that condition means:
* There was an explicit user ID in the @-mention.
* For whatever reason, we don't find the user that the @-mention specifically identified by user ID.
* We don't return yet, fall through, and end up looking to see if there's some *other* user who happens to have the same name, and return that one.

If that fall-through has any effect at all, then it means returning a user that is definitely not the same user object that the person who wrote the @-mention had selected. Very likely they have nothing at all to do with the person who was actually @-mentioned, and are just another person with the same name. That's no good -- much better to be honest about the fact that we simply didn't find the intended user.

In the context of the warning this code is part of providing, I guess the exact right thing to do is whatever the server is going to do to identify who's been @-mentioned. But I certainly hope that what the server is going to do when there's an explicit user ID is to stick to that, and not ignore it in favor of some potentially completely unrelated user who happens to have the same name. If not, then that's a bug that I expect would get fixed. So let's assume the server is doing the sensible thing, and do the same.

src/compose/MentionWarnings.js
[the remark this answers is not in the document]
gnprice: > the error will only confuse the user - did the @-mention get sent, or did it not?

Well, that uncertainty is pretty appropriate! After all, the point of these warnings is that the @-mention won't reach the mentioned user if they're not subscribed.

We'd like people to have confidence in general about whether someone else is going to get notified about their message or not. In cases where really don't know, the way to maintain that confidence is to say we don't know. Otherwise, when we *do* know and don't show a warning, people won't know whether to believe it, or whether instead this is another of those times that we failed to show them a warning before the surprising thing happened.

I think a toast would be the appropriate form to report the issue. `showErrorAlert` pops up a modal dialog box, which is more interruptive than seems appropriate for this condition. Compare:
https://material.io/components/dialogs#usage
https://material.io/components/snackbars#usage
(where "snackbar" is the current Material term for something very similar to a toast.)

Here's another iteration on the text: "Couldn't load information about ${user.full_name}".

gnprice:
Thanks @agrawal-d for the revisions! Getting close. A few more comments for this round.

src/compose/MentionWarnings.js
agrawal-d: Makes sense!

src/compose/MentionWarnings.js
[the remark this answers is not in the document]
agrawal-d: Yeah, makes sense.

src/compose/ComposeBox.js
agrawal-d: The Flow typedef seems to prefer the use of `current`:
```js
  declare export function createRef<T>(
  ): {|current: null | T|};
```

agrawal-d: And the Docs say we should access a ref using `current`:
https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs

I'll use current in all places for now. If you still think we should leave it out, let me know. I'll remove it then.

agrawal-d:
Thanks for the review, @gnprice! I've pushed some changes.

src/compose/MentionWarnings.js
```diff
@@ -0,0 +1,186 @@
...
+    return undefined;
+  };
+
+  showSubscriptionStatusLoadError = (mentionedUser: UserOrBot) => {
+    const _ = this.context;
+
+    const alertTitle = _.intl.formatMessage(
+      {
+        id: "Couldn't load information about {fullName}",
+        defaultMessage: "Couldn't load information about {fullName}'",
```
gnprice, line 80:
```suggestion
        id: "Couldn't load information about {fullName}",
        defaultMessage: "Couldn't load information about {fullName}",
```

This also serves as a reminder that we should really make our own little wrapper for `formatMessage` that saves us from having to repeat the string twice :slightly_smiling_face:

gnprice:
I think I want to merge first and then follow up on these, but here are some comments on the UI design.

The "Subscribe" button has an odd shape:
image

Very tall vertical margins around the text, and quite narrow horizontal margins.

It looks like that ZulipButton has a height of 44px? I believe we'd said elsewhere that the ideal minimum height for a touch target is 48px. So I imagine what's going on is:
* the button is that tall to make it an acceptably tall touch target
* that causes the tall vertical margins seen around the text
* and there isn't any correspondingly large horizontal padding/margin.

I think the ideal visual shape for a button isn't quite so tall. Material specifies 36px height:
https://material.io/components/buttons#specs
But then the touch target can extend into the margin outside the button, beyond the space that's drawn to look like a button.

---

The border at the top of the banner feels pretty heavy:
image
and also feels unbalanced, in that there isn't a corresponding border at the bottom.

How about just not having a border? The orange "warning" color should contrast pretty sharply with the message list's background already.

---

The orange background color doesn't feel great to me, somehow. I appreciate that you have a comment saying something about where it came from:
```js
    backgroundColor: 'hsl(40, 100%, 60%)', // Material warning-color
```
I'm not finding that in a quick look at https://material.io/design/color/the-color-system.html and some related pages. Where in the Material docs did you find the reference to that color?

Also not finding that specific color (aka `#ffbb33`) anywhere in the Material palette as represented here:
https://material.io/resources/color/

---

The orange color is pretty light. Light backgrounds are perfectly fine but it's important to use dark text with them. There's a nice discussion here:
https://material.io/design/color/text-legibility.html

The Material color tool [says](https://material.io/resources/color/#!/?view.left=1&view.right=1&primary.color=ffbb33) when I input that color:
> White Text: NOT LEGIBLE :warning:
> Black Text: min 59% opacity

Very similar [result](https://material.io/resources/color/#!/?view.left=1&view.right=1&primary.color=ffa600) with the color that's used for the button:
> White Text: NOT LEGIBLE :warning:
> Black Text: min 60% opacity

---

The animation when the banner expands in is kind of odd. It's the exact same one we have for some other banners, so I don't want to emphasize this too much. But it's pretty equally odd for all of them -- it really doesn't make any sense for a banner.

Specifically the animation sequence is:
* Start state: no banner.
* Suddenly all at once the rest of the layout moves around, to make room for where the banner is going to appear.
* In the center of that space where the banner will be, there's a tiny rectangle which is the banner scaled down.
* From that center, it expands out to fill the space.
* This scaling is happening *in all directions* -- when it's halfway done expanding, it's half the height it's going to be *and also half the width*. The rest of the space the banner will fill is just blank space -- we made room for it all at once in step 2.

It'd be good to fix that and do something better. In fact it'd quite likely be an improvement to just drop animation entirely for these banners. (This one; the "N unreads" banner, which is by far the most common; and not sure offhand if other banners like "Connecting..." or "No Internet connection" behave this way too.)

Better yet than no animation would be an appropriate animation. The logical thing for a banner is to slide in from -- in this case the bottom since it's extending up from something, and from the top for the UnreadNotice banner and others that appear from the top. I took a quick look at Material [on banners](https://material.io/components/banners#behavior), and it says something similar:
> Banners typically appear when a screen loads content.
>
> Banners that appear after a screen loads should animate on screen from the top of a layout. If the banner is at the same elevation as content, it pushes content downwards.

(Material banners always appear at the top.)

gnprice:
OK, and merged!

Of the UI remarks I made in the previous comment, I went and fixed the color contrast issue, because that's the only one where the issue is really unacceptable -- it's the only one that can make that piece of the app unusable for some people. Also fixed the tiny point mentioned in the previous comment.

Fixing the text contrast required a bit of digging into the ZulipButton component, and that led me to do some other work there. I'll send that separately, as well as some changes addressing the other UI issues.

PR #4103, opened 2020-05-12
docs: Add more troubleshooting info for build and run.

agrawal-d:
Adds command to clean gradle build, and link to upstream
troubleshooting docs.

### DIFF
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -144,6 +144,11 @@ next person with a setup like yours.
 
 ## Troubleshooting
 
+Apart from the steps mentioned below, you may find the
+[React Native troubleshooting docs] to be helpful.
+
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Merged, thanks!

I just made one small tweak: changing `gradlew clean` to `./gradlew clean`. My understanding is that sometimes the shell needs a hint that you want it to run a script in a particular file, as opposed to a built-in command or something in the `PATH` variable — the `./` at the beginning makes it clear that it's a file path. If you didn't `cd` into `android` first, you could say `android/gradlew`, without the leading `./`, because it already looks enough like a file path that it's reasonable to assume it is one.

chrisbobbe:
And without the leading `./`, I get this:

```
$ gradlew clean
bash: gradlew: command not found
```

Issue #4035, opened 2020-04-11
Pin down the main forms of identifying a PM thread, and their conversion helpers


gnprice:
There are a bunch of different ways we identify a PM conversation -- all of them variations on listing the people in the conversation:
* Include self, or not.
  * Or include self only for the self-PM thread.
  * Or include self for group PMs (i.e. >= 3 people total), and for the self-PM thread (i.e. 1 person total), but not for non-self 1:1 threads (i.e. exactly 2 people total.)
    * Yes, this [really is how it works](https://chat.zulip.org/#narrow/stream/3-backend/topic/small.20huddles/near/823600) in one part of the API! Namely the `unread_msgs` initial data.
* Use user IDs, or emails.
* Sort the items, usually.
  * AFAIK they always come sorted, either numerically for user IDs (even when stringified, like `2,11`) or lexicographically for emails. But there may be exceptions in a corner I haven't looked closely into.
* Use arrays, or comma-joined strings.

Much of this complexity comes to us from the Zulip server API -- this is an unfortunately unnecessarily-complex area of the API.

Then some of it is I think self-inflicted within the mobile codebase; partly from cases of #3764, and partly in other ways. And even what comes originally from the API, we could do a lot to handle more cleanly:
* name things to clearly disambiguate
* write comments on the nuances
* where we can and it helps, convert things at the edge into a sane internal format, on the principle of the [crunchy shell](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md).

The spot in the code where this is most concentrated is `src/utils/recipient.js`... but that file is more the incomplete solution than it is the problem -- what's needed is partly to document, rename, and clean up some of those helpers, but also to deal with other places in the code where we're interacting with this tangled API surface, e.g. converting from one form to another, and doing so by hand without using that file's helpers and with no explicit reference to the different pieces of API involved.

---

This came up again just today at https://github.com/zulip/zulip-mobile/pull/3535#issuecomment-612268975 : it's basically the remaining obstacle to us completing that PR, fixing #3133, so that the PMs tab stops feeling broken for people who heavily use PMs.

It's been a recurring drag on development in certain areas for a long time, too, so it's really past time we took care of it.

I have a couple of draft branches with a lot of comments in them, from when I studied some of this a month or two ago. So, assigning to myself to at least kick things off by cleaning up and sending some of those changes.

gnprice:
I've just pushed some commits that make progress on this:
[...]

PR #4104, opened 2020-05-12
Use recent_private_conversations on PMs screen, when available

### LINKED ISSUE #3133: PMs screen doesn't show many conversations

rk-for-zulip:
Attempt to acquire `recent_private_conversations` from the server (zulip/zulip#11946). Add a new implementation for `getRecentConversations` which will use the data provided therein.

Since there are many Zulip servers still active which predate this server-side feature, we continue to use the old implementation when connecting to those servers, to avoid a significant regression.

This PR is intended to supersede PR #3535, with which it shares some code and history. Notable functional differences include:
 * *(the first two paragraphs, above)*
 * Incoming IDs for recent conversations are always sorted (per [this comment](https://github.com/zulip/zulip-mobile/pull/3535#issuecomment-626071303)).
 * New functions added have been to `exampleData` to ease test construction.
 * The new `EVENT_NEW_MESSAGE` reducer now properly filters `display_recipient`. (Noticed while converting its tests to use the new functions.)

Fixes #3133.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -509,5 +509,6 @@ export const eventNewMessageActionBase /* \: $Diff<EventNewMessageAction, {| mes
   type: EVENT_NEW_MESSAGE,
   id: 1001,
   caughtUp: {},
+  ownId: selfUser.user_id,
   ownEmail: selfAccount.email,
 };
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -188,16 +188,17 @@ export const stream: Stream = makeStream({
  * Messages
  */
 
-const displayRecipientFromUser = (user: User): PmRecipientUser => {
-  const { email, full_name, user_id: id } = user;
-  return deepFreeze({
-    email,
-    full_name,
-    id,
-    is_mirror_dummy: false,
-    short_name: '', // what is this, anyway?
+const displayRecipientFromUsers = (users: User[]): PmRecipientUser[] =>
+  users.map(user => {
+    const { email, full_name, user_id: id } = user;
+    return deepFreeze({
+      email,
+      full_name,
+      id,
+      is_mirror_dummy: false,
+      short_name: '', // what is this, anyway?
+    });
   });
-};
 
 /** Boring properties common to all example Message objects. */
 const messagePropertiesBase = deepFreeze({
@@ -239,14 +240,14 @@ const messagePropertiesFromSender = (user: User) => {
 };
 
 /** Beware! These values may not be representative. */
-export const pmMessage = (extra?: $Rest<Message, {}>): Message => {
+export const pmMessageFromTo = (from: User, to: User[], extra?: $Rest<Message, {}>): Message => {
   const baseMessage: Message = {
     ...messagePropertiesBase,
-    ...messagePropertiesFromSender(otherUser),
+    ...messagePropertiesFromSender(from),
 
     content: 'This is an example PM message.',
     content_type: 'text/markdown',
-    display_recipient: [displayRecipientFromUser(selfUser)],
+    display_recipient: displayRecipientFromUsers([from, ...to]),
     id: 1234567,
     recipient_id: 2342,
     stream_id: -1,
@@ -257,6 +258,9 @@ export const pmMessage = (extra?: $Rest<Message, {}>): Message => {
 
   return deepFreeze({ ...baseMessage, ...extra });
 };
+
+export const pmMessage = (extra?: $Rest<Message, {}>): Message =>
+  pmMessageFromTo(selfUser, [otherUser], extra);
 
 const messagePropertiesFromStream = (stream1: Stream) => {
   const { stream_id, name: display_recipient } = stream1;
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -3,7 +3,7 @@ import deepFreeze from 'deep-freeze';
 import { createStore } from 'redux';
 
 import type { CrossRealmBot, Message, PmRecipientUser, Stream, User } from '../../api/modelTypes';
-import type { Action, GlobalState, RealmState } from '../../reduxTypes';
+import type { Action, GlobalState, MessagesState, RealmState } from '../../reduxTypes';
 import type { Auth, Account, Outbox } from '../../types';
 import { ZulipVersion } from '../../utils/zulipVersion';
 import {
@@ -17,6 +17,7 @@ import rootReducer from '../../boot/reducers';
 import { authOfAccount } from '../../account/accountMisc';
 import { HOME_NARROW } from '../../utils/narrow';
 import { NULL_OBJECT } from '../../nullObjects';
+import { objectFromEntries } from '../../jsBackport';
 
 /* ========================================================================
  * Utilities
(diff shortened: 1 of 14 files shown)


ray-kraesig:
Suggested future work:

* In `EVENT_NEW_MESSAGE` handlers, switch over from using `ownEmail` to using `ownId` where possible.

* Remove the "nastiness" mentioned in the previous PR's 534387209e79d954444d077a5a8d974f1038593e – preferably not by pushing it into `recipient.js`, which has quite enough nastiness in it already, but instead by making the sole downstream consumer of this code accept a #3764-compliant data structure.

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -51,10 +67,114 @@ export const getRecentConversations: Selector<PmConversationData[]> = createSele
...
+ * recent message from each.
+ *
+ * Switches between implementations as appropriate for the current server
+ * version.
+ */
+export const getRecentConversations: Selector<PmConversationData[]> = createSelector(
+  state => state,
+  getMetaselector,
+  (state, metaselector) => metaselector(state),
+);
```
gnprice, line 176: This is an example of a selector that doesn't do any usefully cacheable work, so there's no need for `createSelector`.

(Also the `state => state` means the cache will never be hit unless `state` is identically unchanged, and in that case we generally shouldn't be getting this far anyway.)

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -51,10 +67,114 @@ export const getRecentConversations: Selector<PmConversationData[]> = createSele
...
+
+/**
+ * The server version in which 'recent_private_conversations' was first made
+ * available.
+ */
+const DIVIDING_LINE = new ZulipVersion('2.1-dev-384-g4c3c669b41');
+
+// Private. Selector to choose between other selectors. (This avoids needlessly
+// recomputing the old version when we're on a new server, or vice versa.)
+const getMetaselector: Selector<Selector<PmConversationData[]>> = createSelector(
```
gnprice, line 147: Having this "metaselector" makes the code rather more complex than if we simply put this logic in `getRecentConversations`.

The thing it accomplishes is basically that the result of `isAtLeast` gets memoized. I'm dubious that that's much of a win -- we've designed that to precompute most of the work when constructing the `ZulipVersion` instance.

If you're keen on memoizing that, though, a straightforward way would be to do so directly:
```js
const getServerIsOld: Selector<boolean> = createSelector(
  getServerVersion,
  version => !(version && version.isAtLeast(DIVIDING_LINE),
);
```
(Or flip it around for a different name.) Then the memoized data is plain old data, and the control flow has fewer indirections to follow.

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -1,46 +1,62 @@
...
-      ids: pmUnreadsKeyFromMessage(msg, ownUser.user_id),
-      emails: normalizeRecipientsSansMe(msg.display_recipient, ownUser.email),
-      msgId: msg.id,
-    }));
+/**
+ * Given a list of PmConversationPartial or PmConversationData, trim it down to
+ * contain only the most recent message from any conversation, and return them
+ * sorted by recency.
+ */
+const collateByRecipient = <T: { recipients: string, msgId: number }>(
```
gnprice, line 25: Looks like this is only used in the one place? I don't think it's a win to pull it out -- the interface is pretty complicated in isolation.

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -1,46 +1,62 @@
...
+ *
+ * Note that, since this will only ever be called by other selectors, the inner
+ * function doesn't need to do any memoization of its own.
+ */
+const getAttachUnread: Selector<UnreadAttacher> = createSelector(
+  getUnreadByPms,
+  getUnreadByHuddles,
+  (unreadPms: { [number]: number }, unreadHuddles: { [string]: number }) => (
+    partials: $ReadOnlyArray<PmConversationPartial>,
+  ) =>
```
gnprice, line 58: There's nothing to memoize here, because the body is just a lambda.

The logic inside that lambda -- certainly the expression after `unread:`, complete with comments -- is indeed helpful to factor out into a helper function. That helper can take `unreadPms` and `unreadHuddles` among its arguments.

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -51,10 +67,114 @@ export const getRecentConversations: Selector<PmConversationData[]> = createSele
...
+  ) => {
+    const getEmail = (id: number): string => {
+      const user = allUsers.get(id);
+      if (user) {
+        return user.email;
+      }
+      throw new Error('getRecentConversations: unknown user id');
+    };
+
+    const recipients = recentPCs.map(conversation => {
```
gnprice, line 118: As you astutely noted on the legacy implementation, there's probably a better name to give to this local :wink:

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -51,10 +67,114 @@ export const getRecentConversations: Selector<PmConversationData[]> = createSele
...
+      const userIds = conversation.user_ids.slice();
+      if (userIds.length !== 1) {
+        userIds.push(ownUser.user_id);
+        userIds.sort((a, b) => a - b);
+      }
```
gnprice, line 123: This should also go in `recipients.js`, with a name and jsdoc.

Specifically, a good name/interface might be:
* include the `.join(',')` part;
* call it `pmUnreadsKeyFromRecentPmsKey`.

src/pm-conversations/pmConversationsSelectors.js
```diff
@@ -51,10 +67,114 @@ export const getRecentConversations: Selector<PmConversationData[]> = createSele
...
+
+      return {
+        ids: userIds.join(','),
+        recipients: normalizeRecipientsSansMe(
+          userIds.map(id => ({ email: getEmail(id) })),
+          ownUser.email,
+        ),
```
gnprice, line 130: Oof. I think this ends up having the right behavior... but it sure is confusing from a #4035 perspective. Note that we're giving this function a different version of "the users for this PM conversation" than we do at its other call site. And that not all possible versions of that would work -- notably if we used `conversation.user_ids` directly, that would not work.

At a minimum I'd want a comment here explaining why this ends up with the right set of users (and what the right set of users is supposed to be.)

Better would be another small function in `recipients.js` with its own name and jsdoc.

A good followup would be to eliminate this `recipients` property entirely. It flows to just a handful of places, all under our direct control.

gnprice:
(Several of these commits now appear in #4116, as the 6 commits to 3c7791532 along the series to 27580526d, in revised versions responding to my comments above.)

src/__tests__/lib/exampleData.js
```diff
@@ -88,12 +95,12 @@ const userOrBotProperties = ({ name: _name }) => {
...
     is_admin: false,
     timezone: 'UTC',
-    user_id: randUserId(),
+    user_id: user_id ?? randUserId(),
   });
 };
 
 /** Beware! These values may not be representative. */
-export const makeUser = (args: { name?: string } = {}): User =>
+export const makeUser = (args: UserOrBotPropertiesArgs = NULL_OBJECT): User =>
```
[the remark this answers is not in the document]
gnprice: This was answered at https://github.com/zulip/zulip-mobile/pull/4116#discussion_r426357258 . Seems like a fine workaround in the circumstances.

gnprice:
I've just sent #4392 as a new approach to this issue.

PR #4108, opened 2020-05-13
message list: Fix copying links to clipboard.

### LINKED ISSUE #4107: Copy to clipboard copies a pratial URL for in-realm links
For example, a link to a narrow would not include the realm domain and only include the `/#narrow/stream/topic` part.

For a demonstration of this bug, try opening [this message](https://chat.zulip.org/#narrow/stream/7-test-here/topic/mobile.20copy.20to.20clipboard.20demo/near/876477) in the app and copy the link to clipboard

agrawal-d:
Previously, in-realm links copied to the clipboard didn't include
the realm domain, and were essentially invalid.

Now, the complete URL is copied to the clipboard.

Closes #4107.

### DIFF
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -6,7 +6,7 @@ import type { Dispatch, GetText, Message, Narrow, Outbox } from '../types';
 import type { BackgroundData } from './MessageList';
 import type { ShowActionSheetWithOptions } from '../message/messageActionSheet';
 import { showToast } from '../utils/info';
-import { isUrlAnImage } from '../utils/url';
+import { isUrlAnImage, getFullUrl } from '../utils/url';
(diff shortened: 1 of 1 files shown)


ray-kraesig:
It's possibly worth noting that:
  * a) this will fail for path-relative URLs (`<a href="./topic/Venice3">`) or protocol-relative URLs (`<a href="//www.google.com">`);
  * b) thankfully, these don't seem to be allowed by the Zulip server in user-created links; and
  * c) even if they were, 
     1. we can't do anything better for path-relative links,
     2. protocol-relative links are rare enough not to worry about, and
     3. by the time b) changes, if it ever does, we should have a real URL library anyway.

It's probably *not* worth trying to fit this note into a comment anywhere but here, though.

Merging!

PR #4109, opened 2020-05-13
ligthbox: Toggle header & footer on tapping the entire view.

### LINKED ISSUE #4098: lightbox: Tap outside image to show header/footer again
@eugenolman reports [in chat](https://chat.zulip.org/#narrow/stream/9-issues/topic/Lightbox.20context.20menu.20mobile/near/873685):
> Once you click on an image you have it full screen resized and outside the picture you have black screen. [...]
> Now if you have a slim picture you have alot of black. And to return to menu ere you have an X to close the image you need to tap on the image. Nothing happen if you tap on the black part. [...]
>
> Double tap on black increase zoom. Single tap nothing.
>
> Normally you should be able to tap on black and get the menu back then click X to return to the thread.

I agree. This would give us the same behavior as other apps I briefly checked:
* FB Messenger on iOS
* (Google) Photos on Android
* built-in Photos app on iOS

As is, once you hide the header and footer I don't see a way to get them back, short of exiting the lightbox and coming back to it.

agrawal-d:
As opposed to only when the visible image is tapped, as we
currently do.
This is done by simply adding another prop `onViewTap` ( for the
black part) in the react-native-photo-view component[1].
Note that `onTap` and `onViewTap` respond mutually exclusive
areas, the union of which makes up the entire component.

Closes #4098.

[1] https://github.com/alwx/react-native-photo-view#properties

### DIFF
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -103,6 +103,7 @@ class Lightbox extends PureComponent<Props, State> {
           style={[styles.img, { width }]}
           resizeMode="contain"
           onTap={this.handleImagePress}
+          onViewTap={this.handleImagePress}
         />
         <SlideAnimationView
           property="translateY"


ray-kraesig:
It'd be nice to test this on iOS, too, but I think there's little risk of a regression here. Merging!

PR #4110, opened 2020-05-13
spinner [nfc]: Fix misaligned images.

### LINKED ISSUE #4080: Loading spinner ring is wobbly
Peek 2020-05-01 13-40 trimmed twice

It's tricky to tell, but I don't *think* it's an optical illusion.

agrawal-d:
spinner

The spinner is still  *slightly* wobbly, but it's only visible if you zoom in a lot. If this is not acceptable, feel free to close this PR. :) 

The image was misaligned, and had more empty space on top than in
the bottom, causing it to appear wobbly when rotated. The
image has been edited to fix this.

Closes #4080.


chrisbobbe:
Nice! Could you update `spinning-progress-white.png` exactly the same way? I made that file in bfb884a0e by copying `spinning-progress-black.png` and just inverting the colors, so it'll have the same wobble. Not sure about `spinning-progress.png`, the one with the default green color.

agrawal-d:
Sure. Yay. it's really simple with GIMP - it's a built in command to invert colours. What did you origianlly use?

agrawal-d:
I just checked `spinning-progress.png`, it's okay.

chrisbobbe:
Merged, thank you! I just removed the `[nfc]` from the commit message since there's a visible difference after this change. True, there aren't any code changes, but I wouldn't necessarily describe it as a "pure refactor".

chrisbobbe:
> (Yay. it's really simple with GIMP - it's a built-in command to invert colours. What did you originally use?)

I think I used Preview on my mac? 🤔 I had to Google around a lot to find the tool that would invert the colors. I think GIMP would have been easier but I guess I didn't feel like getting it all set up and stuff. 😛

PR #4113, opened 2020-05-15
ios deps [nfc]: Mark two deps that we're explicitly not linking.

chrisbobbe:
Add some comments to the Podfile on two RN-dependent pods that we *could* link, but don't, for reasons.


gnprice:
Merged -- thanks!

PR #4114, opened 2020-05-15
Remove direct `react-navigation-tabs` dependency.

chrisbobbe:
This is almost NFC, considering how thin `react-navigation`'s wrapper around `react-navigation-tabs`'s `createBottomTabNavigator` is.

    ```
    get createBottomTabNavigator() {
      return require('react-navigation-tabs').createBottomTabNavigator;
    },
    ```


chrisbobbe:
Oh, oops. There's one more usage of `react-navigation-tabs` — should be able to take care of that the same way; I'll update the PR and re-post here.

gnprice:
I like the reasoning for removing the direct dependency on react-navigation-tabs.

We should also remove the corresponding libdef in flow-typed/npm/.

For the fixmes, compare the fixmes that were removed in 0d2066f41 when we switched to using react-navigation-tabs. They were narrower, which meant we got type-checking on the rest of what we were passing to react-navigation there. That seems worth preserving -- there's a bunch of other options involved that it's good to check on.

In the analysis of what's causing the need for the fixmes:

    One summary of the error on, e.g., `HomeTab`, is, "Hey, you should
    write down the fact that HomeTab gets special treatment because it's
    passed into `createBottomTabNavigator`. It gets certain props, for
    example, like the `navigation` prop."

I don't think this is quite it, because if you look at the error it doesn't mention a name `navigation`. It seems to be entirely about the `navigationOptions` property. Which in particular is a property on the component *type*, i.e. a property on the value `HomeTab` itself, i.e. if it's a class then a static property on the class.

It's actually kind of puzzling to me what Flow is doing there, because the libdef says the property is optional, so it should be fine to simply not have it. (As indeed we don't.) I suspect the story involves something like React.ComponentType<…> being handled specially and not as "an object type" at all.

I think I can fix the issue that our typed `connect` doesn't pass through the information about static properties on the component type. The main annoying part will be testing that the fix doesn't cause Flow to lose type-checking precision somewhere else.

But also, on testing one possible fix for that, I find that the react-navigation libdef doesn't actually allow a useful `navigationOptions` anyway! I tried putting
```
 class HomeTab extends PureComponent<Props> {
+  static navigationOptions = { tabBarLabel: 'Home' };
+
```
and that produces an error saying `tabBarLabel` isn't allowed. And in fact it's the same error if I cut out `connect` entirely. So... we wouldn't get useful type-checking if we actually *used* this feature of the react-navigation API, which rather lessens the motivation to fix the `connect` type so that we could try.

chrisbobbe:
Oh no! I'm about 80% sure I posted a lengthy reply to this, referring to my earlier thoughts on this extra property on the component, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-navigation.20types/near/878759). Hmm.

chrisbobbe:
> Oh no! I'm about 80% sure I posted a lengthy reply to this, referring to my earlier thoughts on this extra property on the component, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-navigation.20types/near/878759). Hmm, maybe I didn't follow through, or it got lost somehow.

Ah! I think @gnprice and I discussed it on the phone; that must have been it.

chrisbobbe:
Thanks for the review! Just pushed my latest changes.

chrisbobbe:
> We should also remove the corresponding libdef in flow-typed/npm/.

Ah, just did this, and rebased.

chrisbobbe:
I'm closing this, as we'll need to depend on it directly in react-navigation v4: https://reactnavigation.org/docs/4.x/upgrading-from-3.x/#install-the-new-packages (see the first subheader, "Install the new packages").

PR #4116, opened 2020-05-18
Remove redundant, stringly-typed fields from PmConversationData

rk-for-zulip:
Replace the excessively quirky `ids: string` and `recipients: string` fields in `PmConversationData` with a `users: Users[]` field, unconditionally containing all participants.

Partly in support of #3764 and #4035. Blocks PR #4104 (although see below).


ray-kraesig:
Somewhere around my 3.7th _[sic]_ rebase-slash-rewrite of PR #4104, I got angry and frustrated enough to do this first. This may not be the only way to keep the final version of that PR reasonable, but it's probably the best.

There is some overlap here with the commits in #4104: the commits for well-typing the relevant test code are taken from it. I believe I have addressed all of @gnprice's feedback for those commits in this version, with one notable exception (the `NULL_OBJECT` bit); I'll make a note of that as a "review" comment below. (In retrospect, I suppose these should have been split out into a separate PR.)

The final five commits (starting at `recipient [nfc]: Generalize `filterRecipients`.`) are somewhat speculative, and I shall be unsurprised if they provoke lengthy discussion. This is fine. It's certainly better that we merge the new logic in `PmConversationList` with the existing logic in `recipient.js`, but discussion on exactly how to go about doing that shouldn't block the preceding parts of the commit, and those are the parts that will unblock #4104.

src/__tests__/lib/exampleData.js
```diff
@@ -88,12 +94,12 @@ const userOrBotProperties = ({ name: _name }) => {
...
     is_admin: false,
     timezone: 'UTC',
-    user_id: randUserId(),
+    user_id: user_id ?? randUserId(),
   });
 };
 
 /** Beware! These values may not be representative. */
-export const makeUser = (args: { name?: string } = {}): User =>
+export const makeUser = (args: UserOrBotPropertiesArgs = NULL_OBJECT): User =>
```
ray-kraesig, line 103: **@&zwj;gnprice** wrote:
> What's reasoning for `NULL_OBJECT` here? `{}` is shorter and is one fewer indirection to understand.

`{}` is not exact, and so doesn't typecheck (Flow issue 2386 and/or 2977); and `Object.freeze({})`, which is the usual way to get an explicitly-exact empty object, is longer.

... I thought we had some other examples of using `NULL_OBJECT` as a default parameter in this context, but apparently not. What would you prefer?

src/__tests__/lib/exampleData.js
```diff
@@ -291,6 +292,10 @@ export const streamMessage = (extra?: $Rest<Message, {}>): Message => {
   return deepFreeze({ ...baseMessage, ...extra });
 };
 
+/** Construct a MessagesState from a list of messages. */
+export const makeMessagesState = (messages: Message[]): MessagesState =>
+  objectFromEntries(messages.map(m => [m.id, m]));
```
ray-kraesig, line 297: Feedback at https://github.com/zulip/zulip-mobile/pull/4104#discussion_r424735090 resolved by _a)_ stripping the exactness from both `MessagesState` itself (until Flow v0.111.0) and its indexer-property's value type, and _b)_ eliminating the `$FlowFixMe` entirely.

src/types.js
```diff
@@ -319,8 +319,26 @@ export type TabNavigationOptionsPropsType = {|
...
+
+  /** The most recent message in this conversation. */
   msgId: number,
+
+  /**
+   * A comma-separated sequence of the emails of the users involved in this
+   * conversation, sorted by user ID. Does not include the self-user, unless the
+   * self-user is the only recipient.
+   */
   recipients: string,
```
gnprice: Is this true?

I don't think it is. See the implementation of `normalizeRecipientsSansMe`, and `normalizeRecipients` which it invokes. We sort there, by email address.

src/pm-conversations/PmConversationList.js
```diff
@@ -29,8 +29,8 @@ export default class PmConversationList extends PureComponent<Props> {
     this.props.dispatch(doNarrow(privateNarrow(email)));
   };
 
-  handleGroupNarrow = (email: string) => {
-    this.props.dispatch(doNarrow(groupNarrow(email.split(','))));
+  handleGroupNarrow = (users: UserOrBot[]) => {
+    this.props.dispatch(doNarrow(groupNarrow(users.map(s => s.email))));
```
gnprice: The order here matters, because we use it for the key to look up the narrow in our data structures. So it's important to preserve the order used by the existing code, or else to carefully untangle to make it stop mattering.

src/pm-conversations/PmConversationList.js
```diff
@@ -24,17 +31,33 @@ type Props = $ReadOnly<{|
...
+   * self-conversation, this is the self-user.)
+   *
+   * If this is a group PM, null.
+   */
+  getOtherUser = (users: UserOrBot[]): UserOrBot | null => {
+    const ownUserId = this.props.ownUser.user_id;
+    if (users.length <= 2) {
+      return users.filter(u => u.user_id !== ownUserId)[0];
+    }
+    return null;
```
gnprice: Hmm, does this work in the self-1:1 case? It looks like it'd end up returning `[][0]`, which is `undefined`.

src/pm-conversations/PmConversationList.js
gnprice: Also, doesn't this now see the full set of users? Where previously it saw the set excluding self.

src/pm-conversations/PmConversationList.js
```diff
@@ -24,17 +31,33 @@ type Props = $ReadOnly<{|
...
   handleUserNarrow = (email: string) => {
     this.props.dispatch(doNarrow(privateNarrow(email)));
   };
 
   handleGroupNarrow = (users: UserOrBot[]) => {
-    this.props.dispatch(doNarrow(groupNarrow(users.map(s => s.email))));
+    const { dispatch, ownUser } = this.props;
+    const emails = users.filter(u => u.user_id !== ownUser.user_id).map(u => u.email);
+    dispatch(doNarrow(groupNarrow(emails)));
+  };
```
gnprice: Ah, I see. This change belongs in the earlier commit where we changed what set of users gets passed here.

src/unread/UnreadCards.js
```diff
@@ -34,14 +34,12 @@ class UnreadCards extends PureComponent<Props> {
...
-    type Card =
-      | UnreadStreamItem
-      | { key: 'private', data: Array<$PropertyType<PmConversationList, 'props'>> };
+    const { conversations, unreadStreamsAndTopics } = this.props;
+    type Card = UnreadStreamItem | { key: 'private', data: Array<PmConversationData[]> };
     const unreadCards: Array<Card> = [
       {
         key: 'private',
-        data: [{ conversations, ...restProps }],
+        data: [conversations],
```
gnprice, line 42: Hmm, can this really work with how this data gets used below?
```
          section.key === 'private' ? (
            <PmConversationList {...item} />
```

How does it know to call the prop `conversations`?

src/__tests__/lib/exampleData.js
gnprice: Hmm, I see. Seems an OK workaround for that, then.

src/reduxTypes.js
```diff
@@ -153,9 +153,10 @@ export type FlagName = $Keys<FlagsState>;
  * See also `NarrowsState`, which is an index on this data that identifies
  * messages belonging to a given narrow.
  */
-export type MessagesState = {|
-  [id: number]: $Exact<Message>,
-|};
+// This type cannot be made exact until Flow v0.111.0.
+export type MessagesState = {
+  [id: number]: Message,
```
gnprice, line 158: Instead of removing exactness here, why not add it on `Message` itself? (And remove the then-redundant mention of it here.)

Otherwise we're regressing on how tightly we type-check.

For a quick experiment I made `Message` exact. Flow reported only one error -- and in fact it looks like a real one, which would have been caught in 3dc2d8fe7 if we'd had `Message` exact then.

gnprice:
> The final five commits (starting at `` recipient [nfc]: Generalize `filterRecipients`. ``) are somewhat speculative, and I shall be unsurprised if they provoke lengthy discussion. This is fine.

As I think you guessed, I am not a fan of these last few changes. :slightly_smiling_face: They make this code significantly more complex, and I don't see a corresponding gain.

src/types.js
ray-kraesig: Whoops! Will fix.

src/pm-conversations/PmConversationList.js
ray-kraesig: That's odd – I specifically remember seeing that issue and fixing it. (The second one, not the first; the first is a carryover of my mistake in describing `recipients`.) Will fix again.

src/pm-conversations/PmConversationList.js
ray-kraesig: – oh, there it is. I must have squashed the fix into the wrong commit. :sweat:

src/unread/UnreadCards.js
ray-kraesig: This can't possibly have worked. I think when I tested it last night I wasn't actually running the right version of the code. :sweat:

src/reduxTypes.js
ray-kraesig: `Message` really *isn't* exact, though: it's full of unlisted fields from (possibly future versions of) the server. We shouldn't make the claim that it is. (The bug you cite is a good argument that `eg.streamMessage` and friends should take `$Exact<Message>`, though.)

I assume what we really want to do is to deconstruct the inexact, server-supplied `ApiMessage` object (to coin a name) into a fresh `LocalMessage` object that really is exact. But not, I think, today. 

(I'm pretty sure this has come up before. I know I've documented it in `Message` at some point, but I suppose that hasn't made it into a PR yet. I'll try to dig it up and submit it separately. Or just rewrite it.)

ray-kraesig:
The new version may be politely described as less ambitious. It may be less politely described as "just pushing the junk from one room into another room".

The semantic mess that is `normalizeRecipientsSansMe` is no longer in a place where it gets in the way of work on #4104, but it hasn't been eliminated – just moved, with a new coat of paint on it.
 
(But at least we're no longer passing two different stringly-typed values in.)

gnprice:
The main goals of this PR have been taken care of separately since then, with the PR series up to #4356.

I've just sent #4358 with (rebased versions of) the parts of this PR that I think are useful and still relevant, plus some related new changes. Closing this one.

PR #4122, opened 2020-05-20
Improve typing of `normalizeRecipients`

rk-for-zulip:
There's a `TODO` next to `normalizeRecipients`, so do it (fixing a minor bug in the process).

Also, add documentation for it. (The documentation is based on its original implementation and use in a2419662d431f2fe4f9a17f9f618371197fe1b20, the comment noted in the commit. Its use has expanded beyond that... although not necessarily for the better.)


ray-kraesig:
Withdrawn pending analysis of test failures.

rk-for-zulip:
(Test failure was due to the mock `Message` in the test missing fields. Revised.)

Issue #3588, opened 2019-08-13
On iOS Zulip will log out of realm/organization after force closing app and re opening.


dalanmiller:
Currently using a self-hosted instance with both Android and iOS. Android runs fine at the moment but on iOS the app will require logging into the organization every time the app is closed and started or opened after a long period of time.

How can I try to debug this issue or provide more information on what is going on?

ferman:
I am experiencing the same issues. I am on a iPhone 6s on iOS 12.4.

I was just running Zulip on an android phone and it stays logged in perfectly fine.

gnprice:
Thanks @dalanmiller and @ferman for these reports! One followup question for both of you:
* Have you been seeing this on devices where the app has been installed for a while, and previously was staying logged in just fine?
* Or have you seen it only on fresh installs that have never successfully stayed logged in?

We've also heard about this in [a chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Testers.20being.20logged.20out.20of.20app/near/779249) today, and we heard a report of it Friday in a support thread.

I reproduce this myself when I make a fresh install of v26.1.124, the latest, on an iPhone 7 with iOS 12.3. The report on Friday was on v26.0.123, so it also affects that version.

For me, the app was working just fine (staying logged in) before that fresh install. If it does only affect fresh installs (or, newly logged-into accounts on existing installs), then that's a clue because it suggests we're reading the stored data fine, and the bug specifically affects writing changes to it.

This is a pretty showstopper symptom, so we'll certainly be making a priority of getting to the bottom of it.

dalanmiller:
Confirm that we're observing iOS app version v26.1.124.
We're also running the most recent version of the Zulip docker container
available.

Le mar. 13 août 2019 à 22:40, Greg Price <notifications@github.com> a
écrit :

gnprice:
Update from some attempts to debug:
* The issue continues to reproduce 100% of the time with v26.1.124 (from TestFlight, but should be the same binary as in the App Store) on my physical iPhone 7.
* In a debug build on a simulated iPhone X, the issue *did* reproduce a few times... but then it didn't always, and now I'm finding it hard to reproduce there at all.

  That's slowed down the investigation -- it's much easier to make progress when looking at a debug build with debug logs, adding a few lines of ad-hoc logging, rerunning with that, etc. Also the uncertainty of what might be confounding variables causes me to be paranoid about resetting the state extra-clean between trials, making the iteration cycle longer.

[...]

PR #4124, opened 2020-05-25
[android] Handle receiving shares from other apps

### LINKED ISSUE #117: Handle sharing from other apps (Android)

agrawal-d:
Closes #117.

Two bugs exist as of now:
 -   uploading non-image files causes a network error
 -   User picker screen component shows the back button, event when canGoBack={false} is passed

[Discussion on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M117.20Handle.20sharing.20from.20other.20apps)

Screenshots:
image

image


agrawal-d:
I'm not sure the commit `android [nfc]: Format 'AndroidManifest.xml' acc Android Studio defaults.` is required, I'll remove it if not.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+                          @Nullable params: WritableMap) {
+        Log.d(TAG, "Sending event with shared data")
+        emit(reactContext, eventName, params)
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        WebView.setWebContentsDebuggingEnabled(true)
+        val application = application as ReactApplication
+        launchMainActivity(application as Context)
```
gnprice, line 43: Comparing this logic with `NotifyReact.kt`, one difference I notice is that this one unconditionally calls `launchMainActivity`, while that one only calls it if the app isn't already in the foreground.

Is that difference intentional? What's the effect of it?

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,467 @@
...
+    };
+  }
+  toggleChoosingRecipients = () => {
+    this.setState(state => ({ choosingRecipients: !state.choosingRecipients }));
+  };
+  handleChooseRecipients = (users: Array<User>) => {
+    this.setState({ users });
+    this.toggleChoosingRecipients();
+  };
+  finishShare() {
```
gnprice, line 279: blank lines between all these definitions, please -- it makes it a lot easier to read :slightly_smiling_face:

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,467 @@
...
+  users: Map<number, User>,
+  dispatch: Dispatch,
+  auth: Auth,
+|}>;
+
+class SharingScreen extends PureComponent<Props> {
+  render() {
+    const { dispatch, navigation, subscriptions, auth } = this.props;
+    const share: SharedData = navigation.state.params.sharedData;
+    const Tabs = createMaterialTopTabNavigator(
```
gnprice, line 411: Hmm interesting. I think this call needs to happen at the top level, outside of the `render()` method. It's basically returning a component *type*, i.e. something at the same level as `SharingScreen` itself rather than a particular instance of it. If `render` gets called twice and we get two of these as a result, I expect that will cause the old instance of the old `Tabs` to be discarded and completely replaced by an instance of the new `Tabs`, rather than just have its props updated. Which would mean any UI state is lost, like which tab is selected and what text the user has entered.

I think the `share` value can maybe be passed through like so?
```
          screen: (share) => (
// ...
        <Tabs share={share} />
```
Not sure what the API is that `react-navigation` provides here, but there ought to be *some* way to do it.

For `dispatch`, `subscriptions`, and `auth`, probably simplest to have `StreamScreen` and `PmScreen` each do their own `connect` calls.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,467 @@
...
+
+type StreamScreenState = $ReadOnly<{|
+  stream: string,
+  topic: string,
+  message: string,
+  isStreamFocused: boolean,
+  isTopicFocused: boolean,
+|}>;
+
+class StreamScreen extends React.Component<StreamScreenProps, StreamScreenState> {
```
gnprice, line 82: This name is confusing because we already have a `StreamScreen`. The computer doesn't get confused, because this isn't exported, but us human readers (and readers of things like stack traces in the future!) might be :slightly_smiling_face: 

This one is really somewhat more specific. Perhaps `StreamShareScreen`?

Also, though, it's kind of odd calling this a "screen" when it's inside something else that is a "screen". I think we already have a number of examples of that oddity in the codebase already, but perhaps we can avoid adding more. :wink: How about just `StreamShare`? Or still more explicitly: `ShareToStream`?

src/nav/navReducer.js
```diff
@@ -63,7 +63,10 @@ const rehydrate = (state, action) => {
   if (rehydratedState.users === undefined) {
     return getStateForRoute('loading');
   }
-
+  // Dont switch to main UI if sharing screen is on.
```
gnprice, line 66: there's a reason for that blank line :wink: -- keep it there, and add a similar blank line before this new bit too

In general the basic principle here is: when making a small change to some existing code, notice the style it's written in and match the existing style.

The specific rationale for these blank lines is they help organize the code visually to match its logic.

src/nav/navReducer.js
```diff
@@ -63,7 +63,10 @@ const rehydrate = (state, action) => {
...
+  if (state.routes.find(route => route.routeName === 'sharing')) {
+    return state;
+  }
```
gnprice, line 69: Hmm interesting. Can you say more about why this is needed? What's the scenario in which it takes effect?

One thing that makes me curious here is that we don't have a similar condition in this spot for notifications, even though opening a notification can mean a very similar navigation at startup.

android/app/src/main/AndroidManifest.xml
```diff
@@ -13,61 +13,73 @@
...
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data
+                    android:host="login"
+                    android:scheme="zulip" />
```
gnprice, line 52: Gosh this reformatting sure makes a lot of changes.

I like the newly consistent indentation. I'm not as much a fan of the wrapped lines, or some of the reorderings -- like this one, where logically "scheme" definitely belongs before "host".

What led to this commit? Did you find that Android Studio really wanted to reformat the file, and it was annoying to try to edit it without an auto-reformatting?

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        val host = application.reactNativeHost
+        val reactContext = host.tryGetReactInstanceManager()?.currentReactContext
+        val intent = intent
+        when {
+            intent?.action == Intent.ACTION_SEND -> {
```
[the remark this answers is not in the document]
agrawal-d: I felt so too! That is how I had originally implemented it. But Android Studio suggested to convert it to `when`, so I thought that's the proper way to do it in Kotlin.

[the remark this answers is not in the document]
gnprice: Was this with one of the "intentions" that you get behind a lightbulb icon in the left margin, and that you can see by hitting Alt+Enter?

I think Android Studio provides some of those that go in both directions :smile: . I think this may be one of them. Sometimes they're more in the vein of refactoring tools, like "ways you might refactor the code, depending on what you want and the context" -- rather than lint rules with fixers, like "ways the code is unidiomatic or bad and you can fix it".

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+                when {
+                    "text/plain" == intent.type -> handleSendText(reactContext, intent)
+                    intent.type?.startsWith("image/") == true -> handleSendImage(reactContext, intent)
```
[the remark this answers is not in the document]
gnprice: Speaking of (at https://github.com/zulip/zulip-mobile/pull/4124#discussion_r430870654 ) those Android Studio "intentions" i.e. little refactoring tools: if you hit Alt+Enter on each of these lines, or anyway on the "handleSendText" and similar calls, I think you'll see an option like "Add braces around `when` body". That will make an edit like the one I suggest at the end here.

Then if you hit Alt+Enter again, you'll see an option to take the braces out again :stuck_out_tongue: I think in fact starting from the current revision, if you hit Alt+Enter you can actually see both options suggested in the very same menu -- one for adding braces on these inner `when` bodies, and one for removing the braces on the enclosing `when` body, i.e. the brace in `… == Intent.ACTION_SEND -> {` and the one that matches it. At least, when I was looking at this in Android Studio earlier today I saw both of those at once.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        WebView.setWebContentsDebuggingEnabled(true)
+        val application = application as ReactApplication
+        launchMainActivity(application as Context)
+        val host = application.reactNativeHost
+        val reactContext = host.tryGetReactInstanceManager()?.currentReactContext
+        val intent = intent
```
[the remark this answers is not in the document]
agrawal-d: I have no idea how that happened :smile:

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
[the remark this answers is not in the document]
agrawal-d: I got a green squiggle underline with this popup:
image

agrawal-d:
I just realized why I'm no longer leaving line breaks after function definitions - the GitLens VS Code extension. It shows the function history above it, and it looks like a new line :face_with_head_bandage: 

image

src/sharing/SharingScreen.js
agrawal-d: I've extracted `createMaterialTopTabNavigator` out of `render`.

The snippet you commented does not seem to work though,

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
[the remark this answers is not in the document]
gnprice: Ah, that looks like the other `when`, though :slightly_smiling_face: . In particular "Cascade if" describes the inner `when`, the one with several different options, but not the outer one which has just one option.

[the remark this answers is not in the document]
gnprice: (And yeah, that one isn't the "intentions" thing I was thinking of -- it's definitely saying something in the spirit of "this code is unidiomatic or bad and you should do it this other way".)

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+    //   because the behavior without it only makes sense when starting from
+    //   an Activity.  Our `context` is a service, so it's required.
+    //
+    // * The flag FLAG_ACTIVITY_CLEAR_TOP is mentioned as being what the
+    //   notification manager does; so use that.  It has no effect as long
+    //   as we only have one activity; but if we add more, it will destroy
+    //   all the activities on top of the target one.
+    intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TOP
+    context.startActivity(intent)
+}
```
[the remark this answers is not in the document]
gnprice: > The only reason I used it here was the `TAG` in `Log.d.` - I wanted the TAG to be the one declared in `ReceiveSharingActivity.ky`.

Hmm. For a bit when I was looking at this originally, I'd thought maybe that was the difference (though a small one). But what the code actually says is:
```kt
    Log.d(com.zulipmobile.notifications.TAG, "Launching main activity")
```
So it uses the same "TAG" value as in the other code. :slightly_smiling_face: 

(Which in itself I would have commented on if this function were otherwise good to merge, but that's a finer-grained matter.)

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,98 @@
...
+        val host = application.reactNativeHost
+        val reactContext = host.tryGetReactInstanceManager()?.currentReactContext
+
+        val params = (if ("text/plain" == intent.type) {
+            getParamsFromTextIntent(intent)
+        } else if (intent.type?.startsWith("image/") == true) {
+            getParamsFromImageIntent(intent)
+        } else {
+            getParamsFromFileIntent(intent)
+        })
```
gnprice, line 57: Yeah, *this* one would definitely benefit from `when` :slightly_smiling_face:

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,98 @@
...
+            ?: throw Exception("Could not parse received sharing data")
```
gnprice, line 58: If for some reason we get one of these in an unexpected shape, we don't want to crash -- instead we should just ignore it, after logging a warning.

See how the `onReceived` in `FCMPushNotifications.kt` catches `FcmMessageParseException`, and then the code in `FcmMessage.kt` which parses FCM messages makes sure to throw exceptions of that type if there's a parse error.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,98 @@
...
+        return params
+    }
+
+    private fun getParamsFromFileIntent(intent: Intent): WritableMap? {
+        val uri = intent.getParcelableExtra(Intent.EXTRA_STREAM) as? Uri ?: return null
+        val params = Arguments.createMap()
+        params.putString("type", "file")
+        params.putString("sharedFileUri", uri.toString())
+        return params
+    }
```
gnprice, line 90: I think we can deduplicate some more of this by having a method like `getParamsFromIntent` -- which contains the `"text/plain" == intent.type` etc. logic, and also contains the contents of these three methods.

That way the `params = Arguments.createMap()`, and the couple of lines of overhead for making a separate method, get shared across the three cases. Also the new method has a bit more clear and self-contained of an interface than each of these three.

Then `handleSend` would just call that method, like `val params = getParamsFromIntent(intent)`.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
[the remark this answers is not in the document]
gnprice: Quite helpful, thanks!

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        (intent.getParcelableExtra<Parcelable>(Intent.EXTRA_STREAM) as? Uri)?.let {
+            val params = Arguments.createMap()
+            params.putString("type", "image")
+            params.putString("sharedImageUri", it.toString())
+            if (reactContext == null) {
+                SharingModule.initialSharedData = params
+            } else {
+                sendEvent(reactContext, "shareReceived", params)
+            }
+        }
```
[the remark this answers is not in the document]
gnprice: No -- that wouldn't get uniformity, because more things use "URL" anyway.

Quick empirical demo:
```
$ grep -Rhi url node_modules/react-native | wc -l
2142

$ grep -Rhi uri node_modules/react-native | wc -l
829
```

A lot of people chose "URI" as the name for various things because it was officially the standard for a number of years. But it was never the name most people would normally use in a sentence -- most people said "URL", and would say "URI" only if they were really trying hard to be pedantically standards-correct. On top of that, the official line for all those years was that "URL" totally *was* the name of a subset of URIs that just happened to encompass virtually all of the URIs that anyone ever encountered in practice.

So now that even the relevant current standard recommends dropping "URI", it really has nothing going for it at all outside of contexts with specific legacy constraints.

In some places we do have those legacy constraints: so for example in the Zulip server API, which is hard to make changes to because there are lots of existing servers and clients, there's a field called `realm_uri`. Internally we have things we call `realmUri` because they correspond very directly to a value of that field.

But we shouldn't make *new* API choices that propagate the old confusing name. "URL" is the winner. Fighting it just prolongs the confusion.

Concretely, this key `sharedImageUri` in a data blob we're sending here over to JS is a new API choice, in the internal API between the Kotlin and JS sides of our app. So we should use the standard name for that new choice.

For a local variable, it doesn't matter as much. But the only way to get toward uniformity is to make a habit of picking the standard name, when there isn't a strong reason to go the other way.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+                SharingModule.initialSharedData = params
+            } else {
+                sendEvent(reactContext, "shareReceived", params)
+            }
+
+        }
+    }
+
+    private fun handleSendImage(reactContext: ReactContext?, intent: Intent) {
+        (intent.getParcelableExtra<Parcelable>(Intent.EXTRA_STREAM) as? Uri)?.let {
```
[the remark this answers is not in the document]
gnprice: Looks like the version in this revision is instead:
```kt
        val uri = intent.getParcelableExtra(Intent.EXTRA_STREAM) as? Uri ?: return null
```
with only the cast `as? Uri`, and no type parameter `<Uri>`.

I wasn't sure what that would do, so I hit Ctrl+Shift+P to get the type of the expression `intent.getParcelableExtra(Intent.EXTRA_STREAM)`. That informed me the type was `Uri!` (which means "`Uri` or `Uri?`, sorry it comes from a Java API and the Java type system doesn't express that distinction".) So I *think* that means the type parameter gets inferred as `Uri`, and that gets passed through, so that the cast inside the implementation is to `Uri`, and our own cast doesn't actually do anything at runtime.

Given that that's the case, probably best to be explicit and have the code match what's really going on:
```kt
        val uri = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM) ?: return null
```

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
gnprice: The logic is still different from what's in `NotifyReact.kt`, but in a different way:
```kt
        if (null == reactContext) {
            launchMainActivity(application as Context)
            SharingModule.initialSharedData = params
        } else {
            sendEvent(reactContext, "shareReceived", params)
        }
```

If you compare to the logic in `notifyReact`, you'll find that that existing logic will call `launchMainActivity` if the status is NOT_RUNNING or BACKGROUND, and this won't. It doesn't sound like that's intentional.

This logic is pretty tricky and interacts with subtle details of Android and of RN. For that kind of logic, it's best to make a point of not writing it twice.

Ultimately the way to do that is to refactor the existing code so it can be used in both situations. Or, for development to first get the thing working, a fine shortcut can be to copy the code and make only the needed changes (like the event name "notificationOpened"), with the intent to go back and factor it out after we understand exactly what we need.

src/sharing/SharingScreen.js
gnprice: It looks like in the new revision it's in the constructor. That means we end up getting data (for `share`) from props, in the constructor:
```js
  constructor(props) {
    super(props);
    const { navigation } = this.props;
    const share: SharedData = navigation.state.params.sharedData;

    this.Tabs = createMaterialTopTabNavigator(
      {
        Stream: {
          screen: () => <ShareToStream share={share} />,
```

That isn't good either -- React components don't ordinarily. We should try to write this in a way that fits smoothly with the basic React model of how things work. It's always possible to do something that deviates if it's really necessary, but we should first try to make it work properly, and if we can't then understand why not.

The react-navigation library is meant to be used with React, so I'll be kind of surprised if it really doesn't have a reasonable way to pass through React props in a normal React-compatible way.

Please take a look at the react-navigation docs (https://reactnavigation.org/docs/2.x/ for the version we're currently on), and/or the web, and/or its implementation in your `node_modules/react-navigation`, and try to find the appropriate feature of its API for this. If you're having trouble, ask on #mobile-team and say what you've looked at and tried so far.

chrisbobbe: > Please take a look at the react-navigation docs (https://reactnavigation.org/docs/2.x/ for the version we're currently on), and/or the web, and/or its implementation in your `node_modules/react-navigation`, and try to find the appropriate feature of its API for this. If you're having trouble, ask on #mobile-team and say what you've looked at and tried so far.

There's been some discussion of this, starting from Divyanshu's post [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M117.20Handle.20sharing.20from.20other.20apps/near/882122).

Looking back at my reply there, I think `screenProps` is indeed the way React Navigation would like us to handle it. But there's that inconvenient conclusion that, if you do things ["the right way"](https://reactnavigation.org/docs/2.x/common-mistakes/#explicitly-rendering-more-than-one-navigator), only explicitly rendering the root navigator in your own JSX (and no descendent navigators), whatever you pass as `screenProps` is sort of forced to exist in a global context. Unless there's something less well documented, about `screenProps`, that I'm missing.

It would be helpful if we could work out the advantages of following this "render-only-one-navigator" guideline strictly (and bringing our existing code in line with it).

One compromise (and just because it's a compromise doesn't mean it's what we should do) might be to follow their instructions under

> Alternatively, the following would also work because it exposes the `router` static on `AuthenticationScreen` and threads through the `navigation` prop:

That makes it seam like we could wire things up ourselves, and it would give us the means to pass `screenProps` at a more local level.

chrisbobbe: I've just started a #mobile-team topic about this, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60screenProps.60.20in.20react-navigation.20v2).

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+            ReactAppStatus.BACKGROUND, ReactAppStatus.FOREGROUND ->
+                // JS is running and has already reached foreground.It won't check
+                // initialSharedData again, but it will see a shareReceived event.
+                sendEvent(reactContext, "shareReceived", params)
+        }
+        when (appStatus) {
+            null, ReactAppStatus.NOT_RUNNING, ReactAppStatus.BACKGROUND ->
+                launchMainActivity(application as Context)
+            ReactAppStatus.FOREGROUND -> Unit
+        }
```
agrawal-d, line 73: For the time being, this code is a close duplicate of the one in `NotifyReact.kt`, until we decide to either make a common function for both or keep the duplicate here.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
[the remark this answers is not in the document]
agrawal-d: Changed relevant names to `uri`.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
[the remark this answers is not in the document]
agrawal-d: Done. But instead of returning `null`, it throws an exception now.

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
gnprice: Looks good!

android/app/src/main/java/com/zulipmobile/sharing/ReceiveShareActivity.kt
gnprice: Great. The one essential thing it's missing is a comment (on each copy) mentioning it's duplicated, and pointing to the other copy.

I definitely won't want to keep this code duplicated. But it will be straightforward to factor out.

src/sharing/SharingScreen.js
gnprice: Thanks for that suggestion and the demo! And good thought to move to chat for the detailed discussion.

That approach with the `router` static, and passing through the `navigation` prop, sounds good to me.

android/app/src/main/AndroidManifest.xml
```diff
@@ -13,61 +13,96 @@
...
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="image/*" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="*/*" />
+            </intent-filter>
```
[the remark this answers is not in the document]
gnprice: (I went ahead and removed them in the version I merged.)

android/app/src/main/AndroidManifest.xml
```diff
@@ -13,61 +13,96 @@
...
+                    android:scheme="zulip" />
+            </intent-filter>
+        </activity>
+
+        <activity
+            android:name=".sharing.ReceiveShareActivity"
+            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+            android:label="@string/app_name"
+            android:launchMode="singleTask"
+            android:windowSoftInputMode="adjustResize">
```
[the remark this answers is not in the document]
agrawal-d: >Are these extra attributes helpful and do they do things we want?
`windowSoftInputMode` and `configChanges` are not relevant here, because we don't care how the activity responds to the keyboard etc. But we do need `launchMode` to be `singleTask` - we want only one instance of the activity to exist at any given time.

I'll remove the useless attribs.

Reference: https://developer.android.com/guide/topics/manifest/activity-element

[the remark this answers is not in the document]
agrawal-d: >Hmm, in fact. I wonder if it might simplify things to simply add these intent filters to the main activity. After all, the behavior we're actually implementing here amounts to that in the end -- the user goes to send something, i.e. causes an intent with action android.intent.action.SEND, and what we end up doing is passing that all to the main activity.

One possible problem with that would be that we always have to launch an activity. So, in case the app is already running, we will definitely launch MainActivity for every share action. We can close it soon, but I would expect the `ZulipMobile` component to be pretty heavy in terms of resources.

agrawal-d:
I've made two changes in this new revision:
1. Used the `react-navigation` code suggested by @chrisbobbe  (Thanks!)
2. If there is no active account when receiving a share, the account picker screen is presented.

src/api/messages/sendMessage.js
```diff
@@ -9,10 +9,10 @@ export default async (
   params: {|
```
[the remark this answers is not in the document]
agrawal-d: Yeah, even I found this strange.

gnprice:
> One thing that I haven't been able to reproduce again, but that did happen, is an error in `SharingModule.kt` on the `promise.resolve` line in `getInitialSharedContent`; it said something like "Error: Map already consumed". I tried several times to get it to happen again, without success

Interesting. Here's a lead on what the error might have looked like, from some quick googling:
https://stackoverflow.com/questions/59025712/error-map-already-consumed-react-native

One guess for what that might mean is that perhaps you're sending the same `WritableMap` (hence the "Map" part) value over the bridge more than once, and perhaps that's not allowed. That seems quite plausible, and I think it'd explain an error message like that.

One thing that might affect reproducing it: I wouldn't particularly expect the behavior of the bridge to vary between debug and release mode... but it definitely *does* vary, and this is the kind of thing I could easily believe it varies in, between *remote JS debugging* and not remote JS debugging. The reason is that when remote JS debugging, the JS code of the RN environment actually runs *in the Chrome browser you're debugging from*, rather than in a JS implementation built into the app. Naturally some parts of the implementation of the bridge are written separately for talking to Chrome vs. talking to that in-app JS implementation... and so there can be subtle differences in behavior.

(Why yes, that can indeed be a very unpleasant surprise when trying to debug something! See #3588 and https://github.com/zulip/zulip-mobile/commit/5cf7f107e3a4904e461a4fcca27136b786717ce8 .)

android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
gnprice: > We always have to launch an activity.

Do we? I think this is what `android:launchMode` controls, and we can make a different choice.

In fact, I think the value `singleTask` already means it won't launch a new activity. Here's [the documentation for that attribute](https://developer.android.com/guide/topics/manifest/activity-element#lmode), and here's what it says about `singleTask`:
> The system creates the activity at the root of a new task and routes the intent to it. However, if an instance of the activity already exists, the system routes the intent to existing instance through a call to its onNewIntent() method, rather than creating a new one.

(I definitely agree we don't want to create a new MainActivity when we already have one running.)

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+            progress={sending}
+            disabled={!this.isSendButtonEnabled()}
+          />
+        </View>
+      </>
+    );
+  }
+}
+
+const ShareToPm = connect(state => ({
```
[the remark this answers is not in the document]
gnprice: I think I may have at some point suggested having some of these components in a single file. But I agree, this file is pretty long and it'd be good to apply our usual convention where each of the component classes, plus its `connect` wrapper, gets its own file.

src/types.js
```diff
@@ -324,3 +324,20 @@ export type PmConversationData = {|
...
+  type: 'image',
+  sharedImageUrl: string,
+|};
+
+export type SharedFile = {|
+  type: 'file',
+  sharedFileUrl: string,
+|};
+
+export type SharedData = SharedText | SharedImage | SharedFile;
```
gnprice, line 343: This type should have jsdoc to explain what it's about. In particular that it represents a user action of sharing something from another app.

src/sharing/index.js
```diff
@@ -0,0 +1,57 @@
...
+  /** Private. */
+  unlistenAll() {
+    while (this.unsubs.length > 0) {
+      this.unsubs.pop()();
+    }
+  }
+
+  handleShareReceived = (data: SharedData) => {
+    handleReceivedData(data, this.dispatch);
+  };
```
gnprice, line 44: This can be inlined, as an arrow function. Unlike in the case where we're passing a callback as a React prop, nothing's going to be looking to see if one value is `===` to a previous one and optimizing if it is -- which is the reason we have to have these separate named handler functions so often elsewhere.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+const SharingTopTabNavigator = createMaterialTopTabNavigator(
+  {
+    Stream: {
+      // $FlowFixMe
+      screen: ShareToStream,
+    },
+    'Private Message': {
+      // $FlowFixMe
+      screen: ShareToPm,
+    },
```
gnprice, line 445: What's the reason for these fixmes?

In general where there's a `$FlowFixMe`, we always want at least a line or so of comment identifying why it was necessary.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+    const { sharedData } = this.props.navigation.state.params;
+
+    if (sharedData.type === 'text') {
+      return message !== '' && users.length > 0;
+    }
+
+    return users.length > 0;
+  };
+
+  getUsersPreview = () => {
```
gnprice, line 316: This is basically a render method, a helper for the component's main render method. So let's put "render" in the name, like `renderUsersPreview`.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+          email={user.email}
+          fullName={user.full_name}
+          onPress={() => {}}
+        />,
+      );
+    });
+    return preview;
+  };
+
+  handleSend = async () => {
```
gnprice, line 336: Let's put the action handlers above all the render logic -- so above `isSendButtonEnabled` as well as `getUsersPreview` / `renderUsersPreview`.

src/api/messages/sendMessage.js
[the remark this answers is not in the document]
gnprice: `subject` is the old name for `topic`.

Dunno what the story is with `local_id` and `queue_id` -- it looks like those aren't documented, but I'd sure expect that they're still supported, as they're how we match up the eventual `message` / EVENT_NEW_MESSAGE event with the right outbox message to forget about. They may just never have been documented, though I thought our API-docs infrastructure was at the point now where it'd flag that. Would be good to ask in chat.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+  };
+
+  render() {
+    const { sharedData } = this.props.navigation.state.params;
+    const { stream, topic, message, isStreamFocused, isTopicFocused, sending } = this.state;
+    const narrow = streamNarrow(stream);
+
+    return (
+      <>
+        <ScrollView style={styles.wrapper} keyboardShouldPersistTaps>
```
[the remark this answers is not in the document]
agrawal-d: Changed to 'always'.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+  getUsersPreview = () => {
+    const { users } = this.state;
+
+    if (users.length === 0) {
+      return <Label text="Please choose recipients to share with" />;
+    }
+    const preview = [];
+    users.forEach((user: User) => {
+      preview.push(
+        <UserItem
```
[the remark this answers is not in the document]
agrawal-d: Added the `key` attribute.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+}
+
+type ShareToPmProps = $ReadOnly<{|
+  ...$Exact<NavigationNavigatorProps<{||}, {| params: {| sharedData: SharedData |} |}>>,
+  dispatch: Dispatch,
+  auth: Auth,
+|}>;
+
+type ShareToPmState = $ReadOnly<{|
+  users: User[],
```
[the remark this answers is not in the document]
agrawal-d: Changed to `selectedRecipients`.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+    this.setState({ sending: true });
+  };
+
+  toggleChoosingRecipients = () => {
+    this.setState(state => ({ choosingRecipients: !state.choosingRecipients }));
+  };
+
+  handleChooseRecipients = (users: Array<User>) => {
+    this.setState({ users });
+    this.toggleChoosingRecipients();
```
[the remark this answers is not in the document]
agrawal-d: I agree, it can be confusing. Removed the toggle function.

>There's also the fact that the "Choose recipients" button doesn't really work as a toggle switch between the two states, but rather to just open the modal (i.e., change the state from false to true); it gets hidden behind the modal, and therefore effectively disabled, when the modal opens. It might actually be good to reinforce this pattern by using ZulipButton's disabled prop.

The button does not get hidden behind, it's not rendered at all when `choosingRecipients` is true.

```js
    if (choosingRecipients) {
      return (
        <Modal>
          <ChooseRecipientsScreen onComplete={this.handleChooseRecipients} />
        </Modal>
      );
    }
```

So I think it's okay to not use the `disabled` prop.

src/sharing/SharingScreen.js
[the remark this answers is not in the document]
agrawal-d: Okay, I've split the two sharing components into their own files.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+  static router = SharingTopTabNavigator.router;
+
+  render() {
+    const { navigation, auth, dispatch } = this.props;
+
+    // If there is no active logged-in account, abandon the sharing attempt,
+    // and present the account picker screen to the user.
+    if (auth === undefined) {
+      dispatch(navigateToAccountPicker());
+      return <></>;
```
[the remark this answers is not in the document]
agrawal-d: Changed to `null`.

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+import StreamAutocomplete from '../autocomplete/StreamAutocomplete';
+import TopicAutocomplete from '../autocomplete/TopicAutocomplete';
+import AnimatedScaleComponent from '../animation/AnimatedScaleComponent';
+import ChooseRecipientsScreen from './ChooseRecipientsScreen';
+import { sendMessage, uploadFile } from '../api';
+import { streamNarrow } from '../utils/narrow';
+import { getAuth, tryGetAuth } from '../selectors';
+import { showToast } from '../utils/info';
+import { navigateBack, navigateToAccountPicker } from '../nav/navActions';
+import { fetchTopicsForActiveStream } from '../topics/topicActions';
```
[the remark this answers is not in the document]
agrawal-d: note to self: TODO

src/sharing/SharingScreen.js
```diff
@@ -0,0 +1,498 @@
...
+
+  handleSend = async () => {
+    this.setSending();
+    const { topic, stream, message } = this.state;
+    let messageToSend = message;
+    const { auth } = this.props;
+    const { sharedData } = this.props.navigation.state.params;
+
+    try {
+      showToast('Sending Message...');
```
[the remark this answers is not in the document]
agrawal-d: Added `_` to `showToast` calls.

>I think there are a few more strings that are also not being translated; what's passed as the placeholder prop to Input components, for example. Let's get all the user-facing strings translated that we can.


I think the Zulip `Input` translates placeholders automatically?

```js
/**
 * A light abstraction over the standard TextInput component
 * that allows us to seamlessly provide internationalization
 * capabilities and also style the component depending on
 * the platform the app is running on.
 *
 * @prop placeholder - Translated before passing to TextInput as
 *   a prop of the same name.
 */
export default class Input extends PureComponent<Props, State> {
```

src/types.js
agrawal-d: Added a short jsdoc.

src/sharing/index.js
agrawal-d: Okay, removed `handleReceivedData` and made the calls inline.

src/sharing/SharingScreen.js
agrawal-d: Removing the `$FlowFixMe` causes the following error:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/sharing/SharingScreen.js:23:15

Cannot call createMaterialTopTabNavigator with object literal bound to routeConfigs because in property Stream.screen:
 • Either property navigationOptions is missing in React.ComponentType [1] but exists in
   withOptionalNavigationOptions [2].
 • Or React.ComponentType [1] is incompatible with React.StatelessFunctionalComponent [3].

     src/sharing/SharingScreen.js
      18│ |}>;
      19│
      20│ const SharingTopTabNavigator = createMaterialTopTabNavigator(
      21│   {
      22│     Stream: {
      23│       screen: ShareToStream,
      24│     },
      25│     'Private Message': {
      26│       // $FlowFixMe
      27│       screen: ShareToPm,
      28│     },
      29│   },
      30│   {
      31│     tabBarPosition: 'top',
      32│     animationEnabled: true,

     flow-typed/npm/react-navigation_v2.x.x.js
 [2] 313│     withOptionalNavigationOptions<Options>;
        :
 [3] 628│   > = React$StatelessFunctionalComponent<{
     629│     ...Props,
     630│     ...NavigationContainerProps<State, Options>,
     631│   }> &

     src/react-redux.js
 [1]  98│ ): C => ComponentType<$ReadOnly<OwnProps<C, SP>>> {
```

@chrisbobbe do you know what might be causing this? I can't figure out. I've added a generic comment:
```
      // Requires additional props that we dont need. $FlowFixMe.
```

src/sharing/SharingScreen.js
agrawal-d: Changed to `renderUsersPreview`.

src/sharing/SharingScreen.js
agrawal-d: Okay, moved `renderUsersPreview` above. Looks more uniform now :)

agrawal-d:
Thanks for the review, @chrisbobbe, @gnprice! I've made the changes you suggested.

>One thing that I haven't been able to reproduce again, but that did happen, is an error in SharingModule.kt on the promise.resolve line in getInitialSharedContent; it said something like "Error: Map already consumed". I tried several times to get it to happen again, without success; it could have been something that only happens in development. But keep an eye out in case you see something like this; if you do, maybe you'll have better luck than me in reproducing it. I haven't found much of anything useful by Googling the error either.

Yes, I've encountered this error several times - I've only encountered it when remote debugging was enabled, and never when running the app without remote JS debugging.

src/sharing/SharingScreen.js
chrisbobbe: I think https://github.com/zulip/zulip-mobile/pull/4114#issuecomment-634255590 is our latest knowledge on this error. Thanks, Greg, for writing that up! 🙂

src/api/messages/sendMessage.js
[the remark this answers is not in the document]
chrisbobbe: > `subject` is the old name for `topic`.

True, I know, but it would be good to know more; maybe we can start using the new name (maybe deeper changes are needed to allow this?), when was the switch made, etc.

src/sharing/SharingScreen.js
gnprice: Ah, and that in turn leads me to this existing fixme:
```
$ rg -C2 FixMe.*nav 
src/nav/AppNavigator.js
34-
35-export default createStackNavigator(
36:  // $FlowFixMe react-navigation types :-/ -- see a36814e80
37-  {
38-    account: { screen: AccountPickScreen },
```

Which, when I try removing it, produces exactly the same kind of error as @agrawal-d quoted above.

That more recent writeup has more information than the commit mentioned in that fixme. But that certainly serves as an example of how these explanations don't need to be long :wink: .

A couple of smaller points on what goes in the fixme:
* The keyword `$FlowFixMe` goes at the start of the line, right after the `// `. That helps it stand out when scanning the code by eye. (Same thing goes for TODO, or anything else of that ilk.)
* Thinking about it a bit more, I think the key information I want in the explanation comes down to: why haven't we just already fixed it? So with "Requires additional props that we don't need", the natural next thing is: OK, let's make whatever's requiring those stop requiring them, then. But then the answer to that is: ah but it's over in the react-navigation library (or anyway its types, specified in its libdef). That doesn't mean we *can't* fix it -- it's just code, we can change it -- but it's more work. So identifying a specific third-party place as the location of the problem is a common kind of explanation for these.

chrisbobbe: > Ah, and that in turn leads me to this existing fixme: [...]

One minor point about that example is that the `$FlowFixMe` is less precise than it could be; in this PR and https://github.com/zulip/zulip-mobile/pull/4114/files (in the latter, at Greg's suggestion) we get some nice precision by putting it directly above the `screen` property in all those places. (I just tried that with one case in AppNavigator.js and commented out the other cases, and it was fine, as we'd expect.)

But adding 29 `$FlowFixMe`s might be cumbersome, and we don't use anything other than the `screen` property. In fact, I'm pretty sure (untested though) the API would allow us to do this for all of them there:

```diff
- account: { screen: AccountPickScreen },
+ account: AccountPickScreen,
```

src/api/messages/sendMessage.js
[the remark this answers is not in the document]
chrisbobbe: Tim responded [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/https.3A.2F.2Fzulip.2Ecom.2Fapi.2Fsend-message/near/915346).

agrawal-d:
@gnprice based on your message on [CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Android.20activity.20to.20handle.20intent), I've modified the code, so that the handling of sharing is done in MainActivity itself.

I've pushed to GitHub, but due to some issues with GitHub today, the PR is not getting updated, even after multiple force pushes. However, I am able to compare the changes from here: https://github.com/zulip/zulip-mobile/compare/master...agrawal-d:sharing

gnprice:
Reviewing this new version now -- sorry for the delay.

> I've pushed to GitHub, but due to some issues with GitHub today, the PR is not getting updated, even after multiple force pushes. However, I am able to compare the changes from here:

Huh -- that is odd! Good to mention it and include that alternate way to see the changes.

It looks like it eventually did update, fortunately. Here's what I have locally, after using zulip:tools/reset-to-pull-request to fetch this PR:
`$ git log --oneline --reverse upstream/master..pr/4124`
e0df13759 android: Refactor MainActivity to Kotlin.
b8b1462dc android: Move sharing logic to MainActivity.
a4f7e0e6b topic actions [nfc]: Rename 'fetchTopicsForActiveStream' to 'fetchTopicsForStream'.
572b9a92e send message: Make certain fields optional.
d344d16db (pr/4124) android: Receive and handle shares from other apps.

And those are the same commit IDs as I see on this PR webpage, and the same as I see on the comparison page you linked to.

One GitHub bug (or class of bugs?) that I regularly run into which potentially explains your experience: when I have a PR or issue page open in a browser tab and then something happens that updates the PR or issue, the page *often* updates in place in the existing tab... but very often it doesn't. (In fact I was going to write that it "usually" updates, and then realized I'm not sure the success rate is even over 50%. :disappointed: ) So if you were watching it in an existing tab, then your experience would be another example of that.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.os.Bundle
+import android.webkit.WebView
+import com.facebook.react.ReactActivity
+
+open class MainActivity : ReactActivity() {
+    /**
+     * Returns the name of the main component registered from JavaScript.
+     * This is used to schedule rendering of the component.
+     */
+    override fun getMainComponentName(): String? {
```
gnprice, line 26: Let's make it `String` -- that seems more true to the intent of the method.

(I'm guessing that the Android Studio auto-conversion went for `String?`, which is natural because Java doesn't make the distinction. The cleanup to `String` can be in a followup commit -- it's fine and good for that sort of cleanup to make the Kotlin more idiomatic to be in a separate commit from the conversion.)

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -1,8 +1,22 @@
...
+import androidx.annotation.Nullable
 import com.facebook.react.ReactActivity
+import com.facebook.react.ReactApplication
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.bridge.WritableMap
+import com.zulipmobile.notifications.*
+import com.zulipmobile.sharing.SharingModule
+
+const val TAG = "MainActivityShare"
```
gnprice, line 18: Should just be "MainActivity", because it's global to this file so it's the natural thing that other logging calls in this file will use. Or leave it specific to sharing but name it something specific like `SHARE_TAG`.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -13,8 +27,81 @@ open class MainActivity : ReactActivity() {
...
+        val application = application as ReactApplication
+        val host = application.reactNativeHost
+        val reactContext = host.tryGetReactInstanceManager()?.currentReactContext
+        val appStatus = reactContext?.appStatus
+        Log.d(TAG, "app status is $appStatus")
+        when (appStatus) {
+            null, ReactAppStatus.NOT_RUNNING ->
+                // Either there's no JS environment running, or we haven't yet reached
+                // foreground.  Expect the app to check initialSharedData on launch.
+            {
```
gnprice, line 68: nit: open-brace at end of previous line, after `->` (like in other examples)

src/topics/topicActions.js
```diff
@@ -19,7 +19,7 @@ export const fetchTopics = (streamId: number) => async (dispatch: Dispatch, getS
   dispatch(initTopics(topics, streamId));
 };
 
-export const fetchTopicsForActiveStream = (narrow: Narrow) => async (
+export const fetchTopicsForStream = (narrow: Narrow) => async (
```
gnprice, line 22: Ah, thanks for this cleanup! You may find 5542b01da informative for what that was about.

gnprice: In commit message: summary line is too long. Main solution is to drop the old or the new name (or both, and say something like "Fix misleading "...ActiveStream" name.")

gnprice:
OK, merged the following commits!
bdce7d0a9 android: Refactor MainActivity to Kotlin.
c8b56e182 android [nfc]: Tighten return type on a converted-from-Java method.
f9597fae5 android: Move sharing logic to MainActivity.
e9ebc89dc android [nfc]: Inline the sendEvent helper.
c13f889ad android [nfc]: Cut a bit of redundant logging in share-receiving code.
d8be9dcb9 android [nfc]: Push a try/catch into initializer expression.
64ca47360 topic actions [nfc]: Rename fetchTopicsForStream.
b4692e321 api: Mark certain fields optional in sendMessage.

These correspond to the first N-1 commits of this PR, plus a few tiny commits of my own.

static/translations/messages_en.json
```diff
@@ -213,5 +213,13 @@
   "🚌 Commuting": "🚌 Commuting",
   "🤒 Out sick": "🤒 Out sick",
   "🌴 Vacationing": "🌴 Vacationing",
-  "🏠 Working remotely": "🏠 Working remotely"
+  "🏠 Working remotely": "🏠 Working remotely",
```
gnprice, line 216: nit: add items somewhere other than the end of the list, to avoid noise in the diff (because JSON doesn't permit a comma after the last item)

src/types.js
```diff
@@ -324,3 +324,24 @@ export type PmConversationData = {|
...
+export type SharedFile = {|
+  type: 'file',
+  sharedFileUrl: string,
+|};
+
+/**
+ * The type of data that can be shared to this app from 3rd party apps - either
+ * plain text or an image or a file.
+ */
+export type SharedData = SharedText | SharedImage | SharedFile;
```
gnprice, line 347:
```suggestion
/**
 * The data we get when the user "shares" to Zulip from another app.
 */
export type SharedData = SharedText | SharedImage | SharedFile;
```

One thing I want to communicate in this jsdoc is that this is the data we *got* from some act of sharing. The "can be shared" sounds like it's something that might in the future be shared, or might not.

One thing that helps tighten up the jsdoc -- makes it shorter, and perhaps also a bit easier to write -- is that the jsdoc on a type focuses on *a value of the type*. Like how `PmConversationData` just above is a "Summary of a PM conversation", and `Identity` is "An identity belonging to this user in some Zulip org", etc. So no need for "The type of …"; and that perhaps helps with the motivation for "can be", too.

The "either plain text or …" is an implementation detail and so can just be left out. That also avoids a risk of another alternative being added later and the jsdoc not getting updated :wink:

src/sharing/index.js
gnprice: Sorry, this was ambiguous -- I meant that this function `handleShareReceived` could be inlined :slightly_smiling_face: 

This is fine too, though; IIRC I actually waffled on whether I wanted to inline `handleShareReceived`, or inline `handleReceivedData` at its two call sites including this one.

src/nav/navReducer.js
gnprice: Huh! That is peculiar. It leaves me still curious what's different between this and notifications that would make it needed here and not there.

I don't see this mentioned in the group PM thread with you and me and Chris. Probably #mobile-team would be the best place for any further debugging, in any case.

One possibility I wonder about is that perhaps it's related to how in a previous version of this PR there was a separate activity. Now that that's gone, the way the rest of the system works is more like it does for notifications.

gnprice:
OK, I am merging this! Leaving out just two things:
* The uncommenting of the bit of the manifest that would actually enable the feature. (We'll definitely need to fix that crash/red-screen mentioned above, at least, before taking that step.)
* The bit in the rehydrate handler of `navReducer`. I'm still puzzled about whether it's really necessary -- and as described in my last few comments above (not the code-inline comments but the top-level ones), there are some bits of odd nav-related behavior in the UX, and I wonder if one or more of those are related. But in particular it broke tests :slightly_smiling_face: so I wasn't going to merge it in that state. We can always add it later along with updates to those tests.

This PR thread has quite a lot in it, so hopefully merging this will give us a clean baseline for another PR, or handful of PRs, to complete the work.

Note that a number of the code comments above from this round of review still apply to the code I'm merging. So I'll want to see those taken care of in a followup PR before we consider this feature complete.

Thanks again @agrawal-d for all your work on this, and looking forward to seeing it across the finish line!

PR #4125, opened 2020-05-26
[WIP] Add feature to enable reply with @-mention.

### LINKED ISSUE #3436: Provide "reply with mention" option

agrawal-d:
Closes #3436.

Todo
- [x] Add the @-mention to the message input.
- [ ] Show the keyboard automatically.


chrisbobbe:
I'm seeing some strange behavior in the WebView when I press "Reply with mention" on iOS; it's spending a really long time in the "LOADING" state; details [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/WebView.20.22LOADING.22.20state/near/890935).

agrawal-d:
Needs clean up of the edit-message architecture first, as suggested by Greg [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M2757.20Cannot.20cancel.20editing.20messages/near/879116)

gnprice:
> Needs clean up of the edit-message architecture first, as suggested by Greg [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M2757.20Cannot.20cancel.20editing.20messages/near/879116).

Aka #4141, which I just filed a moment ago.

PR #4126, opened 2020-05-27
RN v60 upgrade preparation

chrisbobbe:
These commits are meant to be mergeable before the RN v60 upgrade. It'll make a main PR that does the upgrade smaller and easier to reason about.


package.json
```diff
@@ -65,7 +65,7 @@
     "react-native-safe-area": "^0.4.1",
     "react-native-screens": "^1.0.0-alpha.23",
     "react-native-simple-toast": "^1.0.0",
-    "react-native-sound": "^0.10.9",
+    "react-native-sound": "~0.10.9",
```
gnprice, line 68: I think the existing `^0.10.9` already has the same meaning.

You can try it empirically here:
https://semver.npmjs.com/

android/app/build.gradle
```diff
@@ -22,6 +22,9 @@ import com.android.build.OutputFile
  *   // the entry file for bundle generation
  *   entryFile: "index.android.js",
  *
+ *   // https://facebook.github.io/react-native/docs/performance#enable-the-ram-format
+ *   bundleCommand: "ram-bundle",
```
gnprice, line 26: I was curious what this was about, so I followed the link to read the docs.

Hilariously, the link is broken already. Here's a permalink for what it originally linked to:
https://reactnative.dev/docs/0.60/performance#enable-the-ram-format
and it looks like this is its new home:
https://reactnative.dev/docs/ram-bundles-inline-requires#enable-the-ram-format

(Regrettably the versioned URLs don't seem to work for the current version -- so for the current version of the docs it's impossible to create a link that both works today and will work in the future. But if you're reading this in the future, stick an `0.62/` in there.)

Let's use the updated URL. Probably that means in a future upgrade there'll be a patch that we've already taken and just conveniently drops out.

Meanwhile my conclusion from reading that bit of docs is that this won't be a priority for us to do (it requires a lot of setup and tuning to make it really a net positive), and it's obsoleted by Hermes so we'll probably be on that before we'd turn our attention to try to use this feature.

android/app/proguard-rules.pro
```diff
@@ -9,13 +9,6 @@
 
 # Add any project specific keep options here:
 
-# If your project uses WebView with JS, uncomment the following
-# and specify the fully qualified class name to the JavaScript interface
-# class:
-#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
-#   public *;
-#}
-
```
gnprice, line 18: Technically, this should happen after the upgrade commit because RN
    v0.60.0 newly omits WebView in RN core. But we're not using the
    WebView in RN core, we're using `react-native-webview`, from
    `react-native-community`. So the absence of this hint will be fine.

Also it will be fine because it's a hint we've been ignoring all this time anyway :wink: 

(I don't even really understand what it's trying to say.)

android/app/build.gradle
gnprice: > Probably that means in a future upgrade there'll be a patch that we've already taken and just conveniently drops out.

Ha, nope, it's still broken in master :shrug:

package.json
chrisbobbe: Ah OK! I'd mistaken the caret's behavior here as the caret's behavior if the leftmost spot wasn't a zero.

From a relevant-looking [doc](https://docs.npmjs.com/misc/semver):
> Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4
>
> Allows changes that do not modify the left-most non-zero digit in the `[major, minor, patch]` tuple.

chrisbobbe:
Fantastic, I'm so pleased to have much less to work with at a time, on my `rn-60` branch! I'll try to get that PR put together tomorrow.

Issue #4120, opened 2020-05-20
Upgrade to ESLint 6.8.0


chrisbobbe:
EDIT: See https://github.com/zulip/zulip-mobile/issues/4120#issuecomment-632911819, which will make much of these details obsolete — let's go for 6.8.0, instead of 6.1.0, if we can.

-------

Upgrading ESLint to 6.1.0 is included in the [RN v0.59.10 to v0.60.0 upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.59.10&to=0.60.0). I've tried doing this before and after the main upgrade commit, and I don't see anything new being flagged, so I think we can treat this as possible to do at any time. As with the Babel-related upgrades there, I haven't been able to trace this to a particular commit in `template/package.json` in the react-native repo. (🤔)

I think the version of `eslint-config-airbnb` we want is ~18.0.0. Both 18.0.0 and 18.0.1 list ESLint under peerDependencies as "^5.16.0 || ^6.1.0". (https://github.com/airbnb/javascript/blob/a204cdf38a3e67a405c191773456d18296a9b37b/packages/eslint-config-airbnb/package.json)

18.1.0 and above have ESLint under peerDependencies as "^5.16.0 || ^6.8.0". (https://github.com/airbnb/javascript/blob/a3355f90c4b52445dbe828fd4e71a7e7d12ef202/packages/eslint-config-airbnb/package.json)

Let's see if we can emulate 58cb26e75 and do the upgrade with a simple script; this is the most easily reproducible way to do it.

One thing that should help simplify things is to remove our direct dependency on `eslint-config-airbnb-base`. We don't ever import it ourselves, and `eslint-config-airbnb` already has it, since it's listed as a `dependency` by that package. It's never been listed there as a `peerDependency`, even as far back as v15.0.2 (https://github.com/airbnb/javascript/blob/c68fa632624f088b4bbe81192a50554b0c6af970/packages/eslint-config-airbnb/package.json), the version we were on when we added `eslint-config-airbnb-base` as a direct dependency, in 984f77157. 984f77157 didn't explain the addition, and the main focus of that commit was elsewhere, so it seems unlikely that it was well-considered.

Running `rm -rf node_modules && yarn` after removing `eslint-config-airbnb-base` gives no errors.

chrisbobbe:
One diff to `package.json` that gets `rm -rf node_modules && yarn` working with no errors at all is this:

```diff
     "babel-eslint": "^10.0.2",
[...]

PR #4128, opened 2020-05-29
ESLint 6.8.0 upgrade

chrisbobbe:
Closes: #4120

This was started with the RN v60 upgrade (#3548) in mind, since the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.59.10&to=0.60.6) had us going to ESLint 6 from 5.

But running

`rm -rf node_modules && yarn && tools/test lint prettier --full`

(yes, newly ensuring to use `--full`! see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4120.20ESLint.206.2E8.2E0.20upgrade/near/888186) for more on that)

with these commits positioned before or after the RN 60 upgrade gives no warnings or errors whatsoever. So it seems like we're in good shape to get this done at any time, and no mention is made of the RN v60 upgrade in this revision.


chrisbobbe:
@gnprice If you'd like the reorganization of those Airbnb overrides to look different, please feel free to change as you see fit, if you prefer that to a round-trip review. (Of course I'll make changes, as requested, if you prefer that, too.) I do always examine any changes you make before merging, when you mention them, to increase my chances of doing things better next time. 🙂

chrisbobbe:
Oops! As also noted at https://github.com/zulip/zulip-mobile/pull/4133#issuecomment-637219322, I'll fix this to say `Fixes: #...` instead of `Closes: #...`.

chrisbobbe:
> Oops! As also noted at [#4133 (comment)](https://github.com/zulip/zulip-mobile/pull/4133#issuecomment-637219322), I'll fix this to say `Fixes: #...` instead of `Closes: #...`.

Done. I also rebased and fixed some conflicts.

gnprice:
Thanks @chrisbobbe ! Just read this, and it looks good. The main thing it needs is a rebase.

I might make some adjustments to how the eslintrc is organized. I agree that's better to just do rather than round-trip through a review.

Small comments:

#### cea62b754 tools/test lint: Make globs passed to `prettier-eslint` absolute.

There was a chat discussion about this, right? May be worth linking to in the commit message, to help explain the syntax for any reader unfamiliar with this sort of Bash arcana.

#### 859fee6e7 deps: Remove direct `eslint-config-airbnb-base` dependency.

    Also, run `yarn yarn-deduplicate && yarn`, as prompted by
    `tools/test deps`.

There's no change to `yarn.lock` here, so it seems like this change wound up in a different commit.

... Hmm, in fact -- in an earlier commit in the series:

#### 9c6aebb2f deps: Upgrade babel-eslint to ^10.0.3.

the `yarn.lock` changes look like they introduce duplicates which I expect `yarn yarn-deduplicate` would remove.

chrisbobbe:

> There was a chat discussion about this, right?

Ah, yes—thanks for the reminder! Added a link.

About `yarn yarn-deduplicate && yarn`:

Yup. Looks like I'd marked that I'd run that command on...both commits. Sure enough, running it at the `babel-eslint ^10.0.3` commit resulted in some changes, so, I committed those, and `tools/test deps` didn't complain at the `eslint-config-airbnb-base` commit, so I removed that bit of the commit message.

chrisbobbe:
OK, just pushed my latest changes, for another review. 🙂

gnprice:
Looks great, thanks -- merged!

PR #4129, opened 2020-05-29
uploads: Show file upload progress.

### LINKED ISSUE #2684: Image upload: Show progress feedback

agrawal-d:
When a user chooses a file for upload, show a local echo which
    indicates that the file is being uploaded. When the upload
    finishes, replace the echo with the actual uploaded file, and send
    that message to the server. If the upload fails, remove the local
    echo.

Todos:
- [x] Show upload progress.
- [ ] Translate the '[Uploading file...]' string.

upload-status

    
Closes #2684.


agrawal-d:
upload-status

src/types.js
```diff
@@ -201,6 +201,12 @@ export type Outbox = {|
   subject: string,
   timestamp: number,
   type: 'stream' | 'private',
+
+  /**
+   * Whether the message is for local echo only. If true, the message is not
+   * sent to the server.
+   */
+  isLocalEcho?: boolean,
```
chrisbobbe, line 209: Hmm, I think this needs a more precise name. The whole architecture for the "outbox" is kind of an implementation of the "local echo" concept; we've got a brief mention of "local echo" in a doc [here](https://github.com/zulip/zulip-mobile/blob/8bf48939771a4eacc8a170a4620ef2f01ff97d15/docs/glossary.md#local-echo).

At first, I thought, maybe, `isLocalEchoOnly`? But maybe that requires more effort than necessary to understand what it means.

I wonder about something like `deferSending`?

agrawal-d: `deferSending` sounds good. Changed.

src/outbox/outboxActions.js
```diff
@@ -64,6 +64,10 @@ export const trySendMessages = (dispatch: Dispatch, getState: GetState): boolean
         return; // i.e., continue
       }
 
+      if (item.deferSending !== undefined && item.deferSending === true) {
```
chrisbobbe, line 67: Just `if (item.deferSending) {` should be fine here.

src/outbox/outboxReducer.js
```diff
@@ -17,7 +17,15 @@ const initialState = NULL_ARRAY;
 const messageSendStart = (state, action) => {
   const message = state.find(item => item.timestamp === action.outbox.timestamp);
   if (message) {
-    return state;
+    // If we find there is already a message in the outbox with the same
```
chrisbobbe, line 20: Looks like there's a test in `outboxReducer-test.js` that should be changed to test the new behavior. 🙂

chrisbobbe: Also, an extra-principled option would be to add an `outboxActions-test.js`, like we have `fetchActions-test.js`, on the action side of Redux.

Mainly, `fetchActions-test.js` tests the async `thunk` action creators to make sure they dispatch the appropriate plain object actions when they're supposed to. Note that `redux-mock-store` describes in its README that it doesn't do anything with the reducers; there's not currently a clear way to test the actions and the reducers together.

Could just start with something small, so that there's at least something to add to, for the next person making changes to the outbox actions.

src/types.js
```diff
@@ -201,6 +201,9 @@ export type Outbox = {|
   subject: string,
   timestamp: number,
   type: 'stream' | 'private',
+
+  /** If true, we don't send the message to the server. */
+  deferSending?: boolean,
```
gnprice, line 206: What's the difference in meaning between a missing value and `false`?

If there is no difference in meaning, then the property should just be always present. We'll want a migration to add it to existing outbox objects, as `false`.

src/outbox/outboxReducer.js
```diff
@@ -17,7 +17,15 @@ const initialState = NULL_ARRAY;
...
+    // timestamp, replace it with the new message from the action. This is used
+    // to update & send mesasges with `deferSending` as `true`.
+    return state.map(item => {
+      if (item.timestamp === action.outbox.timestamp) {
+        return action.outbox;
+      }
+      return item;
+    });
```
gnprice, line 28: This behavior is a bit peculiar, and it only really becomes possible to understand given the context of the next commit. (One way of thinking of that: it's presenting an API which that next commit is using, and the API is quite particular and starts really making sense only when the way the API is used becomes visible.) So for that reason it'd be best to just squash these commits together.

src/message/fetchActions.js
```diff
@@ -286,8 +286,16 @@ export const uploadFile = (narrow: Narrow, uri: string, name: string) => async (
...
+  const timestamp = await dispatch(addToOutbox(narrow, '[Uploading file...]', true));
+  let response;
+  try {
+    response = await api.uploadFile(auth, uri, name);
+  } catch (err) {
+    dispatch(deleteOutboxMessage(timestamp));
+    return;
+  }
+  const messageToSend = `[${name}](${response.uri})`;
+  dispatch(addToOutbox(narrow, messageToSend, false, timestamp));
```
gnprice, line 300: It makes me kind of nervous to be constructing this whole outbox object (and all its properties) a second time here. What if it gets some different answers the second time? Which should prevail? I'd rather not have to worry about that.

(Also it means doing a variety of work twice, which is the sort of thing that can contribute to making things slow. In this case I think there's nothing expensive that gets done twice, but because it's spread over a decent bit of code that *looks* like it gets called once per message sent, that could change innocently in the future.)

Instead let's have a new action type like UPDATE_OUTBOX_MESSAGE, which just contains a `timestamp` property (really `id` would be better, but the existing code uses `timestamp` as an ID and we can follow that for now), `deferSending`, and `content` and `markdownContent`.

Then instead of this second call to `addToOutbox`, we can call a new thunk action creator, something like `dispatch(completeOutboxMessage(timestamp, messageToSend));`, and that one will dispatch one of the new action type to update Redux and then dispatch a `sendOutbox()`.

src/outbox/outboxActions.js
```diff
@@ -176,7 +181,9 @@ export const addToOutbox = (narrow: Narrow, content: string) => async (
       avatar_url: userDetail.avatar_url,
       isOutbox: true,
       reactions: [],
+      deferSending,
     }),
   );
   dispatch(sendOutbox());
```
gnprice, line 187: This line can be skipped when `deferSending`, and that might make it a bit easier to think about what's intended to happen here.

src/outbox/outboxReducer.js
chrisbobbe: > Looks like there's a test in `outboxReducer-test.js` that should be changed to test the new behavior. 🙂

This is no longer true, with the implementation Greg suggests at https://github.com/zulip/zulip-mobile/pull/4129#discussion_r463413900.

chrisbobbe: > Also, an extra-principled option would be to add an `outboxActions-test.js`, like we have `fetchActions-test.js`, on the action side of Redux.

I took a stab at doing this, but I think it would be best to do this along with #3881, so that we don't end up doing work that we'll just discard.

chrisbobbe:
I've just pushed a new revision, resolving conflicts and addressing Greg's feedback above. 🙂

chrisbobbe:
Hi, @agrawal-d! I have an update on the status of this PR. It's looking like it'll be superseded by something like #4590, which aims to put the link in the compose box instead of sending the image as a message right away. This will be really helpful, as it'll mean the user has another opportunity to cancel if it turns out that they don't actually want to send a particular message.

Unfortunately, that PR goes in a pretty different direction, so it's not likely that most of the work done here will be useful there. Given that, I'll go ahead and close this. We still really appreciate that you did this work, though, as it helped move our thinking forward! 🙂

PR #4132, opened 2020-06-01
More pre-RN 60 commits

chrisbobbe:
These two are small and NFC but mergeable before the RN upgrade, to make that PR easier.


gnprice:
Merged -- thanks!

Issue #4093, opened 2020-05-07
Apple warning about UIWebView (2020-11 deadline)


gnprice:
Apple sends the following warning by email when I upload a new version of the app:

> **ITMS-90809: Deprecated API Usage** - App updates that use UIWebView will no longer be accepted as of December 2020. Instead, use WKWebView for improved security and reliability. Learn more (https://developer.apple.com/documentation/uikit/uiwebview).

I believe we don't actually use `UIWebView` -- we switched to `WKWebView` in 24.0.113, over a year ago. Unless we messed that up somehow. But `react-native-webview` at the version we're using still has code that refers to the old `UIWebView`, which is almost surely enough to explain the warnings.

Looks like I started getting these warnings with the upload of 26.23.146, on 2020-03-09. That may reflect a change in our app then, or it may just reflect Apple starting to issue these warnings.

It looks like react-native-webview dropped UIWebView in v7.0.1:
https://github.com/react-native-community/react-native-webview/releases/tag/v7.0.1

So we should upgrade to that version, and then see if that eliminates this warning.

That version in turn comes after the r-n-webview v6 change to use AndroidX. So after we ourselves migrate (#3852), we should be able to try that upgrade.

chrisbobbe:
I'm currently planning to upgrade `react-native-webview` to 7.0.0 in the same commit as the RN v0.60.0 upgrade, because of peer dependency constraints. Then, as a follow-up, I'm planning to take it to v7.0.3 because of a want expressed in 1982f3fa8183502b5388dec0377c518fdeedf. If you want, I can stop by v7.0.1 in a separate commit before v7.0.3, with a note about dropping UIWebView, or I can just give a shoutout to that in the v7.0.3 commit.

Issue #4026, opened 2020-04-08
Build: Use "Autolinking" (RN >= v0.60)


chrisbobbe:
Once we're on RN >= v0.60 (that's #3548), we can further increase the ease of managing our native code dependencies, even more than we did on iOS with CocoaPods in #3987.

EDIT: Ah, I just realized that there'd been some previous discussion of this! That's [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/RN.20autolinking/near/813280).

The autolinking docs are [here](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md).

The key step that we will no longer have to do is `react-native link`. Note that this distinguishes "autolinking" from "automatic linking" (what a choice of names!), described in the ["Linking Libraries" doc](https://reactnative.dev/docs/linking-libraries-ios#automatic-linking). Note that the "Manual Linking" step, there, describing steps you can take through Xcode to modify the `ios/project.pbxproj` file, is completely obsolete following the use of CocoaPods in #3987.

I expect we'll want to improve this issue heading (I'm open to this message being edited, wiki-style), but for now, I wanted a place to put a specific comment about one of our dependencies, which I'll do in the next comment.

chrisbobbe:
At https://github.com/zulip/zulip-mobile/pull/3987#discussion_r402002988, Greg was puzzled about the presence of a `libz.tbd` in `project.pbxproj` (our project's Xcode config): 

> This one isn't in the upstream template -- but it also looks like it never was. It entered our project file in [a18c95b](https://github.com/zulip/zulip-mobile/commit/a18c95b5976867117a04d23734f5d9a7e32194c8) (#881).
> 
> Well, I guess that makes any cleanup of it orthogonal to this PR, then.

As that commit and PR hint at, but don't really say, `libz` was introduced when we first installed [`@sentry/react-native`](https://github.com/getsentry/sentry-react-native). It still receives some special treatment in the latest version: it appears under `platforms.ios.sharedLibraries` in `@sentry/react-native`'s `react-native.config.js`, a file that we [know](https://github.com/zulip/zulip-mobile/pull/3987#issuecomment-605359371) is capable of causing changes in the `project.pbxproj` file when you run `react-native link`. The doc for that bit of a `react-native.config.js` file seems to be [here](https://github.com/react-native-community/cli/blob/master/docs/dependencies.md#platformsiossharedlibraries).

[...]

PR #4133, opened 2020-06-01
RN 60 upgrade

### LINKED ISSUE #3548: Upgrade to RN v0.60

chrisbobbe:
Closes #3548, #4026. Yay!

In this revision, there are two `prettier [nfc]` commits that are fine to merge first; just those two are at #4132.

I ran `rm -rf node_modules; rm -rf ios/Pods; yarn; tools/test --full` at every commit, and they all passed. I opened the app on my iOS device and on the Android emulator (the Android devices has been having "Could not connect to development server" errors), poked around, sent a message, etc.; it all seems to work.

There's just a bit of follow-up work to do, but I'll put that in its own PR so I don't crowd this one.


src/webview/js/js.js
```diff
@@ -604,6 +604,10 @@ const eventUpdateHandlers = {
 };
 
 // See `handleInitialLoad` for how this gets subscribed to events.
+
+// Flow 0.98.0 (in an upcoming commit) wants this to be
```
gnprice: nit: I'd use a line like `//` to keep the above comment visually attached to the code it's commenting on, while still getting this blank line between paragraphs.

Doesn't much matter here, because this comes and goes within the span of the series.

chrisbobbe:
Oops! I was just reminded by https://github.com/zulip/zulip-mobile/pull/4089#discussion_r433566019 and the doc linked from there that we're sticking with `Fixes: #...`, not `Closes: #...` for consistency's sake; I'll fix this tomorrow. I'd seen a few instances of `Closes: #...` recently and thought there might be a new tendency to distinguish fixing broken things from doing tasks, generally. But I agree, consistency is nice.

gnprice:
#### d8329e609 android build: Prepare to activate Hermes.

Tiny nit, but I'd more typically say "enable Hermes" or "switch to Hermes".

    This part corresponds to
    
    - facebook/react-native@8e375850d
    - facebook/react-native@4bb0b4f20 (a partial reversion)
    
    Then:
    
    - In 0.60.1, facebook/react-native@e857d7066 was released,
      giving Hermes support.
    - In 0.60.2, facebook/react-native@0738fe573 was released,
      fixing an error in the template app.

I read this as saying this part (this commit?) corresponds to the first two of these. But I think it corresponds to all of them together? I think squashing it like this probably is indeed the cleanest way to show it. We can probably also tell the story more simply to match -- like just list the four commits and say it's the combination of them.

chrisbobbe:
> I read this as saying this part (this commit?) corresponds to the first two of these.

Ah, right, thanks for catching that! I think I meant "this part" in the context of "RN did this in two parts" above that, and "Then:" referred to the second part of RN's strategy. I agree it's ambiguous; I'll find a way to clear that up.

chrisbobbe:
Thanks for your comments so far! They're all very helpful; I'll address them tomorrow. I wasn't sure if it was possible to adequately summarize all the information but clearly you've gotten most of the way there, so I'll take it from there! 😄

chrisbobbe:
> Tiny nit, but I'd more typically say "enable Hermes" or "switch to Hermes".

Ah, haha. I think the variable they use in the template is indeed `enableHermes`.

Now I have some monstrosity in my mind about a mad scientist in some sci-fi comic evilly commanding "Activate Hermes!" and Percy Weasley's owl Hermes shoots out from somewhere and squawks.

src/webview/js/js.js
chrisbobbe: Sure, good idea!

chrisbobbe:
I remembered that we need a libdef for `react-native-webview` 7. FlowTyped doesn't have one, so I'm putting one together from the FlowTyped one for version 6, plus a diff of some TypeScript files in the `react-native-webview` repo.

chrisbobbe:
Updated this wth the `react-native-webview` 7 libdef.

gnprice:
Very helpful, thanks!

#### 2db86bb1c Upgrade React Native from v0.59 to v0.60.

    - Upgrade `react-native-webview` to satisfy peer dependencies and
      remove the outdated libdef (to be replaced in an upcoming commit).

* The removal doesn't happen in this commit, so it's confusing to have it in this list of things that happen.

  It was already explained pretty clearly in the commit where it happened, so I think it can just not be mentioned here.

    - Upgrade `react-native-sound` and `rn-fetch-blob` to versions with
      podspecs compatible with the new pod layout in RN v0.60. A lot of
      work has already been done toward this:
    
    - We ignore or have already handled several changes to the Xcode and
      Gradle configs.

* The "A lot of work has …" looks like an editing error.

* Oh, and I guess the "We ignore or have already handled …" is maybe better as one of the list after this (of things previously done), too.

* This first list could use a short heading saying what it's a list of -- like "In this commit:"

* Then, lower down: the discussion of the two declared breaking changes in r-n-webview is quite informative! It kind of gets drowned out a bit by all the details of the peer-dep warnings. How about replacing the first few paragraphs of that section, from "After the upgrade, we get" through "We take 7.6.0, the latest.", with something like this?

  > We upgrade react-native-webview in this commit because versions before 7.x aren't compatible with RN v0.60, and versions 7.x aren't compatible with RN v0.59. We take the latest 7.x version, 7.6.0.

* Hmm, also, why not go all the way to the latest?

* One thing that could simplify this a bit: upgrade first to the latest 6.x. That would cover the AndroidX update and also the PR we sent.

* Another thing that would simplify it, IIUC, is actually to go in this commit only to 7.0.0 -- then to the latest (or latest 7.x) in a followup commit. That way the one thing that actually requires a code change explained in this giant commit message gets to be pulled out.

    later than 7.0.0. We take 7.6.0, the latest. Nicely, this gets us
    the changes from one of our PRs, released in 7.0.3; see 1982f3fa8
    and its reversion in the commit that followed, bbfac73f0.

* (This is kind of sad to look at because what that PR did was remove some noise from the Gradle output... and unimodules is now about 20 times noisier :cry:, plus a bunch of color.)

    There are no updates on the Android side that must happen atomically
    with the RN upgrade.
    
    Some Android changes never appear in this series:

* I think this whole section can just be dropped; it looks like everything in it is something already explained in another commit.

  In general this commit message is quite long, and has a lot of different things explained in it. That can make it a bit of a challenge to follow the explanations of the key points about the changes it does make. So cutting material about changes it doesn't make is helpful for that.

    Some iOS changes from the upgrade guide don't appear in this series.
    …

* Similarly, most of this is summarizing stuff already done and explained in previous commits.

  Hmm, I guess one thing that's going on in this list, and also the Android one, is that it provides a sort of cross-reference from the diff hunks in the upgrade-helper output. That's definitely a useful thing to do! It was basically an essential part of the work you did to produce the upgrade, and it's good to keep those resulting notes.

  I think this commit message isn't the right medium for it, though, because most of it isn't about this commit and it's quite a lot of material.

  A good medium for it might be the GitHub issue, #3548, as a comment. In fact one major advantage that would have is that, as is, this cross-reference is a bit scattered through different parts of the commit message (here; the Android list above; the short paragraph about "scripts" in `package.json`); if it's its own GitHub comment, you can collect all those together.

  Also as a comment you can edit it as needed; at this point there won't be many edits left, but perhaps a few to link to follow-up commits.

  In retrospect and for future upgrades, that could be a good way to track those notes while the work is in progress, too.

android/app/src/main/java/com/zulipmobile/MainApplication.java
```diff
@@ -43,25 +36,21 @@ public boolean getUseDeveloperSupport() {
...
+            // manually here
+            packages.add(new ZulipNativePackage());
+            packages.add(new NotificationsPackage());
+            packages.add(new SharingPackage());
+
+            // Add unimodules
+            List<ReactPackage> unimodules = Arrays.<ReactPackage>asList(
                     new ModuleRegistryAdapter(mModuleRegistryProvider)
             );
+            packages.addAll(unimodules);
```
gnprice, line 51: Is this different from just `packages.add(new ModuleRegistryAdapter(…))`?

android/app/build.gradle
```diff
@@ -239,3 +267,6 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
         jvmTarget = "1.8"
     }
 }
+
+// Necessary for Autolinking; see facebook/react-native@261197d85
+apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
```
gnprice, line 272: nit: newline at end of line

Also, let's uncramp this line by putting the statements on separate lines. I can see why people jam them into one line when writing instructions and saying to copy-paste; but now that it's part of our source code, it's better to put the code in a reasonable format.

android/settings.gradle
```diff
@@ -1,33 +1,7 @@
 rootProject.name = 'ZulipMobile'
-
+apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
```
gnprice, line 2: similarly, let's uncramp this onto two lines

Also the blank lines seem helpful, separating logical stanzas of related code.

gnprice:
#### 747145fbf libdefs: Compose and use libdef for `react-native-webview` at 7.6.

This looks like a lot of work -- thanks!

Would be great to send this as a PR to the flow-typed repo, too.

I tried spot-checking some of the changes, as seen in `git diff @~9 -- flow-typed/`. They seemed fine, but I didn't attempt to read through all of them.

There was a converter from TS to Flow libdefs that we'd talked about before... ah, Flowgen:
https://github.com/joarwilk/flowgen
and you've [tried it once before](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Android.20build.3A.20unimodules/near/845833). Did you try it out here? How'd that go?


#### 0b03a5083 android build: Enable "autolinking".

    2. Make the changes indicated by `facebook/react-native@261197d85`,
       where it touches the Android side of the React Native template
       app. In this example, `new MainReactPackage()` is nowhere to be
       seen [3], so, remove that from the list. …

    [3]: It isn't present in
         https://github.com/facebook/react-native/blob/769e35ba5f4c31ef913035a5cc8bc0e88546ca55/template/android/app/src/main/java/com/helloworld/MainApplication.java#L22-L28,
         which the "autolinking" doc (footnote 1, above) links to.

It sounds like this is just describing the fact that `new MainReactPackage()` is removed by that commit. Is there a subtlety here that I'm missing? If not, then it seems like the first sentence covers that, and the second sentence and its footnote aren't needed.

Or, for this whole step:

> 2. Apply the Android-side changes from facebook/react-native@261197d85. In `getPackages`, this removes `MainReactPackage` and gives a new form to put the references to `ZulipNativePackage`, `NotificationsPackage`, and `SharingPackage`.

    4. Add a `react-native.config.js` file, as the recommended way [2]
       to say we don't want certain modules to be linked.

Great!

How did you determine that's the right set of config tweaks to make? In particular that there wasn't some other library that's in our node_modules but we don't want to autolink.


#### a1d3609a6 ios build: Enable "autolinking".

    2. Make the changes recommended by these commits in
       `facebook/react-native`. They follow an A, revert-A, A' pattern:
    
       - 261197d85 Implement changes to enable native modules auto
       linking (#24506)
    
       - da7d3dfc7 Partially back out #24506, fixing iOS e2e tests
       (#24788)
    
       - 86a97e783 – Bring back autolinking to the iOS template (#25314)
    
       It simply imports and runs a command, `use_native_modules`.

I think the method's name is `use_native_modules!` I.e. the `!` is just the last character of its name.

In particular if you look in the file that implements it, the definition looks like:
```ruby
def use_native_modules!(root = "..", config = nil)
  …
```

Also it looks like this can be summarized by mentioning just the last commit -- the changes made in A that weren't reverted in A' are all outside `template/ios/`.


    6. Run `pod install` and see that the only change to
       `ios/Podfile.lock` is in the checksum. So, job done.

:tada:

chrisbobbe:
Thanks for the detailed review! I'll respond in chat, starting [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3548.20RN.20v0.2E60.2E0.20upgrade/near/893725).

android/app/src/main/java/com/zulipmobile/MainApplication.java
chrisbobbe: Hmm, interesting! I don't know what else might be supposed to go in this list, eventually. I was following this: https://gist.github.com/mczernek/9de9e184abc430e9e3508d26738c8a14/revisions

which is linked to from https://github.com/unimodules/react-native-unimodules, as "this diff for react-native >= 0.60".

I think it makes sense to just do `packages.add(new ModuleRegistryAdapter(…))`.

gnprice:
I think you're still working on the r-n-webview libdefs.

On the rest, just small comments:

    5. To compare the list of linked packages before and after this
       commit, note that in both cases, the list to be observed is the
       one returned by the `getPackages` implementation given to the
       abstract class `ReactNativeHost`, where it's constructed in
       `MainApplication.java`.

This isn't entirely right -- the abstract class `ReactNativeHost` has the implementation it has, and we don't give it a new one. The name for the syntax you're seeing in `MainApplication.java` looks to be "anonymous class": https://docs.oracle.com/javase/tutorial/java/javaOO/anonymousclasses.html That is, we make a subclass of `ReactNativeHost`, don't bother to give the class a name, and immediately just make an instance of it (as suggested by the `new`.)

I think just saying "the one returned by the `getPackages implementation in `MainApplication.java`" is probably enough.


    2. Make the change recommended by `facebook/react-native@86a97e783`:
       import and call a method, `use_native_modules!`.

Why the backticks around the commit ID? Without them, the GitHub view of the commit message will linkify it, which seems useful.

Otherwise all looks good!

chrisbobbe:
> This isn't entirely right -- the abstract class `ReactNativeHost` has the implementation it has, and we don't give it a new one. The name for the syntax you're seeing in `MainApplication.java` looks to be "anonymous class": https://docs.oracle.com/javase/tutorial/java/javaOO/anonymousclasses.html That is, we make a subclass of `ReactNativeHost`, don't bother to give the class a name, and immediately just make an instance of it (as suggested by the `new`.)

Ah OK, good to know, thanks! I wasn't aware of anonymous classes.

chrisbobbe:
> Why the backticks around the commit ID?

Habit, I think — using backticks means something is definitely not an English word. But...this is already definitely not an English word! And those links are helpful. It'll autolink (ha) to this issue from the facebook/react-native repo, but that's fine.

chrisbobbe:
Thanks for the review! I've pushed the changes with the `react-native-webview` libdef.

gnprice:
Great! This all looks good -- just a couple of small nits.

The PR is marked as draft; is there a reason that's intentional or did you mean to lift that?

#### 320ab3a5a Upgrade React Native from v0.59 to v0.60.

    Fixes: #3548
    
    Fixes: #4093

nit: The "Foo: something" metadata lines at the end go together as one stanza; see for example the commit messages in linux.git. For a specific example, here's the latest non-merge commit at the moment in Linus's tree: https://github.com/torvalds/linux/commit/db33ec371be8e45956e8cebb5b0fe641f008430b


#### 37cbe114f libdefs: Compose and use libdef for `react-native-webview` at 7.x.x.

The `style` prop is probably optional, right?

chrisbobbe:
> The `style` prop is probably optional, right?

Mmm, thanks for catching this! Also, I don't remember why I put the `+` in there; I'm not finding a really suitable example that I would have modeled that on, either in `expo-apple-authentication` ( a few enum-like types use `+` there, but that's all) or in `ViewProps` in `react-native/Libraries/Components/View/ViewPropTypes`, which is what informs this type.

In fact, it looks like the actual line (line 378 for me) is this:

```
style?: ?ViewStyleProp,
```

Not really something we've made a pattern of doing in our code, but maybe I'll use this just for consistency.

chrisbobbe:
and I just un-marked it as a draft! Yeah, I should have done that earlier.

gnprice:
OK -- merged! :confetti_ball: This was a big upgrade -- glad it's now complete. :smile:

Issue #4141, opened 2020-06-08
Move "edit message" state to be local UI state


gnprice:
The "edit message" feature has some significant UI problems, like #2757 and #3413, and missing features like #3975.

A root cause of some of these, and an obstacle to improving the others, is that the current implementation is a pretty bad hack that isn't compatible with the design of React. This issue is to rebuild it with a better architecture. After we've done this, it'll be straightforward to address those other issues.

Right now the edit-message feature lives in a piece of global Redux state, with code in several different places arbitrarily going and looking at it and then changing their behavior. Plus some dubious logic in `componentWillReceiveProps` in `ComposeBox`.

The proper React-style way to handle it would instead be, I think, something like:
* The state of "we're editing a message now, namely message 12345" is UI state. It's UI state that is local to the message-list screen aka `ChatScreen`. So it should be part of the React state on that `ChatScreen` component.
  * (In particular that's the least common ancestor of the components that care about that state -- ComposeBox cares, and MessageList cares because it's responsible for the message action sheet which starts an edit. Also in our current UI the ChatNavBar component cares.)

* Then that information gets passed down as props to the child components that need to know about it.
  * So `ComposeBox` gets a prop `editMessage`.
  * So does `ChatNavBar`. Then it passes the same prop down to `Title` (which is basically a private helper for `ChatNavBar` already, despite its generic-sounding name), and `Title` doesn't need to get it from Redux.
  * And `MessageList` gets a prop for a callback to start editing a message. This can get passed down to the message action sheet and used instead of dispatching a `START_EDIT_MESSAGE` Redux action.

It's possible the state should even live not on `ChatScreen` but on `ComposeBox`. The code in its `componentWillReceiveProps` looks like what it really wants to be is a method that `ComposeBox` exposes for changing its edit-message state. This would be a lot like treating `ComposeBox` as an ["uncontrolled component"](https://reactjs.org/docs/uncontrolled-components.html).

But in any case I think it will be easier to switch between different flavors of that React-style design than to switch to any one of them from the current design where `editMessage` is in the global Redux state. So I'd like us to do that, and then we'll be in a much better position to make further changes to the edit-message UX.

PR #4143, opened 2020-06-09
chat screen: Move "edit message" state to be local UI state.

agrawal-d:
Currently, the message edit state lives in the global redux state.
This is not the react way of handling that and leads to a lot of
complications.

This commit moves the edit message state to `ChatScreen`, and it's
passed down as props to the required child components.

Fixes: #4141


src/webview/MessageList.js
```diff
@@ -278,9 +280,9 @@ class MessageList extends Component<Props> {
             // Android. On iOS, though, it involves some unpredictable UUIDs which
             // RN provides no good interface to. (`react-native-fs` is awful.)
             if (
-              url.startsWith('file://')
-              && url.endsWith(baseUrlTail)
-              && !unsafeUrlRegex.test(url)
+              url.startsWith('file://') &&
```
agrawal-d, line 283: As @chrisbobbe suggested [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SOLVED.3A.20Prettier.20auto-formatting/near/893193), I downgraded Prettier VS Code extension to v4.7, but this formatting change still happens. :confused:

gnprice: That would be because of this line :slightly_smiling_face: :
```diff
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -1,3 +1,4 @@
+/* eslint-disable operator-linebreak */
 /* @flow strict-local */
```

Effectively that line requests this formatting change, by disabling the way we control that aspect of formatting.

src/compose/ComposeBox.js
```diff
@@ -265,13 +266,21 @@ class ComposeBox extends PureComponent<Props, State> {
...
     }
   };
 
+  cancelEdit = () => {
+    const { onEditComplete } = this.props;
+
+    if (this.messageInput) {
+      this.messageInput.blur();
+    }
+    onEditComplete();
```
gnprice, line 281: These come in the opposite order from above in `handleEdit`.

Probably both are fine :slightly_smiling_face: but I'd rather not have to think about both -- better to pick one.

src/compose/ComposeBox.js
```diff
@@ -361,6 +386,14 @@ class ComposeBox extends PureComponent<Props, State> {
...
     return (
       <View style={this.styles.wrapper}>
+        {editMessage && (
+          <View style={this.styles.editingBanner}>
+            <Label text="Editing message" style={this.styles.editStatus} />
+            <Touchable onPress={this.cancelEdit}>
+              <Label text="Cancel" style={this.styles.editCancel} />
+            </Touchable>
+          </View>
+        )}
```
gnprice, line 396: I like this direction! Much better than changing the app bar at the top.

I think there are some tweaks I'd like to make to the UI. How about we do this as a separate followup PR? We'll want screenshots, and it'll just be easier to have that thread if it's not mixed into the same thread with code comments on the architectural changes.

This would also be great to start a thread on in #mobile with screenshots, to get feedback there.

src/chat/ChatScreen.js
```diff
@@ -32,16 +35,41 @@ type Props = $ReadOnly<{|
...
+    editMessage: null,
+  };
+
   static contextTypes = {
     styles: () => null,
   };
 
+  onEdit = async (message: Message) => {
+    const { auth } = this.props;
+    const { raw_content } = await getRawMessageContent(auth, message.id);
```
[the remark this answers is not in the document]
gnprice: Done: 33ebcb22d and https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#zulip-api-bindings .

src/compose/ComposeBox.js
```diff
@@ -266,7 +265,11 @@ class ComposeBox extends PureComponent<Props, State> {
         showErrorAlert(error.message, 'Failed to edit message');
       });
     }
-    dispatch(cancelEditMessage());
+
+    onEditComplete();
+    if (this.messageInput) {
+      this.messageInput.blur();
+    }
```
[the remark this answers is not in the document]
agrawal-d: UI improvement was the reason. Agreed, it should be in a separate commit.

src/compose/ComposeBox.js
agrawal-d: I was excited about this change, so I included it in this PR. But yeah, a followup PR will work.

src/chat/ChatScreen.js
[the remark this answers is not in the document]
agrawal-d: Moved the logic to action sheet.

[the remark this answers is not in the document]
agrawal-d: Added the `api` namespace.

src/message/messageActionSheet.js
```diff
@@ -311,11 +315,12 @@ export const constructNonHeaderActionButtons = ({
...
   message,
   narrow,
+  onEdit,
 }: ConstructSheetParams<>): ButtonCode[] => {
   if (message.isOutbox) {
-    return constructOutboxActionButtons({ backgroundData, message, narrow });
+    return constructOutboxActionButtons({ backgroundData, message, narrow, onEdit });
   } else {
-    return constructMessageActionButtons({ backgroundData, message, narrow });
+    return constructMessageActionButtons({ backgroundData, message, narrow, onEdit });
```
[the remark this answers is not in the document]
agrawal-d: Added a commit.

src/chat/ChatScreen.js
```diff
@@ -18,11 +18,13 @@ import { canSendToNarrow } from '../utils/narrow';
 import { getLoading } from '../directSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow } from './narrowsSelectors';
+import { getAuth } from '../selectors';
 
 type SelectorProps = {|
   fetching: Fetching,
   haveNoMessages: boolean,
   loading: boolean,
+  auth: Auth,
```
gnprice, line 27: I think this isn't used, now that the API call is made at a different layer.

src/chat/ChatScreen.js
```diff
@@ -32,16 +34,35 @@ type Props = $ReadOnly<{|
...
 
+  state = {
+    editMessage: null,
+  };
+
   static contextTypes = {
     styles: () => null,
   };
 
+  onEditStart = async (editMessage: EditMessage) => {
```
gnprice, line 52: no longer needs async

src/chat/ChatScreen.js
```diff
@@ -32,16 +34,35 @@ type Props = $ReadOnly<{|
...
+    this.setState({
+      editMessage,
+    });
```
gnprice, line 55: nit: so short, best on one line (much like the `editMessage: null` below)

src/chat/ChatScreen.js
```diff
@@ -32,16 +34,35 @@ type Props = $ReadOnly<{|
...
+
+  onEditComplete = () => {
+    this.setState({ editMessage: null });
+  };
+
   render() {
     const { styles: contextStyles } = this.context;
     const { fetching, haveNoMessages, loading, navigation } = this.props;
+    const { editMessage } = this.state;
     const { narrow } = navigation.state.params;
```
gnprice, line 66: nit: the line getting stuff from out of `navigation` is directly connected to the line that got `navigation` from out of `this.props`, so natural to keep them right next to each other

src/message/messageActionSheet.js
```diff
@@ -70,8 +72,15 @@ const copyToClipboard = async ({ _, auth, message }) => {
 copyToClipboard.title = 'Copy to clipboard';
 copyToClipboard.errorMessage = 'Failed to copy message to clipboard';
 
-const editMessage = async ({ message, dispatch }) => {
-  dispatch(startEditMessage(message.id, message.subject));
+const editMessage = async ({ message, dispatch, startEdit, auth }) => {
+  if (!message.isOutbox) {
```
gnprice, line 76: The case where `message.isOutbox` is true should be impossible -- it would represent a bug. (We only show this option for real, sent messages.) So the case where it's false should be treated as the main line of the logic, rather than put inside a condition.

A way to do that with the same semantics as this version would be an early return:
```js
  if (message.isOutbox) {
    return;
  }

  // …
```

But also as long as we have logic to detect this case -- which is here basically because we haven't expressed its impossibility in a way that the type-checker is able to prove for us -- we should have something to let us know if it starts impossibly happening:
```js
  if (message.isOutbox) {
    logging.warn('Attempted "Edit message" for outbox message');
    return;
  }

  // …
```

Another alternative would be to throw an exception. In this case just a warning is fine, because there's a pretty clear thing we can do to carry on -- just ignore the attempted action.

src/webview/webViewEventHandlers.js
```diff
@@ -187,12 +188,21 @@ const handleLongPress = (
...
-    message,
-    narrow,
-  });
+  const { dispatch, showActionSheetWithOptions, backgroundData, narrow, onEditStart } = props;
+  showActionSheet(
+    target === 'header',
+    showActionSheetWithOptions,
+    {
+      dispatch,
+      startEdit: onEditStart,
```
gnprice, line 197: Let's pick a single name for this callback :slightly_smiling_face: Makes it easier to trace things through the various layers.

I'm happy leaving out the `on`. I think the name can be improved further by adding the word "message", so it's clear this is about editing a message -- vs. e.g. editing a draft. Perhaps `startEditMessage`, then.

agrawal-d:
@gnprice Do we not have CI running on GitHub anyomre?

src/chat/ChatScreen.js
agrawal-d: Removed

src/webview/webViewEventHandlers.js
agrawal-d: Used `startEditMessage` everywhere.
Also changed `onEditComplete` to `completeEditMessage` for consistency.

gnprice:
Looks great, thanks!

Merging, with one formatting tweak:
```diff
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -192,16 +192,8 @@ const handleLongPress = (
   showActionSheet(
     target === 'header',
     showActionSheetWithOptions,
-    {
-      dispatch,
-      startEditMessage,
-      _,
-    },
-    {
-      backgroundData,
-      message,
-      narrow,
-    },
+    { dispatch, startEditMessage, _ },
+    { backgroundData, message, narrow },
   );
 };
```

Much more compact this way. :slightly_smiling_face:

gnprice:
> @gnprice Do we not have CI running on GitHub anymore?

We do, but it seems CircleCI made a surprise change that caused it to stop working in a common configuration:
https://chat.zulip.org/#narrow/stream/43-automated-testing/topic/circleci.20backlog/near/899195

The workaround is to stop having the jobs run on your personal CircleCI account, and instead leave them to run in the organization's CircleCI account. But the consequence of that is that they'll tend to pile up in a backlog, because they're all running against the same small quota (and CircleCI is pretty slow to begin with, even when you're paying them a lot of money.) So that's not a great solution, and it's likely that this means it's a good moment for us to try switching to GitHub Actions, which didn't exist yet when we adopted CircleCI.

In the meantime, I'll just try to be aware that I need to be sure to run `tools/test` before merging :slightly_smiling_face: -- I usually do anyway, because I've usually rebased, if nothing else.

gnprice:
> We do, but it seems CircleCI made a surprise change that caused it to stop working in a common configuration:

I should make a correction here, because I evidently wasn't thinking clearly when I wrote this: we don't use CircleCI in the first place for zulip-mobile, and never have. (We use it for most of our zulip/zulip tests, which is why it came up in that thread.) Rather, we use Travis CI.

And Travis continues to run on PRs -- here's the most recent PR builds in this repo:
https://travis-ci.org/github/zulip/zulip-mobile/pull_requests
and they include a half-dozen of them from today.

The thing that's missing is that for some reason the results aren't getting posted to the PRs. I guess I should look into that. I'll make a chat thread.

gnprice:
> The thing that's missing is that for some reason the results aren't getting posted to the PRs.

And fixed!

PR #4145, opened 2020-06-09
Add `react-native-url-polyfill`.

chrisbobbe:
Fixes: #4081 

I'll follow this up by creating an issue to fix the many uses of regexes and string concatenation; I think #4081 was mainly for settling on a workable solution.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -53,6 +53,7 @@ node_modules/react-native/flow/
 flow-typed/@sentry/react-native_v1.x.x.js
 flow-typed/@react-native-community/async-storage_v1.x.x.js
 flow-typed/react-native-webview_v7.x.x.js
+flow-typed/react-native-url-polyfill_vx.x.x.js
 
 
 [strict]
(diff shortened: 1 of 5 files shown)


gnprice:
Thanks!

What's the origin of the libdef -- is that from... ah, I see, there are links in comments.

Those links are to libdefs built into Flow, for the types `URLSearchParams` and `URL`. Hmm, it sure would be nice to just refer to those, if that's the effect we want, rather than copy them. But then I'm not sure how to actually spell that, because the names are the same.

Nit in commit message:

    Many thanks to charpeni for the changes [2] that allow RN v0.60

I think that's meant to be `[3]`.

Otherwise LGTM. I'll fix that one nit and merge.

chrisbobbe:
> What's the origin of the libdef -- is that from... ah, I see, there are links in comments.
> 
> Those links are to libdefs built into Flow, for the types URLSearchParams and URL. Hmm, it sure would be nice to just refer to those, if that's the effect we want, rather than copy them. But then I'm not sure how to actually spell that, because the names are the same.

Yeah. The first thing I tried was to use those built-ins, but it didn't work.

Issue #4118, opened 2020-05-19
Replace unmaintained dep `react-native-orientation`.


chrisbobbe:
When we upgrade to RN v0.60.0 (that's #3548), we'll get this warning:

```
warn The following packages use deprecated "rnpm" config that will stop working from next release:
  - react-native-orientation: https://github.com/yamill/react-native-orientation#readme
  - rn-fetch-blob: https://npmjs.com/package/rn-fetch-blob
Please notify their maintainers about it. You can find more details at https://github.com/react-native-community/cli/blob/master/docs/configuration.md#migration-guide.
```

`rn-fetch-blob` is easy to fix; we just have to upgrade to v0.11; I'm planning to do that in my PR for #3548.

`react-native-orientation` is unmaintained, though. The fix for this warning is very simple, but the [PR for it](https://github.com/yamill/react-native-orientation/pull/353) has gathered dust, and the repo isn't getting updates.

This will take some assembly in native code, which should all be spelled out in the replacement library's installation instructions.

I recommend trying Expo's [ScreenOrientation](https://docs.expo.io/versions/latest/sdk/screen-orientation/); Expo libraries are generally frequently maintained, and we can newly use them since #4016 landed. If we're going with that, we'll follow the instructions for a "bare React Native app" (ours isn't "managed" by Expo), and use `yarn add expo-screen-orientation` instead of `npm install`. Since Expo uses TypeScript instead of Flow, we'll have to put together a libdef, unless FlowTyped has one (I don't think it does). We have a new [doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/libdefs.md) for that, and it would be nice for that to be updated with anything new we learn along the way.

chrisbobbe:
Huh, it would seem that we may actually want [DeviceMotion](https://docs.expo.io/versions/latest/sdk/devicemotion/), not ScreenOrientation...

From the [ScreenOrientation doc](https://docs.expo.io/versions/latest/sdk/screen-orientation/):
> Screen Orientation is defined as the orientation in which graphics are painted on the device. For example, the figure below has a device in a vertical and horizontal physical orientation, but a portrait screen orientation. For physical device orientation, see the orientation section of Device Motion.

chrisbobbe:
> Huh, it would seem that we may actually want DeviceMotion, not ScreenOrientation...

[...]

Issue #4131, opened 2020-05-30
Use Hermes


chrisbobbe:
This will be possible following the RN 60 upgrade, https://github.com/zulip/zulip-mobile/issues/3548.

The guide we'll want to follow is at https://reactnative.dev/docs/0.60/hermes, including all the steps to ensure it's activated properly.

gnprice:
I think this may also block on Hermes growing support for `Intl`:
https://github.com/facebook/hermes/issues/23
and in particular growing enough support that `react-intl` works. The good news is that they seem to be working on that now.

timabbott:
Interesting.  We probably want to at least prototype running Hermes even if it breaks i18n, just to get a sense of what the performance impact and discover any other things we need to fix on our end.

agrawal-d:
I'm very excited about this!

felippepuhle:
You can import some polyfills for android, ex:

```
if (Platform.OS === 'android') {	
  require('intl')	
  require('intl/locale-data/jsonp/en')	

  require('date-time-format-timezone/build/src/code/polyfill.js').default(	
    global	
  )	
  require('date-time-format-timezone/build/src/code/data-loader.js').default(	
    global	
  )	
  require('date-time-format-timezone/build/src/data/metazone.js')(global)	
  require('date-time-format-timezone/build/src/data/locales/locale-en.js')(	
    global	
  )	
  require('date-time-format-timezone/build/src/data/timezones/tzdata-america-los_angeles.js')(	
    global	
  )	
}
```

We've used to do something like this in the past but we're currently using react-native-v8.

gnprice:
> but we're currently using react-native-v8.

@felippepuhle Interesting! I hadn't been really aware of react-native-v8. I just looked around on the Web and found some benchmark reports, and it seems like react-native-v8 makes different tradeoffs from Hermes.

I'd be very curious to hear more, if you'd be up for sharing, about why you switched from Hermes to react-native-v8 :slightly_smiling_face: . Either on this thread, or on #mobile on chat.zulip.org, which might be a more convenient venue for discussion.

chrisbobbe:
Hmm, I'm getting a "Too many handles allocated in GCScope" error. It resembles https://github.com/facebook/react-native/issues/25730 and https://github.com/facebook/hermes/issues/26.

chrisbobbe:
[...]

Issue #4091, opened 2020-05-07
Upgrade to a react-native-unimodules without privacy-sensitive dead code


gnprice:
See [chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/apple.20purpose.20strings/near/869917).

We've successfully worked around this for now by including the boilerplate that the r-n-unimodules authors recommended as of the older version we're on. From our workaround commit c8d0c8df5e:

> These are needed because (a) some of the code in react-native-unimodules apparently refers to these APIs, and (b) for Calendars, Contacts, and Microphone in particular, Apple refuses to allow an app to be uploaded if it has code referring to these APIs and doesn't have a purpose string. That means this [before the workaround] blocks uploading a release for iOS.

But r-n-unimodules upstream has since fixed the problem so that's not needed, and we should upgrade. Ray finds:

> The need to do this [seems to have been removed as of 0.8.0](https://github.com/unimodules/react-native-unimodules/commit/3cac0d72a6cfe6f4ba6f4e874dc6b32c70172a48)

(Now moot because we've completed the RN v0.60 upgrade:
~~There's some uncertainty about just how far we can upgrade while still on RN v0.59. The details we've learned are described by @chrisbobbe [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/apple.20purpose.20strings/near/869947).~~

~~Possibly the cleanest way to do this is to first finish the RN v0.60 upgrade, #3548 -- I believe at this point we've done all the hard parts of that.~~)

chrisbobbe:
Hmm. I'm investigating a reason why we might not be able to do this yet, even on RN v0.60: https://github.com/unimodules/react-native-unimodules/issues/97

chrisbobbe:
There may be a kind of awkward workaround; I've posted some more findings at https://github.com/unimodules/react-native-unimodules/issues/97#issuecomment-637180616.

chrisbobbe:
See discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Podfile.20dependency.20of.20a.20dependency/near/892373).

chrisbobbe:
Blocking this on the RN v0.61 upgrade (#3781); see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Podfile.20dependency.20of.20a.20dependency/near/898933). We're optimistic that this won't take as long as the last one, though! 🙂

[...]

PR #4147, opened 2020-06-09
RN v0.60 upgrade immediate-term followup

chrisbobbe:
This doesn't include as much as I was hoping to include; see details at https://github.com/zulip/zulip-mobile/issues/3548#issue-465925049.

In particular (these are the major things), it doesn't include using Hermes (#4131) or upgrading Unimodules (#4091). But Hermes, IIUC, can be done any time; the Unimodules upgrade should Just Work, but only following the RN v0.61 upgrade.

Fixes: #4118


gnprice:
Second commit message has a `[1]` but there's no footnote -- I'm curious to follow that link :slightly_smiling_face:

chrisbobbe:
Thanks for catching that! I'll add that in—along with a note about how the "default screen orientation" bit, in those instructions, is misleading (see https://github.com/zulip/zulip-mobile/issues/4118#issuecomment-641561847).

gnprice:
Mentioning the issue one of these fixes, so it gets a backlink to the PR: this fixes #4118.

chrisbobbe:
> Mentioning the issue one of these fixes, so it gets a backlink to the PR: this fixes #4118.

I was just about to do that, when I saw you posted. Thanks!

src/boot/AppEventHandlers.js
```diff
@@ -69,13 +69,30 @@ type Props = $ReadOnly<{|
...
   netInfoDisconnectCallback: (() => void) | null = null;
 
-  handleOrientationChange = (orientation: OrientationT) => {
+  handleOrientationChange = (event: ScreenOrientation.OrientationChangeEvent) => {
     const { dispatch } = this.props;
-    dispatch(appOrientation(orientation));
+
+    const {
+      orientationInfo: { orientation },
+    } = event;
```
gnprice, line 93: The multi-line formatting is kind of a sign here of the fact that this syntax takes a bit  of effort for the reader to unpack :wink: 

The usual way we'd write this is:
```js
const { orientation } = event.orientationInfo;
```

chrisbobbe:
Thanks for the comments! I've pushed a new revision.

gnprice:
Looks good, thanks! I've added a few words to the commit message with your findings from https://github.com/zulip/zulip-mobile/issues/4118#issuecomment-641621448 ; merging.

PR #4148, opened 2020-06-10
Update build-run instructions in docs

agrawal-d:
Required after the RN 0.60 upgrade.

### DIFF
--- a/docs/howto/build-run.md
+++ b/docs/howto/build-run.md
@@ -149,20 +149,6 @@ Apart from the steps mentioned below, you may find the
 
 [React Native troubleshooting docs]: https://reactnative.dev/docs/troubleshooting
 
-### `react-native run-{android/ios}` launches the app, then crashes immediately
-
-After running `react-native run-{android/ios}`, if you see an error like this
-on the device:
-```
-Unable to load script.Make sure you are either running a Metro server
-(run 'react-native start') or that your bundle 'index.android.bundle' is
-packaged correctly for release.
-```
-then this means you are not running the Metro bundler.
-
-Starting from React Native 0.60, you need to launch the Metro bundler server
-separately, using `react-native start`. Once the server starts up, run
-`react-native run-android` again, and the app should not crash.
 
 ### `yarn install` failure, at `fsevents`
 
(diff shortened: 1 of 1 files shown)


docs/howto/build-run.md
```diff
@@ -75,6 +75,8 @@ If you're starting with iOS development, be sure to [install
 CocoaPods](https://guides.cocoapods.org/using/getting-started.html),
 as the guide instructs, then rerun `yarn`.
 
+Start the Metro bundler server using `react-native start`.
+
```
gnprice, line 79: This is one of the steps of the upstream instructions:
https://reactnative.dev/docs/environment-setup
just before `react-native run-android` (or actually `npx react-native run-android`, in their version.) There are a lot of steps there, and I don't want to try to duplicate them.

It looks like that instruction is new in v0.62 -- it's not in the v0.61 docs:
https://reactnative.dev/docs/0.61/getting-started.html
So in short I think the audience that needs to be specifically told about this step is people who've *already* gone through these instructions before the v0.62 release. Like you :slightly_smiling_face: and potentially other existing contributors, but not new readers of these instructions.

docs/howto/build-run.md
```diff
@@ -149,6 +151,12 @@ Apart from the steps mentioned below, you may find the
 
 [React Native troubleshooting docs]: https://reactnative.dev/docs/troubleshooting
 
+### `react-native run-{android/ios}` launches the app, then crashes immediately
+
+Starting from React Native 0.60, you need to launch the Metro bundler server
+separately, using `react-native start`. Once the server starts up, run
+`react-native run-android` again, and the app should not crash.
+
```
gnprice, line 159: A troubleshooting entry would definitely be helpful!

One key element in troubleshooting entries is generally a copy of the error message. That's helpful because it means people can Ctrl+F search for the message they're seeing, and quickly find the entry relevant to them. So let's include a copy of that here.

docs/howto/build-run.md
agrawal-d: Makes sense.

docs/howto/build-run.md
agrawal-d: Updated with the error message.

docs/howto/build-run.md
agrawal-d: Deleted the commit.

gnprice:
Thanks @agrawal-d ! Merged, with a few tweaks as additional commits.

PR #4150, opened 2020-06-11
Pre-RN v0.61 changes

chrisbobbe:
These will help smooth the RN v0.61 upgrade. I have a few more planned (hence the `1` in the branch name!) that I'm still working on, but @gnprice, you might like to see these.


android/app/src/main/java/com/zulipmobile/MainActivity.java
```diff
@@ -22,12 +20,4 @@ protected void onCreate(Bundle savedInstanceState) {
...
         WebView.setWebContentsDebuggingEnabled(true);
     }
-
-    @Override
-    public void onConfigurationChanged(Configuration newConfig) {
-        super.onConfigurationChanged(newConfig);
-        Intent intent = new Intent("onConfigurationChanged");
-        intent.putExtra("newConfig", newConfig);
-        this.sendBroadcast(intent);
-    }
```
gnprice, line 32: Interesting, good catch!

Let's remove this from `ReceiveShareActivity.kt` too. I puzzled a bit over that when reviewing #4124 which added it... and then saw it was copied from `MainActivity` and I guess I figured it was generic RN boilerplate, and didn't study it closer. (But in retrospect this story makes more sense -- if it were generic RN boilerplate they could just put it in the base class.)

android/build.gradle
```diff
@@ -49,6 +49,7 @@ allprojects {
 
         google()
         jcenter()
+        maven { url 'https://jitpack.io' }
```
gnprice, line 52: I think we can skip this. The PR doesn't give any further information, and searching the web I don't see a lot of evidence of people using this service.

It'll be easy to add a line like this in the future as part of adopting any library that is published there.

jest/jestSetup.js
```diff
@@ -36,3 +36,10 @@ jest.mock('react-native-device-info', () => ({
   getSystemName: jest.fn().mockReturnValue('ios'),
   getSystemVersion: jest.fn().mockReturnValue('13.3.1'),
 }));
+
+jest.mock('@react-native-community/cameraroll', () => ({
+  // Incomplete, but error-free
+  saveToCameraRoll: jest.fn(),
+  save: jest.fn(),
+  getPhotos: jest.fn(),
+}));
```
gnprice, line 45: Hmm, it's interesting that we need this now and didn't before! Do you understand what changed?

src/animation/AnimatedScaleComponent.js
```diff
@@ -22,7 +22,7 @@ export default class AnimatedScaleComponent extends PureComponent<Props, State>
 
   animatedValue = new Animated.Value(this.props.visible ? 1 : 0);
 
-  componentWillReceiveProps(nextProps: Props) {
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
```
gnprice, line 25: The effect of this is basically to acknowledge that we've seen the warning, to tell it to stop warning us.

That's good to do, but before we do we should file an issue about the need to move away from these methods. It can link to that upstream announcement https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html to handle most of the specifics.

jest/jestSetup.js
chrisbobbe: Hmm, not precisely. It looks like a similar story was seen in 33c87b5f585ab7ffe70762a86012f0ca6cfd2653, where we started getting async-storage from its @react-native-community module instead of react-native:

>     * With the extraction of `async-storage` out of `react-native` proper,
>      Jest needs a separate mock implementation made available.

In that case, though, there was a handy file at `node_modules/@react-native-community/async-storage/jest/async-storage-mock.js` (with syntax that Jest recognizes) with an export that we could pass to `jest.mock` in our jestSetup.js. There isn't such a file in @react-native-community/cameraroll.

There is a `jest.setup.js` file that does some mocking work without exporting anything. But if I just add the following to `jestSetup.js`:

```
import '@react-native-community/cameraroll/jest.setup';
```

then I get this:

```
    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/chrisbobbe/dev/zulip-mobile/node_modules/@react-native-community/cameraroll/jest.setup.js:12
    import {NativeModules} from 'react-native';
           ^

    SyntaxError: Unexpected token {

      1 | import mockAsyncStorage from '@react-native-community/async-storage/jest/async-storage-mock';
      2 | 
    > 3 | import '@react-native-community/cameraroll/jest.setup';
        | ^
      4 | 
      5 | jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
      6 | 

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
      at Object.<anonymous> (jest/jestSetup.js:3:1)
```

chrisbobbe: The original error I get, using `@react-native-community/cameraroll` without adding anything to `jestSetup.js` is this:

```
   Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/chrisbobbe/dev/zulip-mobile/node_modules/@react-native-community/cameraroll/js/CameraRoll.js:11
    import {Platform} from 'react-native';
           ^

    SyntaxError: Unexpected token {

      1 | /* @flow strict-local */
      2 | import { Platform, PermissionsAndroid } from 'react-native';
    > 3 | import CameraRoll from '@react-native-community/cameraroll';
        |                                          ^
      4 | import RNFetchBlob from 'rn-fetch-blob';
      5 | 
      6 | import type { Auth } from '../api/transportTypes';

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
      at Object.<anonymous> (src/lightbox/downloadImage.js:3:42)
```

src/animation/AnimatedScaleComponent.js
chrisbobbe: Makes sense! Opened as #4154.

chrisbobbe:
Thanks for the review! Ready for another one.

gnprice:
Thanks for the revisions! Looks great -- merged.

PR #4151, opened 2020-06-11
RN v61 upgrade

chrisbobbe:
~~Awaiting https://github.com/zulip/zulip-mobile/pull/4150.~~ And, it's merged! Thanks, @gnprice !

Fixes: #3781

PR #4152, opened 2020-06-11
RN v0.61 upgrade followup.

chrisbobbe:
~~Awaiting #4151.~~ Done!

Fixes: #4091


package.json
```diff
@@ -116,6 +116,7 @@
     "jest-cli": "^24.9.0",
     "jest-environment-jsdom": "^24.9.0",
     "jest-environment-jsdom-global": "^1.2.0",
+    "jest-expo": "^37.0.0",
```
chrisbobbe, line 119: b706f4d47 jest: Add `jest-expo` preset, to be used in the next commit.

This commit is in this PR (instead of a PR that doesn't have RN v0.61) because I didn't want to bother figuring out all the appropriate versions based on RN v0.60, which we'll move away from soon.

chrisbobbe: One thing I haven't looked into further is that this particular version of jest-expo isn't compatible with Jest 26.1.0, which we're thinking of switching to soon (see [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Lolex.20timers/near/927725)).

chrisbobbe:
Cool, and #4151 has been merged! Thanks, @gnprice! I'll fix these conflicts and push this branch.

chrisbobbe:
Mentioning this here, as #4091 will be closed automatically once this PR lands:

There'll be another important step after upgrading Unimodules, which would be easy to forget about, I think—https://github.com/zulip/zulip-mobile/commit/c8d0c8df5efe808214db78c2a5daad186c9dc91f should be reverted once we're on the new version, IIUC.

I could make that part of this PR if you want, @gnprice.

gnprice:
> [c8d0c8d](https://github.com/zulip/zulip-mobile/commit/c8d0c8df5efe808214db78c2a5daad186c9dc91f) should be reverted once we're on the new version, IIUC.

SGTM!

chrisbobbe:
(huh, looking into the `tools/verify-webview-js` failure now, that's odd)

chrisbobbe:
> (huh, looking into the `tools/verify-webview-js` failure now, that's odd, and reproduces locally)

OK, just pushed a fix! Yikes, the culprit was a `yarn yarn-deduplicate && yarn` upon installing `jest-expo`, where tons of Babel-related things were updated. One more reason to stop considering `jest-expo`, I suppose—the idea to use it was mostly a whim, though I slightly like not having to make boring mocks for `expo-` modules.

gnprice:
Merged! Thanks again.

I looked into the Babel-related changes [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/changes.20in.20generatedEs3.2Ejs/near/935752), and as discussed there I merged some changes before this that isolated the Babel upgrades and simplified that commit back again.

Issue #2757, opened 2018-07-04
Cannot cancel editing a message


Shayan-To:
You cannot get out of editing mode once started editing a message (unless you really edit the message!).

![zulipmobileedit](https://user-images.githubusercontent.com/12458563/42278248-5fee34a0-7faf-11e8-9785-1f40dbde4a77.gif)

ray-kraesig:
As noted in #3413, you can't even exit the "Edit message" state by going to another narrow where it's absent. (It *is* possible to do so by killing the app.)

gnprice:
I've just filed #4141 to describe the main work that I think we'll want to do to resolve this.

PR #4153, opened 2020-06-12
compose box: Introduce new edit message UI.

agrawal-d:
This new UI is displayed as a part of the compose box and allows
cancelling the edit in progress. The navbar no longer shows the
edit status as it's shown in the compose box itself.

image

As discussed in chat, you mentioned it might be good to try rounded corners for the compose box in general, and I tried it - It didn't look good. I think it's because the regular compose box is the part of the base UI. However, the editing message state can be thought of as an action above the normal UI, and so the rounded corners look good.

Fixes: #2757


agrawal-d:
I just updated the PR, the edit-message UI no longer has rounded borders.

chrisbobbe:
Nice, thanks, @agrawal-d!

One thing that I've noticed, without these changes, is that sometimes the "Topic" input isn't pre-filled when you go to a message. This can lead to a surprising (and annoying) failure to send the message; it only alerts you about the blank topic when you click the send button.

I just tried out editing a message ([this one](https://chat.zulip.org/#narrow/stream/7-test-here/topic/Chris/near/937316)), and the same thing happened—the message body input was filled out with the message's contents, but the topic was blank.

Are you able to reproduce this, and if so, would it be an easy fix in this PR?

agrawal-d:
>Are you able to reproduce this

I'm unable to reproduce this :( 

>would it be an easy fix in this PR?

The logic for updating the topic is handled near line 285:

```js
  UNSAFE_componentWillReceiveProps(nextProps: Props) {
    if (nextProps.editMessage !== this.props.editMessage) {
      const topic =
        isStreamNarrow(nextProps.narrow) && nextProps.editMessage
          ? nextProps.editMessage.topic
          : '';
      const message = nextProps.editMessage ? nextProps.editMessage.content : '';
      this.setMessageInputValue(message);
      this.setTopicInputValue(topic);
      if (this.messageInput) {
        this.messageInput.focus();
      }
    }
  }
```

I'm not sure what can cause topic to be blank!

Issue #4111, opened 2020-05-14
Enable `react-native-screens`.


chrisbobbe:
This issue should start with a code change that's pretty small and straightforward~~, after "autolinking", which is #4026. I'll mark this issue as `blocked` on that work.~~ The bulk of the work required will be stress-testing to see if anything breaks, and perhaps collecting some performance data.

As foreshadowed in [this blog post](https://reactnavigation.org/blog/2018/11/17/react-navigation-3.0/), and discussed in [this talk](https://www.youtube.com/watch?v=Z0Jl1KCWiag) (notable timestamps are 18:43 and 27:04), there's supposed to be a way we can substantially free up resources, to make our navigation go a lot faster. More discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.3A.20r-n-screens.20.2F.20r-n-document-picker/near/877155) and following.

Assuming we're still on React Navigation v2, just follow [this doc](https://reactnavigation.org/docs/2.x/react-native-screens/), choosing the "Setup in normal react-native applications" path. Instead of following the link to the instructions for the latest version of `react-native-screens`, you'll want to find the doc corresponding to the version we're using, which is 1.0.0-alpha.23 as of this writing. To do that, make sure you've run `yarn`, and open node_modules/react-native-screens/README.md. Look at the instructions under "Usage with react-navigation (without Expo)".

Skip the `yarn add` instruction unless you have a particular `react-native-screens` version in mind that will satisfy all peer dependency constraints. We did that step in 0d2066f41.

~~Skip the `react-native link` instruction; this is taken care of by "autolinking".~~ Actually, we'll want to edit our `react-native.config.js` to not blacklist `react-native-screens` for Android and/or iOS.

Do everything else.

chrisbobbe:
Once it's all set up (you can refer to [this doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/build-run.md) for build failures, hopefully they won't happen), we'll want to make sure things haven't broken.

The main claim of this optimization is to free up resources on screens that aren't visible. There may be a snag for us, here, since we've never had to think through anything on a screen getting garbage collected if it's still in the navigation stack but not currently visible. In particular, we know it's possible to push an unbounded number of MessageLists to the history stack. So, try doing that, and then return to one of them that's kind of buried deep. See if it's totally broken and the app crashes, or if we merely lose our scroll position, etc.

[...]

PR #4161, opened 2020-06-15
[TESTABLE] Enable react-native-screens.

chrisbobbe:
Fixes: #4111 

This branch contains my current revision of https://github.com/zulip/zulip-mobile/pull/4160. Without that, it's hard to see any performance differences clearly because one has to wait so long for a link to be visible that would add a route to the stack.

I haven't done extensive testing of this yet, and a few trials suggest that the performance difference isn't huge. But nothing seems to break. I thought I'd file this so I'd have something to return to, if we want to do further testing. Or others can check this out and run some tests too. Or, I suppose, we could just merge it and release it knowing that it's theoretically a boost to performance.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -272,6 +272,8 @@ PODS:
     - React
   - RNDeviceInfo (0.21.5):
     - React
+  - RNScreens (1.0.0-alpha.23):
+    - React
   - RNSentry (1.0.9):
     - React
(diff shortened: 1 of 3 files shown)


gnprice:
I've just sent #5421, which functions as an updated version of this. (Some details are different because we've upgraded react-native-screens since this PR.) Thanks for the investigation!

Issue #4115, opened 2020-05-16
Use @react-native-community/push-notifications-ios.


chrisbobbe:
This issue is a stub; we can fill it in later, but for now, I need an issue number to include in a TODO I'm writing for #3548.

Currently, we're depending on some combination of [this](https://github.com/wix/react-native-notifications) (less popular, less maintained) and [this](https://reactnative.dev/docs/pushnotificationios.html) (deprecated).

Discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20push.20notifications/near/825050) will be relevant.

chrisbobbe:
@gnprice, when you get a chance, could you please record your thoughts on the future of iOS notification-handling that we discussed on the phone the other day?

I think we'd like any solution we land on to include something like https://github.com/zulip/zulip-mobile/pull/4163/commits/15c79843f8f8117ea86a30d859d12e50a047a3dc, which addresses a problem we'll have to solve as part of #4182.

PR #4163, opened 2020-06-17
Start using @react-native-community/push-notification-ios

chrisbobbe:
And remove our dependency on wix/react-native-notifications.

Fixes: https://github.com/zulip/zulip-mobile/issues/4115

I took this up before I saw Greg's comment [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/905165) about https://github.com/zulip/zulip/pull/15179:
> I think this issue is actually a lot simpler than that. The PR will cause the server to start sending the information, and it's doing it in a way that iOS should already respond to even without any client-side code of ours.

Ah, well; that particular need doesn't seem to require any code changes on mobile. But I've been really wanting to get this done because

(1) I've often been kind of confused by our dependency on two different libraries for different iOS push notification needs, when there's no obvious reason why everything couldn't be done by one library, and
(2) some peace and clarity comes from being on the latest versions of things. PushNotificationIOS from RN core is deprecated, and there are complaints about how well the Wix library is maintained; see the issue, and discussion linked from there.

So I went ahead and put this together.


ios/ZulipMobile/AppDelegate.m
```diff
@@ -78,30 +83,51 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserAct
...
-  [RCTPushNotificationManager didReceiveLocalNotification:notification];
+  [RNCPushNotificationIOS didReceiveLocalNotification:notification];
 }
+
+// iOS 10.0+: Called when a notification is delivered to a foreground app.
+-(void)userNotificationCenter:(UNUserNotificationCenter *)center
+      willPresentNotification:(UNNotification *)notification
+        withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler
+{
+  // Update the badge count. Do not play sound or show an alert. For
```
chrisbobbe, line 127: Here, we explicitly tell it to update the badge count if a notification is received while the app is in the foreground. I'm not sure what the previous behavior was, but this sure seems helpful with https://github.com/zulip/zulip/pull/15179 in mind.

chrisbobbe: I've just tested on 26.28.151 (from TestFlight) on CZO. Indeed, I can send myself messages from a test account (increasing the unread count) and go back to the home screen and see that the badge count did not change. Then I send myself another message, while the app is in the background, and the notification appears and the badge count updates. (It jumps to 609...but that's a server-side issue [being investigated](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/907169)).

Being foregrounded should not suppress badge count updates; this PR should fix this. 🙂

src/notification/index.js
```diff
@@ -118,8 +119,8 @@ const getInitialNotification = async (): Promise<Notification | null> => {
   }
 
   // This is actually typed as ?Object (and so effectively `any`); but if
-  // present, it must be a JSONable dictionary. (See PushNotificationIOS.js and
-  // RCTPushNotificationManager.m in Libraries/PushNotificationIOS.)
+  // present, it must be a JSONable dictionary. (See js/index.js and
+  // ios/RNCPushNotificationIOS.m in @rnc-push-notification-ios.)
```
gnprice, line 123: this module name wants a slash, I think

src/notification/index.js
```diff
@@ -265,29 +259,32 @@ export class NotificationListener {
...
+ * Our 'register' listener will fire on that event.
+ */
 export const getNotificationToken = () => {
   if (Platform.OS === 'ios') {
-    // This leads to a call (in wix's, or RN upstream's, NotificationsIOS) to this:
+    // This leads to a call in RNC's PushNotificationIOS to this:
     //   https://developer.apple.com/documentation/uikit/uiapplication/1622932-registerusernotificationsettings
     // (deprecated after iOS 10, yikes!); which after possibly prompting the
-    // user causes "the app" (i.e. the platform part) to call this, I think,
-    // though I haven't successfully traced all the steps there:
```
gnprice, line 275: Does this mean you have traced through those steps? :slightly_smiling_face:

src/notification/index.js
```diff
@@ -265,29 +259,32 @@ export class NotificationListener {
...
+ * Try to cause a `remoteNotificationsRegistered` event.
+ *
+ * Our 'register' listener will fire on that event.
+ */
 export const getNotificationToken = () => {
   if (Platform.OS === 'ios') {
-    // This leads to a call (in wix's, or RN upstream's, NotificationsIOS) to this:
+    // This leads to a call in RNC's PushNotificationIOS to this:
     //   https://developer.apple.com/documentation/uikit/uiapplication/1622932-registerusernotificationsettings
     // (deprecated after iOS 10, yikes!); which after possibly prompting the
```
gnprice, line 271: :-/ Is this still true in the current library? Dang.

chrisbobbe: Yeah. 😦

src/notification/index.js
chrisbobbe: I did, but I don't remember offhand what they all were, and there's a chance that I missed something—I'll come back to this and write up the steps.

chrisbobbe:
Hmm, for the big picture, I've actually just remembered that Expo has this whole framework that we may (or may not) want to buy into, where you can send push notifications through Expo's servers: https://github.com/expo/expo/tree/master/packages/expo-notifications#add-your-projects-credentials-to-expo-server-optional

Apparently `expo-notifications` is a thing, and you can install it just like any other Unimodule. This didn't occur to me when I followed the deprecation notice at https://reactnative.dev/docs/0.60/pushnotificationios, oops.

chrisbobbe:
Pinging @gnprice on the above (https://github.com/zulip/zulip-mobile/pull/4163#issuecomment-651461201) as it should have occurred to me earlier and it may be worth being aware of in some recent conversations about notifications ([here](https://chat.zulip.org/#narrow/stream/141-kandra-ops/topic/push.20queue.20pages/near/919025) maybe?).

src/notification/index.js
chrisbobbe: Er, actually, no! https://github.com/react-native-community/push-notification-ios/pull/66

chrisbobbe:
OK, just fixed conflicts and adjusted some comments. I also filed https://github.com/react-native-community/push-notification-ios/issues/179 for something I noticed in the library.

src/notification/index.js
chrisbobbe: In my current revision, these comments contain more details, which I'm more sure about.

src/notification/index.js
```diff
@@ -265,29 +256,56 @@ export class NotificationListener {
...
-    // which certainly in the wix case leads to a call to this:
-    //   https://developer.apple.com/documentation/uikit/uiapplication/1623078-registerforremotenotifications
-    // which "initiate[s] the registration process with [APNs]".  Then the
-    // methods that calls on success/failure in turn are implemented to send
-    // an event with name `remoteNotificationsRegistered` etc., in both the
-    // wix and RN-upstream case.  (Though NB in the *failure* case, the
-    // event names differ!  wix has s/Error/Failed/ vs. upstream; also
-    // upstream has singular for failure although plural for success, ouch.)
+    // This leads to a call in RNCPushNotificationIOS to this, to
+    // maybe prompt the user to grant authorization, with options for
```
chrisbobbe, line 267: The Apple documentation for these newer things seems to prefer the term "authorization" over "permissions"; might as well echo that.

chrisbobbe:
I've just fixed some conflicts, and updated a few issues having both the `a-iOS` and `a-notifications` labels (seen in GitHub's UI, just above this comment) as blocked on this work. It would be best to settle what packages we use for notifications on iOS (as this does) before fixing issues with notifications on iOS, at least if they're not p0 issues.

gnprice (commented):
Thanks @chrisbobbe !

We discussed [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20push.20notifications/near/1061130), and I think we want to stick with this line of libraries -- upgrading to the react-native-community library as this PR does -- rather than switch to the Expo notifications library.

Comments below -- all small, I think.

src/notification/index.js
```diff
@@ -234,7 +221,7 @@ export class NotificationListener {
 
   /** Private. */
   handleIOSRegistrationFailure = (err: NotificationRegistrationFailedEvent) => {
-    logging.warn(`Failed to register iOS push token: ${err.domain}:#${err.code}`, {
+    logging.warn(`Failed to register iOS push token: ${err.message}`, {
```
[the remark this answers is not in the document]
chrisbobbe: > Let's be sure to keep the code

Makes sense. I think I was counting on either the code being part of `message` (along with the domain), or that `message` was composed in some way I don't really understand, but that Apple is confident will describe the error well. From the [doc on `localizedDescription`](https://developer.apple.com/documentation/foundation/nserror/1414418-localizeddescription?language=objc):

```
The object in the user info dictionary for the key NSLocalizedDescriptionKey. If the user info dictionary doesn’t contain a value for NSLocalizedDescriptionKey, a default string is constructed from the domain and code.
```

It totally makes sense to at least keep the code by itself in the first argument to `logging.warn`, since they give it to us.

src/notification/index.js
```diff
@@ -247,17 +234,22 @@ export class NotificationListener {
...
-      // straight off the wire from the server.
-      this.listenIOS('notificationOpened', (note: { getData(): JSONableDict }) => {
-        const data = fromAPNs(note.getData());
-        if (data) {
-          this.handleNotificationOpen(data);
+      this.listenIOS('notification', (notification: PushNotificationIOS) => {
+        // This is actually typed as ?Object (and so effectively `any`); but if
+        // present, it must be a JSONable dictionary. (See PushNotificationIOS.js and
+        // RCTPushNotificationManager.m in Libraries/PushNotificationIOS.)
+        const data: ?JSONableDict = notification.getData();
```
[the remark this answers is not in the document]
chrisbobbe: > It might be simpler to appeal to what this method's job is: it's giving us the notification data, which was passed over APNs as JSON.

Would there still be reason to use `JSONableDict`, or would we have to just say `JSONable` on that justification?

src/notification/index.js
```diff
@@ -247,17 +234,22 @@ export class NotificationListener {
...
+        if (!data) {
+          return;
         }
+        const dataFromAPNs: Notification | void = fromAPNs(data);
```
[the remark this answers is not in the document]
chrisbobbe: > (and returning `undefined` when invalid)

It also throws a special kind of error sometimes; it says `@throws An ApnsMsgValidationError on interesting failure.`. Would you say a nullish `data` is an interesting failure? I'm not so sure what would be going on to make `data` nullish.

ios/ZulipMobile/AppDelegate.h
```diff
@@ -1,8 +1,9 @@
 #import <React/RCTBridgeDelegate.h>
 #import <UIKit/UIKit.h>
 #import <UMReactNativeAdapter/UMModuleRegistryAdapter.h>
+#import <UserNotifications/UNUserNotificationCenter.h>
 
-@interface AppDelegate : UIResponder <UIApplicationDelegate, RCTBridgeDelegate>
+@interface AppDelegate : UIResponder <UIApplicationDelegate, RCTBridgeDelegate, UNUserNotificationCenterDelegate>
```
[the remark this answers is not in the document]
chrisbobbe: Mmm, thanks for catching this! I'll add it to that preceding commit.

Those lines don't seem to be documented in React Native's [doc](https://reactnative.dev/docs/0.62/pushnotificationios) for `PushNotificationIOS`, even though that related-looking method did make it into the doc. I think it was written back when iOS 10 (and `UNUserNotificationCenter`) was still new, and they forgot that crucial-looking change. I'm not sure why my manual testing (described in the commit message) didn't turn up any problems, but I think you're quite right that we should include those lines there.

chrisbobbe:
> > Being foregrounded should not suppress badge count updates; this PR should fix this.

> Hmm! Does that mean this fixes #4182? It sounds like it probably does.

It's necessary but not sufficient, I think; we'll have to turn on sending badge counts other than zero, IIRC, and play with that for a bit. I seem to remember we ran into some additional issues when we temporarily turned that on.

chrisbobbe:
(Still working on this; I'll post back here when it's ready for another review.)

chrisbobbe:
OK, I've pushed another revision for review. 🙂 Please also see my comments at https://github.com/zulip/zulip-mobile/pull/4163#discussion_r529070939 and https://github.com/zulip/zulip-mobile/pull/4163#discussion_r529078477.

chrisbobbe:
(Just fixed some conflicts.)

chrisbobbe:
Huh, I've been wondering why PushNotificationsIOS hasn't been removed yet; it's been marked as deprecated for a few new RN releases now.

Turns out it's still used internally at Facebook: https://github.com/facebook/react-native/pull/30026#issuecomment-698627185

chrisbobbe:
(Just rebased and resolved some conflicts.)

ios/ZulipMobile/AppDelegate.m
```diff
@@ -94,30 +99,45 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserAct
...
                                                        fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler
 {
-  [RNNotifications didReceiveRemoteNotification:userInfo];
-  [RCTPushNotificationManager didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
+  [RNCPushNotificationIOS didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
+}
+
+// Required for localNotification event
+- (void)userNotificationCenter:(UNUserNotificationCenter *)center
+didReceiveNotificationResponse:(UNNotificationResponse *)response
```
gnprice, line 125: Hmm, is this comment still right? (The method was `application:didReceiveLocalNotification:`, which more evidently matches this description.)

src/notification/index.js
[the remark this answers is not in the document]
gnprice: Yeah, more precisely: it looks like `fromAPNsImpl` has that `@throws`, and `fromAPNs` catches such errors, logs them, and returns `undefined`.

I'm not sure when that would happen either. That means it'd be sensible to log a warning in that case, just like we do for malformed data blobs. Probably "interesting" wasn't a very clear way to write that description; the distinction that I think it's intended to get at is "unexpected", as in "buggy".

That is:
```
// @returns A `Notification` on success, `undefined` on suppressible failure.
// @throws An ApnsMsgValidationError on interesting failure.
```
* it returns `undefined` when the given data has a form that we know happens, and we understand why, but the app can't currently do anything with.

  There's one such case now, which is when the notification-message is of type `remove` or some unknown future type, rather than type `message` (or no type, which implicitly means `message`.)
* it throws an error when the given data has a form that we don't expect -- meaning there's a bug either in the app or in the server.

And then its caller logs a warning in the latter case -- indicating a bug -- and in both cases returns `undefined` so we do nothing further with the notification.

src/notification/index.js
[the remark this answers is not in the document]
gnprice: I think `JSONableDict` is still fine. The APNs protocol has `data` as a blob which is supposed to be a JSON dictionary.

Though if we can say `JSONable` here and that doesn't cause complications anywhere downstream, then that'd be cleaner still.

gnprice (commented):
Thanks and sorry for the delay!

Small comments below.

The one remaining item that's more significant is: I've lost track a bit of what the state is of badge counts. It seems like this PR will have us start applying badge-count updates more. In a subthread last year https://github.com/zulip/zulip-mobile/pull/4163#discussion_r441739804 , you mentioned testing the change and seeing the badge counts update:

> I've just tested on 26.28.151 (from TestFlight) on CZO. Indeed, I can send myself messages from a test account (increasing the unread count) and go back to the home screen and see that the badge count did not change. Then I send myself another message, while the app is in the background, and the notification appears and the badge count updates. (It jumps to 609...but that's a server-side issue [being investigated](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/907169)).
> 
> Being foregrounded should not suppress badge count updates; this PR should fix this.

But then in that linked chat thread, it looks like a few weeks later we turned off sending badge counts other than zero, at least for iOS.

What's your understanding of the current state of what this will do with badge counts? With current server behavior, will badge counts continue to not appear?

As long as that's the case -- really, as long as this won't make current behavior any worse -- I'm happy merging this and then proceeding to get badge counts working as a separate project.

ios/ZulipMobile/AppDelegate.m
chrisbobbe: Hmm, good question. This is what the doc says in the version we take in this commit: https://github.com/react-native-push-notification-ios/push-notification-ios/blob/v1.5.0/README.md

Also on `master` (as of posting this): https://github.com/react-native-push-notification-ios/push-notification-ios/blob/master/README.md

Looks like it was added in https://github.com/react-native-push-notification-ios/push-notification-ios/pull/104/files.

chrisbobbe:
> badge counts

Good thought to check in about that. I don't expect this PR to change any badge-count behavior for current servers.

That's because I believe current servers send a badge count of 0 with all notifications. (So it would be surprising if we started seeing badges, which only show if the badge count is above zero.) This is corroborated by these observations:
- Without changes to mobile-app code, we were able to see badges for the brief period you mention, when we knew CZO was sending nonzero badge counts.
- Also without changing mobile-app code, we stopped seeing badges when CZO stopped sending badge counts in zulip/zulip@2f66c825a.
- On current `master`, the `get_apns_badge_count_future` from zulip/zulip@2f66c825a still exists and isn't used except in tests, and `get_apns_badge_count` still returns zero.

It seems that iOS delegates control of badge counts to an app's code when the app is foregrounded. The commit `ios notifications: Update the badge count if the app is foregrounded.` accepts that responsibility by updating the badge count according to the notification's badge-count value, instead of not doing so. (And it declines to do other things like play a sound or show the notification banner.) I don't think there's much room for this code to set a wrong badge-count value, if the right value is in the notification; [`UNNotificationPresentationOptionBadge`](UNNotificationPresentationOptionBadge) is what you say when you want to "[a]pply the notification's badge value to the app’s icon".

In chat, you've [mentioned](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/938530) that you think servers might not be sending totally the right shape of thing to convey the badge count. I haven't yet been able to follow the reasoning there; do you still think that might be the case? 🙂 I think you may have also mentioned (on a video call?) that we might have a bug in client-side code for parsing the badge-count value; I don't think I've followed that either, and possibly I'm mis-remembering.

I do think there may have been a server bug where things in the database didn't look right; here's a [message](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/907177) on CZO I'm looking at for that.

I know of one bug that's indicated by Tim's comment, "likely mobile app side issue with handling notifications while the app is open" from zulip/zulip@2f66c825a. It's the one we hope to fix in the badge-count commit in this PR. 🙂

chrisbobbe:
Thanks for the review! Revision pushed. 🙂

ios/ZulipMobile/AppDelegate.m
gnprice: Cool, thanks for the investigation.

I still kind of have a suspicion that that wording is just cloned from the old version -- the diff in the PR looks like this:
```objc
// Required for the localNotification event.
```
→
```objc
// IOS 10+ Required for localNotification event
- (void)userNotificationCenter:(UNUserNotificationCenter *)center
didReceiveNotificationResponse:(UNNotificationResponse *)response
         withCompletionHandler:(void (^)(void))completionHandler
{
  [RNCPushNotificationIOS didReceiveNotificationResponse:response];
  completionHandler();
}
// IOS 4-10 Required for the localNotification event.
```

But :shrug: that's good enough. In a future where we're writing more code of our own here, we'll spend the time to understand this platform API more deeply ourselves, and then we can write our own comment (either more accurate, or just more convincingly accurate, e.g. with a link to docs.)

gnprice:
Cool, that's all very helpful, thanks. I'll go ahead and merge.


> In chat, you've [mentioned](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/938530) that you think servers might not be sending totally the right shape of thing to convey the badge count. I haven't yet been able to follow the reasoning there; do you still think that might be the case?

If we were to swap `get_apns_badge_count_future` back in on the server, I think what it would start sending still might not be right.

But as long as it's hard-coding a badge count of 0, any issues there are still dormant. So we're good to get this PR in, and then consider the next steps there.

There were a few different bugs going on in the investigation in that chat thread:
* A bug where the server thought the right value for the badge count was something like 2, when it should be 0. That's the one discussed where you link to here:
  > I do think there may have been a server bug where things in the database didn't look right; here's a [message](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/907177) on CZO I'm looking at for that.

  I *think* that one is all cleared up; see that thread [at 2020-07-05](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/925672).

* zulip/zulip#18072, the "Server never sends notification, if recipient recently used webapp" bug. That's long before I filed that particular issue thread, but [this message](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/926362) is a clear signature of that bug:
  > Hmm but worse than the summary of that issue [meaning #3659] suggests -- I had a test message at T + 5 minutes, and it never produced a notification

* This bug: we weren't applying the badge-count update, on the client, when the app was in the foreground.

* Some other bug somewhere, which results in the badge count being stuck at 1 once everything's read. Detailed repro [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/926367), 2020-07-06 in that thread.

  In particular, at [this 2020-07-15 message](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20badge.20count/near/938503) (just above the one you linked to about being unsure the server was doing the right thing), my recollection was that I didn't have the app open at all when I did that repro. That'd rule out being caused by the change we're talking about in this PR.

  In that repro, I also saw the badge count successfully updating from 3 down to 1, which wouldn't be consistent with this bug being involved. It's just going from 1 down to 0 that didn't happen.

I think that last bug is the one I had in mind in the message you linked to.

gnprice:
Oh, one other mystery, but this one is OK if it stays a mystery:

From reading the Apple docs following links from [`UNNotificationPresentationOptionBadge`](https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/unnotificationpresentationoptionbadge?language=objc), I don't actually understand why the behavior in the status quo (before this PR) is the way it is.

Namely, when we get a notification while the app is in the foreground, no notification appears in the UI -- there's no banner at the top, and it's not in Notification Center when I swipe from the top of the screen to see that. I just tested that empirically:
* on an iPad mini 4, running iOS 14.6;
* with v27.164, the release I just made yesterday (and [successfully uploaded today](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/ios.20upload) to alpha, and plan to send to beta right after I merge this PR);
* by sending a PM from user A to user B, where user B is logged in on the device.
  * User B has the "send mobile notifications even if offline" setting enabled, so the server should be sending the notification regardless of what the client is doing. But I haven't directly verified that it's doing so.
* What I found is that I reliably get a notification -- both a banner, and in the list in Notification Center -- if the app is not foregrounded, and none if it is.

That behavior reflects what we're explicitly asking for with the change in this PR, passing just `UNNotificationPresentationOptionBadge` to the callback from `userNotificationCenter:willPresentNotification:withCompletionHandler:`.

But if you look at the [docs for that method](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter?language=objc), they say:
> If your delegate does not implement this method, the system behaves as if you had passed the `UNNotificationPresentationOptionNone` option to the `completionHandler` block. If you do not provide a delegate at all for the `UNUserNotificationCenter` object, the system uses the notification’s original options to alert the user.

The behavior we're seeing is consistent with `UNNotificationPresentationOptionNone`.

But in the status quo, I don't think it's the case that we have a delegate for `UNUserNotificationCenter` that doesn't implement this method. Rather, it's that we don't provide a delegate at all. Providing a delegate is what this diff is doing:
```diff
+  // Define UNUserNotificationCenter
+  UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+  center.delegate = self;
```
(… also, now that I've actually read some of the relevant API docs, I see that that comment is totally wrong. We're not defining that object -- the object already exists. The first line has us just getting a pointer to it, for us to use in the second line. The second line has us telling the center what object to use as its delegate object: namely, the `AppDelegate` object, the same one we use as the delegate [for `UIApplication`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate).)

And with some grepping I'm not finding anywhere that our dependencies are providing such a delegate on our behalf, either. E.g. `rg -C4 UNUserNotificationCenter node_modules/` finds references in a handful of files, and they're doing various things you might expect, but none of them appear to be setting the `delegate` property.

So it sure seems like the docs say we should be seeing this behavior:

> If you do not provide a delegate at all for the `UNUserNotificationCenter` object [as we apparently don't], the system uses the notification’s original options to alert the user.

And what does it mean to "use" those "original options"? Well, there's [this reference doc](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification?language=objc#2943363) for composing an APNs payload.

We're providing an `alert` property -- that's what produces the banner and the Notification Center item, when the app isn't in the foreground. So it seems like the behavior we should expect to be seeing now is that it'd be producing the same banner and Notification Center item when the app is in the foreground, too.

We're also providing a `badge` property, of course, with a value of 0. So the same reading of the docs would also predict that we'd see the badge getting set to 0 (and more generally, to whatever the server said in the payload) regardless of whether we're in the foreground or background.

Anyway. EIBTI -- I'm glad to be making it explicit what we want in this situation, and no longer relying on that corner of the API's behavior. Regardless of whether the solution to the mystery is that it's not doing what the docs say it does, or that we are in fact "provid[ing] a delegate" in some nonobvious way.

gnprice:
I've also just sent a quick PR with a small followup refactor to the `fromAPNs` interface: #4810.

chrisbobbe:
Thanks for the review and merge! 🙂

Issue #4158, opened 2020-06-15
Support receiving message ids to delete in bulk.


amanagr:
After https://github.com/zulip/zulip/commit/cda7b2f5393b803f521f45f566b520ebcd9b8d20, client can specify if it supports bulk deletion of messages via passing `bulk_message_deletion = True` in [client_capabilities](https://chat.zulip.org/api/register-queue#arguments) on `/register` request.

See [`do_delete_messages` in action.py on zulip/zulip](https://github.com/zulip/zulip/blob/cda7b2f5393b803f521f45f566b520ebcd9b8d20/zerver/lib/actions.py#L4586) for expected event data formatting in response.

Note that the old event data formatting will be removed in Zulip 2.3 release, so this should be fixed with priority.

chrisbobbe:
Thanks for the heads-up, @amanagr!

IIUC, the `type` of the new bulk-delete event is "delete_message", which is identical to what it is for the now-deprecated event (that only deletes one message): https://github.com/zulip/zulip/blob/cda7b2f5393b803f521f45f566b520ebcd9b8d20/zerver/lib/actions.py#L4596.

This is fine; we can distinguish the new one by its having "message_ids" instead of "message_id".

chrisbobbe:
@gnprice, should we split this into two Redux actions, e.g., the current EVENT_MESSAGE_DELETE action and a new EVENT_MESSAGES_DELETE action, or just modify the existing EVENT_MESSAGE_DELETE action, kind of following the server's choice here?

chrisbobbe:
> Let's follow the structure of the server's event.

Hmm, just so I understand correctly, are you saying we should overwrite our `EVENT_MESSAGE_DELETE` action, reducers, etc., in-place to handle the new (bulk-delete) shape? I agree that we should try to name and arrange things as much as possible to mirror the server API.

But if we want to preserve compatibility with older servers, I think there might have to be some differences, such as having the `EventMessageDeleteAction` type basically be a union of the old shape and the new shape. Does that sound right, and is it a goal to still be able to handle the old shape?

timabbott:
What we do in some server-side testing code is just convert the old shape into the new shape on receipt, and then write all the rest of the code against the new shape.  I think that's an attractive pattern for this (and most other forms of data format migrations, actually).

gnprice:
I agree with Tim's answer completely :slightly_smiling_face:

PR #4164, opened 2020-06-17
events: Add support for bulk delete_message events.

chrisbobbe:
I haven't yet tested end-to-end an actual bulk delete; I'm still looking for a way to do that in the UI on CZO. Maybe it's an admin feature, or it's not accessible through the UI yet?

Fixes: #4158

### DIFF
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -88,10 +88,13 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction => {
         ownEmail: getOwnEmail(state),
       };
 
+    // Before server feature level 13, or if we don't specify the
+    // `bulk_message_deletion` client capability (which we do) this
+    // event has `message_id` instead of `message_ids`.
     case 'delete_message':
       return {
         type: EVENT_MESSAGE_DELETE,
-        messageId: event.message_id,
+        messageIds: event.message_ids ?? [event.message_id],
       };
 
     case EventTypes.stream:
--- a/src/message/__tests__/messagesReducer-test.js
+++ b/src/message/__tests__/messagesReducer-test.js
@@ -90,7 +90,7 @@ describe('messagesReducer', () => {
       const initialState = deepFreeze({ 1: { id: 1 } });
       const action = deepFreeze({
         type: EVENT_MESSAGE_DELETE,
-        messageId: 2,
+        messageIds: [2],
       });
       const newState = messagesReducer(initialState, action);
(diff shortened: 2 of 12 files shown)


chrisbobbe:
OK, just tested this on a dev server, and it works as expected. I see the `EVENT_MESSAGE_DELETE` action with multiple items in `messageIds`, and the messages do disappear.

src/message/__tests__/messagesReducer-test.js
```diff
@@ -117,7 +129,7 @@ describe('messagesReducer', () => {
       });
       const action = deepFreeze({
         type: EVENT_UPDATE_MESSAGE,
-        messageId: 3,
+        messageId: [3],
```
gnprice, line 132: Hmm, it's surprising this one makes it past the type-checker! Would be good to identify why.

(I think this change is unintentional and can be simply dropped.)

src/events/eventToAction.js
```diff
@@ -88,11 +88,15 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction => {
...
+    // Before server feature level 13, or if we don't specify the
+    // `bulk_message_deletion` client capability (which we do) this
+    // event has `message_id` instead of `message_ids`.
+    case 'delete_message': {
       return {
         type: EVENT_MESSAGE_DELETE,
-        messageId: event.message_id,
+        messageIds: event.message_ids ?? [event.message_id],
       };
+    }
```
gnprice, line 99: nit: extra braces not needed

(I'm guessing they came from a previous draft of this commit where you perhaps had a local variable.)

src/message/__tests__/messagesReducer-test.js
chrisbobbe: Thanks for catching this!

> Hmm, it's surprising this one makes it past the type-checker!

Alas, this file isn't type-checked at all.

src/events/eventToAction.js
chrisbobbe: Yep, that was it—thanks for catching this.

chrisbobbe:
Merged. Thanks for the review!

src/message/__tests__/messagesReducer-test.js
gnprice: Ah, of course :-/ Looks like that's still true of a little over half our test files:
```
$ git ls src/**/__tests__/*.js | wc -l
92

$ git grep -l '@flow' src/**/__tests__/*.js | wc -l
39

Issue #4165, opened 2020-06-17
Time out initial fetch, and go to account-picker screen


gnprice:
If you launch the app, and the last server you'd been using is one that is now impossible to reach... currently we just stay at the loading screen forever, endlessly retrying the initial fetch.

If that's the only server you ever wanted to use, then this isn't so bad, given that there's not much useful we can do until the server is back. But if you'd like to go use some *other* server, then you're completely blocked from doing so. If the server doesn't come back, your only remedy is to clear the app's data entirely (or uninstall and reinstall) and log into those other servers from scratch. This is #4159, as described at https://github.com/zulip/zulip-mobile/issues/4159#issuecomment-645574458 .

We should do better. As a minimal version, if the initial fetch takes a completely unreasonable amount of time (maybe one minute), we should give up and take you to the account-picker screen so you can try a different account and server.

(An improved followup version might continue loading but show a message like "Still working..." and an action "Switch account" on the loading screen, after a much shorter period like 7 seconds.)

Here's how I'd go about implementing this:
* The key logic to change is in `doInitialFetch`, around `api.registerForEvents` with the pair of `tryUntilSuccessful` calls.
* That logic already handles the case where the load is outright rejected by the server -- that's the `isClientError` case inside `tryUntilSuccessful`, and the `catch` block outside it that calls `dispatch(logout())`.
* What's needed is to also handle the case where the server simply doesn't respond, or responds with a 5xx error. We don't want to log out in that case, because the credentials may still be perfectly good -- but we do want to abort.
* For actually aborting and going to the account-picker screen, we can probably do the same thing as we do for the "Switch account" button in `ProfileCard.js`: `dispatch(navigateToAccountPicker())`. Then return out of `doInitialFetch`.
* For the logic to get there: first, let's move `tryUntilSuccessful` into the same file as `doInitialFetch` and just above it, and give it a more specific name like `tryFetch`. (This move can be its own commit.) It's always been closely tied to that function's specific needs with its handling of 4xx errors, and we're about to make it more so.
* Then, in the retry loop, check how long it's been since we started -- if it's more than a timeout constant like 60 seconds, throw an exception. In the catch block around the requests, still do `dispatch(logout())` if `isClientError(e)`; but otherwise, abort without logging out.

chrisbobbe:
[...]

PR #4166, opened 2020-06-18
Time out initial fetch, and go to account-picker screen

chrisbobbe:
Fixes: #4165

N.B.: Extracting `promiseTimeout` into a helper function, as @gnprice suggested [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Stuck.20on.20loading.20screen/near/907701), led me to add some time-elapsed arithmetic near the call site so we're not too generous and give each attempt a full 60 seconds. We shrink `timeLimitMs` with each successive call.


chrisbobbe:
Oh, goodness—I do need to add tests for `promiseTimeout`, I think. And probably `tryFetch`; I removed the tests for that under its previous name, `tryUntilSuccessful`, partly because they weren't really doing much. Glad to see that there are already tests for `isClientError`.

gnprice:
Yeah, let's have tests for `promiseTimeout`. :slightly_smiling_face: Then for `tryFetch`, two of the three old tests were pretty trivial, but the third tests that it does retry. It'd also be good to test its new timeout functionality, and the existing `isClientError` feature.

src/utils/async.js
```diff
@@ -1,5 +1,47 @@
...
+
+/**
+ * Time-out a Promise after `timeLimitMs` has passed.
+ *
+ * Returns a new Promise with the same outcome as `promise`, if
+ * `promise` completes in time.
+ *
+ * If `promise` does not complete before `timeLimitMs` has passed,
+ * `onTimeout` is called, and its outcome is used as the outcome of
+ * the returned Promise.
```
gnprice, line 15: I like this term "outcome". It's a good way of conveying this should apply to completion and failure uniformly.

src/utils/async.js
```diff
@@ -1,5 +1,47 @@
...
+ *
+ * If `onTimeout` is omitted, a timeout will cause the returned
+ * Promise to reject with a TimeoutError.
+ *
+ * Interface modeled after
+ * https://api.dart.dev/stable/2.8.4/dart-async/Future/timeout.html.
+ *
+ * @param promise
+ * @param timeLimitMs
+ * @param onTimeout
```
gnprice, line 25: These don't seem to add anything -- the names are already in the signature.

src/utils/async.js
```diff
@@ -1,5 +1,47 @@
...
+  onTimeout?: () => T | Promise<T>,
+): Promise<T> {
+  const timeoutPromise = new Promise((_, reject) =>
+    setTimeout(() => reject(new TimeoutError()), timeLimitMs),
+  );
+  try {
+    return await Promise.race([promise, timeoutPromise]);
+  } catch (e) {
+    if (e instanceof TimeoutError && onTimeout !== undefined) {
+      return onTimeout();
```
gnprice, line 39: This feels like ideally it belongs at the same place as the `new TimeoutError()`, and replacing that in the `onTimeout` case. I think that would simplify the code, because this whole try/catch would then get to go away.

IOW, the whole try/catch could become just
```js
  return await Promise.race([promise, timeoutPromise]);
```
with a suitable version of `timeoutPromise` that itself handles the `onTimeout`.

(And then in fact the `await` could be deleted, along with the `async`: just `return Promise.race(…);`.)

src/message/fetchActions.js
```diff
@@ -222,6 +227,42 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState)
...
+    // seconds, then the second attempt should have a 57-second
+    // timeout. Etc.
+    const timeLimitMs = MAX_TIME_MS - msElapsed;
+    try {
+      return await promiseTimeout<T>(func(), timeLimitMs);
+    } catch (e) {
+      if (isClientError(e) || e instanceof TimeoutError) {
+        throw e;
+      }
+      await backoffMachine.wait();
```
gnprice, line 255: Ideally this would also be covered by the logic expressed by `timeLimitMs` above.

src/message/fetchActions.js
```diff
@@ -222,6 +227,42 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState)
...
+    }
+  }
```
gnprice, line 257: Looking at this and particularly with the comment I just made about the `backoffMachine.wait()`, I wonder: with the spiffy new `promiseTimeout` helper, could all this timeout logic be simplified by using that around the whole thing? I think it could.

src/message/fetchActions.js
gnprice, line 255: A small simplification I notice that applies already to the existing code, but might be helpful for the changes you're making: the `await backoffMachine.wait();` could be pulled out of the `catch` block, to just be at the end of the `while` body.

I think that actually makes the most conceptual sense, too -- the backoff is happening at the bottom of the loop, just before we start a new iteration.

src/message/fetchActions.js
```diff
@@ -222,6 +227,42 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState)
...
 };
 
+/**
+ * Calls an async function and if unsuccessful retries the call.
+ *
+ * If the function is an API call and the response has HTTP status code 4xx
+ * the error is considered unrecoverable and the exception is rethrown, to be
+ * handled further up in the call stack.
+ */
+export async function tryFetch<T>(func: () => Promise<T>): Promise<T> {
```
gnprice, line 237: The jsdoc should grow a mention of the timeout.

src/utils/async.js
chrisbobbe: Ah, I thought of doing it that way—but then, IIUC, it's not a race between (A) the outcome of `promise` and (B) the timer (which is what it says on the tin), but rather, between (A) the outcome of `promise` and (B) the timer *plus* the outcome of a Promise returned by `onTimeout`.

src/message/fetchActions.js
chrisbobbe: Hmm, if I understand correctly, something like this?

```javascript
export async function tryFetch<T>(func: () => Promise<T>): Promise<T> {
  const MAX_TIME_MS: number = 60000;
  const backoffMachine = new BackoffMachine();

  return promiseTimeout<T>(
    (async () => {
      // eslint-disable-next-line no-constant-condition
      while (true) {
        try {
          return await func();
        } catch (e) {
          if (isClientError(e) || e instanceof TimeoutError) {
            throw e;
          }
          await backoffMachine.wait();
        }
      }
      // Without this, Flow 0.92.1 does not know this code is unreachable,
      // and it incorrectly thinks Promise<undefined> could be returned,
      // which is inconsistent with the stated Promise<T> return type.
      // eslint-disable-next-line no-unreachable
      throw new Error();
    })(),
    MAX_TIME_MS,
  );
}
```

chrisbobbe: Oh, hmm, no, that wouldn't do it—the infinite loop could continue infinitely even after the timer was up. I'll see what I can think of.

chrisbobbe: Er, maybe this isn't the most elegant solution, but it might work?

```javascript
export async function tryFetch<T>(func: () => Promise<T>): Promise<T> {
  const MAX_TIME_MS: number = 60000;
  const backoffMachine = new BackoffMachine();

  let timerHasExpired = false;

  return promiseTimeout(
    (async () => {
      // eslint-disable-next-line no-constant-condition
      while (true) {
        if (timerHasExpired) {
          // No one is listening for this Promise's outcome, so stop
          // doing more work.
          throw new Error();
        }
        try {
          return await func();
        } catch (e) {
          if (isClientError(e)) {
            throw e;
          }
          await backoffMachine.wait();
        }
      }
      // Without this, Flow 0.92.1 does not know this code is unreachable,
      // and it incorrectly thinks Promise<undefined> could be returned,
      // which is inconsistent with the stated Promise<T> return type.
      // eslint-disable-next-line no-unreachable
      throw new Error();
    })(),
    MAX_TIME_MS,
    () => {
      timerHasExpired = true;
      throw new TimeoutError();
    }
  );
}
```

src/message/fetchActions.js
chrisbobbe: Makes sense.

src/message/fetchActions.js
chrisbobbe: Ah, right; thanks, I'll do that.

src/message/fetchActions.js
chrisbobbe: Hmm, true—see one possible way this might look, at https://github.com/zulip/zulip-mobile/pull/4166#discussion_r442655214.

src/utils/async.js
chrisbobbe: Mm, bumping this, @gnprice, as I don't think we got around to it on the call (we were mostly talking about https://github.com/zulip/zulip-mobile/pull/4166#discussion_r442571136).

chrisbobbe: Concretely, I think you're suggesting something like this:

```javascript
  const timeoutPromise = new Promise((resolve, reject) =>
    setTimeout(
      () => (onTimeout !== undefined ? resolve(onTimeout()) : reject(new TimeoutError())),
      timeLimitMs,
    ),
  );
  return await Promise.race([promise, timeoutPromise]);
```

chrisbobbe: But this extends the effective timeout duration by the duration of a Promise returned by `onTimeout`; I'm writing a test against this subtle case right now. 😛

chrisbobbe: `resolve(onTimeout())` might look a bit weird if `onTimeout` returns a Promise; I'm not used to doing that. But I think this excerpt supports the idea that the outcome of `onTimeout()` (the Promise returned by `onTimeout`) is assigned to be the outcome of `timeoutPromise`.

It's from a big green note [here](https://tc39.es/ecma262/#sec-promise-executor), in a section on the Promise constructor in an ECMAScript spec (formatting lost):

> The resolve function that is passed to an executor function accepts a single argument. The executor code may eventually call the resolve function to indicate that it wishes to resolve the associated Promise object. The argument passed to the resolve function represents the eventual value of the deferred action and can be either the actual fulfillment value or another Promise object which will provide the value if it is fulfilled.

chrisbobbe: I'll include a test for this case and you can play around with it if you have doubts. 🙂

chrisbobbe:
> It'd also be good to test its new timeout functionality, and the existing isClientError feature.

It looks like there are already tests for `isClientError`; did you have more thoroughness in mind?

I just pushed my latest revision, so this is ready for another review.

src/utils/async.js
gnprice: Hmm, good point. And good thought to add a test for it. :slightly_smiling_face: 

It still feels like this logic is sort of awkwardly split between the `new TimeoutError()` and the later `catch` and `if (e instanceof TimeoutError && onTimeout !== undefined)`. But evidently it'd take some deeper thinking about these promise and timer semantics to fix that than I'd thought. And there isn't *that* much to gain -- the behavior seems correct as it is, and the code is already short and not hard to understand.

chrisbobbe:
Closing, as this revision has grown old and doesn't have working tests—https://github.com/zulip/zulip-mobile/pull/4193 is working on that.

PR #4168, opened 2020-06-20
[WIP] Add npx to build commands in Build-run.md

onurtemiz:
Changed commands such as "react-native run android" with "npx react-native run-android" in Build-run.md file.


chrisbobbe:
I'm thinking of closing this. Did you encounter an error after following the instructions in our setup docs? I haven't found a need to use `npx` in my work, so far, but possibly that's an incidental feature of my environment.

Also, the two occurrences of `react-native run-android` have been replaced with `tools/run-android`, so these changes won't apply there. The only thing this PR will touch once it's rebased is the `react-native run-ios` line.

Issue #4157, opened 2020-06-15
Add mobile support for new `user_avatar_url_field_optional` client capability


timabbott:
To resolve https://github.com/zulip/zulip/pull/15287, we're introducing a new client_capability that should allow the mobile app to have good performance when talking to servers with thousands of long-term-idle users (and email_address_visibility configured; that last bit being relevant mainly in that our previous attempt at solving this problem. `client_gravatar` feature works only with EMAIL_ADDRESS_VISIBILITY_EVERYONE), since the client needs real email addresses to compute gravatar hashes.

One should be able to prototype today against https://github.com/zulip/zulip/pull/15359; it should work aside from having the wrong capability name.  I expect that to get cleaned up and this to merge in the next few days.

Tagging as a priority since this issue makes chat.zulip.org a lot slower to load on mobile.

PR #4171, opened 2020-06-24
Prep commits for new `avatar_url` handling (#4157)

chrisbobbe:
In a revision of my work for #4157, I noticed that there were some necessary changes to some test files, and some of these files haven't been type-checked or taken advantage of the work we've put into `src/__tests__/lib/exampleData.js`.

Rather than continue editing these files in this unstable state, pave the way for better edits by type-checking them and using that example data. 🙂  And fix a few bugs I've noticed.

The revision that informed these changes was one where the type for `avatar_url` on the `Message` type (not just that field on the `User` type) was changed. I *think* this might be a direction we'll go in (e.g., the type might be a union of all subclasses of `AvatarURL` except `IdleAvatarURL`); I'm thinking of the discussion around [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Mobile.20API.20changes/near/912176). In any case, that explains why `fetchActions-test.js` was touched (`fetchMessages` is in there).

There will be some more of this; I know at least `src/webview/__tests__/webViewHandleUpdates-test.js` has some stuff that I'm still figuring out.

### DIFF
--- a/src/api/messages/__tests__/migrateMessages-test.js
+++ b/src/api/messages/__tests__/migrateMessages-test.js
@@ -1,12 +1,53 @@
+/* @flow strict-local */
+import omit from 'lodash.omit';
+
 import { migrateMessages } from '../getMessages';
+import * as eg from '../../../__tests__/lib/exampleData';
+import type { ServerMessage, ServerReaction } from '../getMessages';
+import type { Message } from '../../modelTypes';
 
 describe('migrateMessages', () => {
-  test('Replace user object with `user_id`', () => {
-    const messages = [
-      { reactions: [{ user: { id: 1, full_name: 'name', email: 'a@a.com' }, code: 'code' }] },
-    ];
-    const expectedOutput = [{ reactions: [{ user_id: 1, code: 'code' }] }];
+  const reactingUser = eg.makeUser();
 
-    expect(migrateMessages(messages)).toEqual(expectedOutput);
+  const serverReaction: ServerReaction = {
+    emoji_name: '+1',
+    reaction_type: 'unicode_emoji',
+    emoji_code: '1f44d',
+    user: {
+      email: reactingUser.email,
+      full_name: reactingUser.full_name,
+      id: reactingUser.user_id,
+    },
+  };
+
+  const serverMessage: ServerMessage = {
+    // The `omit` shouldn't be necessary with Flow v0.111: "Spreads
+    // now overwrite properties instead of inferring unions"
+    // (https://medium.com/flow-type/spreads-common-errors-fixes-9701012e9d58).
+    ...(omit(eg.streamMessage(), 'reactions'): $Diff<Message, { reactions: mixed }>),
+    reactions: [serverReaction],
+  };
+
+  const input: ServerMessage[] = [serverMessage];
+
+  const expectedOutput: Message[] = [
+    {
+      // The `omit` shouldn't be necessary with Flow v0.111.
+      ...(omit(serverMessage, 'reactions'): $Diff<ServerMessage, { reactions: mixed }>),
+      reactions: [
+        {
+          user_id: reactingUser.user_id,
+          emoji_name: serverReaction.emoji_name,
+          reaction_type: serverReaction.reaction_type,
+          emoji_code: serverReaction.emoji_code,
+        },
+      ],
+    },
+  ];
+
+  const actualOutput: Message[] = migrateMessages(input);
+
+  test('Replace user object with `user_id`', () => {
+    expect(actualOutput.map(m => m.reactions)).toEqual(expectedOutput.map(m => m.reactions));
   });
 });
--- a/src/webview/html/__tests__/render-test.js
+++ b/src/webview/html/__tests__/render-test.js
@@ -1,14 +1,16 @@
+/* @flow strict-local */
 import messageTypingAsHtml from '../messageTypingAsHtml';
+import * as eg from '../../../__tests__/lib/exampleData';
 
 describe('typing', () => {
-  it('escapes values', () => {
-    expect(
-      messageTypingAsHtml('&<r', [
-        {
-          avatar_url: '&<av',
-          email: '&<e',
-        },
-      ]),
-    ).not.toContain('&<');
+  it('escapes &< (e.g., in `avatar_url` and `email`', () => {
+    const name = '&<name';
+    const user = {
+      ...eg.makeUser({ name }),
+      avatar_url: `https://zulip.example.org/yo/avatar-${name}.png`,
+      email: `${name}@example.com`,
+    };
+
+    expect(messageTypingAsHtml('&<r', [user])).not.toContain('&<');
   });
 });
--- a/src/api/messages/getMessages.js
+++ b/src/api/messages/getMessages.js
@@ -19,7 +19,7 @@ type ApiResponseMessages = {|
  * Note that reaction events have a *different* variation; see their
  * handling in `eventToAction`.
  */
-type ServerReaction = $ReadOnly<{|
+export type ServerReaction = $ReadOnly<{|
   ...$Diff<Reaction, {| user_id: mixed |}>,
   user: $ReadOnly<{|
     email: string,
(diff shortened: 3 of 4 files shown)


flow-typed/redux-mock-store_v1.2.x.js
```diff
@@ -5,12 +5,15 @@ declare module 'redux-mock-store' {
   */
 
   declare type mockStore = { <S, A>(state: S): mockStoreWithoutMiddleware<S, A>, ... };
-  declare type DispatchAPI<A> = (action: A) => A;
-  declare type Dispatch<A: $ReadOnly<{ type: string, ... }>> = DispatchAPI<A>;
+  declare interface Dispatch<S, A> {
```
chrisbobbe, line 8: Mm, the `S` type argument is unnecessary (making a comment so I remember to address this in my next revision, along with code-review feedback).

chrisbobbe: (and a code comment here saying that we've edited this bit would be good too)

chrisbobbe:
Just resolved some conflicts; awaiting review. 🙂

chrisbobbe:
Just rebased again, following the RN v0.61 upgrade (no conflicts), awaiting review. 🙂

chrisbobbe:
Ah, oops: Just found a few things to do better, so I pushed again; hopefully this didn't interrupt a review. I'll move on to something else now so you don't have to worry about possible interruptions again, and I'll post here if for some reason I need to make more changes.

chrisbobbe:
Mm, looks like there are some conflicts—fixing those now.

chrisbobbe:
And now more conflicts, from commits in https://github.com/zulip/zulip-mobile/pull/4187—I'll fix these now.

src/api/messages/__tests__/migrateMessages-test.js
```diff
@@ -1,12 +1,49 @@
...
+    emoji_code: '1f44d',
+    user: {
+      email: reactingUser.email,
+      full_name: reactingUser.full_name,
+      id: reactingUser.user_id,
+    },
+  };
+
+  const serverMessage: ServerMessage = {
+    ...(omit(eg.streamMessage(), 'reactions'): $Diff<Message, { reactions: mixed }>),
```
gnprice, line 24: Is the `omit` needed here? The property mentioned after the spread should clobber it anyway.

src/webview/html/__tests__/render-test.js
```diff
@@ -1,14 +1,11 @@
...
   it('escapes values', () => {
-    expect(
-      messageTypingAsHtml('&<r', [
-        {
-          avatar_url: '&<av',
-          email: '&<e',
-        },
-      ]),
-    ).not.toContain('&<');
+    const user = eg.makeUser({ name: '&<av' });
```
gnprice, line 7: Does this still have the `&<` sequence in the avatar_url? One of the things this is testing is that that gets properly encoded.

src/message/__tests__/fetchActions-test.js
```diff
@@ -72,6 +74,7 @@ describe('fetchActions', () => {
 
       expect(actions).toHaveLength(1);
       expect(actions[0].type).toBe('MESSAGE_FETCH_START');
+      expect(actions[0].numBefore).not.toBeGreaterThan(0);
     });
 
     test('when no messages to be fetched after the anchor, fetchingNewer is false', () => {
```
gnprice: (missed this one)

src/api/messages/__tests__/migrateMessages-test.js
chrisbobbe: Definitely sounds reasonable.

It seems like Flow happily verifies that there will be a correct type at `reactions` after the spread, but only with the `omit` (or something like it) in place. Even though it's in the spec (er, I'm pretty sure) that later-listed properties will clobber earlier-listed ones, it seems like Flow maybe doesn't know that, or in any case doesn't like it when there are two competing (and both present) values for a single property.

If I take away the `omit`:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/api/messages/__tests__/migrateMessages-test.js:23:40

Cannot assign object literal to serverMessage because:
 • property user is missing in object type [1] but exists in object type [2] in array element of property reactions.
 • property user_id is missing in object type [2] but exists in object type [1] in array element of property
   reactions.

     src/api/messages/__tests__/migrateMessages-test.js
      20│     },
      21│   };
      22│
      23│   const serverMessage: ServerMessage = {
      24│     ...eg.streamMessage(),
      25│     reactions: [serverReaction],
      26│   };
      27│
      28│   const input: ServerMessage[] = [serverMessage];
      29│

     src/api/messages/getMessages.js
 [2]  33│   reactions: $ReadOnlyArray<ServerReaction>,

     src/api/modelTypes.js
 [1] 493│   reactions: $ReadOnlyArray<Reaction>,
```

src/webview/html/__tests__/render-test.js
chrisbobbe: It currently does, meaning `eg.makeUser` has this for `avatar_url`:

```
`https://zulip.example.org/yo/avatar-${name}.png`,
```

But that's up to `exampleData` to choose. Might as well pin to a particular `avatar_url` and `email` here by clobbering those properties; I'll do that.

chrisbobbe:
Thanks for the review! Just pushed those changes. You may want to see my note at https://github.com/zulip/zulip-mobile/pull/4171#discussion_r459224854 about the `omit`, but it's not Earth-shattering; I'll merge as-is in a day or two if you don't get around to reading it.

chrisbobbe:
Actually, since you said at https://github.com/zulip/zulip-mobile/pull/4171#pullrequestreview-453847396 that it otherwise looks good, I'll go ahead and merge the changes to the `fetchActions` tests; that'll be helpful for me to get #4205 to a more reviewable state. The one comment you made on that (https://github.com/zulip/zulip-mobile/pull/4171#discussion_r459226185) was straightforwardly addressed (thanks for catching it!).

chrisbobbe:
> Actually, since you said at [#4171 (review)](https://github.com/zulip/zulip-mobile/pull/4171#pullrequestreview-453847396) that it otherwise looks good, I'll go ahead and merge the changes to the `fetchActions` tests; that'll be helpful for me to get #4205 to a more reviewable state. The one comment you made on that ([#4171 (comment)](https://github.com/zulip/zulip-mobile/pull/4171#discussion_r459226185)) was straightforwardly addressed (thanks for catching it!).

OK, I ended up merging all the commits except for two you commented on that I don't have an immediate need for, so you can see my responses to them when you get a chance (https://github.com/zulip/zulip-mobile/pull/4171#discussion_r461234078 and https://github.com/zulip/zulip-mobile/pull/4171#discussion_r461235748).

chrisbobbe:
I [said](https://github.com/zulip/zulip-mobile/pull/4171#issue-438964898):
> There will be some more of this; I know at least `src/webview/__tests__/webViewHandleUpdates-test.js` has some stuff that I'm still figuring out.

OK, just pushed this. It was a little tricky getting `prevProps` and `nextProps` working with Flow, but I think I've found a good balance. I've decided not to bother faking the `intl` property of `props._`, but we don't currently have tests that use that, and I don't think we will soon.

src/api/messages/__tests__/migrateMessages-test.js
gnprice: Hmm, indeed.

It looks like this is one of the issues discussed here:
https://medium.com/flow-type/spreads-common-errors-fixes-9701012e9d58
> Spreads now overwrite properties instead of inferring unions

and it's fixed in Flow v0.111.

Which, checking our next RN upgrade issue #3782... we'll get with that upgrade to RN v0.62! Should be a lot of Flow improvement in that upgrade -- that'll be a happy day.

gnprice: So this is a fine workaround; but let's add a comment mentioning Flow v0.111. (So that when we do that upgrade we can easily find this in a grep and make it one of the spots where we improve the types as a followup.)

gnprice:
The new commit lgtm!

Re `omit`, see reply above about a quick comment to add. Then please merge at will. Thanks for making all these tests better!

chrisbobbe:
OK, done, thanks for the review!

PR #4172, opened 2020-06-26
ios native modules: Expose "AppIdentifierPrefix".

chrisbobbe:
Following discussion [1] about a SO answer (not currently the
accepted one) [2], we find it's not too hard, when we want the App
ID prefix (sometimes called the Team ID), to

1. Add a custom entry in ios/ZulipMobile/Info.plist that gets set to
   that string at build time, and
2. Write some minimal Objective-C to add a native module that calls
   an iOS API to grab that custom entry and exposes it to the
   JavaScript.

We followed a very clear React Native doc [3] on how to do this.

We also considered using a different doc [4] that would have helped
us package our code up very conveniently for distribution to NPM,
with a third-party tool called `create-react-native-module`. The
output of that tool seemed to have all the appropriate metadata that
we're used to seeing in modules in the React Native ecosystem, such
as a "podspec" file.

But our goal isn't to distribute this code for wide use. It would
still be possible to, e.g., host it on GitHub (or even locally) and
point to it in `dependencies` in our `package.json`. But this seems
like a lot more overhead than it'll ever be worth, and we avoided
doing something like that in 0b84717e4, an already-merged part of
issue #4124, where we just wrote some Kotlin code and a few lines in
our `MainApplication.java` to import it.

We put the two new files in the same directory that has
`UtilManager.h` and `UtilManager.m`; these were also created for
needs particular to Zulip. This time, we follow the pattern of
adding a short, all-caps prefix ZM, for ZulipMobile, as seen in

- Expo -> EX
- Unimodules -> UM
- React Native -> RCT or RN

To link the new files, we used the Xcode UI:

1. Found the PBXGroup called "ZulipMobile" in the left sidebar; in
   Xcode 11.5, it has a folder icon and includes the
   `Info.plist` as well as `UtilManager.h` and `UtilManager.m`.

2. Right-clicked it and selected `Add files to "ZulipMobile"...`,
   and chose those files.

3. Selected the ZulipMobile target in the "project and targets
   list", and went to Build Phases -> Compile Sources to add the
   files there.

4. Saw that, surprisingly, they were already added there. So, job
   done.

Then, we tested with some in-app logging to see that we can indeed
access the Team ID. With NativeModules imported from react-native,
we see it at `NativeModules.ZMInfoProperties.appIdentifierPrefix`.

[1]: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20native.20module/near/916750
[2]: https://stackoverflow.com/a/28714850/6792075
[3]: https://reactnative.dev/docs/native-modules-ios
[4]: https://reactnative.dev/docs/native-modules-setup#docsNav

### DIFF
--- a/ios/ZulipMobile.xcodeproj/project.pbxproj
+++ b/ios/ZulipMobile.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		13B07FC11A68108700A75B9A /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
 		42689E9C23466FF7007540AA /* webview in Resources */ = {isa = PBXBuildFile; fileRef = 42689E9B23466FF7007540AA /* webview */; };
+		8BEA1EF024A6983D00674040 /* ZLPInfoProperties.m in Sources */ = {isa = PBXBuildFile; fileRef = 8BEA1EEE24A6983D00674040 /* ZLPInfoProperties.m */; };
 		A148FEFC1E9D8CB900479280 /* zulip.mp3 in Resources */ = {isa = PBXBuildFile; fileRef = A148FEFB1E9D8CB900479280 /* zulip.mp3 */; };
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Thanks for the review ([in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/iOS.20native.20module/near/917025))! Just pushed a new revision; ready for another review.

gnprice:
Looks good, thanks! Merged.

Issue #4134, opened 2020-06-03
Allow viewing message edit history


chrisbobbe:
The endpoint it looks like we'll want to use is documented [here](https://zulipchat.com/api/get-message-history).

We have a binding for it, in `src/api/messages/getMessageHistory.js`, that currently isn't used.

Conveniently, just as with messages themselves, it looks like the diff we want to display, `content_html_diff`, is HTML, which we can render in a WebView.

I think a major question we'll need to answer first is **whether the edit history should be viewed in the same WebView used to display the message list, or if it should be a new WebView**.

There's been some discussion around [here](https://chat.zulip.org/#narrow/stream/2-general/topic/zulip-mobile.20settings/near/872463) of introducing a new WebView for the particular reason that we want to replicate a complicated (but useful) bit of the web app's UI so we don't have to reinvent it ourselves: in that case, the settings UI (which indeed has a lot of stuff going on).

Here, though:

* The stuff we want to display isn't particularly various or complicated: a list of edit items, where each edit item shows the diff, the timestamp, and the person who made the edit (I'm guessing there's an admin feature to allow non-authors to edit messages).

* Also, Tim says this about the web app's implementation:
  
  @timabbott [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Edited.20history/near/893875):
  > the webapp's implementation of it is one of our least polished features, but people clearly do like it despite that.

Given these facts, I'm currently inclined against dropping the web app's implementation into its own WebView (with the tweaks necessary to make that possible), even though it might be a win for developing our code-sharing infrastructure.

So, if we do put it in its own WebView (separating it from the MessageList), then we should do it for different reasons. Possibly, just wanting to avoid complicating the existing WebView is sufficient. I'm leaning that way, currently; presumably, we'd be responding to a long-press, or something in the action sheet, and we're already outside the message list's WebView at that point. But I could anticipate hesitation about a multiplicity of WebViews.

@gnprice, thoughts?

gnprice:
> if we do put it in its own WebView (separating it from the MessageList), then we should do so for different reasons. Possibly, just wanting to avoid complicating the existing WebView is sufficient. I'm leaning that way, currently; presumably, we'd be responding to a long-press, or something in the action sheet, and the bulk of the logic for those handlers is already already outside the message list's WebView.

That sounds like a good reason to me.

[...]

PR #4173, opened 2020-06-29
Allow viewing message edit history

agrawal-d:
image

Fixes: #4134


agrawal-d:
>I'm not sure why, but I'm just getting a blank page in the WebView on my iOS device; it just says about:blank. Did you run into this on Android, and was there something special you had to do to get it to work?

Oh no!, that's bad!
Nope, I never ran into this on Android, and didn't do anything special to get it to work. The only command I needed to run this is `yarn run android`.

agrawal-d:
The [API Guide](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#reference) says:
>Note that using static HTML requires the WebView property originWhiteList to ['*'].

Does changing `originWhitleist` in `EditHistory.js `to this fix the issue for you on iOS?  I can't test on iOS as I don't have an iOS device.

src/webview/html/editHistoryHtml.js
```diff
@@ -0,0 +1,105 @@
...
+      displayedContent = template`Topic:
+      <span class="highlight_text_inserted">${snapshot.topic}</span>
+      <span class="highlight_text_deleted">${snapshot.prev_topic}</span>
+      <br/>$!${snapshot.content_html_diff}`;
+      break;
+    }
+
+    default:
+      // eslint-disable-next-line no-console
+      console.error('Invalid edit_flag');
```
chrisbobbe, line 75: Let's get this logged to Sentry, with `logging.error`, and include the value of `edit_flag` in the second argument, `extras`.

src/webview/html/editHistoryHtml.js
```diff
@@ -0,0 +1,105 @@
...
+    edit_flag = CONTENT_EDITED;
+  } else if (snapshot.prev_content === undefined && snapshot.prev_topic !== undefined) {
+    edit_flag = TOPIC_EDITED;
+  } else {
+    edit_flag = BOTH_EDITED;
+  }
+
+  switch (edit_flag) {
+    case NOT_EDITED: {
+      tag += template`<span class="message-tag">Original message</span>`;
```
chrisbobbe, line 33: When you're looking at the original message, is the Markdown rendered properly?

I'm seeing the raw text; not sure why. But it seems like it's been possible to show everything properly formatted when viewing edits in the list, so it would be great to get the original message formatted properly too if we can (and it's that way in the web app). 🙂

agrawal-d: Oops, I placed `content` instead of `rendered_content`!

src/webview/html/editHistoryHtml.js
agrawal-d: Ok
(It's actually impossible to reach here :p)

agrawal-d:
Thanks for the review, @chrisbobbe, and for the detailed investigation on CZO. I've pushed some changes, and I think it may work on iOS too, for now.

src/webview/html/editHistoryHtml.js
gnprice: > (It's actually impossible to reach here :p)

Quick tip: you can add `ensureUnreachable(edit_flag);` here to assert this to the type-checker (and ask it to verify that it can prove it.) That helps prevent errors before they even reach users :slightly_smiling_face: if say a new case is added here in the future.

chrisbobbe:
(For the record: the comments I made on CZO were [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/924227)).

src/webview/html/editHistoryHtml.js
agrawal-d: I'm getting the following error on adding that:
```js
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/webview/html/editHistoryHtml.js:76:25

Cannot call ensureUnreachable with edit_flag bound to x because:
 • string [1] is incompatible with empty [2].
 • string [3] is incompatible with empty [2].
 • string [4] is incompatible with empty [2].
 • string [5] is incompatible with empty [2].

     src/webview/html/editHistoryHtml.js
 [1] 10│ const TOPIC_EDITED = 'TOPIC_EDITED';
 [3] 11│ const CONTENT_EDITED = 'CONTENT_EDITED';
 [4] 12│ const BOTH_EDITED = 'BOTH_EDITED';
 [5] 13│ const NOT_EDITED = 'NOT_EDITED';
       :
     73│     }
     74│
     75│     default:
     76│       ensureUnreachable(edit_flag);
     77│       logging.error('Invalid edit_flag', { editFlag: edit_flag });
     78│       break;
     79│   }

     src/types.js
 [2] 42│ export function ensureUnreachable(x: empty) {}



Found 4 errors
```

I think Flow is not able to understand that this case is unreachable?

src/api/modelTypes.js
```diff
@@ -350,19 +350,16 @@ export type Reaction = $ReadOnly<{|
 export type MessageSnapshot = $ReadOnly<{|
   user_id: number,
   timestamp: number,
+  topic: string,
+  content: string,
+  rendered_content: string,
 
-  /** Docs unclear but suggest absent if only content edited. */
-  topic?: string,
```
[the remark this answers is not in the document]
gnprice: That discussion here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/get-message-history.20docs/near/921587

Whenever we find a place where the API docs seem to be wrong, I always want to discuss it in chat and bring it to Tim's attention there. Reasons include:
* I want the API docs to converge on being right too :slightly_smiling_face: -- if we have a steady state where our types say one thing and the API docs say another, that'll be confusing to a later reader
* If for whatever reason the API docs don't get updated promptly, I'll want the comments in our code to be super explicit that the API docs say X but the truth is Y, and here's how we've determined the truth is Y. That gives the future reader a hope of not being confused by the contradiction, because at least they can see that the author of the types was aware of what the API docs said, and wasn't just misreading them, or working from what at that future time is an out-of-date version
* The answer might be that the documented API is as intended, and the server's behavior that disagrees is a bug!

As of right now (since a few minutes ago), the status is that Tim confirms the current behavior is what you're seeing empirically -- but potentially wants to fix it by changing the behavior to match the docs instead of vice versa.

src/api/modelTypes.js
```diff
@@ -350,19 +350,16 @@ export type Reaction = $ReadOnly<{|
...
-   * They definitely say "prev"/"diff" properties absent on the first snapshot.
-   */
-  content?: string,
-  rendered_content?: string,
+  // The prev_* field is only present if the respective attribute
+  // of the message was changed.
   prev_content?: string,
   prev_rendered_content?: string,
   content_html_diff?: string,
+  prev_topic?: string,
```
gnprice, line 362: This summary is concise in a similar way to the corresponding bit of the API docs -- but I think it would actually be clearer (and not even many more words, if more at all) to be less abstract, more concrete and explicit:
```js
  // These are present just if the content was edited.
  prev_content?: string,
￼  prev_rendered_content?: string,
￼  content_html_diff?: string,

  // Present just if the topic was edited.
  prev_topic?: string,
```

(... OK, and I just counted: that's 16 words instead of 15. :grin: )

src/webview/html/editHistoryHtml.js
gnprice: Ah -- it's that the type of `edit_flag` is simply `string`, instead of being an enum `'TOPIC_EDITED' | 'CONTENT_EDITED' | 'BOTH_EDITED' | 'NOT_EDITED'`.

You could make it an enum by annotating it with that type (and then probably doing something to avoid initializing to `undefined` and mutating.)

Or, looking a bit more at the code: this flag is so short-lived, just from an if/else-if chain to a `switch` right after it. I think it'd simplify things to just fuse those together: have an if/else-if chain with those conditions, and with the cases' bodies as their bodies. Or, a bit further: a two-level tree of if/else statements:
```js
if (snapshot.prev_content === undefined) {
  if (snapshot.prev_topic === undefined) {
  …
  } else {
  …
  }
} else {
  …
}

File: src/webview/html/editHistoryHtml.js [hunk 5]
```diff
@@ -0,0 +1,105 @@
...
+    html += renderSnapshot(snapshot, usersById);
+  }
+
+  return template`
+$!${script(null, auth)}
+$!${css(theme)}
+<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">
+<body style="overflow-x: hidden;">
+$!${html}
+</body>`;
```
gnprice, line 104: There are some subtle bits here :slightly_smiling_face:. Particularly that `meta` element -- I can only guess at what it does, and would have to look it up.

And this sure has a lot in common with the code in `html.js`! In particular, the subtle bits are exactly the same. (Then the latter has one additional subtle bit, `webkitBugWorkaround`, which is likely unnecessary here but would be harmless.)

Anything that's subtle, I am keen on not duplicating. When code is duplicated, future bugfixes and incidental changes tend to go on just one or the other -- and then it's self-reinforcing: as they diverge, it gets harder to make the same change on both of them or be sure the change would do the same thing on both, so more changes get applied to just the one where the bug happened to be spotted. So because we need this in two places, let's factor it out into one place we can call from both of them.

src/webview/html/editHistoryHtml.js
```diff
@@ -0,0 +1,105 @@
...
+    default:
+      logging.error('Invalid edit_flag', { editFlag: edit_flag });
+      break;
+  }
+
+  return template`
+    <div class="edit-history-block">
+      <div class="static-timestamp">${editedTime}</div>
+      <span class="edit-history-content">$!${displayedContent}</span>
+      <div class="message-tags">$!${tag}</div>
```
gnprice, line 83: I appreciate the careful use of ``template`…` ``, `${…}`, and `$!${…}`! You may already know that this is an important surface area for preventing HTML injections.

Which we actually had, until a couple of years ago -- the day I discovered that, I [fixed it](https://github.com/zulip/zulip-mobile/commit/994145383f9478e1e2ae1b40c9b77f222ebe72d7) and then built this little templating/escaping system to help us prevent it ever happening again.

I've looked closely at these templates, and I believe everything gets escaped correctly.

A couple of things I try to do to help keep it that way, as people make changes in the future:
* Background ideas:
  * Conceptually there are two types here: text string, and HTML string. Sadly we haven't taught the type-checker the difference -- they're both `string` -- so we have to keep track of the distinction ourselves.
  * The ultimate return value is an HTML string. Most of the inputs are text strings, and in general any string that isn't explicitly meant to be used as raw HTML is a text string. The only way to get an HTML string from a text string is to hand it to `template`; and an interpolation in `template` should have `$!$` if it's an HTML string and plain `$` if it's a text string.
  * So the goal is to be able to look at a use of `$` or `$!$` -- especially a use of `$!$` -- and immediately tell, without a lot of reading through other code, whether it looks right, or should be the other way instead.
* One thing that helps: For each variable that represents a fragment that'll get used in templates, give it a name that's explicit about whether it's a text string or an HTML string. Especially if it's an HTML string, make that explicit. That way when there's an interpolation like `$!${someVar}`, and it's not a name like `$!${someHtml}`, it jumps out as suspicious.
  * This is why `messageHeaderAsHtml.js` has variables `topicHtml`, for example.
  * So here, `tagHtml` would be a good name for `tag`. Similarly `displayedContentHtml`.
* Another thing that helps: make it so that every last backtick-interpolated string in the HTML-rendering code uses `template`. That way any that don't are suspicious.
  * This commit adds one non-`template` backtick-literal: `editedTime`. The behavior is fine, because it truly is a text string, and it gets duly escaped when it's used. But it'd be good to maintain that invariant, to help make any future oversights stand out. One simple way would be to just inline that expression in the place where `editedTime` is used.

For history see 994145383f9478e1e2ae1b40c9b77f222ebe72d7, where I fixed the HTML injection (in 2018-05.) Some discussion of these hygiene practices in af5a56f41d6229faa8b60b42781951892a5a70f1, a few commits later; `template` itself was introduced in the latter's parent.

src/webview/static/base.css
```diff
@@ -475,3 +475,30 @@ blockquote {
...
+.highlight_text_inserted {
+  color: #9effa1;
+  background-color: hsla(120,64%,95%,.3);
+}
+.highlight_text_deleted {
+  color: red;
+  background-color: #fcdcd8;
+  text-decoration: line-through;
+  word-break: break-all;
+}
```
gnprice, line 504: Are these based on something in the webapp, or did you develop them afresh?

Either is fine :slightly_smiling_face: -- I'd just like to know (and have the commit message say so for future readers), and if it's from the webapp have a quick comment in the file saying where exactly it's from. (Including if it's based loosely or in part on something in the webapp.) That will save us some archaeological work in any future effort to unify the message CSS on web and mobile.

src/message/EditHistory.js
```diff
@@ -0,0 +1,88 @@
...
+        <WebView
+          source={{ baseUrl, html }}
+          originWhitelist={['file://']}
+          style={{ backgroundColor: 'transparent' }}
+          onShouldStartLoadWithRequest={getOnShouldStartLoadWithRequest(baseUrl)}
+          onError={(msg: mixed) => {
+            // eslint-disable-next-line no-console
+            console.error(msg);
+          }}
+        />
```
gnprice, line 79: This is another example of a subtle set of details I'm keen on not duplicating :slightly_smiling_face: .

Most of this should get factored out to go along with `getOnShouldStartLoadWithRequest` -- the `originWhitelist`, the transparent-background `style`, the `onShouldStartLoadWithRequest`. The bits that the factored-out code (perhaps a new React component?) should take as parameters (or React props) are something like: `html`, the `onMessage` and `onError` callbacks, and maybe `baseUrl`.

agrawal-d:
I've somehow requested a review from you twice :stuck_out_tongue: 

image

src/webview/MessageList.js
```diff
@@ -112,52 +111,11 @@ export type Props = $ReadOnly<{|
...
-//
-// [4] https://github.com/react-native-community/react-native-webview/blob/v5.12.1/ios/RNCWKWebView.m#L376
-//     https://github.com/facebook/react-native/blob/v0.59.10/React/Base/RCTConvert.m#L85
-
 class MessageList extends Component<Props> {
   static contextType = ThemeContext;
   context: ThemeColors;
 
-  webview: ?WebView;
+  webview = null;
```
chrisbobbe, line 118: Hmm, what was the reason to lose the type annotation here? Is it possible to bring it back? If it can't be brought back simply as-is, maybe we could try with the newer `React.createRef` API? (My example at https://github.com/zulip/zulip-mobile/pull/4101#discussion_r449259543 might be helpful.)

src/common/WebView.js
```diff
@@ -0,0 +1,165 @@
+/* @flow strict-local */
+import React from 'react';
+import { Platform } from 'react-native';
+import { WebView as RNWebView } from 'react-native-webview';
+import type { WebViewMessageEvent, WebViewNavigation } from 'react-native-webview';
+import type {} from '../types';
```
chrisbobbe, line 6: 🤔

src/common/WebView.js
```diff
@@ -0,0 +1,165 @@
+/* @flow strict-local */
```
chrisbobbe, line 1: f1a7acaa1 webview: Refactor core logic into a common component.

Cool! It's great to see this core logic centralized so we don't have to duplicate it.

With the choice to do it in a new React component, one thing to be aware of (even if nothing actually changes, and it turns out to be fine) is that it adds a layer of complexity if we want to change the behavior (or debug something) around React components' lifecycles, particularly updates, a.k.a. re-`render`s. I mentioned this briefly [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/926743) but didn't go into very much detail.

The `WebView` component from `react-native-webview` has its own ideas about when it should update; it can choose to be a `PureComponent` or a regular `Component`, and, if the latter, it can give hints about its desired behavior in `shouldComponentUpdate`.

There's now another layer between that `WebView` and our `MessageList` (and between that `WebView` and the new `EditHistory`, but I think it's much more important to think about `MessageList`).

Our `MessageList` component implements `shouldComponentUpdate` to say that we just never want the `MessageList` component to update. (React would probably say this is controversial; they recommend, "Do not rely on it to “prevent” a rendering, as this can lead to bugs", in [this doc](https://reactjs.org/docs/react-component.html#shouldcomponentupdate). But so far we haven't run into problems with this.)

I haven't tested, but I think everything should actually turn out fine. I would be concerned if our new `WebView` component did some additional updates of its own, not prompted by an update in its parent. The only case I can think of where this would likely happen is if our new `WebView` component had some `state` that it decided, on its own, to change. ("Note that returning `false` does not prevent child components from re-rendering when `their` state changes," from the [`shouldComponentUpdate` doc](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).)

One (optional) simple thing that could be done about this case, for a bit of peace of mind, I think, is to make our new `WebView` a stateless "function component" ([doc](https://reactjs.org/docs/components-and-props.html#function-and-class-components)) and add a note that we shouldn't introduce `state` without considering the extra complexity it would add around re-`render`ing.

chrisbobbe: I also think, just in case we do eventually have to do some hairy debugging around components' lifecycle methods, or just to be clear about what's what, in general, it would be nice to name our new `WebView` something like `ZulipWebView`. We have this kind of name for, e.g., `ZulipButton`, which wraps a component called `Button`.

chrisbobbe:
Thanks, @agrawal-d! I think https://github.com/zulip/zulip-mobile/pull/4183 may be about to land (I'll go resolve some conflicts), which will change the logic to get the `baseUrl`.

src/webview/MessageList.js
agrawal-d: The innerRef is actually a ref to `react-native-webview`, so to annotate the type, I need to import both `ZulipWebView` and `WebView` in `MessageList.js`. Done this now.

agrawal-d:
Thanks for the review @chrisbobbe, I've made the changes you suggested. Now I guess I'll wait for #4183 to land.

chrisbobbe:
> Now I guess I'll wait for #4183 to land.

Just landed, in the series ending in https://github.com/zulip/zulip-mobile/commit/ae7ad996583ca40553af67fe86d32268bde3ab26! 🙂

agrawal-d:
@chrisbobbe I've rebased this PR. It was a bit convoluted, so if I missed something from your commits, please let me know.

chrisbobbe:
I just rebased and resolved some conflicts. I also made a small change (getting the theme name for `EditHistory` using `react-redux` instead of as a navigation param), but I haven't yet done a more thorough review. I hope to do so soon! 🙂

Issue #4162, opened 2020-06-17
New markdown feature: Mention a time


timabbott:
We added a new markdown syntax, `!time(2020-06-18T17:32:00-07:00)`, that is meant to be a timezone-aware time displayed to all clients in local time (using the same mechanisms used in the timestamps in Zulip messages).

There's likely some discussion that will happen on how these should be rendered; my current leaning is that we should put a clock icon/emoji at the start of the pills.  But what's important for now is the core logic to make this feature work at all.  In the webapp, it's this part of `rendered_markdown.js`:

```
    content.find('span.timestamp').each(function () {                                      
        // Populate each timestamp span with mentioned time                                
        // in user's local timezone.                                                       
        const timestamp = moment.unix($(this).attr('data-timestamp'));                     
        if (timestamp.isValid() && $(this).attr('data-timestamp') !== null) {              
            const text = $(this).text();                                                   
            const rendered_time = timerender.render_markdown_timestamp(timestamp,          
                                                                       null, text);        
            $(this).text(rendered_time.text);                                              
            $(this).attr('title', rendered_time.title);                                    
        } else {                                                                           
            $(this).removeClass('timestamp');                                              
            $(this).attr('title', 'Could not parse timestamp.');                           
        }                                                                                  
    });                                                                               

     // Renders the timestamp returned by the !time() markdown syntax.                          
exports.render_markdown_timestamp = function (time, now, text) {                           
    now = now || moment();                                                                 
    if (page_params.timezone) {                                                            
        now = now.tz(page_params.timezone);                                                
        time = time.tz(page_params.timezone);                                              
    }                                                                                      
    const timestring = time.format('ddd, MMM D YYYY, h:mm A');                             
    const titlestring = "This time is in your timezone. Original text was '" + text + "'."\
;                                                                                          
    return {                                                                               
        text: timestring,                                                                  
        title: titlestring,                                                                
    };                                                                                     
};                                                                                         
```

Implementing this should be a priority, since the whole point is that the time gets rendered properly :).

timabbott:
The syntax for these is changing to use HTML 5 time elements: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

So the markdown will be `<time datetime="{datetime_string}">{original_text}</time>` once we merge  https://github.com/zulip/zulip/pull/15433.

timabbott:
[...]

PR #4175, opened 2020-07-01
webview: Add support for 'time' elements. 

agrawal-d:
Add support for rendering 'time' mentions in message text. The
mentioned time is formatted to be human-readable, and tapping on it
shows an alert dialog with the original text.

**Screenshots**

image

image

image

Fixes: #4162

### DIFF
--- a/src/webview/js/__tests__/rewriteHtml.js
+++ b/src/webview/js/__tests__/rewriteHtml.js
@@ -1,10 +1,10 @@
 /** @jest-environment jsdom-global */
 // @flow strict-local
 
-import rewriteImageUrls from '../rewriteImageUrls';
+import rewriteHtml from '../rewriteHtml';
 import type { Auth } from '../../../types';
 
-describe('rewriteImageUrls', () => {
+describe('rewriteHtml', () => {
   const realm = 'https://realm.example.com';
   global.jsdom.reconfigure({ url: 'file:///nowhere_land/index.html' });
 
@@ -18,7 +18,7 @@ describe('rewriteImageUrls', () => {
     const img = document.createElement('img');
     img.setAttribute('src', src);
     const before = img.src;
-    rewriteImageUrls(auth, img);
+    rewriteHtml(auth, img);
     const after = img.src;
     return { input: src, before, after };
   };
--- a/src/webview/js/rewriteHtml.js
+++ b/src/webview/js/rewriteHtml.js
@@ -14,11 +14,8 @@ const inlineApiRoutes: RegExp[] = ['^/user_uploads/', '^/thumbnail$', '^/avatar/
  *      than the document location.
  *   2. If the source URL names an endpoint known to require authentication,
  *      inject an API key into its query parameters.
- *
- * DEPRECATED: If no root element is specified, transform every <img> in the
- * entire document.
  */
-const rewriteImageUrls = (auth: Auth, element: Element | Document = document) => {
+const rewriteImageUrls = (auth: Auth, element: Element | Document) => {
   const realm = new URL(auth.realm);
 
   // Find the image elements to act on.
@@ -56,4 +53,14 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
   });
 };
 
-export default rewriteImageUrls;
+/**
+ * Modify relevant parts of the rendered DOM tree under the given root,
+ * inclusive, to ensure functionality.
+ *
+ * DEPRECATED: If no root element is specified, the entire document is considered.
+ */
+const rewriteHTML = (auth: Auth, element: Element | Document = document) => {
+  rewriteImageUrls(auth, element);
+};
+
+export default rewriteHTML;
--- a/src/webview/js/rewriteHtml.js
+++ b/src/webview/js/rewriteHtml.js
@@ -54,6 +54,39 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document) => {
 };
 
 /**
+ * Rewrite the 'time' elements under the given root, inclusive, to make them more readable.
+ *
+ * 1. Changes text to localized human readable date-time.
+ * 2. Adds 'original-text' attribrute, to show original text alert dialog on press.
+ */
+const rewriteTime = (element: Element | Document) => {
+  // Find the time elements to act on.
+  const timeElements = [].concat(
+    element instanceof HTMLTimeElement ? [element] : [],
+    Array.from(element.getElementsByTagName('time')),
+  );
+
+  timeElements.forEach(elem => {
+    // Present only because the Flow DOM libdef does not have a specific definition
+    // for `getElementsByTagName('time')`, so the typedef of `timeElements` is wider
+    // than it actually is. See https://github.com/facebook/flow/issues/8450.
+    if (!(elem instanceof HTMLTimeElement)) {
+      return;
+    }
+
+    const timeStamp = elem.dateTime;
+    const text = elem.innerText;
+
+    const d = new Date(timeStamp);
+    elem.setAttribute('original-text', text);
+    elem.innerText = `🕒  ${d.toLocaleString(undefined, {
+      dateStyle: 'full',
+      timeStyle: 'short',
+    })}`;
+  });
+};
+
+/**
  * Modify relevant parts of the rendered DOM tree under the given root,
  * inclusive, to ensure functionality.
  *
@@ -61,6 +94,7 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document) => {
  */
 const rewriteHTML = (auth: Auth, element: Element | Document = document) => {
   rewriteImageUrls(auth, element);
+  rewriteTime(element);
 };
 
 export default rewriteHTML;
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -475,3 +475,9 @@ blockquote {
   height: 32px;
   fill: hsla(0, 0%, 100%, 0.75);
 }
+time {
+  display: inline-block;
+  padding: 4px 6px;
+  background-color: hsla(0, 0%, 50%, 0.125);
+  border-radius: 8px;
+}
--- a/src/webview/webViewEventHandlers.js
+++ b/src/webview/webViewEventHandlers.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import { Clipboard } from 'react-native';
+import { Clipboard, Alert } from 'react-native';
 import * as api from '../api';
 import config from '../config';
 import type { Dispatch, GetText, Message, Narrow, Outbox, EditMessage } from '../types';
@@ -119,6 +119,11 @@ type MessageListEventMention = {|
   userId: number,
 |};
 
+type MessageListEventTimeDetails = {|
+  type: 'time',
+  originalText: string,
+|};
+
 export type MessageListEvent =
   | MessageListEventReady
   | MessageListEventScroll
@@ -132,7 +137,8 @@ export type MessageListEvent =
   | MessageListEventDebug
   | MessageListEventWarn
   | MessageListEventError
-  | MessageListEventMention;
+  | MessageListEventMention
+  | MessageListEventTimeDetails;
 
 type Props = $ReadOnly<{
   backgroundData: BackgroundData,
@@ -263,6 +269,18 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
     case 'mention': {
       const { dispatch } = props;
       dispatch(navigateToAccountDetails(event.userId));
+      break;
+    }
+
+    case 'time': {
+      const alertText = _.intl.formatMessage(
+        {
+          id: "This time is in your timezone. Original text was '{originalText}'.",
+          defaultMessage: "This time is in your timezone. Original text was '{originalText}'.",
+        },
+        { originalText: event.originalText },
+      );
+      Alert.alert('', alertText);
       break;
     }
 
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -14,38 +14,6 @@ export default `
 var compiledWebviewJs = (function (exports) {
   'use strict';
 
-  var inlineApiRoutes = ['^/user_uploads/', '^/thumbnail$', '^/avatar/'].map(function (r) {
-    return new RegExp(r);
-  });
-
-  var rewriteImageUrls = function rewriteImageUrls(auth) {
-    var element = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : document;
-    var realm = new URL(auth.realm);
-    var imageTags = [].concat(element instanceof HTMLImageElement ? [element] : [], Array.from(element.getElementsByTagName('img')));
-    imageTags.forEach(function (img) {
-      var actualSrc = img.getAttribute('src');
-
-      if (actualSrc == null) {
-        return;
-      }
-
-      var fixedSrc = new URL(actualSrc, realm);
-
-      if (fixedSrc.origin === realm.origin) {
-        if (inlineApiRoutes.some(function (regexp) {
-          return regexp.test(fixedSrc.pathname);
-        })) {
-          var delimiter = fixedSrc.search ? '&' : '';
-          fixedSrc.search += "".concat(delimiter, "api_key=").concat(auth.apiKey);
-        }
-      }
-
-      if (img.src !== fixedSrc.toString()) {
-        img.src = fixedSrc.toString();
-      }
-    });
-  };
-
   function _classCallCheck(instance, Constructor) {
     if (!(instance instanceof Constructor)) {
       throw new TypeError("Cannot call a class as a function");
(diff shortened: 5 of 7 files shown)


agrawal-d:
The first commit 5f6b7a4 is borrowed from my PR https://github.com/zulip/zulip-mobile/pull/4169.

src/webview/webViewEventHandlers.js
```diff
@@ -125,7 +130,8 @@ export type MessageListEvent =
   | MessageListEventReactionDetails
   | MessageListEventDebug
   | MessageListEventError
-  | MessageListEventMention;
+  | MessageListEventMention
+  | MessageListTime;
```
chrisbobbe, line 134: Let's have this start with `MessageListEvent`, like the others. Then maybe have something a bit more descriptive; `MessageListEventTimeDetails` (looking at `MessageListEventReactionDetails`, above) or `MessageListEventPressTime`, maybe.

The names for the events going into and out of the WebView are very confusing; one part of #3913 is to fix that (see https://github.com/zulip/zulip-mobile/pull/3913/commits/6c9ff80864780c2e51164a642202d9de408d5ca5). But that PR has gone rather stale, and I'm hoping to get back to it before too long—likely after your PR lands, though.

src/webview/js/rewriteHtml.js
```diff
@@ -58,11 +58,37 @@ export const rewriteImageUrls = (auth: Auth, element: Element | Document = docum
...
+ * 1. Changes text to localized human readable date-time.
+ * 2. Adds 'original-text' attribrute, to show original text alert dialog on press.
+ */
+const rewriteTime = () => {
+  const timeElements: NodeList<HTMLTimeElement> = document.querySelectorAll('time');
+  timeElements.forEach(elem => {
+    const timeStamp = elem.getAttribute('datetime');
+    const text = elem.innerText;
+
+    // 'text' and 'timeStamp' are guaranteed to be present.
```
chrisbobbe, line 72: This comment, and the code below it, look a little funny to me, like, "This will definitely be here. If it's not...".

I think the check is probably being done to satisfy Flow that these things will, in fact, be present. A brief explanation of that fact would help this make more sense, so it doesn't seem self-contradictory. 🙂

src/webview/js/rewriteHtml.js
```diff
@@ -58,11 +58,37 @@ export const rewriteImageUrls = (auth: Auth, element: Element | Document = docum
...
+    if (text == null || timeStamp == null) {
+      return;
+    }
+
+    const d = new Date(timeStamp);
+    elem.setAttribute('original-text', text);
+    elem.innerText = `🕒  ${d.toLocaleString(undefined, {
```
chrisbobbe, line 79: If you pass `undefined`, I wonder if it's smart enough to detect the correct locale, e.g., for putting the month before the day or vice versa: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#Parameters

I think maybe it is; this is running in a WebView, which the device has some control over—hopefully enough to tell it the appropriate device-level setting.

agrawal-d:
Thanks for the review! I've pushed some changes. 
I've removed the `If no root element is specified, transform every  in the entire document.` comment from `rewriteImageUrls` because it's private, and is only called by `rewriteHtml`, which always passes an element or document to it - it's redundant and also incorrect.

Instead, as you suggested, I've added a similar comment to `rewriteHtml` instead.

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+      timeStyle: 'short',
+    })}`;
+  });
+};
+
+/**
+ * Modify relevant parts of the rendered DOM tree under the given root,
+ * inclusive, to ensure functionality.
+ *
+ * If no root element is specified, the eintire document is considered.
```
chrisbobbe, line 93: nit: 'entire"

src/webview/js/__tests__/rewriteHtml.js
```diff
@@ -1,7 +1,7 @@
 /** @jest-environment jsdom-global */
 // @flow strict-local
 
-import rewriteImageUrls from '../rewriteImageUrls';
```
chrisbobbe, line 4: fa7f7539a webview [nfc]: Refactor rewriteImageUrls to rewriteHtml.

Commit message nits: "Rename", "create".

agrawal-d: I thought this form ("Renames") of commit message body is the one recommended by Zulip [Docs](https://zulip.readthedocs.io/en/latest/contributing/version-control.html)?

>The body is written in prose, with full paragraphs; each paragraph should be separated from the next by a single blank line.

chrisbobbe: Ah, I think the mobile project diverges from the server project in this respect. We acknowledge that the "what" of a commit message may overflow the summary line, into the body, and we prefer keeping the same syntactic form (often called the imperative; I'm on the fence about whether it's kind of a base infinitive, it would be good to figure that out 😆). @gnprice has posted some thoughts about this at https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622188734.

gnprice: Note that as described in that comment, there's variation on this within the server project :slightly_smiling_face: -- some of us consistently use the same syntactic form as in the summary line, others consistently use the indicative (like "Renames"), and Tim varies between the two. But in the mobile repo we've generally been consistent in using the same form as in the summary line.

src/webview/webViewEventHandlers.js
```diff
@@ -278,6 +284,17 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
...
+        {
+          id: "This time is in your timezone. Original text was '{originalText}'.",
+          defaultMessage: "This time is in your timezone. Original text was '{originalText}'.",
+        },
+        { originalText: event.originalText },
+      );
+      Alert.alert('', alertText);
+      break;
+    }
     default:
```
gnprice, line 298: nit: blank line

src/webview/webViewEventHandlers.js
```diff
@@ -278,6 +284,17 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
       logging.error('WebView error', event.details);
       break;
 
+    case 'time': {
+      const alertText = _.intl.formatMessage(
+        {
+          id: "This time is in your timezone. Original text was '{originalText}'.",
+          defaultMessage: "This time is in your timezone. Original text was '{originalText}'.",
```
gnprice, line 291: This feels redundant, as it's just the same string as `id`. We don't use it in the definition of `_` -- we just pass `id` there. (I searched for `formatMessage` to look for examples.)

Can you take this `defaultMessage` line out, and does it all still work as expected?

Based on a quick look at the API docs:
https://formatjs.io/docs/react-intl/api/#formatmessage
https://formatjs.io/docs/react-intl/api/#message-formatting-fallbacks
they agree with our practice at `_`, that it should work.

src/webview/static/base.css
```diff
@@ -475,3 +475,9 @@ blockquote {
   height: 32px;
   fill: hsla(0, 0%, 100%, 0.75);
 }
+time{
```
gnprice, line 478: nit: space before `{`

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
   });
 };
 
-export default rewriteImageUrls;
+/* Rewrite the 'time' elements under the given root, inclusive, to make them more readable.
```
gnprice, line 58:
```suggestion
/**
 * Rewrite the 'time' elements under the given root, inclusive, to make them more readable.
```

This way it's marked as jsdoc.

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+ *
+ * 1. Changes text to localized human readable date-time.
+ * 2. Adds 'original-text' attribrute, to show original text alert dialog on press.
+ */
+const rewriteTime = (element: Element | Document) => {
+  // Find the image elements to act on.
```
gnprice, line 64: The copy-paste is showing :wink:

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+    element instanceof HTMLTimeElement ? [element] : [],
+    Array.from(element.getElementsByTagName('time')),
+  );
+
+  timeElements.forEach(elem => {
+    const timeStamp = elem.getAttribute('datetime');
+    const text = elem.innerText;
+
+    // 'text' and 'timeStamp' are guaranteed to be present. The conditional is
+    // present only to satisfy Flow.
```
gnprice, line 75: This kind of comment makes me ask: how do you know / why are they guaranteed? Or put another way: what do you know that Flow doesn't? That's the question I most want answered from a comment here.

There's a similar conditional above in this file which provides a nice example of answering that question:
```
    // Not `img.src`, because that getter absolutizes relative URLs itself.
    const actualSrc = img.getAttribute('src');

    // A missing `src` is invalid per the HTML spec.
    if (actualSrc == null) {
      return;
    }
```

OK, Flow doesn't know that fact. It *does* actually know that an `HTMLImageElement` must have a `src: string` -- but we avoided `img.src` for the reason mentioned in the other comment, and instead wrote `img.getAttribute('src')`, and Flow can't make the connection between the string passed to `getAttribute` and the type.

But it's not obvious to me that there's a good excuse like that here.

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+ */
+const rewriteTime = (element: Element | Document) => {
+  // Find the image elements to act on.
+  const timeElements = [].concat(
+    element instanceof HTMLTimeElement ? [element] : [],
+    Array.from(element.getElementsByTagName('time')),
+  );
+
+  timeElements.forEach(elem => {
+    const timeStamp = elem.getAttribute('datetime');
```
gnprice, line 71: Is there a reason not to use the more direct `elem.datetime` here?

If you're copying from the example of `rewriteImageUrls`, note that there's a comment there explaining the reason we sadly can't use `img.src` at the corresponding spot there.

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,46 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
 
-export default rewriteImageUrls;
+/* Rewrite the 'time' elements under the given root, inclusive, to make them more readable.
+ *
+ * 1. Changes text to localized human readable date-time.
+ * 2. Adds 'original-text' attribrute, to show original text alert dialog on press.
+ */
+const rewriteTime = (element: Element | Document) => {
+  // Find the image elements to act on.
+  const timeElements = [].concat(
```
gnprice, line 65: When I hover over `elem` below, Flow reports the type:
> HTMLElement | HTMLTimeElement | any

I think that's the root of at least some of the need you're finding to add a conditional that should be impossible. In reality these elements should all be of type `HTMLTimeElement`.

Note how that's arranged in the other function:
```js
  const imageTags: $ReadOnlyArray<HTMLImageElement> = [].concat(
    element instanceof HTMLImageElement ? [element] : [],
    Array.from(element.getElementsByTagName('img')),
  );
```
with the type annotation. Try that here?

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +55,14 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
   });
 };
 
-export default rewriteImageUrls;
+/**
+ * Modify relevant parts of the rendered DOM tree under the given root,
+ * inclusive, to ensure functionality.
+ *
+ * If no root element is specified, the entire document is considered.
```
gnprice, line 93: This feature had been marked as "DEPRECATED". Is there a reason to drop that?

src/webview/js/rewriteHtml.js
```diff
@@ -15,10 +15,9 @@ const inlineApiRoutes: RegExp[] = ['^/user_uploads/', '^/thumbnail$', '^/avatar/
  *   2. If the source URL names an endpoint known to require authentication,
  *      inject an API key into its query parameters.
  *
- * DEPRECATED: If no root element is specified, transform every <img> in the
- * entire document.
+ * Private; exported only for tests.
  */
-const rewriteImageUrls = (auth: Auth, element: Element | Document = document) => {
+export const rewriteImageUrls = (auth: Auth, element: Element | Document) => {
```
gnprice, line 20: Instead of exporting this, let's just have the tests refer to the renamed `rewriteHTML`.

That way it's entirely an implementation detail of this module that its functionality is broken up into a couple of different functions. That seems like the most appropriate way to think of where its interface is.

src/webview/js/rewriteHtml.js
agrawal-d: I did not know the correct meaning of deprecated. I thought it meant 'no longer used' when I removed it - because it was being used and so I thought that the "DEPRECATED" tag was wrong.

Just now I searched for the meaning, and it means "usable but regarded as obsolete and best avoided, typically because it has been superseded", so the comment is not incorrect. I'll revert it.

src/webview/js/rewriteHtml.js
agrawal-d: I was not aware that it's possible to refer to element-specific attributes directly.

agrawal-d:
Thanks for the review, @gnprice! I've made the changes you suggested.

src/webview/webViewEventHandlers.js
agrawal-d: It works.

gnprice: Hmm, I see. I agree, then. Seems like rather a wart in that API.

src/webview/js/__tests__/rewriteHtml.js
```diff
@@ -1,10 +1,10 @@
...
+import rewriteHTML from '../rewriteHtml';
```
gnprice, line 4: Something's not quite right about this line :wink:

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +53,54 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+    Array.from(element.getElementsByTagName('time')),
+  );
+
+  timeElements.forEach(elem => {
+    // Present only because the Flow DOM libdef does not have a specific definition
+    // for `getElementsByTagName('time')`, so the typedef of `timeElements` is wider
+    // than it actually is.
+    if (!(elem instanceof HTMLTimeElement)) {
+      return;
+    }
```
gnprice, line 75: Hmm, really? I see this in `dom.js` in my latest /tmp/flow/flowlib_* directory:
```js
  getElementsByTagName(name: 'time'): HTMLCollection<HTMLTimeElement>;
```

gnprice: ... Oho. It has that on `Document`... but not on `Element`.

This would be great to report on the Flow issue tracker :slightly_smiling_face: : https://github.com/facebook/flow/issues
Then the comment can just link to the issue.

src/webview/js/rewriteHtml.js
```diff
@@ -56,4 +53,54 @@ const rewriteImageUrls = (auth: Auth, element: Element | Document = document) =>
...
+
+    const timeStamp = elem.dateTime;
+    const text = elem.innerText;
+
+    // 'text' is guaranteed to be present, as the server always supplies it.
+    // The conditional is present only to satisfy Flow.
+    if (text == null) {
+      return;
+    }
```
gnprice, line 84: Is this conditional actually needed? The code gets a lot shorter if it's taken out -- and simpler to think about, as there's one fewer path through the control flow.

At a quick experiment, I get no Flow errors if I just remove this.

src/webview/webViewEventHandlers.js
```diff
@@ -270,6 +276,18 @@ export const handleMessageListEvent = (props: Props, _: GetText, event: MessageL
       console.debug(props, event); // eslint-disable-line
       break;
 
+    case 'time': {
```
gnprice, line 279: Let's not break up `debug` from `warn` and `error` -- as is, they form a nice sequence of closely related cases.

src/webview/js/__tests__/rewriteHtml.js
agrawal-d: I can't figure out what's wrong! :P

src/webview/js/rewriteHtml.js
agrawal-d: Done. https://github.com/facebook/flow/issues/8450

src/webview/webViewEventHandlers.js
agrawal-d: Oh yes! Sorry about this :(

agrawal-d:
Thanks for the review, @gnprice! I've pushed some changes.

I'm still confused by https://github.com/zulip/zulip-mobile/pull/4175#discussion_r463424159 :p

src/webview/js/__tests__/rewriteHtml.js
chrisbobbe: Ah, well spotted, Greg! I see it too—check the variable against the path, letter-by-letter. 🙂

agrawal-d: Oh, got it now! Capitalization!

agrawal-d:
Fixed the capitalization error now :)

gnprice:
And merged! Thanks for the revisions, and for reporting that Flow issue :memo:

Issue #3800, opened 2020-01-14
New messages simply disappear when starting the app without internet


agrawal-d:
A spinner is shown when you were connected when you launched the app, but while staying on the app, became disconnected, and posted a new message
![image](https://user-images.githubusercontent.com/7714968/72384295-25848600-3743-11ea-8636-5b0d57f7df35.png)

However, when starting the app **already disconnected from the internet**, any new message posted **simply disappears**, and no spinner/error is shown. Happens both in PMs and in Streams.

This message appears back upon re-connecting to the internet.

chrisbobbe:
Thanks for the report, @agrawal-d! Possibly related is issue https://github.com/zulip/zulip-mobile/issues/3281. A PR was made to fix that issue, but it was reverted when some tests failed. Greg suspects the code change itself is OK, but the tests need to be changed.

agrawal-d:
I tested #3281, it does not fix this issue.

gnprice:
Interesting! I don't think we've otherwise known about this issue. Marking as a priority, because messages disappearing (even temporarily so that it looks like they have completely) is a bad symptom.

The one issue I see with similar symptoms is #3117. But those reports don't mention starting up without internet.

I think the next step for debugging this would be to identify what step is going wrong in our normal sequence for sending a message. This goes through `addToOutbox` in `outboxActions.js`; after the `messageSendStart(..)` action is dispatched, the message ought to be in Redux under `state.outbox`, and that should cause it to appear in the message list. So questions include:
* Is that happening? (If not, why not?)
* If yes, is the outbox message promptly disappearing again from `state.outbox`? (If so, why?)
* If not, why is the outbox message not appearing in the message list?

A good tool for answering some of these initial questions is to use RN's remote JS debugging and look at the console, at the information [we have `redux-logger` log](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#redux-logger) there. For questions that doesn't answer, I'd add some `console.log` calls.

chrisbobbe:
I'm curious whether we might still be affected by the upstream issues with `fetch` that surfaced in https://github.com/zulip/react-native/pull/2 (June 2018). That's a PR to our fork of react-native, which we'd abandoned in 4b95b124bf2b2fa4c7698c3a013cebffb3d43c57 (March 2017), in favor of using the published react-native. So, apparently the issues addressed in that PR were serious enough that we considered going back to our react-native fork after almost a year without it.

[...]

Issue #3829, opened 2020-01-21
Time out sending a message, and indicate failure


gnprice:
Currently, when we try to send a message and can't, we keep the message around (as an `Outbox` message) and keep retrying indefinitely until it succeeds.

(Specifically, we do so in a loop with backoff, then at each subsequent message send and at each "initial fetch", so in particular at each launch of the app. See `sendOutbox` and its call sites.)

This isn't great, because most messages are eventually better to send never than send late. In particular a short conversational reply like "sure" or "no, sorry" may have a completely different meaning if the discussion has moved on.

Instead, when the oldest outbox message reaches a certain age (perhaps 60 seconds? 120 seconds?) and we've been unable to send, we should stop trying to send it; give the user feedback that it failed; and give an option to retry the message or delete it. This is similar to the experience in an SMS client, or in SMS-style chat apps like FB Messenger.

The feedback is important (so that users can be confident that messages *are* delivered whenever they don't see the feedback), and a bit tricky UI-wise:
* In the message list, the failed-outbox message should continue to show up, but instead of a spinner should have a "warning" or "problem" indicator. E.g., a red ⚠ (U+2640 WARNING SIGN, :warning:) or similar.
  * Perhaps also a bit of text right there with an error message like "Failed to send this message. Tap for options." (Or maybe long-press instead?)
  * On tap (or long-press), we'd show an action sheet with two options labelled like "Retry" and "Delete".
* Here's the trickier part: what if the user hit send and then navigated away, and assumed the message got sent? We'll want to somehow call their attention to the fact it didn't, so they can decide what to do.
  * I think what an SMS-style app would do here is make the situation visible in the conversation list.
  * For us: maybe treat the failed-send message as unread?

gnprice:
See also #3731 .

ray-kraesig:
> * On tap (or long-press), we'd show an action sheet with two options labelled like "Retry" and "Delete".

Also "Edit message", I think.

> * For us: maybe treat the failed-send message as unread?

That would probably be better than nothing, but ideally we should have a distinction (visible and otherwise) between the two. Notably, if the user sends a message and then mutes the relevant topic, they'll probably still want to be informed that the message couldn't be sent.

chrisbobbe:
This may warrant a separate issue, but I wonder if the progressiveTimeout, which is used here, a) might be improved with the option to reject after a certain number of tries, and b) is doing what it's currently intended to do.

[...]

PR #4179, opened 2020-07-01
outbox: Show outbox messages when 'caughtUp.newer' is false.

agrawal-d:
Currently, we return a NULL_ARRAY in outboxMessagesForNarrow when
'caughtUp.newer' is false - but there is no reason to do this. We
should always show outbox messages regardless of the caughtUp state
because there can always be unsent messages for various reasons.

So, remove the conditional that causes this.

Fixes: #3800.

### DIFF
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -75,7 +75,7 @@ describe('getMessagesForNarrow', () => {
     expect(anchor).toEqual(expectedState);
   });
 
-  test('do not combine messages and outbox if not caught up', () => {
+  test('combine messages and outbox if not caught up', () => {
     const state = deepFreeze({
       narrows: {
         [HOME_NARROW_STR]: [123],
@@ -94,7 +94,7 @@ describe('getMessagesForNarrow', () => {
       ],
     });
 
-    const expectedState = deepFreeze([state.messages[123]]);
+    const expectedState = deepFreeze([state.messages[123], ...state.outbox]);
 
     const anchor = getMessagesForNarrow(state, HOME_NARROW);
 
--- a/src/chat/narrowsSelectors.js
+++ b/src/chat/narrowsSelectors.js
@@ -36,9 +36,6 @@ export const outboxMessagesForNarrow: Selector<Outbox[], Narrow> = createSelecto
   getCaughtUpForNarrow,
   state => getOutbox(state),
   (narrow, caughtUp, outboxMessages) => {
-    if (!caughtUp.newer) {
-      return NULL_ARRAY;
-    }
     const filtered = outboxMessages.filter(item => narrowContains(narrow, item.narrow));
     return isEqual(filtered, outboxMessages) ? outboxMessages : filtered;
   },
--- a/src/topics/__tests__/topicsSelectors-test.js
+++ b/src/topics/__tests__/topicsSelectors-test.js
@@ -30,6 +30,7 @@ describe('getLastMessageTopic', () => {
   test('when no messages in narrow return an empty string', () => {
     const state = deepFreeze({
       narrows: {},
+      outbox: [],
     });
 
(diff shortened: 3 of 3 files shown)


chrisbobbe:
Interesting, thanks for this PR, and the debugging that led to it, @agrawal-d!

Over at https://github.com/zulip/zulip-mobile/issues/3800#issuecomment-652485459, you said:

> I don't get the logic behind the `if` condition I mentioned above - what I understand from it is: 'If we haven't fetched the newest messages, don't show unsent messages'.

> That does not make much sense to me. In fact think there is no need for this condition at all, because whatever may be our perceived knowledge about the fetched messages, the app can still have unsent messages, and they should always be shown.

I traced the origin of this bit of logic to b23628a16, which links to https://github.com/zulip/zulip-mobile/pull/1262. Neither the commit nor the PR offers any explanation at all, which is unfortunate—not only because it doesn't tell us the motivation for this particular change, but because, looking at the commit, it seems plausible that there are numerous *other* changes associated with this change, and those are also unexplained.

These days, we have much stricter standards about what gets committed, and it pays off even if it feels a bit laborious to get things right. 😄

Do the changes in b23628a16 suggest to you that there might be more work to be done in this PR, if the goal is to undo the work of the commit message, "Do not show outbox if not caught up"?

chrisbobbe:
For the reason for b23628a16e88c6ca9e2f996a871f6fb2c197d6be, I have one guess.

You [said](https://github.com/zulip/zulip-mobile/issues/3800#issuecomment-652485459):

> whatever may be our perceived knowledge about the fetched messages, the app can still have unsent messages):

Very true.

Then:

> and they should always be shown.

I mostly agree, but I suppose there's an argument for not showing them *directly following* some messages that we know might be old. While we're loading new messages, we might mistakenly think that the outbox messages are going to be sent directly following some messages that may be old in any of several ways, e.g.:

- more messages arrived after them
- they'd been edited or deleted
- they're just...old. Like a message saying "Shall we meet in 10 minutes?" that was actually sent a day ago.

That being said, if we're still retrying sending some messages, then our latest knowledge of the narrow should be...reasonably up-to-date, I think? The latest knowledge will likely line up with the time you first tried to send the messages, and we shouldn't be constantly retrying for an unreasonable amount of time (that's https://github.com/zulip/zulip-mobile/issues/3829, I think). With this in mind, I think it's probably okay to proceed as we do in this PR. But does that sound reasonable to you?

agrawal-d:
Thanks for the link to the original change. Yikes, the PR gives no explanation at all on why the change was made!


>That being said, if we're still retrying sending some messages, then our latest knowledge of the narrow should be...reasonably up-to-date, I think? The latest knowledge will likely line up with the time you first tried to send the messages. And we shouldn't be constantly retrying for an unreasonable amount of time (that's #3829, I think).

Yeah, I agree, if I responded to message X with a message Y, and Y is still in the outbox, it is very unlikely that there are un-fetched messages between X and Y. An if there is a UX penalty for showing Y, the UX penalty for **not** showing Y is much higher, I think.

And if I am replying offline, then I will be reasonably aware that the messages are not latest. What would be very bad UX is that I sent a message when offline, and it disappeared.

I think both logics have drawbacks, but I feel the current logic has more drawbacks than my proposed one.

chrisbobbe:
Linking to [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3800.20outbox.20messages.20not.20shown/near/926437).

agrawal-d:
Closing as the solution discussed in the link above is very different to the one proposed in this PR.

gnprice:
Thanks @agrawal-d for driving this forward!

If you'd like to tackle giving an error on send in this condition as discussed at that chat link, that would be excellent :slightly_smiling_face:

agrawal-d:
@gnprice sure! Working on it.

gnprice:
(And that's #4208.)

Issue #4156, opened 2020-06-15
Message list loads forever, except when very few messages


pohutukawa:
Unless visiting a stream with only **very few** (up to about a hand-full) messages, the mobile UI never loads the topic/message list. It remains in the stylised "message list loading" type of screen (fancy spinner). To me, this seems possibly related to #4033, but it may be different (more severe?).

In our environment pretty much all users experience this on the mobile Zulip client on both Android as well as iOS. The web and desktop clients all work fully as expected, with a very responsive UI.

Side note: After an issue with the RabbitMQ queue handling on the server (issue https://github.com/zulip/zulip/issues/14862) at the time we have resolved to upgrading the server to the current Git master (beginning of May).

gnprice:
Thanks @pohutukawa for the report! That is a pretty bad experience.

One thing reported on #4033 was this workaround:

> Restarting the app (force quit and reload) fixes it immediately 100% of the time

Does that fix it for you? That's still a bad experience -- maybe *less* terrible but still bad -- but it's potentially an informative clue for us in debugging it.

pohutukawa:
Thanks @gnprice for taking this on. Yes, I have tried the force quit & reload, but it didn't work.

I've also deleted all the cache & data of the app and re-connected to our Zulip server: Same thing, nothing loads.

If I can help with gathering insights, I'm happy to provide them to my means of ability.

pohutukawa:
@gnprice any further input on this? Anything we can do to provide further data to debug the problem?

chrisbobbe:
The loading view you're seeing is this one, right; gray squares and lines covering most of the screen?

![image](https://user-images.githubusercontent.com/22248748/85504072-12060280-b5a0-11ea-8a72-77c7aa5bb8a1.png)

Or is it a circular spinner?

pohutukawa:
@chrisbobbe Yes, it's the "stripey one" in the screen shot above.

pohutukawa:
Ping.
Any help I can provide for further insights on debugging. For quite a while the mobile Zulip client is completely unusable, now. People are starting to scream to go back to Slack again, which I'd like to avoid.

gnprice:
@pohutukawa Thanks for your persistence with this!

I think in order to get helpful logs for debugging this, we'll need to add some new code. I hope to have a beta version out soon with some code for that. If you haven't already, one step you can take now is to join the beta:
https://github.com/zulip/zulip-mobile/#using-the-beta
[...]

PR #4180, opened 2020-07-01
MessageList long loading logs

chrisbobbe:
To help debugging #4156; see also discussion around https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/921700, where we'll have the debugging conversations.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -179,10 +179,24 @@ class MessageList extends Component<Props> {
    * Initiate round-trip handshakes with the WebView, until one succeeds.
    */
   setupSendUpdateEvents = (): void => {
+    const timeAtSetup = Date.now();
+    let attempts: number = 0;
+    let hasLogged: boolean = false;
+
     clearInterval(this.readyRetryInterval);
     this.readyRetryInterval = setInterval(() => {
+      const timeElapsedMs: number = Date.now() - timeAtSetup;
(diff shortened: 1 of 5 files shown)


gnprice:
Thanks @chrisbobbe !

I'm about to merge this, in the interest of getting out a beta today that will help debug #4156. As you say in the last commit message, though, it'd be good to structure that logging code a bit better. A few thoughts:
* The log-long-load state and all the logic referring to that state could be encapsulated in a class, with an instance created at toplevel (where the `setTimeout(maybeLogLongLoad, …)` is now.)
* The logic added to `handleMessageEvent` could then be one line just invoking a method on that instance, just passing `uevent`.
* The `map` callback that produces `loggableEvents` could be pulled out and given a name like `scrubUpdateEvent` -- then `logLongLoad` would be much shorter.
* In that `map` callback, I think it's more robust (even though it's a bit more boilerplate) to have each case name all the properties it's passing through, rather than say `...rest`. In particular this means that if we add a new property to one of these, it doesn't get logged until we explicitly consider how to handle it.
  * A related illustration, perhaps: in the `typing` case I think the `placeholdersDivTagFromContent` will always just return null, because the `typing` content is focused on just the typing-status elements. Probably the boolean `content !== ''` is as much information from that as could be useful, but also there wouldn't be much lost by just omitting it entirely.

chrisbobbe:
Cool, thanks for those suggestions!

Issue #4058, opened 2020-04-20
Give early error when "Zulip server URL" is an email address


gnprice:
We have [an event](https://sentry.io/organizations/zulip/discover/zulip-mobile:fcab692fe3bd4149af994c91ee1c1939/?field=title&field=event.type&field=project&field=count_unique%28user%29&name=All+Events&project=191284&query=&sort=-count_unique_user&statsPeriod=30d&widths=-1&widths=-1&widths=-1&widths=-1&yAxis=count%28%29) in Sentry where we tried to fetch from a URL like `https://example@hotmail.com/api/v1/server_settings?`. (Where I've edited only the username, to `example`.) 

Evidently the user entered an email address at the prompt "Enter your Zulip server URL:". The `https://` prefix would then come from our "smart URL input".

We should detect this and give an appropriate error message before even trying to make the HTTP request. In particular, although `https://jdoe@example.com/` is a well-formed URL, it parses as having a (non-empty) username in it -- which is a rare feature on the Web in general, and which I think there is no use case for with a Zulip server. So it should be enough to parse the URL and look for its [username property](https://url.spec.whatwg.org/#dom-url-username).

First spotted at https://github.com/zulip/zulip-mobile/issues/4050#issuecomment-616754046 .

PR #4181, opened 2020-07-02
realm screen: Give early error when server URL is an email address.

agrawal-d:
Show an error message when the user tries to enter an email in the
server URL input field, even before trying to connect to the
server.

Fixes: #4058

### DIFF
--- a/src/start/RealmScreen.js
+++ b/src/start/RealmScreen.js
@@ -46,7 +46,6 @@ class RealmScreen extends PureComponent<Props, State> {
     const { realm } = this.state;
 
     this.setState({
-      realm,
       progress: true,
       error: null,
     });
--- a/src/start/RealmScreen.js
+++ b/src/start/RealmScreen.js
@@ -7,7 +7,7 @@ import { ZulipVersion } from '../utils/zulipVersion';
 import type { ApiResponseServerSettings, Dispatch } from '../types';
 import { connect } from '../react-redux';
 import { ErrorMsg, Label, SmartUrlInput, Screen, ZulipButton } from '../common';
-import { isValidUrl } from '../utils/url';
+import { isValidUrl, tryParseUrl } from '../utils/url';
 import * as api from '../api';
 import { realmAdd, navigateToAuth } from '../actions';
 
@@ -45,13 +45,21 @@ class RealmScreen extends PureComponent<Props, State> {
   tryRealm = async () => {
     const { realm } = this.state;
 
+    const parsedRealm = tryParseUrl(realm);
+    if (!parsedRealm) {
+      this.setState({ error: 'Please enter a valid URL' });
+      return;
+    }
+    if (parsedRealm.username !== '') {
+      this.setState({ error: 'Please enter the server URL, not your email' });
+      return;
+    }
+
+    const { dispatch } = this.props;
     this.setState({
       progress: true,
       error: null,
     });
-
-    const { dispatch } = this.props;
-
     try {
       const serverSettings: ApiResponseServerSettings = await api.getServerSettings(realm);
       dispatch(realmAdd(realm, new ZulipVersion(serverSettings.zulip_version)));
--- a/src/common/ErrorMsg.js
+++ b/src/common/ErrorMsg.js
@@ -2,13 +2,11 @@
 import React, { PureComponent } from 'react';
 import { StyleSheet, View } from 'react-native';
 
-import { CONTROL_SIZE } from '../styles';
 import Label from './Label';
 
 const styles = StyleSheet.create({
   field: {
     flexDirection: 'row',
-    height: CONTROL_SIZE / 2,
     marginTop: 5,
     marginBottom: 5,
     justifyContent: 'center',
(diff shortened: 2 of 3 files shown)


src/start/RealmScreen.js
```diff
@@ -53,6 +54,15 @@ class RealmScreen extends PureComponent<Props, State> {
 
     const { dispatch } = this.props;
 
+    if (new WhatwgURL(realm).username !== '') {
```
chrisbobbe, line 57: `new WhatwgURL(realm)` will throw an error if `realm` is an invalid URL: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

src/start/RealmScreen.js
```diff
@@ -53,6 +54,15 @@ class RealmScreen extends PureComponent<Props, State> {
...
+      this.setState({
+        progress: false,
+        error:
+          'Please enter the server URL, not your email. You can enter your ceredentials in the next screen.',
```
chrisbobbe, line 61: Nit: "credentials" (also in `messages_en.json`)

src/start/RealmScreen.js
chrisbobbe: I just tested this by hardcoding something wrong for 'realm'. I got this:

```
Possible Unhandled Promise Rejection (id: 0):
TypeError: Invalid URL: asdf
```

I don't get a white-screen crash (the error was thrown from a React event handler), but the spinner does go on indefinitely. I think we can isolate `new WhatwgUrl(realm)` into its own try/catch block, and, on this error, do a `setState` similar to this one, with an appropriate error message saying it's an invalid URL.

Then the email can be checked below that.

agrawal-d: Ah! Initially, I did add a try-catch for the URL constructor. But then I saw that we already validate the URL in the login button using a regular expression - but I never considered that the form could be submitted by simply tapping the enter key from the keyboard, without needing to touch the button.

agrawal-d: Okay, I just tested this, and I *cant* submit the form if the URL is incorrect - by tapping enter or by tapping the send button.

agrawal-d:
Thanks for the review. I've made the suggested changes.

src/start/RealmScreen.js
```diff
@@ -53,6 +54,26 @@ class RealmScreen extends PureComponent<Props, State> {
 
     const { dispatch } = this.props;
 
+    let parsedRealm;
+    try {
+      parsedRealm = new WhatwgURL(realm);
+    } catch (err) {
```
gnprice, line 60: This bit can be simplified with `tryParseUrl` (which I added last week for a very similar pattern in #4034 :slightly_smiling_face: )

src/start/RealmScreen.js
```diff
@@ -53,6 +54,26 @@ class RealmScreen extends PureComponent<Props, State> {
...
+        progress: false,
+        error:
+          'Please enter the server URL, not your email. You can enter your credentials in the next screen.',
+      });
+      return;
+    }
+
     try {
       const serverSettings: ApiResponseServerSettings = await api.getServerSettings(realm);
       dispatch(realmAdd(realm, new ZulipVersion(serverSettings.zulip_version)));
```
gnprice, line 79: This function gets somewhat difficult to follow with all the stuff going on in it now. I think one reason is that the `setState` at the top is setting up the start of a story you basically have to read the whole function to see the different endings of.

One thing that would simplify it is to not set `progress: true` until after the realm-parsing is done and those errors haven't happened. That way those don't need to say `progress: false`, and the whole question of the `progress` state is limited to just where we're actually contacting the server, which is what it's for.

Relatedly, we can cut the line already that sets `realm` in that first `setState`. It's a no-op because we just read `realm` *from* the state. Looks like it's a legacy of an unfinished refactor in 252fed22fa1432e4cf6469423099521ecc3ed7ab -- before then, this line actually set `realm` to a different value than it already had, and so was doing something useful.

Similarly to moving the `setState` but smaller: the `const { dispatch } = …` line can wait until just above the `try`.

src/start/RealmScreen.js
```diff
@@ -53,6 +54,26 @@ class RealmScreen extends PureComponent<Props, State> {
...
+        error: 'Please enter a valid URL',
+      });
+      return;
+    }
+
+    if (parsedRealm.username !== '') {
+      this.setState({
+        progress: false,
+        error:
+          'Please enter the server URL, not your email. You can enter your credentials in the next screen.',
```
gnprice, line 72: I think most people don't know what "credentials" means in this context. Someone who enters an email address at the "Zulip server URL" prompt is especially unlikely to know what it means :wink:

So let's just drop that sentence.

(It's also not quite true -- often the next screen will be a menu of options for how to log in. Then the *next* screen after that might involve entering credentials, depending on the auth method chosen.)

src/start/RealmScreen.js
agrawal-d: Okay, also, I moved the `error:null` to just above the `try` block too.

agrawal-d:
Thanks for the review, @gnprice. I've pushed some changes.

gnprice:
Thanks @agrawal-d ! Merged.

I made one change first, inserting a commit to remove `realm` from the `setState`. The reason for that isn't really explained by the commit message for the main change, so it belongs in its own little commit in order to get its own commit message.

Issue #4174, opened 2020-06-30
Migrate to GitHub Actions (from Travis CI)


agrawal-d:
GitHub Actions is a new product by GitHub that we can use to automate our CI workflow. I've used it on some of my own projects, and it's pretty good. It runs fast and the dependency downloads are not flaky like in Travis CI. 

We should try out GitHub actions to evaluate it.

Documentation: https://help.github.com/en/actions

chrisbobbe:
I seem to remember some discussion about CI...just not being run anymore, in this repo. Is that what led you to consider GitHub Actions? Do you remember where that discussion was, or am I imagining things?

agrawal-d:
> . Is that one thing that led you to consider GitHub Actions?

Partly yes, and partly because it's integrated into the GitHub UI, so we won't have to switch between services.

>Do you remember where that discussion was, or am I imagining things

:smile: Yes, there was a discussion some time ago, as a comment on a PR: https://github.com/zulip/zulip-mobile/pull/4143#issuecomment-643004206

gnprice:
Turns out:
* Our CI was *running* just fine all along -- but failing to post results to the PRs;
* I was able to [fix that by just](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/travis.20results/near/933188) revoking Travis's access to my account and authorizing it again.

So we're now back to Travis working just as well as it always did.

Faster and less flaky would still be quite attractive, though! So this remains a useful thing to try.

gnprice:
I noticed on #4250 that Travis was back to not posting results on the PRs.

Looking again at [the build history](https://travis-ci.org/github/zulip/zulip-mobile/pull_requests) on Travis's own site, it's still the case that it is running the builds, just apparently failing to post results. But it's really not very useful when it doesn't post results.

I could repeat the "turn it off and on again" fix that seemed to work in July, and maybe I will. But if I have to do that every couple of months when we notice it's broken, that will get old real fast -- really it already has. So we should probably consider Travis to be on its last legs.

gnprice:
I'd noticed again recently that some PRs weren't getting CI results posted from Travis. Today @hackerkid [discovered](https://chat.zulip.org/#narrow/stream/43-automated-testing/topic/travis.20ci/near/1077862) that this is true across the Zulip project, and that the reason is that Travis CI has recently cut off free service for most open-source work, including Zulip.

[...]

PR #4183, opened 2020-07-07
Use absolute URL for webview `baseUrl` on iOS.

chrisbobbe:
Following discussion at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/924227.

~~This should be NFC on Android, but I haven't tested—so I'm marking this as a draft until I do so.~~ Done!

I've tried to be faithful to the discussions around the labels we choose, etc., at 28bd508ea and at links from there. But I may have missed something.


src/webview/MessageList.js
```diff
@@ -309,8 +285,16 @@ class MessageList extends Component<Props> {
...
-        originWhitelist={['file://']}
+        source={{ baseUrl: baseUrl.toString(), html }}
+        originWhitelist={[
+          // `baseUrl.origin` may be "null" (an "opaque origin") for
+          // `file://` URLs. [1] [2] `react-native-webview` doesn't
+          // know about this.
+          //
+          // [1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/926650
+          // [2] https://github.com/jsdom/whatwg-url/tree/v7.0.0/#specification-conformance
+          baseUrl.origin === new WhatwgURL('file:///').origin ? 'file://' : baseUrl.origin,
```
chrisbobbe, line 296: Oh, right—in this revision, I've acted on what I said [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/926696) and abandoned ever pointing to what's served by the packager.

In effect, this means the URL should always start with `file://`—so we could just say "file://", if that's preferred.

gnprice: I was just about to ask what this code was doing and why this part became more complicated :slightly_smiling_face: I guess that's the answer!

Yeah, let's leave this in the simple form it's in until/unless we find a need to do something more complex.

src/webview/MessageList.js
```diff
@@ -261,23 +245,15 @@ class MessageList extends Component<Props> {
...
-              url.startsWith('file://')
-              && url.endsWith(baseUrlTail)
-              && !unsafeUrlRegex.test(url)
-            ) {
-              loaded_once = true;
-              return true;
-            }
+          const parsedUrl: WhatwgURL | void = tryParseUrl(url);
+          if (
+            loaded_once === false
```
gnprice, line 251: This is already boolean :wink: -- so just write the boolean operation `!loaded_once`

src/webview/MessageList.js
```diff
@@ -261,23 +245,15 @@ class MessageList extends Component<Props> {
...
+            && parsedUrl !== undefined
```
gnprice, line 252: This can be just `parsedUrl`.

src/webview/MessageList.js
```diff
@@ -261,23 +245,15 @@ class MessageList extends Component<Props> {
...
-            // RN provides no good interface to. (`react-native-fs` is awful.)
-            if (
-              url.startsWith('file://')
-              && url.endsWith(baseUrlTail)
-              && !unsafeUrlRegex.test(url)
-            ) {
-              loaded_once = true;
-              return true;
-            }
+          const parsedUrl: WhatwgURL | void = tryParseUrl(url);
```
gnprice, line 249: This annotation isn't needed, is it? It's just the plain return type of `tryParseUrl`.

src/webview/MessageList.js
```diff
@@ -240,12 +201,35 @@ class MessageList extends Component<Props> {
...
+     *   so we do.
+     *
+     * [1] Oddly, this essential feature doesn't seem to be documented!  It's
+     *     widely described in how-tos across the web and StackOverflow answers.
+     *     It's assumed in some related docs which mention it in passing, and
+     *     treated matter-of-factly in some Chromium bug threads.  Details at:
+     *     https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/android.20filesystem/near/796440
      */
-    const baseUrl = `${assetsPath}/index.html`;
+    const baseUrl: WhatwgURL =
```
gnprice, line 229: This annotation is surely not needed :slightly_smiling_face: -- we're initializing it with `new WhatwgURL(…)`

src/webview/MessageList.js
```diff
@@ -309,8 +285,16 @@ class MessageList extends Component<Props> {
     // https://github.com/react-native-community/react-native-webview/pull/697
     return (
       <WebView
-        source={{ baseUrl, html }}
-        originWhitelist={['file://']}
+        source={{ baseUrl: baseUrl.toString(), html }}
```
gnprice, line 288: Hmm, I get a Flow error on this line!

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/webview/MessageList.js:288:17

Could not decide which case to select. Since case 1 [1] may work but if it doesn't
case 2 [1] looks promising too. To fix add a type annotation to call of method
toString [2].

     src/webview/MessageList.js
     285│     // https://github.com/react-native-community/react-native-webview/pull/697
     286│     return (
     287│       <WebView
 [2] 288│         source={{ baseUrl: baseUrl.toString(), html }}
     289│         originWhitelist={[
     290│           // `baseUrl.origin` may be "null" (an "opaque origin") for
     291│           // `file://` URLs. [1] [2] `react-native-webview` doesn't

     flow-typed/react-native-webview_v8.x.x.js
 [1] 495│     source?: WebViewSource,



Found 1 error
```

Puzzling that I see that even though it seems to have passed in CI. If you're not seeing the error, let's debug in chat.

chrisbobbe: > it seems to have passed in CI

Ah, I think CI isn't a thing anymore, hence https://github.com/zulip/zulip-mobile/issues/4174?

chrisbobbe:
Just pushed my changes, including some more, discussed [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Resource.20URLs/near/927320) and following, so, ready for another review. 🙂

chrisbobbe:
Just tweaked the 'react-native' mocking commit. FYI, that commit is cherry-picked (with some changes to the commit message) into my current revision of the RN v0.61 upgrade (https://github.com/zulip/zulip-mobile/pull/4151). Its appearance in both PRs is reviewable, but when one PR gets merged, we should remove the commit from the other PR branch. 🙂

src/webview/MessageList.js
gnprice: Ah indeed :slightly_smiling_face: (and fixed now.)

gnprice: I see the fix was:
```
        source={{ baseUrl: (baseUrl.toString(): string), html }}
```
i.e. adding an explicit cast to `string`. Weird that that makes a difference -- this method's return type is shown as simply `string` already. Seems like a Flow bug.

But the behavior with that workaround seems like the right behavior, so :shrug:

src/webview/MessageList.js
```diff
@@ -113,45 +115,33 @@ export type Props = $ReadOnly<{|
 |}>;
 
 /**
- * A URL-like magic string denoting the webview-assets folder.
+ * The URL of the platform-specific assets folder.
  *
- * Not suitable for reuse or export.
+ * - On iOS: We can't easily hardcode this because it includes UUIDs.
+ *   So we bring it over the React Native bridge in ZLPConstants.m.
+ * - On Android: Different apps' WebViews see different (virtual) root
```
gnprice, line 122:
```suggestion
 * - On iOS: We can't easily hardcode this because it includes UUIDs.
 *   So we bring it over the React Native bridge in ZLPConstants.m.
 *
 * - On Android: Different apps' WebViews see different (virtual) root
```

src/webview/MessageList.js
```diff
@@ -113,45 +115,33 @@ export type Props = $ReadOnly<{|
...
+const assetsUrl =
+  Platform.OS === 'ios'
+    ? new WhatwgURL(NativeModules.ZLPConstants.resourceURL)
+    : new WhatwgURL('file:///android_asset/');
+
+/**
+ * The URL of the webview-assets folder.
+ *
+ * See `check_path_name` in `tools/build-webview` for information on
+ * what folder this is.
```
gnprice, line 142: This is implementation details, not interface, so doesn't belong in jsdoc.

gnprice: And thinking a bit more about this, the real value of this cross-reference is about finding what files go into this directory, not finding where it lives in the build tree. (The existing comment isn't very specific about this distinction.) So instead of mentioning `check_path_name`, maybe:
```suggestion
 *
 * This is the folder populated at build time by `tools/build-webview`.
```
(and jsdoc is a fine place for that.)

chrisbobbe:
Thanks for the review! Just pushed my changes.

chrisbobbe:
Just resolved that conflict.

chrisbobbe:
From your most recent review, at https://github.com/zulip/zulip-mobile/pull/4183#pullrequestreview-447627190:

> The one substantive thing this needs is a rebase across the Jest mocking changes (as revised and merged via #4151). See also small comments below -- then this will be ready to merge!

I've done that, and those comments were indeed small (and I believe I've addressed them), so I'll go ahead and merge this. 🙂

Issue #3950, opened 2020-03-04
Use Map rather than object-as-Map in our Redux state


gnprice:
This is an optimization, and a code cleanup, that will also help us toward #3949.

Our Redux state tree is currently made up entirely of primitive values, plain arrays, and plain objects. These are very nice for struct-like data structures like `User` or `Message`. They're not so nice for key-value maps: we end up using objects as maps, and that's messier in the code and for type-checking than a proper `Map`. It also turns out to be about 2x slower for building them up and making new ones, as measured in https://github.com/zulip/zulip-mobile/issues/3339#issuecomment-462963242 .

So, where we use an object as a map, we should use a real `Map` instead. (Better, perhaps, an `Immutable.Map` -- that's #3949. But this will be a step on the way.)

We've already converted lots of objects-as-maps to real `Map`s in much of our code, notably in the return values of caching selectors. The reason we haven't yet done so at all in our Redux state tree is a technical obstacle that shouldn't fundamentally be complicated once we work out how, but will require some working out: we'll need to serialize and deserialize these data structures for `redux-persist`.

Specifically we'll need to study the docs of `redux-persist` and friends, then possibly their implementations where docs are inadequate, in order to see how to wire up appropriate serialization and deserialization functions.

Relatedly, we'll want to think through and test the migration path: what happens when a version of the app using a new data structure (for a given spot in the Redux state tree) first launches and there's existing data serialized by a version of the app that used the old data structure.

Some quick notes (adapted from notes I made after some reading in 2019-04) on possible tools for the persistence side:
* The redux-persist feature we want is "transforms": https://github.com/rt2zz/redux-persist#transforms
* In particular `remotedev-serialize` looks like a pretty good-quality implementation of serializing things like `Map` and `Set`. It'll also support Immutable.js data structures if we go on to adopt those in the future (#3949). The implementation is here:
  https://github.com/kolodny/jsan/blob/master/lib/cycle.js
  and enabled here:
  https://github.com/zalmoxisus/remotedev-serialize/blob/master/constants/options.js

The migration will happen one data structure at a time; each one is pretty much independent of others. Most of the work will probably come as part of doing the first one, though -- figuring out just how to handle persistence and any migration, and getting that set up.

gnprice:
I think this is now subsumed in #3949.

[...]

Issue #3949, opened 2020-03-03
Use better data structures in our Redux state


gnprice:
As discussed in #3339, there are a lot of places where we maintain large amounts of data -- like all the users in the realm -- in data structures like `Array` which make it inefficient to find the data we want at a given time, like a particular user.

Some other data structures are maintained as objects used as maps: notably, the collection of all the messages we have from the server. These are efficient for lookup... but both these objects-as-maps, and the `Array`s, are extremely inefficient to build up for large amounts of data in the Redux style we use. Specifically, building a new array like `[...state, newItem]` or a new object like `{ ...state, [newKey]: newItem }` has to copy the entire existing array or object, taking linear time -- which means building up an array or object with N items this way takes *quadratic* time O(N^2). Demo at https://github.com/zulip/zulip-mobile/issues/3339#issuecomment-462963242 .

So, we should fix that.

There are a couple of potential stages here:
* As a modest but useful improvement, we can use `Map` instead of object-as-map. This is still quadratic time to build, but over 2x faster (https://github.com/zulip/zulip-mobile/issues/3339#issuecomment-462963242), a solid constant-factor improvement. It's also cleaner in the code and for type-checking.
* For a more complete solution, we'd want to be maintaining our data structures in an efficient way. Efficiently maintaining data structures without mutating them is a bit subtle, but there are well-understood techniques for doing so (called "persistent data structures"). One implementation of those in JS is [Immutable.js](https://github.com/facebook/immutable-js/).

---

The main technical obstacle I believe we'll need to resolve is how to serialize and deserialize these data structures for `redux-persist`. This shouldn't fundamentally be complicated once we work out how; but it'll require some studying of the docs of `redux-persist` and friends and of Immutable.js, then possibly of implementations where docs are inadequate, in order to see how to wire up appropriate serialization and deserialization functions.

Relatedly, we'll want to think through and test the migration path: what happens when a version of the app using a new data structure (for a given spot in the Redux state tree) first launches and there's existing data serialized by a version of the app that used the old data structure.

One reason to start with converting some existing object-as-map into a `Map` is that it would give us an opportunity to figure out that persistence side of things without yet introducing any of the more interesting changes to our code that Immutable.js may bring.

Some quick notes (adapted from notes I made after some reading in 2019-04) on possible tools for the persistence side:
* The redux-persist feature we want is "transforms": https://github.com/rt2zz/redux-persist#transforms
[...]

PR #4187, opened 2020-07-10
Make `state.narrows` an Immutable.Map.

chrisbobbe:
An instance of #3949 and #3950, this is planned followup to #4047. See https://github.com/zulip/zulip-mobile/pull/4047#issuecomment-616877699.

~~Still getting all these tests type-checked, but thought I'd push my work today.~~ 🙂

Done! Well, with a few things to note:

- `src/message/__tests__/fetchActions-test.js` isn't type-checked and made to use example data in this PR. Those are done in my current branch for #4171, which is open. No need to cherry-pick that work, I think; that'll just make things more confusing. I don't think this has a strong influence on which PR we should aim to merge first: I just want to generally be sure that the test files I touch will be type-checked at some point.

- I did cherry-pick a few commits from my branch for #4171, only in support of type-checking other test files for the first time.

- ~~I also cherry-picked one commit from #3946, which is also open.~~ #3946 was recently merged, so no need—just rebased across this. 🙂 

I've called out all of these cherry-picks with a note at the bottom of the commit message.


chrisbobbe:
Just resolved some conflicts; this is ready for review.

src/chat/__tests__/narrowsSelectors-test.js
```diff
@@ -120,11 +114,9 @@ describe('getMessagesForNarrow', () => {
       ],
     });
 
-    const anchor = getMessagesForNarrow(state, privateNarrow('john@example.com'));
-
-    const expectedState = deepFreeze([{ id: 123 }]);
```
gnprice: Much better -- thanks for cleaning these up :slightly_smiling_face: 

I think with these names you're looking at the results of two different cases of someone copy-pasting without thinking it through (or perhaps not understanding.) The `expectedState` is, as you say, a reasonable name in many reducer tests.

The `anchor`... I suspect comes from the last few test cases below, testing things like `getFirstMessageId`. There, the result is at least plausibly an "anchor" -- it's a message ID that we're going to use to identify a spot in the message stream. And then those got copied elsewhere, by a contributor who didn't understand what it meant and in a period where we weren't doing meaningful code review.

OK, and having said that I was curious to find out. :slightly_smiling_face: Indeed -- and another layer is that those tests for `getFirstMessageId` etc. were originally for a function called... `getAnchor`. a89d897c2 changed that. And before then, 8dd208120 propagated the `anchor` name to these other test cases.

src/pm-conversations/__tests__/pmConversationsSelectors-test.js
```diff
@@ -23,103 +27,117 @@ describe('getRecentConversations', () => {
...
+      display_recipient: [eg.displayRecipientFromUser(eg.selfUser)],
+    });
+
+    const meJohnAndMarkPm = eg.pmMessage({
+      id: 0,
+      display_recipient: [
+        eg.displayRecipientFromUser(eg.selfUser),
+        eg.displayRecipientFromUser(userMark),
+        eg.displayRecipientFromUser(userJohn),
+      ],
```
gnprice, line 67: These are probably a sign that it'd be helpful to refactor `eg.displayRecipientFromUser` to be called, say, `eg.displayRecipientFromUsers` -- and to take `User[]` and return `PmRecipientUser[]`, and be just a `map` around what it currently is.

(Maybe even both versions should exist.)

gnprice:
... Err, I guess that came out slightly wrong, in that this PR is marked as merged but with that headline final commit dropped, oops. Please go ahead and open a new PR with (a rebased version of) 1a80243f6.

The set of commits merged to master were exactly what I intended, though :slightly_smiling_face:

chrisbobbe:
OK, opened as #4201!

Issue #4188, opened 2020-07-10
Hide user emails in message reaction list according to email_address_visibility setting.


sahil839:
Email of users in the message rection list (which is opened by long-press on an existing reaction or by clicking on "See who reacted" from message action sheet) should be hidden when the original email is not visible to the user according to the email_address_visibility settings of the organization.

There is no use of showing the emails generated by user-id (of the form user1888@example.com), as it does not provide any information about the user.

gnprice:
Thanks @sahil839 !

In fact I think the best thing to do is to go farther than that: we should just leave out the email addresses on this screen unconditionally. Even when real email addresses are available, in this context they feel mostly like clutter.

The code that implements this screen is in `src/reactions/MessageReactionList.js`.

PR #4189, opened 2020-07-13
message reaction list: Hide emails from user list.

agrawal-d:
image

Removes user emails from the user list in the message reactions
screen because in this context, the emails mostly feel like
clutter.

As a side effect of this change, we no longer have to worry about
the server 'email_address_visibility' setting.

Fixes: #4188.


chrisbobbe:
Merged, thanks! I made two small tweaks to the commit message:

- Removed the period on 'Fixes: #4188.' to keep with our [style](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md) for this line.
- Changed "Removes" to "Remove".

PR #4190, opened 2020-07-13
msg action sheet: Add 'Reply with mention' option.

agrawal-d:
Adds the reply with mention featrue, that populates the compose box
message input with an @-mention, the topic input (if present) with
the topic of the message and shows the keyboard.

Fixes: #3436.


agrawal-d:
>Would you mind rebasing and see if you also get a Flow error?

Yep, I got the Flow error. I've fixed it now.

PR #4191, opened 2020-07-13
apiTypes [nfc]: Cut some require cycles.

chrisbobbe:
Unfortunately, we'll need to do more—but this is a fairly straightforward bit of progress I think we've landed on.

From [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3745.20Server.20version.20to.20Sentry/near/930324).

### DIFF
--- a/src/api/apiTypes.js
+++ b/src/api/apiTypes.js
@@ -4,5 +4,3 @@ export type * from './transportTypes';
 export type * from './modelTypes';
 export type * from './eventTypes';
 export type * from './initialDataTypes';
-
-export type * from './settings/getServerSettings';
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -1,14 +1,7 @@
 /* @flow strict-local */
 import NavigationService from '../nav/NavigationService';
-import type {
-  Narrow,
-  Dispatch,
-  GetState,
-  GlobalState,
-  Message,
-  Action,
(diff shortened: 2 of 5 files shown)


chrisbobbe:
Heh, actually, debatable whether this is NFC—as it's meant to fix a bug.

gnprice:
This change looks good, thanks!

Just one quibble, in the commit message: this doesn't actually cut many cycles. We go from 18 cycles reported by `tsort`, down to 17:
```
$ tools/tsort-modules pairs | tsort 2>&1 >/dev/null | grep -c 'contains a loop'
17
```

Or in terms of whole cycle-blobs (the technical term is "strongly-connected components" or SCCs), it actually increases the number of them... by breaking up the biggest of them, with 2 of its 12 modules moving off into their own tiny cycle-blob and 3 others getting freed entirely, leaving 7 behind.

This is still quite helpful, though! For basically the reason described in the body of the commit message: it prevents an anticipated future import from creating a much larger import cycle-blob, and triggering that Flow bug. Details [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/import.20cycles/near/1057133).

chrisbobbe:
Thanks for the review and analysis! I just pushed some tweaks to the commit message; please take a look.

gnprice:
Thanks! Merged.

PR #4195, opened 2020-07-17
Improve long-loading logging code from the WebView.

chrisbobbe:
Prompted by Greg's suggestions at https://github.com/zulip/zulip-mobile/pull/4180#issuecomment-654466175.

There are a few differences to call out. If they're problematic, I'll fix them right away, of course!

> The log-long-load state and all the logic referring to that state could be encapsulated in a class

Since I was making a class anyway, and putting it in its own JS file, I didn't want to entangle that class with the DOM detail of whether the `message-loading` div was visible. Instead, I made it so you can call "start" and "stop" methods on the class. I think this will also be useful if we want to log based on other, unrelated states. The code added in `js.js` is still much smaller than in the old way.

> The `map` callback that produces `loggableEvents` could be pulled out and given a name like `scrubUpdateEvent` -- then `logLongLoad` would be much shorter.

Makes sense. I've pulled out the "scrubbing" logic into its own function, but the difference is: I don't use it in a `map`. As soon as the class instance becomes aware of the event, it scrubs it right away, before pushing it to the array.

### DIFF
--- a/src/webview/js/InboundEventLogger.js
+++ b/src/webview/js/InboundEventLogger.js
@@ -1,0 +1,159 @@
+/* @flow strict-local */
+import type {
+  WebViewUpdateEvent as WebViewInboundEvent,
+  WebViewUpdateEventContent as WebViewInboundEventContent,
+  WebViewUpdateEventFetching as WebViewInboundEventFetching,
+  WebViewUpdateEventTyping as WebViewInboundEventTyping,
+  WebViewUpdateEventReady as WebViewInboundEventReady,
+  WebViewUpdateEventMessagesRead as WebViewInboundEventMessagesRead,
+} from '../webViewHandleUpdates';
+import type { JSONable } from '../../utils/jsonable';
+import sendMessage from './sendMessage';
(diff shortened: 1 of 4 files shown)


gnprice:
Looks good! Two comments:
* When at the 10-second mark we decide not to send, we should be sure to clear the array of log/event items. That way we don't hang on to those objects indefinitely when they'll never be used again.
* The main disadvantage of scrubbing up front vs. scrubbing before send is that most of the time we don't end up logging anything, and in those cases it's a waste of work to bother scrubbing. But on the other hand we're not doing anything particularly expensive in this scrubbing -- it's not like we're searching through a big string or big object tree, we're just picking out some specific properties and in one case searching for one regexp -- so that's not a big deal. And it's also potentially nice to not hang onto the entire event with the entire content string -- in particular if there are a number of updates and so a number of copies of that string that we'd be keeping around until we decide whether to log. So it's likely this way is best in any case.

chrisbobbe:
Sounds good! I just pushed my latest changse.

gnprice:
Merged -- thanks!

PR #4196, opened 2020-07-20
[android] Handle receiving shares from other apps 

agrawal-d:
Fixes: #117

Continuation of work done in #4124

Resolves several issues pointed out by @gnprice in https://github.com/zulip/zulip-mobile/pull/4124.


src/nav/navReducer.js
```diff
@@ -64,6 +64,11 @@ const rehydrate = (state, action) => {
     return getStateForRoute('loading');
   }
 
+  // Dont switch to main UI if sharing screen is on.
+  if (state.routes.find(route => route.routeName === 'sharing')) {
+    return state;
+  }
+
```
gnprice, line 71: In the chat thread you started for this question, I think I figured out the mystery of what's happening here (and why notifications don't have the same trouble), and [found](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/need.20for.20conditional.20in.20navReducer's.20.60rehydrate.60.20for.20Sharing/near/945368) a cleaner solution. (And more complete! This one still leaves it the case that the sharing UI will briefly try to render in a situation where the state it has is bogus.)

android/app/src/main/java/com/zulipmobile/sharing/SharingModule.kt
```diff
@@ -11,7 +12,11 @@ internal class SharingModule(reactContext: ReactApplicationContext)
 
     @ReactMethod
     fun getInitialSharedContent(promise: Promise) {
-        promise.resolve(initialSharedData)
+        try{
+            promise.resolve(initialSharedData)
+        }catch(e:Exception){
```
gnprice, line 17: nit: spaces before `{` and after `}`

android/app/src/main/java/com/zulipmobile/sharing/SharingModule.kt
```diff
@@ -11,7 +12,11 @@ internal class SharingModule(reactContext: ReactApplicationContext)
...
+            promise.resolve(null)
+        }
```
gnprice, line 19: Hmmm.

I'm skeptical when I see a pattern like this -- it looks like trying to shut up an error without understanding it.

In particular this description in the commit message seems like a contradiction:

> Catching this exception and returning null fixes this error. This
> does not change the behaviour of this function because the error
> only happens when the map is re-read - and we never actually re-read
> the initialSharedData map, we only ever read it once.

Doesn't the fact that we're seeing the error appear indicate that we do actually re-read the Map in question?

Or another way of looking at basically the same point: if the behavior of the function isn't affected by this change, then why are we making the change? It's certainly not simplifying or clarifying the code. If this change causes those red-screen errors with "Map already consumed" to stop appearing, then that's very much a change in the behavior of the function.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -43,6 +43,11 @@ open class MainActivity : ReactActivity() {
     }
 
     private fun handleSend(intent: Intent) {
+        // Intent is reused after quitting, skip it.
+        if((getIntent().flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY)!=0){
+            return;
+        }
```
gnprice, line 49: Huh, that is *wacky* that it behaves that way! Good find.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -43,6 +43,11 @@ open class MainActivity : ReactActivity() {
     }
 
     private fun handleSend(intent: Intent) {
+        // Intent is reused after quitting, skip it.
+        if((getIntent().flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY)!=0){
```
gnprice, line 47: nit: spaces -- see the style of the existing code

(In Android Studio you should be able to hit Ctrl+Alt+L and have it reformat the code for you. In steady state that should have no effect.)

gnprice:
Observations from playing with the UI:

* I left the Zulip app in the background, went and opened another app, and hit "share" from there and chose Zulip.
* The sharing screen came up. A second or two later, I saw this happen again (like at https://github.com/zulip/zulip-mobile/pull/4124#pullrequestreview-451025610):
  > Then a moment later it flickered and sort of redrew. Not sure what was happening underneath.

  Specifically, now that I've seen this a few times: it looked like a new copy of the sharing screen was appearing and animating in, in a way that I associate with navigating to a screen. So it was as if I was on the sharing screen and then navigating to another copy of the sharing screen.
* I chose PMs, chose a recipient, and sent.
* This time, the app didn't close! Instead it took me back to the screen I'd been on (a particular stream narrow) when I put the app in the background in the first place. The toasts appeared in the usual sequence, over that screen.
  * Seems like ideally we would navigate to the conversation I shared to, instead.
* I hit the home button (this is Android 8 O, on an emulated Pixel 2) to get back to the launcher, and then hit the "recent apps" button and chose Zulip.
  * The app came back to foreground showing the same screen it was on! So, that's good.
* Then I hit the home button again, and this time hit the app's icon in the launcher.
  * ... This time, it came back to the foreground showing that same screen... but then a fraction of a second later, took me back to the main "unreads" screen. Hmm.
  * I tried a few times repeating this. It reproes 100% of the time, the first time I use the launcher icon after having shared something. (If I then use the launcher icon again, it behaves normally. And I can use the recent-apps screen instead of the launcher any number of times, and it behaves normally.)
  * This calls for some further debugging, I think. What's causing this navigation back to the main screen?

src/sharing/ShareToPm.js
```diff
@@ -169,7 +153,7 @@ class ShareToPm extends React.Component<Props, State> {
 
     if (choosingRecipients) {
       return (
-        <Modal>
+        <Modal onRequestClose={this.handleModalClose}>
```
gnprice, line 156: sharing: Handle hardware back button when selecting PM recipients.
    
    Use modal's 'onRequestClose' to hide the user selection screen when
    the back button is pressed while choosing recipients in the share to
    PM tab.

Hmm, did this not work in the previous version? Certainly good to fix so that this works.

I'm less happy about relying completely on the back button here. It's acceptable as a temporary workaround because we only support this flow in the first place on Android -- but that won't work when in the future we offer sharing on iOS as well. And in the UX of this even on Android: everywhere else in the app (including other uses of this user-picker widget!), when it's possible and natural to go back, we have a "back" icon at the top left of the screen that offers that. It feels like a glitch for that to be missing here.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
agrawal-d: Took me several hours to figure this out, I tried a lot of different things 😰

gnprice:
Just merged 5 of these commits!
6bccf4c68 sharing: Switch to 'react-navigation-tabs' for creating tab bars.
38aa9927f nav reducer tests: Use an accurate initial state.
63fee0f7c sharing: De-duplicate the 'handleSend()' function.
a91424121 android: Handle new intents in 'onNewIntent' when activity is resumed.
6ca95f13c sharing: Handle hardware back button when selecting PM recipients.

Looking closer at the rest...

gnprice:
OK, after some investigation [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/need.20for.20conditional.20in.20navReducer's.20.60rehydrate.60.20for.20Sharing/near/947450), ended up sending #4204 which I think will resolve at least a bit of the oddness that we've both been seeing on versions of this branch. Want to try it on top of that and see?

For anything that that branch does fix, another way to avoid the issue is to not rely on `react-native run-android` to start the app. Instead, perhaps after using that command to build and install the app: quit it, and start it from the launcher (or don't start it at all, to see how the sharing feature behaves when the app isn't already running.)

agrawal-d:
Thanks, @gnprice!

I posted a comment on your PR thread: https://github.com/zulip/zulip-mobile/pull/4204#issuecomment-662295317

gnprice:
(Rebased, after having merged #4204 .)

gnprice:
I reproduce the issue you describe at that comment. Let's debug [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/double.20initialization.20on.20sharing/near/948765).

gnprice:
OK, merged three more commits:
742415123 sharing: Involve a thunk action in navigateToSharing so it's buffered.
70e9ad9d4 android: Skip sharing intent when it's reused after quitting.
21a6713fb modal search nav bar: Accept prop `canGoBack`.

(One of them comes from squashing two of the commits from the branch.)

That leaves just two others outstanding:
1ace60ad2 android: Enable receiving 'shares' from other apps.
6ac549cc3 sharing: Reset 'initialSharedData' after sending in SharingModule.

I'd really like to fix the double-initialization issue before turning this on for everyone -- i.e. the issue where we can end up with two different Android activities, and two different React component trees, in the same process and talking to the same JS engine and the same Redux store. Made some debugging progress in that chat thread linked above, and suggested a possible solution (but you should confirm whether it seems to actually work :wink: ).

android/app/src/main/java/com/zulipmobile/sharing/SharingModule.kt
```diff
@@ -1,6 +1,7 @@
 package com.zulipmobile.sharing
 
 import com.facebook.react.bridge.*
+import java.lang.Exception
```
gnprice, line 4: now redundant

android/app/src/main/java/com/zulipmobile/sharing/SharingModule.kt
```diff
@@ -12,6 +13,7 @@ internal class SharingModule(reactContext: ReactApplicationContext)
     @ReactMethod
     fun getInitialSharedContent(promise: Promise) {
         promise.resolve(initialSharedData)
+        initialSharedData = null
     }
```
gnprice, line 17: This is certainly plausible code now. I wonder if it's clearly the behavior we really want, though. Why are we ending up trying twice to get the initial shared data? What's the situation where that happens?

agrawal-d:
Thanks, @gnprice! I've updated the PR.

gnprice:
This one looks like some kind of rebase error, perhaps -- the description and contents don't match:
```
commit 4394d6c6d745610c1bb79fb86fe844aedbfa7885
Author: Divyanshu Agrawal <agrawal.divyanshu@outlook.com>
Date:   Mon Jul 20 12:39:33 2020 +0530

    sharing: Switch to 'react-navigation-tabs' for creating tab bars.
    
    Since the RN 0.60 upgrade, using `createMaterialTopTabNavigator`
    exported by 'react-navigation' causes an error - that
    'ViewPagerAndroid has been removed from React Native'. Switching to
    the one exported by 'react-navigation-tabs' fixes this.
---
 src/sharing/SharingScreen.js | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/src/sharing/SharingScreen.js b/src/sharing/SharingScreen.js
index e69303d76..12a74206c 100644
--- a/src/sharing/SharingScreen.js
+++ b/src/sharing/SharingScreen.js
@@ -71,7 +71,12 @@ class SharingScreen extends PureComponent<Props> {
     }
 
     return (
-      <Screen canGoBack={false} title="Share on Zulip" shouldShowLoadingBanner={f
alse}>
+      <Screen
+        canGoBack={false}
+        title="Share on Zulip"
+        shouldShowLoadingBanner={false}
+        keyboardShouldPersistTaps="always"
+      >
         <SharingTopTabNavigator navigation={navigation} />
       </Screen>
     );
```

gnprice:
For the "Exit the process when the activity is destroyed" commit, I'm concerned about collateral damage and don't want to make that change before better understanding what's happening in the problem we're trying to solve. I have some questions in chat [to help you debug more fully](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/double.20initialization.20on.20sharing/near/950930).

Issue #3994, opened 2020-03-22
Night mode toggle also switches to Home card


harshbisle:
![20200322_222025](https://user-images.githubusercontent.com/41483548/77255174-8ec8ce00-6c8b-11ea-9a39-3ec7c1d4c42c.gif)

Like many folks I am a big fan of night mode. Sometimes I would even love to play around it but in zulip mobile when I click on on the night mode button it directly takes me to other mode and I can't immediately switch between the modes . I think it's an issue to be solved.

jainkuniya:
It is a recent regression, I remembered it was working as expected some time back. You can use `git bisect` to figure out which commit actually caused this.
My notes of `git bisect` https://jainkuniya.github.io/git-bisect-the-saviour/

harshbisle:
I will look into it thanks @jainkuniya

ray-kraesig:
Interestingly, if you move the theme toggle into the notifications card, toggling it doesn't kick you back out to the home screen; it leaves you on the notifications card. However, if you then hit Back, you're presented with the home tab, rather than the settings tab.

Currently, theme changes are handled in `src/boot/StylesProvider.js`, via the React Context API. The use of `key={theme}` there causes a re-`render()` of all visible UI components; presumably this is the point at which the "current-tab" navigation state is lost. (Removing `key={theme}` causes the theme to mostly not be applied.)

It's plausible that what we really want to do here is to eliminate the ThemeContext in favor of just keeping the theme state in the Redux store, like everything else, and keeping watch on it from there.

gnprice:
> (Removing `key={theme}` causes the theme to mostly not be applied.)

I think this is because in many places we still use React's legacy context API. As its docs say:
https://reactjs.org/docs/legacy-context.html#updating-context

> Updating Context
>
> Don’t do it.
>
> React has an API to update context, but it is fundamentally broken and you should not use it.

So the `key={theme}` is our workaround, for causing updates even while the (legacy) context API doesn't.

In new code, or where we're substantially changing the logic that consumes the theme anyway, we've generally moved to React 16's new context API. That one properly causes updates:
https://reactjs.org/docs/context.html#contextprovider

> All consumers that are descendants of a Provider will re-render whenever the Provider’s `value` prop changes.

Whatever we do, we should move off the legacy context API before too long; in addition to this problem, it'll be dropped in a future post-16.x release of React.

[...]

Issue #4009, opened 2020-03-27
Automatically follow user's OS-level dark/light preference


gnprice:
Over the last year or two it's become common for apps to have both a dark and a light color scheme, and let the user choose between them. Zulip has this.

Since 2019 this has become a thing at the OS level. And in recent months, I've seen more and more apps noticing my OS-level setting and following along. We should do the same, at least by default -- when the user goes into system-wide settings and says they want the device's screen to stay dark instead of bright, we shouldn't wait for an individual invitation.

This becomes especially significant now that it's an OS-level feature, in both Android and iOS, to have the light/dark mode switch back and forth every morning and evening. The only way that can work as a good experience is when apps all follow along automatically.

[RN v0.62 adds an API](https://reactnative.dev/blog/2020/03/26/version-0.62#new-dark-mode-features) for this. ~~So I think it makes sense to hold off on trying to implement this behavior until we've done that upgrade, aka #3782.~~ Done!

chrisbobbe:
As part of this, let's deprecate the 'default' name, like so:

```diff
- export type ThemeName = 'default' | 'night'
+ export type ThemeName = 'light' | 'night'
```

As I said over at https://github.com/zulip/zulip-mobile/pull/4046#discussion_r414032884:

> Hmm. Taking a look at [Apple's HIG document on Dark Mode](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/dark-mode/), though, it makes it sound like "default" semantics shouldn't apply to either theme at the application level, but rather at the device level:
> 
> > In Settings, people can choose Dark Mode as their default interface style and schedule automatic changes between the appearance modes. Because people make these choices at a systemwide level, they generally expect all apps to respect their preferences.
> 
> Perhaps we might actually deprecate the "default" name, and use "night" or "light" everywhere (with our `ThemeName` type being `'light' | 'night'` instead of `'default' | 'night'` as it is currently), at or on the path to #4009.

and Greg [agreed](https://github.com/zulip/zulip-mobile/pull/4046#discussion_r414886033).

gnprice:
It'd also be good to switch to calling it "dark mode" instead of "night mode" or "night theme" -- as "dark" is what both Google and Apple call it, and also is more logical because there's no inherent reason to only use it at night.

That, or at least the UI side of it, should happen in coordination with the webapp, though, so the same terminology is used across platforms.

gnprice:
[...]

PR #4199, opened 2020-07-20
Stop using legacy Context API for styles (continued)

chrisbobbe:
Fixes: #1946 
Fixes: #3994

A continuation of https://github.com/zulip/zulip-mobile/pull/4046, which has gotten cluttered with several unrelated conversations (sorry!). The work being done here isn't as complicated as the length of those conversations might suggest.

Greg has already merged several commits from that older PR, as he lists at https://github.com/zulip/zulip-mobile/pull/4046#issuecomment-622204993:

> I've gone and merged 10 of these commits, along with a small added commit at the end:
> f9caf224c SmartUrlInput: Switch to new React Context API for theme colors.
> a197c9916 StreamItem: Switch to new React Context API for theme colors.
> 64e142226 miscStyles [nfc]: Remove unused `color` in `miscStyles`.
> ecba735a7 StreamItem [nfc]: Readability refactor (no need to repeat object literals).
> aa4f3ad6b StreamItem [nfc]: Refactor textColor to look more like iconColor above it.
> be26c744b StreamItem [nfc]: Remove `isSelected` prop that is never passed.
> 452edbc5e StreamItem: Confirm '' cases are irrelevant, remove check, add JSDoc.
> 2a4d9ee98 StreamAutocomplete [nfc]: Rename `streams` to `matchingSubscriptions`.
> 67be6ac43 subscriptions reducer test: Test for `subscriptions` being empty array.
> 3faf5a4a7 subs reducer: Remove needless check for void `subscriptions` in REALM_INIT.
> 29c22e029 StreamItem [nfc]: Reorder props for logical grouping.
> 
> [...]
> 
> Still reviewing the rest, but this will help move things forward and also reduce the work of future rebases and reviews :slightly_smiling_face:

### DIFF
--- a/docs/architecture/react.md
+++ b/docs/architecture/react.md
@@ -103,7 +103,8 @@ long as the code adheres to core Redux principles:
 ### Context
 
 We're on board with the current API where possible; there's a
-third-party library we use that isn't there yet.
+third-party library we use that isn't there yet, but we deal with that
+at the edge by "translating" the old API to the new one.
 
 #### Current Context API
 
@@ -164,29 +165,21 @@ updates to the context, and not just by the consumer's own
 We have to think about the legacy Context API in just one place. The
 `react-intl` library's `IntlProvider` uses it to provide the `intl`
 context. The only consumer of `intl` is
-`TranslationContextTranslator`, but we've made that a `PureComponent`,
-so it doesn't get updated when `intl` changes. Our workaround until
-now has been a "`key` hack":
-
-If `locale` changes, we make the entire React component tree at and
-below `IntlProvider` remount. (Not merely re-`render`: completely
-vanish and start over with `componentDidMount`; see the note at [this
-doc](https://5d4b5feba32acd0008d0df98--reactjs.netlify.app/docs/reconciliation.html)
-starting with "Keys should be stable, predictable, and unique".) We do
-this with the [`key`
-attribute](https://reactjs.org/docs/lists-and-keys.html), which isn't
-recommended for use except in lists.
-
-In the next commit, we stop using the `key` hack and instead make
-`TranslationContextTranslator` "speak" the old API better.
+`TranslationContextTranslator`, which "speaks" the old API by being
+the direct child of `IntlProvider` and by being a `Component`, not a
+`PureComponent` (whose under-the-hood `shouldComponentUpdate` would
+suppress updates on context changes)—all to make sure that it
+re-`render`s whenever `intl` changes. Then,
+`TranslationContextTranslator` is itself a provider, and it provides
+the same value, but it does so in the new Context API way. All its
+consumers are updated appropriately, which is what we want.
 
 ### The exception: `MessageList`
 
-We have one React component that we wrote (beyond `connect` calls) that
-deviates from the above design: `MessageList`.  This is the only
-component that extends plain `Component` rather than `PureComponent`,
-and it's the only component in which we implement
-`shouldComponentUpdate`.
+We have one React component that we wrote (beyond `connect` calls)
+that deviates from the above design: `MessageList`.  It extends plain
+`Component` rather than `PureComponent`, and it's the only component
+in which we implement `shouldComponentUpdate`.
 
 In fact, `MessageList` does adhere to the Pure Component Principle -- its
 `render` method is a pure function of `this.props` and `this.context`.  So
--- a/docs/architecture/react.md
+++ b/docs/architecture/react.md
@@ -102,38 +102,100 @@ long as the code adheres to core Redux principles:
 
 ### Context
 
-Our "Pure Component Principle" says `render` is a pure function of props,
-state, *and context*.  But `PureComponent` only checks for changes to props
-and state, and skips re-render when just those two are unchanged.  Doesn't
-that open up bugs if just `this.context` changes?
+We're on board with the current API where possible; there's a
+third-party library we use that isn't there yet.
 
-[Yes, it
-would](https://reactjs.org/docs/legacy-context.html#updating-context).  For
-this reason, when something provided in context is updated, we force the
-entire React component tree under that point (in our usage of `context`,
-this is nearly the entire tree) to re-render.  This means we use `context`
-sparingly -- only for things where its benefit for code readability is very
-large, and where updates are rare so we're OK with that global re-render.
+#### Current Context API
 
-In `StylesProvider`, for example, this is done with a `key`.
+We should use the [current Context
+API](https://reactjs.org/docs/context.html) instead of the [legacy
+one](https://reactjs.org/docs/legacy-context.html) wherever possible.
+The new API aggressively ensures consumers will be updated
+(re-`render`ed) on context changes, and the old one doesn't (see
+below). From the [new API's
+doc](https://reactjs.org/docs/context.html):
 
-Relatedly, the `this.context` API we use is a legacy API.  Recent React
-versions offer a [new context API](https://reactjs.org/docs/context.html)
-that works much more like Redux and `connect`, above.
+> All consumers that are descendants of a Provider will re-render
+> whenever the Provider’s `value` prop changes.
+
+It's so aggressive that there's a potential "gotcha" with the new API:
+context consumers are the first occurrence of the following that we're
+aware of (from the [doc on
+`shouldComponentUpdate`](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)):
+
+> In the future React may treat `shouldComponentUpdate()` as a hint
+> rather than a strict directive, and returning `false` may still
+> result in a re-rendering of the component.
+
+We gather this from the following (in the [new API's
+doc](https://reactjs.org/docs/context.html)):
+
+> The propagation from Provider to its descendant consumers (including
+> [`.contextType`](https://reactjs.org/docs/context.html#classcontexttype)
+> [...])
+> is not subject to the shouldComponentUpdate method
+
+Concretely, this means that our `MessageList` component updates
+(re-`render`s) when the theme changes, since it's a `ThemeContext`
+consumer, *even though its `shouldComponentUpdate` always returns
+`false`*. So far, this hasn't been a problem because the UI doesn't
+allow changing the theme while a `MessageList` is in the navigation
+stack. If it were possible, it would be a concern: setting a short
+interval to automatically toggle the theme, we see that the message
+list's color scheme changes as we'd want it to, but we also see the
+bad effects that `shouldComponentUpdate` returning `false` is meant to
+prevent: losing the scroll position, mainly (but also, we expect,
+discarding the image cache, etc.).
+
+#### Legacy Context API
+
+The legacy Context API is
+[declared](https://reactjs.org/docs/legacy-context.html#updating-context)
+fundamentally broken because consumers could be blocked from receiving
+updates to the context, and not just by the consumer's own
+`shouldComponentUpdate`:
+
+> The problem is, if a context value provided by component changes,
+> descendants that use that value won’t update if an intermediate
+> parent returns `false` from `shouldComponentUpdate`. This is totally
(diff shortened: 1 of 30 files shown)


chrisbobbe:
I also removed the tip commit that was based on Ray's recommendation for testing that things don't break. You can find it on my non-PR branch `test-theme-timer`.

src/subscriptions/__tests__/subscriptionsReducer-test.js
```diff
@@ -1,46 +1,67 @@
...
+  desktop_notifications: false,
+  email_address: `${eg.randString()}@example.org`,
+  is_old_stream: false,
+  push_notifications: null,
+  stream_weekly_traffic: 0,
+});
+
+const subscriptionPropertiesFromStream: $Shape<Subscription> = deepFreeze(eg.stream);
+
+export const makeSubscription = (args: $Shape<Subscription> = {}): Subscription =>
```
chrisbobbe, line 26: (copied from https://github.com/zulip/zulip-mobile/pull/4046#discussion_r452568253)

Note: we may instead want to cherry-pick https://github.com/zulip/zulip-mobile/commit/947dd8d26e71d8b48714ede3e24f09f9cae80263 (not yet merged) from #3946.

src/styles/theme.js
```diff
@@ -2,14 +2,16 @@
...
+
 export type ThemeColors = {|
   color: string,
   backgroundColor: string,
   cardColor: string,
   dividerColor: string,
 |};
 
-export const themeColors: { [string]: ThemeColors } = {
+export const themeColors: { [name: ThemeName | 'light']: ThemeColors } = {
```
gnprice: Happy to have this be made more specific, but I was a bit confused by the commit message.

* Instead of "fix annotation", the message can be more specific: namely that it makes the type (or the annotation) more specific.

* I kept wanting to read the paragraph in the body as an explanation of the "why" for this change. But I couldn't see the connection -- and now I think the idea is that it's describing a future change we probably want to make to this code. Is that right?

gnprice:
Merged 11/15 commits:
a06bfae7f miscStyles [nfc]: Remove needless commented-out property.
ad429e35c Screen: Switch to new React Context API for theme colors.
b697cf148 ChatScreen: Switch to new React Context API for theme colors.
af3190e54 miscStyles [nfc]: Delete unused `screen` in `miscStyles`.
0550230df Input: Switch to new React Context API for theme colors.
8cf897e0e miscStyles [nfc]: Delete unused `input` in `miscStyles`.
2690dc7b2 Label: Switch to new React Context API for theme colors.
ac0cb58be RawLabel: Switch to new React Context API for theme colors.
6b80f4c77 miscStyles [nfc]: Delete unused `label`.
47d1108f2 theme: Remove remaining consumers of legacy Context for styles.
b530f6cba theme: Fix latent bug with colors not updating on rerender.

src/subscriptions/__tests__/subscriptionsReducer-test.js
```diff
@@ -51,83 +53,35 @@ describe('subscriptionsReducer', () => {
...
-            stream_id: 2,
-          },
-        ],
+        id: 1,
+        subscriptions: [sub1, sub2],
       });
 
-      const expectedState = [
-        {
-          color: 'red',
```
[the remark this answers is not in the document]
chrisbobbe: Mm, I think that makes sense, but just to make sure:

> But that kind of substantive change in what behavior is expected should be its own commit, separate from the big mostly-mechanical change to make things well-typed, so that it can be described in its own commit message. Could come either before or after.

The "substantive change" is the one that this commit currently does, not an additional one you're requesting, right? (I gather this from, "It's probably fine that we keep the code simple and effectively assume this case doesn't happen".)

If it comes before the commit to start type checking, I think it's most straightforward to just add `color: 'red'` to the subscription in the action.

If it comes after the type-checking commit, I think it's more complicated. There needs to be a "color" property (and all the other required properties) *at* the type-checking commit. To emulate the current expectation at the type-checking commit (that the sub in state does not get clobbered), we need to change some properties of the sub in the action while keeping the ID the same. Then, after that commit, the new commit would bring it back to the normal, "boring" sub with the same ID and properties.

Best, I guess, if things are as boring as possible all the way through, so I'll maybe do it before the type-checking commit?

docs/architecture/react.md
```diff
@@ -100,27 +100,6 @@ long as the code adheres to core Redux principles:
...
-Our "Pure Component Principle" says `render` is a pure function of props,
-state, *and context*.  But `PureComponent` only checks for changes to props
-and state, and skips re-render when just those two are unchanged.  Doesn't
-that open up bugs if just `this.context` changes?
-
-[Yes, it
-would](https://reactjs.org/docs/legacy-context.html#updating-context).  For
-this reason, when something provided in context is updated, we force the
-entire React component tree under that point (in our usage of `context`,
-this is nearly the entire tree) to re-render.  This means we use `context`
```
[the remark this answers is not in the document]
chrisbobbe: Ah, right—I wonder if I wasn't thinking of this way back when I first created this PR. Anyway, do you think this means I should remove the marker that the PR fixes #1946?

src/styles/theme.js
chrisbobbe: > Instead of "fix annotation", the message can be more specific: namely that it makes the type (or the annotation) more specific.

~~I think the "fix" was aimed at `[string]`, which looks like a subtle mistake (though not a large one) — `string` isn't interpreted as a type; it's just what we were calling the indexer property.~~ Huh, looks like I was wrong about this! ([doc](https://flow.org/en/docs/types/objects/#toc-objects-as-maps))

Makes sense.

> I kept wanting to read the paragraph in the body as an explanation of the "why" for this change. But I couldn't see the connection -- and now I think the idea is that it's describing a future change we probably want to make to this code. Is that right?

Mm, yeah (that's right), I'll make that clearer.

chrisbobbe: In fact, maybe I'll move it to a code comment on the `ThemeName` type declaration in `src/reduxTypes`.

chrisbobbe:
OK, just resolved conflicts and pushed my changes! I responded your comments above; in particular, what do you think about [this](https://github.com/zulip/zulip-mobile/pull/4199#discussion_r460275716):

> Anyway, do you think this means I should remove the marker that the PR fixes #1946?

chrisbobbe:
> **edit**: Having expanded on the doc in `docs/architecture/react.md`, one thing pops out to me: could we not remove the need for the `key` hack in `TranslationProvider` by making `TranslationContextTranslator` a regular `Component` or a "stateless function component", instead of a `PureComponent`?

OK, I just pushed a commit that does this! If all goes well, no more key hacks at all, even there.

src/subscriptions/__tests__/subscriptionsReducer-test.js
```diff
@@ -108,6 +108,7 @@ describe('subscriptionsReducer', () => {
         subscriptions: [
           {
             name: 'some stream',
+            color: 'red',
```
gnprice: It does this in a way that doesn't look intentional; it may just be
    a consequence of not being forced to use well-typed (and therefore
    more likely to be representative) data in tests. The `color`

My reading is that it was probably intentional -- the two subscription objects looked different so that the test could confirm that the reducer kept the old one rather than putting in the new one. Partly that's my reading because the description says "do not add them", and if it replaced the old object with the new one then that'd be a way of "adding" the new one. So this seems like a natural thing to try to check if you're trying to write a test to be "thorough", rather than thinking about what the expected interface ought to be and sticking to that.

src/reduxTypes.js
```diff
@@ -254,6 +254,9 @@ export type RealmState = {|
   isAdmin: boolean,
 |};
 
+// TODO: Stop using the 'default' name. As I explain in #4009, any
+// 'default' semantics should only apply the device level (at least on
+// iOS), not within the app.
```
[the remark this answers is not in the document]
chrisbobbe: Makes sense. Just updated https://github.com/zulip/zulip-mobile/issues/4009#issuecomment-619280681 and pointed the code comment at that GitHub comment.

src/subscriptions/__tests__/subscriptionsReducer-test.js
chrisbobbe: Huh, interesting, I guess that makes sense. I had read "do not add them" as the negation of "add them" in the sense it's used in the other `op: 'add'` test: "(do not) concatenate items onto the state". But my reasoning turned out to be unhelpful because "add them" quite plausibly means "add or replace them" when there are partially matching items in the state already—and this isn't possible in the "add them" test, so it won't be addressed by negating "add them" in that sense.

docs/architecture/react.md
```diff
@@ -102,24 +102,62 @@ long as the code adheres to core Redux principles:
...
+> [`.contextType`](https://reactjs.org/docs/context.html#classcontexttype)
+> and
+> [`useContext`](https://reactjs.org/docs/hooks-reference.html#usecontext))
+> is not subject to the shouldComponentUpdate method, so the consumer is
+> updated even when an ancestor component skips an update.
+
+(I think word "even" means to emphasize "ancestor" -- i.e., it's not
+only the case that a consumer's *own* `shouldComponentUpdate` is
+ignored, but that, remarkably, so is the `shouldComponentUpdate` of
+ancestor components. I think it's just as remarkable that a
```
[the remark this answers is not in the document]
chrisbobbe: Mm, you're right, I think this whole section can be made clearer; I'll rework it.

> The latter point isn't really about shouldComponentUpdate in the end...

As I understand it: true, some of those ancestors may skip an update—but if they do, their `shouldComponentUpdate`s are the reason why. They might not themselves consume the context, but `shouldComponentUpdate` can (hint to) skip an update based on nothing, or on anything it has available at the time.

docs/architecture/react.md
```diff
@@ -102,24 +102,62 @@ long as the code adheres to core Redux principles:
...
+clear instance of the
+[following](https://reactjs.org/docs/react-component.html#shouldcomponentupdate):
+"In the future React may treat `shouldComponentUpdate()` as a hint
+rather than a strict directive, and returning `false` may still result
+in a re-rendering of the component.")
+
+The legacy Context API is
+[declared](https://reactjs.org/docs/legacy-context.html#updating-context)
+fundamentally broken because `shouldComponentUpdate` and
+`PureComponent`s can block updates.
```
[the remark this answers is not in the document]
chrisbobbe: I'll push a revision where the doubts are (hopefully) clearer; anyway, I use a concrete example. Then we can decide if it's worth mentioning at all, or at such length.

chrisbobbe:
OK, just pushed my latest revision, fixing the straightforward things and hopefully clarifying the things about context.

gnprice:

docs/architecture/react.md
```diff
@@ -102,38 +102,100 @@ long as the code adheres to core Redux principles:
...
+(re-`render`s) when the theme changes, since it's a `ThemeContext`
+consumer, *even though its `shouldComponentUpdate` always returns
+`false`*. So far, this hasn't been a problem because the UI doesn't
+allow changing the theme while a `MessageList` is in the navigation
+stack. If it were possible, it would be a concern: setting a short
+interval to automatically toggle the theme, we see that the message
+list's color scheme changes as we'd want it to, but we also see the
+bad effects that `shouldComponentUpdate` returning `false` is meant to
+prevent: losing the scroll position, mainly (but also, we expect,
+discarding the image cache, etc.).
```
gnprice, line 149: Huh interesting.

That seems like a bug in the implementation of `WebView`, really -- it sounds like it's responding inappropriately to a re-`render`, and not behaving properly in the lifecycle of a React component. As React starts treating `shouldComponentUpdate` as more of a hint and less of a strict directive, this is the kind of bug that'll be getting exposed. It's possible we'll end up having to deal with the bug ourselves, if nobody else runs into it harder and fixes it first.

PR #4201, opened 2020-07-21
Store `state.narrows` as an Immutable.Map (continued)

chrisbobbe:
@gnprice [said](https://github.com/zulip/zulip-mobile/pull/4187#issuecomment-661642480):
> Please go ahead and open a new PR with (a rebased version of) [1a80243](https://github.com/zulip/zulip-mobile/commit/1a80243f6055c6640375021cd7a276d949062b7d).

Here you go!

An instance of #3949 and #3950.


chrisbobbe:
Huh, redux-logger has a way to make it easier to look at logs of Immutable things: https://github.com/LogRocket/redux-logger#transform-immutable-with-combinereducers

chrisbobbe:
OK, just fixed some conflicts.

> Huh, redux-logger has a way to make it easier to look at logs of Immutable things: https://github.com/LogRocket/redux-logger#transform-immutable-with-combinereducers

And added something based on this.

chrisbobbe:
Just fixed a recent conflict. This would set a good example for, e.g., https://github.com/zulip/zulip-mobile/issues/4252.

src/boot/store.js
```diff
@@ -228,6 +234,20 @@ function listMiddleware() {
       // and ours:
       //   https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#redux-logger
       createLogger({
+        stateTransformer: state => {
+          // Beware: errors thrown in here will silently drop actions.
+          const newState = {};
+          Object.keys(state).forEach(key => {
+            if (Immutable.isImmutable(state[key])) {
+              // Make things like `state.narrows` appear as plain
+              // JS objects in the logs (much easier to read)
```
gnprice, line 243: Neat.

Have you tried using this implementation? https://github.com/andrewdavey/immutable-devtools It's a little more code, and specifically aims at the Chrome Dev Tools API... and the result seems somewhat cleaner than is possible to get by translating to plain objects-as-maps and arrays. (One difference that seems fundamental is that if we start using some map keys that aren't strings, those can't translate properly to an object-as-map.)

Ran across that in looking at the Immutable.js wiki: https://github.com/immutable-js/immutable-js/wiki/Modules-that-work-with-Immutable.js

gnprice (commented):
Thanks @chrisbobbe ! Sorry for delay -- this looks great. See comments below.

The other major comment is: I'm pretty confident that this refactor is correct, thanks to Flow. But then the other important question about it (and future further changes to our major data structures) is to have some idea of what its effect is on the app's performance. This doesn't have to be super precise, but I'd like to (a) be confident we aren't accidentally making things a lot slower, and (b) ideally be able to confirm that as we start using good immutable-data-structure algorithms, we're making things faster.

I'll go make a chat thread to discuss.

src/chat/narrowsReducer.js
```diff
@@ -35,58 +35,54 @@ const messageFetchComplete = (state, action) => {
...
-      newState[key] = [...state[key], action.message.id];
+      return [...value, action.message.id];
     } else {
-      newState[key] = state[key];
+      return value;
     }
   });
+
   return stateChange ? newState : state;
 };
```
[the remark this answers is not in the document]
chrisbobbe: Sure, sounds good!

src/boot/store.js
chrisbobbe: Ooh, that looks helpful. Another reason to prefer something like that (as noted in a code comment): the current implementation will silently drop actions if there's an error thrown from the code that makes plain objects out of `Immutable` objects.

chrisbobbe:
Thanks for the review! I just pushed a new revision. I'll put together some performance observations soon.

docs/howto/debugging.md
```diff
@@ -158,6 +159,32 @@ call to `createLogger` in `src/boot/middleware.js`.
...
+NPM package provides the formatter for `zulip-mobile` just as
+effectively without making widespread changes in behavior when you
+browse the Web.)
+
+To enable it, open Chrome Dev Tools and press F1 to open the settings.
+In the "Console" section, tick "Enable custom formatters". If it isn't
+working, please consult the project's issue tracker before opening an
+issue in `zulip-mobile`.
+
 <div id="webview" />
```
gnprice, line 188: nit: double blank line between sections (helps with visually parsing the source)

gnprice (commented):
Thanks for the revisions! All LGTM pending those performance results, except a nit below.

That custom Chrome DevTools formatter looks great. There are some UI choices that aren't the best -- I'd rather that the expando for an element appeared at the left of the key, like it does for objects in the built-in behavior, rather than at the right of the key (which is potentially long, and potentially varies in length between different keys of a map) -- but definitely better than trying to interpret the Immutable.Map internals. Perhaps at some point I'll feel moved to try to tweak the formatter.

chrisbobbe:
Thanks for the review! And, in particular, the performance measurements, analysis, and a solution to a regression, as discussed in [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60Immutable.2EMap.60.20performance.20for.20.60state.2Enarrows.60/near/1068904).

src/unread/__tests__/unreadHuddlesReducer-test.js
```diff
@@ -115,20 +120,16 @@ describe('unreadHuddlesReducer', () => {
     test('if message is sent by self, do not mutate state', () => {
       const initialState = deepFreeze([]);
 
+      const message2 = eg.pmMessage({
+        sender: eg.selfUser,
+        recipients: [eg.makeUser({ name: 'john' }), eg.makeUser({ name: 'mark' }), eg.selfUser],
```
gnprice, line 125: These are fine, but I usually just turn them into `eg.otherUser` and `eg.thirdUser`, if there don't appear to be other places that depend on the specific names.

(If there are such places, usually it's ideal to replace those too -- it just makes it more work and so I don't always.)

src/chat/narrowsReducer.js
```diff
@@ -35,62 +35,76 @@ const messageFetchComplete = (state, action) => {
...
+      // it's the most recent message, i.e., unless we know we're
+      // currently looking at (caught up with) the newest messages in
+      // the narrow.
+      const isCaughtUp = action.caughtUp[key] && action.caughtUp[key].newer;
+      // Don't add a message that's already been added. It's probably
+      // very rare for a message to have already been added when we
+      // get an EVENT_NEW_MESSAGE, and perhaps impossible. (TODO:
+      // investigate?)
+      const messageDoesNotExist = value.find(id => action.message.id === id) === undefined;
+      if (isCaughtUp && messageDoesNotExist) {
```
gnprice, line 79: One subtle but potentially performance-significant wrinkle that's [in my fastest version](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60Immutable.2EMap.60.20performance.20for.20.60state.2Enarrows.60/near/1067739):
> I tried another optimization: [graph]
> This is one I spotted because of staring at this code, but it's actually orthogonal to Immutable vs. objects-as-maps, and we should just do it in any case:
```diff
     for (const [key, value] of state.entries()) {
-      const isInNarrow = isMessageInNarrow(action.message, flags, JSON.parse(key), action.ownEmail);
-      const isCaughtUp = action.caughtUp[key] && action.caughtUp[key].newer;
-      const messageDoesNotExist = value.find(id => action.message.id === id) === undefined;
-
-      if (isInNarrow && isCaughtUp && messageDoesNotExist) {
+      if (
+        isMessageInNarrow(action.message, flags, JSON.parse(key), action.ownEmail)
+        && (action.caughtUp[key] && action.caughtUp[key].newer)
+        && value.find(id => action.message.id === id) === undefined
+      ) {
         stateMut.set(key, [...value, action.message.id]);
       }
     }
```
> I.e., no need to go looking through the whole narrow to see if the message is already there, if it's a narrow it doesn't even go in.

In general, the pattern this existing code is doing is a bad practice: computing several different complex or potentially-expensive conditions, and only at the end looking to see that the first one already gave us our answer and the rest were irrelevant. Instead, it's better to check each condition after computing it and act on it immediately. This probably isn't as big a deal for performance as it was in my measurements, because those were mostly reflecting the `isMessageInNarrow` condition, but best to not have to worry about it.

Probably the cleanest way to format that, in the presence of your helpful comments in the next commit, is to use the early-return pattern:
```js
      if (!(action.caughtUp[key] && action.caughtUp[key].newer)) {
        // Don't add a message to the end of the list unless we know
        // it's the most recent message, i.e., unless we know we're
        // currently looking at (caught up with) the newest messages in
        // the narrow.
        return;
      }
```
though a long chain of `&&` in one `if` statement (like in my demo version) is another time-honored alternative.

gnprice (commented):
Thanks again @chrisbobbe ! Just small comments below.

src/chat/narrowsReducer.js
chrisbobbe: Ah, right. I had read and considered that optimization, but I skipped it for readability with those new comments in mind (not having considered the quite readable early-return pattern, oops) since the number of iterations has decreased to a constant ~5, as you say:

> This probably isn't as big a deal for performance as it was in my measurements, because those were mostly reflecting the `isMessageInNarrow` condition

But you're right; EVENT_NEW_MESSAGE handling is a place where we should take any easy performance gains enthusiastically. 🙂

src/unread/__tests__/unreadHuddlesReducer-test.js
chrisbobbe: Would that be OK to do in the same commit?

src/unread/__tests__/unreadStreamsReducer-test.js
```diff
@@ -36,29 +38,33 @@ describe('unreadStreamsReducer', () => {
...
           unread_msgs: {
+            ...eg.action.realm_init.data.unread_msgs,
             streams: [
               {
-                stream_id: 1,
-                topic: 'some topic',
-                unread_message_ids: [1, 2],
+                stream_id: message1.stream_id,
+                topic: message1.subject,
+                unread_message_ids: [message1.id, 2],
```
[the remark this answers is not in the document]
chrisbobbe: > These `[message1.id, 2]` feel kind of incongruous.

I think the `2` here can be removed in this spot without losing anything important. And in other cases, where it couldn't be removed, I think you're saying that we should probably represent the `2` as something like `message2.id`, so it's like `[message1.id, message2.id]`, right?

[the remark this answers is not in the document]
chrisbobbe: > Probably `eg.streamMessage({ id: 1 })` would be a good solution.

Ah, right—in the definition of `message1`, right, so

```diff
- const message1 = eg.streamMessage();
+ const message1 = eg.streamMessage({ id: 1 });
```

chrisbobbe:
Thanks for the review! I've just pushed another revision.

gnprice:
LGTM, thanks -- merged!

src/unread/__tests__/unreadHuddlesReducer-test.js
gnprice: Yeah, or can break them up if the commit is getting too unwieldy.

PR #4204, opened 2020-07-22
Add our own run-android script; cut dummy-activity hack.

gnprice:
As a comment here explains, `react-native run-android` is unable to run our app in debug mode with the correct app ID and activity name.

We've had a series of hacks that, first, convince it of the right app ID (which it misdetects on its own), and then provide a dummy activity with the name it expects: com.zulipmobile.debug.MainActivity.  See commit 0d16b604b, which introduced these hacks and explains the issues in `react-native run-android` that led to them.

But having a dummy activity leads to surprising, subtly broken, behavior when combined with launching the app through other means that use the normal activity name instead of the dummy one: the launcher, or the new "sharing" feature.  See chat discussions involving some very puzzled debugging:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/need.20for.20conditional.20in.20navReducer's.20.60rehydrate.60.20for.20Sharing/near/946142
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/double.20initialization/near/947471

So, here we take that dummy activity out.

The new `tools/run-android` script provides an alternative to `react-native run-android`, one that works correctly with our actual app ID and main-activity name.

Then we change references to that command to point to this script instead, and remove the dummy activity and the other hacks so that it's no longer possible to encounter the broken behavior they enabled.


agrawal-d:
I tested this on my `sharing-2` branch on top of this PR, and the sharing screen disappears in this specific case:

- Start the app using `yarn run android` with JS Debugging on.
- Kill the app once it launches.
- Go to another app (GitHub) and share the link to a PR with Zulip
- Send the link via a PM.
- After sending, Zulip exits and returns me to GitHub
- Share the PR link once again with Zulip
- The sharing screen appears, but I cant interact  with it ( switch tabs etc. )
- It disappears immediately.

The following redux action is responsible for removing the sharing screen from the nav stack:


```
action persist/REHYDRATE @ 12:59:39.989 (in 5.93 ms)

redux-logger.js:1  prev state
nav: {key: "StackRouterRoot", isTransitioning: true, index: 1, routes: Array(2)}

redux-logger.js:1  action 
{type: "persist/REHYDRATE", payload: {…}, error: null}

redux-logger.js:1  next state
nav: {key: "StackRouterRoot", isTransitioning: false, index: 0, routes: Array(1)}
```

gnprice:
OK, I've gone ahead and merged because I think this improves the situation. Sounds like we still have something to debug, though, at least in the sharing case.

Issue #4186, opened 2020-07-09
When message fetch fails, show error instead of loading-animation


gnprice:
When you go to read a conversation, we load the messages that are in it if we don't already have them. While we're doing that fetch, we show a "loading" animation of placeholder messages. But then if that fetch fails with an exception, we carry on showing that loading-animation forever.

That's bad because after the fetch has failed, we are not in fact still working on loading the messages, so the animation is effectively telling the user something that isn't true. It's also misleading when trying to debug, as it obscures the fact there was an error and not just something taking a long time.

Cases where we've seen this come up recently:
* #4156 [turned out to be](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders) a server bug, where the server returned garbled data on fetching any messages that had emoji reactions.
  * We'd actually seen this bug [before](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Streams.20not.20loading.20in.20android/near/883392), when it was live on chat.zulip.org and caused the same symptom. It was fixed within a couple of days, but perhaps unsurprisingly there was a deployment that happened to have upgraded to master within that window of a couple of days, and stayed there.
* #4033 may in part reflect another case of this -- at least when the loading is "endless" and not merely long. (Though because that one goes away on quitting and relaunching the app, it's definitely not the same server bug and probably is a purely client-side bug.)

Instead, when the fetch fails we should show a widget that's not animated and says there was an error. Preferably also with a button (low-emphasis, like a [text button](https://material.io/components/buttons#text-button)) to retry.

Some chat discussion of how to implement this starts [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/928698), and particularly [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/928758) and after.

PR #4205, opened 2020-07-24
When message fetch fails, show error instead of loading animation

chrisbobbe:
Fixes: #4186

There are some subtle things going on here, so I'll want to do some further testing and explaining (in commit messages and JSDocs), so I've marked this as a draft. But having code to look at will likely help along the conversation happening in [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/951023).

So glad to be this much closer to rescuing people from the dreaded infinite loading indicator! 🙂


src/message/__tests__/messageActions-test.js
```diff
@@ -16,93 +16,23 @@ const streamNarrowStr = JSON.stringify(streamNarrowObj);
...
-      const [action0, action1, action2] = actions;
-      expect(action0.type).toBe('DO_NARROW');
-      expect(action1.type).toBe('MESSAGE_FETCH_START');
-      if (action1.type === 'MESSAGE_FETCH_START') {
-        expect(action1.narrow).toEqual(streamNarrowObj);
-      }
-      expect(action2.type).toBe('Navigation/PUSH');
-    });
-
-    test('when no messages in new narrow but caught up, only switch to narrow, do not fetch', () => {
```
chrisbobbe, line 48: Probably not great to remove these tests without something to replace them—it seems like these are mostly testing work that's the job of `isFetchNeededAtAnchor`? So, maybe I should write some tests for `isFetchNeededAtAnchor`.

src/chat/FetchError.js
```diff
@@ -0,0 +1,35 @@
+/* @flow strict-local */
+
+import React, { PureComponent } from 'react';
```
chrisbobbe, line 3: This file was copied from `NoMessages`; if anything looks weird, that's a likely reason.

chrisbobbe:
Mm, from the issue:

> Preferably also with a button (low-emphasis, like a [text button](https://material.io/components/buttons#text-button)) to retry.

I haven't done this yet, in this revision.

src/message/__tests__/messageActions-test.js
chrisbobbe: OK, done.

chrisbobbe:
OK, and I just cleaned this up some more, so it's ready for a first review (I'm unmarking it as a draft). Still not sure how to proceed with `fetchOlder` and `fetchNewer`; brainstorming is ongoing [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/953303). I'm not sure whether it should block this PR or not. But I think users would be grateful for even an improvement that didn't yet handle `fetchOlder` and `fetchNewer`.

chrisbobbe:
(Just addressed some things flagged in CI and pushed again.)

src/message/fetchActions.js
```diff
@@ -43,6 +44,16 @@ const messageFetchStart = (narrow: Narrow, numBefore: number, numAfter: number):
   numAfter,
 });
 
+/* eslint-disable-next-line no-unused-vars */
+const messageFetchError = (args: {| narrow: Narrow, error: Error |}): Action => {
```
gnprice, line 47: Is this eslint-disable needed? It seems like not -- perhaps it dates from an earlier draft.

src/message/fetchActions.js
```diff
@@ -192,11 +192,13 @@ const isFetchNeededAtAnchor = (state: GlobalState, narrow: Narrow, anchor: numbe
 export const fetchMessagesInNarrow = (
   narrow: Narrow,
   anchor: number = FIRST_UNREAD_ANCHOR,
-) => async (dispatch: Dispatch, getState: GetState) => {
+) => async (dispatch: Dispatch, getState: GetState): Promise<Message[] | void> => {
   if (!isFetchNeededAtAnchor(getState(), narrow, anchor)) {
-    return;
+    // Just `return;` would be fine, but ESLint doesn't want us to do
+    // that while elsewhere returning something else.
+    return undefined;
```
gnprice, line 203: I think this doesn't even need a comment. Just `return;` would result in the same behavior, but `return undefined;` is a way of being explicit that this codepath is choosing to return `undefined`, in a function that otherwise can return a more interesting value.

(Perhaps a shorter way to say this is: I agree with that ESLint rule :wink: )

src/message/__tests__/fetchActions-test.js
```diff
@@ -21,6 +22,55 @@ describe('fetchActions', () => {
     fetch.reset();
   });
 
+  describe('isFetchNeededAtAnchor', () => {
+    test("false if we're caught up, even if there are no messages", () => {
+      const state = eg.reduxState({
```
gnprice, line 27: Let's squash this and the next commit together. That makes it a bit easier to read this commit, because it becomes visible that a lot of the code is simply moved, or moved with small edits.

src/message/fetchActions.js
```diff
@@ -314,17 +294,11 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
...
   dispatch(initialFetchComplete());
   dispatch(startEventPolling(initData.queue_id, initData.last_event_id));
 
   dispatch(fetchPrivateMessages());
 
-  const session = getSession(getState());
-  if (session.lastNarrow) {
-    dispatch(fetchMessagesInNarrow(session.lastNarrow));
-  }
-
```
gnprice, line 280: Huh, I hadn't quite spotted before that we do the "topmost narrow" fetch *before* INITIAL_FETCH_COMPLETE and the start of polling, and the "last narrow" fetch *after* those. (And the PM fetch after them as well.)

Seems likely there's a subtle bug in there, one way or the other.

src/message/fetchActions.js
```diff
@@ -280,8 +256,12 @@ const fetchTopMostNarrow = () => async (dispatch: Dispatch, getState: GetState)
  * and a mainly server-side perspective:
  *   https://zulip.readthedocs.io/en/latest/subsystems/events-system.html
  *
- * Also fetch some messages eagerly, and do some miscellaneous other work
- * we want to do when starting up, or regaining a network connection.
+ * Also do some miscellaneous other work we want to do when starting
+ * up, or regaining a network connection. Mostly, we don't fetch
+ * messages here; that's dispatched from the UI component responsible
+ * for showing the messages and error states about the fetch. The
```
gnprice, line 262: Probably best to mention one or both examples -- that way the reader who's interested in understanding how we maintain the messages data can go look there to follow along.

src/message/fetchActions.js
chrisbobbe: Ah—in the commit that adds it, it isn't used, so the ignore is needed there. But then when we start using it, it should disappear; I'll do that.

gnprice:
Looks good, merged, thanks!

Issue #4200, opened 2020-07-20
Use new logo!


chrisbobbe:
With the release of [Zulip Server 3.0](https://blog.zulip.com/2020/07/16/zulip-3-0-released/), we got a new logo! Let's use it.

Here are some [Apple docs](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/) and [Android docs](https://developer.android.com/google-play/resources/icon-design-specifications) on the app icon.

### Here's a list of places we'll need to update with the new logo.

A couple of these are external to the app itself, in the app stores.

- [ ] iOS home screen etc. (white on color/gradient background, full-bleed; PNG, sizes [here](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/))
- [ ] App Store (see iOS specs)
- [ ] Play Store (same series of PNGs as for iOS but another size; [specs](https://developer.android.com/google-play/resources/icon-design-specifications))
- [ ] Android launcher as WebP (white on color/gradient background, in circle; sizes according to “48dp” example [here](https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp)) *(the old-fashioned way — but required even if we do an adaptive icon, because those are several OS versions newer than the oldest we support)*
- [ ] [static/img/logo.png](https://github.com/zulip/zulip-mobile/blob/master/static/img/logo.png) (white on color/gradient background, in circle; PNG; 40x40) *(used only in one uncommon situation, but whatever)*
- [ ] Android notifications icon (White Z on transparent background; in fact all channels but alpha get ignored, per [docs](https://developer.android.com/about/versions/android-5.0-changes#BehaviorNotifications). … Then I can’t actually find documentation for the size, but our current ones are “24dp” i.e. half the sizes in [this example](https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp), and are WebP.)
- [ ] [static/img/message-loading.png](https://github.com/zulip/zulip-mobile/blob/master/static/img/message-loading.png), used in the splash screen on launch (white on gray background, in circle; PNG; 512x512)
- ?? [static/img/zulip-logo.png](https://github.com/zulip/zulip-mobile/blob/master/static/img/zulip-logo.png) (25 KB, 512×512, RGB, has alpha channel) *(is this used anywhere?)*
[...]

PR #4206, opened 2020-07-25
Update to the new logo; add a script to do so

gnprice:
Hopefully the script will make it easier to regenerate everything if
there's a v2.1 or v3 of the logo in the future, or if we want to tweak
the graphical variation of the logo we use in a particular context, or
just to add more sizes as Apple comes out with new hardware and
requires particular new odd sizes for the app icon there.

A few changes, other than simply using the new v2 logo:

 * On Android we include a higher-res version of the launcher icon
   than we had before: xxxhdpi.  Nominally this is for a display
   density of 640ppi, which is beyond any real device; but there are
   devices with more than the 480ppi that corresponds to the next
   density down, and they'll generally prefer to scale the higher-res
   image down than vice versa.

   * For the debug variant, we also copy the xxhdpi version over to
     xxxhdpi.  Otherwise, on a debug build, the launcher on my
     emulated Pixel 2 ends up preferring the higher-res version rather
     than the one for the right variant.

 * For the app icon on iOS, we drop a bunch of sizes that we had lying
   around but that aren't recommended in the current documentation
   from Apple.  In particular this includes a bunch of "1x" images for
   low-density screens... which turns out to mean the iPhone 3GS and
   older, which haven't been supported since iOS 7.

   * We also normalize the filenames, which has some variation in them
     that seems to be meaningless.

There'll be one additional step to deploy this, beyond the usual
builds and uploads: the icon in the Play Store needs to be updated
through the Play Console.  For the App Store this will IIUC be taken
care of by the usual upload process, because the App Store icon is
part of the "icon set" and gets built into the artifact we upload;
but for the Play Store it isn't.

Fixes: #4200

---

This passes the most basic smoke tests, in a debug build on Android. Needs more testing before release.


gnprice:
OK, @chrisbobbe and I have now [tested this](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/new.20logo/near/954121) and all seems well -- merging!

PR #4207, opened 2020-07-27
Slightly upgrade some chunks of our dependencies.

chrisbobbe:
Following discussion around [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/changes.20in.20generatedEs3.2Ejs/near/935879).

This includes
- Jest
- Sentry
- Flow

Two things from that discussion are left out, for now

- ESLint (see [what happened](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/changes.20in.20generatedEs3.2Ejs/near/954082); #4128 shouldn't have this problem)
- An "everything else" command, probably `yarn upgrade`; this would include the ESLint stuff, which causes problems as noted.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -96,6 +96,9 @@ rules:
   # don't have a way to do that, and it doesn't make sense to keep these
   # out of the repo.
   jest/no-disabled-tests: off
+  # Only using string literals for test names isn't obviously helpful;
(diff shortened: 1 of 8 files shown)


gnprice:
Merged. Thanks for taking care of these, and for the clear writeups!

PR #4208, opened 2020-07-28
compose box: Show error on sending message when not caught up.

agrawal-d:
Check if we are caught up in the current narrow, and if so, show an
error when trying to send a message asking the user to try again
after some time.

This prevents the user from possibly responding to a very old
message when the messages after it have not been loaded yet - this
can lead to the reply being out of context without the sender
realizing it.

Fixes: #3800

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -265,7 +265,7 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   handleEdit = () => {
-    const { auth, editMessage, completeEditMessage } = this.props;
+    const { auth, editMessage, completeEditMessage, _ } = this.props;
     if (!editMessage) {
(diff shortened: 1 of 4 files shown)


gnprice:
Thanks @agrawal-d !

What approach did you find for manually testing this? With a debug build, simply putting the device into airplane mode before launching the app is no good, because then it can't load the JS bundle from Metro. (The app just shows a red error screen on startup.)

agrawal-d:
I disabled remote JS debugging, this will prevent the red screen error.

I went to a narrow, with the internet turned on, so that the messages are cached and then I killed the app.

I then turned off the device internet and opened Zulip again.

Then I went to the same narrow and tried sending a message.

gnprice:
That works, thanks!

Merged, with a couple of additional commits on top.

PR #4216, opened 2020-08-04
More prep commits for new `avatar_url` handling (#4157)

chrisbobbe:
Following #4171, here's another chunk of work that I think it would be helpful to land, to make it easier to focus on the changes to start handling avatar URLs at the edge. It includes changes to `eventMiddleware` as discussed around [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/965896).


src/api/eventTypes.js
```diff
@@ -13,13 +13,27 @@
...
+  static update_display_settings: 'update_display_settings' = 'update_display_settings';
   static user_status: 'user_status' = 'user_status';
+  static realm_user: 'realm_user' = 'realm_user';
+  static realm_bot: 'realm_bot' = 'realm_bot';
+  static reaction: 'reaction' = 'reaction';
+  static heartbeat: 'heartbeat' = 'heartbeat';
+  static update_message_flags: 'update_message_flags' = 'update_message_flags';
+  static typing: 'typing' = 'typing';
+  static user_group: 'user_group' = 'user_group';
 }
```
gnprice, line 37: Originally the order of these had matched the order of the event type definitions below. (Which is the main place they're used.)

Then that got out of sync, though. These `type` values stayed alphabetical -- previously both had been, which I don't recall if I'd intended -- but the type definitions didn't.

Fine to add these others, but perhaps simplest to keep them in alphabetical order. The order of appearance in `eventToAction` is a bit chaotic, as it's governed to a large extent by things like how we group them for the `return { ...event, type: actionTypeOfEventType[event.type] }` case.

src/api/messages/getMessages.js
```diff
@@ -56,11 +56,11 @@ export const migrateMessages = (messages: ServerMessage[]): Message[] =>
     };
   });
 
-const migrateResponse = response => {
+const migrateResponse = (response, auth: Auth) => {
```
gnprice, line 59: Let's pass an `Identity` instead. From the `Auth` jsdoc:
```
 * Prefer using `Identity` where possible, which identifies the user but
 * leaves out the secret API key; use `identityOfAuth` to extract one from
 * an `Auth`.
```

(Or passing `auth.realm` would be good too.)

gnprice:
BTW this is part of the work for #4157.

(I see you've mentioned that in the title -- but a GitHub quirk is it doesn't parse titles for issue references, so it doesn't turn that into a link nor into a backlink from the mentioned issue.)

src/events/eventActions.js
```diff
@@ -73,7 +72,6 @@ export const startEventPolling = (queueId: number, eventId: number) => async (
 
       const actions = eventsToActions(getState(), events);
 
-      actionCreator(dispatch, actions, getState());
       dispatchOrBatch(dispatch, actions);
 
       actions.forEach(action => {
```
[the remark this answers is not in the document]
chrisbobbe: > If later we have some that need to come before and others after for some reason, we can have two different functions handling the different side effects

Or, I guess, move these side effects into a new custom Redux Middleware; in there, we can choose when to call `next`.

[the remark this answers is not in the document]
chrisbobbe: > This has a subtle effect on behavior which it's not obvious to me whether it matters.

Hmmm, right. I assumed it wouldn't matter, because `clearTypingNotification` starts off by doing nothing for 15 seconds. But there's another thing to think about: `clearTypingNotification` isn't dispatched unconditionally:

```
    case EVENT_TYPING_START:
      if (Object.keys(state.typing).length === 0) {
        dispatch(clearTypingNotification());
      }
      break;
```

I'll try and see if I can understand what's going on with that condition; it seems kind of wrong, like "If there are no typing notifications, clear the typing notifications".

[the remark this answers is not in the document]
chrisbobbe: OK, I resolved this confusion by reimplementing the client-side timer for the typing state; take a look, and see what you think.

chrisbobbe:
Thanks for the review! I just resolved a conflict and addressed your comments.

In particular, I reimplemented the logic around `clearTypingNotification`; I think it caused both of us some unnecessary confusion.

Hopefully I made it better; I think it makes some more sense this way. I may have gotten so deep into it that I lost some control of the narrative of the branch, so if you're not sure what's going on, feel free to ping me rather than give yourself a headache trying to guess at what I was trying to do. 🙂

src/boot/store.js
```diff
@@ -204,6 +227,44 @@ function listMiddleware() {
...
+
+    /**
+     * Drop an action from an account that is not the active account.
+     *
+     * Will only be given a plain-object action as input, because of
+     * its placement after `thunkMiddleware`.
+     *
+     * If the action has the special `initiatingIdentity` field, check
+     * it against the identity of the currently active account, and
+     * drop the action if it's not a match.
```
gnprice, line 239: Hmm, interesting solution!

I think this is not actually needed for the present use case. IIUC the TYPING_CLIENT_TIMER_EXPIRED action effectively just tells the reducer "hey BTW time T has come and gone, I heard a rumor you might want to update your state at key X at that time, take a look and see." And the reducer won't actually change anything unless it turns out it did indeed want to make that change at time T.

I could imagine cases where this is helpful for something else; say perhaps where we have some code make an API request and then dispatch an action with the resulting data, though that may not be a great example because we hope to cover those more systematically by canceling fetches on logout or account switch.

Perhaps keep this around in your back pocket for if we spot another case where we want to use it?

src/boot/store.js
```diff
@@ -204,6 +227,44 @@ function listMiddleware() {
...
+     *
+     * We could take care of this at all the dispatch sites, but it's
+     * easier to get this (privacy-relevant) check right in all those
+     * places if all we have to do is include the `initiatingIdentity`
+     * field.
+     */
+    ({ getState, dispatch }) => next => (action: Action) => {
+      // TODO: In `eventToAction`, stop spreading raw events straight
+      // from the server; we don't want any actions to include
+      // `initiatingIdentity` by mistake.
```
gnprice, line 253: FWIW I think we're actually pretty safe from this particular risk... because the server API systematically uses `snake_case`, not `camelCase` like this name. :slightly_smiling_face: 

(Though that spreading of raw events is certainly a thing we want to fix for other reasons!)

src/typing/typingReducer.js
```diff
@@ -51,30 +92,10 @@ const eventTypingStop = (state, action) => {
...
-
-  if (newTypingUsers.length > 0) {
-    return {
-      ...state,
-      [normalizedRecipients]: {
-        time: action.time,
-        userIds: newTypingUsers,
-      },
-    };
-  }
```
gnprice: This isn't a pure refactor of the `eventTypingStop` handler, but it
    makes it better by fixing a small bug. Before this change, if I'm in
    a group PM with many people, and several are typing at once, an
    EVENT_TYPING_STOP for a single person will clear the typing state of
    *everyone* who was typing, not just the person who stopped typing.

Huh, really? This code looks to me like it correctly handled that case. What am I missing?

src/typing/typingReducer.js
```diff
@@ -14,6 +14,47 @@ import { NULL_OBJECT } from '../nullObjects';
...
+  }
+
+  const newTypingUsers = state[normalizedRecipients].userIds.filter(userId => userId !== senderId);
+
+  if (newTypingUsers.length > 0) {
+    return {
+      ...state,
+      [normalizedRecipients]: {
+        ...state[normalizedRecipients],
+        time,
```
gnprice, line 47: Another thing that occurs to me on looking at this code, and in light of your later commits about TYPING_CLIENT_TIMER_EXPIRED:

It doesn't seem to make a lot of sense to bump `time` here. (Which we do both before and after this commit.) Effectively we're saying that if people A and B were typing, and then B stops typing... that will push back the time at which we should decide A isn't typing any more if we haven't heard another start from them.

And in fact because we only have the one "expired" action dispatched per start, we'll ignore that action and just keep thinking they're still typing forever.

In order to accurately track all this, I think what we really need is a separate time per sender -- namely the time we last heard of a typing start from them.

src/boot/store.js
chrisbobbe: Are you leaning toward it being fine to let the state get updated as normal, without checking whether the same user that received the EVENT_TYPING_START is still logged in?

Maybe that's fine; I guess if there's a rapid user switch and they happen to have a conversation with the exact same people, it wouldn't feel like a glitch if the new user's state was informed by the previous user's state. (A person is either typing in a conversation, or they're not.) And, thinking about privacy of the knowledge that a different account has an active conversation with the same people as the currently active account, it's more reasonable to suppose the two accounts are owned by the same person than it would be if we were building a web app (which people will log into on public computers).

Unfortunately, we can't leave the decision to ignore (or treat differently) actions for no-longer-active accounts up to the individual reducers. `typingReducer` doesn't know anything about the `accounts` state, which we'd need to determine the active account.

gnprice: > And, thinking about privacy of the knowledge that a different account has an active conversation with the same people as the currently active account, it's more reasonable to suppose the two accounts are owned by the same person than it would be if we were building a web app (which people will log into on public computers).

Right -- there's no privacy issue with information moving around the app. A potential privacy issue only arises when sending information to the server (or outside the app in general, but for us that almost always means to the server.)

> Maybe that's fine; I guess if there's a rapid user switch and they happen to have a conversation with the exact same people, it wouldn't feel like a glitch if the new user's state was informed by the previous user's state.

Right but I think even more than that is true here: because the reducer checks the timestamp, the only information the reducer is really relying on from the action is the fact that the current time is at least what the action says it is. If the reducer gets an action which was really intended for the old account but ends up acting on it, then it must also have been the case that it was already late enough that it ought to be getting any moment now an action with the same effect that was intended for the new account.

gnprice:
Some chat discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Typing-state.20client.20timer/near/986619). Merging all the changes apart from the typing-status changes:

78e00a938 eventActions [nfc]: Remove unnecessary checks in `responseToActions`.
ef3c7fd37 eventToAction: Make shell a little crunchier for `message` events.
f2b5b7146 eventMiddleware: Stop mutating events.
871742a9f eventMiddleware: Rename; run on actions instead of events.
2c5d03c29 eventActions [nfc]: Rename `responseToActions` to `eventsToActions`.
bf504a939 doEventActionSideEffects [nfc]: Turn into an action creator.
6217f891e fetchActions tests: Mock raw server message for fetch response.
eb2fd3316 eventTypes [nfc]: Synchronize EventTypes class with `eventToAction`.
99f959f62 migrateMessages tests [nfc]: Make type alias for something we write twice.
c72ec2c68 getMessages [nfc]: Plumb `identity` down to `migrateMessages`.

I definitely still want to see that area cleaned up, but we need some more iteration on how to do so.

chrisbobbe:
Sounds good, thanks for the review!

PR #4219, opened 2020-08-06
UI tweaks on mention-unsubscribed warning

gnprice:
This follows up on some of the UI points I noted in this comment before merging #4101: https://github.com/zulip/zulip-mobile/pull/4101#issuecomment-669764675

Specifically, going through that comment in order:
- [x] button shape/size in visual appearance -- done here
- [ ] button touch target to 48px (extending beyond visual button) -- still TODO
- [x] no border -- done here
- [ ] background color -- still TODO
- [x] text/background color contrast, for legibility -- done in #4101 before merge
- [x] cut animation -- done here
- [ ] use a nice slide-in animation instead -- still TODO

### DIFF
--- a/src/common/ZulipButton.js
+++ b/src/common/ZulipButton.js
@@ -89,6 +89,7 @@ type Props = $ReadOnly<{|
       paddingRight?: mixed,
       height?: mixed,
       width?: mixed,
+      borderWidth?: mixed,
       borderRadius?: mixed,
       flex?: mixed,
       alignSelf?: mixed,
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -10,7 +10,6 @@ import { isPrivateNarrow } from '../utils/narrow';
 import * as api from '../api';
 import { showToast } from '../utils/info';
 
-import AnimatedScaleComponent from '../animation/AnimatedScaleComponent';
 import MentionedUserNotSubscribed from '../message/MentionedUserNotSubscribed';
 
 type State = {|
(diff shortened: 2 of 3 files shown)


agrawal-d:
Regarding the border, its present to separate multiple warnings.

gnprice:
> Regarding the border, its present to separate multiple warnings.

Yeah. As I say in the commit message:

    When there are several of these banners, this means there won't be a
    dividing line between them.  I tried that out and it looked fine.

Here's a screenshot:
Screenshot_1596763759

When there are several banners, the effect is a lot like a [Material list](https://material.io/components/lists). A list like this one typically won't have dividers between the elements:
> Place a divider between rows with lots of content, such as those with three-line lists.

chrisbobbe:
LGTM!

> * [ ]  button touch target to 48px (extending beyond visual button) -- still TODO

For this, we'll likely want to use the [`hitSlop` prop](https://reactnative.dev/docs/view#hitslop) of the component that gets the `onPress` prop. We may have to adjust the dimensions of its parent, though:

> The touch area never extends past the parent view bounds and the Z-index of sibling views always takes precedence if a touch hits two overlapping views.

gnprice:
Merged! Thanks for the reviews.

I've filed #4224 for the remaining followups here. Including the mention of `hitSlop` -- that sounds very helpful! I hadn't known about that one and had imagined doing it by having `ZulipButton` contain some padding inside the touchable while outside the part that looks like a button. I think that could work, but this sounds potentially simpler.

PR #4220, opened 2020-08-07
i18n [nfc]: Add interpolation support to our `_` wrapper.

gnprice:
This allows us to stop using the `intl.formatMessage` API in its most extra-verbose form.  (When there are variables to interpolate, both `defaultMessage` and `id` are required, even though for us they're the same thing -- because while react-intl will fall back to `id` as the message when a translation isn't available, it will do so literally, without any interpolation. @agrawal-d pointed this out at https://github.com/zulip/zulip-mobile/pull/4175#discussion_r459261420 .)


chrisbobbe:
Merged, thanks!

PR #4221, opened 2020-08-07
StyleSheet.create and other refactors

gnprice:
This is the other set of followups from https://github.com/zulip/zulip-mobile/pull/4101#issuecomment-669814479 where I merged #4101: some small refactors to the ZulipButton component, and then that led me to some digging into StyleSheet.create and responding to an unpleasant discovery there, and to a bit of Flow metaprogramming.

(I actually still have some further changes following up on these, going deeper into tools for the Flow metaprogramming with some of the things I learned while writing this `SubsetProperties`. Those aren't quite ready and I'll send them separately.)

### DIFF
--- a/src/generics.js
+++ b/src/generics.js
@@ -1,0 +1,80 @@
+/* @flow strict-local */
+/**
+ * Tools for manipulating generic types in the Flow type system.
+ */
+
+/* eslint-disable flowtype/generic-spacing */
+
+/**
+ * The type `S`, plus a check that `S` is a supertype of `T`.
+ *
+ * Gives an error unless `S` is a supertype of `T` -- i.e., a `T` can always
+ * be used where an `S` is required.
+ */
+export type IsSupertype<+S, +T: S> = S; // eslint-disable-line no-unused-vars
+
+/**
+ * An inverse of the operation `(A, B) => {| ...A, ...B |}`, with checking.
+ *
+ * Similar to `$Diff`, but checks that the subtraction makes sense.
+ * Also always subtracts a given property completely, even if its type on
+ * `L` happens to be more specific than on `U`.
+ *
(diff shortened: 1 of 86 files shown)


src/generics.js
```diff
@@ -47,6 +47,41 @@ export type BoundedDiff<-U, -L> = $Diff<
...
+//
+// Sadly that doesn't work (as of Flow v0.130) when T has any optional
+// properties not found in U: the first diff type will still have them
+// optional, this will still be true after the $ObjMap, and so the second
+// diff won't remove them and they'll all wind up in the resulting type.
+//
+// Possibly this will get better after a project to fix the handling of
+// $Diff and other "type destructors", which as of 2020-08 the Flow team is
+// taking up as a priority:
+//   https://github.com/facebook/flow/issues/7886#issuecomment-669977952
```
chrisbobbe, line 82: So glad to see such a quick (and promising!) response from Flow! It's awesome that you wrote such a detailed comment for them to focus on.

chrisbobbe: I have a friend who's working at Facebook, and I mentioned that we use Flow. He said something like, "Whoa, actual Flow users in the wild!!"

gnprice:
Thanks for the review!

Yeah, the added blank lines were not intended. I noticed a couple of them while developing the change -- they were actually introduced by `eslint --fix`, as when it deduplicates two imports from the same module:
```js
import styles, { ThemeContext } from '../styles';
import { createStyleSheet } from '../styles';
```
->
```js
import styles, { createStyleSheet, ThemeContext } from '../styles';

```
it leaves behind a blank line where the removed import was. Seems like a bug in that ESLint rule, though not a super bad one.

I'll fix those up and merge.

PR #4223, opened 2020-08-11
Revert "deps: Upgrade ESLint to ^6.8.0, and related deps."

chrisbobbe:
See details in commit message; this follows from [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SOLVED.3A.20Prettier.20auto-formatting/near/981661).


gnprice:
Thanks for tracking this down!

PR #4226, opened 2020-08-12
Prep for initial-fetch timeout.

### LINKED ISSUE #4165: Time out initial fetch, and go to account-picker screen

chrisbobbe:
This work helps prepare for fixing #4165.

Work on this issue started at https://github.com/zulip/zulip-mobile/pull/4166, then we ran into a bunch of issues with getting the tests to work. https://github.com/zulip/zulip-mobile/pull/4193 is currently open as a draft, for discussion about how to fix those; Greg and I had some of that discussion over the phone, and it continues on CZO [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Lolex.20timers/near/983381).

These are some commits from #4193 that aren't being held up by anything, so they're ready for review. They don't fix #4165 at all, but they do prepare for us to get working (and not absurd-looking) tests.


package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@expo/react-native-action-sheet": "^3.4.0",
+    "@jest/source-map": "^26.1.0",
```
gnprice, line 35: This can go in `devDependencies`, alongside the other Jest-related packages.

(Discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SOLVED.3A.20Prettier.20auto-formatting/near/985105) apropos of a different recent change.)

chrisbobbe:
Thanks for the review! Just merged, after moving that dependency to `devDependencies`.

PR #4227, opened 2020-08-13
Clean up "actionCreator" and part of typing-status expiry

gnprice:
This follows up on #4216 to clean out the very odd `actionCreator` function and a bit of the logic around expiring typing status.

This doesn't include fixes for all the things that are off about how we expire typing status -- notably the loop at a fixed 15s period, which means that a status might wait as long as 30s before we get around to expiring it. But it should at least isolate all the oddness to specifically the typing-status code, and get it out of the way of the general event-handling code.


src/typing/typingActions.js
```diff
@@ -30,7 +30,10 @@ const typingStatusExpiryLoop = () => async (dispatch: Dispatch, getState: GetSta
 };
 
 /** Start the typing-status expiry loop, if there isn't one already. */
-export const ensureTypingStatusExpiryLoop = () => async (dispatch: Dispatch, getState: GetState) => {
+export const ensureTypingStatusExpiryLoop = () => async (
```
chrisbobbe, line 33: To me, this line-break change suggests that you're also affected by the auto-formatting issue I mentioned [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SOLVED.3A.20Prettier.20auto-formatting/near/985123), hmm. IIUC this should have been done in the "ensureTypingStatusExpiryLoop" rename commit.

gnprice: Ah indeed, thanks. Pushed a fix.

chrisbobbe:
Thanks! This looks good to me; looks like you fixed that one formatting nit. Did you find any others, though? Currently I've been running `tools/test prettier --fix` at each commit.

chrisbobbe:
Merged, thanks, @gnprice! (I ran `tools/test` at each commit first.)

There's a planned followup to this, currently at your `dev-expire-typing` branch; discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Typing-state.20client.20timer/near/987163). Would you like me to take that up? Or, later in that discussion, you said this:

> ... As I think more about that, the conclusion I'm coming to is that the good systematic solution is really for us to get on with maintaining the user's data from all their logged-in accounts at once, rather than focusing on one "active" account as we do.

Does that mean we should hold off on the followup until we do that reorganization?

PR #4234, opened 2020-08-21
i18n: Add script tools/tx-sync to automate more steps.

gnprice:
This brings the process for cleanly syncing with Transifex down
much closer to a single command.

That in turn will hopefully help us do it routinely right after
merging any change that adds or modifies message strings, so that
translators who are active can more consistently have an
opportunity to supply translations before the new message strings
appear in even one release.

### DIFF
--- a/docs/howto/translations.md
+++ b/docs/howto/translations.md
@@ -130,21 +130,19 @@ You'll want Transifex's CLI client, `tx`.
 [tx-docs-maintainers]: https://docs.transifex.com/teams/understanding-user-roles#project-maintainers
 
 
-### Uploading strings to translate
+### Regular operation
 
-Run `tx push -s`.
+To sync with Transifex, run `tools/tx-sync`.
 
-This uploads from `static/translations/messages_en.json` to the
-set of strings Transifex shows for contributors to translate.
-(See `.tx/config` for how that's configured.)
+This syncs in both directions, and makes local commits with any
+changes.  Review the results and then push to the central repo.
 
+The sync uploads from `static/translations/messages_en.json` to the
+set of strings Transifex shows for contributors to translate, and
+downloads translations to files `static/translations/messages_*.json`.
+(See `.tx/config` for how those paths are configured.)
 
-### Downloading translated strings
-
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Nice! I haven't tried it out at all, but I'd be happy to, if you'd like.

One thought has occurred to me so far. IIUC, there's not currently something in place to prevent absent-mindedly running the script when the repo is in a bad state; in particular, this may make the side effect in Transifex a bad one. Do I understand correctly, and are you concerned about this? If so, I wonder if there's some kind of trustworthy automated check (not sure) or if it's best left as a simple "Warning... Y/N" prompt.

chrisbobbe:
> One thought has occurred to me so far. IIUC, there's not currently something in place to prevent absent-mindedly running the script when the working tree is in a bad state; in particular, this may make the side effect in Transifex a bad one. Do I understand correctly, and are you concerned about this? If so, I wonder if there's some kind of trustworthy automated check that could be done (not sure) or if it's best left as a simple "Warning... Y/N" prompt.

This can easily be done in a follow-up, if it's important. 🙂 I'm merging this to reduce the backlog.

gnprice:
Good strategy. :slightly_smiling_face: Thanks!

> One thought has occurred to me so far. IIUC, there's not currently something in place to prevent absent-mindedly running the script when the working tree is in a bad state; in particular, this may make the side effect in Transifex a bad one. Do I understand correctly, and are you concerned about this? If so, I wonder if there's some kind of trustworthy automated check that could be done

I agree it'd be good to have a check like this. In fact I remember having that thought when I was first writing the script, and then apparently I forgot about it before I sent the PR.

So then what happened is I saw your comment, thought "ah yes, I should add that check", and then went and dealt with something else instead. That repeated a couple of times over a few weeks.

In retrospect I guess that means the right move the first time I saw your reply would have been to skip right away to where it is now: say "yes, good idea, let's do that as a followup" and merge it, perhaps with a TODO comment.

I think I'll naturally be reminded of this when I use this script, so I'll probably end up writing this check one of the next few times I do that.

Issue #4146, opened 2020-06-09
Use URL web API polyfill for URLs, instead of string concatenation / regexes.


chrisbobbe:
Now that #4081 has landed, adding the `react-native-url-polyfill` package, we can finish using URL objects instead of string concatenation and regexes!

See 8f0f5aba8 for an example of how to do this in a simple case.

One good place to start would be `getFullUrl` in `url.js`. I'm optimistic that we can replace all of its callers with something more transparent, using this URL API, and dissolve `getFullUrl`. If not, then it should at least be made much clearer and bug-free.

We'll also want to look closely at every helper function in `src/utils/url.js`. Most of them aren't tailored to logic specific to Zulip, which is a hint that there's been some reinvention of the wheel that we won't have to do anymore.

There are likely a few idiosyncratic cases that don't even use the helpers in `src/utils/url.js`. We'll want to track down all of these and fix them.

I think it wouldn't be a bad idea for, e.g., `realm` in `Account` objects, to use a URL object instead of a string. A URL instance is not serializable, so we'd need correct replace/revive logic, as in bfe794955 and following.

agrawal-d:
Does this polyfill contain exported functions for decoding/encoding links? If so, this can also fix https://github.com/zulip/zulip-mobile/issues/4136.

chrisbobbe:
I just added a link to an MDN doc for the thing this is polyfilling, which should help answer questions like these. 🙂  Here's one quote:

> URLs are encoded according to the rules found in RFC 3986.

I'm not sure this will fulfill what @gnprice describes at https://github.com/zulip/zulip-mobile/issues/4136#issue-631242811, though:
> So it seems like we need to, in `openLink` on iOS before passing the URL to `SafariView.show`:
> * %-encode non-ASCII characters -- that's #3315;
> * but *not* %-encode `%` itself, instead leave it alone;
> * and presumably also leave alone all the other characters that `encodeURI` leaves alone, like `a` and `Z` and `/`;
> * and it's not clear if we should %-encode the remaining characters that `encodeURI` affects, like ` ` and `"` and a bunch of other punctuation.

I'm wondering whether this selective encoding of some things and not others might introduce its own bugs? Probably not larger ones that those we've described so far, though.

chrisbobbe:
[...]

PR #4235, opened 2020-08-21
Store `state.accounts[].realm` as a WHATWG URL object.

chrisbobbe:
Prompted recently by a performance concern in #4230 (see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/993671)), this is also something we've been thinking of in the background as part of #4146.

It was refreshing to see a recent, relevant issue in Jest get resolved quite quickly, and (maybe even more refreshing!) to be able to just bump up a few minor versions to get the fix. It was so easy because of the recent jump in a few major Jest versions we did recently. 🎉

### DIFF
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -27,7 +27,7 @@ import styles from '../styles';
 import { Centerer, Screen, ZulipButton } from '../common';
 import { getCurrentRealm } from '../selectors';
 import RealmInfo from './RealmInfo';
-import { encodeParamsForUrl, tryParseUrl } from '../utils/url';
+import { encodeParamsForUrl } from '../utils/url';
 import * as webAuth from './webAuth';
 import { loginSuccess, navigateToDev, navigateToPassword } from '../actions';
 import IosCompliantAppleAuthButton from './IosCompliantAppleAuthButton';
@@ -168,7 +168,7 @@ export const activeAuthentications = (
 
 type Props = $ReadOnly<{|
   dispatch: Dispatch,
-  realm: string,
+  realm: URL,
   navigation: NavigationScreenProp<{ params: {| serverSettings: ApiResponseServerSettings |} }>,
 |}>;
 
@@ -260,7 +260,7 @@ class AuthScreen extends PureComponent<Props> {
       id_token: credential.identityToken,
     });
 
-    openLink(`${this.props.realm}/complete/apple/?${params}`);
+    openLink(`${this.props.realm.toString()}/complete/apple/?${params}`);
 
     // Currently, the rest is handled with the `zulip://` redirect,
     // same as in the web flow.
@@ -275,12 +275,7 @@ class AuthScreen extends PureComponent<Props> {
       return false;
     }
 
-    const host = tryParseUrl(this.props.realm)?.host;
-    if (host === undefined) {
-      // `this.props.realm` invalid.
-      // TODO: Check this much sooner.
-      return false;
-    }
+    const { host } = this.props.realm;
 
     // The native flow for Apple auth assumes that the app and the server
     // are operated by the same organization, so that for a user to
@@ -347,5 +342,5 @@ class AuthScreen extends PureComponent<Props> {
 }
 
 export default connect(state => ({
-  realm: getCurrentRealm(state).toString(),
+  realm: getCurrentRealm(state),
 }))(AuthScreen);
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -226,7 +226,7 @@ class AuthScreen extends PureComponent<Props> {
     const { dispatch, realm } = this.props;
     const auth = webAuth.authFromCallbackUrl(event.url, otp, realm);
     if (auth) {
-      dispatch(loginSuccess(auth.realm.toString(), auth.email, auth.apiKey));
+      dispatch(loginSuccess(auth.realm, auth.email, auth.apiKey));
     }
   };
 
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -260,6 +260,7 @@ class AuthScreen extends PureComponent<Props> {
       id_token: credential.identityToken,
     });
 
+    // TODO: Use a `URL` computation, for #4146.
     openLink(`${this.props.realm.toString()}/complete/apple/?${params}`);
 
     // Currently, the rest is handled with the `zulip://` redirect,
--- a/src/start/AuthScreen.js
+++ b/src/start/AuthScreen.js
@@ -226,7 +226,7 @@ class AuthScreen extends PureComponent<Props> {
     const { dispatch, realm } = this.props;
     const auth = webAuth.authFromCallbackUrl(event.url, otp, realm);
     if (auth) {
-      dispatch(loginSuccess(auth.realm, auth.email, auth.apiKey));
+      dispatch(loginSuccess(auth.realm.toString(), auth.email, auth.apiKey));
     }
   };
 
@@ -347,5 +347,5 @@ class AuthScreen extends PureComponent<Props> {
 }
 
 export default connect(state => ({
-  realm: getCurrentRealm(state),
+  realm: getCurrentRealm(state).toString(),
 }))(AuthScreen);
--- a/src/start/webAuth.js
+++ b/src/start/webAuth.js
@@ -98,7 +98,7 @@ export const authFromCallbackUrl = (
     && otpEncryptedApiKey.length === otp.length
   ) {
     const apiKey = extractApiKey(otpEncryptedApiKey, otp);
-    return { realm, email, apiKey };
+    return { realm: new URL(realm), email, apiKey };
   }
 
   return null;
--- a/src/start/webAuth.js
+++ b/src/start/webAuth.js
@@ -66,11 +66,7 @@ export const closeBrowser = () => {
  */
 const extractApiKey = (encoded: string, otp: string) => hexToAscii(xorHexStrings(encoded, otp));
 
-export const authFromCallbackUrl = (
-  callbackUrl: string,
-  otp: string,
-  realm: string,
-): Auth | null => {
+export const authFromCallbackUrl = (callbackUrl: string, otp: string, realm: URL): Auth | null => {
   // callback format expected: zulip://login?realm={}&email={}&otp_encrypted_api_key={}
   const url = tryParseUrl(callbackUrl);
   if (!url) {
@@ -82,8 +78,7 @@ export const authFromCallbackUrl = (
     return null;
   }
   const callbackRealm = tryParseUrl(callbackRealmStr);
-  // TODO: Check validity of `realm` much sooner
-  if (!callbackRealm || callbackRealm.origin !== new URL(realm).origin) {
+  if (!callbackRealm || callbackRealm.origin !== realm.origin) {
     return null;
   }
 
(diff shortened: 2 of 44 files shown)


chrisbobbe:
Just resolved a conflict.

src/start/webAuth.js
```diff
@@ -71,19 +71,42 @@ export const authFromCallbackUrl = (
...
+  if (callbackRealmStr === null) {
+    return null;
+  }
+
+  let callbackRealm: URL;
+  try {
+    callbackRealm = new URL(callbackRealmStr);
+  } catch (e) {
+    return null;
+  }
```
gnprice, line 86: This bit can be simplified with `tryParseUrl` too, I think.

src/start/webAuth.js
```diff
@@ -71,19 +71,42 @@ export const authFromCallbackUrl = (
...
+
+  // TODO: Check validity of `realm` much sooner
+  if (callbackRealm.origin !== new URL(realm).origin) {
+    return null;
+  }
+
+  const email = url.searchParams.get('email');
+  const otpEncryptedApiKey = url.searchParams.get('otp_encrypted_api_key');
 
   // callback format expected: zulip://login?realm={}&email={}&otp_encrypted_api_key={}
```
gnprice, line 95: This comment probably logically belongs above all the parsing above.

gnprice (commented):
Thanks @chrisbobbe ! I've now read through the first half or so of this branch:
3f9892d8c jest: Prepare for `eslint-plugin-jest` upgrade.
1c2d73082 deps: Add `typescript` as dev dependency.
affc9205b jest: Update all Jest-related dependencies to their latest.
10622c6a3 render test: Use representative `realm` value for `messageTypingAsHtml`.
ceedce91f url tests: Use representative `realm` value.
ee2764b86 url: Make `getFullUrl` use URL constructor.
20427e70e url [nfc]: Inline all uses of `getFullUrl` and remove it.
26272d815 webAuth: Better compare web auth callback's `realm` with current realm.
0e8820e70 store: Replace/revive URL instances.

and have just the comments below.

chrisbobbe:
Thanks for the review! I've pushed changes to address your comments.

.eslintrc.yaml
```diff
@@ -169,6 +169,18 @@ rules:
...
+  # outside a `test` or `it` block (e.g., in a `beforeEach` or
+  # `afterEach`) means that the assertion isn't executed. Empirically,
+  # this seems just wrong [2], and there are a few places where we
+  # want to call `expect` in a `beforeEach` or `afterEach` to assert
+  # that multiple tests in the same file aren't interfering with each
+  # other by leaving bits of state lying around.
+  #
+  # [1] https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-standalone-expect.md
+  # [2] https://github.com/zulip/zulip-mobile/pull/4235#discussion_r489984717
+  jest/no-standalone-expect: off
```
gnprice, line 191: Adapt to the first two and disable the third, which seems unhelpful,
    as noted in code comments. It did, however, lead to an improvement
    in `sentry-test`, where we move some `expect`s from `beforeEach` and
    `afterEach` into the tests themselves, since logically they're part
    of the tests themselves [1].

I do appreciate that these lint rules highlighted this improvement for us. But I think the credit can be assigned just as well to `jest/expect-expect`, which we're keeping :slightly_smiling_face: . The two tests that were improved were tests that had no `expect` calls of their own.

In fact I think it's not a coincidence that the tests that were improved were tests that would have been flagged by `expect-expect` alone, while the tests that wouldn't have been improved were tests that that rule thought was fine and only `no-standalone-expect` complained about. The fact that those Sentry tests had no `expect` of their own is closely tied to the fact that they were relying on `afterEach` to do the actual testing; and the fact that the heartbeat tests were more self-contained tests and their `afterEach` was more about protecting the tests from each other is closely tied to the fact that those tests had their own `expect` calls where they did their job as tests.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -1,4 +1,8 @@
 /* @flow strict-local */
+// Tell Jest to recognize `expectStream` as a helper function that
+// runs assertions.
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectStream"] }] */
```
gnprice, line 4: similarly, probably best next to the definition of `expectStream`

src/start/webAuth.js
```diff
@@ -71,19 +71,37 @@ export const authFromCallbackUrl = (
...
   realm: string,
 ): Auth | null => {
-  const url: Url = parseURL(callbackUrl, true);
-
   // callback format expected: zulip://login?realm={}&email={}&otp_encrypted_api_key={}
+  const url = tryParseUrl(callbackUrl);
+
+  if (!url) {
+    return null;
+  }
```
gnprice, line 75: nit: for this sort of "check return value" pattern, no blank line -- the check and the line above it that produces the value are closely connected, so best to keep them visually together

gnprice (commented):
Updates look good! Just the small comments below.

I still need to read the remaining commits in the branch; but for the commits mentioned above at https://github.com/zulip/zulip-mobile/pull/4235#pullrequestreview-490249761 , feel free to merge after these comments. Or a prefix of them, if you'd rather e.g. save "store: Replace/revive URL instances." to go along with the subsequent commits that use it.

.eslintrc.yaml
chrisbobbe: Makes sense!

chrisbobbe:
> I still need to read the remaining commits in the branch; but for the commits mentioned above at [#4235 (review)](https://github.com/zulip/zulip-mobile/pull/4235#pullrequestreview-490249761) , feel free to merge after these comments. Or a prefix of them, if you'd rather e.g. save "store: Replace/revive URL instances." to go along with the subsequent commits that use it.

OK, done, thanks!

chrisbobbe:
OK, just resolved some conflicts.

I also found a wrinkle and pushed a fix, but I feel slightly uneasy about it; maybe there's a better way? It's implemented in

8b966423807c6af839800adcffb8914bbe3c3489 webview: Prepare to "revive" realm in `handleInitialLoad`.

and a small tweak in that specific area of the code in

8c0cc3f409f2a3ed89dd2950a2c80714077d3311 accounts: Make `Auth.realm` a URL object, including in `state.accounts`.

gnprice:
> I also found a wrinkle and pushed a fix, but I feel slightly uneasy about it; maybe there's a better way? It's implemented [...] in
> 
> [8c0cc3f](https://github.com/zulip/zulip-mobile/commit/8c0cc3f409f2a3ed89dd2950a2c80714077d3311) accounts: Make `Auth.realm` a URL object, including in `state.accounts`.

Hmm, I see!

Seems like one underlying awkward thing here is that the call site of `handleInitialLoad` is not type-checked. Which is because it's generated code, from this template string:
```js
export default (scrollMessageId: number | null, auth: Auth): string => `
<script>
// …
document.addEventListener('DOMContentLoaded', function() {
  ${compiledWebviewJs}
  compiledWebviewJs.handleInitialLoad(
    ${JSON.stringify(Platform.OS)},
    ${JSON.stringify(scrollMessageId)},
    ${JSON.stringify(auth)}
  );
});
</script>
`;

... But then when thinking about ways to address that, I find another underlying awkward thing: our `JSONable` type does not quite mean what it says it means, which is values that can be round-tripped through JSON. In particular, it has `URL` as a subtype. Demo [here](https://flow.org/try/#0FAehAIFMA8BcCcCGBjWkAm4Bm8D2BbbAG1wHciBLAIxHQIDoArAZ2HUmSMXknE8WbNwAVQBKAGXABvYOHDNYiWBWTgeAN1wBrSAHkqjDrDHiAFAFd4RAFzyEFAHYBzAJS3NFdAG5ZfXA4V4c1RceAsrW0DHJwAacCoBSAB+SPtncAAfEQk3cA9vXwALAULU+GifOULcBTKKoprYB0R8SDrnSvBCnix2p07QiidHPq9wMDVIRDoHIgBPXwAHAWZSUPRRpaVC5tbNuUXQ2H3wRbxYXGRcGztyjt9mKfhkUtv6uXNH+F22t-uAX2AwFgc0WvAAUgBlXQAOUQVCIvAAvL4sg5zEQiKi-k5sej8FRIPBsVRcNcpg5sVIsgBqADaUWcAF1bFDYfDEZlAXIsgASURTdC6WZzACC8CQcwAPGy4QjIAA+HxAq4BWDgaC2EzgJHgByQUjZMwAcnoxpcytMmvAso5kAt4wgDlwUAloQAhHJtShkJBFmhMAIbdC5Yj6EDoPQLpC0k5TBbgEA).

So I don't have a great idea for how to arrange this so that the computer checks we haven't missed something, for when these types potentially change again in the future. Failing that, this seems like a fine way to handle the types as they currently are.

File: src/webview/js/js.js [hunk 6]
```diff
@@ -544,8 +544,11 @@ const handleUpdateEventContent = (uevent: WebViewUpdateEventContent) => {
 export const handleInitialLoad = (
   platformOS: string,
   scrollMessageId: number | null,
-  auth: Auth,
+  // In an upcoming commit, the `realm` part of an `Auth` object will
+  // be a URL object. It'll be passed here in its stringified form.
+  rawAuth: {| ...$Diff<Auth, { realm: mixed }>, realm: string |},
 ) => {
+  const auth: Auth = { ...rawAuth, realm: rawAuth.realm };
   // Since its version 5.x, the `react-native-webview` library dispatches our
```
gnprice: nit: blank line to separate these

src/start/AuthScreen.js
```diff
@@ -260,7 +260,7 @@ class AuthScreen extends PureComponent<Props> {
       id_token: credential.identityToken,
     });
 
-    openLink(`${this.props.realm}/complete/apple/?${params}`);
+    openLink(`${this.props.realm.toString()}/complete/apple/?${params}`);
```
gnprice, line 263: This probably wants to be a `URL` computation.

(That'd be a followup, another thing we can do to take advantage of having `URL` objects everywhere; for this branch a TODO comment would be fine.)

src/utils/internalLinks.js
```diff
@@ -20,16 +22,18 @@ const getPathsFromUrl = (url: string = '', realm: string) => {
 };
 
 /** PRIVATE -- exported only for tests. */
-export const isInternalLink = (url: string, realm: string): boolean =>
-  isUrlOnRealm(url, realm) ? /^(\/#narrow|#narrow)/i.test(url.split(realm).pop()) : false;
+export const isInternalLink = (url: string, realm: URL): boolean =>
+  isUrlOnRealm(url, realm)
+    ? /^(\/#narrow|#narrow)/i.test(url.split(realm.toString()).pop())
```
gnprice, line 27: This bit deserves a TODO comment about replacing it with `URL`, similar to a couple of other such comments added in this branch.

gnprice (commented):
Looks good! Small comments below.

I went through the codebase with the search `rg '[rR]ealm.toString'` to look for likely places to convert further toward URL objects, and only found a couple of items, mentioned below. Also `rg 'new URL\(.*realm'` and didn't find anything.

On another side of things, I did a pair of searches for comparisons that potentially need `toString()` added: `rg -i 'realm.*[!=]=='` and `rg -i '[!=]==.*realm'`. Everything looked to be covered.

chrisbobbe:
Thanks for the review! Glad to catch those things. 🙂  I've just pushed my changes.

chrisbobbe:
In fact, this might not need another review; your suggestions are small and were easily addressed. I'll merge this in the next few minutes unless you say otherwise. 🙂

Issue #4245, opened 2020-09-02
Upgrade to RN v0.63


chrisbobbe:
~~We won't start real work on this until the upgrade to v0.62, https://github.com/zulip/zulip-mobile/issues/3782, is complete.~~ Done!

Released 2020-07-06.

- [Release announcement](https://reactnative.dev/blog/2020/07/06/version-0.63)
- [Changelog](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0632)
- [Upgrade guide](https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.2) (the diff doesn't look huge, and big chunks of it make sense at a glance!!)

Looks like we'll get:

- LogBox, a nice-looking new way to see errors, warnings, and logs
- Pressable, a new API that helps React Native apps' interactions seem more purely native
- Native Colors (PlatformColor, DynamicColorIOS), an API for grabbing system-defined colors from the native platform
- Dropping iOS 9 and Node.js 8 support. Both of these are fine; we're targeting iOS 10.3 (see top of `ios/Podfile`), and we're on Node 12
- Changed iOS LaunchScreen from `xib` to `storyboard` (we did this in #2756)
- Flow v0.122 (see [changelog](https://github.com/facebook/flow/blob/master/Changelog.md#01220))

chrisbobbe:
We'll get at least one Flow bug fixed:

- https://github.com/facebook/flow/issues/8178 (see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3782.20RN.20v0.2E61.20-.3E.20v0.2E62.20upgrade/near/1006272))

chrisbobbe:
In RN v0.62, we started getting (and suppressing) this new error on calling various methods on a `TextInput` stored in a ref (e.g.,`textInputRef.current.focus()`):

```
Cannot call textInputRef.current.focus because:
 • Either property focus is missing in AbstractComponent [1].
 • Or property focus is missing in object type [2].
```

At first, I thought something was wrong with how we're annotating the variable storing the ref, or that Flow didn't fully understand the `React.createRef` API, which we started using in more places just before the RN v0.62 upgrade. Rather, it seems to be a known issue that wasn't observed on React Native's `master` as of 2020-04-06 [2]. Checking commits around that date, I'm pretty sure we'll have a fix in RN v0.63 (#4245).

[...]

PR #4247, opened 2020-09-02
RN v0.61 -> v0.62 upgrade

chrisbobbe:
~~To follow ~~#4244 and~~ the react-navigation v4 upgrade (#4248); see my note at https://github.com/zulip/zulip-mobile/issues/4248#issuecomment-689879451.~~

~~Currently a draft, since it's quite incomplete. But I'm sending this now so I can conveniently point to things in it.~~

Ready for review! 🙂 

Fixes: #3782


src/api/transport.js
```diff
@@ -3,5 +3,9 @@ import base64 from 'base-64';
...
 import type { Auth } from './transportTypes';
 
-export const getAuthHeaders = (auth: Auth): { [string]: string } =>
-  auth.apiKey ? { Authorization: `Basic ${base64.encode(`${auth.email}:${auth.apiKey}`)}` } : {};
+export const getAuthHeaders = (auth: Auth): {| Authorization?: string |} =>
+  // The redundant `Authorization` in the `:` case avoids a Flow issue:
+  // https://github.com/facebook/flow/issues/2386
+  auth.apiKey
+    ? { Authorization: `Basic ${base64.encode(`${auth.email}:${auth.apiKey}`)}` }
+    : { Authorization: undefined };
```
gnprice, line 11: Hmm, so this actually changes what gets returned -- instead of `{}` we return `{ Authorization: undefined }`. That *probably* doesn't ultimately matter, but that's not immediately obvious -- in at least one call site (in `apiFetch.js`) it propagates through to the calling function's own return value.

How about `Object.freeze({})` instead? I learned about that solution just now from browsing a couple of these issue threads:
https://github.com/facebook/flow/issues/2386#issuecomment-695064325

src/api/apiFetch.js
```diff
@@ -10,9 +10,11 @@ import { makeErrorFromApi } from './apiErrors';
 
 const apiVersion = 'api/v1';
 
-export const getFetchParams = <P>(auth: Auth, params: P): { headers: mixed, ...P } => {
-  // $FlowFixMe This is purely a no-op, and Flow even knows that. :-/
-  const { body } = (params: { body?: mixed });
+export const getFetchParams = <P: $Rest<$Exact<RequestOptions>, {| headers: mixed |}>>(
```
gnprice, line 13: `$Rest` is very similar to `$Diff`, and rereading what the difference (ha) is between them, I think `$Diff` is more on point here:
https://flow.org/en/docs/types/utilities/#toc-rest
because it's more purely "this object type, just without these particular properties".

src/api/apiFetch.js
```diff
@@ -10,9 +10,11 @@ import { makeErrorFromApi } from './apiErrors';
 
 const apiVersion = 'api/v1';
 
-export const getFetchParams = <P>(auth: Auth, params: P): { headers: mixed, ...P } => {
```
gnprice, line 13: It looks like we've gestured toward not allowing `params` to contain
    `headers`, so, make that explicit.

I think the old code doesn't so much gesture toward `params` not having that property, as say that the return value will have that property. P could have that property, and then the return type would just have it with the same type as P.

But certainly the actual interface of this function is that you can't include `headers`, and it'd break the functionality if you did. So, glad to have that expressed in the type. :slightly_smiling_face:

src/reactions/MessageReactionList.js
```diff
@@ -86,7 +86,7 @@ const getReactionsTabs = (
       // that's since been removed.  Ignore the nav hint in that case.
       ...(reactionName !== undefined && reactionsTabs[reactionName]
         ? { initialRouteName: reactionName }
-        : {}),
+        : { initialRouteName: undefined }),
```
gnprice, line 89: Perhaps `Object.freeze({})` here too?

Otherwise, this code would be simplified by inlining the spread and just having an `initialRouteName` property with a conditional expression inside its initializer.

.flowconfig
```diff
@@ -100,7 +100,8 @@ esproposal.optional_chaining=enable
 esproposal.nullish_coalescing=enable
 
 # The main warning is "Unused suppression comment".
-include_warnings=true
+# Temporarily disabled to smooth the RN upgrade.
+# include_warnings=true
```
gnprice: I think you want this a few commits earlier in the branch, before "flagsReducer: Add a $FlowFixMe for something Flow v0.113 will catch."

src/common/Input.js
```diff
@@ -8,10 +8,10 @@ import type { ThemeData } from '../styles';
 import { ThemeContext, HALF_COLOR, BORDER_COLOR } from '../styles';
 
 export type Props = $ReadOnly<{|
-  // TextInput's definition changes across the RN v0.61 -> v0.62
-  // upgrade; we'll handle that change after the upgrade.
+  // See point 2 at
+  // https://github.com/facebook/react-native/issues/28459#issuecomment-609957836
   // $FlowFixMe
```
gnprice: It takes a bit of unpacking to see from this that we expect this to go away with RN v0.63.

How about mentioning this in a comment on #4245, and then pointing there?

gnprice (commented):
Thanks @chrisbobbe ! This looks great. Small comments below.

.flowconfig
chrisbobbe: How about moving

"flagsReducer: Add a $FlowFixMe for something Flow v0.113 will catch."

later in the branch? I think this is the only one that's strayed away from the unused-fixmes-allowed window.

chrisbobbe: Oh, hang on—actually I think there's a subtlety that I didn't keep track of. If the Flow upgrade proceeded without the changes in

"flagsReducer: Add a $FlowFixMe for something Flow v0.113 will catch."

, there would be an error at a different site than where that `$FlowFixMe` is:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/chat/flagsReducer.js:48:10

Cannot return object literal because inexact object literal [1] is incompatible with exact FlagsState [2].

 [2] 33│ ): FlagsState => {
       :
     45│     });
     46│   });
     47│
 [1] 48│   return {
     49│     ...state,
     50│     ...newState,
     51│   };
     52│ };
     53│
     54│ const removeFlagForMessages = (state: FlagsState, messages: number[], flag: string): FlagsState => {
```

The way I suppress that error is to pretend that `{}` for `newState` is a `FlagsState`, which is a lie, and that lie gets caught even before v0.113.

chrisbobbe:
Thanks for the review! I just have one query, at https://github.com/zulip/zulip-mobile/pull/4247#discussion_r491217203. (Basically: should I change my strategy, or just describe it better?0

.flowconfig
gnprice: Mmm, I see.

Yeah, that strategy makes sense. In describing it, I'd just emphasize that you're casting these values, lying about their types. The change isn't primarily about adding a fixme (as it might be for covering an existing lie / type error that Flow is about to start spotting) as adding a new lie, with a fixme to cover it up.

chrisbobbe:
Thanks for the review! I just pushed my latest changes.

gnprice:
Looks good! Merged.

Issue #4246, opened 2020-09-02
Don't generate Jitsi URLs when not the selected provider


gnprice:
This is a followup to #3546.

Following #4229, we now have a "video call" button. When a Jitsi server URL is configured, we offer the button, and it inserts a Jitsi link.

There's more in the video-call config the server provides, though, and the intended behavior is more complicated than that.
* There are several other video-chat providers.
* Each can have its own set of details for how to generate a URL.
* But there's also some details that identify *which* provider is the one to be used. This choice is controlled by the realm/org admins. Quoting @chrisbobbe from https://github.com/zulip/zulip-mobile/issues/3546#issuecomment-597260426 :
  > * The choice of video call provider is made by the available and the selected video call providers: `.realm_available_video_chat_providers` and `.realm_video_chat_provider`, respectively, on `page_params`. On chat.zulip.org, Jitsi is available and selected. These params are made available in the `/register` response, so the mobile app has access to them.
  * (I missed this in my subsequent comment there.)
* The intended behavior is that only the selected provider will be used, even if config details exist for other providers.

Details of the webapp's logic are in static/js/compose.js .

I'm not excited about replicating the full logic, particularly for Zoom, where it's fairly messy. But as a minimal fix: when the selected provider isn't Jitsi, we should not offer the icon at all, rather than show it when it'll do the wrong thing.

wpj:
@gnprice just to clarify: is the intention here to disable the video call button when the selected video call provider is unimplemented (all but jitsi) as a short term fix and later implement the other providers?

gnprice:
> @gnprice just to clarify: is the intention here to disable the video call button when the selected video call provider is unimplemented (which is currently all of them except Jitsi) as a short term fix and later implement the other providers?

Yes, exactly. Thanks!

PR #4250, opened 2020-09-07
Hide video call button for unsupported video chat providers

wpj:
Clients that connect to a server that has configured a video call provider not currently supported by the mobile app should not display a video call button in the compose menu. This PR updates the realm reducer and compose box to disable the video call button for unsupported video chat providers (currently, only Jitsi is supported) and lays the groundwork for implementing other providers in the future.

Fixes: #4246


src/reduxTypes.js
```diff
@@ -247,7 +261,7 @@ export type RealmState = {|
   nonActiveUsers: User[],
   filters: RealmFilter[],
   emoji: RealmEmojiById,
-  jitsiServerUrl: string | null,
+  videoChatProvider: VideoChatProvider | null,
```
gnprice, line 264: I like this way of representing the data!

src/reduxTypes.js
```diff
@@ -215,6 +215,19 @@ export type PresenceState = {
...
 };
 
+/**
+ * Disjoint union (tagged by the `name` property) that represents the set of
+ * video chat providers supported by the mobile app. Each member of the union
+ * represents a distinct video chat provider and encapsulates the associated
+ * data for creating video call links for that provider. For example, the type
+ * for the "jitsi_meet" provider contains a field for the jitsiServerUrl
+ * returned by the Zulip server.
+ *
```
gnprice, line 225: A couple of bits of doc-comment style:
* Generally we start a doc comment with a short summary, just a line or so.

  The Dart project generally has excellent documentation, and they have a great style guide for it, almost all of which can apply just as well in JS (or really most any language.) [This item](https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary), and the next few after it, are on point here.

* For a type's doc comment, the summary focuses on a value of that type, rather than on the type definition. See examples elsewhere in this file; e.g. `MessagesState` is "A map with all messages we've stored locally, indexed by ID", because `MessagesState` is the type *of* values like that. (Subsequent paragraphs can focus on the type definition, in cases where that's helpful.)

src/realm/realmReducer.js
```diff
@@ -30,6 +35,34 @@ const initialState = {
...
+  providerId,
+}: {
+  availableProviders: AvailableVideoChatProviders,
+  jitsiServerUrl?: string,
+  providerId: number,
+}): VideoChatProvider | null {
+  if (
+    availableProviders.jitsi_meet &&
+    availableProviders.jitsi_meet.id === providerId &&
+    jitsiServerUrl != undefined
```
gnprice, line 58: If you run our test suites with `tools/test`, I think you'll see some lint errors here. You can fix some of them automatically with `--fix`; see `tools/test --help`.

(This should run automatically as CI when you send a PR, but our CI provider has been flaking on us lately; we should probably switch. That's #4174. Fortunately it's easy to run locally.)

src/realm/__tests__/realmReducer-test.js
```diff
@@ -25,7 +25,6 @@ describe('realmReducer', () => {
         user_id: undefined,
         isAdmin: false,
         twentyFourHourTime: false,
-        jitsiServerUrl: null,
```
gnprice, line 28: Hmm, this doesn't seem to be quite enough to update the tests -- see the results of running `tools/test`.

src/compose/ComposeBox.js
```diff
@@ -403,12 +409,16 @@ class ComposeBox extends PureComponent<Props, State> {
...
       isSubscribed,
-      jitsiServerUrl,
       stream,
+      videoChatProvider,
     } = this.props;
 
+    // Flow is erroneously widening the type of videoChatProvider to also allow
+    // `string` types here. The type is refined back down to the correct type by
+    // explicitly annotating `ComposeBox.prototype.insertVideoCallLink`, but be
+    // aware of that if you need to use this value outside of that method.
```
gnprice, line 419: Hmm, what's the type that you're seeing widened to `string`? When I remove the annotation on the argument type of `insertVideoCallLink`, and then hover on various tokens here (in VS Code) to see the types Flow infers, I don't spot any `string` types.

Over in the implementation of `insertVideoCallLink`, if I hover on the `name` in `videoChatProvider.name`, I do see `string | "jitsi_meet"`. That is puzzling. But it seems to be the same with or without that type annotation on the `videoChatProvider` argument. Fortunately it also doesn't seem to do any harm.

gnprice (commented):
Thanks @wpj ! Comments below.

src/compose/ComposeBox.js
wpj: Huh, that's puzzling indeed. I'm updating my comment to clarify that it's `videoChatProvider.name` that's being widened. One occasion where this would cause problems is if you were to do something like this:

```javascript
let isBigBlueButton = false;
if (videoChatProvider !== null) {
  // For the purposes of this example, the correct name for the below provider is `big_blue_button`.
  if (videoChatProvider.name === 'bigbluebutton') {
    isBigBlueButton = true;
  }
}
```

The above example passes flow's checker when done in the body of the render method, but strangely, it doesn't pass if moved into the body of `insertVideoCallLink`, despite the fact that the type widening appears to be happening there too, as you correctly pointed out.

src/reduxTypes.js
wpj: Thank you for those pointers! I've updated the doc comment to reflect those guidelines.

src/realm/__tests__/realmReducer-test.js
wpj: Hmm, I'd been running the test suite locally so I'm not sure how this slipped through, but I've updated to fix the test.

src/api/initialDataTypes.js
```diff
@@ -45,7 +49,7 @@ export type InitialDataRealm = {|
   realm_allow_message_deleting: boolean,
   realm_allow_message_editing: boolean,
   realm_authentication_methods: { GitHub: true, Email: true, Google: true },
-  realm_available_video_chat_providers: string[],
+  realm_available_video_chat_providers: AvailableVideoChatProviders,
```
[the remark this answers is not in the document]
wpj: That is a better prefix, I've updated the commit message, thank you!

src/reduxTypes.js
wpj: Excellent! Yeah, learning about disjoint unions in Flow and discriminated unions in Typescript was a life-changing moment for me.

gnprice:
Thanks again @wpj ! Merged, with a couple of small additional commits on top.

PR #4256, opened 2020-09-12
Add a clear button to the Topic input. Fixes #4192

### LINKED ISSUE #4192: Add a clear button to the Topic input of the compose field
I've found that it would be super helpful when composing a message to be able to quickly clear the 'Topic' input field during compose (and perhaps in other places), with an X button.

<details>
<summary>
In this image, the ideal would be to clear out the word 'Streams' from the topic box.
</summary>

</details>

I can see two ways of handling that within this codebase:

- https://reactnative.dev/docs/textinput.html#clearbuttonmode
  This has React Native provide its own clear button, but will only work on iOS.

- https://github.com/zulip/zulip-mobile/blob/be2663f239aa560513c89901588ab210e5d29c5e/src/common/InputWithClearButton.js
  This component is currently only used for search (and possibly scaled for it and the like). Perhaps this could be integrated with `common/Input.js`.

I'm happy to take a crack at this issue if it seems reasonable -- would love guidance on which of the above two approaches would be preferable to the project.

tekacs:


chrisbobbe:
Thanks for the PR! There's been some recent discussion about that particular prop on chat.zulip.org, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/RN.20TextInput.20.60clearButtonMode.60.20prop/near/1005175). For the moment, at least, we're not super satisfied with the quality of React Native's implementation. Would you be up for giving it a try with our custom `InputWithClearButton` component? I wouldn't say it's perfect, but we've got some screenshots posted at the CZO thread that suggest the clear button looks better, and in any case, it works on both iOS and Android, which is a win. 🙂

chrisbobbe:
Oh, and I see you already mentioned `InputWithClearButton` in your issue, [here](https://github.com/zulip/zulip-mobile/issues/4192#issue-658452666)! 😄 

I'd be interested to hear how using `InputWithClearButton` goes. 🙂

Issue #4257, opened 2020-09-17
Text in code blocks has a stripy inline background/padding/border


gnprice:
A few months ago I started noticing behavior like this whenever there's a code block:
![image](https://user-images.githubusercontent.com/28173/93419235-33098700-f861-11ea-99db-a68863ba36df.png)

That is, we're showing not only a block-level border around the whole code block (as intended), but also an inline border around just the text inside the block.  That means a border starts on the left of the text; runs above and below it to the end of just the text on the line; then the border lines break and resume at the start of the next line; etc., until wrapping around the right end of the text on the last line.  We also show a contrasting gray background enclosed by that border, just as we do for inline code snippets, giving the text in the code block a stripey appearance. Overall it looks pretty obviously broken, at least to my eyes.

I mentioned this in [a priorities conversation in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/priorities/near/995968) last month and said I'd file an issue for it; here it is. :slightly_smiling_face: 

I also just today dug in and diagnosed this and made a fix, which I'll post shortly.

PR #4258, opened 2020-09-17
msglist css: Fix inline border on text inside code blocks.

gnprice:
When the Markdown implementation on the server was upgraded in the
3.0-dev period, one subtle change was that code blocks started looking
like <pre><code>...</code></pre> instead of just <pre>...</pre>.

As a consequence, our CSS rules that were meant to apply to inline
code snippets started applying also to the text inside a code block.
Most conspicuously, we started showing not only a block-level border
around the whole code block (as intended), but also an inline border
around just the text inside the block.  That means a border starts on
the left of the text; runs above and below it to the end of just the
text on the line; then the border lines break and resume at the start
of the next line; etc., until wrapping around the right end of the
text on the last line.  We also showed a contrasting gray background
enclosed by that border, just as we do for inline code snippets,
giving the text in the code block a stripey appearance.

This fix is inspired by the webapp's changes that addressed the same
issue: zulip/zulip@43ac901ad and zulip/zulip@c00f626f8.

Fixes: #4257

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -39,6 +39,15 @@ code, pre {
   background-color: hsla(0, 0%, 50%, 0.125);
   font-family: Monaco, Menlo, Consolas, "Courier New", monospace;
 }
+pre code {
+  /* Starting with Zulip Server 3.0, code blocks have `pre > code`.
+     Undo the stuff we have for `code` that isn't meant to apply there. */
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Looks great to me on iOS, both light and dark mode! Thanks for the fix!

PR #4259, opened 2020-09-17
msglist css [nfc]: Group and reorder CSS rules for better readability.

gnprice:
For many aspects of the message list's layout and styling, the relevant CSS rules are scattered through the file. That can make it hard to get a clear picture of how things work, or to compare to their counterparts in the webapp where applicable (i.e. for message content.)

These changes add somewhat more order to the CSS rules. There's still plenty of opportunity to clean them up further.

I was moved to make these changes by working in the CSS for #4258, and so I developed them on top of that fix; it's the first commit of this branch.

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -1,3 +1,7 @@
+/*
+ Top level.
+ */
+
 html {
   -webkit-user-select: none; /* Safari 3.1+ */
   -moz-user-select: none; /* Firefox 2+ */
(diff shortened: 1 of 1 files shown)


src/webview/static/base.css
```diff
@@ -404,42 +404,32 @@ body {
 
 /*
  Message contents.
+
+ For the body of a message, we use nearly the same HTML as the webapp does.
+ The styles therefore have a lot in common.  For the webapp's version of
+ these styles, compare static/styles/rendered_markdown.scss .
```
chrisbobbe, line 410: .css, I think: https://github.com/zulip/zulip/commit/a3d26d701eafc8f4d2ad260764dee111225741c4

src/webview/static/base.css
```diff
@@ -517,17 +507,53 @@ blockquote {
...
+.highlight {
+  background-color: hsl(51, 94%, 74%);
+}
+
+/* `a`, which the webapp styles with Bootstrap. */
+a {
+  color: hsl(200, 100%, 40%);
+}
+
+/* `hr`, not found in the webapp's rendered_markdown.scss. */
```
chrisbobbe, line 549: (here, too, .css, I think)

chrisbobbe:
This looks good! I also tested it on iOS (light and dark mode) and I didn't notice any changes. There's just a small thing I noticed (see above) then this is good to merge.

src/webview/static/base.css
gnprice: Hmm indeed! Happened just this week, it looks like.

gnprice:
Thanks for the review!

Issue #4240, opened 2020-08-26
Android build: Address deprecation warning in `react-native-device-info`


chrisbobbe:
I see the following warning on running `./gradlew help --warning-mode=all` from `android` (recommended by [this doc](https://docs.gradle.org/5.6/userguide/upgrading_version_5.html#changes_5.6)):

```
> Configure project :react-native-device-info
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed soon. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html
```

It looks like this package is still being maintained, by react-native-community: https://github.com/react-native-community/react-native-device-info#readme. So, we might be able to upgrade to a version that doesn't have this warning, following their upgrade guides and handling any breaking changes.

Another route we could take is to get rid of this dependency and consider having [`expo-application`](https://docs.expo.io/versions/latest/sdk/application/) do its job, as we did partially in 18c37ce0f. [`expo-device`](https://docs.expo.io/versions/latest/sdk/device/) might be needed, too.

PR #4261, opened 2020-09-18
RN v0.61 -> v0.62 upgrade followup

chrisbobbe:
Fixes: #4240

Here is a small handful of commits to tie up some known loose ends after this upgrade (open as #4247, for #3782).

Then, starting from this commit:

4abb1e404 deps: Upgrade `react-native-unimodules` to ^0.10.1, the latest.

I've thrown in a few upgrades to some dependencies. I don't have any particular reason to think these can't go before the RN upgrade, but I thought I might as well not attempt them before the upgrade, and risk something going wrong and having to do them after the upgrade anyway. I can put these into another PR, if you'd like, @gnprice.


gnprice:
This all LGTM, thanks! Once #3782 is in, please merge at will.

I noticed in the jest-expo upgrade commit that it makes `yarn.lock` a lot longer (something like 1600 lines), much more than the commit it's reverting made it shorter. It looks like the reason, or a major reason, is that it pulls in Jest 25, whereas we're now on Jest 26. That's mildly annoying, but I guess it's harmless, and presumably with some future release of jest-expo it'll get reconciled.

chrisbobbe:
> This all LGTM, thanks! Once #3782 is in, please merge at will.

Thanks for the review! Done.

> I noticed in the jest-expo upgrade commit that it makes `yarn.lock` a lot longer (something like 1600 lines), much more than the commit it's reverting made it shorter. It looks like the reason, or a major reason, is that it pulls in Jest 25, whereas we're now on Jest 26. That's mildly annoying, but I guess it's harmless, and presumably with some future release of jest-expo it'll get reconciled.

Yeah, that's annoying. The latest release of jest-expo, 39.0.0, requires RN v0.63 (via its "react" peer-dep) and is still using Jest 25: https://github.com/expo/expo/blob/2196cebae55bda181ccceadec809942f51ee9e39/packages/jest-expo/package.json.

Issue #3780, opened 2020-01-09
Upgrade to Node v12


rk-for-zulip:
At present (2020-01), Node v12 is [the latest active LTS](https://nodejs.org/en/about/releases/). We should probably be ready to move up to v12 before v10 goes into "Maintenance" mode in 2020-04.

ray-kraesig:
Current status:
* **Linux:** I've been using Node 12 for the last month or so. Seems OK so far.
* **Windows:** Known to be blocked by facebook/metro#453.
* **macOS**: Untested; unknown.

gnprice:
> We should probably be ready to move up to v12 before v10 goes into "Maintenance" mode in 2020-04.

v10 will still be supported (as LTS) up to 2021-04-30. So I think it'd be OK to still be on it up to then, for as long as there's a reason to be.

Definitely it's good to be on current versions of tools when we can, though. So once that blocker on Windows is resolved, we should upgrade unless there's some other blocker in the way.

gnprice:
Also, as you may well know but for concreteness here in the thread: what it means that we're currently on Node v10 is
* we use Node 10 in CI (see `.travis.yml`)
* we tell people to use Node 10 in the setup instructions (`docs/howto/build-run.md`)
* if people have build trouble and turn out to be using some other version of Node, we ask them to please use Node 10 instead.

So upgrading is a matter of confirming that everything works on the new version, and then updating those two places.

gnprice:
> * but that probably means we'll need to update to React Native 0.62

I went and filed #3782 for that (and #3781 for the upgrade before it), so as to have a central name for it.

Because the comment was labeled "Current status", I took the liberty of updating it accordingly, wiki-style.

gnprice:
We learned today ([chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/SOLVED.3A.20Prettier.20auto-formatting/near/896340)) about one major improvement that comes with Node 12: the startup time goes from painful (like 350ms for `node -e 1` in my quick testing) to in the range you'd expect (~50ms; vs 40ms `python -c 1` or 70ms `ruby -e 1`.) So that would be very nice -- everything we do from a script will get significantly faster!

[...]

PR #4262, opened 2020-09-18
deps: Upgrade to Node v12, in setup docs and in CI.

gnprice:
We've been using this for local development for a little while,
and all seems well.

The major visible benefit to this upgrade is that startup times are
much improved:
  https://github.com/zulip/zulip-mobile/issues/3780#issuecomment-639804863
which means everything we do from a script becomes much faster.

Fixes: #3780

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ notifications:
     on_failure: always
 
 install:
-  - nvm install 10
+  - nvm install 12
   - node --version
   - npm i yarn -g
   - yarn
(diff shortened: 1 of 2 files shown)


gnprice:
I plan to merge this as soon as the Travis build succeeds:
https://travis-ci.org/github/zulip/zulip-mobile/builds/728452817

One of the changes here is specifically to our Travis config, so the only reliable way to test that is to actually push it. (It's rare that we make changes to that file.)

chrisbobbe:
Great, thanks!

gnprice:
CI passed; merged.

PR #4265, opened 2020-09-22
Fix a regression from #4235.

chrisbobbe:
There was an accidental duplication of the trailing slash on the realm in all API requests, starting in 865914f0e. So, fix that and make some more progress toward #4146.


gnprice:
Thanks! About to merge.

    `isValidUrl` failed to alert us of a double-forward-slash issue in
    this string concatenation: the realm gained a trailing slash in its
    new spelling as `realm.origin` from 865914f0e, and we kept the
    leading slash in what we were adding to the end of the realm. This
    is a sign that `isValidUrl` should be retired ASAP; we'll do that in
    an upcoming commit.

I think this isn't a problem with `isValidUrl` (although still good to retire it now that we have `URL`) -- the resulting URL is a perfectly valid URL, and `new URL` agrees. It just isn't the URL we intended, and the server doesn't understand it as pointing to the API route we wanted.

```
$ node
> (new URL('https://example.com//foo///bar')).toString()
'https://example.com//foo///bar'
```

chrisbobbe:

gnprice:
It seems likely that this line has the same issue, then, in the Apple auth flow:

```
+    // TODO: Use a `URL` computation, for #4146.
     openLink(`${this.props.realm.toString()}/complete/apple/?${params}`);
```

chrisbobbe:
Ah, yeah, probably—testing with a fix now.

chrisbobbe:
> Ah, yeah, probably—testing with a fix now.

OK, just sent #4266.

PR #4266, opened 2020-09-23
Fix a few more regressions from #4235.

chrisbobbe:
Explanations in the commit messages.

For the second (doubled slashes in Apple auth endpoint), following Greg's comment at https://github.com/zulip/zulip-mobile/pull/4265#issuecomment-697040991—chat.zulip.org seems perfectly happy to accept the double-slash version, like it did before cfebe3f53752968235e33ee428e2239025b5c11c 😆. At first I was confused about why it seemed to work both before and after my commit here. Anyway, that's probably why.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -619,7 +619,16 @@ const eventUpdateHandlers = {
 const handleMessageEvent: MessageEventListener = e => {
   scrollEventsDisabled = true;
   const decodedData = decodeURIComponent(escape(window.atob(e.data)));
-  const updateEvents: WebViewUpdateEvent[] = JSON.parse(decodedData);
+  const rawUpdateEvents = JSON.parse(decodedData);
+  const updateEvents: WebViewUpdateEvent[] = rawUpdateEvents.map(updateEvent => ({
+    ...updateEvent,
+    // A URL object doesn't round-trip through JSON; we get the string
+    // representation. So, "revive" it back into a URL object.
+    ...(updateEvent.auth
+      ? { auth: { ...updateEvent.auth, realm: new URL(updateEvent.auth.realm) } }
+      : {}),
+  }));
+
   updateEvents.forEach((uevent: WebViewUpdateEvent) => {
     eventLogger.maybeCaptureInboundEvent(uevent);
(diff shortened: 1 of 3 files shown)


src/webview/js/js.js
```diff
@@ -619,7 +619,16 @@ const eventUpdateHandlers = {
...
-  const updateEvents: WebViewUpdateEvent[] = JSON.parse(decodedData);
+  const rawUpdateEvents = JSON.parse(decodedData);
+  const updateEvents: WebViewUpdateEvent[] = rawUpdateEvents.map(updateEvent => ({
+    ...updateEvent,
+    // A URL object doesn't round-trip through JSON; we get the string
+    // representation. So, "revive" it back into a URL object.
+    ...(updateEvent.auth
+      ? { auth: { ...updateEvent.auth, realm: new URL(updateEvent.auth.realm) } }
+      : {}),
+  }));
```
gnprice, line 630: Hmm, kind of ugly, isn't it. :-/ Seems fine for a workaround, and I don't have any immediate ideas for a way to do this (and the similar spot in `handleInitialLoad`) more cleanly without a fair amount more code; but we'll want to keep an eye on this area whenever adding more complexity to it.

gnprice (commented):
Thanks for the fixes! Merging.

src/webview/js/js.js
chrisbobbe: Yeah, my thoughts exactly. 😦

gnprice:
> chat.zulip.org seems perfectly happy to accept the double-slash version, like it did before [cfebe3f](https://github.com/zulip/zulip-mobile/commit/cfebe3f53752968235e33ee428e2239025b5c11c) . At first I was confused about why it seemed to work both before and after my commit here. Anyway, that's probably why.

Hmm! Now that you mention that, I do seem to recall finding that some classes of server API routes tolerated duplicate slashes, while others didn't, just because they went through different pieces of infrastructure code that handled that case differently. The web auth flows certainly use some distinctive pieces of infrastructure code, so I guess it fits that they might be in the group that tolerates such variations in the URL while our main API endpoints aren't.

Issue #3804, opened 2020-01-15
Decouple navigation from Redux


chrisbobbe:
This won't be a priority until we're closer to upgrading to React Navigation v5 (which currently hasn't been released yet, and we're on v2 now), but React Navigation will eventually stop supporting the pattern of storing navigation state in Redux. Here are the indications:

* The [v2 Redux integration doc](https://reactnavigation.org/docs/en/2.x/redux-integration.html) (the version we're on), rather colorfully: "**Warning: in the next major version of React Navigation, to be released in Fall 2018, we will no longer provide any information about how to integrate with Redux and it may cease to work**", and "_Think twice before you consider doing this, there is an incredibly good chance that you do not need to do this!_"
* The [v3 Redux integration doc](https://reactnavigation.org/docs/en/3.x/redux-integration.html#can-i-store-the-navigation-state-in-redux-too), a little less frantic but still firm: "This is technically possible, but we don't recommend it - it's too easy to shoot yourself in the foot and slow down / break your app. We encourage you to leave it up to React Navigation to manage the navigation state."
* The [v4 Redux integration doc](https://reactnavigation.org/docs/en/4.x/redux-integration.html): "404: File not found"
* A [draft of a v4 -> v5 upgrade doc](https://reactnavigation.org/docs/en/next/upgrading-from-4.x.html#navigation-state-in-redux): "We have long recommended not to store navigation state in Redux. We have finally dropped support for storing navigation state in Redux in React Navigation 5.x."

See the discussion at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/redux.20.2F.20navigation/near/808659.

In the meantime, this issue is a place to document what this coupling currently provides for us, so we don't forget about it when we eventually have to make the change. Since this pattern works for us, currently, incremental changes will be favored over large, dramatic ones.

ray-kraesig:
Possibly also relevant: the existing "upgrade to v3" issue, #3649.

chrisbobbe:
As @gnprice mentions [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3745.20Server.20version.20to.20Sentry/near/930326), fixing this issue may help us in a somewhat surprising way: it appears that many of our import cycles go through `navReducer.js`. When there are too many import cycles, Flow and Jest sometimes panic and spit out lots of error output.

[...]

PR #4269, opened 2020-09-24
ChatScreen: Get `isFocused` from React Navigation.

chrisbobbe:
(Moves us closer toward https://github.com/zulip/zulip-mobile/issues/3804.)

-----

As noted in discussion [1], we haven't been using quite the right
information to decide whether it's time for a ChatScreen to fetch
new messages. As Greg says, there,

"""
[T]he condition we'd like to have governing this isn't "is this
ChatScreen showing the same narrow that's topmost or last", but
rather "is this the topmost or last ChatScreen".
"""

To grab that bit, we first considered [2] something like adding a
unique ID to each ChatScreen, which seems hacky.

But while I was doing the recent React Navigation v3 and v4
upgrades, I came across this `withNavigationFocus` HOC, which
provides an `isFocused` prop [3], which looks more or less exactly
like what we'd want. "Is this ChatScreen focused" isn't *exactly*
the same question as "is this the topmost or last ChatScreen", but
the differences seem trivial, and I think this commit's approach is
simpler and clearer.

There's also an `isFocused` *method* directly on the `navigation`
prop [4]; we get the `navigation` prop for free [5] since we've
stuck `ChatScreen` directly on one of our navigators. The problem
with just using that method is that `ChatScreen` doesn't naturally
update (re-`render`) when it goes in and out of focus, so our code
in `componentDidUpdate` that would ask for the `isFocused` value
would never get run.

So, use this HOC, which does set up `ChatScreen` to listen for a
change of focus.

[1] https://github.com/zulip/zulip-mobile/pull/4205#pullrequestreview-461268523
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/978256
[3] https://reactnavigation.org/docs/4.x/function-after-focusing-screen#triggering-an-action-with-the-withnavigationfocus-higher-order-component
[4] https://reactnavigation.org/docs/4.x/navigation-prop/#isfocused---query-the-focused-state-of-the-screen
[5] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4156.20Message.20List.20placeholders/near/953303

### DIFF
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -2,7 +2,9 @@
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
 import type { NavigationScreenProp } from 'react-navigation';
+import { withNavigationFocus } from 'react-navigation';
 import { ActionSheetProvider } from '@expo/react-native-action-sheet';
+import { compose } from 'redux';
 
 import { connect } from '../react-redux';
 import type { ThemeData } from '../styles';
@@ -19,7 +21,6 @@ import ComposeBox from '../compose/ComposeBox';
 import UnreadNotice from './UnreadNotice';
 import { canSendToNarrow } from '../utils/narrow';
 import { getLoading, getSession } from '../directSelectors';
-import { getTopMostNarrow } from '../nav/navSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow } from './narrowsSelectors';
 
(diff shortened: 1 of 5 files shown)


chrisbobbe:
I've tested on iOS and Android; all seems well.

src/boot/store.js
```diff
@@ -199,6 +199,9 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
     })),
   }),
 
+  // Drop `session.lastNarrow`.
+  '16': dropCache,
+
```
gnprice, line 204: The `session` subtree isn't persisted at all, so no migration is needed.

gnprice (commented):
LGTM! Good find. Just a small comment below.

The one thing that's nicer about using "topmost ChatScreen" vs. "ChatScreen which is the topmost screen" is that it can let us start the fetch a bit sooner, when it's very likely we'll want the fetch's results soon.

But I think that difference is pretty immaterial for us in practice. For one thing, this is just not that common a case to be in -- you have to have lost the old event queue, typically from having the app in the background for ≥ 10 minutes, but OTOH still have the app actually in the background (vs. killed by the OS to free up memory), and also you have to be actually still interested in one of the ChatScreens that was on your nav stack (vs. going straight back to the main screen.) So it wouldn't be worth a lot of effort (or complexity) to try to finely optimize its timing; the only important thing is that the behavior is correct.

If there were a good clean handy way to identify "topmost ChatScreen", then that'd be nice; but unsurprisingly there doesn't seem to be, so it's not worth the effort it'd take to do that instead of "topmost screen".

chrisbobbe:

Issue #4270, opened 2020-09-25
Crash on sending message


gnprice:
In a release build of v27.155 -- the release I tagged yesterday but haven't yet rolled out even to beta -- I had the app crash shortly after I sent a message.

See investigation [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/crash.20at.20.2EgetWrappedInstance%28%29/near/1024873); I pulled the uncaught exception from the log, and we seem to have a diagnosis.

chrisbobbe:
Thanks for catching this! I just sent #4271; I'll merge it in ten minutes or so (or at your approval) unless you'd like to change anything.

PR #4271, opened 2020-09-25
ComposeBox: Fix a crash on sending a message.

chrisbobbe:
This is a regression caused by something that was missed in
fd4debbae; we partly but incompletely removed our use of
`getWrappedInstance` with a react-redux API change. From that commit
message:

"""
So, make that change, including the implied removal of
`getWrappedInstance`; that bit of UI works fine on Android and iOS
after that removal.
"""

It looks like "that bit of UI" was something other than sending a
message, and I didn't do a full-project search for the term
`getWrappedInstance` in that revision. This should be the last of
them now.

Fixes: #4270


gnprice:
Looks good, thanks!

PR #4273, opened 2020-10-01
Prep work toward decoupling navigation from Redux.

chrisbobbe:
Some preparation for #3804.


src/account-info/AccountDetailsScreen.js
```diff
@@ -30,7 +30,9 @@ type SelectorProps = $ReadOnly<{|
 |}>;
 
 type Props = $ReadOnly<{|
-  navigation: NavigationScreenProp<{ params: {| userId: number |} }>,
+  // Since we've put this screen in a stack-nav route config, it gets the
+  // `navigation` prop for free.
+  navigation: NavigationStackProp<{| ...NavigationStateRoute, params: {| userId: number |} |}>,
```
gnprice, line 35: The main example in the TypeScript doc is the following:
    
    ```
    type Props = {
      navigation: NavigationStackProp<{ userId: string }>;
    };
    ```
    
    We find `NavigationStackProp` is a good thing to use, but
    `{ userId: string }` is bad for a couple of reasons:
    
    - It's off by a level of nesting; surely they mean to describe
      `navigation.state.params`. To do that, the type needs to look
      something like `{ params: { userId: string } }`.
    
    - It leaves out all the other things on `navigation.state` including
      `key` and `routeName`, which might be nice to have someday.
      Experimentally, these are conveniently filled in by saying
      `NavigationStackProp<NavigationStateRoute>` [3].

Yeah, that's odd!

I wondered if this might be a difference between the TS definitions and the Flow libdef, in what they have `NavigationStackProp` take as a type parameter. But AFAICT it's equally wrong there -- the first type parameter is called `State` and is used for the `state` property, not `state.params`. (Though there is a *second* type parameter for the latter.)

src/account/AccountPickScreen.js
```diff
@@ -11,6 +12,10 @@ import AccountList from './AccountList';
 import { navigateToRealmScreen, accountSwitch, removeAccount } from '../actions';
 
 type Props = $ReadOnly<{|
+  // Since we've put this screen in a stack-nav route config, it gets the
+  // `navigation` prop for free.
```
gnprice, line 16: IIUC the precise logic is: because the *only* way we invoke this component is as a screen in such a route config, it will always get that prop.

Or I guess, equally correct and broader: because that's the only *non-type-checked* way we invoke it.

Typically I wouldn't expect us to use one of these components in any other way. But as a way of helping confirm that, I did this quick search: `rg -w '[A-Z]\w+Screen' src/` to find all references to things that fit the naming pattern common to almost all these components. Other than their own definitions, including react-redux `connect` calls, the only ones I found referred to anywhere other than `AppNavigator` were:
* `CompatibilityScreen`
* `ChooseRecipientsScreen`

And indeed those don't appear in this commit, so that checks out.

I repeated the search on the components that appear in this commit and don't fit that naming pattern: `rg -we SmartUrlInput -e MainScreenWithTabs -e MessageReactionList src/`

That actually found one reference that doesn't line up! Did you mean to include `SmartUrlInput`?

src/common/SmartUrlInput.js
```diff
@@ -45,7 +49,7 @@ type Props = $ReadOnly<{|
    * it appears not to contain an explicit domain.
    */
   defaultDomain: string,
-  navigation: NavigationScreenProp<mixed>,
+  navigation: NavigationScreenProp<NavigationState>,
```
gnprice, line 52: I think this component may not belong in this commit -- see previous comment.

But also, the type here appears to vary from the pattern of the other changes.

src/sharing/ShareToPm.js
```diff
@@ -54,7 +54,13 @@ const styles = createStyleSheet({
 });
 
 type Props = $ReadOnly<{|
-  ...$Exact<NavigationNavigatorProps<{||}, {| params: {| sharedData: SharedData |} |}>>,
+  // Since we've put this screen in a tab-nav route config, it gets the
+  // `navigation` prop for free.
+  navigation: NavigationTabProp<{|
+    ...NavigationStateRoute,
+    params: {| sharedData: SharedData |},
+  |}>,
```
gnprice, line 62: Hmm -- for some reason it seems like this type annotation is not working (it's resulting in `any` types), whereas in at least some spot-checking the ones from the previous commit with `NavigationStackProp` are working.

Specifically, if I go and do something silly like
```diff
   state = (() => {
-    const { sharedData } = this.props.navigation.state.params;
+    const { sharedData, asdfasdf } = this.props.navigation.state.params;
```

then I don't get any type error at the attempt to pull an `asdfasdf` property from `params`.

Also, hovering on the identifier `navigation` in that line I just quoted shows "any(implicit)" as the type, and ditto for the subsequent `state` or `params`. (That's the symptom I first noticed; I tried the `asdfasdf` thing just to confirm that it wasn't some IDE glitch.)

Perhaps relatedly, if in VS Code I hit F12 on `NavigationTabProp` to try to go to its definition, it doesn't find it -- it can only take me to the `import type` line at the top of this file. Whereas it does work fine for `NavigationStackProp` in a file like `AccountDetailsScreen.js`. There *does* seem to be a libdef in `flow-typed/`, so not sure what the issue is.

gnprice (commented):
Thanks! This direction makes sense to me. Some comments below -- notably I think we should be able to simplify the "wait until the store is hydrated" part. Which IIUC runs through these three commits:
844646a4e messagesActions: Don't let `doNarrow` cancel an expected navigation (2/2).
dfac22abb AppEventHandlers: Make `handleInitialShare` wait for rehydration.
3a90d903d AppEventHandlers: Use `handleInitialCare` correctly.

src/chat/InvalidNarrow.js
```diff
@@ -0,0 +1,34 @@
...
+
+type Props = $ReadOnly<{|
+  narrow: Narrow,
+|}>;
+
+export default class InvalidNarrow extends PureComponent<Props> {
+  render() {
+    return (
+      <View style={styles.container}>
+        <Label style={styles.text} text="Invalid narrow." />
```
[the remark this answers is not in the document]
gnprice: > Also I think the term "narrow" may not be helpful -- it's a bit of an internal term. It doesn't seem to appear at https://zulipchat.com/help/ , for example.

See also d1acd6636 and b0203671a, which resulted from me looking at this just now :slightly_smiling_face: -- we don't currently use the term in the mobile app's UI at all, but until those commits we had it in a couple of strings for translation which we weren't using but were accidentally still around.

src/boot/AppEventHandlers.js
```diff
@@ -121,9 +123,33 @@ class AppEventHandlers extends PureComponent<Props> {
     // Release memory here
   };
 
-  componentDidMount() {
+  /**
+   * Do one-off tasks that require persisted data to have loaded.
```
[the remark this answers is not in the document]
chrisbobbe: > Hmm dang -- did you observe this crash empirically?

I did observe a crash while investigating my suspicion of a crash, but I'm not reproducing it right now; I'll try more tomorrow?

> The existing strategy we have that should prevent this kind of thing is, we use `redux-action-buffer` to cause all actions to be held up in a queue until a REHYDRATE action comes along.

True, but I was specifically thinking to prepare for the future when navigation actions aren't dispatched through Redux and `redux-action-buffer` will have no power over them. But you're right; now I think more closely about it, being in a thunk action creator should do the trick (and it seems to in my testing just now).

src/account-info/AccountDetailsScreen.js
chrisbobbe: Right, I did see some wrong things in the TypeScript (and noted them briefly in the commit message) but didn't record any specifics like this.

src/account/AccountPickScreen.js
chrisbobbe: Ah—SmartUrlInput should probably go in its own commit. It's a reusable component that expects...*a* `navigation` prop, which it doesn't get for free (from being plugged into a navigator's route config) but it shouldn't care whether that `navigation` prop comes from a stack navigator, drawer navigator, etc. I landed on `NavigationScreenProp<NavigationState>` without much consideration from https://github.com/react-navigation/react-navigation/issues/3643#issuecomment-370044086, and it could be wrong.

src/common/SmartUrlInput.js
chrisbobbe: Ah, right; I just responded above, at https://github.com/zulip/zulip-mobile/pull/4273#discussion_r498003062.

src/sharing/ShareToPm.js
chrisbobbe: I have a hunch that I'll look into tomorrow—we've got two libdefs for `react-navigation-tabs` in `flow-typed/npm`, and one of them is a stub where everything is `any`.

chrisbobbe: Aha...that's it. Yeah, getting rid of that stub libdef (which we don't want anyway) reveals lots of Flow errors that seem productive to address.

src/nav/navSelectors.js
```diff
@@ -8,15 +8,16 @@ export const getNav = (state: GlobalState): NavigationState => state.nav;
...
+  getNavigationRoutes(state)[getNavigationIndex(state)];
+
+export const getCurrentRouteName = (state: GlobalState): string => getCurrentRoute(state).routeName;
 
 export const getCurrentRouteParams: Selector<void | { narrow?: Narrow }> = createSelector(
-  getNavigationRoutes,
-  getNavigationIndex,
-  (routes, index) => routes[index] && routes[index].params,
+  getCurrentRoute,
+  currentRoute => currentRoute.params,
```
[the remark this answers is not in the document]
chrisbobbe: Ah, right, sounds good. I've also added the same check in `getCurrentRouteName`, where it wasn't before.

src/actionConstants.js
```diff
@@ -15,7 +15,6 @@ export const HTTP_UNAUTHORIZED: 'HTTP_UNAUTHORIZED' = 'HTTP_UNAUTHORIZED';
 export const LOGOUT: 'LOGOUT' = 'LOGOUT';
 
 export const REALM_INIT: 'REALM_INIT' = 'REALM_INIT';
-export const DO_NARROW: 'DO_NARROW' = 'DO_NARROW';
```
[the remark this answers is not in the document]
chrisbobbe: Ah, right, thanks.

src/boot/AppEventHandlers.js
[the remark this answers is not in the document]
chrisbobbe: > > Hmm dang -- did you observe this crash empirically?
> 
> I did observe a crash while investigating my suspicion of a crash, but I'm not reproducing it right now; I'll try more tomorrow?

OK, I think I've solved this mystery—I was logged out when I observed the crash. If I set it up so that `handleInitialNotification` dispatches `doNarrow` unconditionally (I had it do so before `await getInitialNotification()`), it *does* crash if the user is logged out; I get the "no server data" message. No real surprise there.

But this case isn't representative—`narrowToNotification` will handle checking to see that the account the notification is meant for is logged in.

chrisbobbe:
Thanks for the review! Just pushed some changes and responded to your comments above.

src/start/RealmScreen.js
```diff
@@ -17,7 +17,7 @@ type SelectorProps = {|
 
 type Props = $ReadOnly<{|
   navigation: NavigationScreenProp<{
-    params: ?{|
+    params: {|
```
chrisbobbe, line 25: e1922546c RealmScreen: Assume `navigation.state.params` will exist.

Oops! This commit is wrong; please don't merge until I've pushed a fix. I forgot about reaching RealmScreen through `navReducer`.

chrisbobbe: OK, done.

src/boot/AppEventHandlers.js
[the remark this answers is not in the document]
gnprice: Great -- glad that's diagnosed.

src/sharing/ShareToPm.js
gnprice: Aha! Good catch.

src/start/RealmScreen.js
```diff
@@ -17,7 +17,7 @@ type SelectorProps = {|
 
 type Props = $ReadOnly<{|
   navigation: NavigationScreenProp<{
-    params: ?{|
+    params?: {|
```
gnprice, line 25: Why optional at all? I think the logic in your previous version (at 6f6a3ee4a) was sound for why it shouldn't simply go missing; and following that logic we make `params` non-optional on other screens too.

src/common/SmartUrlInput.js
gnprice, line 52: Not a big deal, but I suspect you can make this change *after* the "Tell all screens on AppNavigator" commit (and include the `RealmScreen` change in that one.) If you do, then I think the story to tell in this one's commit message becomes simpler, because no foreshadowing is necessary.

src/start/RealmScreen.js
gnprice: Ah I see, you answered that here :slightly_smiling_face: : https://github.com/zulip/zulip-mobile/pull/4273#discussion_r499095804

I think this calls for a comment saying why it's optional, in any case -- since that is out of the ordinary compared to how all, I think, of our other screens work.

... Then I went reading this code some more to see how it handles the case where `params` is missing. OK, effectively `realm` gets defaulted to `''`, but `initial`? It seems to default to `false`. So when is it ever true -- is that just dead code? ... No, it's not. The answer is this hack in `navReducer.js`:
```js
export const getStateForRoute = (route: string): NavigationState => {
  // TODO: this is kind of a hack!  Refactor to a better way.
  //  * Perhaps pass `initial: true` unconditionally in this initialization
  //    code, to all routes?  Then that'd just be part of the interface of
  //    making a screen work as an initial screen.
  //  * Alternatively, we could replace the whole system of `ModalNavBar`,
  //    `canGoBack`, etc., with our own "navigation view" that would be more
  //    directly integrated into the navigation framework:
  //      https://reactnavigation.org/docs/en/2.x/navigation-views.html
  const params = route === 'realm' ? { initial: true } : undefined;
```

So in fact in `navReducer` when we navigate to this screen, we *also* pass a `params` object (which I guess means we always do.) It has a property `initial: true`, and no `params`.

I guess now this extra definitely calls for a comment here :stuck_out_tongue: Which really I should have added at the same time as adding that hack: 7a04a5638

gnprice (commented):
Thanks for the revisions!

This all LGTM. A few comments below (and above, thanks GitHub) on small things; please merge at will after going through those.

src/start/RealmScreen.js
chrisbobbe: Yeah, I've already got plans to get rid of that hack soon, as part of the process :)

src/common/SmartUrlInput.js
chrisbobbe: I get an error if I leave this as `mixed` with the `RealmScreen` change applied:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/start/RealmScreen.js:121:10

Cannot create SmartUrlInput element because in property navigation:
 • Either cannot create exact type from mixed [1].
 • Or property getParam is missing in object type [2] but exists in NavigationScreenProp [3].

        src/start/RealmScreen.js
        118│         shouldShowLoadingBanner={false}
        119│       >
        120│         <Label text="Enter your Zulip server URL:" />
        121│         <SmartUrlInput
        122│           style={styles.input}
        123│           navigation={navigation}
        124│           defaultProtocol="https://"
        125│           defaultOrganization="your-org"
        126│           defaultDomain="zulipchat.com"
        127│           defaultValue={initialRealm}
        128│           onChangeText={this.handleRealmChange}
        129│           onSubmitEditing={this.tryRealm}
        130│           enablesReturnKeyAutomatically
        131│         />
        132│         {error !== null ? (
        133│           <ErrorMsg error={error} />
        134│         ) : (

        flow-typed/npm/react-navigation-stack_v1.x.x.js
    [2] 595│     & {
        596│         pop: (n?: number, params?: {| immediate?: boolean |}) => boolean,
        597│         popToTop: (params?: {| immediate?: boolean |}) => boolean,
        598│         push: (
        599│           routeName: string,
        600│           params?: NavigationParams,
        601│           action?: NavigationNavigateAction
        602│         ) => boolean,
        603│         replace: (
        604│           routeName: string,
        605│           params?: NavigationParams,
        606│           action?: NavigationNavigateAction
        607│         ) => boolean,
        608│         reset: (actions: NavigationAction[], index: number) => boolean,
        609│         ...
        610│       };

        src/common/SmartUrlInput.js
 [3][1]  52│   navigation: NavigationScreenProp<mixed>,
```

chrisbobbe: I could do a temporary `$FlowFixMe`, but I wonder if that would mean roughly the same amount of explanation that's already in this commit?

chrisbobbe:
> Not a big deal, but I suspect you can make this change _after_ the "Tell all screens on AppNavigator" commit (and include the `RealmScreen` change in that one.) If you do, then I think the story to tell in this one's commit message becomes simpler, because no foreshadowing is necessary.

Since you said it's not a big deal, I'll go ahead as-is, if that's OK; see the error in https://github.com/zulip/zulip-mobile/pull/4273#discussion_r499905037 that I got when I tried the suggested adjustment.

src/common/SmartUrlInput.js
gnprice: Huh. That's a bit puzzling to me: it feels like it should be perfectly fine, and also when I look at the actual type definitions it seems like it should be accepted. In particular, the new type `NavigationScreenProp<NavigationState>` should be a subtype of the old type `NavigationScreenProp<mixed>` -- `NavigationState` is a subtype of `mixed` because everything is, and `NavigationScreenProp` is declared in the react-navigation libdef as covariant (with a `+`) in its type parameter:
```js
  declare export type NavigationScreenProp<+S> = {
```
so S a subtype of T should make `NavigationScreenProp<S>` a subtype of `NavigationScreenProp<T>`.

But not worth trying to debug further, since you've already made a fine workaround.

PR #4274, opened 2020-10-06
Decouple Navigation from Redux (Part 1)

chrisbobbe:
Contributing to https://github.com/zulip/zulip-mobile/issues/3804.

Following #4273, which was merged today.

Some interesting things happening in this PR:

- Add a [`NavigationService`](https://reactnavigation.org/docs/4.x/navigating-without-navigation-prop) so we can talk directly to React Navigation instead of going through Redux
- Use the `NavigationService` instead of Redux to read the navigation state, in all but a very few cases that will disappear naturally with further work
- Transplant logic in our helpers in `navReducer` that handle some of our custom actions, i.e., actions that don't look like `Navigation/POP` and friends.

While my observations suggest that actions dispatched using `NavigationService` still affect `state.nav` in the ways we'd want them to [1], it seems safest to minimize the places where we read from the `state.nav` source before we ~~stop updating it~~ potentially alter how we update it to an extent that isn't known.

I take it that `NavigationService` is a safe source for the navigation state because

- we've been quite comfortable using the `navigation` prop to read the navigation state, even with our react-navigation-redux-helpers setup; we use `props.navigation.state.params` quite often
- the docs [suggest](https://reactnavigation.org/docs/4.x/navigating-without-navigation-prop) that `NavigationService` can reasonably be considered a way to grab things from the `navigation` prop
- empirically, I haven't yet seen `NavigationService` give the wrong nav state

The next main piece of work after this PR will be to use `NavigationService` to dispatch navigation actions, then it should be a pretty easy path to removing our react-navigation-redux-helpers setup completely.

[1] Also, an example app with react-navigation v2 and react-navigation-redux-helpers seems just as happy to dispatch navigation actions with the [`navigation` prop](https://github.com/react-navigation/react-navigation/blob/2.x/examples/ReduxExample/src/components/LoginScreen.js#L28) (which is interesting for us and our `NavigationService`) as they are to do so with [Redux's `dispatch`](https://github.com/react-navigation/react-navigation/blob/2.x/examples/ReduxExample/src/components/AuthButton.js#L27).


src/nav/navActions.js
```diff
@@ -13,11 +13,35 @@ import { getSameRoutesCount } from '../selectors';
...
 export const navigateBack = (): NavigationAction => StackActions.pop({ n: getSameRoutesCount() });
 
-// Other stack routes reached through `navReducer`:
-//    StackActions.push({ routeName: 'loading' });
-//    StackActions.push({ routeName: 'realm' });
-//    StackActions.push({ routeName: 'account' });
-//    StackActions.push({ routeName: 'main' });
+/**
+ * "Reset" actions, to explicitly prohibit back navigation.
+ */
```
gnprice, line 18: nit: plain non-jsdoc comment, because this doesn't attach to a particular single item that comes just after it.

(As is, this will appear when hovering on a call site of `resetToLoading` in an IDE, because that happens to be the first thing after this.)

src/start/LoadingScreen.js
```diff
@@ -15,15 +19,95 @@ const styles = createStyleSheet({
...
+    this.doInitialNavigation();
+  };
+
+  /**
+   * Data has been loaded, so open the app to the right screen.
+   *
+   * Not to be called before the REHYDRATE action, and not to be
+   * called more than once.
+   */
+  doInitialNavigation = () => {
```
gnprice, line 59: (We discussed this commit today by video chat. One thing we figured out is that the use of a ref on the navigator means that it belongs in an ancestor component of the navigator, not a descendant like `LoadingScreen`.)

src/start/AuthScreen.js
```diff
@@ -234,6 +234,7 @@ class AuthScreen extends PureComponent<Props> {
     const { dispatch, realm } = this.props;
     const auth = webAuth.authFromCallbackUrl(event.url, otp, realm);
     if (auth) {
+      dispatch(resetToLoading());
       dispatch(loginSuccess(auth.realm, auth.email, auth.apiKey));
```
gnprice, line 238: There's a couple of other places we dispatch `loginSuccess`. We probably want this there too, right?

src/account-info/ProfileCard.js
```diff
@@ -57,6 +58,7 @@ class LogoutButton extends PureComponent<{| +dispatch: Dispatch |}> {
   onPress = () => {
     const { dispatch } = this.props;
     dispatch(tryStopNotifications());
+    dispatch(resetToAccountPicker());
     dispatch(logout());
```
gnprice, line 62: It's kind of awkward that we end up having to repeat these separately at each call site. It seems like when logging out, we're consistently always going to want to go to the account picker. (Or maybe an improvement: go conditionally to either the account picker if you have another account to use, or straight to the login flow (RealmScreen) if you don't! That would mean additional logic, which we certainly wouldn't want to repeat at each call site.)

The above comment about `loginSuccess` highlights this -- it looks like a couple of call sites there might have been missed.

One thing that would probably help is to wrap these two lines up in a thunk action creator. Then `accountActions` might not even export the plain-action creators like `logout`; all call sites would call the wrapper, and that thunk would dispatch the plain action as well as cause a navigation.

src/nav/navReducer.js
```diff
@@ -29,7 +30,7 @@ export const initialState = getStateForRoute('loading');
 export default (state: NavigationState = initialState, action: Action): NavigationState => {
   switch (action.type) {
     case INITIAL_FETCH_COMPLETE:
-      return state.routes[0].routeName === 'main' ? state : getStateForRoute('main');
+      return getCurrentRouteName() === 'main' ? state : getStateForRoute('main');
```
gnprice: This is actually changing the behavior here, I think -- the `[0]` was intentional. Probably deserved a comment, or something in the code making it explicit.

The idea of this code:
```js
      return state.routes[0].routeName === 'main' ? state : getStateForRoute('main');
```
was: if we're *anywhere in the normal UI of the app* when an initial fetch completes, then remain there. Only if we're elsewhere -- which I think is just a way of saying, only if we're on the loading screen -- do we navigate, and in that case go to the main screen.

So in particular if you reopen the app after it was in the background for a while, and you were on some conversation, then it shows a loading animation there for a while, and when the fetch completes it shows you the conversation with its data. Similarly if you launch the app by opening a notification.

src/nav/AppWithNavigation.js
```diff
@@ -2,10 +2,17 @@
...
+// $FlowFixMe - should use a type-checked `connect`
+export default connect(
+  state => ({
+    state: getNav(state),
+  }),
+  null,
+  null,
+  { forwardRef: true },
+  // $FlowFixMe - not sure what's going on here
+)(createReduxContainer(AppNavigator, 'root'));
```
gnprice, line 17: I think this takes care of it:

```suggestion
)((createReduxContainer(AppNavigator, 'root'): ComponentType<{||}>));
```

Here's what the error message was:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/nav/AppWithNavigation.js:16:3

Missing type annotation for ContainerProps. ContainerProps is a type parameter
declared in function [1] and was implicitly instantiated at call of
createReduxContainer [2].

     src/nav/AppWithNavigation.js
     13│   null,
     14│   null,
     15│   { forwardRef: true },
 [2] 16│ )(createReduxContainer(AppNavigator, 'root'));
     17│

     node_modules/react-navigation-redux-helpers/src/create-redux-container.js
 [1] 28│ function createReduxContainer<
     29│   State: NavigationState,
     30│   Options: {},
     31│   NavigatorProps: NavigationNavigatorProps<Options, State>,
     32│   NavigatorType: NavigationNavigator<
     33│     State,
     34│     Options,
     35│     NavigatorProps,
     36│   >,
     37│   ContainerProps: {
     38│     ...$Diff<NavigatorProps, InjectedProps<State>>,
     39│     ...$Exact<RequiredProps<State>>,
     40│     theme: SupportedThemes | 'no-preference',
     41│   },
     42│ >(
     43│   Navigator: NavigatorType,
     44│   key?: string = "root",
     45│ ): React.ComponentType<ContainerProps> {
```
So it can't tell what the type parameter `ContainerProps` should be for this `createReduxContainer` call. And that's apparently the props type of the return value. Adding an annotation with that return type fixes the issue.

I suspect the reason this appeared with this change is that before this change, with a well-typed `connect`, the type on `connect` propagated information through so that Flow could infer the type of its argument here, the return value of `createReduxContainer`. And then when `connect` is marked with a fixme, Flow just doesn't know what to think here.

gnprice (commented):
Thanks @chrisbobbe! I like what this branch is doing. Some comments below.

I haven't read the REHYDRATE commit in detail, because I believe you're going to rework that one a fair amount.

src/nav/NavigationService.js
```diff
@@ -0,0 +1,55 @@
...
+const reduxContainerRef = React.createRef<ReduxContainer>();
+
+const getState = (): NavigationState => {
+  if (reduxContainerRef.current === null) {
+    throw new Error('Tried to use NavigationService before reduxContainerRef was set.');
+  }
+  return (
+    reduxContainerRef.current
+      // $FlowFixMe - how to tell Flow about this method?
+      .getCurrentNavigation().state
```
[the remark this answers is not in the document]
chrisbobbe: > Looks like the existing libdef for `react-navigation-redux-helpers` is a stub, so that's no help. Given that we're aiming to rip that out soon, it may not be worth the effort to prepare a real one.

Makes sense. I expect we'll have a very similar problem with the ref we'll set to React Navigation's own component (when r-n-r-h's `createReduxContainer` out of the picture), but we can cross that bridge when we come to it.

src/nav/NavigationService.js
```diff
@@ -0,0 +1,55 @@
...
+import type {
+  NavigationAction,
+  NavigationNavigatorProps,
+  NavigationState,
+  NavigationDispatch,
+  SupportedThemes,
+} from 'react-navigation';
+
+type ReduxContainerProps = {
+  ...$Diff<NavigationNavigatorProps<{}, NavigationState>, { navigation: mixed }>,
```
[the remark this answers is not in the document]
chrisbobbe: Makes sense. It's `{}` in react-navigation-redux-helpers, which doesn't help us decide which is meant. With `{}` being implicitly inexact, I guess the safe option is to make it be explicitly inexact?

src/nav/NavigationService.js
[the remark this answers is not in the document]
chrisbobbe: > Looks like the existing libdef for `react-navigation-redux-helpers` is a stub, so that's no help.

True, but there are some Flow types in node_modules/react-navigation-redux-helpers/src/create-redux-container.js. But I didn't glean anything that I found particularly helpful for this.

src/nav/NavigationService.js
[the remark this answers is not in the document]
gnprice: Here's the libdef's definition of the type it's being passed as a type argument to:
```js
  declare export type NavigationNavigatorProps<O: {...}, S: {...}> = $Shape<{
// …
```

So the default according to the libdef is an (explicitly) inexact type. So yeah, let's make it explicitly inexact to match that default.

src/nav/navReducer.js
chrisbobbe: I could go further and just make it check whether it's on the loading screen?

chrisbobbe:
Thanks for the review! Just pushed my latest revision.

src/start/DevAuthScreen.js
```diff
@@ -74,13 +74,13 @@ class DevAuthScreen extends PureComponent<Props, State> {
...
   tryDevLogin = async (email: string) => {
-    const { partialAuth } = this.props;
+    const { partialAuth, dispatch } = this.props;
 
     this.setState({ progress: true, error: undefined });
 
     try {
       const { api_key } = await api.devFetchApiKey(partialAuth, email);
-      this.props.dispatch(loginSuccess(partialAuth.realm, email, api_key));
+      dispatch(loginSuccess(partialAuth.realm, email, api_key));
```
gnprice, line 83: nit: this hunk is now a no-op and best left out

src/nav/navReducer.js
gnprice: That'd probably be good; we'd just want to check to be confident that there isn't another case where we'd want this behavior (and where the current code is getting it.) Probably best as a separate commit, either within this branch or afterward.

gnprice (commented):
Thanks @chrisbobbe! These revisions look good, with some small comments below (and above.)

The big gap is that I haven't yet read the new commit for the REHYDRATE case. I'll return to that; but it's dinnertime, and I wanted to send the rest of this off so you'll have it.

src/nav/navReducer.js
```diff
@@ -11,17 +11,7 @@ import { INITIAL_FETCH_COMPLETE, ACCOUNT_SWITCH, LOGIN_SUCCESS, LOGOUT } from '.
  * Private; exported only for tests.
  */
 export const getStateForRoute = (route: string): NavigationState => {
-  // TODO: this is kind of a hack!  Refactor to a better way.
```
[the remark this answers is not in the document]
chrisbobbe: Aha, interesting—I think this is probably a case of there being a previous revision where something different came after "a", such that "a" was definitely correct, but then I put that `initial: true` bit after the "a" and forgot to change "a" to "an". Possibly because the backtick made the incorrectness not stand out to me as much? 🤔

src/message/fetchActions.js
```diff
@@ -156,10 +158,20 @@ const initialFetchStart = (): Action => ({
   type: INITIAL_FETCH_START,
 });
 
-const initialFetchComplete = (): Action => ({
+const initialFetchCompletePlain = (): Action => ({
   type: INITIAL_FETCH_COMPLETE,
 });
 
+export const initialFetchComplete = () => async (dispatch: Dispatch, getState: GetState) => {
+  if (getNavigationRoutes()[0] !== 'main') {
```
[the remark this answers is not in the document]
chrisbobbe: Thanks for catching this!

chrisbobbe:
Thanks for the review! I've just addressed those comments.

src/nav/navReducer.js
```diff
@@ -41,43 +34,10 @@ export const getStateForRoute = (route: string): NavigationState => {
...
-  }
-
-  // If there's an active, logged-in account but no server data, then behave
-  // like `ACCOUNT_SWITCH`: show loading screen.  Crucially, `sessionReducer`
-  // will have set `needsInitialFetch`, too, so we really will be loading.
-  //
-  // (Valid server data must have a user: the self user, at a minimum.
-  // Compare getHaveServerData; this has an extra check because `users`
-  // may be missing entirely.)
-  if (rehydratedState.users === undefined || rehydratedState.users.length === 0) {
```
gnprice, line 65: The last part of this comment was meant to explain why this conditional can't simply use `getHaveServerData`. Since with this change we get to delete that bit of comment and the reason for it, we can also switch to simply using that selector :slightly_smiling_face: , like we do in `MainScreenWithTabs`. That'll make this code more explicit about what it's about.

src/nav/InitialNavigationDispatcher.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+
+    // If there's an active, logged-in account but no server data, then behave
+    // like `ACCOUNT_SWITCH`: show loading screen.  Crucially, `sessionReducer`
+    // will have set `needsInitialFetch`, too, so we really will be loading.
+    //
+    // (Valid server data must have a user: the self user, at a minimum.)
+    if (users.length === 0) {
+      // We're already on the loading screen.
```
gnprice, line 64: This calls for a few words about why we believe that's the case. (I think these lines originated with https://github.com/zulip/zulip-mobile/commit/57f1a7630f9d0b69f15349fa6ab61afd81242dac in your first revision of this branch, where this code is in a method on `LoadingScreen` and so the context already makes that clear.)

src/nav/InitialNavigationDispatcher.js
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * Not to be called before the REHYDRATE action, and not to be
+   * called more than once.
+   */
+  doInitialNavigation = () => {
+    const { hasAuth, accounts, users, dispatch } = this.props;
+
+    // If there are accounts but the active account is not logged in,
+    // show account screen.
+    if (!hasAuth) {
```
gnprice, line 48: The "If there are accounts" part no longer applies -- in the old code, it's there to make a contrast with the preceding condition:
```js
  // If there's no data to rehydrate, or no account data, show login screen.	
  if (!action.payload || !action.payload.accounts) {	
    return getStateForRoute('realm');	
  }	
```
but that's dissolved by the fact that with this design we're looking at a nice well-formed state post-REHYDRATE, rather than the REHYDRATE payload which is a bunch of pieces in roughly the shape of our state.

gnprice (commented):
Thanks again! All those commits LGTM now, and I've just read the remaining one. A few small comments below, mostly about comments; then please merge at will.

src/nav/InitialNavigationDispatcher.js
chrisbobbe: You're saying the comment should be adjusted, right:

```diff
-   // If there are accounts but the active account is not logged in,
-   // show account screen.
+   // If the active account is not logged in, show account screen.
```

chrisbobbe: I've stared at this a bit, and I think something is still off:

```js
    // If the active account is not logged in, show account screen.
    if (!hasAuth) {
      if (accounts.length > 1) {
        dispatch(resetToAccountPicker());
        return;
      } else {
        dispatch(resetToRealmScreen({ initial: true }));
        return;
      }
    }
```

I think maybe

```diff
-     if (accounts.length > 1) {
+     if (accounts.length >= 1) {
```

?

chrisbobbe: Still doesn't quite line up with the comment; maybe something like this, to replace the code in the multi-line quote [above](https://github.com/zulip/zulip-mobile/pull/4274#discussion_r508171639):

```js
    // If we don't know about any accounts, go to the realm-input screen.
    if (accounts.length === 0) {
      dispatch(resetToRealmScreen({ initial: true }));
      return;
    }

    // If the active account is not logged in, show account screen.
    if (!hasAuth) {
      dispatch(resetToAccountPicker());
      return;
    }
```

chrisbobbe:
Thanks again! I just responded to your comments, with one small question, and I also added one additional commit at the tip.

src/nav/InitialNavigationDispatcher.js
gnprice: (Discussed [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/decouple.20nav.20from.20redux.20%28.23M3804%29/near/1042584), and the result is in your new tip commit.)

gnprice:
LGTM! Thanks for all the revisions. Merging.

chrisbobbe:

I seemed to have dropped a ball here, and I didn't mark this PR as blocked by #4222 before this PR was merged.

In discussion, I'd [said](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/decouple.20nav.20from.20redux.20(.23M3804)/near/1032019):

> OK, I've just realized that #M4222 will make this a lot easier. With navigation state stored in Redux, it doesn't matter that the `key` hack in `TranslationProvider` unmounts and remounts the root navigator. But when we solely rely on that navigator for the navigation state, that `key` hack becomes a problem—changing the language puts you on LoadingScreen, 'loading' being the `initialRouteName` of `AppNavigator`, and you stay there indefinitely.

But having said that, I didn't then go and mark this PR as blocked, either on #4222 or a more complicated solution. Oops.

The effect is that, on `master`, changing the language is broken. The symptom I quoted above...isn't the one I'm seeing now, but it's still unstable and something we should fix.

Now, if I select a different language, it resets to the main-tabs screen. A lot (!) better than getting stuck indefinitely on the loading screen, but I'm not currently sure why it's not doing that anymore. I've pushed a revision for #4222 just now after resolving some conflicts.

Issue #4275, opened 2020-10-07
Crash on logout


gnprice:
This is a regression in the unreleased v27.155. @chrisbobbe [writes in a chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Crash.20on.20logout/near/1027082):
> Hmm, it appears that we've got a crash on pressing the logout button that was introduced in the React Navigation v4 upgrade commit (f3b6c1f9e) and is still active on `master`.
>
> -----
>
> At that commit, the error is the following:
>
> ```
> "ExceptionsManager.js:76 TypeError: Cannot read property 'state' of undefined"
> ```
>
> I've painstakingly traced the problem to a [particular line](https://github.com/react-navigation/react-navigation/blob/4.x/packages/core/src/getChildNavigation.js#L92) in react-navigation. (React, or React Native, tries to be helpful by *only* showing the component hierarchy—but in this case, an actual stack trace would have been helpful!) At that line, a function `getChildNavigation` is called (it's actually a recursive call), but with the first argument (the value of `children[childKey]`) as `undefined`. It was not anticipated that `undefined` might be passed, and the above-quoted type error throws on [this line](https://github.com/react-navigation/react-navigation/blob/4.x/packages/core/src/getChildNavigation.js#L18). […]
>
> -----
>
> The symptoms of the crash change at the commit where we upgrade to react-redux v6 (fd4debbae). There, I  newly don't get the above-quoted error message; instead, I get the familiar "no server data found", which usually indicates that we're trying to show a user-data-dependent screen without user data. This is the error we see on `master` too.
>
> If I just revert the `react-navigation` v4 upgrade commit (f3b6c1f9e) atop `master` (keeping our react-redux upgrades), there is no crash at all—which is why I suspect the react-redux v6 upgrade means a change of symptoms of the crash, and not a whole new crash altogether.

See that chat conversation for debugging details, particularly starting [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Crash.20on.20logout/near/1027476) after we dealt with a tools issue. I believe what's happening is that as we navigate away from MainScreenWithTabs as a result of the logout, that component and some of its descendants get `render`ed again. The "no server data found" error occurs exactly when that happens to a component that needs server data, like UnreadCards, and there isn't any. And the `getChildNavigation` crash appears to come from that happening to a child navigator after the parent has forgotten about it.

Because this is a crash, it's a blocker for making a release.

PR #4276, opened 2020-10-07
Fix logout crash, by rendering nothing in main screen if no data.

gnprice:
Fixes: #4275


chrisbobbe:
Looks great, thanks! Tested and merged.

Issue #3503, opened 2019-05-29
Navigation - 'Back' button should always bring you to 'Home'


borisyankov:
The intended behavior of the 'Back' button in the navbar intends to return the user always to the 'Home' screen (`MainScreenWithTabs`), even if the user have navigated from one narrow to another. We even use some custom logic for that.

This mostly works, except in few cases that may confuse our current logic. One such case:
* go to a stream narrow
* open the list of topics
* navigate to a topic
* go Back

Expected behavior - you are in 'Home'. Current behavior - you are in the `TopicsScreen`.

ishammahajan:
Why is this the intended behaviour? The user when pressing the back button expects to go back one navigation level, i.e. where they arrived on the current screen from.

Infact I thought there might have been a bug where clicking back from the `topic` screen doesn't take me back to the `stream` where it came from, which is why I filed #3455. There has been a [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/switching.20back.20from.20topic.20to.20stream.20narrow) (well, sort of) on this too.

borisyankov:
That is a good point. We used to have this behavior in the past. I actually thought this to be the default sensical behavior.

Sometimes though you might go several levels deep but apparently people still want to go back to the Home/Root screen so they navigate elsewhere. Enough users complained about it.

At least currently this is the intended behavior (likely to not change soon without a redesign of the UI, which itself is not a priority) I did file the issue because I already had the commit that fixes it.

gnprice:
I basically agree with @ishammahajan -- to the extent that the code currently does successfully have this behavior, I usually find it annoying and not what I wanted.

Good to have this issue to discuss it on! And thanks @ishammahajan for the links to previous discussions. I just went looking to see if we have an older issue thread on this subject; I didn't find one.

ishammahajan:
> Enough users complained about it.

I haven't really thought this through, but if there are enough users on both sides of the argument we could have a setting which changes the behaviour dynamically. I've seen some apps that do this (albeit IIRC, they were Reddit apps). Just a thought.

borisyankov:
> could have a setting which changes the behaviour dynamically

Death by a thousand app settings?

borisyankov:
The decision was made a long time ago but people used to complain and now they don't. @timabbott might have thoughts on this (or not, has been a while).
Switching to the granular back behavior is straight-forward so this is not a technical problem.

ray-kraesig:
[...]

PR #4278, opened 2020-10-13
Decouple Navigation from Redux (Part 2)

chrisbobbe:
Fixes: #3804

~To follow #4274, which is still in review.~


chrisbobbe:
Now unblocking, as both #4274 and #4222 have landed! 🙂 So this is ready for a review, at your convenience, @gnprice.

src/ZulipMobile.js
```diff
@@ -31,7 +31,12 @@ export default (): React$Node => (
             <ThemeProvider>
               <InitialNavigationDispatcher>
                 <BackNavigationHandler>
-                  <AppWithNavigation ref={NavigationService.reduxContainerRef} />
+                  <AppContainer
+                    // `static navigationOptions` and `static router` not
+                    // being handled properly
+                    // $FlowFixMe
+                    ref={NavigationService.appContainerRef}
```
gnprice: Hmm interesting. Here's the error message I get when I remove the fixme:
```
Cannot create AppContainer element because:
 • Either mixed [1] is incompatible with React.ComponentType [2] in property
   current of property ref.
 • Or a call signature declaring the expected parameter / return type is missing
   in object type [3] but exists in function type [4] in property ref.
 • Or withRouter [5] is not a React component.
 • Or withOptionalNavigationOptions [6] is not a React component.

     src/ZulipMobile.js
      29│           <TranslationProvider>
      30│             <ThemeProvider>
      31│               <InitialNavigationDispatcher>
      32│                 <AppContainer
      33│                   // `static navigationOptions` and `static router` not
      34│                   // being handled properly
      35│                   ref={NavigationService.appContainerRef}
      36│                 />
      37│               </InitialNavigationDispatcher>
      38│             </ThemeProvider>
      39│           </TranslationProvider>

     flow-typed/npm/react-navigation_v4.x.x.js
 [2] 413│   > = React$ComponentType<Props> &
 [5] 414│     withRouter<State, Options> &
 [6] 415│     withOptionalNavigationOptions<Options>;

     /tmp/flow/flowlib_1bc53cbd/react.js
 [1] 159│ declare type React$ComponentType<-Config> = React$AbstractComponent<Config, mixed>;
        :
 [4] 197│   | ((React$ElementRef<ElementType> | null) => mixed)
        :
 [3] 246│   ): {|current: null | T|};
```

I tried removing those statics from the `NavigationContainer` definition in the libdef:
```diff
--- a/flow-typed/npm/react-navigation_v4.x.x.js
+++ b/flow-typed/npm/react-navigation_v4.x.x.js
@@ -410,9 +410,9 @@ declare module 'react-navigation' {
     State: NavigationState,
     Options: {...},
     Props: NavigationContainerProps<Options, State>,
-  > = React$ComponentType<Props> &
+  > = React$ComponentType<Props> /* &
     withRouter<State, Options> &
-    withOptionalNavigationOptions<Options>;
+    withOptionalNavigationOptions<Options> */;
```
That eliminated the last two of those four errors, but the first two remain. So I think the issue is more basic.

... OK, on some experimentation, I think I see the cause! It looks like the type parameter to `React.createRef` should be a `React$ElementRef` type:
```diff
 const appContainerRef = React.createRef<
-  NavigationContainer<NavigationState, { ... }, NavigationContainerProps<{ ... }, NavigationState>>,
+  React$ElementRef<
+    NavigationContainer<NavigationState, { ... }, NavigationContainerProps<{ ... }, NavigationState>>,
+  >,
 >();
```
That resolves this fixme, and also the one on `appContainerRef.current.dispatch`.

I got there by looking at the type definitions mentioned in that Flow error message, and in particular looking closely at the `createRef` declaration:
```js
  declare export function createRef<T>(
  ): {|current: null | T|};
```
and comparing with this one in the same file (and also mentioned in the error message):
```js
/**
 * The type of the ref prop available on all React components.
 */
declare type React$Ref<ElementType: React$ElementType> =
  | { -current: React$ElementRef<ElementType> | null, ... }
  | ((React$ElementRef<ElementType> | null) => mixed)
  | number | string;
```
Note the types of the `current` property.

I'll push a commit with that change to the tip of this PR branch.

In the course of finding that, I also discovered some small bugs in the libdef and other small ways it can be improved. I pushed those fixes to [another branch](https://github.com/zulip/zulip-mobile/compare/master...gnprice:dev-nav-libdef), though they turn out not to be necessary here.

src/nav/BackNavigationHandler.js
```diff
@@ -1,33 +0,0 @@
...
-  }
-
-  componentWillUnmount() {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBackButtonPress);
-  }
-
-  handleBackButtonPress = () => {
-    const canGoBack = NavigationService.getState().index > 0;
-    if (canGoBack) {
-      NavigationService.dispatch(navigateBack());
```
gnprice: So this line was actually doing something quite different from this default behavior of React Navigation:

    """
    By default, when user presses the Android hardware back button,
    react-navigation will pop a screen or exit the app if there are no
    screens to pop. This is a sensible default behavior, but there are
    situations when you might want to implement custom handling.
    """

Namely:
```js
export const navigateBack = (): NavigationAction => StackActions.pop({ n: getSameRoutesCount() });
```
It pops all the topmost screens that happen to be on the same route. In particular, if you've navigated from one narrow (route `chat`) to another, it will pop both/all of those narrows.

This is the behavior discussed at #3954, #3503, and other places linked from [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/priorities/near/995971).

I don't really *like* this behavior, and it might well be an improvement to just drop it in favor of a straightforward "pop one screen" behavior. But we should be sure to consider that explicitly as a UI decision; in this commit it looks like it's happening as an unintended consequence of an infrastructural change.

gnprice (commented):
Thanks @chrisbobbe ! Glad to see this. Comments below.

src/nav/BackNavigationHandler.js
chrisbobbe: Oh, you're totally right—you had already pointed this out, on a video call, and it seems to have fallen off my radar.

src/nav/NavigationService.js
```diff
@@ -7,8 +7,13 @@ import type {
   NavigationContainerProps,
 } from 'react-navigation';
 
+/* prettier-ignore */
 const appContainerRef = React.createRef<
-  NavigationContainer<NavigationState, { ... }, NavigationContainerProps<{ ... }, NavigationState>>,
+  React$ElementRef<
```
chrisbobbe, line 12: Amazing!! It sounds like we've finally found what `React$ElementRef` is for—thanks! See, e.g., 333c31255af7d90f438c938c74220f1ae4a04eba and its parent c1f9d1543c92b536861085c7d30ef65bdbf0e617; the latter links to https://github.com/zulip/zulip-mobile/pull/4101#discussion_r449259543 (in which I used `React$ElementRef` in very much the wrong way).

chrisbobbe: Hmmmm, I'm actually getting an error when I remove the line

```
// $FlowFixMe - how to tell Flow about `.state`?
```

(below; it seems GitHub won't let me add a comment to a line this far away from a diff chunk)

chrisbobbe: Looks like it's on the thing we're trying to pass as a type argument to `React.createRef`, and it looks familiar:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/nav/NavigationService.js:12:3

Cannot instantiate React.ElementRef because:
 • Either withRouter [1] is not a React component.
 • Or withOptionalNavigationOptions [2] is not a React component.

     src/nav/NavigationService.js
       9│
      10│ /* prettier-ignore */
      11│ const appContainerRef = React.createRef<
      12│   React$ElementRef<
      13│     NavigationContainer<
      14│       NavigationState,
      15│       { ... },
      16│       NavigationContainerProps<{ ... }, NavigationState>>>
      17│ >();
      18│
      19│ const getState = (): NavigationState => {

     flow-typed/npm/react-navigation_v4.x.x.js
 [1] 414│     withRouter<State, Options> &
 [2] 415│     withOptionalNavigationOptions<Options>;
```

chrisbobbe: @gnprice, could you please check and see if you get that error too? I was expecting to be able to remove this `FlowFixMe` just like this WIP commit removes this other fixme:

```
// $FlowFixMe - how to tell Flow about this method?
```

chrisbobbe:
I just pushed a new revision; please take a look, @gnprice. 🙂  It's not mergeable yet—please see my comments at https://github.com/zulip/zulip-mobile/pull/4278#discussion_r516401809.

src/nav/NavigationService.js
gnprice: (We discussed this [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/decouple.20nav.20from.20redux.20%28.23M3804%29/near/1056167).) It seems like the `React$ElementRef` part is correct, and there's something else going on, probably in the react-navigation libdef. For now the right strategy is probably to type the ref the way we now think is right, and just include whatever fixmes we need to, like this one.

gnprice:
LGTM! One commit-message nit:
d10e393da BackNavigationHandler: Get `canGoBack` from `NavigationService`.
9c4a3cd56 navActions: Stop dispatching `navigateBack()` via Redux.
dcfc3c71c navActions: Stop dispatching `navigateToAccountDetails()` via Redux.
e8e808d87 navActions: Stop dispatching all actions via Redux.
4262acb6d reduxTypes: Cut out `NavigationAction` from `Dispatch` and `PlainDispatch`.
f4f8e9c0d nav: Stop using react-navigation-redux-helpers.
b9699b48f deps: Remove react-navigation-redux-helpers.
36dac0efb BackNavigationHandler [nfc]: Comment on our use of `navigateBack`.

I'd s/navActions:/nav:/ in those three commits toward the beginning of the branch. They're about how we navigate; and they're not particularly specific to the `navActions` module (they don't actually touch it), and saying just `nav:` helps group them with the other `nav:` commit later on.

Please merge at will after adjusting that fixme discussed above.

chrisbobbe:
> Please merge at will after adjusting that fixme discussed above.

Done, thanks for the review!

PR #4279, opened 2020-10-20
Add launch screens.

### LINKED ISSUE #2756: Add splash screen for before React has started up

chrisbobbe:
Fixes: #2756

We've [discussed](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5/near/1039507) a change in how we set the initial route when that becomes possible, i.e., when rehydration is complete.

In the proposed change (which should be possible even on our current version of React Navigation), we would just not render our navigation system's "Redux container" (a variant on React Navigation's own "App container", which we'll be using when #3804 is complete) while we're waiting for the Redux store to be rehydrated. (Then, when we do show it, we'll know exactly what to put for `initialRouteName`.)

That raises the question of what to show during that time. Currently, we show the `LoadingScreen`, but that's a component that's built to be embedded in our navigation system, and it probably wouldn't be great to duplicate the parts that aren't, and suit them to this purpose.

One attractive option is to extend the duration of the splash screen to cover the rehydration time. `expo-splash-screen` will let us do that (see its [example](https://docs.expo.io/versions/latest/sdk/splash-screen/#example-keep-splash-screen-open-while-loading) for this pattern), and it also gives us lots of detailed instructions ([ios](https://github.com/expo/expo/tree/master/packages/expo-splash-screen#-configure-ios), [android](https://github.com/expo/expo/tree/master/packages/expo-splash-screen#-configure-android)) for setting up a splash/launch screen in the first place, even without using `expo-splash-screen`. That lets us fix #2756, though without anything fancy—just a solid brand-color background. But [Apple's HIG document on the launch screen](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/launch-screen/) is pretty conservative about what should go in a launch screen, and a solid color seems quite likely to pass review.

For the record: I've found that Apple consistently prefers the term "launch screen"; I'm not sure about Android, but I think "splash screen" (what the Expo package name uses) is fine.


ios/ZulipMobile/Assets.xcassets/LaunchScreen.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    {
+      "filename" : "blank-1.png",
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "filename" : "blank-2.png",
+      "idiom" : "universal",
+      "scale" : "3x"
```
gnprice, line 16: Why three different files? It appears that they're all identical (same sequence of bytes.)

Also I read [those Expo instructions](https://github.com/expo/expo/tree/master/packages/expo-splash-screen#-add-an-image-to-imagesxcassets) as not really saying "you need to use an image"; rather they're assuming that the thing you want is an image, and they're providing instructions for how to do so. In particular the reference to "*the* image file that [is desired to] serve as a splash screen" sounds to me like it's assuming there is such an image, meaning an image you want to serve as a splash screen. So I expect you can simplify this with something like the suggestion you make in the commit message: an empty view with a background color, instead of an image view.

android/app/src/main/res/values/color.xml
```diff
@@ -2,4 +2,5 @@
 <resources>
     <color name="primaryColor">#ffffff</color>
     <color name="primaryColorDark">#d7ccc8</color>
+    <color name="splashscreen_background">#51c2af</color>
```
gnprice, line 5: nit: follow surrounding capitalization convention, which is camelCase

android/app/src/main/res/values/styles.xml
```diff
@@ -6,6 +6,9 @@
         <item name="colorPrimary">@color/primaryColor</item>
         <item name="colorPrimaryDark">@color/primaryColorDark</item>
         <item name="android:textColor">#000000</item>
+        <!-- From `expo-splash-screen` setup instructions. -->
+        <item name="android:drawable">@color/splashscreen_background</item>
```
gnprice, line 10: I think this line isn't meant to be here: seems redundant with the next, and I don't see it in the linked instructions.

android/app/src/main/res/values/styles.xml
```diff
@@ -6,6 +6,9 @@
...
+        <item name="android:windowBackground">@drawable/splashscreen</item>
```
gnprice, line 11: Hmm -- what if instead of `@drawable/splashscreen`, you wrote here `@color/splashscreen_background`? (Or `@color/splashscreenBackground`, after renaming.) Seems like that's all that `@drawable/splashscreen` consists of... and it itself is invoking the color from an attribute called `android:drawable`, so that suggests that a color counts as a drawable.

The Expo instructions need the extra indirection because the drawable XML is where you manually insert the image, if you so choose (with the `SplashScreenImageResizeMode.NATIVE` option.)

android/app/src/main/res/values/color.xml
gnprice: (If using that Expo package, it apparently relies on the specific resource names. That's not a great API choice, and it's not real clearly stated in those setup instructions, but it's e.g. here in the implementation:
https://github.com/expo/expo/blob/227455e3d6f4bc9d05ca84109db5cbc03a56e342/packages/expo-splash-screen/android/src/main/java/expo/modules/splashscreen/NativeResourcesBasedSplashScreenViewProvider.kt#L32-L36
But we're not, so we don't.)

android/app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,7 @@
         android:theme="@style/AppTheme">
         <activity
             android:name=".MainActivity"
+            android:theme="@style/AppTheme"
```
gnprice, line 28: I believe this is redundant -- we already have it on the application element, above.

android/app/src/main/res/values/color.xml
gnprice: ... Oh I see, using the Expo package is the next commit. :slightly_smiling_face:

gnprice (commented):
Thanks @chrisbobbe ! This looks like a good set of changes to make.

I'm a bit puzzled by the last commit, the one that brings in the expo-splash-screen library. It seems like we're doing quite well already as of the commit before :slightly_smiling_face: , just using the platforms' respective features directly, with the library having served just as a guide for what features to use and what steps to take. The library adds some machinery and doesn't seem to make our own code any simpler.

Ah, rereading the PR description, I see the idea is to use expo-splash-screen's machinery to have the native launch screen persist until rehydration is done. I think it may be simpler to just control that in the standard React way, like we do most things in our app, as I suggested [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5/near/1039507) in that thread.

android/app/src/main/res/values/color.xml
chrisbobbe: Through trial and error, I noticed that it did depend on it literally being called "splashscreen_background" and thought it was odd, but I didn't locate it in the Expo package.

My preference was actually something like "brand"—does that sound right? It would be nice to have something reusable on the Android native side for the brand color. And we may be able to bring it over the bridge with RN v0.63.0's fancy new [`PlatformColor` API](https://reactnative.dev/docs/platformcolor).

android/app/src/main/res/values/styles.xml
chrisbobbe: Oh interesting; thanks for catching this. Looks a lot like what goes in `splashscreen.xml`, and I guess I might have thrown it in here too.

android/app/src/main/res/values/styles.xml
chrisbobbe: Ooh, and it seems like we could remove splashscreen.xml altogether—yes, just tested, and that works.

ios/ZulipMobile/Assets.xcassets/LaunchScreen.imageset/Contents.json
chrisbobbe: Makes sense—it works perfectly fine without an image.

chrisbobbe:
Thanks for the review! We continued the conversation on CZO (around [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5/near/1042787)), and decided to not use `expo-splash-screen`, so this revision cuts that out. I also responded to your comments above.

One thing to point out—the "add splash screen for Android" commit is really quite silly in how small it is, and it basically just looks like "set a root-level background color". I think Android doesn't have a native concept of a launch/splash screen, like iOS does? 🤔 

Anyway, with that background color being set, I notice something that reminds me of the (iOS-side) annoyance I mention in 

95c57f286 ios: Set `rootView.loadingView` instead of `rootView.backgroundColor`.

I don't see anything odd when I rotate the device 90 degrees, but there's a kind of annoying appearance of that green background color when you open and close the keyboard. Try this reproduction, on Android:

- In any narrow, tap the compose box, to type a message.
- While the keyboard-appearance animation is in progress, watch as some solid brand-color space appears.
- Tap away from the compose box, to make the keyboard disappear
- Watch some solid brand-color space appear again (though I notice less of it here).

docs/architecture/platform-versions.md
```diff
@@ -88,6 +88,9 @@ History:
   discovered that we'd been mistaken in thinking since 2018 that its
   WebView browser got updated independently; in fact it's pinned at a
   version a couple of years older than any other browser we support.
+* We dropped iOS 10 support in 2020-10. It was <0.3% of iOS users
```
gnprice, line 91: nit: just "was 0.3%" -- in the message cited, that figure is rounded and the exact figure might have been anything from 0.25% to less than 0.35%

android/app/src/main/res/values/color.xml
gnprice: > My preference was actually something like "brand"—does that sound right? It would be nice to have something reusable on the Android native side for the brand color. And we may be able to bring it over the bridge with RN v0.63.0's fancy new [`PlatformColor` API](https://reactnative.dev/docs/platformcolor).

Sure, SGTM!

gnprice:
> I don't see anything odd when I rotate the device 90 degrees, but there's a kind of annoying appearance of that green background color when you open and close the keyboard.

[In the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5/near/1044222) we concluded that
* the same problem exists in the Expo solution;
* if we try the Android upstream recommendation for implementing a launch screen, it's foiled by RN -- there's a white flash in between the native launch screen and our `LoadingScreen` (i.e. between the native launch screen and the first React component tree);

and it seems like there isn't any readily-available solution for RN apps on Android that doesn't have one of those problems. So for Android, the best option is probably the status quo.

gnprice:
LGTM other than that final Android commit and the one nit above (https://github.com/zulip/zulip-mobile/pull/4279#discussion_r509698602) -- after those please merge at will.

chrisbobbe:
OK, sounds good. I just fixed that nit and removed that last commit for Android, and merged. Thanks again!

PR #4280, opened 2020-10-21
flipper: Upgrade Flipper to 0.39.0.

chrisbobbe:
See discussion [1], where Greg reported that 7c7271598 introduced an
issue where uploading images from the camera was broken on Android.

The issue is facebook/react-native#28551; it seems to be a problem
specifically with Flipper's "network" plugin [2].

The discussion on that issue links to facebook/flipper#993, which
suggests [3] upgrading Flipper to 0.39.0. So, do. On Android, it may
be necessary to do a `cd android && ./gradlew clean` after this
commit.

It's nice that we can upgrade Flipper independently of React Native;
it seems this will stay possible [4].

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flipper/near/1042657
[2] https://github.com/facebook/react-native/issues/28551#issuecomment-611085378
[3] https://github.com/facebook/flipper/issues/993#issuecomment-619823916
[4] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flipper/near/1042764

Fixes: #4281

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -28,4 +28,4 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
-FLIPPER_VERSION=0.33.1
+FLIPPER_VERSION=0.39.0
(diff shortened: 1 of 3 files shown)


gnprice:
LGTM! And you mentioned in the chat thread that you'd tested it some, I believe on both iOS and Android.

I just filed #4281 for the issue this fixes. I'll add that reference to the commit message, and merge!

PR #4285, opened 2020-10-22
i18n: Offer zh_TW instead of zh-Hant.

gnprice:
The existing zh-Hant translation appears to use Taiwanese vocabulary, so it seems like it was effectively a zh_TW translation in the first place under another name.  Between the two of them, the zh_TW translation is the one that's now nearly complete, so prefer that one.  See chat discussion: https://chat.zulip.org/#narrow/stream/58-translation/topic/zh-Hant.20.2F.20zh_TW/near/1044288

There are several ways to name the different zh* translations.  We follow Wikipedia's choices.  Facebook does the same (except with ASCII parens rather than full-width parens).  Apple and Google write zh_TW as "繁體中文（台灣）", i.e. leading with "繁體" which identifies the traditional script.

Also it seems like past time we sorted the language-picker UI by the language's endonym, rather than its English name -- so do that.


chrisbobbe:
Cool! I've posted a comment at https://chat.zulip.org/#narrow/stream/58-translation/topic/zh-Hant.20.2F.20zh_TW/near/1045443.

chrisbobbe:
Looks like this needs a rebase past a `react-intl` upgrade, also a change to use 繁體中文 （台灣） instead of 中文 （台灣）.

gnprice:
Rebased; but haven't yet tested that it works right with the react-intl changes, or made the changes to the language names.

gnprice:
OK, took care of both of these:

> but haven't yet tested that it works right with the react-intl changes, or made the changes to the language names.

as well as added a couple of other cleanups in this subsystem.

The reason this is still a draft is that we should first supply the missing migration 31 for `locale` -> `language`:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/state.2Esettings.2Elanguage/near/1246409
I would just go and add that here, were it not that it's time I should be going AFK and to bed. :-)

gnprice:
OK, rebased atop the merged #4959 to get that migration, and I believe this is now ready.

chrisbobbe:
LGTM, thanks!

PR #4295, opened 2020-10-31
UserItem [nfc]: Use more whole user objects, fewer bags of properties

gnprice:
These changes were sitting in an old branch of mine from February where I experimented with prototyping a major overhaul of the `Narrow` data structures. These are quite independent of that and easily pulled out, and help us advance #3764 (converting from using emails to using user IDs); so might as well get them in.


chrisbobbe:
Thanks, this LGTM! Merged.

Issue #4298, opened 2020-11-03
Group PM in outbox can be missing in alternate version of same thread


gnprice:
Repro:
* Find a group PM conversation where the users land in a different order if sorted by email than if by user ID.
* Send a message there, just so the conversation is easy to reach in the next couple of steps.
* Visit that conversation from the recent-PMs screen. Go back out.
* Go to the "all messages" narrow via the "globe" icon on the main/unreads screen, and then visit the same conversation via a recipient bar there.
* Put your device into airplane mode, so that sending a message will fail and it'll hang out in the outbox.
* Send a message. The message should appear with an outbox spinner.
* Go back out, and again to the recent-PMs screen and again visit the conversation from there.
* Expected: The message appears, with an outbox spinner.
* Actual: The message doesn't appear.

The root of the issue is that we can end up describing the same group-PM narrow by different orderings of the same users. Details are in this comment on `groupNarrow`:
```js
/**
 * A group PM narrow.
 *
 * The users represented in `emails` should agree, as a (multi)set, with
 * `pmKeyRecipientsFromMessage`.  But this isn't checked, and we've had bugs
 * where they don't; some consumers of this data re-normalize to be sure.
 *
 * They might not have a consistent sorting.  (This would be good to fix.)
 * Consumers of this data should sort for themselves when making comparisons.
 */
// Ideally, all callers should agree on how they're sorted, too.  Because
// they don't, we have latent bugs (possibly a live one somewhere) where we
// can wind up with several distinct narrows that are actually the same
// group PM conversation.
//
// For example this happens if you have a group PM conversation where email
// and ID sorting don't happen to coincide; visit a group PM conversation
// from the main nav (either the unreads or PMs screen) -- which sorts by
// email; and then visit the same conversation from a recipient bar on the
// "all messages" narrow -- which sorts by ID.  The Redux logs in the
// debugger will show two different entries in `state.narrows`.  This bug is
// merely latent only because it doesn't (as far as we know) have any
// user-visible effect.
```
I wrote that comment just recently, in 7727e2999. As described there, the bug is latent.

Then the complementary piece that causes this user-visible symptom is that when picking which outbox messages should be shown in a given narrow, we use a function called `narrowContains`... and when the narrow is any individual conversation, that function comes down to an attempted equality comparison:
```js
  return JSON.stringify(needle) === JSON.stringify(haystack);
```
Which will naturally give the wrong result when the users appear in different orders.

[...]

PR #4299, opened 2020-11-03
outbox: Fix some bugs from Narrow confusion; remove Narrow

gnprice:
This grew out of #3889. It follows #4294 by returning to the outbox code, and includes versions of all the changes made in #3889.

Along the way it fixes additional bugs, and the "Replace narrowContains" commit turns out to fix some bugs of its own. (That is, the old implementation turns out to be outright buggy. Meanwhile the new implementation here is from #4294 and well tested, and avoids the bugs in #3889's new implementation.)

Fixes: #4298
Fixes: #4301

### DIFF
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -81,7 +81,7 @@ describe('getMessagesForNarrow', () => {
         [JSON.stringify(privateNarrow('john@example.com'))]: [123],
       },
       messages,
-      outbox: [{ ...outboxMessage, narrow: streamNarrow('denmark') }],
+      outbox: [outboxMessage],
       realm: eg.realmState({ email: eg.selfUser.email }),
     });
 
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -34,6 +34,7 @@ describe('getMessagesForNarrow', () => {
       },
       messages,
       outbox: [],
+      realm: eg.realmState({ email: eg.selfUser.email }),
     });
 
     const result = getMessagesForNarrow(state, HOME_NARROW);
@@ -51,6 +52,7 @@ describe('getMessagesForNarrow', () => {
       caughtUp: {
         [HOME_NARROW_STR]: { older: false, newer: true },
       },
+      realm: eg.realmState({ email: eg.selfUser.email }),
     });
 
     const result = getMessagesForNarrow(state, HOME_NARROW);
@@ -65,6 +67,7 @@ describe('getMessagesForNarrow', () => {
       },
       messages,
       outbox: [outboxMessage],
+      realm: eg.realmState({ email: eg.selfUser.email }),
     });
 
     const result = getMessagesForNarrow(state, HOME_NARROW);
@@ -79,6 +82,7 @@ describe('getMessagesForNarrow', () => {
       },
       messages,
       outbox: [{ ...outboxMessage, narrow: streamNarrow('denmark') }],
+      realm: eg.realmState({ email: eg.selfUser.email }),
     });
 
     const result = getMessagesForNarrow(state, privateNarrow('john@example.com'));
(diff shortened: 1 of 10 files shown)


src/chat/__tests__/narrowsSelectors-test.js
```diff
@@ -34,6 +34,7 @@ describe('getMessagesForNarrow', () => {
       },
       messages,
       outbox: [],
+      realm: eg.realmState({ email: eg.selfUser.email }),
```
chrisbobbe, line 37: It looks like this is setting `realm.email` in the state because that value gets passed directly as `ownEmail` to `isMessageInNarrow`, having been selected by `getOwnEmail` (that `getOwnEmail` use having been added in this same commit).

I do notice that `getOwnEmail` has this in its JSDoc:

```
 * Prefer using `getOwnUserId` or `getOwnUser`; see #3764.
```

(#3764)

It looks like `getOwnUser` itself uses `getOwnEmail`, so it's not obviously better, I guess.

But I could imagine that being changed someday; I could see `getOwnUser` instead using, e.g., `getOwnUserId` (there's even a comment on `getOwnUserId` suggesting to do so) and `getUserForId`.

If such a change were made, it would no longer be useful to just set `state.realm.email`; we'd want to set `state.realm.user_id`. Possibly that's totally fine, and a bridge we'll happily cross when we get to it. But I wondered if we might like to consider providing a simple way to grab `state.realm` from the example data such that it contains all the stuff that's specific to the self-user: there's `email`, `user_id`, probably more.

chrisbobbe:
Looks good; it's great news to have this tidied up! Just one small query, below.

chrisbobbe (commented):
Looks good; it's great news to have this tidied up! Just one small query, below.

src/chat/__tests__/narrowsSelectors-test.js
gnprice: > But I wondered if we might like to consider providing a simple way to grab state.realm from the example data such that it contains all the stuff that's specific to the self-user: there's email, user_id, probably more.

Yeah, I agree -- that would be useful, and we'll want to add something like that. A more "batteries included" example Redux state, one might call it.

The reason I haven't yet gone and done that is that there are some API choices to be made about it:
* Probably it includes the self-user's email and related properties, as given by `eg.selfUser`, in `state.realm`.
* Probably it also includes `eg.selfUser` in `state.users`, and a corresponding account (with `eg.realm`) in `state.accounts`.
* Does it include `eg.otherUser` in `state.users`? Also `eg.thirdUser`? `eg.crossRealmBot`?
* There's `eg.stream` and `eg.subscription`. Does it include those in `state.streams` and `state.subscriptions`?
* Maybe there's even several different levels of "batteries included" that we want? In addition to probably still the initial state, the thing that's now `eg.reduxState`.
* What about when we add new example values along those lines, like the recent `thirdUser`?
  * If they're included, that potentially changes the behavior of existing tests. If it makes them fail, that's one thing; more insidiously, it could potentially make them pass trivially, even if the code they're meant to test stops behaving the way they're meant to check that it does.
  * If they're left out, then the "batteries included" state comes to include only some batteries, and potentially a less and less predictable subset of them.
  * Probably the right answer is to include new example values, and make a practice of systematically writing tests so that they don't depend on any assumptions that the state doesn't have miscellaneous other data they weren't expecting. That is, if a test makes its own `eg.makeStream()` call, it can assume *that* stream isn't in the example state; but it can't make an assumption like "there's only one stream in the state".
    * This requires a bit of work finding a clear pattern to prescribe for how to systematically do that.

This diff (as you saw, there are a good handful of lines like this one) definitely made me think about adding this. :slightly_smiling_face: Not enough that I felt I wanted to do it before shipping this branch, but it moved that feeling closer.

chrisbobbe: > Not enough that I felt I wanted to do it before shipping this branch, but it moved that feeling closer.

Makes sense, that's fine! So I'd say you can merge this branch at will.

gnprice:
Thanks @chrisbobbe for the review! Merged.

Issue #4303, opened 2020-11-11
Consecutive messages from same-named senders shown as if same sender


gnprice:
Repro:
* Get two user accounts with the same "full name". (E.g., change a test user's name: webapp > Settings > Your account > Full name.)
* On a given conversation -- either a stream and topic, or a PM conversation -- send a message from one of them, then from the other.
* Expected: we show a sender-header on the second message, to indicate it's sent by a different person from the previous message. Because the names are the same -- and the avatars could be too, if someone were actively trying to be deceptive -- it looks a little funny, but at least we avoid making it look just like a normal series of messages from a single sender.
* Actual: we show no sender-header, just as if the new message and the previous one were from the same person.

This is a bug I discovered by reading the code, specifically while studying `renderMessages` to understand #4301. It's pretty blatant in the code:
```js
    const shouldGroupWithPrev =
      !diffRecipient
      && !diffDays
      && prevItem
      && prevItem.sender_full_name === item.sender_full_name;
```
and it's been that way essentially forever, since f1b10f42a8a115f562d4091d1bd1d94546c7aa42 in 2016.

Instead of the sender's display name / "full name", we should use an actual ID for them: their numeric user ID, but even using their email (the legacy way of identifying a Zulip user) would fix this issue.

This bug isn't likely to be very commonly hit -- but it could be hit by someone deliberately impersonating another user, and would potentially amplify the confusion they cause. Marking therefore as priority.

PR #4304, opened 2020-11-11
msglist: Don't assume people with the same name are the same.

gnprice:
Fixes: #4303

Sadly we can't use the sender's actual, permanent, numeric user ID here, because outbox messages (values of type Outbox) don't have a sender_id. For now, using the email suffices to fix this bug, and leaves this code no worse than a lot of the rest of our code.

We also add a sender_id property to Outbox... but don't yet switch to using it, because existing outbox messages, if there were any just before the user upgrades to a release with this change, won't yet have it. Once such a release has been out for a few weeks, we can make the property required and have a migration just drop any values lacking it. There's a TODO comment marking this as one of the many sites of #3674.


chrisbobbe:
> There's a TODO comment marking this as one of the many sites of #3674.

Do you mean https://github.com/zulip/zulip-mobile/issues/3764?

src/message/renderMessages.js
```diff
@@ -31,11 +31,11 @@ export default (
         data: [],
       });
     }
+
+    // TODO(#3674): Use sender_id, not sender_email.  Needs adding
```
chrisbobbe: I think you mean #3764?

chrisbobbe:
LGTM except for one tiny comment, above—looks like a typo in an issue number.

gnprice:
> Do you mean #3764?

Ha, indeed I do. Thanks for the catch!

I'll edit that and merge.

PR #4308, opened 2020-11-12
webview build: Update target browser versions, and clarify.

gnprice:
These were kind of tucked away in the middle of a bunch of logic.
Perhaps more importantly, they had a comment pointing to js.js but
nothing vice versa -- so they weren't discoverable when updating
our canonical list of supported platforms.

And in fact, we didn't remember this line and update it when we most
recently updated that list, in c953bc336 a few weeks ago... nor the
time before that, in b3eced058 a year ago... and the latter commit
came just a few weeks after we wrote this code in the first place.

So, apply those two updates; and pull these minimums out with more
explanation, and linked both to and from the canonical list in js.js.

This update doesn't cause any change to the generated output; in fact
it turns out that there are no Babel plugins at all that are affected
by upgrading Chrome between versions 13 and 38.  That's perhaps
unsurprising when you remember that Chrome 37 is what shipped with
Android 5, in 2014, as the JS language was still just emerging from
its long period of stagnation.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -17,6 +17,8 @@ import rewriteHtml from './rewriteHtml';
 /*
  * Supported platforms:
  *
+ * * (When updating these, be sure to update tools/generate-webview-js too.)
+ *
  * * We support iOS 11.  So this code needs to work on Mobile Safari 11.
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Merged, thanks!

PR #4314, opened 2020-11-17
docs/profiling: Add a rough first doc on profiling and benchmarking.

gnprice:
Here's a rough writeup of what I've learned from experimenting with the tool called `react-native-performance-monitor`. It links to a branch where I've encoded all the trickier steps.

@chrisbobbe, would you try these instructions out and see if you can get working results? If something doesn't work for you, describe it in chat and we'll debug. In particular I haven't yet tested this on iOS.

Then I think probably the best workflow is for you to add a commit on top of this one that adds any further steps you needed, clarifications, or other changes; and we can merge that.

With some further work, I think everything in the setup could be merged, too. If we find ourselves doing this regularly, we'll want to do that, because it's just a trade of work up front for a bit of work each time. We can update these instructions accordingly when that happens.

### DIFF
--- a/docs/howto/ios-tips.md
+++ b/docs/howto/ios-tips.md
@@ -132,6 +132,9 @@ You should now be able to enter `http://zulipdev.com:9991` (not
 `https://`), see the "Sign in with Apple" button, and use it
 successfully.
 
+
+<div id="disable-ats" />
+
 ## Adding `http://` exceptions to `NSAppTransportSecurity` in `Info.plist`
 
 If you need to connect to `http://zulipdev.com` or another host with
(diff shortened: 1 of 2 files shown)


chrisbobbe:
> Then I think probably the best workflow is for you to add a commit on top of this one that adds any further steps you needed, clarifications, or other changes; and we can merge that.

OK, done! One change I made is to say that your separate branch enables plain HTTP for iOS, which it currently doesn't; I think you'll have to do that since it's not in a PR with "allow edits" checked.

gnprice (commented):
Oops, it seems I never submitted these review comments! They were "pending" just now; I think I wrote them soon after your change above.

I'm not sure where I was in reading this at the time (likely not 100% done, or I'd have submitted the review), but wanted to not delay these any longer. Was looking at this because I'm about to do some more profiling and benchmarking and use these techniques.

gnprice:
I'm going to tweak this a bit and then merge it -- it's documentation, so best to simply fix forward for anything that needs changing.

gnprice:
And done. Thanks for the revisions!

PR #4315, opened 2020-11-17
Stop storing `safeAreaInsets` in Redux; use react-native-safe-area-context instead.

chrisbobbe:
react-native-safe-area is getting a bit old; it's now been two years
since the last change in the GitHub repo
(miyabi/react-native-safe-area). And, according to NPM,
react-native-safe-area-context has almost 300,000 weekly
downloads [1], to react-native-safe-area's 7,000 [2].

This more popular library gives us a built-in way to consume the
safe-area insets, so that we don't have to go through Redux.
Actually, three ways: a HOC, a React Context consumer, and a React
Hook. The inclusion of the Hook, in particular, signals that this
library is still maintained and forward-looking. See discussion [3]
for one reason we might prefer to stop using Redux to handle the
safe-area insets.

We installed it in f3b6c1f9e to satisfy peer-dependency constraints,
but at that point we hadn't been planning on actually using it. So,
now, make a libdef for it (using Flowgen) and add the
`SafeAreaProvider` to our various wrappers in src/ZulipMobile.js.

[1] https://www.npmjs.com/package/react-native-safe-area-context
[2] https://www.npmjs.com/package/react-native-safe-area
[3] https://github.com/zulip/zulip-mobile/pull/4300#discussion_r524881747


src/common/Screen.js
```diff
@@ -151,6 +150,4 @@ class Screen extends PureComponent<Props> {
   }
 }
 
-export default connect(state => ({
-  safeAreaInsets: getSession(state).safeAreaInsets,
-}))(Screen);
+export default withSafeAreaInsets(Screen);
```
gnprice, line 153: Hmm -- this seems to lose us type-checking on these props. :-/

Try it by going to a call site of `Screen`, and editing e.g. `title=` to `titl=`. Flow gives no error.

I think one problem here may be that the type of `withSafeAreaInsets` is written with some types that were apparently invented for the TypeScript types for React, like `React.ForwardRefExoticComponent`. I guess those are effectively being treated as `any`, perhaps.

But really the root problem here seems to be within the `react-native-safe-area-context` library, and its TypeScript types. I don't see how this could possibly work to express the actual type of this function:
```js
// node_modules/react-native-safe-area-context/lib/typescript/src/SafeAreaContext.d.ts
export declare function withSafeAreaInsets<T>(WrappedComponent: React.ComponentType<T>): React.ForwardRefExoticComponent<React.PropsWithoutRef<T> & React.RefAttributes<T>>;
```
The inner component should take a prop `insets`, with a certain type... and the outer component should take no such prop. It looks like nothing in that type knows anything about a property name `insets`.

We might be best off with either our own small handwritten libdef, or a type-wrapper. The libdef (or type-wrapper) can have just the type `EdgeInsets` and the types of `withSafeAreaInsets` and `useSafeArea`.

The type of `withSafeAreaInsets` should look like a much simplified version of our `connect` type-wrapper: it should be just like what you get from `connect` if you pass it an argument of type `() => {| insets: EdgeInsets |}`.

gnprice (commented):
Thanks! One comment below on the code.

What devices (either physical or simulated) have you tested this on? It'd be good to test it on at least one device that has nontrivial insets, like an iPhone X-something, and one that doesn't.

src/common/Screen.js
chrisbobbe: Ah, thanks for catching that. How about bypassing the problem by using the [`useSafeAreaInsets` hook](https://github.com/th3rdwave/react-native-safe-area-context#usesafeareainsets)?

chrisbobbe: > `Screen` would be quite easy to convert to a Hooks-based component.

Ah, right, but `ComposeBox` wouldn't.

chrisbobbe: I'll see about making a small type-wrapper, so we can use the handy `BoundedDiff` type you wrote. 😄

chrisbobbe:
> What devices (either physical or simulated) have you tested this on? It'd be good to test it on at least one device that has nontrivial insets, like an iPhone X-something, and one that doesn't.

I've tested it on my iPhone 6s device, and on an iPhone X simulator, and it doesn't seem to make anything worse on either device, although there are still plenty of problems covered by #3066. I've done some logging just now, to report the following results.

One thing that the logging made plain was that the old code wasn't updating the safe areas when the device is rotated, while the new code does. Or it tries to, with glitches (see below). I'm not aware that we've ever tried to use the left-right inset measurements while the device is in landscape mode, but at least we'll have ready access to them, if we can work through these glitches. I.e., problems like the following are still present, but the path to fixing them might be shorter (though there's still a design discussion to be had):

image

If I log the insets in the `render` method of src/common/Screen.js, I see the following if I visit a screen that uses it (I used `UserStatusScreen`):

- iPhone 6 physical device:
  - Before this branch:
    - On navigating to the screen: `{ right: 0, top: 20, left: 0, bottom: 0 }`
    - While there, on changing the orientation: (no log)
  - After this branch:
    - On navigating to the screen: `{ right: 0, top: 20, left: 0, bottom: 0 }`
    - While there, on changing the orientation:
      - `{ left: 0, top: 0, right: 0, bottom: 0 }` whenever I switch to landscape
      - Ah! Whenever I switch back to portrait, I get three logs in quick succession (looks like a bug):
        1. `{ left: 0, top: 20, right: 0, bottom: 0 }`
        2. `{ left: 0, top: 0, right: 0, bottom: 0 }`
        3. `{ left: 0, top: 20, right: 0, bottom: 0 }`

- iPhone X simulator:
  - Before this branch:
    - On navigating to the screen: `{ bottom: 34, left: 0, right: 0, top: 44 }`
    - While there, on changing the orientation: (no log)
  - After this branch:
    - On navigating to the screen: `{ bottom: 34, left: 0, right: 0, top: 44 }`
    - While there, on changing the orientation—hmm, this seems a little random, but I think I've found a pattern. I'm consistently seeing the following on each cycle through four clockwise 90-degree turns, starting from upside-up portrait:
      - First rotation: Four logs in quick succession:
        1. `{ left: 44, top: 0, right: 0, bottom: 21 }`
        2. `{ left: 0, top: 0, right: 0, bottom: 0 }`
        3. `{ left: 44, top: 0, right: 0, bottom: 0 }`
        4. `{ left: 44, top: 0, right: 44, bottom: 21 }` (So, landing with right-looking values anyway...)
      - Second rotation: (no logs)
      - Third rotation: (no logs)
      - Fourth rotation: Four logs in quick succession:
        1. `{ left: 0, top: 44, right: 0, bottom: 0 }`
        2. `{ left: 0, top: 0, right: 0, bottom: 0 }`
        3. `{ left: 0, top: 44, right: 0, bottom: 0 }`
        4. `{ left: 0, top: 44, right: 0, bottom: 34 }` (So, landing with the same upside-up portrait values we saw when first navigating to the screen.)

chrisbobbe:
I've pushed a new revision with the small type-wrapper, but I'm concerned about the erratic way the insets update upon changing the orientation, hmm.

gnprice:
Thanks! The code looks great.

I agree that behavior is concerning. OTOH we already don't really make very correct use of this data... that's #3066 . So it may not be a regression.

If you're confident from testing that the app's behavior doesn't seem to get any worse, at least in simulators, then I think it'll make sense to merge. We'll have some debugging of this library to do, though. (Even if it turns out to be buggy only in a simulator, that's a meaningful bug because it makes it hard to detect or debug other bugs.) Probably that should be a comment on #3066 , as it's something we'll have to dig into when fixing that.

chrisbobbe:
> We'll have some debugging of this library to do, though. (Even if it turns out to be buggy only in a simulator, that's a meaningful bug because it makes it hard to detect or debug other bugs.) Probably that should be a comment on #3066 , as it's something we'll have to dig into when fixing that.

Yeah, makes sense; I'll go make that comment. Also, when we address #3066 more thoroughly, I'd like to look into not using `withSafeAreaInsets`/`useSafeAreaInsets` at all (or using them sparingly)—the [preferred thing](https://github.com/th3rdwave/react-native-safe-area-context#consumers) to use is `SafeAreaView`, specifically because it's supposed to behave better with screen rotations.

chrisbobbe:
> If you're confident from testing that the app's behavior doesn't seem to get any worse, at least in simulators, then I think it'll make sense to merge.

OK, done! Thanks for the review. 🙂

Issue #4239, opened 2020-08-26
Remove `react-native-text-input-reset` dependency, if possible


chrisbobbe:
I see this deprecation warning when running `./gradlew help --warning-mode=all` from `android` (recommended by [this doc](https://docs.gradle.org/5.6/userguide/upgrading_version_5.html#changes_5.6)):

```
> Configure project :react-native-text-input-reset
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed soon. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html
```

Its [GitHub page](https://github.com/nikolaiwarner/react-native-text-input-reset/tree/master) says it was only meant to be temporary, so I'm not confident that this deprecation warning will ever be addressed. But from a quick skim of the RN issue it was aimed to be a workaround for, https://github.com/facebook/react-native/pull/12462, it's not immediately clear that that issue has been fixed in any version of React Native.

A complete fix must include:
- this dependency is removed
- we're confident that https://github.com/facebook/react-native/pull/12462 doesn't affect us

aryanshridhar:
Working on it :+1:

aryanshridhar:
Taking this issue :)

aryanshridhar:
Hey , I see no difference while removing <code>TextInputReset</code> from <a href = 'https://github.com/zulip/zulip-mobile/blob/7e36838cc69a9d181839bc1ead72e45274da92e5/src/compose/ComposeBox.js#L102'>here</a> as the behaviour of ComposeBox still remains the same (contents of textinput being properly cleared)

Should I go on with it , and open up an PR removing this package ?

chrisbobbe:
Thanks for looking into this, and it sounds promising that it's working on your device! It would be great if you could read through https://github.com/facebook/react-native/pull/12462 and confirm that we can be reasonably sure that it won't affect other people's devices, too. I'm not very familiar with the details, but there seemed to be some nuances that suggest some people might see the bug, while others don't.

aryanshridhar:
As rightly said by you , looks like the same problem still persists in many android device and hasn't been merged due to lack of <a href = 'https://github.com/facebook/react-native/pull/12462#issuecomment-477295348'>verification</a> of the same .
However this issue can still be solved using native module method by integrating this <a href = 'https://github.com/facebook/react-native/pull/12462#issuecomment-298812731'>code</a> to zulip-mobile . 
[...]

PR #4317, opened 2020-11-25
narrow: Add `getNarrowsForMessage`.

chrisbobbe:
narrow: Add `getNarrowsForMessage`.

We've never had a function for getting all the narrows that a
message appears in. Such a function will be important
algorithmically for avoiding a performance regression with the
EVENT_NEW_MESSAGE action if/when we start storing `state.narrows` as
an `Immutable.Map`; see discussion of this strategy at
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60Immutable.2EMap.60.20performance.20for.20.60state.2Enarrows.60/near/1068912.

Might as well put it in `narrow.js` and make it reusable, too.

### DIFF
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -316,6 +316,47 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
   });
 
 /**
+ * Answers the question, "What narrows do this message appear in?"
+ *
+ * This function does not support search narrows, and it always
+ * excludes them.
+ *
+ * The message's flags must be in `flags`; `message.flags` is ignored.  This
+ * makes it the caller's responsibility to deal with the ambiguity in our
+ * Message type of whether the message's flags live in a `flags` property or
+ * somewhere else.
+ */
+export const getNarrowsForMessage = (
+  message: Message | Outbox,
+  ownUser: User,
+  flags: $ReadOnlyArray<string>,
+): Narrow[] => {
+  const result: Narrow[] = [];
+
+  // All messages are in the home narrow.
+  result.push(HOME_NARROW);
+
+  if (message.type === 'private') {
+    result.push(ALL_PRIVATE_NARROW);
+    result.push(pmNarrowFromEmails(pmKeyRecipientsFromMessage(message, ownUser).map(x => x.email)));
+  } else {
+    const streamName = streamNameOfStreamMessage(message);
+    result.push(topicNarrow(streamName, message.subject));
+    result.push(streamNarrow(streamName));
+  }
+
+  if (flags.includes('mentioned') || flags.includes('wildcard_mentioned')) {
+    result.push(MENTIONED_NARROW);
+  }
+
+  if (flags.includes('starred')) {
+    result.push(STARRED_NARROW);
+  }
+
+  return result;
+};
+
+/**
  * Answers the question, "Where should my reply to a message go?"
  *
  * For stream messages, chooses a topic narrow over a stream narrow.
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -318,22 +318,15 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
 /**
  * Answers the question, "Where should my reply to a message go?"
  *
- * Careful: quirky behavior on a stream message with empty topic.
- *
- * If you think you want to reuse this function: study carefully; maybe
- * refactor it to something cleaner first; if not, then definitely document
- * its quirky behavior.
+ * For stream messages, chooses a topic narrow over a stream narrow.
  */
-// TODO: do that, or just make this a private local helper of its one caller
+// TODO: probably make this a private local helper of its one caller,
+//   now that it's free of fiddly details from the Narrow data structure
 export const getNarrowForReply = (message: Message | Outbox, ownUser: User) => {
   if (message.type === 'private') {
     return pmNarrowFromEmails(pmKeyRecipientsFromMessage(message, ownUser).map(x => x.email));
   } else {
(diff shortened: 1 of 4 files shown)


src/utils/narrow.js
```diff
@@ -304,7 +304,14 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
 
-export const getNarrowFromMessage = (message: Message | Outbox, ownEmail: string) => {
+/**
+ * Answers the question, "Where should my reply to a message go?"
+ *
+ * Chooses a topic narrow over a stream narrow if the message has a
+ * topic. (FIXME: What does that mean; don't all messages have
+ * topics?)
+ */
+export const getNarrowForReply = (message: Message | Outbox, ownEmail: string) => {
```
gnprice: Great -- this is a much better name :slightly_smiling_face:

src/utils/narrow.js
```diff
@@ -358,19 +359,29 @@ export const getNarrowsForMessage = (
...
 
-  return topicNarrow(message.display_recipient, message.subject);
+  const [sendableStreamNarrow, sendableTopicNarrow] = [
+    sendableNarrows.find(n => isStreamNarrow(n)),
+    sendableNarrows.find(n => isTopicNarrow(n)),
+  ];
+  if (sendableStreamNarrow && sendableTopicNarrow) {
+    return sendableTopicNarrow;
+  } else if (sendableNarrows[0]) {
+    return sendableNarrows[0];
```
gnprice: This feels pretty fiddly to me. I think a cleaner way to avoid the duplication here is to factor just the duplicated code out: effectively this version is calling `getNarrowsForMessage` but then undoing some of what that function did, and we can get a cleaner result by instead having both functions call a common simpler helper that just does the part that's duplicated.

Following #4239, `pmKeyRecipientsFromMessage` is that helper and that's what  `getNarrowFromMessage` now uses.

src/utils/narrow.js
```diff
@@ -304,6 +304,55 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
+    if (message.display_recipient.length > 1) {
+      result.push(
+        groupNarrow(message.display_recipient.filter(x => x.email !== ownEmail).map(x => x.email)),
+      );
+    } else if (message.display_recipient.length === 1) {
+      result.push(privateNarrow(message.display_recipient[0].email));
+    } else {
+      // FIXME(crunchy): This shouldn't happen; reject at the edge.
+      result.push(groupNarrow([]));
+    }
```
gnprice: This part can be simplified following the model of #4329.

src/utils/narrow.js
```diff
@@ -304,6 +304,55 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
+  }
+
+  if (flags.includes('mentioned') || flags.includes('wildcard_mentioned')) {
+    result.push(MENTIONED_NARROW);
+  }
+
+  if (flags.includes('starred')) {
+    result.push(STARRED_NARROW);
+  }
+
```
gnprice: Let's add a comment mentioning search narrows, saying they're excluded.

In fact, that should probably be in the jsdoc -- just like `isMessageInNarrow`.

gnprice (commented):
Thanks @chrisbobbe ! Comments below. As you'll see, some of these refer to #4329, which I hadn't written yet when you originally wrote this PR :slightly_smiling_face:

src/utils/narrow.js
```diff
@@ -306,11 +306,16 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
-      message.display_recipient.length > 1
-        ? message.display_recipient.filter(x => x.email !== ownEmail)
-        : message.display_recipient;
-    return groupNarrow(recipient.map(x => x.email));
+    if (message.display_recipient.length > 1) {
+      return groupNarrow(
+        message.display_recipient.filter(x => x.email !== ownEmail).map(x => x.email),
+      );
+    } else if (message.display_recipient.length === 1) {
+      return privateNarrow(message.display_recipient[0].email);
```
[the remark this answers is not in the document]
gnprice: > I have a change that unifies the two of them in one of my draft branches for refactoring `Narrow`, but it's atop a bunch of other bigger changes. I noticed this same confusion in this bit of code the other day when preparing #4329, and I think I've seen it somewhere else recently too. So it might be past time for me to try making just that unification on its own -- it should simplify any further Narrow-refactoring anyway.

OK, and sent this as #4330!

Turned out the `privateNarrow` / `groupNarrow` unification was actually at the very start of that branch. :slightly_smiling_face: But I went and included some other pieces as well.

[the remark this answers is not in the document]
chrisbobbe: Great! I just reviewed and merged #4330; it LGTM.

src/utils/narrow.js
chrisbobbe: > Following #4239, `pmKeyRecipientsFromMessage` is that helper and that's what `getNarrowFromMessage` now uses.

(#4329, I think. 🙂)

chrisbobbe: > I think a cleaner way to avoid the duplication here is to factor just the duplicated code out: effectively this version is calling `getNarrowsForMessage` but then undoing some of what that function did, and we can get a cleaner result by instead having both functions call a common simpler helper that just does the part that's duplicated.

Ooh, nice! Thanks.

chrisbobbe:
Thanks for the review! 🙂 I've just pushed a new revision.

gnprice:
Thanks! Merged, with small tweaks.

chrisbobbe:
Thanks! I saw the tweaks; thanks for making them. 👍

PR #4318, opened 2020-11-26
Fix a few Flow complaints we'd get with RN 0.63 and Flow 0.122.0.

chrisbobbe:
The RN v0.63 upgrade issue is #4245. There'll likely be quite a bit more work to do for Flow when we get closer to doing that upgrade, but this is a start! 🎉

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -83,7 +83,6 @@ untyped-type-import=warn
 nonstrict-import=warn
 deprecated-type=warn
 unsafe-getters-setters=warn
-inexact-spread=warn
 unnecessary-invariant=warn
 signature-verification-failure=warn
 deprecated-utility=error
--- a/src/sharing/ShareToPm.js
+++ b/src/sharing/ShareToPm.js
@@ -162,12 +162,7 @@ class ShareToPm extends React.Component<Props, State> {
     let sharePreview = null;
     if (sharedData.type === 'image') {
       sharePreview = (
-        <Image
-          source={{ uri: sharedData.sharedImageUrl }}
-          width={200}
-          height={200}
-          style={styles.imagePreview}
-        />
+        <Image source={{ uri: sharedData.sharedImageUrl }} style={styles.imagePreview} />
       );
     }
 
--- a/src/start/IosCompliantAppleAuthButton/index.js
+++ b/src/start/IosCompliantAppleAuthButton/index.js
@@ -1,10 +1,11 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
-import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
+import type { ViewStyle } from 'react-native/Libraries/StyleSheet/StyleSheet';
 import * as AppleAuthentication from 'expo-apple-authentication';
 import { connect } from '../../react-redux';
 
+import type { SubsetProperties } from '../../generics';
 import Custom from './Custom';
 import type { ThemeName } from '../../reduxTypes';
 import type { Dispatch } from '../../types';
@@ -15,7 +16,15 @@ type SelectorProps = $ReadOnly<{|
 |}>;
 
 type Props = $ReadOnly<{|
-  style?: ViewStyleProp,
+  // See `ZulipButton`'s `style` prop, where a comment discusses this
+  // idea.
+  /* eslint-disable flowtype/generic-spacing */
+  style?: SubsetProperties<
+    ViewStyle,
+    {|
+      marginTop?: mixed,
+    |},
+  >,
   onPress: () => void | Promise<void>,
 
   dispatch: Dispatch,
--- a/src/caughtup/__tests__/caughtUpReducer-test.js
+++ b/src/caughtup/__tests__/caughtUpReducer-test.js
@@ -3,7 +3,7 @@ import deepFreeze from 'deep-freeze';
 
 import * as eg from '../../__tests__/lib/exampleData';
 import caughtUpReducer from '../caughtUpReducer';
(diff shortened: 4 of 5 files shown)


chrisbobbe:
Hmm, that CI failure might be transient:

```
Running android...
```
(lots of Unimodules output)
```
* What went wrong:
Execution failed for task ':app:mergeDebugAssets'.
> Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
   > Could not download appcompat-1.1.0.aar (androidx.appcompat:appcompat:1.1.0)
      > Could not get resource 'https://dl.google.com/dl/android/maven2/androidx/appcompat/appcompat/1.1.0/appcompat-1.1.0.aar'.
         > Connection reset
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
```

src/start/IosCompliantAppleAuthButton/index.js
```diff
@@ -15,7 +16,13 @@ type SelectorProps = $ReadOnly<{|
...
 
 type Props = $ReadOnly<{|
-  style?: ViewStyleProp,
+  /* eslint-disable flowtype/generic-spacing */
+  style?: SubsetProperties<
+    ViewStyle,
+    {|
+      marginTop?: mixed,
+    |},
+  >,
```
gnprice, line 25: Partly in the spirit of 8b30fa9d6, but also because this helps us
    avoid a mystifying error reported by Flow when we take the upgrade
    to Flow v0.122.0, with the React Native v0.63 upgrade (#4245).
    
    I was seeing the error on the use of `ZulipButton` in `AuthScreen`;
    it didn't like what I was passing for the `style` prop. When I
    temporarily commented out the `style` prop passed to
    `IosCompliantAppleAuthButton`, the error disappeared. It doesn't
    make a lot of sense to me, but it sounds like
    `IosCompliantAppleAuthButton`'s type for its `style` prop was
    influencing what type `styles.halfMarginTop` was inferred to have,
    when it was passed to `ZulipButton`.

Huh wacky. That seems like a Flow bug... quite likely reporting a real type mismatch, but a bug in at least where it places the error. I'd be curious to take a look at the error -- want to paste it somewhere, perhaps in chat?

In any case this seems like a fine tightening of the type. Perhaps add a comment pointing at ZulipButton's props, where a comment discusses this idea.

src/sharing/ShareToPm.js
```diff
@@ -162,12 +162,7 @@ class ShareToPm extends React.Component<Props, State> {
     let sharePreview = null;
     if (sharedData.type === 'image') {
       sharePreview = (
-        <Image
-          source={{ uri: sharedData.sharedImageUrl }}
-          width={200}
-          height={200}
-          style={styles.imagePreview}
-        />
```
gnprice, line 170: Worth a quick mention in the commit message that this isn't a change in Flow -- these props are there in the type in RN at our current version, and it's that type definition in RN that's getting stricter in the upgrade, to match the docs as you describe.

Here's the relevant bit of the type definition as of our current v0.62.x version:
```js
  /**
   * See https://facebook.github.io/react-native/docs/image.html#style
   */
  style?: ?ImageStyleProp,

  // Can be set via props or style, for now
  height?: ?DimensionValue,
  width?: ?DimensionValue,
```

.flowconfig
```diff
@@ -83,7 +83,6 @@ untyped-type-import=warn
 nonstrict-import=warn
 deprecated-type=warn
 unsafe-getters-setters=warn
-inexact-spread=warn
```
gnprice, line 86: Here's the explanation in the linked RN commit https://github.com/facebook/react-native/commit/87370c4c0 :
> After changes to the semantics of spread, this lint is no longer relevant and will be removed in an upcoming version of Flow.

Based on that, I think ideally we'd remove this after the upgrade -- that way we have the lint as long as we're still on versions of Flow where it might be relevant.

Is it causing an error or warning at the upgrade? If so it'd be fine to remove it before, but probably I'd still want it as part of the same PR/branch as the upgrade.

... OK, I went and did a quick history search (`git log --stat -p -S inexact-spread --pickaxe-all`) in the Flow repo, and it looks like
* the lint was outright removed, so it probably becomes an error to have it in .flowconfig, in v0.117 at https://github.com/facebook/flow/commit/759970c1b6dc0a25171d9969e34eaf5dae70f130;
* the lint had already stopped doing anything in v0.111 at https://github.com/facebook/flow/commit/0410e050246f43d280a793c8a153930390c3b6a7 -- or at least, it had stopped ever firing in any situation exercised by Flow's tests. (The later commit that removed it didn't have to touch the tests, except to remove the lint from a flowconfig.)

We're on v0.113.0, in between those -- so in fact now is a perfect time to remove this lint from our flowconfig.

(Seems like an oversight that this isn't mentioned in the Flow changelog, at https://github.com/facebook/flow/blob/master/Changelog.md#01170 .)

gnprice (commented):
Thanks @chrisbobbe ! Small comments below; after taking care of them, please merge at will.

I agree, that error message looks like a transient issue -- either at `dl.google.com`, or in Travis CI, or in the network between them. (So probably Travis.) Tests all pass for me locally now.

src/start/IosCompliantAppleAuthButton/index.js
chrisbobbe: Sure, I've just posted the error(s!) in chat, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20errors.20in.20.23M4318/near/1075420).

chrisbobbe:
> Thanks @chrisbobbe ! Small comments below; after taking care of them, please merge at will.

OK, done!

PR #4319, opened 2020-11-27
android build: Upgrade Gradle to 6.2.

chrisbobbe:
Part of the RN v0.62 -> v0.63 changes to the template app [1],
corresponding to the following commits:

- facebook/react-native@aa0ef1533
- facebook/react-native@a89a55353
- facebook/react-native@947a5593e
- facebook/react-native@d4d8887b5

I didn't find that the additions to android/gradlew.bat (from
facebook/react-native@d4d8887b5) were reproducible using our upgrade
command for any upgrade between 6.0.1 (where we were before this
commit) and 6.2.

But I could reproduce them by running that command to upgrade to
6.3. So, we might as well include them. It's possible that there are
some difference between the upgrade command we run and the command
React Native maintainers run, or (maybe more likely) React Native
initially went for an upgrade to 6.3 then partially backed it out,
to only go to 6.2.

The build scan revealed a handful of new deprecation notices not
flagged at our last Gradle upgrade (2c425116f), also for things that
will be removed in Gradle 7.0 [2].

[1] https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.3
[2] https://scans.gradle.com/s/3ocxcu7uxryfe/deprecations


chrisbobbe:
(Linking the RN v0.63 upgrade issue, #4245.)

gnprice:
> I didn't find that the additions to android/gradlew.bat (from
> [facebook/react-native@d4d8887](https://github.com/facebook/react-native/commit/d4d8887b5)) were reproducible using our upgrade
> command for any upgrade between 6.0.1 (where we were before this
> commit) and 6.2.
> 
> But I could reproduce them by running that command to upgrade to
> 6.3.

Remember what we discovered at the last Gradle upgrade, at https://github.com/zulip/zulip-mobile/pull/4244#discussion_r486789169:
> * Maybe the update to the `gradlew` script depends on the version you _start_ from, rather than the one you upgrade to?
> 
> It's at least pretty easy to imagine how that behavior could arise -- you have one version installed, you ask that version to upgrade, but while that existing version is still running it also goes and tries to update your wrapper scripts to the latest and best that it knows about. Not _good_ behavior but easy to imagine.

I think in fact I can confirm that that's what we're seeing. Once you're already on 6.2, if you *rerun* that "upgrade" command with the exact same arguments -- so "upgrade again" to 6.2 -- then you get that gradlew.bat change. The same is true if you upgrade to some other version like 6.3.

The other script change here, we actually get if you do the same with the version 6.0.1 we're already on.

I'll add a commit that does the 6.0.1 re-upgrade, and amends the instructions-comment.

gnprice:
Merged, thanks! I also bumped it further to the latest bugfix release of that 6.2.x series.

chrisbobbe:
> I think in fact I can confirm that that's what we're seeing. Once you're already on 6.2, if you _rerun_ that "upgrade" command with the exact same arguments -- so "upgrade again" to 6.2 -- then you get that gradlew.bat change. The same is true if you upgrade to some other version like 6.3.

Ohh, OK, right! That first commit, with the instruction to run the same command twice, is helpful. I remembered there was some oddness about how these upgrades worked, but I was fuzzy on the details. Thanks for the reminder, and for fixing it for the future. 🙂

PR #4322, opened 2020-12-01
Follow some of RN's Flipper changes before taking RN v0.63 upgrade.

chrisbobbe:
These are all of the Flipper-related changes that it makes sense (to me, currently) to take before the main-upgrade commit for RN v0.63 (#4245). When we get there, there will be large changes to make to our Podfile, mostly because React Native stops having us put lots of code (including Flipper-specific code) in our Podfile and instead maintains it themselves, in its `scripts/{autolink-ios.rb => react_native_pods.rb}` (that rename happens in facebook/react-native@4118d7982). Cool; glad not to have to write out a bunch of stuff in our Podfile, eventually.

The commits where that code is moved back to RN are these:

- facebook/react-native@a35efb940, for lots of RN-related, but not Flipper-related, pods, like "React-RCTImage".
- facebook/react-native@619d5d60d, for the Flipper-related pods [1].

We've long preferred to isolate some changes that make sense to do (and are possible to do) before a main RN upgrade commit. These commits do just that, but it's important to note that I've transposed some of the changes from `scripts/{autolink-ios.rb => react_native_pods.rb}` onto our Podfile because of the above. Those changes don't appear in the [RN Upgrade Helper](https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.4), which only looks at the template app. I still think they're quite important to do; facebook/react-native@29639e7b9, for example, has parts that are reflected in the Upgrade Helper that quite plausibly could be done before the main upgrade commit. But it's marked as depending on Flipper v0.41.0—an upgrade that gets done in that same commit, but (for iOS) in `scripts/{autolink-ios.rb => react_native_pods.rb}`, which we'd miss if we were only looking at the portion of that commit that touches the template app (and that makes it to the Upgrade Helper).

I'll post more about individual RN commits in some comments, to follow.

[1] It looks like this commit was incompletely backported to RN v0.62, in facebook/react-native@b4d1fcfb2; we took that incomplete backport in our 9a144c5b3. (At the time, I think I didn't realize this was an incomplete backport or even a backport.) I say it's incomplete because, even though it adds the same RN code to `scripts/{autolink-ios.rb => react_native_pods.rb}`, it doesn't invoke it; instead, it has the template app largely duplicate the code (with the Flipper version off by two patch versions; not sure why).


chrisbobbe:
The following is a list of commits filtered down (with order preserved) from the output of

```
git lsp v0.62.2..v0.63.4 --reverse -- template scripts/autolink-ios.rb scripts/react_native_pods.rb
```

in the RN repo. Filtered by...whether it seems plausibly related to Flipper, when I read the commit message and/or the changes. 🙂 

- facebook/react-native@f11b4895d: We already took this, in our 9a144c5b3, when it was backported to v0.62 in facebook/react-native@8f954b365.
- facebook/react-native@68bf0e32e: We already took this, in our f2e611995, when it was backported to v0.62 in facebook/react-native@3c3b4cf57.
- facebook/react-native@939df8a01: We already took this, in our 2fc3bc10a, when it was backported to v0.62 in facebook/react-native@e7bfa1fc2.
- facebook/react-native@52cd9cd6f: We already took this, in our 7c7271598, when it was backported to v0.62 in facebook/react-native@05f5cb534.
- facebook/react-native@8f93dedc6: We already took this, in our 9a144c5b3, when it was backported to v0.62 in facebook/react-native@e3218a0d9.
- **facebook/react-native@4bb0b4f20: Done in this PR branch.**
- facebook/react-native@619d5d60d: See [above](https://github.com/zulip/zulip-mobile/pull/4322#issue-530560415). This was incompletely backported to v0.62; see facebook/react-native@b4d1fcfb2. We took that incomplete backport in our 9a144c5b3. The following still needs to be done: calling RN's code instead of duplicating it in our Podfile. Best to do so in the main upgrade commit, I think.
- facebook/react-native@4118d7982: We should handle this in the main upgrade commit.
- facebook/react-native@fbb94a30b: We went past Flipper 0.32.2 in our f7005c50c.
- facebook/react-native@db1225564: We went past Flipper 0.33.1 on Android in our f7005c50c.
- facebook/react-native@85a4b0f5b: We went past Flipper 0.33.1 on iOS in our f7005c50c.
- facebook/react-native@e5497ca8f: We already took this, in our 9a144c5b3, when it was backported to v0.62 in facebook/react-native@8858d879e.
- facebook/react-native@87f5b8b79: We don't take this; it's cleanly reverted in facebook/react-native@574447ae3.
- facebook/react-native@574447ae3: Cleanly reverts facebook/react-native@87f5b8b79, which we didn't take; ignore.
- facebook/react-native@27ccc6019: We went past Flipper 0.37.0 in our f7005c50c.
- **facebook/react-native@29639e7b9: Done in this PR branch.**
- **facebook/react-native@a642c3f26: Done in this PR branch.**
- facebook/react-native@c72ecdb90: Haven't done this, but we could. It lets you specify which iOS build configurations you want Flipper to be active for, with ["Debug"] being the default. Currently, it's just hard-coded as "Debug". We just have the default two build configurations, "Release" and "Debug", and we don't immediately plan to make more ([this](https://chat.zulip.org/#narrow/stream/48-mobile/topic/measuring.20performance/near/1065928) is the only discussion I know of that might prompt adding more).
- **facebook/react-native@66c8cd393: Done in this PR branch.**

gnprice:
Thanks for this detailed and easy-to-follow analysis! This all looks great.

> The following still needs to be done: calling RN's code instead of duplicating it in our Podfile. Best to do so at or after the main upgrade commit, I think.

Yeah. When we do that, we'll be getting the version that's in the RN we're running. The version we're currently on doesn't have this code at all (not as a standalone, invokable script), so we can't really do this before the upgrade.

Probably best is as a commit coming in the upgrade PR after the upgrade commit.


> * [facebook/react-native@c72ecdb](https://github.com/facebook/react-native/commit/c72ecdb90): Haven't done this, but we could. It lets you specify which iOS build configurations you want Flipper to be active for, with ["Debug"] being the default. Currently, it's just hard-coded as "Debug". We just have the default two build configurations, "Release" and "Debug", and we don't immediately plan to make more ([this](https://chat.zulip.org/#narrow/stream/48-mobile/topic/measuring.20performance/near/1065928) is the only discussion I know of that might prompt adding more).

We'll effectively get this when we switch to using the script from RN, right? In that case probably cleanest to take this just as part of following along with RN's version of this code. Then when we switch over, the switch can be NFC or nearly so, which is nice.

Please merge at will after that!

chrisbobbe:
Thanks for the review!

> We'll effectively get this when we switch to using the script from RN, right? In that case probably cleanest to take this just as part of following along with RN's version of this code. Then when we switch over, the switch can be NFC or nearly so, which is nice.

Sure, makes sense! Done.

> Please merge at will after that!

OK, done! Thanks again for the review.

Issue #4306, opened 2020-11-12
Status icons on avatars lack contrast


gnprice:
This is a followup I originally posted at https://github.com/zulip/zulip-mobile/pull/3417#issuecomment-474641128 just before merging that PR.

The "unavailable" icon lacks contrast on top of an avatar:
  ![image](https://user-images.githubusercontent.com/28173/54651283-66523680-4a6f-11e9-915d-336ccbc6deb3.png)

* Possible fix: give it a partly-opaque background? Or even fully opaque.

* Something similar would probably benefit the other icons too -- if someone's avatar happens to have green or similar bright colors in the lower right, even the "active" color might not stand out. Perhaps a contrasting (white?) border would be good.

I just did a quick test and the behavior is the same as it was then, so this is still live.

abhi0504:
hey greg , 
I Think we can do two things or either of one of below for this issue

1. We can make our back ground partially-opaque for every status , 

2. Or , we can make our icon circular by increasing border radius and then put that status symbol half on avatar and half outside 
for eg. 
![moss_status (1)](https://user-images.githubusercontent.com/56453541/98977056-29895c00-253e-11eb-93dc-8d2c2a654efc.png)

gnprice:
Hi @abhi0504 !

Zulip's avatars are meant to be displayed as rounded squares, not circles. And quoting Tim at #205:
> This is based on a lot of consideration -- circular avatars are problematic because many images that we'd like to use in avatars (e.g. bot logos) don't actually look good in circular form, whereas just about anything looks good in a rounded square.

I think an opaque or partly-opaque background around each status icon, and perhaps an opaque border, would be a good solution. Your example image demonstrates that (in addition to the avatar shape) and I think it works well.

abhi0504:
@gnprice  thanks for your help , ok so we are keeping that rounded squares with partly-opaque background and perhaps an opaque border so should I started working this issue or any other suggestions

gnprice:
@abhi0504 You've already claimed #4307, so let's start with that, and then you can take this one up next.

abhi0504:
@gnprice I have also done some changes and submitted a PR for this issue as we have discussed 
## Before 

![](https://user-images.githubusercontent.com/28173/54651283-66523680-4a6f-11e9-915d-336ccbc6deb3.png)

## After 
[...]

PR #4324, opened 2020-12-02
Avatar-UI : giving a white background and cornering the status Indicator

abhi0504:
## **User's Avatar**

I have made 2 changes majorly

1. Gives a white background larger than the status indicator due to which no matter what will be the Avatar Image of the user the status indicator looks clear .
for example If the user's Avatar Image is greenish then the present indicator not looks that well 
Also above point can be easily understand by Greg's Issue #4306 

2. We need to check whether the status is offline or Its a Zulip's Bot and if it is then there is no need to add any white background as no status indicator is there but If we doesn't check then we get a white background over there 
                                                                              
### _not considering that If case_
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_58_02 (2)          Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_57_38 (2)

### _considering that If case_
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_58_23 (2)          Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_58_33 (2)

So after the above 2 changes Avatars of user's in User's List looks like this

### Before 
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_35 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_34_23 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_33_59 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_55 (2)

### After
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_22_58 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_23_27 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_25_15 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_26_06 (2)

Fixes : #4306

### DIFF
--- a/src/common/PresenceStatusIndicator.js
+++ b/src/common/PresenceStatusIndicator.js
@@ -17,6 +17,20 @@ const styles = createStyleSheet({
     height: 12,
     borderRadius: 6,
   },
+  status: {
+    bottom: 1.5,
+    right: 1.5,
+    position: 'absolute',
+  },
+  opaqueBackground: {
+    width: 15,
+    height: 15,
+    borderRadius: 7.5,
+    backgroundColor: 'white',
+    bottom: 0,
+    right: 0,
+    position: 'absolute',
+  },
   active: {
     backgroundColor: 'hsl(106, 74%, 44%)',
   },
@@ -81,6 +95,7 @@ type Props = $ReadOnly<{|
   style?: ViewStyleProp,
   email: string,
   hideIfOffline: boolean,
+  useOpaqueBackground: boolean,
 |}>;
 
 /**
@@ -95,7 +110,15 @@ type Props = $ReadOnly<{|
  */
 class PresenceStatusIndicator extends PureComponent<Props> {
   render() {
-    const { email, presence, style, hideIfOffline, allUsersByEmail, userStatus } = this.props;
+    const {
+      email,
+      presence,
+      style,
+      hideIfOffline,
+      allUsersByEmail,
+      userStatus,
+      useOpaqueBackground,
+    } = this.props;
 
     const userPresence = presence[email];
     const user = allUsersByEmail.get(email);
@@ -112,15 +135,44 @@ class PresenceStatusIndicator extends PureComponent<Props> {
 
     switch (status) {
       case 'active':
+        if (useOpaqueBackground) {
+          return (
+            <View style={styles.opaqueBackground}>
+              <PresenceStatusIndicatorActive style={styles.status} />
+            </View>
+          );
+        }
+
         return <PresenceStatusIndicatorActive style={style} />;
 
       case 'idle':
+        if (useOpaqueBackground) {
+          return (
+            <View style={styles.opaqueBackground}>
+              <PresenceStatusIndicatorIdle style={styles.status} />
+            </View>
+          );
+        }
         return <PresenceStatusIndicatorIdle style={style} />;
 
       case 'offline':
+        if (useOpaqueBackground) {
+          return (
+            <View style={styles.opaqueBackground}>
+              <PresenceStatusIndicatorOffline style={styles.status} />
+            </View>
+          );
+        }
         return <PresenceStatusIndicatorOffline style={style} />;
 
       case 'unavailable':
+        if (useOpaqueBackground) {
+          return (
+            <View style={styles.opaqueBackground}>
+              <PresenceStatusIndicatorUnavailable style={styles.status} />
+            </View>
+          );
+        }
         return <PresenceStatusIndicatorUnavailable style={style} />;
 
       default:
--- a/src/common/PresenceStatusIndicator.js
+++ b/src/common/PresenceStatusIndicator.js
@@ -27,9 +27,6 @@ const styles = createStyleSheet({
     height: 15,
     borderRadius: 7.5,
     backgroundColor: 'white',
-    bottom: 0,
-    right: 0,
-    position: 'absolute',
   },
   active: {
     backgroundColor: 'hsl(106, 74%, 44%)',
@@ -66,11 +63,16 @@ const styles = createStyleSheet({
 function MaybeOpaqueBackgroundWrapper(
   props: $ReadOnly<{|
     useOpaqueBackground: boolean,
+    style?: ViewStyleProp,
     children: React$Node,
   |}>,
 ) {
-  const { useOpaqueBackground, children } = props;
-  return useOpaqueBackground ? <View style={[styles.opaqueBackground]}>{children}</View> : children;
+  const { useOpaqueBackground, style, children } = props;
+  return useOpaqueBackground ? (
+    <View style={[styles.opaqueBackground, style]}>{children}</View>
+  ) : (
+    children
+  );
 }
 
 const PresenceStatusIndicatorActive = ({ style }: { style: ViewStyleProp }) => (
@@ -146,28 +148,28 @@ class PresenceStatusIndicator extends PureComponent<Props> {
     switch (status) {
       case 'active':
         return (
-          <MaybeOpaqueBackgroundWrapper useOpaqueBackground={useOpaqueBackground}>
+          <MaybeOpaqueBackgroundWrapper style={style} useOpaqueBackground={useOpaqueBackground}>
             <PresenceStatusIndicatorActive style={useOpaqueBackground ? styles.status : style} />
           </MaybeOpaqueBackgroundWrapper>
         );
 
       case 'idle':
         return (
-          <MaybeOpaqueBackgroundWrapper useOpaqueBackground={useOpaqueBackground}>
+          <MaybeOpaqueBackgroundWrapper style={style} useOpaqueBackground={useOpaqueBackground}>
             <PresenceStatusIndicatorIdle style={useOpaqueBackground ? styles.status : style} />
           </MaybeOpaqueBackgroundWrapper>
         );
 
       case 'offline':
         return (
-          <MaybeOpaqueBackgroundWrapper useOpaqueBackground={useOpaqueBackground}>
+          <MaybeOpaqueBackgroundWrapper style={style} useOpaqueBackground={useOpaqueBackground}>
             <PresenceStatusIndicatorOffline style={useOpaqueBackground ? styles.status : style} />
           </MaybeOpaqueBackgroundWrapper>
         );
 
       case 'unavailable':
         return (
-          <MaybeOpaqueBackgroundWrapper useOpaqueBackground={useOpaqueBackground}>
+          <MaybeOpaqueBackgroundWrapper style={style} useOpaqueBackground={useOpaqueBackground}>
             <PresenceStatusIndicatorUnavailable
               style={useOpaqueBackground ? styles.status : style}
             />
--- a/src/common/PresenceStatusIndicator.js
+++ b/src/common/PresenceStatusIndicator.js
@@ -63,6 +63,16 @@ const styles = createStyleSheet({
   },
 });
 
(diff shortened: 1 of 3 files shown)


abhi0504:
@chrisbobbe plz review this PR

gnprice:
Hi @abhi0504! Here's some comments from a quick look:
* Let's avoid duplicating here any of the logic from `PresenceStatusIndicator` that figures out what kind of status indicator is going to show.
  * Instead, let's have `PresenceStatusIndicator` itself provide this background. This would be controlled by a boolean prop so that it's enabled here but not in other places we use `PresenceStatusIndicator` where we might not want the background.
* The specific mention of `email === 'welcome-bot@zulip.com'` is a bad sign -- that is definitely not going to be the right logic for this. Whatever's going on for welcome-bot is going to be true of some other users too, so let's take the time to figure that out so we can have the right logic to cover them. (Similarly, it might be true of welcome-bot in your testing now but not true of welcome-bot in some other circumstances.)
* In your comment here on the issue: https://github.com/zulip/zulip-mobile/issues/4306#issuecomment-727253127 you had some screenshots where the status icons looked somewhat different from these. Is that from a previous version of your changes, or are there some places where it looks like that and others where it looks like your screenshots above?
* In the screenshots above, the white borders look pretty thick. Let's try them thinner -- like about the same thickness as the gray or orange borders in two of the status icons.

abhi0504:
Hi @gnprice Thanks for your comment , 
I am making a new commit which will clear all above points , this commit will do following things :

- Now Instead of duplicating the logic from `PresenceStatusIndicator` in `UserAvatarWithPresence` I am using a boolean prop `isUserAvatar` from `UserAvatarWithPresence` which tells whether `PresenceStatusIndicator` is used for User Avatar or not and if Its true then we are returning with that extra `View` which gives it a white background.

- Now there is also no need of specific mention of `email === 'welcome-bot@zulip.com'` as we are returning that white background only in the case of `IDLE` and `ACTIVE` user 
- In my comment here on the issue:[#4306](https://github.com/zulip/zulip-mobile/issues/4306#issuecomment-727253127) the screenshots that I have mentioned are from my previous version of changes , and the screen shots present above (in the current PR) are the main screen shots 
- Yes In the above screen shot the border is more thicker than the previous one which in my opinion looks good and gives it notch like feel but if you would save I can decrease its thickness like the [previous ones](https://github.com/zulip/zulip-mobile/issues/4306#issuecomment-727253127) or like about the same thickness as the gray or orange borders in two of the status icons as you have mentioned plz reply what looks good here 
- Also If you would say then I will squash all the commits into One by which this PR have only one final commit in it before merging

src/common/UserAvatarWithPresence.js
```diff
@@ -12,8 +12,8 @@ import PresenceStatusIndicator from './PresenceStatusIndicator';
 
 const styles = createStyleSheet({
   status: {
-    bottom: 0,
-    right: 0,
+    bottom: 4,
+    right: 4,
```
chrisbobbe, line 16:
```diff
-    bottom: 0,
-    right: 0,
+    bottom: 4,
+    right: 4,
```

Hmm, what's going on here? I haven't yet run the app with this revision, but it's not totally clear how this small code change contributes to the behavior change described in the commit message.

src/common/PresenceStatusIndicator.js
```diff
@@ -110,6 +128,22 @@ class PresenceStatusIndicator extends PureComponent<Props> {
       return null;
     }
 
+    if (isUserAvatar && status === 'active') {
```
chrisbobbe, line 131: Hmm, what's the reason that `isUserAvatar` is only checked for the `active` and `idle` statuses? How about `offline` and `unavailable` too?

chrisbobbe (changes requested):
Hi @abhi0504! Here are some comments from a quick look at the code just now.

src/common/PresenceStatusIndicator.js
```diff
@@ -17,6 +17,15 @@ const styles = createStyleSheet({
     height: 12,
     borderRadius: 6,
   },
+  commonbkg: {
```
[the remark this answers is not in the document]
abhi0504: sure I will changed it from commonbkg to opaqueBackground

src/common/PresenceStatusIndicator.js
```diff
@@ -81,6 +90,7 @@ type Props = $ReadOnly<{|
   style?: ViewStyleProp,
   email: string,
   hideIfOffline: boolean,
+  isUserAvatar: boolean,
```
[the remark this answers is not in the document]
abhi0504: sure I will changed it from isUserAvatar to useOpaqueBackground

src/common/UserAvatarWithPresence.js
abhi0504: this changes helps us to centralize the status indicator on the white background

src/common/PresenceStatusIndicator.js
abhi0504: for offline and unavailable statuses there is no need of white background required

abhi0504:
@gnprice and @chrisbobbe should I squash that second commit into first one or its ok for now. Also plz review the above changes

src/common/PresenceStatusIndicator.js
chrisbobbe: Let's do a white background for all four statuses. Like you've done in your "after" screenshots from https://github.com/zulip/zulip-mobile/pull/4324#issue-530742703, right? (Or, if not: can you say why you don't want to do it for "offline" or "unavailable" anymore? 🙂) Quoting from your message:

> ### Before
> Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_35 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_34_23 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_33_59 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_55 (2)
> 
> ### After
> Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_22_58 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_23_27 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_25_15 (2) Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_26_06 (2)

Under "After", I see an opaque white background for all status indicators.

-----

I noticed that, in this revision, the "unavailable" indicator still doesn't stand out very well:

image

Also, the "offline" indicator, when it appears—i.e., when `PresenceStatusIndicator` gets `false` for its `hideIfOffline` prop—doesn't stand out quite enough when the colors are similar. (I do see that `UserAvatarWithPresence` gives `true` for `hideIfOffline`. But `PresenceStatusIndicator` should do the right thing for all inputs it claims to accept.)

image

chrisbobbe:
> @gnprice and @chrisbobbe should I squash that second commit into first one or its ok for now.

Yes, please do squash the second commit into the first one. 😉

chrisbobbe (changes requested):
Thanks, @abhi0504! See some comments below, and another one [above](https://github.com/zulip/zulip-mobile/pull/4324#discussion_r551460168).

I'd also like to see a version with thinner borders; I like Greg's idea from https://github.com/zulip/zulip-mobile/pull/4324#issuecomment-745128290:
> about the same thickness as the gray or orange borders in two of the status icons

abhi0504:
@chrisbobbe Thank You so much for review .
So the things is we have two different versions of borders thinner and broader screenshot of both the versions are below
### Before The PR
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_35 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_34_23 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_33_59 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_32_55 (2)

### Broader or thick Borders
Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_22_58 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_23_27 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_25_15 (2)    Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_26_06 (2)

### Thinner Borders
Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_07_59     Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_08_51     Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_10_09     Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_10_36

The thing is If we have thin borders (the same thickness as the gray or orange borders) then there is no need of white Background for the   "Active" and "Offline" status as they are fully covered by status indicator .

So as screenshots for both the cases are above plz tell what looks good I will commit accordingly and squash all commits into one single commit .

chrisbobbe:
> The thing is If we have thin borders (the same thickness as the gray or orange borders) then there is no need of white Background for the "Active" and "Offline" status as they are fully covered by status indicator .

I think you might be confusing "thin borders" with "no borders"; I don't see any white border in any of the four screenshots under "Thinner borders".

I'm thinking thinner than this border

Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_26_06 (2)

But not so thin that the border disappears (as it does here):

Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_10_09

You could try a width of 1.5 or 2; I'm getting those numbers from `idleWrapper` and `unavailableWrapper`, which look like they control the width of the orange and gray borders in two of the status indicators.

abhi0504:
> > The thing is If we have thin borders (the same thickness as the gray or orange borders) then there is no need of white Background for the "Active" and "Offline" status as they are fully covered by status indicator .
> 
> I think you might be confusing "thin borders" with "no borders"; I don't see any white border in any of the four screenshots under "Thinner borders".
> 
> I'm thinking thinner than this border
> 
> Android Emulator - Pixel_3_API_29_5554 02-12-2020 09_26_06 (2)
> 
> But not so thin that the border disappears (as it does here):
> 
> Android Emulator - Pixel_3_API_29_5554 05-01-2021 01_10_09
> 
> You could try a width of 1.5 or 2; I'm getting those numbers from `styles.idleWrapper` and `styles.unavailableWrapper`, which look like they control the width of the orange and gray borders in two of the status indicators.

Oh Okay Now I understood Thanks @chrisbobbe 
I am making borders a little more broader than the last one also adding the screenshots of all four cases below 
Android Emulator - Pixel_3_API_29_5554 05-01-2021 16_18_31 (2)     Android Emulator - Pixel_3_API_29_5554 05-01-2021 20_48_14 (2)     Android Emulator - Pixel_3_API_29_5554 05-01-2021 20_59_33 (2)     Android Emulator - Pixel_3_API_29_5554 05-01-2021 21_00_30 (2)

Also You have mentioned some changes above which are also done and I squash all commits into one kindly plz review The changes and PR .

chrisbobbe:
Thanks, @abhi0504! I like the new thickness of the borders.

Merged, with a revised commit message and a few [NFC](https://github.com/zulip/zulip-mobile/blob/570db381dd518c37dba61f715ee35b4e3a89c35b/docs/glossary.md#nfc-or-nfc) commits on top of yours; please take a look! 🙂 The code at your commit works, but I saw that we could tighten up the logic by making it easier to see what precisely in `PresenceStatusIndicator` was controlled by `useOpaqueBackground`.

Issue #3745, opened 2019-12-31
Store server's version number, include in Sentry events


gnprice:
When an error happens only on certain versions of the Zulip server, that can be a powerful clue for diagnosing it. I've often noticed when looking at Sentry events that I'd like to have that information. #3732 is a recent example that brought this to mind.

In implementing this, one layer is to use the Sentry API to get that information attached. I'm not sure how best to do that; @ray-kraesig has studied those docs recently and may have ideas.

The other layer is to get and keep the server-version information around in the first place. The server reports its version number in the `server_settings` response, but we don't currently store it anywhere; the only things we use the `server_settings` response for are at initial login, so we fetch it once then. It wouldn't do to just store it then, either, because of course the version (and other parts of that response) can change over the months and years the user might stay logged in.

Here's a plausible approach:
* Add a property like `serverVersion` to the `Account` type, found in the `accounts` state-subtree. Can be `null | string`.
* In the login flow, save the version we get.
* At the same time we call `/register` aka `registerForEvents`, in `doInitialFetch` in `fetchActions.js`, make a `/server_settings` request again and save the new version value.

The server version number could also be useful if we decide to condition on it for certain API changes, as suggested at https://github.com/zulip/zulip-mobile/issues/3732#issuecomment-569468538 .

ray-kraesig:
> In implementing this, one layer is to use the Sentry API to get that information attached. I'm not sure how best to do that; @ray-kraesig has studied those docs recently and may have ideas.

Sounds like a job for either [tags](https://docs.sentry.io/enriching-error-data/context/?platform=javascript#tagging-events) or [extras](https://docs.sentry.io/enriching-error-data/context/?platform=javascript#extra-context). We probably want something like this when handling the `server_settings` response:
```js
Sentry.configureScope(scope => { scope.setTag("server-version", "2.2.7723"); });
```

Note that I haven't tried tags out yet, so I'm not sure what they look like in Sentry – but extras support is used in #3733.

gnprice:
> Sounds like a job for either [tags](https://docs.sentry.io/enriching-error-data/context/?platform=javascript#tagging-events) or [extras](https://docs.sentry.io/enriching-error-data/context/?platform=javascript#extra-context).

Thanks!

Between those two, I agree with your example that we probably want tags. Seems like the main difference is that tags get indexed, searched by, and aggregated on -- all of which sound like useful things for the server version.

[...]

PR #4326, opened 2020-12-03
logging: Send server version to Sentry.

chrisbobbe:
We include a few tags here:

- `rawServerVersion`: Just what it sounds like.
- `coarseServerVersion`: E.g., 2.1.x or 3.x
- `fineServerVersion`: E.g., 2.1.0 or 3.2

(Note that our numbering conventions changed, effective with 3.0, so
3.x and 4.x are each the same level of granularity as 2.1.x or
2.0.x.)

Also, update our custom libdef to include `scope.setTags`, which we
use here.

Fixes: #3745

### DIFF
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -2,9 +2,11 @@
 import type { Scope, SeverityType, EventHint } from '@sentry/react-native';
 import { getCurrentHub, Severity, withScope as withScopeImpl } from '@sentry/react-native';
 
+import store from '../boot/store';
 import type { JSONable } from './jsonable';
 import objectEntries from './objectEntries';
 import config from '../config';
+import { tryGetActiveAccount } from '../account/accountsSelectors';
 
 /** Type of "extras" intended for Sentry. */
 // This type should be exact, but cannot be until Flow v0.126.0. (See note in
@@ -20,6 +22,43 @@ function withScope<R>(callback: Scope => R): R {
   // Flow can't know that `ret` has actually been initialized
   return ((ret: $FlowFixMe): R);
 }
+
+/**
+ * Get server-version tags at various levels of granularity.
+ */
+const getServerVersionTags = () => {
+  const zulipVersion = tryGetActiveAccount(store.getState())?.zulipVersion;
+
+  // Why might we not have the server version? If there's no active
+  // account.
+  if (!zulipVersion) {
+    return {};
+  }
+
+  const raw = zulipVersion.raw();
+
+  const OMITTED = 'x';
+  const UNKNOWN = '?';
+
+  const elements = zulipVersion.elements();
+  const major = elements.major ?? UNKNOWN;
+  const minor = elements.minor ?? UNKNOWN;
+  const patch = elements.patch ?? UNKNOWN;
+
+  let coarseServerVersion;
+  let fineServerVersion;
+  // Effective with 3.0, we changed our numbering conventions; 3.x and
+  // 4.x are each the same level of granularity as 2.1.x or 2.0.x.
+  if (zulipVersion.isAtLeast('3.0')) {
+    coarseServerVersion = [major, OMITTED].join('.');
+    fineServerVersion = [major, minor].join('.');
+  } else {
+    coarseServerVersion = [major, minor, OMITTED].join('.');
+    fineServerVersion = [major, minor, patch].join('.');
+  }
+
+  return { rawServerVersion: raw, coarseServerVersion, fineServerVersion };
+};
 
 /**
  * Log an event (a string or Error) at some arbitrary severity.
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Following discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3745.20Server.20version.20to.20Sentry/near/1059031).

chrisbobbe:
Here's an example event: https://sentry.io/organizations/zulip/issues/2071017458/events/7bd84b976b11436696459ec371bc149c/

src/utils/logging.js
```diff
@@ -21,6 +23,43 @@ function withScope<R>(callback: Scope => R): R {
...
+  // 4.x are each the same level of granularity as 2.1.x or 2.0.x.
+  if (zulipVersion.isAtLeast('3.0')) {
+    result.coarseServerVersion = [major, OMITTED].join('.');
+    result.fineServerVersion = [major, minor].join('.');
+  } else {
+    result.coarseServerVersion = [major, minor, OMITTED].join('.');
+    result.fineServerVersion = [major, minor, patch].join('.');
+  }
+
+  return result;
```
gnprice, line 60: Let's put this in a form where it ends with something like
```js
return { rawServerVersion: raw, coarseServerVersion, fineServerVersion };
```
That way when you're looking down at the `setTags` call, and trying to see what are all the tags our code sets here, when you look up at this function you only have to glance at the `return` statements to see the inventory of them. (And then if there's one you're interested in knowing more about, you can follow up to where that value gets computed.)

src/utils/zulipVersion.js
```diff
@@ -26,32 +27,32 @@ type VersionElements = {
...
   _comparisonArray: number[];
-  _loggingArray: number[];
+  _elements: VersionElements;
 
   constructor(raw: string) {
     this._raw = raw;
-    this._comparisonArray = this._getComparisonArray(raw);
-    this._loggingArray = this._getLoggingArray(raw);
+    this._comparisonArray = this._getComparisonArray();
+    this._elements = this._getElements();
```
gnprice, line 35: Hmm. I like moving these methods away from being own-properties on the instances. But one thing I don't like about this change as it stands is that it causes these two functions to have implicit dependencies on the order in which we initialize things.

Looking at them again, I think what they really want to be is pure functions. Probably as static methods on this class, since they're closely tied to the details of its implementation. (The alternative would be as module-level, unexported functions.)

src/utils/zulipVersion.js
```diff
@@ -112,13 +113,13 @@ export class ZulipVersion {
...
-  };
+  }
 
   /**
    * Compute a number[] to be used in .isAtLeast comparisons.
    */
-  _getComparisonArray = (raw: string): number[] => {
-    const { major, minor, patch, flag, numCommits } = this._getElements(raw);
+  _getComparisonArray(): number[] {
+    const { major, minor, patch, flag, numCommits } = this._getElements();
```
gnprice, line 122: Another thing I notice as I'm rereading this code with attention to how these initialization methods depend on each other: this ends up unnecessarily calling `_getElements` twice. Instead, this function could take a `VersionElements`, which the constructor can cheerfully pass it.

gnprice (commented):
Thanks @chrisbobbe ! Glad to be adding this functionality. A few comments below (mostly on the follow-up refactor.)

src/utils/logging.js
chrisbobbe: What do you think about something like this:

```javascript
  // If there's no active account, we don't know the server version.
  if (zulipVersion) {
    // [...]
    return { rawServerVersion: raw, coarseServerVersion, fineServerVersion };
  } else {
    return {};
  }
```

That way, we can maintain that
- all values in the returned object are strings (none are undefined)
- it's quite easy for Flow to know the return type; either the returned object will have those three properties, or it will be empty.

chrisbobbe:
Thanks for the review! I just pushed a new revision.

src/utils/logging.js
```diff
@@ -21,6 +23,42 @@ function withScope<R>(callback: Scope => R): R {
...
+      fineServerVersion = [major, minor].join('.');
+    } else {
+      coarseServerVersion = [major, minor, OMITTED].join('.');
+      fineServerVersion = [major, minor, patch].join('.');
+    }
+
+    return { rawServerVersion: raw, coarseServerVersion, fineServerVersion };
+  } else {
+    return {};
+  }
```
gnprice, line 59: Let's make this an early return, to keep the main/happy/interesting path on a straight line. So like:
```js
  if (!zulipVersion) {
    return {};
  }

  const raw = zulipVersion.raw();
  // …
```

src/utils/logging.js
gnprice: Yeah, those are good things to maintain. Looks good except let's make the quick `return {}` case an early return, to avoid letting that edge-case tail wag the dog of how the main case is formatted and read. Also keeps the edge-case return next to the condition that leads to it.

gnprice:
LGTM except the pair of (closely related) comments above -- please merge at will after that. Thanks!

chrisbobbe:
> LGTM except the pair of (closely related) comments above -- please merge at will after that. Thanks!

OK, done! Thanks!

Issue #4253, opened 2020-09-10
Upgrade to ESLint 7.


chrisbobbe:
Filing this quickly as a placeholder so we have something to refer to.

Currently, this is known to be blocked by prettier-eslint not supporting ESLint 7 yet: https://github.com/prettier/prettier-eslint/issues/401.

chrisbobbe:
Version 12.0.0 of `prettier-eslint` [reportedly](https://github.com/prettier/prettier-eslint/pull/407#issuecomment-731610600) just got ESLint 7 support, so I'm unblocking this! 🎉

PR #4327, opened 2020-12-04
deps: Upgrade ESLint-related deps to their latest.

chrisbobbe:
Done with

```
yarn upgrade --latest $(tools/test deps | grep eslint)
```

which was easier than 01593b3af.

One thing we neglected to mention in 01593b3af, which still applies
here, is that the version of ESLint used by `prettier-eslint-cli` is
still 5, and it's unclear when (or if) that'll change [1], which is
unfortunate.

[1] https://github.com/prettier/prettier-eslint-cli/issues/304

Fixes: #4253


gnprice:
I get a few lint errors on this branch:
```
$ tools/test --full lint prettier
Running lint...

/home/greg/z/mobile/src/message/__tests__/fetchActions-test.js
  312:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect
  313:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect
  338:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect
  363:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect
  405:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect
  498:9  error  Avoid calling `expect` conditionally`  jest/no-conditional-expect

✖ 6 problems (6 errors, 0 warnings)

Running prettier...
501 files were unchanged

FAILED: lint
```

Reading that rule's docs:
https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-conditional-expect.md
they're really aimed at something we're not doing here. But what we are doing isn't really natural to express as a conditional like this, either:
```js
      expect(action.type).toBe('MESSAGE_FETCH_START');
      if (action.type === 'MESSAGE_FETCH_START') {
        expect(action.numBefore).toBeGreaterThan(0);
        expect(action.numAfter).toBeGreaterThan(0);
      }
```
What we really intend there is an assertion -- we're sure that at that point the condition will be true, and the conditional is there just to let Flow know that.

Happily, I've recently learned of a convenient way to express such a condition to Flow! I went and made a small demo for adding it to our codebase, and it turned into a bigger demo. Take a look: #4329 The last commit in that series makes the small fix we need here for these tests.

chrisbobbe:
> I get a few lint errors on this branch:

Thanks for finding these! I think I forgot to use `--full`.

gnprice:
Rebased now that #4329 is merged, and all looks good -- merged!

PR #4329, opened 2020-12-05
Introduce `invariant`; use it to handle display_recipient more cleanly.

gnprice:
This starts us using Flow's support for an assertion function named `invariant`, which I learned about recently: https://github.com/facebook/flow/issues/6052

The immediate inspiration is the last commit in this series, a small test cleanup which clears things up for #4327. I thought a bit about more substantial ways to make use of it, and landed on a refactoring of how we handle the notorious `display_recipient` property on messages: at the end of this branch, when we consume `display_recipient` we always do so through a pair of helpers in `recipient.js`, which explicitly check our assumptions and return well-typed results.

There's also a few other cleanups in here which either came up as prerequisites for the main changes, or were suggested by them and quick to add in.

### DIFF
--- a/src/message/__tests__/fetchActions-test.js
+++ b/src/message/__tests__/fetchActions-test.js
@@ -1,4 +1,5 @@
 /* @flow strict-local */
+import invariant from 'invariant';
 import configureStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import omit from 'lodash.omit';
@@ -308,10 +309,9 @@ describe('fetchActions', () => {
       expect(actions.length).toBeGreaterThanOrEqual(1);
       const [action] = actions;
       expect(action.type).toBe('MESSAGE_FETCH_START');
-      if (action.type === 'MESSAGE_FETCH_START') {
-        expect(action.numBefore).toBeGreaterThan(0);
-        expect(action.numAfter).toBeGreaterThan(0);
-      }
+      invariant(action.type === 'MESSAGE_FETCH_START', 'expect failed');
+      expect(action.numBefore).toBeGreaterThan(0);
+      expect(action.numAfter).toBeGreaterThan(0);
     });
 
     test('when no messages to be fetched before the anchor, numBefore is not greater than zero', async () => {
@@ -334,9 +334,8 @@ describe('fetchActions', () => {
       expect(actions.length).toBeGreaterThanOrEqual(1);
       const [action] = actions;
       expect(action.type).toBe('MESSAGE_FETCH_START');
-      if (action.type === 'MESSAGE_FETCH_START') {
-        expect(action.numBefore).not.toBeGreaterThan(0);
-      }
+      invariant(action.type === 'MESSAGE_FETCH_START', 'expect failed');
+      expect(action.numBefore).not.toBeGreaterThan(0);
     });
 
     test('when no messages to be fetched after the anchor, numAfter is not greater than zero', async () => {
@@ -359,9 +358,8 @@ describe('fetchActions', () => {
       expect(actions.length).toBeGreaterThanOrEqual(1);
       const [action] = actions;
       expect(action.type).toBe('MESSAGE_FETCH_START');
-      if (action.type === 'MESSAGE_FETCH_START') {
-        expect(action.numAfter).not.toBeGreaterThan(0);
-      }
+      invariant(action.type === 'MESSAGE_FETCH_START', 'expect failed');
+      expect(action.numAfter).not.toBeGreaterThan(0);
     });
   });
 
@@ -401,9 +399,8 @@ describe('fetchActions', () => {
       expect(actions).toHaveLength(1);
       const [action] = actions;
       expect(action.type).toBe('MESSAGE_FETCH_START');
-      if (action.type === 'MESSAGE_FETCH_START') {
-        expect(action.numBefore).toBeGreaterThan(0);
-      }
+      invariant(action.type === 'MESSAGE_FETCH_START', 'expect failed');
+      expect(action.numBefore).toBeGreaterThan(0);
     });
 
     test('when caughtUp older is true, no action is dispatched', async () => {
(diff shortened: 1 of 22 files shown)


src/types.js
```diff
@@ -2,6 +2,7 @@
 import type { IntlShape } from 'react-intl';
 import type { DangerouslyImpreciseStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 
+import type { SubsetProperties } from './generics';
 import type { Auth, Topic, Message, PmRecipientUser, Reaction, ReactionType } from './api/apiTypes';
```
chrisbobbe, line 6: d220cc82f types: Use SubsetProperties to base Outbox more directly on Message.

`tools/test lint` tells me `Reaction` and `PmRecipientUser` are defined but never used, in this commit.

src/outbox/outboxActions.js
```diff
@@ -121,12 +113,10 @@ const mapEmailsToUsers = (emails, allUsersByEmail, ownUser) => {
   return result;
 };
 
-// TODO type: `string | NamedUser[]` is a bit confusing.
-type DataFromNarrow = {|
-  type: 'private' | 'stream',
-  display_recipient: string | NamedUser[],
-  subject: string,
-|};
+type DataFromNarrow = SubsetProperties<
```
chrisbobbe, line 116: > In fact, use our SubsetProperties type-helper to get
the real thing directly from the actual Outbox type so we don't
have to duplicate the definition.

src/utils/__tests__/narrow-test.js
```diff
@@ -322,7 +329,8 @@ describe('getNarrowFromMessage', () => {
     expect(actualNarrow).toEqual(expectedNarrow);
   });
 
-  test('if recipient of a message is string, returns a stream narrow', () => {
+  test('for stream message with empty topic, returns a stream narrow', () => {
+    // TODO this behavior seems pretty dubious
```
chrisbobbe, line 331: > `// TODO this behavior seems pretty dubious`

Indeed! 🙂 See https://github.com/zulip/zulip-mobile/commit/5bf51c7c14427508e00ea2fbf7dc4814da511302 on #M4317, which removes it.

src/utils/narrow.js
```diff
@@ -304,18 +311,23 @@ export const canSendToNarrow = (narrow: Narrow): boolean =>
...
-  if (message.subject && message.subject.length) {
-    return topicNarrow(message.display_recipient, message.subject);
+/**
+ * Careful: quirky behavior on a stream message with empty topic.
+ *
+ * If you think you want to reuse this function: study carefully; maybe
+ * refactor it to something cleaner first; if not, then definitely document
+ * its quirky behavior.
+ */
+// TODO: do that, or just make this a private local helper of its one caller
```
chrisbobbe, line 321: > // TODO: do that, or just make this a private local helper of its one caller

Yeah, this might be the way to go. #4317 contains an attempt to clean it up, but I'm not sure about the value of that particular attempt, especially since there is just one caller, as you mention.

chrisbobbe:
This looks great! `invariant` reminds me of "User-defined Type Guards" in [TypeScript](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), though I'm sure there are important differences.

I've made just a few comments, all small. It's so nice to be able to clean up `display_recipient`. IIUC, #3133 (among our current [priorities](https://chat.zulip.org/#narrow/stream/48-mobile/topic/priorities/near/1000872)) got stalled on cleaning up something in `recipients.js`; I'm not yet clear on whether it's this sort of cleanup, but I'm sure it wouldn't hurt! I happen to have just rebased #3535 (at a separate branch; see https://github.com/zulip/zulip-mobile/pull/3535#pullrequestreview-545299959), but I did so before (re)discovering the potentially preferable #4104, oops.

src/message/__tests__/fetchActions-test.js
```diff
@@ -334,9 +334,8 @@ describe('fetchActions', () => {
       expect(actions.length).toBeGreaterThanOrEqual(1);
       const [action] = actions;
       expect(action.type).toBe('MESSAGE_FETCH_START');
-      if (action.type === 'MESSAGE_FETCH_START') {
-        expect(action.numBefore).not.toBeGreaterThan(0);
-      }
+      invariant(action.type === 'MESSAGE_FETCH_START', 'expect failed');
```
chrisbobbe, line 337: Nice! Yeah, I felt weird adding these conditionals. But less weird than adding `$FlowFixMe`s. Glad to have this solution.

src/types.js
gnprice: > `tools/test lint` tells me `Reaction` and `PmRecipientUser` are defined but never used, in this commit.

Ah, thanks for the catch! Fixed.

gnprice:
Merged, with that one fix. Thanks for the review!

> It's so nice to be able to clean up `display_recipient`. IIUC, #3133 (among our current [priorities](https://chat.zulip.org/#narrow/stream/48-mobile/topic/priorities/near/1000872)) got stalled on cleaning up something in `recipients.js`; I'm not yet clear on whether it's this sort of cleanup, but I'm sure it wouldn't hurt! I happen to have just rebased #3535 (at a separate branch; see [#3535 (review)](https://github.com/zulip/zulip-mobile/pull/3535#pullrequestreview-545299959)), but I did so before (re)discovering the potentially preferable #4104, oops.

Yeah. It did stall on that when I picked it up in... February, IIRC. I then spent a solid few days studying the `display_recipients` and related "how to name a PM conversation" messes. A lot of that has landed since then, some is still in draft branches, and this branch overlaps with some of those but the main idea (with the two helpers to explicitly get a stream name or a PM recipients list) is something I hadn't tried before.

I'm not sure offhand how the bits that have landed compare to what was blocking #3133, or what might still be needed for that; I'd have to go take a look to page that back into my head.

PR #4330, opened 2020-12-08
narrow: Treat group PMs and 1:1 PMs more symmetrically.

gnprice:
We have two "constructors" `privateNarrow` and `groupNarrow` for 1:1 PMs and group PMs respectively, but they end up building basically the same data structure.  Then in a lot of places in our code we have control flow to handle the group-PM case on the one hand and the 1:1 case on the other, and it ends up just saying the same thing twice to treat them both the same way.  (In a few places, like `getNarrowFromMessage`, we have no such logic and already implicitly rely on the fact that they're the same underneath.)

In general the Zulip model is that group PMs and 1:1 PMs are very much alike, and the Zulip API generally treats them alike.  So let's move toward more consistently handling them alike, and only having conditionals to distinguish between them when we actually want to do something different.

I wrote the bulk of these changes back in February, as part of a larger draft branch for refactoring our handling of narrows. Prompted today by https://github.com/zulip/zulip-mobile/pull/4317#discussion_r537995930 , I went and took another look at that branch. Here's a portion of it -- rebased forward, and with proper commit messages added where it didn't have them. (For some other portions of the original branch, I'm either not convinced they're the right direction or they just need a bit more work.)

### DIFF
--- a/src/chat/__tests__/narrowsSelectors-test.js
+++ b/src/chat/__tests__/narrowsSelectors-test.js
@@ -9,7 +9,7 @@ import {
 import {
   HOME_NARROW,
   HOME_NARROW_STR,
-  privateNarrow,
+  pmNarrowFromEmail,
   streamNarrow,
   topicNarrow,
   STARRED_NARROW,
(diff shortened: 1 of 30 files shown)


src/chat/__tests__/narrowsSelectors-test.js
```diff
@@ -278,7 +278,7 @@ describe('isNarrowValid', () => {
     expect(result).toBe(false);
   });
 
-  test('narrowing to a group chat with non-existing user is not valid', () => {
+  test('narrowing to a group chat with existing users is valid', () => {
```
chrisbobbe, line 281: 4f14e6bcc narrow: Fix isNarrowValid for group PMs.

```
Oddly, there was a test checking that this behavior was this way, so
it was noticed before.  It's not clear why it was thought desirable,
though.  The test was added in cefbe6d in 2018-05, long after the
logic itself was written in b0c7323 in 2017-12, and neither commit
explains the discrepancy.
```

Weird!

chrisbobbe:
LGTM, merged, thanks!

Issue #4323, opened 2020-12-01
Open links in default browser, not embedded webview


jryans:
We should change tapping on links to open them in the default browser, not the webview. (description edited by @WesleyAC, original below)

---

In many iOS apps these days, tapping links will open them by default in your preferred browser (recently added as option in iOS 14).

Zulip on iOS doesn’t appear to offer this, and only has an embedded webview, which adds extra steps, since each link needs to be shared over to the actual browser.

Please add some route to open links in the user’s chosen default browser, preferably by just tapping the link.

shaurya2612:
I am working in this

shaurya2612:
I am working on this !

gnprice:
Thanks @jryans for the report!

It looks like we actually switched from opening in Safari to an embedded Safari view, a few years ago, in #1721. There's no record of what the motivation for that was, though, unfortunately.

I'm not personally a real frequent iOS user, so I don't have a strong view myself on the best UX here. From a quick web search, I found this article fairly persuasive that an embedded Safari view might have been best a few years ago but opening in a full browser is the best choice now:
https://pspdfkit.com/blog/2019/open-links-in-safari-not-safari-view-controller/
So I agree, we should make this change.

WesleyAC:
I think we should do this for both iOS and Android, see #4562 for the reasons there. I'm going to edit this issue to be for both, rather than iOS specific.

gnprice:
For future reference: We ended up making this change by default on iOS, but leaving the in-app browser the default on Android, and offering a setting in either case. Rationale in [chat discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1143087).

PR #4336, opened 2020-12-12
Open urls in browser not WebView

shaurya2612:
#4323 replaced NativeModules with Linking to open the url in a browser instance instead of a WebView

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -38,12 +38,12 @@ PODS:
     - Flipper-Folly (~> 2.2)
     - Flipper-RSocket (~> 1.1)
   - Flipper-DoubleConversion (1.1.7)
-  - Flipper-Folly (2.2.0):
+  - Flipper-Folly (2.3.0):
     - boost-for-react-native
     - CocoaLibEvent (~> 1.0)
     - Flipper-DoubleConversion
     - Flipper-Glog
-    - OpenSSL-Universal (= 1.0.2.19)
+    - OpenSSL-Universal (= 1.0.2.20)
   - Flipper-Glog (0.3.6)
   - Flipper-PeerTalk (0.0.4)
   - Flipper-RSocket (1.1.0):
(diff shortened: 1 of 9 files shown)


shaurya2612:
Previous WebView Screen - 
Screen Shot 2020-12-12 at 12 42 40 PM

After using the Linking API -
Screen Shot 2020-12-12 at 12 43 42 PM

gnprice:
Thanks @shaurya2612 for the PR and the screenshots!

Here's a few comments on this change:
* This is something that'll be different between Android and iOS -- in fact the existing code has a conditional to do different things for each one. So we'll need to evaluate the change separately for each platform. To keep things simple, let's focus on just one for now, so let's start with a version that doesn't change what we do on Android.
* This `openLink` function is used in a number of places in the app. You can find them through the IDE, or with a command-line search like `rg -C2 'openLink\('`. We should make sure the change makes sense for all of them and won't break how they work.
  * In particular there are two calls to it in our code for signing into a Zulip account, in `AuthScreen.js` and `webAuth.js`. I suspect that for those two, using an external browser will be a worse experience, and we should stick with the embedded Safari view for those.
  * There's also a couple of calls in our code for the lightbox, where the user is trying to download or share an image, and there's a call in `messageLinkPress` that corresponds to when the user taps on an image that's part of a message. Here I'm not sure if the external browser will be a better or worse experience.

We can have two versions of this function, named like `openLink` and `openLinkEmbedded`, where the latter uses a Safari view. (And, at least for now, on Android both functions will still use an embedded browser view.) Then we can call the two versions in different places.

Then the main work that'll be needed in order to finish this from here is to go through each of the different flows that use `openLink` to open a link, and test how the experience is with this change so we can decide which ones it makes sense to use the external browser for and which ones we should continue to use the Safari view for.

shaurya2612:
This PR has been continued at #4343 as the current branch faced some rebasing issues.

PR #4341, opened 2020-12-15
UserAvatarWithPresence: Remove dubious default prop for `email`.

chrisbobbe:
As Greg points out, the effect of this default prop has been to
suppress the presence-status indicator for the sender in the
lightbox header. And since that omission is quite unlikely to be
intentional, we should fix it by passing down the sender's email.
So, do.

[1] https://github.com/zulip/zulip-mobile/pull/4230#discussion_r542953311


gnprice:
Looks good, merged -- thanks!

Issue #3916, opened 2020-02-19
[Pending server work] Support downgrading to a long-lived event queue.


chrisbobbe:
Tim mentions at https://chat.zulip.org/#narrow/stream/137-feedback/topic/mobile/near/818555 that https://github.com/zulip/zulip/pull/12926 is the PR thread tracking a server feature for the ability to downgrade to a long-lived event queue that ignores high-traffic events.

This would cover https://github.com/zulip/zulip-mobile/issues/1804 by making it unnecessary to fetch all users at startup; long initial load times are a common problem for users (see also the czo link, above), but I've made a new issue because there are separate benefits and implementation details to consider.

gnprice:
> Unrecognized events will be ignored all the way back to mobile commit [d2836e1](https://github.com/zulip/zulip-mobile/commit/d2836e1277a94e5721636064d9ee1f57e481d7b9), 2017-03-11, v1.0.11 (see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/update_message.20events/near/818281)).

It looks like that reference was about events with the `update_message` type, but with a new field.

We do ignore events of unexpected type (in `responseToActions` in `eventActions.js`), though, and I think we have for a long time. Fortunately also it doesn't matter how long we have (or even if we already do or not), because this will be a feature that new clients opt into.

gnprice:
This blocks on #4841 and #3408.

Then when the blockers are cleared, the work of this issue will be to a large extent on the server, as well as in the client.

PR #4344, opened 2020-12-16
Address a performance regression from #4230 by deferring `new URL` computations.

chrisbobbe:
Screen Shot 2020-12-16 at 3 11 09 PM

Following some profiling with Chrome DevTools, and [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1081038) on CZO, keep going in the direction of deferring `new URL` computations, since the computations done in `api.registerForEvents` have made the initial fetch take quite a lot longer after #4230 than before, and URL computations were observed to be a big part of that in profiling.

The above measurements are taken with release builds on my 4+ year-old iPhone 6s running iOS 13.7, on CZO, at notable commits (see the chart's key):

- 998947d807 (narrow [nfc]: Unexport pmNarrowFromEmails!) — Before #4230
- c642677685 (registerForEvents: Start sending `user_avatar_url_field_optional`.) — Tip of #4230
- Then each commit in this series. Note that they're based on #4230; I haven't rebased them atop `master`, which contains more recent commits.

The logging code is the following (also pasted in [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1080938)):

```diff
diff --git src/message/fetchActions.js src/message/fetchActions.js
index a610c11ea..46cb1e189 100644
--- src/message/fetchActions.js
+++ src/message/fetchActions.js
@@ -4,6 +4,7 @@ import type { Narrow, Dispatch, GetState, GlobalState, Message, Action } from '.
 import type { ApiResponseServerSettings } from '../api/settings/getServerSettings';
index a610c11ea..46cb1e189 100644
--- src/message/fetchActions.js
+++ src/message/fetchActions.js
@@ -4,6 +4,7 @@ import type { Narrow, Dispatch, GetState, GlobalState, Message, Action } from '.
 import type { ApiResponseServerSettings } from '../api/settings/getServerSettings';
 import type { InitialData } from '../api/initialDataTypes';
 import * as api from '../api';
+import * as logging from '../utils/logging';
 import { resetToMainTabs } from '../actions';
 import { isClientError } from '../api/apiErrors';
 import {
@@ -309,6 +310,7 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
   let initData: InitialData;
   let serverSettings: ApiResponseServerSettings;
 
+  const before = Date.now();
   try {
     [initData, serverSettings] = await Promise.all([
       tryFetch(() =>
@@ -331,6 +333,8 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
     dispatch(logout());
     return;
   }
+  const after = Date.now();
+  logging.logToProfiling({ durationMs: after - before, ip: '192.168.1.7' });
```

It seems to confirm our [expectation](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1081247) that the biggest difference is made by removing the `new URL` in `FallbackAvatarURL`'s `validateAndConstructInstance`. That's 9ec9e20684 in this revision.

Also, removing a `new URL` from `fromUserOrBotData` (which was used to discern which subclass to use) is helpful, as [expected](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1081272). That's 1d3c38a725 in this revision.

Removing two `new URL`s in `GravatarURL`'s `validateAndConstructInstance` seems definitely helpful; that's 1de50ccd49 in this revision.

The remaining changes in this revision, 720547ff2b and 190eb154f4...aren't large improvements, it seems. The former might have shortened the time by just a hair; the latter (the tool chose purple, like the before-#4230 measurements, oops) actually ended up with a larger mean than the mean for 1de50ccd49.

Unfortunately, nothing was able to clearly beat the before-#4230 measurements (the lowest purple line shows its mean), but it's close.

### DIFF
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -3,7 +3,6 @@
 /* eslint-disable no-use-before-define */
 import md5 from 'blueimp-md5';
 
-import { tryParseUrl } from './url';
 import * as logging from './logging';
 import { ensureUnreachable } from '../types';
 
@@ -53,7 +52,16 @@ export class AvatarURL {
       // which to generate the correct hash; see
       // https://github.com/zulip/zulip/issues/15287. Implemented at
       // `do_events_register` in zerver/lib/events.py on the server.)
-      if (tryParseUrl(rawAvatarUrl)?.origin === GravatarURL.ORIGIN) {
+      if (
+        rawAvatarUrl.startsWith(
+          // Best not to use an expensive `new URL` call, when the
+          // following is equivalent (assuming `rawAvatarUrl` is also
+          // valid through its /pathname, ?query=params, and
+          // #fragment). The trailing slash ensures that we've
+          // examined the full origin in `rawAvatarUrl`.
+          `${GravatarURL.ORIGIN}/`,
+        )
+      ) {
         const hashMatch = /[0-9a-fA-F]{32}$/.exec(new URL(rawAvatarUrl).pathname);
         if (hashMatch === null) {
           const msg = 'Unexpected Gravatar URL shape from server.';
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -62,13 +62,7 @@ export class AvatarURL {
           `${GravatarURL.ORIGIN}/`,
         )
       ) {
-        const hashMatch = /[0-9a-fA-F]{32}$/.exec(new URL(rawAvatarUrl).pathname);
-        if (hashMatch === null) {
-          const msg = 'Unexpected Gravatar URL shape from server.';
-          logging.error(msg, { value: rawAvatarUrl });
-          throw new Error(msg);
-        }
-        return GravatarURL.validateAndConstructInstance({ email, hash: hashMatch[0] });
+        return GravatarURL.validateAndConstructInstance({ email, urlFromServer: rawAvatarUrl });
       }
 
       // Otherwise, it's a realm-uploaded avatar, either absolute or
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Next, I'll make a similar chart for the Recurse realm. I expect to get different results than we see on CZO because Recurse has EMAIL_ADDRESS_VISIBILITY_EVERYONE set.

chrisbobbe:
OK, here's the same thing but on the Recurse realm instead of CZO:

image

Looks like 9ec9e20684 (the one for FallbackAvatarURL) is less helpful on Recurse. The "fallback" case kicks in when `.avatar_url` is missing; and, while the contract is that (if the client sets `user_avatar_url_field_optional`) the server can omit it at its sole discretion, we happen to know (I think it's still true) that the current implementation omits it for long-term idle users. It seems likely that there are more long-term idle users on CZO than Recurse, which might be a reason we see this difference in effectiveness.

Looks like 1de50ccd49 is a lot less helpful (if it's helpful at all) on Recurse. I expect that's because Recurse has EMAIL_ADDRESS_VISIBILITY_EVERYONE set, while CZO doesn't. 1de50ccd49 improves the case where the server sends the full Gravatar URL string, which it does if EMAIL_ADDRESS_VISIBILITY_EVERYONE is not set.

Looks like 190eb154f4 is a lot more helpful on Recurse than on CZO. That optimization affects uploaded avatar URLs that are absolute URLs not on the realm. Recurse sends those absolute URLs (it uses the s3 backend), while CZO does not; that's probably why.

chrisbobbe:
> Looks like, still, nothing's quite as fast as it was before #4230, hmm.

I think we do expect the initial load to speed up dramatically once we start using long-lived event queues (#3916), in any case. That might be a long way off, though.

src/utils/avatar.js
```diff
@@ -314,14 +314,32 @@ export class UploadedAvatarURL extends AvatarURL {
...
+      // not depend on `this._standardUrl` being present, and that
+      // might get messy.
+      //
+      // If it's absolute, we can defer `new URL` to the first
+      // `.get()` call: we should be able to trust that the server
+      // gives us properly formatted URLs; if it didn't, it seems like
+      // the kind of bug that would be fixed quickly.
+      absoluteOrRelativeUrl.startsWith('/')
+        ? new URL(absoluteOrRelativeUrl, realm)
+        : absoluteOrRelativeUrl,
```
gnprice, line 341: A couple of thoughts on this one:
* This logic is completely correct that if the URL string starts with a `/`, it's definitely a relative-URL string (provided it's any kind of valid URL string, and I guess has no fragment). With just a bit more logic, I think we can actually be more specific: we can make it definitely a ["path-absolute-URL string"](https://url.spec.whatwg.org/#path-absolute-url-string), and then we can say `${realm.origin}${absoluteOrRelativeURL}` and skip the `new URL` here too.
* If the string doesn't start with a slash, it isn't necessarily absolute; it could be one of a few other flavors of relative-URL strings. I think *probably* a well-behaved server will never want to give us one of those... but I'm not sure of that. So in the spirit of the crunchy shell, I'd like to validate it just a little more to be sure it's absolute, and if it's neither absolute nor the flavor of relative that we expect, then give an error.

How about I try developing a commit for this uploaded-avatar case? I have the [URL Standard](https://url.spec.whatwg.org/#url-parsing) open, and I think I can make something short and simple that reliably does the small part of it we need here.

gnprice (commented):
Thanks for these changes and the detailed timing information!

These all look good to me except the last one, 190eb154f41d0bfe6e164bd94184ca7119c7dae6, for uploaded avatars -- comments on that one below. I'll go ahead and merge all the others, and then I'll see about trying to write an improved version of that one.

I do think we should be able to continue to speed this up so that the initial load is faster than before #4230, rather than slower. The purpose of the `user_avatar_url_field_optional` feature was all about speeding it up, after all. So for the next steps, I'm curious what a profile shows about where the time is going after these changes. We can continue that discussion [in the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1082039).

src/utils/avatar.js
chrisbobbe: > How about I try developing a commit for this uploaded-avatar case? I have the [URL Standard](https://url.spec.whatwg.org/#url-parsing) open, and I think I can make something short and simple that reliably does the small part of it we need here.

Cool, sounds good, thanks! 🙂

PR #4345, opened 2020-12-17
    avatar: Remove remaining URL construction for initial load.

gnprice:
This follows up on #4344 to address part of the remaining performance regression from #4230.

We still construct URL objects in the `get` methods, lazily and memoized; but never up front when simply taking in all the initial data from the server, or the rehydrated data we've stored.  This should save a further slice of time in the initial load.

To do so, we introduce predicates `isUrlAbsolute` and `isUrlPathAbsolute`, which have very simple implementations but clearly defined semantics in terms of the URL Standard.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -106,10 +106,10 @@ const userOrBotProperties = ({ name: _name }) => {
   const name = _name ?? randString();
   const capsName = name.substring(0, 1).toUpperCase() + name.substring(1);
   return deepFreeze({
-    avatar_url: UploadedAvatarURL.validateAndConstructInstance({
-      realm: new URL('https://zulip.example.org'),
-      absoluteOrRelativeUrl: `/yo/avatar-${name}.png`,
-    }),
+    // Internally the UploadedAvatarURL mutates itself for memoization.
+    // That conflicts with the deepFreeze we do for tests; so construct it
+    // here with a full-blown URL object in the first place to prevent that.
+    avatar_url: new UploadedAvatarURL(new URL(`https://zulip.example.org/yo/avatar-${name}.png`)),
 
     date_joined: `2014-04-${randInt(30)
       .toString()
--- a/src/utils/__tests__/url-test.js
+++ b/src/utils/__tests__/url-test.js
@@ -7,9 +7,68 @@ import {
   fixRealmUrl,
   autocompleteRealmPieces,
   autocompleteRealm,
+  isUrlAbsolute,
+  isUrlRelative,
+  isUrlPathAbsolute,
 } from '../url';
 import type { Auth } from '../../types';
 import type { AutocompletionDefaults } from '../url';
+
+const urlClassifierCases = {
+  // These data are mostly a selection from this resource:
+  //   https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json
+  // which is referred to at the top of the URL Standard.
+  absolute: ['https://example.com/foo', 'a1234567890-+.:foo/bar', 'AB://c/d'],
+  pathAbsolute: ['/', '/foo/bar', '/.//path', '/../localhost/', '/:23', '/a/ /c'],
+  otherRelative: [
+    '//example.com/foo',
+    '//foo/bar',
+    '//',
+    '///',
+    '///test',
+    '//www.example2.com',
+    '10.0.0.7:8080/foo.html',
+    'a!@$*=/foo.html',
+    '#β',
+  ],
+};
+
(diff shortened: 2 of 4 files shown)


chrisbobbe:
Great, thanks! This LGTM except I wonder about commenting on the new functions with something like, "why might we use these instead of `new URL`? Because `new URL` is expensive... (etc.)". That way a well-meaning contributor might be less likely to try to sweep them away for #4146.

Shall I do another round of benchmarking for this change, maybe just before and just after it? It could probably be merged before then, anyway; we certainly don't expect it to make anything worse.

chrisbobbe:
> Shall I do another round of benchmarking for this change, maybe take measurements just before and just after it? I'm sure it could be merged before I do so, anyway; we certainly don't expect it to make anything worse.

Ha, well, I went to go and do this, with the timing code that captures both the network time and the `transform` time ([this code](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user_avatar_url_field_optional/near/1080938)). My measurements for before this PR went like this:

image

(I kept going past 12 because of the surprising drop in times.)

So it seems something happened with my network connection quality that threw the measurements off. I'll set it up to just time `transform` in `api.registerForEvents`, then.

chrisbobbe:
> I'll set it up to just time `transform` in `api.registerForEvents`, then.

OK, here we go. With this logging code (and the other logging code undone):

```diff
diff --git src/api/registerForEvents.js src/api/registerForEvents.js
index 655849892..50445f24a 100644
--- src/api/registerForEvents.js
+++ src/api/registerForEvents.js
@@ -5,6 +5,7 @@ import type { ApiNarrow } from './apiTypes';
 import type { CrossRealmBot, User } from './modelTypes';
 import { apiPost } from './apiFetch';
 import { AvatarURL } from '../utils/avatar';
+import * as logging from '../utils/logging';
 
 type RegisterForEventsParams = {|
   apply_markdown?: boolean,
@@ -69,5 +70,9 @@ export default async (auth: Auth, params: RegisterForEventsParams): Promise<Init
     fetch_event_types: JSON.stringify(fetch_event_types),
     client_capabilities: JSON.stringify(client_capabilities),
   });
-  return transform(rawInitialData, auth);
+  const before = Date.now();
+  const transformResult = transform(rawInitialData, auth);
+  const after = Date.now();
+  logging.logToProfiling({ durationMs: after - before, ip: '192.168.1.7' });
+  return transformResult;
 };
```

image

chrisbobbe:
Er, hang on, I'm getting an error:

image

gnprice:
Thanks for the review and the new measurements!

> I wonder about commenting on the new functions with something like, "why might we use these instead of `new URL`? Because `new URL` is expensive... (etc.)". That way a well-meaning contributor might be less likely to try to sweep them away for #4146.

I think it's fine because AFAICT there isn't a direct way to answer these questions using `URL`. Like if I try to write a sentence for the jsdoc like "This is equivalent to [something involving `new URL`], but more efficient.", I don't see how to fill in the blank.

At the one call site, in `avatar.js`, there is a comment explaining that, so anyone trying to remove these should pretty quickly find that.

I'm also coming to think that #4146 may not be entirely the right direction, and we'll want to amend what it says, just because `new URL` turns out to be so slow. (Though there's probably some relatively easy improvements that could be made within its implementation to make it noticeably less so, like maybe 2x.) I guess I'll go write some thoughts on that over on that thread.

gnprice:
Merged, with one bugfix: `Https://example/` is an absolute URL, just like `https://example/`.

I discovered that one while adding more tests, as part of an experimental followup branch working through some of our existing tangly URL-handling functions in pursuit of a different angle on #4146.

chrisbobbe:
> Thanks for the review and the new measurements!
> 
> > I wonder about commenting on the new functions with something like, "why might we use these instead of `new URL`? Because `new URL` is expensive... (etc.)". That way a well-meaning contributor might be less likely to try to sweep them away for #4146.
> 
> I think it's fine because AFAICT there isn't a direct way to answer these questions using `URL`. Like if I try to write a sentence for the jsdoc like "This is equivalent to [something involving `new URL`], but more efficient.", I don't see how to fill in the blank.
> 
> At the one call site, in `avatar.js`, there is a comment explaining that, so anyone trying to remove these should pretty quickly find that.

Ah OK, right—thanks for pointing those things out, and thanks again for the PR. 🙂 

> I'm also coming to think that #4146 may not be entirely the right direction, and we'll want to amend what it says, just because `new URL` turns out to be so slow. (Though there's probably some relatively easy improvements that could be made within its implementation to make it noticeably less so, like maybe 2x.) I guess I'll go write some thoughts on that over on that thread.

Mmm, yeah—thanks.

Issue #4333, opened 2020-12-10
Identify narrows with user IDs and stream IDs, not emails and names


gnprice:
This is a major component of both #3764 and #3918. Currently we describe PM narrows (both 1:1 and group) by emails of the people involved, and stream and topic narrows by the stream name (and in the latter case the topic.) We should instead use the numeric user IDs and stream IDs, for all the reasons described in #3764 and #3918.

This has been a stubborn issue to resolve because there are *so many* places in the code that we create and consume `Narrow` objects, and so many different places the data for them comes from, and all of those need to migrate. I've had several draft branches for doing this, starting with some in 2020-02. The work I'm currently doing on #4035, of which #4332 just now is the latest, will bring it closer by centralizing the fiddly bits in the PM case, and in particular making it so that we always have user IDs on hand at the points where we construct PM narrows.

One symptom of this issue is described at #3230. (Though probably a rare one; found internally, and has no user reports.)

Ayushdubey86:
I would like to work on this!

gnprice:
Hi @Ayushdubey86 ! This issue is a pretty complex one, and I'm already working on it. To find good issues to take on, take a look at our ["help wanted" label](https://github.com/zulip/zulip-mobile/labels/help%20wanted). See also our [contribution guidelines](https://github.com/zulip/zulip-mobile/blob/master/CONTRIBUTING.md), and say hello [in #mobile on chat.zulip.org](https://github.com/zulip/zulip-mobile/#discussion).

timabbott:
Just adding a note that we'll want to open a zulip/zulip issue for doing the cleanup discussed here:
https://chat.zulip.org/#narrow/stream/3-backend/topic/stream.20administrators/near/1124311 (noting we need to wait 6 months before doing it) once we're no longer using the `name` field in subscription/update events.

chrisbobbe:
Thanks, Tim!

> once we're no longer using the `name` field in subscription/update events.

I was mistaken about this; we actually don't use it, so the `name` field can be removed right away, as Greg points out in the [CZO discussion](https://chat.zulip.org/#narrow/stream/3-backend/topic/stream.20administrators/near/1126755).

PR #4346, opened 2020-12-18
narrow: Switch to a reasonable representation, and store user IDs

gnprice:
This follows on #4342 and its predecessors, and accomplishes part of #4333: at the end of this PR, we now have user IDs in our internal representation of PM narrows.

After this, in the next PR or two we'll go on to switch all the various places we consume narrows to use the IDs instead of the emails -- which as mentioned at https://github.com/zulip/zulip-mobile/issues/4035#issuecomment-746896403 will include completing #4035 -- and stop storing emails in our narrows at all.

Thanks to all the prep work in the preceding several PRs, not much actually has to change in this one -- we've already quite thoroughly encapsulated the details of the Narrow type, and also eliminated nearly all the places where we were constructing PM narrows without passing a user ID to the constructor along with an email. Most of what's in this PR is just taking care of the last few of those.

### DIFF
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -1,6 +1,5 @@
 /* @flow strict-local */
 import invariant from 'invariant';
-import isEqual from 'lodash.isequal';
 
 import type { ApiNarrow, Message, Outbox, User, UserOrBot } from '../types';
 import {
@@ -32,9 +31,6 @@ import {
  *    server, and `apiNarrowOfNarrow` for converting to it.
  */
 export opaque type Narrow = ApiNarrow;
-
-export const isSameNarrow = (narrow1: Narrow, narrow2: Narrow): boolean =>
-  Array.isArray(narrow1) && Array.isArray(narrow2) && isEqual(narrow1, narrow2);
 
 export const HOME_NARROW: Narrow = [];
 
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -63,15 +63,6 @@ export const HOME_NARROW_STR: string = keyFromNarrow(HOME_NARROW);
 const pmNarrowFromEmails = (emails: string[]): Narrow => Object.freeze({ type: 'pm', emails });
 
 /**
- * DEPRECATED.  Use `pm1to1NarrowFromUser` instead.
- *
- * This function is being replaced by `pm1to1NarrowFromUser`, as part of
- * migrating from emails to user IDs to identify users.  Don't add new uses
- * of this function; use that one instead.
- */
-export const pmNarrowFromEmail = (email: string): Narrow => pmNarrowFromEmails([email]);
-
-/**
  * A PM narrow, either 1:1 or group.
  *
  * The argument's type guarantees that it comes from
--- a/src/utils/narrow.js
+++ b/src/utils/narrow.js
@@ -33,26 +33,13 @@ import {
 export opaque type Narrow =
   | {| type: 'stream', streamName: string |}
   | {| type: 'topic', streamName: string, topic: string |}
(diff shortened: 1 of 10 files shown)


src/utils/narrow.js
```diff
@@ -282,11 +289,13 @@ export const parseNarrow = (narrowStr: string): Narrow => {
     }
 
     case 'pm:': {
-      if (!rest.startsWith('s:')) {
+      const match = /^d:(.*?):(.*)/.exec(rest);
```
chrisbobbe, line 292: Ah OK, I think I've worked out what we're doing here:

The `?` in the first capture group, `(.*?)`, makes the `*` quantifier lazy instead of greedy, which it would be by default. This is fine because the IDs are numeric and obviously don't contain a `:`. It's also necessary, because email addresses might contain colons (er, I think? [Wikipedia](https://en.wikipedia.org/wiki/Email_address#Domain) gives the example "jsmith@[IPv6:2001:db8::1]", which I think may be permitted by [IETF RFC 2821 section 4.1.3](https://tools.ietf.org/html/rfc2821#section-4.1.3)).

It'll be simpler to think about once we've taken emails out of the picture. If that weren't planned, I'd say we might want to have an explanatory code comment here.

This reminds me of what we're doing for topic narrows, starting in af1713a76 (though it's seeming like that'll need to be reworked because it caused a regression; the `\x00` character we use there (which has a very similar role to the `:` here) doesn't round-trip into being stored as the value of the `data-narrow` attribute in our message-list HTML; see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4339.20regression/near/1085234)).

chrisbobbe (commented):
Thanks, @gnprice! This LGTM pending a few small comments, below.

src/utils/narrow.js
gnprice: > The `?` in the first capture group, `(.*?)`, makes the `*` quantifier lazy instead of greedy, which it would be by default.

Yep!


> It's also necessary, because email addresses might contain colons (er, I think? [Wikipedia](https://en.wikipedia.org/wiki/Email_address#Domain) gives the example "jsmith@[IPv6:2001:db8::1]", which I think may be permitted by [IETF RFC 2821 section 4.1.3](https://tools.ietf.org/html/rfc2821#section-4.1.3)).

Yeah, good find with that example. I didn't even think about whether an email address could contain a colon. But the great thing about using `.*?` here is that we don't have to, because it will match up to that first colon (which has to be the one we put there as a delimiter), and then everything (*) after that is the email addresses.

In general for capturing contents up to a delimiter, `.*?` is what you want rather than `.*`, for just this reason. It similarly appears in the other regexps in this `parseNarrow` function.

I did deliberately put the IDs first and the emails at the end, so that the emails would be the part in that position where we don't have to think about what's there.

(*) Though looking again at the `topic:` case reminds me that there is an asterisk on that, because without the `/s` modifier there are a few characters the `.` pattern rejects. I'm not sure if those characters are possible in email addresses -- I'm sure ASCII newline isn't, but there are a couple of other "newline" characters that wrinkle applies to and I'm not certain if they can appear. I'll add `/s` there too so we don't have to think about it.


> (though it's seeming like that'll need to be reworked because it caused a regression; the `\x00` character we use there (which has a very similar role to the `:` here) doesn't round-trip into being stored as the value of the `data-narrow` attribute in our message-list HTML; see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4339.20regression/near/1085234)).

(We ended up fixing that issue at a different layer, in #4352.)

gnprice:
Thanks for the review! Merged after making the suggested changes.

PR #4350, opened 2020-12-19
isInternalLink: Clarify, fix bugs, add tests, move to use URL object

gnprice:
After studying closely the URL Standard and its definition of the [syntax](https://url.spec.whatwg.org/#url-writing) and [parsing algorithm](https://url.spec.whatwg.org/#url-parsing) for URLs in order to write some small predicates like `isUrlAbsolute` for #4345, I took a fresh look yesterday at some of our URL-processing code that we've been wanting to fix for a while, i.e. #4146.

Here's a branch fixing one tangly part of that code: the `isInternalLink` function, used for deciding whether a link the user tries to follow in a message is one we should try to interpret within the app to navigate to some other conversation, or is one we should instead handle more generically, with a download or going out to the browser.

We make the changes in a series of very small steps, so that we can spot exactly where the behavior changes. After a long string of those, including a couple of bugfixes, we have a version that's pretty reasonable and is still in explicit string/regexp terms... and then in the final commit, we switch it to using `new URL`. The effect of this is that we start accepting a wide range of URLs we didn't before -- for example it turns out that the URL parser produces exactly the same result for `https://example/#narrow` as for
* `https://example/.#narrow`
* `https://example/foo/bar/../../../baz/.././#narrow`
* `https://%65%78%61%6d%70%6c%65/#narrow`
* `https://example:/#narrow`
* `https://example:0000443/#narrow`
* ... and further kinds of variation are possible for domains with non-ASCII characters and for IP addresses instead of domains
* ... and that's just among "valid URL strings", which aren't all the inputs the URL parser accepts.

So in that final commit we also add a wide variety of test cases to illustrate this variation. These are perhaps overkill for this particular function; but as we revise or write other URL-handling functions, they can serve as a helpful source of material for tests.

### DIFF
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -30,6 +30,9 @@ describe('isInternalLink', () => {
     [true, 'path-absolute, with numeric IDs', '/#narrow/stream/123-jest'],
     [true, 'path-absolute, with numeric IDs', '/#narrow/pm-with/123-mark'],
 
+    [false, 'same domain, double slash', 'https://example.com//#narrow/stream/jest'],
+    [false, 'same domain, triple slash', 'https://example.com///#narrow/stream/jest'],
+
     // These examples may seem weird, but a previous version accepted most of them.
     [
       false,
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -16,11 +16,16 @@ const realm = new URL('https://example.com');
 
 describe('isInternalLink', () => {
   const cases = [
-    [false, 'another domain', 'https://another.com'],
-    [false, 'same domain, nontrivial path', 'https://example.com/user_uploads'],
-    [true, 'same domain, to a narrow', 'https://example.com/#narrow/stream/jest'],
     [true, 'fragment-only, to a narrow', '#narrow/stream/jest/topic/topic1'],
+    [false, 'fragment-only, wrong fragment', '#nope'],
     [true, 'path-absolute, to a narrow', '/#narrow/stream/jest'],
+    [false, 'path-absolute, wrong fragment', '/#nope'],
+    [false, 'path-absolute, wrong path', '/user_uploads/#narrow/stream/jest'],
+    [true, 'same domain, to a narrow', 'https://example.com/#narrow/stream/jest'],
+    [false, 'same domain, wrong fragment', 'https://example.com/#nope'],
+    [false, 'same domain, wrong path', 'https://example.com/user_uploads/#narrow/stream/jest'],
+    [false, 'wrong domain', 'https://another.com/#narrow/stream/jest'],
+
     [true, 'fragment-only, with numeric IDs', '#narrow/stream/123-jest/topic/topic1'],
     [true, 'path-absolute, with numeric IDs', '/#narrow/stream/123-jest'],
     [true, 'path-absolute, with numeric IDs', '/#narrow/pm-with/123-mark'],
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -15,28 +15,22 @@ import * as eg from '../../__tests__/lib/exampleData';
 const realm = new URL('https://example.com');
 
 describe('isInternalLink', () => {
-  test('when link is external, return false', () => {
-    expect(isInternalLink('https://example.com', new URL('https://another.com'))).toBe(false);
-  });
+  const cases = [
+    [false, 'another domain', 'https://another.com'],
+    [false, 'same domain, nontrivial path', 'https://example.com/user_uploads'],
+    [true, 'same domain, to a narrow', 'https://example.com/#narrow/stream/jest'],
+    [true, 'fragment-only, to a narrow', '#narrow/stream/jest/topic/topic1'],
+    [true, 'path-absolute, to a narrow', '/#narrow/stream/jest'],
+    [true, 'fragment-only, with numeric IDs', '#narrow/stream/123-jest/topic/topic1'],
+    [true, 'path-absolute, with numeric IDs', '/#narrow/stream/123-jest'],
+    [true, 'path-absolute, with numeric IDs', '/#narrow/pm-with/123-mark'],
+  ];
 
-  test('when link is internal, but not in app, return false', () => {
-    expect(isInternalLink('https://example.com/user_uploads', realm)).toBe(false);
-  });
-
-  test('when link is internal and in app, return true', () => {
-    expect(isInternalLink('https://example.com/#narrow/stream/jest', realm)).toBe(true);
-  });
-
-  test('when link is relative and in app, return true', () => {
-    expect(isInternalLink('#narrow/stream/jest/topic/topic1', realm)).toBe(true);
-    expect(isInternalLink('/#narrow/stream/jest', realm)).toBe(true);
-  });
-
-  test('links including IDs are also recognized', () => {
-    expect(isInternalLink('#narrow/stream/123-jest/topic/topic1', realm)).toBe(true);
-    expect(isInternalLink('/#narrow/stream/123-jest', realm)).toBe(true);
-    expect(isInternalLink('/#narrow/pm-with/123-mark', realm)).toBe(true);
-  });
+  for (const [expected, description, url] of cases) {
+    test(`${expected ? 'accept' : 'reject'} ${description}: ${url}`, () => {
+      expect(isInternalLink(url, realm)).toBe(expected);
+    });
+  }
 });
 
 describe('isMessageLink', () => {
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -30,6 +30,10 @@ describe('isInternalLink', () => {
     [true, 'path-absolute, with numeric IDs', '/#narrow/stream/123-jest'],
     [true, 'path-absolute, with numeric IDs', '/#narrow/pm-with/123-mark'],
 
+    [false, 'fragment-only, #narrowly', '#narrowly/stream/jest'],
+    [false, 'path-absolute, #narrowly', '/#narrowly/stream/jest'],
+    [false, 'same domain, #narrowly', 'https://example.com/#narrowly/stream/jest'],
+
     [false, 'same domain, double slash', 'https://example.com//#narrow/stream/jest'],
     [false, 'same domain, triple slash', 'https://example.com///#narrow/stream/jest'],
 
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/internalLinks-test.js
```diff
@@ -37,6 +37,100 @@ describe('isInternalLink', () => {
...
+      new URL('https://example.भारत/'),
+    ],
+    [
+      true,
+      'same domain, punycodable host',
+      'https://example.भारत/#narrow/stream/jest',
+      new URL('https://example.भारत/'),
+    ],
+    // This one fails because our polyfilled URL implementation has IDNA stripped out.
+    // [true, 'same domain, IDNA-mappable', 'https://ℯⅩªm🄿ₗℰ.ℭᴼⓂ/'], // FAILS
```
chrisbobbe, line 81: Nit: `https://ℯⅩªm🄿ₗℰ.ℭᴼⓂ/#narrow/stream/jest` (e.g.), I think.

Also, I think we're not using the polyfilled implementation in these tests; we don't (e.g.) have a line like `import 'react-native-url-polyfill/auto';` here or in our Jest setup.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -37,6 +37,100 @@ describe('isInternalLink', () => {
...
+    [
+      true,
+      'same IPv4 address, %-encoded',
+      'http://%31%39%32%2e168%2e0%2e1/#narrow/stream/jest',
+      new URL('http://192.168.0.1/'),
+    ],
+    // This one fails because our polyfilled URL implementation has IDNA stripped out.
+    // [true, 'same IPv4 address, IDNA-mappable', 'http://１𝟗𝟚。①⁶🯸．₀｡𝟭/#narrow/stream/jest', new URL('http://192.168.0.1/'),],
```
chrisbobbe, line 89: Same comment as above, about using the polyfill.

This one passes when I `nvm use 13` or `nvm use 14`, but not when I `nvm use 12`; that points to the polyfill not being used here.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -37,6 +37,100 @@ describe('isInternalLink', () => {
     [false, 'same domain, double slash', 'https://example.com//#narrow/stream/jest'],
     [false, 'same domain, triple slash', 'https://example.com///#narrow/stream/jest'],
 
+    // These examples may be odd-looking URLs, but are nevertheless valid.
+    [true, 'scheme-relative', '//example.com/#narrow/stream/jest'],
+    [true, 'path-relative, nop path', '.#narrow/stream/jest'],
```
chrisbobbe, line 42: Interesting, I hadn't heard of "nop" before, but sure enough, I suppose it's at least as commonly used as "no-op": https://en.wikipedia.org/wiki/NOP_(code).

src/utils/__tests__/internalLinks-test.js
```diff
@@ -37,6 +37,100 @@ describe('isInternalLink', () => {
...
+    [true, 'path-relative, nop path', 'foo/..#narrow/stream/jest'],
+    [true, 'path-relative, nop path', 'foo/../#narrow/stream/jest'],
+    [true, 'path-relative, nop path', 'foo/bar/../../../baz/.././#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/.#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/./#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/..#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/../#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/foo/..#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/foo/../#narrow/stream/jest'],
+    [true, 'path-absolute, nop path', '/foo/bar/../../../baz/.././#narrow/stream/jest'],
```
chrisbobbe, line 55: Following all the `../`'s—this would put you up a level _above_ `/`, right, but `new URL` just sets the pathname to `/` instead of throwing an error or something. Odd.

chrisbobbe (commented):
Thanks, @gnprice! It's interesting to see the behavior changes that get revealed when this fix is done step-by-step, like this. It seems like it'll help a lot in debugging if we get bug reports both from people running a version with the fix, and a version without the fix—we'll know exactly how the previous code was broken, and we'll have a head-start in investigating any regressions the fix might have introduced (though I don't expect there to be any).

src/utils/__tests__/internalLinks-test.js
gnprice: Yep -- once the path is at the root, additional `..` path components have no effect. You can see this in a browser: visit e.g.  https://github.com/../../../../zulip/zulip-mobile/pull/4350 . In Chrome for me, the location bar immediately shows the version with the `..` components canonicalized out, before it's even had time to load anything from the server.

In the URL Standard, this is a consequence basically of the "If path is empty" case in the definition of ["shorten a url's path"](https://url.spec.whatwg.org/#shorten-a-urls-path), which is the effect of a "double-dot path segment" [in path state](https://url.spec.whatwg.org/#path-state) in the parser.

It mirrors the behavior of Unix filesystem paths, which URL paths are modeled on / inspired by: for any other directory `..` points to its parent, but for the root directory `/` there is no parent and it points to itself, just like `.`.

src/utils/__tests__/internalLinks-test.js
gnprice: Hmm interesting! That sure seems informative. I'll have to investigate a bit more.

gnprice: Yeah, I see. We get our polyfilled `URL` implementation in the app because of this line by the top of `src/ZulipMobile.js`:
```js
import 'react-native-url-polyfill/auto';
```
But none of our tests import that file, and we don't do anything else to apply the polyfill, so we don't get it in tests.

If I add that polyfill line at the top of the test file:
```
import 'react-native-url-polyfill/auto'
```
then the test fails even with Node 14. So does another one, which in fact had been kind of puzzling before!
```js
    [
      true,
      'same domain, punycoded host',
      'https://example.xn--h2brj9c/#narrow/stream/jest',
      new URL('https://example.भारत/'),
    ],
```
I guess that puzzlingness was a sign telling me to look deeper.

Seems like the right thing to do here is that we should be applying the same polyfill in tests as we do in the app -- otherwise we're at high risk of confusing ourselves anywhere that the polyfill and the implementation in Node have any different behavior. I'll add a commit to that effect.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -15,28 +15,22 @@ import * as eg from '../../__tests__/lib/exampleData';
...
-  });
+  const cases = [
+    [false, 'another domain', 'https://another.com'],
+    [false, 'same domain, nontrivial path', 'https://example.com/user_uploads'],
+    [true, 'same domain, to a narrow', 'https://example.com/#narrow/stream/jest'],
+    [true, 'fragment-only, to a narrow', '#narrow/stream/jest/topic/topic1'],
+    [true, 'path-absolute, to a narrow', '/#narrow/stream/jest'],
+    [true, 'fragment-only, with numeric IDs', '#narrow/stream/123-jest/topic/topic1'],
+    [true, 'path-absolute, with numeric IDs', '/#narrow/stream/123-jest'],
+    [true, 'path-absolute, with numeric IDs', '/#narrow/pm-with/123-mark'],
```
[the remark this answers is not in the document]
gnprice: That's (part of) why I have the URLs included in the descriptions for Jest :slightly_smiling_face: 

I wanted all the descriptions as seen by Jest to be distinct, but in some of these cases I didn't feel like adding more words would be an improvement for clarity. See also the "path-relative, nop path" cases and friends below, in a later commit.

gnprice:
Thanks for the review! New revision pushed.

src/utils/__tests__/internalLinks-test.js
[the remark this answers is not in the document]
chrisbobbe: Ah OK, right, makes sense. 🙂

src/utils/__tests__/internalLinks-test.js
chrisbobbe: > It mirrors the behavior of Unix filesystem paths, which URL paths are modeled on / inspired by: for any other directory `..` points to its parent, but for the root directory `/` there is no parent and it points to itself, just like `.`.

Huh, interesting—I didn't know that.

chrisbobbe:
Great, thanks!

I'm seeing lots of instances of the following, just after the commit that has us using the URL polyfill in Jest tests:

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

      at node_modules/react-native-url-polyfill/index.js:11:69
      at Object.getValue [as URL] (node_modules/react-native/Libraries/Utilities/defineLazyObjectProperty.js:43:16)
          at Array.forEach (<anonymous>)
```

gnprice:
> I'm seeing lots of instances of the following at the commit that has us using the URL polyfill in Jest tests:

Hmm! Indeed I reproduce. Thanks for the catch.

I think before sending the revised version I might have been running only the one test file for `internalLinks` -- which does show the same symptom, but it's at the top above a ton of other output and I apparently didn't notice.

... Ah, aha, I think I might understand. The offending code is here:
```
$ nl -ba node_modules/react-native-url-polyfill/index.js
…
     8	export function setupURLPolyfill() {
     9	  global.REACT_NATIVE_URL_POLYFILL = `${name}@${version}`;
    10	
    11	  polyfillGlobal('URL', () => require('./js/URL').URL);
    12	  polyfillGlobal(
    13	    'URLSearchParams',
    14	    () => require('./js/URLSearchParams').URLSearchParams,
    15	  );
    16	}
```
So lines 11 and 14, the ones Jest is complaining about, are `require` calls.

They're `require` calls inside a getter it's passing to `polyfillGlobal`. Reading the latter function, it's intended to polyfill the given properties with lazy getters -- I think the point is to do exactly what react-native-url-polyfill is doing here, by making the `require` lazy, so that if the polyfill is in something you don't necessarily use or don't immediately use, you don't pay the cost at startup.

But that means that then when some code in Jest tries to use `URL`, or `URLSearchParams`, it goes through this lazy getter, which calls `require`. I think this might even be code in Jest proper, outside any of our test code -- e.g. I've noticed that when it prints test results:
```
 PASS  src/topics/__tests__/topicsReducer-test.js
```
the filenames have dotted underlines and are actually links, in my terminal, to `file:` URLs. Presumably Jest is spitting out some terminal magic to do that, and it'd be very sensible for it to use `URL` to construct the URLs to put there.

In which case, it tries to get `URL`, it goes through that getter -- and there it is, trying to import a file after the Jest environment has been torn down.

I'll change it to skip the laziness and do the import up front as an `import` statement.

gnprice:
New revision pushed with that fix!

chrisbobbe:
Great, thanks! Pushed.

PR #4352, opened 2020-12-22
msglist: Fix regression on tapping topic narrows in stream-narrow msg…

chrisbobbe:
…list.

In af1713a76, from #4339, a subtle bug was introduced by our use of
the `\x00` character in making keys for narrows (`keyFromNarrow`).
That character doesn't round-trip through being the value of the
`data-narrow` attribute in our message-list HTML, so we got an error
on pressing a topic-narrow header in the message list for a stream
narrow.

So, make it so that that attribute never contains `\x00` (or indeed
any other non-handled character [1]) by encoding it with our
`base64UtfEncode` function.

See discussion for how we arrived at this solution [2]. In
particular, it has the nice quality of also reducing our exposure to
bugs stemming from stream names (or topics, or users' email
addresses -- anything that ends up in the result of `keyFromNarrow`)
containing characters that aren't allowed in HTML-attribute values.

[1] https://html.spec.whatwg.org/multipage/syntax.html#attributes-2
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4339.20regression/near/1085234

### DIFF
--- a/src/utils/encoding.js
+++ b/src/utils/encoding.js
@@ -37,13 +37,10 @@ export const hexToBase64 = (hex: string) => base64.encode(hexToAscii(hex));
  *
  * This lets us pass an arbitrary string through a channel (like the
  * `postMessage` on RN's WebViews on Android) that tries to do something
- * like percent-decode it.
+ * like percent-decode it, or (like an HTML attribute) that forbids certain
+ * characters.
  *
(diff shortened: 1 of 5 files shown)


gnprice:
Thanks, and thanks for spotting this bug! Merged, with a new commit on top tweaking the comments.

chrisbobbe:
Great, thanks!

PR #4355, opened 2020-12-22
accessibility: Added captions for visually impaired users

### LINKED ISSUE #4232: Button Captions for Screen Readers (Accessibility)
My organization is currently rolling out this platform, and it appears to be fulfilling all of our needs accept for one small issue.

For the visually impaired mobile users (specifically ios) , there appears to be a lack of pronounceable widgets in the user interface. It has been pointed out to me that the most likely buttons without this "caption" are the buttons to send a message and the button to back out of a conversation. There are other buttons without labels, but these are obviously the important ones. 

The fix for this was described to be similar to adding alt-text to images in HTML.

Would it be possible to add better usability for the visually impaired?

Gautam-Arora24:
**Why and how the change was made**
This PR proposed to add captions for visually impaired users. I added captions for buttons like back button and send button. I used React Native's accessibilityLabel prop to add captions. 

**Tests done**
I tested this app on both simulator and on the actual device, and it worked well for both the scenarios.

**Fixes #4232**

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -506,6 +506,7 @@ class ComposeBox extends PureComponent<Props, State> {
             />
           </View>
           <FloatingActionButton
+            accessibilityLabel="Send message"
             style={this.styles.composeSendButton}
             Icon={editMessage === null ? IconSend : IconDone}
             size={32}
--- a/src/nav/NavButtonGeneral.js
+++ b/src/nav/NavButtonGeneral.js
@@ -9,6 +9,7 @@ import { Touchable } from '../common';
 type Props = $ReadOnly<{|
   children: React$Node,
   onPress: () => void,
+  accessibilityLabel?: string,
 |}>;
 
 export default class NavButtonGeneral extends PureComponent<Props> {
@@ -22,10 +23,10 @@ export default class NavButtonGeneral extends PureComponent<Props> {
   });
 
   render() {
-    const { children, onPress } = this.props;
+    const { children, onPress, accessibilityLabel } = this.props;
 
     return (
-      <Touchable onPress={onPress}>
+      <Touchable onPress={onPress} accessibilityLabel={accessibilityLabel}>
         <View style={this.styles.navButtonFrame}>{children}</View>
       </Touchable>
     );
--- a/src/common/FloatingActionButton.js
+++ b/src/common/FloatingActionButton.js
@@ -22,6 +22,7 @@ type Props = $ReadOnly<{|
   size: number,
   Icon: SpecificIconType,
   onPress: () => void,
+  accessibilityLabel?: string,
 |}>;
 
 /**
@@ -38,7 +39,7 @@ type Props = $ReadOnly<{|
  */
 export default class FloatingActionButton extends PureComponent<Props> {
   render() {
-    const { style, size, disabled, onPress, Icon } = this.props;
+    const { style, size, disabled, onPress, Icon, accessibilityLabel } = this.props;
     const iconSize = Math.trunc(size / 2);
     const customWrapperStyle = {
       width: size,
@@ -51,7 +52,11 @@ export default class FloatingActionButton extends PureComponent<Props> {
     };
 
     return (
-      <Touchable style={style} onPress={disabled ? undefined : onPress}>
+      <Touchable
+        style={style}
+        onPress={disabled ? undefined : onPress}
+        accessibilityLabel={accessibilityLabel}
+      >
         <View style={[styles.wrapper, customWrapperStyle]}>
           <Icon style={iconStyle} size={iconSize} color="white" />
         </View>
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -62,6 +62,7 @@ class ChatNavBar extends PureComponent<Props> {
         >
           <NavButton
             name="arrow-left"
+            accessibilityLabel="Navigate up"
             color={color}
             onPress={() => {
               NavigationService.dispatch(navigateBack());
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -4,16 +4,14 @@ import React, { PureComponent } from 'react';
 import { View } from 'react-native';
 import Color from 'color';
 
-import * as NavigationService from './NavigationService';
 import type { Dispatch, Narrow, EditMessage } from '../types';
 import { LoadingBanner } from '../common';
 import { connect } from '../react-redux';
 import { BRAND_COLOR, NAVBAR_SIZE } from '../styles';
 import Title from '../title/Title';
-import NavButton from './NavButton';
+import NavBarBackButton from './NavBarBackButton';
 import { DEFAULT_TITLE_BACKGROUND_COLOR, getTitleBackgroundColor } from '../title/titleSelectors';
 import { foregroundColorFromBackground } from '../utils/color';
-import { navigateBack } from '../actions';
 import { ExtraButton, InfoButton } from '../title-buttons/titleButtonFromNarrow';
 
 type SelectorProps = {|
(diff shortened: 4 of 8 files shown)


src/nav/NavButtonGeneral.js
```diff
@@ -26,7 +26,14 @@ export default class NavButtonGeneral extends PureComponent<Props> {
 
     return (
       <Touchable onPress={onPress}>
-        <View style={this.styles.navButtonFrame}>{children}</View>
+        <View
+          accessible={true}
+          accessibilityLabel="Go Back!"
```
gnprice, line 31: This is much too broad, because this component is used in a number of other places in the app.

For a clean approach to apply this change in only the relevant spots, see these commits in the earlier PR:
https://github.com/zulip/zulip-mobile/pull/2753/commits/6935e5bfb54b1279805228f32737c8d51b5e5686
https://github.com/zulip/zulip-mobile/pull/2753/commits/d26962d0116927e4bdeb8e3d810ec1a4645b6bb6
https://github.com/zulip/zulip-mobile/pull/2753/commits/a0d74d9b8d92d2bab226d45b598267bba8c4d05d

src/compose/ComposeBox.js
```diff
@@ -512,13 +512,19 @@ class ComposeBox extends PureComponent<Props, State> {
...
-            style={this.styles.composeSendButton}
-            Icon={editMessage === null ? IconSend : IconDone}
-            size={32}
-            disabled={message.trim().length === 0}
-            onPress={editMessage === null ? this.handleSend : this.handleEdit}
-          />
+          <View
+            accessible={true}
+            accessibilityLabel="Send Message"
+            accessibilityHint="Sends a message !"
```
gnprice, line 518: No need for a hint when the label already conveys all the same information.

gnprice (commented):
Thanks @Gautam-Arora24 for these changes!

The most important piece of feedback I have is that these changes need to be manually tested: try navigating the app by audio, and find out what effect your changes have. Here's the upstream guides for doing so with either Android or iOS:
https://developer.android.com/guide/topics/ui/accessibility/testing
https://developer.apple.com/documentation/uikit/accessibility_for_ios_and_tvos/supporting_voiceover_in_your_app
Either one would be enough for a PR. (Later, before we really consider the accessibility support complete, we'll want to try it out on both because there may be relevant differences.)

gnprice:
Also be sure to take a look at the Zulip project's guidelines for commit messages:
https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages
as well as the other information in the mobile-specific style guide:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md

Gautam-Arora24:
@gnprice Can you review this PR?

gnprice:
Thanks @Gautam-Arora24 ! This is looking like a promising structure.

Let's split this up into several commits. That will help make it easier to analyze the different changes that are happening. You can read more about our approach to small, coherent commits here:
https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-discipline

Specifically, for this PR, let's do:
* One commit for `FloatingActionButton`, `NavButtonGeneral`, and `NavButton`, which just add the `accessibilityLabel` prop to be passed through. (This could even be two commits, one for `FloatingActionButton` and one for the others.)
* One commit for `ComposeBox`, adding the label on the send-message button.
* One commit for `ChatNavBar`, `ModalNavBar`, and `ModalSearchNavBar`, adding the label on the "navigate back/up" button.

Then for the commits that actually add specific labels:
* Please mention in the commit message what testing you did to find the most natural wording to use for the given action.
  * I've never been a regular user of this kind of UI (and I believe you haven't either), so I don't feel like I automatically know what fits in naturally with the UX that someone will get when they're relying on it. It's best to choose something that'll be similar to what the user gets from other apps.
  * For the back/up button, a good model would be the system settings app, or other Google-written apps that have plenty of that kind of navigation. From when I tried this out when first reviewing this PR, I think the wording there was like "Navigate up" -- note "up", not "back". For background on what that's about, see this article: https://developer.android.com/guide/navigation/navigation-principles A user won't have read that article, but they will know that the same wording they hear in other apps means the thing they're looking for.
    * (This should probably be different between iOS and Android. I'm OK with merging a PR like this with just the Android wording for now, and we can handle the iOS wording in a followup.)
  * For the "send message" button, a good model would be the system Messages (i.e. SMS and the like) app, and also FB Messenger, Slack, Discord, WhatsApp, Signal, and so on. No need to try every one of them if you don't have some of them conveniently lying around, but at least two or three would be good.

I'm not sure how much experience you have already with Git, but this is one of those things that might look like a pain but becomes very quick to do once you know the right commands. Here, I'd do it with commands like these:
```
$ git reset HEAD^
$ git status # look at what changes you have
$ git commit file1 file2
$ git status # look again
$ git commit file3
$ # … repeat until all changes are committed
```
There's a bunch of related useful information in Zulip's [Git guide](https://zulip.readthedocs.io/en/latest/git/using.html#examine-and-tidy-your-commit-history), and you can also always ask questions on chat.zulip.org in `#development help`, `#general`, or `#mobile`.

gnprice:
Oh, one other thought: ideally we should really have these strings translated. :slightly_smiling_face: 

I'll be fine with merging this PR with them just always in English, though -- and then we can work through wiring them up for translation as a followup.

Gautam-Arora24:
This PR is now splitted into several commits as per the requirements!

gnprice:
Thanks @Gautam-Arora24 ! Please also do this part:

> Then for the commits that actually add specific labels:
> 
> * Please mention in the commit message what testing you did to find the most natural wording to use for the given action.
>   
>   * I've never been a regular user of this kind of UI (and I believe you haven't either), so I don't feel like I automatically know what fits in naturally with the UX that someone will get when they're relying on it. It's best to choose something that'll be similar to what the user gets from other apps.
>   * For the back/up button, a good model would be the system settings app, or other Google-written apps that have plenty of that kind of navigation. From when I tried this out when first reviewing this PR, I think the wording there was like "Navigate up" -- note "up", not "back". For background on what that's about, see this article: https://developer.android.com/guide/navigation/navigation-principles A user won't have read that article, but they will know that the same wording they hear in other apps means the thing they're looking for.
>     
>     * (This should probably be different between iOS and Android. I'm OK with merging a PR like this with just the Android wording for now, and we can handle the iOS wording in a followup.)
>   * For the "send message" button, a good model would be the system Messages (i.e. SMS and the like) app, and also FB Messenger, Slack, Discord, WhatsApp, Signal, and so on. No need to try every one of them if you don't have some of them conveniently lying around, but at least two or three would be good.

---

Separately, the last commit on top looks like a fix to a previous commit, which should be squashed into that one. See the discussion of coherent commits in the Zulip project's Git style guide:

You can make both of these changes conveniently -- squashing that commit into a previous one, and editing the commit messages of other commits -- with `git rebase -i`. See the quick guide to fixing commits here:
https://zulip.readthedocs.io/en/latest/git/fixing-commits.html
and Git questions are also always welcome on chat.zulip.org.

gnprice:
Thanks @Gautam-Arora24 for the revision! Looks good -- merged.

I made some small style edits to the commit messages before merge. Take a look:

commit 4775de532a4624a59f684a0fbf0b29287c7469eb

    a11y: Add accessibilityLabel prop on FloatingActionButton.

commit 311ffaff4cd20f039fae6e1d58cf78caec6f481e

    a11y: Add accessibilityLabel prop on NavButton and NavButtonGeneral.

commit 0c66376b54ac3ca336258d8d9f29f38cbd8d3293

    a11y: Add label on "send message" button.
    
    Through various tests on different messaging apps like WhatsApp,
    Messenger and Discord etc., I found that the label was either
    "Send message" or "Send button". So I used "Send message".

commit 7a17dd615ed3bb7c41d83fc176040049a54f5fea

    a11y: Add "Navigate up" label on up/back button.
    
    Through tests on system settings app and system messages app on
    Android, a user hears Navigate up when they click on back button.
    So I used this as a label on back button.
    
    For iOS the label should probably be something else, but this is
    better than not having it.

The `a11y:` prefix refers to a common abbreviation for "accessibility". We prefix our commit summary lines with the general area that a commit is about, and don't use the "feat:" system that some other projects do. I also wrapped the text to a width of 68 columns.

I highly recommend browsing through our Git history -- it's good for seeing examples of our commit style, among many other things. See our tips for reading Git history here: https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md

I also added a commit on top, just making a small refactor that this branch caused me to notice we should do.

PR #4356, opened 2020-12-23
recipient: Finish reconciling the ways we identify a PM conversation

gnprice:
This is the next PR, following #4346, in our series straightening out how we identify narrows and especially PM narrows. It takes some further steps toward #4333 and #3764, converting some narrow-consumers to use user IDs rather than emails; and it completes the last remaining pieces of #4035.

Along the way, we fix some bugs that potentially hit if a user changes their email address, so that we potentially have data reflecting both their old and new email addresses. This will probably be a recurring feature of many of the individual conversions to using user IDs that make up #3764.

Fixes: #4035

### DIFF
--- a/src/unread/unreadSelectors.js
+++ b/src/unread/unreadSelectors.js
@@ -11,11 +11,11 @@ import {
   getUnreadHuddles,
   getUnreadMentions,
 } from '../directSelectors';
-import { getOwnEmail, getAllUsersByEmail } from '../users/userSelectors';
+import { getOwnUserId } from '../users/userSelectors';
 import { getSubscriptionsById } from '../subscriptions/subscriptionSelectors';
 import { isTopicMuted } from '../utils/message';
 import { caseNarrow } from '../utils/narrow';
-import { NULL_SUBSCRIPTION, NULL_USER } from '../nullObjects';
+import { NULL_SUBSCRIPTION } from '../nullObjects';
 
 /** The number of unreads in each stream, excluding muted topics, by stream ID. */
 export const getUnreadByStream: Selector<{ [number]: number }> = createSelector(
@@ -221,24 +221,13 @@ export const getUnreadByHuddlesMentionsAndPMs: Selector<number> = createSelector
 export const getUnreadCountForNarrow: Selector<number, Narrow> = createSelector(
   (state, narrow) => narrow,
   state => getStreams(state),
-  state => getAllUsersByEmail(state),
-  state => getOwnEmail(state),
+  state => getOwnUserId(state),
   state => getUnreadTotal(state),
   state => getUnreadStreams(state),
   state => getUnreadHuddles(state),
   state => getUnreadPms(state),
   state => getMute(state),
-  (
-    narrow,
-    streams,
-    allUsersByEmail,
-    ownEmail,
-    unreadTotal,
-    unreadStreams,
-    unreadHuddles,
-    unreadPms,
-    mute,
-  ) => {
+  (narrow, streams, ownUserId, unreadTotal, unreadStreams, unreadHuddles, unreadPms, mute) => {
     const sumLengths = unreads => unreads.reduce((sum, x) => sum + x.unread_message_ids.length, 0);
 
     return caseNarrow(narrow, {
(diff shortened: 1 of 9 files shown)


src/utils/recipient.js
```diff
@@ -273,6 +275,29 @@ export const pmUnreadsKeyFromMessage = (message: Message, ownUserId?: number): s
...
+ */
+// See comment on pmUnreadsKeyFromMessage for details on this form.
+export const pmUnreadsKeyFromPmKeyIds = (
+  userIds: $ReadOnlyArray<number>,
+  ownUserId: number,
+): string => {
+  if (userIds.length === 1) {
+    // A 1:1 PM.  Both forms include just one user: the other user if any,
+    //   and self for a self-1:1.
+    return userIds[0].toString();
```
chrisbobbe, line 293: Are you thinking we'll eventually combine `state.unread.pms` and `state.unread.huddles` into one bucket with a shared key space, as part of dissolving unhelpful distinctions between group and 1:1 PMs?

src/unread/unreadSelectors.js
```diff
@@ -257,9 +258,8 @@ export const getUnreadCountForNarrow: Selector<number, Narrow> = createSelector(
...
       pm: (emails, ids) => {
         if (ids.length > 1) {
-          // TODO this should go somewhere central like recipient.js
-          const userIds = [...ids, ownUserId].sort((a, b) => a - b).join(',');
-          const unread = unreadHuddles.find(x => x.user_ids_string === userIds);
+          const unreadsKey = pmUnreadsKeyFromPmKeyIds(ids, ownUserId);
+          const unread = unreadHuddles.find(x => x.user_ids_string === unreadsKey);
           return unread ? unread.unread_message_ids.length : 0;
         } else {
           const senderId = ids[0];
```
chrisbobbe, line 265: Could call `pmUnreadsKeyFromPmKeyIds` to get `senderId` here, I suppose—as long as `senderId` and `unreadPms[].sender_id` are made to agree in being a `number` or a `string` (for the `.find` that they're used in here).

chrisbobbe (commented):
Looks great, thanks! Just a few small comments/queries, below.

src/utils/recipient.js
gnprice: Yeah, I think that'd make sense. There's this comment already in `pmConversationsSelectors.js`:
```js
      unread:
        // This business of looking in one place and then the other is kind
        // of messy.  Fortunately it always works, because the key spaces
        // are disjoint: all `unreadHuddles` keys contain a comma, and all
        // `unreadPms` keys don't.
        /* $FlowFixMe: The keys of unreadPms are logically numbers, but because it's an object they
         end up converted to strings, so this access with string keys works.  We should probably use
         a Map for this and similar maps. */
        unreadPms[recipient.unreadsKey] || unreadHuddles[recipient.unreadsKey],
```

That one is looking in a couple of cached-selector results, not `state.unread.pms` and `state.unread.huddles` themselves, but the selectors correspond directly to those.

src/unread/unreadSelectors.js
gnprice: Yeah. I think that will most naturally come after converting these to our own `Immutable.Map` or similar data structure, and perhaps after converting the both of these (`state.unread.pms` and `state.unread.huddles`) into a single such data structure.

With the data structures as they are, with `.sender_id` a number, if we ran `ids` through `pmUnreadsKeyFromPmKeyIds` we'd have to convert it back from string to number in order to do the search (or else convert every `sender_id` into a string to do the comparison, which would be inefficient.) That conversion would make sense only because we know the key isn't just any string but has a particular structure... but the same facts about its structure mean we can just use the number directly and skip converting to a string and back.

chrisbobbe: > Yeah. I think that will most naturally come after converting these to our own `Immutable.Map` or similar data structure, and perhaps after converting the both of these (`state.unread.pms` and `state.unread.huddles`) into a single such data structure.

Makes sense—I figured that was in the plan. 🙂

gnprice:

PR #4358, opened 2020-12-25
Refresh the still-useful parts of #4116, on path to using recent-PMs data

gnprice:
After taking care of #4035 (cleanly handling PM recipients, in place of our complex old handling of them) with my series of PRs up to #4356, I'm taking a look back at #3133/#3535, to start using the recent_private_conversations data structure, which had been blocked by that work.

Ray had a PR #4116 that sought to do some of that. Some of it goes in some of the same directions as I've since done with my recent PRs for #4035, and there are some parts that I don't think are helpful or needed. But there are other changes there that were and still are useful. Here I've rebased those changes, and added some related commits of my own.

After this, I'll return to Ray's further PR #4104 (which builds on the changes in #4116) to see what parts of that are also helpful to use now.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -101,8 +101,13 @@ export const diverseCharacters =
  * Users and bots
  */
 
+type UserOrBotPropertiesArgs = {|
+  name?: string,
+  user_id?: number,
+|};
+
 const randUserId: () => number = makeUniqueRandInt('user IDs', 10000);
-const userOrBotProperties = ({ name: _name }) => {
+const userOrBotProperties = ({ name: _name, user_id }: UserOrBotPropertiesArgs) => {
   const name = _name ?? randString();
   const capsName = name.substring(0, 1).toUpperCase() + name.substring(1);
   return deepFreeze({
@@ -119,12 +124,12 @@ const userOrBotProperties = ({ name: _name }) => {
     full_name: `${capsName} User`,
     is_admin: false,
     timezone: 'UTC',
-    user_id: randUserId(),
+    user_id: user_id ?? randUserId(),
   });
 };
 
 /** Beware! These values may not be representative. */
-export const makeUser = (args: { name?: string } = {}): User =>
+export const makeUser = (args: UserOrBotPropertiesArgs = Object.freeze({})): User =>
   deepFreeze({
     ...userOrBotProperties(args),
 
@@ -138,7 +143,9 @@ export const makeUser = (args: { name?: string } = {}): User =>
   });
 
 /** Beware! These values may not be representative. */
-export const makeCrossRealmBot = (args: { name?: string } = {}): CrossRealmBot =>
+export const makeCrossRealmBot = (
+  args: UserOrBotPropertiesArgs = Object.freeze({}),
+): CrossRealmBot =>
   deepFreeze({
     ...userOrBotProperties(args),
     is_bot: true,
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -346,6 +346,9 @@ export const pmMessage = (args?: {|
   return deepFreeze({ ...baseMessage, ...extra });
 };
 
+export const pmMessageFromTo = (from: User, to: User[], extra?: $Rest<Message, {}>): Message =>
+  pmMessage({ sender: from, recipients: [from, ...to], ...extra });
+
 const messagePropertiesFromStream = (stream1: Stream) => {
   const { stream_id, name: display_recipient } = stream1;
   return deepFreeze({
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -332,6 +332,8 @@ export const pmMessage = (args?: {|
 
     content: 'This is an example PM message.',
     content_type: 'text/markdown',
+    // We don't sort the recipients, because they're inconsistently sorted
+    // in real messages.  (See comments on the Message type.)
     display_recipient: recipients.map(displayRecipientFromUser),
     id: randMessageId(),
     recipient_id: 2342,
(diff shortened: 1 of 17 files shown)


src/chat/__tests__/narrowsSelectors-test.js
```diff
@@ -113,11 +113,11 @@ describe('getFirstMessageId', () => {
       narrows: Immutable.Map({
         [HOME_NARROW_STR]: [1, 2, 3],
       }),
-      messages: {
```
chrisbobbe, line 116:
```
tests: Use new eg.makeMessagesState where applicable.

Note that this actually changes the test data in a couple of spots --
where we were producing a malformed data structure!  Looks like a
copy-paste error... in which it's probably not a coincidence that the
old form here was pretty repetitive.
```

Does it change the test data? I'm not sure I follow; it looks like a pure refactor to me.

src/types.js
```diff
@@ -320,9 +320,22 @@ export type TabNavigationOptionsPropsType = {|
  * Summary of a PM conversation (either 1:1 or group PMs).
  */
 export type PmConversationData = {|
+  /**
+   * A comma-separated (numerically-)sorted sequence of the IDs of the users
+   * involved in this conversation. Does not includes the self-user iff there
```
chrisbobbe, line 325: nit: `Does not includes` -> (maybe) `Excludes`?

chrisbobbe (commented):
Thanks! This LGTM except for a few small comments/queries below.

src/types.js
gnprice: Thanks, yeah, the negative on the left of this "iff" is confusing.

src/chat/__tests__/narrowsSelectors-test.js
gnprice: Ah, it's subtle -- there's this (twice):
```diff
-      messages: {
-        ...eg.baseReduxState.messages,
-        [message1.id]: message1,
-        [message2.id]: message1,
-      },
+      messages: eg.makeMessagesState([message1, message2]),
```

Note the `message1` at `message2.id`.

I'll make the commit message clearer about that.

src/__tests__/lib/exampleData.js
```diff
@@ -119,12 +126,12 @@ const userOrBotProperties = ({ name: _name }) => {
...
     is_admin: false,
     timezone: 'UTC',
-    user_id: randUserId(),
+    user_id: user_id ?? randUserId(),
   });
 };
 
 /** Beware! These values may not be representative. */
-export const makeUser = (args: { name?: string } = {}): User =>
+export const makeUser = (args: UserOrBotPropertiesArgs = NULL_OBJECT): User =>
```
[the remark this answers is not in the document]
gnprice: Good question :stuck_out_tongue: https://github.com/zulip/zulip-mobile/pull/4104#discussion_r424080618

I guess it'd be good to mention that in the commit message.

src/chat/__tests__/narrowsSelectors-test.js
chrisbobbe: Oh, yikes, yeah—that `message1`/`message2` misalignment sure looks fishy; thanks for fixing. I actually noticed this when working on my draft (not yet pushed) for storing `state.messages` as an `Immutable.Map` and made a note to return to it in case it was somehow done on purpose.

src/__tests__/lib/exampleData.js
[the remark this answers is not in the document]
gnprice: ... Or I'll just do the other workaround discussed there, namely to say `Object.freeze({})` directly. That's what `NULL_OBJECT` is defined as anyway, and it perhaps needs less explanation.

src/pm-conversations/__tests__/pmConversationsSelectors-test.js
```diff
@@ -29,45 +29,33 @@ describe('getRecentConversations', () => {
...
       users: [eg.selfUser, userJohn, userMark],
       narrows: Immutable.Map({
-        [ALL_PRIVATE_NARROW_STR]: [
-          meJohnAndMarkPm.id,
-          meAndJohnPm1.id,
-          meAndMarkPm.id,
-          meAndJohnPm2.id,
-          meOnlyPm.id,
-        ],
+        [ALL_PRIVATE_NARROW_STR]: [0, 1, 2, 3, 4],
```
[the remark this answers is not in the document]
gnprice: There is a tradeoff here, but I think on net this version is a bit easier to understand.

One key condition of that is that these IDs all stay local -- their meaning is all contained within the test case, and doesn't involve looking out to anything further afield. (I'd like to say the scope of code that can care about one of these IDs all fits on a screen, but that may not be true on laptop screens. It'd be better if each test's code were shorter, so that it did.)

Then on the other hand these names were feeling like a bit of work to decode in themselves. Partly because they're long, perhaps? They're sensible names, and I don't have better ideas for names for these values, so that seems potentially inherent to naming them in these tests.

... In fact. It looks like the data don't actually match up with each other!
```js
          {
            sender_id: userJohn.user_id,
            unread_message_ids: [meAndMarkPm1.id, meAndMarkPm2.id],
          },
          {
            sender_id: userMark.user_id,
            unread_message_ids: [meAndJohnPm1.id],
          },
```
The fact that we didn't notice that until now suggests to me that the names weren't providing a lot of value in making this test data easy to understand. I'll add a fix for that to this branch.

Another thing on the other hand is that the ordering of the message IDs is important -- the output is supposed to be sorted by message ID (descending), and the second test is all about that property. So with the names you have to go consulting the definitions and matching them up to see that the ordering is right, whereas with the IDs it's immediate.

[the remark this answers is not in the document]
chrisbobbe: Ah OK, makes sense. And thanks for catching that mismatch!

gnprice:
Thanks for the review! Pushed a new revision.

chrisbobbe:
Great, thanks! Merged.

Issue #3557, opened 2019-07-14
iOS: scrolling in views with messages is slow


antifuchs:
Since the iOS app update prior to last, scrolling in views where there are messages (inside a topic or stream) feels much more "heavy" than previously: I used to be able to "flick" the content up and scroll a significant chunk of channel contents (like you would on a Safari page). Now, it takes me maybe a screenful or two forward/backward.

This isn't so annoying when catching up on a topic, but it gets really difficult when I try to find an older message in a long-running stream (my thumbs are very nasty RSI hot spots).

chrisbobbe:
We've had another report of this, [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Deprecate.20mobile.20app.3F/near/852292).

gnprice:
I reproduce this on both an iPhone XR running iOS 13 and an iPhone 7 running iOS 12.

I *don't* reproduce on Android -- scrolling with a "flick" gesture works great on a Pixel 4 running Android 10. So something's happening here that's specific to iOS.

PR #4360, opened 2020-12-28
webview: Use native scroll deceleration on iOS

titanous:
The default WebView `decelerationRate` is `fast`, which slows down momentum scrolling through messages significantly. This changes it to `normal` which is the native rate on iOS and makes it behave like other apps, allowing for much faster scrolling.

Test Plan: confirmed that this has the expected scroll momentum on an iPhone 12.

### DIFF
--- a/src/webview/MessageList.js
+++ b/src/webview/MessageList.js
@@ -307,6 +307,7 @@ class MessageList extends Component<Props> {
         source={{ baseUrl: (baseUrl.toString(): string), html }}
         originWhitelist={['file://']}
         onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
+        decelerationRate="normal"
         style={{ backgroundColor: 'transparent' }}
         ref={webview => {
           this.webview = webview;


chrisbobbe:
Good find, thanks for the PR! This fixes #3557, a long-standing issue. 🎉 Merged, with a few tweaks:

- I checked out the branch and ran `tools/test` (this has stopped running automatically for PRs and posting the results on GitHub; that's #4174), and ESLint told me it would like double quotes there instead of singled quotes; fixed.
- Added `Fixes: #3557` to the commit message, and the post at the top of this thread, so that issue gets linked up and closed automatically. 🎉 
- Fixed a commit-message style nit by changing "This changes it" to "Change it"; one obscure difference (we should probably document it somewhere) between the mobile-app project and the main Zulip commit-message style guide ([here](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages)) is that we explicitly prefer the imperative in the commit message body, not just the summary line, for describing what work is done in the commit. The https://github.com/zulip/zulip project itself is inconsistent on this point; Greg discusses that at https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622188734.

PR #4361, opened 2020-12-29
narrow: Convert more email usage to user IDs

gnprice:
This is the next tranche of changes in the series of PRs up to #4356. It converts several more of the places we use the emails from a PM narrow to use user IDs instead. Along the way we make some other conversions from user IDs to emails, as well as other cleanups in the code we're operating on.

### DIFF
--- a/src/common/UserAvatarWithPresence.js
+++ b/src/common/UserAvatarWithPresence.js
@@ -19,7 +19,6 @@ type Props = $ReadOnly<{|
   avatarUrl: AvatarURL,
   email: string,
   size: number,
-  shape: 'rounded' | 'square',
   onPress?: () => void,
 |}>;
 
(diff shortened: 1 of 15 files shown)


src/common/UserAvatarWithPresence.js
```diff
@@ -41,3 +51,33 @@ export default class UserAvatarWithPresence extends PureComponent<Props> {
...
+    const user = getUserForId(state, props.userId);
+    return { avatarUrl: user.avatar_url, email: user.email };
+  },
+)(
+  // The type inference embedded in our `connect` type relies on seeing the
+  // exact Props type for the underlying component, complete with all the
+  // expected props.  Normally that's just how our Props types are in the
+  // first place, but here we have to provide it explicitly.
+  /* eslint-disable flowtype/generic-spacing */
+  (UserAvatarWithPresence: ComponentType<
```
chrisbobbe, line 80: I think using react-redux's `connect` here is fine and convenient because we want to create a separate thing with a new interface, instead of modifying the original thing.

But when we settle on using this one and removing the old one, I wonder about trying out React Hooks, in particular, react-redux's `useReducer` (which is already available with the react-redux version we're on). True, we've gotten a pretty good handle on treating the "inner" and "outer" props in most places, with `SelectorProps`—but we won't have to do even that, when there's no wrapping component involved. 🙂

chrisbobbe (commented):
Thanks! See just a few small comments below.

src/title/TitlePrivate.js
```diff
@@ -18,7 +18,7 @@ type SelectorProps = $ReadOnly<{|
 |}>;
 
 type Props = $ReadOnly<{
-  email: string,
+  userId: number,
```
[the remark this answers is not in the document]
chrisbobbe: It looks like we could also conveniently use `UserAvatarWithPresenceById` in `UserItem` without too much fuss, too?

src/common/UserAvatarWithPresence.js
gnprice: Sure, that'd be interesting!

(`useSelector`, I think.)

chrisbobbe: > (`useSelector`, I think.)

Ah of course 🤪.

src/title/TitlePrivate.js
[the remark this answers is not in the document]
gnprice: Sure, added those changes.

gnprice:
New revision pushed!

chrisbobbe:
Great, thanks! Merged.

PR #4362, opened 2020-12-29
replaceRevive: Use switch/case for what's interesting in `replacer`.

chrisbobbe:
As Greg suggests, at
https://github.com/zulip/zulip-mobile/pull/4348#discussion_r548764896.

Also add some comments / jsdoc.

### DIFF
--- a/src/boot/replaceRevive.js
+++ b/src/boot/replaceRevive.js
@@ -30,28 +30,29 @@ const replacer = function replacer(key, value) {
   // The value at the current path before JSON.stringify called its
   // `toJSON` method, if present.
   //
-  // When identifying what kind of thing we're working with, be sure
-  // to examine `origValue` instead of `value`, if calling `toJSON` on
-  // that kind of thing would remove its identifying features -- which
-  // is to say, if that kind of thing has a `toJSON` method.
+  // When identifying what kind of thing we're working with, it's good
+  // practice to examine `origValue` instead of `value`, just in case
+  // calling `toJSON` on that kind of thing would remove its
+  // identifying features -- which is to say, just in case that kind
+  // of thing has a `toJSON` method.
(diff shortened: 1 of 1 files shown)


src/boot/replaceRevive.js
```diff
@@ -46,24 +46,31 @@ const replacer = function replacer(key, value) {
...
+      return {
+        data: UploadedAvatarURL.serialize(value),
+        [SERIALIZED_TYPE_FIELD_NAME]: 'UploadedAvatarURL',
+      };
+    case (FallbackAvatarURL.prototype: $FlowFixMe):
+      return {
+        data: FallbackAvatarURL.serialize(value),
+        [SERIALIZED_TYPE_FIELD_NAME]: 'FallbackAvatarURL',
+      };
+    case (Immutable.Map.prototype: $FlowFixMe):
```
gnprice, line 76: This bit isn't NFC because `instanceof Immutable.Map` would accept an `Immutable.OrderedMap` (the one subclass `Immutable` has for that class), and this won't.

... Which is a fix to a latent bug! We don't currently use `OrderedMap`, but if we did, the old code would cheerfully serialize it, and then deserialize it as a plain `Immutable.Map`.

gnprice (commented):
Thanks @chrisbobbe ! All LGTM and ready to merge, modulo a small comment below.

src/boot/replaceRevive.js
chrisbobbe: Ah, right, good catch—I'll make a note in the commit message and remove the [nfc] mark.

chrisbobbe:
Thanks for the review! I just pushed a new revision with that commit-message change.

gnprice:
Thanks -- merged!

PR #4364, opened 2020-12-30
narrow: Convert more emails to user IDs

gnprice:
This is the next tranche after #4361.

There's some additional places the new `UserItemById` can be used, much like `UserAvatarWithPresenceById`. In a draft branch I've converted those, but haven't finished following the chain of further simplifications that those enable. Those will come in a future PR.

After this PR, there's just a handful of places remaining where we consume the emails from a PM narrow. We'll shortly eliminate those, and then stop storing emails in `Narrow` values altogether.

### DIFF
--- a/src/title-buttons/titleButtonFromNarrow.js
+++ b/src/title-buttons/titleButtonFromNarrow.js
@@ -19,7 +19,11 @@ export const InfoButton: ComponentType<Props> = props =>
       stream: streamName => <InfoNavButtonStream {...props} />,
       topic: (streamName, topic) => <InfoNavButtonStream {...props} />,
       pm: (emails, ids) =>
-        ids.length === 1 ? <InfoNavButtonPrivate {...props} /> : <InfoNavButtonGroup {...props} />,
+        ids.length === 1 ? (
+          <InfoNavButtonPrivate userId={ids[0]} color={props.color} />
+        ) : (
+          <InfoNavButtonGroup {...props} />
+        ),
     },
     () => false,
   );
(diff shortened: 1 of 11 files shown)


src/title-buttons/titleButtonFromNarrow.js
```diff
@@ -3,51 +3,33 @@ import React from 'react';
...
+export const InfoButton: ComponentType<Props> = props =>
+  caseNarrowDefault(
+    props.narrow,
+    {
+      stream: streamName => <InfoNavButtonStream {...props} />,
+      topic: (streamName, topic) => <InfoNavButtonStream {...props} />,
+      pm: (emails, ids) =>
+        ids.length === 1 ? <InfoNavButtonPrivate {...props} /> : <InfoNavButtonGroup {...props} />,
+    },
+    () => false,
```
chrisbobbe, line 28: `title buttons [nfc]: Use a more straightforward React idiom, and caseNarrow.`

This commit makes a nice simplification!

Returning `false` in the default of `caseNarrowDefault` is totally fine, and it's the most NFC-thing to do in this commit. But when we want to explicitly set the entire return value of a React component('s `render`) to be nothing, I think we often favor `null` over `false`. I don't feel very strongly either way (do you?), but this stood out to me just a tiny bit. 🙂

chrisbobbe (commented):
Thanks, this LGTM! Just a tiny nit below, if that.

src/title-buttons/titleButtonFromNarrow.js
gnprice: I think it's fine. :slightly_smiling_face:

Certainly we often use `false` implicitly as a React node, in a form like `{someCondition() && <SomeComponent />}`. So I think the reader needs to be aware it's possible anyway.

gnprice:

PR #4366, opened 2020-12-30
Switch shebang to use /usr/bin/env

WesleyAC:
This is more portable, allowing me to run the install script on NixOS, which
does not have a `/bin/bash` binary, but does have `/usr/bin/env`. Any system
that has `/bin/bash` should have `/usr/bin/env` as well.


gnprice:
Thanks @WesleyAC ! LGTM -- merged.

PR #4368, opened 2020-12-30
narrow: Convert typing-status code to all user IDs

gnprice:
This follows #4364 and the chain of PRs that led to that one.

It eliminates one of the two remaining places we use the emails from a PM narrow, in the typing-status code. Along the way it pulls in https://github.com/zulip/zulip/pull/16987 and some other changes to the code shared from the webapp.

The last remaining consumer of the email data in a PM narrow is an important one: it's where we actually fetch messages from the server, or rather in `apiNarrowOfNarrow` where we prepare the `ApiNarrow` value we'll use to do so. As documented in the tip commit of this PR, the server only started accepting IDs there in Zulip Server 2.1. So we can't eliminate that code, or the code that stores emails into `Narrow` values in the first place, until we drop support for server versions 2.0.x and older.

Zulip Server 2.1 was released a little over a year ago, so it's possible that that's already something that would be reasonable to do. I want to make that decision after getting some data, though, about how many of our users are still on an older server version -- I believe the app currently works with a server as old as 1.8.x, maybe older. We'll get that data in Sentry once the next release is out, containing #4326; it's past time for a release anyway, so that should be soon.

In the meantime (and particularly in case the answer turns out to be that a lot of users are still on older servers), I plan to do some small refactoring so that all the other code that consumes narrows can be completely unaware that they contain emails, other than the few bits that still have to touch them.

### DIFF
--- a/docs/howto/shared.md
+++ b/docs/howto/shared.md
@@ -61,9 +61,10 @@ $ cd static/shared  # the root of the @zulip/shared package's source
 $ git checkout master
 $ git pull --ff-only
 
-$ npm version patch --no-git-tag-version \
-    --message 'shared: Bump version to %s.'
+   # (These steps can probably become a `version` NPM script.)
+$ npm version patch --no-git-tag-version
    # Suppose the new version is 0.0.3.  Then:
+$ git commit -am 'shared: Bump version to 0.0.3.'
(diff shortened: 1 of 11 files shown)


chrisbobbe:
LGTM, thanks! Merged.

Issue #4357, opened 2020-12-24
Narrow lands far in the past, every first time, on iOS


gnprice:
This is an instance of symptoms like #3457, with a specific repro recipe. It can't explain some of the previous reports on #3457, so it doesn't replace that one. It's a pretty annoying bug that will come up frequently for anyone on iOS, so it's a priority to diagnose and fix.

To reproduce:
* Launch the app, on iOS.
* Navigate to "streams", and then to some particular stream that has a long history. (I'm using one that has no unreads.)
* Observe that the messages being shown are from months ago, and several screenfuls up from the end. That's the bug.

Then for further color:
* Navigate back out to "streams", and then back in to that stream.
* Observe that the messages being shown are from very near the bottom.
  * For me it's actually not quite at the bottom -- in fact it looks like I'm getting the bottom of the viewport right at the top of the very last message. Or in other words, it's showing the last messages *except* the very last one, which I have to scroll down to see. That seems likely to be its own separate bug. In any case it's a somewhat milder symptom than the "many screenfuls back" symptom we're focusing on in this discussion.
* To repeat the repro: close the app entirely, then launch it again.

My repros just now were on an iPhone XR, running iOS 14.2, and v27.157 of the app.

By contrast, when I try this on Android -- specifically on a Pixel 4 running Android 11, and again v27.157 of the app -- the issue doesn't reproduce. I get shown the very latest messages, just as I should, every time.

The behavior that's going wrong here is in the scroll position of the message list, so inside the webview we use for the message list. We have some logic in the JS inside that webview to try to scroll the right part of the message list into view. Because this is iOS-specific, the issue is probably in some way that Safari's behavior differs from Chrome, and how our scrolling logic interacts with that.

The next step for debugging this will probably involve connecting the Safari developer tools to the webview:
https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#debug-webview-on-ios
and using those to study what's happening.

Because the bug happens just as we open the webview, this might not be enough on its own -- by the time you connect the dev tools, it'll be after the bug happened. You can get further information by editing the webview JS code (in `js.js`) to report information with e.g. `console.log`, or making other changes and seeing what happens.

gnprice:
A few more details from my repro observations, in case they're helpful for further debugging:
* From the point in the past where the message list first shows, it's 5 screenfuls (exactly, to within the precision of my measuring by scrolling a screenful at a time) to get down to the bottom. The messages are from about 8 months ago.
[...]

Issue #3921, opened 2020-02-20
Narrow sometimes lands one screenful in the past


gnprice:
This is forked off from #3457 -- it's a specific bug we discovered on that thread, which falls into that class of symptoms broadly defined, but appears to be a different bug from the one causing the early reports in that thread.

Specifically, as @ray-kraesig [reported](https://github.com/zulip/zulip-mobile/issues/3457#issuecomment-555234717):
> The test case involved repeatedly entering and leaving a stream (with all messages read) containing a handful of unusually-long individual messages. [...] No reloading of the app or its assets was performed.
>
> When `handleInitialLoad` is called, `document.body.scrollHeight` and `.clientHeight` would take on one of two pairs of values: [one correct, one with a mysterious zero]
>
> The latter, of course, is incorrect; its appearance was consistently associated with the message-list initially being scrolled to a point in the past. Conversely, the former was consistently associated with the message-list initially being scrolled to the bottom, as expected.

We investigated some more. Quoting my summary of what we learned:

---

OK, here are some things we learned in the chat discussion [[here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/clientHeight.20and.20friends/near/798730)]:

* Both in @ray-kraesig 's repro, and when I repro it (both of us in Android emulators), the symptom is we land right about one screenful too high. (To within <2px in Ray's case; about 54px more than a screenful in mine.)
* The observations above, where `scrollHeight` is too big, are kind of puzzling as a cause for that symptom! Because in that case `scrollToAnchor` should be trying to scroll us too far _down_.
* But, we also found a diagnosis for something that puts us a screenful too high! The `resize` handler is getting called as we go from a `clientHeight` of 0 to the actual `clientHeight`. When the viewport gets taller, that handler is meant to scroll us up (as measured from the top of the viewport to the top of the content) in order to keep the bottom of the viewport at the same spot in the content. So, it dutifully scrolls us up by one screenful.
* The following patch, which cuts out that spurious resize, resolves the issue completely in my repro case:

```diff
 let viewportHeight = documentBody.clientHeight;
 
 window.addEventListener('resize', event => {
+  if (viewportHeight === 0) {
+    viewportHeight = documentBody.clientHeight;
+    return;
+  }
   const difference = viewportHeight - documentBody.clientHeight;
```

[...]

PR #4370, opened 2020-12-31
webview: Fix race when setting scroll position

titanous:
Setting innerHTML triggers an asynchronous render of the DOM. On iOS this reliably happens _after_ the code that comes immediately after the `innerHTML` set. This means that if we calculate a scroll position right after a content update, we can get the wrong `scrollHeight` and present the wrong position.

A nested `requestAnimationFrame` callback reliably runs code after the `scrollHeight` is updated, so use that to get the accurate scroll target.

References:
https://twitter.com/rauschma/status/1288868746682081285
https://macarthur.me/posts/when-dom-updates-appear-to-be-asynchronous

Test Plan: Manually loaded streams on iPhone 12 (iOS 14.3)

Fixes: #4357

I don't have an Android device to test on, but I suspect this will resolve or improve at least some variants of #3457 and #3921.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -511,6 +511,19 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
   window.scrollBy(0, newBoundRect.top - prevBoundTop);
 };
 
+// Run a function after the next repaint. This is important if we have just set
+// innerHTML and need to read from the DOM, as the repaint happens asynchronously.
+// See: https://macarthur.me/posts/when-dom-updates-appear-to-be-asynchronous
+const runAfterRepaint = (fn: () => void) => {
+  requestAnimationFrame(() => {
+    // this runs before the next repaint
+    requestAnimationFrame(() => {
+      // this runs after the repaint
+      fn();
+    });
+  });
+};
+
 const handleUpdateEventContent = (uevent: WebViewUpdateEventContent) => {
   let target: ScrollTarget;
   if (uevent.updateStrategy === 'replace') {
@@ -531,15 +544,17 @@ const handleUpdateEventContent = (uevent: WebViewUpdateEventContent) => {
 
   rewriteHtml(uevent.auth);
 
-  if (target.type === 'bottom') {
-    scrollToBottom();
-  } else if (target.type === 'anchor') {
-    scrollToMessage(target.messageId);
-  } else if (target.type === 'preserve') {
-    scrollToPreserve(target.msgId, target.prevBoundTop);
-  }
+  runAfterRepaint(() => {
+    if (target.type === 'bottom') {
+      scrollToBottom();
+    } else if (target.type === 'anchor') {
+      scrollToMessage(target.messageId);
+    } else if (target.type === 'preserve') {
+      scrollToPreserve(target.msgId, target.prevBoundTop);
+    }
 
-  sendScrollMessageIfListShort();
+    sendScrollMessageIfListShort();
+  });
 };
 
 // We call this when the webview's content first loads.
@@ -584,7 +599,7 @@ const handleUpdateEventTyping = (uevent: WebViewUpdateEventTyping) => {
   const elementTyping = document.getElementById('typing');
   if (elementTyping) {
     elementTyping.innerHTML = uevent.content;
-    setTimeout(() => scrollToBottomIfNearEnd());
+    runAfterRepaint(() => scrollToBottomIfNearEnd());
   }
 };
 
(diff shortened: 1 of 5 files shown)


src/webview/js/js.js
```diff
@@ -511,6 +511,19 @@ const scrollToPreserve = (msgId: number, prevBoundTop: number) => {
...
 
+// Run a function after the next repaint. This is important if we have just set
+// innerHTML and need to read from the DOM, as the repaint happens asynchronously.
+// See: https://macarthur.me/posts/when-dom-updates-appear-to-be-asynchronous
+const runAfterRepaint = (fn: () => void) => {
+  requestAnimationFrame(() => {
+    // this runs before the next repaint
+    requestAnimationFrame(() => {
+      // this runs after the repaint
+      fn();
```
gnprice, line 522: So one thing that's less than optimal about this is that it means we're guaranteed to have a frame appear with an intermediate state: we've introduced the new content, but we're at some uncontrolled scroll position that may not be the one we want.

That's still a huge improvement over the bug this fixes -- better to be wrong for a frame than to be wrong and stay that way.

This seems likely to be the cause of the jank this introduces on Android, described above. I imagine it might also cause some jank on iOS, too.

gnprice (commented):
Thanks @titanous for this investigation and fix!

I tried this out on Android, in an emulated Pixel 3a running Android 11 [**EDIT**: and Chrome 83.0.4103.106]. There, I found:
* At master 602bcc16c, I actually can't reproduce #3921 (which has a repro recipe that *used* to repro a good fraction of the time), nor #3457 (though that's less remarkable because we've never had a reliable repro recipe for that one.)
* With this change, I reproduced #3921 once (that's the one-screenful-too-high bug.) That's out of a bunch of tries, so it's possible that it reproes at the same low rate both with and without this change.
* At master, when I go through the repro recipe for #4357 -- entering a given stream for the first time after launching the app from scratch -- the behavior is perfect: we show the end of the history, with no jank. (No surprise there, as we'd already observed this issue was iOS-only.)
* With this change, when I go through the same repro, we wind up in the right spot but there's some fairly bad jank: the messages appear, we're in the right spot, then maybe a few hundred ms later there's a brief flash where some other messages appear, then it's back to the right spot at the end of the history.

So probably the right thing to do is to make this conditional on platform. I can add a bit of logic for that in the course of merging.

src/webview/js/js.js
titanous: One thing we could try instead is `setTimeout(fn, 0)` inside the first rAF. In theory this could run a bit earlier but I don't have a good idea if that'll be less reliable.

titanous:
>With this change, when I go through the same repro, we wind up in the right spot but there's some fairly bad jank: the messages appear, we're in the right spot, then maybe a few hundred ms later there's a brief flash where some other messages appear, then it's back to the right spot at the end of the history.

I believe that this is the `oldMessagesAdded` content update [described here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Wrong.20scroll.20position.20on.20iOS/near/1090401).

gnprice:
And merged! With the extra logic to make it iOS-only, as I mentioned. Thanks again.

gnprice:
> I believe that this is the `oldMessagesAdded` content update [described here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Wrong.20scroll.20position.20on.20iOS/near/1090401).

Yeah, that seems quite likely given the timing (a few hundred ms after the messages first show up.)

src/webview/js/js.js
gnprice: Yeah. And I think it'll still result in showing a frame based on the intermediate state -- AIUI the point of rAF is that its callback gets invoked just at the last moment before the frame gets turned to pixels.

What about just one rAF, and invoking the callback directly in that? Not sure if you've already tested and found that `scrollHeight` isn't yet updated at that point, but a priori it seems like it might be. The blog post is doing a different test, of whether what's visible on the screen has been updated -- and it'd actually make a lot of sense for the API to be that when a `requestAnimationFrame` callback is invoked, all the DOM layout properties are guaranteed to be ready and just the repaint hasn't happened yet.

titanous: Hmm, I think I may have messed something up when testing a single rAF earlier (I probably failed to regenerate the webview JS). It seems to be working with only one call, which I agree makes sense. I'm going to do some more testing but I'm not seeing a difference. The same content update jank is still there, but that's unsurprising.

gnprice: Ah, aha, here's some previous discussion of this: https://github.com/zulip/zulip-mobile/issues/3457#issuecomment-555252129

I *thought* I remembered a discussion of `requestAnimationFrame` in a context like this, but I was having trouble finding it. There it is.

The conclusion there is... that either a `setTimeout` inside the first rAF, or just running the code directly inside a single rAF, are likely solutions :slightly_smiling_face: . Just needs testing to determine which, if either, actually seems to work well.

PR #4372, opened 2020-12-31
Use function declarations at toplevel.

chrisbobbe:
Following discussion yesterday [here](https://github.com/zulip/zulip-mobile/pull/4367#discussion_r550311033).

### DIFF
--- a/src/common/UserAvatar.js
+++ b/src/common/UserAvatar.js
@@ -23,7 +23,7 @@ type Props = $ReadOnly<{|
  * @prop [children] - If provided, will render inside the component body.
  * @prop [onPress] - Event fired on pressing the component.
  */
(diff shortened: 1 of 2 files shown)


src/common/UserAvatar.js
```diff
@@ -53,6 +53,6 @@ const UserAvatar = function UserAvatar(props: Props) {
       </Touchable>
     </View>
   );
-};
+}
 
 export default UserAvatar;
```
gnprice, line 58: How about putting this directly on the function definition, as we typically do where applicable for class components? So:
```js
export default function UserAvatar(props: Props) {
```

That seems just a bit simpler still.

gnprice (commented):
LGTM, thanks! Merging. One followup comment below.

PR #4373, opened 2020-12-31
redux-persist: Stop saving `REHYDRATE` payload to disk.

chrisbobbe:
The expensive part of persisting the state to disk is running
`stringify` (see replaceRevive.js). While #4348 brought this time
from almost 2000ms down to, perhaps, 300ms on one account on my
device [1], that's still a long time, especially for work that
doesn't need to be done.

See discussion [2] for this solution.

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60remotedev-serialize.60.20and.20.60jsan.60/near/1083700
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60remotedev-serialize.60.20and.20.60jsan.60/near/1089466

### DIFF
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -114,7 +114,12 @@ export default function createPersistor (store, config) {
     rehydrate: adhocRehydrate,
     pause: () => { paused = true },
     resume: () => { paused = false },
-    purge: (keys) => purgeStoredState({storage, keyPrefix}, keys)
+    purge: (keys) => purgeStoredState({storage, keyPrefix}, keys),
+
+    // Only used in `persistStore`, to force `lastState` to update
+    // with the results of `REHYDRATE` even when the persistor is
+    // paused.
+    _resetLastState: () => { lastState = store.getState() }
   }
 }
 
--- a/src/third/redux-persist/persistStore.js
+++ b/src/third/redux-persist/persistStore.js
@@ -30,6 +30,34 @@ export default function persistStore (store, config = {}, onComplete) {
         }
         try {
           store.dispatch(rehydrateAction(restoredState, err))
+
+          // Ensure the payload of `REHYDRATE` isn't persisted.
+          //
+          // The `REHYDRATE` payload contains exactly every useful
+          // piece of state; there's nothing useful in the existing
+          // state before it arrives that it has to merge with. Since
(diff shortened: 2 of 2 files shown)


chrisbobbe:
Hmm, I guess there's something I don't yet understand going on. The chart shows six startups of the app (release builds, on CZO, with my iPhone). As expected, there's just one tall bar per startup, instead of two. But the tall bars in the "after" results are significantly taller than the ones in "before"...

image

chrisbobbe:
> Hmm, I guess there's something I don't yet understand going on.

Ah, I think I might havej it. We shouldn't set `lastState` to be the `REHYDRATE` payload itself; we should set it to the value of `store.getState()` immediately after `REHYDRATE` has gone through.

chrisbobbe:
Here's a chart for starting up the app (on a release build on one account on CZO on my iPhone) six times: before the change, and after.

image

Hmm, interesting! I see the expected result that there's just one tall bar per startup (that's the `/register` response), instead of two (`REHYDRATE` payload and the `/register` response). But I also see that the number of `stringify` calls has been cut pretty much in half, and I don't yet understand why.

gnprice:
> Hmm, interesting! I see the expected result that there's just one tall bar per startup (that's the `/register` response), instead of two (`REHYDRATE` payload and the `/register` response). But I also see that the _total_ number of `stringify` calls has been cut pretty much in half, and I don't yet understand why.

One thing is that when we've made updates across the state tree, that doesn't result in just one call to `stringify` -- each top-level subtree like `state.messages`, `state.users`, etc., is persisted separately, so will cause a separate call to `stringify`.

So the tall bars come from `/register` and rehydrate, but so do a lot of shorter bars. The tall bars are presumably the ones for `state.users`, since that's the giant part of the state.

gnprice:
Thanks -- merged! I made a tiny edit to add a blank line above each of the two new comments, to separate it visually from the code above.

chrisbobbe:
> One thing is that when we've made updates across the state tree, that doesn't result in just one call to `stringify` -- each top-level subtree like `state.messages`, `state.users`, etc., is persisted separately, so will cause a separate call to `stringify`.
> 
> So the tall bars come from `/register` and rehydrate, but so do a lot of shorter bars. The tall bars are presumably the ones for `state.users`, since that's the giant part of the state.

Oh OK, right, that makes sense! Thanks for the review, tweak, and merge.

PR #4374, opened 2020-12-31
webview: Call requestAnimationFrame once on iOS.

titanous:
After extensive testing, it's clear that we only need one call in order to reliably get the `scrollHeight`. I also confirmed that the jank is gone in an Android emulator, but I left the conditional intact as it works reliably without rAF at all on Android.

References: #4370


gnprice:
Great, thanks for that investigation! Merged.

PR #4332, opened 2020-12-10
narrow/recipient: Fix more bugs, and prep for bigger refactors

gnprice:
This follows on some work prompted by #4317, begun in #4330.  I have a longer branch on top of this which is still in development, which does much more to encapsulate the fiddly details of how we describe PM conversations, and as a result takes care of most of #4035 which has been the main obstacle to #3133.  This PR contains the changes from that branch which are ready and stand on their own.

Among these changes are some small bugfixes, and other cleanups in areas of the code where the larger branch will do further work.  This branch starts with a couple of infrastructural changes to smooth the way for using error logging in more parts of our code.

PR #4382, opened 2021-01-05
narrow: Stop storing emails!

gnprice:
This is the next in the long chain of PRs #4330, #4332, #4335, #4339, #4342, #4346, #4356, #4361, #4364, and most recently #4368. This one completes a major objective of all that refactoring: at the end of it, we no longer store emails in Narrow values.

After this change, the portion of #4333 that's about PMs, emails, and user IDs -- aka the portion of #3764 that's about narrows -- is complete.

Still open from #4333 is to convert stream and topic narrows from stream names to stream IDs.  I'm hopeful that that will be simpler: (a) unlike the situation for PMs, there's just one stream mentioned at a time, so there's no question of sorting, and (b) there isn't the "include self or not" complication that's bedeviled much of our code related to PMs.  In other words, stream and topic narrows don't suffer from #4035.

As discussed at #4368, in one of the spots this touches we still need emails, namely to make get-messages requests to the server (so long as we support servers older than 2.1). But we can handle that the same way we've handled such mismatches internally to the app during this refactoring, and the same way we've handled the reverse of them where we receive emails and now want to store user IDs: use the `allUsersById` map (or its inverse) to convert one form to the other. That means we don't need the emails from the Narrow value, and the rest of these changes can proceed.

There's still plenty more places around the app where we can replace emails with user IDs, i.e. #3764. I still have a draft branch on top of this one with some of them. I'm hopeful that much of what remains will be easier now that emails are gone from narrows, because that's such a central data structure that a lot of code ultimately has to feed or is fed by.

### DIFF
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -18,8 +18,6 @@ import ZulipAsyncStorage from './ZulipAsyncStorage';
 import createMigration from '../redux-persist-migrate/index';
 import { provideLoggingContext } from './loggingContext';
 import { tryGetActiveAccount } from '../account/accountsSelectors';
-import { keyFromNarrow } from '../utils/narrow';
-import { objectFromEntries } from '../jsBackport';
 
 if (process.env.NODE_ENV === 'development') {
   // Chrome dev tools for Immutable.
@@ -226,31 +224,17 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
   // `AvatarURL`.
   '18': dropCache,
 
-  // Change format of keys representing narrows, from JSON to our format.
-  '19': state => ({
+  // Change format of keys representing narrows: from JSON to our format,
+  // then for PM narrows adding user IDs.
+  '21': state => ({
     ...dropCache(state),
-    drafts: objectFromEntries(
-      // Note this will migrate straight to our current format, even after
-      // that changes from when this migration was written!  That saves us
-      // from duplicating `keyFromNarrow` here... but calls for care in
-      // migrations for future changes to `keyFromNarrow`.
-      Object.keys(state.drafts).map(key => [keyFromNarrow(JSON.parse(key)), state.drafts[key]]),
-    ),
-  }),
-
-  // Change format of keys representing PM narrows, adding user IDs.
-  '20': state => ({
-    ...dropCache(state),
-    drafts: objectFromEntries(
-      Object.keys(state.drafts)
-        // Just drop any old-style, email-only PM keys.  Converting them
-        // would require using additional information to look up the IDs,
-        // which would make this more complex than any of our other
-        // migrations.  Drafts are inherently short-term, and are already
-        // discarded whenever switching between accounts.
-        .filter(key => !key.startsWith('pm:s:'))
-        .map(key => [key, state.drafts[key]]),
-    ),
+    // The old format was a rather hairy format that we don't want to
+    // permanently keep around the code to parse.  For PMs, there's an
+    // extra wrinkle in that any conversion would require using additional
+    // information to look up the IDs.  Drafts are inherently short-term,
+    // and are already discarded whenever switching between accounts;
+    // so we just drop them here.
+    drafts: {},
   }),
 
   // TIP: When adding a migration, consider just using `dropCache`.
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -18,6 +18,7 @@ import ZulipAsyncStorage from './ZulipAsyncStorage';
 import createMigration from '../redux-persist-migrate/index';
 import { provideLoggingContext } from './loggingContext';
 import { tryGetActiveAccount } from '../account/accountsSelectors';
+import { objectFromEntries } from '../jsBackport';
 
 if (process.env.NODE_ENV === 'development') {
   // Chrome dev tools for Immutable.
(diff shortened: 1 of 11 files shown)


chrisbobbe:
> #4336

#4335, I think? 🙂

src/utils/narrow.js
```diff
@@ -399,14 +399,28 @@ export const isSearchNarrow = (narrow?: Narrow): boolean =>
...
+    pm: (emails_, ids) => {
+      const emails = [];
+      for (const id of ids) {
+        const email = allUsersById.get(id)?.email;
+        if (email === undefined) {
+          throw new Error('apiNarrowOfNarrow: missing user');
+        }
+        emails.push(email);
+      }
+      // TODO(server-2.1): just send IDs instead
```
chrisbobbe, line 411: If it makes a big difference to the amount of work the server has to do, we could check the stored server version in Redux and courteously send IDs to the newer servers, while still sending emails to older ones. But it might not make much of a difference. 🙂

gnprice:
> #4335, I think?

Indeed, thanks :slightly_smiling_face:

src/boot/store.js
```diff
@@ -253,6 +253,16 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
     ),
   }),
 
+  // Change format of keys representing PM narrows, dropping emails.
+  '21': state => ({
+    ...dropCache(state),
+    drafts: objectFromEntries(
+      Object.keys(state.drafts)
+        .map(key => key.replace(/^pm:d:(.*?):.*/s, 'pm:$1'))
```
chrisbobbe, line 261: Good, I think `'21'` is safe from the potential issue mentioned in `'19'`:

```
      // Note this will migrate straight to our current format, even after
      // that changes from when this migration was written!  That saves us
      // from duplicating `keyFromNarrow` here... but calls for care in
      // migrations for future changes to `keyFromNarrow`.
```

In particular:

- If [`'21'`] is the sequence of migrations that run at startup, drafts in `'21'`'s input will be in the `pm:d:` format, which this string replacement handles correctly.
- If [`'20'`, `'21'`] is the sequence of migrations that run at startup, drafts in `'21'`'s input will be...empty. (Drafts in `'20'`'s input would have all had `pm:s:` keys, and `'20'` would have removed them all.) Drafts being empty in `'21'`'s input is fine; it just gives an empty output.
- If [..., `'19'`, `'20'`, `'21'`] is the sequence of migrations that run at startup, drafts in `'21'`'s input will have keys that are already in the up-to-date `pm:` format (because of the quirk with `'19'`). This is fine; `'21'` is idempotent on those keys because the `.replace` neither throws an error with that input nor changes it at all.

chrisbobbe (commented):
Great, thanks! This LGTM; just a small query and a comment below.

src/utils/narrow.js
gnprice: Yeah, I don't think it materially affects the work the server has to do. It'll want to look up the users in the database in any event (for one thing, to make sure they're valid users that are in your realm); and the database will have indexes so that looking up by ID or by email are both fast, because those are both common operations.

src/boot/store.js
gnprice: Thanks for checking this!

> * If [`'20'`, `'21'`] is the sequence of migrations that run at startup, drafts in `'21'`'s input will be...empty.

(It'll have no PMs, that is -- it'll still have drafts for stream messages.)


> * If [(...,) `'19'`, `'20'`, `'21'`] is the sequence of migrations that run at startup, drafts in `'21'`'s input will have keys that are already in the up-to-date `pm:` format

Hmm. I think there's actually a bug here, though! In migration 19:
```js
      Object.keys(state.drafts).map(key => [keyFromNarrow(JSON.parse(key)), state.drafts[key]]),
```
the `JSON.parse` for a PM narrow will return something like `{ operator: 'pm-with', operand: 'foo@example.com,bar@example.org' }`. In fact this is really two bugs:
* That's the old shape of `Narrow`, which `keyFromNarrow` no longer accepts.
* That has emails and no user IDs. So there's nothing `keyFromNarrow` could possibly do to convert it to the new format.

The bug was introduced with #4346, where `Narrow` changed and so the type expected by `keyFromNarrow` changed. The reason Flow didn't catch that there was this hole in the encapsulation is that `JSON.parse` there.

This probably means that my plan in #4339 (at https://github.com/zulip/zulip-mobile/pull/4339#discussion_r543603022 ) wasn't the right one in retrospect. It would have been better to simply drop the drafts, or else bite the bullet and duplicate `keyFromNarrow`.

Happily no general users can have been affected at all, because we haven't made a release since this was introduced.

I think I'll fix this by rolling up migrations 19 and 20 into a new migration 21, which just drops drafts entirely (plus `dropCache`). That'll have the same effect in practice as if I'd gone with a "drop drafts" plan in #4339 from the beginning -- each user will have drafts dropped once. And the current migration 20 is already meant to drop all PM drafts, so only stream drafts are newly affected.

chrisbobbe: Oh OK, great—I'm glad you found this! 🙂

> I think I'll fix this by rolling up migrations 19 and 20 into a new migration 21

SGTM.

gnprice:
New revision pushed!

src/boot/store.js
```diff
@@ -226,31 +226,17 @@ const migrations: { [string]: (GlobalState) => GlobalState } = {
...
+  // Change format of keys representing narrows: from JSON to our format,
+  // then for PM narrows adding user IDs.
+  '21': state => ({
     ...dropCache(state),
-    drafts: objectFromEntries(
-      // Note this will migrate straight to our current format, even after
-      // that changes from when this migration was written!  That saves us
-      // from duplicating `keyFromNarrow` here... but calls for care in
-      // migrations for future changes to `keyFromNarrow`.
-      Object.keys(state.drafts).map(key => [keyFromNarrow(JSON.parse(key)), state.drafts[key]]),
```
chrisbobbe, line 237: Looks like `keyFromNarrow` and `objectFromEntries` need their imports removed (but then don't forget to bring back `objectFromEntries`'s for migration 22 🙂).

chrisbobbe (commented):
Thanks, @gnprice! LGTM except for one comment below.

gnprice:
Thanks for the review! Merged, with that fix.

PR #4385, opened 2021-01-07
Fix a `recipients` bug, preparing to use recent-PMs data

gnprice:
I wrote up today a draft branch that will take care of #3133, replacing the previous #3535 and #4104, to use the server's handy `recent_private_conversations` feature in order to populate the "PM conversations" screen both more completely and more efficiently.

Along the way -- actually while rebasing and studying #4104, before I decided it'd be better to take a different approach than it or #3535 did -- I discovered a bug in `pmKeyRecipientsFromIds`: it's supposed to work whether the input contains or excludes self, but it doesn't live up to that spec in the case of a self-1:1. This has been a latent bug because all of the representations for PM conversations that we deal with so far represent the self-1:1 thread with a singleton list, not empty. But the `recent_private_conversations` data structure does use an empty list for the self-1:1 thread, so it would trip on this bug and make it a live one.

This PR fixes that bug, and also takes care of several other issues that that change turns up. (Including discovering a Flow bug which I then reported. Though that part was actually a couple of months ago -- I was looking at some of this code as part of #4299 and #4304, and that work led to the first two commits here and that Flow bug report.)

### DIFF
--- a/src/__tests__/collections-test.js
+++ b/src/__tests__/collections-test.js
@@ -1,0 +1,23 @@
+/* @flow strict-local */
+
+const { mapOrNull } = require('../collections');
+
+describe('mapOrNull', () => {
+  test('behaves right in happy case', () => {
+    expect(mapOrNull([10, 20], (x, i, a) => x + i + a.length)).toEqual([12, 23]);
+  });
+
+  test('returns null on nullish item', () => {
+    expect(mapOrNull([1, 2, 3], x => (x === 2 ? null : x))).toEqual(null);
+    expect(mapOrNull([1, 2, 3], x => (x === 2 ? undefined : x))).toEqual(null);
+  });
+
+  test('aborts on nullish item', () => {
+    const log = [];
+    mapOrNull([1, 2, 3], x => {
+      log.push(x);
+      return x === 2 ? null : x;
+    });
+    expect(log).toEqual([1, 2]);
+  });
+});
--- a/src/collections.js
+++ b/src/collections.js
@@ -1,0 +1,41 @@
+/**
+ * Handy generic functions for operating on JS collections.
+ *
+ * Any function here is the sort of thing that perhaps should be in the
+ * language's standard library.
+ *
+ * See also
+ *  * `jsBackport.js`, for things that actually are in newer versions of JS,
+ *    or late-stage proposals for future versions
+ *  * `generics.js`, for things that operate on types
+ *
+ * @flow strict-local
+ */
+
+/**
+ * Like `Array#map`, but returns null if `f` ever returns a nullish value.
+ *
+ * If `f` returns nullish at any element, the iteration stops and `f` is not
+ * called on any later elements.
+ *
+ * This replaces either (a) writing an explicit loop, in order to be able to
+ * abort on a nullish value, or (b) using `Array#map` and then running
+ * through the list again to look for nullish values.
+ */
+export function mapOrNull<T, U>(
+  array: $ReadOnlyArray<T>,
+  f: (T, number, $ReadOnlyArray<T>) => U,
+  // `Array#map` has one more parameter, "thisArg".  Skipping that because
+  //   it seems like there isn't a way to type it properly in Flow (flowlib
+  //   has `any`), and we haven't had a need for it.
+): Array<$NonMaybeType<U>> | null {
+  const result = [];
+  for (let i = 0; i < array.length; i++) {
+    const r = f(array[i], i, array);
+    if (r == null) {
+      return null;
+    }
+    result.push(r);
+  }
+  return result;
+}
--- a/src/utils/__tests__/recipient-test.js
+++ b/src/utils/__tests__/recipient-test.js
@@ -1,8 +1,11 @@
+/* @flow strict-local */
+
 import {
   normalizeRecipientsAsUserIds,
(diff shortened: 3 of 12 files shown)


src/collections.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * If `f` returns nullish at any element, the iteration stops and `f` is not
+ * called on any later elements.
+ *
+ * This replaces either (a) writing an explicit loop, in order to be able to
+ * abort on a nullish value, or (b) using `Array#map` and then running
+ * through the list again to look for nullish values.
+ */
+export function mapOrNull<T, U>(
+  ths: $ReadOnlyArray<T>,
```
chrisbobbe, line 26: Hmm, what does `ths` stand for? After reading the comment about `thisArg`, it makes me think it might have to do with that, but I think it doesn't.

src/collections.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ * Like `Array#map`, but returns null if `f` ever returns a nullish value.
+ *
+ * If `f` returns nullish at any element, the iteration stops and `f` is not
+ * called on any later elements.
+ *
+ * This replaces either (a) writing an explicit loop, in order to be able to
+ * abort on a nullish value, or (b) using `Array#map` and then running
```
chrisbobbe, line 22: > abort on a nullish value

~~nit: abort on a null value~~

Hmm! Right, it does abort on a nullish value, not just `null`. The `== null` (`==`, not `===`) makes it abort on a nullish value. I wasn't expecting the double-equals, but I suppose it makes sense to do one operation instead of two (i.e., `r === null || r === undefined`).

src/__tests__/collections-test.js
```diff
@@ -0,0 +1,23 @@
+/* @flow strict-local */
+
+const { mapOrNull } = require('../collections');
+
+describe('mapOrNull', () => {
+  test('behaves right in happy case', () => {
+    expect(mapOrNull([10, 20], (x, i, a) => x + i + a.length)).toEqual([12, 23]);
```
chrisbobbe, line 7: This would be an odd way to use `mapOrNull` in the wild, right, but it catches all the potential pitfalls, and quite succinctly! Staring at it, I suppose its oddness is good because it makes me go and make sure I understand `mapOrNull` well. 🙂

src/utils/__tests__/recipient-test.js
```diff
@@ -49,6 +50,28 @@ describe('normalizeRecipientsAsUserIdsSansMe', () => {
   });
 });
 
+describe('pmKeyRecipientsFromIds', () => {
+  const allUsersById = new Map([eg.selfUser, eg.otherUser, eg.thirdUser].map(u => [u.user_id, u]));
+  const [self, other, third] = [eg.selfUser, eg.otherUser, eg.thirdUser];
+  // prettier-ignore
+  /* eslint-disable no-multi-spaces */
+  /* eslint-disable array-bracket-spacing */
+  for (const [desc, users, expectedSet] of [
```
chrisbobbe, line 59: nit: Maybe `description` instead of `desc`?; For a split-second I thought of `descending` because that's another thing that comes to mind when I look at arrays (though I may be biased from all the recent stuff about sorting).

chrisbobbe (commented):
Great, glad to be getting closer to #3133! Small comments and a question below.

src/collections.js
gnprice: Ah, I should make it clearer, then. :slightly_smiling_face:

It's meant to stand for "this" -- as in, if you imagine this *were* in the stdlib it'd be a method on Array, so the array it's mapping over would be `this`. But of course `this` is a keyword and I can't just call the parameter that, so I have to munge the name.

In Python there's a fairly common pattern that calls for a function to take an argument that is a class and you'd like to call it "class", and similarly `class` is a keyword and you can't. The standard solution there -- [here it is](https://docs.python.org/3/library/functions.html#classmethod) right in the language reference docs -- is to call it `cls`. So I think in my head I was echoing that usage. But that's not so helpful for a reader who hasn't spent time writing Python :wink: 

Given the "thisArg" thing, probably anything that tries to say "this" will be confusing unless separately explained. Maybe I'll call it `array`.

src/collections.js
gnprice: Yep!

For the API choice, I think "nullish" is most helpful because it means it can work with things like `Map#get` that return undefined as well as with things that, like this function itself, return null.

For the code-style choice: I've gone back and forth, but my current leaning is that it makes sense to use `==` for basically this one pattern, of `== null`.
* It does exactly mean `r === null || r === undefined` -- which makes it just about the only usage of `==` that can be described as cleanly as that.
* It's rather shorter; and it avoids duplicating the expression, which doesn't matter here but can meaningfully reduce mess when it's more complex.
* It's a pretty tightly constrained and bright-line pattern -- so I think it's still quite doable to see `==` in any context other than `== null` and have that stand out as a code smell.

gnprice: (Similarly `!= null`, naturally.)

src/__tests__/collections-test.js
gnprice: :smile:

gnprice:
Thanks for the review! Merged with those tweaks.

PR #4387, opened 2021-01-07
redux-persist: Fix a recent regression in handling migrations.

chrisbobbe:
08b77bb79f went a long way in reducing unnecessary work being done,
but unfortunately I wasn't careful enough about what happens with
migrations, and it introduced a source of hard-to-diagnose bugs that
might show up when migrations are added.

In particular, it's appropriate and necessary to turn around and
save the result of `REHYDRATE` back to the disk in cases where there
has been a migration. See discussion [1].

Rather than just revert 08b77bb79f, though, introduce a condition
that lets us get the benefit of the optimization in the far more
common case (a startup without a migration, vs. a startup with one).

The way we do that does break the separation between our
redux-persist and redux-persist-migrate, by making it so
redux-persist knows about the specifics of redux-persist-migrate.
But, as Greg says [2],

"""
I think that's fine, because a good "how to persistently store your
application's state" solution has to include migrations, and I think
the really high-quality libraries for it have ended up integrating a
specific way to do migrations. So breaking down the separation
between those two seems like the right architectural direction
anyway.
"""

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60dropCache.60/near/1093989
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60dropCache.60/near/1094088

### DIFF
--- a/src/third/redux-persist/persistStore.js
+++ b/src/third/redux-persist/persistStore.js
@@ -29,35 +29,62 @@ export default function persistStore (store, config = {}, onComplete) {
           else purgeKeys.forEach((key) => delete restoredState[key])
         }
         try {
+          // The version (in redux-persist-migrate's terms) that was
+          // current in the previous session.
+          //
+          // Caution: this same expression would give a different
+          // value if it were run after the `store.dispatch` line,
+          // because redux-persist-migrate's store enhancer
+          // `createMigration` mutates `migrations.version` in the
+          // action's payload (see `realVersionSetter` in
+          // redux-persist-migrate).
+          const prevVersion = restoredState.migrations?.version;
+
           store.dispatch(rehydrateAction(restoredState, err))
 
-          // Ensure the payload of `REHYDRATE` isn't persisted.
-          //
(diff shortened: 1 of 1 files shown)


chrisbobbe:
I'm not so sure about the second commit, but I thought I'd include it for discussion at least.

gnprice:
Thanks!

I'm not sure the second commit is necessary, but I think it does simplify thinking about what happens on a migration. So it's probably helpful to include it.

One small code comment is that instead of `type: 'DUMMY'` with a fixed string, you could borrow a tactic from Redux itself and generate a random string (which Redux does for the initial action), to be sure nothing matches it. Like (copying the implementation from `randString` in `exampleData.js`):
```js
{ type: `PERSIST_DUMMY/${randInt(2 ** 54).toString(36)}` }
```

Otherwise this looks good!

chrisbobbe:
> Otherwise this looks good!

Great, thanks for the review! Merged, with that one tweak.

PR #4389, opened 2021-01-07
UserAvatar: Fail gently on bug where auth info is missing.

### LINKED ISSUE #4388: Crash on trying to log into a second account
Repro:
* Be logged into an account, and in the main app UI.
* Profile > Switch account > Add new account.
* Type an org URL, and hit enter.
* The app crashes.

Here's an alternative repro that doesn't quite as much show how easy this bug is to hit, but is easier to reproduce repeatedly in a loop while changing things:
* Have at least two accounts in the list: one logged in, and one logged out. (E.g., log out of one account, if you already have two logged in.)
* Choose the logged-in account, to get the main app UI.
* Profile > Switch account.
* Choose the logged-out account.
* The app crashes.
* (On next startup, you'll be at the pick-account screen, and can choose the logged-in account to repeat.)

I did a `git bisect` (using that handy second repro), and the bug begins at commit c59e286ab. That indeed introduces a `getAuth` where we didn't have one before:
```diff
+  const auth = useSelector((state: GlobalState) => getAuth(state));
```

That should be perfectly fine to do. It's in `UserAvatar`, which is part of the app's normal UI. If we're trying to render this component when we don't have our normal set of data, that's a bug. So we should fix that bug.

Because of the severity of this issue, and particularly because I'd like to get a release out soon because it's been a while, I may try fixing this first with a workaround: e.g., if we're not logged in, return normally from `UserAvatar` but just don't render anything.

gnprice:
This fixes #4388 with a quick workaround.  The underlying issue is
still present -- we shouldn't have this component mounted at all in
this situation.  But fixing that may involve some tricky changes to
our navigation, and this is a high-severity issue -- a crash in an
essential flow for using the app for many people -- so we want to fix
it promptly even if the underlying fix takes some time.

Fixes: #4388

### DIFF
--- a/src/common/UserAvatar.js
+++ b/src/common/UserAvatar.js
@@ -4,7 +4,7 @@ import { ImageBackground, View, PixelRatio } from 'react-native';
 
 import { useSelector } from '../react-redux';
 import { getAuthHeaders } from '../api/transport';
-import { getAuth } from '../selectors';
+import { tryGetAuth } from '../account/accountsSelectors';
 import Touchable from './Touchable';
 import { AvatarURL } from '../utils/avatar';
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, merged! I've just gone through your repro at #4388 (the first one) and the app didn't crash.

gnprice:
Thanks for the fast review!

Issue #4296, opened 2020-11-03
Upgrade react-navigation to v5


chrisbobbe:
Currently blocked by #3804, which has https://github.com/zulip/zulip-mobile/pull/4278 in progress.

Release announcement: https://reactnavigation.org/blog/2020/02/06/react-navigation-5.0/

Upgrade guide: https://reactnavigation.org/docs/upgrading-from-4.x

Docs: https://reactnavigation.org/docs/getting-started

gnprice:
(Mainly to help find this issue from these others: this follows #4248 for v4, #3649 for v3, and #3573 for v2.)

chrisbobbe:
Added the `multi-org` label, since I expect we'll want to settle into React Navigation v5 (which comes with a pretty big API change) before we make the necessary nav UI changes for better multi-org support.

PR #4393, opened 2021-01-11
React Navigation v5 upgrade (part 2)

chrisbobbe:
This is the next PR toward fixing #4296, the React Navigation v5 upgrade, and it includes making the upgrade itself! 🎉 It follows (from earliest to latest) the resolution of #3804 in #4273, #4274, #4278; #4315; and (merged last Friday) #4300.

It starts with some more prep commits for the main upgrade commit, to reduce the size of its diff and to address some quirky differences between v4 and v5. In the main upgrade commit, we drop in @react-navigation/compat to further reduce the size of the diff, but we promptly stop using @react-navigation/compat over several commits (rather than leaving it in semi-indefinitely) because it's completely untyped.

Tested on my physical iPhone 6s, an iPhone 12 Pro simulator, and the office Samsung Galaxy S9.

Fixes: #4296


chrisbobbe:
I'm not sure this PR is the place to make any changes to address it, but one thing that'll be good to keep in mind is that, while the `NavigationService` approach is still supported, it's now explicitly not recommended, when you have access to the `navigation` prop. From a [v5 doc](https://reactnavigation.org/docs/navigating-without-navigation-prop/):

> If you're looking for a way to navigate from inside a component without needing to pass the `navigation` prop down, see [`useNavigation`](https://reactnavigation.org/docs/use-navigation). **Do not** use this method when you have access to a `navigation` prop or `useNavigation` since it will behave differently, and many helper methods specific to screens won't be available.

src/nav/navActions.js
```diff
@@ -29,89 +32,83 @@ export const resetToMainTabs = (): GenericNavigationAction =>
...
 
 /** Only call this via `doNarrow`.  See there for details. */
-export const navigateToChat = (narrow: Narrow): NavigationAction =>
-  StackActions.push({ routeName: 'chat', params: { narrow } });
+export const navigateToChat = (narrow: Narrow): GenericNavigationAction =>
+  StackActions.push('chat', { narrow });
 
-export const navigateToUsersScreen = (): NavigationAction =>
-  StackActions.push({ routeName: 'users' });
+export const navigateToUsersScreen = (): GenericNavigationAction => StackActions.push('users');
```
gnprice, line 38: They do note (as they do on their "navigating without the navigation
    prop" doc [3]) that it's best to use the `navigation` object when
    one can:
    
    """
    It's highly recommended to use the methods on the navigation object
    instead of using action creators and `dispatch`. It should only be
    used for advanced use cases.
    """

Interesting. Seems like migrating to that would be a good followup commit, then (not necessarily within this PR.)

I think probably the bulk of these `navActions` actions are only ever invoked from within a React component. Perhaps most of them even from a React component that already has a `navigation` prop, and there's `useNavigation` for a convenient way to handle some of the others. So that might let us eliminate most of these.

And when we do, it looks like we'd get proper type-checking! When I try `this.props.navigation.navigate()` in a random component and start entering arguments, it looks like I get Flow telling me whether I have a proper route name and then whether I have the right arguments for that route. That's pretty nifty.

src/chat/ChatScreen.js
```diff
@@ -48,9 +43,10 @@ const componentStyles = createStyleSheet({
  * whole process.
  */
 const useFetchMessages = args => {
-  const { isFocused, narrow } = args;
+  const { narrow } = args;
 
   const dispatch = useDispatch();
+  const isFocused = useIsFocused();
```
gnprice, line 50: src/chat/ChatScreen.js | 25 ++++++-------------------
    1 file changed, 6 insertions(+), 19 deletions(-)

Very nice!

src/nav/AppNavigator.js
```diff
@@ -80,66 +84,98 @@ export type AppNavigatorParamList = {|
...
+export type AppNavigationRouteProp<
+  RouteName: $Keys<AppNavigatorParamList> = $Keys<AppNavigatorParamList>,
+> = RouteProp<GlobalParamList, RouteName>;
+
+const Stack = createStackNavigator<GlobalParamList, AppNavigatorParamList, AppNavigationProp<>>();
+
+type SelectorProps = $ReadOnly<{|
+  hasAuth: boolean,
+  accounts: Account[],
+  haveServerData: boolean,
```
gnprice, line 98: I think this means that the caller, ZulipNavigationContainer, can stop getting these three from Redux. It'd be good to pair the deletion with the addition, so that the diff is visibly just moving the logic.

I think this is also logically more or less independent of the change in the nav API, so it'd simplify things to make it a separate diff. (This commit in particular has a large quantity of boring mechanical stuff happening, so it'd be good to keep more-interesting changes out of it.) I think it could pretty well happen either before or after this one -- either `createAppNavigator` starts taking these instead of its old arguments and calling `getInitialRouteInfo` itself, or `AppNavigator` at first still takes the old arguments and later starts connecting to Redux for itself.

gnprice (commented):
Thanks @chrisbobbe ! There was a lot of complexity that had to be handled in this project -- glad to see it nearing completion.

Generally this looks great. Some comments below.

src/nav/NavigationService.js
```diff
@@ -1,32 +1,25 @@
...
+    throw new Error('Tried to use NavigationService before `NavigationContainer` was ready.');
+  }
+  return navigationContainerRef.current;
 };
 
-export const getState = (): NavigationState =>
-  // https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/decouple.20nav.20from.20redux.20%28.23M3804%29/near/1056167
-  // $FlowFixMe
-  getContainer().state.nav;
+export const getState = (): PossiblyStaleNavigationState => getContainer().getRootState();
```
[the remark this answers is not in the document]
chrisbobbe: Yikes, yeah—responding [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5.3A.20.60PossiblyStaleNavigationState.60/near/1096924).

src/styles/tabs.js
```diff
@@ -57,7 +53,10 @@ export const materialTopTabNavigatorConfig: Props => MaterialTopTabNavigatorConf
       indicatorStyle: {
         backgroundColor: BRAND_COLOR,
       },
-      upperCaseLabel: false,
+      // TODO: `upperCaseLabel` vanished in react-navigation v5; we
+      // used to use `false` for this, but it appears that the
+      // effective default value is `true`:
+      // https://github.com/react-navigation/react-navigation/issues/7952
```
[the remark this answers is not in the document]
chrisbobbe: Ah OK, I've figured out why this doesn't have a visible effect.

It _would_ have had a visible effect if we had used a string instead of a React element for `tabBarLabel`, or if we hadn't specified `tabBarLabel` and defined [`title`](https://reactnavigation.org/docs/material-top-tab-navigator/#title).

By passing a React element, we get to say exactly how the tab bar label should look, and that includes case.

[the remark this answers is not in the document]
chrisbobbe: That might mean I can delete this TODO, but I have a feeling we'll run up against it again if we ever stop using our strategy here, or we make a material top tab nav somewhere else that doesn't use that strategy.

src/nav/navActions.js
chrisbobbe: > And when we do, it looks like we'd get proper type-checking! When I try `this.props.navigation.navigate()` in a random component and start entering arguments, it looks like I get Flow telling me whether I have a proper route name and then whether I have the right arguments for that route. That's pretty nifty.

Nice! Happy to do that as a followup.

src/nav/AppNavigator.js
chrisbobbe: > or `AppNavigator` at first still takes the old arguments and later starts connecting to Redux for itself.

OK! I've taken this approach, with a new commit at the tip.

chrisbobbe:
I've just pushed a revision—but I just remembered that it doesn't address the results of the discussion about `PossiblyStaleNavigationState` (from your question at https://github.com/zulip/zulip-mobile/pull/4393#discussion_r555408509). That'll come soon.

chrisbobbe:
> I've just pushed a revision—but I just remembered that it doesn't address the results of the discussion about `PossiblyStaleNavigationState` (from your question at [#4393 (comment)](https://github.com/zulip/zulip-mobile/pull/4393#discussion_r555408509)). That'll come soon; perhaps this evening but more likely tomorrow.

OK, revision pushed!

flow-typed/@react-navigation/bottom-tabs_v5.x.x.js
```diff
@@ -245,9 +239,7 @@ declare module '@react-navigation/bottom-tabs' {
     label?: string,
     ...
   }>;
-  declare type AccessibilityActionEvent = SyntheticEvent<
-    $ReadOnly<{actionName: string, ...}>,
-  >;
+  declare type AccessibilityActionEvent = SyntheticEvent<$ReadOnly<{ actionName: string, ... }>>;
```
gnprice, line 242: react-navigation libdef [nfc]: Format files.

I think it'd be better to leave them in the format we found them in. The reason is that ideally we'd be contributing these fixes back upstream to the flow-typed repo, and preserving the format will help keep it smooth for us to do so.

So probably that means telling Prettier and ESLint to ignore these files.

src/styles/tabs.js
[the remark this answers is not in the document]
gnprice: Hmm interesting!

Yeah, I think that means the TODO still belongs here. It'd be good to also add a comment with that information you just learned about when it would start affecting things.

gnprice:
Thanks for the revision! All looks great except the two comments just above (one of them quite small.)

src/styles/tabs.js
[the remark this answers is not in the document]
chrisbobbe: Oh, hmm. With these findings in mind, I've just found a place where I'd expect to see a change: The bottom tab nav in `MainTabs` shows tab-bar labels, which we define with strings like "Home", "Streams", etc., if `!!Platform.isPad`. On an iPad simulator, I do indeed see those labels—but I don't see them in all caps.

I suspect the all-caps quirk applies to the "material top tab" navigator, but doesn't apply to the "bottom" tab navigator. I wouldn't be totally surprised to see differences spring up between them; the React Nav v5 upgrade took us from depending on `react-navigation-tabs` to `@react-navigation/material-top-tabs` and `@react-navigation/bottom-tabs`.

chrisbobbe:
Thanks for the review! Revision pushed. I posted an additional finding, about the `upperCaseLabel` thing, [above](https://github.com/zulip/zulip-mobile/pull/4393#discussion_r556920477); unfortunately it complicates the story a bit.

src/styles/tabs.js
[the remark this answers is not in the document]
gnprice: Yeah, I think that's an intentional design choice.

Or more specifically, it's a faithful implementation (at least on that point) of Material Design, which those top tabs are advertised as being for. Material calls those top tabs simply "tabs", and they are in caps:
https://material.io/components/tabs
image

For our bottom tabs, Material's term is "bottom navigation":
https://material.io/components/bottom-navigation
and the titles are not in caps.

Meanwhile, Apple says "tab bar" for something very much like our bottom tabs, or Material's bottom navigation:
https://developer.apple.com/design/human-interface-guidelines/ios/bars/tab-bars/
and those also aren't in caps. I'm not sure there's any equivalent to top tabs, or Material tabs, in Apple's iOS guidelines.

Anyway, I think that means that we shouldn't expect the upper-case thing to just randomly change on us in the future due to something intended as a bugfix.

.prettierignore
```diff
@@ -2,4 +2,9 @@
 src/emoji/__tests__/data-test.js
 
 # We're not allowing Prettier to touch some of our vendored code.
-src/third/redux-persist
\ No newline at end of file
+src/third/redux-persist
+
+# We'd like to keep the original formatting for some Flow libdefs from
+# FlowTyped, to make it easier to eventually make a PR back to
+# FlowTyped.
```
gnprice, line 9: ESLint already ignores them because of the `**/__flow-tests__/**`
    line in `.eslintignore`.

nit: I think you mean the other line, `**/flow-typed/**`

.prettierignore
```diff
@@ -2,4 +2,9 @@
...
+flow-typed/@react-navigation
```
gnprice, line 10: nit: last line should end with a newline, like other lines

(Git will display this like so:
```diff
+flow-typed/@react-navigation
\ No newline at end of file
```
)

gnprice (commented):
Thanks for the revision! Looks great -- two nits below, and I replied about upper-case in tab labels above: https://github.com/zulip/zulip-mobile/pull/4393#discussion_r556949209 Please merge at will.

.prettierignore
chrisbobbe: Ah, yeah, thanks for the catch. This is one of those things that I know we're supposed to do, but that I don't usually think about—probably because our auto-formatting takes care of it in most files.

.prettierignore
chrisbobbe: Indeed, thanks.

Issue #4297, opened 2020-11-03
Fix build on Windows, at webview rsync step


gnprice:
The build is currently failing on Windows, inside tools/build-webview. The error is from rsync, which apparently is taking the absolute paths we give it to copy from and to (like `/d/zulip-mobile/…`) and interpreting them somehow as remote locations, where the leading `/d/` means to go to a remote host called `d`. [See chat discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/rsync.20windows.20issue/near/1043864), including diagnosis and workaround.

Here's that workaround as a branch:
https://github.com/zulip/zulip-mobile/compare/master...gnprice:dev-rsync-relative
The solution is to use relative paths. We already take care to *pass* relative paths to this script -- see its caller in `android/` -- but we absolutize them inside the script. So just the script itself needs to be edited to not do that.

That workaround branch does that, but has some collateral damage. To solve this issue, we'll want to make similar changes to those in the branch, but:
* There are some checks on the paths, where we use absolute paths. The branch just deletes those checks, but let's keep them. To do that, we'll want to make absolute paths, but just keep them in their own separate variables.
* The branch assumes the current directory when running the script is the root of the zulip-mobile repo. That happens to be true when it's invoked from our build process, but the script should avoid depending on that.
  * The script should `cd` to that root directory, at the same point where the current code determines it and stores it in the variable `$root`.
  * But: if the current directory starts out somewhere else, the `dest` path argument passed to the script should be interpreted according to the original current directory. That's normal behavior for any command-line interface.

tusharojha:
I am also facing the same issue. Also, I am struggling with installing rsync, please help.

gnprice:
Hi @tusharojha ! Glad to help out with the setup steps. Let's use [the chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Unable.20to.20build.20and.20run.20android.20app/near/1075786) for that.

palakyadav1807:
Is this issue resolved? I am facing same problem , what steps should I follow?

gnprice:
@palakyadav1807 This issue is now fixed! If you update to master and just use our normal setup instructions, it should now work.

If you find you're still running into any trouble, please go ahead and ask on #mobile on chat.zulip.org, and we'll debug.

PR #4395, opened 2021-01-11
tools/build-webview: Work in terms of relative paths.

gnprice:
On Windows (at least in Git Bash, which is our recommended way to set up the dev environment on Windows), the absolute paths end up looking like `/d/zulip-mobile/…`.  When one of those gets passed to rsync, it somehow decides to interpret it as a remote location, where the leading `/d/` means to go to a remote host called `d`.  Naturally, that doesn't work out.

To solve the problem, use relative paths.

This is made slightly tricky by the fact that the two different arguments we pass to rsync come to us in the form of paths relative to two different locations: the sources we want to copy are at particular places relative to the source tree, while the destination directory comes from the command line and is therefore relative to the current directory the script was started with.  So we can't pass them both directly to rsync without doing some kind of normalization on them first: if we're not at the source tree's root, then a source path like `src/webview/static` won't be right, but if we cd to the root then the destination path from the command line won't be right.  That's the problem we'd previously cut right through by normalizing everything into absolute paths.

Happily, GNU's coreutils offers not only `readlink` but `realpath`, which includes a `--relative-to` option.  So we can use that to turn `dest` relative to the starting directory into a `dest` that's still relative, but now relative to the source root, just like the source paths are.

(We already require GNU coreutils; that's the `ensure-coreutils.sh` script invoked near the top of this one.)

Also add a cleaned-up, optional version of the verbose logging that was helpful in debugging this.

Fixes: #4297

### DIFF
--- a/tools/build-webview
+++ b/tools/build-webview
@@ -36,6 +36,15 @@ err() {
     exit 1;
 }
 
+# If verbose, format and log message.  First arg is format string.
+verbose_printf() {
+    if ! (( verbose )); then
+        return
+    fi
+    # shellcheck disable=SC2059 # format string from parameter
+    printf >&2 "%s: $1\n" "$0" "${@:2}"
+}
+
 ################################################################################
 # Environment
 ################################################################################
@@ -63,6 +72,7 @@ readonly root
 unset target
 unset dest
 check_path_name=0
+verbose=0
 while (( $# )); do
     case "$1" in
         --help|-h|help)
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Great, thanks! Issues with development on Windows aren't high-visibility for me (I've been using my Mac only), so I'm quite glad for your debugging and fixing this. 🙂 

I don't have a Windows computer to test the fix, but I can confirm that

- the changes don't add more problems that `shellcheck` flags
- I can still build/run the app for iOS and Android on my Mac

So, merging!

gnprice:
Thanks for the review! And the testing on macOS -- I did think through that it should be fine (because we already require coreutils in this script), but had not actually tested that myself :slightly_smiling_face:

PR #4400, opened 2021-01-12
unreads: Prepare for converting to efficient data structures

gnprice:
Currently the app can be very sluggish in responding to new data: I [measured it](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data-structure.20performance/near/1068970) a few weeks ago at several hundred ms on any of a new message; or marking messages as read; or handling a batch of fetched messages. And that was on a Pixel 2 XL, which is a fairly recent (2017) flagship device, so it'd be slower on many others. I was viewing the app's home screen (so there wasn't even a message list involved), with a test user on czo who -- like many real users -- has a lot of unreads.

Then [breaking it down](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data-structure.20performance/near/1069171), a large chunk of that time (for some of those events) is going into updating our data structures, and a larger chunk (or almost all, for some events) is going into updating React components. A root cause of both of these ([including the React work](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data-structure.20performance/near/1069202)) is that many of our data structures are inefficient. We've recently started moving our central data structures to Immutable.js, as in #4201 for `state.narrows`, #4392 for `state.pmConversations`, and #4390 (open) for `state.messages`.

Another of our inefficient data structures, one particularly implicated in those slow React updates, is `state.unreads`. I have a draft of a branch, begun Friday, to convert `state.unreads.streams` to something efficient, as a start and as the part of `state.unreads` that's most likely to be large for many users. The actual conversion still needs a bit more work, but here's a series of prep commits.

The largest part of the change is that we take the tests which had been meddling in the details of the internal representation of `state.unreads.streams`, and convert them to a form where they instead (a) build up the test states by using the actual reducer, and (b) inspect the resulting states by using selectors, or a helper which would be a reasonable selector. This continues in the direction of organizing reducers and selectors into encapsulated "models", discussed last week in 7c697fb74d312975a9df649d088f4ae712a7a381 (#4392): the tests test the model as a whole, not its internal pieces.

Converting the tests to this whole-model style makes them a lot shorter and clearer. It also saves us from going and fiddling with all the test cases each time we change the internals of that data structure -- the draft of the actual conversion commit doesn't touch the tests at all, except updating one selector-like helper function. And it keeps the tests in place so they can check all the same facts about the new implementation as they do now about the old one... which is good because yesterday they already caught several bugs in my initial draft :wink: .


src/unread/__tests__/unreadSelectors-test.js
```diff
@@ -34,22 +34,22 @@ const unreadStreamData = [
...
 const unreadHuddlesData = [
   {
     user_ids_string: '1,2,3',
-    unread_message_ids: [1, 2],
+    unread_message_ids: [21, 22],
   },
   {
-    user_ids_string: '4,5',
-    unread_message_ids: [3, 4, 5],
+    user_ids_string: '1,4,5',
```
chrisbobbe: `unread tests: Fix some impossible data.`

Hmm, I don't yet follow why `'4,5'` (changed to `'1,4,5'`) is impossible for `user_ids_string` here.

src/unread/__tests__/unreadSelectors-test.js
```diff
@@ -83,9 +48,7 @@ describe('getUnreadByStream', () => {
           in_home_view: true,
         },
       ],
-      unread: {
-        streams: unreadStreamData,
-      },
+      unread: unreadState,
```
chrisbobbe, line 51: Separate from this refactor, I think there's something a bit wrong with the description for this test: 'when there are unread stream messages, returns a list with counts per stream_id '.

The thing returned isn't a list, it's an object-as-map.

src/unread/__tests__/unreadSelectors-test.js
```diff
@@ -13,55 +14,19 @@ import {
...
-    user_ids_string: '1,4,5',
-    unread_message_ids: [23, 24, 25],
-  },
-];
-
-const unreadMentionsData = [1, 2, 3];
+import * as eg from '../../__tests__/lib/exampleData';
+import {
+  initialState,
+  initialState as initialUnreadState,
```
chrisbobbe, line 20: Hmm, what's the reason for importing `initialState` twice (once retaining the `initialState` name, once giving it the `initialUnreadState` name)?

src/topics/__tests__/topicsSelectors-test.js
```diff
@@ -112,21 +106,16 @@ describe('getTopicsForStream', () => {
...
-      },
+      unread: [
+        eg.streamMessage({ stream_id: 1, subject: 'topic 2', id: 1 }),
+        eg.streamMessage({ stream_id: 1, subject: 'topic 2', id: 5 }),
+        eg.streamMessage({ stream_id: 1, subject: 'topic 2', id: 6 }),
+        eg.streamMessage({ stream_id: 1, subject: 'topic 4', id: 7 }),
+        eg.streamMessage({ stream_id: 1, subject: 'topic 4', id: 8 }),
+      ].reduce(
+        (st, message) => unreadReducer(st, mkMessageAction(message)),
+        eg.baseReduxState.unread,
```
chrisbobbe, line 117: Would the `initialState` export from unread-testlib.js do just as well as `eg.baseReduxState.unread`, or should we stick with the latter?

chrisbobbe (commented):
Thanks! I continue to be excited for this new way of organizing our selectors and reducers. Just a few comments, below; otherwise, this LGTM.

src/unread/__tests__/unreadSelectors-test.js
gnprice: Ah -- it's because in this data structure, the key for a group PM thread includes all users including self. So in particular there's some user ID (namely self's) that is in every key in the map. So there can't be both `1,2,3` and `4,5` as keys.

I'll add a remark about this.

src/unread/__tests__/unreadSelectors-test.js
gnprice: Hmm indeed. I'd guess that the description was true when originally written, and then the data structure was improved :slightly_smiling_face: ... huh, nope, `git log -S` on that description finds 81a481e76 where it was already wrong in just this way.

Anyway, I'll fix it.

src/unread/__tests__/unreadSelectors-test.js
gnprice: Hmm, good catch. That is probably a consequence of letting the IDE add the import :wink: I'll reduce to just `initialUnreadState`.

src/unread/__tests__/unreadSelectors-test.js
chrisbobbe: Ah OK, got it, thanks! I thought it might have something to do with the self-user, but I couldn't find anything labelling `1` as the self-user's user ID.

It makes sense though now you put it that way; it's impossible to say that `1,2,3` and `4,5` are both valid keys under the constraint that each key must include the self-user, but it _is_ possible for `1,2,3 and 1,4,5` to both be valid keys given that it's possible for `1` to be the self-user's ID.

src/unread/__tests__/unreadSelectors-test.js
gnprice: ... Or maybe just `initialState`. This is after all the *unread* selector tests... plus that's the name used in `unreadModel-test.js`, and using the same name in both will simplify unifying those in the future.

(The main work to do to unify them is to make all these `unreadSelectors` tests well-typed. I started on that right after converting the reducer tests, then realized it wasn't actually needed for what this branch is doing, and paused that in favor of pursuing the rest of the branch.)

src/topics/__tests__/topicsSelectors-test.js
gnprice: Ah, yes. I think it was as I was fixing up this test file (after writing most of the other changes) that I realized that `eg.baseReduxState.unread` should be exactly equivalent to that `initialState` export -- after all, the point of the way we construct that `initialState`, with `{ type: eg.randString() }`, is that it exercises exactly the same path through the reducer that Redux uses to initialize the state. (And `eg.baseReduxState` comes via Redux's `createStore`.)

I'm not entirely sure what the most helpful style is for referring to that state.
* Here, `eg.baseReduxState.unread` feels natural because it fits in with the rest of this global state we're assembling.
* In the unread-model tests (including `unreadSelectors-test.js`), `initialState` or `initialUnreadState` feels natural because it makes it clear that it's local -- these are the tests of this very code, so they shouldn't need to appeal to some central library to get an instance of this state.

I'm inclined to leave the style different between these places at least for now, and perhaps in the future we'll settle on one of them.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -0,0 +1,188 @@
...
+        allMessages: {},
+        all,
+        messages,
+        flag,
+        operation,
+      };
+    };
+
+    const baseState = (() => {
+      const streamAction = args => mkMessageAction(eg.streamMessage(args));
```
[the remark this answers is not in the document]
gnprice: It could, but I'm inclined to keep the name short: this helper is local to a quite short stretch of code, so it's always easy to go look at its definition. And I like that the `state = reducer(state, …);` statements fit on one line each, and those lines are already kind of long.

If this were a more global name (like if it were exported, or even exposed to this whole test file), then I'd definitely want to give it a more explicit name like `streamMessageAction`.

src/topics/__tests__/topicsSelectors-test.js
chrisbobbe: > I'm inclined to leave the style different between these places at least for now, and perhaps in the future we'll settle on one of them.

Sounds good; I don't have strong feelings either way.

gnprice:
New revision pushed! Take a look.

chrisbobbe:
Great, thanks! Merged.

gnprice:
I've just filed #4438 to give a name to the issue that this work was in preparation for fixing.

chrisbobbe:
> I've just filed #4438 to give a name to the issue that this work was in preparation for fixing.

Thanks!

Issue #4401, opened 2021-01-13
Regression: Some avatars (those stored on s3 backend) don't show up.


chrisbobbe:
See [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/302.20redirect.20for.20avatar/near/1097413); filing this issue before making a PR so that we have something to refer to if we get reports, as we've recently made a (beta) release.

The symptom is that some people's avatars don't show up. This bug is only observable on realms that use the s3 backend for avatar storage, e.g., realms on `zulipchat.com`.

PR #4402, opened 2021-01-13
UserAvatar: Don't send auth headers to s3 backend.

chrisbobbe:
See discussion at
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/302.20redirect.20for.20avatar/near/1097413.

In #4367, I'd assumed that it would be fine to include the auth
headers even when they're unnecessary. That turns out not to be
true: the s3 backend doesn't want them.

Fixes: #4401


gnprice:
Merged! Thanks for spotting this issue and for the fix.

PR #4404, opened 2021-01-13
user: Convert to UserItemById from whole-User-taking UserItem

gnprice:
This is the small series of followup conversions I mentioned at #4364. Not on a critical path to our major data-refactoring priorities, but a small nice cleanup.

The branch doesn't end up eliminating the old UserItem. (Or making private, really -- the code wasn't going to be deleted in any case, because it does most of the actual work of the new UserItem, which just wraps it with `connect`.) That's because it turns out that in one place, the autocomplete suggestions, we actually make use of the ability to supply fake synthetic users that don't exist in the users data.

That's likely not the best way that could work, at either a UI level or a code level. But doing anything about that would be a bigger project than this series otherwise is, and doesn't seem like it has a big payoff compared to the priorities we're working on, so we leave it be for now, and just rename and re-doc things to isolate it to that one spot.

### DIFF
--- a/src/reactions/ReactionUserList.js
+++ b/src/reactions/ReactionUserList.js
@@ -5,7 +5,7 @@ import { connect } from '../react-redux';
 
 import * as NavigationService from '../nav/NavigationService';
 import type { Dispatch, UserOrBot } from '../types';
-import { UserItemById } from '../users/UserItem';
+import UserItem from '../users/UserItem';
 import { navigateToAccountDetails } from '../actions';
 
 type Props = $ReadOnly<{|
@@ -30,9 +30,7 @@ class ReactionUserList extends PureComponent<Props> {
       <FlatList
         data={reactedUserIds}
         keyExtractor={userId => `${userId}`}
-        renderItem={({ item }) => (
-          <UserItemById key={item} userId={item} onPress={this.handlePress} />
-        )}
+        renderItem={({ item }) => <UserItem key={item} userId={item} onPress={this.handlePress} />}
       />
     );
   }
--- a/src/reactions/ReactionUserList.js
+++ b/src/reactions/ReactionUserList.js
@@ -5,13 +5,12 @@ import { connect } from '../react-redux';
 
 import * as NavigationService from '../nav/NavigationService';
 import type { Dispatch, UserOrBot } from '../types';
-import UserItem from '../users/UserItem';
+import { UserItemById } from '../users/UserItem';
 import { navigateToAccountDetails } from '../actions';
 
 type Props = $ReadOnly<{|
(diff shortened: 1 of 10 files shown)


src/reactions/ReactionUserList.js
```diff
@@ -25,19 +24,15 @@ class ReactionUserList extends PureComponent<Props> {
...
         keyExtractor={userId => `${userId}`}
-        renderItem={({ item }) => {
-          const user = allUsersById.get(item);
-          if (!user) {
-            return null;
-          }
-          return <UserItem key={user.user_id} user={user} onPress={this.handlePress} />;
-        }}
+        renderItem={({ item }) => (
+          <UserItemById key={item} user={item} onPress={this.handlePress} />
```
chrisbobbe: 17bf4a53fe546ea3e9fcaadfb7f1a9ca76f312a7  user: Use UserItemById in ReactionUserList; simplify whole call stack.

Hmm, I'm seeing that `UserItemById`'s props aren't getting type-checked; do you see that too? I notice that the `user` prop given here is a `number`, and that the `userId` prop isn't given. I also notice that the callsite in `GroupDetailsScreen` does pass a `number` for `userId` (that sounds right to me), and the callsite in `PmConversationList` gives a `number` for `user` and doesn't give `userId`.

I believe the intended interface is that callers pass `userId` (a `number`) and don't pass `user` because that'll be supplied (as a `UserOrBot`) by `connect`.

I also find that I can give, e.g.,`foo="nonsense"` in one of `UserItemById`'s callsites without getting Flow errors.

chrisbobbe (commented):
Thanks! One comment below.

src/reactions/ReactionUserList.js
gnprice: Hmm! Thanks for catching this. Will investigate.

gnprice:
New revision pushed!

src/reactions/ReactionUserList.js
gnprice: We discussed this [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20and.20.60connect.60/near/1098203) and found a satisfying workaround. It seems to be a Flow bug -- likely the same underlying Flow bug that was behind #3768.

chrisbobbe:
Great, thanks! I've just rebased across the changes in #4393, which was just merged, and I pushed the result to this PR branch. There were a few moderately sized conflicts, so I wonder if you'd like to take a quick look and spot any rebase errors I might have made? I don't expect any.

Otherwise, please merge at will!

gnprice:
The rebase looks good, thanks! Merged.

PR #4406, opened 2021-01-13
Clarify some message-list code (part 1).

chrisbobbe:
A (partial) resumption of #3913, which has grown stale. There are various cleanups I'd like to do before fixing #3851 with a diffing algorithm, and this is a chunk of those.

In a PR after this one, I'd like to flatten the return value of `renderMessages` (renamed `getHtmlPieceDescriptors` in this series), so that it's something like `Array<TimePieceDescriptor | HeaderPieceDescriptor | MessagePieceDescriptor>`. The current nesting of the "time" and "message" pieces under `HtmlPieceDescriptor.data` is an unnecessary complication, since `renderMessagesAsHtml` (renamed `contentHtmlFromPieceDescriptors` in this series) just goes and undoes that nesting.


chrisbobbe:
Hmm, seeing some CI failures. Will investigate.

src/webview/MessageList.js
```diff
@@ -149,7 +149,7 @@ class MessageList extends Component<Props> {
   webview: ?WebView;
   readyRetryInterval: IntervalID | void;
   sendUpdateEventsIsReady: boolean;
-  unsentUpdateEvents: WebViewUpdateEvent[] = [];
+  unsentUpdateEvents: WebViewInboundEvent[] = [];
```
gnprice, line 152: Grepping for `pdateEvent` (or `updateevent` case-insensitive), there's still a few identifiers here and also in js.js which use that terminology. Those would be good to update too, to match. Maybe cleanest to do in a second commit right after this one.

gnprice (commented):
This looks great, thanks! Glad to be getting some of this tangly logic cleared up.

One comment below, which is almost just a followup, but ideally would really happen within this branch, in a commit right after the one it's about.


> contentHtmlFromPieceDescriptors

This name feels a little awkwardly long... but that's a lot better than being confusing, which the old names (where the `renderedMessages` got "rendered" again) certainly were. And I don't have a shorter suggestion that's equally clear.


> In a PR after this one, I'd like to flatten the return value of `renderMessages` (renamed `getHtmlPieceDescriptors` in this series), so that it's something like `Array<TimePieceDescriptor | HeaderPieceDescriptor | MessagePieceDescriptor>`. The current nesting of the "time" and "message" pieces under `HtmlPieceDescriptor.data` is an unnecessary complication, since `renderMessagesAsHtml` (renamed `contentHtmlFromPieceDescriptors` in this series) just goes and undoes that nesting.

I agree -- sounds great.

chrisbobbe:
> > contentHtmlFromPieceDescriptors
> 
> This name feels a little awkwardly long... but that's a lot better than being confusing, which the old names (where the `renderedMessages` got "rendered" again) certainly were. And I don't have a shorter suggestion that's equally clear.

Yeah. We can always change it if we think of something better. 🙂 

gnprice:
Merged, thanks!

Issue #4405, opened 2021-01-13
AuthScreen.js UI : should use Scrollable View


abhi0504:
## Issue description
While working on a issue , I noticed that `AuthScreen.js`  UI is not looking good in landscape Orientation 

## How to Reproduce 

Step 1 : Log out from your account and then login again 
Step 2 : Now you are on login screen(AuthScreen.js) and then rotate your phone in landscape orientation 

And no you can see the below screen shot 

![Android Emulator - Pixel_3_API_29_5554 13-01-2021 12_25_12](https://user-images.githubusercontent.com/56453541/104416833-761f0100-559a-11eb-8ca2-5fa3d9025795.png)

Here you can clearly see that the "Zulip" logo is not looking that good also the last option that is "Sign in with Apple" is also not gettin proper space .

## Fix 
This can be fixed by using a `ScrollView` instead of `View` or `Screen`

abhi0504:
I am working on this issue @gnprice @chrisbobbe plz comment and give your suggestions

PR #4408, opened 2021-01-14
Using ScrollView instead of View

abhi0504:
In `AuthScreen.js` when we change our orientation from
    portrait to landscape then the screen UI not looks good
    as the logo of zulip and the lower sign in button not fits
    in the screen so we should use a `ScrollView`

https://user-images.githubusercontent.com/56453541/104576765-a213b280-567e-11eb-8a7a-b21fb8f46cf0.mp4

 
 Fixes: #4405

### DIFF
--- a/src/common/Screen.js
+++ b/src/common/Screen.js
@@ -2,7 +2,7 @@
 
 import React, { PureComponent } from 'react';
 import type { Node as React$Node } from 'react';
-import { View, ScrollView } from 'react-native';
+import { ScrollView } from 'react-native';
 import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, @abhi0504! In particular, thanks for noticing this annoying bug. 🙂 

I'd rather not add another `ScrollView` here; we already have some subtle logic around what code exactly is responsible for handling scrolling, with `Screen`'s `scrollEnabled` prop, etc.

In fact, I expect this problem would be solved by something we discussed on CZO today, [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Screen.20component.20in.20landscape.20mode/near/1115698). That solution is open in #4471, so I'll close this.

Issue #3918, opened 2020-02-20
Convert from names to stream IDs to identify streams


gnprice:
This is a meta-issue which covers a large number of independent tasks. Filing it in order to have a place to point to refer to the overall effort.

This is parallel to #3764, but for streams. Just like users can change their email addresses, streams can have their names changed, and so the permanent, numeric stream IDs are a better way to refer to a stream.

Most of the discussion in #3764 applies to this as well.

Tasks for specific instances of this include #3244, and #2760 which we fixed last summer.

chrisbobbe:
> Tasks for specific instances of this include #3244, and #2760 which we fixed last summer.

Do I understand correctly that #3244 is the last remaining instance of this? #4635 has a commit that's marked as fixing that.

gnprice:
> > Tasks for specific instances of this include #3244, and #2760 which we fixed last summer.
> 
> Do I understand correctly that #3244 is the last remaining instance of this? #4635 has a commit that's marked as fixing that.

No, unfortunately -- those were just the instances that had specific tasks at the time.

#4333 is the biggest remaining instance of this -- we identify narrows by stream name rather than stream ID. It connects to other instances which may not have specific tasks in our tracker.

#3998 will help, giving us stream IDs on Outbox values. That will let various code that consumes those -- especially all the code that consumes `Message | Outbox` -- stop using names and use the IDs instead.

gnprice:
The victory condition for this issue is basically: if you try removing the `name` property from the definitions of the `Stream` and `Subscription` types, Flow should find only a handful of errors (i.e. places we're still using the property), all of which should be purely legitimate uses. (Which basically means just showing the name to the user in the UI.)

One quick approximation to that is to grep for `streamName` -- I've tried to shift to that name for most locals and function parameters that refer to a stream name. If you try that now, it finds 131 matches (or 103 with `-w`). The bulk of those are uses that ought to be using stream IDs. I believe most of them at this point are part of #4333 -- they're using the stream name to identify a stream or topic narrow.

So I think at this point the sequence for resolving this issue will be:
* Finish #3998. With #5000 merged, the next step there is to wait for a few weeks after a release with #5000 is out; then we can merge a migration that finishes #3998.
  * Then do any easy followups on that.
* Then do #4333. This may be fairly complex. We did the PMs side of this in the series of PRs leading up to #4382.
* Then mop up any remaining instances of this issue. I think at that point these will be small, if there are any.

gnprice:
[...]

PR #4413, opened 2021-01-14
internal_link: allow streamID without stream name

### LINKED ISSUE #4185: Allow internal links to use stream ID without stream name
This is the mobile counterpart to zulip/zulip#15402:

> Apparently we don't support using links of this form to access to a topic:
> 
> https://chat.zulip.org/#narrow/stream/3/topic/send_message.20API
> 
> requiring the stream name (or anything) to be encoded after a dash, like here:
> 
> https://chat.zulip.org/#narrow/stream/3-backend/topic/send_message.20API

The relevant code in our codebase is here, in src/utils/internalLinks.js:
```js
/** Parse the operand of a `stream` operator, returning a stream name. */
const parseStreamOperand = (operand, streamsById): string => {
  // "New" (2018) format: ${stream_id}-${stream_name} .
  const match = /^(\d+)-/.exec(operand);
  if (match) {
    const stream = streamsById.get(parseInt(match[0], 10));
    if (stream) {
      return stream.name;
    }
  }

  // Old format: just stream name.  This case is relevant indefinitely,
  // so that links in old conversations continue to work.
  return decodeHashComponent(operand);
};
```

The webapp now accepts the ID-only form, and we should too. (It still won't *generate* links of that form -- but bots may start doing so.)

AdityaYadavalli1:
This commit allows for the internal links to be of the form:
https://chat.zulip.org/#narrow/stream/3/topic/send_message.20API. Previously, only internal links of the form:
https://chat.zulip.org/#narrow/stream/3-backend/topic/send_message.20API were supported. 

Fixes #4185

### DIFF
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -288,7 +288,6 @@ describe('getNarrowFromLink', () => {
 
     test('on malformed stream link: treat as old format', () => {
       const expectAsName = name => expectStream(name, [streamGeneral], name);
-      expectAsName(`${streamGeneral.stream_id}`);
       expectAsName(`-${streamGeneral.stream_id}`);
       expectAsName(`${streamGeneral.stream_id}nonsense-general`);
     });
@@ -310,7 +309,7 @@ describe('getNarrowFromLink', () => {
 
       test('on ambiguous new- or old-style: new wins', () => {
         const collider = { ...eg.makeStream({ name: 'collider' }), stream_id: 311 };
-        expectStream('311', [numbers, collider], '311'); // malformed for new-style
+        expectStream('311', [numbers, collider], '311'); // will soon recognize this as the new style
         expectStream('311-', [numbersHyphen, collider], 'collider');
         expectStream('311-help', [numbersPlus, collider], 'collider');
       });
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -289,9 +289,7 @@ describe('getNarrowFromLink', () => {
     test('on malformed stream link: treat as old format', () => {
       const expectAsName = name => expectStream(name, [streamGeneral], name);
       expectAsName(`-${streamGeneral.stream_id}`);
-      // This case will come back when we take the fix in
-      //   zulip/zulip@d67cc2d08, soon.
-      // expectAsName(`${streamGeneral.stream_id}nonsense-general`);
+      expectAsName(`${streamGeneral.stream_id}nonsense-general`);
     });
 
     {
--- a/src/utils/__tests__/internalLinks-test.js
+++ b/src/utils/__tests__/internalLinks-test.js
@@ -289,7 +289,9 @@ describe('getNarrowFromLink', () => {
     test('on malformed stream link: treat as old format', () => {
       const expectAsName = name => expectStream(name, [streamGeneral], name);
       expectAsName(`-${streamGeneral.stream_id}`);
-      expectAsName(`${streamGeneral.stream_id}nonsense-general`);
+      // This case will come back when we take the fix in
+      //   zulip/zulip@d67cc2d08, soon.
+      // expectAsName(`${streamGeneral.stream_id}nonsense-general`);
     });
 
     {
@@ -309,7 +311,7 @@ describe('getNarrowFromLink', () => {
 
       test('on ambiguous new- or old-style: new wins', () => {
         const collider = { ...eg.makeStream({ name: 'collider' }), stream_id: 311 };
-        expectStream('311', [numbers, collider], '311'); // will soon recognize this as the new style
+        expectStream('311', [numbers, collider], 'collider');
         expectStream('311-', [numbersHyphen, collider], 'collider');
         expectStream('311-help', [numbersPlus, collider], 'collider');
       });
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -121,7 +121,7 @@ export const decodeHashComponent = (string: string): string => {
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
-  const match = /^([\d]+)(?:-.*)?/.exec(operand);
+  const match = /^([\d]+)(?:-.*)?$/.exec(operand);
   if (match) {
     const stream = streamsById.get(parseInt(match[1], 10));
     if (stream) {
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -121,7 +121,7 @@ export const decodeHashComponent = (string: string): string => {
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
-  const match = /^(\d+)-/.exec(operand);
+  const match = /^([\d]+)(-.*)?/.exec(operand);
   if (match) {
     const stream = streamsById.get(parseInt(match[0], 10));
     if (stream) {
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -123,7 +123,7 @@ const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
   const match = /^([\d]+)(-.*)?/.exec(operand);
   if (match) {
-    const stream = streamsById.get(parseInt(match[0], 10));
+    const stream = streamsById.get(parseInt(match[1], 10));
     if (stream) {
       return stream.name;
     }
--- a/src/utils/internalLinks.js
+++ b/src/utils/internalLinks.js
@@ -121,7 +121,7 @@ export const decodeHashComponent = (string: string): string => {
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
-  const match = /^([\d]+)(-.*)?/.exec(operand);
+  const match = /^([\d]+)(?:-.*)?/.exec(operand);
   if (match) {
     const stream = streamsById.get(parseInt(match[1], 10));
     if (stream) {


gnprice:
Thanks @AdityaYadavalli1 !

Some discussion in [this chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/First.20Contribution/near/1098953).

src/utils/internalLinks.js
```diff
@@ -120,7 +120,7 @@ export const decodeHashComponent = (string: string): string => {
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
-  const match = /^(\d+)-/.exec(operand);
+  const match = /^(\d+)(-.*)?/.exec(operand);
```
chrisbobbe, line 123: nit: I would favor a "non-capturing group" instead of a "capturing group" (you can search [this doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges) for those terms) for the added part, since we don't need to remember the part after the stream ID. (This should remain true as we move toward identifying streams with IDs, not names; that's #3918.)

Though, I notice that the web app [uses](https://github.com/showell/zulip/commit/c84f939d828274f69d69018fa99174e1ec0f0dd3#diff-e39c2df6624a5f0a813790e84aa1a9eed567fb0e70b1fa8df99b6e40d7c6ecedR289) a capturing group, hmm. @gnprice, what do you think; should we match that, just for consistency?

src/utils/internalLinks.js
```diff
@@ -120,7 +120,7 @@ export const decodeHashComponent = (string: string): string => {
...
   if (match) {
     const stream = streamsById.get(parseInt(match[0], 10));
```
chrisbobbe, line 125: Hmm, in fact! I think something is slightly wrong with the code before the change made in this PR. In particular, `match[0]`, here, will be what gets matched by the _entire pattern_, which means it might include `-`. Or, after the change in this commit, it might include, e.g., `-asdfjkl;` (if "asdfjkl;" is the stream name). Oddly, both `parseInt('123-', 10)` and `parseInt('123-asdfjkl;', 10)` seem to give `123`—but that seems like too odd of a quirk to rely on.

What `parseInt` wants (or should want, anyway), instead of `match[0]`, is just the stream ID string, with none of the stuff after it. Handily, the parentheses around the just-stream-ID-string pattern—i.e., the `(` and `)` in `(\d+)`—make a "capturing group", which I've mentioned in a comment above.

Capturing groups will be stored in `match` in order of their appearance in the pattern, starting from index 1 (note: not index 0, which is reserved for what the entire pattern matched). 

I find https://regex101.com/ quite helpful in debugging regular expressions, just make sure to select "EcmaScript (JavaScript)" instead of the default "PCRE (PHP)". 🙂

So, to fix (unless @gnprice has a different idea): Just a simple commit, [separate](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html) from the one you've already made, with the following change:

```diff
- const stream = streamsById.get(parseInt(match[0], 10));
+ const stream = streamsById.get(parseInt(match[1], 10));
```

I see that the web app already has [`m[1]`](https://github.com/zulip/zulip/blob/08d716c/static/js/stream_data.js#L320), instead of `m[0]`, in its corresponding code.

chrisbobbe (commented):
Thanks, @AdityaYadavalli1! See a few comments, below—one about some code that looks wrong before the changes in this PR, and would be good to clean up. 🙂

src/utils/internalLinks.js
AdityaYadavalli1: Yes, I noticed it earlier too but it didn't change it because it directly impact this issue (or so I thought). I have made that change. I also changed the test cases so that it passes now. Thanks for the help!

chrisbobbe:
Thanks for the revisions, @AdityaYadavalli1!

When I run `tools/test` on each commit in the branch, I see that not all commits pass.

From your `streamID` branch (i.e., after running `git checkout streamID`, you can do `git rebase -i --exec 'tools/test'` as a shortcut for running `tools/test` on each new commit in the branch, i.e., each commit that's on `streamID` but isn't on the branch that `streamID` is set up to track.

(Normally, it's helpful to set up `streamID` to track `master` on https://github.com/zulip/zulip-mobile. E.g., if the remote pointing to https://github.com/zulip/zulip-mobile is called "origin", the default name, `streamID` would be set up to track it if you've run `git branch -u origin/master`. In my own dev environment, I've decided to use the name "upstream" instead of the default "origin", but that's up to you.)

AdityaYadavalli1:
> Thanks for the revisions, @AdityaYadavalli1!
> 
> When I run `tools/test` on each commit in the branch, I see that not all commits pass.
> 
> From your `streamID` branch (i.e., after running `git checkout streamID`, you can do `git rebase -i --exec 'tools/test'` as a shortcut for running `tools/test` on each new commit in the branch, i.e., each commit that's on `streamID` but isn't on the branch that `streamID` is set up to track.
> 
> (Normally, it's helpful to set up `streamID` to track `master` on https://github.com/zulip/zulip-mobile. E.g., if the remote pointing to https://github.com/zulip/zulip-mobile is called "origin", the default name, `streamID` would be set up to track it if you've run `git branch -u origin/master`. In my own dev environment, I've decided to use the name "upstream" instead of the default "origin", but that's up to you.)

AdityaYadavalli1:
> Thanks for the revisions, @AdityaYadavalli1!
> 
> When I run `tools/test` on each commit in the branch, I see that not all commits pass.
> 
> From your `streamID` branch (i.e., after running `git checkout streamID`, you can do `git rebase -i --exec 'tools/test'` as a shortcut for running `tools/test` on each new commit in the branch, i.e., each commit that's on `streamID` but isn't on the branch that `streamID` is set up to track.
> 
> (Normally, it's helpful to set up `streamID` to track `master` on https://github.com/zulip/zulip-mobile. E.g., if the remote pointing to https://github.com/zulip/zulip-mobile is called "origin", the default name, `streamID` would be set up to track it if you've run `git branch -u origin/master`. In my own dev environment, I've decided to use the name "upstream" instead of the default "origin", but that's up to you.)

Hi @chrisbobbe!, sorry for the late reply. Yes, you are right. Other than the last commit, all the others don't pass the tests. Would you like me to squash those commits?

AdityaYadavalli1:
> > Thanks for the revisions, @AdityaYadavalli1!
> > When I run `tools/test` on each commit in the branch, I see that not all commits pass.
> > From your `streamID` branch (i.e., after running `git checkout streamID`, you can do `git rebase -i --exec 'tools/test'` as a shortcut for running `tools/test` on each new commit in the branch, i.e., each commit that's on `streamID` but isn't on the branch that `streamID` is set up to track.
> > (Normally, it's helpful to set up `streamID` to track `master` on https://github.com/zulip/zulip-mobile. E.g., if the remote pointing to https://github.com/zulip/zulip-mobile is called "origin", the default name, `streamID` would be set up to track it if you've run `git branch -u origin/master`. In my own dev environment, I've decided to use the name "upstream" instead of the default "origin", but that's up to you.)
> 
> Hi @chrisbobbe! Sorry for the late reply. Yes, you are right. Other than the last commit, all the others don't pass the tests. Would you like me to squash those commits?

As per our discussions in [this chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/First.20Contribution/near/1114676), I have squashed the commits. You can take a look at it now.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -292,9 +292,9 @@ describe('getNarrowFromLink', () => {
 
     test('on malformed stream link: treat as old format', () => {
       const expectAsName = name => expectStream(name, [streamGeneral], name);
-      expectAsName(`${streamGeneral.stream_id}`);
-      expectAsName(`-${streamGeneral.stream_id}`);
-      expectAsName(`${streamGeneral.stream_id}nonsense-general`);
+      expectAsName(`${streamGeneral.name}`);
+      expectAsName(`-${streamGeneral.name}`);
+      expectAsName(`${streamGeneral.name}nonsense-general`);
```
chrisbobbe, line 297: I see that changing `.stream_id` to `.name` makes the first and third of these assertions pass, when they'd otherwise fail with the new pattern.

Looking a bit closer, I think we'll want to make changes in a different direction, though. 🙂

- I think the first should just be removed: a stream ID by itself isn't malformed, it's acceptable in the new style, as I've mentioned elsewhere in this review (quoting zulip/zulip@2574efde3).
- The third... hmm, this reveals something interesting, possibly a bug! The new pattern happily takes "999nonsense-general" and thinks it's well-formed in the new style, taking 999 as the stream ID. That's wrong, at least according to the comment in the web app added in zulip/zulip@2574efde3 (quoting again):
  
  ```
    Modern stream slugs look like this, where 42
    is a stream id:
  
        42
        42-stream-name
  ```
  
  I've just posted on CZO [here](https://chat.zulip.org/#narrow/stream/6-frontend/topic/stream_data's.20slug_to_name/near/1115316), asking whether the web app has got this quite right, and what we should do. So, this test has been quite useful: it's alerted us to some behavior, after we make a change, that we might or might not want! 🙂

src/utils/__tests__/internalLinks-test.js
```diff
@@ -314,7 +314,7 @@ describe('getNarrowFromLink', () => {
 
       test('on ambiguous new- or old-style: new wins', () => {
         const collider = { ...eg.makeStream({ name: 'collider' }), stream_id: 311 };
-        expectStream('311', [numbers, collider], '311'); // malformed for new-style
+        expectStream('311', [numbers, collider], 'collider'); // malformed for new-style
```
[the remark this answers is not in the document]
AdityaYadavalli1: Okay, on it.

src/utils/internalLinks.js
```diff
@@ -120,9 +120,9 @@ export const decodeHashComponent = (string: string): string => {
 /** Parse the operand of a `stream` operator, returning a stream name. */
 const parseStreamOperand = (operand, streamsById): string => {
   // "New" (2018) format: ${stream_id}-${stream_name} .
-  const match = /^(\d+)-/.exec(operand);
+  const match = /^([\d]+)(-.*)?/.exec(operand);
```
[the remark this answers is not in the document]
AdityaYadavalli1: Yes, I remember you mentioning it earlier but I didn't use "non-capturing group" (I have been using regex for years now but never heard of it. Looked at the link you posted before I made any changes. Surprised to know I made all the way through without knowing about it.) to maintain consistency with the web app. I'll make another commit for this.

src/utils/__tests__/internalLinks-test.js
AdityaYadavalli1: So, I'll wait for their response before making any changes to this file then?

AdityaYadavalli1: I have removed both the 1st and the 3rd assertions. Is that okay?

src/utils/internalLinks.js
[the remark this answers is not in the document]
AdityaYadavalli1: I have made a different commit for this change. You can review it now.

src/utils/__tests__/internalLinks-test.js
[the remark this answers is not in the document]
AdityaYadavalli1: Done. Let me know if you want anymore changes.

chrisbobbe (changes requested):
Thanks, @AdityaYadavalli1! See one comment on the code, below.

Also, let's get some more detail into the commit messages, and tighten them up to match our [style requirements](https://zulip.readthedocs.io/en/latest/contributing/version-control.html). 🙂 I'd highly recommend using Greg's ["secret" to `git log -p`](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md#2-the-secret-to-using-git-log--p) to easily browse the history of commits in this project, so you can see examples of commit messages that follow our chosen style.

- The commit that fixes #4185 should be marked as fixing it, with our chosen formatting. It should indicate the zulip/zulip commit(s) that it replicates.
- The commit that changes `match[0]` to `match[1]` should briefly explain the reason for the change, in the commit message. (I gave an explanation at https://github.com/zulip/zulip-mobile/pull/4413#discussion_r557736601; you can include a link to that if you'd like.)
- The commit that starts using a non-capturing group should explain the reason for doing so. For me, the reason is pretty simple; in https://github.com/zulip/zulip-mobile/pull/4413#discussion_r557728692, I just said "we don't need to remember the part after the stream ID".

src/utils/__tests__/internalLinks-test.js
```diff
@@ -288,9 +288,7 @@ describe('getNarrowFromLink', () => {
 
     test('on malformed stream link: treat as old format', () => {
       const expectAsName = name => expectStream(name, [streamGeneral], name);
-      expectAsName(`${streamGeneral.stream_id}`);
-      expectAsName(`-${streamGeneral.stream_id}`);
-      expectAsName(`${streamGeneral.stream_id}nonsense-general`);
+      expectAsName(`-${streamGeneral.name}`);
```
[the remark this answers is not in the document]
AdityaYadavalli1: Hi @chrisbobbe, extremely sorry for the delay! I have made the suggested changes and I have tried to follow the [style requirements](https://zulip.readthedocs.io/en/latest/contributing/version-control.html). I feel that the last two commits should be squashed with the previous ones. However, it'd be easier to look through changes later if the commits are separate like they are now. Let me know what you think. 

Thank you for the help!

chrisbobbe:
Thanks, @AdityaYadavalli1!

Merged, with a few tweaks:

- Added a separate commit, at the beginning, dedicated to clarifying in the tests that an integer by itself isn't malformed in the new format.

- Removed the following change from one of the commits, as it wasn't in quite the right direction (see https://github.com/zulip/zulip-mobile/pull/4413/commits/70558725dfa90475749ec2f724ba795d9a814b12#r576315663):

  ```diff
  - expectAsName(`-${streamGeneral.stream_id}`);
  + expectAsName(`-${streamGeneral.name}`);
  ```

- Changed "internal_link" to "internalLinks" in the subject lines of the commits; "internal_link" doesn't correspond to an area of the mobile project's code.

- Squashed the last two commits together, so we adjust the tests at exactly the same time as we change the behavior that we're testing.

- Added "Fixes: #4185" to the commit that fixes the issue, so the fix is official (:tada:)!

- Edited the commit messages a bit, adding a bit more detail in the commit-message bodies, and shortening summary line lengths to 76 or below (https://zulip.readthedocs.io/en/latest/contributing/version-control.html).

AdityaYadavalli1:
I am glad I was of help (however small it is). I will keep these in mind when I contribute in the future :))

chrisbobbe:
Thanks for your help, @AdityaYadavalli1! 🙂

PR #4414, opened 2021-01-14
Anticipate and fix a few Flow errors we'd get with RN v0.63 and Flow 0.122.

chrisbobbe:
On #4245, the issue for upgrading to React Native v0.63, I [said](https://github.com/zulip/zulip-mobile/issues/4245#issuecomment-753239808):

> Much of the work has already been done, but the new Flow version that we'll take along with it has lots of issues with our use of react-navigation, so I've been hoping to finish #4296 (for which #4300 is open) before taking up the RN upgrade again.

Well, #4296 has been resolved! As we'd hoped, the new Flow version (0.122.0) doesn't complain about anything that has to do with react-navigation anymore, which brings us closer to being able to take the Flow version along with RN v0.63.

This branch whittles down the new Flow version's output to the single warning below, with the helpful comment that we can remove a $FlowFixMe along with the upgrade.

```
Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/presence/presenceReducer.js:30:9

Unused suppression comment.

     27│     case PRESENCE_RESPONSE:
     28│       return {
     29│         ...state,
     30│         // $FlowFixMe - Flow bug; should resolve in #4245
     31│         ...action.presence,
     32│       };
     33│
```

### DIFF
--- a/src/common/SlideAnimationView.js
+++ b/src/common/SlideAnimationView.js
@@ -46,7 +46,7 @@ export default class SlideAnimationView extends PureComponent<Props, State> {
     const { property, from, to, movement, style } = this.props;
     const animationValue = this.state.animationIndex.interpolate({
       inputRange: [0, 1],
-      outputRange: movement === 'out' ? [from, to] : [to, from],
+      outputRange: ((movement === 'out' ? [from, to] : [to, from]): number[]),
     });
 
     const slideStyle = { transform: [{ [property]: animationValue }] };
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -324,8 +324,22 @@ type EventUpdateMessageAction = {|
   message_id: number,
   // TODO is it really right that just one of the orig_* is optional?
   orig_content: string,
-  // $FlowFixMe clarify orig_subject vs. other orig_*, and missing vs. empty
+
+  // TODO: The doc for this field isn't yet correct; it turns out that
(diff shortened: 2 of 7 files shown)


src/common/SlideAnimationView.js
```diff
@@ -46,7 +46,7 @@ export default class SlideAnimationView extends PureComponent<Props, State> {
     const { property, from, to, movement, style } = this.props;
     const animationValue = this.state.animationIndex.interpolate({
       inputRange: [0, 1],
-      outputRange: movement === 'out' ? [from, to] : [to, from],
+      outputRange: movement === 'out' ? ([from, to]: number[]) : ([to, from]: number[]),
```
gnprice, line 49: Could this annotation be applied once to the whole expression? Would make it a bit shorter.

gnprice (commented):
Looks great, thanks! One small suggestion -- otherwise please merge at will.

src/common/SlideAnimationView.js
chrisbobbe: Yep! Thanks, done.

chrisbobbe:

Issue #4417, opened 2021-01-15
Remove most or all definitions in `navActions`.


chrisbobbe:
React Navigation 5, which we've recently upgraded to (#4296), discourages the `NavigationService` strategy we've been using, as we mentioned in 140c28cc1.

From their [doc](https://reactnavigation.org/docs/5.x/navigating-without-navigation-prop) about that strategy:

> **Do not** use this method when you have access to a `navigation` prop or `useNavigation` since it will behave differently, and many helper methods specific to screens won't be available.

And from their [doc](https://reactnavigation.org/docs/5.x/upgrading-from-4.x/#action-creators) on upgrading from v4 to v5:

> It's highly recommended to use the methods on the `navigation` object instead of using action creators and `dispatch`. It should only be used for advanced use cases.

We should aim to remove all the action creators in `navActions` and use, e.g., `navigation.navigate` instead. See discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5.3A.20Planned.20follow-up/near/1098324) and [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/navigation.2Enavigate/near/1098334).

The callsites for the action creators in `navActions` seem to fall into three categories:

- We're in a React component that functions only as a "screen" component in React Navigation (i.e., the component is used in one way: it gets passed as the `component` prop to a `<*.Screen />` component). Thus, it gets passed the `navigation` prop, and it's quite easy to just use it.
- We're in a React component that doesn't function as a "screen" component, so it doesn't automatically get the `navigation` prop.
  - These can often easily be converted to Hooks-based components, so that `useNavigation` can provide the navigation object.
  - Sometimes, (as in the case of `ComposeMenu`) the conversion to Hooks would take a bit of effort. We can consider 
    - just pushing through with the conversion,
    - passing down the `navigation` prop from the nearest component that has it, or
    - making a trivial Hooks-based wrapper component that uses `useNavigation`, in the same file, and export that.
- We're not in a React component; e.g., we're somewhere deep in the event-handling code for a button being pressed on the action sheet, or for an outbound event coming from the WebView. It might actually end up being appropriate to keep the use of `NavigationService` for these callsites (as an "advanced use case"), rather than finding a way to thread the `navigation` object through from React. But we may find that something forces our hand.

gnprice:
This was done in #5408.

[...]

PR #4418, opened 2021-01-15
Prepare to remove several definitions in `navActions`.

chrisbobbe:
Preparation for part of #4417.

> We're in a React component that doesn't function as a "screen" component, so it doesn't automatically get the `navigation` prop.
> * These can often easily be converted to Hooks-based components, so that `useNavigation` can provide the navigation object.

This PR does those easy conversions to Hooks-based components. It also adds a type-wrapper and makes a tweak to the react-navigation libdef so that our use of `useNavigation` will be well-typed.


gnprice:
Looks good, thanks! Merged.

Issue #4394, opened 2021-01-11
Make the app more accessible with a screen-reader


gnprice:
We've just completed #4232 with #4355, by adding captions to two particularly critical bits of the UI. But there are still more things we should do to make using the app a good experience with a screen reader.

The essential thing for working on this issue is to take some time using the app this way -- and before that, to take some time learning how to use one of these systems, if you're not already a regular user of them. Here's the upstream guides for doing so with either Android or iOS:
https://developer.android.com/guide/topics/ui/accessibility/testing
https://developer.apple.com/documentation/uikit/accessibility_for_ios_and_tvos/supporting_voiceover_in_your_app

In particular, the following steps are highly educational, and essential for working in this area:
* Take your actual physical device, either Android or iOS.
* Turn on its screen reader. (On Android this is called TalkBack; you may have to install it from the Play Store. On iOS it's called VoiceOver.)
* Go through the tutorial the system provides for how to use the screen reader and how to navigate around the device using it.
* Then leave the screen reader on, and keep relying on it as you try to do things on the device (including exploring the screen reader's own settings.) This way you'll keep learning more about how to use it in practice.

When I did that for the first time, on Android, I spent about a half-hour and learned quite a lot. As you continue working in this area, you'll keep doing more of it.

One general principle in choosing how a given widget should work is that we want things to feel natural and as-expected for people who regularly use one of these systems. So the labels we choose, and other aspects like what's selectable, what order things come in, etc., should match what other apps do, especially apps from the platform maker (Google or Apple) and other major apps.

Here's a list of some of the things we should improve or investigate:
- [ ] The up/back button (`NavBarBackButton.js`) gives the right label for Android, but it should probably be something else on iOS.
- [ ] I don't know if the onboarding flow (logging into Zulip, the first time you open the app) works well or at all. We should try it, find issues, and fix them.
- [ ] The pick-account screen has "trash" icons for deleting an account. These don't seem to work in TalkBack.
- [ ] … (to be continued)

gnprice:
For anyone who regularly uses a screenreader, I'd be especially interested in your feedback on what's most annoying for you in using Zulip, or thoughts on how a specific part of the UI should work, or any other aspect of this that you have something to say about :slightly_smiling_face:

gnprice:
Another note: the old PR #2753 contains fixes for some of the items above. In particular:
[...]

PR #4419, opened 2021-01-18
a11y: Make the app more accessible with screen reader

Gautam-Arora24:
**What these changes proposes?**
Add accessibility options to several parts of UI.

Still in progress...

#4394


src/compose/ComposeMenu.js
```diff
@@ -202,10 +206,20 @@ class ComposeMenu extends PureComponent<Props> {
           </View>
         </AnimatedComponent>
         {!expanded && (
-          <IconPlusCircle style={this.styles.expandButton} size={24} onPress={onExpandContract} />
+          <IconPlusCircle
+            style={this.styles.expandButton}
+            size={24}
+            onPress={onExpandContract}
+            accessibilityLabel="Attach button"
```
chrisbobbe, line 213: Hmm, I think "Attach button" isn't precisely the description for this button. One of the buttons doesn't attach anything; it takes you to a PM narrow, after you've selected the participants.

I see we've got "Collapse menu" for the button that shows in the same position when the menu has been expanded. What do you think about "Expand menu" for this one, as its opposite?

src/compose/ComposeMenu.js
```diff
@@ -183,17 +184,20 @@ class ComposeMenu extends PureComponent<Props> {
...
             />
             <IconCamera
+              accessibilityLabel="Camera"
               style={this.styles.composeMenuButton}
               size={24}
               onPress={this.handleCameraCapture}
             />
             {insertVideoCallLink !== null ? (
               <IconVideo
+                accessibilityLabel="Video Call button"
```
chrisbobbe, line 200: I think we don't need to include the word "button" in the label. It looks like there's another way to say what kind of thing it is: the [`accessibilityRole` prop](https://reactnative.dev/docs/accessibility#accessibilityrole). If we set that prop to "button", VoiceOver on iOS automatically says "button" after reading the label.

Since the value passed to `accessibilityRole` is a special identifier, I expect it doesn't have to be translated.

chrisbobbe: Also, I think the label could be a bit more specific about what it does. Maybe, "Insert video call link"? That way, it's clear that it's not going to, e.g., open a video call immediately. (The user might not like the idea of opening a video call immediately; they may want to prepare their environment a bit before doing so.)

The icon by itself is already ambiguous that way, and it will stay ambiguous for people who don't use a screen reader. But we might as well clarify it here, for people who do. 🙂

src/compose/ComposeMenu.js
```diff
@@ -169,6 +169,7 @@ class ComposeMenu extends PureComponent<Props> {
         >
           <View style={this.styles.composeMenu}>
             <IconPeople
+              accessibilityLabel="Other Users"
```
chrisbobbe, line 172: I actually haven't used this button much, myself. In fact, I had forgotten what it does, exactly.

It turns out that it's a way to open a PM narrow. That sounds useful!

One of my first guesses was that it adds an `@` mention of a specified user to the edit-message input. After all, the nearby video-call button adds something to the edit-message input. (And the other buttons end up sending a message, in the same narrow.)

I think it would be helpful to describe the action more precisely here. Maybe, "narrow to private-message conversation"?

Similarly to my comment on the video call button, the ambiguity of what action will be taken will still exist for people who don't use a screen reader. But we might as well have a good amount of description, for people who do.

src/compose/ComposeMenu.js
```diff
@@ -183,17 +184,20 @@ class ComposeMenu extends PureComponent<Props> {
               />
             )}
             <IconImage
+              accessibilityLabel="Gallery"
```
chrisbobbe, line 187: Maybe, "Send an image from the media gallery"?

src/compose/ComposeMenu.js
```diff
@@ -183,17 +184,20 @@ class ComposeMenu extends PureComponent<Props> {
...
               style={this.styles.composeMenuButton}
               size={24}
               onPress={this.handleImagePicker}
             />
             <IconCamera
+              accessibilityLabel="Camera"
```
chrisbobbe, line 193: Maybe, "Send an image from the camera"?

chrisbobbe (commented):
Thanks, @Gautam-Arora24! See a few comments below.

Also, I think it would be pretty straightforward to translate the new text strings, so let's do that. See the comment on the `GetText` type in src/types.js, and an example of using `_` in, maybe, `src/user-status/UserStatusScreen.js`. (So, e.g., `accessibilityLabel="Collapse menu"` should become `accessibilityLabel={_('Collapse menu')}`. Then, all the strings to be translated should get added to static/translations/messages_en.json.

(Possibly, there's something even smarter we can do to encode the semantics of collapsing and expanding that menu. Or to encode other things. React Native [documents](https://reactnative.dev/docs/accessibility) a *lot* of props you can pass to components!)

Ah, also: Let's remove `Fixes: #4394` from the opening comment, while still mentioning #4394 there. This PR contributes to solving that issue, but it doesn't solve all of it; I see Greg has put together quite a large checklist there. 🙂

src/compose/ComposeMenu.js
Gautam-Arora24: Thank you @chrisbobbe for your precise reply. I will consider the points mentioned by you and do the following changes.

src/compose/ComposeMenu.js
Gautam-Arora24: "Expand menu" is looking apt here :)

Gautam-Arora24:
Adding more commits for other parts of this issue...

src/streams/StreamItem.js
```diff
@@ -91,6 +91,7 @@ export default class StreamItem extends PureComponent<Props> {
       showSwitch,
       isSwitchedOn,
       unreadCount,
+      accessibilityLabel,
```
chrisbobbe, line 94: Similarly to my comment on `TopicItem`, let's not add an `accessibilityLabel` to `StreamItem`. Instead, let's have `StreamItem` build `accessibilityLabel` itself, since it already has the data it needs, in other props.

I see one of `StreamItem`'s callers doesn't provide an `unreadCount`. That's fine; `StreamItem` can decide to do one thing if it's passed an `unreadCount`, and a different thing if it's not.

src/streams/StreamList.js
```diff
@@ -75,6 +75,11 @@ export default class StreamList extends PureComponent<Props> {
             isSwitchedOn={item.subscribed}
             onPress={onPress}
             onSwitch={onSwitch}
+            accessibilityLabel={`${item.name} ${
```
chrisbobbe, line 78: This text can be translated too, even though it has special data values (the topic's name and unread count) in it. See, e.g., `alertTitle` in src/compose/MentionWarnings.js:

```js
    const alertTitle = _('Couldn’t load information about {fullName}', {
      fullName: mentionedUser.full_name,
    });
```

src/nav/NavBarBackButton.js
```diff
@@ -21,10 +21,11 @@ import * as NavigationService from './NavigationService';
 // TODO: on iOS, give the right label for a back button
 export default function NavBarBackButton(props: {| +color?: string |}) {
   const { color } = props;
+  const accessibilityLabel = Platform.OS === 'ios' ? 'Back Button' : 'Navigate up';
```
chrisbobbe, line 24: Let's make sure this `accessibilityLabel` gets translated.

`NavBarBackButton` is a function component that's set up to use [React Hooks](https://reactjs.org/docs/hooks-overview.html); we've recently started converting more of our components into this form. This looks like one of the first components that started its life as a function component (that was in 4e524e89e2182bfc6774d3649d757558e656ec05). There have also been several commits, like dd7cb8866dcb4a87bdf5c33400c57fbc57d59158, that convert a class component to a function component, to prepare for using Hooks.

A consequence of this being a Hooks-based component is that there's a different (and easier!) way to use React Context. Instead of doing `static contextType`, etc., you'll want to do this:

```js
import React, { useContext } from 'react';

// …

export default function NavBarBackButton(props: {| +color?: string |}) {
  // …
  const _ = useContext(TranslationContext`);
  // …

```

chrisbobbe: Also, the "Button" in "Back Button" seems redundant. If it's not already clear that it's a button, that seems like something that can be fixed with an `accessibilityRole` somewhere. Perhaps `Touchable` can accept an `accessibilityRole` prop, and it could have 'button' as the default value, using [`static defaultProps = { … }`](https://reactjs.org/docs/react-component.html#defaultprops)?

chrisbobbe (commented):
Thanks, @Gautam-Arora24!

In each commit that introduces any user-facing strings, we'll also need to add those strings to static/translations/messages_en.json.

A few notes on the commit messages (these are covered in Zulip's main [doc](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) on version control, which the mobile project's [style doc](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md) links to):

- Please end each summary line with a period.
- I see some commit messages where the line length is a bit too long; we aim for about 68 characters for the body. The time I spent figuring out how to get my text editor to do this automatically was well worth it. 🙂

One item that isn't covered there, but that we like to maintain in the mobile project, is that the "what" of the commit, even in the commit-message body, is described using the same syntactical form that we use in the summary line. So, e.g., "In this commit, I have added" can be shortened to simply "Add". We haven't yet documented this difference, but until we do so, I've been linking people to https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622188734. 🙂 

See also some comments below.

src/compose/ComposeMenu.js
```diff
@@ -65,7 +67,8 @@ class ComposeMenu extends PureComponent<Props> {
     const { dispatch, destinationNarrow } = this.props;
     dispatch(uploadFile(destinationNarrow, uri, chooseUploadImageFilename(uri, fileName)));
   };
-
+  static contextType = TranslationContext;
+  context: GetText;
```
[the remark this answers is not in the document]
Gautam-Arora24: Done the changes. In ```HomeScreen``` however, I am not able to add an extra line before ```static contextType = TranslationContext;```. All thanks to Prettier

src/streams/StreamItem.js
```diff
@@ -87,9 +88,15 @@ export default function StreamItem(props: Props) {
     backgroundColor !== undefined
       ? (foregroundColorFromBackground(backgroundColor): string)
       : themeColor;
+  const unreads =
+    unreadCount !== undefined && unreadCount > 0 ? `${unreadCount} unread messages` : '';
+  const accessibilityLabel = _('{name} {unreads}', {
```
chrisbobbe, line 93: In the first argument passed to `_`, the things in curly braces aren't meant to include English labels about the data they represent. Here, I'm looking specifically at "unreads", which I see sometimes contains " unread messages". A translator won't see that label, and they won't be able to translate it. 🙂

To simplify a bit, let's include the unread count whenever it's present, even if it's zero. It doesn't hurt to know that there are zero unread messages in a stream, and if it turns out to be too verbose we can always change it.

Then, let's have the code make a choice between two expressions for the entire `accessibilityLabel`. I'm thinking something like this:

```js
  const accessibilityLabel =
    unreadCount === undefined
      ? _('Stream: {streamName}', {
          streamName: name,
        })
      : _('Stream: {streamName} ({unreadCount} unread messages)', {
          streamName: name,
          unreadCount,
        });
```

Then `'Stream: {streamName}'` and `'Stream: {streamName} ({unreadCount} unread messages)'` should go in messages_en.json, and I expect they'll be quite translatable. 🙂

src/compose/ComposeMenu.js
```diff
@@ -5,7 +5,8 @@ import type { DocumentPickerResponse } from 'react-native-document-picker';
 import ImagePicker from 'react-native-image-picker';
 
 import * as NavigationService from '../nav/NavigationService';
-import type { Dispatch, Narrow } from '../types';
+import type { Dispatch, Narrow, GetText } from '../types';
```
chrisbobbe, line 8: nit in commit message: Check the spelling of `accessibleLabel`.

src/compose/ComposeMenu.js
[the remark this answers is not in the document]
chrisbobbe: Ah, right—I should have looked closer.

Right; `HomeScreen` won't allow an empty line before `static contextType = TranslationContext;`. In fact, Prettier is right, and we don't want one there! 🙂

In `ComposeMenu`, I should have recommended that the added lines go at the top, like you've done in `HomeScreen`:

```js
class ComposeMenu extends PureComponent<Props> {
  static contextType = TranslationContext;
  context: GetText;
```

We've generally been putting this pair of lines at the very top.

chrisbobbe (commented):
Thanks, @Gautam-Arora24! See some comments below.

src/nav/NavBarBackButton.js
```diff
@@ -20,11 +22,16 @@ import * as NavigationService from './NavigationService';
...
+
+  const accessibilityLabel = Platform.OS === 'ios' ? _('Back') : _('Navigate up');
   const { color } = props;
+
   return (
     <NavButton
       name="arrow-left"
-      accessibilityLabel="Navigate up"
+      accessibilityLabel={accessibilityLabel}
+      accessibilityRole="button"
```
chrisbobbe, line 34: Hmm, it seems like `accessibilityRole="button"` should be decided in one of the less specific components that `NavBarBackButton` depends on:

- `Touchable`
- `NavButtonGeneral`
- `NavButton`

In particular, it sounds to me like the only right way to use `NavButtonGeneral` is where its role is a button, so it could probably be decided there, without even exposing an `accessibilityRole` prop itself. So, like this:

src/nav/NavButtonGeneral
```diff
   render() {
-    const { children, onPress, accessibilityLabel, accessibilityRole } = this.props;
+    const { children, onPress, accessibilityLabel } = this.props;
 
     return (
       <Touchable
         onPress={onPress}
         accessibilityLabel={accessibilityLabel}
-        accessibilityRole={accessibilityRole}
+        accessibilityRole="button"
       >
         <View style={this.styles.navButtonFrame}>{children}</View>
       </Touchable>
```

It's probably less certain that `Touchable` should have a "button" role. It might be a sensible default (I can't think of a way we currently use it that isn't as a button), but it's a good idea to let that be overridden.

A simple way to give `Touchable` that default would be the following:

src/common/Touchable.js
```diff
-    const { accessibilityLabel, style, onPress, onLongPress, accessibilityRole } = this.props;
+    const {
+      accessibilityLabel,
+      style,
+      onPress,
+      onLongPress,
+      accessibilityRole = 'button',
+    } = this.props;
```

chrisbobbe (commented):
Thanks, @Gautam-Arora24! I wonder if there was some work that you forgot to push; I've replied "bump" on a few points where I haven't found a resolution. 🙂 

Oh, unfortunately GitHub's UI isn't doing a good job of showing my "bump" comments below in the actual context where I wrote them. You'll find that context above, or at least I can see it—please let me know if something isn't clear. 🙂

static/translations/messages_en.json
```diff
@@ -5,6 +5,14 @@
...
   "Starred": "Starred",
   "Mentions": "Mentions",
+  "Mentions ({count} unread)":"Mentions ({count} unread)",
+  "Stream: {streamName}":"Stream: {streamName}",
+  "Stream: {streamName} ({unreadCount} unread messages)":"Stream: {streamName} ({unreadCount} unread messages)",
+  "Topic: {name}":"Topic: {name}",
+  "Topic: {name} ({unreadCount} unread messages)":"Topic: {name} ({unreadCount} unread messages)",
+  "Back":"Back",
+  "Navigate up":"Navigate up",
+  "Starred messages":"Starred messages",
```
chrisbobbe, line 15: Nit (for all these added entries): put a space after the colon, like in the rest of the file. So, `"...": "..."` instead of `"...":"..."`.

chrisbobbe (changes requested):
Thanks, @Gautam-Arora24! The GitHub UI tells me that I had most of a review ready to post, from a while ago, but I never actually sent it—oops!

So, here's that review. Also, while I was looking at the code locally again, I went ahead and rebased the branch atop the latest `master` and resolved some rebase conflicts for you; please pull the changes from the remote branch to your local branch, and start your next round of changes from there. 🙂

src/utils/narrow.js
```diff
@@ -537,3 +537,24 @@ export const getNarrowForReply = (message: Message | Outbox, ownUserId: UserId)
     return topicNarrow(streamName, message.subject);
   }
 };
+
+/**
+ * Answers, this question, "What should be the right accessibilityLabel for
+ * send message button?"
```
chrisbobbe, line 543: So, the reason I suggested putting it in narrow.js was that I figured it could end up being useful for things besides the send message button; my thought was that it could be broadly useful for a lot of different areas in the UI. So this comment is too specific. 🙂 Possibly most of its uses will be for improving our interface with screen readers.

How about something like, "A translated description of a narrow, useful for screen-reader interactions."

Then, in that case, please do put useful descriptions for all types of narrows (`allPrivate`, `home`, etc.).

src/utils/narrow.js
```diff
@@ -537,3 +537,24 @@ export const getNarrowForReply = (message: Message | Outbox, ownUserId: UserId)
...
+ *
+ * This function only provide labels for stream, topic and pm narrow.
+ * For other narrows, there is no point of label.
+ */
+export function narrowUiDescription(narrow: Narrow, _: GetText): string {
+  return caseNarrow(narrow, {
+    stream: name => _('Narrow: stream {streamName}', { streamName: name }),
+    topic: (streamName, topic) =>
+      _('Narrow: stream {streamName}, topic {topic}', { streamName, topic }),
+    pm: () => _('Narrow: PM'),
```
chrisbobbe, line 553: This should list everyone in the PM conversation, as I mentioned [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Adding.20Accessibilty/near/1130272).

src/compose/ComposeBox.js
```diff
@@ -444,6 +446,7 @@ class ComposeBox extends PureComponent<Props, State> {
     }
 
     const placeholder = getComposeInputPlaceholder(narrow, ownUserId, allUsersById);
+    const accessibilityLabel = narrowUiDescription(narrow, _);
```
chrisbobbe, line 449: This shouldn't be called `accessibilityLabel` because it's not the entire accessibility label. Let's call it something like `narrowDescription`.

src/utils/narrow.js
```diff
@@ -537,3 +537,24 @@ export const getNarrowForReply = (message: Message | Outbox, ownUserId: UserId)
...
 };
+
+/**
+ * Answers, this question, "What should be the right accessibilityLabel for
+ * send message button?"
+ *
+ * This function only provide labels for stream, topic and pm narrow.
+ * For other narrows, there is no point of label.
+ */
+export function narrowUiDescription(narrow: Narrow, _: GetText): string {
```
chrisbobbe, line 548: Ah, looking again at the other things in this file: I think a name like `uiDescriptionOfNarrow` would be better. (See that there are several functions that end in `OfNarrow` and take a `Narrow`.

src/compose/ComposeBox.js
```diff
@@ -510,7 +513,7 @@ class ComposeBox extends PureComponent<Props, State> {
             />
           </View>
           <FloatingActionButton
-            accessibilityLabel="Send message"
+            accessibilityLabel={_('Send Message, {accessibilityLabel}', { accessibilityLabel })}
```
chrisbobbe, line 516: This should say something like `Send message to: {narrowDescription}`. I'm hoping that

1. This makes plenty of sense to English speakers, and
2. Translators can see that string and recognize what it means (remember, they don't generally see the code!), and
3. Translators can find a way to render it in the target language.

src/nav/NavBarBackButton.js
```diff
@@ -18,13 +20,16 @@ import * as NavigationService from './NavigationService';
  * https://material.io/design/navigation/understanding-navigation.html
  */
 // TODO: on iOS, show the right icon for a back button
-// TODO: on iOS, give the right label for a back button
 export default function NavBarBackButton(props: {| +color?: string |}) {
+  const _ = useContext(TranslationContext);
+
+  const accessibilityLabel = Platform.OS === 'ios' ? _('Back') : _('Navigate up');
```
chrisbobbe, line 25: nit: Let's inline this instead of putting it in a new variable. So, down below,

```
accessibilityLabel={Platform.OS === 'ios' ? _('Back') : _('Navigate up')}
```

chrisbobbe (changes requested):
Thanks, @Gautam-Arora24! See a few comments below. Same thing as last time; I've gone ahead and resolved a rebase conflict and pushed the result to the PR branch; please pull that and work from there. 🙂

Issue #4365, opened 2020-12-30
Avatars don't work on iOS 14, except in message list


titanous:
The profile picture in the bottom tab bar is missing on local builds targeting the current version of iOS (see screenshot below). Additionally, the margin for the "home swipe bar" appears to be missing. I have tested this in a few different configurations, and here's what I found:

**Working**
- Release version (v27.157) from the App Store on an iPhone 12 with iOS 14.3
- master (05048c7007d30a6967333a3354f38c3ece68b3e8) built with Xcode 12.3 running in Simulator 13.7

**Broken**
- master built with Xcode 12.3 on an iPhone 12 with iOS 14.3
- master built with Xcode 12.3 running in Simulator 14.3
- master built with Xcode 12.3 running in Simulator 14.0
- v27.157 built with Xcode 12.3 running in Simulator 14.3

![Simulator Screen Shot - iPhone 12 - 2020-12-29 at 21 29 32](https://user-images.githubusercontent.com/13026/103326172-17e01300-4a1d-11eb-8c8e-094d0e9d14e7.png)

chrisbobbe:
Thanks for the detailed report, @titanous!

@gnprice, you mentioned [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60Immutable.2EMap.60.20performance.20for.20.60state.2Enarrows.60/near/1066833) the possibility of upgrading your test device's iOS version to 14. Did you end up doing that, or would it be best to keep it on its current version while plenty of people are still using that version? 🙂 (The latter has been my strategy for my iPhone so far.)

gnprice:
Thanks @titanous for the detailed report!

I just tried this in an iOS 14 test device, and it's working. That's:
* v27.157 from the App Store (well, TestFlight) on an iPhone XR with iOS 14.2

I see the profile picture there, and there's a margin below those five bottom icons.

One other variable which could affect this behavior is the type of avatar -- Zulip avatars can be hosted in a few different ways and places, and we've had bugs in the past which affected one or another type differently. In particular there's:
* A Gravatar, the default for a new user;
* An avatar the user uploaded, and which the Zulip server hosts locally (as chat.zulip.org and most other Zulip servers do);
* An avatar the user uploaded, and which the Zulip server stores in S3 and serves a redirect to there (as zulipchat.com does);
[...]

PR #4321, opened 2020-12-01
deps: Follow @babel/runtime upgrade from RN template app.

chrisbobbe:
Part of the RN v0.62 -> v0.63 changes to the template app [1],
corresponding to facebook/react-native@82e823933.

That RN commit upgrades a few other deps; we don't take those
changes, for the following reasons:

- We're already past ^7.6.2 of @babel/core.

- We don't currently use @react-native-community/eslint-config
  (though we're considering it, in #4119).

- We don't currently use babel-jest.

- We're already past ^25.1.0 of jest.

- We're already past ^0.58.0 of metro-react-native-babel-preset.

Run our usual `yarn yarn-deduplicate && yarn`, as prompted by
`tools/test deps`.

[1] https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.4

PR #4320, opened 2020-11-27
android build: Upgrade Android Gradle Plugin to v3.5.3.

chrisbobbe:
Release notes here:
  https://developer.android.com/studio/releases/gradle-plugin#3-5-0
No major potentially-incompatible changes identified there.

Part of the RN v0.62 -> v0.63 changes to the template app [1],
corresponding to facebook/react-native@e1e081b00.

[1] https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.3

PR #4420, opened 2021-01-20
Upgrade to React Native v63.

chrisbobbe:
Following #4296, #4318, #4319, #4320, #4321, #4322, and #4414, complete the RN v0.63 upgrade! 🎉

If, during testing, you encounter an error during the `postinstall` script like either of the errors described in https://github.com/zulip/zulip-mobile/pull/4420#issuecomment-763883027, try running `rm -rf ios/Pods` and rerunning `yarn`; no further action should be necessary. This has worked for me, in testing, when hopping between before and after the upgrade. (As described in that comment, I've already committed the result of running a `pod update` command in the upgrade commit. That seems to have fixed something that wouldn't be fixed by merely removing ios/Pods.)

If, when testing on Android, you get an instant crash, try removing `android/.gradle` and running `./gradlew clean` from the `android` directory, as recommended in a few places in our [build-run doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/build-run.md).

Fixes: #4245
Fixes: #4365


chrisbobbe:
These are the RN commits that touch the [RN template app](https://github.com/facebook/react-native/tree/master/template) that are in v0.63.4 but are not in v0.62.2.

I.e., they're the commits in the output of `git log --oneline --reverse v0.62.2..v0.63.4 -- template`, when I run it in a clone of the facebook/react-native repo.

Along with whether, how, and when we're propagating the changes to our app.

We shouldn't forget to update this, if necessary, along with any changes made during code review and merge.

| RN commit | When done | Comment |
| --- | --- | --- |
| facebook/react-native@e1e081b00 Update Android Gradle plugin to 3.5.3 (#27462) | #4320 | … |
| facebook/react-native@f11b4895d Edits to flipper pods comments (#27485) | 9a144c5b3 | This was backported to v0.62 in facebook/react-native@8f954b365, so we took it in that upgrade. See also #4322. |
| facebook/react-native@aa7845734 v0.114.0 | Upgrade commit | … |
| facebook/react-native@68bf0e32e Fix RNTester integration with Flipper (#27482) | f2e611995 | This was backported to v0.62 in facebook/react-native@3c3b4cf57, so we took it in that upgrade. See also #4322. |
| facebook/react-native@7980615d3 remove `xcshareddata` from gitignore (#27563) | N/A | We never took the addition of this line from facebook/react-native@d57cdac62; it was done in error. (The removal was backported to v0.61 in facebook/react-native@3aaebe60c, and to v0.62 in facebook/react-native@b0eddefcb, so we would have gotten it in one of those upgrades anyway.) See also my comment on facebook/react-native@a715decd2. |
| facebook/react-native@98ebc1ea2 Adjust HelloWorld-tvOSTests/Info.plist `CFBundleIdentifier` to use PR ODUCT_BUNDLE_IDENTIFIER (#27601) | N/A | We don't have a ZulipMobile-tvOSTests/Info.plist because we don't do anything with tvOS. |
| facebook/react-native@87370c4c0 Remove use of inexact-spread Flow lint | #4318 | … |
| facebook/react-native@674b59180 iOS: Deprecate iOS 9 / tvOS 9 SDK support | N/A | We've already got our iOS deployment target at 11 (we went from 10 to 11 in c953bc336). |
| facebook/react-native@829a2237d iOS/tvOS 9.0 ==> 10.0 in .xcodeproj | N/A | We've already got our iOS deployment target at 11 (we went from 10 to 11 in c953bc336). |
| facebook/react-native@15ce226f5 v0.115 in xplat/js | Upgrade commit | … |
| facebook/react-native@159624d90 React Native sync for revisions 6cff70a...19f6fe1 | Upgrade commit | We don't take the react-test-renderer upgrade because we don't use that package. |
| facebook/react-native@aa0ef1533 Applied missing changes from bumping Gradle wrapper to 6.0.1 (#27639) | #4319 | … |
| facebook/react-native@c40988cc2 Deploy Flow v0.116 to xplat | Upgrade commit | … |
| facebook/react-native@a89a55353 Update Gradle wrapper to 6.1 (#27800) | #4319 | … |
| facebook/react-native@edcbfb982 Bump Android build-tools to 29.0.2, compileSdk to 29 | N/A | Our `compileSdkVersion` and `targetSdkVersion` are already on 29, so no need to change those. We don't need to specify a `buildToolsVersion`; see e0b0b97d2 and [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/buildToolsVersion.20in.20RN.20v0.2E63.20upgrade/near/1073261). |
| facebook/react-native@9c3fa5733 Remove FB copyright notices from iOS template (#27725) | cc51ea2bf | This was backported to v0.62 in facebook/react-native@6a10d5dfd, so we took it in that upgrade. |
| facebook/react-native@7bc2b9179 Deploy Flow v0.117.0 to xplat | Upgrade commit | … |
| facebook/react-native@947a5593e Update Gradle Wrapper to 6.1.1 (#27865) | #4319 | … |
| facebook/react-native@939df8a01 Upgrade Flipper version (#27837) | 2fc3bc10a | This was backported to v0.62 in facebook/react-native@e7bfa1fc2, so we took it in that upgrade. See also #4322. |
| facebook/react-native@52cd9cd6f initalizeFlipper should be set in template app by default (#27569) | 7c7271598 | This was backported to v0.62 in facebook/react-native@05f5cb534, so we took it in that upgrade. See also #4322. |
| facebook/react-native@8f93dedc6 Remove empty Swift file for Flipper (#27922) | 9a144c5b3 | This was backported to v0.62 in facebook/react-native@e3218a0d9, so we took it in that upgrade. See also #4322. |
| facebook/react-native@82e823933 Upgrade all dependencies in package.json template (#27928) | #4321 | See #4321 for why we didn't take all of this commit. |
| facebook/react-native@2a0c4c878 Remove types-first from flowconfig template | N/A | A clean reversion of facebook/react-native@d9b84450a. We didn't take react-native@d9b84450a because its clean reversion (this commit) was backported to v0.62 in facebook/react-native@cae662bfe. |
| facebook/react-native@82165673f chore: add eslint back to template (#27983) | N/A | We're happy with our scripts in package.json, so we don't take this. |
| facebook/react-native@cd194b833 Deploy Flow v0.118 | Upgrade commit | … |
| facebook/react-native@a35efb940 Use autolink script in template on iOS (#27984) | Upgrade commit | Mentioned in #4322, this is when the RN-related but non-Flipper-related code was moved to a function defined by React Native (`use_react_native!`). We're now expected to call this function instead of copy a bunch of stuff to our own Podfile (which is nice!). I haven't yet been moved to propagate all the v0.62 -> v0.63 changes to RN's script over to the code that runs in our Podfile, so that we could make the switch to the RN call as an NFC commit after the upgrade commit. (See my comment on facebook/react-native@619d5d60d, where I said I'd done so for the Flipper-related code.) |
| facebook/react-native@4bb0b4f20 Remove `pickFirst` arguments (#27964) | #4322 | … |
| facebook/react-native@29d3dfbd1 add Detox integrations to Android RNTester app | N/A | We have these lines already (but with a different value for the Kotlin version) because we're already using Kotlin. |
| facebook/react-native@619d5d60d Smoothen Flipper iOS integration (#28044) | After the upgrade commit. Partial (some has already been done). | As I mentioned in #4322, *most* of this was backported to RN v0.62, in facebook/react-native@b4d1fcfb2. We took the portion that was, in 9a144c5b3. What remains to do in this upgrade is, as I mentioned in #4322, "calling RN's code instead of duplicating it in our Podfile." On Greg's [recommendation](https://github.com/zulip/zulip-mobile/pull/4322#issuecomment-745780259), I propagated all (not just most) of the v0.62 -> v0.63 changes to RN's script over to the code that runs in our Podfile. This means we can make the switch to the RN call as an NFC commit after the upgrade commit. |
| facebook/react-native@a1b14deb3 Add androidx.swiperefreshlayout to template build.gradle (#28071) | d179e901a | This was backported to v0.62 in facebook/react-native@99ef72cad, so we took it in that upgrade. |
| facebook/react-native@0e4bcaa29 Resolve localization warnings (#28046) | 1ebd4d4b4 | This was backported to v0.62 in facebook/react-native@ebb629d05, so we took it in that upgrade. |
| facebook/react-native@d4d8887b5 Update Gradle Wrapper to 6.2 (#28112) | #4319 | … |
| facebook/react-native@4118d7982 Rename autolinking-ios.rb script and bring RNTester and template in line. (#28077) | Upgrade commit and after. | The `../node_modules/react-native/scripts/{ autolink-ios -> react_native_pods }` file rename here is why we use the `react_native_pods` name when we handle facebook/react-native@a35efb940 in the upgrade commit. The `add_flipper_pods!` -> `use_flipper!` rename here is why we use the `use_flipper!` name when we handle the remainder of facebook/react-native@619d5d60d, after the upgrade commit (see our comment at facebook/react-native@619d5d60d for what "the remainder" means). |
| facebook/react-native@c0d8c1db9 Updating the URLs to point at new domain name reactnative.dev | 2afbcf706 | … |
| facebook/react-native@6ea963de6 @allow-large-files deploy 0.119 to xplat | Upgrade commit | … |
| facebook/react-native@a715decd2 Add xcscheme files back in the template (#28198) | N/A | We never took facebook/react-native@3e4c5c09c, which removed these files in error. (This fix was backported to v0.62 in facebook/react-native@4efa3baf0 so we would have gotten it then anyway.) |
| facebook/react-native@80e6d672f UIViewController-based status bar management (#25919) | N/A | We don't take this; it's cleanly reverted (unclear why, though) in facebook/react-native@f4538e577. |
| facebook/react-native@f4538e577 Back out "UIViewController-based status bar management" | N/A | Cleanly reverts facebook/react-native@80e6d672f (unclear why, though), which we didn't take; ignore. |
| facebook/react-native@1beef1e18 deploy 120 to xplat | Upgrade commit | … |
| facebook/react-native@db1225564 Upgrade Flipper Android to 0.33.1 (#28275) | #4322 | … |
| facebook/react-native@44ae0fea0 Re-sync with internal repository | N/A | We don't take this; it's cleanly reverted in facebook/react-native@067b7a836. |
| facebook/react-native@067b7a836 Revert "Merge pull request #28292 from hramos/fixup-T63861364-master" | N/A | Cleanly reverts facebook/react-native@44ae0fea0, which we didn't take; ignore. |
| facebook/react-native@07def5539 fbshipit-source-id: da15f69185e724eaf7d4bc78dbc61fcdcb3074d5 | N/A | No need to include; accidentally removes CRLF line endings. This is reverted in facebook/react-native@8988a073b (released RN v0.64-rc.0). |
| facebook/react-native@bac9e6959 React Native sync for revisions 241c446...b5c6dd2 | Upgrade commit | We don't take the react-test-renderer upgrade because we don't use that package. |
| facebook/react-native@9637d6214 Update React package | Upgrade commit | … |
| facebook/react-native@6fe7cd88d Deploy Flow 0.121 to Xplat (#901) | Upgrade commit | … |
| facebook/react-native@ad86a1830 chore: update lint config in template (#28443) | N/A | We don't use @react-native-community/eslint-config. We might eventually start to; see #4119. We don't take the metro-react-native-babel-preset upgrade because we're already on v0.63.0. |
| facebook/react-native@4a48b021d upgrade to flow 0.122.0 | Upgrade commit | … |
| facebook/react-native@696fb5588 Update default Podfile to not depend on a path (#28572) | Upgrade commit | … |
| facebook/react-native@7acd6672a chore: remove Kotlin version from the default template | N/A | We should leave these lines in because we use Kotlin; see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Kotlin.20in.20RN.20v0.2E63.20upgrade/near/1073268). |
| facebook/react-native@e2dd18d8b [0.63.0-rc.0] Bump version numbers | Upgrade commit | … |
| facebook/react-native@87f5b8b79 Remove the post install step (#28651) | N/A (tracked in #4322) | We don't take this; it's cleanly reverted in facebook/react-native@574447a. |
| facebook/react-native@574447ae3 Revert D21064653: Remove the post install step | N/A (tracked in #4322) | Cleanly reverts facebook/react-native@87f5b8b, which we didn't take; ignore. |
| facebook/react-native@208bd0569 Bump @react-native-community/eslint-config in new app template | N/A | We don't use this. We might eventually start to; see #4119. |
| facebook/react-native@48413a443 [0.63.0-rc.1] Bump version numbers | Upgrade commit | … |
| facebook/react-native@27ccc6019 Upgrade Flipper to 0.37.0 (#28545) | f7005c50c and #4322 (271d9c482) | … |
| facebook/react-native@29639e7b9 Enable with CocoaPods `:configuration` (#28796) | #4322 | … |
| facebook/react-native@0225f180f Changed iOS LaunchScreen from xib to storyboard (#28239) | 78a62b249 | … |
| facebook/react-native@4f897336b [0.63.0] Bump version numbers | Upgrade commit | … |
| facebook/react-native@ce0fd0a57 [0.63.1] Bump version numbers | Upgrade commit | … |
| facebook/react-native@a642c3f26 Exclude okhttp from flipper dependency (#29260) | #4322 | … |
| facebook/react-native@3c9e5f147 [0.63.2] Bump version numbers | Upgrade commit | … |
| facebook/react-native@66c8cd393 Update Flipper (#29787) | f7005c50c and #4322 (271d9c482) | … |
| facebook/react-native@7100756bb [0.63.3] Bump version numbers | Upgrade commit | … |
| facebook/react-native@27651720b [0.63.4] Bump version numbers | Upgrade commit | … |

chrisbobbe:
This comment describes an issue that I ran into and solved with a simple command; the result is in the current revision.

On running `yarn` after updating the dependencies `react-native`, `react`, etc., in package.json, I got the following red error output, from the `pod install` step of our `postinstall` script:

```
[!] CocoaPods could not find compatible versions for pod "Folly":
  In snapshot (Podfile.lock):
    Folly (from `../node_modules/react-native/third-party-podspecs/Folly.podspec`)

  In Podfile:
    FBReactNativeSpec (from `../node_modules/react-native/Libraries/FBReactNativeSpec`) was resolved to 0.63.4, which depends on
      Folly (= 2020.01.13.00)

    Folly (from `../node_modules/react-native/third-party-podspecs/Folly.podspec`)
```

I then `rm -rf ios/Pods` and run `yarn` again, and get the following:

```
[!] CocoaPods could not find compatible versions for pod "Folly":
  In snapshot (Podfile.lock):
    Folly (from `../node_modules/react-native/third-party-podspecs/Folly.podspec`)

  In Podfile:
    Folly (from `../node_modules/react-native/third-party-podspecs/Folly.podspec`)

It seems like you've changed the version of the dependency `Folly` and it differs from the version stored in `Pods/Local Podspecs`.
You should run `pod update Folly --no-repo-update` to apply changes made locally.
```

This second message matches the one reported in facebook/react-native#30015. The suggested solution there is to run the recommended command, `pod update Folly --no-repo-update`. This seemed to work for a lot of people, and no one has yet responded to say it doesn't work or causes problems. There are also posts on other threads, like [this post](https://github.com/react-native-community/upgrade-support/issues/113#issuecomment-705744522), reporting that this command removed a roadblock in their RN upgrade.

The command seems to work for us. I observe that the `yarn` command by itself (i.e., with `postinstall` giving one of the above errors) causes no change to ios/Podfile.lock. After running `pod update Folly --no-repo-update` (with or without having removed `ios/Pods` first), I see quite a lot of changes to ios/Podfile.lock, and a quick skim through them suggests they're the ones we'd expect to see with this RN upgrade.

So, run and commit the result of `pod update Folly --no-repo-update` in the main upgrade commit (and leave this comment explaining why).

gnprice:
Thanks @chrisbobbe ! I'll look closer at this today.

> * We're still waiting for RN v0.63 support from react-native-cameraroll

If your PR there (or some alternative) isn't merged upstream when we're otherwise ready to merge this upgrade, we can always handle this by pointing to your PR commit. But the project does seem to be actively merging PRs, and I see you left a ping on the issue just today, so hopefully that state won't last long.

src/compose/ComposeBox.js
```diff
@@ -348,9 +348,6 @@ class ComposeBox extends PureComponent<Props, State> {
     }
     completeEditMessage();
     if (this.messageInputRef.current !== null) {
-      // Should be fixed in RN v0.63 (#4245); see
-      // https://github.com/zulip/zulip-mobile/issues/4245#issuecomment-695104351.
-      // $FlowFixMe
```
gnprice, line 353: Since the `$FlowFixMe`s start getting flagged as "unused suppression
    comments", remove them.
    
    Strangely, `this.textInputRef.current` and friends are
    `any(implicit)` after this.

Hmm -- that is concerning. And indeed if I go and change these method calls to nonsense names, I get no error.

I think this may ultimately all be down to bugs in Flow's special support for React. Both `React$ElementRef`, and `React$AbstractComponent` aka `React.AbstractComponent`, which is part of the `TextInput` type, are implemented by special logic within Flow. I suspect that that logic isn't fully working as intended.

If we're not going to get type-checking on these anyway, I think I'd rather have the fixmes there, so it's obvious looking at the code that it isn't type-checked and you need to be careful about the types by hand.

src/common/InputWithClearButton.js
```diff
@@ -32,7 +32,7 @@ export default class InputWithClearButton extends PureComponent<Props, State> {
     canBeCleared: false,
     text: '',
   };
-  textInputRef = React.createRef<typeof TextInput>();
+  textInputRef = React.createRef<React$ElementRef<typeof TextInput>>();
```
gnprice, line 35: A similar change should also happen at `Input` in the type of its prop that we pass this to.

(Though maybe not now after all; see previous comment.)

ios/Podfile
```diff
@@ -2,87 +2,17 @@
...
-  pod 'FBLazyVector', :path => "../node_modules/react-native/Libraries/FBLazyVector"
-  pod 'FBReactNativeSpec', :path => "../node_modules/react-native/Libraries/FBReactNativeSpec"
-  pod 'RCTRequired', :path => "../node_modules/react-native/Libraries/RCTRequired"
-  pod 'RCTTypeSafety', :path => "../node_modules/react-native/Libraries/TypeSafety"
+  use_react_native!
 
-  pod 'React', :path => '../node_modules/react-native/'
-  pod 'React-Core', :path => '../node_modules/react-native/'
-  pod 'React-CoreModules', :path => '../node_modules/react-native/React/CoreModules'
-  pod 'React-Core/DevSupport', :path => '../node_modules/react-native/'
```
gnprice, line 62: One change here is that this particular pod no longer gets included. The upstream function has this:
```ruby
  unless production
    pod 'React-Core/DevSupport', :path => "#{prefix}/"
  end
```

and `production` defaults to `false`.

I'm not sure what that does. It sounds like something that's good to leave out in release builds, but we might miss in dev builds. But I guess if things still seem to work well in development, then we didn't need it?

gnprice (commented):
This looks great! Thanks for all your work on this upgrade, and for the clear table above. A few comments below.

ios/Podfile
chrisbobbe: Since `production` defaults to `false`, and we don't take it away from the default (which we would do by passing an argument with our `use_react_native!` call in the Podfile), I think this pod is still included in all builds.

```
  unless false
    // This code will definitely run
  end
```

chrisbobbe: I've just changed it to `unless true` in node_modules and run `pod install` (via `yarn`), and I see a line with `React-Core/DevSupport` removed from `Podfile.lock`.

(The `/` says to look for `DevSupport` as a "subspec" of `React-Core`. I think all subspecs of `React-Core` would normally have been pulled in by the `pod 'React-Core'` line, above this conditional—but this isn't happening here because `React-Core`'s podspec declares a "default subspec"; [doc](https://guides.cocoapods.org/syntax/podspec.html#subspec).)

Probably best to keep pulling in `DevSupport` for development, but I agree that it should be fine to exclude it in production builds. I've just tried doing a release build with it excluded, and it built with no errors and is so far running just fine on my iPhone 6s.

I'd be nice if React Native took the approach it took with `use_flipper!` in facebook/react-native@c72ecdb906f29fbf9f97a2de2e9997a0f41d0352 (see also our 61bda2a20dfce2a21e3a485e73b43cbfe71a3985) toward `use_react_native!`. The "build configuration" (the thing we use to say whether we're building for production or development) isn't known at `pod install` time, which is the time the Podfile is used. So I don't see a way we could pass a correct value for the `production` param that `use_react_native!` expects.

chrisbobbe: > So I don't see a way we could pass a correct value for the `production` param that `use_react_native!` expects.

Given this, it's possibly just fine to not do anything for now. Including `React-Core/DevSupport` in release builds is the status quo anyway.

chrisbobbe:
Thanks for the review, @gnprice! I've just pushed a new revision that addresses https://github.com/zulip/zulip-mobile/pull/4420#discussion_r561355599 and an additional type-checking thing that had slipped through the cracks. I've also responded to the point about `React-Core/DevSupport`, at https://github.com/zulip/zulip-mobile/pull/4420#discussion_r562167018.

Next, I'll go ahead and try pointing to my PR commit for react-native-cameraroll, as you've suggested at https://github.com/zulip/zulip-mobile/pull/4420#issuecomment-763973831. I'm not 100% sure there isn't some kind of packaging step that contributes to what we get from NPM; I'll find out soon enough. 🙂

ios/Podfile
gnprice: > `unless false` is equivalent to `if true`, I think. Is that right?

Ha, yes, I guess I misread that code :upside_down_face:

gnprice: Yeah, I agree with all the above. This change is NFC -- we were always getting this pod, and we're still always getting it. And the appropriate thing would be for that function to do something like what `use_flipper!` does, and take a list of configurations for it to apply in.

(Or else have some other way of being conditioned on the configuration.)

gnprice:
All looks good, thanks! And that sounds like a good plan.

chrisbobbe:
> All looks good, thanks! And that sounds like a good plan.

Great, thanks; I've just added a commit that points to the PR commit on react-native-cameraroll/react-native-cameraroll#264. I suppose it would be slightly neater to have this commit live on a branch owned by `zulip` instead of `chrisbobbe`, but I don't have the permissions to fork to `zulip`.

gnprice:
Looks great -- merged!

> I suppose it would be slightly neater to have this commit live on a branch owned by `zulip` instead of `chrisbobbe`

I think that would definitely be better if we found ourselves maintaining a fork of that repo, with ongoing changes. (As we did for remotedev-serialize, starting at 9d0aa954e28c54075a86af542cefefdc94b57e4a, before we ended up inlining that code and integrating it more closely into the main mobile repo.)

But when the intent really is "just this one PR, until it's hopefully merged soon", I think pointing straight to the branch the PR was sent from fits pretty well. If in a few months we notice that it's still there and the PR isn't merged, and there's some other maintenance it needs, we might end up moving to the fork model; but hopefully we'll instead get to just switch back to pointing upstream, and then get to forget about this package's maintenance again for a while.

chrisbobbe:
Great, thanks! 🙂

> But when the intent really is "just this one PR, until it's hopefully merged soon", I think pointing straight to the branch the PR was sent from fits pretty well. 

Makes sense. I've just opened #4425 so we don't totally forget to switch back to a published version on NPM.

PR #4421, opened 2021-01-20
types: Use a distinct UserId type for user IDs.

gnprice:
This makes a new `UserId` type as a Flow [opaque type alias](https://flow.org/en/docs/types/opaque-types/), like so:
```js
export opaque type UserId: number = number;
```
The effect of this is that:
* at runtime, a `UserId` is simply a `number`, with no added overhead;
* a `UserId` can be freely converted to a plain `number` (because of the `: number` bound);
* but to create a `UserId` from a plain `number` can only be done by code in the same module as the type definition. The effect of this part is that it requires invoking an explicit function `makeUserId` which we export from that module.

This requires touching a lot of code and, frankly, I'm not sure it was worth the effort required to prepare the changes. But with that work done, I think it definitely makes the code nicer than it was without it. It moves one more set of facts about our code out of the realm of unchecked commenting and naming, and into the realm of machine-checked types.

### DIFF
--- a/src/api/idTypes.js
+++ b/src/api/idTypes.js
@@ -1,0 +1,54 @@
+/* @flow strict-local */
+
+/**
+ * A user ID.
+ *
+ * This is a number that identifies a particular Zulip user.  Different
+ * users on the same Zulip server will have different user IDs.  On the
+ * other hand, between different Zulip servers the same user ID may be used
+ * to refer to completely unrelated users.
+ *
+ * In general, if something calls for a value of this type, then you should
+ * be getting it from something that already has this type: like the
+ * `user_id` property of a `User` object, or a data structure that stores
+ * user IDs.
+ *
+ * The only other way to create a `UserId` is to call the `makeUserId`
+ * function provided by this module.
+ *
+ * See also type `User`, for the thing that one of these identifies.
+ */
+// How this type works: it's an "opaque type alias" for simply a number.
+// See Flow docs: https://flow.org/en/docs/types/opaque-types/
+//
+// This means that:
+//  * At runtime, a `UserId` value is just a number.  The use of `UserId`
+//    involves zero runtime overhead compared with simply `number`.
+//  * Because we've written a "bound" of `: number`, all code that has one
+//    of these values can freely use it as if it were simply `number`.
+//  * On the other hand, the only way to *create* such a value is to invoke
+//    something from this module to do it for you.
+//
+// For more background discussion of opaque types, see `PmKeyRecipients`.
+export opaque type UserId: number = number;
+
+/**
+ * Take a number, and declare that it truly is a user ID.
+ *
+ * This does nothing at all at runtime, just returning the value it's
+ * passed.  Its only effect is to inform the type-checker that it's OK to
+ * use this value where a user ID is required.
+ *
+ * In general the only legitimate use case for this function, outside of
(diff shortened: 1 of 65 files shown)


src/webview/js/js.js
```diff
@@ -9,6 +9,7 @@ import type {
   WebViewInboundEventReady,
   WebViewInboundEventMessagesRead,
 } from '../generateInboundEvents';
+import { makeUserId } from '../../api/idTypes';
```
chrisbobbe, line 12: We should run `tools/generate-webview-js` at this commit.

src/api/idTypes.js
```diff
@@ -0,0 +1,54 @@
...
+ * use this value where a user ID is required.
+ *
+ * In general the only legitimate use case for this function, outside of
+ * tests, is when parsing a user ID from a string.  When getting a user ID
+ * from any other source, if the values really are user IDs then the type of
+ * that source should be adjusted to say so.
+ */
+export const makeUserId = (id: number): UserId => id;
+
+/* Possible future work:
```
chrisbobbe, line 49: > Possible future work:

Makes sense. Do you think it'll always be pretty clear whether a given ID type belongs here in src/api/idTypes.js, vs. somewhere else?

I'm thinking of the return value of `keyFromNarrow`, which we've [discussed](https://github.com/zulip/zulip-mobile/pull/4339#issuecomment-745057068) making an opaque type for. I think that pretty clearly belongs right alongside `keyFromNarrow`, in narrows.js. I think I'd say it doesn't belong in src/api because we never pass those keys to the API, and we never receive them from the API. We do pass the result of `apiNarrowOfNarrow` to the API.

I think this is a sensible criterion for putting ID types in src/api/idTypes.js: if we ever talk to the API using that type of ID, its opaque type should live there. Perhaps that goes without saying, but we do a lot of passing-around of IDs within the app (including of user IDs, stream IDs, and our own narrow keys), so I wonder if a small comment might be helpful. 🙂

chrisbobbe (commented):
Great, thanks!

See also a small comment and a query, below; otherwise, this LGTM.

src/webview/js/js.js
gnprice: Indeed, thanks for the catch!

src/api/idTypes.js
gnprice: The basic principle is that `src/api/` is self-contained, so that code there shouldn't import code elsewhere -- it's describing the server API, and behaves like an independent library for doing that. We've made a few exceptions to that principle for convenience, but are generally pretty consistent about it.

So that means that any type that's going to be used in the API should be defined in the API part of the code. If we introduce a `StreamId` or `MessageId`, we'll want to use it in the API, so it'll be defined there.

The narrow-keys returned by `keyFromNarrow` don't appear in the API; they're our own invention for within the app. So they naturally wouldn't be defined in `src/api/`, and instead would appear in `narrows.js` as you say.

Because this is a more general principle I don't think it calls for a comment on this specific instance. But it'd be good to write down that general principle, which I'm not sure we have. Any suggestions on where would be most helpful?

chrisbobbe: Makes sense, thanks! I've just merged ba7108f7e5b04364bff4d2a1650f363a30bbf272, wiki-style.

gnprice: Looks good, thanks!

Issue #4422, opened 2021-01-21
Crash on attempting @-mention starting with "a" or "e"


gnprice:
This repro works on v27.158 (the current beta) and I believe on current master as well:
* Go to a compose box.
* Type `@a`, or `@e`.
* The app's UI disappears and the app just shows a white screen.

The culprit is the fake user-like data that our autocomplete code generates for the `@all` and `@everyone` mentions, and that data's interaction with `UserAvatarWithPresenceById`, which is passed a user ID and expects that user ID to belong to an actual user.

Instead, when we're handling such fake data, we should simply return an empty React element like `null` in place of the avatar.

(Also we can probably do better than writing this in terms of fake pseudo-users in the first place. But that's a bigger refactor.)

gnprice:
I'm looking through [Sentry issues that were new in v27.158](https://sentry.io/organizations/zulip/issues/?project=191284&query=is%3Aunresolved+firstRelease%3Acom.zulipmobile%4027.158%2B158&sort=user), the current beta, and the most widespread is this one:

PR #4423, opened 2021-01-21
autocomplete: Fix crash in showing avatar for pseudo-users

gnprice:
Along the way, add `tryGetUserForId` as a non-throwing variant of `getUserForId`, and use it in a couple of existing spots where we already do something equivalent.

Fixes: #4422

### DIFF
--- a/src/common/UserAvatarWithPresence.js
+++ b/src/common/UserAvatarWithPresence.js
@@ -6,7 +6,7 @@ import { createStyleSheet } from '../styles';
 import UserAvatar from './UserAvatar';
 import PresenceStatusIndicator from './PresenceStatusIndicator';
 import { AvatarURL } from '../utils/avatar';
-import { getUserForId } from '../users/userSelectors';
+import { tryGetUserForId } from '../users/userSelectors';
 import { useSelector } from '../react-redux';
 
 const styles = createStyleSheet({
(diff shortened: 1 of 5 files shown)


src/common/UserAvatarWithPresence.js
```diff
@@ -1,13 +1,13 @@
 /* @flow strict-local */
 import React, { type ComponentType, PureComponent } from 'react';
```
chrisbobbe, line 2: ESLint tells me we should remove `type ComponentType`.

src/users/UserItem.js
```diff
@@ -36,7 +36,10 @@ type Props = $ReadOnly<{|
   onPress: UserOrBot => void,
 |}>;
 
-// See UserAvatarWithPresence for discussion of this inexact object type.
+// The underlying class gets an inexact Props type to express that it's fine
+// for it to be passed extra props by the implementation of …ById.
```
chrisbobbe, line 40: IIUC, it's not clear what "…ById" refers to in this comment's new home, here, since 912b92a73e8e286711bc1e32b5afd7dc605fd43f.

chrisbobbe (commented):
Great, thanks for finding and fixing this bug! Just a few nits, below.

src/common/UserAvatarWithPresence.js
gnprice: Ah thanks, indeed.

src/users/UserItem.js
gnprice: Hmm true! In fact... there's no need for this inexact type in the first place here on `UserItem`, because its Redux-using wrapper is already using Hooks and therefore can pass exactly the desired set of props. So I think I'll just fix that in a prep commit.

gnprice:
New revision pushed!

chrisbobbe:
Merged, thanks again!

Issue #4426, opened 2021-01-22
Upgrade to RN v0.64


gnprice:
The next round after #4245. This is the first time I'm opening one of these when we're already on the previous latest release, and before the new one is even out yet!

The RN community release managers are working on the next release now, and it's anticipated in the next few weeks. Details at https://github.com/react-native-community/releases/issues/214 and particularly at https://github.com/react-native-community/releases/issues/214#issuecomment-758203239 .

There's already a release candidate, so that would be a way to get an early check on any issues we'll run into.

chrisbobbe:
We'll be jumping to Flow 0.137.0, which means (among other things) that something called ["types-first"](https://flow.org/en/docs/lang/types-first/) has been both introduced and enabled by default. We can turn it off, though, by adding `types_first=false` under `[options]` in our Flow config.

The thing we'll have to do to enable it is annotate almost all the exports of our modules:

> The caveat of this new version is that it requires exported parts of the code to be annotated with types, or to be expressions whose type can be trivially inferred (for example numbers and strings).

Not doing so for an export will raise a `signature-verification-failure` error.

It looks like we'll be able to work toward that incrementally by selecting certain directories whose exports should be checked.

They mention a [codemod](https://flow.org/en/docs/lang/types-first/#toc-a-codemod-for-large-codebases), but I suspect, if we use it at all, we'll want to use it on small sections of the code, one at a time, and check its work.

gnprice:
> We'll be jumping to Flow 0.137.0, which means (among other things) that something called ["types-first"](https://flow.org/en/docs/lang/types-first/) has been both introduced and enabled by default

Cool! The performance benefits of that won't be as big for us as they are for a giant codebase like Facebook's, but we should see some. I think the quality of the type errors may improve, too.

Re the codemod, we'll definitely want to check its work: making readable commits and reviewing them normally. I suspect we'll want to start by trying it on relatively self-contained swathes of code like `src/api/` or even `src/third/`; maybe then reducers+selectors, the other files that don't know about React, and finally swathes of React components. Depending on how much manual work is needed (and how much effort is needed to review the output), we can scale up or down how much code we cover per commit.

And if the codemod is really bad at its job, we can always ignore it and do the changes manually. But I'm hopeful that it'll be good enough, automating enough boring stuff accurately enough, that it'll be worth using.

---

FTR here's the full list of changes we'll see -- up to this point in the changelog:
[...]

PR #4433, opened 2021-01-25
flow [nfc]: Mark specific rules that are being suppressed.

chrisbobbe:
When we start using Flow v0.137+, along with React Native v0.64
(that's issue #4426), we'll be asked to specify what rule(s) we're
suppressing with each suppression comment. Conveniently, the
warnings will tell us the applicable rule(s) we need to write down.

So, having peeked ahead at the warnings seen in the RN 0.64.0-rc.2 
world, get a head start on doing so. There will probably be more of
these (we haven't yet started enforcing the requirement now, so some
unmarked `$FlowFixMe`s will naturally accumulate), but we might as
well get this chunk out of the way.

### DIFF
--- a/src/boot/replaceRevive.js
+++ b/src/boot/replaceRevive.js
@@ -57,25 +57,25 @@ function replacer(key, value) {
 
   switch (Object.getPrototypeOf(origValue)) {
     // Flow bug: https://github.com/facebook/flow/issues/6110
-    case (ZulipVersion.prototype: $FlowFixMe):
+    case (ZulipVersion.prototype: $FlowIssue):
       return { data: value.raw(), [SERIALIZED_TYPE_FIELD_NAME]: 'ZulipVersion' };
-    case (URL.prototype: $FlowFixMe):
+    case (URL.prototype: $FlowIssue):
       return { data: origValue.toString(), [SERIALIZED_TYPE_FIELD_NAME]: 'URL' };
-    case (GravatarURL.prototype: $FlowFixMe):
+    case (GravatarURL.prototype: $FlowIssue):
       return { data: GravatarURL.serialize(value), [SERIALIZED_TYPE_FIELD_NAME]: 'GravatarURL' };
-    case (UploadedAvatarURL.prototype: $FlowFixMe):
+    case (UploadedAvatarURL.prototype: $FlowIssue):
       return {
         data: UploadedAvatarURL.serialize(value),
         [SERIALIZED_TYPE_FIELD_NAME]: 'UploadedAvatarURL',
       };
-    case (FallbackAvatarURL.prototype: $FlowFixMe):
+    case (FallbackAvatarURL.prototype: $FlowIssue):
       return {
         data: FallbackAvatarURL.serialize(value),
         [SERIALIZED_TYPE_FIELD_NAME]: 'FallbackAvatarURL',
       };
-    case (Immutable.List.prototype: $FlowFixMe):
+    case (Immutable.List.prototype: $FlowIssue):
       return { data: value, [SERIALIZED_TYPE_FIELD_NAME]: 'ImmutableList' };
-    case (Immutable.Map.prototype: $FlowFixMe): {
+    case (Immutable.Map.prototype: $FlowIssue): {
       const firstKey = origValue.keySeq().first();
       return {
         data: value,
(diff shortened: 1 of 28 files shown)


src/compose/ComposeMenu.js
```diff
@@ -83,7 +83,8 @@ class ComposeMenu extends PureComponent<Props> {
       return;
     }
 
-    // $FlowFixMe sketchy falsiness check, because upstream API is unclear
+    // Upstream API is unclear.
+    // $FlowFixMe[sketchy-null-bool]
     const error: string | null = response.error || null;
```
gnprice, line 88: Let's try to keep the explanatory text after the $FlowFixMe, so between the $FlowFixMe and the code it's about. That helps keep it clear what text is there to explain the fixme, and what's there to explain something else about the code.

In particular that will help us delete the explanatory text when we delete the fixme, because of either a Flow upgrade or an improvement in the types. Otherwise it'd be easy to pretty often leave it behind because the person deleting the fixme didn't fully understand the comment at that moment, and then there'd be a puzzling comment lingering there to confuse people.

When it's too long for one line, we can use a `/* … */` comment, like in a few of our existing comments and in the docs at https://flow.org/en/docs/errors/ .

src/boot/store.js
```diff
@@ -343,7 +345,8 @@ const reduxPersistConfig: Config = {
 
   // Store data through our own wrapper for AsyncStorage, in particular
   // to get compression.
-  // $FlowFixMe: https://github.com/rt2zz/redux-persist/issues/823
+  // https://github.com/rt2zz/redux-persist/issues/823
+  // $FlowFixMe[incompatible-variance]
   storage: ZulipAsyncStorage,
```
gnprice, line 350: This is another nice illustration of how it's helpful to keep the text explaining the fixme explicitly attached to the fixme. As is, it looks like this issue link is probably there to add more context on the "Store data through … compression" comment, rather than to explain the fixme.

src/webview/html/template.js
```diff
@@ -14,7 +14,7 @@ import escape from 'lodash.escape';
  *   template`Hello $\!${&<world}` -> 'Hello $!&amp;&lt;world'
  */
 export default (strings: string[], ...values: Array<string | number>) => {
-  // $FlowFixMe github.com/facebook/flow/issues/2616
+  // $FlowFixMe[prop-missing] github.com/facebook/flow/issues/2616
```
gnprice, line 17: Separately: this change in Flow called my attention to a feature/convention that's long existed, and would be handy to use here!

> `$FlowIssue`: for a type error that you suspect is an issue with Flow

https://flow.org/en/docs/errors/

No rush to do this at the same time as adding the error codes, but probably we should use `$FlowIssue` for at least the cases where we actually link to an issue in the Flow tracker :slightly_smiling_face:

gnprice (commented):
Thanks! Will be glad to start using this feature. Comments below.

chrisbobbe:

src/compose/ComposeBox.js
```diff
@@ -449,7 +449,10 @@ class ComposeBox extends PureComponent<Props, State> {
 
     return (
       <View style={this.styles.wrapper}>
-        {/* $FlowFixMe - `MentionWarnings` should use a type-checked `connect` */}
+        {/*
+          `MentionWarnings` should use a type-checked `connect`
+          $FlowFixMe[incompatible-use]
+        */}
```
gnprice, line 455: Here's one remaining comment that gets out of order.

src/utils/__tests__/internalLinks-test.js
```diff
@@ -157,7 +157,8 @@ describe('isInternalLink', () => {
     ],
   ];
 
-  // $FlowFixMe realm_ is URL | void, but complains of out-of-bounds access
+  /* realm_ is URL | void, but complains of out-of-bounds access
+     $FlowFixMe[invalid-tuple-index] */
```
gnprice, line 161: Here's another (I think the only other.)

src/boot/replaceRevive.js
```diff
@@ -57,25 +57,25 @@ function replacer(key, value) {
 
   switch (Object.getPrototypeOf(origValue)) {
     // Flow bug: https://github.com/facebook/flow/issues/6110
-    case (ZulipVersion.prototype: $FlowFixMe):
+    case (ZulipVersion.prototype: $FlowIssue):
```
gnprice, line 60: nit in commit message:

    flow [nfc]: Better indicate flow bugs in a few suppressions.

Should read "Flow bugs".

gnprice (commented):
Thanks for the revision! Small comments below -- otherwise looks good.

src/compose/ComposeBox.js
chrisbobbe: Ah, indeed, thanks (and for the other one too).

chrisbobbe:

PR #4434, opened 2021-01-26
user: Convert more code to IDs-only, fixing warning-spew in tests

gnprice:
We've had recently a couple of warnings regularly showing up when
running tests.  Generally we should keep our test suite clean of
warning spew, partly to keep our own development experience good
and partly to help us notice new warnings so we can investigate,
in case they're telling us about potential new bugs in the app.

This PR fixes them, by converting a bit more of our code to handle
users purely by ID.  I ended up writing a surprisingly long commit message
for one of these commits, so I'll copy it here just lightly edited:

---

The two warnings are very similar; here's one of them:

```
 PASS  src/notification/__tests__/notification-test.js
  ● Console

    console.warn
      pmKeyRecipientsFromIds: missing data on user

      125 |
      126 |     if (config.enableErrorConsoleLogging) {
    > 127 |       toConsole(event);
          |       ^
      128 |
      129 |       const data = objectEntries(extras)
      130 |         .map(([key, value]) => `    ${key}: ${JSON.stringify(value)}`)

      at Object.<anonymous> (src/utils/logging.js:127:7)
      at pmKeyRecipientsFromIds (src/utils/recipient.js:211:13)
      at getNarrowFromNotificationData (src/notification/index.js:112:53)
      at Object.<anonymous> (src/notification/__tests__/notification-test.js:81:54)
```

This is a warning emitted by our own app code; it says we're trying
to look up data on a user, in order to construct a Narrow object for
a PM conversation involving them, and didn't find them in our
`state.users` data.

And the test that's causing this warning to be emitted turns out to
be... a test specifically for this error case!  In particular, it's
for where this arises as we're trying to construct the Narrow object
for a group-PM conversation a notification came from.  So it's
perfectly appropriate that we have a test exercising that case,
and it will naturally generate a warning log message.

The right way to fix the test would be to have it mock `logging.warn`
and test for the warning instead of letting it spew to the console.
We don't currently have any examples of that, but for past examples
see `git log -p -S logging src/*/__tests__/*.js`.

But we can actually do better than that: this is an edge case that no
longer needs to exist at all.  In this code, we have user IDs as
input, and we're going and looking up these whole UserOrBot objects
in the allUsersById map... but then the only thing we do with them is
pass them straight to pmNarrowFromUsers, which just picks off the
user IDs and ignores the rest of the data.

So we can simplify everything by working purely with the IDs: use a
new helper pmKeyRecipientsFromIds instead of pmKeyRecipientUsersFromIds,
and pass the result to pmNarrowFromRecipients.  That's one less giant
data structure passed around, and one less selector needed by this
code's own caller; and it means this error-case test disappears
entirely, because there is no error case.

Now, what happens if we do get a notification where this error case
would have previously shown up?  The old code would have returned
`null` from getNarrowFromNotificationData, and we would have ignored
the notification from there.  The new code will unsuspectingly return
a Narrow object encoding the list of users the notification gave it,
and we'll attempt to navigate to ChatScreen for that narrow.

(This case can absolutely happen in principle, even with a perfectly
well-behaved server: for example, perhaps a user just signed up and
immediately sent us a PM, and we haven't yet learned about them in
the event queue.  Fundamentally the reason it can happen is that
notifications are independent of the event queue, and don't
participate in its nice atomicity and ordering guarantees.)

Well, that's already a case we've always had to handle for
notifications from stream messages, where we take the stream name
verbatim from the notification without comparing against our data.
And in fact the behavior we have for it is probably a better UX than
the old behavior here: we go ahead and do the navigation, and then
the resulting `ChatScreen` consults the `isNarrowValid` selector,
which does the lookup in our data structures.  When that finds we
don't have the relevant data, `ChatScreen` displays a reasonable
error message.

---

The PR also adds the new recipient-helper in question, and then
fixes the other of the two warnings, which is a very similar story.

### DIFF
--- a/src/message/messagesActions.js
+++ b/src/message/messagesActions.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import * as NavigationService from '../nav/NavigationService';
 import type { Narrow, Dispatch, GetState } from '../types';
-import { getAuth, getAllUsersById } from '../selectors';
+import { getAuth } from '../selectors';
 import { getMessageIdFromLink, getNarrowFromLink } from '../utils/internalLinks';
 import openLink from '../utils/openLink';
 import { navigateToChat } from '../nav/navActions';
@@ -28,10 +28,9 @@ export const messageLinkPress = (href: string) => async (
 ) => {
   const state = getState();
   const auth = getAuth(state);
-  const allUsersById = getAllUsersById(state);
   const streamsById = getStreamsById(state);
   const ownUserId = getOwnUserId(state);
-  const narrow = getNarrowFromLink(href, auth.realm, allUsersById, streamsById, ownUserId);
+  const narrow = getNarrowFromLink(href, auth.realm, streamsById, ownUserId);
   if (narrow) {
     const anchor = getMessageIdFromLink(href, auth.realm);
     dispatch(doNarrow(narrow, anchor));
--- a/src/notification/__tests__/notification-test.js
+++ b/src/notification/__tests__/notification-test.js
@@ -2,7 +2,7 @@
 import deepFreeze from 'deep-freeze';
 
 import type { Notification } from '../types';
-import type { UserId, UserOrBot } from '../../api/modelTypes';
+import type { UserOrBot } from '../../api/modelTypes';
 import type { JSONableDict } from '../../utils/jsonable';
 import { getNarrowFromNotificationData } from '..';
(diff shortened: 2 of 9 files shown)


Gautam-Arora24:
I also saw these warnings in my latest commit tests

gnprice:
Yeah, they've been appearing consistently whenever you run tests that happen to include these test files. That's new -- we normally make sure to keep things in a state where you don't see any warnings unless something new has gone wrong which you should pay attention to. So this PR should take us back to that state :slightly_smiling_face:

chrisbobbe:
Merged, thanks!

PR #4437, opened 2021-01-27
redux: Provide the global state to each sub-reducer

gnprice:
There are a lot of places in our various reducers where to best do their job, they really want some data from elsewhere in our app state, outside the bit that that particular reducer maintains.

For example, several different reducers want to know when handling an EVENT_NEW_MESSAGE whether the message was sent by the self user, because they need to treat it differently in that case.  We currently accommodate this with a hack, where in `eventToAction` where we turn the server event into a Redux action, we stick in an extra property `ownUserId` that's selected straight off the current state.

We have similar hacks to provide `state.caughtUp` on the same event for the sake of some other reducers, and to provide `ownUserId` or `state.messages` on some other event actions.  These hacks aren't limited to actions for server events, either: we have `ownUserId` on `MESSAGE_FETCH_COMPLETE` actions, too.

And then there's more places where we should be using information from elsewhere in the state but haven't been, and as a result doing far more computation than should be necessary to perform some tasks that ought to be very fast.

For example, when some messages get marked as read -- a very common operation, sitting in the critical path for updating the "N unreads" banner, which the user is probably noticing right at that moment -- the "unread" reducer has to sweep through its entire data structure searching for the given message IDs so it can remove them.  If there are a lot of unreads, this can take hundreds of ms, which is painfully long. (This is #4438; it's an example of #3339 and related to #3949.) What we should be doing there is to consult `state.messages`, see exactly what conversation each message appears in, and go straight to that part of the unreads data structure and touch that and nothing else.

Why haven't we been better using the information available to us in places like that?  One factor is that it's rather a pain to add one of these hacks.  Another, probably, is that it's visibly a hack -- and perhaps even because we've felt that the Redux Way is to have each sub-reducer know nothing about the rest of the state and so there's something somehow wrong with doing so.

Well, that version of the Redux Way might be a good idea in situations where different sub-reducers are managing areas of data that are truly independent of each other, and benefit from being isolated.  But as these examples show, many or most of our state subtrees are in reality closely interconnected.  In particular they're full of IDs that refer to things in other subtrees -- messages, users, streams -- and they're naturally structured by facts about those things, facts which are maintained in those other subtrees.  So let's put that idea to rest:

* Our reducers often *should* use data from elsewhere in the state.

  If you think one should, it probably should.

And let's start providing a non-hacky way for them to do that.  (Or at least a less-hacky way.)

---

At the end of this PR, each sub-reducer is now simply passed the global state as an argument.  For the moment, most of them just ignore it.

Converting any given sub-reducer to actually use that argument will be a bit of work, just because its tests will all need to be updated to pass it.

To make an end-to-end demo, we go ahead and do this for the `unread` reducer, and use that to eliminate one use of the state-embedded-in-action hack.

Before making these changes, we need to take control of the overall structure of our main reducer, which is currently built using Redux's `combineReducers`. The initial part of the PR does that, inlining `combineReducers` and then simplifying the result, so that when we get to actually passing the global state it's a very straightforward change.


src/boot/reducers.js
```diff
@@ -61,8 +61,7 @@ function applyReducer<Key: $Keys<GlobalState>, State>(
   return nextState;
 }
 
-// Inlined just now from React upstream.
-// We'll clean this up in the next few commits.
+// Based on React upstream's combineReducers.
```
chrisbobbe, line 80: > Based on React upstream's combineReducers.

Redux upstream, I think?

src/boot/reducers.js
```diff
@@ -52,7 +52,22 @@ function applyReducer<Key: $Keys<GlobalState>, State>(
     startMs = Date.now();
   }
 
-  const nextState = reducer(state, action, globalState);
+  /* $FlowFixMe - We make a small lie about the type, pretending that
+     globalState is not void.
+
+     This is OK because it's only ever void at the initialization action,
+     and no reducer should do anythng there other than return its initial
```
chrisbobbe, line 59: > anythng

nit: "anything"

src/__tests__/lib/exampleData.js
```diff
@@ -453,15 +453,75 @@ export const makeOutboxMessage = (data: $Shape<$Diff<Outbox, {| id: mixed |}>>):
...
+/**
+ * A global Redux state, adding the given data to `plusReduxState`.
+ *
+ * This automatically includes standard example data like `selfUser` and
+ * `otherUser`, so that there's no need to add those explicitly.  See
+ * `plusReduxState` for details on what's included.
+ *
+ * See `reduxState` for a version starting from a minimal state.
+ */
+export const reduxStatePlus = (extra?: $Rest<GlobalState, {}>): GlobalState =>
```
chrisbobbe, line 522: Could you say a bit about the choice of names for `reduxStatePlus` and `plusReduxState`? They seem pretty similar.

src/__tests__/lib/exampleData.js
```diff
@@ -453,15 +453,75 @@ export const makeOutboxMessage = (data: $Shape<$Diff<Outbox, {| id: mixed |}>>):
 
 const privateReduxStore = createStore(rootReducer);
 
-/** The global Redux state, at its initial value. */
+/**
```
chrisbobbe, line 456: > example data: Add a batteries-included version of the Redux state.

Cool, sounds good! Linking some prior brainstorming about this: https://github.com/zulip/zulip-mobile/pull/4299#discussion_r518437296.

chrisbobbe (commented):
This looks great, thanks! Just some small comments below.

src/boot/reducers.js
```diff
@@ -61,7 +60,78 @@ const reducers = {
 
 export const ALL_KEYS: string[] = Object.keys(reducers);
 
-const combinedReducer: CombinedReducer<GlobalState, Action> = combineReducers(
+// Inlined just now from React upstream.
```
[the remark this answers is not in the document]
gnprice: Ah indeed, thanks :slightly_smiling_face:

src/__tests__/lib/exampleData.js
gnprice: Yeah, I don't love the names, and probably better names are possible. Happy to hear suggestions :slightly_smiling_face: 

I think having them similar to each other is pretty OK because they're basically the same underlying content with different forms of input. (In particular `eg.plusReduxState` is exactly equivalent to `eg.reduxStatePlus()`, with no argument.) So the only differences between them are the ones you see right in the source where they're being used; there's no hidden subtlety to have to keep straight, and if you just think of them as basically the same thing then that's pretty much ideal.

The "plus" is there because we'll still want to have `eg.baseReduxState` and the helper for adding stuff to it -- that one is canonical in a way that this one won't be. And because I want to communicate explicitly somehow that this is a state that has extra things added to it, things which we try to make more or less generic and reusable across our tests, but which are arbitrary and made up for the purpose of tests.

src/__tests__/lib/exampleData.js
gnprice: Ah, thanks for locating that :slightly_smiling_face: Rereading, it looks like this covers all the thoughts there, either directly or as TODO comments; so that's reassuring.

gnprice:
Pushed a new revision, with those nits fixed.

chrisbobbe:
Great, thanks! Merged.

PR #4440, opened 2021-01-27
Clean up navigation logic (2/x).

chrisbobbe:
This is the second PR in the series that began with #4428.

It converts some more components to Hooks-based components, to prepare for future work, and it also makes some preparatory changes to `ZulipStatusBar`, which I'd like to focus on in the next PR in this series. (`ZulipStatusBar`, I've found, isn't the best place to put logic that handles the safe-area view at the top; that PR will relocate that logic.)

I found that I could reasonably include a fix for #4267 in this PR, to help it land sooner. The status bar still appears in the lightbox after this PR, despite passing `hidden` to the lightbox's `ZulipStatusBar`. I'll get to the bottom of that, later, but it's orthogonal to what's really needed for fixing #4267:

1. having the lightbox header clear the unsafe area at the top (which contains the status bar, if present) when it slides down
2. having the footer clear the unsafe area at the bottom, when it slides up
3. padding the contents of the header and footer horizontally, so they don't appear in the horizontal unsafe areas while the device is in landscape mode.

(I neglected 2 and 3 in my #4268; so, I'm closing that.)

After those fixes, the lightbox looks like this on a newer iPhone (showing the image in [this message](https://chat.zulip.org/#narrow/stream/7-test-here/topic/zt1/near/1100506)). These screenshots show the state when the header and footer are visible; tapping the screen makes them slide cleanly off the screen. I've also tested on the office Android device, and on my own, older, iPhone; all works as expected.

Screen Shot 2021-01-27 at 12 14 19 PM Screen Shot 2021-01-27 at 12 14 59 PM

### DIFF
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -21,6 +21,7 @@ import { canSendToNarrow } from '../utils/narrow';
 import { getLoading, getSession } from '../directSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './narrowsSelectors';
+import { getTitleBackgroundColor } from '../title/titleSelectors';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'chat'>,
@@ -112,11 +113,13 @@ export default function ChatScreen(props: Props) {
   const sayNoMessages = haveNoMessages && !isFetching;
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
+  const titleBackgroundColor = useSelector(state => getTitleBackgroundColor(state, narrow));
+
   return (
     <ActionSheetProvider>
       <View style={[componentStyles.screen, { backgroundColor }]}>
         <KeyboardAvoider style={styles.flexed} behavior="padding">
-          <ZulipStatusBar narrow={narrow} />
+          <ZulipStatusBar backgroundColor={titleBackgroundColor} />
           <ChatNavBar narrow={narrow} editMessage={editMessage} />
           <OfflineNotice />
           <UnreadNotice narrow={narrow} />
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -59,7 +59,7 @@ export default function Lightbox(props: Props) {
       ? `Shared in #${streamNameOfStreamMessage(message)}`
       : 'Shared with you';
   const resource = getResource(src, auth);
-  const { width, height } = Dimensions.get('window');
+  const { width: windowWidth, height: windowHeight } = Dimensions.get('window');
 
   const animationProps = {
     easing: Easing.bezier(0.075, 0.82, 0.165, 1),
@@ -71,14 +71,14 @@ export default function Lightbox(props: Props) {
     <View style={styles.container}>
       <PhotoView
         source={resource}
-        style={[styles.img, { width }]}
+        style={[styles.img, { width: windowWidth }]}
         resizeMode="contain"
         onTap={handleImagePress}
         onViewTap={handleImagePress}
       />
       <SlideAnimationView
         property="translateY"
-        style={[styles.overlay, styles.header, { width }]}
+        style={[styles.overlay, styles.header, { width: windowWidth }]}
         from={-NAVBAR_SIZE}
         to={0}
         {...animationProps}
@@ -95,9 +95,9 @@ export default function Lightbox(props: Props) {
       </SlideAnimationView>
       <SlideAnimationView
         property="translateY"
-        style={[styles.overlay, { width, bottom: height - 44 }]}
-        from={height}
-        to={height - 44}
+        style={[styles.overlay, { width: windowWidth, bottom: windowHeight - 44 }]}
+        from={windowHeight}
+        to={windowHeight - 44}
         {...animationProps}
       >
         <LightboxFooter
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -60,10 +60,6 @@ class Lightbox extends PureComponent<Props, State> {
     }));
   };
 
-  handlePressBack = () => {
-    NavigationService.dispatch(navigateBack());
-  };
-
   render() {
     const { src, message, auth } = this.props;
     const footerMessage =
@@ -96,7 +92,9 @@ class Lightbox extends PureComponent<Props, State> {
           {...animationProps}
         >
           <LightboxHeader
-            onPressBack={this.handlePressBack}
+            onPressBack={() => {
+              NavigationService.dispatch(navigateBack());
+            }}
             timestamp={message.timestamp}
             avatarUrl={message.avatar_url}
             senderName={message.sender_full_name}
--- a/src/lightbox/Lightbox.js
+++ b/src/lightbox/Lightbox.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
+import React, { useState, useCallback } from 'react';
 import { View, Dimensions, Easing } from 'react-native';
 import PhotoView from 'react-native-photo-view';
 import { connectActionSheet } from '@expo/react-native-action-sheet';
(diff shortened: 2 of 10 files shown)


gnprice:
Thanks!

I think you mean #4267 as the issue this will fix.

I'll look at the code shortly, but first a couple of thoughts on those helpful screenshots:
* It looks like this causes the header and footer to get separated from the top and bottom of the screen. That seems like an odd effect. Better would be for the background of the header and footer to extend all the way to the top and bottom, respectively.
  * This is what Apple's Photos app does, for example. They have a nav bar at top and a tab bar at bottom, and they extend all the way up and down (but keep their actual controls within the safe area.)
  * Actually, more than that: when the header is in view, we should be showing the actual status bar. (With a background coordinated with our header: see https://developer.apple.com/design/human-interface-guidelines/ios/bars/status-bars/. In the Photos app, the nav bar / header is visually continuous with the status bar.) We should hide the status bar just when we're hiding the header and footer, and showing purely the image.
* This part is maybe a separate issue from #4267 but I think is also part of #3066: it looks like we're trying to show the image, the one the user is trying to look at, all the way out to the edges of the full screen, and as a result trying to show it in places like the sensor notch where nothing can show. Instead, we should be scaling the image so it fits in the area you can see.
  * ... Well, I wrote that and then I went and looked at Apple's Photos app, viewing some screenshots (which conveniently, like I guess the image in your example message, have exactly the screen's size as their natural size) -- and they show them across the full screen, just like we do.
  * Ah, but if I rotate the device into landscape mode, then the screenshot gets scaled down to fit -- it gets "pillarboxed". We should do the same, I think. It looks like we're instead doing some unhelpful cropping.

src/nav/ModalSearchNavBar.js
```diff
@@ -13,32 +12,26 @@ type Props = $ReadOnly<{|
...
+      {canGoBack && <NavBarBackButton />}
+      <SearchInput autoFocus={autoFocus} onChangeText={searchBarOnChange} />
+    </View>
+  );
 }
 
-export default ModalSearchNavBar;
+ModalSearchNavBar.defaultProps = {
+  canGoBack: true,
+};
```
gnprice, line 37: It looks like this is our first Hooks conversion involving defaultProps!

What do you think of handling this in one of the normal ways for just a function that takes some arguments? That seems like it'd be most in the spirit of Hooks. I think it'd also help make it clearer what's going on. As is, there's actually a bit of a regression in the readability of this code, I think, because the default (and the fact that this one *has* a default, and parent components aren't required to pass it) gets moved down here, far from the `Props` definition and the top of the function.

For example, we could say
```js
  const { autoFocus, searchBarOnChange, canGoBack = true } = props;
￼```
and then the `Props` type could say
```js
  canGoBack?: boolean,
```

That style would also have the added benefit that the `Props` type would now reflect the actual external interface. In the class-component world, we need to define a `Props` type that's inward-facing, reflecting the defaults already being applied, because the component's implementation may have a lot of entry points (the various methods that use `this.props`), and we need to express that they can all count on those props already being filled in, even if they're optional for callers. So we've accepted not actually having the interface written down anywhere, because the Props type is mostly close enough and the `defaultProps` are right there (at the top of the class, a few lines below the `Props` definition) and the duplication of writing both would be worse. But with a function component, there's just the one entry point (the top of the function), and we can handle filling in the defaults explicitly and do so just once.

src/lightbox/LightboxFooter.js
```diff
@@ -36,10 +37,17 @@ export default class LightboxFooter extends PureComponent<Props> {
...
     return (
-      <View style={[styles.wrapper, style]}>
+      <SafeAreaView
+        mode="padding"
+        // Why we don't do anything about the unsafe area at the
+        // bottom here: `Lightbox` ensures that the footer clears it
+        // when it slides in.
+        edges={['right', 'left']}
+        style={[styles.wrapper, style]}
+      >
```
gnprice, line 47: Ah, so perhaps adding `'bottom'` here would be the way to get the footer's padding to extend through to the bottom?

src/chat/ChatScreen.js
```diff
@@ -112,11 +113,13 @@ export default function ChatScreen(props: Props) {
...
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
+  const titleBackgroundColor = useSelector(state => getTitleBackgroundColor(state, narrow));
+
   return (
     <ActionSheetProvider>
       <View style={[componentStyles.screen, { backgroundColor }]}>
         <KeyboardAvoider style={styles.flexed} behavior="padding">
-          <ZulipStatusBar narrow={narrow} />
+          <ZulipStatusBar backgroundColor={titleBackgroundColor} />
```
gnprice, line 122: Ah, this is much nicer!

gnprice (commented):
OK, and read through the code! All looks good modulo one comment below, and my comment above about the UI.

One more thought as I reread your PR description:

> The status bar still appears in the lightbox after this PR, despite passing `hidden` to the lightbox's `ZulipStatusBar`. I'll get to the bottom of that, later, but it's orthogonal to what's really needed to fix #4267:

Hmm, does the status bar really appear in the lightbox? In your screenshot above in portrait mode, there's a status bar... but I think it is the status bar from the image being shown (which is itself a screenshot from a similar phone.) The time in it is "1:04", and that's the same as I see in that image.

As discussed above, I think the desired behavior is:
* We should show the status bar -- the real live one from the device right now :slightly_smiling_face: -- when we're showing the header and footer. I think we currently don't, and don't in your screenshots from after this PR.
* We should hide it when we're hiding the header and footer. I think we currently get this right (and from the code in the PR, I think it doesn't change that.)

src/nav/ModalSearchNavBar.js
chrisbobbe: Sounds good!

chrisbobbe:
I think I'll tackle #4267 in a separate PR after this one—I've just pushed a revision that excludes that part. 🙂

gnprice:
Looks good, thanks! Merged.

I edited the PR description first to replace "what's really needed to fix #4267" with "what's really needed for fixing #4267", so it wouldn't accidentally cause that issue to be closed :slightly_smiling_face:

src/lightbox/Lightbox.js
```diff
@@ -95,9 +95,9 @@ export default function Lightbox(props: Props) {
       </SlideAnimationView>
       <SlideAnimationView
         property="translateY"
-        style={[styles.overlay, { width, bottom: height - 44 }]}
-        from={height}
-        to={height - 44}
+        style={[styles.overlay, { width: windowWidth, bottom: windowHeight - 44 }]}
+        from={windowHeight}
+        to={windowHeight - 44}
```
gnprice, line 100: I also notice, thanks to this cleanup calling my attention to it, that this longstanding bit of code is all unnecessarily complicated -- if you subtract `windowHeight` from all three of these vertical values, it gets a lot simpler. Maybe simpler still if you subtract `windowHeight - 44`, so that it's:
* `bottom: 0`
* `from={44}`
* `to={0}`

But that might all be made moot by the changes you're [already working on](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60View.60's.20.60onLayout.60.20prop/near/1110028) to replace this `translateY` animation entirely.

chrisbobbe: I've got a commit that does exactly that simplification; about to send the PR. 😄

karanb1:
@chrisbobbe Can I help you with the cleanup and in conversion of component to hooks based components?
I can start with subscriptions and user status screens.

chrisbobbe:
Sure, @karanb1, thanks! Several of our components don't have any lifecycle methods beyond `render`, so they're low-hanging fruit for Hooks conversions. For one that's even a little bit complicated, it would be helpful to handle it in multiple commits, as you can see I've done in this PR with `Lightbox`. I'd stay away from the really complicated components like `ComposeBox`, for now; there's a good chance that some subtlety of how things work isn't clearly documented, which makes it unfortunately easy to introduce bugs in a refactor like this. If you'd like to follow or join the discussion about converting to Hooks, it's [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks). 🙂

chrisbobbe:
(Just split off some work on the lightbox in #4442.)

src/nav/ModalSearchNavBar.js
gnprice: For cross-reference: I've now added this to the style guide:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#react-function-prop-defaults
https://github.com/zulip/zulip-mobile/commit/316070448a9d168972c3be47121c1e51854df5bd

chrisbobbe: Thank you!

Issue #4217, opened 2020-08-04
Replace outdated `react-native-photo-view` library.


chrisbobbe:
In d873e22fc, we checked out a commit on this project's `master` branch that hadn't been released at the time, which was September 21, 2018. It still hasn't been released—in fact, there haven't been any commits after that date.

Right now, if I do `yarn upgrade && yarn yarn-deduplicate && yarn`, Android fails to build (looks like it's something to do with the Android Support Library; it reminds me of 01cf59800 and 48ef40d98).

Unfortunately, I haven't found an obvious choice for a replacement lightbox library. Here are some candidates:

- `react-native-lightbox` ([GitHub](https://github.com/oblador/react-native-lightbox), [npm](https://www.npmjs.com/package/react-native-lightbox)). I recognize the author's name, at least—they run `react-native-vector-icons`. Looks like pinch-to-zoom isn't supported (though there's https://github.com/oblador/react-native-lightbox/issues/5 from 2015 for it), and it's not super frequently maintained.
- Brent Vatne (who's quite active in the React Native ecosystem) says, on an [Expo feature request page](https://expo.canny.io/feature-requests/p/pinchzoomphotoview), that `react-native-reanimated` ([GitHub](https://github.com/software-mansion/react-native-reanimated), [npm](https://www.npmjs.com/package/react-native-reanimated)) can be used for a lightbox with pinch-and-zoom. `react-native-reanimated` v2 is a reimplementation of the [RN Animated library](https://reactnative.dev/docs/animated) entirely in JavaScript ([blog post](https://blog.swmansion.com/introducing-reanimated-2-752b913af8b3): "Therefore, we wanted for the next Reanimated to allow for building animations and interactions using just JavaScript."). I tried out the [lightbox example](https://github.com/software-mansion/react-native-reanimated/blob/master/Example/src/LightboxExample.js) on the iOS simulator, following the [instructions](https://github.com/software-mansion/react-native-reanimated/blob/master/Example/README.md), and pinch-and-zoom didn't work. It animated from a small thumbnail into a size that covered the screen, but once it was there, I couldn't zoom in at all (you can use the option key to do a pinch gesture in the simulator).

[...]

PR #4442, opened 2021-01-28
Put lightbox header/footer text and buttons in the safe area.

chrisbobbe:
This branched off of #4440 when the fix became a bit more complicated than what was mainly being done in that PR. 🙂 

I've addressed some of Greg's points there:

@gnprice [said](https://github.com/zulip/zulip-mobile/pull/4440#issuecomment-768603140):
> It looks like this causes the header and footer to get separated from the top and bottom of the screen. That seems like an odd effect. Better would be for the background of the header and footer to extend all the way to the top and bottom, respectively.

Yeah, you're right. I'd hesitated to obscure more of the image, which is why I had just those bands, instead of extending the space through the whole top and bottom. But then, yeah—it doesn't really matter at all if we obscure that bit of the image, since the user can just tap, and the header and footer will go away completely. 🙂

> Ah, but if I rotate the device into landscape mode, then the screenshot gets scaled down to fit -- it gets "pillarboxed". We should do the same, I think. It looks like we're instead doing some unhelpful cropping.

Yeah. Unfortunately, that doesn't get fixed in this PR; I haven't quite found the right way to do it. There's a `resizeMode` prop that I add a code comment on, saying that it basically doesn't work at all on iOS. It's not fixed in `react-native-photo-view-ex`, which you pointed out at https://github.com/zulip/zulip-mobile/issues/4217#issuecomment-668913786. I notice that if the device is already in landscape mode when you open the lightbox, it shows the whole image, as we'd want it to:

But then, from there, if you go into portrait mode, the image is incorrectly small:

I suspect the native component gets initialized with proper dimensions but doesn't bother to properly update its dimensions after that.

I tried a React `key` hack to unmount and remount the `PhotoView` component when the orientation changes. That worked for maybe 75% of rotations, but the rest of the time it wouldn't unmount-remount upon rotation, instead updating when I tapped the screen, despite the `key` actually changing. Not sure what's going on there. Also I'm pretty sure the unmount-remount means the image gets re-downloaded.

Anyway, even with the improper framing when you're in landscape mode, it is possible to scroll so that you can see all of the image; it just doesn't all show at once.

> Hmm, does the status bar really appear in the lightbox?

Ah, I think it might only be showing up after a fast-refresh (from when save changes to a file), and those don't happen in production.

> * We should show the status bar -- the real live one from the device right now 🙂 -- when we're showing the header and footer. I think we currently don't, and don't in your screenshots from after this PR.
> * We should hide it when we're hiding the header and footer. I think we currently get this right (and from the code in the PR, I think it doesn't change that.)

I'll tackle this after some more work in the clean-up-navigation PRs. Currently, `ZulipStatusBar` is improperly doing the job of protecting content from the unsafe areas, with a sized `View`, so inserting it or removing it somewhere will unhelpfully alter the layout. I've got a draft that takes that `View` out of `ZulipStatusBar`.

Here are the screenshots after the changes in this PR:

landscape
portrait

I also tested on Android, and my iPhone 6s, and all seems fine.

Fixes: #4267


src/lightbox/Lightbox.js
```diff
@@ -80,7 +80,7 @@ export default function Lightbox(props: Props) {
           styles.overlay,
           styles.header,
           { width: windowWidth },
-          movement === 'out' ? { top: 0 } : { bottom: windowHeight },
```
gnprice: Oh wow, `'out'` meant they were *visible*? That sure was confusing.

src/lightbox/Lightbox.js
```diff
@@ -76,6 +76,8 @@ export default function Lightbox(props: Props) {
         //   https://github.com/alwx/react-native-photo-view/issues/98
         // TODO: Figure out how to make it work.
         resizeMode="contain"
+        showsHorizontalScrollIndicator={false}
+        showsVerticalScrollIndicator={false}
```
gnprice, line 80: Lightbox: Remove scrollbars when panning and zooming on iOS.
    
    I don't think these really add anything.

Agreed.

From trying the current beta on Android, I think the "on iOS" means that currently we have scrollbars on iOS but not Android, and after this change we don't have them on either platform. Does that agree with what you have in mind? Would be good to make explicit -- otherwise it sounds like perhaps we still have them but only on Android.

src/lightbox/Lightbox.js
```diff
@@ -71,6 +71,10 @@ export default function Lightbox(props: Props) {
       <PhotoView
         source={resource}
         style={[styles.img, { width: windowWidth }]}
+        // Doesn't seem to do anything on iOS:
+        //   https://github.com/alwx/react-native-photo-view/issues/62
+        //   https://github.com/alwx/react-native-photo-view/issues/98
+        // TODO: Figure out how to make it work.
         resizeMode="contain"
```
gnprice, line 78: Huh, I see! Indeed, when I try on Android (with the current beta) I get exactly the "pillarboxed" behavior that I'd hope for. It's just on iOS that I don't, and the image stays full size.

Hmm. I've been doing this testing mainly with an image that's a screenshot from a similar phone, so its natural size is the screen size. What if the image's natural size is much larger, or smaller? Does this issue in PhotoView mean that on iOS it'll stay too large, or too small?

src/lightbox/Lightbox.js
```diff
@@ -59,6 +59,11 @@ export default function Lightbox(props: Props) {
       ? `Shared in #${streamNameOfStreamMessage(message)}`
       : 'Shared with you';
   const resource = getResource(src, auth);
+
+  // Since we're using `Dimensions.get` (below), we'll want a rerender
+  // when the orientation changes. No need to store the value.
+  useSelector(state => getSession(state).orientation);
```
gnprice, line 65: Oh, I think I see this as a bug in the existing implementation too!

Start in portrait mode, open the lightbox, and then rotate to landscape. The image stays limited to a pillarbox, even if the image's natural aspect ratio is wider. And the lightbox header and footer stay narrow, too:
image

I'd noticed that a few minutes ago, and was going to mention it as another small design issue that'd be good to fix while we're working on this -- the header and footer should go the whole width of the screen. Then I realized that it's caused by the bug this change fixes :slightly_smiling_face: 

Not sure if I'd ever spotted that before; if I had, I must have put it down to a suboptimal design choice then too.

Conversely, if you start in landscape and rotate to portrait, everything stays wide, and the header and footer are unusable because the controls are off-screen. That must be this same bug. (And, even before seeing this code, would be a signal that it's an outright bug and not a dubious design choice.)

src/lightbox/Lightbox.js
```diff
@@ -49,6 +49,7 @@ export default function Lightbox(props: Props) {
 
   // Pulled out here just because this function is used twice.
   const handleImagePress = useCallback(() => {
+    LayoutAnimation.configureNext(LayoutAnimation.create(100, 'easeInEaseOut', 'opacity'));
```
gnprice, line 52: It'd be good to be clear that this doesn't just reimplement the old animation -- it replaces it with a different animation.

In particular the old animation had the header and footer slide in from the top and bottom, while this one has them just fade in.

It's a shame that (AFAICT) this LayoutAnimation API doesn't seem to support translation at all -- I think it would not be possible to implement an animation like the old one using this API.

I kind of like the slide-in/out animation. But it looks like both Google's and Apple's "Photos" apps have the top and bottom bars fade in and out, with no sliding; so even though I'm not sure I like that quite as much, it's clearly a perfectly fine way to do it.

gnprice: ... Hmm, and now I look back at your (animated) screenshots, and it looks like it still is doing slide-in/out! I like the effect -- but I sure am confused about what that `'opacity'` argument is doing. According to the doc:
https://reactnative.dev/docs/layoutanimation
it should be "the layout property to animate"; but in your screencaps it looks like `top` and `bottom`, or really the overall position that they together work out to, is animating, and `opacity` isn't at all.

gnprice (commented):
Thanks @chrisbobbe ! Comments below.

> I'll tackle [the status bar] after some more work in the clean-up-navigation PRs. Currently, `ZulipStatusBar` is improperly doing the job of protecting content from the unsafe areas, with a sized `View`, so inserting it or removing it somewhere will unhelpfully alter the layout. I've got a draft that takes that `View` out of `ZulipStatusBar`.

Sounds like a good plan to me!

src/lightbox/Lightbox.js
chrisbobbe: Right -- these are iOS-only props. I'll make a note.

src/lightbox/Lightbox.js
chrisbobbe: There's something else that looks wrong in that screenshot: there's too much blank dark space at the *top* of the footer, and at the *bottom* of the header. I'm not seeing that; is it possible you had made a code adjustment that was causing that? Or perhaps it's this same bug (though I don't really see how)?

src/lightbox/Lightbox.js
chrisbobbe: Huh, yeah, odd—I actually don't see anything happening with the opacity. I took `Presets.easeInEaseOut` and just adjusted the duration from 300 to 100, since it seemed kind of long. (The new revision will be clearer about this.)

chrisbobbe: Not sure what `Presets.easeInEaseOut` was doing with 'opacity'.

src/lightbox/Lightbox.js
chrisbobbe: Here are some screenshots from the iOS simulator, for [an extremely small image](https://chat.zulip.org/#narrow/stream/7-test-here/topic/pictures/near/1111014), and [an extremely large image](https://chat.zulip.org/#narrow/stream/7-test-here/topic/pictures/near/1111013). I've included screenshots of portrait and landscape mode, but it's important to note that, in both cases, I started out in portrait mode, since that seems to make a difference.

Screen Shot 2021-01-29 at 3 33 32 PM
Screen Shot 2021-01-29 at 3 33 22 PM
Screen Shot 2021-01-29 at 3 33 07 PM
Screen Shot 2021-01-29 at 3 32 57 PM

chrisbobbe: I think the native component sets its dimensions when it first initializes, but it doesn't update them at all afterward.

chrisbobbe:
Thanks for the review, revision pushed! I made another change; I've started using RN v0.63's new [`Pressable` API](https://reactnative.dev/docs/pressable).

I've had some luck viewing touchable areas by doing the following:

- shake
- "Show Inspector"
- "Touchables" (then the app annoyingly reloads, taking me away from the lightbox)
- shake
- "Hide Inspector"
- go back to the lightbox

In this screenshot, it's extremely faint, but there's a red dotted-line square around the "x" and the three-dots button, showing the area of the touch target:

image

src/lightbox/Lightbox.js
gnprice: > I think the native component showing the image sets its dimensions when it first initializes, but it doesn't update them at all afterward.

Cool, yeah, that seems to be what's happening. So it appropriately scales the image at the start, but doesn't update that when the component gets resized.

src/lightbox/LightboxHeader.js
```diff
@@ -73,9 +73,11 @@ export default class LightboxHeader extends PureComponent<Props> {
               {subheader}
             </Text>
           </View>
-          <Touchable style={styles.rightIconTouchTarget} onPress={onPressBack}>
```
gnprice: On iOS, the "x" button in the header was putting a brand-colored
    rectangle around the icon when it was pressed, which seemed jarring
    and artificial to me.

Hmm, yeah.

I actually reproduce that on Android too, now that I look closely. I think it's less visible there, because the animation that takes us back to the message list is quicker to get it out of view. On Android we get the standard Android "fade down" animation for returning to a previous screen, and on iOS we get the standard iOS "slide out right" animation, and I guess the latter has a slower start.

src/lightbox/LightboxFooter.js
```diff
@@ -39,7 +41,11 @@ export default class LightboxFooter extends PureComponent<Props> {
       <SafeAreaView mode="padding" edges={['right', 'bottom', 'left']}>
         <View style={[styles.wrapper, style]}>
           <Text style={styles.text}>{displayMessage}</Text>
-          <Icon style={styles.icon} color="white" name="more-vertical" onPress={onOptionsPress} />
+          <Pressable style={styles.iconTouchTarget} onPress={onOptionsPress} hitSlop={10}>
```
gnprice: Neither button was giving the touch target the minimum 48 x 48 px
    size recommended by Material Design [1]; so, do that with the
    `hitSlop` prop [2]. Passing one number like this adds that amount of
    additional distance outside the element in which a press can be
    detected. So, with a font size of 28, the new width and height of
    the pressable area are approx. 10 + 28 + 10 = 48.

One thing I see in that handy Pressable doc is this:

> The touch area never extends past the parent view bounds and the Z-index of sibling views always takes precedence if a touch hits two overlapping views.

So this `hitSlop` will only take effect to the extent that there's that much room around this view within its parent view. Is that the case here? We might need to arrange some margin here, or padding in the parent, to make sure of that.

src/lightbox/LightboxFooter.js
```diff
@@ -20,9 +20,11 @@ const styles = createStyleSheet({
     fontSize: 14,
     alignSelf: 'center',
   },
+  iconTouchTarget: {
+    alignSelf: 'center',
+  },
   icon: {
     fontSize: 28,
```
gnprice: detected. So, with a font size of 28, the new width and height of
    the pressable area are approx. 10 + 28 + 10 = 48.

It is kind of odd that we're specifying this with something called `fontSize`. There's not really a font involved here. And it makes me wonder if this size is getting scaled by the user's font-size setting -- which would be the wrong behavior, because that setting should only affect actual text that the user may be trying to read as text.

OK, a quick test finds that it isn't (as of v27.158, current beta), neither on Android nor iOS. If I crank the system font-size setting way up, the text in the header and footer get bigger (as they should), but the icons stay the same size (as they should.)

Still, it'd be good if we can avoid that confusion. And as your "approx." suggests, it's not entirely clear that `fontSize` of 28 means a size of 28px.

It looks like in other places where we create an icon, we tend to use its own `size` prop, like `size={24}`, instead of using the `style` prop. If we do that here (with the same number as for `fontSize`), does that keep the behavior unchanged?

src/lightbox/LightboxFooter.js
gnprice: Ah, I think your screenshot in the comment above may answer this question. It looks like we're probably in good shape except for the top side of the three-dots button.

It'd be good to add some padding or margin to give that target its full size, and perhaps also to structure things so that it's clear they both do. Looking at the header code, I think that one is getting enough room only because the avatar holds up the header's height. Perhaps add a `margin` to each `Pressable`'s style, equal to its `hitSlop`?

gnprice (commented):
Thanks for the revision! Some comments below on the new `Pressable` commit; otherwise all LGTM.

src/lightbox/LightboxFooter.js
chrisbobbe: > Perhaps add a `margin` to each `Pressable`'s style, equal to its `hitSlop`?

image

Hmm, this feels like too much vertical space. The `margin` I've added gets added to the `paddingVertical: 8` in `styles.wrapper`.

chrisbobbe: (I suppose the same is happening with the horizontal padding, too.)

gnprice: Ah, indeed.

I think it'd be fine to add margin that only supplements the parent's padding. We want 12 total space on each side, so where the parent's padding is 16 that's enough by itself, but where it's 8 we want a margin of 4.

chrisbobbe:

gnprice:
Looks good! Merged.

gnprice:
(For cross-reference: this fixed part of #3066.)

PR #4443, opened 2021-01-29
Clean up navigation logic (3/x).

### LINKED ISSUE #3066: Safe areas are not handled correctly on iPhone X and Androids with notches

chrisbobbe:
Following #4428 and #4440, this PR takes the job of handling the top inset away from `ZulipStatusBar`.

It doesn't yet establish a consistent scheme for how we should handle safe areas (for #3066 and put a ratchet on a fix for it), but it brings us closer to being able to do that, and I aim to do that in a future PR. 🙂

My goal for that scheme is this:

- Each screen on a stack navigator should use its [`header` prop](https://reactnavigation.org/docs/stack-navigator/#header) to either
  - Show a custom, non-trivial header, e.g., with nav buttons, etc., that avoids the top inset. Its content should also be padded on the left and right to avoid the horizontal insets when in landscape mode. (`ChatNavBar`, `ModalNavBar`, etc., can pretty much be dropped in here after this PR.) Or,
  - Fill the top inset, and that's it.
  - (React Navigation is happy to provide a default, batteries-included header that handles safe areas and shows the current route name and a back button, etc.—but we've so far avoided that, and the default styling doesn't really work with our app. Still, even with custom headers, I'm expecting the use of the `header` prop to help make consistency across our headers more natural than it is now.)
  - Special notes about `MainTabsScreen`:
    - `MainTabsScreen` itself is a screen on a stack navigator, `AppNavigator`. It'll get the latter treatment; just a simple `View` that covers the top inset.
    - The tab screens directly *on* `MainTabsScreen` [aren't configurable with a `header` prop](https://reactnavigation.org/docs/bottom-tab-navigator/). They fill the entire area between the bottom tab bar and `MainTabsScreen`'s trivial space-covering header (just mentioned). So they just need to worry about their horizontal padding.
    - The bottom tab bar already avoids the bottom inset because of some logic under the hood, I think.
    - If we were planning to [put a stack navigator on each tab](https://reactnavigation.org/docs/tab-based-navigation/#a-stack-navigator-for-each-tab) (or some tabs), we'd have to think about double headers stacking on each other: (1) `MainTabsScreen`'s header, and (2) the header for each screen on such a stack navigator. (Likely meaning handling the top inset *twice*, which is one too many times.) But we're [not currently planning to use this pattern](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Stack.20nav.20for.20settings.20tab/near/1107642).
- Each screen component on a stack navigator is responsible for keeping important content out of the right, bottom, and left insets. It doesn't have to worry about the top inset because its header does that, as mentioned above.

Linking React Navigation's [doc on handling safe areas](https://reactnavigation.org/docs/handling-safe-area/).


gnprice:
(I'll look at the actual content here in a bit, but:

> It doesn't yet […] fix #3066

this is a classic example of the kind of phrasing that trips up GitHub's "this is supposed to fix that" detector :wink:. Should reword to either not say "fix", or squeeze a word in between "fix" and the issue number.)

chrisbobbe:
Ah, right; thanks for catching that.

src/start/LoadingScreen.js
```diff
@@ -27,8 +30,21 @@ type Props = $ReadOnly<{|
...
     <View style={componentStyles.center}>
+      {orientation === 'PORTRAIT' && (
+        <View
+          style={{
+            height: statusBarHidden ? 0 : insets.top,
+            backgroundColor: 'black',
+          }}
+        />
+      )}
       <ZulipStatusBar backgroundColor={BRAND_COLOR} />
```
gnprice: Did you mean `backgroundColor: BRAND_COLOR` here?

src/main/MainTabsScreen.js
```diff
@@ -67,14 +66,7 @@ export default function MainTabsScreen(props: Props) {
 
   return (
     <View style={[styles.flexed, { backgroundColor }]}>
-      {orientation === 'PORTRAIT' && (
-        <View
-          style={{
-            height: insets.top,
-            backgroundColor: DEFAULT_TITLE_BACKGROUND_COLOR,
```
gnprice: The value of this constant is `'transparent'`, which is the default
    background color of `View`s anyway.

It looks like after this change a further simplification is possible: I believe it's the case that `DEFAULT_TITLE_BACKGROUND_COLOR` is only ever used as a sentinel value! That is, it signifies to our own code "there's nothing here, just use your default behavior", and we have conditionals that test for equality to it... but I believe we never pass it as an actual style prop, or otherwise to any code outside our own.

If so, that means that we could replace the constant's initializer with `undefined`, or for that matter `{}` or anything that makes a fresh unique object. And probably better -- we could just replace every use of it with `undefined`, and augment the relative types with `| void` to make that possibility explicit.

src/chat/ChatScreen.js
```diff
@@ -116,21 +116,18 @@ export default function ChatScreen(props: Props) {
...
 
   const titleBackgroundColor = useSelector(state => getTitleBackgroundColor(state, narrow));
-  const orientation = useSelector(state => getSession(state).orientation);
   const insets = useSafeAreaInsets();
 
   return (
     <ActionSheetProvider>
       <View style={[componentStyles.screen, { backgroundColor }]}>
         <KeyboardAvoider style={styles.flexed} behavior="padding">
-          {orientation === 'PORTRAIT' && (
```
gnprice: I can't think of a device that has a nonzero top inset in landscape
    mode, so it's possible that this commit is NFC. Still, we might as
    well (1) handle that case, so we don't have to think about it, and
    (2) remove something that can cause confusion.

Hmm, maybe a foldable? Or some odd-shaped tablet?

In any case I definitely agree with this reasoning.

src/nav/ChatNavBar.js
```diff
@@ -31,36 +32,46 @@ export default function ChatNavBar(props: Props) {
       ? 'default'
       : foregroundColorFromBackground(backgroundColor);
 
+  const insets = useSafeAreaInsets();
+
   return (
-    <View
```
gnprice, line 35: (The default
    algorithm used to display the diff in `ChatNavBar` gives a lot of
    noise; `git diff -w` should reduce that quite a lot.)

It will, but perhaps this is a good time to plug a handy Git feature! This is one I relatively recently discovered and enabled in [my gitconfig](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3):
```
# Use color to identify code that was moved around verbatim.
[diff]
	# Enable the feature.
	colorMoved = zebra

	# Supercharge it to do the equivalent of `git diff -b`.
	# …
	colorMovedWS = ignore-space-change
```
(And `git diff -b` is a close relative of `-w`; they're pretty similar but I tend to find `-b` is closer to what I'm really looking for in principle.)

So here's how the main part of that diff comes out with my normal `git usp`, without going back and adding any other flags:

image

The "moved" colors let me see immediately that the code in each of those chunks didn't change beyond getting indented. Then I can match up the chunks by eye, and pretty quickly isolate what's changed. Not as simple as the `git diff -b` or `-w` view, but simple enough that I usually don't feel the need anymore to go reach for that one.

(The default colors are different from those and I think less helpful; those colors are also in my posted gitconfig.)

src/settings/SettingsScreen.js
```diff
@@ -52,7 +51,6 @@ class SettingsScreen extends PureComponent<Props> {
 
     return (
       <ScrollView style={styles.optionWrapper}>
-        <ModalNavBar canGoBack={false} title="Settings" />
```
gnprice, line 55: Oh gosh this was inside the ScrollView! That never made any sense.

Which... oh, I see, this was quite noticeable on iOS.

On Android it could only come up if there was more there than fit vertically. I can make that happen by cranking up the system "Display size" setting plus going into landscape; but otherwise it's pretty unlikely, because we only have a few items on this settings screen.

But iOS has the animation where you can tug the scrollable content around even when there's no scrolling to be done, and the "Settings" header would move right along.

src/nav/ModalNavBar.js
```diff
@@ -24,17 +24,14 @@ export default function ModalNavBar(props: Props) {
...
+    <View style={{ borderBottomWidth: 1, backgroundColor, paddingTop: insets.top }}>
       <View
         style={[
           {
             borderColor: 'hsla(0, 0%, 50%, 0.25)',
             flexDirection: 'row',
             height: NAVBAR_SIZE,
             alignItems: 'center',
-            borderBottomWidth: 1,
-            backgroundColor,
```
gnprice: Hmm, is this quite right? The `borderColor` is getting separated from the `borderBottomWidth`.

For that matter: is there a reason not to collapse these into a single wrapper View? I don't think I'm spotting a way that we're using the distinction.

gnprice: I guess the next commit turns the outer one into a `SafeAreaView`. But that's supposed to be just like a `View` with one extra feature, so I think the same question applies.

gnprice (commented):
Thanks @chrisbobbe! Glad to see all these things straightened out. Some comments below.

> My goal for that scheme is this:

This plan sounds good to me.

src/main/MainTabsScreen.js
chrisbobbe: Hmm, interesting! I think I see just one place where we pass it to code that's not ours, at this commit:

src/chat/ChatScreen
```js
  const titleBackgroundColor = useSelector(state => getTitleBackgroundColor(state, narrow));
  // …
  return (
    <ActionSheetProvider>
      <View style={[componentStyles.screen, { backgroundColor }]}>
        <KeyboardAvoider style={styles.flexed} behavior="padding">
          {orientation === 'PORTRAIT' && (
            <View
              style={{
                height: insets.top,
                backgroundColor: titleBackgroundColor,
              }}
            />
          )}
```

`getTitleBackgroundColor` will return `DEFAULT_TITLE_BACKGROUND_COLOR` if the narrow is not a stream or topic narrow.

src/nav/ChatNavBar.js
chrisbobbe: Hmm, interesting! 

> Not as simple as the `git diff -b` or `-w` view, but simple enough that I usually don't feel the need anymore to go reach for that one.

It is quite a lot more lines than `git diff -b`. I'm not sure how long it'll take until I see this output as simpler, but I'll give it a go and find out! 🙂

src/nav/ModalNavBar.js
chrisbobbe: > is there a reason not to collapse these into a single wrapper View? I don't think I'm spotting a way that we're using the distinction.

The existing `View` sets an explicit height of `NAVBAR_SIZE` (58), so the part that holds the content gets shrunk. Though I guess we could get rid of that explicit height; we recently removed one in #4442.

chrisbobbe: I think the situation is similar with `ModalSearchNavBar`.

src/main/MainTabsScreen.js
chrisbobbe: Ah, right—but passing `undefined` would be just the same as passing 'transparent' there.

chrisbobbe:

src/main/MainTabsScreen.js
chrisbobbe: My new revision makes `DEFAULT_TITLE_BACKGROUND_COLOR` go away. I've also renamed `getTitleBackgroundColor` to `getStreamColorForNarrow`, but I've kept the part of its interface that says it accepts narrows other than stream or topic narrows (for which it returns `undefined`). The `react-hooks/rules-of-hooks` rule doesn't like it if I conditionally use a hook, as in

```
const backgroundColor = isStreamNarrow(narrow)
    ? useSelector(state => getStreamColorForNarrow(state, narrow))
    : '…';
```

```
React Hook "useSelector" is called conditionally. React Hooks must be called in the exact same order in every component (render.eslintreact-hooks/rules-of-hooks)
```

src/nav/ChatNavBar.js
gnprice: To be clear, the output of `git diff -b` is definitely simpler than this!

I just mean that this is simpler than reading the plain diff (and having to look closely to spot the few bits that changed); and it's enough simpler, giving me enough of what I'd get from `git diff -b`, that when I'm in the middle of a review I typically don't feel the need to go reach for `git diff -b`.

(And I don't just have `-b` or `-w` on all the time, because I do want to know about the changes those ignore; when they're relevant, I want to both see those changes, and see the changes with those simplified away, just separately.)

src/nav/ModalNavBar.js
gnprice: Ah I see, because the padding comes out of the `height`.

We could also handle that directly, I think, by adding the desired padding to the height when combining these two `View`s:
```js
  paddingTop: insets.top,
  height: insets.top + NAVBAR_SIZE,
```

But removing NAVBAR_SIZE sounds good in any case.

src/nav/ModalNavBar.js
```diff
@@ -24,24 +24,22 @@ export default function ModalNavBar(props: Props) {
...
+        {
+          borderColor: 'hsla(0, 0%, 50%, 0.25)',
+          flexDirection: 'row',
+          alignItems: 'center',
+          borderBottomWidth: 1,
+          backgroundColor,
+          paddingTop: insets.top,
+        },
+      ]}
+    >
```
gnprice: This commit is marked NFC, but I think it does change something: it causes `backgroundColor` to extend up through the top inset. Right?

I think that's a good change -- just want to be sure I'm not missing something.

src/nav/ModalNavBar.js
gnprice: > We could also handle that directly, I think, by adding the desired padding to the height when combining these two `View`s:
> 
> ```js
>   paddingTop: insets.top,
>   height: insets.top + NAVBAR_SIZE,
> ```

Oh, but this gets stickier when converting to `SafeAreaView`, doesn't it.

Anyway, happy to remove `NAVBAR_SIZE` :slightly_smiling_face:

src/title/titleSelectors.js
```diff
@@ -4,12 +4,11 @@ import { isStreamOrTopicNarrow, streamNameOfNarrow } from '../utils/narrow';
...
 /**
- * Background color to use for the app bar in narrow `narrow`.
+ * The stream's color for the given stream or topic narrow.
  *
- * If `narrow` is a stream or topic narrow, this is based on the stream color.
- * Otherwise, it takes a default value.
+ * Gives undefined for narrows that are not stream or topic narrows.
  */
-export const getTitleBackgroundColor = (state: GlobalState, narrow: Narrow) => {
+export const getStreamColorForNarrow = (state: GlobalState, narrow: Narrow) => {
```
gnprice: With its new clearer name, I think this also is best moved to a different file: `subscriptionSelectors.js`, where it will fit right in.

That also lets us remove `titleSelectors.js`, which never made as much sense as the selectors files like `subscriptionSelectors.js` that are clearly about a particular part of the app's data model.

src/common/ZulipStatusBar.js
```diff
@@ -46,7 +45,7 @@ type Props = $ReadOnly<{
 class ZulipStatusBar extends PureComponent<Props> {
   static defaultProps = {
     hidden: false,
-    backgroundColor: DEFAULT_TITLE_BACKGROUND_COLOR,
+    backgroundColor: undefined,
```
gnprice, line 44: This can also be dropped (probably in a small followup commit, like you've done already for similar cleanups.)

gnprice (commented):
Thanks for the revision! I like the new further cleanup. Small comments above and below.

chrisbobbe:

gnprice:
All looks good, thanks! Merged.

Issue #4449, opened 2021-01-30
Message view gives error on malformed image URL


gnprice:
Recently I've occasionally been seeing a red error banner at the top of the message list (the one that says "Oh no! Something went wrong. Try again?".) I've only seen it when looking at `#test here` on czo -- so presumably there's something exciting happening in some message there that triggers whatever bug is involved.

Then yesterday I was looking at `adb logcat` output in order to do some performance measurements (for #4448), and I hit the same symptom and saw this in the log:
```
01-28 22:00:15.545 23892 23927 E ReactNativeJS: WebView error
01-28 22:00:15.553 23892 23927 E ReactNativeJS:     message: "Uncaught TypeError: Failed to construct 'URL': Invalid URL"
01-28 22:00:15.553 23892 23927 E ReactNativeJS:     source: "file:///android_asset/webview/index.html"
01-28 22:00:15.553 23892 23927 E ReactNativeJS:     line: 263
01-28 22:00:15.553 23892 23927 E ReactNativeJS:     column: 22
01-28 22:00:15.553 23892 23927 E ReactNativeJS:     error: {}
01-28 22:00:15.554 23892 23927 E ReactNativeJS: index.android.bundle:559:1322
01-28 22:00:15.554 23892 23927 E ReactNativeJS: handleWebViewOutboundEvent@index.android.bundle:2170:2490
01-28 22:00:15.554 23892 23927 E ReactNativeJS: handleMessage@index.android.bundle:2143:2114
01-28 22:00:15.554 23892 23927 E ReactNativeJS: onMessage@index.android.bundle:2145:5068
01-28 22:00:15.554 23892 23927 E ReactNativeJS: __callFunction@index.android.bundle:27:3751
01-28 22:00:15.554 23892 23927 E ReactNativeJS: index.android.bundle:27:812
01-28 22:00:15.554 23892 23927 E ReactNativeJS: __guard@index.android.bundle:27:2958
01-28 22:00:15.554 23892 23927 E ReactNativeJS: callFunctionReturnFlushedQueue@index.android.bundle:27:784
01-28 22:00:15.554 23892 23927 E ReactNativeJS: callFunctionReturnFlushedQueue@[native code]
```

Specifically, I'd just narrowed to [the topic "abc"](https://chat.zulip.org/#narrow/stream/7-test-here/topic/abc) on that stream. That should make this bug now highly reproducible.

[...]

PR #4450, opened 2021-01-30
msglist: Gracefully ignore ill-formed image URLs.

gnprice:
Recently I've occasionally been seeing a red error banner at the top
of the message list (the one that says "Oh no! Something went
wrong. Try again?".)  I've only seen it when looking at #**test here**
on czo -- so figured there must be an oddity in some message there
that triggered whatever bug it was.

And indeed, it turns out that the trigger is a message with an image
element where the image's URL isn't a well-formed URL.  We'd been
implicitly assuming that that couldn't happen -- the server is after
all generating this HTML, and could prevent that.  But it turns out
that it doesn't.

Quite possibly the server should and in the future will refuse to
generate such elements.  But the Zulip server renders messages to HTML
immediately at send time, and in general we never go back and
re-render old messages.  (That's because if we did, then every subtle
change in Zulip's Markdown format, or even changes in its
implementation that were intended to have no effect on behavior, would
carry a risk of silently corrupting old messages.)  So even if the
server becomes stricter about this in the future, clients will still
need to handle messages with this kind of malformedness indefinitely.

(For that matter, in fact, it's possible the server already has made
that change in the two years since the message I found that triggers
this issue was sent.)

From the code, it looks like this bug has been here since we started
explicitly absolutizing these URLs, in 7ef0aebe0 in 2019.  Yet I was
pretty confident I hadn't seen it before, when I saw it sometime in
the last few weeks; and every time I've seen it since then, it's been
on the same stream and likely due to the same one message there.
So that's a little reassuring that this isn't a super common issue
to hit, at least.

Fixes: #4449


chrisbobbe:
Merged, thanks!

Issue #4453, opened 2021-01-30
Tried to use NavigationService before appContainerRef was set


gnprice:
In our [Sentry error reports](https://sentry.io/organizations/zulip/issues/?project=191284&query=is%3Aunresolved+level%3Aerror&sort=user), the issue affecting the most users (after the "issue" that consists of all API errors returned by the server combined: #3864) is [this one](https://sentry.io/organizations/zulip/issues/2033201611/events/a7e1d4eeec0a41419a10b179082f1a84/?project=191284), thrown here in `NavigationService.js`:
```js
const getState = (): NavigationState => {
  if (appContainerRef.current === null) {
    throw new Error('Tried to use NavigationService before appContainerRef was set.');
  }
```

The stack above that is `getNavState` < `getNavigationRoutes` < `initialFetchComplete` < `doInitialFetch`.

Based on that stack, it looks like this exception would interrupt our trying to navigate to the app's main UI, away from the loading screen, after completing the initial data fetch. So this could mean that people start the app and are unable to actually get to use it. That's a high-impact symptom, and this seems to be affecting a significant number of people, so marking this as a top priority.

The error reports for this are on 27.157 and 27.158 (the current beta), but no older versions that I see. So it seems to have been a regression in 27.157.

It's happening *almost* entirely on Android -- but not quite; there are a small number of instances of it on iOS. It's happening across roughly our overall mix of Android versions: plenty of Android 10, Android 11, and Android 9, and some older versions.

PR #4432, opened 2021-01-25
fix-4036 swipe back message

Aman-Maheshwari:
## Proposed changes

Fix: issue handling go back on swipe on android
## Issue(s)	

Fix issue #4036 .
#4398 is closed because of merge issue.

## Video

https://user-images.githubusercontent.com/50165440/105683728-f8ec7800-5f19-11eb-9faa-94dbcbd68342.mp4

PR #4454, opened 2021-01-30
Fix "Tried to use NavigationService before appContainerRef was set".

chrisbobbe:
See [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4453.20.22Tried.20to.20use.20NavigationService.20before.20appContaine.2E.2E.2E/near/1111344).

@gnprice, feel free to make any tweaks as you see fit (or ask me to, and I will). Also, let me know if you'd like to try a different approach, of course. 🙂 

Fixes: #4453

### DIFF
--- a/src/main/MainTabsScreen.js
+++ b/src/main/MainTabsScreen.js
@@ -23,6 +23,7 @@ import ProfileScreen from '../account-info/ProfileScreen';
 import { connect } from '../react-redux';
 import { getHaveServerData } from '../selectors';
 import styles, { ThemeContext } from '../styles';
+import LoadingScreen from '../start/LoadingScreen';
 
 export type MainTabsNavigatorParamList = {|
   home: RouteParamsOf<typeof HomeScreen>,
@@ -68,7 +69,7 @@ function MainTabsScreen(props: Props) {
     //
     // Avoid rendering any of our main UI in this case, to maintain
     // the guarantee that it can all rely on server data existing.
-    return null;
+    return <LoadingScreen />;
   }
 
   return (
--- a/src/main/MainTabsScreen.js
+++ b/src/main/MainTabsScreen.js
@@ -8,6 +8,7 @@ import {
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import type { RouteProp, RouteParamsOf } from '../react-navigation';
+import type { Dispatch } from '../types';
 import type { AppNavigationProp } from '../nav/AppNavigator';
 import type { GlobalParamList } from '../nav/globalTypes';
 import { bottomTabNavigatorConfig } from '../styles/tabs';
@@ -19,7 +20,7 @@ import { IconInbox, IconSettings, IconStream } from '../common/Icons';
 import { OwnAvatar, OfflineNotice, ZulipStatusBar } from '../common';
 import IconUnreadConversations from '../nav/IconUnreadConversations';
 import ProfileScreen from '../account-info/ProfileScreen';
-import { useSelector } from '../react-redux';
+import { connect } from '../react-redux';
 import { getHaveServerData } from '../selectors';
 import styles, { ThemeContext } from '../styles';
 
@@ -41,14 +42,21 @@ const Tab = createBottomTabNavigator<
   MainTabsNavigationProp<>,
 >();
 
+type SelectorProps = $ReadOnly<{|
+  haveServerData: boolean,
+|}>;
+
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'main-tabs'>,
   route: RouteProp<'main-tabs', void>,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
 |}>;
 
-export default function MainTabsScreen(props: Props) {
+function MainTabsScreen(props: Props) {
   const { backgroundColor } = useContext(ThemeContext);
-  const haveServerData = useSelector(getHaveServerData);
+  const { haveServerData } = props;
 
   const insets = useSafeAreaInsets();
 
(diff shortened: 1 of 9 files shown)


chrisbobbe:
@gnprice, I've just pushed a new revision, following our conversations yesterday ([here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4453.20.22Tried.20to.20use.20NavigationService.20before.20appContaine.2E.2E.2E/near/1111875) and [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1111919)). Thanks for your help!

I'm not sure if the reversion to `connect` for `MainTabsScreen` is what we'll land on, but I've [mentioned](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1112271) that I haven't quite gotten it to work with `ReactReduxContext` yet.

In case we do want to keep the reversion to `connect`, would you mind distilling what you've learned about the reason for doing so, into a code comment and the commit message? I think this would help us land the fix more quickly.

src/message/fetchActions.js
```diff
@@ -172,16 +172,7 @@ const initialFetchCompletePlain = (): Action => ({
...
-    // If we're anywhere in the normal UI of the app, then remain
-    // where we are. Only reset the nav state if we're elsewhere,
-    // and in that case, go to the main screen.
-    //
-    // TODO: "elsewhere" is probably just a way of saying "on the
-    // loading screen", but we're not sure. We could adjust the
-    // conditional accordingly, if we found out we're not depending on
-    // the more general condition; see
-    //   https://github.com/zulip/zulip-mobile/pull/4274#discussion_r505941875
+  if (getNavigationRoutes().some(navigationRoute => navigationRoute.name === 'loading')) {
```
gnprice: Could this drop the `some`, and instead look just at the root route, i.e. `getNavigationRoutes()[0]`?

That would make the reasoning quoted in the commit message apply more directly -- it says

    * The possible roots of the nav stack are main-tabs, loading,
      account-pick, and realm-input. Those are the routes that we either
      reset to, or set as the initial route.
    
    * So we're talking about a possible change in behavior when the root
      route is account-pick or realm-input.

after concluding the old logic is equivalent to "root route is not main-tabs". Implicitly it's using that the new logic is equivalent to "root route is loading", in order to say that the remaining two possible root routes are the only cases where the behavior might change.

src/main/MainTabsScreen.js
```diff
@@ -119,3 +127,28 @@ export default function MainTabsScreen(props: Props) {
...
+// `MainTabsScreen`'s early return on `!haveServerData` preventing
+// that from happening.
+//
+// Thankfully, we've found that placing a `connect` here will prevent
+// that from happening:
+//   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1112465.
+//
+// When our other `connect` callsites are gone, or at least the ones
+// in `MainTabsScreen`'s descendants, we expect that we can safely
+// switch this to a `useSelector`.
```
gnprice, line 152: I think the conclusion is actually that we'll likely want to keep this here even in a future when all its descendants are using `useSelector`, because it's doing something useful for us that `useSelector` doesn't do: it interposes a new `ReactReduxContext.Provider` component, which proxies subscriptions so that the descendant components only rerender if this one continues to say their subtree should be kept around.

I guess right now it seems to break as long as a relevant descendant uses `connect`, and works if it's converted to `useSelector` too. After investigation, I wound up puzzled that `useSelector` would work here when `connect` doesn't:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1111970
So clearly there is something I still don't understand. But, the job `connect` is doing for us here is one that makes sense to me as being a useful one.

src/message/fetchActions.js
```diff
@@ -172,9 +169,6 @@ const initialFetchCompletePlain = (): Action => ({
 });
 
 export const initialFetchComplete = () => async (dispatch: Dispatch, getState: GetState) => {
-  if (getNavigationRoutes().some(navigationRoute => navigationRoute.name === 'loading')) {
-    NavigationService.dispatch(resetToMainTabs());
```
gnprice, line 185: > Fixes: #4432

Should this say #4453?

gnprice (commented):
Thanks @chrisbobbe for this fix! Just a few comments below.

src/message/fetchActions.js
chrisbobbe: Ah, yes—thanks!

src/main/MainTabsScreen.js
chrisbobbe: Ah OK, thanks! If I haven't got it quite right in this revision (which borrows heavily from your comment here), I'd appreciate it if you'd jump in and make any necessary changes to set the record straight. 🙂

chrisbobbe:

src/message/fetchActions.js
chrisbobbe: Yep, sounds right!

gnprice:
Merged, thanks!

I made one fix, in an intermediate commit in some code a later commit deletes anyway:
```diff
 export const initialFetchComplete = () => async (dispatch: Dispatch, getState: GetState) => {
-  if (getNavigationRoutes()[0] === 'loading') {
+  /* flowlint-next-line unnecessary-optional-chain:off */ // [0] may not exist
+  if (getNavigationRoutes()[0]?.name === 'loading') {
```

chrisbobbe:
Thanks for that catch! 🙂

PR #4462, opened 2021-02-02
Convert three more components to use Hooks and `useSelector`.

chrisbobbe:
~~See discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1111919). The use of `connect` in these components instead of `useSelector` has been contributing to a crash.~~

(Never mind; I've got a fix for this open in #4454, where we add a `connect` in `MainTabsScreen`. Still, might as well not lose the work of Hooksifying these components, so I'll leave this PR open.)


gnprice:
Thanks -- merged!

PR #4463, opened 2021-02-02
replaceRevive [nfc]: Fix a latent bug caused by a sketchy null check.

chrisbobbe:
This is only latent because, so far, we only use
`ImmutableMapNumKeys` for `state.messages`. Greg confirms that it
would be very surprising if a message could have an ID of zero [1].

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/replaceRevive.20bug/near/1112480


gnprice:
Merged, thanks!

PR #4465, opened 2021-02-03
Start type checking `MessageList`'s props.

chrisbobbe:
By adding a type-wrapper for `connectActionSheet`, and annotating the export of MessageList.js with `ComponentType<OuterProps>`, as @gnprice mentioned in [a recent discussion of Flow and `connect`](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20and.20.60connect.60/near/1098230).

I've removed the kind of hacky cases where we use the same name for an "outer" prop as we use for a prop that `connect` provides. Let me know if you see a better way to do this, and in particular, if you don't like the changes to `MessageList`'s interface. 🙂 

Also, in this one—

MessageList [nfc]: Remove htmlPieceDescriptorsForShownMessages prop (3/x).

—there's a potential performance regression, where we stop using a caching selector. After I'd removed that, it occurred to me that some of the work it caches (calling `getHtmlPieceDescriptors` on each shown message) is enough that we'll want to keep the caching and find something else to do about the `htmlPieceDescriptorsForShownMessages` prop. What do you think?

### DIFF
--- a/src/react-native-action-sheet.js
+++ b/src/react-native-action-sheet.js
@@ -1,0 +1,29 @@
+/* @flow strict-local */
+import type { ComponentType, ElementConfig } from 'react';
+import { connectActionSheet as connectActionSheetInner } from '@expo/react-native-action-sheet';
+
+import type { BoundedDiff } from './generics';
+
+/* eslint-disable flowtype/generic-spacing */
+
+export type ShowActionSheetWithOptions = (
+  { options: string[], cancelButtonIndex: number },
+  (number) => void,
+) => void;
+
+/**
+ * Exactly like the `connectActionSheet` in
+ *   `react-native-action-sheet` upstream, but more typed.
+ */
+export function connectActionSheet<P, C: ComponentType<P>>(
+  WrappedComponent: C,
+): ComponentType<
+  $ReadOnly<
+    BoundedDiff<
+      $Exact<ElementConfig<C>>,
+      {| showActionSheetWithOptions: ShowActionSheetWithOptions |},
+    >,
+  >,
+> {
+  return connectActionSheetInner(WrappedComponent);
+}
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -22,8 +22,6 @@ type Props = $ReadOnly<{|
 |}>;
 
 export default class SearchMessagesCard extends PureComponent<Props> {
-  static NOT_FETCHING = { older: false, newer: false };
-
   render() {
     const { isFetching, messages } = this.props;
 
@@ -52,7 +50,6 @@ export default class SearchMessagesCard extends PureComponent<Props> {
           messages={messages}
           narrow={HOME_NARROW}
           htmlPieceDescriptorsForShownMessages={htmlPieceDescriptors}
-          fetching={SearchMessagesCard.NOT_FETCHING}
           showMessagePlaceholders={false}
         />
       </View>
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -41,15 +41,19 @@ export default class SearchMessagesCard extends PureComponent<Props> {
       return <SearchEmptyState text="No results" />;
     }
 
-    const htmlPieceDescriptors = getHtmlPieceDescriptorsForMessages(messages, HOME_NARROW);
+    // TODO: This is kind of a hack.
+    const narrow = HOME_NARROW;
 
     return (
       <View style={styles.results}>
         <MessageList
           initialScrollMessageId={messages[0].id}
           messages={messages}
-          narrow={HOME_NARROW}
-          htmlPieceDescriptorsForShownMessages={htmlPieceDescriptors}
+          narrow={narrow}
+          htmlPieceDescriptorsForShownMessages={getHtmlPieceDescriptorsForMessages(
+            messages,
+            narrow,
+          )}
           showMessagePlaceholders={false}
         />
       </View>
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -50,6 +50,9 @@ export default class SearchMessagesCard extends PureComponent<Props> {
           messages={messages}
           narrow={narrow}
           showMessagePlaceholders={false}
+          // TODO: handle editing a message from the search results,
+          // or make this prop optional
+          startEditMessage={() => undefined}
         />
       </View>
     );
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -8,7 +8,7 @@ import { createStyleSheet } from '../styles';
 import { LoadingIndicator, SearchEmptyState } from '../common';
 import { HOME_NARROW } from '../utils/narrow';
 import MessageList from '../webview/MessageList';
-import getHtmlPieceDescriptors from '../message/getHtmlPieceDescriptors';
+import { getHtmlPieceDescriptorsForMessages } from '../message/messageSelectors';
 
 const styles = createStyleSheet({
   results: {
@@ -41,7 +41,7 @@ export default class SearchMessagesCard extends PureComponent<Props> {
       return <SearchEmptyState text="No results" />;
     }
 
-    const htmlPieceDescriptors = getHtmlPieceDescriptors(messages, HOME_NARROW);
+    const htmlPieceDescriptors = getHtmlPieceDescriptorsForMessages(messages, HOME_NARROW);
 
     return (
       <View style={styles.results}>
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -8,7 +8,6 @@ import { createStyleSheet } from '../styles';
 import { LoadingIndicator, SearchEmptyState } from '../common';
 import { HOME_NARROW } from '../utils/narrow';
 import MessageList from '../webview/MessageList';
-import { getHtmlPieceDescriptorsForMessages } from '../message/messageSelectors';
 
 const styles = createStyleSheet({
   results: {
@@ -50,10 +49,6 @@ export default class SearchMessagesCard extends PureComponent<Props> {
           initialScrollMessageId={messages[0].id}
           messages={messages}
           narrow={narrow}
-          htmlPieceDescriptorsForShownMessages={getHtmlPieceDescriptorsForMessages(
-            messages,
-            narrow,
-          )}
           showMessagePlaceholders={false}
         />
       </View>
--- a/src/search/SearchMessagesCard.js
+++ b/src/search/SearchMessagesCard.js
@@ -9,7 +9,6 @@ import { LoadingIndicator, SearchEmptyState } from '../common';
 import { HOME_NARROW } from '../utils/narrow';
 import MessageList from '../webview/MessageList';
 import getHtmlPieceDescriptors from '../message/getHtmlPieceDescriptors';
-import { NULL_ARRAY } from '../nullObjects';
 
 const styles = createStyleSheet({
   results: {
@@ -55,7 +54,6 @@ export default class SearchMessagesCard extends PureComponent<Props> {
           htmlPieceDescriptorsForShownMessages={htmlPieceDescriptors}
           fetching={SearchMessagesCard.NOT_FETCHING}
           showMessagePlaceholders={false}
-          typingUsers={NULL_ARRAY}
         />
       </View>
     );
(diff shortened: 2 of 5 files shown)


gnprice:
> Also, in this one—
> 
> MessageList [nfc]: Remove htmlPieceDescriptorsForShownMessages prop (3/x).
> 
> —there's a potential performance regression, where we stop using a caching selector. After I'd removed that, it occurred to me that some of the work it caches (calling `getHtmlPieceDescriptors` on each shown message) is enough that we'll want to keep the caching and find something else to do about the `htmlPieceDescriptorsForShownMessages` prop. What do you think?

Yeah, I think we'll want to keep that work memoized.

There's likely a related perf regression at
c7487a2868c8b0638c3a0262cb7e7ed42d34b9c3 MessageList [nfc]: Remove htmlPieceDescriptorsForShownMessages prop (6/x).
where instead of having that data as a prop, it becomes something that gets recomputed in several different places and potentially several times.

Oh, I guess two of those get removed a couple of commits later. :slightly_smiling_face: That mitigates it. Still, should be memoized so that we only compute it once each time the inputs change.

One thing I notice from this refactoring is that in the old code, we aren't getting memoization in the search case! Each time the `SearchMessagesCard` got rerendered, for any reason, it would go recompute `htmlPieceDescriptors` from scratch. So one benefit of centralizing this logic in `MessageList`, as this branch does, is that we can be sure to consistently memoize.

chrisbobbe:
> Yeah, I think we'll want to keep that work memoized.

Makes sense.

Looking at how best to memoize `getHtmlPieceDescriptors`, I notice that it requires some messages and a narrow. With `MessageList` newly requiring `props.narrow` and `props.messages`, I think it makes sense to use those. Which means…`state`, as it's passed to `mapStateToProps`, isn't an input. Which means we're not tied to using a caching selector from `reselect`; in fact, I think it would look a bit odd to use one.

Does that sound right? Can we, e.g., use [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo) inside `mapStateToProps`?

chrisbobbe:
Thanks, @gnprice! New revision pushed.

chrisbobbe:
@gnprice, I've just rebased this past #4518. Thanks for reviewing #4518 so quickly!

Getting `MessageList`'s prop type-checking to work would be really helpful for the next part in my Flow-related work toward the RN v0.64 upgrade. Quoting from #4465's description at https://github.com/zulip/zulip-mobile/pull/4518#issue-588844365:

> I have a draft on top of this that goes further toward #3452. The "implicit-inexact-object" lint Greg mentions at https://github.com/zulip/zulip-mobile/issues/3452#issuecomment-657778760 finds 192 warnings (142 in our own code, the rest in libdefs) at the current `master` (before this PR). At the tip of that draft in its current state, it finds just 54 warnings, with 19 in our own code, and those numbers are decreasing 😄.

I can be more confident in that draft after a PR like the current one is merged. 🙂

chrisbobbe:
Just fixed some rebase conflicts. 🙂

src/webview/MessageList.js
```diff
@@ -93,18 +91,20 @@ type SelectorProps = {|
...
 |};
 
 // TODO get a type for `connectActionSheet` so this gets fully type-checked.
 export type Props = $ReadOnly<{|
   narrow: Narrow,
+  messages: $ReadOnlyArray<Message | Outbox>,
+  initialScrollMessageId: number | null,
   showMessagePlaceholders: boolean,
   startEditMessage: (editMessage: EditMessage) => void,
+  disableFetching?: true,
```
gnprice, line 106: This name seems a bit misleading because it doesn't affect any logic that controls actually *fetching*. Rather it suppresses the bits of UI that indicate that we are fetching. Perhaps `suppressFetchingIndicator`, similar to the `suppressTypingIndicator` elsewhere in the branch?

Huh, or in fact: I think we could just never pass this prop, and we'd get the same behavior. We'd go look up the narrow in the fetching state, and not find it, and return...
```js
/** The value implicitly represented by a missing entry in FetchingState. */
export const DEFAULT_FETCHING = { older: false, newer: false };
```

That seems better. After all if we *did* start doing fetching when you scroll to the top of a search narrow, we'd want that to get indicated just the same way as in other narrows.

src/webview/MessageList.js
```diff
@@ -93,18 +91,20 @@ type SelectorProps = {|
...
+  suppressTypingIndicator?: true,
```
gnprice, line 107: Similarly, we could just drop this prop entirely and get the right result -- and really it'd be a more natural way to get that result.

src/webview/MessageList.js
```diff
@@ -365,7 +365,7 @@ export default connect<SelectorProps, _, _>((state, props: OuterProps) => {
       ? { older: false, newer: false }
       : getFetchingForNarrow(state, props.narrow),
     htmlPieceDescriptorsForShownMessages: props.htmlPieceDescriptorsForShownMessages
-      ? props.htmlPieceDescriptorsForShownMessages
+      ? getHtmlPieceDescriptorsForMessages(props.messages, props.narrow)
       : getHtmlPieceDescriptorsForMessages(props.messages, props.narrow),
```
gnprice: SearchMessagesCard [nfc]: Don't pass htmlPieceDescriptorsForShownMessages.
    
    Instead, have the conditional branch in `MessageList` that was
    exercised by passing this prop compute the value it needs for
    itself.

I'm a bit confused by this commit. We stop passing the prop in this caller; and then we also change the behavior in the case where the prop *is* passed?

(But at this point there is no case where it's passed, right?)

I think this and the next two commits can perhaps be clearer as just two commits:
* First, stop passing the prop in the search case (the only case where we were passing it.) This is NFC because the falsy branch of this conditional computes the same thing we had been passing.
* Then, eliminate the prop from the interface, and always just compute it here. This is NFC because at this point we're never passing the prop anyway.

src/webview/MessageList.js
```diff
@@ -335,6 +335,7 @@ type OuterProps = {|
   messages: $ReadOnlyArray<Message | Outbox>,
   initialScrollMessageId: number | null,
   showMessagePlaceholders: boolean,
+  startEditMessage: (editMessage: EditMessage) => void,
```
gnprice, line 338: MessageList types: Copy `startEditMessage` to `OuterProps`, where it belongs.
    
    Now, `OuterProps` matches the subset of `Props` that callers are
    meant to pass. If `OuterProps` must exist, this is what it should
    look like.

Ah, good catch, thanks.

Indeed this type definition ideally shouldn't need to exist. The reason for it lives in one of those commit messages where I probably should have put some of the information in a code comment:

commit 3168ff5c1b36cb03947bd9e431de98a266fdcfef
Date:   Wed Sep 5 18:13:37 2018 -0700

    message list: Give a real type to the exported MessageList.
    
    By annotating the argument to `connect`, we enable Flow to infer the
    type of its return value -- aka `MessageList`, as seen by the rest of
    the code -- and type-check call sites.
    
    The repetition is unfortunate and I'd be glad to find a way to avoid
    it.  I tried `Props = OuterProps & { /* ... */ }`, but unfortunately
    that left Flow still thinking that `anchor`, `fetching`, etc. could be
    missing because they could in `OuterProps`.  This is much better than
    not having types, though.
    
    Also, because `connectActionSheet` lacks a type annotation, the
    relationship *between* `OuterProps` and `Props`, and the code that
    (*ahem*) connects them, isn't checked.  That'd be good to fix.
    
    We abuse this `Props` type in a lot of places where we really mean
    some other type that merely overlaps, and that contributes to some
    mysteries still to be resolved here.  I cleared up one --
    `highlightUnreadMessages` just never appears in something claiming to
    use this type -- but a few others are marked with comments.

I think the key point there is that `connectActionSheet` didn't have a type. (The `Props = OuterProps & /* … */` thing was an attempt to use [intersection types](https://flow.org/en/docs/types/intersections/); if I were attempting the same thing now, I'd write `Props = {| ...OuterProps, /* … */ |}` and work from there. Intersection types are potentially helpful for inexact object types -- which we were using then because we didn't entirely know better -- but don't have a lot of other applications.)

With the next commit in this PR, though, it does! So maybe we can drop `OuterProps` after that?

Or, given that `OuterProps` comes in independently handy in the last commit of the branch: maybe define it at the top and save the duplication by spreading it into `Props`? Just like we do `SelectorProps`, basically.

src/react-native-action-sheet.js
```diff
@@ -0,0 +1,29 @@
...
+/* eslint-disable flowtype/generic-spacing */
+
+export type ShowActionSheetWithOptions = (
+  { options: string[], cancelButtonIndex: number },
+  (number) => void,
+) => void;
+
+/**
+ * Exactly like the `connectActionSheet` in
+ *   `react-native-action-sheet` upstream, but more typed.
```
gnprice, line 16: action sheet: Add type wrapper for `connectActionSheet`.
    
    `connectActionSheet`'s poor typing is one thing stopping
    `MessageList`'s props from being type checked.
    
    Tightening up the types of those props will take some work, starting
    from sorting out the various hacks the search-messages screen uses
    when it passes props with names that match the names we assign to
    data from react-redux.
    
    Like we did for react-native-safe-area-context in 8f569641f, use a
    type-wrapper instead of a libdef so that we can use our handy
    `BoundedDiff` type.

Excellent :smile: 

I think the "Tightening up the types … will take some work" graf can pretty much be dropped -- it mostly describes the work that's now done in the earlier commits of this branch. If a version of it is kept, it should be updated accordingly.

src/search/SearchMessagesCard.js
```diff
@@ -52,6 +52,9 @@ export default class SearchMessagesCard extends PureComponent<Props> {
           disableFetching
           showMessagePlaceholders={false}
           suppressTypingIndicator
+          // TODO: handle editing a message from the search results,
+          // or don't let callers pass this.
+          startEditMessage={() => undefined}
```
gnprice, line 57: What do you have in mind for "don't let callers pass this"? The two natural options I see here are
* handle editing a message from the search results, or
* make this optional.

Probably we don't want to try to handle editing from here, so that'd point to the latter. We already have this where we construct the action sheet:
```js
  if (
    // Our "edit message" UI only works in certain kinds of narrows.
    && (isStreamOrTopicNarrow(narrow) || isPmNarrow(narrow))
  ) {
    buttons.push(editMessage);
  }
```

Perhaps that filter on the narrow could move up to `ChatScreen`, and control whether we pass a non-void `startEditMessage` callback at all? (And here in search, just never pass it.) Then in the action sheet we could just check if the callback is truthy.

gnprice (commented):
Thanks @chrisbobbe ! This looks great -- sorry for the delayed review. Comments below.

src/webview/MessageList.js
chrisbobbe: > The reason for it lives in one of those commit messages where I probably should have put some of the information in a code comment

IIRC, I found this commit message, noted that it said `OuterProps` was necessary, then also realized that it's worth keeping for the last commit in the branch. I elided all that with just "If `OuterProps` must exist", probably because I didn't feel like explaining everything? 😛

src/react-native-action-sheet.js
chrisbobbe: Ah, right—looks like this commit started its life much earlier in the branch, then I moved it later without updating the commit message.

src/search/SearchMessagesCard.js
chrisbobbe: > don't let callers pass this

I think I meant to say "don't require callers to pass this".

chrisbobbe: > Perhaps that filter on the narrow could move up to `ChatScreen`, and control whether we pass a non-void `startEditMessage` callback at all? (And here in search, just never pass it.) Then in the action sheet we could just check if the callback is truthy.

Yeah; just this week, it's come up that it'd be nice to do this conditional in just one place: https://github.com/zulip/zulip-mobile/pull/4799#discussion_r651298734.

To check if the callback is truthy, in the action sheet, are we OK with `constructMessageActionButtons` taking another param? Currently it looks like this:

```js
export const constructMessageActionButtons = ({
  backgroundData: { ownUser, flags },
  message,
  narrow,
}: {
  backgroundData: $ReadOnly<{
    ownUser: User,
    flags: FlagsState,
    ...
  }>,
  message: Message,
  narrow: Narrow,
}): Button<MessageArgs>[] => {
  // ...
}
```

chrisbobbe:

src/search/SearchMessagesCard.js
gnprice: Hmm, yeah, it is a little awkward to pass it down there (through `constructNonHeaderActionButtons`, too. But I think it's OK.

gnprice:
Thanks, looks good! Merged.

There's one open thread about a possible followup: https://github.com/zulip/zulip-mobile/pull/4465#discussion_r652996826

chrisbobbe:
Thanks for the review!

PR #4466, opened 2021-02-04
Convert `StreamItem` and `TopicItem` to use Hooks.

chrisbobbe:
To simplify translating some user-facing strings. `StreamItem`, at least, will need two contexts: the theme context and the translation context. With the new Context API, a class component can't declare multiple `contextType`s. But Hooks components work just fine with two different `useContext` calls (see [example](https://reactjs.org/docs/hooks-overview.html#other-hooks)).

From [discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Adding.20Accessibilty/near/1113306).

### DIFF
--- a/src/streams/TopicItem.js
+++ b/src/streams/TopicItem.js
@@ -41,14 +41,7 @@ export default function TopicItem(props: Props) {
   } = props;
 
   return (
-    <Touchable
-      onPress={() => {
-        onPress(stream, name);
-      }}
-      onLongPress={() => {
-        showToast(name);
-      }}
-    >
+    <Touchable onPress={() => onPress(stream, name)} onLongPress={() => showToast(name)}>
       <View
         style={[
           styles.listItem,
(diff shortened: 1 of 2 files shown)


src/streams/TopicItem.js
```diff
@@ -38,21 +38,18 @@ export default class TopicItem extends PureComponent<Props> {
...
 
     return (
-      <Touchable onPress={this.handlePress} onLongPress={this.handleLongPress}>
+      <Touchable
+        onPress={() => {
+          onPress(stream, name);
+        }}
+        onLongPress={() => {
+          showToast(name);
+        }}
```
gnprice: nit: could be kept shorter by saying simply e.g. `() => onPress(stream, name)`.

(Both callbacks already return `void`, so that's exactly equivalent.)

gnprice (commented):
Thanks! Looks good -- merged, with a small commit on top for the nit below.

PR #4467, opened 2021-02-04
Add a margin and accent color to the Android notification icon

andersk:
* generate-logos [nfc]: Expand inkscape --export-png for ≥ 1.0 compatibility.

* generate-logos [nfc]: Add z-white.svg to the repository.

* android assets: Pad notification icon for consistency with other apps.
  
  Add a 10% margin on all sides, so the height of the Z comes out to an integer at all sizes we generate, which are multiples of 12px.

* android notif: Set an accent color for notifications.
  
  rgb(100, 146, 253.5) is the sRGB midpoint of our logo’s gradient.

Before:
before

After:
after

Before:
before

After:
after


gnprice:
Looks great, merged -- thanks @andersk!

PR #4468, opened 2021-02-04
Clean up navigation logic (4/x): Tidy up our use of `ZulipStatusBar` a bit more.

chrisbobbe:
I've removed the work toward using React Navigation's `header` option in this revision, and changed the title.

Toward: #3066.

-----

This PR is the next in the series started by #4428, #4440, and #4443.

Here, we use the [`header` option](https://reactnavigation.org/docs/stack-navigator/#header)—for the first time!—for `ChatNavBar`, as part of the safe-area-handling pattern I described at https://github.com/zulip/zulip-mobile/pull/4443#issue-563657336.

I'm a bit disappointed to see that Flow isn't tracking what route we're on when we use the `header` option, so I have to suppress it when it says it's suspicious about grabbing specific route params to give to `ChatNavBar`. I'm not sure how easy this would be to fix. With React Navigation v5's component-based API, I'm not sure if it's a case where we'd need to ask more from the React types than is reasonable.

Also, on iOS, we start showing and hiding the status bar as the lightbox header/footer slide in and out, as Greg mentioned wanting to do at https://github.com/zulip/zulip-mobile/pull/4440#issuecomment-768603140.

I think I worked out something odd in how RN's `KeyboardAvoidingView` works, and I made a note on that.

### DIFF
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -24,7 +24,7 @@ import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'chat'>,
-  route: RouteProp<'chat', {| narrow: Narrow |}>,
+  route: RouteProp<'chat', {| narrow: Narrow, editMessage: EditMessage | null |}>,
 |}>;
 
 const componentStyles = createStyleSheet({
@@ -98,9 +98,12 @@ const useFetchMessages = args => {
 };
 
 export default function ChatScreen(props: Props) {
+  const { route, navigation } = props;
   const { backgroundColor } = React.useContext(ThemeContext);
 
-  const [editMessage, setEditMessage] = React.useState<EditMessage | null>(null);
+  const { editMessage } = route.params;
+  const setEditMessage = (value: EditMessage | null) =>
+    navigation.setParams({ editMessage: value });
 
   const { narrow } = props.route.params;
 
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -101,11 +101,9 @@ export default function ChatScreen(props: Props) {
   const { route, navigation } = props;
   const { backgroundColor } = React.useContext(ThemeContext);
 
-  const { editMessage } = route.params;
+  const { narrow, editMessage } = route.params;
   const setEditMessage = (value: EditMessage | null) =>
     navigation.setParams({ editMessage: value });
-
-  const { narrow } = props.route.params;
 
   const isNarrowValid = useSelector(state => getIsNarrowValid(state, narrow));
 
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -1,12 +1,11 @@
 /* @flow strict-local */
 import React from 'react';
-import { View } from 'react-native';
 import { useIsFocused } from '@react-navigation/native';
 
 import { useSelector, useDispatch } from '../react-redux';
 import type { RouteProp } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
-import styles, { ThemeContext, createStyleSheet } from '../styles';
+import { ThemeContext, createStyleSheet } from '../styles';
 import type { Narrow, EditMessage } from '../types';
 import { KeyboardAvoider, OfflineNotice } from '../common';
 import ChatNavBar from '../nav/ChatNavBar';
@@ -114,36 +113,34 @@ export default function ChatScreen(props: Props) {
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
   return (
-    <View style={[componentStyles.screen, { backgroundColor }]}>
-      <KeyboardAvoider style={styles.flexed} behavior="padding">
-        <ChatNavBar narrow={narrow} editMessage={editMessage} />
-        <OfflineNotice />
-        <UnreadNotice narrow={narrow} />
-        {(() => {
-          if (!isNarrowValid) {
-            return <InvalidNarrow narrow={narrow} />;
-          } else if (fetchError !== null) {
-            return <FetchError narrow={narrow} error={fetchError} />;
-          } else if (sayNoMessages) {
-            return <NoMessages narrow={narrow} />;
-          } else {
-            return (
-              <MessageList
-                narrow={narrow}
-                showMessagePlaceholders={showMessagePlaceholders}
-                startEditMessage={setEditMessage}
-              />
-            );
-          }
-        })()}
-        {showComposeBox && (
-          <ComposeBox
-            narrow={narrow}
-            editMessage={editMessage}
-            completeEditMessage={() => setEditMessage(null)}
-          />
-        )}
-      </KeyboardAvoider>
-    </View>
+    <KeyboardAvoider style={[componentStyles.screen, { backgroundColor }]} behavior="padding">
+      <ChatNavBar narrow={narrow} editMessage={editMessage} />
+      <OfflineNotice />
+      <UnreadNotice narrow={narrow} />
+      {(() => {
+        if (!isNarrowValid) {
+          return <InvalidNarrow narrow={narrow} />;
+        } else if (fetchError !== null) {
+          return <FetchError narrow={narrow} error={fetchError} />;
+        } else if (sayNoMessages) {
+          return <NoMessages narrow={narrow} />;
+        } else {
+          return (
+            <MessageList
+              narrow={narrow}
+              showMessagePlaceholders={showMessagePlaceholders}
+              startEditMessage={setEditMessage}
+            />
+          );
+        }
+      })()}
+      {showComposeBox && (
+        <ComposeBox
+          narrow={narrow}
+          editMessage={editMessage}
+          completeEditMessage={() => setEditMessage(null)}
+        />
+      )}
+    </KeyboardAvoider>
   );
 }
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -8,7 +8,7 @@ import type { RouteProp } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
 import styles, { ThemeContext, createStyleSheet } from '../styles';
 import type { Narrow, EditMessage } from '../types';
-import { KeyboardAvoider, OfflineNotice, ZulipStatusBar } from '../common';
+import { KeyboardAvoider, OfflineNotice } from '../common';
 import ChatNavBar from '../nav/ChatNavBar';
 import MessageList from '../webview/MessageList';
 import NoMessages from '../message/NoMessages';
@@ -21,7 +21,6 @@ import { canSendToNarrow } from '../utils/narrow';
 import { getLoading, getSession } from '../directSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './narrowsSelectors';
-import { getStreamColorForNarrow } from '../subscriptions/subscriptionSelectors';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'chat'>,
@@ -113,12 +112,9 @@ export default function ChatScreen(props: Props) {
   const sayNoMessages = haveNoMessages && !isFetching;
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
-  const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
-
   return (
     <View style={[componentStyles.screen, { backgroundColor }]}>
       <KeyboardAvoider style={styles.flexed} behavior="padding">
-        <ZulipStatusBar backgroundColor={streamColor} />
         <ChatNavBar narrow={narrow} editMessage={editMessage} />
         <OfflineNotice />
         <UnreadNotice narrow={narrow} />
--- a/src/common/KeyboardAvoider.js
+++ b/src/common/KeyboardAvoider.js
@@ -9,6 +9,25 @@ type Props = $ReadOnly<{|
   children: React$Node,
   style?: ViewStyleProp,
   contentContainerStyle?: ViewStyleProp,
+
+  /** How much the top of `KeyboardAvoider`'s layout *parent* is
+   * displaced downward from the top of the screen.
+   *
+   * If this isn't set correctly, the keyboard will hide some of
+   * the bottom of the screen (an area whose height is what this
+   * value should have been set to).
+   *
+   * I think `KeyboardAvoidingView`'s implementation mistakes `x`
+   * and `y` from `View#onLayout` to be a `View`'s position
+   * relative to the top left of the screen. In reality, I'm
+   * pretty sure they represent a `View`'s position relative to
+   * its parent:
+   *   https://github.com/facebook/react-native-website/issues/2056#issuecomment-773618381
+   *
+   * But at least `KeyboardAvoidingView` exposes this prop, which
+   * we can use to balance the equation if we need to.
+   */
+  keyboardVerticalOffset?: number,
 |}>;
 
 /**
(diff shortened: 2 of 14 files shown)


chrisbobbe:
- ZulipMobile [nfc]: Use just one `ActionSheetProvider`, here. 
- ChatScreen [nfc]: Remove unnecessary wrapping `View`.

I made these two commits before I quite figured out what was going on with the `KeyboardAvoidingView`, so they can freely be dropped if they don't seem helpful.

src/chat/ChatScreen.js
```diff
@@ -24,7 +24,7 @@ import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'chat'>,
-  route: RouteProp<'chat', {| narrow: Narrow |}>,
+  route: RouteProp<'chat', {| narrow: Narrow, editMessage: EditMessage | null |}>,
```
chrisbobbe, line 26:
```
ChatScreen: Store `editMessage` as a nav param, not React state.

The fact that `ChatScreen`'s header changes with this bit of state
is a clue that it really wants to be a route param. When we start
using the `header` option [1] for the header, instead of smushing
the header in with the rest of the screen's content, it'll be
convenient to grab `editMessage` from the route params.

[1] https://reactnavigation.org/docs/stack-navigator/#header
```

I think we should aim to reduce the contents of `editMessage`, probably to just the ID of the message being edited, along the lines of [what React Navigation recommends should go in these params](https://reactnavigation.org/docs/params/#what-should-be-in-params).

I think the biggest drawback to storing the whole original content and topic equally applies in the status quo (i.e., where `editMessage` is stored in React state): it's model data, so it can become stale. So, perhaps a fix can be done as a followup, or we can add a TODO.

The article gives a couple of other good reasons, but some of them (like gracefully handling deep-linking) don't yet really apply to us. But I think their guidance that "You can also think of the route object like a URL" makes sense. For this screen, I'm thinking of the analogy of a URL like `/msglist?msg-edit-id=12345`.

src/nav/navActions.js
```diff
@@ -33,7 +33,7 @@ export const resetToMainTabs = (): GenericNavigationAction =>
 
 /** Only call this via `doNarrow`.  See there for details. */
 export const navigateToChat = (narrow: Narrow): GenericNavigationAction =>
-  StackActions.push('chat', { narrow });
+  StackActions.push('chat', { narrow, editMessage: null });
```
chrisbobbe, line 36: `editMessage: null` could go here, or it could go in `AppNavigator` where we define the screen ([doc](https://reactnavigation.org/docs/params/#updating-params)):

```
<Stack.Screen name="chat" component={ChatScreen} initialParams={{ editMessage: null }} />
```

src/chat/ChatScreen.js
chrisbobbe: In my draft that moves `ModalSearchNavBar` to the header, the search term is stored in the nav state. This seems fine; I'm thinking of the analogy of a URL that has something like `?q=foo`.

chrisbobbe: > In my draft that moves `ModalSearchNavBar` to the header

While at first I thought we might not want the search bar to go in the header/nav bar, I think it's a fine place to put it, at least on Android.

From an [Android doc](https://developer.android.com/training/appbar):

> The key functions of the app bar are as follows:
> - [...]
> - Access to important actions in a predictable way, such as search.

chrisbobbe:
Just rebased and resolved some conflicts.

src/nav/AppNavigator.js
```diff
@@ -119,7 +120,19 @@ export default function AppNavigator(props: Props) {
...
+      <Stack.Screen
+        name="chat"
+        component={ChatScreen}
+        options={{
+          header: stackHeaderProps => {
+            const { route } = stackHeaderProps.scene;
+            // $FlowFixMe - Flow doesn't know what route we're on
+            const params: RouteParamsOf<typeof ChatScreen> = route.params;
+            const { editMessage, narrow } = params;
+            return <ChatNavBar editMessage={editMessage} narrow={narrow} />;
```
gnprice, line 132: Hmm -- it seems like a bit of a step backward for the details of this definition to appear here at the navigator, rather than next to the `ChatScreen` component. In particular it makes it rather indirect to be looking at `ChatScreen` and work out where the header is coming from.

I think one pretty reasonable solution would be to put this code over in `ChatScreen.js`, either as a separately exported function or perhaps better as a property on `ChatScreen`. Like `ChatScreen.header = (params: RouteParamsOf<typeof ChatScreen>) => { … };`.

Even then, though, that still seems more complicated than just invoking `ChatNavBar` at the top of the component's own rendered tree. Does this buy us something that's worth the complexity?

(further discussion in main thread)

gnprice (commented):
Thanks @chrisbobbe ! Sorry for the delayed review.

Some comments below. Generally this looks good except I'm not convinced the final commit is a helpful direction. (And a couple of the prep commits may not be needed if we leave that one out.)

---

Specifically: this way of specifying the header seems rather more complicated than just invoking something like `ChatNavBar` at the top of the component's own rendered tree. Does this buy us something that's worth the complexity?

I'm looking back at https://github.com/zulip/zulip-mobile/pull/4443#issue-563657336 , and I see this at the endpoint:

> Each screen component on a stack navigator is responsible for keeping important content out of the right, bottom, and left insets. It doesn't have to worry about the top inset because its header does that, as mentioned above.

I think that means we'll still want to be handling right, bottom, and left insets somewhere generic like `Screen` for the sake of screens that don't have particular other needs, as well as handling all four sides in screen-specific ways where needed.

Given that, I don't see that we necessarily benefit from handling one of the four sides in a different way. If anything, it's probably better to handle all four together, or close by each other -- it helps in spotting gaps and understanding how things fit together.

... For example. Looking at `Screen`, it strikes me that we handle the bottom, and then the top (via Modal(Search)?NavBar), but don't handle the left and right. That seems like it's probably the cause of this symptom: https://github.com/zulip/zulip-mobile/issues/3066#issuecomment-498863856 (Assuming that symptom is still present -- my test iPhone XR is out of battery, just started it charging so I can confirm.) I guess not that *exact* symptom because those screenshots are from MainTabsScreen, but I'd predict from that code that we have that symptom when `Screen` is involved too.

chrisbobbe:
> Sorry for the delayed review.

I know you've got a lot of stuff on your list. 🙂 Thanks for the review!

It's useful to see this position against React Navigation's "magic" here; certainly they don't get everything right, so it's good to have some skepticism. I think we agree right away that React Navigation's _default_ header wouldn't work out of the box, but it might be helpful to see it, for some context, if you haven't already. On `master`, if I just remove the `headerMode="none"` from `AppNavigator`, that looks like this on iOS:

https://user-images.githubusercontent.com/22248748/109678117-50e55280-7b48-11eb-9648-cf9b49f7bc71.mp4

On iOS, the animation looks consistent with `headerMode="float"`, which makes sense because the React Nav doc [says](https://reactnavigation.org/docs/stack-navigator/#headermode) it's a common iOS pattern. (Notice how the default header doesn't slide, while our homemade header in the content area slides. Obviously if we were going with the default header, it wouldn't coexist with the homemade header.)

The default header doesn't work for a few reasons (here's an incomplete list, and this is just from looking on iOS):

- We don't want a title and nav button on `MainTabsScreen`. That's easy to get rid of, but in doing so, we should also override the `headerMode="float"` default and use "screen" instead. That's because "float" looks wrong when different screens have differently structured headers. (An alternative would be to think harder about animations; apparently they give an escape hatch if you really want to use "float".)
- To show each screen's title, it's doing the best it can—but it doesn't know that "Variables" should be capitalized, for example. Of course that'd be easy to fix, by e.g. passing the [title option](https://reactnavigation.org/docs/stack-navigator/#title).
- The background color isn't quite right, I think?
- It's way too simple for, e.g., the chat screen nav bar, which needs more than a back button.

Still, these and other issues should be fixable by either using various options like `headerLeft`, `headerRight`, [`headerBackTitleVisible`](https://reactnavigation.org/docs/stack-navigator/#headerbacktitlevisible) (hmm that one's got an interesting-looking default), `headerTitleAlign`—there are maybe two dozen of these options, which we haven't exercised—or just supplying our own component for the whole header, as this PR does.

Ah, looking back, I agree that my comments about handling safe-area insets don't read as an argument for using the `header` option. I think I'd assumed that we'd be interested in `header`, as a semantically correct and not-too-hard thing to do, but that it'd be chaos (specifically for our safe-area-handling plans) to do it in some places and not others, so I picked using it everywhere (with this PR taking one step to get there).

So, I owe some points in favor of picking that 🙂:

-----

- Being a little bit "forced" (or at least encouraged) to write our headers a certain way (e.g., if we eventually use `headerLeft` and those options I mentioned above) is intuitively likely to help keep a consistent look-and-feel on all screens. E.g., you've pointed out somewhere that some back buttons look like ← and some look like <.
- IIUC, we have three substantive headers (four if you count `MainTabsScreen`'s `<View style={{ height: insets.top }} />`): `ChatNavBar`, `ModalNavBar`, and `ModalSearchNavBar`. Currently, 
  - `Screen` switches between `ModalNavBar` and `ModalSearchNavBar` based on the `search` prop, since several screens use each.
  - `ChatScreen`, being the only screen to use `ChatNavBar`, chooses to eject from `Screen` rather than ask `Screen` to be aware of its special header preferences. I guess this is defensible; we wouldn't want `Screen`'s header-rendering logic to turn into a giant switch-case block on the current screen identifier ('account-pick', 'account-details', etc.). Would we? 🤔

  But the consequences for `ChatScreen` are a bit unfortunate: it has to duplicate code for showing the `OfflineNotice`, `UnreadNotice`, and for avoiding the keyboard, and that invites an inconsistent look-and-feel. Possibly `Screen` would be easier for `ChatScreen` (and others) to reach for if it didn't have to deal with headers.
- If we want, we could "set and forget" a default header in the navigator, with [`screenOptions`](https://reactnavigation.org/docs/stack-navigator/#screenoptions), for new screens to have. Then, screens with specific needs (like `ChatScreen`) could override the default with their special headers, but naturally not before checking that
  - the change is necessary and well-considered, and
  - the "specific needs" aren't actually more general needs that the default should handle.
- The header should always participate in the layout in pretty much the same way: it stays at the top. Using `header` gives us this guarantee, and we shouldn't hit or have to think about problems like [one you've mentioned](https://github.com/zulip/zulip-mobile/pull/4443#discussion_r567133246) where a header was scrolling up and down along with the screen's content.
- It seems odd to use the wrong semantics for setting a header; I think my strongest thought starting out was that using `header` would just be "right", and that not using it would be "wrong", because we're pretending that
  - the header is something that it's not (i.e., null), and that
  - the screen content is something that it's not (i.e., content plus header).
  
  Possibly this is just putting too much faith in React Navigation's ability to structure things nicely? Still, if we keep things how they are, maybe let's add a code comment about this? 🙂
- It seems a bit lucky that we're not going for the iOS [`headerMode="float"`](https://reactnavigation.org/docs/stack-navigator/#headermode) effect. I don't think we should go for it—but if we wanted to, I don't see how we could do it without using `header`. There might be other features like this that we'd potentially be interested in, but that we're blocked from considering with the current implementation.
- I guess I owe some data about this, but I figured there was a decent chance that accessibility (including aspects I don't yet understand) would be improved, possibly a lot, by using the common/documented approach. (As I look for that data, though, I notice that https://github.com/react-navigation/react-navigation/pulls?q=label%3Aaccessibility comes up empty, hmmmm…still, we could get any improvements naturally when taking new React Nav versions?)

-----

I'd understand if these considerations don't convince you that the trouble (including the trouble with Flow I mentioned in the description) is worth it, but I wanted to put them out there to help inform the decision, since I didn't do so before. 🙂

src/chat/ChatScreen.js
chrisbobbe: > I think the thing that makes that tricky is mainly just the message content.

Ahh—right, you're right. I saw that but forgot about it when commenting.

src/common/KeyboardAvoider.js
```diff
@@ -9,11 +9,35 @@ type Props = $ReadOnly<{|
...
+      //
+      // I think `KeyboardAvoidingView`'s implementation mistakes `x`
+      // and `y` from `View#onLayout` to be a `View`'s position
+      // relative to the top left of the screen. In reality, I'm
+      // pretty sure they represent a `View`'s position relative to
+      // its parent:
+      //   https://github.com/facebook/react-native-website/issues/2056#issuecomment-773618381
+      //
+      // But at least `KeyboardAvoidingView` exposes this prop, which
+      // we can use to balance the equation.
```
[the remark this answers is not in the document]
chrisbobbe: > it looks like it's an implementation comment

It sort of is an implementation comment, in that it explains a weirdness in some code that it calls as part of doing its job, right? If (for some strange reason 😬) that weirdness were common or easily-obtained knowledge, the jsdoc might do OK to just say "passes some props straight through to `KeyboardAvoidingView`", I suppose? I guess if it were well-known, the implementation comment wouldn't be needed either, hmm.

I agree that this is easy to miss where it currently is, so it'll be smart to propagate it to the jsdoc where it'll be surer to be found.

gnprice:
Thanks for all those details! Replied [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-nav.3A.20headers/near/1126188).

src/common/KeyboardAvoider.js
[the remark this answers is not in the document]
gnprice: Yeah, as written it does both. But this contains some key information that the reader who's simply using this component will need to know -- so that information is part of the interface and belongs in the jsdoc.

I think one way to write this would be:
* Take the first two paragraphs and put them in jsdoc. Also s/KeyboardAvoidingView/KeyboardAvoider/ there.
  * I think that covers all the information needed for simply using this component.
* Then leave the remaining paragraphs formatted as an implementation comment -- they're really information that's not needed for using the component, but is there to explain how the implementation works for the sake of someone modifying or debugging it.

> If for some strange reason () that weirdness were common (or easily obtained) knowledge, the component's jsdoc might do OK to just say "passes some props straight through to `KeyboardAvoidingView`", I suppose? I guess if it were well-known, the implementation comment wouldn't be needed either, hmm.

Yep and yep :slightly_smiling_face: 

In other circumstances, where we're just passing some props through but they *are* reasonably well documented (or at any rate we don't have any thoughts to add to what's in their docs), it can indeed make sense for our interface to just say that:

https://github.com/zulip/zulip-mobile/blob/0735ff117baa22fee9f1ead1662d2c3d48951f4e/src/common/Input.js#L28-L42

chrisbobbe:
Thanks for the review! Revision pushed, with the `header`-option work removed.

src/common/KeyboardAvoider.js
```diff
@@ -9,6 +9,25 @@ type Props = $ReadOnly<{|
   children: React$Node,
   style?: ViewStyleProp,
   contentContainerStyle?: ViewStyleProp,
+
+  /** How much the top of `KeyboardAvoidingView`'s layout *parent*
```
gnprice, line 13: Should say `KeyboardAvoider` here, I think.

(They happen to work out to the same thing, but that's only evident from looking at the implementation of `KeyboardAvoider`.)

src/common/KeyboardAvoider.js
```diff
@@ -20,7 +39,13 @@ type Props = $ReadOnly<{|
  */
 export default class KeyboardAvoider extends PureComponent<Props> {
   render() {
-    const { behavior, children, style, contentContainerStyle } = this.props;
+    const {
+      behavior,
+      children,
+      style,
+      contentContainerStyle,
+      keyboardVerticalOffset = 0,
```
gnprice, line 47: Does this need a default here? Docs say it has the same value as a default at the layer below:
https://reactnative.dev/docs/keyboardavoidingview#keyboardverticaloffset

Seeing this, I was wondering if by starting to pass 0 here we were changing any behavior. It looks like the answer is no but I think it'd be easier to see that if it's just another prop we pass straight through.

src/nav/ModalNavBar.js
```diff
@@ -24,26 +23,23 @@ export default function ModalNavBar(props: Props) {
   ];
 
   return (
-    <>
-      <ZulipStatusBar />
```
gnprice: I like this cleanup!

gnprice (commented):
Thanks @chrisbobbe ! This looks great -- small comments below. Please merge at will after taking a look through those.

chrisbobbe:
> Please merge at will after taking a look through those.

Done, thanks for the reviews!

PR #4471, opened 2021-02-08
Fix:AccountDetails & StreamSettings screen in landscape mode

rajprakash00:
**Issue**:
In landscape mode the accountDetails and streamSettings screen were not completely visible.
**Changes Done**:
Made it scrollable 

**Before**:

https://user-images.githubusercontent.com/42652941/107241571-1dd00780-6a51-11eb-9437-7ad862039ce5.mp4

**After**:

https://user-images.githubusercontent.com/42652941/107241829-5e2f8580-6a51-11eb-8f9f-79db5f697069.mp4

### DIFF
--- a/src/common/Screen.js
+++ b/src/common/Screen.js
@@ -137,7 +137,17 @@ class Screen extends PureComponent<Props> {
           contentContainerStyle={[padding && styles.padding]}
         >
           <ScrollView
-            contentContainerStyle={[styles.flexed, centerContent && componentStyles.content, style]}
+            contentContainerStyle={[
+              // If `Screen` is responsible for managing scrolling,
+              // keep its childrens' height unbounded. If not, set a
+              // bounded height on its children, e.g., as required for
+              // a child `FlatList` or `SectionList` to work (or even
+              // another `ScrollView`, but hopefully we don't nest too
+              // many of these!).
+              !scrollEnabled ? styles.flexed : null,
+              centerContent && componentStyles.content,
+              style,
+            ]}
             style={componentStyles.childrenWrapper}
             keyboardShouldPersistTaps={keyboardShouldPersistTaps}
             scrollEnabled={scrollEnabled}


chrisbobbe:
Hello! It does seem annoying that you can't see the full content of those screens; thanks for noticing that!

I haven't had the chance to give this the full review it deserves, but one comment is that I'd prefer not to introduce more `ScrollView`s than necessary. Let's try the `scrollEnabled` prop of the `<Screen />` components used in both of those files. 🙂 Would that work?

chrisbobbe:
Thanks, @rajprakash00! We've discussed this [in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Screen.20component.20in.20landscape.20mode/near/1115493).

Please [squash](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html) the "fix previous commit" commit into the previous commit. 🙂

chrisbobbe:
Ah, in fact—I think #4405 would also be solved by the solution we discussed in chat today. 🙂

rajprakash00:
@chrisbobbe Done all the changes requested. Is it OK now?

src/common/Screen.js
```diff
@@ -137,7 +137,11 @@ class Screen extends PureComponent<Props> {
           contentContainerStyle={[padding && styles.padding]}
         >
           <ScrollView
-            contentContainerStyle={[styles.flexed, centerContent && componentStyles.content, style]}
+            contentContainerStyle={[
+              !scrollEnabled ? styles.flexed : null,
```
chrisbobbe, line 141: Let's add a code comment describing the reason for the conditional, just above it. That way the next person won't think it was accidental or poorly considered. 🙂 Maybe something like this:

```diff
           <ScrollView
             contentContainerStyle={[
+              // If `Screen` is responsible for managing scrolling,
+              // keep its childrens' height unbounded. If not, set a
+              // bounded height on its children, e.g., as required for
+              // a child `FlatList` or `SectionList` to work (or even
+              // another `ScrollView`, but hopefully we don't nest too
+              // many of these!).
               !scrollEnabled ? styles.flexed : null,
               centerContent && componentStyles.content,
               style,
```

chrisbobbe (commented):
> @chrisbobbe Done all the changes requested. Is it OK now?

Thanks! It's easier to review more thoroughly now. 🙂

See a small comment on the code, below.

Also, in the commit message, let's include a link to the discussion on CZO. That way, someone reading this in the future will have the entire discussion available to them, so they can be better informed when they make changes on top of this. It could even help them identify any mistakes we might have made here. 🙂 I've mostly settled on a pattern for including links unobtrusively in commit messages. To see what I usually do, try using Greg's ["secret" to using `git log -p`](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md#2-the-secret-to-using-git-log--p) to browse past commits, but include `--author=cbobbe@zulip.com` in the command so it just shows mine. 🙂

Let's change the [prefix](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages) in the summary line (`Fix:`, in this revision) so it indicates what part of the code the change affects. How about `Screen: ` (remember the space after the colon, too), because the change is made in the `Screen` component's file. The doc I've just linked to says that lowercase names are preferred, but in the mobile project Greg and I both freely use uppercase when we're referring to a specific identifier that uses uppercase (you can see this in the commit history, using the `git log -p` "secret" mentioned above).

Then, for the rest of the summary line: remember that the fix isn't limited to `AccountDetailsScreen` and `StreamSettingsScreen`! 🙂 As I mentioned [above](https://github.com/zulip/zulip-mobile/pull/4471#issuecomment-776236154), I expect `AuthScreen` is also affected. Not all the affected components will fit in the summary line, so we'll have to be more general: how about something like

```
Screen: Stop preventing scrolling when `scrollEnabled` is true.`
```

(Also note the period at the end.)

In the body, let's mention that we've taken care to avoid the particular bug that the `styles.flexed` was added to fix.

Finally, since we know of one issue this fixes (by the way, I've just confirmed that it does fix it), let's mention that issue in the commit message.

So, to put it all together 😄, I'm thinking of something similar to this:

```
Screen: Stop preventing scrolling when `scrollEnabled` is true.

While taking care not to cause a particular regression.      
`styles.flexed` was added here, in zulip/zulip-mobile@9fa95a9b7, to 
fix a bug where child `FlatList`s weren't scrolling as they should. 
The `styles.flexed` should be used when `Screen`'s children (not
`Screen` itself) are managing scrolling, but it should not be used
otherwise. See discussion [1].

[1] https://chat.zulip.org/#narrow/stream/48-mobile/topic/Screen.20component.20in.20landscape.20mode/near/1115682

Fixes: #4405
```

src/common/Screen.js
rajprakash00: Done.

chrisbobbe:
Thanks! Merged, in 6a59363ee0f6bd9a2f0f3722f43bf74712ad4fa6.

PR #4473, opened 2021-02-12
eslint: Specify Jest version in config.

chrisbobbe:
This fixes a problem where format-on-save had stopped working for
me, probably after updating vscode-eslint. I'd been getting this
error:

```
Error: Error while loading rule 'jest/no-deprecated-functions':
  Unable to detect Jest version - please ensure jest package is
  installed, or otherwise set version explicitly

Occurred while linting
  /Users/chrisbobbe/dev/zulip-mobile/src/compose/ComposeMenu.js

	at detectJestVersion
	  (/Users/chrisbobbe/dev/zulip-mobile/node_modules/eslint-plugin-jest/lib/rules/no-deprecated-functions.js:38:9)

	at create
	  (/Users/chrisbobbe/dev/zulip-mobile/node_modules/eslint-plugin-jest/lib/rules/no-deprecated-functions.js:61:256)

	[...]
```

Now I don't get that error, and format-on-save works again.

See these issues:

- jest-community/eslint-plugin-jest#686
- microsoft/vscode-eslint#1145

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -51,6 +51,19 @@ settings:
     node:
       extensions: [.js, .ios.js, .android.js, .json]
 
+  jest:
+    # Some Jest rules depend on the Jest version. In particular,
+    #   "jest/no-deprecated-functions". If we don't provide it here,
+    #   that rule will use `process.cwd()` to grab the Jest version
+    #   from node_modules. But our VSCode plugin for ESLint seems to
+    #   be changing the working directory in a way that interferes
+    #   with that, and I haven't found the right value for that
+    #   plugin's `eslint.workingDirectories` setting to fix the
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Hmm, from the CI failure, looks like this isn't quite the way this should be done. I'll take another look soon.

gnprice:
Thanks! I've been seeing the same symptom -- glad to have that fixed. Merged, with a tiny formatting tweak to add a blank line above the start of that `jest:` stanza.

Where did you see that error message? I'm not finding an error message anywhere, even when I'm on the commit before this fix, and hit save, and it doesn't auto-format.

chrisbobbe:
> Where did you see that error message?

If I click on "Prettier" in the bar at the bottom of VSCode, toward the right:

image

Issue #4460, opened 2021-02-01
Crash in getUserFromMention


gnprice:
One of our more widespread issues reported by Sentry is [this one](https://sentry.io/organizations/zulip/issues/1968822455/events/4fa22161f82c49e2b6171095ccb6e67c/):
> undefined is not an object (evaluating 'o.split')

with stack
```js
    const [userFullName, userIdRaw] = unformattedMessage.split('|');
```
in `getUserFromMention` < `MentionWarnings`.

So it seems that `unformattedMessage` here is `undefined`. The fuller source listng is
```js
  getUserFromMention = (completion: string): UserOrBot | void => {
    const { usersById } = this.props;

    const unformattedMessage = completion.split('**')[1];
    const [userFullName, userIdRaw] = unformattedMessage.split('|');
```
so it looks like this would happen if `completion.split('**')` returns an array of less than two elements. In particular this would happen if `completion` doesn't contain `'**'`.

Because this is regularly affecting users, and I believe it's causing a crash for them, it would be good to fix.

I think this is ripe for some debugging.
* The first goal would be to try to reproduce it -- go to the compose box and try to enter some text that triggers this bug.
* A useful step for trying to do that would be to read a bit more of the code that leads to here, and try to see when `getUserFromMention` is called. It seems like that function is assuming that its argument string will contain `**` whenever it gets called, but that that isn't always true.
* Once we have a clear reproduction, I suspect the fix may be relatively easy -- the debugging may be the main part of the work.

im-adithya:
Hey @gnprice! I have tried to reproduce this but failed to do so...
I found that the function getUserFromMention is being called because of handleMessageAutocomplete which is used whenever we type :, # or @... So when the case is @ --> `**Username**` is added by default (hence it is checking whether the lastWordPrefix is '@' or not in order to ensure the presence of **s) which is then used to split to get `unformattedMessage`.
> It seems like that function is assuming that its argument string will contain ** whenever it gets called, but that that isn't always true.

I tried to replicate this by typing @username but it then doesn't trigger autocomplete so that doesn't even get invoked.
So I think it is correct to assume that the function's argument always contains ** whenever it gets called...

Gautam-Arora24:
@gnprice I would like to work on this?

im-adithya:
> @gnprice I would like to work on this?

Hey Gautam, I'm still investigating this issue to reproduce it and will work on this once I find it...

chrisbobbe:
[...]

PR #4474, opened 2021-02-12
crash: Fix crash in getUserFromMention as unformattedMessage gives undefined.

Gautam-Arora24:
**What was the error?**
The **unformattedMessage** in **getUserFromMention** was giving undefined which was the cause of the crash.

**How I solved this issue?**
Since on mentioning the **group** in the ComposeBox was providing the same error. I used a constant called **splittedMessage** which splits the string through "*". This computation can provide me an array of length either 3 or 5. Thus, according to the size, I specified which index's value is further needed in the code. Hence, the bug was solved. 

**Before**
Screenshot-2021-02-05-at-10 52 28-AM

**After changes**
 No error
Screenshot 2021-02-13 at 1 58 09 AM

Fixes: #4460

### DIFF
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -57,7 +57,6 @@ class MentionWarnings extends PureComponent<Props, State> {
     const { allUsersById } = this.props;
 
     const unformattedMessage = completion.split('**')[1];
-    const [userFullName, userIdRaw] = unformattedMessage.split('|');
 
     // We skip user groups, for which autocompletes are of the form
     // `*<user_group_name>*`, and therefore, message.split('**')[1]
@@ -65,6 +64,8 @@ class MentionWarnings extends PureComponent<Props, State> {
     if (unformattedMessage === undefined) {
       return undefined;
     }
+
+    const [userFullName, userIdRaw] = unformattedMessage.split('|');
 
     if (userIdRaw !== undefined) {
       const userId = makeUserId(Number.parseInt(userIdRaw, 10));


chrisbobbe:
Hi, @Gautam-Arora24! I've commented on the issue ([here](https://github.com/zulip/zulip-mobile/issues/4460#issuecomment-779358253)) with a potentially cleaner way of handling this.

In particular, I get the feeling that future readers of the code might prefer not to have to read and understand exactly what it means if `completion.split('**')` has a length of 3 or not. And if that did turn out to be necessary, I'd like to have a very clear code comment, so a future reader (or debugger!) could understand the code just as clearly as you do. 🙂

chrisbobbe:
Thanks, @Gautam-Arora24!

In the summary line, let's use `MentionWarnings:` for the prefix, instead of `crash:`. We tend to use the prefix to help specify ["what part of the code is affected"](https://zulip.readthedocs.io/en/latest/contributing/version-control.html). And the rest of the summary line makes it clear that a crash is being fixed, as well as narrowing it down to a particular function that can be found in the `MentionWarnings` component/file. 🙂

In the commit-message body, let's say "Use a conditional" instead of "Used a conditional". The mobile-app project differs from the main project in explicitly preferring the first form in the commit message body, not just the summary line, to explain the "what" of the commit message (i.e., what work is being done). We should get around to documenting this, but I usually refer to Greg's comment at https://github.com/zulip/zulip-mobile/pull/4079#issuecomment-622188734 to explain our reasoning. 🙂

I notice that the line lengths are well within bounds; thanks! 🙂 I do notice that the lines of the body are a bit uneven, though; the second is quite a lot shorter than the third. Do you mean the third line to begin a new paragraph? I tend to separate paragraphs with an extra line break, e.g.,

```
Use a conditional to check whether the unformattedMessage
is undefined or not.

Now, there is no error related to unformattedMessage as
undefined.
```

Gautam-Arora24:
@chrisbobbe I hope now the commit is fine.

chrisbobbe:
Merged, thanks!

Issue #4340, opened 2020-12-15
 BUG. Sending pictures with apple device doesn't work.


ThomasBaeuml:
Hello,

some IPhone users noticed that sending pictures doesn't work. We try taking pictures and send them doesn't work in the Zulip app ... Also taking photos  with the IPhone camera and just choose them doesn't work. They using IOS 14.2. Is this an Apple bug? Is there a solution for this? Android, Windows client and webapp work for us. Do you need more informations?

Thanks for your answer.

Greetings from germany 

Thomas

pratyaksh123:
I've tested this on my iPhone 7+ running iOS 14.2 , I'm experiencing the same behaviour. Sending pictures through camera doesn't work.

shaurya2612:
@ThomasBaeuml Hi Thomas ! Could you send a short vid/gif of what is exactly happening, as the iphone camera is not available on emulators

pratyaksh123:
@ThomasBaeuml I think this was just a permissions issue.Just go to settings and check the permissions for photos.

ThomasBaeuml:
Hey, 
@pratyaksh123 No zulip has every permissions.
@shaurya2612  Sorry for late response. I was on a vacation. Below you can see the video. 
At first i try du choose a picture from my recordings (Bottom left picture image).
After that i try to take a picture in Zulip and send it... but both doesn't work for all Apple users...

https://user-images.githubusercontent.com/63281013/104292390-16065d00-54bd-11eb-8b64-f6c4592562f6.MP4

ThomasBaeuml:
Are there some news?

chrisbobbe:
Thanks for the report!

There's one bug we're tracking in this area: #2684. We don't yet show a loading indicator while we're waiting for the image to be uploaded. #4129 is open as a way to fix that bug; we should review and resolve that.

But I think you're saying that the image is *never* unloaded, even after waiting a reasonable amount of time, say, a few minutes. Is that right?

There have been a few bugs that have started to surface with iOS 14, like #4365; this could be one of them. But it's possible that some other variable is involved, or that it's not only happening on iOS 14. I did just test on an iPhone 12 Pro simulator running iOS 14.3, and I was able to send a picture.

[...]

PR #4475, opened 2021-02-13
Use `origURL` from react-native-image-picker, when provided.

chrisbobbe:
Major update (marking this as "blocked" until we investigate more): Looks like React Native has fixed issue facebook/react-native#27099 (linked here), with PR facebook/react-native#31457?! 🎉 I'd predict that this fix would be released in RN v0.65.

-----

Part of #4340.

The added comments are pretty absurdly long, and I'd appreciate any recommendations for how to fix that while still having enough detail. 🙂 

This follows the discussion on CZO up to [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/JPEG.20file.20sizes/near/1117131). My comment there still applies:

> > If I use `origURL`, then I can vary the quality in the React Native snippet above, and I don't see a change in the size the server reports: it's 4267773 whether that quality is 1.0, 0.4, or 0.5.
> 
> I'd love to have a better idea of why using `origURL` seemingly lets us bypass React Native's `UIImageJPEGRepresentation` call.

In particular, this change should eliminate the size inflation that happens when selecting an existing image from the media library on iOS—_for some iOS versions_. Unfortunately, we don't know the iOS version at which Apple will decide to remove a deprecated iOS API that this optimization depends on (discussion [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/JPEG.20file.20sizes/near/1116304)).

We'd obviously prefer to use a library that didn't depend on a deprecated API for this optimization. Unfortunately, if we were to upgrade to the latest version of react-native-image-picker, we'd [lose the optimization](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/JPEG.20file.20sizes/near/1116283) entirely. And on the most obvious alternative to react-native-image-picker—expo-image-picker—the symptom of the deprecated API being removed would be much worse than missing out on an optimization: it looks like [the app would just crash](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/JPEG.20file.20sizes/near/1116301).

As mentioned in the code comments, images coming straight from the camera will still be inflated to almost the same level as they were before. The major obstacle to fixing this is that React Native rudely [interferes](https://github.com/facebook/react-native/issues/27099#issuecomment-602016225) by uncompressing and recompressing such an image before uploading it, with a `compressionQuality` of 1.0 (the highest possible value; see discussion on CZO for more detail).

### DIFF
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -78,6 +78,7 @@ class ComposeMenu extends PureComponent<Props> {
       // https://github.com/react-native-community/react-native-image-picker/issues/1271
       fileName: ?string,
     }>,
+    source: 'camera' | 'library',
   ) => {
     if (response.didCancel) {
       return;
@@ -103,7 +104,7 @@ class ComposeMenu extends PureComponent<Props> {
           path: 'images',
         },
       },
-      this.handleImagePickerResponse,
+      response => this.handleImagePickerResponse(response, 'library'),
     );
   };
 
@@ -115,7 +116,9 @@ class ComposeMenu extends PureComponent<Props> {
       },
     };
 
-    ImagePicker.launchCamera(options, this.handleImagePickerResponse);
+    ImagePicker.launchCamera(options, response =>
+      this.handleImagePickerResponse(response, 'camera'),
+    );
   };
 
   handleFilePicker = async () => {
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Marking as blocked for now; see edit at top of PR description.

chrisbobbe:
> Looks like React Native has fixed issue [facebook/react-native#27099](https://github.com/facebook/react-native/issues/27099) (linked here), with PR [facebook/react-native#31457](https://github.com/facebook/react-native/pull/31457)?! 🎉 I'd predict that this fix would be released in RN v0.65.

Welp this was reverted; see: https://github.com/facebook/react-native/issues/33760#issuecomment-1354053780

And react-native-image-picker has changed significantly since I sent this PR, and it's possible that the inflation there has gone away (https://github.com/react-native-image-picker/react-native-image-picker/pull/2006) but we should verify experimentally.

Closing as stale.

Issue #3593, opened 2019-08-22
message list: Show an actual star for starred messages


gnprice:
Following up on #2627 :
> In the message list, we should display an indication of whether a message has been starred. For consistency and ease of understanding, it should look a lot like what we do in the webapp, which is a "star" icon in the upper-right of the message.

We actually already had code to try to show this, but the mechanics of updating the data to drive that were buggy. That's #2676 and was fixed.

Still, UX-wise the small and faint word "starred" leaves room for improvement. This issue is to replace it with an actual star.

Steps to do that will look like:
* Consult the webapp to see what HTML and CSS it uses for this case. The source code will likely be all under `static/` in zulip/zulip. I'd start by inspecting a starred Zulip message in the Chrome DevTools (or another browser's equivalent) to see the final HTML and the relevant CSS rules.
* Add needed bits of HTML to `messageAsHtml.js`, and remove the `starred` bit from `messageTagsAsHtml` there.
* Add needed bits of CSS to `css.js`, and to `cssNight.js` if relevant.
* Use a desktop browser's dev tools to inspect and fiddle with the result. See our doc here: https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#webview

agrawal-d:
@zulipbot claim

agrawal-d:
**@zulipbot** claim

chrisbobbe:
Would you say the star should be tappable to toggle starred/unstarred, or is that too small of a target for people to interact with? Currently all message actions are done by long-pressing a message, which makes sense, but I think users generally expect a star on a message to be tappable, as in Gmail, for example.

ray-kraesig:
My instinct here is that associating an always-visible ☆ or ★ with every message will exacerbate issue #3632, although tapping a star just to remove it probably wouldn't be a bad idea.

Unrelatedly, I wouldn't necessarily suggest trying to use the webapp's star code: it uses FontAwesome, and while we have that in the app already, I recall there being some annoying subtleties involved in making that font accessible to the WebView on iOS.

chrisbobbe:
Hmm, are you mostly concerned with vertical space? An always-visible ☆ or ★ wouldn't necessarily have to add vertical space; it could go on the left of the timestamp, as in the web app (as viewed from either mobile or desktop), right:

![image](https://user-images.githubusercontent.com/22248748/72105850-2badd900-32e3-11ea-8f81-3d2a954d611e.png)

[...]

PR #2758, opened 2018-07-05
Add support for right-to-left messages

Shayan-To:
Add support for correct viewing of right-to-left messages.

Issue #2755, opened 2018-07-03
Add support for right-to-left messages


Shayan-To:
Add support for messages to be right to left. For ex

Shayan-To:
@zulipbot claim

zulipbot:
Hello @Shayan-To!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-mobile/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/zulip-mobile/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

gnprice:
#2758 is a previous PR that aimed to fix this issue. (Thanks @Shayan-To !)

That PR didn't get finished because the author has been busy. We'd still love to have this issue resolved. Anyone is welcome to make a try at it -- just mention here first when you start doing so, to reduce the chance of duplicate work.

When attempting this issue, I recommend reading #2758; it looks like it contains a lot of the pieces that we'll want in a solution, and the review comments on it will be helpful too. Definitely borrow whatever pieces of that PR turn out to be helpful in a new PR.

im-adithya:
@zulipbot claim

chrisbobbe:
Hi, @im-adithya! I've assigned the issue to you. Please see Greg's comment, [above](https://github.com/zulip/zulip-mobile/issues/2755#issuecomment-450105363), about a PR that likely has a lot of useful work you can borrow from. 🙂

im-adithya:
> Hi, @im-adithya! I've assigned the issue to you. 

Thanks @chrisbobbe. 

> Please see Greg's comment, [above](https://github.com/zulip/zulip-mobile/issues/2755#issuecomment-450105363), about a PR that likely has a lot of useful work you can borrow from.

Sure!

PR #4477, opened 2021-02-16
Add support for right-to-left messages

im-adithya:
## Description
Wrapped the message box (message, message-tag and reactions) with a div having dir='auto'

## Additional Fix
I have also styled the scroll-to-bottom button by increasing the button's opacity in place of a blue rectangle.

## Earlier
Screenshot from 2021-02-16 10-59-55

## Now
Screenshot from 2021-02-16 11-53-48


im-adithya:
@gnprice @chrisbobbe Can you please review this? :)

Gautam-Arora24:
@im-adithya I can see that the commits are not following the commit discipline stated in the [docs](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-discipline) which have been told to me by @chrisbobbe several times.

**Things you should keep in mind**
1. Summary line of the commit should contain prefixes.
2. Each summary line should end with a period.
3. Any paragraph content in the commit message should be line-wrapped to about 68 characters per line.

im-adithya:
Yeah, I'm sorry for that! I didn't know these guidelines. Is it possible to change it now?

Gautam-Arora24:
Yes @im-adithya It's very easy to change it. You can use Git's interactive rebase command ```git rebase -i HEAD~n``` where n stands for the number of commits you have to change before the head (n would be 3 in your's case). Then a window will pop up where you can use ```reword``` to change the commit message. After changing the commit message you can force push your branch.

gnprice:
Thanks @im-adithya ! Would you include a "before" screenshot for the same messages? That'd help in understanding the "after" screenshot.

How does it look when the message contains things like a bulleted list, a quote, or a spoiler block? If those look wrong, then to fix them, the places where the CSS contains references to "left" and "right" may need to be converted to say "start" and "end". Take a look at https://github.com/zulip/zulip-mobile/pull/2758/files#r211116856 for some previous discussion.

In the screenshot, it looks kind of disorienting that the tags and reactions appear on the left for some messages and on the right for others. It may work better to put the `dir="auto"` on specifically the message content. See how the previous PR did it:
https://github.com/zulip/zulip-mobile/pull/2758/files#diff-1b9adebe8633a4a43eaee72a0bffa9d21f5cff355635536c0871a8691e83b5ef

How does the webapp handle this? (Or does the webapp handle this?)

im-adithya:
Added the screenshot for your reference @gnprice. Please have a look!

> How does it look when the message contains things like a bulleted list, a quote, or a spoiler block? If those look wrong, then to fix them, the places where the CSS contains references to "left" and "right" may need to be converted to say "start" and "end". Take a look at https://github.com/zulip/zulip-mobile/pull/2758/files#r211116856 for some previous discussion.

It works fine for a bulleted list but fails for quotes and spoiler blocks. In fact, the same thing is observed on the website as well. The quotes also work fine only when the first word in the message is RTL (because it uses dir='auto' which checks only the first letter) which is kinda fine but the spoiler blocks don't even work in pure RTL (both in the website and app) and result in removing the header and dropdown of the spoiler block and result in displaying the inner content `like this`. Reference:

(The message below is the markdown of the one above) 
> In the screenshot, it looks kind of disorienting that the tags and reactions appear on the left for some messages and on the right for others. It may work better to put the `dir="auto"` on specifically the message content.

It was actually done on purpose 😬, as the message tags were always set to align on the right. But, I thought it would be better to shift the tags depending on the dir as it didn't look so good and I have also seen an issue related to removing the starred and other tags anyways... 
(And here I think it is better to copy the style of the mobile site of marking `(EDITED)` using the space on the left like this:)

And for the star, there's already an issue filed as #3593 for which my idea is to include it inside the time tag which pops out on tapping the message.

Also, if we see the above screenshot it appears like the star is actually present on top of the RTL message which shouldn't be the case.

> How does the webapp handle this? (Or does the webapp handle this?)
Yes, it does but has the same errors as mentioned above...

gnprice:
> It works fine for a bulleted list but fails for quotes and spoiler blocks. In fact, the same thing is observed on the website as well.

Hmm interesting! Do we have an issue for that in zulip/zulip? If not, it'd be very helpful to file an issue there with a clear description and repro.

I don't quite follow the screenshot you posted, though. It looks like the spoiler block doesn't begin with ```` ```spoiler ````, but with ```` ``` ```` followed by some text in Arabic script. So that's just a [code block](https://zulip.com/help/format-your-message-using-markdown#code), with syntax highlighting for the language named by that text -- which probably isn't the name of any programming language Zulip knows about :wink: 



> (And here I think it is better to copy the style of the mobile site of marking `(EDITED)` using the space on the left like this:)

Hmm, well, as that screenshot demonstrates, the webapp keeps the layout consistent from message to message: reactions are at the left, and that "(EDITED)" marker is also at the left.

Similarly, we should keep the same overall layout (for the UI surrounding the message) from message to message, regardless of the message contents.

gnprice:
> Also, if we see the above screenshot it appears like the star is actually present on top of the RTL message which shouldn't be the case.

Indeed, well spotted! That would be another issue that would be good to file on zulip/zulip so it can get fixed.

im-adithya:
Done @gnprice! Please review it now!

chrisbobbe:
Thanks, @im-adithya! From Greg's comments above, I see that he was wondering a few things about the screenshots. I think it'd be helpful to post some new screenshots from the new revision of this PR, clarifying the points where he ran into some confusion. 🙂

I also see that Greg mentioned the idea of filing some issues in the zulip/zulip repo. If you've done so, could you post the issue numbers here so the GitHub UI makes it easy to follow links there and back?

im-adithya:
I have updated the screenshots in the first comment itself!

> could you post the issue numbers here?

Actually, the first one wasn't an issue. As Greg said the block doesn't begin with ` ```spoiler `. 
And the second issue star on top of the message has been filed as https://github.com/zulip/zulip/issues/17986

gnprice:
Thanks @im-adithya ! 

Because there were a lot of things discussed in this thread, I wasn't sure initially what the status was. After looking closely, I see that you've revised it to follow this suggestion:

> It may work better to put the `dir="auto"` on specifically the message content. See how the previous PR did it:
> https://github.com/zulip/zulip-mobile/pull/2758/files#diff-1b9adebe8633a4a43eaee72a0bffa9d21f5cff355635536c0871a8691e83b5ef

I just went and played with using the app with this code. It seems to work great for plain text and for not interfering with the surrounding UI outside the message content.

So then I went looking to see what other cases we should be sure to consider in order to get it right. In particular I went back to that same previous PR to see what it did:
https://github.com/zulip/zulip-mobile/pull/2758/files
I noticed there are several bits there related to `code` and `pre` elements. So that's clearly a potential pitfall, and something to try out and make sure is handled right.

For some test data, I went to the "test here" stream on CZO and just searched for where the previous PR's author might have tested things. Here's a thread:
https://chat.zulip.org/#narrow/stream/7-test-here/topic/rtl
And with this version it does not turn out right. It looks like this:



Those inline code snippets should say `code(`, but they read as `)code`. Awfully confusing if you're trying to read some code!

And the code blocks should say `multiline code(!`, but they read as `!)multiline code`.

The webapp has some CSS that looks relevant to this. We should try using the exact same solution, and if that works then use that.

---

A couple of other cases to take care of too -- both of which just come from looking at that previous PR:
* Blockquotes. The current version here ends up with the bar marking the quote on the left, even when the text is way off on the right.
* Lists. The current version here ends up with the bullet points correctly over on the right, but with way more indentation than it should have.

So the work that's needed in order to have a solution to this issue is:
* Try out each of those cases.
* Try copying the exact same bits of CSS that the webapp has that solves this problem -- they all work great currently in the webapp, so that should be possible.
* If for a given case that doesn't seem to work, find another solution. Probably start by figuring out exactly what's happening in the webapp that makes it work there.
* Whatever solution we use, test to confirm that all these cases work correctly in the end.
* Also try to think of other cases that might be tricky -- what the previous author did is an important baseline, but it's quite possible they didn't think of everything! There may also be features that didn't exist yet, or worked differently, when that PR was written.

gnprice:
BTW here's a screenshot of the webapp handling all those cases:
image

See [here](https://chat.zulip.org/#narrow/stream/7-test-here/topic/rtl/near/1170879) for those test messages.

im-adithya:
Thanks, @gnprice! I'll try checking the CSS of the web app and come up with a solution to pass these test cases!

bluddy:
@Shayan-To  do you have those commits or branch where you did this work?

bluddy:
@gnprice What did @Shayan-To refer to by
"When I added RTL support to the web app"?

#2758 appears to be on zulip-mobile so I'm confused what he meant by that.
If we have a fully functional example from the web app, wouldn't that be the easiest reference?

gnprice:
Closing this PR because it doesn't look like it's going to develop further. The issue #2755 is still very much open; we can continue discussion of fixing it there.

PR #4481, opened 2021-02-17
webview errors: Report browser version

### LINKED ISSUE #4452: webview errors: Report browser version
At #4451, which is about an error reported by our code inside the WebView, I wrote:

> Not sure what browser versions this is happening on. [... then some approximate information to make a rough guess ...] (We should report browser versions in WebView error events; I'll file an issue for that.)

This is that issue.

The relevant code is first in `js.js` at this call:
```js
  sendMessage({
    type: 'error',
    details: {
      message,
      source,
      line,
      column,
      error,
    },
  });
```
where we'll want to include one more property in that "details" object, for the browser version. This might be just the user-agent string, or possibly there's some better-structured information we can get our hands on. Because this is an Android or iOS webview, the only browsers we're concerned with here are Chrome and Safari.

Then that information gets passed up through `handleOutboundEvents.js`, and a `logging.error` call there sends it to Sentry.

im-adithya:
## Description
I have updated the details object to include additional 
userAgent property which returns browser version during errors.

## Issue(s) linked to this PR
This PR closes #4452 

## Screenshots

In **react-native terminal**:

Screenshot from 2021-02-18 01-46-10

In **adb logcat**:

Screenshot from 2021-02-18 01-46-28

In **app**:

### DIFF
--- a/src/webview/handleOutboundEvents.js
+++ b/src/webview/handleOutboundEvents.js
@@ -109,6 +109,7 @@ type WebViewOutboundEventError = {|
     source: string,
     line: number,
     column: number,
+    userAgent: string,
     error: Error,
   },
 |};
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -405,11 +405,13 @@ var compiledWebviewJs = (function (exports) {
   };
 
   window.onerror = function (message, source, line, column, error) {
+    var userAgent = window.navigator.userAgent;
+
     if (window.enableWebViewErrorDisplay) {
       var elementJsError = document.getElementById('js-error-detailed');
 
       if (elementJsError) {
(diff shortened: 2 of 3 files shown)


im-adithya:
Please review this :D

gnprice:
Thanks @im-adithya ! Merged, with a few tweaks:
* I squashed the three commits together, because each one of the changes in them only makes sense combined with the others. (In particular the first commit alone, or the first two commits alone, don't pass our tests.) See discussion of coherent commits here: https://zulip.readthedocs.io/en/latest/contributing/version-control.html
* I also rewrote the commit message, to align with our style. See the later part of that same doc, and also our [mobile style guide](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#github-prs-issues).
* I added a commit on top to leave the user-agent out from that red error banner, in part because it actually only shows up in development. (See its commit message for more details.) This also wasn't at all obvious from the code -- it's perfectly reasonable if you didn't notice it was development-only -- so I added some comments in the code to hopefully make that clearer.

Issue #4458, opened 2021-02-01
"t.get is not a function" on state.narrows at startup


gnprice:
Looking at our [Sentry errors that are new in v27.158](https://sentry.io/organizations/zulip/issues/?project=191284&query=is%3Aunresolved+firstRelease%3Acom.zulipmobile%4027.158%2B158&sort=user&statsPeriod=30d), the current beta, [this one](https://sentry.io/organizations/zulip/issues/2138980382/events/oldest/?project=191284&query=is%3Aunresolved+firstRelease%3Acom.zulipmobile%4027.158%2B158&sort=user) has affected a substantial number of our beta users:
> t.get is not a function. (In 't.get(v.ALL_PRIVATE_NARROW_STR)', 't.get' is undefined)

where the stack is
```js
    const pmIds = narrows.get(ALL_PRIVATE_NARROW_STR) || NULL_ARRAY;
```
in `getPrivateMessages` < reselect < `UnreadCards`'s mapStateToProps < react-redux < redux-persist dispatching its REHYDRATE action. It looks like this would cause a crash; the symptom might look like staying at the loading screen forever upon launch.

So here `narrows.get` is undefined. Which means `narrows` wasn't *itself* undefined -- else we'd have had an error trying to read the property `narrows.get` -- but it's some object that doesn't have a `get` property. E.g. it could be an empty object.

This is concerning because `narrows` there comes straight from a selector that returns `state.narrows`, and the latter is supposed to be an `Immutable.Map`. Which would certainly have a `get` property.

This error is the kind of thing that could happen if a migration didn't run. Indeed, in previous releases including v27.157, `state.narrows` was a plain Object used as a map, and it would not have had a `get` property. But we did have a migration for that, and indeed v27.158 has some later migrations which are simply `dropCache`; I don't see anything wrong with our migrations here. So if the migration *did* go wrong, or somehow get skipped, that's concerning because it suggests something wrong in the underlying layer that manages migrations.

I think we won't want to roll the release out to users in general until we've fixed this, or otherwise are confident it won't affect lots of users in general. So that makes this a top priority.

chrisbobbe:
Ack; thanks for reporting this.

gnprice:
> So here `narrows.get` is undefined. Which means `narrows` wasn't _itself_ undefined -- else we'd have had an error trying to read the property `narrows.get` -- but it's some object that doesn't have a `get` property. E.g. it could be an empty object.

[...]

PR #4482, opened 2021-02-18
Fix "t.get is not a function" on state.narrows at startup.

chrisbobbe:
Thanks to @gnprice for his investigation and diagnosis of this issue!

I've gone pretty light on detail in code comments, while being sure to link to the relevant conversation on CZO. I'm wary of muddling an already-subtle area of the code with another wall of text, especially for a temporary fix that we expect will be replaced with something more thorough soon.

Fixes: #4458


gnprice:
Thanks! Merged, with a small comment-tweak on top.

PR #4483, opened 2021-02-18
webview errors: Terminate if HTMLTimeElement is undefined

### LINKED ISSUE #4451: HTMLTimeElement is not defined, says a WebView error
Looking in Sentry at our WebView errors (while investigating #4449), I'm seeing error reports that look like [this one](https://sentry.io/organizations/zulip/discover/zulip-mobile:e9525e624b6a4774889bf8e3efeebfb8/):

> Uncaught ReferenceError: HTMLTimeElement is not defined
> source: file:///android_asset/webview/index.html
> column: 53
> line: 276

The impact of an error like that is:
* A red banner shows at the top of the message list, saying "Oh no! Something went wrong. Try again?". This means the user knows that they're seeing a bug.
* Whatever our code inside the webview was trying to do at the time got interrupted. This can mean that arbitrary other functionality in the webview breaks.
  * For this particular error, I think the main effect is likely to be that (a) scroll behavior will be wrong, because the last part of `handleInboundEventContent` doesn't get to run; (b) if visiting a conversation that's so short it fits on one screen, the messages won't get marked as read. Basically one annoying symptom for long conversations, and a different annoying symptom for short ones :slightly_smiling_face:

We don't currently have a great way to sort through the different WebView errors -- they all get aggregated together -- but when I look through a [firehose](https://sentry.io/organizations/zulip/discover/results/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=191284&query=message%3A%22WebView+error%22&sort=-timestamp&statsPeriod=24h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) of the most recent of them and just sample a few to look at individually, the ones I find are
* some instances of #4449 which appear to be from myself;
* and some of these, from at least a few different people (I see three different OS versions.)

So this is at least one of the most frequent WebView errors we have, likely the majority of them. And WebView errors in aggregate are one of our handful of [most frequent and widespread types of errors](https://sentry.io/organizations/zulip/issues/?project=191284&query=is%3Aunresolved+level%3Aerror&sort=user).

From the error message, it's presumably thrown here:
```js
const rewriteTime = (element: Element | Document) => {
  // Find the time elements to act on.
  const timeElements = [].concat(
    element instanceof HTMLTimeElement ? [element] : [],
    Array.from(element.getElementsByTagName('time')),
  );
```
(We have one other reference too, later in that function.)

Seems like the browser doesn't have `HTMLTimeElement`.

Not sure what browser versions this is happening on. The handful of these events I'm looking at are on Android versions from 6.0 to 7.1.2. According to our notes in docs/architecture/platform-versions.md, Android 7 Nougat shipped with Chrome 51, so the device with Android 7.1.2 must have at least that new of a Chrome. (We should report browser versions in WebView error events; ~~I'll file an issue for that.~~ filed as #4452.) The great majority of Android devices have a current version of Chrome, no matter how old they started with, but some don't.

Our support levels at the top of `js.js` say we currently accept graceful degradation for older than Chrome 58 (which shipped with Android 8 Oreo.) So it's OK if this particular feature (the Zulip ["mention a time" feature](https://zulip.com/help/format-your-message-using-markdown#mention-a-time)) doesn't work on those browsers; but we need to ensure core functionality keeps working, which we support back to Chrome 44 which shipped with Android 6.

To fix this issue, it will probably suffice to check if HTMLTimeElement exists, and have `rewriteTime` just return without doing anything if it does. Before merging that solution, the work required in order to verify it is:
* Reproduce the error. The easiest way to do this is probably to run the app in an Android 6 Marshmallow emulator. (Stock emulator images keep the original Chrome version that their version of Android shipped with.)
* Verify that on Chrome 58, the "mention a time" feature works properly. The easiest way to do this is probably to run the app in an Android 8 Oreo emulator. Double-check that the Chrome version you're getting is indeed Chrome 58.

im-adithya:
## Description
HTMLTimeElement is supported from 62 in chrome android.

So, I checked whether it exists or not and ended the function if it doesn't exist
as it prevents the error which has further impacts as mentioned by @gnprice 
in #4451 

## Issue(s) linked to this PR

This PR closes #4451

### DIFF
--- a/docs/architecture/platform-versions.md
+++ b/docs/architecture/platform-versions.md
@@ -192,6 +192,21 @@ General policy:
     emulator image for some past Android release, for practicality of
     testing.
 
+ * Further, we allow *graceful degradation* on all but recent versions
+   of each browser.  This means that although core functionality needs
+   to work, for fancier features of Zulip it's acceptable for them not
+   to be available on older browsers.
+
+   * Typically this will apply to versions older than those shipped in
+     about the last two years' OS releases.  Like the minimum support
+     thresholds, these are described in the block comment at the top of
+     `js.js`.
+
+   * (Implicitly we do much the same thing with old OS versions; we
+     just have less code that interacts with specific OS versions than
+     with browser versions, so we don't write down explicit threshold
+     versions there.)
+
 
 ### Data and commentary
 
@@ -223,6 +238,8 @@ Empirical details on Android Chrome versions found in WebViews:
   * Android versions 5 L, 6 M, 7 N, 8 O had originally shipped
     with Chrome versions 37, 44, 51, 58 respectively (based on
     looking at stock emulator images.)
+    * Later data: Android 9, 10, 11 ship with Chrome versions
+      69, 74, 83 respectively.
   * At that time about 17% of our Android users were on Android
     versions <=6 M -- far more than the 2% or so of Android users with
     Chrome versions older than what shipped with Android 7 N.
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -269,6 +269,10 @@ var compiledWebviewJs = (function (exports) {
   };
 
   var rewriteTime = function rewriteTime(element) {
+    if (typeof HTMLTimeElement !== 'function') {
+      return;
+    }
+
     var timeElements = [].concat(element instanceof HTMLTimeElement ? [element] : [], Array.from(element.getElementsByTagName('time')));
     timeElements.forEach(function (elem) {
       if (!(elem instanceof HTMLTimeElement)) {
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -25,10 +25,10 @@ import { toggleSpoiler } from './spoilers';
  *   Graceful degradation is acceptable below iOS 13 / Mobile Safari 13.
  *
  * * For Android, core functionality needs to work on Chrome 44.
- *   Graceful degradation is acceptable below Chrome 58.
+ *   Graceful degradation is acceptable below Chrome 74.
  *
  *   * These versions are found in stock images for Android 6 Marshmallow
- *     and Android 8 Oreo, respectively, for convenient testing.
+ *     and Android 10, respectively, for convenient testing.
  *
  *   * (Note that Android's Chrome auto-updates independently of the OS, and
  *     the large majority of Android users have a fully-updated Chrome --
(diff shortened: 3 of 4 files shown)


gnprice:
Thanks @im-adithya ! 

It looks like this carries out the solution I mentioned at the end of #4451. As I said there, most of the work required for completing this issue will be in verifying that the solution works as intended.

How does the "mention a time" feature behave after this change, on an affected browser version? It's OK if the feature doesn't work there, but it's important that core functionality should keep working.

Also, how does this work on Chrome 58 in particular? As mentioned at #4451, that's the oldest version that we currently try to fully support (i.e. to keep all features working, not just core functionality.)

---

Separately: please squash together these changes into one commit. See our guide on clear and coherent commits:

im-adithya:
> How does the "mention a time" feature behave after this change, on an affected browser version? It's OK if the feature doesn't work there, but it's important that core functionality should keep working.

It works absolutely fine in the supported browsers (62 and above), but it shows the raw time in unsupported ones (no errors or functionality-breaks though) like this:
Screenshot_1615126287

> Also, how does this work on Chrome 58 in particular? As mentioned at #4451, that's the oldest version that we currently try to fully support (i.e. to keep all features working, not just core functionality.)

The version that my emulator uses is 51. But according to the Browser compatibility table above, it must work similarly till 61.
Screenshot_1615125350

im-adithya:
> Please squash together these changes into one commit.

Done @gnprice :D

gnprice:
> > Also, how does this work on Chrome 58 in particular? As mentioned at #4451, that's the oldest version that we currently try to fully support (i.e. to keep all features working, not just core functionality.)
> 
> The version that my emulator uses is 51.

This question ended up not being critical to answer, because of that helpful screenshot showing exactly what it looks like at Chrome 51 and because I updated the full-support threshold to a later version (Chrome 74). But in case you need to test at a particular version of Chrome in the future, here's how you do it -- quoting from our `js.js` (from before the changes I just merged):

```
 *   Graceful degradation is acceptable below Chrome 58.
 *
 *   * These versions are found in stock images for Android 6 Marshmallow
 *     and Android 8 Oreo, respectively, for convenient testing.
```

So to get webviews with Chrome 58, you set up an emulator image running Android 8 Oreo.

There's more notes in our `platform-versions.md` doc:
```
  * Android versions 5 L, 6 M, 7 N, 8 O had originally shipped
    with Chrome versions 37, 44, 51, 58 respectively (based on
    looking at stock emulator images.)
    * Later data: Android 9, 10, 11 ship with Chrome versions
      69, 74, 83 respectively.
```
So you can get any of the above Chrome versions for your WebViews by installing an emulator image with the respective Android version. Your emulated device with Chrome 51 is presumably running Android 7 Nougat.

Getting other Chrome versions (in between those versions) is more annoying... but that's why we deliberately choose our support thresholds to always be one of these particular versions :wink:

im-adithya:
Oh! So that's how it works! I was trying to downgrade the chrome version in the emulator all the time. I'll set up some more emulator images now!

PR #4485, opened 2021-02-18
StatusBar: Fix bug causing invalid color names

WesleyAC:
Color.darken() can return a color that is rounded in a way that causes
this error when used:

`StatusBar._updatePropsStack`: Color hsl(186.39999999999998, 51.9%, 70.9%) parsed to null or undefined

This seems to be because something cannot handle a color with that many
decimal points of precision. This doesn't result in a visible bug (my
guess is that it gets passed in as a string, and then the rendering
engine can handle it just fine, although I haven't validated that), but
it does create a bunch of noise in the logs, which this commit fixes.


gnprice:
Thanks @WesleyAC ! Merged.

Issue #4480, opened 2021-02-16
TextInput: Increase Text opacity in landscape mode in nightmode


rajprakash00:

abhi0504:
@rajprakash00 are you working on this issue ?

rajprakash00:
@abhi0504 Yes I'm working on it

PR #4486, opened 2021-02-18
Input: Fix text color in landscape orientation in nightmode.

rajprakash00:
**Description:**
In landscape orientation in night mode,the color of text written in
`<TextInput>` was too much transparent to be viewed properly.
Applied a different shade of previous color to fix it.

**Screenshots:**

zulipAfter2
zulipAfterPortrait2

Fixes: #4480

### DIFF
--- a/src/styles/theme.js
+++ b/src/styles/theme.js
@@ -6,6 +6,7 @@ import type { ThemeName } from '../reduxTypes';
 
 export type ThemeData = {|
   color: string,
+  textColor: string,
   backgroundColor: string,
   cardColor: string,
   dividerColor: string,
@@ -14,6 +15,7 @@ export type ThemeData = {|
 export const themeData: { [name: ThemeName | 'light']: ThemeData } = {
   night: {
     color: 'hsl(210, 11%, 85%)',
+    textColor: 'hsl(200, 11%, 65%)',
     backgroundColor: 'hsl(212, 28%, 18%)',
     cardColor: 'hsl(212, 31%, 21%)',
     // Dividers follow Material Design: opacity 12% black or 12% white.
@@ -22,6 +24,7 @@ export const themeData: { [name: ThemeName | 'light']: ThemeData } = {
   },
   light: {
     color: 'hsl(0, 0%, 20%)',
+    textColor: 'hsl(0, 0%, 20%)',
     backgroundColor: 'white',
     cardColor: 'hsl(0, 0%, 97%)',
     // Dividers follow Material Design: opacity 12% black or 12% white.
--- a/src/common/Input.js
+++ b/src/common/Input.js
@@ -82,7 +82,7 @@ class Input extends PureComponent<Props, State> {
 
     return (
       <TextInput
-        style={[this.styles.input, { color: this.context.color }, style]}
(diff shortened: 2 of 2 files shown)


src/styles/theme.js
```diff
@@ -6,6 +6,7 @@ import type { ThemeName } from '../reduxTypes';
 
 export type ThemeData = {|
   color: string,
```
chrisbobbe, line 8: `color` has been what we use for text (see e.g. `Label`); let's not add something new with such a similar role.

The problem here is that the text input's full-screen UI (which is Android-only) doesn't respect the theme choice; the background is showing up white even when the user has selected dark mode. I suspect that's because the native code doesn't have a piece of state that tracks with the theme selection we make at the JavaScript level and is stored in Redux. If it did, I think it's plausible that the native UI element would consult that and we'd see it using a dark background.

It looks like you can disable the full-screen UI by setting a `TextInput`'s [`disableFullScreenUI`](https://reactnative.dev/docs/textinput#disablefullscreenui) to `true`. Unfortunately, when I tried that just now, the keyboard took up so much room that the layout glitched and I couldn't see any of the text input. So that doesn't seem like a solution here.

It doesn't look like `TextInput` gives you any control over what the full-screen UI looks like; just whether it gets used or not.

I'd be inclined toward a solution that makes the native-code layer aware of what theme we want its elements to naturally present in. I suspect that might start with a change to `android/app/src/main/res/values/styles.xml` like the one made to React Native's template app in facebook/react-native@2b56011f9; we'll take that naturally (or do some other appropriate thing) with the RN v0.64 upgrade, #4426.

If we change the text color, we'd want to make sure it's done in a really targeted way: only on Android (use `Platform.OS` for this), only here in the compose box (or, I guess, in any text input that doesn't set `disableFullScreenUI`), and only when in landscape mode (grab `state.session.orientation` in the `connect` call for that).

What do you think; want to try that targeted change, or wait until the RN v0.64 upgrade?

chrisbobbe (changes requested):
Thanks, @rajprakash00! Are those before or after screenshots? It'd be good to have both before and after. But before you go get more, please see my comment, which indicates some different directions in which I think the fix should be made. 🙂

src/styles/theme.js
rajprakash00: >  I suspect that's because the native code doesn't have a piece of state that tracks with the theme selection we make at the JavaScript level and is stored in Redux

I didn't know about this big problem with `TextInput`. Thanks for the info 🙂

> I'd be inclined toward a solution that makes the native-code layer aware of what theme we want its elements to naturally present in.

Hmm.. agreed with this logic .

> If we change the text color, we'd want to make sure it's done in a really targeted way: only on Android (use `Platform.OS` for this), only here in the compose box (or, I guess, in any text input that doesn't set `disableFullScreenUI`), and only when in landscape mode (grab `state.session.orientation` in the `connect` call for that).
> 
> What do you think; want to try that targeted change, or wait until the RN v0.64 upgrade?

Um.. this seems a good deal of work,I don't know it's priority at this moment ,nevertheless I'll surely want to try and implement this (I don't know how far I can go, but seems worth a try )

PR #4487, opened 2021-02-19
Improve the debugging doc a bit.

chrisbobbe:
Some of these ideas come from [this discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/webview.20log/near/1120527) on CZO.


gnprice:
Thanks -- merged!

Issue #4039, opened 2020-04-13
Swipe an image down to close it.


chdinesh1089:
Instead of clicking on `x` to close an image, dragging an image down to close it will make it much easier and user-friendly. It also improves reachability issues when using one-handed.

An example would be WhatsApp's image viewing.

ray-kraesig:
I can understand swiping left or right to close the lightbox, but I'm surprised by swiping *down*.

Does WhatsApp also have their lightbox-equivalent slide up from the bottom when it comes into view?

agrawal-d:
> I can understand swiping left or right to close the lightbox, but I'm a bit surprised by swiping _down_.
> 
> Does WhatsApp also have their lightbox-equivalent slide up from the bottom when it comes into view?

Swiping down is the only way ( apart from the back button ) to close the lightbox there.


In almost every app, swiping left / right does not close the lightbox, becaue usually the lighbox is a gallery of of images and left / right swipes would change the image displayed.

chdinesh1089:
> I can understand swiping left or right to close the lightbox, but I'm a bit surprised by swiping down.

Are you saying that we swipe from edge which acts as clicking back on android?

> Does WhatsApp also have their lightbox-equivalent slide up from the bottom when it comes into view?

No, it just opens up in a pop up kind of fashion. I think I understand how you are imagining. Are you thinking that it looks like a different container that can be closed by sliding down?

In the issue I was trying to say that we should be able to close the image by swiping it down anywhere on the image.

As @agrawal-d said, swiping left or right takes us to previous/next image on whatsapp and gallery. I don’t really know how it works on android, I’m guessing it’ll be same. I’m thinking that on android we can also close the image by swiping from left/right edge as that acts as back button.

chrisbobbe:
I think swiping down to dismiss makes the most sense in a slightly different case than ours.

Check out the GIFs at the demo here:

https://github.com/oblador/react-native-lightbox#demo

There, the image expands to fill the screen, as an overlay on top of the current route—there isn't another route that we navigate to. So it can make sense that you could drag it and swipe in pretty much any direction, and it would go away. In the GIF, they show it going away if you swipe down.

In our app, we do navigate to a different route to show the lightbox. So, to make it go away, we're pretty limited to using whatever "go back" is in our navigation system. Swiping down isn't used in any other screens to make them go away, and it would be surprising and pattern-breaking to add it here.

[...]

PR #4489, opened 2021-02-19
Experiment: Set up to use full-screen modals, React Navigation's way.

chrisbobbe:
EDIT: I haven't compared this approach with using React Native's [`Modal`](https://reactnative.dev/docs/modal).

-----

@gnprice and I discussed this on the phone the other day, and we're a bit cautious about trusting React Navigation's magic here, but we thought it'd be good to try this out and see if we like it. The main question to answer is whether this setup, using React Navigation's API, is helpful; individual details about the lightbox might be good to talk about but won't necessarily answer that conclusively.

Here we have a few nav-structure-changing commits, followed by turning the lightbox into a full-screen modal.

On iOS, I like how you can swipe down from the top to dismiss the modal (#4039). It doesn't seem to work if your finger starts out _on_ the image, which makes some sense because the image handles gestures to pan/zoom. Still, other apps manage to handle both types of gestures: I see that Facebook on iOS lets you pan/zoom an image that's open in a lightbox modal, but it also recognizes when it's impossible to pan (the image is fully visible and zoomed out; this is the image's initial state anyway) and in that case a swipe-down closes the modal. More awkwardly, while I seem to be able to dismiss the modal by swiping down from a point *above* the image, I don't seem to be able to dismiss it by swiping down from a point *below* the image. I'm not sure how much of this is due to React Navigation, and how much to our react-native-photo-view.

See Apple's [HIG document on modals](https://developer.apple.com/design/human-interface-guidelines/ios/app-architecture/modality/). I'm not sure what the corresponding doc for Android might be; [this "Dialogs" page](https://developer.android.com/guide/topics/ui/dialogs) comes up first in my search, but a brief look suggests it's mostly for "yes/no/OK" dialogs, not something like the lightbox? Does the lightbox not deserve special treatment on Android, as I think it does on iOS?

### DIFF
--- a/src/nav/MainStackScreen.js
+++ b/src/nav/MainStackScreen.js
@@ -31,7 +31,6 @@ import DiagnosticsScreen from '../diagnostics/DiagnosticsScreen';
 import VariablesScreen from '../diagnostics/VariablesScreen';
 import TimingScreen from '../diagnostics/TimingScreen';
 import StorageScreen from '../diagnostics/StorageScreen';
-import LightboxScreen from '../lightbox/LightboxScreen';
 import CreateGroupScreen from '../user-groups/CreateGroupScreen';
 import InviteUsersScreen from '../streams/InviteUsersScreen';
 import StreamSettingsScreen from '../streams/StreamSettingsScreen';
@@ -60,7 +59,6 @@ export type MainStackNavigatorParamList = {|
   'search-messages': RouteParamsOf<typeof SearchMessagesScreen>,
   users: RouteParamsOf<typeof UsersScreen>,
   language: RouteParamsOf<typeof LanguageScreen>,
-  lightbox: RouteParamsOf<typeof LightboxScreen>,
   'create-group': RouteParamsOf<typeof CreateGroupScreen>,
   'invite-users': RouteParamsOf<typeof InviteUsersScreen>,
   diagnostics: RouteParamsOf<typeof DiagnosticsScreen>,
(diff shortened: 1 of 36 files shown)


chrisbobbe:
(Just rebased.)

chrisbobbe:
> (Just rebased, resolving some conflicts.)

(And again.)

src/nav/MainStackScreen.js
```diff
@@ -87,7 +88,10 @@ const Stack = createStackNavigator<
   MainStackNavigationProp<>,
 >();
 
-type Props = $ReadOnly<{||}>;
+type Props = $ReadOnly<{|
+  navigation: RootStackNavigationProp<'main-stack'>,
+  route: RouteProp<'main-stack', void>,
+|}>;
 
 export default function MainStackScreen(props: Props) {
```
WesleyAC, line 94: Same here about a comment — now that we have a more complicated hierarchy, it'd be good to document it more.

WesleyAC (commented):
A couple comments, the code looks good other than that.

I tried it on Android, and it causes a flash of `BRAND_COLOR` when i open the lightbox. Any idea what's up with that?

WesleyAC:
Here's a video of what I see: https://user-images.githubusercontent.com/5001092/122320509-3193ea80-ced7-11eb-8c9a-6ba9021a0c71.mp4

chrisbobbe:
Thanks for the review! Closing as not planned after all: https://chat.zulip.org/#narrow/stream/2-general/topic/Flutter/near/1492568

Issue #4407, opened 2021-01-14
Black screen appears when user click on `Share` option.


nikhilmaske-2001:
**Issue Description:**
Black screen appears when user click on `Share` option. It's a bug and it is also discussed in this [thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Black.20screen.20on.20share).

**How to Reproduce**

Step 1 : Open **Zulip v27.157** android app.
Step 2 : Long tap on any message.
Step 3 : Choose `Share` option.

**Screenshot** 
![Black screen](https://user-images.githubusercontent.com/59444243/104544801-b344cb00-564e-11eb-9558-70ed9ffd1ced.png)

abhi0504:
@nikhilmaske-2001 are you working on this or can I take this?

nikhilmaske-2001:
No @abhi0504 I am not working on it. You can take it.

gnprice:
Thanks for the report!

This reproduces for me on v27.158 too, the current beta release. That was on a Pixel 4 running Android 11.

It did *not* reproduce for me when I tried it just now on iOS: specifically on v27.157, on an iPhone 7 running iOS 12.4.1. Sharing is a feature that involves some platform-specific code, so it's natural that a bug like this might exist on only one platform. I've marked the issue with our "area Android" label because it looks like it's specifically on Android.

gnprice:
For cross-reference: there's some later discussion in [this chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Black.20screen.20appears.20when.20user.20click.20on.20.60Share.60.20option.2E/near/1119224).

gnprice:
The upstream issue https://github.com/expo/react-native-action-sheet/issues/203 has been closed as fixed, and we've upgraded to the version with the fix. That happened in #5203, for the v27.181 release.

I just tried the original repro recipe (thanks @nikhilmaske-2001 for a nice clear recipe!) in v27.182, the current beta, and I no longer reproduce.

So I believe this has been fixed! Thanks again @nikhilmaske-2001 for the report, and @abhi0504 for the debugging and the report upstream.

PR #4490, opened 2021-02-19
messageActionSheet : using `Modal` instead of `@expo/react-native-action-sheet`

abhi0504:
@nikhilmaske-2001 reported an issue #4407 that is :
when we long press on any message then `messageActionSheet` opens up and it took a extra black screen on `Share` option when we share our message 
104544801-b344cb00-564e-11eb-9558-70ed9ffd1ced

[here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Black.20screen.20appears.20when.20user.20click.20on.20.60Share.60.20option.2E/near/1106837) i and @gnprice have discussed about the problem and found that this issue coming from `@expo/react-native-action-sheet` for which i have made a tiny app and report this issue on facebook/expo page [here](https://github.com/expo/react-native-action-sheet/issues/203) and also created a snack for better explaination of the issue [here](https://appetize.io/embed/xc1w6f1krd589zhp22a0mgftyw?autoplay=false&debug=true&device=nexus5&deviceColor=black&embed=true&launchUrl=exp:%2F%2Fexpo.io%2F@abhi0504%2F0d4260%2BIjv94!N!zI&orientation=portrait&screenOnly=false&xDocMsg=true&xdocMsg=true&params=%7B%22EXKernelLaunchUrlDefaultsKey%22:%22exp:%2F%2Fexpo.io%2F@abhi0504%2F0d4260%2BIjv94!N!zI%22,%22EXKernelDisableNuxDefaultsKey%22:true%7D&scale=75&osVersion=8.1)

then we have decide to use  `Modal` instead of @expo/react-native-action-sheet` as it restricts us upto a certain extent on basis
of UI for eg if we use `Modal` instead of `@expo/react-native-action-sheet` then we have many options for eg colored messageActionSheet , popping out with diff animations , can set some icons before any option which and greg have
already discussed [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Black.20screen.20appears.20when.20user.20click.20on.20.60Share.60.20option.2E/near/1113084)

Also we can do many things if we use `Modal` and can tweak our `messageActionSheet`'s UI to an another level for eg 
WhatsApp-Image-2021-02-04-at-3 18 02-PM

- We can add icons in starting of different options of our `messageActionSheet` 
- We can make our text colorful in `messageActionSheet`
- We can add different animations in our `messageActionSheet` 

Fixes: #4407

### DIFF
--- a/src/chat/ChatScreen.js
+++ b/src/chat/ChatScreen.js
@@ -1,8 +1,7 @@
 /* @flow strict-local */
-import React from 'react';
+import React, { useState } from 'react';
 import { View } from 'react-native';
 import { useIsFocused } from '@react-navigation/native';
-import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 
 import { useSelector, useDispatch } from '../react-redux';
 import type { RouteProp } from '../react-navigation';
@@ -23,6 +22,7 @@ import { getLoading, getSession } from '../directSelectors';
 import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './narrowsSelectors';
 import { getStreamColorForNarrow } from '../subscriptions/subscriptionSelectors';
+import { ActionSheetModalHandler } from '../message/messageActionSheet';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'chat'>,
@@ -113,43 +113,48 @@ export default function ChatScreen(props: Props) {
   const showMessagePlaceholders = haveNoMessages && isFetching;
   const sayNoMessages = haveNoMessages && !isFetching;
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
+  const [modalVisible, setModalVisible] = useState(false);
 
   const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
 
+  const longPressModalHandler = () => {
+    setModalVisible(!modalVisible);
+  };
+
   return (
-    <ActionSheetProvider>
-      <View style={[componentStyles.screen, { backgroundColor }]}>
-        <KeyboardAvoider style={styles.flexed} behavior="padding">
-          <ZulipStatusBar backgroundColor={streamColor} />
-          <ChatNavBar narrow={narrow} editMessage={editMessage} />
-          <OfflineNotice />
-          <UnreadNotice narrow={narrow} />
-          {(() => {
-            if (!isNarrowValid) {
-              return <InvalidNarrow narrow={narrow} />;
-            } else if (fetchError !== null) {
-              return <FetchError narrow={narrow} error={fetchError} />;
-            } else if (sayNoMessages) {
-              return <NoMessages narrow={narrow} />;
-            } else {
-              return (
-                <MessageList
-                  narrow={narrow}
-                  showMessagePlaceholders={showMessagePlaceholders}
-                  startEditMessage={setEditMessage}
-                />
-              );
-            }
-          })()}
-          {showComposeBox && (
-            <ComposeBox
-              narrow={narrow}
-              editMessage={editMessage}
-              completeEditMessage={() => setEditMessage(null)}
-            />
-          )}
-        </KeyboardAvoider>
-      </View>
-    </ActionSheetProvider>
+    <View style={[componentStyles.screen, { backgroundColor }]}>
+      <KeyboardAvoider style={styles.flexed} behavior="padding">
+        <ZulipStatusBar backgroundColor={streamColor} />
+        <ChatNavBar narrow={narrow} editMessage={editMessage} />
+        <OfflineNotice />
+        <UnreadNotice narrow={narrow} />
+        {(() => {
+          if (!isNarrowValid) {
+            return <InvalidNarrow narrow={narrow} />;
+          } else if (fetchError !== null) {
+            return <FetchError narrow={narrow} error={fetchError} />;
+          } else if (sayNoMessages) {
+            return <NoMessages narrow={narrow} />;
+          } else {
+            return (
+              <MessageList
+                narrow={narrow}
+                showMessagePlaceholders={showMessagePlaceholders}
+                startEditMessage={setEditMessage}
+                handleLongPressModal={longPressModalHandler}
+              />
+            );
+          }
+        })()}
+        {showComposeBox && (
+          <ComposeBox
+            narrow={narrow}
+            editMessage={editMessage}
+            completeEditMessage={() => setEditMessage(null)}
+          />
+        )}
+        <ActionSheetModalHandler modalVisible={modalVisible} modalHandler={longPressModalHandler} />
+      </KeyboardAvoider>
+    </View>
   );
 }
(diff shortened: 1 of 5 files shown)


chrisbobbe:
Thanks for the PR, @abhi0504! 

I see that GitHub is reporting some CI failures; let's get those fixed up. 🙂 If you're not sure how to fix something, please ask. You should start by running `tools/test`, and a list of problems will appear. I see failures reported by Flow, ESLint, and Prettier.

Also, could you please post some screenshots of the new UI on Android, in both night and day modes? Here are some for iOS:

 

-----

- The first thing I notice is that this is very different from what it was on iOS before. 🙂 That might be OK; Apple's design guidelines do allow [modal interactions](https://developer.apple.com/design/human-interface-guidelines/ios/app-architecture/modality/). I think the decision to stop using Apple's native action sheet will depend on how closely our modal's purpose matches what Apple says [action sheets](https://developer.apple.com/design/human-interface-guidelines/ios/views/action-sheets/) are for. I haven't yet thought deeply enough about this; in particular, we'd want to consider the future direction of the action sheet, including the follow-up improvements you've suggested. For reference, here's how the action sheet looks on iOS before the changes in this PR:

  

- I also notice that, at least on iOS (I haven't tested on Android yet) I can't tap somewhere above the top of the action sheet to make it disappear (this was possible before); I have to tap the cancel button. Is there an easy fix to let me tap above the top of the action sheet to make it disappear?

- Another thing I notice is that, if I've rotated the device into a landscape orientation, something weird happens: everything seems to turn sideways when I open the action sheet, as though it's trying to go back to a portrait orientation. Then, when I close the action sheet, things turn right-side-up again. Here's what it looks like:

  https://user-images.githubusercontent.com/22248748/108567264-e9bcd800-72d5-11eb-91b6-d1e203df154b.mp4

  Do you see this happening on Android too? That could provide a clue for debugging.

abhi0504:
Thank you so much @chrisbobbe for your reply 
I am fixing CI failures and need some help in fixing them for now some of them are fixed and some of them are in process

Here are the screenshots of the new UI on Android, in both night and day modes
Android Emulator - Pixel_3_API_29_5554 20-02-2021 14_50_52    Android Emulator - Pixel_3_API_29_5554 20-02-2021 14_51_38

Also you have mentioned that there is a different UI in `ios` which might be OK for now also we can change that in future commits using `Platform`  

you have mentioned that on clicking outside the `messageActionSheet` the `modal` must be closed so for that I have added an commit [here](https://github.com/zulip/zulip-mobile/pull/4490/commits/5abbc99ada6f3aa83192ce1453a90cc86f1fd0a1) If you would say then i will squash both commit into one
Here is a video which can reffer above statement

https://user-images.githubusercontent.com/56453541/108592186-aaa08c80-7392-11eb-95a7-f441dc07b728.mp4

Another thing that you have noticed is that in landscape orientation, something weird happens: everything seems to turn sideways when I open the action sheet, as though it's trying to go back to a portrait orientation. Then, when I close the action sheet, things turn right-side-up again but this seems perfectly fine for android device here is an video for that

https://user-images.githubusercontent.com/56453541/108592247-f6ebcc80-7392-11eb-9b1c-03936d1b7ee4.mp4

@chrisbobbe I hope I have cleared all the points can you plz help in solving CI failures on CZO [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Black.20screen.20appears.20when.20user.20click.20on.20.60Share.60.20option.2E)

chrisbobbe:
Thanks, @abhi0504!

> Another thing that you have noticed is that in landscape orientation, something weird happens: everything seems to turn sideways when I open the action sheet, as though it's trying to go back to a portrait orientation. Then, when I close the action sheet, things turn right-side-up again but this seems perfectly fine for android device here is an video for that

Ah, good to know this problem doesn't happen on Android. We still need to either debug the issue on iOS, or just decide not to use this new UI on iOS 😉. I currently prefer the latter (keeping the current UI on iOS), for a few reasons:

- This bug
- I'm not sure this new UI is an improvement on iOS, as I mentioned in https://github.com/zulip/zulip-mobile/pull/4490#issuecomment-782411394.
- #4407, the black-screen issue, doesn't seem to happen on iOS (I've just tested)
- You might not happen to have a Mac computer, to run Xcode, which is required to test on iOS

So, if you're still interested in the new UI on Android, it'll be necessary to write some conditional logic (using [`Platform`](https://reactnative.dev/docs/platform-specific-code#platform-module)) so that we can keep the current UI on iOS. Unfortunately, this area of the code is already tricky, but it's also quite important, so we should try not to make the code harder to understand than it is now.

That'll likely mean several [NFC](https://github.com/zulip/zulip-mobile/blob/master/docs/glossary.md#nfc-or-nfc) commits, to refactor the code into a state where adding such a conditional makes sense. In particular, I'm imagining an approach like this:

- Write code to implement the new UI, but don't call it yet
- Refactor the show-action-sheet logic so it uses a simple conditional, e.g., `Platform.OS === 'android' ? … : …`, but both branches of the conditional call the code for the current UI.

Then, after that,

- A small, simple commit that changes the `true` branch of the platform-is-Android conditional, so that it calls the code for the new UI.

That way, we can maintain the invariant that no commits have the new UI active on iOS. Since we know we don't want it right now (for the reasons I mentioned, like because it's buggy), that'll be really helpful for the code review. 🙂

One place the platform-is-Android conditional might go is in `showActionSheet` in src/message/messageActionSheet.js. The iOS branch could call `showActionSheetWithOptions`, and the Android branch would (eventually) call the function that sets the new-UI modal to be visible, named `handleLongPressModal` in this revision. 

I see that `handleLongPressModal` sets a piece of state that lives in `ChatScreen`. Instead, let's have that state live in a new wrapper component that lives in src/webview/MessageList.js—the `ChatScreen` shouldn't have to care how the message-action-sheet operates. (Also, leaving `MessageList`'s callers to manage the modal state means that state-management logic has to be copied around each caller. So, while the modal works on `ChatScreen`, it doesn't work on `SearchMessagesCard` in this revision, since it wasn't copied to there.)

chrisbobbe:
> I see that `handleLongPressModal` sets a piece of state that lives in `ChatScreen`. Instead, let's have that state live in a new wrapper component that lives in src/webview/MessageList.js

You might find it helpful to have the `MessageList`'s props get properly type-checked before working on that, though. I've got #4465 open for that.

chrisbobbe:
Also, it's important to note—the approach I've outlined [above](https://github.com/zulip/zulip-mobile/pull/4490#issuecomment-786912159) has more complexity than we'd normally expect a new contributor to handle (as mentioned, this area of the code is tricky, but it also deserves lots of attention to correctness and readability). You may find it more pleasant and educational to choose something else to work on; up to you. 🙂

abhi0504:
Thanks @chrisbobbe for your suggestion and your opinion love to know that you appreciate my efforts 😇 and I also think that this area of the code is already tricky, but it's also quite important, so we should try not to make the code harder to understand than it is now and we could be friendlier to new contributors by making this code easier to work with so I think I should leave this PR for now for future refactoring but once again thank you so much @chrisbobbe and @gnprice for your constant help and support 😊

gnprice:
#4407 has now been fixed in Expo upstream (thanks @abhi0504 for making that upstream bug report!).

It's possible we'll eventually want to switch anyway to something like `Modal` where we can have more control, but from this thread it looks like there will be some tricky things to get right in order to do so. We'll make a new PR if we end up tackling that at some point.

Issue #4416, opened 2021-01-15
Track in-progress login state as local UI state, not global Redux app state


gnprice:
This issue describes a plan for solving #2380 and #4415.

The root cause of both of those is that when you start to log into a Zulip account (whether at first startup, or from the "Switch account" screen), we immediately put our partial information about it into the global Redux state, in `state.accounts`, and then we use that global state to track your progress through the login flow.

Fundamentally this information (like "the user is in the middle of logging into an account at `https://chat.example/`; they haven't yet supplied an email or completed authentication") is a poor fit for the global state, because the rest of the app needs different things from the data about the user's accounts from what the login flow needs.
* The complications we have in the accounts data to make it work for the login flow have been a longtime source of bugs where those complications trip up the rest of the app; the most recent was #4388.
* And at the same time, the logic of the login flow is limited by what fits in the accounts data we use for the rest of the app, which leads to code like what's quoted in https://github.com/zulip/zulip-mobile/issues/4415#issuecomment-760546036 which is the code that causes #4415 and parts of #2380.

Instead:
* First, `DevAuthScreen` and `PasswordAuthScreen` should not use `getPartialAuth` to get data to pass to their API requests. They should take `realm: URL` as route parameters, and `AuthScreen` should pass that information when navigating to them.
* Then, `PasswordAuthScreen` should not use `getPartialAuth` to get an email to pre-fill-in either. Better for it to not pre-fill any email at all -- that would fix the symptoms in #4415.
  * At this point, `getPartialAuth` can go away entirely.
* `AuthScreen` should not get the "current realm" from Redux. ~~It should take `realm: URL` as a route parameter, alongside `serverSettings`.~~ Better: there's already `realm_uri` in that `serverSettings` data, and that's the realm URL that the server identifies for itself. We should simply use that.
* `RealmScreen` should not dispatch `realmAdd(…)` to add any data to Redux.
  * This means `realmAdd` can disappear, along with the rest of the code related to `REALM_ADD`.
  * At this point I think the symptoms in #2380 will be fixed too.

Some other changes that would be good as followups:
* In `AccountPickScreen` we should not be navigating to `RealmScreen`, only to have that screen flash by -- that's not good UX. The code that does that is quoted in https://github.com/zulip/zulip-mobile/issues/4415#issuecomment-760546036 . Instead we should navigate directly to `AuthScreen` if the account is not logged in.
[...]

Issue #4415, opened 2021-01-14
"Switch account" can go to wrong account on chosen realm, if not logged in


gnprice:
I've been seeing this recently on v27.158, the current beta version. But now that I investigate, it reproduces on v27.157 too, the current general release -- it just requires a condition I don't often have. (I haven't tested older versions; it may have been present for a long time.)

This is a high priority because showing one account when the user wanted another could be very concerning for the user.

I'd been seeing versions of this for a few days, but it wasn't immediately clear what the pattern was. Here's a repro that seems to work reliably for me:
* Have three accounts in the list: call them A and B, both on one realm, and C on another.
* Be logged into accounts C and A, but not B.
* Be in the main app UI for account C, the one on the other realm.
* Profile > Switch account > choose account B:
  ![image](https://user-images.githubusercontent.com/28173/104659091-e603b900-5678-11eb-8ce3-d7de6d62c585.png)
* The RealmScreen (with the URL input) flashes by, and then AuthScreen (with buttons for different types of auth) shows up. That's all as expected for choosing a logged-out account from this screen.
* Choose "Sign in with password".
* Expected: Get a login form with the chosen account's email (B) filled in.
* Actual: Get a login form with the *other* account's email (A) filled in -- the one I didn't pick:
  ![image](https://user-images.githubusercontent.com/28173/104659239-2105ec80-5679-11eb-8dc9-0ec9a950c5f9.png)

On the other hand this doesn't reproduce if I'm already logged into account B, the one I choose. Instead I get the loading screen and then the main app UI for account B, as expected. I usually don't have logged-out accounts lying around, which explains why I never saw this before on v27.157.

gnprice:
A variation of that repro:
* Again, have accounts A and B on one realm and C on another.
* Again, be logged into C and A but not B.
* This time, start from the main app UI for account A.
* Again, try to switch to B.
* Again, wind up with a login form with account A's email filled in -- the one we didn't pick.

So whether A or C is the account we start from when we go to pick B, either way it's A we land on. That suggests that "same realm" is indeed a key element of this bug.

gnprice:
Different variation:
* Again, A and B on one realm and C on another.
* This time, be logged into only C, and not A nor B.
* Again, start in the main app UI for C -- which is now the only one we're logged into anyway.
* Again, try to switch to B.
* Again, wind up with a login form with account A's email filled in.

[...]

PR #4491, opened 2021-02-19
Track in-progress login state as local UI state, not global Redux app state

### LINKED ISSUE #2380: Improve login flow.
Here are few things which are currently going on.
* Currently new account is added as soon as user enters `realm` url.
* Currently active account is decided based the index of the account list.
 * If account is at top of the list, then it is active.

Problems associated with it:
If user is in account of realm lets say A and start process of adding another account (on realm B).
Clicks `Switch Account` -> Clicks `Add new account` -> Enters another realm url -> (Optional) Select auth type -> (Optional) enters wrong password. -> Navigate back -> back -> back

* `Optional` here means, this steps can be skipped 

Logically the new account has not setup up, but as per our current implementation, this new account is at the top of the list and thus active.

On navigating back stream, users, messages etc of account of realm A are present, 
* but if user tries to find his/her active account in Account pick screen (after clicking `Switch Account`), it is showing this newly added account as active in realm B (not completely logged in)
* if user tries to fetch new messages, send message they will not succeed as now auth of this half logged in account is active.
* if user closes app and re-open, they he is transferred to account pick screen, but I was expecting it to be at home screen of account with realm A, as user has not completely logged in new account.

Solution:
Add account only after he successfully logs in. Send realm url to auth screen & further via screen props instead of via Global state.
As https://github.com/zulip/zulip/commit/383c62fb03267ce09dccb5e0780ebe27075b62fb got merged, so (in dev login) realm url to which user is registered is also included in the user object. (which may not be the same as entered in the realm screen). Basically then app will also support multiple realm on single server.

@zulipbot claim

chrisbobbe:
This PR follows the plan of #4416, to fix #2380 and #4415.

It doesn't yet address the followups listed there:

> Some other changes that would be good as followups:
> 
> * In `AccountPickScreen` we should not be navigating to `RealmScreen`, only to have that screen flash by -- that's not good UX. The code that does that is quoted in [#4415 (comment)](https://github.com/zulip/zulip-mobile/issues/4415#issuecomment-760546036) . Instead we should navigate directly to `AuthScreen` if the account is not logged in.
>   
>   * This may require some further refactoring to pull relevant logic out of `RealmScreen`. It will require careful testing that things continue to work in all kinds of scenarios.
> * In `AccountPickScreen` we should display an account differently if it's logged out -- it should look "grayed-out" or disabled, or be marked with a label like "Signed out".
>   
>   * Google does this with a "Signed out" label; see screenshots like [here](https://vaadin.com/learn/tutorials/google-login) or [here](https://www.carlsetzer.com/2020/03/03/google-tip-deleting-old-google-sign-in-accounts/).

So, if this gets merged (closing #4416), we'll probably want to open a new issue with those so we don't forget about them.

Fixes: #4416
Fixes: #2380
Fixes: #4415


src/account/__tests__/accountsReducer-test.js
```diff
@@ -153,67 +72,74 @@ describe('accountsReducer', () => {
...
+    });
 
     const prevState = deepFreeze([account1, account2]);
 
-    test('on login, update initial account with auth information, without clobbering zulipVersion', () => {
-      const newAccount = eg.makeAccount({
-        realm: account1.realm,
-        zulipVersion: null,
-      });
+    test("on login, update initial account with apiKey, set ackedPushToken to null, don't clobber anything else", () => {
```
gnprice, line 86: I think this test can be dropped now as a duplicate of "on login, if account does exist, …". By "initial account" I think the description means the thing this PR gets rid of -- the storage of partial login information here in Redux.

src/account/accountsReducer.js
```diff
@@ -62,8 +34,7 @@ const accountSwitch = (state, action) => {
 const findAccount = (state: AccountsState, identity: Identity): number => {
   const { realm, email } = identity;
   return state.findIndex(
-    account =>
-      account.realm.toString() === realm.toString() && (!account.email || account.email === email),
+    account => account.realm.toString() === realm.toString() && account.email === email,
```
gnprice, line 37: Along with this cleanup, we can delete the `, email` here below:
```js
  return [
    { ...state[accountIndex], email, apiKey, ackedPushToken: null },
```

That's now a no-op because at that point we've already ensured that the account we find at `accountIndex` is one that has a matching `email`. It used to be needed for the case of filling in a partial auth.

gnprice (commented):
Thanks @chrisbobbe ! Small comments below.

chrisbobbe:
Thanks for the review, @gnprice! Revision pushed. 🙂

gnprice:
Thanks for the revision! Looks good -- merged.

chrisbobbe:
Thanks for the review! 🎉

PR #4493, opened 2021-02-22
StoreProvider: Set server version tags on global scope, not one-off scope.

chrisbobbe:
We discovered [1] that we'd only been including the server version
tags when making a `logging.foo` call. This excludes the large
volume of Sentry events that come from uncaught exceptions.

So, set the tags on the global scope, and be sure to keep them
up-to-date by listening for changes to the server version stored in
Redux.

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4458.3A.20.22t.2Eget.20is.20not.20a.20function.22.20on.20state.2Enarrows.20at.20sta.2E.2E.2E/near/1119488
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Custom.20tags.20for.20Sentry's.20exception.20tracking/near/1121623

### DIFF
--- a/src/boot/StoreProvider.js
+++ b/src/boot/StoreProvider.js
@@ -1,8 +1,11 @@
 /* @flow strict-local */
 import React, { PureComponent } from 'react';
 import { Provider } from 'react-redux';
-
 import type { Node as React$Node } from 'react';
+
+import { observeStore } from '../redux';
+import * as logging from '../utils/logging';
+import { tryGetActiveAccount } from '../selectors';
 import store, { restore } from './store';
 import timing from '../utils/timing';
 
@@ -11,11 +14,28 @@ type Props = $ReadOnly<{|
 |}>;
 
 export default class StoreProvider extends PureComponent<Props> {
+  unsubscribeStoreObserver: () => void;
+
   componentDidMount() {
     timing.start('Store hydration');
     restore(() => {
       timing.end('Store hydration');
     });
+
+    this.unsubscribeStoreObserver = observeStore(
+      store,
+      // onChange will fire when this value changes
+      state => tryGetActiveAccount(state)?.zulipVersion,
+      zulipVersion => {
+        logging.setTagsFromServerVersion(zulipVersion);
+      },
+    );
+  }
+
+  componentWillUnmount() {
+    if (this.unsubscribeStoreObserver) {
+      this.unsubscribeStoreObserver();
+    }
   }
 
   render() {
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -22,16 +22,29 @@ function withScope<R>(callback: Scope => R): R {
   return ((ret: $FlowFixMe): R);
 }
 
+type ServerVersionTags = {|
+  rawServerVersion: string | void,
+  coarseServerVersion: string | void,
+  fineServerVersion: string | void,
+|};
+
 /**
  * Get server-version tags at various levels of granularity.
+ *
+ * If the server version isn't found in the logging context, all the
+ * tags will be `undefined`.
  */
-const getServerVersionTags = () => {
+const getServerVersionTags = (): ServerVersionTags => {
   const zulipVersion = getLoggingContext()?.serverVersion;
 
   // Why might we not have the server version? If there's no active
   // account.
   if (!zulipVersion) {
-    return {};
+    return {
+      rawServerVersion: undefined,
+      coarseServerVersion: undefined,
+      fineServerVersion: undefined,
+    };
   }
 
   const raw = zulipVersion.raw();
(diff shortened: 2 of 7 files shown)


src/boot/StoreProvider.js
```diff
@@ -11,11 +15,38 @@ type Props = $ReadOnly<{|
...
+          // Set server version tags on Sentry's global scope, so
+          // all events will have them.
+          //
+          // If `zulipVersion` is falsy, all values in the object
+          // passed to `setTags` will be undefined. This means the
+          // tags will be removed from the scope, though this is
+          // unofficial as a way to remove tags:
+          //   https://github.com/getsentry/sentry-javascript/pull/3108#issue-534072956
+          scope.setTags(getServerVersionTags(zulipVersion));
+        });
```
gnprice, line 41: This part is all pretty specific to the nuances of the Sentry API. So let's have it live in the code whose job naturally includes knowing about Sentry, like `src/utils/logging.js`.

As a bonus, then it can stay next to `getServerVersionTags`. The call that this StoreProvider code would make could then look like
```js
  logging.setTagsFromServerVersion(zulipVersion);
```

src/utils/logging.js
```diff
@@ -88,8 +100,7 @@ const logToSentry = (event: string | Error, level: SeverityType, extras: Extras)
   }
 
   const tags = {
-    ...getServerVersionTags(),
-    // Other tags go here; they're useful for event aggregation. See
+    // Tags can go here; they're useful for event aggregation. See
     // https://docs.sentry.io/platforms/javascript/enriching-events/tags/.
   };
```
gnprice, line 105: This should probably have a comment pointing to where we actually do set the tags.

Or better yet: once that `setTags` call moves back into this file (per other review-comment above), this and the `setTags` call below can be deleted. It's possible that in the future we'll have some tags that apply only in the explicit-logging case, but I think the ideal is that we won't and instead all tags will go onto scopes so that they apply equally to uncaught exceptions.

src/utils/logging.js
```diff
@@ -22,16 +22,28 @@ function withScope<R>(callback: Scope => R): R {
...
+ * `undefined`.
  */
-const getServerVersionTags = () => {
-  const zulipVersion = getLoggingContext()?.serverVersion;
-
+export const getServerVersionTags = (zulipVersion: ?ZulipVersion): ServerVersionTags => {
   // Why might we not have the server version? If there's no active
-  // account.
+  // account. N.B: an account may be the active account, but not
+  // logged in.
```
gnprice, line 40: This can perhaps point at the glossary entry:
https://github.com/zulip/zulip-mobile/blob/5f454954f/docs/glossary.md#active-account

gnprice (commented):
Thanks @chrisbobbe ! Comments below.

chrisbobbe:

chrisbobbe:
Oh, hmm: What's the fate of src/boot/loggingContext.js, from 0b9d381e0, after these changes?

chrisbobbe:
Huh, what could that test failure be? Here's the output (it's Jest):

```
Summary of all failing tests
FAIL src/unread/__tests__/unreadModel-test.js
  ● stream substate › EVENT_NEW_MESSAGE › if stream with topic does not exist, append to state

    expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 3

      Immutable.Map {
        2: Immutable.Map {
    +     "some topic": Array [
    +       1,
    +     ],
          "another topic": Array [
            4,
          ],
        },
      }

      139 |       const state = reducer(baseState, action(message), eg.plusReduxState);
      140 |       // prettier-ignore
    > 141 |       expect(summary(state)).toEqual(Immutable.Map([
          |                              ^
      142 |         [eg.stream.stream_id, Immutable.Map([['some topic', [1]]])],
      143 |         [message.stream_id, Immutable.Map([['another topic', [4]]])],
      144 |       ]));

      at Object.<anonymous> (src/unread/__tests__/unreadModel-test.js:141:30)
```

gnprice:
Thanks, looks good -- merged!

> Oh, hmm: What's the fate of src/boot/loggingContext.js, from [0b9d381](https://github.com/zulip/zulip-mobile/commit/0b9d381e0ba1f11fb58320feb77c93031ce00486), after these changes?

Yeah, I think that can now disappear. The new observer basically does its job.


> Huh, what could that test failure be? Here's the output (it's Jest):

Indeed, sounds like a flake! I guess I should debug.

chrisbobbe:
> I guess I should debug.

If you don't see something right away (I didn't), we could just wait for it to happen again, and debug further then?

gnprice:
No, I think I see the issue -- it's that the test assumes that `eg.stream.stream_id` and `message.stream_id` will be different. But the latter is hardcoded to 2, and the former is randomized in `exampleData`.

gnprice:
Pushed a fix: c7d373540.

PR #4494, opened 2021-02-23
Hooks: convert `AccountItem`, `AccountList`.

chrisbobbe:
I had a spare minute or two, so I converted these more or less at random (they're at the top of a search results for `extends PureComponent`) and thought I'd send it as a PR. 🙂


gnprice:
Looks good -- merged!

Issue #4492, opened 2021-02-20
Excessive space around numbered list


gnprice:
@kasperg reported the following behavior at zulip/zulip#17284:
![image](https://user-images.githubusercontent.com/28173/108585274-66fb4380-72fc-11eb-855e-d356c6587ded.png)

There are two things that are quite wrong here:
1. There's too much space between the paragraph and the following list.
2. The list items are too deeply indented.

By contrast, we have pretty reasonable behavior with a plain bulleted list. We should use the same spacing for numbered lists as for bulleted lists.

(It's possible the spacing for bulleted lists should be slightly tweaked, too. But first let's make both kinds behave the same way, and then we can tweak them both together.)

A good comparison to look at is the webapp's behavior. See https://github.com/zulip/zulip/commit/03e4ccbc2d6f092e84cda20b06c20cef3e94e565 which recently adjusted this for the webapp.

rajprakash00:
@gnprice Can i work on this issue?

PR #4495, opened 2021-02-23
Message-view: Fix excessive spacing for lists followed by paragraphs.

rajprakash00:
Previously there was an excessive space for ordered lists coming after
a paragraph unlike plain bulleted lists.This was already fixed in webapp
message-list view in [#17285](https://github.com/zulip/zulip/pull/17285)

We fixed this in mobile app by following same spacing as of bulleted
lists coming after paragraph.

**Screenshots:**
zulipAfter3

Fixes: #4492


rajprakash00:
@gnprice Please take a look

gnprice:
Thanks @rajprakash00 !

Would you please post both a "before" screenshot as well as an "after" screenshot of the same test data? That would make this a lot easier to review.

I like the idea of treating `ul` and `ol` the same. When I look at the webapp's `rendered_markdown.css`, that's what we seem to do there.

I notice in this code that there's another existing rule that also refers to `ul`, and it doesn't get updated in this commit:
```diff
 p {
   margin: 0;
 }
 .message p + p {
   margin-top: 16px;
 }
-.message ul {
+.message ul, ol {
   padding-left: 20px;
   margin: 4px 0 0 0;
 }
 .message ul + p {
   margin-top: 16px;
 }
```

Is there a reason to leave that one unchanged? Other than making `ul` and `ol` still not quite treated the same, I wonder what the effect is for `ol` of having just one of these rules apply and not the other.

rajprakash00:
Thanks @gnprice for the review.

I have added the before screenshot .

Hmm, I tested this just now and `ul` and `ol` styles becomes different when a paragraph comes after that, and you are right about treating `ol` and `ul` same , it should be same as of `ul + p` style.

Issue #4496, opened 2021-02-23
Logout Confirmation


Kunal-2001:
While logging out the account, the app should ask confirmation from the user 
There could be a dialog box displayed asking for confirmation and then fire the logout behaviour
@gnprice Can you please review this issue ?

abhi0504:
Interesting I also just checked it out and I also thinks that there must be an Alert box or some confirmation message for log out @chrisbobbe Whats your suggestion about this

PR #4497, opened 2021-02-24
Adding Alert for logout

abhi0504:
On the profile pic screen when we click on `logout` button then we suddenly logged out without any confirmation message or without any `Alert` which is not that good their must be an `Alert` message before someone logged out like in other apps.

So this PR added an `Alert` message before logout so that if user mistakenly clicks on logout button then it will not suddenly logged out. Attaching a video below which shows that logout `Alert` box

https://user-images.githubusercontent.com/56453541/108977036-27db4280-76ae-11eb-84e7-7e43e03910b0.mp4

Also we can add a spinner or loader before logging out to prevent that rapid log out effect @chrisbobbe @gnprice whats your opinion about that

Fixes: #4496

### DIFF
--- a/src/account-info/ProfileScreen.js
+++ b/src/account-info/ProfileScreen.js
@@ -1,7 +1,8 @@
 /* @flow strict-local */
-import React from 'react';
-import { ScrollView, View } from 'react-native';
+import React, { useContext } from 'react';
+import { ScrollView, View, Alert } from 'react-native';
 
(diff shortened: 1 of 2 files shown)


abhi0504:
Also I am thinking of adding a loader untill the user logs out @gnprice @chrisbobbe what's your suggestions over this and also please review this PR

chrisbobbe:
Thanks, @abhi0504! Merged, with a few tweaks; please take a look. 🙂

PR #4498, opened 2021-02-25
loggingContext: Remove, as now unnecessary.

chrisbobbe:
In 195d5fd9e, we found a better way to include the server version in
Sentry events than the one we landed on in 0b9d381e0, and it still
doesn't cause problems with import cycles.

So, take down the previous setup, as discussed at
  https://github.com/zulip/zulip-mobile/pull/4493#issuecomment-784673377.


gnprice:
Looks good, thanks -- merged!

PR #4499, opened 2021-02-25
emoji picker: Allow searching for literal emoji.

### LINKED ISSUE #3961: In "add reaction" search, accept actual emoji as well as names

WesleyAC:
This allows the user to use the system emoji picker to select reaction
emoji, should they prefer that to typing the emoji name.

There is some complexity around skin-tone modifiers that this commit
does not attempt to address, see the comment for details.

Fixes #3961.

emoji_literal_search
emoji_string_search

### DIFF
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -74,13 +74,6 @@ describe('getFilteredEmojis', () => {
     ]);
   });
 
-  // This is sort of odd behaviour in some sense: the UI is showing the user
-  // multiple emojis, when they can only react with one of them (if they
-  // react twice with :+1: and :thumbs_up:, it'll only be stored as a single
-  // reaction). We could canonicalize the emoji name in cases where there
-  // are multiple aliases, but it seems somewhat useful to allow the user to
-  // discover that the aliases exist, and unlikely to confuse in practice
-  // (since people will likely just pick an arbitrary one and be done with it)
   test('returns multiple literal emoji', () => {
     const list = getFilteredEmojis('👍', {});
     expect(list).toEqual([
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -67,6 +67,28 @@ describe('getFilteredEmojis', () => {
     ]);
   });
 
+  test('returns literal emoji', () => {
+    const list = getFilteredEmojis('🖤', {});
+    expect(list).toEqual([
+      { emoji_type: 'unicode', code: '1f5a4',  name: 'black_heart' },
+    ]);
+  });
+
+  // This is sort of odd behaviour in some sense: the UI is showing the user
+  // multiple emojis, when they can only react with one of them (if they
+  // react twice with :+1: and :thumbs_up:, it'll only be stored as a single
+  // reaction). We could canonicalize the emoji name in cases where there
+  // are multiple aliases, but it seems somewhat useful to allow the user to
+  // discover that the aliases exist, and unlikely to confuse in practice
+  // (since people will likely just pick an arbitrary one and be done with it)
+  test('returns multiple literal emoji', () => {
+    const list = getFilteredEmojis('👍', {});
+    expect(list).toEqual([
+      { emoji_type: 'unicode', code: '1f44d',  name: '+1' },
+      { emoji_type: 'unicode', code: '1f44d',  name: 'thumbs_up' },
+    ]);
+  });
+
   test('search in realm emojis as well', () => {
     expect(
       getFilteredEmojis('qwerty', {
--- a/src/emoji/data.js
+++ b/src/emoji/data.js
@@ -4,6 +4,7 @@ import 'string.fromcodepoint';
 import type { ImageEmojiType, EmojiType } from '../types';
 import { objectFromEntries } from '../jsBackport';
 import { unicodeCodeByName, override } from './codePointMap';
+import objectEntries from '../utils/objectEntries';
 
 const unicodeEmojiNames = Object.keys(unicodeCodeByName);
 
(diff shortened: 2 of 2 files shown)


src/emoji/data.js
```diff
@@ -28,12 +28,34 @@ export const getFilteredEmojis = (
...
+    // the user the default skin color emoji when they've typed in a
+    // modified one than to display nothing. We should circle back to
+    // this later, and figure out what our overall strategy for skin-tone
+    // modifiers across the app should be.
+    const matches_emoji_literal = parseUnicodeEmojiCode(x[1]) === query;
+    const matches_emoji_name = Math.min(1, x[0].indexOf(query));
+    return [x[0], matches_emoji_literal ? 0 : matches_emoji_name];
+  });
+
+  const matchingCustomEmoji = Object.keys(activeImageEmojiByName).map(x => [
```
gnprice, line 50: Let's call this "image" emoji rather than "custom". That matches the name of the data it's processing, `activeImageEmojiByName`... and it also reflects the fact that it includes not only the custom emoji but also the Zulip extra emoji. (Of which there's just one: `:zulip:`.)

src/emoji/data.js
```diff
@@ -28,12 +28,34 @@ export const getFilteredEmojis = (
...
+    Math.min(1, x.indexOf(query)),
+  ]);
+
   // Map from emoji names to 0 for prefix matches, 1 for others.
   // (Nonmatching names are excluded.)
   const nameMap: Map<string, number> = new Map(
-    [...unicodeEmojiNames, ...Object.keys(activeImageEmojiByName)]
-      .map(x => [x, Math.min(1, x.indexOf(query))])
-      .filter(([_, i]) => i !== -1),
+    [...matchingUnicodeEmoji, ...matchingCustomEmoji].filter(([_, i]) => i !== -1),
```
gnprice, line 58: The names with "matching" are a bit misleading: these lists actually contain all the emoji, just with markers indicating whether they match (and with what priority if so.)

One way to fix that would be to propagate the `filter` to the two definitions. That should also be slightly more efficient, as we then don't allocate a long array here if there are only a few matches.

src/emoji/data.js
```diff
@@ -28,12 +28,34 @@ export const getFilteredEmojis = (
...
+    }
+    // We may want to have fancier emoji-matching logic in the future: this
+    // code will not work with skin-tone modifiers, for instance, and
+    // there may be other strange unicode things that warrant special
+    // handling. It would be simple to add code to strip off the
+    // skin-tone modifiers in the query, but it seems worse to suggest to
+    // the user the default skin color emoji when they've typed in a
+    // modified one than to display nothing. We should circle back to
+    // this later, and figure out what our overall strategy for skin-tone
+    // modifiers across the app should be.
```
gnprice, line 44: Yeah. We have an issue for supporting skin-tone modifiers, which this should link to: zulip/zulip#992.

I agree it's probably unhelpful to provide a partial match, like the version with the default/nonhuman skin tone when the user enters an emoji with a specific skin tone.

I think this *code* will actually work perfectly fine with skin-tone modifiers, once Zulip supports them in emoji reactions at all. When that happens, the "code" for the emoji will be something that `parseUnicodeEmojiCode` can turn into the actual sequence of codepoints: e.g. '1f44d-1f3fb' would be the code for 👍🏻. For the format of these codes, see the comment on `Reaction.emoji_code` here:
https://github.com/zulip/zulip/blob/15c4a72ac5297b1213ba93e986147c4c27a75af2/zerver/models.py#L2206-L2219
(That's linked to by the jsdoc on our `unicodeCodeByName`. There's probably other places we could usefully add links.)

Maybe what that boils down to is:
* This emoji-literal logic will only find exact matches. That's a product/UI choice we're making that deserves a comment calling it out explicitly.
* This means if the user enters an emoji that's in Unicode but not in Zulip, we won't show a match. One particular case of that is emoji with explicit skin tone (zulip/zulip#992); it would be simple to add logic to return the default-skin-tone version in that case, but it seems worse to do that than to display nothing.

src/emoji/__tests__/data-test.js
```diff
@@ -67,6 +67,13 @@ describe('getFilteredEmojis', () => {
     ]);
   });
 
+  test('returns literal emoji', () => {
+    const list = getFilteredEmojis('🖤', {});
+    expect(list).toEqual([
+      { emoji_type: 'unicode', code: '1f5a4',  name: 'black_heart' },
+    ]);
+  });
```
gnprice, line 75: How does this behave in a case where we have several names for a given emoji? Per the jsdoc on `unicodeCodeByName`:
```
 * Many codes have multiple names.  For example: as of late 2018, there were
 * approximately 1600 names for 1000 codes, with about 400 codes having >=2
 * names, about 100 having >=3, up to as many as 6 names for the two
 * record-holders ('2620' and '1f5dd').
```

Probably a screenshot of the UI in that case would be helpful, as well as a test case demonstrating it.

gnprice (commented):
Thanks @WesleyAC! Some comments below.

src/emoji/data.js
WesleyAC: Yeah, I think the thought around this is mostly if #992 ends up resulting in having a setting for "default skin tone modifier", we might want to always include that in the list in addition to the literal match, even if they search for a different one? Or maybe not, I'm not super sure. But it seems like something to think about when we put in the work to decide how things should look for #992. In any case, I reworded the comment a bit to mention that.

src/emoji/__tests__/data-test.js
WesleyAC: It shows the same emoji multiple times. I think this is somewhat reasonable, I added a test for it and a comment explaining my reasoning.

WesleyAC:
Should be ready for a second pass @gnprice :)

src/emoji/__tests__/data-test.js
chrisbobbe: > approximately 1600 names for 1000 codes, with about 400 codes having >=2

Not that it really matters, but 400 looks too low to me. Doesn't it have to be at least 600, and is that by the pigeonhole principle? 🤔

alexmv: The 1st code could have 601 names, and all of the rest could have one.  Pigeonhole principle here just says "1600 names and 1000 codes means _at least one code_ has more than one name."

chrisbobbe: Ah, right, interesting—thanks, @alexmv!

src/emoji/__tests__/data-test.js
```diff
@@ -67,6 +67,28 @@ describe('getFilteredEmojis', () => {
...
+
+  // This is sort of odd behaviour in some sense: the UI is showing the user
+  // multiple emojis, when they can only react with one of them (if they react
+  // twice with :+1: and :thumbs_up:, it'll only be stored as a single
+  // reaction). We could canonicalize the emoji name in cases where there are
+  // multiple aliases, but it seems somewhat useful to allow the user to
+  // discover that the aliases exist, and unlikely to confuse in practice
+  // (since people will likely just pick an arbitrary one and be done with it)
+  test('returns multiple literal emoji', () => {
```
gnprice: Yeah. There's another reason this is the right behavior, too -- it lets the user pick which name they want to use. For example if you enter 🗽 aka U+1F5FD STATUE OF LIBERTY, on a version that has the changes in this PR, you get this:
image
with `:new_york:`, `:statue:`, and `:statue_of_liberty:` as your choices. They all result in reactions with the same `emoji_code` and hence the same glyph 🗽, but the `emoji_name` will be different, and the latter controls what other users will see as the name when (on web/desktop) they hover over the reaction.

(In this particular example, probably the actual right answer is that we should not have added `:statue:` as an alias for STATUE OF LIBERTY, and maybe not `:new_york:` either. But as long as we have aliases where the names have any difference in meaning, we'll want users to be able to choose.)

This is unfortunately not really documented in our emoji subsystem doc:
https://zulip.readthedocs.io/en/latest/subsystems/emoji.html
but you can get some of it from the Reaction class (mostly its AbstractReaction base class) and the comments on it, in the server's models file:
https://github.com/zulip/zulip/blob/15c4a72ac5297b1213ba93e986147c4c27a75af2/zerver/models.py#L2179

gnprice (commented):
Thanks again @WesleyAC ! All looks good -- merged.

I added a commit on top that makes some comment/jsdoc changes, following up on the remarks below.

PR #4501, opened 2021-02-26
actionTypes: Remove `email` from `EventSubscriptionUpdateAction`.

chrisbobbe:
We don't use this, and Tim reports [1] that the server would like to
get rid of it.

[1] https://chat.zulip.org/#narrow/stream/3-backend/topic/stream.20administrators/near/1124276


gnprice:
Thanks for the cleanup! Merged, with a small tweak: I left the property there but commented out, to help the type serve as notes for understanding what we get from the API. (Unlikely to really matter in this case, but something that's helpful for us to make a practice of in general.)

PR #4502, opened 2021-02-28
deps: Use `@react-native-clipboard/clipboard`

rajprakash00:
`Clipboard` in React Native core is deprecated. See [here](https://reactnative.dev/docs/clipboard).
 Although no visible bug occurs but warning errors are 
shown in console  to use `@react-native-community/clipboard`
which was recently migrated to `@react-native-clipboard/clipboard` [here](https://github.com/react-native-clipboard/clipboard/pull/87).

So,migrated to it.

**Steps to reproduce the bug**:
- Go to any message-view
- try copying the message using **Copy to Clipboard**
- See console for warning errors

**Additional Context:**
Since I don't have MacOS so haven't linked & tested it in iOS.

### DIFF
--- a/flow-typed/@react-native-clipboard/clipboard_vx.x.x.js
+++ b/flow-typed/@react-native-clipboard/clipboard_vx.x.x.js
@@ -1,0 +1,11 @@
+// node_modules/@react-native-clipboard/clipboard/dist/Clipboard.d.ts
+declare module '@react-native-clipboard/clipboard' {
+  declare export default {
+    /**
+     * Set content of string type.
+     *
+     * @param the content to be stored in the clipboard.
+     */
+    setString(content: string): void,
+  };
+}
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@expo/react-native-action-sheet": "^3.8.0",
+    "@react-native-clipboard/clipboard": "^1.7.0",
     "@react-native-community/async-storage": "^1.6.3",
     "@react-native-community/cameraroll": "^4.0.4",
     "@react-native-community/masked-view": "^0.1.10",
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -1,5 +1,6 @@
 /* @flow strict-local */
-import { Clipboard, Share, Alert } from 'react-native';
+import { Share, Alert } from 'react-native';
+import Clipboard from '@react-native-clipboard/clipboard';
 
 import * as NavigationService from '../nav/NavigationService';
 import type {
--- a/src/webview/handleOutboundEvents.js
+++ b/src/webview/handleOutboundEvents.js
@@ -1,5 +1,6 @@
 /* @flow strict-local */
-import { Clipboard, Alert } from 'react-native';
+import { Alert } from 'react-native';
+import Clipboard from '@react-native-clipboard/clipboard';
 
 import * as NavigationService from '../nav/NavigationService';
 import * as api from '../api';
--- a/yarn.lock
+++ b/yarn.lock
@@ -2110,6 +2110,11 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@react-native-clipboard/clipboard@^1.7.0":
+  version "1.7.0"
+  resolved "https://registry.yarnpkg.com/@react-native-clipboard/clipboard/-/clipboard-1.7.0.tgz#43320841870b82b2f311f375dd5f178da46e244e"
+  integrity sha512-i5dJgR+wM8Om+hFEB/PqNb65/x5WxpaZG+UjEBX2/gmmIrmAWI72tI9rVL1gjPA9RWNpdpzvp+ioGjpdl7MyWQ==
+
 "@react-native-community/async-storage@^1.6.3":
   version "1.6.3"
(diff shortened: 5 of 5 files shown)


chrisbobbe:
Interesting, thanks for finding this! I see that deprecation warning in RN's [`Clipboard` module doc](https://reactnative.dev/docs/clipboard).

I actually haven't seen the errors you describe in the commit message. When and where are you seeing those? It makes me wonder if I'm missing an important source of information in my normal development setup. 🙂

rajprakash00:
Oh! My bad  ,I should have written steps to reproduce the bug. 
Anyways the warning errors appears in console whenever we try the **copy to clipboard** option to copy any message from message-view 
@chrisbobbe  Is it helpful now?

chrisbobbe:
> whenever we try the **copy to clipboard** option to copy any message from message-view

OK, great, that's helpful! Let's include that in the commit message. 🙂

flow-typed/npm/@react-native-clipboard/clipboard_vx.x.x.js
```diff
@@ -0,0 +1,63 @@
+// flow-typed signature: 12bef3bad479f254c9e8da5a51023c58
```
chrisbobbe, line 1: From the commit message:

```
Also,this library has built-in Flow support, so we don't need to make our
own libdef.
```

Is this true? I see that you've added a stub libdef with no types in it. If you delete that, do we get good type-checking (i.e., Flow correctly complains when you try to do something wrong)?

chrisbobbe (commented):
I've just tested on iOS, and it works! See one question below (and a note [above](https://github.com/zulip/zulip-mobile/pull/4502#issuecomment-833637597)); otherwise, this looks good.

flow-typed/npm/@react-native-clipboard/clipboard_vx.x.x.js
rajprakash00: @chrisbobbe Thanks for review!

Hmm... I remember now , the problem occured was that this error `Importing from an untyped module makes it any and is not safe! Did you mean to add // @flow to the top of @react-native-clipboard/clipboard?Flow(untyped-import)`  always occured 
And after some research I found I have to add a stub libdef for this(which Ithink it should be optional here :thinking:)

chrisbobbe: OK; that sounds like a sign that "Also,this library has built-in Flow support" is wrong. 🙂 Is there still something that leads you to believe that the library has built-in Flow support?

As mentioned above, if it turns out that it doesn't come with types, we'll have to write our own with a meaningful libdef.

rajprakash00: OK , now I get it...seems earlier I misunderstood the built-in Flow support thing pretty much :( .
> As mentioned above, if it turns out that it doesn't come with types, we'll have to write our own with a meaningful libdef.

I'm a little confused about this,
AFAIK `@react-native-clipboard/clipboard` has published libdef available(so we can get autogenerated libdef stub for it) , 
Or do we have to write separate libdef mentioning proper types , for e.g, we have for `react-native-webview`

chrisbobbe: Ah, OK, so the thing we usually do to see if there's a libdef published by `flow-typed` is this:

```
flow-typed install @react-native-clipboard/clipboard
```

I've tried this command just now, and it gave this output:

```
[@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead
• Searching for 1 libdefs...
• rebasing flow-typed cache...
!! No flow@v0.128.0-compatible libdefs found in flow-typed for the explicitly requested libdefs. !!

Consider using `flow-typed create-stub @react-native-clipboard/clipboard` to generate an empty libdef that you can fill in.
```

(I have no idea what the `@octokit/rest` thing is about; ignoring that.)

Notice the "No […] libdefs found in flow-typed […]."

The file you made looks like it was created by running the recommended command, `flow-typed create-stub @react-native-clipboard/clipboard`. But, as the rest of that sentence says (and as the generated file says), the file you get from that is basically empty; there's some boilerplate but nothing substantive. It won't have useful types until we write some. 🙂

The libdef for `react-native-webview` is an example where we wrote one, but I expect this one would be much easier than that—we use just one function it provides, `Clipboard.setString`, and that function's type seems pretty simple. Its parameters are described [here](https://github.com/react-native-clipboard/clipboard#setstring), and I think it probably returns `undefined`.

rajprakash00:
@chrisbobbe I've updated the changes, please take a look.

chrisbobbe:
Looks good, thanks! Please squash the second commit into the first; it's neatest to add a libdef in the same commit that we add the library itself. Also (and I'm not sure this is documented anywhere), our custom or at-all edited libdefs should go in `flow-typed/`, not `flow-typed/npm`. Also, when you squash the commits together, please make sure the commit message doesn't say that the library has built-in Flow types, since that's confusing (our custom libdef is where the types live). 🙂 

Thanks!

gnprice:
Updated by @chrisbobbe and merged as #5648. Thanks @rajprakash00 for the report and the fix!

PR #4505, opened 2021-03-03
Baby steps toward detangling `ComposeBox`.

chrisbobbe:
Including a new comment on the `subject` field on our `Message` type, and fixes for some at least latent (though possibly not-super-common?) bugs.

### DIFF
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -543,6 +543,12 @@ export type PmMessage = $ReadOnly<{|
    */
   display_recipient: $ReadOnlyArray<PmRecipientUser>,
 
+  /**
+   * PMs have (as far as we know) always had the empty string at this
+   * field. Though the doc warns that this might change if Zulip adds
+   * support for topics in PMs; see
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Just rebased across #4506.

WesleyAC (approved):

WesleyAC:
Thanks for the detailed commit messages :)

I wouldn't have been able to figure out what was going on here without that, it was quite a rube goldberg machine to set the topic :grimacing:

chrisbobbe:
Thanks for the review! Merging. (@gnprice, hopefully it's OK to skip your review, but this has been open for a bit. Also, I've tried to leave a careful trace of my thought process in the commit messages, in case this causes new bugs that don't camouflage into the component's existing bugs. 🙂)

gnprice:
Yeah, sorry I let this sit for a while. I just read through it now -- that was indeed quite a complicated sequence! Thanks for cleaning up this bit of the logic, and leaving that careful trace of your analysis.

PR #4506, opened 2021-03-03
Make `PmMessage` and `StreamMessage` types.

chrisbobbe:
(Removed some outdated text here as our thinking proceeded.)

There's a site in the last commit where the handling for `subject` isn't great, and things get worse at that same site when I try to do `subject_links`; it seems to panic a bit (complaining about things other than `subject_links`; it's like it's forgotten some things it would otherwise know) unless I remove the code that gives `subject_links` a new value. So I'll pause on this for today and take a fresh look tomorrow (and you could give it a go if you like 🙂).

But what do you think of this general approach, @gnprice? I think it's similar to what you proposed [in CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60PmMessage.60.2C.20.60StreamMessage.60.20types.3F/near/1126689) but it doesn't use `MessageBase`; in fact, that can remain unexported, which I kind of like (since it doesn't represent a whole object we'd actually pass around anywhere).

You might find it helpful to start by looking at these two consecutive commits:

5c008bc18cc142cc946e3339887024a782b1ea60 messagesReducer types: Handle `Message` being `PmMessage` | `StreamMessage`.
68c67426a7b9df8c0546faa27564d8a6e424170e getMessages types: Handle `Message` being `PmMessage` | `StreamMessage`.

### DIFF
--- a/src/api/messages/getMessages.js
+++ b/src/api/messages/getMessages.js
@@ -46,27 +46,22 @@ type ServerApiResponseMessages = {|
 
 /** Exported for tests only. */
 export const migrateMessages = (messages: ServerMessage[], identity: Identity): Message[] =>
-  messages.map(message => {
-    /* eslint-disable-next-line no-unused-vars */
-    const { reactions, avatar_url: rawAvatarUrl, ...restMessage } = message;
-
-    return {
-      ...message,
-      avatar_url: AvatarURL.fromUserOrBotData({
-        rawAvatarUrl: message.avatar_url,
-        email: message.sender_email,
-        userId: message.sender_id,
-        realm: identity.realm,
-      }),
-      reactions: message.reactions.map(reaction => {
-        const { user, ...restReaction } = reaction;
-        return {
-          ...restReaction,
-          user_id: user.id,
-        };
-      }),
-    };
-  });
+  messages.map(message => ({
+    ...message,
+    avatar_url: AvatarURL.fromUserOrBotData({
+      rawAvatarUrl: message.avatar_url,
+      email: message.sender_email,
+      userId: message.sender_id,
+      realm: identity.realm,
+    }),
+    reactions: message.reactions.map(reaction => {
+      const { user, ...restReaction } = reaction;
+      return {
+        ...restReaction,
+        user_id: user.id,
+      };
+    }),
+  }));
 
 const migrateResponse = (response, identity: Identity) => {
   const { messages, ...restResponse } = response;
--- a/src/api/messages/getMessages.js
+++ b/src/api/messages/getMessages.js
@@ -51,7 +51,7 @@ export const migrateMessages = (messages: ServerMessage[], identity: Identity): 
     const { reactions, avatar_url: rawAvatarUrl, ...restMessage } = message;
 
     return {
-      ...restMessage,
+      ...message,
       avatar_url: AvatarURL.fromUserOrBotData({
         rawAvatarUrl: message.avatar_url,
         email: message.sender_email,
--- a/src/api/messages/getMessages.js
+++ b/src/api/messages/getMessages.js
@@ -2,7 +2,7 @@
 import type { Auth, ApiResponseSuccess } from '../transportTypes';
 import type { Identity } from '../../types';
 import type { Message, ApiNarrow } from '../apiTypes';
-import type { Reaction, UserId } from '../modelTypes';
+import type { PmMessage, StreamMessage, Reaction, UserId } from '../modelTypes';
 import { apiGet } from '../apiFetch';
 import { identityOfAuth } from '../../account/accountMisc';
 import { AvatarURL } from '../../utils/avatar';
@@ -31,11 +31,15 @@ export type ServerReaction = $ReadOnly<{|
   |}>,
 |}>;
 
-export type ServerMessage = $ReadOnly<{|
-  ...$Exact<Message>,
+// How `ServerMessage` relates to `Message`, in a way that applies
+// uniformly to `Message`'s subtypes.
+type ServerMessageOf<M: Message> = $ReadOnly<{|
+  ...$Exact<M>,
   avatar_url: string | null,
   reactions: $ReadOnlyArray<ServerReaction>,
 |}>;
+
+export type ServerMessage = ServerMessageOf<PmMessage> | ServerMessageOf<StreamMessage>;
 
 // The actual response from the server.  We convert the data from this to
 // `ApiResponseMessages` before returning it to application code.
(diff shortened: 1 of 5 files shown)


src/api/modelTypes.js
```diff
@@ -486,7 +486,7 @@ export type Submessage = $ReadOnly<{|
  * See also `MessagesState` for discussion of how we fetch and store message
  * data.
  */
-export type Message = $ReadOnly<{|
+type MessageBase = $ReadOnly<{|
```
chrisbobbe, line 489: 6d6ce7720 modelTypes: Start separating `PmMessage` and `StreamMessage`.

(The jsdoc should probably also be moved/refactored, somewhere around this commit.)

src/api/messages/getMessages.js
```diff
@@ -45,27 +49,28 @@ type ServerApiResponseMessages = {|
...
-    };
-  });
+export const migrateMessages = (
+  messages: $ReadOnlyArray<ServerMessage>,
+  identity: Identity,
+): Message[] =>
+  messages.map(<M: Message>(message: ServerMessageOf<M>): M => ({
+    // (This casting looks unnecessary, but seems to nudge Flow in the
+    // right direction.)
+    ...(message: ServerMessageOf<M>),
```
gnprice, line 59: Hmm, does it? If I remove this cast, it seems like everything works fine.

Perhaps it was needed in some previous draft version, but isn't now?

src/message/messagesReducer.js
```diff
@@ -83,9 +83,9 @@ export default (state: MessagesState = initialState, action: Action): MessagesSt
     case EVENT_REACTION_ADD:
       return state.update(
         action.message_id,
-        oldMessage =>
+        <M: Message>(oldMessage: M): M =>
           oldMessage && {
-            ...oldMessage,
+            ...(oldMessage: M),
```
gnprice, line 88: Similarly, this cast seems not to be needed.

src/api/messages/getMessages.js
```diff
@@ -31,11 +31,15 @@ export type ServerReaction = $ReadOnly<{|
   |}>,
 |}>;
 
-export type ServerMessage = $ReadOnly<{|
```
gnprice, line 34: Let's keep the `$ReadOnly` here.

src/unread/unreadStreamsReducer.js
```diff
@@ -16,11 +16,11 @@ import { getOwnUserId } from '../users/userSelectors';
...
 
 const eventNewMessage = (state, action, globalState) => {
-  if (action.message.type !== 'stream') {
+  if (getOwnUserId(globalState) === action.message.sender_id) {
     return state;
   }
 
-  if (getOwnUserId(globalState) === action.message.sender_id) {
+  if (action.message.type !== 'stream') {
     return state;
```
gnprice, line 24: - When we start putting things together to call
      `addItemsToStreamArray`, we need to be sure `action.message` is a
      stream message. We were already sure of that; there's already an
      early return on `action.message.type !== stream`. But Flow seems
      to think it's done too early, and that something in the
      intervening code might thwart the job it's doing. So, move that
      early return a bit later.

Ah, yep. Specifically what's happening is: we check `action.message.type`, but then we call `getOwnUserId` -- and that's some function that might go and do anything. So in particular it might go and mutate the same object that we're looking at as `action.message`. So that object *had* had a `type` of `'private'`, but now it might not and so it might not have a `stream_id`.

src/api/modelTypes.js
```diff
@@ -569,27 +579,15 @@ type MessageBase = $ReadOnly<{|
...
-   * self-user is always included.
+   * This lists the sender as well as all (other) recipients, and it
+   * lists each user just once.  In particular the self-user is always
+   * included.
    *
    * The ordering is less well specified; if it matters, sort first.
-   *
-   * For stream messages, prefer `stream_id`; see #3918.
    */
-  display_recipient: string | $ReadOnlyArray<PmRecipientUser>, // `string` for type stream, else PmRecipientUser[]
```
gnprice, line 580: :confetti_ball:

src/api/modelTypes.js
gnprice: Yeah. Probably it should simply move to the new `Message` definition -- that's what will cause it to appear when looking in the IDE at uses of `Message`.

And then at this spot, and at `PmMessage` and `StreamMessage`, perhaps add just a few words pointing at `Message`.

gnprice:
Thanks @chrisbobbe ! Small comments above, but generally this looks great.

I see also that the last commit is a draft, and there's one more property `subject_links` which presumably you plan to move before completing this. LMK if there are further questions I can answer to help out.

chrisbobbe:
Thanks for the review!

Yeah, I'd actually appreciate it if you could try splitting `subject_links` and see what you might do about the resulting Flow error. I also don't love the solution I landed on for `subject`; it involves a `$FlowFixMe`.

src/api/messages/getMessages.js
chrisbobbe: > Perhaps it was needed in some previous draft version, but isn't now?

Huh, I guess so! Thanks for catching this!

src/message/messagesReducer.js
chrisbobbe: For this one (and likely the others in this file), I find that it's not needed in the commit that introduces them, but it *is* needed as of the commit that splits `type` between `PmMessage` and `StreamMessage`.

chrisbobbe:
I've just pushed a new revision, and responded to your comments above; please see some queries at https://github.com/zulip/zulip-mobile/pull/4506#discussion_r590905300 and https://github.com/zulip/zulip-mobile/pull/4506#discussion_r590890582. 🙂

I haven't yet found the right thing to do for `subject` and `subject_links`; I'm trying something just now and will let you know if it works.

chrisbobbe:
OK, and I have indeed found a working approach at that site for `subject` and `subject_links`! Revision pushed. 😅

chrisbobbe:
Just rebased this past #4518.

chrisbobbe:
Just rebased again and fixed some conflicts; including some from #4448.

src/api/messages/getMessages.js
gnprice: Yep, `ServerMessageOf` (as you've done in the revision.) I.e. where the object-type syntax is. Then `ServerMessage` gets the benefit too, as it just uses `ServerMessageOf`.

gnprice:
Thanks @chrisbobbe ! All looks great -- merging, with a fix to just a nit I spotted in the jsdoc, which stood out with the help of Git's "color moved" feature:

image

image

Probably a small conflict-resolution error on rebasing past #4518.

chrisbobbe:
Great, thanks for the review! 🎉

PR #4507, opened 2021-03-04
api types: Drop `name` from subscription update events.

gnprice:
A quick followup to #4501.

We already ignore this in favor of `stream_id`.  On the server side
we intend to stop sending it soon, because neither mobile nor
terminal is using it:
  https://chat.zulip.org/#narrow/stream/3-backend/topic/stream.20administrators/near/1124277

While we're here, also reorder the remaining properties a bit to be
more logically in general-to-specific order.

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -274,10 +274,11 @@ type EventSubscriptionUpdateAction = {|
   ...ServerEvent,
   type: typeof EVENT_SUBSCRIPTION,
   op: 'update',
-  name: string,
-  property: string,
   stream_id: number,
+  property: string,
   value: boolean | number | string,
+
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Ah, yep! My mistake was seeing that we use a subscription object's `name` field to identify the object (e.g., in `getStreamColorForNarrow` in subscriptionSelectors.js) and assuming we use the `name` field on the event to keep the subscription's name up-to-date. We don't, as I can see looking back at the reducer code that handles the event

Whereas `email` isn't even a field on subscription objects, and somehow I did correctly conclude from the reducer code that we don't use the `email` field there—though without also noticing that the same was true for `name`.

chrisbobbe:
Merged, thanks!

Issue #4508, opened 2021-03-04
MessageReactionsScreen: Overlapping of emojis tabs 


rajprakash00:
Currently if we have a bunch of emoji reactions on a message and then on visiting **See who reacted** screen
 we can see the emoji tabs are overlapping upon each other thereby making it difficult to view reaction count no. & 
giving a weird view for any user.

**Screenshots/Media**:

https://user-images.githubusercontent.com/42652941/109925782-2cc07780-7ce8-11eb-952a-604171ed77c2.mp4

**Additional Context**:

Working on a PR for this

chrisbobbe:
Thanks, @rajprakash00! Yeah; I agree this is annoying. I'm closing this as a duplicate of #3798, though, which has gathered more design discussion about what to do about it. Please let me know if I've misunderstood the symptom here, though. 🙂

Issue #3798, opened 2020-01-14
Emojis cut-off in Reactions screen when number of unique reactions is large


agrawal-d:
![Screenshot_20200115-015143_Zulip](https://user-images.githubusercontent.com/7714968/72379691-d66f2e80-370b-11ea-811c-89df142d4d7e.jpg)

chrisbobbe:
Aha, you beat me to it! Thanks for opening this issue @agrawal-d as this was exactly what I was testing for.

The problem also occurs on iOS:

![image](https://user-images.githubusercontent.com/22248748/72380677-c3883a00-36ca-11ea-87f1-6a9cfe5c5156.png)

I'll post more soon.

chrisbobbe:
So I see two ways we could fix this:

1) Let the tabs maintain enough width to show the entire emoji, and allow scrolling through the tabs. This would be an easy fix, mainly adding `scrollEnabled: true` in tabBarConfig in the TabNavigatorConfig (https://reactnavigation.org/docs/en/2.x/material-top-tab-navigator.html#tabnavigatorconfig). I've confirmed that this simple config change works on iOS, though I would add a rule to decrease the max width of each tab:

![image](https://user-images.githubusercontent.com/22248748/72381306-0dbdeb00-36cc-11ea-85a2-4a7e275deea9.png)

2) We could remove the tab format and use a SectionList (https://facebook.github.io/react-native/docs/sectionlist) instead, where headers are emoji and items are users.

In either case, I think it would be nice to sort the emoji descending by the number of users, so if 15 people reacted with, e.g., 👍 , and 14 people reacted with 👎 , the 👍 would appear first.

I don't have very strong feelings either way, but one advantage of a SectionList is that it gives a (much) larger area for users to interact with, and it allows interacting from the bottom of the screen, which is easier to reach. This consideration may make up for the downside that the amount of required scrolling increases with the total number of users who have reacted, not just the number of unique reactions.

agrawal-d:
This is how Slack does it:
![image](https://user-images.githubusercontent.com/7714968/72383440-7bf0c500-3741-11ea-91de-d2e69ae3a956.png)

agrawal-d:
During the design process of this feature, the initially proposed design was similar to that of slack (https://github.com/zulip/zulip-mobile/pull/3110)

chrisbobbe:
> During the design process of this feature, the initially proposed design was similar to that of slack (#3110), and would not have faced this issue.

[...]

PR #4509, opened 2021-03-04
MessageReactionsScreen: Set `scrollEnabled` to true.

rajprakash00:
Currently in `MessageReactionsScreen` when we have a bunch of emojis then
the emoji tabs are overlapping with each other making it hard for any user
to view emojis & it's reaction count.

So add a scrolling effect by setting `scrollEnabled` to true in `Tab.Navigator`.

**Screenshots/Media**:

https://user-images.githubusercontent.com/42652941/109989091-2ead2980-7d2e-11eb-9e15-dbc52016c9e2.mp4

Fixes : #4508

### DIFF
--- a/src/reactions/MessageReactionsScreen.js
+++ b/src/reactions/MessageReactionsScreen.js
@@ -99,6 +99,7 @@ class MessageReactionsScreen extends PureComponent<Props> {
                 },
               })}
               swipeEnabled
+              tabBarOptions={{ scrollEnabled: true }}
             >
               {
                 // Generate tabs for the reaction list. The tabs depend


chrisbobbe:
Hi! Before reviewing an approach like this, I'd like to hear more thoughts on some alternative designs; there's been some discussion on an existing issue around https://github.com/zulip/zulip-mobile/issues/3798#issuecomment-574370290.

chrisbobbe:
Hi, @rajprakash00! You mentioned at https://github.com/zulip/zulip-mobile/issues/3798#issuecomment-813921147 that you'd like to try a different approach than the one here; I think that makes sense. 🙂 Quoting that comment:

> > If we go the vertical list way (see my previous comment here, and still I think my preference), the shuffling would still affect tap attempts, but it would be noticeably less annoying because there's less of a demand to tap to see who the users are, because you can already see avatars. It's also a much larger area than the row of tabs.
> 
> @chrisbobbe I'm thinking the design path to be more or less like this 🙂
> Also in this way we don't have to deal with tediuos amount of scrolling for unique reactions(if tab view design is kept) and another good thing is we are also keeping the design in sync with other `SectionList`s used in zulip mobile)(e.g: `UserList` screen)

Should we close this PR, then, or are you thinking you'll update it with that new approach? Either is fine.

rajprakash00:
I think #3798  goes for the whole `MessageReactionsScreen` UI change, so better to have a separate PR for that 🤔

Issue #4510, opened 2021-03-07
Depreciation warning for "PushNotificationIOS" module of core React-Native.


Gautam-Arora24:
**What is this issue for?**
The core ```PushNotificationIOS``` module has been depreciated from React-Native. We can see this error in the console window when we start the server.
This is where official RN [docs](https://reactnative.dev/docs/pushnotificationios) mention this.
![Screenshot 2021-03-06 at 2 36 35 PM](https://user-images.githubusercontent.com/53913514/110248495-81d5e500-7f97-11eb-9b91-113ef556cbfc.png)

**Steps to solve this issue**
We will use the react-native-community/push-notification-ios package. Instead of using PushNotificationIOS from react-native we will use the same from the above-mentioned package in /src/notification/index.js. The docs of the package mention the use of the PushNotificationIOS.

Gautam-Arora24:
Working on this!

chrisbobbe:
Thanks, @Gautam-Arora24!

This deprecation warning is covered in issue #4115, which has PR #4163 open for it, so I'll go ahead and close this.

I probably could have done a better job of writing #4115, to include search terms like `PushNotificationIOS`, to make it easier for you to find; sorry about that. 🙂

PR #4511, opened 2021-03-07
deps: Use '@react-native-community/push-notification-ios'.

Gautam-Arora24:
**How I solved this issue?**
I used the react-native-community/push-notification-ios package. Instead of using PushNotificationIOS from react-native I used the same from the above-mentioned package in /src/notification/index.js. The docs of the package mentions the use of the PushNotificationIOS.
Fixes: #4510

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -358,6 +358,8 @@ PODS:
     - React
   - RNCMaskedView (0.1.10):
     - React
+  - RNCPushNotificationIOS (1.8.0):
+    - React-Core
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Thanks for noticing this, @Gautam-Arora24!

It turns out that a PR for using @react-native-community/push-notification-ios is already open, as #4163, and it's gathered more detail than this one, through a couple rounds of review. 🙂 So I'm closing this one.

Thanks again, though!

Gautam-Arora24:
Thanks, @chrisbobbe I didn't notice that :)

PR #4512, opened 2021-03-07
Added text labels to active destinations

### LINKED ISSUE #3083: layout: Add text labels to active destinations on the bottom navigation bar.
From [this Material Design article](https://material.io/design/components/bottom-navigation.html#anatomy) on bottom navigation bars:

> The way bottom navigation destinations are represented can depend on how many are used ... 
> - Four destinations: Active destinations display an icon and text label. Inactive destinations display icons, and text labels are recommended.

I like the animation and styling of active destinations of the bottom navigation bar in [this particular example](https://storage.googleapis.com/spec-host-backup/mio-design%2Fassets%2F1-Y09wDWkBCQtc96nxBzY-e0c34Aok4Dl%2Fbottomnav-icons-1a.mp4), which comes from the same article.

Here's how Zulip's bottom navigation bar looks at the moment (screenshot from an iPhone 7 simulation):
<p align="center">

</p>

(@zulipbot add "area: layout")
(@zulipbot claim)

im-adithya:
## Description
Adds the labels to active destinations in bottom navigation bar
<p>

</p>
## Issues related
This PR closes #3083


src/main/MainTabsScreen.js
```diff
@@ -89,31 +89,48 @@ function MainTabsScreen(props: Props) {
           component={HomeScreen}
           options={{
             tabBarLabel: 'Home',
-            tabBarIcon: ({ color }) => <IconInbox size={24} color={color} />,
+            tabBarIcon: properties => (
+              <View style={styles.center}>
+                <IconInbox size={24} color={properties.color} />
+                {properties.focused && <Text style={styles.bottomNav}>Home</Text>}
+              </View>
+            ),
```
gnprice, line 97: Note there's already a label here: `tabBarLabel: 'Home'`.

Currently we don't usually show that label when on a phone, but we do on a tablet -- I see them on my iPad mini 4. I think if you create an emulated device that's an Android tablet, you'll see the same thing.

We should probably just control when those existing labels are shown. Certainly we should avoid showing double labels. :slightly_smiling_face:

src/main/MainTabsScreen.js
```diff
@@ -89,31 +89,48 @@ function MainTabsScreen(props: Props) {
...
           }}
         />
         <Tab.Screen
           name="pm-conversations"
           component={PmConversationsScreen}
           options={{
             tabBarLabel: 'Conversations',
-            tabBarIcon: ({ color }) => <IconUnreadConversations color={color} />,
+            tabBarIcon: properties => (
+              <IconUnreadConversations color={properties.color} focused={properties.focused} />
```
gnprice, line 119: Why no label here? Seems like that'd be awfully inconsistent in the UI.

src/main/MainTabsScreen.js
```diff
@@ -89,31 +89,48 @@ function MainTabsScreen(props: Props) {
...
+            ),
           }}
         />
         <Tab.Screen
           name="stream-tabs"
           component={StreamTabsScreen}
           options={{
             tabBarLabel: 'Streams',
-            tabBarIcon: ({ color }) => <IconStream size={24} color={color} />,
+            tabBarIcon: properties => (
```
gnprice, line 105: I think the `({ color })` destructuring syntax (that we're currently using) is a bit cleaner here, for these very short functions.

gnprice (commented):
Thanks @im-adithya ! Comments below.

src/main/MainTabsScreen.js
im-adithya: It has a label Greg, the problem is that the component `IconUnreadConversations` isn't actually an Icon that accepts text but a view component (that has an icon inside it), so I added the label in IconUnreadConversations.js instead of placing it here!

src/main/MainTabsScreen.js
im-adithya: But I had to check whether the component is active or not, fetched from the `focussed` key of the object, to render the label... 😕

src/main/MainTabsScreen.js
im-adithya: Sure, Greg! I didn't check in a tablet-sized-emulator so I thought it was something placed for accessibility, will use media queries and sort this!

src/main/MainTabsScreen.js
gnprice: Right, so you can get that the same way: `({ color, focused }) => …`

src/main/MainTabsScreen.js
gnprice: Ahh, I see. But why wouldn't it work to do the same thing for this one as for the others?

As is, it looks like it'd cause the `CountOverlay` to enclose the label as well as the icon. I'd expect that won't look right.

src/main/MainTabsScreen.js
im-adithya: Oh sorry! I didn't read `destructuring syntax` 😅

src/main/MainTabsScreen.js
im-adithya: > But why wouldn't it work to do the same thing for this one as for the others?

Sorry, Greg! It does work that way. Seems like I've made some mistake previously...
Now it works as expected!

src/main/MainTabsScreen.js
im-adithya: Done! Had to set showLabel to `false` in place of `Platform.isPad` BTW, I didn't see the labels on my (android) tablet emulator... does it only recognize Apple devices?

Issue #161, opened 2016-12-30
Display Current Server in a prominent place


borisyankov:
It is quite confusing currently, if you have accounts for several servers.

zulipbot:
Hello @borisyankov, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

kunall17:
@borisyankov Would the settings tab be the perfect component to show this information?

borisyankov:
Not really. We should just make the account list more prominent or something.

vishwesh3:
Might be duplicate of https://github.com/zulip/zulip-mobile/issues/392

chrisbobbe:
> Might be same as #392

I think #392 is about showing an icon for each realm in the org switcher; I think this issue is asking for a prominent display of the realm info (icon, name, etc.?) while you're logged in and active with an account.

Sometimes it takes me a second or two to check and see if I'm in the Recurse realm or the Zulip realm.

alya:
This should be implemented after #M4851.

PR #4517, opened 2021-03-09
issue #161 Displaying current server in home-page

vaibhavs2:
Issue #161 
Please refer to PR #4516  for details that I closed by mistake and also it was failing CI tests.
- with small update, both email and realm in toast message


vaibhavs2:
I am working on it

chrisbobbe:
Hi, @vaibhavs2! Thanks for your work on this.

I see this has some CI failures and has gathered some rebase conflicts, and the commits aren't yet clear and coherent in this particular revision. Are you looking for any feedback to help get the PR back on track, or are you thinking of moving on to other work or starting over? 🙂 Anything is fine, I'm just wondering if you think it might be time to close this to help clean up the PR tracker.

vaibhavs2:
yes , that time I did't familiar with the code quality, flow-types, indentation etc. that's why it got conflict errors, I have a ready updated code for it but I didn't push that yet, I am waiting for comments is this idea worth? if it worth then I am ready to work on it too, now I have to make changes as realmAdd and some more functionality has changed now. Details are here  #4516 , by mistake I opend two PR because its the first time I faced something like `CI check` and got confused what just happened that time.

chrisbobbe:
> If the idea is not providing any value then close it :)

Ideas can certainly provide value even if we end up closing the issue/PR they were presented in. 🙂 Sometimes we start out thinking we'll do something one way, then we explore doing it that way and that causes us to try a different way, etc., and our thinking moves forward.

I appreciate all your screenshots (here and on #4516); thanks for those! I have a few concerns with the proposed UI; I'm basing these on those screenshots and your descriptions, as I haven't yet checked out the PR to try it locally.

- If meaning is only expressed through color, people with colorblindness can have a worse experience.
- I think the dot on the tab icon looks a bit too much like a notification indicator, saying "please check out this tab, it has news for you!" If I understand correctly, that's not what it really means; it means to show what account/server is active by showing its corresponding color.
- Long-press interactions can be hard to make discoverable.
  - Ah, and in fact—when the app's performance is doing well and things are responding quickly, it may actually be faster to learn the account/realm details with a _short_-press on that tab icon, than with a long-press. That's because a short-press (which is shorter) opens the account details screen that should have all this stuff. I think this might lead to the long-press not getting used much.

@gnprice, @WesleyAC, @alya, and I have recently been chatting about a [drawer-navigation](https://material.io/components/navigation-drawer) idea, and some other ideas, for #161 (and better multi-account support in general). Those ideas go in different directions than this PR, but of course we're happy to hear feedback. So I'll close this, but again I appreciate your work on it. 🙂

PR #4518, opened 2021-03-10
Flow-related work on the path to RN v0.64 upgrade, part 1.

### LINKED ISSUE #3452: Convert all object types to exact, or explicitly-inexact

chrisbobbe:
This resolves the long-standing comments dotting our code like "this really should be an exact object type -- see note in jsonable.js about objects with indexer properties."

As one might remember from d1affaf78, that was something we weren't expecting to be able to do until Flow v0.126. We're on Flow v0.122, and the latest release candidate for RN v0.64 [is on Flow v0.137](https://github.com/facebook/react-native/blob/v0.64.0-rc.4/package.json#L120). So we'd generally wouldn't expect this work to be doable until after we're on RN v0.64; that's #4426.

Well, my solution is to actually take the Flow v0.126 upgrade now, even before upgrading React Native. I also take a couple more Flow upgrades, through v0.128, for some other things.

Our normal practice is to change the Flow version only in the same commit that we change the React Native version. That's because the code for a React Native release (the code in node_modules/react-native) is only guaranteed to pass type-checking with the Flow version that was used to check the code at that release. Using a different version of Flow means we might run into errors in node_modules/react-native, and we depend on that being error-free, and without suppressions, because we want our use of React Native's APIs to be type-checked.

But I think the benefits outweigh the costs in this case:

- It lets us make certain changes more incrementally, without crowding those changes at and around the RN v0.64 upgrade commit. That kind of crowding is a major source of complexity when doing RN upgrades.
  - Getting this Flow bugfix will help us fix #3452 much more productively: we'll get to skip the step of marking all these object types as explicitly inexact when we really want them to be exact. #3452 is basically a prerequisite for taking facebook/react-native@050a7dd01 (switching on `exact_by_default`), and that's on the path to the RN v0.64 upgrade. That commit message warns that React Native's code may assume that projects turn on `exact_by_default`.
  - We get to drop `$FlowMigrationFudge` and our other uses of `suppress_comment` (as Greg [noted](https://github.com/zulip/zulip-mobile/issues/4426#issuecomment-767226886) we'd have to) now instead of later. Including the ability, with Flow v0.128, to use `$FlowIgnore` instead of `$FlowFixMe` in the places we used to use `$FlowMigrationFudge`.
- We got lucky this time: going from v0.122 to v0.129 only introduced a Flow error in one React Native file: `node_modules/react-native/Libraries/ReactNative/PaperUIManager.js`. I've never heard of this file, we never import it, and I don't expect we'll cause any problems by temporarily (until we're on RN v0.64) telling Flow to ignore it.

I have a draft on top of this that goes further toward #3452. The "implicit-inexact-object" lint Greg mentions at https://github.com/zulip/zulip-mobile/issues/3452#issuecomment-657778760 finds 192 warnings (142 in our own code, the rest in libdefs) at the current `master` (before this PR). At the tip of that draft in its current state, it finds just 54 warnings, with 19 in our own code, and those numbers are decreasing 😄.

### DIFF
--- a/flow-typed/@react-navigation/bottom-tabs_v5.x.x.js
+++ b/flow-typed/@react-navigation/bottom-tabs_v5.x.x.js
@@ -26,7 +26,7 @@ declare module '@react-navigation/bottom-tabs' {
     | false
     | ''
     | $ReadOnlyArray<StyleObj>
-    | { [name: string]: any, ... };
+    | {| [name: string]: any |};
   declare type ViewStyleProp = StyleObj;
   declare type TextStyleProp = StyleObj;
   declare type AnimatedViewStyleProp = StyleObj;
@@ -425,7 +425,7 @@ declare module '@react-navigation/bottom-tabs' {
    * Actions, state, etc.
    */
 
-  declare export type ScreenParams = { +[key: string]: mixed, ... };
+  declare export type ScreenParams = {| +[key: string]: mixed |};
 
   declare export type BackAction = {|
     +type: 'GO_BACK',
@@ -474,7 +474,7 @@ declare module '@react-navigation/bottom-tabs' {
 
   declare export type GenericNavigationAction = {|
     +type: string,
-    +payload?: { +[key: string]: mixed, ... },
+    +payload?: {| +[key: string]: mixed |},
     +source?: string,
     +target?: string,
   |};
@@ -540,10 +540,9 @@ declare module '@react-navigation/bottom-tabs' {
   declare type ActionCreators<
     State: NavigationState,
     Action: GenericNavigationAction,
-  > = {
+  > = {|
     +[key: string]: (...args: any) => (Action | State => Action),
-    ...
-  };
+  |};
 
   declare export type DefaultRouterOptions = {
     +initialRouteName?: string,
@@ -556,7 +555,7 @@ declare module '@react-navigation/bottom-tabs' {
     RouterOptions: DefaultRouterOptions,
   > = (options: RouterOptions) => Router<State, Action>;
 
-  declare export type ParamListBase = { +[key: string]: ?ScreenParams, ... };
+  declare export type ParamListBase = {| +[key: string]: ?ScreenParams |};
 
   declare export type RouterConfigOptions = {|
     +routeNames: $ReadOnlyArray<string>,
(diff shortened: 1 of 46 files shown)


flow-typed/react-intl_vx.x.x.js
```diff
@@ -75,8 +74,7 @@ declare module 'react-intl' {
     | 'textComponent'
     | 'defaultLocale'
     | 'defaultFormats'
-    | 'onError',
-  >;
+    | 'onError', >;
```
gnprice, line 77: Huh, weird. But I reproduce that that's what auto-formatting does... and this is not code we spend a lot of time in or really fully maintain, so probably better to just let it do as it likes here rather than try to persuade it to do something else. :shrug:

flow-typed/@react-navigation/bottom-tabs_v5.x.x.js
```diff
@@ -540,10 +540,9 @@ declare module '@react-navigation/bottom-tabs' {
   declare type ActionCreators<
     State: NavigationState,
     Action: GenericNavigationAction,
-  > = {
+  > = {|
     +[key: string]: (...args: any) => (Action | State => Action),
-    ...
-  };
+  |};
```
gnprice, line 545: Here, there were substantially more *explicitly* inexact object
    types with indexer properties. It doesn't make a lot of sense why
    one would want an inexact object with an indexer property, whether
    implicitly or explicitly inexact. So, make those exact too.

Hmm, yeah.

These are mostly lightly-edited flowgen output, right? Seems like probably a (minor) bug in flowgen.

gnprice (commented):
Looks great, thanks! Merged.

chrisbobbe:
Thanks for the quick review!

PR #4476, opened 2021-02-14
Task debug icon

im-adithya:
## Description
I added the logo with DEBUG banner and then modified the scripts accordingly to generate the logos using inkscape

## Earlier
image
## Now
zulip-icon-debug

### After generating icons using modified script:
Screenshot from 2021-02-14 13-25-54

## Issues related
This PR closes #4210

## Additional Information
I changed the `make_android_icon()` function to take an additional argument to decide between main and debug locations

Issue #4130, opened 2020-05-29
Run `shellcheck`, add to `tools/test`.


chrisbobbe:
When @gnprice [said](https://github.com/zulip/zulip-mobile/pull/4028#pullrequestreview-390311786):

> @andersk Do you know of another good resource to point to for the key points of writing shell scripts? I linked to two below but I feel like there are a few other good things to mention that they don't cover.

@andersk [replied](https://github.com/zulip/zulip-mobile/pull/4028#issuecomment-611209553):

> Yes! Run ShellCheck on all of your shell scripts and get them down to zero diagnostics. Run it every time you commit. Run it in CI. Run it automatically in your editor. ShellCheck knows things that you didn’t know you needed to know. ShellCheck knows things that _I_ didn’t know I needed to know. ShellCheck is excellent. ShellCheck is magical. ShellCheck is the one shining beacon of hope in the vast fiery hellscape that is the world of UNIX shell scripting. Run ShellCheck.
> 
> Run it.

Running `shellcheck tools/test` on master currently gives five or six warnings in that script. I haven't investigated any of them yet, but it sounds like a good goal to do so, and to maybe make `tools/test shell` a thing.

PR #4519, opened 2021-03-10
Some tweaks to tools/generate-logos

gnprice:
A followup to #4476.
* Fix it so that `git status` stays clean unless something changed.
* Document explicitly how we keep the output tracked in Git.
* Switch `make_one_android_icon` to something like named parameters rather than positional, as it now has 6 of them.

(@im-adithya fyi)

### DIFF
--- a/tools/generate-logos
+++ b/tools/generate-logos
@@ -138,7 +138,8 @@ make_webp() {
 }
 
 make_one_android_icon() {
-    local src="$1" sourceset="$2" restype="$3" name="$4" size_px="$5" density="$6"
+    # SET BY CALLER: src sourceset restype name
+    local size_px="$1" density="$2"
     local output=android/app/src/"${sourceset}"/res/"${restype}"-"${density}"/"${name}".webp
     mkdir -p "${output%/*}"
     make_webp "${src}" "${size_px}" "${output}"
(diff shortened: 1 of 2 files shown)


tools/generate-logos
```diff
@@ -78,7 +78,7 @@ make_one_ios_app_icon() {
 }
 
 make_ios_app_icon() {
-    local iconset=ios/ZulipMobile/Images.xcassets/AppIcon.appiconset
+    local iconset=ios/ZulipMobile/Assets.xcassets/AppIcon.appiconset
```
chrisbobbe, line 87: Ah, indeed—thanks for catching this!

tools/generate-logos
```diff
@@ -138,7 +138,8 @@ make_webp() {
 }
 
 make_one_android_icon() {
-    local src="$1" sourceset="$2" restype="$3" name="$4" size_px="$5" density="$6"
+    # SET BY CALLER: src sourceset restype name
```
chrisbobbe, line 141: Interesting; `make_one_android_icon` uses these variables as they're defined in its caller, even when the caller uses `local`.

From the Bash Manual section 3.3 ([link](https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions), formatting lost):

> The shell uses dynamic scoping to control a variable’s visibility within functions. With dynamic scoping, visible variables and their values are a result of the sequence of function calls that caused execution to reach the current function. The value of a variable that a function sees depends on its value within its caller, if any, whether that caller is the "global" scope or another shell function. This is also the value that a local variable declaration "shadows", and the value that is restored when the function returns.
> 
> For example, if a variable var is declared as local in function func1, and func1 calls another function func2, references to var made from within func2 will resolve to the local variable var from func1, shadowing any global variable named var.
> 
> The following script demonstrates this behavior. […]

chrisbobbe (approved):
Thanks, @gnprice! I verified that `shellcheck` (#4130) doesn't find any additional errors in the script after this PR, vs. just before. Merged!

tools/generate-logos
gnprice: Yep. `local` protects the given variables between this function and its caller, but not between this function and its callees.

And actually it doesn't 100% protect the function from caring about what its caller might do with those variable names, either -- see this wrinkle in `tools/build-webview`:
```bash
readonly target dest check_path_name verbose

# …

sync() {
    # "_dest" because bash's `readonly` means can't shadow with `local`
    local src="$1" _dest="$2"
```

That is, even with `local` the function isn't totally free to pick a name of its choice without worrying about what names its calling code might have already used: if the caller has already used some name *and made it `readonly`*, then Bash won't let this function reuse the name even with `local`.

(This probably means it's not actually a good idea to routinely mark things `readonly`, as we do in that script.)

Issue #4533, opened 2021-03-14
Fix alignment of embedded link previews


rajprakash00:
Currently all embedded images are center aligned in mobile unlike in zulip web where all are left aligned.
Center-aligned images seems fine but if we have back to back different orientation images then it becomes 
quite weird to view in mobile unlike in web.

An example **screenshot**:
![zulipBefore5](https://user-images.githubusercontent.com/42652941/111075092-480f5c00-850c-11eb-8fe0-3f94d8a42aca.png)

**Web:**
![zulipWeb5](https://user-images.githubusercontent.com/42652941/111075777-8b1efe80-850f-11eb-88a0-8a8ed9b654dd.png)

**Solution:**
Keep the styling of embedded message images same as zulip web(all left aligned images.)

**Additonal Context:**
Working on a PR for this

PR #4534, opened 2021-03-15
Fix alignment of embedded link previews.

rajprakash00:
Currently all embedded images are center aligned in mobile unlike in zulip
web where all are left aligned.Center-aligned images seems fine but for
different orientation images it becomes quite weird to view in mobile unlike
in zulip web.

So ,make the link previews left aligned.(Twitter images are already
within an enclosed box & it seems perfectly fine being centred so no change
in that.)

**Screenshot:**
zulipAfter5

Fixes : #4533

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -484,12 +484,13 @@ time {
 }
 
 .message_inline_image {
-  text-align: center;
+  margin-top: 5px;
 }
 .message_inline_image img,
 .message_inline_ref img,
 .twitter-image img {
-  width: 100%;
+  width: auto; 
+  max-width: 100%; 
   height: 160px;
   object-fit: contain;
 }


src/webview/static/base.css
```diff
@@ -484,12 +484,13 @@ time {
 }
 
 .message_inline_image {
-  text-align: center;
+  margin-top: 5px;
```
gnprice, line 487: What's the reasoning for adding to the margin here? Is this related to the new left-alignment, or is it for another reason?

src/webview/static/base.css
```diff
@@ -484,12 +484,13 @@ time {
...
 }
 .message_inline_image img,
 .message_inline_ref img,
 .twitter-image img {
-  width: 100%;
+  width: auto; 
+  max-width: 100%;
```
gnprice, line 493: What's the effect of this change? It doesn't seem to be described in the commit message.

gnprice (commented):
Thanks @rajprakash00 !

I'm not sure I love the left alignment, but it probably is best to have a consistent layout with the webapp in that respect, so that sounds like a good change.

Comments below -- mainly it looks like there are some other changes here too and I'm curious about them.

src/webview/static/base.css
gnprice, line 493: Oh, and a nit: no trailing whitespace at end of line.

src/webview/static/base.css
rajprakash00: No its not related to new alignment, rather it's just to make sure mobile has same styling design with web-app, since there it has been kept 
`margin:3px 0 3px 0`([here](https://github.com/zulip/zulip/blob/6b7a3fb74acb9e8992d256a2da3b75fcb22befb9/static/styles/rendered_markdown.css#L4)) and that provides necessary 3px gap between image link and image.(and since unlike web ,here image link underline remains as it is so it seemed better with 5px)

src/webview/static/base.css
rajprakash00: Hmm...  a change I should have written description of it, but as far as I remember this is the main logic for left alignment of image.

I tried with `float:left` first(just like in web) but it didn't work(🤔 ) and upon using `max-width:100%` or `width:100%` it gave desired result but still 1 problem remained,i.e, message reactions got shifted to the row of respective image and the reaction and image both got aligned somewhat like`flex-direction:row` property(weird,..I think because of `<span>` property but I'm not sure why)

So the final solution was to keep `width:auto` and also giving `max-width` as 100% and it worked without the need for `float` property unlike in web.

Issue #4530, opened 2021-03-13
Markdown tables with long text can get cut off


WesleyAC:
A markdown table such as:

```
| a | b |
| --- | --- |
| `ajkljdslkgjalkfjksldjfk lsjdf jsdlfkjds lkfjdksl jfkldsj kldsjfl kjds lkjdsklfj ldskjf lsdkjfl ksdjfkl jsdlkfj klsdj fldsj klfjdslkfjldsk jkldsj fdsl kfdsj l` | supercalifragilisticexpialidocious |
| 1 | 2 |
```

Will have the text "supercalifragilisticexpialidocious" cut off on the right side. Instead of cutting off the text, the table should scroll horizontally.

gnprice:
Agreed.

This is probably just a line or two of CSS to fix -- the main work needed is finding the right change like that to make.

For whoever wants to take this up, I'd suggest taking a look at how we handle LaTeX math displays: https://zulip.com/help/format-your-message-using-markdown#latex . I believe if you have a message with a long equation that doesn't fit on the screen, we'll have that scroll horizontally. So it may be a matter of using the same bit of CSS here as we do there.

rajprakash00:
@gnprice  Can i work on this issue?

timabbott:
It's also worth looking at `static/styles/rendered_markdown.css`, both for whether we have the same bug and if not, whether we can use the same solution.

PR #4536, opened 2021-03-16
Markdown tables: Add horizontal scrolling.

rajprakash00:
Currently a markdown table having long text gets cut off from the right side
instead of scrolling.

Added horizontal scrolling effect.

**ScreenShots/Media**:

https://user-images.githubusercontent.com/42652941/111325017-4e7b1080-8691-11eb-8d26-9b935eddf39a.mp4

Fixes : #4530

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -437,7 +437,9 @@ blockquote {
 
 table {
   border-collapse: collapse;
-  width: 100%;
+  max-width: fit-content;
+  display: block;
+  overflow-x: auto;
 }
 table, th, td {
   border: 1px solid hsla(0, 0%, 50%, 0.25);


rajprakash00:
@gnprice Please take a look

WesleyAC (approved):
This looks good to me, thanks for doing this :)

WesleyAC:
Actually, on second thought, shouldn't a `table` element be `display: table` (the default)? Is the `display: block` needed here? [this stack overflow answer](https://stackoverflow.com/questions/7455650/setting-a-table-to-display-block) suggests that `display: block` on a table could cause problems, although doesn't go into any details.

WesleyAC:
Specifically, it seems like `display: block` on a table may [break screen reader support](https://www.tpgi.com/short-note-on-what-css-display-properties-do-to-table-semantics/). Unfortunately, it seems like the only solutions for this that don't require `display: block` rely on wrapping the `<table>` in another element. For what it's worth, tables are `display: block` in the webapp.

Perhaps doing `display: block` and setting the [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) `role=table` on the table would be easier.

Then again, I'm not even sure that screen-readers on mobile devices trip up on this — it seems this was fixed in firefox [three years ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1005271), hopefully the webview UAs have fixed this by now as well?

@rajprakash00, would it be possible for you to test how the table is read by [talkback](https://support.google.com/accessibility/android/answer/6007100) on android, before and after this change, to ensure that it's the same?

WesleyAC:
@Gautam-Arora24 FYI re: accessibility, since I've seen you working on that on CZO — feel free to chime in if you have an idea of what the behaviour is here :)

rajprakash00:
> Actually, on second thought, shouldn't a `table` element be `display: table` (the default)? Is the `display: block` needed here? [this stack overflow answer](https://stackoverflow.com/questions/7455650/setting-a-table-to-display-block) suggests that `display: block` on a table could cause problems, although doesn't go into any details.

@WesleyAC  Yes it should be `display:table` but it seems horizontal scrolling becomes a problem upon doing that
.Although `display:block` seems to come to rescue many developers have strictly denied to use it as it breaks semantics and 
has accessibility issues.


> @rajprakash00, would it be possible for you to test how the table is read by [talkback](https://support.google.com/accessibility/android/answer/6007100) on android, before and after this change, to ensure that it's the same?

Hmm, weird I tried it the spoken feedback was not coming even for normal texts in message-list ,I wonder why it could be?🤔

WesleyAC:
Ok, I just tried out this PR with talkback on android on my phone, and it seems to work the same as it does in the current app, which is the desired behaviour: when you tap a cell, it reads the column name, and when you change rows, it reads the new row number and the column name. Not sure why it wasn't working for you — you do need to tap on the messages in the message list for the spoken feedback to happen, which can be tricky to get the hang of, but it could also be some other bug stopping it from working on your device.

I'm not sure how [VoiceOver](https://support.apple.com/guide/iphone/turn-on-and-practice-voiceover-iph3e2e415f/ios) on iOS handles this, but the fact that Android gets it right gives me a fair amount of confidence.

WesleyAC (approved):

rajprakash00:
Weird , I tried it in whatsapp it worked fine though :(.

Gautam-Arora24:
> Hmm, weird I tried it the spoken feedback was not coming even for normal texts in message-list, I wonder why it could be?🤔
@rajprakash00 For TalkBack to work on Android you need to double-tap to perform an action and just a single tap to hear the screen reader. You can also directly tap with two fingers to perform an action without listening to the screen reader. And yes, to scroll, use two fingers.

gnprice:
Thanks @rajprakash00, and thanks @WesleyAC for the detailed review!

Merged, with just small tweaks to the commit message:
```diff
-    Markdown tables: Add horizontal scrolling.
+    webview styles: Let tables scroll horizontally.
 
-    Currently a markdown table having long text gets cut off from the right side
-    instead of scrolling.
+    Currently a markdown table having long text gets cut off from the
+    right side instead of scrolling.
 
     So, add horizontal scrolling effect.
 
-    Fixes : #4530
+    Fixes: #4530
     Signed-off-by: rajprakash00 <rajprakash1999@gmail.com>
```

timabbott:
Does the Zulip webapp have anything we should fix as a follow-up to this?  If so, it'd be great to have something filed there.

rajprakash00:
@timabbott I have checked in webapp and markdown tables seems to work perfectly fine in it. 🙂

timabbott:
Great, thanks for the confirmation!

Issue #4521, opened 2021-03-11
Display problem in the "@" icon on the home screen


blinry:
On version 27.159 of the Android app (latest version from the Play Store), and on Android 9, the "@" icon in the top bar has a display problem where the path is not filled with color where it's self-intersecting:

![](https://user-images.githubusercontent.com/81277/110769914-fdb28480-8258-11eb-9109-ae1eb1c502d5.png)

Can other people reproduce this? I've been noticing this for a while. I don't know enough about the involved technologies to fix this myself –  it might be an upstream issue with react-native-vector-icons, or just a setting to say that the [fill-rule](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule) of the SVG should be "nonzero" instead of "evenodd"?

Gautam-Arora24:
Seeing this issue on iOS as well. 
My iOS version - 14.3

chrisbobbe:
Thanks for the report, @blinry, and for confirming on iOS, @Gautam-Arora24! Good to have an issue for this in the tracker.

@WesleyAC has debugged this and filed oblador/react-native-vector-icons#1311. Once that fix lands upstream and we can take a new release with it, it should resolve.

WesleyAC:
Honestly, it might be good to just revert RNVI to the previous version until they fix this, since it's been several weeks since I filed that PR, it could be sitting there for quite a while. I don't *think* there are any cutting-edge features we need, although I haven't looked closely. Looks like `6.4.2` was the last version to not have this issue, we're on `^8.0.0`.

Gautam-Arora24:
But moving back to ```6.4.2``` will cut many features for us which may be important.

WesleyAC:
Looking at the commit log for all the upgrades, I think #2209 is the only thing that would revert. And personally, I'd rather have #2209 than this, but maybe there are features that we rely on in upgrades that weren't mentioned in the commit log.

Gautam-Arora24:
Here how it looks on an iPhone Simulator. Certainly, the @ symbol is empty and the path is not filled
![Screenshot 2021-03-16 at 10 02 46 PM](https://user-images.githubusercontent.com/53913514/111345490-593ea100-86a3-11eb-9c1c-49b58644b928.png)

chrisbobbe:
> Looks like `6.4.2` was the last version to not have this issue, we're on `^8.0.0`.

[...]

PR #4537, opened 2021-03-16
Fix appearance of "@" icon in the home screen.

chrisbobbe:
By downgrading react-native-vector-icons while we wait for oblador/react-native-vector-icons#1311 to land, as @WesleyAC suggests at https://github.com/zulip/zulip-mobile/issues/4521#issuecomment-797219057.

This PR downgrades to 7.0.0, instead of 6.4.2, following my observations at https://github.com/zulip/zulip-mobile/issues/4521#issuecomment-800448323.

I haven't seen any noticeable regressions from doing the downgrade.

One thing stands out in the changelog: a claim that 8.0.0 has better Xcode 12 compatibility. It doesn't go into detail (though I believe oblador/react-native-vector-icons#1251 is meant), and I was able to build and run for iOS with version 7.0.0 of react-native-vector-icons, in both debug and release modes, using Xcode 12.4. So possibly there was a wrong diagnosis, or possibly an issue was encountered with some Xcode 12.x that was fixed at or before Xcode 12.4, etc.

If it turns out to be correct to go all the way back to 6.4.2, we should do that instead.

Fixes: #4521

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -372,7 +372,7 @@ PODS:
     - RNSound/Core (= 0.11.0)
   - RNSound/Core (0.11.0):
     - React
-  - RNVectorIcons (7.1.0):
+  - RNVectorIcons (7.0.0):
     - React
   - Sentry (6.1.4):
     - Sentry/Core (= 6.1.4)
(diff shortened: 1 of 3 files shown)


WesleyAC (approved):
Thanks!

gnprice:
Looks great, thanks! Merged.

Issue #4378, opened 2021-01-04
Mobile app can create post without a topic, when permissions are set not to allow that.


vanclute:
Mobile app users can still create "no topic" posts even if permissions are set to disallow it and desktop users cannot.  Tested on iOS.

gnprice:
@vanclute Thanks for the report. I've split off the second issue in your latter comments as #4379.

This would be a good issue for someone to fix. The approach will look something like:
* Store the relevant setting in `RealmState`, and populate it in `realmReducer`, much like the existing property `canCreateStreams`.
  * If it's not already in the `InitialData` type, add it.
* At `ComposeBox`, add the setting to the props and `connect` call, and use it.

dhanmoni:
I want to work on this. Can you assign me?

abhi0504:
@gnprice I would to like to work on this issue as i think dhanmoni is not working on it

abhi0504:
@zulipbot claim

arjunrkaushik:
@gnprice I would like to work on this issue, if it's not been resolved yet!

PR #4543, opened 2021-03-18
chat: Fix create post without a topic

AlaaElden98:
When the option 'require topics in stream messages' is checked,the user
shouldn't be able to post a message without a topic.

*Store the setting in `RealmState`
*Populate it in `realmReducer`
*Add a selector for it in `directSelector`
*Add the setting to `ComposeBox` props and `connect` call,and use it

Fixes: #4378


AlaaElden98:
@gnprice, please review the changes and let me know if something needs to be different

src/compose/ComposeBox.js
```diff
@@ -320,14 +322,18 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   handleSend = () => {
-    const { dispatch, narrow, caughtUp, _ } = this.props;
+    const { dispatch, narrow, caughtUp, _, isTopicMandatory } = this.props;
     const { message } = this.state;
+    const topic = this.state.topic;
```
chrisbobbe, line 327: nit: combine with previous line instead of making a new line, like

```js
const { message, topic } = this.state;
```

src/compose/ComposeBox.js
```diff
@@ -320,14 +322,18 @@ class ComposeBox extends PureComponent<Props, State> {
...
 
     if (!caughtUp.newer) {
       showErrorAlert(_('Failed to send message'));
       return;
     }
-
+    if (isTopicMandatory && (topic === '(no topic)' || !topic)) {
+      showErrorAlert(_('Please specify a topic'));
```
chrisbobbe, line 334: We should this new user-facing string to static/translations/messages_en.json, in the commit that adds it, so it gets localized; see our [translations doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/translations.md).

src/realm/realmReducer.js
```diff
@@ -24,6 +24,7 @@ const initialState = {
   twentyFourHourTime: false,
   canCreateStreams: true,
   isAdmin: false,
+  isTopicMandatory: false,
```
chrisbobbe, line 27: Let's add a `dropCache` migration in src/boot/store.js for this. Something like,

```js
  // Add `isTopicMandatory` to `state.realm`.
  '27': dropCache,
```

src/realm/realmReducer.js
```diff
@@ -72,6 +73,8 @@ export default (state: RealmState = initialState, action: Action): RealmState =>
         twentyFourHourTime: action.data.twenty_four_hour_time,
         canCreateStreams: action.data.can_create_streams,
         isAdmin: action.data.is_admin,
+
+        isTopicMandatory: action.data.realm_mandatory_topics,
```
chrisbobbe, line 77: nit: Instead of starting a new group with an empty line, this could maybe be moved up to join the group that goes from `nonActiveUsers` to `videoChatProvider`. I'm just basing that on the fact that those use things from `action.data` that begin with `realm_`.

src/compose/ComposeBox.js
```diff
@@ -320,14 +322,18 @@ class ComposeBox extends PureComponent<Props, State> {
...
+    const { dispatch, narrow, caughtUp, _, isTopicMandatory } = this.props;
     const { message } = this.state;
+    const topic = this.state.topic;
 
     if (!caughtUp.newer) {
       showErrorAlert(_('Failed to send message'));
       return;
     }
-
+    if (isTopicMandatory && (topic === '(no topic)' || !topic)) {
```
chrisbobbe, line 333: It looks like the web app only does the empty-string check; it doesn't check for '(no topic)': https://github.com/zulip/zulip/blob/master/static/js/compose.js#L732

Let's do that, to match the web app's behavior but also to keep low the number of places where we use the string literal '(no topic)'. Ideally, I think, that string would be stored in a constant, with a comment pointing to a corresponding constant in zulip/zulip.git.

chrisbobbe (commented):
Thanks, @AlaaElden98! See a few small comments below.

Also, I think the commit message can be tightened up a bit:

```
chat: Fix create post without a topic

Why i did these changes?
-When the option 'require topics in stream messages' is checked,the user
shouldn't be able to post a message without a topic.

What changes:
* Store the setting in `RealmState`
* Populate it in `realmReducer`
* Add a selector for it in `directSelector`
* Add the setting to `ComposeBox` props and `connect` call,and use it
* Edit `realmReducer-test` to test the new changes

Fixes: #4378
```

- One thing is that the "What changes" list duplicates what's already clear by looking at the code changes, so I think that can just be removed. 🙂 The changes are pretty small, straightforward, and unsurprising.
- Let's give the summary line a bit more detail. How about something like, "ComposeBox: Prevent sending empty-topic messages if topics are mandatory."
- Then, the summary line plus the `Fixes:` line should pretty much answer the question of why the changes were made. So the "Why i did these changes?" section can disappear too, to leave us with the following:
  
  ```
  ComposeBox: Prevent sending empty-topic messages if topics are mandatory.
  
  Fixes: #4378
  ```
  
  That's shorter, and still gets the job done. 🙂

src/compose/ComposeBox.js
AlaaElden98: Make sense, done!

src/realm/realmReducer.js
AlaaElden98: That's new, I understand what this does, but when should I add a `dropCache`?

src/realm/realmReducer.js
AlaaElden98: done

src/realm/realmReducer.js
AlaaElden98: And i think you mean `'28': dropCache` ? there is already 
`'27': state => ({
    ...state,
    accounts: state.accounts.filter(a => a.email !== ''),
  }),`

AlaaElden98:
@chrisbobbe, Thanks for the review, the PR is ready for another one.

gnprice:
This issue was resolved by #4798. Thanks @AlaaElden98 for your work on this!

Issue #4546, opened 2021-03-18
Reacting with "0️⃣" results in "Invalid emoji name" error


WesleyAC:
If you try to react with the "0️⃣" (`:0:`) emoji on CZO, you get the following error:

```
[Thu Mar 18 2021 17:55:58.353]  LOG      {"httpStatus": 400, "json": {"code": "BAD_REQUEST", "msg": "Invalid emoji name.", "result": "error"}, "params": {"body": "reaction_type=unicode_emoji&emoji_code=0030-20e3&emoji_name=0", "method": "post"}, "route": "messages/1135041/reactions"}
[Thu Mar 18 2021 17:56:00.388]  WARN     Possible Unhandled Promise Rejection (id: 3):
Error: Invalid emoji name.
```

We should fix this, and add an error toast on a 400 response (currently there is no feedback given to the user about why this failed)

WesleyAC:
Ok, filed a PR to show a toast when there's an error.

As for the thing with `:0:` — the app is trying to add `:0:`, but the server only responds to `:zero:`. I think this is because we ship a `src/emoji/codePointMap.js` file that's gotten out of sync with the server version of that file? But I'm not sure how that works overall.

WesleyAC:
@gnprice let me know if you have context on what's going on here, otherwise I'm happy to investigate further.

gnprice:
Indeed, good find! As I mentioned [in the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/near/1135688), this (well, "Invalid emoji name" in general) turns out to be an error that users are regularly hitting.

We have an old issue #3128 which describes this, so I'll close this one and add an update there, to consolidate threads.

PR #4547, opened 2021-03-18
emoji: Show toast on error adding reaction emoji

WesleyAC:
Related: #4546

Screenshot:

Screenshot_20210318-231849

### DIFF
--- a/src/emoji/EmojiPickerScreen.js
+++ b/src/emoji/EmojiPickerScreen.js
@@ -16,6 +16,8 @@ import { connect } from '../react-redux';
 import { getAuth, getActiveImageEmojiByName } from '../selectors';
 import { navigateBack } from '../nav/navActions';
 import zulipExtraEmojiMap from './zulipExtraEmojiMap';
+import * as logging from '../utils/logging';
+import { showToast } from '../utils/info';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'emoji-picker'>,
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks! Will be glad to fix this. As I mentioned [in the chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/force.20handling.20promise.20errors.3F/near/1135688) after you spotted this issue and sent this PR, this is an issue that Sentry shows us is happening regularly for a steady stream of users.

Showing the user this feedback is definitely an improvement, but I believe as is this change will also have the effect of suppressing the error in Sentry -- we're currently seeing it because the promise is throwing an unhandled exception. Because this condition is something we should fix in the software, I'd like to be sure we continue to get those reports; even after we fix #4546, that will mean we learn in the future if there's a regression.

We can do that by including a call to `logging.error`. See that function's jsdoc, and see the existing places we call it for examples.

WesleyAC:
@gnprice done!

in debug mode, the error looks like this in the console:

```
[Fri Mar 19 2021 15:39:02.820]  LOG      {"httpStatus": 400, "json": {"code": "BAD_REQUEST", "msg": "Invalid emoji name.", "result": "error"}, "params": {"body": "reaction_type=unicode_emoji&emoji_code=0030-20e3&emoji_name=0", "method": "post"}, "route": "messages/1136226/reactions"}
[Fri Mar 19 2021 15:39:02.910]  ERROR    Error adding reaction emoji
[Fri Mar 19 2021 15:39:02.113]  ERROR        data: {}
    code: "BAD_REQUEST"
    httpStatus: 400
    line: 24092
    column: 28
    sourceURL: "http://localhost:8081/index.bundle?platform=android&dev=true&minify=false"
```

gnprice:
> IIRC, server errors are translated, so the error will at least be in the language that the server admin has chosen for the server? Which is likely to correlate with the mobile app user's preferred language, I think.

I believe in fact it's better than that: for any logged-in requests they're translated into the user's preferred language, as set in their settings on the server (i.e. in/for the webapp.)

See the server code (and docstring) here:
https://github.com/zulip/zulip/blob/08212ef74a4703105b8e8ab563db76e5da68f91e/zerver/lib/exceptions.py#L57-L89

We actually introduced that abstraction, and the `code` property in the server's JSON responses, because the webapp had a bug where it wasn't properly handling the event queue going away if the user had chosen certain languages -- it was trying to interpret the text of the error message.

chrisbobbe:
Ah OK, great, then! Thanks for that detail! 🙂

PR #4550, opened 2021-03-19
Replacing connecting-spinner with indeterminate loading-bar,  Fixes: #4545

### LINKED ISSUE #4545: "Connecting" spinner pushes around content
The "Connecting" spinner at the top of the main screen pushes down all of the content below it. This results in things that the user might be trying to tap on jumping around, which is a poor experience. We should find a way of indicating that this is happening without pushing around the content.

This is a similar problem to #4526, although we could have a different solution. The constraints are pretty similar, though:

* We don't want any jumping of content that the user may try to click on
* We always want all of the content to be visible

I think a good solution to this might be a horizontal loading spinner (an "indefinite linear progress indicator" in [material design](https://material.io/components/progress-indicators) language) — see the gif at the top of that page for an idea of what I mean.

This doesn't completely solve the problem, because we'll still need to have that spinner disappear, but there are a few solutions there:

* We can have the linear progress indicator overlap the content, since it's not large enough to actually occlude the text (I don't think — we'd need to test this)
* We can animate the linear progress indicator away, which will spread out the few pixels of jump over time, which is probably good enough, and certainly better than the current behaviour.

This is a pretty popular style of loading indicator these days (the youtube and github webapps both have the "determinate" version of this), so i think it should be quite understandable.

vaibhavs2:
Fixed: #4545 
It prevents jumping of content (unread message cards), when loading ends.

LoadingBar has two types one as shown in image and one is regular type depends on decimal ∈ [0,1]


WesleyAC:
Thanks for doing this!

Two quick comments:

* Did you copy the `ProgressBar` code from another library? It's great to use other libraries, but we should try to include them in `package.json` so we can easily keep them up to date. This looks like a slightly edited version of [Bar.js from oblador/react-native-progress](https://github.com/oblador/react-native-progress/blob/master/Bar.js) — is there a reason you can't pull that file in from NPM instead?
* It's great that you included the `Fixes: ` in the commit, but that should probably go at the end of the message. The first line should be easily readable without knowing the issue number, maybe something like `HomeScreen: Replace "Connecting" spinner with indeterminate progress bar`?

vaibhavs2:
yes i copied that from library that you mentioned because I really didn't find any way to tackle the warning showing below

- And before that I thought we need only horizontal progressbar then why to add the whole module, also we already have circular loading etc.
- Please look at #4516 also how I got stuck with lot of failed tests , while writing this I realy didn't care about flow type or mentioned rules, so  this time tried without any warnings. I will be glad if you write something on the Idea of the PR #4516 and that's not copied from anywhere for sure.
- I learned how types works and helps :)

WesleyAC:
I started a [CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flow.20types.20on.20external.20libs) to talk about how to handle this, I expect someone who knows more about flow than I do will chime in there :)

vaibhavs2:
@WesleyAC 
Should I force push the changes I did based on the solution in [CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flow.20types.20on.20external.20libs) ?
I changed the connecting spinner with indeterminate progress-bar in StreamTabsScreen also.
see how its looking :)

WesleyAC:
Yes, go ahead and push!

vaibhavs2:
I made changes in  ChatNav and PMConversationScreen too, they were also containing connecting-Spinner and had same issues

flow-typed/npm/react-native-progress_vx.x.x.js
```diff
@@ -0,0 +1,87 @@
+// flow-typed signature: cda95545105831ec898fa23792e07ae9
+// flow-typed version: <<STUB>>/react-native-progress_v4.1.2/flow_v0.128.0
+
+/**
+ * This is an autogenerated libdef stub for:
+ *
+ *   'react-native-progress'
+ *
+ * Fill this stub out by replacing all the `any` types.
```
WesleyAC, line 9: Seems like we should maybe do this? Not sure how hard that is.

flow-typed/npm/react-native-progress_vx.x.x.js
```diff
@@ -0,0 +1,87 @@
...
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'react-native-progress' {
+  declare module.exports: any;
+}
+
+/**
+ * We include stubs for each file inside this npm package in case you need to
+ * require those files directly. Feel free to delete any files that aren't
```
WesleyAC, line 22: Please do delete all the files you don't use here.

src/common/progressBar.js
```diff
@@ -0,0 +1,68 @@
+/* @flow strict-local */
```
WesleyAC, line 1: You should delete this file, and switch to using the version in `package.json`

src/common/index.js
```diff
@@ -13,7 +13,7 @@ export { default as KeyboardAvoider } from './KeyboardAvoider';
 export { default as Label } from './Label';
 export { default as LineSeparator } from './LineSeparator';
 export { default as LoadingIndicator } from './LoadingIndicator';
-export { default as LoadingBanner } from './LoadingBanner';
+export { default as ProgressBar } from './progressBar';
```
WesleyAC, line 16: Once you're using the version from `package.json`, you won't need this line.

src/main/StreamTabsScreen.js
```diff
@@ -44,35 +49,38 @@ type Props = $ReadOnly<{|
...
-        options={{
-          tabBarLabel: ({ color }) => (
-            <Text style={[styles.tab, { color }]}>
-              <FormattedMessage id="All streams" defaultMessage="All streams" />
-            </Text>
-          ),
-        }}
-      />
-    </Tab.Navigator>
+    <SafeAreaView style={styles.container}>
```
WesleyAC, line 52: Can you include a screenshot of what this looks like before and after on iOS, so I can see what the effect `SafeViewArea` has on the layout?

WesleyAC (changes requested):
I'm confused about how the loading bar goes away with this code — can you explain what triggers that?

You should be able to use `git rebase -i origin/master` or similar to fix up the commit history.

flow-typed/npm/react-native-progress_vx.x.x.js
vaibhavs2: its the stubs required to tackle with flow warning as suggested in [CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flow.20types.20on.20external.20libs), well I removed some library from here that are not in use.

src/common/progressBar.js
vaibhavs2: I created a seperate file for progress-bar because we will have to use selector in every file(screen) to get loading status and override default props of Bar [see here](https://github.com/oblador/react-native-progress) as it has defalut `borderWidth=1, borderRadius=4, height=6, indeterminate` etc. and each time all these have to write again and again. According to user interface I used `borderWidth=0,radius=0, height=3` etc, you can see in common/progressBar.js I defined default props and overridring props as our need and also have to contain it in View to provide positional style, as Bar from react-native-progress doesn't have style props to accept. And the last, by doing all these we need to write only <ProgressBar /> in the screen and that's it. This is the same way as loadingBanner has implemented.

vaibhavs2: from loadingBanner i am referring to connecting-spinner

src/main/StreamTabsScreen.js
vaibhavs2: I can't include the screenshots because I dont have MacBook to compile and run  ( _if nyone looking at it please include screenshots I will be thankful to you_ 

previously the loadingBanner was in both seperate Tabs ie. /streams/SubscriptionsCard and /subscriptions/StreamListCard and both component includes in main/StreamTabsScreen as Tabs, so I removed loadingBanner from both and added single progressBar in parent screen(StreamTabsScreen) so I can show the loading status in the NavigationBar or say parallel to tab's Indicator as shown in image.

Again there need a **parent** to hold the tab's parent compontent(Tab.Navigator) and progressBar and for this I could use simple View or SafeAreaView, once look at the [link](https://stackoverflow.com/a/60557715/9929519) it's not got many upvotes but i agreed with the answer. As progressBar is placed out side of Tab.Navigator component so I used SafeAreaView well it doesn't have any effect because of progressBar `style={position:'absolute', top:NAVBAR_SIZE}` it will always lie at bottom of Navbar but to be in safer side I used it, it will act as simple View in android and IOS-version 10 or lower.

vaibhavs2:
> I'm confused about how the loading bar goes away with this code — can you explain what triggers that?


Within the file common/progressBar.js as image below


A pre-defined selector provides the loading status and makes progress bar go away, so visibility of progress is independent from the screens same as previously implemented you can see in the file common/loadingBanner.js

src/common/progressBar.js
```diff
@@ -0,0 +1,73 @@
...
+    tension?: number,
+    speed?: number,
+    bounciness?: number,
+  |},
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
+
+class ProgressBar extends PureComponent<Props> {
```
WesleyAC, line 39: Could you rename this file and class? It's not really a "progress" bar (since we're not measuring progress), I think "LoadingBar" would be better.

src/common/progressBar.js
```diff
@@ -0,0 +1,73 @@
...
+import { getLoading } from '../selectors';
+import { BRAND_COLOR, NAVBAR_SIZE } from '../styles';
+import { connect } from '../react-redux';
+import type { Dispatch } from '../types';
+
+type SelectorProps = $ReadOnly<{|
+  loading: boolean,
+|}>;
+
+type Props = $ReadOnly<{|
```
WesleyAC, line 15: Can you remove the props you aren't currently using here? That seems to be all of them except for `dispatch` and `SelectorProps`.

And then move the parts that you're just passing into `Bar` into the JSX part of the code.

src/common/progressBar.js
```diff
@@ -0,0 +1,73 @@
...
+    bounciness?: number,
+  |},
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
+
+class ProgressBar extends PureComponent<Props> {
+  static defaultProps = {
+    animated: true,
```
WesleyAC, line 41: `animated`, `progress`, and `animationConfig` are at the upstream defaults here, I think it'd be better to remove them.

src/main/StreamTabsScreen.js
WesleyAC: I don't have any apple devices either, so I can't test it, unfortunately.

But I think it makes more sense to just use a plain `View`, unless there's a specific layout problem that the `SafeAreaView` is solving. Feel free to start a thread in #mobile-team on CZO if you want more people's input on this, though, since I don't have much iOS experience.

This seems especially true given this bit from the [React Navigation docs](https://reactnavigation.org/docs/handling-safe-area/):

> While React Native exports a `SafeAreaView` component, it has some inherent issues, i.e. if a screen containing safe area is animating, it causes jumpy behavior.

WesleyAC (changes requested):
Looking much better! Just a few more changes :)

Also, can you delete `src/common/LoadingBanner.js` in this commit? Since we won't be using it anymore.

src/main/StreamTabsScreen.js
vaibhavs2: Lets work harder to get one then :D 
Now I think I should wrap them in View :)

src/common/progressBar.js
vaibhavs2: ok , It will be better, will do it

src/common/progressBar.js
vaibhavs2: not all of them , I defined them as props because maybe later we need progressbar or some change like we need smaller loading width than full screen, or to change color/background or loading-bar with more height(thick) so we can adjust bar with minimal props, even then if you think that needs to be removed then some can be eleminated like animated, indeterminateAnimationDuration,useNativeDriver and animationConfig I will remove these

vaibhavs2:
> Also, can you delete `src/common/LoadingBanner.js` in this commit? Since we won't be using it anymore.

No, once look at common/screen.js and this screen component used in many screens and connecting spinner looks pretty cool too! that's why I leave it.

src/common/progressBar.js
WesleyAC: If we need them later, we can add them later.

For a case like this, where this component should basically always look the same across the different places it's called, it's better not to provide a lot of props like this, since if we change them somewhere, we're probably going to want to change them everywhere.

src/common/loadingBar.js
```diff
@@ -0,0 +1,52 @@
...
+  static defaultProps = {
+    color: BRAND_COLOR,
+    unfilledColor: 'white',
+  };
+
+  render() {
+    const { loading, ...rests } = this.props;
+
+    return (
+      <View style={{ position: 'absolute', left: 0, right: 0, top: NAVBAR_SIZE - 3 }}>
```
WesleyAC, line 32: Can you make a local variable `const height = 3`, and use that here and to set the height below? that way we can avoid this magic number

src/common/loadingBar.js
```diff
@@ -0,0 +1,52 @@
...
+
+type Props = $ReadOnly<{|
+  color: string | number,
+  unfilledColor: string,
+
+  dispatch: Dispatch,
+  ...SelectorProps,
+|}>;
+
+class LoadingBar extends PureComponent<Props> {
```
WesleyAC, line 22: So, I probably should have realized this earlier, but I think we should keep the name `LoadingBanner`. That way, names like `shouldShowLoadingBanner`, which we have in various places right now will keep on making sense, and the diff for this file will show which bits of the code come from the old `LoadingBanner` file vs which ones are new.

WesleyAC (changes requested):
Sorry for the delay in looking at this! This is really close, just a couple more small changes around names :)

src/common/loadingBar.js
vaibhavs2: as you said :)
> Could you rename this file and class? It's not really a "progress" bar (since we're not measuring progress), I think "LoadingBar" would be better.

I would suggest to keep file name as LoadingBar, and  change  the `shouldShowLoadingBanner` to `shouldShowLoadingBar` everywhere.
wouldn't it be good?

WesleyAC: Yeah, that's probably a better name on the whole. The reason I suggested keeping it `LoadingBanner` is because that would make the diff nicer, but I do think `LoadingBar` describes it better. It'd be ideal for you to change `LoadingBanner` to use the `ProgressBar` code without renaming it in one commit, and then rename it (and the `shouldShowLoadingBanner`/etc) in a second commit.

vaibhavs2:
@WesleyAC 
Now it is ready for another review,
Instead of height i changed to bottomMargin, as height is NAVBAR_SIZE

vaibhavs2:
@WesleyAC 
Now it is ready for another review, i changed magic number :D to barHeight and it is using as bottom margin and height of bar also, 
:)

src/nav/ChatNavBar.js
```diff
@@ -56,7 +54,7 @@ export default function ChatNavBar(props: Props) {
         <ExtraButton color={color} narrow={narrow} />
         <InfoButton color={color} narrow={narrow} />
       </View>
-      <LoadingBanner spinnerColor={spinnerColor} backgroundColor={streamColor} textColor={color} />
+      <LoadingBanner unfilledColor={streamColor} color={color} />
```
WesleyAC: Can you post a gif of what this looks like, with a couple different colors of streams? It seems conceivable that this would look pretty weird.

One thing that's concerning is that `foregroundColorFromBackground` returns `'black' | 'white'`, but I think if this is white, the bar might be invisible?

My suspicion is that just having the bar always be `BRAND_COLOR` (by just not passing in a color) would probably be better, and we can deal with it looking bad when a stream happens to be close to `BRAND_COLOR`, but I'd like to see gifs first. Could you post one gif for each of `black`, `white`, and `BRAND_COLOR`?

WesleyAC (commented):
Looking good! Just one comment about the color of the bar in the stream view :)

src/nav/ChatNavBar.js
vaibhavs2: here they are
image | image |
| --- | --- |
|  |  |
| |  |

it is same as previously implemented and you already know that  ` <LoadingBanner unfilledColor={streamColor} color={color} />`, | 
accordig to what I have interpreted from review i think loading-bar always have `BRAND_COLOR` would not be a good idea in case of stream's NavBar, and the visuals of loading-bar always same as Navar and their texts, navBar color is the unfilled color of bar and text color will be the animating bar.
As in all other screen it is `BRAND_COLOR` but if we always apply `BRAND_COLOR`  in streams then we have a white horizontal line inside that the bar with `BRAND_COLOR` animate other wise it will be not visible with some colors.

when stream comes near to `BRAND_COLOR` their text also have color that can visible on navbar and the same way this loading-bar will behave.
I just caught, bar will be invisible if `streamColor =undefined` and `color='white'` see image below it would look something like

now we can have a condition  `if streamColor =undefined ` then `color=BRAND_COLOR` then BRAND_COLOR bar animating inside white horizontal view and be visible

vaibhavs2: From discussion on [CZO thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/loading.20bar.20colors) , come up with these, after adding 1px to height
| | |
|---|---|
|  |  |
|||

src/main/StreamTabsScreen.js
```diff
@@ -44,35 +48,38 @@ type Props = $ReadOnly<{|
...
+          options={{
+            tabBarLabel: ({ color }) => (
+              <Text style={[styles.tab, { color }]}>
+                <FormattedMessage id="All streams" defaultMessage="All streams" />
+              </Text>
+            ),
+          }}
+        />
+      </Tab.Navigator>
+      <LoadingBanner />
```
gnprice: It looks like this will put the loading bar/banner at the bottom of this view, instead of the top. Is that intended? Can you post a gif of what that looks like?

src/common/LoadingBanner.js
```diff
@@ -2,75 +2,56 @@
...
-        </View>
-        <Label
-          style={{
-            fontSize: 14,
-            margin: 2,
-            color: textColor,
-          }}
-          text="Connecting..."
-        />
+      <View style={{ position: 'absolute', left: 0, right: 0, top: NAVBAR_SIZE - barHeight }}>
```
gnprice: Can this say something like `bottom: 0`, instead of `top:` with arithmetic? That would help keep it more robust to being used in places that have a height different from `NAVBAR_SIZE`.

src/common/LoadingBanner.js
```diff
@@ -2,75 +2,56 @@
...
- * Display a notice that the app is connecting to the server, when appropriate.
+ * Display a indeterminate loading-bar while the app is connecting to the server, when appropriate.
  */
 class LoadingBanner extends PureComponent<Props> {
   static contextType = ThemeContext;
   context: ThemeData;
 
+  static defaultProps = {
+    color: BRAND_COLOR,
+    unfilledColor: null,
```
gnprice: nit: can simplify slightly by dropping this default, and saying `unfilledColor?: string` in the type.

After all, we're not really using `null` in particular here :slightly_smiling_face: -- all we do is hand it to `??` which tests if the value is either null or undefined.

gnprice (commented):
Thanks @vaibhavs2 and @WesleyAC ! A few comments below.

src/main/StreamTabsScreen.js
vaibhavs2: No it will be at top, as LoadingBar has `style={{position: 'absolute'}}`, and yes its intended, as it is the last component and has `position:absolute` it will be in most upper layer(It will have higher zIndex)

src/common/LoadingBanner.js
vaibhavs2: Before, I implented just as you said, but that time I didn't try night mode, after a number of reviews given by @WesleyAC  , I encountered with night mode but the way loadingBar was implented doesn't suit to night mode that thanks to @WesleyAC , If `unfilledColor` dropped from default then according to flow-type we have to pass unfilled color from every screen where we are using this loading-bar as of now Props is being passed from only one screen that is stream(ChatNav)

according to my logic, we are passing unfilled color props only from stream(ChatNav), if we are passing this unfilled color from every screen or if we don't have the night mode then type String with having some default white color would be good. But here I added null because we cant fetch `this.context.backgroundColor`(we need it when app turns to night mode) and assigned to unfilled color in default parenthesis. That's why I added null, and if any screen passes props then null will be replaced by props and if not then condition ` unfilledColor={unfilledColor ?? this.context.backgroundColor}`  will take the contex backgroundcolor based on nightMode or daymode, so here null is being used to decide `unfilledcolor`

src/common/LoadingBanner.js
vaibhavs2: yes right but, look into the file `main/StreamTabsScreen.js ` something i already said above and now if `bottom:0` will be used then It will definetly be at the bottom of screen, same as you said 

> It looks like this will put the loading bar/banner at the bottom of this view

In first push I did same as you are saying but because of the StreamTabsScreen.js I changed to the current we have, I assume you maybe think about put loadingBar in `StreamListCard` and `SubscriptionsCard` for that we need `top=0` and the reason I didn't implement like that is we will have two bars(one loadingBar below and other (in half screen) TabBar-Indicator above and that will look weired. To make them concurrent I used loading-bar in parent component of `StreamListCard` and `SubscriptionsCard` that is `main/StreamTabsScreen.js `

>That would help keep it more robust to being used in places that have a height different from NAVBAR_SIZE.

we can have props to accomplish that

src/main/StreamTabsScreen.js
gnprice: Hmm, I see.

Would you post a screencap showing what this looks like on that screen?

src/common/LoadingBanner.js
gnprice: Yeah, now that I see what's going on with the `position: 'absolute'`, it sounds like this is already very fragile. This is placing itself at a particular spot on the screen, which is dependent on the rest of what's on the screen around it looking a very particular way -- if there isn't some kind of bar at the top of height NAVBAR_SIZE, then this is going to look completely nonsense.

Maybe we end up doing that, but if so it's necessary that when you're looking at the code of any of the places where this is invoked, it's easy to tell that there's this unusual extra constraint that it makes on the layout of the rest of that screen. Otherwise as future changes happen to the app, we will end up with that kind of nonsense layout somewhere, with a loading indicator just plopping itself across the middle of some content in some other widget.

gnprice: Here's a way to address this that I think will give in the code a much better handle on how this layout works:

* Take the wrapper `View` with the `position: absolute`, and push it out to this component's callers. That way, the `position: absolute` is right there next to the source code of each surrounding `View`, the one that's being used to lay it out in each different context where we use it.
  * You can say something like `export const loadingBarHeight = 3` so that the callers can import that value, for those callers that need it.
  * This is enough to solve the really bad problem with the way this current version is using `position: 'absolute'`. We'll at least be able to see it's there, next to the code that controls each of the layouts it interacts with.
* Then, for I think probably all the places where we use this other than `StreamTabsScreen`, we should be able to push the `LoadingBar` inside the navbar, or whatever else it is that's expected to have height NAVBAR_SIZE.
  * That way, we avoid having a dependency on particular things *actually* being height NAVBAR_SIZE.
    * For example: in `HomeScreen`, is that row of icons at the top exactly NAVBAR_SIZE tall? Maybe it is, but it's sure not easy to tell that from its source code. Even if it is now, it might not be in the future, and it'd be hard to know when we're breaking that.
  * This will typically also mean that we can replace `top:` and arithmetic with `bottom: 0`.
* Then, for `StreamTabsScreen` and also if there are any other places where we end up still having a long-range dependency on the assumption that something is NAVBAR_SIZE tall, we make sure to add comments pointing out that dependency.

src/common/LoadingBanner.js
gnprice: > If `unfilledColor` dropped from default then according to flow-type we have to pass unfilled color

Did you try making the prop optional? That's this syntax that I mentioned:
```js
unfilledColor?: string,
```
with the `?` before the `:`. That's how you tell Flow that a prop (or a property of any object) should have a certain type if it's there, but also it's OK if it's not there at all.

I agree we don't want to be having to explicitly pass this in from each caller.

src/common/LoadingBanner.js
vaibhavs2: >kind of nonsense layout somewhere, with a loading indicator just plopping itself across the middle of some content in some other widget

:D :D
well , I changed loadingBarHeight to LOADING_BAR_THICKNESS, I think its more suitable

vaibhavs2:
@gnprice 
>Would you post a screencap showing what this looks like on that screen?

there it is!


one is [above](https://github.com/zulip/zulip-mobile/pull/4550#issuecomment-803074392) also

PR #4551, opened 2021-03-19
emoji: Fix some obsolete emoji names

WesleyAC:
This removes names like `:8:` for `:eight:`, since modern server
versions do not support these. It also changes `:8_ball:` to `:8ball:`,
since that's what modern servers expect.

We already had the new names, so there's no need for this commit to add
them.

This was all done manually based on errors seen in Sentry.

Related: #4546


WesleyAC:
Ok, this should be good to go now.

gnprice:
Merged, thanks!

Issue #4053, opened 2020-04-18
Huge amount of space left empty in top navigation bar


sumukhah:
![screenshot-2020-04-18_21 11 12 93](https://user-images.githubusercontent.com/23723464/79642203-c60e9a00-81b9-11ea-95a6-cfd733117e16.png)

sumukhah:
Working on the issue

shubhamkakkar:
@sumukhah can you describe this more ?
PS: steps to recreate this would also be helpful

sumukhah:
As you could see on the screenshot, more space left empty on top navigation bar(where all messages, mention, search are placed).
Use the "touchable" option that will be available along with the inspect elements to check this.
Ask this on the slack before diving fixing this.

abhi0504:
@gnprice I want to work on this issue can you plz assign it to me ?

chrisbobbe:
Hi, @abhi0504! Could you please describe your plan for your work on this issue? I don't see any alternatives to the current layout presented here, so the first job will be to think of some. 🙂

abhi0504:
@chrisbobbe thank you for your reply
i think we can increase the touchable opacity area for every icon such that it takes 0.25 out of 1 of the width in this way empty spaces in the top navigation bar will be filled up what's your opinion over this

chrisbobbe:
Sure, I think that sounds fine!

PR #4553, opened 2021-03-19
adjusting extra amount of space in top navigation bar

abhi0504:
using View with width 0.25*TotalWidthOfScreen such that it covers all the vacant space left in top navigation bar on the main screen

## Before 
Android Emulator - Pixel_3_API_29_5554 19-03-2021 16_32_21

## After

Android Emulator - Pixel_3_API_29_5554 19-03-2021 16_32_44

Fixes: #4053


abhi0504:
@chrisbobbe please have a look of this PR

src/main/HomeScreen.js
```diff
@@ -42,31 +44,39 @@ class HomeScreen extends PureComponent<Props> {
...
-          >
-            <IconUnreadMentions color={BRAND_COLOR} />
-          </NavButtonGeneral>
-          <NavButton
-            name="search"
-            onPress={() => {
-              NavigationService.dispatch(navigateToSearch());
-            }}
-          />
+          <View style={{ width: windowWidth * 0.25 }}>
```
chrisbobbe, line 47: Let's see if we can use [flexbox](https://reactnative.dev/docs/flexbox) styling here.

One reason is that I expect it'll be more concise, but another is more subtle: [`Dimensions.get`](https://reactnative.dev/docs/dimensions) doesn't hook the React component into changes in the dimensions and cause the `render` method to run again. It just tells you what the dimensions are when `render` happens to run. The dimensions will at least change when the device is rotated, and probably at other times.

chrisbobbe (commented):
Thanks, @abhi0504!

See one code comment, below.

Also, looking at the screenshot (thanks for including it! 🙂), I don't love the new visual layout in which the icons have drifted to the left:

image

I'd be happy to see the buttons' touchable areas expanded to fill the previously empty space in the row, but I'd prefer that each icon remains centered in its touchable area.

abhi0504:
@chrisbobbe thank you so much for your review 
so I have removed the `Dimensions.get` as you have mentioned above and also used the flexbox 
Now I have give every icon 25% of width via flexbox and also centralized the icons as well so for now there are no huge spaces around the icons 
Here are the current screenshots of this PR
Android Emulator - Pixel_3_API_29_5554 02-04-2021 17_51_29       Android Emulator - Pixel_3_API_29_5554 02-04-2021 17_50_21

Accepting the fact that the gap is still there but now the gap is divided on both of the sides so doesn't looks that bad

Issue #4539, opened 2021-03-16
Sharing from lightbox broken on Android


AkashDhiman:
Sharing from the image preview window is not implemented properly, it downloads the content it needs to share, and then supposedly passes the path of the downloaded content, Ideally it should pass the content URI.
The way it's done, leads to problems of wrong image being parsed by the receiving app.

https://user-images.githubusercontent.com/40268170/111384383-70938380-86cf-11eb-8a3a-35ce5a3b0a7a.mp4

In the given screen capture, I am sharing a file called: `cat.jpg`, it gets downloaded in my Downloads folder, where I have another (different) `cat.jpg` already present. This causes the downloaded file to be renamed as `cat (1).jpg` automatically, but the path that is shared to another app is still `Downloads/cat.jpg` which picks up my original cat.jpg.

Many other apps (slack, signal) fail to receive the shared content altogether.  
[discussion link](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Sharing.20from.20lightbox.20broken.20on.20Android)

PR #4554, opened 2021-03-20
lightbox: Refactor sharing module in android which is used in lightbox.

AkashDhiman:
This was required because sharing from lightbox screen was not
implemented properly:
- **Redundant Download**: It downloaded the image in the device's storage
before picking it up to share.
- **Faulty Sharing 1**: If a file with same name as the downloaded file was
already present in the downloaded directory, it would get picked up for
sharing instead of the file we downloaded.
- **Faulty Sharing 2**: Zulip was unable to share to some apps.
- **Hard Coded Implementation**: Sharing only worked for JPEG files properly.

This implementation uses `FileProvider` to create a cache directory for storing 
content to send, and in general improves the logic of the share from lightbox
screen functionality so that content of any type can be shared from it.

I also refactored the name of the module used for sharing, previously it was
called `ShareImageAndroid`. The new `shareContent` method I created is generic
and can be used for any type of content be it images, files, video etc, and hence I
changed the module name to `ShareContent`.

related links:
- https://developer.android.com/training/secure-file-sharing
- https://stackoverflow.com/questions/33222918/sharing-bitmap-via-android-intent/45563876

Logic for iOS sharing is unchanged.

Fixes: #4539


android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+            out.close();
+            Uri uri = FileProvider.getUriForFile(
+                getReactApplicationContext(),
+                "com.zulipmobile.fileprovider",
+                file
+            );
+
+            // invoke share intent
+            shareContentUri(uri, mimeType);
+        } catch (Exception e) {
```
WesleyAC, line 69: Can we also log this error so that it will get sent to sentry? I'm not sure off the top of my head how to do that in native code, but if you ask in #mobile-team on CZO someone will know.

android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+    public void shareContent(@Nullable ReadableMap source) {
+        try {
+            String urlString = source.getString("uri");
+            String[] urlSplit = urlString.split("/");
+            String fileName = urlSplit[urlSplit.length - 1];
+            URL url = new URL(urlString);
+            String auth = source.getMap("headers").getString("Authorization");
+            HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+            conn.addRequestProperty("Authorization", auth);
+            String mimeType = conn.getContentType();
```
WesleyAC, line 50: Do we want to limit the possible mimetypes we'll share? Possibly it makes sense to just share whatever we get, I don't know if there are security implications there or anything.

I think probably it makes sense to leave this as-is, I just wanted to bring it up in case anyone knows more about this than I do and has a reason to restrict mimetypes.

android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+            InputStream in = conn.getInputStream();
+
+            // save content in cache directory
+            File contentsFolder = new File(getReactApplicationContext().getCacheDir(), "contents");
+            contentsFolder.mkdir();
+            File file = new File(contentsFolder, fileName);
+            FileOutputStream out = new FileOutputStream(file);
+            IOUtils.copy(in, out);
+            out.flush();
```
WesleyAC, line 59: I don't think this `flush()` call does anything — see [here](https://stackoverflow.com/questions/31456660/fileoutputstream-does-the-close-method-calls-also-flush)

android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+        super(reactContext);
+    }
+
+    @Override
+    public String getName() {
+        return "ShareContent";
+    }
+
+    @ReactMethod
+    public void shareContent(@Nullable ReadableMap source) {
```
WesleyAC, line 41: Can you change this to explicitly take a `url` and `auth` with the correct types? That will make this function significantly more type-safe.

WesleyAC (changes requested):
Thanks for working on this! I left a handful of comments.

Also, I appreciate that your commit message is very detailed and follows the style guide, thanks for that :)

android/app/src/main/java/com/zulipmobile/ShareContent.java
WesleyAC: I started a CZO thread for this [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/logging.20to.20sentry.20from.20native.20android.20code) :)

android/app/src/main/java/com/zulipmobile/ShareContent.java
WesleyAC: I started a CZO thread for this [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/logging.20to.20sentry.20from.20native.20android.20code) :)

android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Allows Sending Content from zulip to other apps.
+ */
+
+public class ShareContent extends ReactContextBaseJavaModule {
+    private String TAG = "ShareContent";
```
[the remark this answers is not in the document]
AkashDhiman: Removed the variable.

android/app/src/main/java/com/zulipmobile/ShareContent.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Allows Sending Content from zulip to other apps.
+ */
+
+public class ShareContent extends ReactContextBaseJavaModule {
```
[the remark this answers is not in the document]
AkashDhiman: Refactored the name to `ShareFile`.

android/app/src/main/java/com/zulipmobile/ShareContent.java
AkashDhiman: I've introduced a call to `Sentry.captureException` in the catch block. As far as I could tell by reading the [documentation](https://docs.sentry.io/platforms/android/), this should work, since when we call `Sentry.init` in JavaScript it should also call the corresponding `Sentry.init` in android.

android/app/src/main/java/com/zulipmobile/ShareContent.java
AkashDhiman: removed the call to `flush`.

android/app/src/main/java/com/zulipmobile/ShareContent.java
AkashDhiman: I have changed the signature now it is `shareFile(@Nullable String authHeader, String resourceUrl)`.

AkashDhiman:
Issues addressed after first review by @WesleyAC:

- renamed shareContent to shareFile.
- removed unused variable.
- modified method shareFile (previously shareContent) to take auth, resource-url as argument.
- removed flush method call.
- wrote code to send Exception to Sentry. (not sure if it works as intended, couldn't test)

Concern raised by @WesleyAC:

> Do we want to limit the possible mimetypes we'll share? Possibly it makes sense to just share whatever we get, I don't know if there are security implications there or anything.

> I think probably it makes sense to leave this as-is, I just wanted to bring it up in case anyone knows more about this than I do and has a reason to restrict mimetypes.

This PR is ready for re-review @WesleyAC.

android/app/src/main/java/com/zulipmobile/ShareContent.java
AkashDhiman: I just tested this out by using my own sentry key, and it works fine.

src/lightbox/shareFile.js
```diff
@@ -0,0 +1,41 @@
+/* @flow strict-local */
```
WesleyAC, line 1: I think this file should still be `shareImage` rather than `shareFile`, since it's lightbox-specific — the lightbox only shows images, not other types of files.

That'll also make the diff a bit nicer for me to look at :)

WesleyAC (changes requested):
Thanks! Just one more little change :)

src/lightbox/shareFile.js
AkashDhiman: Right makes sense, refactored accordingly.

AkashDhiman:
@WesleyAC it's ready for another round of review, thanks!

WesleyAC:
Hmmm, we should actually probably call `shareFile.js` and `shareFile.java` `shareFileAndroid` instead, since that's what they were named previously. Apologies again for all of the back and forth on names here, I should've thought of that earlier :sweat_smile:

AkashDhiman:
> Hmmm, we should actually probably call `shareFile.js` and `shareFile.java` `shareFileAndroid` instead, since they had the `Android` suffix previously. Apologies again for all of the back and forth on names here, I should've thought of that earlier :sweat_smile:

My thought process was that in future we may make a shareFile module for iOS as well and hence the name should be common, but then again that change should be done when we do eventually make a shareFileModule for iOS. Will do the refactoring ASAP.

AkashDhiman:
changed file names as suggested. @WesleyAC you may review it again.

WesleyAC (approved):

chrisbobbe (changes requested):
Thanks for catching these issues and working on this, @AkashDhiman! Particularly concerning is that the app could end up accidentally sharing a `cat.jpg` that the user has no intention of sharing at all; I'll be glad to have that privacy issue fixed.

I ran into a few issues while testing this on Android:

-----

`tools/run-android` failed with the following message:

```
> com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: INSTALL_FAILED_CONFLICTING_PROVIDER: Package couldn't be installed in /data/app/com.zulipmobile.debug-b13SHlXJ34jXQFRJSUlYyQ==: Can't install because provider name com.zulipmobile.fileprovider (in package com.zulipmobile.debug) is already used by com.zulipmobile
```

Have you encountered a message like that?

My understanding from a bit of Googling is that, in the added code in `android/app/src/main/AndroidManifest.xml`, we need to set a different value for `android:authorities` for debug vs. release builds. I'm not certain that I have the full picture, though; as far as I know, I've only tried to do a debug build on my device, and not a release build, so I'm not totally sure why I encountered the error. But the error did go away when I did the following, to make `android:authorities` different for debug vs. release. Does this sound reasonable?

- Set `android:authorities="@strings/file_provider_authority"`, where "file_provider_authority" is the "name" of a string that will differ between debug and release. (We could choose a different name but this seemed right to me.)
- Set the string "com.zulipmobile.debug.fileprovider" in android/app/src/debug/res/values/strings.xml, and give it the name "file_provider_authority"
- Set the string "com.zulipmobile.fileprovider" in android/app/src/main/res/values/strings.xml, and give it the name "file_provider_authority"

This solution was based on [this blog post](https://gradlewhy.ghost.io/overcoming-install-failed-conflicting-provider/). If we need to commit a solution (such as this one) to the INSTALL_FAILED_CONFLICTING_PROVIDER issue, and it's not something weird with my own environment, let's aim to do so in a separate prep commit(s), before the more interesting commits.

-----

I got the "Sharing Failed." message just after tapping the "Share image" button in the lightbox's action sheet, with nothing in between (it didn't ask me where to share it).

I added a line to log `e` in the `catch` block, and I saw it go by in `adb logcat` when I pressed the "Share image" button again:

```
03-23 14:01:24.854 30355 30534 W MainActivity: java.lang.IllegalArgumentException: Failed to find configured root that contains /data/data/com.zulipmobile.debug/cache/files/IMG_20190808_153140.JPG
03-23 14:01:24.854 30355 30534 W MainActivity: 	at androidx.core.content.FileProvider$SimplePathStrategy.getUriForFile(FileProvider.java:800)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at androidx.core.content.FileProvider.getUriForFile(FileProvider.java:442)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.zulipmobile.ShareFileAndroid.shareFile(ShareFileAndroid.java:61)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at java.lang.reflect.Method.invoke(Native Method)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:151)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at android.os.Handler.handleCallback(Handler.java:873)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at android.os.Handler.dispatchMessage(Handler.java:99)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at android.os.Looper.loop(Looper.java:214)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
03-23 14:01:24.854 30355 30534 W MainActivity: 	at java.lang.Thread.run(Thread.java:764)
```

I haven't yet dug deeper to debug that; did you run into that error, and do you know what might be causing it?

-----

See also some comments below.

android/app/src/main/AndroidManifest.xml
```diff
@@ -72,5 +72,14 @@
         </service>
 
         <activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />
+        <provider
+        android:name="androidx.core.content.FileProvider"
```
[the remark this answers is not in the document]
AkashDhiman: there were pre-existing issues, made an NFC commit as suggested.

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.java
```diff
@@ -0,0 +1,92 @@
...
+            Sentry.captureException(e);
+        }
+    }
+
+    private void shareUri(Uri uri, String mimeType) {
+        Intent shareIntent = new Intent();
+        shareIntent.setAction(Intent.ACTION_SEND);
+        shareIntent.putExtra(Intent.EXTRA_STREAM, uri);
+        shareIntent.setType(mimeType);
+
```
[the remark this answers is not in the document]
AkashDhiman: ran through auto format, it's fixed now.

src/lightbox/LightboxActionSheet.js
```diff
@@ -2,7 +2,7 @@
 import type { Auth } from '../types';
 import downloadImage from './downloadImage';
 import share from './share';
-import shareImage from './shareImage';
+import { shareImage } from './shareImage';
```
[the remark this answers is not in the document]
AkashDhiman: this was unintentional, reverted back to default export.

src/lightbox/shareImage.js
```diff
@@ -3,26 +3,31 @@ import { Platform, Share } from 'react-native';
...
+    const tempUrl = await api.tryGetFileTemporaryUrl(url, auth);
+
+    if (tempUrl === null) {
+      showToast('Please share the image from your browser');
+      openLink(new URL(url, auth.realm).toString());
+      return;
+    }
+
     try {
+      const fileName = url.split('/').pop();
```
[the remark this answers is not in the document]
AkashDhiman: moved it outside of the try catch block.

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public String getName() {
+        return "ShareFileAndroid";
+    }
+
+    @ReactMethod
+    public void shareFile(@Nullable String authHeader, String resourceUrl) {
+        try {
+            String[] urlSplit = resourceUrl.split("/");
+            String fileName = urlSplit[urlSplit.length - 1];
```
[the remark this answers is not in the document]
AkashDhiman: now I am using `String fileName = FilenameUtils.getName(url.getPath());` from package `org.apache.commons.io.FilenameUtils`

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.java
```diff
@@ -0,0 +1,92 @@
...
+                "com.zulipmobile.fileprovider",
+                file
+            );
+
+            // invoke share intent
+            shareUri(uri, mimeType);
+        } catch (Exception e) {
+            Toast.makeText(
+                getReactApplicationContext(),
+                "Sharing Failed.",
```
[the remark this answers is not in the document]
AkashDhiman: made the method resolve a promise and now I am handling it inside JavaScript.

src/lightbox/shareImage.js
```diff
@@ -3,26 +3,31 @@ import { Platform, Share } from 'react-native';
 
 import downloadImage from './downloadImage';
 import type { Auth } from '../types';
-import ShareImageAndroid from '../nativeModules/ShareImageAndroid';
+import ShareFile from '../nativeModules/ShareFileAndroid';
```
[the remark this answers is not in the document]
AkashDhiman: refactored to `import ShareFileAndroid from '../nativeModules/ShareFileAndroid'`.

src/lightbox/shareImage.js
```diff
@@ -3,26 +3,31 @@ import { Platform, Share } from 'react-native';
...
 import { showToast } from '../utils/info';
 import * as api from '../api';
 import openLink from '../utils/openLink';
+import { getResource } from '../utils/url';
 
-export default async (url: string, auth: Auth) => {
-  const tempUrl = await api.tryGetFileTemporaryUrl(url, auth);
```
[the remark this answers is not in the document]
AkashDhiman: I thought it was an extra step, why bother creating a temp URL when you can manage to fetch the resource regardless, I didn't see the history of the file to understand why it was introduced and assumed the initial implementation worked that way. I wasn't familiar with looking into history of a file like that and will surely use it a lot now, thanks for that!
If you would like I will revert it back to use temp URL, just give me another cue. But I still have my doubts on this, wasn't this problem in the backend rather than the fetch code? @chrisbobbe If possible, can you clarify where it won't work with directly using auth & url, where it would work using temp url.

AkashDhiman:
regards to the following error:
```
com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: INSTALL_FAILED_CONFLICTING_PROVIDER: Package couldn't be installed in /data/app/com.zulipmobile.debug-b13SHlXJ34jXQFRJSUlYyQ==: Can't install because provider name com.zulipmobile.fileprovider (in package com.zulipmobile.debug) is already used by com.zulipmobile
```
I did encounter this while installing from `./tools/run-android` on a device that already had the playstore version of zulip, I suspected it was because I had two versions of zulip installed in the same device, and so tried to build again after uninstalling the play store version and the build succeeded after that. Suspecting that it was nothing more than that (and users are always going to have single app on there device)I  didn't dive deeper into this issue. Did you happen to have another zulip already installed as well? I will try to look into this and let you know if I find something interesting to share.

AkashDhiman:
> I got the "Sharing Failed." message just after tapping the "Share image" button in the lightbox's action sheet, with nothing in between (it didn't ask me where to share it).

looking at your error logs it seems that it was not able to create the `cacheDir`, This is a wild guess but perhaps when you made modifications with relation to `android:authorities` you may not have configured it properly, there is a hard coded value of referencing fileprovider in line 62 of `android/app/src/main/java/com/zulipmobile/ShareFileAndroid.java` that might also require changing.
I will try to do the modifications you suggested and see if I get similar behavior.

gnprice:
> I did encounter this while installing from `./tools/run-android` on a device that already had the playstore version of zulip, I suspected it was because I had two versions of zulip installed in the same device, and so tried to build again after uninstalling the play store version and the build succeeded after that.

Yeah, the reason we give a different Android package name to the debug vs. release build (`com.zulipmobile.debug` vs `com.zulipmobile`) is because that makes it possible to have both installed at the same time. That's very convenient for development, because it means you can have the published release build installed on your phone for normal use and then also a debug version which you're frequently replacing with your changes. (You can't simply upgrade the published version to a debug version and back again, because they're signed with different keys and so Android won't allow that without completely uninstalling the app and deleting its data, as a security/privacy measure.)

So let's be sure to keep it the case that you can have both of them installed at the same time. The solution @chrisbobbe described above sounds like a good one to me.

gnprice (commented):
Thanks @AkashDhiman ! Will definitely be glad to have this fixed. Here's just some high-level comments.

First, here's some useful background about our Android native code:
https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/android.md#kotlin-and-java
In particular, the normal way we'd approach a change like this is we'd first convert the file `ShareImageAndroid.java` to Kotlin, and then write the new code in Kotlin. Kotlin is a pretty great improvement over Java, and it's remarkably smooth to move to if you've been writing Java -- which together are why Google has been investing heavily in making it the new preferred language for Android.

Because this is your first contribution, though (and because this isn't a huge amount of Java code), we can do things in the other order here: I'll be happy to see a version of this PR merged that's in Java, and we can convert it to Kotlin as a followup.

---

Second, about ways to structure this. In this PR's new version of this code, there's basically two steps:
* We download the file.
* Then we share it.

(I've marked out the boundary between those in comments below.)

We have the same two steps in the current code: first we download the file, then we share it.

I think it would be helpful to keep those two steps separate. Specifically, let's keep them as two separate calls that the JS code makes into the native Java/Kotlin code. There are a few reasons I think that will be helpful:
* Generally we try to keep our JS code "in charge" of directing the flow of things, and the native Kotlin/Java code (or Swift/Obj-C code, on iOS) limited to the things that have to be done there because they're interacting with the platform APIs.
* It'll help keep it the case that you can almost always focus on the JS code and see what's happening, which is one of the reasons we do that in general.
* It'll help keep the code the same for the Android and iOS cases as much as possible, which is another reason we do that in general.
  * Here, it's OK if the code in e.g. `shareImage.js` becomes a bit more different between iOS and Android than it is now. But it'll still be helpful if the code in the two cases is at least pretty similar.
* Another reason, more specific to this change, is that I think it'll help isolate and make clearer what the important things are that this PR is changing.

Expanding on the last point, I'd like to understand better the reason why this switches away from using `rn-fetch-blob`. We're downloading the file to a different location -- to the app-specific cache directory, instead of the user's cross-app downloads directory. But in our `downloadImage`, we tell `rn-fetch-blob` what path to download to -- we could make the same change there. Have you tried doing it that way? What goes wrong?

If there is a good reason why `rn-fetch-blob` won't work, or isn't as good as some code we can write ourselves, that's perfectly fine -- we should just be sure the reason is clearly explained.

Similarly, I'm interested in the question about `tryGetFileTemporaryUrl` that @chrisbobbe explained at https://github.com/zulip/zulip-mobile/pull/4554#discussion_r599875119 .

AkashDhiman:
> Expanding on the last point, I'd like to understand better the reason why this switches away from using rn-fetch-blob

I overlooked the fact that I could use `rn-fetch-blob` here, apologies for that, I just tested the workflow using `rn-fetch-blob` and its much simpler this way. so I have pushed the new implementation. (Turns out under the hood `rn-fetch-blob` also uses `FileProvider`)
Since I was basically starting from scratch again I took the opportunity to convert the module to Kotlin as @gnprice suggested.
I also made the newer implementation using `tryGetFileTemporaryUrl` as well as separating the sharing part from the downloading part.
as for the explanation as to why I didn't use `tryGetFileTemporaryUrl` I failed to understand its purpose initially since I assumed it was there from the beginning and not introduced to avoid a bug, although I am still a bit confused with its use, Is it only there to terminate the function early so that `rn-fetch-blob` does not get called when the link gets redirected to S3?

> Second, about ways to structure this. In this PR's new version of this code, there's basically two steps:

My primary reason to do sharing in one step was that I wanted to make the sharing process as a single entity. I didn't suppose we would want to download a file to cache directory individually and that the downloading step was just a subpart of sharing hence it should stay in the same module. In fact I tried to figure out a way to not even have to download it, Since we technically already have the image present in the app (inside lightbox imageView), The way I would have done this is by extracting the bitmap of the image, saving it in cache dir and then sending its content uri. But since react native is modular it's hard for me to do what I described unless I modify the PhotoView module.

@gnprice @chrisbobbe This PR is ready for another review

src/lightbox/shareImage.js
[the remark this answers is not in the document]
chrisbobbe: > If possible, can you clarify where it fails to work when using just auth & url but works using temp url.

Zulip Cloud realms on zulipchat.com use the s3 backend to store files; chat.zulip.org doesn't. So a good first step, if you'd like to investigate whether ed645fe91 went in the wrong direction, would be to experiment on a realm there. (In addition to looking for clues in its PR, #4089, and links from there.)

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.kt
```diff
@@ -36,6 +37,7 @@ class ShareFileAndroid(reactContext: ReactApplicationContext?) : ReactContextBas
         reactApplicationContext.startActivity(chooserIntent)
         promise.resolve("success")
     } catch (e: Exception) {
+        Sentry.captureException(e)
```
chrisbobbe, line 40: Nit: Let's have the JavaScript be "in charge" of reporting this error, in the pattern Greg mentioned at https://github.com/zulip/zulip-mobile/pull/4554#pullrequestreview-620338958 of having the JavaScript do what it can.

A `logging.error` call, passing the error, should be sufficient; you can search the project for other examples of using `logging.error`. 🙂

chrisbobbe (changes requested):
Thanks, @AkashDhiman! I've just tried this out on Android, and it seems to be working correctly; I was able to share a PNG image ([this one](https://chat.zulip.org/#narrow/stream/7-test-here/topic/Chris/near/1149302)) to the Gmail app after these commits, and I wasn't able to do so before these commits.

I noticed one nit, below, and I expect @gnprice would like to take another pass, as his previous review consisted of "just some high-level comments". 🙂

I also responded to the point about using a temp URL, at https://github.com/zulip/zulip-mobile/pull/4554#discussion_r605333447.

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.kt
AkashDhiman: done.

src/lightbox/downloadFileToCache.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Download a remote file to the app's cache directory.
+ *
+ * @param tempUrl A URL to the file. Should be a valid temporary URL generated
+ *     using `getFileTemporaryUrl`.
+ * @param fileName Name of the file to be downloaded. Should include the
+ *     extension.
+ */
+
+export default async (tempUrl: string, fileName: string): Promise<mixed> =>
```
gnprice, line 14: nit: no blank line between jsdoc and the thing it's documenting -- that helps keep it visually clear what it's attached to

src/lightbox/downloadFileToCache.js
```diff
@@ -0,0 +1,21 @@
...
+  RNFetchBlob.config({
+    path: `${RNFetchBlob.fs.dirs.CacheDir}/${fileName}`,
+    fileCache: true,
+    useDownloadManager: true,
+    mime: getMimeTypeFromFileExtension(fileName.split('.').pop()),
+    title: fileName,
+  }).fetch('GET', tempUrl);
```
gnprice, line 21: This is quite similar to the code in the Android case of `downloadImage`. That's fine -- it's more important to have it working correctly than to share code, and when I look closely I see that it actually wouldn't be simple to have the two versions share much code in a useful way -- but let's get the two versions right next to each other in the code. That way it's easy to compare them, and also when we're changing one it'll be easy to notice the other and consider whether we should be changing it the same way.

I think the best way to do that probably looks like:
* First, have a prep commit that renames `downloadImage.js` to like `download.js`, and turns the existing default-export into an export named `downloadImage`.
* Then, the commit that adds this version can just add another function to that file, with a name like `downloadFileToCache`.

gnprice (commented):
Thanks @AkashDhiman! This looks great. A few comments below, which I think will all be straightforward to fix.

> In fact I tried to figure out a way to not even have to download it, Since we technically already have the image present in the app (inside lightbox imageView), The way I would have done this is by extracting the bitmap of the image, saving it in cache dir and then sending its content uri. But since react native is modular it's hard for me to do what I described unless I modify the PhotoView module.

Yeah. I'm not even sure there's a straightforward way to do this if modifying that library -- if you take a look in node_modules/react-native-photo-view/**/PhotoView.java, it seems to use something from `com.facebook` (so probably pulled in as part of React Native itself) to actually show the image, and that thing takes the URL and handles the request for itself. So it'd be a matter of getting the image back out from that Facebook library.

We wouldn't so much want the raw bitmap -- the original file is likely to be compressed, and we want to keep that compression rather than trying to share an uncompressed version (which may be giant) or a newly re-compressed version (which is likely to end up lower in quality.)

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.kt
```diff
@@ -0,0 +1,41 @@
...
+        val shareIntent = Intent()
+        shareIntent.action = Intent.ACTION_SEND
+        shareIntent.type = reactApplicationContext.contentResolver.getType(sharedFileUri)
+        shareIntent.putExtra(Intent.EXTRA_STREAM, sharedFileUri)
+
+        val chooserIntent = Intent.createChooser(shareIntent,
+            reactApplicationContext.resources.getText(R.string.send_to))
+        chooserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        reactApplicationContext.startActivity(chooserIntent)
+        promise.resolve("success")
```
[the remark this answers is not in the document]
AkashDhiman: I changed it to `promise.resolve(null)` in the latest revision, at JS side it resolves to null, which would make the type of function - `Promise<null>`. I couldn't find a way to resolve `undefined` explicitly from Kotlin since `Promise.resolve` always requires a parameter, if there is a way to do this let me know and I will make the necessary changes.

AkashDhiman:
@gnprice this PR is ready for re-review.

android/app/src/main/java/com/zulipmobile/ShareFileAndroid.kt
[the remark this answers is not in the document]
gnprice: Sounds good, thanks! `Promise<null>` should be clear enough.

gnprice:
Thanks! Merged :tada: 

I made a couple of small changes in merging:
* Added a commit on top rearranging the new comment in `ShareFileAndroid.kt` a bit.
* Adjusted the summary line in the main bugfix commit: it'd said "Refactor", but in my vocabulary a refactor is a change that's mainly about how the code looks or works on the inside, and either completely avoids altering how it behaves on the outside or only does so as a minor side effect. Whereas in this commit, the main point was the behavior changes (the bugfix), so that's the part to highlight in the summary line.

PR #4561, opened 2021-03-23
webview: Mark messages as read when bottom comes into view

WesleyAC:
Previously, we marked messages as read as soon as 20px of the message
was visible. However, this is a poor experience: messages are marked as
read before the user has seen the entire message. It's also different
from what the webapp does: in the webapp, it's impossible to mark a
message as read by scrolling until the bottom is in view.

We can't just change `isVisible` to do this, since that's used in
someVisibleMessage, which is used to find a message for resetting the
scroll position - my understanding is that that assumes that if there
are any messages loaded, at least one will be visible. However, it's
possible for a single large message to be the only thing that's visible,
but we don't want to mark it as read, since the bottom hasn't come into
view yet. Hence, this commit adds a `isRead` which is used in
conjunction with `isVisible` to choose when to mark messages as read.

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -307,12 +307,51 @@ function lastMessage(): ?Element {
   return walkToMessage(documentBody.lastElementChild, 'previousElementSibling');
 }
 
-/** The minimum height (in px) to see of a message to call it visible. */
-const minOverlap = 20;
+/** The message before the given message, if any. */
+function previousMessage(start: Element): ?Element {
+  return walkToMessage(start.previousElementSibling, 'previousElementSibling');
+}
 
+/**
+ * An element is visible if any part of it is visible on screen.
+ *
+ * @param top The top of the screen (typically 0)
+ * @param bottom The bottom of the screen (typically body.clientHeight)
+ */
 function isVisible(element: Element, top: number, bottom: number): boolean {
   const rect = element.getBoundingClientRect();
-  return top + minOverlap < rect.bottom && rect.top + minOverlap < bottom;
+  return top < rect.bottom && rect.top < bottom;
+}
+
+/**
+ * The number of pixels of the message that are allowed to be scrolled off
+ * the bottom of the screen when we mark the message as "read".
+ */
+// Picked based on the bottom padding of messages in CSS. This doesn't work
+// so great in cases where messages have the "edited" badge or reaction emojis,
+// since the user needs to scorll to the bottom of the reaction emoji/edited
+// badge section for the message to be "read", but it's more important for the
+// single-line message case to not get inadvertently read than it is for
+// messages to get marked read exactly when the bottom of the text of the
+// message comes onto the screen. In the future, we may want more complicated
+// behaviour here, but this should be fine for now.
+const messageReadSlop = 16;
+
+/**
+ * A visible message is read when its bottom isn't too far down out of view.
+ *
+ * See `messageReadSlop` for specifically how far the bottom might be. The
+ * idea is that a message becomes read when the bottom of its content
+ * scrolls into view, excluding the padding between messages.
+ *
+ * This function doesn't check that the message is visible; it probably
+ * makes sense to call only when `isVisible` is already known to be true.
+ *
+ * @param top The top of the screen (typically 0)
+ * @param bottom The bottom of the screen (typically body.clientHeight)
+ */
+function isRead(element: Element, top: number, bottom: number): boolean {
+  return bottom + messageReadSlop >= element.getBoundingClientRect().bottom;
 }
 
 /** Returns some message element which is visible, if any. */
@@ -333,6 +372,25 @@ function someVisibleMessage(top: number, bottom: number): ?Element {
   );
 }
 
+/** Returns some message element which is both visible and read, if any. */
+function someVisibleReadMessage(top: number, bottom: number): ?Element {
+  function checkReadAndVisible(candidate: ?Element): ?Element {
+    return candidate && isRead(candidate, top, bottom) && isVisible(candidate, top, bottom)
+      ? candidate
+      : null;
+  }
+
+  // Algorithm: If there's a visible message that isn't read, that means
+  // it's partway off the bottom of the screen. Therefore, either:
+  // * the message above it will be visible and read, or
+  // * there are no visible read messages.
+  const visible = someVisibleMessage(top, bottom);
+  if (!visible) {
+    return visible;
+  }
+  return checkReadAndVisible(visible) || checkReadAndVisible(previousMessage(visible));
+}
+
 /**
  * The Zulip message ID of the given message element; throw if not a message.
  */
@@ -345,23 +403,23 @@ function idFromMessage(element: Element): number {
 }
 
 /**
- * Returns the IDs of the first and last visible messages, if any.
+ * Returns the IDs of the first and last visible read messages, if any.
  *
- * If no messages are visible, the return value has first > last.
+ * If no messages are both visible and read, the return value has first > last.
  */
-function visibleMessageIds(): {| first: number, last: number |} {
-  // Algorithm: We find some message that's visible; then walk both up and
-  // down from there to find all the visible messages.
+function visibleReadMessageIds(): {| first: number, last: number |} {
+  // Algorithm: We find some message that's both visible and read; then walk
+  // both up and down from there to find all the visible read messages.
 
   const top = 0;
   const bottom = viewportHeight;
   let first = Number.MAX_SAFE_INTEGER;
   let last = 0;
 
-  // Walk through visible elements, observing message IDs.
+  // Walk through visible-and-read elements, observing message IDs.
   function walkElements(start: ?Element, step: 'nextElementSibling' | 'previousElementSibling') {
     let element = start;
-    while (element && isVisible(element, top, bottom)) {
+    while (element && isVisible(element, top, bottom) && isRead(element, top, bottom)) {
       if (element.classList.contains('message')) {
         const id = idFromMessage(element);
         first = Math.min(first, id);
@@ -372,7 +430,7 @@ function visibleMessageIds(): {| first: number, last: number |} {
     }
   }
 
-  const start = someVisibleMessage(top, bottom);
+  const start = someVisibleReadMessage(top, bottom);
   walkElements(start, 'nextElementSibling');
   walkElements(start, 'previousElementSibling');
 
@@ -419,11 +477,14 @@ const setMessagesReadAttributes = rangeHull => {
  *
  */
 
-/** The range of message IDs visible whenever we last checked. */
-let prevMessageRange = visibleMessageIds();
+/**
+ * The range of message IDs that were both visible and read whenever we last
+ * checked.
+ */
+let prevMessageRange = visibleReadMessageIds();
 
 const sendScrollMessage = () => {
-  const messageRange = visibleMessageIds();
+  const messageRange = visibleReadMessageIds();
   // rangeHull is the convex hull of the previous range and the new one.
   // When the user is actively scrolling, the browser gives us scroll events
   // only occasionally, so we use this to interpolate scrolling past the
@@ -442,7 +503,16 @@ const sendScrollMessage = () => {
     endMessageId: rangeHull.last,
   });
   setMessagesReadAttributes(rangeHull);
-  prevMessageRange = messageRange;
+  // If there are no visible + read messages (for instance, the entire screen
+  // is taken up by a single large message), then we don't want to update
+  // prevMessageRange.  This way, if the user scrolled past some messages to
+  // get here, then even though `messageRange` was empty this time and so we
+  // didn't mark any messages as read just now, we'll include those in
+  // `rangeHull` the next time the user scrolls and so we'll mark them as read
+  // then.
+  if (messageRange.first < messageRange.last) {
+    prevMessageRange = messageRange;
+  }
 };
 
 // If the message list is too short to scroll, fake a scroll event
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -515,11 +515,19 @@ var compiledWebviewJs = (function (exports) {
     return walkToMessage(documentBody.lastElementChild, 'previousElementSibling');
   }
 
-  var minOverlap = 20;
+  function previousMessage(start) {
+    return walkToMessage(start.previousElementSibling, 'previousElementSibling');
+  }
 
   function isVisible(element, top, bottom) {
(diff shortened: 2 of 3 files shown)


src/webview/js/js.js
```diff
@@ -295,12 +295,23 @@ function lastMessage(): ?Element {
...
+// based on the bottom padding of messages in CSS. This doesn't work so great
+// in cases where messages have the "edited" badge or reaction emojis, but it's
+// more important for the single-line message case to not get inadvertently read
+// than it is for messages to get marked read exactly when the bottom of the
+// text of the message comes onto the screen. In the future, we may want more
+// complicated behaviour here, but this should be fine for now.
+const messageReadSlop = 16;
+
+function isRead(element: Element, top: number, bottom: number): boolean {
+  return bottom + messageReadSlop >= element.getBoundingClientRect().bottom;
```
gnprice, line 314: IIUC this function's jsdoc might look like:
```js
/**
 * A visible message is read when its bottom isn't too far down out of view.
 *
 * See `messageReadSlop` for specifically how far the bottom might be. The
 * idea is that a message becomes read when the bottom of its content
 * scrolls into view, excluding the padding between messages.
 *
 * This function doesn't check that the message is visible; it probably
 * makes sense to call only when `isVisible` is already known to be true.
 *
 * See `isVisible` for the meaning of `top` and `bottom`.
 */
function isRead(element: Element, top: number, bottom: number): boolean {
```

Does that agree with your understanding?

One thing this brings to mind is that, suppose we do find ourselves scrolled down into the middle of the unread history; then I believe with these changes, as you scroll up, we'll mark a message as read as soon as the first px of its bottom (in fact of its bottom padding) scrolls into view. With the existing code, we'll only do so once 20px scroll into view (including the padding, I guess -- so once the whole bottom padding plus 4px of the message's inner box are in view.)

Probably both those behaviors are almost equally annoying, plus that's a situation that I think we should avoid getting into (but there are almost certainly bugs that do get us into it), so this isn't a reason to block merging this if we don't see an easy way to avoid it.

gnprice (commented):
Thanks @WesleyAC ! I agree this will be a good UI change, and I think this algorithm makes sense. Some comments below that are mainly stylistic, including requests for more doc comments :slightly_smiling_face: 

There's also one potential behavior issue that I just thought of, prompted by writing the last comment below: I think this can cause some glitches if the user scrolls quickly past a lot of messages. I'll discuss in [the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4526.2C.20unread.20bar.20jank/near/1140324).

src/webview/js/js.js
```diff
@@ -343,10 +354,11 @@ function visibleMessageIds(): {| first: number, last: number |} {
   let first = Number.MAX_SAFE_INTEGER;
   let last = 0;
 
-  // Walk through visible elements, observing message IDs.
+  // Walk through visible and read elements, observing message IDs.
   function walkElements(start: ?Element, step: 'nextElementSibling' | 'previousElementSibling') {
+    // what happens if element is not read?
```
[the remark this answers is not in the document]
WesleyAC: Nope, I just wrote this to myself and forgot to remove it.

src/webview/js/js.js
```diff
@@ -295,12 +295,23 @@ function lastMessage(): ?Element {
...
+}
+
+// The number of pixels of the message that are allowed to be scrolled off the
+// bottom of the screen when we mark the message as "read". Picked
+// based on the bottom padding of messages in CSS. This doesn't work so great
+// in cases where messages have the "edited" badge or reaction emojis, but it's
+// more important for the single-line message case to not get inadvertently read
+// than it is for messages to get marked read exactly when the bottom of the
+// text of the message comes onto the screen. In the future, we may want more
+// complicated behaviour here, but this should be fine for now.
```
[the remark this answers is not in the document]
WesleyAC: I edited the comment, the behavior is that you have to scroll to the bottom of the edited bar/reaction emojis/etc for the message to be marked as read.

src/webview/js/js.js
WesleyAC: Yep, that's correct. Added a jsdoc.

src/webview/js/js.js
```diff
@@ -295,12 +295,23 @@ function lastMessage(): ?Element {
...
+const messageReadSlop = 16;
```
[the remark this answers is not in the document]
WesleyAC: Yep, this is correct.

This function doesn't try to determine whether a message is visible, so every message above the bottom of the screen will be "read" according to this function.

WesleyAC:
@gnprice I've addressed all of the comments, should be ready for another review :)

src/webview/js/js.js
```diff
@@ -426,8 +481,11 @@ const sendScrollMessage = () => {
     startMessageId: rangeHull.first,
     endMessageId: rangeHull.last,
   });
-  setMessagesReadAttributes(rangeHull);
-  prevMessageRange = messageRange;
+  // Only update things if we have visible + read messages.
+  if (messageRange.first < messageRange.last) {
+    setMessagesReadAttributes(rangeHull);
+    prevMessageRange = messageRange;
+  }
```
gnprice, line 488: This comment provides a "what" but I think a "why" would be helpful. We did after all just send a message to the RN side; why is it desirable not to make these updates? If I didn't remember the [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4526.2C.20unread.20bar.20jank/near/1144664) about this, I think I'd be quite mystified :slightly_smiling_face: 

In fact I think `setMessagesReadAttributes` should get called whenever we're sending that message (or "webview outbound event" -- "message" is a confusing word to use here, sorry). The event causes our RN code to go tell the server to please mark those messages as read, while `setMessagesReadAttributes` marks them immediately as read in the CSS, causing the green bar to fade away. If in some conditions one of those gets called but not the other, that makes it easy to have a bug where some message gets marked as read in one place but not the other.

src/webview/js/js.js
```diff
@@ -321,6 +355,25 @@ function someVisibleMessage(top: number, bottom: number): ?Element {
...
+    return start;
+  }
+
+  // If the message we pick is visible but not read, that means it's partway
+  // off the bottom of the screen. Therefore, either:
+  //
+  // * The message above it will be visible and read
+  // * There are no visible read messages
+  if (!isRead(start, top, bottom)) {
+    return walkToMessage(start, 'previousElementSibling');
```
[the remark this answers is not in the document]
gnprice: Added that `walkToMessage` jsdoc: ec5e07bef

gnprice:
BTW a workflow thing that'd be helpful for me as a reviewer: when you address a comment, please leave the GitHub comment thread unresolved. That way the reviewer (here, me) can conveniently look back through them when reading your next revision -- good for a reminder of what they previously said, and also to see your replies. I'll usually then resolve the GitHub threads at that point.

For keeping one's own checklist as the author addressing the feedback, I'll sometimes just leave replies like "Done!" Or an idea I owe to Chris is to just leave a thumbs-up reaction on the comment once I've addressed it, if there's nothing more to say.

src/webview/js/js.js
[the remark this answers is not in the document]
WesleyAC: Done, with basically this approach :)

src/webview/js/js.js
WesleyAC: Done. I wasn't sure whether to include the `setMessagesReadAttributes`/`sendMessage` in the if statement — i opted not to, since the `sendMessage` seems like it could possible be important even if we don't update the `prevMessageRange`, but I'm open to doing it the other way if that seems better.

WesleyAC:
Ok, I think I addressed all the feedback, ready for another review :)

src/webview/js/js.js
```diff
@@ -333,6 +372,23 @@ function someVisibleMessage(top: number, bottom: number): ?Element {
...
+  // Algorithm: If there's a visible message that isn't read, that means
+  // it's partway off the bottom of the screen. Therefore, either:
+  // * the message above it will be visible and read, or
+  // * there are no visible read messages.
+  const visible = someVisibleMessage(top, bottom);
+  if (visible === null || visible === undefined) {
+    return visible;
+  }
+  return checkRead(visible) || checkRead(previousMessage(visible));
+}
```
gnprice, line 390: Hmm, I know this is following the code I suggested, but rereading I see it isn't quite right! I think with a small fix it is, though.

Specifically, if we're in the middle of a long message, then we'll get that as `visible`; we'll try that in the first `checkRead`, and it'll fail; and then we'll find the previous message (supposing there is one) and pass that to `checkRead`. And it will accept it and we'll return that, even though it's out of view above the screen and isn't visible.

One fix that I believe does the job is to add an `isVisible` check in `checkRead` (and also change its name to match: like `checkVisibleAndRead`, or `checkQualified`, or just `check`.)

Alternatively, could inline the `checkRead` helper and just write out the sequence of conditionals, with something like `const previous = previousMessage(visible);` at the point where we're about to consider that one.

src/webview/js/js.js
```diff
@@ -333,6 +372,23 @@ function someVisibleMessage(top: number, bottom: number): ?Element {
...
+  }
+
+  // Algorithm: If there's a visible message that isn't read, that means
+  // it's partway off the bottom of the screen. Therefore, either:
+  // * the message above it will be visible and read, or
+  // * there are no visible read messages.
+  const visible = someVisibleMessage(top, bottom);
+  if (visible === null || visible === undefined) {
+    return visible;
+  }
```
gnprice, line 388: nit: can shorten condition to `!visible`

It's important not to do that with a string or number, of course, but in the situation where it's either an object, or null or undefined, then just testing falsiness is equally correct and helps make it not feel like a lot of boilerplate. And the type-checker is there to make sure we know the difference -- in fact it will complain if you use `!foo` when `foo` could be string or number, which is a feature I quite appreciate :slightly_smiling_face:

src/webview/js/js.js
```diff
@@ -345,23 +401,23 @@ function idFromMessage(element: Element): number {
...
  *
- * If no messages are visible, the return value has first > last.
+ * If no messages are both visible and read, the return value has first > last.
  */
-function visibleMessageIds(): {| first: number, last: number |} {
-  // Algorithm: We find some message that's visible; then walk both up and
-  // down from there to find all the visible messages.
+function visibleReadMessageIds(): {| first: number, last: number |} {
+  // Algorithm: We find some message that's both visible and read; then walk
+  // both up and down from there to find all the visible messages.
```
gnprice, line 410:
```suggestion
  // Algorithm: We find some message that's both visible and read; then walk
  // both up and down from there to find all the visible read messages.
```

gnprice (commented):
Thanks! Comments below.

src/webview/js/js.js
gnprice: Yep, this way is right! In fact there are two things we do in the outer code when we get one of these outbound events -- here's from `handleWebViewOutboundEvent.js` (which is also the file that defines these types like WebViewOutboundEventScroll`:
```js
    case 'scroll':
      fetchMore(props, event);
      markRead(props, event);
      break;
```

The latter of those two functions is telling the server to mark things as read, so it'll do nothing given an empty range like this (or in general if there were no still-unread messages in the range.) But the former is important for fetching more messages if the user has scrolled near the edge of what we have so far.

src/webview/js/js.js
```diff
@@ -442,7 +499,14 @@ const sendScrollMessage = () => {
     endMessageId: rangeHull.last,
   });
   setMessagesReadAttributes(rangeHull);
-  prevMessageRange = messageRange;
+  // If there are no visible + read messages (for instance, the entire screen
+  // is taken up by a single large message), then we don't want to update
+  // prevMessageRange, since doing so could cause unread messages that are
+  // above the currently visible message to not get marked read next time
+  // sendScrollMessage is called.
```
gnprice, line 506: This is accurate but can perhaps be a bit clearer by describing directly how it does behave instead of how it avoids behaving:
```suggestion
  // prevMessageRange.  This way, if the user scrolled past some messages
  // to get here, then even though `messageRange` was empty this time and so
  // we didn't mark any messages as read just now, we'll include those in `rangeHull`
  // the next time the user scrolls and so we'll mark them as read then.
```

(Except wrapped properly -- that's more annoying to do in the GitHub comment form than in a full editor.)

Writing it out that way, I also get the feeling that this is a bit of a kludgy way of handling this case: it'd be better to go ahead and mark now the messages we *do* already know we want to mark as read. So in particular if the user scrolled down to get here, that means the messages from wherever they were up through the message that's just above the current visible one.

But getting that right may open up a new set of subtleties to debug, so it's probably best left for a followup. I think this behavior is good enough that it doesn't need to block merging this PR and getting the benefits of the main change it makes.

gnprice:
(Also replied on a thread above: https://github.com/zulip/zulip-mobile/pull/4561#discussion_r607563653 which is something GitHub's UI doesn't necessarily make it easy to spot.)

src/webview/js/js.js
WesleyAC: Makes sense :) Writing code like this makes me miss the `?` operator in Rust...

src/webview/js/js.js
WesleyAC: Ah, yup. Updated to be `checkReadAndVisible`, since inlining it seemed like a pain.

src/webview/js/js.js
WesleyAC: Yeah, I agree that this is kludgy, but it seemed like the most likely to be correct thing. In general, it seems like it would be nice to rewrite a lot of this code to use a more abstract model of messages (ie, not directly operate on the DOM), so that it would be more easy to test.

WesleyAC:
Thanks! Updated.

src/webview/js/js.js
gnprice: Yeah. Kotlin has `?:`, which provides much the same convenience. JS has not done as well at this.

gnprice:
Thanks! Looks good -- merged.

I made one small wording fix, which I noticed on a final reading:
```diff
-/** The range of message IDs that are both visible and read whenever we last
- * checked. */
+/**
+ * The range of message IDs that were both visible and read whenever we last
+ * checked.
+ */
 let prevMessageRange = visibleReadMessageIds();
```

(Also while I was editing that, adjusted the formatting; generally when it goes past one line we expand the formatting like that.)

src/webview/js/js.js
gnprice: Let's continue discussion in [the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/marking.20read.20on.20scroll/near/1155043).

PR #4566, opened 2021-03-26
action sheets [nfc]: Factor out makeButtonCallback.

gnprice:
This deduplicates a bit of code, notably the try/catch and the alert on error.

A small followup to https://github.com/zulip/zulip-mobile/pull/4541#pullrequestreview-622564692 .


WesleyAC (approved):
Oh, nice! I didn't think to pass in the entire Args object when I was writing this, that makes sense.

Issue #3927, opened 2020-02-21
Drop caching (`createSelector`) that isn't caching any real work


gnprice:
This is something I originally described at https://github.com/zulip/zulip-mobile/issues/3015#issuecomment-440890524 , a comment on #3015 which was about attempted caching that didn't actually create working caches at all.

Something I learned from studying `reselect` and `react-redux` after reviewing #2450, back in 2018-11: in many selectors where we use `createSelector`, we really don't get a meaningful benefit from it! For example, until quite recently we had:
```js
export const getFetchingForNarrow = (narrow: Narrow): Selector<Fetching> =>
  createSelector(
    getFetching,
    fetching => fetching[JSON.stringify(narrow)] || DEFAULT_FETCHING,
  );
```
This should just be, and now (9556994cf) is:
```js
export const getFetchingForNarrow = (state: GlobalState, narrow: Narrow): Fetching =>
  getFetching(state)[JSON.stringify(narrow)] || DEFAULT_FETCHING;
```
(Also it should be renamed by s/Active//.)

The reason is that even if we'd used `createSelector` correctly (vs. #3015), like so:
```js
export const getFetchingForNarrow = Selector<Fetching, Narrow> =>
  createSelector(
    (state, narrow) => narrow,
    state => getFetching(state),
    (narrow, fetching) => fetching[JSON.stringify(narrow)] || DEFAULT_FETCHING,
  );
```
then what the resulting selector would do when called is basically
* (We're here because either `state` or the component's `props` has changed.)
* Check if either `state` or `narrow` has changed.
* If so, recompute `fetching`; else we're done.
* Check if either `fetching` or `narrow` has changed.
* If so, recompute the final result.

Whereas the simpler thing, when called, does
* (We're here because either `state` or the component's `props` has changed.)
* Compute `fetching`.
* Compute the final result.

This will always recompute `fetching`, but `getFetching` is cheap, and if it weren't it should have its own caching... and it will always recompute the final result. But on the other hand there's some small cost to those "check if changed" lines -- and most importantly, the simpler thing is much simpler to reason about and fully understand.

So in general, whenever the function that computes the final result -- i.e. that last argument to `createSelector` -- is doing very little work, we're better off dropping the `createSelector`. In particular if it's just a property access or two and some boolean logic, let's cut `createSelector` -- my guess is that it's actually spending *more* time going through the abstractions to check the cache than we would spend just computing the result directly.

[...]

PR #4568, opened 2021-03-27
Stream title actionsheet

WesleyAC:
This adds an actionsheet when you long-press a stream title

### DIFF
--- a/src/title/TitleStream.js
+++ b/src/title/TitleStream.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
+import React from 'react';
 import { Text, View, TouchableWithoutFeedback } from 'react-native';
 
 import type { Narrow, Stream, Subscription, Dispatch } from '../types';
@@ -23,8 +23,8 @@ type Props = $ReadOnly<{|
   ...SelectorProps,
 |}>;
 
-class TitleStream extends PureComponent<Props> {
-  styles = createStyleSheet({
+const TitleStream = (props: Props) => {
+  const componentStyles = createStyleSheet({
     outer: {
       flex: 1,
       flexDirection: 'column',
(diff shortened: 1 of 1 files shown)


WesleyAC:
screenshot

src/title/TitleStream.js
```diff
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
+import React from 'react';
```
chrisbobbe:
```
I've chosen to wrap it in React.memo to mirror the usage of
PureComponent, although I'm not sure that we actually get a performance
benefit from doing so with this particular component.
```

Good thought!

Hmm, yeah, I don't see an obvious potential for an important performance benefit here.

Based on an earlier discussion, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1092351), I think this means we should do without `React.memo` here. Greg says (edited just so the issue link works):
> I think this will be a lot like `createSelector`: #3927. We'll want to introduce explicit memoization where there's significant computation happening in a render method -- either with `React.memo`, or perhaps with `useMemo` inside the component -- but because it makes the code more complicated, we'll want to do so only then, and not just all the time.

src/title/TitleStream.js
```diff
@@ -70,4 +104,11 @@ const TitleStream = React.memo<Props>((props: Props) => {
 
 export default connect<SelectorProps, _, _>((state, props) => ({
   stream: getStreamInNarrow(state, props.narrow),
+  backgroundData: {
+    auth: getAuth(state),
+    mute: getMute(state),
+    subscriptions: getSubscribedStreams(state),
```
chrisbobbe, line 110: I notice that `backgroundData.subscriptions` is `getSubscribedStreams(state)`, here, whereas in `MessageList`, it's `getSubscriptions(state)`.

I haven't yet worked out any differences in behavior this might cause, but I was wondering if this was intentional and if you had a prior feeling about which one is more correct in either place.

chrisbobbe (commented):
Thanks, @WesleyAC, this looks great! See some small comments, below.

Also, here's an iOS screenshot, since I was just testing on iOS:

image

src/title/TitleStream.js
gnprice: Yeah, agreed.

gnprice (commented):
Thanks @WesleyAC! Looks great -- one other small comment, in addition to the items Chris spotted.

src/title/TitleStream.js
WesleyAC: Hmmm, interesting. I don't think there's a point in using `getSubscribedStreams`, so I've switched to `getSubscriptions`.

WesleyAC:
Updated, should be ready for another review :)

src/title/TitleStream.js
```diff
@@ -23,8 +23,8 @@ type Props = $ReadOnly<{|
   ...SelectorProps,
 |}>;
 
-class TitleStream extends PureComponent<Props> {
-  styles = createStyleSheet({
+const TitleStream = (props: Props) => {
+  const titleStreamStyles = createStyleSheet({
```
gnprice: nit: This gets renamed to `titleStreamStyles` in the first commit, then `componentStyles` in the second. It'd make the diffs slightly cleaner to read if it becomes `componentStyles` directly in the first commit.

I'd make this change with `git rebase -i`, choose `edit` (or `e`) for the first commit, let it run, make the edit, say `git add -u && git rebase --continue` (actually I [have an alias](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3) `git rac` to save typing that), probably resolve conflict which probably just means picking the appropriate lines from what's there, say `git rac` again. Takes about a minute.

This example isn't a critical thing to fix, but it is a real (though small) cleanup and is an easy edit to make, as edits to commit series go. I think it's worth the practice to do easy ones like this -- there will be plenty of other cases where it's trickier :slightly_smiling_face:, and it's good to get to where the mechanics of it are so familiar they're easy and almost automatic to do.

gnprice (commented):
Thanks @WesleyAC for the revision! One small comment below -- then after taking care of that, please merge at will.

Issue #4238, opened 2020-08-24
Use `op` rather than `operation` for the previously anomalous `update_message_flags` events with feature level 32+


timabbott:
See https://github.com/zulip/zulip/pull/16168 for details.  We'll need to check `feature_level` for which field to look at, but this should let us eventually delete the server-side compatibility code to send `operation` as well on these events.

PR #4581, opened 2021-03-31
eventToAction: Use `op` on `update_message_flags` when possible.

chrisbobbe:
Fixes: #4238

### DIFF
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -1,4 +1,5 @@
 /* @flow strict-local */
+import invariant from 'invariant';
 import { EventTypes } from '../api/eventTypes';
 
 import * as logging from '../utils/logging';
@@ -34,7 +35,7 @@ import {
 } from '../actionConstants';
 import { getOwnUserId, tryGetUserForId } from '../users/userSelectors';
 import { AvatarURL } from '../utils/avatar';
-import { getCurrentRealm } from '../account/accountsSelectors';
+import { tryGetActiveAccount, getCurrentRealm } from '../account/accountsSelectors';
 
 const opToActionUserGroup = {
   add: EVENT_USER_GROUP_ADD,
@@ -89,6 +90,11 @@ const actionTypeOfEventType = {
 // This FlowFixMe is because this function encodes a large number of
 // assumptions about the events the server sends, and doesn't check them.
 export default (state: GlobalState, event: $FlowFixMe): EventAction | null => {
+  invariant(
+    tryGetActiveAccount(state),
+    'Expected to have an active account when `eventToAction` was called.',
+  );
+
   switch (event.type) {
     // For reference on each type of event, see:
     // https://zulip.com/api/get-events#events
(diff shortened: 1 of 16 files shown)


src/events/eventToAction.js
```diff
@@ -226,6 +238,12 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction => {
       return {
         ...event,
         type: EVENT_UPDATE_MESSAGE_FLAGS,
+        op:
+          zulipFeatureLevel !== null
+            ? zulipFeatureLevel >= 32
+              ? event.op
+              : event.operation
+            : event.op ?? event.operation,
```
gnprice, line 246: Can this just unconditionally say `event.op ?? event.operation`? I think it can.

The cases where that would behave differently from this are where either an older server somehow sends `op`, or a newer server fails to but does send `operation`. If either of those does ever happen for some reason, I think the behavior of `event.op ?? event.operation` is at least as likely to be the right one as this logic's is.

Where a straightforward "use this if present, else fall back to that" approach works, I think it's best to prefer that. It means less conditionals to reason through, and I think it pretty directly expresses what we really mean.

The places where `zulipFeatureLevel` is critical to use are in making requests (where we have to decide whether to try to use some newer feature), or in interpreting data where the semantics changed in some more complex way so that a simple fallback doesn't suffice.

gnprice (commented):
Thanks @chrisbobbe! I like the approach of converting everything on the inside to use the new form. All looks good modulo the comment below.

src/events/eventToAction.js
chrisbobbe: Cool, yeah, that makes sense. I was erring toward encoding our assumptions as exactly as possible, in case one day we find a high level of exactness is required—which would be for reasons I don't yet understand. If you don't see a need to be this exact, then I'm also happy to use the simpler version 🙂.

One of the earlier, prep commits adds an invariant at the top of `eventToAction`, to require that there be an active account when `eventToAction` is called. This seems like something we'd want, in principle, but now that it's not strictly necessary, I wonder if we should keep it? There might be some nasty surprises where it catches a case we haven't thought of yet.

chrisbobbe:
Thanks, @gnprice! Revision pushed. Also please see my comment at https://github.com/zulip/zulip-mobile/pull/4581#discussion_r607315106.

src/events/eventToAction.js
chrisbobbe: Ah, certainly: I've already got one on the `UpdateMessageFlagsEvent` type, but this is a place we'll want to see it too.

gnprice:
Merged, thanks!

src/events/eventToAction.js
gnprice: > One of the prep commits adds an invariant at the top of `eventToAction`, to require that there be an active account when `eventToAction` is called. This seems like something we'd want, in principle, but I wonder if we should or shouldn't keep it? There might be some nasty surprises where it catches a case we haven't thought of yet.

Hmm. I merged this (after reading this comment) and then came to write a reply here, and realized: we can indeed do better! We can use `logging.error` so we see it in Sentry but don't cause additional failures. I'll send a PR with that momentarily.

gnprice: #4627

chrisbobbe:
Thanks for the review!

Issue #4579, opened 2021-03-31
Show logged-in/logged-out state in account picker.


chrisbobbe:
`AccountPickScreen` shows all the accounts the app knows about, whether or not they're logged in.

Each item in the list should indicate if the account is logged in (it has a non-empty `.apiKey`) or logged out (it has an empty `.apiKey`).

gnprice:
More details, copied from #4491:

In `AccountPickScreen` we should display an account differently if it's logged out -- it should look "grayed-out" or disabled, or be marked with a label like "Signed out".
* Google does this with a "Signed out" label; see screenshots like [here](https://vaadin.com/learn/tutorials/google-login) or [here](https://www.carlsetzer.com/2020/03/03/google-tip-deleting-old-google-sign-in-accounts/).

vaibhavs2:
can it be assign to me?

chrisbobbe:
Thanks, @vaibhavs2!

PR #4583, opened 2021-04-01
Add sign-out label to account-list item in AccountPickScreen, Fixes: #4579

vaibhavs2:
Fixes: #4579 
| | |
|---|---|
|  | |

### DIFF
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -33,7 +33,7 @@ const styles = createStyleSheet({
     padding: 12,
     margin: 12,
   },
-  signOutText: {
+  signedOutText: {
     fontStyle: 'italic',
     fontSize: 12,
     textAlign: 'right',
@@ -67,7 +67,7 @@ export default function AccountItem(props: Props) {
           { backgroundColor: backgroundItemColor },
         ]}
       >
-        {!isLoggedIn && <Label style={styles.signOutText} text="Signed out" />}
+        {!isLoggedIn && <Label style={styles.signedOutText} text="Signed out" />}
         <View style={styles.details}>
           <RawLabel style={[styles.text, { color: textColor }]} text={email} numberOfLines={1} />
           <RawLabel
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -5,6 +5,7 @@ import { View } from 'react-native';
 import { BRAND_COLOR, createStyleSheet } from '../styles';
 import { RawLabel, Touchable } from '../common';
 import { IconDone, IconTrash } from '../common/Icons';
+import type { AccountStatus } from './accountsSelectors';
 
 const styles = createStyleSheet({
   wrapper: {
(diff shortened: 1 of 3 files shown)


src/account/AccountItem.js
```diff
@@ -26,34 +26,54 @@ const styles = createStyleSheet({
...
+    top: 3,
+    color: 'rgba(0,0,0,.5)',
+  },
 });
 
 type Props = $ReadOnly<{|
-  index: number,
-  email: string,
-  realm: URL,
+  accounts: AccountStatus,
```
chrisbobbe, line 48: Let's use a singular name (e.g., 'account') for a value that just represents one thing; the plural 'accounts' is a bit confusing.

src/account/AccountItem.js
```diff
@@ -26,34 +26,54 @@ const styles = createStyleSheet({
...
     <Touchable style={styles.wrapper} onPress={() => props.onSelect(props.index)}>
-      <View style={[styles.accountItem, showDoneIcon && styles.selectedAccountItem]}>
+      <View
+        style={[
+          styles.accountItem,
+          showDoneIcon && styles.selectedAccountItem,
+          { backgroundColor: backgroundItemColor },
+        ]}
+      >
+        {!isLoggedIn && <Text style={styles.signOutText}>Signed out</Text>}
```
chrisbobbe, line 69: Let's use `Label` instead of `Text`, and add the new user-facing string to static/translations/messages_en.json, to wire it up for [translation](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/translations.md).

src/account/AccountList.js
```diff
@@ -24,9 +24,7 @@ export default function AccountList(props: Props) {
         renderItem={({ item, index }) => (
           <AccountItem
             index={index}
-            showDoneIcon={index === 0 && item.isLoggedIn}
-            email={item.email}
-            realm={item.realm}
+            accounts={item}
```
chrisbobbe, line 27: I think there's enough pure refactoring of `AccountItem`'s props going on to motivate separating that work from the more interesting changes in the PR, on the [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) principle.

So, I'm thinking of at least one separate commit, for replacing the `email` and `realm` props with the `accounts` (should be singular, as mentioned elsewhere) prop.

And then maybe another commit for the `showDoneIcon` change, but I'm not sure that change is helpful; it puts responsibility on `AccountList` for knowing the details of the `state.accounts` data structure, which is something I think it wouldn't have to worry about, ideally. I think a better interface for `AccountList` than its current one, where it takes `index`, might be to have it take an `Identity` (see its jsdoc in src/types.js). But that improvement isn't necessary to get this feature in.

Then the more interesting changes would stand on their own and be easier to review and talk about in isolation.

chrisbobbe (commented):
Thanks, @vaibhavs2! See some comments below.

Please also change the commit message to follow the [commit-message style](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages) and put a "Fixes: #4579" line at the end of the commit message.

src/account/AccountList.js
vaibhavs2: Pardon me, I considered showDoneIcon as a part of refactor, because if a component have all the resources then component itself should calculate what it's need

chrisbobbe:
Thanks, @vaibhavs2! Glad to see the signed-out state in the accounts list. Merged, with a few small tweaks, including:

- Added the [`[nfc]` label](https://github.com/zulip/zulip-mobile/blob/master/docs/glossary.md#nfc-or-nfc) to the pure-refactor commit
- Squashed the last commit, adding the user-facing string to messages_en.json, into the commit that added the string to the UI. That way, the commit that added the string to the UI won't break when you select a different language.
- Added the "Fixes: #4579" text to the commit with the fix (🎉), as I suggested at https://github.com/zulip/zulip-mobile/pull/4583#pullrequestreview-626441822. 🙂
- Added two commits on top to tweak a few small things

Issue #4584, opened 2021-04-01
Add a React "error boundary" near the root of the component tree.


chrisbobbe:
White-screen crashes almost certainly mean an error somewhere in the React component tree. A React ["error boundary"](https://reactjs.org/docs/error-boundaries.html) is the thing React offers to catch such errors. We should use one, near the root of the component tree.

It can start by just showing a generic "something went wrong" message, but we'll want to expand on it with any useful debugging information that would help us diagnose and fix these high-priority issues.

chrisbobbe:
Here's a UI I've got so far. On Android and iOS, it shows the error and stack trace in a scrollable TextInput (with editing disabled), including the component stack, and it has a button at the top for copying all the details. A toast appears when you tap the copy button to confirm that the text has been copied.

The component stack is a lot less useful in release builds than debug builds.

In all these trials, I had an error with "sneaky error" throw at the top of `ZulipButton`'s `render` method. On iOS, I was logged in, and I triggered the error by going to the profile screen, which has some `ZulipButtons`. On iOS, I wasn't logged in, so the error was triggered right away on the realm input screen.

-----

[ios-debug.txt](https://github.com/zulip/zulip-mobile/files/6246462/ios-debug.txt), [ios-release.txt](https://github.com/zulip/zulip-mobile/files/6246460/ios-release.txt)
<img width=300 src="https://user-images.githubusercontent.com/22248748/113353662-0522f600-930c-11eb-8e65-1f5597a88330.png" /> <img width=300 src="https://user-images.githubusercontent.com/22248748/113353663-05bb8c80-930c-11eb-8313-0b7bf0deabec.png" />

-----

[android-debug.txt](https://github.com/zulip/zulip-mobile/files/6246461/android-debug.txt), [android-release.txt](https://github.com/zulip/zulip-mobile/files/6246463/android-release.txt)
<img width=300 src="https://user-images.githubusercontent.com/22248748/113353665-06542300-930c-11eb-949e-1a8def154b49.jpg" /> <img width=300 src="https://user-images.githubusercontent.com/22248748/113353664-06542300-930c-11eb-85d1-251eaa3057db.jpg" />

PR #4585, opened 2021-04-01
ZulipMobile: Add and use RootErrorBoundary.

chrisbobbe:
Fixes: #4584

### DIFF
--- a/src/RootErrorBoundary.js
+++ b/src/RootErrorBoundary.js
@@ -1,0 +1,147 @@
+/* @flow strict-local */
+import React from 'react';
+import { View, Clipboard, TextInput, ScrollView, Button, Platform } from 'react-native';
+import Toast from 'react-native-simple-toast';
+
+import * as logging from './utils/logging';
+
+type Props = $ReadOnly<{|
+  children: React$Node,
+|}>;
+
+type State = $ReadOnly<{|
+  error: Error | null,
+|}>;
+
+/**
+ * A React error boundary to go as near the root as possible.
+ *
+ * React error boundaries [1] shouldn't be used as part of a normal
+ * control flow; it's best and simplest to strive to keep our
+ * component hierarchy error-free. But mistakes happen, and when they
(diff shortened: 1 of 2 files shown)


chrisbobbe:
See screenshots/discussion at https://github.com/zulip/zulip-mobile/issues/4584#issuecomment-812172389.

chrisbobbe:
Hmm, potentially interesting: Sentry provides a ready-to-use component [`Sentry.ErrorBoundary`](https://docs.sentry.io/platforms/javascript/guides/react/components/errorboundary/). It looks like it was designed for use on the web, so in particular its built-in [user feedback widget](https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/user-feedback/) hasn't been working for me.

But when I pass a function for the `fallback` prop, I get the Sentry event ID along with the error info.

chrisbobbe:
> It looks like it was designed for use on the web, so in particular its built-in [user feedback widget](https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/user-feedback/) hasn't been working for me.

Though it looks like it might not be too hard to set something like this up manually: https://github.com/getsentry/sentry-react-native/issues/500

gnprice:
Looks good, thanks! Merging.

I have one bit of UI feedback that I think is worth doing (naturally we don't want to spend a lot of time on this UI): instead of having a lot of text in the button:
image

a more idiomatic thing to do would be to have some text to say most of that, and then have the button very concisely say what the button itself will do. Like maybe "Copy details". See e.g.:
https://material.io/components/buttons#text-button

But given that I'd like to ship a beta today with #4588, and it'd be good to include this in that too, that is not something worth blocking on and instead can be a small followup :slightly_smiling_face:

gnprice:
(Also I made a tiny formatting tweak to the output:
```diff
 Component Stack:
-
 ${
```
because it seemed like empirically there were one too many blank lines in there. My screenshot just above reflects that change.)

chrisbobbe:
Thanks for the review!

Issue #4587, opened 2021-04-02
App perma-crashes on partial update to persisted state


gnprice:
We've had a couple of reports like [the following](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1136722):
> Over the past few months when I log into Zulip via my Android phone, I get a white screen and no further. In the past I would close the app and log in again. Eventually it would let me in. But for the past 2 days it just keeps hanging on the white screen
> * Force Zulip app to stop
> * Start Zulip app
> * Zulip logo appears for a brief second
> * White screen

Thanks to logs kindly provided by Adam Birds, we've been able to diagnose what's happening:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1150343

Fundamentally this issue is a reminder that the architecture where we rely on redux-persist is not a good one, and a sign that it's overdue for replacement. Discussion in [this thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1150381).

But for the present, there are some straightforward [workarounds](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1150365) we can do that should resolve the issue in the most likely cases.

Crucially, those workarounds can rely for now on the fact that shortly after startup we go fetch fresh data from the server anyway -- so they only have to let the app survive that long. They will not suffice for a future where we keep a long-lived event queue on the server and use that to maintain the app's state incrementally over long periods without re-fetching from scratch.

PR #4588, opened 2021-04-02
user: Expand checks in getHaveServerData.

gnprice:
The one check we had would be perfectly sufficient in a world where
we used a reliable form of storage for keeping data from the server
on the device between runs of the app.  It does not suffice given
our model with redux-persist.

This doesn't fix the problem -- that will take something more
fundamental -- but it hopefully mitigates the worst consequences.

In particular, empirically we already reasonably gracefully handle
in the unreads view and the PM-conversations view a situation where
either `state.unreads` or `state.pmConversations` refers to streams,
subscriptions, or users that we don't have data about:
  https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1151982

Fixes: #4587

### DIFF
--- a/src/users/userSelectors.js
+++ b/src/users/userSelectors.js
@@ -193,6 +193,59 @@ export const getUserIsActive = (state: GlobalState, userId: UserId): boolean =>
 
 /**
  * Whether we have server data for the active account.
+ *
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, LGTM! Merged.

Issue #3996, opened 2020-03-22
Scroll-to-bottom button has an awkward highlight


harshbisle:
![Screenshot_20200322-102914](https://user-images.githubusercontent.com/41483548/77255782-f8e37200-6c8f-11ea-9671-18ea1be08214.png)
When the user clicks on the scroll to bottom button the hover effect looks a little bit buggy(Basically the square around the button).

ray-kraesig:
This highlight is generated by the WebView; as far as I can tell, we have no direct control over its shape – we'd have to eliminate it (`-webkit-top-highlight-color: rgba(0,0,0,0)`) and synthesize a new one.

PR #4589, opened 2021-04-03
webview: Remove highlight on scroll-to-bottom.

im-adithya:
This adds transparency to the a tag and background-color
onpress after converting the button into a circle with
border-radius 50% so as to remove the default highlight
on tap.

## Earlier
Screenshot from 2021-02-16 10-59-55

## Now
Screenshot from 2021-02-16 11-53-48

Fixes: #3996

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -205,6 +205,7 @@ body {
   z-index: 200;
   right: 5px;
   bottom: 15px;
+  border-radius: 50%;
 }
 #scroll-bottom a {
   display: block;
@@ -212,6 +213,7 @@ body {
   height: 32px;
   border-radius: 50%;
   background: hsla(170, 48%, 54%, 0.5);
+  -webkit-tap-highlight-color: transparent;
 }
 #scroll-bottom .text {
   clip: rect(0 0 0 0);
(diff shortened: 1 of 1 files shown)


im-adithya:
@gnprice Can you please review this :)

WesleyAC (approved):
Looks good to me, I'm going to go ahead and merge.

We should break out the colors into CSS variables or something, but I'll file an issue for that, that's not a new problem in this PR (`170, 48%, 54%` appears 6 times in this CSS file already, one doesn't really matter on the margin, IMO).

im-adithya:
I too felt the same but @chrisbobbe didn't mention this issue while changing the colors in #4544 (while mentioning to change the colors in base.css, refer to this [comment](https://github.com/zulip/zulip-mobile/pull/4544#discussion_r600005187)), so I thought it was intentional as it was mentioned in the constants.js file as a code comment.

Will submit a PR if required!

gnprice:
> We should break out the colors into CSS variables or something, but I'll file an issue for that, that's not a new problem in this PR (`170, 48%, 54%` appears 6 times in this CSS file already, one doesn't really matter on the margin, IMO).

Sure, that'd be a nice improvement.

Note that we support some fairly old browsers -- see the block comment at the top of `js.js`. Some might not support that feature.

OTOH we added a little while ago a build step for our webview assets, in `tools/build-webview`. As is it does nothing more than copy some files in from various sources, but we could certainly add something there that does more processing on our stylesheets if it'd help us write things more conveniently.

PR #4593, opened 2021-04-03
AccountItem: Tweak the layout of "Signed out" text.

chrisbobbe:
(A proposed followup to #4583; tagging @vaibhavs2. 🙂 See screenshots of the current layout at that PR.)

The previous position of this text looked just a little odd: to my
eye, it didn't play nicely with the vertical centering of the
trash-can icon. Now, it joins the existing left-aligned labels as
the last child of the `styles.details` View.

I'm not sure I've arrived at the best layout choice, but at least
there can be no question about vertical centering. It also makes the
code slightly easier to reason about, since it mostly fits into an
existing pattern: it's just like the existing left-aligned labels,
except it's italic instead of bold (which I like).

### DIFF
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -35,12 +35,8 @@ const styles = createStyleSheet({
   },
   signedOutText: {
     fontStyle: 'italic',
-    fontSize: 12,
-    textAlign: 'right',
-    position: 'absolute',
-    end: 10,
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks! I agree.

chrisbobbe:
Thanks for the review!

Issue #4595, opened 2021-04-05
Timestamp is not visible on notifications


im-adithya:
## Description
Currently, the notification doesn't have a timestamp making it difficult for the user to see when he/she has received the message.

## Screenshot
![original_45cd74ad-5de0-474b-ba74-73d442a5dcd1_Screenshot_20210405-124339](https://user-images.githubusercontent.com/64399555/113548149-aa8be180-960c-11eb-8ba4-a4f694f94f2f.jpg)

PR #4596, opened 2021-04-05
notifications: Add Timestamp.

im-adithya:
## Description
This adds a setShowWhen method to set timestamps
to true as it is set to false by default.

More information can be found [here](https://developer.android.com/reference/android/app/Notification.Builder#setWhen(long))

## Screenshot
Screenshot_1617606370

Fixes: #4595


im-adithya:
@gnprice @chrisbobbe Caan you please review this :)

gnprice:
Thanks @im-adithya for tracking this down and fixing it! This is an issue that'd been bugging me :slightly_smiling_face: 

From the linked doc:

> Add a timestamp pertaining to the notification (usually the time the event occurred). For apps targeting Build.VERSION_CODES.N and above, this time is not shown anymore by default and must be opted into by using setShowWhen(boolean)

Aha. So this used to work, and got broken with #3075. But they didn't document the change in the central what-changed documentation :disappointed: :
https://developer.android.com/about/versions/nougat/android-7.0-changes
and I guess we didn't notice the regression in time to make the connection.

Issue #4575, opened 2021-03-31
Feature Request: List of Current Active Users as we have on Web App


adambirds:
Would love for us to be able to see a list of current active users, as would be extremely useful to know if the person you want to talk to is online or not without having to have them on a previous direct message.

abhi0504:
@adambirds yeah i guess there must be `presence status indicator` in PM conversation too I would like to work on this issue 
😊

abhi0504:
@chrisbobbe @gnprice  whats your suggestions over this issue

abhi0504:
@adambirds if we add status indicator for offline users it looks like this 
![Android Emulator - Pixel_3_API_29_5554 03-04-2021 15_29_57 (2)](https://user-images.githubusercontent.com/56453541/113475266-1f86dc00-9492-11eb-8526-143565ab19da.png)
Is it ok ?

im-adithya:
@adambirds We already have this feature.
You can navigate to the Search bar in the PMs tab to find a list of all the Active, Idle, and Offline users!

abhi0504:
Also @im-adithya I think we can add this grey box for the offline case

PR #4597, opened 2021-04-05
removing `hideIfOffline` prop[nfc]

abhi0504:
removing `hideIfOffline` prop will show a grey box for the offline users

### DIFF
--- a/src/common/UserAvatarWithPresence.js
+++ b/src/common/UserAvatarWithPresence.js
@@ -45,7 +45,7 @@ export default class UserAvatarWithPresence extends PureComponent<Props> {
         <PresenceStatusIndicator
           style={styles.status}
           email={email}
-          hideIfOffline
+          hideIfOffline={false}
           useOpaqueBackground
         />
       </UserAvatar>


chrisbobbe:
@abhi0504 I don't see how this is a complete solution for #4575 based on its description. Could you help me with that?

abhi0504:
Thanks @chrisbobbe for review 

- It basically tells us about the user (offline or online) I accept the fact that it doesn't not resolves #4575 completely its just can be a enhancement 😊 
- I have resolved all the CI failures

chrisbobbe:
> I accept the fact that it doesn't not resolves #4575 completely its just can be a enhancement 😊

Ah, OK! Enhancements are nice. 🙂 But let's be sure not to say `Fixes: #4575`, either here on GitHub or in the commit message, then, since it's not true. 😉

chrisbobbe:
> * It basically tells us about the user (offline or online)

Could you talk more about the "why" for this change? It looks like the prop was set to `true` in #2545; do you think that was a mistake, or is there anything else that stands out from reading the history?

abhi0504:
> > I accept the fact that it doesn't not resolves #4575 completely its just can be a enhancement 😊
> 
> Ah, OK! Enhancements are nice. 🙂 But let's be sure not to say `Fixes: #4575`, either here on GitHub or in the commit message, then, since it's not true. 😉

Sure just removing that  `Fixes: #4575` tag

abhi0504:
> > * It basically tells us about the user (offline or online)
> 
> Could you talk more about the "why" for this change? It looks like the prop was set to `true` in #2545; do you think that was a mistake, or is there anything else that stands out from reading the history?

I think in my opinion there must be the `statusIndiactor` dot in offline cases

chrisbobbe:
Possibly there's something we want to do here, but it's not at all clear to me; some more explanation and evidence might have helped make this a viable PR. But some things naturally have to be closed to keep the workload manageable, and this is one of those times.

If someone wants to resume something like this in the future, I'd recommend looking at the history I mention above in https://github.com/zulip/zulip-mobile/pull/4597#issuecomment-814384746.

Issue #2792, opened 2018-07-12
Do not show edit message option in message action sheet if edit time is passed or message edit is not allowed on that realm.


jainkuniya:
This will fix issues like 

https://chat.zulip.org/#narrow/stream/48-mobile/topic/Inability.20to.20edit.20a.20message.20right.20after.20posting

@zulipbot claim

chrisbobbe:
I believe there are two realm-specific server settings we'll need to read, in order to fix this issue:

- Whether the realm allows editing messages
- The time limit on editing messages

So in particular we can't do something like assume message-editing is allowed and use a hard-coded time limit of 60 minutes, just because those are the settings on CZO. 🙂

Issue #2793, opened 2018-07-12
Do not show delete message option in message action sheet if delete time is passed or message delete is not allowed on that realm.


jainkuniya:
Similiar to #2792.

@zulipbot claim

borisyankov:
We might want to keep the option but show an error.
Hidden options are usually very opaque to users, they might even report the missing item as a bug.

jainkuniya:
ya May be disabled option, with little bit of info , why it is disabled?

chrisbobbe:
I believe there are two realm-specific server settings we'll need to read, in order to fix this issue. (I'm less sure about these than I am for #2792, but I believe these points follow fairly clearly from the issue description.)

- Whether the realm allows deleting messages
- The time limit on deleting messages

So in particular we can't do something like assume message-deleting is allowed and use a hard-coded time limit of 60 minutes, just because those are the settings on CZO. 🙂

PR #4599, opened 2021-04-05
messageActionSheet: Remove "edit","delete" option if time limit exceeds.

rajprakash00:
Currently in zulip web app when edit/delete time limit
exceeds, edit/delete option button is removed.

Handle this feature in mobile version too, like this:

1.If it is "Private messages" narrow remove the buttons.
2.If it is "Topic or Stream messages" narrow keep edit
  button for editing of topic names, remove delete button.

**ScreenShots/Media**:

https://user-images.githubusercontent.com/42652941/113569359-01ef7900-9630-11eb-926b-4d0975720778.mp4

Fixes: #2792,#2793


chrisbobbe:
Thanks, @rajprakash00! Please see a comment I've just posted on the two issues.

rajprakash00:
@chrisbobbe , coincidentally I was just examinig that and just now came to conclusion :(

Hmmm , I just found out that I did a mistake by assuming timelimit for edit/delete option as 60 mins(got confused thinking that zulip has set it as fixed..so silly! 😞 ) 
And due to that too I didnt check for another condition: "Does the realm allows edit,delete option?"

So I'll keep this PR as a **WIP** for now and soon provide the complete fix

rajprakash00:
@chrisbobbe This is ready for review .

chrisbobbe:
Thanks, @rajprakash00! One thing I notice is that the second commit's commit message says "Previous commit was handled in a wrong way".

The thing to do in this case is to make a PR that doesn't contain commits that we know are wrong. 🙂 That makes review easier, and it makes it easier to read the history of the project once the PR is merged. See [this doc](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html) for fixing up a PR branch locally and force-pushing it with a new revision. 🙂

rajprakash00:
@chrisbobbe Please take a look

src/message/messageActionSheet.js
```diff
@@ -301,12 +314,24 @@ export const constructMessageActionButtons = ({
   }
   if (
     message.sender_id === ownUser.user_id
+    // if realm allows users to "edit message", show edit option button
+    && realmAllowMessageEditing
     // Our "edit message" UI only works in certain kinds of narrows.
-    && (isStreamOrTopicNarrow(narrow) || isPmNarrow(narrow))
```
WesleyAC, line 305: You should keep this line the same as it was, and just add the `&& updateMessageTimeLimit < realmMessageContentEditLimit)` on the next line. I'm pretty sure that'll do the same thing, and it's a lot clearer.

src/realm/realmReducer.js
```diff
@@ -18,6 +18,10 @@ const initialState = {
   filters: [],
   emoji: {},
   videoChatProvider: null,
+  realmAllowMessageEditing: true,
+  realmAllowMessageDeleting: true,
+  realmMessageContentEditLimit: 10,
+  realmMessageContentDeleteLimit: 10,
```
WesleyAC, line 24: Where do these values come from?

src/reduxTypes.js
```diff
@@ -232,6 +232,10 @@ export type VideoChatProvider = {| name: 'jitsi_meet', jitsiServerUrl: string |}
  * @prop videoChatProvider - The video chat provider configured by the
  *   server; null if none, or if the configured provider is one we don't
  *   support.
+ * @prop realmAllowMessageEditing - server's permission for "edit message".
+ * @prop realmAllowMessageDeleting - server's permission for "delete message".
+ * @prop realmMessageContentEditLimit - server's timelimit for "edit message".
+ * @prop realmMessageContentDeleteLimit - server's timelimit for "delete message".
```
WesleyAC, line 238: nit: "time limit", not "timelimit"

WesleyAC (commented):
Thanks for working on this! Left a few comments below :)

src/realm/realmReducer.js
rajprakash00: @WesleyAC  I think this is the initial value fixed when a user opens any new org in zulip( that's why I kept this,  although I had a little doubt  to keep it as `initialState` value 🤔 )

src/message/messageActionSheet.js
rajprakash00: Hmm.. actually it wouldn't be same, here I'm excluding the edit option in only PM narrow but in stream/topic narrow even though time limit exceeds a message sender should have edit message option to **edit topic** atleast and I'm pretty sure we don't need that in PM narrow

chrisbobbe:
(Note for when we pick this back up: see https://chat.zulip.org/#narrow/stream/378-api-design/topic/message_content_delete_limit_seconds/near/1240862 about an API change.)

PR #4601, opened 2021-04-05
Add reusable component for screens that need server data.

chrisbobbe:
As discussed [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1149960). The intention is to use this for `ChatScreen` which needs something like this, with at least P1 urgency; see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1112199).

Marked as a draft because there are some Flow errors that I haven't managed to work through yet; I'd appreciate if @gnprice would take a look. 🙂


chrisbobbe:
Resumed in #4603.

PR #4603, opened 2021-04-05
ChatScreen: Add have-server-data gate.

chrisbobbe:
(Continuing from #4601, an earlier draft.)

As discussed:
  https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1149960
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1112199

### DIFF
--- a/src/main/MainTabsScreen.js
+++ b/src/main/MainTabsScreen.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import React, { useContext, type ComponentType, type ElementConfig } from 'react';
+import React, { useContext } from 'react';
 import { Platform, View } from 'react-native';
 import {
   createBottomTabNavigator,
@@ -10,7 +10,6 @@ import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import type { RouteProp, RouteParamsOf } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
 import type { GlobalParamList } from '../nav/globalTypes';
-import type { Dispatch } from '../types';
 import { bottomTabNavigatorConfig } from '../styles/tabs';
 import HomeScreen from './HomeScreen';
 import StreamTabsScreen from './StreamTabsScreen';
@@ -20,10 +19,8 @@ import { IconInbox, IconSettings, IconStream } from '../common/Icons';
 import { OwnAvatar, OfflineNotice, ZulipStatusBar } from '../common';
 import IconUnreadConversations from '../nav/IconUnreadConversations';
 import ProfileScreen from '../account-info/ProfileScreen';
-import { connect } from '../react-redux';
-import { getHaveServerData } from '../selectors';
 import styles, { ThemeContext } from '../styles';
-import FullScreenLoading from '../common/FullScreenLoading';
+import withHaveServerDataGate from '../withHaveServerDataGate';
 
 export type MainTabsNavigatorParamList = {|
   home: RouteParamsOf<typeof HomeScreen>,
@@ -107,59 +104,6 @@ function MainTabsScreen(props: Props) {
         />
       </Tab.Navigator>
     </View>
-  );
-}
-
-/**
- * A HOC for any server-data-dependent screen component that might be
- *   mounted when we don't have server data.
- *
- * Prevents rerendering of the component's subtree unless we have
- * server data.
- *
- * Passing a `dispatch` prop or a `haveServerData` prop to the
- * returned component will lead to undefined behavior; don't.
- */
-function withHaveServerDataGate<P: { ... }, C: ComponentType<$Exact<P>>>(
-  Comp: C,
-): ComponentType<$Exact<ElementConfig<C>>> {
-  // `connect` does something useful for us that `useSelector` doesn't
-  // do: it interposes a new `ReactReduxContext.Provider` component,
-  // which proxies subscriptions so that the descendant components only
-  // rerender if this one continues to say their subtree should be kept
-  // around. See
-  //   https://github.com/zulip/zulip-mobile/pull/4454#discussion_r578140524
-  // and some discussion around
-  //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/converting.20to.20Hooks/near/1111970
-  // where we describe some limits of our understanding.
-  //
-  // We found these things out while investigating an annoying crash: we
-  // found that `mapStateToProps` on a descendant of `MainTabsScreen`
-  // was running -- and throwing an uncaught error -- on logout, and
-  // `MainTabsScreen`'s early return on `!haveServerData` wasn't
-  // preventing that from happening.
-  return connect(state => ({ haveServerData: getHaveServerData(state) }))(
-    ({
-      dispatch,
-      haveServerData,
-      ...props
-    }: {|
-      dispatch: Dispatch,
-      haveServerData: boolean,
-      ...$Exact<P>,
-    |}) =>
-      haveServerData ? (
-        <Comp {...props} />
-      ) : (
-        // Show a full-screen loading indicator while waiting for the
-        // initial fetch to complete, if we don't have potentially stale
-        // data to show instead. Also show it for the duration of the nav
-        // transition just after the user logs out (see our #4275).
-        //
-        // And avoid rendering any of our main UI, to maintain the
-        // guarantee that it can all rely on server data existing.
-        <FullScreenLoading />
-      ),
   );
 }
 
--- a/src/main/MainTabsScreen.js
+++ b/src/main/MainTabsScreen.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import React, { useContext } from 'react';
+import React, { useContext, type ComponentType, type ElementConfig } from 'react';
 import { Platform, View } from 'react-native';
 import {
   createBottomTabNavigator,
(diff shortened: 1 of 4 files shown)


chrisbobbe:
One important difference from the previous draft is that I'm grabbing the `haveServerData` value as it's supplied by the `connect` call, instead of having the `connect` call just sit there without supplying anything (but `dispatch`) and using `useSelector`.

Manual testing on the earlier approach showed that the no-server-data state was not preventing the component's subtree from rerendering.

chrisbobbe:
This PR has `MainTabsScreen` and `ChatScreen` as the only two screen components that use the HOC.

When I [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1149943)

> Ideally we should do these things on all screens that depend on server data.

Greg [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1149945)

> Do we need to do it for all those screens (which is most of our screens), or just the ones it's possible to navigate to without using the UI of another such screen? I think the latter.

And I agree.

src/main/MainTabsScreen.js
```diff
@@ -117,32 +117,46 @@ function MainTabsScreen(props: Props) {
...
-);
+/**
+ * A HOC for any server-data-dependent screen component that might be
+ *   mounted when we don't have server data.
+ *
+ * Prevents rerendering of the component's subtree unless we have
+ * server data.
+ *
+ * Passing a `dispatch` prop or a `haveServerData` prop to the
+ * returned component will lead to undefined behavior; don't.
```
gnprice: Will it? I believe what will happen is `connect` will just clobber those props.

Also I think if Flow sees you're doing that, you'll get an error because the returned component-type doesn't accept those props -- it accepts what the argument `Comp` accepts, minus those two. That Flow error will probably be pretty obscure, though.

src/main/MainTabsScreen.js
```diff
@@ -153,4 +163,4 @@ function withHaveServerDataGate(Comp) {
   );
 }
 
-export default withHaveServerDataGate(MainTabsScreen);
+export default withHaveServerDataGate<Props, ComponentType<Props>>(MainTabsScreen);
```
gnprice, line 110: I think you can simplify this like so:
```js
export default withHaveServerDataGate<Props, _>(MainTabsScreen);
```
and let it infer the second type argument as `ComponentType<Props>`.

We'll basically always want it to be that, which is also the totally boring thing for it to be. So I think a convention of saying just `_` is the most helpful one -- like we do with `connect`.

Oh, in fact, we can simplify further: drop `C: ComponentType<$Exact<P>>` from the definition, and just substitute `ComponentType<$Exact<P>>` in the one place the definition mentions `C`.

In the type of `connect` I do something like this `C` argument, because it appears several times and there's a readability benefit to not repeating it. But it has this trade-off that it awkwardly adds a type parameter to the interface. Here we can just skip that.

gnprice (commented):
> One important difference from the previous draft is that I'm grabbing the `haveServerData` value as it's supplied by the `connect` call, instead of having the `connect` call just sit there without supplying anything (but `dispatch`) and using `useSelector`.
> 
> Manual testing on the earlier approach showed that the no-server-data state was not preventing the component's subtree from rerendering.

Ah, that set of observations makes sense now that I think about it. I don't quite have in my head all the mechanics of how this scenario would work, but: if we have a `connect` sitting there but then its inner component uses `useSelector` to control the conditional, then the `connect` will go ahead and cause the inner component to rerender before deciding whether to notify descendant components to rerender... but that inner component will be getting its data from a source that doesn't necessarily get the updated information at the same time as `connect` would have provided it.

Hmm, in fact: the inner component calling `useSelector(getHaveServerData)` means it's getting that information in the same way as all those other components below in the tree, via the new `ReactReduxData.Provider` that's interposed by `connect`. The point for us of making sure to get that new provider interposed here is so that those components below in the tree continue to get the old version of the data until we've decided that the new version is OK to render from. But that means that to make that decision correctly, we need to be looking at the new version at a time when those descendant components are still seeing the old version.

I still don't feel I have that entirely pinned down, but I think that may be an accurate intuition for why `useSelector(getHaveServerData)` wasn't working here, and actually getting the information from the `connect` is necessary.

---


> This PR has `MainTabsScreen` and `ChatScreen` as the only two screen components that use the HOC.
> 
> When I [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Android.3A.20White.20screen/near/1149943)
> […]

Hmm. But didn't we then discuss how in the case of a forced logout, this issue could arise on any screen that needs server data? Regardless of how one navigates to it, that is, because the forced logout can come when we're already on it.

Fortunately I think the tip commit of this branch (adding it to `ChatScreen`) demonstrates that it takes very little code per screen to do that, after the infrastructure built in the rest of the branch.

src/main/MainTabsScreen.js
chrisbobbe: > Will it? I believe what will happen is `connect` will just clobber those props.

Ah, I meant "undefined" in the sense of "we don't care to define", not "we can't define". 🙂 To make the component's interface more stable across implementation changes.

Does that sound like a good strategy? If so, I think this bit of jsdoc might be good to keep, especially with the likelihood that the Flow error would be obscure. There may be a better way to write it, though.

chrisbobbe:
Thanks for the review!

My new revision doesn't yet use the HOC on all screen components that expect server data; I'm questioning an approach for where to call the HOC in a "no merge" commit at the tip.

Also, please see my comment at https://github.com/zulip/zulip-mobile/pull/4603#discussion_r608205783. :)

gnprice:
the same file as the components we're calling them on. Also,
    interacting with the type-param interface gets awkward. Forgetting
    to pass the type param is (as far as I've seen) allowed, when the
    calls are here, but it means we lose type-checking of the resulting
    component (it didn't complain when I tried to put a
    `withHaveServerDataGate(MainTabsScreen)` for e.g. the `emoji-picker`
    screen.

Hmmm.

Well, here's a version that works!
```diff
-export default function withHaveServerDataGate<P: { ... }>(
-  Comp: ComponentType<$Exact<P>>,
-): ComponentType<$Exact<P>> {
+export default function withHaveServerDataGate<P: { ... }, C: ComponentType<$Exact<P>>>(
+  Comp: C,
+): ComponentType<$Exact<ElementConfig<C>>> {
```

That allows leaving off the type parameter at the call site:
```
-      <Stack.Screen
-        name="main-tabs"
-        component={withHaveServerDataGate<ElementConfig<typeof MainTabsScreen>>(MainTabsScreen)}
-      />
+      <Stack.Screen name="main-tabs" component={withHaveServerDataGate(MainTabsScreen)} />
```

plus now if you use a component that doesn't match up with the route name (like `MainTabsScreen` with `emoji-picker`), you get an error. Actually a series of errors, but some of them include the key identifiers to let you know what's wrong.

I don't totally have an explanation of why Flow doesn't give an error in the other case. It seems like in principle it should. I think with type-level operators like `$Diff` and `ComponentType` (the latter of which is built into Flow along with some other React-specific logic), sometimes the type information doesn't quite flow bidirectionally in all the ways that it should, and it's possible that one of those issues is the cause here.

What this version does is basically copy the way `connect` works:
```js
export function connect<SP, P, C: ComponentType<P>>(
  mapStateToProps?: (GlobalState, OwnProps<C,
    // Error "property `foo` is missing"?  Add to inner component's props.
    SP>) => SP,
): C => ComponentType<$ReadOnly<OwnProps<C, SP>>> {
```
but with the return type simplified a bit because it should be passing through the same props.

gnprice:
And then I agree it's probably best to have the calls to `withHaveServerDataGate` at the navigator, rather than at the components' definitions. It'll certainly make it easier to scan through and spot any that should have it and don't.

chrisbobbe:

I did a bit of tweaking and testing with `TimingScreen`, a random one of those screens that gets wrapped. Part of that was to make it a function component, so I've left that conversion commit in. 🙂

src/main/MainTabsScreen.js
gnprice: Hmm, rereading, I think what I said above is correct about the runtime behavior -- but if the inner component wants to take those props, and you pass them to the outer component, Flow won't give an error.

That's seen in the fact that the signature takes a component and returns a component which takes the same props:
```js
export default function withHaveServerDataGate<P: { ... }, C: ComponentType<$Exact<P>>>(
  Comp: C,
): ComponentType<$Exact<ElementConfig<C>>> {
```
so if Flow were going to complain about how the implementation doesn't totally agree with that, it should do so already when type-checking the implementation.

And indeed empirically:
```js
function example(c: ComponentType<{| dispatch: Dispatch |}>) {
  const MyComponent = withHaveServerDataGate(c);
  <MyComponent />; // error
  <MyComponent dispatch={(null: $FlowFixMe)} />; // no error
}
```

I think that's basically OK, because trying to pass these two props through doesn't seem like a likely mistake to make, and I don't think one would get very far before seeing it wasn't working.

But yeah, a jsdoc mention of it seems good. I'll try tweaking the wording.

gnprice:
Looks good, thanks! Merged.

I added a comment-commit on top reflecting that remaining thread above.

chrisbobbe:
Thanks for the review and that tweak!

Issue #4605, opened 2021-04-06
PasswordAuthScreen: Crash in `WebLink` following #4491.


chrisbobbe:
The `WebLink` in `PasswordAuthScreen` depends on the current realm being stored in Redux. We stopped storing it there in #4491, so we get a crash in `PasswordAuthScreen`.

gnprice:
Glad we caught this before it went into a release :smile: 

Looking at your PR now.

gnprice:
Ah, is this specific to the "Forgot password?" link?

If so I think this is P1 but not P0. That's a flow that any given user rarely needs and probably most users never need; and there's always the workaround of doing it on web and coming back to mobile. If this were preventing a large class of users from logging in at all, that'd definitely be P0.

Still -- good to have caught this!

chrisbobbe:
> If this were preventing a large class of users from logging in at all, that'd definitely be P0.

Ah, I should have been more specific: the symptom _is_ that a large class of users are prevented from logging in at all. It's a white-screen "crash" on viewing the password-auth screen at all; it doesn't require clicking on the link.

gnprice:
Ahh I see, because the exception happens at the selector passed to `connect`. :+1:

PR #4606, opened 2021-04-06
PasswordAuthScreen: Fix a crash.

chrisbobbe:
Fixes: #4605

### DIFF
--- a/src/common/WebLink.js
+++ b/src/common/WebLink.js
@@ -2,15 +2,13 @@
 
 import React from 'react';
 
-import { useSelector } from '../react-redux';
 import Label from './Label';
 import openLink from '../utils/openLink';
-import { getCurrentRealm } from '../selectors';
 import { BRAND_COLOR, createStyleSheet } from '../styles';
 
 type Props = $ReadOnly<{|
   label: string,
-  href: string,
+  url: URL,
 |}>;
 
 const componentStyles = createStyleSheet({
@@ -26,14 +24,12 @@ const componentStyles = createStyleSheet({
  * A button styled like a web link.
  */
 export default function WebLink(props: Props) {
-  const realm = useSelector(getCurrentRealm);
   return (
     <Label
       style={componentStyles.link}
       text={props.label}
       onPress={() => {
-        const { href } = props;
-        openLink(new URL(href, realm).toString());
+        openLink(props.url.toString());
       }}
     />
   );
--- a/src/common/WebLink.js
+++ b/src/common/WebLink.js
@@ -27,10 +27,6 @@ const componentStyles = createStyleSheet({
 
 /**
  * A button styled like a web link.
- *
- * @prop label - Text of the button.
- * @prop href - URL address to open on press.
- * @prop realm - Current realm. Used if the `href` property is relative.
  */
 function WebLink(props: Props) {
   return (
--- a/src/common/WebLink.js
+++ b/src/common/WebLink.js
@@ -16,6 +16,15 @@ type Props = $ReadOnly<{|
   realm: URL,
 |}>;
 
+const componentStyles = createStyleSheet({
+  link: {
+    marginTop: 10,
+    fontSize: 15,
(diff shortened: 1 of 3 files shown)


src/common/WebLink.js
```diff
@@ -1,48 +1,56 @@
...
+
+  let url;
+  try {
+    url = new URL(props.href, baseUrl).toString();
+  } catch (e) {
+    invariant(
+      baseUrl != null,
+      "Can't decide what URL to use for a `WebLink`. Did you mean to provide a `baseUrl`?",
+    );
+    throw e;
```
chrisbobbe, line 44: As long as we have total control over the URL that the `WebLink` opens, I think it makes sense to error inside the `render` method. (Now at least we have the error boundary, from #4585, so it won't just be a white screen.) Does that sound right?

If we start using user-defined or even server-defined strings, we may want to postpone this error checking until the user actually presses the link, and then we can show a toast or something if it fails to open.

src/common/WebLink.js
```diff
@@ -1,48 +1,56 @@
...
 
 type Props = $ReadOnly<{|
-  dispatch: Dispatch,
   label: string,
   href: string,
-  realm: URL,
+
+  // What to use to absolutize relative URLs, if different from the
+  // current realm or we don't expect there to be a current realm.
+  baseUrl?: URL,
```
gnprice, line 18: This feels a bit fragile, in that even if a `baseUrl` is passed, the behavior will depend on whether a current realm is present. As we saw, that fact about a call site can change without there being any obvious reason to look through to the details of this component.

How about we do away with this component's looking at the current realm entirely, and have it always take `baseUrl` from the caller? And if that's missing, it just passes `undefined` through to `new URL`, requiring the href itself to be absolute.

src/common/WebLink.js
gnprice: Certainly as long as the `href` is a literal in the caller's source code, that seems like the right behavior. That'll help make sure we spot any failure early, likely before it makes it to users.

src/common/WebLink.js
gnprice: (There's currently just the one callsite, so this doesn't involve any major refactor.)

gnprice (commented):
Thanks @chrisbobbe ! Glad we've caught this. Comments below.

src/common/WebLink.js
chrisbobbe: > How about we do away with this component's looking at the current realm entirely, and have it always take `baseUrl` from the caller? And if that's missing, it just passes `undefined` through to `new URL`, requiring the href itself to be absolute.

Hmm, actually, what do you think of just asking all callers to provide the entire absolute URL as a URL object, so the validation happens outside `WebLink`? If we're going to always require (a) an absolute URL, or (b) a base URL plus a relative URL (which together make an absolute URL), we might as well make it simpler and always require an absolute URL?

gnprice: Sure, SGTM.

chrisbobbe:
Thanks for the review! I've left in the commit that adds `tryGetCurrentRealm`, in case that ends up being useful one day. But since we don't have a current use for it, I'd be just as happy to see it go away, if you think that's best.

src/common/WebLink.js
```diff
@@ -1,48 +1,56 @@
...
+    textAlign: 'right',
+  },
+});
+
 /**
  * A button styled like a web link.
- *
- * @prop label - Text of the button.
- * @prop href - URL address to open on press.
- * @prop realm - Current realm. Used if the `href` property is relative.
```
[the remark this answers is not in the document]
chrisbobbe: Since after https://github.com/zulip/zulip-mobile/pull/4606#discussion_r607423727 we're requiring a full, validated URL for the prop (and Flow enforces it), I haven't left such a comment in; let me know if I should do something different here.

gnprice:
LGTM! Merged.

PR #4611, opened 2021-04-07
events: Better handle internal errors; clean some logic up

gnprice:
Prompted by discussion here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/stream.20typing.20notifications/near/1154301
Now if there's an exception in a reducer as we handle an event, we'll log an error and skip just that event and move on.

Also fix a bug that I discovered as a consequence of the refactoring in this branch: if we got a batch of several events in one response (e.g. because we'd been offline for a bit and they'd accumulated), then we'd get the Redux state once at the start of the batch and pass that to `eventToAction` for converting all the batch's events into actions. The use of the state there at all is kind of a hack (see jsdoc added here to `eventToAction`), and we only use it in bits and pieces for a few types of events... but clearly the correct behavior is that each event's action should be based on the state after completely applying any preceding events.


chrisbobbe (approved):
Thanks, Greg, this LGTM! Normally I'd merge; you may like to do so or await Wesley's review.

src/events/eventActions.js
```diff
@@ -77,10 +67,9 @@ export const startEventPolling = (queueId: number, eventId: number) => async (
         // are explicitly not the place for side effects (see
         // https://redux.js.org/faq/actions).
         dispatch(doEventActionSideEffects(action));
+        dispatch(action);
       });
 
-      dispatchOrBatch(dispatch, actions);
```
timabbott, line 82: This is not a request to change something, just an FYI -- in the Zulip webapp, we process the vast majority of events individually, but have this (very early/old) code to batch certain classes of events differently.  I think the goal is to avoid unnecessary rerenders -- e.g. if you're getting 10 new messages, avoiding adding each one individually 10 times, which might be fine in the webapp but certainly isn't something we've explicitly optimized there.

That behavior ends up forcing the message-related events (message edit, etc.) to also be batched, etc., which means it probably has bugs (e.g. from reading the below, I think batched up `reaction` events may get dropped from being processed before the message that they're reacting to; apparently we did this carefully when adding `submessage` but not `reaction`):

```
    const dispatch_event = function dispatch_event(event) {                                          
        switch (event.type) {                                                                        
            case "message": {                                                                        
                const msg = event.message;                                                           
                msg.flags = event.flags;                                                             
                if (event.local_message_id) {                                                        
                    msg.local_id = event.local_message_id;                                           
                    sent_messages.report_event_received(event.local_message_id);                     
                }                                                                                    
                messages.push(msg);                                                                  
                break;                                                                               
            }                                                                                        
                                                                                                     
            case "update_message":                                                                   
                update_message_events.push(event);                                                   
                break;                                                                               
                                                                                                     
            case "delete_message":                                                                   
            case "submessage":                                                                       
            case "update_message_flags":                                                             
                post_message_events.push(event);                                                     
                break;                                                                               
                                                                                                     
            default:                                                                                 
                server_events_dispatch.dispatch_normal_event(event);                                 
        }                                                                                            
    };                                                                                               

```

For non-message events where we're confident it'd be correct to do batching, once we have long-term event queues, probably the server will eventually do whatever batching transformations are safe/correct just to safe memory.

So mostly this is agreement with a "You probably don't want to do batching unless you really have to".

gnprice:
Thanks for the reviews! Merged.

PR #4612, opened 2021-04-07
Add "Stream settings" to HeaderActionSheet

WesleyAC:
This is the same thing that the (i) menu on the stream title currently
displays. Since we plan to remove that icon soon, allowing people to
access it here makes sense.

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -24,7 +24,7 @@ import {
 import { isTopicMuted } from '../utils/message';
 import * as api from '../api';
 import { showToast } from '../utils/info';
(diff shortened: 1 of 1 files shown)


gnprice:
LGTM, merged, thanks!

chrisbobbe:
Ah, the one comment I was about to add was that it'd be good to add these new user-facing strings to static/translations/messages_en.json.

gnprice:
Ah good catch, thanks! Want to push a quick change to do that?

And then we should perhaps do something to make that easier to notice when writing the change in the first place. A block comment about it, at the top or bottom of the list of these options, might be a good step.

chrisbobbe:
Sure, just sent #4619!

PR #4613, opened 2021-04-07
TitleStream: Increase long-press target size.

WesleyAC:


src/title/TitleStream.js
```diff
@@ -68,37 +68,37 @@ const TitleStream = (props: Props) => {
...
-          }}
-        >
+    <TouchableWithoutFeedback
+      onLongPress={() => {
+        showHeaderActionSheet({
+          showActionSheetWithOptions,
+          callbacks: { dispatch, _ },
+          backgroundData,
+          stream: stream.name,
+          topic: topicOfNarrow(narrow),
```
chrisbobbe, line 78: That `topicOfNarrow(narrow)` should go behind an `isTopicNarrow(narrow)` check; else there's an error when long-pressing the header if it's a stream narrow.

chrisbobbe (commented):
Thanks, @WesleyAC! See one comment below.

chrisbobbe:
Here's what the inspector shows for the touchable area after this change: 



I got there by doing this:

- Shake gesture
- "Show inspector" on the dev menu
- "Touchables" in the inspector panel
- Then either shake and "Hide inspector", or unselect "Inspect" from the inspector panel

gnprice:
Ah, good thought checking that. Certainly the ideal behavior would be for the touchable area to go the full height of the app bar.

If that seems tricky it doesn't need to block this PR (the PR is already an improvement in the same direction), but it'd be a good followup.

src/title/TitleStream.js
WesleyAC: Done, thanks for catching that!

WesleyAC:
Added a commit to increase the size of the touchable area to the full with of the navbar, and fixed the bug that @chrisbobbe pointed out. Ready for another review!

src/title/TitleStream.js
```diff
@@ -68,37 +69,39 @@ const TitleStream = (props: Props) => {
...
-              showActionSheetWithOptions,
-              callbacks: { dispatch, _ },
-              backgroundData,
-              stream: stream.name,
-              topic: topicOfNarrow(narrow),
-            });
-          }}
-        >
+    <TouchableWithoutFeedback
+      onLongPress={() => {
```
chrisbobbe, line 73: We hope to handle long presses when the narrow is a stream narrow too, right? (Whether `showHeaderActionSheet` is made explicitly topic-narrow-only and we grow something similar for stream narrows, or we adapt `showHeaderActionSheet` to be used for both kinds of narrows; I'm looking at discussion at https://github.com/zulip/zulip-mobile/pull/4608/commits/d193f787e9cc58fe9341907b258363aab2935639#r609373770.)

Until that time, it may be worth a small note that we plan to do so, so it doesn't fall off our radar; I think it's nice to avoid no-op press handlers, for how they may interfere with accessibility.

src/title/TitleStream.js
```diff
@@ -56,7 +56,8 @@ const TitleStream = (props: Props) => {
       flex: 1,
       flexDirection: 'column',
       alignItems: 'flex-start',
-      justifyContent: 'flex-start',
+      justifyContent: 'center',
```
chrisbobbe, line 59: `justifyContent: 'center'` makes sense, and I think `height: '100%'` would probably be fine.

In another direction for the latter, though, I found another option I'm now wondering about: to have `alignItems: 'stretch'` instead of `'center'` on the `View` that wraps `<NavBarBackButton>`, `<Title>`, etc., in `ChatNavBar`. I think it may make more sense for the nav bar to say it wants all its children to stretch from top to bottom, than to say it wants everything centered.

I did come at this from the angle of generally preferring to reach for flexbox in React Native (there are a few [differences](https://reactnative.dev/docs/flexbox) from how it works on the web), and possibly that's not the best thing to reach for here—or I may have landed on the wrong layout choice just trying to do something with flexbox at all. 🙂 But I thought I'd mention it since it was something that worked in my testing.

chrisbobbe (commented):
Thanks, @WesleyAC! I've just put a few small comments below; otherwise, this LGTM.

src/title/TitleStream.js
gnprice: > I think it's nice to avoid no-op press handlers, for how they may interfere with accessibility.

Ah, good catch. Yeah, so long as this is a no-op, let's make it simply `undefined` instead (which should have the same effect as not passing it at all): just put the conditional outside the function expression, like `isTopicNarrow(narrow) ? () => { … } : undefined`. That way it won't be there to mislead a11y systems.

WesleyAC: Done.

src/title/TitleStream.js
WesleyAC: I made this change, although I'm not 100% convinced that it's better. I don't have a strong preference, so I just went ahead and did it.

WesleyAC:
@chrisbobbe ready for another reveiew :)

src/title/TitleStream.js
chrisbobbe: Sure, sounds good; we can always change later.

src/title/TitleStream.js
```diff
@@ -69,17 +69,19 @@ const TitleStream = (props: Props) => {
...
+          ? () => {
+              showHeaderActionSheet({
+                showActionSheetWithOptions,
+                callbacks: { dispatch, _ },
+                backgroundData,
+                stream: stream.name,
+                topic: topicOfNarrow(narrow),
+              });
+            }
+          : undefined
```
chrisbobbe, line 83: Mm, yeah; that's a good fix for the potential a11y issue here; hopefully `TouchableWithoutFeedback` can tell the difference between `undefined` and a function that does nothing, and do the right thing for a11y.

But let's squash the fix into the same commit that introduced the potential issue, I think.

chrisbobbe (commented):
Thanks, @WesleyAC! Just one nit, below; after addressing that, please merge at will.

WesleyAC:
@chrisbobbe I thought about this a bit more, and realized that the `height: 100%` solution makes more sense — this PR as it was before broke the PM screen, since that was no longer centered. In general, it seems cleaner to fix the stream title bar by just editing its file, rather than also changing a bunch of other stuff.

chrisbobbe:
> this PR as it was before broke the PM screen, since that was no longer centered

Good catch, thanks! Merged, to help along this improvement for topic/stream narrows. This has also moved my thinking forward on that flexbox approach; see below.

> In general, it seems cleaner to fix the stream title bar by just editing its file, rather than also changing a bunch of other stuff.

I think the rightness or wrongness of those follow-on changes could reasonably be considered a factor. Often, I'll end up doing some one-time independent cleanups alongside more targeted changes. 🙂 `TitlePrivate` is in a very similar situation to `TitleStream`, and is left without its height (and so its touch target) expanded after this PR; see screenshots. An approach like `alignItems: 'stretch'` in `ChatNavBar` wouldn't centralize the logic for vertically centering the title (a `justifyContent: 'center'` in `TitlePrivate` could do that, as you've done in `TitleStream`), but it _would_ centralize the logic for expanding its touch target, and I think `ChatNavBar` could reasonably be the thing that controls that. That kind of advantage, if it is an advantage, is more accessible when flexbox is used, and [React Native promotes flexbox](https://reactnative.dev/docs/flexbox) as a sensible way to do layout. They also specifically prefer flex over percentage dimensions (different [doc](https://reactnative.dev/docs/height-and-width#percentage-dimensions)): "If you want to fill a certain portion of the screen, but you _don't_ want to use the `flex` layout, you _can_ use **percentage values** in the component's style."



My takeaway is that I/we could helpfully take note, as we go, on where flexbox vs. percentage dimensions really shine for us and should definitely be preferred in principle. As I've said in an earlier comment, I'm not hoping to use flexbox uncritically everywhere. But it can have some benefits that are harder to get without it. 🙂

PR #4618, opened 2021-04-07
Add "bash" to `tools/postinstall`

AlaaElden98:
When running `yarn` with git bash on windows, it errors with `tools\postinstall' is not recognized.`

Adding ['bash' before `tools/postinstall`](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Setting.20up.20development.20environment.20for.20zulip-app/near/1102688) resolve this.

Fixes: zulip#4427

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "postinstall": "tools/postinstall",
+    "postinstall": "bash tools/postinstall",
     "start": "react-native start",
     "ios": "react-native run-ios",
     "ios-min": "react-native run-ios --simulator 'iPhone 5s'",


docs/howto/build-run.md
```diff
@@ -152,6 +152,37 @@ Apart from the steps mentioned below, you may find the
...
+yarn install v1.22.5
+[1/4] Resolving packages...
+success Already up-to-date.
+$ tools/postinstall
+'tools\postinstall' is not recognized as an internal or external command,
+operable program or batch file.
+error Command failed with exit code 1.
+info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
+```
+then, in our `Package.json` :
```
WesleyAC, line 170: Do you literally have a `Package.json` file, with a capital `P`? Or do you mean `package.json`?

docs/howto/build-run.md
```diff
@@ -152,6 +152,37 @@ Apart from the steps mentioned below, you may find the
...
+  "name": "ZulipMobile",
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "postinstall": "tools/postinstall",
+```
+
+To fix this, edit that last line so that it reads:
+```
+    "postinstall": "bash tools/postinstall",
```
WesleyAC, line 183: Can we just make this change in the `package.json`, so it works for everyone? I don't see a downside to doing that, but I could be missing something.

WesleyAC (changes requested):
Thanks for figuring this out, and working to improve the experience for other people!

I think we can just make this change in our `package.json`, so it'll be better for everyone, do you want to just make that PR?

WesleyAC:
Seems like we [wanted to do a PR with this before](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Setting.20up.20development.20environment.20for.20zulip-app/near/1103503), but didn't get around to it :sweat_smile:

AlaaElden98:
> I think we can just make this change in our `package.json`, so it'll be better for everyone, do you want to just make that PR?

I thought of that, but i don't know if these changes in `package.json` will have a diffrent behave on macOS or not. I think it's not since you asks. 
On it.

AlaaElden98:
> Seems like we [wanted to do a PR with this before](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Setting.20up.20development.20environment.20for.20zulip-app/near/1103503), but didn't get around to it 😅

Yes,  i mentioned the issue to that 😃

docs/howto/build-run.md
AlaaElden98: Just a typo, fixing ...

docs/howto/build-run.md
AlaaElden98: I thought of that, but I don't know if these changes in package.json will have a different behave on macOS or not.
if you could test this on macOS, or you know it will not affect the behavior on macOS, I will be happy to open a PR with the changes.

gnprice: Yeah, this should have no effect on macOS or Linux.

The script itself already starts with instructions to run it under `bash`:
```
$ head -1 tools/postinstall 
#!/usr/bin/env bash
```
which is always how we want it to run. The issue is just that the Windows command interpreter apparently doesn't understand that instruction, and `yarn` is apparently trying to use the Windows command interpreter to run these lines from `package.json`.

Here's how I put it [earlier in that thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Setting.20up.20development.20environment.20for.20zulip-app/near/1102365):
````quote
The extra `bash` should be redundant, and it's a bit ugly, but if it fixes this problem on Windows then it's worth having.
````

AlaaElden98: I see, on it!

AlaaElden98:
[Check PR](https://github.com/zulip/zulip-mobile/pull/4626)

gnprice:
@AlaaElden98 BTW when you're revising a PR after review, it's generally best to update the existing PR rather than close it and make a new one. See our docs here:
https://zulip.readthedocs.io/en/latest/git/pull-requests.html#update-a-pull-request

WesleyAC:
Thanks! I adjusted the commit message to match our style — you should take a look at it and try to match that style in the future. Specifically:

* Add the area that you changed in front of the first line — `package.json` in this case (it's not always the name of a file, that's just what makes the most sense in this case, i think)
* You don't need to describe what you did in a commit like this, since `git show` will show people that. Focus on describing why you made the changes you did, rather than what you did.

Thanks for finding this and fixing it, I'm sure future contributors will appreciate the smoother experience :)

PR #4619, opened 2021-04-08
Translate some user-facing strings added to an action sheet.

chrisbobbe:
Followup as noted at https://github.com/zulip/zulip-mobile/pull/4612#issuecomment-815340027.

See also [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Warn.20if.20user-facing.20strings.20untreated/near/1156609), where I mentioned briefly a way we might automate something to help us remember to do this, since it's so easy to forget (especially since the `_(...)` call isn't right there in front of you).

(I haven't included that in this PR; if it seems too complicated, we can move on to other things. I've added some code comments that might be a bit useful, possibly not.)

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -57,11 +57,18 @@ type MessageArgs = {
 
 type Button<Args: HeaderArgs | MessageArgs> = {|
   (Args): void | Promise<void>,
+
+  /** The label for the button. */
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks! Merged. Also added a line in the last commit with the resolution of the mystery of why it was missing :slightly_smiling_face:

Issue #4559, opened 2021-03-22
Retry failed "mark as read" messages


WesleyAC:
Currently, if a API request to mark messages as read fails, we never retry it. Instead, we should save a list of messages we have tried to mark as read (this should be persisted to disk, ideally, although maybe just putting it in the Redux state is a good enough start), and retry them periodically (probably using exponential backoff, maybe with some jitter?)

I think we should probably do this for every `messagesFlags` call, not just read state: thing like starring messages deserve to not be lost as well.

One thing I'm not sure about is how to deal with stateful logic in the `src/api/` code — I don't know what dependencies we're OK with that code having (can it save stuff in Redux?). I think it would be good to aim to have fairly generic code for retrying any sort of API request, which we can at first just have apply to `messagesFlags`, but can add to other methods later, if we want.

It's also worth thinking about how the `auth` will get passed around when we do this — we want whatever is clearing out the retry queue to have the most recent `auth` rather than the one that the message was first sent with, most likely. We should also make sure that the code doesn't accidentally send messages to the wrong realm.

Related: #4558, #4526

chrisbobbe:
> fairly generic code for retrying any sort of API request

I believe `tryFetch` might be helpful here?

I've also got a `promiseTimeout` in #M4193 (not yet merged) that might be helpful if we want to give up retrying after a time. This came up in a fairly recent discussion on timeout/retry logic for sending outbox messages, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1096373).

gnprice:
This will be good to do! Among other things it's a prerequisite to #3549.

Some discussion [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/retrying.20requests/near/1139254) about how to approach the implementation.

gnprice:
> Among other things it's a prerequisite to #3549 [eagerly setting the flag locally before hearing back from the server].

Actually, thinking about this again I think it's not. (Though it'd still be good to do.) We can go ahead and do #3549 without blocking on this.

I explained why at an old comment https://github.com/zulip/zulip-mobile/issues/3549#issuecomment-525491386 :

[...]

PR #4620, opened 2021-04-08
queueMarkAsRead: Resend messages on failure

WesleyAC:


src/api/messages/messagesFlags.js
```diff
@@ -1,6 +1,16 @@
 /* @flow strict-local */
-import type { ApiResponse, Auth } from '../transportTypes';
+import type { ApiResponseSuccess, Auth } from '../transportTypes';
 import { apiPost } from '../apiFetch';
 
-export default (auth: Auth, messages: number[], op: string, flag: string): Promise<ApiResponse> =>
+export type ApiResponseMessagesFlags = {|
```
chrisbobbe, line 5:
```
messagesFlags: Return list of messages from API response.

The API gives us a list of messages we successfully marked as read,
which is useful to expose to the rest of our code.
```

Commit message nit: Not just marked-as-read messages, I think; the [doc](https://zulip.com/api/update-message-flags) says "An array with the IDs of the modified messages." (So, e.g., this includes changing the starred/unstarred status.)

src/api/queueMarkAsRead.js
```diff
@@ -30,12 +30,13 @@ const processQueue = (auth: Auth) => {
     return;
   }
 
-  messagesFlags(auth, unsentMessageIds, 'add', 'read');
-  unsentMessageIds = [];
+  messagesFlags(auth, unackedMessageIds, 'add', 'read').then(success => {
+    unackedMessageIds = unackedMessageIds.filter(id => !success.messages.includes(id));
+  });
```
chrisbobbe, line 35: We've more often used `async`/`await` syntax for Promises than `.then`, etc., but I'm not sure if that's something we'd like to enforce or if it just ends up being convenient most of the time. One nice benefit is its use in combination with `try`/`catch`, and we don't have `catch`-like logic here.

If we wanted to do that, one way is to make `processQueue` itself `async`, which
- allows the use of `await` in its body, and
- tells callers that it returns a Promise (hopefully a Promise that some callers might be interested in `await`ing). 

Then we could do,

```suggestion
  const responsePromise = messagesFlags(auth, unackedMessageIds, 'add', 'read');
  lastSentTime = Date.now();
  const success = await responsePromise;
  unackedMessageIds = unackedMessageIds.filter(id => !success.messages.includes(id));
```

or perhaps more directly,

```suggestion
  lastSentTime = Date.now();
  const success = await messagesFlags(auth, unackedMessageIds, 'add', 'read');
  unackedMessageIds = unackedMessageIds.filter(id => !success.messages.includes(id));
```

But I think I'm not fussy about using `async`/`await` here; as I'll mention in my top-level review comment, this isn't the code we plan to settle on, and making `processQueue` `async` is a change to its interface that we may or may not like.

(Another way to use `async`/`await`, one that doesn't mean making `processQueue` `async`, would be to use an `async` IIFE, but I don't do this much and I don't think it's clearer than `.then`. E.g.,

```js
  lastSentTime = Date.now();
  (async () => {
    const success = await messagesFlags(auth, unackedMessageIds, 'add', 'read');
    unackedMessageIds = unackedMessageIds.filter(id => !success.messages.includes(id));
  })();
```

)

chrisbobbe (commented):
Thanks, @WesleyAC!

I see your comment [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/retrying.20requests/near/1157357)—

> yeah, so i think what this project looks like is first rewriting `queueMarkAsRead` to use `BackoffMachine` and handle errors/being offline properly, and then once that's done, i can focus on persisting to disk.

—so I understand this PR as a small change in a slightly different direction than we'll eventually go, but I'd be happy to see it merged as it's an easy way to make the experience better.

src/api/messages/messagesFlags.js
WesleyAC: Done.

src/api/queueMarkAsRead.js
WesleyAC: That makes sense. My preference is to leave this as `.then` for now, and figure out whether we actually want an async API or not in the future, as I adjust and refactor this code more. (although I'm mostly familiar with Rust `async`/`await`, not javascript, so my intuition could be off here).

Let me know if you'd rather I go in a specific direction in this PR.

WesleyAC:
@chrisbobbe ready for another review :)

chrisbobbe:
Thanks, @WesleyAC! Merged.

gnprice:
Adding a cross-reference: this is part of #4559 .

I was just looking at that issue a few minutes ago because it came up for a different reason, and didn't see this. Which illustrates why I [always want cross-references](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#mention-related-issues) :slightly_smiling_face:

src/api/queueMarkAsRead.js
gnprice: The async logic here is pretty simple and I don't have a strong preference on whether it's written with `async`/`await` or with `.then`.

In general, though, I think we'll generally prefer `await` when things get at all complicated -- in that case it keeps things a lot easier to follow and understand than the corresponding `.then`/`.catch`/etc code.

Detailed discussion in chat:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/async/near/1160961

gnprice: In further discussion there, Anders convinced me that we should use `await` rather than `.then`, here and systematically. I've filed #4652 for the handful of existing spots in general, but in particular it'd be a good followup to this PR to do so here.

src/api/queueMarkAsRead.js
```diff
@@ -30,12 +30,13 @@ const processQueue = (auth: Auth) => {
     return;
   }
 
-  messagesFlags(auth, unsentMessageIds, 'add', 'read');
-  unsentMessageIds = [];
+  messagesFlags(auth, unackedMessageIds, 'add', 'read').then(success => {
+    unackedMessageIds = unackedMessageIds.filter(id => !success.messages.includes(id));
```
gnprice, line 34: This is quadratic -- we go through the list `unackedMessageIds`, and for each of them we go through the list `success.messages` to look for the ID.

If there are thousands of IDs in the list, that could be significant. I think it's a bit tricky to make that happen -- because we only fetch 100 messages at a time it's tricky to have thousands of messages in the message list at once without having already scrolled through them, and if we're successfully fetching more as you scroll then we're probably successfully marking them off as read -- but it can happen.

And fortunately this only takes a line to fix: use `new Set` to make a JS `Set` from `success.messages`, and replace the `.includes` with checking that.

gnprice:
Thanks @WesleyAC !

Two comments I'd like to see followup fixes for: one just above, and https://github.com/zulip/zulip-mobile/pull/4620#discussion_r612044658 in the earlier thread.

Issue #4614, opened 2021-04-07
Avatars pixelated in user view


WesleyAC:
If you click on an image of a user, you see a large image of their avatar, along with some info about them. That image is pixelated.

This probably has a solution like the other problems like this, which are described in #4305.

chrisbobbe:
Interesting, thanks for the report! Are you seeing this pixelation on all avatars, or just some? Shortly after merging work to prevent pixelation, I did identify an exception that that work didn't cover; it'd be good for me to load that back into my brain and write it down. But if it's happening to all avatars, that's something else.

WesleyAC:
It's happening on every avatar I've looked at.

WesleyAC:
Oh, wait, nevermind, your avatar on CZO looks fine... Will investigate further.

chrisbobbe:
> Shortly after merging work to prevent pixelation, I did identify an exception that that work didn't cover; it'd be good for me to load that back into my brain and write it down.

Interesting; this seems to be different from the exception I had in mind.

(That one was, when we use the `/avatar/{user_id}` redirect endpoint, we can't express the size we want. You can read more about `/avatar/{user_id}` in `FallbackAvatarURL` which we construct at the edge for [crunchy shell](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md), and on the point about `user_avatar_url_field_optional` at https://zulipchat.com/api/register-queue. So in cases where we use that endpoint, avatars might be pixelated even when we have the right idea of what size we want to use. In practice, though, the current implementation means we use that endpoint for long-term idle users (though this is a server implementation detail that may change without notice), and naturally people don't run into those users often.)

[...]

Issue #4305, opened 2020-11-12
Avatars look pixelated in some parts of UI, on high-density displays


gnprice:
(Originally diagnosed this at https://github.com/zulip/zulip-mobile/pull/4230#discussion_r521713861 , while reviewing the big avatar refactor in #4230 earlier today. But I've been noticing these symptoms for a long time.)

Repro:
* Take a device with a high-density display, like about 400ppi or higher.
  * Any flagship phone from the last few years will do; I see this on my Pixel 4 with its 444ppi display.
  * Or fire up an emulation of such a device, on a computer with a high-resolution monitor. This can make the effect easier to see, because the physical pixels are bigger.
* Optionally, to make the effect stronger: go to the device's display settings and turn up the display size to max. (E.g., Settings > Display > Display size > Larger, on stock Android 11.)
* Open the app and:
  * go to the PM conversations screen and look at the user avatars there;
  * or to "Create group" or "Search" from there, and look at the user avatars on the user list/search screen there;
  * or just look at your avatar in the lower-right corner from the main screen of the app;
  * or tap that one and look at your avatar in a large size at the top of the your-account screen.
* Expected: each avatar shows up nice and crisp.
* Actual: many avatars show up as some combination of pixelated, jagged, and blurry.
  * In short, they're low-res images which have been scaled up for display. The fact they're sometimes jagged, and aren't just blatantly pixelated or blurry, is presumably due to some sophisticated visual scaling algorithm that some layer in the stack has brought to bear to valiantly try to scale them up as cleanly as possible.

The exact symptoms are different between avatars that were uploaded to the Zulip server and avatars hosted on Gravatar; Gravatar will serve an avatar at any size, but Zulip provides avatars at just two discrete sizes, which can either aggravate this bug or mask it. They're also different between Gravatars on an org with hidden emails (like chat.zulip.org) vs. Gravatars on an org without hidden emails.

On a test account with an autogenerated Gravatar, here's what it looks like in that lower-right icon (as seen on an emulated Pixel 3, because using the emulator on my desktop makes cropped screenshots easy; and this is with default display settings):
![image](https://user-images.githubusercontent.com/28173/98883856-bc919b80-2443-11eb-9a83-c487ceb04a3c.png)
Yuck. That's pretty much how it looks on my actual Pixel 4, too; though it's a little less conspicuous there, just because the physical pixels are smaller.

Here's how it should look -- concretely, how it looks with a quick hack to fix this issue for just that widget:
[...]

PR #4625, opened 2021-04-08
Fix more avatar pixelation issues.

chrisbobbe:
("More", since we fixed others in #4305.)

(cc @timabbott if you're interested; thanks for your help in the [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4614.20Avatars.20pixelated.20in.20user.20view/near/1156704).)

I think we can still end up with a pixelated image if we use the `/avatar/{user_id}` / `/avatar/{user_id}/medium` redirect and the image we end up with happens to be from Gravatar; I'm not aware that this endpoint forwards the default-vs.-medium size preference on to Gravatar, and there doesn't seem to be a way to pass an integer size, like we do when getting a Gravatar image directly.

But #4614 was pretty high-impact on zulipchat.com, so I'm glad to have that fixed, and to have the other, lower-impact improvement that follows in the branch, for using `/avatar/{user_id}/medium` at all.

Ideally, we'd show a pixelated image during the load time of a larger image, if possible; see the results of my brief exploration of this in [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4614.20Avatars.20pixelated.20in.20user.20view/near/1157651), that unfortunately didn't come up with anything.

Fixes: #4614

### DIFF
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -398,10 +398,10 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      const match = new RegExp(/(\w+)\.png/).exec(result.pathname);
-      if (match !== null) {
-        result.pathname = result.pathname.replace(match[0], `${match[1]}-medium.png`);
-      }
+      result.pathname = result.pathname.replace(
+        new RegExp(/(\w+)\.png/),
+        (_, p1) => `${p1}-medium.png`,
+      );
     }
     return result;
   }
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -398,7 +398,12 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      result.pathname = result.pathname.replace(/\.png$/, '-medium.png');
+      result.pathname = result.pathname.replace(
+        // `.png` is optional: s3 uploads don't use it, local ones do.
+        // See TODO in zulip/zulip@c03615b98.
+        /(?:\.png)?$/,
+        '-medium.png',
+      );
     }
     return result;
   }
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -398,7 +398,7 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      const match = new RegExp(/(\w+)\.png/g).exec(result.pathname);
+      const match = new RegExp(/(\w+)\.png/).exec(result.pathname);
       if (match !== null) {
         result.pathname = result.pathname.replace(match[0], `${match[1]}-medium.png`);
       }
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -398,10 +398,7 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      result.pathname = result.pathname.replace(
-        new RegExp(/(\w+)\.png/),
-        (_, p1) => `${p1}-medium.png`,
-      );
+      result.pathname = result.pathname.replace(/(\w+)\.png/, (_, p1) => `${p1}-medium.png`);
     }
     return result;
   }
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -398,7 +398,7 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      result.pathname = result.pathname.replace(/(\w+)\.png/, (_, p1) => `${p1}-medium.png`);
+      result.pathname = result.pathname.replace(/\.png$/, '-medium.png');
     }
     return result;
   }
--- a/src/utils/avatar.js
+++ b/src/utils/avatar.js
@@ -8,6 +8,15 @@ import { ensureUnreachable, type UserId } from '../types';
 import { isUrlAbsolute, isUrlPathAbsolute } from './url';
 
 /**
+ * Pixel dimensions of different size choices we have (they're all
+ * square) when requesting an uploaded avatar.
+ */
+// DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+export const DEFAULT_UPLOAD_SIZE_PX = 100;
+// MEDIUM_AVATAR_SIZE in zerver/lib/upload.py.
+export const MEDIUM_UPLOAD_SIZE_PX = 500;
+
+/**
  * A way to get a standard avatar URL, or a sized one if available
  *
  * This class is abstract. Only instantiate its subclasses.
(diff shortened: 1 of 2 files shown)


src/utils/avatar.js
```diff
@@ -284,7 +286,18 @@ export class FallbackAvatarURL extends AvatarURL {
       this._standardUrl = new URL(this._standardUrl);
     }
 
-    return this._standardUrl;
+    let result: URL = this._standardUrl;
+    if (sizePhysicalPx > 100) {
```
WesleyAC, line 290: Can you leave a comment explaining where `100` comes from?

src/utils/avatar.js
```diff
@@ -389,7 +389,12 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      result.pathname = result.pathname.replace(/\.png$/, '-medium.png');
+      result.pathname = result.pathname.replace(
+        // `.png` is optional: s3 uploads don't use it, local ones do.
+        // See TODO in zulip/zulip@c03615b98.
+        /(?:\.png)?$/,
```
WesleyAC, line 408: Why the non-capturing group here? Is there a reason `/(\.png)?$/` wouldn't work?

WesleyAC (commented):
A nit on the commit message:

> We don't expect there to be more than one occurrence of `.png`.

is I think a sort of weak statement — given that we're looking for end-of-line, and this is a URL, we kind of know that there will only be one. Unless, of course, we have a bug somewhere that causes a literal newline character in a URL. I think `Since the regex matches end-of-line, there should be at most one match` would be clearer.

src/utils/avatar.js
chrisbobbe: I tend to use non-capturing groups when I don't intend to read the captured value on the match object I get from executing the pattern.
- It has seemed neater to me to get rid of unused (and often never-useful) bits of data there.
- Also, the signal that we don't intend to use the captured value can help make patterns easier to read, I think. When I use a capturing group, I'm foreshadowing that I'll care about the value later on; when I use a non-capturing group, its role is clearer as just an instrument to say "treat this sequence of items together". I guess I'd prefer if the capturing-group syntax were the one that required extra characters. 🙂

What do you think?

src/utils/avatar.js
chrisbobbe: Ah, good point—this was copied from `UploadedAvatarURL#get`, which grew around some lines first added in aec2ff255, I think, without an implementation comment there—and I never added one.

chrisbobbe:
Thanks for the review, @WesleyAC! Revision pushed.

src/utils/avatar.js
chrisbobbe: Er, rather, I addressed this in a commit that comes after this one; hopefully that's fine. So, there was an existing lack (the one in `UploadedAvatarURL`), I propagated it just a bit, here, then fixed both.

src/utils/avatar.js
```diff
@@ -391,7 +391,7 @@ export class UploadedAvatarURL extends AvatarURL {
 
       const match = /\.png$/.exec(result.pathname);
       if (match !== null) {
-        result.pathname = result.pathname.replace(match[0], '-medium.png');
+        result.pathname = result.pathname.replace(/\.png$/, '-medium.png');
```
gnprice: As the commit message suggests, this potentially does the wrong thing before this commit. I believe none of the preceding commits change when that can happen, but it's a bit tricky to reason about.

How about putting the next commit:

    UploadedAvatarURL [nfc]: Remove `if (match !== null)` check.
    
    `.replace` already has the behavior of doing nothing if a match
    isn't found.

at the beginning? That will fix this bug from the start, and simplify reasoning about the rest.

src/utils/avatar.js
```diff
@@ -389,7 +389,7 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      const match = new RegExp(/\.png$/g).exec(result.pathname);
```
gnprice: UploadedAvatarURL: Remove global flag in regex.
    
    Since the regex matches end-of-line, there should be at most one
    match.

Yeah. Also even if there were several matches (like with the original pattern), we only call this `RegExp` object once before discarding it -- so it'll return the first match then anyway, and we never go back to ask it about a next match.

Quick demo in a browser console:
```
> r = /(\w+)\.png/g
/(\w+)\.png/g

> r.exec("foo.png/bar.png")
(2) ["foo.png", "foo", index: 0, input: "foo.png/bar.png", groups: undefined]

> r.exec("foo.png/bar.png")
(2) ["bar.png", "bar", index: 8, input: "foo.png/bar.png", groups: undefined]

> r.exec("foo.png/bar.png")
null
```

gnprice: To be clear, I think the existing pattern's behavior isn't a *live* bug, only a latent one -- I believe Zulip avatar URLs all follow highly structured shapes where the only place the pathname can match this is at the end. But certainly good to fix! As well as to simplify the logic, and to fix the original motivating issue here.

gnprice (approved):
Thanks @chrisbobbe ! Looks good modulo small comments below.

src/utils/avatar.js
chrisbobbe: Yeah—I spotted that we could remove the `g` because of its different behavior when used as a pattern in `.replace`. IIRC/IIUC it was trying to replace multiple things at once; I didn't investigate super closely after realizing it should just go away in any case.

chrisbobbe: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
> To perform a global search and replace, include the g switch in the regular expression.

src/utils/avatar.js
chrisbobbe: Sure; done. There was a bit of a wrinkle and I put the remove-`g` commit before that one, if that's OK.

chrisbobbe:
Thanks for the review, @gnprice! Revision pushed, with the change you mentioned at https://github.com/zulip/zulip-mobile/pull/4625#discussion_r610222635 (note a bit of a wrinkle where I put the `g`-removing commit before the `if (match !== null)`-removing one, hopefully that's OK and makes things clearer on net).

I also addressed my own comment at https://github.com/zulip/zulip-mobile/pull/4625#discussion_r610096706 by moving earlier the commit that adds `DEFAULT_UPLOAD_SIZE_PX` and `MEDIUM_UPLOAD_SIZE_PX`.

chrisbobbe:
(that last force-push just now was just a small commit-message edit)

src/utils/avatar.js
```diff
@@ -398,10 +398,10 @@ export class UploadedAvatarURL extends AvatarURL {
          */
       result = new URL(this._standardUrl);
 
-      const match = new RegExp(/(\w+)\.png/).exec(result.pathname);
-      if (match !== null) {
-        result.pathname = result.pathname.replace(match[0], `${match[1]}-medium.png`);
-      }
+      result.pathname = result.pathname.replace(
+        new RegExp(/(\w+)\.png/),
+        (_, p1) => `${p1}-medium.png`,
```
gnprice: Ah I see, the needs of writing this `replace` call when the replacement is still dependent on the match make this more complicated than I'd quite thought through when I made that suggestion :sweat_smile: 

This revision looks good, though. I'll go ahead and merge.

chrisbobbe:
Thanks for the review! :)

PR #4627, opened 2021-04-09
event: Log error but don't throw, if no active account at eventToAction.

gnprice:
Chris asked if adding this error was still a good idea:
  https://github.com/zulip/zulip-mobile/pull/4581#discussion_r607315106
as most event types will get by OK in this case.  It's a bug if we
hit this case, but perhaps we'd rather learn of such a bug without
causing an event to fail that didn't have to.

After merging the PR but while writing a reply to the comment, I
realized that there's a useful middle ground: we can log an error
without actually immediately failing.


chrisbobbe:
Thanks! Merged.

Issue #4628, opened 2021-04-09
Compose Box shifts up and down when menu is toggled


im-adithya:
## Description
Whenever the menu is collapsed, the compose box shifts a bit more up than required and falls down.

## Video
https://user-images.githubusercontent.com/64399555/114175586-b999c900-9957-11eb-8d40-3cabd936209f.mp4

PR #4629, opened 2021-04-09
ComposeBox: Add multiLine to state object.

im-adithya:
## Description
This adds multiLine which keeps track of the lines in composebox
and changes the multiline prop of the Input field only after the
menu is completely collapsed, i.e., after 300 milliseconds as it
is an Animated Component.

## Issue
Fixes: #4628

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -101,6 +101,7 @@ type State = {|
   message: string,
   height: number,
   selection: InputSelection,
+  multiLine: boolean,
 |};
 
 function randomInt(min, max) {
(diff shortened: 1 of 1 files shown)


src/compose/ComposeBox.js
```diff
@@ -221,6 +223,15 @@ class ComposeBox extends PureComponent<Props, State> {
     this.setState(({ isMenuExpanded }) => ({
       isMenuExpanded: !isMenuExpanded,
     }));
+    setTimeout(
```
WesleyAC, line 226: Instead of using `setTimeout` here, we should pass a callback into `start()` in `AnimatedComponent`.

WesleyAC (changes requested):
Thanks! This is good to fix, but I don't think it's the right approach. I think we should pass a callback that does this into the `.start()` call in `AnimatedComponent` instead. Feel free to ask on CZO if you have questions about how to go about this.

chrisbobbe:
Closing as superseded by #5635.

PR #4632, opened 2021-04-09
i18n: Rename id_ID -> id.

gnprice:
This reflects a change we just made in Transifex:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Using.20device.20language/near/1158948

Tested manually: switching to Indonesian and back, and also switching to Indonesian before the upgrade and things working fine (and still being in Indonesian) after the upgrade.


WesleyAC:
LGTM!

PR #4634, opened 2021-04-09
ensure-coreutils: Work with arbitrary Homebrew prefix.

gnprice:
In particular this lets it work with a default Homebrew install on
macOS ARM, which uses /opt/homebrew (hooray!) rather than their
unfortunate older choice of taking over /usr/local.

Debugging thread here:
  https://chat.zulip.org/#narrow/stream/48-mobile/topic/Error.20running.20zulip-mobile.20app/near/1158901

### DIFF
--- a/tools/lib/ensure-coreutils.sh
+++ b/tools/lib/ensure-coreutils.sh
@@ -20,6 +20,7 @@
 # it installed but just not in their PATH.  We write our scripts
 # for a GNU environment, so we bring it into the PATH.
 
+# Check, silently, for a working coreutils on the PATH.
 check_coreutils() {
     # Check a couple of commands for GNU-style --help and --version,
     # which macOS's default BSD-based implementations don't understand.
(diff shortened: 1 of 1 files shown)


tools/lib/ensure-coreutils.sh
```diff
@@ -18,33 +18,75 @@
...
 check_coreutils() {
     # Check a couple of commands for GNU-style --help and --version,
     # which macOS's default BSD-based implementations don't understand.
     fmt --help >/dev/null 2>&1
     readlink --version 2>&1 | grep -q GNU
 }
 
+# Either get Homebrew's coreutils on the PATH, or error out.
+try_homebrew() {
+    local homebrew_prefix="$1"
```
WesleyAC, line 31: Can you add a `#!/usr/bin/env bash`, since `local` is a non-POSIX feature?

WesleyAC:
Looks good other than one nit :)

gnprice:
Thanks @WesleyAC for the review!

I've merged with a variation on your suggestion -- I don't want to put a shebang line here because this isn't an executable script, but I've added a ShellCheck ["shell" directive](https://github.com/koalaman/shellcheck/wiki/Directive#shell) which should be just as informative for human readers.

Issue #3244, opened 2018-12-28
Use new support for muting topics by stream ID when available


timabbott:
Related to https://github.com/zulip/zulip/issues/11126 and https://github.com/zulip/zulip/issues/11136, we now (in master) support sending a `stream_id` in the API for muting/unmuting topics.  We can't take advantage of this in mobile yet since no clients have that code, but I figured it's worth opening the issue for migrating.  

I expect several similar changes to be integrated in the server codebase in the coming weeks to support IDs in the API where possible.  We'll want to come up with a reasonable general strategy for detecting this sort of capability; possibly that's just a server version number check in the app codebase.  No urgency to migrate, in any case.

borisyankov:
I think the optimal way is to do that once the server supports 'everything by id' and provide this as a value on `server_settings`.

The actual benefit to clients (including mobile) is once we can solely rely on IDs. 

One of the major things would be narrowing by IDs. Considering this might be a different format, we might include @gnprice suggestion for more 'readable' narrow format that I remember we discussed with you too, Tim (not 100% sure about that part)

timabbott:
Yeah, that's a reasonable plan.  Narrowing by IDs is the main one that hasn't been started, though I think @showell has talked about doing it a few times.  We're probably merging "Send messages to PM recipients by IDs" in the next week or two.

timabbott:
Just a small update that I'm adjusting the API documentation in https://github.com/zulip/zulip/issues/11136 to note that the old format is deprecated.  Presumably we can't remove it until some time after mobile adds support for it, though.

chrisbobbe:
It looks like support for using stream IDs was https://github.com/zulip/zulip/commit/c35afb942, released in 2.0.0. There's been a [mention](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/compatibility.20documentation/near/1174950) that as of March, dropping support for 1.9.x and older might affect 2-4% of users.

Are we OK to switch completely over to using stream IDs for this feature, or should we check the server version and only use stream IDs if the server is at least 2.0.0?

gnprice:
Looks like this was accidentally closed by #4635 -- the original revision of that PR would have resolved this, but in the end it didn't, because https://github.com/zulip/zulip-mobile/pull/4635#discussion_r652455585 .

I'll take care of this in an upcoming followup to #5192, as part of relying on server 2.0+.

PR #4194, opened 2020-07-17
home screen: Add ability to mark messages in bulk, mute topics.

agrawal-d:
Fixes: #3175

preview

Issue #3175, opened 2018-11-29
feature, home screen: Add ability to mark messages in bulk, mute topics etc


jainkuniya:
Issue: 
> to feel fully caught-up, I have to open topics that I'm not interested in, and mark them as read. It'd be nice if I can long press on the topic and either mute them or mark them as read. Making this a bulk action, where I can select multiple topics and mark them as read would be even better!

Solution:
>I'm imagining is that you long-press, and the result is the screen enters a "select several" mode. Like this:
https://material.io/design/interaction/selection.html

Read more at https://chat.zulip.org/#narrow/stream/48-mobile/topic/muting.20topics.20or.20mark.20as.20read

borisyankov:
The upstream issue is here:
https://github.com/facebook/react-native/pull/22025

This affects our unread screen `UnreadCards`, all users screen and user picker screen (that use the `UserList` component)

chrisbobbe:
A user in the Recurse realm, over [here](https://recurse.zulipchat.com/#narrow/stream/102312-consciousness/topic/Victoria.20Grey.20Mitchell/near/203644976) (only accessible to realm members, like @gnprice, Tim, and me), reports that it's not clear that you can mute topics/streams.

In a high-participation realm like the Recurse Center (or [CZO](chat.zulip.org), I suppose), it's especially important to be able to easily opt out of content you're not interested in. The user identifies the difficulty of muting conversations as a main reason they don't like or recommend Zulip. As an additional data point, I've often found the RC realm overwhelming; it's a common experience to need to quite actively weed out unwanted topics and streams just to get down to a manageable volume of content you can reasonably consume. (This may indeed be a factor in my recently not being very active on the Recurse realm!)

Another user points out to this user that you can mute topics from the message list while looking at a stream, by long-pressing the topic header—but that necessarily means marking some messages as read, which is unfortunate. It looks like that part is also covered in this issue.

From that discussion Vishwesh links to:

@gnprice [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/muting.20topics.20or.20mark.20as.20read/near/662438):
> That sounds like a good UI feature to me, though it'll require some iteration and playing-with to get to a version that's really good.

PR #4635, opened 2021-04-10
Add mark as read option in TopicActionSheet and Modify TopicActionSheet to use `stream_id`.

AkashDhiman:
## Work Done:
- Renamed `HeaderActionSheet` to `TopicActionSheet`
- Added **mark as read** option in the `TopicActionSheet`
- Wrote test cases to ensure it's visible only when the topic is not actually read.
- Modified `TopicActionSheet` code to use stream_id instead of stream names, refactored test cases and function calls to `showTopicActionSheet` accordingly.

This PR is an alternative to #4194 for marking topics as read.

For more context: https://chat.zulip.org/#narrow/stream/48-mobile/topic/mark.20as.20read

mark-as-read

Related: #3918
Related: #3175
Fixes: #3244

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -759,6 +759,7 @@ export const backgroundData: BackgroundData = deepFreeze({
   streams: getStreamsById(baseReduxState),
   streamsByName: getStreamsByName(baseReduxState),
   subscriptions: [],
+  unread: baseReduxState.unread,
   theme: 'default',
   twentyFourHourTime: false,
 });
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -39,6 +39,7 @@ import rootReducer from '../../boot/reducers';
 import { authOfAccount } from '../../account/accountMisc';
 import { HOME_NARROW } from '../../utils/narrow';
 import type { BackgroundData } from '../../webview/MessageList';
+import { getStreamsById } from '../../selectors';
 
 /* ========================================================================
  * Utilities
@@ -755,6 +756,7 @@ export const backgroundData: BackgroundData = deepFreeze({
   mute: [],
   mutedUsers: Immutable.Map(),
   ownUser: selfUser,
+  streams: getStreamsById(baseReduxState),
   subscriptions: [],
   theme: 'default',
   twentyFourHourTime: false,
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -39,7 +39,7 @@ import rootReducer from '../../boot/reducers';
 import { authOfAccount } from '../../account/accountMisc';
 import { HOME_NARROW } from '../../utils/narrow';
 import type { BackgroundData } from '../../webview/MessageList';
-import { getStreamsById } from '../../selectors';
+import { getStreamsById, getStreamsByName } from '../../selectors';
 
 /* ========================================================================
  * Utilities
@@ -757,6 +757,7 @@ export const backgroundData: BackgroundData = deepFreeze({
   mutedUsers: Immutable.Map(),
   ownUser: selfUser,
   streams: getStreamsById(baseReduxState),
+  streamsByName: getStreamsByName(baseReduxState),
   subscriptions: [],
   theme: 'default',
   twentyFourHourTime: false,
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -5,6 +5,8 @@ import { streamNameOfStreamMessage } from '../../utils/recipient';
 
 import * as eg from '../../__tests__/lib/exampleData';
 import { constructMessageActionButtons, constructTopicActionButtons } from '../messageActionSheet';
+import { reducer } from '../../unread/unreadModel';
+import { initialState } from '../../unread/__tests__/unread-testlib';
 
 const buttonTitles = buttons => buttons.map(button => button.title);
 
@@ -48,6 +50,34 @@ describe('constructTopicActionButtons', () => {
   const streamName = streamNameOfStreamMessage(streamMessage);
   const topic = streamMessage.subject;
   const streamId = streamMessage.stream_id;
+
+  const baseState = (() => {
+    const r = (state, action) => reducer(state, action, eg.plusReduxState);
+    let state = initialState;
+    state = r(state, eg.mkActionEventNewMessage(streamMessage));
+    return state;
+  })();
+
+  test('show mark as read if topic is unread', () => {
+    const unread = baseState;
+    const buttons = constructTopicActionButtons({
+      backgroundData: { ...eg.backgroundData, unread },
+      streamName,
+      streamId,
+      topic,
+    });
+    expect(buttonTitles(buttons)).toContain('Mark topic as read');
+  });
+
+  test('do not show mark as read if topic is read', () => {
+    const buttons = constructTopicActionButtons({
+      backgroundData: eg.backgroundData,
+      streamName,
+      streamId,
+      topic,
+    });
+    expect(buttonTitles(buttons)).not.toContain('Mark topic as read');
+  });
 
   test('show Unmute topic option if topic is muted', () => {
     const mute = deepFreeze([[streamName, topic]]);
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -48,7 +48,7 @@ describe('constructTopicActionButtons', () => {
     const mute = deepFreeze([['electron issues', 'issue #556']]);
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute },
-      stream: 'electron issues',
+      streamName: 'electron issues',
       topic: 'issue #556',
     });
     expect(buttonTitles(buttons)).toContain('Unmute topic');
@@ -57,7 +57,7 @@ describe('constructTopicActionButtons', () => {
   test('show mute topic option if topic is not muted', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute: [] },
-      stream: streamNameOfStreamMessage(eg.streamMessage()),
+      streamName: streamNameOfStreamMessage(eg.streamMessage()),
       topic: eg.streamMessage().subject,
     });
     expect(buttonTitles(buttons)).toContain('Mute topic');
@@ -67,7 +67,7 @@ describe('constructTopicActionButtons', () => {
     const subscriptions = [{ ...eg.subscription, in_home_view: false }];
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
-      stream: streamNameOfStreamMessage(eg.streamMessage()),
+      streamName: streamNameOfStreamMessage(eg.streamMessage()),
       topic: eg.streamMessage().subject,
     });
     expect(buttonTitles(buttons)).toContain('Unmute stream');
@@ -77,7 +77,7 @@ describe('constructTopicActionButtons', () => {
     const subscriptions = [{ ...eg.subscription, in_home_view: true }];
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
-      stream: streamNameOfStreamMessage(eg.streamMessage()),
+      streamName: streamNameOfStreamMessage(eg.streamMessage()),
       topic: eg.streamMessage().subject,
     });
     expect(buttonTitles(buttons)).toContain('Mute stream');
@@ -87,7 +87,7 @@ describe('constructTopicActionButtons', () => {
     const ownUser = { ...eg.selfUser, is_admin: true };
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, ownUser },
-      stream: streamNameOfStreamMessage(eg.streamMessage()),
+      streamName: streamNameOfStreamMessage(eg.streamMessage()),
       topic: eg.streamMessage().subject,
     });
     expect(buttonTitles(buttons)).toContain('Delete topic');
@@ -96,7 +96,7 @@ describe('constructTopicActionButtons', () => {
   test('do not show delete topic option if current user is not an admin', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: eg.backgroundData,
-      stream: streamNameOfStreamMessage(eg.streamMessage()),
+      streamName: streamNameOfStreamMessage(eg.streamMessage()),
       topic: eg.streamMessage().subject,
     });
     expect(buttonTitles(buttons)).not.toContain('Delete topic');
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -1,7 +1,6 @@
 // @flow strict-local
 import deepFreeze from 'deep-freeze';
 import { HOME_NARROW } from '../../utils/narrow';
-import { streamNameOfStreamMessage } from '../../utils/recipient';
 
 import * as eg from '../../__tests__/lib/exampleData';
 import { constructMessageActionButtons, constructTopicActionButtons } from '../messageActionSheet';
@@ -46,10 +45,11 @@ describe('constructActionButtons', () => {
 });
 
 describe('constructTopicActionButtons', () => {
-  const streamMessage = eg.streamMessage();
-  const streamName = streamNameOfStreamMessage(streamMessage);
+  const stream = eg.makeStream();
+  const streamMessage = eg.streamMessage({ stream });
   const topic = streamMessage.subject;
   const streamId = streamMessage.stream_id;
+  const streams = deepFreeze(new Map([[stream.stream_id, stream]]));
 
   const baseState = (() => {
     const r = (state, action) => reducer(state, action, eg.plusReduxState);
@@ -61,8 +61,7 @@ describe('constructTopicActionButtons', () => {
   test('show mark as read if topic is unread', () => {
     const unread = baseState;
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, unread },
-      streamName,
+      backgroundData: { ...eg.backgroundData, streams, unread },
       streamId,
       topic,
     });
@@ -71,8 +70,7 @@ describe('constructTopicActionButtons', () => {
 
   test('do not show mark as read if topic is read', () => {
     const buttons = constructTopicActionButtons({
-      backgroundData: eg.backgroundData,
-      streamName,
+      backgroundData: { ...eg.backgroundData, streams },
       streamId,
       topic,
     });
@@ -80,10 +78,9 @@ describe('constructTopicActionButtons', () => {
   });
 
   test('show Unmute topic option if topic is muted', () => {
-    const mute = deepFreeze([[streamName, topic]]);
+    const mute = deepFreeze([[stream.name, topic]]);
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, mute },
-      streamName,
+      backgroundData: { ...eg.backgroundData, streams, mute },
       streamId,
       topic,
     });
@@ -92,8 +89,7 @@ describe('constructTopicActionButtons', () => {
 
   test('show mute topic option if topic is not muted', () => {
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, mute: [] },
-      streamName,
+      backgroundData: { ...eg.backgroundData, streams, mute: [] },
       streamId,
       topic,
     });
@@ -101,10 +97,9 @@ describe('constructTopicActionButtons', () => {
   });
 
   test('show Unmute stream option if stream is not in home view', () => {
-    const subscriptions = [{ ...eg.subscription, in_home_view: false }];
+    const subscriptions = [{ ...eg.subscription, in_home_view: false, ...stream }];
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, subscriptions },
-      streamName,
+      backgroundData: { ...eg.backgroundData, subscriptions, streams },
       streamId,
       topic,
     });
@@ -112,10 +107,9 @@ describe('constructTopicActionButtons', () => {
   });
 
   test('show mute stream option if stream is in home view', () => {
-    const subscriptions = [{ ...eg.subscription, in_home_view: true }];
+    const subscriptions = [{ ...eg.subscription, in_home_view: true, ...stream }];
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, subscriptions },
-      streamName,
+      backgroundData: { ...eg.backgroundData, subscriptions, streams },
       streamId,
       topic,
     });
@@ -125,8 +119,7 @@ describe('constructTopicActionButtons', () => {
   test('show delete topic option if current user is an admin', () => {
     const ownUser = { ...eg.selfUser, is_admin: true };
     const buttons = constructTopicActionButtons({
-      backgroundData: { ...eg.backgroundData, ownUser },
-      streamName,
+      backgroundData: { ...eg.backgroundData, ownUser, streams },
       streamId,
       topic,
     });
@@ -135,8 +128,7 @@ describe('constructTopicActionButtons', () => {
 
   test('do not show delete topic option if current user is not an admin', () => {
     const buttons = constructTopicActionButtons({
-      backgroundData: eg.backgroundData,
-      streamName,
+      backgroundData: { ...eg.backgroundData, streams },
       streamId,
       topic,
     });
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -47,12 +47,14 @@ describe('constructTopicActionButtons', () => {
   const streamMessage = eg.streamMessage();
   const streamName = streamNameOfStreamMessage(streamMessage);
   const topic = streamMessage.subject;
+  const streamId = streamMessage.stream_id;
 
   test('show Unmute topic option if topic is muted', () => {
     const mute = deepFreeze([[streamName, topic]]);
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute },
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).toContain('Unmute topic');
@@ -62,6 +64,7 @@ describe('constructTopicActionButtons', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute: [] },
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).toContain('Mute topic');
@@ -72,6 +75,7 @@ describe('constructTopicActionButtons', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).toContain('Unmute stream');
@@ -82,6 +86,7 @@ describe('constructTopicActionButtons', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).toContain('Mute stream');
@@ -92,6 +97,7 @@ describe('constructTopicActionButtons', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, ownUser },
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).toContain('Delete topic');
@@ -101,6 +107,7 @@ describe('constructTopicActionButtons', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: eg.backgroundData,
       streamName,
+      streamId,
       topic,
     });
     expect(buttonTitles(buttons)).not.toContain('Delete topic');
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -44,12 +44,16 @@ describe('constructActionButtons', () => {
 });
 
 describe('constructTopicActionButtons', () => {
+  const streamMessage = eg.streamMessage();
+  const streamName = streamNameOfStreamMessage(streamMessage);
+  const topic = streamMessage.subject;
+
   test('show Unmute topic option if topic is muted', () => {
-    const mute = deepFreeze([['electron issues', 'issue #556']]);
+    const mute = deepFreeze([[streamName, topic]]);
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute },
-      streamName: 'electron issues',
-      topic: 'issue #556',
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).toContain('Unmute topic');
   });
@@ -57,8 +61,8 @@ describe('constructTopicActionButtons', () => {
   test('show mute topic option if topic is not muted', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, mute: [] },
-      streamName: streamNameOfStreamMessage(eg.streamMessage()),
-      topic: eg.streamMessage().subject,
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).toContain('Mute topic');
   });
@@ -67,8 +71,8 @@ describe('constructTopicActionButtons', () => {
     const subscriptions = [{ ...eg.subscription, in_home_view: false }];
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
-      streamName: streamNameOfStreamMessage(eg.streamMessage()),
-      topic: eg.streamMessage().subject,
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).toContain('Unmute stream');
   });
@@ -77,8 +81,8 @@ describe('constructTopicActionButtons', () => {
     const subscriptions = [{ ...eg.subscription, in_home_view: true }];
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, subscriptions },
-      streamName: streamNameOfStreamMessage(eg.streamMessage()),
-      topic: eg.streamMessage().subject,
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).toContain('Mute stream');
   });
@@ -87,8 +91,8 @@ describe('constructTopicActionButtons', () => {
     const ownUser = { ...eg.selfUser, is_admin: true };
     const buttons = constructTopicActionButtons({
       backgroundData: { ...eg.backgroundData, ownUser },
-      streamName: streamNameOfStreamMessage(eg.streamMessage()),
-      topic: eg.streamMessage().subject,
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).toContain('Delete topic');
   });
@@ -96,8 +100,8 @@ describe('constructTopicActionButtons', () => {
   test('do not show delete topic option if current user is not an admin', () => {
     const buttons = constructTopicActionButtons({
       backgroundData: eg.backgroundData,
-      streamName: streamNameOfStreamMessage(eg.streamMessage()),
-      topic: eg.streamMessage().subject,
+      streamName,
+      topic,
     });
     expect(buttonTitles(buttons)).not.toContain('Delete topic');
   });
--- a/src/message/__tests__/messageActionSheet-test.js
+++ b/src/message/__tests__/messageActionSheet-test.js
@@ -4,7 +4,7 @@ import { HOME_NARROW } from '../../utils/narrow';
 import { streamNameOfStreamMessage } from '../../utils/recipient';
 
 import * as eg from '../../__tests__/lib/exampleData';
-import { constructMessageActionButtons, constructHeaderActionButtons } from '../messageActionSheet';
+import { constructMessageActionButtons, constructTopicActionButtons } from '../messageActionSheet';
 
 const buttonTitles = buttons => buttons.map(button => button.title);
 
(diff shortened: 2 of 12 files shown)


src/message/messageActionSheet.js
```diff
@@ -252,10 +261,16 @@ export const constructTopicActionButtons = ({
     buttons.push(muteTopic);
   }
   const sub = subscriptions.find(x => x.name === stream);
-  if (sub && !sub.in_home_view) {
-    buttons.push(unmuteStream);
-  } else {
-    buttons.push(muteStream);
+  if (sub) {
+    const unreadCount = unreadStreams.get(sub.stream_id)?.get(topic)?.size;
+    if (unreadCount !== undefined && unreadCount > 0) {
```
WesleyAC, line 271: Either `if (unreadCount) {` or `if (unreadCount && unreadCount > 0) {` should work here. I'd go with the second, even though it's redundant (well, unless we have a negative number of unreads, in which case we have a second problem...), just since it reads more clearly to me.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -58,6 +58,34 @@ describe('constructActionButtons', () => {
 });
 
 describe('constructTopicActionButtons', () => {
+  test('show mark as read if topic is unread', () => {
```
WesleyAC, line 60: You should add the tests in the same commit that you add the code in, in general.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -58,6 +58,34 @@ describe('constructActionButtons', () => {
...
+      topic: 'test topic',
+    });
+    expect(buttonTitles(buttons)).toContain('Mark topic as read');
+  });
+
+  test('do not show mark as read if topic is read', () => {
+    const stream = eg.makeStream({ name: 'test stream', streamId: 123 });
+    const subscriptions = [{ ...eg.subscription, ...stream }];
+    const unreadStreams = eg.makeUnreadStreamsState([
+      { streamId: 123, topics: [{ topicName: 'unread topic', messageIds: [1, 2] }] },
```
WesleyAC, line 78: I don't think there's a reason to include this stream in `unreadStreams`, or even to override `unreadStreams` at all, since this test doesn't look at it.

src/message/messageActionSheet.js
```diff
@@ -38,8 +38,8 @@ export type ShowActionSheetWithOptions = (
 
 type TopicArgs = {
   auth: Auth,
-  stream: string,
   topic: string,
+  streamId?: number,
```
WesleyAC, line 42: This shouldn't be optional, we should throw an error before we construct this if there's no stream ID.

src/message/messageActionSheet.js
```diff
@@ -119,28 +119,33 @@ const deleteMessage = async ({ auth, message, dispatch }) => {
...
+    await api.markTopicAsRead(auth, streamId, topic);
   }
 };
 markTopicAsRead.title = 'Mark topic as read';
 markTopicAsRead.errorMessage = 'Failed to mark topic as read';
 
-const unmuteTopic = async ({ auth, stream, topic }) => {
-  await api.unmuteTopic(auth, stream, topic);
+const unmuteTopic = async ({ auth, streamId, topic, subscriptions }) => {
+  const sub = subscriptions.find(x => x.stream_id === streamId);
```
WesleyAC, line 131: Instead of adding this code, just change `api.unmuteTopic` and `api.muteTopic` to take a stream ID. I'd do that in the commit before this one.

src/message/messageActionSheet.js
```diff
@@ -167,34 +172,34 @@ const deleteTopic = async ({ auth, stream, topic, dispatch, _ }) => {
       );
     });
   if (await AsyncAlert()) {
-    await dispatch(deleteMessagesForTopic(stream, topic));
+    const sub = subscriptions.find(x => x.stream_id === streamId);
```
WesleyAC, line 175: It should be fairly simple to rewrite `deleteMessagesForTopic` to take a stream ID — do that instead of finding the stream name here. Again, I'd do that in a commit before this one.

src/message/messageActionSheet.js
```diff
@@ -427,12 +432,17 @@ export const showTopicActionSheet = ({
     unreadStreams: UnreadStreamsState,
     ...
   }>,
-  stream: string,
+  streamId?: number,
```
WesleyAC, line 435: This should also not be optional. I suspect it may be simpler to keep this taking a `stream: string` for now, so that we can centralize the error handling (since currently, both of the callers get the stream id by searching the subscriptions), but I don't have a strong feeling on that.

WesleyAC (changes requested):
Thanks! Looks good overall, just a few changes on the stream name/stream ID refactor :)

WesleyAC:
Hmm, I just tested this on my device, and pressing "Mark topic as read" from the home screen doesn't seem to do anything (the topic is still unread). Any idea what's up with that?

src/message/messageActionSheet.js
AkashDhiman: On doing this, flow gives me 2 error.
> Avoid using \`&&\` to check the value of  number [1]. Consider handling falsy values (0 and NaN) by using a conditional to choose an explicit default instead.

> Sketchy null check on  number [1] which is potentially 0. Perhaps you meant to check for  null or undefined [2]?

Your condition checks if `unreadCount` gives a truthy value which is a bad idea for number data type since 0 results in a falsy value (hence we have this rule set in our `.flowconfig`).
I understand that it won't affect this condition but this is a special case, hence, to avoid the error essentially, I am checking for `undefined` explicitly.

src/message/__tests__/messageActionSheet-test.js
AkashDhiman: Sure, squashed into previous commit in the new revision.

src/message/__tests__/messageActionSheet-test.js
AkashDhiman: I kept it here to make sure that having any random topic in `unreadStream` doesn't affect the `markTopicAsRead` button to the actionsheet for another topic (this isn't true in the current scenario but may be true for a possible future bad implementation.)
I haven't removed it in the new revision, give me another cue if you think that this is unnecessary and I will remove it.

src/message/messageActionSheet.js
AkashDhiman: I have added a try catch block before `showTopicActionSheet` in the new revision and made `stream_id` required.
In case try fails I am also displaying a toast with a generic "Something went wrong". I can't figure out a message concise enough to display here so that the user gets an idea what went wrong so I went with that for now, if you have any suggestions let me know.

src/message/messageActionSheet.js
AkashDhiman: done.

src/message/messageActionSheet.js
AkashDhiman: yes, did the necessary changes.

src/message/messageActionSheet.js
AkashDhiman: using `streamId` does cause me to repeat the error handling code at the calling site. But If we use `stream` we would have a mixture of two types of ActionSheet (one wanting stream and one wanting streamId) since there are many places where I can easily use `StreamId` (without searching). If we don't do this mixup we would be promoting the use of `stream` where we could easily have used `streamId` which may make using `showTopicActionSheet` difficult in some new places or it may make the refactor we would eventually do, difficult.

This is not a very convincing argument by me but I am still using `streamId`.

AkashDhiman:
> Hmm, I just tested this on my device, and pressing "Mark topic as read" from the home screen doesn't seem to do anything (the topic is still unread). Any idea what's up with that?

I am speculating that this was related to [syncing read/unread state](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/syncing.20read.2Funread.20state/near/1166265).
Rebasing with master fixed this. This was only reproducible if previously I had not read any message in the app (in case I did that, it behaved normally until i cleared its data), hence I missed this before.

---
@WesleyAC This PR is ready for a re-review,

I had a bit of doubt regarding [nfc] commits It would be great if you could clear that --
Are my commits in this PR appropriately labelled as [nfc]? For me functional requirement is anything client facing, and I have marked all the commits that don't really change for the client as [nfc], even when they do from a developer point of view.

Are we ever going to do the Outbox refactor to work along with stream_id, if so would it be a good idea to also add TODOs at places that were changed in this PR but would require another change once the Outbox refactor is done, for example at the call site of `showActionSheet`.

src/__tests__/lib/exampleData.js
```diff
@@ -225,12 +225,13 @@ export const crossRealmBot: CrossRealmBot = makeCrossRealmBot({ name: 'bot' });
 
 const randStreamId: () => number = makeUniqueRandInt('stream IDs', 1000);
 export const makeStream = (
-  args: {| name?: string, description?: string |} = Object.freeze({}),
+  args: {| name?: string, description?: string, streamId?: number |} = Object.freeze({}),
 ): Stream => {
   const name = args.name ?? randString();
   const description = args.description ?? `On the ${randString()} of ${name}`;
+  const streamId = args.streamId ?? randStreamId();
```
WesleyAC, line 232: I'd do this inline, rather than making a variable.

src/message/messageActionSheet.js
WesleyAC: Ah, I see. I'd personally go with `if (!!unreadCount && unreadCount > 0) {`, but it doesn't really matter, since I don't think that `unreadCount` can be `null`. Seems fine as-is.

src/message/messageActionSheet.js
```diff
@@ -436,12 +413,22 @@ export const showTopicActionSheet = ({
...
 |}): void => {
+  const sub = backgroundData.subscriptions.find(x => x.stream_id === streamId);
+  try {
+    if (!sub) {
+      throw Error('Stream id does not exist.');
+    }
+  } catch (err) {
+    logging.error(err);
+    return;
+  }
```
WesleyAC, line 427: I don't think there's a reason to use `try`/`catch` here, we can just do:

```
if (!sub) {
  logging.error('Stream id does not exist.');
  return;
}
```

instead.

src/streams/TopicItem.js
```diff
@@ -52,15 +54,27 @@ export default function TopicItem(props: Props) {
...
       onLongPress={() => {
+        try {
+          if (!subscription) {
+            throw Error('Could not find subscription with given stream name.');
+          }
+        } catch (err) {
+          logging.error(err);
+          showToast(_('Something went wrong.'));
+          return;
+        }
```
WesleyAC, line 71: Same comment as above re: `try`/`catch`

src/streams/TopicItem.js
```diff
@@ -52,15 +54,27 @@ export default function TopicItem(props: Props) {
...
     <Touchable
       onPress={() => onPress(stream, name)}
       onLongPress={() => {
+        try {
+          if (!subscription) {
+            throw Error('Could not find subscription with given stream name.');
+          }
+        } catch (err) {
+          logging.error(err);
+          showToast(_('Something went wrong.'));
```
WesleyAC, line 69: Personally, I'd opt to just not show an error here — I don't think there's anything useful the user can do about this condition (which I think we expect to never happen), and just not showing the actionsheet seems better than popping up a cryptic error.

src/webview/handleOutboundEvents.js
```diff
@@ -214,11 +214,24 @@ const handleLongPress = (
...
+
+      try {
+        if (!subscription) {
+          throw Error('Could not find subscription with given stream name.');
+        }
+      } catch (err) {
+        logging.error(err);
+        showToast(_('Something went wrong.'));
+        return;
+      }
```
WesleyAC, line 228: Same comment here as above.

WesleyAC (changes requested):
Looks pretty good! Left a few comments about your recent changes.

WesleyAC:
Oh, and your `[nfc]` commit labeling looks good :)

WesleyAC (approved):
LGTM, thanks!

src/message/messageActionSheet.js
```diff
@@ -173,7 +173,10 @@ const deleteTopic = async ({ auth, stream, topic, dispatch, _ }) => {
       );
     });
   if (await AsyncAlert()) {
-    await dispatch(deleteMessagesForTopic(stream, topic));
+    const sub = subscriptions.find(x => x.name === stream);
+    if (sub) {
+      await dispatch(deleteMessagesForTopic(sub.stream_id, topic));
```
chrisbobbe: Is the added `if (sub)` conditional a functional change? If so, let's be sure it's a change that we want, describe it in the commit message, and un-mark the commit as NFC.

If it is NFC, I'm not seeing that that's obvious; that should get an explanation if it turns out to be the case, I think. 🙂

src/__tests__/lib/exampleData.js
```diff
@@ -225,12 +225,12 @@ export const crossRealmBot: CrossRealmBot = makeCrossRealmBot({ name: 'bot' });
 
 const randStreamId: () => number = makeUniqueRandInt('stream IDs', 1000);
 export const makeStream = (
-  args: {| name?: string, description?: string |} = Object.freeze({}),
+  args: {| name?: string, description?: string, streamId?: number |} = Object.freeze({}),
```
chrisbobbe, line 228: nit: Several of these functions that make data objects (`streamMessage`, `pmMessage`, etc.) accept the property names directly, without bothering to use camelCase. Best to do so here, for consistency, I think: so, `stream_id` instead of `streamId`.

src/message/messageActionSheet.js
```diff
@@ -436,12 +413,18 @@ export const showTopicActionSheet = ({
     unreadStreams: UnreadStreamsState,
     ...
   }>,
-  stream: string,
+  streamId: number,
   topic: string,
 |}): void => {
+  const sub = backgroundData.subscriptions.find(x => x.stream_id === streamId);
+  if (!sub) {
+    logging.error('Stream id does not exist.');
```
chrisbobbe, line 421: Is this quite the right error message? We've got a stream ID; it's the subscription that we haven't found, right?

Also, do we expect the `if (!sub)` condition to ever happen, and why or why not?

If we never expect it to happen, let's use [`invariant`](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#invariant-assertions) instead (here and in similar places in this commit); the number of places we've used that is small but growing. 🙂

src/streams/TopicItem.js
```diff
@@ -52,15 +53,22 @@ export default function TopicItem(props: Props) {
...
   }));
 
+  const subscription = backgroundData.subscriptions.find(x => x.name === stream);
+
   return (
     <Touchable
       onPress={() => onPress(stream, name)}
       onLongPress={() => {
+        if (!subscription) {
+          logging.error('Stream id does not exist.');
```
chrisbobbe, line 63: (same comment as above, about the error message, whether we expect this to happen, and `invariant`)

src/webview/handleOutboundEvents.js
```diff
@@ -214,11 +214,19 @@ const handleLongPress = (
   const { dispatch, showActionSheetWithOptions, backgroundData, narrow, startEditMessage } = props;
   if (target === 'header') {
     if (message.type === 'stream') {
+      const stream = streamNameOfStreamMessage(message);
+      const subscription = backgroundData.subscriptions.find(x => x.name === stream);
+
+      if (!subscription) {
```
chrisbobbe, line 220: (same comment as above, about whether we expect this to happen, and `invariant`)

chrisbobbe (changes requested):
Thanks, @AkashDhiman! See a few comments below, most of them small.

There's one thing I'm hesitating on before merging: https://github.com/zulip/zulip-mobile/issues/3244#issuecomment-840200325. It'd be great to have @gnprice's and/or @WesleyAC's thoughts on whether we can drop support here for servers 1.9.x and below, or if we should write a conditional to keep compatibility, or if we should do more thinking and work about establishing what servers are officially supported, and how to communicate that, first. Discussion [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/compatibility.20documentation/near/1179737).

Also, I notice that there are some commits where the commit message body goes past 70 columns in a line; let's wrap those to be in line with the [style guide](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#message-body):

> Any paragraph content in the commit message should be line-wrapped to about 68 characters per line, but no more than 70, so that your commit message will be reasonably readable in `git log` in a normal terminal. You may find it helpful to […]

The summary line can go a bit longer just because it has to, sometimes; that guide says no more than 76 characters for the summary line.

src/message/messageActionSheet.js
chrisbobbe: Ah, looking back after posting the review: if we really expect `sub` to exist every single time, this is probably another good place to use `invariant` (see my other comments where I mention that).

src/message/messageActionSheet.js
```diff
@@ -118,19 +119,25 @@ const deleteMessage = async ({ auth, message, dispatch }) => {
 deleteMessage.title = 'Delete message';
 deleteMessage.errorMessage = 'Failed to delete message';
 
-const unmuteTopic = async ({ auth, stream, topic }) => {
-  await api.setTopicMute(auth, stream, topic, false);
+const markTopicAsRead = async ({ auth, streamId, topic }) => {
+  await api.markTopicAsRead(auth, streamId, topic);
+};
+markTopicAsRead.title = 'Mark topic as read';
```
chrisbobbe, line 125: Ah, I meant to post in my review but forgot:

The title and error message are translated via `_` (the function with type `GetText`), so we should make sure these user-facing strings have entries in static/translations/messages_en.json; looks like 'Mark topic as read' is already there but the error message is not.

src/message/messageActionSheet.js
AkashDhiman: Using invariant in the new revision,

I think `subscriptions` will not change at this position i.e. it will keep using the same value that `showTopicActionSheet` was called with. even if it is updated by react in between the execution of this function.
I am not sure if that is intentional though (since such a change can happen when the user is about to interact with the action sheet.)

So if `showTopicActionSheet` is called with right stream name and Subscription object here,  `(sub !== undefined)` will always be true.
Hence I am using invariant.

**Why i think `subscriptions` will not change at this position?**
I think this is an example of "stale closure" (more info at: https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function)

Essentially what I could infer from it was that when we pass a state to a function and the state updates, our function creates a closure with the value of state that was passed to it when the function was called and it keeps using that outdated data even when the state updates.
Just in case this is unnecessary a way to solve this is by using `useRef` (for functional components) and its equivalent for class. Here is a blog post discussing this: https://css-tricks.com/dealing-with-stale-props-and-states-in-reacts-functional-components/

src/message/messageActionSheet.js
AkashDhiman: Using invariant in the new revision. The reason is explained at https://github.com/zulip/zulip-mobile/pull/4635#discussion_r632862010

Also I changed the error message in `invariant`.

src/streams/TopicItem.js
AkashDhiman: On exploring this I realized that the current implementation allows a user to Long Press `TopicItem` on an unsubscribed stream (from `TopicListScreen`) (I think ideally this shouldn't be the case).
In that scenario subscription will be undefined.
Given that case I am not sure if we should be `logging.error` as well, so I removed that statement in the new revision and now the code simply returns on not finding subscription.

src/webview/handleOutboundEvents.js
AkashDhiman: using invariant here, the reason is same as https://github.com/zulip/zulip-mobile/pull/4635#discussion_r632862010

AkashDhiman:
@chrisbobbe Thanks for the review, I think I have addressed your comments and you may review the changes again. Please pay particular attention to https://github.com/zulip/zulip-mobile/pull/4635#discussion_r632862010, as I highlight what seems to me as something that can be considered an issue.

src/message/messageActionSheet.js
```diff
@@ -226,31 +225,38 @@ const cancel = params => {};
...
+    unreadStreams: UnreadStreamsState,
     ...
   }>,
-  stream: string,
+  streamId: number,
   topic: string,
-|}): Button<HeaderArgs>[] => {
+|}): Button<TopicArgs>[] => {
+  const sub = subscriptions.find(x => x.stream_id === streamId);
+  invariant(sub, 'Subscription with provided stream id does not exist.');
```
AkashDhiman, line 244: Just highlighting, that I have used an invariant here as well (this wasn't asked in @chrisbobbe's review). necessary for the subsequent call to `isTopicMuted`.
for the same reason as https://github.com/zulip/zulip-mobile/pull/4635#discussion_r632862010

gnprice: Thanks, writing this expectation explicitly like this is helpful. I think the expectation is not correct. :grin: I'll elaborate in a review comment.

gnprice (changes requested):
Thanks @AkashDhiman for these changes, and @WesleyAC and @chrisbobbe for the reviews!

I haven't read all of this in detail, but one thing I see is that this code appears to assume that for any topic we validly try to make a topic action sheet for, there'll be an entry in `subscriptions` for the stream. I think that isn't right, because we'll only have one of those if the user is subscribed to the stream. For information about the stream in general, not specific to the user's individual relationship to the stream, we want a `Stream` value, gotten from `getStreamsById` or its relatives.

Here's a general sequence of changes that I think would work well. Each of these might be one or several commits.
* (Before anything else, the rename to "topic" from "header". That is overdue :wink: .)
* Then, first, have the action-sheet code take a stream ID as well as a stream name, redundantly.
  * Can call them `streamName` and `streamId`.
  * The caller can look the ID up, if it doesn't have it, in `getStreams(state)` (just like the code in the current revision of this PR looks in the result of `getSubscriptions(state)`.) Then this code doesn't need to do that, and the presence of stream IDs can be pushed further upstream in later followups.
* Then, make the `setTopicMute` and `deleteMessagesForTopic` refactors -- but this time without the `subscriptions.find`, and instead using the stream ID we were passed.
* Now, make the changes supplying the unreads data and adding the option to mark as read.
* Meanwhile, if nothing is using the stream name anymore, stop passing it in. Or if something is, perhaps refactor that away too as a followup.
  * Can switch from `subscriptions: getSubscriptions(state)` in the background data to `subscriptionsById: getSubscriptionsById(state)`.
  * Can also add `streamsById: getStreamsById(state)`, if that helps.

src/__tests__/lib/exampleData.js
```diff
@@ -263,6 +264,28 @@ export const makeSubscription = (args: {| stream?: Stream |} = Object.freeze({})
...
+    mockData.map(unreadStream => [
+      unreadStream.streamId,
+      Immutable.Map(
+        unreadStream.topics.map(unreadTopic => [
+          unreadTopic.topicName,
+          Immutable.List(unreadTopic.messageIds),
+        ]),
+      ),
+    ]),
+  );
```
gnprice, line 287: Instead of directly building up the internal data structure that `state.unread.streams` comes in, let's use the unreads reducer to build it up. That way:
* The details of the data structure stay internal to the unreads model. This makes it easier to change that data structure around later (as we recently did, in fact, earlier this year.)
* By keeping those details out, it also keeps this code that's building the data more abstract in a way that I think will make it more readable: it'll say "start from an initial/empty state, then insert each of the messages in this list".
* Because the reducer takes basically message objects, I think this will ultimately help simplify the tests that use this, too: instead of referring to a stream by an ID here and a name there, and counting on tying those together elsewhere in the test, or using the same ID or name in several places and implicitly using that they're the same value, they can use a real `Stream` object throughout.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -56,60 +56,98 @@ describe('constructActionButtons', () => {
   });
 });
 
-describe('constructHeaderActionButtons', () => {
+describe('constructTopicActionButtons', () => {
+  test('show mark as read if topic is unread', () => {
+    const stream = eg.makeStream({ name: 'test stream', stream_id: 123 });
+    const subscriptions = [{ ...eg.subscription, ...stream }];
+    const unreadStreams = eg.makeUnreadStreamsState([
+      { streamId: 123, topics: [{ topicName: 'test topic', messageIds: [1, 2] }] },
```
gnprice, line 64: Instead of saying `streamId: 123` here and `stream_id: 123` above where `stream` is constructed, this can say `streamId: stream.stream_id` here. That way the connection between this and the ID of `stream` is explicit.

I think that also makes it unnecessary to add a `stream_id` option to `eg.makeStream`. In general I think in almost any place we might use that, it'd be better to turn the flow around just like here: let the stream get whatever ID it gets, and where we want to have that stream's ID just refer to it directly.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -56,60 +56,98 @@ describe('constructActionButtons', () => {
   });
 });
 
-describe('constructHeaderActionButtons', () => {
+describe('constructTopicActionButtons', () => {
+  test('show mark as read if topic is unread', () => {
+    const stream = eg.makeStream({ name: 'test stream', stream_id: 123 });
```
gnprice, line 61: Similarly, instead of explicitly setting `name: 'test stream'` and then writing `test stream` again later in the test (in one of the intermediate commits in this branch), we can just say `stream.name` at that later point.

That makes the test code simpler and shorter -- and in particular it makes it clear that there's nothing interesting about the particular name and it isn't something the reader needs to pay attention to.

That at that point we're passing no arguments to `eg.makeStream` at all, so we might simplify further by using `eg.stream` instead of a local, to further make clear that there's nothing interesting about this particular stream.

src/__tests__/lib/exampleData.js
gnprice: For an example, see:
https://github.com/zulip/zulip-mobile/blob/31c4bc808053532499399e0dc241810e4a0113b9/src/unread/__tests__/unreadModel-test.js#L149-L160

For another example:
https://github.com/zulip/zulip-mobile/blob/31c4bc808053532499399e0dc241810e4a0113b9/src/unread/__tests__/unread-testlib.js#L36-L58

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -56,60 +56,98 @@ describe('constructActionButtons', () => {
...
+    const subscriptions = [{ ...eg.subscription, ...stream }];
+    const unreadStreams = eg.makeUnreadStreamsState([
+      { streamId: 123, topics: [{ topicName: 'test topic', messageIds: [1, 2] }] },
+    ]);
+    const buttons = constructTopicActionButtons({
+      backgroundData: { ...baseBackgroundData, subscriptions, unreadStreams },
+      streamId: stream.stream_id,
+      topic: 'test topic',
```
gnprice, line 69: In a related spirit, instead of saying the string `'test topic'` twice and implicitly using that they're the same, it'd be better to have a `const topic = 'test topic'` and refer to that in both places.

src/__tests__/lib/exampleData.js
gnprice: > Instead of directly building up the internal data structure that `state.unread.streams` comes in, let's use the unreads reducer to build it up.

(This isn't a principle we have written down anywhere -- in fact I really only started doing it in our code just recently, with that refactor to this very unreads model. I think it becomes more important as our data structures get interesting, like that one is. @chrisbobbe noticed it at the time and we had a conversation about it, which might be informative to read and I should perhaps turn into a style-guide entry or something; maybe it was on one of the PRs where I was making those changes?)

gnprice (commented):
Here's also some comments from looking specifically at the test code. Thanks again!

AkashDhiman:
@gnprice thank you for the review, I have made the changes you requested.

---
> but one thing I see is that this code appears to assume that for any topic we validly try to make a topic action sheet for, there'll be an entry in subscriptions for the stream

Yes I did make this assumption in the previous revision (changed now). It felt a bit unnatural to me that we provide options like mute topic to a topic in a stream we aren't really subscribed to and it could also cause conflicts with future options that would be even more irrelevant for topics in muted streams.

src/streams/TopicItem.js
```diff
@@ -45,20 +54,25 @@ export default function TopicItem(props: Props) {
...
     auth: getAuth(state),
     mute: getMute(state),
+    streams: getStreamsById(state),
     subscriptions: getSubscriptions(state),
+    unreadStreams: getUnreadStreams(state),
     ownUser: getOwnUser(state),
     flags: getFlags(state),
   }));
 
+  const currentStream = [...backgroundData.streams.entries()].find(x => x[1].name === streamName);
```
AkashDhiman, line 64: I wanted to de-structure this similar to
```javascript
const [currentStreamId, currentStream] = [...backgroundData.streams.entries()].find(x => x[1].name === streamName);    
```
Because it seemed more clean, but flow doesn't allow this because it thinks that this can potentially be `undefined`. Is their a way to handle that?

gnprice, line 64: I'd like to avoid introducing this sort of linear scan through all the streams.

We have a number of those now, but it's a pattern that's potentially quite bad for performance and I'd like to get rid of over time; that's #3339.

Fortunately it's not too hard to avoid: we just need an appropriate data structure like a `Map`.

If you look in `subscriptionSelectors.js` (which has selectors for both streams and subscriptions -- that's itself something we should perhaps do something to clarify), you'll see that we currently have memoizing selectors for three of the four cells in a 2x2 matrix:
```js
export const getStreamsById: Selector<Map<number, Stream>> = createSelector(
  getStreams,
  streams => new Map(streams.map(stream => [stream.stream_id, stream])),
);

export const getSubscriptionsById: Selector<Map<number, Subscription>> = createSelector(
  getSubscriptions,
  subscriptions =>
    new Map(subscriptions.map(subscription => [subscription.stream_id, subscription])),
);

export const getSubscriptionsByName: Selector<Map<string, Subscription>> = createSelector(
  getSubscriptions,
  subscriptions => new Map(subscriptions.map(subscription => [subscription.name, subscription])),
);
```

The one we need here is the remaining case: a `getStreamsByName`. So the first thing to do is to add a selector like that, following the examples of those three.

Then we can add a `streamsByName: getStreamsByName(state)` to the `backgroundData` found here and wherever else we need to turn a stream name into a `Stream` and/or a stream ID; and then we can do that by saying `streamsByName.get(streamName)`.

src/api/subscriptions/setTopicMute.js
```diff
@@ -5,12 +5,12 @@ import { apiPatch } from '../apiFetch';
...
 export default async (
   auth: Auth,
-  stream: string,
+  streamId: number,
   topic: string,
   value: boolean,
 ): Promise<ApiResponse> =>
   apiPatch(auth, 'users/me/subscriptions/muted_topics', {
-    stream,
+    stream_id: streamId,
```
gnprice, line 13: This will change what interface we're assuming from the server when we make a request there. When we do that, we want to be sure to say in the commit message why we believe the new interface will work.

Following the link to #3244, it looks like this new way is accepted starting with server 2.0. We're planning to desupport servers older than that soon, but we're not quite ready to do so -- #4750 added a banner to warn people on such old servers that it's time to upgrade, and we'll want to have the release with that banner out for a bit before having a further release that actually stops supporting the older servers.

So let's hold this commit off for now, leaving it out of this PR.

gnprice (commented):
Thanks @AkashDhiman for the revision! About to go to bed, but I've read the first 7 commits:
28d86d070 ActionSheet [nfc]: Rename HeaderActionSheet to TopicActionSheet.
5f1cc4ab7 ActionSheet [nfc]: Rename variable `stream` to `streamName`.
4e8848448 ActionSheet [nfc]: Pass `StreamsById` to actionsheet.
bf6add072 ActionSheet [nfc]: Parameterize some values used in `tests`.
f44af8b2e ActionSheet [nfc]: Update signature of actionsheet to use streamId.
288edc1b0 ActionSheet [nfc]: Use streamId in actionsheet functions wherever possible.
2fb84bb10 api [nfc]: Update `setTopicMute` to use `stream_id`.

and they generally look great. Comments below.

That leaves 4 further commits:
0206d8988 topicActions [nfc]: Use `streamId` in `deleteMessagesForTopic`.
5b0bf1bc2 ActionSheet [nfc]: Pass unreadStreams data to action sheet.
9e80f627c ActionSheet: Add mark as read option to topic action sheet.
7c396415d ActionSheet [nfc]: Update signature of actionsheet to not use `streamName`.

which I'll look at later.

src/topics/topicActions.js
```diff
@@ -40,25 +40,22 @@ export const fetchTopicsForStream = (narrow: Narrow) => async (
...
 };
 
-export const deleteMessagesForTopic = (streamName: string, topic: string) => async (
+export const deleteMessagesForTopic = (streamId: number, topic: string) => async (
   dispatch: Dispatch,
   getState: GetState,
 ) => {
   const state = getState();
   const outbox = getOutbox(state);
+  const currentStream = getStreamsById(state).get(streamId);
```
gnprice, line 49: Better to just call this `stream`. The extra word "current" isn't really adding any meaning here.

(Sometimes "current" is helpful when there's also a "previous" or "next", or some other notion of another thing of the same kind.)

... Oh I see, this is the name the existing code had.

But we're already touching all the places the old name was used. So the diff won't really get any bigger or more complex if we also fix up this name while we're here :slightly_smiling_face:

src/topics/topicActions.js
```diff
@@ -40,25 +40,22 @@ export const fetchTopicsForStream = (narrow: Narrow) => async (
...
+
   outbox.forEach((outboxMessage: Outbox) => {
     if (
       outboxMessage.type === 'stream'
-      && streamNameOfStreamMessage(outboxMessage) === streamName
+      && streamNameOfStreamMessage(outboxMessage) === currentStream?.name
```
gnprice, line 54: Let's add a TODO comment here that mentions #3998 and #3918. That will help us find this spot and clean it up to use the stream ID too, once we've added stream IDs to our outbox messages.

static/translations/messages_en.json
```diff
@@ -242,5 +242,6 @@
   "Signed out": "Signed out",
   "Remove account?": "Remove account?",
   "This will make the mobile app on this device forget {email} on {realmUrl}.": "This will make the mobile app on this device forget {email} on {realmUrl}.",
-  "Remove account": "Remove account"
+  "Remove account": "Remove account",
+  "Failed to mark topic as read": "Failed to mark topic as read"
```
gnprice, line 246: nit: add a new string here on any line other than the last -- that way the commas don't cause the diff to unnecessarily touch another string :stuck_out_tongue:

src/message/messageActionSheet.js
```diff
@@ -226,30 +226,38 @@ const cancel = params => {};
...
   topic: string,
-|}): Button<HeaderArgs>[] => {
+|}): Button<TopicArgs>[] => {
   const buttons = [];
   if (ownUser.is_admin) {
     buttons.push(deleteTopic);
   }
-  if (isTopicMuted(stream, topic, mute)) {
+  const unreadCount = unreadStreams.get(streamId)?.get(topic)?.size;
+  if (unreadCount !== undefined && unreadCount > 0) {
```
gnprice, line 250: Hmm interesting.

This is reasonable, but I'd kind of like to keep this kind of detail of the unreads data structure encapsulated inside the unreads model. I'm thinking perhaps a function in `unreadModel.js` that takes an `UnreadState` and a stream and topic, and returns a count.

It looks like we already have one other spot outside the unreads code that similarly looks inside this data structure -- in `topicSelectors.js`:
```js
      const unreadCount = unreadStreams.get(stream.stream_id)?.get(name)?.size ?? 0;
```
(I found that in a kind of crude, but effective, way: by editing the definition of `UnreadStreamsState` to say `export opaque type UnreadStreamsState = …`, and looking through the resulting Flow errors to find any outside the unreads code.)

I'll write up a quick change to factor that `topicSelectors.js` line out into the function I have in mind, and push that to this PR branch. Then you can rebase it (after fetching the new version of your branch) to come before this commit, and call that function instead.

src/message/__tests__/messageActionSheet-test.js
```diff
@@ -56,61 +60,93 @@ describe('constructActionButtons', () => {
   });
 });
 
-describe('constructHeaderActionButtons', () => {
+describe('constructTopicActionButtons', () => {
+  const stream = eg.makeStream();
+  const streamMessage = eg.streamMessage({ stream });
+  const topic = streamMessage.subject;
+  const streamId = streamMessage.stream_id;
+  const streams = deepFreeze(new Map([[streamMessage.stream_id, stream]]));
```
gnprice, line 68: nit: cleaner to say `[stream.stream_id, stream]` -- that makes it more obvious that it's correct, and also more stably correct in case of changes to `streamMessage` or something

gnprice (commented):
OK, and read the remaining commits! All look good, modulo small comments below.

src/message/messageActionSheet.js
gnprice: > I'll write up a quick change to factor that `topicSelectors.js` line out into the function I have in mind, and push that to this PR branch.

OK, that change grew a bit as I wrote it, so I sent it as its own PR: #4805 .

If this PR is otherwise ready before that one is merged, no need to wait for it; we can refactor this to use that as a followup. But that PR is pretty short and simple, so it's likely we'll have it merged soon anyway.

src/streams/TopicItem.js
```diff
@@ -45,20 +55,26 @@ export default function TopicItem(props: Props) {
...
-      onPress={() => onPress(stream, name)}
+      onPress={() => onPress(streamName, name)}
       onLongPress={() => {
-        showHeaderActionSheet({
+        showTopicActionSheet({
           showActionSheetWithOptions,
           callbacks: { dispatch, _ },
           backgroundData,
-          stream,
+          streamId: stream[0],
```
gnprice, line 77: I think you want `stream.stream_id` here :wink:

(And in the other spot like it.)

gnprice (commented):
Thanks @AkashDhiman for the revision! All looks good in the code except one item mentioned below. Please go ahead and fix that, and then we can merge :smile: 

---

Meanwhile, here's a nit in a commit message and a tip:

    ActionSheet [nfc]: Update signature of actionsheet to not use
    `streamName`.

That's a long summary line. The first thing you can do to shorten it is to drop the second mention of "action sheet" -- the prefix already tells the reader that that's the area we're talking about. So:

> ActionSheet [nfc]: Update signature to not use `streamName`.

Then I'd reword with a verb that lets us get faster to the point. Here, "drop" carries the meaning of "update … to not use". So we can say:

> ActionSheet [nfc]: Drop `streamName` from signature.

A bonus of that wording is that it gets the "`streamName`" bit earlier in the summary line, and that's the part that's most specific to this change.

AkashDhiman:
@gnprice done.

gnprice:
Thanks for the quick revision! Merged.

PR #4639, opened 2021-04-10
tools/run-android: Pass through args to gradle

WesleyAC:


chrisbobbe:
LGTM, thanks, @WesleyAC!

Issue #4638, opened 2021-04-10
Not able to react with some emojis.


nikhilmaske-2001:
Issue: I am not able to react with some emojis like 😄 1️⃣ 2️⃣  and some more in the chat on the messages.
I am using the latest version of mobile app (i.e 27.161) that is available on play store.

WesleyAC:
For the numbers, this was fixed in #4551, which I guess hasn't hit the play store yet. @gnprice, any idea when we're going to do a release?

We were actually just talking about the smile one last night, I'll make a PR fixing that soon as well.

gnprice:
BTW @nikhilmaske-2001 thanks for the report!

If you see any other emoji with these symptoms (any other than `:smile:`, `:slight_smile:`, and the numerals), please file an issue about it. We should make this systematically not a problem -- that's #2956 -- but until then, fixing any given emoji ad hoc is quick and we want to do so for things people run into.

> @gnprice, any idea when we're going to do a release?

Yeah, we should do this soon. #mobile-team is probably the best place for discussion.

PR #4640, opened 2021-04-10
emoji: Fix smile/slight_smile emoji names

WesleyAC:
This updates the codePointMap to reflect changes in modern server
versions. We should do this automatically, but this is a good enough fix
for now.

Fixes: #4638


chrisbobbe:
Merged, thanks!

> We should do this automatically

Yeah. That's #2956.

If you happen to know whether this particular fix brings us completely up-to-date with modern servers, that'd be good to note here; otherwise it's probably fine as #2956 will have us naturally up-to-date with a server we're connected too.

PR #4643, opened 2021-04-12
messageActionSheet: Add reply with mention option

AlaaElden98:
Fixes part of: #3436


src/webview/MessageList.js
```diff
@@ -102,6 +103,7 @@ export type Props = $ReadOnly<{|
   narrow: Narrow,
   showMessagePlaceholders: boolean,
   startEditMessage: (editMessage: EditMessage) => void,
+  startReplyWithMention: (replyWithMention: ReplyWithMention) => void,
```
WesleyAC, line 106: Why does this prop get passed into `MessageList`, when it doesn't seem to use it?

src/types.js
```diff
@@ -133,7 +133,9 @@ export type EditMessage = {|
   content: string,
   topic: string,
 |};
-
+export type ReplyWithMention = {|
+  content: string,
```
WesleyAC, line 137: I don't think we need a type for this.

src/compose/ComposeBox.js
```diff
@@ -372,6 +374,14 @@ class ComposeBox extends PureComponent<Props, State> {
         this.messageInputRef.current.focus();
       }
     }
+    if (nextProps.replyWithMention !== this.props.replyWithMention) {
+      const message = nextProps.replyWithMention ? nextProps.replyWithMention.content : '';
+      this.setMessageInputValue(message);
+      if (this.messageInputRef.current !== null) {
```
WesleyAC, line 380: We don't need the `!== null` here, we can just do `if (this.messageInputRef.current)`, or even better, no if statement and `this.messageInputRef.current?.focus();` should work.

src/message/messageActionSheet.js
```diff
@@ -82,6 +86,21 @@ const reply = ({ message, dispatch, ownUser }) => {
 reply.title = 'Reply';
 reply.errorMessage = 'Failed to reply';
 
+const replyWithMention = ({ message, dispatch, ownUser, narrow, startReplyWithMention }) => {
+  const replyNarrow = getNarrowForReply(message, ownUser.user_id);
+  const replyMessage = `@**${message.sender_full_name}**`;
```
WesleyAC, line 91: One thing that this doesn't handle is the case where two users have the same display name — you might want to look at how the webapp code handles that.

src/message/messageActionSheet.js
```diff
@@ -82,6 +86,21 @@ const reply = ({ message, dispatch, ownUser }) => {
...
+  if (JSON.stringify(replyNarrow) !== JSON.stringify(narrow)) {
```
WesleyAC, line 92: It's better to use `isEqual` from `lodash.isequal`, rather than `JSON.stringify`.

src/message/messageActionSheet.js
```diff
@@ -82,6 +86,21 @@ const reply = ({ message, dispatch, ownUser }) => {
...
+    dispatch(draftUpdate(replyNarrow, replyMessage));
```
WesleyAC, line 93: Is there a reason `draftUpdate` doesn't work for both of these cases?

src/webview/MessageList.js
AlaaElden98: That seems odd to me! I don't recall why!

src/message/messageActionSheet.js
AlaaElden98: `draftUpdate` actions work when `ComposeBox` first gets rendered, so if we are at the same narrow, the `draftUpdate` action will fire, but the `ComposeBox` will not be updated it gets re-rendered.

src/message/messageActionSheet.js
AlaaElden98: Aaah, I see, I need to use the `user_id`

src/message/messageActionSheet.js
AlaaElden98: Actually, Greg answered that question for me, [here is the answer](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Reply.20with.20mention/near/1160754)

src/webview/MessageList.js
AlaaElden98: That's make me wonder, why `startEditMessage` passed to MessageList?

chrisbobbe:
Let's come back to this when a certain refactor on `ComposeBox` has been done. As Greg [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Reply.20with.20mention/near/1161069),

> […]
> 
> Which actually connects to this other thread today:
> https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Next.20steps.20for.20.60ComposeBox.60/near/1160876
> 
> As I think about how to write a callback to do what we need here in the case where you're already looking at the relevant narrow, I think if we try to do it the same way as we have `editMessage`, then the interaction between that and `editMessage` itself would quickly get pretty messy.
> 
> But once we do the refactoring that @**Chris Bobbe** was describing in that thread, then I think it will be much more straightforward.

AlaaElden98:
Update on this : 
I manage to add the "Reply with mention" option to the action sheet when the user isn't in the same narrow - same behaviour "Reply" have - however it's important to be able to have "Reply with mention" if the user is in the same narrow of the message, i tried to implement this but I couldn't do it, I need some help there. 
Otherwise, the PR is ready for another review

Issue #3398, opened 2019-03-12
Avatars should be squares with rounded corners, not circles


gnprice:
Same thing as #205, but round 2 :wink: . The same reasoning applies:

> Zulip's avatars should be displayed as squares (with rounded corners), not as circles.
>
> This is based on a lot of consideration -- circular avatars are problematic because many images that we'd like to use in avatars (e.g. bot logos) don't actually look good in circular form, whereas just about anything looks good in a rounded square.

One example: our own bot avatars look quite bad when cropped to a circle, as they now are on the user-profile screen (`AccountDetailsScreen`).

---

Relatedly: we shouldn't have had to have a round 2 of this, and we should avoid a round 3.

One root cause of round 2 is that in round 1 (in #208) we left the "circle" case lying around in the code, even though we had determined that we had no intention of actually using it. That's a bad pattern for a number of reasons -- most of all, it makes the code more complex and confusing to understand -- and this is another of them.

Issue #205, opened 2017-01-11
Avatars should be squares with rounded corners, not circles


timabbott:
Zulip's avatars should be displayed as squares (with rounded corners), not as circles.

This is based on a lot of consideration -- circular avatars are problematic because many images that we'd like to use in avatars (e.g. bot logos) don't actually look good in circular form, whereas just about anything looks good in a rounded square.

PR #4647, opened 2021-04-12
Add realm icon to AccountItems in organisation switcher, Fixes: #392

### LINKED ISSUE #392: Use realm icons for quick mobile organization switcher

vaibhavs2:
Fixes: #392
|1|2|3|
|---|---|---|
||||

The third image above, I created that mannualy to show some realm icons together :)

If some wonder why I made icon container round look at the image below(not looking good), sometime we can get round icon sometime square shaped, so its better to have round container, we can adjust square shaped Icons in round container.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -164,11 +164,14 @@ export const zulipVersion = new ZulipVersion('2.1.0-234-g7c3acf4');
 
 export const zulipFeatureLevel = 1;
 
+export const realmIcon = new URL('/icon.png', realm);
+
 export const makeAccount = (
   args: {|
     user?: User,
     email?: string,
     realm?: URL,
+    realmIcon?: URL,
     apiKey?: string,
     zulipFeatureLevel?: number | null,
     zulipVersion?: ZulipVersion | null,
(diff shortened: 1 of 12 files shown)


chrisbobbe:
> If some wonder why I made icon container round look at the image below(not looking good), sometime we can get round icon sometime square shaped, so its better to have round container, we can adjust square shaped Icons in round container.

I disagree; I think we'd be better off using the image with no shape change, or at most make it a square with (slightly) rounded corners. I think the basic reason is that some realm logos weren't built to look best as a circle, and we shouldn't assume they were; see #205 and #3398 where we settled on rounded-corner squares for avatar images for that reason. 🙂

vaibhavs2:
Made corner slightly rounded :)



_again above list mannualy created_

vaibhavs2:
@chrisbobbe thanks for review, now its ready for another review :)

PR #4650, opened 2021-04-12
getDraftForNarrow selector [nfc]: Return undefined for none-found.

chrisbobbe:
It seems useful for callers to be able to distinguish between "no
draft found for this narrow" and "empty-string draft found for this
narrow".


gnprice:
Seems sensible.

We should in this case make sure that also when we're clearing a draft, we remove it from the map rather than writing back `''`. I think the main case of that is here:
```js
  handleMessageChange = (message: string) => {
    this.setState({ message, isMenuExpanded: false });
    const { dispatch, narrow } = this.props;
    dispatch(sendTypingStart(narrow));
  };
```
in the case where `message` -- i.e. the just-updated contents of the message input -- is empty; not sure if there are others.

chrisbobbe:
I think the reducer code will remove the draft from state if it's been cleared:

```js
const draftUpdate = (state, action) => {
  const narrowStr = keyFromNarrow(action.narrow);

  if (action.content.trim().length === 0) {
    // New content is blank; delete the draft.
    if (!state[narrowStr]) {
      return state;
    }
    const newState = { ...state };
    delete newState[narrowStr];
    return newState;
  }

  return state[narrowStr] === action.content ? state : { ...state, [narrowStr]: action.content };
};
```

But possibly the reducer shouldn't be where that happens, and dispatchers should signal clearing a draft by (e.g.) sending `undefined` or even using a new kind of action like `DRAFT_CLEAR`.

If the reducer doesn't recognize draft-clearing, dispatchers will have to (this is currently the only one).

Since Flow can't distinguish a cleared-draft string (e.g., the empty string) from a non-cleared-draft string, it's not likely to help ensure we do the right thing anywhere, I think. That makes me look to the reducer, as one central place where we have to get it right, but I'm not sure that's what we'd normally do in circumstances where Flow helps more.

WesleyAC:
What's the case where we want to have a blank-string draft hanging around? It seems sensible to explicitly make this optional, in general, but I'm curious if there's a usecase here, or if it's just doing it since it feels right.

Issue #4528, opened 2021-03-13
Respect system 24-hour time setting


WesleyAC:
If the user's system is set to 24-hour time, we should display the time tags on messages and the timezone widgets in 24-hour time.

How to check this: [Android](https://stackoverflow.com/questions/4392636/how-do-i-check-if-system-is-12-or-24-hour) [iOS](https://stackoverflow.com/questions/28162729/nsdateformatter-detect-24-hour-clock-in-os-x-and-ios)

Here's a RN library to do this, I have not made any attempt to evaluate whether it's good or not: [react-native-hour-format](https://github.com/software-mansion-labs/react-native-hour-format)

abhi0504:
@WesleyAC I want to work on this issue can you kindly please assign it to me.

WesleyAC:
Thanks for taking this on, @abhi0504! Feel free to ping me on chat.zulip.org if you have any questions :)

gnprice:
I agree.

There is a Zulip user setting for this, and we currently use that -- search for `twenty_four_hour_time` and `twentyFourHourTime`. But for a mobile app I think it makes most sense to follow the setting the user has chosen on the device. I just tried several other apps that show times for things, and none of them have their own time-format setting that I can find, and each of them follows the system setting when I change it.



> Here's a RN library to do this, I have not made any attempt to evaluate whether it's good or not: [react-native-hour-format](https://github.com/software-mansion-labs/react-native-hour-format)

It looks probably fine.

It doesn't support a 100% high-quality solution -- it reads this setting at app startup, and then never notices if the user goes and changes it. Apps from Google and Apple, in my testing, do notice immediately if the setting is changed. On Android, the underlying API for this would be to use `ContentResolver#registerContentObserver`, like in this example:
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/DigitalClock.java#72
and then the library would have to provide a bit of API for the app's JS code to listen in turn. I haven't looked up what API iOS provides for doing this, but I imagine it's fundamentally similar.

That wouldn't be super hard to do... but it would be significantly more work than just using a library like this one. And it doesn't seem like a priority. Few people ever change that setting except when they're first setting up their device, I imagine.

abhi0504:
@gnprice Thank you so much for your explanation and ideal solution for the current issue 😊

PR #4653, opened 2021-04-13
setting up time format according to phone setting

abhi0504:
Using `react-native-hour-format` in `MessageList.js` to check whether the phone's default setting for time format is of 24-hour format or 12-hour format and if its 24-hour format then sending true instead of using `getRealm(state).twentyFourHourTime` prop

Also attaching some screenshots for more clarification :

Android Emulator - Pixel_3_API_29_5554 13-04-2021 16_38_59  Android Emulator - Pixel_3_API_29_5554 13-04-2021 16_37_07

Fixes: #4528


ios/Podfile
```diff
@@ -23,6 +23,8 @@ target 'ZulipMobile' do
   # Note that if you have use_frameworks! enabled, Flipper will not work and
   # you should disable these next few lines.
   use_flipper!
+  pod 'RNHourFormat', :path => '../node_modules/react-native-hour-format'
```
chrisbobbe, line 26: Was this (and some similarly purposed lines in Android files) added as a result of running `react-native link`? Unfortunately this library's setup instructions don't mention it, but I'd expect additions like this to be made unnecessary by React Native's autolinking feature (there's a doc for that somewhere).

I'd suggest reverting all changes except the ones in `src/` and running `yarn add react-native-hour-format`, and don't run a `react-native link` command. I suspect that will be sufficient setup, assuming autolinking treats this library the same way it's been treating others. (I can't remember if that ends up making a small change to an Android file after running on Android or running `tools/test android --full`, but I definitely don't expect it to change the Podfile.)

chrisbobbe (changes requested):
Thanks, @abhi0504! CI is failing; please fix that, and see a comment below.

If CI is failing, that's a signal that the PR isn't ready for a full review; I've just taken a quick look so far. If CI is failing and you don't understand what the failures mean, or you understand what they mean but you're not sure how best to fix them, please investigate a bit and then mention it if you're still having trouble.

ios/Podfile
abhi0504: Sure I have restore these linking changes and also test it out every thing is working fine

abhi0504:
@chrisbobbe thanks for your review
I have fixed the CI error which is due to using an untyped module that is `react-native-time-format` by using the libdef

Issue #4651, opened 2021-04-12
De-emphasize account deletion on "Pick account" screen


alya:
There are two problems with the account deletion interface on the "Pick account" screen:

1. In the current UI the, button for deleting the account is too prominent (bright red trash can). We expect account deletion to be quite rare, so it should be less emphasized than other actions the user can take here.

A good starting point for this would be to look at other apps to identify a set of options that users will be familiar with.

2. A confirm dialog should be added to prevent accidental account deletion.

AlaaElden98:
I can work on this

PR #4654, opened 2021-04-13
AccountPickScreen: Ask for confirmation when pressing on 'remove account'

AlaaElden98:
Fixes part of #4651

### DIFF
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -27,6 +27,7 @@ type Props = $ReadOnly<{|
 class AccountPickScreen extends PureComponent<Props> {
   static contextType = TranslationContext;
   context: GetText;
+
   handleAccountSelect = (index: number) => {
     const { accounts, dispatch } = this.props;
     const { realm, isLoggedIn } = accounts[index];
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -1,11 +1,13 @@
 /* @flow strict-local */
 
 import React, { PureComponent } from 'react';
+import { Alert } from 'react-native';
 
+import { TranslationContext } from '../boot/TranslationProvider';
 import type { RouteProp } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
 import * as NavigationService from '../nav/NavigationService';
-import type { Dispatch } from '../types';
+import type { GetText, Dispatch } from '../types';
 import { connect } from '../react-redux';
 import { hasAuth, getAccountStatuses } from '../selectors';
 import type { AccountStatus } from './accountsSelectors';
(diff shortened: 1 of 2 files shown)


src/account/AccountPickScreen.js
```diff
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 
 import React, { PureComponent } from 'react';
-
```
chrisbobbe, line 4: nit: Keep a blank line, and keep third-party imports (the one from 'react' and the new one from 'react-native') above the blank line.

src/account/AccountPickScreen.js
```diff
@@ -36,7 +36,20 @@ class AccountPickScreen extends PureComponent<Props> {
   };
 
   handleAccountRemove = (index: number) => {
-    this.props.dispatch(removeAccount(index));
+    Alert.alert(
+      'Remove Account?',
```
chrisbobbe, line 40: Let's be sure to wire these new user-facing strings up for translation, by using `_` (see the jsdoc on our `GetText` type for how to get access to that in a class component, and examples like `UserStatusScreen` or `LanguagePicker`) and adding the new strings to messages_en.json.

Also, let's go for the same capitalization, punctuation, and button style as we use for the logout confirmation in src/account-info/ProfileScreen.js.

chrisbobbe (changes requested):
Thanks, @AlaaElden98! I'll be glad to see a confirmation here; it seems disruptive for the device to totally forget about an account without checking to make sure that's actually what the user wanted to do.

I have a commit message nit:

```
AccountPickScreen: Ask for confirmation

Ask for confirmation when the user press on the remove button

Fixes part of: #4651
```

The summary line has room to be more specific than that; what do you think about something like `AccountPickScreen: Confirm before removing account from device.`? Then the "Ask for confirmation..." line in the body can disappear because it's already been said in the summary.

chrisbobbe:
Thanks! Merged, with a few tweaks; see the commits I added after yours, and also I put a note mentioning #4651 back into your commit. 🙂

AlaaElden98:
> Thanks! Merged, with a few tweaks; see the commits I added after yours, and also I put a note mentioning #4651 back into your commit. 🙂

Noted, thanks!  
Is the rules of adding blank lines between codes exists in our docs 🤔

gnprice:
> Is the rules of adding blank lines between codes exists in our docs

The basic rule here is to use blank lines to separate things that don't belong grouped together. There isn't a formula for what belongs grouped together, so it's hard to codify in a rule. In general a good practice is to look at other code (in the same file, or the same codebase) and follow what it's doing.

gnprice:
One source that may be helpful to read through is the Google JavaScript style guide. We don't necessarily align with all the choices they make, but probably the large majority of what's there is good advice in our codebase or any other JS codebase. Here's what they say about blank lines:
https://google.github.io/styleguide/jsguide.html#formatting-vertical-whitespace
and it's pretty close to what we do, and in particular explains why we have one here.

PR #4662, opened 2021-04-15
Replace switch with pressable-icon and rename props,Fixes: #4658

### LINKED ISSUE #4658: Improve UI for "All streams" subscription status
The "All streams" view currently uses slider toggles to indicate stream subscription status and allow the user to change it. This UI feels off, because toggles are generally used for settings, and the stream itself is not a setting.

We should try changing the UI to something that feels more appropriate. A good candidate is using "+" for streams the user is not subscribed to, and a checkmark for streams the user is currently subscribed to. See screenshots below.

Related Zulip thread: https://chat.zulip.org/#narrow/stream/48-mobile/topic/.22All.20streams.22.20subscribe.20toggles

Current Zulip UI:
toggles

Alternative pattern (screenshot from Pocket Casts podcast app):
Screenshot_20210411-150500_Pocket-Casts

vaibhavs2:
Fix: #4658 
I tried [pressable](https://reactnative.dev/docs/pressable) here,

|||

- The image below, if I use padding

-  Still the sensible area is equal to above(padding) which will recieve touch event (for icon-button). I used [hitSlop](https://reactnative.dev/docs/pressable#hitslop) to do it(without padding)

### DIFF
--- a/src/common/ZulipSubscribe.js
+++ b/src/common/ZulipSubscribe.js
@@ -1,0 +1,41 @@
+/* @flow strict-local */
+import React, { useState } from 'react';
+import { Pressable } from 'react-native';
+import { IconPlus, IconDone } from './Icons';
+import { BRAND_COLOR } from '../styles';
+
+type Props = $ReadOnly<{|
+  isSubscribed: boolean,
+  disabled: boolean,
+  onPress: (arg: boolean) => void,
+|}>;
+
+export default function ZulipSubscribe(props: Props) {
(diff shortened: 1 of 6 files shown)


src/common/ZulipSubscribe.js
```diff
@@ -0,0 +1,41 @@
+/* @flow strict-local */
+import React, { useState } from 'react';
+import { Pressable } from 'react-native';
```
AlaaElden98, line 3: https://github.com/zulip/zulip-mobile/pull/4654#discussion_r612593879

PR #4664, opened 2021-04-15
ios: Drop iOS 11 support; require iOS 12.1+.

chrisbobbe:
ios: Drop iOS 11 support; require iOS 12.1+.

We'd like to use `Array.prototype.flatMap`, which it seems won't
work in React Native on iOS until iOS 12. This was noted in a
comment [1] on facebook/react-native#26283. While the reasoning
there wasn't as thorough as we'd like, we looked closer [2], and we
believe the conclusion is correct.

A fresh look at iOS installations by platform (or the proxy we use
for that, anyway) [3] shows iOS 11 usage is 0.1% of our iOS usage,
so this should be fine to do. Also, 5.9% of our iOS users are on iOS
<= 13, so graceful degradation at that level is fine.

[1] https://github.com/facebook/react-native/issues/26283#issuecomment-607615346
[2] https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/1163034
[3] https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/1164857


chrisbobbe:
(For comparison, c953bc3367343df8b204d37173e5f809ad9db5b7 was the commit in which we dropped iOS 10 support.)

gnprice:
Thanks @chrisbobbe ! Small comments:

* Let's also update `tools/generate-webview-js`; see 84fae481c. I think it won't actually change the output (because our minimum Chrome version is rather older), but good to keep in sync.
* In the "history" entry in platform-versions.md, I've liked having links to the chat threads where we made the decision; sometimes handy for looking those up, and easy to do at the time we're making the change.
* In the commit message: I don't feel like our reasoning on the "RN uses the system JSC" point is totally thorough either, so I'd just drop that bit.

chrisbobbe:
> * Let's also update `tools/generate-webview-js`; see [84fae48](https://github.com/zulip/zulip-mobile/commit/84fae481ce2ca78aa1fc8de50d4eb71c70dc2300). I think it won't actually change the output (because our minimum Chrome version is rather older), but good to keep in sync.

Goodness, right! Thanks for the reminder; I see you carefully made sure there were bidirectional comments, and I looked right past them, only focusing on what I'd done in my own commit last time. 🙂

docs/architecture/platform-versions.md
```diff
@@ -91,11 +91,15 @@ History:
 * We dropped iOS 10 support in 2020-10. It was 0.3% of iOS users who
   tried Zulip, and we wanted to use a feature introduced in iOS 11,
   called "named colors".
+* We [dropped iOS 11 support][] in 2021-04. It was 0.1% of iOS users
```
WesleyAC, line 94: nit: the empty `[]` doesn't do anything here, I don't think

docs/architecture/platform-versions.md
```diff
@@ -91,11 +91,15 @@ History:
...
+  who tried Zulip. We started iOS 12 support at 12.1 because Xcode's
+  dropdown for "Deployment Target" didn't have a 12.0.
```
WesleyAC, line 96: Could you add a comment here like:

```
<!-- When updating this, please update `docs/developer-guide.md` as well. -->
```

Alternatively, we might want to consider just having fewer places where we keep this information :) I'd be fine with removing it from `docs/developer-guide.md` altogether, or keeping a link to this doc but not explicitly writing down the versions.

WesleyAC (approved):
Two minor comments, but looks good other than those — feel free to merge once you've fixed them.

chrisbobbe:

gnprice:
Thanks!

> * In the commit message: I don't feel like our reasoning on the "RN uses the system JSC" point is totally thorough either, so I'd just drop that bit.

FTR, what I meant here (and, rereading, I see this was thoroughly unclear in my comment) was that the reasoning that we want to use `flatMap` and we believe that's only available on iOS 12+ is fine -- just that I'd like to drop the part about how the reasoning in that other comment wasn't thorough.

Issue #4657, opened 2021-04-13
Migrate from `realm_filters` event format to new `realm_linkifiers` format


timabbott:
In https://github.com/zulip/zulip/pull/17596, we changed the format for `realm_filters` events to a new `realm_linkifiers` format that is more rational and extensible.  We should migrate clients to this new format.  (See the backwards-compatibility notes in the API documentation, but I expect a correct implementation will choose whether to send `realm_filters` or `realm_linkifiers` in `event_types` when doing a `register` post, and then have handled code for `GET /events` for both formats).

(This will be an important prerequisite for adding stream-specific linkifiers in the future).

chrisbobbe:
As I understand it, https://github.com/zulip/zulip-mobile/issues/4242 is needed for this. We do one thing with `realm_filters`, which is pass it to a `parseMarkdown` function for local echo. That function was forked from the web app in 2017 and naturally hasn't been kept up-to-date.

Once #4242 is done, we can take the new event format—presumably the shared code we use will be able to handle it. The next step for #4242 is to respond to my review of his PR zulip/zulip-markdown-parser#5.

timabbott:
I'm not sure this needs to block on that -- can't we just write a block of event handler code that takes a `realm_linkifiers` object and turns it into a `realm_filters` style tuple before passing the data to the markdown implementation?  

(And then when moving to a shared markdown implementation, one would potentially switch that around?)

I would expect that as a general rule, we want to handle feature level logic in the code directly adjacent to the network (E.g. a first step after receiving responses from the server) -- so that other code can be clean and simple (either against the old API, or the new, depending on how we want to convert the data).

chrisbobbe:
> The format change is I think just `obj => [obj.pattern, obj.url_format_string, obj.id]`

Interesting; yeah, if the new `realm_linkifiers` objects can be converted to old-format `realm_filters` ones, this shouldn't have to block.

chrisbobbe:
Just sent #4666.

PR #4666, opened 2021-04-16
events: Accept new `realm_linkifiers` format.

chrisbobbe:
Fixes: #4657

### DIFF
--- a/src/api/registerForEvents.js
+++ b/src/api/registerForEvents.js
@@ -50,6 +50,18 @@ const transformCrossRealmBot = (
 
 const transform = (rawInitialData: RawInitialData, auth: Auth): InitialData => ({
   ...rawInitialData,
+
+  // Transform the newer `realm_linkifiers` format, if present, to the
+  // older `realm_filters` format. We do the same transformation on
+  // 'realm_linkifiers' events.
+  realm_filters: rawInitialData.realm_linkifiers
+    ? rawInitialData.realm_linkifiers.map(({ pattern, url_format, id }) => [
+        pattern,
+        url_format,
+        id,
+      ])
+    : rawInitialData.realm_filters,
+
   realm_users: rawInitialData.realm_users.map(rawUser => transformUser(rawUser, auth.realm)),
   realm_non_active_users: rawInitialData.realm_non_active_users.map(rawNonActiveUser =>
     transformUser(rawNonActiveUser, auth.realm),
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -313,8 +313,18 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
   try {
     [initData, serverSettings] = await Promise.all([
       tryFetch(() =>
+        // Currently, no input we're giving `registerForEvents` is
+        // conditional on the server version / feature level. If we
+        // need to do that, make sure that data is up-to-date -- we've
+        // been using this `registerForEvents` call to update the
+        // feature level in Redux, which means the value in Redux will
+        // be from the *last* time it was run. That could be a long
+        // time ago, like from the previous app startup.
         api.registerForEvents(auth, {
+          // Event types not supported by the server are ignored; see
+          //   https://zulip.com/api/register-queue#parameter-fetch_event_types.
           fetch_event_types: config.serverDataOnStartup,
+
           apply_markdown: true,
           include_subscribers: false,
           client_gravatar: true,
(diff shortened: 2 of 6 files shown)


WesleyAC:
It seems like the first commit here will likely increase the time that the initial fetch takes. Is that the case? If so, how significant is the impact?

timabbott:
I guess the alternative implementation would be to include `realm_filters` and `realm_linkifiers` in the requested event types and then just ignore `realm_filters` events if the feature level came back high enough.

chrisbobbe:
> (Or we could change the server to check if the client accepts `realm_linkifiers` events when deciding whether to deliver a `realm_filters` event?).

For this particular event, I don't think we need to do that; I think mobile is happy to respond to both events as they're delivered. Both event types give the entire list of linkifiers, so there's no issue of potentially pushing the same item to a list twice; we'll just overwrite the list one more time than necessary, and that doesn't seem like a performance concern given how uncommonly people will be changing the linkifiers.

chrisbobbe:
> It seems like the first commit here will likely increase the time that the initial fetch takes. Is that the case? If so, how significant is the impact?

Not trivial, I think; am I doing this right?

```
chrisbobbe:~/dev/zulip-mobile (master u+1-9) 
$ time curl -s https://chat.zulip.org/api/v1/server_settings > /dev/null

real    0m0.315s
user    0m0.018s
sys     0m0.007s
chrisbobbe:~/dev/zulip-mobile (master u+1-9) 
$ time curl -s https://chat.zulip.org/api/v1/server_settings > /dev/null

real    0m0.327s
user    0m0.018s
sys     0m0.006s
chrisbobbe:~/dev/zulip-mobile (master u+1-9) 
$ time curl -s https://chat.zulip.org/api/v1/server_settings > /dev/null

real    0m0.279s
user    0m0.018s
sys     0m0.006s
chrisbobbe:~/dev/zulip-mobile (master u+1-9) 
$ time curl -s https://chat.zulip.org/api/v1/server_settings > /dev/null

real    0m0.946s
user    0m0.021s
sys     0m0.007s
chrisbobbe:~/dev/zulip-mobile (master u+1-9) 
$ time curl -s https://chat.zulip.org/api/v1/server_settings > /dev/null

real    0m0.489s
user    0m0.019s
sys     0m0.006s
```

chrisbobbe:
OK, revision pushed—I've removed the commit that has us wait for `/server_settings` before starting `/register`, so that slowdown is no longer an issue. 🙂

src/message/fetchActions.js
```diff
@@ -310,8 +310,18 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
...
+        // Currently, no input we're giving `registerForEvents` is
+        // conditional on the server version / feature level. If we
+        // need to do that, make sure that data is up-to-date -- we've
+        // been using this `registerForEvents` call to update the
+        // feature level in Redux, which means the value in Redux will
+        // be from the *last* time it was run. That could be a long
+        // time ago, like from the previous app startup.
         api.registerForEvents(auth, {
+          // Event types not supported by the server are ignored; see
+          //   https://zulip.com/api/register-queue#parameter-fetch_event_types.
```
WesleyAC, line 322: The commit message about this not being deployed yet can go now :)

src/api/registerForEvents.js
```diff
@@ -50,6 +50,18 @@ const transformCrossRealmBot = (
...
+  // Transform the newer `realm_linkifiers` format, if present, to the
+  // older `realm_filters` format. We do the same transformation on
+  // 'realm_linkifiers' events.
+  realm_filters: rawInitialData.realm_linkifiers
+    ? rawInitialData.realm_linkifiers.map(({ pattern, url_format, id }) => [
+        pattern,
+        url_format,
+        id,
+      ])
+    : rawInitialData.realm_filters,
```
WesleyAC, line 63: Do we want to make it so that if the server doesn't send either back, it's `[]`, or does that not matter?

WesleyAC (commented):
A couple small comments, looks good other than that!

src/api/registerForEvents.js
chrisbobbe: I think if the server didn't send either back, it would be a server bug (as long as we didn't suddenly stop requesting them for some reason). That might need some defensive code, but I'd be inclined to investigate that at the time it's discovered.

chrisbobbe:

I also made some changes to the comments, which only occurred to me while working on this revision. Mostly changing which comments are implementation comments (with `//`) and which ones are interface comments (jsdoc). Rather than wait for another review on those, I've gone ahead and merged, since we have a practice of editing comments wiki-style. (Please let me know if you happen to see something wrong, though!)

gnprice:
Thanks for implementing this, and thanks @WesleyAC for the reviews! Looks good to me.

Issue #3998, opened 2020-03-24
Add sender_id and stream_id properties to Outbox


gnprice:
We have a type called `Outbox` that's meant to be a lot like `Message`. It represents messages the user has hit "send" on from the compose box, and that we're still in the process of actually getting to the server (and then hearing back from the server as new messages.)

It doesn't quite have all the same properties as `Message` has, which is fine -- we don't use all those properties.

But in particular, while it has a `sender_email: string`, it does not have `sender_id: number`. This is basically a case of #3764.

Similarly it has this property:
```
  display_recipient: $FlowFixMe, // `string` for type stream, else PmRecipientUser[].
```
which, just like on `Message`, is for a stream message the name of the stream. But it lacks this other property which `Message` has:
```
  stream_id: number, // FixMe: actually only for type `stream`, else absent.
```
So this is basically a case of #3918.

We should add those two properties. Then all Outbox objects (outside tests) are created by `addToOutbox`, so we should fill the properties in appropriately there.

The scope of this issue does *not* include switching to actually use them yet -- and in fact we can't immediately, without some more work. The reason is that we persist outbox messages between runs of the app, so a new version of the app could get `Outbox` values stored by an old version that didn't fill in these properties.

---

Fortunately, outbox messages are inherently perishable -- we already discard them at one week of age, figuring that's better than sending a message which may be badly out of date and out of context. So, here's a low-effort way we can solve this problem.

First, let's:
* Add the two properties to `Outbox`, as *optional*.
* Start filling them in for all `Outbox` values we create.
* Next to the two properties' definitions in the `Outbox` type, have a TODO comment saying we introduced them at such-and-such time.
* Close this issue as having been fixed. :tada: 

Then we wait until a release with the above changes has been out a month or two.

After that, at our leisure, we:
* Switch the properties to required. Add a migration that simply deletes any outbox messages where they're missing.
  * This will only affect users who upgrade from the old version before the above changes, straight to this version, within a week -- and moreover have a message they tried to send on the old version that's still hanging around, unsent, when they start this new version.
  * I expect almost all users, almost all of the time, have zero outbox messages hanging around between runs of the app. So with this timeline, my hope is to reduce to a probable zero the number of users for whom this migration has any effect at all.
  * And if it's not quite zero, that's not a big deal. Once a message is hanging around unsent for more than a few minutes, it's already not an awesome experience even if we do go on to send it. And we already discard such messages if you do something like switch to another account and back, which is much more common than this migration.
[...]

PR #4667, opened 2021-04-16
Outbox type: Make `sender_id` required; add optional `stream_id`.

chrisbobbe:
Finish most of #3998! 🎉

- Make `sender_id` required, since the change that added it as optional has been out in a release for a while
- Add `stream_id`, as optional, so the process starts again: once _that's_ been out for a few weeks, we should make that required, and that'll finish of #3998.

Also start type-checking another test file.


src/__tests__/lib/exampleData.js
```diff
@@ -444,6 +444,9 @@ export const makeOutboxMessage = (data: $Shape<$Diff<Outbox, {| id: mixed |}>>):
   const outputTimestamp = timestamp ?? makeTime() / 1000;
   return deepFreeze({
     ...outboxMessageBase,
+    /* $FlowFixMe[exponential-spread] - Yeah, we should probably deal
```
WesleyAC, line 447: Can you make a ticket to track this, or do it in this PR?

I guess it's not a huge deal, since this is just a problem with typechecking performance, not runtime performance, but it does seem like something we should fix.

WesleyAC (approved):
LGTM, except for either tracking or actually fixing the `exponential-spread` warning.

Thanks for doing this! Having the `stream_id` in `Outbox` I think will let us get rid of `stream_name` in the actionsheet code, which I'm excited about :)

chrisbobbe:

For the Flow error, it seemed sufficient to tweak the type of `outboxMessageBase` to be based on `OutboxBase` rather than `Outbox`. So I did that and changed its jsdoc accordingly.

WesleyAC:
This looks good to me, feel free to merge after you fix the conflict :)

chrisbobbe:
> This looks good to me, feel free to merge after you fix the conflict :)

Thanks for the review! Done; I also added a commit at the beginning to put a quick comment on migration 28.

src/types.js
```diff
@@ -208,6 +194,51 @@ export type Outbox = $ReadOnly<{|
...
+
+  ...SubsetProperties<PmMessage, {||}>,
+|};
+
+export type StreamOutbox = {|
+  ...OutboxBase,
+
+  // TODO(#3764): Make stream_id required.  Needs a migration to drop
+  //   StreamOutbox values that lack it; that'll be fine once the
+  //   release that adds it has been out for a few weeks. (Also drop
```
gnprice, line 208: There's another key step missing before we can do this: we need to start actually populating this property in the values we create for `state.outbox` :wink: See a1fad7ca9 for comparison.

gnprice (commented):
Thanks @chrisbobbe and @WesleyAC ! Glad to have this taken care of.

I've just pushed the following commits following up on these changes:
6fe63e3e2 outbox types: Mark PmOutbox and StreamOutbox as read-only.
8171f2d7b outbox types [nfc]: Fix TODO comment on adding stream_id.
606591b63 outbox types [nfc]: Add TODO comments on PmOutbox vs. StreamOutbox.

One is trivial. The others are comment-only changes, corresponding to the comments below -- take a look.

I've also drafted up a fix for one of these, completing the distinction between `PmOutbox` and `StreamOutbox`. Branch here:
https://github.com/zulip/zulip-mobile/compare/master...gnprice:dev-outbox-types
but it needs some cleaning-up before merge, and it's late here now and I should go to bed :slightly_smiling_face:

chrisbobbe:
Thanks for catching those things, @gnprice! Please let me know if you'd like me to take over https://github.com/zulip/zulip-mobile/compare/master...gnprice:dev-outbox-types; otherwise I'll take a look when you make it into a PR. 🙂

PR #4670, opened 2021-04-16
Clarify some message-list code (part 2).

chrisbobbe:
EDIT: See https://github.com/zulip/zulip-mobile/pull/4670#issuecomment-821586117 where I mention the cause of the CI failures.

-----

A continued resumption of #3913 (which has grown stale), toward #3851. Part 1 was #4406 .

This PR focuses on the refactor to flatten the funny shape of `HtmlPieceDescriptor[]`, and it has tests specifically aimed at confirming the refactor.

Those tests are snapshot tests, something I think it could be neat to explore. I'd like to know if there's a better file I can put them in (I've made a `contentHtmlFromPieceDescriptors-test.js` and put them there); they test how `getHtmlPieceDescriptors` and `contentHtmlFromPieceDescriptors` work together, not separately. But I don't think I run on enough caffeine to work with a filename that concatenates those names 🤔.

### DIFF
--- a/src/message/getHtmlPieceDescriptors.js
+++ b/src/message/getHtmlPieceDescriptors.js
@@ -11,7 +11,7 @@ export default (
   const showHeader = !isPmNarrow(narrow) && !isTopicNarrow(narrow);
 
   let prevMessage = undefined;
-  const sections = [{ key: 0, data: [], message: null }];
+  const pieces: HtmlPieceDescriptor[] = [];
   messages.forEach(message => {
     const diffDays =
       !!prevMessage
@@ -23,15 +23,15 @@ export default (
         timestamp: message.timestamp,
         subsequentMessage: message,
       };
-      sections[sections.length - 1].data.push(renderedTimeDescriptor);
+      pieces.push(renderedTimeDescriptor);
     }
 
     const diffRecipient = !isSameRecipient(prevMessage, message);
     if (showHeader && diffRecipient) {
-      sections.push({
+      pieces.push({
+        type: 'header',
         key: `header${message.id}`,
-        message,
-        data: [],
+        subsequentMessage: message,
       });
     }
 
@@ -44,9 +44,9 @@ export default (
       isBrief: shouldGroupWithPrev,
       message,
     };
-    sections[sections.length - 1].data.push(renderedMessageDescriptor);
+    pieces.push(renderedMessageDescriptor);
 
     prevMessage = message;
   });
-  return sections;
+  return pieces;
 };
--- a/src/message/getHtmlPieceDescriptors.js
+++ b/src/message/getHtmlPieceDescriptors.js
@@ -17,12 +17,13 @@ export default (
       !!prevMessage
       && !isSameDay(new Date(prevMessage.timestamp * 1000), new Date(message.timestamp * 1000));
     if (!prevMessage || diffDays) {
-      sections[sections.length - 1].data.push({
+      const renderedTimeDescriptor = {
         key: `time${message.timestamp}`,
         type: 'time',
         timestamp: message.timestamp,
         subsequentMessage: message,
-      });
+      };
+      sections[sections.length - 1].data.push(renderedTimeDescriptor);
     }
 
     const diffRecipient = !isSameRecipient(prevMessage, message);
@@ -37,12 +38,13 @@ export default (
     const shouldGroupWithPrev =
       !diffRecipient && !diffDays && !!prevMessage && prevMessage.sender_id === message.sender_id;
 
-    sections[sections.length - 1].data.push({
+    const renderedMessageDescriptor = {
       key: message.id,
       type: 'message',
       isBrief: shouldGroupWithPrev,
       message,
-    });
+    };
+    sections[sections.length - 1].data.push(renderedMessageDescriptor);
 
     prevMessage = message;
   });
--- a/src/message/getHtmlPieceDescriptors.js
+++ b/src/message/getHtmlPieceDescriptors.js
@@ -19,13 +19,12 @@ export default (
       !!prevMessage
       && !isSameDay(new Date(prevMessage.timestamp * 1000), new Date(message.timestamp * 1000));
     if (!prevMessage || diffDays) {
-      const renderedTimeDescriptor = {
+      pieces.push({
         key: `time${message.timestamp}`,
         type: 'time',
         timestamp: message.timestamp,
         subsequentMessage: message,
-      };
-      pieces.push(renderedTimeDescriptor);
+      });
     }
 
     const diffRecipient = !isSameRecipient(prevMessage, message);
@@ -40,13 +39,12 @@ export default (
     const shouldGroupWithPrev =
       !diffRecipient && !diffDays && !!prevMessage && prevMessage.sender_id === message.sender_id;
 
-    const renderedMessageDescriptor = {
+    pieces.push({
       key: message.id,
       type: 'message',
       isBrief: shouldGroupWithPrev,
       message,
-    };
-    pieces.push(renderedMessageDescriptor);
+    });
   }
   return pieces;
 };
--- a/src/message/getHtmlPieceDescriptors.js
+++ b/src/message/getHtmlPieceDescriptors.js
@@ -10,9 +10,11 @@ export default (
 ): $ReadOnlyArray<HtmlPieceDescriptor> => {
   const showHeader = !isPmNarrow(narrow) && !isTopicNarrow(narrow);
 
-  let prevMessage = undefined;
   const pieces: HtmlPieceDescriptor[] = [];
-  messages.forEach(message => {
+  for (let i = 0; i < messages.length; i++) {
+    const message = messages[i];
+    const prevMessage: typeof message | void = messages[i - 1];
+
     const diffDays =
       !!prevMessage
       && !isSameDay(new Date(prevMessage.timestamp * 1000), new Date(message.timestamp * 1000));
@@ -45,8 +47,6 @@ export default (
       message,
     };
     pieces.push(renderedMessageDescriptor);
-
-    prevMessage = message;
-  });
+  }
   return pieces;
 };
--- a/src/message/getHtmlPieceDescriptors.js
+++ b/src/message/getHtmlPieceDescriptors.js
@@ -7,7 +7,7 @@ import { isSameDay } from '../utils/date';
 export default (
   messages: $ReadOnlyArray<Message | Outbox>,
   narrow: Narrow,
-): HtmlPieceDescriptor[] => {
+): $ReadOnlyArray<HtmlPieceDescriptor> => {
   const showHeader = !isPmNarrow(narrow) && !isTopicNarrow(narrow);
 
   let prevMessage = undefined;
--- a/src/utils/date.js
+++ b/src/utils/date.js
@@ -3,12 +3,31 @@ import format from 'date-fns/format';
 import isToday from 'date-fns/is_today';
 import isYesterday from 'date-fns/is_yesterday';
 import isSameYear from 'date-fns/is_same_year';
+import addMinutes from 'date-fns/add_minutes';
 import tz from 'timezone/loaded';
 
 export { default as isSameDay } from 'date-fns/is_same_day';
 
+/**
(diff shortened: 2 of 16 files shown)


chrisbobbe:
(Blocked for the moment on #4664; I'm taking another look at that now. 🙂)

src/webview/html/__tests__/contentHtmlFromPieceDescriptors-test.js
```diff
@@ -0,0 +1,235 @@
...
+ *     update the snapshots, and commit the result.
+ *   - If they don't look correct, look for bugs that were caused or
+ *     revealed by your changes. Please fix them and run the tests
+ *     again.
+ *
+ * - Different input was given on this run of the tests, and naturally
+ *   our code produced different output.
+ *   - This will be the case when extending these tests to increase
+ *     coverage. Thanks for doing that! :)
+ *   - But one big "gotcha!" is that we can't vary the input
```
chrisbobbe, line 34: Is there another sub-point to go here? Representative values for our model data might have changed; e.g., if Gravatar suddenly told us to change the URL we use to request images, and we did that.

chrisbobbe:
Ah, hilarious. The CI failure looks like a timezone difference between where I run the tests and where CI runs them:

```diff
- Snapshot  - 3
+ Received  + 3

@@ -32,11 +32,11 @@

  <div class="subheader">
    <div class="username">
      Nonrandom name one User
    </div>
-   <div class="static-timestamp">7:15 PM</div>
+   <div class="static-timestamp">11:15 PM</div>
  </div>
```

chrisbobbe:
> Ah, hilarious. The CI failure looks like a timezone difference between where I run the tests and where CI runs them. I'll have to think of what to do about that.

I guess the thing to do (if we want to keep the snapshot tests) is set the timezone to be the same thing (UTC, probably) wherever the tests get run. But I think we might run into some issues on Windows if we use `process.env.TZ = 'UTC'` for that: https://github.com/nodejs/node/issues/4230

That issue has a comment mentioning the NPM package https://github.com/capaj/set-tz, which is for setting the timezone on all OSes. On Windows it says "it effectively changes your whole OS timezone so beware" so maybe that's not the best thing to do either. 🤔

chrisbobbe:
(Discussion continuing [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Snapshot.20tests/near/1167492).)

chrisbobbe:
Hmm, it's occurred to me that there are a lot of variables that the new snapshot tests could fruitfully test! Like whether consecutive messages were on the same day; whether they were sent by the same user or different users; what kind of narrow we're looking at; what flags messages have, etc.

Do we think we want to draw up a table of transformations we want to apply toward varying those things, then write code to automatically apply all combinations of those transformations? That'd mean tons of snapshots and being a bit overwhelmed with every snapshot failure (since lots of failures would happen at once).

chrisbobbe:
(Pausing work on this while awaiting discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4670.3A.20Snapshot.20tests.20for.20HTML.20generation/near/1179236).)

src/utils/date.js
```diff
@@ -3,12 +3,29 @@ import format from 'date-fns/format';
...
+// We're only counting on this if we can't set the timezone `Date` uses to
+// UTC in the first place; see note where we attempt to do that in
+// jest/globalSetup.js.
+function maybeAsUtc(date: Date) {
+  /* eslint-disable operator-linebreak */
+  /* eslint-disable-next-line id-match */
+  return __TEST__
+    ? // Negate the UTC offset, using an offset we get from `date` itself, so
+      // it'll be correct for DST-type changes through the year.
+      addMinutes(date, date.getTimezoneOffset())
```
gnprice, line 23: Have you confirmed empirically that this is the right way around? This is what someone suggested on an upstream issue thread, but it looked backward to me:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Snapshot.20tests/near/1167551
but I could have a sign flipped.

src/utils/date.js
```diff
@@ -3,12 +3,29 @@ import format from 'date-fns/format';
 import isToday from 'date-fns/is_today';
 import isYesterday from 'date-fns/is_yesterday';
 import isSameYear from 'date-fns/is_same_year';
+import addMinutes from 'date-fns/add_minutes';
 import tz from 'timezone/loaded';
 
 export { default as isSameDay } from 'date-fns/is_same_day';
 
+/**
+ * Use consistent timezone (UTC) in local and CI tests, for snapshot tests.
```
gnprice, line 12: I think this isn't quite accurate (because unfortunately the thing this function is doing / the thing the underlying APIs leave as the best we were able to find how to do is not as clean as this description). I'll push to the branch a commit which can go at the end, expanding this jsdoc.

src/message/getHtmlPieceDescriptors.js
```diff
@@ -7,17 +7,17 @@ import { isSameDay } from '../utils/date';
...
   messages: $ReadOnlyArray<Message | Outbox>,
   narrow: Narrow,
-): HtmlPieceDescriptor[] => {
+): $ReadOnlyArray<HtmlPieceDescriptor> => {
   const showHeader = !isPmNarrow(narrow) && !isTopicNarrow(narrow);
 
-  let prevMessage = undefined;
-  const sections = [{ key: 0, data: [], message: null }];
-  messages.forEach(message => {
+  return messages.flatMap((message, index, arr) => {
```
gnprice, line 13: I'm happy to see `prevMessage` turned into a `const` at the top of the loop body, instead of a piece of state that gets mutated.

But the thing about this `flatMap` approach is that it'll create a lot of little arrays (mostly of length 1) which promptly get thrown away; one of them per message. That's quite a bit of allocation, and allocation (plus the GC work it makes necessary) is typically a large fraction of the CPU work an application ends up having to do. That means it's likely that the `flatMap` version of this function is a noticeable constant factor slower -- I'd guess it's somewhere between 10% slower and 2x slower -- than a version that pushes directly onto a single array.

Because this codepath where we're computing what to show in the message list is a potentially performance-sensitive area, it's one where it's worth taking care with that kind of constant factor as long as it doesn't make the source code substantially worse.

I'll push a quick WIP commit onto the branch which converts this to a plain loop but keeps the non-stateful `prevMessage`. Then please go ahead and squash that into the appropriate spot earlier in the branch, and/or make similar changes in a different order.

src/message/getHtmlPieceDescriptors.js
```diff
@@ -7,17 +7,17 @@ import { isSameDay } from '../utils/date';
...
+    const prevMessage = arr[index - 1];
```
gnprice, line 14: This makes sense but unfortunately Flow doesn't notice that it can be `undefined`.

(That's an understandable design choice on Flow's part -- it's unusual to have an array access that can actually be out of bounds like this.)

For that reason, probably best to explicitly annotate it so that Flow can check we handle the `undefined` case reasonably in the code below.

gnprice (commented):
Thanks @chrisbobbe ! These piece descriptors are some legacy code that I'm happy to see cleaned up like this. Also interested to see how the snapshot tests work out for us over time.

Comments below. As mentioned below, I'll also push a couple of additional commits for where showing seems easier/clearer than telling.

src/utils/date.js
chrisbobbe: I did, but looks like I didn't write it down. At the tip of this revision (er, actually after the squashing you mentioned at https://github.com/zulip/zulip-mobile/pull/4670#discussion_r647965060; that shouldn't matter), I just did the following:

In `jest/globalSetup.js`:

```diff
-  process.env.TZ = 'UTC';
+  process.env.TZ = 'America/New_York';
```

Then I ran `tools/test jest`, and got only this failure (which I'd expect):

```
Summary of all failing tests
 FAIL  src/utils/__tests__/date-test.js
  ● Timezones (for test environment only) › Timezone has been set to UTC, unless on Windows

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

       8 |         // https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Snapshot.20tests/near/1167560
       9 |         || process.platform === 'win32',
    > 10 |     ).toBe(true);
         |     ^
      11 |   });
      12 | });
      13 | 

      at Object.<anonymous> (src/utils/__tests__/date-test.js:10:5)


Test Suites: 1 failed, 55 passed, 56 total
Tests:       1 failed, 1 skipped, 1706 passed, 1708 total
Snapshots:   36 passed, 36 total
Time:        28.567 s, estimated 41 s
Ran all test suites related to changed files.
```

Then, I made this change in `maybePretendUtc` in `src/utils/date.js`:

```diff
-     addMinutes(date, date.getTimezoneOffset())
+     addMinutes(date, -date.getTimezoneOffset())
```

I ran `tools/test jest` again (er actually `tools/test jest > jest.txt 2>&1`), and I got that failure plus lots of others: 
[jest.txt](https://github.com/zulip/zulip-mobile/files/6627015/jest.txt)

src/utils/date.js
chrisbobbe: Ah, yeah, thanks for doing that. It's tricky! 🤕

chrisbobbe:
Thanks for the review and those commits! Revision pushed.

src/message/getHtmlPieceDescriptors.js
```diff
@@ -12,7 +12,9 @@ export default (
 
   let prevMessage = undefined;
   const pieces: HtmlPieceDescriptor[] = [];
-  messages.forEach(message => {
+  for (let i = 0; i < messages.length; i++) {
```
WesleyAC, line 14: This commit message:

```
 getHtmlPieceDescriptors [nfc]: Use a for loop. 
```

Should mention why you're doing this. Or just squash it into the next commit :)

chrisbobbe: Hmm, yeah. Greg made it a `for` loop in his WIP commits, so I kept that, but it'd totally be possible to keep a `forEach` because the callback receives the index of the item. I'm inclined to go back to `forEach` (not `flatMap` though) in my next revision; @gnprice, what do you think?

chrisbobbe:

src/message/getHtmlPieceDescriptors.js
chrisbobbe: > Or just squash it into the next commit :)

I've opted not to do this; I think it might have muddled the next commit more than it clarified, and it would leave the change unexplained.

src/utils/date.js
gnprice: So be it! Thanks for confirming.

src/message/getHtmlPieceDescriptors.js
gnprice: The loop structure in the new revision LGTM. (I haven't looked at what the commit sequence looked like at the start of this subthread.)

```js
  messages.forEach((message, i, arr) => {
    const prevMessage = arr[i - 1];
```

It's just a little awkward because `message` and `prevMessage` aren't gotten in a symmetrical way: we get the latter from indexing into `arr` using `i`, but the former directly as an argument to the callback. It's implicit that `message` is the same as `arr[i]`, but the parallelism between that and `arr[i - 1]` doesn't appear explicitly.

I think that's why I went for a plain `for` loop in my demo. But this is fine too.

src/message/getHtmlPieceDescriptors.js
gnprice: Bump :slightly_smiling_face:

gnprice:
Thanks for the revision!

All LGTM except the one small item at https://github.com/zulip/zulip-mobile/pull/4670#discussion_r654678980 . Please merge at will after that.

src/message/getHtmlPieceDescriptors.js
chrisbobbe: Oh OK, got it! Thanks for that explanation; it makes sense.

chrisbobbe:
Thanks for the review! Merged, with that tweak, and one other: in the first commit, we were using `__TEST__`; I changed this to `global.__TEST__` because I just discovered we were getting a `ReferenceError: __TEST__ is not defined` when running the app and opening the message list.

PR #4672, opened 2021-04-17
queueMarkAsRead fixes

WesleyAC:
Two fixes from #4620, as requested by @gnprice.

### DIFF
--- a/src/api/queueMarkAsRead.js
+++ b/src/api/queueMarkAsRead.js
@@ -16,7 +16,7 @@ export const resetAll = () => {
   timeout = null;
 };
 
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @WesleyAC ! Merged.

I made one small fix in a commit message:
```
    messagesFlags API call, but using a Set, we can reduce it to
    O(unackedMessageIds.length + ackedMessageIds.length).
```

This had said `O(unackedMessageIds.length)`, but there's the time building the `Set` too, so if the acked list is longer than the unacked for some reason (some kind of race?) then it could be more than that.

PR #4677, opened 2021-04-19
make statusBar visible in landscape mode with matching color, Fixes:#4668

### LINKED ISSUE #4668: statusBar color not matches when orientation is in landscape mode

vaibhavs2:
Fixes: #4668 

horizontal

stable

### DIFF
--- a/src/common/ZulipStatusBar.js
+++ b/src/common/ZulipStatusBar.js
@@ -1,13 +1,13 @@
 /* @flow strict-local */
 
 import React, { PureComponent } from 'react';
-import { Platform, StatusBar } from 'react-native';
+import { StatusBar } from 'react-native';
 import Color from 'color';
 
-import type { Orientation, ThemeName, Dispatch } from '../types';
+import type { ThemeName, Dispatch } from '../types';
 import { connect } from '../react-redux';
 import { foregroundColorFromBackground } from '../utils/color';
-import { getSession, getSettings } from '../selectors';
+import { getSettings } from '../selectors';
 
 type BarStyle = $PropertyType<React$ElementConfig<typeof StatusBar>, 'barStyle'>;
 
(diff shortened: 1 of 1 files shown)


src/common/ZulipStatusBar.js
```diff
@@ -55,27 +54,26 @@ class ZulipStatusBar extends PureComponent<Props> {
   };
 
   render() {
-    const { theme, hidden, orientation } = this.props;
+    StatusBar.setTranslucent(true);
```
chrisbobbe, line 57: Interesting; what's the reason this is included in the same commit vs. a different commit? They might go together; it just hasn't been made obvious to me.

Also, what's the reason for using `.setTranslucent` instead of using the [`translucent` prop](https://reactnative.dev/docs/statusbar#translucent-android)?

chrisbobbe (commented):
Thanks, @vaibhavs2! See a comment below. Also, are you able to test on iOS, or shall I? A macOS is needed for that.

src/common/ZulipStatusBar.js
vaibhavs2: Its my fault, I just saw and used it, both caters the same purpose and it would be better to use it as prop.
And about
>same commit vs. a different commit?

Its been around 18 days, as I remember I couldn't found a commit message to describe changes that i removed a condition that restrict statusbar to be visible in landscape mode in android but it was not catering the purpose and in landscape mode default status-bar had been visible.
For the next, `translucent` will make lightbox visible in full screen without jumping, [CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/status-bar/near/1164588)
I will do it :) .

I will be very thankfull to you, If you test it on IOS. I don't have macOS

vaibhavs2: mentioning you @chrisbobbe  :) would be a good way to notify, thank you

Issue #4655, opened 2021-04-13
Add support for muted users


timabbott:
We've finished merging the backend for muting users in https://github.com/zulip/zulip/pull/16915.

This is an important feature for large open communities.

https://chat.zulip.org/api/mute-user is relevant for when mobile wants to implement muting, but the part that is likely to be more important is having the mobile apps hide messages from muted users.  We should update this issue with a link to the Help Center article explaining the behavior once that's complete, but in short:
* Messages sent by muted users are automatically marked as read (implemented in server).
* All prior messages sent by muted users are marked read when they are muted (implemented in server).
* Clients should arrange to avoid displaying messages sent by muted users in stream/huddle views, and hide such messages in some way.  (Details not fully settled and will likely evolve on the server).
* Muted users should not appear in lists of PM conversations / presence lists.

gnprice:
See [chat discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/muted.20users/near/1163894).

In particular here's my description of what'll be involved in the implementation:

Basically the plumbing work will all look like sort of a zulip-mobile version of the server/webapp's new-feature tutorial:
* There's some new data in the `/register` response. Add it in `src/api/initialDataTypes.js`.
* The new data needs a new key to be mentioned in `fetch_event_types`. Add that.
* We'll need to keep the new data in Redux. In this case, it can go in `state.realm` -- that's not a good name, but it's generally where we put miscellaneous small pieces of data that don't live as part of some other object like a particular user or stream or message.
  * Add an appropriate type for the data to the relevant part of the `GlobalState` type, in `src/reduxTypes.js`.
    * Often the appropriate type will be different from the wire format found in the server API. Here, probably `Immutable.Map<UserId, number>` is good, where the values are the `timestamp` fields from the API's format. That one is good because we can efficiently look a user up in it and efficiently make changes.
  * Add code to the reducer to give an initial empty value, and to read the `/register` response data from a `REALM_INIT` action and put that in the new data structure in the Redux state.
* We'll need to handle the events.
  * Add a type in `src/api/eventTypes.js`, following the API docs.
  * Add an event action type in `src/actionTypes.js` and `src/actionConstants.js`, following the example of `EventUserStatusUpdateAction`.
  * Add it to `eventToAction.js` following the examples in `actionTypeOfEventType`.
  * Now add code to the reducer to handle that action type by updating the data structure.
* Write tests for the new reducer code.
[...]

PR #4678, opened 2021-04-19
Start implementing muted users

WesleyAC:
Related: #4655

Tested by opening the app, and running the following commands in my terminal:

```
curl -sSX POST https://chat.zulip.org/api/v1/users/me/muted_users/19335 -u wesleyac@zulip.com:MY_API_KEY
curl -sSX DELETE https://chat.zulip.org/api/v1/users/me/muted_users/19335 -u wesleyac@zulip.com:MY_API_KEY
```

And using the debug console to verify that the correct data got added to the redux state.


src/mute/__tests__/mutedUsersReducer-test.js
```diff
@@ -0,0 +1,84 @@
...
+      const expectedState = Immutable.Map([
+        [makeUserId(42), 1618822632],
+        [makeUserId(1234), 1618822635],
+      ]);
+
+      const newState = mutedUsersReducer(initialState, action);
+
+      expect(newState).toEqual(expectedState);
+    });
+  });
```
gnprice, line 83: So this is a style that you'll see in a lot of our legacy tests, but I think the verbosity of it isn't helpful. It's possible to write all the same content with a lot less boilerplate; when we do, I think that (a) makes it easier to see what the test is really saying, and (b) makes it easier to scan through all the tests in the file and see what conditions are covered and think about what might not be.

For good examples to follow, you can look at tests written relatively recently by me or Chris. Scanning through `git log --stat -p --author 'greg\|cbobbe' -- src/**/*-test.js` and looking for newly-written tests (i.e. scanning past where we're just touching an existing test), there's 0a218e04fe70be7696001f6db7f6783b5d829944 as a nice though short example. Another good example is `src/session/__tests__/sessionReducer-test.js`. And a longer example, covering more complexity than we need in this file, is `src/unread/__tests__/unreadModel-test.js`.

src/boot/store.js
```diff
@@ -276,6 +276,11 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     };
   },
 
+  '27': state => ({
+    ...state,
+    muted_users: Immutable.Map({}),
+  }),
```
gnprice, line 282: As you observed in the chat thread, this isn't needed when adding a new key.

gnprice (commented):
Thanks @WesleyAC ! Will be great to have this feature in.

Comments below. I'm also going to go expand our [testing style guide](https://github.com/zulip/zulip-mobile/blob/33ae07cbd50f424bdc6a6daa4f0e62b83f359512/docs/howto/testing.md#unit-tests-js) with some of this information, and some other parts will feed into the "how to add new data from the server" doc that https://github.com/zulip/zulip-mobile/issues/4655#issuecomment-820931874 was a draft of.

src/config.js
```diff
@@ -27,6 +27,7 @@ const config: Config = {
     'alert_words',
     'message',
     'muted_topics',
+    'muted_users',
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/api/initialDataTypes.js
```diff
@@ -34,6 +34,15 @@ export type InitialDataMutedTopics = {|
   muted_topics: MuteTuple[],
 |};
 
+export type MutedUser = {|
+  id: number,
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/api/initialDataTypes.js
```diff
@@ -34,6 +34,15 @@ export type InitialDataMutedTopics = {|
...
+  timestamp: number,
+|};
+
+export type InitialDataMutedUsers = {|
+  muted_users: MutedUser[],
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/api/initialDataTypes.js
```diff
@@ -34,6 +34,15 @@ export type InitialDataMutedTopics = {|
   muted_topics: MuteTuple[],
 |};
 
+export type MutedUser = {|
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/actionTypes.js
```diff
@@ -428,6 +435,12 @@ type EventMutedTopicsAction = {|
   muted_topics: MuteState,
 |};
 
+type EventMutedUsersAction = {|
+  ...ServerEvent,
+  type: typeof EVENT_MUTED_USERS,
+  muted_users: MutedUser[],
+|};
+
```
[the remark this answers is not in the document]
WesleyAC: Done, with the `UserStatusEvent` style.

src/api/eventTypes.js
```diff
@@ -144,3 +145,8 @@ export type UpdateMessageFlagsEvent = {|
   all: boolean,
   messages: number[],
 |};
+
+export type MutedUser = {|
+  id: number,
+  timestamp: number,
+|};
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/reduxTypes.js
```diff
@@ -166,6 +166,9 @@ export type MigrationsState = {|
 
 export type MuteState = MuteTuple[];
 
+/** A map from user IDs to the timestamp when they were muted. */
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/mute/mutedUsersReducer.js
```diff
@@ -0,0 +1,35 @@
...
+  } else {
+    return initialState;
+  }
+}
+
+export default (state: MutedUsersState = initialState, action: Action): MutedUsersState => {
+  switch (action.type) {
+    case LOGOUT:
+    case ACCOUNT_SWITCH:
+      return initialState;
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/mute/mutedUsersReducer.js
```diff
@@ -0,0 +1,35 @@
+/* @flow strict-local */
+import Immutable from 'immutable';
+import type { MutedUsersState, Action, UserId } from '../types';
+import { REALM_INIT, LOGOUT, ACCOUNT_SWITCH, EVENT_MUTED_USERS } from '../actionConstants';
+import { makeUserId } from '../api/idTypes';
+import type { MutedUser } from '../api/apiTypes';
+
+const initialState: MutedUsersState = Immutable.Map({});
```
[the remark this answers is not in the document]
WesleyAC: Done, here and in the tests.

src/mute/mutedUsersReducer.js
```diff
@@ -0,0 +1,35 @@
...
+
+function mutedUsersToMap(muted_users: MutedUser[]): Immutable.Map<UserId, number> {
+  if (muted_users) {
+    return Immutable.Map(
+      muted_users.map(muted_user => [makeUserId(muted_user.id), muted_user.timestamp]),
+    );
+  } else {
+    return initialState;
+  }
```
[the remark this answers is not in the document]
WesleyAC: This was guarding against it being undefined for the javascript `.map`, but I've switched to the cleaner way you suggested in a later comment. Done.

src/mute/mutedUsersReducer.js
```diff
@@ -0,0 +1,35 @@
...
+import { REALM_INIT, LOGOUT, ACCOUNT_SWITCH, EVENT_MUTED_USERS } from '../actionConstants';
+import { makeUserId } from '../api/idTypes';
+import type { MutedUser } from '../api/apiTypes';
+
+const initialState: MutedUsersState = Immutable.Map({});
+
+function mutedUsersToMap(muted_users: MutedUser[]): Immutable.Map<UserId, number> {
+  if (muted_users) {
+    return Immutable.Map(
+      muted_users.map(muted_user => [makeUserId(muted_user.id), muted_user.timestamp]),
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/mute/mutedUsersReducer.js
```diff
@@ -0,0 +1,35 @@
+/* @flow strict-local */
+import Immutable from 'immutable';
+import type { MutedUsersState, Action, UserId } from '../types';
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/mute/__tests__/mutedUsersReducer-test.js
```diff
@@ -0,0 +1,84 @@
...
+
+describe('mutedUsersReducer', () => {
+  describe('REALM_INIT', () => {
+    test('when `muted_users` data is provided init state with it', () => {
+      const initialState = Immutable.Map({});
+      const action = deepFreeze({
+        type: REALM_INIT,
+        data: {
+          muted_users: [
+            {
```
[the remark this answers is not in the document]
WesleyAC: Done.

src/mute/__tests__/mutedUsersReducer-test.js
WesleyAC: Done.

src/boot/store.js
WesleyAC: Removed.

WesleyAC:
Ok, fixed all of those comments, and also went ahead and pushed the initial version of hiding messages from muted users in the MessageList. Should be ready for another look :)

WesleyAC:
Ah, a few of my fixes here are in the wrong commit. I'll fix that in a moment.

WesleyAC:
Ok, fixed up the commit history.

Also, here are screenshots of what the UI looks like.

src/reduxTypes.js
```diff
@@ -166,6 +166,9 @@ export type MigrationsState = {|
 
 export type MuteState = MuteTuple[];
 
+/** A map from user IDs to the Unix timestamp in seconds when they were muted. */
+export type MutedUsersState = Immutable.Map<UserId, number>;
```
chrisbobbe, line 170: I'm glad to have another case of `Immutable.Map`s with numeric keys! 🎉 b54d68d9c should ensure that these get "revived" from storage with their keys as numbers, instead of strings. The data is persisted as JSON so we've had to be careful about that.

src/api/eventTypes.js
```diff
@@ -144,3 +145,9 @@ export type UpdateMessageFlagsEvent = {|
   all: boolean,
   messages: number[],
 |};
+
+export type MutedUsersEvent = {|
```
gnprice, line 149: nit: we've had the event types in alphabetical order in this file (same as in the `EventTypes` enum/class), so let's maintain that

src/mute/mutedUsersReducer.js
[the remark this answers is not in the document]
gnprice: Ah, that makes sense. In fact I think I was forgetting that an empty JS array is truthy (forgetting because an empty Python list is falsy).

In a situation like this but where `?? []` wasn't a handy solution so that we did need some form of conditional like this, I'd probably do it with the early-return pattern:
```js
  if (!muted_users) {
    return initialState;
  }
  return Immutable.Map(muted_users.map(/* … */));
```

Oh, also in that case I'd definitely want the function to reflect in its signature that the thing could be undefined :slightly_smiling_face: . So `muted_users: MutedUser[] | void`, or perhaps `muted_users?: MutedUser[]`. That way the type-checker sees what I'm doing with the check, and future people touching the code won't accidentally refactor away the check without noticing.

src/webview/static/base.css
```diff
@@ -569,7 +581,8 @@ hr {
   border-top: 1px solid hsla(0, 0%, 50%, 0.5);
 }
 
-/* Our own "sorry, unsupported" message for widgets. */
+/* Our own "sorry, unsupported" message for widgets.
+   Also used for showing that a message was sent by a muted user. */
 .widget {
```
gnprice: I was going to comment that it's a fragile solution to reuse something called "widget" for a purpose unrelated to widgets.

The next commit renames this to `special-message`, which seems like a fine solution. That commit should just come before this one so there isn't a concerning intermediate state :slightly_smiling_face:

src/webview/static/base.css
```diff
@@ -142,6 +142,18 @@ body {
   padding: 0 16px 16px 80px;
 }
 
+.message-muted {
+  /* If a message is both brief and muted, we don't want to save the extra
+   * space for the avatar, since it's not shown anyways. */
+  padding: 16px;
```
gnprice: I don't think I follow what this comment is saying about "both brief and muted". Perhaps that'll end up being moot in a final version of this branch? I see this goes away in a later commit in this revision of the branch.

src/webview/html/messageAsHtml.js
```diff
@@ -85,15 +85,28 @@ export const flagsStateToStringList = (flags: FlagsState, id: number): string[]
...
+     class="message ${isBrief ? 'message-brief' : 'message-full'} ${
+    isUserMuted ? 'message-muted' : ''
+  }"
      id="msg-${id}"
      data-msg-id="${id}"
      $!${flagStrings.map(flag => template`data-${flag}="true" `).join('')}
     >`;
   const messageTime = shortTime(new Date(timestamp * 1000), backgroundData.twentyFourHourTime);
 
+  // TODO: i18n
```
gnprice, line 100: Is your intention to fix this TODO before merge? If so, always good to mention that in the commit's summary line and/or the PR thread.

src/webview/html/messageAsHtml.js
```diff
@@ -89,20 +89,19 @@ export default (backgroundData: BackgroundData, message: Message | Outbox, isBri
...
      $!${flagStrings.map(flag => template`data-${flag}="true" `).join('')}
     >`;
   const messageTime = shortTime(new Date(timestamp * 1000), backgroundData.twentyFourHourTime);
 
   // TODO: i18n
   const mutedMessageHtml = isUserMuted
     ? template`
-    <div class="special-message">
-      This message was hidden because it is from a user you have muted.
+    <div class="special-message muted-message-explaination">
```
gnprice, line 104: nit: spelling

src/webview/js/js.js
```diff
@@ -926,9 +926,25 @@ const handleLongPress = (target: Element) => {
     return;
   }
 
+  const targetType = target.matches('.header')
+    ? 'header'
+    : target.matches('a')
+    ? 'link'
+    : 'message';
```
gnprice, line 952: Prettier has some kind of bug affecting nested `?:` operators. Use `prettier-ignore`, and give it a sensible indentation instead :slightly_smiling_face: 

Search the codebase for "ternary" to find a couple of other spots as examples.

src/webview/js/js.js
```diff
@@ -908,6 +913,20 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
   }
 });
 
+const revealMutedMessages = (target: Element) => {
```
gnprice, line 916: This should get a bit of jsdoc to describe what its behavior is supposed to be. It looks like it's fairly subtle.

gnprice, line 916: Separately, this should be somewhere up above the `click` listener. It's helpful that the `click` listener and this long-press handler are together pretty compact, with just a few lines for each of the different actions the user might take -- that's helpful in scanning through to see all the logic for dispatching between them.

src/webview/js/js.js
```diff
@@ -908,6 +913,20 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
...
+  let first_run = true;
+  while (
+    messageNode
+    && messageNode instanceof Element
+    && (messageNode.classList.contains('message-brief') || first_run)
+  ) {
+    messageNode.setAttribute('data-mute-state', 'shown');
+    messageNode = nextMessage(messageNode);
+    first_run = false;
+  }
```
gnprice, line 927: The logic controlling this loop is somewhat messy. For example `nextMessage` should always be returning an `Element` which is a message, or else `null` -- so the `instanceof Element` check and the `classList` check shouldn't be needed. (Unless there's an intended contrast between `message-brief` and `message-full`? If so, that calls for an explanatory comment.)

src/webview/js/js.js
```diff
@@ -908,6 +913,20 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
   }
 });
 
+const revealMutedMessages = (target: Element) => {
+  let messageNode = getMessageNode(target);
```
gnprice, line 917: As the jsdoc says, this function is deprecated :slightly_smiling_face: It looks like `target.closest('.message-brief')` will give you what you're looking for.

src/webview/static/base.css
```diff
@@ -142,6 +142,18 @@ body {
   padding: 0 16px 16px 80px;
 }
 
+.message-muted {
```
gnprice: I wonder if `message-muted-user` would be a clearer name for this class. I keep thinking `message-muted` sounds like it could apply somehow to a topic mute, which is a pretty different-feeling use case that gets treated differently in the UI.

gnprice (commented):
Thanks for the revisions! Comments below, mainly on the later commits.

src/api/eventTypes.js
WesleyAC: This file isn't quite in alphabetical order (since `Submessage` wants to be next to `Message`, it seems?), but I made it a bit closer :)

src/mute/__tests__/mutedUsersReducer-test.js
```diff
@@ -0,0 +1,75 @@
...
+      const initialState = Immutable.Map();
+      const action = deepFreeze({
+        ...eg.eventMutedUsersActionBase,
+        muted_users: [
+          {
+            id: makeUserId(42),
+            timestamp: 1618822632,
+          },
+        ],
+      });
```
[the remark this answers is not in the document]
WesleyAC: Updated the test to match the description.

src/mute/__tests__/mutedUsersReducer-test.js
```diff
@@ -0,0 +1,75 @@
...
+
+      const expectedState = Immutable.Map([
+        [makeUserId(42), 1618822632],
+        [makeUserId(1234), 1618822635],
+      ]);
+
+      const newState = mutedUsersReducer(initialState, action);
+
+      expect(newState).toEqual(expectedState);
+    });
```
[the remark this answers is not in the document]
WesleyAC: I changed it to look more like `src/session/__tests__/sessionReducer-test.js`, but that didn't make it much less verbose — this doesn't benefit too much from `baseReduxData`, since the state isn't a struct.

Let me know if there's a specific way that I could make this less verbose, though.

src/webview/__tests__/generateInboundEvents-test.js
```diff
@@ -15,6 +16,7 @@ describe('generateInboundEvents', () => {
     debug: eg.baseReduxState.session.debug,
     flags: eg.baseReduxState.flags,
     mute: [],
+    mutedUsers: Immutable.Map({}),
```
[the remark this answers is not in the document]
WesleyAC: Fixed.

src/webview/static/base.css
WesleyAC: Fixed.

src/webview/static/base.css
WesleyAC: Yeah, this goes away later.

A "brief" message is one that doesn't show the avatar or username (not a great name, but the one we use). This is so that we don't have the message about the fact that it's muted have a big margin on the left side, the way there normally is when there's no user avatar.

This goes away because I coalesce messages later, so brief messages never have any content in them.

src/webview/html/messageAsHtml.js
WesleyAC: It's my intention to fix this before we launch this feature, but I think we could merge an initial version of this before that.

Currently this is the same as the status quo as for widgets, which maybe isn't a good excuse, but does make more more comfortable merging it before this is done. I'm happy to block on this if you'd prefer, though.

src/webview/html/messageAsHtml.js
WesleyAC: Fixed.

src/webview/js/js.js
WesleyAC: Fixed.

src/webview/js/js.js
WesleyAC: Done. Let me know if this needs any more elaboration.

WesleyAC: Done. I couldn't find a great place to put this, but I moved it up a bit.

src/webview/js/js.js
WesleyAC: Rewrote this as `do { } while ()`, which is much nicer.

The reason for the `instanceof Element` was because of `getMessageNode`, which returns `Node` rather than `Element`, which I've now fixed.

Not sure what you mean about `message-brief` vs `message-full`, the `classList` check is definitely needed. Hopefully the jsdoc makes it clear what's going on?

src/webview/js/js.js
WesleyAC: Fixed, and added a note to use that instead in the jsdoc deprecation comment :)

src/webview/static/base.css
WesleyAC: This goes away in a later commit, so I'm not too inclined to change it.

WesleyAC:
@gnprice should be ready for another review :)

src/webview/html/messageAsHtml.js
WesleyAC: Added i18n in one of the more recent commits :)

src/api/eventTypes.js
gnprice: Hmm true! And `UserStatusEvent` in what seems like a completely arbitrary place. Well, thanks for making it closer :slightly_smiling_face:

src/webview/js/js.js
gnprice: Good idea, thanks :)

src/webview/js/js.js
gnprice: Definitely helpful, thanks!

I'll pick up this thread with a comment on the new version.

gnprice (commented):
Thanks @WesleyAC for the revision! Comments below.

Because this PR thread is getting unwieldy, I haven't looked at the last few new commits:
f78745bde PmConversationList [nfc]: Convert to functional component.
e9298d3d3 PmConversationList: Hide PMs from muted users.
0267617b1 contentHtmlFromPieceDescriptors: Switch to object for args.
13e530e46 messageAsHtml: Translate muted-user text.

I'm also going to go ahead and try merging some of these commits that are ready, in order to simplify things for the rest.

gnprice:
OK, so of these commits:

2d77cfeeb initialDataTypes: Add MutedUsers.
01b763872 Redux: Add muted_users reducer.
2298150bc directSelectors: add getMutedUsers.
82d92ab4d backgroundData: Add muted_users.
ff1773692 css: Rename widget class to special-message.
d445419eb webview: Add initial support for hiding muted messages.
5cc0ee049 webview: Allow revealing messages from muted users.
dff038514 caughtUpReducer: Fix bug that overwrote CaughtUp state.
aad5b966c webview: Coalesce consecutive muted messages.
2c2dfbcf5 webview: Add note about alternative to getMessageNode.
f78745bde PmConversationList [nfc]: Convert to functional component.
e9298d3d3 PmConversationList: Hide PMs from muted users.
0267617b1 contentHtmlFromPieceDescriptors: Switch to object for args.
13e530e46 messageAsHtml: Translate muted-user text.

I'm merging versions of these:

2d77cfeeb initialDataTypes: Add MutedUsers.
01b763872 Redux: Add muted_users reducer.
2298150bc directSelectors: add getMutedUsers.
82d92ab4d backgroundData: Add muted_users.
ff1773692 css: Rename widget class to special-message.
2c2dfbcf5 webview: Add note about alternative to getMessageNode.
0267617b1 contentHtmlFromPieceDescriptors: Switch to object for args.

The merged commits are:

10a07b9a9 api: Add types for muted users; add muted_users to requested data.
14ee13e1a muted users: Add data to Redux state.
c438b54b9 muted users tests [nfc]: Compact object literals to 1 line, vs. 4.
efb301b56 muted users tests [nfc]: Inline small actual/expected expressions.
b67e440c3 muted users tests [nfc]: Factor out boring arbitrary base state.
c8ed0b08f muted users tests: Use eg.otherUser and eg.thirdUser.
c7c58c894 example data [nfc]: Describe the "action fragments" idea more clearly.
0a66a1af0 example data [nfc]: Describe when to define example actions here.
f0c01d079 msglist: Add muted_users to backgroundData.
aeb56ec23 msglist css [nfc]: Rename widget class to special-message.
954a67fb5 webview js [nfc]: Add note about alternative to getMessageNode.
713f1c6f7 msglist html [nfc]: Use an args object for contentHtmlFromPieceDescriptors.

Do take a look. The main changes are:
* I squashed the `directSelectors` commit into the Redux commit, and moved a few of the changes from the Redux commit ahead into the API commit -- specifically the parts that are basically more of the API binding, so that all the API-binding changes are in one commit.
  * That means the API commit now includes the changes to `src/api/eventTypes.js`, and also the corresponding bits to turn that event into a Redux action, because those bits are (thankfully) just boilerplate.
* I added `[nfc]` to summary lines where appropriate.
  * Hmm, I guess that isn't in our style guide! It is documented in our [glossary](https://github.com/zulip/zulip-mobile/blob/master/docs/glossary.md#nfc). I should go add it to the style guide (which I think we hadn't started yet when I wrote that entry.)
* I revised some summary lines mostly to make the prefixes broader, like changing "contentHtmlFromPieceDescriptors" to "msglist html", and putting the specific function name in the later part of the summary line.
  * This looks like it may not really be explained anywhere in any detail; it probably should be! But as the [Zulip-wide Git style guide](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages) says briefly, that prefix is generally "the name of the subsystem". So "the message list" or "the HTML generation for the message list" is a good match for the "subsystem" level of generality we're going for there. Sometimes I'll do a quick `git log src/foo/` to see how we've referred to a given subsystem in past commits.
* I wrote some additional commits to tighten up the new test file along the lines I had in mind up at https://github.com/zulip/zulip-mobile/pull/4678#discussion_r616363607 . These are split out into commits more finely than I probably would otherwise do, to help make visible the different things I'm doing there. In net the file gets 22 lines shorter, and the two very simple tests become one line each.
  * I also added there a couple of changes to `exampleData.js` that hopefully make clear some things that really weren't written down there before -- so thank you for running into those :slightly_smiling_face:

gnprice:
And merged those commits -- thanks again @WesleyAC!

Because this thread has gotten unwieldy for GitHub (with parts of it hidden to keep their page-load numbers good, etc.), let's pick up the rest on a new PR. I've pushed a rebased version to my repo as a branch [dev-muted-users](https://github.com/zulip/zulip-mobile/compare/master...gnprice:dev-muted-users) -- the commits are:
9cb90e820 caughtUp: Fix bug that overwrote CaughtUp state.
84aa24f32 webview: Add initial support for hiding muted messages.
c83092f8b webview: Allow revealing messages from muted users.
b738acea0 webview: Coalesce consecutive muted messages.
1391b41ca PmConversationList [nfc]: Convert to functional component.
16fbb4735 PmConversationList: Hide PMs from muted users.
89b8f95cc messageAsHtml: Translate muted-user text.

---

One thing I noticed on the `caughtUp` commit, just before I was going to merge with it included:
```diff
-        caughtUp = { older: action.foundOldest, newer: action.foundNewest };
+        caughtUp = {
+          // $FlowIssue[unnecessary-optional-chain]
+          older: state[key]?.older || action.foundOldest,
+          // $FlowIssue[unnecessary-optional-chain]
+          newer: state[key]?.newer || action.foundNewest,
+        };
```
What happens if `state[key]` is indeed absent?

I believe we'll end up with the right result -- `false` or `true` for each value, depending on the data in `action` -- but it's a bit subtle. In particular if we had the two sides of the `||` in the other order, which looks pretty innocuous, I believe it'd go wrong and stick an `undefined` in there when `foundFooest` was false.

So it'd be good to make the fallback logic here more explicit. I think the logic at the end of `legacyInferCaughtUp` above is pretty effective for this:
```js
  const { older: prevOlder, newer: prevNewer } = prevCaughtUp || DEFAULT_CAUGHTUP;

  return {
    older: prevOlder || caughtUpOlder,
    newer: prevNewer || caughtUpNewer,
  };
```

Where the jsdoc on `DEFAULT_CAUGHTUP` is:
```js
/** The value implicitly represented by a missing entry in CaughtUpState. */
```

PR #4343, opened 2020-12-16
Open links in different ways (browser and webview)

shaurya2612:
This PR is in continuation of #4336 and resolves issue #4323 

I've added a seperate function called openLinkEmbedded for opening links in webview, as stated in -https://github.com/zulip/zulip-mobile/pull/4336#issuecomment-745109552 - I have made changes to open a webview in `AuthScreen.js` and `webAuth.js`.

PR #4679, opened 2021-04-19
Open links in external browser

WesleyAC:
Fixes: #4323

Based on #4343

### DIFF
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -283,6 +283,14 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     accounts: state.accounts.filter(a => a.email !== ''),
   }),
 
+  '28': state => ({
+    ...state,
+    settings: {
+      ...state.settings,
+      browser: 'default',
+    },
+  }),
+
   // TIP: When adding a migration, consider just using `dropCache`.
 };
 
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -51,6 +51,7 @@ const makeIcon = <Glyphs: string>(
   return <FixedIcon name={name} {...props} />;
 };
 
+export const IconBrowser = makeIcon(Feather, 'chrome');
 export const IconInbox = makeIcon(Feather, 'inbox');
 export const IconMention = makeIcon(Feather, 'at-sign');
 export const IconSearch = makeIcon(Feather, 'search');
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -263,6 +263,21 @@ export type RealmState = {|
 // https://github.com/zulip/zulip-mobile/issues/4009#issuecomment-619280681.
 export type ThemeName = 'default' | 'night';
 
+/**
+ * The values for this mean:
+ *
+ * * embedded: The in-app browser
+ * * external: The user's default browser app
+ * * default: 'external' on iOS, 'embedded' on Android
+ *
+ * Use the `shouldUseInAppBrowser` function from src/utils/openLink.js in order to
+ * parse this.
+ *
+ * See https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser
+ * for the reasoning behind these options.
+ */
+export type BrowserPreference = 'embedded' | 'external' | 'default';
+
 export type SettingsState = {|
   locale: string,
   theme: ThemeName,
@@ -270,6 +285,7 @@ export type SettingsState = {|
   onlineNotification: boolean,
   experimentalFeaturesEnabled: boolean,
   streamNotification: boolean,
+  browser: BrowserPreference,
 |};
 
 export type StreamsState = Stream[];
--- a/src/settings/SettingsScreen.js
+++ b/src/settings/SettingsScreen.js
@@ -6,12 +6,13 @@ import { ScrollView } from 'react-native';
 import type { RouteProp } from '../react-navigation';
 import type { MainTabsNavigationProp } from '../main/MainTabsScreen';
 import * as NavigationService from '../nav/NavigationService';
-import type { Dispatch } from '../types';
+import type { Dispatch, BrowserPreference } from '../types';
 import { createStyleSheet } from '../styles';
 import { connect } from '../react-redux';
 import { getSettings } from '../selectors';
 import { OptionButton, OptionRow } from '../common';
 import {
+  IconBrowser,
   IconDiagnostics,
   IconNotifications,
   IconNight,
@@ -25,6 +26,7 @@ import {
   navigateToDiagnostics,
   navigateToLegal,
 } from '../actions';
+import { shouldUseInAppBrowser } from '../utils/openLink';
 
 const styles = createStyleSheet({
   optionWrapper: {
@@ -37,6 +39,7 @@ type Props = $ReadOnly<{|
   route: RouteProp<'settings', void>,
 
   theme: string,
+  browser: BrowserPreference,
   dispatch: Dispatch,
 |}>;
 
@@ -47,7 +50,7 @@ class SettingsScreen extends PureComponent<Props> {
   };
 
   render() {
-    const { theme } = this.props;
+    const { dispatch, theme, browser } = this.props;
 
     return (
       <ScrollView style={styles.optionWrapper}>
@@ -56,6 +59,14 @@ class SettingsScreen extends PureComponent<Props> {
           label="Night mode"
           value={theme === 'night'}
           onValueChange={this.handleThemeChange}
+        />
+        <OptionRow
+          Icon={IconBrowser}
+          label="Open links with in-app browser"
+          value={shouldUseInAppBrowser(browser)}
+          onValueChange={value => {
+            dispatch(settingsChange({ browser: value ? 'embedded' : 'external' }));
+          }}
         />
         <OptionButton
           Icon={IconNotifications}
@@ -92,4 +103,5 @@ class SettingsScreen extends PureComponent<Props> {
 
 export default connect(state => ({
   theme: getSettings(state).theme,
+  browser: getSettings(state).browser,
 }))(SettingsScreen);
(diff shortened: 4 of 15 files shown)


chrisbobbe:
Thanks, @WesleyAC, this LGTM! I tested on iOS and it works as expected.

I added one commit on top, to bump the new `openLinkExternal` toward more visibility/adoption. If that looks good, please merge at will.

WesleyAC:
@chrisbobbe Thanks! I added a setting as per the [CZO discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser), would be good to get a review on the last commit in this stack :)

WesleyAC:
Ok, pushed a new version with defaults as discussed on CZO :)

src/settings/settingsReducer.js
```diff
@@ -14,6 +14,7 @@ const initialState: SettingsState = {
   onlineNotification: true,
   experimentalFeaturesEnabled: false,
   streamNotification: false,
+  browser: 'default',
```
chrisbobbe, line 17: Let's add a migration for adding this to the settings state. I was a bit surprised to see things working, in my local testing, without one.

It turns out that the `REHYDRATE` action doesn't clobber the `initialState` with the state that's just been grabbed from storage. Rather, it does a "shallow merge": https://github.com/zulip/zulip-mobile/blob/v27.162/src/third/redux-persist/autoRehydrate.js#L69. That's why there's still a `state.settings.browser`, post-rehydration.

It took me a while staring at the code to find that fact, and I'm not aware that we intentionally depend on it anywhere. Also, `redux-persist` and `redux-persist-migrate` sort of work together, so this logic could plausibly change as we overhaul `redux-persist`.

We've also found that it's good to err on the side of being explicit and literal in what goes in migrations; e.g., in 1193a57cc346c9f2d684bcd6fee5bb0fa2d0ed11.

b058fa266881cf84570d6c9390c03eaabed15119 has an example of adding a migration for more data being stored. That one was definitely necessary, I think; the "shallow merge" couldn't have helped there.

chrisbobbe (commented):
Thanks, @WesleyAC! I'm not sure I've mentioned this yet but I'm quite looking forward to this. I often uninstall and reinstall the app on my iPhone, since debug and release builds share an identifier (you can't have one of each installed, like on Android). That means the embedded browser doesn't have a long memory for me being logged into things; I want to start using my normal cookies, as Greg mentions in his reasoning on [CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118). A few comments below.

src/reduxTypes.js
```diff
@@ -263,13 +263,26 @@ export type RealmState = {|
 // https://github.com/zulip/zulip-mobile/issues/4009#issuecomment-619280681.
 export type ThemeName = 'default' | 'night';
 
+// The values for this mean:
+// * embedded: The in-app browser
+// * external: The user's default browser app
+// * default: 'external' on iOS, 'embedded' on Android
+//
```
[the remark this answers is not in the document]
WesleyAC: Done!

src/settings/SettingsScreen.js
```diff
@@ -57,6 +60,14 @@ class SettingsScreen extends PureComponent<Props> {
           value={theme === 'night'}
           onValueChange={this.handleThemeChange}
         />
+        <OptionRow
+          Icon={IconBrowser}
+          label="Open links with in-app browser"
```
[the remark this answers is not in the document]
WesleyAC: Ah, good catch! A PR to error on untranslated strings in dev sounds great.

src/utils/openLink.js
```diff
@@ -13,3 +16,21 @@ export function openLinkEmbedded(url: string): void {
...
+
+export function shouldUseInAppBrowser(browser: BrowserPreference): boolean {
+  if (browser === 'default') {
+    return Platform.OS === 'android';
+  } else {
+    return browser === 'embedded';
+  }
+}
+
+export function openLinkWithUserPreference(url: string, getState: GetState): void {
```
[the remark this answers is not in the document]
WesleyAC: (waiting on @gnprice to weigh in here)

src/settings/settingsReducer.js
WesleyAC: Done!

WesleyAC:
Should be ready for another look @chrisbobbe :)

src/reduxTypes.js
```diff
@@ -263,13 +263,29 @@ export type RealmState = {|
...
+ *
+/* * embedded: The in-app browser
+/* * external: The user's default browser app
+/* * default: 'external' on iOS, 'embedded' on Android
+/*
+/* Use the `shouldUseInAppBrowser` function from src/utils/openLink.js in order to
+/* parse this.
+/*
+/* See https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser
+/* for the reasoning behind these options.
```
chrisbobbe, line 277: nit: there are extra `/`'s on the left.

chrisbobbe:
LGTM, pending that spot where we're waiting for @gnprice to weigh in, [here](https://github.com/zulip/zulip-mobile/pull/4679#discussion_r627138670). As I mentioned, I'd be happy enough with a small comment and no further action there, so I'd imagine we could time out waiting for Greg in a day or two. 🙂

src/reduxTypes.js
WesleyAC: Fixed :)

WesleyAC:
Added that comment, going ahead and merging. Happy to circle back later if we decide we want a thunk action creator instead.

src/common/Icons.js
```diff
@@ -51,6 +51,7 @@ const makeIcon = <Glyphs: string>(
   return <FixedIcon name={name} {...props} />;
 };
 
+export const IconBrowser = makeIcon(Feather, 'chrome');
```
gnprice, line 54: Another thing that I'd like to improve is the icon - I'm using the
    Feather icon for Google Chrome, since the only other browser-related
    icons I could find were the globe icon we're using for the language
    setting in this same screen, and the compass icon, which is less
    recognizable than the Chrome one. I'm unhappy with using a branded icon
    here, but Feather doesn't give us very many options, sadly.

Hmm. Yeah, this icon isn't very appropriate:
* It looks like it's suggesting the setting means something like "use Chrome". But actually its effect is independent of whether you use Chrome or some other browser, or at least it is on Android -- with either value of this setting, you get the same choice of browsers when you open a link.
* For most iOS users, with Safari as their browser, it doesn't much match what the setting does at all.

It might be best to just leave out the icon. In general, settings UIs often have icons for high-level groups of settings, but usually don't for individual settings.

src/reduxTypes.js
```diff
@@ -263,13 +263,29 @@ export type RealmState = {|
...
+ * * external: The user's default browser app
+ * * default: 'external' on iOS, 'embedded' on Android
+ *
+ * Use the `shouldUseInAppBrowser` function from src/utils/openLink.js in order to
+ * parse this.
+ *
+ * See https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser
+ * for the reasoning behind these options.
+ */
+export type BrowserPreference = 'embedded' | 'external' | 'default';
```
gnprice, line 279: Thanks for this detailed jsdoc!

Like all our jsdoc, it should also get a one-line summary at the top. It seems I haven't written that down in our style guide, hmm. But when I do, I don't expect to improve on this item and the one after it (just some details are different because it's a different language):
https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary
In particular:
> Provide just enough context for the reader to orient themselves and decide if they should keep reading or look elsewhere for the solution to their problem.

Here, one way of looking at it would be: this nicely provides the answer corresponding to each possible value. But it doesn't really say what question the answers are answering. That'd be a good subject for the summary line.

So one version would be:
```
/**
 * The user's preference on what browser to open links with.
 *
 * The values for this mean:
 *  * embedded: The in-app browser
 *  * external: The user's default browser app
 *  * default: 'external' on iOS, 'embedded' on Android
 *
…
```
(That also demonstrates a slight reformatting of the list that I think makes it a bit easier to visually parse.)

src/utils/openLink.js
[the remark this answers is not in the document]
gnprice: Hmm. I agree that this interface taking `getState` feels odd. I feel like the direction I'd rather go in is to separate this function from being concerned with our Redux store at all -- that feels to me like a layer that lives above this one.

Probably the simplest resolution is to have this just take a `BrowserPreference`. The caller is already looking around at various other bits of the state, and it's no trouble for it to to pick out that one too and pass it here.

OTOH if we had more call sites, it would start to feel weird that part of the interface of this function was effectively that you had to say something like `getSettings(state).browser` in its second argument. The exact bit of settings that this depends on feels like it should be an internal implementation detail.

Maybe the best solution is a point in between: have it take a `SettingsState`. Then:
* The exact setting it needs is an internal detail, just like in the current version.
* But it's no longer exposed to the whole Redux state.
* Also the fact that it does depend on *some* user setting becomes explicit in the interface. I think it's good for that to be explicit -- I wouldn't e.g. want this to take just one argument `url: string`, and go find the Redux store as a global or something, or if it did then I'd at least want that connection to be drawn clearly in its jsdoc.

src/boot/store.js
```diff
@@ -283,6 +283,14 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     accounts: state.accounts.filter(a => a.email !== ''),
   }),
 
+  '28': state => ({
```
gnprice, line 286: Should get a quick explanation, like the other migrations above. (Chris already took care of this in 74ae528f0.)

gnprice (commented):

Just took a proper look through these changes today. Comments below, including one on that thunk-action-creator thread and one on the icon.

PR #4680, opened 2021-04-20
shared: Bump version to 0.0.5.

WesleyAC:


chrisbobbe:
Is there anything specific that we're looking for in the new version, or is this just the release that was made because one hadn't been made in a while? That might be good to put in the commit message, along with any known highlights (or breaking API changes, hopefully there aren't those) that we get with this version. I guess currently we just use `typing_status` from this, so any news outside of that will be less interesting.

WesleyAC:
Updated the commit message to cover both of those :)

WesleyAC:
I'm going to go ahead and merge this, since I addressed @chrisbobbe's comments and I think it's pretty low risk.

PR #4681, opened 2021-04-20
action sheet: Add "edit message" option for admins to edit any topic.

rajprakash00:
Currently in zulip web any admin can edit any topic name
irrespective of whether admin has any sent messages in that
topic or not.

So, add that feature in zulip mobile.

**Before**: 

https://user-images.githubusercontent.com/42652941/115412997-74587f80-a212-11eb-8329-1102af8cb235.mp4

**After** :

https://user-images.githubusercontent.com/42652941/115413194-9baf4c80-a212-11eb-9e87-e24519678949.mp4

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -300,9 +300,13 @@ export const constructMessageActionButtons = ({
     buttons.push(shareMessage);
   }
   if (
-    message.sender_id === ownUser.user_id
-    // Our "edit message" UI only works in certain kinds of narrows.
-    && (isStreamOrTopicNarrow(narrow) || isPmNarrow(narrow))
+    (message.sender_id === ownUser.user_id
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks @rajprakash00 !

It would indeed be good to add the feature for people to edit the topic of other people's messages, where they have permission to do that.

But it looks like what this PR does is to offer the "edit *message*" UI in some cases for other people's messages. That is, the option is labeled "Edit message", and when you choose it the UI you get includes an input for the message content as well as the message's topic. We don't want to do that -- the server won't actually accept such an edit anyway, but we don't want to make it look like that's what the feature does.

The main work we'll need to do to add this feature is to add some UI for specifically editing just the topic. I'm closing this PR because I think that'll be pretty independent of what's in this PR.

PR #4683, opened 2021-04-20
More prep work for handling safe areas.

chrisbobbe:
I think a good chunk of the work toward making #3066 work without things looking too awkward will be to add appropriate left and right padding to various row elements that are meant to extend to the left and right edges of the screen.

For example, we don't want this (how it is now):

But we _also_ don't want to just throw everything in a `View` with its own margin or padding—rows are cut off abruptly and don't extend to the edge; you can see this in the stream header, but it also affects the touch highlight (not shown):

Instead, we want this—rows extend to the edge, but content is padded so it doesn't get hidden:

(For the left inset there, it looks to me like we're wasting some space—but the padding I added is exactly the length of the inset announced by the device. While there's no notch there, it seems Apple has decided it needs a large inset because of how rounded the corners are, which I guess makes sense.)

This PR does

- a bit of cleanup on some of our row components,
- some other things toward #3066 (like making `UserPickerCard` a function component), and
- some independent cleanups.

Then I've got a draft on top of this that adds padding for (I think) all the relevant "row" components.

I'll post screenshots of how the appearance changes in this PR (it's not a lot) soon.


chrisbobbe:
The icon size changes from these two commits:

79d9b2ea71d48138dd3098906df266a9c4bb6474 NavButton: Use standard icon size of 24, instead of using 26.
d4475f780777a70619dde7e75b27010ac472a899 TitleStream: Use standard icon size of 24, instead of 20, for StreamIcon.

Before this PR (left); after this PR (right):

chrisbobbe:
The change to the user-status-suggestion list from

361b0db40d599ae83133942e0daf77bfb7e6fa79 UserStatusScreen [nfc]: Use new `SelectableOptionRow`.

Before this PR:



After this PR:

src/common/SelectableOptionRow.js
```diff
@@ -0,0 +1,51 @@
+/* @flow strict-local */
```
WesleyAC, line 1: It would be good to include in the commit message here what you intend to use this component for, it's not clear to me from the name.

src/common/SelectableOptionRow.js
```diff
@@ -33,8 +33,8 @@ type Props<TItemKey: string | number> = $ReadOnly<{|
   // position that requires it).
   itemKey: TItemKey,
 
-  subtitle: string,
```
WesleyAC: Could this commit message be `SelectableOptionRow [nfc]: Reorder variables.` instead of `SelectableOptionRow [nfc]: Tiny refactor.`? That makes it more clear what's going on.

src/common/NavButtonRow.js
```diff
@@ -12,10 +12,18 @@ import styles, { ThemeContext } from '../styles';
...
 |}>;
 
-export default class OptionButton extends PureComponent<Props> {
+/**
+ * A button that navigates to a "nested" screen.
+ *
+ * Shows a right-facing arrow to indicate its purpose. If you need a
+ * selectable option row instead, use `SelectableOptionRow`.
+ */
+export default class NavButtonRow extends PureComponent<Props> {
```
WesleyAC, line 26: IMO `NavButtonRow` is a confusing name for this, maybe `NestedNavRow` or something similar? `NavButton` makes me think of a button in the NavBar, which this isn't.

WesleyAC (changes requested):
Looks good overall, just a few comments!

For the commit messages in the `SelectableOptionRow [nfc]: Widen interface to match name` series, could you add more description to each commit? It'd be useful to have a summary of what each one does.

And for the `/n` commits, it'd be good to replace those with the final number, although I presume you're planning to do that already?

src/common/NavButtonRow.js
chrisbobbe: Good point; `NestedNavRow` sounds good.

src/react.js
```diff
@@ -0,0 +1,20 @@
+/* @flow strict-local */
```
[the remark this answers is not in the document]
chrisbobbe: I'd been following the pattern of `src/react-redux.js`, `src/react-native-safe-area-context.js`, and `src/react-navigation.js` (possibly others?). These have so far been for "type wrappers"; things like

```
/**
 * Exactly like the `useSelector` in `react-redux` upstream, but more typed.
 * […]
```

But

1. Maybe that's not the best place for those?
2. This one isn't a type wrapper; it's not modeled on an existing thing, it's a new thing.

So, putting it in `utils` sounds good.

chrisbobbe:
(Just rebased.)

src/title/TitleStream.js
```diff
@@ -91,7 +91,7 @@ const TitleStream = (props: Props) => {
             isMuted={!stream.in_home_view}
             isPrivate={stream.invite_only}
             color={color}
-            size={20}
+            size={24}
```
gnprice, line 94: This actually ends up looking kind of oversized to me, in your handy screenshot at https://github.com/zulip/zulip-mobile/pull/4683#issuecomment-823584357 -- like it's out of proportion with the text that comes after it.

I think perhaps it makes sense here to think of this as not our usual use of an icon. The reasoning in previous adjustments like this (like the commit before this one) is about icon buttons -- but this icon isn't a button, it's not something you can activate. Instead it's sort of a sigil attached to the stream name, and it wants to be in a size that corresponds to that text.

It looks like the existing code ends up using the same size here as the font size on the title. Perhaps just replace the literal 20 here with an explicit reference to that?

gnprice (commented):
Thanks @chrisbobbe and @WesleyAC  ! This looks good -- small comments below, and I'll reply on a thread above.

src/react.js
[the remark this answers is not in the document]
gnprice: I've generally avoided putting new things in `src/utils/`, because I tend to feel that a "utils" directory makes things harder to find more than it makes them easier. It makes the list of things in the parent directory shorter; but now when you want to browse for something, the first thing you have to guess is whether it got called a "utility". Better for the top-level organization to be by what general area or subsystem the code belongs to.

If you take a look in the server/webapp repo, you'll see there are no directories with "util" names there, either. But there is this pattern, which gives me an idea here:
```
$ git ls-files | grep util
static/js/hash_util.js
static/js/keydown_util.ts
static/js/list_util.js
static/js/message_util.js
…
```

So for example there are a lot of files related to messages; and one of them, for some miscellaneous functions with minimal dependencies, is named `message_util.js`. That feels a lot better to me than a `util/messages.js` would, because you can still navigate it by starting with "the thing I'm looking for would be about messages".

Here, how about `src/reactUtils.js`?

chrisbobbe:
Thanks for the review! Conflicts resolved and revision pushed. 🙂

gnprice:
Looks good, thanks! Merging.

chrisbobbe:
Thanks for the review!

Issue #4684, opened 2021-04-20
Efficiently locate read messages in "unreads" state


gnprice:
This is a followup to #4438.

One of the optimizations in #4448, which closed that issue, was that when we learn a message has been read, we'd efficiently look up the relevant bit of the unreads state and delete it there. Previously, we'd been searching through the whole data structure to find each newly-read message.

As implemented, this optimization had a bug: it relied on `state.messages` to tell it the conversation the message was in. And `state.messages` can be far less complete than `state.unread` -- in fact that's the point of the server feature that initializes `state.unread`. So we'd fail to find the message, and `state.unread` would keep recording it as read. @WesleyAC [discovered this](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/syncing.20read.2Funread.20state/near/1166265
) and bisected it to the relevant commit.

For an immediate fix, I'm about to revert that optimization. But fundamentally we have the data we need to do it correctly: if a message is in the unreads state, then we know where in the unreads state it belongs, and we just need that information indexed by message ID. So we should build that data structure, and use it to restore this optimization.

Effectively that data structure will mean there's a small amount of per-message data we have on potentially a large number of messages, while we have a full `Message` object -- notably, the message content -- on typically a much smaller number of messages.

Design discussion [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/syncing.20read.2Funread.20state/near/1166725).

timabbott:
In the webapp, `static/js/unread.js` has a set of data structures for precisely this problem, `Bucketer` and friends.  @showell can help explain them if useful.

PR #4685, opened 2021-04-20
unread: Revert "Use state.messages to optimize updating stream unreads."

gnprice:
See commit messages, #4684, and/or [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/syncing.20read.2Funread.20state/near/1166265).

### DIFF
--- a/src/unread/unreadModel.js
+++ b/src/unread/unreadModel.js
@@ -44,153 +44,53 @@ export const getUnreadMentions = (state: GlobalState): UnreadMentionsState => st
 
 const initialStreamsState: UnreadStreamsState = Immutable.Map();
 
-// Like `Immutable.Map#update`, but prune returned values equal to `zero`.
-function updateAndPrune<K, V>(
+// Like `Immutable.Map#map`, but with the update-only-if-different semantics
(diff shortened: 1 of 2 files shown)


src/unread/unreadModel.js
```diff
@@ -271,7 +156,8 @@ function streamsReducer(
     }
 
     case EVENT_MESSAGE_DELETE:
-      return deleteMessages(state, action.messageIds, globalState.messages);
+      // TODO optimize by using `state.messages` to look up directly
```
chrisbobbe, line 159: IIUC "by using `state.messages`" is no longer accurate (here and below), but we do have #4684 to track the planned solution that doesn't use `state.messages`, so this might be fine as long as we don't lose track of #4684.

chrisbobbe (approved):
Thanks for the report, @WesleyAC, and the fix, @gnprice!

See one small comment below; no need to change anything there if you don't think we need to.

I tested and I don't reproduce the bug here.

src/unread/unreadModel.js
gnprice: Ah indeed, good catch. Updated the comments.

gnprice:
Thanks for the review; merged!

PR #4686, opened 2021-04-21
css: Style headings to be similar to webapp.

WesleyAC:
The webapp styles the `<h1>` through `<h6>` tags to be all the same size,
bold, and underlined. This commit makes them look similar on mobile,
instead of the goofy giant text from the default sytlesheet that we had
previously.

old

old

new

new

webapp

webapp

### DIFF
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -569,6 +569,17 @@ hr {
   border-top: 1px solid hsla(0, 0%, 50%, 0.5);
 }
 
+h1, h2, h3, h4, h5, h6 {
+  font-size: 100%;
+  font-weight: 600;
+  /* line-height doesn't seem to be needed on mobile, but it's here just for
+   * the sake of being consistent with the webapp */
+  line-height: 1.4;
+  margin-top: 0;
+  margin-bottom: 5px;
+  text-decoration: underline;
+}
+
 /* Our own "sorry, unsupported" message for widgets. */
 .special-message {
   display: flex;


src/webview/static/base.css
```diff
@@ -569,6 +569,14 @@ hr {
   border-top: 1px solid hsla(0, 0%, 50%, 0.5);
 }
 
+h1, h2, h3, h4, h5, h6 {
+  font-weight: bold;
+  font-size: 100%;
+  text-decoration: underline;
+  margin-top: 0;
+  margin-bottom: 0.5rem;
+}
```
chrisbobbe, line 578: Since these aren't identical to the web app's corresponding styles—

(from static/styles/rendered_markdown.css)
```css
{
  font-size: 14px;
  font-weight: 600;
  line-height: 1.4;
  margin-top: 0;
  margin-bottom: 5px;
  text-decoration: underline;
}
```

—we could put a brief comment saying what they do. That way, the default interpretation for this group of styles is that the differences aren't there to make a dramatic departure from the web app. Maybe a comment like "All heading levels have the same style, like in the web app."? It looks like https://github.com/zulip/zulip/issues/11418#issuecomment-459573917 is why the web app makes the slightly surprising choice of not differentiating the styles between h1, h2, etc., at all.

I think I don't mind leaving out comments on each individual difference from the web app, as long as we've made a reasonable effort to match it closely, with necessary changes so it works on mobile. Could, e.g., reorder them to be in the same order the web app uses. When someone revisits these particular styles, they'll likely have the corresponding web app code open at the same time, for comparison.

chrisbobbe: Hmm, or maybe the section header is strong enough—

```js
/*
 Message contents.

 For the body of a message, we use nearly the same HTML as the webapp does.
 The styles therefore have a lot in common.  For the webapp's version of
 these styles, compare static/styles/rendered_markdown.css .
 */
```

—and the thing to do is mark any surprising/dramatic departures, and leave the boring differences unmarked.

chrisbobbe (approved):
Thanks, @WesleyAC! I agree that it makes sense to follow the web app here, although I was a bit surprised to see that it doesn't differentiate the styles at all.

Looks good; feel free to merge after seeing a small comment below.

chrisbobbe:
Oh, hmm, one thing I notice from the screenshots. There's this comment in the web app code:

```
/* Headings: We need to make sure our headings are less prominent than our sender names styling. */
```

On mobile, the headings now look pretty similar to the sender names styling we use. Do you think that might be an issue?

WesleyAC:
Hmmm, looks like the difference between usernames and headings in the webapp is a weight of `700` vs `600`.

We haven't used numbers for `font-weight` in the past, but [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight), `bold` == `700`, so using `600` for headings sounds good. I'll make that change.

On whether to comment the differences, my inclination is not to, unless the differences are for an explicit departure from the webapp look. In this case, the departures are all because we don't use the same units as the webapp, which I don't think needs to be commented, people reading the CSS presumably already understand that. I'll go ahead and reorder the lines though, since I don't see a reason not to do that.

WesleyAC:
I ommitted the `line-height`, btw, since it doesn't seem to be needed on Android. Could you take a look on iOS as well? it seems to me that the `margin-bottom` and `font-size` control the height of the entire thing, presumably a difference in the base CSS for mobile browsers and desktop ones?

src/webview/static/base.css
```diff
@@ -569,6 +569,14 @@ hr {
   border-top: 1px solid hsla(0, 0%, 50%, 0.5);
 }
 
+h1, h2, h3, h4, h5, h6 {
+  font-size: 100%;
+  font-weight: 600;
+  margin-top: 0;
+  margin-bottom: 0.5rem;
```
gnprice, line 576: Margin should be in `px`:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#px-rem-no-em

gnprice (commented):
Thanks for this fix! One tweak below.

> I omitted the `line-height`, btw, since it doesn't seem to be needed on Android.

Did you test with a heading that was long enough that it wrapped onto several lines? That's the case where I'd particularly expect that to make a difference.

Even if it doesn't seem to make a difference, though, I'd be inclined to keep it for the sake of alignment with the webapp. I think for these styles, having what looks like a difference from the webapp (but on close study turns out not to be) is a greater complexity cost than having a line of code.

And then I agree that the right approach for comments is to mark differences that are surprising. In particular I think that means
* places where we make a different substantive choice;
* places where the reason that our different code has the same behavior is a subtle one. I think the line-height being redundant would be an example of this :wink:

WesleyAC:
> Did you test with a heading that was long enough that it wrapped onto several lines? That's the case where I'd particularly expect that to make a difference.

Yes, I tested this, it doesn't make a difference. And for what it's worth, I see changes when I remove `line-height` on desktop even with non-wrapped headings.

In any case, I added the line height just for the sake of consistency.

Should be good to go now.

src/webview/static/base.css
WesleyAC: Done.

gnprice:
Looks good, thanks -- merged!

> And then I agree that the right approach for comments is to mark differences that are surprising. In particular I think that means
> 
> * places where we make a different substantive choice;
> * places where the reason that our different code has the same behavior is a subtle one. I think the line-height being redundant would be an example of this :wink:

Just to clarify this last bit: I meant that if we were to leave out the line-height, then we would definitely want to have a comment explaining it. (Because our code would be different there from the webapp's similar code, but for subtle reasons it would turn out to have the same behavior. At least on Chrome/Android, I guess.) Which then reinforces that it's simpler to have that line than to leave it out and end up having to think about the difference.

But I also don't mind having the comment even when we do have the line-height line, as in your revision I just merged. It is a surprising fact that might be helpful for someone making changes there in the future.

PR #4688, opened 2021-04-21
Haptic feedback on long press, Fixes: #4649

### LINKED ISSUE #4649: Haptic feedback on long press
When the user does a long-press, we should respond with haptic feedback. This is standard on both platforms. Here's an explanation of why it's needed:
https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/haptics/
In particular I think our lack of it is contributing to our long-press responses feeling slow, even though we're responding at the same 500ms threshold as is standard on both platforms. Chat discussion:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/long.20press/near/1160638

There's basically four environments for this, some of which we might be able to cover at once and some of which might be independent tasks:
* [x] Inside the WebView, on iOS. This one seems to already work (see chat discussion) on iOS 13+. No action needed, but if we hear of it not working on iOS 13+ then we'll investigate.
* [ ] Inside the WebView, on Android.
* [ ] In the pure-RN views (so e.g. long-pressing a topic in the unreads list or in the app bar), on iOS.
* [ ] In the pure-RN views (so e.g. long-pressing a topic in the unreads list or in the app bar), on Android.

On Android there's a [specific haptic response](https://developer.android.com/reference/android/view/HapticFeedbackConstants#LONG_PRESS) that is standard for long-press, and we should be sure to use that. On iOS I'm not sure there's a single standard, but we should use something similar to other iOS apps, as standard as possible -- which empirically means it'll be pretty different from the Android standard. Ideally we can get both at once by using some library made for RN.

@chrisbobbe [adds some notes](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/long.20press/near/1160717) on implementation approaches:
```quote
I'm not sure what RN v0.63's new [`Pressable` API](https://reactnative.dev/docs/pressable) ([release notes](https://reactnative.dev/blog/2020/07/06/version-0.63#pressable)) does by default, but we could try using that more than we do. Possibly it just makes it easier to manage state such that haptic feedback is given, without actually giving haptic feedback by default.

If it seems useful we could look at Expo's [`Haptics` library](https://docs.expo.io/versions/latest/sdk/haptics/).
```

vaibhavs2:
Fixes: #4649


src/webview/handleOutboundEvents.js
```diff
@@ -267,9 +268,13 @@ export const handleWebViewOutboundEvent = (
       handleImage(props, event.src, event.messageId);
       break;
 
-    case 'longPress':
+    case 'longPress': {
+      if (Platform.OS === 'android') {
```
WesleyAC, line 272: Why is this Android only? We should have haptic feedback on iOS as well.

flow-typed/npm/expo-haptics_vx.x.x.js
```diff
@@ -0,0 +1,53 @@
...
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'expo-haptics' {
+  declare module.exports: any;
+}
+
+/**
+ * We include stubs for each file inside this npm package in case you need to
+ * require those files directly. Feel free to delete any files that aren't
```
WesleyAC, line 22: Please delete these, given that you don't use them.

WesleyAC (changes requested):

src/webview/handleOutboundEvents.js
vaibhavs2: > Inside the WebView, on iOS. This one seems to already work (see chat discussion) on iOS 13+. No action needed, 

said by @gnprice  [here](https://github.com/zulip/zulip-mobile/issues/4649#issue-856264579) in comment :)

WesleyAC: Ah, I see. Please add a comment stating that. Particularly, we should note that this means that haptic feedback will not work on iOS <13.

I would add a `longPressHapticFeedback` function in this file, so that you can have the check and the comment only in one place. This will also be useful if we, for instance, choose to [detect the iOS version](https://reactnative.dev/docs/platform-specific-code#detecting-the-ios-version).

WesleyAC:
@gnprice wrote in #4649:

> On Android there's a [specific haptic response](https://developer.android.com/reference/android/view/HapticFeedbackConstants#LONG_PRESS) that is standard for long-press, and we should be sure to use that. On iOS I'm not sure there's a single standard, but we should use something similar to other iOS apps, as standard as possible -- which empirically means it'll be pretty different from the Android standard. Ideally we can get both at once by using some library made for RN.

This doesn't do that, right? We should consider using a library like [react-native-haptic-feedback](https://github.com/junina-de/react-native-haptic-feedback) instead of the expo thing, since that actually supports the long-press specific haptic feedback API.

src/utils/hapticFeedbacks.js
```diff
@@ -0,0 +1,7 @@
+/* @flow strict-local */
```
WesleyAC, line 1: I don't think we need this file, given that the implementation is just one line.

IMO, this obscures things more than it helps.

vaibhavs2:
> We should consider using a library like [react-native-haptic-feedback](https://github.com/junina-de/react-native-haptic-feedback) 

should we favour this  "if Haptic is disabled in the Android system settings,  allow ignoring the setting and trigger haptic feedback"?

src/utils/hapticFeedbacks.js
vaibhavs2: I added 'selection' for IOS , because there is no deault given for IPhone 6s and 6s+ in [code](https://github.com/junina-de/react-native-haptic-feedback/blob/7372436c2cd70a77447312f9e47cbc7be1709fa2/ios/RNReactNativeHapticFeedback.m#L50), 
and now It has more than one line, so leaved this file :)

src/utils/hapticFeedbacks.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * In Android feedback trigger for all devices irrespective of system settings
+ */
+export const longPressHapticFeedback = () => {
+  const hapticTriggerType = Platform.select({
+    ios: 'selection',
+    android: 'longPress',
+  });
+  ReactNativeHapticFeedback.trigger(hapticTriggerType, {
+    ignoreAndroidSystemSettings: true,
```
WesleyAC, line 17: We shouldn't ignore the system settings here.

src/utils/hapticFeedbacks.js
```diff
@@ -0,0 +1,19 @@
...
+import ReactNativeHapticFeedback from 'react-native-haptic-feedback';
+
+/**
+ * Haptic feedback will trigger if iOS >= 10 OR Device >= iPhone6s.
+ *
+ * In Android feedback trigger for all devices irrespective of system settings
+ */
+export const longPressHapticFeedback = () => {
+  const hapticTriggerType = Platform.select({
+    ios: 'selection',
```
WesleyAC, line 13: Just FYI, this is the default that `react-native-haptic-feedback` will do on iOS if we do `longPress` — you can see that in the [upstream code](https://github.com/junina-de/react-native-haptic-feedback/blob/8e4d598ad9be886325316b0e2140df8df624a91f/ios/RNReactNativeHapticFeedback.m#L51).

I think it's still reasonable to have this, just so if that changes, we can change this ourselves, but I don't think we really need this, strictly speaking.

WesleyAC (changes requested):
Thanks for the changes! Once you remove `ignoreAndroidSystemSettings`, this should be good to go.

src/utils/hapticFeedbacks.js
vaibhavs2: Look at the next line [upstream code](https://github.com/junina-de/react-native-haptic-feedback/blob/8e4d598ad9be886325316b0e2140df8df624a91f/ios/RNReactNativeHapticFeedback.m#L54) there  it do not have default for supportsHapticFor6SAnd6SPlus, that's why I explicitly provide the `selection` type

WesleyAC: Ah, cool, that makes sense. Looking at [this table](https://iosref.com/ios), it seems like the 6s/6s Plus support iOS 14, which means we still support them.

It's probably worth filing a upstream bug about that, I'll go do that now.

WesleyAC: Looks like this is junina-de/react-native-haptic-feedback#47

src/utils/hapticFeedbacks.js
vaibhavs2: >should we favour this "if Haptic is disabled in the Android system settings, allow ignoring the setting and trigger haptic feedback"?
I aksed about that before :)
well, I will remove that, should i remove and push -f or wait for request that we did on upstream?

WesleyAC: Go ahead and push with the fix. I don't see a reason to wait on upstream, since the helper function that you wrote gets around the problem.

gnprice:
I posted some questions about this [in the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/long.20press/near/1171547).

PR #4690, opened 2021-04-21
sentry: Document, and partly clean up, handling Sentry in release process

gnprice:
There's a step in the process I've had for making releases which hasn't been written down in the release how-to doc, partly because it feels like kind of a silly way of doing it: I have a local branch with a commit that just fills in a couple of secrets, and I rebase it atop each release commit before building.

Here, convert one of the things that branch was doing into a more reasonable approach, and document it. And I didn't as easily find today a way to replace the other thing the branch was doing with some more reasonable approach, so for that one just document the existing local-branch solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,10 @@ npm-debug.log
 /ios/webview
 # /android/app/build/... is already covered above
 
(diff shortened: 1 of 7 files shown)


chrisbobbe:
Merged, thanks!

Issue #3395, opened 2019-03-12
Some emojis (digits and single letters) are displayed as text


ungps:
Hello,

When composing a message, although `:0:` `:1:` `:2:`  ... `:10:` emojis are detected by the auto-completion system, they are displayed as plain text, not as emojis (:zero: :one: :two: ... :keycap_ten:). Also, they cannot be added as reactions. It seems that `:zero:` `:one:` `:two:` `:three:` ... `:ten:` work just fine in both cases. [Preview](https://gfycat.com/OilyIdioticAcornwoodpecker)

A similar problem is with `:a:` `:b:`, which are detected by the auto-completion system, but are not displayed at all (not even as `:a:` `:b:`). They can be added as reactions without any problem. [Preview](https://gfycat.com/ImpressiveSaltyGuineapig)

In the web application, `:0:` `:1:` `:2:` ... `:10:` are not in the list of emojis (but `:zero:` ... `:ten:` are) and are displayed as plain text. `:a:` and `:b:` are displayed correctly as :a: and :b: in messages. Neither `:digit:`, nor `:a:` `:b:` can be auto-completed.

I want to fix this, but I am not sure which would be the best solution. The easy solution would be to remove these emojis (`:digit:`) from emoji list (so the mobile app would be on par with web app). The other solution would be to add support for them.

P.S: Tested on Android 9.0, Zulip 23.3.112.

timabbott:
@HarshitOnGitHub may remember this issue on the webapp.

HarshitOnGitHub:
We need to update emoji code mapping from zulip master. Last update was based on the following commit https://github.com/zulip/zulip/commits/169de2f9cb9c4b0d30b943508d44453d3a4571f9 after which a change to remove number emojis was merged. https://github.com/zulip/zulip/commits/fc7e654d7b247eaf69afa513c62ae1922871f0eb
@jainkuniya FYI

HarshitOnGitHub:
Okay just checked #3100 seems a complete fix for these issues. :tada:

gnprice:
Thanks @ungps for this report!

The issue with `:1:` etc. and `:one:` etc. was fixed recently in #4551. (It would also be fixed by #2956, which is what #3100 was aimed at.)

The issue with `:a:` etc. turns out to be different -- it has the same cause as #3517. The issue is that the Zulip emoji database gives `:a:` a code of `1f170`, meaning that it should correspond to the Unicode codepoint U+1F170:
[...]

PR #4693, opened 2021-04-23
emoji: Fix keypad emoji, overriding broken codes from server's list.

gnprice:
The Zulip emoji database has codepoint sequences for these emoji that
aren't actually valid emoji sequences according to the Unicode spec.

Some browsers accept the forms we've been using, but others don't.
In particular, newer Chrome versions don't, since about 2019.
We discussed fixing the server's data here in 2019:
  https://chat.zulip.org/#narrow/stream/3-backend/topic/unicode.20emojis/near/758828
but never did, and then didn't come back to this.  Hopefully it
will get fixed sometime, but first let's just work around it here.

Fixes: #3517

---

Diagnosing this issue today caused me to look back at an earlier related report #3395. That one turns out to have a very similar fix, so I've just added that on to this PR:

---

emoji: Fix `:a:`, `:b:`, `:o:`, `:p:` to have emoji-style presentation.

This causes them to render (at least on Android) with bigger, colored
glyphs that look a lot like the glyphs we use in the webapp.

Without this, they render with text-style glyphs: transparent letter
on white background, which also means that in a light theme they're
completely invisible.

These are the only emoji we have in this range of letter characters.
There are some others like :ok: that are from neighboring characters
representing multiple letters, but those all seem to be doing fine
already.

Fixes: #3395

### DIFF
--- a/src/emoji/__tests__/data-test.js
+++ b/src/emoji/__tests__/data-test.js
@@ -18,12 +18,16 @@ describe('codeToEmojiMap', () => {
   test('works for some single-codepoint emoji', () => {
     check('1f44d', '👍', '\u{1f44d}');
     check('1f308', '🌈', '\u{1f308}');
-    check('1f308', '🌈', '\u{1f308}');
   });
-  test('works for some multi-codepoint emoji', () => {
-    check('0030-20e3', '0⃣', '0\u{20e3}');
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks, @gnprice! Merged, after checking these on iOS and the office Android device.

PR #4694, opened 2021-04-23
Structure redux-persist code toward doing writes in batches with `AsyncStorage.multiSet`.

chrisbobbe:
I'd like this to be reviewed/merged after #4709. While that PR is advertised as just adding some tests, it also does its own cleanup of some stuff in redux-persist, so I'd rather not go through reordering it with this PR if it can be avoided.

-----

I ended up doing this more incrementally than I think we'd said I needed to, on the call today. But it's a tricky and important area, and it doesn't have any tests, so this'll hopefully make it easier to track bugs in review than if I'd done it in just a commit or two. I also found that this incremental approach wasn't unusually hard, after all—at least, not when making the changes to get to this point. (It might get harder later on.)

After this series of commits, I believe we give `AsyncStorage` instructions to write out the right updates, in the right order. One flaw that remains after this series is that we don't wait for one write to complete before starting the next. We'll want to fix that, probably after changes like these have landed (but could do it before if we found a good way).

Related: #4587, #4588

### DIFF
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -32,36 +32,32 @@ export default function createPersistor (store, config) {
   // initialize stateful values
   let lastState = {}
   let paused = false
-  let storesToProcess = []
   let writeInProgress = false
 
   store.subscribe(() => {
-    if (paused) return
+    if (paused || writeInProgress) return
+    writeInProgress = true
 
     const state = store.getState()
+    const storesToProcess = []
 
     Object.keys(state).forEach((key) => {
       if (!passWhitelistBlacklist(key)) return
       if (lastState[key] === state[key]) return
-      if (storesToProcess.indexOf(key) !== -1) return
       storesToProcess.push(key)
     })
 
-    if (!writeInProgress) {
-      writeInProgress = true
-      const timeIterator = setInterval(() => {
-        if (storesToProcess.length === 0) {
-          clearInterval(timeIterator)
-          writeInProgress = false
-          return
-        }
+    const timeIterator = setInterval(() => {
+      if (storesToProcess.length === 0) {
+        clearInterval(timeIterator)
+        writeInProgress = false
+        return
+      }
 
-        const key = storesToProcess.shift()
-        const storageKey = createStorageKey(key)
-        const endState = store.getState()[key]
-        storage.setItem(storageKey, serializer(endState)).catch(warnIfSetError(key))
-      }, 0)
-    }
+      const key = storesToProcess.shift()
+      const storageKey = createStorageKey(key)
+      storage.setItem(storageKey, serializer(state[key])).catch(warnIfSetError(key))
+    }, 0)
 
     lastState = state
   })
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -37,8 +37,27 @@ export default function createPersistor (store, config) {
   store.subscribe(() => {
     if (paused || writeInProgress) return;
 
-    writeOnce(store.getState());
+    write();
   })
+
+  async function write() {
+    // Take the lock.
+    writeInProgress = true;
+    // Then yield so the `subscribe` callback can promptly return.
+    await new Promise(r => setTimeout(r, 0));
+
+    try {
+      let state = undefined;
+      // eslint-disable-next-line no-cond-assign
+      while ((state = store.getState()) !== lastWrittenState) {
+        await writeOnce(state);
+      }
+    } finally {
+      // Release the lock, so the next `subscribe` callback will start the
+      // loop again.
+      writeInProgress = false;
+    }
+  }
 
   /**
    * Update the storage to the given state.
@@ -47,8 +66,6 @@ export default function createPersistor (store, config) {
    * On completion, sets `lastWrittenState` to `state`.
    */
   async function writeOnce(state) {
-    writeInProgress = true
-
     // Atomically collect the subtrees that need to be written out.
     const updatedSubstates = [];
     for (const key of Object.keys(state)) {
@@ -82,7 +99,6 @@ export default function createPersistor (store, config) {
     }
 
     // Record success.
-    writeInProgress = false
     lastWrittenState = state
   }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -49,13 +49,17 @@ export default function createPersistor (store, config) {
   async function writeOnce(state) {
     writeInProgress = true
 
-    const updatedSubstates = []
-
-    Object.keys(state).forEach((key) => {
-      if (!passWhitelistBlacklist(key)) return
-      if (lastWrittenState[key] === state[key]) return
-      updatedSubstates.push([key, state[key]])
-    });
+    // Atomically collect the subtrees that need to be written out.
+    const updatedSubstates = [];
+    for (const key of Object.keys(state)) {
+      if (!passWhitelistBlacklist(key)) {
+        continue;
+      }
+      if (state[key] === lastWrittenState[key]) {
+        continue;
+      }
+      updatedSubstates.push([key, state[key]]);
+    }
 
     const writes = []
     while (updatedSubstates.length > 0) {
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -57,9 +57,12 @@ export default function createPersistor (store, config) {
         writes.push([key, serializer(substate)])
       }
 
-      writes.forEach(([key, serializedSubstate]) => {
-        storage.setItem(createStorageKey(key), serializedSubstate).catch(warnIfSetError(key))
-      })
+      if (writes.length > 0) { // `multiSet` doesn't like an empty array
+        // Warning: not guaranteed to be done in a transaction.
+        storage.multiSet(
+          writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
+        ).catch(warnIfSetError(writes.map(([key, _]) => key)));
+      }
 
       writeInProgress = false
       lastState = state
@@ -116,9 +119,17 @@ export default function createPersistor (store, config) {
   }
 }
 
-function warnIfSetError (key) {
+function warnIfSetError (keys) {
   return function setError (err) {
-    if (err) { logging.warn(err, { message: 'Error storing data for key:', key }) }
+    if (err) {
+      logging.warn(
+        err,
+        {
+          message: 'An error was encountered while trying to persist this set of keys',
+          keys
+        }
+      );
+    }
   }
 }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -35,19 +35,20 @@ export default function createPersistor (store, config) {
   let writeInProgress = false
 
   store.subscribe(() => {
-    if (paused || writeInProgress) return
-    writeInProgress = true
-
-    const state = store.getState()
-    const updatedSubstates = []
-
-    Object.keys(state).forEach((key) => {
-      if (!passWhitelistBlacklist(key)) return
-      if (lastState[key] === state[key]) return
-      updatedSubstates.push([key, state[key]])
-    });
+    if (paused || writeInProgress) return;
 
     (async () => {
+      writeInProgress = true
+
+      const state = store.getState()
+      const updatedSubstates = []
+
+      Object.keys(state).forEach((key) => {
+        if (!passWhitelistBlacklist(key)) return
+        if (lastState[key] === state[key]) return
+        updatedSubstates.push([key, state[key]])
+      });
+
       while (updatedSubstates.length > 0) {
         await new Promise(r => setTimeout(r, 0));
 
@@ -56,9 +57,8 @@ export default function createPersistor (store, config) {
         storage.setItem(storageKey, serializer(substate)).catch(warnIfSetError(key))
       }
       writeInProgress = false
+      lastState = state
     })()
-
-    lastState = state
   })
 
   function passWhitelistBlacklist (key) {
@@ -97,9 +97,16 @@ export default function createPersistor (store, config) {
     resume: () => { paused = false },
     purge: (keys) => purgeStoredState({storage, keyPrefix}, keys),
 
-    // Only used in `persistStore`, to force `lastState` to update
-    // with the results of `REHYDRATE` even when the persistor is
-    // paused.
+    /**
+     * Set `lastState` to the current `store.getState()`.
+     *
+     * Only to be used in `persistStore`, to force `lastState` to update
+     * with the results of `REHYDRATE` even when the persistor is paused.
+     *
+     * If this is going to be called, it should be before any writes have
+     * begun. Otherwise it may not be effective; see
+     *   https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691739007.
+     */
     _resetLastState: () => { lastState = store.getState() }
   }
 }
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -61,11 +61,10 @@ export default function createPersistor (store, config) {
       updatedSubstates.push([key, state[key]]);
     }
 
+    // Serialize those subtrees, with yields after each one.
     const writes = []
-    while (updatedSubstates.length > 0) {
-      const [key, substate] = updatedSubstates.shift()
+    for (const [key, substate] of updatedSubstates) {
       writes.push([key, serializer(substate)])
-
       await new Promise(r => setTimeout(r, 0));
     }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -48,18 +48,14 @@ export default function createPersistor (store, config) {
     });
 
     (async () => {
-      while (true) { // eslint-disable-line no-constant-condition
+      while (updatedSubstates.length > 0) {
         await new Promise(r => setTimeout(r, 0));
-
-        if (updatedSubstates.length === 0) {
-          writeInProgress = false
-          break
-        }
 
         const [key, substate] = updatedSubstates.shift()
         const storageKey = createStorageKey(key)
         storage.setItem(storageKey, serializer(substate)).catch(warnIfSetError(key))
       }
+      writeInProgress = false
     })()
 
     lastState = state
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -30,7 +30,7 @@ export default function createPersistor (store, config) {
   const storage = config.storage;
 
   // initialize stateful values
-  let lastState = {}
+  let lastWrittenState = {}
   let paused = false
   let writeInProgress = false
 
@@ -45,7 +45,7 @@ export default function createPersistor (store, config) {
 
       Object.keys(state).forEach((key) => {
         if (!passWhitelistBlacklist(key)) return
-        if (lastState[key] === state[key]) return
+        if (lastWrittenState[key] === state[key]) return
         updatedSubstates.push([key, state[key]])
       });
 
@@ -58,14 +58,19 @@ export default function createPersistor (store, config) {
       }
 
       if (writes.length > 0) { // `multiSet` doesn't like an empty array
-        // Warning: not guaranteed to be done in a transaction.
-        storage.multiSet(
-          writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
-        ).catch(warnIfSetError(writes.map(([key, _]) => key)));
+        try {
+          // Warning: not guaranteed to be done in a transaction.
+          await storage.multiSet(
+            writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
+          )
+        } catch (e) {
+          warnIfSetError(writes.map(([key, _]) => key))(e)
+          throw e
+        }
       }
 
       writeInProgress = false
-      lastState = state
+      lastWrittenState = state
     })()
   })
 
@@ -106,16 +111,17 @@ export default function createPersistor (store, config) {
     purge: (keys) => purgeStoredState({storage, keyPrefix}, keys),
 
     /**
-     * Set `lastState` to the current `store.getState()`.
+     * Set `lastWrittenState` to the current `store.getState()`.
      *
-     * Only to be used in `persistStore`, to force `lastState` to update
-     * with the results of `REHYDRATE` even when the persistor is paused.
+     * Only to be used in `persistStore`, to force `lastWrittenState` to
+     * update with the results of `REHYDRATE` even when the persistor is
+     * paused.
      *
      * If this is going to be called, it should be before any writes have
      * begun. Otherwise it may not be effective; see
      *   https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691739007.
      */
-    _resetLastState: () => { lastState = store.getState() }
+    _resetLastWrittenState: () => { lastWrittenState = store.getState() }
   }
 }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -47,12 +47,10 @@ export default function createPersistor (store, config) {
       storesToProcess.push(key)
     })
 
-    const len = storesToProcess.length
-
     if (!writeInProgress) {
       writeInProgress = true
       const timeIterator = setInterval(() => {
-        if ((paused && len === storesToProcess.length) || storesToProcess.length === 0) {
+        if (storesToProcess.length === 0) {
           clearInterval(timeIterator)
           writeInProgress = false
           return
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -33,7 +33,7 @@ export default function createPersistor (store, config) {
   let lastState = {}
   let paused = false
   let storesToProcess = []
-  let timeIterator = null
+  let writeInProgress = false
 
   store.subscribe(() => {
     if (paused) return
@@ -49,12 +49,12 @@ export default function createPersistor (store, config) {
 
     const len = storesToProcess.length
 
-    // time iterator (read: debounce)
-    if (timeIterator === null) {
-      timeIterator = setInterval(() => {
+    if (!writeInProgress) {
+      writeInProgress = true
+      const timeIterator = setInterval(() => {
         if ((paused && len === storesToProcess.length) || storesToProcess.length === 0) {
           clearInterval(timeIterator)
-          timeIterator = null
+          writeInProgress = false
           return
         }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -93,7 +93,13 @@ export default function createPersistor (store, config) {
           writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
         )
       } catch (e) {
-        warnIfSetError(writes.map(([key, _]) => key))(e)
+        logging.warn(
+          e,
+          {
+            message: 'An error was encountered while trying to persist this set of keys',
+            keys: writes.map(([key, _]) => key)
+          }
+        );
         throw e
       }
     }
@@ -150,20 +156,6 @@ export default function createPersistor (store, config) {
      *   https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691739007.
      */
     _resetLastWrittenState: () => { lastWrittenState = store.getState() }
-  }
-}
-
-function warnIfSetError (keys) {
-  return function setError (err) {
-    if (err) {
-      logging.warn(
-        err,
-        {
-          message: 'An error was encountered while trying to persist this set of keys',
-          keys
-        }
-      );
-    }
   }
 }
 
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -37,42 +37,49 @@ export default function createPersistor (store, config) {
   store.subscribe(() => {
     if (paused || writeInProgress) return;
 
-    (async () => {
-      writeInProgress = true
-
-      const state = store.getState()
-      const updatedSubstates = []
-
-      Object.keys(state).forEach((key) => {
-        if (!passWhitelistBlacklist(key)) return
-        if (lastWrittenState[key] === state[key]) return
-        updatedSubstates.push([key, state[key]])
-      });
-
-      const writes = []
-      while (updatedSubstates.length > 0) {
-        await new Promise(r => setTimeout(r, 0));
-
-        const [key, substate] = updatedSubstates.shift()
-        writes.push([key, serializer(substate)])
-      }
-
-      if (writes.length > 0) { // `multiSet` doesn't like an empty array
-        try {
-          // Warning: not guaranteed to be done in a transaction.
-          await storage.multiSet(
-            writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
-          )
-        } catch (e) {
-          warnIfSetError(writes.map(([key, _]) => key))(e)
-          throw e
-        }
-      }
-
-      writeInProgress = false
-      lastWrittenState = state
-    })()
+    writeOnce(store.getState());
   })
+
+  /**
+   * Update the storage to the given state.
+   *
+   * The storage is assumed to already reflect `lastWrittenState`.
+   * On completion, sets `lastWrittenState` to `state`.
+   */
+  async function writeOnce(state) {
+    writeInProgress = true
+
+    const updatedSubstates = []
+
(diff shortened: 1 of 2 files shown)


src/reduxTypes.js
```diff
@@ -337,6 +337,14 @@ export type GlobalState = {|
   users: UsersState,
 |};
 
+// No substate should allow `undefined`; our use of AsyncStorage
+// depends on it. (This check will also complain on `null`, which I
+// don't think we'd have a problem with. We could try to write this
+// differently if we want to allow `null`.)
+type NonMaybeProperties<O: { ... }> = $ObjMap<O, <V>(V) => $NonMaybeType<V>>;
+type NonMaybeGlobalState = NonMaybeProperties<GlobalState>;
+((s: NonMaybeGlobalState) => undefined: GlobalState => void);
```
WesleyAC, line 346: I'm sort of curious how this works — just defining this type, but not using it puts constraints on `GlobalState`? That seems like some spooky action-at-a-distance to me, but probably I just don't understand flow.

src/boot/ZulipAsyncStorage.js
```diff
@@ -58,6 +58,23 @@ export default class ZulipAsyncStorage {
     );
   }
 
+  static async multiSet(
```
WesleyAC, line 61: Is it reasonable to test this? If not, why not?

src/third/redux-persist/createPersistor.js
```diff
@@ -116,9 +122,13 @@ export default function createPersistor (store, config) {
...
 
-function warnIfSetError (storageKey) {
-  return function setError (err) {
-    if (err && process.env.NODE_ENV !== 'production') { console.warn('Error storing data for storage key:', storageKey, err) }
+function warnIfSetErrors (storageKeys) {
+  return function setErrors (errors) {
+    if (errors && errors.length > 0 && process.env.NODE_ENV !== 'production') {
+      errors.forEach(err => {
+        console.warn('Error storing data for one of these storage keys:', storageKeys, err)
+      })
```
WesleyAC, line 130: Can we make this error function better?

How about we have it print out storageKeys once at the top, then each error individually after that?

src/third/redux-persist/createPersistor.js
```diff
@@ -66,8 +68,12 @@ export default function createPersistor (store, config) {
           return
         }
 
-        let [storageKey, serializedValue] = storesToProcess.shift()
-        storage.setItem(storageKey, serializedValue, warnIfSetError(storageKey))
+        let updatesFromQueue = storesToProcess.shift()
+        // Warning: not guaranteed to be done in a transaction.
+        storage.multiSet(
+          updatesFromQueue,
+          warnIfSetErrors(updatesFromQueue.map(([storageKey, serializedValue]) => storageKey).join(', '))
```
WesleyAC, line 75: How about we pass the `updatesFromQueue.map(([storageKey, serializedValue]) => storageKey)` in to `warnIfSetErrors`, and then do the formatting in there, rather than `join`ing here?

WesleyAC (changes requested):
Thanks for doing this! I'm excited to have this in.

Left a few comments, nothing major.

src/reduxTypes.js
chrisbobbe: Yeah, I'd be happy (and not too surprised) if we found a good, clearer alternative. The fundamental idea is that we're doing a type cast that will fail if `GlobalState` is wrong. We use that idea in `exampleData`:

```js
// Ensure every `eg.action.foo` is some well-typed action.  (We don't simply
// annotate `action` itself, because we want to keep the information of
// which one has which specific type.)
/* eslint-disable-next-line no-unused-expressions */
(action: {| [string]: Action |});
```

Since casting requires a concrete value at runtime, and we're not looking at a value of type `GlobalState`—we want to check `GlobalState` itself—I write a simple function expression, with its input annotated as desired, and cast the value of that. (The function expression is evaluated at runtime, but the function is never actually called.) This expansion of the last line might be clearer:

```js
const func1 = (s: NonMaybeGlobalState) => undefined;
type Func2 = GlobalState => void;
(func1: Func2); // This cast will fail if `GlobalState` is wrong.
```

Or, could have done

```js
const func1 = (s: NonMaybeGlobalState) => undefined;
const func2 = (s: GlobalState) => undefined;
(func1: typeof func2); // This cast will fail if `GlobalState` is wrong.
```

or

```js
type Func1 = NonMaybeGlobalState => void;
type Func2 = GlobalState => void;
const func1: Func1 = s => undefined;
(func1: Func2); // This cast will fail if `GlobalState` is wrong.
```

or even

```js
type Func1 = NonMaybeGlobalState => void;
type Func2 = GlobalState => void;
const func1: Func1 = s => undefined;
const func2: Func2 = s => undefined; // eslint-disable-line no-unused-vars
(func1: Func2); // This cast will fail if `GlobalState` is wrong.
```

chrisbobbe: > Since casting requires a concrete value at runtime

I guess another way we could do it is to fool Flow into thinking we _do_ have a value of type `GlobalState` right in front of us:

```js
// $FlowIgnore[incompatible-type]
const s: GlobalState = 'nothing-to-see-here';
(s: NonMaybeGlobalState);
```

src/third/redux-persist/createPersistor.js
```diff
@@ -72,7 +72,7 @@ export default function createPersistor (store, config) {
         let storageKey = createStorageKey(key)
         let endState = transforms.reduce((subState, transformer) => transformer.in(subState, key), stateGetter(store.getState(), key))
         if (typeof endState !== 'undefined') storage.setItem(storageKey, serializer(endState), warnIfSetError(key))
-      }, false)
```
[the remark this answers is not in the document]
chrisbobbe: Yeah that was weird.

src/boot/ZulipAsyncStorage.js
chrisbobbe: Replied in chat, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60jest-expo.2Fios.60.20and.20.60jest-expo.2Fandroid.60/near/1169139).

chrisbobbe:
Thanks for the review, @WesleyAC! Revision pushed.

src/boot/ZulipAsyncStorage.js
chrisbobbe: #4700

chrisbobbe:
Revision pushed again. This time, I'd like to block on #4700 so we can get coverage of Android-only codepaths for some tests for `ZulipAsyncStorage.multiSet`. This branch is on top of my current revision for #4700.

chrisbobbe:
Revision pushed; this is now based on #4709, which comes after #4700.

chrisbobbe:
(This needs rebasing and some tweaks following comments in #4709; I'll aim to finish those this week.)

chrisbobbe:
OK, revision pushed!

src/third/redux-persist/autoRehydrate.js
```diff
@@ -71,7 +71,9 @@ function defaultStateReconciler (state, inboundState, reducedState, log) {
     if (isStatePlainEnough(inboundState[key]) && isStatePlainEnough(state[key])) newState[key] = {...state[key], ...inboundState[key]} // shallow merge
     else newState[key] = inboundState[key] // hard set
 
-    if (log) logging.warn('redux-persist/autoRehydrate: rehydration for a key complete.', { key, newStateValue: newState[key] })
+    // This is a noisy log that runs when things go as expected; it should
+    // only be used for debugging.
+    // if (log) logging.warn('redux-persist/autoRehydrate: rehydration for a key complete.', { key, newStateValue: newState[key] })
```
WesleyAC, line 76: Can we just remove this instead?

WesleyAC (commented):
One (minor) comment, LGTM other than that :)

src/reduxTypes.js
```diff
@@ -357,6 +357,17 @@ export type GlobalState = $ReadOnly<{|
   users: UsersState,
 |}>;
 
+// No substate should allow `undefined`; our use of AsyncStorage
+// depends on it. (This check will also complain on `null`, which I
+// don't think we'd have a problem with. We could try to write this
+// differently if we want to allow `null`.)
+type NonMaybeProperties<O: { ... }> = $ObjMap<O, <V>(V) => $NonMaybeType<V>>;
+type NonMaybeGlobalState = NonMaybeProperties<GlobalState>;
```
gnprice, line 365: Neat!

src/reduxTypes.js
```diff
@@ -357,6 +357,17 @@ export type GlobalState = $ReadOnly<{|
...
+// $FlowIgnore[incompatible-type]: Pretend `s` is a `GlobalState`.
+const s: GlobalState = 'nothing-to-see-here';
+// This cast will fail typechecking if GlobalState is wrong.
+(s: NonMaybeGlobalState);
```
gnprice, line 369: Here's a slightly cleaner way to write this -- avoids creating a bogus value, by making the `GlobalState` just the hypothetical parameter to a certain function:

```suggestion
// This function definition will fail typechecking if GlobalState is wrong.
(s: GlobalState): NonMaybeGlobalState => s; // eslint-disable-line no-unused-expressions
```

src/third/redux-persist/createPersistor.js
```diff
@@ -86,12 +87,10 @@ export default function createPersistor (store, config) {
...
         try {
           let data = deserializer(subState)
-          let value = transforms.reduceRight((interState, transformer) => {
-            return transformer.out(interState, key)
-          }, data)
+          let value = data
           state = stateSetter(state, key, value)
         } catch (err) {
-          if (process.env.NODE_ENV !== 'production') console.warn(`Error rehydrating data for key "${key}"`, subState, err)
+          logging.warn('Error rehydrating data for a key', { key, subState, err })
```
gnprice, line 93: `subState` here looks like potentially a lot of data. Maybe just drop it?

Alternatively could try to include a prefix or something, to help identify whether it was compressed (and labeled with our compression header.) But I'm not sure that will necessarily add anything vs. the error object.

gnprice (commented):
Thanks @chrisbobbe and @WesleyAC!

Small code comments below.

At a higher level, I am a bit concerned about this particular algorithm and the potential performance implications.

Suppose we had a situation somewhere in the app where we make a burst of updates all in a row to some piece of the state, like `state.users`. Then:
* The old code will just repeatedly note that that key should be written out; then it'll get around to writing it out once.
  * (It does have a rather https://accidentallyquadratic.tumblr.com/ problem, in the case of many different keys, in that it runs through the list of updated keys again and again. Really it should be using a `Set`, not an array. But that's another matter.)
* The new code will serialize out each new version of that substate's value to a string, building up an array with all the different versions. Then it'll work through that queue to write out each of the different versions in turn.

Because the unit by which we store the state is a whole subtree, and because some of those can be giant (especially `state.users`, `state.unread`, and `state.messages`, and potentially others), that may mean we spend a lot of CPU to serialize, a lot of memory to store those different versions at once, and a lot of IO to write the updates out.

Also, independent of the possible repetition, there's an issue with calling `serialize` eagerly: it means we're doing that directly in the `store.subscribe` callback, which I expect will be called synchronously after the reducer on dispatch. That means that whatever CPU time it needs -- which may be significant for something big like `state.users` -- may delay React Redux in getting to update the React state so that the UI updates. Calling `serialize` in the deferred section, i.e. the `setInterval` callback, avoids that problem.

I have some ideas for possible alternative algorithms, but I'll move that to a chat thread to avoid this PR thread getting too complex.

src/reduxTypes.js
gnprice: The original version here was correct, but definitely a bit tricky to understand :slightly_smiling_face: 

> just defining this type, but not using it puts constraints on `GlobalState`? That seems like some spooky action-at-a-distance to me

In that original version it's important that it's not just defining a type -- as Chris said, it's defining a value (albeit a very boring function), and casting it. In general if you *just* define a type my experience has been that Flow may not bother to reason through all the implications and detect that the type can't possibly work; but once values are flowing around, it'll take notice.

One way of thinking about what that original version is doing -- and this is basically an explanation of variance, the same concept as at that article in Chris's last comment -- is:
* The function on the left demands a `NonMaybeGlobalState`.
* The cast says, let's treat it as a function that demands a `GlobalState`.
* The only way that can work is if every `GlobalState` can be used as a `NonMaybeGlobalState` -- i.e. if `GlobalState` is a subtype of `NonMaybeGlobalState`.
* So that's exactly what Flow checks when you take a function of one type and try to use it as another function type. (That plus a check on the return values.)
* Note that that subtype relationship between the parameters runs in the opposite direction from the one between the function types. We say that the function's parameters' types are *contravariant* on the function type itself.

In my review below I suggested an alternate simpler version: https://github.com/zulip/zulip-mobile/pull/4694#discussion_r654042192
```
// This function definition will fail typechecking if GlobalState is wrong.
(s: GlobalState): NonMaybeGlobalState => s; // eslint-disable-line no-unused-expressions
```

src/third/redux-persist/createPersistor.js
```diff
@@ -64,11 +65,11 @@ export default function createPersistor (store, config) {
...
-        let key = storesToProcess.shift()
-        let storageKey = createStorageKey(key)
-        let endState = transforms.reduce((subState, transformer) => transformer.in(subState, key), stateGetter(store.getState(), key))
-        if (typeof endState !== 'undefined') storage.setItem(storageKey, serializer(endState)).catch(warnIfSetError(key))
-      }, debounce)
+        let updatesFromQueue = storesToProcess.shift()
+        // Warning: not guaranteed to be done in a transaction.
+        storage.multiSet(updatesFromQueue)
+          .catch(warnIfSetError(updatesFromQueue.map(([storageKey, serializedValue]) => storageKey)))
+      }, 0)
```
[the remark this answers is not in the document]
chrisbobbe: Thanks for that investigation!

gnprice:
Looks like you sent a revised version of the first part of this branch as #4892. Just merged that!

chrisbobbe:
Thanks for the review! Revision pushed, following the conversation [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1212683) to [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1214393), where you suggested a better algorithm.

I'm not sure I found quite the right sequence of minimal, coherent steps to get there. There might be a better one or we might end up with some commits that squash a few changes together; let me know and I'm happy to take a different approach. 🙂

src/third/redux-persist/createPersistor.js
```diff
@@ -56,9 +57,8 @@ export default function createPersistor (store, config) {
         storage.setItem(storageKey, serializer(substate)).catch(warnIfSetError(key))
       }
       writeInProgress = false
+      lastState = state
     })()
-
-    lastState = state
```
gnprice: createPersistor [nfc]: Include more code in an async IIFE.

There's actually a change happening here which is easy to miss in the source code, and it's not obvious whether it could affect visible behavior.

In the old code, we invoke the async IIFE, then update `lastState` and return. Then later when the whole JS stack is empty, the task created by that async IIFE gets run, and the code in its body starts executing. If the loop condition is true so it hits the `await` inside there, it creates a new task and returns, and later that task gets run so the rest of the loop body gets executed, and so on with further tasks until the loop condition is false so there's no more `await`.

In the new code, we invoke the (larger) async IIFE and return. The update to `lastState` doesn't happen immediately, not until some subsequent task: whichever one finds the loop condition is false.

I *think* this doesn't end up changing behavior, but only because I think `_resetLastState` can't end up getting called between when this code runs and when a write is complete, because I think we don't ever do a write until after the only time we'd call `_resetLastState`. But that fact is a bit nonlocal and delicate, and I'd have to look harder even to be sure it's true.

I think the right fix here is that the update `lastState = state` should really happen atomically with the point where we compare `state` to the old `lastState`:
```js
       Object.keys(state).forEach((key) => {
         if (!passWhitelistBlacklist(key)) return
         if (lastState[key] === state[key]) return
```

That would emphasize the point that it represents the last state we wrote out, and we're comparing the new state against the last one we wrote out, so that that comparison is just an optimization to skip writing out redundantly the parts that already have the latest version written out.

In the old code, that's happening because the only thing that comes between the comparison and the update is the async IIFE, which doesn't do anything immediately -- though that'd be clearer if they were next to each other in the source.

In the new code, if we just do the update right after the comparison loop, then I think things will be good.

src/third/redux-persist/createPersistor.js
```diff
@@ -35,19 +35,20 @@ export default function createPersistor (store, config) {
...
-    const updatedSubstates = []
+    (async () => {
+      writeInProgress = true
 
-    Object.keys(state).forEach((key) => {
-      if (!passWhitelistBlacklist(key)) return
-      if (lastState[key] === state[key]) return
-      updatedSubstates.push([key, state[key]])
-    });
+      const state = store.getState()
```
gnprice: Moving this inside the async IIFE actually represents a change that I think is worth noting in the commit message: it means we have a yield before reading what the new state is that we're going to be writing out.

I think that's fine, and indeed probably an improvement whenever it makes a difference. If there's a burst of updates that happen synchronously (e.g., if anywhere in the code we dispatch one plain action and then another, with no `await` in between), then the code before this commit would fire this callback on the first update; set `writeInProgress` and read the state; then on the second update, do nothing; and have a task come back later to start writing, and it'd write out the first updated state even though the second one exists by then. With this change, we wait to read the new state until that task gets run, which means we take the latest state that's then available and we write that.

src/third/redux-persist/createPersistor.js
```diff
@@ -49,13 +49,18 @@ export default function createPersistor (store, config) {
...
         const [key, substate] = updatedSubstates.shift()
-        const storageKey = createStorageKey(key)
-        storage.setItem(storageKey, serializer(substate)).catch(warnIfSetError(key))
+        writes.push([key, serializer(substate)])
       }
+
+      writes.forEach(([key, serializedSubstate]) => {
+        storage.setItem(createStorageKey(key), serializedSubstate).catch(warnIfSetError(key))
+      })
+
```
gnprice: createPersistor [nfc]: Separate serializing from writing.

This is definitely not NFC -- it's basically the main behavior improvement this series makes :wink: It should narrow the race window quite a lot.

The `multiSet` is a nice bonus, which probably narrows the race window further by making just one trip from JS over to the native side. But this change already means that the `setItem` calls are no longer separated by (a) the CPU time to do the serialization, or (b) yields.

gnprice: Separately, it'd be good to have a yield between the last serialization and doing the writes.

I see that happens later in the series (by moving the yield to the bottom instead of the top of the loop body). That's fine, then, because that bit isn't critical.

src/third/redux-persist/createPersistor.js
```diff
@@ -57,9 +57,10 @@ export default function createPersistor (store, config) {
         writes.push([key, serializer(substate)])
       }
 
-      writes.forEach(([key, serializedSubstate]) => {
-        storage.setItem(createStorageKey(key), serializedSubstate).catch(warnIfSetError(key))
-      })
+      // Warning: not guaranteed to be done in a transaction.
+      storage.multiSet(
+        writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
+      ).catch(warnIfSetError(updatedSubstates));
```
gnprice: This argument to `warnIfSetError` is always an empty array at this point, right?

The code around this changes later in the branch, but a thing that'd be good to change about the final code would actually already help here. In the code at the tip, we hang on to the whole `updatedSubstates` including all the substate data:
```js
    for (const [key, substate] of updatedSubstates) {
      writes.push([key, serializer(substate)])
      await new Promise(r => setTimeout(r, 0));
    }
```
even as we generate the serialized strings. For something like `state.users`, that can be quite a big string and also a big `User[]` array, which this code might be holding the last reference to if there've been subsequent updates. It'd be good to discard each subtree as soon as we've serialized it out.

The code as of this commit does that, by mutating `updatedSubstates`. That's probably the most natural way to keep doing it, even later in the series.

Then in order to keep around the list of keys for the sake of this error logging, the serialization loop could push each key onto a new array that's just the keys. Like this:
```js
       while (updatedSubstates.length > 0) {
         const [key, substate] = updatedSubstates.shift();
         writes.push([createStorageKey(key), serializer(substate)]);
         keys.push(key);
         await new Promise(r => setTimeout(r, 0));
       }
```

gnprice (commented):
Thanks @chrisbobbe ! Sorry for the delayed review. This direction is looking good.

I've just read up through the commit that switches to `multiSet`:
e809b9f21 createPersistor [nfc]: Use a stateful `writeInProgress` flag.
6b89f2719 createPersistor: Simplify "paused" logic a bit.
c1e46f19c createPersistor: Stop blending snapshots when we initiate writes.
aeb1cacbc createPersistor [nfc]: Rename `storesToProcess` to `updatedSubstates`.
26e30d942 createPersistor: Rewrite a loop with `while` and `await`.
41e33c001 createPersistor: Clean up a `while` loop, using a loop condition.
6eef51a46 createPersistor [nfc]: Include more code in an async IIFE.
7b7db1720 createPersistor [nfc]: Separate serializing from writing.
bfa7c87b8 createPersistor: Use `storage.multiSet` instead of many `.setItem`s.
fda001a32 createPersistor: Await `storage.multiSet` before marking state as written.

Some comments below; none of them fundamental, I think.

Going AFK for dinner, so I'll return another time to review the remaining commits:
89c7e2865 createPersistor [nfc]: Define `writeOnce`, with a jsdoc.
3d880df14 createPersistor [nfc]: Tidy up some code, and comment.
40062c152 createPersistor: Move a yield from before to after serializing.
99c24da71 createPersistor [nfc]: Tidy up some more code, and comment.
f4f4232c1 createPersistor [nfc]: Comment some more.
d437384fe createPersistor: Stop ignoring state changes that happen during a write.
0b48a6760 createPersistor [nfc]: Inline a complicated helper function we use once.

src/third/redux-persist/createPersistor.js
gnprice: > For something like `state.users`, that can be quite a big string and also a big `User[]` array, which this code might be holding the last reference to if there've been subsequent updates. It'd be good to discard each subtree as soon as we've serialized it out.

Hmm, I guess actually we hang on to each subtree in any case, in the form of `lastState`, in order to be able to check later if it's changed. So there isn't really a benefit to be had from discarding it from `updatedSubstates`.

This memory-consumption issue gets solved by converting the large data structures to Immutable -- if we had an `Immutable.List<User>` instead of a `User[]` there, then it'd share most of its memory with later versions of itself. So that will make it definitely not a concern in the future.

If it were really essential to solve it in a context where we did still have large arrays that didn't get to share memory like that, we could do it by having `lastState` keep around only weak references (using [JS's `WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef), probably.) That'd ensure that `lastState` never caused a piece of state to stick around when it otherwise wouldn't, and it'd still give us all the information we need because the only thing we really care about in an old state subtree is whether it's the same value that is the new state subtree. But I don't think the complication is worth it.

As of this commit, you could do the `keys.push(key)` thing, or could also just have the log message not mention the keys -- that'd be fine as an intermediate state in the branch.

src/third/redux-persist/createPersistor.js
```diff
@@ -57,13 +57,18 @@ export default function createPersistor (store, config) {
...
+          writes.map(([key, serializedSubstate]) => [createStorageKey(key), serializedSubstate])
+        )
+      } catch (e) {
+        warnIfSetError(updatedSubstates)(e)
+        throw e
+      }
 
       writeInProgress = false
-      lastState = state
+      lastWrittenState = state
```
gnprice: As mentioned above, this update should happen atomically with where the state is compared with the old value. That'll make this new name not so accurate.

I agree it'd be good to make the name more specific like this. I'm not immediately thinking of an equally pithy name, though -- it's basically the last state we've *started* writing.

… Hmmm. I'm now thinking about the possibility that the write can fail, and realizing that that points back in the direction of the "last written" semantics that this branch gives it -- effectively that means that when a write fails, the next time we have an update we'll retry writing the keys that we tried to write last time, as well as any that have newly changed. (Though it isn't perfect at that: if some `state.foo` changes from A to B, we have an error writing that update, and then it changes back to A, we won't try to write A for it, even though it's possible that the update to B actually made it to disk. This mostly… hmm, I was going to say this doesn't end up mattering for us because our subtrees are objects that get new identities every time, but it can actually matter where we use Immutable collections, because e.g. a new empty `Immutable.Map` is identical to any previous empty `Immutable.Map`. … Thinking some more, I think to properly handle this it's probably best to do so explicitly, with like a `Set` of keys that we've started writes to and aren't sure we succeeded. I can try doing that after this PR is in -- I don't think it's critical compared to the holes this subsystem inherently has due to the races that this PR mitigates.)

I guess the upshot of that last paragraph is: I think it still makes best sense to do this update atomically with where the state gets compared with the old value. That makes it clear what this variable's job is, and it makes it easy to reason through at least what the sequence of *attempted* writes is: it's the diffs from one state, to another, to another, etc., with the endpoint of one attempted write being the starting point of the next, and `_resetLastState` effectively suppresses one link in that chain from getting written out. The chain of *actual* writes gets a gap if a link fails, but at least it's straightforward to see that, and we'll fix it separately.

src/third/redux-persist/createPersistor.js
```diff
@@ -49,13 +49,14 @@ export default function createPersistor (store, config) {
...
-    Object.keys(state).forEach((key) => {
-      if (!passWhitelistBlacklist(key)) return
-      if (lastWrittenState[key] === state[key]) return
-      updatedSubstates.push([key, state[key]])
-    });
+    // Atomically collect the subtrees that need to be written out.
+    const updatedSubstates = [];
+    for (const key of Object.keys(state)) {
+      if (state[key] === lastWrittenState[key] || !passWhitelistBlacklist(key)) {
+        continue;
```
gnprice, line 72: I think these are actually cleanest as two separate conditionals. The reasoning behind them -- what you'd say in a comment to explain them, if commenting more verbosely like that -- is pretty different.

src/third/redux-persist/createPersistor.js
```diff
@@ -37,18 +37,34 @@ export default function createPersistor (store, config) {
...
+    await new Promise(r => setTimeout(r, 0));
+
+    try {
+      let state = undefined;
+      // eslint-disable-next-line no-cond-assign
+      while ((state = store.getState()) !== lastWrittenState) {
+        await writeOnce(state);
+      }
+    } finally {
+      // Release the lock, so the next `subscribe` will start the loop again.
```
gnprice, line 56: nit: next `subscribe` *callback*, really -- `subscribe` itself we call just the once

src/third/redux-persist/createPersistor.js
```diff
@@ -139,18 +143,6 @@ export default function createPersistor (store, config) {
   }
 }
 
-function warnIfSetError (updatedSubstates) {
-  return function setError (err) {
-    if (err) {
-      logging.warn(
```
gnprice: createPersistor [nfc]: Inline a complicated helper function we use once.
    
    This is the first commit in a series that will rework
    `createPersistor` to use `storage.multiSet` instead of many
    `storage.setItem`s, taking care not to detract from performance. For
    background, please see
      https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1214393.

nit: can drop that commit-message body, as this is now the last commit in the series instead of the first :slightly_smiling_face:

src/third/redux-persist/createPersistor.js
gnprice: Ha, and I see this line comes straight from my draft:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1214393
:shrug:

src/third/redux-persist/createPersistor.js
gnprice: Aha, and I see that both the name `lastWrittenState`, and its getting updated here at the end, are as in [my draft](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20storage/near/1214393). Heh.

Well, it does have this hole I mentioned above:
> if some `state.foo` changes from A to B, we have an error writing that update, and then it changes back to A, we won't try to write A for it, even though it's possible that the update to B actually made it to disk.

And probably we should fix that. But as I said:
> I don't think it's critical compared to the holes this subsystem inherently has due to the races that this PR mitigates

and really that applies to anything about the case where a write fails, because I don't think that's likely, and if it does happen I'm not sure we have much chance of not just being doomed. (If it does happen, most likely the filesystem is full or the storage device is failing, and any subsequent write attempt is likely to fail too.)

So I'm happy merging this series with the logic it has in this respect and fixing it later. No need to move the update next to the comparison.

gnprice (commented):
OK, and read the rest! Comments below.

src/third/redux-persist/createPersistor.js
gnprice: OK, as described below: https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691735576 I see that updating `lastState` here at the end was actually in my draft, and I'm fine with putting it there. (As discussed in that thread there's a bit of a loophole in the error case where the write fails, but that's a lower-order concern than the problems this subsystem has anyway and that this PR mitigates, and we can handle that later.)

Should identify in the commit message that that's happening, though. And should also mention in… well, `_resetLastState` should never be called after a write may already be in progress, because if so it may not be effective. Should mention that in the comment on that method.

gnprice:
OK, and in the end I think my requested changes are all either in commit messages or comments, apart from one tiny refactor:
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691728052
and one bugfix in an error case:
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691719814
which could just consist of leaving out some of the debug data at that point, because that issue is fixed later in the series anyway.

Some comments ask for a couple of more-substantive changes, but then I realized one of those wouldn't help at all, and the other was something that was in my own draft :stuck_out_tongue: and will have a small gap either way that's best fixed in followup.

src/third/redux-persist/createPersistor.js
```diff
@@ -32,36 +32,32 @@ export default function createPersistor (store, config) {
   // initialize stateful values
   let lastState = {}
   let paused = false
-  let storesToProcess = []
   let writeInProgress = false
 
   store.subscribe(() => {
-    if (paused) return
+    if (paused || writeInProgress) return
```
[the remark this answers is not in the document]
chrisbobbe: Right, exactly.

chrisbobbe:
Thanks for the review! Revision pushed. Please let me know if I've missed something, or if you have ideas for followup you'd like me to do instead of you. 🙂

gnprice:
Thanks for the revision! The changes to the branch all look good.

There are a few more comment / commit-message changes I mentioned above that would still be good to do:
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691739007
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691687531
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691689199

src/third/redux-persist/createPersistor.js
chrisbobbe: > it means we have a yield before reading what the new state is that we're going to be writing out.

Interesting. I don't think I follow, yet, though. To find the yield you're talking about, I'm looking for an `await` in the async IIFE's body, before the `const state = store.getState()` line. I don't see one, so I don't see the yield.

Is there a hidden yield that gets done without `await`, maybe at the start of invoking any async function?

src/third/redux-persist/createPersistor.js
chrisbobbe: > Should mention that in the comment on that method.

I've made that comment a jsdoc, since it seemed more of an interface question than implementation.

chrisbobbe:
Thanks for pointing those out! Revision pushed. I have a question on one of them, https://github.com/zulip/zulip-mobile/pull/4694#discussion_r707831513.

src/third/redux-persist/createPersistor.js
gnprice: Hmm, no, that's right -- I think what I wrote above is mistaken.

This kind of question can be confusing (evidently!), in part because it can vary between languages. Here's notes I took on it a few years ago, with an empirical demo:
https://github.com/gnprice/async-playground/blob/7da42982af9e05540833a96ef7cb82dbaa397b92/async.js

Those confirm that there's no yield at entry to an async function. There is at exit (i.e., when the caller awaits the result), as well as at any awaits inside its body.

I also appear to have been confused about this when I wrote this other comment:
https://github.com/zulip/zulip-mobile/pull/4694#discussion_r691685866
> In the old code, we invoke the async IIFE, then update `lastState` and return. Then later when the whole JS stack is empty, the task created by that async IIFE gets run, and the code in its body starts executing. If the loop condition is true so it hits the `await` inside there, it creates a new task and returns […]

This point doesn't change much there, though: it basically makes the sequence a step shorter.

gnprice:
OK, and merged!

Your question caused me to realize that another of my comments was mistaken. I corrected the text you'd quoted from that comment, producing the commit message in b54737f5956bac214b49cb7386de65e73dc919ef.

PR #4697, opened 2021-04-25
NavBarBackButton: Give the right icon for back button.

Gautam-Arora24:
Provide a suitable icon for the back button. The icon will be
platform-specific i.e the name of the icon will be 'arrow-left'
for android and 'chevron-left' for iOS.

**On iOS**
Screenshot 2021-04-25 at 2 30 38 PM

### DIFF
--- a/src/nav/NavBarBackButton.js
+++ b/src/nav/NavBarBackButton.js
@@ -1,5 +1,6 @@
 /* @flow strict-local */
 import React from 'react';
+import { Platform } from 'react-native';
 
 import { navigateBack } from '../actions';
 import NavButton from './NavButton';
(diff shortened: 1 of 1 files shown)


WesleyAC (approved):
This looks good to me, thanks for fixing this!

It's possible we'll end up wanting to do this in more places in the future, and we'll want to have a helper function for this elsewhere, but this looks good for now.

Gautam-Arora24:
I think the code in this PR will change the icon for every back button used in the mobile app. So, there would be no need for creating any other helper function, right?

WesleyAC:
Yeah, that's my understanding as well — my point was that if we add more back buttons in other places in the future, we may want to make a helper function. We certainly don't need one right now.

gnprice:
Merged, thanks!

PR #4698, opened 2021-04-27
caughtUp: Fix bug that overwrote CaughtUp state.

WesleyAC:
When we made a request for a narrow, we would overwrite the CaughtUp
state for that narrow in the other direction - i.e. if we were fetching
old messages, then we would forget that we've seen the most recent
message.

This commit fixes that by only updating each bit of the CaughtUp state
from false to true (on a fetch-messages result), and never the other
way around.


gnprice:
Looks good, thanks, merged!

For cross-reference in the future when this is no longer obvious: a previous version of this change was in drafts of #4678 and discussed there.

PR #4699, opened 2021-04-27
Muted users: part 2

WesleyAC:
The continuation of #4678. I moved the CaughtUp fix to #4698, since that's somewhat orthogonal (although this PR doesn't work terribly well without it).

All of the comments by @gnprice in #4678 should be resolved in this.

### DIFF
--- a/src/pm-conversations/PmConversationList.js
+++ b/src/pm-conversations/PmConversationList.js
@@ -1,8 +1,9 @@
 /* @flow strict-local */
 import React, { useCallback } from 'react';
 import { FlatList } from 'react-native';
+import { useDispatch } from '../react-redux';
 
-import type { Dispatch, PmConversationData, UserOrBot } from '../types';
+import type { PmConversationData, UserOrBot } from '../types';
 import { createStyleSheet } from '../styles';
 import { type PmKeyUsers } from '../utils/recipient';
 import { pm1to1NarrowFromUser, pmNarrowFromUsers } from '../utils/narrow';
@@ -18,7 +19,6 @@ const styles = createStyleSheet({
 });
 
 type Props = $ReadOnly<{|
-  dispatch: Dispatch,
   conversations: PmConversationData[],
 |}>;
 
@@ -26,18 +26,20 @@ type Props = $ReadOnly<{|
  * A list describing all PM conversations.
  * */
 export default function PmConversationList(props: Props) {
+  const dispatch = useDispatch();
+
   const handleUserNarrow = useCallback(
     (user: UserOrBot) => {
-      props.dispatch(doNarrow(pm1to1NarrowFromUser(user)));
+      dispatch(doNarrow(pm1to1NarrowFromUser(user)));
     },
-    [props.dispatch],
+    [dispatch],
   );
 
   const handleGroupNarrow = useCallback(
     (users: PmKeyUsers) => {
-      props.dispatch(doNarrow(pmNarrowFromUsers(users)));
+      dispatch(doNarrow(pmNarrowFromUsers(users)));
     },
-    [props.dispatch],
+    [dispatch],
   );
 
   const { conversations } = props;
--- a/src/pm-conversations/PmConversationList.js
+++ b/src/pm-conversations/PmConversationList.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import React, { PureComponent } from 'react';
+import React, { useCallback } from 'react';
 import { FlatList } from 'react-native';
 
 import type { Dispatch, PmConversationData, UserOrBot } from '../types';
@@ -25,45 +25,49 @@ type Props = $ReadOnly<{|
 /**
  * A list describing all PM conversations.
  * */
-export default class PmConversationList extends PureComponent<Props> {
-  handleUserNarrow = (user: UserOrBot) => {
-    this.props.dispatch(doNarrow(pm1to1NarrowFromUser(user)));
-  };
+export default function PmConversationList(props: Props) {
+  const handleUserNarrow = useCallback(
+    (user: UserOrBot) => {
+      props.dispatch(doNarrow(pm1to1NarrowFromUser(user)));
+    },
+    [props.dispatch],
+  );
 
-  handleGroupNarrow = (users: PmKeyUsers) => {
-    this.props.dispatch(doNarrow(pmNarrowFromUsers(users)));
-  };
+  const handleGroupNarrow = useCallback(
+    (users: PmKeyUsers) => {
+      props.dispatch(doNarrow(pmNarrowFromUsers(users)));
+    },
+    [props.dispatch],
+  );
 
-  render() {
-    const { conversations } = this.props;
+  const { conversations } = props;
 
-    return (
-      <FlatList
-        style={styles.list}
-        initialNumToRender={20}
-        data={conversations}
-        keyExtractor={item => item.key}
-        renderItem={({ item }) => {
-          const users = item.keyRecipients;
-          if (users.length === 1) {
-            return (
-              <UserItem
-                userId={users[0].user_id}
-                unreadCount={item.unread}
-                onPress={this.handleUserNarrow}
-              />
-            );
-          } else {
-            return (
-              <GroupPmConversationItem
-                users={users}
-                unreadCount={item.unread}
-                onPress={this.handleGroupNarrow}
-              />
-            );
-          }
-        }}
-      />
-    );
-  }
+  return (
+    <FlatList
+      style={styles.list}
+      initialNumToRender={20}
+      data={conversations}
+      keyExtractor={item => item.key}
+      renderItem={({ item }) => {
+        const users = item.keyRecipients;
+        if (users.length === 1) {
+          return (
+            <UserItem
+              userId={users[0].user_id}
+              unreadCount={item.unread}
+              onPress={handleUserNarrow}
+            />
+          );
+        } else {
+          return (
+            <GroupPmConversationItem
+              users={users}
+              unreadCount={item.unread}
+              onPress={handleGroupNarrow}
+            />
+          );
+        }
+      }}
+    />
+  );
 }
--- a/src/pm-conversations/PmConversationList.js
+++ b/src/pm-conversations/PmConversationList.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import React, { useCallback } from 'react';
 import { FlatList } from 'react-native';
-import { useDispatch } from '../react-redux';
+import { useDispatch, useSelector } from '../react-redux';
 
 import type { PmConversationData, UserOrBot } from '../types';
 import { createStyleSheet } from '../styles';
@@ -10,6 +10,7 @@ import { pm1to1NarrowFromUser, pmNarrowFromUsers } from '../utils/narrow';
 import UserItem from '../users/UserItem';
 import GroupPmConversationItem from './GroupPmConversationItem';
 import { doNarrow } from '../actions';
+import { getMutedUsers } from '../selectors';
 
 const styles = createStyleSheet({
   list: {
@@ -43,6 +44,7 @@ export default function PmConversationList(props: Props) {
   );
 
   const { conversations } = props;
+  const mutedUsers = useSelector(getMutedUsers);
 
   return (
     <FlatList
@@ -53,13 +55,14 @@ export default function PmConversationList(props: Props) {
       renderItem={({ item }) => {
         const users = item.keyRecipients;
         if (users.length === 1) {
-          return (
-            <UserItem
-              userId={users[0].user_id}
-              unreadCount={item.unread}
-              onPress={handleUserNarrow}
-            />
-          );
+          const user_id = users[0].user_id;
+          if (mutedUsers.has(user_id)) {
+            return null;
+          } else {
+            return (
+              <UserItem userId={user_id} unreadCount={item.unread} onPress={handleUserNarrow} />
+            );
+          }
         } else {
           return (
             <GroupPmConversationItem
--- a/src/users/UserList.js
+++ b/src/users/UserList.js
@@ -1,12 +1,14 @@
 /* @flow strict-local */
 import React from 'react';
 import { SectionList } from 'react-native';
+import { useSelector } from '../react-redux';
 
 import type { PresenceState, UserOrBot } from '../types';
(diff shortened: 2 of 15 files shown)


src/webview/static/base.css
```diff
@@ -142,14 +142,18 @@ body {
   padding: 0 16px 16px 80px;
 }
 
-.message-muted {
+.message[data-mute-state="hidden"] {
```
gnprice: Then the rest of these first four commits (i.e. everything in those commits apart from adding the `_` plumbing) would be best squashed together. That will simplify the changes quite a bit, because there's a lot of bits like this one that do one thing in an earlier commit and then go a different way in a later commit.

Which is a very natural and useful thing to do while drafting how something should work! Just when it's done, it's quite helpful to clean up the series so it's optimized for the reader to understand.

src/webview/html/messageAsHtml.js
```diff
@@ -134,6 +143,14 @@ $!${divOpenHtml}
   <div class="static-timestamp">${messageTime}</div>
 </div>
 `;
+  // TODO: i18n
```
gnprice, line 146: I think this line can be deleted :slightly_smiling_face:

src/webview/js/js.js
```diff
@@ -926,9 +943,27 @@ const handleLongPress = (target: Element) => {
...
+  const targetType = target.matches('.header')
+    ? 'header'
+    : target.matches('a')
+      ? 'link'
+      : 'message';
+  const messageNode = target.closest('.message');
+
+  if (
+    targetType === 'message'
+    && messageNode
```
gnprice, line 957: This is still a bit messy, but it's only extending a way that the existing logic in this function is already messy, and I think it's fine for now.

When I merge this, I may try splitting the conditionals in this function up to follow the style of the click handler above:
* first, see if target matches `.header`, if so act on that and return.
* then, see if matches `a`, if so act on that and return.
* then, see if muted message, if so act on that and return.
* finally, act on default case.

because I think that'll make it somewhat clearer, and also the difference in clarity will grow rapidly if we add any more cases here.

static/translations/messages_en.json
```diff
@@ -231,5 +231,6 @@
   "Signed out": "Signed out",
   "Remove account?": "Remove account?",
   "This will make the mobile app on this device forget {email} on {realmUrl}.": "This will make the mobile app on this device forget {email} on {realmUrl}.",
-  "Remove account": "Remove account"
+  "Remove account": "Remove account",
+  "This message was hidden because it is from a user you have muted. Long-press to view.": "This message was hidden because it is from a user you have muted. Long-press to view."
```
gnprice, line 235: Very much a nit: I like to add entries in the middle here -- next to logical neighbors if any, but in any case somewhere other than the end. The reason is that that lets the diff avoid touching lines it doesn't have to.

src/pm-conversations/PmConversationList.js
```diff
@@ -25,45 +27,48 @@ type Props = $ReadOnly<{|
 /**
  * A list describing all PM conversations.
  * */
-export default class PmConversationList extends PureComponent<Props> {
-  handleUserNarrow = (user: UserOrBot) => {
-    this.props.dispatch(doNarrow(pm1to1NarrowFromUser(user)));
+export default function PmConversationList(props: Props) {
+  const handleUserNarrow = (user: UserOrBot) => {
```
gnprice, line 31: Should use `useCallback` -- otherwise this is a new value each time this gets re-rendered, which in turn means each `UserItem` will get a new value and may have to re-render.

src/pm-conversations/PmConversationList.js
```diff
@@ -25,45 +27,48 @@ type Props = $ReadOnly<{|
...
+    props.dispatch(doNarrow(pm1to1NarrowFromUser(user)));
   };
 
-  handleGroupNarrow = (users: PmKeyUsers) => {
-    this.props.dispatch(doNarrow(pmNarrowFromUsers(users)));
+  const handleGroupNarrow = (users: PmKeyUsers) => {
+    props.dispatch(doNarrow(pmNarrowFromUsers(users)));
```
gnprice, line 36: Let's add a quick refactoring commit to have this component start using `useDispatch`, now that it's a functional component using Hooks. That will simplify the callers somewhat, as well as make this component more idiomatic again. As is, it winds up feeling sort of half-converted.

src/pm-conversations/PmConversationList.js
```diff
@@ -25,45 +27,48 @@ type Props = $ReadOnly<{|
...
+  return (
+    <FlatList
+      style={styles.list}
+      initialNumToRender={20}
+      data={conversations}
+      keyExtractor={item => item.key}
+      renderItem={({ item }) => {
+        const users = item.keyRecipients;
+        if (users.length === 1) {
+          if (mutedUsers.has(users[0].user_id)) {
```
gnprice, line 51: Let's factor out this `users[0].user_id`, since it's being used twice -- that helps make it real clear it's the same user ID both times.

Also the `users[0]` only makes sense because we've just entered a condition `users.length === 1` -- basically it's some manual pattern-matching -- and so it's good to keep the `users[0]` reference right up close next to that condition. (If for example this block were to use other parts of the user object, I'd want to say `const user = users[0];` right after the `if`, and go from there.)

src/unread/unreadSelectors.js
```diff
@@ -69,7 +69,11 @@ export const getUnreadByPms: Selector<{| [number]: number |}> = createSelector(
  */
 export const getUnreadPmsTotal: Selector<number> = createSelector(
   getUnreadPms,
-  unreadPms => unreadPms.reduce((total, pm) => total + pm.unread_message_ids.length, 0),
+  getMutedUsers,
+  (unreadPms, mutedUsers) =>
+    unreadPms
+      .filter(pm => !mutedUsers.has(pm.sender_id))
+      .reduce((total, pm) => total + pm.unread_message_ids.length, 0),
```
gnprice, line 76: This logic isn't necessary -- the server guarantees that messages from muted users will never be unread. See API docs:
https://chat.zulip.org/api/mute-user

src/users/UserList.js
```diff
@@ -22,48 +22,46 @@ type Props = $ReadOnly<{|
...
+          // $FlowFixMe[incompatible-type]
+          <SectionHeader text={section.key} />
+        )
+      }
+    />
+  );
 }
+
+UserList.defaultProps = {
+  selected: [],
```
gnprice, line 71: Don't use `defaultProps` for a functional component; instead use plain old JS destructuring:
```js
export default function UserList(props: Props) {
  const { foo, bar, selected = [] } = props;
```

The main reason is that it keeps the default visible in a natural spot at the top of the function.

This also means the type in `Props` gets to start being optional, reflecting the API actually exposed to the caller.

src/users/UserList.js
```diff
@@ -24,13 +26,16 @@ type Props = $ReadOnly<{|
...
 
-  if (shownUsers.length === 0) {
+  if (filteredUsers.length === 0) {
     return <SearchEmptyState text="No users found" />;
   }
 
-  const groupedUsers = groupUsersByStatus(shownUsers, presences);
+  const sortedUsers = sortUserList(filteredUsers, presences);
+
+  const groupedUsers = groupUsersByStatus(sortedUsers, presences);
```
gnprice: nit: no blank line between these -- they're closely related

gnprice (commented):
Thanks @WesleyAC ! This looks good -- comments below, all I think pretty small.

src/users/UserList.js
gnprice: More on this here:
https://github.com/zulip/zulip-mobile/pull/4440#discussion_r565684617

and I guess I should go add it to the style guide.

gnprice: https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#react-function-prop-defaults

static/translations/messages_en.json
chrisbobbe: I've sometimes thought that it'd be nice to enforce alphabetical order of the entries. It's impossible (or would take way too much effort, anyway) to be rigorous about grouping together similar items in all cases, and alphabetizing would eliminate cases where it's hard to decide where something goes.

I think I wouldn't want to alphabetize unless we found a sensible way to automate it, that we could also count on contributors using too. That'd take a bit of work, I think, and it's possibly not worth it.

src/unread/unreadSelectors.js
WesleyAC: Well, we've got a server bug, then :)

I think even once we fix that, it's prudent to keep this code in case of similar future problems.

src/users/UserList.js
WesleyAC: Fixed.

src/users/UserList.js
WesleyAC: Thanks for pointing that out! I had a feeling there was a better way of doing this, but didn't know what it was :)

src/unread/unreadSelectors.js
WesleyAC: Started a discussion on [#backend > muted users](https://chat.zulip.org/#narrow/stream/3-backend/topic/muted.20users) on CZO for the backend bug part of this.

src/pm-conversations/PmConversationList.js
WesleyAC: Done.

src/pm-conversations/PmConversationList.js
WesleyAC: Done.

src/pm-conversations/PmConversationList.js
WesleyAC: Done.

static/translations/messages_en.json
WesleyAC: We could just add new entries at the top, instead of the bottom :p

(I agree that making tooling or whatever for this is probably not worth it, and I moved it up a bit in this commit)

src/webview/html/messageAsHtml.js
WesleyAC: Done :sweat_smile:

src/webview/static/base.css
WesleyAC: Done! I sort of assumed that the whole diff would be larger than I wanted, but looking back at it it's actually pretty simple, which is nice.

src/webview/generateInboundEvents.js
```diff
@@ -51,6 +51,7 @@ const updateContent = (prevProps: Props, nextProps: Props): WebViewInboundEventC
       backgroundData: nextProps.backgroundData,
       narrow: nextProps.narrow,
       htmlPieceDescriptors: nextProps.htmlPieceDescriptorsForShownMessages,
+      _: nextProps._,
```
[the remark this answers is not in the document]
WesleyAC: Done.

WesleyAC:
@gnprice Thanks for the review! I made all the changes, should be good to go.

src/unread/unreadSelectors.js
gnprice: Interesting, and it looks like the outcome of that thread is that this is an existing mobile bug! We're not maintaining the unread state correctly.

Specifically, when we handle `EVENT_NEW_MESSAGE` (i.e. the Redux action we make from a `message` event) in each of the `unread` sub-reducers, we have logic like this:
```js
      if (message.sender_id === getOwnUserId(globalState)) {
        return state;
      }
```

and instead it should be checking something like `message.flags.includes('read')`. So we should fix that.

Then I don't think it makes sense to also have logic here specific to muted users. The job of this bit of code is to count up part of the unreads data; if we have future bugs where the unreads data is wrong, we should fix the bugs by making the unreads data correct.

(If for example we did have a server bug in the future that gave us incorrect unreads data, the place I'd want to put a workaround like this is at the input end, to make the data correct in Redux. One reason is that that makes sure it's fixed consistently for all the different places we might be using that data.)

src/webview/static/base.css
gnprice: Yeah :smile: 

There's actually one further squashing I had in mind relative to your new revision, which is:

`$ git uvr --shortstat`
3dc5af683 webview: Add initial support for hiding muted messages.
 4 files changed, 63 insertions(+), 2 deletions(-)
229ddf0d5 webview [nfc]: Pass GetText into messageAsHtml.
 4 files changed, 14 insertions(+), 3 deletions(-)
a1e07759e messageAsHtml: Translate user-muted message.
 2 files changed, 5 insertions(+), 5 deletions(-)

to squash those first and third changes (therefore coming after the "pass GetText in" change). The resulting change is
 5 files changed, 63 insertions(+), 2 deletions(-)

so actually not any bigger than the version that doesn't translate the text, and it cleans the history up further.

Everything else looks good, so I'll just merge now and make that change on the way.

gnprice:
Looks good, thanks -- merged!

I edited the series slightly from this:

3dc5af683 webview: Add initial support for hiding muted messages.
229ddf0d5 webview [nfc]: Pass GetText into messageAsHtml.
a1e07759e messageAsHtml: Translate user-muted message.
b7016f84f PmConversationList [nfc]: Convert to functional component.
8f67bab9b pm-conversations [nfc]: Get Dispatch via useDispatch.
a88f2f383 PmConversationList: Hide PMs from muted users.
9b72a8c0a autocomplete: Hide muted users.
3d21aedca unreads: Don't show PMs from muted users in unread count.
82197a53a UserList [nfc]: Convert to functional component.
571df7767 UserList: Hide muted users.

to this:

cd28e8854 webview [nfc]: Pass GetText into messageAsHtml.
5f816db8c webview: Hide messages from muted users, with option to show them.
a56e0b4ba PmConversationList [nfc]: Convert to functional component.
16ec0dc56 PmConversationList [nfc]: Get Dispatch via useDispatch.
e21138e9c PmConversationList: Hide PMs from muted users.
f688bfe0a autocomplete: Hide muted users.
b351de24e UserList [nfc]: Convert to functional component.
3682ee4d3 UserList: Hide muted users.

by squashing two commits as mentioned at https://github.com/zulip/zulip-mobile/pull/4699#discussion_r624397589 , and dropping the unreads change discussed at https://github.com/zulip/zulip-mobile/pull/4699#discussion_r623538882 .

One thing we'll still need before #4655 is complete is to fix the unread-state bug that we learned about in that latter thread.

WesleyAC:
Thanks! Followed up with #4710 for the unreads change.

PR #4700, opened 2021-04-27
Have Jest cover Android-only codepaths.

chrisbobbe:
Forking from @WesleyAC's comment at https://github.com/zulip/zulip-mobile/pull/4694#discussion_r619023107, which prompted me to realize that it's past time to have Jest include coverage of codepaths with both 'ios' and 'android' for `Platform.OS`.

Both platforms will be covered in CI, but only one (iOS) will run by default in local testing, for speed. That can be changed by passing `--platform both` to `tools/test`.

See discussion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60jest-expo.2Fios.60.20and.20.60jest-expo.2Fandroid.60/near/1169339).


jest/jestSetup.js
```diff
@@ -47,6 +47,12 @@ jest.mock('react-native', () => {
       resourceURL:
         'file:///private/var/containers/Bundle/Application/4DCD4D2B-F745-4C70-AD74-8E5F690CF593/ZulipMobile.app/',
     };
+  } else if (ReactNative.Platform.OS === 'android') {
+    ReactNative.NativeModules.TextCompressionModule = {
+      header: 'z|zlib base64|',
+      compress: async (input: string) => `pretend-this-is-compressed:${input}`,
+      decompress: async (input: string) => input.replace('pretend-this-is-compressed:', ''),
```
gnprice, line 54: The `header` property and the mock compress/decompress behavior here don't match -- the header is supposed to be the prefix of anything the `compress` method returns, and in `ZulipAsyncStorage.js` we take care not to pass to `decompress` anything that doesn't have the header as a prefix.

A good solution would be to make the header something like `z|mock|` (the `z|…|` syntax is part of the overall protocol as implemented by `ZulipAsyncStorage`), and have `compress` return `` `${header}${input}` ``.

docs/howto/testing.md
```diff
@@ -18,6 +18,16 @@ To write a test, place a Javascript file with the `-test.js` suffix in the
...
+one with an iOS preset, and one with an Android preset.
+`tools/test jest` will run one or the other, or both, depending on
+what you pass for `--platform`.
+
+If it makes sense for a test file to be dedicated to just one
+platform, and that's unlikely to change, you can instead suffix it
+with `-test.ios.js` or `-test.android.js`, and it will be ignored when
+the other platform's project runs.
 
 ### Test style guide
```
gnprice, line 32: nit: two blank lines between sections (here and below)

This helps a bit in visually parsing the file when reading it directly.

docs/howto/testing.md
```diff
@@ -18,6 +18,16 @@ To write a test, place a Javascript file with the `-test.js` suffix in the
...
+We've configured Jest to run two different
+["projects"](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig):
+one with an iOS preset, and one with an Android preset.
+`tools/test jest` will run one or the other, or both, depending on
+what you pass for `--platform`.
+
+If it makes sense for a test file to be dedicated to just one
+platform, and that's unlikely to change, you can instead suffix it
+with `-test.ios.js` or `-test.android.js`, and it will be ignored when
+the other platform's project runs.
```
gnprice, line 30: This feels more detailed about the Jest specifics than is needed here -- those details can be left to the code that controls them. (E.g. I think the concept of a Jest "project" can be left to `jest.config.js` and to that `case` block in `tools/test` that sets `--selectedProjects`.)

For this doc the ideal thing is to be more direct about what happens by default, and what you might want to do instead (and when.)

gnprice (commented):
Thanks @chrisbobbe ! This will be great. Comments below.

chrisbobbe:
Thanks for the quick review, @gnprice! Reading now.

src/boot/__tests__/replaceRevive-test.js
```diff
@@ -44,10 +44,14 @@ const data = {
...
+
+Object.keys(data).forEach(key => {
+  stringified[key] = stringify(data[key]);
+});
+
 describe('Stringify', () => {
   Object.keys(data).forEach(key => {
     it(key, () => {
-      stringified[key] = stringify(data[key]);
       expect(stringified[key]).toMatchSnapshot();
```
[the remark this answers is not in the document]
chrisbobbe: Ah—the bug that's fixed in this commit is that describe block labeled "Parse" did nothing. I'll fix the commit message (including the summary) to be clearer about that.

tools/test
```diff
@@ -117,6 +117,10 @@ run_native() {
     )
 }
 
+run_native() {
+    run_native_android || return
```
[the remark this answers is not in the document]
chrisbobbe: Huh, interesting! It looks like something strange is happening.

I added `|| return` because I found that `run_native` wasn't exiting when `run_native_android` failed.

I've tried a few things just now, and I think I found the culprit. From the Bash manual on `set -e` (emphasis mine),

> The shell does not exit if the command that fails is part of the command list immediately following a `while` or `until` keyword, part of the test in an `if` statement, _**part of any command executed in a `&&` or `||` list except the command following the final `&&` or `||`**_, any command in a pipeline but the last, or if the command’s return status is being inverted with `!`.

If I make this change just after the `for suite in "${suites[@]}"` loop, it works as expected with the `|| return` removed:

```diff
-    esac || failed+=($suite)
+    esac
```

🤯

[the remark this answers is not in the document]
chrisbobbe: Does that sound right, and if so, how would you like to proceed?

tools/test
```diff
@@ -117,8 +134,32 @@ run_native_android() {
...
 run_native() {
-    run_native_android || return
+    if [[ $platform == android || $platform == both || $platform == sloppy ]]; then
+        echo 'Running Android native tests...';
+        run_native_android || return
+    else
+        echo -e $'\e[31mSkipping native Android tests. Use --platform to control this.\e[0m';
+    fi
+
+    if [[ $platform == ios || $platform == both ]]; then
```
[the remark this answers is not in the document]
chrisbobbe: Ah OK, sure; I think I'd been thinking of this as another place (like where we choose iOS for Jest) where we'd sloppily pick one platform for the sake of speed.

But the iOS and Android native tests are quite independent of each other. Unlike with running the 'ios' and 'android' projects in Jest, failure/success on one side is quite unlikely to be a signal for something going wrong on the other side.

tools/test
```diff
@@ -17,6 +17,14 @@ Options:
               with --full.
   --fix       Fix issues found, where possible.
   --full      Run on all files, not only changed files.
+  --platform [ios|android|both|sloppy]
+              Target iOS-specific or Android-specific code, or both.
+              "sloppy", the default, chooses one platform arbitrarily
+              for some test suites, so it doesn't give full coverage.
+              It exists just to decrease the running time for local
+              testing, given that much of our code isn't
+              platform-specific.
```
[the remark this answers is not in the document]
chrisbobbe: > Maybe I'll just make a pass over it after merge.

Sure, that'd be helpful, thanks. 🙂

> where it can tell the difference

I'm not sure I follow the reasoning for choosing to run on both platforms vs. just one. Is it that in one case (Jest), we're likely to see the same successes and failures on both platforms?

tools/test
```diff
@@ -117,8 +134,32 @@ run_native_android() {
...
+run_native_ios() {
+    :
+    # TODO: At least make an iOS build when $platform is `ios` or
+    # `both`.
+}
+
 run_native() {
-    run_native_android || return
+    if [[ $platform == android || $platform == both || $platform == sloppy ]]; then
+        echo 'Running Android native tests...';
```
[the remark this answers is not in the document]
chrisbobbe: > Maybe I'll try to experiment with something later.

Cool, that'd be helpful, thanks! I'm also happy to act on some sketched-out ideas, but it might be more efficient for you to implement what you have in mind.

tools/test
```diff
@@ -117,8 +134,32 @@ run_native_android() {
...
+        run_native_android || return
+    else
+        echo -e $'\e[31mSkipping native Android tests. Use --platform to control this.\e[0m';
```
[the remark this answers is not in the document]
chrisbobbe: Ah, OK; yeah, this is the case after I've made the change to have the default (`sloppy`) do both Android and iOS for the `native` suite.

tools/test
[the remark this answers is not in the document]
gnprice: Urgh, right. Thanks for catching that. This is the main gotcha in `set -e`, where it doesn't do everything you'd think it does (because any reasonable version of it absolutely would.) :exploding_head: is right on.

This also explains why I wrote a similar `|| return` in one spot in the existing code, at `check_node || return`. When I wrote this comment I was wondering about that...

I guess the short of it is that inside all these functions, we effectively *don't* have the benefit of `set -e`. And we can't even re-enable it from inside each function, as this bit later in that same manual item explains:
> If a compound command or shell function sets `-e` while executing in a context where `-e` is ignored, that setting will not have any effect until the compound command or the command containing the function call completes.

I think the right immediate fix is to use `|| return` just as this branch does. Basically we need that wherever we have a nontrivial command that isn't the final command of its function. I think the existing code is all OK in this respect, thanks to that `|| return` in one place we did need it.

That and maybe a comment on the `set -e` at the top, warning that it doesn't (and sadly can't) actually apply to most of the code.

Then... if we want to get back `set -e` behavior inside these functions, while still being able to "catch" its propagation at the outer loop so that we can go on to run the other suites, I think basically what's required is to fork a subshell (with the `( … )` syntax -- like we already do inside `run_android` in order to use `cd` and contain its effect.) I think that's the real point of what that quoted bit of docs is saying: `set -e` stops having any effect if it isn't going to be able to go all the way to exit the whole shell. That's a bit of overhead, but only like a millisecond each:
```
$ time bash -c 'for i in {1..100}; do ( : ); done'

time: 0.099s wall (0.065s u, 0.041s s)
```
so not a problem at all. (That fast startup time is one of Bash's handful of great strengths, compared to almost any other language that doesn't require a compile step.)

tools/test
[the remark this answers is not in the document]
gnprice: By "it" I meant tools/test (but didn't say that clearly).

So for the Android-native and future iOS-native builds and tests, we go ahead and run them in the default configuration (the one optimized for interactive use), because we know we can do so without duplicating anything.

Also in a future where we had some clever way of identifying which Jest tests actually depended meaningfully on the value of `Platform.OS` (or whatever else might vary between those two Jest "platform" settings), I'd want the default behavior to be to run those tests both ways, while continuing to run the other Jest tests only once. (And if we were confident that that clever way didn't miss anything -- maybe had false positives, but no false negatives -- then I'd want to do that even in CI.)

But as it stands, we know that the great majority of the Jest tests don't depend on the platform, and we can't in any automated way tell which are the exceptions. So by default we only run the tests once, and pick arbitrarily which platform setting to use.

tools/test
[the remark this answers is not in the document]
gnprice: Right, exactly.

tools/test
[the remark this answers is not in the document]
chrisbobbe: OK, sounds good! Glad to see the 🤯 feeling is shared! 😅 I'll put a comment on the `set -e`.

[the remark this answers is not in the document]
chrisbobbe: I'm not sure what's the best strategy for forking a subshell where appropriate; we'd of course like to leave a good signpost if future added `run_foo` functions have to remember to do something.

jest/jestSetup.js
chrisbobbe: Something like this?

```js
    const header = 'z|mock|';
    ReactNative.NativeModules.TextCompressionModule = {
      header,
      compress: async (input: string) => `${header}${input}`,
      decompress: async (input: string) => input.replace(header, ''),
    };
```

I'm looking at `decompress`, in particular. Possibly we don't need to mock that yet, though, since we don't have any tests that use it?

tools/test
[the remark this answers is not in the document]
chrisbobbe: Sounds good. I've revised the description with this in mind, but I'm not sure I've found the best/most concise wording—please feel free to discard and start fresh if you like. 🙂

chrisbobbe:
Thanks again for the review! Revision pushed. I've also replied inline to some of your comments, and you've already responded to some.

tools/test
```diff
@@ -1,4 +1,21 @@
...
+# because our logic to catch a failing suite, add it to a list, and
+# continue to the other suites, puts a bunch of commands in a list
+# before a `||`. That puts them in a special case, as the Bash manual
+# says:
+#
+# """
+# The shell does not exit if the command that fails is [...] part of
+# any command executed in a `&&` or `||` list except the command
+# following the final `&&` or `||` [...].
+# """
```
WesleyAC, line 14: nit: I would use:

```
> blockquote style
```

rather than the triple quotes, since I think that's clearer, and uses less vertical space.

tools/test
```diff
@@ -1,4 +1,21 @@
 #!/usr/bin/env bash
+
+# Careful! This doesn't do everything you'd think it does. In fact, we
```
WesleyAC, line 3: I would replace `This` here with `'set -e'`, since otherwise it's unclear what this comment is referring to until you read the entire thing.

tools/test
```diff
@@ -35,6 +35,16 @@ Options:
...
+  --platform <ios|android|both|sloppy>
+              Run tests as if on iOS, or Android, or both. The default,
+              "sloppy", takes a shortcut in some suites, so the tests
+              run faster but full coverage of both platforms isn't
+              guaranteed. In particular, while the vast majority of
+              our Jest tests don't depend meaningfully on the platform,
+              we don't yet have a clever way to identify the ones that
+              do. So, "sloppy" just picks one platform to simulate
+              (iOS). Don't use "sloppy" if you want coverage of the
+              other platform.
```
WesleyAC, line 47: I think this could be clearer:

```
              Run tests as if on iOS, or Android, or both. The default,
              "sloppy", takes a shortcut in some suites, so the tests
              run faster but full coverage of both platforms isn't
	      guaranteed. Specifically, "sloppy" will run both Android and iOS
	      native tests, but will run Jest tests on only one platform (iOS).
	      This is usually fine, because the vast majority of our Jest tests
	      don't depend meaningfully on the platform.
```

tools/test
```diff
@@ -179,6 +214,17 @@ run_jest() {
...
 
+    case "$platform" in
+        ios) jest_args+=( --selectProjects ios );;
+        android) jest_args+=( --selectProjects android );;
+        both) jest_args+=( --selectProjects ios android );;
+
+        # This is where `sloppy` is sloppy: we just choose iOS, so the
+        # tests will run faster, but at the expense of a (relatively
+        # small) loss of coverage.
+        sloppy) jest_args+=( --selectProjects ios );;
```
WesleyAC, line 228: A possibly dumb idea — we could pick whether it runs as iOS or Android randomly. The idea there being that if it's only failing on one, picking randomly would cause you to notice sooner during local development, rather than it only being caught in CI. And while that makes it flaky, it's a pretty reasonable type of flake, since it'll always be catching a real test failure.

tools/test
```diff
@@ -82,6 +84,7 @@ while (( $# )); do
             esac
             shift
             ;;
+        --all) files=all; platform=both; shift;;
```
WesleyAC, line 87: I think it someone does `tools/test --all --platform=android` or similar, the `--platform` will overwrite the `--all`, is that right?

Probably we should just fail if someone passes both `--all` and `--platform` or `--diff`?

jest/jestSetup.js
```diff
@@ -45,6 +45,13 @@ jest.mock('react-native', () => {
       resourceURL:
         'file:///private/var/containers/Bundle/Application/4DCD4D2B-F745-4C70-AD74-8E5F690CF593/ZulipMobile.app/',
     };
+  } else if (ReactNative.Platform.OS === 'android') {
+    const header = 'z|mock|';
+    ReactNative.NativeModules.TextCompressionModule = {
+      header,
+      compress: async (input: string) => `${header}${input}`,
+      decompress: async (input: string) => input.replace(header, ''),
```
WesleyAC, line 53: This is fine as is, but I'd be a bit more comfortable with something that mangles the whole string. Possibly base64encoding it, in addition to adding the header? (I'm not sure what JS runtime our jest tests execute in, so I'm not sure if we have a reasonable base64 function easily available to us, but if we do, that seems like a good option).

src/boot/__tests__/ZulipAsyncStorage-test.js
```diff
@@ -0,0 +1,33 @@
...
+// and Android, which might make them a bit harder to understand. But
+// most of the logic is appropriately in common between the platforms,
+// so we haven't split the different logic up into different tests.
+//
+// We may in the future decide to add (or replace this file with) a
+// `ZulipAsyncStorage-test.ios.js` and
+// `ZulipAsyncStorage-test.android.js`, to target entire test suites
+// to Android and iOS.
+describe('ZulipAsyncStorage', () => {
+  describe('setItem', () => {
```
WesleyAC, line 17: Can we test `getItem` as well?

WesleyAC (commented):
Thanks for doing this! I'm very happy that we'll be getting more coverage here. Just a few comments :)

tools/test
chrisbobbe: Interesting; I kind of like this idea. @gnprice, what do you think?

tools/test
chrisbobbe: I like it!

tools/test
chrisbobbe: Makes sense. I also wondered if we should write the first line of the usage string differently, to reflect that:

```
usage: tools/test [--all-files | --diff COMMIT] [--platform <ios|android|both|sloppy>] [--all] [SUITES]
```

(I.e., from reading that it looks like it's perfectly fine to do `--all` and `--platform` or `--diff`.)

But I'm not sure how best to do that while preserving readability. Maybe it's fine as-is.

jest/jestSetup.js
chrisbobbe: Sounds good; there is such a thing, I think. I can use `Buffer.from`.

The Jest tests execute in Node, and we currently recommend the latest 12.x and use that in CI (see f83ccc6cc6ccfc3cb843428006f231b1d057346d for what this means more concretely). #M4263 is open to upgrade to 14.

tools/test
chrisbobbe: Hmm, looking again, I'm not sure what's the best way to enforce these rules on what gets passed. (I see that we're already not enforcing that only one of `--diff` and `--all-files` gets passed.)

I'm not super familiar with Bash, yet. It seems plausible to me that the best way to start enforcing these things will be after we've started using `getopt` for something different; Greg suggested that [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60jest-expo.2Fios.60.20and.20.60jest-expo.2Fandroid.60/near/1169550). Do you think that's right?

src/boot/__tests__/replaceRevive-test.js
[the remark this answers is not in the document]
gnprice: Very clear, thanks!

tools/test
gnprice: For `--all-files` and `--diff`, I think the behavior is actually fine: they're each setting the same option to different values, and the last one wins. That's a common CLI convention, and it can come in handy when you want to have some default command line (e.g. an alias in your interactive shell) and sometimes pass additional arguments, and sometimes override an option that was already in the default version of the command.

I think the way `--all` interacts with those two and `--platform` is also fine for the same reason. It means that, for example, you could have an alias that expands to `tools/test --all`, and effectively it'd be `tools/test` but with your custom choice of a different default -- you'd still be able to override it to get other behavior when you wanted.

(For the real thorough version of that, we'd also add a flag that means the default `files=branch` behavior as opposed to `--all-files` or `--diff`.)

If we did want behavior where some options are mutually exclusive, that's actually orthogonal to getopt; the way you'd implement it if we were using getopt is basically the same as how you'd implement it now.

tools/test
gnprice: Hmm, maybe. I like the idea of causing you to notice sooner. I think this would definitely best be done as a separate change.

tools/test
```diff
@@ -35,6 +35,16 @@ Options:
   --fix       Fix issues found, where possible.
   --all-files
               Run on all files, not only changed files.
+  --platform <ios|android|both|sloppy>
```
gnprice: nit: Let's keep `--all-files` next to `--diff`, because they're closely related -- in particular because they're two possible values of the same underlying option.

docs/howto/testing.md
```diff
@@ -4,7 +4,7 @@ Run `tools/test` to run all our test suites.  This command is
 typically quite fast (5-10s, sometimes less), because it only rechecks
 tests related to the files you've changed.
 
-You can run all our tests with `tools/test --full`.
+You can run all our tests with `tools/test --all-files --platform both`.
```
gnprice, line 7: should say the nice short `--all` in the end

gnprice (commented):
This looks great, thanks!

Other than the two small things below, and the things in threads above where I feel like I need to explore a bit in order to have a concrete suggestion and so I may try for myself later, I have just one other comment:

I think the series can be simplified somewhat by taking the three commits where `--full` gets renamed, `--platform` introduced, and we have both `--all-files` and `--all`, and rearranging to:
* One commit renames `--full` directly to `--all`. After all, in the end that's what we want in all the known use cases of the old `--full`. At this stage it means the same thing as `--all-files` would, because there's no `--platform` yet.
* Then in the commit that adds `--platform`, `--all` starts including the meaning of `--platform both`, and we introduce `--all-files` for the behavior that's only about the `files` variable.

In particular this means the places we mention `--full` change only in the first of those commits, instead of all three of the existing sequence.

src/boot/__tests__/ZulipAsyncStorage-test.js
chrisbobbe: #4709

tools/test
```diff
@@ -1,11 +1,26 @@
...
+# Careful! `set -e` doesn't do everything you'd think it does. In
+# fact, we don't get its benefit in any of the `run_foo` functions.
+# That's because our logic to catch a failing suite, add it to a list,
+# and continue to the other suites, puts a bunch of commands in a list
+# before a `||`. That puts them in a special case, as the Bash manual
+# says:
+#
+# > The shell does not exit if the command that fails is [...] part of
+# any command executed in a `&&` or `||` list except the command
+# following the final `&&` or `||` [...].
```
gnprice, line 12: nit: using `>` on each line makes it clearer this is all part of the quote

tools/test
```diff
@@ -39,21 +66,31 @@ EOF
 coverage=
 files=branch
 fix=
+platform=sloppy
 suites=()
```
gnprice, line 70: tiny nit: clearest if these are kept in the same order as the cases that operate on them, below

gnprice (commented):
Thanks @chrisbobbe ! Looks great -- tiny comments below.

I'm about to merge, so I'll just fix these along the way.

chrisbobbe:
Thanks for the reviews!

> I'm about to merge, so I'll just fix these along the way.

I've gone ahead and made the two targeted fixes for the nits you pointed out (and nothing else), and pushed that to this branch, just in case that's helpful. 😛

gnprice:
Thanks @chrisbobbe! Sorry I dropped this for a bit. I actually thought I'd merged it an hour or two before you eventually did (i.e., late that morning here); must be that that failed with a "there are upstream changes, you need to rebase first" and I didn't notice.

Here's the version I thought I'd pushed:
a3e4c0fff replaceRevive tests: Fix bug where "Parse" describe block did nothing.
5a90e13fc replaceRevive tests: Stop using snapshot tests.
cf6f72535 deps: Use Jest 26, the latest, with a `resolutions` hack.
2a53ad6a4 jest: Use `jest-expo/ios` preset.
80a68beee jest: Expand coverage to include Android-only codepaths.
7f93fc2bc jestSetup [nfc]: Start type-checking.
fe6829047 jestSetup [nfc]: Remove blank line between third-party imports.
99ea911d0 tools/test [nfc]: Write down a big gotcha with `set -e`.
662326360 tools/test [nfc]: Explain the `set -e` thing a bit more broadly.
9009158ba yarn scripts: Remove several aliases for simple commands.
4d628c79d tools/test: Rename `--full` to `--all`.
716f8a66e tools/test: Rename `android` suite to `native`.
6efa08a69 tools/test [nfc]: Bring Android-native test code to its own function.
0ee8fd667 tools/test: Add `--platform` option.
03b799235 tools/test [nfc]: Reorganize usage message.
2d632b5dd tools/test [nfc]: Simplify description of default list of suites.
b43928a20 docs/testing: Explain more on Jest platform switch; shorten in howto.

I'll go ahead and push the additional commits that were in that version -- all comment- and doc-only changes. Other than those, the differences were just
* I had the changes in an order that felt a bit more logical, e.g. grouping the Jest-setup changes together.
* One clarified summary line:

      -    tools/test [nfc]: Write down a big gotcha.
      +    tools/test [nfc]: Write down a big gotcha with `set -e`.

  I.e. the gotcha is in `set -e`, not in using `tools/test`.

So indeed, merging this version as it was was just fine. :slightly_smiling_face: 

Then after those I have some draft changes which split `native` back up into `android` plus a future/placeholder `ios` (along the lines of https://github.com/zulip/zulip-mobile/pull/4700#discussion_r621487064 ), and then go on to experiment with some refactorings to help manage the should-skip logic for each of these suites while communicating clearly to the user what's happening. I'll see about finishing those up into a PR.

gnprice:
OK, pushed those additional commits:
f6985b057 tools/test [nfc]: Explain the `set -e` thing a bit more broadly.
8e4c0961d tools/test [nfc]: Reorganize usage message.
15cd56bb7 tools/test [nfc]: Simplify description of default list of suites.
47c83c432 docs/testing: Explain more on Jest platform switch; shorten in howto.

chrisbobbe:
Great, thanks!

PR #4703, opened 2021-04-28
ios: Fix "Brand" color set, which was slightly wrong.

chrisbobbe:
The thing to match is rgb(100, 146, 253.5), as it says in the
comment on BRAND_COLOR in the JS. So, convert that to hex and use
the result, instead of converting from RGB to HSL and back to RGB as
we'd been doing before.

See discussion at
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/colors/near/1170829.


WesleyAC (approved):
Thanks!

chrisbobbe:
Thanks for the review, @WesleyAC! Pushed.

PR #4704, opened 2021-04-29
Add basic support for polls (viewing and voting)

WesleyAC:
This PR adds basic support for viewing and voting on polls.

Some notes:

* It's pretty ugly, largely because the underlying widget/submessage/polls code is ugly. I'm open to changing things where it's easy, but I don't want to go down the rabbit hole of cleaning up the fundamental way that polls work.
* I've implemented local echo, although I don't know how I feel about it — if the `sendSubmessage` fails, we wouldn't communicate that to the user. Happy to have thoughts on what the right approach is here. One thought is that maybe we want to change the background color, but not the number? That would be pretty ugly, but could communicate what's going on without building a bunch more UI. We could also add a loading spinner like we do on outbox messages.
* I have a `$FlowFixMe` for adding types to `PollData`. I'm would like to do this, although probably not complete types: there are some fundamental limits to how much we can get from typing here, since there's so much dynamism in how widgets work.
* I made the poll buttons 36x36px, since 48x48 seemed really large. I understand that we usually aim for 48x48px touch targets, but I think that for instance, reaction emoji are good precedent here for not doing that.

screenshot

### DIFF
--- a/src/webview/html/messageAsHtml.js
+++ b/src/webview/html/messageAsHtml.js
@@ -1,6 +1,10 @@
 /* @flow strict-local */
 import { PixelRatio } from 'react-native';
+import invariant from 'invariant';
 import distanceInWordsToNow from 'date-fns/distance_in_words_to_now';
+// $FlowFixMe[untyped-import]
+import { PollData } from '@zulip/shared/js/poll_data';
+
 import template from './template';
 import type {
   AggregatedReaction,
@@ -10,14 +14,17 @@ import type {
   MessageLike,
   Outbox,
   Reaction,
+  SubmessageData,
   ImageEmojiType,
   UserId,
+  WidgetData,
 } from '../../types';
 import type { BackgroundData } from '../MessageList';
 import { shortTime } from '../../utils/date';
 import aggregateReactions from '../../reactions/aggregateReactions';
 import { codeToEmojiMap } from '../../emoji/data';
 import processAlertWords from './processAlertWords';
+import * as logging from '../../utils/logging';
 
 const messageTagsAsHtml = (isStarred: boolean, timeEdited: number | void): string => {
   const pieces = [];
@@ -72,13 +79,105 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
 `;
 };
 
-const widgetBody = (message: Message | Outbox) => template`
+/**
+ * Render the body of a message that has submessages.
+ *
+ * Must not be called on a message without any submessages.
+ */
+const widgetBody = (message: Message, ownUserId: UserId) => {
+  invariant(
+    message.submessages !== undefined && message.submessages.length > 0,
+    'should have submessages',
+  );
+
+  const widgetSubmessages: Array<{
+    sender_id: number,
+    content: SubmessageData,
+    ...
+  }> = message.submessages
+    .filter(submessage => submessage.msg_type === 'widget')
+    .sort((m1, m2) => m1.id - m2.id)
+    .map(submessage => ({
+      sender_id: submessage.sender_id,
+      content: JSON.parse(submessage.content),
+    }));
+
+  const errorMessage = template`
 $!${message.content}
 <div class="special-message"
  ><p>Interactive message</p
  ><p>To use, open on web or desktop</p
 ></div>
 `;
+
+  const pollWidget = widgetSubmessages.shift();
+  if (!pollWidget || !pollWidget.content) {
+    return errorMessage;
+  }
+
+  /* $FlowFixMe[incompatible-type]: The first widget submessage should be
+       a `WidgetData`; see jsdoc on `SubmessageData`. */
+  const pollWidgetContent: WidgetData = pollWidget.content;
+
+  if (pollWidgetContent.widget_type !== 'poll') {
+    return errorMessage;
+  }
+
+  if (pollWidgetContent.extra_data == null) {
+    // We don't expect this to happen in general, but there are some malformed
+    // messages lying around that will trigger this [1]. The code here is slightly
+    // different the webapp code, but mostly because the current webapp
+    // behaviour seems accidental: an error is printed to the console, and the
+    // code that is written to handle the situation is never reached.  Instead
+    // of doing that, we've opted to catch this case here, and print out the
+    // message (which matches the behaviour of the webapp, minus the console
+    // error, although it gets to that behaviour in a different way). The bug
+    // tracking fixing this on the webapp side is zulip/zulip#19145.
+    // [1]: https://chat.zulip.org/#narrow/streams/public/near/582872
+    return template`$!${message.content}`;
+  }
+
+  const pollData = new PollData({
+    message_sender_id: message.sender_id,
+    current_user_id: ownUserId,
+    is_my_poll: message.sender_id === ownUserId,
+    question: pollWidgetContent.extra_data.question ?? '',
+    options: pollWidgetContent.extra_data.options ?? [],
+    // TODO: Implement this.
+    comma_separated_names: () => '',
+    report_error_function: (msg: string) => {
+      logging.error(msg);
+    },
+  });
+
+  for (const pollEvent of widgetSubmessages) {
+    pollData.handle_event(pollEvent.sender_id, pollEvent.content);
+  }
+
+  const parsedPollData = pollData.get_widget_data();
+
+  return template`
+<div class="poll-widget">
+  <p class="poll-question">${parsedPollData.question}</p>
+  <ul>
+    $!${parsedPollData.options
+      .map(
+        option =>
+          template`
+        <li>
+          <button
+            class="poll-vote"
+            data-voted="${option.current_user_vote}"
+            data-key="${option.key}"
+          >${option.count}</button>
+          <span class="poll-option">${option.option}</span>
+        </li>`,
+      )
+      .join('')}
+  </ul>
+</div>
+  `;
+};
 
 export const flagsStateToStringList = (flags: FlagsState, id: number): string[] =>
   Object.keys(flags).filter(key => flags[key][id]);
@@ -112,7 +211,7 @@ export default (
 `;
   const bodyHtml =
     message.submessages && message.submessages.length > 0
-      ? widgetBody(message)
+      ? widgetBody(message, backgroundData.ownUser.user_id)
       : messageBody(backgroundData, message);
 
   if (isBrief) {
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -918,6 +918,27 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.matches('.poll-vote')) {
+    const messageElement = target.closest('.message');
+    if (!messageElement) {
+      throw new Error('Message element not found');
+    }
+    // This duplicates some logic from PollData.handle.vote.outbound in
+    // @zulip/shared/js/poll_data.js, but it's much simpler to just duplicate
+    // it than it is to thread a callback all the way over here.
+    const current_vote = requireAttribute(target, 'data-voted') === 'true';
+    const vote = current_vote ? -1 : 1;
+    sendMessage({
+      type: 'vote',
+      messageId: requireNumericAttribute(messageElement, 'data-msg-id'),
+      key: requireAttribute(target, 'data-key'),
+      vote,
+    });
+    target.setAttribute('data-voted', (!current_vote).toString());
+    target.innerText = (parseInt(target.innerText, 10) + vote).toString();
+    return;
+  }
+
   if (target.matches('time')) {
     const originalText = requireAttribute(target, 'original-text');
     sendMessage({
--- a/src/webview/static/base.css
+++ b/src/webview/static/base.css
@@ -696,3 +696,50 @@ h1, h2, h3, h4, h5, h6 {
 .spoiler-block .spoiler-arrow.spoiler-button-open::after {
   transform: rotate(90deg) translate(10px, 0);
 }
+
+/* Poll styling */
+
+.poll-widget {
+  border: hsl(0, 0%, 50%) 1px solid;
+  padding: 2px 8px 2px 10px;
+  border-radius: 10px;
+}
+
+.poll-question {
+  font-size: 1.2rem;
+  margin-bottom: 8px;
+  border-bottom: 1px solid hsla(0, 0%, 60%, 0.2);
+}
+
+.poll-widget > ul {
+  padding: 0px;
+}
+
+.poll-widget > ul > li {
+  list-style: none;
+  margin-bottom: 4px;
+  display: flex;
+  align-items: center;
+}
+
+.poll-vote {
+  background-color: hsla(0, 0%, 0%, 0);
+  border: 1.5px solid hsl(0, 0%, 50%);
+  border-radius: 8px;
+  height: 36px;
+  min-width: 36px;
+  padding: 1px 8px;
+  font-weight: bold;
+  font-size: 18px;
+  flex-shrink: 0;
+}
+
+.poll-vote[data-voted="true"] {
+  border: 1.5px solid hsl(222, 99%, 69%);
+  background-color: hsla(222, 99%, 69%, 25%);
+}
+
+.poll-option {
+  margin-left: 8px;
+  width: 100%;
+}
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@react-navigation/stack": "^5.9.3",
     "@sentry/react-native": "^2.4.3",
     "@unimodules/core": "~5.3.0",
-    "@zulip/shared": "^0.0.5",
+    "@zulip/shared": "^0.0.6",
     "base-64": "^0.1.0",
     "blueimp-md5": "^2.10.0",
     "color": "^3.0.0",
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -40,6 +40,7 @@ import toggleMobilePushSettings from './settings/toggleMobilePushSettings';
 import createStream from './streams/createStream';
 import getStreams from './streams/getStreams';
 import updateStream from './streams/updateStream';
+import sendSubmessage from './submessages/sendSubmessage';
 import getSubscriptions from './subscriptions/getSubscriptions';
 import subscriptionAdd from './subscriptions/subscriptionAdd';
 import subscriptionRemove from './subscriptions/subscriptionRemove';
@@ -86,6 +87,7 @@ export {
   createStream,
   getStreams,
   updateStream,
+  sendSubmessage,
   getSubscriptions,
   setTopicMute,
   subscriptionAdd,
--- a/src/api/modelTypes.js
+++ b/src/api/modelTypes.js
@@ -457,6 +457,47 @@ export type PmRecipientUser = $ReadOnly<{|
 |}>;
 
 /**
+ * The data encoded in a submessage to make the message a widget.
+ *
+ * Note that future server versions might introduce new types of widgets, so
+ * `widget_type` could be a value not included here.  But when it is one of
+ * these values, the rest of the object will follow this type.
+ */
+// Ideally we'd be able to express both the known and the unknown widget
+// types: we'd have another branch of this union which looked like
+//   | {| +widget_type: (string *other than* those above), +extra_data?: { ... } |}
+// But there doesn't seem to be a way to express that in Flow.
+export type WidgetData =
+  | {|
+      +widget_type: 'poll',
+      +extra_data?: {| +question?: string, +options?: $ReadOnlyArray<string> |},
+    |}
+  // We can write these down more specifically when we implement these widgets.
+  | {| +widget_type: 'todo', +extra_data?: { ... } |}
+  | {| +widget_type: 'zform', +extra_data?: { ... } |};
+
+/**
+ * The data encoded in a submessage that acts on a widget.
+ *
+ * The interpretation of this data, including the meaning of the `type`
+ * field, is specific to each widget type.
+ *
+ * We delegate actually processing these to shared code, so we don't specify
+ * the details further.
+ */
+export type WidgetEventData = { +type: string, ... };
+
+/**
+ * The data encoded in a `Submessage`.
+ *
+ * For widgets (the only existing use of submessages), the submessages array
+ * consists of:
+ *  * One submessage with `WidgetData`; then
+ *  * Zero or more submessages with `WidgetEventData`.
+ */
+export type SubmessageData = WidgetData | WidgetEventData;
+
+/**
  * Submessages are items containing extra data that can be added to a
  * message. Despite what their name might suggest, they are not a subtype
  * of the `Message` type, nor do they share almost any fields with it.
@@ -473,7 +514,9 @@ export type Submessage = $ReadOnly<{|
   message_id: number,
   sender_id: UserId,
   msg_type: 'widget', // only this type is currently available
-  content: string, // JSON string
+
+  /** A `SubmessageData` object, JSON-encoded. */
+  content: string,
 |}>;
 
 /**
--- a/src/api/submessages/sendSubmessage.js
+++ b/src/api/submessages/sendSubmessage.js
@@ -1,0 +1,13 @@
+/* @flow strict-local */
+
+import type { ApiResponse, Auth } from '../transportTypes';
+import { apiPost } from '../apiFetch';
+
+/** See https://zulip.readthedocs.io/en/latest/subsystems/widgets.html#poll-todo-lists-and-games */
+// `msg_type` only exists as widget at the moment, see #3205.
+export default async (auth: Auth, messageId: number, content: string): Promise<ApiResponse> =>
+  apiPost(auth, 'submessage', {
+    message_id: messageId,
+    msg_type: 'widget',
+    content,
+  });
--- a/src/webview/css/cssNight.js
+++ b/src/webview/css/cssNight.js
@@ -5,6 +5,9 @@ body {
   color: hsl(210, 11%, 85%);
   background: hsl(212, 28%, 18%);
 }
+.poll-vote {
+  color: hsl(210, 11%, 85%);
+}
 .topic-header {
   background: hsl(212, 13%, 38%);
 }
--- a/src/webview/handleOutboundEvents.js
+++ b/src/webview/handleOutboundEvents.js
@@ -132,6 +132,13 @@ type WebViewOutboundEventTimeDetails = {|
   originalText: string,
 |};
 
+type WebViewOutboundEventVote = {|
+  type: 'vote',
+  messageId: number,
+  key: string,
+  vote: number,
+|};
+
 export type WebViewOutboundEvent =
   | WebViewOutboundEventReady
   | WebViewOutboundEventScroll
@@ -146,7 +153,8 @@ export type WebViewOutboundEvent =
   | WebViewOutboundEventWarn
   | WebViewOutboundEventError
   | WebViewOutboundEventMention
-  | WebViewOutboundEventTimeDetails;
+  | WebViewOutboundEventTimeDetails
+  | WebViewOutboundEventVote;
 
 // TODO: Consider completing this and making it exact, once
 // `MessageList`'s props are type-checked.
@@ -310,6 +318,19 @@ export const handleWebViewOutboundEvent = (
         originalText: event.originalText,
       });
       Alert.alert('', alertText);
+      break;
+    }
+
+    case 'vote': {
+      api.sendSubmessage(
+        props.backgroundData.auth,
+        event.messageId,
+        JSON.stringify({
+          type: 'vote',
+          key: event.key,
+          vote: event.vote,
+        }),
+      );
       break;
     }
 
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -993,6 +993,26 @@ var compiledWebviewJs = (function (exports) {
       return;
     }
 
+    if (target.matches('.poll-vote')) {
+      var _messageElement = target.closest('.message');
+
+      if (!_messageElement) {
+        throw new Error('Message element not found');
+      }
+
+      var current_vote = requireAttribute(target, 'data-voted') === 'true';
+      var vote = current_vote ? -1 : 1;
+      sendMessage({
+        type: 'vote',
+        messageId: requireNumericAttribute(_messageElement, 'data-msg-id'),
+        key: requireAttribute(target, 'data-key'),
+        vote: vote
+      });
+      target.setAttribute('data-voted', (!current_vote).toString());
+      target.innerText = (parseInt(target.innerText, 10) + vote).toString();
+      return;
+    }
+
     if (target.matches('time')) {
       var originalText = requireAttribute(target, 'original-text');
       sendMessage({
--- a/yarn.lock
+++ b/yarn.lock
@@ -2857,10 +2857,10 @@
   resolved "https://registry.yarnpkg.com/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz#e77a97fbd345b76d83245edcd17d393b1b41fb31"
   integrity sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==
 
-"@zulip/shared@^0.0.5":
-  version "0.0.5"
-  resolved "https://registry.yarnpkg.com/@zulip/shared/-/shared-0.0.5.tgz#6c5984cecb3b56b415ebfa82d4f17eab14ef75e3"
-  integrity sha512-mQNuFO4IvsvCRXTppLAAarBTdtObHE9cpqlGIqJD5rsQSv1wq3PTFxUgQuEAA+9p2L9ND2ucHzFFbMIku587JQ==
+"@zulip/shared@^0.0.6":
+  version "0.0.6"
+  resolved "https://registry.yarnpkg.com/@zulip/shared/-/shared-0.0.6.tgz#024450692670b572b6eb095845a1203d3001890e"
+  integrity sha512-17Shi/YJMln4h/B1CE3I2JiEO99tZDClTDEEXEk/XI2SGYCi1Tdb4hB48QKnlV0cy9k93KAX3oC6cydZyZUwiQ==
   dependencies:
     katex "^0.12.0"
     lodash "^4.17.19"


chrisbobbe (changes requested):
Thanks for working on this, @WesleyAC!

For the two polls on CZO [here](https://chat.zulip.org/#narrow/stream/7-test-here/topic/Test.20poll/near/1154607), here are some screenshots from iOS:



I have a few thoughts:

- In night mode, the vote count should be the same light color as the rest of the text, to contrast with the background.
- Should the vote counts be in a slightly bigger font size? They look pretty small, and it looks like they have plenty of space in the rounded rectangles to be bigger.
- The "New Choice" input isn't there yet. I agree that it's probably best to do after we've got a more basic version in.
- Even the un-selected choices give a splash of brand color that I think is rare in the message list, and (maybe?) so far reserved to mark something the user has done; emoji reactions come to mind. Would it make sense to reserve the brand color for just the choice(s) the user has chosen?
- I think the minimum touch target of 36 by 36 may end up being OK; as you've mentioned, reaction emoji aren't at least 48 by 48. One thought I had was to make the entire row touchable (so, extending the width by a lot, without extending the height). But then I thought that might conflict with a future interaction where we saw a truncated list of voters and wanted to expand to see more. If we do go for an interaction like that, and find that it doesn't conflict, perhaps something like that would be possible.
- I don't think the button layout does quite the right thing when the number of voters is very large. Here's 4 digits:
  
  I find that it does something nicer when I remove the `display: flex;` from the `.poll-widget > ul > li`, but that's just an observation and not necessarily the solution:

src/webview/html/messageAsHtml.js
```diff
@@ -71,13 +75,72 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+
+  const widgets = message.submessages
+    .filter(submessage => submessage.msg_type === 'widget')
+    .map(submessage => ({
+      ...submessage,
+      content: JSON.parse(submessage.content),
+    }));
+
+  const poll_widgets = widgets.filter(widget => widget.content.widget_type === 'poll');
+  const events = widgets.filter(widget => 'type' in widget.content);
```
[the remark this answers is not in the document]
WesleyAC: The [widget](https://zulip.readthedocs.io/en/latest/subsystems/widgets.html) docs are the closes we have to docs for this, but it won't answer your question. [`PollData`](https://github.com/zulip/zulip/blob/master/static/shared/js/poll_data.js) is where the code for this lives, which is the closest thing we have to documentation.

I'll go add some comments to that file about the format of what these messages look like. In order to develop this, I just looked at the messages from the API in order to develop this, since there weren't any docs.

[Here's](https://gist.github.com/WesleyAC/7cd8e7a94ecd597d739197b83949171e) the message data for [this message](https://chat.zulip.org/#narrow/stream/101-design/topic/bottom.20row.20design/near/1170377), in case that helps you review.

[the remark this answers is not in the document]
WesleyAC: "events" is I guess a term that I made up for the submessages that aren't the initial widget submessage (`vote` and `new_option`, and a few others), I'll add a comment explaining that.

src/webview/html/messageAsHtml.js
```diff
@@ -71,13 +75,72 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+  const poll_widget = poll_widgets[0];
+
+  const poll_data = new PollData({
+    current_user_id: ownUserId,
+    is_my_poll: message.sender_id === ownUserId,
+    question: poll_widget.content.extra_data.question,
+    options: poll_widget.content.extra_data.options,
+    comma_separated_names: () => {},
+    report_error_function: (msg: string) => {
+      Sentry.captureMessage(msg);
```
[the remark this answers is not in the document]
WesleyAC: Nope, changed.

src/webview/html/messageAsHtml.js
```diff
@@ -71,13 +75,72 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
-<div class="special-message"
- ><p>Interactive message</p
- ><p>To use, open on web or desktop</p
-></div>
+<div class="special-message">
+  <p>Interactive message</p>
+  <p>To use, open on web or desktop</p>
+</div>
 `;
+  if (!message.submessages) {
```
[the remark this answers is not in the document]
WesleyAC: Good catch, changed.

src/webview/html/messageAsHtml.js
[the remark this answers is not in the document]
WesleyAC: Added a comment about this term.

src/webview/js/js.js
```diff
@@ -887,6 +887,23 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
...
 
+  if (target.matches('.poll-vote')) {
+    // This duplicates some logic from PollData.handle.vote.outbound, but it's
+    // much simpler to just duplicate it than it is to thread a callback all
+    // the way over here.
+    const current_vote = target.getAttribute('data-voted') === 'true';
+    const vote = current_vote ? -1 : 1;
+    sendMessage({
+      type: 'vote',
+      messageId: parseInt(target.closest('.message')?.getAttribute('data-msg-id'), 10),
```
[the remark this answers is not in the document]
WesleyAC: `target.closest('.message')` should always be there, rewrote this to do that and also use some of the `requireAttribute` functions, which should make this more robust in general.

WesleyAC:
@chrisbobbe should be ready for another look.

Fixed the CSS so that it doesn't wrap, the font is the right color in night mode, and the edges aren't BRAND_COLOR until voted on.

Screenshot_20210521-074514
Screenshot_20210521-074426

src/webview/static/base.css
```diff
@@ -690,3 +690,49 @@ h1, h2, h3, h4, h5, h6 {
 .spoiler-block .spoiler-arrow.spoiler-button-open::after {
   transform: rotate(90deg) translate(10px, 0);
 }
+
+/* Poll styling */
+
+.poll-widget {
+  border: hsl(0, 0%, 50%) 1px solid;
+  padding: 2px 8px 2px 10px;
```
chrisbobbe, line 698: Why more left padding than right padding? Is it something like what Greg said at https://github.com/zulip/zulip-mobile/pull/4750#discussion_r639285725, about one of Material Design's decisions:

> Speculating, I think the idea in giving the text more left padding than right padding may be that because it's justified left, there's naturally some space between the actual text and the right edge of its box -- and so the text will look more centered if the box is actually somewhat right of center.

I notice that it makes the dividing line between the question and the options a bit off-center, horizontally:

image

This might be fine; is it deliberate?

chrisbobbe:
Cool, thanks for the revision! See one small comment below; also this comment is pretty small.

Here are some iOS screenshots:



Huh, it looks like the buttons are a bit wider on iOS? I've just tried it on Android and I'm seeing a width that looks the same as the Android screenshots you've posted. Hmm.

I think it's a left and right padding of 1em, coming from my user-agent stylesheet? See the screenshot below. If I give the element 1em left and right padding in an inline style with the dev tools, I see no change in its appearance. The `base.css:130` reference is to the following, and if I change the padding there, I don't see the button's padding changing:

```
/* Messages! And "loading" pseudomessages. */
.message,
.loading {
  display: flex;
  word-wrap: break-word;
  /* This value is used for messageReadSlop in src/webview/js/js.js, if
   * updating one, please update the other */
  padding: 16px;
  -webkit-tap-highlight-color: transparent;
}
```

Screen Shot 2021-06-02 at 8 58 15 PM

Possibly the fix is to give an explicit padding that's smaller than 1em (and doesn't use em); does that sound right? Or do you think the user-agent stylesheet is trying to tell us something important that we should pay attention to?

chrisbobbe (approved):
Cool, thanks for the revision! See one small comment below; also this comment is pretty small. Feel free to merge if you find a nice clear path to addressing them.

Here are some iOS screenshots:



Huh, it looks like the buttons are a bit wider on iOS? I've just tried it on Android and I'm seeing a width that looks the same as the Android screenshots you've posted. Hmm.

I think it's a left and right padding of 1em, coming from my user-agent stylesheet? See the screenshot below. If I give the element 1em left and right padding in an inline style with the dev tools, I see no change in its appearance. The `base.css:130` reference is to the following, and if I change the padding there, I don't see the button's padding changing:

```
/* Messages! And "loading" pseudomessages. */
.message,
.loading {
  display: flex;
  word-wrap: break-word;
  /* This value is used for messageReadSlop in src/webview/js/js.js, if
   * updating one, please update the other */
  padding: 16px;
  -webkit-tap-highlight-color: transparent;
}
```

Screen Shot 2021-06-02 at 8 58 15 PM

Possibly the fix is to give an explicit padding that's smaller than 1em (and doesn't use em); does that sound right? Or do you think the user-agent stylesheet is trying to tell us something important that we should pay attention to?

src/webview/static/base.css
WesleyAC: This was copied from the `spoiler-block` css, I'm not sure why it's like that.

WesleyAC:
Updated the padding to be explicit :) Will merge EOD Friday if I don't hear otherwise by then.

chrisbobbe:
Hmm, actually, could you try visiting https://chat.zulip.org/#narrow/stream/7-test-here/topic/poll.20test in the app and see if you get an error like this:

```
ExceptionsManager.js:179 TypeError: Cannot read property 'question' of null
    at widgetBody (messageAsHtml.js:121)
    at _default (messageAsHtml.js:190)
    at contentHtmlFromPieceDescriptors.js:29
    at Array.forEach (<anonymous>)
    at contentHtmlFromPieceDescriptors.js:25
    at Array.forEach (<anonymous>)
    at _default (contentHtmlFromPieceDescriptors.js:21)
    at updateContent (generateInboundEvents.js:50)
    at generateInboundEvents (generateInboundEvents.js:109)
    at MessageList._this.shouldComponentUpdate (MessageList.js:221)
```

It looks like the `poll_widget` it's having a problem with looks like this (JSON.stringified); `extra_data` is null:

```
{
    "id": 299,
    "message_id": 582872,
    "sender_id": 703,
    "msg_type": "widget",
    "content": {
        "widget_type": "poll",
        "extra_data": null
    }
}
```

chrisbobbe:
> Updated the padding to be explicit

Cool, looks good now.

WesleyAC:
Hmmm, I'm curious how that was generated. I see a similar error in the console in the webapp:

```
in process_submessages: null has no properties
```

I'll make it so it doesn't crash, at least, but I think that message is generally malformed.

chrisbobbe:
> Hmmm, I'm curious how that was generated. I see a similar error in the console in the webapp:
> 
> ```
> in process_submessages: null has no properties
> ```
> 
> I'll make it so it doesn't crash, at least, but I think that message is generally malformed.

Interesting; @showell, do you know what might be going on? It's at least one message in https://chat.zulip.org/#narrow/stream/7-test-here/topic/poll.20test

WesleyAC:
It's the first message that's the problem, specifically.

WesleyAC:
Anyways, added a check for that :)

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+  <p>Interactive message</p>
+  <p>To use, open on web or desktop</p>
+</div>
 `;
+  }
+  const poll_widget = poll_widgets[0];
+
+  if (!poll_widget.content || !poll_widget.content.extra_data) {
+    // We don't expect this to happen in general, but there are some malformed
+    // messages lying around that will trigger this.
```
chrisbobbe, line 118: It might be good for the comment (or possibly the commit message) to point to the specific message(s) we know about? Without that, I could see the check lingering far after the problem gets fixed, because no one knows whether it's been fixed and there's no clear path to finding that out.

Or possibly the fact that messages are out there in the wild is enough to say that this logic has to remain permanently, since servers don't generally re-render old messages when the rendering logic changes ([doc](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#html)). Which would be a slightly unfortunate side effect on the codebase.

WesleyAC: IMO it should remain as long as there are messages like this out there. (Also, clients can send messages like this, and we don't want this as a DOS vector). Unfortunate, yeah, but there's not much we can do about it :/

chrisbobbe: Makes sense. Before merging, what do you think about seeing what Steve says, with a timeout on that of a few days?

showell:
We should keep the defensive code here in the client.

src/webview/js/js.js
```diff
@@ -904,6 +904,28 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
...
+      sendMessage({
+        type: 'vote',
+        messageId: requireNumericAttribute(messageElement, 'data-msg-id'),
+        key: requireAttribute(target, 'data-key'),
+        vote,
+      });
+      target.setAttribute('data-voted', (!current_vote).toString());
+      target.innerText = (parseInt(target.innerText, 10) + vote).toString();
+    } else {
+      throw new Error('Message element not found');
```
gnprice, line 924: Let's put the error case first, so it's immediately next to the condition that triggers it.

That also means the normal happy case doesn't get indented, and can be read just linearly down the screen.

src/webview/js/js.js
```diff
@@ -904,6 +904,28 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.matches('.poll-vote')) {
+    const messageElement = target.closest('.message');
+    if (messageElement) {
+      // This duplicates some logic from PollData.handle.vote.outbound, but it's
+      // much simpler to just duplicate it than it is to thread a callback all
+      // the way over here.
```
gnprice, line 912: Yeah definitely. Plus that code is expecting a different set of data structures from what we have -- so really all that's in common is the -1 / 1 thing, and the shape of the submessage.

I think this comment doesn't even need the justification. But it is definitely helpful to mention that this corresponds to the logic in that particular other bit of code -- and one thing it would be quite helpful to add is the filename of that code. Otherwise it's not necessarily obvious where to find `PollData`.

src/webview/js/js.js
```diff
@@ -904,6 +904,28 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
...
+      const current_vote = requireAttribute(target, 'data-voted') === 'true';
+      const vote = current_vote ? -1 : 1;
+      sendMessage({
+        type: 'vote',
+        messageId: requireNumericAttribute(messageElement, 'data-msg-id'),
+        key: requireAttribute(target, 'data-key'),
+        vote,
+      });
+      target.setAttribute('data-voted', (!current_vote).toString());
+      target.innerText = (parseInt(target.innerText, 10) + vote).toString();
```
gnprice, line 922: Hmm, interesting.

So one consequence of this local-echo strategy is that if for whatever reason the message gets rerendered -- for example because another vote came in, or because some other message changed or a new message arrived in this narrow -- this change will get forgotten. That seems awkward.

What does the webapp do to handle the latency of voting?

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+    return template`
 $!${message.content}
-<div class="special-message"
- ><p>Interactive message</p
- ><p>To use, open on web or desktop</p
-></div>
+<div class="special-message">
+  <p>Interactive message</p>
+  <p>To use, open on web or desktop</p>
+</div>
```
gnprice, line 111: This adds whitespace which can potentially affect how the HTML gets rendered. In general I prefer to avoid doing that, because when it does turn out to matter it isn't fun tracking down why there's a stray space somewhere.

(In any case this would be an independent change.)

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+ *
+ * Must not be called on a message without any submessages.
+ */
+const widgetBody = (message: Message, ownUserId: UserId) => {
+  invariant(
+    message.submessages !== undefined && message.submessages.length > 0,
+    'should have submessages',
+  );
+
+  const widgets = message.submessages
```
gnprice, line 91: This doesn't seem like the right name -- the whole thing is a widget. These submessages are things that happened to the widget.

Perhaps `widgetSubmessages`?

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+    }));
+
+  const poll_widgets = widgets.filter(widget => widget.content.widget_type === 'poll');
+  // "events" are submessages other than the main widget submessage - for
+  // instance, `vote`, `new_option`, and similar submessages. These are not
+  // documented, look at the PollData code if you're interested in what the
+  // possibilities are.
+  const events = widgets.filter(widget => 'type' in widget.content);
+
+  if (poll_widgets.length !== 1) {
```
gnprice, line 105: Does this logic correspond to what the webapp does?

I think it'd be best to try to make this logic track what's in the webapp. Particularly in the absence of documentation.

It looks like the corresponding logic here is around `do_process_submessages` in `static/js/submessage.js`, and its callees.

gnprice (commented):
Thanks @WesleyAC, and thanks @chrisbobbe for the reviews! Will be good to have this implemented. Some comments below.

gnprice:
> * I have a `$FlowFixMe` for adding types to `PollData`. I'm would like to do this, although probably not complete types: there are some fundamental limits to how much we can get from typing here, since there's so much dynamism in how widgets work.

Yeah, we should definitely do at least something minimal here.

I think there actually isn't even all that much dynamism -- a novel unknown type of widget can have arbitrary shapes of data, but for a specific widget type (like polls) it's pretty well specified, and `PollData` is about a specific widget type.

But in any case we should do at least something minimal, so that if there's anything difficult we can see concretely what it is. See `typing_status.js.flow` for an example of how to add types in the shared tree.

src/webview/html/messageAsHtml.js
WesleyAC: That code does this:

```javascript
    // Right now, our only use of submessages is widgets.

    const data = events[0].data;

    if (data === undefined) {
        return;
    }
```

Which seems strictly worse, to me. (It assumes `events[0]` is the widget submessage, and it assumes that there's only one widget in the message).

I guess it'd probably be for the best to do `if (poll_widgets.length == 0)`, so we're less likely to break in a different way from the webapp in the future. But I don't think we should replace the `.filter()` here with `[0]`, for instance.

src/webview/js/js.js
WesleyAC: Good call, done.

src/webview/js/js.js
WesleyAC: Yeah, I don't think the justification is super needed, but I'm going to leave it in unless you'd specifically prefer I didn't, since my general intuition is that it's better to have too much comment than too little, and this is the good sort of comment (explaining justification for why the code is the way it is, rather than explaining what the code is).

src/webview/js/js.js
WesleyAC: I'm not sure, @showell, any idea about that?

My inclination is that building a particularly fancy local-echo is not worth it, but I'd be open to suggestions about small tactical fixes.

src/webview/html/messageAsHtml.js
WesleyAC: Removed this change.

src/webview/html/messageAsHtml.js
WesleyAC: Renamed :)###

WesleyAC:
Should be ready for another review! I'll open a PR for adding basic types in the shared repo. I'd rather not block this PR on that, though.

src/webview/html/messageAsHtml.js
chrisbobbe: Interesting; I'll keep this in mind about whitespace. I looked at the old code and thought it was super weird; now I understand why we'd want it that way 🙂

src/webview/js/js.js
WesleyAC: There is no local echo in the webapp. I don't think that's acceptable in the mobile app, where it's easy to think you mistapped if the feedback is delayed.

src/webview/js/js.js
gnprice: That's OK, but please do add the filename. It's otherwise not all that obvious where to find the logic this is talking about. (Especially so since `PollData` isn't imported in this file.)

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
  ><p>To use, open on web or desktop</p
 ></div>
 `;
+  }
+  const poll_widget = poll_widgets[0];
+
+  if (!poll_widget.content || !poll_widget.content.extra_data) {
+    // We don't expect this to happen in general, but there are some malformed
+    // messages lying around that will trigger this.
+    return template`$!${message.content}`;
```
gnprice, line 119: Tracing through to the callees of the webapp's `do_process_submessages` as mentioned at https://github.com/zulip/zulip-mobile/pull/4704#discussion_r646821062 , it looks like the webapp has a different behavior here:
```js
    extra_data: {question = "", options = []} = {},
```

So in particular if there's a poll that starts with no options, the webapp will go ahead and present it normally -- with `new_option` submessages, it can end up as a perfectly functional poll. We should do the same thing, unless we know of a reason to behave differently.

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+  }
+
+  const poll_data = new PollData({
+    current_user_id: ownUserId,
+    is_my_poll: message.sender_id === ownUserId,
+    question: poll_widget.content.extra_data.question,
+    options: poll_widget.content.extra_data.options,
+    comma_separated_names: () => {},
```
gnprice, line 127: This should get a TODO comment -- it's a stub we'll need to fill in to get the feature fully working.

Also it should return a string, probably "".

src/webview/html/messageAsHtml.js
gnprice: Please also propagate the rename to where we iterate through this array :) -- like here:
```js
  const events = widgetSubmessages.filter(widget => 'type' in widget.content);
```

Each of those is a submessage, and isn't a widget.

gnprice (commented):
Comment below, and others just above and on threads in my original review https://github.com/zulip/zulip-mobile/pull/4704#pullrequestreview-677706852 .

One somewhat more high-level thing that I'd like to see before merging this is to have effective types on this code. It'd be good to have them on the API boundary with the shared `PollData`, but the priority is that we should have them on the data we're getting from the server -- in particular, the result of that `JSON.parse(submessage.content)`. As is, that ends up just being `any`.

The fundamental reason I want types on that data is that it's part of making explicit what our assumptions are about the shape of the data. That's important here because those assumptions come basically from some reverse-engineering work -- and it's a big pain to repeat that work just in order to read this code. Effectively that means that when revisiting this code we just won't realistically know exactly what assumptions it's relying on, which makes it fragile.

Much better to have the results written down; that lets us read the code and straightforwardly check it against the assumptions, and at the same time it makes it much easier to check the assumptions themselves against a new server/webapp version, or against API docs if and when this subsystem gets some of those.

I'll push to the PR branch a commit that adds relevant types to `src/api/modelTypes.js`, based on what I've looked at while reviewing this PR. Then you can use a type annotation to say that the result of that `JSON.parse` is a `SubmessageData`. From there it should be possible to write the code so that Flow accepts it with no fixmes, though you might need an assertion or two for assumptions that aren't encoded in the type itself.

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +77,87 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+      ...submessage,
+      content: JSON.parse(submessage.content),
+    }));
+
+  const poll_widgets = widgetSubmessages.filter(widget => widget.content.widget_type === 'poll');
+  // "events" are submessages other than the main widget submessage - for
+  // instance, `vote`, `new_option`, and similar submessages. These are not
+  // documented, look at the PollData code if you're interested in what the
+  // possibilities are.
+  const events = widgetSubmessages.filter(widget => 'type' in widget.content);
```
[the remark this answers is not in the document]
WesleyAC: Yeah, that makes sense. I hadn't realized that PollData was robust to receiving the initial `widget` submessage in addition to the "event" submessages, but it seems that it is. Done.

src/webview/html/messageAsHtml.js
WesleyAC: Done. FWIW, I expect to change this to operate based on user IDs when we actually build this feature.

src/webview/html/messageAsHtml.js
WesleyAC: I don't believe that's what this is checking for — if you do `/poll` in the webapp, the mobile app handles that fine. This is for a particular weird message that as far as I can tell, no official client will send anymore. From what I saw, the webapp threw an error in the console for the only message that I found that triggered this condition.

Do let me know which specific part of the webapp codebase you're looking at that you think handles this if that's not the case, though.

src/webview/html/messageAsHtml.js
WesleyAC: Added the sorting code, thanks for catching that.

> It also wouldn't make any sense for votes to come after the creation of the poll itself.

Presumably you mean before, not after?

> Basically I would much rather be encoding the assumptions that are in the web implementation -- which was written by Steve, who also specced out this subsystem in the first place -- than encoding a different set of implicit assumptions from our own differently-shaped implementation.

I think that this is reasonable in general, be we do actually have a slightly different set of constraints than the webapp — the mobile app needs to be more robust to future server versions, whereas the webapp can safely assume that (for instance), all submessages will be widgets, since the clientside code will change at the same time as the server side code.

I've rewritten the logic to be much more similar to the webapp, but there are still a couple checks that I've added that I think make sense to be there.

src/webview/html/messageAsHtml.js
WesleyAC: Done. I also made a bunch of stuff camel case while I was at it :)

WesleyAC:
Pushed some changes that should address most of this :)

I wasn't able to get flow to figure out that checking that `pollWidget.content.widget_type` is `poll` means that `extra_data` will always have `question` and `options`, I'd be happy for you to take a look at writing that in a way that flow understands.

showell:
We should block this until #18805 is merged on the server side. (It has a one-line change to how `PollData` is constructed.)

WesleyAC:
Updated to use the new version of the shared PollData code from zulip/zulip#18805.

@gnprice, can you take a look at the flow errors here?

WesleyAC:
Ah, I figured out the flow issue on my own, should be good for another review now :)

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +78,90 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+    // messages lying around that will trigger this.
+    return template`$!${message.content}`;
+  }
+
+  const pollData = new PollData({
+    message_sender_id: message.sender_id,
+    current_user_id: ownUserId,
+    is_my_poll: message.sender_id === ownUserId,
+    question:
+      pollWidget.content.extra_data.question == null ? pollWidget.content.extra_data.question : '',
```
gnprice, line 127: nit: Can equivalently say `pollWidget.content.extra_data.question ?? ''`, and that makes it a bit more obvious what's going on.

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +78,90 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+ * Must not be called on a message without any submessages.
+ */
+const widgetBody = (message: Message, ownUserId: UserId) => {
+  invariant(
+    message.submessages !== undefined && message.submessages.length > 0,
+    'should have submessages',
+  );
+
+  const sortedSubmessages = message.submessages.concat();
+  sortedSubmessages.sort((m1, m2) => m1.id - m2.id);
```
gnprice, line 93: Ah neat, I hadn't seen `Array#concat` used for this.

Another thing that can simplify this is that `Array#sort` returns the array. So there's no need to initialize first and then sort; can just initialize with the result of `sort`.

Then in turn that expression can just be inlined in the one place we're using this.

src/webview/html/messageAsHtml.js
```diff
@@ -72,13 +78,90 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+      pollWidget.content.extra_data.options == null ? pollWidget.content.extra_data.options : [],
+    // TODO: Implement this.
+    comma_separated_names: () => '',
+    report_error_function: (msg: string) => {
+      logging.error(msg);
+    },
+  });
+
+  for (const pollEvent of widgetSubmessages) {
+    pollData.handle_event(pollEvent.sender_id, pollEvent.content);
```
gnprice, line 138: > I hadn't realized that PollData was robust to receiving the initial `widget` submessage in addition to the "event" submessages, but it seems that it is.

Hmm yeah, it looks like it is. It looks for `type` and drops anything that doesn't have that, and the initial submessage doesn't have it.

I'm not sure that's really part of its interface, though -- it seems like a bit of an odd thing to do. The web caller doesn't re-pass it that first submessage; in `activate` in `widgetize.js` the flow looks like this:
```js
    events.shift();

    // …

    // Replay any events that already happened.  (This is common
    // when you narrow to a message after other users have already
    // interacted with it.)
    if (events.length > 0) {
        widget_elem.handle_events(events);
    }
```

So that `shift` drops the first element so that it doesn't go to the `handle_events` call. (Which in turn is basically a loop with `handle_event`.)

It's pretty easy to similarly skip the first element from what we use for `handle_event`, and that makes it a bit clearer to understand what's going on, so let's do that.

gnprice (commented):
Thanks for the revision! Here's a couple of comments. More to come; but GitHub won't seem to let me reply to threads above with this review pending, so here it is so far.

src/webview/html/messageAsHtml.js
gnprice: > Presumably you mean before, not after?

Ah, yep.


> the mobile app needs to be more robust to future server versions, whereas the webapp can safely assume that (for instance), all submessages will be widgets, since the clientside code will change at the same time as the server side code.

True! I'm happy to have checks for things like that, to handle forward-compatibility that the webapp can handle by just changing the client code when the time comes.

The really essential thing I want to avoid here is the opposite direction, where we encode assumptions that aren't in the web implementation nor documented for the API. That's how we wind up with bugs like the one where we [assumed a new message sent by self was never unread](https://chat.zulip.org/#narrow/stream/3-backend/topic/muted.20users/near/1171634).

And then the practical way to accomplish that -- at least for the situation this subsystem is in, where it's impossible to use only assumptions that are documented in the API, because there is no API documentation -- is to keep our implementation's logic generally shaped in a similar way to the logic in the web implementation, so that it's possible to compare without a lot of work.

Another aspect of forward compatibility is that if we encounter something from the server that's outside the current known patterns, it's not enough to just do something; we have to do the right thing, or do an appropriate kind of nothing. If we go and give meanings to a wide swath of currently-unused protocol space, that's just as capable of conflicting with future API extensions as any bad assumption.

Then in order to effectively avoid doing that, what's needed is basically to have some kind of coherent, and more or less explicit, theory of how the API is meant to be extended in the future. That's a key part of what our API types are for. For example the new `WidgetData` type discusses this pretty directly:
```js
/**
 * The data encoded in a submessage to make the message a widget.
 *
 * Note that future server versions might introduce new types of widgets, so
 * `widget_type` could be a value not included here.  But when it is one of
 * these values, the rest of the object will follow this type.
 */
// Ideally we'd be able to express both the known and the unknown widget
// types: we'd have another branch of this union which looked like
//   | {| +widget_type: (string *other than* those above), +extra_data?: { ... } |}
// But there doesn't seem to be a way to express that in Flow.
export type WidgetData =
  | {|
      +widget_type: 'poll',
      +extra_data?: {| +question?: string, +options?: $ReadOnlyArray<string> |},
    |}
  // We can write these down more specifically when we implement these widgets.
  | {| +widget_type: 'todo', +extra_data?: { ... } |}
  | {| +widget_type: 'zform', +extra_data?: { ... } |};
```

Then it's possible to
* look at the code, and compare it to the explicit expectations, and check they're aligned;
* and look at the explicit expectations, and discuss them or think about them from an API perspective and evaluate whether they seem right, without having to reverse-engineer the implementation to extract its assumptions as part of that discussion.

src/webview/html/messageAsHtml.js
WesleyAC: That sounds lovely, but when I try it, flow reports:

```
Cannot get pollWidget.content.extra_data.question because property question is missing in property
extra_data of unknown type [1]. [incompatible-use]

 [1] 116│   if (pollWidget.content.extra_data == null) {
        :
     123│     message_sender_id: message.sender_id,
     124│     current_user_id: ownUserId,
     125│     is_my_poll: message.sender_id === ownUserId,
     126│     question: pollWidget.content.extra_data.question ?? '',
```

src/webview/html/messageAsHtml.js
WesleyAC: Oh, wait, that's because I got the conditional backwards, oops :disappointed: 

Which means that the flow error I was fighting with before is real :/

src/webview/html/messageAsHtml.js
WesleyAC: (I actually had the conditional backwards in the code I pushed, but even with the conditional correct, it passes flow, but doesn't with the `??`)

src/webview/html/messageAsHtml.js
WesleyAC: Yeah, that all makes sense :+1:

WesleyAC:
Ok, I pushed a revision that addresses everything except the `??`, which flow seems to treat differently than the ternary operator.

Should be ready for another look :)

WesleyAC:
@gnprice Pushed a revision based on what we talked about on VC + in person.

src/webview/html/messageAsHtml.js
gnprice: FTR this is now fixed by the cast, with a fixme, to `WidgetData`. The error is correct -- this logic is relying on assumptions that aren't in the types, because the structure of these submessages arrays is kind of quirky and can't be expressed in the type system.

(We never did figure out why there wasn't an error in the `? :` case.)

src/webview/html/messageAsHtml.js
```diff
@@ -72,14 +79,98 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+
+  if (pollWidgetContent.widget_type !== 'poll') {
+    return errorMessage;
+  }
+
+  if (pollWidgetContent.extra_data == null) {
+    // We don't expect this to happen in general, but there are some malformed
+    // messages lying around that will trigger this.
+    return template`$!${message.content}`;
+  }
```
gnprice, line 130: So from the webapp logic I quoted at https://github.com/zulip/zulip-mobile/pull/4704#discussion_r648546322 , it looks like the webapp's behavior in this case should be more like:
```js
  const { extra_data = {} } = pollWidgetContent;
```

We can vary from that, but I'd like to be explicit about doing so, because that makes it possible to compare the two implementations and spot discrepancies.

I think you showed me on your screen the other day the old message you had in mind, in the webapp, and it showed up with a behavior more like this, with just some text like "/poll". (There was an error in the JS console, suggesting the code may have errored out before it reached the code I quoted.) I think you may have suggested some reasoning for why showing the raw content like this is preferable to giving `extra_data` a default, like: at least then it'll say "/poll" instead of just looking empty.

Would you put that reasoning, and a link to that message, into the commit? And in the code, a comment saying this is different from web's behavior in this error case but we think this is better.

src/webview/html/messageAsHtml.js
```diff
@@ -72,14 +79,98 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
 ></div>
 `;
 
+  const pollWidget = widgetSubmessages.shift();
+  if (!pollWidget || !pollWidget.content || pollWidget.content.widget_type !== 'poll') {
+    return errorMessage;
+  }
+
+  // $FlowFixMe[incompatible-type]
+  const pollWidgetContent: WidgetData = pollWidget.content;
```
gnprice, line 120:
```suggestion
  /* $FlowFixMe[incompatible-type]: The first widget submessage should be
       a `WidgetData`; see jsdoc on `SubmessageData`. */
  const pollWidgetContent: WidgetData = pollWidget.content;
```

When writing a fixme, we always want some kind of comment saying why it's appropriate -- like, what kind of assumption is being expressed and where that assumption comes from.

src/webview/html/messageAsHtml.js
```diff
@@ -72,14 +79,98 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+  const errorMessage = template`
 $!${message.content}
 <div class="special-message"
  ><p>Interactive message</p
  ><p>To use, open on web or desktop</p
 ></div>
 `;
 
+  const pollWidget = widgetSubmessages.shift();
+  if (!pollWidget || !pollWidget.content || pollWidget.content.widget_type !== 'poll') {
```
gnprice, line 115: This last condition is redundant with the one below.

I think when we were figuring out the `WidgetData` thing the other day while looking at your screen, you added the second copy of the condition as a quick experiment to check that that was enough to resolve the remaining errors, and indeed it was. It's OK for this logic to wind up a little messy, but we should delete the outright duplicate :slightly_smiling_face:

src/webview/html/messageAsHtml.js
```diff
@@ -72,14 +79,98 @@ $!${messageReactionListAsHtml(reactions, ownUser.user_id, allImageEmojiById)}
...
+  );
+
+  const widgetSubmessages: Array<{
+    sender_id: number,
+    content: SubmessageData,
+    ...
+  }> = message.submessages
+    .concat()
+    .sort((m1, m2) => m1.id - m2.id)
+    .filter(submessage => submessage.msg_type === 'widget')
```
gnprice, line 100: Not at all essential, but: if we put the `filter` before the `sort`, we can drop the `concat`, because the `filter` will already be creating a new array anyway.

gnprice (commented):
Thanks @WesleyAC for the revisions! A few comments below, to complete the threads we were talking about the other day; then I think this will be ready to merge.

(If not for the `extra_data` bit, I'd probably add the types comment as an extra commit, merge, and send a quick followup PR with the other, non-comment tweaks. But the `extra_data` item is more complex because it involves some data you've looked at that I don't have a link to.

And since we are doing a quick last round here, please take care of those other items too -- they'll be just about as quick for you as they would for me :slightly_smiling_face: )

src/webview/html/messageAsHtml.js
WesleyAC: I commented this case, filed #19145, and added a link to the message. I didn't modify the commit message, since it seems redundant to do that given the comment.

I tried to word it carefully, since the behaviour of this code is identical to the behaviour of the webapp as a whole, except for the error message printed to the console. The *local* behaviour is different, so it seems worth commenting still.

src/webview/html/messageAsHtml.js
WesleyAC: Moved this condition down :)

WesleyAC:
@gnprice revision pushed :)

gnprice:
Thanks! Looks good -- merging :tada:

PR #4707, opened 2021-04-29
accountsReducer: Use `restart` events to update server info when we can.

chrisbobbe:
Following from Greg's idea in discussion:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Server.20feature.20level/near/1165178.

> * when we handle the event, we can immediately update that data in Redux -- in particular, we can do so atomically vs. any future events we might try to use the feature level to interpret;


src/api/eventTypes.js
```diff
@@ -151,3 +152,18 @@ export type UpdateMessageFlagsEvent = {|
...
+
+  // Servers at feature level 59 or above send these
+  // (zulip/zulip@65c400e06). The fields describe the server after the
+  // ugprade; handling an event that includes them is the fastest way
+  // to learn about a server upgrade.
+  //
+  // They have the same shape and meaning as the same-named fields in
+  // the /server_settings and /register responses.
+  zulip_version?: string,
+  zulip_feature_level?: number,
```
WesleyAC, line 168: Should we also have `server_generation` and `immediate`?

src/account/accountsReducer.js
```diff
@@ -110,6 +112,34 @@ export default (state: AccountsState = initialState, action: Action): AccountsSt
...
+              {
+                ...state[0],
+                zulipVersion: new ZulipVersion(zulip_version),
+                zulipFeatureLevel: zulip_feature_level ?? 0,
+              },
+              ...state.slice(1),
+            ];
+          } else {
+            return state;
+          }
```
WesleyAC, line 136: It might be simpler to do:

```javascript
return [
  {
    ...state[0],
    zulipVersion: zulip_version ? new ZulipVersion(zulip_version) : state[0].zulipVersion,
    zulipFeatureLevel: zulip_feature_level ? zulip_feature_level : state[0].zulipFeatureLevel,
  },
  ...state.slice(1),
];
```

I don't have a strong feeling that this would be better, just something to think about.

src/account/accountsReducer.js
```diff
@@ -110,6 +112,34 @@ export default (state: AccountsState = initialState, action: Action): AccountsSt
...
+            // TODO (?): Detect if these are different from the values
+            // we had before, so we know it's an upgrade, not just a
+            // restart. Then, implement logic like "be sure to refetch
+            // from scratch within N hours" (to avoid thundering
+            // herding the server).
+            return [
+              {
+                ...state[0],
+                zulipVersion: new ZulipVersion(zulip_version),
+                zulipFeatureLevel: zulip_feature_level ?? 0,
```
WesleyAC, line 130: Do we expect `zulip_feature_level` to be null ever? It seems weird be me tat we need the `?? 0` here given the `undefined` check above.

src/account/accountsReducer.js
```diff
@@ -110,6 +112,34 @@ export default (state: AccountsState = initialState, action: Action): AccountsSt
...
+          const { zulip_feature_level, zulip_version } = event;
+          if (zulip_feature_level !== undefined && zulip_version !== undefined) {
+            // TODO (?): Detect if these are different from the values
+            // we had before, so we know it's an upgrade, not just a
+            // restart. Then, implement logic like "be sure to refetch
+            // from scratch within N hours" (to avoid thundering
+            // herding the server).
+            return [
+              {
+                ...state[0],
```
WesleyAC, line 128: Will the active account always be `state[0]`? This seems like a brittle way of doing things in general, although possibly it's too embedded to easily change at this point. (This isn't really a comment on this PR, more of a general question)

WesleyAC (commented):
Looks good overall! Just a few comments.

src/account/accountsReducer.js
chrisbobbe: Yeah, possibly. One feature of the code in this revision is that `state` won't be replaced in cases where it doesn't meaningfully change (and so components that subscribe to it using react-redux won't rerender, etc.). But that optimization isn't really important at all because these events will be rare.

I think we could still get that feature while using a conditional operator, and so only having one `return` statement which I kind of like; I'll try that?

src/account/accountsReducer.js
chrisbobbe: Good point; no, we don't expect any servers to send `null` for this (I confirmed by checking `ApiResponseServerSettings`).

src/account/accountsReducer.js
chrisbobbe: > Will the active account always be `state[0]`

Yeah, that's how it is. There's been some discussion on the brittleness of that; I don't remember where.

chrisbobbe:

src/account/accountsReducer.js
WesleyAC: In that case, could you remove the `?? 0`? (perhaps you'd need to change the check from `zulip_feature_level !== undefined` to `zulip_feature_level != null` or similar in order to make flow happy, I'm not sure).

WesleyAC:
One more comment about the `?? 0`, feel free to merge once you've resolved that :)

src/account/accountsReducer.js
chrisbobbe: > In that case, could you remove the `?? 0`

Oops—I thought I'd done that for this revision; looks like not. 🙂 Will fix.

chrisbobbe:
Thanks for the review! Merged, after making that change (which I'd meant to do in the previous revision and somehow didn't 😛).

src/account/accountsReducer.js
```diff
@@ -110,6 +112,33 @@ export default (state: AccountsState = initialState, action: Action): AccountsSt
...
+          return zulip_feature_level !== undefined && zulip_version !== undefined
+            ? [
+                {
+                  ...state[0],
+
+                  // TODO (?): Detect if these are different from the values
+                  // we had before, so we know it's an upgrade, not just a
+                  // restart. Then, implement logic like "be sure to refetch
+                  // from scratch within N hours" (to avoid thundering
+                  // herding the server).
```
timabbott, line 129: I think the "check if it's different" part is pretty valuable for load on servers that upgrade often from Git.

Also, I suspect we only need to check zulip_feature_level -- the Zulip version will change with every Git update.

chrisbobbe: > I think the "check if it's different" part is pretty valuable for load on servers that upgrade often from Git.

That sounds right. This PR doesn't have us start doing a `/register` on restart events, it only updates the state, which we'll use for the rest of the session. Also, that state is persisted, so it'll inform the beginning of the next session, until that session naturally does its own `/register` which we'll grab a new value from.

So I think the load on servers is unchanged. When we do start having restart events cause a `/register`, we'll want to be careful about thundering herding.

gnprice:
Thanks @chrisbobbe and @WesleyAC !

I have some small code-style adjustments here, which I've just sent as #4794. I've also filed #4793 for the remaining TODO here (which currently doesn't matter at all, and isn't urgent), expanding a bit on what will cause us to need it in the future.

PR #4711, opened 2021-05-01
css: Don't show mention highlight for muted messages.

WesleyAC:
We don't want messages that are muted and hidden to show the user that
they are mentioned in the message.


chrisbobbe:
Thanks, this LGTM! Merged.

PR #4716, opened 2021-05-02
AccountDetails: Hide local time and last active for current user.

WesleyAC:
The user should know their local time, and always be "Active now", so
it's just clutter to show these.


chrisbobbe (approved):
Thanks, @WesleyAC! This makes sense to me.

PR #4717, opened 2021-05-02
ChatNavBar: Don't use BRAND_COLOR for text.

WesleyAC:
The BRAND_COLOR is a pretty aggressive color, that doesn't look good as
text. Instead, just use the default text color.

before:
chatnav_before

after:
chatnav_after

### DIFF
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -23,7 +23,7 @@ type Props = $ReadOnly<{|
 export default function ChatNavBar(props: Props) {
   const { narrow, editMessage } = props;
   const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
-  const color =
+  const buttonColor =
     streamColor === undefined ? BRAND_COLOR : foregroundColorFromBackground(streamColor);
   const themeColor = useContext(ThemeContext).color;
   const textColor =
@@ -56,10 +56,10 @@ export default function ChatNavBar(props: Props) {
             alignItems: 'center',
           }}
         >
-          <NavBarBackButton color={color} />
+          <NavBarBackButton color={buttonColor} />
           <Title color={textColor} narrow={narrow} editMessage={editMessage} />
-          <ExtraButton color={color} narrow={narrow} />
-          <InfoButton color={color} narrow={narrow} />
+          <ExtraButton color={buttonColor} narrow={narrow} />
+          <InfoButton color={buttonColor} narrow={narrow} />
         </View>
         <LoadingBanner
           spinnerColor={spinnerColor}
--- a/src/nav/ChatNavBar.js
+++ b/src/nav/ChatNavBar.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React from 'react';
+import React, { useContext } from 'react';
 import { View } from 'react-native';
 import Color from 'color';
 import { SafeAreaView } from 'react-native-safe-area-context';
@@ -8,7 +8,7 @@ import { SafeAreaView } from 'react-native-safe-area-context';
 import type { Narrow, EditMessage } from '../types';
 import { LoadingBanner, ZulipStatusBar } from '../common';
 import { useSelector } from '../react-redux';
-import { BRAND_COLOR, NAVBAR_SIZE } from '../styles';
+import { BRAND_COLOR, NAVBAR_SIZE, ThemeContext } from '../styles';
 import Title from '../title/Title';
 import NavBarBackButton from './NavBarBackButton';
 import { getStreamColorForNarrow } from '../subscriptions/subscriptionSelectors';
@@ -25,6 +25,9 @@ export default function ChatNavBar(props: Props) {
   const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
   const color =
     streamColor === undefined ? BRAND_COLOR : foregroundColorFromBackground(streamColor);
+  const themeColor = useContext(ThemeContext).color;
+  const textColor =
+    streamColor === undefined ? themeColor : foregroundColorFromBackground(streamColor);
   const spinnerColor =
     streamColor === undefined ? 'default' : foregroundColorFromBackground(streamColor);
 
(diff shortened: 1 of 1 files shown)


src/nav/ChatNavBar.js
```diff
@@ -54,14 +54,14 @@ export default function ChatNavBar(props: Props) {
...
         >
           <NavBarBackButton color={color} />
-          <Title color={color} narrow={narrow} editMessage={editMessage} />
+          <Title color={textColor} narrow={narrow} editMessage={editMessage} />
           <ExtraButton color={color} narrow={narrow} />
           <InfoButton color={color} narrow={narrow} />
         </View>
         <LoadingBanner
-          spinnerColor={spinnerColor}
+          spinnerColor={textColor}
```
chrisbobbe, line 62: Huh, yeah; I'd say `LoadingBanner` should always get the same color for its spinner and its text. Probably would be good to change its interface to enforce that, except that we're on the path to reworking/replacing `LoadingBanner` in a way that this detail will become irrelevant; so, needn't bother, I think.

src/nav/ChatNavBar.js
```diff
@@ -54,14 +54,14 @@ export default function ChatNavBar(props: Props) {
           }}
         >
           <NavBarBackButton color={color} />
-          <Title color={color} narrow={narrow} editMessage={editMessage} />
+          <Title color={textColor} narrow={narrow} editMessage={editMessage} />
```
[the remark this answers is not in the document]
chrisbobbe: If 'default' turns out to cause any problems, then:

In cases where we don't have to adapt dynamically to a stream-colored background color (i.e., for narrows other than stream and topic narrows), I think a sensible replacement for 'default' would be `useContext(ThemeContext).color`, i.e., the theme-appropriate text color we use in `Label` and other places.

chrisbobbe (commented):
Thanks! I agree; the current coloring doesn't look right, and probably wouldn't pass good, rigorous color-contrast thresholds.

Do I understand correctly from reading the code that this doesn't affect the navbar on stream/topic narrows?

See also a few comments below.

src/nav/ChatNavBar.js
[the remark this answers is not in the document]
WesleyAC: Hmmm, you're right. This worked, but I suspect that was only by accident.

I switched it to keep `spinnerColor`, but use `useContext(ThemeContext).color` for the text colors.

src/nav/ChatNavBar.js
WesleyAC: It turns out the `LoadingBanner` code is really janky, I think there probably shouldn't be a `default` here at all. But like you say, we're on our way to replacing it, so I'm inclined not to fix it.

WesleyAC:
@chrisbobbe Thanks for the review! Updated, should be ready for another look.

src/nav/ChatNavBar.js
chrisbobbe: > It turns out the `LoadingBanner` code is really janky, I think there probably shouldn't be a `default` here at all.

Yeah, possibly we could have done something better with `LoadingBanner` when it was added in #3901.

I don't think I've understood what you've identified as janky, though; should we not have ever used the brand color for the spinner (e.g., as [here](https://github.com/zulip/zulip-mobile/pull/4544#issuecomment-801634074) you've noticed we do), or should that case not have been given "default" semantics (e.g., bfb884a0ec131c9cc02ab664a18d446530c89a5c could have used a name like "brand" instead of "default), or something else? I think `LoadingIndicator`'s `color` prop appropriately takes a very few specific values because it makes a choice between a few specific assets.

src/nav/ChatNavBar.js
```diff
@@ -25,6 +25,9 @@ export default function ChatNavBar(props: Props) {
   const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
   const color =
```
chrisbobbe, line 26: As a small followup improvement:

Now that `color` isn't used for text anymore, it's just used for `<FooButton />`s, it'll be easier to give it a more specific name, which should help solidify its purpose. What do you think about `buttonColor`? Maybe something with "button," if we suppose the `<FooButton />`s are appropriately named, which I guess they are (though they make me think of `ZulipButton` which is quite different, hmm).

chrisbobbe: And another potential small follow-up improvement:

I think there should be a nice way to avoid repeating the `streamColor === undefined` conditional, which might make things a bit more readable.

Is this it something like this:

```js
  const streamColor = useSelector(state => getStreamColorForNarrow(state, narrow));
  const themeColor = useContext(ThemeContext).color;

  const [buttonColor, textColor, spinnerColor] =
    streamColor === undefined // <----- (conditional happens just once, here)
      ? [BRAND_COLOR, themeColor, 'default']
      : [
          foregroundColorFromBackground(streamColor),
          foregroundColorFromBackground(streamColor),
          foregroundColorFromBackground(streamColor),
        ];
```

Or, could use object destructuring instead of array; it'd be more verbose because we'd repeat `buttonColor`, `textColor`, and `spinnerColor` a few times.

chrisbobbe (approved):
Thanks for the revision! Just a few small comments; feel free to merge after seeing those.

src/nav/ChatNavBar.js
WesleyAC: I added the `color` → `buttonColor` commit to this, since that seems obviously better.

For the other thing, how would you feel about just moving `streamColor === undefined` into a named variable, and then keeping the three conditionals? That feels cleaner to me, but I'm open to being convinced otherwise.

chrisbobbe: Yeah, I think that'd be good.

gnprice:
Thanks @WesleyAC for spotting and fixing this, and @chrisbobbe for the reviews! This code all looks good to me, and the UI change is a distinct improvement.

src/nav/ChatNavBar.js
gnprice: I can't speak for what Wesley had in mind, but rereading `LoadingBanner` now, one thought I have is that it might be cleaner to have it take just the one parameter `backgroundColor`, and derive the others from that; it has all the information it needs to make the same choices `ChatNavBar` is making. As it stands, it's effectively an implicit part of its interface that those colors need to contrast and coordinate appropriately.

> I think `LoadingIndicator`'s `color` prop appropriately takes a very few specific values because it makes a choice between a few specific assets.

I agree with this reasoning. One further step that might be helpful -- might be helpful even with `LoadingBanner` on its way out, because `LoadingIndicator` has other callers -- is to give a name to that enum `'default' | 'black' | 'white'`. That'd let the same name be used in the types of `SpinningProgress`, `LoadingIndicator`, and `LoadingBanner`, helping make clear the linkage between that prop in their respective interfaces; it'd also provide a place to put a few words of jsdoc about what "default" means here.

Issue #4715, opened 2021-05-01
Stream settings screen doesn't make sense for non-subscribed streams


WesleyAC:
Currently, the stream settings screen shows options like "Pinned", "Muted", and "Notifications", even if you aren't subscribed to the stream. They do nothing when you tap them.

Instead, it should hide those options, and we should add a "Subscribe" button (which we don't have currently??? and amusingly, you can't add yourself via "Add subscribers", since `UserPicker` filters you out.)

PR #4718, opened 2021-05-02
Improve stream settings (especally while unsubscribed)

WesleyAC:
Fixes: #4715 

before (unsubscribed):

after (unsubscribed):

before (subscribed):

after (subscribed):


src/streams/StreamCard.js
```diff
@@ -36,6 +36,7 @@ export default class StreamCard extends PureComponent<Props> {
 
     const name = subscription.name || stream.name;
     const description = subscription.description || stream.description;
+    const subscribed = subscription !== NULL_SUBSCRIPTION;
```
chrisbobbe, line 39: I would probably write `isSubscribed` instead of `subscribed`, to be explicit that it's a boolean.

Hmm, also: is there another way to compute it that doesn't use `NULL_SUBSCRIPTION`? It's marked as deprecated, so I imagine there might be a scenario where we remove one or more uses of it and end up with a surprise here (where we think `subscription` might be `===` to it but it isn't).

src/streams/StreamIcon.js
```diff
@@ -8,12 +8,20 @@ type Props = $ReadOnly<{|
...
-  const StreamIcon = isMuted ? IconMute : isPrivate ? IconPrivate : IconStream;
+export default ({
+  color,
+  style,
+  isPrivate = false,
+  isMuted = false,
+  isSubscribed = true,
+  size,
+}: Props) => {
+  const StreamIcon = isMuted && !isMuted ? IconMute : isPrivate ? IconPrivate : IconStream;
```
chrisbobbe, line 24: I think you meant to use `isSubscribed` in here somewhere.

src/streams/StreamCard.js
chrisbobbe: (Same things on `StreamSettingsScreen` in a different commit.)

src/streams/StreamSettingsScreen.js
```diff
@@ -60,6 +62,11 @@ class StreamSettingsScreen extends PureComponent<Props> {
     NavigationService.dispatch(navigateToStreamSubscribers(stream.stream_id));
   };
 
+  handleSubscribe = () => {
```
chrisbobbe, line 65: How about a name that says that it's handling a request to add a subscription, to make it clear it's not for unsubscribing, or a toggle between them?

I have mixed feelings about the naming pattern that's developed for these methods we use as event handlers. On the one hand, I guess I can be glad that [most of them] say "handle", since that defends them against being haphazardly co-opted into use as something other than an event handler. At the same time, "ToggleMuteStream", "Edit", "EditSubscribers", etc., aren't the best descriptions for events that these things fire on. Rather, they fire on the user _requesting_ to toggle stream muting, to edit, to edit subscribers, etc., by doing the appropriate interaction. Possibly `handleRequestEdit` would be a reasonable name, but I'm not sure I'm up to enforcing that for anything much longer; `handleRequestToggleMuteStream` might obscure more than it clarifies. Do you have thoughts on this?

I think I'd be fine with keeping the general pattern but adding "add" in there somewhere, like `handleSubscriptionAdd` or something similar.

src/streams/StreamItem.js
```diff
@@ -48,13 +48,13 @@ type Props = $ReadOnly<{|
...
  * @prop isMuted - false for a Stream; !sub.in_home_view for Subscription
  * @prop isPrivate - .invite_only for a Stream or a Subscription
+ * @prop isSubscribed - whether the user is subscribed to the stream
  * @prop color - if provided, MUST be .color on a Subscription
  * @prop backgroundColor - if provided, MUST be .color on a Subscription
  *
  * @prop unreadCount - number of unread messages
  * @prop iconSize
  * @prop showSwitch - whether to show a toggle switch (ZulipSwitch)
- * @prop isSwitchedOn - initial value of the toggle switch, if present
```
chrisbobbe, line 57: 🎉

I'd put this `isSwitchedOn` -> `isSubscribed` change into its own NFC commit, so the pure refactor can stand and be verified on its own.

chrisbobbe (commented):
Thanks, @WesleyAC! I haven't yet run this and poked around with it yet, but I've left a few small comments from reading the code, below.

src/streams/StreamItem.js
WesleyAC: Good call, done.

src/streams/StreamIcon.js
WesleyAC: Good catch! Fixed.

src/streams/StreamCard.js
WesleyAC: Renamed! Will look into not using `NULL_SUBSCRIPTION`.

src/streams/StreamSettingsScreen.js
WesleyAC: How do you feel about `handleSubscribeClick`? `Click` is sort of anachronistic, but used a fair amount, so I don't think it'd be confusing.

src/streams/StreamCard.js
WesleyAC: Removed NULL_SUBSCRIPTION from `StreamSettingsScreen`. Still using `NULL_SUBSCRIPTION.color` in `StreamCard`, but I don't think that's a problem, that'll be easy to remove when we feel like removing `NULL_SUBSCRIPTION` altogether, and in the meantime, it's nice to have it in one place.

src/streams/StreamSettingsScreen.js
chrisbobbe: Yeah, probably fine, maybe better, to have it be explicit about the type of interaction the user uses to say they want to subscribe. Better because in cases where we actually use an event object passed to the handler, the event object will have details specific to the type of interaction.

Except my sense is that "press" is more idiomatic on mobile; `ZulipButton` and the things from RN that it uses have `onPress`. So, `handleSubscribePress` or `handlePressSubscribe` would be fine; I think I prefer the latter.

chrisbobbe (commented):
Thanks, @WesleyAC! A few comments below.

I think we might add `StreamSettingsScreen` to the list of things we could try doing in the style of #4489; what do you think?

src/streams/StreamItem.js
```diff
@@ -111,13 +117,13 @@ export default function StreamItem(props: Props) {
         <UnreadCount color={iconColor} count={unreadCount} />
         {showSwitch && (
           <ZulipSwitch
-            value={!!isSwitchedOn}
+            value={!!isSubscribed}
```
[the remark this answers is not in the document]
WesleyAC: I don't see that in practice, I just wrote `!!` out of habit in order to be defensive. Didn't realize about the typechecking hole, so I'm glad I did :)

src/streams/StreamItem.js
```diff
@@ -66,9 +66,9 @@ export default function StreamItem(props: Props) {
     backgroundColor,
     isPrivate = false,
     isMuted = false,
+    isSubscribed = true,
     iconSize,
     showSwitch = false,
-    isSwitchedOn = false,
```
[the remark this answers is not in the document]
WesleyAC: Well, I shouldn't have done that in a NFC commit, definitely. I'll fix that, and then add another commit to remove the default.

src/streams/StreamSettingsScreen.js
```diff
@@ -112,13 +119,22 @@ class StreamSettingsScreen extends PureComponent<Props> {
             secondary
             onPress={() => delay(this.handleEditSubscribers)}
           />
+          {!isSubscribed && (
+            <ZulipButton
+              style={styles.marginTop}
+              text="Subscribe"
```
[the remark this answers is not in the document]
WesleyAC: Thanks! Done :)

src/streams/StreamList.js
```diff
@@ -72,7 +72,7 @@ export default class StreamList extends PureComponent<Props> {
             unreadCount={unreadByStream[item.stream_id]}
             isMuted={item.in_home_view === false} // if 'undefined' is not muted
             showSwitch={showSwitch}
-            isSwitchedOn={item.subscribed}
+            isSubscribed={item.subscribed}
```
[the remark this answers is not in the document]
WesleyAC: Hmmm, interesting, thanks for catching that!

This is unfortunately sort of annoying to fix, since we've written `PseudoSubscription` as a enum, which is I think not actually what we want here. See [#mobile-team > Stream/Subscription split](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Stream.2FSubscription.20split) for my thoughts on the right approach to fixing that — but I'd prefer for that to be in a different PR, since I think it's somewhat involved.

src/streams/StreamSettingsScreen.js
WesleyAC: Ok, changed to `handlePressFoo`. Did this as a commit at the end, since that felt cleanest (given that I was updating existing code as well)

WesleyAC:
@chrisbobbe Ok, should be good to go!

chrisbobbe (commented):
Thanks! See a few comments below, all small, then please merge at will.

WesleyAC:
Made all those changes and rebased, thanks for the review! I'd appreciate you looking it over once more just to be sure, I think it'll be pretty quick.

src/unread/UnreadCards.js
```diff
@@ -50,6 +50,7 @@ export default function UnreadCards(props: Props) {
             iconSize={16}
             isMuted={section.isMuted}
             isPrivate={section.isPrivate}
+            isSubscribed
```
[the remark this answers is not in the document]
WesleyAC: Turns out `isSubscribed` was only being used when `showSwitch` was set to `true`, so I don't think we need to make it required.

Maybe we want to replace `isSubscribed` and `showSwitch` with a `subscribedSwitchValue?: boolean` or something, but in the meantime, just having a arbitrary default for `isSubscribed` seems fine.

chrisbobbe:
Thanks for the revision! Merged.

src/streams/StreamIcon.js
```diff
@@ -8,12 +8,13 @@ type Props = $ReadOnly<{|
...
   isMuted: boolean,
+  isSubscribed: boolean,
   size: number,
   style?: TextStyleProp,
 |}>;
 
-export default ({ color, style, isPrivate = false, isMuted = false, size }: Props) => {
-  const StreamIcon = isMuted ? IconMute : isPrivate ? IconPrivate : IconStream;
+export default ({ color, style, isPrivate, isMuted, isSubscribed, size }: Props) => {
+  const StreamIcon = isSubscribed && isMuted ? IconMute : isPrivate ? IconPrivate : IconStream;
```
gnprice, line 17: StreamIcon: Don't show as muted when unsubscribed.

This isn't necessary -- before this commit it was already the case that we didn't show a stream as muted when unsubscribed. There was a bug in this in StreamCard, but it was caused by the use of NULL_SUBSCRIPTION and was fixed in the commit that removed that, earlier in this branch.

So this change doesn't seem helpful.

It is kind of awkward that this component's callers are expected to compute `isMuted` as false when there's no subscription. A version of this change that moved that logic inside the component, probably by having it take `stream: Stream, subscription?: Subscription` instead of their individual properties, would improve things.

I'll send a PR with a revert.

gnprice (commented):

Some comments below. For two of them I have followups I'll send; the third is open.

Issue #4722, opened 2021-05-05
Play Store doesn't offer app when no camera on device


gnprice:
A user [reports](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Zebra.20Mobile.20Computer.20compatibility/near/1174306) that they can't find the app in the Play Store on a Zebra MC3300X, a mobile computer for warehouse that's running Android 10.

And indeed the Play Console [tells me this](https://play.google.com/console/developers/8060868091387311598/app/4976350040864490411/devices?deviceGroupId=Zebra-mc3300x) about that device:
> Reasons the device is unsupported
> Doesn't support required feature: android.hardware.camera

So it seems the Play Store believes our app requires a camera. Of course in reality the situation is that the app can *use* a camera if you have one, but the app is perfectly functional for its core use cases without one. So it seems that some metadata is incorrect.

This is puzzling because in the app's manifest, the closest thing to this that I can find is:
```xml
    <uses-permission
        android:name="android.permission.CAMERA"
        android:required="false" />
```
which seems like it should say that a camera is not required. But evidently something else is going on.

The next step for debugging this is probably some research in the Android docs and/or on the web, to understand just what that bit of the manifest means and to understand how the Play Store decides what hardware features an app requires.

Until we've fixed it, a workaround is to download the APK (e.g. [via GitHub](https://github.com/zulip/zulip-mobile/releases)) and install it manually.

andreasoc:
great!

PR #1843, opened 2018-01-30
[Android] Make camera & touch permissions optional.

jainkuniya:
To make sure support on Chromebooks.

#1834

PR #4724, opened 2021-05-05
Fix bug in AndroidManifest that prevented play store from displaying Zulip to certain user as an installable app.

AkashDhiman:
Fixes: #4722

All of the relevant documentation is given in https://developer.android.com/guide/topics/manifest/uses-feature-element, it explains how play store filtering works and what a developer can do to control that.

gist is explained in the commit message of https://github.com/zulip/zulip-mobile/commit/199b73b6a72fe4f3c74fdba6e5a305bea13b67c1 of this PR.

> Google Play filters which app to display to a user based on `uses-sdk`,
`uses-permission` and `uses-feature` tags.

> In case `uses-feature` is not declared `uses-permission` is used to do
this filtering, and by default it treats the permission it needs as a
compulsory requirement for the app. Hence our app will not show in play
store for devices without the hardware of concern.

> With `uses-feature` we override the filtering done by `uses-permission`
and by setting the required attribute to false we say that our app isn't
dependent on the hardware of concern to work properly (camera in case of
this commit).

**Why did I specifically add `android.hardware.camera` and `android.hardware.camera.autofocus` in `uses-feature`?**
Table 2 at https://developer.android.com/guide/topics/manifest/uses-feature-element#permissions gives a list which hardware requirements are affected by which specific permission. I am speculating here that without `android.hardware.camera.autofocus` an android device with camera but without autofocus hardware won't be shown our app as well.

I am wondering now what exactly will happen if a user without camera does press the button to access camera in our app, and if something like [checking hardware availability](https://developer.android.com/training/permissions/declaring#determine-hardware-availability) be required?

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -15,6 +15,15 @@
         android:name="android.hardware.touchscreen"
         android:required="false" />
 
+    <!-- The CAMERA permission causes these two camera-related features to be
+         implicitly treated as required, so we have to explicitly say they're not. -->
+    <uses-feature
+        android:name="android.hardware.camera"
+        android:required="false" />
+    <uses-feature
+        android:name="android.hardware.camera.autofocus"
+        android:required="false" />
+
     <application
         android:name=".MainApplication"
         android:allowBackup="true"
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,9 +8,7 @@
     <uses-permission
         android:name="android.permission.READ_PHONE_STATE"
         tools:node="remove" />
-    <uses-permission
-        android:name="android.permission.CAMERA"
-        android:required="false" />
+    <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature


gnprice:
Well spotted, thanks!

Looks like that ineffective `android:required="false"` on the `uses-permission` element was added in a836e7442 / #1843 (back in 2018), with the explanation that it was to make the app support Chromebooks. I wonder if anyone actually tested at the time that the change had any effect for Chromebooks.

> In case `uses-feature` is not declared `uses-permission` is used to do
> this filtering, and by default it treats the permission it needs as a
> compulsory requirement for the app.

Yuck. Thanks for debugging this.

I took a look at that table:
https://developer.android.com/guide/topics/manifest/uses-feature-element#permissions
and compared it to the permissions in our manifest, to see if there are any others we should take care of in the same way. It looks like there's one more: `ACCESS_WIFI_STATE`, which implies feature `android.hardware.wifi`.

* That one isn't in our own source code for the manifest, but I found it from looking through the manifest of the built app as installed on my device, using the handy [Dexplorer](https://play.google.com/store/apps/details?id=com.dexplorer) app.
* Dexplorer is good for browsing through all the different apps you have installed, but then if you open up our AndroidManifest.xml source file in Android Studio, and choose the "Merged Manifest" tab at the bottom, you get another view of the manifest of the built app:
  image
  including a handy cross-reference to what source file contributed any given bit of the manifest. So in that "Merging Log" section in the bottom right of that screenshot, we can see that three different libraries all asked for that permission in their own `AndroidManifest.xml` fragments: `@react-native-community/netinfo`, `rn-fetch-blob`, and Flipper. (The blue bits of text are links, and if you follow them it opens up the actual source files in question.)
* I definitely recommend opening up that view in Android Studio and browsing around it a bit. There turns out to be quite a lot in our manifest that isn't in our own `AndroidManifest.xml` source files (and comes instead from various libraries), and it's educational to look around and get an idea of what's there and where it comes from.

There probably are very few Android devices indeed that don't have wifi. But the app should work fine without it -- it just needs some other way to talk to the Internet. So as long as we're here, we should probably make that feature optional too. Want to do that as a quick followup?

---

> I am wondering now what exactly will happen if a user without camera does press the button to access camera in our app, and if something like [checking hardware availability](https://developer.android.com/training/permissions/declaring#determine-hardware-availability) be required?

Good question. I agree with Wesley that I'd sure hope that would be treated as if the user had denied permission. I'm not finding docs addressing that question, but there is a mention here:
https://developer.android.com/training/permissions/requesting#handle-denial
that "[i]n certain situations, the permission might be denied automatically, without the user taking any action." Hopefully this is just one of those situations.

I think ideally we would indeed check hardware availability, and just not show the "camera" icon in the first place if there isn't a camera to use. Not a priority, though -- it'd be just a small polish issue, and only for very few users. If we were writing against the native Android SDK in the first place, I'd want to do that (because it'd be just a couple of lines of code); but because we'd have to wire that up through RN to the JS side, I think it won't be worth it.

gnprice:
OK, and merged! Thanks again.

I made a few small changes before merge:
* Added a short comment explaining the new `uses-feature` elements; also deleted the blank line between them, in order to visually group them together and group them with the comment.
* Added a bit more to the first commit message.
* s/redundant/meaningless/ in the first commit message -- I would interpret "redundant" as meaning that there was already something else doing the same job, whereas the story here is more like that this attribute can't possibly do anything in the first place, and in fact nothing was successfully doing the job this attribute was intended to do.

See also the suggested followup above about feature `android.hardware.wifi`.

PR #4728, opened 2021-05-07
intl: Highlight in dev user-facing strings not added to messages_en.json.

chrisbobbe:
As discussed around
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Warn.20if.20user-facing.20strings.20untreated/near/1156727,
this seems pretty low-risk and high-reward.

-----

Also a follow-up commit that removes some places where we were passing `defaultMessage` to a `FormattedMessage` component. Those aren't suppressing any errors; I did check that the strings there were all in messages_en.json. But our habit has been to use `Label` instead of using `FormattedMessage` directly, so, do that.

### DIFF
--- a/src/boot/TranslationProvider.js
+++ b/src/boot/TranslationProvider.js
@@ -36,7 +36,18 @@ export function withGetText<P: { +_: GetText, ... }, C: ComponentType<P>>(
 
 const makeGetText = (intl: IntlShape): GetText => {
   const _ = (message, values) =>
-    intl.formatMessage({ id: message, defaultMessage: message }, values);
+    intl.formatMessage(
+      {
+        id: message,
+
+        // If you see this in dev, it means there's a user-facing
+        // string that hasn't been added to
+        // static/translations/messages_en.json. Please add it! :)
+        defaultMessage:
+          process.env.NODE_ENV === 'development' ? `UNTRANSLATED—${message}—UNTRANSLATED` : message,
+      },
+      values,
+    );
   _.intl = intl;
   return _;
 };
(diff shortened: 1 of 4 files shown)


WesleyAC:
Looks good overall!

Is there a way that we could make this error more visible? For me, it only shows up in the console log (no popup on the screen), and the error message is pushed off my screen immediately by the traceback. Could we make it so there's a popup with the error, or maybe have the fallback string be "UNTRANSLATED STRING" or something like that? Maybe that's too obnoxious when developing, but I think I might have missed this message as it is.

I guess maybe a good solution would be to put error messages below the stacktrace in general, although I don't know how involved that is.

I'm also happy to merge this as-is, since it's an improvement on the status quo even without changing how the error is surfaced, just thought I'd bring up making it more visible.

WesleyAC (approved):

chrisbobbe:
I'm not sure how we'd go about making a popup for it (maybe a place to start is to look around in react-intl's internals), and I'm not sure about how to put error messages below the stacktrace. In my setup, the stacktrace is a clear signal that something's wrong because it's included in the block of red in Chrome devtools:

image

and I just have to scroll up to see the error.

Still, if I'm not doing the kind of work that I expect to give an error (maybe I'm changing straight quotes to curly quotes in a user-facing string, I dunno), then I won't necessarily have the devtools open at all.

I think having the default be "UNTRANSLATED STRING" is a good solution. I think that'll mean that we don't get any errors in the console, but that seems OK; it'll ~never be the case that you're investigating some UI thing with the console visible and the app hidden away somewhere. What do think of including the string (so it's clear which is the offending one) but prefixed with `UNTRANSLATED STRING: `?

Ah, or maybe prefixed with `UNTRANSLATED—` and suffixed with `—UNTRANSLATED`, in case for whatever reason the first part of the string is truncated?

WesleyAC (approved):
Hmm, it's a shame that react-intl doesn't give us the API we need to handle this well, but I think the `UNTRANSLATED-<string>-UNTRANSLATED` solution is a reasonable one. LGTM!

chrisbobbe:

PR #4730, opened 2021-05-07
Muted users placeholder in reactions screen

WesleyAC:
This commit shows a placeholder for the name/avatar of a muted user in the reactions list (and anywhere else that uses `UserItem`, but I think we've filtered out muted users in most of the other places `UserItem` shows up).

I've pulled in a modified version of a FontAwesome icon for this, and I'm not sure exactly what attribution is required for it — would love to hear thoughts on that.

Screenshot_20210507-142257
Screenshot_20210507-142233

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -8,6 +8,7 @@ import type { FeatherGlyphs } from 'react-native-vector-icons/Feather';
 import IoniconsIcon from 'react-native-vector-icons/Ionicons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
 import SimpleLineIcons from 'react-native-vector-icons/SimpleLineIcons';
+import FontAwesome from 'react-native-vector-icons/FontAwesome';
 
 /**
  * The props actually accepted by icon components in r-n-vector-icons.
@@ -85,3 +86,4 @@ export const IconMoreHorizontal = makeIcon(Feather, 'more-horizontal');
 export const IconEdit = makeIcon(Feather, 'edit');
 export const IconPlusSquare = makeIcon(Feather, 'plus-square');
 export const IconVideo = makeIcon(Feather, 'video');
+export const IconUserMuted = makeIcon(FontAwesome, 'user');
--- a/src/common/UserAvatar.js
+++ b/src/common/UserAvatar.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import React, { type Node as React$Node } from 'react';
+import React, { type Node as React$Node, useContext } from 'react';
 import { Image, View, PixelRatio } from 'react-native';
 
 import { useSelector } from '../react-redux';
@@ -7,10 +7,13 @@ import { getAuthHeaders } from '../api/transport';
 import { tryGetAuth } from '../account/accountsSelectors';
 import Touchable from './Touchable';
 import { AvatarURL, FallbackAvatarURL } from '../utils/avatar';
+import { IconUserMuted } from './Icons';
+import { ThemeContext } from '../styles';
 
 type Props = $ReadOnly<{|
   avatarUrl: AvatarURL,
   size: number,
+  isMuted?: boolean,
   children?: React$Node,
   onPress?: () => void,
 |}>;
@@ -24,13 +27,20 @@ type Props = $ReadOnly<{|
  * @prop [onPress] - Event fired on pressing the component.
  */
 function UserAvatar(props: Props) {
-  const { avatarUrl, children, size, onPress } = props;
+  const { avatarUrl, children, size, isMuted = false, onPress } = props;
   const borderRadius = size / 8;
   const style = {
     height: size,
     width: size,
     borderRadius,
   };
+  const iconStyle = {
+    height: size,
+    width: size,
+    textAlign: 'center',
+  };
+
+  const { color } = useContext(ThemeContext);
 
   const auth = useSelector(state => tryGetAuth(state));
   if (!auth) {
@@ -48,16 +58,20 @@ function UserAvatar(props: Props) {
     <View>
       <Touchable onPress={onPress}>
         <View>
-          <Image
-            source={{
-              uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
-              ...(avatarUrl instanceof FallbackAvatarURL
-                ? { headers: getAuthHeaders(auth) }
-                : undefined),
-            }}
-            style={style}
-            resizeMode="cover"
-          />
+          {!isMuted ? (
+            <Image
+              source={{
+                uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
+                ...(avatarUrl instanceof FallbackAvatarURL
+                  ? { headers: getAuthHeaders(auth) }
+                  : undefined),
+              }}
+              style={style}
+              resizeMode="cover"
+            />
+          ) : (
+            <IconUserMuted size={size} color={color} style={iconStyle} />
+          )}
           {children}
         </View>
       </Touchable>
--- a/src/common/UserAvatar.js
+++ b/src/common/UserAvatar.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 import React, { type Node as React$Node } from 'react';
-import { ImageBackground, View, PixelRatio } from 'react-native';
+import { Image, View, PixelRatio } from 'react-native';
 
 import { useSelector } from '../react-redux';
 import { getAuthHeaders } from '../api/transport';
@@ -46,21 +46,20 @@ function UserAvatar(props: Props) {
 
   return (
     <View>
-      <Touchable onPress={onPress} style={style}>
-        <ImageBackground
-          style={style}
-          source={{
-            uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
-            ...(avatarUrl instanceof FallbackAvatarURL
-              ? { headers: getAuthHeaders(auth) }
-              : undefined),
-          }}
-          resizeMode="cover"
-          /* ImageBackground seems to ignore `style.borderRadius`. */
-          borderRadius={borderRadius}
-        >
+      <Touchable onPress={onPress}>
+        <View>
+          <Image
+            source={{
+              uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
+              ...(avatarUrl instanceof FallbackAvatarURL
+                ? { headers: getAuthHeaders(auth) }
+                : undefined),
+            }}
+            style={style}
+            resizeMode="cover"
+          />
           {children}
-        </ImageBackground>
+        </View>
       </Touchable>
     </View>
   );
--- a/src/common/UserAvatar.js
+++ b/src/common/UserAvatar.js
@@ -55,27 +55,25 @@ function UserAvatar(props: Props) {
   }
 
   return (
-    <View>
-      <Touchable onPress={onPress}>
-        <View>
-          {!isMuted ? (
-            <Image
-              source={{
-                uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
-                ...(avatarUrl instanceof FallbackAvatarURL
-                  ? { headers: getAuthHeaders(auth) }
-                  : undefined),
-              }}
-              style={style}
-              resizeMode="cover"
-            />
-          ) : (
-            <IconUserMuted size={size} color={color} style={iconStyle} />
-          )}
-          {children}
-        </View>
-      </Touchable>
-    </View>
+    <Touchable onPress={onPress}>
+      <View>
+        {!isMuted ? (
+          <Image
+            source={{
+              uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
+              ...(avatarUrl instanceof FallbackAvatarURL
+                ? { headers: getAuthHeaders(auth) }
+                : undefined),
+            }}
+            style={style}
+            resizeMode="cover"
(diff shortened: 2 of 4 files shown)


chrisbobbe:
> I've pulled in a modified version of a FontAwesome icon for this, and I'm not sure exactly what attribution is required for it — would love to hear thoughts on that.

Hmm, interesting; how is it modified? It looks pretty close to the original, to me.

In fact, could we just make an `IconUserSlash` component in src/common/Icons.js (RNVI supports FontAwesome I think) and use that?

WesleyAC:
It's modified in size and color, but the path is the same. Perhaps I should go with the Material `person_off` icon, which seems to have a completely free license, and we're using other material icons already. Unfortunately, I don't think RNVI will help us here, since the `UserAvatar` needs to be able to render child components in the icon component (for the presence indicators) — the RNVI `Icon` class doesn't support that, I don't think

WesleyAC:
Ok, switched to the material icon.

chrisbobbe:
> Unfortunately, I don't think RNVI will help us here, since the `UserAvatar` needs to be able to render child components in the icon component (for the presence indicators) — the RNVI `Icon` class doesn't support that, I don't think

Hmm, yeah, I think I see what you mean:

- `UserAvatar` wants to pass the `children` it receives—i.e., `foo` in `<UserAvatar>{foo}</UserAvatar>`—as the `children` of an `ImageBackground`.
- `Icon` has no interface for callers to pass `children`.

Looking at the first point, `ImageBackground`'s [implementation](https://github.com/facebook/react-native/blob/v0.63.4/Libraries/Image/ImageBackground.js#L60-L86) is essentially this. The component is introduced with, `Very simple drop-in replacement for <Image> which supports nesting views`:

```js
    return (
      <View {...}>
        <Image {...} />
        {children}
      </View>
    );
```

On the second point, here's what RNVI uses in its `Icon` class (in lib/create-icon-set.js, [here](https://github.com/oblador/react-native-vector-icons/blob/v7.0.0/lib/create-icon-set.js#L91-L96)), which I think is the base for all our `IconFoo`s:

```js
      return (
        <Text {...props}>
          {glyph}
          {children}
        </Text>
      );
```

So, it looks like it *does* take `children`. But it does have an interesting difference with `ImageBackground` in what it does with the `children` it receives—it puts `children` as the children of a `Text`, not a `View`.

`View`s can take "children of any type" ([doc](https://reactnative.dev/docs/0.63/view)). `Text`..."supports nesting" ([doc](https://reactnative.dev/docs/0.63/text)), but I think you're just meant to use strings and other `Text`s as its children? I'd expect it not to work if we tried passing a `PresenceStatusIndicator` as `children` to an `IconUserSlash`. Hmm.

Still, I could imagine using our knowledge of `ImageBackground`'s implementation to have `UserAvatar` change from using that to something like

```js
        <View {...}>
          {isMuted ? <IconUserSlash {...} /> : <Image {...} />}
          {children}
        </View>
```

If it turned out to be preferable to maintain the pattern of using `IconFoo` for icons.

chrisbobbe:
I'm also not that sure that `UserAvatar` should be using `children` at all; it's not like we ever pass anything other than a <PresenceStatusIndicator />` there.

chrisbobbe:
> If it turned out to be preferable to maintain the pattern of using `IconFoo` for icons.

One thing that's in `IconFoo`'s favor, besides consistency, is that using PNGs means tighter restraints on adjusting color and size.

WesleyAC:
Ok, switched to the `Icon` approach. Using the FontAwesome icons, since the version of RNVI we're on doesn't have the material `person_off` icon :disappointed:

alya:
@timabbott just merged a PR from @abhijeetbodas2001 for recent topics that uses a "user" icon (no slash) for muted users in **Recent topics**. It was chosen because the version of Font Awesome that we are using does not have the slashed icon.

Maybe we should go with the same "user" icon (no slash) for consistency on mobile?

src/users/UserItem.js
```diff
@@ -63,16 +66,17 @@ export function UserItemRaw<UserT: { user_id: UserId, email: string, full_name:
...
           size={48}
           userId={user.user_id}
+          isMuted={isMuted}
           onPress={onPress && handlePress}
         />
         <View style={componentStyles.textWrapper}>
           <RawLabel
             style={[componentStyles.text, isSelected && componentStyles.selectedText]}
-            text={user.full_name}
+            text={isMuted ? 'Muted User' : user.full_name}
```
alya, line 75: Same capitalization comment as in the other PR -- "Muted user" would be more consistent with web app.

src/common/Icons.js
```diff
@@ -8,6 +8,7 @@ import type { FeatherGlyphs } from 'react-native-vector-icons/Feather';
 import IoniconsIcon from 'react-native-vector-icons/Ionicons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
 import SimpleLineIcons from 'react-native-vector-icons/SimpleLineIcons';
+import FontAwesome5 from 'react-native-vector-icons/FontAwesome5';
```
chrisbobbe, line 11:
```
avatars [nfc]: Add muted user placeholder avatars.
```

nit: this commit isn't NFC 🙂

chrisbobbe: I got an error with "Unrecognized font family" on iOS, so I followed a [troubleshooting entry](https://github.com/oblador/react-native-vector-icons#troubleshooting) and saw that we hadn't added some things to our Info.plist. This was sufficient to get it working:

```diff
--- ios/ZulipMobile/Info.plist
+++ ios/ZulipMobile/Info.plist
@@ -63,6 +63,9 @@
                <string>Entypo.ttf</string>
                <string>EvilIcons.ttf</string>
                <string>FontAwesome.ttf</string>
+               <string>FontAwesome5_Brands.ttf</string>
+               <string>FontAwesome5_Regular.ttf</string>
+               <string>FontAwesome5_Solid.ttf</string>
                <string>Foundation.ttf</string>
                <string>Ionicons.ttf</string>
                <string>Feather.ttf</string>
```

src/common/UserAvatar.js
```diff
@@ -48,16 +53,20 @@ function UserAvatar(props: Props) {
...
+                uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
+                ...(avatarUrl instanceof FallbackAvatarURL
+                  ? { headers: getAuthHeaders(auth) }
+                  : undefined),
+              }}
+              style={style}
+              resizeMode="cover"
+            />
+          ) : (
+            <IconUserMuted size={size} color={color} />
```
chrisbobbe, line 68: I notice that the Font Awesome icon isn't square; it looks like its height is getting set to 48px while its width is greater, which disturbs the alignment with the 48x48 avatars:

image

I guess that's a design choice that Font Awesome has made, hmm. My sense from [a bit of reading](https://material.io/design/iconography/system-icons.html#design-principles) is that some icon libraries use a square for the boundaries of each icon? Alya pointed out that we'll probably want to use a user icon without the slash. That may let us choose an icon from a different library, where they're all nice and square.

Being short and wide is more inconvenient than being tall and narrow. The underlying component that gets used is a `Text`, as we've discussed elsewhere in this PR. I don't yet know how to scale down the contents of a `Text` by specifying a max width, letting the height take a computed value, without cutting stuff off. The `size` prop (which is used as the `Text`'s `style.fontSize`) seems to control the height. If the icon were tall and narrow (from some experimentation, I think Font Awesome's ['user' icon](https://fontawesome.com/icons/user) is this way), I think we could set the height to 48 using `size` and just make sure it's centered in a 48x48 square (I have some working code for this centering if you'd like me to share).

src/common/UserAvatar.js
```diff
@@ -46,21 +46,20 @@ function UserAvatar(props: Props) {
...
-            ...(avatarUrl instanceof FallbackAvatarURL
-              ? { headers: getAuthHeaders(auth) }
-              : undefined),
-          }}
-          resizeMode="cover"
-          /* ImageBackground seems to ignore `style.borderRadius`. */
-          borderRadius={borderRadius}
-        >
+      <Touchable onPress={onPress}>
+        <View>
```
chrisbobbe, line 55: Makes sense. At first, I thought we could just remove this `View`, but then I got the `React.Children.only expected to receive a single React element child` from somewhere in `Touchable`.

I think a nice independent cleanup would be to remove the outer, unstyled `<View />` that wraps the `<Touchable />` in this component; that doesn't seem to be doing anything important, and I poked around without it and things looked just the same as before.

src/common/Icons.js
```diff
@@ -84,3 +85,4 @@ export const IconMoreHorizontal = makeIcon(Feather, 'more-horizontal');
 export const IconEdit = makeIcon(Feather, 'edit');
 export const IconPlusSquare = makeIcon(Feather, 'plus-square');
 export const IconVideo = makeIcon(Feather, 'video');
+export const IconUserMuted = makeIcon(FontAwesome5, 'user-slash');
```
chrisbobbe, line 88: If we end up using a `FontAwesome5` icon, and if you're interested, I have some working code for letting Flow know that the component can take `brand`, `light`, and `solid` props: https://github.com/oblador/react-native-vector-icons/blob/master/FONTAWESOME5.md#usage

src/users/UserItem.js
chrisbobbe: Ah, and: this 'Muted User' text won't be translated, even with an entry in messages_en.json. 😛

I think `RawLabel` was chosen so that people's full names don't get localized; note its jsdoc:

```
Unlike `Label` it does not translate its contents.
```

The solution is probably to do `_('Muted user')`, where `_` is `useContext(TranslationContext)`.

chrisbobbe (changes requested):
Thanks, @WesleyAC!

See a few comments below. I'm glad it turned out to be easy to switch to the `IconFoo` approach. I do wonder whether that's still the right approach, given my notes below on the `FontAwesome` icon's dimensions, hmm. I wonder if we should explore something like [`getImageSourceSync`](https://github.com/oblador/react-native-vector-icons#static-methods) (or make a PNG manually as you did in an earlier revision) or if it'd be fine to just choose a square or taller-than-wide icon.

src/users/UserItem.js
WesleyAC: Fixed :)

src/common/Icons.js
WesleyAC: Shouldn't be needed now that I'm not using FontAwesome5 (see [CZO thread](https://chat.zulip.org/#narrow/stream/101-design/topic/Hiding.20user-muted.20messages))

src/common/UserAvatar.js
WesleyAC: Good catch! `user` has the same problem but with the reverse aspect ratio, I've fixed it by centering the text, let me know if you'd prefer a different approach.

WesleyAC:
Should be ready for another review!

chrisbobbe:
Thanks! Merged, with the small cleanup I mentioned at https://github.com/zulip/zulip-mobile/pull/4730#discussion_r631345432.

src/common/UserAvatar.js
```diff
@@ -45,24 +55,25 @@ function UserAvatar(props: Props) {
   }
 
   return (
-    <View>
-      <Touchable onPress={onPress} style={style}>
-        <ImageBackground
```
gnprice, line 50: The commit message says

    This will allow us to use an `Icon` in place of a avatar URL in the
    future. The implementation of `ImageBackground` is very simple, we
    essentially inline it.

But when I look at the implementation in `node_modules/react-native/Libraries/Image/ImageBackground.js`, it seems like there are some subtleties there that aren't carried through in this version.

* The upstream implementation has an `accessibilityIgnoresInvertColors={true}`. I'm not sure what that does; is it not needed for our case?
* The upstream implementation provides the following style props:
```js
{
  position: 'absolute',
  left: 0,
  right: 0,
  top: 0,
  bottom: 0,
}
```
(under the name of `StyleSheet.absoluteFill`). What's the effect of those?

src/common/UserAvatar.js
```diff
@@ -45,24 +55,25 @@ function UserAvatar(props: Props) {
...
-          style={style}
-          source={{
-            uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
-            ...(avatarUrl instanceof FallbackAvatarURL
-              ? { headers: getAuthHeaders(auth) }
-              : undefined),
-          }}
-          resizeMode="cover"
-          /* ImageBackground seems to ignore `style.borderRadius`. */
-          borderRadius={borderRadius}
```
gnprice, line 60: In the old code, this translated to a `borderRadius={borderRadius}` on the `Image`. Was dropping that intentional?

gnprice (commented):

I have a few comments below just on the first commit, which cut out the RN `ImageBackground` component. Most particularly I'm a bit concerned that the a11y-related prop which was left out would mean a regression in the experience for people using certain a11y settings.

src/common/UserAvatar.js
WesleyAC: Re: `accessibilityIgnoresInvertColors` ­— this apparently overrides the user's "Invert Colors" setting for just this view (see the [Apple docs](https://developer.apple.com/documentation/uikit/uiview/2865843-accessibilityignoresinvertcolors)). I guess it makes sense to keep this, I'll do that in a followup commit.

I found that setting the `width` and `height` was sufficient to cause the avatar to take up the required space, so I omitted the options in `absoluteFill` to keep things simple. It's possible there is a circumstance that causes them to be necessary, but if so, I don't think it appears in our app.

src/common/UserAvatar.js
WesleyAC: We apply the border radius via `style`, so doing it here is redundant. I probably should have done that as a separate commit, though.

PR #4731, opened 2021-05-07
watchman [nfc]: Remove empty config file.

WesleyAC:
According to upstream (https://github.com/facebook/watchman/issues/463),
the .watchmanconfig file is not needed, in general, and we have no
special configuration for it.

This file dates back to the initial commit in this repo, and I suspect
was added by accident as part of some template.


chrisbobbe:
Merged, thanks!

gnprice:
Thanks for tracking that down and this cleanup!

Issue #4723, opened 2021-05-05
Logout gives endless loading screen


gnprice:
This seems to reproduce reliably for me, on v27.162, on both iOS and Android. Not sure if it's new in v27.162 or was present earlier.

Repro steps:
* I'm logged into an account.
* I hit the "Log out" button. Get a confirmation modal, and confirm.

Expected result:
* Navigate to the account-pick screen. During the nav transition animation, the old account's UI is replaced with a loading screen.

Actual result:
* See just the loading screen -- no nav transition happening.
* [Sometimes](https://chat.zulip.org/#narrow/stream/48-mobile/topic/login.20bug/near/1174900), this goes on apparently indefinitely, at least for a good few seconds, until I kill the app. (Ken Clary [observed](https://chat.zulip.org/#narrow/stream/48-mobile/topic/logout.20bug/near/1173539) this continuing for 30 seconds or so before killing the app.)
* [Sometimes](https://chat.zulip.org/#narrow/stream/48-mobile/topic/logout.20bug/near/1173143), it turns into an error screen, the one our new React error boundary displays. The error message is "Error: Active account not logged in".

We [probably want to change](https://chat.zulip.org/#narrow/stream/48-mobile/topic/logged-out.20state/near/1174548) what the "Log out" button does entirely. But that isn't a trivial change to make, and I suspect that taking care of it will involve basically debugging this issue anyway -- the key thing that should be happening here and doesn't seem to be is the navigation, and we'll still want to make the same navigation after that change.

chrisbobbe:
From a bisect (luckily starting with the first and last commits in #4603) I think a80b4e8da3e7918604430e8ff3ef9b627e7d0396 may have introduced the issue; I'm looking into it.

PR #4733, opened 2021-05-07
Fix: Logout gives endless loading screen

chrisbobbe:
Fixes: #4723

### DIFF
--- a/src/nav/AppNavigator.js
+++ b/src/nav/AppNavigator.js
@@ -42,7 +42,7 @@ import EmojiPickerScreen from '../emoji/EmojiPickerScreen';
 import LegalScreen from '../settings/LegalScreen';
 import UserStatusScreen from '../user-status/UserStatusScreen';
 import SharingScreen from '../sharing/SharingScreen';
-import withHaveServerDataGate from '../withHaveServerDataGate';
+import { useHaveServerDataGate } from '../withHaveServerDataGate';
 
 export type AppNavigatorParamList = {|
   'account-pick': RouteParamsOf<typeof AccountPickScreen>,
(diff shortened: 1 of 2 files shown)


WesleyAC (approved):
LGTM, thanks!

chrisbobbe:
Thanks for the review! I'd like @gnprice to take a look as well, as I think he said on the call that this diagnosis was surprising or at least not as precise as it could be (which, it's true, it's not super precise).

WesleyAC:
Did we end up talking about this again? Somehow I thought we made progress towards merging this, are we still waiting on @gnprice?

chrisbobbe:
> Did we end up talking about this again? Somehow I thought we made progress towards merging this, are we still waiting on @gnprice?

Yes, I think we chatted a bit on a video call but Greg hasn't yet had a chance to review the code in more detail.

gnprice:
Thanks @chrisbobbe for this fix (and the underlying diagnosis)!

It's still kind of mysterious to me how this bug causes the particular symptoms we're seeing -- it feels like there's a bug in react-navigation involved here. But the fact that we're making a new `ComponentType` value, for each of these screens, in each render of `AppNavigator` is definitely a bug in our code, and this is a good fix.

I think the boilerplate is avoidable, though. I've pushed an additional commit onto this branch that demonstrates how, by adding a React hook `useHaveServerDataGate` that just factors out what your fix does for each of these screens.

If that looks good to you, please go ahead and merge after fixing up the branch -- I think either squashing the commits together, or putting a commit that adds `useHaveServerDataGate` first and then a commit that uses it, would be fine.

chrisbobbe:
Thanks for the review!

> If that looks good to you, please go ahead and merge after fixing up the branch -- I think either squashing the commits together, or putting a commit that adds `useHaveServerDataGate` first and then a commit that uses it, would be fine.

Done.

Issue #4734, opened 2021-05-07
Presence indicators always have white background


WesleyAC:
Presence indicators always have a white background. Instead, the background color should follow the theme background color, as on the webapp.

PR #4735, opened 2021-05-07
presence: Use theme background color.

WesleyAC:
Fixes: #4734

### DIFF
--- a/src/common/PresenceStatusIndicator.js
+++ b/src/common/PresenceStatusIndicator.js
@@ -1,10 +1,10 @@
 /* @flow strict-local */
-import React, { PureComponent } from 'react';
+import React, { PureComponent, useContext } from 'react';
 import { View } from 'react-native';
 import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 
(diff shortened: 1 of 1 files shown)


WesleyAC:
before:

Screenshot_20210508-032452

after:

Screenshot_20210508-032312

src/common/PresenceStatusIndicator.js
```diff
@@ -67,11 +66,13 @@ function MaybeOpaqueBackgroundWrapper(
   |}>,
 ) {
   const { useOpaqueBackground, style, children } = props;
+  const context = useContext(ThemeContext);
```
chrisbobbe, line 69: tiny nit:

I've more often skipped defining `context`, and done something like this instead:

```js
const { backgroundColor } = useContext(ThemeContext);
```

Or I guess if that conflicts with something, give it a different name like

```js
const { backgroundColor: themeBackgroundColor } = useContext(ThemeContext);
```

What do you think?

chrisbobbe (approved):
LGTM, thanks!

src/common/PresenceStatusIndicator.js
WesleyAC: Good call, done :)

PR #4736, opened 2021-05-07
Muted Users: Hide name in group PM

WesleyAC:


static/translations/messages_en.json
```diff
@@ -208,6 +208,7 @@
   "Reactions": "Reactions",
   "No reactions": "No reactions",
   "See who reacted": "See who reacted",
+  "Muted User": "Muted User",
```
alya, line 211: The webapp uses sentence case ("Muted user") for all translatable strings, so we should probably do that here for consistency.

src/pm-conversations/GroupPmConversationItem.js
```diff
@@ -1,10 +1,13 @@
 /* @flow strict-local */
-import React from 'react';
+import React, { useContext } from 'react';
 import { View } from 'react-native';
+import { useSelector } from 'react-redux';
```
chrisbobbe, line 5:
```suggestion

import { useSelector } from '../react-redux';
```

Our "type wrapper" gives a bit stronger type checking on uses of `useSelector`, so I've made a habit of using it.

`mutedUsers` does end up with the correct type in this case, but that's because the type of `getMutedUsers` is aware of our `GlobalState`. If we tried to do `const foo = useSelector(state => state.foo)` without using the type wrapper, then I think

1. the `.foo` access wouldn't get an error, and
2. `foo` wouldn't be type-checked anywhere it's used.

chrisbobbe (approved):
Thanks, @WesleyAC!

One small comment below; feel free to merge after seeing that and Alya's comment. Also, here's what it looks like for me:

image

I think ideally there'd be a marker in the (visual and screen-reader) interface to make it clear that it's actually a muted user, and not someone whose full name is "Muted user". I notice that the web app doesn't do this, so it's maybe better to wait for that to happen and then follow its lead.

(In fact it looks like it doesn't even hide the names of muted users in group PMs: .)

alya:
Not hiding the name in group PMs in the left side bar in the web app is definitely a bug -- good eye @chrisbobbe ! I posted to https://chat.zulip.org/#narrow/stream/101-design/topic/Hiding.20user-muted.20messages/near/1179851.

WesleyAC:
Made the requested changes, going ahead and merging.

gnprice:
Thanks @WesleyAC and all!

I've just pushed a couple of small fixes to these changes:

9e27f2329 pm-conversations: Use `useCallback` for an event handler.
49a931641 pm-conversations [nfc]: Fix grouping of imports.

chrisbobbe:
```
The solution to that problem is `useCallback`.  Generally we always
want to use that when translating a constructor-initialized callback
like this onto a function component.
```

Noted, thanks!

Issue #4604, opened 2021-04-05
Stop having `RealmInputScreen` flash by between `AccountPickScreen` and `AuthScreen`.


chrisbobbe:
Planned followup to #4491. (The other piece of followup, #4579, has been done.)

> * In `AccountPickScreen` we should not be navigating to `RealmScreen`, only to have that screen flash by -- that's not good UX. The code that does that is quoted in [#4415 (comment)](https://github.com/zulip/zulip-mobile/issues/4415#issuecomment-760546036) . Instead we should navigate directly to `AuthScreen` if the account is not logged in.
>   
>   * This may require some further refactoring to pull relevant logic out of `RealmScreen`. It will require careful testing that things continue to work in all kinds of scenarios.

PR #4740, opened 2021-05-10
Fix: Stop having `RealmInputScreen` flash by between `AccountPickScreen` and `AuthScreen`. 

chrisbobbe:
Fixes: #4604

### DIFF
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -3,6 +3,7 @@
 import React, { PureComponent } from 'react';
 import { Alert } from 'react-native';
 
+import * as api from '../api';
 import { TranslationContext } from '../boot/TranslationProvider';
 import type { RouteProp } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
@@ -13,7 +14,14 @@ import { hasAuth, getAccountStatuses } from '../selectors';
 import type { AccountStatus } from './accountsSelectors';
 import { Centerer, ZulipButton, Logo, Screen, ViewPlaceholder } from '../common';
 import AccountList from './AccountList';
-import { navigateToRealmInputScreen, accountSwitch, removeAccount } from '../actions';
+import {
+  navigateToRealmInputScreen,
+  accountSwitch,
+  removeAccount,
+  navigateToAuth,
+} from '../actions';
+import type { ApiResponseServerSettings } from '../api/settings/getServerSettings';
+import { showErrorAlert } from '../utils/info';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'account-pick'>,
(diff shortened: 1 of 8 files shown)


src/account/AccountPickScreen.js
```diff
@@ -28,15 +36,21 @@ class AccountPickScreen extends PureComponent<Props> {
...
       setTimeout(() => {
         dispatch(accountSwitch(index));
       });
     } else {
-      NavigationService.dispatch(navigateToRealmInputScreen({ realm }));
+      try {
+        const serverSettings: ApiResponseServerSettings = await api.getServerSettings(realm);
+        NavigationService.dispatch(navigateToAuth(serverSettings));
+      } catch (e) {
+        showErrorAlert(_('Failed to connect to realm: {realm}', { realm: realm.toString() }));
```
WesleyAC, line 52: AFAIK we don't want to use "realm" in user-facing strings, switch this to "organization"?

WesleyAC (commented):
One small thing, looks good other than that!

src/account/AccountPickScreen.js
chrisbobbe: Oh, really? I didn't know that; I've been careful to say "realm" for the same reason I'm careful to say "stream" instead of "channel"; I thought it was a product decision. But I guess it could be an outdated and non-user-facing thing, like the way we say "topic" to users even though some legacy parts of the code say "subject". Has there been a discussion on this somewhere, or maybe @alya has more context?

alya: I don't know the history, but "organization" is indeed the term we use on all the help pages (e.g. https://zulip.com/help/join-a-zulip-organization) and in the settings menu, so we should stay consistent with that in user-facing strings.

chrisbobbe: Will do, thanks!

gnprice: Yep, and we have this documented in the mobile repo's glossary:
https://github.com/zulip/zulip-mobile/blob/master/docs/glossary.md#realm

gnprice: In this context, probably "server" is better yet: that's the thing we're failing to connect to.

gnprice, line 52: This looks to be a new user-facing string, so it should go in the `messages_en.json` database for translation.

src/nav/navActions.js
```diff
@@ -62,12 +62,8 @@ export const navigateToGroupDetails = (
   recipients: $ReadOnlyArray<UserId>,
 ): GenericNavigationAction => StackActions.push('group-details', { recipients });
 
-export const navigateToRealmInputScreen = (
-  // The `Object.freeze`` in the `:` case avoids a Flow issue:
-  // https://github.com/facebook/flow/issues/2386#issuecomment-695064325
-  args: {| realm?: URL, initial?: boolean |} = Object.freeze({}),
```
gnprice, line 68: Callers haven't passed `initial` for a long time, and the last
    caller that passed `realm` was removed in the previous commit.

In "a recent commit", perhaps? I think the previous commit doesn't involve a call to this function (and in fact it provides `initial: true`.)

gnprice (commented):
Thanks for fixing this! Small comments below and above; then please merge at will.

chrisbobbe:
Thanks for the reviews! Merged. I've posted about what looks like a Jest flake in `pmConversationsModel-test.js`, [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flake.20in.20.60pmConversationsModel-test.2Ejs.60.3F/near/1185732).

PR #3116, opened 2018-11-06
WIP: Move our iOS code to Swift

borisyankov:

PR #4741, opened 2021-05-10
Modernize / strip dead code (a little bit) from AppDelegate.m.

chrisbobbe:
This is the result of searching #3116 for (1) clearly useful things that (2) haven't grown stale.

It's possible that we may have more to do before we start using Swift in our iOS code, but we should probably address that closer to the point where it becomes necessary, to avoid the risk of writing something that'll just go stale and need to be changed in the future.


WesleyAC (approved):
I don't feel very qualified to review this, but it's also all quite small and all looks reasonable, so LGTM, assuming you've tested on-device.

chrisbobbe:
Thanks for the review!

> […] LGTM, assuming you've tested on-device.

I have, so I'll merge.

Issue #4747, opened 2021-05-12
Editing stream name / description puts double quotes around it.


AkashDhiman:
![edit-stream-name](https://user-images.githubusercontent.com/40268170/118048087-658c6500-b399-11eb-8f17-07e3c59a7021.gif)

The code at fault is in `/src/streams/streamsActions.js/updateExistingStream.js`.

where `api.updateStream` is called with the response of `JSON.stringify(newValue.name)` as the `value` parameter.

This code was added in  https://github.com/zulip/zulip-mobile/commit/3e1becac570dc8e2d3a163673c5403a47285c820 and  https://github.com/zulip/zulip-mobile/commit/f86e9147cf6101cfd07858440ab7b31667ccf8d0.

Removing the call to `JSON.stringify` will remove this issue.

It is unclear to me what the "unsafe characters" are described in that commit, I don't think removing the call to `JSON.stringify` can have any problem, but still posting it as an issue to ask what other people think of this.

Gautam-Arora24:
Thanks for the detailed explanation @AkashDhiman! Working on this 🙂

Gautam-Arora24:
We need to be sure what does it mean by "unsafe characters" before removing the `JSON.stringify` method.

PR #4748, opened 2021-05-13
Handle the unnecessary double quotes while editing Stream name / description.

Gautam-Arora24:
**Earlier**
https://user-images.githubusercontent.com/53913514/118182566-c8900180-b456-11eb-9e8a-f69305ae05bc.mov

**After**
Removed the `JSON.stringify()` method from `updateExistingStream` function. There is no need to take care of the safe characters and encoding since it was removed from these parameters from Zulip 4.0.
Link to the docs -> https://zulip.com/api/update-stream#parameter-description
https://user-images.githubusercontent.com/53913514/118183002-40f6c280-b457-11eb-9bd0-3b1e08f2052a.mov

Fixes: #4747

### DIFF
--- a/src/streams/streamsActions.js
+++ b/src/streams/streamsActions.js
@@ -18,17 +18,10 @@ export const updateExistingStream = (
   newValues: {| name: string, description: string, isPrivate: boolean |},
 ) => async (dispatch: Dispatch, getState: GetState) => {
   if (initialValues.name !== newValues.name) {
-    // Stream names might contain unsafe characters so we must encode it first.
-    await api.updateStream(getAuth(getState()), id, 'new_name', JSON.stringify(newValues.name));
+    await api.updateStream(getAuth(getState()), id, 'new_name', newValues.name);
   }
   if (initialValues.description !== newValues.description) {
-    // Description might contain unsafe characters so we must encode it first.
-    await api.updateStream(
-      getAuth(getState()),
-      id,
-      'description',
-      JSON.stringify(newValues.description),
-    );
+    await api.updateStream(getAuth(getState()), id, 'description', newValues.description);
   }
   if (initialValues.invite_only !== newValues.isPrivate) {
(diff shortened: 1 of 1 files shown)


Gautam-Arora24:
@chrisbobbe and @gnprice This PR is ready for a review 🙂

WesleyAC:
What happens when the stream name/description has a `"`, `\`, or similar character in it? Does this still work?

Gautam-Arora24:
Yes, @WesleyAC I have checked by adding `"`, `\` and other similar characters and it just works fine.

Screenshot 2021-05-22 at 11 02 21 PM

timabbott:
Hrm, it appears @ligmitz and I made some sort of mistake in attempting to check whether the mobile apps had the "rename stream" functionality; this needs to be conditional on API feature level 64 in https://zulip.com/api/changelog to preserve compatibility with older servers.  (I'm pretty sure I know what part of my mistake was, and it had to do with my `zulip-mobile` checkout being misconfigured in a way that led to it pulling from timabbott/zulip-mobile not zulip/zulip-mobile, which I've since discovered and fixed.  Though I'd also thought I'd asked, which I'm now pretty confident didn't happen).

In any case, we've accidentally broken compatibility of current mobile clients for using this functionality with now-or-older mobile clients and Zulip 4.0+.  If we think that's important, we can try adding in a Zulip 4.3 or 4.4 release a `client_capabilities` entry to undo the stream name API change for old mobile clients, but I think it's possible the cost/benefit ratio isn't worth it given that (1) probably some folks will be on 4.2 forever, so we can't eliminate the problem entirely and (2) this is an administrative feature that's probably rarely used and possibly just fixing this in a new mobile app release is good enough. and (3) that's a significant amount of extra work that could be better put into more visible issues.

chrisbobbe:
> I think it's possible the cost/benefit ratio isn't worth it

I think that's right, and the thing to do here is for mobile to use a conditional on feature level 64.

PR #4749, opened 2021-05-14
OfflineNotice: Use gray background instead of red, to be less jarring.

chrisbobbe:
Before:

After:

-----

An offline state shouldn't be an error / failure state. The app's
capabilities will be reduced, so the user needs to be informed --
but it's not quite right to use a color that conveys that something
unhandled or unexpected has happened.

And switch to the theme-appropriate text color so it shows up in
light mode.

### DIFF
--- a/src/common/OfflineNotice.js
+++ b/src/common/OfflineNotice.js
@@ -3,9 +3,8 @@
 import React from 'react';
 import { View } from 'react-native';
 
-import type { Dispatch } from '../types';
 import { createStyleSheet } from '../styles';
-import { connect } from '../react-redux';
+import { useSelector } from '../react-redux';
 import { getSession } from '../selectors';
 import Label from './Label';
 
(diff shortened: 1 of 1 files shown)


WesleyAC (approved):
LGTM! Thanks for the cleanup and fix :)

chrisbobbe:

gnprice:
Thanks @chrisbobbe and @WesleyAC !

Reading this now, I have just one quibble: the commit
7c60271b2 OfflineNotice [nfc]: Stop giving `key` attribute to `View`.
should have come before
fffbf3352 OfflineNotice [nfc]: Return `null` instead of empty `View` when online.

As is, when the latter change (coming first) replaces with `null` a `View` that has a `key` attribute, I worry a bit that there might have been a reason for that `key` and so the commit might not be NFC and have some unintended effect. It works out OK, though, and the reason is what the other commit explains: that the `key` isn't actually doing anything useful and can just be deleted anyway.

chrisbobbe:
> Thanks @chrisbobbe and @WesleyAC !
> 
> Reading this now, I have just one quibble: the commit
> [7c60271](https://github.com/zulip/zulip-mobile/commit/7c60271b20b9c144ca6cd7f4c87f484b77601188) OfflineNotice [nfc]: Stop giving `key` attribute to `View`.
> should have come before
> [fffbf33](https://github.com/zulip/zulip-mobile/commit/fffbf3352261f432c333bec757e13e3b562ab9f7) OfflineNotice [nfc]: Return `null` instead of empty `View` when online.
> 
> As is, when the latter change (coming first) replaces with `null` a `View` that has a `key` attribute, I worry a bit that there might have been a reason for that `key` and so the commit might not be NFC and have some unintended effect. It works out OK, though, and the reason is what the other commit explains: that the `key` isn't actually doing anything useful and can just be deleted anyway.

Makes sense, thanks!

PR #4750, opened 2021-05-14
ServerCompatNotice: Add, to nag users connected to servers <2.1.0.

chrisbobbe:
EDIT: updated the threshold from <2.0.0 to <2.1.0; see https://github.com/zulip/zulip-mobile/pull/4750#issuecomment-843645693.

-----

As the jsdoc says, at some point we'll settle on a non-hard-coded threshold; see
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/compatibility.20documentation/near/1174966.

-----

The banner can go pretty much anywhere that makes sense. Here, I've got it on `HomeScreen` only, so it doesn't show up when you're in the message list or even on another tab like the settings screen. I considered putting it in `MainTabsScreen` just above or below the `OfflineNotice`; that'd be fine, but then I figured the current way is obtrusive enough.

The "x" button makes the banner disappear, but that state is deliberately stored in `state.session`, which isn't persisted, so the banner will come back and nag the user on the next startup.

The "Learn more" link goes to https://zulip.readthedocs.io/en/stable/overview/release-lifecycle.html.

Note that the correct handling of safe areas (noticeable in landscape mode on iOS) will look incongruous until #3066 is fixed; #4683 is my next step for that.

I may add an ease-in-ease-out animation when our `usePrevious` util from #4683 is available.

**NOTE:** The has-dismissed-the-banner state isn't stored per-realm, so there's currently a bug where dismissing the banner makes it go away for all <2.0.0 realms the device knows about (until app restart, anyway). I figured this was minor enough to be acceptable for the time being, unless we can think of a nice and easy way to store the state both per-realm and per-session, while being sure to clean up after ourselves when the user wants to forget the realm.

(The screenshots don't show an actual <2.0.0 version; I've tweaked the should-it-show conditional. It shows the `git describe` output we get from `/server_settings`, unmodified, and this is CZO.)

-----

# iOS

-----

# Android


alya:
Hmm, do we need to display the version? I think it makes the message harder to understand, at least for everyone who isn't the admin for the server.

alya:
By the way, do we display a different message if the user is the administrator?

alya:
The "learn more" link should probably go lower on the page: https://zulip.readthedocs.io/en/stable/overview/release-lifecycle.html#compatibility-and-upgrading

timabbott:
The message will usually be like "1.9.3" unless they're running a forked version from Git (like chat.zulip.org is).

chrisbobbe:
> The "learn more" link should probably go lower on the page: https://zulip.readthedocs.io/en/stable/overview/release-lifecycle.html#compatibility-and-upgrading

Cool, sounds good. I'm seeing this, in that section:

> The Zulip mobile and desktop apps maintain backwards-compatibility code to support any Zulip server since 2.1.0. 

Does that mean I can make the banner show for <2.1.0, not just <2.0.0?

chrisbobbe:
> By the way, do we display a different message if the user is the administrator?

I've updated the PR with this text for admins; what do you think?

image

chrisbobbe:
> Hmm, do we need to display the version? I think it makes the message harder to understand, at least for everyone who isn't the admin for the server.

Right; as Tim says, most people won't see the long and complicated string; they'll see "1.9.3" or "2.0.4" or something.

alya:
Ok, either way seems OK for showing the version number to non-admins then.

alya:
For admins, can we make the call to action more explicit? E.g. for the second sentence: "Please upgrade your server as soon as possible by following the simple process described [here](link)." -- or something along those lines?

chrisbobbe:
> For admins, can we make the call to action more explicit?

Yeah, that makes sense. After looking a bit closer, I think we may like do that in a different direction; the "Learn more" button in my previous revision really wants to be a separate button, not so much an inline link on a word in a sentence.

I took a look at the [Material Design doc for banners](https://material.io/components/banners), and made some adjustments in this revision based on that. It says we shouldn't use an "x" to close; it says "All actions should be shown as text buttons." So I've made "DISMISS" button instead of the "x" icon. It also says "Don’t include links in a banner message. All available actions should be represented as buttons."

Here are some screenshots for what an admin sees. "Upgrade now" seemed like a more obvious wording for the action button than "fix now", but I think "upgrade now" gets used a lot when apps ask people to pay for stuff, and I didn't want an admin to be misled by that (maybe they have the power to make decisions but they're not really in the loop about how Zulip works).




And for non-admins, and one screenshot on Android:



In an earlier revision, I considered adding a Zulip logo to signify that it was a message from us, e.g.,



Images are allowed in the Material Design spec, but I think the text is running long, and that would give it less space. In the ["Specs"](https://material.io/components/banners#specs) section, there's a diagram that says "One to two lines is preferable on mobile and tablet", and I'm seeing three on the devices I'm testing on, which are pretty standard.

alya:
Cool, looks good to me! I like that we now know a lot more about adding
banners -- it may come in handy.

src/common/RawLabel.js
```diff
@@ -32,11 +34,18 @@ export default class RawLabel extends PureComponent<Props> {
   };
 
   render() {
-    const { text, style, ...restProps } = this.props;
+    const { text, children, style, ...restProps } = this.props;
+
+    invariant(
+      (text === undefined && children !== undefined)
+        || (text !== undefined && children === undefined),
```
WesleyAC, line 35: I'd think to write `!!text !== !!children`, but perhaps that's more cryptic.

src/session/sessionReducer.js
```diff
@@ -56,6 +57,18 @@ export type SessionState = {|
   pushToken: string | null,
 
   debug: Debug,
+
+  /**
+   * Whether `ServerCompatNotice` (which we'll add soon) has been
+   *   dismissed this session.
+   *
+   * We put this in `SessionState` deliberately, so that users see the
+   * notice on every startup until the server is upgraded. That's a
```
WesleyAC, line 66: What's the mechanism by which this gets reset?

src/common/ZulipTextButton.js
```diff
@@ -0,0 +1,109 @@
+/* @flow strict-local */
```
WesleyAC, line 1: Is there a reason we can't pull in a button like this from a library of material components? I guess this is BRAND_COLOR by default, is there anything else?

WesleyAC (commented):
A few comments, looks good overall :)

src/common/ZulipTextButton.js
chrisbobbe: Hmm, good idea—maybe https://callstack.github.io/react-native-paper/button.html

src/session/sessionReducer.js
chrisbobbe: Good point; I'll mention it in the comment. `state.session` isn't persisted to the disk, and I've set `state.session.hasDismissedServerCompatNotice` to be `false` by default. That means that on startup it'll be initialized to `false`, no matter what its value was in the previous session.

src/common/RawLabel.js
chrisbobbe: Interesting. It's pretty opaque, to me, and I'm not sure I'd choose it everywhere.

But in this setting, it has the unusual property of not really needing to be understood. `RawLabel`'s callers won't violate the invariant if they've read and understood its jsdoc, and if they do violate it, the error message will tell them clearly what they've done wrong. And there's no other reason the invariant would be violated.

It also has the really nice property of letting us fit the whole `invariant(...)` on a single line. So I'll do it; thanks for the idea! 🙂

src/common/ZulipTextButton.js
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Give correct in-between space for a right-aligned set of buttons.
+   */
+  leftMargin?: true,
+
+  /**
+   * Give correct in-between space for a left-aligned set of buttons.
+   */
+  rightMargin?: true,
```
gnprice, line 61: This will put a margin on the right, which is what we want if the buttons are *right*-aligned, no?

I guess it depends on whether the container already has padding such that the edge-most button needs no margin. If so, then for N buttons we only want N-1 of these spaces.

Maybe a better description would be like: "Give a right margin of the correct in-between spacing for a set of buttons."

src/common/ZulipTextButton.js
```diff
@@ -0,0 +1,109 @@
...
+import { View } from 'react-native';
+
+import type { LocalizableText } from '../types';
+import { BRAND_COLOR, createStyleSheet } from '../styles';
+import { Label } from '.';
+import Touchable from './Touchable';
+
+// TODO(?): Support, e.g., "small", "medium", "large" if it makes
+// sense. It probably doesn't.
+type Variant = 'standard';
```
gnprice, line 12: What would small, medium, or large mean? I don't see those at https://material.io/components/buttons .

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+  buttonsRow: {
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+  },
+});
+
+type Props = $ReadOnly<{||}>;
+
+/**
+ * A "nag banner" saying that the server version is unsupported.
```
gnprice, line 35:
```suggestion
 * A "nag banner" saying that the server version is unsupported, if so.
```

Otherwise it looks like this is something that needs to guarded by a conditional :slightly_smiling_face:

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+                store.getState,
+              );
+            }}
+          />
+        </View>
+      </SafeAreaView>
+    );
+  } else {
+    return null;
+  }
```
gnprice, line 100: nit: swap the order of "then" and "else", and make this an early return

That way the reader isn't in suspense about the other case while reading through the bulk of this function's code, and also the code doesn't get an extra level of indentation.

Also that way we can easily have two separate conditionals, one for the version check and one for `hasDismissedServerCompatNotice`. I think that may be a bit cleaner to read.

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+          <ZulipTextButton
+            leftMargin
+            label="DISMISS"
+            onPress={() => {
+              dispatch(dismissCompatNotice());
+            }}
+          />
+          <ZulipTextButton
+            leftMargin
+            label={isAdmin ? 'FIX NOW' : 'LEARN MORE'}
```
gnprice, line 87: If we can, I think it'd be better to have the string begin as "Learn more", and use styling to transform it to all-caps. Among other things that seems less confusing for translators.

That's in fact what the canonical Material implementation for the web seems to do:
https://material.io/components/buttons/web#text-button
image

And it looks like RN's text styles offer the necessary feature for that:
https://reactnative.dev/docs/text-style-props#texttransform
i.e `textTransform: 'uppercase'`.

(Which means this aspect should move into the `ZulipTextButton` implementation.)

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+            onPress={() => {
+              openLinkWithUserPreference(
+                'https://zulip.readthedocs.io/en/stable/overview/release-lifecycle.html#compatibility-and-upgrading',
+                store.getState,
```
gnprice, line 91: Yeah, I think this example further makes me think (https://github.com/zulip/zulip-mobile/pull/4679#discussion_r638429090 ) that having this function take a `SettingsState` would be a preferable API.

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+                text:
+                  '{realm} is running Zulip Server {serverVersionRaw}, which is unsupported. Please contact your administrator about upgrading.',
+                values: { realm: realm.toString(), serverVersionRaw: zulipVersion.raw() },
+              }}
+            />
+          )}
+        </View>
+        <View style={styles.buttonsRow}>
+          <ZulipTextButton
+            leftMargin
```
gnprice, line 79: I don't think we want a left margin here. There's nothing to separate this button from except the left edge of the banner, and the banner already has its own padding.

src/common/ServerCompatBanner.js
```diff
@@ -0,0 +1,101 @@
...
+  },
+  textRow: {
+    padding: 8,
+    flexDirection: 'row',
+  },
+  buttonsRow: {
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+  },
+});
```
gnprice, line 30: From the spec:
https://material.io/components/banners#specs

I read `leftPadding: 16` for the wrapper (while 8 at right and bottom, and the top is specced in different terms that I'm not sure we can get in RN). Then no additional padding for `textRow`, except 12px vertical space between it and `buttonsRow`.

Speculating, I think the idea in giving the text more left padding than right padding may be that because it's justified left, there's naturally some space between the actual text and the right edge of its box -- and so the text will look more centered if the box is actually somewhat right of center.

gnprice (commented):
Thanks @chrisbobbe , and thanks all for the reviews! Comments below.

Glad also to see these cleanups to Label and RawLabel and friends, and hopeful that we can reuse some of this nice careful work on text-buttons and banners.

src/common/ServerCompatBanner.js
chrisbobbe: Ah OK, got it! I presume `textTransform: 'uppercase'` wouldn't error on languages that don't have an 'uppercase' style; that seems like something that they would have thought of.

gnprice:
On the version threshold, commenting here in the top-level thread for a bit broader visibility: rather than go straight to 2.1.0, I'd prefer to get there gradually by first rolling this banner out to a smaller number of users on older servers. That way we get a chance to hear any feedback from people who see it, and adjust as needed before rolling it out to a wider set of users, because this is inherently a sort of message that may make people unhappy.

A threshold of 2.0 would apply to 2-4% or so of our users [as of a couple of months ago](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/emoji.20list/near/1136563), which I think is a good place to start. In fact if we were more of a giant company with tons of users, I'd start another version or two older than that; but as is, I think 2.0 is about right for getting meaningful data from the feedback we hear, or don't hear, from users as it rolls out.

src/common/ZulipTextButton.js
```diff
@@ -0,0 +1,109 @@
...
+type Props = $ReadOnly<{|
+  /**
+   * See note on the `Variant` type.
+   */
+  variant?: Variant,
+
+  /**
+   * Give correct in-between space for a right-aligned set of buttons.
+   */
+  leftMargin?: true,
```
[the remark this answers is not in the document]
chrisbobbe: Huh, my auto-wrap settings would prefer that it break between "of" and "buttons". But I can resist that without lots of effort and get it on one line; done.

src/common/ZulipTextButton.js
chrisbobbe: I don't have particular variants in mind right now, but I wanted to pretty firmly give a good path of least resistance in case we come up with some. In particular, a path that would have us identify a legitimate variant and write its style in a coherent way. A path that doesn't look like "hmm let's add a flag to allow callers to specify leopard print, and another for this one caller where it's running a little too close to a FAB so it needs to scrunch up its character spacing a bit", etc.

Does this approach to that seem useful?

src/common/ServerCompatBanner.js
gnprice: Yeah, agreed.

src/common/ZulipTextButton.js
gnprice: I see -- so IIUC the idea is to try to avoid this situation that `ZulipButton` has:
```js
  style?: SubsetProperties<
    ViewStyle,
    {|
      // The use of specific style properties for callers to micromanage the
      // layout of the button is pretty chaotic and fragile, introducing
      // implicit dependencies on details of the ZulipButton implementation.
      // TODO: Assimilate those usages into a more coherent set of options
      //   provided by the ZulipButton interface explicitly.
      //
      // Until then, by listing here the properties we do use, we at least
      // make it possible when working on the ZulipButton implementation to
      // know the scope of different ways that callers can mess with the
      // styles.  […]
      /* … a list of 14 particular style props … */
```

That is a very sensible thing to want to avoid!

I guess maybe just replace the comment with an explanation more like the one you just gave? Instead of these examples that the comment acknowledges probably don't make sense :slightly_smiling_face:

src/common/ZulipTextButton.js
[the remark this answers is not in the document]
chrisbobbe: Got it, thanks! I now have those same settings; looks like I'd hard-coded that extension's settings to use 70, instead of adding `editor.rulers` and having it consult those

src/common/ServerCompatBanner.js
chrisbobbe: > _Spec_-ulating

(lol)

> the top is specced in different terms that I'm not sure we can get in RN

Looks like maybe we can get that by applying this to the text:

```js
  text: {
    marginTop: 16,
    lineHeight: 20,
  },
```

I'll give it a go in my next revision.

chrisbobbe:
Thanks for the reviews! Revision pushed.

gnprice (commented):
Thanks! Merged, after fixing the one item below and with a couple of small tweaks at the end.

chrisbobbe:
Thanks for the review!

WesleyAC:
FYI, the first commit here introduced a regression in dev, I've pushed a fix in #4801.

WesleyAC:
Amusingly, I was the one who suggested the change that caused that, oops :)

PR #4753, opened 2021-05-20
Prep for timing out the initial fetch.

chrisbobbe:
This supersedes the first part of #4193, which has grown unwieldy. I'm hoping this smaller PR will be easier to review, and then I'll follow it up with another PR (EDIT: #4754) to finish the work of #4193. 🙂

Related: #4165

### DIFF
--- a/src/message/__tests__/fetchActions-test.js
+++ b/src/message/__tests__/fetchActions-test.js
@@ -18,7 +18,8 @@ import type { ServerMessage } from '../../api/messages/getMessages';
 import { streamNarrow, HOME_NARROW, HOME_NARROW_STR, keyFromNarrow } from '../../utils/narrow';
 import { GravatarURL } from '../../utils/avatar';
 import * as eg from '../../__tests__/lib/exampleData';
-import { sleep } from '../../utils/async';
+import { fakeSleep } from '../../__tests__/lib/fakeTimers';
+import { BackoffMachine } from '../../utils/async';
 
 const mockStore = configureStore([thunk]);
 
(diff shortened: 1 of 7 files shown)


src/message/__tests__/fetchActions-test.js
```diff
@@ -82,7 +83,10 @@ describe('fetchActions', () => {
 
   describe('tryFetch', () => {
     test('resolves any promise, if there is no exception', async () => {
-      const tryFetchFunc = () => new Promise(resolve => setTimeout(() => resolve('hello'), 100));
+      const tryFetchFunc = async () => {
+        await sleep(10);
```
WesleyAC: FWIW, Re: the commit message, I don't think 10ms is a very realistic API request response time, although I don think it's reasonable to use in a test. You'd need to be quite close to the server, and calling a fast API endpoint to get 10ms.

src/message/__tests__/fetchActions-test.js
```diff
@@ -116,13 +116,19 @@ describe('fetchActions', () => {
       jest.runAllTimers();
     });
 
-    test('retries a call, if there is an exception', async () => {
+    test('retries a call if there is a non-client error', async () => {
```
WesleyAC, line 119: Could you add something like the commit message as a comment here? I think that's more useful.

WesleyAC (approved):
LGTM overall :) Feel free to merge once you've added the comment I asked for in the last commit.

src/message/__tests__/fetchActions-test.js
chrisbobbe: Yeah; I see that I didn't back up that statement with anything. 🤔 I think I was going for "more realistic than no sleep at all (what the error case was doing) but still wouldn't slow down the tests as noticeably as 100ms." I'll fix it.

Could you please confirm that you meant "do" instead of "don"?

chrisbobbe: Looks like `GET /server_settings` on CZO is taking ~300ms for me.

WesleyAC: Ah, yeah, I meant "do" — 10ms seems fine, it's hard to imagine what wouldn't break at 10ms but would at 100ms.

chrisbobbe:
Thanks for the review! Merged, after making those changes.

gnprice:
Thanks @chrisbobbe and @WesleyAC ! I've just read through this, and have just two small comments:
* One, I turned into 4854a8904, fixing up some comments that I'd left in a confusing state.
* The other is that I think a TODO comment this adds isn't quite right. Discussed already at https://github.com/zulip/zulip-mobile/pull/4754#discussion_r637254865 , on the followup PR where that TODO comment gets acted on.

PR #4754, opened 2021-05-20
Time out initial fetch and go to account-picker screen.

chrisbobbe:
~~This is meant to follow #4753~~ Done! I've split #4193 into two PRs, that one and this one, to make it easier to review.

This is the "minimal version" of a fix Greg describes in the issue; as he mentions there, it would be good to make follow-up improvements soon after merging this, or open a new issue for those.

Fixes: #4165


chrisbobbe:
Also, note to self: I should make it so we _do_ retry the initial fetch on 5xx's if we have server data and the UI is visible; otherwise, there's no point in retrying, if we're looking at the full-screen loading indicator where the user can't do anything useful.

chrisbobbe:
> Also, note to self: I should make it so we _do_ retry the initial fetch on 5xx's if we have server data and the UI is visible; otherwise, there's no point in retrying, if we're looking at the full-screen loading indicator where the user can't do anything useful.

Done; revision pushed!

chrisbobbe:
Note to self: Probably a good idea to alert the user (and tell Sentry) of which bad outcome occurred: a 5xx error or a timeout.

jest/jestSetup.js
```diff
@@ -18,6 +20,11 @@ import mockAsyncStorage from '@react-native-community/async-storage/jest/async-s
 polyfillGlobal('URL', () => URL);
 polyfillGlobal('URLSearchParams', () => URLSearchParams);
 
+/**
+ * Default should be set with `timers: 'modern'` in Jest config.
+ */
+assertUsingModernFakeTimers();
```
gnprice, line 26: nit: this isn't a declaration of anything, so there's nothing the comment is describing the interface of -- so it should be a plain non-jsdoc implementation-comment

src/chat/FetchError.js
```diff
@@ -28,7 +29,13 @@ export default class FetchError extends PureComponent<Props> {
...
     return (
       <View style={styles.container}>
-        <Label style={styles.text} text="Oops! Something went wrong." />
+        {(() => {
+          if (this.props.error instanceof TimeoutError) {
+            return <Label style={styles.text} text="Request timed out." />;
+          } else {
+            return <Label style={styles.text} text="Oops! Something went wrong." />;
+          }
+        })()}
```
gnprice, line 38: Probably simpler as a `… ? … : …` expression, no?

gnprice (commented):
Thanks @chrisbobbe ! Generally this looks good -- small comments below, and one more-significant question about the set of errors we're handling.

As of the end of the branch I believe we're handling 5xx, 4xx, and timeouts appropriately. But there could be other ways a network failure shows up: if the device is unable to make a TCP connection (perhaps because it's in airplane mode), or a TLS connection atop that (perhaps because there's something wrong operationally with the server), then we'll treat that as an unexpected error that represents a bug. Can we work out what those would look like here, and handle them appropriately so that there's no operational issue that could cause an uncaught exception?

Separately: there's a string of 8 commits at the end that convert over from Lolex to Jest's fake timers. Do those need to be at the end, or can they come earlier -- perhaps even as a separate PR?

src/chat/FetchError.js
chrisbobbe: > Probably simpler as a `… ? … : …` expression, no?

Sure. I used `if` / `else` because I thought there might be other cases to consider in the future, and this form would make it easier to expand on.

gnprice: I see. Yeah, that seems like a fine reason.

chrisbobbe:

Two (possibly related?) things have occurred to me:
- Is `INITIAL_FETCH_ABORT` quite the right name, or should it be more like `INITIAL_FETCH_ERROR`? (Compare `MESSAGE_FETCH_ERROR`.)
- One place in `api.registerForEvents` where things could conceivably fail is in the step where we transform the data we receive from the server (with a function named `transform`). Do we need to change our error handling for that at all?

chrisbobbe:
(Just rebased.)

src/api/apiErrors.js
```diff
@@ -61,8 +61,32 @@ export const makeErrorFromApi = (httpStatus: number, data: mixed): Error => {
...
   e instanceof ApiError && e.httpStatus >= 400 && e.httpStatus <= 499;
+
+/**
+ * Is exception caused by a Server Error (5xx)?
+ *
+ * A Server (5xx) error may or may not be resolved by waiting a short
+ * time and retrying the same request.
+ */
+export const isServerError = (e: Error): boolean =>
+  e instanceof ApiError && e.httpStatus >= 500 && e.httpStatus <= 599;
```
gnprice, line 77: Hmm. Does this actually happen in practice?

Looking at `makeErrorFromApi` above, which is where we construct `ApiError` objects, it looks like on a typical 5xx response we'll currently throw just an `Error` object, with some text in the message. (We'd throw an `ApiError` on a 5xx response if it comes with a body that's valid JSON and decodes to a nice Zulip API error object `{ result: 'error', msg: string, code?: string | void, ... }`. But a 5xx is a server failure, and may not carry nice well-formed JSON like that.)

So I think we want the code in `src/api/` to be throwing a more distinctive error in that case, and then to have `isServerError` detect that.

src/message/__tests__/fetchActions-test.js
```diff
@@ -102,69 +102,131 @@ describe('fetchActions', () => {
...
+          }, withRetry);
+
+          await fakeSleep(60000);
+          return expect(tryFetchPromise).rejects.toThrow(TimeoutError);
+        });
+      },
+    );
+
+    describe('if asked to retry', () => {
+      test('retries a call if there is a recoverable error', async () => {
```
gnprice, line 164: fetchActions: Fail early on all non–known-retryable errors in `tryFetch`.

I appreciate your careful use of both an en-dash and a hyphen. Wouldn't want a reader to think we're talking about errors that are retryable in a non-known fashion :-P

src/message/fetchActions.js
```diff
@@ -310,37 +393,62 @@ export const doInitialFetch = () => async (dispatch: Dispatch, getState: GetStat
...
+    } else if (isNetworkRequestFailedError(e)) {
+      dispatch(initialFetchAbort('network'));
+    } else if (e instanceof TimeoutError) {
+      // We always want to abort if we've kept the user waiting an
+      // unreasonably long time.
+      dispatch(initialFetchAbort('timeout'));
+    } else {
+      logging.warn(e, {
+        message: 'Unexpected error during initial fetch and serverSettings fetch.',
+      });
```
gnprice, line 450: We probably want to be aborting in this case too, right?

Basically it's an internal error in the app. But the user is probably better off getting kicked back to the account picker than having the loading screen continue indefinitely -- it means at least they can try again and see if the issue recurs, or try another account and see if they can at least use that.

gnprice (commented):
Thanks for the revision! Sorry I've let this sit for a bit.

The app-level (i.e., outside `src/api/`) code looks good, with one comment below.

I think the `apiErrors` code doesn't quite work together properly with the existing code in that file. That existing code should change, I think, to handle this properly. I'll sketch out a version of how that could look, and either push some draft commits to this branch or send a PR.

src/message/__tests__/fetchActions-test.js
chrisbobbe: I wish I still had access to the Chicago Manual of Style; it looks like this would be covered in Chapter 6: https://www.chicagomanualofstyle.org/16/ch06/ch06_toc.html 😛 (but I no longer have my college-sponsored subscription, alas).

gnprice:
> I think the `apiErrors` code doesn't quite work together properly with the existing code in that file. That existing code should change, I think, to handle this properly. I'll sketch out a version of how that could look, and either push some draft commits to this branch or send a PR.

OK, and sent #4896 for that.

chrisbobbe:
Thanks for the review, and for #4896! Revision pushed.

gnprice:
Looks good, thanks! Merging.

chrisbobbe:
Thanks for the review! 🙂

PR #4755, opened 2021-05-20
postinstall: Don't run `pod install` unless it's necessary.

chrisbobbe:
This takes several seconds each time I run `yarn`.

### DIFF
--- a/tools/postinstall
+++ b/tools/postinstall
@@ -18,7 +18,15 @@ EOF
         exit 1
     fi
 
-    pod install --project-directory="$ROOT_DIR/ios"
+    # Do nothing if the installed pods already match the Podfile.lock.
+    # If they don't match, run `pod install`. Conditional copied from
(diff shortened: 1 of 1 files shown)


WesleyAC (approved):
Haven't tested this, but it looks reasonable and shellcheck is happy, so it seems fine to merge :)

chrisbobbe:

gnprice:
Thanks @chrisbobbe for spotting and fixing this! This is a nice quality-of-life improvement for developing on a Mac -- on the Mac Mini I have at my desk, `yarn` is now down to 4 seconds when there's nothing to do, and it was more like 16.

Seems like if this is a correct optimization then it's really a performance bug in `pod install` that it doesn't do this itself, but :shrug: . The fact that CocoaPods sticks this same optimization into the Xcode project file seems like a good enough promise that it is indeed a correct optimization.

Issue #4758, opened 2021-05-23
[android] Opening the app from background opens it in a specific narrow, if it was initially started via pressing a notification.


AkashDhiman:
If the app is initially started by pressing on the notification, on closing the app (not killing but quitting via back presses), and starting it from the overview screen it will start at the topic the notification was directed to.

Similar to the issue solved in https://github.com/zulip/zulip-mobile/commit/bb1bba0e3e9b87d4246c895e85dd6a1b540cac03, Originally speculated to exist by @gnprice at https://github.com/zulip/zulip-mobile/pull/4514#discussion_r620874033.

Steps to reproduce:
1. Don't have the app running in background, recent tasks.
2. Receive a notification.
3. Tap on the notification, the app will initiate startup and will open at the topic the notification was pointing to.
4. Now go back, and quit the app (make sure to quit and not kill the app, i.e. use back presses).
5. go to overview screen (by pressing the overview soft key), Notice the app activity is still present here.
6. Tap on it, it will start and navigate to the topic it was initially navigated to.

Screen capture:
![notification-bug](https://user-images.githubusercontent.com/40268170/119274474-bae43400-bc2d-11eb-9722-f9498d1100dc.gif)

PR #4759, opened 2021-05-23
Make `getInitialNotification` a consumer.

AkashDhiman:
Quoting from commit of this PR:

> It is speculated that when an app quits android system does not
flush the entire memory of the app until absolutely necessary
(to facilitate faster startup next time). This also involves the
static variable `initialNotification`.

> This situation creates a problem when the app is started next time,
where it processes `initialNotification` again and falsely navigates
to its narrow.

Fixes: #4758

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/NotificationsModule.java
+++ b/android/app/src/main/java/com/zulipmobile/notifications/NotificationsModule.java
@@ -40,11 +40,12 @@ class NotificationsModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
-    public void getInitialNotification(Promise promise) {
+    public void readInitialNotification(Promise promise) {
         if (null == initialNotification) {
             promise.resolve(null);
         } else {
             promise.resolve(Arguments.fromBundle(initialNotification));
+            initialNotification = null;
         }
     }
 }
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -111,6 +111,14 @@ export const getNarrowFromNotificationData = (
   return pmNarrowFromRecipients(pmKeyRecipientsFromIds(ids, ownUserId));
 };
 
+/**
+ * Read the notification the app was started from, if any.
+ *
+ * This consumes the data; if called a second time, the result is always
+ * null.
+ *
+ * (TODO: Well, it does on Android, anyway.  #4763 is for doing so on iOS.)
+ */
 const readInitialNotification = async (): Promise<Notification | null> => {
   if (Platform.OS === 'android') {
     const { Notifications } = NativeModules;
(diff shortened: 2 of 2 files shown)


src/notification/index.js
```diff
@@ -114,7 +114,7 @@ export const getNarrowFromNotificationData = (
 const getInitialNotification = async (): Promise<Notification | null> => {
```
AkashDhiman, line 114: Not sure about the name of this function, android provides us with a `readInitialNotification` and iOS provides us with a `getInitialNotification` this function calls one of them based on the platform. Should I split this and write platform conditional inside `handleInitialNotification`?

gnprice: Good question.

I just tested your #4758 repro instructions on iOS (translated slightly to the iOS UI), and it reproduces there. So the `getInitialNotifications` is an accurate name for that function that `PushNotificationIOS` gives us, it seems.

I think probably the right thing would be:
* Rename this function of ours to `readInitialNotification`, because that describes the behavior we *want* it to have, and that on Android it now does have.
* File an issue as the iOS counterpart of #4758, and add a TODO comment here pointing at that issue.

android/app/src/main/java/com/zulipmobile/notifications/NotificationsModule.java
```diff
@@ -40,11 +40,12 @@ public void getToken(Promise promise) {
...
 
     @ReactMethod
-    public void getInitialNotification(Promise promise) {
+    public void readInitialNotification(Promise promise) {
         if (null == initialNotification) {
             promise.resolve(null);
         } else {
             promise.resolve(Arguments.fromBundle(initialNotification));
         }
+        initialNotification = null;
```
gnprice, line 49:
```suggestion
    public void readInitialNotification(Promise promise) {
        if (null == initialNotification) {
            promise.resolve(null);
        } else {
            promise.resolve(Arguments.fromBundle(initialNotification));
            initialNotification = null;
        }
```

This has the same behavior but I think is a bit cleaner -- it more closely ties the act of clearing `initialNotification` with the act of passing its contents on elsewhere. That makes it more like just a single atomic act of consuming it, which is how we want to think of this.

gnprice (commented):
Thanks @AkashDhiman for this fix! One small comment below, and I replied to a question you asked above.

Just to be explicit: you've tested empirically that this fixes #4758?

---

In the commit message:

    It is speculated that when an app quits android system does not
    flush the entire memory of the app until absolutely necessary
    (to facilitate faster startup next time).

Yeah, in fact this is a documented feature of the Android platform. Here's one discussion of it:
https://developer.android.com/guide/components/activities/activity-lifecycle#alc
> As the user begins to leave the activity, the system calls methods to dismantle the activity. In some cases, this dismantlement is only partial; the activity still resides in memory (such as when the user switches to another app), and can still come back to the foreground. If the user returns to that activity, the activity resumes from where the user left off. [… more discussion …]

AkashDhiman:
Yes, I tested and made sure it fixed the issue.

src/notification/index.js
gnprice: > * File an issue as the iOS counterpart of #4758,

I've just filed this, as #4763.

AkashDhiman:
@gnprice made the changes you requested and edited the commit message a bit based on the doc you provided. You may review again.

gnprice:
Thanks @AkashDhiman ! Merged.

I added a commit on top with some jsdoc, including that TODO comment mentioned above. The only other tweak I made was to the summary line of the main commit.

PR #4760, opened 2021-05-24
example data: Better support EVENT_NEW_MESSAGE.

gnprice:
This follows up on the thread at https://github.com/zulip/zulip-mobile/pull/4710#discussion_r626113834.

We replace the "action fragments" idea with more-flexible factory functions. In particular all the places in test code where we used `eg.eventNewMessageActionBase` to create an `EVENT_NEW_MESSAGE` action now make a call to the new `eg.mkActionEventNewMessage` which subsumes it -- and unlike the old fragment, the new factory can ensure that `action.message.flags` is always present, as it should be.

Then that in turn lets us clean up some app code, in reducers, that had provided bogus default behavior in case of a missing `flags` in order to accommodate tests that didn't supply one.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -722,6 +722,14 @@ export const eventNewMessageActionBase /* \: $Diff<EventNewMessageAction, {| mes
  * suffice.
  */
 
+/**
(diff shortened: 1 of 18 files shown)


chrisbobbe:
LGTM, thanks, @gnprice! Merged.

PR #4761, opened 2021-05-24
[android] Make wifi hardware requirement optional. (follow up to #4724)

AkashDhiman:
Usage of `@react-native-community/netinfo` adds the permission
requirement for ACCESS_WIFI_STATE in the final AndroidManifest of
the app, which makes wifi hardware a compulsory requirement to
install the app.

Adding an appropriate `uses-feature` tag makes this requirement
optional.

This was discovered by looking into "Merged Manifest" option while
previewing `AndroidManifest.xml` in Android Studio.

Follow up to: #4724 as suggested by @gnprice.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -24,6 +24,14 @@
         android:name="android.hardware.camera.autofocus"
         android:required="false" />
 
+    <!-- We use certain modules that adds the permission requirement for
+         `ACCESS_WIFI_STATE` in the final AndroidManifest of the app, which makes
+         wifi hardware a compulsary requirement to install the app. The following
+         statement makes the hardware requirement optional. -->
+    <uses-feature
+        android:name="android.hardware.wifi"
+        android:required="false" />
+
     <application
         android:name=".MainApplication"
         android:allowBackup="true"


android/app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +24,15 @@
         android:name="android.hardware.camera.autofocus"
         android:required="false" />
 
+    <!-- Usage of @react-native-community/netinfo, rn-fetch-blob and Flipper
+         adds the permission requirement for ACCESS_WIFI_STATE in the final 
+         AndroidManifest of the app which makes wifi hardware a compulsary 
+         requirement to install the app. The following statement makes the
```
gnprice, line 30: nit: no trailing whitespace

This is one bit of linting that's cross-language enough that Git itself highlights it when you use `git diff`:
image

(At least it does if you have `git config --global color.ui auto`, which I definitely recommend. See in general [my gitconfig](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3).)

gnprice, line 30: Let's have this commit message link to the same piece of docs as the previous one did:
https://developer.android.com/guide/topics/manifest/uses-feature-element

Actually, better yet, to the relevant section within that doc:
https://developer.android.com/guide/topics/manifest/uses-feature-element#permissions

It's really pretty counterintuitive, at least to me, that the permission implicitly means a hard requirement for the hardware feature, so seeing the doc helps a lot.

android/app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +24,15 @@
         android:name="android.hardware.camera.autofocus"
         android:required="false" />
 
+    <!-- Usage of @react-native-community/netinfo, rn-fetch-blob and Flipper
+         adds the permission requirement for ACCESS_WIFI_STATE in the final 
+         AndroidManifest of the app which makes wifi hardware a compulsary
```
gnprice, line 29: Small comment/doc-writing tip: better to leave this less specific, just saying that some of our dependencies add this, rather than naming which ones in particular.

The reason is that the exact list may change over time: a future version of one of these may stop doing so, a future version of another dependency may start; we might drop one of these, or follow it across a rename, or pull in a new dependency that does the same thing too. When the comment is very specific like this, any of those changes will make it untrue and potentially confusing to a future reader -- *but*, for the point it's really trying to make, none of those details matter, only the fact that there's any dependency at all that does this. So that makes the point just as well, and is much more likely to remain true over time.

OTOH in the *commit message* these details are great to include (as you do) -- they explain your/our thinking at the time we're making this change, which can be helpful to a future reader; and as part of the commit message they're inherently contextualized as being about this particular moment in time. There's a bit more on this theme in our style guide:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#commit-messages-vs-comments

gnprice (commented):
Thanks @AkashDhiman! A couple of small comments below.

And one tiny nit in the commit message, at the end:

    Follow up to: #4724

The style for this sort of metadata line at the end of a Git commit message is to use dashes, not spaces. You can see this in the examples in this style-guide section:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#mentioning-people
like "Suggested-by:" and "Co-developed-by:". I'll add a quick note there mentioning that aspect explicitly.

gnprice:
> I'll add a quick note there mentioning that aspect explicitly.

e822d1b05

android/app/src/main/AndroidManifest.xml
AkashDhiman: Thanks for the `.gitconfig`.

android/app/src/main/AndroidManifest.xml
AkashDhiman: I see, will keep this in mind.

AkashDhiman:
@gnprice, did the changes you mentioned.

AkashDhiman:
I have now also clarified the last paragraph of the commit message a bit.

gnprice:

Issue #4765, opened 2021-05-27
ImageBackground followups from #4730


gnprice:
This is to track the post-merge review comments at https://github.com/zulip/zulip-mobile/pull/4730#pullrequestreview-666388091 , on the changes there to UserAvatar.

PR #4767, opened 2021-05-28
UserAvatar: Ignore color inversion on iOS.

WesleyAC:
ed69e1ec0 inadvertently removed this when we switched away from
ImageBackground, but it seems like something that we should include,
since avatars will look better non-inverted, and inverting arbitrary
images isn't likely to be very useful for contrast/accessibility
reasons.

Fixes: #4765


gnprice:
Looks good, thanks, merged!

Issue #4764, opened 2021-05-27
Make compose box topic empty when starting from interleaved stream view


alya:
At present, when one starts composing a message from an interleaved stream view, the topic defaults to the topic of the latest message in the stream. 

This is confusing, and means that the user has to be careful to change it if, e.g., they just want to start a new topic. It's especially confusing that the compose box uses the topic of the most recent message even if you're currently viewing earlier messages in a totally different topic.

When starting from an interleaved stream view, we should change the behavior to make the topic box blank and put the cursor there to start with. 

The typeahead will help if you are trying to respond to a topic, or you can respond to a topic from a non-interleaved topic view.

[CZO thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/compose.20box.20topic)

PR #4769, opened 2021-05-29
ComposeBox: Don't set topic by default in stream narrows.

WesleyAC:
This is typically confusing to users, so just remove it and force people
to specify the topic they want.

We should follow up by having the "Topic" input be the default thing
that's focused, rather than the "Message" input.

Fixes: #4764


WesleyAC:
Added two commits to this PR that do some cleanup on ComposeBox (making the types reflect reality more).

WesleyAC:
Hmmm, the last commit in this stack seems pretty buggy, I'm not sure if RN offers a nice way to do what I'm trying to do.

The other commits should still be good for review, though.

WesleyAC:
Superseded by #4773.

PR #4772, opened 2021-06-01
ComposeBox: Don't require caughtUp to send in stream narrow.

WesleyAC:
See CZO [1] for some background on this. There are many reasons why the
destination narrow might not be the caughtUp narrow in the stream view,
and it's worse to prevent the user from sending with a cryptic error
message than it is to allow them to send to a narrow they aren't fully
caught up in.

[1]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Failed.20to.20send.20on.20Android


WesleyAC:
Superseded by #4773.

PR #4773, opened 2021-06-02
Composebox cleanup

WesleyAC:


WesleyAC:
The first two commits here a kinda big and messy, sorry about that :( but the rest are reasonably small and nice.

src/compose/ComposeBox.js
```diff
@@ -320,23 +324,22 @@ class ComposeBox extends PureComponent<Props, State> {
...
+      this.setMessageInputValue('');
 
-    this.setMessageInputValue('');
+      if (this.mentionWarnings.current) {
+        this.mentionWarnings.current.clearMentionWarnings();
+      }
 
-    if (this.mentionWarnings.current) {
-      this.mentionWarnings.current.clearMentionWarnings();
+      dispatch(sendTypingStop(narrow));
```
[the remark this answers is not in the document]
chrisbobbe: Oh, also: looks like this commit starts having us use the destination narrow for the "Failed to send message" check, which is relevant to [a known bug](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Failed.20to.20send.20on.20Android/near/1155882):

> Hmm, I'm getting a "failed to send message" popup when trying to message in #**test here** , and I'm not sure why.

[…]

> yeah, it bases it on whether you have any unfetched messages in the narrow you're in, not the topic you're trying to send to. that's just a bug no matter how you slice it.

So that's something else that would be great to stand on its own, and link to that conversation (I'm not finding a corresponding issue). 🙂 Hopefully extracting that out will leave this as a commit we can mark NFC (good for reviewing and history-reading), or not too many steps off from being NFC.

src/compose/ComposeBox.js
```diff
@@ -45,7 +45,6 @@ import MentionWarnings from './MentionWarnings';
 import {
   getAuth,
   getIsAdmin,
-  getLastMessageTopic,
```
chrisbobbe, line 48:
```
We should follow up by having the "Topic" input be the default thing
that's focused, rather than the "Message" input.
```

This seems like it'd be pretty easy to do, with the `autoFocusTopic` prop you introduce later in the PR? 😅

src/compose/ComposeBox.js
```diff
@@ -63,7 +62,7 @@ type Props = $ReadOnly<{|
   insets: EdgeInsets,
 
   narrow: Narrow,
-  editMessage: EditMessage | null,
+  editing: boolean,
```
chrisbobbe, line 64: nit: I might name this `isEditing`, for more explicitness that it's a boolean

src/compose/ComposeBox.js
```diff
@@ -69,6 +69,11 @@ type Props = $ReadOnly<{|
   defaultMessage?: string,
   defaultTopic?: string,
 
+  autoFocusTopic: boolean,
+  autoFocusMessage: boolean,
+  useTopicAutocomplete: boolean,
+  alwaysShowTopic: boolean,
```
chrisbobbe, line 74:
```
ComposeBox [nfc]: Add autoFocus props.
```

`useTopicAutocomplete` and `alwaysShowTopic` don't quite fit the description in the commit message.

Also, do you have a specific use for them in mind? It looks like we're not passing either one of them, in `ComposeBox`'s only callsite.

One thing I'd caution about is that there are cases where it'll be inappropriate to show anything about a topic: when editing a PM, and when composing a new PM to send.

Instead of these two props, I wonder if there could be just one prop that says whether the compose box should be aimed at working on a PM message or a stream message.

Maybe most of `getCanSelectTopic` (all but the `isFocused` part, I think) can be pulled out of `ComposeBox` to get the value for that prop? That seems like a refactoring commit that might go in the right direction. (Separately, though, I don't like the way we use `props.narrow` to decide whether the compose box should be aware of topics—eventually, I'd like to stop refusing to show the compose box in narrows like the all-messages narrow, from which reasonable users will want to edit a message without having to find it in a different narrow first. That'll mean inspecting the actual message the user wants to edit, in some appropriate place, to see if it's a stream message or a PM.)

src/chat/ChatScreen.js
```diff
@@ -23,6 +23,7 @@ import { getFetchingForNarrow } from './fetchingSelectors';
 import { getShownMessagesForNarrow, isNarrowValid as getIsNarrowValid } from './narrowsSelectors';
 import { addToOutbox } from '../actions';
 import { getAuth, getCaughtUpForNarrow } from '../selectors';
+import { getDraftForNarrow } from '../drafts/draftsSelectors';
```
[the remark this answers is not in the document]
chrisbobbe: > It'd be good to keep the commit that removes `UNSAFE_componentWillReceiveProps` as focused as possible.

Well, here's one behavior change I've just found, and it's a nice one! Just before this commit, I get the following (and also on `master` I get something that looks the same) when I start editing a message, along with lots of lag. It goes away at this commit; I suspect that's because of removing `UNSAFE_componentWillReceiveProps`.

image

chrisbobbe (changes requested):
Thanks, @WesleyAC! I'm excited for `ComposeBox` to continue getting detangled! Comments below.

WesleyAC:
I pushed a bunch of revisions that make the commit history nicer here, and fix some of the comments you brought up. Will reply inline.

src/chat/ChatScreen.js
```diff
@@ -188,6 +188,7 @@ export default function ChatScreen(props: Props) {
           onSend={sendCallback}
           defaultTopic={editMessage ? editMessage.topic : undefined}
           defaultMessage={editMessage ? editMessage.content : draft}
+          autoFocusMessage={editMessage !== null}
```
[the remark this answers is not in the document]
WesleyAC: My understanding was that this was a regression when switching to use `key`, although I'm not 100% sure why/how. I've refactored the commits so that this line is added in the commit that switches to using `key` in the first place, so as not to introduce the problem.

src/compose/ComposeBox.js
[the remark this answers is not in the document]
WesleyAC: Added a commit to just do the `const narrow = this.getDestinationNarrow()` change, and added a link to that CZO discussion in the commit.

Removing `state.topic` as you describe seems like a good plan to me.

src/chat/ChatScreen.js
```diff
@@ -180,7 +182,14 @@ export default function ChatScreen(props: Props) {
...
       })()}
       {showComposeBox && (
-        <ComposeBox narrow={narrow} editMessage={editMessage} onSend={sendCallback} />
+        <ComposeBox
+          narrow={narrow}
+          editMessage={editMessage}
+          onSend={sendCallback}
+          defaultTopic={editMessage ? editMessage.topic : undefined}
+          defaultMessage={editMessage ? editMessage.content : draft}
+          key={JSON.stringify([editMessage, narrow])}
```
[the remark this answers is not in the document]
WesleyAC: Fixed :)

src/chat/ChatScreen.js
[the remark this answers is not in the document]
WesleyAC: > Hmm, could you say more about how adding `defaultTopic` and `defaultMessage` lets us get rid of `UNSAFE_componentWillReceiveProps`? What problems might we run into if we tried to get rid of it before adding those props? The changes look orthogonal to me, but I might be missing something. Or, if there would be problems, how about putting it after?

You're correct, these are orthogonal. I've refactored the commits to avoid mixing them up.

src/chat/ChatScreen.js
```diff
@@ -112,6 +115,26 @@ export default function ChatScreen(props: Props) {
   const sayNoMessages = haveNoMessages && !isFetching;
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
+  const sendMessage = (
+    getState: GetState,
+    dispatch: Dispatch,
+    _: GetText,
+    message: string,
+    callbackNarrow: Narrow,
```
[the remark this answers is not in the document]
WesleyAC: Renamed :)

src/compose/ComposeBox.js
WesleyAC: Sadly not — this is referring to when you go and tap on the "Message" input box, it should actually focus the topic input box, but we don't want to `autoFocus` it in the sense that it'll grab your keyboard as soon as you load the message.

src/compose/ComposeBox.js
WesleyAC: Renamed :)

src/compose/ComposeBox.js
WesleyAC: > `useTopicAutocomplete` and `alwaysShowTopic` don't quite fit the description in the commit message.

Ugh, these accidentally got mixed in when I was rebasing, I used them later in the new topic UI work. They're gone in my new revision.

I don't think that the refactor you describe would work for the new topic UI, although it seems like a reasonable idea in general.

The idea behind these was:

`useTopicAutocomplete`: When you're making a new topic, we don't want to autocomplete old topic names.
`alwaysShowTopic`: This makes it show the "Topic" input box even when the composebox isn't focused, which is desirable for the new topic UI, but not for general use. What we probably actually want is a `showTopicBox: 'always' | 'focus' | 'never';` with `focus` as the default, or something like that.

WesleyAC:
Should be ready for another review @chrisbobbe :)

src/compose/ComposeBox.js
chrisbobbe: Got it, makes sense.

> I wonder if there could be just one prop that says whether the compose box should be aimed at working on a PM message or a stream message.

I think exploring a prop like the one I've mentioned could be work for a different PR. I'm interested in looking into that at some point, if it sounds OK to you; I don't think there would be a lot of obstacles to adding it, but you're right that it wouldn't get you any farther toward making the new-topic UI.

For the new-topic UI, instead of `useTopicAutocomplete` and `alwaysShowTopic`, what do you think about a single prop like `isRequestingNewTopic` (or some better name)? If that sounds OK to you, I think it could be a good anchor for this part of `ComposeBox`'s interface, to help keep it steady until `ComposeBox` (eventually) has better support for PMs. When that time comes, we'll naturally be curious about adding the analogous "new PM" UI that the web app has alongside "new topic"—

image

—and I could imagine it being nice to add a simple `isRequestingNewPm` prop (or perhaps coalescing the two into `isRequestingNewDestinationNarrow` if that turns out to be worth considering).

WesleyAC: My reason for doing separate props was so that it would be clearer what they actually do, but I see the argument for bundling them as well.

Perhaps we should have a `mode: 'stream' | 'topic' | 'pm' | 'newTopic' | 'newPm';` or something like that? We wouldn't want to let someone set both `isRequestingNewPm` and `isRequestingNewTopic`.

chrisbobbe (changes requested):
Thanks for the revision! Comments below.

src/compose/ComposeBox.js
chrisbobbe: > My reason for doing separate props was so that it would be clearer what they actually do

Fair, although a good jsdoc can clarify what a prop does (and we're trying to be better about that) 🙂.

> Perhaps we should have a `mode: 'stream' | 'topic' | 'pm' | 'newTopic' | 'newPm';` or something like that? We wouldn't want to let someone set both `isRequestingNewPm` and `isRequestingNewTopic`.

Seems worth a try; I like that this means we don't have to think about the possibility of contradictory settings being set.

src/chat/ChatScreen.js
```diff
@@ -139,6 +141,8 @@ export default function ChatScreen(props: Props) {
           narrow={narrow}
           editMessage={editMessage}
           completeEditMessage={() => setEditMessage(null)}
+          defaultTopic={editMessage ? editMessage.topic : undefined}
+          defaultMessage={editMessage ? editMessage.content : draft}
```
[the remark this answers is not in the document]
WesleyAC: > I'm having some trouble verifying this commit as a pure refactor, and I think a main reason is that I'm not sure why `editMessage` is making an appearance in it, and I don't find it explained.

Hmmm, this might be an artifact of the way that I split up these commits. (Or maybe it's needed, I'm not actually sure).

It's basically because of the code in `UNSAFE_componentWillReceiveProps`, which is `editMessage`-dependent. Perhaps that doesn't matter for this commit, and should only be added once I remove `UNSAFE_componentWillReceiveProps`, though?

It's been sort of frustrating for me to try to get this PR into a state with a bunch of atomic commits, since the fundamental thing that I'm trying to fix is that `ComposeBox` is sort of a mess where changing one thing has a bunch of cascading effects most of the time.

I can also appreciate that it's much easier to review things that are small atomic commits, so I'm happy to work towards that, I'm just not confident in my ability to figure out exactly what each commit does.

I do think you can probably convince yourself that this commit is NFC, though — if `editMessage` is non-null, that means that `UNSAFE_componentWillReceiveProps` will run, and so none of what we set in the lines that are changed will matter. I guess that means that I should do that as part of a separate commit, I'll try to fix that up...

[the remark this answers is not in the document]
WesleyAC: > I'm also wondering why, in the case where `editMessage` is `null`, is there the following asymmetry between `defaultTopic` and `defaultMessage`: The value passed as `defaultTopic` is an emptiness value (`undefined`), leaving `ComposeBox` to fill in a meaningful value, while the value passed as `defaultMessage` is meaningful (`draft`).

This isn't quite correct, since `draft` can be `undefined` — it's possible that both will be unset, or it's possible that just message will be set.

The reason for doing it this way is that we don't save drafts of topic names (except sort of incidentally, in the fact that we store them by narrow). I'm not really sure what the best path forward here is — I think in involves rethinking how drafts work in general.

src/chat/ChatScreen.js
```diff
@@ -114,6 +117,26 @@ export default function ChatScreen(props: Props) {
   const sayNoMessages = haveNoMessages && !isFetching;
   const showComposeBox = canSendToNarrow(narrow) && !showMessagePlaceholders;
 
+  const sendMessage = (
```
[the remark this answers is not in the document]
WesleyAC: Can we write down in our styleguide when we want to use `useCallback`? I tend to see conflicting reports online, and I don't have a good understanding of what the tradeoffs are.

WesleyAC:
I've done some of these changes, will do the rest when I'm back at my computer. Specifically, I think I've done everything except:

- [ ] Turn `Pull sending logic out of ComposeBox` into a NFC commit
- [ ] Use `useCallback` in `src/chat/ChatScreen.js`

Let me know if that's incorrect, GitHub's UI is quite unwieldy for me for finding things in a review with this many comments :)

WesleyAC:
(Also, I've not tested all of these changes on-device yet, since the `yarn update` that I rebased onto seems to have broken the Android build for me, will also do that when I get back)

WesleyAC:
Ok, I beleive I've addressed all of your comments @chrisbobbe :)

WesleyAC:
Just rebased this and fixed up a merge conflict. Would love to get this in soon :)

src/compose/ComposeBox.js
```diff
@@ -77,8 +63,15 @@ type Props = $ReadOnly<{|
...
 
   narrow: Narrow,
-  editMessage: EditMessage | null,
-  completeEditMessage: () => void,
+  isEditing: boolean,
+
+  onSend: (GetState, Dispatch, GetText, string, Narrow) => void,
+
+  defaultMessage?: string,
+  defaultTopic?: string,
```
gnprice, line 71: Is "default" an accurate name for these? It feels like "initial" would be more appropriate -- like `initialMessage` and `initialTopic`.

In particular, there isn't a particular expectation that this message value is one that the user will actually go with -- it's a draft and probably incomplete, or the existing content of a message they've decided to edit.

Saying something like "initial" would also highlight that their behavior is somewhat unusual for props, in that they get consulted only when the component is first initialized:
```js
  state = {
    // …
    topic:
      this.props.defaultTopic !== undefined
        ? this.props.defaultTopic
        : /* … */,
    message: this.props.defaultMessage || '',
```

Which is good because that's kind of a gotcha in the interface -- it's the reason that it's important that we pass a `key` that changes when these initial values change. (Or, at the point in the branch where these are introduced: it's not yet a gotcha in the component's interface, but making that the case requires a hack in its implementation, namely the use of `UNSAFE_componentWillReceiveProps`.)

And in fact we don't have the key change in all the cases where these initial values do -- we have it depend on `editMessage` but not `draft`. That gets the right behavior, because we're continuously updating the draft in Redux as the user edits, and the input is uncontrolled so there's no need to change anything to reflect it… but it is odd, so it's good for it to stick out more.

src/compose/ComposeBox.js
```diff
@@ -320,15 +315,11 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   handleSend = () => {
-    const { dispatch, narrow, caughtUp, _ } = this.props;
+    const { dispatch, _ } = this.props;
     const { message } = this.state;
+    const narrow = this.getDestinationNarrow();
```
gnprice, line 320: ComposeBox: Send "stop typing" notification to destination narrow.
    
    This sends the "stop typing" notification to the destination narrow,
    rather than the narrow that the ComposeBox happened to be loaded into.
    
    This isn't totally correct - we may still miss sending some stop typing
    notifications that we should send, if a user makes edits to the message
    and then later edits the topic, but this change is still an improvement.

This change is good, but the commit message should mention that all of this is in the future when we start sending typing notifications for stream messages at all; i.e., this is a latent bug, lying in wait for when we add that feature.

Otherwise it's confusing for someone who knows that we only send typing notifications for PMs, as they try to figure out what scenario the commit message is talking about in which we're sending typing notifications and then the destination narrow changes.

Hmm, in fact: I think even in that future, this won't at least primarily be about the destination narrow *changing*. The typical case where `this.getDestinationNarrow()` is different from `this.props.narrow` -- the only case outside of message-editing scenarios -- is when you're in a stream narrow, in which case the destination narrow is always some topic narrow, so always different from the stream narrow. So the latent bug this is fixing is that when we go to add typing notifications for stream messages, when you're in a stream narrow we'd try to send these stopped-typing notifications to the stream narrow, instead of the specific topic. Probably that would look like not sending them at all -- probably `sendTypingStop` would ignore them, just like it currently does everything but PM narrows.

---

… Hmm and actually, looking at the implementation of `getDestinationNarrow`, a `this.props.narrow` of a stream narrow is the *only* case where it returns something different from the prop. That seems like a bug -- if you're in a topic narrow, start editing a message, and edit the topic input, the right answer should use the value from the topic input.

Does that affect any of the other changes in this branch? I feel like there were some refactorings that relied on `getDestinationNarrow`, and may have turned a latent bug (that `getDestinationNarrow` returns a wrong value) into a live one (by using that value in a case where we didn't before.)

src/compose/ComposeBox.js
```diff
@@ -320,15 +315,11 @@ class ComposeBox extends PureComponent<Props, State> {
...
 
-    if (!caughtUp.newer) {
-      showErrorAlert(_('Failed to send message'));
-      return;
-    }
```
gnprice, line 329: Will this reintroduce #3800? If not, why not?

In any case I don't think this benefits from being rolled into a long cleanup series like this one -- it's entirely about a behavior change, not anything to do with cleaning up the code. Better to have its own PR thread, to avoid adding complexity to this one.

src/compose/ComposeBox.js
```diff
@@ -278,13 +276,11 @@ class ComposeBox extends PureComponent<Props, State> {
   };
 
   handleMessageFocus = () => {
-    this.setState((state, { lastMessageTopic }) => ({
-      ...state,
-      topic: state.topic || lastMessageTopic,
```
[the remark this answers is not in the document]
gnprice: Oh, a small other thing from the commit message:

    Fixes: #4764

This should also appear in the PR description. That way GitHub links from the issue directly to the PR. See style guide:
https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#mention-fixed-issue

gnprice (commented):
Thanks @WesleyAC, and thanks @chrisbobbe for the previous rounds of review! Comments below.

src/compose/ComposeBox.js
WesleyAC: This does — I hadn't realized this history of this. I had thought that something about this PR made the symptoms described in [#mobile > Failed to send on Android](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Failed.20to.20send.20on.20Android) worse, but looking at it again, I don't think that's the case.

WesleyAC:
Thanks for the review @gnprice! I've addressed everything except the `onLayout` comment — that's the last commit here, so I'd appreciate a re-review and merging the beginning of the series while I look into that.

src/compose/ComposeBox.js
WesleyAC: I fixed the `getDestinationNarrow` bug in a new commit, but I think that may actually break editing messages. Looking into this more...

WesleyAC: Oh, I was seeing #4830, which reproduces on master :/

I believe that the code as it is now is no more incorrect than it was previously, but I obviously could be missing something.

src/compose/ComposeBox.js
```diff
@@ -336,23 +336,6 @@ class ComposeBox extends PureComponent<Props, State> {
     dispatch(sendTypingStop(narrow));
   };
 
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
```
chrisbobbe, line 362:
```
ComposeBox: Stop using UNSAFE_componentWillReceiveProps.

This replaces UNSAFE_componentWillReceiveProps with a key prop, so that
we rerender the ComposeBox when we look at a different narrow or
start/stop editing a message.
[...]
```

This isn't correct: it doesn't cause the ComposeBox to rerender in the normal sense (that `render` is called, and any `componentDidUpdate` is called, etc.). The crucial thing that's happening, and that makes this strategy work, is that the old one gets completely torn down and a new one is created. For example, we have this code that sets the initial state when a `ComposeBox` instance is constructed:

```
  state = {
    isMessageFocused: false,
    isTopicFocused: false,
    isFocused: false,
    isMenuExpanded: false,
    topic:
      this.props.initialTopic !== undefined
        ? this.props.initialTopic
        : isTopicNarrow(this.props.narrow)
        ? topicOfNarrow(this.props.narrow)
        : '',
    message: this.props.initialMessage || '',
    selection: { start: 0, end: 0 },
  };
```

That's how `state.topic` and `state.message` end up being set to the desired values when the `key` changes. It's this ["instance field"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields#public_instance_fields) bit, which I think of as a shortcut to setting things in the [constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor). It only runs once during the lifecycle of an instance.

chrisbobbe: Also (and this is more of a nit), the `key` isn't a prop; I think I usually call it the "`key` attribute`. IIRC you can't access it on `this.props`: https://reactjs.org/docs/lists-and-keys.html

chrisbobbe: Here's a nice explanation of the strategy we've chosen, which I linked to in the CZO discussion: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

WesleyAC: Thanks for the explanation! Reworded the commit.

WesleyAC:
I've rebased this and fixed several merge conflicts. @gnprice / @chrisbobbe I'd appreciate another review on this, since I think it's blocking several other changes people are interested in making on the composebox.

src/chat/ChatScreen.js
```diff
@@ -125,20 +128,41 @@ export default function ChatScreen(props: Props) {
...
-  const sendMessage = useCallback(
+  const sendCallback = useCallback(
     (message: string, destinationNarrow: Narrow) => {
-      if (!caughtUp.newer) {
-        showErrorAlert(_('Failed to send message'));
-        return;
-      }
+      if (editMessage) {
+        const content = editMessage.content !== message ? message : undefined;
+        const subject = caseNarrowDefault(
```
chrisbobbe, line 140: Why the change to start using `caseNarrowDefault`? I think it belongs in a different commit; this one is described as a simple "switch from A to B" commit.

Ah, looking again: possibly this was a small rebase error? 🙂

src/chat/ChatScreen.js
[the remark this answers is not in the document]
chrisbobbe: Cool, glad to start having `autoFocusMessage`, `initialTopic`, and `initialMessage` depend on `editMessage` in this commit. In the commit message, let's write down our reason for making those changes in this commit, as discussed above (https://github.com/zulip/zulip-mobile/pull/4773/commits/4691ff273c7806a56e8319c03dad66ce693d7a7d#r648445186), so that it's clear and inspectable in the future. 🙂

Also, as I look again, I'm glad to have the mention of the "Warning: Cannot update a component from inside…" fix in the commit message, since this sure seems to be the commit that changes that behavior for the better. But while it's nice (and potentially really helpful for responding to bug reports) to know that the specific behavior is improved, it's also true that we didn't end up with a really precise diagnosis, and I think we should mention that. As-is, I think it's possible for someone to follow the link after "See the CZO thread" in hopes of finding a diagnosis, and they'll be disappointed and possibly confused when they don't find one.

chrisbobbe (commented):
Thanks, @WesleyAC, this is getting closer! I think at some point we left this as @gnprice's to review, but I've looked again myself and found some points that I hope will be productive.

src/compose/ComposeBox.js
```diff
@@ -363,7 +364,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
   getDestinationNarrow = (): Narrow => {
     const { narrow } = this.props;
-    if (isStreamNarrow(narrow)) {
+    if (isStreamNarrow(narrow) || isTopicNarrow(narrow)) {
```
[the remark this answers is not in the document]
WesleyAC: Done :)

src/compose/ComposeBox.js
```diff
@@ -83,6 +81,9 @@ type Props = $ReadOnly<{|
   editMessage: EditMessage | null,
   completeEditMessage: () => void,
 
+  initialMessage?: string,
```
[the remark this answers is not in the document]
WesleyAC: Done :)

src/chat/ChatScreen.js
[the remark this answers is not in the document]
WesleyAC: Added a comment about this in the commit message. I don't think it's as mysterious as our initial conversation made it out to be — there was some logic in `UNSAFE_componentWillReceiveProps`, and those props are a way to replicate that in a non-imperative way, but it still seems worth mentioning. Also updated the bit about the error message disappearing.

src/compose/ComposeBox.js
```diff
@@ -84,6 +84,9 @@ type Props = $ReadOnly<{|
   initialMessage?: string,
   initialTopic?: string,
 
+  autoFocusTopic: boolean,
```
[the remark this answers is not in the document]
WesleyAC: Done :)

src/chat/ChatScreen.js
WesleyAC: Hmmm, I think I had this in there the whole time, and I agree that the commit doesn't describe that. The reason that it's needed is that the previous code could look directly at the `state.topic`, but the callback doesn't have access to the state.

I've added a commit to make that change in the original code.

WesleyAC:
@chrisbobbe I addressed all of those comments, should be ready for another look!

src/chat/ChatScreen.js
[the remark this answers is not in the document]
chrisbobbe: > I don't think it's as mysterious as our initial conversation made it out to be

Understood. 🙂 But we depend heavily on reading a clear commit history, and I think I've mentioned elsewhere that the `key` change is somewhat likely to introduce puzzling regressions, almost no matter how carefully we do it. So it'll be nice to have these details at hand if that happens, and not have to guess as much about why we made certain changes and whether the reasons were good based on what we knew at the time.

Also, I think explaining these prop changes does more than just answer why we made them: it clarifies a bit the limits of what we considered. Without these explanations, I think someone in the future is more likely to assume that we deduced that these changes were _all_ we had to do toward replicating the logic in `UNSAFE_componentWillReceiveProps` in a non-imperative way. We didn't deduce that; we started with some prominent behaviors that we wanted to keep, without considering some small-looking things at all, like a `textInput.setNativeProps` call (whatever that's about) and everything that `TextInputReset` does. I think that's probably an OK tradeoff, so that we don't spend too long on this. But it is a tradeoff, and I think it'll be more obvious that we made the tradeoff after listing these things that we did do something about.

src/compose/ComposeBox.js
```diff
@@ -363,7 +377,7 @@ class ComposeBox extends PureComponent<Props, State> {
 
   getDestinationNarrow = (): Narrow => {
     const { narrow } = this.props;
-    if (isStreamNarrow(narrow)) {
+    if (isStreamNarrow(narrow) || isTopicNarrow(narrow)) {
       const streamName = streamNameOfNarrow(narrow);
       const topic = this.state.topic.trim();
       return topicNarrow(streamName, topic || '(no topic)');
```
gnprice, line 383: Does this do the right thing in the case where you're in a topic narrow and simply composing a new message, rather than editing? We don't have a topic input then, so it's not obvious that we would have a `state.topic` and have the right value there.

It looks like the answer is that it doesn't, as of the commit where this change is made. But then a later change adds more logic to the `state.topic` initializer so that it does.

Probably the easiest fix is to pull that logic out to its own small prep commit: a commit that makes `state.topic` always have the topic we're going to send to (if we'd send a stream message at all), which will simplify several changes we want to make in this branch.

src/compose/ComposeBox.js
```diff
@@ -470,7 +426,7 @@ class ComposeBox extends PureComponent<Props, State> {
             onAutocomplete={this.handleMessageAutocomplete}
           />
         </View>
-        <View style={[this.styles.composeBox, style]} onLayout={this.handleLayoutChange}>
```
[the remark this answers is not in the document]
gnprice: Looks like the revised version drops this change, which is fine.

Meanwhile we have an issue now about the autocomplete widget's placement, so I copied my digression above into a comment there:
https://github.com/zulip/zulip-mobile/issues/4871#issuecomment-876028076

src/compose/ComposeBox.js
WesleyAC: > It looks like the answer is that it doesn't, as of the commit where this change is made. But then a later change adds more logic to the `state.topic` initializer so that it does.

I don't think this is true? In this commit, `state.topic` is set like so:

```javascript
    topic: this.props.lastMessageTopic,
```

Which we get here:

```javascript
    lastMessageTopic: getLastMessageTopic(state, props.narrow),
```

So I think `state.topic` should come from whatever's the last message is in `props.narrow` — in a topic narrow, that should be the correct topic.

Am I missing something here?

gnprice: Hmm, I guess that gets it closer! I hadn't made that connection.

I think there's still buggy behavior at least latently, in that if you're looking at a topic narrow and it's *empty*, `getLastMessageTopic` will return `''` instead of the topic from the narrow. I say "latently" because I'm not sure there's a way to navigate to a narrow with no messages in it...

Ah, no, actually I think that `''` case will happen all the time. The thing is that because this is the state initializer (in the constructor), it's getting run only when the `ComposeBox` instance is first created, and doesn't reflect updates to the props. And any time you visit a given narrow for the first time, we'll have nothing in `state.narrows` at that point.

Anyway. If some chain of reasoning like that does end up finding that the behavior works out correctly, it's definitely the sort of thing I'd want spelled out explicitly :slightly_smiling_face: (And then if it weren't already going away later in the branch, would want to try to find some refactor so that if possible we didn't rely on that kind of non-local connection in steady state.)

src/compose/ComposeBox.js
```diff
@@ -155,8 +164,13 @@ class ComposeBox extends PureComponent<Props, State> {
     isFocused: false,
     isMenuExpanded: false,
     height: 20,
-    topic: this.props.lastMessageTopic,
-    message: this.props.draft,
+    topic:
+      this.props.initialTopic !== undefined
+        ? this.props.initialTopic
```
gnprice, line 169: nit: can use `??` to make this slightly simpler to read

src/compose/ComposeBox.js
```diff
@@ -155,8 +164,13 @@ class ComposeBox extends PureComponent<Props, State> {
...
+        : isTopicNarrow(this.props.narrow)
+        ? topicOfNarrow(this.props.narrow)
+        : '',
+    message: this.props.initialMessage || '',
```
gnprice, line 173: nit: `??` seems a bit more precisely what we mean here, though the behavior is the same; also makes it parallel the `topic` case once that uses `??`

gnprice (commented):
Thanks for the revisions -- this is looking good! I think my most significant comment is the one above at https://github.com/zulip/zulip-mobile/pull/4773#discussion_r665786059 . A couple of nits below.

src/compose/ComposeBox.js
gnprice: That fix looks good, thanks! (Modulo the thread at https://github.com/zulip/zulip-mobile/pull/4773#discussion_r665786059 .)

PR #4778, opened 2021-06-04
dependabot: Add config file.

chrisbobbe:
Currently, dependabot will automatically open PRs to address
reported security vulnerabilities in packages. I'm interested to see
if it can help us be better about keeping our packages up-to-date in
general.

There are plenty of config options we can choose from (see the
linked doc). For now, I'm just curious to see what it does with this
simple config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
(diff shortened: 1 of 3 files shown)


WesleyAC (approved):
LGTM! I'm a bit worried about noise, but happy to try it out and see what happens.

chrisbobbe:
Thanks for the review!

We can always lower `open-pull-requests-limit` down from 5 to reduce noise (5 is the default, but I added it for explicitness so it's visible as something that's super simple to change).

chrisbobbe:
I like that Dependabot links (and sometimes quotes) the changelog for the package; it saves some clicking around in GitHub which can be slow.

PR #4779, opened 2021-06-04
deps: Remove prop-types from devDependencies.

chrisbobbe:
We added it in 5b6014e83 to silence a peer-dep warning that
react-intl hasn't been throwing for a while:
  https://github.com/formatjs/formatjs/issues/1382

Note that react-intl's package.json doesn't have prop-types in its
peerDependencies anymore:
  https://github.com/formatjs/formatjs/blob/react-intl%405.8.6/packages/react-intl/package.json#L144-L147

The package is for runtime checking for React props; we don't need
it because we use Flow:
  https://www.npmjs.com/package/prop-types

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,6 @@
     "prettier": "^1.18.2",
     "prettier-eslint": "^12.0.0",
     "prettier-eslint-cli": "^5.0.0",
-    "prop-types": "^15.7.2",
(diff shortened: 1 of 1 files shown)


WesleyAC (approved):
LGTM, thanks!

chrisbobbe:

PR #4787, opened 2021-06-07
dependabot: Remove config file.

chrisbobbe:
After trying out Dependabot with some simple config for a few days,
we're not happy with the noise it produces; see discussion on CZO:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Dependabot/near/1202812


gnprice:
Merged, thanks!

PR #4789, opened 2021-06-08
Run `yarn upgrade`, after treating some deps individually.

chrisbobbe:
This is a lot at once, but all our tests pass, and manual testing
confirms that basic functionality is preserved on Android and iOS.
We should make sure this spends some time out in a beta release
before going out to all users.

Greg points out that "it's more practical to seriously test an
infrequent bigger upgrade than many small ones" [1], and that "if we
have a regression that we bisect to a big upgrade commit, we can
still do a bisect on the upgrades themselves within the commit" [2].

[1] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Dependabot/near/1202825
[2] https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Dependabot/near/1202866

-----

In this revision, `tools/test --all` passed for me on every commit.

For the react-native-safari-view -> expo-web-browser change, I tested social auth, a link within a message, and the "Forgot password?" link.


ios/Podfile.lock
```diff
@@ -362,7 +362,7 @@ PODS:
     - React
   - RNReanimated (1.13.0):
     - React
-  - RNSentry (2.2.1):
+  - RNSentry (2.4.3):
```
WesleyAC, line 365: > We should have done this in 7dce4e7. It looks like it was missed
> because the quality-of-life improvement in 683f387 came with the
> side effect of skipping `pod install` even when autolinking wants to
> upgrade the Podfile.lock because of an NPM package upgrade. Hmm.

Any ideas/plans on how to fix this in general?

package.json
```diff
@@ -124,7 +124,7 @@
     "prettier": "^1.18.2",
     "prettier-eslint": "^12.0.0",
     "prettier-eslint-cli": "^5.0.0",
-    "react-dom": "16.11.0",
+    "react-dom": "16.13.1",
```
WesleyAC, line 128: >(see history)

(in the commit message) — can you give a commit hash to look at here?

package.json
```diff
@@ -57,7 +57,7 @@
     "lodash.union": "^4.6.0",
     "lodash.uniqby": "^4.4.0",
     "react": "16.13.1",
-    "react-intl": "5.8.6",
+    "react-intl": "5.20.1",
```
WesleyAC, line 60: Did you test using the app in a different language?

yarn.lock
```diff
@@ -9556,9 +9556,9 @@ prettier-linter-helpers@^1.0.0:
     fast-diff "^1.1.2"
 
 prettier@^1.18.2, prettier@^1.7.0:
-  version "1.18.2"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-1.18.2.tgz#6823e7c5900017b4bd3acf46fe9ac4b4d7bda9ea"
-  integrity sha512-OeHeMc0JhFE9idD4ZdtNibzY0+TPHSpSSb9h8FqtP+YnoZZ1sl8Vc9b1sasjfymH3SonAF4QcA2+mzHPhMvIiw==
+  version "1.19.1"
```
WesleyAC, line 9630: > Looks like this requires quite a few formatting changes, and I'm not
> sure I like all of them -- in test files, especially, it's nice when
> we can fit something simple onto a single line. Ah, well.

Can we change the settings to not require the changes that we don't like? I'd prefer to just drop this commit from this PR, so we can talk about formatting changes separately from the rest of these updates.

WesleyAC (commented):
Thanks! Looks good overall, just a few comments.

yarn.lock
chrisbobbe: Sure. Prettier isn't very configurable, and my hopes aren't very high that we'll be able to get what we want in Prettier's config. Possibly there's something we could do in `tools/formatting.eslintrc.yaml`, which (with prettier-eslint-cli) we use to overrule some of Prettier's opinions by running ESLint with some targeted rules.

For strategy on removing the commit from this PR (while keeping the headline big-upgrade commit):

I'm wary of making the big upgrade commit more complicated; it would be great not to have to change the command there from `yarn upgrade` to something like `yarn upgrade $(tools/deps | grep -vx -e prettier)` and explain what `prettier` is doing in the command there.

One option would be to have a commit that pins `prettier` to its current version (using a range without a caret), and then `yarn upgrade` would see that and not try to upgrade it.

package.json
chrisbobbe: Yes; I'll make a note that I did, in the commit message.

yarn.lock
gnprice: > One option would be to have a commit that pins `prettier` to its current version (using a range without a caret), and then `yarn upgrade` would see that and not try to upgrade it. Then after this PR is in, we can think about upgrading `prettier` and adding a caret back.

Yeah, that'd be a fine alternative to the `$(tools/deps | grep -v …)` thing.

src/pm-conversations/__tests__/pmConversationsModel-test.js
```diff
@@ -38,7 +38,11 @@ describe('reducer', () => {
         { user_ids: [2, 1].map(makeUserId), max_message_id: 345 }, // user_ids out of order
       ];
       const expected = {
-        map: Immutable.Map([['', 234], ['1', 123], ['1,2', 345]]),
+        map: Immutable.Map([
+          ['', 234],
+          ['1', 123],
+          ['1,2', 345],
+        ]),
```
gnprice, line 45: (For when we do the Prettier upgrade, whether now or later:)

If you feel like just adding some prettier-ignore comments on this kind of diff hunk, to keep these blobs of test data compact, I'd be for it.

src/account/accountsSelectors.js
```diff
@@ -96,15 +96,12 @@ export const tryGetCurrentRealm = (state: GlobalState): URL | void =>
...
-    }
-    return authOfAccount(account);
-  },
-);
+export const tryGetAuth: Selector<Auth | void> = createSelector(tryGetActiveAccount, account => {
+  if (!account || account.apiKey === '') {
+    return undefined;
+  }
+  return authOfAccount(account);
+});
```
gnprice, line 104: This change OTOH seems like an improvement.

src/account/accountsSelectors.js
```diff
@@ -14,10 +14,10 @@ export type AccountStatus = {| ...Identity, isLoggedIn: boolean |};
...
  * actually need the API keys, but just need to know whether we have them.
  */
-export const getAccountStatuses: Selector<$ReadOnlyArray<AccountStatus>> = createSelector(
-  getAccounts,
-  accounts =>
-    accounts.map(({ realm, email, apiKey }) => ({ realm, email, isLoggedIn: apiKey !== '' })),
+export const getAccountStatuses: Selector<
+  $ReadOnlyArray<AccountStatus>,
+> = createSelector(getAccounts, accounts =>
+  accounts.map(({ realm, email, apiKey }) => ({ realm, email, isLoggedIn: apiKey !== '' })),
```
gnprice, line 20: And this change seems one step forward, one step back -- the more compact `createSelector` arguments seem nice, but the type parameter set off on its own line is silly. On net, probably not worth fighting Prettier over.

yarn.lock
gnprice: Also note that we can always disable Prettier locally with `// prettier-ignore` -- that applies to just the immediately following statement or expression, so it's generally pretty finely targeted. We have a few dozen of those already, and I think it's appropriate to have a fairly low threshold for applying it when Prettier tries to do something ugly.

gnprice (commented):
Thanks for working through these! All looks good modulo Wesley's comments above and a few small other comments.

> For the react-native-safari-view -> expo-web-browser change, I tested social auth, a link within a message, and the "Forgot password?" link.

This tidbit would be helpful to add to that commit's commit message. For one thing, it helps us find that to jog our memory for things to test in future major upgrades to that library.

ios/Podfile.lock
chrisbobbe: Hmm. I think I'll revert 683f3875e for now, and ponder a better fix later.

chrisbobbe:
Thanks for the reviews! Revision pushed.

gnprice:
Thanks for the revision! All LGTM.

I'll let Wesley comment on whether they think the remaining Prettier format changes still need more discussion (https://github.com/zulip/zulip-mobile/pull/4789#discussion_r647705953 ), but otherwise I think this is ready to merge.

WesleyAC:
I'm fine to merge this with the prettier changes.

chrisbobbe:
Great, thanks for the reviews! Merged.

Issue #4600, opened 2021-04-05
Support a future event (shape TBD) that marks messages unread.


chrisbobbe:
See [discussion](https://chat.zulip.org/#narrow/stream/137-feedback/topic/Make.20the.20message.20unread.20if.20it's.20edited.2E/near/1152595).

gnprice:
API discussion is in [this thread](https://chat.zulip.org/#narrow/stream/378-api-design/topic/mark-as-unread.20.28update_message_flags.29/near/1204350) (long).

A server-side draft is at https://github.com/zulip/zulip/pull/19307 .

gnprice:
Oops, we ended up duplicating this issue as #5361.

I think that one has wound up with more things linking to it, so I'll close this one as a duplicate in favor of that one.

PR #4790, opened 2021-06-08
unreads: Support removing read flag from stream messages.

WesleyAC:
Issue: ~~#4600~~ #5361

### DIFF
--- a/src/unread/unreadPmsReducer.js
+++ b/src/unread/unreadPmsReducer.js
@@ -54,7 +54,7 @@ const eventUpdateMessageFlags = (state, action, ownUserId) => {
       return state;
     }
 
-    let newState = [...state];
+    let newState = state;
 
     for (const id of action.messages) {
       const message = message_details.get(id);
@@ -68,12 +68,10 @@ const eventUpdateMessageFlags = (state, action, ownUserId) => {
       }
     }
 
-    return [
-      ...newState.map(({ sender_id, unread_message_ids }) => ({
-        sender_id,
-        unread_message_ids: [...unread_message_ids].sort((a, b) => a - b),
-      })),
-    ];
+    return newState.map(({ sender_id, unread_message_ids }) => ({
+      sender_id,
+      unread_message_ids: [...unread_message_ids].sort((a, b) => a - b),
+    }));
   }
 
   return state;


gnprice:
Thanks @WesleyAC !

* With what it [looks like](https://chat.zulip.org/#narrow/stream/378-api-design/topic/update.20message.20flags.20event/near/1204473) the API is going to become (reusing the unreads data structure from the `/register` response), we'll be taking in the same data here that's handled in the `REALM_INIT` case. So you'll want to factor that out and reuse it.

  See also the commit message of 5725e7d8f375f2feba6e0116d1f5dc78afda4cf3 (which added that) for some reasoning that drove the design of that bit of code, and can be useful for processing other data, particularly into complex data structures like from `Immutable`.

* This point might become moot with that API change, but FFR: for using `Immutable.Map#update`, see the `updateIn` call in the `EVENT_NEW_MESSAGE` case above. (In addition to `updateIn` itself being handy, the use of default arguments cleans it up a bit.)

* An `Immutable` nit: the documented usage is to write just `Immutable.Map()` (and similarly `List` etc.), with no `new`.

* I also agree with [Anders's suggestion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flow.20filter.2Fmap/near/1204493) to avoid `mergeDeepWith` because it lacks effective types.

WesleyAC:
Ok, I think this should be ready for review! It's not using the new API (since I think we'll want to support the old one as well, so that we can support mark-as-unread on mobile on old servers).

Other than that, I think I got to all of your comments @gnprice :)

chrisbobbe:
Do we also need to do something to `state.flags.read`, in `flagsReducer`? If so, see also a potential oddness in how we're currently handling `UPDATE_MESSAGE_FLAGS` there: https://chat.zulip.org/#narrow/stream/378-api-design/topic/UPDATE_MESSAGE_FLAGS.20.60all.3A.20true.60/near/1222451

src/unread/unreadModel.js
```diff
@@ -190,8 +190,36 @@ function streamsReducer(
...
+        let newlyUnreadState = new Immutable.Map();
+
+        for (const id of action.messages) {
+          const message = action.allMessages.get(id);
+
+          if (message && message.type === 'stream') {
+            newlyUnreadState = newlyUnreadState.updateIn(
+              [message.stream_id, message.subject],
+              new Immutable.List(),
+              messages => messages.push(message.id).sort(),
```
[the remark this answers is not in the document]
WesleyAC: This sort is needed in the case where the topic is not already in the state, I think.

I'm not sure what the best way to deal with that is.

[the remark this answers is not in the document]
gnprice: Why's it needed here, though? To expand a bit on what I said about a sort at the end: I believe the one place this `messages` list ends up getting looked at, other than on future iterations of this same line, is this spot:
```js
              ) => oldUnreadMessages.concat(newUnreadMessages).sort(),
```
where it gets called `newUnreadMessages`.

So at that point we concatenate it with another list, and then sort the result.

When we do that, it doesn't matter whether `newUnreadMessages` had been sorted or not -- the end result will be sorted in any case.

(We do need that sort at the end, though. Even with this earlier sorting, the new unread messages might be earlier or later than the ones in `oldUnreadMessages`.)

src/unread/unreadModel.js
```diff
@@ -190,8 +190,40 @@ function streamsReducer(
...
+          // This server is too old to support marking messages as unread.
+          return state;
+        }
+
+        let newlyUnreadState = Immutable.Map();
+
+        for (const id of action.messages) {
+          const message = action.message_info_dict?.get(id);
+
+          if (message && message.type === 'stream') {
```
gnprice, line 203: The `?.` and `message &&` indicate uncertainty about whether `action.message_info_dict` is present. But that shouldn't be necessary at this point, because we have a conditional above bailing out if it's absent -- we should be able to write this code without these extra control-flow paths to worry about.

The reason Flow isn't sure that it's present is because of that function call in between, `Immutable.Map()` -- it doesn't have a way to prove that that doesn't have side effects on the object `action`. So the usual way to fix that and make the condition be as effective as it should be is to make a local:
```js
  const { message_info_dict } = action;
  if (!message_info_dict) {
    // This server etc.
    return state;
  }
```
and use that local thereafter.

gnprice: Then I guess the `message &&` reflects another uncertainty, which is from the `get` -- it's possible that `message_info_dict` is missing something that's in `messages`.

That would be a server bug, though, IIUC. So really our logic is assuming that doesn't happen. That should be made explicit, either with a comment or by logging a warning while bailing out. Here a comment seems fine; the alternative would be more complicated, and this seems like an unlikely server bug so there's not much to gain from logging a warning for it.

src/unread/unreadModel.js
```diff
@@ -190,8 +190,36 @@ function streamsReducer(
...
+              messages => messages.push(message.id).sort(),
+            );
+          }
+        }
+
+        return state.mergeWith(
+          (
+            oldTopicsMap: Immutable.Map<string, Immutable.List<number>>,
+            newTopicsMap: Immutable.Map<string, Immutable.List<number>>,
+            streamKey: number,
```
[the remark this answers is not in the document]
gnprice: Ditto on `topicKey` below :-)

src/events/eventToAction.js
```diff
@@ -292,6 +292,10 @@ export default (state: GlobalState, event: $FlowFixMe): EventAction | null => {
         // stop sending `operation`; see #4238.
         op: event.op ?? event.operation,
 
+        message_info_dict: event.message_info_dict
+          ? new Map(Object.entries(event.message_info_dict).map(([k, v]) => [parseInt(k, 10), v]))
+          : undefined,
+
```
gnprice, line 298: Probably cleanest to push this conversion down into the one reducer that consumes this action, so that the action type doesn't need to diverge from the event (more than it does already).

(We discussed this when you asked about it in person; mentioning here FTR in the thread.)

src/unread/__tests__/unreadModel-test.js
```diff
@@ -212,9 +220,38 @@ describe('stream substate', () => {
...
+              stream_id: 123,
+              topic: 'foo',
+              unmuted_stream_msg: true,
+            },
+          ],
+        ]),
+      });
+      const newState = reducer(baseState, action, eg.plusReduxState);
+      expect(newState.streams.valueSeq().flatten()).toContain(42);
+      expect(newState.streams.valueSeq().flatten()).toContain(99);
```
gnprice, line 254: Let's make these expectations more specific. The `summary` helper used elsewhere in this file makes it pretty clean and compact to just say exactly what the state should look like.

One bit that would be good to include in a test case is that the IDs for a given conversation should be sorted, even if that means they're interleaved with the existing ones.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -212,9 +220,38 @@ describe('stream substate', () => {
...
+          [
+            42,
+            {
+              type: 'stream',
+              mentioned: false,
+              sender_id: eg.selfUser.user_id,
+              stream_id: 123,
+              topic: 'foo',
+              unmuted_stream_msg: true,
+            },
```
gnprice, line 237: Let's put this in a little helper (perhaps even local to this test case), taking just `stream_id` and `topic` as arguments. Then all this mess only appears once, and this test data gets a lot more compact -- like one line per message. That helps a lot in reading through these tests to see what kinds of input they're covering.

gnprice (commented):
Thanks for the revision! This generally looks good. Comments below, and also above because GitHub.

WesleyAC:
Revision pushed — I still need to add the test case you mentioned, and update the mentioned messages datastructure (as mentioned in person), will also bump this once I do that.

gnprice:
Thanks for the revision! I took a look although there are still more changes you plan to make. All looks good modulo the two things you mentioned:

> I still need to add the test case you mentioned, and update the mentioned messages datastructure (as mentioned in person), will also bump this once I do that.

plus one commit-message comment: should mention the issue, #4600.

WesleyAC:
Getting back to this, I'm having a hard time remembering/parsing what the test you wanted added is, @gnprice. You wrote:

> One bit that would be good to include in a test case is that the IDs for a given conversation should be sorted, even if that means they're interleaved with the existing ones.

In reference to to the 'when operation is "remove", add stream message to unreads' test. What does "interleaved with existing ones" mean in this context? Were you referring to the message IDs in the message_info_dict as possibly being interleaved? If so, it looks like I did fix that at some point. Let me know if there's anything you're still unhappy about regarding the tests :)

WesleyAC:
@gnprice should be all working and ready for review now!

It's stacked on top of #5227, since one of the lints in `unreadModel.js` has the problem where prettier and prettier-eslint disagree.

I've tested this with a dev server running https://github.com/zulip/zulip/pull/18743 and it works fine, in addition to the automated tests.

gnprice (commented):
Thanks @WesleyAC for the revisions! Small comments below.

> In reference to to the 'when operation is "remove", add stream message to unreads' test. What does "interleaved with existing ones" mean in this context? Were you referring to the message IDs in the message_info_dict as possibly being interleaved? If so, it looks like I did fix that at some point. Let me know if there's anything you're still unhappy about regarding the tests :)

What I meant here was to have a test case where a message we're marking as unread falls in between some messages that are already unread, in a given conversation.

E.g., give the starting state (in the same way as in the "when removing, don't touch" case above) another message with ID 100, in any of the conversations where you're adding messages. Then the expected state will have that message at the end of the list. That will exercise the fact that we're sorting the lists together at the end, vs. a possible incorrect implementation where the merge just concatenated the lists.

---

Zooming out and thinking about the workflow here:
* I think we probably want to hold off actually merging this until the new API is settled.
  * This doesn't mean waiting for the server side to be actually merged -- I think the server and mobile PRs can go in in either order.
  * It probably does mean having a server PR that's *close* to merge, as doing the implementation can lead to feedback on the API choices. Historically we've gotten that feedback only from the server and the web frontend before committing to an API; this feature will be the first of a lovely new pattern where we benefit from a mobile implementation before that point too!
* So I think the right next step here, other than the small changes in comments below, is to revive the `#api design` [thread](https://chat.zulip.org/#narrow/stream/378-api-design/topic/mark-as-unread.20.28update_message_flags.29/near/1253338). I guess I'll go bump that now.
  * (Because we're currently in the last few weeks of the 5.0 release cycle, I don't expect anything to be merged until that's out, but possibly there can be a revision of the server PR for the API feedback we've given in that thread.)

---

Oh, and one small commit-message-level thing: the reference to the issue shouldn't be "Fixes:", because this is only the stream-messages side so we don't want this to close it. Could do "Fixes-part-of:".

src/api/eventTypes.js
```diff
@@ -154,6 +154,18 @@ export type UpdateMessageFlagsEvent = $ReadOnly<{|
   flag: empty, // TODO fill in
   all: boolean,
   messages: $ReadOnlyArray<number>,
+
+  message_info_dict?: {|
```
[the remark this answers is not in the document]
WesleyAC: I'm not sure I see what the idea behind the `TODO` would be — other examples of it I see are for marking things as required or dropping them, but this field will need to be optional for the foreseeable future, since it's not used for `op: add`.

I've added a comment stating what version it was added in, but I'm not convinced that's generally useful, given that we don't seem to have a habit of doing that elsewhere in this file.

WesleyAC:
Updated!

* Switched to use new API
* Removed "Fixes" from commit message
* Changed mentions test in the way you suggested (not re-adding a already-unread mention)
* Added links to API docs
* Added comment about what server version message_details was added in
* Added warnings for events without `message_details`.

Should be ready for another look @gnprice

src/api/eventTypes.js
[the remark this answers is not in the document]
gnprice: Hmm I see, right, it remains optional even in the new API.

Yeah, I don't think there's a need for a comment when there isn't something actionable for us to eventually do with it.

src/api/eventTypes.js
```diff
@@ -158,6 +160,18 @@ export type UpdateMessageFlagsEvent = $ReadOnly<{|
   flag: empty, // TODO fill in
   all: boolean,
   messages: $ReadOnlyArray<number>,
+
+  // Added in server-5.0
+  message_details?: {|
+    [string]: {|
+      type: string,
```
gnprice, line 167: This can be more specific:
```suggestion
      type: 'stream' | 'private' | 'huddle',
```

src/api/eventTypes.js
```diff
@@ -158,6 +160,18 @@ export type UpdateMessageFlagsEvent = $ReadOnly<{|
...
+      mentioned: boolean,
```
gnprice, line 168: This can be updated with a recent improvement to the API:
```suggestion
      mentioned?: true,
```

gnprice (commented):
Thanks for the revision! The changes generally look good; replied on the small thread at https://github.com/zulip/zulip-mobile/pull/4790#discussion_r817077107 .

The one thing still missing is a test where the message IDs get interleaved:
https://github.com/zulip/zulip-mobile/pull/4790#pullrequestreview-883974830
That and a couple of tweaks below.

Then at that point, I think this will be ready to merge! In particular I think the `#api design` thread looks like we've reached consensus on the API for stream messages. I guess I'd explicitly check that on the thread before merging, but I think the answer will be yes.

WesleyAC:
Ok, @gnprice, I think I addressed all that :)

* Added a message with ID 100 in the test that starts out unread.
* `mentioned` is now optional `true` instead of required `boolean`
* `type` is now `'stream' | 'private'`
* Uses the new new API, with a `user_ids` array
* Now implemented for PMs (as a separate commit)

Still to do:

* Huddles
* Clean up tests a bit (so I don't have to stick `ownUserId: eg.selfUser.user_id` everywhere)

src/api/eventTypes.js
```diff
@@ -159,6 +161,18 @@ export type UpdateMessageFlagsEvent = $ReadOnly<{|
...
+
+  message_details?: {|
+    [string]: {|
+      type: 'stream' | 'private',
+      mentioned?: true,
+      // type: 'stream'
+      stream_id?: number,
+      topic?: string,
+      // type: 'private'
+      user_ids?: UserId,
```
gnprice, line 173: This is best expressed as a union of two object types -- that way the relationship between `type` being e.g. `'stream'`, and `stream_id` and `topic` being definitely present, is in the type, and the code that consumes the object can get simpler. I'll demonstrate in a small added commit.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -288,23 +304,45 @@ describe('stream substate', () => {
...
+        [234, Immutable.Map([['bar', [4, 5, 100]]])],
       ]));
 
       const action = mkAction({ messages: [1, 2] });
       const newState = reducer(state, action, eg.plusReduxState);
       // prettier-ignore
       expect(summary(newState)).toEqual(Immutable.Map([
         [123, Immutable.Map([['foo', [3]], ['qux', [7]]])],
-        [234, Immutable.Map([['bar', [4, 5]]])],
+        [234, Immutable.Map([['bar', [4, 5, 100]]])],
```
gnprice, line 315: We can add the extra message locally for just the one test case where it's being used, to avoid adding noise to the other test cases like this. Same way that the "when removing, don't touch unaffected topics or streams" case already adds a message of its own.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -231,15 +231,24 @@ describe('stream substate', () => {
 
   describe('EVENT_UPDATE_MESSAGE_FLAGS', () => {
     const mkAction = args => {
-      const { all = false, messages, flag = 'read', op = 'add' } = args;
+      const {
+        all = false,
+        allMessages = eg.makeMessagesState([]),
```
gnprice, line 236: Looks like this change doesn't get used.

This property (which is kind of a hack) only gets consumed in the `flags` reducer anyway, so it's natural that none of these tests would need to provide it. Best to let these tests forget its existence as much as possible, until the hack gets removed.

src/unread/unreadModel.js
```diff
@@ -243,8 +244,52 @@ function streamsReducer(
       }
 
       if (action.op === 'remove') {
-        // Zulip doesn't support un-reading a message.  Ignore it.
-        return state;
+        const { message_details } = action;
+        if (message_details === undefined) {
+          logging.warn('Got update_message_flags event without message_details.');
```
gnprice, line 249: As is, this is describing a situation that's perfectly fine and normal -- most update_message_flags events won't have message_details. We can make it more specific, to avoid puzzling someone looking at a Sentry event in the future who has an idea how this part of the API works, but has forgotten the details of how we wrote this code:

```suggestion
        if (message_details === undefined) {
          logging.warn('Got update_message_flags/remove/read event without message_details.');
```

(Also to avoid looking wrong for someone reading this code to try to understand it.)

src/events/eventToAction.js
```diff
@@ -318,7 +318,13 @@ export default (state: PerAccountState, event: $FlowFixMe): EventAction | null =
         // TODO(server-4.0): Simplify to just use `op`.
         op: event.op ?? event.operation,
 
+        message_details: event.message_details
+          ? new Map(Object.entries(event.message_details).map(([k, v]) => [parseInt(k, 10), v]))
+          : undefined,
+
         allMessages: state.messages,
+
+        ownUserId: getOwnUserId(state),
```
gnprice, line 327: As I mentioned in another comment just above, the existing `allMessages` thing is kind of a hack.

We have a cleaner solution now, which I introduced in a2000b9c8 and we've applied in a few places since then: the sub-reducers can just take the whole state (specifically the `PerAccountState`, since introducing that distinction) as a third argument. Then they can apply a selector to get whatever particular data they need.

src/unread/unreadPmsReducer.js
```diff
@@ -46,7 +47,32 @@ const eventUpdateMessageFlags = (state, action) => {
...
     return removeItemsDeeply(state, action.messages);
   } else if (action.op === 'remove') {
-    // we do not support that operation
+    const { message_details } = action;
+    if (message_details === undefined) {
+      logging.warn('Got update_message_flags event without message_details.');
+      return state;
+    }
+
+    let newState = [...state];
```
gnprice, line 56: This copying of the array, and also the one at the `return` below, can be left out; we don't mutate the array anywhere.

(The `addItemsToPmArray` helper already avoids mutation.  That can be seen in its type: the first parameter, where we pass this `newState` to, is a `$ReadOnlyArray`.)

This data structure is never going to be a model of efficiency whatever we do with it -- the real fix is to replace it with an efficient data structure, like we've done for unread stream messages in `unreadModel.js`. But we can avoid extra copies where it's easy to do so.

src/unread/unreadPmsReducer.js
```diff
@@ -46,7 +47,32 @@ const eventUpdateMessageFlags = (state, action) => {
   if (action.op === 'add') {
     return removeItemsDeeply(state, action.messages);
   } else if (action.op === 'remove') {
-    // we do not support that operation
```
gnprice, line 49: nit: the commit message says "unreads: Add support for marking PMs as unread", but it doesn't -- it only adds support for 1:1 PMs, not group PMs.

See e.g. user-facing docs:
https://zulip.com/help/reading-pms
where we say "private messages (PMs)" to mean both kinds of PMs together.

In general we're consistent about that in the UI and user-facing docs, as well as in like 95% of the Zulip codebase. Even in parts of the server where a "huddle" is an important concept, if we say "private message" or "PM" we almost always mean a message that's either to a huddle or an individual user, not only the latter. E.g. search zerver/models.py for /private.message/i.

(I know that this `src/unread/` section of the codebase often says "PM" to mean just 1:1 PMs. That's an unfortunate exception, but we can still use the standard terminology in commit messages.)

gnprice (commented):
Thanks @WesleyAC for the revision and all your work on this! Sorry I've let it sit for a couple of months.

A few comments below. To avoid creating another round-trip, I'll push to the branch shortly with fixes for these. Then I'll merge once I do a bit of (manual) testing with a server. (As predicted in the last few comments above, the server side did get merged, in https://github.com/zulip/zulip/pull/21252 -- so the API is all settled.)

That will leave one remaining piece open for supporting these events (i.e. for #5361): group PMs. @WesleyAC, if you have a bit of time, I'd love to see a PR from you to implement that too. If not, probably either I or Chris will claim the issue sometime in the next few weeks and close it out.

gnprice:
Well, manual testing found a server bug: https://github.com/zulip/zulip/issues/22164 The server will cheerfully mark the same messages unread again and again. In this version, that shows up as the unreads count in our inbox view / main screen getting bigger and bigger, because we get duplicates in our list of unread message IDs for the conversation.

The server should get fixed, but we should perhaps also work around it ourselves. Basically this expression: `oldUnreadMessages.concat(newUnreadMessages).sort()` should instead do a merge of those two lists -- they're both already sorted -- and whenever they have an element in common, it should take that only once. A bonus is that should be a linear-time algorithm, whereas the current version (though more concise) is $ n \log n $.

My testing was for stream messages, but I assume the same server bug applies to PMs. We should therefore handle it there too.

gnprice:
I discovered one further bug in this code while writing the group-PMs version just now. Added as another commit in this branch:

---

unread: Fix sort on marking 1:1 PMs unread

This was an `Array#sort` call without arguments, which falls into
a classic JS gotcha: it sorts by string value, even for numbers,
so 100 sorts before 2.

Fix the logic, and add a test.

There are similar-looking calls `….sort()` in the corresponding
code for stream messages in `unreadModel.js`.  But those do
correctly sort by numeric value, because the receivers of those
method calls are `Immutable.List` values, and those have the
good semantics.

Happily, the stream-messages code also already has a test
exercising the sort.  So if we had had a bug like this there,
we'd have caught it thanks to the test.  (In fact the way I found
this bug is that I was writing the comparable logic for group PMs,
and did write a test exercising the sort, and it failed.)

gnprice:
OK, I'm going to:
* merge this change;
* and then send a PR with my further changes mentioned above to support group PMs, and changes I wrote yesterday to work around that server issue zulip/zulip#22164 .

Thanks again @WesleyAC for your work on this!

PR #4792, opened 2021-06-11
deps: Run `yarn upgrade`.

chrisbobbe:
Notably, this includes an update to Sentry's SDK for React Native
that came out after the last `yarn upgrade` commit, in #4789, was
authored.

I got an error --

```
[!] CocoaPods could not find compatible versions for pod "Sentry":
  In snapshot (Podfile.lock):
    Sentry (= 6.1.4)

  In Podfile:
    RNSentry (from `../node_modules/@sentry/react-native`) was resolved to 2.5.0, which depends on
      Sentry (= 7.0.0)

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * changed the constraints of dependency `Sentry` inside your development pod `RNSentry`.
   You should run `pod update Sentry` to apply changes you've made.
```

-- so I ran `pod update Sentry`, and it went through. After that, I
ran `yarn` (with `pod install` getting run in the postinstall
script), and no further changes were made.

We should aim to have this out in a beta for a little while, to
shake out any bugs. In addition to this Sentry upgrade, quite a lot
of other stuff seems to have been touched.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -362,9 +362,9 @@ PODS:
     - React-Core
   - RNReanimated (1.13.3):
     - React-Core
-  - RNSentry (2.4.3):
+  - RNSentry (2.5.0):
     - React-Core
-    - Sentry (= 6.1.4)
+    - Sentry (= 7.0.0)
   - RNSound (0.11.0):
     - React
     - RNSound/Core (= 0.11.0)
@@ -372,9 +372,9 @@ PODS:
     - React
   - RNVectorIcons (7.0.0):
     - React
-  - Sentry (6.1.4):
-    - Sentry/Core (= 6.1.4)
-  - Sentry/Core (6.1.4)
+  - Sentry (7.0.0):
+    - Sentry/Core (= 7.0.0)
+  - Sentry/Core (7.0.0)
   - Toast (4.0.0)
   - UMAppLoader (1.2.0)
   - UMBarCodeScannerInterface (5.2.1)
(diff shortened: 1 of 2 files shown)


chrisbobbe:
If you take a look at the "rich diff" for yarn.lock, you'll see that all these new versions of things (there are a lot!!) were just released in the past couple days:

image

WesleyAC (approved):
LGTM :)

chrisbobbe:
> all these new versions of things (there are a lot!!) were just released in the past couple days:

Correction, I see some Sentry-related ones that were released 9 days ago; that would have predated the previous `yarn upgrade`. I'd guess that those versions just weren't compatible with the version ranges as they stood after the previous `yarn upgrade` and before this one (this one, as noted, upgraded @sentry/react-native).

chrisbobbe:
> LGTM :)

Thanks for the review!

gnprice:
Thanks @chrisbobbe for taking these updates!

It looks like a lot of new versions in sheer number-of-packages terms, but the bulk of them are either Babel or Sentry. So I think a more accurate picture may be: it's a handful of updates, two of which are to Babel and Sentry -- both of which are projects that have a practice of splitting up their code into lots and lots of packages.

E.g. try `du -sh node_modules/@babel/*/ | sort -rh`. There are 117 packages there… and the majority of them take up 24 kiB on disk. And even that sounds a lot bigger than these packages actually are: `du` counts each file as occupying at least one 4 kiB block. So you get to 24 kiB just by having a few hundred bytes of actual code plus a scattering of metadata:
```
$ tree node_modules/@babel/plugin-transform-unicode-regex/
node_modules/@babel/plugin-transform-unicode-regex/
├── lib
│   └── index.js
├── LICENSE
├── package.json
└── README.md

$ wc node_modules/@babel/plugin-transform-unicode-regex/lib/index.js 
 19  40 527 node_modules/@babel/plugin-transform-unicode-regex/lib/index.js
```

chrisbobbe:
> It looks like a lot of new versions in sheer number-of-packages terms, but the bulk of them are either Babel or Sentry.

Makes sense, thanks!

PR #4794, opened 2021-06-11
accountsReducer [nfc]: Followups on handling restart events

gnprice:
These are some small tweaks in the handy new `restart`-event implementation added in #4707.


chrisbobbe:
LGTM, thanks! Merged.

PR #4796, opened 2021-06-11
ComposeMenu [nfc]: Convert to functional component.

WesleyAC:
This was step one in a UX change I was planning to make. I got sidetracked due to problems with the library I was planning to use (callstack/react-native-paper#2769), but I figure I'd push this refactor commit anyways.

### DIFF
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -1,11 +1,11 @@
 /* @flow strict-local */
-import React, { PureComponent } from 'react';
+import React, { useCallback } from 'react';
 import { Platform, View } from 'react-native';
 import type { DocumentPickerResponse } from 'react-native-document-picker';
 import ImagePicker from 'react-native-image-picker';
 
-import type { Dispatch, Narrow } from '../types';
-import { connect } from '../react-redux';
+import type { Narrow } from '../types';
+import { useDispatch } from '../react-redux';
 import { showErrorAlert } from '../utils/info';
 import { BRAND_COLOR, createStyleSheet } from '../styles';
 import {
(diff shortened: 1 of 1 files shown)


src/compose/ComposeMenu.js
```diff
@@ -58,13 +57,18 @@ export const chooseUploadImageFilename = (uri: string, fileName: ?string): strin
...
-class ComposeMenu extends PureComponent<Props> {
-  uploadFile = (uri: string, fileName: ?string) => {
-    const { dispatch, destinationNarrow } = this.props;
-    dispatch(uploadFile(destinationNarrow, uri, chooseUploadImageFilename(uri, fileName)));
-  };
+export default function ComposeMenu(props: Props) {
+  const { destinationNarrow } = props;
+  const dispatch = useDispatch();
+
+  const uploadFileCallback = useCallback(
```
chrisbobbe, line 64: nit: This `useCallback` isn't doing much; `uploadFileCallback`s' callers are functions that get recreated each time the `ComposeMenu` renders (they aren't using `useCallback`), and those functions are being passed to various children.

chrisbobbe (commented):
Thanks! See one small comment below.

gnprice:
Thanks for this refactor! I agree with Chris's small comment, and everything else looks good.

Other than cutting that `useCallback` which isn't doing anything useful, the other direction to go would be to apply `useCallback` to all the named functions there, so that the callbacks passed as props to child components would be made stable. That's potentially an optimization. But we had a chat discussion a couple of weeks ago about when to use `useCallback`:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.2EuseCallback/near/1226466
and one conclusion was that that optimization may often not be effective at all. If it is effective here, it's unlikely to be meaningful -- it doesn't seem like these icon components (which are what ultimately get some of these callbacks as props) should have much work at all to do in their render functions.

PR #4797, opened 2021-06-11
MainTabsScreen: Turn off lazy loading.

WesleyAC:
This prevents the delay when loading a tab for the first time after app
open, which can often be significant.


chrisbobbe:
Great! I seem to remember some discussion of the rationale on CZO; would you please link to that in the commit message? I'll look at that as part of my next review.

WesleyAC:
Updated the commit message to have more detail and link to the CZO discussion.

chrisbobbe:
LGTM! Please merge at will (I'd do so myself but I'm doing another review at the moment and I don't want to forget this. 🙂)

chrisbobbe:
Merged, thanks!

PR #4798, opened 2021-06-12
Setup the 'realm_mandatory_topics' as 'isTopicMandatory' in Redux

thisisnitish:
Fixes first bullet point of #4378

Changes: 
- added `realm_mandatory_topics` to `InitialDataTypes`
- stored the relevant settings in `RealmState`
- added `isTopicMandatory` to initialState and set to false
- populating the `realmReducer`


WesleyAC:
This looks good to me, but I'm going to wait for @gnprice or @chrisbobbe to give a second opinion before merging :)

src/realm/realmReducer.js
```diff
@@ -24,6 +24,7 @@ const initialState = {
   twentyFourHourTime: false,
   canCreateStreams: true,
   isAdmin: false,
+  isTopicMandatory: false,
```
chrisbobbe, line 27: Let's add a `dropCache` migration in src/boot/store.js for this. Something like,

```js
  // Add `isTopicMandatory` to `state.realm`.
  '30': dropCache,
```

(or a higher number than 30, if new migrations happen to get added after I post this review). See `dropCache`'s jsdoc for what it's for.

chrisbobbe (changes requested):
Thanks @thisisnitish (and @WesleyAC for the review)!

See one comment below, about a migration.

Please format the commit message according to our [style guide](https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages). I think the change is probably simple enough that it can be described in just a summary line (with no body text), but the summary line should be formatted properly with a prefix, a period at the end, etc., as described in the doc.

src/realm/realmReducer.js
WesleyAC: Does this actually matter? My understanding is that an app update will drop the cache anyways, since it'll force a restart, and we aren't persisting this data in any case. I could be misunderstanding how long we intend for this data to persist, though.

chrisbobbe: Ah: we are persisting this data. `realmReducer` deals with `state.realm`, and 'realm' is in `cacheKeys`. Here's the jsdoc on `cacheKeys`:

```js
/**
 * Properties on the global state which we persist for caching's sake.
 *
 * These represent information for which the ground truth is on the
 * server, but which we persist locally so that we have it cached and
 * don't have to re-download it.
 */
```

chrisbobbe: That data won't get cleared automatically on an app restart, but it's easy and harmless for us to tell it to clear on a version upgrade, by using a `dropCache` migration.

WesleyAC: Hmm, but since this is getting overwritten on the `/register` response, doesn't that mean that it's going to get overwritten on app load anyways, since we're going to call `/register` then, before we display anything?

(To be clear, I think it's reasonable to have a `dropCache` in either case, I'm just trying to make sure I have a good understanding of the data flow)

chrisbobbe: > (To be clear, I think it's reasonable to have a `dropCache` in either case, I'm just trying to make sure I have a good understanding of the data flow)

Makes sense! 🙂

We do expect this to get clobbered with an up-to-date value when a `/register` completes. But there are two important possible states to consider while that `/register` is in progress:

- We have no cached server data to show, so we show `FullScreenLoading`.
- We have some stale data for this account, cached from a previous session, so we show an interactable UI with the "Connecting..." banner (`LoadingBanner`).

The conditional for what to show is provided by `useHaveServerDataGate`.

(The logic for what to show used to be more complicated and pretty hard to verify, depending on navigating imperatively at the right time, and setting the initial nav state correctly. If you want, you can read through #4454—particularly, a023b7540—for where we looked at that logic closely and refactored it by removing 'loading' as a navigation route; we then consolidated a reusable "have-server-data gate" in #4603 for use with components besides `MainTabsScreen`, and we fixed a bug in that "gate" logic in 98d19b829 just a few weeks ago.)

In that second state—

> - We have some stale data for this account, cached from a previous session, so we show an interactable UI with the "Connecting..." banner.

—it's possible, in theory (and I guess in practice, if the `/register` takes super long and we fix various UI perf issues) to go poking around the app and try to do things with the stale data that you'd also want to do with up-to-date data. Like, I guess, try to send a message without a topic. So it's good to do something explicit like this to ensure we have the right data from the beginning.

chrisbobbe: > So it's good to do something explicit like this to ensure we have the right data from the beginning.

Hmm, even more explicit than `dropCache` would be if the migration were written to add the key explicitly. I think `dropCache` would work, but only because of the "shallow-merge" quirk in `redux-persist` I found over at https://github.com/zulip/zulip-mobile/pull/4679#discussion_r626022077. (And that same quirk may have the effect that things would work even if we omitted a migration—but as I said there, it's nice to be as explicit as possible with migrations.)

I ~copied the `dropCache` idea from https://github.com/zulip/zulip-mobile/pull/4543#discussion_r627462602.

So, now I'm thinking something like

```js
  // Add `isTopicMandatory` to `state.realm`.
  '30': state => ({
    ...state,
    realm: {
      ...state.realm,
      isTopicMandatory: false,
    },
  }),
```

Sorry for the confusion!

src/api/initialDataTypes.js
```diff
@@ -120,6 +120,7 @@ export type RawInitialDataRealmUser = {|
   is_admin: boolean,
   realm_users: Array<{| ...User, avatar_url?: string | null |}>,
   realm_non_active_users: Array<{| ...User, avatar_url?: string | null |}>,
+  realm_mandatory_topics: boolean,
```
gnprice, line 123: We already have this on `InitialDataRealm`, so no need to add it elsewhere.

The right place for it is indeed `InitialDataRealm`. The organization of these different types is described at the place where they're all combined together:
```js
// Initial data snapshot sent in response to a `/register` request,
// after validation and transformation.
export type InitialData = {|
  // The server sends different subsets of the full available data,
  // depending on what event types the client subscribes to with the
  // `fetch_event_types` field of the `/register` request. We name these
  // subsets after the event types that cause them to be included.
```

This field is controlled by the `realm` event type being present; from the docs at https://zulip.com/api/register-queue :
> `realm_mandatory_topics`: boolean Present if `realm` is present in `fetch_event_types`.

so `InitialDataRealm` is the right place.

It'd belong on `InitialDataRealmUser` (and therefore here on `RawInitialDataRealmUser`, which that's based on) if it were instead controlled by `realm_user`.

(This organization isn't totally obvious, even though there is that one comment about it. I may go and add some short jsdoc on the individual types here.)

src/reduxTypes.js
```diff
@@ -256,6 +256,7 @@ export type RealmState = {|
   twentyFourHourTime: boolean,
   canCreateStreams: boolean,
   isAdmin: boolean,
+  isTopicMandatory: boolean,
```
gnprice, line 259: Let's keep this name very similar to the name in the API -- that helps simplify understanding what the code is doing, without the reader having to trace through all the layers.

The API's name for this is `realm_mandatory_topics`, and we're already on the "realm" state. So `mandatoryTopics` would be a good name.

gnprice, line 259: See the note in the jsdoc on this `RealmState` type about how the properties are organized: stuff about the server, then about the org/realm, then about the user.

This property is about the org aka the realm -- that's why its name in the API is prefixed `realm_` and it's controlled by the `realm` event type -- so it belongs in the "realm" section.

(I realize this is kind of easy to miss. I may try moving that jsdoc material inline into the type definition, next to the respective properties.)

src/realm/realmReducer.js
gnprice: > And that same ["shallow-merge"] quirk may have the effect that things would work even if we omitted a migration altogether

Yes, that's right. Effectively it means that when we introduce a new top-level property to any of our top-level state subtrees -- anything of the form `state.foo.bar` -- then without doing anything about it in a migration, we get the same effect as if we had a migration to give it the same value that we have for it in the initial state, as specified in the reducer.

A key bit of the background to how it works out that way is that the top-level subtrees `state.foo` are the units of what we store with `redux-persist`.

> I think `dropCache` would work, but only because of the "shallow-merge" quirk

The reason `dropCache` works is similar in spirit but I think it's more basic: it's that if a given top-level subtree is missing entirely, then at rehydrate time we just let the initial state (as specified in the reducer) stand untouched.

---

So in a situation like this, where we've introduced a new property `state.foo.bar` on one of the top-level state subtrees `state.foo`, the result is that any of:
1. no migration
2. a `dropCache` migration
3. a migration that explicitly adds the new property, like in Chris's last comment above

will produce the same, correct results.

It's good to be explicit, as with option 3. On the other hand, it means we duplicate the detail of what the default is -- new installs will get the default from the reducer's initial state, while old installs upgrading will get it from the migration. It also means more boilerplate, in a spot that isn't really type-checked (and is fundamentally tricky to properly type-check) and where getting it wrong could be pretty unpleasant.

So I'm kind of inclined to actually make our practice for this situation be option 1: no migration at all.

However! If we do that, I think it's important to still be explicit about making that choice. That helps us be in the habit of thinking about migrations when updating our types, so that we explicitly think through whether the update is one that falls into this situation and doesn't need a migration. So that could look like adding a comment here, the same place we'd add a migration, that looks like:
```js

  // Add `isTopicMandatory` to `state.realm`.
  // No migration; just use default from initial state.
```

---

Scanning back through previous migrations, I think the only one to date that we'd actually have skipped under this approach is 28, adding `state.settings.browser`. All the others either do something more complicated than adding a property, or add a property somewhere deeper in the tree (like in each element of `state.accounts`.)

As we add more things like miscellaneous settings, though, this could be something that becomes more common.

gnprice (commented):
Thanks @thisisnitish! Comments below, in addition to Chris's above.

thisisnitish:
Hey @chrisbobbe and @WesleyAC , I've added a summary, Please have a look and let me know what I can do more 🙂✌️.

chrisbobbe:
> Hey @chrisbobbe and @WesleyAC , I've added a summary, Please have a look and let me know what I can do more 🙂✌️.

I think you haven't force-pushed your new revision yet. 🙂 Also, I see that you've changed the title of this PR. Have you added more commits that make the new title an accurate description?

Also, please address Greg's review, too, if you haven't yet.

thisisnitish:
> > Hey @chrisbobbe and @WesleyAC , I've added a summary, Please have a look and let me know what I can do more 🙂✌️.
> 
> I think you haven't force-pushed your new revision yet. 🙂 Also, I see that you've changed the title of this PR. Have you added more commits that make the new title an accurate description?
> 
> Also, please address Greg's review, too, if you haven't yet.

Hey @chrisbobbe, I've forced push the commit. Yes, I've changed the title of the PR so that it looks more descriptive, and I've added the suitable commit message. Please have a look and Yes I've to address the Greg's review, I'll do it 🙂👍🏻.

chrisbobbe:
```
Fix: Setup the 'real_mandatory_topics' as 'isTopicMandatory' in Redux by adding relevant property
```

The prefix should be "the name of the subsystem and a colon", and it looks like that summary line is 97 characters long. Please take another look at the commit-message style guidelines, linked above. Also, if it's helpful, you can [read the history of the mobile project](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/git.md#reading-history) to see example commit messages.

Also, it looks like the PR is still only setting up some data plumbing. This is totally fine; breaking larger work into smaller chunks is generally a good strategy. 🙂 But the PR's new title, "Confirms that messages have topics, If there is "no topic" don't allow", because that user-facing behavior isn't implemented in the PR.

thisisnitish:
Sorry @chrisbobbe, @WesleyAC and @gnprice for being late, actually I was busy in some personal things, now getting back on track 😊.
Also, I will edit the commit message and PR title and I will also address the Greg's review 😊

thisisnitish:
Hey @chrisbobbe sorry for being late again, now I have edited the commit message again and also the PR title and I think the commit is also short. Please have a look and btw I'm addressing the greg review right now 🙂

thisisnitish:
Hey @chrisbobbe, I've read the @gnprice review and at the end as he mentioned to change property name from `isTopicMandatory` to `mandatoryTopics` , I've done it 😊. Please have a look 😊.

src/realm/realmReducer.js
gnprice: Made a chat thread to help make this discussion easier to find: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20migrations/near/1230029

chrisbobbe:
Thanks, @thisisnitish! I've made a few small tweaks to your commit, and also added some more commits that
- go ahead and use the settings value, now that you've gotten it to be stored in Redux (so I'm updating the PR description to say `Fixes: #4378`), and
- do some independent cleanups (one of these is Greg's, cherry-picked from #5000)

Please take a look! Also, @gnprice or @WesleyAC, this should be ready for review

src/reduxTypes.js
```diff
@@ -266,6 +266,7 @@ export type RealmState = $ReadOnly<{|
   twentyFourHourTime: boolean,
   canCreateStreams: boolean,
   isAdmin: boolean,
+  mandatoryTopics: boolean,
```
gnprice, line 269: nit: this goes in the previous group in this definition:
https://github.com/zulip/zulip-mobile/pull/4798#discussion_r652275167

src/config.js
```diff
@@ -20,6 +21,12 @@ const config: Config = {
   // better off giving up.
   requestLongTimeoutMs: 60 * 1000,
 
+  // The topic recognized by servers as "there is no topic". This should
+  // match
+  //   https://github.com/zulip/zulip/blob/5.0-dev/zerver/lib/actions.py#L2714
+  // or similar logic at the latest `main`.
+  noTopicTopic: '(no topic)',
```
gnprice, line 28: Centralizing a single definition for this constant makes a lot of sense!

I don't love "config" as a home for it. It's not really so much about a choice of how to configure the app, as it is about something we expect from the server.

Hmm, I guess in fact that suggests a good place for it would be under `src/api/` somewhere -- that's where we generally put facts about how the Zulip server behaves. Usually those are in the form of code to invoke, or type definitions, but a constant like this would fit too.

I think a file like `src/api/constants.js`, or `src/api/apiConstants.js`, would work well.

gnprice: Oh, and https://github.com/zulip/zulip/issues/3639 is the issue for the fact that this is hardcoded in the server (and therefore untranslated.)

gnprice (commented):
Thanks @chrisbobbe for the revised and extended version of this! Looks good, with small comments below.

src/api/constants.js
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * The topic servers understand to mean "there is no topic". This should
+ * match
+ *   https://github.com/zulip/zulip/blob/5.0-dev/zerver/lib/actions.py#L2714
+ * or similar logic at the latest `main`.
+ */
+// This is hardcoded in the server, and therefore untranslated; that's
+// zulip/zulip#3639.
+export const NO_TOPIC_TOPIC: '(no topic)' = '(no topic)';
```
gnprice, line 11: I think this is one where the type is best written as `string` rather than the specific string: as far as we want the type-checker to be concerned, this is just some string, much like how a timeout duration would be just some number. It's not a string that we want to have participate in determining our types, like the tag strings that we put in the `type` property of a `Message` or an `Action`.

src/reduxTypes.js
```diff
@@ -245,6 +245,8 @@ export type VideoChatProvider = $ReadOnly<{| name: 'jitsi_meet', jitsiServerUrl:
  * @prop videoChatProvider - The video chat provider configured by the
  *   server; null if none, or if the configured provider is one we don't
  *   support.
+ * @prop mandatoryTopics - Whether topics are required in stream messages
+ * (see https://zulip.com/help/require-topics)
```
gnprice, line 249: nit: indentation

src/realm/__tests__/realmReducer-test.js
```diff
@@ -7,29 +8,40 @@ import {
...
+        filters: action.data.realm_filters,
+        emoji: {}, // update as necessary if example data changes
+        videoChatProvider: null, // update as necessary if example data changes
+
+        email: action.data.email,
+        user_id: action.data.user_id,
+        twentyFourHourTime: action.data.twenty_four_hour_time,
+        canCreateStreams: action.data.can_create_streams,
+        isAdmin: action.data.is_admin,
+        mandatoryTopics: action.data.realm_mandatory_topics,
```
gnprice, line 30: nit: match ordering of properties (like elsewhere in branch)

src/api/constants.js
```diff
@@ -0,0 +1,11 @@
+/* @flow strict-local */
+
+/**
+ * The topic servers understand to mean "there is no topic". This should
+ * match
+ *   https://github.com/zulip/zulip/blob/5.0-dev/zerver/lib/actions.py#L2714
+ * or similar logic at the latest `main`.
+ */
```
gnprice, line 8: nit: one-line summary:

```suggestion
/**
 * The topic servers understand to mean "there is no topic".
 *
 * This should match
 *   https://github.com/zulip/zulip/blob/5.0-dev/zerver/lib/actions.py#L2714
 * or similar logic at the latest `main`.
 */
```

src/compose/ComposeBox.js
```diff
@@ -393,17 +401,26 @@ class ComposeBoxInner extends PureComponent<Props, State> {
...
+    if (
+      isTopicNarrow(destinationNarrow)
+      && topicOfNarrow(destinationNarrow) === apiConstants.NO_TOPIC_TOPIC
+      && mandatoryTopics
+    ) {
+      showErrorAlert(_('Message not sent'), _('Please specify a topic.'));
+      return;
+    }
+
     this.props.onSend(message, destinationNarrow);
```
gnprice, line 424: So, this code actually runs not only when you send a message you've just composed in the normal way, but also when you hit save/submit on a message you've just gone and edited. (Maybe that's a sign we should rename this method, and also the `onSend` callback -- more like `handleSubmit` / `onSubmit`.)

How does this behave in the editing case? Seems like it'll give a confusing error message, if nothing else. What's even the desired semantics in that case -- if a message is already to `(no topic)`, and then the setting is enabled, and then you go to edit it, can you edit it without changing its topic?

… I kind of don't want to let those questions block this PR, because that's a pretty rare case (editing is already uncommon, and editing with a target topic of `(no topic)` when that's forbidden is surely much more so), and because they seem like questions that may lead to a discussion of what the web app's behavior is and should be. Better to get the main issue fixed here, and leave that for another time.

But let's leave at least a TODO comment, like `// TODO how should this behave in the isEditing case?`. That way this at least doesn't much add to the confusingness of this code with respect to its reuse for editing vs. composing: the reader sees it's specifically about sending a message, but the comment warns that doesn't mean the function is only for the sending case.

gnprice (commented):
Thanks @chrisbobbe! A few comments below, all I think easy to resolve.

(Well, one points at a hairier issue, but I think a TODO comment is a fine solution for this PR.)

chrisbobbe:
Thanks for the review!

Issue #4671, opened 2021-04-17
Allow editing in all types of narrows


WesleyAC:
From `src/message/messageActionSheet.js`:

```javascript
  if (
    message.sender_id === ownUser.user_id
    // Our "edit message" UI only works in certain kinds of narrows.
    && (isStreamOrTopicNarrow(narrow) || isPmNarrow(narrow))
  ) {
    buttons.push(editMessage);
  }
```

It would be nice if editing worked in all types of narrows.

I think the easiest solution here is to just narrow to the relevant topic or PM narrow and then show the edit box, but we could also just fix the edit box so it works everywhere.

Gautam-Arora24:
Can I work on this?

AlaaElden98:
@Gautam-Arora24 I see that you working on #4675 if it's okay with you and you didn't do much work on this. I would like to work on it since I'm familiar with this part of code because I worked at #4643

Gautam-Arora24:
@AlaaElden98 I am not working in #4675. If you want you can surely work on this :)

AlaaElden98:
> @AlaaElden98 I am not working in #4675. If you want you can surely work on this :)

I meant #4674, my bad :))

alya:
Bad user experience resulting from this reported in #5179.

PR #4799, opened 2021-06-14
actionsheet: Allow editing in all narrows

shaurya2612:
Earlier, the messageActionSheet only allowed the 'Edit message' UI for
stream, topic and PM (both 1:1 and group) narrows. This change now
allows editing in all types of narrows.
Fixes: #4671

### DIFF
--- a/src/message/messageActionSheet.js
+++ b/src/message/messageActionSheet.js
@@ -17,7 +17,9 @@ import type {
 } from '../types';
 import {
   getNarrowForReply,
+  isHomeNarrow,
   isPmNarrow,
+  isSpecialNarrow,
   isStreamOrTopicNarrow,
   isTopicNarrow,
 } from '../utils/narrow';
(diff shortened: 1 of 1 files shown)


src/message/messageActionSheet.js
```diff
@@ -301,8 +303,10 @@ export const constructMessageActionButtons = ({
   }
   if (
     message.sender_id === ownUser.user_id
-    // Our "edit message" UI only works in certain kinds of narrows.
-    && (isStreamOrTopicNarrow(narrow) || isPmNarrow(narrow))
+    && (isStreamOrTopicNarrow(narrow)
+      || isPmNarrow(narrow)
+      || isSpecialNarrow(narrow)
+      || isHomeNarrow(narrow))
```
chrisbobbe, line 309: Have you tested that editing works in narrows where `isSpecialNarrow(narrow)` or `isHomeNarrow(narrow)` is true?

I wouldn't expect it to work for either of those, because I think we currently refuse to show the compose box in those cases. That's something I hope we can change soon, but it'll take more work toward refactoring `ComposeBox`.

chrisbobbe (commented):
Thanks, @shaurya2612! Comment below; unfortunately I don't have an estimate of when that issue will be resolved.

src/message/messageActionSheet.js
shaurya2612: Thanks @chrisbobbe for pointing that out.

I just saw the edit message option pop up and assumed it was done, my bad.

Maybe I could help out with the compose box (if necessary) after that PR is merged. Or are there any steps you could suggest to make this work for the meantime?

shaurya2612:
Thanks, @chrisbobbe!

Maybe I could help out with the compose box (if necessary) after that PR is merged. Or are there any steps you could suggest to make this work in the meantime?

gnprice:
Closing this PR. Making the edit-message UI actually work in those other narrows will take some significant UI work. When in the future someone tackles that, the step where we turn it on here in the action sheet will be straightforward.

Issue #4766, opened 2021-05-27
Followups on open-links-where setting, #4679


gnprice:
This is to track the post-merge review comments at https://github.com/zulip/zulip-mobile/pull/4679#pullrequestreview-667346423 :
> Comments below, including one on that thunk-action-creator thread and one on the icon.

PR #4721, opened 2021-05-05
flipper ios: Pin to some specific versions, to fix build with Xcode 12.5.

chrisbobbe:
Builds are failing with Xcode 12.5, and mine auto-updated this morning; here's a fix. (And a follow-up commit that resolves an unrelated issue.)

I've marked this P0 because iOS development/testing for any purpose (e.g., code review) is halted, for me and for anyone else whose Xcode got (auto-)updated; please let me know if this is wrong. I don't think Apple lets you downgrade Xcode after upgrading?

PR #4800, opened 2021-06-14
 changelog: Document changes since v27.163.

chrisbobbe:


docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+  authentication screen, on some app startups or when the user tapped a
+  logged-out account in the account picker. (#4604)
+* Fixed misconfigurations causing the Play Store to not offer the Android
+  app on devices that don't have cameras (#4722) or WiFi capability (PR
+  #4761).
+* Continued work toward enabling Android share-to-Zulip. (#4514)
+* Unclear, little-used "create group PM" option in the compose box removed.
+  (PR #4777)
+* Miscellaneous UI improvements. (#4734; PRs #4717, #4749, #4716, #4697,
+  #4686, #3517, #3395)
```
WesleyAC, line 64: #3517 and #3395 seem like bugfixes, not UI improvements, and thus probably deserve their own bullet point.

gnprice:
Thanks @chrisbobbe !

The most salient thing is that the "Highlights for users" needs to be a lot shorter. More precisely: the release notes we enter for the Play Store and App Store need to be a lot shorter than that; and this spot is where we do the work of drafting those up :slightly_smiling_face:

See previous entries for examples. Those have sometimes bumped up against the limits imposed by the Play Store and App Store; I think 500 characters is the limit on at least one of them.

Apart from the platform-imposed hard limits, people just aren't realistically going to read more than a small handful of items here. So the main thing I do in this section is to cut it down to just the few that users are most likely to actually feel are meaningful for them.

I'll also make some specific comments inline.

docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+  narrow when opening the app from the background. (#4758)
+* Fixed a glitch where the org URL input screen would flash by before the
+  authentication screen, on some app startups or when the user tapped a
+  logged-out account in the account picker. (#4604)
+* Fixed misconfigurations causing the Play Store to not offer the Android
+  app on devices that don't have cameras (#4722) or WiFi capability (PR
+  #4761).
+* Continued work toward enabling Android share-to-Zulip. (#4514)
+* Unclear, little-used "create group PM" option in the compose box removed.
+  (PR #4777)
```
gnprice, line 62: This doesn't go in this section -- most users probably never noticed that icon, and nobody will be excited it's gone. (Some will be pleased it's gone, from a UI-connoisseur perspective; but from the direct user perspective, it's the removal of an annoyance one hadn't even quite put a finger on.)

docs/changelog.md
gnprice: Yeah, a bullet point about fixing the keypad and letter emoji (those two issues) would likely be a good one.

Otherwise, a catch-all list like this doesn't go in this section. Short of going and looking up the issues/PRs by number, it doesn't add anything beyond the "other fixes and improvements" line at the end. And on the App Store or Play Store they don't get linkified, so by the time someone's going to the trouble of pulling up our GitHub repo and looking up an issue, if they just want to browse through changes then they have plenty of other ways to do so (including the GitHub release notes, which contain the developer highlights too.)

gnprice (commented):
Comments below on the "Highlights for users" section.

I'll take a look next at the developer-facing section but not quite as closely, because it's not user-facing and doesn't have tight length constraints.

docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+
+* Resolved issues (latest to earliest): PR #4794, PR #4707, PR #4789, PR
+  #4520, #4766, PR #4787, PR #4785, PR #4784, PR #4780, PR #4786, PR #4721,
+  PR #4779, PR #4778, PR #4334, PR #4727, PR #4574, PR #4737, #4715, PR
+  #4777, PR #4774, PR #4709, PR #4750, #4765, PR #4730, #4758, PR #4761, PR
+  #4762, PR #4731, PR #4717, PR #4700, PR #4710, PR #4760, PR #4749, PR
+  #4736, PR #4667, PR #4755, PR #4753, #4722, PR #4514, PR #4741, #4604, PR
+  #4728, PR #4739, PR #4732, PR #3955, #3540, #4006, #4323, #4734, #4264, PR
+  #4716, PR #4711, PR #4634, PR #4699, PR #4697, PR #4703, PR #4680, PR
+  #4544, PR #4698, PR #4686, PR #4690, PR #4678, PR #4689, #3517, #3395
```
gnprice, line 99: So the way I've been compiling this "Resolved issues" list is that each item in it really is an *issue*, i.e. either a bug or a feature request; when I list a PR, it's because we didn't file a GitHub issue for it.

So for example, taking from the top of this list:
* #4794 wouldn't be on that list
* #4707 would, as it implements a feature
* #4789 wouldn't
* #4520 wouldn't
* #4766 wouldn't, even though it's a GitHub issue -- it was really a code review, reified as a GitHub issue just to help with task-tracking
* #4787 wouldn't. Though there might be an item above saying we tried out Dependabot, and mentioning #4787.
* #4785, #4784, #4780, #4786 wouldn't -- just automated upgrades
* #4721 would, as it fixes a build issue. Oh but it'd go closer to the end of the list, because it was merged May 6.
* #4779, #4778, #4334, #4727, #4574, #4737 wouldn't -- Dependabot upgrades, and the change to enable it
* #4715 definitely would -- a buggy bit of UI
* (OK, that gives the idea.)

It wouldn't be unreasonable to have a list that includes all kinds of changes, even those that aren't issues. Possibly it'd be less work to compile that way -- though I think probably not, as it makes the list quite a bit longer.

Definitely quite reasonable to experiment with different ways of writing these notes.

If we do make this list broader, though, it should have a correspondingly different heading. "Resolved issues and merged PRs"?

docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+  `redux-persist`. (#4709)
+* Miscellaneous testing improvements. (PRs #4760, #4753)
+* Fixed a bug where `CaughtUp` state was being wrongly overwritten. (PR
+  #4698)
+* In dev mode only, a user-facing string passed to a `_: GetText` will now
+  be highlighted in the UI if it doesn't have an entry in
+  `messages_en.json`. (#4728)
+* Cleaned up some code for particular API endpoints. ( PR #3955)
+* Bumped `@zulip/shared` to v0.0.5. (PR #4680)
+* Added a migration that should have been released with v27.162. (PR #4689)
```
gnprice, line 88: This sounds kind of alarming. Could use another word or two to make clear that the impact of the missing migration was harmless.

(Alternatively, this item doesn't super need to be here at all -- it's not something that even developers especially need to know happened.)

docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+* Improved documentation for the release process. (PR #4690)
```
gnprice, line 89: So in this section there aren't any length constraints -- plus I'm not sure just who actually reads it anyway -- so having more text than necessary is harmless, and there's no need to spend a bunch of effort tightening it down.

But mainly in the interest of saving effort next time by simply not writing some items:
* I'd skip "Miscellaneous testing improvements" and "Cleaned up some code for particular API endpoints". They don't really tell you much without going and looking at the PRs; and there isn't enough information there to say why you might be interested enough to do so.
* I'd skip also "Bumped `@zulip/shared`". That doesn't in itself convey anything that's interesting for a developer to know.
* I'd skip the `yarn upgrade`, if not for the fact that it'd been nearly two years since the last one. So I might mention that fact :slightly_smiling_face: but then hopefully we'll never have a need to highlight future upgrades, because we'll be doing them more regularly.
* I'd skip the belated migration -- unless there'd been a serious issue related to it that developers should be aware of.

gnprice (commented):
And comments on the developer-facing section. Thanks again!

docs/changelog.md
```diff
@@ -37,6 +37,68 @@ It doesn't include
...
+* Added a setting to choose whether external links should open in the
+  embedded browser or the external browser. (PR #4679)
+* Added a server-compatibility notice for self-hosted orgs running Zulip
+  Server versions before 2.0.0. (PR #4750)
+* Stream settings screen now has a "subscribe" button. And, when
+  unsubscribed, useless toggle switches ("Pinned", "Muted", and
+  "Notifications") are hidden. (#4715)
+* Updated the primary app color to match the new logo. (PR #4544)
+* Fixed a bug on Android where the app sometimes opened to an unexpected
+  narrow when opening the app from the background. (#4758)
```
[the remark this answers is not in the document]
chrisbobbe: Cool, makes sense. I've cut the user highlights section down so it'll be <=500 characters after these transformations.

docs/changelog.md
chrisbobbe: Would the many muted-user PRs all get listed here? The umbrella issue, #4655, is quite near to being resolved but hasn't been yet.

chrisbobbe:
Thanks for the reviews! Revision pushed. Please see a question at https://github.com/zulip/zulip-mobile/pull/4800#discussion_r652797793; I've left all the muted-users PRs in the "Resolved issues" list for this revision but it'd be easy to do something else.

docs/changelog.md
gnprice: I think in that situation it'd be OK to say something like "most of #4655".

But also, that is perhaps a hint that it's time to close #4655 as resolved, and file new issues for specific known things that are missing :wink:

docs/changelog.md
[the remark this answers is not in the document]
gnprice: Ha, cutting it awfully close :stuck_out_tongue:

docs/changelog.md
chrisbobbe: > I think in that situation it'd be OK to say something like "most of #4655".

Done in my next revision 🙂

gnprice:
Thanks for the revision! Looks good modulo the bit about the muted-users PRs.

chrisbobbe:

gnprice:
Thanks! Looks good except for this stray heading at the top, new in this revision:
```diff
+## 27.163 (2021-06-04)
+
```

I'll go ahead and merge after removing that.

chrisbobbe:
> Thanks! Looks good except for this stray heading at the top, new in this revision:
> 
> ```diff
> +## 27.163 (2021-06-04)
> +
> ```
> 
> I'll go ahead and merge after removing that.

Oops, thanks; I added this temporarily so that `tools/changelog czo` would work, so I could easily get the issue/PR numbers linkified :)

PR #4801, opened 2021-06-14
RawLabel: Fix crash when text is empty string.

WesleyAC:
This was causing a crash when trying to autocomplete a user group with a
blank name. The empty string was passed into RawLabel, which caused this
invariant to incorrectly trigger.

Instead of using `!!` to check for undefined, we can do `!= null`, which
will trigger for undefined and null, but not an empty string.

If you want to trigger this on CZO, just type `@test` into the composebox :)

### DIFF
--- a/src/common/RawLabel.js
+++ b/src/common/RawLabel.js
@@ -30,7 +30,7 @@ export default class RawLabel extends PureComponent<Props> {
   render() {
     const { text, children, style, ...restProps } = this.props;
 
-    invariant(!!text !== !!children, 'pass either `text` or `children`');
+    invariant((text != null) !== (children != null), 'pass either `text` or `children`');
 
     // These attributes will be applied unless specifically overridden
     // with the `style` prop -- even if this `<RawLabel />` is nested


src/common/RawLabel.js
```diff
@@ -30,7 +30,7 @@ export default class RawLabel extends PureComponent<Props> {
   render() {
     const { text, children, style, ...restProps } = this.props;
 
-    invariant(!!text !== !!children, 'pass either `text` or `children`');
+    invariant((text != null) !== !!children, 'pass either `text` or `children`');
```
chrisbobbe, line 33: It occurs to me that `children` can be a number or a string; it's typed `?Node` ([link](https://github.com/facebook/react-native/blob/v0.63.4/Libraries/Text/TextProps.js#L60)) and I think the type Flow uses for `Node` includes strings and numbers ([doc](https://flow.org/en/docs/react/types/#toc-react-node)).

It'd be super weird, but I guess one could pass "foo" for `text` and `0` for `children`, and circumvent this check, ending up with "foo0" on the screen?

I'm trying to think whether that case is really pathological and you'd really have to be _trying_ to find it in order for it to be active. Or if you could stumble on it innocently.

I think it's not _totally_ impossible to stumble on it innocently: there's this gotcha in this [conditional rendering doc from React](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator):

> Note that returning a falsy expression will still cause the element after `&&` to be skipped but will return the falsy expression. In the example below, `<div>0</div>` will be returned by the render method.

If you're passing `children` _and_ you've accidentally stumbled on that gotcha there, and you're also passing a string for `text`, there could be a case like `foo0`.

chrisbobbe (commented):
Thanks for catching this!! One tiny correctness hole that I think I've found, below; possibly not worth addressing, but also wouldn't be too hard to address if you chose to.

src/common/RawLabel.js
gnprice: Yeah, seems cleanest to use the same `!= null` for `children` as well.

Thanks @WesleyAC for catching this!

WesleyAC:
Revision pushed :)

gnprice:
Looks good, thanks -- merged!

PR #4802, opened 2021-06-15
[nitpick] Fix parameter order for `onPress`.

AkashDhiman:
nitpick
It was listed as
```javascript
onPress: (topic: string, stream: string) => void,
```
it Should be 
```javascript
onPress: (stream: string, topic: string) => void,
```


WesleyAC (approved):
LGTM, thanks! I'm going to go ahead and merge this.

PR #4804, opened 2021-06-15
webview: Initiate 'ready' handshake from WebView.

WesleyAC:
This is a performance optimization which according to my measurements
reduces the average load time from 487ms to 429ms, on the following
device:

* Pixel 3a
* Android 11
* Chrome 91.0.4472.101
* Release mode

Where load time was determined by calling `performance.now()` in
`componentDidMount` and again in `handleMessage` when the first ready
event is received, and comparing the results. The numbers above are an
average from 25 different runs. The standard deviation was also reduced,
from 58ms to 35ms.

This should still be as safe as it was before, since we still require a
roundtrip before exchanging any data, we just start that roundtrip in a
different place.

Starting the handshake from the webview does subtly change the
properties of the system, but not in a way that should cause any
problems. Specifically, if it was the case that the React Native ->
webview communication came up after the webview -> React Native
communication, there could be dropped message sent after the Webview
sent a 'ready' message but before it was able to receive messages.
However, due to the implementation of the React Native -> webview
channel, we expect that it will always be up by the time the code that
sends the 'ready' message to React Native runs.

This does change the retry frequency, since my testing showed that in
most cases, the retrying logic was unnecessary. Whereas previously we
sent a ready event every 30ms (and typically lost the first 3 or 4 ready
events while the webview was starting up), we now send a ready event
every 100ms in the (hopefully uncommon) case where the first one is
lost.

See CZO for more details and histograms:

https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/webview.20setup.20code

### DIFF
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -718,11 +718,25 @@ const handleInboundEventTyping = (uevent: WebViewInboundEventTyping) => {
   }
 };
 
+let readyRetryInterval: IntervalID | void = undefined;
+
+const signalReadyForEvents = () => {
+  sendMessage({ type: 'ready' });
+
+  // Keep retrying sending the ready event, in case the first one is sent while
+  // the queue isn't ready. While this isn't something I've observed in testing,
+  // we've previously had bugs that were caused by this (for instance, #3078)
+  readyRetryInterval = setInterval(() => {
+    sendMessage({ type: 'ready' });
+  }, 100);
+};
+
 /**
- * Echo back the handshake message, confirming the channel is ready.
+ * Stop resending the handshake message once we confirm that the channel is
+ * ready.
  */
 const handleInboundEventReady = (uevent: WebViewInboundEventReady) => {
-  sendMessage({ type: 'ready' });
+  clearInterval(readyRetryInterval);
 };
 
 /**
@@ -1009,3 +1023,9 @@ documentBody.addEventListener('touchmove', (e: TouchEvent) => {
 documentBody.addEventListener('drag', (e: DragEvent) => {
   clearTimeout(longPressTimeout);
 });
+
+// It's possible we could call this earlier, and would see some performance
+// benifit from doing so, since js.js takes about 16ms to run on a Pixel 3a.
+// However, I don't see that as being worth the possible bugs from things
+// loading too early.
+signalReadyForEvents();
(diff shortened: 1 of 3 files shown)


src/webview/js/js.js
```diff
@@ -1009,3 +1012,16 @@ documentBody.addEventListener('touchmove', (e: TouchEvent) => {
...
+// However, I don't see that as being worth the possible bugs from things
+// loading too early.
+sendMessage({ type: 'ready' });
+
+// Keep retrying sending the ready event, in case the first one is sent while
+// the queue isn't ready. While this isn't something I've observed in testing,
+// we've previously had bugs that were caused by this (for instance, #3078)
+readyRetryInterval = setInterval(() => {
+  sendMessage({ type: 'ready' });
+}, 100);
```
gnprice, line 1027: I'd be a bit more comfortable seeing these grouped together in a function.

I'd also quite like to have this logic located next to the `clearInterval` and the declaration of `readRetryInterval`, so that all that logic is in one place.

How about sticking these in a function up there, and then invoking that function from here? Could be named like `signalReadyForEvents`.

gnprice (commented):
Thanks!

> This does change the retry frequency, since my testing showed that in
> most cases, the retrying logic was unnecessary. Whereas previously we
> sent a ready event every 30ms (and typically lost the first 3 or 4 ready
> events while the webview was starting up), we now send a ready event
> every 100ms in the (hopefully uncommon) case where the first one is
> lost.

This sounds like a fine reason to lengthen the retry interval, but I don't follow why it means the retrying logic was unnecessary. If previously the first few were typically lost, then the retrying was indispensable, right? (Either that, or pre-emptively waiting even longer.)

WesleyAC:
> This sounds like a fine reason to lengthen the retry interval, but I don't follow why it means the retrying logic was unnecessary. If previously the first few were typically lost, then the retrying was indispensable, right? (Either that, or pre-emptively waiting even longer.)

This is referring to the new retrying logic, but I can see how that's confusing. What I specifically mean is that in all the cases I saw, the first message that we sent from webview to RN got through. If that's the case in general (and not just because of quirks of the device I happened to test on, or whatever), that would mean that we don't need this retrying logic. But that's a confusing way of writing that, I'll update the message.

src/webview/js/js.js
WesleyAC: Done :)

gnprice:
Thanks for the revision! Merged.

PR #4805, opened 2021-06-16
unread [nfc]: Add `getUnreadCountForTopic`

gnprice:
This is a quick refactoring inspired by https://github.com/zulip/zulip-mobile/pull/4635#discussion_r652981999 .

There's a number of useful things we could do further in this direction, but stopping here for now in order to get this out.

### DIFF
--- a/src/unread/unreadModel.js
+++ b/src/unread/unreadModel.js
@@ -29,6 +29,9 @@ import {
 // Selectors.
 //
 
+/** The unread-messages state as a whole. */
+export const getUnread = (state: GlobalState): UnreadState => state.unread;
+
 export const getUnreadStreams = (state: GlobalState): UnreadStreamsState => state.unread.streams;
 
 export const getUnreadPms = (state: GlobalState): UnreadPmsState => state.unread.pms;
--- a/src/unread/unreadModel.js
+++ b/src/unread/unreadModel.js
@@ -28,6 +28,8 @@ import {
 //
 // Selectors.
 //
+// These take the global state as their input.
+//
 
 /** The unread-messages state as a whole. */
 export const getUnread = (state: GlobalState): UnreadState => state.unread;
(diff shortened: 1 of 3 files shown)


src/unread/unreadModel.js
```diff
@@ -40,6 +40,13 @@ export const getUnreadHuddles = (state: GlobalState): UnreadHuddlesState => stat
 
 export const getUnreadMentions = (state: GlobalState): UnreadMentionsState => state.unread.mentions;
 
+/** The total number of unreads in the given topic. */
+export const getUnreadCountForTopic = (
+  unread: UnreadState,
```
chrisbobbe, line 45: Is it fine for a selector to take a substate (like `UnreadState`) as the input state, or should all selectors start with `GlobalState`?

chrisbobbe (approved):
Thanks! This LGTM overall; see one small question below.

src/unread/unreadModel.js
gnprice: Yeah, perhaps calling this a "selector" is confusing and I should label it something else. Maybe "getter"?

The reason I want it taking just the unread state is so that we can conveniently use it in a context like #4635, in code where we have a `BackgroundData` style of object but don't have the whole Redux state.

More generally, a direction I'm interested in taking our data-model code is toward having each subsystem's model be more of a self-contained thing with its own identity and interface, vs. the status quo which kind of tries (but inconsistently so, because this doesn't really work if you try to push it hard) to treat them all as undifferentiated implementation details of the global state. For example, we'd have the whole "unreads" model as one such subsystem. So having a getter like this directly on that model's state is a small experiment in that direction.

chrisbobbe: I think the name "selector" is probably fine, and the use case makes sense. Thanks for explaining, and I'm excited to see our data-model code structured more sensibly. 🙂 

We'd call both `rootReducer` and `messagesReducer` "reducers", and I think it's fine to call this function (and those that take `GlobalState`) selectors.

gnprice:

chrisbobbe:
Great, thanks! Merged.

PR #4807, opened 2021-06-16
Show muted messages in mentioned narrow

WesleyAC:


WesleyAC:
This isn't exactly the right approach in the long term, I think (#3542 has the start of a better solution), but I think that this is a useful tactical fix anyways.

src/utils/__tests__/message-test.js
```diff
@@ -70,4 +70,22 @@ describe('shouldBeMuted', () => {
 
     expect(isMuted).toBe(true);
   });
+
+  test('mention narrow messages are never muted', () => {
+    const message = {
```
chrisbobbe, line 75: Hmm, these new objects aren't well-typed. Want to get the file type-checked and start using `exampleData` (e.g., `eg.streamMessage`)?

src/utils/message.js
```diff
@@ -20,6 +20,10 @@ export const shouldBeMuted = (
     return false; // never hide a message when narrowed to topic
   }
 
+  if (isMentionedNarrow(narrow)) {
+    return false;
+  }
```
chrisbobbe, line 25: I see the other early returns each have a comment explaining the reasoning. Could we add one that makes this one clearer (in particular, that this is just a tactical fix for now)?

chrisbobbe (commented):
Thanks, @WesleyAC! See a few small comments below.

src/utils/message.js
WesleyAC: Ah, yeah. I don't really like the other comments (they're explaining *what* the code's doing, not *why*), but this is a case where explaining why would be useful. I'll add that.

src/utils/__tests__/message-test.js
WesleyAC: I don't really *want* to, but I've gone ahead and done it. It is nice to have this typed :)

WesleyAC:
Updated :)

src/utils/__tests__/message-test.js
chrisbobbe: I can relate! 😅 Thanks for doing this. 🙂

src/utils/message.js
chrisbobbe: Yeah, exactly.

src/utils/__tests__/message-test.js
```diff
@@ -1,24 +1,29 @@
...
     expect(isMuted).toBe(false);
   });
 
   test('messages when narrowed to a topic are never muted', () => {
-    const message = {
-      display_recipient: 'stream',
-      subject: 'some topic',
-    };
-    const narrow = topicNarrow('some topic');
+    const narrow = topicNarrow('stream', 'some topic');
```
chrisbobbe, line 26: Ooh, nice to get this `topicNarrow` call fixed!

src/__tests__/lib/exampleData.js
```diff
@@ -250,12 +250,14 @@ export const stream: Stream = makeStream({
...
-  const { stream: streamInner = stream } = args;
+export const makeSubscription = (
+  args: {| stream?: Stream, in_home_view?: boolean |} = Object.freeze({}),
+): Subscription => {
+  const { stream: streamInner = stream, in_home_view } = args;
   return deepFreeze({
     ...streamInner,
     color: '#123456',
-    in_home_view: true,
+    in_home_view: in_home_view === undefined ? true : in_home_view,
```
chrisbobbe, line 260: nit: Could shorten to `in_home_view ?? true`.

This expression has the one difference that it also gives `true` when `in_home_view` is `null`. But `in_home_view` won't be `null` because of its type annotation as an [optional `boolean` property](https://flow.org/en/docs/types/objects/#toc-optional-object-type-properties) of the `args` object.

chrisbobbe (commented):
Thanks! See one nit below.

Also I wonder if @gnprice wants to check that this is a good tactical fix to make? I think it is, but I might be missing something.

src/__tests__/lib/exampleData.js
WesleyAC: Done! Thanks :)

src/utils/message.js
gnprice: The new comment is good!

I agree about the existing comments. The one on "when narrowed to topic" is especially unhelpful. I think a proper fix for that comment (and that conditional certainly does need some kind of comment) would involve writing some jsdoc to articulate what this function's semantics is supposed to be in the first place.

… And then where *that* leads, I think, is that this is really a helper function specialized for its one caller, `getShownMessagesForNarrow`, and should appear next to that function. (That is: it's hard to coherently write down what its meaning is supposed to be without effectively saying "this says whether `getShownMessagesForNarrow` should filter this message out".) And in turn looking at that call site… it should really get inlined there, because the top level of its control flow is mostly about switching on the type of narrow, and in many cases we can skip the `filter` entirely.

Anyway, none of that needs to happen in this PR. Perhaps after this is merged, I'll go and try cleaning that up and writing associated jsdoc.

src/utils/message.js
```diff
@@ -20,6 +20,15 @@ export const shouldBeMuted = (
     return false; // never hide a message when narrowed to topic
   }
 
+  // This logic isn't quite right - we want to make sure we never hide a
+  // message that has a mention, even if we aren't in the "Mentioned" narrow.
+  // However, it's more complex to do that, and this code fixes the largest
+  // problem we have with muted mentioned messages, which is that they show up
+  // in the count for the "Mentions" tab, but not in the actual narrow.
```
gnprice, line 27: One confusing thing in this comment: it describes the state before this bit of code was added, but at the time the comment is added, it's no longer accurate. Muted mentioned messages do now show up in the actual narrow under the "Mentions" tab.

I didn't notice that when first reading this comment -- it's subtle when reading now, when we're thinking about this change itself :wink: and so the previous state is salient in our minds. But in a future when someone (us included) is just trying to understand how the code is and doesn't remember this bugfix, it'll be confusing because the comment is saying one thing and then the code is ensuring the opposite.

About to merge, so I'll just fix this.

gnprice (commented):
Thanks @WesleyAC and @chrisbobbe ! One comment below, which I'll fix and then merge.

gnprice:
And merged! I also added a couple of followup commits -- the nice new well-typed, shorter version of the tests made it easier to see what was going on there, and surfaced some things to fix. I have some more I'll send as a new PR, too.

gnprice:
(I also just added a mention of #3472 in the description, so that it's possible to navigate to this PR from that issue.)

PR #4810, opened 2021-06-18
notif ios [nfc]: Factor out fromPushNotificationIOS.

gnprice:
This is a small followup refactor after #4163.

---

We have this somewhat fiddly logic twice; better to have it
(and explain it) just once.


chrisbobbe:
LGTM, thanks! Merged.

Issue #4745, opened 2021-05-11
Error toast when @-mentioning in a PM conversation


MMagueta:
### Bug description
When you try @-mentioning someone in a private message (either group or 1-1) you get an exception message saying "Couldn't load information about <NAME>" (Check the image for reference).

## For further reference, this issue was discussed in depth [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Quoting.20exception.20message) and a solution has been described

An excerpt of what was discussed at the zulip chat server:

> OK, probably actually the right thing is that `MentionWarnings` should take just a `Narrow`. After all, that "stream" value is just derived straight from `narrow` anyway, by the caller.

> Anyway:
> * We should definitely fix this logic to say `isPmNarrow` instead of `is1to1PmNarrow`. That should fix this bug. That would be a great small fix for someone to do.
> * Then there's a deeper refactor to be made here. That shouldn't block making the fix, but it would be a nice followup to it.

> Along with removing the `stream` prop, I think `MentionWarnings` could reasonably grow an invariant that the "reply narrow" or "destination narrow"—the answer to the question, "where should replies to this message go?"—is a stream narrow or a topic narrow. Its caller could conditionally not call it, otherwise, and it could do a check, e.g., in its constructor.
> > "reply narrow" or "destination narrow"
> - `ComposeBox` has an instance method `getDestinationNarrow` for this, which takes the data we'll give the server when we want to send a message. This looks like the thing to use here (or something with an interface closer to it), since we want the warning to appear before the message has been sent.
> - We've also got `getNarrowForReply` in src/utils/narrow.js, but that currently requires a `Message` or `Outbox` object.

> (Hmm, looks like that has a bug or two, but it should work for this use case without any fixes. In particular, it should probably explicitly return a stream narrow, a topic narrow, or a PM narrow, and nothing else—it's wrong if it ends up giving us a `HOME_NARROW`, `STARRED_NARROW`, etc.)

## Image

![Screenshots_2021-04-26-14-35-45](https://user-images.githubusercontent.com/14153897/117833159-7aa7ad80-b24c-11eb-902f-8894215ab1d6.png)

shaurya2612:
Hi, I am working on this issue. I changed the `is1to1PmNarrow` calls inside MentionWarnings and now I'm trying to refactor the part where stream should be removed as a prop from MentionWarnings.

[...]

PR #4812, opened 2021-06-19
compose: Fix quoting exception message in GroupPM

shaurya2612:
The `handleMentionSubscribedCheck` function in MentionWarnings used to
bail out correctly only when the narrow was a 1:1 PM thread. This showed
an exception message when using quotations in groupPMs. Changing
`is1to1PmNarrow` instances to `isPmNarrow` inside MentionWarnings fixes
this issue.

Fixes: #4745


WesleyAC:
Thanks for doing this! This looks good to me. I'm a bit surprised that this is correct behaviour (the webapp doesn't show a warning when @-mentioning someone who isn't in the group PM you're composing to), but I just tired it out, and it seems that's what the webapp does, so I think we should merge this.

cc @gnprice or @chrisbobbe to take a second look here.

shaurya2612:
@gnprice @chrisbobbe waiting for your perusal.

PR #4813, opened 2021-06-21
webview: Animate "Scroll to bottom" button show/hiding.

WesleyAC:


src/webview/static/base.css
```diff
@@ -247,6 +249,10 @@ body {
   height: 32px;
   fill: hsla(0, 0%, 100%, 0.75);
 }
+#scroll-bottom.hidden {
+  display: block;
```
chrisbobbe, line 253: Do I understand correctly, from a TODO added in #4259, that we'd like to refactor to remove the following rule before adding this one:

```
/* TODO: make specific versions of this rule for where we actually use it.
   That way, no surprises when some contrary rule has higher specificity.
   (Until then, we avoid using this class for any element where we have
   another rule specifying a `display:` attribute at all.) */
.hidden {
  display: none;
}
```

chrisbobbe (commented):
Looks good, thanks! I've just tried it on iOS. With current jank, I had trouble seeing the animation when the duration was at `0.25s`, so I temporarily increased it to `1.0s` and I could see it. I suspect that the right (longer-term) solution is to reduce the jank (or maybe it's much better in a release build), and merge this with `0.25s`, which is probably appropriate.

See one comment below.

src/webview/static/base.css
WesleyAC: I don't think we need to worry about that refactor, although I see how opinions could vary on that. What's happening in this PR is that I'm purposefully making a selector with a higher specificity than `.hidden`, since I want to change the behaviour. That makes us more likely to avoid surprises, since now this element has a specific selector, as we wanted.

From reading that TODO, you might think that means we need a `.scroll-bottom-hidden` class or something, but `.scroll-bottom.hidden` is actually more specific than `.scroll-bottom-hidden`, and it means that we can continue to use the same code to toggle the `hidden` class on this element, so I think the best resolution to this TODO would be making `.foo.hidden` rules for each of the classes we want to hide, and then removing this generic `.hidden` rule.

Does that make sense?

chrisbobbe: > From reading that TODO, you might think that means we need a `.scroll-bottom-hidden` class or something

This isn't quite what I had in mind.

> I think the best resolution to this TODO would be making `.foo.hidden` rules for each of the classes we want to hide, and then removing this generic `.hidden` rule.

This is more what I was thinking (with the current change being something that goes in this direction), but I read the "Until then…" part as preferring that we complete this work kind of in one sitting, like in one PR. That way, we wouldn't indefinitely have a situation (like we would if we just merged this) where someone could see the existing, general `.hidden` rule, not see the more specific `#scroll-bottom.hidden` rule, and then not understand why the `.hidden` rule isn't applying in some situations.

Ah, looking again, maybe that's not quite it: maybe the thing to care about most of all is to avoid having some arbitrary rule that doesn't involve `.hidden`, which _itself_ sets a `display: <something>`, if we plan to add the `hidden` class to any elements affected by that rule. This change would be OK by that.

chrisbobbe: I think we would plan to remove the `display: block;` line when the generic `.hidden` rule is removed, right? Should we add a quick comment for that?

WesleyAC: Good call, done :)

chrisbobbe:
Thanks! Merged.

src/webview/static/base.css
gnprice: > maybe the thing to care about most of all is to avoid having some arbitrary rule that doesn't involve `.hidden`, which _itself_ sets a `display: <something>`, if we plan to add the `hidden` class to any elements affected by that rule.

Yeah, that's the thing that would be super confusing, and that that general `.hidden` rule leaves us open to.

As long as a more-specific rule explicitly mentions `.hidden`, like this one does, the potential for confusion is reduced because once you look at this rule it's clear it's intended to override that one.

But: it's still potentially confusing to have `hidden` defined to mean one thing globally, and then switched to a different thing for one widget. Especially confusing because the comment on that global rule says we avoid doing a thing that we now in fact do :wink: 

So let's take this as a prompt to go and do that cleanup: replace this rule with per-widget rules, which can vary as needed. From a quick grep, I think the only places we ever use this class are here on `#scroll-bottom`; on `#spinner-older` and `#spinner-newer`; and on `#message-loading`. So this won't be a lot of work.

I'll file a quick issue.

gnprice: #4877

PR #4815, opened 2021-06-21
webview: Remove "Possible infinite loop" error.

WesleyAC:
This error wasn't something we expected to actually happen, but because
we set the threshold far too low, it's actually our most common error.
Given that we just changed the webview setup code in a way that we
expect to make this error more impossible than we previously thought, we
might as well just get rid of it and save ourselves the noise in Sentry.


chrisbobbe:
Thanks! Merged.

PR #4816, opened 2021-06-21
package.json [nfc]: Re-alphabetize dependencies.

WesleyAC:
The `yarn` tool alphabetizes dependencies, so it's nicest to keep them
that way in order to keep diffs clean.


chrisbobbe:
Yep, thanks for catching this—looking at the diff, my guess is I made a small mistake rebasing #4163 past a React Navigation upgrade. Merged!

Issue #4309, opened 2020-11-12
Opening a notification for nonexistent user/stream causes crash


gnprice:
This is a situation that's only possible in the first place when caused by some bug in either the app or the server. But such bugs happen, and particularly because it involves the server, we should be robust enough not to crash on it.

I don't think there's a known way to trigger this with a current version from master (or with today's release v27.157, now in alpha) and a stock server. But one way to trigger it is by reproducing #4290 on the current production release, v27.156. As described at https://github.com/zulip/zulip-mobile/issues/4290#issuecomment-726387004 , the app crashes, and (on Android) `adb logcat` shows output like this:

```
11-12 14:38:49.964 27558 27719 W ReactNativeJS: notification realm_uri not found in accounts
11-12 14:38:49.965 27558 27719 W ReactNativeJS:     realm_uri: "https://chat.zulip.org"
11-12 14:38:49.965 27558 27719 W ReactNativeJS:     known_urls: …
11-12 14:38:50.002 27558 27719 E ReactNativeJS: Error: getUserForEmail: missing user: email …
```
That is, we try to navigate to the right conversation as if it were on the active account... and when there isn't a user by that email, we crash.

Instead, we should handle this in `getNarrowFromNotificationData` -- that's the place where we take the raw names/IDs the server put in the notification and turn them into values we'll pass around to the ["soft center"](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/crunchy-shell.md) of our normal app code which assumes the data it gets is well-formed.

In `getNarrowFromNotificationData`, we should confirm that the specified user(s) or stream exist in our data. If they don't, then treat it much like we do in `getAccountFromNotificationData`:
* Log a warning (to Sentry.)
* Return null, to skip trying to do anything with the data.

That way the app will be open and running, and not crashed.

gnprice:
> I don't think there's a known way to trigger this with a current version from master (or with today's release v27.157, now in alpha) and a stock server.

A correction: this bug can be triggered on v27.157 by interacting with this remaining piece of #2295 (quoting my description at https://github.com/zulip/zulip-mobile/issues/2295#issuecomment-548612532 ):

> * If the user has several known accounts in the same realm, we
>   don't switch to either of them, because we can't tell which.

So if you've been using one realm/org; and you get a notification for a different one; and you have multiple known accounts in that other realm; then we don't switch to it, we stay in the wrong realm/org, and so we end up exercising this bug and crashing.

[...]

PR #4817, opened 2021-06-21
notifications: Match on realm domain, not href.

WesleyAC:
We had a steady stream of errors in Sentry caused by differences between
the `known_urls` and `realm_uri`, where one of those had an extra `/` on
the end (or some more surprising junk, such as a `/join/` link). Since
we don't support running Zulip on a non-root path [1], we can simply
compare the domain instead of the href. This is probably more of a
band-aid than a real fix (we should figure out how the strange values
get into `known_urls` and fix that), but we might as well just check the
domain, since the path doesn't matter.

Related: #4309.

[1] https://chat.zulip.org/#narrow/stream/3-backend/topic/support.20for.20running.20on.20non-webroot.20url.3F

### DIFF
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -48,7 +48,7 @@ export const getAccountFromNotificationData = (
 
   const urlMatches = [];
   identities.forEach((account, i) => {
-    if (account.realm.href === realmUrl.href) {
+    if (account.realm.origin === realmUrl.origin) {
       urlMatches.push(i);
     }
   });


src/notification/index.js
```diff
@@ -48,7 +48,7 @@ export const getAccountFromNotificationData = (
 
   const urlMatches = [];
   identities.forEach((account, i) => {
-    if (account.realm.href === realmUrl.href) {
+    if (account.realm.host === realmUrl.host) {
```
chrisbobbe, line 51: We usually compare `.origin`, not `.host` (e.g., as in e47a14308). `.origin` should have everything in `.host`, plus the scheme, like `http` or `https`, which I'd be inclined to differentiate on. What do you think?

chrisbobbe (commented):
Great, thanks! I haven't yet traced through how much of #4309 this will fix, but this seems like an obvious improvement to make. See one small comment below.

src/notification/index.js
WesleyAC: Ah, yeah, I was considering that when I made this, but didn't realize `.origin` was a thing. I'll change that, since in theory someone could run two different zulip servers, one on `http://example.com` and the other on `https://example.com`, even if that would be... extremely chaotic :p

WesleyAC:
Revision pushed :)

chrisbobbe:
Discussion is continuing at https://chat.zulip.org/#narrow/stream/176-mobile-core/topic/.23M4817.20notifications.3A.20Match.20on.20realm.20domain.2C.20not.20href.2E/near/1217189.

gnprice:

Issue #4818, opened 2021-06-21
Crash opening notif: `getAccountFromNotificationData` is undefined


gnprice:
Reported by @AkashDhiman [in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Notification.20Crash/near/1215247), with `adb logcat` output including a traceback.

Error message:

> TypeError: (0,c.getAccountFromNotificationData) is not a function. (In '(0,c.getAccountFromNotificationData)(t,(0,P.getIdentities)(o))', '(0,c.getAccountFromNotificationData)' is undefined)

@WesleyAC found we're seeing this [in Sentry](https://sentry.io/organizations/zulip/discover/zulip-mobile:bc9bf199beba406bb303e18e3d1f7c00), infrequently.

We've investigated (see chat thread), and have enough of a diagnosis to work around the issue. The bug appears to be in Metro, the JS bundler used by RN.

gnprice:
Indeed it looks like this was a bug in Metro v0.59.0 (which we're currently using), fixed in Metro v0.60.0 by facebook/metro#540.

We'll take that upgrade as part of the RN v0.64 upgrade, #4426. When we do, we should revert the workaround #4819.

PR #4819, opened 2021-06-22
import: Work around broken imports of '.'.

gnprice:
A bug somewhere in our stack -- presumably Metro or one of its
dependencies -- means that in src/notification/notificationActions.js
the import from `.` was getting resolved not to the module built from
its neighboring `index.js` file, as it should… but to the module
corresponding to the directory `src/common/`.

That also happens to be the directory where our only other imports
from `.` are found.  Seems like the bug may involve the resolution
of `.` being inappropriately shared globally.

In any case, the bug goes away when we make the import reference a
little less special-looking, as `./`.

Thread: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Notification.20Crash/near/1215247
Fixes: #4818

### DIFF
--- a/src/common/CountOverlay.js
+++ b/src/common/CountOverlay.js
@@ -3,7 +3,8 @@
 import React, { PureComponent } from 'react';
 import { View } from 'react-native';
 
-import { ComponentWithOverlay, UnreadCount } from '.';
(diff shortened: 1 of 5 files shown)


chrisbobbe:
LGTM, thanks! Merged.

PR #4820, opened 2021-06-22
queueMarkAsRead: Don't do anything when array is empty.

WesleyAC:
This should fix the following errors, which I see not infrequently in
the console and in Sentry:

```
{
  "httpStatus": 400,
  "json": {
    "code": "BAD_REQUEST",
    "msg": "Invalid message(s)",
    "result": "error"
  },
  "params": {
    "body": "messages=%5B%5D&flag=read&op=add",
    "method": "post"
  },
  "route": "messages/flags"
}
```

### DIFF
--- a/src/api/__tests__/queueMarkAsRead-test.js
+++ b/src/api/__tests__/queueMarkAsRead-test.js
@@ -56,4 +56,13 @@ describe('queueMarkAsRead', () => {
     lolex.runOnlyPendingTimers();
     expect(messagesFlags.default).toHaveBeenCalledTimes(2);
   });
+
+  test('empty array should not cause anything to happen', () => {
+    queueMarkAsRead(eg.selfAuth, []);
+
+    lolex.advanceTimersByTime(2500);
+
+    lolex.runOnlyPendingTimers();
+    expect(messagesFlags.default).toHaveBeenCalledTimes(0);
+  });
 });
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Thanks for this investigation!

Between both of `processQueue`'s callsites, it looks like the recursive callsite is the only one at which `unackedMessageIds` might be empty. Just above the other callsite, it looks like we're always sure to push at least one thing to `unackedMessageIds`—

```js
export default (auth: Auth, messageIds: number[]): void => {
  unackedMessageIds.push(...messageIds);
  processQueue(auth);
};
```

—see the only callsite of `api.queueMarkAsRead`—

```js
  if (unreadMessageIds.length > 0) {
    api.queueMarkAsRead(auth, unreadMessageIds);
  }
```

. I think that means that a more targeted fix might be to put the conditional just above the recursive callsite, skipping the recursive call when `unackedMessageIds` is empty. I've tried to think it through a bit, and I haven't yet found any strong practical advantages to doing it that way vs. the current way. Would you mind confirming that that way isn't a better alternative? 🙂

WesleyAC:
Thanks for checking that! My preference would be to keep the check where it is, and remove the check at the callsite, that way future callers don't need to worry about checking the length of the thing they pass in. Does that sound good?

chrisbobbe:
Yep, sounds good!

WesleyAC:
Revision pushed :)

chrisbobbe:
Thanks! Merged.

Issue #3549, opened 2019-07-09
Eagerly set `read` flags in Redux, before hearing from server


gnprice:
From [a chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/unread.20markers/near/760066) the other day.

Rishi reported:
> Getting a bug where the unread marker is clearing but the message is not being marked as read. Eg if I go back to the home view it's still there as unread, and I can click it again and get the animation again.
>
> This is on pretty bad internet.
>
> Though, it's still happening on somewhat reasonable internet. Another clue is that the "1 unread message" banner at the top is not clearing
>
> And sending messages here is working fine
>
> [...] I guess the animation is not connected with talking to the server?

As I said there: the animation is indeed not made to wait on talking to the server -- and I think that's probably important for having it reliably happen promptly. I believe the webapp has the same behavior, though I'm not as confident of that.

What's definitely a bug is that we have two conflicting ideas of whether each message is read, causing us to show the inconsistencies Rishi described. The Redux store controls the home view and the "1 unread message" banner, and the data inside the webview controls the unread marker. I believe the unread flag in the Redux store only gets cleared when we hear back from the server.

I think the right fix is that we should go ahead and eagerly clear that flag when the user reads a message locally, just like we clear the visual unread marker in the message list.

This is related to #3509; and @jainkuniya described a version of this in a comment there, too.

ishammahajan:
This would result in inconsistencies in data if the server rejects the request for some reason, or due to unideal network conditions. We would need to catch that as well.

As a parallel, it would be good to do a similar thing in `Reactions` and `Poll Votes` as well, since you would then see an immediate reflection of your action in the app. Checked several other commercial apps, they seem to do the same thing -- eagerly put the reaction / vote / something similar which depends on user interaction, and later remove it in case the server didn't register the action.

I tried to do this with the recent PR I sent, #3554, but I was unsuccessful, and recoiled from it, because the implementations I was coming up with were either too hard to understand, or involved too much of an architecture change (for example, using a flag to detect whether a request was sent or rejected, together, or another data structure like `Outbox` (which is for unsent Messages) to cross-check.

Huh, now that I think of it, the following system might work :-
1. Client sets a reaction eagerly once user performs an action.
2. Client sets a timeout with the aforementioned reaction, so that if there is a network failure, the reaction gets removed automatically.
3. Once client receives a response from the server, if `rejected`, does nothing, and if `success`, clears the timeout.

[...]

PR #4821, opened 2021-06-22
queueMarkAsRead: Reduce debouncing time to 500ms.

### LINKED ISSUE #374: Debounce posts to `messages/flags`
Otherwise we run into API usage limits

WesleyAC:
This debouncing doesn't buy us a lot (the webapp doesn't do anything
equivalent, from looking at the network tab as I mark messages as read),
and we have a strong reason not to make this take too long - our unread
messages banner is based on when we get the flag updates back from the
server, so the sooner we send them, the more responsive it will be.

We added this debouncing in 70295bd in order to fix #374, which wasn't
solving a particularly active problem - it was in order to avoid
possible API rate limits. The default rate limit for the server is 200
requests per minute [1], which works out to a request every 300 ms.
Given that, doing a request every 500ms should give us plenty of
breathing room to also request messages/etc, even in the unlikely case
that someone is pretty continuously scrolling through unread messages.

While we should make an effective local echo solution for marking
messages as read (#3549), this change is an good way to make things
better until we do that.

[1] https://chat.zulip.org/#narrow/stream/378-api-design/topic/Rate.20limits/near/1215716

### DIFF
--- a/src/api/queueMarkAsRead.js
+++ b/src/api/queueMarkAsRead.js
@@ -3,7 +3,7 @@ import type { Auth } from './transportTypes';
 import messagesFlags from './messages/messagesFlags';
 
 /** We batch up requests to avoid sending them twice in this much time. */
-const debouncePeriodMs = 2000;
+const debouncePeriodMs = 500;
 
 let unackedMessageIds = [];
 let lastSentTime = -Infinity;


chrisbobbe:
LGTM, thanks! Merged.

Issue #4631, opened 2021-04-09
Opening notif in one of several accounts on same realm doesn't switch accounts


gnprice:
When you open a notification that's in a different account from the one you were last using, we should switch to that account so that we can show you the right conversation.

Typically we do switch to that account. But if the account is one of several that you have on the same organization/realm, then we don't. That's because we can't tell which one it was for.

Since server version 2.1-dev-540-g447a517e6f (zulip/zulip@447a517e6f), the server sends in the notification the user ID of the relevant account. That's now been around a while, so most users' servers have that feature, and we can happily close this issue by solving it for the case where they do.

The main piece needed for this is described in `getAccountFromNotificationData` in src/notification/index.js:
```js
  if (urlMatches.length > 1) {
    // The user has several accounts in the notification's realm.  We should
    // be able to tell the right one using the notification's `user_id`...
    // except we don't store user IDs in `accounts`, only emails.  Until we
    // fix that, just ignore the information.
    logging.warn('notification realm_uri ambiguous; multiple matches found', {
      realm_uri,
      parsed_url: realmUrl,
      match_count: urlMatches.length,
    });
    // TODO get user_id into accounts data, and use that
    return null;
```

So we need to start storing the user ID in `state.accounts`.

The steps for implementing this will look something like:
* ~~First, complete #4951 with the steps described there.~~ done!
* Then we should be able to start using that information in `getAccountFromNotificationData`. Remember to handle gracefully (a) the case where the `userId` on some account/identity is still null because the user hasn't used that account since upgrading to the new version of the app that adds it, and (b) the case where the notification has no `user_id` because the server is old. 

(This is one of the pieces that remained from #2295.)

gnprice:
> * First, complete [Store user ID in accounts state #4951](https://github.com/zulip/zulip-mobile/issues/4951) with the steps described there.

This is now done!

So we can now proceed to this step:

> * Then we should be able to start using that information in `getAccountFromNotificationData`. Remember to handle gracefully (a) the case where the `userId` on some account/identity is still null because the user hasn't used that account since upgrading to the new version of the app that adds it, and (b) the case where the notification has no `user_id` because the server is old.

PR #4831, opened 2021-06-24
notifications: Add more detail to "realm_uri ambiguous" error.

WesleyAC:
There are a lot of users seeing this - more than I would expect to have
multiple accounts on the same server. I want to collect this data so
that we can have a more complete picture of what's going on in these
cases.

This is somewhat personal data (URLs of self-hosted realms), but we log
this same array elsewhere, so it's not changing the status quo
significantly to log it here.

### DIFF
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -77,6 +77,8 @@ export const getAccountFromNotificationData = (
       realm_uri,
       parsed_url: realmUrl,
       match_count: urlMatches.length,
+      unique_identities_count: new Set(urlMatches.map(matchIndex => identities[matchIndex].email))
+        .size,
     });
     // TODO get user_id into accounts data, and use that
     return null;


WesleyAC:
https://sentry.io/organizations/zulip/issues/1426821626/?project=191284&query=is%3Aunresolved&sort=freq&statsPeriod=14d

gnprice:
Yikes, yeah -- that is a surprisingly large fraction of our users. I'd expected that this issue would only be hit by Zulip developers, and others with a need to have test or demo accounts.

I think in the current state of things (pre-#4817) we only count exact URL matches as matches, so we already know that these cases involve several `state.accounts` entries with the exact same URL. Post-#4817, it'd make sense to include in the logging a count of how many of the matches are exact matches.

Probably the next thing to try adding would be a count of how many distinct emails are involved in the matching items. That doesn't raise any privacy concerns, and would help narrow down what's happening here: are people ending up with the same account listed several times, or several different accounts?

I don't have a great hypothesis for what the story might be in either of those cases (the first is probably some bug of ours; the second is probably some use case we're not thinking enough about?), but that'd advance us a step in figuring it out.

gnprice:
#4631 is the existing issue for the fact that this warning case exists at all. I'd considered it low-priority, because I'd figured that while it might be routine for us the developers of the Zulip mobile app, a very small fraction of other users would ever be in the situation where it applies.

But this observation that we're seeing that warning so frequently in Sentry suggests that we should maybe just do the work to fix that issue. Particularly if it turns out that it isn't caused by people somehow having duplicate entries in `state.accounts` (which would be a bug we should fix in itself), then we should probably assume the multi-account-same-org use case is more widespread than we'd thought, and just go and handle it properly.

src/notification/index.js
```diff
@@ -77,6 +78,10 @@ export const getAccountFromNotificationData = (
       realm_uri,
       parsed_url: realmUrl,
       match_count: urlMatches.length,
+      known_urls: knownUrls,
+      uniqueIdentities: identities
+        .map(identity => `${identity.realm.origin} ${identity.email}`)
+        .filter((value, index, self) => self.indexOf(value) === index).length,
```
gnprice, line 84: This should probably count only the ones that were in the list of matches, right? Otherwise it gets some noise that makes it harder to interpret -- we never know for sure that there wasn't a dupe, because it could just represent some unrelated accounts in other orgs.

You can use `urlMatches` to reuse that computation without duplicating it -- it's a list of indexes.

src/notification/index.js
```diff
@@ -77,6 +78,10 @@ export const getAccountFromNotificationData = (
       realm_uri,
       parsed_url: realmUrl,
       match_count: urlMatches.length,
+      known_urls: knownUrls,
+      uniqueIdentities: identities
```
gnprice, line 82: nits:
* use snake_case to match the other keys -- at least the ones in this same warning-event blob :slightly_smiling_face: 

* name should say "count", like `match_count` does, to indicate it's the number of them and not a list of them

gnprice (commented):
Thanks! Comments below.

WesleyAC:
Thanks! Revision pushed.

src/notification/index.js
```diff
@@ -77,6 +78,10 @@ export const getAccountFromNotificationData = (
       realm_uri,
       parsed_url: realmUrl,
       match_count: urlMatches.length,
+      known_urls: knownUrls,
```
gnprice, line 81: Does this really give us useful information for debugging this issue? This case is already all about when the URLs do match, so I'm not sure it does.

There is this that I mentioned above:

> I think in the current state of things (pre-#4817) we only count exact URL matches as matches, so we already know that these cases involve several `state.accounts` entries with the exact same URL. Post-#4817, it'd make sense to include in the logging a count of how many of the matches are exact matches.

But beyond that count, I don't think the actual array adds anything.

It's true that we already log this array in the no-matches case, a bit above this code. (And perhaps we should stop doing that.) Still, if we can avoid even having to give that explanation, that'd be better :slightly_smiling_face:

src/notification/index.js
```diff
@@ -77,6 +78,10 @@ export const getAccountFromNotificationData = (
...
+      unique_identities_count: urlMatches
+        .map(matchIndex => `${identities[matchIndex].realm.origin} ${identities[matchIndex].email}`)
```
gnprice, line 83: Now that we're using `urlMatches` here, I believe the `origin` part of this is redundant. If so, better to remove it so the reader doesn't wonder if they're missing something.

src/notification/index.js
```diff
@@ -77,6 +78,10 @@ export const getAccountFromNotificationData = (
...
+        .filter((value, index, self) => self.indexOf(value) === index).length,
```
gnprice, line 84: This line is clever! But also quadratic. In this context the list is unlikely to be long enough that that matters; but nevertheless writing something linear-time, like `new Set(…).size`, would make me feel more at ease.

gnprice (commented):
Thanks! Comments below -- one significant, and a couple of smaller ones.

WesleyAC:
Revision pushed :)

gnprice:
Thanks, looks good! Merging.

There was a paragraph in the commit message that reflected an earlier revision:

    This is somewhat personal data (URLs of self-hosted realms), but we log
    this same array elsewhere, so it's not changing the status quo
    significantly to log it here. I've purposefully avoided logging email
    addresses, instead logging the total number of unique identities. If we
    need more information to debug this further, we can add that later.

I updated that to reflect what we ended up with, and say a bit about what the data we're actually adding here is for:

    These counts will tell us enough to determine whether people are ending
    up somehow with duplicate accounts in our state, or on the other hand
    have several accounts on one realm that actually seem to be distinct.

PR #4834, opened 2021-06-29
generateInboundEvents: Remove early return on sending a 'content' event.

chrisbobbe:
I've played around with this and inspected the code, and I haven't found any obvious issues yet.

Related: #3851

### DIFF
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -106,20 +106,30 @@ export type FetchingState = {|
  *
  * Unlike almost all other subtrees of our state, this one can be
  * incomplete, always in exactly the same way that `MessagesState` is.
+ *
+ * We expect most of these flags to be very sparse: the number of messages
+ * you have starred, or where you're @-mentioned, is typically a very small
+ * fraction of the total number of messages you have. That's why it probably
+ * doesn't make sense for this data structure to explicitly indicate "not
+ * starred", "not @-mentioned", etc., for every known message. That could
+ * significantly increase the memory and storage requirement when we know
+ * about a lot of messages. If we need to distinguish "unknown message" from
+ * "message without flag", we can look up the message ID in
+ * `state.messages`.
  */
 export type FlagsState = {|
-  read: {| [messageId: number]: boolean |},
-  starred: {| [messageId: number]: boolean |},
-  collapsed: {| [messageId: number]: boolean |},
-  mentioned: {| [messageId: number]: boolean |},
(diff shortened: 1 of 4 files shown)


src/reduxTypes.js
```diff
@@ -106,20 +106,24 @@ export type FetchingState = {|
  *
  * Unlike almost all other subtrees of our state, this one can be
  * incomplete, always in exactly the same way that `MessagesState` is.
+ *
+ * TODO: We currently represent "unknown message" and "message without flag"
+ * the same way, by just omitting the message ID. We may want to
+ * differentiate these.
```
gnprice, line 112: Hmm, interesting.

I think we probably don't want to do that here, because the thing is that most of these flags are very sparse: the number of messages you have starred, or where you're @-mentioned, is typically a very small fraction of the total number of messages you have. So explicitly indicating "not starred", "not @-mentioned", etc., for all those messages could be a significant increase in memory and storage use when you have a lot of messages in the state.

(That also suggests that we should be storing the *unread*, rather than the read, messages here. But even as is, we're getting the benefit of this optimization for all the other flags. And wow, there are a lot of flags there -- I have no idea what some of these mean.)

I think probably the right way to distinguish "unknown message" from "message without flag" is going to be to look up the message ID in `state.messages`.

src/webview/generateInboundEvents.js
```diff
@@ -99,19 +99,23 @@ export default function generateInboundEvents(
...
     || !equalFlagsExcludingRead(prevProps.backgroundData.flags, nextProps.backgroundData.flags)
   ) {
-    return [updateContent(prevProps, nextProps)];
+    uevents.push(updateContent(prevProps, nextProps));
   }
 
-  const uevents = [];
-
   if (prevProps.backgroundData.flags.read !== nextProps.backgroundData.flags.read) {
+    // TODO: Don't consider messages outside the narrow we're viewing.
```
gnprice, line 115: Yeah, wow, that would be quite an optimization in some cases.

gnprice (commented):
Thanks @chrisbobbe ! All looks good except the small comments below, which are all in comments and commit messages.

I don't feel 100% confident there isn't some way we're subtly depending on the current quirkier logic; but I don't see any, and this is definitely cleaner. So given your manual testing, I'm happy merging the change and fixing forward if we do run into some issue. Please go ahead and merge after the tweaks below.

chrisbobbe:
> Please go ahead and merge after the tweaks below.

Issue #4837, opened 2021-06-30
Replace `connect` with `useSelector` in all our component definitions.


chrisbobbe:
We've found another reason to do this: it'll help simplify the React Native v0.64 upgrade (#4426). We'll want to get closer to enabling a Flow feature called ["Types-First"](https://flow.org/en/docs/lang/types-first/), and add a few targeted suppressions where that turns out to be substantial work. See https://github.com/zulip/zulip-mobile/issues/4426#issuecomment-767044329.

Having tried Types-First experimentally, I get some 60+ complaints that our exported component definitions that are made with `connect` aren't sufficiently annotated. The codemod Flow provides seems to stumble and not be able to fix these automatically. While we could do them by hand, it's probably cleanest and easiest just to get rid of the `connect` calls.

We've already gotten a good start on converting our components to function components. We should continue, with particular interest in the ones that use `connect`. There are 36 as of writing this.

gnprice:
Sounds good! There's one place where we know we need `connect` instead of `useSelector` because of the subtle differences in how they actually work; but I guess we've already tucked that inside `withHaveServerDataGate`, so its result isn't directly exported and doesn't affect this.

I've just merged #4838 and #4844, and it looks like `git grep -E '\bconnect\('` is now down to 26 matches, from 36 (one of which is that `connect` in `withHaveServerDataGate`; the rest are indeed exported.)

There's also a number of places where we do the same thing except we provide type arguments to `connect`:
```
$ git grep -E '\bconnect<' src/ 
src/account-info/AccountDetails.js:export default connect<SelectorProps, _, _>((state, props) => ({
…

$ git grep -El '\bconnect<' src/  | wc -l
26
```
One of those is the type-wrapper `src/react-redux.js` itself; the other 25 are components, all in `export default` definitions except some that are exported after another step or two. Those include `MessageList` and `ComposeBox`, which may be among the most complex to convert to function components.

Do those cases also cause this sort of trouble, or do the type parameters make it OK?

In any case, if you run out of easy conversions and there are still a few left, we can always do annotations for those, even if the codemod is no help. For `ComposeBox` in particular, I expect it will be less work to write an annotation than to convert it to a function component. For one thing, the stakes are lower, because a mistake in the annotation isn't a live bug.

PR #4838, opened 2021-06-30
AccountPickScreen [nfc]: Use `useSelector` instead of `connect`.

chrisbobbe:
An instance of #4837.

(The use of `navigation.canGoBack()` in the last commit in the branch is an independent, unrelated cleanup.)

### DIFF
--- a/src/account/AccountPickScreen.js
+++ b/src/account/AccountPickScreen.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
+import React, { useContext, useCallback } from 'react';
 import { Alert } from 'react-native';
 
 import * as api from '../api';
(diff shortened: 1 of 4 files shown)


src/account/AccountPickScreen.js
```diff
@@ -26,102 +24,81 @@ import { showErrorAlert } from '../utils/info';
...
-        </Centerer>
-      </Screen>
-    );
-  }
+  return (
+    <Screen
+      title="Pick account"
+      centerContent
+      padding
+      canGoBack={navigation.canGoBack()}
```
gnprice, line 84: Interesting! Good find. This does seem like an improvement.

gnprice (commented):
Thanks! Looks good -- merged.

chrisbobbe:
Thanks for the review!

PR #4839, opened 2021-06-30
lightbox [nfc]: Remove meaningless properties in rnfb config.

AkashDhiman:
Done in `downloadFileToCache`, the object passed to
`RNFetchBlob.config` contained some meaningless properties.

These property are technically the sub properties of the property
`addAndroidDownloads`. But we would want to remove them instead of
adding them there since these essentially configure the android
download manager to download the file, which we do not want in case
of this function.

Normally rnfb will use its internal implementation for downloading
based on the `OkHttp` android library.
Originally reported [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Sharing.20from.20lightbox.20broken.20on.20Android/near/1223148)

### DIFF
--- a/src/lightbox/download.js
+++ b/src/lightbox/download.js
@@ -73,7 +73,4 @@ export const downloadFileToCache = async (tempUrl: string, fileName: string): Pr
   RNFetchBlob.config({
     path: `${RNFetchBlob.fs.dirs.CacheDir}/${fileName}`,
     fileCache: true,
-    useDownloadManager: true,
-    mime: getMimeTypeFromFileExtension(fileName.split('.').pop()),
-    title: fileName,
   }).fetch('GET', tempUrl);


src/lightbox/download.js
```diff
@@ -73,7 +73,4 @@ export const downloadFileToCache = async (tempUrl: string, fileName: string): Pr
   RNFetchBlob.config({
     path: `${RNFetchBlob.fs.dirs.CacheDir}/${fileName}`,
     fileCache: true,
-    useDownloadManager: true,
```
chrisbobbe, line 76:
```
These property are technically the sub properties of the property
`addAndroidDownloads`.
```

A quick link to the documentation would help someone in the future verify this. 🙂

```
Normally rnfb will use its internal implementation for downloading
based on the `OkHttp` android library.
```

Could you say more in the commit message, ideally with a link to documentation, about why this is relevant to the change?

chrisbobbe: Also, I think a link to my report on CZO might be helpful in the commit message. It has some additional context about how the error was discovered; specifically, that I was working on `exact_by_default`. (That context isn't necessary to justify fixing the error, so I don't think it needs to be explained in the commit message. But a link to the discussion might help satisfy people's curiosity about how it was discovered.)

chrisbobbe (commented):
Thanks! Just a few small comments on the commit message; otherwise, this looks good to me.

src/lightbox/download.js
AkashDhiman: done.

gnprice:
Thanks @AkashDhiman, and @chrisbobbe for the review! Looks good -- merging.

PR #4844, opened 2021-07-01
Use `useSelector` instead of `connect` in several places.

chrisbobbe:
Several instances of #4837 that are pretty straightforward.

There are more of these easy ones, and I almost did those here, but then I figured this was already a big enough chunk to review. 🙂

### DIFF
--- a/src/chat/MarkAsReadButton.js
+++ b/src/chat/MarkAsReadButton.js
@@ -2,9 +2,9 @@
 
 import React, { useCallback } from 'react';
 
-import type { Auth, Narrow, Stream, Dispatch } from '../types';
+import type { Narrow } from '../types';
 import { createStyleSheet } from '../styles';
-import { connect } from '../react-redux';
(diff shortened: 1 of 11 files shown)


src/chat/MarkAsReadButton.js
```diff
@@ -0,0 +1,74 @@
...
+    paddingLeft: 12,
+    paddingRight: 12,
+  },
+});
+
+type Props = $ReadOnly<{|
+  narrow: Narrow,
+|}>;
+
+export default function MarkAsReadButton(props: Props) {
```
gnprice, line 32: MarkAsReadButton [nfc]: Rename, from MarkUnreadButton.
    
    The button marks things as read, not unread.

:upside_down_face: :laughing: :+1:

gnprice (commented):
Looks good, thanks, merged!

chrisbobbe:
Thanks for the review!

PR #4845, opened 2021-07-02
logging: Use Sentry's captureMessage and captureException in logToSentry.

chrisbobbe:
captureException takes an Error object, and captureMessage takes a
string. It makes sense to use captureException when we have an
Error, then, and captureMessage when we have a string -- instead of
coercing both into a string and synthesizing a stack trace to
attach.

Greg said, about the comment "the static API's `captureException`
doesn't allow passing strings",

> [...] I think the idea of that comment is that we want to pass an
> `Error` when we have an `Error`, and just a string when we have
> just a string.
>
> And then implicitly it's looking for one method to use in both
> cases. Neither of those works, so they were both rejected.
>
> But the conditional seems like a fine solution to that problem.

### DIFF
--- a/flow-typed/@sentry/react-native_v2.x.x.js
+++ b/flow-typed/@sentry/react-native_v2.x.x.js
@@ -117,6 +117,11 @@ declare module '@sentry/react-native' {
   // More methods are available.
   declare export class Scope {
     /**
+     * Sets the level on the scope for future events.
+     * @param level string {@link Severity}
+     */
+    setLevel(level: SeverityType): this;
+    /**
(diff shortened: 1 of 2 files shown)


WesleyAC (approved):
LGTM. I see [on CZO](https://chat.zulip.org/#narrow/stream/176-mobile-core/topic/sentry/near/1223956) you mentioned testing this — I haven't tried to verify this, but it looks good.

gnprice:
Thanks! Looks good -- merged.

chrisbobbe:
Thanks for the reviews!

PR #4846, opened 2021-07-02
android: styles.xml change for dark theme, on the path to RN v0.64 upgrade.

chrisbobbe:
We need to test on Android 10 and record any changes; the linked doc says the dark theme is available in Android 10 (API level 29) and higher. I've tested on the office Android device, a Samsung Galaxy S9, but it's running Android 9 and I'm not sure it's time to upgrade it yet (I'm not sure if we can go back to Android 9 after doing so).

I found no visible changes in the app on Android 9 when switching the system-level dark theme setting on and off. I'm not sure how I was able to switch that setting on and off when the doc says that the Dark theme is available starting in Android 10. Anyway, best to test on Android 10.

@AkashDhiman, would you be interested in testing on a device running Android 10?

Related: #4426 
Related: #4009

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -1,7 +1,7 @@
 <resources>
 
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
+    <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
         <!-- Customize your theme here. -->
         <item name="colorPrimary">@color/primaryColor</item>
         <item name="colorPrimaryDark">@color/primaryColorDark</item>


AkashDhiman:
Sure, I will have a look.

AkashDhiman:
checked by using `../zulip/tools/reset-to-pull-request 4846` and then launching the app, the usual way.

The following observation is made on Android 10 and 11:
- white screen that comes up on starting the app changes to darker color when system UI is set to dark mode, this does not happen in the current master build.

There are **no** visual changes within the app.

For reference here are some screen captures on an Android 10 emulator:

## when the system UI is set to light mode
on-sys-light-mode

## when the system UI is set to dark mode
on-sys-dark-mode

WesleyAC (approved):
LGTM :+1:

chrisbobbe:
Thanks for testing it out, @AkashDhiman, and thanks for the review, @WesleyAC! Merged.

gnprice:
Thanks all!

> I'm not sure how I was able to switch that setting on and off when the doc says that the Dark theme is available starting in Android 10.

FWIW my guess for this is that Samsung added a version of this feature for their own devices before it was in Android itself.

I'd expect their version to apply to their own apps, and possibly some third-party apps that explicitly use some Samsung-specific API to get the theme. It probably doesn't succeed in setting the theme in a way that apps relying on the Android 10 API will notice, because they probably designed it before that API existed. If you didn't see an effect like the one Akash showed above, then that fits with that expectation.

Issue #4811, opened 2021-06-19
Send "not typing anymore" notifications more


gnprice:
When the user is typing a PM, we send typing-status notifications.

The semantics is that these expire naturally: if the user keeps typing, we keep sending them periodically, and if a client on the other end sees a timeout interval (15 sec, I think) go by without getting another, then it considers the user no longer typing.

But there are also situations when the user has actively stopped typing. For those, the API supports a "not typing anymore" notification, which causes the receiving client to cancel the "Greg is typing…" display immediately.

Currently we handle one of these situations: when the user sends their message. This works by invoking the `sendTypingStop` thunk action.

We should do the same thing in several more situations:
- [ ] when the user blurs the compose box, i.e. when it loses focus because the user has interacted with something else in the UI;
- [ ] when the user navigates away to some other screen -- either by navigating back, or by navigating forward to somewhere else;
- [ ] when the user clears the whole contents of the compose box.

ConaGo:
Hi, this seems like a good first issue and i would like to try and implement the requested changes :)

ConaGo:
@zulipbot claim

ConaGo:
Hi @gnprice, i implemented the change and would love to get some feedback :)

ConaGo:
Hey, any updates regarding this?

PR #4856, opened 2021-07-05
a-compose-send: added more "not typing anymore" notifications

ConaGo:
task 1 and 2: the handleMessageBlur gets called when the user navigates away and when the compose box looses focus, sending the notification here fixes both issue-tasks.
task 3: by checking the message length in handleMessageChange the notification gets sent on emptying the input.
Fixes: #4811


gnprice:
Hello @ConaGo! Welcome, and apologies for the delayed review.

This change looks great. Merging it now, with just some tweaks to the commit message.

PR #4864, opened 2021-07-06
Rename "Search" to "New PM" in PmConversationScreen

WesleyAC:
[See CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/slow.20disappear.20from.20unreads)

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -76,6 +76,7 @@ export const IconRight = makeIcon(Feather, 'chevron-right');
 export const IconPlusCircle = makeIcon(Feather, 'plus-circle');
 export const IconLeft = makeIcon(Feather, 'chevron-left');
 export const IconPeople = makeIcon(Feather, 'users');
+export const IconPerson = makeIcon(Feather, 'user');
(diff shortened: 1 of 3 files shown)


chrisbobbe:
> [See CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/slow.20disappear.20from.20unreads)

I think this meant to link to the same discussion that the commits link to: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/renaming.20.22Search.22.20in.20users.20view

chrisbobbe (commented):
Thanks, @WesleyAC!

This looks good to me except for one commit-message nit that affects all three commits: when linking to CZO, I tend to use a link to a specific comment, rather than a whole topic. We don't know what discussions might happen in the topic after the commit is merged, and linking to a specific message is an easy way to avoid the kind of confusion where a reader mistakenly (but understandably) takes that later conversation to be part of the context that informed the commit. This confusion would be worse in a case like this where the link to a discussion is given as the entire explanation for the change.

gnprice:
Thanks! LGTM except the small thing Chris mentioned above.

PR #4865, opened 2021-07-06
UserList getItemLayout

WesleyAC:
A revision of #4809, now that we've talked about making `UserList` not a `SectionList` in our brainstorming. Changes `UserList` to be a `FlatList` instead of a `SectionList`, and then uses `getItemLayout` on it to improve performance.


src/users/UserList.js
```diff
@@ -47,7 +47,6 @@ export default function UserList(props: Props) {
       keyboardShouldPersistTaps="always"
       initialNumToRender={20}
       sections={sections}
-      keyExtractor={item => item}
```
chrisbobbe, line 50:
```
UserList [nfc]: Remove keyExtractor.

This was redundant with the `key` prop on the UserItems.
```

`key` isn't quite correctly called a prop; I tend to call it the "`key` attribute". You can't access it on `props`, and of course it has its special role that props don't have (https://reactjs.org/docs/lists-and-keys.html, https://reactjs.org/docs/reconciliation.html#keys).

src/users/UserList.js
```diff
@@ -1,13 +1,13 @@
 /* @flow strict-local */
 import React from 'react';
-import { SectionList } from 'react-native';
+import { FlatList } from 'react-native';
```
chrisbobbe, line 3:
```
[...] all elements will be a static 64px
high.
```

This isn't true yet. 🙂 The elements still have text in them (the user's full name, and possibly their email), the user can change the system font size, and the items expand with the font size increases too much. At the tip of this PR branch, if I add `fontSize: 72` to `componentStyles` in `UserItem`, I can see this:

src/users/UserList.js
```diff
@@ -41,6 +41,11 @@ export default function UserList(props: Props) {
       data={sortedUsers}
       keyboardShouldPersistTaps="always"
       initialNumToRender={20}
+      getItemLayout={(data, index) => ({
+        length: USER_ITEM_HEIGHT,
```
chrisbobbe, line 45: (See previous comment about the fixed height.)

chrisbobbe (changes requested):
Thanks, @WesleyAC! Comments below.

PR #4876, opened 2021-07-07
Use `useSelector` instead of `connect` in several more places.

chrisbobbe:
Following #4844, toward issue #4837, here are some more that I did today.

### DIFF
--- a/src/sharing/SharingScreen.js
+++ b/src/sharing/SharingScreen.js
@@ -1,5 +1,5 @@
 /* @flow strict-local */
-import React, { PureComponent } from 'react';
+import React from 'react';
 import {
   createMaterialTopTabNavigator,
   type MaterialTopTabNavigationProp,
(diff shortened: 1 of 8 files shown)


src/sharing/SharingScreen.js
```diff
@@ -50,51 +47,45 @@ const styles = createStyleSheet({
...
-
-    // If there is no active logged-in account, abandon the sharing attempt,
-    // and present the account picker screen to the user.
-    if (auth === undefined) {
-      NavigationService.dispatch(navigateToAccountPicker());
-      return null;
-    }
+export default function SharingScreen(props: Props) {
+  const { params } = props.route;
+  const auth = useSelector(tryGetAuth);
```
gnprice, line 52: Looking at this, I see that we're not actually using it for more than its truthiness value. So we could use `getHasAuth` instead, which would be a bit more transparent about the purpose and also make one fewer place there's an `Auth` value floating around.

gnprice (commented):
Thanks! All looks good; merging.

One comment below on a possible followup.

src/sharing/SharingScreen.js
chrisbobbe: Makes sense! #4878.

PR #4879, opened 2021-07-07
lightbox [nfc]: Remove `fileCache` property from rnfb config.

AkashDhiman:
This is not required as 'path' with `CacheDir` fulfills the same
purpose as `fileCache`.

The difference is 'path' gives specific path to the file while
`fileCache` gives a randomly generated path (including the
filename and its extension). See [1].

We would want to preserve the filename and its extension while
sharing from lightbox and hence `path` is used instead of
`fileCache`.

Originally reported at [2] by Greg.

reference:
[1] https://github.com/joltup/rn-fetch-blob#download-to-storage-directly
[2] https://chat.zulip.org/#narrow/stream/48-mobile/topic/Sharing.20from.20lightbox.20broken.20on.20Android/near/1226489

**tested to verify that this is nfc.**

### DIFF
--- a/src/lightbox/download.js
+++ b/src/lightbox/download.js
@@ -72,5 +72,4 @@ export const downloadImage = async (url: string, fileName: string, auth: Auth): 
 export const downloadFileToCache = async (tempUrl: string, fileName: string): Promise<mixed> =>
   RNFetchBlob.config({
     path: `${RNFetchBlob.fs.dirs.CacheDir}/${fileName}`,
-    fileCache: true,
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks, looks good! Merging.

I added a line to the commit message to say explicitly that `fileCache` is documented to have no effect when `path` is present, and link to the corresponding bit of docs:
https://github.com/joltup/rn-fetch-blob/wiki/Classes#pathstring-1
(Which I guess I didn't point very clearly to when initially reporting this, oops.)

Issue #4850, opened 2021-07-02
Make "Do not mark messages read on scroll" option findable


alya:
We have a "Do not mark messages read on scroll" setting, which can be quite useful for certain workflows. At present, this setting is hidden in Gear menu > Diagnostics > Debug.

To make it discoverable, we should move this setting to the top level of the gear menu. It can be placed right below the "Open links with in-app browser" option.

We should fix #4849 before making this setting finable.

Issue #4849, opened 2021-07-02
Preserve left-side unread indicator when not marking messages as read


alya:
Within the debug menus, we have an option not to mark messages as read on scroll. We want to make this option more prominent (#4850).

Before doing so, we should fix the fact that the left-side line indicating unread messages still disappears upon scrolling even though messages are not being marked as read.

gnprice:
To do this, we'll want to pass the value of this setting into the webview. The left-side unread indicators are controlled by the within-webview code, in the function `sendScrollMessage` and its callee `setMessagesReadAttributes`. We'll want to just do nothing there, in the case where this setting is enabled.

PR #4880, opened 2021-07-07
Fix do not mark as read, and make it more visible

AkashDhiman:
Done by first passing `doNotMarkAsRead` information to webview then
migrating `doNotMarkAsRead` from debug setting to general setting.

Fixes: #4849
Fixes: #4850

### DIFF
--- a/src/boot/store.js
+++ b/src/boot/store.js
@@ -302,6 +302,9 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     outbox: state.outbox.filter(o => o.sender_id !== undefined),
   }),
 
+  // Add `doNotMarkMessagesAsRead` in `SettingsState`.
+  // (Handled automatically by merging with the new initial state.)
+
   // TIP: When adding a migration, consider just using `dropCache`.
 };
 
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -295,6 +295,7 @@ export type SettingsState = {|
   experimentalFeaturesEnabled: boolean,
   streamNotification: boolean,
   browser: BrowserPreference,
+  doNotMarkMessagesAsRead: boolean,
 |};
 
 export type StreamsState = Stream[];
--- a/src/session/__tests__/sessionReducer-test.js
+++ b/src/session/__tests__/sessionReducer-test.js
@@ -110,7 +110,7 @@ describe('sessionReducer', () => {
     const action = deepFreeze({ type: DEBUG_FLAG_TOGGLE, key: 'someKey', value: true });
     expect(sessionReducer(baseState, action)).toEqual({
       ...baseState,
-      debug: { doNotMarkMessagesAsRead: false, someKey: true },
+      debug: { someKey: true },
     });
   });
 
--- a/src/webview/js/js.js
+++ b/src/webview/js/js.js
@@ -49,6 +49,13 @@ import { toggleSpoiler } from './spoilers';
  */
 declare var platformOS: string;
 
+/**
+ * used to control behavior based on debug settings.
+ * defined in `handleInitialLoad`.
+ * declared globally so as to use across functions.
+ */
+declare var doNotMarkMessagesAsRead: boolean;
+
 /* eslint-disable no-extend-native */
 
 /* Polyfill Array.from. Native in Chrome 45 and at least Safari 13.
@@ -507,7 +514,9 @@ const sendScrollMessage = () => {
     startMessageId: rangeHull.first,
     endMessageId: rangeHull.last,
   });
-  setMessagesReadAttributes(rangeHull);
+  if (!doNotMarkMessagesAsRead) {
+    setMessagesReadAttributes(rangeHull);
+  }
   // If there are no visible + read messages (for instance, the entire screen
   // is taken up by a single large message), then we don't want to update
   // prevMessageRange.  This way, if the user scrolled past some messages to
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -752,7 +752,7 @@ export const backgroundData: BackgroundData = deepFreeze({
   allImageEmojiById: action.realm_init.data.realm_emoji,
   auth: selfAuth,
   debug: baseReduxState.session.debug,
-  doNotMarkMessagesAsRead: baseReduxState.session.debug.doNotMarkMessagesAsRead,
+  doNotMarkMessagesAsRead: baseReduxState.settings.doNotMarkMessagesAsRead,
   flags: baseReduxState.flags,
   mute: [],
   mutedUsers: Immutable.Map(),
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -752,6 +752,7 @@ export const backgroundData: BackgroundData = deepFreeze({
   allImageEmojiById: action.realm_init.data.realm_emoji,
   auth: selfAuth,
   debug: baseReduxState.session.debug,
+  doNotMarkMessagesAsRead: baseReduxState.session.debug.doNotMarkMessagesAsRead,
   flags: baseReduxState.flags,
   mute: [],
   mutedUsers: Immutable.Map(),
--- a/src/session/sessionReducer.js
+++ b/src/session/sessionReducer.js
@@ -85,9 +85,7 @@ const initialState: SessionState = {
   orientation: 'PORTRAIT',
   outboxSending: false,
   pushToken: null,
-  debug: {
-    doNotMarkMessagesAsRead: false,
-  },
+  debug: Object.freeze({}),
   hasDismissedServerCompatNotice: false,
 };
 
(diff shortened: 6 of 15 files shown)


src/boot/store.js
```diff
@@ -99,6 +99,7 @@ function dropCache(state: GlobalState): $Shape<GlobalState> {
     // $FlowFixMe[prop-missing]
     // $FlowFixMe[incompatible-variance]
     // $FlowFixMe[incompatible-type-arg]
+    // $FlowFixMe[incompatible-shape]
```
AkashDhiman, line 102: Is this an appropriate thing to do? I have made the type `Debug = $Shape<{||}>` and flow was giving incompatible shape error on that, I am assuming here that we won't pass through `SessionState` (which contains `debug: Debug`) here because `storeKeys` don't contain that value and hence we can ignore the error related to that. Is that correct?

src/webview/js/js.js
```diff
@@ -49,6 +49,8 @@ import { toggleSpoiler } from './spoilers';
  */
 declare var platformOS: string;
 
+let doNotMarkMessagesAsReadFlag = false;
```
AkashDhiman, line 52: I am using a global variable here since we need to use this in multiple functions inside `js.js`, it is defined within `handleInitialLoad` and used within `sendScrollMessage`, Is that appropriate? does there exists some better design pattern that I can follow here?

AkashDhiman (commented):
I am only partially removing things related to debug settings (like `DebugScreen`, and `debug` property inside `BackgroundData`) since I feel like the code may be reusable when we add another debug setting in the future.

We can squash the last 2 commits, I have made them separate so that It will be easy to review. Let me know if this is required or not.

It would be great if someone reviewing this can take a look at a few doubts I had (see below)

src/boot/store.js
gnprice: Good question. I think the answer here is to not use `$Shape`:
* In general, `$Shape` has some funny unsound behavior (https://flow.org/en/docs/types/utilities/#toc-shape ) and so we usually avoid it.
* Here, `{||}` already means exactly the type of objects with no properties. So there isn't anything for `$Shape` to do, other than its funny unsoundness.

It looks like if I delete the `$Shape` I get an error at the point where it's initialized to `{}`. That is a known Flow quirk, where the empty object literal `{}` can't be used as the empty-object type `{||}`. Fortunately there is a clean workaround, which is to say `Object.freeze({})` instead: https://github.com/facebook/flow/issues/2386#issuecomment-695064325

src/webview/js/js.js
gnprice: Yeah, good question.

I think a somewhat better pattern would be to follow what we do for `platformOS`:
* It's defined directly in `script.js`, before the code in `compiledWebviewJs` (which has all the code in `js.js` and its friends) runs. This means it's initialized with the correct value from the beginning, and there's no need to worry about initialization order.
* Then there's a `declare var` line here at the top of `js.js`, so that Flow knows about it.

That's still kind of awkward, because there's no type-checking that what we do in `script.js` matches what `js.js` expects. But that's equally true with the parameters we pass to `handleInitialLoad`.

I think having it in a global is fine, so long as it has a nice explicit name like this. There's kind of an underlying messiness in the fact that this file is pretty long and does quite a lot of different things. So that makes it kind of uncomfortable to add more things to its top-level namespace. But I think the main thing for us to do there is to take some of the larger clusters of related code in this file and move them to separate modules.

gnprice:
Thanks @AkashDhiman ! I replied to your inline questions above.


> We can squash the last 2 commits, I have made them separate so that It will be easy to review. Let me know if this is required or not.

Yeah, I think those two changes will be best squashed into one commit. That's because they have the nature of removing a thing, and then adding back a revised version of the same thing. When squashed together, both halves of that change become easier to understand and review, because the whole thing can be seen as some moving and editing.

The other branch-level thing I see is that the first commit adds some plumbing to bring the whole `debug: Debug` through into the webview, then the later commits remove that and replace it with plumbing for just `doNotMarkMessagesAsRead`. The series could be simplified by having the first commit plumb through just `doNotMarkMessagesAsRead`, which it can get from `debug`. Then the later changes don't need to touch any of the message-list or webview internals at all -- just switch where that flag comes from, at the outer layer of that subsystem.

I have some small code-level comments too, which I'll make inline.

src/session/__tests__/sessionReducer-test.js
```diff
@@ -106,14 +105,6 @@ describe('sessionReducer', () => {
     ).toEqual({ ...baseState, outboxSending: true });
   });
 
-  test('DEBUG_FLAG_TOGGLE', () => {
-    const action = deepFreeze({ type: DEBUG_FLAG_TOGGLE, key: 'someKey', value: true });
-    expect(sessionReducer(baseState, action)).toEqual({
-      ...baseState,
-      debug: { doNotMarkMessagesAsRead: false, someKey: true },
-    });
-  });
```
gnprice, line 115: This is a test for some logic we still have, in `sessionReducer.js`. So we should delete it only when we're actually deleting that logic.

That'll be a separate commit from any of the changes in this branch -- basically we'd have a commit ripping out the `Debug` type and `DebugScreen`, and all the remaining references to those. I think it'd be pretty reasonable to include that at the end of this PR, but I'm also happy to let it sit around for a bit and see if we find another use for it soon.

src/reduxTypes.js
```diff
@@ -295,6 +295,7 @@ export type SettingsState = {|
   experimentalFeaturesEnabled: boolean,
   streamNotification: boolean,
   browser: BrowserPreference,
+  doNotMarkMessagesAsRead: boolean,
```
gnprice, line 298: We're making a change to the Redux state type, so we should have a migration for it.

(More specifically, it's in a subtree that we actually store between runs -- see `storeKeys` and `cacheKeys` in `src/boot/store.js`. This didn't apply to the change to `SessionState` via changing `Debug`, because `state.session` isn't stored; it's in `discardKeys`.)

Actually, we had a recent discussion where the conclusion was that maybe we shouldn't quite have a migration in this case, because the way the persisted state gets merged with the reducer's initial state will have the right effect already. But still we should have a comment in the list of migrations, saying what changed and that we explicitly decided no migration was needed.

(I'm regretting now that that discussion was on GitHub instead of Zulip, though -- it may not be easy to find.)

src/webview/js/js.js
```diff
@@ -679,9 +683,10 @@ export const handleInitialLoad = (
   // The `realm` part of an `Auth` object is a URL object. It's passed
   // in its stringified form.
   rawAuth: {| ...$Diff<Auth, {| realm: mixed |}>, realm: string |},
+  doNotMarkMessagesAsRead: boolean,
 ) => {
   const auth: Auth = { ...rawAuth, realm: new URL(rawAuth.realm) };
-
+  doNotMarkMessagesAsReadFlag = doNotMarkMessagesAsRead;
   // Since its version 5.x, the `react-native-webview` library dispatches our
```
gnprice, line 690: nit: keep the blank line -- it's serving a purpose by separating this comment from what's above, so that it's visually attached to the whole stanza below

gnprice (commented):
Done reading through this; inline comments below. Thanks again!

src/reduxTypes.js
chrisbobbe: https://github.com/zulip/zulip-mobile/pull/4798#discussion_r652281059 🙂

src/boot/store.js
AkashDhiman: Good to know! thanks.

src/webview/js/js.js
AkashDhiman: passing it like `platformOS` in new revision.

src/session/__tests__/sessionReducer-test.js
AkashDhiman: added back, I am keeping `debug` related code as it is, for potential future use.

src/reduxTypes.js
AkashDhiman: is there a difference when using `dropCache` vs returning a modified state like some migrations are doing? example - '28'

src/boot/store.js
```diff
@@ -302,6 +302,9 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     outbox: state.outbox.filter(o => o.sender_id !== undefined),
   }),
 
+  // Add `doNotMarkMessagesAsRead` in `SettingsState`.
+  '30': dropCache,
```
gnprice, line 306: This isn't the right thing here; it actually has no effect on the `settings` state. (You asked about this line in a subthread above, too.)

I guess I just said above that I regretted having a previous discussion on a GitHub PR thread instead of in Zulip because it's harder to find here, so I'll make a chat thread to go into more detail :slightly_smiling_face:

gnprice (commented):
Thanks for the revision!

All looks good, except the migration (which you asked about above.) I'll fix that up and merge.

src/boot/store.js
gnprice: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20migrations/near/1230029

src/reduxTypes.js
gnprice: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/data.20migrations/near/1230029 (same link as below, but here too for cross-reference)

Issue #4886, opened 2021-07-08
Add release-build option to `tools/run-android`.


chrisbobbe:
So it's easier to test changes (such as those made for a React Native upgrade) that we reasonably expect might cause different behaviors in debug vs. release builds.

chrisbobbe:
https://github.com/zulip/zulip-mobile/issues/4929#issuecomment-908603249 is somewhere where a user wanted to test something on Android in production mode instead of development.

PR #4887, opened 2021-07-08
Follow some template-app changes on the way to RN v0.64.

chrisbobbe:
(Another one was #4846.)

Toward #4426, but this doesn't do the actual upgrade in `package.json` yet.

I've tested this in debug and release builds on my iPhone, and in a debug build on the office Android device (see #4886 for release). From some poking around in the message list and elsewhere, all seems fine.

The most important-looking change, to me, looks like

```
metro config: Turn on inline requires.
```

This is a feature that's been available since RN v0.59, but that they've now started turning on in the template app. It gets its own section in the RN v0.64 release announcement: https://reactnative.dev/blog/2021/03/12/version-0.64#inline-requires-enabled-by-default. The [RN v0.64 changelog entry](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0640) warns of a breaking change: "slightly different JS execution order". I haven't yet noticed any problems this seems to cause.

I plan to post screenshots of the Xcode UI for a few commits where I think that'll be helpful. (**edit**: done; see the two comments that follow.)


chrisbobbe:
To follow some of facebook/react-native@6685aba46:

```
ios build: Run "Update to recommended settings" for ZulipMobile project.
```

update-to-recommended-settings-rn-6685aba46

chrisbobbe:
To follow facebook/react-native@538446e75:

```
ios build: Bump Xcode compatibility to 12.0.
```

bump-xcode-rn-538446e75

gnprice:
> metro config: Turn on inline requires.

It looks like this is the RN docs on this feature:
https://reactnative.dev/docs/ram-bundles-inline-requires

It's not real clear about the role of that Metro config option. The option doesn't appear to be documented in the Metro docs:
https://facebook.github.io/metro/docs/configuration

My guess as to what the option means is that when we have a `require` function call (as opposed to an `import` statement), when this option is false, the required module actually ends up getting imported statically. (I.e., its top-level code gets executed before the requiring module's top-level code is executed.) And when the option is true, then the required module only gets imported lazily, if and when the `require` call actually happens.

If that's right, then the new behavior certainly makes more sense. Also, if that's right then the effect is limited to places where we have `require` calls, which we have very few of … in fact, a quick grep tells me we have just one outside of tests, namely where we import `immutable-devtools` conditioned on a debug build.

Anyway, we probably don't need to pin down exactly what this undocumented feature actually does. As long as things still seem to work, I'm happy to assume it didn't break anything.

ios/Podfile
```diff
@@ -7,12 +7,9 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
...
 
 target 'ZulipMobile' do
-  # Pods from React Native
-  use_react_native!
+  config = use_native_modules!
 
-  # "Autolinking": automatically link pods that depend on React
-  # Native, unless omitted in our own `react-native.config.js`.
-  use_native_modules!
+  use_react_native!(:path => config[:reactNativePath])
```
gnprice, line 12: This isn't super well explained, but it looks like their claim for
    this being needed starts by mentioning
    @react-native-community/cli@e949e234b, which was released in v4.10.1
    of @react-native-community/cli.
    
    And `yarn why` tells me that we're on v4.14.0 of that package, so it

This probably means the intermediate state with `config["reactNativePath"]` is broken, right? So probably we should just squash these two changes together.

.prettierrc.js
```diff
@@ -13,4 +13,5 @@ module.exports = {
   parser: 'babel-flow',
   singleQuote: true,
   trailingComma: 'all',
+  arrowParens: 'avoid',
```
gnprice, line 16: commit-message comment:

    That commit reports that the default behavior before Prettier 2.0 is
    like setting `arrowParens: 'avoid'`. It's looking unlikely that we
    have an easy path to using Prettier 2.0 (see
    prettier/prettier-eslint-cli#304), but we might as well follow the
    RN template app.

This looks like a previous draft of the same material that's in the rest of the commit message. Probably you meant to delete this version?

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -315,45 +310,12 @@
...
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Foundation.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Ionicons.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/MaterialCommunityIcons.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/MaterialIcons.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Octicons.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/SimpleLineIcons.ttf",
-				"${PODS_ROOT}/../../node_modules/react-native-vector-icons/Fonts/Zocial.ttf",
-				"${PODS_CONFIGURATION_BUILD_DIR}/React-Core/AccessibilityResources.bundle",
+			inputFileListPaths = (
+				"${PODS_ROOT}/Target Support Files/Pods-ZulipMobile/Pods-ZulipMobile-resources-${CONFIGURATION}-input-files.xcfilelist",
```
gnprice, line 314: Neat -- this looks like a definite improvement in decoupling our Xcode project file from the details of the libraries we use.

From the way you produced this commit, it sounds like this is probably something CocoaPods has supported doing for at least a little while, but depends on a newish Xcode feature (this `inputFileListPaths` option) that wasn't available in Xcode 3.2 (!), so CocoaPods couldn't make use of it as long as the `compatibilityVersion` setting was so ancient.

package.json
```diff
@@ -88,9 +88,9 @@
     "zulip-markdown-parser": "^1.0.6"
   },
   "devDependencies": {
-    "@babel/core": "^7.11.4",
+    "@babel/core": "^7.12.9",
     "@babel/preset-env": "^7.11.0",
-    "@babel/runtime": "^7.8.4",
+    "@babel/runtime": "^7.12.5",
```
gnprice, line 93: deps: Follow RN in updating some dev dependencies.

Worth mentioning somewhere in the commit message that this doesn't change any of the versions we actually end up using -- our `yarn.lock` already contained versions as new as all of these. It just bumps the explicit lower bounds in the `package.json`.

ios/ZulipMobile.xcodeproj/project.pbxproj
```diff
@@ -414,8 +376,11 @@
 				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
 				HEADER_SEARCH_PATHS = "$(inherited)";
 				INFOPLIST_FILE = ZulipMobile/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/Frameworks",
+				);
```
gnprice, line 383: This is also a nice change. The old version looks like it probably would have broken if you have a directory name with a space (` `) in it somewhere above your worktree! That is a classic kind of bug for old shell scripts to have… but even there it rather went out of style sometime in the aughts.

gnprice (commented):
Thanks for all this upgrade work! I appreciate the clear descriptions of where these changes came from -- particularly the Xcode project file, where the changes can easily get gnarly.

Comments below, all small (and some are just remarks on the nice changes from upstream, nothing to change.) Please merge at will modulo those.

.prettierrc.js
chrisbobbe: Thanks!

PR #4836, opened 2021-06-30
Flow-related work on the path to RN v0.64 upgrade, part 3.

chrisbobbe:
Parts 1 and 2 were #4518 and #4520.

As I explained at https://github.com/zulip/zulip-mobile/issues/3452#issuecomment-796388640,

> We'll want to fix [#3452] on the path to the RN v0.64 upgrade (#4426). [facebook/react-native@050a7dd](https://github.com/facebook/react-native/commit/050a7dd01) turns on `exact_by_default` with the warning that RN code might assume that projects have turned it on.
> 
> #4518 has been merged, toward this, and I've just opened #4520.

The first few commits fix a few small things that have crept into our own code since #4520, and then the rest deal with the libdefs.

Fixes: #3452

PR #4891, opened 2021-07-09
Flow-related work on the path to RN v0.64 upgrade, part 4.

chrisbobbe:
Parts 1, 2, and 3 were #4518, #4520, and #4836.

This doesn't upgrade Flow at all or add any new config; it's just cleaning up some stuff that I noticed (one commit is another round like #4433, for example). These kinds of small messes could easily creep back in before we add config to prevent them for good, but might as well sweep things up as we see them.

After this, the work I see before the RN v0.64 upgrade (#4426) is:

- Types-first (as noted in the issue, we can probably do some selective suppressions there for now, but I'm not sure that'll hold in future Flow versions)
- Turning on `exact_by_default`
- (Maybe an issue with react-intl's libdef; see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-intl.20libdef/near/1228275))

For `exact_by_default`, I think I have a handle on all the little issues in our libraries' built-in Flow definitions, so that won't be hard unless new problems come up (e.g., with libraries or library versions we aren't using right now). One of those issues might best be solved with a small one-line PR to @react-native-community/cameraroll; hopefully they'd take it.

### DIFF
--- a/flow-typed/react-intl_vx.x.x.js
+++ b/flow-typed/react-intl_vx.x.x.js
@@ -158,19 +158,19 @@ declare module 'react-intl' {
     +MISSING_INTL_API: 'MISSING_INTL_API', // "MISSING_INTL_API"
   |};
   declare interface FieldData {
-    // $FlowFixMe - illegal name (these fixmes added in TS to Flow translation)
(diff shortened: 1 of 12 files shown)


gnprice:
Looks good, thanks! Merging.

PR #4896, opened 2021-07-14
Clean up API error-distinguishing logic

gnprice:
This is what I ended up with after my comment https://github.com/zulip/zulip-mobile/pull/4754#pullrequestreview-705680950 :

> [The] existing code [in `apiErrors`] should change, I think, to handle this properly. I'll sketch out a version of how that could look, and either push some draft commits to this branch or send a PR.

It was kind of a tangle, and now I hope it's clearer. In particular, I think it's now a lot more straightforward to think of the set of possible errors on an API request as completely partitioned into meaningful subcategories, and to write conditionals on those.

The `isClientError` predicate goes away in this branch, replaced by a direct `instanceof` check at its former callsite. I think the same thing can be done with `isServerError` and `isNetworkRequestFailedError` in #4754; they can become `instanceof` checks referring to the new error classes added here.

The last few commits in this branch are some cleanups to the API result types, which aren't needed for anything else in this branch. I wrote those on the way toward trying to make `interpretApiResponse` return an `ApiSuccessResponse` instead of `mixed`, but the change to actually do that is still a draft.

This inherits P1 priority from #4754 .


src/api/apiFetch.js
```diff
@@ -52,15 +52,18 @@ export const apiCall = async (
...
+    const { httpStatus, data } = error;
+
     // eslint-disable-next-line no-console
-    console.log({ route, params, httpStatus: response.status, json });
+    console.log({ route, params, httpStatus, data });
     Sentry.addBreadcrumb({
       category: 'api',
       level: 'info',
-      data: { route, params, httpStatus: response.status, json },
+      data: { route, params, httpStatus, data },
```
chrisbobbe, line 82:
```
api errors [nfc]: Rely on error object in logging code.
```

nit: Breadcrumbs sent to Sentry will look different after this commit, so I think that makes it non-NFC.

src/api/apiFetch.js
```diff
@@ -37,12 +37,13 @@ export const apiFetch = async (
   params: $Diff<$Exact<RequestOptions>, {| headers: mixed |}>,
 ) => fetch(new URL(`/${apiVersion}/${route}`, auth.realm).toString(), getFetchParams(auth, params));
 
+/** (Caller beware! Return type is the magic `empty`.) */
```
chrisbobbe, line 45: > `fetch` having a type of `any` in the libdef RN provides

Yeah; it'll be nice to get facebook/react-native@6651b7c59, looks like in RN v0.65.

src/api/apiErrors.js
```diff
@@ -11,7 +21,7 @@ import * as logging from '../utils/logging';
  * See docs: https://zulip.com/api/rest-error-handling
  */
 // TODO we currently raise these in more situations; fix that.
-export class ApiError extends Error {
+export class ApiError extends RequestError {
   code: ApiErrorCode;
```
chrisbobbe, line 23: Could also have a commit that makes `ApiError` and `ServerError` read-only, right?

chrisbobbe (commented):
Thanks for this cleanup, @gnprice! See a few tiny comments below; otherwise, please merge at will.

src/api/apiFetch.js
gnprice: Mmm, indeed, good catch.

I think actually the name should ideally say something like "result" or "response" -- neither "json" nor "data" is real clear that this is the data the server sent back in the body, rather than say the data that we sent in the request. I'll pick a name like that.

src/api/apiFetch.js
gnprice: Ah indeed, neat!

src/api/apiErrors.js
gnprice: Sure, yeah.

I guess when I did that on the base class:
```js
export class RequestError extends Error {
  +httpStatus: number | void;
  +data: mixed;
```
it was just because that's necessary in order for the subclasses to give more-specific types for those properties. But the same thing would be appropriate on the subclasses' properties too.

gnprice:
Thanks for the review! Merged, with those changes.

Issue #4859, opened 2021-07-05
Enable fetching more search results


alya:
At present, we fetch the 20 most recent search results, and do not provide a way for the user to get more. This can lead to a bad experience when the user is looking for an older message. We should:

1. Fetch additional search results when the user scrolls up from the search results that have already been fetched.
2. Consider fetching more results (100?) in the first place, as 20 is quite low when the query is not very specific.

gnprice:
The implementation of this will require some refactoring and be a bit tricky.

For the message list in all cases other than search, we get the messages from our global Redux store, and the way to fetch more is to go and do a fetch that puts the results into the global Redux store. So that's what the `fetchMore` function in `src/webview/handleOutboundEvents.js` does, dispatching a `fetchOlder` or `fetchNewer` thunk action.

But in the case of search, the list of messages isn't in the global Redux state; instead it's in the local state of the `SearchMessagesScreen` component. We do that because, especially with search-as-you-type causing lots of searches, we don't want to be keeping around indefinitely the results of all the different searches you've done.

As a result, in order to fetch more messages and have them show up in the list, we need to do a fetch that will put the results into that screen's local state.

The first step we'll need for this is to wire things up so that the parent of `MessageList` passes in a pair of callbacks to tell it how to fetch more messages. These can probably just be props called `fetchOlder` and `fetchNewer`, and then in `handleOutboundEvents.js` we'll have
```diff
-    dispatch(fetchOlder(narrow));
+    fetchOlder();
```
and similarly for `fetchNewer`.

That step should be a pure refactor with [no functional change](https://github.com/zulip/zulip-mobile/blob/cf8f1a841dc27f2ba1bf32513e886a299f340822/docs/glossary.md#nfc-or-nfc); in the search case we can pass no-op callbacks for those props.

Then, we'll want to implement those callbacks for search. This will interact with the other state on `SearchMessagesScreen`. Here's some thoughts on that:
* The existing state structure has some trickiness because the query whose results we're showing isn't necessarily the latest one the user entered, if that latest one hasn't yet gotten a response or didn't succeed.
* We should be sure we never end up showing a mixture of results from one query and results from another.
* On scroll, the query we fetch more messages for should probably be the one we're currently showing messages for. That means we should start remembering what that query was, with a state property like `query: string` that we update at the same time as `messages`.
[...]

PR #4899, opened 2021-07-15
Implement stream action sheets

AkashDhiman:
Also involves moving `messageActionSheet` to `src/action-sheet/index.js`.

Preview:
WhatsApp Image 2021-07-15 at 5 43 39 PM


alya:
Tiny comment on the screenshot: "Pin to Top" should be "Pin to top".

src/webview/html/messageHeaderAsHtml.js
```diff
@@ -62,7 +62,7 @@ export default (
 
   if (item.type === 'stream' && headerStyle === 'full') {
     const streamName = streamNameOfStreamMessage(item);
-    const stream = subscriptions.find(x => x.name === streamName);
+    const stream = [...subscriptions.values()].find(x => x.name === streamName);
```
gnprice, line 65: The change at this line risks being significantly slower: the old code is linear (which isn't good), and the new code is still linear, but the new code does a linear amount of *allocation*, building up this new array, and that's potentially quite a bit more costly than just searching through an array.

And this line is in the code that generates the HTML for the message list, so it's something we potentially do many times if the user is looking at a view with a lot of history (provided it's an interleaved view where the stream name appears in headers -- for example the @-mentions or starred-messages views, or a search view once we have #4859.)

Happily, I think there is an easy solution! We already have the stream ID here, because `item` is a `Message | Outbox` with `type: 'stream'`, and those have a `stream_id`.

So we can just use `subscriptions.get(item.stream_id)`, much like in the spot in `action-sheets/index.js` that motivates this change.

src/action-sheets/index.js
```diff
@@ -232,6 +288,45 @@ const cancel = params => {};
...
+    }
+    if (sub.pin_to_top) {
+      buttons.push(unpinFromTop);
+    } else {
+      buttons.push(pinToTop);
+    }
+    if (sub.push_notifications === true) {
+      buttons.push(disableNotifications);
+    } else {
+      buttons.push(enableNotifications);
```
gnprice, line 319: This logic isn't quite right. See the existing place we look at this `push_notifications` property, in `StreamSettingsScreen`:
```js
    const currentValue = subscription?.push_notifications ?? userSettingStreamNotification;
```

That is, if `push_notifications` is nullish, then that doesn't necessarily mean the same as false -- it means it falls back to the value the user has chosen as their default, which may be true.

We should probably put that logic in one place; it's already a little awkward that it's duplicated within `StreamSettingsScreen.js`, and it'd be much more awkward (and much more risk of divergence) to have it duplicated across multiple files.

src/streams/StreamItem.js
```diff
@@ -75,6 +90,21 @@ export default function StreamItem(props: Props): Node {
...
 
+  const showActionSheetWithOptions: ShowActionSheetWithOptions = useActionSheet()
+    .showActionSheetWithOptions;
+  const _ = useContext(TranslationContext);
+  const dispatch = useDispatch();
+  const backgroundData = useSelector(state => ({
+    auth: getAuth(state),
+    ownUser: getOwnUser(state),
+    streams: getStreamsById(state),
+    streamsByName: getStreamsByName(state), // TODO might need to pass stream by id here.
```
gnprice, line 101: Is this TODO still applicable?

gnprice: Also, it looks like `streamsByName` isn't needed for `backgroundData`, only for finding `stream` (and thereby the stream's ID.) So it'd be a bit cleaner to have that outside of `backgroundData`.

That would also then be helpful when we go on to have `StreamItem`'s parent pass it the stream ID, rather than making `StreamItem` look it up; if `streamsByName` is on its own outside `backgroundData`, it'll be easier then to see that it's no longer needed and cut it out.

gnprice (commented):
Thanks @AkashDhiman! Generally this looks good. A few comments below.

> [a479b4b](https://github.com/zulip/zulip-mobile/commit/a479b4bacc92dc72a04a93512c80048fb25e3a69) seems like it might have been better as three separate commits, although I'm alright with it as-is.

Indeed, one advantage of splitting those changes would be that it'd make it easy to hold off on the "Enable push notifications" part (which has the only comment that isn't real simple to fix) while merging all the other changes in the PR :wink:

src/webview/html/messageHeaderAsHtml.js
gnprice: Hmm, I guess `Outbox` doesn't actually have `stream_id` yet: #3998. (Since #4667 it's there on the type as an optional property, but we never actually provide it.) So that simple solution won't yet work.

The most direct solution here would be to use the `subscriptions.values()` [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) just like you're doing, but use it directly with something like a for/of loop, to avoid allocating a big array of all the elements.

Another solution would be to go ahead and pass down a `subscriptionsByName` value into this code, gotten from `getSubscriptionsByName`.

AkashDhiman: using `for ... of`.

src/action-sheets/index.js
AkashDhiman: deduplicated it in `StreamSettingsScreen` by creating a common function in `src/stream`; although I am not sure if it's placed at the right location.

src/streams/StreamItem.js
AkashDhiman: removed streamsByName.

AkashDhiman:
@gnprice Did the required chances, you may review again.

1 more thing I did on my own in this revision is that I am using `api.setSubscriptionProperty` instead of `dispatch(setSubscriptionProperty)` used previously.
is there any difference among them?

gnprice:
(I've just rebased this and resolved conflicts.)

src/streams/getIsNotificationEnabled.js
```diff
@@ -0,0 +1,7 @@
+/* @flow strict-local */
+import type { Subscription } from '../types';
+
+export default (
+  subscription: Subscription | void,
+  userSettingStreamNotification: boolean,
+): boolean => subscription?.push_notifications ?? userSettingStreamNotification;
```
gnprice, line 7: Cool, this solves the problem.

Some nits I would do differently:
* This is really about a subscription (the user's relationship to a stream, represented by a `Subscription` value) rather than a stream itself (represented by `Stream`), so I'd put it in `src/subscriptions/` or perhaps `src/settings/`. (This is a distinction our existing code isn't super consistent about.)
* I've found it helpful to shift toward fewer, medium-size files rather than lots of tiny files. So instead of giving this its own file, I would make it the first thing added to a file named `src/subscriptions/subscriptionsModel.js` ~~or `src/settings/settingsModel.js`~~ (definitely "subscriptions"; see below). See b21d7c58de80f40b15ca7de68c2575c5824fd9de (`muteModel`) for a recent example that's equally tiny for now. See `src/unread/unreadModel.js` for a larger example, with the reducer in the same file; this function would be a "getter", a lot like the function `getUnreadCountForTopic` (added in 0b4c2a446a62d424456cf61005cb23db8876f184) there.
* I think my ideal version of this function would take the whole `PerAccountSettingsState` object, i.e. the return value of `getSettings`, rather than this one flag from it. That way the details of which piece of the user settings is relevant are internal to this function. That also means that where we have code which is ultimately going to call both this function and a few more like it, that code only needs to pass around a single settings object, and doesn't have to keep adding more bits of settings to pass down through the chain each time it has to call another function like this one.
  * That is *very* recent, though -- I might not have felt the same way before splitting up `PerAccountSettingsState` vs `GlobalSettingsState`, which just happened in 3d98cf9c405da325ae740ea58b79c9fb749fbcc3 / #5017 a couple of weeks ago (and after you wrote this revision.) In any case I definitely haven't discussed or demonstrated this idea much in the past.
  * Also this may involve a bit of refactoring in the layers above where this gets called. Like I said, I think this will lead to less churn there as we add more functions like this, but it means a bit more right now.
* I'd like the name to somehow indicate that it's per-stream, as contrasted with something like `state.settings.offlineNotification` which feels like this same name `getIsNotificationEnabled` could describe it too. The types already mean there isn't a great danger of outright mixing this up with something that isn't per-stream, but still it'd be good to make it more immediately obvious to the reader.
  * Something I quite like about this name is that it makes clear it's getting an on/off setting! The first few names I think of that do indicate it's per-stream, plus align with the property name `push_notifications`, are like `getStreamPushNotifications` and they sound like they're getting some actual notifications, whatever that would mean. A name that has both these good properties will probably be somewhat wordy, but I think that's OK.

All of those are things I'm happy to take care of after we merge this PR, though -- no need to revise them here. This already solves the really key problem, which is that I want this logic in one place and with its own name.

---

Oh, but one pair of things I think I would like to have before merge:
* A bit of jsdoc saying what this is intended to mean.
* A jsdoc comment on the `push_notifications` property where `Subscription` is defined, saying to use this function to interpret it.

Those are key because the problem this is meant to solve is, when someone goes to use that property (just as you did in the original revision of this PR), we want the easy natural thing to be that they spot this function and use it so that they get the right behavior. So the comment from the property to help you find this function is a key part of that.

Hmm, and I guess thinking about that answers the question I was waffling on above about whether this goes in a "subscription" model or "settings" model: it should go in a `subscriptionsModel.js`, because this function is something you need in order to correctly interpret the subscriptions data. Or from another angle, the question this function is an answer to is a question that comes up when you're thinking about a subscription, not when you're thinking about settings.

src/webview/html/header.js
```diff
@@ -55,7 +55,13 @@ export default (
     `;
     } else if (headerStyle === 'full') {
       const streamName = streamNameOfStreamMessage(message);
-      const stream = subscriptions.find(x => x.name === streamName);
+      let stream = null;
+      for (const sub of subscriptions.values()) {
+        if (sub.name === streamName) {
```
gnprice, line 60: Let's add a comment here that says `TODO(#3339)` (with parens just like that -- a consistent format is helpful for grepping.) That's because this spot was and still is an instance of #3339, but in this form with an explicit loop it's not as easy to spot when searching for instances of that issue. So the TODO tag will help make sure we find it when we're closing in on finishing that issue.

static/translations/messages_en.json
```diff
@@ -118,7 +118,10 @@
   "Type a message": "Type a message",
   "Message group": "Message group",
   "starred": "starred",
-  "Enable notifications": "Enable notifications",
+  "Failed to subscribe": "Failed to subscribe",
+  "Failed to unsubscribe": "Failed to unsubscribe",
+  "Unpin from top": "Unpin from top",
+  "Failed to unpin from top": "Failed to unpin from top",
```
gnprice: In this and the next couple of commits, it looks like the changes in this file got a bit messed up.

I think the net changes at the end of the branch look right, though.

gnprice (commented):
Thanks @AkashDhiman for the revision! Sorry for the slow review; comments below.

There's just a few and they're all small things that are quick for me to fix myself. So I think I'm going to go do that and then merge.

---

> 1 more thing I did on my own in this revision is that I am using `api.setSubscriptionProperty` instead of `dispatch(setSubscriptionProperty)` used previously.
> is there any difference among them?

Good question. As you can see in the implementation of the `setSubscriptionProperty` thunk action, it's a pretty trivial wrapper around `api.setSubscriptionProperty` -- it's not really adding any new behavior.

I think the answer is that that wrapper (and the similar trivial thunk-action wrappers we have for some other API calls) shouldn't exist, and we should just use the `api.foo` methods directly, just as you do in this revision.

That's because I don't think there's any meaningful semantics being added by those wrappers. That is, not only do they not add any significant behavior or logic, but there also isn't any way their semantics meaningfully differs from the underlying `api.foo` methods -- it's not like there's something they're abstracting away, that just happens to have a trivial implementation but where the fact the implementation is trivial is itself a nonobvious fact. (For some examples of functions with trivial implementations that *do* nevertheless have nontrivial semantics, see `src/recipient.js` -- particularly `streamNameOfStreamMessage` or `pmTypingKeyFromPmKeyIds`.) So they don't gain us anything.

And conversely, the extra wrapper has a real cost -- an extra layer of indirection to jump through when you're reading the code and working out what it does.

src/action-sheets/index.js
gnprice: Looks good, thanks! See nits at https://github.com/zulip/zulip-mobile/pull/4899#discussion_r722831289 . In particular the location I'd put it is in a new file `src/subscriptions/subscriptionsModel.js`; discussion in that comment.

PR #4901, opened 2021-07-16
changelog + i18n: Document changes since v27.165; fix language tags

chrisbobbe:
EDIT: And add a few more languages to the UI, and do a Transifex sync.


static/translations/messages_ar.json
```diff
@@ -6,6 +6,9 @@
   "Private messages": "رسائل خاصة",
   "Starred": "مميزة بنجمة",
   "Mentions": "الإشارات",
+  "Message": "Message",
+  "Send": "إرسال",
+  "Please choose recipients to share with": "Please choose recipients to share with",
```
gnprice, line 11: Note: The "Portuguese (Brazil)" message, added in messages_en.json
    in this series, is added to the other languages in this commit,
    instead of in a separate "Sync recently-added message strings"
    commit. That's because I ran `tools/tx-sync` twice after that string
    was added, and I kept the result from the second time. This was a
    small mistake that I don't expect to cause real problems, but it
    does explain what happened with that string.

Yeah, and it looks like the same thing happened with a number of these other strings.

Nothing all that bad happens -- nothing at all goes wrong at runtime. The only effect is that it makes it hard to spot the actual new translations in the history. (You can get some idea by looking at the diffstat and seeing that some languages have more edits than others; but then there's also cases like this one where a new string already has a translation.)

Definitely not worth trying to go back and pick apart the changes into two commits. But that's the reason why I developed a habit of doing first a pull for new translations on existing strings, committing that, then a push + pull to sync translations across. And then codifying that habit into the tools/tx-sync script :slightly_smiling_face: 

Basically, the Transifex push that happens in the last step of tools/tx-sync has side effects -- it pushes strings to Transifex, and so can immediately change what the next pull from Transifex will get -- so you always want to push (in Git) the result of the first step of tools/tx-sync, where it pulled translations down from Transifex.

gnprice (commented):
Thanks! Reading the changelog now. Meanwhile, comments on the translations part of the branch.

docs/changelog.md
```diff
@@ -36,6 +36,24 @@ It doesn't include
...
+  (#4850)
+* Basic support for polls. More to come! (#3205)
+
+### Highlights for developers
+
+* All object types are now exact or explicitly inexact. (#3452)
+* We're now using @react-native-community/push-notifications-ios instead of
+  two different libraries. (#4115)
+* Resolved issues (latest to earliest): #117, #4165, #4858, #4850, #4849,
+  #3205, part of #4309 (PR #4817), #3244, #3452, parts of #4540 and #2366
```
gnprice, line 55: Also PR #4635, which added a (small) user-facing feature that we don't seem to have filed an issue for. (Goes right next to #3244, which was fixed in the same PR.)

And between #4115 and #4657, there's
  PR #4821 shorter debounce on mark-as-read
  PR #4820 fix a BAD_REQUEST 
  PR #4815 rm "Possible infinite loop"
  PR #4797 turn off lazy loading in MainTabsScreen
Those didn't have GitHub issues filed for them, but they each fixed a bug or made a small improvement in user-visible behavior.

gnprice (commented):
And comments on the changelog. Thanks again!

chrisbobbe:
Thanks for the review! Revision pushed. I haven't run `tools/tx-sync` after these tweaks to the translation logic; should I do that?

src/settings/languages.js
```diff
@@ -33,41 +34,42 @@ export type Language = {|
...
+  { tag: 'it', name: 'Italian', nativeName: 'Italiano' },
+  { tag: 'ja', name: 'Japanese', nativeName: '日本語' },
+  { tag: 'ko', name: 'Korean', nativeName: '한국어' },
+  { tag: 'lt', name: 'Lithuanian', nativeName: 'Lietuvių' },
+  { tag: 'ml', name: 'Malayalam', nativeName: 'മലയാളം' },
+  { tag: 'fa', name: 'Persian', nativeName: 'فارسی' },
+  { tag: 'pl', name: 'Polish', nativeName: 'Polski' },
+  { tag: 'pt', name: 'Portuguese', nativeName: 'Português' },
+  { tag: 'pt-BR', name: 'Portuguese (Brazil)', nativeName: 'Português (Brasil)' },
+  { tag: 'pt-PT', name: 'Portuguese (Portugal)', nativeName: 'Português (Portugal)' },
```
gnprice, line 63: Hmm, something seems wrong when we have these three choices all next to each other.

It looks like the state of things:
https://www.transifex.com/zulip/zulip/mobile/
is that the pt-BR translation is quite fragmentary, but the pt and pt-PT translations are much more complete: 11% vs 89% and 64%. So I think the simplest reasonable thing to do for now is to leave the fragmentary one out of it. I'd hope that the pt translation is actually Brazilian Portuguese.

Some previous discussion here of how we might straighten out this area of our translations:
https://chat.zulip.org/#narrow/stream/58-translation/topic/language.20cleanup

docs/changelog.md
```diff
@@ -36,6 +36,30 @@ It doesn't include
...
 
+### Highlights for users
+
+* (Android) You can now share content from other apps to Zulip. (#117)
+* Fixed a bug where network or server issues could cause an infinite
+  full-screen loading spinner. It now times out after 60 seconds. (#4165)
+* New setting to not mark messages as read when you view them, which is
+  useful for certain workflows. (#4850)
+* Basic support for polls. More to come! (#3205)
+
```
gnprice, line 47: nit: two blank lines between sections

gnprice (commented):
Thanks for the revisions! Looks good except two comments below -- I'll fix those and merge. (One is tiny, and the other just means dropping one commit.)

chrisbobbe:
Thanks for the review!

gnprice:
> I haven't run `tools/tx-sync` after these tweaks to the translation logic; should I do that?

In general I think the mode that `tools/tx-sync` works best in is to be more or less synchronous -- after running it, you push the results to the Git repo soon afterward. That's because it has side effects in Transifex, through the `tx push` it runs.

It also should be the case that any commits it makes are pretty much guaranteed to be OK: they're just syncing down what's in Transifex.

So I think the right strategy when using it is probably:
* go ahead and push directly the commits it makes, without a PR;
* if there are associated changes you make, like adding new languages or changing how some of the system works, that feel like they should go in a PR, then separate those and send them in a PR after pushing the automated sync commits.

Does that make sense? We should probably say something about that in the howto/translations doc :slightly_smiling_face:

chrisbobbe:
> Does that make sense?

Yes, I think so, thanks.

> then separate those and send them in a PR after pushing the automated sync commits

So, while the `tx-sync` commits should be pushed immediately, there isn't a requirement in general that the `tx-sync`-and-push task is done just before (and in the same push as) making the new release tag, with `tools/bump-version`, right?

Or maybe it's OK to do a `tx-sync`-and-push at any time, but we should be sure to also do one along with `tools/bump-version`?

gnprice:
> Or maybe it's OK to do a `tx-sync`-and-push at any time, but we should be sure to also do one along with `tools/bump-version`?

Yes, exactly.

In fact ideally we should be doing one each time we merge a change that adds or changes message strings. That way the new strings get uploaded to Transifex, so that our translation contributors have an opportunity to translate them.

PR #4829, opened 2021-06-24
i18n: Tag "Oops! Something went wrong." for translation.

WesleyAC:
This string appears in `src/chat/FetchError.js`.

PR #4902, opened 2021-07-16
i18n: Remove a duplicate message key.

chrisbobbe:
I got the following error when I ran `tools/tx-sync`:

```
Step 3: Upload strings to translate...
+ tx --quiet push -s
tx ERROR: Error received from server: Duplicate string key ('Oops! Something went wrong\.') in line 247
tx ERROR: Could not upload source file. You can use --skip to ignore this error and continue the execution.
tx ERROR: Error received from server: Duplicate string key ('Oops! Something went wrong\.') in line 247
```

Looks like maybe a rebase error when I was working on #4754. The
string was already added, in #4829.

### DIFF
--- a/static/translations/messages_en.json
+++ b/static/translations/messages_en.json
@@ -38,7 +38,6 @@
   "Why not start the conversation?": "Why not start the conversation?",
   "That conversation doesn't seem to exist.": "That conversation doesn't seem to exist.",
   "Request timed out.": "Request timed out.",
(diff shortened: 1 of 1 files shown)


chrisbobbe:
P0 because this is an easy fix for a problem that's blocking a release. I guess I also don't know what consequences this duplication might have at runtime; I haven't yet tested. 🙂

gnprice:
Looks good, thanks -- merging.

PR #4905, opened 2021-07-19
[android] Multiple share to zulip / share to Zulip from other apps [part 3]

AkashDhiman:
This PR makes changes to enable users to share multiple files.
Followup for https://github.com/zulip/zulip-mobile/pull/4757

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
+++ b/android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
@@ -25,13 +25,12 @@ fun maybeHandleIntent(
     contentResolver: ContentResolver
 ): Boolean {
     // We handle intents from "sharing" something to Zulip.
-    if (intent?.action == Intent.ACTION_SEND) {
-        handleSend(intent, application, contentResolver)
-        return true
+    when (intent?.action) {
+        Intent.ACTION_SEND, Intent.ACTION_SEND_MULTIPLE -> {
+            handleSend(intent, application, contentResolver)
+            return true
+        }
     }
-    // TODO also handle ACTION_SEND_MULTIPLE?
-    //   See: https://developer.android.com/training/sharing/receive#receiving-data-activity
-
     // For other intents, let RN handle it.  In particular this is
     // important for VIEW intents with zulip: URLs.
     return false
@@ -63,6 +62,14 @@ private fun handleSend(intent: Intent, application: ReactApplication, contentRes
     }
 }
 
+private fun urlToSharedFile(url: Uri, contentResolver: ContentResolver): WritableMap {
+    val file = Arguments.createMap()
+    file.putString("name", getFileName(url, contentResolver))
+    file.putString("mimeType", contentResolver.getType(url)?: "application/octet-stream")
+    file.putString("url", url.toString())
+    return file
+}
+
 private fun getParamsFromIntent(intent: Intent, contentResolver: ContentResolver): WritableMap {
     // For documentation of what fields to expect in the Intent, see:
     //   https://developer.android.com/reference/android/content/Intent#ACTION_SEND
@@ -80,14 +87,24 @@ private fun getParamsFromIntent(intent: Intent, contentResolver: ContentResolver
         params.putString("type", "text")
         params.putString("sharedText", sharedText)
     } else {
-        val file = Arguments.createMap()
-        val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
-            ?: throw ShareParamsParseException("Could not extract URL from File Intent")
-        file.putString("name", getFileName(url, contentResolver))
-        file.putString("mimeType", contentResolver.getType(url)?: "application/octet-stream")
-        file.putString("url", url.toString())
+        val files = Arguments.createArray()
+        when (intent.action) {
+            Intent.ACTION_SEND -> {
+                val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
+                    ?: throw ShareParamsParseException("Could not extract URL from File Intent")
+                val file = urlToSharedFile(url, contentResolver)
+                files.pushMap(file)
+            }
+            Intent.ACTION_SEND_MULTIPLE -> {
+                val urls = intent.getParcelableArrayListExtra<Uri>(Intent.EXTRA_STREAM)
+                    ?: throw ShareParamsParseException("Could not extract URLs from File Intent")
+                for (url in urls) {
+                    files.pushMap(urlToSharedFile(url, contentResolver))
+                }
+            }
+        }
         params.putString("type", "file")
-        params.putMap("file", file)
+        params.putArray("files", files)
     }
     return params
 }
--- a/android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
+++ b/android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
@@ -75,26 +75,19 @@ private fun getParamsFromIntent(intent: Intent, contentResolver: ContentResolver
     //
     // it corresponds with `src/types.js#SharedData`.
     val params = Arguments.createMap()
-    when {
-        "text/plain" == intent.type -> {
-            val sharedText = intent.getStringExtra(Intent.EXTRA_TEXT)
-            params.putString("type", "text")
-            params.putString("sharedText", sharedText)
-        }
-        intent.type?.startsWith("image/") == true -> {
-            val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
-                ?: throw ShareParamsParseException("Could not extract URL from Image Intent")
-            params.putString("type", "image")
-            params.putString("sharedImageUrl", url.toString())
-            params.putString("fileName", getFileName(url, contentResolver))
-        }
-        else -> {
-            val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
-                ?: throw ShareParamsParseException("Could not extract URL from File Intent")
-            params.putString("type", "file")
-            params.putString("sharedFileUrl", url.toString())
-            params.putString("fileName", getFileName(url, contentResolver))
-        }
+    if ("text/plain" == intent.type) {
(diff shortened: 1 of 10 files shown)


AkashDhiman:
@gnprice here is the PR to multi-share to zulip.

src/types.js
```diff
@@ -393,19 +393,18 @@ export type SharedText = {|
...
 |};
 
-export type SharedImage = {|
-  type: 'image',
-  sharedImageUrl: string,
-  fileName: string,
+export type File = {|
+  name: string,
+  mimeType: string,
+  url: string,
```
gnprice, line 399: This is a very general name, and it's here in the global `types.js` so it has a very general scope. But the real meaning isn't so general -- this is really specific to having a file shared with us (and for that matter specific to Android's API for that; I'm not sure what it looks like on iOS, and we'll figure that out when we add the feature there.)

`SharedFile` would be a good name, except that it's taken (and even at the tip of the branch, it'd be confusing to have it look so parallel to `SharedText` when it isn't.)

I may try doing a quick refactor of these types and pushing that to the PR branch.

android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
```diff
@@ -25,16 +25,16 @@ fun maybeHandleIntent(
     contentResolver: ContentResolver
 ): Boolean {
     // We handle intents from "sharing" something to Zulip.
-    if (intent?.action == Intent.ACTION_SEND) {
-        handleSend(intent, application, contentResolver)
-        return true
+
+    when (intent?.action) {
+        Intent.ACTION_SEND -> handleSend(intent, application, contentResolver)
+        Intent.ACTION_SEND_MULTIPLE -> handleSend(intent, application, contentResolver)
```
gnprice, line 31:
```suggestion
        Intent.ACTION_SEND, Intent.ACTION_SEND_MULTIPLE ->
            handleSend(intent, application, contentResolver)
```

This makes it a bit clearer that these bodies are indeed the same.

android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
```diff
@@ -25,16 +25,16 @@ fun maybeHandleIntent(
...
+        // For other intents, let RN handle it.  In particular this is
+        // important for VIEW intents with zulip: URLs.
+        else -> return false
```
gnprice, line 34: The cases we handle should continue to return true, as the doc comment says we will.

gnprice: ... Oh I see, it does that below.

That's confusing -- at least, evidently I was confused :wink:. Better to pair the handling of the intent with the returning of true to say we did so.

android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
```diff
@@ -75,26 +83,30 @@ private fun getParamsFromIntent(intent: Intent, contentResolver: ContentResolver
...
+        when (intent.action) {
+            Intent.ACTION_SEND -> {
+                val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
+                    ?: throw ShareParamsParseException("Could not extract URL from File Intent")
+                val file = urlToSharedFile(url, contentResolver)
+                files.pushMap(file)
+            }
+            Intent.ACTION_SEND_MULTIPLE -> {
+                val urls = intent.getParcelableArrayListExtra<Uri>(Intent.EXTRA_STREAM)
+                    ?: throw ShareParamsParseException("Could not extract URL from File Intent")
```
gnprice, line 101: Let's make this message different from the other line above -- it's easy, and gives us a better chance of working out what's happening if we start spotting these in Sentry.

Saying "URLs", plural, would be enough.

android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
```diff
@@ -75,26 +83,30 @@ private fun getParamsFromIntent(intent: Intent, contentResolver: ContentResolver
...
+        params.putString("type", "text")
+        params.putString("sharedText", sharedText)
+    } else {
+        val files = Arguments.createArray()
+        when (intent.action) {
+            Intent.ACTION_SEND -> {
+                val url = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
+                    ?: throw ShareParamsParseException("Could not extract URL from File Intent")
+                val file = urlToSharedFile(url, contentResolver)
+                files.pushMap(file)
```
gnprice, line 97: Could tighten a bit by inlining `file` here -- the resulting line is only slightly longer.

src/sharing/ShareWrapper.js
```diff
@@ -178,19 +210,13 @@ class ShareWrapper extends React.Component<Props, State> {
...
             {sharedData.type === 'file' && (
-              <View style={styles.imagePreview}>
-                <IconAttachment size={200} color={BRAND_COLOR} />
-                <Text style={styles.previewText}>{sharedData.fileName}</Text>
-              </View>
+              <FlatList
+                data={this.state.files}
+                renderItem={this.renderItem}
+                keyExtractor={i => i.url}
+                horizontal
```
gnprice, line 218: Horizontal sounds good here. Does it scroll if the previews don't fit in the width of the screen?

src/sharing/ShareWrapper.js
```diff
@@ -170,6 +172,36 @@ class ShareWrapper extends React.Component<Props, State> {
...
+      this.onShareCancelled();
+    }
+  };
+
+  renderItem = ({ item, index, separators }) => (
+    <ComponentWithOverlay
+      overlaySize={20}
+      overlayColor="white"
+      overlayPosition="bottom-right"
+      overlay={<IconCancel color="gray" size={20} onPress={() => this.deleteItem(item)} />}
```
gnprice, line 190: Buttons should be at least size 24. (Do we have this in, or linked from, our style guide somewhere? If not, we should.)

src/sharing/ShareWrapper.js
```diff
@@ -170,6 +172,36 @@ class ShareWrapper extends React.Component<Props, State> {
     }
   };
 
+  deleteItem = toDelete => {
+    const files = [...this.state.files];
+    const filteredItems = files.filter(item => item.url !== toDelete.url);
+    this.setState({ files: filteredItems });
```
gnprice, line 178: This should get merged into a `setState` call with a callback:
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

src/sharing/ShareWrapper.js
```diff
@@ -170,6 +172,36 @@ class ShareWrapper extends React.Component<Props, State> {
...
+    if (filteredItems.length === 0) {
+      showToast('Cancelled Share');
+      this.onShareCancelled();
+    }
```
gnprice, line 182: And then I guess this needs to go inside that `setState` callback too.

The length check should then also check that the old length was nonzero -- so if these get repeated for some reason, we only do the cancel once.

And the `showToast` and `onShareCancelled` should probably get put inside a `setTimeout` -- so they don't happen synchronously inside the state update. Maybe that'd be fine, but it feels like the sort of thing that might be glitchy.

gnprice (commented):
Thanks @AkashDhiman ! This generally looks great -- some comments below.

I'll take a crack at those `SharedData` types and see what I can come up with that I like.

gnprice:
OK, just added some prep commits to clean up the `SharedData` type, and pushed a revision with your changes rebased on top of those. Please take a look! Now back to you for the other comments.

(I also marked a couple of the commits as needing commit-message updates; I resolved code conflicts in the rebase, but didn't update the commit messages.)

src/types.js
gnprice: (done -- see below in main thread)

src/sharing/ShareWrapper.js
AkashDhiman: Yes.

I just tested this, by temporarily setting the image width > my screen width, since images have fixed size of 200 x 200 which is smaller than most practical device widths and single image scrolls if its width exceeds screen width.

src/sharing/ShareWrapper.js
AkashDhiman: changed overlaySize and `IconCancel`'s size to 24.

I couldn't find a style guide that said 24, and referred to the size from similar code in `AvatarItem` which is 20. Should that be changed as well?

AkashDhiman:
@gnprice made the changes you requested and replied to your questions above, you may take a look.

src/sharing/ShareWrapper.js
gnprice: Great. What if you have several images, so that each one fits but combined they don't?

AkashDhiman: Then the flatlist scrolls horizontally

src/sharing/ShareWrapper.js
gnprice: Hmm, good question.

So, actually the most important aspect of this is that the *touch target* should be at least 48px, to make it accessible for people to accurately hit:
https://material.io/design/usability/accessibility.html#layout-and-typography
(I just went and looked that up, and it reminded me that the needed touch-target size is that big.)

Looking at `AvatarItem`, the whole avatar is the touch target, and it's 50px across. (You can see this in the "Group PM" / "New group PM" screen: select some users, then touch anywhere on their avatar to remove them.) So we're covered there. That also means that the icon isn't really functioning as a button: it's more like a label on the whole avatar as a button, indicating what touching it will do.

Here, I think making the whole 200px preview a touch target for removing it would be counterintuitive. But perhaps we can make a whole 48px square in the corner around the "cancel" icon be the touch target?

gnprice:
Thanks for the revision! I have a couple of comments above about the UI, but this is definitely a major improvement to the UI already. So I'm merging, and those can be followups.

I made one commit-message tweak:
```
    -    sharing types [nfc]: Update `SharedData`.
    +    sharing types [nfc]: Unify SharedData for 'image' and 'file'.
```
just to make the summary line somewhat more informative about what's happening.

PR #4906, opened 2021-07-19
Use `useSelector` instead of `connect` in more places.

chrisbobbe:
Following #4876, toward issue #4837, here are some more that I did today.

### DIFF
--- a/src/autocomplete/TopicAutocomplete.js
+++ b/src/autocomplete/TopicAutocomplete.js
@@ -1,6 +1,6 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
+import React, { useEffect } from 'react';
 import { FlatList } from 'react-native';
 
 import type { Dispatch, Narrow } from '../types';
@@ -31,9 +31,10 @@ type Props = $ReadOnly<{|
   ...SelectorProps,
 |}>;
 
-class TopicAutocomplete extends PureComponent<Props> {
-  componentDidMount() {
-    const { dispatch, narrow } = this.props;
+function TopicAutocomplete(props: Props) {
+  const { dispatch, narrow, isFocused, topics, text, onAutocomplete } = props;
+
+  useEffect(() => {
     // The following should be sufficient to ensure we're up-to-date
     // with the complete list of topics at all times that we need to
     // be:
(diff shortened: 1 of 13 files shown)


src/start/IosCompliantAppleAuthButton/index.js
```diff
@@ -45,40 +32,35 @@ type State = $ReadOnly<{|
...
-          cornerRadius={22}
-          style={[{ height: 44 }, style]}
-          onPress={onPress}
-        />
-      );
-    } else {
-      return <Custom style={style} onPress={onPress} theme={theme} />;
+  useEffect(() => {
+    async function getAndSetIsAvailable() {
+      setIsNativeButtonAvailable(await AppleAuthentication.isAvailableAsync());
```
gnprice, line 42: Hmm, this nesting is kind of odd-looking.

Google tells me that if you just make the `useEffect` callback an async function, you get a warning:
https://stackoverflow.com/questions/53332321/react-hook-warnings-for-async-function-in-useeffect-useeffect-function-must-ret
so I guess that's why you didn't do that. And this thread is informative:
https://github.com/facebook/react/issues/14326

In general this can be a bad pattern, but it's fine here because this always asks the same question, and the answer can't change in the lifetime of the component -- we can't end up with data that's mismatched or out of sync with the rest of the component.

I guess if/when React delivers "suspense", this will be a spot that can switch to that.

For now: maybe just a link to that React issue thread, to explain this odd-looking pattern?

gnprice (commented):
LGTM, thanks! One comment below, just asking for a quick comment; then please merge at will.

gnprice:
... Ah, that was on reading the first 10 commits; looking at the rest now :slightly_smiling_face:

chrisbobbe:
Ah, looks like your review crossed with my force-push! Thanks for the review; there's more, now, though. 😛

src/autocomplete/TopicAutocomplete.js
```diff
@@ -17,23 +17,19 @@ const styles = createStyleSheet({
...
   isFocused: boolean,
   text: string,
   onAutocomplete: (name: string) => void,
-
-  dispatch: Dispatch,
-  ...SelectorProps,
 |}>;
 
-class TopicAutocomplete extends PureComponent<Props> {
-  componentDidMount() {
```
gnprice, line 35: I think this conversion is not actually NFC -- in fact I think it fixes a bug!

Namely, in the existing code we do a `fetchTopicsForStream(narrow)` here when the component first mounts. If `narrow` changes, we won't fetch topics again for the new stream.

Now, when `TopicAutocomplete` appears in `ComposeBox`, this is only a latent bug: `narrow` is the narrow we're viewing, and that only changes by navigation, making new components.

But it also appears in `ShareToStream`. I believe this means we had a visible bug there, where if you
* focused the topic input,
* then edited the stream input,
* and focused the topic input again,

we'd never go and fetch topics for the new stream, and so we'd populate the autocomplete only with whatever topics we happened to already know about for that stream.

With this change, we make a new fetch when the narrow changes. Because the fetch goes to Redux, that doesn't introduce any issue of being out of sync by the time the fetch returns. And the magic of `useEffect` and its explicit dependencies (and ESLint's ability to check the dependencies for completeness) means that it'd actually be rather harder now to write this code *with* this sort of bug than without it.

This commit is otherwise simple enough that I don't think there's a need for a separate commit for this fix -- just the commit message should reflect that it's this bugfix.

gnprice (commented):
OK, and the remaining commits look good too. Just one comment below -- which is that I think one of these commits labeled NFC actually fixes a bug! So the only edit to make is to its commit message.

src/autocomplete/TopicAutocomplete.js
chrisbobbe: Great! 😄

chrisbobbe:
Thanks for the reviews! Merged, with those small tweaks.

Issue #4907, opened 2021-07-19
Flow: Switch to new "Types-First" mode.


chrisbobbe:
We generally keep our Flow version at the same version as the React Native project uses at the time of the RN release we're using. That's because different Flow versions tend to flag different errors and warnings; this happens naturally as Flow fixes bugs and adds/improves features.

- This means we should aim for Flow v0.137 in the upcoming React Native v0.64 upgrade ([link](https://github.com/facebook/react-native/blob/v0.64.2/template/_flowconfig#L66)); that's #4426.
- RN v0.65 hasn't been released yet (and we don't have an issue for it yet). But from a release candidate, it looks like we should expect to aim for v0.149 or later ([link](https://github.com/facebook/react-native/blob/v0.65.0-rc.2/template/_flowconfig#L63)).

From the [doc](https://flow.org/en/docs/lang/types-first/), it looks like Types-First was released as experimental in v0.102 and released officially in v0.125. It'll be enabled by default starting in v0.134, and that's why we'll have to either switch to it before the RN v0.64 upgrade, or temporarily opt out with `types_first=false`.

From the [Flow changelog](https://github.com/facebook/flow/blob/master/Changelog.md#01430), I see that v0.143 removed the ability to opt out:

> Support for Classic mode has been dropped and Types-First mode is now always enabled (Types-First has been the default mode since v0.134). The `types_first` and `well_formed_exports` flowconfig options are no longer recognized. See https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb/ for more about Types-First mode.

So, to keep up with React Native, we'll want to have Types-First enabled by the time we upgrade to RN v0.65, at the latest, since we'll want to aim for Flow v0.149 for that upgrade.

As I mentioned at https://github.com/zulip/zulip-mobile/issues/4426#issuecomment-767044329:

> The thing we'll have to do to enable it is annotate almost all the exports of our modules:
> 
> > The caveat of this new version is that it requires exported parts of the code to be annotated with types, or to be expressions whose type can be trivially inferred (for example numbers and strings).
> 
> Not annotating an export sufficiently will raise a `signature-verification-failure` error.

chrisbobbe:
OK, I've just made another pass through the errors that appear if I temporarily set `well_formed_exports=true`. A few exports stood out to me as particularly challenging to annotate, and I'll probably start discussions for on CZO once we take care of the easier things:

[...]

PR #4908, opened 2021-07-19
Increase module export annotations, where very easy.

chrisbobbe:
Toward #4907. With `well_formed_exports=true` temporarily added in the `.flowconfig` under `[options]` (as suggested in the [doc](https://flow.org/en/docs/lang/types-first/)), this cuts the number of errors approximately in half; it's 396 to 205.


gnprice:
Looks good, thanks! Merging.

Issue #4909, opened 2021-07-20
Starred-messages view should show even topic- or stream-muted messages


gnprice:
This is much like #3472, but for starred messages. I came across it while working on this code, to clean it up and to fix #3472, after looking closer at it as a result of reviewing and merging #4807.

Possibly we should *also* let starred messages break through stream and topic mutes in other views, like @-mentions do. But the web app currently doesn't, so that'd start with a chat thread in #general or #issues or something.

PR #4910, opened 2021-07-20
Fix muted @-mentions and starred messages; clean up muting logic

gnprice:
After merging #4807 I went about doing the cleanup and comment-improvements I mentioned at https://github.com/zulip/zulip-mobile/pull/4807#discussion_r672647504 , and some cleanups to the tests… and kept going, ending up fixing #3472 plus finding and fixing #4909.

Over at #3472 when it was filed a couple of years ago, I'd recommended basing the fix on the web implementation. Here's the key piece of that for topic mutes:
https://github.com/zulip/zulip/blob/89724e5bb4fcbe3aa68477297cfcfc58760f5a77/static/js/message_list_data.js#L168-L181
together with:
https://github.com/zulip/zulip/blob/89724e5bb4fcbe3aa68477297cfcfc58760f5a77/static/js/narrow_state.js#L334-L341
For stream mutes, the implementation seems to be more indirect and I haven't tracked it down.

Seeing that code, I don't exactly want to follow all its structure, and didn't here. But it does provide a good basis for comparison that we're covering important cases.

Fixes: #3472
Fixes: #4909


src/chat/narrowsSelectors.js
```diff
@@ -164,7 +144,35 @@ export const getShownMessagesForNarrow: Selector<$ReadOnlyArray<Message | Outbox
         return messagesForNarrow;
       }
 
-      return messagesForNarrow.filter(item => !shouldBeMuted(item, narrow, subscriptions, mute));
+      // TODO: What about starred messages, and the starred-messages view?
+      //   Do we really want to filter starred messages out because the
+      //   stream or topic is muted?
```
chrisbobbe: Nit:

```diff
      // TODO: What about starred messages, and the starred-messages view?
      //   Do we really want to filter starred messages out because the
-     //   stream or topic is muted?
+     //   topic is muted?
```

It looks like the code at this point is just filtering out starred messages when the topic is muted, without checking the stream.

(I see this is made irrelevant later in the series, with the fix for the starred narrow, though.)

chrisbobbe (approved):
Thanks, @gnprice! This LGTM except for a few small comments below; please merge at will after addressing those. 🙂 (Or wait for @WesleyAC's review, if you'd like, since this is a followup to their #4807.)

src/chat/__tests__/narrowsSelectors-test.js
```diff
@@ -96,86 +96,99 @@ describe('getMessagesForNarrow', () => {
   });
 });
 
-describe('shouldBeMuted', () => {
+/* eslint-disable no-shadow */
```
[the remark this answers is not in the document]
gnprice: Ah indeed, thanks.

src/chat/narrowsSelectors.js
gnprice: Ah, yeah.

Filtering on account of stream *or* topic would be the thing that makes any sense, if we were going to filter for either of them here. That way it would be parallel to what we do in the all-messages narrow.

But this change was written to be NFC, so it reflects what the old code was doing before this branch. I guess it turns out that meant filtering the starred-messages narrow for muted topics only, but I didn't notice that when writing this comment and instead imagined it was doing the arguably-reasonable thing.

gnprice: Changed it to say "because the topic (or stream) is muted". In the context of asking this question, the stream case is still relevant -- the question is about hypothetical behaviors we could choose, and if the answer were "yes" it'd probably be "yes" to both. But hopefully that's less confusing in conjunction with actual behavior that's just about the topics.

gnprice:
Thanks @chrisbobbe for the review! Pushed a revision.

@WesleyAC, it'd be great if you can take a look too, but if not then I can go ahead and merge.

chrisbobbe:
I've just rebased and fixed a small Prettier thing that CI caught.

gnprice:
Thanks for that fix! And I've just merged.

PR #4915, opened 2021-07-21
i18n [nfc]: Document translation-sync process more.

chrisbobbe:
See
  https://github.com/zulip/zulip-mobile/pull/4901#issuecomment-883697861
for what prompted this.


gnprice:
Thanks for writing this down! Merging, with some additional changes.

PR #4921, opened 2021-07-22
android: remove default textColor of #0 on Android.

AkashDhiman:
This was introduced in 8919a6aeb, to achieve consistancy with iOS.
With the changes made in 4db472a2e, this becomes the cause of an
inconsistancy where setting system dark mode in android, makes the
text colors black within a dark colored alert dialog box, giving a
poor color contrast and overall bad look to the dialog box.

This commit makes the color white on dark background and black on
light background, based on the theme.

Similar issue on upstream:
https://github.com/facebook/react-native/issues/31345

Discussion on CZO:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Dark.20theme.20issue.20on.20Android

## Before
WhatsApp-Image-2021-07-15-at-6 16 17-AM

## After (Dark)
WhatsApp-Image-2021-07-15-at-7 06 10-AM

## After (Light)
WhatsApp-Image-2021-07-15-at-7 06 26-AM


AkashDhiman:
@chrisbobbe you may take a look at this.

chrisbobbe:
Thanks! Merged, with a small tweak: I changed the CZO link so it points to a specific message instead of the whole topic, since we don't know what future discussion might happen in the topic. 🙂

PR #4926, opened 2021-07-28
Finish removing `connect` calls for exported components, where it's easy.

chrisbobbe:
Obviously the threshold for "easy" isn't set in stone, but for types-first (#4907) I think I'd like to try leaving the remaining exported `connect`ed components (after this PR) intact for now, and just annotate them following Greg's suggestion [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20types-first/near/1237927) of a straightforward way to do that. Looks like there will be about a dozen of those.

I guess, for #4837, we're not ready to mark it as fixed yet because the title says "all our component definitions". Ah, well, at least we'll have processed the low-hanging fruit. 🙂


src/topics/TopicListScreen.js
```diff
@@ -1,67 +1,52 @@
...
+
+  const handlePress = useCallback(
+    (streamObj: string, topic: string) => {
+      // TODO: use Suspense, when available, in case of race conditions when
+      // fetching on changes to `stream`:
+      //   https://github.com/zulip/zulip-mobile/pull/4906#discussion_r673557179
+      //   https://github.com/facebook/react/issues/14326#issuecomment-441680293
+      // Currently `stream` doesn't change, so this ends up being OK.
+      dispatch(doNarrow(topicNarrow(stream.name, topic)));
+    },
```
gnprice, line 36: Hmm, I think this is actually fine.

The concern from those threads would apply here if we were taking `stream`, and then doing something with it that can take time, like fetching some data… and then we took the result and applied it to what we show in this component, but by the time we do so, `stream` may have changed so that the data we got may be an answer to an out-of-date question. We could be showing a new input value, and the output for an old input value, at the same time, so that what's on the screen is inconsistent and misleading.

But here we're not doing that. We take `stream`, and immediately use it to go initiate a navigation, and that's it.

When we do that, the important thing is just that the `stream` we use is the one the user intended -- which basically means the one that was used for rendering the UI that the user acted on.

(There is a gap in that logic, which is that the value the user actually intended to refer to is the one that was used to render the UI a few frames ago, like 30-100ms, because it takes the human eye and brain and finger that long to act. There isn't a perfect general solution to that issue, but it's an excellent reason to try to avoid jerking the UI around more than necessary, and especially right after first showing some UI, when the user may be most likely to try to interact with it.)

Anyway, so we want the `stream` that was used in rendering the UI the user acted on.

And that is exactly the `stream` that we had when we last rendered this component, because that's what gets passed down to `TopicList`. So using `stream` here is correct.

gnprice: One angle on the data-fetching race described at that upstream React thread is that we're basically completely avoiding it here by an indirection through Redux:

* The UI doesn't get the data from this component's state (which would be for a particular stream, opening up the possibility of a race where what stream that should be changes and the data doesn't get updated in sync). Instead it gets it from the global Redux state, where it looks up the data corresponding to the desired stream.
* When we fetch, we don't fetch the data into this component's state, or anything else that's about some particular stream that might be identified implicitly somewhere else. Instead we fetch it into Redux, storing it filed under whatever stream we fetched data for.

Because the data is stored in a place (our global Redux store) where it's always keyed by the stream the data corresponds to, and because the UI is always told just the stream and goes and looks up the corresponding data there, we avoid the possibility of this kind of race where the data shown in the UI isn't in sync with which stream it's supposed to be about.

gnprice (commented):
Thanks @chrisbobbe! I've read through the bulk of these, and am about to go AFK for dinner so leaving a review now with what I have.

In these first two commits:

4e9913734 flow: Annotate some exported React components, where very easy.
6c480599e userHelpers types [nfc]: Annotate return type of `getUsersAndWildcards`.

there are some spots where I'd like to avoid duplicating the annotations, similar to #4948.

Then in all these component conversions:

fb27fd289 AccountDetails [nfc]: Convert to function component.
0c3527d2c AccountDetails [nfc]: Use `useSelector` instead of `connect`.
201e14e14 AccountDetailsScreen [nfc]: Convert to function component.
a00d743a2 AccountDetailsScreen [nfc]: Use `useSelector` instead of `connect`.
4eb6efdc7 ZulipStatusBar [nfc]: Convert to function component.
d85898a77 ZulipStatusBar [nfc]: Use `useSelector` instead of `connect`.
18be9b30d Emoji [nfc]: Pull out `componentStyles`.
3658bf659 Emoji [nfc]: Convert to function component.
59c1effc4 Emoji [nfc]: Use `useSelector` instead of `connect`.
5888e8894 EmojiPickerScreen [nfc]: Convert to function component.
9932ecbaa EmojiPickerScreen [nfc]: Use `useSelector` instead of `connect`.
f7ccafe5b MentionedUserNotSubscribed [nfc]: Convert to function component.
af06902fa MentionedUserNotSubscribed [nfc]: Use `useSelector` instead of `connect`.
828da6227 NotSubscribed [nfc]: Convert to function component.
88953bc56 NotSubscribed [nfc]: Use `useSelector` instead of `connect`.
f5dbb4871 ZulipNavigationContainer [nfc]: Convert to function component.
6b530d5a1 ZulipNavigationContainer [nfc]: Use `useSelector` instead of `connect`.
b79da02f7 InviteUsersScreen [nfc]: Convert to function component.
76ed1273a InviteUsersScreen [nfc]: Use `useSelector` instead of `connect`.
62e3b40da InviteUsersScreen [nfc]: Inline `handleFilterChange`.
35b9583e1 StreamSettingsScreen [nfc]: Convert to function component.
749b1aa34 StreamSettingsScreen [nfc]: Use `useSelector` instead of `connect`.
0b71c2fe3 SubscriptionsCard [nfc]: Convert to function component.
dc21f8886 SubscriptionsCard [nfc]: Use `useSelector` instead of `connect`.
c9ff859ad StreamListCard [nfc]: Convert to function component.
29c4988ff StreamListCard [nfc]: Use `useSelector` instead of `connect`.
a7f9c7e53 ActivityText [nfc]: Convert to function component.
8829d4114 ActivityText [nfc]: Use `useSelector` instead of `connect`.
ab4dde568 TopicListScreen [nfc]: Add a TODO that we'll address soon.
1f9507f93 TopicListScreen [nfc]: Convert to function component.
45f4c4efa TopicListScreen [nfc]: Use `useSelector` instead of `connect`.
14f4af783 TopicListScreen [nfc]: Inline `handleFilterChange`.

just a couple of small comments below. Please feel free to go ahead and merge those changes (i.e. the branch up to this point, other than the first two commits), modulo those comments.

And the last few commits remain for me to look at later:

9f0d53884 UserStatusScreen [nfc]: Convert to function component.
18071d3fa UserStatusScreen [nfc]: Use `useSelector` instead of `connect`.
e5c4b36f5 UserStatusScreen [nfc]: Inline `setStatusTextState`.
f8343fe66 UserStatusScreen [nfc]: Improve some variable names.
dca1e0c75 EditStreamScreen [nfc]: Convert to function component.
d6921b736 EditStreamScreen [nfc]: Use `useSelector` instead of `connect`.
869d4676b flow: Annotate some more exported React components.

gnprice:
OK, read those remaining commits and they look good! So just the remarks in my comment above.

chrisbobbe:
Thanks for the review! But:

> In these first two commits:
> 
> [b592e32](https://github.com/zulip/zulip-mobile/commit/b592e32eae01f2e7beeef714af3e4895588bf897) flow: Annotate some exported React components, where very easy.
> [e0e629b](https://github.com/zulip/zulip-mobile/commit/e0e629ba370005923bc116b8723f3c455e96bdcc) userHelpers types [nfc]: Annotate return type of `getUsersAndWildcards`.
> 
> there are some spots where I'd like to avoid duplicating the annotations, similar to #4948.

I'm actually not yet finding these spots. Would you mind looking again?

src/topics/TopicListScreen.js
chrisbobbe: Thanks, that makes sense!

PR #4927, opened 2021-07-29
flow: Annotate some more exports.

chrisbobbe:
Toward #4907.

These are the four items I'd flagged as "tricky." Here, I follow Greg's suggestions for handling them at https://github.com/zulip/zulip-mobile/issues/4907#issuecomment-885992786. As it turns out, some of them are not all that tricky! 🙂

### DIFF
--- a/src/api/apiFetch.js
+++ b/src/api/apiFetch.js
@@ -2,6 +2,7 @@
 import * as Sentry from '@sentry/react-native';
 import type { UrlParams } from '../utils/url';
 import type { Auth } from './transportTypes';
+import type { FixmeUntypedFetchResult } from './apiTypes';
 import { getAuthHeaders } from './transport';
 import { encodeParamsForUrl } from '../utils/url';
 import userAgent from '../utils/userAgent';
@@ -36,7 +37,7 @@ export const getFetchParams = <P: $Diff<$Exact<RequestOptions>, {| headers: mixe
   };
 };
 
-export const apiFetch = async (
+const apiFetch = async (
   auth: Auth,
   route: string,
   params: $Diff<$Exact<RequestOptions>, {| headers: mixed |}>,
@@ -48,7 +49,7 @@ export const apiCall = async (
   route: string,
   params: $Diff<$Exact<RequestOptions>, {| headers: mixed |}>,
   isSilent: boolean = false,
-): Promise<empty> => {
+): Promise<FixmeUntypedFetchResult> => {
   try {
     networkActivityStart(isSilent);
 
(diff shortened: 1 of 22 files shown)


src/api/notifications/forgetPushToken.js
```diff
@@ -7,7 +7,7 @@ import { apiDelete } from '../apiFetch';
  *
  * @param mobileOS - Choose the server-side API intended for iOS or Android clients.
  */
-export default (auth: Auth, mobileOS: 'ios' | 'android', token: string) => {
+export default (auth: Auth, mobileOS: 'ios' | 'android', token: string): Promise<empty> => {
```
gnprice, line 10: And for this one and `savePushToken`, it looks like we don't actually inspect the returned result at all -- so this could just be `Promise<mixed>`.

src/api/checkCompatibility.js
```diff
@@ -3,7 +3,7 @@ import userAgent from '../utils/userAgent';
 
 // check with server if current mobile app is compatible with latest backend
 // compatibility fails only if server responds with 400 (but not with 200 or 404)
-export default () =>
+export default (): Promise<empty> =>
```
gnprice, line 6: This could be `Promise<{ status: number, ... }>`.

src/api/apiFetch.js
```diff
@@ -40,7 +40,8 @@ export const apiFetch = async (
   auth: Auth,
   route: string,
   params: $Diff<$Exact<RequestOptions>, {| headers: mixed |}>,
-) => fetch(new URL(`/${apiVersion}/${route}`, auth.realm).toString(), getFetchParams(auth, params));
+): Promise<empty> =>
```
gnprice, line 43: This one actually doesn't need to be exported at all! So that's the simplest fix for this one.

gnprice (commented):
Thanks! Comments below.

src/account/__tests__/accountsReducer-test.js
```diff
@@ -25,7 +25,7 @@ describe('accountsReducer', () => {
       expect(
         accountsReducer(
           deepFreeze([account1, account2, account3]),
-          deepFreeze({ ...eg.action.realm_init, zulipVersion: newZulipVersion }),
+          deepFreeze({ ...eg.realmInitAction, zulipVersion: newZulipVersion }),
```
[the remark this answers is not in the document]
chrisbobbe: > Is the motivation for these names to put the definitions at the top level of the module, in order to add annotations?

It was. But looking back at your [message](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20types-first/near/1237928) on CZO, I see I misread what you wanted me to try. 😛 I'd thought you were asking for the members of `eg.action` to be exported individually, IOW, to pull them out into their own exported variables. Now I see you were talking about exporting the types themselves, in a different file, like `RealmInitAction`.

chrisbobbe:

gnprice:

PR #4931, opened 2021-07-30
Stop using Lolex, since we're using Jest's "modern" fake timers.

chrisbobbe:
Planned followup to #4754.


gnprice:
Thanks! Merging.

chrisbobbe:
Thanks for the review!

PR #4934, opened 2021-08-02
sentry: Remove non-Kandra realm URLs from breadcrumbs.

chrisbobbe:


src/sentry.js
```diff
@@ -45,6 +51,82 @@ const preventNoise = (): void => {
...
+ *
+ * - Removes the realm hostnames in API requests, except for Kandra-hosted
+ *   realms.
+ */
+function scrubBreadcrumb(breadcrumb: Breadcrumb, hint?: BreadcrumbHint): Breadcrumb {
+  switch (breadcrumb.type) {
+    case 'http': {
+      // $FlowIgnore[incompatible-indexer]
+      // $FlowIgnore[incompatible-type]
+      const httpBreadcrumb: HttpBreadcrumb = breadcrumb;
```
gnprice, line 82: Should have a comment explaining why we believe the ignores are OK.

Basically we're assuming that if `type` is `'http'`, then the breadcrumb has the shape of an `HttpBreadcrumb`. That assumption isn't expressed in the libdef, but is in the API docs at the page you link to.

src/sentry.js
```diff
@@ -45,6 +51,82 @@ const preventNoise = (): void => {
...
+            // realms in the account-statuses state, including those without
+            // `/api/v1`, like `/avatar/{user_id}` (zulip/zulip@0f9970fd3
+            // confirms that this and potentially others don't / won't use
+            // `/api/v1`).
+            || accountStatuses.some(({ realm }) => parsedUrl.origin === realm.origin)
+          )
+          && !config.appOwnDomains.some(
+            domain => parsedUrl.host === domain || parsedUrl.host.endsWith(`.${domain}`),
+          )
+        ) {
```
gnprice, line 108: Quite an expression there :grin: 

This could be pulled out as a helper function with a name like `shouldScrubHost`, taking a `URL` and the `accountStatuses` list. Then in that function you can use control flow like `return` instead of some of these boolean operators, which I think will make it simpler.

src/sentry.js
```diff
@@ -45,6 +51,82 @@ const preventNoise = (): void => {
...
+            // Definitely a request to a Zulip realm. Will catch requests to
+            // realms in the account-statuses state, including those without
+            // `/api/v1`, like `/avatar/{user_id}` (zulip/zulip@0f9970fd3
+            // confirms that this and potentially others don't / won't use
+            // `/api/v1`).
+            || accountStatuses.some(({ realm }) => parsedUrl.origin === realm.origin)
+          )
+          && !config.appOwnDomains.some(
+            domain => parsedUrl.host === domain || parsedUrl.host.endsWith(`.${domain}`),
+          )
```
gnprice, line 107: This is doing the same thing as in the other spot where we use `config.appOwnDomains`, right?

Let's pull this out as its own little named helper so that there's only one copy of it, because it is a bit tricky. Not sure there's a great place to group it with other things; so perhaps just its own tiny file `src/isAppOwnDomain.js`.

src/sentry.js
```diff
@@ -45,6 +51,82 @@ const preventNoise = (): void => {
...
+        ) {
+          parsedUrl.host = `zulip.example${
+            // So different realms are still distinguishable
+            md5(parsedUrl.host).substring(0, 6)
+          }.com`;
```
gnprice, line 112: The `example.foo` domains are reserved, but I believe other names with the word "example" as just part of a name component, like `example012345.com`, are not.

So for the cleanest thing here, we should use names that are reserved and guaranteed never to collide with something real.

Saying something like `zulip-${…}.example.com`, or even `zulip-${…}.example` ([it's a TLD!](https://en.wikipedia.org/wiki/.invalid)), would work.

Perhaps even better: use a name with "invalid":
https://datatracker.ietf.org/doc/html/rfc6761#section-6.4
So e.g. `hidden-${…}.zulip.invalid`.

gnprice (commented):
Thanks @chrisbobbe ! This looks good. Some small comments below.

chrisbobbe:

src/sentry.js
```diff
@@ -45,6 +52,86 @@ const preventNoise = (): void => {
...
+ */
+function scrubBreadcrumb(breadcrumb: Breadcrumb, hint?: BreadcrumbHint): Breadcrumb {
+  switch (breadcrumb.type) {
+    case 'http': {
+      // $FlowIgnore[incompatible-indexer] | We assume it's an
+      // $FlowIgnore[incompatible-type]    | HttpBreadcrumb; see jsdoc.
+      const httpBreadcrumb: HttpBreadcrumb = breadcrumb;
+
+      const unscrubbedUrl = httpBreadcrumb.data.url;
+      let scrubbedUrl;
```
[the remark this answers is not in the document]
gnprice: (Also I think actually both of these can be made somewhat cleaner by using control flow with `return`, instead of data flow mutating a `let`-bound variable. Maybe I'll try doing that at or after merging.)

gnprice (commented):
Thanks for the revision! This looks good; small comments below.

chrisbobbe:

gnprice:
Thanks! Looks good -- merging, with the changes I mentioned in the last subthread above.

src/sentry.js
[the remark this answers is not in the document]
gnprice: (Indeed I did; take a look.)

chrisbobbe:
Thanks for the review and those tweaks!

src/sentry.js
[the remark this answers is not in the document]
gnprice: e62493475

PR #4935, opened 2021-08-03
ComposeMenu: Offer file picker on iOS, not just Android.

chrisbobbe:
ComposeMenu: Offer file picker on iOS, not just Android

I've just tested on my iPhone (13 Pro, running iOS 16.1), and I was
able to upload a file just fine after this change. That was easy!

Fixes: #4586

cc @alya for the UX change

(edited for revision on 2023-01-10; screenshots in a later message)


chrisbobbe:
Inheriting P1 from the issue.

gnprice:
(Chat discussion of this / #4586 was [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/PDF.20in.20iOS/near/1243359).)

chrisbobbe:
Revision pushed, and I'm re-adding the P1 label given the renewed interest in this feature.

Here's a screenshot of the feature working (except that the inserted text isn't visible in the compose box; that'll be caused by the `multiline` bug described in #5291 and fixed in PR #5635):

Jan-10-2023 15-00-07

gnprice:
Thanks for pushing on this! The code looks great.



> (except that the inserted text isn't visible in the compose box; that'll be caused by the `multiline` bug first described in #5291 and to be fixed in PR #5635):

Hmm, that seems like a pretty bad symptom -- it's likely to make the user think it didn't work at all. I'll go take a look at #5635 now and see if we can merge that first.

gnprice:
OK, just reviewed #5635. Not quite ready to merge, but I think we'll be able to do so soon. So let's let this PR just hang out for a bit, and we can merge it once that's in.

gnprice:
OK, and #5635 is now merged. Would you rebase this PR and resolve the conflicts?

gnprice:
Thanks! Looks good; merging.

PR #4937, opened 2021-08-03
deps: Remove expo-splash-screen.

chrisbobbe:
Empirically, this is no longer needed: we added it in c4fca9d9d to
satisfy a peer-dependency constraint, but now I'm not getting a
peer-dep warning if I remove it.

See #4279 and discussion at
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/React.20Navigation.20v5/near/1042787
for some warnings from Greg about the design of expo-splash-screen,
should we choose to actually use it one day.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -22,9 +22,6 @@ PODS:
   - EXScreenOrientation (1.1.1):
     - React-Core
     - UMCore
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Closing as superseded by #4949.

PR #4938, opened 2021-08-03
android: Drop support for API < 23.

AkashDhiman:
As Discussed in [CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Updating.20.60minSdkVersion.60/near/1236327), Last time (9-2020) we checked, the total
user base using API < 23 is less than 1 percent and hence it is
safe for us to update the API version (according to the [update
policy](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/platform-versions.md#android-and-ios-versions) [2] we have).

The advantage of this change would be simplification in some of the
code that we use / will be using which will be changed / added in
other commits.

The changes in this commit Involved:
* Changing `android/build.gradle`.
* Setting up chrome version to 44 (set according to what is
  written in `platform-versions.md`).
* Modifications to the documentation.

Marking P1 since Notification UI is built on top of this PR.

### DIFF
--- a/docs/architecture/platform-versions.md
+++ b/docs/architecture/platform-versions.md
@@ -94,6 +94,8 @@ History:
 * We [dropped iOS 11 support] in 2021-04. It was 0.1% of iOS users
   who tried Zulip. We started iOS 12 support at 12.1 because Xcode's
   dropdown for "Deployment Target" didn't have a 12.0.
+* We [dropped Android 5 Lollipop support][dropped-android-l] in
+  2021-08.
 <!-- When updating this, please update `docs/developer-guide.md` as well. -->
 
 [dropped-android-j]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/625585
@@ -101,6 +103,7 @@ History:
 [dropped iOS 9 support]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/771761
 [dropped-android-k]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/794551
 [dropped iOS 11 support]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/1165087
+[dropped-android-l]: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Updating.20.60minSdkVersion.60/near/1236327
 
 
 Related observations:
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // The oldest Android version we support.  Increasing this stops updates
         // for users on old devices, but can simplify the system.  Synced with
         // our dev docs: docs/developer-guide.md
-        minSdkVersion = 21
+        minSdkVersion = 23
 
         // A bit subtle: we should try to keep this at the latest version,
         // but some testing is required and code changes are often required.
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -1,6 +1,6 @@
 # Developer guide
 
-We target operating systems >= Android 5 Lollipop (API 21)
+We target operating systems >= Android 6 Marshmallow (API 23)
 and >= iOS 12.1.  (Details [here](architecture/platform-versions.md).)
 
 ## Why React Native?
--- a/src/webview/js/generatedEs3.js
+++ b/src/webview/js/generatedEs3.js
@@ -434,12 +434,12 @@ var compiledWebviewJs = (function (exports) {
     sendMessage({
       type: 'error',
       details: {
-        message: message,
-        source: source,
-        line: line,
-        column: column,
-        userAgent: userAgent,
-        error: error
+        message,
+        source,
+        line,
+        column,
+        userAgent,
+        error
       }
     });
     return true;
@@ -593,8 +593,8 @@ var compiledWebviewJs = (function (exports) {
     walkElements(start, 'nextElementSibling');
     walkElements(start, 'previousElementSibling');
     return {
-      first: first,
-      last: last
+      first,
+      last
     };
   }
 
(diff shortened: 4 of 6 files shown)


docs/architecture/platform-versions.md
```diff
@@ -94,13 +94,16 @@ History:
...
+* We [dropped Android 5 Lollipop support][dropped-android-l] in
+  2021-08.
 <!-- When updating this, please update `docs/developer-guide.md` as well. -->
 
 [dropped-android-j]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/625585
 [dropped iOS 8 support]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/628412
 [dropped iOS 9 support]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/771761
 [dropped-android-k]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/794551
 [dropped iOS 11 support]: https://chat.zulip.org/#narrow/stream/48-mobile/topic/platform.20versions/near/1165087
+[dropped-android-l]: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Updating.20.60minSdkVersion.60/near/1236327
```
AkashDhiman, line 106: would be nice if someone could move conversations of that topic, to platform versions, to make this consistent with others.
(I think only admins can move conversations)

AkashDhiman:
@gnprice, @chrisbobbe let me know if I missed anything, thank you.

chrisbobbe (changes requested):
Thanks! See one small comment below. Also, I think there's one paragraph that can be deleted from a comment in src/webview/js/js.js 🎉:

```
```

Hmm, and that makes me think we should maybe bump the numbers in the following, a bit above that:

```
 *   Graceful degradation is acceptable below Chrome 74.
```

From a comment in docs/architecture/platform-versions.md, I _think_ that the pattern would be to bump those two numbers to 51 and 83:

```
  * Android versions 5 L, 6 M, 7 N, 8 O had originally shipped
    with Chrome versions 37, 44, 51, 58 respectively (based on
    looking at stock emulator images.)
    * Later data: Android 9, 10, 11 ship with Chrome versions
      69, 74, 83 respectively.
```

Maybe @gnprice could confirm? 🙂

AkashDhiman:
@chrisbobbe did the changes you asked for waiting on @gnprice to comment on
> Hmm, and that makes me think we should maybe bump the numbers in the following, a bit above that:
> 
> ```
>  * * For Android, core functionality needs to work on Chrome 44.
>  *   Graceful degradation is acceptable below Chrome 74.
> ```

docs/architecture/platform-versions.md
gnprice: I don't mind the links not all being to the same topic. In fact it's convenient for the "platform versions" topic to not have much detailed discussion on it, to help make it easy to scan through it for the history of the numbers we've posted there, so I'm happy for the recent discussion to stay on a distinct topic.

It would be good to have a quick message there just recording that we've made this change, linking to the discussion thread and to this PR.

gnprice:
> Hmm, and that makes me think we should maybe bump the numbers in the following, a bit above that:
> 
> ```
>  * * For Android, core functionality needs to work on Chrome 44.
>  *   Graceful degradation is acceptable below Chrome 74.
> ```

Yeah, there's probably a helpful related change to do here along those lines. It doesn't actually have a particular need to be coupled to this minSdkVersion change, though -- this one is about the minimum version of Android that we support, so the data it's based on is the distribution of Android versions our users have, while that one would be about the minimum versions of Chrome that we support to various degrees, so the data it'd be based on is the distribution of Chrome versions our users have. Those distributions are quite different from each other, because Android's Chrome auto-updates independently of the OS.

Before this change, we nominally had three levels of support for different Chrome versions:
* 37 to below 44: sometimes fix issues, when trivial
* 44 to below 74: core functionality needs to work, but graceful degradation acceptable
* 74 and up: fully supported

I'm kind of inclined to drop that lowest tier, because I'm not sure the extra distinction was really buying us that much. If we run into something that's busted only on ancient Chrome versions like 44, we can consider reintroducing it, or perhaps just dropping support for Android 6 M altogether (and consequently for Chrome older than 51.)

The current version of this PR does that. Keeping it that way means leaving the mention of Chrome 44 as is, to be in sync with Android 6 M as our oldest supported Android version.

For updating the "fully supported" / "graceful degradation below" version, it looks like our last update was 92d24025171a672ad2475d877fd8c484c71de89c, earlier this year. Updating it to Chrome 83 would mean updating to a Chrome version released 2020-05 (thank you [Wikipedia](https://en.wikipedia.org/wiki/Google_Chrome_version_history) -- one of the niches Wikipedia tends to do a great job at is chronicling old versions of all kinds of major software). That seems pretty reasonable; that's over a year ago, 9 versions before the current latest, and [our previous look at some data](https://github.com/zulip/zulip-mobile/blob/88763612d4e14cdce01f9f1e63991f82b180643b/docs/architecture/platform-versions.md#data-and-commentary-1) suggests that 90% of users will be on just the two latest versions.

gnprice:
This all looks good; merging. Thanks @AkashDhiman for taking care of this!

In the previous comment, I guess I reached the conclusion that a good followup would be to replace that mention of Chrome 74 with Chrome 83. I don't think that involves changing any other lines.

A separate nice followup would be to remove the handful of spots in the `js.js` source code that are there just for the sake of old browsers we no longer support. Searching the file for "Chrome" finds them. Looks like there's a couple of polyfills, and a conditional inside `midMessagePeer`. Cutting out that conditional will also mean we get to delete a paragraph of the long comment above it.

For cutting those all out, it'd be good to try the app with Chrome 44 (by firing up an emulator with a stock Android 6 M image) and confirm that things seem to work. Also ideally on iOS 12, I guess (for the oldest Safari we support); but iOS 12 is a few years newer than Chrome 44, so even if our notes are mistaken it's a lot less likely that these features would be missing there.

gnprice:
And merged! I made just one change, expanding the summary line:
2e7ea833e android: Drop support for API < 23, aka Android 6 Marshmallow.

I tend to find the API numbers hard to keep track of, so when they appear in text I like to accompany them with the Android version number and letter (or the dessert the letter stands for, though starting with Android 9 P they've de-emphasized the letter and dessert, and starting with Android 10 Q the letter doesn't stand for anything.) Those have consistently advanced by one each year since before the oldest Android versions we care about, so I can always count backward from the latest to orient myself to how old a given number/letter is.

PR #4940, opened 2021-08-03
Improve internet-connectivity state a bit; upgrade @react-native-community/netinfo.

chrisbobbe:
Now, we're specifically using a bit of state that says whether we can expect to reach the Internet. And we show a notice if it's taken unexpectedly long to determine the state (10 seconds).


flow-typed/@react-native-community/netinfo_v5.x.x.js
```diff
@@ -0,0 +1,172 @@
...
+  declare export type NetInfoOtherState = NetInfoConnectedState<typeof NetInfoStateType.other>;
+  declare export type NetInfoConnectedStates =
+    | NetInfoCellularState
+    | NetInfoWifiState
+    | NetInfoBluetoothState
+    | NetInfoEthernetState
+    | NetInfoWimaxState
+    | NetInfoVpnState
+    | NetInfoOtherState;
+  declare export type NetInfoState = NetInfoDisconnectedStates | NetInfoConnectedStates;
```
gnprice, line 100: Neat! This type is pretty detailed.

src/boot/AppEventHandlers.js
```diff
@@ -93,12 +53,52 @@ class AppEventHandlers extends PureComponent<Props> {
...
+        //    native code, signaling a change in the network state. If that
+        //    change would make `netInfoState.type` become or remain
+        //    something good (i.e., not 'none' or 'unknown'), and this
+        //    (`.isInternetReachable`) is currently `false`, then this will
+        //    be made `null`, and the request loop described above will
+        //    start again.
+        //
+        // (Several of those parameters are configurable -- timeout durations,
+        // URL, etc.)
+        netInfoState.isInternetReachable,
```
gnprice, line 99: online state: Switch to `isInternetReachable`, from `connectionType`.
    
    This is hopefully more accurate about what we actually want to know:
    "can the app expect to reach the Internet?" Not just "is there some
    network connection?".

Cool.

One thing happening here is that our old logic was equivalent to using the `isConnected` property (except that with `@react-native-community/netinfo` v6 that property started being `null` sometimes instead of `false`.) So using that would have been a bit cleaner for us to do in the first place, plus with the v6 upgrade it gives us the distinction between `null` and `false`.

Then switching from (effectively) `isConnected` to `isInternetReachable` makes it false or possibly null in some cases where it previously would have been true. That's the result of asking the more on-point question.

src/common/OfflineNotice.js
```diff
@@ -24,19 +26,76 @@ const styles = createStyleSheet({
...
+    if (value) {
+      timeoutId.current = setTimeout(() => {
+        setResult(true);
+      }, duration);
+    } else {
+      clearTimeout(timeoutId.current);
+      timeoutId.current = null;
+      setResult(false);
+    }
+  }, [value, duration]);
```
gnprice, line 43: Hmm. Won't this clobber the old timeout? That is, if we have:
* Render with value true.
* Render again with value true, before that timeout completes.

then we'll forget about the old timeout and set a new one.

I guess the original timeout will still end up firing… but because `timeoutId.current` won't know about it anymore, it'll end up firing even if the value *then* changes to false.

Oh, and I guess this is a latent bug as it stands, because it only triggers if the *effect* gets rerun. Which only happens if `value` or `duration` changes, and currently we never change `duration`. Still, good to fix the latent bug in case we reuse this -- plus I suspect the code will be somewhat easier to reason through in a fixed version.

gnprice: Separately: we should clear this timeout if the component unmounts, so we don't have a leaked timeout trying to belatedly call `setResult`.

`useEffect` supports this -- have its callback return the cleanup function:
https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

gnprice (commented):
Thanks @chrisbobbe ! These changes will be good. Comments below, mainly on the implementation of `useHasStayedTrueForMs`.

src/common/OfflineNotice.js
chrisbobbe: > Oh, and I guess this is a latent bug as it stands, because it only triggers if the _effect_ gets rerun.

Ah, yeah, I think I was sort of counting on this fact for the implementation. I think I should use `usePrevious(value)` in there somewhere; I'll try that.

chrisbobbe:

gnprice (commented):
Thanks! Partial thoughts below.

gnprice:
I've just pushed a revision that takes a bit of a different approach to the issue of `duration` changing in `useHasStayedTrueForMs`: we declare that the caller isn't supposed to do that, plus assert that they don't. (With the assertion skipped in production mode, to avoid adding a `useRef` to track the previous value.) Please take a look!

Also added a couple of tweaks atop the changes to `usePrevious`. All the changes to `usePrevious` end up being moot with this revision, but I think the improvements to it are useful anyway and would be good to have for the future. A demo of how my added commits would be helpful is in 59c2c5985, in a demo branch: we get to have the initial value be `false`, and the resulting type is `boolean` rather than `boolean | null`, simplifying the conditionals on it.

PR #4941, opened 2021-08-03
Simplify "update strategy" logic

gnprice:
As discussed here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/getMessageTransitionProps/near/1243171

This cleans up the structure of this code a bit, hopefully making it easier to work out exactly what it's doing and to change what it's doing.

The actual behavior is unchanged. Some of the conditions don't seem to make a ton of sense -- in particular they seem like they can at best be heuristics for what the names (converted here to comments) suggest they're trying to get at, and it's not at all clear (and there are no comments to explain) why those particular conditions should lead to these particular behaviors. So there's plenty of room to change the behavior from here; in particular, we don't need to give the existing code a lot of deference in being super careful about preserving the existing behavior, as long as we're doing some thinking about the desired new behavior with any changes.

### DIFF
--- a/src/message/__tests__/messageUpdates-test.js
+++ b/src/message/__tests__/messageUpdates-test.js
@@ -1,181 +1,11 @@
 /* @flow strict-local */
 import { streamNarrow } from '../../utils/narrow';
-import { getMessageTransitionProps, getMessageUpdateStrategy } from '../messageUpdates';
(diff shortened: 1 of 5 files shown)


chrisbobbe:
Thanks a lot for this! Merging. 🙂

PR #4942, opened 2021-08-04
MentionWarnings [nfc]: Convert to function component.

chrisbobbe:
For annotating the default export of `MentionWarnings`, for #4907, I think it actually may be easier to first convert it to a function component and remove the `connect` call. This is a little surprising because it had been on my list of components that I thought would be pretty tricky to convert.

I didn't find a way to get thorough type-checking of `MentionWarnings`'s methods that `ComposeBox` was calling on a `ref`, until I converted `MentionWarnings` to a function component. After this PR, we should get that type-checking, without having to manually repeat any types.

I tested the component on my iPhone and it worked as I expected it to.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -138,14 +138,7 @@ class ComposeBox extends PureComponent<Props, State> {
   messageInputRef = React.createRef<$FlowFixMe>();
   topicInputRef = React.createRef<$FlowFixMe>();
 
-  // TODO: Type-check this, once we've adjusted our `react-redux`
-  // wrapper to do the right thing. It should be
-  //
-  //   mentionWarnings = React.createRef<React$ElementRef<MentionWarnings>>()
-  //
-  // but we need our `react-redux` wrapper to be aware of
-  // `{ forwardRef: true }`, since we use that.
-  mentionWarnings = React.createRef();
+  mentionWarnings = React.createRef<React$ElementRef<typeof MentionWarnings>>();
 
   inputBlurTimeoutId: ?TimeoutID = null;
 
@@ -503,10 +496,6 @@ class ComposeBox extends PureComponent<Props, State> {
 
     return (
       <View style={this.styles.wrapper}>
-        {/*
-          $FlowFixMe[incompatible-use]:
-          `MentionWarnings` should use a type-checked `connect`
-        */}
         <MentionWarnings narrow={narrow} stream={stream} ref={this.mentionWarnings} />
         <View style={[this.styles.autocompleteWrapper, { marginBottom: height }]}>
           <TopicAutocomplete
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -19,7 +19,19 @@ type Props = $ReadOnly<{|
 |}>;
 
 type ImperativeHandle = {|
+  /**
+   * Check whether the message text entered by the user contains
+   * an @-mention to a user unsubscribed to the current stream, and if
+   * so, shows a warning.
+   *
+   * This function is expected to be called by `ComposeBox` using a ref
+   * to this component.
+   *
+   * @param completion The autocomplete option chosend by the user.
+      See JSDoc for AutoCompleteView for details.
+   */
   handleMentionSubscribedCheck(completion: string): Promise<void>,
+
   clearMentionWarnings(): void,
 |};
 
@@ -81,17 +93,6 @@ function MentionWarnings(props: Props, ref) {
   useImperativeHandle(
     ref,
     () => ({
-      /**
-       * Check whether the message text entered by the user contains
-       * an @-mention to a user unsubscribed to the current stream, and if
-       * so, shows a warning.
-       *
-       * This function is expected to be called by `ComposeBox` using a ref
-       * to this component.
-       *
-       * @param completion The autocomplete option chosend by the user.
-          See JSDoc for AutoCompleteView for details.
-       */
       handleMentionSubscribedCheck: async (completion: string) => {
         if (is1to1PmNarrow(narrow)) {
           return;
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -1,9 +1,9 @@
 /* @flow strict-local */
 
-import { connect } from 'react-redux';
 import React, { useState, useCallback, useContext, forwardRef, useImperativeHandle } from 'react';
+import { useSelector } from 'react-redux';
 
-import type { Auth, Stream, Dispatch, Narrow, UserOrBot, Subscription, UserId } from '../types';
+import type { Stream, Narrow, UserOrBot, Subscription, UserId } from '../types';
 import { TranslationContext } from '../boot/TranslationProvider';
 import { getAllUsersById, getAuth } from '../selectors';
 import { is1to1PmNarrow } from '../utils/narrow';
@@ -13,17 +13,9 @@ import { showToast } from '../utils/info';
 import MentionedUserNotSubscribed from '../message/MentionedUserNotSubscribed';
 import { makeUserId } from '../api/idTypes';
 
-type SelectorProps = {|
-  auth: Auth,
-  allUsersById: Map<UserId, UserOrBot>,
-|};
-
 type Props = $ReadOnly<{|
   narrow: Narrow,
   stream: Subscription | {| ...Stream, in_home_view: boolean |},
-
-  dispatch: Dispatch,
-  ...SelectorProps,
 |}>;
 
 type ImperativeHandle = {|
@@ -32,7 +24,10 @@ type ImperativeHandle = {|
 |};
 
 function MentionWarnings(props: Props, ref) {
-  const { stream, narrow, auth, allUsersById } = props;
+  const { stream, narrow } = props;
+
+  const auth = useSelector(getAuth);
+  const allUsersById = useSelector(getAllUsersById);
 
   const [unsubscribedMentions, setUnsubscribedMentions] = useState<UserId[]>([]);
 
@@ -169,13 +164,4 @@ function MentionWarnings(props: Props, ref) {
   return mentionWarnings;
 }
 
-// $FlowFixMe[missing-annot]. TODO: Use a type checked connect call.
-export default connect(
-  state => ({
-    auth: getAuth(state),
-    allUsersById: getAllUsersById(state),
-  }),
-  null,
-  null,
-  { forwardRef: true },
-)(forwardRef<Props, ImperativeHandle>(MentionWarnings));
+export default forwardRef<Props, ImperativeHandle>(MentionWarnings);
--- a/src/compose/MentionWarnings.js
+++ b/src/compose/MentionWarnings.js
@@ -1,18 +1,9 @@
 /* @flow strict-local */
 
-import React, { PureComponent } from 'react';
 import { connect } from 'react-redux';
+import React, { useState, useCallback, useContext, forwardRef, useImperativeHandle } from 'react';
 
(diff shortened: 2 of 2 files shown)


src/compose/ComposeBox.js
```diff
@@ -138,13 +138,7 @@ class ComposeBox extends PureComponent<Props, State> {
...
 
-  // TODO: Type-check this, once we've adjusted our `react-redux`
-  // wrapper to do the right thing. It should be
-  //
-  //   mentionWarnings = React.createRef<React$ElementRef<MentionWarnings>>()
-  //
-  // but we need our `react-redux` wrapper to be aware of
-  // `{ forwardRef: true }`, since we use that.
+  // Here, Flow seems to infer the correct type based on usage.
   mentionWarnings = React.createRef();
```
chrisbobbe, line 142: Unfortunately, none of these three worked (see errors alongside):

- `React.createRef<React$ElementRef<MentionWarnings>>()`

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:142:54

Cannot use MentionWarnings as a type. A name can be used as a type only if it refers to a type, interface, class, or
enum definition. To get the type of a non-class value, use typeof. [value-as-type]

     139│   topicInputRef = React.createRef<$FlowFixMe>();
     140│
     141│   // Here, Flow seems to infer the correct type based on usage.
     142│   mentionWarnings = React.createRef<React$ElementRef<MentionWarnings>>();
     143│
     144│   inputBlurTimeoutId: ?TimeoutID = null;
     145│
```

- `React.createRef<MentionWarnings>()` (from https://github.com/zulip/zulip-mobile/pull/4843#discussion_r665701353)

```
Cannot use MentionWarnings as a type. A name can be used as a type only if it refers to a type, interface, class, or
enum definition. To get the type of a non-class value, use typeof. [value-as-type]

     139│   topicInputRef = React.createRef<$FlowFixMe>();
     140│
     141│   // Here, Flow seems to infer the correct type based on usage.
     142│   mentionWarnings = React.createRef<MentionWarnings>();
     143│
     144│   inputBlurTimeoutId: ?TimeoutID = null;
     145│
```

- `React.createRef<typeof MentionWarnings>()`

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:304:37

Cannot call this.mentionWarnings.current?.handleMentionSubscribedCheck because property handleMentionSubscribedCheck is
missing in AbstractComponent [1]. [incompatible-use]

 [1] 142│   mentionWarnings = React.createRef<typeof MentionWarnings>();
        :
     301│     if (lastWordPrefix === '@') {
     302│       // https://github.com/eslint/eslint/issues/11045
     303│       // eslint-disable-next-line no-unused-expressions
     304│       this.mentionWarnings.current?.handleMentionSubscribedCheck(completion);
     305│     }
     306│   };
     307│


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:382:36

Cannot call this.mentionWarnings.current.clearMentionWarnings because property clearMentionWarnings is missing in
AbstractComponent [1]. [incompatible-use]

 [1] 142│   mentionWarnings = React.createRef<typeof MentionWarnings>();
        :
     379│     this.setMessageInputValue('');
     380│
     381│     if (this.mentionWarnings.current) {
     382│       this.mentionWarnings.current.clearMentionWarnings();
     383│     }
     384│
     385│     dispatch(sendTypingStop(narrow));


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:500:63

Cannot create MentionWarnings element because in property ref: [incompatible-type]
 • Either object type [1] is incompatible with AbstractComponent [2] in property current.
 • Or a call signature declaring the expected parameter / return type is missing in object type [3] but exists in
   function type [4].

     src/compose/ComposeBox.js
 [2] 142│   mentionWarnings = React.createRef<typeof MentionWarnings>();
        :
     497│
     498│     return (
     499│       <View style={this.styles.wrapper}>
     500│         <MentionWarnings narrow={narrow} stream={stream} ref={this.mentionWarnings} />
     501│         <View style={[this.styles.autocompleteWrapper, { marginBottom: height }]}>
     502│           <TopicAutocomplete
     503│             isFocused={isTopicFocused}

     /private/tmp/flow/flowlib_38b4cf0c/react.js
 [4] 197│   | ((React$ElementRef<ElementType> | null) => mixed)
        :
 [3] 248│   ): {|current: null | T|};

     src/compose/MentionWarnings.js
 [1] 170│ const MentionWarnings: AbstractComponent<Props, ImperativeHandle> = forwardRef<



Found 3 errors
```

-----

I found that the following would work, but I would appreciate hearing thoughts about it 🙂. I haven't yet been able to reconcile it with our findings in that linked discussion.

```
React.createRef<React$ElementRef<typeof MentionWarnings>>()
```

gnprice: > `React.createRef<React$ElementRef<typeof MentionWarnings>>()`

Yeah, that looks correct to me.

In that discussion I wrote this:
> Hmm, actually, now that I've reasoned through it more fully like that, I see that if my understanding of what's going on here is right, then `React$ElementRef<typeof Foo>` -- for any `Foo` we'll ever write -- is actually just a long way of spelling `Foo`.

But that was based in part on this, which I now see is… incomplete!
> Hmm, I guess `ref` can't be used on function components anyway.

What I really meant was that for a `Foo` that is a *class* component-type, `React$ElementRef<typeof Foo>` is actually just a long way of spelling `Foo`.

As I now understand it, it is indeed true that `ref` can't be used directly on a function component-type… but there are a bunch of other flavors of component type besides functions and classes, in particular the return values of `forwardRef`.

More generally, I think the story is that `React$ElementRef` takes anything that looks like `React$AbstractComponent<Config, Instance>`, and it picks out that `Instance` type. Here, we say right in the type annotation (new in this branch):
```js
const MentionWarnings: AbstractComponent<Props, ImperativeHandle> = // …
```
so `typeof MentionWarnings` is `AbstractComponent<Props, ImperativeHandle>`;
so `React$ElementRef<typeof MentionWarnings>` is `ImperativeHandle`. Which is the type you want here, and the same as the type that Flow infers for it given just `React.createRef()` with no type parameter. (It's using the fact that we pass this as the `ref` pseudoprop to `MentionWarnings`, down below.)

I started writing down some more notes on what I've worked out from studying these core-React types, and ended up expanding it into a new page in our docs tree:
https://github.com/zulip/zulip-mobile/blob/470ad571782501039369963eaf8d262802a2f07b/docs/background/react.md

See also where the [Flow upstream docs describe `ElementRef`](https://flow.org/en/docs/react/types/#toc-react-elementref), in a page I actually only just now found after writing that up. It doesn't cover what happens with `forwardRef`. But I think it does explain what the major motivation for `ElementRef` was, given that its behavior on class component-types and function component-types is so boring: it's for DOM element-types, where it maps `'div'` to `HTMLDivElement` and `'input'` to `HTMLInputElement` and so on.

gnprice: Anyway, given that Flow successfully infers the type here just fine, and because spelling it out is kind of a mouthful, I'm happy to just leave it to be inferred.

gnprice (commented):
Thanks @chrisbobbe ! Comments below. One is long, but the length is all background; all are simple to handle.

src/compose/ComposeBox.js
chrisbobbe: Thanks for all that detail and the new doc! 😄

> Anyway, given that Flow successfully infers the type here just fine, and because spelling it out is kind of a mouthful, I'm happy to just leave it to be inferred.

Now that I'm confident about what to annotate it as, I think including the annotation could be better. It seems to help Flow give better error messages.

If I do

```diff
- this.mentionWarnings.current.clearMentionWarnings();
+ this.mentionWarnings.current.clearMentionWarnings(123);
```

without the annotation, then I get this error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:500:63

Cannot create MentionWarnings element because a call signature declaring the expected parameter / return type is missing
in object type [1] but exists in function type [2] in property ref. [incompatible-type]

     src/compose/ComposeBox.js
     497│
     498│     return (
     499│       <View style={this.styles.wrapper}>
     500│         <MentionWarnings narrow={narrow} stream={stream} ref={this.mentionWarnings} />
     501│         <View style={[this.styles.autocompleteWrapper, { marginBottom: height }]}>
     502│           <TopicAutocomplete
     503│             isFocused={isTopicFocused}

     /private/tmp/flow/flowlib_3f0b09e6/react.js
 [2] 197│   | ((React$ElementRef<ElementType> | null) => mixed)
        :
 [1] 248│   ): {|current: null | T|};
```

But if I do that _with_ the annotation, I get this error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/compose/ComposeBox.js:382:57

Cannot call this.mentionWarnings.current.clearMentionWarnings because no arguments are expected by function type [1].
[extra-arg]

     src/compose/ComposeBox.js
     379│     this.setMessageInputValue('');
     380│
     381│     if (this.mentionWarnings.current) {
     382│       this.mentionWarnings.current.clearMentionWarnings(123);
     383│     }
     384│
     385│     dispatch(sendTypingStop(narrow));

     src/compose/MentionWarnings.js
 [1]  37│   clearMentionWarnings(): void,
```

chrisbobbe:
Thanks for the review and that doc! Revision pushed.

src/compose/ComposeBox.js
gnprice: Hmm indeed -- good thought to check that!

That's a good reason to write down the annotation.

gnprice:
Thanks! Merged.

PR #4943, opened 2021-08-04
Annotate most exported `connect`ed components.

chrisbobbe:
With the exception of `MentionWarnings`, which I've opened #4942 for, and `ComposeMenu`, which we have #4796 for.

Related: #4907

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -75,18 +75,24 @@ type SelectorProps = {|
   stream: Subscription | {| ...Stream, in_home_view: boolean |},
 |};
 
-type Props = $ReadOnly<{|
-  insets: EdgeInsets,
-
+type OuterProps = $ReadOnly<{|
   narrow: Narrow,
   editMessage: EditMessage | null,
   completeEditMessage: () => void,
+|}>;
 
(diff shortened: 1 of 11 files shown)


src/compose/ComposeBox.js
```diff
@@ -75,13 +76,19 @@ type SelectorProps = {|
...
   completeEditMessage: () => void,
+|}>;
 
+type Props = $ReadOnly<{|
+  ...OuterProps,
+
+  // from withSafeAreaInsets
+  insets: EdgeInsets,
+
+  // from `connect`
```
gnprice, line 91: nit: Let's put these (and the prop from `withGetText`, below) in a consistent order relative to how we do the wrapping, so it's easy to compare the two and see that they line up.

In the common case with just `connect`, we generally put those at the end, so I guess let's do it consistent with that: `OuterProps` first, then the next thing inward (which is `withGetText`), then the next thing inward, etc.

gnprice (commented):
Thanks @chrisbobbe! All looks good, with one nit below; then please merge at will.

chrisbobbe:
Thanks for the review! Merged, with that tweak.

Issue #4861, opened 2021-07-06
Reply from notification


alya:
It is common for a user to want to reply to a message they receive a notification about. At present, this requires clicking on the notification to get to the message, which takes some time.

It would be convenient to be able to reply to the message directly from the notification.

gnprice:
On Android, this will build on the APIs we'll start using as part of #2691.

On iOS, I'm not sure if this is possible.

PR #4945, opened 2021-08-05
[wip] Reply from notification in Android

AkashDhiman:
Builds up on #4861, ignore common commits.

This allows user to respond to notification from notifications itself, without having to open the app.
The idea is to create a [headless](https://reactnative.dev/docs/headless-js-android) js service which
gives control to javascript without having to start the react native tree.
within javascript we restore our redux store and call api.sendMessage
then if that succeeds we call a newly introduced react method UpdateNotification
to update the corresponding notification with the data we just sent.
(this updation is a requirement of android sdk)

Caveats:
- notifications aren't updated after reply is sent see 65ea451
- the whole process is a little slow atleast in debug build.
- cannot reply for an account other than the active one, won't be able to find auth. 

references:
- headless js: https://reactnative.dev/docs/headless-js-android
- tutorial: https://medium.com/software-engineers-toolbox/direct-reply-notification-for-android-in-4-steps-5612b49b5487
- https://developer.android.com/training/notify-user/build-notification#reply-action

Fixes: #4861


gnprice:
Thanks @AkashDhiman! Looking forward to getting this functionality into the app.

> cannot reply for an account other than the active one, won't be able to find auth.

We can fix this by using `getAccountsByIdentity`. The auth for all the accounts the user is logged in as is there in `state.accounts`.

Relatedly, the data the Android-native side sends over to the JS side will need to include the identity...

Hmm. There's actually an annoying mismatch here:
* The notification payload has the user's ID, but not their email. I.e., it's already living in the future where we identify users exclusively by ID.
* But the `state.accounts` data doesn't have the user's ID; only their email.

The real fix for that is that we should store the user ID in `state.accounts`, in the `Account` objects.

Until then:
* We can still make this work just fine for the case where the user doesn't have multiple accounts they're logged in as on the same realm/org. So we should do that.
* When there are several accounts in the same org, though, it's going to be ambiguous which one this notification was for, and which one the user intended to reply as. We should be sure not to guess, because if the user writes something they intend to send as one account and we go and send it as another, that is potentially quite bad. So in that case... for now we just give an error, I guess. Having several of these in the first place seems like a use case that should only apply to us Zulip developers and a few other users who have some reason to have test or demo accounts.
  * Though see #4831 -- there's some puzzling data to suggest it's a larger fraction of our users.

> the whole process is a little slow atleast in debug build.

How slow is it -- how long does it take?

I'll be happy merging a version of this even while it's fairly slow. There are several things we can potentially do to speed it up:
* If a lot of time is spent rehydrating the data, we can do something so that we only rehydrate `state.accounts`. That's all we need for making the reply -- that's where the API keys are.
* If a lot of time is spent importing all the JS code, we can arrange things so that for this task we only import what we need. RN supports lazy imports. This might involve some effort spent pruning our import graph, so that it's possible to import what we need here without importing most of the rest of the app.
* If a lot of time is spent starting up the JS engine, and/or either of the above things, we can potentially start it up eagerly before the user tries to send a reply. Perhaps we can do so when they open the input box to type a reply? If we can't get a signal then, perhaps we can start it up in the background when we show the notification.

We don't need to do any of those as a blocker for merging this feature, unless without them it's so very slow that it isn't really usable. Even if fairly slow, it'll be a useful feature for people, and between the options above I'm confident we'll be able to speed it up later.

gnprice:
> Hmm. There's actually an annoying mismatch here: […] The real fix for that is that we should store the user ID in `state.accounts`, in the `Account` objects.

I've just filed #4951 for that task.

gnprice:
> * notifications aren't updated after reply is sent see [65ea451](https://github.com/zulip/zulip-mobile/commit/65ea451ede51958afc2735ee99089a9834f31a27)

Yeah, sounds like there's some debugging to do there. Some questions I'd ask myself if trying to debug that include: At what point is it failing? Is it not finding the existing notification? Is it failing with some kind of exception at some point before it reaches the `notify` method call at the end of `updateNotification`? I'd add `Log.d` calls, like the ones you have in the existing draft, and watch them with `adb logcat` using filters like the ones in [our debugging HOWTO doc](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/debugging.md#adb-logcat).

Another thought I have looking at that code, or a pair of related thoughts:
* It feels like a lot of that is duplicated between this and the code from #4842. It'd be good to deduplicate a lot of that, by factoring out into a common helper or two -- like one returning, and/or acting on, a `NotificationCompat.Builder`.
* Let's avoid writing nontrivial amounts of code like this in Java, and keep it in Kotlin. (Per [our Kotlin/Java strategy](https://github.com/zulip/zulip-mobile/blob/master/docs/architecture/android.md#migrating-our-code): all new code is in Kotlin, not Java.) Two ways to do that:
  * The easiest way: pull the whole content of that function into a function in `FcmPushNotifications.kt`, next to the other similar functions, and make this function a trivial wrapper that just calls that.
  * The more thorough way, and probably actually not much more work: have a prep commit that converts `NotificationsModule` to Kotlin.

Issue #4916, opened 2021-07-21
Search syntax auto-complete


alya:
We should offer syntax suggestions for the search box even before the user starts typing. Otherwise, it's much harder to figure out how to use search syntax.

A possible list:
```
stream:
topic:
pm-with:
sender:
is:
has:
```

Further suggestions on how to complete the query should be given when search syntax is inserted, similar to the webapp search.

This issue is blocked on #4903.

chrisbobbe:
We've had another report of this: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Bug.20reports/near/1296919

PR #4946, opened 2021-08-05
Modify Search functionality 

AkashDhiman:
This PR does the prep work for implementing #4859 it changes the behavior of search from on query change to on submit.

Fixes: #4964

### DIFF
--- a/src/common/Screen.js
+++ b/src/common/Screen.js
@@ -5,6 +5,7 @@ import type { Node } from 'react';
 import { ScrollView } from 'react-native';
 import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 import { SafeAreaView } from 'react-native-safe-area-context';
+import type { EditingEvent } from 'react-native/Libraries/Components/TextInput/TextInput';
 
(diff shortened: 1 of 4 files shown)


gnprice (commented):
Thanks @AkashDhiman !

It looks like the current version of this PR does just the first of those two things:
> changing the search behavior to search on submit instead of search on text change

Which is perfect -- I'd be happy to merge a PR that does just that. The "fetch more on scroll" feature #4859 will be fairly complex, so its PR thread might get long and it's probably best to keep it separate from PRs for other changes anyway.

A couple of comments below; please take a look.

(Is there an issue for this change?)

AkashDhiman:
@gnprice thanks for looking into it, I made the changes, This PR only changes the search functionality from on Change to on Submit, I will push the code for fetch more on search on a separate PR.

gnprice:
Thanks, looks good -- merged!

I tweaked the commit message a bit: I don't think it's likely that we'll make the Zulip search backend start doing prefix searches in the future. This new only-on-submit behavior in the client is also helpful because it opens up space in the UI for search suggestions, #4916.

PR #4948, opened 2021-08-05
Annotate more exports.

chrisbobbe:
I promise we're getting nearer to done with these! 🙂

After this, and after the currently open PRs #4943, #4942, #4926, and after we do something about `ZulipAsyncStorage` (see [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/ZulipAsyncStorage.20extends.20AsyncStorage.20.28.3F.29/near/1244517) and #4947), there will be just a few small things that I expect will basically disappear naturally with #4898 and #4868 (and maybe a third PR that's not occurring to me right now—if so, it's tiny).

This also adds a lot of read-only annotations that aren't strictly necessary for our work on annotating things for #4907. But I figured it's a good thing to do, and I'd happily split that off to a different PR if you'd like.

Related: #4907

### DIFF
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -242,7 +242,7 @@ export type HuddlesUnreadItem = {|
   user_ids_string: string,
 
   // Sorted.
-  unread_message_ids: number[],
+  unread_message_ids: $ReadOnlyArray<number>,
 |};
 
 /** The unreads in a 1:1 PM thread, as represented in `unread_msgs`. */
@@ -259,7 +259,7 @@ export type PmsUnreadItem = {|
   sender_id: UserId,
 
   // Sorted.
-  unread_message_ids: number[],
+  unread_message_ids: $ReadOnlyArray<number>,
 |};
 
 /** Initial data for `update_message_flags` events. */
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -16,40 +16,40 @@ import type {
   UserStatusMapObject,
 } from './apiTypes';
 
-export type InitialDataBase = {|
+export type InitialDataBase = $ReadOnly<{|
   last_event_id: number,
   msg: string,
   queue_id: number,
-|};
+|}>;
 
-export type InitialDataAlertWords = {|
+export type InitialDataAlertWords = $ReadOnly<{|
   alert_words: string[],
-|};
+|}>;
 
-export type InitialDataMessage = {|
+export type InitialDataMessage = $ReadOnly<{|
   max_message_id: number,
-|};
+|}>;
 
 export type MuteTuple = [string, string];
 
-export type InitialDataMutedTopics = {|
+export type InitialDataMutedTopics = $ReadOnly<{|
   muted_topics: MuteTuple[],
-|};
+|}>;
 
 /** Added in server version 4.0, feature level 48 */
-export type InitialDataMutedUsers = {|
+export type InitialDataMutedUsers = $ReadOnly<{|
   muted_users?: MutedUser[],
-|};
+|}>;
 
-export type InitialDataPresence = {|
+export type InitialDataPresence = $ReadOnly<{|
   presences: {| [email: string]: UserPresence |},
-|};
+|}>;
 
-export type AvailableVideoChatProviders = {|
-  [providerName: string]: {| name: string, id: number |},
-|};
+export type AvailableVideoChatProviders = $ReadOnly<{|
+  [providerName: string]: $ReadOnly<{| name: string, id: number |}>,
+|}>;
 
-export type InitialDataRealm = {|
+export type InitialDataRealm = $ReadOnly<{|
   jitsi_server_url?: string,
   max_icon_file_size: number,
   realm_add_emoji_by_admins_only: boolean,
@@ -57,7 +57,7 @@ export type InitialDataRealm = {|
   realm_allow_edit_history: boolean,
   realm_allow_message_deleting: boolean,
   realm_allow_message_editing: boolean,
-  realm_authentication_methods: { GitHub: true, Email: true, Google: true, ... },
+  realm_authentication_methods: $ReadOnly<{ GitHub: true, Email: true, Google: true, ... }>,
   realm_available_video_chat_providers: AvailableVideoChatProviders,
   realm_bot_creation_policy: number,
   realm_bot_domain: string,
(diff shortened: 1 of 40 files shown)


src/api/modelTypes.js
```diff
@@ -339,7 +339,7 @@ export type Topic = {|
 // See docs: https://zulip.com/api/construct-narrow
 // prettier-ignore
 /* eslint-disable semi-style */
-export type NarrowElement =
+export type NarrowElement = $ReadOnly<
  | {| +operator: 'is' | 'in' | 'topic' | 'search', +operand: string |}
```
gnprice, line 343: This `$ReadOnly` is redundant with the marks on each of the individual properties, where it's spelled `+`. Here's where that syntax is documented:

The per-property version is actually the fundamental one -- Flow doesn't have a notion of an object type as a whole being read-only, only of individual properties being that way. When we say a whole type is read-only, it's a shorthand for saying all its properties are. And similarly, `$ReadOnly`'s meaning is defined in terms of making each of the properties read-only:

(Similarly at a couple of widget-related types below.)

src/reduxTypes.js
```diff
@@ -298,24 +298,24 @@ export type SettingsState = {|
...
+|}>;
 
-export type TypingState = {|
+export type TypingState = $ReadOnly<{|
   [normalizedRecipients: string]: {|
     time: number,
     userIds: UserId[],
   |},
-|};
+|}>;
```
gnprice, line 316: The inner objects here should also get read-only marks. Easy to miss those :wink:

gnprice (commented):
Thanks for going through these! Small comments below. Probably the main thing is that I'd like to avoid duplicating a lot of the type annotations, where we already have annotations that are just not quite in the spot that types-first needs them to be in.

src/api/modelTypes.js
chrisbobbe: Hmm, thanks! That makes sense.

As it says in the commit message, I intended to make changes only where Flow specifically suggested them. But I don't see an error if I revert this change, and the others I think you've indicated. 🤷

gnprice: This is in `modelTypes`, which is one of the files that it looks like you started with (which makes sense, because these are primarily the types we use to cast the arbitrary JSON we get from the server.)

chrisbobbe: Oh, got it, right!

src/reduxTypes.js
chrisbobbe: Ah, right. I've found a lot more of these, I think all of them. Thanks!

chrisbobbe:

src/settings/LanguagePicker.js
```diff
@@ -28,7 +29,7 @@ export default class LanguagePicker extends PureComponent<Props> {
       };
     });
 
-  getFilteredLanguageList = (filter: string): Language[] => {
+  getFilteredLanguageList: string => Language[] = (filter: string): Language[] => {
```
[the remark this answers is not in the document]
gnprice: Still have a redundancy at this one :slightly_smiling_face: :
```diff
-  getFilteredLanguageList = (filter: string): Language[] => {
+  getFilteredLanguageList: string => Language[] = (filter: string) => {
```
Also `isSendButtonEnabled`.

src/pm-conversations/pmConversationsModel.js
```diff
@@ -190,7 +194,15 @@ export function reducer(
...
-        st = insert(st, keyOfExactUsers(r.user_ids), r.max_message_id);
+        st = insert(
+          st,
+          keyOfExactUsers(
+            r.user_ids
+              // Copy the array (keyOfExactUsers sorts its input in-place)
+              .map(id => id),
+          ),
+          r.max_message_id,
+        );
```
gnprice, line 205: Hmm, awkward that this makes this one line sprawl across 9 lines. I'd take that as a cue to pull out some subexpression as a local :wink: 

E.g. like this:
```diff
       for (const r of action.data.recent_private_conversations ?? []) {
-        st = insert(
-          st,
-          keyOfExactUsers(
-            r.user_ids
-              // Copy the array (keyOfExactUsers sorts its input in-place)
-              .map(id => id),
-          ),
-          r.max_message_id,
-        );
+        const key = keyOfExactUsers([...r.user_ids]);
+        st = insert(st, key, r.max_message_id);
       }
```

I think the comment isn't actually needed, because after all the types say we need a mutable array and that we have a read-only one. This bit of code doesn't actually need to care about *why* the helper demands a mutable array.

(Meanwhile although the `map` works fine, the spread feels more direct to me, and also seems more likely to be reliably efficient -- no callback to invoke for each element.)

src/api/initialDataTypes.js
```diff
@@ -121,67 +121,67 @@ export type RawInitialDataRealmFilters = {|
...
+|}>;
 
-export type RawInitialDataRealmUser = {|
+export type RawInitialDataRealmUser = $ReadOnly<{|
   avatar_source: 'G',
   avatar_url: string | null,
   avatar_url_medium: string,
   can_create_streams: boolean,
-  cross_realm_bots: Array<{| ...CrossRealmBot, avatar_url?: string | null |}>,
+  cross_realm_bots: $ReadOnlyArray<{| ...CrossRealmBot, avatar_url?: string | null |}>,
```
gnprice, line 139: Yet another layer of nesting: this `avatar_url` property isn't read-only :stuck_out_tongue: 

This is a good application for the `+` syntax -- where there's just one or two properties and the whole object type fits on a line, using `+` vs `$ReadOnly` can often save it from spilling into several lines.

src/api/modelTypes.js
```diff
@@ -350,8 +350,7 @@ export type NarrowElement =
  | {| +operator: 'pm-with', +operand: string | $ReadOnlyArray<UserId> |}
  | {| +operator: 'sender', +operand: string | UserId |}
  | {| +operator: 'group-pm-with', +operand: string | UserId |}
- | {| +operator: 'near' | 'id', +operand: number |} // message ID
- ;
+ | {| +operator: 'near' | 'id', +operand: number |}; // message ID
```
gnprice, line 353: nit: small noise formatting change

(The existing formatting here is intentional, for making things slightly easier when adding more branches.)

src/reduxTypes.js
```diff
@@ -298,24 +298,24 @@ export type SettingsState = {|
...
+export type TopicsState = $ReadOnly<{|
+  [number]: $ReadOnlyArray<Topic>,
+|}>;
 
-export type TypingState = {|
-  [normalizedRecipients: string]: {|
+export type TypingState = $ReadOnly<{|
+  [normalizedRecipients: string]: $ReadOnly<{|
     time: number,
     userIds: UserId[],
```
gnprice, line 314: Another bit of still-deeper nesting: that `UserId` array.

gnprice (commented):
Thanks for the revision! Small comment above:
https://github.com/zulip/zulip-mobile/pull/4948#discussion_r688094871
and some others below.

Some of these represent yet more opportunities to add read-only marks. There's no need for those to all happen atomically (beyond what's needed to keep Flow passing), so it's fine to merge some changes now and handle more stragglers later.

src/settings/LanguagePicker.js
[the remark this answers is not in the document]
chrisbobbe: Indeed! Thanks.

chrisbobbe:

gnprice:

There was still a small noise diff at `NarrowElement`:
```diff
@@ -351,7 +351,7 @@ export type NarrowElement =
  | {| +operator: 'sender', +operand: string | UserId |}
  | {| +operator: 'group-pm-with', +operand: string | UserId |}
  | {| +operator: 'near' | 'id', +operand: number |} // message ID
- ;
+;
 
 /**
```
but I just fixed that before merge.

chrisbobbe:
Oh, oops—thanks for the fix and review! 🙂

PR #4950, opened 2021-08-09
Upgrade reselect, date-fns, color, and base-64 to their latest, bumping major versions.

chrisbobbe:


gnprice:
We're fine with ES6 syntax. So we shouldn't have to do anything
    special to take this upgrade.
    
    ...Ah, but we're not yet fine with some *ES12* syntax that was snuck
    in unannounced: the "NumericLiteralSeparator" [2] [3]. So, bring in
    a Babel plugin for that, for now. I see this has caused some
    controversy:
      - https://github.com/Qix-/color/issues/204
      - https://github.com/Qix-/color/issues/206
      - https://github.com/Qix-/color/pull/208
      - https://github.com/Qix-/color/issues/209

Fun.

It seems like the root of this is that there's some ambiguity in the NPM-using world about whether the published packages should have the code transpiled down to an older version of JS. I agree with this observation in a comment on that issue 206:
> I have ~100 direct dependencies (all of them kept to the latest version with https://renovatebot.com/), thousands of transitive dependencies. I don't have to transpile any of the other dependencies. So think that the industry standard is that npm packages already contain the code transpiled (if needed).

In principle it would also be perfectly reasonable to have the alternative protocol, where packages are published as up-to-date modern JS and the transpilation happens at application build time. That's the world the `color` maintainer evidently wants to live in, and it has advantages as well as disadvantages.

But it really does not work well for a library to try to live in that world while applications expect another. I'm not a fan of the `color` maintainer's belligerent attitude here, and some of the things they're saying (there's no such thing as "Vue's parser"; none of this is about what fancy syntax they use in their source code, but about the protocol for what's published in the built artifact) indicate they don't really understand how this works or what the problem is.

Anyway, glad you sorted it out.

gnprice (commented):
Looks good, thanks! Merging.

chrisbobbe:
Thanks for the review!

PR #4952, opened 2021-08-10
deps: Upgrade 6 more libraries across major versions.

chrisbobbe:


package.json
```diff
@@ -109,15 +108,15 @@
...
+    "flow-typed": "^3.3.1",
     "immutable-devtools": "^0.1.5",
     "jest": "^26.6.3",
     "jest-cli": "^26.4.1",
     "jest-environment-jsdom": "^26.3.0",
     "jest-environment-jsdom-global": "^2.0.4",
     "jest-expo": "^40.0.1",
     "jest-extended": "^0.11.5",
-    "jetifier": "^1.6.5",
+    "jetifier": "^2.0.0",
```
gnprice, line 119: Right now, for us, the warning looks like this:
    
    ```
    Jetifier is deprecated. Future versions of react-native CLI may not
      run it by default.
    Jetifier found 1372 file(s) to forward-jetify. Using 12 workers...
    Jetifier: propose an AndroidX conversion PR to this repository:
      node_modules/react-native-photo-view/android/src/main/java/com/reactnative/photoview/ImageEvent.java
    Jetifier: propose an AndroidX conversion PR to this repository:
      node_modules/react-native-photo-view/android/src/main/java/com/reactnative/photoview/PhotoView.java
    ```
    
    I'm not sure if we have to wait for 1372 files to move to AndroidX
    before we can stop using this tool...but at least we have an issue
    (#4217) for using something more maintained than
    react-native-photo-view, which should take care of the two mentions
    of that library.

I think what's happening is that there are 1372 files that it has to look at the contents of, basically all the Java and Kotlin files (and some others) in our `node_modules` tree… and then it prints a warning for each one of them where it finds there's actually something for it to rewrite. Which thankfully is just those two files in one library.

If you do something like `rg -lw androidx node_modules`, you find there are lots of files that already mention AndroidX; most of those are importing from it, i.e. they've done this migration by now. Looks like react-native-photo-view is the one straggler among our dependencies.

gnprice (commented):
Thanks for going through these! All looks good; merging.

chrisbobbe:
Thanks for the review!

PR #4955, opened 2021-08-11
deps: Upgrade flow-coverage-report to 0.8.0, the latest.

chrisbobbe:
No breaking changes are announced:
  https://github.com/rpl/flow-coverage-report/releases.

Note that I haven't tested it at all to see if it still works,
because it's not part of our normal workflow and I'm not familiar
with its previously intended behavior.

Greg points out that it'd be useful to see what this tool has to say
now, but it's mostly been unused for a while:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Dependabot/near/1203387.

Anyway, apart from it being nice in general to be on the latest
versions of things, this also removes all five low-severity
vulnerabilities reported by `yarn audit` that implicate
flow-coverage-report.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "eslint-plugin-react-hooks": "^4.2.0",
     "fbjs-scripts": "1.2.0",
     "flow-bin": "^0.137.0",
-    "flow-coverage-report": "^0.6.0",
+    "flow-coverage-report": "^0.8.0",
     "flow-typed": "^3.3.1",
     "immutable-devtools": "^0.1.5",
     "jest": "^26.6.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2031,7 +2031,7 @@
     estree-walker "^1.0.1"
     picomatch "^2.2.2"
 
-"@rpl/badge-up@2.2.0":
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks for spotting this upgrade!

The workflow where we've used this is described in #2052: you run `yarn test:flow-coverage`. When you do, it thinks for a while and when it's done it spits out a line like:
```
View generated HTML Report at file:///home/greg/z/mobile/flow-coverage/index.html
```

That page looks like this:
image
(That's after I've clicked on "Uncovered" to sort by that column, foregrounding the biggest opportunities to cover more code.)

And you can click on any of the files mentioned to get a view of that file highlighting the untyped expressions, like this:
image

I took a look just now at the output before and after this change. It's visually identical in the browser; the only detectable difference is that the new version is about twice as fast :slightly_smiling_face: -- about 30s on my desktop, vs. about 60s.

It also distinctly finds some things we should do! Most notably, we should really put types on redux-persist. I'll update #2052, and probably file a new issue or two.

I'll also merge this change, after adjusting the commit message.

chrisbobbe:
Great, thanks for the review and testing!

Issue #4853, opened 2021-07-03
Fetch data when notification received


alya:
It is common for a user to open the Zulip app right after receiving a notification. At present, when the user does so, they still have to wait for the app to fetch the message that triggered the notification. This is a confusing and potentially frustrating experience, as it feels to the user as if the app should already "know" about the message.

To improve the experience, we should experiment with fetching messages immediately when a notification is received, in expectation that the user may open the app shortly thereafter.

gnprice:
One thing that complicates this is that we currently only store data from the server for one account at a time (aka "the active account"). For purposes of this issue -- which is an optimization anyway -- it'll be enough to have it work in the case when the notification is for the active account.

The native-side notifications code doesn't know which account is the active account, so it'll always want to start up the JS code. But the JS code we write for this can then look and see if the notification is for the active account, and promptly return if it's not.

In order to implement starting up the JS code, on the iOS side there should be a feature we can start using in the underlying RN notifications library that will do this. On the Android side, we'll want to add a bit of Kotlin code. A useful source of example code will be the code we currently have for actually launching the app when the user opens the notification -- this code will be different, but similar.

gnprice:
A major open question we'll want to answer with this experiment is whether the OSes will tolerate the amount of time, CPU, and network traffic we spend by doing this. Both Android and iOS warn that when your app is in the background and receives a notification, it should take care to finish its work handling the notification quickly, and if it goes too long it'll just get shut down.

The iOS / APNs docs also warn somewhere, IIRC (and I forget if the Android or FCM docs say something like this too) that if your app is real egregious in spending a lot of time handling its notifications, the OS may decide to just not always give it all its notifications.

So we'll want to (a) go back to those docs and review what they say -- in particular, if they give a sense of whether "too long" is more like 1 second or 10 seconds or 60 seconds; (b) experiment with how long we do end up spending, when we try this, so we can compare to those guidelines; (c) try to see if empirically we get into any trouble :wink:, i.e. if we spot any evidence of the OS clamping down.

gnprice:
Some more discussion starting here about the fetches this will need to do, and why:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/fetch.20data.20when.20notification.20received.20.23M4853/near/1246572

We'll want to do something like `doInitialFetch`, and we'll want to invoke `fetchMessagesInNarrow`.

(And all of that only if the notification is for the active account, as I mentioned above.)

PR #4956, opened 2021-08-12
Fetch data when notification is received

AkashDhiman:
Fixes: #4853

The first commit for this PR is similar to reply from notifications, where we configure a headless task following https://reactnative.dev/docs/headless-js-android.

The idea I had in mind to get this to work was
* on receiving a firebase message we start the headless service which starts up javascript.
* from javascript we access the redux store and dispatch `EVENT_ADD_MESSAGE`, (we can also consider other associated events here as well)
* this will construct a redux state with our new message in the unread state.
* then we quit from javascript and this saves the unreadState with redux-persist.
* on opening the app normally, it restores this unread state and sees the latest messages.

Issues I currently face:
* Persistence isn't actually working, EVENT_ADD_MESSAGE is triggered, and state does change but it doesn't get stored, new notification restore old state and forget the previous ones, moreover opening the app also doesn't show a change of behavior. Possible reason could be that since a service doesn't die when finished we don't actually finish hence we don't trigger redux-persist to store the data. (I will try to do an exit call from service to see if that is indeed the case)
* Flow is giving me error on dispatching EVENT_ADD_MESSAGE, I am unable to dispatch functions as well it seems, I usually look at other places where similar code is used to try to resolve such scenarios and type hinting fails at many places dispatch is called.
* message provided via notification and what is expected in the payload with EVENT_ADD_MESSAGE are somewhat different, I was wondering if it is possible for them to be more in sync with each other.

The commits are roughly implemented as of now.

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/FcmListenerService.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FcmListenerService.kt
@@ -18,7 +18,9 @@ class FcmListenerService : FirebaseMessagingService() {
         run {
             val service = Intent(applicationContext, HeadlessFetchService::class.java)
             val bundle = Bundle()
-            bundle.putString("foo", "bar")
+            for ((key, value) in message.data) {
+                bundle.putString(key, value)
+            }
             service.putExtras(bundle)
             applicationContext.startService(service)
         }
--- a/src/headless/FetchTask.js
+++ b/src/headless/FetchTask.js
@@ -1,8 +1,80 @@
 /* @flow strict-local */
+import invariant from 'invariant';
+import { EVENT_NEW_MESSAGE } from '../actionConstants';
+import { makeUserId } from '../api/idTypes';
+import store, { restore } from '../boot/store';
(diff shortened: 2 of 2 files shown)


gnprice:
Thanks @AkashDhiman! Seeing this is very helpful.

It's not going to work to take just the data that's in the notification payload; we really want to make some requests to the server to fetch the data we need, because it doesn't all fit in a notification payload in general. I went and wrote more details in chat, starting here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/fetch.20data.20when.20notification.20received.20.23M4853/near/1246572

The Flow error at `EVENT_NEW_MESSAGE` is moot given that, but I took a quick look at the cause of the error and discussed it starting here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/fetch.20data.20when.20notification.20received.20.23M4853/near/1246567

Re persistence, I don't have as much to say -- I look forward to the results of your debugging :slightly_smiling_face:. A few thoughts here, though:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/fetch.20data.20when.20notification.20received.20.23M4853/near/1246587

PR #4957, opened 2021-08-13
api types: Mark most remaining types read-only

gnprice:
This largely covers the remaining types here, inspired by #4948, with the help of some Perl find-and-replace.

There isn't anything special about the API that means its objects and arrays should be read-only -- after all, the actual server itself can't possibly care. But we reuse a lot of these for Redux actions and for pieces of our Redux store, where we do want things to be read-only, so it's helpful to just consistently mark here too.


chrisbobbe:
Thanks, LGTM! Merged.

PR #4960, opened 2021-08-13
redux-persist: Improve logging of errors.

chrisbobbe:
`logging.warn` has the best chance of helpfully reporting an Error
object if we pass it as the first argument. In some of these cases,
we weren't even passing an Error at all, when we had one available
-- so, do so, and have it be the first argument.

The second argument should be a plain, JSONable object. So, make
that happen, too, so we don't get Sentry events with things like

  "extra": {
    "0": "u",
    "1": "s",
    "2": "e",
    "3": "r",
    "4": "G",
    "5": "r",
    "6": "o",
    "7": "u",
    "8": "p",
    "9": "s"
  }

### DIFF
--- a/src/third/redux-persist/createPersistor.js
+++ b/src/third/redux-persist/createPersistor.js
@@ -84,7 +84,7 @@ export default function createPersistor (store, config) {
           let value = data
           state[key] = value
         } catch (err) {
-          logging.warn('Error rehydrating data for a key', { key, err })
+          logging.warn(err, { message: 'Error rehydrating data for a key', key })
         }
       })
     } else state = incoming
@@ -113,7 +113,7 @@ export default function createPersistor (store, config) {
 
 function warnIfSetError (key) {
   return function setError (err) {
-    if (err) { logging.warn('Error storing data for key:', key, err) }
+    if (err) { logging.warn(err, { message: 'Error storing data for key:', key }) }
   }
 }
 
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Bugs like this will be helped by type-checking our code in `src/third/redux-persist`—I have a growing backlog of cleanups there that are waiting for #4694; it'd be great to get that in. 🙂

chrisbobbe:
For those with access, the Sentry event quoted above is this one: https://sentry.io/organizations/zulip/issues/2573194717/?project=191284

src/third/redux-persist/createPersistor.js
```diff
@@ -113,7 +113,7 @@ export default function createPersistor (store, config) {
 
 function warnIfSetError (key) {
   return function setError (err) {
-    if (err) { logging.warn('Error storing data for key:', key, err) }
+    if (err) { logging.warn(err, { message: 'Error storing data for key:', key }) }
```
gnprice, line 116: Looks like this is the line that generated the Sentry event you quoted.

Ah, so I guess what happened is that the second argument was a string -- and our types say it should really be one of these objects:
```js
type Extras = {| +[key: string]: JSONable |};
```
and it ultimately gets passed to a `setExtras` method. And I guess that really wants the value to indeed be an object, not a string, and ends up enumerating its properties, perhaps like this:
```
> Object.entries("abc")
[ [ '0', 'a' ], [ '1', 'b' ], [ '2', 'c' ] ]
```

Yeah, it would indeed be good to have type-checking in this file.

gnprice (commented):

chrisbobbe:
Thanks for the review!

Issue #4841, opened 2021-07-01
Store persistent data in a sound way


gnprice:
The way we currently store the user's data on the device's storage, between runs of the app, is with the design from redux-persist:
* The units of storage are the top-level subtrees of the Redux state: `state.accounts` is one, `state.users` is another, and so on.
* When something changes, we write out the new version of each top-level subtree that changed: if a user changed their avatar then we write out the entire new `state.users`, and so on.
* When things change across several different top-level subtrees, we write out each of them separately.

As a result, it's possible for the stored data to get into a state that was never the state inside the app, and that's internally inconsistent. This was the cause of #4587, for example.

For the most part, this doesn't cause user-visible problems today -- the crashes in #4587 were an exception to that pattern.

But the reason it doesn't is that we currently don't really rely that heavily on the stored data: each time the app starts up, we go and do an initial fetch of the server data from scratch. That means that for all the data that ultimately comes from the server -- which includes most of our data and most of the complexity in our data -- any inconsistencies that creep into the stored data will typically get wiped away almost immediately after we first even read it from storage. (The way that issue #4587 managed to evade that protection is that it caused a crash before we got far enough to do that initial fetch.) So we're largely protected from consequences of this issue… but the price is that we're not really getting much use out of storing that data.

There are several things we'd like to do that would mean relying more firmly on the stored data, and not constantly refreshing it from scratch:
* The status quo means that when you launch the app, the data you're looking at is stale and there's a long lag before you see up-to-date data. This is an especially bad experience when you got there by opening a notification -- there's a message you know you got, it was right there in the notification, but we don't initially know about it. We have the basic design for how to fix this, as #3916; but it means maintaining your data in device storage and relying on that, rather than refreshing from scratch.
* We'd like to improve the experience of switching between several accounts (in several orgs): #1321. Doing a really good job of that would require maintaining all the same data about all your accounts as we currently do about the one active account; and fundamentally, doing much at all in that direction will require maintaining at least *some* data about all your accounts. In principle I guess we could maintain all that data while keeping the reload-from-scratch-on-each-launch strategy; but it'd mean even *more* data getting reloaded from scratch, so even more lag.
* We'd like to offer fuller-featured, higher-quality notifications: with an inline reply action (#2512), with end-to-end encryption from the Zulip server (#1190), pre-fetching the conversation's messages so it's ready to go if you open the notification (#3599), among other things.

[...]

PR #4961, opened 2021-08-13
Check active account, too, in getHaveServerData

gnprice:
This isn't itself server data -- it's what we use for acquiring server data -- which is why we didn't cover it when expanding these checks in 916dc4b28 / #4588.

But like the server data, it's something we assume is present when rendering the main UI of the app.  And because of the same issue #4841, the same lack of consistency in our data-storage model with redux-persist that caused #4587 and made that fix necessary, it's possible for it to be lacking data when the existing checks in this function all pass.

So, add a check for this too.

Also expand the comments in this function, particularly on the tricky balance it has to maintain. Some care is needed when adding checks here -- too much care, really, and I'd love to find a way to arrange things to make these delicate properties be more automatically true. Pending that, add some comments to make the needed reasoning more explicit, so that at least it's easier to re-confirm without reasoning back through everything from scratch.

### DIFF
--- a/src/account/AccountItem.js
+++ b/src/account/AccountItem.js
@@ -51,7 +51,12 @@ type Props = $ReadOnly<{|
 export default function AccountItem(props: Props): Node {
   const { email, realm, isLoggedIn } = props.account;
 
+  // Don't show the "remove account" button (the "trash" icon) for the
+  // active account when it's logged in.  This prevents removing it when the
+  // main app UI, relying on that account's data, may be on the nav stack.
+  // See `getHaveServerData`.
   const showDoneIcon = props.index === 0 && isLoggedIn;
+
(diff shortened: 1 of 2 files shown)


gnprice:
I tested manually that this resolves a possible crash:
* Hacked up `usersReducer` and `realmReducer` so that on LOGOUT they don't clear state.
* Logged out of the active account.
* Removed that account, i.e. used the "trash" icon. Had no other accounts.
* Then logged in afresh.

Without these changes, the result was a crash upon logging in. (Also at the logout, step 2; but killing the app and restarting it got me to the account picker for step 3.)

With these changes, the result was the loading screen followed by the normal main app UI.

chrisbobbe (approved):
Thanks, this LGTM! See some tiny comments below, which I've addressed before merging.

gnprice:
Thanks for the review and fixes!

PR #4965, opened 2021-08-19
initialFetchAbort: Don't kick user out if we have server data.

chrisbobbe:
I have a draft in progress for a better solution to this; see
discussion at
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/connection.20failed/near/1248170.

But for now this should fix an annoying situation that's so far been
observed on Alya's Android device, where a "Gave up trying to
connect" alert appears immediately on startup. The theory is that
the app stops making progress on the network request when the app is
backgrounded, while of course it doesn't stop approaching the
timeout. Then, when the app is foregrounded again, the request
hasn't succeeded but the timeout has fired, and the alert is shown.

### DIFF
--- a/src/message/fetchActions.js
+++ b/src/message/fetchActions.js
@@ -219,6 +219,7 @@ const initialFetchAbortPlain = (reason: InitialFetchAbortReason): Action => ({
 export const initialFetchAbort = (
   reason: InitialFetchAbortReason,
 ): ThunkAction<Promise<void>> => async (dispatch, getState) => {
+  dispatch(initialFetchAbortPlain(reason));
   showErrorAlert(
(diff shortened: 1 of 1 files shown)


gnprice:
This code looks good! Have you been able to test it out empirically?

(More thoughts in [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/connection.20failed/near/1248224).)

chrisbobbe:
Yep, it seems to work just as expected; see more in the chat thread. 🙂

gnprice:
Thanks! Merged; just added a bit about that manual testing to the commit message.

PR #4967, opened 2021-08-25
notif: Require sender_id, server, and realm_uri, rejecting ancient servers' data

gnprice:
These fields are all present since Zulip Server 1.9, and two of them even since 1.8, so it's long past time we should be able to not worry about servers that don't send them. Having them always present will be helpful for the work @AkashDhiman is doing on our Android notification UI, as discussed here:
https://github.com/zulip/zulip-mobile/pull/4842#discussion_r685546372

---

~~This PR is a draft because before we start ignoring these notifications from old (pre-1.9) servers, I'd like to set things up so that we log warnings to Sentry when we do so. There's a couple of commits at the start of this branch aimed at doing that, but then I realized that they can't possibly work yet because the Sentry destination (aka "DSN" or "client key") lives only in the JS code, at `src/sentryConfig.js`, and without that there's no way that the native-Android Sentry SDK can know where a Sentry event should go. So I'll see about fixing that and getting Sentry reporting working from the native Android side, and then this can go atop that.~~

(Sentry reporting from the native iOS side would be great too -- not so much for our own code, because we don't have much there, but because hopefully it can let us know if there are crashes happening there. That's also another benefit of getting it set up on Android.)

~~In the meantime, I think all the commits reflect what I'd want to merge once that preparatory work is in.~~

### DIFF
--- a/src/notification/__tests__/notification-test.js
+++ b/src/notification/__tests__/notification-test.js
@@ -25,6 +25,7 @@ describe('getNarrowFromNotificationData', () => {
 
   test('recognizes stream notifications and returns topic narrow', () => {
     const notification = {
+      realm_uri,
       recipient_type: 'stream',
       stream: 'some stream',
       topic: 'some topic',
@@ -37,6 +38,7 @@ describe('getNarrowFromNotificationData', () => {
     const users = [eg.selfUser, eg.otherUser];
     const allUsersByEmail: Map<string, UserOrBot> = new Map(users.map(u => [u.email, u]));
     const notification = {
+      realm_uri,
       recipient_type: 'private',
       sender_email: eg.otherUser.email,
     };
@@ -49,6 +51,7 @@ describe('getNarrowFromNotificationData', () => {
     const allUsersByEmail: Map<string, UserOrBot> = new Map(users.map(u => [u.email, u]));
 
     const notification = {
+      realm_uri,
       recipient_type: 'private',
       pm_users: users.map(u => u.user_id).join(','),
     };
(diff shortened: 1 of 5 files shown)


gnprice:
> This PR is a draft because before we start ignoring these notifications from old (pre-1.9) servers, I'd like to set things up so that we log warnings to Sentry when we do so.

Sent that as #4996. Once that's merged, I'll rebase this and mark as ready.

gnprice:
OK, we now have Sentry reporting in our Android-native code, and this PR is ready!

src/notification/__tests__/notification-test.js
```diff
@@ -11,6 +11,8 @@ import * as eg from '../../__tests__/lib/exampleData';
 import { fromAPNsImpl as extractIosNotificationData } from '../extract';
 import objectEntries from '../../utils/objectEntries';
 
+const realm_uri = eg.realm.href;
```
chrisbobbe, line 14: Tiny nit: I've slightly preferred `.toString()` over `.href`. What do you think?

We're asking for a different representation of the whole URL, and I think `.toString()` is a bit clearer about that. `.href` looks a little bit like `.origin`, `.hostname`, and friends: the things you'd use to pull out just a component of the URL.

chrisbobbe (approved):
Thanks, this LGTM! I've also tested on Android that some simple notifications worked as expected.

src/notification/__tests__/notification-test.js
gnprice: Sure, that's probably a bit nicer. Changing that and merging.

gnprice:

PR #4968, opened 2021-08-26
redux: Store userId in state.accounts

### LINKED ISSUE #4951: Store user ID in accounts state
Currently for each of the accounts the user is logged in as, we store:
* the URL for that org/realm;
* the user's email that they use to log in there;
* their API key, their secret credential that represents being logged in there;
* a few other bits of metadata.

This is OK, except that instead of email we really want to be identifying a user by their numeric ID: #3764.

And what makes this particularly acute is that the payloads of notification-messages from the server have already performed that migration: they don't contain the user's email, only their ID. As a result, if the user is logged into several accounts on the same org, we have #4631 if they try to open a notification from one of those (we're unable to switch to the appropriate account, because we don't know which one it is); and we'll have a similar issue in that case with #4945/#4861 replying to a notification (we won't be able to do it from one of those, because we don't know which one it is.)

So we need to start storing the user ID in `state.accounts`.

The steps for implementing this will look something like:
* Add a property `userId` to the `Account` type. Follow the examples of how we handled `zulipVersion` and `zulipFeatureLevel`. Use a migration (in `src/boot/store.js`) to fill in the value as `null` for existing accounts.
* Each time we complete an initial fetch, we'll learn the user ID then and so we can fill it in. This will come in a `REALM_INIT` action; see `accountsReducer.js` for where we update `state.accounts`, and follow the example of `zulipFeatureLevel`.
* Then we should be able to start using that information, in particular for #4631.

shaurya2612:
Earlier we used to identify users by their emails, making it difficult
to keep track of which user to switch to if multiple accounts are logged
in to the same org when opening the app via a notification.

Since server version 2.1-dev-540-g447a517e6f the server sends the
user ID of the relevant account in the notification. Making it sensible
to store the userId in the redux state.

The `accountsReducer-test.js` file used to expect the random userId
which the account got from the `makeAccount` method, but since a
`REALM_INIT` action changes it now to whatever payload it was supplied
we need to have a constant userId in makeAccount (in this case 4).
fixes #4951


src/account/__tests__/accountsReducer-test.js
```diff
@@ -13,10 +13,11 @@ import accountsReducer from '../accountsReducer';
 import { ZulipVersion } from '../../utils/zulipVersion';
 
 import * as eg from '../../__tests__/lib/exampleData';
+import { makeUserId } from '../../api/idTypes';
 
 describe('accountsReducer', () => {
   describe('REALM_INIT', () => {
-    const account1 = eg.makeAccount();
+    const account1 = eg.makeAccount({ userId: makeUserId(4) });
```
chrisbobbe, line 20: Interesting; why do we need to set `account1`'s user ID to 4?

shaurya2612: Earlier the account Id was unaffected in the redux state, in this change the `REALM_INIT` action modifies it to whatever user ID is provided in the payload. In these test cases the example data provides a user ID of 4 (line 657 of `exampleData.js`) hence the need to keep it 4 while initializing the account. 

In contrast, earlier we used to generate and expect a random user ID but that would always be changed to 4 in this case. 4 could be any other number as long as it is the same in this file and in `exampleData.js`.

gnprice: Hmm, does this test actually depend on using the same ID as in that line of `exampleData`?

If so, we should fix that. In general, we want the assumptions that a test is making to be clear when just reading the test. If it depends on any facts about test data elsewhere, like in `exampleData`, then it should only depend on that data being "reasonable" and boring, not on any arbitrary details.

For example we should probably make `eg.action.realm_init.user_id` be `eg.selfUser.userId`. If we did, it'd be fine for tests to assume that was the case, because in general we make a lot of the `exampleData` data consistent with each other, and that's the only particular value it could have that would be boring and reasonable for it to have. But otherwise, tests shouldn't assume any particular value there at all.

gnprice: Yeah, I see, it does. Here's a fix -- this makes the test not depend on what specific value `exampleData` supplies there, by making the relationship explicit in the code:
```suggestion
    const account1 = eg.makeAccount({ userId: eg.action.realm_init.data.user_id });
```

If we adjusted `exampleData` so it used `selfUser.user_id` there, then an alternative would be:
```js
    const account1 = eg.selfAccount;
```
But the explicit version works in any case.

src/boot/store.js
```diff
@@ -343,6 +343,15 @@ const migrations: {| [string]: (GlobalState) => GlobalState |} = {
     },
   }),
 
+  // Add Accounts.userId, as number | null.
```
gnprice, line 346: This comment isn't right -- the type is `UserId | null`.

gnprice (commented):
Thanks for the revision! Looks good except for the small comment below and the thread above. I'll fix those and merge.

gnprice:
Merged! In addition to the edits mentioned above, I just made some small stylistic tweaks to the jsdoc:
```js
  /**
   * The user's numeric ID.
   *
   * We learn the user ID each time we complete an initial fetch.
   *
   * This is `null` briefly when we've logged in but not yet completed our
   * first initial fetch on the account.  It's also `null` when representing
   * an account which was last used on a version of the app which didn't
   * record this information.
   */
  userId: UserId | null,
```

src/api/initialDataTypes.js
```diff
@@ -99,6 +99,7 @@ export type InitialDataRealm = $ReadOnly<{|
    * https://zulip.com/api/get-server-settings
    */
   zulip_feature_level?: number,
+  user_id?: UserId,
```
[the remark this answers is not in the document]
gnprice: > I should probably go and update that comment

#5110

PR #4973, opened 2021-08-27
ChatScreen: Use fetching instead of caughtUp to refuse sending message.

WesleyAC:
If a user is scrolled up in the message list, it's possible that
caughtUp.newer will be false, but the user knows that, and still wants
the message to be sent.

See CZO:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Failed.20to.20send.20on.20Android/near/1217321


gnprice:
Thanks both!

This logic looks good, but I think the description isn't quite right. I'll revise the comment and commit message, and merge.

gnprice:
Done!

PR #4974, opened 2021-08-27
Fix some libdef issues with AsyncStorage.

chrisbobbe:
The useful part of #4947, which I'll close now.


gnprice:

chrisbobbe:
Thanks for the review!

PR #4976, opened 2021-08-27
ZulipAsyncStorage types [nfc]: Add a quick annotation, for types-first.

chrisbobbe:
The in-progress PR #4974 can help verify that these annotations are appropriate.

Related: #4907


chrisbobbe:
(#4947 was an earlier approach to this.)

gnprice:
Thanks, looks good -- merging!

I went and merged #4974 first, as it sounds like this should be sequenced after that :wink:

PR #4979, opened 2021-08-30
Fix #4904 unread-PMs count blank when too big

### LINKED ISSUE #4904: Unread-PMs count is blank if too big
The count we show on the "people" icon that leads to the PMs tab can look like this:
image
I.e., there's a little bubble there… but instead of having a number inside, it's just blank, and so it looks broken.

I get that when logged in as a test account that has 139 unread PMs. I suspect what's happening involves the fact that the number is too long to really fit in that space.

Instead, when this count is big, we should probably just say something like "99+".

ShristiJain:
Fixes: #4904

### DIFF
--- a/src/chat/UnreadNotice.js
+++ b/src/chat/UnreadNotice.js
@@ -31,10 +31,10 @@ const styles = createStyleSheet({
     paddingRight: 4,
   },
   unreadNumberExceeds: {
-      fontSize: 12,
-      color: 'white',
-      paddingRight: 4,
-    },
+    fontSize: 12,
+    color: 'white',
+    paddingRight: 4,
+  },
   unreadText: {
     fontSize: 14,
     color: 'white',
--- a/src/chat/UnreadNotice.js
+++ b/src/chat/UnreadNotice.js
@@ -30,6 +30,11 @@ const styles = createStyleSheet({
     color: 'white',
     paddingRight: 4,
   },
+  unreadNumberExceeds: {
+      fontSize: 12,
+      color: 'white',
+      paddingRight: 4,
+    },
   unreadText: {
     fontSize: 14,
     color: 'white',
@@ -47,7 +52,7 @@ export default function UnreadNotice(props: Props): Node {
   return (
     <AnimatedScaleComponent visible={unreadCount > 0} style={styles.unreadContainer}>
       <View style={styles.unreadTextWrapper}>
-        <RawLabel style={styles.unreadNumber} text={unreadCount.toString()} />
+        <RawLabel style={unreadCount > 99 ? styles.unreadNumberExceeds : styles.unreadNumber} text={unreadCount > 99 ? '99+' : unreadCount.toString()} />
         <Label
           style={styles.unreadText}
           text={unreadCount === 1 ? 'unread message' : 'unread messages'}


src/chat/UnreadNotice.js
```diff
@@ -30,6 +30,11 @@ const styles = createStyleSheet({
     color: 'white',
     paddingRight: 4,
   },
+  unreadNumberExceeds: {
+      fontSize: 12,
```
chrisbobbe, line 34: It looks like this new `styles.unreadNumberExceeds` is just meant to decrease the font size, from 14 to 12.

Can we find a solution that doesn't require an inconsistent font size? If the problem is that you're running into a maximum width of that `BRAND_COLOR`ed bubble, can we instead remove that constraint, and let the width grow just enough to show "99+" without shrinking the text?

src/chat/UnreadNotice.js
```diff
@@ -30,6 +30,11 @@ const styles = createStyleSheet({
     color: 'white',
     paddingRight: 4,
   },
+  unreadNumberExceeds: {
```
chrisbobbe, line 33:
```
Issue #4904 Solved
```

Please follow our [commit message style](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#git-commits-commit-messages).

chrisbobbe (changes requested):
Thanks, @ShristiJain! Comments below.

src/chat/UnreadNotice.js
ShristiJain: Thankyou for pointing this out. I have made the edit.

chrisbobbe: Ah, I see you've changed the title, but I was talking about your first comment in the PR, [this one](https://github.com/zulip/zulip-mobile/pull/4979#issue-722716437). I think GitHub also calls this the "description" of the PR.

chrisbobbe:
Ah, and I see GitHub is pointing out some failed checks: https://github.com/zulip/zulip-mobile/pull/4979/checks?check_run_id=3465031609

It says `FAILED: prettier`. You should be able to fix that with `tools/test prettier --fix`.

ShristiJain:
hey @chrisbobbe I am new to this open source contribution. It would be great if you could help me with where exactly can I fix this issue of FAILED: prettier.

chrisbobbe:
> hey @chrisbobbe I am new to this open source contribution. It would be great if you could help me with where exactly can I fix this issue of FAILED: prettier.

`FAILED: prettier` means there's some code that isn't formatted correctly. You can also see that output by running the command `tools/test prettier`, or just `tools/test`.

`tools/test prettier --fix` will try to fix the problem. 🙂 What output do you get from that command? Does `git status` show that some formatting corrections have been made?

I think `tools/fmt` would also work.

ShristiJain:
> > hey @chrisbobbe I am new to this open source contribution. It would be great if you could help me with where exactly can I fix this issue of FAILED: prettier.
> 
> `FAILED: prettier` means there's some code that isn't formatted correctly. You can also see that output by running the command `tools/test prettier`, or just `tools/test`.
> 
> `tools/test prettier --fix` will try to fix the problem. 🙂 What output do you get from that command? Does `git status` show that some formatting corrections have been made?
> 
> I think [`tools/fmt`](https://github.com/zulip/zulip-mobile/blob/master/docs/howto/editor.md#other-editors) would also work.

'tools/test prettier' shows 'passed' in my laptop but fails on Github CI. Git status doesn't show any new changes either. What should I do now?

chrisbobbe:
> 'tools/test prettier' shows 'passed' in my laptop but fails on Github CI. Git status doesn't show any new changes either.

It sounds like there may be some commits that you haven't pushed yet. Could you paste the output of `git show`? Currently, I see GitHub showing d7e721175 as the tip of the remote branch. If that's different from what you see on your local branch, you'll probably want to [force-push](https://zulip.readthedocs.io/en/latest/git/using.html#force-push-changes-to-github-after-you-ve-altered-your-history) your latest changes.

ShristiJain:
> > 'tools/test prettier' shows 'passed' in my laptop but fails on Github CI. Git status doesn't show any new changes either.
> 
> It sounds like there may be some commits that you haven't pushed yet. Could you paste the output of `git show`? Currently, I see GitHub showing [d7e7211](https://github.com/zulip/zulip-mobile/commit/d7e721175394853268334d5d4984caf83eb503cb) as the tip of the remote branch. If that's different from what you see on your local branch, you'll probably want to [force-push](https://zulip.readthedocs.io/en/latest/git/using.html#force-push-changes-to-github-after-you-ve-altered-your-history) your latest changes.

image
It shows the same number!

gnprice:
Thanks @ShristiJain for this contribution! I agree with @chrisbobbe's comment above:
https://github.com/zulip/zulip-mobile/pull/4979#discussion_r698715684

Would you please also post screenshots of what this icon looks like when the count is large, both before and after your change?

gnprice:
After #5147, the code this applies to has been replaced. If the issue still exists, the solution will look quite different from this.

PR #4980, opened 2021-08-30
unreadModel: Handle moving messages between topics/streams.

### LINKED ISSUE #4840: unreadModel: respond to message topic/stream edits
We currently don't update unreads when a message topic is edited. We should do so.

See [CZO discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.22No.20messages.20with.20this.20topic.22).

WesleyAC:
Fixes #4840.

### DIFF
--- a/src/unread/__tests__/unreadModel-test.js
+++ b/src/unread/__tests__/unreadModel-test.js
@@ -1,7 +1,11 @@
 /* @flow strict-local */
 import Immutable from 'immutable';
 
-import { ACCOUNT_SWITCH, EVENT_UPDATE_MESSAGE_FLAGS } from '../../actionConstants';
+import {
+  ACCOUNT_SWITCH,
+  EVENT_UPDATE_MESSAGE_FLAGS,
+  EVENT_UPDATE_MESSAGE,
+} from '../../actionConstants';
 import { reducer } from '../unreadModel';
 import { type UnreadState } from '../unreadModelTypes';
 import * as eg from '../../__tests__/lib/exampleData';
@@ -59,6 +63,114 @@ describe('stream substate', () => {
       // prettier-ignore
       expect(summary(reducer(initialState, action, eg.plusReduxState))).toEqual(Immutable.Map([
         [message1.stream_id, Immutable.Map([[message1.subject, [1, 2]]])],
+      ]));
+    });
+  });
+
+  describe('EVENT_UPDATE_MESSAGE', () => {
+    const mkAction = args => {
+      const {
+        message_ids,
+        stream_id = 123,
+        new_stream_id = undefined,
+        orig_subject = 'foo',
+        subject = 'foo',
+      } = args;
+      return {
+        id: 1,
+        type: EVENT_UPDATE_MESSAGE,
+        user_id: eg.selfUser.user_id,
+        message_id: message_ids[0],
+        message_ids,
+        edit_timestamp: 10000,
+        stream_id,
+        new_stream_id,
+        propagate_mode: 'change_later',
+        orig_subject,
+        subject,
+        subject_links: [],
+        orig_content: '',
+        orig_rendered_content: '',
+        prev_rendered_content_version: 0,
+        content: '',
+        rendered_content: '',
+        is_me_message: false,
+        flags: [],
+      };
+    };
+
+    const baseState = (() => {
+      const streamAction = args => eg.mkActionEventNewMessage(eg.streamMessage(args));
+      const r = (state, action) => reducer(state, action, eg.plusReduxState);
+      let state = initialState;
+      state = r(state, streamAction({ stream_id: 123, subject: 'foo', id: 1 }));
+      state = r(state, streamAction({ stream_id: 123, subject: 'foo', id: 2 }));
+      state = r(state, streamAction({ stream_id: 123, subject: 'foo', id: 3 }));
+      state = r(state, streamAction({ stream_id: 123, subject: 'foo', id: 4 }));
+      state = r(state, streamAction({ stream_id: 456, subject: 'zzz', id: 6 }));
+      state = r(state, streamAction({ stream_id: 456, subject: 'zzz', id: 7 }));
+      state = r(state, streamAction({ stream_id: 123, subject: 'foo', id: 15 }));
+      return state;
+    })();
+
+    test('(base state, for comparison)', () => {
+      // prettier-ignore
+      expect(summary(baseState)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2, 3, 4, 15]]])],
+        [456, Immutable.Map([['zzz', [6, 7]]])],
+      ]));
+    });
+
+    test('if topic/stream not updated, return original state', () => {
+      const state = reducer(baseState, mkAction({ message_ids: [5] }), eg.plusReduxState);
+      expect(state.streams).toBe(baseState.streams);
+    });
+
+    test('if topic updated, but no unreads, return original state', () => {
+      const state = reducer(
+        baseState,
+        mkAction({ message_ids: [100], orig_subject: 'foo', subject: 'bar' }),
+        eg.plusReduxState,
+      );
+      expect(state.streams).toBe(baseState.streams);
+    });
+
+    test('if topic updated, move unreads', () => {
+      const state = reducer(
+        baseState,
+        mkAction({ message_ids: [3, 4, 15], orig_subject: 'foo', subject: 'bar' }),
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]], ['bar', [3, 4, 15]]])],
+        [456, Immutable.Map([['zzz', [6, 7]]])],
+      ]));
+    });
+
+    test('if stream updated, move unreads', () => {
+      const state = reducer(
+        baseState,
+        mkAction({ message_ids: [3, 4, 15], new_stream_id: 456 }),
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]]])],
+        [456, Immutable.Map([['zzz', [6, 7]], ['foo', [3, 4, 15]]])],
+      ]));
+    });
+
+    test('if moved to topic with existing unreads, ids stay sorted', () => {
+      const state = reducer(
+        baseState,
+        mkAction({ message_ids: [3, 4, 15], new_stream_id: 456, subject: 'zzz' }),
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]]])],
+        [456, Immutable.Map([['zzz', [3, 4, 6, 7, 15]]])],
       ]));
     });
   });
--- a/src/unread/unreadModel.js
+++ b/src/unread/unreadModel.js
@@ -18,6 +18,7 @@ import {
   ACCOUNT_SWITCH,
   EVENT_MESSAGE_DELETE,
   EVENT_NEW_MESSAGE,
+  EVENT_UPDATE_MESSAGE,
   EVENT_UPDATE_MESSAGE_FLAGS,
   LOGOUT,
   MESSAGE_FETCH_COMPLETE,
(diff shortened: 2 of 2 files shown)


src/actionTypes.js
```diff
@@ -344,6 +344,9 @@ type EventUpdateMessageAction = {|
   type: typeof EVENT_UPDATE_MESSAGE,
   edit_timestamp: number,
   message_id: number,
+  message_ids: number[],
+  stream_id: number,
+  new_stream_id?: number,
```
gnprice, line 349: Let's accompany this kind of change by having a comment at the top of the event link to the relevant bit of docs:
https://zulip.com/api/get-events#update_message
That helps a lot in reviewing a change here, or in comparing in the future to the possibly-by-then-further-changed API. It also makes it easy to look up what each of these properties is documented as being supposed to mean.

We mostly don't have those now (in fact it looks like there's just one such example, `RestartEvent` in `eventTypes.js` -- which not coincidentally looks like the latest event we've added to our types), mainly because the bulk of these type definitions were written down in the reverse-engineering era, when there were no API docs to link to.

src/actionTypes.js
```diff
@@ -344,6 +344,9 @@ type EventUpdateMessageAction = {|
   type: typeof EVENT_UPDATE_MESSAGE,
   edit_timestamp: number,
   message_id: number,
+  message_ids: number[],
```
gnprice, line 347: I just mentioned linking to the API docs, and I had the thought that that substitutes for writing down any documentation here about what the properties mean... but as I read this event's docs, I see that this `message_id` vs. `message_ids` thing is odd enough, and enough of a potential gotcha, that it probably merits making an exception and writing something down here.

In particular, any content changes apply to just `message_id`, and any stream or topic changes apply to all of `message_ids`, which is guaranteed to include `message_id`.

src/actionTypes.js
```diff
@@ -344,6 +344,9 @@ type EventUpdateMessageAction = {|
...
+  stream_id: number,
```
gnprice, line 348: This should be optional, according to the docs. (It's there only when the message was sent to a stream.)

src/actionTypes.js
gnprice: The docs also kind of make it sound like `message_ids` isn't always present -- emphasis added:
> *If the stream or topic was changed*, the set of moved messages is encoded in the separate message_ids field, which is guaranteed to include message_id.

even though the `message_ids` bit doesn't say it's optional. That might be good to clarify in chat on #api-design.

gnprice (commented):
Thanks! Comments below on the first commit. The second commit is more interesting, and I'll return to it after looking at some other things, but wanted to get these out without waiting :slightly_smiling_face:

src/unread/unreadModel.js
```diff
@@ -200,6 +201,31 @@ function streamsReducer(
...
+        (action.orig_subject !== undefined && action.subject !== action.orig_subject)
+        || (action.new_stream_id !== undefined && action.new_stream_id !== action.stream_id)
+      ) {
+        const moved_ids = new Set(action.message_ids);
+        const updated_unreads = state
+          .getIn([action.stream_id, action.orig_subject ?? action.subject], Immutable.List())
+          .filter(id => moved_ids.has(id));
+
+        if (updated_unreads.size > 0) {
+          return state
```
gnprice, line 215: These conditionals (this, and the `if` above) have the feeling that the main line of the code's logic gets nested deeper and deeper as the code goes on. This pattern also has the effect that what happens in the alternative case (here, when `updated_unreads` is empty) is left in suspense for the reader until after all the other code.

I tend to prefer the "early return" pattern for arranging code flow: exceptional cases like errors and fast-path shortcuts go off to the side and then return, so that the main narrative of the code gets to continue vertically all in the same column. This also means that the condition for each exceptional case, and what we do in that case, get to stay right next to each other.

I can try demonstrating that rearrangement for this code at merge time, or after merge.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,95 @@ describe('stream substate', () => {
...
+
+    test('if topic not updated, return original state', () => {
+      const state = reducer(baseState, mkAction({ message_id: 5 }), eg.plusReduxState);
+      expect(state.streams).toBe(baseState.streams);
+    });
+
+    test('if topic updated, but no unreads, return original state', () => {
+      const state = reducer(baseState, mkAction({ message_id: 100 }), eg.plusReduxState);
+      expect(state.streams).toBe(baseState.streams);
+    });
```
gnprice, line 121: Does this test match its description? It doesn't look like a topic is updated.

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,95 @@ describe('stream substate', () => {
...
+          message_id: 3,
+          message_ids: [3, 4, 5],
+          new_stream_id: 456,
+        }),
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]]])],
+        [456, Immutable.Map([['zzz', [6, 7]], ['foo', [3, 4, 5]]])],
```
gnprice, line 154: Let's also have a test case where they get moved to an existing topic (like 456 > 'zzz'), showing that the list gets properly combined with the existing one.

gnprice (commented):
And read the main commit. The logic generally looks good! Small comments below.

One commit-message nit:

    unreadModel: Handle moving messages between topics/streams.
    
    Fixes #4840.

See https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#fixes-format .

chrisbobbe:
FYI I've just sent #5155; it might be convenient for that to be reviewed/merged before this.

src/actionTypes.js
chrisbobbe: Confirmed in [chat](https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1293909) that we do expect it to be always present.

gnprice:
I've just pushed a revision of this PR, rebasing atop the API types from #5155 and applying my comments above. @chrisbobbe please take a look!

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
     });
   });
 
+  describe('EVENT_UPDATE_MESSAGE', () => {
+    const mkAction = args => {
+      const {
+        message_ids,
+        stream_id = 123,
```
chrisbobbe, line 74: I think this default parameter means that `mkAction`'s callers can't request an action that omits `stream_id`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#passing_undefined_vs._other_falsy_values

Whether those callers will end up wanting to do that will depend on whether servers want to convey something relevant by omitting `stream_id`, I guess. You've written down the state of that question in `unreadModel`:

```js
      if (stream_id == null) {
        // Not stream messages, or else a pure content edit (no stream/topic change.)
        //
        // As of this writing, the docs actually promise this field for all
        // updates to stream messages.  Empirically it's absent on pure
        // content edits, but present on edits affecting either the stream
        // or topic:
        //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1296823
        return state;
      }
```

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
...
+
+    test('(base state, for comparison)', () => {
+      // prettier-ignore
+      expect(summary(baseState)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2, 3, 4, 15]]])],
+        [456, Immutable.Map([['zzz', [6, 7]]])],
+      ]));
+    });
+
+    test('if topic not updated, return original state', () => {
```
chrisbobbe, line 124:
```suggestion
    test('if topic/stream not updated, return original state', () => {
```

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
...
+      const state = reducer(baseState, mkAction({ message_ids: [5] }), eg.plusReduxState);
+      expect(state.streams).toBe(baseState.streams);
+    });
+
+    test('if topic updated, but no unreads, return original state', () => {
```
chrisbobbe, line 129:
```suggestion
    test('if topic updated (and not stream), but no unreads, return original state', () => {
```

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
...
+      const state = reducer(
+        baseState,
+        mkAction({ message_ids: [100], orig_subject: 'foo', subject: 'bar' }),
+        eg.plusReduxState,
+      );
+      expect(state.streams).toBe(baseState.streams);
+    });
+
+    test('if topic updated, move unreads', () => {
```
chrisbobbe, line 138:
```suggestion
    test('if topic updated (and not stream), move unreads', () => {
```

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
...
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]], ['bar', [3, 4, 15]]])],
+        [456, Immutable.Map([['zzz', [6, 7]]])],
+      ]));
+    });
+
+    test('if stream updated, move unreads', () => {
```
chrisbobbe, line 151:
```suggestion
    test('if stream updated (and not topic), move unreads', () => {
```

src/unread/__tests__/unreadModel-test.js
```diff
@@ -63,6 +67,114 @@ describe('stream substate', () => {
...
+        eg.plusReduxState,
+      );
+      // prettier-ignore
+      expect(summary(state)).toEqual(Immutable.Map([
+        [123, Immutable.Map([['foo', [1, 2]]])],
+        [456, Immutable.Map([['zzz', [6, 7]], ['foo', [3, 4, 15]]])],
+      ]));
+    });
+
+    test('if moved to existing topic, ids stay sorted', () => {
```
chrisbobbe, line 164:
```suggestion
    test('if stream/topic both updated, and stream/topic thread has existing unreads, ids stay sorted', () => {
```

src/unread/unreadModel.js
```diff
@@ -203,6 +204,47 @@ function streamsReducer(
...
+        // content edits, but present on edits affecting either the stream
+        // or topic:
+        //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1296823
+        return state;
+      }
+
+      if (
+        (action.subject === action.orig_subject || action.orig_subject == null)
+        && (action.new_stream_id === stream_id || action.new_stream_id == null)
+      ) {
```
chrisbobbe, line 223: I think a check like this makes a lot of sense, given that we don't have confirmation on what we've observed empirically; see above:

```
        // As of this writing, the docs actually promise this field for all
        // updates to stream messages.  Empirically it's absent on pure
        // content edits, but present on edits affecting either the stream
        // or topic:
        //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1296823
```

And even if we did, this is a nice and intuitive conditional, even though I guess it would be kind of belt-and-suspenders, right.

chrisbobbe (commented):
Thanks! See a few small comments/suggestions below. Let me know if I can be clearer on any of them.

src/unread/__tests__/unreadModel-test.js
gnprice: True. At present this test file is just for the "streams" part of the unreads data, because that's the part that has been modernized. In the future when we similarly modernize the rest and have its tests in the same place, we'll want to either add a different helper for those tests or modify this one.

src/unread/unreadModel.js
gnprice: Updated that above comment to reflect the further discussion in that thread: we plan to make `stream_id` present on all updates to stream messages. Then this check will be directly needed in order to skip pure content edits.

gnprice:
Thanks for the review!

I applied some of the test-name suggestions, but left out the "(and not stream)" ones; I think it's generally fine for the test names to highlight the thing that does happen and leave implicit that other things that would be interesting don't happen.

The revision is only in comments and test names, so going ahead and merging.

src/unread/__tests__/unreadModel-test.js
chrisbobbe: Makes sense. Also, I see from your other comment that we've settled that servers will start always including `stream_id` for stream messages, so there's no reason for this modernized streams-focused part of the model to be tested with an event that's missing `stream_id`.

PR #4982, opened 2021-08-31
Fix react-intl libdef, on the path to RN v0.64 upgrade.

chrisbobbe:
When we upgrade Flow for the RN v0.64 upgrade (#4426), Flow understandably starts complaining about the `React_2` stuff in this libdef. We've had a TODO to deal with it; so, do that. From my experimentation so far, these changes should be enough to satisfy those complaints from the new Flow version with RN v0.64.

See discussion at https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-intl.20libdef/near/1228275.

### DIFF
--- a/flow-typed/react-intl_vx.x.x.js
+++ b/flow-typed/react-intl_vx.x.x.js
@@ -1,14 +1,9 @@
 // `react-intl`, translated from the relevant-looking TypeScript .d.ts
 // file, with minimal tweaks to make it work.
 
-// Anything at `React_2.` is `any`; see
-// https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/libdef.3A.20react-native-webview/near/896571.
-// We should use Flow's builtin React type annotations, like
-// React$Node instead of React.Node.
-
-// Not sure why the TypeScript uses React_2 instead of just React, but
-// we've kept it intact.
-import * as React_2 from 'react';
+// This used to try to use types from React's TypeScript libdef,
+// @types/react. We've translated those places to use Flow instead,
+// hopefully without errors or too much loss of information.
 
 // react-intl/react-intl.d.ts
 declare module 'react-intl' {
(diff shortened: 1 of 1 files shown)


flow-typed/react-intl_vx.x.x.js
```diff
@@ -1,14 +1,9 @@
...
-// https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/libdef.3A.20react-native-webview/near/896571.
-// We should use Flow's builtin React type annotations, like
-// React$Node instead of React.Node.
-
-// Not sure why the TypeScript uses React_2 instead of just React, but
-// we've kept it intact.
-import * as React_2 from 'react';
+// This used to try to use types from React's TypeScript libdef,
+// @types/react. We've translated those places to use Flow instead,
+// hopefully without errors or too much loss of information.
```
gnprice, line 6: Thanks for doing this!

It's too bad that (as you say in the commit message) imports don't work inside Flow libdefs. That would simplify a good bit of this.

Probably an ideal version of Flowgen would work around that by copying in the relevant definitions from the referenced other libdef.

flow-typed/react-intl_vx.x.x.js
```diff
@@ -328,8 +323,14 @@ declare module 'react-intl' {
...
-  ): React_2.ForwardRefExoticComponent<
-    React_2.PropsWithoutRef<WithIntlProps<React_2.PropsWithChildren<P>>> & React_2.RefAttributes<T>, > & {
+  ): React$AbstractComponent<
+    {|
+      ...WithIntlProps<P>,
+      children?: React$Node,
+      ref?: React$Ref<T>,
+    |},
+    mixed,
+  > & {
```
gnprice, line 333: This looks like it reflects another missing feature (not to say it'd necessarily be an easy one to build…) in Flowgen: the TS type here uses `&`, presumably reflecting inexact object types on the TS side, but we have exact object types here and so they really need to be spreads.

gnprice (commented):
Thanks for taking care of this! Looks good -- merging.

chrisbobbe:
Thanks for the review!

PR #4983, opened 2021-09-01
Use PR branch for camera-roll library, on the path to RN v0.64.

chrisbobbe:


chrisbobbe:
Thanks for the review!

gnprice:
Done. I also added a sentence about the other thing I checked before deciding this looked good:

    So instead, the PR just makes all the implicit inexact object types
    explicitly inexact.  It's based directly on v4.0.4, the version we've
    been using, so it doesn't pull in any other changes.

PR #4985, opened 2021-09-01
flow: Add missing error codes in some new suppressions.

chrisbobbe:
When we bump the Flow version along with the RN v0.64 upgrade (#4426), Flow will enforce having all of our Flow error suppressions indicate which specific error code they're meant to suppress. These few snuck in after we'd gone through the whole codebase and added those indications.


gnprice:
Thanks! Merging.

PR #4987, opened 2021-09-03
flow: Enable Types-First mode, on the path to RN v0.64 upgrade.

chrisbobbe:
From Flow's perspective, we're fine to delay types-first until we're
on v0.143, which we'd normally take along with the RN v0.65 upgrade.

But, from experimentation, React Native v0.64's internal code seems
to have been written with Types-First, such that a few Flow errors
would show up if we tried to use RN v0.64 without Types-First
enabled. So, enable it, now that we can, and now that the RN v0.64
upgrade is on the horizon (#4426).

Fixes: #4907

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -113,5 +113,8 @@ module.file_ext=.js
 module.file_ext=.json
 module.file_ext=.ios.js
 
+; We can remove this on Flow v0.134 or later, when it's on by default.
+types_first=true
+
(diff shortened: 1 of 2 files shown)


gnprice:
:confetti_ball: Thanks! Looks good -- merging.

chrisbobbe:
Yay, thanks for the review!

PR #4988, opened 2021-09-03
@react-navigation/stack libdef: Fix setParams by dropping $Shape.

gnprice:
This is a followup to the small type-safety regression in #4987.

See comment.  The underlying Flow bug is that $Shape doesn't
really work:
  https://github.com/facebook/flow/issues/5702

Happily there is a workaround, namely `$Rest<T, {...}>`.
One might give that a handy alias, like `$Partial<T>` -- and
that's exactly what this libdef already does.  So just use it here.

Also add a type-test.  The most natural place for this might be a
folder like `flow-typed/__tests__/`, next to the libdefs themselves…
but if we do that, then every time we touch the file, Flow restarts
from scratch, just as it does for actual libdefs in `flow-typed/`.
Seems like something odd there.  But we can work around it by just
using a separate folder.

### DIFF
--- a/__libdef-tests__/stack_v5.x.x-test.js
+++ b/__libdef-tests__/stack_v5.x.x-test.js
@@ -1,0 +1,36 @@
+/* @flow strict-local */
+import { type NavigationProp } from '@react-navigation/stack';
+
+/* eslint-disable no-unused-vars */
+
+function test_setParams() {
+    type NavProp<P> = NavigationProp<{| r: P |}, 'r'>;
+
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks! LGTM, except let's put the specific error code in the `$FlowExpectedError` suppressions, which will be required when we upgrade Flow soon. Please feel free to merge after that.

gnprice:
Good catch, thanks! Merged with that change.

PR #4989, opened 2021-09-04
deps: Remove two peer-deps we don't need anymore (helpful for RN v0.64 upgrade).

chrisbobbe:
We added these in 62621ef8e, so we could add `jest-expo`.
Empirically, it's fine to remove these and keep `jest-expo`; we
don't get any peer-dep warnings after doing so.

(Inheriting priority from #4426, the issue for the RN v0.64 upgrade.)


gnprice:
Thanks! Looks good, merging.

Issue #4866, opened 2021-07-06
Add link to user profile from sender's name


alya:
In the message feed, a user's avatar opens the profile menu, but the sender's name is currently not a link. We should make the name link to the profile menu as well, which is likely what the user expects, and is consistent with the web app UI.

jay719:
@zulipbot claim

SilentCruzer:
Can I work on this issue?

jay719:
Yes please take it away from me, my computer has not been working for a
while

chrisbobbe:
(Discussion has continued [on CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Edit.20WebView/near/1252673).)

PR #4990, opened 2021-09-07
Add link to user profile from sender's name

SilentCruzer:
Fixes: #4866 

Changes:
- The first commit was just to rename the web outbound event from avatar to request-user-profile since we are going to use it on multiple events.
- The second commit added a new condition when the username is pressed. 'UserId' is required to navigate a component to the profile page. In the avatar-img tag, there was already an attribute names 'data-sender-id, but for the username tag, there was no such attribute. So used .closest() function to get the avatar tag and hence get the id.

-


src/webview/js/js.js
```diff
@@ -867,6 +867,19 @@ documentBody.addEventListener('click', (e: MouseEvent) => {
     return;
   }
 
+  if (target.matches('.username')) {
+    const doc = target.closest('.message');
+    if (doc != null) {
+      sendMessage({
+        type: 'request-user-profile',
+        fromUserId: makeUserId(
+          requireNumericAttribute(doc.getElementsByClassName('avatar-img')[0], 'data-sender-id'),
```
WesleyAC, line 876: Rather than finding the `avatar-img` here, I'd add a `data-sender-id` attribute to the username element, and then you can just change the above if statement to `if (target.matches('.avatar-img') || target.matches('.username')) {`.

WesleyAC (changes requested):
Thanks! This looks good overall — one comment about the code, and one comment on the commit message: `webview[nfc]: Renamed Avatar Outbound event` should be instead something like `webview[nfc]: Rename WebViewOutboundEventAvatar`. That's in the present tense (`Rename` instead of `Renamed`), and the thing that you're renaming is a little more accurate.

src/webview/js/js.js
SilentCruzer: I have added the suggested changes. Added the data-sender-id attribute in the messageAsHtml.js file. Also added updated snapshots for the tests.

SilentCruzer:
Thanks for the feedback, I have made the suggested changes. Added the data-sender-id attribute in the messageAsHtml.js file. Also added updated snapshots for the tests.

src/webview/handleOutboundEvents.js
```diff
@@ -55,8 +55,8 @@ type WebViewOutboundEventScroll = {|
   endMessageId: number,
 |};
 
-type WebViewOutboundEventAvatar = {|
-  type: 'avatar',
+type WebViewOutboundEventRequestProfile = {|
```
chrisbobbe, line 58: Since the string is 'request-user-profile', I would probably name this Flow type alias `WebViewOutboundEventRequestUserProfile`. It's pretty hilariously long, but at least it's less ambiguous that they're meant to go together.

chrisbobbe (changes requested):
Thanks, @SilentCruzer! See one small comment below; also, it looks like this PR has gathered some rebase conflicts. Would you mind rebasing across those?

gnprice:
Thanks @SilentCruzer, and thanks @WesleyAC and @chrisbobbe for the reviews! This looks good -- merging.

I've made a couple of small fixes to the commits. The main one is:
* The type was getting renamed to `WebViewOutboundEventRequestProfile` in the first commit, and then to `WebViewOutboundEventRequestUserProfile` in the second commit, which otherwise isn't related to the rename. I edited the first commit to rename it all the way to the desired name (using `git rebase -i`; edit the first commit; `git checkout -p main`; select the desired changes; `git rebase --continue`.)

PR #4995, opened 2021-09-08
RN v0.64 upgrade followups

chrisbobbe:
I think these are all the planned followups, but I could be forgetting one or two. These just remove workarounds that are no longer necessary.

Related: #4426

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -25,13 +25,6 @@ node_modules/react-native/Libraries/polyfills/.*
 [untyped]
 .*/node_modules/@react-native-community/cli/.*/.*
 
-; TODO: Remove this line when RN v0.64 upgrade is complete. We took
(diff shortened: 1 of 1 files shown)


src/common/LoadingBanner.js
```diff
@@ -7,7 +7,7 @@ import { View } from 'react-native';
 import { useSelector } from '../react-redux';
 import { getLoading } from '../selectors';
 // eslint-disable-next-line import/no-useless-path-segments
```
gnprice, line 9: This line can probably disappear too, right?

gnprice (commented):
Thanks! Looks good, modulo a tiny comment below. Otherwise please merge at will.

src/common/LoadingBanner.js
chrisbobbe: Yep, thanks!

chrisbobbe:

PR #4996, opened 2021-09-08
Set up Sentry for Android-native code

gnprice:
This means we'll start seeing reports in Sentry of both uncaught exceptions, and error or warning cases we explicitly catch, in our Kotlin code just as we do already in our RN-based JS code.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -87,5 +87,12 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
+
+        <!-- To make a build that reports errors to Sentry, fill in the Sentry DSN below
+             and uncomment.  See src/sentryConfig.js for another site and more discussion. -->
+        <!--
+        <meta-data android:name="io.sentry.auto-init" android:value="true" tools:node="replace" />
+        <meta-data android:name="io.sentry.dsn" android:value="https://123@sentry.example/45" />
+        -->
     </application>
 </manifest>
--- a/android/app/src/main/java/com/zulipmobile/SentryUtils.kt
+++ b/android/app/src/main/java/com/zulipmobile/SentryUtils.kt
@@ -1,0 +1,26 @@
+package com.zulipmobile
+
+import io.sentry.Sentry
+import io.sentry.SentryLevel
+
+/**
+ * A home for things that ought to be static extensions of `Sentry`.
+ *
+ * Extending Java classes with static members isn't currently a feature
+ * available in Kotlin:
+ *   https://youtrack.jetbrains.com/issue/KT-11968
+ * so this is our substitute.
+ */
+class SentryX {
+    companion object {
+        /**
+         * Like `Sentry.captureException`, but at level `SentryLevel.WARNING`.
+         */
+        public fun warnException(e: Throwable) {
+            Sentry.withScope { scope ->
+                scope.level = SentryLevel.WARNING
+                Sentry.captureException(e)
+            }
+        }
+    }
+}
--- a/android/app/src/main/java/com/zulipmobile/ZLog.kt
+++ b/android/app/src/main/java/com/zulipmobile/ZLog.kt
@@ -1,0 +1,31 @@
+package com.zulipmobile
+
+import android.util.Log
+import io.sentry.Sentry
+
+/**
+ * Zulip-specific logging helpers.
+ *
+ * These mirror part of the interface of `android.util.Log`, but they log
+ * to Sentry as well as to the device console.
+ *
+ * We basically always want to use these instead of plain `Log.e` or `Log.w`.
+ */
+class ZLog {
+    companion object {
+        /** Log an error to both Sentry and the device log. */
+        public fun e(tag: String, e: Throwable) {
+            // Oddly there is no `Log.e` taking just a throwable, like there is for `Log.w`.
+            // Have a message that just repeats the first line of how the throwable prints.
+            val msg = "${e.javaClass.name}: ${e.message}"
+            Log.e(tag, msg, e)
+            Sentry.captureException(e)
+        }
+
+        /** Log a warning to both Sentry and the device log. */
+        public fun w(tag: String, e: Throwable) {
+            Log.w(tag, e)
+            SentryX.warnException(e)
+        }
+    }
+}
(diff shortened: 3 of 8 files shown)


WesleyAC (commented):
I would do the `Log.w` in the helper, rather than manually. Other than that, looks good!

gnprice:
Thanks for the review! Yeah, a good improvement to this code would be to unify things so that each error handler only has to make one call instead of two. Our `logging.js` functions provide something a lot like that. I may try making that change.

@chrisbobbe also pointed out (in the office yesterday) that we already have been getting some Sentry reports that look like they come from the Android-native side. For example here's one with a Java stack trace:
https://sentry.io/organizations/zulip/issues/1964700767/events/71cabef082ab476b9d55d6d11b2393f0/
And here's one reporting a segfault in libjsc, the JS implementation that RN supplies on Android:
https://sentry.io/organizations/zulip/issues/1968181388/events/1e76508e14aa44b5dd9686ac30fdc037/

I haven't actually done a negative test of this: I didn't try reporting a warning, or inducing an exception, before this "Enable Sentry" commit. I *suspect* what's happening is:
* In the existing code, once we initialize Sentry from the JS side, the Sentry RN implementation also goes and initializes the Sentry Android or Sentry iOS implementation. That'd mean we're already covered… as long as the JS has already started running.
* In order to have Sentry working when we're processing an incoming notification, though, as we want for #4967, we'll still need to make this change.
* We probably also still need to initialize Sentry on the JS side (even on Android), in order to have it working there. Hopefully the redundancy is harmless.

I'll do some investigation to confirm.

gnprice:
OK, some findings:
* If I add a DSN in `sentryConfig.js` but not in `AndroidManifest.xml`, so that we initialize Sentry from JS only -- just as we do in the status quo for published releases -- then:
  * `Sentry.captureMessage` in Java/Kotlin works fine after the JS has started up. In particular I added it in `TextCompressionModule#decompress`, and it showed up.
  * OTOH it has no effect before the JS has started up. In particular I tried `MainActivity#onCreate`.
  * An uncaught JVM exception doesn't get reported to Sentry, even after the JS has started up. I tried the same places as for `Sentry.captureMessage`.
* OTOH if I add a DSN in both places, following the instructions in this branch, then:
  * `Sentry.captureMessage` in Java/Kotlin now works fine both before and after the JS has started up.
  * An uncaught exception from `MainActivity#onCreate` now gets reported just fine.
  * … Hmm, but an uncaught exception in `TextCompressionModule#decompress` still doesn't!

    So that seems like a limitation of the Sentry library for React Native. That method is one that's exposed to JS via the React Native API, with the `@ReactMethod` annotation. An uncaught exception there causes a red screen in dev; in prod I expect it would cause a crash.

    In any case, that's not the situation that this PR is concerned with. AFAIK we've never had a bug of this form (and an uncaught exception in one of these methods is always a bug, in the method or something it calls -- exceptions should be turned into promise rejections, in order to propagate properly to the JS caller), so it's not a priority.

gnprice:
> OTOH if I add a DSN in both places, following the instructions in this branch, then: […]
> * An uncaught exception from `MainActivity#onCreate` now gets reported just fine.
> * … Hmm, but an uncaught exception in `TextCompressionModule#decompress` still doesn't!
>
>   So that seems like a limitation of the Sentry library for React Native. That method is one that's exposed to JS via the React Native API, with the `@ReactMethod` annotation. An uncaught exception there causes a red screen in dev; in prod I expect it would cause a crash.

Looks like this is probably https://github.com/getsentry/sentry-react-native/issues/1102 . The person reporting that issue says it was limited to a debug build, and worked fine in a release build. (Easy to imagine how that could come about, with the red-screen mechanism interfering somehow.)

If that's the case then this is… still an annoying Sentry bug that they should really fix, because it causes confusion for developers, but it doesn't have a direct effect on error reporting in the field, which is nice.

chrisbobbe:
Great, thanks for that investigation! I'm satisfied with it.

gnprice:
Pushed a revision with `Log.e` and `Log.w` folded inside new wrappers `ZLog.e` and `ZLog.w` that cover both the device log and Sentry, as well as with commit-message edits to reflect that investigation. Please take a look!

android/app/src/main/AndroidManifest.xml
```diff
@@ -87,5 +87,12 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
+
```
chrisbobbe, line 90: (IIUC)

I think it makes sense that https://github.com/getsentry/sentry-react-native/issues/1102 would be the issue with `TextCompressionModule#decompress`, and that that report would be accurate about the problem being debug-mode-only.

```
This should immediately mean that we start reporting to Sentry when
an exception propagates uncaught and crashes the app.
```

I wonder if this deserves an asterisk on the word "should", though, just in case we find an uncaught exception not being reported to Sentry in the future? Maybe just a link to https://github.com/zulip/zulip-mobile/pull/4996#issuecomment-916501317.

android/app/src/main/java/com/zulipmobile/SentryUtils.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+class SentryX {
+    companion object {
+        /**
+         * Like `Sentry.captureException`, but at level `SentryLevel.WARNING`.
+         */
+        public fun warnException(e: Throwable) {
+            Sentry.withScope { scope ->
+                scope.level = SentryLevel.WARNING
+                Sentry.captureException(e)
```
chrisbobbe, line 22: Interesting, I think I'd expect that you'd have to pass the `scope` param somewhere in the `Sentry.captureException` call. But I guess you don't: https://docs.sentry.io/platforms/java/enriching-events/scopes/#local-scopes

chrisbobbe (approved):
LGTM, thanks! Feel free to merge after seeing a small thing or two below.

android/app/src/main/java/com/zulipmobile/SentryUtils.kt
gnprice: Yeah. I think this API actually makes more sense than it would if you did have to pass it: basically the point is that there's always a current scope (there has to be, to handle uncaught exceptions where you don't explicitly invoke Sentry at all), and so `Sentry.captureException` uses that.

The `withScope` function is sort of doing two things at once: it pushes a new scope onto the scope stack (and undertakes to pop it at the end), and then it passes you the scope much like `Sentry.configureScope` does.  I think that's a bit confusing, and if I were designing this API, I'd probably separate those, so that you'd write:
```kt
Sentry.pushScope {
  Sentry.configureScope { scope -> scope.level = SentryLevel.WARNING }
  Sentry.captureException(e)
}
```

(Well, and I'd probably make `configureScope` a bit less cumbersome too, like:
```kt
Sentry.pushScope {
  Sentry.currentScope.level = SentryLevel.WARNING
  Sentry.captureException(e)
}
```
)

android/app/src/main/AndroidManifest.xml
gnprice: Sure; added a note about that.

PR #4997, opened 2021-09-09
reactUtils tests: Start testing `useHasStayedTrueForMs`.

chrisbobbe:
This is our first use of `react-test-renderer`. It piggy-backs on
our incorporation of Jest's "modern" fake-timer implementation in
PRs #4754 and #4931. That was handy!

I haven't yet found any test cases that fail with our
implementation. (And I'd been hoping to, to debug an unexpected
error!)

But I did try pasting in an earlier iteration of the hook's
implementation, from #4940, that Greg had found bugs in by reading
the code. Many of these tests failed on that buggy implementation,
which is a good sign.

Might as well keep these new tests, then, if they're not an
unreasonable maintenance burden.

### DIFF
--- a/src/__tests__/reactUtils-test.js
+++ b/src/__tests__/reactUtils-test.js
@@ -1,0 +1,246 @@
+/* @flow strict-local */
+import React from 'react';
+import type { ComponentType } from 'react';
+// $FlowFixMe[untyped-import]
+import { create, act } from 'react-test-renderer';
+
+import { fakeSleep } from './lib/fakeTimers';
+import { useHasStayedTrueForMs } from '../reactUtils';
+
+describe('useHasStayedTrueForMs', () => {
+  /**
+   * Simulate a mock component using the hook, and inspect hook's value.
+   *
+   * - The constructor mounts the component. (`ms` won't change through the
+   *   lifetime of the TestMachine instance.)
+   * - Use `updateValue` to change the `value` arg passed to the hook.
+   * - Use `hookOutput` to get the return value of the hook from the latest
+   *   render.
+   * - (Important) To wait for a duration, use the instance's `sleep` method
+   *   instead of the util `fakeSleep` or similar. It wraps some
+   *   `react-test-renderer` boilerplate.
+   * - When done, call 'cleanup'.
+   *
+   * Encapsulates a few things:
+   * - react-test-renderer is untyped (so far)
+   * - boilerplate for using react-test-renderer, like calling `act`
+   *   repeatedly
+   * - boring details like how the mock component is implemented
+   */
+  // I'm not totally clear on everything `act` does, but react-test-renderer
+  // seems to recommend it strongly enough that we actually get errors if we
+  // don't use it. Following links --
+  //   https://reactjs.org/docs/test-renderer.html#testrendereract
+  //   https://reactjs.org/docs/test-utils.html#act
+  //   https://reactjs.org/docs/testing-recipes.html
+  // -- I see the following, which I think does the best job of explaining.
+  // (The `act` in `react-dom/test-utils` might not be identical to the
+  // `act` in `react-test-renderer`, but `react-test-renderer` says they're
+  // similar.)
+  // > When writing UI tests, tasks like rendering, user events, or data
+  // > fetching can be considered as “units” of interaction with a user
+  // > interface. `react-dom/test-utils` provides a helper called `act()`
+  // > that makes sure all updates related to these “units” have been
+  // > processed and applied to the DOM before you make any assertions
+  class TestMachine {
+    static HOOK_VALUE_TRUE = 'HOOK_VALUE_TRUE';
+    static HOOK_VALUE_FALSE = 'HOOK_VALUE_FALSE';
+
+    _TestComponent: ComponentType<{| value: boolean |}>;
+    _testRenderer: $FlowFixMe;
+
+    constructor(ms: number, initialValue: boolean) {
+      this._TestComponent = function _TestComponent(props: {| value: boolean |}) {
+        const hookOutput = useHasStayedTrueForMs(props.value, ms);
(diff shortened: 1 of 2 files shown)


chrisbobbe:
The current revision uses a very simple test-only function component.

It looks like we could use a third-party library, @testing-library/react-hooks (https://github.com/testing-library/react-hooks-testing-library), to avoid doing that, and just test the hook by itself.

I'm inclined to keep the current approach because the function component is easy to define, for our needs here. And `react-test-renderer` is maintained by React (it has a doc on the React website at https://reactjs.org/docs/test-renderer.html).

src/__tests__/reactUtils-test.js
```diff
@@ -0,0 +1,232 @@
...
+        { waitBefore: MS / 5 - 1, expectedOutput: false, thenUpdateInputTo: true },
+        { waitBefore: MS / 5 + 1, expectedOutput: true },
+      ],
+    },
+  ];
+
+  for (let i = 0; i < sequencesToTest.length; i++) {
+    const currentSequence = sequencesToTest[i];
+    test(JSON.stringify(currentSequence, null, 2), async () => {
+      await testSequence(sequencesToTest[i]);
```
gnprice, line 229: nit:
```suggestion
    const currentSequence = sequencesToTest[i];
    test(JSON.stringify(currentSequence, null, 2), async () => {
      await testSequence(currentSequence);
```
Or is there something subtle here I'm missing?

src/__tests__/reactUtils-test.js
```diff
@@ -0,0 +1,232 @@
...
+        { waitBefore: MS / 5, expectedOutput: false, thenUpdateInputTo: true },
+        { waitBefore: MS / 5 - 1, expectedOutput: false, thenUpdateInputTo: true },
+        { waitBefore: MS / 5 + 1, expectedOutput: true },
+      ],
+    },
+  ];
+
+  for (let i = 0; i < sequencesToTest.length; i++) {
+    const currentSequence = sequencesToTest[i];
+    test(JSON.stringify(currentSequence, null, 2), async () => {
```
gnprice, line 228: These test "names" get pretty long in the Jest output. Let's add something like a `description` property to each of these test-case objects, and use that here.

Another benefit of that is that it'd provide a good place for a few words explaining the scenario each of these is meant to cover.

For example:
```js
    {
      description: "start false, wait short time",
      initialValue: true,
      sequence: [{ waitBefore: MS / 2, expectedOutput: false }],
    },
    {
      description: "start false, wait long time",
      initialValue: true,
      sequence: [{ waitBefore: MS * 2, expectedOutput: true }],
    },
```

src/__tests__/reactUtils-test.js
```diff
@@ -0,0 +1,232 @@
...
+    {
+      initialValue: false,
+      sequence: [{ waitBefore: MS / 2, expectedOutput: false }],
+    },
+    {
+      initialValue: false,
+      sequence: [
+        { waitBefore: MS / 2, expectedOutput: false, thenUpdateInputTo: true },
+        { waitBefore: MS * 2, expectedOutput: true, thenUpdateInputTo: false },
+        { waitBefore: MS / 2, expectedOutput: false },
```
gnprice, line 153: nit:
```suggestion
      sequence: [
        { waitBefore: MS / 2, expectedOutput: false, thenUpdateInputTo: true },
        { waitBefore: 2 * MS, expectedOutput: true, thenUpdateInputTo: false },
        { waitBefore: MS / 2, expectedOutput: false },
```
and similarly the other `MS * 2` instances.

That makes the "twice" and "half" more visually distinct, making it a bit easier to scan each of these scenarios and see what it's saying. Perhaps not coincidentally, it's also more conventional in math and physics: numeric coefficient first, then the variable or dimensional quantity it's multiplying goes after that.

src/__tests__/reactUtils-test.js
```diff
@@ -0,0 +1,232 @@
...
+    hookOutput() {
+      const result = this._testRenderer.root.children[0] === TestMachine.HOOK_VALUE_TRUE;
+      return result;
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    async sleep(ms: number): Promise<void> {
+      // The hook uses `useState`, which seems to make `react-test-renderer`
+      // complain if we don't use `act`.
+      return act(() => fakeSleep(ms));
```
gnprice, line 73: Yeah. Here's the error if I remove the `act`:
```
  console.error
    Warning: An update to _TestComponent inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at _TestComponent (/home/greg/z/mobile/src/__tests__/reactUtils-test.js:54:71)

      70 |   useEffect(() => {
      71 |     if (value) {
    > 72 |       const id = setTimeout(() => setResult(true), duration);
```

I think the point is that `fakeSleep` causes that timer to run -- that's why we're calling it, after all -- which means we set the state. That causes an update to the test component, so we need it wrapped in `act` to make sure the updates get applied.

gnprice (commented):
Great! Thanks for exploring this.

These tests look good. Small comments below, then please merge at will.

src/__tests__/reactUtils-test.js
chrisbobbe: Nope, you're right! Thanks!

src/__tests__/reactUtils-test.js
chrisbobbe: Yeah. It occurs to me after re-reading that error you quoted: I think they did a pretty good job raising this error and choosing its error message. It happens early, it's specific, and it's actionable.

chrisbobbe:
> Great! Thanks for exploring this.
> 
> These tests look good. Small comments below, then please merge at will.

PR #4998, opened 2021-09-11
Distinguish types PmOutbox vs. StreamOutbox more fully

gnprice:
This expresses that the types of certain properties on `Outbox` vary together, and that they do so in the same way as on `Message`.
    
That in turn means that when we have a condition like `type === 'stream'`, Flow can infer that `display_recipient` will be a string (the stream name) rather than an array, and so on.

Also make some followup simplifications that this opens up.

(~~One upcoming place I suspect this will simplify the code a bit is at https://github.com/zulip/zulip-mobile/pull/4899#discussion_r703932340~~ Hmm, no -- that code wants `stream_id`, and that's something we don't yet have on `Outbox`: #3998. Ah well, this will simplify that code once we also have #3998.)


chrisbobbe:
Thanks, LGTM! Merged.

PR #5002, opened 2021-09-14
api errors: Make the invalid-JSON case a bit more explicit

gnprice:
The way these errors are handled in `apiFetch.js` and sent to
Sentry, if `data` is undefined then the corresponding property
just disappears in the Sentry breadcrumb.  That can be a bit
confusing when matching up the Sentry data with the source code.

Instead, use a string that says the data was invalid JSON.
(Which it was -- that's how `data` here can be undefined.)


chrisbobbe:
As we discussed in the office today:

This will have the right effect in the 200 case, but it won't have an effect in the 4xx case. Should a change instead be made at the place where we consume `MalformedResponseError`s?

gnprice:
Thanks for the review! Pushed a revision.

chrisbobbe:
Thanks! LGTM; merged.

PR #5003, opened 2021-09-14
Extract and improve reusable banner component; fix eventQueueId type; get feature level from /register

chrisbobbe:
I'd like to make these improvements on the path to a better alternative to 6d2647f2730665215c68e3c0e98a7d8666e2314a. I think they're also good to do independently of my plan for that.

See discussion: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/connection.20failed/near/1248170

(screenshots coming soon)

### DIFF
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -19,7 +19,7 @@ import type {
 export type InitialDataBase = $ReadOnly<{|
   last_event_id: number,
   msg: string,
-  queue_id: number,
+  queue_id: string,
 |}>;
 
 export type InitialDataAlertWords = $ReadOnly<{|
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -21,10 +21,10 @@ export type InitialDataBase = $ReadOnly<{|
   msg: string,
   queue_id: string,
 
-  // The feature level (below) is included unconditionally as of 3.0
-  // (zulip/zulip@2c6313019), which is why we put it in `InitialDataBase`.
-  // To get it from servers older than that, we have to pass `zulip_version`
-  // in `fetch_event_types`, which we do.
+  // The feature level and server version, below, are included
+  // unconditionally as of 3.0 (zulip/zulip@2c6313019), which is why we put
+  // them in `InitialDataBase`. To get them from servers older than that, we
+  // have to pass `zulip_version` in `fetch_event_types`, which we do.
   //
   // TODO(server-3.0): We can stop passing `zulip_version` in
   // `fetch_event_types` and remove this comment.
@@ -35,6 +35,14 @@ export type InitialDataBase = $ReadOnly<{|
    * https://zulip.com/api/get-server-settings. See also the comment above.
    */
   zulip_feature_level?: number,
+
+  /**
+   * Should be present as far back as 1.6.0 (zulip/zulip@d9b10727f). Same
+   * meaning as in the server_settings response:
+   * https://zulip.com/api/get-server-settings. See also the comment above
+   * `zulip_feature_level`, above.
+   */
+  zulip_version: string,
 |}>;
 
 export type InitialDataAlertWords = $ReadOnly<{|
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -20,6 +20,21 @@ export type InitialDataBase = $ReadOnly<{|
   last_event_id: number,
   msg: string,
   queue_id: string,
+
+  // The feature level (below) is included unconditionally as of 3.0
+  // (zulip/zulip@2c6313019), which is why we put it in `InitialDataBase`.
+  // To get it from servers older than that, we have to pass `zulip_version`
+  // in `fetch_event_types`, which we do.
+  //
+  // TODO(server-3.0): We can stop passing `zulip_version` in
+  // `fetch_event_types` and remove this comment.
+
+  /**
+   * Added in server version 2.2, feature level 1.
+   * Same meaning as in the server_settings response:
+   * https://zulip.com/api/get-server-settings. See also the comment above.
+   */
+  zulip_feature_level?: number,
 |}>;
 
 export type InitialDataAlertWords = $ReadOnly<{|
@@ -92,13 +107,6 @@ export type InitialDataRealm = $ReadOnly<{|
   realm_uri: string,
   realm_video_chat_provider: number,
   realm_waiting_period_threshold: number,
-
-  /**
-   * Added in server version 2.2, feature level 1.
-   * Same meaning as in the server_settings response:
-   * https://zulip.com/api/get-server-settings
-   */
-  zulip_feature_level?: number,
 |}>;
 
 export type InitialDataRealmEmoji = $ReadOnly<{|
--- a/src/common/ServerCompatBanner.js
+++ b/src/common/ServerCompatBanner.js
@@ -2,52 +2,23 @@
 
 import React from 'react';
 import type { Node } from 'react';
-import { View } from 'react-native';
-import { SafeAreaView } from 'react-native-safe-area-context';
 
-import { createStyleSheet, HALF_COLOR } from '../styles';
+import ZulipBanner from './ZulipBanner';
 import { useSelector, useDispatch } from '../react-redux';
-import Label from './Label';
 import { getIdentity, getServerVersion } from '../account/accountsSelectors';
 import { getIsAdmin, getSession, getSettings } from '../directSelectors';
 import { dismissCompatNotice } from '../session/sessionActions';
-import ZulipTextButton from './ZulipTextButton';
 import { openLinkWithUserPreference } from '../utils/openLink';
 
 // The oldest version we currently support. Should match what we say at
 //   https://zulip.readthedocs.io/en/stable/overview/release-lifecycle.html#compatibility-and-upgrading.
 const minSupportedVersion = '2.1.0';
 
-const styles = createStyleSheet({
-  wrapper: {
-    backgroundColor: HALF_COLOR,
-    paddingLeft: 16,
-    paddingRight: 8,
-    paddingBottom: 8,
-  },
-  textRow: {
-    flexDirection: 'row',
-  },
-  text: {
-    marginTop: 16,
-    lineHeight: 20,
-  },
-  buttonsRow: {
-    marginTop: 12,
-    flexDirection: 'row',
-    justifyContent: 'flex-end',
-  },
-});
-
 type Props = $ReadOnly<{||}>;
 
 /**
  * A "nag banner" saying the server version is unsupported, if so.
  */
-// Made with somewhat careful attention to
-// https://material.io/components/banners. Please consult that before making
-// layout changes, and try to make them in a direction that brings us closer
-// to those guidelines.
 export default function ServerCompatBanner(props: Props): Node {
   const dispatch = useDispatch();
   const hasDismissedServerCompatNotice = useSelector(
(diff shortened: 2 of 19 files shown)


chrisbobbe:
# Before:






# After:

src/api/initialDataTypes.js
```diff
@@ -19,7 +19,30 @@ import type {
...
+  queue_id: string,
+
+  // The feature level and server version, below, are included
+  // unconditionally as of 3.0 (zulip/zulip@2c6313019), which is why we put
+  // them in `InitialDataBase`. To get them from servers older than that, we
+  // have to either
+  // - pass `zulip_version` in `event_types`, or
+  // - not specify `event_types`.
+  // We've been doing the latter for a long time, so we're fine.
+
```
gnprice, line 31: Huh. We *do* specify `fetch_event_types`, which controls most of what appears in the initial data. If it's `event_types` here, that seems like an anomaly.

I guess in that case we fixed it long ago by sending these unconditionally :shrug:

src/api/initialDataTypes.js
```diff
@@ -19,7 +19,30 @@ import type {
...
+   */
+  zulip_feature_level?: number,
+
+  /**
+   * Should be present as far back as 1.6.0 (zulip/zulip@d9b10727f). Same
+   * meaning as in the server_settings response:
+   * https://zulip.com/api/get-server-settings. See also the comment above
+   * `zulip_feature_level`, above.
+   */
+  zulip_version: string,
```
gnprice, line 45: The commit message is just

    register: Note that server feature level is also in /register response.

but this `zulip_version` seems like the more interesting change.

gnprice: Hmm, also it looks like as of that early commit zulip/zulip@d9b10727f, this is conditional in exactly the same way that both this and `zulip_feature_level` are conditional until the later change zulip/zulip@2c6313019.

Maybe we should just send `zulip_version` in our `fetch_event_types`, so that this property is unconditionally present? With a `TODO(server-3.0)` comment, to help us remove it in the future.

src/common/ZulipBanner.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  return (
+    <SafeAreaView mode="padding" edges={['right', 'left']} style={styles.wrapper}>
+      <View style={styles.textRow}>
+        <Label style={styles.text} text={text} />
+      </View>
+      <View style={styles.buttonsRow}>
+        {buttons.map(({ id, label, onPress }) => (
+          <ZulipTextButton key={id} label={label} onPress={onPress} />
```
gnprice, line 86: Is `key` needed here?

I think it shouldn't be -- the caller isn't likely to have a dynamic list of buttons, but even if they do, I don't think `ZulipTextButton` has any state that would make it matter to preserve the identity of which button is which across re-renders.

src/common/ServerCompatBanner.js
```diff
@@ -58,50 +29,50 @@ export default function ServerCompatBanner(props: Props): Node {
...
+      buttons={[
+        {
+          id: 'dismiss',
+          label: 'Dismiss',
+          onPress: () => {
             dispatch(dismissCompatNotice());
-          }}
-        />
-        <ZulipTextButton
-          leftMargin
```
gnprice, line 95: This `leftMargin` gets dropped, in the `[nfc]` commit that factors out `ZulipBanner`.

Did we need it? It looked like it was there to provide spacing between the two buttons.

Probably using flexbox is better, and I see that's the direction the later commits aim at. If it's a pain to reproduce this behavior in the initial factoring-out, then I don't mind having the layout a bit wonky as an intermediate state; just mention in the commit message.

gnprice (commented):
Thanks @chrisbobbe! I've just done a quick read partway through, and should probably call it a night. Comments below on these commits:
758ae3ba4 session state: Use a better type for eventQueueId.
a5d1440dc register: Note that server feature level is also in /register response.
cc60b491c register: Get both server version and feature level from register response.
d622ecf2a docs [nfc]: Remove mentions of `fetchInitialData`.
2fb17dfb6 ZulipBanner [nfc]: Extract reusable component, from ServerCompatBanner.

That leaves these for me to read another time:
e1cfe581d ZulipBanner [nfc]: Better support single-row layout (1/3).
6c04b60d4 ZulipBanner [nfc]: Better support single-row layout (2/3).
d5af39b1e ZulipBanner: Better support single-row layout (3/3).
dabbf0a42 fetchActions: Use `loading` instead of `needsInitialFetch`.

src/common/ServerCompatBanner.js
gnprice: OK, having now read the remaining commits I see they don't appear to affect this between-button spacing. And looking at the screenshots above, I see there was a change. So I think a fix to keep this spacing is needed.

Looking at the available flexbox properties, I think those won't actually give us what we want: 8px between items. Which maybe makes sense because after all that isn't really about *flex*ing.

So probably the simplest thing is to keep using `leftMargin`. You can use the second parameter to the `Array#map` callback (i.e., the index) to add it back without much change to the new code.

src/message/fetchActions.js
```diff
@@ -170,9 +169,9 @@ export const fetchOlder = (narrow: Narrow): ThunkAction<void> => (dispatch, getS
   const firstMessageId = getFirstMessageId(state, narrow);
   const caughtUp = getCaughtUpForNarrow(state, narrow);
   const fetching = getFetchingForNarrow(state, narrow);
-  const { needsInitialFetch } = getSession(state);
+  const { loading } = getSession(state);
 
-  if (!needsInitialFetch && !fetching.older && !caughtUp.older && firstMessageId !== undefined) {
+  if (!loading && !fetching.older && !caughtUp.older && firstMessageId !== undefined) {
```
gnprice, line 174: Hmm. What are the situations where this would produce a different result? I feel like I need to think through some more reasoning to be confident this won't result in some behavior that feels buggy.

I definitely like the idea of reducing entanglement with `needsInitialFetch`. Grepping after this change, it looks like… the only remaining place we consume it is in `AppDataFetcher` itself! The other references are all in comments, tests, or where we update it (but never read it) in `sessionReducer`.

gnprice (commented):
OK, review complete! Another comment below, as well as those above.

Another thing I notice is that it looks like this is several pretty independent changes: a 4-commit series on `ZulipBanner`, a 2-commit series on the server version and feature level, and 3 other commits that are all independent. So if some of these prove trickier to complete, we can go ahead and merge the others promptly while work continues on the former.

src/api/initialDataTypes.js
chrisbobbe: I agree, it's odd for it to be `event_types` instead of `fetch_event_types`. But I think it is. From the quoted commit https://github.com/zulip/zulip/commit/2c6313019:

```
**Feature level 3**:

* `zulip_version` and `zulip_feature_level` are always returned
  in `POST /register`; previously they were only returned if `event_types`
  included `zulip_version`.
```

and `want`, as in the removed lines

```diff
-    if want('zulip_version'):
-        state['zulip_version'] = ZULIP_VERSION
-        state['zulip_feature_level'] = API_FEATURE_LEVEL
```

, is

```
    if event_types is None:
        # return True always
        want: Callable[[str], bool] = always_want
    else:
        want = set(event_types).__contains__
```

src/api/initialDataTypes.js
chrisbobbe: > Maybe we should just send `zulip_version` in our `fetch_event_types`, so that this property is unconditionally present? With a `TODO(server-3.0)` comment, to help us remove it in the future.

This is what I'd thought we had to do, and what made it into my earlier [post](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/connection.20failed/near/1248170) on CZO.

But looking closer, it's actually controlled by `event_types`, not `fetch_event_types`. So we should get the server version and (from servers that have one) the server feature level unconditionally, without having to do anything but keep passing `null` for `event_types`.

src/common/ZulipBanner.js
chrisbobbe: Without it, I see the expected warning message:

image

That warning does go away if I use the list index for the `key` though. Should I do that?

src/message/fetchActions.js
chrisbobbe: It looks like we started using `needsInitialFetch` in 7a7a63fb1, with the commit message

```
Prevent extra fetches before initial fetch is complete
```

chrisbobbe: So if `needsInitialFetch` is `true`, then `loading` should be either
- `true` because the initial fetch is in progress, or
- `false`, but about to follow `needsInitialFetch` in switching to `true`, because an initial fetch is about to start (`AppDataFetcher`'s `useEffect` is about to call `doInitialFetch`).

In the second case, I guess it technically matters whether there's a yield between `needsInitialFetch` becoming `true` and `loading` becoming `true`. I think there isn't such a yield, but I could be wrong.

If there is, then I guess it becomes possible that we do a fetch while `needsInitialFetch` is `true` but `loading` is still false. I think that environment exists in the following ways:
- the app has just started up and rehydrated (and we discover we have an active, logged-in account)
- the server has on the server invalidating an existing event queue after 10 minutes of inactivity

I think it's not really realistic for `fetchOlder` or `fetchNewer` to be called in the first case, at startup. They're called from a UI that has to be reached by navigating.

For the second case...I guess (and again this is only if there's a yield) it would mean we pretty minimally extend the existing window during which it's possible for this code to start a message fetch that ends up still being in progress when an initial fetch starts.

chrisbobbe: If `needsInitialFetch` is `false`, then `loading` should be `false` too:

- All of the cases in `sessionReducer` that set `needsInitialFetch` to `false` also set `loading` to false.
- There's one case where `loading` is set to `true` and `needsInitialFetch` isn't touched. This is on a `INITIAL_FETCH_START`, and at that point `needsInitialFetch` will be `true` too (and will stay that way until the initial fetch completes or aborts).

src/api/initialDataTypes.js
gnprice: But that's the same `want` that is used in the rest of that function too, for all the other parts of the initial data.

Looking at that function's callsite lower in that file, its `event_types` argument is computed like so:
```py
    if fetch_event_types is not None:
        event_types_set: Optional[Set[str]] = set(fetch_event_types)
    elif event_types is not None:
        event_types_set = set(event_types)
    else:
        event_types_set = None

    # …
        ret = fetch_initial_state_data(
            user_profile,
            event_types=event_types_set,
```

So `fetch_event_types` controls if it's present. I'm guessing that that name (a parameter to *that* function) corresponds to the same name in the HTTP request.

src/common/ZulipBanner.js
gnprice: Hmm, I see. Looks like what's happening is that the old code with two separate `<ZulipTextButton … />` expressions in the source resulted in two separate arguments to `React.createElement`, while this version with `{buttons.map(…)}` results in passing the list as a single argument to `React.createElement`. Here's [a demo](https://babeljs.io/repl/#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABACwKYBt1wBQEpEDeAUIogE6pQhlIA8AJjAG4B8tAhogB4C8BAjAF9EAejadeBAEzCxtEY1YBuIoKJFQkWAhQYsUvIRKIICAM5RE6RD0QBtfgBpEUgLorSFKjUQNmLY1ICdAA6AFt2AAdsbC58HhZEbEDSXwk-LlkA1NTcXDVSeUUWFTUgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.15.6&externalPlugins=&assumptions=%7B%7D) (starting from the handy link at the top of ["JSX In Depth"](https://reactjs.org/docs/jsx-in-depth.html).)

And then the warning is presumably driven by seeing that list of elements, and pretty reasonably figuring that it's dynamic and should use `key`.

In that case, let's go ahead and use keys like you have them. I wouldn't feel good about using the list index for the key -- that's such a well-known bad pattern that using it here to shut up a warning would look like hubris inviting a bug :wink:

src/message/fetchActions.js
gnprice: > In the second case, I guess it technically matters whether there's a yield between `needsInitialFetch` becoming `true` and `loading` becoming `true`. I think there isn't such a yield, but I could be wrong.

I'm not sure exactly what happens between the reducer updating `needsInitialFetch` to true, and the effect in `AppDataFetcher` being run. There must be a bunch of React Redux machinery. I agree there probably isn't a yield.

Once that effect starts executing, we go quite promptly back to the reducer to update `loading` to true, with no yield, because `dispatch(initialFetchStart());` is the first line of the `doInitialFetch` body.


> * the server has on the server invalidating an existing event queue after 10 minutes of inactivity

Yeah, I think this is the plausible case where this could make any difference.

Looking at `sessionReducer`, there are a couple of other cases that also give us `needsInitialFetch` true while `loading` still false:
* we've just logged in;
* we've just switched accounts.

But just like just starting up and rehydrating, those lead to a UI where you can't trigger a `fetchOlder` or `fetchNewer` without first navigating somewhere.


> If `needsInitialFetch` is `false`, then `loading` should be `false` too:

I agree with this reasoning.

One correction that doesn't change the conclusion: there's a case that can set `needsInitialFetch` to false that doesn't touch `loading`, which is on rehydrate. But `loading` is always false then, being untouched from its initial value.

gnprice: Thanks for going through that logic! I think this reasoning suffices to conclude that this change is OK.

The key points are:
* `needsInitialFetch` is never false when `loading` is true
* whenever `needsInitialFetch` is true and `loading` isn't, `AppDataFetcher` will immediately cause `loading` to become true
* so at most, this means
  > we pretty minimally extend the existing window during which it's possible for this code to start a message fetch that ends up still being in progress when an initial fetch starts.

It's a bug, and a potential source of wrong data, that that window exists -- or rather, that we don't cancel that earlier message fetch and may use its results, which are potentially missing updates that happened before we created our new event queue. But this change doesn't meaningfully expand that bug, and it helps us toward simplifying the `needsInitialFetch` / `AppDataFetcher` logic.

src/api/initialDataTypes.js
chrisbobbe: Huh, interesting. Thanks for noticing that!

chrisbobbe:

chrisbobbe:
Just rebased.

src/common/ZulipBanner.js
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <SafeAreaView mode="padding" edges={['right', 'left']} style={styles.wrapper}>
+      <View style={styles.textRow}>
+        <Label style={styles.text} text={text} />
+      </View>
+      <View style={styles.buttonsRow}>
+        {buttons.map(({ id, label, onPress }, index) => (
+          <ZulipTextButton
+            key={id}
+            leftMargin={index !== 0 || undefined}
```
gnprice, line 88: This `|| undefined` is kind of an odd wrinkle :slightly_smiling_face: I see it's to deal with that component's interface:
```js
  leftMargin?: true,
```
We should probably just make that `leftMargin?: boolean,` and say it defaults to false.

gnprice (commented):
Thanks for the revision!

This looks good -- one comment below, which I think is fine to do as a followup. Merging.

chrisbobbe:
Thanks for the review!

> one comment below, which I think is fine to do as a followup

Sure, will do.

Issue #4994, opened 2021-09-08
Test flake on ZulipAsyncStorage


gnprice:
Our unit tests for ZulipAsyncStorage have a flake: the round-trip tests occasionally fail.

This happened on #4992, for example -- though I can't seem to find the record of that CI run anymore in GitHub's CI interface, after the PR got updated as part of merging it.

Filing this issue to track fixing it.

PR #5008, opened 2021-09-17
ZulipAsyncStorage: Document and assert JSON requirement

gnprice:
We've always passed JSON-encoded data in the one place we actually
use this storage, inside the redux-persist code; and our `getItem`
has relied on that assumption for a long time.

But our tests for this module have been passing random strings
instead.  When those happen to start with a `z`, that collides
with our compression header, and the test flakes.

Document that requirement on this module's actual interface, and
add some debug asserts to enforce it.  That then makes the flakes
much easier to reproduce; fix those tests to conform to the
interface's requirements.

Fixes: #4994


chrisbobbe:
Thanks! LGTM; merged.

PR #5011, opened 2021-09-20
recipient [nfc]: Small cleanups

gnprice:
I was looking at this code and saw this comment about a hack, and realized it was now easy to clean it up.


chrisbobbe:
Thanks! LGTM, merged.

PR #5012, opened 2021-09-20
webview html: Skip expensive isSameRecipient where we can

gnprice:
When making the changes in #5011, I saw the comment in `isSameRecipient` mentioning that it's in a hot loop where we're already doing too much computation. There's a description there of how we can fix that for all cases; but in the meantime, for the case where the user is just looking at a single topic or PM conversation we can skip calling this function entirely.


chrisbobbe:
Thanks! LGTM, merged.

Issue #5006, opened 2021-09-16
Multi-org-compatible schemas for data in Redux and persistent store


gnprice:
We'd like to store server data for all the accounts the user is logged in as, #5005, instead of just a single "[active account](https://github.com/zulip/zulip-mobile/blob/2627cfa6fbdc21c2aa3c9e6a9d09695f2893ead8/docs/glossary.md#active-account)".

As part of that, we'll need to arrange the data in a way that makes room for data from multiple accounts, both in Redux and in the store that persists between runs of the app.

This issue is for developing that new pair of schemas and migrating to it, *without* yet making any changes to what data we actually store.

I'll add in a comment some specific design thoughts we have so far on what that might look like.

gnprice:
(placeholder first comment)

gnprice:
(placeholder second comment, for new design)

gnprice:
(placeholder comment for the persistent-state schema)

gnprice:
Another question is how to do migration. This was actually the reason we forked/vendored redux-persist in the first place: upstream had a new major version with a new design for migrations, which insisted on doing migrations locally in each subtree and was not compatible with changes that seriously rearranged the subtrees themselves, as we need to do here.

I think the existing migration logic will *mostly* give us what we need. But this is something I'll need to look at more closely.

One thing that I know is missing is that when we stop using an old key -- for example `users`, in favor of `byAccount:1:users` and `byAccount:2:users` and so on -- and then we write out a migrated version of the state using the new keys, it would be good to actually delete the old key. The current code will let it sit around, which is inefficient at best and a hazard for causing subtle bugs. To date this hasn't mattered because we have never, at least not since 2018 when we started doing migrations at all, stopped using an old key.

gnprice:
Here's a different aspect that requires some more design and experimentation: what will we want the app *code* to look like?

One part of this, about reducers, is discussed above:

> * We'd like all the server data for a given account to be bundled together as one subtree somewhere inside the state.
>   That way, when handling a server event or the result of another fetch (like a message fetch), we can have a sub-reducer function that gets passed the new data and just that one subtree -- as far as that sub-reducer is concerned, there's only one account in the world, and it has no access to other accounts' data.
[...]

PR #5016, opened 2021-09-21
redux: Various small prep changes for multi-org support

gnprice:
This is the first PR in a series produced from the branch described at https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909 .

This first PR doesn't yet get into the main substance of the changes. It collects a variety of small, mostly independent, prep commits needed along the way, that were quickest to polish into a PR and should be generally easy to review.

### DIFF
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -384,8 +384,6 @@ type NonMaybeGlobalState = NonMaybeProperties<GlobalState>;
 // the corresponding parameter when used in `createSelector`, and this does.
 export type Selector<TResult, TParam = void> = InputSelector<GlobalState, TParam, TResult>;
 
-export type GetState = () => GlobalState;
-
 export type PlainDispatch = <A: Action>(action: A) => A;
 
 export interface Dispatch {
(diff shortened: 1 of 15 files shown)


src/sharing/index.js
```diff
@@ -12,25 +11,20 @@ const Sharing = NativeModules.Sharing ?? {
...
+    goToSharing(initialSharedData);
   }
 };
 
 export class ShareReceivedListener {
-  dispatch: Dispatch;
   unsubs: Array<() => void> = [];
 
-  constructor(dispatch: Dispatch) {
-    this.dispatch = dispatch;
```
chrisbobbe, line 31:
```
sharing [nfc]: Stop taking dispatch, no longer used here

This module hasn't needed a Redux `dispatch` since 543c07c, when
we stopped using Redux for dispatching navigation actions.  So,
simplify by not taking one in the first place.
```

This is great!!

src/reduxTypes.js
```diff
@@ -384,13 +384,11 @@ type NonMaybeGlobalState = NonMaybeProperties<GlobalState>;
 // the corresponding parameter when used in `createSelector`, and this does.
 export type Selector<TResult, TParam = void> = InputSelector<GlobalState, TParam, TResult>;
 
-export type GetState = () => GlobalState;
```
chrisbobbe, line 387: Wow, really? Nice!

chrisbobbe (approved):
Thanks! LGTM, merging.

PR #5017, opened 2021-09-22
redux: Distinguish per-account vs global state (1/n)

gnprice:
This is the second PR in the series that began with #5016, produced from the branch described at https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909 .

At the end of this PR, we have a type PerAccountState which is distinct from GlobalState, exposing only the per-account parts of the data. A GlobalState can be freely used implicitly as a PerAccountState, but not vice versa. And to start exercising the distinction, most selectors have been marked as per-account, accepting a mere PerAccountState -- which also means that any selectors they go on to invoke must also be per-account.

The next steps in my draft branch after this PR are to convert the remaining selectors to be per-account where appropriate. Mainly this means those consuming `state.accounts` in `accountsSelectors`, which involves teasing apart just what's expected at their various callers -- whether they truly want "the active account", or (like the bulk of our other selector callsites) want "*this* account", i.e. the particular account the caller is already acting on, which is just assumed to be the active account because that assumption typically holds across our codebase.

After that, we'll convert thunk actions and React components so that they get a PerAccountState rather than a GlobalState.

### DIFF
--- a/src/emoji/emojiSelectors.js
+++ b/src/emoji/emojiSelectors.js
@@ -1,13 +1,14 @@
 /* @flow strict-local */
 import { createSelector } from 'reselect';
 import type { Selector, RealmEmojiById, ImageEmojiType } from '../types';
+import { assumeSecretlyGlobalState } from '../reduxTypes';
 import { getRawRealmEmoji } from '../directSelectors';
 import { getIdentity } from '../account/accountsSelectors';
 import zulipExtraEmojiMap from './zulipExtraEmojiMap';
 import { objectFromEntries } from '../jsBackport';
 
 export const getAllImageEmojiById: Selector<RealmEmojiById> = createSelector(
-  getIdentity,
+  state => getIdentity(assumeSecretlyGlobalState(state)),
   getRawRealmEmoji,
   (identity, realmEmoji) => {
     const result: {| [string]: ImageEmojiType |} = {};
--- a/src/session/sessionReducer.js
+++ b/src/session/sessionReducer.js
@@ -20,21 +20,12 @@ import {
 import { getHasAuth } from '../account/accountsSelectors';
 
 /**
- * Miscellaneous non-persistent state about this run of the app.
+ * Miscellaneous non-persistent state specific to a particular account.
  *
- * These state items are stored in `session.state`, and 'session' is
- * in `discardKeys` in src/boot/store.js. That means these values
- * won't be persisted between sessions; on startup, they'll all be
- * initialized to their default values.
+ * See {@link SessionState} for discussion of what "non-persistent" means.
  */
-export type SessionState = $ReadOnly<{|
+export type PerAccountSessionState = $ReadOnly<{
   eventQueueId: number,
-
-  // `null` if we don't know. See the place where we set this, for what that
-  // means.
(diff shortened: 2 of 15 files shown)


src/emoji/emojiSelectors.js
```diff
@@ -1,13 +1,14 @@
...
 import type { Selector, RealmEmojiById, ImageEmojiType } from '../types';
+import { assumeSecretlyGlobalState } from '../reduxTypes';
 import { getRawRealmEmoji } from '../directSelectors';
 import { getIdentity } from '../account/accountsSelectors';
 import zulipExtraEmojiMap from './zulipExtraEmojiMap';
 import { objectFromEntries } from '../jsBackport';
 
 export const getAllImageEmojiById: Selector<RealmEmojiById> = createSelector(
-  getIdentity,
+  state => getIdentity(assumeSecretlyGlobalState(state)),
```
chrisbobbe, line 11:
```
redux types: Add two assumeSecretlyGlobalState markings
```

Just to check my understanding: Is it intentional that `getIdentity` is still a `Selector` at the time this change is made, and the switch to `GlobalSelector` comes after?

In the commit message, I see "[…] invokes one of those selectors (which demands a GlobalState)", and I think I'm a little confused because at this commit `getIdentity` doesn't yet demand a `GlobalState`, by that name, anyway.

src/session/sessionReducer.js
```diff
@@ -19,6 +19,41 @@ import {
...
+   * Whether `ServerCompatNotice` (which we'll add soon) has been
+   *   dismissed this session.
+   *
+   * We put this in the per-session state deliberately, so that users
+   * see the notice on every startup until the server is upgraded.
+   * That's a better experience than not being able to load the realm
+   * on mobile at all, which is what will happen soon if the user
+   * doesn't act on the notice.
+   */
+  hasDismissedServerCompatNotice: boolean,
```
chrisbobbe, line 52: Cool! This looks like progress toward solving a pretty small bug, but a bug nonetheless.

From https://github.com/zulip/zulip-mobile/pull/4750#issue-892286064:

> NOTE: The has-dismissed-the-banner state isn't stored per-realm, so there's currently a bug where dismissing the banner makes it go away for all <2.0.0 realms the device knows about (until app restart, anyway). I figured this was minor enough to be acceptable for the time being, unless we can think of a nice and easy way to store the state both per-realm and per-session, while being sure to clean up after ourselves when the user wants to forget the realm.

chrisbobbe: (I think per-account is a totally fine proxy for per-realm.)

chrisbobbe (commented):
Thanks! I think this looks good; I've posted one question, below, mostly to check my understanding.

src/emoji/emojiSelectors.js
gnprice: Yeah, I guess that is kind of talking about the future. Probably it'd be clearer to switch things to GlobalSelector where appropriate, then add these two markings, then switch other selectors to PerAccountState where appropriate (and then make the change that distinguishes PerAccountState from GlobalState.) I'll quick make that change.

gnprice:

PR #5021, opened 2021-09-23
Use PmMessage, StreamMessage, PmOutbox, StreamOutbox more, replacing runtime checks.

chrisbobbe:
This spun off of a draft I've been working on to resume message-list diffing. I've forgotten the specifics of why this cleanup is helpful or necessary for that work (and it would take some effort to go back and check), but I'm hoping it will still seem worthwhile without that context. 🙂 

Mainly, this PR takes a small handful of functions in `recipient.js` like this:

```js
/** The stream name a stream message was sent to.  Throws if a PM. */
export const streamNameOfStreamMessage = (message: Message | Outbox): string => {
  if (message.type !== 'stream') {
    throw new Error('streamNameOfStreamMessage: got PM');
  }
  return message.display_recipient;
};
```

And turns them into functions like this:

```js
/** The stream name a stream message was sent to. */
export const streamNameOfStreamMessage = (message: StreamMessage | StreamOutbox): string =>
  message.display_recipient;
```

(These improvements are now possible thanks to #4998! 🎉)

Callers have so far been really attentive in passing the appropriate kind of message or outbox. And I don't think we've seen any reports of those runtime errors in Sentry. But anyway, now we can enlist the type-checker to help these functions make sure their callers keep doing the right thing going forward.

### DIFF
--- a/src/pm-conversations/pmConversationsModel.js
+++ b/src/pm-conversations/pmConversationsModel.js
@@ -11,7 +11,7 @@ import {
 } from '../actionConstants';
 import { makeUserId } from '../api/idTypes';
 
-import type { Action, Message, Outbox, UserId } from '../types';
+import type { Action, PmMessage, PmOutbox, UserId } from '../types';
 import { recipientsOfPrivateMessage } from '../utils/recipient';
 import { ZulipVersion } from '../utils/zulipVersion';
 
@@ -47,7 +47,7 @@ function keyOfUsers(ids: UserId[], ownUserId: UserId): PmConversationKey {
 }
 
 // Input must indeed be a PM, else throws.
-function keyOfPrivateMessage(msg: Message | Outbox, ownUserId: UserId): PmConversationKey {
+function keyOfPrivateMessage(msg: PmMessage | PmOutbox, ownUserId: UserId): PmConversationKey {
   return keyOfUsers(
     recipientsOfPrivateMessage(msg).map(r => r.id),
     ownUserId,
--- a/src/pm-conversations/pmConversationsModel.js
+++ b/src/pm-conversations/pmConversationsModel.js
@@ -46,7 +46,6 @@ function keyOfUsers(ids: UserId[], ownUserId: UserId): PmConversationKey {
   return keyOfExactUsers(ids.filter(id => id !== ownUserId));
 }
 
-// Input must indeed be a PM, else throws.
 function keyOfPrivateMessage(msg: PmMessage | PmOutbox, ownUserId: UserId): PmConversationKey {
   return keyOfUsers(
     recipientsOfPrivateMessage(msg).map(r => r.id),
--- a/src/unread/unreadHuddlesReducer.js
+++ b/src/unread/unreadHuddlesReducer.js
@@ -18,20 +18,22 @@ import { NULL_ARRAY } from '../nullObjects';
 const initialState: UnreadHuddlesState = NULL_ARRAY;
 
 const eventNewMessage = (state, action) => {
-  if (action.message.type !== 'private') {
+  const { message } = action;
+
+  if (message.type !== 'private') {
     return state;
   }
 
-  if (recipientsOfPrivateMessage(action.message).length < 3) {
+  if (recipientsOfPrivateMessage(message).length < 3) {
     return state;
   }
 
-  invariant(action.message.flags, 'message in EVENT_NEW_MESSAGE must have flags');
-  if (action.message.flags.includes('read')) {
+  invariant(message.flags, 'message in EVENT_NEW_MESSAGE must have flags');
+  if (message.flags.includes('read')) {
     return state;
   }
 
-  return addItemsToHuddleArray(state, [action.message.id], pmUnreadsKeyFromMessage(action.message));
+  return addItemsToHuddleArray(state, [message.id], pmUnreadsKeyFromMessage(message));
 };
 
 const eventUpdateMessageFlags = (state, action) => {
--- a/src/message/messageSelectors.js
+++ b/src/message/messageSelectors.js
@@ -1,7 +1,8 @@
 /* @flow strict-local */
 import { createSelector, defaultMemoize } from 'reselect';
+import invariant from 'invariant';
 
-import type { Message, Outbox, Narrow, Selector, MessageListElement } from '../types';
+import type { Message, PmMessage, Outbox, Narrow, Selector, MessageListElement } from '../types';
 import {
   getAllNarrows,
   getFlags,
@@ -32,17 +33,26 @@ function truncateForLogging<T: JSONable>(arr: $ReadOnlyArray<T>, len = 10): JSON
   };
 }
 
-export const getPrivateMessages: Selector<Message[]> = createSelector(
+export const getPrivateMessages: Selector<PmMessage[]> = createSelector(
   getAllNarrows,
   getMessages,
   (narrows, messages) => {
-    const privateMessages: Message[] = [];
+    const privateMessages: PmMessage[] = [];
     const unknownIds: number[] = [];
 
     const pmIds = narrows.get(ALL_PRIVATE_NARROW_STR) || NULL_ARRAY;
     pmIds.forEach(id => {
       const msg = messages.get(id);
       if (msg !== undefined) {
+        // It seems like the code involved in maintaining this invariant
+        // extends to the server: we go and make a fetch for this narrow,
+        // and then whatever the server gives us gets put here, and we don't
+        // seem to try to verify that as it goes in. So in fact ideally this
+        // check would be at a crunchy shell. But getting that wrong is a
+        // pretty unlikely bug for the server to have, so just doing
+        // `invariant` here is fine.
+        invariant(msg.type === 'private', 'msg is a PM');
+
         privateMessages.push(msg);
       } else {
         unknownIds.push(id);
--- a/src/pm-conversations/pmConversationsSelectors.js
+++ b/src/pm-conversations/pmConversationsSelectors.js
@@ -2,7 +2,7 @@
 import invariant from 'invariant';
 import { createSelector } from 'reselect';
 
-import type { PerAccountState, Message, PmConversationData, Selector } from '../types';
+import type { PerAccountState, PmMessage, PmConversationData, Selector } from '../types';
 import { getPrivateMessages } from '../message/messageSelectors';
 import { getAllUsersById, getOwnUserId } from '../users/userSelectors';
 import { getUnreadByPms, getUnreadByHuddles } from '../unread/unreadSelectors';
@@ -37,7 +37,7 @@ export const getRecentConversationsLegacy: Selector<PmConversationData[]> = crea
   getAllUsersById,
   (
     ownUserId,
-    messages: Message[],
+    messages: PmMessage[],
     unreadPms: {| [number]: number |},
     unreadHuddles: {| [string]: number |},
     allUsersById,
--- a/src/utils/recipient.js
+++ b/src/utils/recipient.js
@@ -4,15 +4,19 @@ import isEqual from 'lodash.isequal';
 
 import { mapOrNull } from '../collections';
 import * as logging from './logging';
-import type { PmRecipientUser, Message, Outbox, UserId, UserOrBot } from '../types';
+import type {
+  PmRecipientUser,
+  Message,
+  StreamMessage,
+  Outbox,
+  StreamOutbox,
+  UserId,
+  UserOrBot,
+} from '../types';
 
-/** The stream name a stream message was sent to.  Throws if a PM. */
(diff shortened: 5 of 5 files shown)


gnprice (commented):
Thanks @chrisbobbe! I like to see the types get tighter like this.

A couple of code comments below. Some branch-level comments:
* In the summary lines, I prefer to reserve "type" prefixes for changes that are purely at type level, not affecting any of the runtime code. Many of these add or remove asserts, or make other non-type refactors (like making a `message` local in the last commit).

  So those are NFC (so long as we're certain the asserts are true, which I think we are in each of these (though not the assert I request adding in the first comment below)), but not types-only.
* Some of the changes are to make a function take a smaller type like `PmMessage` (and remove a now-redundant assert), while also refactoring around one of its callsites in order to satisfy the new type. I think those would be clearer as first the refactor, then the change of type.
* The commit on `recipientsOfPrivateMessage` mentions only that one in its commit message, but also updates the types on a couple of its callers.
* Conversely, there are several `pm*FromMessage` functions that are all very similar. Those could be edited in the same commit, once those aren't mixed with refactors or added asserts at callers.

src/pm-conversations/pmConversationsModel.js
```diff
@@ -47,7 +47,7 @@ function keyOfUsers(ids: UserId[], ownUserId: UserId): PmConversationKey {
 }
 
 // Input must indeed be a PM, else throws.
-function keyOfPrivateMessage(msg: Message | Outbox, ownUserId: UserId): PmConversationKey {
+function keyOfPrivateMessage(msg: PmMessage | PmOutbox, ownUserId: UserId): PmConversationKey {
```
gnprice, line 50: This comment can get deleted at the same time as the type is updated.

chrisbobbe:

src/pm-conversations/pmConversationsModel.js
gnprice: Ah, specifically a later commit. So that later change has a hunk like:
```diff
-// Input must indeed be a PM, else throws.
 function keyOfPrivateMessage(msg: PmMessage | PmOutbox, ownUserId: UserId): PmConversationKey {
```

This is fine, but I think it's actually quite legitimate to remove the comment at the time the type gets narrowed to PM-only.

At that point, the comment is talking about a situation that the type says shouldn't exist. In general, if you manage to pass to a function an argument that its type says isn't permitted, the default from the caller's perspective is that anything might happen (much like [undefined behavior](https://en.wikipedia.org/wiki/Undefined_behavior) in language standards), and we don't usually comment on the specifics; if there's a reason to think the caller might need to care, that usually means that either the type or the callsites should be adjusted to bring them into agreement.

src/unread/unreadHuddlesReducer.js
```diff
@@ -18,20 +18,22 @@ import { NULL_ARRAY } from '../nullObjects';
 const initialState: UnreadHuddlesState = NULL_ARRAY;
 
 const eventNewMessage = (state, action) => {
-  if (action.message.type !== 'private') {
+  const { message } = action;
+
+  if (message.type !== 'private') {
```
gnprice, line 23: Now, after the early return on `message.type !== 'private'`, Flow
    will remember that `message` is a PM, even after some intervening
    things happen, like function calls. This will help when
    `pmUnreadsKeyFromMessage` starts taking a `PmMessage` soon.

Yeah. Without this change, it's not so much that Flow doesn't *remember* the type as it is that Flow doesn't *trust* that it will still have that type. The basic issue is that an intervening function call (like to `recipientsOfPrivateMessage`) could go and mutate `action`, causing `action.message` to be some completely other object from the object it was.

Here's the docs on this behavior:
https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations

gnprice (commented):
Thanks for the revision! Looks good; merging. I'll tweak a commit message per the comment below.

src/unread/unreadHuddlesReducer.js
chrisbobbe: Looking again, I think you're right.

I think initially I wanted to say something like that ('trust' instead of 'remember), but then I thought something like, "well, if that's all that's happening, and in the old code Flow didn't want to trust that `action` wasn't being mutated, then why is Flow happy in the new code to trust that `message` isn't being mutated to have a different `message.type`?" And I didn't investigate further.

I think the answer to that is that `PmMessage`, the type that `message` is refined to after the early return, is `$ReadOnly`.

Which I think suggests that we might have taken a different angle than the one in this commit: we might have marked `EventNewMessageAction` as `$ReadOnly`. (Untested.) We should do that anyway, I think, but not urgently, and this change should be fine to make.

gnprice:
Made the edit, ran the test suite, then apparently got distracted before pushing :slightly_smiling_face: . Merged now.

The other small commit-message edit I made, while I was editing that commit, was:

    -    unreadHuddlesReducer [nfc]: Refactor to help a Flow refinement
    +    unread [nfc]: Refactor to help a Flow refinement

I like to think of "unread" as the name of that whole subsystem that maintains the `state.unread` subtree, including the specific files like `unreadHuddlesReducer`.

chrisbobbe:
Great, thanks for the review and those improvements!

src/unread/unreadHuddlesReducer.js
gnprice: > I think initially I wanted to say something like that ('trust' instead of 'remember'), but then I thought something like, "well, if that's all that's happening, and in the old code Flow didn't want to trust that `action` wasn't being mutated, then why is Flow happy in the new code to trust that `message` isn't being mutated to have a different `message.type`?" And I didn't investigate further.

That is indeed a good question. The answer isn't awesome.


> I think the answer to that is that `PmMessage`, the type that `message` is refined to after the early return, is `$ReadOnly`.

Well, the invalidation happens even if the function in question isn't passed any arguments -- which is appropriate, because the object we're looking at may be known by another name somewhere else (e.g., it could be a global somewhere), and so the function we call might happen to somehow know about it and get to it.

And if that callee knows about the object by another name, it might know it by a non-`$ReadOnly` type. When we have a read-only type, that only reflects a promise this code is *making* to any other code that passes it the value; it doesn't come with any promises in the other direction, that the code that supplies the value won't be keeping a reference that it might later allow to be used to mutate it.

So that explanation doesn't seem to be the story.


What the behavior actually seems to be is something like: if you have an object type with some refinements on its properties, and you call a function, then the refinements get invalidated. But if the top-level type is a union and you have some refinements on it, then calling a function doesn't invalidate that.

In particular, if you have `x: { t: 'n' } | { t: 's' }` and a refinement `t !== 'n'`, then that will persist across a function call. But if you have `x: { t: 'n' | 's' }` -- which sure looks awfully similar -- and the same refinement, then that will not persist. [Here's a Try Flow demo](https://flow.org/try/#0C4TwDgpgBAKgFgSwHYHMoF4oG8rAFxQDkShANFAIYFICuAtgEYQBOAvlAD7a4GEDOZKAwJ9gzZGlYBuAFAyAZjSQBjYAgD2SKPIAUADwLwJASmwyoFqAnlR9AOmAZ0mYoVPMIwGsySzLUAHoAqAAlCHlkCDoIJGA+SjAwABsQO3NLe3wiAWM-DL07YShRcVRc9IsUHXL-INDwyOjY+NEEJKSE5JByYDh1GhQ4KAADFGGoZX6kgBMoOAoAN2g6GmAKYAhZ4b1htP9M3hy8i3sikpNZVjlFFTVNbV09A25no1QoVlMsCqsbfQKCo5nC4SO5PN5fD86mEIkgojE4p0Unt8gCHIc3MdbE87AUzmILj8qjVLNCGnCmoikOooElNCgWEjUhYAILyDbMSjtcijcaTGgzOaLCBQ4IrNYbLZPcZ8Gm9dYjHE7IQQZQUaLFdTRXoSKAQJJ8CAok5K9HZTE-f64woiAllKT+GRXBRKVQaLRVZ50BB6TZfaRyGR1ADKBLAUH6jl60AgejEFCgYGY6kgzDUED4DtE6mYICg03UGZIjmU81QRrkoEgsEQqAATBhuFlXB9OE2MR9rq67lp5HX9IZayg618ftZsQ4nCC3FAPF4fFiybD4c0mcaJ82jkTqovgjDGgiWmp2kyen0BkNeRMprN5ksoOL1ptFbtLYCMeVnTc3fd5H3-gQOCvEODafGY-jjlagJTkQoKzuCC6ivUy6UvEFCJMib5opuFr+MSu7IQeq7UrS9KMuhXR2Ky7LkdyIxjNeAq3sKSGPpKirKrKuDzI42wAuMTBqhqsrakOeoGhW+ymjhn6BnUACy6j3tGUBKO6VhIHwCDTNAKnqAwABWqpRuAFZ1AAcuoADu3HQNmub5oW8SlhQ5YAISVqZNYSAAzI2ODNiQbb8IQraBMEdLqAA1mhpGOEkCBRdAbzDuQBZFoQjgIMAAD8Xa3Op-7Ab5o4QX80HArBM5zhCBH7hSh5rm+ZohSSlQ7kh9UroiABUJF0uWzA9U1tTBGyHJckkPIMfygp3ssqxPlKr5SS1W5ft2hVFc8QGDr5HylZYkHSTBrhgvOkKjYRDWrhRmFSdhH5YvhnXkt18T9WRnJ3cyUDjbRU30XyN5CksrGLexfEynKPEcbiAmquqdlap4Yn6oa65QWtFpXEAA).

This leaves the behavior not completely sound in the face of mutation. I suspect their reasoning here reflects a balance: with the way it does behave, you can always make the refinement stick by just picking out the relevant property as a variable, perhaps a property-of-property, however deep the type goes as object types before you reach a union. That's straightforward to do and not messy in the code, and it does really protect you against possible mutation of those objects (even if that isn't a real likely sort of bug, which it isn't in our codebase.) But if they refused more thoroughly to keep refinements across function calls, there isn't a great way we could adjust the code to respond to that. A lot of code would have to become messier, not worth the exclusion of this fairly unlikely class of bug.

gnprice: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flow.20refinements/near/1262824

andersk: I think Flow is being consistent here.

* It’s justified in assuming that the `t` property of `x: { t: 'n' } | { t: 's' }` won’t be mutated by external functions, because mutating it would be a type error in those functions.
* The difference with `xx: { x: { t: 'n' } | { t: 's' } }` is *not* that the `t` property of `xx.x` might be mutated, but rather that the `x` property of `xx` might be mutated.
* The difference with `x: { t: 'n' | 's' }` is that mutating *this* `t` property would not be a type error.

[Demo](https://flow.org/try/#0DYUwLgBA5hC8EAoCUcB8EDeBfA3AKDzAE8AHECAFQAsBLAOygCY5MIwAuCAcjq4iwgAfVh24BnPlgIAzAK50AxmBoB7OhGmMEAD07V6TFBjwRTEGtMTaAdJFj3uvFACdws53XxmrtzlwlIXmZQyEGmOr7iXIEQAPSxECoA1nhSeApqYpDajHq0DMzwGGx+Evz4MPDIaJgmZjm2LDxcOHEJIM7OKs4AhKn4mjqMgTLySqrq0oPauqyz+gX8RnWmFlY2NnYOzS5uHmHr1pul0QchI946G5H+p20QHV29qQQZdFkQM7lzeQbMAkVPpxiqJbvxynhKogULB0MZvF8jiwQX5eOV7sl+ngploviNCKRyAsoABmZElRx8YRgtJyRTKNQaXG-BgkgA00E41VhEAAbioaAATZbeNYRLbwHYQVxgdyeFY+UEBM6hBXik4xeIPTrdPppN4fbQklmk8lKyQVFjcuEKo2NSW8VpazG0sYMybTWbFeb5U1YEVmMUzI727ZOaV7eWXYPHKIXYKq6PXc2a9o6576zLZI3AoGUX1kgE-EQncG4SFWmE2hF2zZNR0YlI1klIwGgtG4RtYoA).

gnprice: > * It’s justified in assuming that the `t` property of `x: { t: 'n' } | { t: 's' }` won’t be mutated by external functions, because mutating it would be a type error in those functions.

Hmm, thanks! I hadn't expected that that would be a type error, but on studying this further, I agree. More in that chat thread.

One observation that helped make sense of this for me is:
```js
(x: {t: 'n'} | {t: 's'}): {+t: 'n' | 's'} => x; // ok
(x: {t: 'n' | 's'}): {+t: 'n'} | {+t: 's'} => x; // error!
```
So if you have a `{+t: 'n'} | {+t: 's'}`, it can't have come from something that had a `{t: 'n' | 's'}`, and was thereby holding on to an option to mutate `t` from 'n' to 's' and back.

PR #5027, opened 2021-09-29
Message-list diffing: some renames

chrisbobbe:
I'm not sure about the new `messageListElementHtml` directory. We have `src/start/IosCompliantAppleAuthButton/index.js`, but we've been wary of this pattern because it can make things annoying in code editors; it means we have to think about this extra name "index.js" and that requirement can be unhelpful… Maybe there's a better way?


src/types.js
```diff
@@ -341,30 +341,30 @@ export type UnreadStreamItem = {|
...
 |};
 
-export type HtmlPieceDescriptor =
-  | TimePieceDescriptor
-  | MessagePieceDescriptor
-  | HeaderPieceDescriptor;
+export type MessageListElement =
+  | MessageListElementTime
+  | MessageListElementMessage
+  | MessageListElementHeader;
```
gnprice, line 367: I keep wanting to read these type names as some kind of time, some kind of message, etc. How about `TimeMessageListElement` and so on?

src/webview/js/generatedEs3.js
```diff
@@ -411,7 +411,7 @@ var compiledWebviewJs = (function (exports) {
   };
 
   window.onerror = function (message, source, line, column, error) {
-    if (window.enableWebViewErrorDisplay) {
+    if (window.isDevelopment) {
```
gnprice, line 414: Hmm, yikes, this global isn't type-checked by Flow. Let's get some kind of type-checking on this (and checking that we have the right variable name, even) by dropping the `window.` and using the same thing we do for `platformOS`. That can go either before or after changing to the new name.

gnprice: Separately:

    And rename the variable in the WebView. In the WebView, it's very
    natural to want to know if we're in development or not. The old
    thing was telling us the same information, but in a roundabout way
    that wasn't explicit.

I agree this new version is better. FWIW I think the idea with the old version wasn't to tell us whether we're in development -- it was to have the flag in `config.js` control specifically what the name here said, i.e. whether to show the error display in the webview.

The reason this new way is better is that there isn't really an advantage to having that controlled in `config.js`, instead of being controlled by the code here at the spot where we're actually doing it. There are possible use cases for something like the `config.js` architecture -- it resembles the `settings.py` architecture that the server uses, which is valuable for lots of people installing the Zulip server -- but that doesn't so much apply here, and I'm not aware of us ever having much real benefit from the `config.js` structure. If someone's editing `config.js`, they can just as well edit the directly relevant code instead.

And conversely, having the logic for the decision here at the spot where we actually execute on it makes for less referring back and forth to figure out what's going on.

gnprice (commented):
Thanks @chrisbobbe! Small comments below. We also talked just now in the office about alternatives to the `messageListElementHtml` directory.

chrisbobbe:

src/message/getMessageListElements.js
```diff
@@ -39,6 +39,16 @@ export default (
...
         key: `header${message.id}`,
+
+        style: isStreamNarrow(narrow)
+          ? 'topic+date'
+          : // eslint-disable-line operator-linebreak
+            // If it's not a conversation narrow (AKA a topic or PM narrow),
+            // we're not making a HeaderMessageListElement in the first place;
+            // see `canShowRecipientHeaders` above. So this case doesn't mean
+            // headers will show up in conversation narrows.
+            'full',
```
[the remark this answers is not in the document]
chrisbobbe: Yep, thanks for catching that!

chrisbobbe:
> Tiny comments below; then please merge at will.

Done, thanks for the review!

PR #5028, opened 2021-10-01
ios: Fix some build issues with Xcode 13.

chrisbobbe:
Here are the error messages:

One issue:

> Undefined symbol: __swift_FORCE_LOAD_$_swiftDataDetection
> Undefined symbol: __swift_FORCE_LOAD_$_swiftFileProvider

The other issue:

> Multiple methods named 'orientation' found with mismatched result, parameter type or attributes

I almost gave this P0, but then I remembered I can still do development work, I just have to do it on Android. 🙂


chrisbobbe:
The first commit

```
ios: Run `pod update DoubleConversion` and `pod update glog`
```

has nothing to do with the build failures, it's just something I noticed we should probably do.

gnprice:
Looks good -- merged! Thanks for tracking those down.

chrisbobbe:
Thanks for the review!

PR #5030, opened 2021-10-04
per-account 3/n: Sort out remaining selectors

gnprice:
This is the next PR in the series after #5016, #5017, and #5023, produced from the branch described at [#5006 (comment)](https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909).

> After this PR, the conversion of `tryGetAuth` is still incomplete, so that's next up. Then we handle a few remaining selectors elsewhere, in particular those for `state.settings` and `state.session`.

That's this PR.

Next, continuing what we mentioned last time:

> After that, we'll convert thunk actions and React components so that they get a PerAccountState rather than a GlobalState.

### DIFF
--- a/src/account/__tests__/accountsSelectors-test.js
+++ b/src/account/__tests__/accountsSelectors-test.js
@@ -1,43 +1,6 @@
 import deepFreeze from 'deep-freeze';
 
-import { getAuth, tryGetAuth } from '../accountsSelectors';
-
-describe('tryGetAuth', () => {
-  test('returns undefined when no accounts', () => {
-    const state = deepFreeze({
-      accounts: [],
-    });
-
-    const auth = tryGetAuth(state);
-
-    expect(auth).toBe(undefined);
-  });
-
-  test('returns undefined when no API key on active account', () => {
-    const state = deepFreeze({
-      accounts: [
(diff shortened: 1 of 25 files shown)


chrisbobbe:
Thanks, LGTM! Please merge at will.

One nit is that "remaining selectors" doesn't quite cover it, I think—remember this bit, from your 59d8813889c69083e66902b140a742e93fd80d66:

> A more boring class of exceptions is the many small anonymous
> selectors (mostly as callbacks to `useSelector`) that get bits of
> the settings and session data. Many of those are really
> per-account, but for now we leave them all marked as global.

gnprice:

> One nit is that "remaining selectors" isn't quite accurate, I think—remember this bit, from your [59d8813](https://github.com/zulip/zulip-mobile/commit/59d8813889c69083e66902b140a742e93fd80d66):
> 
> > A more boring class of exceptions is the many small anonymous
> > selectors (mostly as callbacks to `useSelector`) that get bits of
> > the settings and session data. Many of those are really
> > per-account, but for now we leave them all marked as global.

True. I guess what I really meant is more like "remaining named selectors", or "remaining selectors in `*Selectors.js` files".

Those small anonymous selectors, we'll get to as part of migrating the React components that they're part of.

PR #5033, opened 2021-10-06
BackNavigationHandler: Delete, as it's not doing anything

chrisbobbe:
We tested on two devices (one Android and one iOS), and it looks
like this component wasn't actually doing anything.

In particular, it was meant to change the app's response to the
native go-back interaction (swipe-from-left-edge on iOS; or a
gesture or "hardware" back button on Android) so that it used our
own custom "navigateBack" function, which pops everything off the
stack.

The behavior we observed on those devices was that just one screen
was popped off the stack -- which is the same behavior as if we
didn't have this component in the first place.

We think the pop-just-one behavior is fine, and maybe even nicer.
So, sweep this component away.


gnprice:
Thanks! Looks good; merging, with just a tweak to make a bit of the commit message more specific: s/gesture/swip-from-edge/.

PR #5035, opened 2021-10-07
Fix layout of app bars

gnprice:
The main commit here is:

    app bar: Give buttons the proper size and spacing
    
    This adjusts the size of the app bar's buttons, and the
    horizontal layout of all the buttons and the title, to
    match the Material Design spec:
      https://material.io/components/app-bars-top#specs
    
    The motivating issue was that when there are two "action item"
    buttons at the end of the app bar, the icons were too widely
    spaced apart. […]

This work started from noticing today that excessive spacing. But along the way there, I ended up making a number of refactors in this area of the code, and several other fixes. In particular:

    app bar: Fix height, especially when no back button
    
    These app bars -- all our app bars outside the message list, which
    uses ChatNavBar -- didn't have anything to actually specify how tall
    they should be.
    
    When the back button is present, it serves as a strut to keep the app
    bar at least as tall as the button is, which gets us pretty close:
    just a fraction of a pixel off in my testing on an Android 11
    emulator.  (In fact I don't understand why it isn't exactly right:
    `NavButtonGeneral` sets the height to NAVBAR_SIZE.)
    
    But when the back button is not present, the app bar would collapse to
    the height of the text, which is typically much shorter.
    
    Fix both cases by simply specifying the intended height, as a minimum.
    […]

.

    app bar: Make the padding in ModalNavBar halfway reasonable
    
    We'll fix this up to more precisely match the intended Material
    Design spec in a bit.
    
    But first, just fix up two aspects that were totally absurd:
     * NAVBAR_SIZE, which is 58px, is vastly too much right margin here.
       Make it 16px, the same as the other margin here.
     * On the other hand, when there's no back button, we were using
       no right margin at all.  […]

.

    app bar: Use flex properly in titles, preventing collisions
    
    Before this fix, if you visit a stream or topic narrow with a very
    long stream name, the text of the name will run over on top of the
    button icons that go after it. […]
    
    Fix those by using Flexbox properly so that it can see this is the
    component to put the residual width into, after the fixed amounts for
    buttons and padding and so on. […]


chrisbobbe (commented):
Thanks, @gnprice, this looks good!

I made a few screenshots of the chat screen. See before and after screenshots; for each of those, one with touch targets highlighted and one without.

Before:





After:

Issue #5038, opened 2021-10-07
Improve search bar UI in stream topic view


alya:
We should make a couple of small improvements to the search bar for topics in the stream topic view:

1. Change the placeholder text to "Search topics" for clarity.
2. Bump up the contrast on the placeholder text, which is quite difficult to see right now

anishwar-007:
Is there any way ,i can work on this issue?

alya:
Awesome, this is super helpful @AlaaElden98 ! I think both of the new versions (light and dark) look pretty good in all the contexts above. Please ago ahead and submit a PR.

ArchitJain1201:
@alya Can you assign me this issue ?

rutujapate:
@alya can you assign this issue to me

alya:
@rutujapate Please start by reviewing the attached PR and commenting on whether it looks good or if any changes should be made.

aritroCoder:
Hello I want to contribute in this issue, but I am having a issue while trying to do `yarn install` it says 
```
$ bash tools/postinstall
node_modules/.bin/jetify: 12: node: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
Can you suggest how to fix this? I have also tried `yarn install --force` but result was same

alya:
@aritroCoder Please ask for help in the #mobile team stream in the [Zulip development community](https://zulip.com/development-community/).

codewithbharat:
@alya is this issue still open..... If's yes I would like to work on it.

Komalbelge345:
@alya I’d like to work on this!  
I will update the placeholder to “Search topics”
 and bump up the contrast to make it more readable.  
Please assign me this issue.

gnprice:
@Komalbelge345 As the README says, we're no longer accepting contributions in this legacy app. If you'd like to contribute to the Zulip mobile app, see the new repo linked from the README (https://github.com/zulip/zulip-flutter), and the contributor guide linked from there.

PR #5039, opened 2021-10-07
[SearchInput] Change the placeholder text to "Search topics" for clarity

AlaaElden98:
Fixes part of: #5038


src/common/SearchInput.js
```diff
@@ -44,7 +44,7 @@ export default function SearchInput(props: Props): Node {
         selectTextOnFocus
         underlineColorAndroid="transparent"
         autoCapitalize="none"
-        placeholder="Search"
+        placeholder="Search topics"
```
chrisbobbe, line 47: This changes the placeholder text everywhere `SearchInput` is used, not just in the place we want to update. 😉 The issue #5038 just says "the search bar for topics in the stream topic view".

AlaaElden98: Got it!

AlaaElden98:
@chrisbobbe Thanks for the clarification, it's ready for a review.

src/common/Screen.js
```diff
@@ -47,6 +47,7 @@ type Props = $ReadOnly<{|
   searchBarOnChange?: (text: string) => void,
   searchBarOnSubmit?: (e: EditingEvent) => void,
   shouldShowLoadingBanner?: boolean,
+  +placeholder?: LocalizableText,
```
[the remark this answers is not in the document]
gnprice: Good question. The `+` means this property is read-only:

That is the behavior we want. But the `$ReadOnly<…>` that we have around the whole object type already applies a `+` to each property:

So putting `+` on an individual property is redundant, and we should leave it out.

I see there's a `+` already on some of the properties in this type. [In the Git history](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#reading-history), it looks like we added those before we'd put `$ReadOnly` on the whole thing. Now that we have `$ReadOnly` there, I'll go remove the redundant individual `+` marks.

src/common/Screen.js
```diff
@@ -69,6 +70,7 @@ type Props = $ReadOnly<{|
  * @prop [autoFocus] - If search bar enabled, should it be focused initially.
  * @prop [searchBarOnChange] - Event called on search query change.
  * @prop [searchBarOnSubmit] - Event called on search query submit.
+ * @prop [placeholder] - text shown as place holder, required if `search` is true.
```
gnprice, line 73: It's not true that this is required in that case, is it? After all, we don't pass it in any of the other places where we use `Screen` with `search` true.

(You can find those other places with e.g. `git grep searchBarOn`, because each such screen will want to use at least one of the callbacks `searchBarOnChange` and `searchBarOnSubmit`.)

gnprice: Let's call this prop `searchPlaceholder` here on `Screen`, to help make clear what it's a placeholder for.

OTOH for `SearchInput`, just `placeholder` is an appropriate name.

gnprice (commented):
Thanks @AlaaElden98! Comments below.

Another comment, at the commit level: These changes will be clearest as two commits.
* First, add the new option on Screen, threaded through ModalSearchNavBar to SearchInput. This change should be [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc).
* Then, use the new option at TopicListScreen. (And add the new string in `messages_en.json`.)

src/common/Screen.js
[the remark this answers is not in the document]
gnprice: > Now that we have `$ReadOnly` there, I'll go remove the redundant individual `+` marks.

#5047.

gnprice:
Thanks for the revision! Looks good, with small issues -- fixing those and merging.

* There was one spot in the first commit that still said `placeholder` when it should be `searchPlaceholder`. This was fixed in the second commit, but that still meant that if you check out the first commit, the test suite (`tools/test`) would fail with a Flow error. It's important that the code at every commit be fully working, so I edited the first commit to include that fix.
* I've edited the summary lines to match our style. See `git log --oneline` for a sample of what our summary lines look like, or `git log --stat -p` (and use [the "secret"](https://github.com/zulip/zulip-mobile/blob/5c92f033449e836a6eef058df04076a53965617c/docs/howto/git.md#reading-history) for reading that output) to see them in context.

gnprice:
And merged -- thanks again @AlaaElden98!

One other edit I made before merge: in the commit that fixes (part of) the issue #5038, we want it to mention the issue:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mention-fixed-issue
That way it's easy for people reading the commit history to see what the context was that motivated this change, and also it helps people reading the issue to see the commits that took care of it, or major parts of it.

PR #5040, opened 2021-10-07
Format redux-persist code so it's closer to our style.

chrisbobbe:
And a few small steps at the end toward getting Flow to cover all this code.


src/third/redux-persist/getStoredState.js
```diff
@@ -1,98 +1,93 @@
...
+    return state;
   }
 
-  if (typeof onComplete !== 'function' && !!Promise) {
-    return new Promise((resolve, reject) => {
-      onComplete = (err, restoredState) => {
-        if (err) reject(err)
-        else resolve(restoredState)
-      }
-    })
```
gnprice, line 92: Hmm, this is actually something where I'd be inclined to take it in the opposite direction: this `getStoredState` shouldn't take a callback at all, and should always just return a Promise. That's because it's a function whose whole job is to eventually call that callback, and now that JS has promises and a nice async/await syntax, it's cleaner to write that in the form of `await getStoredState(config)`.

It looks like the story is that we don't call this `geStoredState` function at all in our code outside this library; but where it is called in `persistStore`, that uses the callback form. So probably it's best to defer refactoring that until we have types on this code.

src/third/redux-persist/persistStore.js
```diff
@@ -120,25 +129,25 @@ export default function persistStore (store, config = {}, onComplete) {
     store.dispatch({
       // Include a random string, to be sure no other action type
       // matches this one. Like Redux does for the initial action.
-      type: `PERSIST_DUMMY/${Math.floor(Math.random() * 2 ** 54).toString(36)}`
+      type: `PERSIST_DUMMY/${Math.floor(Math.random() * 2 ** 54).toString(36)}`,
     });
 
-    onComplete && onComplete(err, restoredState)
+    onComplete(err, restoredState);
```
gnprice, line 135: OTOH for `persistStore` (by contrast with `getStoredState` as discussed above), a callback makes sense because the callback is just an incidental hook, not the main point of the function.

Here's the callback we end up passing as this `onComplete`, from `StoreProvider.js`:
```js
    timing.start('Store hydration');
    restore(() => {
      timing.end('Store hydration');
    });
```

Given that, I think the appropriate API is for this one to stay optional. That timing is harmless and cheap, but something we rarely use; it's easy to imagine we might cut it out sometime in the future, and then it'd look silly to pass a no-op callback.

I think that means the net for this commit:

    redux-persist [nfc]: Assume some on-complete callbacks will be passed.

is:
* We'll drop it for now, i.e. from this PR.
* After adding types, we'll do a version of it that's just for `getStoredState`, and makes it never take a callback and always return a `Promise`.

src/third/redux-persist/getStoredState.js
```diff
@@ -1,98 +1,93 @@
...
+        key,
+      });
     }
 
-    return state
-  }
-
-  function complete (err, restoredState) {
-    onComplete(err, restoredState)
-  }
```
gnprice, line 74: getStoredState [nfc]: Inline `complete`.

Huh. I am mystified as to why this indirection was here in the first place. Even looking at the very first version we vendored (with `git log --stat -p --reverse src/third/redux-persist/getStoredState.js`), it seems like it doesn't have any effect and the code would be improved by this inlining.

Perhaps it made more sense either in an earlier version of redux-persist (before the one we vendored from), or in a contemplated future API for it that hadn't been implemented.

In any case, I think this inlining makes just as much sense with the preceding require-`onComplete` change dropped as it does with that change included.

gnprice (commented):
Thanks @chrisbobbe for taking care of these! Comments below, mainly on just one of these commits.

Then one commit-message comment, which applies to much of the branch: I'd like the summary-line prefixes to be somewhat broader.

Saying `redux-persist:` hits the right level of generality: this library as a whole feels to me like a "subsystem" that has a meaningful identity, whereas the individual files in it feel more like arbitrary internal divisions of the implementation details -- they mostly don't mean much to me, and some of them might get rearranged significantly in refactoring this code.

Better yet, I think, is `persist:` (as we discussed in the office yesterday, and is in your latest few commits at the end). That conveys basically the same meaning (there's no competing system in our codebase for "persisting" things), and is shorter :slightly_smiling_face: 

(Some of these commits apply to lots of the files in this library, and have summary prefixes naming just one of them. So for those I'd want a broader prefix even if the individual files did feel like submodules.)

The first commit touches only `.eslintrc.yaml`, and so the prefix `eslint config:` is fine for that (though I usually make it shorter: `eslint:`, or even just `lint:`.) But also it's only touching the redux-persist overrides in that file, so `persist [nfc]:` would also be a fine prefix for it. After all, many of these other commits are at heart about deleting overrides from the ESLint config, and are properly prefixed `persist [nfc]:`; this one is like those, just even more NFC :wink:

gnprice:
Thanks! Looks good -- merged.

PR #5043, opened 2021-10-08
generics: Add tests, doubling as Flow demo; small fixes to BoundedDiff (per-account interlude, 3.1/n)

gnprice:
As the commits' author-dates suggest, this is some work I wrote up a while ago, returning to a couple of times over the past year and a half or so, but kept feeling unsure that it mattered enough to polish up and send.

As it turned out, the version of this branch I had from January needed little more than some squashing to produce a fine branch. I added the small commit at the end, fixing a wart in `BoundedDiff` (working around an underlying Flow bug) that the preceding commit's testing had exposed; added the last comment at the end of `test_BoundedDiff`; and made a couple of tiny other edits.

The immediate prompt for looking back at this branch and turning it into a PR is that the next part of my draft for distinguishing per-account vs global data, for the next PR after #5030, contains a small change to BoundedDiff. I'd like to write tests for that change, and those should naturally go on top of having tests for BoundedDiff at all.

---

commit f0ba576951cafb33e878afee1148dc6806dae5de
Date:   Thu May 7 14:58:14 2020 -0700

    generics tests: Test IsSupertype
    
    In addition to testing our `IsSupertype`, these tests double as:
    
     * A demo of some Flow type-system features like `empty`, `mixed`,
       and read-only vs. invariant/read-write object properties.
    
     * A demo of some techniques for writing type-tests.

commit 51e8c26bac7b4e5d19769f326c60974b10a7e468
Date:   Thu Aug 6 06:52:57 2020 -0700

    generics: Add typesEquivalent, with tests

commit fb5836daf879e266e46e3910567dcba9eebf1667
Date:   Tue Jan 26 21:06:09 2021 -0800

    generics [nfc]: Test BoundedDiff; fix some details in its jsdoc
    
    Turns out the resulting type is actually exact, not inexact as the
    jsdoc had said!  That's behavior inherited from $Diff.
    
    Not sure if that changed at some point since I originally wrote this,
    or if it was always this way -- could certainly be the latter, as
    after all this didn't have tests. ;-)
    
    Also don't try to characterize the behavior if the arguments aren't
    read-only.  The read-only case is the only one we care about, and we
    don't try to write tests for beyond that.

commit c2a2dadfcfd1c6b9973fe72a4e7e06fd1ddee0fa
Date:   Wed Oct 6 13:16:00 2021 -0700

    generics: Fix variance in BoundedDiff result, working around Flow bug
    
    We had been working around this at every place we use BoundedDiff.
    Instead, put the workaround inside BoundedDiff itself, because it's
    needed for getting the behavior we actually want.


src/__flow-tests__/generics-tests.js
```diff
@@ -0,0 +1,272 @@
...
+
+  // Exact <: inexact, strictly.
+  (a: IsSupertype<{ a: number, ... }, { a: number, ... }>);
+  (a: IsSupertype<{| a: number |}, {| a: number |}>);
+  (a: IsSupertype<{ a: number, ... }, {| a: number |}>);
+  // $FlowExpectedError[incompatible-exact]
+  (a: IsSupertype<{| a: number |}, { a: number, ... }>);
+
+  // Read-only properties are covariant.
+  (a: IsSupertype<{| +a: number |}, {| +a: empty |}>);
```
chrisbobbe, line 80: Could put, e.g., `{| +a: 3 |}` instead of `{| +a: empty |}`, since we don't mean to invoke all the reasons why `empty` is special here.

gnprice: Hmm, well, this line is meant to be read together with the one below it (which make up a little stanza):
```js
  // Read-only properties are covariant.
  (a: IsSupertype<{| +a: number |}, {| +a: empty |}>);
  // $FlowExpectedError[incompatible-type-arg]
  (a: IsSupertype<{| +a: number |}, {| +a: mixed |}>);
```

The types `empty` and `mixed` are equally special, and in inverse ways: `empty` is a subtype of everything (can flow to anything), and `mixed` is a supertype of everything (anything can flow to it.)

So it's true that `empty` is special, but that specialness is really just a pure form of what this test needs here: it wants some type T that's clearly a subtype of `number`, in order to test that that implies `{| +a: T |}` is a subtype of `{| +a: number |}`. (And more generally: if `S` supertype `T` then `{| +a: S |}` supertype `{| +a: T |}`.)

We saw in `test_IsSupertype` above that if you have `empty` appearing at the input end of a data flow, then Flow may stop caring what type is at the output end, and not notice even if that latter type has a type expression like `IsSupertype<empty, number>` that should be an error. But it's equally true that if you have `mixed` appearing at the output end of a data flow, then Flow may stop caring what type is at the input end, and not notice even if *that* type has a type expression that should be an error.

Here, the `empty` is flowing simply to `number`. That will succeed, because `empty` can flow to anything. The test demonstrates that as a result, `{| +a: empty |}` can flow to `{| +a: number |}` (and that `IsSupertype` accurately captures that behavior.) On the other hand, `{| +a: mixed |}` can't flow to `{| +a: number |}`. That's because that would work only if `mixed` could flow to `number`, and it can't. That's even though if we needed the flow in the *other* direction, from `number` to `mixed`, it would succeed, because anything can flow to `mixed`.

This isn't real clear in the existing version of this file, I think. The discussion of `empty` at the top is a little overbroad, and there isn't a demonstration of how `mixed` can produce the same kind of short-circuit behavior when it's simply at the output end of the flow instead of the input end. I think I probably didn't understand that symmetry at the time I originally wrote that text; I spent some time digging into Flow this past December and January, and understand it a lot better now than I did before.

I think I'll merge this as is in this respect, but then try to make a further change to clarify.

gnprice: #5053

Issue #5020, opened 2021-09-23
User's own name is missing from "share to Zulip"


alya:
When sharing to Zulip, the user's own name is missing. This is problematic, as sharing a file with yourself is a reasonable workflow for uploading the file (e.g. a photo) from the phone to the computer. 

As @gnprice [describes in the CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/share.20to.20Zulip/near/1258427), the cause of that in the code is that our UserPickerCard component filters out yourself:

```
// The users we want to show in this particular UI.
// We exclude (a) users with `is_active` false; (b) cross-realm bots; (c) self.
const getUsersToShow: Selector<User[]> = createSelector(
  getUsers,
  getOwnUserId,
  (users, ownUserId) => users.filter(user => user.user_id !== ownUserId),
);
```

We need the behavior to be different for this context.

PR #5044, opened 2021-10-09
Show own user when sharing to zulip

AlaaElden98:
Fixes : #5020


src/user-picker/UserPickerCard.js
```diff
@@ -41,9 +42,9 @@ const getUsersToShow: Selector<User[]> = createSelector(
 );
 
 export default function UserPickerCard(props: Props): Node {
-  const { filter } = props;
+  const { filter, showOwnUser } = props;
 
-  const users = useSelector(getUsersToShow);
+  const users = useSelector(showOwnUser ? getUsers : getUsersToShow);
```
chrisbobbe, line 47: If we add a case here where `getUsersToShow` is not used, then `getUsersToShow`'s name becomes inaccurate, as well as the following comment on it:

```
// The users we want to show in this particular UI.
```

AlaaElden98: My first thought was to change the name of `getUsersToShow` to `getUsersExceptSelf` or something better and shorter, then create a function that takes showOwnUser as a parameter if true returns`getUsersExceptSelft` otherwise return `getUsers`.
and add the comment you mentioned above the function

AlaaElden98:
> Also, could you say more about how you've addressed (or would like to address) this comment from Greg, especially his last paragraph here:
> 
> @gnprice [said](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/share.20to.20Zulip/near/1258436):
> > Oh, though also one pitfall to watch out for when implementing this is to just make sure that those two possibilities _do_ produce the same result.
> 
> Thanks again! 🙂

So, I just notice that if I select myself and another person the message will not be sent to me!
The only case this works is if the user selects only himself to share the file with. 
I will investigate more.

src/user-picker/UserPickerCard.js
chrisbobbe: @gnprice, is the `createSelector` in `getUsersToShow` an instance of #3927?

Issue #5018, opened 2021-09-22
Make typeahead boxes more visible


alya:
At present, typeahead boxes are not visually distinct from the background in night mode. This is especially problematic when the box opens automatically when the user does not expect it, as it now does for topic typeahead when sending a message from an interleaved view.

Possible solutions are to give the box more of a border or a different background color. The light mode typeahead box has a more clear boundary, but perhaps would benefit from more distinct styling as well.

Screenshot:
![Screenshot_20210921-175759_Zulip](https://user-images.githubusercontent.com/2090066/134291355-6cf8f918-40c5-4edf-9287-776c10976657.jpg)

[CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/compose.20topic.20and.20navigation)

Jumeb:
Working on this.

Jumeb:
@zulipbot claim

Jumeb:
> Thanks @Jumeb for picking this up and sharing the screenshots! In the dark mode ones, it looks like the blue border overlaps with the contents. Could we have some margin between the border and the text?
> 
> I also wonder if there's something we can do to make it feel more like the auto-complete box is on top of the other content. Perhaps playing with the background color?



> > I also wonder if there's something we can do to make it feel more like the auto-complete box is on top of the other content. Perhaps playing with the background color?
> 
> I like that idea! It would be great to figure out what [Material Design](https://material.io/design) would want us to do here. slightly_smiling_face

There are two options from [Material Design](https://material.io/design)
1) Change the background color
2) Increase the depth perception by increasing the elevation
I would love to go with the second, reason being, the background of the component is controlled using the useContext hook and the only way will be to modify the background color in the theme.js file which will affect other components.
As for the overlap, I will love to put slight padding in the Popup.js component for the contents to avoid the corners.

Jumeb:
> > Thanks @Jumeb for picking this up and sharing the screenshots! In the dark mode ones, it looks like the blue border overlaps with the contents. Could we have some margin between the border and the text?
> > I also wonder if there's something we can do to make it feel more like the auto-complete box is on top of the other content. Perhaps playing with the background color?
> 
> > > I also wonder if there's something we can do to make it feel more like the auto-complete box is on top of the other content. Perhaps playing with the background color?
> > 
> > 
> > I like that idea! It would be great to figure out what [Material Design](https://material.io/design) would want us to do here. slightly_smiling_face
> 
[...]

Issue #4872, opened 2021-07-06
Open typeahead immediately when @ is entered


alya:
At present, we open @-mention typeahead suggestions when the user enters `@` (not within a word) followed by another character. Instead, we should pop up typeahead right away without waiting for the first character. This will make it clear to the user that they can use @-mentions in Zulip, and is consistent with expectations from other chat apps (e.g. Slack, Discord).

alya:
Implemented in the web app: https://github.com/zulip/zulip/issues/19142

alya:
Yeah, I think because of `:` and `#` potentially being used at the start of unrelated words (esp. `#` for issue numbers), let's just do this for `@`. That also has the advantage of being consistent with the webapp. thanks!

PR #5046, opened 2021-10-12
typeahead:Open the typeahead box immediately the '@' character is entered

Jumeb:


Jumeb:
Yes, I wanted to make two separate PRs
1) The first to solve issue [#5018](https://github.com/zulip/zulip-mobile/issues/5018)
2) The second to solve issue [#4872](https://github.com/zulip/zulip-mobile/issues/4872) 
But, I messed up by not creating separate branches for each solve, that is why I have duplicated work done.

chrisbobbe:
Cool, that makes sense. Can you tell from your end whether GitHub will let you change the PR branch for this PR? Maybe with the "Edit" button next to the PR's title? Otherwise, it may be easiest to make a new PR with a new branch, so we can separate the work being done for the two different issues. Thanks!

PR #5047, opened 2021-10-12
types [nfc]: Remove some redundant read-only-property annotations.

gnprice:
Each of these `+` marks means the property is read-only:

In each case that's the behavior we want.  But in each case we
already have a `$ReadOnly<…>` around the whole object type, which
applies a `+` to each property:

So putting `+` on an individual property is redundant.  Take it out,
to avoid confusing people reading the codebase and trying to
understand what the various type annotations mean:
  https://github.com/zulip/zulip-mobile/pull/5039#discussion_r724636461

(In each of these cases, we added the `+` marks at a time when there
wasn't a `$ReadOnly` there, and then added the latter as part of a
broad automated refactoring.)

### DIFF
--- a/src/api/transportTypes.js
+++ b/src/api/transportTypes.js
@@ -31,8 +31,8 @@ export type Auth = $ReadOnly<{|
  *  * {@link ApiResponseErrorData}
  */
 export type ApiResponse = $ReadOnly<{
-  +result: string,
-  +msg: string,
+  result: string,
+  msg: string,
   ...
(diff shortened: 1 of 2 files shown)


chrisbobbe:
LGTM, thanks! Please merge at will, assuming tests pass (which I confidently predict they do, but CI isn't finished yet and I haven't pulled the branch to look at it locally).

gnprice:

PR #5048, opened 2021-10-12
msglist-diffing: Add and use new '.msglist-element' class in WebView HTML

chrisbobbe:
(And fix an unrelated bug where a _**[bold, italic link](https://www.chat.zulip.org/)**_ in a Zulip message failed to open when you tapped on it.)


gnprice:
Thanks! Looks good -- merged.

chrisbobbe:
Thanks for the review!

PR #5052, opened 2021-10-14
Search for more than 20 messages.

AkashDhiman:
~currently WIP~

Changed `fetchNewer`, `fetchOlder` to be callbacks passed on to `MessageList` as suggested in https://github.com/zulip/zulip-mobile/issues/4859#issuecomment-874977103.

TODO:
- [x] implement callbacks in search message screen.
- [x] fix failing tests.

Fixes: #4859


gnprice:
* I don't understand why I need to do `await await` when
      dispatching `fetchMessages`,

Yeah, that is weird. Sure looks like `dispatch` should be returning a `Promise` of an object type, and then one `await` should get you that object type. Kind of looks like a possible Flow bug.

The `await await` should be harmless at runtime, anyway.

      this only happens because I am
      calling setState in the next line.

This part is basically a version of the same principle of Flow described here:
https://github.com/zulip/zulip-mobile/blob/41dd6a7b72944920e6e7aaa14da64e9691b9917f/src/__flow-tests__/generics-tests.js#L7-L20

If (after removing the second `await`) you comment out the `setState` call, but add back a line like this:
```js
        (foundOldest: boolean);
```
then you get the same error.

Though really I think if that principle is coming into play here, that's more evidence that a Flow bug is afoot. If Flow is going to think it's invalid to get the `foundNewest` property off that value returned by the first `await`, it should be invalid already, even if we don't go on to try to use that value elsewhere.


    * Is there a Need for `fetchNewer` here?

Yeah, I think there isn't.

That may also let you stop tracking `caughtUp.newer`, which may let you not bother with looking at `foundNewest` (though `fetchMessages` should still return it, for symmetry with `foundOldest`.)


    * Scrolling up a lot and then coming back down, causes a glitch in
      screen which makes the messagelist content blank possibly due to
      large message size in state.

Can you reproduce the same thing already in a stream narrow?

If you're scrolling really a *lot* a lot, then this is something I've seen sometimes. It happens reliably if I intentionally provoke it -- like fling-scroll up again, and again, and again, in `#test here` -- but have never seen in normal usage. Certainly would be good to fix.


    * foundOldest, foundNewest can be `undefined` need to handle these
      properly.

It looks like that was possible only before Zulip Server 1.8. See our jsdoc on `api.getMessages`:
```js
/**
 * See https://zulip.com/api/get-messages
 *
 * These values exist only in Zulip 1.8 or newer:
 *   * found_anchor
 *   * found_newest
 *   * found_oldest
 */
```

That's already older than we support, so we can actually already assume those are present: we can update the types in `src/api/messages/getMessages.js`, propagate that to where those are used, and in particular simplify the existing fallback logic in `caughtUpReducer.js`.

gnprice:
> That's already older than we support, so we can actually already assume those are present

Sent #5100 to do that. Once that's merged and you rebase on it, these will be booleans.

chrisbobbe:
> Sent #5100 to do that. Once that's merged and you rebase on it, these will be booleans.

(And that's now been merged! 🎉)

AkashDhiman:
@gnprice, I think this should be ready for a proper review now.

I have removed some failing test since they corresponded with the `fetchOlder`, `fetchNewer` function in `fetchActions.js` which are no longer present. I am curious if am required to write these for `ChatScreen` and `SearchMessageScreen`, if yes then the fact that now `fetchOlder` and `fetchNewer` are methods of a component and passed in as `props` makes it a bit confusing for me to figure out how I can do this.

Also in general I am failing to understand when I have to write tests for the things that I implement, what cases they are required to cover, when to modify or discard a test.

AkashDhiman:
> ```
> * Scrolling up a lot and then coming back down, causes a glitch in
>   screen which makes the messagelist content blank possibly due to
>   large message size in state.
> ```
> Can you reproduce the same thing already in a stream narrow?

Yes filed an issue here #5124

gnprice:
> Also in general I am failing to understand when I have to write tests for the things that I implement, what cases they are required to cover, when to modify or discard a test.

Yeah, good question. It's a subtle one -- we don't have a bright-line rule for it.

Here, I'm happy to let these tests get deleted, because they really weren't testing much. (They date to 2017, from bc14f0d35afa7a52e589c9bc394ed93c311d80d7.) Nearly 200 lines, and they just tested that the conditions `!loading && !fetching.older && !caughtUp.older` (and `.newer` respectively) were present, that a `MESSAGE_FETCH_START` got dispatched, and that the two had `numBefore` vs `numAfter` the right way around. (And not even quite that -- the tests would have accepted passing positive values for both of those in both places.)

If it were important to have tests for this code, we could do it a lot more compactly and readably. See `unreadModel-test.js` for some tests in a better style. (Or 9d08464d1765ba9522cb4ece40b3481f5eedc21b in `recipient-test.js`, or the `isInternalLink` tests at the top of `internalLinks-test.js`, for tests much more compact than that… but in a way that may only be possible for some kinds of functions, with relatively crisp-yet-tricky semantics.)

Some projects, including zulip/zulip, set a goal of 100% test coverage, and then it's necessary to test this kind of code and to try to do so reasonably cleanly. But for something like `!loading && !fetching.older && !caughtUp.older`, where all you can really say in a test is just to restate that those three conditions should be present, I don't think there's much value in having that test -- the only reason to do it is as part of maintaining 100% coverage, for the sake of being sure you get testing in other places where it's more useful but one might otherwise let it slide.

Generally I'll want to have tests for code that's tricky, or especially where there's something you can say to describe its interface that's simpler than just restating the logic that's supposed to be in it.

The existing async fetching and state-updating code in `SearchMessagesScreen` is actually a prime example of code that's tricky enough it should really have tests. The history of subtle bugs in it illustrates that: e.g. e2e9eed7be86cfb32bbd3ac890e3c96a5ab4624b. We lack tests for it basically because of a combination of: it's annoying and tricky to write tests for async code (but we have done so, e.g. `async-test.js`); and we don't currently have any tests for React components (but React totally has facilities for doing so, we just haven't used them.)

src/search/SearchMessagesScreen.js
```diff
@@ -131,6 +149,35 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+            narrow: SEARCH_NARROW(this.state.query),
+            numBefore: config.messagesPerRequest,
+            numAfter: 0,
+          }),
+        );
+        // Last fetched message === first current message, hence remove it
+        fetchedMessages.pop();
+        this.setState(prevState => ({
+          messages: [...fetchedMessages, ...prevState.messages],
+          foundOldest,
```
gnprice, line 169: There's a tricky bug here. We're updating the state to add the messages we just fetched… but the query we fetched them for may no longer be the current query, the one that the other messages we're showing are from. That's because when we did the `await` above to wait for the results, the user may have entered a new query and those results may have come in. (Or for that matter a query they'd already entered, for which we had a request already in flight, may have had its results come in.)

A first attempt at fixing that would be to have the `setState` callback check that the current query is still the one that we did this fetch for, and if not then drop the update. So for example:
```js
    const { query } = this.state;
    const { messages: fetchedMessages, foundOldest } = await await this.props.dispatch(
        // …
        narrow: SEARCH_NARROW(query),
    // …
    this.setState(prevState => {
      if (prevState.query !== query) return {};
      return {
        messages: [...fetchedMessages, ...prevState.messages],
        foundOldest,
      };
```

I think that still isn't quite right, though, because it's possible that the query has changed and changed back, in which case the list of messages may no longer be the same -- so there might be a gap, or even an overlap, between the list of messages that's now in the state and the messages we've just fetched.

I think a sufficient fix would be to add a second condition, checking that the anchor message we fetched from is still the first message in the list in the current state: `prevState.messages && prevState.messages[0]?.id === anchor`, where `anchor` is a value computed before the `await` and passed to `fetchMessages`. (In addition to checking that the query is the same.)

An alternate approach would be to look at `this.lastIdSuccess` before the `await`, save that in a local variable, and check here that it's still the same. That would cover things completely, and it would make this logic mesh more with what's in `handleQuerySubmit`. But I kind of like having this *not* care about the implementation of `handleQuerySubmit` -- I think I like better the approach of checking `query` and `messages[0].id`, because I think that directly covers everything this fetch-older code actually cares about here.

src/search/SearchMessagesScreen.js
```diff
@@ -131,6 +149,35 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+        }));
+      } finally {
+        this.setState({ isFetching: false });
```
gnprice, line 172: This uses the same state property that's managed by `handleQuerySubmit` to describe whether there's an ongoing fetch for a new search query, and I think it doesn't interact correctly with that -- in particular it can clear `isFetching` when in fact there's still a fetch ongoing for a new search query.

In general it's probably quite tricky to have this managed by both this code and also `handleQuerySubmit`. It's easier to reason about when there's just one pair of lines setting and clearing it.

Fortunately I think there isn't actually a need to have this operate on the same bit of state -- the one existing place that consumes `isFetching`, in `SearchMessagesCard`, only cares about it if there are no messages in the current state, so it should never even notice whether `fetchOlder` has set it. We do want this `fetchOlder` to have a flag that allows it to avoid running twice concurrently, but we can just make it a separate thing.

src/search/SearchMessagesScreen.js
```diff
@@ -131,6 +149,35 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
     this.handleQuerySubmit(e);
   };
 
+  fetchOlder = async () => {
+    if (!this.props.loading && !this.state.foundOldest && !this.state.isFetching) {
+      this.setState({ isFetching: true });
```
gnprice, line 154: I'm also a bit nervous about the flag that's used as a mutex here, the `isFetching`, be a React state property. The thing about those is that `setState` may not update it immediately -- so I'm not sure it would be the case that we can't have a second call to this function see `!this.state.isFetching`, and enter this block and start a second request, while a first call is still in the middle of it too.

That can be fixed by making this use a direct property on the object, just like the existing `this.lastIdSuccess` etc. (We'll still want `this.state.isFetching` for the purpose it already serves, but this `fetchOlder` logic will use a new `this.isFetchingOlder` or whatever instead.)

---

There's a second issue around this mutex logic, which is that the query can change. In this version, if we have an outstanding `fetchOlder` request for a previous version of the query, and it's hanging for a while for whatever reason, then that'll stop us from fetching older messages for a new query until it finally finishes.

So instead, we can remember which query it is that we're fetching older messages for. There may be several of these, but all we really care about is whether we're doing so for the latest query -- so that can be summarized by another `lastIdFoo` property, similar to the three we have. That is, we can say: … Hmm, this gets fairly tricky. I think it should work out, but it'll take some careful looking at the code, and that's also a hint that maybe there's a better way to architect this.

This second issue is lower-priority, because it can't produce wrong results, only delayed results. And it isn't a super likely case to happen. So I'll be OK with merging a version of this PR that just uses a single boolean like `this.isFetchingOlder`, and then trying to work that out later.

gnprice (commented):
Thanks @AkashDhiman! This is looking good, except for some subtle issues in maintaining the state as things happen concurrently. See comments below.

gnprice:
Hi @AkashDhiman! Do you think you'll be up for returning to this PR and completing it? There's good work here, and it'd be great to get it merged.

Rereading my last comments above, there's a couple of easy nits in the prep commits and then several more interesting issues in the main commit at the end. But for each of those it looks like there's a concrete idea for the direction to take to resolve it.

AkashDhiman:
Hi @gnprice, I will most likely be able to make time for this from January not before that. If things change, I will make sure to inform.

gnprice:
Sounds great -- looking forward to it. Thanks!

AkashDhiman:
update: Apologies for the delay, I have started working on this again, expect the revision to come soon enough.

src/search/SearchMessagesScreen.js
```diff
@@ -153,6 +202,7 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
 
 const SearchMessagesScreen: ComponentType<OuterProps> = connect(state => ({
   auth: getAuth(state),
+  loading: getSession(state).loading,
```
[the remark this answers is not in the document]
AkashDhiman: done

src/message/fetchActions.js
```diff
@@ -111,7 +102,9 @@ export const fetchMessages = (fetchArgs: {|
   anchor: number,
   numBefore: number,
   numAfter: number,
-|}): ThunkAction<Promise<Message[]>> => async (dispatch, getState) => {
+|}): ThunkAction<
+  Promise<{| messages: Message[], foundNewest: boolean, foundOldest: boolean |}>,
+> => async (dispatch, getState) => {
```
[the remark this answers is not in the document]
AkashDhiman: done

src/search/SearchMessagesScreen.js
```diff
@@ -131,6 +149,35 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+        const { messages: fetchedMessages, foundOldest } = await await this.props.dispatch(
+          fetchMessages({
+            anchor: this.state.messages ? this.state.messages[0].id : LAST_MESSAGE_ANCHOR,
+            narrow: SEARCH_NARROW(this.state.query),
+            numBefore: config.messagesPerRequest,
+            numAfter: 0,
+          }),
+        );
+        // Last fetched message === first current message, hence remove it
+        fetchedMessages.pop();
```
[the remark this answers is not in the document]
AkashDhiman: Not using `LAST_MESSAGE_ANCHOR` here anymore. I have added an invariant call here, since flow was complaining when I removed `LAST_MESSAGE_ANCHOR`, given that we always expect `this.state.messages` to be not null. Let me know if that's not a correct assumption.

src/search/SearchMessagesScreen.js
AkashDhiman: > There's a tricky bug here.
Indeed!
I have gone with the first approach, did test it using a fake delay and it does work out as far as I could tell.
```javascript
const sleep = ms => new Promise(
  resolve => setTimeout(resolve, ms));
``` 
Before, there were overlap between messages of different queries - if I changed queries in the middle of fetch, now that doesn't happen. A UX related problem that I didn't address is that the moment user presses the search button we should clear away the messages array and the loading spinner should again be visible, that is not the behavior we have currently.

src/search/SearchMessagesScreen.js
AkashDhiman: added a non react state `isFetchingOlder` to address this.

src/search/SearchMessagesScreen.js
AkashDhiman: the delay still happens in the latest revision, i.e. if I search foo1, then switch to foo2, then to foo3, the page renders foo1 then moves to foo2 then to foo3 which can be a little annoying.

AkashDhiman:
Hi @gnprice, I think you can take a look at this now.

gnprice:
Excellent, thanks @AkashDhiman! Looking forward to reviewing this next week.

gnprice:
I ended up spending the week focused hard on [a different issue](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flutter.20scrolling/near/1502128) and didn't get to this, sorry. Haven't forgotten about it, and will plan to return next week.

src/search/SearchMessagesScreen.js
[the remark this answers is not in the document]
gnprice: Sounds good.

I *think* it will always be non-null, but I'd have to stare at this concurrency logic harder to be certain there isn't a case where that could fail. Above I'd said:

> this callback should only get invoked when there are already some messages in the first place, and if there somehow aren't, it can just return without doing anything. That's what the old `fetchOlder` for `ChatScreen` does, after all.

which would mean to add that check to the early return at the top of this function. But in this case there's no practical difference to the user between this function returning early and throwing — its call site, the `props.fetchOlder()` in `handleOutboundEvents`, doesn't await it so won't know the difference. The only difference is that throwing will cause a Sentry event to be logged (because the Promise rejection will go unhandled.) And that seems for the best given that we believe the case where it's null here shouldn't be possible.

src/search/SearchMessagesScreen.js
```diff
@@ -130,6 +154,50 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+          narrow: SEARCH_NARROW(this.state.query),
+          numBefore: config.messagesPerRequest,
+          numAfter: 0,
+        }),
+      );
+      this.setState(prevState => {
+        if (
+          prevState.query === query
+          && prevState.messages
+          && anchor === prevState.messages[0].id
```
gnprice, line 184: Flow will tell you to drop the `?.` from my suggested condition at https://github.com/zulip/zulip-mobile/pull/5052#discussion_r755599428 , but it's needed because the array can be empty. So:
```suggestion
          // FlowIssue: ?. is needed because array can be empty
          // flowlint-next-line unnecessary-optional-chain:off
          && prevState.messages?.[0]?.id === anchor
```

(This is basically a bug in flowlint. It stems from Flow deciding to let `foo[i]` where `foo: Array<T>` return `T` rather than `T | void`, which is an understandable pragmatic choice given that programmers usually check the bounds first before getting the element, but is unsound.)

src/search/SearchMessagesScreen.js
```diff
@@ -130,6 +154,50 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+    ) {
+      return;
+    }
+
+    invariant(this.state.messages !== null, 'must have atleast 1 message already fetched.');
+    this.isFetchingOlder = true;
+    try {
+      const { query } = this.state;
+      const anchor = this.state.messages[0].id;
+      const { messages: fetchedMessages, foundOldest } = await await this.props.dispatch(
```
gnprice, line 172: This `await await` looks real weird. :-) You noticed that too when you first wrote this code, and I looked into it:
https://github.com/zulip/zulip-mobile/pull/5052#issuecomment-960296275
and it seems to be some harmless Flow bug.

I'd forgotten all about that, though, and it caught my attention again. Let's add a comment so the next reader knows what we know:
```suggestion
      // FlowIssue: Flow insists on this `await await`.  Should be equivalent to `await`, so harmless.
      //   See: https://github.com/zulip/zulip-mobile/pull/5052#issuecomment-960296275
      const { messages: fetchedMessages, foundOldest } = await await this.props.dispatch(
```

src/search/SearchMessagesScreen.js
```diff
@@ -62,11 +70,17 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+  ): Promise<{
+    foundNewest: boolean,
+    foundOldest: boolean,
+    messages: $ReadOnlyArray<Message>,
+  }> => {
     const fetchArgs = {
       narrow: SEARCH_NARROW(query),
       anchor: LAST_MESSAGE_ANCHOR,
-      numBefore: 20,
+      numBefore: config.messagesPerRequest,
```
gnprice, line 83: Let's do this as its own little commit before the main one. It's an important and useful user-visible change on its own, logically independent of the main change, and it deserves its own commit and commit message.

I think basically the point here is that now that we're no longer trying to do a search on every keystroke, there's no need to be hyper-frugal on how much we fetch up front. By the time we actually do a search request, the user has said this is a search they actually want the results for, just like they might by navigating to some other narrow like a conversation. So we should fetch the same number of messages as for other narrows — enough so that they're not just going to immediately want more.

src/search/SearchMessagesScreen.js
```diff
@@ -130,6 +154,50 @@ class SearchMessagesScreenInner extends PureComponent<Props, State> {
...
+        }
+        return {};
+      });
+    } finally {
+      this.isFetchingOlder = false;
+    }
+  };
+
+  fetchNewer = async () => {
+    // Implementing this should not be necessary.
```
gnprice, line 198: Let's have this throw — it should be impossible for it to get called, and that will let us know if that expected invariant fails.

src/search/SearchMessagesScreen.js
gnprice: With plain `.id` instead of `?.id`, this would blow up if the query successfully completed and found no results — that's what an empty `state.messages` means.

That case is not easy to trigger, because this comes after checking `prevState.query === query` — so the current query is the same one that this `fetchOlder` was started for — and that query must have previously returned some messages for us to get as far as calling `fetchOlder`. (I think; see first comment of this review, above. But once we expand the invariant as per that comment, then the query definitely must have previously returned some messages for us to get past that check.)

But it's a case that can happen. It just needs the query's results to have changed between when we originally tried it (and got some messages that the user scrolled up past to trigger calling `fetchOlder`), and when we tried it again (after changing the query and changing it back, all while this `fetchOlder`'s request was pending.) Specifically, there must have previously been some results, and then those messages disappeared from the search results — they were deleted, or were moved or edited so they no longer matched, or were in a stream we lost access to, maybe some other possibilities.

gnprice (commented):
Thanks @AkashDhiman for the revision (and for originally writing this!) The code looks great. I also played with the new UI, and it works very nicely — a huge improvement over the existing search UX.

A few comments below, all small. Do you think you'll have time to follow up on them? Then I'll be looking forward to getting to merge this.

> A UX related problem that I didn't address is that the moment user presses the search button we should clear away the messages array and the loading spinner should again be visible, that is not the behavior we have currently.

Seems right. I'll file a separate issue for that further improvement. (→ #5659)

src/search/SearchMessagesScreen.js
gnprice: > A UX related problem that I didn't address is that the moment user presses the search button we should clear away the messages array and the loading spinner should again be visible, that is not the behavior we have currently.

Good point, yeah. Filed that as #5659 .

(As it sounds like you know, but just to make it explicit for anyone reading this quickly: that UX problem exists already in the current version of the app.)

src/search/SearchMessagesScreen.js
AkashDhiman: I don't think it is impossible for it to get called, in fact it does get called a lot of times, based on the condition for it to get triggered in `handleOutboundEvents.js`
```javascript
  if (innerHeight + scrollY >= offsetHeight - config.messageListThreshold) {
    props.fetchNewer();
  }
```
As far as I could understand, This gets called whenever we are close enough to the bottom (latest) of the messages in our search screen, so throwing an error would not be practical unless we change our conditions.

I have left this unchanged, let me know what you think about this.

AkashDhiman:
@gnprice I did the required changes, you may check again.

src/search/SearchMessagesScreen.js
gnprice: Mmm, I see, right. It's this callback's responsibility to see that there's nothing to do, just like how in `fetchOlder` we return earlier if `this.state.foundOldest`.

The comment should just make that clear, then. When I see "should not be necessary", I immediately wonder *why* it shouldn't be necessary — what assumption this code is relying on when it chooses not to do anything.

I'll revise that bit, and then go ahead and merge.

gnprice:
And merged! Thanks again @AkashDhiman for all your work on this, and for pushing it to the finish line.

PR #5056, opened 2021-10-18
outbox: Make stream_id required on StreamOutbox

gnprice:
This completes the last step of #3998 by making this property required, now that the code which started supplying it (0c251bfe0 / #5000) has been out to users for a few weeks (in v17.171 which rolled out 2021-09-23) and the bulk of our users have upgraded.

We also take care of the two specific followups mentioned in #3998. There's still the more general followup of carrying on with #3918, converting uses of stream names to stream IDs.

Fixes: #3998


gnprice:

Issue #4659, opened 2021-04-14
Turn `api.foo(auth, …)` into fully featured `makeApi(auth).foo(…)`.


chrisbobbe:
This implementation plan is on the path to #4170. It will expose methods to cancel in-progress fetches, and for #4170 we'd just have to call those methods in the right places.

I've put together this plan from discussion with Greg in [chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3745.20Server.20version.20to.20Sentry/near/974729) (@gnprice, please edit this plan liberally as needed! 🙂):

> * Just like `fetch` itself, [something like `api.sendMessage`] can take as an optional argument (or really, as an optional property in an optional argument which is an options-object) a `signal: AbortSignal`.
> * If so, then before we start the `fetch` request, we'll listen on *that* AbortSignal, with a handler which will in turn abort the AbortController whose signal we're going to pass to the actual `fetch` request.

> IOW we can make `api.sendMessage` and its friends behave in the same way that APIs in the web platform itself are supposed to behave:
https://dom.spec.whatwg.org/#abortcontroller-api-integration
> (which includes details […] like if the signal you're passed is already aborted, then reject immediately.)

This is an area where I expect we'll be able to fruitfully write solid tests; see #4193 for examples of testing with Jest's fake timers.

We've never used React Native's `AbortController` polyfill (from facebook/react-native#24419, released in v0.60) before. There may be surprising differences from its standard, documented behavior. Anyway:

- [ ] Make the pure refactor to replace `api.foo(auth, …)` with `makeApi(auth).foo(…)`.
- [ ] Have `makeApi` implement a cache, keyed by the email, realm, and API key from `auth`. We want `makeApi` to return the same object for the same auth, and a different object for a different auth.
- [ ] Allow canceling all in-progress fetches at once:
  - When `makeApi(auth)` is called, have it create an empty array or set called something like `abortControllers`. `abortControllers` will be maintained so that it always comprises one item per in-progress request.
  - When it's time to do a `fetch`, create a new [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) object and pass its [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) in the `fetch` call (note `fetch`'s [`signal` param](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters)). Add the `AbortController` object to `abortControllers`.
[...]

PR #5059, opened 2021-10-19
api: Adapt to API change in update-stream, conditionally

### LINKED ISSUE #4747: Editing stream name / description puts double quotes around it.

gnprice:
This subsumes #4748, and fixes #4747.

I started adding the needed conditional to the code in #4748, and then faced the question of what to do with the case where `zulipFeatureLevel` was null. After some thought I realized that that case is possible, but just barely, and we can make a change to make it no longer possible. Then we can have a `getZulipFeatureLevel` selector that returns a non-null feature level, and encapsulates an assertion and the reasoning about why it holds.

This appears to be the first time we're actually conditioning on the feature level, and we have just a couple of conditionals on the Zulip version number.  Hopefully these selectors will simplify their use a bit, as we go to use them in more places.

Fixes: #4747

### DIFF
--- a/src/api/streams/updateStream.js
+++ b/src/api/streams/updateStream.js
@@ -2,6 +2,13 @@
 import type { ApiResponse, Auth } from '../transportTypes';
 import { apiPatch } from '../apiFetch';
 
+/**
+ * https://zulip.com/api/update-stream
+ *
+ * NB the caller must adapt for old-server compatibility; see comment.
+ */
+// TODO(#4659): Once we pass the feature level to API methods, this one
+//   should encapsulate a switch at feature level 64.  See its call sites.
 export default (
   auth: Auth,
   id: number,
--- a/src/streams/streamsActions.js
+++ b/src/streams/streamsActions.js
@@ -1,7 +1,7 @@
 /* @flow strict-local */
 import type { Stream, ThunkAction } from '../types';
 import * as api from '../api';
-import { getAuth } from '../selectors';
+import { getAuth, getZulipFeatureLevel } from '../selectors';
 
 export const createNewStream = (
   name: string,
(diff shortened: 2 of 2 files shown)


src/api/streams/updateStream.js
```diff
@@ -2,6 +2,13 @@
 import type { ApiResponse, Auth } from '../transportTypes';
 import { apiPatch } from '../apiFetch';
 
+/**
+ * https://zulip.com/api/update-stream
+ *
+ * NB the caller must adapt for old-server compatibility; see comment.
+ */
+// TODO(#4659): Once we pass the feature level to API methods, this one
+//   should encapsulate a switch at feature level 64.  See its call sites.
```
chrisbobbe, line 11: Could mention that we have a plan sketched out for that at https://github.com/zulip/zulip-mobile/issues/4659#issuecomment-914996061.

chrisbobbe (approved):
Thanks @gnprice, this LGTM! Please merge at will after seeing the two small comments below.

src/streams/streamsActions.js
```diff
@@ -17,20 +17,23 @@ export const updateExistingStream = (
...
-      getAuth(getState()),
-      id,
-      'description',
-      JSON.stringify(newValues.description),
-    );
+    await api.updateStream(auth, id, 'description', maybeEncode(newValues.description));
   }
   if (initialValues.invite_only !== newValues.isPrivate) {
-    await api.updateStream(getAuth(getState()), id, 'is_private', newValues.isPrivate);
+    await api.updateStream(auth, id, 'is_private', newValues.isPrivate);
```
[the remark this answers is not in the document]
gnprice: Sure, I'll reword.

src/api/streams/updateStream.js
gnprice: I think I'm content to let the issue number stand for that. If in the future when we go to implement that issue we overlook that feature, then when we look at this TODO comment and go looking for how the issue was meant to relate to feature levels I think we'll find those comments.

chrisbobbe: Oh, right, makes sense! I read right past the issue number in the comment, without noticing it. Then I went and dug up the link to the specific comment, just because I remembered writing it.

gnprice:
Thanks for the review! Merged, with a small change discussed above.

Issue #5060, opened 2021-10-20
Upgrade react-native-image-picker to ≥v3


gnprice:
We use the library https://npmjs.com/package/react-native-image-picker for letting the user take a photo, or select an existing image from their device, in order to attach it to a Zulip message. It has a new major version, v3.x.

We intend to upgrade to that version shortly, in particular because it promises changes that will work with Android "shared storage", which is becoming a requirement: https://github.com/zulip/zulip-mobile/issues/4283#issuecomment-825181815 @chrisbobbe has already started on this as part of #4283, and has some draft changes for it.

Filing this as its own issue mainly so we have something distinct to refer to for the upgrade itself, from other issues which relate to the library.

PR #5063, opened 2021-10-20
deps: Upgrade react-native-image-picker to 4.1.1, the latest

chrisbobbe:
Fixes: #5060

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -22,6 +22,12 @@ node_modules/react-native/Libraries/polyfills/.*
 .*/node_modules/flow-coverage-report/.*
 .*/node_modules/@snyk/.*
 
+; As of v4.1.1, this package only has TypeScript. That's fine as far as
+; running the code goes -- Metro transpiles it -- but this line causes Flow
+; to stop looking for types in that TypeScript, and then go and look in our
+; libdef.
+.*/node_modules/react-native-image-picker/.*
+
 [untyped]
 .*/node_modules/@react-native-community/cli/.*/.*
 
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,20 +8,20 @@
     <uses-permission
         android:name="android.permission.READ_PHONE_STATE"
         tools:node="remove" />
-    <uses-permission android:name="android.permission.CAMERA" />
+
+    <!-- CAMERA has an unexpected wrinkle, and we want to avoid having any
+         dependency requesting it. See
+         https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/react-native-image-picker.20upgrade/near/1267245,
+         https://github.com/react-native-image-picker/react-native-image-picker/commit/839d2afe1,
+         and https://issuetracker.google.com/issues/37063818 -->
+    <uses-permission
+        android:name="android.permission.CAMERA"
+        tools:node="remove" />
+
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature
         android:name="android.hardware.touchscreen"
-        android:required="false" />
-
-    <!-- The CAMERA permission causes these two camera-related features to be
-         implicitly treated as required, so we have to explicitly say they're not. -->
-    <uses-feature
-        android:name="android.hardware.camera"
-        android:required="false" />
-    <uses-feature
-        android:name="android.hardware.camera.autofocus"
         android:required="false" />
 
     <!-- We use certain modules that adds the permission requirement for
--- a/src/compose/ComposeMenu.js
+++ b/src/compose/ComposeMenu.js
@@ -69,8 +69,14 @@ export const chooseUploadImageFilename = (uri: string, fileName: string): string
    * automatically converted to JPEG format... but the `fileName` property in
    * the react-native-image-picker response still has the `.HEIC` extension.
    */
+  // TODO: Is it still true, after the react-native-image-picker upgrade,
+  // that `fileName` can end in .HEIC?
   if (/\.jpe?g$/i.test(uri)) {
-    return nameWithoutPrefix.replace(/\.heic$/i, '.jpeg');
+    const result = nameWithoutPrefix.replace(/\.heic$/i, '.jpeg');
+    if (result !== nameWithoutPrefix) {
+      logging.info('OK, so .HEIC to .jpeg replacement still seems like a good idea.');
+    }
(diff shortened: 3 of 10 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -8,22 +8,12 @@
     <uses-permission
         android:name="android.permission.READ_PHONE_STATE"
         tools:node="remove" />
-    <uses-permission android:name="android.permission.CAMERA" />
```
gnprice, line 11: Let's go farther and explicitly say we don't want a `<uses-permission>` element like this one. The `READ_PHONE_STATE` element above this provides an example.

That way, if some future dependency wants to introduce such an element for some reason, it won't actually show up in the manifest. That's useful because we have this new constraint that having this permission in the manifest would be bad, potentially causing an issue that's annoying for users and subtle to debug.

(And OTOH if that library does actually need the permission, hopefully we'll have a better shot at debugging that -- because it'll be a breakage in the thing we're actively working on and thinking about -- than we would at it subtly causing the breakage we're thinking about now.)

src/compose/ComposeMenu.js
```diff
@@ -63,72 +69,79 @@ export const chooseUploadImageFilename = (uri: string, fileName: ?string): strin
    * automatically converted to JPEG format... but the `fileName` property in
    * the react-native-image-picker response still has the `.HEIC` extension.
    */
+  // TODO: Is it still true, after the react-native-image-picker upgrade,
+  // that `fileName` can end in .HEIC?
   if (/\.jpe?g$/i.test(uri)) {
-    return name.replace(/\.heic$/i, '.jpeg');
+    const result = nameWithoutPrefix.replace(/\.heic$/i, '.jpeg');
+    if (result !== nameWithoutPrefix) {
+      logging.info('OK, so .HEIC to .jpeg replacement still seems like a good idea.');
```
gnprice, line 77: Looks good but I would not call this commit NFC. The new log event is a quite salient kind of behavior change for us as the developers (and we are after all the audience for the commit history :wink:), even if it's generally not noticeable for users.

.flowconfig
```diff
@@ -116,6 +116,16 @@ suppress_type=$FlowFixMeProps
...
 module.file_ext=.js
+
+# Rarely, a package only has .ts files and no .js files. For that case, we
+# tell Flow to pay attention to those .ts files, just in case Flow happens
+# to be able to parse them, i.e., in case the TypeScript syntax used in the
+# file also happens to be recognized as Flow syntax. When that's not the
+# case, we should write a Flow libdef (see docs/howto/libdefs.md). We should
+# probably write a Flow libdef anyway: just because Flow recognizes some
+# TypeScript doesn't mean it interprets it the way the writer intended.
+module.file_ext=.ts
```
gnprice, line 127: Do we need this bit? I see there a libdef in this commit too.

src/compose/ComposeMenu.js
```diff
@@ -63,72 +69,79 @@ export const chooseUploadImageFilename = (uri: string, fileName: ?string): strin
...
-      ...
-    }>,
-  ) => {
-    if (response.didCancel) {
+    if (response.didCancel === true) {
+      return;
+    }
+
+    const errorCode: string | void = response.errorCode;
+    if (errorCode != null) {
```
gnprice, line 94: Can delete the `string | void` annotation -- the handy libdef gives a more precise type.

src/compose/ComposeMenu.js
```diff
@@ -63,72 +69,79 @@ export const chooseUploadImageFilename = (uri: string, fileName: ?string): strin
...
+    const { uri, fileName } = firstAsset ?? {};
+
+    if (!firstAsset || uri == null || fileName == null) {
+      // TODO: If we need to keep this, wire up the user-facing string for
+      // translation.
+      showErrorAlert('Error', 'Something went wrong, and your message was not sent.');
+      logging.error('Unexpected response from image picker', {
+        firstAssetUndefined: firstAsset === undefined,
+        uriUndefined: uri === undefined,
+        fileNameUndefined: fileName === undefined,
```
gnprice, line 112: Probably these three diagnostic booleans should match the three `||` operands above.

gnprice (commented):
Thanks @chrisbobbe for taking this on! This looks good modulo small comments below, and a couple of errors that Flow is reporting.

There's also some TODO comments in the code that will be good to turn into followup issues (or find existing issues to update) after merge.

src/compose/ComposeMenu.js
```diff
@@ -3,9 +3,11 @@ import React, { PureComponent } from 'react';
 import type { ComponentType } from 'react';
 import { Platform, View } from 'react-native';
 import type { DocumentPickerResponse } from 'react-native-document-picker';
-// $FlowFixMe[untyped-import]
-import ImagePicker from 'react-native-image-picker';
+/* $FlowIgnore[cannot-resolve-module] Package only has TS files. Metro
+   transpiles those, and Flow shouldn't care because we wrote a libdef. */
+import { launchCamera, launchImageLibrary } from 'react-native-image-picker';
```
gnprice, line 8: Huh. Does that mean it's not seeing the libdef?

… I think it's not seeing the libdef. If I go munge one of the property names we pass to these functions:
```diff
     launchCamera(
       {
-        mediaType: 'photo',
+        mediaTy: 'photo',
```
I don't get an error.

Perhaps we need some other line in the .flowconfig to tell Flow to ignore the .ts file? And/or to look at the libdef instead. One model might be these lines:
```
[libs]
node_modules/react-native/interface.js
node_modules/react-native/flow/

; If the following files are not explicitly named, Flow ignores them in favor of
; their autogenerated versions in flow-typed/npm. (Contrast commit 6219e5a23,
; where the opposite behavior is exhibited!)
flow-typed/@sentry/react-native_v1.x.x.js
flow-typed/react-native-url-polyfill_vx.x.x.js
flow-typed/expo-screen-orientation_vx.x.x.js

; Explicitly name these so that Flow sees them instead of their
; counterparts in node_modules/react-native:
flow-typed/jest_v26.x.x.js
```

But if with some quick experimentation you don't find a way to get it working, then… we'll ship the version telling Flow to look at the .ts file, I guess, and troubleshoot from there.

gnprice (commented):
Thanks for the revision! Looks good except the comment below.

chrisbobbe:

gnprice:
Thanks! Looks good; merged. Glad to have this upgrade done :smile:

PR #5067, opened 2021-10-21
Input [nfc]: Bump up the contrast on placeholder

AlaaElden98:
This fixes the second item in #5038

### DIFF
--- a/src/styles/constants.js
+++ b/src/styles/constants.js
@@ -20,3 +20,4 @@ export const HIGHLIGHT_COLOR: string = Color(BRAND_COLOR)
 
 export const HALF_COLOR = 'hsla(0, 0%, 50%, 0.5)';
 export const QUARTER_COLOR = 'hsla(0, 0%, 50%, 0.25)';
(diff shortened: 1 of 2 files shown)


src/styles/constants.js
```diff
@@ -20,3 +20,4 @@ export const HIGHLIGHT_COLOR: string = Color(BRAND_COLOR)
 
 export const HALF_COLOR = 'hsla(0, 0%, 50%, 0.5)';
 export const QUARTER_COLOR = 'hsla(0, 0%, 50%, 0.25)';
+export const THREE_QUARTERS_COLOR = 'hsla(0, 0%, 50%, 0.75)';
```
AlaaElden98, line 23: I'm not sure if this naming is ok, this how I came up with it : 
`HALF_COLOR = 'hsla(0, 0%, 50%, 0.5)';`  alpha = 0.5 so it's `HALF`
`QUARTER_COLOR = 'hsla(0, 0%, 50%, 0.25)';` alpha = 0.25 so it's `QUARTER`
So, if alpha in  'hsla(0, 0%, 50%, 0.75)' = 0.75 , then the variable name should be `THREE_QUARTERS` .

PR #5069, opened 2021-10-21
stream IDs: StreamItem and TopicItem props and callbacks

gnprice:
This is an installment on #3918, converting from stream names to stream IDs; I'd taken a look at it recently following #5056.

In this PR, we make those conversions for the two components StreamItem and TopicItem: they now take the stream ID as a prop along with the stream name and other details, and pass the stream ID to their callbacks as well as the stream name and in TopicItem's case the topic.

This allows a couple of simplifications, to skip looking up the ID from the name in a couple of places where this code was already calling something that demands an ID. It also makes stream IDs available in various places we currently use names, including several Narrow-constructor call sites.


chrisbobbe:
Thanks, LGTM! Merged.

Issue #4284, opened 2020-10-21
Target Android 11 (set targetSdkVersion to 30), by 2021-11 deadline


gnprice:
This is the successor to #3665 (and #3563 and #3075). We should update our `targetSdkVersion` to 30, meaning Android 11.

Google Play [has set a deadline](https://android-developers.googleblog.com/2020/11/new-android-app-bundle-and-target-api.html) of [2021-11-01](https://support.google.com/googleplay/android-developer/answer/9859152?visit_id=637547176866559252-1000181464&rd=1#targetsdk) for this update, consistent with the pattern of the last few years.
(~~Based on [the last few years' experience](https://support.google.com/googleplay/android-developer/answer/113469#targetsdk), Google Play will probably set a deadline for this of late 2021. It's been the first business day of November in each of the last few years.~~)

The important steps for this upgrade are:
1. Read about the [potentially-breaking changes](https://developer.android.com/about/versions/11/behavior-changes-11), and identify those that might affect us.
2. Make a WIP change to `targetSdkVersion`.
3. Test the WIP change thoroughly, especially in the areas highlighted in step 1. Fix things as needed, and repeat.

One known blocker:
* #4283. The library we use for taking a photo to send via Zulip, or picking an existing image on the device to send via Zulip, hasn't yet adapted to "scoped storage", which arrived with Android 10. We've taken an opt-out for that, but the opt-out stops being available with Android 11.

Apart from that issue, we could do this any time. It'd be nice to do it well before the deadline -- or at least to learn about any other, unknown blockers.

Issue #4283, opened 2020-10-21
Fix image capture + image upload to work with Android "scoped storage".


gnprice:
Android 10 introduces significant changes in how apps access the device's photos, other media of several types, and files in general. Gone, in that future, are the days where your photos and documents live in a single shared storage area which lots and lots of apps have unrestricted access to, like on a desktop OS. The new model is called "scoped storage".

That future is opt-out for apps targeting Android 10, and we've had to take that opt-out -- see https://github.com/zulip/zulip-mobile/issues/3665#issuecomment-713238913 and #4282. But it will be mandatory for apps targeting Android 11, and targeting Android 11 will presumably be mandatory for updates on Google Play in about a year. So our opt-out is only a temporary extension of the deadline; we'll need to adapt to scoped storage

The known issues we currently run into with scoped storage are all within the react-native-image-picker library, which we use both for taking a photo to send via Zulip and picking an existing image on the device to send via Zulip. Both of those features encounter trouble:
* On trying to take a photo (the "camera" icon), an error modal appears before the camera widget even shows up. Detailed report: https://github.com/zulip/zulip-mobile/issues/3665#issuecomment-713199695
* On trying to pick an existing image (the "photo" icon), the symptoms are different; I saw them earlier today but have forgotten exactly how it went wrong. ISTR the picker appeared successfully, but then an error occurred upon actually choosing an image.

The basic problem is that r-n-image-picker tries to handle the photos as if it had unrestricted access to them as files, and with scoped storage it no longer does. Several upstream issues seem related (I think some of them are duplicates): at least
* react-native-image-picker/react-native-image-picker#1269,
* react-native-image-picker/react-native-image-picker#1350,
* react-native-image-picker/react-native-image-picker#1393,
* react-native-image-picker/react-native-image-picker#1443;
* and there's an open PR react-native-image-picker/react-native-image-picker#1439.

The PR is quite simple, and the main reason it isn't already merged seems to be (https://github.com/react-native-image-picker/react-native-image-picker/pull/1439#issuecomment-702876251) that it gives up entirely on putting the captured image, when you use the camera, into the central "Pictures" folder shared with other apps. That is a useful feature. As a Zulip user, I prefer to keep that feature too.

[...]

PR #5070, opened 2021-10-22
android: Update target SDK version to 30, i.e. Android 11

chrisbobbe:
Along with moving us over to "scoped storage", which will be newly enforced on Android 11.

Fixes: #4283
Fixes: #4284


gnprice:
Thanks @chrisbobbe for all your work leading up to this! Looks great; merged.

Issue #4873, opened 2021-07-06
Full-screen compose box


alya:
At present, the compose box is quite small on mobile. Especially with Zulip's recommended style of slightly longer messages, it would be very helpful to have a full-size compose box option for composing long messages.

The toggle between the small vs. full-size compose box can be double arrows pointing in or out in the corner, a la Slack.

SilentCruzer:
Can I work on this issue?

alya:
@gnprice can commend on how approachable this issue is.

SilentCruzer:
(chat continued on [CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Expand.20compose.20box/near/1261834))

SilentCruzer:
@alya , should I add the expoand icon inside the text-box like this:

![expand](https://user-images.githubusercontent.com/73246484/136096973-83e73c4a-e71b-47c6-88e1-0000e2504f08.png)

The only problem with this is that we need to add padding to the left of the icon, so text-overflow doesn't happen.

alya:
@SilentCruzer Yeah, I think that's the right place for it. I think it could be closer to the right edge of the text box in your screenshot.

alya:
@sumj25 please discuss with @SilentCruzer whether they are still working on this issue.

SilentCruzer:
@alya, I have added the expand feature, this is how it looks now:

https://user-images.githubusercontent.com/73246484/137984611-f41bbdba-9a0a-492f-8359-103b48197974.mp4

Can you suggest changes if any?

alya:
We should just use different gray tones in light vs dark mode, as we do with many UI elements.

PR #5085, opened 2021-10-27
compose: Add full-screen compose box

SilentCruzer:
Closes: #4873 

Changes:
- Main changes are in `src/compose/ComposeBox.js`, added an expand button in the right corner of the text input box.
- When pressed on the button a full screen compose box will pop up. Added a shrink button at the top right so the user can go back to the normal input.
- `src/compose/ComposeMenu.js`, in this added a condition to check where the menu is called from. The reason for adding this is because I reused the compose menu from the normal compose box and wanted to hide the collapse button if the compose box is in expanded mode.

Changes demo:

https://user-images.githubusercontent.com/73246484/139150667-b5d0f1ff-c9fc-4e50-8f14-b69c6d62da7a.mp4

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -89,3 +89,5 @@ export const IconAttach: SpecificIconType = makeIcon(Feather, 'paperclip');
 export const IconAttachment: SpecificIconType = makeIcon(IoniconsIcon, 'document-attach-outline');
 export const IconGroup: SpecificIconType = makeIcon(FontAwesome, 'group');
 export const IconPublic: SpecificIconType = makeIcon(FontAwesome, 'globe');
+export const IconExpand: SpecificIconType = makeIcon(FontAwesome, 'expand');
+export const IconCompress: SpecificIconType = makeIcon(FontAwesome, 'compress');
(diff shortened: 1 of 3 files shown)


SilentCruzer:
Modified the PR, replaced the KeyBoard avoiding view with a modal this time. Also made the changes mentioned in the [CZO discussions](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4873.20full-screen.20compose/near/1273627).

Video of current changes: 

https://user-images.githubusercontent.com/73246484/140660416-386b6edf-7303-410b-a270-ce8008044ded.mp4

alya:
Thanks for the updated video -- sorry I missed it earlier!

In this new version, we aren't showing the stream name, which I think is confusing. It also seems a bit hard to tell that the topic name is a topic name. 

I'm not sure what the best solution is. Perhaps we should (1) keep the top header bar, and (2) keep the lighter background around the topic box to separate it out better?

chrisbobbe:
> In this new version, we aren't showing the stream name, which I think is confusing. It also seems a bit hard to tell that the topic name is a topic name.

Makes sense; I agree. I've posted some more thoughts on CZO [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4873.20full-screen.20compose/near/1283784).

SilentCruzer:
@alya @chrisbobbe , I have made the suggested changes. The screenshots of the changes can be found in the CZO [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M4873.20full-screen.20compose/near/1284873).

PR #5090, opened 2021-10-30
notifs: Warn when server not registered for push notifications

chrisbobbe:
I've just updated this to exclude the part where we respond to realm-update events. It's "good enough" for now to just use what we get in the `/register` response, I think. And I've got #5123 in flight toward the events part.

Fixes: #1507


src/api/eventTypes.js
```diff
@@ -171,3 +172,118 @@ export type RestartEvent = $ReadOnly<{|
...
+  video_chat_provider?: number,
+  waiting_period_threshold?: number,
+  digest_weekday?: number,
+  private_message_policy?: number,
+  user_group_edit_policy?: number,
+  default_code_block_language?: string,
+  message_content_delete_limit_seconds?: number,
+
+  // Fails `value?: string | boolean | number` check on RealmUpdateEvent.
+  // authentication_methods?: {| +[authentication_method_name: string]: boolean |},
```
gnprice, line 208: It looks like all (or most?) of the properties here are meant to correspond to properties in `InitialDataRealm`.

Awkwardly we can't actually use that type, or express in the type system its relationship to this one, because the names differ: there's that `realm_` prefix on the initial-data side.

But let's use the same types for both, and where they need to be updated let's do so in tandem. We can also keep them listed in the same order, to facilitate visually comparing them (and spotting where one side has a property that the other doesn't.)

We can also make a type-test, to confirm they stay in sync for each given property. (This will be most useful for the properties like this one where the type might get updated over time.) The type-test could have a long list of lines like:
```js
  check<'realm_authentication_methods', 'authentication_methods'>();
  check<'realm_avatar_changes_disabled', 'avatar_changes_disabled'>();
  // … etc. etc. etc., one per property
```
where `check` can be written somewhat like `typesEquivalent`. I can make a go at writing that later.

Then we'll still need to rely on comments to remind us to keep them in sync when adding new properties, but at least the test will help when updating one of these complex properties' types.

gnprice (commented):
I see this is WIP and didn't read through closely. But here's a couple of thoughts that may be helpful.

chrisbobbe:
Thanks for the review! Revision pushed, and I've just updated the description.

src/types.js
```diff
@@ -126,6 +126,13 @@ export type Account = {|
    * See the `pushToken` property in `SessionState`, and docs linked there.
    */
   ackedPushToken: string | null,
+
+  /**
+   * When the user last dismissed the server-not-set-up-for-notifs notice.
+   *
+   * `null` if never or if the server reports it's set up for notifications.
```
gnprice, line 133: Can we just make this "`null` if never"? That seems simpler to reason about. Then the interaction with whether the server reports it's set up for notifications is for the UI to handle, where it's deciding whether to show a banner.

(And I think the code there in this version does handle that just as one would want.)

src/common/ServerPushSetupBanner.js
```diff
@@ -0,0 +1,88 @@
...
+  const isAdmin = useSelector(getIsAdmin);
+  const settings = useGlobalSelector(getGlobalSettings);
+
+  let visible = false;
+  let text = '';
+  if (pushNotificationsEnabled) {
+    // don't show
+  } else if (
+    lastDismissedServerPushSetupNotice !== null
+    && isAfter(
```
gnprice, line 38: This can be just `>=` -- comparison between Date objects behaves just as you'd hope.

(And in fact if you look at the implementation of `date-fns/isAfter`, that's just what it does, after it's taken its possibly-not-actual-Date arguments and converted them into `Date`s.)

src/common/ServerPushSetupBanner.js
```diff
@@ -0,0 +1,88 @@
...
+      lastDismissedServerPushSetupNotice,
+      subWeeks(
+        // TODO: Could rerender this component on an interval, to give an
+        //   upper bound on how old this "current date" can be.
+        new Date(),
+        2,
+      ),
+    )
```
gnprice, line 46: nit: this whole comparison fits nicely on a line, if the comment is pulled to the outside of it:
```js
    // TODO: Could rerender this component on an interval, to give an
    //   upper bound on how old this "current date" can be.
    && isAfter(lastDismissedServerPushSetupNotice, subWeeks(new Date(), 2))
```

Needs a bit of rewording that way, because its physical location no longer points specifically to the `new Date()` it's about but rather to the whole comparison in general. But 7 lines of code turn into 1 line, and without a comment in the middle, making it much easier to take it all in at once. That's a good trade.

src/common/ServerPushSetupBanner.js
```diff
@@ -0,0 +1,88 @@
...
+        {
+          id: 'dismiss',
+          label: isAdmin ? 'Fix later' : 'Remind me later',
+          onPress: () => {
+            dispatch(dismissServerPushSetupNotice());
+          },
+        },
+        {
+          id: 'resolve',
+          label: isAdmin ? 'Fix now' : 'Learn more',
```
gnprice, line 77: "Fix now" doesn't feel right to me here -- as if it's a button that *will* fix it, rather than give you instructions on how you can fix it. (And it may not even be that, if you're an admin within Zulip but not a server admin with access to the underlying machine -- then it's instructions you can send a colleague in asking them to fix it.)

I think the same "Learn more" / "Remind me later" text as for non-admins might work best.

gnprice (commented):
Thanks for taking care of this! Small comments below.

(And it looks like the changes I was commenting on above got separated into #5123, which makes sense.)

src/types.js
chrisbobbe: The other part is aimed at potential issues that could happen when the `realm_push_notifications_enabled` value changes back and forth in the two-week period after a user dismisses the notice.

Here's an example sequence of events, I guess:

1. Tell the admin to fix it
2. Dismiss the banner
3. Admin fixes it
4. A week goes by
5. Admin reverts the fix
6. (Do you want to see the banner right away, or wait another week?)

chrisbobbe: It's very possible this is a small enough edge case that we don't need to worry about it.

chrisbobbe:
Thanks for the review! Revision pushed, and I've responded to a point at https://github.com/zulip/zulip-mobile/pull/5090#discussion_r749832647.

src/types.js
gnprice: Hmm, I see.

Yeah, I think ideally in that scenario we'll promptly start showing the banner again -- you may have gotten notifications, so you think it's fixed, and want to know it's broken again. But also it is probably pretty uncommon that it gets fixed and promptly broken again.

The semantics that points to is still a bit different from this jsdoc, though: it'd be that this is when the user last dismissed the notice, since the last time it became applicable, and `null` if they haven't since then.

In particular, in your example at step 5, this value will be `null` even though the user had dismissed the notice (more recently than never), and even though the server doesn't report it's set up for notifications.

gnprice:
Thanks! Replied there, and otherwise looks good.

gnprice:
Oh, also the last commit message should be updated for #5123 being merged. It's now that there's simply a followup task to apply #5123 to this.

chrisbobbe:

chrisbobbe:
I've just removed the TODO for handling the value being updated in an event. It turns out that it isn't updated in an event after all; see https://github.com/zulip/zulip-mobile/pull/5131#issuecomment-971949204.

chrisbobbe:
(Oops, I accidentally mis-reported this one as merged on our call yesterday. Still, I think it's pretty close.)

chrisbobbe:
Revision pushed, with a change to show the banner persistently in the notification-settings screen, and also an independent small cleanup (making some object types read-only).

src/common/ServerPushSetupBanner.js
```diff
@@ -0,0 +1,93 @@
...
+ * applicable notice. If the server does get setup for push notifications,
+ * then gets un-setup, a new notice will apply.)
+ */
+export default function PushNotifsSetupBanner(props: Props): Node {
+  const dispatch = useDispatch();
+
+  // If we can detect that this banner is on the notification-settings
+  // screen, it shouldn't be dismissable.
+  const route = useRoute();
+  const isDismissable = route.name !== 'notifications';
```
gnprice, line 35: Hmm, I don't like having this reach upward like this to have its behavior depend on where it finds itself -- it feels like a sort of action-at-a-distance that's going to be a surprise for someone making changes in the future.

It's good that the jsdoc clearly mentions this, and that'd be an essential thing to do to mitigate the confusingness if we did go down this path.

But how about instead making this flag just be a boolean prop? Then `NotificationsScreen` will set it, and `HomeScreen` will leave it out, and it'll be clear from looking at the two call sites why they behave differently, just like normal.

gnprice (commented):
Thanks @chrisbobbe! The first three commits look good -- I'll merge.

A comment below on the implementation strategy for the three new commits at the end. Let's handle that functionality in a new separate PR.

src/common/ServerPushSetupBanner.js
chrisbobbe: Sure, that's reasonable! 🙂

I think I was thinking of a possible future where `<ServerPushSetupBanner />`, `<ServerCompatBanner />`, etc., are called by something like a `BannerManager` component, instead of called directly by our screen components. That `BannerManager` might somehow take care of the [don't-show-two-banners-at-once requirement](https://material.io/components/banners), like by popping up a second applicable banner just after you dismiss the first.

In that case, it might be cumbersome to plumb a prop down to `ServerPushSetupBanner` from the screen component.

But we aren't using something like a `BannerManager` component, and if we were, we might still be concerned about action-at-a-distance.

gnprice: Yeah. We can cross that bridge when we come to it -- we'll see if there's a reasonable way to pass the information through directly, and if not then we can do this or something else indirect.

PR #5091, opened 2021-10-30
android sharing [nfc]: Rename and javadoc the shim activity

gnprice:
This is an activity, not an intent, so its name should end with
"Activity" and not with "Intent".

We can also name it a bit more for what its purpose is at a
product level; the fact that it gets intents whose action is
SEND (and SEND_MULTIPLE) is expressed more clearly in the
relevant code in the manifest anyway.  Also add a bit of javadoc
to explain what's going on and why it doesn't look like it's
doing much.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -91,7 +91,7 @@
         </activity>
 
         <activity
-            android:name=".HandleSendIntent"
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks, LGTM! Please merge at will.

gnprice:
Done. Thanks for the review!

PR #5092, opened 2021-11-01
lint: Drop unhelpful rule flowtype/space-after-type-colon

gnprice:
From the comment, it doesn't seem I thought this rule was a good idea
even in 2018, when I moved it from our top-level ESLint config to this
file that also applies in our prettier-eslint runs.

But mostly it just hasn't come up much.  It did again in a recent PR,
though:
  https://github.com/zulip/zulip-mobile/pull/5084#discussion_r740514987

And it turns out it's quite easy to remove!  Just one place changes.
(Even the one place we've explicitly disabled the rule, it seems to
have no effect; must have applied to an earlier version or draft.)
So do so.

### DIFF
--- a/src/__flow-tests__/generics-tests.js
+++ b/src/__flow-tests__/generics-tests.js
@@ -2,8 +2,6 @@
 import { type BoundedDiff, typesEquivalent } from '../generics';
 import type { IsSupertype } from '../types';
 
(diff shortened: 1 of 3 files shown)


chrisbobbe:
LGTM, thanks! Merged.

PR #5093, opened 2021-11-02
lint: Redirect react-redux imports through our type-wrapper module

gnprice:
This lets us catch bugs where, for example, a selector passed to
`useSelector` doesn't return the type that the caller wants it to.

Now that we have a distinction between `useSelector` and
`useGlobalSelector`, it also lets us catch bugs where `useSelector` is
mentioned but `useGlobalSelector` is what's actually wanted.  For the
moment these are both purely type-wrappers of the upstream
`useSelector`, so such a bug would have no effect at runtime... but in
the future a per-account state will be a different actual object from
a global state, so one or both of them will not pass its selector the
same state object as the upstream `useSelector` would, and then such a
bug would likely cause a crash.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -227,7 +227,10 @@ rules:
     - devDependencies: ['**/__tests__/**/*.js', tools/**]
   no-restricted-imports:
     - error
-    - patterns: ['**/__tests__/**']
+    - patterns:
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Thanks, LGTM! Merged.

PR #5094, opened 2021-11-02
deps: Upgrade jest-expo to 43, which officially targets RN v0.64

chrisbobbe:
And, as expected (see v27.170 in the changelog), we stop seeing a
few warnings in our regular dev workflow.

- When running Jest, we'd get the following warning, which was fixed
  in expo/expo@24bce422c:

  react-native/jest-preset contained different
    transformIgnorePatterns than expected

- When running Yarn, we'd get the following warning, fixed by
  removing a `resolutions` line, which we can now do (see below):

  warning Resolution field "react-test-renderer@17.0.1" is
    incompatible with requested version
    "react-test-renderer@~16.11.0"

And remove some workarounds we'd had in place:

Remove `fbjs-scripts`, which we added in 42bb7128f. It looks like
the relevant issue was fixed in expo/expo@d46578571.

Remove a `resolutions` line, which we added in 4a5780d65. They've
released the relevant change to their peer dependencies
(expo/expo@d49e7070a).

And run `yarn yarn-deduplicate && yarn`, as prompted by
`tools/test deps`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,6 @@
     "eslint-plugin-prettier": "^3.2.0",
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "fbjs-scripts": "1.2.0",
     "flow-bin": "^0.137.0",
     "flow-coverage-report": "^0.8.0",
     "flow-typed": "^3.3.1",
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks for the update! Merged.

chrisbobbe:
Thanks!

PR #5095, opened 2021-11-02
msglist-diffing: Add new div#msglist-elements

chrisbobbe:
I tested the message-list on iOS, and I didn't notice any change in behavior. It seems like the kind of thing where regressions might be sneaky, though.

Related: #3851

### DIFF
--- a/src/webview/html/htmlBody.js
+++ b/src/webview/html/htmlBody.js
@@ -32,7 +32,9 @@ export default (content: string, showMessagePlaceholders: boolean): string => te
 <div id="js-error-plain-dummy">Oh no! Something went wrong. Try again?</div>
 <div id="spinner-older" class="hidden loading-spinner"></div>
 
-$!${content}
+<div id="msglist-elements">
+  $!${content}
+</div>
 
 <div id="message-loading" class="${showMessagePlaceholders ? '' : 'hidden'}">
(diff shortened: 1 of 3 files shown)


chrisbobbe:
(And a few cleanups of backward-compatibility code we don't need anymore.)

gnprice:
Thanks! This looks good; I studied the code and looked for other ways we were depending on the behavior that changes here, and didn't see any. Merging.

This does change the set of elements that count as a message-list element:
```
 * A "message-list element" is a message or one of their siblings that get
 * laid out among them: e.g. a recipient bar or date separator, but not an
 * absolutely-positioned overlay. All message-list elements are direct
 * children of a div#msglist-elements.
```
in that several siblings of the new `#msglist-elements` are laid out with it (and its children), and previously counted as message-list elements: the spinners and the typing indicator.

But AFAICT the only way we currently use the concept of a message-list element (previously called a "message-peer") is to find an actual message, using the fact that they're siblings -- we don't care about those elements *as* message-list elements at all. So that shouldn't change any behavior. And then basically it serves to narrow the concept to include only the things that we want included for the new use we're going to put to it soon, which is the list of items to diff.

chrisbobbe:
Thanks for the review!

Issue #5097, opened 2021-11-03
Share to Android: After sharing once, app opened to "share" screen on startup the next few times


chrisbobbe:
@timabbott reports that, after sharing to Zulip once, the next few times he opened the app, it had him on the "share" screen.

I just looked at the code, and I don't have a specific diagnosis yet.

I don't think it's a problem with how we're organizing our React Navigation screens; I can say more about this.

I think we're immediately navigating to the sharing screen, from some other initial screen, when we shouldn't be.

I don't know if the navigation is happening because of an "initial share" (`await Sharing.readInitialSharedContent()` being non-null) or because of a 'shareReceived' event.

I suspect the app might be receiving the share intent multiple times, and we should only be handling it the first time. Could it be some other flavor of 70e9ad9d4? The relevant code there was refactored in 1a20529e0; could there have been an issue there? Neither of those commits were released very recently.

PR #5098, opened 2021-11-03
Fix some bugs Tim reported in share-to-Android experience.

chrisbobbe:
This doesn't include a fix for one of them; I've filed #5097 for that.

Tim said:

> For sending that message, I used the Android sharing feature for the first time. I like the UI it took me to but ran into a few input issues:
> 
> - The default keyboard setting for the topic only was sentence case (which is usually not how we spell topics)
> - Some bug prevented adding a space at the end of what I was typing specifically on the topic and maybe stream input (the space would appear and then immediately disappear, I think) ... so I was only able to create the topic "bog plants" by typing "bogplants" and then moving my cursor.
> - The typeahead for the stream seemed to only do prefix match, so that typing just "sf" didn't find any matches.


src/sharing/ShareToStream.js
```diff
@@ -77,33 +76,27 @@ class ShareToStreamInner extends React.Component<Props, State> {
...
-    const { mandatoryTopics } = this.props;
-    const { stream, topic } = this.state;
     const { sharedData } = this.props.route.params;
 
-    if (sharedData.type !== 'text') {
-      return stream !== '' && (topic !== '' || !mandatoryTopics);
-    }
-
-    return stream !== '' && (topic !== '' || !mandatoryTopics) && message !== '';
+    return sharedData.type !== 'text' || message !== '';
```
gnprice, line 99: Hmm, this seems like a regression -- it means that when you actually have an empty topic, we'll show the send button as enabled even when that means you can't send, right?

And similarly for streams.

gnprice (commented):
Thanks! Comment below.

    share to android: Enforce mandatory-topics on submit, not during compose

Let's leave `android` out of the prefix here; this code is currently only reachable on Android, but it's code that will be in common with sharing to Zulip on iOS, once we do the iOS-specific work to make it reachable there too.

src/sharing/ShareToStream.js
chrisbobbe: > when you actually have an empty topic, we'll show the send button as enabled even when that means you can't send, right?

That's right. But I think this may actually be an improvement? It's good to tell the user specifically why they can't send a message, right, rather than having them guess?

I notice that the web app doesn't disable the send button, and waits for you to press it before giving you an alert. This is what I see just after pressing "Send":

image

One thing that I prefer in the web app's UX, for accessibility, is that the alert persists right alongside the relevant input, even while you go and change the value of the input. (Until you dismiss the alert anyway.) When it's in a modal, like I have here in this revision, you have to keep the validation requirement in your working memory while you try to go fix the error.

But I guess I still think it's an improvement over the status quo, where there isn't a validation message at all.

gnprice: Hmm interesting. It's true, it's good to be explicit about what's missing.

Still it seems like not good UX for the "send" button to be right there brightly beckoning to you when it's not actually possible to send. I don't think that's a good UX choice in the web app either.

For a quick comparison, I just opened up Gmail on Android and went to compose a new email. It actually does both:
* When I haven't entered anything (even in the To: field), the "send" icon is grayed out.
* But if I tap it anyway, I get a little modal error dialog: "Add at least one recipient.", with an "OK" button.
* Then once I enter a valid email address, the icon gets its normal blue color.

chrisbobbe:

gnprice:

chrisbobbe:
Thanks for the review!

PR #5100, opened 2021-11-04
api: Start relying more on non-ancient servers

gnprice:
The change here that caused me to look into this today was this one, after https://github.com/zulip/zulip-mobile/pull/5052#issuecomment-960296275 :

024853c2f caught-up: Drop legacy inference for servers pre-1.8

That turns out to delete for us a surprising amount of code! Especially in tests -- that was some subtle logic.

But then I also went and did a sweep to help lay the groundwork for other such changes, to make more use than we do of the assumption that the server is not ancient:

---

a0d86af95 api [nfc]: Use the greppable `TODO(server-N.M)` form consistently.

We've recently started to actually drop old conditionals that added
complexity and were needed for supporting old versions of the server.

To do more of that, it's good to have a systematic format we use when
marking that something depends on the server version, so that as we
desupport an old version we can come sweep through and find all the
new opportunities to simplify.

We've generally tried to make these references reasonably clear, but
hadn't had a concise form to use consistently.  A year ago in
e47aa004b, I introduced a form of comments like this:
    // TODO(server-2.0): switch to numeric user IDs, not emails.
and I think that works well.  Sweep through now, using a variety of
less-precise search patterns to try to find existing comments of this
kind, and convert them all to that standard form.

Many of these refer to server versions we already don't support --
with ServerCompatBanner we tell users their server is unsupported if
it's older than 2.1.  So a lot of these are opportunities we can
already take.  But we'll leave those for separate commits.

---

We don't need to rush to sweep through all of those cleanup opportunities, particularly as some are more work than others, some (like dropping that legacy caught-up inference, using the "new" `found_oldest` / `found_newest` properties) have more of a benefit to us than others, and some are more likely to break things for users than others. E.g., that caught-up inference is pretty high on the scale of likely breakage to users… to the extent users are on that version, which is very little.

But I may do a sweep through these soon, to do the easier and/or higher-value items and triage the rest.

### DIFF
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -713,8 +713,8 @@ export const action = Object.freeze({
     anchor: 0,
     numBefore: 50,
     numAfter: 50,
-    foundNewest: undefined,
-    foundOldest: undefined,
+    foundNewest: false,
+    foundOldest: false,
     ownUserId: selfUser.user_id,
   }): MessageFetchCompleteAction),
   // If a given action is only relevant to a single test file, no need to
--- a/src/__tests__/lib/exampleData.js
+++ b/src/__tests__/lib/exampleData.js
@@ -328,7 +328,6 @@ const messagePropertiesFromSender = (user: User) => {
   const { user_id: sender_id, email: sender_email, full_name: sender_full_name } = user;
 
   return deepFreeze({
-    sender_domain: '',
     avatar_url: user.avatar_url,
     client: 'ExampleClient',
     gravatar_hash: 'd3adb33f',
--- a/src/api/initialDataTypes.js
+++ b/src/api/initialDataTypes.js
@@ -30,10 +30,11 @@ export type InitialDataBase = $ReadOnly<{|
   // `fetch_event_types` and remove this comment.
 
   /**
-   * Added in server version 2.2, feature level 1.
+   * Added in server version 3.0, feature level 1.
    * Same meaning as in the server_settings response:
    * https://zulip.com/api/get-server-settings. See also the comment above.
    */
+  // TODO(server-3.0): Mark as required.
   zulip_feature_level?: number,
 
   /**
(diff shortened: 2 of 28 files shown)


timabbott:
> // TODO(server-2.0):

We use `TODO/compatibility` as a name for these; idunno if that's better, but I think the `/` separator might feel more convenient than parenthesis for doing greps.

gnprice:
> We use `TODO/compatibility` as a name for these; idunno if that's better, but I think the `/` separator might feel more convenient than parenthesis for doing greps.

Good to know, thanks. For greps indeed I tend to gloss over the paren by writing a dot:
```
$ rg TODO.server-1.8
```
and that works well. If there were something it sweeps up that isn't this syntax, then it's probably a malformed version of it that I wanted to find anyway.

I use the parens syntax because it's what I see some other projects doing; I think it was popularized as a Googleism. Compared to the slash idea, it's a bit more annoying to type, but OTOH it lends itself well to contents that themselves contain a slash (like the URL of an issue thread, when it's not in our own tracker.)

I do think having a specific version reference is a helpful element for the compatibility flavor of these TODO comments. That's what makes it possible to focus on the really old compatibility items that are already actionable, while ignoring the ones that are recent and addressed to the future.

src/__tests__/lib/exampleData.js
```diff
@@ -713,8 +713,8 @@ export const action = Object.freeze({
     anchor: 0,
     numBefore: 50,
     numAfter: 50,
-    foundNewest: undefined,
-    foundOldest: undefined,
+    foundNewest: false,
```
chrisbobbe, line 716:
```
[…] We may at some point want to have the app
outright refuse to operate on sufficiently ancient servers, where
it's just not going to be a good experience. […]
```

I've filed #5102 for this task.

chrisbobbe (approved):
Looks great, thanks for all these cleanups! Please merge at will.

src/api/initialDataTypes.js
```diff
@@ -30,10 +30,11 @@ export type InitialDataBase = $ReadOnly<{|
   // `fetch_event_types` and remove this comment.
 
   /**
-   * Added in server version 2.2, feature level 1.
+   * Added in server version 3.0, feature level 1.
```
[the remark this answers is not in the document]
gnprice: Yeah, to be clear, this wasn't wrong, just outdated -- 3.0 is the name we used for the next major release after 2.1, what was originally planned as 2.2. (That's the point at which we shortened the version numbers, following the industry-wide trend of recent years.)

src/__tests__/lib/exampleData.js
gnprice: Sounds good, thanks!

PR #5103, opened 2021-11-04
Rename some "initial fetch" plain Redux actions.

chrisbobbe:
In particular:

```
  INITIAL_FETCH_START                 -> REGISTER_START
  INITIAL_FETCH_ABORT                 -> REGISTER_ABORT
  INITIAL_FETCH_COMPLETE / REALM_INIT -> REGISTER_COMPLETE
```

This is taken from a draft branch I put together a few weeks ago, for a non-hacky alternative to the bugfix in 6d2647f27.


src/__tests__/lib/exampleData.js
```diff
@@ -595,7 +595,7 @@ export const action = Object.freeze({
     apiKey: selfAccount.apiKey,
   }): LoginSuccessAction),
   realm_init: (deepFreeze({
-    type: REALM_INIT,
+    type: REGISTER_COMPLETE,
```
gnprice, line 598: Should also update the name of the example action itself.

(I found this with the following search:
```
$ rg -i 'realm.?init'
```
looking for any variation of the name with or without caps or underscores.)

src/message/fetchActions.js
```diff
@@ -465,24 +461,23 @@ export const doInitialFetch = (): ThunkAction<Promise<void>> => async (dispatch,
...
     } else {
-      dispatch(initialFetchAbort('unexpected'));
+      dispatch(registerAbort('unexpected'));
       logging.warn(e, {
         message: 'Unexpected error during /register.',
       });
     }
     return;
   }
+  dispatch(registerComplete(initData));
```
gnprice, line 479: A good small other commit to add would be to move this action creator into this file, next to its "start" and "abort" counterparts. It's currently in `realmActions.js` (the only thing there). Which sort of makes sense if you accept the old "realm init" name, but doesn't make sense in any other way.

gnprice (commented):
Good idea, thanks for this cleanup! Small comments below.

There's also a couple of other matches to that search, in comments and test names.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Merging.

I fixed up where a couple of references got changed to "REGISTER_QUEUE" where "REGISTER_COMPLETE" was intended:
```diff
-        //   check for that; it looks like it only detects a new REGISTER_QUEUE.
+        //   check for that; it looks like it only detects a new REGISTER_COMPLETE.

-  describe('REGISTER_QUEUE', () => {
-    test('updates as appropriate on a boring but representative REGISTER_QUEUE', () => {
+  describe('REGISTER_COMPLETE', () => {
+    test('updates as appropriate on a boring but representative REGISTER_COMPLETE', () => {
```

chrisbobbe:
> a couple of references got changed to "REGISTER_QUEUE" where "REGISTER_COMPLETE" was intended

🤦 Mm, thanks for that!

PR #5073, opened 2021-10-24
Fix app restart on split screen (Android)

Somena1:
This commit fixes app restart on split screen and screen resize 
issue #5068

PR #5104, opened 2021-11-04
Android: Fix AppComponent reconstructed on split screen

### LINKED ISSUE #5068: Switching between regular and split view restarts the app
When opening the android app in split view and making it fullscreen or the other way around the app restarts, losing which thread I was reading. This is very annoying when I want to switch away from split view to be easier to read a thread as I will have to navigate back to it. Other apps simply resize their view when switching without losing ui state.

Somena1:
This commit fixes #5068
Switching between screen size or split screen causes Android AppComponent to be reconstructed and I'm initiate a restart.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -70,7 +70,7 @@
         android:theme="@style/AppTheme">
         <activity
             android:name=".MainActivity"
-            android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
+            android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize|uiMode"
             android:label="@string/app_name"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustResize">


chrisbobbe:
Thanks, @Somena1! I see two commits:

c4a976060cb9f023727875783672368b7f752bf8 (Android) Fix AppComponent reconstructed on split screen #5068
20648f1310713b52c05945785d88f28bfb981947 Android config changes set to alphabetical

It looks like the second commit is fixing up a small problem that was introduced in the first commit. Please squash the two commits together, so there's just one commit that doesn't (knowingly) introduce any problems.

For that single commit, please make sure it follows our commit-message style.

- The summary line shouldn't start with "(Android)". We have guidelines on how to write the prefix in the summary line: https://zulip.readthedocs.io/en/latest/contributing/version-control.html#good-summaries
- The commit-message body should end with `Fixes: #5068`: https://zulip.readthedocs.io/en/latest/contributing/version-control.html#message-body

Thanks!

Somena1:
Hello @gnprice, I've resolved the squash and commit message. 
Thanks for putting me through

Somena1:
Not sure if this is the right place to say this, my PR on RN addressing this issue was accepted 
[commit](https://github.com/facebook/react-native/commit/fc962c9)

Somena1:
Hello @gnprice please review my changes

Somena1:
@chrisbobbe please review changes in commit

gnprice:
> Not sure if this is the right place to say this, my PR on RN addressing this issue was accepted [commit](https://github.com/facebook/react-native/commit/fc962c9)

Cool! Thanks for getting that fixed upstream.

No need to @-mention maintainers pinging for a review. It can take a few business days before we review a change, but we do see them.

For cross-reference, this PR is the continuation of #5073.

This change looks good! Thanks for your work on it. Merging.

gnprice:
And done. I also revised the commit message -- take a look:

    android: Fix restart on resizing or splitting screen
    
    This fixes an issue where the app would restart on entering split
    screen, or on resizing the app's side of a split screen.
    
    Basically this line of the Android manifest is making an assurance to
    the system that the app is prepared to handle these changes; docs here:
      https://developer.android.com/guide/topics/manifest/activity-element#config
    When the system doesn't get that assurance, it just restarts the app
    from scratch to be sure.
    
    And empirically the app indeed does just fine when we add those
    assurances and then go ahead and switch to split-screen mode, or
    resize the app's portion of the split screen.  That layer of things is
    basically all RN's job, so RN is handling this just fine -- we just
    haven't been letting the system know that we're ready to do so.
    
    Also got this fixed in the RN template app upstream:
      https://github.com/facebook/react-native/pull/32536
      https://github.com/facebook/react-native/commit/fc962c9b6
    
    [greg: expanded commit message]
    
    Fixes: #5068

Thanks again!

PR #5105, opened 2021-11-05
per-account 6/n: Stop letting GlobalState cast to PerAccountState

gnprice:
This is the next PR in the series after #5083, #5066, #5030, #5016, #5017, and #5023, produced from the branch described at [#5006 (comment)](https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909).

> Coming next, we'll tighten the distinction between `GlobalState` and `PerAccountState` in the same way as we did for `GlobalDispatch` and `Dispatch`.

That's this PR.

Then, continuing as described last time:

> After that, we'll start distinguishing plain actions: just like other areas of our code before the start of this PR series, most of them implicitly refer to the active account while others don't, and we'll start tracking which is which so that we can later reinterpret the former group as referring to whichever specific account their caller had in mind.

### DIFF
--- a/src/reduxTypes.js
+++ b/src/reduxTypes.js
@@ -405,7 +405,7 @@ export type UsersState = $ReadOnlyArray<User>;
  * parts of our code will in that future operate on a particular account and
  * which parts will operate on all accounts' data or none.
  */
-export type PerAccountState = $ReadOnly<{
+type PerAccountStateImpl = $ReadOnly<{
   // TODO(#5006): Secretly we assume these objects also have `Account` data,
   //   like so:
   // accounts: [Account, ...mixed],
@@ -445,6 +445,8 @@ export type PerAccountState = $ReadOnly<{
   ...
 }>;
 
+export opaque type PerAccountState: PerAccountStateImpl = PerAccountStateImpl;
+
 /**
  * Our complete Redux state tree.
  *
@@ -474,10 +476,6 @@ export type GlobalState = $ReadOnly<{|
   accounts: AccountsState,
 |}>;
 
-// For now, under our single-active-account model, we want a GlobalState
-// to be seamlessly usable as a PerAccountState.
-(s: GlobalState): PerAccountState => s; // eslint-disable-line no-unused-expressions
-
 /**
  * Assume the given per-account state object is secretly a GlobalState.
  *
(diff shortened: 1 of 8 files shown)


src/reduxTypes.js
```diff
@@ -434,6 +434,8 @@ export type PerAccountState = $ReadOnly<{
   ...
 }>;
 
+export opaque type PerAccountState: PerAccountStateImpl = PerAccountStateImpl;
```
chrisbobbe, line 437:
```
redux types: Stop letting GlobalState pose as PerAccountState
```

In particular (and I'm not sure if this needs a mention in the commit, I just want to check my understanding), this prevents GlobalState from posing as a PerAccountState _in code outside this file_. We totally can do that within the file, and in fact we do:

```js
// For now, under our single-active-account model, we want a GlobalState
// to be seamlessly usable as a PerAccountState.
(s: GlobalState): PerAccountState => s; // eslint-disable-line no-unused-expressions
```

chrisbobbe (commented):
Thanks! Just a few small comments below to check my understanding; if all is good please merge at will.

src/reduxTypes.js
```diff
@@ -482,6 +482,24 @@ export function assumeSecretlyGlobalState(state: PerAccountState): GlobalState {
   return state;
 }
 
+/**
+ * Use the given state object as a per-account state.
+ *
+ * TODO(#5006): We'll have to fix and eliminate each call to this.
+ */
+export function dubPerAccountState(state: GlobalState): PerAccountState {
```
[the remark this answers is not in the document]
gnprice: It basically is. I had a distinction in my mind when I picked a different name here... but I think at the end of this PR, where `GlobalState` is no longer a subtype of `PerAccountState` any more than vice versa, the distinction no longer means anything, so probably it wasn't helpful in the first place.

(I think the idea was that because at the time `GlobalState <: PerAccountState`, it wasn't "secret" but openly there in the types that a `state: GlobalState` was in fact a `PerAccountState`, and the function was just a way of making that conversion explicitly.)

src/reduxTypes.js
gnprice: Yes, that's right. Mmm, and I should update that comment and/or cast, too, because they become misleading with this change.

gnprice:
OK, merged. Thanks for the review!

I fixed the second thing, but I let the confusing asymmetry of "dub" vs. "assume" stand for now.

Once I finish applying the per-account vs. global distinction generally, with the remaining phase of that being actions, the next bundle of work for me in this direction will be cleaning up one by one the specific remaining places we break it, most of which are calls to those functions. So as part of that cleanup I might rename for more symmetry.

PR #5106, opened 2021-11-05
tools/test: Use `local` throughout

gnprice:
Shell variables are global by default.  Thankfully Bash has a feature
where you can at least make them local, if you remember to do so.

This fixes a bug since 1db6b89ed where running the `lint` suite would
clobber the `files` variable that remembered the `--diff` vs
`--all-files` vs default (since-branch-base) setting.  This meant that
later suites (jest, prettier, and deps, by default) would see an
invalid value for `files`.

For `jest`, this had the same effect as `--all-files`, making it
annoyingly slow for interactive use (but the same as we run in CI.)
For `prettier`, it would cause it not to run at all.

The other accidentally-global variables here didn't happen to collide
with anything, but best to do this consistently so they don't in the
future (and so we're more likely to remember to do it in future code,
and avoid this sort of bug.)

Reported-by: Chris Bobbe <cbobbe@zulip.com>
Debugged-by: Chris Bobbe <cbobbe@zulip.com>

### DIFF
--- a/tools/test
+++ b/tools/test
@@ -116,6 +116,7 @@ files_js() {
             tools/info changed-files | grep '^src/.*\.js$' || :
             ;;
         diff:*)
+            local commitish
             commitish="${files#diff:}"
             git diff --name-only --diff-filter=d "$commitish" \
                 | grep '^src/.*\.js$' || :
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, LGTM! Merged.

PR #5108, opened 2021-11-05
notification: Use user_id to find account if multiple on same realm

chrisbobbe:
Along with making the property required on APNs and FCM payloads.

I noticed the opportunity to do this after reviewing #5100 and #5105.

Fixes: #4631

### DIFF
--- a/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
@@ -112,6 +112,7 @@ data class MessageFcmMessage(
     fun dataForOpen(): Bundle = Bundle().apply {
         // NOTE: Keep the JS-side type definition in sync with this code.
         putString("realm_uri", identity.realmUri.toString())
+        identity.userId?.let { putInt("user_id", it) }
         when (recipient) {
             is Recipient.Stream -> {
                 putString("recipient_type", "stream")
--- a/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
@@ -111,7 +111,7 @@ data class MessageFcmMessage(
      */
     fun dataForOpen(): Bundle = Bundle().apply {
         // NOTE: Keep the JS-side type definition in sync with this code.
-        identity.realmUri.let { putString("realm_uri", it.toString()) }
+        putString("realm_uri", identity.realmUri.toString())
         when (recipient) {
             is Recipient.Stream -> {
                 putString("recipient_type", "stream")
--- a/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
+++ b/android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
@@ -190,6 +190,8 @@ private fun extractIdentity(data: Map<String, String>): Identity =
     Identity(
         serverHost = data.require("server"),
         realmId = data.require("realm_id").parseInt("realm_id"),
+
+        // `realm_uri` was added in server version 1.9.0
         realmUri = data.require("realm_uri").parseUrl("realm_uri"),
 
         // Server versions from 1.6.0 through 2.0.0 (and possibly earlier
--- a/src/notification/extract.js
+++ b/src/notification/extract.js
@@ -2,6 +2,7 @@
 import PushNotificationIOS from '@react-native-community/push-notification-ios';
 
 import type { Notification } from './types';
+import { makeUserId } from '../api/idTypes';
 import type { JSONable, JSONableDict, JSONableInput, JSONableInputDict } from '../utils/jsonable';
 import * as logging from '../utils/logging';
 
@@ -185,6 +186,7 @@ export const fromAPNsImpl = (rawData: ?JSONableDict): Notification | void => {
 
   const identity = {
     realm_uri,
+    ...(user_id === undefined ? Object.freeze({}) : { user_id: makeUserId(user_id) }),
   };
 
   if (recipient_type === 'stream') {
--- a/src/notification/extract.js
+++ b/src/notification/extract.js
@@ -172,11 +172,14 @@ export const fromAPNsImpl = (rawData: ?JSONableDict): Notification | void => {
     throw err('invalid');
   }
 
-  const { realm_uri } = zulip;
+  const { realm_uri, user_id } = zulip;
   if (realm_uri === undefined) {
     throw err('archaic (pre-1.9.x)');
   }
   if (typeof realm_uri !== 'string') {
+    throw err('invalid');
+  }
+  if (user_id !== undefined && typeof user_id !== 'number') {
     throw err('invalid');
   }
 
--- a/src/notification/extract.js
+++ b/src/notification/extract.js
@@ -183,16 +183,20 @@ export const fromAPNsImpl = (rawData: ?JSONableDict): Notification | void => {
     throw err('invalid');
   }
 
+  const identity = {
+    realm_uri,
+  };
+
   if (recipient_type === 'stream') {
     const { stream, topic } = zulip;
     if (typeof stream !== 'string' || typeof topic !== 'string') {
       throw err('invalid');
     }
     return {
+      ...identity,
       recipient_type: 'stream',
       stream,
       topic,
-      realm_uri,
     };
   } else {
     /* recipient_type === 'private' */
@@ -207,16 +211,16 @@ export const fromAPNsImpl = (rawData: ?JSONableDict): Notification | void => {
         throw err('invalid');
       }
       return {
+        ...identity,
         recipient_type: 'private',
         pm_users: ids.sort((a, b) => a - b).join(','),
-        realm_uri,
       };
     }
 
     if (typeof sender_email !== 'string') {
       throw err('invalid');
     }
-    return { recipient_type: 'private', sender_email, realm_uri };
+    return { ...identity, recipient_type: 'private', sender_email };
   }
 
   /* unreachable */
--- a/src/notification/index.js
+++ b/src/notification/index.js
@@ -4,7 +4,7 @@ import PushNotificationIOS from '@react-native-community/push-notification-ios';
 import type { PushNotificationEventName } from '@react-native-community/push-notification-ios';
 
 import type { Notification } from './types';
-import type { Auth, Dispatch, GlobalDispatch, Identity, Narrow, UserId, UserOrBot } from '../types';
+import type { Auth, Dispatch, GlobalDispatch, Account, Narrow, UserId, UserOrBot } from '../types';
 import { topicNarrow, pm1to1NarrowFromUser, pmNarrowFromRecipients } from '../utils/narrow';
 import type { JSONable, JSONableDict } from '../utils/jsonable';
 import * as api from '../api';
@@ -25,14 +25,14 @@ import { getAccounts } from '../directSelectors';
 /**
  * Identify the account the notification is for, if possible.
  *
- * Returns an index into `identities`, or `null` if we can't tell.
+ * Returns an index into `accounts`, or `null` if we can't tell.
  * In the latter case, logs a warning.
  *
- * @param identities Identities corresponding to the accounts state in Redux.
+ * @param accounts The accounts state in Redux.
  */
 export const getAccountFromNotificationData = (
   data: Notification,
-  identities: $ReadOnlyArray<Identity>,
+  accounts: $ReadOnlyArray<Account>,
 ): number | null => {
   const { realm_uri } = data;
   if (realm_uri == null) {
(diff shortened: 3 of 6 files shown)


android/app/src/test/java/com/zulipmobile/notifications/FcmMessageTest.kt
```diff
@@ -142,11 +142,6 @@ class MessageFcmMessageTest : FcmMessageTestBase() {
         )
     }
 
-    @Test
-    fun `optional fields missing cause no error`() {
-        expect.that(parse(Example.pm.minus("user_id")).identity.userId).isNull()
```
gnprice, line 147: This should turn into a case where we test that required fields are required. See 9620629a7571c7d981c54384cfe9f3d72c9fadbc or other recent changes in `git log --stat -p android/app/src/test/`.

src/notification/extract.js
```diff
@@ -70,23 +71,6 @@ const asDict = (obj: JSONableInput | void): JSONableInputDict | void => {
...
-    get either of these from a server via the API, so this would not be possible
-    until the addition of `realm_uri` in 1.8.0-2150-g5f8d193bb7...
-
-    ... which still didn't permit differentiating between multiple accounts on
-    the same realm. This was only made possible by the addition of the `user_id`
-    field, in 2.1-dev-540-g447a517e6f.
-
-    TODO(server-1.8): Simplify this comment a lot.
-    TODO(server-1.9): Simplify further.
-    TODO(server-2.1): Simplify further.
```
gnprice, line 89: In addition to the paragraphs just above, the other simplification I had in mind here is to delete the distinctions between "antiquity" and the earliest supported version on the various properties, and to simplify the paragraphs at the top.

src/notification/extract.js
```diff
@@ -172,13 +156,19 @@ export const fromAPNsImpl = (rawData: ?JSONableDict): Notification | void => {
...
-  const { realm_uri } = zulip;
+  const { realm_uri, user_id } = zulip;
   if (realm_uri === undefined) {
     throw err('archaic (pre-1.9.x)');
   }
   if (typeof realm_uri !== 'string') {
     throw err('invalid');
   }
+  if (user_id === undefined) {
+    throw err('archaic (pre-2.1.x)');
```
gnprice, line 167: This should get a test case. See the existing mentions of `archaic` in the tests.

... Which raises the question why this "baseline" case didn't break:
```js
      // baseline
      expect(make({ realm_uri, recipient_type: 'private', sender_email })).toBeTruthy();
      // missing recipient_type
      expect(make({ realm_uri, sender_email })).toThrow(/archaic/);
```
It looks like the answer is that it's passing because the `make` return value is just a closure, and so always truthy -- it needs to be called to actually test anything. Oops. So a prep commit should fix that.

src/notification/types.js
```diff
@@ -1,4 +1,10 @@
 // @flow strict-local
+import type { UserId } from '../types';
+
+type NotificationBase = {|
+  realm_uri: string,
+  user_id: UserId,
```
gnprice, line 6: notif: Expose payload's realm_uri, present since server 2.1, to inner code

s/realm_uri/user_id/?

src/notification/index.js
```diff
@@ -25,15 +25,15 @@ import { getAccounts } from '../directSelectors';
...
- * Returns an index into `identities`, or `null` if we can't tell.
+ * Returns an index into `accounts`, or `null` if we can't tell.
  *
- * @param identities Identities corresponding to the accounts state in Redux.
+ * @param accounts The accounts state in Redux.
  */
 export const getAccountFromNotificationData = (
   data: Notification,
-  identities: $ReadOnlyArray<Identity>,
+  accounts: $ReadOnlyArray<Account>,
```
gnprice, line 34: The Identity type doesn't have the user ID, and it's not super
    convenient to add it there. That's because the Identity type is
    based on the Auth type (and we make Identity objects out of Auth
    objects at runtime), and the Auth type doesn't have the user ID.
    It's not super convenient to give it to the Auth type because there
    are places where we make Auth objects but we don't have a user ID,
    like `authFromCallbackUrl` in src/start/webAuth.js.

Oof. The `Auth` does of course have the email. So anywhere we have an Auth and can't readily make an Identity, because we lack the user ID, is an obstacle to converting from emails to IDs, #3764.

In `authFromCallbackUrl`, we're decoding one of those `zulip://` URLs the server used to complete a web login and get us the API key. So we should get the server to start giving us user IDs there, even if we won't be able to count on their presence for a while. That should be a thread in `#api design`, then.

src/notification/index.js
```diff
@@ -79,10 +79,8 @@ export const getAccountFromNotificationData = (
       realm_uri,
       parsed_url: realmUrl,
       match_count: urlMatches.length,
-      unique_identities_count: new Set(urlMatches.map(matchIndex => identities[matchIndex].email))
-        .size,
+      unique_accounts_count: new Set(urlMatches.map(matchIndex => accounts[matchIndex].email)).size,
```
gnprice: Best to keep the name for this one commit. :slightly_smiling_face: 
* The commit is marked NFC, which it isn't if a name sent in logging changes.
* "Unique identities" is still an accurate description, and clearer what it means, because we're only counting up distinct emails (for this realm), not looking for other properties of an `Account`.

src/notification/index.js
```diff
@@ -79,10 +79,8 @@ export const getAccountFromNotificationData = (
...
     });
-    // TODO get user_id into accounts data, and use that
```
gnprice, line 85: nit: this TODO comment should go away in the later commit where we start using the user_id

src/notification/index.js
```diff
@@ -71,19 +71,14 @@ export const getAccountFromNotificationData = (
...
+    // based on the notification's `user_id`.
+    const userMatch = urlMatches.find(urlMatch => accounts[urlMatch].userId === user_id);
+    if (userMatch == null) {
+      logging.warn('notification account not found with realm_uri and user_id');
+      return null;
+    }
+    return userMatch;
   }
 
   return urlMatches[0];
```
gnprice, line 84: Let's do the search for a user_id match regardless of whether there are one or several realm-URL matches. That way we notice if the one account in the realm isn't actually for the right user.

android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
```diff
@@ -191,19 +192,7 @@ private fun extractIdentity(data: Map<String, String>): Identity =
...
-        // but it'd be inherently buggy, and the bug it'd introduce seems
-        // likely to affect more users than the bug it'd fix.  So just ignore.
-        // TODO(server-2.0): Delete this comment.
-        // (data["user"] ignored)
-
-        // `user_id` was added in server version 2.1.0 (released 2019-12-12;
-        // commit 447a517e6, PR #12172.)
-        // TODO(server-2.1): Require this.
-        userId = data["user_id"]?.parseInt("user_id")
+        userId = data.require("user_id").parseInt("user_id")
```
gnprice, line 195: Let's leave on this a one-line comment saying it was new in 2.1.  That'll be helpful to refer to when seeing Sentry warnings about dropping FCM messages where it's absent.

(I see I didn't do this in 9620629a7571c7d981c54384cfe9f3d72c9fadbc, but I realize now I should have.)

android/app/src/main/java/com/zulipmobile/notifications/FcmMessage.kt
```diff
@@ -112,6 +112,7 @@ data class MessageFcmMessage(
     fun dataForOpen(): Bundle = Bundle().apply {
         // NOTE: Keep the JS-side type definition in sync with this code.
         identity.realmUri.let { putString("realm_uri", it.toString()) }
+        identity.userId.let { putInt("user_id", it) }
```
gnprice, line 115: The `let` here is redundant; can simplify it out.

(It's redundant on the previous line too. Looks like it got left behind in 9620629a7571c7d981c54384cfe9f3d72c9fadbc when I made `realmUri` non-nullable; before that it was `?.let` and was doing something useful.)

gnprice (commented):
Thanks! It'll be great to fix this bug.

Specific comments below. Then more broadly, we just talked through in the office the question of requiring this 2.1+ feature, and concluded we aren't ready to drop notifications entirely from pre-2.1 servers. But this can be revised to keep it as optional, just pass it through, and make use of it.

chrisbobbe:

> Then more broadly, we just talked through in the office the question of requiring this 2.1+ feature, and concluded we aren't ready to drop notifications entirely from pre-2.1 servers. But this can be revised to keep it as optional, just pass it through, and make use of it.

Indeed. Because of this, there are several of your comments (from before we decided that) that I haven't acted on—please let me know if I've dropped some that I shouldn't have. 🙂

src/notification/index.js
```diff
@@ -71,23 +71,33 @@ export const getAccountFromNotificationData = (
...
+        },
+      );
+      return null;
+    } else {
+      return urlMatches[0];
+    }
   }
 
-  return urlMatches[0];
+  // The user has several accounts in the notification's realm. Pick one
```
gnprice, line 93: This isn't quite true yet -- this code will apply when there's just one element in `urlMatches`.

src/notification/index.js
```diff
@@ -71,23 +71,33 @@ export const getAccountFromNotificationData = (
...
+  // based on the notification's `user_id`.
+  const userMatch = urlMatches.find(urlMatch => accounts[urlMatch].userId === user_id);
+  if (userMatch == null) {
+    logging.warn('notification account not found with realm_uri and user_id');
```
gnprice, line 97: Although at this point we have a `user_id` from the notification, it may be the case that the account it's meant for doesn't have `userId` non-null in the accounts state. That's because:
```js
  /**
   * The user's numeric ID.
   *
   * We learn the user ID each time we complete an initial fetch.
   *
   * This is `null` briefly when we've logged in but not yet completed our
   * first initial fetch on the account.  It's also `null` when representing
   * an account which was last used on a version of the app which didn't
   * record this information.  It's never `null` for an account for which we
   * have server data.
   */
```

Probably if there's just one URL match, and its `userId` is null, then we should go for it.

gnprice:
Thanks for the revision! This looks good, modulo a couple of things in the last commit. Comments below, and it looks like https://github.com/zulip/zulip-mobile/pull/5108#discussion_r744034514 still applies.

chrisbobbe:
> it looks like https://github.com/zulip/zulip-mobile/pull/5108#discussion_r744034514 still applies.

Oh, actually that was one that I thought I'd fixed.

src/notification/index.js
chrisbobbe: > Probably if there's just one URL match, and its `userId` is null, then we should go for it.

My current revision uses a slightly different algorithm. If there are two URL matches, and one of them has a `userId` but it's the wrong one, but the other one is `null`, we still want to go for the `null` one.

src/notification/index.js
```diff
@@ -71,23 +71,44 @@ export const getAccountFromNotificationData = (
...
+      return urlMatches[0];
+    }
   }
 
-  return urlMatches[0];
+  // There may be multiple accounts in the notification's realm. Pick one
+  // based on the notification's `user_id`.
+  const userMatch = urlMatches.find(urlMatch => {
+    const { userId } = accounts[urlMatch];
+    return userId !== null && userId === user_id;
```
gnprice, line 97: This `userId !== null` is always true if `userId === user_id` is, right? So it can be dropped, and this function simplifies back down to just returning an expression.

src/notification/index.js
```diff
@@ -71,23 +71,44 @@ export const getAccountFromNotificationData = (
...
+    // Maybe we didn't get a userId match because the correct account just
+    // hasn't had its userId recorded on it yet. See jsdoc on the Account
+    // type for when that is.
+    const nullUserIdMatches = urlMatches.filter(urlMatch => accounts[urlMatch].userId === null);
+    if (nullUserIdMatches.length === 1) {
+      return nullUserIdMatches[0];
+    }
+
+    logging.warn('notification account not found with realm_uri and user_id');
+    return null;
```
gnprice, line 109: This log message won't always be accurate -- there could have been two accounts that had the right realm_uri and null user_id.

(Well, it could be read as accurate depending on how one interprets "found with realm_uri and user_id". But a reading that makes it accurate would make the same message equally accurate in the case where there's a match we actually use, just above here.)

We've been pretty precise in the different error cases in this function so far, because it's been a fairly common set of errors and it's good to know exactly what conditions are driving that. So let's make the distinction here too between zero matches and multiple matches.

gnprice (commented):
Thanks! A couple of comments below.

gnprice:
Oh also: let's mark this as fixing #4631! This is an issue I run into regularly when using test accounts -- looking forward to getting it fixed. :slightly_smiling_face:

chrisbobbe:
Thanks! Revision pushed.

gnprice:

PR #5110, opened 2021-11-05
api [nfc]: Explain InitialData more; simplify register-queue interface

gnprice:
This started from the note here:
https://github.com/zulip/zulip-mobile/pull/4968#discussion_r696943253
to better explain how `initialDataTypes.js` is organized, and expanded as I noticed that the register-queue binding could be 
cleaned up to make its interface better reflect how it actually behaves.


chrisbobbe:
Ah, looks like the CI failure is that some snapshot-test snapshots should be updated.

gnprice:
Yeah -- the snapshot change just corresponds to the change in the example data. Pushed a revision.

chrisbobbe:

PR #5111, opened 2021-11-05
webview: Have 'content' event only replace what's in msglistElementsDiv

chrisbobbe:
Instead of replacing all of what's in htmlBody.

This is a step we're doing on its own, on the way to message-list
diffing, for #3851. When this is released, we can get some empirical
confirmation that there's not some state being tracked somewhere in
the DOM that's under htmlBody but not under msglistElementsDiv, that
we're no longer resetting, since that could be a buggy experience.

We don't think there is any state like that, from brief testing and
reading the code.

The most likely-looking bit is whether the div#message-loading has
the "hidden" class or not, IOW, whether or not we're covering the
view with our "fetching messages" indicator. After all, before this
change, this event was setting that state based on a
`showMessagePlaceholders` boolean.

It should be OK that we're not doing that anymore. That's the job of
the `updateFetching` inbound event. Before e26a59d5d, we were
failing to send that event along with this 'content' event, because
of an ill-considered early return. But that's gone now, and testing
suggests that that state is being set as it should be.

### DIFF
--- a/src/webview/generateInboundEvents.js
+++ b/src/webview/generateInboundEvents.js
@@ -5,7 +5,6 @@ import isEqual from 'lodash.isequal';
 import type { Auth, FlagsState } from '../types';
 import type { Props } from './MessageList';
 import type { UpdateStrategy } from '../message/messageUpdates';
-import htmlBody from './html/htmlBody';
 import messageListElementHtml from './html/messageListElementHtml';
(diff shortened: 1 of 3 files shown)


gnprice:
Great! Indeed that is a surprisingly simple change. Looks good -- merging.

PR #5112, opened 2021-11-05
ios [nfc]: Finish cleaning up unused `AppIdentifierPrefix` logic

chrisbobbe:
Looks like we missed this in eb37e6c78.


gnprice:
Thanks! Looks good, merging.

PR #5113, opened 2021-11-08
per-account 7/n: Sort out plain actions

gnprice:
This is the next PR in the series after #5105, #5083, #5066, #5030, #5016, #5017, and #5023, produced from the branch described at [#5006 (comment)](https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909).

> After that, we'll start distinguishing plain actions: just like other areas of our code before the start of this PR series, most of them implicitly refer to the active account while others don't, and we'll start tracking which is which so that we can later reinterpret the former group as referring to whichever specific account their caller had in mind.

That's this PR.

And this is the end of that draft branch! :tada: 

There's still more to do for #5006. But at the end of this PR, we've accomplished this (from that comment https://github.com/zulip/zulip-mobile/issues/5006#issuecomment-924341909 where I described the draft branch these came from):

> That lets us use the type-checker to trace how the data flows around to ensure that we're consistent in which code gets which kind of type [i.e. per-account vs. global], and to identify exactly which spots of code rely on assumptions that won't stay true in the future where these objects are actually distinct.

The next task will be to fix those spots one by one.


chrisbobbe:
Thanks, this LGTM! Looks like just a Prettier failure on the last commit, otherwise please merge at will.

gnprice:
Ah indeed -- fixed, and merged. Thanks for the review!

PR #5122, opened 2021-11-10
Moving attachments row below the composeBox

### LINKED ISSUE #5080: Move compose box buttons into a row below the compose box
At present, compose box buttons are accessed via a "plus" symbol to the left of the compose box. Clicking on "plus" expands the buttons on the same line as the compose box. This works OK with the buttons we currently have, but is likely to no longer look good as we start adding more buttons (#3427, #5079).

We should try moving the row of buttons below the compose box. We may need to play with the UI to make sure the interactions feel natural, but we can try the following approach:

1) Putting the cursor into the compose box automatically opens a row of buttons below the compose box. Moving the cursor away from the compose box closes the buttons row.
2) Clicking on "plus" when the cursor is not in the compose box also automatically opens a row of buttons below the compose box.

This issue may be a blocker for #3427 and #5079. It's possible we can implement one of those without this issue, but not both.

[Discussion thread on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user.20feedback/near/1270607)

chetas411:
Fixes #5080
Now the attachments row shows up as soon as the input in the message is focused and it can be trigerred through the plus icon also but now we see it below the composeBox

https://user-images.githubusercontent.com/65273718/141056725-0503e985-190a-4463-8762-d4e40398afc6.mp4

### DIFF
--- a/src/common/Icons.js
+++ b/src/common/Icons.js
@@ -75,6 +75,7 @@ export const IconSettings: SpecificIconType = makeIcon(Feather, 'settings');
 export const IconRight: SpecificIconType = makeIcon(Feather, 'chevron-right');
 export const IconPlusCircle: SpecificIconType = makeIcon(Feather, 'plus-circle');
 export const IconLeft: SpecificIconType = makeIcon(Feather, 'chevron-left');
+export const IconUp: SpecificIconType = makeIcon(Feather, 'chevron-up');
 export const IconPeople: SpecificIconType = makeIcon(Feather, 'users');
 export const IconPerson: SpecificIconType = makeIcon(Feather, 'user');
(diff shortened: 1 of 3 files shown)


chetas411:
@chrisbobbe and @alya Can you reveiw the PR now?

chetas411:
@chrisbobbe and @alya I have made the final changes as you proposed  in #5080. Also I used `git pull --rebase` as described in the git guide provided by Zulip and then I pushed the commits to my branch "composeBox" , so why it's still showing branch can't be resbased here above?

chetas411:
https://user-images.githubusercontent.com/65273718/143764220-f7ccffd3-89e7-4939-a883-ac48f7c39702.mp4

PR #5123, opened 2021-11-10
redux: Fire actions for 'realm' events with op `update` or `update_dict`

chrisbobbe:
The bulk of this is just transferring information from the docs into our types:

- first, updating `InitialDataRealm`, for what we get in the `/register` response
- then, making types for `realm op: update` and `realm op: update_dict` events, based on `InitialDataRealm`

Next, we should do what Greg says [here](https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60value.60.20in.20realm.20op.3A.20update/near/1275442):

> […] I think what I'd want to do (and what really any sane client that has ambitions of handling more than one or two special cases would want to do, unless like the web client it's already been written) would be to convert them both into a common format, basically corresponding to an `update_dict` event, and have the code that handles the details of an event just handle that one form.


src/api/initialDataTypes.js
```diff
@@ -74,53 +74,197 @@ export type AvailableVideoChatProviders = $ReadOnly<{|
 |}>;
 
 export type InitialDataRealm = $ReadOnly<{|
```
gnprice, line 76: In the first commit, is the idea to catch up completely with the current list in the API docs? I think doing so would be good -- it'd help make it more tractable to review, and would also give us a good starting point for incremental updates from here. (Like "this is current to feature level 107", and then we can read the [changelog](https://zulip.com/api/changelog) starting from there to make a future update.)

Currently it isn't -- e.g. it doesn't have `max_stream_name_length`. (I searched on https://zulip.com/api/register-queue#response for "present if realm is", and that was the first result.)

src/api/initialDataTypes.js
```diff
@@ -74,53 +74,197 @@ export type AvailableVideoChatProviders = $ReadOnly<{|
...
+  development_environment: boolean,
+
+  // TODO(server-5.0): Added, at feat. 74.
+  event_queue_longpoll_timeout_seconds?: number,
+
+  jitsi_server_url?: string, // TODO: Really optional?
+  max_avatar_file_size_mib: number,
+  max_file_upload_size_mib: number,
+
+  // TODO(server-5.0): Replaced in feat. 72 by max_max_icon_file_size_mib
```
gnprice, line 91:
```suggestion
  // TODO(server-5.0): Replaced in feat. 72 by max_icon_file_size_mib
```

src/api/initialDataTypes.js
```diff
@@ -74,53 +74,197 @@ export type AvailableVideoChatProviders = $ReadOnly<{|
...
+  realm_night_logo_url: string,
   realm_notifications_stream_id: number,
   realm_password_auth_enabled: boolean,
+  realm_plan_type: number,
   realm_presence_disabled: boolean,
+  realm_private_message_policy: number,
+  realm_push_notifications_enabled: boolean,
+
+  // TODO: Do servers send this, here or in realm-update events?
   realm_restricted_to_domain: boolean,
```
gnprice, line 228: Hmm, yeah. I wonder if this was an old name for `realm_emails_restricted_to_domains` (or for an old version of that.)

src/__tests__/lib/exampleData.js
```diff
@@ -621,47 +621,92 @@ export const action = Object.freeze({
...
       realm_authentication_methods: { GitHub: true, Email: true, Google: true },
       realm_available_video_chat_providers: {},
+      realm_avatar_changes_disabled: false,
       realm_bot_creation_policy: 3,
       realm_bot_domain: 'example.com',
-      realm_create_stream_by_admins_only: true,
+      realm_community_topic_editing_limit_seconds: 600,
+      realm_create_private_stream_policy: 3,
+      realm_create_public_stream_policy: 3,
+      realm_create_web_public_stream_policy: 3,
```
gnprice, line 647: This example action should probably get a comment at the top warning these values may not be representative, like we have on users and messages.

(This particular property is a fun one because apparently the allowed values are 2, 4, 6, and 7. But it's fine to have 3 here! If we ever write a test that cares, we'll notice then.)

src/api/eventTypes.js
```diff
@@ -173,3 +175,236 @@ export type RestartEvent = $ReadOnly<{|
...
+
+  password_min_length: $PropertyType<InitialDataRealm, 'password_min_length'>,
+  password_min_guesses: $PropertyType<InitialDataRealm, 'password_min_guesses'>,
+
+  // TODO(server-5.0): See notes in InitialDataRealm.
+  // realm_add_custom_emoji_policy
+  // realm_add_emoji_by_admins_only
+  // realm_allow_community_topic_editing
+
+  allow_edit_history: $PropertyType<InitialDataRealm, 'realm_allow_edit_history'>,
```
gnprice, line 218: Hmm, using `$PropertyType` is a good idea here!

src/api/eventTypes.js
```diff
@@ -173,3 +175,236 @@ export type RestartEvent = $ReadOnly<{|
...
+
+  // realm_available_video_chat_providers doesn't have an event because it
+  //   can only change when the server is restarted anyway; see
+  //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60value.60.20in.20realm.20op.3A.20update/near/1276269
+
+  avatar_changes_disabled: $PropertyType<InitialDataRealm, 'realm_avatar_changes_disabled'>,
+  bot_creation_policy: $PropertyType<InitialDataRealm, 'realm_bot_creation_policy'>,
+  bot_domain: $PropertyType<InitialDataRealm, 'realm_bot_domain'>,
+
+  // TODO(server-3.0): See note in InitialDataRealm.
```
gnprice, line 234: For these comments that just refer to similar comments in InitialDataRealm, let's omit this copy. That'll help it be more uniform to read through, and let the more specific comments like on `realm_available_video_chat_providers` stand out better.

src/api/eventTypes.js
```diff
@@ -173,3 +175,236 @@ export type RestartEvent = $ReadOnly<{|
...
+
+  invite_to_stream_policy: $PropertyType<InitialDataRealm, 'realm_invite_to_stream_policy'>,
+  is_zephyr_mirror_realm: $PropertyType<InitialDataRealm, 'realm_is_zephyr_mirror_realm'>,
+  logo_source: $PropertyType<InitialDataRealm, 'realm_logo_source'>,
+  logo_url: $PropertyType<InitialDataRealm, 'realm_logo_url'>,
+  mandatory_topics: $PropertyType<InitialDataRealm, 'realm_mandatory_topics'>,
+  message_content_allowed_in_email_notifications: $PropertyType<
+    InitialDataRealm,
+    'realm_message_content_allowed_in_email_notifications',
+  >,
```
gnprice, line 303: I think this type is also a good place to do a `// prettier-ignore` so that it's possible to put the information in a more tabular form: all formatted the same way, and with the right-hand property names all starting at the same column so it's possible to scan down them easily by eye.

For example:
```js
  invite_to_stream_policy:
    $PropertyType<InitialDataRealm, 'realm_invite_to_stream_policy'>,
  is_zephyr_mirror_realm:
    $PropertyType<InitialDataRealm, 'realm_is_zephyr_mirror_realm'>,
  logo_source:
    $PropertyType<InitialDataRealm, 'realm_logo_source'>,
  logo_url:
    $PropertyType<InitialDataRealm, 'realm_logo_url'>,
  mandatory_topics:
    $PropertyType<InitialDataRealm, 'realm_mandatory_topics'>,
  message_content_allowed_in_email_notifications:
    $PropertyType<InitialDataRealm, 'realm_message_content_allowed_in_email_notifications'>,
```

(It's possible to edit from here to there pretty efficiently using multi-cursors: e.g. select `: $`, then hit Ctrl+D a bunch of times, then make the edit.)

Alternatively things could be kept on one line each, but that'd get awfully long in a few cases like `message_content_allowed_in_email_notifications`.

src/api/eventTypes.js
```diff
@@ -173,3 +175,236 @@ export type RestartEvent = $ReadOnly<{|
...
+|}>;
+
+// https://zulip.com/api/get-events#realm-update
+export type RealmUpdateDictEvent = $ReadOnly<{|
+  ...EventCommon,
+  type: typeof EventTypes.realm,
+
+  op: 'update_dict',
+  property: 'default',
+  data: RealmDataForUpdate,
```
gnprice, line 409: It looks like this will mean a type that has *all* the properties, not an arbitrary subset. But using `$Rest<…, { ... }>` should fix that.

src/api/eventTypes.js
```diff
@@ -173,3 +175,236 @@ export type RestartEvent = $ReadOnly<{|
...
+// https://zulip.com/api/get-events#realm-update
+export type RealmUpdateEvent = $ReadOnly<{|
+  ...EventCommon,
+  type: typeof EventTypes.realm,
+
+  op: 'update',
+
+  // TODO(flow): `property` and `value` should correspond
+  property: $Keys<RealmDataForUpdate>,
+  value: $ElementType<$Rest<RealmDataForUpdate, { ... }>, $Keys<RealmDataForUpdate>>,
```
gnprice, line 397: I think the way to do this TODO would be to parameterize the type over its `property` type. Like:
```js
export type RealmUpdateEvent<P: $Keys<RealmDataForUpdate>> = // …
  property: P,
  value: $ElementType<RealmDataForUpdate, P>,
  // …
```

But we can look into that after merge.

gnprice (commented):
Thanks for taking care of these! Comments below.

src/api/initialDataTypes.js
chrisbobbe: Eep, yes, I definitely meant to get all 82 matches for "present if realm is" there; looks like I missed at least the first few. I'm checking again now.

src/api/eventTypes.js
chrisbobbe: Yep! We'll get a nice shortening of all these in a future Flow version: https://flow.org/en/docs/types/utilities/#toc-propertytype

src/api/eventTypes.js
chrisbobbe: Cool, yeah. That occurred to me, and I tried it, but I ran into an issue with some code for handling the events. Let's debug together? After merge (or before) is fine.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+  // TODO: Check if the doc really means this, instead of max_icon_file_size
+  //   and max_icon_file_size_mib?
+  icon_file_size: number,
+
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'max_logo_file_size'>,
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'max_logo_file_size_mib'>,
+  password_min_length:
+    $PropertyType<InitialDataRealm, 'password_min_length'>,
```
gnprice, line 221: It looks like max_message_length and its neighbors didn't make it into this list.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+  //   $PropertyType<InitialDataRealm, 'realm_delete_own_message_policy'>,
+  description:
+    $PropertyType<InitialDataRealm, 'realm_description'>,
+  digest_emails_enabled:
+    $PropertyType<InitialDataRealm, 'realm_digest_emails_enabled'>,
+  digest_weekday:
+    $PropertyType<InitialDataRealm, 'realm_digest_weekday'>,
+
+  // TODO: Number or boolean? Event doc and register-response doc disagree
+  disallow_disposable_email_addresses: empty,
```
gnprice, line 280: Better `mixed` to represent the uncertainty. `empty` would mean that we could write code that consumed this and let that code make absolutely whatever assumptions it wanted to about the value; `mixed` instead means no assumptions allowed.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+ * shape of that data as InitialDataRealm. The properties are named
+ * differently between here and there (InitialDataRealm has lots of
+ * properties that start with "realm_"), but we expect the values to be the
+ * same types.
+ */
+// If there's a property that exists in InitialDataRealm, but we don't use
+// it and the event doc doesn't mention it, write the InitialDataRealm
+// property in a code comment in the place where it would go here.
+/* prettier-ignore */
+type RealmDataForUpdate = $ReadOnly<{
```
gnprice, line 195: Because this type is pretty long and boilerplate-y, and of a different flavor from the event types, I'd like to get it out of this file -- I think that will help keep the rest of the event types more readable.

Perhaps a file `src/api/realmDataTypes.js`?

Then the actual event types would still be in this file -- they'd just import it. This is a bit like how `StreamEvent` refers to the `Stream` type over in `modelTypes.js`.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+
+  // TODO(flow): `property` and `value` should correspond
+  property: $Keys<RealmDataForUpdate>,
+  value: $ElementType<RealmDataForUpdate, $Keys<RealmDataForUpdate>>,
+
+  extra_data: { upload_quota?: number, ... },
+|}>;
+
+// https://zulip.com/api/get-events#realm-update
+export type RealmUpdateDictEvent = $ReadOnly<{|
```
gnprice, line 420:
```suggestion
// https://zulip.com/api/get-events#realm-update_dict
export type RealmUpdateDictEvent = $ReadOnly<{|
```

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+  email_address_visibility:
+    $PropertyType<InitialDataRealm, 'realm_email_address_visibility'>,
+  email_auth_enabled:
+    $PropertyType<InitialDataRealm, 'realm_email_auth_enabled'>,
+  email_changes_disabled:
+    $PropertyType<InitialDataRealm, 'realm_email_changes_disabled'>,
+  emails_restricted_to_domains:
+    $PropertyType<InitialDataRealm, 'realm_emails_restricted_to_domains'>,
+  realm_giphy_rating:
+    $PropertyType<InitialDataRealm, 'realm_giphy_rating'>,
```
gnprice, line 293: Hmm, I wonder if this in the docs is accurate, or if it's actually `giphy_rating`. Seems like the general naming scheme in effect here would want it to be the latter.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+  bot_domain:
+    $PropertyType<InitialDataRealm, 'realm_bot_domain'>,
+  community_topic_editing_limit_seconds:
+    $PropertyType<InitialDataRealm, 'realm_community_topic_editing_limit_seconds'>,
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'realm_create_private_stream_policy'>,
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'realm_create_public_stream_policy'>,
+  create_stream_policy:
+    $PropertyType<InitialDataRealm, 'realm_create_stream_policy'>,
```
gnprice, line 256: And here the API docs seem to say that the event uses the old single property. That's probably not correct -- seems like the feature can't work right if it is.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+
+  // TODO: Do servers send this, here or in InitialDataRealm?
+  //   realm_show_digest_email
+
+  signup_notifications_stream_id:
+    $PropertyType<InitialDataRealm, 'realm_signup_notifications_stream_id'>,
+  upload_quota:
+    $PropertyType<InitialDataRealm, 'realm_upload_quota'>,
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'realm_upload_quota_mib'>,
```
gnprice, line 367: Another case where the API docs on the event side have the old name, and I suspect are wrong.

src/api/eventTypes.js
```diff
@@ -173,3 +175,253 @@ export type RestartEvent = $ReadOnly<{|
...
+  //   $PropertyType<InitialDataRealm, 'realm_create_web_public_stream_policy'>,
+  default_code_block_language:
+    $PropertyType<InitialDataRealm, 'realm_default_code_block_language'>,
+
+  // realm_default_external_accounts doesn't have an event because it can
+  //   only change when the server is restarted anyway; see
+  //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60value.60.20in.20realm.20op.3A.20update/near/1276269
+
+  // <insert_name>:
+  //   $PropertyType<InitialDataRealm, 'realm_default_language'>,
```
gnprice, line 267: Then there's quite a lot of properties like this one that aren't in the event-side doc. I'm not sure if that's an issue in the doc or if we just don't send events when these change; it might be some of each.

gnprice (commented):
Thanks! Small comments below.

Commit-message nits:
b2a2ea32b initialDataTypes: Sync 'realm' data in register response with the doc
fa9056c63 redux: Fire actions for 'realm' events with op `update` or `update_dict`

I'd give the first of these a more general prefix of `api types:`, and the second a prefix like `events:` instead of `redux:` -- it's really about our handling of events, not about Redux. (It does involve making Redux actions; but it's interacting with Redux only as a tool, using it in a boring way as it's meant to be used. The interesting part of the commit is about Zulip events.)

src/api/eventTypes.js
gnprice: Sounds good!

src/api/initialDataTypes.js
```diff
@@ -73,54 +73,211 @@ export type AvailableVideoChatProviders = $ReadOnly<{|
...
+  realm_create_public_stream_policy?: number,
+
+  // TODO(server-5.0): Replaced in feat. 102 by
+  // realm_create_private_stream_policy and realm_create_public_stream_policy
+  realm_create_stream_policy?: number,
+
+  // TODO(server-5.0): Added in feat. 103
+  realm_create_web_public_stream_policy?: number,
+
+  realm_default_code_block_language: string,
```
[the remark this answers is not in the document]
chrisbobbe: https://chat.zulip.org/#narrow/stream/378-api-design/topic/.22type.22.20annotations.20and.20.60null.60/near/1282157

src/api/initialDataTypes.js
```diff
@@ -73,54 +73,211 @@ export type AvailableVideoChatProviders = $ReadOnly<{|
...
+  // TODO(server-5.0): Added in feat. 75, replacing realm_allow_community_topic_editing
+  realm_edit_topic_policy?: number,
+
+  realm_email_address_visibility: number,
   realm_email_auth_enabled: boolean,
   realm_email_changes_disabled: boolean,
+  realm_emails_restricted_to_domains: boolean,
+
+  // TODO(server-4.0): Added in feat. 55.
+  realm_giphy_rating?: number,
```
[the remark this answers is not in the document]
chrisbobbe: https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60giphy_rating_options.60.20in.20.2Fregister.20response/near/1282161

src/api/eventTypes.js
chrisbobbe: [Yeah, I'm not sure. ](https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60value.60.20in.20realm.20op.3A.20update/near/1282173)

chrisbobbe:

gnprice:
Looks good! Merging.

Issue #4687, opened 2021-04-21
Unhandled Zulip API event: type: "attachment"


WesleyAC:
When uploading an image, I see the following error in the log:

```
[Wed Apr 21 2021 23:16:26.796]  ERROR    Unhandled Zulip API event
[Wed Apr 21 2021 23:16:26.844]  ERROR        type: "attachment"
    op: "add"
    attachment: {"id":22548,"name":"Screenshot_20210421-231422.png","path_id":"2/16/aNNbCWo8ptgK6y-FfP6mYEB8/Screenshot_20210421-231422.png","size":148949,"create_time":1619018185000,"messages":[]}
    upload_space_used: 10630145314
    id: 12
```

There's no user-visible impact, as far as I can tell, but we should figure out what's causing this and what the correct fix is.

chrisbobbe:
Looks like that's this event: https://zulip.com/api/get-events#attachment-add

Since we get a new-message event when there's a new (attachment-containing) message, and we get an update-message event when there's an updated (attachment-containing) message, I can't think of a live-updating bug that might come from this.

Maybe if we were keeping and relying on a data structure specifically for attachments, there would be such a bug?

timabbott:
That event is intended for UI on the "Uploaded files" settings page so I'd expect mobile to not want to request it.

PR #5125, opened 2021-11-10
Don't immediately send & send multiple images.

AkashDhiman:
TODO:

- [ ] Choose better name for `toAttachmentResponse` method.
- [ ]  Break into multiple commits.
- [ ]  `FileName` isn't appropriate, `react-native-image-picker` modifies
  it. (it's along the line of `rn_image_picker_lib_temp_*`)

Fixes: #2366
Fixes: #4540


AkashDhiman:
An error without any side effect as far as i can tell right now happens whenever I attach any file (image or otherwise)

chrisbobbe:
> An error without any side effect as far as i can tell right now happens whenever I attach any file (image or otherwise)
> 
> […]
> 
> I wonder what's that about?

That looks like #4687.

chrisbobbe:
Closing as superseded by #5474 and #5638. Thanks for your work! 🙂

PR #5130, opened 2021-11-17
compose: Explain validation errors on press of disabled submit button

chrisbobbe:
Like 630a543de87355dcae67f5eeffd9051738cde1a6, but for the compose box instead of the share-to interface.

We could have done something like bb90a7e48f89881fbfd6c6fed64ccc050f10ea9a to the `FloatingActionButton`, where we add an `isPressHandledWhenDisabled` prop. But we already know that we don't want to keep using that component here, because the button isn't a FAB at all. So, stop using it.

### DIFF
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -594,7 +594,7 @@ class ComposeBoxInner extends PureComponent<Props, State> {
           {(props => {
             // TODO: Integrate this into the surrounding code.
 
-            const { size, disabled, onPress, Icon } = props;
+            const { size, disabled, Icon } = props;
             const iconSize = Math.trunc(size / 2);
             const customWrapperStyle = {
               width: size,
@@ -609,7 +609,7 @@ class ComposeBoxInner extends PureComponent<Props, State> {
             return (
               <Touchable
                 style={this.styles.composeSendButton}
-                onPress={disabled ? undefined : onPress}
+                onPress={disabled ? undefined : this.handleSend}
                 accessibilityLabel="Send message"
               >
                 <View style={[fabStyles.wrapper, customWrapperStyle]}>
@@ -621,7 +621,6 @@ class ComposeBoxInner extends PureComponent<Props, State> {
             Icon: isEditing ? IconDone : IconSend,
             size: 32,
             disabled: message.trim().length === 0 || this.state.numUploading > 0,
-            onPress: this.handleSend,
           })}
         </View>
       </View>
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -484,6 +484,8 @@ class ComposeBoxInner extends PureComponent<Props, State> {
     },
   };
 
+  submitButtonHitSlop = { top: 8, right: 8, bottom: 8, left: 8 };
+
   render() {
     const { isTopicFocused, isMenuExpanded, height, message, topic, selection } = this.state;
     const {
@@ -592,14 +594,27 @@ class ComposeBoxInner extends PureComponent<Props, State> {
             />
           </View>
           <View style={this.styles.submitButtonContainer}>
-            <Touchable
-              style={[this.styles.submitButton, { opacity: submitButtonDisabled ? 0.25 : 1 }]}
-              onPress={submitButtonDisabled ? undefined : this.handleSend}
-              accessibilityLabel="Send message"
-              hitSlop={{ top: 8, right: 8, bottom: 8, left: 8 }}
+            <View
+              // Mask the Android ripple-on-touch so it doesn't extend
+              //   outside the circle…
+              // TODO: `Touchable` should do this, and the `hitSlop`
+              //   workaround below.
+              style={{
+                borderRadius: this.styles.submitButton.borderRadius,
+                overflow: 'hidden',
+              }}
+              // …and don't defeat the `Touchable`'s `hitSlop`.
+              hitSlop={this.submitButtonHitSlop}
             >
-              <SubmitButtonIcon size={16} color="white" />
-            </Touchable>
+              <Touchable
+                style={[this.styles.submitButton, { opacity: submitButtonDisabled ? 0.25 : 1 }]}
+                onPress={submitButtonDisabled ? undefined : this.handleSend}
+                accessibilityLabel="Send message"
+                hitSlop={this.submitButtonHitSlop}
+              >
+                <SubmitButtonIcon size={16} color="white" />
+              </Touchable>
+            </View>
           </View>
         </View>
       </View>
--- a/src/compose/ComposeBox.js
+++ b/src/compose/ComposeBox.js
@@ -597,7 +597,6 @@ class ComposeBoxInner extends PureComponent<Props, State> {
           {(() => {
             // TODO: Integrate this into the surrounding code.
 
-            const iconSize = Math.trunc(32 / 2);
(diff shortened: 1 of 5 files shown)


src/compose/ComposeBox.js
```diff
@@ -581,14 +636,16 @@ class ComposeBoxInner extends PureComponent<Props, State> {
...
-            size={32}
-            disabled={message.trim().length === 0 || this.state.numUploading > 0}
-            onPress={this.handleSend}
-          />
+          <View style={componentStyles.submitButtonContainer}>
+            <Touchable
+              style={[componentStyles.submitButton, { opacity: submitButtonDisabled ? 0.25 : 1 }]}
+              onPress={this.handleSubmit}
+              accessibilityLabel={isEditing ? _('Save message') : _('Send message')}
+              hitSlop={{ top: 8, right: 8, bottom: 8, left: 8 }}
```
gnprice, line 644: Neat!

src/compose/ComposeBox.js
```diff
@@ -432,49 +502,30 @@ class ComposeBoxInner extends PureComponent<Props, State> {
...
-  };
+  getValidationErrors = (): $ReadOnlyArray<ValidationError> => {
+    const { mandatoryTopics } = this.props;
+    const destinationNarrow = this.getDestinationNarrow();
+    const { message } = this.state;
+
+    const result = [];
 
-  styles = {
-    wrapper: {
```
gnprice, line 444: ComposeBox [nfc]: Bring out `componentStyles`

I'd rather keep the styles here where they are -- it means they're there right next to the code in the `render` method that refers to them. Otherwise they end up separated from that code by a few hundred lines of interaction logic, which doesn't really care about the layout details or vice versa.

src/compose/ComposeBox.js
```diff
@@ -432,49 +502,30 @@ class ComposeBoxInner extends PureComponent<Props, State> {
...
-      borderWidth: 0,
-      borderRadius: 5,
-      fontSize: 15,
-      flexShrink: 1,
-      ...this.inputMarginPadding,
-    },
+    if (
+      isTopicNarrow(destinationNarrow)
+      && topicOfNarrow(destinationNarrow) === apiConstants.NO_TOPIC_TOPIC
+      && mandatoryTopics
```
gnprice, line 515: nit: let's put `mandatoryTopics` first -- that sets the context for what the rest of the condition is about, and it also can short-circuit the rest of the logic early (though the latter doesn't seriously matter because this is all pretty cheap.)

gnprice: Ah I see, this is verbatim from the old code at `handleSend`. Shrug.

You can turn it into a code move by squashing that next commit

    ComposeBox [nfc]: Deduplicate expression for mandatory-topic validation

into this one. The diff gets just a bit bigger, but the real complexity of the diff is actually smaller, I think (even just comparing the squashed commit to this first one, and more so if comparing to both the commits), because this code gets moved instead of introduced.

gnprice (commented):
Thanks @chrisbobbe! This all looks good -- just small comments below. Glad to have that FAB that isn't floating cleared up.

I'll also fire this up in an emulator and play with it a bit.

gnprice:
OK, tried it in an emulator and it looks good.

… Well, there's some terrible lag when the send button changes from disabled to enabled and back -- like 500ms or more. But that turns out to be the same on main. I don't usually notice, because the lag is a lot less than that (though still noticeable when I look) on my phone, and I think it's also a lot less on my home desktop -- this desktop in the office has a good bit slower of a CPU.

Also it looks weird that the splash effect on touching the button extends out to a square around the button, instead of being confined to the actual button. But that's also… oh hmm. Actually that one *has* changed.
* In main, it extends to the full height of the compose box (at least when it's one line high).
* With this branch, it extends only to a square the height and width of the button.

I think the new version actually heightens the oddness of that visual effect somewhat, because now the effect stops at a boundary that's in the middle of the gray background of the compose box.

chrisbobbe:
> Also it looks weird that the splash effect on touching the button extends out to a square around the button, instead of being confined to the actual button. But that's also… oh hmm. Actually that one _has_ changed.

Hmm, yeah.

We looked at this in the office today, and found that `ZulipButton` isn't affected by this problem. From some experimentation, this appears to be because the `<Touchable />` there is wrapped in a `<View />` that has a `borderRadius` and an `overflow: 'hidden'`.

Does that sound like something we want to apply locally to this caller of `Touchable`, or would you rather we make a change to `Touchable` itself? I guess in the latter case we'd inspect the passed `style` prop for a `borderRadius`, and if it's there, apply it to the outer view.

chrisbobbe:
I've just pushed a revision that doesn't yet fix the ripple problem you pointed out; see my previous comment, https://github.com/zulip/zulip-mobile/pull/5130#issuecomment-977363753. 🙂

gnprice:
> From some experimentation, this appears to be because the `<Touchable />` there is wrapped in a `<View />` that has a `borderRadius` and an `overflow: 'hidden'`.

Hmm yeah, weird. So the key is that the `borderRadius` lives on the Touchable's *parent*, plus that it has `overflow: 'hidden'`.


> would you rather we make a change to `Touchable` itself? I guess in the latter case we'd inspect the passed `style` prop for a `borderRadius`, and if it's there, apply it to the outer view.

Well, the outer view that `Touchable` creates is already getting the whole `style` prop, so for our send button it's already where the `borderRadius` is going. It seems like what we need is a `borderRadius` (plus `overflow: 'hidden'`) on a parent of *that*.

Hmm, which also defeats the `hitSlop`. That's awkward. (Unless maybe if we put a `hitSlop` on the parent too?)

For the present issue, I'll be happy merging a fix that's just at this call site. This is something that ideally `Touchable` itself would solve, but fixing it there requires careful attention to all its other call sites, so I'd prefer to decouple that from the immediate issue. Once we do have a good fix for this site, though, that'll be a good time to follow up by fixing it on `Touchable` in general, while the nuances are fresh for us.

chrisbobbe:
> Well, the outer view that `Touchable` creates is already getting the whole `style` prop

Hmm, yeah, I see that's true in some cases: if there's no `onPress` and no `onLongPress` passed; else if the platform is iOS.

chrisbobbe:
> Well, the outer view that `Touchable` creates is already getting the whole `style` prop, so for our send button it's already where the `borderRadius` is going. It seems like what we need is a `borderRadius` (plus `overflow: 'hidden'`) on a parent of _that_.

Ah, yeah, I wasn't very clear—by "the outer view", I meant another outer view that we would add in `Touchable`. Like we'd add a view around callsites if we choose to have callers handle it.

> Hmm, which also defeats the `hitSlop`. That's awkward.

Hmm! I didn't think of that. There's a report that, with facebook/react-native@e35a963bf, released in RN v0.66, "the touch area due to hitslop can extend beyond the parent boundaries". See https://github.com/facebook/react-native/commit/e35a963bfb93bbbdd92f4dd74d14e2ad6df5e14a#commitcomment-60424352. If that's true, and if it's true even with `overflow: 'hidden'` (which may be doubtful, actually, looking at the motivation for the commit), then maybe we're safe, maybe by accident? I haven't looked more closely yet.

> (Unless maybe if we put a `hitSlop` on the parent too?)

This is probably the sensible workaround in any case.

gnprice:
Thanks for the revision! Sorry for the delayed review. This looks good; merging.

Issue #4920, opened 2021-07-22
iOS: Narrow sometimes lands very far in the past.


chrisbobbe:
This seems to be happening again on iOS: the initial scroll position (or maybe the scroll position a fraction of a second after opening a narrow) is far higher than it should be. It's not yet clear how this relates to #3457 or #4863, and I don't yet have a clear reproduction recipe.

For anyone wishing to help, a great first step would be to find a good reproduction recipe: a list of steps that I or anyone else can take to see the bug happening. So far it only seems to happen sometimes, and I haven't found a pattern.

chrisbobbe:
This issue only really makes sense to work on if you have an iOS device or you can run an iOS simulator.

To run the simulator, you'll need Xcode, which means you'll need an Apple computer: https://github.com/zulip/zulip-mobile/blob/main/docs/howto/ios-tips.md#running-on-ios-simulator. Do you have one of those?

thisisnitish:
Thanks for your patience @chrisbobbe!, I've gone through it. So far I observed is:
* When I click any topic, stream I see narrow goes so much above and when I click any PM then I sometimes and sometimes not.
* Basically, I've observed is, the number of times narrow lands very far in past for topics, streams >> the number of times narrow lands very far in past for PMs.
I haven't found any pattern yet.
Thanks.

PR #5132, opened 2021-11-18
msglist-diffing: Tidy up scroll-position logic a little bit

chrisbobbe:
Including something we'd like to try for #4920 (I've updated the description there to say so). Not sure yet how much it'll help, but probably some?


chrisbobbe:
(Fixed some small things flagged by eslint.)

gnprice:
Thanks, looks good -- merging.

chrisbobbe:
Thanks!

PR #5135, opened 2021-11-19
sentry: Add `error.name` and `error.message` in apiCall error breadcrumb

chrisbobbe:
Prompted by noticing that we could be more informative here, after
seeing what looks like a failed API request in a Sentry breadcrumb,
but without an httpStatus:
  https://sentry.io/organizations/zulip/issues/2799655412/events/6664a7f3171749279ff724b4e805a132/?project=191284

(The breadcrumb was first noticed by Tim while debugging an issue
that turned out to be unrelated:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/sentry.20IllegalArgumentException.3A.20Bitmap.20must.20not.20be.20null.2E/near/1283848.)


gnprice:
Good idea. Looks good -- merging.

chrisbobbe:
Thanks!

Issue #5136, opened 2021-11-20
Give notifications a distinct vibration pattern, on Android


gnprice:
This is a sort of sibling issue to #3150 -- in addition to using a distinct, recognizable sound, I'd like to use a distinct, recognizable vibration pattern. This is for basically the same reason as the OP of #3150 explained there:
> so I can tell without looking at my phone what's asking for my attention.

Chat discussion [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notification.20vibration.20pattern/near/1280242) with details.

Unlike with sounds, it looks like there isn't an API for us to be able to do this on iOS. I don't see any reference to vibration either in the APIs we'll get to use when we write client-side code to control notifications:
https://developer.apple.com/documentation/usernotifications/unmutablenotificationcontent
or in the API we're currently using, specifying the details from the server side:
https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2943363
So this is an Android-only issue without an iOS counterpart.

PR #5137, opened 2021-11-20
android notif: Set a distinct vibration pattern

gnprice:
For detailed discussion of choosing this particular pattern, see:
  https://chat.zulip.org/#narrow/stream/48-mobile/topic/notification.20vibration.20pattern/near/1284530

This also makes the first time we're discarding an old notification
channel / category and making a new one, in order to change the
settings that can't be changed on an existing channel.  So add a bit
of infrastructure to handle that: just look for channels that aren't
at the current channel ID, and delete them.

Before putting this in a prod release, I'd like to bunch it up with
setting a distinct notification sound too, to avoid repeatedly
clobbering users' settings.  But it's fine to do that for us few
alpha users, so that isn't an obstacle to merging this now.

Also delete the comments about settings being added recently and not
necessarily applying, because with the new channel ID they will now
always have been applied.

Fixes: #5136


chrisbobbe:
Thanks, LGTM! Merged.

PR #5139, opened 2021-11-22
Add organisation icon,name,uri in account picker screen.

sumj25:


chrisbobbe:
Hi, @sumj25, welcome! Please tidy up your branch using our commit style guide at https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#git-commits-commit-messages, before we do a review. Thanks!

alya:
@sumj25 could you please add screenshots (light and dark theme) to this PR? Thanks!

sumj25:
> @sumj25 could you please add screenshots (light and dark theme) to this PR? Thanks!

Yaa Sure.
Screenshot from 2022-01-21 16-13-19



Screenshot from 2022-01-21 16-14-37


Screenshot from 2022-01-21 16-15-16

Screenshot from 2022-01-21 16-16-02

alya:
Cool, looks good to me -- thanks!

src/account/AccountItem.js
```diff
@@ -59,23 +66,55 @@ export default function AccountItem(props: Props): Node {
...
+      } catch (e) {
+        showErrorAlert(_('Failed to connect to server: {realm}', { realm: realm.toString() }));
+      }
+    }
+    getAndSetIsAvailable();
+  }, [realm, _]);
+
+  if (!orgDetail.realm_icon || !orgDetail.realm_name || !orgDetail.realm_uri) {
+    return <LoadingIndicator size={40} />;
+  }
```
gnprice, line 91: No, don't do this. Remember that people aren't always on a fast Internet connection, or indeed any connection. It's very frustrating to be trying to navigate around an app, looking at information it should already have -- like the list of accounts you're logged into -- and have it sit there hanging while it tries to talk to a server.

If the only way to implement this feature of showing the names and icons were to do this, it would not be worth it because on net it would make the experience worse.

---

Fortunately, we have a better way to do this sort of thing. Take a look at the `Account` type in src/types.js, and the properties `zulipVersion` and `zulipFeatureLevel` there. Each of those is data from the server, which we keep around for all of the user's accounts. If you search the codebase for `zulipFeatureLevel` and look at the code around the results, particularly in accountsReducer.js and migrations.js, you'll see how we maintain that data.

We can store the `realm_icon` and `realm_name` information for each account in the same way as we store its `zulip_feature_level`. Then we'll have that data immediately when the user looks at this screen, without having to wait for a network request.

When we do, there will still sometimes be accounts that we don't have the data for -- in particular, when the user first upgrades to a version of the app that has this feature, that will be true at first for any existing accounts they're logged into, until the next time they use that account. This screen needs to handle that case gracefully -- in particular, it should not make the user wait in that case either.

A good-enough solution would be that we just leave the icon and the name blank, and show the URL and email like we currently do.

gnprice (commented):
Thanks @sumj25!

Trying this out, the first thing I notice is that there's a very long lag when I hit the "Switch account" button to go to this screen -- the app just hangs, without yet going anywhere, still showing the old screen. This goes on for about one full second when I try it on the Android emulator on my desktop.

Did you observe the same thing happening? When there's a serious UX issue like that, it's important to highlight it in your PR. We may be able to help debug, or sometimes we may decide the trade-off is worth it; but we definitely don't want to merge something like that without knowing it.

From looking at the code, one major comment below. This approach will make a bad user experience (on top of that lag before we even reach this screen.) Fortunately there is a better way to do it.

gnprice:
> Trying this out, the first thing I notice is that there's a very long lag when I hit the "Switch account" button to go to this screen -- the app just hangs, without yet going anywhere, still showing the old screen. This goes on for about one full second when I try it on the Android emulator on my desktop.

(BTW I've debugged this and sent a fix: #5226.)

PR #5140, opened 2021-11-23
notif settings screen: Always show server-push-setup-banner if it applies

chrisbobbe:
In addition to the dismissable banner on the home screen that we started showing in #5090.

Related: #1507

### DIFF
--- a/src/common/ServerCompatBanner.js
+++ b/src/common/ServerCompatBanner.js
@@ -62,14 +62,14 @@ export default function ServerCompatBanner(props: Props): Node {
       buttons={[
         {
           id: 'dismiss',
-          label: 'Dismiss',
+          label: 'Remind me later',
           onPress: () => {
             dispatch(dismissCompatNotice());
           },
         },
         {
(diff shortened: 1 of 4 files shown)


src/common/ServerCompatBanner.js
```diff
@@ -62,14 +62,14 @@ export default function ServerCompatBanner(props: Props): Node {
...
           id: 'dismiss',
-          label: 'Dismiss',
+          label: 'Remind me later',
           onPress: () => {
             dispatch(dismissCompatNotice());
           },
         },
         {
-          id: 'resolve',
-          label: isAdmin ? 'Fix now' : 'Learn more',
```
gnprice, line 72: I think these were the only two uses of these two strings, so they can be deleted from `messages_en.json`.

gnprice (commented):
Thanks @chrisbobbe! Looks good, with a small comment below; then please merge at will.

chrisbobbe:

PR #5144, opened 2021-12-02
checkout-keystore: Switch to Sequoia PGP's `sq`, from GPG

gnprice:


chrisbobbe:
Thanks, LGTM! Merged.

andersk:
FYI this can be done with less fancy Bash:

```bash
sq decrypt --recipient-key "${recipient_key}" \
  -o "${cleartext}" "${cryptotext}" \
  2> >(grep -vE '^(Compressed|Encrypted) ' >&2)
```

gnprice:
> less fancy Bash:

Neat, thanks! Though hmm, I'm not sure `>(…)` is available on macOS (even with a modern Bash.)


> or even mildly obfuscated POSIX shell (modulo `pipefail` concerns):

Hmm yeah, that might be better too, along with `set -o pipefail`.

I tried something more like that (with `exec 3>&1` because I didn't expect that `{ … } 3>&1`, with redirection on mere curly braces, would work), but in addition to the `pipefail` issue, I was using `|& grep …` and didn't find a solution that left `sq`'s stdout going directly to the original stdout. That'd be fine because the only output we expect is the password prompt… but when `sq decrypt` sees its stdout is a pipe (or, not a terminal?), it prints a warning about how its output isn't in a stable format for scripting.

gnprice:
Just went and pulled this off my to-do list: #5191

Thanks again @andersk for the suggestions!

Issue #5145, opened 2021-12-02
android notif: Replace our `NotificationIntentService` with directly opening MainActivity


chrisbobbe:
> Well, here's one blocking issue [for #5101]:
>   https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines
> We'll need to replace our `NotificationIntentService`, which we use when you tap a notification to open it, with directly opening an Activity. (We learned about this from a lint warning in Android Studio: https://github.com/zulip/zulip-mobile/pull/5121#discussion_r747107709)
> * Ideally this would be our `MainActivity`. We'll need to play with things somewhat to try to get the correct behavior in all three cases of the app being in the foreground, in the background, and not running. (Currently `NotificationIntentService` has `NotifyReact.kt` do this work.)
>   * The main advice from Android upstream won't work for us verbatim because it assumes the canonical Android model where basically every screen in your app is a distinct activity. That structure isn't even the unanimous advice of Android upstream these days -- some of their voices say you should have one or a few activities and use `Fragment`s for navigation instead. But in any case it's not our structure of things using RN, and we have just one activity where the app's whole UI (apart from the notifications) lives. Hence the need to play with things to find a solution.
> * If we can't make that work, it should certainly be possible to make a trampoline activity, analogous to our `ShareToZulipActivity`. That sure won't be in the spirit of what they're going for here, though. It may also cause visible jank in the UI, like an extra transition.

_Originally posted by @gnprice in https://github.com/zulip/zulip-mobile/issues/5101#issuecomment-966039530_

chrisbobbe:
I used a handy "reference in new issue" button in the GitHub UI to make this issue, and tweaked the output a bit:

![image](https://user-images.githubusercontent.com/22248748/144511868-c1e8cda0-64ea-4c6c-b273-e160d142ef85.png)

PR #5146, opened 2021-12-02
android notif: Replace our NotificationIntentService with directly opening MainActivity

chrisbobbe:
Fixes: #5145


android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -11,6 +11,25 @@ import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
 import com.zulipmobile.notifications.*
 import com.zulipmobile.sharing.maybeHandleShareIntent
 
+/* Returns true just if we did handle the intent. */
+fun maybeHandleOpenNotificationIntent(intent: Intent): Boolean {
```
chrisbobbe: I think we probably want this to live somewhere else, and probably have a better name? I haven't settled on anything definite for either yet though.

android/app/src/main/java/com/zulipmobile/notifications/NotifyReact.kt
```diff
@@ -93,11 +93,7 @@ internal fun notifyReact(application: ReactApplication, data: Bundle) {
             // initialNotification again, but it will see a notificationOpened event.
             emit(reactContext, "notificationOpened", Arguments.fromBundle(data))
     }
-    when (appStatus) {
-        null, ReactAppStatus.NOT_RUNNING, ReactAppStatus.BACKGROUND ->
-            launchMainActivity(application as Context)
-        ReactAppStatus.FOREGROUND -> Unit
```
gnprice, line 99: I didn't find any reasons given for the conditional; possibly it's
    an unquestioned relic of the Wix library we used to use.

Indeed the behavior goes back to the Wix library. Looks like it's basically unchanged since 9eddca888e33f6ca8f828582a927cbfe981f2bcc:

    android notif: Use standard upstream API for watching RN lifecycle.
    
    This has exactly the same effect as the 80 lines of code in the wix
    library's ReactAppLifecycleFacade.

with the logic a bit refactored in 9b0157d828b94226cca34004b5be2bb5a76afeb0 and 0feb940fb0af4bd0626200e5caba4d4b7a313a87.

I don't think having the conditional is necessarily bad, given that launching isn't going to do anything useful when already in the foreground. But with the changes upcoming in this branch, it becomes useful to do this unconditionally, so it's a good change now.

android/app/src/main/java/com/zulipmobile/notifications/NotifyReact.kt
```diff
@@ -104,7 +104,7 @@ fun emit(reactContext: ReactContext, eventName: String, data: Any?) {
 }
 
 fun launchMainActivity(context: Context) {
-    Log.d(TAG, "NotifyReact: launching main activity")
+    Log.d(TAG, "android notif: launching main activity")
```
gnprice: android notif: Remove false specificity from a log line
    
    Launching MainActivity doesn't really have anything to do with
    "notifying React", i.e., sending information about the opened
    notification to the JS environment.

Yeah, the prefix there was really to make this log line identifiably correspond to this line of code -- so it was a reference to the filename, not a description of the substance.

I think the simplest thing is just to delete this line, if you haven't had a use for it. (That's what happens later in the branch anyway.)

(If we were to keep this, I'd drop "android" from the prefix -- this is already going to an Android-specific log system.)

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -39,7 +39,7 @@ open class MainActivity : ReactActivity() {
         maybeHandleShareIntent(intent, application as ReactApplication, contentResolver)
     }
 
-    override fun onNewIntent(intent: Intent?) {
+    override fun onNewIntent(intent: Intent) {
```
gnprice: android notif [nfc]: Make MainActivity.onNewIntent's `intent` arg required
    
    ReactActivity's `.onNewIntent`, which we're overriding, requires `intent`.

I think it doesn't. Here's the declaration in `ReactActivity.java`:
```java
  @Override
  public void onNewIntent(Intent intent) {
```
That doesn't mention the possibility of the argument being null. But... in Java, all object types are nullable. So Java `Intent` translates to Kotlin `Intent?`.

Sometimes there's a Java annotation `@NotNull` indicating that a given value isn't supposed to be null. But there doesn't seem to be one of those here. So if a caller passes null, no linter would complain either.

If you look in the Kotlin version of the SDK docs:
https://developer.android.com/reference/kotlin/android/app/Activity#onnewintent
this is reflected in the notation `Intent!`:
> `protected open fun onNewIntent(intent: Intent!): Unit`

The meaning of that `!` notation is defined here:
https://kotlinlang.org/docs/java-interop.html#notation-for-platform-types
> `T!` means "`T` or `T?`"

Practically, the way I ended up with this version of the definition is probably that I started typing "onNewIntent" and let Android Studio autocomplete an empty override for me.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -11,6 +11,25 @@ import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
...
+fun maybeHandleOpenNotificationIntent(intent: Intent): Boolean {
+    // We handle intents to launch MainActivity on tapping notifications
+    if (
+        intent.action == Intent.ACTION_VIEW
+        // The web-auth intent's action is also VIEW; leave that for RN
+        //   to handle.
+        && intent.hasExtra(EXTRA_NOTIFICATION_DATA)
+    ) {
+        val data = intent.getBundleExtra(EXTRA_NOTIFICATION_DATA)
+        logNotificationData("notif opened", data!!)
```
gnprice, line 24: android notif: Move a log line into MainActivity's lifecycle methods

This commit is kind of confusing because it's rather a lot more code than just moving a log line sounds like it should be:
 2 files changed, 26 insertions(+), 4 deletions(-)

The added plumbing makes sense because of the real work we have it do in the next commit. I think the clearest thing is probably to squash that one into this. The result is only slightly bigger than this already is:
 2 files changed, 26 insertions(+), 6 deletions(-)

and it's easier to understand as changing where we do the real work, with the log line coming along for the ride.

android/app/src/main/java/com/zulipmobile/notifications/NotificationIntentService.kt
```diff
@@ -1,17 +1,36 @@
...
+import android.util.Log
+import com.zulipmobile.MainActivity
 
 class NotificationIntentService : IntentService("NotificationIntentService") {
     override fun onHandleIntent(intent: Intent?) {
         val context = applicationContext as? MainApplication ?: return
         if (Intent.ACTION_VIEW == intent!!.action) {
             val data = intent.getBundleExtra(EXTRA_NOTIFICATION_DATA)
-            onOpened(context, data!!)
+            Log.d(TAG, "android notif: launching main activity")
```
gnprice: This commit and the next move this code around twice:

6fd1a19d2 android notif [nfc]: Inline NotificationUiManager.onOpened
 2 files changed, 23 insertions(+), 26 deletions(-)
2521e1df5 android notif: Remove NotificationIntentService
 2 files changed, 21 insertions(+), 50 deletions(-)

If squashed, the resulting diff is just slightly bigger than one of those:
 2 files changed, 21 insertions(+), 53 deletions(-)

That also keeps the move within one file, rather than moving to a different file and back. So I think it's clearest to squash those two together.

With that, the preceding commit:
3cf35647c android notif [nfc]: Rename variables to ease next refactor
 2 files changed, 6 insertions(+), 6 deletions(-)

also becomes unnecessary.

gnprice (commented):
Thanks! Glad to have this getting cleaned up. Comments below.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
gnprice: Yeah. Looking at this function and now looking at the share-to version beside it, I think what I really want is to have those two bits of code next to each other -- and probably cleanest as a single `when (intent.action)`, with sub-conditions as needed (like looking for `EXTRA_NOTIFICATION_DATA`, though actually the cleanest thing there -- which can be a followup -- is probably to parse the outermost parts of the URL.)

IOW rather than align the name of `maybeHandleIntent` to reflect where it lives, align where it lives to reflect its name: move it to be a private method on `MainActivity`, and then later in the branch add the open-notification case to it.

Part of why I'd like them to be next to each other is that they're carving out two different swaths of the same space of possibilities -- if we were to have any two things like this that could overlap in what they covered, then the order we called them in would start mattering, and it'd probably be a bug. So it's helpful to have them next to each other and ideally in the same `when` statement, so that it's easy to see they're mutually exclusive and keep them that way.

gnprice:
(Also just replied to the initial thread above.)

chrisbobbe:

android/app/src/main/java/com/zulipmobile/sharing/SharingHelper.kt
```diff
@@ -7,48 +7,25 @@ import android.content.ContentResolver
...
 import com.zulipmobile.notifications.ReactAppStatus
 import com.zulipmobile.notifications.appStatus
 import com.zulipmobile.notifications.emit
-import com.zulipmobile.notifications.tryGetReactInstanceManager
 
 @JvmField
 val TAG = "ShareToZulip"
 
-/* Returns true just if we did handle the intent. */
-fun maybeHandleIntent(
```
gnprice, line 23: android notif [nfc]: Make maybeHandleIntent a helper method on MainActivity

nit: s/ notif//

Similarly for the next commit:

    android notif [nfc]: Simplify interface of MainActivity's maybeHandleIntent

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -28,6 +29,45 @@ open class MainActivity : ReactActivity() {
...
+            // Launch MainActivity on tapping a notification
+            Intent.ACTION_VIEW -> {
+                // The web-auth intent's action is also VIEW; leave that for RN
+                //   to handle.
+                if (!intent.hasExtra(EXTRA_NOTIFICATION_DATA)) {
+                    return false
+                }
+
+                val data = intent.getBundleExtra(EXTRA_NOTIFICATION_DATA)
+                logNotificationData("notif opened", data!!)
```
gnprice, line 60: This can be cleaned up slightly by putting the `getBundleExtra` first, and then checking if it's null instead of calling `hasExtra`. Then the `!!` becomes unnecessary.

That also means that if somehow we get an intent with an `EXTRA_NOTIFICATION_DATA` extra that isn't a Bundle, then we don't crash at the `!!` and instead just ignore it here as unhandled.

android/app/src/main/java/com/zulipmobile/notifications/NotifyReact.kt
```diff
@@ -76,11 +72,7 @@ val ReactContext.appStatus: ReactAppStatus
         }
     }
 
-internal fun notifyReact(application: ReactApplication, data: Bundle) {
-    // TODO deduplicate this with handleSend in SharingHelper.kt.
-    // Until then, keep in sync when changing.
```
gnprice, line 81: We could deduplicate all the way down to `appStatus`, except

Yeah. There's still quite a bit of trickiness here that I'd like to deduplicate: sometimes we set a static/global variable, and sometimes we emit an event, and we need to pick the right one or we're likely to drop or duplicate an event. My ideal version of this and `handleSend` is that they'd each say: here's a `WritableMap`, and here's a little value identifying where the event should go, and then all that logic like the `when (appStatus)` would be in a common function they both invoke.

This change is a helpful step! But let's keep these TODO comments here.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -28,6 +29,45 @@ open class MainActivity : ReactActivity() {
...
+    private fun maybeHandleIntent(
+        intent: Intent?,
+        contentResolver: ContentResolver
+    ): Boolean {
+        if (intent == null) {
+            return false
+        }
+
+
+        val host = (application as ReactApplication).reactNativeHost
```
gnprice, line 42: nit: just one blank line

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -28,6 +29,45 @@ open class MainActivity : ReactActivity() {
         }
     }
 
+    /* Returns true just if we did handle the intent. */
+    private fun maybeHandleIntent(
+        intent: Intent?,
+        contentResolver: ContentResolver
```
gnprice, line 35: This parameter can be simplified out in much the same way as `application: ReactApplication`: the call sites just get it from a getter on `this`, so this function can do the same thing.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
```diff
@@ -28,6 +29,45 @@ open class MainActivity : ReactActivity() {
...
+    ): Boolean {
+        if (intent == null) {
+            return false
+        }
+
+
+        val host = (application as ReactApplication).reactNativeHost
+        val reactContext = host.tryGetReactInstanceManager()?.currentReactContext
```
gnprice, line 43: Hmm, actually I don't love having these lines run here, above the `when (intent.action)`.

The thing is that that means they run even if they're not needed -- if the intent is something else. They look like they could be cheap and side-effect-free… or could not be. I think there isn't any live case where it'd matter even if they are -- the only intents we handle other than the two cases below are the web-auth intents, where we'll want a React instance/context anyway -- but I'd rather not have to worry about that possibility coming up in the future.

Maybe simplest is I'll just add a commit at the end to address this.

gnprice (commented):
Thanks for the revision! Looks good -- just small comments below.

android/app/src/main/java/com/zulipmobile/MainActivity.kt
chrisbobbe: > Maybe simplest is I'll just add a commit at the end to address this.

OK, SGTM, thanks!

chrisbobbe:
Thanks! Revision pushed.

chrisbobbe:
> Thanks! Revision pushed.

Ah, is there a followup from https://github.com/zulip/zulip-mobile/pull/5146#discussion_r762350660 :

> actually the cleanest thing there -- which can be a followup -- is probably to parse the outermost parts of the URL.

I'm not really sure what you mean there.

gnprice:
Thanks! Looks good -- merging. I made one tweak to cut the `SuppressLint` import in the commit where it stops being used (because Android Studio did that automatically after I made some other edits there.)

I have a draft for that followup and this one: https://github.com/zulip/zulip-mobile/pull/5146#discussion_r764447236, and I'll send those as a separate PR because they're nontrivial refactors.

chrisbobbe:
Great, thanks very much!

gnprice:
#5160

Issue #5143, opened 2021-12-02
Unread-PMs count is below center and can get clipped


gnprice:
From a report [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/unread.20count.20is.20not.20visible.20properly.20in.20overlay.2E/near/1286941):
![image](https://user-images.githubusercontent.com/28173/144490956-f0549274-530c-4d70-8c84-5513e6933524.png)

This would be good to fix, because it's pretty broken-looking and can be pretty conspicuous depending on your usage pattern.

PR #5147, opened 2021-12-03
Fix clipped unread-PMs count

chrisbobbe:
Some screenshots after these changes (for what it looked like before, see #5143):

Screenshot_20211202-190602_Zulip-debug
Screenshot_20211202-190618_Zulip-debug

Fixes: #5143


gnprice:
Neat! Glad this could be resolved in a way that reduces the amount of code. Looks good -- merging.

chrisbobbe:
Thanks!

PR #5153, opened 2021-12-06
ComposeMenu: Change a `logging.info` to `logging.warn`, to send to Sentry

chrisbobbe:
Checking the jsdoc on logging.info, it explicitly says it's not
meant for sending things to Sentry. So, use logging.warn, which is.


gnprice:
Thanks; merged!

PR #5155, opened 2021-12-07
action types: Sync update-message event type with doc

chrisbobbe:
This should help with #4840. Inheriting P1 from that issue.

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -341,19 +341,42 @@ type EventMessageDeleteAction = {|
   type: typeof EVENT_MESSAGE_DELETE,
   messageIds: number[],
 |};
+
+// This is current to feature level 109:
+//   https://zulip.com/api/get-events#update_message
 type EventUpdateMessageAction = {|
   ...ServerEvent,
   type: typeof EVENT_UPDATE_MESSAGE,
   user_id: UserId,
+
+  // Any content changes apply to just message_id.
   message_id: number,
+
+  // Any stream/topic changes apply to all of message_ids, which is
+  //   guaranteed to include message_id.
+  message_ids: $ReadOnlyArray<number>,
+
   edit_timestamp: number,
+  stream_name?: string,
+  stream_id?: number,
+  new_stream_id?: number,
+  propagate_mode: 'change_one' | 'change_later' | 'change_all',
   orig_subject?: string,
   subject: string,
-  subject_links: string[],
+
+  // TODO(server-4.0): Changed in feat. 46 to array-of-objects shape, from string[]
+  topic_links?: $ReadOnlyArray<{| +text: string, +url: string |}> | $ReadOnlyArray<string>,
+
+  // TODO(server-3.0): Replaced in feat. 1 by topic_links
+  subject_links?: string[],
+
   orig_content: string,
   orig_rendered_content: string,
   prev_rendered_content_version: number,
+  content: string,
   rendered_content: string,
+  is_me_message: boolean,
+  flags: $ReadOnlyArray<string>,
 |};
 
 type EventReactionCommon = {|
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -346,24 +346,8 @@ type EventUpdateMessageAction = {|
   type: typeof EVENT_UPDATE_MESSAGE,
   edit_timestamp: number,
   message_id: number,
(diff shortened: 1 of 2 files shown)


chrisbobbe:
See two currently unresolved questions at https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1292941 .

src/actionTypes.js
```diff
@@ -341,35 +341,37 @@ type EventMessageDeleteAction = {|
...
 type EventUpdateMessageAction = {|
   ...ServerEvent,
   type: typeof EVENT_UPDATE_MESSAGE,
-  edit_timestamp: number,
+  user_id: UserId,
   message_id: number,
-  // TODO is it really right that just one of the orig_* is optional?
-  orig_content: string,
-
-  // TODO: The doc for this field isn't yet correct; it turns out that
```
gnprice, line 352: The second TODO is resolved with Tim's clarifications in in
    8295bb918.

Let's clarify what repo that commit ID is in, like `zulip/zulip@8295bb918`.

src/actionTypes.js
```diff
@@ -341,35 +341,37 @@ type EventMessageDeleteAction = {|
   type: typeof EVENT_MESSAGE_DELETE,
   messageIds: number[],
 |};
+type TopicLinks = $ReadOnlyArray<{| +text: string, +url: string |}> | $ReadOnlyArray<string>;
+// This is current to feature level 109.
 type EventUpdateMessageAction = {|
```
gnprice, line 346: This is probably a good occasion to add a link to the API doc. (Which didn't exist when this type was first written down.)

gnprice, line 346: nit: blank line between these definitions

(Not sure why there wasn't one between the existing two; seems like an oversight.)

src/actionTypes.js
```diff
@@ -341,35 +341,37 @@ type EventMessageDeleteAction = {|
...
   orig_subject?: string,
+  subject: string,
 
+  // TODO(server-3.0): Added in feat. 46, replacing subject_links
+  // TODO(server-4.0): Changed in feat. 1 to array-of-objects shape, from string[]
+  topic_links?: TopicLinks,
+
+  // TODO(server-3.0): Replaced in feat. 46 by topic_links
+  // TODO(server-4.0): Changed in feat. 1 to array-of-objects shape, from string[]
+  subject_links?: TopicLinks,
```
gnprice, line 366: I think the feature levels here are swapped.

Also, as I read the doc, `subject_links` stops existing after the rename; it's not simultaneously present e.g. for compatibility. So when present, it's always the old form.

src/actionTypes.js
```diff
@@ -341,35 +341,37 @@ type EventMessageDeleteAction = {|
...
+  message_ids: $ReadOnlyArray<number>,
+  edit_timestamp: number,
+  stream_name?: string,
+  stream_id?: number,
+  new_stream_id?: number,
+  propagate_mode: 'change_one' | 'change_later' | 'change_all',
   orig_subject?: string,
+  subject: string,
 
+  // TODO(server-3.0): Added in feat. 46, replacing subject_links
```
gnprice, line 360: I think this TODO can be dropped, because there's not actually anything to do at this spot when dropping support for these old servers: the field is optional even in the latest version.

gnprice (commented):
Thanks for taking care of this! Small comments below.

gnprice:
Oh, and on the partial version of this that's in #4980 I see there's one other comment that's relevant here: https://github.com/zulip/zulip-mobile/pull/4980#discussion_r700592607

(Plus https://github.com/zulip/zulip-mobile/pull/4980#discussion_r700590443 , which I covered above but more tersely.)

chrisbobbe:
Thanks for the review, revision pushed!

Please note this:

> See two currently unresolved questions at https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60update_message.60.20event/near/1292941 .

chrisbobbe:
Confirmed in chat that we should expect `message_ids` to always be present; I intend to include that in my docs PR to resolve those two questions.

gnprice:
Thanks! This looks good -- merging.

Those two unresolved questions look to not affect these types, because they're both about the conditions under which a particular optional field is or isn't present; the fields are optional regardless.

Issue #5141, opened 2021-11-24
Make "+" open an action sheet instead of row of buttons


alya:
At present, compose box buttons are accessed via a "plus" symbol to the left of the compose box. Clicking on "plus" expands the buttons on the same line as the compose box. This works OK with the buttons we currently have, but is likely to no longer look good as we start adding more buttons (#3427, #5079).

In #5080, we change this behavior so that putting the cursor into the compose box will open buttons *below* the compose box.

Based on the discussion on #5080 and in [this CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M5080.20Move.20compose.20box.20buttons.20into.20a.20row.20below.20the.20compose), we should also make "+" open an action sheet.

Actions (we can discuss/finalize these labels when working on the issue):
* Upload a file (cf. webapp tooltip: Upload files)
* Upload a picture
* Take a picture
* Add a video call (cf. webapp tooltip: Add video call)

This issue may be a blocker for #3427 and #5079. It's possible we can implement one of those without this issue, but not both.

sumj25:
I'm working on this issue and would like to claim it.

chrisbobbe:
As discussed in our meeting on 2022-10-05, I'm marking this as blocked on @alya considering this with the mobile redesign in mind.

It's a priority especially because it's in our plan toward fixing #M5463, a really annoying bug; see https://github.com/zulip/zulip-mobile/issues/5463#issuecomment-1227641703.

chrisbobbe:
Closing, as we took a different design in #5080, with PR #5635.

Issue #5079, opened 2021-10-26
Make it possible to insert local time


alya:
The [local times widget](https://zulip.com/help/format-your-message-using-markdown#mention-a-time) is very useful, but can be difficult to discover. On mobile, it is *very* difficult to insert a new local time, as it requires manually typing the appropriate syntax, which is not designed to be easy to enter by hand.

 To address these challenges, we should add a compose box button to insert a local time.

* Icon: https://fontawesome.com/v4.7/icon/clock-o
* Location: As the rightmost icon in the "plus" compose box menu.
* Action: Open a calendar widget for inserting a timezone-sensitive time, and generate the appropriate local time syntax into the message based on the selection.

The time should be inserted in the position of the cursor.

Typing `<time` directly into the compose box should also pop up a calendar widget to select a local time. This is the current way to access the time widget in the web app.

Corresponding web app issue: https://github.com/zulip/zulip/issues/20045

[Discussion thread on CZO](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/user.20feedback/near/1270607)

MrMischievousX:
Hello, i would like to work on this issue if its not already taken. I have did some of it please check it
![Simulator Screen Shot - iPhone 12 - 2021-11-23 at 00 22 26](https://user-images.githubusercontent.com/721983
![Simulator Screen Shot - iPhone 12 - 2021-11-23 at 00 22 33](https://user-images.githubusercontent.com/72198360/142918971-3449369d-39ef-4150-872d-93bc92e2794d.png)
60/142918961-b6900a7e-62c5-42e4-8c69-96e4049591c4.png)
.

alya:
@MrMischievousX please play around with the `<time` widget in the Zulip webapp to better understand the required functionality.

MrMischievousX:
Something similar to webap

https://user-images.githubusercontent.com/72198360/142973456-7ab73aeb-a49a-4398-8491-26f0bb286101.mov

p

alya:
Cool!

We generally follow Material Design for the design of our app, so I think we'll want a different look for the time picker that's consistent with those guidelines.

https://material.io/components/time-pickers
https://material.io/components/date-pickers

When you have the next iteration or if you have questions, you start a thread in [#mobile-team on chat.zulip.org](https://chat.zulip.org/#narrow/stream/243-mobile-team) to discuss.

Issue #3427, opened 2019-03-25
Inefficient and illogical "mentions" when composing from mobile app


freakystylie:
We are on a three month trial implementation of Zulip, and have been receiving feedback from our field users who interact almost entirely on the mobile app. Part of our internal communication protocol is to @mention anyone who _needs_ to see the message.

On the mobile app, the process to mention is cumbersome, because it always involves going to another keyboard screen to get the "@". This might be fine for one mention, but add a few more and it quickly becomes annoying.

Furthermore, the mobile app provides the ability to press the "+" button next to the compose window and then select the people icon to mention people. However, this starts a private message, it does not simply mention the selected people in the message.

For example: reply to a topic in a stream. Start typing, then press "+", then press the people icon. Rather than mentioning the people in the topic/stream you have been typing in, a new private message is started, and your draft message to the stream/topic is abandoned. This is illogical and unexpected behaviour, because if you press the "+" and then either the photo icon or the pictures icon, the resulting photo/picture is uploaded to the stream/topic.  

@mentioning people is a heavy part of our workflow as it is the only way to cut through the noise. It would be waaaay better if @mentioning was more readily accessible when using the mobile app. Our in-office users LOVE Zulip, but our mobile app users struggle, and this cumbersome @mentioning is inhibiting communication protocol conformance, and therefore adoption.

Thanks, and keep up the good work.

shidarin:
Slack does this by putting an `@` button on the compose section, it's a good solution.

PR #5159, opened 2021-12-08
Open an action sheet instead of row of buttons

sumj25:
At present, compose box buttons are accessed via a "plus" symbol to the left of the compose box. Clicking on "plus" expands the buttons on the same line as the compose box. This works OK with the buttons we currently have but is likely to no longer look good as we start adding more buttons ([#3427](https://github.com/zulip/zulip-mobile/issues/3427), [#5079](https://github.com/zulip/zulip-mobile/issues/5079)). So that's why we have to open an action sheet instead of a row of buttons.
see #5141. For more details see [CZO thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M5080.20Move.20compose.20box.20buttons.20into.20a.20row.20below.20the.20compose).


src/compose/ComposeMenu.js
```diff
@@ -4,24 +4,17 @@ import type { ComponentType } from 'react';
 import { Platform, View, Alert, Linking } from 'react-native';
 import type { DocumentPickerResponse } from 'react-native-document-picker';
 import { launchCamera, launchImageLibrary } from 'react-native-image-picker';
-
```
chrisbobbe, line 7: Please keep the blank line here; it separates third-party imports from imports of our own code.

src/compose/ComposeMenu.js
```diff
@@ -35,7 +28,7 @@ type SelectorProps = $ReadOnly<{||}>;
 
 type Props = $ReadOnly<{|
   ...OuterProps,
-
```
chrisbobbe, line 38: Please keep this blank line, for separation, and add a new one below `showActionSheetWithOptions`.

src/compose/ComposeMenu.js
```diff
@@ -234,56 +227,58 @@ class ComposeMenuInner extends PureComponent<Props> {
...
 
   render() {
-    const { expanded, insertVideoCallLink, onExpandContract } = this.props;
-    const numIcons =
-      2 + (Platform.OS === 'android' ? 1 : 0) + (insertVideoCallLink !== null ? 1 : 0);
+    const { expanded, insertVideoCallLink } = this.props;
+
+    const onOpenActionSheet = () => {
+      const options = [
+        'Upload a file',
```
chrisbobbe, line 234: Please set up these new user-facing strings for translation. They should get entries in static/translations/messages_en.json. Please see https://github.com/zulip/zulip-mobile/blob/main/docs/howto/translations.md and search for `const _ = this.context;` in the codebase.

src/compose/ComposeMenu.js
```diff
@@ -234,56 +227,58 @@ class ComposeMenuInner extends PureComponent<Props> {
...
-        <AnimatedComponent
-          stylePropertyName="width"
-          fullValue={40 * numIcons}
-          useNativeDriver={false}
-          visible={expanded}
-        >
-          <View style={this.styles.composeMenu}>
-            {Platform.OS === 'android' && (
-              <IconAttach
-                style={this.styles.composeMenuButton}
```
chrisbobbe, line 252: `this.styles.composeMenuButton` is no longer used; let's remove its definition.

src/compose/ComposeMenu.js
```diff
@@ -234,56 +227,58 @@ class ComposeMenuInner extends PureComponent<Props> {
...
+      ];
+      const cancelButtonIndex = options.length - 1;
+
+      this.props.showActionSheetWithOptions(
+        {
+          options,
+          cancelButtonIndex,
+        },
+        buttonIndex => {
+          if (buttonIndex === 0) {
```
chrisbobbe, line 248: These repeated `if` statements look pretty fragile.Can you instead make an array and use `buttonIndex`

chrisbobbe, line 248: These repeated `if` statements look pretty fragile.

Instead, let's define an array (maybe call it `buttons`) that we use for two things at once:
- to compute `options`
- to give the right behavior based on the given `buttonIndex`, using that index to access an item in `buttons`

It could be an array of objects like `{ option, callback }`, or something more like we do in src/action-sheets/index.js.

src/compose/ComposeMenu.js
chrisbobbe, line 234: Please set up these new user-facing strings for translation. They should get entries in static/translations/messages_en.json. Please see https://github.com/zulip/zulip-mobile/blob/main/docs/howto/translations.md and search for `const _ = this.context;` in the codebase.

src/compose/ComposeMenu.js
```diff
@@ -234,56 +227,58 @@ class ComposeMenuInner extends PureComponent<Props> {
...
+          if (buttonIndex === 2) {
+            this.handleCameraCapture();
+          }
+          if (buttonIndex === 3) {
+            // Flow actually has a special case for this, this is called an "unknown function".
+            // We know that insertVideoCallLink is a function, but we don't know anything about its
+            // arguments or return type so we can't safely do anything with it except pass it around.
+            // see:https://stackoverflow.com/questions/65314559/flow-type-refine-mixed-to-function
+            if (typeof insertVideoCallLink === 'function') {
+              insertVideoCallLink();
```
chrisbobbe, line 263: > we don't know anything about its arguments or return type so we can't safely do anything with it except pass it around

That's not true 🙂—we're not just passing this function around. We're calling it, and Flow doesn't complain. We're allowed to do so because (unlike the case in that SO thread) Flow knows specifically what the function type will be:

```js
type OuterProps = $ReadOnly<{|
  // …
  insertVideoCallLink: (() => void) | null,
  // …
|}>;
```

i.e., it takes no arguments and it returns `undefined`.

In any case, there's a bug here: we shouldn't offer the action-sheet button in the first place if nothing happens when you press it.

chrisbobbe (changes requested):
Thanks! Here's a partial review. For some reason some of my comments have already appeared, above, I think; they were meant to be part of this review.

Please also write a commit message according to our [style](https://github.com/zulip/zulip-mobile/blob/master/docs/style.md#git-commits-commit-messages), as always.

sumj25:
Hii @chrisbobbe it's ready for another review.

sumj25:
Hey @chrisbobbe I hope you are in the best of your health. I request you to review my PR.

chrisbobbe:
@sumj25, before this PR is reviewable, you'll need to make the commits clear and coherent according to our [style guide](https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#git-commits-commit-messages).

sumj25:
> @sumj25, before this PR is reviewable, you'll need to make the commits clear and coherent according to our [style guide](https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#git-commits-commit-messages).

@chrisbobbe I also have to resolve conflicts?

chrisbobbe:
> @chrisbobbe I also have to resolve conflicts?

Ah, yes, I do see that GitHub is currently reporting one conflicting file:

android/app/src/main/java/com/zulipmobile/generated/BasePackageList.java

Your PR isn't adding any dependencies, so there shouldn't be a need to touch that file. You should resolve the conflict to accept the version of that file in `main`, without any changes.

Issue #5022, opened 2021-09-24
Event-queue polling loop doesn't stop itself on logout


gnprice:
This is perhaps a special case of #5009; it's a salient one because this is a fetch that we're constantly keeping one of open.

The long-poll loop in `startEventPolling` has some logic to look just after a request comes back and try to check if the user has switched accounts, and if so to bail out because the data we're now keeping around will be for a different account than the data we just got back from the old fetch.
```js
      const response = await api.pollForEvents(auth, queueId, lastEventId);
      events = response.events;

      // User switched accounts or logged out
      if (queueId !== getState().session.eventQueueId) {
```

But this logic doesn't seem quite right. It's looking for `state.session.eventQueueId` to change... but looking in `sessionReducer`, it seems like that will only get changed on a `REALM_INIT`, i.e. when you switch accounts and then a new initial fetch completes.

That's good enough that it means we won't mix the old account's data (from this event-queue polling) into the data for the new account, because `REALM_INIT` marks the start of really having server data from the new account around. But:
* Ideally we should stop this as soon as the active account changes, i.e. as soon as we change which account all our server data is *meant* to be about.
* If the user logs out, rather than switches accounts, this check won't stop anything.

Currently one result this causes can be that we try to process the resulting event(s) and promptly get an exception, from the check at the top of `eventToAction`:
```js
// src/events/eventToAction.js
export default (state: GlobalState, event: $FlowFixMe): EventAction | null => {
  if (!tryGetActiveAccount(state)) {
    logging.error('eventToAction: no active account');
  }
```
(This happens if the user not only logs out, but then removes their only account from the list, so that they have no accounts and therefore no active account.) And in fact we [see that exception in Sentry](https://sentry.io/organizations/zulip/issues/2444407871/events/38f45b4b113f49d18ce138806b8ff7f3/), at a low frequency. I *think* the result is that nothing happens and the user never notices, not a crash, but I'm not sure.

A little while ago @chrisbobbe looked at some changes that would update `state.session.eventQueueId` in more circumstances. That might be enough to fix this issue.

Issue #5009, opened 2021-09-18
Discard fetches when the active account has changed


gnprice:
When we go and fetch some data from the server, we should be sure to associate the resulting data with the right account, and not with some other of the user's accounts. In particular this should be the case even if the fetch took a while, and if in the interim the user went and switched to look at a different account (the new ["active account"](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#active-account).)

We'll need to solve this as part of #5005, where we'll potentially have several fetches going concurrently for several different accounts, and each will need to know which account the results should get stored under.

But pending that, in our current world where we have server data for just the one active account at a time, it'd be enough to discard the fetch's result (and ideally cancel the fetch itself, #4170) if the active account has changed by then.

Currently we do this in the event-queue long-polling loop in `startEventPolling`, which is the most important case, but not most other places -- notably, not in `doInitialFetch`, or the various message-fetching actions in `fetchActions.js`.

Issues #4170 and #4659 cover parts of how we might do this.

Issue #3791 is basically an example symptom of this.

gnprice:
We got a report today of what is probably this issue:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/local.20cross-realm.20leakage.20issue/near/1444848

chrisbobbe:
Any objection to changing the title from

> Discard fetches when the active account has changed

to

> Discard fetches when the active, logged-in account has changed

, since we probably want to cancel fetches on logout too?

PR #5161, opened 2021-12-08
Stop event-polling for an account when it's logged out

chrisbobbe:
And make a few other cleanups and comments.

Fixes: #5022
Related: #5009

### DIFF
--- a/src/session/sessionReducer.js
+++ b/src/session/sessionReducer.js
@@ -24,6 +24,15 @@ import { getHasAuth } from '../account/accountsSelectors';
  * See {@link SessionState} for discussion of what "non-persistent" means.
  */
 export type PerAccountSessionState = $ReadOnly<{
+  /**
+   * The event queue ID that we're currently polling on, if any.
+   *
+   * Null when we're not polling on any event queue:
+   * - Between startup and registering a queue
+   * - After the server tells us our old queue was invalid, and before we've
+   *   registered a new one
+   * - While this account is logged out
+   */
   eventQueueId: string | null,
 
   /**
@@ -162,12 +171,20 @@ export default (state: SessionState = initialState, action: Action): SessionStat
         ...state,
         needsInitialFetch: true,
         loading: false,
+
+        // The server told us that the old queue ID is invalid. Forget it,
+        // so we don't try to use it.
+        eventQueueId: null,
       };
 
     case LOGIN_SUCCESS:
       return {
         ...state,
         needsInitialFetch: true,
+
+        // We're about to request a new event queue; no use hanging on to
+        // any old one we might have.
+        eventQueueId: null,
       };
 
     case LOGOUT:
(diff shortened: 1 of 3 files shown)


chrisbobbe:
I've just rebased this past #5259. I'd appreciate a review! 🙂

gnprice (commented):
Thanks! Comments below.

chrisbobbe:
Thanks for the review! Revision pushed. 🙂

src/session/sessionReducer.js
```diff
@@ -24,6 +24,16 @@ import { getHasAuth } from '../account/accountsSelectors';
  * See {@link SessionState} for discussion of what "non-persistent" means.
  */
 export type PerAccountSessionState = $ReadOnly<{
+  /**
+   * The event queue ID that we're currently polling on, if any.
```
gnprice, line 28: I like this summary line! This seems like a good basis for thinking about what the behavior should be.

src/session/sessionReducer.js
```diff
@@ -24,6 +24,16 @@ import { getHasAuth } from '../account/accountsSelectors';
...
+   *
+   * Null when we're not polling on any event queue:
+   * - Between startup and registering a queue
+   * - After the server tells us our old queue was invalid, and before we've
+   *   registered a new one
+   * - While this account is logged out
+   * - While this account is logged in but is not the active account
```
gnprice, line 35: I don't think this item applies.

* At present, the only per-account state we have is the one for the active account. So "this account" here is always the active account.
* In a future where we have any per-account session state for accounts other than the active account, I think that will go together with maintaining an event queue for those accounts.

src/session/sessionReducer.js
```diff
@@ -162,26 +172,43 @@ export default (state: SessionState = initialState, action: Action): SessionStat
...
       };
 
     case LOGOUT:
       return {
         ...state,
         needsInitialFetch: false,
         loading: false,
+
+        // Although it may be valid, forget the event queue ID -- we know
+        // the user won't care about it.
```
gnprice, line 198: I'd frame this a bit differently:
```suggestion
        // Stop polling this event queue.
```

src/session/sessionReducer.js
```diff
@@ -162,26 +172,43 @@ export default (state: SessionState = initialState, action: Action): SessionStat
...
 
     case ACCOUNT_SWITCH:
       return {
         ...state,
         needsInitialFetch: true,
         loading: false,
+
+        // Although it may be valid, forget the old account's event queue --
+        // we know the user won't care about it. We also know we're about to
+        // request a new event queue; no use hanging on to the old one.
```
gnprice, line 210: Similarly:

```suggestion
        // Stop polling this event queue.  (We'll request a new one soon,
        // for the new account.)
```

gnprice (commented):
Thanks for the revision! Just small comments below.

chrisbobbe:

src/session/sessionReducer.js
gnprice: Yeah. Given that in that situation when the request eventually finishes, we're going to throw away the result without doing anything with it, I think it's still fair to say that we're not "currently polling on" that queue -- we do still have an HTTP request open, but as far as the app's logical behavior is concerned it already may as well not exist.

gnprice:
Thanks! Looks good; merging.

Issue #5162, opened 2021-12-09
ios: App unusable if "Prefer Cross-Fade Transitions" a11y setting is set


chrisbobbe:
See diagnosis at https://chat.zulip.org/#narrow/stream/48-mobile/topic/login.20bug/near/1294340, and see https://github.com/facebook/react-native/issues/29974.

I'm putting together a PR; filing this now so that I can give it a `Fixes:` line, as this was a pretty stubborn bug.

The `severe: crash` label is because the usable part of the screen went blank. The only way to undo that was to restart the app...then it would just happen again.

PR #5163, opened 2021-12-09
ios: Fix app-unusable bug with "Prefer Cross-Fade Transitions" setting

chrisbobbe:
Fork React Native's `KeyboardAvoidingView` and apply a fix.

Fixes: #5162

### DIFF
--- a/src/third/react-native/KeyboardAvoidingView.js
+++ b/src/third/react-native/KeyboardAvoidingView.js
@@ -1,0 +1,232 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * docs/THIRDPARTY file from the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const Keyboard = require('./Keyboard');
+const LayoutAnimation = require('../../LayoutAnimation/LayoutAnimation');
+const Platform = require('../../Utilities/Platform');
+const React = require('react');
+const StyleSheet = require('../../StyleSheet/StyleSheet');
+const View = require('../View/View');
+
+import type {ViewStyleProp} from '../../StyleSheet/StyleSheet';
+import {type EventSubscription} from '../../vendor/emitter/EventEmitter';
+import type {
+  ViewProps,
+  ViewLayout,
+  ViewLayoutEvent,
+} from '../View/ViewPropTypes';
+import type {KeyboardEvent} from './Keyboard';
+
+type Props = $ReadOnly<{|
+  ...ViewProps,
+
+  /**
+   * Specify how to react to the presence of the keyboard.
+   */
+  behavior?: ?('height' | 'position' | 'padding'),
+
+  /**
+   * Style of the content container when `behavior` is 'position'.
+   */
+  contentContainerStyle?: ?ViewStyleProp,
+
+  /**
+   * Controls whether this `KeyboardAvoidingView` instance should take effect.
+   * This is useful when more than one is on the screen. Defaults to true.
+   */
+  enabled: ?boolean,
+
+  /**
+   * Distance between the top of the user screen and the React Native view. This
+   * may be non-zero in some cases. Defaults to 0.
+   */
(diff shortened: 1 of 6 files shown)


chrisbobbe:
Oops, working out a bug; will un-mark-as-draft when finished 🙂

src/third/react-native/KeyboardAvoidingView.js
```diff
@@ -0,0 +1,251 @@
...
+        // ), then the keyboard fades in and out instead of sliding. In the
+        // off-screen state, the keyboard-frame values (`height`, `width`,
+        // `screenX`, and `screenY`) all take the special value of zero.
+        //
+        // So, recognize that special representation as a keyboard-closed
+        // state. (We choose `height` for the conditional because it seems
+        // least likely to collide with some valid non-off-screen state.)
+        // See zulip/zulip-mobile#5162 and facebook/react-native#29974.
+        return 0;
+    }
```
gnprice, line 102: I think this can explain much more directly why this logic is correct. In particular the correctness of it doesn't depend on those empirical observations (always a bit concerning of a thing to have to rely on) or on very much detail at all.

Then the explanation of why it's needed comes after that. The possibility that it's redundant is lower-stakes than the possibility that it's incorrect, so the reader can be more relaxed about it.

Here's one version:
```suggestion
    if (keyboardFrame.height === 0) {
      // The keyboard is hidden, and occupies no height.
      //
      // This condition is needed because in some circumstances when the
      // keyboard is hidden we get both `screenY` and `height` (as well as
      // `screenX` and `width`) of zero.  In particular, this happens when
      // the UIAccessibilityPrefersCrossFadeTransitions setting is true,
      // i.e. when the user has enabled both "Reduce Motion" and "Prefer
      // Cross-Fade Transitions" under Settings > Accessibility > Motion.
      // See zulip/zulip-mobile#5162 and facebook/react-native#29974.
    }
```

(Another thing there is that I think saying how to find the setting in the UI is more helpful than that API docs page; we don't want to have any code looking for that setting in any case, at least not in connection with this logic. The page is easy to find via Google given the identifier, anyway.)

src/third/react-native/KeyboardAvoidingView.js
```diff
@@ -0,0 +1,251 @@
+/* @flow strict-local */
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
```
gnprice, line 6: This should be updated; for one thing, with this file in a different source tree, there is a LICENSE file in the root directory but it's a different license.

Relatedly, we'll want to add an entry in `docs/THIRDPARTY`.

I can make those edits before merging.

gnprice: In the commit message, let's mention the exact RN version, and path in its tree, that this file is copied from. That makes it easy to compare, and also to look for upstream changes in the future.

src/third/react-native/KeyboardAvoidingView.js
gnprice: Oh, and this component is written to be cross-platform, right? We only use it on iOS, but the conditional in `componentDidMount` is meant to operate on Android too.

So this description of the relevant circumstance should mention iOS. Adding "on iOS" after "In particular, this happens" should do it.

gnprice (commented):
Thanks @chrisbobbe for getting to the bottom of this! This code looks good; comments below on a comment, and on some license metadata.

chrisbobbe:
Thanks!

> This should be updated; for one thing, with this file in a different source tree, there is a LICENSE file in the root directory but it's a different license.
> 
> Relatedly, we'll want to add an entry in `docs/THIRDPARTY`.
> 
> I can make those edits before merging.

Great, thank you!

src/third/react-native/KeyboardAvoidingView.js
chrisbobbe: > I can make those edits before merging.

Cool, thanks!

gnprice:
OK, and done! Thanks again for fixing this.

Issue #5156, opened 2021-12-07
"Mark all as read" button for PMs


chrisbobbe:
Tim has reported a few times ([here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Mark.20as.20read.20bugs/near/1293359) and [here](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Mark.20as.20read.20in.20PMs/near/1252998)) a bug where there's no "mark all as read" button for a PM conversation. We should add one.

There isn't an API endpoint for marking a whole PM conversation as read (see [chat](https://chat.zulip.org/#narrow/stream/378-api-design/topic/mark.20PM.20narrow.20as.20read/near/1253754)). But Wesley [points out](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Mark.20as.20read.20in.20PMs/near/1253769) that a workaround is to send a list of message IDs. I think that's likely to work in almost all cases, so let's try it.

The most complete list of message IDs we have in the app is found in the unread-messages data, right? A small caveat is that the window included in the initial unreads data is long but not infinite. There could be some missed unreads from before that window that we don't include, so they don't get marked as read.

The following comment from 89cff7c24 suggests that the window really is pretty long:

```js
      // TODO handle MESSAGE_FETCH_COMPLETE here.  This rarely matters, but
      //   could in principle: we could be fetching some messages from
      //   before the (long) window included in the initial unreads data.
      //   For comparison, the webapp does handle this case; see the call to
      //   message_util.do_unread_count_updates in message_fetch.js.
```

timabbott:
I think for all practical purposes, the `unread_msgs` data set should have the complete set of unread message IDs; it'll be the most recent 50K message IDs that are unread; the chances that a user is looking at a PM thread that contains unreads that are older than 50K other unreads are very low (and in that situation, the solution is probably a top-level notice to the user in the UI, not a change to this feature).  So worth a comment that this is theoretically not perfect, but certainly not worth spending more time on than that.

The key thing is to grab the set of IDs to send to the server from the unread messages data set, rather than just what messages are displayed locally.

gnprice:
That strategy sounds good to me.

Once we implement the TODO in that comment quoted above, then the unreads data (what we initialize from `unread_msgs`) will also contain any messages you're looking at now that are unread.

[...]

PR #5164, opened 2021-12-09
msglist: Offer bulk-mark-as-read button for PM conversations

chrisbobbe:
Fixes: #5156

### DIFF
--- a/src/chat/MarkAsReadButton.js
+++ b/src/chat/MarkAsReadButton.js
@@ -8,11 +8,13 @@ import { createStyleSheet } from '../styles';
 import { useSelector } from '../react-redux';
 import { ZulipButton } from '../common';
 import * as api from '../api';
-import { getAuth, getStreams } from '../selectors';
+import { getAuth, getStreams, getOwnUserId } from '../selectors';
+import { getUnread, getUnreadIdsForPmNarrow } from '../unread/unreadModel';
 import {
   isHomeNarrow,
   isStreamNarrow,
   isTopicNarrow,
+  isPmNarrow,
   streamNameOfNarrow,
   topicOfNarrow,
 } from '../utils/narrow';
@@ -34,6 +36,8 @@ export default function MarkAsReadButton(props: Props): Node {
   const { narrow } = props;
   const auth = useSelector(getAuth);
   const streams = useSelector(getStreams);
+  const unread = useSelector(getUnread);
+  const ownUserId = useSelector(getOwnUserId);
 
   const markAllAsRead = useCallback(() => {
     api.markAllAsRead(auth);
@@ -55,6 +59,19 @@ export default function MarkAsReadButton(props: Props): Node {
     }
   }, [auth, narrow, streams]);
 
+  const markPmAsRead = useCallback(() => {
+    // The message IDs come from our unread-messages data, which is
+    //   initialized with "only" the most recent 50K unread messages. So
+    //   we'll occasionally, but rarely, miss some messages here; see #5156.
+    const messageIds = getUnreadIdsForPmNarrow(unread, narrow, ownUserId);
+
+    if (messageIds.length === 0) {
+      return;
+    }
+
+    api.messagesFlags(auth, messageIds, 'add', 'read');
+  }, [auth, unread, narrow, ownUserId]);
+
   if (isHomeNarrow(narrow)) {
     return <ZulipButton style={styles.button} text="Mark all as read" onPress={markAllAsRead} />;
   }
@@ -68,6 +85,12 @@ export default function MarkAsReadButton(props: Props): Node {
   if (isTopicNarrow(narrow)) {
     return (
       <ZulipButton style={styles.button} text="Mark topic as read" onPress={markTopicAsRead} />
+    );
+  }
+
+  if (isPmNarrow(narrow)) {
+    return (
+      <ZulipButton style={styles.button} text="Mark conversation as read" onPress={markPmAsRead} />
     );
   }
 
--- a/src/api/messages/messagesFlags.js
+++ b/src/api/messages/messagesFlags.js
@@ -9,7 +9,7 @@ export type ApiResponseMessagesFlags = {|
 
 export default (
   auth: Auth,
-  messages: number[],
+  messages: $ReadOnlyArray<number>,
   op: string,
   flag: string,
 ): Promise<ApiResponseMessagesFlags> =>
--- a/src/unread/unreadModel.js
+++ b/src/unread/unreadModel.js
@@ -2,6 +2,9 @@
 import Immutable from 'immutable';
 import invariant from 'invariant';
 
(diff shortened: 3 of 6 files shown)


chrisbobbe:
Also posted [a thing I noticed while working on this](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60pmNarrowInternal.60's.20jsdoc/near/1295166).

src/chat/MarkAsReadButton.js
```diff
@@ -55,6 +62,23 @@ export default function MarkAsReadButton(props: Props): Node {
...
 
+  const markPmAsRead = useCallback(() => {
+    const userIds = userIdsOfPmNarrow(narrow);
+    let messageIds = undefined;
+    const key = pmUnreadsKeyFromPmKeyIds(userIds, ownUserId);
+    if (userIds.length > 1) {
+      messageIds = unreadHuddles.find(x => x.user_ids_string === key)?.unread_message_ids ?? [];
+    } else {
+      messageIds = unreadPms.find(x => x.sender_id.toString() === key)?.unread_message_ids ?? [];
+    }
```
gnprice, line 73: Yikes, this is kind of messy, isn't it. Particularly that gotcha commented on in the first commit.

I'd like to have this kind of code, dealing with the messy structure inside a data model, be located inside that model's code. So e.g. a `getUnreadIdsForPmNarrow`, located among the selectors at the top of `unreadModel.js`.

There's an existing spot where we do something similar, in `getUnreadCountForNarrow` in `unreadSelectors.js`:
```js
      pm: ids => {
        if (ids.length > 1) {
          const unreadsKey = pmUnreadsKeyFromPmKeyIds(ids, ownUserId);
          const unreadItem = unread.huddles.find(x => x.user_ids_string === unreadsKey);
          return unreadItem?.unread_message_ids.length ?? 0;
        } else {
          const senderId = ids[0];
          const unreadItem = unread.pms.find(x => x.sender_id === senderId);
          return unreadItem?.unread_message_ids.length ?? 0;
        }
      },
```

So a good structure would be to factor that logic out of there, and then that case of `getUnreadCountForNarrow` just looks like `ids => getUnreadIdsForPmNarrow(ids).length`.

Note that that also has a variation of how to deal with the string/number thing, that's perhaps a bit cleaner.

src/chat/MarkAsReadButton.js
```diff
@@ -55,6 +62,23 @@ export default function MarkAsReadButton(props: Props): Node {
...
+
+    // The message IDs come from our unread-messages data, which is
+    //   initialized with "only" the most recent 50K unread messages. So
+    //   we'll occasionally, but rarely, miss some messages here; see #5156.
+    // TODO(?): Should we use `queueMarkAsRead` here?
+    api.messagesFlags(auth, messageIds, 'add', 'read');
```
gnprice, line 79: Should probably skip this if `messageIds` is empty.

src/chat/MarkAsReadButton.js
```diff
@@ -55,6 +62,23 @@ export default function MarkAsReadButton(props: Props): Node {
...
+    if (userIds.length > 1) {
+      messageIds = unreadHuddles.find(x => x.user_ids_string === key)?.unread_message_ids ?? [];
+    } else {
+      messageIds = unreadPms.find(x => x.sender_id.toString() === key)?.unread_message_ids ?? [];
+    }
+
+    // The message IDs come from our unread-messages data, which is
+    //   initialized with "only" the most recent 50K unread messages. So
+    //   we'll occasionally, but rarely, miss some messages here; see #5156.
+    // TODO(?): Should we use `queueMarkAsRead` here?
```
gnprice, line 78: No need for that, I think; the job of that queue is when you're scrolling and have lots of frequent little batches to mark as read.

gnprice (commented):
Thanks @chrisbobbe! Comments below.

src/chat/MarkAsReadButton.js
chrisbobbe: > Note that that also has a variation of how to deal with the string/number thing, that's perhaps a bit cleaner.

Yeah, maybe. It's not a biggie, but that one doesn't use `pmUnreadsKeyFromPmKeyIds` at all in the 1:1 pm case, even though it's written to be able to handle it.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks for the revision! Looks good. One nit:
```diff
-  invariant(isPmNarrow(narrow));
```

This gets added with the surrounding code, then removed in the subsequent commit. Did you intend to take it out?

(I think it's fine to either have it or not have it; the next line will throw an error if it's not a PM narrow anyway.)

chrisbobbe:
> Thanks for the revision! Looks good. One nit:
> 
> ```diff
> -  invariant(isPmNarrow(narrow));
> ```
> 
> This gets added with the surrounding code, then removed in the subsequent commit. Did you intend to take it out?
> 
> (I think it's fine to either have it or not have it; the next line will throw an error if it's not a PM narrow anyway.)

Eep, thanks for catching this! Yeah, I think I was doing a complicated reordering of commits, and this fell through. Agreed that we don't need the invariant here.

Small correction: at the commit that includes that invariant line, the function (`getUnreadIdsForPmNarrow`) would actually error unconditionally. I've sent https://github.com/zertosh/invariant/pull/48 for that. 😝 I'm not sure `invariant` is actively maintained; if you'd like, I'll make a little libdef for it with that fix.

chrisbobbe:
Merged, with the invariant line taken out.

gnprice:
> the function (`getUnreadIdsForPmNarrow`) would actually error unconditionally. I've sent [zertosh/invariant#48](https://github.com/zertosh/invariant/pull/48) for that.

Ha, I see.

> I'm not sure `invariant` is actively maintained; if you'd like, I'll make a little libdef for it with that fix.

Could try that. Though I'm not sure a libdef (or that `.js.flow` file) will actually have any effect… because Flow actually special-cases calls to `invariant`, by name:
https://github.com/facebook/flow/blob/d247c5eef542a1745d831edeab4e11e305effaf4/src/typing/statement.ml#L221-L223

So this is likely a bug in that logic in Flow… that or a divergence between the `invariant` NPM package and the Facebook-internal `invariant`. The way to find out which probably runs through filing an issue in Flow.

Related Flow issue, with background on why it's special: https://github.com/facebook/flow/issues/ 112

From back-links in that thread, a possible form of a solution if the NPM package and what's inside Facebook have diverged (other than the solution of making the package align with the Facebook version):
https://github.com/facebook/flow/commit/3e35bd475435a4fd282939be21ebeb3a6fab7106

I haven't really been bothered by the fact that `invariant` is special-cased, because the name is specific enough that it doesn't feel likely to be accidentally used for something unrelated. But if the special-cased behavior is wrong, it'd certainly be good to fix it.

PR #5167, opened 2021-12-13
deps: Upgrade RN to v0.64.3, from v0.64.2

chrisbobbe:
This is just a minor version that came out a month ago, which was
after our big v0.64 upgrade, #4991.

The changelog entry [1] has just one item:

  ### Fixed

  - For Android, general fixes to Appearance API and also fixes
    AppCompatDelegate.setDefaultNightMode(). For iOS, now works
    correctly when setting window.overrideUserInterfaceStyle
    ([25a2c608f7](https://github.com/facebook/react-native/commit/25a2c608f790f42cbc4bb0a90fc06cc7bbbc9b95)
    by [@mrbrentkelly](https://github.com/mrbrentkelly))

We don't use those mentioned APIs, so this should be fine and not
disruptive.

There are no changes to the template app.

Tested on my iPhone 13 Pro running iOS 15, and on the office Android
device (Samsung Galaxy S9 running Android 9), and I didn't encounter
any building or running failures, and nothing was obviously wrong in
the app.

[1] https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0643

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -30,14 +30,14 @@ PODS:
     - UMCore
   - EXWebBrowser (9.2.0):
     - UMCore
-  - FBLazyVector (0.64.2)
-  - FBReactNativeSpec (0.64.2):
+  - FBLazyVector (0.64.3)
+  - FBReactNativeSpec (0.64.3):
     - RCT-Folly (= 2020.01.13.00)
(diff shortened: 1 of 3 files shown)


gnprice:
Thanks! Merging.

PR #5131, opened 2021-11-17
api types: Sync RealmDataForUpdate with the recent update of the event doc

chrisbobbe:
In particular, zulip/zulip@864fb7c97 and some previous commits.

See discussion at
  https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60value.60.20in.20realm.20op.3A.20update/near/1282692.

PR #5168, opened 2021-12-15
legal: Link to Zulip's terms and terms for the active realm

chrisbobbe:
And do a few cleanups/improvements along the way, as usual. 🙂

Greg, I'd really appreciate your help resolving a Flow error in the last commit. It would unlock the ability to bold the realm name where it appears inside a string we want translated.

### DIFF
--- a/src/events/eventToAction.js
+++ b/src/events/eventToAction.js
@@ -132,9 +132,30 @@ export default (state: PerAccountState, event: $FlowFixMe): EventAction | null =
         messageIds: event.message_ids ?? [event.message_id],
       };
 
+    case EventTypes.realm:
+      return {
+        type: EVENT,
+        event:
+          /* prettier-ignore */
+          event.op === 'update'
+            // Convert to an equivalent `update_dict` event, so reducers only have
+            //   to handle that one form.
+            // TODO: handle `extra_data` hack property in the `update`
+            //   event, as long as servers still send it
+            ? {
+                id: event.id,
+                type: EventTypes.realm,
+                op: 'update_dict',
+                property: 'default',
+                data: {
+                  [event.property]: event.value,
+                },
+              }
+            : event,
+      };
+
     case EventTypes.restart:
     case EventTypes.stream:
-    case EventTypes.realm:
       return {
         type: EVENT,
         event,
--- a/src/settings/LegalScreen.js
+++ b/src/settings/LegalScreen.js
@@ -6,9 +6,9 @@ import type { Node } from 'react';
 import type { RouteProp } from '../react-navigation';
 import type { AppNavigationProp } from '../nav/AppNavigator';
 import { useSelector } from '../react-redux';
-import { Screen, NestedNavRow } from '../common';
+import { Screen, NestedNavRow, RawLabel } from '../common';
 import { openLinkEmbedded } from '../utils/openLink';
-import { getRealmUrl } from '../selectors';
+import { getRealmUrl, getRealmName } from '../selectors';
 
 type Props = $ReadOnly<{|
   navigation: AppNavigationProp<'legal'>,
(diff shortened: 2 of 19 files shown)


src/events/eventToAction.js
```diff
@@ -132,9 +132,28 @@ export default (state: PerAccountState, event: $FlowFixMe): EventAction | null =
...
       };
 
+    case EventTypes.realm:
+      return {
+        type: EVENT,
+        event:
+          /* prettier-ignore */
+          event.op === 'update'
+            // Convert to an equivalent update_dict event, so reducers only have
+            // to handle update_dict.
```
gnprice, line 142: There's also that one property for which the `update` form has a hack where it provides a second property at the same time. If we're not handling that now (which we probably aren't), don't need to handle it in this change, but should leave a TODO comment to do so.

src/settings/LegalScreen.js
```diff
@@ -18,19 +18,35 @@ type Props = $ReadOnly<{|
...
+        // the org itself (e.g., orgs on Zulip Cloud), as this text might be
+        // read to imply. But:
+        // - We don't want to complicate the wording. Not everyone knows
+        //   what a server is.
+        // - These terms will often differ from Zulip's own terms (the ones
+        //   at the other link), and we have to include them.
+        // - These terms will apply to all users in the org, in all cases.
+        label={{
+          text: 'Terms for {realmName}',
+          values: { realmName: <RawLabel style={{ fontWeight: 'bold' }} text={realmName} /> },
```
gnprice, line 46: Neat!

From that error message, it looks like the relevant bit of the libdef is here:
```js
  declare export interface IntlFormatters<T = React$Node, R = T> {
// …
    formatMessage(
      descriptor: MessageDescriptor,
      // The `+` was added to make the properties covariant rather
      // than invariant, something TypeScript can't do.
      values?: {| +[key: string]: PrimitiveType | FormatXMLElementFn<string, string> |},
    ): string;
    formatMessage(
      descriptor: MessageDescriptor,
      // The `+` was added to make the properties covariant rather
      // than invariant, something TypeScript can't do.
      values?: {| +[key: string]: PrimitiveType | React$Node | FormatXMLElementFn<T, R> |},
    ): React$Node;
```

So there's an overload -- two signatures for the same method `formatMessage`. Up to now we've used the first one, and now we want to use the second one.

I suspect a key piece of why there's an error here is that, via our `GetText` type, we're saying that our `formatMessage` call (in `TranslationProvider.js`) should return string. From examining those overloaded definitions, one can see that that's not what `formatMessage` returns anymore if you pass it a value that's a `React$Node` -- in that case it returns another `React$Node`.

I'll fiddle with it a bit and see what I find.

gnprice (commented):
Thanks! And that's very promising for formatting text in the UI.

I'll spend a few minutes seeing what I can do with the types. One small other comment below.

gnprice:
OK, pushed a couple of commits that I think sort out the types. We end up needing a distinction between two flavors of `LocalizableText`: one to use in a React context, and one to pass to `_` to get a plain string out (e.g. for action sheets or toasts.) Please take a look!

Then some of those changes should go in prep commits, and other parts squashed into the other commits in the branch.

chrisbobbe:
Thanks a lot!!

Writing this down at top-level because it looks like GitHub's iOS app doesn't support commenting directly on lines in a diff: I need to fix the selector for getting the realm description so it gets the realm description, not the realm name.

chrisbobbe:
Note that I haven't yet been able to test the live-updating of the realm name and description. I haven't had time to set up a dev server with Docker lately. I tried to test with the description on CZO by finding a typo and trying to fix it, but the server errored; discussion at https://chat.zulip.org/#narrow/stream/3-backend/topic/502.20on.20updating.20realm.20description/near/1297960.

chrisbobbe:
Pushed a new revision with the `realm_name` -> `name` rename in the realm-update events from #5131.

gnprice:
Thanks! Looks good. Merging with one comment tweak:
```diff
--- a/src/settings/LegalScreen.js
+++ b/src/settings/LegalScreen.js
@@ -35,7 +35,7 @@ export default function LegalScreen(props: Props): Node {
         // These are really terms set by the server admin responsible for
         // hosting the org, and that server admin may or may not represent
         // the org itself, as this text might be read to imply. (E.g.,
-        // that's the case for orgs on Zulip Cloud.) But:
+        // on Zulip Cloud they don't.) But:
```

chrisbobbe:
Thanks!

PR #5172, opened 2021-12-17
nav [nfc]: Delete empty file ConfiguredNavigationContainer.js

gnprice:
This was added in 12284502b, but never referred to there or
elsewhere.  Presumably it was from an earlier draft of that change.

With this change (and after our recent work to apply types to
redux-persist), a grep confirms that all our non-test JS source
is typed:

```
  $ git grep -L '@flow strict' src/**/*.js | grep -v '/__tests__/'
  $
```


chrisbobbe:
Eep, thanks for catching this! Merged.

PR #5173, opened 2021-12-18
recipient [nfc]: Preserve PmKeyRecipients through PM narrows

gnprice:
And to a few more places that that enables.


chrisbobbe:
Thanks, LGTM! Merged.

Issue #5171, opened 2021-12-16
Handle stricter length limit on Android toasts when targeting Android 12.


chrisbobbe:
Here's an item that stands out in the list of behavior changes to watch out for when targeting Andriod 12 for the first time:

["Toast redesign"](https://developer.android.com/about/versions/12/behavior-changes-12#toast-redesign), specifically the new limit to two lines of text in a toast. We use `react-native-simple-toast`, which uses the native toast on Android.

One of our longer toast messages without a variable-length component (like a file name, server error, group-PM participants, etc.) looks like this:

![image](https://user-images.githubusercontent.com/22248748/146462435-4ce1ca8a-c5b1-493e-bf73-46f89a772656.png)

Here's how something longer gets cut off:

![image](https://user-images.githubusercontent.com/22248748/146462605-6f4a5be4-df0e-425a-862d-a6e7d711b5c9.png)

Things to fix:

- [ ] For group-PM messages, there isn't always room in message headers (in the message list) to print all the participants' names. So we offer a toast that's meant to show all the names. That toast will be much worse at its job after this change. We shouldn't just be using full names anyway; sometimes I need more than that to verify that I'm talking to the person I think I am. I may want to see an avatar, email address, and so on. I think we should just link to the group-PM participants screen, which offers a nice scrollable list of user items, where each item has the user's full name, email, and avatar, and it's tappable to get to the full profile.
- [ ] `Uploading {fileName} failed.` should become something like `Failed to upload file: {fileName}`. The "failed" part is really important, and we can't have it get cut off just because a file name is very long.
- (We sometimes put server errors in toasts. I'm not sure how long those can get...probably servers should convey what's most important in a few words at the start of the message?)

chrisbobbe:
For the PM header long-press toast, Greg and I just discussed raising an action sheet on long press instead, like we do for topic recipient headers.

The list of full names makes sense as a title of that action sheet (I'd like to [localize it](https://formatjs.io/docs/react-intl/api/#formatlist); we may have to add some `Intl` polyfills), as long as it can't defeat any interactions by being too long. It's not a problem if the way to achieve that is by cutting off part of the text—we'll add a button to the action sheet that takes you to the full list of recipients, with avatars, emails, links to profiles, etc.

I've experimented with the title being a list of 100 names: "Firstname Lastname0", "Firstname Lastname1", etc., and with a few more buttons that the current one we're planning to add.

[...]

Issue #5101, opened 2021-11-04
Target Android 12 (set targetSdkVersion to 31), by 2022-11 deadline


gnprice:
This is the successor to #4284 (and #3665 and #3563 and #3075). We should update our `targetSdkVersion` to 31, meaning Android 12.

Based on [the last few years' experience](https://support.google.com/googleplay/android-developer/answer/113469#targetsdk), Google Play will probably set a deadline for this of late 2022. It's been the first business day of November in each of the last few years.

The important steps for this upgrade are:
1. Read about the [potentially-breaking changes](https://developer.android.com/about/versions/12/behavior-changes-12), and identify those that might affect us.
2. Make a WIP change to `targetSdkVersion`.
3. Test the WIP change thoroughly, especially in the areas highlighted in step 1. Fix things as needed, and repeat.

We aren't aware of any blockers to this, so we can try doing it any time. It'd be nice to do it well before the deadline -- or at least to go far enough to learn about any unknown blockers.

gnprice:
Well, here's one blocker:
  https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines
We'll need to replace our `NotificationIntentService`, which we use when you tap a notification to open it, with directly opening an Activity. (We learned about this from a lint warning in Android Studio: https://github.com/zulip/zulip-mobile/pull/5121#discussion_r747107709)
* Ideally this would be our `MainActivity`. We'll need to play with things somewhat to try to get the correct behavior in all three cases of the app being in the foreground, in the background, and not running. (Currently `NotificationIntentService` has `NotifyReact.kt` do this work.)
  * The main advice from Android upstream won't work for us verbatim because it assumes the canonical Android model where basically every screen in your app is a distinct activity. That structure isn't even the unanimous advice of Android upstream these days -- some of their voices say you should have one or a few activities and use `Fragment`s for navigation instead. But in any case it's not our structure of things using RN, and we have just one activity where the app's whole UI (apart from the notifications) lives. Hence the need to play with things to find a solution.
* If we can't make that work, it should certainly be possible to make a trampoline activity, analogous to our `ShareToZulipActivity`. That sure won't be in the spirit of what they're going for here, though. It may also cause visible jank in the UI, like an extra transition.

chrisbobbe:
A few items stand out in the list of breaking changes. Blank checkboxes are things we need to fix before targeting Android 12.

[...]

PR #5174, opened 2021-12-21
android: Update target SDK version to 31, i.e., Android 12

chrisbobbe:
Fixes: #5101
Fixes: #5171


chrisbobbe:
Please also see #5175 for some further improvements to toast texts, and more.

src/action-sheets/index.js
```diff
@@ -587,3 +624,53 @@ export const showStreamActionSheet = ({
     }),
   );
 };
+
+export const showPmActionSheet = ({
```
gnprice, line 628: Really more like "PM *conversation* action sheet", right? For an individual PM, there's the message action sheet.

gnprice: (Similarly for the function constructing the buttons.)

src/compose/ComposeBox.js
```diff
@@ -259,11 +259,11 @@ class ComposeBoxInner extends PureComponent<Props, State> {
         try {
           response = await api.uploadFile(auth, attachments[i].uri, fileName);
         } catch (e) {
-          showToast(_('Uploading {fileName} failed.', { fileName }));
+          showToast(_('Failed to upload file: {fileName}.', { fileName }));
           this.setMessageInputValue(
             this.state.message.replace(
               placeholder,
-              `[${_('Uploading {fileName} failed.', { fileName })}]()`,
+              `[${_('Failed to upload file: {fileName}.', { fileName })}]()`,
```
gnprice, line 266: Let's also drop the `.` at the end. That works when the filename is in the middle of the sentence, but here it risks looking like part of the filename; and in general I think we don't put periods at the end of these short messages.

src/chat/PmConversationDetailsScreen.js
```diff
@@ -15,11 +15,14 @@ import { navigateToAccountDetails } from '../actions';
...
 type Props = $ReadOnly<{|
-  navigation: AppNavigationProp<'group-details'>,
-  route: RouteProp<'group-details', {| recipients: PmKeyRecipients |}>,
+  navigation: AppNavigationProp<'pm-conversation-details'>,
+  route: RouteProp<'pm-conversation-details', {| recipients: PmKeyRecipients |}>,
 |}>;
 
-export default function GroupDetailsScreen(props: Props): Node {
+/**
+ * Details about a PM conversation, identified by PM key recipients.
```
gnprice, line 23: The second half of this is redundant with the type. Best to leave it out -- that way it doesn't become wrong if we change the type in the future (renaming it, or changing to something like `PmNarrow`, or something else.)

gnprice (commented):
Thanks for taking care of this! Looks great; small comments below.

chrisbobbe:
Thanks! Revision pushed.

gnprice:

chrisbobbe:
Thanks!

PR #5175, opened 2021-12-21
Fix bad text in toasts; give `RawLabel` and `Label` better names

chrisbobbe:
See our capitalization [policy](https://zulip.readthedocs.io/en/latest/translating/translating.html#capitalization) to explain some of the proposed text changes here.

When a toast has two full ideas, I've punctuated them as sentences, like

> Download failed. Please try again from your browser.

Otherwise, I've left out any punctuation, to leave the message quick and not too prose-like. I'm not sure if that's exactly the choice we want.

Greg, I'd be happy for you to make any improvements/fixes to the user-facing text before merge without another round of review. I mainly just wanted to clean up the more glaring errors in these.


src/lightbox/LightboxActionSheet.js
```diff
@@ -36,7 +36,7 @@ type ButtonType = {|
 const tryToDownloadImage = async ({ src, auth }: DownloadImageType) => {
   const tempUrl = await api.tryGetFileTemporaryUrl(src, auth);
   if (tempUrl === null) {
-    showToast('Please download the image from your browser');
+    showToast('Download failed. Please try again from your browser.');
```
gnprice, line 39: Hmm, is "download failed" accurate here? Or is it more like "one form of download wasn't available, so we'll use another"?

I looked at our jsdoc for this API method, and the one it calls, and didn't come out with a clear answer. :-/ It seems this API route isn't documented. But I *think* the story is basically that on a server that doesn't have this newer feature, we'll just always use this path for the download instead of the other path.

src/lightbox/shareImage.js
```diff
@@ -33,10 +34,10 @@ export default async (url: string, auth: Auth) => {
       try {
         await Share.share({ url: uri, message: url });
       } catch (error) {
-        showToast('Can not share');
+        showToast('Share failed');
       }
     } catch (error) {
-      showToast('Can not download');
+      showToast('Share failed');
     }
```
gnprice, line 41: Making this about "sharing" in both cases makes sense.

This also means we can drop the inner try/catch -- it's now redundant with the outer one.

src/lightbox/LightboxActionSheet.js
gnprice: I kind of think the right thing here may be to not have a toast at all. We're opening the file directly in a browser. The user will see it there, and if they want it further "downloaded" they can find the "download" option in their browser. (… Or if they can't, this toast isn't likely to help them.)

src/lightbox/shareImage.js
```diff
@@ -9,11 +9,12 @@ import * as api from '../api';
 import { openLinkEmbedded } from '../utils/openLink';
 import * as logging from '../utils/logging';
 
+// TODO(i18n): Wire up toasts for translation.
 export default async (url: string, auth: Auth) => {
   const tempUrl = await api.tryGetFileTemporaryUrl(url, auth);
 
   if (tempUrl === null) {
-    showToast('Please share the image from your browser');
+    showToast('Share failed. Please try again from your browser.');
```
gnprice, line 17: Hmm, here it's definitely not that sharing failed. I'm not sure the existing text is great, but I feel like this version might be more confusing.

What's really happening here is:
* The server couldn't give us a link we can use directly.
* So instead we're sending the user to the file in a browser, and inviting them to use the "share" feature of the browser.
* That's probably not something they expected when they hit "share" -- and when it does open a browser, it's likely not obvious what that has to do with the sharing they wanted to do.
* So we're giving them some kind of indication of "yes, we heard you, you want to share the file -- this is the path we're offering for doing that".

gnprice (commented):
Thanks! Small comments below.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

I'll also do a quick tools/tx-sync and we'll see what that Transifex UI looks like.

PR #5186, opened 2022-01-08
Mark a bunch more object and array types read-only

gnprice:
Including all the object and array types in our actions and other Redux-related types, and a bunch of miscellaneous others.

The lack of read-only markings on the type for (Redux actions for) `update_message` events, `EventUpdateMessageAction`, produced some very confusing Flow errors in tests when I started changing those, sending me down a path of at least a half-hour of debugging. (It was a real type error, it turned out -- a missing `subject` property, which for now our types believe is required there.)

And it turns out I had a draft branch from last summer with that and the other action types, and a few more changes. So, here's that branch finished up, along with some further changes in the same direction.

### DIFF
--- a/flow-typed/redux_v4.x.x.js
+++ b/flow-typed/redux_v4.x.x.js
@@ -17,7 +17,7 @@ declare module 'redux' {
 
   declare export type DispatchAPI<A> = (action: A) => A;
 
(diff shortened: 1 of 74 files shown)


chrisbobbe:
Great, thanks a lot! LGTM, merged.

